PR #4975, opened 2021-03-04
CBG-1322: Avoid revID update when only user xattr is changed

JRascagneres:
In the case where an import is triggered by a change of user xattrs we need to avoid the generation of a new revid and only increment the sequence.

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -817,7 +817,7 @@ func (db *Database) Put(docid string, body Body) (newRevID string, doc *Document
 	delete(body, BodyRevisions)
 
 	allowImport := db.UseXattrs()
-	doc, newRevID, err = db.updateAndReturnDoc(newDoc.ID, allowImport, expiry, nil, func(doc *Document) (resultDoc *Document, resultAttachmentData AttachmentData, avoidRevIDGeneration bool, updatedExpiry *uint32, resultErr error) {
+	doc, newRevID, err = db.updateAndReturnDoc(newDoc.ID, allowImport, expiry, nil, func(doc *Document) (resultDoc *Document, resultAttachmentData AttachmentData, newRevIDGenerated bool, updatedExpiry *uint32, resultErr error) {
 
 		var isSgWrite bool
 		var crc32Match bool
@@ -919,7 +919,7 @@ func (db *Database) PutExistingRevWithConflictResolution(newDoc *Document, docHi
 	}
 
 	allowImport := db.UseXattrs()
-	doc, _, err = db.updateAndReturnDoc(newDoc.ID, allowImport, newDoc.DocExpiry, nil, func(doc *Document) (resultDoc *Document, resultAttachmentData AttachmentData, avoidRevIDGeneration bool, updatedExpiry *uint32, resultErr error) {
+	doc, _, err = db.updateAndReturnDoc(newDoc.ID, allowImport, newDoc.DocExpiry, nil, func(doc *Document) (resultDoc *Document, resultAttachmentData AttachmentData, newRevIDGenerated bool, updatedExpiry *uint32, resultErr error) {
 		// (Be careful: this block can be invoked multiple times if there are races!)
 
 		var isSgWrite bool
@@ -1594,7 +1594,7 @@ func (db *Database) IsIllegalConflict(doc *Document, parentRevID string, deleted
 	return true
 }
 
-func (db *Database) documentUpdateFunc(docExists bool, doc *Document, allowImport bool, previousDocSequenceIn uint64, unusedSequences []uint64, callback updateAndReturnDocCallback, expiry uint32) (retSyncFuncExpiry *uint32, retNewRevID string, retStoredDoc *Document, retOldBodyJSON string, retUnusedSequences []uint64, changedAccessPrincipals []string, changedRoleAccessUsers []string, importFromUserXattrs bool, err error) {
+func (db *Database) documentUpdateFunc(docExists bool, doc *Document, allowImport bool, previousDocSequenceIn uint64, unusedSequences []uint64, callback updateAndReturnDocCallback, expiry uint32) (retSyncFuncExpiry *uint32, retNewRevID string, retStoredDoc *Document, retOldBodyJSON string, retUnusedSequences []uint64, changedAccessPrincipals []string, changedRoleAccessUsers []string, newRevIDGenerated bool, err error) {
 
 	err = db.validateExistingDoc(doc, allowImport, docExists)
 	if err != nil {
@@ -1602,7 +1602,7 @@ func (db *Database) documentUpdateFunc(docExists bool, doc *Document, allowImpor
 	}
 
 	// Invoke the callback to update the document and return a new revision body:
-	newDoc, newAttachments, importFromUserXattrs, updatedExpiry, err := callback(doc)
+	newDoc, newAttachments, avoidRevIDGeneration, updatedExpiry, err := callback(doc)
 	if err != nil {
 		return
 	}
@@ -1653,7 +1653,7 @@ func (db *Database) documentUpdateFunc(docExists bool, doc *Document, allowImpor
 		return
 	}
 
-	if doc.CurrentRev != prevCurrentRev || importFromUserXattrs {
+	if doc.CurrentRev != prevCurrentRev || avoidRevIDGeneration {
 		// Most of the time this update will change the doc's current rev. (The exception is
 		// if the new rev is a conflict that doesn't win the revid comparison.) If so, we
 		// need to update the doc's top-level Channels and Access properties to correspond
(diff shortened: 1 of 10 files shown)


db/import.go
```diff
@@ -263,13 +259,30 @@ func (db *Database) importDoc(docid string, body Body, isDelete bool, existingDo
...
+		if !doc.BodyChanged(existingDoc.Body) {
+			doc.RemoveBody()
+			newDoc.UpdateBody(body)
+			if !wasStripped && !isDelete {
+				newDoc._rawBody = rawBodyForRevID
 			}
+			newDoc.RevID = doc.CurrentRev
+
+			return newDoc, nil, true, updatedExpiry, nil
 		}
```
bbrks: Wondering if this will be more maintainable to only regenerate rev id when the body changes instead. Otherwise we have 2 blocks of the same logic to keep in sync for both cases.

bbrks (changes requested):

db/import.go
JRascagneres: Yeah that actually makes more sense imo.

JRascagneres: We do still need to update newDoc.RevID though.

bbrks: Fair enough, but I think can still be covered by an else

JRascagneres: Yeah sure. Just pointing it out :)

JRascagneres:
Review revision cache Put operation. Will need to update this therefore need to figure out a way to update. Likely delete + Put

db/revision_cache_lru.go
```diff
@@ -181,6 +185,15 @@ func (rc *LRURevisionCache) Put(docRev DocumentRevision) {
 	value.store(docRev)
 }
 
+// Update a revision
+func (rc *LRURevisionCache) Update(docRev DocumentRevision) {
+	value := rc.getValue(docRev.DocID, docRev.RevID, false)
```
adamcfraser: As discussed, need to lock for both the get and remove to avoid races.  Can probably simplify and do less work than getValue is doing.

adamcfraser (approved):

db/crud.go
```diff
@@ -1762,7 +1764,7 @@ func (db *Database) updateAndReturnDoc(docid string, allowImport bool, expiry ui
 			}
 
 			docExists := currentValue != nil
-			syncFuncExpiry, newRevID, storedDoc, oldBodyJSON, unusedSequences, changedAccessPrincipals, changedRoleAccessUsers, err = db.documentUpdateFunc(docExists, doc, allowImport, docSequence, unusedSequences, callback, expiry)
+			syncFuncExpiry, newRevID, storedDoc, oldBodyJSON, unusedSequences, changedAccessPrincipals, changedRoleAccessUsers, avoidRevIDGeneration, err = db.documentUpdateFunc(docExists, doc, allowImport, docSequence, unusedSequences, callback, expiry)
```
adamcfraser: I don't think avoidRevIDGeneration makes sense as a variable name here - it sounds like revID generation is still pending, when in fact it's already been completed (or not completed) in documentUpdateFunc.  I think the only place it makes sense is inside documentUpdateFunc itself - for external usage, it's a flag that indicates whether a rev id was generated.

db/import.go
```diff
@@ -263,29 +259,41 @@ func (db *Database) importDoc(docid string, body Body, isDelete bool, existingDo
...
-			return nil, nil, updatedExpiry, err
-		}
-		base.DebugfCtx(db.Ctx, base.KeyImport, "Created new rev ID for doc %q / %q", base.UD(newDoc.ID), newRev)
-		// body[BodyRev] = newRev
-		newDoc.RevID = newRev
-		err := doc.History.addRevision(newDoc.ID, RevInfo{ID: newRev, Parent: parentRev, Deleted: isDelete})
-		if err != nil {
-			base.InfofCtx(db.Ctx, base.KeyImport, "Error adding new rev ID for doc %q / %q, Error: %v", base.UD(newDoc.ID), newRev, err)
-		}
+		avoidRevIDGeneration = !doc.BodyChanged(existingDoc.Body)
```
adamcfraser: Can 'bodyChanged' be returned by doc.IsSGWrite, to avoid computing the crc32c of the existing body twice?

adamcfraser (changes requested):

db/revision_cache_lru.go
```diff
@@ -181,6 +185,31 @@ func (rc *LRURevisionCache) Put(docRev DocumentRevision) {
...
+	// Add new value and overwrite existing cache key, pushing to front to maintain order
+	value := &revCacheValue{key: key}
+	rc.cache[key] = rc.lruList.PushFront(value)
+
+	// Purge oldest item if required
+	for len(rc.cache) > int(rc.capacity) {
+		rc.purgeOldest_()
+	}
+
+	value.store(docRev)
```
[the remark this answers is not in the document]
JRascagneres: Don't think so actually. Will move out.

db/import.go
JRascagneres: I'm not sure this would improve a lot. Its not quite as simple as we're always computing the hash inside of IsSGWrite because in the case of an xattr change we currently only compute the body hash once using doc.BodyChanged and won't check the body at all inside of IsSGWrite. If we were to return BodyChanged from IsSGWrite we would need to always compute the body hash regardless of what happened to the xattr meaning every time we run IsSGWrite we're going to run this hash check. IsSGWrite gets run a few times for an import operation. I think the question would be in the case of non-xattr imports do we want to pay the cost of an additional hash or in the case of an xattr import do we want to pay the cost of multiple body hash checks most of which are 'wasted'.

JRascagneres:
Conflicts and updates have been done but don't appear here. Seems like GitHub having issues.

db/crud.go
```diff
@@ -1862,7 +1864,12 @@ func (db *Database) updateAndReturnDoc(docid string, allowImport bool, expiry ui
 			Deleted:          doc.History[newRevID].Deleted,
 			_shallowCopyBody: storedDoc.Body(),
 		}
-		db.revisionCache.Put(documentRevision)
+
+		if revIDGenerationSkipped {
+			db.revisionCache.Upsert(documentRevision)
```
adamcfraser, line 1874: We're upserting the rev cache after the document has been persisted to the bucket.  I think this means there's a new race here, in the (unlikely but possible) scenario where a client is notified of the new change and gets the previous rev from the rev cache? i.e.:
- document updated in bucket
- client gets notified of update, requests rev, gets previous version of rev 
- revisionCache.upsert runs
I think this race only creates a problem in the case when the user doesn't have security access to the previous version of the rev, but I don't see an obvious fix.  (even if we invalidate the old entry in the rev cache prior to writing the doc, another process could add it back)

adamcfraser (changes requested):
Couple of minor comments, but would like to discuss the potential race.

db/crud.go
```diff
@@ -1707,6 +1708,7 @@ func (db *Database) updateAndReturnDoc(docid string, allowImport bool, expiry ui
 	var docSequence uint64                                       // Must be scoped outside callback, used over multiple iterations
 	var unusedSequences []uint64                                 // Must be scoped outside callback, used over multiple iterations
 	var oldBodyJSON string                                       // Stores previous revision body for use by DocumentChangeEvent
+	var revIDGenerationSkipped bool
```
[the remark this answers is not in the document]
JRascagneres: Changed to `createNewRevIDSkipped` for now. Think that works?

db/crud.go
JRascagneres: Addressed as per discussion.

db/revision_cache_interface.go
```diff
@@ -32,6 +32,12 @@ type RevisionCache interface {
 	// Put will store the given docRev in the cache
 	Put(docRev DocumentRevision)
 
+	// Update will remove existing value and re-create new one
+	Upsert(docRev DocumentRevision)
+
+	// Invalidate
```
bbrks: Can you provide a description for the use-case or behaviour of Invalidate.

It's not clear what it's used for without knowing what was said in the discussion to implement this, which will be forgotten sooner or later.

db/revision_cache_lru.go
```diff
@@ -135,6 +149,33 @@ func (rc *LRURevisionCache) getFromCache(docID, revID string, loadOnCacheMiss bo
 	return docRev, err
 }
 
+func (rc *LRURevisionCache) LoadInvalidRevFromBackingStore(key IDAndRev, includeBody bool, includeDelta bool) (DocumentRevision, error) {
```
bbrks: Add comment to describe what an invalid rev is (i.e. a newer version of a revision that was previously cached, but needs to be replaced).

bbrks (changes requested):
Couple of minor suggestions for comments for unclear methods around invalid revcache values.
I don't think we'll remember what an invalid revcache value is if we need to revisit this down the line.

db/revision_cache_interface.go
JRascagneres: Added comments how do they look?

db/revision_cache_lru.go
JRascagneres: Added comments how do they look?

db/revision_cache_lru.go
```diff
@@ -135,6 +149,35 @@ func (rc *LRURevisionCache) getFromCache(docID, revID string, loadOnCacheMiss bo
...
+// used. This function grabs the value directly from the bucket.
+func (rc *LRURevisionCache) LoadInvalidRevFromBackingStore(key IDAndRev, includeBody bool, includeDelta bool) (DocumentRevision, error) {
+	var delta *RevisionDelta
+	var docRevBody Body
+
+	value := revCacheValue{
+		key:     key,
+		invalid: true,
+	}
+	value.bodyBytes, value.body, value.history, value.channels, value.removed, value.attachments, value.deleted, value.expiry, value.err = revCacheLoader(rc.backingStore, key, includeBody)
```
adamcfraser: I had anticipated we could use revCacheLoaderForDoc here, at least when this is called from GetActive.  Is there a reason why we need to make the additional GetDocument call that's in revCacheLoader in all cases?

adamcfraser (changes requested):

db/revision_cache_lru.go
JRascagneres: Yeah there isn't a specific reason. I've updated this now. My solution is to modify `LoadInvalidRevFromBackingStore` to take a document. In the event a document is passed in which is the case under GetActive then it'll run revCacheLoaderForDoc. In the event that a document isn't passed in (a nil value) which is the case under `getFromCache` then it'll run `revCacheLoader`.
I think this is the best way to do it.

adamcfraser (approved):

PR #4976, opened 2021-03-04
CBG-1323: Include xattr in _raw

JRascagneres:


rest/admin_api.go
```diff
@@ -403,6 +403,17 @@ func (h *handler) handleGetRawDoc() error {
...
+	if h.db.Options.UserXattrKey != "" {
+		rawUserXattrVal := doc.RawUserXattr
+		if len(rawUserXattrVal) == 0 {
+			rawUserXattrVal = []byte("null")
+		}
+		rawBytes, err = base.InjectJSONPropertiesFromBytes(rawBytes, base.KVPairBytes{Key: h.db.Options.UserXattrKey, Val: rawUserXattrVal})
+		if err != nil {
+			return err
+		}
+	}
```
bbrks: This seems a bit mis-representative, given that the xattr in the sync function is nested under "meta", and this wouldn't.

Also what would happen if the doc already contains a property matching the xattr key (e.g. `"channels"`)?

bbrks (changes requested):

rest/admin_api.go
JRascagneres: Thats a fair point... Similarly a user could have a 'meta' field in their doc though.

rest/admin_api.go
```diff
@@ -403,6 +403,21 @@ func (h *handler) handleGetRawDoc() error {
...
+
+	metaMapBytes, err := base.JSONMarshal(metaMap)
+	if err != nil {
+		return err
+	}
+
+	rawBytes, err = base.InjectJSONPropertiesFromBytes(rawBytes, base.KVPairBytes{Key: "_meta", Val: metaMapBytes})
+	if err != nil {
+		return err
+	}
```
bbrks: Nitpick: There's a variant of this util that does the marshalling for you. But don't bother changing unless there's other stuff to do.

```suggestion
	rawBytes, err = base.InjectJSONProperties(rawBytes, base.KVPair{Key: "_meta", Val: metaMap})
	if err != nil {
		return err
	}
```

rest/admin_api.go
```diff
@@ -403,6 +403,21 @@ func (h *handler) handleGetRawDoc() error {
 		return err
 	}
 
+	metaMap, err := doc.GetMetaMap(h.db.Options.UserXattrKey)
```
adamcfraser: Don't show meta unless userXattrKey is set.

adamcfraser (changes requested):

bbrks (approved):

adamcfraser (approved):

PR #4977, opened 2021-03-17
CBG-1337 Add an option to skip TLS certificates verification while loading JavaScript from HTTPS endpoints

sarathkumarsivan:
Changes to provide an unsupported config option that allows the users to skip TLS verification (or allowing users to use self signed certificate) while loading external JavaScript for import filer, sync function, conflict resolver function and Webhook filter function.

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -340,7 +340,7 @@ func (dbConfig *DbConfig) setup(name string) error {
 
 	// Load Sync Function.
 	if dbConfig.Sync != nil {
-		sync, err := loadJavaScript(*dbConfig.Sync)
+		sync, err := loadJavaScript(*dbConfig.Sync, dbConfig.Unsupported.RemoteConfigTlsSkipVerify)
 		if err != nil {
 			return &JavaScriptLoadError{
 				JSLoadType: SyncFunction,
@@ -353,7 +353,7 @@ func (dbConfig *DbConfig) setup(name string) error {
 
 	// Load Import Filter Function.
 	if dbConfig.ImportFilter != nil {
-		importFilter, err := loadJavaScript(*dbConfig.ImportFilter)
+		importFilter, err := loadJavaScript(*dbConfig.ImportFilter, dbConfig.Unsupported.RemoteConfigTlsSkipVerify)
 		if err != nil {
 			return &JavaScriptLoadError{
 				JSLoadType: ImportFilter,
@@ -367,7 +367,7 @@ func (dbConfig *DbConfig) setup(name string) error {
 	// Load Conflict Resolution Function.
 	for _, rc := range dbConfig.Replications {
 		if rc.ConflictResolutionFn != "" {
-			conflictResolutionFn, err := loadJavaScript(rc.ConflictResolutionFn)
+			conflictResolutionFn, err := loadJavaScript(rc.ConflictResolutionFn, dbConfig.Unsupported.RemoteConfigTlsSkipVerify)
 			if err != nil {
 				return &JavaScriptLoadError{
 					JSLoadType: ConflictResolver,
@@ -386,8 +386,8 @@ func (dbConfig *DbConfig) setup(name string) error {
 // If the specified path does not qualify for a valid file or an URI, it returns the input path
 // as-is with the assumption that it is an inline JavaScript source. Returns error if there is
 // any failure in reading the JavaScript file or URI.
-func loadJavaScript(path string) (js string, err error) {
-	rc, err := readFromPath(path)
+func loadJavaScript(path string, insecureSkipVerify bool) (js string, err error) {
+	rc, err := readFromPath(path, insecureSkipVerify)
 	if errors.Is(err, ErrPathNotFound) {
 		// If rc is nil and readFromPath returns no error, treat the
 		// the given path as an inline JavaScript and return it as-is.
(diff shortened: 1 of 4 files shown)


rest/config.go
```diff
@@ -386,8 +386,8 @@ func (dbConfig *DbConfig) setup(name string) error {
 // If the specified path does not qualify for a valid file or an URI, it returns the input path
 // as-is with the assumption that it is an inline JavaScript source. Returns error if there is
 // any failure in reading the JavaScript file or URI.
-func loadJavaScript(path string) (js string, err error) {
-	rc, err := readFromPath(path)
+func loadJavaScript(path string, insecureSkipVerify bool) (js string, err error) {
```
adamcfraser, line 390: What error gets logged if SG attempts to load config javascript via https that's using a self-signed cert, but the new flag isn't set?  If possible, it would be nice to add information about the new flag in the error message.

rest/config_test.go
```diff
@@ -1377,49 +1378,84 @@ func javaScriptHttpErrorEndpoint() func() (path string, teardownFn func()) {
...
 		{
-			name:        "Load JavaScript from an external endpoint that returns an error",
-			jsInput:     javaScriptHttpErrorEndpoint(),
-			jsExpected:  "",
-			errExpected: errInternalServerError,
+			name:               "Load JavaScript from an external https endpoint with ssl verification enabled",
+			jsInput:            javaScriptHttpsEndpoint(t, js),
+			insecureSkipVerify: false,
+			jsExpected:         "",
+			errExpected:        &url.Error{},
```
adamcfraser: If it's possible to have SG detect the specific error and convert it to a SG error type (that this test could check for), it would be preferable.

adamcfraser (changes requested):

rest/config.go
sarathkumarsivan: The blow error gets logged if SG attempts to load config javascript via https that's using a self-signed cert, but the new flag isn't set:

Error loading JavaScript (SyncFunction) from "https://localhost:5000/sync", Err: Get https://localhost:5000/sync: x509: certificate signed by unknown authority. SSL certificate verification is turned on by default when loading config over HTTPS. Can be overriden by the unsupported "remote_config_tls_skip_verify" option

rest/config_test.go
sarathkumarsivan: Yeah, that's possible. Included changes to wrap the underlying certificate verification error in httpSSLVerifyError. This error type can be asserted in the unit test.

rest/config.go
```diff
@@ -382,18 +383,40 @@ func (dbConfig *DbConfig) setup(name string) error {
...
+	rc, err := readFromPath(path, insecureSkipVerify)
 	if errors.Is(err, ErrPathNotFound) {
 		// If rc is nil and readFromPath returns no error, treat the
 		// the given path as an inline JavaScript and return it as-is.
 		return path, nil
 	}
 	if err != nil {
+		if !insecureSkipVerify {
+			switch errors.Unwrap(err).(type) {
```
adamcfraser: Should use errors.Is here - covers the top-level error and any wrapped error.

rest/config_test.go
```diff
@@ -1377,49 +1377,84 @@ func javaScriptHttpErrorEndpoint() func() (path string, teardownFn func()) {
...
 		{
-			name:        "Load JavaScript from an external endpoint that returns an error",
-			jsInput:     javaScriptHttpErrorEndpoint(),
-			jsExpected:  "",
-			errExpected: errInternalServerError,
+			name:               "Load JavaScript from an external https endpoint with ssl verification enabled",
+			jsInput:            javaScriptHttpsEndpoint(t, js),
+			insecureSkipVerify: false,
+			jsExpected:         "",
+			errExpected:        httpSSLVerifyError{},
```
adamcfraser: As discussed, can switch the test to also use errors.Is, and avoid the named wrapping type (i.e just stick x509.UnknownAuthorityError here)

adamcfraser (changes requested):

rest/config.go
sarathkumarsivan: When looking further, the errors.As seems fits best in this case, because the errors.Is function behaves like a comparison to a sentinel error, and the errors.As function behaves like a type assertion.

rest/config_test.go
sarathkumarsivan: Sure, x509.UnknownAuthorityError can be used here based on the above change.

rest/config.go
```diff
@@ -386,14 +387,23 @@ func (dbConfig *DbConfig) setup(name string) error {
...
 		// If rc is nil and readFromPath returns no error, treat the
 		// the given path as an inline JavaScript and return it as-is.
 		return path, nil
 	}
 	if err != nil {
+		if !insecureSkipVerify {
+			var unkAuthErr x509.UnknownAuthorityError
+			if errors.As(err, &unkAuthErr) {
+				return "", fmt.Errorf("%w. SSL certificate verification is turned "+
+					"on by default when loading config over HTTPS. Can be overriden by "+
```
adamcfraser, line 402:
```suggestion
					"on by default when loading config over HTTPS. Can be overridden by "+
```

rest/config.go
```diff
@@ -386,14 +387,23 @@ func (dbConfig *DbConfig) setup(name string) error {
...
+					"the unsupported \"remote_config_tls_skip_verify\" option", err)
```
adamcfraser:
```suggestion
				return "", fmt.Errorf("%w. TLS certificate failed verification.  TLS verification "+
					"can be disabled using the unsupported \"remote_config_tls_skip_verify\" option", err)
```

sarathkumarsivan: This change has been committed in the latest revision.

adamcfraser (approved):

PR #4979, opened 2021-03-27
CBG-1378 Force SG to initialize the OIDC client in the subsequent request when the provider is not reachable in the first request

sarathkumarsivan:
During OpenID Connect authentication, Sync Gateway initializes the client with the specified configuration by performing an initial provider metadata discovery from a well-known endpoint and initiates the periodic discovery sync in the background. Initialization of the client is performed lazily (when the first request comes in) and exactly once throughout the life cycle of an SG instance. The caveat is that if the provider is not reachable at the time of initializing the OIDC client, SG won’t refresh the client instance that is cached in memory until the time of the next metadata sync. If the provider becomes available later on and any auth request that is made before the next metadata sync can end up in ”401 Invalid Login” even when there is a valid Bearer token provided in the request header that is obtained from the provider. Today, customers are performing a restart to establish the connection with the provider and that is not desirable. The PR includes the changes to force Sync Gateway to reinitialize the client by establishing the connection with the provider during the subsequent authentication request and avoid explicitly restarting the server instance.

### DIFF
--- a/auth/oidc.go
+++ b/auth/oidc.go
@@ -170,7 +170,7 @@ type OIDCProvider struct {
 
 	// clientOnce synchronises access to the GetClient() and ensures that
 	// the OpenID Connect client only gets initialized exactly once.
-	clientOnce sync.Once
+	clientOnce base.Once
 
 	// IsDefault indicates whether this OpenID Connect provider (the current
 	// instance of OIDCProvider is explicitly specified as default provider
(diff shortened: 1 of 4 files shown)


bbrks (changes requested):
If we're going to have a resettable sync.Once, I don't think we should be using this implementation. See comments for detail

sarathkumarsivan:
Closing this ticket since the implementation can be done with a flag based approach.

PR #4981, opened 2021-04-07
CBG-1361: Prepare history on principal docs

JRascagneres:
~~- Store previous channels / roles on principal docs at inval~~
~~- At rebuildRoles/ rebuildChannels use previous channels / roles to build out channel history struct~~
- Bunch of auth level tests following Google Sheets plan verifying history

Updated:
 - At invalidation time we set either ChannelInvalSeq in the case of channel invalidation or RoleInvalSeq in the case of role invalidations to the sequence at which the principal was invalidated. Note we no longer will set the Channels / Roles to nil, we retain the information, InvalSeq is used to specify whether they are invalidated.
 - The Channels() / RoleNames() function returns nil if invalid, allowing existing null checks to be maintained. 
 - At rebuildRoles / rebuildChannels we use the Channels / Roles property to build out history. InvalSeq is used as endSeq in history.
 - Channels / Roles are updated following rebuild and the invalSeq is removed.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -168,13 +168,8 @@ func (auth *Authenticator) rebuildChannels(princ Principal) error {
 	// always grant access to the public document channel
 	channels.AddChannel(ch.DocumentStarChannel, 1)
 
-	var previousChannelEntries ch.TimedSet
-	var previousInvalSeq uint64
-
-	if princ.PreviousChannels() != nil {
-		previousChannelEntries = princ.PreviousChannels().Entries
-		previousInvalSeq = princ.PreviousChannels().InvalSeq
-	}
+	previousChannelEntries := princ.InvalidatedChannels()
+	previousInvalSeq := princ.GetChannelInvalSeq()
 
 	removedChannels := ch.TimedSet{}
 	for previousChannelName, previousChannelInfo := range previousChannelEntries {
@@ -185,17 +180,17 @@ func (auth *Authenticator) rebuildChannels(princ Principal) error {
 
 	channelHistory := princ.ChannelHistory()
 	if channelHistory == nil {
-		channelHistory = map[string]ChannelOrRoleHistoryEntries{}
+		channelHistory = map[string]TimeSetHistoryEntries{}
 	}
 
 	for name, channelInfo := range removedChannels {
 		currentChannelHistory, ok := channelHistory[name]
 		if !ok {
-			currentChannelHistory = ChannelOrRoleHistoryEntries{}
+			currentChannelHistory = TimeSetHistoryEntries{}
 		}
 
 		currentChannelHistory.UpdatedAt = time.Now().UnixNano()
-		currentChannelHistory.Entries = append(currentChannelHistory.Entries, ChannelOrRoleHistoryEntry{
+		currentChannelHistory.Entries = append(currentChannelHistory.Entries, TimeSetHistoryEntry{
 			Seq:    channelInfo.Sequence,
 			EndSeq: previousInvalSeq,
 		})
@@ -208,7 +203,7 @@ func (auth *Authenticator) rebuildChannels(princ Principal) error {
 	}
 
 	base.Infof(base.KeyAccess, "Recomputed channels for %q: %s", base.UD(princ.Name()), base.UD(channels))
-	princ.SetPreviousChannels(nil)
+	princ.SetChannelInvaliSeq(0)
 	princ.setChannels(channels)
 
 	return nil
@@ -233,13 +228,8 @@ func (auth *Authenticator) rebuildRoles(user User) error {
 		roles.Add(explicit)
 	}
 
-	var previousRoleEntries ch.TimedSet
-	var previousInvalSeq uint64
-
-	if user.PreviousRoles() != nil {
-		previousRoleEntries = user.PreviousRoles().Entries
-		previousInvalSeq = user.PreviousRoles().InvalSeq
-	}
+	previousRoleEntries := user.InvalidatedRoles()
+	previousInvalSeq := user.GetRoleInvalSeq()
 
 	removedRoles := ch.TimedSet{}
 	for previousRoleName, previousRoleInfo := range previousRoleEntries {
@@ -250,17 +240,17 @@ func (auth *Authenticator) rebuildRoles(user User) error {
 
 	roleHistory := user.RoleHistory()
 	if roleHistory == nil {
-		roleHistory = map[string]ChannelOrRoleHistoryEntries{}
+		roleHistory = map[string]TimeSetHistoryEntries{}
 	}
 
 	for name, roleInfo := range removedRoles {
 		currentRoleHistory, ok := roleHistory[name]
 		if !ok {
-			currentRoleHistory = ChannelOrRoleHistoryEntries{}
+			currentRoleHistory = TimeSetHistoryEntries{}
 		}
 
 		currentRoleHistory.UpdatedAt = time.Now().UnixNano()
-		currentRoleHistory.Entries = append(currentRoleHistory.Entries, ChannelOrRoleHistoryEntry{
+		currentRoleHistory.Entries = append(currentRoleHistory.Entries, TimeSetHistoryEntry{
 			Seq:    roleInfo.Sequence,
 			EndSeq: previousInvalSeq,
 		})
@@ -273,8 +263,8 @@ func (auth *Authenticator) rebuildRoles(user User) error {
 	}
 
 	base.Infof(base.KeyAccess, "Computed roles for %q: %s", base.UD(user.Name()), base.UD(roles))
+	user.SetRoleInvaliSeq(0)
 	user.setRolesSince(roles)
-	user.SetPreviousRoles(nil)
 	return nil
 }
 
@@ -339,14 +329,7 @@ func (auth *Authenticator) InvalidateChannels(p Principal, invalSeq uint64) erro
 
 		base.Infof(base.KeyAccess, "Invalidate access of %q", base.UD(p.Name()))
 
-		if p.PreviousChannels() == nil {
-			p.SetPreviousChannels(&PreviousChannelsOrRole{
-				Entries:  p.Channels(),
-				InvalSeq: invalSeq,
-			})
-		}
-
-		p.setChannels(nil)
+		p.SetChannelInvaliSeq(invalSeq)
 		return p, nil
 	}
 	return auth.casUpdatePrincipal(p, invalidateChannelsCallback)
@@ -365,15 +348,7 @@ func (auth *Authenticator) InvalidateRoles(user User, invalSeq uint64) error {
 		}
 		base.Infof(base.KeyAccess, "Invalidate roles of %q", base.UD(user.Name()))
 
-		if user.PreviousRoles() == nil {
-			roleNames := user.RoleNames()
-			user.SetPreviousRoles(&PreviousChannelsOrRole{
-				Entries:  roleNames,
-				InvalSeq: invalSeq,
-			})
-		}
-
-		user.setRolesSince(nil)
+		user.SetRoleInvaliSeq(invalSeq)
 		return user, nil
 	}
 
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -209,7 +209,7 @@ func (auth *Authenticator) calculateHistory(invalSeq uint64, invalGrants ch.Time
 
 		// Add grant to history
 		currentHistoryForGrant.UpdatedAt = time.Now().UnixNano()
-		currentHistoryForGrant.Entries = append(currentHistoryForGrant.Entries, GrantHistoryEntry{
+		currentHistoryForGrant.Entries = append(currentHistoryForGrant.Entries, GrantHistorySequencePair{
 			Seq:    previousInfo.Sequence,
 			EndSeq: invalSeq,
 		})
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -157,30 +157,55 @@ func (auth *Authenticator) getPrincipal(docID string, factory func() Principal) 
 func (auth *Authenticator) rebuildChannels(princ Principal) error {
 	channels := princ.ExplicitChannels().Copy()
 
-	// Changes for vbucket sequence management.  We can't determine relative ordering of sequences
-	// across vbuckets. To avoid redundant channel backfills during changes processing, we maintain
-	// the previous vb/seq for a channel in PreviousChannels.  If that channel is still present during
-	// this rebuild, we reuse the vb/seq from PreviousChannels (using UpdateIfPresent).  If PreviousChannels
-	// is nil, reverts to normal sequence handling.
-
-	previousChannels := princ.PreviousChannels().Copy()
-	if previousChannels != nil {
-		channels.UpdateIfPresent(previousChannels)
-	}
-
 	if auth.channelComputer != nil {
 		viewChannels, err := auth.channelComputer.ComputeChannelsForPrincipal(princ)
 		if err != nil {
 			base.Warnf("channelComputer.ComputeChannelsForPrincipal returned error for %v: %v", base.UD(princ), err)
 			return err
 		}
-		if previousChannels != nil {
-			viewChannels.UpdateIfPresent(previousChannels)
-		}
 		channels.Add(viewChannels)
 	}
 	// always grant access to the public document channel
 	channels.AddChannel(ch.DocumentStarChannel, 1)
+
+	var previousChannelEntries ch.TimedSet
+	var previousInvalSeq uint64
+
+	if princ.PreviousChannels() != nil {
+		previousChannelEntries = princ.PreviousChannels().Entries
+		previousInvalSeq = princ.PreviousChannels().InvalSeq
+	}
+
+	removedChannels := ch.TimedSet{}
+	for previousChannelName, previousChannelInfo := range previousChannelEntries {
+		if _, ok := channels[previousChannelName]; !ok {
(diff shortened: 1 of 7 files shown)


auth/role.go
```diff
@@ -20,15 +20,33 @@ import (
...
+	ExplicitChannels_ ch.TimedSet             `json:"admin_channels,omitempty"`
+	Channels_         ch.TimedSet             `json:"all_channels"`
+	Sequence_         uint64                  `json:"sequence"`
+	PreviousChannels_ *PreviousChannelsOrRole `json:"previous_channels,omitempty"`
+	ChannelHistory_   ChannelOrRoleHistory    `json:"channel_history,omitempty"`
 	vbNo              *uint16
 	cas               uint64
 }
 
+type PreviousChannelsOrRole struct {
```
adamcfraser: I think it would be better to have a more generic name for this struct, instead of PreviousChannelsOrRole.  InvalidatedTimedSet seems like one possibility.

Can we also add a description for this struct, especially to describe what invalSeq represents, compared to the sequences that are included in the timed set?

auth/role.go
```diff
@@ -20,15 +20,33 @@ import (
...
+	InvalSeq uint64      `json:"inval_seq"`
+	Entries  ch.TimedSet `json:"entries"`
+}
+
+type ChannelOrRoleHistory map[string]ChannelOrRoleHistoryEntries
```
adamcfraser: Same comment above about a more generic name, instead of ChannelOrRoleX.  Same goes for the comments describing the distinction between UpdatedAt and Seq/EndSeq.

auth/role.go
```diff
@@ -20,15 +20,33 @@ import (
...
-	ExplicitChannels_ ch.TimedSet `json:"admin_channels,omitempty"`
-	Channels_         ch.TimedSet `json:"all_channels"`
-	Sequence_         uint64      `json:"sequence"`
-	PreviousChannels_ ch.TimedSet `json:"previous_channels,omitempty"`
+	Name_             string                  `json:"name,omitempty"`
+	ExplicitChannels_ ch.TimedSet             `json:"admin_channels,omitempty"`
+	Channels_         ch.TimedSet             `json:"all_channels"`
+	Sequence_         uint64                  `json:"sequence"`
+	PreviousChannels_ *PreviousChannelsOrRole `json:"previous_channels,omitempty"`
+	ChannelHistory_   ChannelOrRoleHistory    `json:"channel_history,omitempty"`
```
adamcfraser: Is there duplication of information between PreviousChannels_ and ChannelHistory_?  Or would only one be populated at a time on a persisted role document.

auth/auth.go
```diff
@@ -157,31 +157,56 @@ func (auth *Authenticator) getPrincipal(docID string, factory func() Principal)
...
+	}
+
+	removedChannels := ch.TimedSet{}
+	for previousChannelName, previousChannelInfo := range previousChannelEntries {
+		if _, ok := channels[previousChannelName]; !ok {
+			removedChannels[previousChannelName] = previousChannelInfo
+		}
+	}
+
+	channelHistory := princ.ChannelHistory()
```
adamcfraser: What's the reason to not just stick previousChannels into channelHistory at invalidation time, instead of waiting for next auth?

adamcfraser (changes requested):
Generally looks good, a few style/naming questions and request for additional comments, so I can understand what data is actually being stored.  Also want to understand the necessity of separating 'history' vs. 'previous'.

auth/auth.go
```diff
@@ -265,22 +330,30 @@ func (auth *Authenticator) UpdateSequenceNumber(p Principal, seq uint64) error {
...
+func (auth *Authenticator) InvalidateChannels(p Principal, invalSeq uint64) error {
 	invalidateChannelsCallback := func(p Principal) (updatedPrincipal Principal, err error) {
 
 		if p == nil || p.Channels() == nil {
 			return p, base.ErrUpdateCancel
 		}
 
 		base.Infof(base.KeyAccess, "Invalidate access of %q", base.UD(p.Name()))
+
+		if p.PreviousChannels() == nil {
```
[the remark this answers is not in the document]
JRascagneres: I'll put in a comment too but for first question:
- Previous channels is only set once because we only need to know the information since the user last pulled data. If there are a bunch of grants removed and added again between pulls we don't really care about those. 

Second:
- Yeah I've had a think and I can't think of a scenario where you'd end up with both being non-nil but as you say its defensive so doesn't hurt.

auth/role.go
JRascagneres: Again, will add comment but to clarify here>
invalSeq: The sequence the change was triggered by as obtained from MarkPrincipalsChanged, Later used for triggeredBy
TimedSet: Just contains previous data

auth/role.go
JRascagneres: UpdatedAt: Timestamp when data was updated, used for pruning (not yet implemented)
Seq: Sequence when a grant was done to give access to a role/channel (only populated when endSeq is also there)
EndSeq: Sequence when access to a role/channel was revoked.

auth/role.go
JRascagneres: Shouldn't be any duplication no. Both could be populated at once but if they are they will contain different info. When a user performs an auth the PreviousChannels / Role is read to build up History and then is nilled. 

Both bits of data would exist in the Scenario where:
- Channel history exists so basically there has been an invalidation which removed a channel and then an auth occurred to build the channel history.
- Another invalidation occurs which would set PreviousChannels but where another auth has not yet been performed.

auth/auth.go
JRascagneres: Couple reasons:
- Want to know what current assignments are so we can diff with them so find changes. The query / work done to calculate current assignments is already done in rebuild.
- In the case of multiple invalidations you don't want to do all the work until necessary.

To have work done on invalidation:
- Need a way to find current assignments, involving adding a query
- Lose the advantage of only doing the invalidation once. You'd need to recalculate every time there is a change.

auth/auth.go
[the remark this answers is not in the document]
JRascagneres: Removed Previous

auth/role.go
JRascagneres: Removed previous

auth/role.go
JRascagneres: Previous removed

auth/auth.go
JRascagneres: Should be addressed now

JRascagneres:
Marked existing comments as resolved as I'm fairly sure they've either been done or are no longer relevant based on updates.

bbrks (changes requested):
Small comment about how we do the diffing, but otherwise looks OK to me.

auth/auth.go
```diff
@@ -157,37 +157,63 @@ func (auth *Authenticator) getPrincipal(docID string, factory func() Principal)
...
-	princ.SetPreviousChannels(nil)
+	princ.SetChannelInvalSeq(0)
 	princ.setChannels(channels)
 
 	return nil
+}
+
+// Calculates history for either roles or channels
+func (auth *Authenticator) calculateHistory(invalSeq uint64, invalGrants ch.TimedSet, newGrants ch.TimedSet, currentHistory TimeSetHistory) TimeSetHistory {
+	removedGrants := ch.TimedSet{}
```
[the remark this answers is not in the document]
JRascagneres: Fair comment. Half the loops! Done

auth/auth.go
```diff
@@ -157,37 +157,63 @@ func (auth *Authenticator) getPrincipal(docID string, factory func() Principal)
...
+
+	// Iterate over removed grants and build history
+	for name, info := range removedGrants {
+		currentHistoryForGrant, ok := currentHistory[name]
+		if !ok {
+			currentHistoryForGrant = TimeSetHistoryEntries{}
+		}
+
+		currentHistoryForGrant.UpdatedAt = time.Now().UnixNano()
+		currentHistoryForGrant.Entries = append(currentHistoryForGrant.Entries, TimeSetHistoryEntry{
```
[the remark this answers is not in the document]
JRascagneres: Done

adamcfraser (changes requested):
The updated approach looks good to me.  I did have some comments on the struct used for history, and I found the tests difficult to review - made a few requests/suggestions for making the tests more readable and self-describing.

auth/auth_test.go
```diff
@@ -1240,3 +1239,1206 @@ func TestAuthenticateUntrustedJWT(t *testing.T) {
...
+		userPrincipal, ok := principal.(*userImpl)
+		assert.True(t, ok)
+		rolePrincipal, err := auth.GetPrincipal("foo", false)
+		assert.NoError(t, err)
+		return userPrincipal, rolePrincipal
+	}
+
+	auth := NewAuthenticator(testBucket, &testMockComputer)
+
+	user, err := auth.NewUser("alice", "password", ch.SetOf(t))
```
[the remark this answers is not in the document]
JRascagneres: No, can swap out.

auth/auth_test.go
```diff
@@ -1240,3 +1239,1206 @@ func TestAuthenticateUntrustedJWT(t *testing.T) {
...
+	assert.NoError(t, err)
+	err = auth.Save(user)
+	assert.NoError(t, err)
```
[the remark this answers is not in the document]
JRascagneres: `getPrincipals` moved out into a separate func.
Created `initializeScenario` which initiates user and role and returns them.

Although the role and channel grant are also done in all tests I want to keep these in the individual tests. Its only 2 lines and think its important to be explicit about the grants being added / removed in the tests.

auth/auth_test.go
```diff
@@ -316,13 +316,12 @@ func (self *mockComputer) UseGlobalSequence() bool {
...
 
 func TestRebuildUserChannels(t *testing.T) {
-
 	bucket := base.GetTestBucket(t)
 	defer bucket.Close()
 	computer := mockComputer{channels: ch.AtSequence(ch.SetOf(t, "derived1", "derived2"), 1)}
 	auth := NewAuthenticator(bucket, &computer)
 	user, _ := auth.NewUser("testUser", "password", ch.SetOf(t, "explicit1"))
-	user.setChannels(nil)
+	user.SetChannelInvalSeq(2)
```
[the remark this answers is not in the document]
JRascagneres: Yeah fair, makes more sense. I was trying to match the old behaviour best but this is probably better

auth/auth_test.go
```diff
@@ -1240,3 +1239,1206 @@ func TestAuthenticateUntrustedJWT(t *testing.T) {
...
+	err = auth.Save(role)
+	assert.NoError(t, err)
+
+	// Get principals to do initial ops on
+	alicePrincipal, err := auth.GetPrincipal("alice", true)
+	assert.NoError(t, err)
+	aliceUserPrincipal, ok := alicePrincipal.(*userImpl)
+	assert.True(t, ok)
+	fooPrincipal, err := auth.GetPrincipal("foo", false)
+	assert.NoError(t, err)
```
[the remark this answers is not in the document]
JRascagneres: This initial one is now covered by the `initializeScenario` function. All other uses to grab user / principal are done with `getPrincipals` already.

auth/role.go
```diff
@@ -20,15 +20,28 @@ import (
...
+	ExplicitChannels_ ch.TimedSet    `json:"admin_channels,omitempty"`
+	Channels_         ch.TimedSet    `json:"all_channels"`
+	Sequence_         uint64         `json:"sequence"`
+	ChannelHistory_   TimeSetHistory `json:"channel_history,omitempty"`
+	ChannelInvalSeq   uint64         `json:"channel_inval_seq"`
 	vbNo              *uint16
 	cas               uint64
 }
 
+type TimeSetHistory map[string]TimeSetHistoryEntries
```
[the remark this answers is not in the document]
JRascagneres: Yeah np, I think this was probably a typo.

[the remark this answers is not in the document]
JRascagneres: Went with last example. Written custom marshal / unmarshal as discussed.

auth/auth_test.go
```diff
@@ -1240,3 +1239,1206 @@ func TestAuthenticateUntrustedJWT(t *testing.T) {
...
+	err = auth.Save(user)
+	assert.NoError(t, err)
+}
+
+// =======================================================================================================
+// The below 'TestRevocationScenario' tests refer to scenarios in the following Google Sheet
+// https://docs.google.com/spreadsheets/d/1pTLyJqrSdde-dAxDfMkKGXi0ttWF4GVCOBFJg7hRY0U/edit?usp=sharing
+// =======================================================================================================
+
+func TestRevocationScenario1(t *testing.T) {
```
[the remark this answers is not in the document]
JRascagneres: Attempted to add comments that weren't too verbose.

auth/auth_test.go
```diff
@@ -1240,3 +1240,933 @@ func TestAuthenticateUntrustedJWT(t *testing.T) {
...
+	assert.Equal(t, GrantHistorySequencePair{Seq: 65, EndSeq: 95}, userRoleHistory.Entries[1])
+
+	channelHistory, ok := fooPrincipal.ChannelHistory()["ch1"]
+	require.True(t, ok)
+	assert.Equal(t, GrantHistorySequencePair{Seq: 75, EndSeq: 85}, channelHistory.Entries[0])
+
+	assert.Equal(t, 0, len(aliceUserPrincipal.ChannelHistory()))
+}
+
+// Scenario 3
```
adamcfraser, line 1504: I'm not sure these descriptions are that illuminating as written.  I don't know what 'Get Doc' means in the context of this test, and I think the main difference between scenarios is timing of requests (compared to the grants/revocation), and that's not described.

auth/auth_test.go
```diff
@@ -1240,3 +1240,933 @@ func TestAuthenticateUntrustedJWT(t *testing.T) {
...
+	auth := NewAuthenticator(testBucket, &testMockComputer)
+	aliceUserPrincipal, fooPrincipal := initializeScenario(t, auth)
+
+	testMockComputer.addRoleChannels(t, auth, fooPrincipal, "foo", "ch1", 5)
+	testMockComputer.addRole(t, auth, aliceUserPrincipal, "alice", "foo", 20)
+
+	// Get Principals / Rebuild Seq 25
+	aliceUserPrincipal, fooPrincipal = getPrincipals(t, auth)
+
+	// Ensure user can see ch1 (via role)
```
adamcfraser, line 1447: These tests seem to be doing 'ensure user can see channel' as a surrogate for what's shown as changes requests in the spreadsheet, but I don't think that covers the key question, which is 'would a changes request at this sequence trigger a revocation'.  

I don't think you need to implement the entire changes handling to do that, but I feel like it does make sense to have a test function that performs the same evaluation (of role history, user channel history and user role history) for that.  Is the idea that you'd enhance these tests when you implement the changes-side work?

auth/auth_test.go
adamcfraser: If you want to list the steps (as opposed to a verbal description of what differentiates the scenario), it might be better in the form:
```
[describe setup]
- changes request, expected result
[describe what's changed]
- changes request, expected result
[describe what's changed]
- changes request, expected result
```
repeat as needed

adamcfraser: Also, the sequence for role/channel grants/revocations is the same for all these scenarios, is that right?  The only thing changing is the point at which you're simulating a replication?  If so, that sounds like something that would actually make more sense as a single unit test, with an 'ensure user can...' block for each scenario.

adamcfraser (changes requested):

auth/auth_test.go
JRascagneres: Yes. Its true that the same grants and revocations occur in each scenario. However, due to how the history is built you can't really do this in one test. If you mean do it as a case table kind of test that would be possible, I considered it, but it got difficult to read whats going on exactly in each step.

Agreed with the comments. I was trying to cover both being descriptive to know whats going on but do in a few words which I think I ended up failing at both.

auth/auth_test.go
JRascagneres: The can see channel is a part of it but not entirely. The point in these tests are to verify that the history is built correctly and not necessarily that we would revoke. The revoke tests would come later. Once the changes-side work is complete there will be similar tests doing the same scenarios but in full. With actual REST calls to create the user / docs and issue changes request to ensure the document is revoked.

auth/auth_test.go
adamcfraser: I thought the role/channel history should be identical for all the scenarios?  My reading of the spreadsheet is that the user/role/channel revocation history is the same across all scenarios, and the differences are when a client attempts to replicate.  Can you explain more what you mean when you say 'due to how the history is built, you can't really do this in one test'?  If it's due to information being persisted to the user document on authentication, you could invalidate channels/roles prior to doing that check.  However, you'd also want to test to ensure that invalidation isn't required (i.e. one version of the test makes sure the history is as expected without validation, one checks with validation.

auth/auth_test.go
adamcfraser: So is the reason the tests are doing 'ensure user can see channel' just to perform auth, so that the channel history gets resolved?  Probably best to state that explicitly.

JRascagneres: No. GetPrincipals does the auth. Can probably drop the 'ensure user can see channel' really

JRascagneres: Should be addressed as per discussion

auth/auth_test.go
JRascagneres: Added comments like above and other issues should be addressed as per discussion

adamcfraser (approved):

PR #4982, opened 2021-04-09
CBG-1363: Add revocations option to _changes and subChanges

JRascagneres:
Just allows clients to specify revocations option. Will be used in following PR where this is actually used.
Also added config option for ISGR side which:
- passes option to subChanges too for pull
- sends changes with revocations for push

Wasn't really possible to write tests for it at this stage but was manually tested with breakpoint. Can easily be tested once revocation messages are in.

### DIFF
--- a/db/blip_messages.go
+++ b/db/blip_messages.go
@@ -15,14 +15,15 @@ type BLIPMessageSender interface {
 
 // SubChangesRequest is a strongly typed 'subChanges' request.
 type SubChangesRequest struct {
-	Continuous     bool     // Continuous can be set to true if the requester wants change notifications to be sent indefinitely (optional)
-	Batch          uint16   // Batch controls the maximum number of changes to send in a single change message (optional)
-	Since          string   // Since represents the latest sequence ID already known to the requester (optional)
-	Filter         string   // Filter is the name of a filter function known to the recipient (optional)
-	FilterChannels []string // FilterChannels are a set of channels used with a 'sync_gateway/bychannel' filter (optional)
-	DocIDs         []string // DocIDs specifies which doc IDs the recipient should send changes for (optional)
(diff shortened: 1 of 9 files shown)


db/blip_messages.go
```diff
@@ -15,14 +15,15 @@ type BLIPMessageSender interface {
...
-	ActiveOnly     bool     // ActiveOnly is set to `true` if the requester doesn't want to be sent tombstones. (optional)
-	clientType     clientType
+	Continuous      bool     // Continuous can be set to true if the requester wants change notifications to be sent indefinitely (optional)
+	Batch           uint16   // Batch controls the maximum number of changes to send in a single change message (optional)
+	Since           string   // Since represents the latest sequence ID already known to the requester (optional)
+	Filter          string   // Filter is the name of a filter function known to the recipient (optional)
+	FilterChannels  []string // FilterChannels are a set of channels used with a 'sync_gateway/bychannel' filter (optional)
+	DocIDs          []string // DocIDs specifies which doc IDs the recipient should send changes for (optional)
+	ActiveOnly      bool     // ActiveOnly is set to `true` if the requester doesn't want to be sent tombstones. (optional)
+	EnableAutoPurge bool
```
bbrks: Godoc comment as discussed - this should feed into the replicator protocol spec here (which also needs updating):
https://github.com/couchbase/couchbase-lite-core/blob/master/modules/docs/pages/replication-protocol.adoc#subchanges

bbrks (changes requested):

db/blip_messages.go
JRascagneres: Added comment and renamed to Revocations as this is what the actual subChanges request element will be. Will update protocol once agreed with CBL I guess.

bbrks (approved):
LGTM

adamcfraser (approved):

PR #4984, opened 2021-04-15
CBG-1351 Abort SG startup when an undefined environment variable is specified in the config without default value

sarathkumarsivan:
It's reasonable to throw an error and let the users to know about the dangling reference when a nonexistent environment variable is being referenced in the config.

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -833,9 +833,15 @@ func expandEnv(config []byte) (value []byte, errs error) {
 	})), errs
 }
 
-// ErrEnvVarUndefined is returned when a specified variable can’t be resolved from the system environment or no
-// default value is supplied in the configuration — startup will be aborted.
-const ErrEnvVarUndefined = "undefined environment variable '${%s}' is specified in the config without default value"
+// ErrEnvVarUndefined is returned when a specified variable can’t be resolved from
+// the system environment and no default value is supplied in the configuration.
+type ErrEnvVarUndefined struct {
+	key string // Environment variable identifier.
+}
+
+func (e ErrEnvVarUndefined) Error() string {
+	return fmt.Sprintf("undefined environment variable '${%s}' is specified in the config without default value", e.key)
+}
 
 // envDefaultExpansion implements the ${foo:-bar} parameter expansion from
 // https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_02
@@ -849,7 +855,7 @@ func envDefaultExpansion(key string, getEnvFn func(string) string) (value string
 		base.Debugf(base.KeyAll, "Replacing config environment variable '${%s}' with "+
 			"default value specified", key)
 	} else if value == "" && len(kvPair) != 2 {
-		return "", fmt.Errorf(ErrEnvVarUndefined, key)
+		return "", ErrEnvVarUndefined{key: key}
 	} else {
 		base.Debugf(base.KeyAll, "Replacing config environment variable '${%s}'", key)
 	}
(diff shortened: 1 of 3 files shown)


rest/config_test.go
```diff
@@ -1213,7 +1262,13 @@ func TestExpandEnv(t *testing.T) {
...
 			}
 			// Check environment variable substitutions.
-			actualConfig := expandEnv(test.inputConfig)
+			actualConfig, err := expandEnv(test.inputConfig)
+			if test.expectedError != nil {
+				errs, ok := err.(*multierror.Error)
+				require.True(t, ok)
+				require.Equal(t, test.expectedError.Len(), errs.Len())
+				assert.Equal(t, test.expectedError, errs)
+			}
```
bbrks, line 1271: We're not ensuring `err` is nil here.

```suggestion
			} else {
				assert.NoError(t, err)
			}
```

rest/config.go
```diff
@@ -814,21 +817,29 @@ func (config *ServerConfig) setupAndValidateDatabases() (errs error) {
...
+		if err != nil {
+			errs = multierror.Append(errs, err)
+		}
+		return val
+	})), errs
 }
 
+// ErrEnvVarUndefined is returned when a specified variable can’t be resolved from the system environment or no
+// default value is supplied in the configuration — startup will be aborted.
+const ErrEnvVarUndefined = "undefined environment variable '${%s}' is specified in the config without default value"
```
bbrks: Seems like somewhere we'd want a typed error to allow for use of `errors.Is`/`errors.As`

```suggestion
type ErrEnvVarUndefined struct {
	key string
}

func (e ErrEnvVarUndefined) Error() string {
	return fmt.Sprintf("undefined environment variable '${%s}' is specified in the config without default value", e.key)
}
```

rest/config.go
```diff
@@ -814,21 +817,29 @@ func (config *ServerConfig) setupAndValidateDatabases() (errs error) {
...
+		val, err := envDefaultExpansion(key, os.Getenv)
+		if err != nil {
+			errs = multierror.Append(errs, err)
+		}
+		return val
+	})), errs
 }
 
+// ErrEnvVarUndefined is returned when a specified variable can’t be resolved from the system environment or no
+// default value is supplied in the configuration — startup will be aborted.
```
bbrks: SG startup behaviour isn't really related to this error.

```suggestion
// ErrEnvVarUndefined is returned when a specified variable can’t be resolved from the system environment and no
// default value is supplied in the configuration.
```

bbrks (changes requested):
Functionally looks fine, should use an approach that allows for use of `errors.Is`/`errors.As` though.

rest/config_test.go
sarathkumarsivan: Yeah, this is was nice catch. It's committed to latest revision.

rest/config.go
sarathkumarsivan: I've considered including this as a typed error initially, but later decided to not go with it since there was special handling by using the errors.Is/errors.As. When revisiting this now, it seems reasonable to include typed error because it can be used in the unit tests as well.

rest/config.go
sarathkumarsivan: Actually the intention was to include a note saying that SG won't allow even a single undefined variable being referenced in the config file without a default value. This comment has been refactored.

rest/config.go
bbrks: What do you mean by special handling? I think generally speaking, all new errors we're creating in SG should be compatible with Is/As for consistency reasons and ease of use more than anything else.

rest/config.go
bbrks: I think I was more concerned with commenting the current behaviour in a place where we're not actually aborting startup (i.e. returning the error).

If we switched SG to log a warning instead this comment would likely be left out of date because it's not really part of the code.

bbrks (approved):

adamcfraser (approved):

PR #4985, opened 2021-04-19
CBG-1392: Support soft role deletion

JRascagneres:
Changes role deletion to perform a soft delete so that its previous channels can be accessed to calculate revoked channels. Also builds history when role is deleted.
Has REST side and auth side unit test.

- [x] Rebase

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -90,6 +90,14 @@ func (auth *Authenticator) GetUser(name string) (User, error) {
 
 // Looks up the information for a role.
 func (auth *Authenticator) GetRole(name string) (Role, error) {
+	role, err := auth.GetRoleIncDeleted(name)
+	if role != nil && role.IsDeleted() {
+		return nil, err
+	}
+	return role, err
+}
+
+func (auth *Authenticator) GetRoleIncDeleted(name string) (Role, error) {
 	princ, err := auth.getPrincipal(docIDForRole(name), func() Principal { return &roleImpl{} })
 	role, _ := princ.(Role)
 	return role, err
@@ -111,7 +119,7 @@ func (auth *Authenticator) getPrincipal(docID string, factory func() Principal) 
 			return nil, nil, false, pkgerrors.WithStack(base.RedactErrorf("base.JSONUnmarshal() error for doc ID: %s in getPrincipal().  Error: %v", base.UD(docID), err))
 		}
 		changed := false
-		if princ.Channels() == nil {
+		if princ.Channels() == nil && !princ.IsDeleted() {
 			// Channel list has been invalidated by a doc update -- rebuild it:
 			if err := auth.rebuildChannels(princ); err != nil {
 				base.Warnf("RebuildChannels returned error: %v", err)
@@ -442,14 +450,41 @@ func (auth *Authenticator) casUpdatePrincipal(p Principal, callback casUpdatePri
 }
 
 // Deletes a user/role.
-func (auth *Authenticator) Delete(p Principal) error {
-	if user, ok := p.(User); ok {
+func (auth *Authenticator) Delete(p Principal, purge bool, deleteSeq uint64) error {
+	user, ok := p.(User)
+	if ok {
 		if user.Email() != "" {
 			if err := auth.bucket.Delete(docIDForUserEmail(user.Email())); err != nil {
 				base.Debugf(base.KeyAuth, "Error deleting document ID for user email %s. Error: %v", base.UD(user.Email()), err)
 			}
 		}
+		return auth.bucket.Delete(p.DocID())
 	}
+
+	// If not a user, ie. a role
+	if purge {
+		return auth.Purge(p)
+	}
+	return auth.casUpdatePrincipal(p, func(p Principal) (updatedPrincipal Principal, err error) {
+		if p == nil || p.IsDeleted() {
+			return p, base.ErrUpdateCancel
+		}
+		p.setDeleted(true)
+		p.SetSequence(deleteSeq)
+
+		channelHistory := auth.calculateHistory(deleteSeq, p.Channels(), nil, p.ChannelHistory())
+		if len(channelHistory) != 0 {
+			p.SetChannelHistory(channelHistory)
+		}
+
+		p.SetChannelInvalSeq(deleteSeq)
+		return p, nil
+
+	})
+}
+
+// Purge a user/role from bucket
+func (auth *Authenticator) Purge(p Principal) error {
 	return auth.bucket.Delete(p.DocID())
 }
 
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -449,23 +449,20 @@ func (auth *Authenticator) casUpdatePrincipal(p Principal, callback casUpdatePri
 	return err
 }
 
-// Deletes a user/role.
-func (auth *Authenticator) Delete(p Principal, purge bool, deleteSeq uint64) error {
-	user, ok := p.(User)
-	if ok {
-		if user.Email() != "" {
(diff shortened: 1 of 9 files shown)


auth/auth.go
```diff
@@ -90,6 +90,14 @@ func (auth *Authenticator) GetUser(name string) (User, error) {
...
 // Looks up the information for a role.
 func (auth *Authenticator) GetRole(name string) (Role, error) {
+	role, err := auth.GetRoleIncDeleted(name)
+	if role != nil && role.IsDeleted() {
+		return nil, err
+	}
+	return role, err
+}
+
+func (auth *Authenticator) GetRoleIncDeleted(name string) (Role, error) {
```
adamcfraser, line 100: Looks like this can be private.  And since the name isn't immediately obvious, stick a comment in describing how it differs from GetRole.

auth/auth.go
```diff
@@ -90,6 +90,14 @@ func (auth *Authenticator) GetUser(name string) (User, error) {
 
 // Looks up the information for a role.
 func (auth *Authenticator) GetRole(name string) (Role, error) {
+	role, err := auth.GetRoleIncDeleted(name)
+	if role != nil && role.IsDeleted() {
+		return nil, err
```
adamcfraser, line 95: Isn't err going to be nil here if it's soft-deleted?  I'd expected this method to return a NotFound error if it existed but was deleted, to match the previous behaviour.

auth/auth.go
```diff
@@ -442,14 +450,41 @@ func (auth *Authenticator) casUpdatePrincipal(p Principal, callback casUpdatePri
 }
 
 // Deletes a user/role.
-func (auth *Authenticator) Delete(p Principal) error {
-	if user, ok := p.(User); ok {
+func (auth *Authenticator) Delete(p Principal, purge bool, deleteSeq uint64) error {
+	user, ok := p.(User)
+	if ok {
```
adamcfraser: This doesn't look like it needed to be changed from the single line syntax it was using previously.

db/users.go
```diff
@@ -83,6 +83,28 @@ func (dbc *DatabaseContext) GetPrincipal(name string, isUser bool) (info *Princi
 	return
 }
 
+func (db *DatabaseContext) DeletePrincipal(name string, isUser bool, purge bool) (found bool, err error) {
```
adamcfraser: There doesn't really seem to be that much common code between the delete handling for role and user - I think it would make the code much cleaner and easier to maintain if we just switch to DeleteUser and DeleteRole down in auth, and only need this intermediary in the delete role case (to grab the seq).

auth/auth.go
```diff
@@ -442,14 +450,41 @@ func (auth *Authenticator) casUpdatePrincipal(p Principal, callback casUpdatePri
...
 			}
 		}
+		return auth.bucket.Delete(p.DocID())
+	}
+
+	// If not a user, ie. a role
+	if purge {
+		return auth.Purge(p)
 	}
+	return auth.casUpdatePrincipal(p, func(p Principal) (updatedPrincipal Principal, err error) {
```
adamcfraser: casUpdatePrincipal doesn't look like it's using GetRoleIncDeleted - shouldn't it be doing that?  Do we have any tests for a role update that goes into CAS retry because the role has been deleted?

auth/role.go
```diff
@@ -139,6 +150,14 @@ func (role *roleImpl) SetCas(cas uint64) {
 	role.cas = cas
 }
 
+func (role *roleImpl) setDeleted(deleted bool) {
+	role.Deleted = deleted
+}
+
+func (role *roleImpl) IsDeleted() bool {
```
adamcfraser, line 157: It was unexpected to me that the read accessor is public, but the write accessor is private. Is this intentional?

adamcfraser (changes requested):

auth/auth.go
JRascagneres: It needs to be public so that it can be used to calculate revocations. This is used in https://github.com/couchbase/sync_gateway/pull/4986/files#diff-c2dab3b01f62851ce6742b56d378463c64d11ed2a1f1d9da5055781e819c3a8cR231

auth/auth.go
JRascagneres: I was under the impression the previous behaviour would end up returning nil but I can check this.

auth/role.go
JRascagneres: Similarly, needs to be read in the https://github.com/couchbase/sync_gateway/pull/4986/files#diff-c2dab3b01f62851ce6742b56d378463c64d11ed2a1f1d9da5055781e819c3a8cR231 case. Whereas any writing is handled by other functions in auth.

auth/auth.go
JRascagneres: If its deleted I think we still want to return nil here so shouldn't use GetRoleIncDeleted. I can try to write a test for this.

auth/auth.go
JRascagneres: Yeah. The previous behaviour would return nil. Can discuss in sync up if need be

auth/auth.go
JRascagneres: If you want I can make them private in this PR as they have no reason not to be in this PR? Then can make them public again once required.

auth/auth.go
JRascagneres: Will write a test but pretty sure we want to get GetRole here so its nil in the case of a delete. This was the previous behaviour and then the callback would handle what to do in the situation where the value is nil.

db/users.go
JRascagneres: Moved user back to basically using same as before and split the user and role delete.

rest/admin_api.go
```diff
@@ -711,19 +711,17 @@ func (h *handler) deleteUser() error {
...
 	h.assertAdminOnly()
-	role, err := h.db.Authenticator().GetRole(mux.Vars(h.rq)["name"])
-	if role == nil {
-		if err == nil {
-			err = kNotFoundError
-		}
-		return err
+	purge := h.getBoolQuery("purge")
+	found, err := h.db.DeleteRole(mux.Vars(h.rq)["name"], purge)
+	if !found {
```
adamcfraser: It looks like db.DeleteRole is returning found=false for any GetRole error, which results in deleteRole always returning kNotFoundError in that case.  This seems like a change from the previous handling.

In additional, it seems like we could simplify db.DeleteRole to just return error, and have it return kNotFoundError when it's not found.

adamcfraser (changes requested):
Generally looks good, one question about deleteRole error handling.

rest/admin_api.go
JRascagneres: Done this and switched the 'if' ordering a little so it more closely matches the existing implementation.

db/users.go
```diff
@@ -83,6 +83,25 @@ func (dbc *DatabaseContext) GetPrincipal(name string, isUser bool) (info *Princi
...
+func (db *DatabaseContext) DeleteRole(name string, purge bool) error {
+	authenticator := db.Authenticator()
+
+	role, err := authenticator.GetRole(name)
+	if role == nil {
+		if err == nil {
+			return base.ErrNotFound
+		}
+		return err
+	}
```
bbrks: I think this is a bit easier to reason about, as it doesn't make assumptions about role _always_ being nil when we get an error back.

```suggestion
	if err != nil {
		return err
	}

	if role == nil {
		return base.ErrNotFound
	}
```

bbrks (changes requested):
One minor comment - but otherwise looks good.

db/users.go
JRascagneres: I did this so it matched the existing checks but happy to change it.

adamcfraser (approved):

PR #4986, opened 2021-04-21
CBG-1365: Calculate channels to revoke

JRascagneres:
Uses the built history to calculate channels to revoke along with the triggeredBy value for each of the channels.

### DIFF
--- a/auth/user.go
+++ b/auth/user.go
@@ -183,9 +183,17 @@ func (user *userImpl) RoleHistory() TimedSetHistory {
 	return user.RoleHistory_
 }
 
-func (user *userImpl) RevokedChannels(since uint64) map[string]uint64 {
+// Type is used to store a pair of channel names to triggered by sequences
+// If there already exists the channel in the map it'll only update its sequence if the sequence being added is greater
+type RevokedChannels map[string]uint64
 
-	// Grab all channels user has access to
+func (revokedChannels RevokedChannels) add(chanName string, triggeredBy uint64) {
+	if currentVal, ok := revokedChannels[chanName]; !ok || triggeredBy > currentVal {
+		revokedChannels[chanName] = triggeredBy
+	}
+}
+
+func (user *userImpl) RevokedChannels(since uint64) RevokedChannels {
 	accessibleChannels := user.InheritedChannels()
 
 	// Get revoked roles
@@ -194,7 +202,7 @@ func (user *userImpl) RevokedChannels(since uint64) map[string]uint64 {
 	for roleName, history := range roleHistory {
 		if !user.RoleNames().Contains(roleName) {
 			for _, entry := range history.Entries {
-				if entry.StartSeq <= since && entry.EndSeq >= since {
+				if entry.StartSeq <= since && entry.EndSeq > since {
 					mostRecentEndSeq := history.Entries[len(history.Entries)-1]
 					rolesToRevoke[roleName] = mostRecentEndSeq.EndSeq
 				}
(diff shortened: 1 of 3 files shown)


auth/user.go
```diff
@@ -183,6 +183,98 @@ func (user *userImpl) RoleHistory() TimedSetHistory {
 	return user.RoleHistory_
 }
 
+func (user *userImpl) RevokedChannels(since uint64) map[string]uint64 {
```
adamcfraser: Add a comment describing the contents of the returned map[string]uint64.  Even after reviewing the entire function I'm not positive what the sequence value represents (it seems like 'end seq for latest revocation', but I'm not sure how that information should be used).

auth/user.go
```diff
@@ -183,6 +183,98 @@ func (user *userImpl) RoleHistory() TimedSetHistory {
...
+
+	// Grab all channels user has access to
```
adamcfraser: This comment is probably unnecessary - the call to InheritedChannels() is self-documenting.

auth/user.go
```diff
@@ -183,6 +183,98 @@ func (user *userImpl) RoleHistory() TimedSetHistory {
...
+	accessibleChannels := user.InheritedChannels()
+
+	// Get revoked roles
+	rolesToRevoke := map[string]uint64{}
+	roleHistory := user.RoleHistory()
+	for roleName, history := range roleHistory {
+		if !user.RoleNames().Contains(roleName) {
+			for _, entry := range history.Entries {
+				if entry.StartSeq <= since && entry.EndSeq >= since {
```
adamcfraser: Having both StartSeq and EndSeq comparisons as "or equal to" doesn't look correct - my guess is that it should be simply `entry.EndSeq > since`, as a call with since=endSeq usually implies everything up to and including endSeq has already been replicated.  But unit tests with boundary conditions would be the best approach.

auth/user.go
```diff
@@ -183,6 +183,98 @@ func (user *userImpl) RoleHistory() TimedSetHistory {
...
+					mostRecentEndSeq := history.Entries[len(history.Entries)-1]
+					rolesToRevoke[roleName] = mostRecentEndSeq.EndSeq
+				}
+			}
+		}
+	}
+
+	// Store revoked channels to return
```
adamcfraser, line 222: Are both of these inline functions strictly necessary?  For addToCombined, it seems like this would be cleaner as a custom type for the map[string]uint64, with a function like "add(string, uint64)" (that did the check addToCombined is doing).

For revokeChannelHistoryProcessing, what is the benefit of using as an inline function instead of a standalone function?

adamcfraser (changes requested):
A few comments, and  I'd like a walkthrough of RevokedChannels once those have been dealt with, have a few questions that are better talked through.

auth/user.go
JRascagneres: Basically the value that should be used for triggeredBy, but I'll add a comment yeah.

auth/user.go
JRascagneres: I like the idea for the custom type, that works.
revokeChannelHistoryProcessing - benefit was just its only used internally for that function so seems unnecessary to have it 'bloat' externally but I can do. When a function only ever needs to be used inside another I tend to go for these inline functions. Also has access to the scope in this case to add to the map directly rather than via a parameter.

auth/user.go
JRascagneres: Yeah, you're correct would end up with unnecessary revocation message. Confirmed with test.

auth/user.go
JRascagneres: Note to myself: Addressed first part, second part not yet addressed. Can maybe discuss on Monday along with other.

adamcfraser (approved):

PR #4989, opened 2021-04-27
CBG-1410: Add document channel history

JRascagneres:
Adds Channel History pairs to document sync data.
Required changes to GrantHistorySequencePair type to allow entries with only start seq.

Added a test for it.

To Change / Why a draft:
Not happy with where the GrantHistorySequencePair type is. Need to move it really but unsure of where best to put it. Maybe channels? Where ChannelMap is now?

### DIFF
--- a/db/document.go
+++ b/db/document.go
@@ -844,9 +844,17 @@ func (doc *Document) updateChannelHistory(channelName string, seq uint64, additi
 		}
 	}
 
-	// First time addition
-	doc.ChanNames = append(doc.ChanNames, channelName)
-	doc.ChannelHistory = append(doc.ChannelHistory, []auth.GrantHistorySequencePair{{StartSeq: seq}})
+	// Haven't seen this channel before in history
+	// If its an addition go ahead and add an entry
+	// If its not an addition but we've not seen it before then its a legacy document. Add entry with start seq of 1.
+	if addition {
+		doc.ChanNames = append(doc.ChanNames, channelName)
+		doc.ChannelHistory = append(doc.ChannelHistory, []auth.GrantHistorySequencePair{{StartSeq: seq}})
+	} else {
+		doc.ChanNames = append(doc.ChanNames, channelName)
+		doc.ChannelHistory = append(doc.ChannelHistory, []auth.GrantHistorySequencePair{{StartSeq: 1, EndSeq: seq}})
+	}
+
 }
 
 // Updates the Channels property of a document object with current & past channels.
--- a/db/document.go
+++ b/db/document.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/couchbase/sync_gateway/auth"
 	"github.com/couchbase/sync_gateway/base"
 	"github.com/couchbase/sync_gateway/channels"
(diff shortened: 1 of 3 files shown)


db/document.go
```diff
@@ -829,6 +832,23 @@ func (doc *Document) UpdateExpiry(expiry uint32) {
...
+				doc.ChannelHistory[idx][len(doc.ChannelHistory[idx])-1].EndSeq = seq
+			} else {
+				doc.ChannelHistory[idx] = append(doc.ChannelHistory[idx], auth.GrantHistorySequencePair{StartSeq: seq})
+			}
+			return
+		}
+	}
+
+	// First time addition
+	doc.ChanNames = append(doc.ChanNames, channelName)
```
JRascagneres: Needs changing. This assumes that if there is no entry its going to be startSeq entry. In reality this could be removal of a document from before upgrade. Here we can check the addition bool and set start as 1 and add end if its a removal.

db/document.go
```diff
@@ -52,22 +53,24 @@ type AttachmentsMeta map[string]interface{} // AttachmentsMeta metadata as inclu
...
+	RoleAccess      UserAccessMap                     `json:"role_access,omitempty"`
+	Expiry          *time.Time                        `json:"exp,omitempty"`                     // Document expiry.  Information only - actual expiry/delete handling is done by bucket storage.  Needs to be pointer for omitempty to work (see https://github.com/golang/go/issues/4357)
+	Cas             string                            `json:"cas"`                               // String representation of a cas value, populated via macro expansion
+	Crc32c          string                            `json:"value_crc32c"`                      // String representation of crc32c hash of doc body, populated via macro expansion
+	Crc32cUserXattr string                            `json:"user_xattr_value_crc32c,omitempty"` // String representation of crc32c hash of user xattr
+	TombstonedAt    int64                             `json:"tombstoned_at,omitempty"`           // Time the document was tombstoned.  Used for view compaction
+	Attachments     AttachmentsMeta                   `json:"attachments,omitempty"`
+	ChanNames       []string                          `json:"chan_names"`
```
adamcfraser, line 72: Are these just active channels, or do they also include previous channels?  We might want to think a bit more about how these might be indexed and queried in future to determine whether to include both active and previous channels in a single set, or to separate them.  Having to cross-reference against another array to differentiate might cause issues.  An alternative might be to store as ["+chan1", "-chan2"] so that they are indexed together, but that would mean two scans at query time to pick up both cases.

db/document.go
adamcfraser: We'll need to think about when to generate the new format for legacy documents.

adamcfraser (changes requested):
I'm fine moving to either channels or base, I think either one makes sense.

I think we'll want to make sure we're landing on the best format for indexing, which probably means a bit more research on that side.

db/document.go
JRascagneres: ChanNames maps to ChannelHistory so would contain previous channels too. Like index 1 of chan names is associated with additions and removals of ChannelHistory in index 1

JRascagneres: You're right, need to discuss this I think.

PR #4990, opened 2021-04-30
CBG-1388 sg_collect does not package archived log files (*.log.gz)

sarathkumarsivan:
Changes to collect rotated and archived log files from the standard log locations.

### DIFF
--- a/tools/sgcollect_info
+++ b/tools/sgcollect_info
@@ -255,10 +255,27 @@ def make_collect_logs_tasks(zip_dir, sg_url, salt, should_redact):
     sg_tasks = []
 
     def lookup_std_log_files(files, dirs):
-        for std_log_file in [os.path.join(d, f) for f in files for d in dirs]:
-            if std_log_file not in sg_log_file_paths:
-                sg_tasks.append(add_file_task(sourcefile_path=std_log_file))
-                sg_log_file_paths[std_log_file] = std_log_file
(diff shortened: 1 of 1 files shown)


tools/sgcollect_info
```diff
@@ -255,10 +255,27 @@ def make_collect_logs_tasks(zip_dir, sg_url, salt, should_redact):
...
 
     def lookup_std_log_files(files, dirs):
-        for std_log_file in [os.path.join(d, f) for f in files for d in dirs]:
-            if std_log_file not in sg_log_file_paths:
-                sg_tasks.append(add_file_task(sourcefile_path=std_log_file))
-                sg_log_file_paths[std_log_file] = std_log_file
+        for dir in dirs:
+            for file in files:
+                name, ext = os.path.splitext(file)
+                # Collect rotated log files from the default log locations.
```
bbrks: I think the pattern is also capturing active/non-rotated log files, so perhaps include that as a comment so it's clear that's intentional.

bbrks (changes requested):
Looks good aside from a comment to clarify handling

tools/sgcollect_info
sarathkumarsivan: Yeah, the pattern collect both (active and rotated log files). It's reasonable to rephrase this comment.

adamcfraser (approved):

PR #4994, opened 2021-05-04
CBG-1410: Add document channel history

JRascagneres:
JSON field names need tweaking but putting this up as a draft for quick look over.

Adds channel history as discussed to document sync data with two arrays one being 'current' and another being 'old'.
Has a unit test to verify data is being set correctly.

### DIFF
--- a/db/document.go
+++ b/db/document.go
@@ -50,24 +50,32 @@ type UserAccessMap map[string]channels.TimedSet
 
 type AttachmentsMeta map[string]interface{} // AttachmentsMeta metadata as included in sync metadata
 
+type DocumentChannelHistoryEntry struct {
+	Name  string `json:"name"`
+	Start uint64 `json:"start"`
+	End   uint64 `json:"end"`
+}
+
 // The sync-gateway metadata stored in the "_sync" property of a Couchbase document.
 type SyncData struct {
-	CurrentRev      string              `json:"rev"`
-	NewestRev       string              `json:"new_rev,omitempty"` // Newest rev, if different from CurrentRev
-	Flags           uint8               `json:"flags,omitempty"`
-	Sequence        uint64              `json:"sequence,omitempty"`
-	UnusedSequences []uint64            `json:"unused_sequences,omitempty"` // unused sequences due to update conflicts/CAS retry
-	RecentSequences []uint64            `json:"recent_sequences,omitempty"` // recent sequences for this doc - used in server dedup handling
-	History         RevTree             `json:"history"`
-	Channels        channels.ChannelMap `json:"channels,omitempty"`
-	Access          UserAccessMap       `json:"access,omitempty"`
-	RoleAccess      UserAccessMap       `json:"role_access,omitempty"`
-	Expiry          *time.Time          `json:"exp,omitempty"`                     // Document expiry.  Information only - actual expiry/delete handling is done by bucket storage.  Needs to be pointer for omitempty to work (see https://github.com/golang/go/issues/4357)
-	Cas             string              `json:"cas"`                               // String representation of a cas value, populated via macro expansion
-	Crc32c          string              `json:"value_crc32c"`                      // String representation of crc32c hash of doc body, populated via macro expansion
-	Crc32cUserXattr string              `json:"user_xattr_value_crc32c,omitempty"` // String representation of crc32c hash of user xattr
-	TombstonedAt    int64               `json:"tombstoned_at,omitempty"`           // Time the document was tombstoned.  Used for view compaction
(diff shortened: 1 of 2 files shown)


db/document.go
```diff
@@ -50,24 +50,32 @@ type UserAccessMap map[string]channels.TimedSet
 
 type AttachmentsMeta map[string]interface{} // AttachmentsMeta metadata as included in sync metadata
 
+type DocumentChannelHistoryEntry struct {
```
adamcfraser: ChannelSetEntry

db/document.go
```diff
@@ -50,24 +50,32 @@ type UserAccessMap map[string]channels.TimedSet
...
+	RoleAccess        UserAccessMap                 `json:"role_access,omitempty"`
+	Expiry            *time.Time                    `json:"exp,omitempty"`                     // Document expiry.  Information only - actual expiry/delete handling is done by bucket storage.  Needs to be pointer for omitempty to work (see https://github.com/golang/go/issues/4357)
+	Cas               string                        `json:"cas"`                               // String representation of a cas value, populated via macro expansion
+	Crc32c            string                        `json:"value_crc32c"`                      // String representation of crc32c hash of doc body, populated via macro expansion
+	Crc32cUserXattr   string                        `json:"user_xattr_value_crc32c,omitempty"` // String representation of crc32c hash of user xattr
+	TombstonedAt      int64                         `json:"tombstoned_at,omitempty"`           // Time the document was tombstoned.  Used for view compaction
+	Attachments       AttachmentsMeta               `json:"attachments,omitempty"`
+	ChannelHistory    []DocumentChannelHistoryEntry `json:"channel_history"`
```
adamcfraser: ChannelSet/channel_set

db/document.go
```diff
@@ -50,24 +50,32 @@ type UserAccessMap map[string]channels.TimedSet
...
+	OldChannelHistory []DocumentChannelHistoryEntry `json:"old_channel_history"`
```
adamcfraser: ChannelSetHistory/channel_set_history

db/document.go
```diff
@@ -829,6 +837,48 @@ func (doc *Document) UpdateExpiry(expiry uint32) {
...
+			} else {
+				doc.ChannelHistory[idx].End = seq
+			}
+			return
+		}
+	}
+
+	// If we get to this point there is no existing entry
+	// If addition we can just simply add it
+	// If its a removal / not addition then its a legacy document. Start seq is 1
```
adamcfraser: Can you add 'Start seq is 1 because we don't know when legacy channels were added'.

db/document.go
```diff
@@ -829,6 +837,48 @@ func (doc *Document) UpdateExpiry(expiry uint32) {
...
+		if historyEntry.Name == channelName {
+			// If we are here there is an existing entry for this channel
+			// If addition we need to:
+			// - Move existing history entry to old channels
+			// - Remove existing entry from current channels
+			// - Add new entry with start seq
+			// If removal / not addition then we can simply add the end seq
+			if addition {
+				doc.OldChannelHistory = append(doc.OldChannelHistory, historyEntry)
+				doc.ChannelHistory = append(doc.ChannelHistory[:idx], doc.ChannelHistory[idx+1:]...)
```
adamcfraser: Can just replace the existing index in this case instead of these appends.

db/document.go
```diff
@@ -829,6 +837,48 @@ func (doc *Document) UpdateExpiry(expiry uint32) {
...
+	for idx, historyEntry := range doc.ChannelHistory {
+		if historyEntry.Name == channelName {
+			// If we are here there is an existing entry for this channel
+			// If addition we need to:
+			// - Move existing history entry to old channels
+			// - Remove existing entry from current channels
+			// - Add new entry with start seq
+			// If removal / not addition then we can simply add the end seq
+			if addition {
+				doc.OldChannelHistory = append(doc.OldChannelHistory, historyEntry)
```
adamcfraser: Think about whether we should be defensive here and not write items to OldChannelHistory if historyEntry.end isn't present.

adamcfraser (changes requested):

db/document.go
JRascagneres: I'd be very surprised if this happens as we're plugging into the existing channel addition / removal, however, it doesn't hurt to be defensive.
Added a check and if there is no existing 'end' when attempting to do an addition we're going to just return. This means we just keep the existing record with the existing start and no end. In terms of how this could affect revocation I believe worse case is we end up sending a revocation for a doc ID which the user has never had. This isn't perfect but without the information of when a removal was carried out our options are limited.

adamcfraser (approved):

PR #4995, opened 2021-05-07
Add clistruct package to map struct fields to command line flags

bbrks:
Not wired up to SG yet - but this PR adds a package to support future functionality.

### Usage

- Use `cli:"foo"` struct tag to create a flag called `-foo`
- Optionally specify help using `help:"foo is something"` to describe the flag.

Fields must be exported, and of a concrete type. Does not currently support slices or maps.

`MustRegisterFlags` panics for structs that try to map invalid fields or types to allow early detection of invalid usages.

---

### Example

Supports nested structs, and dot-separates flags:
```go
struct{
  A struct {
    B bool `cli:"b"`
  } `cli:"a"`
}
```
results in a flag called `-a.b`

Tags can empty to allow for "embedding" the child flags:
```go
struct {
  A struct {
    B bool `cli:"b"`
  } `cli:""`
}
```
results in a flag called `-b`

---

### Doc

Screenshot 2021-05-07 at 17 28 10

### DIFF
--- a/base/clistruct/clistruct.go
+++ b/base/clistruct/clistruct.go
@@ -32,7 +32,12 @@ func registerFlags(fs *flag.FlagSet, cliStruct interface{}) error {
 		return errors.New("expected flag.FlagSet in RegisterFlags but got nil")
 	}
 
-	if err := register(fs, reflectStructValue(cliStruct), ""); err != nil {
+	refVal, err := reflectStructValue(cliStruct)
+	if err != nil {
+		return err
+	}
+
+	if err := register(fs, refVal, ""); err != nil {
 		return err
 	}
 
--- a/base/clistruct/clistruct.go
+++ b/base/clistruct/clistruct.go
@@ -3,6 +3,7 @@ package clistruct
 
 import (
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"reflect"
@@ -28,11 +29,11 @@ func MustRegisterFlags(fs *flag.FlagSet, cliStruct interface{}) {
 // registerFlags is the non-panicking version of MustRegisterFlags, for testing.
 func registerFlags(fs *flag.FlagSet, cliStruct interface{}) error {
 	if fs == nil {
-		panic("expected flag.FlagSet in RegisterFlags but got nil")
+		return errors.New("expected flag.FlagSet in RegisterFlags but got nil")
 	}
 
 	if err := register(fs, reflectStructValue(cliStruct), ""); err != nil {
-		panic(err.Error())
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


base/clistruct/clistruct.go
```diff
@@ -0,0 +1,152 @@
...
+	cliStructHelpTag = "help"
+)
+
+// MustRegisterFlags will register flags on the given FlagSet from cli struct tags in the given struct.
+// FlagSet.Parse() must be called _after_ this has been run.
+//
+// Panics if the given struct cannot be assigned to a FlagSet (unsupported types, duplicate flags, etc.)
+func MustRegisterFlags(fs *flag.FlagSet, cliStruct interface{}) {
+	if err := registerFlags(fs, cliStruct); err != nil {
+		panic(err.Error())
```
adamcfraser, line 24: What's the benefit of panicking here instead of having MustRegisterFlags return error, and leave the error handling to the caller?

base/clistruct/clistruct.go
```diff
@@ -0,0 +1,152 @@
...
+	case *json.Number:
+		fs.Var((*jsonNumberFlagValue)(v), flagName, flagHelp)
+	default:
+		// See docs: unsafe.Pointer (5) Conversion of the result of ... reflect.Value.UnsafeAddr
+		// uintptr -> unsafe.Pointer conversion MUST be done without an intermediate variable
+		p := unsafe.Pointer(fieldValue.UnsafeAddr())
+
+		// we can only map struct field types which have a matching flag var type
+		switch fieldValue.Type().Kind() {
+		case reflect.Bool:
```
adamcfraser: What's the difference between what we're doing here, and the handling in https://github.com/golang/tools/blob/v0.1.0/internal/tool/tool.go#L187 to generate flags for a reflect.Value?  The latter looks like it's handling these types directly in the type switch above.

adamcfraser (commented):
A couple of questions for discussion.

base/clistruct/clistruct.go
bbrks: Huh, that looks way better, seems easy to get rid of this reflect Kind switch (and unsafe in the process) - which makes me more comfortable putting this directly in `base`.

base/clistruct/clistruct.go
bbrks: Mainly for visibility (to us at dev time) - given we'll be registering this on startup with a struct, and it's known at startup time whether that's valid or not for a given build.

I'd be happy enough switching this to return `err` and ensure callers log as a fatal error.

bbrks:
Closing - will revisit after I've worked through changes based on specific use-cases.

PR #4997, opened 2021-05-12
CBG-1443: Avoid rebuild on invalidate

JRascagneres:
Implementation to avoid a GetUser / GetRole operation before performing an invalidation to avoid rebuild work.

If possible performs subdoc operation rather than a Get / Set (Update).

Added unit tests to ensure same behaviour remains.

- [x] https://github.com/couchbaselabs/walrus/pull/57
- [x] https://github.com/couchbase/sg-bucket/pull/63
- [x] http://uberjenkins.sc.couchbase.com:8080/view/All/job/sync-gateway-integration/764/
Couple of failures but ran with no problems locally

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -322,11 +322,14 @@ func (auth *Authenticator) InvalidateChannels(name string, isUser bool, invalSeq
 		docID = docIDForRole(name)
 	}
 
+	base.Infof(base.KeyAccess, "Invalidate access of %q", base.UD(name))
+
 	if auth.bucket.IsSupported(sgbucket.DataStoreFeatureSubdocOperations) {
 		err := auth.bucket.SubdocInsert(docID, "channel_inval_seq", 0, invalSeq)
-		if base.IsDocNotFoundError(err) {
-			return nil
+		if err != nil && !base.IsDocNotFoundError(err) && !base.IsSubDocPathExistsError(err) {
+			return err
 		}
+		return nil
 	}
 
 	_, err := auth.bucket.Update(docID, 0, func(current []byte) (updated []byte, expiry *uint32, delete bool, err error) {
@@ -362,11 +365,14 @@ func (auth *Authenticator) InvalidateChannels(name string, isUser bool, invalSeq
 func (auth *Authenticator) InvalidateRoles(username string, invalSeq uint64) error {
 	docID := docIDForUser(username)
 
+	base.Infof(base.KeyAccess, "Invalidate roles of %q", base.UD(username))
+
 	if auth.bucket.IsSupported(sgbucket.DataStoreFeatureSubdocOperations) {
 		err := auth.bucket.SubdocInsert(docID, "role_inval_seq", 0, invalSeq)
-		if base.IsDocNotFoundError(err) {
-			return nil
+		if err != nil && !base.IsDocNotFoundError(err) && !base.IsSubDocPathExistsError(err) {
+			return err
 		}
+		return nil
 	}
 
 	_, err := auth.bucket.Update(docID, 0, func(current []byte) (updated []byte, expiry *uint32, delete bool, err error) {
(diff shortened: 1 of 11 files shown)


auth/auth_test.go
```diff
@@ -2431,49 +2430,6 @@ func TestRoleSoftDelete(t *testing.T) {
 	assert.Equal(t, expectedChannelHistory, role.ChannelHistory()["!"].Entries[0])
 }
 
-func TestRoleSoftDeleteCasMismatch(t *testing.T) {
```
JRascagneres, line 2434: This test was removed as it tested the casUpdatePrincipal work done on a role and there is no longer one of these operations being performed for a role.

auth/auth.go
```diff
@@ -309,39 +309,62 @@ func (auth *Authenticator) UpdateSequenceNumber(p Principal, seq uint64) error {
...
+		}
 	}
-	return auth.casUpdatePrincipal(p, invalidateChannelsCallback)
+	return nil
+
 }
 
 // Invalidates the role list of a user by saving its Roles() property as nil.
-func (auth *Authenticator) InvalidateRoles(user User, invalSeq uint64) error {
+func (auth *Authenticator) InvalidateRoles(user string, invalSeq uint64) error {
```
adamcfraser: We haven't discussed previously, but if a single document triggers invalidation of both roles and channels for a user, can we do that in a single operation?

auth/auth.go
```diff
@@ -309,39 +309,62 @@ func (auth *Authenticator) UpdateSequenceNumber(p Principal, seq uint64) error {
...
 
-		if p == nil || p.Channels() == nil {
-			return p, base.ErrUpdateCancel
+		cas, err := auth.bucket.Get(docID, &princ)
+		if err != nil {
+			return err
 		}
 
-		base.Infof(base.KeyAccess, "Invalidate access of %q", base.UD(p.Name()))
+		princ.SetChannelInvalSeq(invalSeq)
```
adamcfraser: Can you remind me why we need to set invalSeq to the most recent invalidation, instead of the earliest invalidation, in the case where there are multiple invalidations without an auth?

auth/auth.go
```diff
@@ -309,39 +309,62 @@ func (auth *Authenticator) UpdateSequenceNumber(p Principal, seq uint64) error {
...
+			princ = &userImpl{}
+			docID = docIDForUser(name)
+		} else {
+			princ = &roleImpl{}
+			docID = docIDForRole(name)
+		}
 
-		if p == nil || p.Channels() == nil {
-			return p, base.ErrUpdateCancel
+		cas, err := auth.bucket.Get(docID, &princ)
```
adamcfraser: This type of Get and CAS-safe Set is already encapsulated in bucket.Update, including CAS retry.  Is there a reason we wouldn't want to use that here?

adamcfraser (commented):
A few comments/questions.

auth/auth.go
JRascagneres: We need to use the earliest. I missed it here, thanks for catching!

auth/auth.go
JRascagneres: It might be possible. Will look into it. Guess it depends how it's triggered

auth/auth.go
JRascagneres: Probably just blanked on that. Will switch

auth/auth.go
JRascagneres: There's quite a lot of work around where we check for changed principals so I think it'd be a bigger piece of work. We could possibly flag this for future improvement?

adamcfraser: Makes sense.

base/error.go
```diff
@@ -182,3 +183,17 @@ func IsDocNotFoundError(err error) bool {
 		return false
 	}
 }
+
+func IsSubDocPathExistsError(err error) bool {
```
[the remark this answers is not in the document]
JRascagneres: As discussed modified existing one to not be on the bucket API and then removed my new one.

adamcfraser (approved):

PR #4998, opened 2021-05-12
CBG-1390 Add warning threshold for excessively large number of channels per user

sarathkumarsivan:
Changes to add a new warning about channel count per user, to give early warning about running into memory issues with extremely high numbers of channels per user.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -24,9 +24,10 @@ import (
 
 /** Manages user authentication for a database. */
 type Authenticator struct {
-	bucket            base.Bucket
-	channelComputer   ChannelComputer
-	sessionCookieName string // Custom per-database session cookie name
+	bucket                   base.Bucket
+	channelComputer          ChannelComputer
+	sessionCookieName        string // Custom per-database session cookie name
+	channelsWarningThreshold *uint32
 }
 
 // Interface for deriving the set of channels and roles a User/Role has access to.
@@ -57,6 +58,14 @@ func (auth *Authenticator) SessionCookieName() string {
 
 func (auth *Authenticator) SetSessionCookieName(cookieName string) {
 	auth.sessionCookieName = cookieName
+}
+
+func (auth *Authenticator) ChannelsWarningThreshold() *uint32 {
+	return auth.channelsWarningThreshold
+}
+
+func (auth *Authenticator) SetChannelsWarningThreshold(channelsWarningThreshold *uint32) {
+	auth.channelsWarningThreshold = channelsWarningThreshold
 }
 
 func docIDForUserEmail(email string) string {
@@ -394,7 +403,7 @@ func (auth *Authenticator) InvalidateRoles(username string, invalSeq uint64) err
 
 		user.SetRoleInvalSeq(invalSeq)
 
-		updated, err = base.JSONMarshal(user)
+		updated, err = base.JSONMarshal(&user)
 		return updated, nil, false, err
 	})
 
--- a/auth/user.go
+++ b/auth/user.go
@@ -15,6 +15,7 @@ import (
 	"math"
 	"net/http"
 	"regexp"
+	"sync"
 
 	"golang.org/x/crypto/bcrypt"
 
@@ -28,6 +29,10 @@ type userImpl struct {
 	userImplBody
 	auth  *Authenticator
 	roles []Role
+
+	// warnChanThresholdOnce ensures that the check for channels
+	// per user threshold is only performed exactly once.
+	warnChanThresholdOnce sync.Once
 }
 
 // Marshallable data is stored in separate struct from userImpl,
@@ -483,6 +488,17 @@ func (user *userImpl) InheritedChannels() ch.TimedSet {
 		roleSince := user.RoleNames()[role.Name()]
 		channels.AddAtSequence(role.Channels(), roleSince.Sequence)
 	}
+
+	user.warnChanThresholdOnce.Do(func() {
+		if channelsPerUserThreshold := user.auth.ChannelsWarningThreshold(); channelsPerUserThreshold != nil {
+			channelCount := len(channels)
+			if uint32(channelCount) >= *channelsPerUserThreshold {
+				base.Warnf("User ID: %v channel count: %d exceeds %d for channels per user warning threshold",
+					base.UD(user.Name()), channelCount, *channelsPerUserThreshold)
+			}
+		}
+	})
+
 	return channels
 }
 
--- a/base/constants.go
+++ b/base/constants.go
@@ -153,9 +153,10 @@ var (
 	}
 
 	// Default warning thresholds
-	DefaultWarnThresholdXattrSize      = 0.9 * float64(couchbaseMaxSystemXattrSize)
-	DefaultWarnThresholdChannelsPerDoc = uint32(50)
-	DefaultWarnThresholdGrantsPerDoc   = uint32(50)
+	DefaultWarnThresholdXattrSize       = 0.9 * float64(couchbaseMaxSystemXattrSize)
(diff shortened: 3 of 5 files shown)


auth/auth.go
```diff
@@ -58,6 +59,14 @@ func (auth *Authenticator) SetSessionCookieName(cookieName string) {
 	auth.sessionCookieName = cookieName
 }
 
+func (auth *Authenticator) ChannelsPerUser() *uint32 {
+	return auth.channelsPerUser
+}
+
+func (auth *Authenticator) SetChannelsPerUser(channelsPerUser *uint32) {
```
bbrks: The variable and method name make it sound like a limit. Can we change it to be called a warn threshold like the others?

bbrks (changes requested):
`channels` is not taking into account channels that come from roles, which means I think we need to move where the check is being done. See inline comment.

auth/auth.go
```diff
@@ -183,6 +192,15 @@ func (auth *Authenticator) rebuildChannels(princ Principal) error {
 	}
 
 	base.Infof(base.KeyAccess, "Recomputed channels for %q: %s", base.UD(princ.Name()), base.UD(channels))
+
+	if channelsPerUserThreshold := auth.ChannelsPerUser(); channelsPerUserThreshold != nil {
+		channelCount := len(channels)
```
bbrks: I played around with this locally and it looks like `channels` is not the full set of expanded user channels (accessible via roles).

```
$ curl http://localhost:4985/db1/_role/bar
{"name":"bar","all_channels":["!","z"]}

$ curl http://localhost:4985/db1/_user/ben
{"name":"ben","all_channels":["!","a","c","d","z"],"roles":["bar"]}
```

```
2021-05-13T12:54:55.642+01:00 [INF] HTTP:  #046: GET /db1/_user/ben (as ADMIN)
2021-05-13T12:54:55.642+01:00 [INF] Access: Recomputed channels for "ben": !:1,a:9,c:14,d:16
2021-05-13T12:54:55.642+01:00 [WRN] User ID: ben channel count: 4 exceeds 2 for channels per user warning threshold -- auth.(*Authenticator).rebuildChannels() at auth.go:199
```

I think you'll have to do this warning for users after rebuilding channels and roles inside `getPrincipal`.

auth/auth.go
```diff
@@ -183,6 +192,15 @@ func (auth *Authenticator) rebuildChannels(princ Principal) error {
...
+		if uint32(channelCount) >= *channelsPerUserThreshold {
+			base.Warnf("User ID: %v channel count: %d exceeds %d for channels per user warning threshold",
```
bbrks: This is logged for roles too, which might be a valid thing to do, but the log message is confusing as it doesn't include the `role:` prefix for the principal name.

```
$ curl -X PUT http://localhost:4985/db1/acc3 -H 'Content-Type: application/json' -d '{"user":"role:bar","access":"z"}'
{"id":"acc3","ok":true,"rev":"1-5f2ab86a9ade4664638a2db2e4fb945c"}

$ curl -X PUT http://localhost:4985/db1/_role/bar -d '{}'
```

```
2021-05-13T12:54:00.712+01:00 [INF] HTTP:  #042: PUT /db1/_role/bar (as ADMIN)
2021-05-13T12:54:00.712+01:00 [INF] Access: Recomputed channels for "bar": !:1,z:11
2021-05-13T12:54:00.712+01:00 [WRN] User ID: bar channel count: 2 exceeds 2 for channels per user warning threshold -- auth.(*Authenticator).rebuildChannels() at auth.go:199
```

auth/auth.go
sarathkumarsivan: I've just refactored it to channelsWarningThreshold.

auth/auth.go
sarathkumarsivan: Yeah, that makes sense. it can be moved to getPrincipal.

auth/auth.go
sarathkumarsivan: We can get rid of by moving the logging to getPrincipal.

auth/auth.go
```diff
@@ -135,6 +144,14 @@ func (auth *Authenticator) getPrincipal(docID string, factory func() Principal)
 				}
 				changed = true
 			}
+
+			if channelsPerUserThreshold := auth.ChannelsWarningThreshold(); channelsPerUserThreshold != nil {
+				channelCount := len(user.Channels())
```
bbrks: As discussed previously, we want this to take into account inherited channels too - those granted via roles. `user.Channels()` only fetches the channels associated directly with the user.

bbrks (changes requested):
We're still not fetching inherited channels.

bbrks:
Also can you fix the commit message above? It seems unrelated.

```
@sarathkumarsivan 13a5561 
CBG-1388 Rephrase the comments to indicate that file filter pattern collect both active and rotated logs
```

auth/auth.go
sarathkumarsivan: Yeah, user.InheritedChannels() can be used instead of user.Channels(), which returns every channel the user has access to, including those inherited from Roles.

sarathkumarsivan:
Fixed the wrong commit message and the test failures.

auth/user.go
```diff
@@ -397,6 +397,13 @@ func (user *userImpl) InheritedChannels() ch.TimedSet {
 		roleSince := user.RoleNames()[role.Name()]
 		channels.AddAtSequence(role.Channels(), roleSince.Sequence)
 	}
+	if channelsPerUserThreshold := user.auth.ChannelsWarningThreshold(); channelsPerUserThreshold != nil {
```
bbrks: We said we should add a `sync.Once` or something to stop this from being repeatedly logged for the same user in a single long-running changes that are issuing repeated `InheritedChannels()` calls.

bbrks (changes requested):

auth/user.go
sarathkumarsivan: Makes sense, added changes to ensures that the check for channels per user threshold is only performed exactly once.

auth/role.go
```diff
@@ -32,6 +33,10 @@ type roleImpl struct {
 	Deleted           bool            `json:"deleted,omitempty"`
 	vbNo              *uint16
 	cas               uint64
+
+	// warnChanThresholdOnce ensures that the check for channels
+	// per user threshold is only performed exactly once.
+	warnChanThresholdOnce sync.Once
```
bbrks: Why is this on `roleImpl` instead of `userImpl`?

bbrks (changes requested):

auth/role.go
sarathkumarsivan: I agree, it should be on userImpl.

auth/role.go
```diff
@@ -12,13 +12,12 @@ package auth
...
+	"github.com/couchbase/sync_gateway/base"
+	ch "github.com/couchbase/sync_gateway/channels"
 	"net/http"
 	"regexp"
 	"strconv"
 	"strings"
-
-	"github.com/couchbase/sync_gateway/base"
-	ch "github.com/couchbase/sync_gateway/channels"
 )
```
bbrks: This is super nit-picky, but I'm fairly sure our tooling (both Goland and `goimports`) should handle this by separating stdlib and third-party imports unless explicitly moved into the same group. Is there a reason why this and the `bcrypt` import in the file below changed?

bbrks (changes requested):

auth/role.go
sarathkumarsivan: This was accidental — committed while removing the old import statement for sync package.

adamcfraser (approved):

PR #5000, opened 2021-05-13
CBG-1367: Send revocation over changes feed

JRascagneres:
Should be good for initial review.
Tests should provide adequate coverage.

- [x] http://uberjenkins.sc.couchbase.com:8080/view/All/job/sync-gateway-integration/771/
One failure for the above but below passed fine so probably one-off
- [x] http://uberjenkins.sc.couchbase.com:8080/view/All/job/sync-gateway-integration/772/

### DIFF
--- a/auth/user.go
+++ b/auth/user.go
@@ -289,6 +289,91 @@ func (user *userImpl) RevokedChannels(since uint64) RevokedChannels {
 	return combinedRevokedChannels
 }
 
+// Calculate periods where user somehow had access to the given channel
+func (user *userImpl) ChannelGrantedPeriods(chanName string, latestSequence uint64) ([]GrantHistorySequencePair, error) {
+	var resultPairs []GrantHistorySequencePair
+
+	// Grab user history and use this to begin the resultPairs
+	userChannelHistory, ok := user.ChannelHistory()[chanName]
+	if ok {
+		resultPairs = userChannelHistory.Entries
+	}
+
+	// Iterate over current user channels and add to resultPairs
+	for channelName, chanInfo := range user.Channels() {
+		if channelName != chanName {
+			continue
+		}
+		resultPairs = append(resultPairs, GrantHistorySequencePair{
+			StartSeq: chanInfo.Sequence,
+			EndSeq:   latestSequence,
+		})
+	}
+
+	// Small function which takes the two start seqs and two end seqs and calculates the intersection
+	compareAndAddPair := func(startSeq1, startSeq2, endSeq1, endSeq2 uint64) {
+		start := base.MaxUint64(startSeq1, startSeq2)
+		end := base.MinUint64(endSeq1, endSeq2)
+		if start < end {
+			resultPairs = append(resultPairs, GrantHistorySequencePair{
+				StartSeq: start,
+				EndSeq:   end,
+			})
+		}
+	}
+
+	// Iterate over current roles
+	for _, currentRole := range user.GetRoles() {
+
+		// Grab pairs from channel history on current roles
+		roleChannelHistory, ok := currentRole.ChannelHistory()[chanName]
+		if ok {
+			for _, roleChannelHistoryEntry := range roleChannelHistory.Entries {
+				compareAndAddPair(roleChannelHistoryEntry.StartSeq, currentRole.Sequence(), roleChannelHistoryEntry.EndSeq, latestSequence)
+			}
+		}
+
+		// Grab pairs from current channels on current roles
+		if currentRole.Channels().Contains(chanName) {
+			for _, channelInfo := range currentRole.Channels() {
+				resultPairs = append(resultPairs, GrantHistorySequencePair{
+					StartSeq: channelInfo.Sequence,
+					EndSeq:   latestSequence,
+				})
+			}
+		}
+	}
+
+	// Iterate over roles the user has had
+	for roleName, historyEntry := range user.RoleHistory() {
+		role, err := user.auth.GetRoleIncDeleted(roleName)
+		if err != nil {
+			return nil, err
+		}
+
+		// Iterate over history on old roles
+		roleChannelHistory, ok := role.ChannelHistory()[chanName]
+		if ok {
+			for _, roleChannelHistoryEntry := range roleChannelHistory.Entries {
+				for _, roleHistoryEntry := range historyEntry.Entries {
+					compareAndAddPair(roleChannelHistoryEntry.StartSeq, roleHistoryEntry.StartSeq, roleChannelHistoryEntry.EndSeq, roleHistoryEntry.EndSeq)
+				}
+			}
+		}
+
+		// Iterate over current channels on old roles
+		if role.Channels().Contains(chanName) {
+			for _, activeChannel := range role.Channels() {
+				for _, roleHistoryEntry := range historyEntry.Entries {
+					compareAndAddPair(activeChannel.Sequence, roleHistoryEntry.StartSeq, latestSequence, roleHistoryEntry.EndSeq)
+				}
+			}
+		}
+	}
+
+	return resultPairs, nil
+}
+
 // Returns true if the given password is correct for this user, and the account isn't disabled.
 func (user *userImpl) Authenticate(password string) bool {
 	if user == nil {
--- a/auth/user.go
+++ b/auth/user.go
@@ -12,6 +12,7 @@ package auth
 import (
 	"bytes"
 	"fmt"
+	"math"
 	"net/http"
 	"regexp"
 
@@ -372,6 +373,24 @@ func (user *userImpl) ChannelGrantedPeriods(chanName string, latestSequence uint
 	}
 
 	return resultPairs, nil
+}
+
+// Find latest revocation status. Used to determine whether we can resume a interrupted revocation feed or if we need
+// to start again from 0
+func (user *userImpl) FindMostRecentRevocationSeq(chanName string) (uint64, error) {
+	grantPeriods, err := user.ChannelGrantedPeriods(chanName, math.MaxUint64)
+	if err != nil {
+		return 0, err
+	}
+
+	highRevocationSeq := uint64(0)
+	for _, pair := range grantPeriods {
+		if pair.EndSeq > highRevocationSeq && pair.EndSeq != math.MaxUint64 {
+			highRevocationSeq = pair.EndSeq
+		}
+	}
+
+	return highRevocationSeq, nil
 }
 
 // Returns true if the given password is correct for this user, and the account isn't disabled.
--- a/auth/user.go
+++ b/auth/user.go
@@ -39,8 +39,8 @@ type userImplBody struct {
 	OldPasswordHash_ interface{}     `json:"passwordhash,omitempty"` // For pre-beta compatibility
 	ExplicitRoles_   ch.TimedSet     `json:"explicit_roles,omitempty"`
 	RolesSince_      ch.TimedSet     `json:"rolesSince"`
-	RoleInvalSeq     uint64          `json:"role_inval_seq,omitempty"`
-	RoleHistory_     TimedSetHistory `json:"role_history,omitempty"`
+	RoleInvalSeq     uint64          `json:"role_inval_seq,omitempty"` // Sequence at which the roles were invalidated. Data remains in RolesSince_ for history calculation.
+	RoleHistory_     TimedSetHistory `json:"role_history,omitempty"`   // Added to when a previously granted role is revoked. Calculated inside of rebuildRoles.
 
 	OldExplicitRoles_ []string `json:"admin_roles,omitempty"` // obsolete; declared for migration
 }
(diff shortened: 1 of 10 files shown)


db/changes.go
```diff
@@ -171,6 +172,151 @@ func (db *Database) AddDocInstanceToChangeEntry(entry *ChangeEntry, doc *Documen
...
 
+func (db *Database) buildRevokedFeed(singleChannelCache SingleChannelCache, options ChangesOptions, triggeredBy uint64) <-chan *ChangeEntry {
+	feed := make(chan *ChangeEntry, 1)
+	sinceVal := options.Since.Seq
+
+	queryLimit := db.Options.CacheOptions.ChannelQueryLimit
+	requestLimit := options.Limit
+
+	paginationOptions := options
+	paginationOptions.Since.Seq = 0
```
adamcfraser, line 183: In the case of an interrupted and resumed revocation (i.e. since value like seq:revocationSeq), we don't want to start again from zero.  The same is handled for backfill on the normal changesFeed today using the since value in ChangesOptions (I think), it feels like we want to do the same here.

adamcfraser (commented):
Still working my way through this PR, but one item to take a look at.

db/changes.go
JRascagneres: Added handling in the latest commit for this. Should handle this scenario and allow a resume if possible.

db/changes.go
```diff
@@ -171,6 +172,160 @@ func (db *Database) AddDocInstanceToChangeEntry(entry *ChangeEntry, doc *Documen
...
+			paginationOptions.Since.Seq = options.Since.TriggeredBy
+		}
+	}
+
+	go func() {
+		defer base.FatalPanicHandler()
+		defer close(feed)
+		var itemsSent int
+		var lastSeq uint64
+
```
adamcfraser, line 197: I don't think it would hurt to clone the comment from changesFeed about how the pagination is being computed.

db/changes.go
```diff
@@ -171,6 +172,160 @@ func (db *Database) AddDocInstanceToChangeEntry(entry *ChangeEntry, doc *Documen
...
+		for {
+			if requestLimit == 0 {
+				paginationOptions.Limit = queryLimit
+			} else {
+				remainingLimit := requestLimit - itemsSent
+				paginationOptions.Limit = base.MinInt(remainingLimit, queryLimit)
+			}
+
+			// Get changes from 0 to latest seq
+			base.TracefCtx(db.Ctx, base.KeyChanges, "Querying channel %q for revocations with options: %+v", base.UD(singleChannelCache.ChannelName()), paginationOptions)
```
adamcfraser:
```suggestion
			base.TracefCtx(db.Ctx, base.KeyChanges, "Querying channel %q for revocation with options: %+v", base.UD(singleChannelCache.ChannelName()), paginationOptions)
```

db/changes.go
```diff
@@ -171,6 +172,160 @@ func (db *Database) AddDocInstanceToChangeEntry(entry *ChangeEntry, doc *Documen
...
+			for _, logEntry := range changes {
+				seqID := SequenceID{
+					Seq:         logEntry.Sequence,
+					TriggeredBy: triggeredBy,
+				}
+
+				idOnlyRevocation := false
+
+				// We need to check whether a change / document sequence is greater than since.
+				// If its less than we can continue. Otherwise:
```
adamcfraser:
```suggestion
				// If its less than we can send a standard revocation with sequenceID as above. Otherwise:
```

db/changes.go
```diff
@@ -171,6 +172,160 @@ func (db *Database) AddDocInstanceToChangeEntry(entry *ChangeEntry, doc *Documen
...
+				// Because we only see the active revision of the document through the GetChanges query, we need to check
+				// whether the user had access to this document at the last since to know whether we need to  end a
+				// revocation message. We don't want to send a revocation message if the document has not seen before.
```
adamcfraser: I think the three lines above might be simplified as:
// Otherwise, we need to determine whether a previous revision of the document was in the 
// channel prior to the since value, and only send a revocation if this was the case.

db/changes.go
```diff
@@ -171,6 +172,160 @@ func (db *Database) AddDocInstanceToChangeEntry(entry *ChangeEntry, doc *Documen
...
+
+			paginationOptions.Since.Seq = lastSeq
+		}
+	}()
+
+	return feed
+}
+
+// Checks if a document needs to be revoked. This is used in the case where the since < doc sequence
+func (db *Database) docRequiresRevocation(docID, chanName string, since uint64) (bool, error) {
```
adamcfraser: I feel like this might be better named for what it's doing (something like wasDocInChannelAtSeq), instead of how we're using the results (revocation check).

db/changes.go
```diff
@@ -171,6 +172,160 @@ func (db *Database) AddDocInstanceToChangeEntry(entry *ChangeEntry, doc *Documen
...
+
+	// Get doc sync data so we can verify the docs grant history
+	syncData, err := db.GetDocSyncData(docID)
+	if err != nil {
+		return false, err
+	}
+
+	// Grab the latest sequence, this is used by the ChannelGrantedPeriods function for an 'endSeq' for current grants
```
adamcfraser: As discussed, hopefully can avoid the overhead of this retrieval.

db/changes.go
```diff
@@ -171,6 +172,160 @@ func (db *Database) AddDocInstanceToChangeEntry(entry *ChangeEntry, doc *Documen
...
+		if docHistoryEntry.Name != chanName {
+			continue
+		}
+
+		for _, accessPeriod := range channelAccessPeriods {
+			if accessPeriod.StartSeq > since || accessPeriod.EndSeq <= since {
+				continue
+			}
+
+			// If there is some overlap we can quit out as we know user has had the doc and we need to revoke
```
adamcfraser:
```suggestion
			// If there is some overlap we can quit out as we know user had access to the doc at seq since 
```

db/changes.go
```diff
@@ -171,6 +172,160 @@ func (db *Database) AddDocInstanceToChangeEntry(entry *ChangeEntry, doc *Documen
...
+
+				idOnlyRevocation := false
+
+				// We need to check whether a change / document sequence is greater than since.
+				// If its less than we can continue. Otherwise:
+				// Because we only see the active revision of the document through the GetChanges query, we need to check
+				// whether the user had access to this document at the last since to know whether we need to  end a
+				// revocation message. We don't want to send a revocation message if the document has not seen before.
+				if logEntry.Sequence > sinceVal {
+					requiresRevocation, err := db.docRequiresRevocation(logEntry.DocID, singleChannelCache.ChannelName(), sinceVal)
```
adamcfraser: docRequiresRevocation is calling ChannelGrantedPeriods, which calls user.GetRoles().  It looks like this should only be loading the role set from the bucket once (based on the nil check in GetRoles()) - just wanted to confirm this is the case.

db/changes.go
```diff
@@ -281,6 +436,17 @@ func makeChangeEntry(logEntry *LogEntry, seqID SequenceID, channelName string) C
 	return change
 }
 
+func makeRevocationChangeEntry(logEntry *LogEntry, seqID SequenceID, channelName string, idOnly bool) ChangeEntry {
```
adamcfraser: Does 'omitRevID' make more sense than idOnly, given that we're sending more than just id?  In particular I wasn't sure at first whether we still send seqID.seq in the case where we're omitting rev.

db/changes.go
```diff
@@ -171,6 +172,160 @@ func (db *Database) AddDocInstanceToChangeEntry(entry *ChangeEntry, doc *Documen
...
+						return
+					}
+
+					idOnlyRevocation = true
+				}
+
+				change := makeRevocationChangeEntry(logEntry, seqID, singleChannelCache.ChannelName(), idOnlyRevocation)
+				lastSeq = logEntry.Sequence
+
+				base.DebugfCtx(db.Ctx, base.KeyChanges, "Channel feed processing seq:%v in channel %s", seqID, base.UD(singleChannelCache.ChannelName()))
```
adamcfraser: Whether or not we're sending a no-revision changes message is useful debugging information - would be nice to include that information here.

db/changes.go
```diff
@@ -171,6 +172,160 @@ func (db *Database) AddDocInstanceToChangeEntry(entry *ChangeEntry, doc *Documen
...
+
+				select {
+				case <-options.Terminator:
+					base.DebugfCtx(db.Ctx, base.KeyChanges, "Revoked thing stopped")
```
adamcfraser: I feel like this debug message could be workshopped a bit?

db/changes.go
```diff
@@ -594,8 +760,14 @@ func (db *Database) SimpleMultiChangesFeed(chans base.Set, options ChangesOption
...
 			}
 
-			// TODO: Use revocations option here in CBG-1367
-			_ = options.Revocations
+			if options.Revocations {
+				channelsToRevoke := db.user.RevokedChannels(options.Since.SafeSequence())
+				for channel, triggeredBy := range channelsToRevoke {
+					feed := db.buildRevokedFeed(db.changeCache.getChannelCache().getSingleChannelCache(channel), options, triggeredBy)
+					feeds = append(feeds, feed)
+					names = append(names, channel)
```
adamcfraser: Is names actually still used?  If it is, should we be modifying the name to indicate that this is a revocation feed, in the same way we're doing for 'late' sequence feeds?

adamcfraser (changes requested):
Generally looks very good - a few questions on some of the details, and a few proposed tweaks to comments/etc.

auth/user.go
```diff
@@ -289,6 +290,109 @@ func (user *userImpl) RevokedChannels(since uint64) RevokedChannels {
...
+	}
+
+	// Iterate over current user channels and add to resultPairs
+	for channelName, chanInfo := range user.Channels() {
+		if channelName != chanName {
+			continue
+		}
+		resultPairs = append(resultPairs, GrantHistorySequencePair{
+			StartSeq: chanInfo.Sequence,
+			EndSeq:   latestSequence,
```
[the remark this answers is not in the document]
JRascagneres: Good point, I'll switch this to MaxUint

auth/user.go
```diff
@@ -289,6 +290,109 @@ func (user *userImpl) RevokedChannels(since uint64) RevokedChannels {
...
+				}
+			}
+		}
+	}
+
+	return resultPairs, nil
+}
+
+// Find latest revocation status. Used to determine whether we can resume a interrupted revocation feed or if we need
+// to start again from 0
```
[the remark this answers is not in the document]
JRascagneres: No longer relevant

db/changes.go
```diff
@@ -171,6 +172,160 @@ func (db *Database) AddDocInstanceToChangeEntry(entry *ChangeEntry, doc *Documen
...
+	paginationOptions := options
+	paginationOptions.Since.Seq = 0
+	paginationOptions.Since.LowSeq = 0
+
+	if options.Since.TriggeredBy > 0 {
+		// Need to check if there has been a change since the triggeredByVal. If there has we need to start again from
+		// 0. Otherwise we can continue where we left off
+		highSeq, err := db.user.FindMostRecentRevocationSeq(singleChannelCache.ChannelName())
+		if highSeq > triggeredBy && err == nil {
+			paginationOptions.Since.Seq = options.Since.TriggeredBy
```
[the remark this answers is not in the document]
JRascagneres: Yeah you're right.

db/changes.go
JRascagneres: Yeah that makes sense

db/changes.go
JRascagneres: Yeah thats makes more sense.

db/changes.go
JRascagneres: Woopsy 😂

db/changes.go
JRascagneres: Doesn't look like it's used, or has been used for as far back as I can see in release history. Will remove it. Only reason I used it is I was following what others were doing.

db/changes.go
JRascagneres: Added `with omit revID %t`

db/changes.go
JRascagneres: Yeah this was the intention, that it would only fetch roles once per user.

db/changes.go
```diff
@@ -171,6 +172,160 @@ func (db *Database) AddDocInstanceToChangeEntry(entry *ChangeEntry, doc *Documen
...
+	queryLimit := db.Options.CacheOptions.ChannelQueryLimit
+	requestLimit := options.Limit
+
+	paginationOptions := options
+	paginationOptions.Since.Seq = 0
+	paginationOptions.Since.LowSeq = 0
+
+	if options.Since.TriggeredBy > 0 {
+		// Need to check if there has been a change since the triggeredByVal. If there has we need to start again from
+		// 0. Otherwise we can continue where we left off
```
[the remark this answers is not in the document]
JRascagneres: Yeah you're right. I had a reason to do this but upon further evaluation it doesn't make sense. Have removed this additional check.

adamcfraser (approved):

PR #5001, opened 2021-05-17
CBG-1424 - Show network interface stats when configured API listen adresses use hostnames instead of IPs

bbrks:
- Lookup non-IP host to resolve addresses in `discoverInterfaceName`, prior to searching for a matching network interface.
  - Note: This resolves hostnames at each stat collection - which can be advantageous for dynamic addressing, and is not expected to be detremental given DNS caching on both the local resolver, and upstream DNS servers.
- Simplify `networkInterfaceStatsForHostnamePort`:
  - Remove intermediate `allInterfaces` bool in favour of single `perNic`
  - Remove special case for `localhost`->`127.0.0.1` (let the resolver deal with that, along with others like `localhost4`, `localhost6`, and proper hostnames like `admin.sgw.example.com`)

## Manual testing:

1. Add localhost alias to /etc/hosts for resolver to see.
```
$ cat /etc/hosts | grep localhost
# localhost is used to configure the loopback interface
127.0.0.1	localhost
::1             localhost
127.0.0.1 speciallocalhost
```
2. Bind admin interface to that alias
```
$ cat config.json
{
  "adminInterface": "speciallocalhost:4985",
  "unsupported": {
    "stats_log_freq_secs": 10
  },
...
```
3. Run SG and wait for first stat collection.
4. See expvar for working (>0) `admin_net_bytes_recv` stat and no warning in logs.
5. Repeat from step 2 with IP address instead of hostname for regression.

### Before:
Screenshot 2021-05-17 at 21 51 47

```
2021-05-17T21:50:58.825+01:00 [WRN] Error getting admin network interface resource stats: Unable to find interface for host: speciallocalhost -- rest.(*ServerContext).logNetworkInterfaceStats() at server_context.go:1122
```

### After:
Screenshot 2021-05-17 at 21 52 53

### DIFF
--- a/rest/stats_context.go
+++ b/rest/stats_context.go
@@ -245,54 +245,41 @@ func goroutineHighwaterMark(numGoroutines uint64) (maxGoroutinesSeen uint64) {
 
 }
 
-func networkInterfaceStatsForHostnamePort(hostPort string) (iocountersStats gopsutilnet.IOCountersStat, err error) {
+func networkInterfaceStatsForHostnamePort(hostPort string) (*gopsutilnet.IOCountersStat, error) {
 
 	host, _, err := net.SplitHostPort(hostPort)
 	if err != nil {
-		return iocountersStats, err
-	}
-
-	allInterfaces := false
-	if host == "" || host == "0.0.0.0" {
-		allInterfaces = true
-	}
-
-	// "localhost" -> "127.0.0.1", since this code only works with IP addresses
-	if host == "localhost" {
-		host = "127.0.0.1"
-	}
-
-	interfaceName := ""
-	if !allInterfaces {
-		interfaceName, err = discoverInterfaceName(host)
-		if err != nil {
-			return iocountersStats, err
-		}
+		return nil, err
 	}
 
 	// Only get interface stats on a "Per Nic (network interface card)" basis if we aren't
 	// listening on all interfaces, in which case we want the combined stats across all NICs.
-	perNic := !allInterfaces
+	perNic := true
+	if host == "" || host == "0.0.0.0" {
+		perNic = false
+	}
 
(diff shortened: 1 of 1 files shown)


rest/stats_context.go
```diff
@@ -320,19 +323,20 @@ func discoverInterfaceName(host string) (interfaceName string, err error) {
...
+
 		for _, ifaceCIDRAddr := range ifaceAddresses {
 
 			ipAddr, _, err := net.ParseCIDR(ifaceCIDRAddr.String())
 			if err != nil {
 				return "", err
 			}
 
-			if ipAddr.String() == host {
+			if _, ok := hosts[ipAddr.String()]; ok {
```
adamcfraser, line 334: In the scenario where a host has multiple addresses, is it possible for those addresses to assigned to different interfaces?

adamcfraser (commented):
Approach looks fine - question about a multiple address/multiple interface scenario.

rest/stats_context.go
bbrks: Yes it's possible, but the interface _actually_ being used for the REST API depends what was first in the resolved list of addresses at the time of `net.Listen()` [which does not support multiple addresses](https://github.com/golang/go/issues/9334).

I think that means that as long as `LookupHost` resolves in the same order as `net.Listen`, you'll get correct results - but if that does not happen, you'll see stats for another interface. I don't know how likely or unlikely this is.

There's a "correct" implementation that resolves an address _before_ starting up the HTTP servers, so we know exactly what to look for in this stat code, but I'm not sure it's worth that to cover this specific case.

adamcfraser (approved):

rest/stats_context.go
adamcfraser: Makes sense - thanks.

PR #5004, opened 2021-05-20
CBG-1426: Include removed in changes deleted field

JRascagneres:
Include removed in deleted bitfield on changes feed.
Modified existing `TestRevocationMessage` test to check that this field is set correctly.

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -310,13 +310,13 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 					if bh.db.DatabaseContext.Options.UnsupportedOptions.ForceBLIPV3 {
 						deletedFlags := changesDeletedFlag(0)
 						if change.Deleted {
-							deletedFlags += changesDeletedFlagDeleted
+							deletedFlags |= changesDeletedFlagDeleted
 						}
 						if change.Revoked {
-							deletedFlags += changesDeletedFlagRevoked
+							deletedFlags |= changesDeletedFlagRevoked
 						}
 						if len(change.Removed) > 0 {
-							deletedFlags += changesDeletedFlagRemoved
+							deletedFlags |= changesDeletedFlagRemoved
 						}
 
 						changeRow = []interface{}{change.Seq, change.ID, item["rev"], deletedFlags}
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -249,6 +249,15 @@ type sendChangesOptions struct {
 	ignoreNoConflicts bool
 }
 
+type changesDeletedFlag uint
+
+const (
+	// Bitfield flags used to build changes deleted property below
+	changesDeletedFlagDeleted changesDeletedFlag = 0b001
+	changesDeletedFlagRevoked changesDeletedFlag = 0b010
+	changesDeletedFlagRemoved changesDeletedFlag = 0b100
+)
+
 // Sends all changes since the given sequence
 func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions) (isComplete bool) {
 	defer func() {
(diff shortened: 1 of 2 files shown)


db/blip_handler.go
```diff
@@ -306,6 +306,9 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 						if change.Revoked {
 							deleted += 2
 						}
+						if len(change.Removed) > 0 {
+							deleted += 4
+						}
```
bbrks: I think I've mentioned this before, but I really dislike this way of setting flags as decimal values in a bitfield. It ultimately just looks like magic numbers, and glancing at this code, it's not obvious that `deleted` is just a set of flags that mean more than just "deleted".

It would be slightly better if:
- `deleted` was renamed to `deletedFlags`
- `1`, `2`, and `4` were named consts (i.e. `changesDeletedFlagRemoved`)

But the `+=` might still throw people off I think.
Adding a new type for the flag consts gives yet more hints that people shouldn't be including magic numbers, and should only be using the defined consts.

```go
type changesDeletedFlag uint

const (
	// could use iota here, but the replication protocol
	// should define them explicitly, so why not here too
	changesDeletedFlagDeleted changesDeletedFlag = 0b001
	changesDeletedFlagRevoked changesDeletedFlag = 0b010
	changesDeletedFlagRemoved changesDeletedFlag = 0b100
)
```

```go
	var deletedFlags changesDeletedFlag
	if change.Deleted {
		deletedFlags |= changesDeletedFlagDeleted
	}
	if change.Revoked {
		deletedFlags |= changesDeletedFlagRevoked
	}
	if change.Removed {
		deletedFlags |= changesDeletedFlagRemoved
	}
```

rest/blip_api_test.go
```diff
@@ -3093,62 +3093,72 @@ func TestRevocationMessage(t *testing.T) {
...
+		},
+	}
 
-				for _, changesMessages := range changesMessage {
-					if docName, ok := changesMessages[1].(string); !ok || docName != "doc" {
+	for _, testCase := range testCases {
+		t.Run(testCase.Name, func(t *testing.T) {
+			// Verify the deleted property in the changes message is "2" this indicated a revocation
+			for _, msg := range messages {
+				if msg.Properties["Profile"] == "changes" {
```
bbrks:
```suggestion
				if msg.Properties[db.BlipProfile] == db.MessageChanges {
```

bbrks (changes requested):
The more I glance at this, the more I feel we need to have actual named bitmask flag values. It wasn't as bad with just 2 cases.

db/blip_handler.go
```diff
@@ -249,6 +249,15 @@ type sendChangesOptions struct {
 	ignoreNoConflicts bool
 }
 
+type changesDeletedFlag uint
+
+const (
```
adamcfraser, line 254: I think it's more idiomatic go to use `1 << iota` for this scenario, isn't it?  Like we're doing for things like LogEntry.Flags?  So something like
```
const (
	// Bitfield flags used to build changes deleted property below
	changesDeletedFlagDeleted changesDeletedFlag = 1 << iota
	changesDeletedFlagRevoked 
	changesDeletedFlagRemoved 
)
```

db/blip_handler.go
```diff
@@ -299,16 +308,19 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 
 					var changeRow []interface{}
 					if bh.db.DatabaseContext.Options.UnsupportedOptions.ForceBLIPV3 {
-						deleted := 0
+						deletedFlags := changesDeletedFlag(0)
 						if change.Deleted {
-							deleted += 1
+							deletedFlags += changesDeletedFlagDeleted
```
adamcfraser: I feel like using the bitwise or operator to set flags (`|=`) is more defensive - it avoids a regression in future where the flag is already set, so `+=` doesn't work as expected.

adamcfraser (changes requested):

db/blip_handler.go
bbrks: I suggested in an above comment explicit values, given the source of these values is the replication protocol, rather than the Go code, to avoid any future const additions from inadvertently changing the values.

adamcfraser: I see what you mean - I think that's reasonable.  In that case, what do you think about 0x01, 0x02, 0x04 instead of the 0b representations?   I think gocb follows that approach for flags, and I haven't actually seen the 0b format used previously.

JRascagneres: I personally prefer the binary representation as it makes it obvious it's a bit flag that has each different bit set. 
I can switch it though if you think it'd be better.

db/blip_handler.go
JRascagneres: Agreed. Forgot to do this earlier.

adamcfraser (approved):

PR #5005, opened 2021-05-21
Add clistruct package to map JSON struct tagged fields to command line flags

bbrks:
Mostly the same as the closed #4995 - but uses `json` struct tags instead of explicit `cli` flags.
Also handles a lot more types now (pointers, arbitrary structs, `[]string`) typically seen in config structs.

---

`json:"foo"` struct tag creates a flag called -foo.

Optionally specify help using `json:"foo" help:"foo is something"` to include a textual description of the flag.

Fields must be exported, in the same way that JSON requires. MustRegisterJSONFlags panics for structs that try to map invalid fiels or unknown types.

Supports nested structs, which dot-separates flags:
```go
    struct{
        A struct {
            B bool `json:"b"`
        } `json:"a"`
    }
```
results in a flag called `-a.b`

Embedded structs are also handled as expected:
```go
    struct {
        A bool `json:"a"`
        structB
    }
    type structB struct {
        B bool `cli:"b"`
    }
```
results in flags `-a` and `-b`

### DIFF
--- a/base/clistruct/clistruct.go
+++ b/base/clistruct/clistruct.go
@@ -1,4 +1,4 @@
-// Package clistruct provides the ability to assign command line flags to fields in structs that declare `cli` and `help` struct tags.
+// Package clistruct provides the ability to automatically assign command line flags to fields in structs that declare `json` and optionally `help` struct tags.
 package clistruct
 
 import (
--- a/base/clistruct/clistruct.go
+++ b/base/clistruct/clistruct.go
@@ -1,0 +1,224 @@
+// Package clistruct provides the ability to assign command line flags to fields in structs that declare `cli` and `help` struct tags.
+package clistruct
+
+import (
+	"encoding/json"
+	"errors"
+	"flag"
+	"fmt"
+	"reflect"
+	"strings"
+	"time"
+)
+
+const (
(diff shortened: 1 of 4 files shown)


base/clistruct/clistruct.go
```diff
@@ -0,0 +1,224 @@
...
+		if v == nil {
+			tmp := jsonNumberFlagValue("0")
+			i = &tmp
+			fieldValue.Set(reflect.ValueOf((*json.Number)(&tmp)))
+		} else {
+			i = (*jsonNumberFlagValue)(v)
+		}
+	}
+
+	switch v := i.(type) {
```
adamcfraser: I don't completely follow why we need a second type switch here, as opposed to just adding the fs.Var to the cases above.

adamcfraser (changes requested):
Looks good - one minor question on the duplicate type switch.

base/clistruct/clistruct.go
bbrks: No reason over having multiple calls to `fs.Var` - changed as it's easier to reason about when explicit.

adamcfraser (approved):

PR #5007, opened 2021-05-25
CBG-1435: Support multiple replication protocols

JRascagneres:
Bring over go-blip changes to allow for multiple protocols.

Included removing the temporary ForceBLIPV3 config option as this functionality is now handled by the set active protocol. 

 - [x] https://github.com/couchbase/go-blip/pull/48
 
 Note: This will require QE testing to ensure that this continues to work with both V2 CBL clients and Pre-3.0 ISGR implementations.

### DIFF
--- a/base/constants.go
+++ b/base/constants.go
@@ -130,6 +130,13 @@ const (
 
 	// Replication filter constants
 	ByChannelFilter = "sync_gateway/bychannel"
+
+	// Blip Context SubProtocols
+
+	// blipCBMobileReplication is the AppProtocolId part of the BLIP websocket subprotocol.  Must match identically with the peer (typically CBLite / LiteCore).
+	// At some point this will need to be able to support an array of protocols.  See go-blip/issues/27.
+	BlipCBMobileReplicationV2 = "CBMobile_2"
+	BlipCBMobileReplicationV3 = "CBMobile_3"
 )
 
 const (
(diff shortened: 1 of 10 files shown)


base/constants.go
```diff
@@ -130,6 +130,13 @@ const (
 
 	// Replication filter constants
 	ByChannelFilter = "sync_gateway/bychannel"
+
+	// Blip Context SubProtocols
+
+	// blipCBMobileReplication is the AppProtocolId part of the BLIP websocket subprotocol.  Must match identically with the peer (typically CBLite / LiteCore).
+	// At some point this will need to be able to support an array of protocols.  See go-blip/issues/27.
```
bbrks: "At some point" being now? 😄

base/constants.go
```diff
@@ -130,6 +130,13 @@ const (
...
+	BlipCBMobileReplicationV2 = "CBMobile_2"
```
bbrks: I think these can stay inside `db/blip.go` now they're only used in one place.

bbrks (changes requested):
Minor comment on the consts, which I think moved due to a previous change that isn't here anymore.

base/constants.go
JRascagneres: Yas

base/constants.go
JRascagneres: Correct

bbrks (dismissed):
LGTM pending go-blip merge/manifest bump

adamcfraser (dismissed):

adamcfraser (approved):

PR #5012, opened 2021-06-01
CBD-4043: Mobile 3.0 BSL

udkyo:
This change adds license header snippets to all source files, and adds associated license files to /licenses


bbrks:
Is there a script which generates these file headers?

I'm wondering how to keep new files up to date with the headers if they are now required by legal.

udkyo:
Yes these are added by script, the idea is to propose the bulk of these changes manually initially, with future incremental updates being proposed by an automated process.

bbrks:
Ok no problem. I was thinking we'd need to run the script locally on a precommit hook or something before pushing new files.

But if that process is something being organised by the build team that's fine 👍

bbrks (approved):

PR #5015, opened 2021-06-02
Remove AdminUI

bbrks:


bbrks:
Closing - will still need these config options in this struct for being able to read pre-3.0 configs correctly, even if they aren't wired up to anything.

PR #5018, opened 2021-06-02
CBG-1386: Handle pruning of history

JRascagneres:
Adds pruning to grant history. Rules as follows:

 - If the history entries for a role / channel are not updated in 30 days then the entire history for that role / channel will be removed.
 - Grant Pairs are pruned to maintain a maximum of 10 entries and a minimum of 1 entry. Numbers in between are calculated in order to maintain grant history no larger than 1MB. 
 Note 1MB can be exceeded if an extreme number of channels / roles have history as each requires at least 1 entry.
 
The calculation for the second rule has a unit test.

Created AuthenticatorOptions to pass through options to the Authenticator. Currently has two attributes but allows for easy additions in future.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -229,7 +229,7 @@ func (auth *Authenticator) calculateHistory(princName string, invalSeq uint64, i
 	}
 
 	if prunedHistory := currentHistory.PruneHistory(auth.ClientPartitionWindow); len(prunedHistory) > 0 {
-		base.Debugf(base.KeyCRUD, "rebuildChannels: Pruned principal history on %s for %s", princName, strings.Join(prunedHistory, ","))
+		base.Debugf(base.KeyCRUD, "rebuildChannels: Pruned principal history on %s for %s", base.UD(princName), base.UD(strings.Join(prunedHistory, ",")))
 	}
 
 	// Ensure no entries are larger than the allowed threshold
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -47,6 +47,15 @@ type userByEmailInfo struct {
 
 const PrincipalUpdateMaxCasRetries = 20 // Maximum number of attempted retries on cas failure updating principal
 
+// Constants used in CalculateMaxHistoryEntriesPerGrant
+const (
+	maximumHistoryBytes          = 1024 * 1024 // 1MB
+	averageHistoryKeyBytes       = 250         // This is an estimate of key size in bytes, this includes channel name, the unix timestamp, "entries" key
+	averageHistoryEntryPairBytes = 14          // Assume each sequence is 7 digits
+	minHistoryEntriesPerGrant    = 1           // Floor of history entries count to ensure there is at least 1 entry
+	maxHistoryEntriesPerGrant    = 10          // Ceiling of history entries count to ensure there is no more than 10 entries
+)
+
 // Creates a new Authenticator that stores user info in the given Bucket.
 func NewAuthenticator(bucket base.Bucket, channelComputer ChannelComputer, options AuthenticatorOptions) *Authenticator {
 	return &Authenticator{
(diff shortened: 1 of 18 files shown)


auth/auth_test.go
```diff
@@ -2648,3 +2648,33 @@ func TestInvalidateChannels(t *testing.T) {
...
+		{
+			Name:                     "Tiny history length",
+			HistoryLengthInput:       1,
+			MaxEntriesExpectedOutput: 10,
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.Name, func(t *testing.T) {
+			assert.Equal(t, CalculateMaxHistoryEntriesPerGrant(testCase.HistoryLengthInput), testCase.MaxEntriesExpectedOutput)
```
bbrks: expected <--> actual

```suggestion
			assert.Equal(t, testCase.MaxEntriesExpectedOutput, CalculateMaxHistoryEntriesPerGrant(testCase.HistoryLengthInput))
```

auth/auth_test.go
```diff
@@ -2648,3 +2648,33 @@ func TestInvalidateChannels(t *testing.T) {
...
+		{
+			Name:                     "Extreme history length",
+			HistoryLengthInput:       1000000,
+			MaxEntriesExpectedOutput: 1,
+		},
+		{
+			Name:                     "Large history length",
+			HistoryLengthInput:       1000,
+			MaxEntriesExpectedOutput: 10,
+		},
```
bbrks: Include these so:
- We can see where the distribution is (these numbers are ballpark, they have a fair amount of leeway in them)
- The test fails if we change things to affect the distribution too significantly

```suggestion
		{
			Name:                     "Upper bound",
			HistoryLengthInput:       3600,
			MaxEntriesExpectedOutput: 1,
		},
		{
			Name:                     "Large history length",
			HistoryLengthInput:       3330,
			MaxEntriesExpectedOutput: 3,
		},
		{
			Name:                     "Median history length",
			HistoryLengthInput:       3075,
			MaxEntriesExpectedOutput: 5,
		},
		{
			Name:                     "Small history length",
			HistoryLengthInput:       2800,
			MaxEntriesExpectedOutput: 7,
		},
		{
			Name:                     "Lower bound",
			HistoryLengthInput:       2550,
			MaxEntriesExpectedOutput: 10,
		},
```

bbrks (changes requested):

auth/role.go
```diff
@@ -35,6 +36,19 @@ type roleImpl struct {
 
 type TimedSetHistory map[string]GrantHistory
 
+func (timedSet TimedSetHistory) PruneHistory(partitionWindow time.Duration) []string {
+	prunedChannelHistory := make([]string, 0)
+	for chanName, grantHistory := range timedSet {
+		grantTime := time.Unix(0, grantHistory.UpdatedAt)
```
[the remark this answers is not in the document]
JRascagneres: Switched to Unix

base/constants.go
```diff
@@ -167,6 +167,8 @@ var (
 	DefaultWarnThresholdChannelsPerDoc  = uint32(50)
 	DefaultWarnThresholdChannelsPerUser = uint32(50000)
 	DefaultWarnThresholdGrantsPerDoc    = uint32(50)
+	DefaultTimedSetHistoryPruneCap      = 10
```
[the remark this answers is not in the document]
JRascagneres: Yeah woops, left this. 
I'll add a const for the max and min.

auth/auth.go
```diff
@@ -222,20 +219,51 @@ func (auth *Authenticator) calculateHistory(invalSeq uint64, invalGrants ch.Time
...
+		currentHistoryForGrant.UpdatedAt = time.Now().Unix()
 		currentHistoryForGrant.Entries = append(currentHistoryForGrant.Entries, GrantHistorySequencePair{
 			StartSeq: previousInfo.Sequence,
 			EndSeq:   invalSeq,
 		})
 		currentHistory[previousName] = currentHistoryForGrant
 	}
 
+	if prunedHistory := currentHistory.PruneHistory(auth.ClientPartitionWindow); len(prunedHistory) > 0 {
+		base.Debugf(base.KeyCRUD, "rebuildChannels: Pruned principal history on %s for %s", base.UD(princName), base.UD(strings.Join(prunedHistory, ",")))
```
bbrks: Would be better to redact each element individually, so each channel name gets a hash, which is already supported for string slices.

```suggestion
		base.Debugf(base.KeyCRUD, "rebuildChannels: Pruned principal history on %s for %s", base.UD(princName), base.UD(prunedHistory))
```

bbrks (changes requested):
Looks good apart from the last change for log redaction

bbrks (dismissed):

auth/auth.go
```diff
@@ -222,20 +218,51 @@ func (auth *Authenticator) calculateHistory(invalSeq uint64, invalGrants ch.Time
...
+			grantHistory.Entries[1].StartSeq = grantHistory.Entries[0].StartSeq
+			currentHistory[grantName] = grantHistory
+		}
+	}
+
 	return currentHistory
 }
 
+func CalculateMaxHistoryEntriesPerGrant(channelCount int) int {
+	const maximumHistoryBytes = 1024 * 1024 // 1MB
```
adamcfraser: All three of these constants are the same units, but that's not evident from naming.  What do you think about maximumHistoryBytes, averageHistoryKeyBytes, and averageHistoryEntryBytes?

adamcfraser (dismissed):
Looks good, just some minor nits about naming.

adamcfraser (approved):

PR #5021, opened 2021-06-04
CBG-1477: Only send revocations if user has no access

JRascagneres:
Adds calculation to **only** send revocations if the user has completely lost access to the document. Means that we do an additional lookup for revocations to check whether the user has access through an alternative channel. 
This means the `{"_removed": true}` body check is no longer necessary and the document can be immediately purged when a revocation is received. 

Removals also follow same

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -302,21 +302,9 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 	_, forceClose := generateBlipSyncChanges(changesDb, channelSet, options, opts.docIDs, func(changes []*ChangeEntry) error {
 		base.DebugfCtx(bh.loggingCtx, base.KeySync, "    Sending %d changes", len(changes))
 		for _, change := range changes {
-
 			if !strings.HasPrefix(change.ID, "_") {
 				for _, item := range change.Changes {
 					changeRow := bh.buildChangesRow(change, item["rev"])
-					pendingChanges = append(pendingChanges, changeRow)
-					if err := sendPendingChangesAt(opts.batchSize); err != nil {
-						return err
-					}
-				}
-
-				// In the event we were unable to determine what the rev ID was at the time of revocation we have to
-				// send a revocation changes message but without a revID present. This will result in change.Changes
-				// being empty so we need to have this special case.
-				if len(change.Changes) == 0 && change.Revoked == true {
-					changeRow := bh.buildChangesRow(change, "")
 					pendingChanges = append(pendingChanges, changeRow)
 					if err := sendPendingChangesAt(opts.batchSize); err != nil {
 						return err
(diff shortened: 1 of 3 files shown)


db/changes.go
```diff
@@ -282,6 +292,20 @@ func (db *Database) buildRevokedFeed(singleChannelCache SingleChannelCache, opti
 	return feed
 }
 
+func UserStillMaintainsAccessToDoc(db *Database, docID, revID string) (bool, error) {
```
adamcfraser: UserHasDocAccess maybe?

db/blip_handler.go
```diff
@@ -302,21 +302,21 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
...
 				for _, item := range change.Changes {
 					changeRow := bh.buildChangesRow(change, item["rev"])
-					pendingChanges = append(pendingChanges, changeRow)
-					if err := sendPendingChangesAt(opts.batchSize); err != nil {
-						return err
+
+					if len(change.Removed) > 0 && !change.Deleted && bh.blipContext.ActiveProtocol() == BlipCBMobileReplicationV3 {
+						userMaintainsAccessToDoc, err := UserStillMaintainsAccessToDoc(bh.db, change.ID, item["rev"])
+						if err != nil {
+							return err
```
adamcfraser: On error, should log at info and continue.

db/blip_handler.go
```diff
@@ -514,7 +514,17 @@ func (bh *blipHandler) handleChanges(rq *blip.Message) error {
 
 		}
 
-		if missing == nil && deletedFlags&(changesDeletedFlagRevoked|changesDeletedFlagRemoved) == 0 {
+		if bh.blipContext.ActiveProtocol() == BlipCBMobileReplicationV3 && deletedFlags&(changesDeletedFlagRevoked|changesDeletedFlagRemoved) != 0 && deletedFlags&changesDeletedFlagDeleted == 0 {
```
adamcfraser: Need to do this only for purge on removal?

adamcfraser: Do we actually need to exempt tombstones from purge on revoked/removal?

db/changes.go
```diff
@@ -282,6 +292,20 @@ func (db *Database) buildRevokedFeed(singleChannelCache SingleChannelCache, opti
...
+	rev, err := db.revisionCache.Get(docID, revID, false, false)
+	if err != nil {
```
adamcfraser, line 297: On a KeyNotFound error, I think we can return "false, nil", as they don't have access to a rev that doesn't exist. Otherwise return error.

adamcfraser (changes requested):

db/blip_handler.go
JRascagneres: Changed to only do on purge on removal and removed deleted checks.

adamcfraser (approved):

PR #5023, opened 2021-06-07
CBG-1412 - JSON strings in some responses not being correctly escaped

IsaacLambat:
Implemented fix for PUT (local, replicator2, with attachment, and standard) and DELETE. All fully tested.

### DIFF
--- a/rest/admin_api_test.go
+++ b/rest/admin_api_test.go
@@ -39,44 +39,44 @@ func TestPutDocSpecialChar(t *testing.T) {
 	rt := NewRestTester(t, nil)
 	defer rt.Close()
 	testCases := []struct {
-		name string
-		pathDocID string
-		method string
-		body string
+		name         string
+		pathDocID    string
+		method       string
+		body         string
 		expectedResp int
-		eeOnly bool
+		eeOnly       bool
 	}{
 		{
-			name: "Double quote PUT",
-			pathDocID: `doc"55"`,
-			method: "PUT",
-			body: "{}",
+			name:         "Double quote PUT",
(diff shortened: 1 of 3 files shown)


rest/admin_api_test.go
```diff
@@ -34,6 +34,81 @@ import (
...
+			eeOnly:       false,
+		},
+		{
+			name:         "Double quote PUT with attachment",
+			pathDocID:    `doc"59"/attachMe`,
+			method:       "PUT",
+			body:         "{}",
+			expectedResp: http.StatusCreated, // Admin Docs expected response http.StatusOK
+			eeOnly:       false,
+		},
```
bbrks, line 80: I see you had a test case for POST but it was removed.

Looks like handlePostDoc still has an unescaped docID field in the response body.
Is there a plan to include that in this PR?

bbrks (changes requested):
Looks good, just a question about the remaining `POST` case.

rest/admin_api_test.go
IsaacLambat: I didn't escape the DocID for handlePostDoc because when you try to use unescaped JSON, it returns HTTP status 400 (http.StatusBadRequest).

The test checked if the 400 error was returned for unescaped JSON to the POST doc endpoint and I didn't think that would be relevant to this ticket.

bbrks: Looks like handlePostDoc is already dealing with ensuring the doc ID is well-formed due to it coming out of the request body, I didn't notice that previously.

Looks good as is!

bbrks (approved):
LGTM 👍

PR #5030, opened 2021-06-08
CBG-1488: Fix and test interrupted revocation logic

JRascagneres:
Fix logic with interrupted replications. Largely involves utilizing the triggeredBy value to calculate revocations rather than only the Seq value.

### DIFF
--- a/db/changes.go
+++ b/db/changes.go
@@ -775,7 +775,7 @@ func (db *Database) SimpleMultiChangesFeed(chans base.Set, options ChangesOption
 
 			if options.Revocations && db.user != nil {
 				revocationCheckSeq := options.Since.SafeSequence()
-				if options.Since.TriggeredBy > 0 {
+				if options.Since.TriggeredBy > 0 && revocationCheckSeq > options.Since.TriggeredBy-1 {
 					revocationCheckSeq = options.Since.TriggeredBy - 1
 				}
 				channelsToRevoke := db.user.RevokedChannels(revocationCheckSeq)
--- a/db/changes.go
+++ b/db/changes.go
@@ -171,7 +171,7 @@ func (db *Database) AddDocInstanceToChangeEntry(entry *ChangeEntry, doc *Documen
 	}
 }
 
-func (db *Database) buildRevokedFeed(singleChannelCache SingleChannelCache, options ChangesOptions, revokedSeq, revocationCheckSeq uint64, to string) <-chan *ChangeEntry {
+func (db *Database) buildRevokedFeed(singleChannelCache SingleChannelCache, options ChangesOptions, revokedSeq, revocationSinceSeq uint64, to string) <-chan *ChangeEntry {
 	feed := make(chan *ChangeEntry, 1)
 	sinceVal := options.Since.Seq
 
@@ -232,7 +232,7 @@ func (db *Database) buildRevokedFeed(singleChannelCache SingleChannelCache, opti
 				// Otherwise: we need to determine whether a previous revision of the document was in the channel prior
 				// to the since value, and only send a revocation if that was the case
 				if logEntry.Sequence > sinceVal {
-					requiresRevocation, err := db.wasDocInChannelAtSeq(logEntry.DocID, singleChannelCache.ChannelName(), revocationCheckSeq)
+					requiresRevocation, err := db.wasDocInChannelAtSeq(logEntry.DocID, singleChannelCache.ChannelName(), revocationSinceSeq)
 					if err != nil {
 						change := ChangeEntry{
 							Err: base.ErrChannelFeed,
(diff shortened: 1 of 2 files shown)


db/changes.go
```diff
@@ -774,9 +774,13 @@ func (db *Database) SimpleMultiChangesFeed(chans base.Set, options ChangesOption
 			}
 
 			if options.Revocations && db.user != nil {
-				channelsToRevoke := db.user.RevokedChannels(options.Since.SafeSequence())
-				for channel, triggeredBy := range channelsToRevoke {
-					feed := db.buildRevokedFeed(db.changeCache.getChannelCache().getSingleChannelCache(channel), options, triggeredBy, to)
+				revocationCheckSeq := options.Since.Seq
```
adamcfraser: Keep SafeSequence(), or see whether RevokedChannels should do that work.

Might be useful to throw a few cases of sequences with lowSeq in the tests as since values, to make sure it's not missed elsewhere.

adamcfraser (changes requested):

db/changes.go
JRascagneres: Switched back to SafeSeq and added a test case in the existing tests.

db/changes.go
```diff
@@ -774,9 +774,13 @@ func (db *Database) SimpleMultiChangesFeed(chans base.Set, options ChangesOption
 			}
 
 			if options.Revocations && db.user != nil {
-				channelsToRevoke := db.user.RevokedChannels(options.Since.SafeSequence())
-				for channel, triggeredBy := range channelsToRevoke {
-					feed := db.buildRevokedFeed(db.changeCache.getChannelCache().getSingleChannelCache(channel), options, triggeredBy, to)
+				revocationCheckSeq := options.Since.SafeSequence()
```
adamcfraser: Thinking about it more, I don't think we want to use SafeSequence() as revocationCheckSeq.  Consider the case:
- sequences 100, 110 and 120 are in channel foo
- user loses access to channel foo at sequence 115
- sequence 105 is skipped
A user that replicated when latestSequence is 112 would end up with a lastSeq value like `100::110` (lowSeq is 100, seq is 110).
If that user later issues a changes request with since=100::110 and detects the revocation, we'd want to revoke everything up to 110, not 100.

db/changes.go
```diff
@@ -774,9 +774,13 @@ func (db *Database) SimpleMultiChangesFeed(chans base.Set, options ChangesOption
...
+				if options.Since.TriggeredBy > 0 {
+					revocationCheckSeq = options.Since.TriggeredBy - 1
+				}
+				channelsToRevoke := db.user.RevokedChannels(revocationCheckSeq)
```
adamcfraser: I'm not sure why the call to RevokedChannels should potentially use TriggeredBy - 1.  I thought it should just be using TriggeredBy, and the -1 is only needed below.

adamcfraser (changes requested):

db/changes.go
JRascagneres: `TestRevocationResumeSameRoleAndLowSeqCheck` shows the reason this is required:
- User has role1
- Role1 has ch1, ch2
- doc1 in ch1 at seq 10
- doc2 in ch2 at seq 11
- client does a pull. Last seq 11, has both docs.
- ch1 revoked from role at seq 20
- ch2 revoked from role at seq 30
At this point the invalSeq on the role is at 20 as we keep the oldest.
- Mutate doc1 seq 40
- Mutate doc2 seq 50

We start a revocation with say since 11. This will find that ch1 and ch2 are revoked and start sending, however, after sending the first the replication is interrupted and we have a since of 20:40.

If we just use '20' as triggered by it won't flag that ch1 and ch2 are revoked because `RevokedChannels` will say 'well at 20 you didn't have any channels so we're good'. So we need -1 for that.

db/changes.go
```diff
@@ -171,7 +171,7 @@ func (db *Database) AddDocInstanceToChangeEntry(entry *ChangeEntry, doc *Documen
 	}
 }
 
-func (db *Database) buildRevokedFeed(singleChannelCache SingleChannelCache, options ChangesOptions, triggeredBy uint64, to string) <-chan *ChangeEntry {
+func (db *Database) buildRevokedFeed(singleChannelCache SingleChannelCache, options ChangesOptions, revokedSeq, revocationCheckSeq uint64, to string) <-chan *ChangeEntry {
```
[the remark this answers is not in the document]
JRascagneres: Yeah thats fair, will rename

db/changes.go
JRascagneres: I think that'd still be the case here no? So 100 the lowSeq would be used to say 'what channels did I have at 100' to calculate which channels to revoke. We still revoke everything up to 110. We don't stop revocations when we reach the since.

adamcfraser: I thought we only sent revocations post-since if the document was in the channel prior to the since, is that not the case?  And in this case, the doc with seq 110 wasn't necessarily in the channel at 100?

db/changes.go
adamcfraser: That makes sense, and explains some of the difficulty in naming revocationCheckSeq.  The name is a bit more appropriate when passed to user.RevokedChannels, but then less clear (at least to me) how it's being used when passed to buildRevokedFeed.

JRascagneres: Yeah the passed in value is then used for wasDocInChannelAtSeq for a similar reason

db/changes.go
JRascagneres: We do but that since check uses the regular seq rather than this `revocationCheckSeq` value.

You may be right though, we might not want SafeSeq here.

db/changes.go
```diff
@@ -774,9 +774,13 @@ func (db *Database) SimpleMultiChangesFeed(chans base.Set, options ChangesOption
 			}
 
 			if options.Revocations && db.user != nil {
-				channelsToRevoke := db.user.RevokedChannels(options.Since.SafeSequence())
-				for channel, triggeredBy := range channelsToRevoke {
-					feed := db.buildRevokedFeed(db.changeCache.getChannelCache().getSingleChannelCache(channel), options, triggeredBy, to)
+				revocationCheckSeq := options.Since.SafeSequence()
+				if options.Since.TriggeredBy > 0 {
+					revocationCheckSeq = options.Since.TriggeredBy - 1
```
adamcfraser: As discussed, revocationCheckSeq should be the lower of lowSeq and (triggeredBy - 1).

db/changes.go
adamcfraser: The post-since check is being done in buildRevokedFeed based on seq, so isn't affected by this - this should be fine as-is.

db/changes.go
```diff
@@ -775,7 +775,7 @@ func (db *Database) SimpleMultiChangesFeed(chans base.Set, options ChangesOption
 
 			if options.Revocations && db.user != nil {
 				revocationCheckSeq := options.Since.SafeSequence()
-				if options.Since.TriggeredBy > 0 {
+				if options.Since.TriggeredBy > 0 && revocationCheckSeq > options.Since.TriggeredBy-1 {
```
adamcfraser: I think this condition merits a comment to provide some context on what the second condition is doing.

adamcfraser (changes requested):
Added one more comment on recent change, in addition to the pending comment on naming.

JRascagneres:
Added some comments explaining use of the revocationSinceSeq value.

adamcfraser (approved):

PR #5031, opened 2021-06-08
CBG-1326 - Warning logged when channels dropped from list when requesting changes

IsaacLambat:
A warning is issued with details about the channels requested and the filtered channels when requesting changes from a list of channels.

### DIFF
--- a/auth/user.go
+++ b/auth/user.go
@@ -520,7 +520,7 @@ func (user *userImpl) FilterToAvailableChannels(channels base.Set) (filtered ch.
 			removed = append(removed, channel)
 		}
 	}
-	return
+	return filtered, removed
 }
 
 func (user *userImpl) GetAddedChannels(channels ch.TimedSet) base.Set {
--- a/auth/user.go
+++ b/auth/user.go
@@ -509,16 +509,15 @@ func (user *userImpl) ExpandWildCardChannel(channels base.Set) base.Set {
 	return channels
 }
 
-func (user *userImpl) FilterToAvailableChannels(channels base.Set) (filtered ch.TimedSet, removed base.Set) {
+func (user *userImpl) FilterToAvailableChannels(channels base.Set) (filtered ch.TimedSet, removed []string) {
 	filtered = ch.TimedSet{}
-	removed = base.Set{}
 	for channel := range channels {
 		if channel == ch.AllChannelWildcard {
 			return user.InheritedChannels().Copy(), nil
 		}
 		added := filtered.AddChannel(channel, user.CanSeeChannelSince(channel))
 		if !added {
-			removed.Add(channel)
+			removed = append(removed, channel)
 		}
 	}
 	return
--- a/auth/user.go
+++ b/auth/user.go
@@ -509,15 +509,19 @@ func (user *userImpl) ExpandWildCardChannel(channels base.Set) base.Set {
 	return channels
 }
 
-func (user *userImpl) FilterToAvailableChannels(channels base.Set) ch.TimedSet {
-	output := ch.TimedSet{}
+func (user *userImpl) FilterToAvailableChannels(channels base.Set) (filtered ch.TimedSet, removed base.Set) {
+	filtered = ch.TimedSet{}
+	removed = base.Set{}
 	for channel := range channels {
 		if channel == ch.AllChannelWildcard {
-			return user.InheritedChannels().Copy()
+			return user.InheritedChannels().Copy(), nil
 		}
-		output.AddChannel(channel, user.CanSeeChannelSince(channel))
+		added := filtered.AddChannel(channel, user.CanSeeChannelSince(channel))
+		if !added {
+			removed.Add(channel)
+		}
 	}
-	return output
+	return
 }
 
 func (user *userImpl) GetAddedChannels(channels ch.TimedSet) base.Set {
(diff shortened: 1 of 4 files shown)


db/changes.go
```diff
@@ -631,6 +631,9 @@ func (db *Database) SimpleMultiChangesFeed(chans base.Set, options ChangesOption
 		var channelsSince channels.TimedSet
 		if db.user != nil {
 			channelsSince = db.user.FilterToAvailableChannels(chans)
+			if !chans.Contains("*") && len(channelsSince.AsSet()) < len(chans) {
+				base.WarnfCtx(db.Ctx, "Channels restricted from %s to %s", chans.String(), channelsSince.AsSet().String())
```
JRascagneres: I think perhaps the channel names should be redacted in the logging (at least we've done this in other places.
```suggestion
				base.WarnfCtx(db.Ctx, "Channels restricted from %s to %s", base.UD(chans.String()), base.UD(channelsSince.AsSet().String()))
```

JRascagneres (changes requested):

db/changes.go
```diff
@@ -631,6 +631,9 @@ func (db *Database) SimpleMultiChangesFeed(chans base.Set, options ChangesOption
 		var channelsSince channels.TimedSet
 		if db.user != nil {
 			channelsSince = db.user.FilterToAvailableChannels(chans)
+			if !chans.Contains("*") && len(channelsSince.AsSet()) < len(chans) {
```
JRascagneres: Is the AsSet() necessary on the len check?

db/changes_test.go
```diff
@@ -22,6 +22,87 @@ import (
...
+		_, _, err = db.Put("doc3", Body{"channels": []string{"ch3"}})
+		require.NoError(t, err)
+
+		db.WaitForPendingChanges(context.Background())
+
+		db.user, _ = auth.GetUser("test")
+
+		warnCount := base.SyncGatewayStats.GlobalStats.ResourceUtilizationStats().WarnCount.Value()
+
+		ch, err := db.GetChanges(base.SetOf("ch1", "ch2", "ch3"), ChangesOptions{Since: lastSeq})
```
JRascagneres: Any reason we're doing Since sequences like this?
I think it'd be simpler if the setup was done in each of the test runs (perhaps in a function if they're the same) and then we can just do since 0. Avoids doing the initial GetChanges to get lastSeq and all that. Would save you a few lines.

JRascagneres (dismissed):

db/changes_test.go
```diff
@@ -22,6 +22,87 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// Unit test for CBG-1326
+func TestLogIfChannelsDroppedFromList(t *testing.T) {
```
JRascagneres: Bit of a nitpick but wondering if this could use testCases? Each test looks the same except the create docs / channels. You could have a parameter that takes "channels list" or something and just create docs based off of that. Reduces a lot of the repetition.
If you think that makes it harder feel free to ignore this one!

db/changes.go
JFlath: @JRascagneres These days, that'll tag each element individually, right? Also, am I right in saying that the `AsSet()` will make the elements identical across both slices by stripping the `:123` from `foo:123` as normally appears in channelsSince type contexts?

JFlath (dismissed):

adamcfraser (changes requested):

db/changes.go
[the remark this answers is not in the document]
IsaacLambat: Alright, I will make FilterToAvailableChannels return the removed channels which shouldn't add much computing cost at all. Did originally think of doing it this way but didn't want to break anything.

auth/user.go
```diff
@@ -509,15 +509,19 @@ func (user *userImpl) ExpandWildCardChannel(channels base.Set) base.Set {
 	return channels
 }
 
-func (user *userImpl) FilterToAvailableChannels(channels base.Set) ch.TimedSet {
-	output := ch.TimedSet{}
+func (user *userImpl) FilterToAvailableChannels(channels base.Set) (filtered ch.TimedSet, removed base.Set) {
```
adamcfraser: Can just return removed as []string to avoid the later conversion at logging time.

db/changes_test.go
```diff
@@ -22,6 +22,65 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// Unit test for CBG-1326
+// This will need to be manually verified by checking the logs in console
+func TestLogIfChannelsDroppedFromList(t *testing.T) {
```
adamcfraser: As discussed, might be better to just target FilterToAvailableChannels directly with the test, as that's the main new code.

adamcfraser (changes requested):

auth/user.go
```diff
@@ -509,15 +509,18 @@ func (user *userImpl) ExpandWildCardChannel(channels base.Set) base.Set {
...
+			return user.InheritedChannels().Copy(), nil
+		}
+		added := filtered.AddChannel(channel, user.CanSeeChannelSince(channel))
+		if !added {
+			removed = append(removed, channel)
 		}
-		output.AddChannel(channel, user.CanSeeChannelSince(channel))
 	}
-	return output
+	return
```
adamcfraser:
```suggestion
	return filtered, removed
```

adamcfraser (dismissed):

bbrks (approved):

PR #5035, opened 2021-06-09
Admin API Auth: Initial version

JRascagneres:
Initial implementation of Admin API auth. Currently missing the ability to grab management endpoints, however, hardcoded with my test server and seems to work.

Tests currently are again, reliant on having the users on the test server however have prepared functions to make and create users for testing.

Can have a discussion / initial review based on this.

### DIFF
--- a/rest/admin_api_test.go
+++ b/rest/admin_api_test.go
@@ -17,13 +17,12 @@ import (
 	"log"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"strings"
 	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
-
-	"github.com/stretchr/testify/require"
 
 	"github.com/couchbase/sync_gateway/auth"
 	"github.com/couchbase/sync_gateway/base"
@@ -31,6 +30,7 @@ import (
 	"github.com/couchbase/sync_gateway/db"
 	goassert "github.com/couchbaselabs/go.assert"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 // Reproduces CBG-1412 - JSON strings in some responses not being correctly escaped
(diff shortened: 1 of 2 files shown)


rest/config.go
```diff
@@ -1284,6 +1285,123 @@ func startServer(config *ServerConfig, sc *ServerContext) {
...
+
+	// Roles ==> bucketName to role
+	requiredRoles := make([]string, 0)
+	if bucket == "" {
+		// Check cluster read only admin
+		requiredRoles = append(requiredRoles, "ro_admin")
+		requiredRoles = append(requiredRoles, "admin")
+	} else {
+		// Check bucket roles
+		requiredRoles = append(requiredRoles, "mobile_sync_gateway")
```
JRascagneres, line 1353: add equivalent roles

rest/config.go
```diff
@@ -1284,6 +1285,123 @@ func startServer(config *ServerConfig, sc *ServerContext) {
...
+	}
+
+	err = base.JSONUnmarshal(bodyResult, &permissions)
+	if err != nil {
+		return http.StatusInternalServerError, err
+	}
+
+	for _, permResult := range permissions {
+		if permResult {
+			fmt.Println("YAY PERMS")
```
JRascagneres, line 1337: remove

rest/admin_api_test.go
```diff
@@ -2966,3 +2966,123 @@ func TestObtainUserChannelsForDeletedRoleCasFail(t *testing.T) {
...
+			BucketName:          "db",
+			requiredPermissions: nil,
+			expectedStatus:      http.StatusUnauthorized,
+		},
+	}
+
+	testBucket := base.GetTestBucket(t)
+	defer testBucket.Close()
+
+	MakeUser(t, "fulladmintest", "password", []string{"ro_admin"})
```
JRascagneres, line 3074: Add make / remove for all users to test with

rest/config.go
```diff
@@ -1284,6 +1285,123 @@ func startServer(config *ServerConfig, sc *ServerContext) {
...
+			return http.StatusOK, nil
+		}
+	}
+
+	// Whoami
+	// If has role accept, otherwise 403
+
+	// Roles ==> bucketName to role
```
adamcfraser, line 1345: Need to discuss whether we should fall back to role checks for server versions that include mobile permissions.  I'm thinking no - that we should grant the permissions to the Sync Gateway role in that scenario.  Needs review of how legacy roles are migrated during upgrade (particularly older roles like application access, etc).

rest/config.go
```diff
@@ -1284,6 +1285,123 @@ func startServer(config *ServerConfig, sc *ServerContext) {
...
 
+func TestAuth(cluster *gocb.Cluster, username, password string, bucket string, requirePermissions ...string) (statusCode int, err error) {
+	_ = cluster
+	// TODO: Replace with cluster get Management endpoints
+	managementEndpoint := "http://work-couchbase.lan:8091"
+
+	// TODO: Get http client from cluster
+	httpClient := http.DefaultClient
+
+	body := strings.Join(requirePermissions, ",")
```
adamcfraser, line 1296: Are you expecting bucket/collection-level permissions to already have that encoded in requirePermissions?  The use of bucket for roles but not permissions is a bit unclear.

rest/config.go
```diff
@@ -1284,6 +1285,123 @@ func startServer(config *ServerConfig, sc *ServerContext) {
...
+	// Roles ==> bucketName to role
+	requiredRoles := make([]string, 0)
+	if bucket == "" {
+		// Check cluster read only admin
+		requiredRoles = append(requiredRoles, "ro_admin")
+		requiredRoles = append(requiredRoles, "admin")
+	} else {
+		// Check bucket roles
+		requiredRoles = append(requiredRoles, "mobile_sync_gateway")
+	}
```
adamcfraser, line 1354: Assuming this is split out into a role-only function, I expect we'd want to pass in the set of roles to check, in the same way we do for permissions.

rest/config.go
```diff
@@ -1284,6 +1285,123 @@ func startServer(config *ServerConfig, sc *ServerContext) {
...
+	// TODO: Replace with cluster get Management endpoints
+	managementEndpoint := "http://work-couchbase.lan:8091"
+
+	// TODO: Get http client from cluster
+	httpClient := http.DefaultClient
+
+	body := strings.Join(requirePermissions, ",")
+	checkPermissionsURL := fmt.Sprintf("%s/pools/default/checkPermissions", managementEndpoint)
+
+	req, err := http.NewRequest("POST", checkPermissionsURL, bytes.NewBufferString(body))
```
adamcfraser, line 1299: Somewhere in this call stack I expect we'll want to add retry for transient HTTP errors.

rest/config.go
JRascagneres: Yeah, it does take an array at the moment but I could split it into another function and have it take them as arguments.

rest/config.go
JRascagneres: Yeah true... I should probably do that encoding like I am for roles

rest/config.go
JRascagneres: Yeah thats fair, not sure where that ended up when we were talking it through

rest/config.go
```diff
@@ -1284,6 +1285,123 @@ func startServer(config *ServerConfig, sc *ServerContext) {
...
+		requiredRoles = append(requiredRoles, "ro_admin")
+		requiredRoles = append(requiredRoles, "admin")
+	} else {
+		// Check bucket roles
+		requiredRoles = append(requiredRoles, "mobile_sync_gateway")
+	}
+
+	whoamiURL := fmt.Sprintf("%s/whoami", managementEndpoint)
+
+	req, err = http.NewRequest("GET", whoamiURL, nil)
```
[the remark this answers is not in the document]
JRascagneres: I am doing?

PR #5037, opened 2021-06-10
CBG-1486 Mobile 2.x replication protocol changes to support the new attachment storage model

sarathkumarsivan:
Mobile 2.x replication protocol changes to support the new attachment storage model.

### DIFF
--- a/db/attachment.go
+++ b/db/attachment.go
@@ -12,6 +12,7 @@ import (
 	"crypto/rand"
 	"crypto/sha1"
 	"encoding/base64"
+	"fmt"
 
 	"github.com/couchbase/sync_gateway/base"
 )
@@ -209,7 +210,7 @@ type AttachmentCallback func(name string, digest string, knownData []byte, meta 
 // its data. The callback is told whether the attachment body is known to the database, according
 // to its digest. If the attachment isn't known, the callback can return data for it, which will
 // be added to the metadata as a "data" property.
-func (db *Database) ForEachStubAttachment(body Body, minRevpos int, callback AttachmentCallback, docID string) error {
+func (db *Database) ForEachStubAttachment(body Body, minRevpos int, docID string, callback AttachmentCallback) error {
 	atts := GetBodyAttachments(body)
 	if atts == nil && body[BodyAttachments] != nil {
 		return base.HTTPErrorf(400, "Invalid _attachments")
@@ -313,6 +314,8 @@ func ToAttachmentStorageMeta(attachments AttachmentsMeta) []AttachmentStorageMet
 					m := AttachmentStorageMeta{digest: digestString}
 					if version, ok := base.ToInt64(attMap["ver"]); ok {
 						m.version = version
+					} else {
+						m.version = int64(AttachmentStorageModelVersion1)
 					}
 					meta = append(meta, m)
 				}
@@ -341,4 +344,21 @@ func Sha1DigestKey(data []byte) string {
 	digester := sha1.New()
 	digester.Write(data)
 	return "sha1-" + base64.StdEncoding.EncodeToString(digester.Sum(nil))
+}
+
+// AttachmentStorageModelVersion is used to distinguish
+// between legacy and non-legacy attachments.
+type AttachmentStorageModelVersion int64
+
+const (
+	// AttachmentStorageModelVersion1 on the attachment metadata indicates that the
+	// attachment is shared across documents — multiple documents can reference a
+	// given attachment, and multiple revisions of the same document can reference
+	// a given attachment.
+	AttachmentStorageModelVersion1 AttachmentStorageModelVersion = iota + 1
+)
+
+// String returns string representation of the attachment version.
+func (v AttachmentStorageModelVersion) String() string {
+	return fmt.Sprintf("%d", int(v))
 }
--- a/db/attachment.go
+++ b/db/attachment.go
@@ -209,7 +209,7 @@ type AttachmentCallback func(name string, digest string, knownData []byte, meta 
 // its data. The callback is told whether the attachment body is known to the database, according
 // to its digest. If the attachment isn't known, the callback can return data for it, which will
 // be added to the metadata as a "data" property.
-func (db *Database) ForEachStubAttachment(body Body, minRevpos int, callback AttachmentCallback) error {
+func (db *Database) ForEachStubAttachment(body Body, minRevpos int, callback AttachmentCallback, docID string) error {
 	atts := GetBodyAttachments(body)
 	if atts == nil && body[BodyAttachments] != nil {
 		return base.HTTPErrorf(400, "Invalid _attachments")
@@ -293,6 +293,33 @@ func AttachmentDigests(attachments AttachmentsMeta) []string {
 		}
 	}
 	return digests
+}
+
+// AttachmentStorageMeta holds the metadata for building
+// the key for attachment storage and retrieval.
+type AttachmentStorageMeta struct {
+	digest  string
+	version int64
+}
+
+// ToAttachmentStorageMeta returns a slice of AttachmentStorageMeta, which is contains the
+// necessary metadata properties to build the key for attachment storage and retrieval.
+func ToAttachmentStorageMeta(attachments AttachmentsMeta) []AttachmentStorageMeta {
+	meta := make([]AttachmentStorageMeta, 0, len(attachments))
+	for _, att := range attachments {
+		if attMap, ok := att.(map[string]interface{}); ok {
+			if digest, ok := attMap["digest"]; ok {
+				if digestString, ok := digest.(string); ok {
+					m := AttachmentStorageMeta{digest: digestString}
+					if version, ok := base.ToInt64(attMap["ver"]); ok {
+						m.version = version
+					}
+					meta = append(meta, m)
+				}
+			}
+		}
+	}
+	return meta
 }
 
 func attachmentKeyToString(key AttachmentKey) string {
--- a/db/attachment.go
+++ b/db/attachment.go
@@ -30,6 +30,7 @@ type DocAttachment struct {
 	Length      int    `json:"length,omitempty"`
 	Revpos      int    `json:"revpos,omitempty"`
 	Stub        bool   `json:"stub,omitempty"`
+	Version     int64  `json:"ver,omitempty"`
 	Data        []byte `json:"-"` // tell json marshal/unmarshal to ignore this field
 }
 
@@ -313,7 +314,7 @@ func ToAttachmentStorageMeta(attachments AttachmentsMeta) []AttachmentStorageMet
 					if version, ok := base.ToInt64(attMap["ver"]); ok {
 						m.version = version
 					} else {
-						m.version = legacyAttachmentVersion
+						m.version = LegacyAttachmentVersion
 					}
 					meta = append(meta, m)
 				}
@@ -344,8 +345,14 @@ func Sha1DigestKey(data []byte) string {
 	return "sha1-" + base64.StdEncoding.EncodeToString(digester.Sum(nil))
 }
 
-// legacyAttachmentVersion on the attachment metadata indicates that the
-// attachment is shared across documents — multiple documents can reference a
-// given attachment, and multiple revisions of the same document can reference
-// a given attachment.
-const legacyAttachmentVersion int64 = 1
+const (
+	// LegacyAttachmentVersion on the attachment metadata indicates that the
+	// attachment is shared across documents — multiple documents can reference a
+	// given attachment, and multiple revisions of the same document can reference
+	// a given attachment.
+	LegacyAttachmentVersion int64 = 1
+
+	// NonLegacyAttachmentVersion on the attachment metadata indicates that the
+	// attachment is not shared across documents.
+	NonLegacyAttachmentVersion int64 = 2
+)
--- a/db/attachment.go
+++ b/db/attachment.go
@@ -12,6 +12,7 @@ import (
 	"crypto/rand"
 	"crypto/sha1"
 	"encoding/base64"
+	"encoding/json"
 	"github.com/couchbase/sync_gateway/base"
 )
 
@@ -72,6 +73,9 @@ func (db *Database) storeAttachments(doc *Document, newAttachmentsMeta Attachmen
 				"stub":   true,
 				"digest": string(key),
 				"revpos": generation,
+			}
+			if version, ok := meta["ver"].(json.Number); ok {
+				newMeta["ver"] = version
 			}
 			if contentType, ok := meta["content_type"].(string); ok {
 				newMeta["content_type"] = contentType
@@ -319,8 +323,10 @@ func ToAttachmentStorageMeta(attachments AttachmentsMeta) []AttachmentStorageMet
 			if digest, ok := attMap["digest"]; ok {
 				if digestString, ok := digest.(string); ok {
 					m := AttachmentStorageMeta{digest: digestString}
-					if version, ok := attMap["ver"].(int); ok {
-						m.version = version
+					if ver, ok := attMap["ver"].(json.Number); ok {
+						if version, err := ver.Int64(); err == nil {
+							m.version = int(version)
+						}
 					} else {
 						m.version = AttVersion1
 					}
--- a/db/attachment.go
+++ b/db/attachment.go
@@ -12,8 +12,6 @@ import (
 	"crypto/rand"
 	"crypto/sha1"
 	"encoding/base64"
-	"fmt"
-
 	"github.com/couchbase/sync_gateway/base"
 )
 
@@ -315,7 +313,7 @@ func ToAttachmentStorageMeta(attachments AttachmentsMeta) []AttachmentStorageMet
 					if version, ok := base.ToInt64(attMap["ver"]); ok {
 						m.version = version
 					} else {
-						m.version = int64(AttachmentStorageModelVersion1)
+						m.version = legacyAttachmentVersion
 					}
 					meta = append(meta, m)
 				}
@@ -346,19 +344,8 @@ func Sha1DigestKey(data []byte) string {
 	return "sha1-" + base64.StdEncoding.EncodeToString(digester.Sum(nil))
 }
 
-// AttachmentStorageModelVersion is used to distinguish
-// between legacy and non-legacy attachments.
-type AttachmentStorageModelVersion int64
-
-const (
-	// AttachmentStorageModelVersion1 on the attachment metadata indicates that the
-	// attachment is shared across documents — multiple documents can reference a
-	// given attachment, and multiple revisions of the same document can reference
-	// a given attachment.
-	AttachmentStorageModelVersion1 AttachmentStorageModelVersion = iota + 1
-)
-
-// String returns string representation of the attachment version.
-func (v AttachmentStorageModelVersion) String() string {
-	return fmt.Sprintf("%d", int(v))
-}
+// legacyAttachmentVersion on the attachment metadata indicates that the
+// attachment is shared across documents — multiple documents can reference a
+// given attachment, and multiple revisions of the same document can reference
+// a given attachment.
+const legacyAttachmentVersion int64 = 1
--- a/db/attachment.go
+++ b/db/attachment.go
@@ -15,6 +15,14 @@ import (
 	"github.com/couchbase/sync_gateway/base"
 )
 
+const (
+	// AttVersion1 attachments are persisted to the bucket based on attachment body digest.
+	AttVersion1 int = 1
+
+	// AttVersion2 attachments are persisted to the bucket based on docID and body digest.
+	AttVersion2 int = 2
+)
+
 // Key for retrieving an attachment from Couchbase.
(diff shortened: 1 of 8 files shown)


db/blip_handler.go
```diff
@@ -1060,48 +1076,66 @@ func (bh *blipHandler) downloadOrVerifyAttachments(sender *blip.Sender, body Bod
 			}
 
 			return nil, proveAttErr
-		})
+		}, docID)
```
adamcfraser: Having the additional parameter trailing the callback separates the parameters from the function call, making the code more difficult to read.  I'd suggest changing the order of parameters for ForEachStubAttachment to move docID into the third position.

db/blip_handler.go
```diff
@@ -940,9 +940,20 @@ func (bh *blipHandler) handleGetAttachment(rq *blip.Message) error {
 	if digest == "" {
 		return base.HTTPErrorf(http.StatusBadRequest, "Missing 'digest'")
 	}
-	if !bh.isAttachmentAllowed(digest) {
+
+	key := digest
```
adamcfraser: 'key' is ambiguous here.  Would be better as attachmentAllowedKey.

db/blip_handler.go
```diff
@@ -968,6 +979,11 @@ func (bh *blipHandler) sendGetAttachment(sender *blip.Sender, docID string, name
 	if isCompressible(name, meta) {
 		outrq.Properties[BlipCompress] = "true"
 	}
+
+	if bh.blipContext.ActiveProtocol() == BlipCBMobileReplicationV3 {
```
adamcfraser: ActiveProtocol is a relatively expensive call (it's doing strings.SplitN every time), so I don't think it makes sense to call it more than once per blipContext.  Can we store this on the context closer to context initialization?

db/attachment.go
```diff
@@ -295,6 +295,33 @@ func AttachmentDigests(attachments AttachmentsMeta) []string {
...
+// ToAttachmentStorageMeta returns a slice of AttachmentStorageMeta, which is contains the
+// necessary metadata properties to build the key for attachment storage and retrieval.
+func ToAttachmentStorageMeta(attachments AttachmentsMeta) []AttachmentStorageMeta {
+	meta := make([]AttachmentStorageMeta, 0, len(attachments))
+	for _, att := range attachments {
+		if attMap, ok := att.(map[string]interface{}); ok {
+			if digest, ok := attMap["digest"]; ok {
+				if digestString, ok := digest.(string); ok {
+					m := AttachmentStorageMeta{digest: digestString}
+					if version, ok := base.ToInt64(attMap["ver"]); ok {
```
adamcfraser: In the case that version isn't specified, we're treating it as version=0.  Is this intentional?

db/attachment_test.go
```diff
@@ -438,44 +438,45 @@ func TestForEachStubAttachmentErrors(t *testing.T) {
 
 	// Call ForEachStubAttachment with invalid attachment; simulates the error scenario.
 	doc := `{"_attachments": "No Attachment"}`
+	docID := "foo"
 	assert.NoError(t, base.JSONUnmarshal([]byte(doc), &body))
-	err = db.ForEachStubAttachment(body, 0x1, callback)
+	err = db.ForEachStubAttachment(body, 0x1, callback, docID)
```
adamcfraser: Why is this test using hex format for ints?

db/blip_handler.go
```diff
@@ -1060,48 +1076,66 @@ func (bh *blipHandler) downloadOrVerifyAttachments(sender *blip.Sender, body Bod
...
-		bsc.allowedAttachments[digest] = bsc.allowedAttachments[digest] + 1
+	for _, attachment := range attMeta {
+		if bsc.blipContext.ActiveProtocol() == BlipCBMobileReplicationV3 {
+			key := docID + attachment.digest
+			if _, found := bsc.allowedAttachments[key]; !found {
+				bsc.allowedAttachments[key] = AllowedAttachment{counter: 1}
+			}
+		} else {
+			att := bsc.allowedAttachments[attachment.digest]
+			att.counter = att.counter + 1
```
adamcfraser: Shouldn't the use of counter depend on attachment version, not peer replication version?

db/blip_handler.go
```diff
@@ -1060,48 +1076,66 @@ func (bh *blipHandler) downloadOrVerifyAttachments(sender *blip.Sender, body Bod
...
+			att.docID = docID
+			att.version = attachment.version
```
adamcfraser: If att already exists in bsc.allowedAttachments, is there a need to update docID/version?

adamcfraser (changes requested):
A few comments on implementation, but particularly want a review of the replication version/attachment version handling in addAllowedAttachments/removeAllowedAttachments.

db/blip_handler.go
sarathkumarsivan: Yeah, use of counter depends on the attachment version.

db/blip_sync_messages.go
```diff
@@ -439,11 +439,15 @@ func (g *getAttachmentParams) digest() string {
 	return g.rq.Properties[GetAttachmentDigest]
 }
 
+func (g *getAttachmentParams) id() string {
+	return g.rq.Properties[GetAttachmentID]
```
bbrks, line 443: As discussed, change to DocID for clarity

db/blip_handler.go
```diff
@@ -1067,41 +1083,63 @@ func (bsc *BlipSyncContext) incrementSerialNumber() uint64 {
...
 	}
 
 	bsc.lock.Lock()
 	defer bsc.lock.Unlock()
 
-	for _, digest := range attDigests {
-		if n := bsc.allowedAttachments[digest]; n > 1 {
-			bsc.allowedAttachments[digest] = n - 1
+	for _, attachment := range attMeta {
+		if bsc.blipContext.ActiveProtocol() == BlipCBMobileReplicationV3 {
```
bbrks: This logic seems related to attachment version more than protocol version

db/blip_handler.go
```diff
@@ -1067,41 +1083,63 @@ func (bsc *BlipSyncContext) incrementSerialNumber() uint64 {
...
 	if bsc.allowedAttachments == nil {
-		bsc.allowedAttachments = make(map[string]int, 100)
+		bsc.allowedAttachments = make(map[string]AllowedAttachment, 100)
 	}
-	for _, digest := range attDigests {
-		bsc.allowedAttachments[digest] = bsc.allowedAttachments[digest] + 1
+	for _, attachment := range attMeta {
+		key := attachment.digest
+		if bsc.blipContext.ActiveProtocol() == BlipCBMobileReplicationV3 {
+			key = docID + attachment.digest
```
bbrks: Again, by moving the protocol check up one level, you can basically always have key be `docID + attachment.digest`, even when `docID` is unset.

db/blip_sync_messages.go
```diff
@@ -439,11 +439,15 @@ func (g *getAttachmentParams) digest() string {
...
+}
+
 func (g *getAttachmentParams) String() string {
 
 	buffer := bytes.NewBufferString("")
 
-	buffer.WriteString(fmt.Sprintf("Digest:%v ", g.digest()))
+	buffer.WriteString(fmt.Sprintf("Digest:%v, ID: %v ", g.digest(), g.id()))
 
 	return buffer.String()
```
bbrks: Old code but seems unnessesary.

```return fmt.Sprintf("Digest:%v, ID: %v ", g.digest(), g.id())```

File: db/blip_sync_context.go [hunk 12]
```diff
@@ -570,9 +579,10 @@ func (bsc *BlipSyncContext) sendRevision(sender *blip.Sender, docID, revID strin
 
 	history := toHistory(rev.History, knownRevs, maxHistory)
 	properties := blipRevMessageProperties(history, rev.Deleted, seq)
-	attDigests := AttachmentDigests(rev.Attachments)
-	base.DebugfCtx(bsc.loggingCtx, base.KeySync, "Sending rev %q %s based on %d known, digests: %v", base.UD(docID), revID, len(knownRevs), attDigests)
-	return bsc.sendRevisionWithProperties(sender, docID, revID, bodyBytes, attDigests, properties, seq, nil)
+	attachmentStorageMeta := ToAttachmentStorageMeta(rev.Attachments)
+	base.DebugfCtx(bsc.loggingCtx, base.KeySync, "Sending rev %q %s based on %d known, digests: %v", base.UD(docID),
+		revID, len(knownRevs), attachmentStorageMeta)
```
bbrks: Do we still want to log digests here rather than metadata.

bbrks (changes requested):

db/blip_handler.go
sarathkumarsivan: Sure, that make sense.

rest/blip_api_test.go
```diff
@@ -1449,54 +1450,70 @@ func TestPutAttachmentViaBlipGetViaRest(t *testing.T) {
...
-	digest := db.Sha1DigestKey([]byte(attachmentBody))
+	var tests = []struct {
+		name                   string
+		inputBlipProtocol      string
+		inputAttachmentVersion int64
+	}{
+		{"TestPutAttViaBlipGetViaBlipCBMobile2AttVer1", db.BlipCBMobileReplicationV2, db.LegacyAttachmentVersion},
+		{"TestPutAttViaBlipGetViaBlipCBMobile2AttVer2", db.BlipCBMobileReplicationV2, db.NonLegacyAttachmentVersion},
+		{"TestPutAttViaBlipGetViaBlipCBMobile3AttVer1", db.BlipCBMobileReplicationV3, db.LegacyAttachmentVersion},
+		{"TestPutAttViaBlipGetViaBlipCBMobile3AttVer2", db.BlipCBMobileReplicationV3, db.NonLegacyAttachmentVersion},
```
adamcfraser: To properly test these cases I expect you'd need multiple docs referencing the same attachment digest - otherwise you're not testing any of the new allowedAttachment indexing.

rest/utilities_testing.go
```diff
@@ -1049,14 +1049,15 @@ func (bt *BlipTester) GetDocAtRev(requestedDocID, requestedDocRev string) (resul
...
-	attachmentDigest string
-	history          []string
-	body             []byte
+	docId             string
+	revId             string
+	attachmentName    string
+	attachmentLength  int
+	attachmentBody    string
+	attachmentDigest  string
+	attachmentVersion int64
```
adamcfraser: As above, simple int would be sufficient.

adamcfraser (changes requested):

rest/blip_api_test.go
sarathkumarsivan: Makes sense, the TestPutAttachmentViaBlipGetViaBlip has been refactored with two documents referencing the same attachment.

sarathkumarsivan (commented):
Refactored allowed attachment handling logic and the TestPutAttachmentViaBlipGetViaBlip.

db/blip_handler.go
```diff
@@ -1083,41 +1099,70 @@ func (bsc *BlipSyncContext) incrementSerialNumber() uint64 {
...
+	}
+	for _, attachment := range attMeta {
+		key := bsc.allowedAttachmentKey(docID, attachment.digest)
+		att, found := bsc.allowedAttachments[key]
+		if found {
+			if att.version == AttVersion1 {
+				att.counter = att.counter + 1
+				bsc.allowedAttachments[key] = att
+			} else if att.version == AttVersion2 {
+				bsc.allowedAttachments[key] = AllowedAttachment{}
```
adamcfraser: Why are you setting the value to an empty AllowedAttachment{} -  won't this lose the docID needed to retrieve the attachment?  Please ensure there is a unit test covering this path.

adamcfraser (changes requested):

db/blip_handler.go
sarathkumarsivan: That was logically incorrect. We will miss the docID to retrieve the attachment if we set the value to an empty AllowedAttachment{}. The else block can be removed since we don't need to increment or decrement the counter for AttVersion2. And the value of the counter will be set to 1 when the attachment is added to allowedAttachments map in the first place. TestPutAttachmentViaBlipGetViaBlip has the coverage for V1 and V2 attachments handling. Also added a change in storeAttachments set the version if it's passed the document body along with attachment metadata. It's necessary for adding the coverage.

db/blip_handler.go
```diff
@@ -1083,41 +1099,68 @@ func (bsc *BlipSyncContext) incrementSerialNumber() uint64 {
...
+		key := bsc.allowedAttachmentKey(docID, attachment.digest)
+		att, found := bsc.allowedAttachments[key]
+		if found {
+			if att.version == AttVersion1 {
+				att.counter = att.counter + 1
+				bsc.allowedAttachments[key] = att
+			}
+		} else {
+			bsc.allowedAttachments[key] = AllowedAttachment{
+				docID:   docID,
```
adamcfraser: Is docID needed inside the AllowedAttachment struct?  Where does it get used?

rest/blip_api_test.go
```diff
@@ -1449,55 +1450,163 @@ func TestPutAttachmentViaBlipGetViaRest(t *testing.T) {
...
+						Revpos:      1,
+						Stub:        true,
+						Digest:      digest,
+						Version:     tt.inputAttachmentVersion,
+						Length:      len(attachmentBody),
+						Data:        []byte(attachmentBody),
+					},
+				},
+			}
+			sent, _, _ := bt.SendRevWithAttachments(input)
```
adamcfraser: As in the comments elsewhere, I don't think we expect Version to be sent from clients?

adamcfraser (changes requested):
I don't believe clients will be sending version - added notes on that.

db/attachment_test.go
```diff
@@ -1359,3 +1360,131 @@ func TestMigrateBodyAttachmentsMergeConflicting(t *testing.T) {
...
+				blipContextDb: &Database{Ctx: context.TODO()},
+				blipContext:   blipContext,
+			}
+			meta := []AttachmentStorageMeta{
+				{digest: "digest1", version: tt.inputAttVersion},
+				{digest: "digest2", version: tt.inputAttVersion},
+			}
+			ctx.addAllowedAttachments("doc1", meta)
+			require.True(t, ctx.isAttachmentAllowed("digest1"))
+			require.True(t, ctx.isAttachmentAllowed("digest2"))
```
adamcfraser: In the case where blipProtocol==BlipCBMobileReplicationV3, won't attachments be stored in allowedAttachments indexed by docID+digest?  These tests don't seem to be accounting for that when calling isAttachmentAllowed.

adamcfraser (changes requested):
Tests don't appear to be actually testing BlipCBMobileReplicationV3 handling.

db/attachment_test.go
sarathkumarsivan: Makes sense, when looking at this further, it seems that the protocol version is alway returned as BlipCBMobileReplicationV2. To fix that we can make use of the NewBlipTesterFromSpecWithRT, but the  TestAllowedAttachments needs to be moved to the rest package. In addition to that, a setter method on BlipSyncContext is necessary to set the blip.Context.

db/attachment_test.go
```diff
@@ -1359,3 +1359,129 @@ func TestMigrateBodyAttachmentsMergeConflicting(t *testing.T) {
...
+
+	// Single document associated with multiple attachments of different digests.
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := &BlipSyncContext{}
+			meta := []AttachmentStorageMeta{
+				{digest: "digest1", version: tt.inputAttVersion},
+				{digest: "digest2", version: tt.inputAttVersion},
+			}
+			var docID = "doc1"
```
adamcfraser: When running with BlipCBMobileReplicationV2, we won't know the docID when calling isAttachmentAllowed.  Shouldn't the test be following that pattern as well?

db/attachment_test.go
```diff
@@ -1359,3 +1359,129 @@ func TestMigrateBodyAttachmentsMergeConflicting(t *testing.T) {
...
+
+			ctx.addAllowedAttachments(docID, meta, tt.inputBlipProtocol)
+			requireIsAttachmentAllowedTrue(t, ctx, docID, meta, tt.inputBlipProtocol)
+
+			ctx.removeAllowedAttachments(docID, meta, tt.inputBlipProtocol)
+			requireIsAttachmentAllowedFalse(t, ctx, docID, meta, tt.inputBlipProtocol)
+		})
+	}
+
+	// Single document associated with multiple attachments of similar digests.
```
adamcfraser:
```suggestion
	// Single document associated with multiple attachments with matching digests.
```

db/attachment_test.go
```diff
@@ -1359,3 +1359,129 @@ func TestMigrateBodyAttachmentsMergeConflicting(t *testing.T) {
...
+
+	// Multiple documents associated with multiple attachments of different digests.
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := &BlipSyncContext{}
+			meta := []AttachmentStorageMeta{
+				{digest: "digest1", version: tt.inputAttVersion},
+				{digest: "digest2", version: tt.inputAttVersion},
+			}
+			var (
```
adamcfraser: Why use a var block for these, as opposed to just docID1 := "doc1"

db/attachment_test.go
```diff
@@ -1359,3 +1359,129 @@ func TestMigrateBodyAttachmentsMergeConflicting(t *testing.T) {
...
+			ctx.addAllowedAttachments(docID, meta, tt.inputBlipProtocol)
+			key := allowedAttachmentKey(docID, meta[0].digest, tt.inputBlipProtocol)
+			require.True(t, ctx.isAttachmentAllowed(key))
+
+			ctx.removeAllowedAttachments(docID, meta, tt.inputBlipProtocol)
+			require.False(t, ctx.isAttachmentAllowed(key))
+		})
+	}
+
+	// Multiple documents associated with multiple attachments of different digests.
```
adamcfraser: You don't seem to have the simple case of two document with different attachments - here you're attaching the same two attachments to different documents.  Might be nice to have the basic case as well, to make sure it doesn't get broken.

db/attachment_test.go
```diff
@@ -1359,3 +1359,129 @@ func TestMigrateBodyAttachmentsMergeConflicting(t *testing.T) {
...
+			var (
+				docID1 = "doc1"
+				docID2 = "doc2"
+			)
+			ctx.addAllowedAttachments(docID1, meta, tt.inputBlipProtocol)
+			requireIsAttachmentAllowedTrue(t, ctx, docID1, meta, tt.inputBlipProtocol)
+			ctx.addAllowedAttachments(docID2, meta, tt.inputBlipProtocol)
+			requireIsAttachmentAllowedTrue(t, ctx, docID2, meta, tt.inputBlipProtocol)
+
+			ctx.removeAllowedAttachments(docID1, meta, tt.inputBlipProtocol)
```
adamcfraser, line 1483: We should be checking isAllowedAttachment for docID2 here, after removing docID1.

adamcfraser (changes requested):
Have a few comments/questions about tests still remaining.

db/attachment_test.go
sarathkumarsivan: Sure, added checking isAllowedAttachment for docID2 here based on replication versions and attachment versions.

db/attachment_test.go
```diff
@@ -1425,63 +1433,101 @@ func TestAllowedAttachments(t *testing.T) {
...
-				requireIsAttachmentAllowedTrue(t, ctx, docID1, meta, tt.inputBlipProtocol)
-			} else {
+
+			if tt.inputBlipProtocol == BlipCBMobileReplicationV2 {
+				if tt.inputAttVersion == AttVersion1 {
+					requireIsAttachmentAllowedTrue(t, ctx, docID1, meta, tt.inputBlipProtocol)
+					requireIsAttachmentAllowedTrue(t, ctx, docID2, meta, tt.inputBlipProtocol)
+				} else if tt.inputAttVersion == AttVersion2 {
+					requireIsAttachmentAllowedFalse(t, ctx, docID1, meta, tt.inputBlipProtocol)
+					requireIsAttachmentAllowedFalse(t, ctx, docID2, meta, tt.inputBlipProtocol)
```
adamcfraser: At this point we haven't removed docID2, so isAttachmentAllowed should still be returning true.  This looks to be related to the question discussed earlier, about whether the counter should be incremented in the BlipCBMobileReplicationV2, AttVersion2 case.  Based on this I think the answer is yes - as long as we're using BlipCBMobileReplicationV2 there may be multiple in-flight revisions with matching allowedAttachments indexes.

adamcfraser (changes requested):

db/attachment_test.go
sarathkumarsivan: Yeah, that makes sense, changes are included to increment the counter when the protocol is BlipCBMobileReplicationV2 and attachment version is AttVersion2.

adamcfraser (approved):

PR #5042, opened 2021-06-14
CBG-1498: Ensure removed notifications only sent when doc not accessible

JRascagneres:
Fixes issue where removals would still be sent when doc is accessible through an alternate channel.

Changes:
 - Swapped `bh.purgeOnRemoval` setting to `opts.revocations` instead. Purge on removal is only set through ISGR config option whereas the revocations option is intended to be used internally and is set when subChanges requests them.
 - When user still has access to doc send the change as non-removal. This is okay to do (and we should do) because the doc has been mutated to trigger the removal so the user needs that change, however, we shouldn't send as a removal because they still have access through another channel.

Added a test which ensures that removals only come over when there is no access left.

Based onto https://github.com/couchbase/sync_gateway/pull/5044
- [x] Merge above

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -320,7 +320,7 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 				for _, item := range change.Changes {
 					changeRow := bh.buildChangesRow(change, item["rev"])
 
-					if opts.revocations && len(change.Removed) > 0 && bh.blipContext.ActiveProtocol() == BlipCBMobileReplicationV3 {
+					if opts.revocations && change.allRemoved && bh.blipContext.ActiveProtocol() == BlipCBMobileReplicationV3 {
 						userHasAccessToDoc, err := UserHasDocAccess(bh.db, change.ID, item["rev"])
 						if err != nil {
 							base.InfofCtx(bh.loggingCtx, base.KeySync, "Unable to obtain the doc: %s %s to verify user access: %v", base.UD(change.ID), item["rev"], err)
@@ -328,7 +328,7 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 						}
 
 						if userHasAccessToDoc {
-							change.Removed = nil
+							continue
 						}
 					}
 
@@ -377,7 +377,7 @@ func (bh *blipHandler) buildChangesRow(change *ChangeEntry, revID string) []inte
 		if change.Revoked {
 			deletedFlags |= changesDeletedFlagRevoked
 		}
-		if len(change.Removed) > 0 {
+		if change.allRemoved {
 			deletedFlags |= changesDeletedFlagRemoved
 		}
 
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -320,7 +320,7 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 				for _, item := range change.Changes {
 					changeRow := bh.buildChangesRow(change, item["rev"])
 
-					if bh.purgeOnRemoval && len(change.Removed) > 0 && bh.blipContext.ActiveProtocol() == BlipCBMobileReplicationV3 {
+					if opts.revocations && len(change.Removed) > 0 && bh.blipContext.ActiveProtocol() == BlipCBMobileReplicationV3 {
 						userHasAccessToDoc, err := UserHasDocAccess(bh.db, change.ID, item["rev"])
 						if err != nil {
 							base.InfofCtx(bh.loggingCtx, base.KeySync, "Unable to obtain the doc: %s %s to verify user access: %v", base.UD(change.ID), item["rev"], err)
@@ -328,7 +328,7 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 						}
 
 						if userHasAccessToDoc {
-							continue
+							change.Removed = nil
 						}
 					}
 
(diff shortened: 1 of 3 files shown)


db/blip_handler.go
```diff
@@ -320,15 +320,15 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 				for _, item := range change.Changes {
 					changeRow := bh.buildChangesRow(change, item["rev"])
 
-					if bh.purgeOnRemoval && len(change.Removed) > 0 && bh.blipContext.ActiveProtocol() == BlipCBMobileReplicationV3 {
+					if opts.revocations && len(change.Removed) > 0 && bh.blipContext.ActiveProtocol() == BlipCBMobileReplicationV3 {
```
adamcfraser: As discussed, we should use allRemoved here.

adamcfraser (changes requested):

db/blip_handler.go
```diff
@@ -320,7 +320,7 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 				for _, item := range change.Changes {
 					changeRow := bh.buildChangesRow(change, item["rev"])
 
-					if bh.purgeOnRemoval && len(change.Removed) > 0 && bh.blipContext.ActiveProtocol() == BlipCBMobileReplicationV3 {
+					if opts.revocations && change.allRemoved && bh.blipContext.ActiveProtocol() == BlipCBMobileReplicationV3 {
```
adamcfraser: This looks like we're checking opts.revocations in the context of deciding whether or not to send a removal when the user still has access to it.  This seems like an overloading of the meaning of opts.revocations, which I thought was "does the peer wish to receive revocation messages for documents in a channel when they lose access to that channel".

adamcfraser (changes requested):
Would like to discuss and clarify the use of opts.revocations.

db/blip_handler.go
JRascagneres: Well `opts.revocations` means that revocations have been requested via `subChanges`. 
If the issue is the difference between revocations and removals then this option is being used to say 'I want revocations and removals'. This was discussed with CBL (and confirmed tonight).

JRascagneres: Although... To be honest on second though I wonder if we want to have that check at all... Should we just always send this removed message (as we did in 2.8 I think?) but then the client can decide whether to actually purge it or not.

JRascagneres: As discussed with CBL. The expectation is that if revocations subChanges message is not set then they fall back to the 2.8.x behaviour. This being that we don't do the user check and just send all removal messages.

db/blip_handler.go
```diff
@@ -377,7 +377,7 @@ func (bh *blipHandler) buildChangesRow(change *ChangeEntry, revID string) []inte
 		if change.Revoked {
 			deletedFlags |= changesDeletedFlagRevoked
 		}
-		if len(change.Removed) > 0 {
+		if change.allRemoved {
```
JRascagneres, line 380: Double check this... Need to confirm previous behaviour.

db/blip_handler.go
```diff
@@ -320,14 +320,20 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
...
+							userHasAccessToDoc, err := UserHasDocAccess(bh.db, change.ID, item["rev"])
+							if err != nil {
+								base.InfofCtx(bh.loggingCtx, base.KeySync, "Unable to obtain the doc: %s %s to verify user access: %v", base.UD(change.ID), item["rev"], err)
+								continue
+							}
+
+							if userHasAccessToDoc {
+								continue
+							}
+						} else if !change.Deleted {
```
adamcfraser: If we're going to send tombstone removals when opts.revocations=false, should we also be sending tombstone removals when opts.revocations==true && userHasAccessToDoc == true?  (i.e. should the !change.Deleted check be part of the condition on line 323?)

adamcfraser (changes requested):

db/blip_handler.go
JRascagneres: Yeah, thats a fair point. I was assuming tombstones would lose access anyway but on second thoughts that'd be a wrong assumption.
Made that change now.

db/blip_handler.go
```diff
@@ -320,14 +320,22 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
...
-							base.InfofCtx(bh.loggingCtx, base.KeySync, "Unable to obtain the doc: %s %s to verify user access: %v", base.UD(change.ID), item["rev"], err)
-							continue
-						}
-
-						if userHasAccessToDoc {
+					// If change is a removal and we're operating with protocol V3 and the change is not a tombstone fall
+					// into 3.0 removal handling
+					if change.allRemoved && bh.blipContext.ActiveProtocol() == BlipCBMobileReplicationV3 && !change.Deleted {
+						// If client requests revocations / removals via subChanges, continue work to possibly send
+						// removal. Otherwise skip sending removals.
```
adamcfraser: The nested continues here add code complexity.  In particular, it took me until now to realize that we're not sending a removal message when we get an error looking up access, which I think is incorrect?

What do you think about reorganizing along the lines of:
```
// If client doesn't want removals/revocations, don't send change
if !opts.revocations {
    continue
}

// If the user can access the removed doc through another channel, don't send change
userHasAccessToDoc, err == ...
if err == nil && userHasAccess {
    continue
}

// If we can't determine access due to error, log error and send change
if err != nil {
   log()
}

adamcfraser (changes requested):

db/blip_handler.go
JRascagneres: Yeah thats fair, it was getting fairly nested.

adamcfraser (approved):

PR #5043, opened 2021-06-14
CBG-1398 - 3.0 Startup Config

bbrks:
- CBG-1398 - Opt-in 3.0 startup config with backwards compatibility for existing configs.

Existing config handling that I know will break with this pr, that may warrant fixes before merging:
- [x] existing command line flags not wired up to anything ... I should fix this
- [x] some tests with commented out assertions (mostly due to lack of databases in sg/_config response)

Intentionally not implemented in this PR:

- CBG-1399 Legacy config upgrade in persistent mode
- CBG-1458 Bootstrap retry logic
- CBG-1457 Initial database bootstrap


rest/main.go
```diff
@@ -0,0 +1,60 @@
...
+	base.InitializeMemoryLoggers()
+	base.LogSyncGatewayVersion()
+
+	fs := flag.NewFlagSet(osArgs[0], flag.ContinueOnError)
+
+	// used by service scripts as a way to specify a per-distro defaultLogFilePath
+	defaultLogFilePath = *fs.String("defaultLogFilePath", "", "Path to log files, if not overridden by --logFilePath, or the config")
+
+	// TODO: Change default when we're ready to enable 3.0/bootstrap/persistent config by default (once QE's existing tests are ready to handle it)
+	persistentConfigFlag := fs.Bool("persistent_config", false, "Can be set to false to disable persistent config handling, and read all configuration from a legacy config file.")
```
bbrks: Should move this back to `disable_persistent_config` based on discussion in [CBG-1461](https://issues.couchbase.com/browse/CBG-1461)

bbrks:
litecore tests failing - find the config they use and repro

<details>
<summary>Details</summary>


```
Waiting for sync gateway to start
2021-06-22T18:08:00.422Z ==== Couchbase Sync Gateway/() EE ====
2021-06-22T18:08:00.422Z [WRN] Running in legacy config mode -- rest.legacyServerMain() at main_legacy.go:12
2021-06-22T18:08:00.422Z [INF] Loading content from [config.json] ...
2021-06-22T18:08:00.426Z [ERR] Unexpected panic: runtime error: invalid memory address or nil pointer dereference - stopping process
goroutine 1 [running]:
runtime/debug.Stack(0xc0003538b8, 0xf8eba0, 0x1bd23b0)
	/root/.gvm/gos/go1.14/src/runtime/debug/stack.go:24 +0x9d
github.com/couchbase/sync_gateway/base.FatalPanicHandler()
	/var/jenkins/workspace/ub_pipeline_sync_gateway_PR-5043/godeps/src/github.com/couchbase/sync_gateway/base/util.go:1395 +0x57
panic(0xf8eba0, 0x1bd23b0)
	/root/.gvm/gos/go1.14/src/runtime/panic.go:967 +0x15d
github.com/couchbase/sync_gateway/rest.(*LegacyConfig).ToStartupConfig(0xc000353d48, 0xc00036369d, 0x2, 0xc0002d0780, 0x0)
	/var/jenkins/workspace/ub_pipeline_sync_gateway_PR-5043/godeps/src/github.com/couchbase/sync_gateway/rest/config_legacy.go:133 +0x5cd
github.com/couchbase/sync_gateway/rest.legacyServerMain(0xc00000e080, 0x2, 0x2, 0x1, 0x0)
	/var/jenkins/workspace/ub_pipeline_sync_gateway_PR-5043/godeps/src/github.com/couchbase/sync_gateway/rest/main_legacy.go:24 +0x135
github.com/couchbase/sync_gateway/rest.serverMain(0x12f24c0, 0xc00003a050, 0xc00000e080, 0x2, 0x2, 0x0, 0x0)
	/var/jenkins/workspace/ub_pipeline_sync_gateway_PR-5043/godeps/src/github.com/couchbase/sync_gateway/rest/main.go:56 +0x32c
github.com/couchbase/sync_gateway/rest.ServerMain()
	/var/jenkins/workspace/ub_pipeline_sync_gateway_PR-5043/godeps/src/github.com/couchbase/sync_gateway/rest/main.go:15 +0x61
main.main()
	/var/jenkins/workspace/ub_pipeline_sync_gateway_PR-5043/godeps/src/github.com/couchbase/sync_gateway/main.go:24 +0x20
 -- base.FatalPanicHandler() at util.go:1395
```

</details>

rest/config_new.go
```diff
@@ -0,0 +1,240 @@
...
+	Console        *base.ConsoleLoggerConfig `json:"console,omitempty"`
+	Error          *base.FileLoggerConfig    `json:"error,omitempty"`
+	Warn           *base.FileLoggerConfig    `json:"warn,omitempty"`
+	Info           *base.FileLoggerConfig    `json:"info,omitempty"`
+	Debug          *base.FileLoggerConfig    `json:"debug,omitempty"`
+	Trace          *base.FileLoggerConfig    `json:"trace,omitempty"`
+	Stats          *base.FileLoggerConfig    `json:"stats,omitempty"`
+}
+
+type AuthConfig struct {
```
adamcfraser: Everything in AuthConfig is deprecated in Lithium.  Worth discussing whether it should only be supported when using legacy config, and can be omitted from dynamic config.

rest/config_new.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+type UnsupportedConfig struct {
+	StatsLogFrequency *time.Duration `json:"stats_log_frequency,omitempty" help:"How often should stats be written to stats logs"`
+	UseStdlibJSON     bool           `json:"use_stdlib_json,omitempty"     help:"Bypass the jsoniter package and use Go's stdlib instead"`
+
+	HTTP2 *HTTP2Config `json:"http2,omitempty"`
+}
+
+type HTTP2Config struct {
```
adamcfraser: Having Http2Config and HTTP2Config in the same package seems error-prone.  Can we remove the old one, since only the annotations differ?

rest/config_new.go
```diff
@@ -0,0 +1,240 @@
...
+type FacebookConfig struct {
+	Register bool `json:"register,omitempty" help:"If true, server will register new user accounts"`
+}
+
+type GoogleConfig struct {
+	Register    bool     `json:"register,omitempty" help:"If true, server will register new user accounts"`
+	AppClientID []string `json:"app_client_id,omitempty" help:"List of enabled client ids"`
+}
+
+type ReplicatorConfig struct {
```
adamcfraser: sg-replicate 1 is marked for removal in 3.0 - can omit this from new config?

rest/main_new.go
```diff
@@ -0,0 +1,53 @@
...
+// serverMainPersistentConfig runs the Sync Gateway server with persistent config.
+func serverMainPersistentConfig(osArgs []string, fs *flag.FlagSet, flagStartupConfig *StartupConfig) error {
+
+	// 3.0 config - bootstrap and pull configs from server buckets.
+	base.Infof(base.KeyAll, "Running in persistent config mode")
+
+	sc := DefaultStartupConfig(defaultLogFilePath)
+	base.Tracef(base.KeyAll, "default config: %#v", sc)
+
+	fileStartupConfig, err := LoadStartupConfigFromPaths(fs.Args()...)
```
adamcfraser: Is specifying multiple paths on startup a new feature/requirement?  If so, can you remind me of the details?  Doing a simple merge based on the order loaded seems like it would have corner cases that would result in error.

rest/config_new.go
```diff
@@ -0,0 +1,240 @@
...
+	CompressResponses  *bool `json:"compress_responses,omitempty"   help:"If false, disables compression of HTTP responses"`
+	HideProductVersion bool  `json:"hide_product_version,omitempty" help:"Whether product versions removed from Server headers and REST API responses"`
+
+	TLS  TLSConfig   `json:"tls,omitempty"`
+	CORS *CORSConfig `json:"cors,omitempty"`
+}
+
+type TLSConfig struct {
+	MinimumVersion string `json:"minimum_version,omitempty"     help:"The minimum allowable TLS version for the REST APIs"`
+	CertPath       string `json:"cert_path,omitempty" help:"The TLS cert file to use for the REST APIs"`
```
adamcfraser: Do you think that naming these properties the same as client certpath/keypath for connecting to Couchbase Server might lead to confusion, even if they are under a different hierarchy?

rest/config_new.go
```diff
@@ -0,0 +1,240 @@
...
+		base.Errorf("Error setting MaxFileDescriptors to %d: %v", sc.MaxFileDescriptors, err)
+	}
+
+	// Given unscoped usage of base.JSON functions, this can't be scoped.
+	if sc.Unsupported.UseStdlibJSON {
+		base.Infof(base.KeyAll, "Using the stdlib JSON package")
+		base.UseStdlibJSON = true
+	}
+
+	// TODO: Move to be scoped under auth.Authenticator ?
```
adamcfraser: Agree with this todo - makes sense there, and that's where users will be looking for it.

rest/main.go
```diff
@@ -0,0 +1,60 @@
...
+	// TODO: Merge legacyFlagStartupConfig onto default config before merging others.
+	legacyFlagStartupConfig := registerLegacyFlags(fs)
+	_ = legacyFlagStartupConfig
+
+	// register config property flags
+	var flagStartupConfig StartupConfig
+	// TODO: Revisit config cli flags after initial persistent config implementation
+	// if err := clistruct.RegisterJSONFlags(fs, &flagStartupConfig); err != nil {
+	// 	return err
+	// }
```
adamcfraser, line 48: Do we have and/or need tracking tickets for the new TODOs in here?

rest/main_new.go
```diff
@@ -0,0 +1,53 @@
...
+	"flag"
+	"fmt"
+
+	"github.com/couchbase/sync_gateway/base"
+	"github.com/imdario/mergo"
+	pkgerrors "github.com/pkg/errors"
+)
+
+// serverMainPersistentConfig runs the Sync Gateway server with persistent config.
+func serverMainPersistentConfig(osArgs []string, fs *flag.FlagSet, flagStartupConfig *StartupConfig) error {
```
adamcfraser: It looks like osArgs would be pushed into fs by the time this is called?  Is there a pending use case for osArgs here?

rest/server_context.go
```diff
@@ -13,15 +13,14 @@ import (
...
-	"net/url"
 	"os"
 	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
 
-	"github.com/couchbase/go-couchbase"
+	"github.com/couchbase/gocb"
```
adamcfraser: As discussed, it would be preferred to encapsulate the gocb/gocbcore-specific handling in base.  I think that means having something other than an explicit gocb.Cluster reference here, and instead a wrapper struct with the subset of required operations.  I think that would improve testability, but review and advise whether that's feasible.

rest/server_context.go
```diff
@@ -42,14 +41,38 @@ const DefaultStatsLogFrequencySecs = 60
...
+	statsContext        *statsContext
+	bootstrapConnection *gocb.Cluster
+	HTTPClient          *http.Client
+	replicator          *base.Replicator
+	cpuPprofFileMutex   sync.Mutex // Protect cpuPprofFile from concurrent Start and Stop CPU profiling requests
+	cpuPprofFile        *os.File   // An open file descriptor holds the reference during CPU profiling
+}
+
+type DatabaseConfig struct {
+	CAS    gocb.Cas
```
adamcfraser: There's no usage of CAS yet, but would be interested in the details of how it's expected to be used (or omit it until it's implemented).

rest/server_context.go
```diff
@@ -177,34 +208,21 @@ func (sc *ServerContext) GetDatabase(name string) (*db.DatabaseContext, error) {
...
-		}
 	}
-	return dbc, nil
-
+	return nil, base.HTTPErrorf(http.StatusNotFound, "no such database %q", name)
 }
 
-func (sc *ServerContext) GetDatabaseConfig(name string) *DbConfig {
+func (sc *ServerContext) GetDatabaseConfig(name string) *DatabaseConfig {
+	// FIXME
```
adamcfraser: Needs a fix?

rest/admin_api.go
```diff
@@ -134,12 +134,15 @@ func (h *handler) handlePutDbConfig() error {
...
+	config.Server = &h.server.config.Bootstrap.Server
 	if err := config.setup(dbName); err != nil {
 		return err
 	}
 	h.server.lock.Lock()
 	defer h.server.lock.Unlock()
-	h.server.config.Databases[dbName] = config
+	dbc := DatabaseConfig{CAS: 0, Config: *config}
+	h.server.bucketDbConfigs[h.db.Bucket.GetName()] = &dbc
+	h.server.dbConfigs[dbName] = &dbc
```
adamcfraser, line 149: I don't think it should be the responsibility of external callers to know they need to update both internal maps - seems easy to have a case where one is missed.  Maybe an internal function on server that also manages the locking?

rest/server_context.go
```diff
@@ -42,14 +41,38 @@ const DefaultStatsLogFrequencySecs = 60
...
-	lock              sync.RWMutex
-	statsContext      *statsContext
-	HTTPClient        *http.Client
-	replicator        *base.Replicator
-	cpuPprofFileMutex sync.Mutex // Protect cpuPprofFile from concurrent Start and Stop CPU profiling requests
-	cpuPprofFile      *os.File   // An open file descriptor holds the reference during CPU profiling
+	config           *StartupConfig
+	persistentConfig bool
+	// bucketDbConfigs is a map of bucket name to DatabaseConfig
+	bucketDbConfigs map[string]*DatabaseConfig
```
adamcfraser: How is bucketDbConfigs going to be used?  Given that, need to consider future compatibility with databases that target collections or sets of collections instead of buckets.

adamcfraser (changes requested):
Overall approach is solid.  Mainly minor comments, but a few functional questions as well.

base/http_listener.go
```diff
@@ -60,38 +60,30 @@ func ListenAndServeHTTP(addr string, connLimit int, certFile *string, keyFile *s
...
-	}
-	if idleTimeout != nil {
-		server.IdleTimeout = time.Duration(*idleTimeout) * time.Second
-	} else {
-		server.IdleTimeout = DefaultIdleTimeout
+	server := &http.Server{
+		Addr:              addr,
+		Handler:           handler,
+		ReadTimeout:       readTimeout,
+		WriteTimeout:      writeTimeout,
```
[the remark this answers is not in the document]
bbrks: Yep

rest/main_new.go
```diff
@@ -0,0 +1,53 @@
...
+		}
+	}
+
+	if flagStartupConfig != nil {
+		base.Tracef(base.KeyAll, "got config from flags: %#v", flagStartupConfig)
+		err := mergo.Merge(&sc, flagStartupConfig, mergo.WithOverride)
+		if err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
bbrks: It's clearer to me what the state of the system is if default values are included in the /_config response, without working backwards to find what the default values are for a given version/edition. Similiarly, it would benefit customers using the /_config endpoint, as it removes the dependency on docs to find what the default values are for a given option.

At least for the file-based config, we should be capturing the raw startup config file to determine exactly what customers are defining via sgcollect.

rest/config_new.go
bbrks: These are a couple of miscellaneous options for BLIP and REST/BLIP changes feeds but couldn't think of a better name or place for them. Not directly related to sg-replicate/ISGR.

rest/main_new.go
bbrks: Nope not a requirement, and agreed on the corner cases. Will remove.

rest/main_new.go
```diff
@@ -0,0 +1,53 @@
...
+
+	base.Debugf(base.KeyAll, "final config: %#v", sc)
```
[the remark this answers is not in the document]
bbrks: Redacted and switched to trace (will consider removal later in development)

rest/main_new.go
bbrks: Removing until needed.

rest/config_new.go
bbrks: I can see how they could be mixed up, although a misconfguration would be immediately obvious (either bootstrap auth fails, or invalid cert errors via HTTPS)

We could rename the Couchbase Server ones to be `x509_cert_path`/`x509_key_path` to make it super explicit what they're used for?

rest/server_context.go
bbrks: Will remove for now, but will be used for an early-exit opportunity for config updates

rest/config_new.go
bbrks: Moved the Facebook/Google auth options to a "DeprecatedConfig" struct, which cannot be set from a 3.0 config, but can still be used from the legacy config/in-memory upgrade case.

rest/main.go
bbrks: Updated comments to include CBG tickets

rest/server_context.go
bbrks: Forgot I moved the usage of this into #5057. I'll remove from this PR, but CollectionName->DatabaseConfig would be the equivalent once collections are introduced.

rest/admin_api_test.go
```diff
@@ -2372,10 +2372,13 @@ func TestHandlePutDbConfigWithBackticks(t *testing.T) {
 }
 
 func TestHandleDBConfig(t *testing.T) {
+	// FIXME: Broken test with new config - tries connecting to localhost:8091 for new db?
```
bbrks: FIXME

rest/admin_api_test.go
```diff
@@ -2464,15 +2467,18 @@ func TestHandleGetConfig(t *testing.T) {
...
-	assert.Equal(t, "127.0.0.1:4985", respBody["AdminInterface"].(string))
-	databases := respBody["Databases"].(map[string]interface{})
-	db := databases["db"].(map[string]interface{})
-	assert.Equal(t, syncFunc, db["sync"].(string))
+	var respBody StartupConfig
+	assert.NoError(t, base.JSONUnmarshal([]byte(resp.Body.String()), &respBody))
+
+	assert.Equal(t, "127.0.0.1:4985", respBody.API.AdminInterface)
+
+	// FIXME: No dbConfig here in 3.0
```
bbrks: FIXME

rest/admin_api_test.go
```diff
@@ -2718,21 +2724,23 @@ func TestConfigRedaction(t *testing.T) {
...
 	// Test default server config redaction
-	var unmarshaledServerConfig ServerConfig
+	var unmarshaledServerConfig LegacyServerConfig
 	response = rt.SendAdminRequest("GET", "/_config", "")
 	err = json.Unmarshal(response.BodyBytes(), &unmarshaledServerConfig)
 	require.NoError(t, err)
 
-	assert.Equal(t, "xxxxx", unmarshaledServerConfig.Databases["db"].Password)
-	assert.Equal(t, "xxxxx", *unmarshaledServerConfig.Databases["db"].Users["alice"].Password)
+	// FIXME: No dbconfig in response in 3.0
```
bbrks: FIXME

rest/admin_api_test.go
```diff
@@ -2718,21 +2724,23 @@ func TestConfigRedaction(t *testing.T) {
...
+	// assert.Equal(t, "xxxxx", *unmarshaledServerConfig.Databases["db"].Users["alice"].Password)
 
 	// Test default server config redaction when redaction disabled
 	response = rt.SendAdminRequest("GET", "/_config?redact=false", "")
 	err = json.Unmarshal(response.BodyBytes(), &unmarshaledServerConfig)
 	require.NoError(t, err)
 
-	assert.Equal(t, base.TestClusterPassword(), unmarshaledServerConfig.Databases["db"].Password)
-	assert.Equal(t, "password", *unmarshaledServerConfig.Databases["db"].Users["alice"].Password)
+	// FIXME: No dbconfig in response in 3.0
```
bbrks: FIXME

rest/config.go
```diff
@@ -317,31 +224,35 @@ func GetTLSVersionFromString(stringV *string) uint16 {
 
 func (dbConfig *DbConfig) setup(name string) error {
 
+	// FIXME: Inherit from bootstrap config?
```
bbrks: FIXME

rest/main_new.go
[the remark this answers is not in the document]
adamcfraser: I agree from the perspective of seeing the state of the system, but what's the expected workflow if someone wants to retrieve an existing database config, tweak some properties, and then update?  Assuming they pull down the default values, it looks like they are explicitly setting the defaults in the updated version.  Which has the potential to create problems if we change the default value in a future release.

rest/config_new.go
adamcfraser: Ah ok, my mistake.  I tend to think of 'replicator' as something on the active side of replication, but I've spent 5 minutes and haven't come up with a better name, so can leave as-is.

rest/config_new.go
adamcfraser: The changes you've made look fine.

rest/server_context.go
```diff
@@ -264,16 +275,18 @@ func (sc *ServerContext) PostUpgrade(preview bool) (postUpgradeResults PostUpgra
 	return postUpgradeResults, nil
 }
 
+// Removes and re-adds a database to the ServerContext.
+func (sc *ServerContext) _reloadDatabaseFromConfig(reloadDbName string) (*db.DatabaseContext, error) {
+	sc._removeDatabase(reloadDbName)
+	config := sc.dbConfigs[reloadDbName]
+	return sc._getOrAddDatabaseFromConfig(*config, true)
```
adamcfraser, line 282: If `_getOrAddDatabaseFromConfig` returns error, we'll end up with mismatched sets in `sc.databases_` and `sc.dbConfigs`.  Is that ok?  I don't see anywhere we're currently blocking if a config already exists in dbConfigs- we're just overwriting in all cases, I think - but wanted to confirm.

adamcfraser (commented):
Looks good - one question for clarification on the error case in reloadDatabaseFromConfig, and a followup question about whether defaults should be included in _config response.  Might be quicker to just discuss the latter.

rest/server_context.go
bbrks: This is an area that will need a bit more attention in the PRs that add the `/db/_config` REST APIs and background config polling to ensure things are consistent and safe.

rest/main_new.go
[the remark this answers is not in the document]
bbrks: In this case it's the StartupConfig so they can't use the REST API to update, as it's coming from the config file. But I see what you're saying, if users use the REST API to retrieve to write a new config file. I'll have a think about that use-case and see if we're able to cover both bases.

adamcfraser (approved):

PR #5044, opened 2021-06-15
Allow BLIP Tester Client to support setting its supported BLIP protocols

JRascagneres:
Allows the BLIP tester client to specify which BLIP sub protocols it supports. This improves our ability to test V2 and V3 replication behaviour.

### DIFF
--- a/rest/blip_client_test.go
+++ b/rest/blip_client_test.go
@@ -27,10 +27,12 @@ import (
 )
 
 type BlipTesterClientOpts struct {
-	ClientDeltas          bool // Support deltas on the client side
-	Username              string
-	Channels              []string
-	SendRevocations       bool
+	ClientDeltas           bool // Support deltas on the client side
+	Username               string
+	Channels               []string
+	SendRevocations        bool
+	SupportedBLIPProtocols []string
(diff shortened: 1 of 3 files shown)


rest/blip_client_test.go
```diff
@@ -27,10 +27,12 @@ import (
...
 type BlipTesterClientOpts struct {
-	ClientDeltas          bool // Support deltas on the client side
-	Username              string
-	Channels              []string
-	SendRevocations       bool
+	ClientDeltas           bool // Support deltas on the client side
+	Username               string
+	Channels               []string
+	SendRevocations        bool
+	SupportedBLIPProtocols []string
```
adamcfraser, line 34: We already have 'protocol' in place to set the client protocol, should either use that or replace that property with the slice if there's a reason for the test client to specify multiple supported protocols.

adamcfraser (changes requested):
Changes to blip.go to add the new constructor with protocols look fine, but I think we can avoid the duplicate methods to specify protocol in the ClientOpts.

adamcfraser (approved):

PR #5045, opened 2021-06-16
CBG-1473 - Redaction is partial by default

IsaacLambat:
Added unset RedactionLevel which gets changed to RedactPartial (or whatever RedactionLevelDefault) is set to.
~Reordered iota for RedactLevel so RedactPartial is 0, which means if no RedactLevel is set in the logs, RedactPartial is used by default~
RedactUserData is true when a non-redact level number is passed to SetRedaction()
Redact level is set to partial by default in sgcollect_info
Added a test to check redaction level and user data redaction variable is correct

### DIFF
--- a/base/redactor.go
+++ b/base/redactor.go
@@ -104,7 +104,7 @@ func SetRedaction(redactionLevel RedactionLevel) {
 	case RedactNone:
 		RedactUserData = false
 	default:
-		RedactUserData = false
+		RedactUserData = true
 	}
 }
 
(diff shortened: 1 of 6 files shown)


base/redactor.go
```diff
@@ -90,8 +90,8 @@ func buildRedactorFuncSlice(valueOf reflect.Value, function func(interface{}) Re
 type RedactionLevel int
 
 const (
-	RedactNone RedactionLevel = iota
-	RedactPartial
+	RedactPartial RedactionLevel = iota
```
bbrks: As discussed add a RedactUnset value to differentiate between none/unset and set default value in setupAndValidateLogging

bbrks (changes requested):

bbrks (approved):
Looks good to me! 👍

PR #5046, opened 2021-06-16
CBG-1504: Add ability to obtain mgmt endpoints from cluster

JRascagneres:
Add a function on the ServerContext which connects to a Couchbase Server cluster and obtains the management endpoints.

This currently connects with the test values (serverAddress, serverUsername, serverPassword, certPath, keyPath) from a temporary function `tempConnectionDetails`. Once the bootstrap work is complete for persistent config these will be replaced.

Also added a function on DatabaseContext which can be used in the event that we have an existing bucket connection.

Unit test is currently grabs endpoints and attempts to assert against the formatted connection string.

### DIFF
--- a/base/bucket_gocb.go
+++ b/base/bucket_gocb.go
@@ -2037,12 +2037,17 @@ func AsGoCBBucket(bucket Bucket) (*CouchbaseBucketGoCB, bool) {
 	return AsGoCBBucket(underlyingBucket)
 }
 
-// Get one of the management endpoints.  It will be a string such as http://couchbase
-func GoCBBucketMgmtEndpoint(bucket *gocb.Bucket) (url string, err error) {
+func GoCBBucketMgmtEndpoints(bucket *gocb.Bucket) (url []string, err error) {
 	mgmtEps := bucket.IoRouter().MgmtEps()
 	if len(mgmtEps) == 0 {
-		return "", fmt.Errorf("No available Couchbase Server nodes")
+		return nil, fmt.Errorf("No available Couchbase Server nodes")
 	}
+	return mgmtEps, nil
+}
+
+// Get one of the management endpoints.  It will be a string such as http://couchbase
+func GoCBBucketMgmtEndpoint(bucket *gocb.Bucket) (url string, err error) {
+	mgmtEps, err := GoCBBucketMgmtEndpoints(bucket)
 	bucketEp := mgmtEps[rand.Intn(len(mgmtEps))]
 	return bucketEp, nil
 }
--- a/base/bucket_gocb.go
+++ b/base/bucket_gocb.go
@@ -2048,6 +2048,9 @@ func GoCBBucketMgmtEndpoints(bucket *gocb.Bucket) (url []string, err error) {
 // Get one of the management endpoints.  It will be a string such as http://couchbase
 func GoCBBucketMgmtEndpoint(bucket *gocb.Bucket) (url string, err error) {
 	mgmtEps, err := GoCBBucketMgmtEndpoints(bucket)
+	if err != nil {
+		return "", err
+	}
 	bucketEp := mgmtEps[rand.Intn(len(mgmtEps))]
 	return bucketEp, nil
 }
--- a/rest/server_context.go
+++ b/rest/server_context.go
@@ -9,6 +9,7 @@
 package rest
 
 import (
+	"crypto/x509"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -22,6 +23,7 @@ import (
 	"time"
 
 	"github.com/couchbase/go-couchbase"
+	"github.com/couchbase/gocbcore"
 	sgbucket "github.com/couchbase/sg-bucket"
 	"github.com/couchbase/sync_gateway/base"
 	"github.com/couchbase/sync_gateway/db"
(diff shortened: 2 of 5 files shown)


rest/server_context_test.go
```diff
@@ -289,3 +291,27 @@ func TestStatsLoggerStopped(t *testing.T) {
...
+	baseSpec, err := connstr.Parse(clusterAddress)
+	require.NoError(t, err)
+
+	spec, err := connstr.Resolve(baseSpec)
+	require.NoError(t, err)
+
+	for _, httpHost := range spec.HttpHosts {
+		formatted := fmt.Sprintf("http://%s:%d", httpHost.Host, httpHost.Port)
+		assert.Contains(t, eps, formatted)
+	}
```
bbrks: I know this is a test, and works fine in our current setup - but this isn't always true if the bootstrap connstr contains non-existent hosts (which can be the case if a customer scales down their backend).

I'm also not entirely sure there's a direct link between the two, because bootstrap hosts might be hostnames, whereas the cluster could return IP addresses, or vice versa.

rest/server_context.go
```diff
@@ -1147,6 +1149,68 @@ func (sc *ServerContext) updateCalculatedStats() {
...
 
+func initClusterAgent(clusterAddress, clusterUser, clusterPass, certPath, keyPath string) (*gocbcore.Agent, error) {
+	authenticator, err := base.GoCBCoreAuthConfig(clusterUser, clusterPass, certPath, keyPath)
+	if err != nil {
+		return nil, err
+	}
+
+	config := gocbcore.AgentConfig{
+		Auth:           authenticator,
+		AuthMechanisms: []gocbcore.AuthMechanism{gocbcore.ScramSha512AuthMechanism},
```
bbrks: I'm not sure why we need to set this specifically - it looks like gocbcore.createAgent has handling for a zero-value to set some sane defaults?

rest/server_context.go
```diff
@@ -1147,6 +1149,68 @@ func (sc *ServerContext) updateCalculatedStats() {
...
+		TLSRootCAProvider: func() *x509.CertPool {
+			return nil
+		},
```
bbrks: Doesn't this need to be set in order to trust a cert from both a known and unkown signing authority on the server? I think if you return nil here, _no_ cert will be trusted.

rest/server_context.go
```diff
@@ -1147,6 +1149,68 @@ func (sc *ServerContext) updateCalculatedStats() {
...
+		return nil, err
+	}
+
+	agent, err := gocbcore.CreateAgent(&config)
+	if err != nil {
+		return nil, err
+	}
+
+	agentReadyErr := make(chan error)
+	_, err = agent.WaitUntilReady(time.Now().Add(5*time.Second), gocbcore.WaitUntilReadyOptions{}, func(result *gocbcore.WaitUntilReadyResult, err error) {
```
bbrks: 5 seconds seems low enough to be flaky under some conditions (test env?).
Is this a deadline for the internal retry loop, rather than a fixed sleep? If so, I think we can increase this a bit (15 seconds?)

bbrks: We only care about waiting for the `gocbcore.MgmtService`, otherwise it's apparently going to try and wait for everything: `ServiceTypes []ServiceType // Defaults to all services`

bbrks (changes requested):

rest/server_context.go
JRascagneres: Yeah its a deadline. Was just concerned that a HTTP call could just block for 15 seconds but I can go with that.
Will set Mgmt service. I guess it knows whats enabled and whats not then? Given my cluster had only KV and Query but didn't block?

rest/server_context_test.go
JRascagneres: Think the IP / hostname should be handled by resolve. In terms of existing hosts I'll switch to check at least one match.

rest/server_context.go
JRascagneres: I think if I return `nil` it basically skips verify but yeah. Adding this.

rest/server_context_test.go
```diff
@@ -325,3 +325,45 @@ outerLoop:
...
+
+	spec, err := connstr.Resolve(baseSpec)
+	require.NoError(t, err)
+
+	existsOneMatchingEndpoint := false
+
+outerLoop:
+	for _, httpHost := range spec.HttpHosts {
+		for _, ep := range eps {
+			formattedHttpHost := fmt.Sprintf("https://%s:%d", httpHost.Host, httpHost.Port)
```
bbrks, line 360: Is it possible to try and send a whoami request to this endpoint?
I think that would verify whether we need additional work around the tls.Config or not.

If this is already planned for a future PR that also makes sense.

base/bucket_gocb.go
```diff
@@ -2037,12 +2037,17 @@ func AsGoCBBucket(bucket Bucket) (*CouchbaseBucketGoCB, bool) {
...
 	if len(mgmtEps) == 0 {
-		return "", fmt.Errorf("No available Couchbase Server nodes")
+		return nil, fmt.Errorf("No available Couchbase Server nodes")
 	}
+	return mgmtEps, nil
+}
+
+// Get one of the management endpoints.  It will be a string such as http://couchbase
+func GoCBBucketMgmtEndpoint(bucket *gocb.Bucket) (url string, err error) {
+	mgmtEps, err := GoCBBucketMgmtEndpoints(bucket)
```
bbrks, line 2050: missing error check

base/bucket_gocb.go
```diff
@@ -2037,12 +2037,17 @@ func AsGoCBBucket(bucket Bucket) (*CouchbaseBucketGoCB, bool) {
...
 	bucketEp := mgmtEps[rand.Intn(len(mgmtEps))]
```
bbrks, line 2054: will panic when len(mgmtEps) == 0 (seems related to error handling)

bbrks (changes requested):
spotted a couple of things

rest/server_context_test.go
JRascagneres: Yeah actually making a request is planned for the next PR.

base/bucket_gocb.go
JRascagneres: Handled by len check in `GoCBBucketMgmtEndpoints`

JRascagneres:
Split out work to create agent and ObtainManagementEndpoints takes agent as a parameter. This works better as agent is also required to obtain the client transport config.

rest/server_context.go
```diff
@@ -1147,6 +1148,70 @@ func (sc *ServerContext) updateCalculatedStats() {
...
+		Auth:              authenticator,
+		TLSRootCAProvider: tlsRootCAProvider,
+	}
+
+	err = config.FromConnStr(clusterAddress)
+	if err != nil {
+		return nil, err
+	}
+
+	agent, err := gocbcore.CreateAgent(&config)
```
adamcfraser, line 1172: If we have an agent, should we be using agent.DoHTTPRequest instead of working with the management endpoints directly?

rest/server_context.go
```diff
@@ -1147,6 +1148,70 @@ func (sc *ServerContext) updateCalculatedStats() {
...
+	if err != nil {
+		return nil, err
+	}
+
+	agentReadyErr := make(chan error)
+	_, err = agent.WaitUntilReady(time.Now().Add(15*time.Second), gocbcore.WaitUntilReadyOptions{ServiceTypes: []gocbcore.ServiceType{gocbcore.MgmtService}}, func(result *gocbcore.WaitUntilReadyResult, err error) {
```
adamcfraser: Have you tested this against a multi-node server, to validate that when clusterAddress is just a single node, you're getting the full set of endpoints back (i.e. ensure agent.WaitUntilReady is actually waiting on the cluster map to be retrieved)?

rest/server_context.go
```diff
@@ -1147,6 +1148,70 @@ func (sc *ServerContext) updateCalculatedStats() {
 
 }
 
+func initClusterAgent(clusterAddress, clusterUser, clusterPass, certPath, keyPath, caCertPath string) (*gocbcore.Agent, error) {
+	authenticator, err := base.GoCBCoreAuthConfig(clusterUser, clusterPass, certPath, keyPath)
+	if err != nil {
+		return nil, err
+	}
+
+	tlsRootCAProvider, err := base.GoCBCoreTLSRootCAProvider(caCertPath)
```
adamcfraser, line 1157: Are we planning to add an unsupported config for testing with self-signed certs?

adamcfraser (changes requested):

rest/server_context.go
JRascagneres: Yes, have confirmed this

rest/server_context.go
JRascagneres: Ultimately yes but as I understand that was going to be part of CBG-1518?

rest/server_context.go
JRascagneres: We had this discussion in a Sync up about this last week (as I originally played around with DoHTTPRequest in my PoC) and I was fairly sure we came to the conclusion that we wanted to do that on our side so we can explicitly control the operation.

We can more easily control retries and can do things like try another endpoint if the first doesn't work

adamcfraser (dismissed):

adamcfraser (approved):

PR #5049, opened 2021-06-17
CBG-1520 Switch to the latest revision of go-blip library

sarathkumarsivan:
A recent change has been included in the go-blip library store the currently used Websocket subprotocol on BLIP context. The ActiveProtocol function exposed on the BLIP context is no more and we should use ActiveSubprotocol instead.

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -322,8 +322,7 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 
 					// If change is a removal and we're running with protocol V3 and change change is not a tombstone
 					// fall into 3.0 removal handling
-					if change.allRemoved && bh.blipContext.ActiveSubprotocol() == BlipCBMobileReplicationV3 && !change.
-						Deleted {
+					if change.allRemoved && bh.blipContext.ActiveSubprotocol() == BlipCBMobileReplicationV3 && !change.Deleted {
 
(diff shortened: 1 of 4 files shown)


db/blip_handler.go
```diff
@@ -322,7 +322,8 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 
 					// If change is a removal and we're running with protocol V3 and change change is not a tombstone
 					// fall into 3.0 removal handling
-					if change.allRemoved && bh.blipContext.ActiveProtocol() == BlipCBMobileReplicationV3 && !change.Deleted {
+					if change.allRemoved && bh.blipContext.ActiveSubprotocol() == BlipCBMobileReplicationV3 && !change.
+						Deleted {
```
bbrks: This linebreak looks unexpected/weird to me, even if it's valid syntax. We're not too strict on line lengths to warrant breaking it up

```suggestion
					if change.allRemoved && bh.blipContext.ActiveSubprotocol() == BlipCBMobileReplicationV3 && !change.Deleted {
```

bbrks (changes requested):
Minor request about a presumably automated linebreak

bbrks (approved):

db/blip_handler.go
sarathkumarsivan: Makes sense, this was an accidental formatting. Removed in the latest revision.

PR #5054, opened 2021-06-21
CBG-1475 - Added Warning threshold for channel name length

IsaacLambat:
Adds configurable warning threshold for max channel name length.

Warning that is logged: "Doc id: _DOC ID_ channel name size: _1_ channel/s exceeds _250_ characters for channel name size warning threshold"

### DIFF
--- a/base/stats.go
+++ b/base/stats.go
@@ -232,31 +232,32 @@ type CBLReplicationPushStats struct {
 }
 
 type DatabaseStats struct {
-	ConflictWriteCount      *SgwIntStat `json:"conflict_write_count"`
-	Crc32MatchCount         *SgwIntStat `json:"crc32c_match_count"`
-	DCPCachingCount         *SgwIntStat `json:"dcp_caching_count"`
-	DCPCachingTime          *SgwIntStat `json:"dcp_caching_time"`
-	DCPReceivedCount        *SgwIntStat `json:"dcp_received_count"`
-	DCPReceivedTime         *SgwIntStat `json:"dcp_received_time"`
-	DocReadsBytesBlip       *SgwIntStat `json:"doc_reads_bytes_blip"`
-	DocWritesBytes          *SgwIntStat `json:"doc_writes_bytes"`
-	DocWritesBytesBlip      *SgwIntStat `json:"doc_writes_bytes_blip"`
-	DocWritesXattrBytes     *SgwIntStat `json:"doc_writes_xattr_bytes"`
-	HighSeqFeed             *SgwIntStat `json:"high_seq_feed"`
-	NumDocReadsBlip         *SgwIntStat `json:"num_doc_reads_blip"`
-	NumDocReadsRest         *SgwIntStat `json:"num_doc_reads_rest"`
-	NumDocWrites            *SgwIntStat `json:"num_doc_writes"`
-	NumReplicationsActive   *SgwIntStat `json:"num_replications_active"`
-	NumReplicationsTotal    *SgwIntStat `json:"num_replications_total"`
-	NumTombstonesCompacted  *SgwIntStat `json:"num_tombstones_compacted"`
-	SequenceAssignedCount   *SgwIntStat `json:"sequence_assigned_count"`
-	SequenceGetCount        *SgwIntStat `json:"sequence_get_count"`
-	SequenceIncrCount       *SgwIntStat `json:"sequence_incr_count"`
-	SequenceReleasedCount   *SgwIntStat `json:"sequence_released_count"`
-	SequenceReservedCount   *SgwIntStat `json:"sequence_reserved_count"`
-	WarnChannelsPerDocCount *SgwIntStat `json:"warn_channels_per_doc_count"`
-	WarnGrantsPerDocCount   *SgwIntStat `json:"warn_grants_per_doc_count"`
-	WarnXattrSizeCount      *SgwIntStat `json:"warn_xattr_size_count"`
+	ConflictWriteCount       *SgwIntStat `json:"conflict_write_count"`
+	Crc32MatchCount          *SgwIntStat `json:"crc32c_match_count"`
+	DCPCachingCount          *SgwIntStat `json:"dcp_caching_count"`
+	DCPCachingTime           *SgwIntStat `json:"dcp_caching_time"`
+	DCPReceivedCount         *SgwIntStat `json:"dcp_received_count"`
+	DCPReceivedTime          *SgwIntStat `json:"dcp_received_time"`
+	DocReadsBytesBlip        *SgwIntStat `json:"doc_reads_bytes_blip"`
+	DocWritesBytes           *SgwIntStat `json:"doc_writes_bytes"`
+	DocWritesBytesBlip       *SgwIntStat `json:"doc_writes_bytes_blip"`
+	DocWritesXattrBytes      *SgwIntStat `json:"doc_writes_xattr_bytes"`
+	HighSeqFeed              *SgwIntStat `json:"high_seq_feed"`
+	NumDocReadsBlip          *SgwIntStat `json:"num_doc_reads_blip"`
+	NumDocReadsRest          *SgwIntStat `json:"num_doc_reads_rest"`
+	NumDocWrites             *SgwIntStat `json:"num_doc_writes"`
+	NumReplicationsActive    *SgwIntStat `json:"num_replications_active"`
+	NumReplicationsTotal     *SgwIntStat `json:"num_replications_total"`
+	NumTombstonesCompacted   *SgwIntStat `json:"num_tombstones_compacted"`
+	SequenceAssignedCount    *SgwIntStat `json:"sequence_assigned_count"`
+	SequenceGetCount         *SgwIntStat `json:"sequence_get_count"`
+	SequenceIncrCount        *SgwIntStat `json:"sequence_incr_count"`
+	SequenceReleasedCount    *SgwIntStat `json:"sequence_released_count"`
+	SequenceReservedCount    *SgwIntStat `json:"sequence_reserved_count"`
+	WarnChannelsPerDocCount  *SgwIntStat `json:"warn_channels_per_doc_count"`
+	WarnGrantsPerDocCount    *SgwIntStat `json:"warn_grants_per_doc_count"`
+	WarnXattrSizeCount       *SgwIntStat `json:"warn_xattr_size_count"`
+	WarnChannelNameSizeCount *SgwIntStat `json:"warn_channel_name_size_count"`
 
 	// These can be cleaned up in future versions of SGW, implemented as maps to reduce amount of potential risk
 	// prior to Hydrogen release. These are not exported as part of prometheus and only exposed through expvars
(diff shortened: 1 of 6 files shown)


base/stats.go
```diff
@@ -232,31 +232,32 @@ type CBLReplicationPushStats struct {
...
+	NumTombstonesCompacted   *SgwIntStat `json:"num_tombstones_compacted"`
+	SequenceAssignedCount    *SgwIntStat `json:"sequence_assigned_count"`
+	SequenceGetCount         *SgwIntStat `json:"sequence_get_count"`
+	SequenceIncrCount        *SgwIntStat `json:"sequence_incr_count"`
+	SequenceReleasedCount    *SgwIntStat `json:"sequence_released_count"`
+	SequenceReservedCount    *SgwIntStat `json:"sequence_reserved_count"`
+	WarnChannelsPerDocCount  *SgwIntStat `json:"warn_channels_per_doc_count"`
+	WarnGrantsPerDocCount    *SgwIntStat `json:"warn_grants_per_doc_count"`
+	WarnXattrSizeCount       *SgwIntStat `json:"warn_xattr_size_count"`
+	WarnChannelNameSizeCount *SgwIntStat `json:"warn_channel_name_size_count"`
```
JRascagneres, line 260: Minor comment but might be nice if we can keep these alphabetically ordered, given the number of stats it makes it easier to scan over

JRascagneres: Minor comment but might be nice if we can keep these alphabetically ordered, given the number of stats it makes it easier to scan over

base/stats.go
```diff
@@ -700,33 +701,34 @@ func (d *DbStats) initDatabaseStats() {
...
+		SequenceGetCount:         NewIntStat(SubsystemDatabaseKey, "sequence_get_count", labelKeys, labelVals, prometheus.CounterValue, 0),
+		SequenceIncrCount:        NewIntStat(SubsystemDatabaseKey, "sequence_incr_count", labelKeys, labelVals, prometheus.CounterValue, 0),
+		SequenceReleasedCount:    NewIntStat(SubsystemDatabaseKey, "sequence_released_count", labelKeys, labelVals, prometheus.CounterValue, 0),
+		SequenceReservedCount:    NewIntStat(SubsystemDatabaseKey, "sequence_reserved_count", labelKeys, labelVals, prometheus.CounterValue, 0),
+		WarnChannelsPerDocCount:  NewIntStat(SubsystemDatabaseKey, "warn_channels_per_doc_count", labelKeys, labelVals, prometheus.CounterValue, 0),
+		WarnGrantsPerDocCount:    NewIntStat(SubsystemDatabaseKey, "warn_grants_per_doc_count", labelKeys, labelVals, prometheus.CounterValue, 0),
+		WarnXattrSizeCount:       NewIntStat(SubsystemDatabaseKey, "warn_xattr_size_count", labelKeys, labelVals, prometheus.CounterValue, 0),
+		WarnChannelNameSizeCount: NewIntStat(SubsystemDatabaseKey, "warn_channel_name_size_count", labelKeys, labelVals, prometheus.CounterValue, 0),
```
JRascagneres: Same with ordering

db/crud.go
```diff
@@ -1921,6 +1921,20 @@ func (db *Database) checkDocChannelsAndGrantsLimits(docID string, channels base.
...
+	if channelNameSizeThreshold := db.Options.UnsupportedOptions.WarningThresholds.ChannelNameSize; channelNameSizeThreshold != nil {
+		var channelsOverCount int
+		for c := range channels {
+			if uint32(len(c)) > *channelNameSizeThreshold {
+				channelsOverCount++
+			}
+		}
+		if channelsOverCount > 0 {
+			db.DbStats.Database().WarnChannelNameSizeCount.Add(1)
+			base.WarnfCtx(db.Ctx, "Doc id: %v channel name size: %d channel/s exceeds %d characters for channel name size warning threshold", base.UD(docID), channelsOverCount, *channelNameSizeThreshold)
```
JRascagneres: I think it'd be nice to include the channel name / names that exceed the limit so a customer doesn't have to go and inspect the doc / xattrs and count themselves.

Hwoever, also thinking that potentially logging multiple > 250 char strings in a long messages might be a bit excessive. Maybe have a play around with it? Perhaps we could even log once per channel thats too large inside the loop?

rest/admin_api_test.go
```diff
@@ -33,6 +33,121 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+// Test warnings being issued when a new channel is created with over 250 characters - CBG-1475
+func TestChannelNameSizeWarning(t *testing.T) {
```
JRascagneres: This test is fairly hard to read. There is a lot going on. 
Couple comments below on this.

rest/admin_api_test.go
```diff
@@ -33,6 +33,121 @@ import (
...
+			expectWarn:         false,
+		},
+		{
+			name:               "Under max channel length",
+			channelLengthToAdd: -1,
+			expectWarn:         false,
+		},
+	}
+
+	boundaryTest := func(warnThresholdLength int) {
```
JRascagneres: Generally we try to avoid inline functions unless there is a really good reason for them as they make things harder to read. Especially as we're running the test cases  inside of it.

rest/admin_api_test.go
```diff
@@ -33,6 +33,121 @@ import (
...
+		before := getChannelNameWarnCount()
+		tr = rt.SendAdminRequest("DELETE", "/db/deleteme?rev="+getRespRev(t, tr), "")
+		assertStatus(t, tr, http.StatusOK)
+		after := getChannelNameWarnCount()
+		assert.Equal(t, before, after)
+	})
+
+	// change value to 500 in config
+	warnThreshold := uint32(500)
+	rt = NewRestTester(t, &RestTesterConfig{
```
JRascagneres: I think this is going to create a new rest tester with the new config rather than just modify the config like you want. 
I think you should be able to just set it with `rt.DatabaseConfig.Unsupported.WarningThresholds`....

rest/admin_api_test.go
```diff
@@ -33,6 +33,121 @@ import (
...
+				}
+			})
+		}
+	}
+
+	boundaryTest(int(base.DefaultWarnThresholdChannelNameSize))
+
+	// Update doc - should warn
+	channelLength := int(base.DefaultWarnThresholdChannelNameSize) + 5
+	t.Run("Update doc without changing channel", func(t *testing.T) {
```
JRascagneres: I think all these subtests make it difficult to read.

Are they independent of whats going on in the testCases loop?

If so:
- I'd prefer them as separate functions, or at least in another Test function that doesn't have the test case loop so its easier to immediately see each are standalone.

If not:
- Move them inside of the test case loop.

rest/admin_api_test.go
```diff
@@ -33,6 +33,121 @@ import (
...
+		DatabaseConfig: &DbConfig{
+			Unsupported: db.UnsupportedOptions{
+				WarningThresholds: db.WarningThresholds{
+					ChannelNameSize: &warnThreshold,
+				},
+			},
+		},
+	})
+	defer rt.Close()
+	boundaryTest(int(warnThreshold))
```
JRascagneres: I think here you're increasing the threshold count and checking the limits again?

Could this be made part of the sub tests? If not its mostly fine like this but would like to avoid the inline func if possible

JRascagneres (changes requested):
Actual implementation looks pretty good. Just got some comments on readability of the test cases really.

rest/admin_api_test.go
IsaacLambat: Had problems getting this modification to work as for some reason it wasn't recognising the change through the "db.Options.UnsupportedOptions.WarningThresholds.ChannelNameSize" line in crud.go.

I have made it so a new rest tester is made per loop iteration which might be quite intensive to do, but I think it will be a better way to test it as, I assume, the user will be setting the Warning Threshold in the config before SG starts. 

Let me know if that seems like an ok way to do it or if there's something I am missing

rest/admin_api_test.go
```diff
@@ -2966,3 +2976,142 @@ func TestObtainUserChannelsForDeletedRoleCasFail(t *testing.T) {
...
+				SyncFn: syncFn,
+				DatabaseConfig: &DbConfig{
+					Unsupported: db.UnsupportedOptions{
+						WarningThresholds: thresholdConfig,
+					},
+				},
+			})
+			defer rt.Close()
+
+			chanNameWarnCountBefore := rt.ServerContext().Database("db").DbStats.Database().WarnChannelNameSizeCount.Val
```
JRascagneres, line 3046: Should use .Value() function for stats.

JRascagneres: Should use .Value() function for stats, couple places to swap out

rest/admin_api_test.go
```diff
@@ -2966,3 +2976,142 @@ func TestObtainUserChannelsForDeletedRoleCasFail(t *testing.T) {
...
+	syncFn := "function sync(doc, oldDoc) { channel(doc.chan); }"
+	rt := NewRestTester(t, &RestTesterConfig{
+		SyncFn: syncFn,
+	})
+	defer rt.Close()
+
+	// Update doc - should warn
+	channelLength := int(base.DefaultWarnThresholdChannelNameSize) + 5
+	t.Run("Update doc without changing channel", func(t *testing.T) {
+		tr := rt.SendAdminRequest("PUT", "/db/replace", fmt.Sprintf("{\"chan\":\"%s\"}", strings.Repeat("B", channelLength))) // init doc
```
JRascagneres: (nitpick) With it being the same channel name can we assign a variable here for the channel?

JRascagneres (dismissed):
Couple small things but otherwise looks good to me now. Much easier to read tests.

rest/admin_api_test.go
```diff
@@ -2966,3 +2976,142 @@ func TestObtainUserChannelsForDeletedRoleCasFail(t *testing.T) {
...
+				DatabaseConfig: &DbConfig{
+					Unsupported: db.UnsupportedOptions{
+						WarningThresholds: thresholdConfig,
+					},
+				},
+			})
+			defer rt.Close()
+
+			chanNameWarnCountBefore := rt.ServerContext().Database("db").DbStats.Database().WarnChannelNameSizeCount.Value()
+			tr := rt.SendAdminRequest("PUT", fmt.Sprintf("/db/doc%v", test.channelLength), fmt.Sprintf("{\"chan\":\"%s\"}", strings.Repeat("A", test.channelLength)))
```
bbrks: Lots going on everywhere we're sending a request - which is difficult to glance over quickly.

I'd pull out each bit of the request into variables (which can be useful for sticking into assertion messages in some more complex test cases, good habit to get into).
You can also use backticks helps with the JSON by avoiding the escaped quotes.

```suggestion
			docID := fmt.Sprintf("/db/doc%v", test.channelLength)
			chanName := strings.Repeat("A", test.channelLength)
			tr := rt.SendAdminRequest("PUT", docID, `{"chan":"`+chanName`"}`)
```

bbrks (changes requested):
Need to go through all usages of `rt.SendAdminRequest` and make sure we're only doing one or two things per line.
The lines as they are are on the long side, and generally doing 3 or 4 things per line, which makes it really hard to glance over without actually reading.

Doc bodies like `{"foo":"+bar+"}` are reasonable to inline, but less so the other stuff.
In addition, having `docID`/`revID` variables can be useful for putting into assertion messages for more complex test cases.

bbrks (dismissed):

bbrks (approved):

PR #5055, opened 2021-06-21
CBG-1505: Specify permissions on handler and check

JRascagneres:
PR aims to check permissions that are specified by the handler against a user specified user on CBS.
Also adds the config options to enable / disable the feature.

Currently no permissions are specified on the handlers and so the check inside the handler is temporarily disabled.

Builds on existing agent / management endpoint work.

Has unit tests checking whether we can verify a users permissions:
- One with multiple test cases
- One small X509 which just verifies we can check one permission to verify X509 cert use

Based off of https://github.com/couchbase/sync_gateway/pull/5046
- [x] Merge https://github.com/couchbase/sync_gateway/pull/5046 & rebase

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -1381,6 +1381,25 @@ func (context *DatabaseContext) ObtainManagementEndpoints() ([]string, error) {
 	return base.GoCBBucketMgmtEndpoints(gocbBucket.Bucket)
 }
 
+func (context *DatabaseContext) ObtainManagementEndpointsAndHTTPClient() ([]string, *http.Client, error) {
+	gocbBucket, ok := base.AsGoCBBucket(context.Bucket)
+	if !ok {
+		base.Warnf("Database %v: Unable to get server management endpoints. Underlying bucket type was not GoCBBucket.", base.MD(context.Name))
+		return nil, nil, nil
+	}
+
+	endpoints, err := base.GoCBBucketMgmtEndpoints(gocbBucket.Bucket)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	httpClient := &http.Client{
+		Transport: gocbBucket.Bucket.IoRouter().HttpClient().Transport,
+	}
+
+	return endpoints, httpClient, nil
+}
+
 func (context *DatabaseContext) GetUserViewsEnabled() bool {
 	if context.Options.UnsupportedOptions.UserViews.Enabled != nil {
 		return *context.Options.UnsupportedOptions.UserViews.Enabled
--- a/db/database.go
+++ b/db/database.go
@@ -1393,8 +1393,11 @@ func (context *DatabaseContext) ObtainManagementEndpointsAndHTTPClient() ([]stri
 		return nil, nil, err
 	}
 
+	// Clone HTTP transport for use in our httpClient
+	transport := gocbBucket.Bucket.IoRouter().HttpClient().Transport.(*http.Transport).Clone()
+
 	httpClient := &http.Client{
-		Transport: gocbBucket.Bucket.IoRouter().HttpClient().Transport,
+		Transport: transport,
 	}
 
 	return endpoints, httpClient, nil
--- a/rest/handler.go
+++ b/rest/handler.go
@@ -89,22 +89,22 @@ const (
 type handlerMethod func(*handler) error
 
 // Creates an http.Handler that will run a handler with the given method
-func makeHandler(server *ServerContext, privs handlerPrivs, checkPermissions []string, storePermissionsResults bool, method handlerMethod) http.Handler {
+func makeHandler(server *ServerContext, privs handlerPrivs, accessPermissions []string, responsePermissions []string, method handlerMethod) http.Handler {
 	return http.HandlerFunc(func(r http.ResponseWriter, rq *http.Request) {
 		runOffline := false
 		h := newHandler(server, privs, r, rq, runOffline)
-		err := h.invoke(method, checkPermissions, storePermissionsResults)
+		err := h.invoke(method, accessPermissions, responsePermissions)
 		h.writeError(err)
 		h.logDuration(true)
 	})
 }
 
 // Creates an http.Handler that will run a handler with the given method even if the target DB is offline
-func makeOfflineHandler(server *ServerContext, privs handlerPrivs, checkPermissions []string, storePermissionsResults bool, method handlerMethod) http.Handler {
+func makeOfflineHandler(server *ServerContext, privs handlerPrivs, accessPermissions []string, responsePermissions []string, method handlerMethod) http.Handler {
 	return http.HandlerFunc(func(r http.ResponseWriter, rq *http.Request) {
 		runOffline := true
 		h := newHandler(server, privs, r, rq, runOffline)
-		err := h.invoke(method, checkPermissions, storePermissionsResults)
+		err := h.invoke(method, accessPermissions, responsePermissions)
 		h.writeError(err)
 		h.logDuration(true)
 	})
@@ -124,7 +124,7 @@ func newHandler(server *ServerContext, privs handlerPrivs, r http.ResponseWriter
 }
 
 // Top-level handler call. It's passed a pointer to the specific method to run.
-func (h *handler) invoke(method handlerMethod, checkPermissions []string, storePermissionsResults bool) error {
+func (h *handler) invoke(method handlerMethod, accessPermissions []string, responsePermissions []string) error {
 
 	var err error
 	if h.server.config.CompressResponses == nil || *h.server.config.CompressResponses {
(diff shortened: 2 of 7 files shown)


rest/server_context.go
```diff
@@ -1200,21 +1203,112 @@ var tempConnectionDetailsForManagementEndpoints = func() (serverAddress string,
...
 	err = agent.Close()
 	if err != nil {
-		return nil, err
+		return nil, nil, err
+	}
+
+	return managementEndpoints, httpClient, nil
+}
+
+func (sc *ServerContext) CheckPermissions(httpClient *http.Client, managementEndpoints []string, username, password string, requestedPermissions []string) (statusCode int, permissionResults map[string]bool, err error) {
```
bbrks: Can we rename this? And probably include a comment for more detail. It does a poor job at conveying what the function is doing.

"Check" is fairly ambiguous:
- Is it checking there are ANY permissions present?
- Is it checking that the user has ALL of the requested permissions?

I think the answer to both of those is no. But I had to read the code to figure that out :)

rest/server_context.go
```diff
@@ -1200,21 +1203,112 @@ var tempConnectionDetailsForManagementEndpoints = func() (serverAddress string,
...
+		}
+
+		req.SetBasicAuth(username, password)
+
+		httpResponse, err = httpClient.Do(req)
+
+		// If no error on request break out and continue
+		if err == nil {
+			break
+		}
```
bbrks: Is there a reason why we're not using the normal retry loop helpers?

rest/handler.go
```diff
@@ -387,6 +400,40 @@ func (h *handler) checkAuth(context *db.DatabaseContext) (err error) {
...
+		if err != nil {
+			return base.HTTPErrorf(http.StatusInternalServerError, "")
+		}
+	}
+
+	username, password := h.getBasicAuth()
+	statusCode, permResults, err := h.server.CheckPermissions(httpClient, managementEndpoints, username, password, checkPermissions)
+	if err != nil {
+		return base.HTTPErrorf(http.StatusInternalServerError, "")
+	}
```
bbrks: At what point do we actually get information about any errors?

None of this code is doing any logging at all - which is OK if the error is being bubbled up enough to where it is being logged, but it doesn't look like that's happening.

rest/handler.go
```diff
@@ -82,27 +83,28 @@ const (
...
 	publicPrivs         // Handler Handler checks auth and falls back to guest if invalid or missing
 	adminPrivs          // Handler ignores auth, always runs with root/admin privs
+	metricsPrivs
 )
 
 type handlerMethod func(*handler) error
 
 // Creates an http.Handler that will run a handler with the given method
-func makeHandler(server *ServerContext, privs handlerPrivs, method handlerMethod) http.Handler {
+func makeHandler(server *ServerContext, privs handlerPrivs, checkPermissions []string, storePermissionsResults bool, method handlerMethod) http.Handler {
```
bbrks: What does `storePermissionResults` do and how do I know whether I want to set it or not for a given route? I can't tell what the `permissionsResults` map is used for...

rest/handler.go
```diff
@@ -201,6 +203,17 @@ func (h *handler) invoke(method handlerMethod) error {
 		}
 	}
 
+	// If an Admin Request we need to check the user credentials
+	// FIXME: Temporarily disable this
+	if false {
+		clusterAddress, _, _, _, _, _ := tempConnectionDetailsForManagementEndpoints()
```
bbrks: Why? Is it because we don't want to break usages of the Admin API? At what point is it safe to enable this permanently?

rest/routing.go
```diff
@@ -40,74 +40,68 @@ func createHandler(sc *ServerContext, privs handlerPrivs) (*mux.Router, *mux.Rou
...
-	dbr.Handle("/_design/{ddoc}", makeHandler(sc, privs, (*handler).handlePutDesignDoc)).Methods("PUT")
-	dbr.Handle("/_design/{ddoc}", makeHandler(sc, privs, (*handler).handleDeleteDesignDoc)).Methods("DELETE")
-	dbr.Handle("/_design/{ddoc}/_view/{view}", makeHandler(sc, privs, (*handler).handleView)).Methods("GET")
-	dbr.Handle("/_ensure_full_commit", makeHandler(sc, privs, (*handler).handleEFC)).Methods("POST")
-	dbr.Handle("/_revs_diff", makeHandler(sc, privs, (*handler).handleRevsDiff)).Methods("POST")
+	dbr.Handle("/_all_docs", makeHandler(sc, privs, nil, false, (*handler).handleAllDocs)).Methods("GET", "HEAD", "POST")
+	dbr.Handle("/_bulk_docs", makeHandler(sc, privs, nil, false, (*handler).handleBulkDocs)).Methods("POST")
+	dbr.Handle("/_bulk_get", makeHandler(sc, privs, nil, false, (*handler).handleBulkGet)).Methods("POST")
+	dbr.Handle("/_changes", makeHandler(sc, privs, nil, false, (*handler).handleChanges)).Methods("GET", "HEAD", "POST")
+	dbr.Handle("/_design/{ddoc}", makeHandler(sc, privs, nil, false, (*handler).handleGetDesignDoc)).Methods("GET", "HEAD")
```
bbrks: Do you see a use-case for permission checking on the public API too? Or is this just because the admin API uses the same methods?

bbrks (changes requested):
see inline comments

rest/server_context.go
JRascagneres: Yeah fair do's will add a comment and think of a name

rest/server_context.go
JRascagneres: Guess not, can look into that

rest/handler.go
JRascagneres: `storePermissionResults` defines whether to store the results of the permissions check on the handler. This will allow handlers to figure out what to do based on any permissions it has. For example, the GET _user will do different things based on different permissions passed in. Although given the cost I might just save results for all of them.

rest/handler.go
JRascagneres: Two reasons:
- Don't want to break usages of admin API at the moment
- Haven't yet defined the list of permissions for each endpoint

I'm confident the testing shows this works so when I go back and add permissions for each endpoint I can enable this along with (probably) some default for tests setting the override config

rest/routing.go
JRascagneres: Just because it uses the same `makeHandler` method. Might be possible to split this out

db/database.go
```diff
@@ -1380,6 +1380,25 @@ func (context *DatabaseContext) ObtainManagementEndpoints() ([]string, error) {
...
+		return nil, nil, nil
+	}
+
+	endpoints, err := base.GoCBBucketMgmtEndpoints(gocbBucket.Bucket)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	httpClient := &http.Client{
+		Transport: gocbBucket.Bucket.IoRouter().HttpClient().Transport,
```
adamcfraser: I think this is going to return a mutable reference to that bucket's transport?  Do we want to make a copy instead, to avoid callers of this method inadvertently modifying the underlying bucket client's transport?

rest/handler.go
adamcfraser: As discussed, may want to differentiate between endpoint access permissions, and endpoint response permissions, as it's only the latter we need to store.

rest/handler.go
adamcfraser: I'd also prefer not to have 'if false' in this way if it's going into master and not a dev branch.  Even having a 'check permissions' function that's stubbed out and always returns true would be preferred I think - at least then it's clear how this code is expected to interact with the permissions set.

rest/handler.go
```diff
@@ -201,6 +203,17 @@ func (h *handler) invoke(method handlerMethod) error {
...
+		if !base.ServerIsWalrus(clusterAddress) && ((h.privs == adminPrivs && *h.server.config.AdminInterfaceAuthentication) || (h.privs == metricsPrivs && *h.server.config.MetricsInterfaceAuthentication)) {
```
adamcfraser: It doesn't feel right to have a check for walrus here - it's more like we should default the InterfaceAuthentication config options to false when loading the bootstrap config.

rest/handler.go
adamcfraser: Agreed that we don't want to expose to end user, but definitely want logging on security failure (and possibly success).

adamcfraser (changes requested):
Agree with Ben's feedback, added a couple of additional points.

rest/server_context.go
JRascagneres: Couldn't think of a better name for now but has a fairly detailed comment.

rest/handler.go
JRascagneres: Okay moved the 'if' work into `shouldCheckAdminAuth` for now.

rest/handler.go
JRascagneres: Agreed but currently there is no way to nicely do this. Talked this through with Ben a little and he has something similar doing an initial check for walrus with his persistent config work so I'll probably want to leave this temporary check in `shouldCheckAdminAuth` for now and this can be replaced later. Might add this to the `Update tempConnectionDetails() to use bootstrap credentials on ServerContext` ticket so we remember to do this if we're happy keeping it in here.

db/database.go
JRascagneres: Yeah, thats true. Can utilize the http.Transport Clone() method here.

rest/server_context.go
JRascagneres: Done

rest/handler.go
JRascagneres: Opted to bubble up errors to where we call `checkAdminAuth` and at that point we do logging. In the case of an error log the error, in the case of a not-OK status code, log the failure and return status code to user otherwise log successful login.

JRascagneres:
Had to do force push to fix conflicts apologies for any extra review difficulty.

adamcfraser (approved):

PR #5057, opened 2021-06-22
CBG-1457 - Load existing database configs on bootstrap

bbrks:
Discovers and loads matching group ID database configs in buckets on startup.

- [x] Depends on #5043

### DIFF
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -75,7 +75,7 @@ func (cc *CouchbaseCluster) GetConfigLocations() ([]string, error) {
 		return nil, err
 	}
 
-	bucketList := make([]string, len(buckets))
+	bucketList := make([]string, 0, len(buckets))
 	for bucketName := range buckets {
 		bucketList = append(bucketList, bucketName)
 	}
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -1,6 +1,7 @@
 package base
 
 import (
+	"errors"
 	"time"
 
 	"github.com/couchbase/gocb"
@@ -8,7 +9,11 @@ import (
 
 // BootstrapConnection is the interface that can be used to bootstrap Sync Gateway against a Couchbase Server cluster.
 type BootstrapConnection interface {
-	// TODO: CBG-1457 - Method(s) to retrieve configs from server (consider Testing/Walrus implementation too)
+	// GetConfigLocations returns a list of location names where a database config could belong (e.g. Bucket names, Collection names)
+	GetConfigLocations() ([]string, error)
+	// GetConfig fetches a database config for a given location and config group ID, along with the CAS of the document.
+	GetConfig(location, groupID string, valuePtr interface{}) (cas uint64, err error)
+	// Close closes the connection
 	Close() error
 }
 
@@ -58,6 +63,41 @@ func NewCouchbaseCluster(server, username, password,
 	}
 
 	return &CouchbaseCluster{c: cluster}, nil
+}
+
+func (cc *CouchbaseCluster) GetConfigLocations() ([]string, error) {
+	if cc == nil {
+		return nil, errors.New("nil CouchbaseCluster")
+	}
+
+	buckets, err := cc.c.Buckets().GetAllBuckets(nil)
+	if err != nil {
+		return nil, err
+	}
+
+	bucketList := make([]string, len(buckets))
+	for bucketName := range buckets {
+		bucketList = append(bucketList, bucketName)
+	}
+
+	return bucketList, nil
+}
+
+func (cc *CouchbaseCluster) GetConfig(location, groupID string, valuePtr interface{}) (cas uint64, err error) {
+	if cc == nil {
+		return 0, errors.New("nil CouchbaseCluster")
+	}
+
+	res, err := cc.c.Bucket(location).DefaultCollection().Get(PersistentConfigPrefix+groupID, nil)
+	if err != nil {
+		return 0, err
+	}
+	err = res.Content(valuePtr)
+	if err != nil {
+		return 0, err
+	}
+
+	return uint64(res.Cas()), nil
 }
 
 func (cc *CouchbaseCluster) Close() error {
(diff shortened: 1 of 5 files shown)


rest/config.go
```diff
@@ -827,13 +828,75 @@ func setupServerContext(config *StartupConfig, persistentConfig bool) (*ServerCo
...
 
+func (sc *ServerContext) fetchConfigs() error {
+	locations, err := sc.bootstrapConnection.GetConfigLocations()
+	if err != nil {
+		return fmt.Errorf("couldn't get buckets from cluster: %w", err)
+	}
+
+	for _, location := range locations {
+		base.Tracef(base.KeyAll, "Checking %q for Sync Gateway config in group %q", location, sc.config.Bootstrap.ConfigGroupID)
+		// populate values from bootstrap, before we overwrite them with any values fetched from the bucket.
```
adamcfraser: What happens in the situation where the bootstrap is using basic auth, and the db-specific is using x509 (or vice versa)?  Will the bootstrap values be zeroed out correctly?

base/bootstrap.go
```diff
@@ -59,6 +64,41 @@ func NewCouchbaseCluster(server, username, password,
...
+	if cc == nil {
+		return nil, errors.New("nil CouchbaseCluster")
+	}
+
+	buckets, err := cc.c.Buckets().GetAllBuckets(nil)
+	if err != nil {
+		return nil, err
+	}
+
+	bucketList := make([]string, len(buckets))
```
adamcfraser:
```suggestion
	bucketList := make([]string, 0, len(buckets))
```

base/bootstrap.go
```diff
@@ -59,6 +64,41 @@ func NewCouchbaseCluster(server, username, password,
...
+
+	return bucketList, nil
+}
+
+func (cc *CouchbaseCluster) GetConfig(location, groupID string, valuePtr interface{}) (cas uint64, err error) {
+	if cc == nil {
+		return 0, errors.New("nil CouchbaseCluster")
+	}
+
+	res, err := cc.c.Bucket(location).DefaultCollection().Get(PersistentConfigPrefix+groupID, nil)
```
adamcfraser: Do we want a bucket.WaitUntilReady after calling Bucket(location) to fail fast on a gocb bootstrapError?  It looks like we'd get back any potential underlying bootstrapError on the Get, when gocb does getKvProvider, but I'm not sure if we're doing unnecessary work by waiting until that point.

base/bootstrap.go
```diff
@@ -59,6 +64,41 @@ func NewCouchbaseCluster(server, username, password,
...
+	if err != nil {
```
adamcfraser, line 105: For other gocb functionality I've been checking for expected errors (e.g. not found) at this level, and returning the SG ErrNotFound to avoid leaking gocb imports out of base.

base/bootstrap.go
```diff
@@ -59,6 +64,41 @@ func NewCouchbaseCluster(server, username, password,
...
+
+	bucketList := make([]string, len(buckets))
+	for bucketName := range buckets {
+		bucketList = append(bucketList, bucketName)
+	}
+
+	return bucketList, nil
+}
+
+func (cc *CouchbaseCluster) GetConfig(location, groupID string, valuePtr interface{}) (cas uint64, err error) {
```
adamcfraser, line 86: In the current implementation location is just bucketName.  Might be clearer to leave that explicit here for now, as this function will need rework anyway if location is changed to [bucket,scope,collection].

rest/config.go
```diff
@@ -827,13 +828,75 @@ func setupServerContext(config *StartupConfig, persistentConfig bool) (*ServerCo
...
+				Username:   sc.config.Bootstrap.Username,
+				Password:   sc.config.Bootstrap.Password,
+				CertPath:   sc.config.Bootstrap.X509CertPath,
+				KeyPath:    sc.config.Bootstrap.X509KeyPath,
+				CACertPath: sc.config.Bootstrap.CACertPath,
+			},
+		}
+
+		cas, err := sc.bootstrapConnection.GetConfig(location, sc.config.Bootstrap.ConfigGroupID, &cnf)
+		if errors.As(err, &gocb.ErrDocumentNotFound) {
```
adamcfraser: As mentioned, would prefer to keep the gocb-specific error checking in base.

rest/config.go
```diff
@@ -827,13 +828,75 @@ func setupServerContext(config *StartupConfig, persistentConfig bool) (*ServerCo
...
+			base.Debugf(base.KeyAll, "%q did not contain config in group %q", location, sc.config.Bootstrap.ConfigGroupID)
+			continue
+		}
+		if err != nil {
+			return fmt.Errorf("couldn't fetch config in group %q from location %q: %w", sc.config.Bootstrap.ConfigGroupID, location, err)
+		}
+
+		base.Tracef(base.KeyAll, "Got config for location %q with cas %d", location, cas)
+
+		sc.lock.Lock()
```
adamcfraser: Given that there are a number of exit paths, maybe paying for a defer Unlock here is worth it?

rest/server_context.go
```diff
@@ -58,7 +59,8 @@ type ServerContext struct {
 }
 
 type DatabaseConfig struct {
-	// TODO: Copy non-legacy properties into this struct
+	// TODO: Copy non-legacy properties into this struct?
+	CAS uint64
```
adamcfraser: It looks like CAS could be private?

adamcfraser (changes requested):
Approach looks good, some questions/comments.

base/bootstrap.go
bbrks: Makes sense. I've just renamed all references to "location" to be "bucket" until we're ready to support collections.

rest/config.go
bbrks: It was because I was locking inside the loop to apply fetched configs.

I've updated so that configs are fetched from the bucket without locking, and then do a 2nd loop which acquires the lock once to apply all the configs.

rest/config.go
bbrks: Good catch... They won't be. Updated so that _any_ auth fields defined on a DbConfig overwrite all of the bootstrap auth fields.

adamcfraser (approved):

PR #5059, opened 2021-06-23
CBG-1530 - Fixed 500 error calling /_config endpoint

IsaacLambat:
Caused by missing case in the redactor Unmarshall Text function. 
Rectified issue and modified unit test to test for this case in future.

### DIFF
--- a/base/redactor.go
+++ b/base/redactor.go
@@ -137,14 +137,14 @@ func (l *RedactionLevel) MarshalText() ([]byte, error) {
 // UnmarshalText unmarshals text to a RedactionLevel.
 func (l *RedactionLevel) UnmarshalText(text []byte) error {
 	switch strings.ToLower(string(text)) {
-	case "unset":
-		*l = RedactionLevelDefault
 	case "none":
 		*l = RedactNone
 	case "partial":
 		*l = RedactPartial
 	case "full":
 		*l = RedactFull
+	case "unset":
+		*l = RedactUnset
 	default:
 		return fmt.Errorf("unrecognized redaction level: %q", text)
 	}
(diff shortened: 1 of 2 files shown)


base/redactor_test.go
```diff
@@ -98,6 +98,10 @@ func TestRedactionLevelUnmarshalText(t *testing.T) {
 	assert.NoError(t, err)
 	goassert.Equals(t, level, RedactFull)
 
+	err = level.UnmarshalText([]byte("unset"))
+	assert.NoError(t, err)
+	goassert.Equals(t, level, RedactionLevelDefault)
```
bbrks, line 103: This should be `RedactUnset` really, in the same way that MarshalText returns "unset".

Want to keep the "default" handling fairly isolated and keep the JSON/Text marshaling accurate to what is being passed in/out.

bbrks (changes requested):

base/redactor.go
```diff
@@ -137,6 +137,8 @@ func (l *RedactionLevel) MarshalText() ([]byte, error) {
 // UnmarshalText unmarshals text to a RedactionLevel.
 func (l *RedactionLevel) UnmarshalText(text []byte) error {
 	switch strings.ToLower(string(text)) {
+	case "unset":
+		*l = RedactionLevelDefault
```
bbrks: This should be `RedactUnset` really, in the same way that MarshalText returns "unset".

Want to keep the "default" handling fairly isolated and keep the JSON/Text marshaling accurate to what is being passed in/out.

bbrks (changes requested):
see inline comment

bbrks (approved):

PR #5061, opened 2021-06-23
CBG-1524 - Fixed race condition in TestEnsureRevocationAfterDocMutation

IsaacLambat:
Implemented WaitForChanges after mutating doc for the first time when it has been revoked from the user.

### DIFF
--- a/rest/api_test.go
+++ b/rest/api_test.go
@@ -7066,9 +7066,8 @@ func TestEnsureRevocationAfterDocMutation(t *testing.T) {
 	// Skip to seq 19 and then update doc foo
 	revocationTester.fillToSeq(19)
 	revID = rt.createDocReturnRev(t, "doc", revID, map[string]interface{}{"channels": "A"})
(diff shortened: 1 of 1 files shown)


rest/api_test.go
```diff
@@ -7066,6 +7066,9 @@ func TestEnsureRevocationAfterDocMutation(t *testing.T) {
 	// Skip to seq 19 and then update doc foo
 	revocationTester.fillToSeq(19)
 	revID = rt.createDocReturnRev(t, "doc", revID, map[string]interface{}{"channels": "A"})
+	changes, err := rt.WaitForChanges(1, "/db/_changes?since=19&revocations=true", "user", true)
+	require.NoError(t, err)
+	assert.Equal(t, 1, len(changes.Results))
```
JRascagneres: Completely fine to leave in and doesn't hurt for extra safety but just FYI this assert should never fail.

WaitForChanges will error if the results != 1 and the require will mean it quits before it gets here 😄

JRascagneres (approved):

JRascagneres (dismissed):
LGTM

bbrks (approved):

PR #5062, opened 2021-06-23
CBG-1506: Check for user roles and complete flow

JRascagneres:
- Add check for roles based on whether db scoped or not
- Refactor to allow for easier testing of checkAdminAuth
- Like above, removed CheckRoles and CheckPermissions from ServerContext

- Tested with unit tests for both CheckRoles and full checkAdminAuth
- Has a small single X509 test for checkAdminAuth to verify it works

Based on https://github.com/couchbase/sync_gateway/pull/5055
- [x] Merge https://github.com/couchbase/sync_gateway/pull/5055

### DIFF
--- a/rest/handler.go
+++ b/rest/handler.go
@@ -49,8 +49,8 @@ const (
 	ReadOnlyAdminRole     = "ro_admin"
 )
 
-var BucketScopedEndpointRoles = [...]string{MobileSyncGatewayRole, BucketFullAccessRole, FullAdminRole}
-var ClusterScopedEndpointRoles = [...]string{ReadOnlyAdminRole}
+var BucketScopedEndpointRoles = []string{MobileSyncGatewayRole, BucketFullAccessRole, FullAdminRole}
+var ClusterScopedEndpointRoles = []string{ReadOnlyAdminRole}
 
 // If set to true, JSON output will be pretty-printed.
 var PrettyPrint bool = false
@@ -445,11 +445,20 @@ func checkAdminAuth(bucketName, basicAuthUsername, basicAuthPassword string, htt
 		return nil, http.StatusInternalServerError, err
 	}
 
+	anyResponsePermFailed := false
 	if len(responsePermissions) > 0 {
 		responsePermissionResults = permResults
+		for _, permResult := range permResults {
+			if !permResult {
+				anyResponsePermFailed = true
+				break
+			}
+		}
 	}
 
-	if statusCode == http.StatusOK {
+	// If user has required accessPerms and all response perms return with statusOK
+	// Otherwise we need to fall through to continue as the user may have access to responsePermissions through roles.
+	if statusCode == http.StatusOK && !anyResponsePermFailed {
 		return responsePermissionResults, http.StatusOK, nil
 	}
 
@@ -462,9 +471,9 @@ func checkAdminAuth(bucketName, basicAuthUsername, basicAuthPassword string, htt
 
 	var requestRoles []string
 	if bucketName != "" {
-		requestRoles = BucketScopedEndpointRoles[:]
+		requestRoles = BucketScopedEndpointRoles
 	} else {
-		requestRoles = ClusterScopedEndpointRoles[:]
+		requestRoles = ClusterScopedEndpointRoles
 	}
 
 	statusCode, err = CheckRoles(httpClient, managementEndpoints, basicAuthUsername, basicAuthPassword, requestRoles, bucketName)
--- a/rest/handler.go
+++ b/rest/handler.go
@@ -41,6 +41,17 @@ const (
 	minCompressibleJSONSize = 1000
 )
 
+// Admin API Auth Roles
+const (
+	MobileSyncGatewayRole = "mobile_sync_gateway"
+	BucketFullAccessRole  = "bucket_full_access"
+	FullAdminRole         = "admin"
+	ReadOnlyAdminRole     = "ro_admin"
+)
+
+var BucketScopedEndpointRoles = [...]string{MobileSyncGatewayRole, BucketFullAccessRole, FullAdminRole}
+var ClusterScopedEndpointRoles = [...]string{ReadOnlyAdminRole}
+
 // If set to true, JSON output will be pretty-printed.
 var PrettyPrint bool = false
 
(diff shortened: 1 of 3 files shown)


rest/handler.go
```diff
@@ -400,38 +407,61 @@ func (h *handler) checkAuth(context *db.DatabaseContext) (err error) {
...
-		h.permissionsResults = permResults
+	var requestRoles []string
+	var bucketName string
+	if dbContext != nil {
+		requestRoles = []string{"mobile_sync_gateway", "bucket_full_access", "admin"}
+		bucketName = dbContext.Bucket.GetName()
+	} else {
+		requestRoles = []string{"ro_admin"}
+		bucketName = ""
 	}
```
bbrks: What do these represent? Roles that are required to use cluster/db-scoped endpoints?

I feel like the roles should be constants in a place a bit more visible than this.

rest/handler.go
```diff
@@ -400,38 +407,61 @@ func (h *handler) checkAuth(context *db.DatabaseContext) (err error) {
 	return nil
 }
 
-func (h *handler) checkAdminAuth(dbContext *db.DatabaseContext, checkPermissions []string, storePermissionsResults bool) error {
+func checkAdminAuth(serverContext *ServerContext, dbContext *db.DatabaseContext, basicAuthUsername, basicAuthPassword string, checkPermissions []string) (map[string]bool, error) {
```
bbrks: Seems cleaner to pass in a slice of management endpoints/httpClient given only one of serverContext or dbContext will be used to obtain them.

rest/handler.go
```diff
@@ -400,38 +407,61 @@ func (h *handler) checkAuth(context *db.DatabaseContext) (err error) {
...
+		return nil, base.HTTPErrorf(statusCode, "")
 	}
 
-	if storePermissionsResults {
-		h.permissionsResults = permResults
+	var requestRoles []string
+	var bucketName string
+	if dbContext != nil {
+		requestRoles = []string{"mobile_sync_gateway", "bucket_full_access", "admin"}
+		bucketName = dbContext.Bucket.GetName()
```
bbrks: If you do remove serverContext/dbContext above, passing down a role scope seems reasonable to me, where `scope:"" == cluster`

bbrks (changes requested):

rest/handler.go
JRascagneres: Yeah db / cluster scoped endpoints have different role requirements. Fair enough, can split it out.

rest/handler.go
JRascagneres: Wanted to encapsulate the admin api work in the same function rather than have that management endpoint work separated.

rest/handler.go
adamcfraser: Agree with Ben's comments here and above - instead of passing in the entire serverContext/dbContext, make this function more stand-alone and pass in only what's needed by the function.  Makes future refactoring of serverContext/dbContext easier, as we know exactly what parts of the two this function is depending on.

adamcfraser (changes requested):

rest/handler.go
JRascagneres: Not sure how keen I am on it given the management endpoint work has to be split out so its not all encapsulated into one function anymore but I understand the benefits so done.

rest/handler.go
```diff
@@ -412,32 +439,40 @@ func (h *handler) checkAuth(context *db.DatabaseContext) (err error) {
...
+		requestRoles = ClusterScopedEndpointRoles[:]
+	}
+
+	statusCode, err = CheckRoles(httpClient, managementEndpoints, basicAuthUsername, basicAuthPassword, requestRoles, bucketName)
+	if err != nil {
+		return nil, http.StatusInternalServerError, err
 	}
 
-	return statusCode, nil
+	return permResults, statusCode, nil
```
adamcfraser, line 494: In the case where they have access through a role, this should also imply that they have access to anything otherwise gated by responsePermissions, shouldn't it?  I.e. set responsePermissionResults to match the incoming set, with all set to true?

rest/handler.go
```diff
@@ -41,6 +41,17 @@ const (
...
 
+// Admin API Auth Roles
+const (
+	MobileSyncGatewayRole = "mobile_sync_gateway"
+	BucketFullAccessRole  = "bucket_full_access"
+	FullAdminRole         = "admin"
+	ReadOnlyAdminRole     = "ro_admin"
+)
+
+var BucketScopedEndpointRoles = [...]string{MobileSyncGatewayRole, BucketFullAccessRole, FullAdminRole}
```
adamcfraser: Is the reason for [...]string vs. []string to do with slice memory allocation?  If so, do we still benefit from that if the only place we use it we're converting it to a regular slice via BucketScopedEndpointRoles[:]?  I'm not sure if the latter results in a new allocation - do you know?

adamcfraser (changes requested):
Couple of questions, generally looks good.

rest/handler.go
JRascagneres: Yeah that's a good point.

rest/handler.go
JRascagneres: It was the closest version I could think of to having a const. Also unsure on the final memory allocation when converting it back to [:]

rest/handler.go
JRascagneres: Yeah thats fair, added that case now.

rest/handler.go
JRascagneres: Did a quick benchmark but wasn't able to verify what impact it has on memory allocation. Like above, main reason for the fixed size array was to get as close to a constant as I could.

rest/handler.go
JRascagneres: This actually raises a question... If say there is accessPermission A and responsePermissions B and C. And user has A, B and has role sync_gateway what would be the expected output?

Currently it'd find the user has access A, then just return with 'you have B responsePermission'. Should we fall through to do the role check anyway for that reason?

adamcfraser: As discussed,  in any case where the user fails a permission check (including responsePermissions), we need to fall through to role check before returning (for as long as we're supporting role-based authorization).

rest/handler.go
adamcfraser: As discussed elsewhere, I think []string is preferable if the only use case is to do [:].

adamcfraser (approved):

PR #5064, opened 2021-06-24
CBG-1522 - Additional waiting for doc deletion for TestSGR2TombstoneConflictHandling

IsaacLambat:
Added waiting when a doc gets deleted to make sure its persisted to DCP
Change t1 and t2 bucket variable names to be more accurate and readable (activeBucket and passiveBucket)

### DIFF
--- a/rest/replicator_test.go
+++ b/rest/replicator_test.go
@@ -3992,9 +3992,9 @@ func TestSGR2TombstoneConflictHandling(t *testing.T) {
 			}
 
 			// Passive
-			tb2 := base.GetTestBucket(t)
+			passiveBucket := base.GetTestBucket(t)
 			remotePassiveRT := NewRestTester(t, &RestTesterConfig{
-				TestBucket: tb2,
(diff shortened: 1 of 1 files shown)


JRascagneres (dismissed):
LGTM

rest/replicator_test.go
```diff
@@ -4067,6 +4066,7 @@ func TestSGR2TombstoneConflictHandling(t *testing.T) {
 				// Delete doc on remote
 				resp = remotePassiveRT.SendAdminRequest("PUT", "/db/docid2?rev=3-abc", `{"_deleted": true}`)
 				assertStatus(t, resp, http.StatusCreated)
+				requireTombstone(t, tb2, "docid2")
```
bbrks: As discussed, I'm not sure this is adding any blocking - it's just doing an immediate assertion which will either pass or fail.

bbrks (changes requested):

bbrks (approved):

PR #5068, opened 2021-06-25
CBG-1453 - Removed unused code from expvar and full test file

IsaacLambat:
Removed most of the code from expvar.go as it was used nowhere except in the tests.
Removed all expvar.go unit tests due to them no longer being needed.

Removed creating expvar "syncgateway" as that expvar was never called anywhere.
Initialization of SyncGatewayStats is now done when it is declared.


base/expvar.go
```diff
@@ -34,20 +20,7 @@ const (
...
-	// Initialize Sync Gateway Stats
-
-	// All stats will be stored as part of this struct. Global variable accessible everywhere. To add stats see stats.go
-	SyncGatewayStats = *NewSyncGatewayStats()
-
-	// Publish our stats to expvars. This will run String method on SyncGatewayStats ( type SgwStats ) which will
-	// marshal the stats to JSON
-	expvar.Publish(StatsGroupKeySyncGateway, &SyncGatewayStats)
-}
+var SyncGatewayStats = *NewSyncGatewayStats()
```
bbrks: I guess whats left in this file can be moved to `stats.go` now?

bbrks (changes requested):

base/stats.go
```diff
@@ -58,6 +70,8 @@ type SgwStats struct {
 	dbStatsMapMutex sync.Mutex
 }
 
+var SyncGatewayStats = *NewSyncGatewayStats()
```
bbrks: @JRascagneres do we still need to do `expvar.Publish(StatsGroupKeySyncGateway, &SyncGatewayStats)` ? Or is that handled by the stats framework itself now?

JRascagneres (changes requested):
As Ben mentioned you need the keep the expvar publish line. 
Would recommend having a quick look (once thats in) at just spinning up SGW and making sure you can see stats on :4986/expvar and :4986/_metrics

JRascagneres (approved):

PR #5075, opened 2021-07-02
CBG-1518 - New ServerTLSSkipVerify config option and singular place to get CA certificates

IsaacLambat:
- [x] http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/902/ 
- [x] (passed with 1 fail, that works locally) http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/899/

Made a everything use the same function for TLS Config generation and for x509 certificate pool is generation.
If no CA Cert is specified then TLS falls back to the system root pool.
Added new unsecure config option ServerTLSSkipVerify which will cause TLS not to fall back to the system root pool if no CA cert is specified. This will also enable the insecureSkipVerify flag for TLS config.

Current help text for new config option: `Allow empty server CA Cert Path without attempting to use system root pool`
However could be changed to `Skip verification of server certificates` if that makes sense.

### DIFF
--- a/base/bucket.go
+++ b/base/bucket.go
@@ -10,9 +10,7 @@ package base
 
 import (
 	"crypto/tls"
-	"crypto/x509"
 	"fmt"
-	"io/ioutil"
 	"net/url"
 	"strconv"
 	"strings"
@@ -111,6 +109,7 @@ type BucketSpec struct {
 	Auth                          AuthHandler
 	CouchbaseDriver               CouchbaseDriver
 	Certpath, Keypath, CACertPath string         // X.509 auth parameters
+	CACertUnsetTlsSkipVerify      bool           // Use insecureSkipVerify when secure scheme (couchbases) is used and cacertpath is undefined
 	KvTLSPort                     int            // Port to use for memcached over TLS.  Required for cbdatasource auth when using TLS
 	MaxNumRetries                 int            // max number of retries before giving up
 	InitialRetrySleepTimeMS       int            // the initial time to sleep in between retry attempts (in millisecond), which will double each retry
@@ -205,7 +204,7 @@ func (b BucketSpec) GetViewQueryTimeoutMs() uint64 {
 }
 
 func (b BucketSpec) TLSConfig() *tls.Config {
-	tlsConfig, err := TLSConfigForX509(b.Certpath, b.Keypath, b.CACertPath)
+	tlsConfig, err := TLSConfigForX509(b.CACertUnsetTlsSkipVerify, b.CACertPath, b.Certpath, b.Keypath)
 	if err != nil {
 		Errorf("Error creating tlsConfig for DCP processing: %v", err)
 		return nil
@@ -213,36 +212,22 @@ func (b BucketSpec) TLSConfig() *tls.Config {
 	return tlsConfig
 }
 
-// Returns a TLSConfig based on the specified certificate paths.  If none are provided, returns tlsConfig with
-// InsecureSkipVerify:true.
-func TLSConfigForX509(certpath, keypath, cacertpath string) (*tls.Config, error) {
-
-	tlsConfig := &tls.Config{}
-	if cacertpath != "" {
-		cacertpaths := []string{cacertpath}
-		rootCerts := x509.NewCertPool()
-		for _, path := range cacertpaths {
-			cacert, err := ioutil.ReadFile(path)
-			if err != nil {
-				return nil, err
-			}
-			ok := rootCerts.AppendCertsFromPEM(cacert)
-			if !ok {
-				return nil, fmt.Errorf("can't append certs from PEM")
-			}
-		}
-		tlsConfig.RootCAs = rootCerts
-	} else {
-		// A root CA cert is required in order to secure TLS communication from a client that doesn't maintain it's own store
-		// of trusted CA certs (i.e. any SDK-based client).  If a root CA cert isn't provided, set InsecureSkipVerify=true to
-		// accept any cert provided by the server. This follows the pattern being used by the SDK for TLS connections:
-		// https://github.com/couchbase/gocbcore/blob/7b68c492c29f3f952a00a4ba97dac14cc4b2b57e/agent.go#L236
-		tlsConfig.InsecureSkipVerify = true
+// TLSConfigForX509 returns a TLSConfig based on the specified certificate paths.  If none are provided then the system
+// root pool is used (except on Windows which will error), unless CACertUnsetTlsSkipVerify is true which will return
+// tlsConfig with InsecureSkipVerify: true.
+func TLSConfigForX509(caCertUnsetTlsSkipVerify bool, caCertPath, certPath, keyPath string) (tlsConfig *tls.Config, err error) {
+	certPool, err := getRootCAs(caCertUnsetTlsSkipVerify, caCertPath)
+	if err != nil {
+		return nil, err
+	}
+	tlsConfig = &tls.Config{
+		RootCAs:            certPool,
+		InsecureSkipVerify: caCertUnsetTlsSkipVerify,
 	}
 
 	// If client cert and key are provided, add to config as x509 key pair
-	if certpath != "" && keypath != "" {
-		cert, err := tls.LoadX509KeyPair(certpath, keypath)
+	if certPath != "" && keyPath != "" {
+		cert, err := tls.LoadX509KeyPair(certPath, keyPath)
 		if err != nil {
 			return nil, err
 		}
--- a/base/bucket.go
+++ b/base/bucket.go
@@ -10,6 +10,7 @@ package base
 
 import (
 	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"net/url"
 	"strconv"
@@ -109,7 +110,7 @@ type BucketSpec struct {
 	Auth                          AuthHandler
 	CouchbaseDriver               CouchbaseDriver
 	Certpath, Keypath, CACertPath string         // X.509 auth parameters
-	CACertUnsetTlsSkipVerify      bool           // Use insecureSkipVerify when secure scheme (couchbases) is used and cacertpath is undefined
+	TLSSkipVerify                 bool           // Use insecureSkipVerify when secure scheme (couchbases) is used and cacertpath is undefined
 	KvTLSPort                     int            // Port to use for memcached over TLS.  Required for cbdatasource auth when using TLS
 	MaxNumRetries                 int            // max number of retries before giving up
 	InitialRetrySleepTimeMS       int            // the initial time to sleep in between retry attempts (in millisecond), which will double each retry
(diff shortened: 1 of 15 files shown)


db/database.go
```diff
@@ -179,6 +179,7 @@ type UnsupportedOptions struct {
 	OidcTlsSkipVerify         bool                    `json:"oidc_tls_skip_verify"`                  // Config option to enable self-signed certs for OIDC testing.
 	SgrTlsSkipVerify          bool                    `json:"sgr_tls_skip_verify"`                   // Config option to enable self-signed certs for SG-Replicate testing.
 	RemoteConfigTlsSkipVerify bool                    `json:"remote_config_tls_skip_verify"`         // Config option to enable self signed certificates for external JavaScript load.
+	CACertUnsetTlsSkipVerify  bool                    `json:"ca_cert_unset_tls_skip_verify"`         // Use insecureSkipVerify when secure scheme (couchbases) is used and cacertpath is undefined
```
adamcfraser: I think server_tls_skip_verify would be a better user-facing name for this property.

rest/config.go
```diff
@@ -138,13 +139,14 @@ func (bc *BucketConfig) MakeBucketSpec() base.BucketSpec {
...
-		Certpath:   bc.CertPath,
-		CACertPath: bc.CACertPath,
-		KvTLSPort:  tlsPort,
-		Auth:       bc,
+		Server:                   server,
+		BucketName:               bucketName,
+		Keypath:                  bc.KeyPath,
+		Certpath:                 bc.CertPath,
+		CACertPath:               bc.CACertPath,
+		CACertUnsetTlsSkipVerify: dc.Unsupported.CACertUnsetTlsSkipVerify,
```
adamcfraser: Is there a valid scenario where users would want to set both CACertPath and the unsupported TlsSkipVerify?  If they are trying to use a self-signed cert they don't need to set CACertPath in the first place, I don't think.  Assuming there isn't a valid scenario, I feel like we should throw a config validation error if both are set.  As mentioned above, this allows us to just check Spec.TlsSkipVerify=true to see whether we're running with tlsSkipVerify, and not have to repeat the CACertPath check everywhere.

base/bucket.go
```diff
@@ -111,6 +109,7 @@ type BucketSpec struct {
 	Auth                          AuthHandler
 	CouchbaseDriver               CouchbaseDriver
 	Certpath, Keypath, CACertPath string         // X.509 auth parameters
+	CACertUnsetTlsSkipVerify      bool           // Use insecureSkipVerify when secure scheme (couchbases) is used and cacertpath is undefined
```
adamcfraser: I think the simpler name TlsSkipVerify would make sense here, and (as mentioned below), only set this when CACertPath isn't specified in the config.  That way we don't need to do the "should I use CACertPath or TLSSkipVerify" every time we use the spec.

base/bucket.go
```diff
@@ -205,44 +204,30 @@ func (b BucketSpec) GetViewQueryTimeoutMs() uint64 {
 }
 
 func (b BucketSpec) TLSConfig() *tls.Config {
-	tlsConfig, err := TLSConfigForX509(b.Certpath, b.Keypath, b.CACertPath)
+	tlsConfig, err := TLSConfigForX509(b.CACertUnsetTlsSkipVerify, b.CACertPath, b.Certpath, b.Keypath)
```
adamcfraser: I don't see TLSConfigForX509 being used anywhere else.  Given that, can we just move it inside this method, since otherwise this method is just doing a bit of error handling?

base/gocb_utils.go
```diff
@@ -114,7 +103,11 @@ func GoCBCoreTLSRootCAProvider(caCertPath string) (func() *x509.CertPool, error)
 	}, nil
 }
 
-func getRootCAs(caCertPath string) (*x509.CertPool, error) {
+func getRootCAs(CACertUnsetTlsSkipVerify bool, caCertPath string) (*x509.CertPool, error) {
```
adamcfraser: I think it's better to leave this function with a single purpose - retrieve the rootCAs - and have callers make the decision about whether to call it in the first place based on TlsSkipVerify.

base/gocb_utils.go
```diff
@@ -125,14 +118,12 @@ func getRootCAs(caCertPath string) (*x509.CertPool, error) {
...
 		}
 
 		return rootCAs, nil
 	}
 
-	// We're purposefully ignoring the error here Due to the fact that the main error case is that this call is not
-	// supported in Windows.
-	rootCAs, _ := x509.SystemCertPool()
-	return rootCAs, nil
+	rootCAs, err := x509.SystemCertPool()
```
adamcfraser, line 144: Have we already had a discussion about how best to handle the windows error case?  Since the error is specifically "unable to load the system cert pool", I think logging that error and then returning an empty (non-nil) cert pool would be a better approach.  We'd expect this to fail to connect over couchbases unless insecureSkipVerify is set, but I don't think we need to pre-emptively error based on the failed SystemCertPool() call.

base/bucket.go
```diff
@@ -205,44 +204,30 @@ func (b BucketSpec) GetViewQueryTimeoutMs() uint64 {
...
-		tlsConfig.RootCAs = rootCerts
-	} else {
-		// A root CA cert is required in order to secure TLS communication from a client that doesn't maintain it's own store
-		// of trusted CA certs (i.e. any SDK-based client).  If a root CA cert isn't provided, set InsecureSkipVerify=true to
-		// accept any cert provided by the server. This follows the pattern being used by the SDK for TLS connections:
-		// https://github.com/couchbase/gocbcore/blob/7b68c492c29f3f952a00a4ba97dac14cc4b2b57e/agent.go#L236
-		tlsConfig.InsecureSkipVerify = true
+// TLSConfigForX509 returns a TLSConfig based on the specified certificate paths.  If none are provided then the system
+// root pool is used (except on Windows which will error), unless CACertUnsetTlsSkipVerify is true which will return
+// tlsConfig with InsecureSkipVerify: true.
```
adamcfraser: This function description makes it difficult to be sure exactly what happens on Windows if insecureSkipVerify is set to true.  Maybe break up the last sentence to better differentiate the cases?

rest/handler.go
```diff
@@ -224,7 +224,7 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []string, respo
 		if dbContext != nil {
 			managementEndpoints, httpClient, err = dbContext.ObtainManagementEndpointsAndHTTPClient()
 		} else {
-			managementEndpoints, httpClient, err = h.server.ObtainManagementEndpointsAndHTTPClient()
+			managementEndpoints, httpClient, err = h.server.ObtainManagementEndpointsAndHTTPClient(dbContext.BucketSpec.CACertUnsetTlsSkipVerify)
```
adamcfraser: Since dbContext == nil here, we can't use it to get the value of tlsSkipVerify.  Might need additional testing for this path. 

@JRascagneres Does this imply that we also need a server-level unsupported insecureSkipVerify flag, with the option to override at the db level?

rest/server_context.go
```diff
@@ -1155,13 +1155,13 @@ func (sc *ServerContext) updateCalculatedStats() {
 
 }
 
-func initClusterAgent(clusterAddress, clusterUser, clusterPass, certPath, keyPath, caCertPath string, timeoutSeconds *int) (*gocbcore.Agent, error) {
+func initClusterAgent(clusterAddress, clusterUser, clusterPass, certPath, keyPath, caCertPath string, CACertUnsetTlsSkipVerify bool, timeoutSeconds *int) (*gocbcore.Agent, error) {
```
adamcfraser:
```suggestion
func initClusterAgent(clusterAddress, clusterUser, clusterPass, certPath, keyPath, caCertPath string, caCertUnsetTlsSkipVerify bool, timeoutSeconds *int) (*gocbcore.Agent, error) {
```

adamcfraser: Avoid capitalization in parameter names, although this name is subject to change anyway.

rest/server_context.go
```diff
@@ -1207,9 +1207,9 @@ var tempConnectionDetailsForManagementEndpoints = func() (serverAddress string,
 	return base.UnitTestUrl(), base.TestClusterUsername(), base.TestClusterPassword(), "", "", ""
 }
 
-func (sc *ServerContext) ObtainManagementEndpointsAndHTTPClient() ([]string, *http.Client, error) {
+func (sc *ServerContext) ObtainManagementEndpointsAndHTTPClient(CACertUnsetTlsSkipVerify bool) ([]string, *http.Client, error) {
```
adamcfraser: I think we'd pull tlsSkipVerify from the same place we get the rest of the tlsConfig (currently tempConnectionDetails), and not pass into this method.

adamcfraser (dismissed):
In general I feel like we're passing the insecureSkipVerify parameter around too much - I've added a few suggestions on how to rework to avoid this.  Beyond that, need to discuss how to handle insecureSkipVerify for non-db-scoped connections to Couchbase Server.

rest/handler.go
JRascagneres: @adamcfraser Eventually this option is going to be at the server level. Especially this is backed up by @bbrks work. There is only one server that the persistent config setup can connect to and this server address cannot be overridden at the db level. Guessing the db-level is temporary?

rest/server_context.go
JRascagneres: Agreed that aligns with it being a server config item.

base/gocb_utils.go
```diff
@@ -134,14 +135,16 @@ func getRootCAs(caCertPath string) (*x509.CertPool, error) {
...
 
 		return rootCAs, nil
 	}
 
-	// We're purposefully ignoring the error here Due to the fact that the main error case is that this call is not
-	// supported in Windows.
-	rootCAs, _ := x509.SystemCertPool()
+	rootCAs, err := x509.SystemCertPool()
+	if err != nil {
+		rootCAs = &x509.CertPool{}
```
bbrks: x509.CertPool contains a couple of maps that would panic if uninitialized when adding things to them.

https://play.golang.org/p/SXWLnrfnWKc

In this case that's probably not going to happen, but it seems safer to use `x509.NewCertPool()` anyway. We don't really know what gocb is doing under the hood with the CertPool we give them.

bbrks (changes requested):
Looks alright in general. A few test comments and one about how we're creating a CertPool.

base/gocb_utils_test.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	//
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			sc, err := GoCBv2SecurityConfig(test.tlsSkipVerify, test.caCertPath)
+			if test.expectError {
+				assert.Error(t, err)
+				assert.Nil(t, sc.TLSRootCAs)
+				return
+			}
```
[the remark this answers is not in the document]
IsaacLambat: GoLint is telling me that there is no need for this because the return would make the else redundant.

Should I still add the else for readability purposes?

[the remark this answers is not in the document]
bbrks: Ah right - I missed the `return` above :) Yeah dropping the else is fine, and just including the `assert.NoError`

bbrks (approved):
lgtm

PR #5076, opened 2021-07-06
[2.8.3] CBG-1454: Removal messages are not sent on push replications

IsaacLambat:
Removal messages are not sent on push replicates when no filtered channels are in the document channels.
Added a unit test to test this behaviour with and without the document channels being one of the filtered channels.

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -235,14 +235,15 @@ const (
 )
 
 type sendChangesOptions struct {
-	docIDs            []string
-	since             SequenceID
-	continuous        bool
-	activeOnly        bool
-	batchSize         int
-	channels          base.Set
-	clientType        clientType
-	ignoreNoConflicts bool
+	docIDs                 []string
+	since                  SequenceID
+	continuous             bool
+	activeOnly             bool
+	batchSize              int
+	channels               base.Set
+	clientType             clientType
+	ignoreNoConflicts      bool
+	disableRemovalMessages bool // stops returning a document removal message when the channels are not any of the filtered channels
 }
 
 // Sends all changes since the given sequence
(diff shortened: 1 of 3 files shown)


db/blip_handler.go
```diff
@@ -292,6 +293,9 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 			if !strings.HasPrefix(change.ID, "_") {
 				for _, item := range change.Changes {
 					changeRow := []interface{}{change.Seq, change.ID, item["rev"], change.Deleted}
+					if opts.disableRemovalMessages && change.allRemoved {
```
adamcfraser, line 296: I feel like it would be preferable to do this check at a lower level - indicating whether we want removals in the options we pass to generateBlipSyncChanges, and filter it out at that level.  
@JRascagneres Is this handled by ChangesOptions.Revocations in Lithium for the removals case in ISGR?

adamcfraser (commented):
Would like to push this down so that all filtering happens at the same layer, review with Jacques what's done in Lithium.

db/blip_handler.go
JRascagneres: @adamcfraser For revocations it's handled by changes. For removals its done at this level.

adamcfraser: Ok - it's reasonable to leave where it is, then.

rest/replicator_test.go
```diff
@@ -968,6 +968,108 @@ func TestActiveReplicatorPullOneshot(t *testing.T) {
...
+			expectDoc := "test"
+			if !test.expectRemovalMessage {
+				// Create document after to verify removal message was not sent on changes endpoint
+				resp = activeRT.SendAdminRequest(http.MethodPut, "/db/docAfter", `{"channels":["chanA"]}`)
+				assertStatus(t, resp, http.StatusCreated)
+				expectDoc = "docAfter"
+			}
+
+			assert.NoError(t, ar.Start())
+
```
adamcfraser: I'm not sure this test is completely verifying the scenarios we want to validate.  
1. It's sort of validating that the updated version of "test" hasn't been sent to the passive peer (by checking for 'docAfter', but I think it would be even better to verify that the previous version of "test" still exists on the remote as rev 1-x.
2. We should also test the case where the replication is running for both chanA and chanB, rev 1-x is in chanA and chanB, and then rev 2-x is just in chanA, to make sure that continues to be replicated.

adamcfraser (changes requested):
Change looks ok, but would like a bit more test coverage.

adamcfraser (approved):

PR #5078, opened 2021-07-07
Fix Principal History Pruning

JRascagneres:
Added tests around principal history pruning. 
Also fix to the history length pruning which would merge old entries but not actually remove the oldest.


bbrks (approved):
LGTM

adamcfraser (approved):

PR #5079, opened 2021-07-07
Add Document ChannelSet History Pruning

JRascagneres:
Adds pruning to the document channel set history.
Has a static maximum length per channel of 5 entries and is pruned when a channel attempts to be added to the history.
Has a unit test to validate the behaviour.

### DIFF
--- a/db/document.go
+++ b/db/document.go
@@ -27,6 +27,9 @@ import (
 // Non-winning bodies smaller than this size are more efficient to store inline.
 const MaximumInlineBodySize = 250
 
+// DocumentHistoryMaxEntriesPerChannel is the maximum allowed entries per channel in the Document ChannelSetHistory
+const DocumentHistoryMaxEntriesPerChannel = 5
+
 type DocumentUnmarshalLevel uint8
 
 const (
@@ -43,10 +46,6 @@ const (
 	RemovedRedactedDocument = `{"` + BodyRemoved + `":true}`
 	// DeletedDocument is returned by SG when a given document has been deleted
 	DeletedDocument = `{"` + BodyDeleted + `":true}`
-)
-
-const (
-	DocumentHistoryMaxEntriesPerChannel = 5
 )
 
 // Maps what users have access to what channels or roles, and when they got that access.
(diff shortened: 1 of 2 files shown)


db/document.go
```diff
@@ -881,6 +886,49 @@ func (doc *Document) updateChannelHistory(channelName string, seq uint64, additi
...
+			if entry.Start < secondOldestEntryStartSeq {
+				secondOldestEntryStartSeq = entry.Start
+				secondOldestEntryIdx = entryIdx
+			}
+		}
+	}
+
+	if entryCount == DocumentHistoryMaxEntriesPerChannel {
+		doc.ChannelSetHistory[secondOldestEntryIdx].Start = oldestEntryStartSeq
+		doc.ChannelSetHistory = append(doc.ChannelSetHistory[:oldestEntryIdx], doc.ChannelSetHistory[oldestEntryIdx+1:]...)
```
bbrks, line 925: edit: nevermind - misread

bbrks (changes requested):

db/document.go
```diff
@@ -881,6 +886,49 @@ func (doc *Document) updateChannelHistory(channelName string, seq uint64, additi
...
+			}
+
+			if entry.Start < secondOldestEntryStartSeq {
+				secondOldestEntryStartSeq = entry.Start
+				secondOldestEntryIdx = entryIdx
+			}
+		}
+	}
+
+	if entryCount == DocumentHistoryMaxEntriesPerChannel {
```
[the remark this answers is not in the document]
JRascagneres: Yeah fair enough. Don't think that should happen but this should limit the damage of that possibility

db/document.go
adamcfraser: Should we be using the memory safe version of a slice cut here?  https://github.com/golang/go/wiki/SliceTricks

adamcfraser (changes requested):
Looks good, question about the slice cut being used.

db/document.go
JRascagneres: Don't think so. It doesn't contain any pointers so the current method should suffice with no memory leak issues.

adamcfraser (approved):

PR #5082, opened 2021-07-07
CBG-1535 - Enforce TLS by default

IsaacLambat:
- [x] http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/929/ (test TestDBOnlineWithDelayAndImmediate failed on integration but works locally)

https://issues.couchbase.com/browse/CM-837
Default bucket is no server is provided is now couchbases://localhost:8091 regardless of flag
If the server is a Walrus bucket, then this is treated as a secure server meaning UseTLSServer does not need to be set.

See https://docs.google.com/document/d/1Az-Fa75WebU1cokZQ1Xh6EOcL1JeqKcCqKso9rr6Jss/edit for config details

### DIFF
--- a/rest/admin_api_test.go
+++ b/rest/admin_api_test.go
@@ -35,8 +35,6 @@ import (
 
 // Reproduces CBG-1412 - JSON strings in some responses not being correctly escaped
 func TestPutDocSpecialChar(t *testing.T) {
-	rt := NewRestTester(t, nil)
-	defer rt.Close()
 	testCases := []struct {
 		name         string
 		pathDocID    string
@@ -80,6 +78,8 @@ func TestPutDocSpecialChar(t *testing.T) {
 	}
 	for _, testCase := range testCases {
 		t.Run(testCase.name, func(t *testing.T) {
+			rt := NewRestTester(t, nil)
+			defer rt.Close()
 			if testCase.eeOnly && !base.IsEnterpriseEdition() {
 				t.Skipf("Skipping enterprise-only test")
 			}
@@ -92,6 +92,8 @@ func TestPutDocSpecialChar(t *testing.T) {
 	}
 
 	t.Run("Delete Double quote Doc ID", func(t *testing.T) { // Should be done for Local Document deletion when it returns response
+		rt := NewRestTester(t, nil)
+		defer rt.Close()
 		tr := rt.SendAdminRequest("PUT", fmt.Sprintf("/db/%s", `del"ete"Me`), "{}") // Create the doc to delete
 		assertStatus(t, tr, http.StatusCreated)
 		var putBody struct {
--- a/rest/admin_api_test.go
+++ b/rest/admin_api_test.go
@@ -35,6 +35,8 @@ import (
 
 // Reproduces CBG-1412 - JSON strings in some responses not being correctly escaped
 func TestPutDocSpecialChar(t *testing.T) {
+	rt := NewRestTester(t, nil)
+	defer rt.Close()
 	testCases := []struct {
 		name         string
 		pathDocID    string
@@ -78,8 +80,6 @@ func TestPutDocSpecialChar(t *testing.T) {
 	}
 	for _, testCase := range testCases {
 		t.Run(testCase.name, func(t *testing.T) {
-			rt := NewRestTester(t, nil)
-			defer rt.Close()
 			if testCase.eeOnly && !base.IsEnterpriseEdition() {
 				t.Skipf("Skipping enterprise-only test")
 			}
@@ -92,8 +92,6 @@ func TestPutDocSpecialChar(t *testing.T) {
 	}
 
 	t.Run("Delete Double quote Doc ID", func(t *testing.T) { // Should be done for Local Document deletion when it returns response
-		rt := NewRestTester(t, nil)
-		defer rt.Close()
 		tr := rt.SendAdminRequest("PUT", fmt.Sprintf("/db/%s", `del"ete"Me`), "{}") // Create the doc to delete
 		assertStatus(t, tr, http.StatusCreated)
 		var putBody struct {
--- a/rest/config.go
+++ b/rest/config.go
@@ -841,6 +841,11 @@ func (sc *StartupConfig) ValidateInsecureTLSConnection() (err error) {
 			err = fmt.Errorf("a TLS key and cert path must be provided when not allowing insecure TLS connections")
 			return err
 		}
+	} else { // Make sure TLS key and cert is not provided
+		if sc.API.HTTPS.TLSKeyPath != "" || sc.API.HTTPS.TLSCertPath != "" {
+			err = fmt.Errorf("cannot use TLS and also use insecure TLS connections")
+			return err
+		}
 	}
 	return nil
 }
--- a/rest/config.go
+++ b/rest/config.go
@@ -833,6 +833,18 @@ func (sc *StartupConfig) validate() (errorMessages error) {
 	return errorMessages
 }
 
+// Validate insecure connections
+func (sc *StartupConfig) ValidateInsecureTLSConnection() (err error) {
+	// Validate SSL is provided if not allowing unsecure connections
+	if sc.API.HTTPS.AllowInsecureTLSConnections == nil || !*sc.API.HTTPS.AllowInsecureTLSConnections {
+		if sc.API.HTTPS.TLSKeyPath == "" || sc.API.HTTPS.TLSCertPath == "" {
+			err = fmt.Errorf("a TLS key and cert path must be provided when not allowing insecure TLS connections")
+			return err
+		}
+	}
+	return nil
+}
+
 // ServerContext creates a new ServerContext given its configuration and performs the context validation.
 func setupServerContext(config *StartupConfig, persistentConfig bool) (*ServerContext, error) {
 	// Logging config will now have been loaded from command line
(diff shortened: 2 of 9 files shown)


bbrks:
Need to do the work to get the `sgw-pipeline-litecore-ee` stage passing (pass in a command line flag for this new option from the litecore test script)

I can help get you started with above when you have chance to look at this @IsaacLambat

rest/config_legacy.go
```diff
@@ -117,12 +117,13 @@ func (lc *LegacyServerConfig) ToStartupConfig() (*StartupConfig, DbConfigMap, er
...
-			CACertPath:   dbConfig.CACertPath,
-			X509CertPath: dbConfig.CertPath,
-			X509KeyPath:  dbConfig.KeyPath,
+			Server:                         *dbConfig.Server,
+			Username:                       dbConfig.Username,
+			Password:                       dbConfig.Password,
+			CACertPath:                     dbConfig.CACertPath,
+			X509CertPath:                   dbConfig.CertPath,
+			X509KeyPath:                    dbConfig.KeyPath,
+			AllowInsecureServerConnections: base.BoolPtr(true),
```
JRascagneres: Still need this to be false by default in the default case. Think this should mean just removing this as the current DefaultStartupConfig() should handle it

rest/config_legacy.go
```diff
@@ -133,6 +134,9 @@ func (lc *LegacyServerConfig) ToStartupConfig() (*StartupConfig, DbConfigMap, er
 			Pretty:             lc.Pretty,
 			CompressResponses:  lc.CompressResponses,
 			HideProductVersion: lc.HideProductVersion,
+			HTTPS: HTTPSConfig{
+				AllowInsecureTLSConnections: base.BoolPtr(true),
```
JRascagneres: Same as above

JRascagneres (changes requested):

rest/config.go
```diff
@@ -72,7 +72,7 @@ type BucketConfig struct {
 
 func (bc *BucketConfig) MakeBucketSpec() base.BucketSpec {
 
-	server := "http://localhost:8091"
+	server := "couchbases://localhost:8091"
```
bbrks: 8091 is a http port, need to remove to allow it to default to 11207 ([memcached_ssl_port](https://docs.couchbase.com/server/current/install/install-ports.html#detailed-port-description))

```suggestion
	server := "couchbases://localhost"
```

rest/config.go
```diff
@@ -809,6 +809,21 @@ func (config *StartupConfig) Serve(addr string, handler http.Handler) error {
 	)
 }
 
+// Validate insecure connections
+func (sc *StartupConfig) validateInsecureTLSConnection() (err error) {
+	// Validate SSL is provided if not allowing unsecure connections
+	if sc.API.HTTPS.AllowInsecureTLSConnections == nil || !*sc.API.HTTPS.AllowInsecureTLSConnections {
+		if sc.API.HTTPS.TLSKeyPath == "" || sc.API.HTTPS.TLSCertPath == "" {
+			return fmt.Errorf("a TLS key and cert path must be provided when not allowing insecure TLS connections")
```
bbrks: The double negative in this message is a bit confusing. Think we can rephrase this to be:

```suggestion
			return fmt.Errorf("TLS key path and cert path must be provided unless api.https.allow_insecure_tls_connections is set")
```

rest/config.go
```diff
@@ -809,6 +809,21 @@ func (config *StartupConfig) Serve(addr string, handler http.Handler) error {
...
+		}
+	} else { // Make sure TLS key and cert is not provided
+		if sc.API.HTTPS.TLSKeyPath != "" || sc.API.HTTPS.TLSCertPath != "" {
+			return fmt.Errorf("cannot use TLS and also use insecure TLS connections")
```
bbrks: likewise

```suggestion
			return fmt.Errorf("cannot use TLS with allow_insecure_tls_connections set")
```

rest/config_legacy.go
```diff
@@ -427,7 +427,7 @@ func ParseCommandLine(args []string, handling flag.ErrorHandling) (*LegacyServer
 	_ = flagSet.Bool("api.admin_interface_authentication", true, "")
 	_ = flagSet.Bool("api.metrics_interface_authentication", true, "")
 	_ = flagSet.Bool("bootstrap.allow_insecure_server_connections", false, "")
-	_ = flagSet.Bool("bootstrap.allow_insecure_tls_connections", false, "")
+	_ = flagSet.Bool("api.https.allow_insecure_tls_connections", false, "")
```
bbrks: I think this is already done, which means the PR should be rebased onto the latest master.

rest/config_startup.go
```diff
@@ -106,9 +109,10 @@ type APIConfig struct {
 }
 
 type HTTPSConfig struct {
-	TLSMinimumVersion string `json:"tls_minimum_version,omitempty"     help:"The minimum allowable TLS version for the REST APIs"`
-	TLSCertPath       string `json:"tls_cert_path,omitempty" help:"The TLS cert file to use for the REST APIs"`
-	TLSKeyPath        string `json:"tls_key_path,omitempty"  help:"The TLS key file to use for the REST APIs"`
+	TLSMinimumVersion           string `json:"tls_minimum_version,omitempty"     help:"The minimum allowable TLS version for the REST APIs"`
+	TLSCertPath                 string `json:"tls_cert_path,omitempty" help:"The TLS cert file to use for the REST APIs"`
+	TLSKeyPath                  string `json:"tls_key_path,omitempty"  help:"The TLS key file to use for the REST APIs"`
+	AllowInsecureTLSConnections *bool  `json:"allow_insecure_tls_connections" help:"Allow TLS to be optional for the REST APIs"`
```
bbrks: It's not just optional, since we fail to start up if you provide TLS cert/key paths.

My phrasing below still isn't perfect but is a bit clearer.

```suggestion
	AllowInsecureTLSConnections *bool  `json:"allow_insecure_tls_connections" help:"Allows REST APIs to start without TLS"`
```

rest/server_context.go
```diff
@@ -372,6 +372,17 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(config DatabaseConfig, useE
 		return nil, err
 	}
 
+	secure := spec.IsTLS()
+	if sc.config.Bootstrap.AllowInsecureServerConnections == nil || !*sc.config.Bootstrap.AllowInsecureServerConnections {
+		if !secure && !spec.IsWalrusBucket() {
+			return nil, fmt.Errorf("couchbase server URL must use secure protocol when disallowing insecure server connections. Current URL: %s", spec.Server)
```
bbrks: "disallowing insecure connections" is the default here, so again the double negative is just going to cause confusion for users encountering this (they won't know about that option).


```suggestion
			return nil, fmt.Errorf("Must use secure scheme in Couchbase Server URL, or opt out using allow_insecure_server_connections: %s", spec.Server)
```

rest/utilities_testing.go
```diff
@@ -42,17 +42,19 @@ import (
...
+	DatabaseConfig                *DbConfig                  // Supports additional config options.  BucketConfig, Name, Sync, Unsupported will be ignored (overridden)
+	InitSyncSeq                   uint64                     // If specified, initializes _sync:seq on bucket creation.  Not supported when running against walrus
+	EnableNoConflictsMode         bool                       // Enable no-conflicts mode.  By default, conflicts will be allowed, which is the default behavior
+	distributedIndex              bool                       // Test with walrus-based index bucket
+	TestBucket                    *base.TestBucket           // If set, use this bucket instead of requesting a new one.
+	adminInterface                string                     // adminInterface overrides the default admin interface.
+	sgReplicateEnabled            bool                       // sgReplicateManager disabled by default for RestTester
+	sgr1Replications              []*ReplicateV1ConfigLegacy // sgr1Replications are a list of replications to enable on the server context.
+	hideProductInfo               bool
+	denyInsecureTLSConnections    bool // If true, TLS will be used with SG
```
bbrks: Is it `denyInsecure...` here rather than `allowInsecure...` so that the zero value of false matches SG's default of true?

Is Could that be a bit misleading when testing?

bbrks (changes requested):
I think it looks OK besides rephrasing some of the error messages to make them more user-friendly and some small cleanup

rest/admin_api_test.go
```diff
@@ -80,6 +78,8 @@ func TestPutDocSpecialChar(t *testing.T) {
 	}
 	for _, testCase := range testCases {
 		t.Run(testCase.name, func(t *testing.T) {
+			rt := NewRestTester(t, nil)
```
JRascagneres: Any reason we've moved this? Seems unrelated to this PR

rest/admin_api_test.go
```diff
@@ -92,6 +92,8 @@ func TestPutDocSpecialChar(t *testing.T) {
 	}
 
 	t.Run("Delete Double quote Doc ID", func(t *testing.T) { // Should be done for Local Document deletion when it returns response
+		rt := NewRestTester(t, nil)
```
JRascagneres: ditto

rest/config_legacy.go
```diff
@@ -112,6 +115,9 @@ func (lc *LegacyServerConfig) ToStartupConfig() (*StartupConfig, DbConfigMap, er
 			Pretty:             lc.Pretty,
 			CompressResponses:  lc.CompressResponses,
 			HideProductVersion: lc.HideProductVersion,
+			HTTPS: HTTPSConfig{
+				UseTLSClient: lc.UseTLSClient,
```
JRascagneres, line 127: What is this going to end up being if its not specified in the config? Defaults to false maybe before flag / Default overrides?

JRascagneres (changes requested):
Just stuck a few comments on and discussion in person about other bits, mostly around defaults / flags

rest/config_legacy.go
IsaacLambat: Tested it and the default of true does get applied if the user has not specified it in their config.

bbrks (changes requested):
help text needs updating - and need to investigate why RestTester works with the client use TLS option defaulting to true (shouldn't it fail to start up the server because we don't have any HTTPS certs?)

bbrks (approved):

JRascagneres (approved):

PR #5088, opened 2021-07-09
CBG-1402 Switch to updated digest computation for new attachments added in Lithium or later

sarathkumarsivan:
Changes to switch to updated digest computation for new attachments added in Lithium or later.


db/attachment.go
```diff
@@ -253,6 +279,21 @@ func (db *Database) ForEachStubAttachment(body Body, minRevpos int, callback Att
...
 }
 
+func GetAttVersion(meta map[string]interface{}) int {
+	if ver, ok := meta["ver"].(json.Number); ok {
+		if version, err := ver.Int64(); err != nil {
+			return AttVersion1
+		} else {
+			return int(version)
+		}
+	} else if ver, ok := meta["ver"].(int); ok {
```
bbrks: What scenarios exist that set this value as an `int`?

Do we want to be made aware (via logging or an error) of invalid types rather than silently return version 1?

db/blip_handler.go
```diff
@@ -931,7 +931,23 @@ func (bh *blipHandler) handleProveAttachment(rq *blip.Message) error {
...
+		if docID == "" {
+			return base.HTTPErrorf(http.StatusBadRequest, "no docID sent with proveAttachment")
+		}
+		attachmentAllowedKey = docID + digest
+	}
+
+	allowedAttachment := bh.allowedAttachment(attachmentAllowedKey)
+	if allowedAttachment.counter <= 0 {
+		return base.HTTPErrorf(http.StatusForbidden, "Attachment's doc not being synced")
+	}
```
bbrks: This looks like new handling that we weren't previously doing. Is it a gap in our `proveAttachment` implementation? What scenario is this check trying to catch?

rest/blip_api_test.go
```diff
@@ -2752,15 +2752,15 @@ func TestBlipDeltaSyncNewAttachmentPull(t *testing.T) {
 		// Check the request body was the actual delta
 		msgBody, err := msg.Body()
 		assert.NoError(t, err)
-		assert.Equal(t, `{"_attachments":[{"hello.txt":{"digest":"sha1-Kq5sNclPz7QV2+lfQIuc6R7oRu0=","length":11,"revpos":2,"stub":true}}]}`, string(msgBody))
+		assert.Equal(t, `{"_attachments":[{"hello.txt":{"digest":"sha1-Kq5sNclPz7QV2+lfQIuc6R7oRu0=","length":11,"revpos":2,"stub":true,"ver":2}}]}`, string(msgBody))
```
bbrks: Do we need to send `"ver":2` to Couchbase Lite? Or is this an internal SG implementation detail?

rest/replicator_test.go
```diff
@@ -339,10 +339,8 @@ func TestActiveReplicatorPullAttachments(t *testing.T) {
 	assert.Equal(t, revID, doc2.SyncData.CurrentRev)
 	assert.Equal(t, "rt2", doc.GetDeepMutableBody()["source"])
 
-	// Because we're targeting a Hydrogen node that supports proveAttachment, we only end up sending the attachment once.
-	// If targeting a pre-hydrogen node, GetAttachment would be 2.
-	assert.Equal(t, int64(1), ar.Pull.GetStats().GetAttachment.Value())
-	assert.Equal(t, int64(1), ar.Pull.GetStats().ProveAttachment.Value())
+	assert.Equal(t, int64(2), ar.Pull.GetStats().GetAttachment.Value())
+	assert.Equal(t, int64(0), ar.Pull.GetStats().ProveAttachment.Value())
```
bbrks, line 355: Can we leave a comment here explaining why we're not expecting proveAttachment to work in this case?

bbrks (changes requested):
Left some initial feedback

db/blip_handler.go
sarathkumarsivan: Yeah, this change is not required, proveAttachment handling can be refactored instead.

db/attachment.go
```diff
@@ -200,7 +218,13 @@ func (db *Database) setAttachments(attachments AttachmentData) error {
 
 	for key, data := range attachments {
 		attachmentSize := int64(len(data))
-		_, err := db.Bucket.AddRaw(attachmentKeyToString(key), 0, data)
+		var attKey string
+		if strings.Contains(string(key), ":") {
+			attKey = attachmentV2KeyToString(key)
+		} else {
+			attKey = attachmentKeyToString(key)
+		}
```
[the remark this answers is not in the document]
sarathkumarsivan: Yeah, we can make use of the existing AttachmentData which maps attachment keys to value bytes.

db/attachment.go
sarathkumarsivan: This value is set as int from the unit test cases, when looking at this further, it make sense to use the base.ToInt64 while   obtaining the attachment version from meta.

bbrks (changes requested):
Looks mostly good, but 2 questions:

db/attachment.go
```diff
@@ -218,29 +236,37 @@ type AttachmentCallback func(name string, digest string, knownData []byte, meta
...
-			data, err := db.GetAttachment(AttachmentKey(digest))
-			if err != nil && !base.IsDocNotFoundError(err) {
-				return err
+			version, ok := GetAttachmentVersion(meta)
+			if !ok {
+				return ErrAttachmentVersion
+			}
+			var data []byte
+			var err error
+			if version == AttVersion1 {
```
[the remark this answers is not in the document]
sarathkumarsivan: Yeah, the issue was that the ["data"] getting populated in the attachment meta because of the getAttachment call and the actual data was not getting removed from the meta.

rest/blip_api_test.go
```diff
@@ -2856,6 +2855,7 @@ func TestBlipDeltaSyncPushAttachment(t *testing.T) {
 // 5. Update doc in the test client by adding another attachment
 // 6. Have that update pushed using delta sync via the continuous replication started in step 2
 func TestBlipDeltaSyncPushPullNewAttachment(t *testing.T) {
+	t.Skip("Skipping this test temporarily")
```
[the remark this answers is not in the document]
sarathkumarsivan: Yeah this was intentional, but this has been enable in the latest commit.

rest/blip_api_test.go
```diff
@@ -2931,6 +2930,77 @@ func TestBlipDeltaSyncPushPullNewAttachment(t *testing.T) {
...
+	attachments, ok := respBody[db.BodyAttachments].(map[string]interface{})
+	require.True(t, ok)
+	assert.Len(t, attachments, 1)
+	hello, ok := attachments["hello.txt"].(map[string]interface{})
+	require.True(t, ok)
+	assert.Equal(t, "sha1-Kq5sNclPz7QV2+lfQIuc6R7oRu0=", hello["digest"])
+	assert.Equal(t, float64(11), hello["length"])
+	assert.Equal(t, float64(2), hello["revpos"])
+	assert.True(t, hello["stub"].(bool))
+}
```
bbrks, line 3005: I think we should do some assertions on the stats around `HandleGetAttachment`/`HandleGetAttachmentBytes` to ensure we're not sending the attachment data for the 2nd revision

bbrks (changes requested):
Looks like the new test should be asserting on some attachment stats

rest/blip_api_test.go
sarathkumarsivan: Sure, included assertions on the stats around HandleGetAttachment/HandleGetAttachmentBytes.

db/attachment.go
```diff
@@ -218,32 +236,40 @@ type AttachmentCallback func(name string, digest string, knownData []byte, meta
...
+				data, err = db.GetAttachment(attachmentKey)
+				if err != nil && !base.IsDocNotFoundError(err) {
+					return err
+				}
 			}
-
 			if newData, err := callback(name, digest, data, meta); err != nil {
 				return err
-			} else if newData != nil {
+			} else if newData != nil && version == AttVersion1 {
```
adamcfraser: Why are we only doing this if the version is AttVersion1?

db/attachment.go
```diff
@@ -218,32 +236,40 @@ type AttachmentCallback func(name string, digest string, knownData []byte, meta
...
 			}
 			digest, ok := meta["digest"].(string)
 			if !ok {
-				return base.HTTPErrorf(400, "Invalid attachment")
+				return base.HTTPErrorf(http.StatusBadRequest, "Invalid attachment")
 			}
-			data, err := db.GetAttachment(AttachmentKey(digest))
-			if err != nil && !base.IsDocNotFoundError(err) {
-				return err
+			version, ok := GetAttachmentVersion(meta)
```
adamcfraser: As this PR stands, the incoming `body` is never going to have the attachment version set.  We've discussed a subsequent enhancement to retrieve the current revision from the bucket in order to be able to properly compute revpos, but if we're at this point and there's an attachment on the incoming document with revpos < minRevpos, then that's an attachment Sync Gateway doesn't know about, and should be stored in the new format.

db/blip_handler.go
```diff
@@ -1011,10 +1014,10 @@ func (bh *blipHandler) sendGetAttachment(sender *blip.Sender, docID string, name
...
 	}
 
-	lNum, metaLengthOK := meta["length"].(json.Number)
-	metaLength, err := lNum.Int64()
-	if err != nil {
-		return nil, err
+	lNum, metaLengthOK := meta["length"]
+	metaLength, ok := base.ToInt64(lNum)
+	if !ok {
+		return nil, fmt.Errorf("invalid attachment length found in meta")
```
adamcfraser, line 1032: There doesn't seem to be a functional change in the above code  (other than error message)- what's the motivation for changing it?

db/attachment.go
```diff
@@ -180,27 +190,35 @@ func (db *Database) loadAttachmentsData(attachments AttachmentsMeta, minRevpos i
 	return newAttachments, nil
 }
 
-// Retrieves an attachment given its key.
-func (db *Database) GetAttachment(key AttachmentKey) ([]byte, error) {
-	v, _, err := db.Bucket.GetRaw(attachmentKeyToString(key))
+// DeleteAttachmentVersion removes attachment versions from the AttachmentsMeta map specified.
+func DeleteAttachmentVersion(attachments AttachmentsMeta) (newAttachments AttachmentsMeta) {
+	newAttachments = attachments.ShallowCopy()
```
adamcfraser: Why do we need to make a shallow copy here?  It looks like it's being called based on the rev cache value, which is already a shallow copy.

adamcfraser (changes requested):

db/attachment.go
sarathkumarsivan: It was actually wrong, we need this handling for both legacy and new attachments.

db/attachment.go
sarathkumarsivan: This logic has been refactored. It assumes the attachment is always AttVersion2 while checking whether it has already been uploaded. CBG-1590 needs to be implemented to determine whether incoming attachment is AttVersion1 or AttVersion2.

db/blip_handler.go
adamcfraser: Reviewed with Sarath - makes sense for unit tests that aren't doing a decode with UseNumber.

rest/replicator_test.go
```diff
@@ -4657,14 +4662,15 @@ func TestLocalWinsConflictResolution(t *testing.T) {
...
-		},
+		// TODO: Enable this test once CBG-1590 is implemented.
+		// {
+		// 	// simpleMutation includes an attachment prior to conflict, validates it's preserved
+		// 	name:           "localAttachment",
+		// 	initialState:   newRevisionState(3, "a", false, 2),
+		// 	localMutation:  newRevisionState(4, "b", false, 0),
+		// 	remoteMutation: newRevisionState(4, "c", false, 0),
+		// 	expectedResult: newRevisionState(5, "b", false, 3), // revpos==3 here because the revision isn't replicated until rev 3
+		// },
```
adamcfraser: Can you share the specifics of why these conflict resolution tests are failing under this PR and require CBG-1590, but are passing on master?  I would have expected at most an additional GetAttachment call, but nothing that would break replication.

adamcfraser (changes requested):

rest/replicator_test.go
sarathkumarsivan: Sure Adam, the scenario is like this:

1. Create two SG instances rt1 and rt2 and start pull replication from rt2 to rt1.
2. Create initial revision of the document with an attachment on rt2
{"_attachments":{"initialAtt.txt":{"data":"aW5pdGlhbA=="}},"channels":["alice"]}
3. Wait for the document revision to get replicated to rt1.
4. Stop the replication process on rt1 and update the document by removing the attachment to create a local conflicting revision.
 {"source":"rt1","channels":["alice"]}
5. Create the remote conflicting revision by updating the doc on rt2 but keep the attachment stub.
 {"source":"rt2","_attachments":{"initialAtt.txt":{"stub":true,"revpos":1}},"channels":["alice"]}
6. Starts the replicator to trigger conflict resolution to merge both attachments in the conflict.

When rt1 receives the rev message, it tries to download the attachments. Subsequently rt1 identifies that the attachment is already uploaded and asks rt2 to prove it has the attachment without sending it. At that point, it panics and fails to prove the existence of the attachment — The issue is that in handleProveAttachment,  we're alway trying to grab the attachment by using the legacy format key, but the attachment is stored in the new storage format.

db/blip_handler.go
```diff
@@ -1068,7 +1071,7 @@ func (bh *blipHandler) sendProveAttachment(sender *blip.Sender, docID, name, dig
 // For each attachment in the revision, makes sure it's in the database, asking the client to
 // upload it if necessary. This method blocks until all the attachments have been processed.
 func (bh *blipHandler) downloadOrVerifyAttachments(sender *blip.Sender, body Body, minRevpos int, docID string) error {
-	return bh.db.ForEachStubAttachment(body, minRevpos,
+	return bh.db.ForEachStubAttachment(body, minRevpos, docID,
```
adamcfraser, line 1092: As discussed, as a defensive fallback for the case where we issue a proveAttachment request against SG (ISGR) for a 3.0 attachment, we should return a typed 'AttachmentNotFound' error.  We should check for that error here on line 1088, and fall back to getAttachment handling.

db/blip_handler.go
```diff
@@ -931,7 +931,7 @@ func (bh *blipHandler) handleProveAttachment(rq *blip.Message) error {
 		return base.HTTPErrorf(http.StatusBadRequest, "no digest sent with proveAttachment")
 	}
 
-	attData, err := bh.db.GetAttachment(AttachmentKey(digest))
+	attData, err := bh.db.GetAttachment(base.AttPrefix + digest)
```
adamcfraser, line 940: If we get an error here, check bh.clientType to see if it's ISGR.  If it's ISGR, return the 'AttachmentNotFound' error here.

adamcfraser (changes requested):

db/blip_handler.go
sarathkumarsivan: Sure, that makes sense. Added changes to check for bh.clientType == BLIPClientTypeSGR2 and return ErrAttachmentNotFound.

db/blip_handler.go
sarathkumarsivan: Sure, this change has been committed.

adamcfraser (changes requested):

db/blip_handler.go
```diff
@@ -1057,6 +1066,18 @@ func (bh *blipHandler) sendProveAttachment(sender *blip.Sender, docID, name, dig
...
+		resp.Properties["Error-Domain"] == "HTTP" &&
+		resp.Properties["Error-Code"] == "404" {
+		return ErrAttachmentNotFound
+	}
+
+	if resp.Type() == blip.ErrorType &&
+		resp.Properties["Error-Domain"] == blip.BLIPErrorDomain &&
+		resp.Properties["Error-Code"] == "500" {
+		return ErrAttachmentNotFound
+	}
```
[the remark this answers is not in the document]
sarathkumarsivan: Committed changest to avoid treating any 500 error as ErrAttachmentNotFound.

db/blip_handler.go
```diff
@@ -942,6 +942,9 @@ func (bh *blipHandler) handleProveAttachment(rq *blip.Message) error {
 		if bh.clientType == BLIPClientTypeSGR2 {
 			return ErrAttachmentNotFound
 		}
+		if IsMissingDDocError(err) {
```
adamcfraser: This is intended for checking missing design docs (ddoc) - you want to use IsKeyNotFound, I think.

adamcfraser (changes requested):

adamcfraser (approved):

db/blip_handler.go
sarathkumarsivan: Sure, that makese sense, I didn't realize this earlier, thank you, Adam.

PR #5093, opened 2021-07-12
CBG-1533: Specify permissions for each handler

JRascagneres:
Changes:

- Added new 'Permission' type to facilitate easy conversion between basic name and name required when requesting permission from cluster. These require formatting with cluster! or cluster.bucket[bucketName]!.
- Removed unused Vaccum endpoint
- Added permission requirements to each endpoint
- Added response permission check for /_user and /_role
- Set all response permissions to true if admin auth disabled

- [x] Merge https://github.com/couchbase/sync_gateway/pull/5086 and rebase

Permissions used are as follows:
https://docs.google.com/spreadsheets/d/1cE2MptvrcFF2cVyIGphwWuGERRMrKhuXD5kl8UgqOCY/edit?usp=sharing

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -716,8 +716,8 @@ func (h *handler) getUserInfo() error {
 		}
 		return err
 	}
-
-	_, includeDynamicGrantInfo := h.permissionsResults[PermReadPrincipalAppData.PermissionName]
+	// If not specified will default to false
+	includeDynamicGrantInfo := h.permissionsResults[PermReadPrincipalAppData.PermissionName]
 	bytes, err := marshalPrincipal(user, includeDynamicGrantInfo)
 	h.writeRawJSON(bytes)
 	return err
@@ -732,7 +732,8 @@ func (h *handler) getRoleInfo() error {
 		}
 		return err
 	}
-	_, includeDynamicGrantInfo := h.permissionsResults[PermReadPrincipalAppData.PermissionName]
+	// If not specified will default to false
+	includeDynamicGrantInfo := h.permissionsResults[PermReadPrincipalAppData.PermissionName]
 	bytes, err := marshalPrincipal(role, includeDynamicGrantInfo)
 	_, _ = h.response.Write(bytes)
 	return err
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -610,10 +610,10 @@ func marshalPrincipal(princ auth.Principal, includeDynamicGrantInfo bool) ([]byt
 	if user, ok := princ.(auth.User); ok {
 		info.Email = user.Email()
 		info.Disabled = user.Disabled()
-		info.ExplicitRoleNames = user.ExplicitRoles().AllChannels()
+		info.ExplicitRoleNames = user.ExplicitRoles().AllKeys()
 		if includeDynamicGrantInfo {
 			info.Channels = user.InheritedChannels().AsSet()
-			info.RoleNames = user.RoleNames().AllChannels()
+			info.RoleNames = user.RoleNames().AllKeys()
 		}
 	} else {
 		if includeDynamicGrantInfo {
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -601,20 +601,24 @@ func externalUserName(name string) string {
 	return name
 }
 
-func marshalPrincipal(princ auth.Principal) ([]byte, error) {
+func marshalPrincipal(princ auth.Principal, includeDynamicGrantInfo bool) ([]byte, error) {
 	name := externalUserName(princ.Name())
 	info := db.PrincipalConfig{
 		Name:             &name,
 		ExplicitChannels: princ.ExplicitChannels().AsSet(),
 	}
 	if user, ok := princ.(auth.User); ok {
-		info.Channels = user.InheritedChannels().AsSet()
 		info.Email = user.Email()
 		info.Disabled = user.Disabled()
 		info.ExplicitRoleNames = user.ExplicitRoles().AllChannels()
-		info.RoleNames = user.RoleNames().AllChannels()
+		if includeDynamicGrantInfo {
+			info.Channels = user.InheritedChannels().AsSet()
+			info.RoleNames = user.RoleNames().AllChannels()
+		}
 	} else {
-		info.Channels = princ.Channels().AsSet()
+		if includeDynamicGrantInfo {
+			info.Channels = princ.Channels().AsSet()
+		}
 	}
 	return base.JSONMarshal(info)
 }
@@ -713,7 +717,8 @@ func (h *handler) getUserInfo() error {
 		return err
 	}
 
-	bytes, err := marshalPrincipal(user)
+	_, includeDynamicGrantInfo := h.permissionsResults[PermReadPrincipalAppData.PermissionName]
+	bytes, err := marshalPrincipal(user, includeDynamicGrantInfo)
 	h.writeRawJSON(bytes)
 	return err
 }
@@ -727,7 +732,8 @@ func (h *handler) getRoleInfo() error {
 		}
 		return err
 	}
-	bytes, err := marshalPrincipal(role)
+	_, includeDynamicGrantInfo := h.permissionsResults[PermReadPrincipalAppData.PermissionName]
+	bytes, err := marshalPrincipal(role, includeDynamicGrantInfo)
 	_, _ = h.response.Write(bytes)
 	return err
 }
--- a/rest/routing.go
+++ b/rest/routing.go
@@ -43,34 +43,34 @@ func createCommonRouter(sc *ServerContext, privs handlerPrivs) (*mux.Router, *mu
 	r.Handle("/", makeHandler(sc, privs, nil, nil, (*handler).handleRoot)).Methods("GET", "HEAD")
 
 	// Operations on databases:
-	r.Handle("/{db:"+dbRegex+"}/", makeOfflineHandler(sc, privs, nil, nil, (*handler).handleGetDB)).Methods("GET", "HEAD")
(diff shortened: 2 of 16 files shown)


rest/api_test.go
```diff
@@ -7719,3 +7719,114 @@ func TestDocChannelSetPruning(t *testing.T) {
 	assert.Equal(t, uint64(1), syncData.ChannelSetHistory[0].Start)
 	assert.Equal(t, uint64(12), syncData.ChannelSetHistory[0].End)
 }
+
+func TestAdminAPIAuth(t *testing.T) {
```
bbrks: Should this belong in `admin_api_test.go` or `admin_api_auth_test.go` if we anticipate more to be written.

```suggestion
func TestAdminAPIAuthNoAccess(t *testing.T) {
```

rest/routing.go
```diff
@@ -28,6 +29,56 @@ const docRegex = "[^_/][^/]*"
 //  - A user name with an escaped "/" character
 var docWithSlashPathRegex *regexp.Regexp
 
+// Permission stores the name of a permission along whether it is database scoped. This is used to later obtain a
+// formatted permission string for checking.
+type Permission struct {
+	PermissionName string
+	DatabaseScoped bool
+}
```
bbrks: Maybe move to a new `routing_permissions.go` and move the list of permission variables up to the top so they're easier to find?

rest/routing.go
```diff
@@ -28,6 +29,56 @@ const docRegex = "[^_/][^/]*"
...
+
+func (perm *Permission) FormattedName(bucketName string) string {
+	if perm.DatabaseScoped {
+		return fmt.Sprintf("cluster.bucket[%s]!%s", bucketName, perm.PermissionName)
+	}
+	return fmt.Sprintf("cluster!%s", perm.PermissionName)
+}
+
+func FormatPermissionNames(perms []Permission, bucketName string) (formattedPerms []string) {
+	for _, perm := range perms {
```
bbrks:
```suggestion
	formattedPerms = make([]string, 0, len(perms))
	for _, perm := range perms {
```

rest/routing.go
```diff
@@ -28,6 +29,56 @@ const docRegex = "[^_/][^/]*"
...
+	}
+	return formattedPerms
+}
+
+func GetPermissionsNameFromFormatted(formattedName string) string {
+	return strings.Split(formattedName, "!")[1]
+}
+
+func GetPermissionNameFromFormattedStrings(formattedNames []string) (perms []string) {
+	for _, formattedName := range formattedNames {
```
bbrks:
```suggestion
	perms = make([]string, 0, len(formattedNames))
	for _, formattedName := range formattedNames {
```

rest/server_context_test.go
```diff
@@ -719,6 +724,6 @@ func TestAdminAuthWithX509(t *testing.T) {
 	managementEndpoints, httpClient, err := ctx.ObtainManagementEndpointsAndHTTPClient()
 	require.NoError(t, err)
 
-	_, _, err = checkAdminAuth("", base.TestClusterUsername(), base.TestClusterPassword(), httpClient, managementEndpoints, []string{"cluster!admin"}, nil)
+	_, _, err = checkAdminAuth("", base.TestClusterUsername(), base.TestClusterPassword(), httpClient, managementEndpoints, []Permission{{"cluster", false}}, nil)
```
bbrks: Shouldn't this be the admin permission scoped to the cluster? This is going to end up as `cluster!cluster` when formatted.

Do we need to assert more than just NoError to catch this?

```suggestion
	_, _, err = checkAdminAuth("", base.TestClusterUsername(), base.TestClusterPassword(), httpClient, managementEndpoints, []Permission{{"admin", false}}, nil)
```

bbrks (changes requested):
general approach seems reasonable

rest/api_test.go
JRascagneres: Not renamed test yet as it'll likely include non-noaccess perms but point taken.

rest/admin_api.go
```diff
@@ -713,7 +717,8 @@ func (h *handler) getUserInfo() error {
 		return err
 	}
 
-	bytes, err := marshalPrincipal(user)
+	_, includeDynamicGrantInfo := h.permissionsResults[PermReadPrincipalAppData.PermissionName]
```
bbrks: This returns `true` if the permission exists but is explicitly `false` ... (`val, ok`)

rest/admin_api.go
```diff
@@ -727,7 +732,8 @@ func (h *handler) getRoleInfo() error {
 		}
 		return err
 	}
-	bytes, err := marshalPrincipal(role)
+	_, includeDynamicGrantInfo := h.permissionsResults[PermReadPrincipalAppData.PermissionName]
```
bbrks: ditto

bbrks (changes requested):

rest/admin_api.go
JRascagneres: 🤦

rest/admin_api.go
```diff
@@ -601,20 +601,24 @@ func externalUserName(name string) string {
...
 	}
 	if user, ok := princ.(auth.User); ok {
-		info.Channels = user.InheritedChannels().AsSet()
 		info.Email = user.Email()
 		info.Disabled = user.Disabled()
 		info.ExplicitRoleNames = user.ExplicitRoles().AllChannels()
-		info.RoleNames = user.RoleNames().AllChannels()
+		if includeDynamicGrantInfo {
+			info.Channels = user.InheritedChannels().AsSet()
+			info.RoleNames = user.RoleNames().AllChannels()
```
adamcfraser: What do you think about refactoring timedSet.AllChannels() to timedSet.AllKeys() in this PR?  This isn't the first time I've spent a few minutes trying to work out why we're calling AllChannels() on a set of roles.

rest/routing.go
```diff
@@ -136,160 +136,159 @@ func CreateAdminRouter(sc *ServerContext) *mux.Router {
...
 
 	dbr.Handle("/_user/",
-		makeHandler(sc, adminPrivs, nil, nil, (*handler).getUsers)).Methods("GET", "HEAD")
+		makeHandler(sc, adminPrivs, []Permission{PermReadPrincipal}, []Permission{PermReadPrincipal}, (*handler).getUsers)).Methods("GET", "HEAD")
 	dbr.Handle("/_user/",
-		makeHandler(sc, adminPrivs, nil, nil, (*handler).putUser)).Methods("POST")
+		makeHandler(sc, adminPrivs, []Permission{PermWritePrincipal}, nil, (*handler).putUser)).Methods("POST")
 	dbr.Handle("/_user/{name}",
-		makeHandler(sc, adminPrivs, nil, nil, (*handler).getUserInfo)).Methods("GET", "HEAD")
+		makeHandler(sc, adminPrivs, []Permission{PermReadPrincipal, PermReadPrincipalAppData}, []Permission{PermReadPrincipal, PermReadPrincipalAppData}, (*handler).getUserInfo)).Methods("GET", "HEAD")
```
adamcfraser: I expected only PermReadPrincipal to be in the access permissions, and only PermReadPrincipalAppData in the response permissions.

rest/routing.go
```diff
@@ -136,160 +136,159 @@ func CreateAdminRouter(sc *ServerContext) *mux.Router {
...
 	dbr.Handle("/_user/{name}/_session",
-		makeHandler(sc, adminPrivs, nil, nil, (*handler).deleteUserSessions)).Methods("DELETE")
+		makeHandler(sc, adminPrivs, []Permission{PermWritePrincipal}, nil, (*handler).deleteUserSessions)).Methods("DELETE")
 	dbr.Handle("/_user/{name}/_session/{sessionid}",
-		makeHandler(sc, adminPrivs, nil, nil, (*handler).deleteUserSession)).Methods("DELETE")
+		makeHandler(sc, adminPrivs, []Permission{PermWritePrincipal}, nil, (*handler).deleteUserSession)).Methods("DELETE")
 
 	dbr.Handle("/_role/",
-		makeHandler(sc, adminPrivs, nil, nil, (*handler).getRoles)).Methods("GET", "HEAD")
+		makeHandler(sc, adminPrivs, []Permission{PermReadPrincipal, PermReadPrincipalAppData}, []Permission{PermReadPrincipal, PermReadPrincipalAppData}, (*handler).getRoles)).Methods("GET", "HEAD")
```
adamcfraser: I expected this to be just readPrincipal, similar to /_user/, as it's also just returning a list of IDs.

rest/routing.go
```diff
@@ -136,160 +136,159 @@ func CreateAdminRouter(sc *ServerContext) *mux.Router {
...
 	dbr.Handle("/_role/",
-		makeHandler(sc, adminPrivs, nil, nil, (*handler).putRole)).Methods("POST")
+		makeHandler(sc, adminPrivs, []Permission{PermWritePrincipal}, nil, (*handler).putRole)).Methods("POST")
 	dbr.Handle("/_role/{name}",
-		makeHandler(sc, adminPrivs, nil, nil, (*handler).getRoleInfo)).Methods("GET", "HEAD")
+		makeHandler(sc, adminPrivs, []Permission{PermReadPrincipal, PermReadPrincipalAppData}, []Permission{PermReadPrincipal, PermReadPrincipalAppData}, (*handler).getRoleInfo)).Methods("GET", "HEAD")
```
adamcfraser: Same question as above (expected read for access, read..appdata for response).

adamcfraser (changes requested):
Generally looks good, a couple of comments.

rest/routing.go
JRascagneres: Yup this was my mistake.

rest/admin_api.go
JRascagneres: Yeah that makes sense. To be honest reading this again it took me a sec to realize that wasn't getting the channels on a role...

adamcfraser (approved):

PR #5094, opened 2021-07-13
CBG-1553 - Inherit migrated 2.x configs from a DefaultStartupConfig

bbrks:
In-memory migrated 2.x configs were not setting the default values appropriately, causing things like stats logging to be disabled.

This PR covers all of the node-level default config by merging the 2.x config options onto a DefaultStartupConfig in the same way the 3.0 config does.

The default config options for databases, like cache sizing, etc. is covered by existing code.


rest/main_legacy.go
```diff
@@ -21,12 +22,19 @@ func legacyServerMain(osArgs []string) error {
...
 
-	sc, databases, err := lc.ToStartupConfig()
+	sc := DefaultStartupConfig(defaultLogFilePath)
+
+	migratedStartupConfig, databases, err := lc.ToStartupConfig()
+	if err != nil {
+		return err
+	}
+
+	err = mergo.Merge(&sc, migratedStartupConfig, mergo.WithOverride)
```
adamcfraser: Are there any scenarios where DefaultStartupConfig sets a non-zero/non-empty value, and setting that to an empty value in migratedStartupConfig is valid?  Are we guarding against that by defining all such values as pointers?  (my other comment may be related to this)

rest/config_startup.go
```diff
@@ -145,8 +138,8 @@ type ReplicatorConfig struct {
 }
 
 type UnsupportedConfig struct {
-	StatsLogFrequency *base.ConfigDuration `json:"stats_log_frequency,omitempty" help:"How often should stats be written to stats logs"`
-	UseStdlibJSON     bool                 `json:"use_stdlib_json,omitempty"     help:"Bypass the jsoniter package and use Go's stdlib instead"`
+	StatsLogFrequency base.ConfigDuration `json:"stats_log_frequency,omitempty" help:"How often should stats be written to stats logs"`
```
adamcfraser: I'm not following how we differentiate between stats_log_frequency being absent vs. set to zero with this change.

bbrks: Yeah you're right. I shouldn't have changed this to a non-pointer for that reason.

adamcfraser (dismissed):

PR #5097, opened 2021-07-14
CBG-1378 - OIDC client will be initalized until successful

IsaacLambat:
- [x]  http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/901/ (1 test failed, runs locally fine, and on integration and EE unit tests so assumed to be racy test).

Sync.once behaviour is copied except the atomic int is only set to 1 when there is no errors from initializing the OIDC client.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -649,9 +649,9 @@ func (auth *Authenticator) AuthenticateUntrustedJWT(token string, providers OIDC
 // Returns identity claims extracted from the token if the verification is successful and an identity error if not.
 func verifyToken(token string, provider *OIDCProvider, callbackURLFunc OIDCCallbackURLFunc) (identity *Identity, err error) {
 	// Get client for issuer
-	client := provider.GetClient(callbackURLFunc)
-	if client == nil {
-		return nil, fmt.Errorf("OIDC client was not initialized")
+	client, err := provider.GetClient(callbackURLFunc)
+	if err != nil {
+		return nil, fmt.Errorf("OIDC initialization error: %v", err)
 	}
 
 	// Verify claims and signature on the JWT; ensure that it's been signed by the provider.
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -651,7 +651,7 @@ func verifyToken(token string, provider *OIDCProvider, callbackURLFunc OIDCCallb
 	// Get client for issuer
 	client, err := provider.GetClient(callbackURLFunc)
 	if err != nil {
-		return nil, fmt.Errorf("OIDC initialization error: %v", err)
+		return nil, fmt.Errorf("OIDC initialization error: %w", err)
 	}
 
 	// Verify claims and signature on the JWT; ensure that it's been signed by the provider.
--- a/auth/oidc.go
+++ b/auth/oidc.go
@@ -167,10 +167,13 @@ type OIDCProvider struct {
 	// use the accessor method GetClient() instead.
 	client *OIDCClient
 
-	// clientOnce synchronises access to the GetClient() and ensures that
+	// clientInitLock synchronises access to the GetClient() and ensures that
 	// the OpenID Connect client only gets initialized exactly once when
-	// successfully connected to the OIDC provider
-	clientOnce base.AtomicBool
+	// the client has been successfully initialized.
+	clientInitLock sync.Mutex
+
+	// clientInit tracks whether the client has been successfully initialized or not
+	clientInit base.AtomicBool
 
 	// IsDefault indicates whether this OpenID Connect provider (the current
 	// instance of OIDCProvider is explicitly specified as default provider
@@ -246,20 +249,25 @@ func (opm OIDCProviderMap) Stop() {
 func (op *OIDCProvider) GetClient(buildCallbackURLFunc OIDCCallbackURLFunc) *OIDCClient {
 	// Initialize the client on first request. If the callback URL isn't defined for the provider,
 	// uses buildCallbackURLFunc to construct (based on current request)
-	if op.clientOnce.CompareAndSwap(false, true) { // If false, make true
-		var err error
-		// If the redirect URL is not defined for the provider generate it from the
-		// handler request and set it on the provider
-		if op.CallbackURL == nil || *op.CallbackURL == "" {
-			callbackURL := buildCallbackURLFunc(op.Name, op.IsDefault)
-			if callbackURL != "" {
-				op.CallbackURL = &callbackURL
+	if !op.clientInit.IsTrue() { // If client not initialized
+		op.clientInitLock.Lock()
+		if !op.clientInit.IsTrue() {
+			var err error
+			// If the redirect URL is not defined for the provider generate it from the
+			// handler request and set it on the provider
+			if op.CallbackURL == nil || *op.CallbackURL == "" {
+				callbackURL := buildCallbackURLFunc(op.Name, op.IsDefault)
+				if callbackURL != "" {
+					op.CallbackURL = &callbackURL
+				}
+			}
+			if err = op.initOIDCClient(); err != nil {
+				base.Errorf("Unable to initialize OIDC client: %v", err)
+			} else {
+				op.clientInit.Set(true)
 			}
 		}
-		if err = op.initOIDCClient(); err != nil {
-			base.Errorf("Unable to initialize OIDC client: %v", err)
-			op.clientOnce.Set(false)
-		}
+		op.clientInitLock.Unlock()
 	}
 	return op.client
 }
--- a/auth/oidc.go
+++ b/auth/oidc.go
@@ -168,8 +168,9 @@ type OIDCProvider struct {
 	client *OIDCClient
 
 	// clientOnce synchronises access to the GetClient() and ensures that
-	// the OpenID Connect client only gets initialized exactly once.
-	clientOnce sync.Once
+	// the OpenID Connect client only gets initialized exactly once when
+	// successfully connected to the OIDC provider
+	clientOnce base.AtomicBool
 
 	// IsDefault indicates whether this OpenID Connect provider (the current
 	// instance of OIDCProvider is explicitly specified as default provider
@@ -245,7 +246,7 @@ func (opm OIDCProviderMap) Stop() {
 func (op *OIDCProvider) GetClient(buildCallbackURLFunc OIDCCallbackURLFunc) *OIDCClient {
 	// Initialize the client on first request. If the callback URL isn't defined for the provider,
 	// uses buildCallbackURLFunc to construct (based on current request)
-	op.clientOnce.Do(func() {
+	if op.clientOnce.CompareAndSwap(false, true) { // If false, make true
 		var err error
 		// If the redirect URL is not defined for the provider generate it from the
 		// handler request and set it on the provider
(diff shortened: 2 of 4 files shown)


auth/oidc.go
```diff
@@ -245,7 +246,7 @@ func (opm OIDCProviderMap) Stop() {
 func (op *OIDCProvider) GetClient(buildCallbackURLFunc OIDCCallbackURLFunc) *OIDCClient {
 	// Initialize the client on first request. If the callback URL isn't defined for the provider,
 	// uses buildCallbackURLFunc to construct (based on current request)
-	op.clientOnce.Do(func() {
+	if op.clientOnce.CompareAndSwap(false, true) { // If false, make true
```
IsaacLambat: Could instead check if false, and then change to true when the OIDC client is set successfully so if this is called concurrently, before the OIDC Client is initialized, then nil is not returned.

The current CompareAndSwap way more emulates the old behaviour with the difference that it can be retried if the client failed to get initialized. 

Would like to hear your thoughts

bbrks: I don't think this follows the old behaviour.

`sync.Once` blocks concurrent requests until the function has finished, allowing all of the requests to use a client once the first has initialised it.

I think this blocking is the behaviour we _want_ to do, as even checking for false is not sufficinet to allow simultaenous requests to initialise and use a client for all of the requests.

The comment in CBG-1378 describes the approach we want to take, using a combination of atomic and mutex, in the same way that sync.Once is doing under the hood.

bbrks (dismissed):
Need to revisit approach to follow the suggestion in the Jira comment - as this doesn't sufficiently handle concurrent requests in the way we want.

auth/oidc.go
adamcfraser: A key feature of sync.Once.Do() is that concurrent invocations will block until the initial call completes.  We'll need to maintain the same functionality with the replacement.  The approach you've got here doesn't address that scenario, as far as I can tell.

adamcfraser (changes requested):

bbrks (commented):
Looks good 👍 will let @adamcfraser take a final look

auth/oidc.go
```diff
@@ -245,21 +249,26 @@ func (opm OIDCProviderMap) Stop() {
...
+			// If the redirect URL is not defined for the provider generate it from the
+			// handler request and set it on the provider
+			if op.CallbackURL == nil || *op.CallbackURL == "" {
+				callbackURL := buildCallbackURLFunc(op.Name, op.IsDefault)
+				if callbackURL != "" {
+					op.CallbackURL = &callbackURL
+				}
+			}
+			if err = op.initOIDCClient(); err != nil {
+				base.Errorf("Unable to initialize OIDC client: %v", err)
```
adamcfraser: As discussed, let's change GetClient to return (*OIDCClient, error), and have callers check for error instead of nil client.  With that in place can remove the Errorf logging here and rely on the callers to log appropriately (review existing callers to include the err when logging).

adamcfraser (changes requested):

auth/auth.go
```diff
@@ -649,9 +649,9 @@ func (auth *Authenticator) AuthenticateUntrustedJWT(token string, providers OIDC
 // Returns identity claims extracted from the token if the verification is successful and an identity error if not.
 func verifyToken(token string, provider *OIDCProvider, callbackURLFunc OIDCCallbackURLFunc) (identity *Identity, err error) {
 	// Get client for issuer
-	client := provider.GetClient(callbackURLFunc)
-	if client == nil {
-		return nil, fmt.Errorf("OIDC client was not initialized")
+	client, err := provider.GetClient(callbackURLFunc)
+	if err != nil {
+		return nil, fmt.Errorf("OIDC initialization error: %v", err)
```
bbrks: As of Go 1.13, you can use `%w` (wrap) in `fmt.Errorf`, which behaves the same as `%v` but will also let you access the underlying error using `errors.Is/errors.As`

```suggestion
		return nil, fmt.Errorf("OIDC initialization error: %w", err)
```

auth/oidc.go
```diff
@@ -242,25 +246,34 @@ func (opm OIDCProviderMap) Stop() {
...
+
+	if op.clientInit.IsTrue() { // If client not initialized
+		return op.client, nil
+	}
+	// Acquire lock
+	op.clientInitLock.Lock()
+	defer op.clientInitLock.Unlock()
 
-	return op.client
+	if op.clientInit.IsTrue() { // Make sure past thread didn't init in lock
```
bbrks: This is a good example of a "why" comment :) but advise we keep comments on their own lines, it's easy to lose them when reading if they're tacked on to the end of code.

```suggestion
	// Check again to see if the previous lock holder initialized the client
	if op.clientInit.IsTrue() {
```

rest/oidc_api.go
```diff
@@ -95,10 +95,11 @@ func (h *handler) handleOIDCCommon() (redirectURLString string, err error) {
...
 	}
 
-	client := provider.GetClient(h.getOIDCCallbackURL)
-	if client == nil {
+	client, err := provider.GetClient(h.getOIDCCallbackURL)
+	if err != nil {
+		base.Errorf("OIDC initialization error: %v", err)
 		return redirectURLString, base.HTTPErrorf(
-			http.StatusInternalServerError, fmt.Sprintf("Unable to obtain client for provider:%s", providerName))
+			http.StatusInternalServerError, fmt.Sprintf("Unable to obtain client for provider: %s", providerName))
```
bbrks: Include error message for the user making the request, now that we have access to it

```suggestion
			http.StatusInternalServerError, fmt.Sprintf("Unable to obtain client for provider: %s - %v", providerName, err))
```

rest/oidc_api.go
```diff
@@ -168,9 +169,9 @@ func (h *handler) handleOIDCCallback() error {
 		http.SetCookie(h.response, stateCookie)
 	}
 
-	client := provider.GetClient(h.getOIDCCallbackURL)
-	if client == nil {
-		return err
+	client, err := provider.GetClient(h.getOIDCCallbackURL)
+	if err != nil {
+		return fmt.Errorf("OIDC initialization error: %v", err)
```
bbrks:
```suggestion
		return fmt.Errorf("OIDC initialization error: %w", err)
```

rest/oidc_api.go
```diff
@@ -221,9 +222,9 @@ func (h *handler) handleOIDCRefresh() error {
 		return base.HTTPErrorf(http.StatusBadRequest, "Unable to identify provider for callback request")
 	}
 
-	client := provider.GetClient(h.getOIDCCallbackURL)
-	if client == nil {
-		return err
+	client, err := provider.GetClient(h.getOIDCCallbackURL)
+	if err != nil {
+		return fmt.Errorf("OIDC initialization error: %v", err)
```
bbrks:
```suggestion
		return fmt.Errorf("OIDC initialization error: %w", err)
```

rest/oidc_api.go
```diff
@@ -95,10 +95,11 @@ func (h *handler) handleOIDCCommon() (redirectURLString string, err error) {
 		return redirectURLString, err
 	}
 
-	client := provider.GetClient(h.getOIDCCallbackURL)
-	if client == nil {
+	client, err := provider.GetClient(h.getOIDCCallbackURL)
+	if err != nil {
+		base.Errorf("OIDC initialization error: %v", err)
```
bbrks: We can be rely on the normal HTTP error logging now, and remove this.

```suggestion
```

HTTP error logs appear like this:

```
2021-07-16T15:18:22.307+01:00 [INF] HTTP:  #010: PUT /db1/ (as ADMIN)
2021-07-16T15:19:16.816+01:00 [ERR] #010: 500 couldn't load database: couldn't reload database: 502  Unable to connect to Couchbase Server (connection refused). Please ensure it is running and reachable at the configured host and port.  Detailed error: no access -- rest.(*handler).writeError() at handler.go:890
2021-07-16T15:19:16.816+01:00 [INF] HTTP: #010:     --> 500 couldn't load database: couldn't reload database: 502  Unable to connect to Couchbase Server (connection refused). Please ensure it is running and reachable at the configured host and port.  Detailed error: no access  (54508.3 ms)
```

bbrks (changes requested):
Only a small improvement to the REST API response (to include the now available error) and some general rules about code comments that I try and follow (why not what).

rest/oidc_api.go
IsaacLambat: Will do. Originally didn't do this because I thought the error message wasn't the most helpful but on second thoughts it could help.
Example error message when OIDC Provider Unreachable: `Unable to obtain client for provider: foo - unable to discover config: unsuccessful response: <nil>`

bbrks: Ah right - yeah that `<nil>` is a bit misleading. Looks like another usage of an incorrect "err" variable...

```go
	if resp.StatusCode != http.StatusOK {
		return ProviderMetadata{}, MaxProviderConfigSyncInterval, false, fmt.Errorf("unsuccessful response: %v", err)
	}
```

I think there we should include the repsonse body, in the event that the provider is giving us more detail in the error repsonse.

rest/oidc_api_test.go
```diff
@@ -2090,3 +2095,85 @@ func TestOpenIDConnectAuthCodeFlowWithUsernameClaim(t *testing.T) {
 		})
 	}
 }
+
+// CBG-1378 - test when OIDC provider is not reachable, and then becomes reachable
+// at a later request
+func TestEventuallyReachableOIDCClient(t *testing.T) {
+	// Modified copy of TestOpenIDConnectImplicitFlow
+	unreachableAddr := "0.0.0.0:1234"
```
bbrks: It looks like this needs to be a HTTP URL, not just an address. I think the test works but not for the reason you're expecting.

```
2021-07-20T13:48:49.532+01:00 [INF] Auth: Fetching provider config from explicitly defined discovery endpoint: <ud>0.0.0.0:1234/foo/.well-known/openid-configuration</ud>
2021-07-20T13:48:49.532+01:00 [DBG] Auth+: Fetching custom provider config from <ud>0.0.0.0:1234/foo/.well-known/openid-configuration</ud>
2021-07-20T13:48:49.532+01:00 [DBG] Auth+: Error building new request for URL <ud>0.0.0.0:1234/foo/.well-known/openid-configuration</ud>: parse "0.0.0.0:1234/foo/.well-known/openid-configuration": first path segment in URL cannot contain colon
2021-07-20T13:48:49.532+01:00 [INF] HTTP:  #001: POST /db/_session
2021-07-20T13:48:49.532+01:00 [INF] HTTP: #001:     --> 401 Invalid login  (0.2 ms)
```

Note the errors in the middle there, related to building a URL to make the OIDC config request.

bbrks (changes requested):
Minor change required to test, but it should be a small one-line tweak.

rest/oidc_api_test.go
IsaacLambat: Nice spot.
Replaced address with `http://0.0.0.0` which now just refuses the connection. Could have also put a IP to time out (IE. 1.2.3.4), or a completely unreachable host (any URL that does not exist) if you think any of them seem more appropriate but ultimately leads to the same results.

bbrks (approved):
LGTM

adamcfraser (approved):

PR #5098, opened 2021-07-14
Added temporary flag overrides for a couple config options

JRascagneres:
- Added some temporary flags which allow us to override a few items which are required to run litecore tests with SGW 3.x
 - Auth ones are hooked up, TLS / Server ones will be hooked up as part of the PR that adds those options
 - Additionally we pass down flag options to legacy config in the same way we do for persistent config
 
 - [x] Merge https://github.com/couchbase/sync_gateway/pull/5094 and rebase


bbrks (dismissed):
LGTM

bbrks (approved):

PR #5103, opened 2021-07-15
Temp rename agent label for testing docker builders

bbrks:
- [ ] Get pipeline builds running through docker container (so we're able to run CI jobs on laptops and with more parallelism)


JRascagneres:
Closing PR as this work is 'on-hold' for now. Can always re-open when we have some free time to investigate.

PR #5104, opened 2021-07-16
CBG-1421 - Bootstrap.Server is now a required field

IsaacLambat:
- [ ] http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/890/

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -72,7 +72,7 @@ type BucketConfig struct {
 
 func (bc *BucketConfig) MakeBucketSpec() base.BucketSpec {
 
-	server := "http://localhost:8091"
+	server := ""
 	bucketName := ""
 	tlsPort := 11207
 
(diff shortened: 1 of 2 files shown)


rest/config_test.go
```diff
@@ -1265,6 +1265,7 @@ func TestSetupServerContext(t *testing.T) {
 	defer base.SetUpTestLogging(base.LevelInfo, base.KeyAll)()
 	t.Run("Create server context with a valid configuration", func(t *testing.T) {
 		config := DefaultStartupConfig("")
+		config.Bootstrap.Server = "walrus:" // Valid config requires server to be explicitly defined
```
JRascagneres: Should use `base.UnitTestUrl()` instead I think so this can work with CBS

JRascagneres (changes requested):

JRascagneres:
Needs a master rebase to pickup litecore fix.

rest/config.go
```diff
@@ -809,6 +809,13 @@ func (config *StartupConfig) Serve(addr string, handler http.Handler) error {
 	)
 }
 
+func (sc *StartupConfig) validateServerProvided() (err error) {
```
JRascagneres: Worth getting opinion from Mr Config @bbrks but if we're going to any other validation / checking on startup config then it may be worth using the same function / renaming to validationServerConfig.

IsaacLambat: Yes, would be best as my other PR uses it's own validation function as well.

JRascagneres: Ah okay, if you're doing that in another PR then its probably fine not to need brrks input. Go for it :)

JRascagneres (approved):
Looks fine to me

bbrks (approved):
LGTM

PR #5109, opened 2021-07-19
Store closable http.Server on ServerContext for graceful shutdown

bbrks:
Keep references to the underlying `http.Server` in `ServerContext` so they can be closed gracefully on `ServerContext.Close()`

Allows for "end-to-end" type testing (serverMain) for bootstrap config.

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -786,7 +786,7 @@ func SetMaxFileDescriptors(maxP *uint64) error {
 	return nil
 }
 
-func (config *StartupConfig) Serve(addr string, handler http.Handler) error {
+func (sc *ServerContext) Serve(config *StartupConfig, addr string, handler http.Handler) error {
 	http2Enabled := false
 	if config.Unsupported.HTTP2 != nil {
 		http2Enabled = *config.Unsupported.HTTP2.Enabled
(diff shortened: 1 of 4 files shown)


rest/config.go
```diff
@@ -807,6 +807,19 @@ func (config *StartupConfig) Serve(addr string, handler http.Handler) error {
 		http2Enabled,
 		tlsMinVersion,
 	)
+	if err != nil {
+		return err
+	}
+
+	sc.addHTTPServer(server)
+
```
adamcfraser, line 815: If someone called serverContext.Close at this point, it would attempt to Close() a server that hadn't run server.Serve() yet.  Is that a no-op/safe sequence of operations?

adamcfraser (commented):
Looks fine, one request for clarification on the interaction between close and serve.

rest/config.go
bbrks: Close is safe to call on a server that has not yet started. The server errors on startup after Close has been called. It's also safe to double close.

https://play.golang.org/p/o1CswqN55pe 

```go
func main() {
	s := http.Server{}
	// close before listen/serve
	fmt.Println(s.Close())
	// double close
	fmt.Println(s.Close())
	// listen/serve after close
	fmt.Println(s.ListenAndServe())
	fmt.Println(s.Close())
}
```
```
<nil>
<nil>
http: Server closed
<nil>
```

adamcfraser (approved):

PR #5112, opened 2021-07-20
CBG-1544: Handle SyncWriteInProgress as recoverable write error

JRascagneres:
Adds retry handling for `SyncWriteInProgress` error. Only occurs on write.
This is handled by us in gocbv1 by `isRecoverableWriteError`.
On gocbv2 this is done internally by gocb and they retry up to the timeout. Behaviour will be determined as part of CBG-1142

### DIFF
--- a/base/bucket_gocb.go
+++ b/base/bucket_gocb.go
@@ -810,8 +810,19 @@ func (bucket *CouchbaseBucketGoCB) isRecoverableWriteError(err error) bool {
 	}
 
 	_, ok := recoverableGocbV1Errors[pkgerrors.Cause(err).Error()]
+	if ok {
+		return ok
+	}
 
-	return ok
+	// In some circumstances we are unable to supply errors in the recoverable error map so need to check the error
+	// codes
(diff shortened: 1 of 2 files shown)


base/bucket_gocb.go
```diff
@@ -810,8 +810,19 @@ func (bucket *CouchbaseBucketGoCB) isRecoverableWriteError(err error) bool {
...
+	if ok {
+		return ok
+	}
 
-	return ok
+	// In some circumstances we are unable to supply errors in the recoverable error map so need to check the error
+	// codes
+	gocbcoreKvError, ok := err.(*gocbcore.KvError)
+	// 0xA2 error is "SyncWriteInProgress" error
+	if gocbcoreKvError.Code == 0xA2 {
```
adamcfraser: Can you use `memd.StatusSyncWriteInProgress` here instead of `0xA2`?  There's also already a kvError checking helper - base.isKvError - can you use that here?

adamcfraser (changes requested):

adamcfraser (approved):

PR #5113, opened 2021-07-21
CBG-1238 - Removed deprecated logging

IsaacLambat:
cgLevel now uses new log level type base.LogLevel
**Removed** 
base.Level and callers
EnableLogHTTPBodies and usages (LoggingTeeResponseWriter, TeeReadCloser)
LoggerWriter
PrependContextID (unused function)


base/logger_external.go
```diff
@@ -125,7 +125,7 @@ func (CBGoUtilsLogger) Level() logging.Level {
 	if consoleLogger == nil || consoleLogger.LogLevel == nil {
 		return logging.INFO
 	}
-	return ToDeprecatedLogLevel(*consoleLogger.LogLevel).cgLevel()
+	return consoleLogger.cgLevel()
```
bbrks: Given the usage of this is isolated now, it might be worth considering inlining this specifically for `CBGoUtilsLogger`, to avoid putting any old stuff on the `ConsoleLogger`.

```suggestion
	switch *consoleLogger.LogLevel {
	case LevelTrace:
		return logging.TRACE
	case LevelDebug:
		return logging.DEBUG
	case LevelInfo:
		return logging.INFO
	case LevelWarn:
		return logging.WARN
	case LevelError:
		return logging.ERROR
	}
	return logging.NONE
```

bbrks (changes requested):
Looks good, just one suggestion to inline and avoid introducing "old" logging stuff into ConsoleLogger.

bbrks (approved):
LGTM

PR #5114, opened 2021-07-21
Admin Auth Config options & cleanup

JRascagneres:
Adds config options and does some cleanup:
- Re-enable admin api and metrics api auth by default
- Add admin api and metrics api config options to the legacy config to allow opt-out in upgrade case
- Add `EnableAdminAuthenticationPermissionsCheck` option so user can opt-in to the permissions based check - Required due to this being a DP feature
- Tweak Permission format to include "!" as part of the name. This should be done as some stats are in a 'category' so you have stats such as cluster.bucket[db].stats!read. This means we could add a Permission with the name `.stats!read`. Unsure at the moment but this may be required in the SGW case, perhaps they will be in a category?
- Add request context to admin auth logging
- Add authorized users name to http logging where we currently have (as ADMIN)
- Move admin auth tests to specific file

- [x] http://uberjenkins.sc.couchbase.com:8080/view/All/job/sync-gateway-integration/910/


rest/handler.go
```diff
@@ -254,20 +255,22 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
...
+		permissions, statusCode, err := checkAdminAuth(authScope, username, password, httpClient, managementEndpoints,
+			*h.server.config.API.EnableAdminAuthenticationPermissionsCheck, accessPermissions, responsePermissions)
 		if err != nil {
 			base.Warnf("An error occurred whilst checking whether a user was authorized: %v", err)
 			return base.HTTPErrorf(http.StatusInternalServerError, "")
 		}
 
 		if statusCode != http.StatusOK {
-			base.Infof(base.KeyAuth, "User %s failed to auth as an admin statusCode: %d", username, statusCode)
+			base.Infof(base.KeyAuth, "%s: User %s failed to auth as an admin statusCode: %d", h.formatSerialNumber(), username, statusCode)
```
bbrks:
```suggestion
			base.Infof(base.KeyAuth, "%s: User %s failed to auth as an admin statusCode: %d", h.formatSerialNumber(), base.UD(username), statusCode)
```

rest/handler.go
```diff
@@ -254,20 +255,22 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
...
 			return base.HTTPErrorf(statusCode, "")
 		}
 
+		h.authorizedAdminUser = username
 		h.permissionsResults = permissions
 
-		base.Infof(base.KeyAuth, "User %s was successfully authorized as an admin", username)
+		base.Infof(base.KeyAuth, "%s: User %s was successfully authorized as an admin", h.formatSerialNumber(), username)
```
bbrks:
```suggestion
		base.Infof(base.KeyAuth, "%s: User %s was successfully authorized as an admin", h.formatSerialNumber(), base.UD(username))
```

rest/handler.go
```diff
@@ -681,13 +691,18 @@ func (h *handler) getBearerToken() string {
 // taggedEffectiveUserName returns the tagged effective name of the user for the request.
 // e.g: '<ud>alice</ud>' or 'GUEST'
 func (h *handler) taggedEffectiveUserName() string {
-	if h.privs == adminPrivs {
+	if h.authorizedAdminUser != "" {
+		return h.authorizedAdminUser + " as ADMIN"
```
bbrks:
```suggestion
		return base.UD(h.authorizedAdminUser).Redact() + " as ADMIN"
```

bbrks (changes requested):
Looks OK, need to apply user-data redaction in the logging though

bbrks (approved):

PR #5115, opened 2021-07-21
Renamed temporary flag override for LiteCore tests to their new names

IsaacLambat:
Renamed bootstrap.allow_insecure_server_connections to bootstrap.use_tls_server
Renamed api.https.allow_insecure_tls_connections to api.https.use_tls_client
Also changed the defaults as appropriate

### DIFF
--- a/rest/main.go
+++ b/rest/main.go
@@ -49,8 +49,8 @@ func serverMain(ctx context.Context, osArgs []string) error {
 	adminInterfaceAuthFlag := fs.Bool("api.admin_interface_authentication", false, "")
 	metricsInterfaceAuthFlag := fs.Bool("api.metrics_interface_authentication", false, "")
 
-	allowInsecureServerConnections := fs.Bool("bootstrap.allow_insecure_server_connections", false, "")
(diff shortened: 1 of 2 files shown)


rest/main.go
```diff
@@ -49,8 +49,8 @@ func serverMain(ctx context.Context, osArgs []string) error {
 	adminInterfaceAuthFlag := fs.Bool("api.admin_interface_authentication", false, "")
 	metricsInterfaceAuthFlag := fs.Bool("api.metrics_interface_authentication", false, "")
 
-	allowInsecureServerConnections := fs.Bool("bootstrap.allow_insecure_server_connections", false, "")
-	allowInsecureTLSConnections := fs.Bool("api.https.allow_insecure_tls_connections", false, "")
+	useTLSServer := fs.Bool("bootstrap.use_tls_server", true, "")
```
IsaacLambat, line 52: Wouldn't this being true cause problems by it overriding the config the user provided if the user has set it to false?
I think this because the flags override the config.

bbrks: can address this in the pr that uses these flag values

bbrks (approved):

PR #5118, opened 2021-07-23
CBG-1335: Use appropriate rev when utilizing a custom conflict resolver

JRascagneres:
Use original local rev and remote rev during conflict resolution except when calculating local/remote winner and for the merged body

Tested with unit test

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -962,7 +962,6 @@ func (db *Database) PutExistingRevWithConflictResolution(newDoc *Document, docHi
 		allowConflictingTombstone := forceAllowConflictingTombstone && doc.IsDeleted()
 
 		if !allowConflictingTombstone && db.IsIllegalConflict(doc, parent, newDoc.Deleted, noConflicts, docHistory) {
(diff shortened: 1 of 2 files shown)


db/crud.go
```diff
@@ -1094,6 +1093,14 @@ func (db *Database) resolveConflict(localDoc *Document, remoteDoc *Document, doc
 		return "", nil, resolveFuncError
 	}
 
+	// Restore rev ID onto localDoc and remoteDoc in the event that these have been mutated inside a custom conflict
+	// resolver
+	localDoc.RevID = localRevID
```
adamcfraser: The versions of localDoc and remoteDoc we're passing to the conflict resolver (inside Conflict) should already be distinct copies - I didn't think this reassignment would be necessary?

adamcfraser (changes requested):
Changes to not use ExtractRev post-resolution look good, but would like to understand why the reassignment is necessary.

adamcfraser (approved):

PR #5119, opened 2021-07-23
CBG-1401 Start background polling for new buckets/database configs

bbrks:
Starts a background polling loop on bootstrap that discovers new buckets and new/updated database configs based on the configured interval (default 10s)

### DIFF
--- a/base/util.go
+++ b/base/util.go
@@ -1557,3 +1557,19 @@ func (d *ConfigDuration) UnmarshalJSON(b []byte) error {
 	*d = ConfigDuration{Duration: dur}
 	return nil
 }
+
+// TerminateAndWaitForClose will close the given terminator, and wait up to timeout for the done channel to be closed in response.
+func TerminateAndWaitForClose(terminator chan struct{}, done chan struct{}, timeout time.Duration) error {
+	if terminator == nil && done == nil {
+		return errors.New("terminateAndWaitForClose requires both terminator and done channels")
+	}
+	close(terminator)
+	t := time.NewTimer(timeout)
+	select {
+	case <-done:
+		t.Stop()
+	case <-t.C:
+		return fmt.Errorf("terminateAndWaitForClose timed out waiting for done channel after %v", timeout)
+	}
+	return nil
+}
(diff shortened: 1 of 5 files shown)


rest/server_context.go
```diff
@@ -159,6 +165,8 @@ func (sc *ServerContext) Close() {
 	sc.lock.Lock()
 	defer sc.lock.Unlock()
 
+	sc.ctxCancel()
```
adamcfraser: Do we need to wait for the bootstrap polling loop to terminate before we continue with server context close?  Potentially prior to acquiring the sc.lock?  It seems like there's a potential race where fetchAndLoadConfigs has fires in the polling goroutine after we acquire the lock.  So applyConfigs is blocked on acquire sc.lock, but then runs after we've finished Close() and possibly recreates the databases we just closed.  Unlikely to be a problem in production as sc.Close() doesn't get called other than during shutdown, but seems like the kind of thing that will cause integration testing headaches.

adamcfraser (changes requested):

rest/server_context.go
bbrks: Yep makes sense, thanks. I'll take a closer look at this close behaviour once I'm back.

bbrks: Reworked this to align with the stats process, and added a util for the `terminator` -> `done` channel pattern we're using for these syncronous closes.

base/util.go
```diff
@@ -1557,3 +1557,19 @@ func (d *ConfigDuration) UnmarshalJSON(b []byte) error {
 	*d = ConfigDuration{Duration: dur}
 	return nil
 }
+
+// TerminateAndWaitForClose will close the given terminator, and wait up to timeout for the done channel to be closed in response.
+func TerminateAndWaitForClose(terminator chan struct{}, done chan struct{}, timeout time.Duration) error {
+	if terminator == nil && done == nil {
```
adamcfraser: Should be or?

```suggestion
	if terminator == nil || done == nil {
```

adamcfraser (changes requested):

adamcfraser (dismissed):

PR #5122, opened 2021-07-26
CBG-1518 - Fixed failing unit tests on Windows

IsaacLambat:
- Fixed the failing unit tests on Windows
- Added new test case to test when no CA provided and TLS skip verify is false (so Root pool will be retrieved except on Windows).

TestTLSConfig was failing due to assert.Empty being used to check if the TLS config was empty but it was not as an empty X509 pool was being passed in, and the TLS Skip Verify value.

TestGoCBv2SecurityConfig was failing due to a CA Cert being provided which meant the returned value would have the CA cert provided and not be empty. The test was asserting this would be empty.

### DIFF
--- a/base/bucket_test.go
+++ b/base/bucket_test.go
@@ -14,7 +14,6 @@ import (
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rand"
-	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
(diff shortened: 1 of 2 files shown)


JRascagneres:
Fix looks mostly fine but as static analysis says assert.Equal ends up copying a mutex that is included in tls.Config which go vet doesn't like (and you shouldn't do). Would recommend asserting on the two items inside the config.

JRascagneres (dismissed):

JRascagneres (approved):

PR #5125, opened 2021-07-27
CBG-1564: Apply query limit to N1QL and allow configuration

JRascagneres:
Applies the viewQueryOps limit to N1QL queries and allow this option to be configured by the user.

### DIFF
--- a/base/bucket.go
+++ b/base/bucket.go
@@ -106,19 +106,19 @@ type CouchbaseBucketType int
 
 // Full specification of how to connect to a bucket
 type BucketSpec struct {
-	Server, BucketName, FeedType     string
-	Auth                             AuthHandler
-	CouchbaseDriver                  CouchbaseDriver
-	Certpath, Keypath, CACertPath    string         // X.509 auth parameters
-	TLSSkipVerify                    bool           // Use insecureSkipVerify when secure scheme (couchbases) is used and cacertpath is undefined
-	KvTLSPort                        int            // Port to use for memcached over TLS.  Required for cbdatasource auth when using TLS
-	MaxNumRetries                    int            // max number of retries before giving up
-	InitialRetrySleepTimeMS          int            // the initial time to sleep in between retry attempts (in millisecond), which will double each retry
-	UseXattrs                        bool           // Whether to use xattrs to store _sync metadata.  Used during view initialization
-	ViewQueryTimeoutSecs             *uint32        // the view query timeout in seconds (default: 75 seconds)
-	ViewQueryMaxConcurrentOpsPerNode int            // maximum number of concurrent view / query operations (default: 100)
-	BucketOpTimeout                  *time.Duration // How long bucket ops should block returning "operation timed out". If nil, uses GoCB default.  GoCB buckets only.
-	KvPoolSize                       int            // gocb kv_pool_size - number of pipelines per node. Initialized on GetGoCBConnString
+	Server, BucketName, FeedType  string
+	Auth                          AuthHandler
+	CouchbaseDriver               CouchbaseDriver
+	Certpath, Keypath, CACertPath string         // X.509 auth parameters
+	TLSSkipVerify                 bool           // Use insecureSkipVerify when secure scheme (couchbases) is used and cacertpath is undefined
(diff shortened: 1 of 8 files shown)


base/bucket_gocb.go
```diff
@@ -169,14 +169,14 @@ func GetCouchbaseBucketGoCBFromAuthenticatedCluster(cluster *gocb.Cluster, spec
 	// to avoid gocb queue overflow issues
 	singleOpsQueue := make(chan struct{}, MaxConcurrentSingleOps*nodeCount*numPools)
 	bucketOpsQueue := make(chan struct{}, MaxConcurrentBulkOps*nodeCount*numPools)
-	viewOpsQueue := make(chan struct{}, MaxConcurrentViewOps*nodeCount)
+	viewOpsQueue := make(chan struct{}, spec.ViewQueryMaxConcurrentOpsPerNode*nodeCount)
```
adamcfraser: I don't think we should multiple by node count here, as the limit we're trying to avoid is the number of sockets being used on this node, not the CBS node.  (this differs from kv ops, where we're limiting the number of concurrent ops in the per-node pipeline established to CBS nodes).

rest/config.go
```diff
@@ -59,15 +59,16 @@ const (
...
+	Username                                string  `json:"username,omitempty"`                                      // Username for authenticating to server
+	Password                                string  `json:"password,omitempty"`                                      // Password for authenticating to server
+	CertPath                                string  `json:"certpath,omitempty"`                                      // Cert path (public key) for X.509 bucket auth
+	KeyPath                                 string  `json:"keypath,omitempty"`                                       // Key path (private key) for X.509 bucket auth
+	CACertPath                              string  `json:"cacertpath,omitempty"`                                    // Root CA cert path for X.509 bucket auth
+	KvTLSPort                               int     `json:"kv_tls_port,omitempty"`                                   // Memcached TLS port, if not default (11207)
+	MaxConcurrentViewQueryOperationsPerNode *int    `json:"max_concurrent_view_query_operations_per_node,omitempty"` // Max concurrent view / query ops
```
adamcfraser: Should be renamed to max_concurrent_query_ops, since it's not limited to views, and it's not per node.

rest/config.go
```diff
@@ -88,14 +89,20 @@ func (dc *DbConfig) MakeBucketSpec() base.BucketSpec {
 		tlsPort = bc.KvTLSPort
 	}
 
+	maxConcurrentViewOpsPerNode := base.MaxConcurrentViewOps
```
adamcfraser: Can rename constant to DefaultMaxConcurrentQueryOps.

base/bucket.go
```diff
@@ -106,18 +106,19 @@ type CouchbaseBucketType int
...
+	Auth                             AuthHandler
+	CouchbaseDriver                  CouchbaseDriver
+	Certpath, Keypath, CACertPath    string         // X.509 auth parameters
+	TLSSkipVerify                    bool           // Use insecureSkipVerify when secure scheme (couchbases) is used and cacertpath is undefined
+	KvTLSPort                        int            // Port to use for memcached over TLS.  Required for cbdatasource auth when using TLS
+	MaxNumRetries                    int            // max number of retries before giving up
+	InitialRetrySleepTimeMS          int            // the initial time to sleep in between retry attempts (in millisecond), which will double each retry
+	UseXattrs                        bool           // Whether to use xattrs to store _sync metadata.  Used during view initialization
+	ViewQueryTimeoutSecs             *uint32        // the view query timeout in seconds (default: 75 seconds)
+	ViewQueryMaxConcurrentOpsPerNode int            // maximum number of concurrent view / query operations (default: 100)
```
adamcfraser: I think `default: DefaultMaxConcurrentQueryOps` is probably better.  On a related note, I think 100 is probably too low of a default and should be updated from the previous default.  Given that we default MaxFileDescriptors to 5000, I think it's probably reasonable to set this default to 1000.

adamcfraser (changes requested):
Generally looks good, a few comments on the specifics.

adamcfraser (approved):

PR #5126, opened 2021-07-28
CBG-1518 - Added ServerTLSSkipVerify to legacy config

IsaacLambat:


rest/config_legacy.go
```diff
@@ -289,7 +291,13 @@ func (config *LegacyServerConfig) validate() (errorMessages error) {
...
 		}
 	}
-
+	if config.ServerTLSSkipVerify != nil && *config.ServerTLSSkipVerify {
+		for name, database := range config.Databases {
+			if database.CACertPath != "" {
+				errorMessages = multierror.Append(errorMessages, fmt.Errorf("cannot skip server TLS validation and use CA Cert for database %s", name))
+			}
+		}
+	}
```
bbrks: I think this validation is already handled as part of `setupServerContext` -> `config.validate` - we'll just be hitting it a bit later in the startup process.

bbrks (changes requested):

rest/config_legacy.go
IsaacLambat: All done. Manually tested it and it does hit the persistent config validate function.

rest/config_legacy_test.go
```diff
@@ -119,3 +119,55 @@ func TestLegacyConfigXattrsDefault(t *testing.T) {
 		})
 	}
 }
+
+func TestCACertSkipVerify(t *testing.T) {
```
bbrks: also needs removing

bbrks (changes requested):

bbrks (approved):

PR #5130, opened 2021-07-29
Test enhancements related to proveAttachment handling

sarathkumarsivan:
Enhance the unit tests to create a revision on the client with history, and immediately sends a changes request for it. This is specifically for testing proveAttachment handling scenarios.

### DIFF
--- a/rest/blip_api_test.go
+++ b/rest/blip_api_test.go
@@ -2489,7 +2489,7 @@ func TestBlipDeltaSyncPush(t *testing.T) {
 	// create doc1 rev 2-abcxyz on client
 	newRev, err := client.PushRev("doc1", "1-0335a345b6ffed05707ccc4cbc1b67f4", []byte(`{"greetings":[{"hello":"world!"},{"hi":"alice"},{"howdy":"bob"}]}`))
 	assert.NoError(t, err)
-	assert.Equal(t, "2-abcxyz", newRev)
+	assert.Equal(t, "2-abc", newRev)
 
 	// Check EE is delta, and CE is full-body replication
 	msg, ok := client.pushReplication.WaitForMessage(2)
@@ -2522,7 +2522,7 @@ func TestBlipDeltaSyncPush(t *testing.T) {
 	var respBody db.Body
 	assert.NoError(t, base.JSONUnmarshal(resp.Body.Bytes(), &respBody))
 	assert.Equal(t, "doc1", respBody[db.BodyId])
-	assert.Equal(t, "2-abcxyz", respBody[db.BodyRev])
+	assert.Equal(t, "2-abc", respBody[db.BodyRev])
 	greetings := respBody["greetings"].([]interface{})
 	assert.Len(t, greetings, 3)
 	assert.Equal(t, map[string]interface{}{"hello": "world!"}, greetings[0])
@@ -2533,7 +2533,7 @@ func TestBlipDeltaSyncPush(t *testing.T) {
 	resp = rt.SendAdminRequest(http.MethodDelete, "/db/doc1?rev="+newRev, "")
 	assert.Equal(t, http.StatusOK, resp.Code)
 
-	data, ok = client.WaitForRev("doc1", "3-f3be6c85e0362153005dae6f08fc68bb")
+	data, ok = client.WaitForRev("doc1", "3-fcc2db8cdbf1831799b7a39bb57edd71")
 	assert.True(t, ok)
 	assert.Equal(t, `{}`, string(data))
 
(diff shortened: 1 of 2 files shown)


rest/blip_client_test.go
```diff
@@ -715,6 +716,151 @@ func (btc *BlipTesterClient) PushRev(docID, parentRev string, body []byte) (revI
 	return newRevID, nil
 }
 
+// PushRevWithHistory creates a revision on the client with history, and immediately sends a changes request for it.
+func (btc *BlipTesterClient) PushRevWithHistory(docID, parentRev string, body []byte, revCount, revpos int) (revID string, err error) {
```
adamcfraser: I expected that PushRev would now just turn around and call PushRevWithHistory, with revCount=1.  Is there a reason that can't be done, to avoid the code duplication below?

rest/blip_api_test.go
```diff
@@ -3286,3 +3286,117 @@ func TestRemovedMessageWithAlternateAccess(t *testing.T) {
...
+
+	attachments, ok := respBody[db.BodyAttachments].(map[string]interface{})
+	require.True(t, ok)
+	assert.Len(t, attachments, 1)
+	hello, ok := attachments["hello.txt"].(map[string]interface{})
+	require.True(t, ok)
+	assert.Equal(t, "sha1-Kq5sNclPz7QV2+lfQIuc6R7oRu0=", hello["digest"])
+	assert.Equal(t, float64(11), hello["length"])
+	assert.Equal(t, float64(2), hello["revpos"])
+	assert.True(t, hello["stub"].(bool))
```
adamcfraser, line 3503: The motivation for these tests is to check whether SG is issuing a proveAttachment or getAttachment call in each of the scenarios.  If there aren't stats in the SG-side code, we should add stats to the test client to track the number of getAttachment / proveAttachment requests recieved.  That will let us validate that the planned future enhancement to revpos handling avoids unnecessary invocation of getAttachment/proveAttachment.

adamcfraser (changes requested):
The tests should be asserting the issue we're trying to add coverage for here - the number of GetAttachment/proveAttachment calls being made in these scenarios.

rest/blip_client_test.go
sarathkumarsivan: PushRev was originally accepting three parameters (docID, parent revision and body bytes), and when we introduced the new parameter revCount as part of this enhancement, two more additional parameters were required to specify the revpos, and skipping document not found error. The reason is that we need to have the ability to set the revpos in the new unit tests that track the calls to getAttachment/proveAttachment requests. Moreover, we don’t need to return an error when the docID doesn’t exist in the client's in-memory store while pushing the revisions with history. In addition, introducing more parameters to the PushRevWithHistory looked a little ugly and I thought of having a separate function for pushing revisions with history. However, when considering the suggestion here, it makes sense to refactor both PushRev and PushRevWithHistory and reduce the duplicate lines of code.

rest/blip_api_test.go
sarathkumarsivan: The stats for capturing both GetAttachment and ProveAttachment are encapsulated in BlipSyncStats and it seems that grabbing those stats from BlipTesterClient is not straight forward. Both AttachmentPushBytes and AttachmentPushCount were used to validate CBLReplicationPush instead. But, let's review this once again.

rest/blip_client_test.go
```diff
@@ -573,11 +582,25 @@ func (btr *BlipTesterReplicator) sendMsg(msg *blip.Message) (err error) {
...
 // PushRev creates a revision on the client, and immediately sends a changes request for it.
-// The rev ID is always: "N-abcxyz", where N is rev generation for predictability.
+// The rev ID is always: "N-abc", where N is rev generation for predictability.
 func (btc *BlipTesterClient) PushRev(docID, parentRev string, body []byte) (revID string, err error) {
+	return btc.PushRevWithHistory(docID, parentRev, body, 1, 0, false)
+}
 
-	// generate fake rev for gen+1
+// PushRevWithHistory creates a revision on the client with history, and immediately sends a changes request for it.
+func (btc *BlipTesterClient) PushRevWithHistory(docID, parentRev string, body []byte, revCount, revpos int, skipDocNotFoundErr bool) (revID string, err error) {
```
adamcfraser: As discussed, change skipDocNotFoundErr to 'prunedRevisionCount', and have the test pass parentRev="" and prunedRevisionCount = 1 for the scenario where we want to send a pruned history.  Line 593 would calculate revGen as parentRevGen+revCount+prunedRevisionCount.

adamcfraser (changes requested):

rest/blip_client_test.go
sarathkumarsivan: Sure, the PushRevWithHistory has been refactored and included an additional method to store the revision only on the client-side.

adamcfraser (approved):

PR #5132, opened 2021-07-30
CBG-1568 - Added Windows testing to CI

IsaacLambat:


IsaacLambat:
TODO: Make fail with windows test.
Make fail with unix test and see if it stops win test.
Remove echo "GOPATH ...." from CI categories.

.github/workflows/ci.yml
```diff
@@ -39,35 +53,48 @@ jobs:
         run: go vet github.com/couchbase/sync_gateway/...
 
   test:
+    needs: run-bootstrap
     env:
       GO111MODULE: off
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
```
JRascagneres, line 62: I think we want to disable fail-fast.

Means that if one item in the matrix fails we still run the other. ie. if one OS fails we still want to test the other. When I say 'fail' this will only occur if the whole thing errors, ie. a panic as opposed to a test failure but still might be nice

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

.github/workflows/ci.yml
```diff
@@ -39,35 +53,48 @@ jobs:
...
+        include:
+          - os: ubuntu-latest
+            set-gopath: echo "GOPATH=$GITHUB_WORKSPACE/godeps" >> $GITHUB_ENV
+
+          - os: windows-latest
+            set-gopath: echo "GOPATH=$Env:GITHUB_WORKSPACE\godeps" >> $Env:GITHUB_ENV
```
JRascagneres, line 69: I know its not really in the scope of this PR but maybe just for completeness we could add macOS? Fairly sure it'd be the same steps as ubuntu but as we already have the matrix setup it should be easy and the tests wouldn't take any longer as they're in parallel.

Just means we'd be covering all the supported github actions OSs for completeness.

Should just be `macos-latest`. Maybe just stick it in, if its not the same as ubuntu don't spend time on it.

JRascagneres (changes requested):
Thrown a couple minor comments on. All looks pretty good though. Good job :)

JRascagneres (approved):
LGTM will request final review from @bbrks

bbrks (approved):
LGTM!

PR #5135, opened 2021-08-02
CBG-1399: Pre-3.0 config automatic upgrade

JRascagneres:
PR handles automatic migration of pre-3.0 configs to 3.0 persistent configs:
1. Converts config to startup config
2. Validates / sanitizes db config - Verifies servers are all the same and removes unused info like server, users & roles
3. Connects to CBS and uploads the new db configs
4. Saves current config file as a backup 
5. Overwrites old config file with new startup file
6. 
7. Copied out the work done in `setupServerContext` when connecting to the CBS cluster as this is required in `automaticConfigUpgrade`.

Added a basic implementation of PutConfig which works for the required use-case in the PR. Can be updated later down the road by @bbrks for his later requirements.

### DIFF
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -2,6 +2,7 @@ package base
 
 import (
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/couchbase/gocb"
@@ -124,17 +125,17 @@ func (cc *CouchbaseCluster) PutConfig(location, groupID string, cas *uint64, val
 	}
 	docID := PersistentConfigPrefix + groupID
 	collection := cc.c.Bucket(location).DefaultCollection()
-	// TODO: CBG-1452 - Optimistic concurrency control (CAS or Opaque field in config rather than in URL?)
-	// if cas != nil {
-	// 	res, err = collection.Replace(docID, value, &gocb.ReplaceOptions{Cas: gocb.Cas(*cas)})
-	res, err := collection.Upsert(docID, value, nil)
-	if err != nil {
-		switch err {
-		case gocb.ErrCasMismatch:
+
+	if cas != nil && *cas == 0 {
+		res, err := collection.Insert(docID, value, nil)
+		if err != nil {
+			return 0, err
 		}
-		return 0, err
+
+		return uint64(res.Cas()), nil
 	}
-	return uint64(res.Cas()), nil
+
+	return 0, fmt.Errorf("Not implemented")
 }
 
 func (cc *CouchbaseCluster) Close() error {
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -118,9 +118,23 @@ func (cc *CouchbaseCluster) GetConfig(location, groupID string, valuePtr interfa
 	return uint64(res.Cas()), nil
 }
 
-func (cc *CouchbaseCluster) PutConfig(bucket, groupID string, cas *uint64, value interface{}) (newCAS uint64, err error) {
-	// TODO: Implement
-	return 0, nil
+func (cc *CouchbaseCluster) PutConfig(location, groupID string, cas *uint64, value interface{}) (newCAS uint64, err error) {
+	if cc == nil {
+		return 0, errors.New("nil CouchbaseCluster")
+	}
+	docID := PersistentConfigPrefix + groupID
+	collection := cc.c.Bucket(location).DefaultCollection()
+	// TODO: CBG-1452 - Optimistic concurrency control (CAS or Opaque field in config rather than in URL?)
+	// if cas != nil {
+	// 	res, err = collection.Replace(docID, value, &gocb.ReplaceOptions{Cas: gocb.Cas(*cas)})
+	res, err := collection.Upsert(docID, value, nil)
+	if err != nil {
+		switch err {
+		case gocb.ErrCasMismatch:
+		}
+		return 0, err
+	}
+	return uint64(res.Cas()), nil
 }
 
 func (cc *CouchbaseCluster) Close() error {
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/couchbase/gocb"
+	"github.com/couchbase/gocbcore/memd"
 )
 
 // BootstrapConnection is the interface that can be used to bootstrap Sync Gateway against a Couchbase Server cluster.
@@ -129,6 +130,11 @@ func (cc *CouchbaseCluster) PutConfig(location, groupID string, cas *uint64, val
 	if cas != nil && *cas == 0 {
 		res, err := collection.Insert(docID, value, nil)
 		if err != nil {
+
+			if isKVError(err, memd.StatusKeyExists) {
+				return 0, ErrAlreadyExists
+			}
+
 			return 0, err
 		}
 
--- a/base/collection_xattr.go
+++ b/base/collection_xattr.go
@@ -402,6 +402,10 @@ func isKVError(err error, code memd.StatusCode) bool {
 	return false
 }
 
+func IsAlreadyExistsError(err error) bool {
+	return isKVError(err, memd.StatusKeyExists)
+}
+
 // If v is []byte or *[]byte, converts to json.RawMessage to avoid duplicate marshalling by gocb.
 func bytesToRawMessage(v interface{}) interface{} {
 	switch val := v.(type) {
--- a/base/collection_xattr.go
+++ b/base/collection_xattr.go
@@ -402,10 +402,6 @@ func isKVError(err error, code memd.StatusCode) bool {
 	return false
 }
 
-func IsAlreadyExistsError(err error) bool {
-	return isKVError(err, memd.StatusKeyExists)
-}
-
 // If v is []byte or *[]byte, converts to json.RawMessage to avoid duplicate marshalling by gocb.
 func bytesToRawMessage(v interface{}) interface{} {
 	switch val := v.(type) {
--- a/rest/main.go
+++ b/rest/main.go
@@ -109,9 +109,13 @@ func serverMainPersistentConfig(fs *flag.FlagSet, flagStartupConfig *StartupConf
 		if pkgerrors.Cause(err) == base.ErrUnknownField {
 
 			// Attempt to perform automatic config upgrade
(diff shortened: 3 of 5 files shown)


rest/persistent_config_test.go
```diff
@@ -0,0 +1,118 @@
...
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("CBS required")
+	}
+
+	tb := base.GetTestBucket(t)
+	defer tb.Close()
+
+	rawConfig := `
+	{
+		"databases": {
```
bbrks: Can you add some node-level config to make sure they're migrated properly?

rest/main.go
```diff
@@ -146,3 +158,144 @@ func serverMainPersistentConfig(fs *flag.FlagSet, flagStartupConfig *StartupConf
...
+	}
+
+	// Attempt to backup current config
+	err = backupCurrentConfigFile(configPath)
+	if err != nil {
+		return nil, false, err
+	}
+
+	// Overwrite old config with new migrated startup config
+	jsonStartupConfig, err := json.Marshal(startupConfig)
```
bbrks: unhandled error

rest/main.go
```diff
@@ -146,3 +158,144 @@ func serverMainPersistentConfig(fs *flag.FlagSet, flagStartupConfig *StartupConf
...
+	var databaseServerAddress string
+
+	for dbName, dbConfig := range configMap {
+		if databaseServerAddress == "" {
+			databaseServerAddress = *dbConfig.Server
+		}
+
+		if *dbConfig.Server != databaseServerAddress {
+			return nil, fmt.Errorf("server addresses specified in dbConfig do not match. This is required for " +
+				"persistent config")
```
bbrks: Rephrase to be related to auto-upgrade, not the persistent config feature.

```suggestion
			return nil, fmt.Errorf("Automatic upgrade to persistent config requires matching server addresses in 2.x config")
```

bbrks (changes requested):

rest/main.go
```diff
@@ -146,3 +158,144 @@ func serverMainPersistentConfig(fs *flag.FlagSet, flagStartupConfig *StartupConf
...
+		}
+
+		if dbConfig.Bucket == nil || *dbConfig.Bucket == "" {
+			*dbConfig.Bucket = dbName
```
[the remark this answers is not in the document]
JRascagneres: Yikes... yeah.

rest/main.go
JRascagneres: I'll switch but technically its true that even if we're running legacy we only support one server now right?

rest/main.go
```diff
@@ -102,10 +107,17 @@ func serverMainPersistentConfig(fs *flag.FlagSet, flagStartupConfig *StartupConf
...
 		if pkgerrors.Cause(err) == base.ErrUnknownField {
-			// TODO: CBG-1399 Do automatic legacy config upgrade here
-			base.Warnf("Couldn't parse bootstrap config and legacy config migration not yet implemented: %v", err)
-			// When automatic legacy config upgrade is done return when disable_persistent_config=true
-			return true, nil
+
+			// Attempt to perform automatic config upgrade
+			fileStartupConfig, disablePersistentConfigFallback, err = automaticConfigUpgrade(configPath[0])
+			if err != nil {
+				return false, err
```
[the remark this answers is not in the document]
JRascagneres: Understood... See if you think my solution covers it

base/bootstrap.go
```diff
@@ -118,9 +118,23 @@ func (cc *CouchbaseCluster) GetConfig(location, groupID string, valuePtr interfa
...
+	// if cas != nil {
+	// 	res, err = collection.Replace(docID, value, &gocb.ReplaceOptions{Cas: gocb.Cas(*cas)})
+	res, err := collection.Upsert(docID, value, nil)
+	if err != nil {
+		switch err {
+		case gocb.ErrCasMismatch:
+		}
+		return 0, err
+	}
+	return uint64(res.Cas()), nil
```
bbrks: I guess for this PR, just make `PutConfig` to do an Insert op when cas==0, and otherwise error with "not implemented". I can just extend this function when I need to perform Replace or Update ops.

At least then you can test your cas-safe insert and have the upgrade do what it's supposed to do.

rest/main.go
```diff
@@ -102,10 +108,31 @@ func serverMainPersistentConfig(fs *flag.FlagSet, flagStartupConfig *StartupConf
...
+			if disablePersistentConfigFallback {
+				return true, nil
+			}
+
+			// If we have got this far the automatic config upgrade was successful so we should nil out the error so we
+			// can continue
+			err = nil
+
 		}
 		if err != nil {
```
bbrks: This is effectively the same thing now we're not always returning, feels cleaner and more obvious

```suggestion
		} else if err != nil {
```

rest/main.go
```diff
@@ -146,3 +173,158 @@ func serverMainPersistentConfig(fs *flag.FlagSet, flagStartupConfig *StartupConf
...
+	}
+	defer func() {
+		_ = source.Close()
+	}()
+
+	// Grab file extension
+	fileExtension := filepath.Ext(sourcePath)
+
+	// Remove file extension so we can easily modify filename
+	fileNameWithoutExtension := strings.ReplaceAll(sourcePath, fileExtension, "")
```
bbrks: This seems safer...

```suggestion
	// Remove file extension so we can easily modify filename
	fileNameWithoutExtension := strings.TrimSuffix(sourcePath, fileExtension)
```

rest/persistent_config_test.go
```diff
@@ -0,0 +1,122 @@
...
+		t.Skip("CBS required")
+	}
+
+	tb := base.GetTestBucket(t)
+	defer tb.Close()
+
+	rawConfig := `
+	{
+		"interface": ":4444",
+  		"adminInterface": ":4445",
```
bbrks:
```suggestion
		"adminInterface": ":4445",
```

rest/persistent_config_test.go
```diff
@@ -0,0 +1,122 @@
...
+		"databases": {
+			"db": {
+			  "server": "%s",
+			  "username": "%s",
+			  "password": "%s",
+			  "bucket": "%s"
```
bbrks:
```suggestion
				"server": "%s",
				"username": "%s",
				"password": "%s",
				"bucket": "%s"
```

rest/persistent_config_test.go
```diff
@@ -0,0 +1,122 @@
...
+						  "server": "%s",
+						  "username": "%s",
+						  "password": "%s",
+						  "bucket": "%s"
+						},
+						"db2": {
+						  "server": "rand",
+						  "username": "",
+						  "password": "",
+						  "bucket": ""
```
bbrks:
```suggestion
							"server": "%s",
							"username": "%s",
							"password": "%s",
							"bucket": "%s"
						},
						"db2": {
							"server": "rand",
							"username": "",
							"password": "",
							"bucket": ""
```

rest/persistent_config_test.go
```diff
@@ -0,0 +1,122 @@
...
+	require.NoError(t, err)
+
+	startupConfig, _, err := automaticConfigUpgrade(configPath)
+	assert.NoError(t, err)
+
+	assert.Equal(t, base.UnitTestUrl(), startupConfig.Bootstrap.Server)
+	assert.Equal(t, base.TestClusterUsername(), startupConfig.Bootstrap.Username)
+	assert.Equal(t, base.TestClusterPassword(), startupConfig.Bootstrap.Password)
+	assert.Equal(t, ":4444", startupConfig.API.PublicInterface)
+	assert.Equal(t, ":4445", startupConfig.API.AdminInterface)
```
bbrks, line 56: Are we able to include assertions for the backup file and newly written config file too? Not just the returned in-memory versions.

bbrks (changes requested):

rest/main.go
JRascagneres: I was looking for a "ReplaceLast" or something....

rest/main.go
```diff
@@ -146,3 +168,161 @@ func serverMainPersistentConfig(fs *flag.FlagSet, flagStartupConfig *StartupConf
...
+	}
+
+	// Attempt to backup current config
+	err = backupCurrentConfigFile(configPath)
+	if err != nil {
+		return nil, false, err
+	}
+
+	// Overwrite old config with new migrated startup config
+	jsonStartupConfig, err := json.Marshal(startupConfig)
```
bbrks: pretty print JSON for new config file

```suggestion
	jsonStartupConfig, err := json.MarshalIndent(startupConfig, "", "  ")
```

rest/main.go
```diff
@@ -102,12 +108,28 @@ func serverMainPersistentConfig(fs *flag.FlagSet, flagStartupConfig *StartupConf
...
+					return false, err
+				}
+
+				return false, upgradeError
+			}
+
+			if disablePersistentConfigFallback {
+				return true, nil
+			}
+
```
bbrks: There is no logging around any of this. It makes it really difficult to tell what's happening on startup/migration/upgrade. Given it's a rather disruptive change, it would make sense to be pretty verbose about what we're doing (e.g. "backing up 2.x config to '...'")

```
> go build && ./sync_gateway cnf_dbg.json
2021-08-03T20:01:32.031+01:00 ==== Couchbase Sync Gateway/() CE ====
2021-08-03T20:01:32.031+01:00 [INF] Loading content from [cnf_dbg.json] ...
2021-08-03T20:01:32.031+01:00 [INF] Loading content from [cnf_dbg.json] ...
2021-08-03T20:01:33.267+01:00 [INF] Config: Starting in persistent mode using config group "default"
2021-08-03T20:01:33.267+01:00 [INF] Logging: Console to stderr
...
```

base/collection_xattr.go
```diff
@@ -402,6 +402,10 @@ func isKVError(err error, code memd.StatusCode) bool {
 	return false
 }
 
+func IsAlreadyExistsError(err error) bool {
+	return isKVError(err, memd.StatusKeyExists)
+}
```
bbrks: I think we've generally been removing these single-error functions haven't we? Or at least making them support all SDKs/buckets like `IsCasMismatch`

bbrks (changes requested):
some general feedback after using it and running into small issues

bbrks (approved):
lgtm!

PR #5137, opened 2021-08-03
CBG-1580: Modify Admin Auth Cluster-Level roles

JRascagneres:
Modified cluster-level role access to be:

PUT / POST operations can be performed by:
- Full Admin
- Cluster Admin

All other cluster-level operations can be performed by:
- Full Admin
- Cluster Admin
- Read Only Admin

Bucket level operations have been modified a little (remove read-only admin ):
- Mobile Sync Gateway 
- Bucket Full Access / Application Access
- Bucket Admin
- Full Admin 

Also limited the permissions-based check to only be available to EE.

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -842,6 +842,11 @@ func (sc *StartupConfig) validate() (errorMessages error) {
 			errorMessages = multierror.Append(errorMessages, fmt.Errorf("Cannot use supplied TLS key or cert when api.https.use_tls_client is false"))
 		}
 	}
+
+	if !base.IsEnterpriseEdition() && sc.API.EnableAdminAuthenticationPermissionsCheck != nil && *sc.API.EnableAdminAuthenticationPermissionsCheck {
+		errorMessages = multierror.Append(errorMessages, fmt.Errorf("enable_admin_authentication_permissions_check is only supported in enterprise edition"))
+	}
+
 	return errorMessages
 }
 
--- a/rest/admin_api_auth_test.go
+++ b/rest/admin_api_auth_test.go
@@ -415,7 +415,7 @@ func TestAdminAuth(t *testing.T) {
 				defer DeleteUser(t, managementEndpoints[0], testCase.CreateUser)
 			}
 
-			permResults, statusCode, err := checkAdminAuth(testCase.BucketName, testCase.Username, testCase.Password, httpClient, managementEndpoints, true, testCase.CheckPermissions, testCase.ResponsePermissions)
+			permResults, statusCode, err := checkAdminAuth(testCase.BucketName, testCase.Username, testCase.Password, "", httpClient, managementEndpoints, true, testCase.CheckPermissions, testCase.ResponsePermissions)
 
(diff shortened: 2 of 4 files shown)


rest/config.go
```diff
@@ -842,6 +842,11 @@ func (sc *StartupConfig) validate() (errorMessages error) {
 			errorMessages = multierror.Append(errorMessages, fmt.Errorf("Cannot use supplied TLS key or cert when api.https.use_tls_client is false"))
 		}
 	}
+
+	if !base.IsEnterpriseEdition() && sc.API.EnableAdminAuthenticationPermissionsCheck != nil && *sc.API.EnableAdminAuthenticationPermissionsCheck {
+		errorMessages = multierror.Append(errorMessages, fmt.Errorf("enable_admin_authentication_permissions_check is only supported in enterprise edition"))
```
IsaacLambat, line 847: Would it be better if the section was included where the config value is?
```suggestion
		errorMessages = multierror.Append(errorMessages, fmt.Errorf("api.enable_admin_authentication_permissions_check is only supported in enterprise edition"))
```

rest/config.go
```diff
@@ -842,6 +842,11 @@ func (sc *StartupConfig) validate() (errorMessages error) {
 			errorMessages = multierror.Append(errorMessages, fmt.Errorf("Cannot use supplied TLS key or cert when api.https.use_tls_client is false"))
 		}
 	}
+
+	if !base.IsEnterpriseEdition() && sc.API.EnableAdminAuthenticationPermissionsCheck != nil && *sc.API.EnableAdminAuthenticationPermissionsCheck {
```
IsaacLambat, line 846: Could the check for enterprise edition be in an outer if statement then the config value check be nested so it is easy to add enterprise config value checks in the future without doing a redundant enterprise edition check?
```suggestion
        if !base.IsEnterpriseEdition() {
	        if sc.API.EnableAdminAuthenticationPermissionsCheck != nil && *sc.API.EnableAdminAuthenticationPermissionsCheck {
```

rest/config.go
JRascagneres: Reason I omitted that was because that option is available in both legacy and  but you have a point... I think its probably fine to leave as is as a user would have to have explicitly set it to get this error so they probably would know where it is and wouldn't have to search for it....

You do have an interesting point tho

rest/config.go
JRascagneres: Didn't bother doing this as this is the only check for EE in this path. If someone else wants to use it its fairly easy to split it out.

rajagp:
@JRascagneres  The roles look fine. Also, let's include Application Access (although "deprecated").  It has same treatment as "Bucket admin" - so bucket level operations.

JRascagneres:
> @JRascagneres The roles look fine. Also, let's include Application Access (although "deprecated"). It has same treatment as "Bucket admin" - so bucket level operations.

This is handled - Application Access == Bucket Full Access - Have modified description to make clear

adamcfraser (approved):

PR #5141, opened 2021-08-04
CBG-1557 - Amazon Linux 2 case for running as service

IsaacLambat:
Amazon Linux 2 now uses `systemctl` instead of default `service` to run the service.

Tested sync_gateway_service_install.sh, sync_gateway_service_uninstall.sh, and sync_gateway_service_upgrade.sh on [Amazon Linux 2 LTS 2.0.20210721.2 x86_64](https://cdn.amazonlinux.com/os-images/2.0.20210721.2/virtualbox/) on VirtualBox.

Final toy build tested was [couchbase-sync-gateway-enterprise_3.0.0-203_x86_64.rpm]( http://mobile.jenkins.couchbase.com/job/sgw-toy-build/203/).

The sg_accel_service_*.sh scripts where **not** tested as SG Accel is fully deprecated. I did add the case to these scripts though, which was the same modifications as the normal service scripts (as none of the other cases had any differences).

### DIFF
--- a/manifest/default.xml
+++ b/manifest/default.xml
@@ -32,7 +32,7 @@ licenses/APL2.txt.
 
 
   <!-- Sync Gateway -->
-  <project name="sync_gateway" path="godeps/src/github.com/couchbase/sync_gateway" remote="couchbase" revision="CBG-1557"/>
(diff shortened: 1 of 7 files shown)


bbrks (approved):
LGTM

PR #5143, opened 2021-08-04
CBG-1591: Order expectedSeqs when calculating checkpoint

JRascagneres:
Order expectedSeqs in `_calculateSafeExpectedSeqsIdx` which is used when calculating checkpoints during replication.

Ordering expectedSeqs when used for checkpoint calculation rather than ordering on insert to cover all possible cases where expectedSeqs is mutated. 
Checkpointing only occurs every 10 seconds whereas inserts happen more often, therefore sorting is more likely to be cheaper when done as part of `_calculateSafeExpectedSeqsIdx` calculation.


bbrks (dismissed):
LGTM but would like @adamcfraser 's review

adamcfraser (approved):

PR #5151, opened 2021-08-06
Modified perms to be inline with 7.0.1 and test

JRascagneres:
The perms have been added to the latest test version of 7.0.1, however, not in the expected format so had to make some changes to support them here.
Additionally wrote a test for each endpoint exercising the roles, ensuring that the expected role can access and the other SGW roles cannot.

Has been tested on build 6002 of CBS 7.0.1 and proven to work.

- [ ] Integration test (for altered pre-7.0 tests)
http://uberjenkins.sc.couchbase.com:8080/view/All/job/sync-gateway-integration/979/

### DIFF
--- a/rest/admin_api_auth_routing_permissions.go
+++ b/rest/admin_api_auth_routing_permissions.go
@@ -8,21 +8,13 @@ import (
 // Permission stores the name of a permission along whether it is database scoped. This is used to later obtain a
 // formatted permission string for checking.
 type Permission struct {
-	PermissionName     string
-	DatabaseScoped     bool
-	IsCollectionFormat bool
+	PermissionName string
+	DatabaseScoped bool
 }
 
 func (perm *Permission) FormattedName(bucketName string) string {
-	if perm.IsCollectionFormat {
-		if perm.DatabaseScoped {
-			return fmt.Sprintf("cluster.collection[%s:_default:_default]%s", bucketName, perm.PermissionName)
-		}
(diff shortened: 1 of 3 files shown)


rest/admin_api_auth_routing_permissions.go
```diff
@@ -8,11 +8,19 @@ import (
...
 }
 
 func (perm *Permission) FormattedName(bucketName string) string {
+	if perm.IsCollectionFormat {
+		if perm.DatabaseScoped {
+			return fmt.Sprintf("cluster.collection[%s:_default:_default]%s", bucketName, perm.PermissionName)
+		}
+		return fmt.Sprintf("cluster.collection[*:*:*]%s", perm.PermissionName)
+	}
+
```
adamcfraser: Do we have (or anticipate) any permissions where DatabaseScoped != IsCollectionFormat?  I don't see any today - am curious which permissions would be using the `cluster.bucket[%s]%s` format.

adamcfraser (changes requested):

rest/admin_api_auth_routing_permissions.go
JRascagneres: A few existing permissions on CBS still use that older format but true, there are none that we use that are in that format now so we can probably remove that. 
Similarly the case of `cluster.collection[*:*:*]%s` is now no longer needed as Bryan changed the format of the dev_ops permission in the latest fix. Will remove those things and probably strip down just to database scoped.

adamcfraser (approved):

PR #5154, opened 2021-08-06
Reduce work TestTLSSkipVerifyCombinations is doing

bbrks:
Narrows down testing to only the validate function.
This avoids some of the disruptive parts of the test that reset global logging, because we're no longer setting up a full server context.

### DIFF
--- a/rest/server_context_test.go
+++ b/rest/server_context_test.go
@@ -376,14 +376,8 @@ func TestTLSSkipVerifyCombinations(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			startupConfig := &StartupConfig{
 				Bootstrap: BootstrapConfig{
-					Server:              base.UnitTestUrl(),
 					CACertPath:          test.caCert,
 					ServerTLSSkipVerify: test.serverTLSSkipVerify,
-				},
-				API: APIConfig{
-					HTTPS: HTTPSConfig{
-						UseTLSClient: base.BoolPtr(false),
-					},
 				},
 			}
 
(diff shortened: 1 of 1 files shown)


rest/server_context_test.go
```diff
@@ -378,19 +376,23 @@ func TestTLSSkipVerifyCombinations(t *testing.T) {
...
 
-			sc, err := setupServerContext(startupConfig, false)
+			err := startupConfig.validate()
 			if test.expectError {
 				assert.Error(t, err)
 				assert.Contains(t, err.Error(), errorText)
-				assert.Empty(t, sc)
-			} else if err != nil {
-				// check if unrelated error
-				assert.NotContains(t, err.Error(), errorText)
```
IsaacLambat, line 393: The point of this was to check that the error from the validation was not to do with TLS Skip Verify so then the validate function can have additional checks to the Startup Config in the future without having to change this test and without the test failing.

For example, this meant that the UseTLSClient config option didn't have to be added as it's errors would be ignored. 

assert.NotContains was the best way I could come up with to assert that the server TLS Skip verify error did not appear in the multiple error messages.

I would be interested in finding out if this is not a good practise to follow though.

IsaacLambat (changes requested):

rest/server_context_test.go
bbrks: That's a fair comment. I think I can leave these checks as-is actually and have it all still work just fine.

IsaacLambat (approved):
LGTM

PR #5155, opened 2021-08-09
CBG-1405 Persist database configs created/updated through REST API

bbrks:
- [x] WIP/Draft PR for self-review

## Integration tests
- [ ] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1013/
- [ ] `xattrs=false` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1014/

### DIFF
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -182,6 +182,10 @@ func (cc *CouchbaseCluster) UpdateConfig(location, groupID string, updateCallbac
 
 		replaceRes, err := collection.Replace(docID, newConfig, &gocb.ReplaceOptions{Transcoder: gocb.NewRawJSONTranscoder(), Cas: res.Cas()})
 		if err != nil {
+			if errors.Is(err, gocb.ErrCasMismatch) {
+				// retry on cas failure
+				continue
+			}
 			return 0, err
 		}
 
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -3,10 +3,12 @@ package base
 import (
 	"errors"
 	"fmt"
+	"reflect"
 	"time"
 
 	"github.com/couchbase/gocb/v2"
 	"github.com/couchbase/gocbcore/v10/memd"
+	"github.com/imdario/mergo"
 )
 
 // BootstrapConnection is the interface that can be used to bootstrap Sync Gateway against a Couchbase Server cluster.
@@ -17,6 +19,8 @@ type BootstrapConnection interface {
 	GetConfig(bucket, groupID string, valuePtr interface{}) (cas uint64, err error)
 	// PutConfig saves a database config for a given bucket and config group ID, along with the CAS of the document.
 	PutConfig(bucket, groupID string, cas *uint64, value interface{}) (newCAS uint64, err error)
+	// UpdateConfig updates a database config for a given bucket and config group ID.
+	UpdateConfig(bucket, groupID string, updateCallback func(rawBucketConfig []byte) (updatedConfig []byte, err error)) (newCAS uint64, err error)
 	// Close closes the connection
 	Close() error
 }
@@ -63,6 +67,7 @@ func NewCouchbaseCluster(server, username, password,
 		RetryStrategy: &goCBv2FailFastRetryStrategy{},
 	})
 	if err != nil {
+		_ = cluster.Close(nil)
 		return nil, err
 	}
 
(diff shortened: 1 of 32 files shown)


rest/config.go
```diff
@@ -1000,9 +1003,11 @@ func (sc *ServerContext) fetchAndLoadDatabase(dbName string) (found bool, err er
 		// TODO: This code is mostly copied from fetchConfigs, move into shared function with DbConfig REST API work?
 
 		// inherit properties the bootstrap config
-		cnf.Server = &sc.config.Bootstrap.Server
+		// cnf.Server = &sc.config.Bootstrap.Server
```
adamcfraser: Can remove this line and the comment.

rest/config.go
```diff
@@ -1000,9 +1003,11 @@ func (sc *ServerContext) fetchAndLoadDatabase(dbName string) (found bool, err er
...
 		cnf.CACertPath = sc.config.Bootstrap.CACertPath
-		cnf.Bucket = &bucket
+
+		bucketCopy := bucket
+		cnf.Bucket = &bucketCopy
```
adamcfraser, line 1039: I didn't think that empty string was a valid bucket name - is the pointer handling in cnf.Bucket just to avoid custom handling for mergo?

adamcfraser (changes requested):
General approach looks good - a couple of details questions.

rest/config.go
bbrks: No but most of the `DatabaseConfig` fields are inherited from `DbConfig` due to complexities in ripping out or changing DbConfig due to retaining 2.x handling.

rest/admin_api_auth_test.go
```diff
@@ -478,518 +478,420 @@ func TestAdminAPIAuth(t *testing.T) {
...
 
+	rt := NewRestTester(t, &RestTesterConfig{
+		adminInterfaceAuthentication:   true,
+		metricsInterfaceAuthentication: true,
+	})
+	defer rt.Close()
+
+	// init RT bucket so we can get the config
+	_ = rt.Bucket()
+	dbConfig := rt.DatabaseConfig
```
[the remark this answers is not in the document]
bbrks: This test is making authorized as well as unauthorized requests for most endpoints (determined by whether `SkipSuccessTest` is set)

The test was `PUT`ing blank database config, which effectively moves the database's bucket from `sg_test_...` to a `db` bucket (bucket name is inherited if unspecified), which ultimately fails and returns a 502 to pass the `!= 401/403` assertion but then results in a missing database.

This then causes 404s on all other sub-tests that come after, rather than the expected 401s.
The change here is to basically PUT the existing database config (including the correct bucket name).

rest/admin_api.go
```diff
@@ -30,17 +30,33 @@ const kDefaultDBOnlineDelay = 0
...
+		// Attempt to apply configuration to node first, to ensure successful before persisting config for other nodes.
+		_, err := h.server.applyConfig(*config)
+		if err != nil {
+			return base.HTTPErrorf(http.StatusInternalServerError, "couldn't load database: %v", err)
+		}
+		// Create DB only, don't update or overwrite
+		cas := base.Uint64Ptr(0)
+		config.cas, err = h.server.bootstrapContext.connection.PutConfig(*config.Bucket, h.server.config.Bootstrap.ConfigGroupID, cas, config)
+		if err != nil {
+			return base.HTTPErrorf(http.StatusInternalServerError, "couldn't save database config: %v", err)
```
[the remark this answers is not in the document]
bbrks: Agreed, although roll back in this case would be to remove the database from the node (it's on a create DB handler, not update DB).

base/bootstrap.go
```diff
@@ -127,21 +132,66 @@ func (cc *CouchbaseCluster) PutConfig(location, groupID string, cas *uint64, val
...
+				}
 
-			if isKVError(err, memd.StatusKeyExists) {
-				return 0, ErrAlreadyExists
+				return 0, err
 			}
 
+			return uint64(res.Cas()), nil
+		} else {
+			// TODO: CAS safe update?
```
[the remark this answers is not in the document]
bbrks: Yep, makes sense. I've also renamed to "InsertConfig" for clarity.

rest/admin_api.go
```diff
@@ -219,24 +232,104 @@ func (h *handler) handlePutConfig() error {
...
+				if err := base.JSONUnmarshal(rawBucketConfig, &bucketDbConfig); err != nil {
+					return nil, err
+				}
+
+				// TODO: CBG-1452 Optimistic Concurrency Control/RevID check
+				if err := base.ConfigMerge(&bucketDbConfig, dbConfig); err != nil {
+					return nil, err
+				}
+
+				// TODO: We're validating but we're not actually starting up the database before we persist the update!
```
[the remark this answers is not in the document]
bbrks: as discussed, going to drop extensive validation/starting db before persisting for the beta - will revisit in CBG-1619

adamcfraser (approved):

adamcfraser:
@bbrks Approved, will let you merge in case the integration test failures need a look.

rest/config_database.go
```diff
@@ -0,0 +1,16 @@
...
+	"github.com/couchbase/sync_gateway/db"
+)
+
+// DatabaseConfig is a 3.x/persisted database config.
+// TODO: Review whether DatabaseConfig should maintain its own list of valid config options, or should just continue inheriting them from DbConfig
+type DatabaseConfig struct {
+	// cas is the Couchbase Server CAS of the database config in the bucket
+	cas uint64
+
+	Guest *db.PrincipalConfig `json:"guest,omitempty"`
```
adamcfraser, line 13: Why is Guest here, as opposed to in DbConfig?  I overlooked earlier that we're using DatabaseConfig (and not DbConfig) for the REST payload.  Is there an intentional distinction?

adamcfraser (commented):
Sorry - just noticed one additional point for clarification while looking at the optimistic concurrency PR.

rest/config_database.go
bbrks: The `DatabaseConfig` struct is currently a bit muddled - in that it stores both 3.x only config options, and fields persisted to the bucket but not returned by REST APIs.

I think we will need to revisit this anyway to strip out now invalid 2.x config options inside DbConfig (by removing the embed completely)

PR #5164, opened 2021-08-11
TestAdminAPIAuth shorten profile times to speed up test

bbrks:
Shortens test by ~240 seconds total


bbrks:
Was already done in another now merged PR

PR #5167, opened 2021-08-11
CBG-1617 - use_tls_server validation done on StartupConfig validation

IsaacLambat:
- [x] http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/982

- Added ServerIsTLS to base where ServerIsWalrus is
- Validation now done in StartupConfig.Validation()
- Tests changed to test StartupConfig validation function
- Added SystemData redaction around server URL (let me know if this is unnecessary)
- HTTPS is not considered as a TLS server protocol any more as it is unsupported
- Fixed broken integration tests

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -56,6 +56,8 @@ const (
 
 	// Default number of index replicas
 	DefaultNumIndexReplicas = uint(1)
+
+	DefaultUseTLSServer = true
 )
 
 // Bucket configuration elements - used by db, index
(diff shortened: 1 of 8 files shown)


rest/config.go
```diff
@@ -830,6 +830,18 @@ func (sc *StartupConfig) validate() (errorMessages error) {
 	if sc.Bootstrap.Server == "" {
 		errorMessages = multierror.Append(errorMessages, fmt.Errorf("a server must be provided in the Bootstrap configuration"))
 	}
+
+	secureServer := base.ServerIsTLS(sc.Bootstrap.Server)
+	if base.BoolDefault(sc.Bootstrap.UseTLSServer, true) {
```
bbrks: I guess we should have a `DefaultUseTLSServer` constant for this value if we need to know it in more than just `DefaultStartupConfig()`

bbrks (changes requested):
Code seems OK (aside from the failing tests (integration have a few failures so make sure they get run and resolved)) but suggest we add a constant for the default value to make it more obvious.

bbrks (approved):
LGTM

PR #5168, opened 2021-08-11
CBG-1608 - Convert http:// server addresses to couchbase:// when upgrading legacy config

IsaacLambat:
- [x] http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/978/
- [x] http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/993/ (flaky test fail)
- [x] http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/9​97

- Added test cases to test behaviour and test cases to new function
- Added code to convert prefix "http://" to "couchbase://"
- Username and password is handled if in server URL (though username and password fields in db config are prioritised) 
- If port is not 8091 or blank for http then error: `automatic migration of connection string from http:// to couchbase:// scheme doesn't support non-default ports.  Please change the server field to use the couchbase(s):// scheme` is returned. For couchbase:// the port is used.
- Get parameters are preserved (and alphabeticalized by url.Encode() )

### DIFF
--- a/rest/config_legacy.go
+++ b/rest/config_legacy.go
@@ -110,7 +110,7 @@ func (lc *LegacyServerConfig) ToStartupConfig() (*StartupConfig, DbConfigMap, er
 		}
 
 		// Prioritise config fields over credentials in host
-		if dbConfig.Username != "" && dbConfig.Password != "" {
+		if dbConfig.Username != "" || dbConfig.Password != "" {
 			username = dbConfig.Username
 			password = dbConfig.Password
 		}
@@ -284,14 +284,14 @@ func gocbv1tov2AddressUpgrade(server string) (newServer, username, password stri
 
 	var hosts string
 	for _, addr := range connSpec.Addresses {
+		if addr.Port != -1 && addr.Port != 8091 && connSpec.Scheme == "http" {
+			return "", "", "", fmt.Errorf("automatic connection string conversion does not support non-default host ports. " +
+				"Please change the server field to use the couchbase(s):// scheme")
+		}
+
 		var port string
-		if addr.Port != 8091 && addr.Port != -1 {
-			if connSpec.Scheme != "http" {
-				port = ":" + strconv.Itoa(addr.Port)
-			} else {
-				return "", "", "", fmt.Errorf("automatic connection string conversion does not support non-default host ports. " +
-					"Please change the server field to use the couchbase(s):// scheme")
-			}
+		if addr.Port != -1 && connSpec.Scheme != "http" {
+			port = ":" + strconv.Itoa(addr.Port)
 		}
 
 		hosts = hosts + addr.Host + port + ","
--- a/rest/config_legacy.go
+++ b/rest/config_legacy.go
@@ -97,11 +97,16 @@ func (lc *LegacyServerConfig) ToStartupConfig() (*StartupConfig, DbConfigMap, er
 	// find a database's credentials for bootstrap (this isn't the first database config entry due to map iteration)
 	bsc := &BootstrapConfig{}
 	for _, dbConfig := range lc.Databases {
-		if dbConfig.Server == nil || *dbConfig.Server == "" {
+		server := dbConfig.Server
+		if server == nil || *server == "" {
 			continue
 		}
+		if strings.HasPrefix(*server, "http://") {
+			*server = strings.Replace(*server, "http", "couchbase", 1)
+		}
+
 		bsc = &BootstrapConfig{
-			Server:              *dbConfig.Server,
+			Server:              *server,
 			Username:            dbConfig.Username,
 			Password:            dbConfig.Password,
 			CACertPath:          dbConfig.CACertPath,
--- a/rest/config_legacy.go
+++ b/rest/config_legacy.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/couchbase/gocbcore/v10/connstr"
-
 	"github.com/couchbase/sync_gateway/base"
 	"github.com/couchbase/sync_gateway/db"
 	"github.com/hashicorp/go-multierror"
@@ -105,7 +104,8 @@ func (lc *LegacyServerConfig) ToStartupConfig() (*StartupConfig, DbConfigMap, er
 
 		server, username, password, err := legacyServerAddressUpgrade(*dbConfig.Server)
 		if err != nil {
-			return nil, nil, err
+			server = *dbConfig.Server
+			base.Errorf("Error upgrading server address: %v", err)
 		}
 
 		// Prioritise config fields over credentials in host
@@ -284,7 +284,7 @@ func legacyServerAddressUpgrade(server string) (newServer, username, password st
 	if connSpec.Scheme == "http" {
 		connSpec.Scheme = "couchbase"
 		for i, addr := range connSpec.Addresses {
-			if addr.Port != 8091 {
+			if addr.Port != 8091 && addr.Port > 0 {
 				return "", "", "", fmt.Errorf("automatic migration of connection string from http:// to couchbase:// scheme doesn't support non-default ports. " +
 					"Please change the server field to use the couchbase(s):// scheme")
 			}
--- a/rest/config_legacy.go
+++ b/rest/config_legacy.go
@@ -5,10 +5,11 @@ import (
 	"fmt"
 	"io"
 	"net/url"
+	"strconv"
 	"strings"
 	"time"
 
-	"gopkg.in/couchbaselabs/gocbconnstr.v1"
+	"github.com/couchbase/gocbcore/v10/connstr"
 
 	"github.com/couchbase/sync_gateway/base"
 	"github.com/couchbase/sync_gateway/db"
@@ -96,49 +97,28 @@ type UnsupportedServerConfigLegacy struct {
 // ToStartupConfig returns the given LegacyServerConfig as a StartupConfig and a set of DBConfigs.
 // The returned configs do not contain any default values - only a direct mapping of legacy config options as they were given.
 func (lc *LegacyServerConfig) ToStartupConfig() (*StartupConfig, DbConfigMap, error) {
-
 	// find a database's credentials for bootstrap (this isn't the first database config entry due to map iteration)
 	bsc := &BootstrapConfig{}
 	for _, dbConfig := range lc.Databases {
-		server := dbConfig.Server
-		if server == nil || *server == "" {
+		if dbConfig.Server == nil || *dbConfig.Server == "" {
 			continue
 		}
 
-		connSpec, err := gocbconnstr.Parse(*server)
+		server, username, password, err := gocbv1tov2AddressUpgrade(*dbConfig.Server)
 		if err != nil {
 			return nil, DbConfigMap{}, err
 		}
 
-		// support for username and password if not specified in username and password fields
-		splitServers := strings.Split(*server, ",")
-		// Only first URL is supported to have username and pass or else gocbconnstr.Parse does not work correctly
-		if strings.Contains(splitServers[0], "@") && dbConfig.Username == "" && dbConfig.Password == "" {
-			u, err := url.Parse(splitServers[0])
-			if err != nil {
-				return nil, DbConfigMap{}, err
-			}
-			username := u.User.Username()
-			password, passwordSet := u.User.Password()
-			if username != "" && passwordSet {
-				dbConfig.Username = username
-				dbConfig.Password = password
-			}
-		}
-
-		if connSpec.Scheme == "http" {
-			*server = "couchbase://"
-			for _, addr := range connSpec.Addresses {
-				*server = fmt.Sprintf("%s%s,", *server, addr.Host)
-			}
-			// Trailing , is valid but removed to make field neater
-			*server = strings.TrimSuffix(*server, ",")
+		// Prioritise config fields over credentials in host
+		if dbConfig.Username != "" && dbConfig.Password != "" {
+			username = dbConfig.Username
+			password = dbConfig.Password
 		}
 
 		bsc = &BootstrapConfig{
-			Server:              *server,
-			Username:            dbConfig.Username,
-			Password:            dbConfig.Password,
+			Server:              server,
+			Username:            username,
+			Password:            password,
 			CACertPath:          dbConfig.CACertPath,
 			X509CertPath:        dbConfig.CertPath,
 			X509KeyPath:         dbConfig.KeyPath,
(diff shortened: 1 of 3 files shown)


bbrks (changes requested):

rest/config_legacy.go
```diff
@@ -97,11 +97,17 @@ func (lc *LegacyServerConfig) ToStartupConfig() (*StartupConfig, DbConfigMap, er
...
 	for _, dbConfig := range lc.Databases {
-		if dbConfig.Server == nil || *dbConfig.Server == "" {
+		server := dbConfig.Server
+		if server == nil || *server == "" {
 			continue
 		}
+		if strings.HasPrefix(*server, "http://") {
+			*server = strings.TrimPrefix(*server, "http://")
+			*server = fmt.Sprintf("couchbase://%s", *server)
+		}
```
bbrks: I think there are a few cases here that will need to be handled for the [variations of `server` that we allow to be used](https://docs.couchbase.com/sync-gateway/current/configuration-properties.html#databases-this_db-server).

- `dbConfig.Server` can contain a list of server addresses, not just one. This is so that SG can fall back if the first server it attempts is unreachable.
  - Will need to do this protocol scheme switch for each address found.
- The HTTP port needs stripping when using `couchbase://` or else you get this error from the SDK:
  - `couchbase://host:8091 not supported for couchbase:// scheme. Use couchbase://host`
  - This is because when using `couchbase(s)://`, the SDK avoids HTTP altogether and talks directly with port `11210`/`11207`

It's starting to feel like we might want to make use of `gocbconnstr.Parse()` instead of string manipulation to transform the server URLs.

bbrks (changes requested):
Tested config upgrade with `http://localhost:8091` and it doesn't work as expected. I've left more info inline in the PR.

```
$ cat cnf_dbg.json
{
  "logging": {
    "console": {
      "log_level": "trace",
      "log_keys": ["*"]
    }
  },
  "use_tls_server": false,
  "server_tls_skip_verify": true,
  "databases": {
    "db1": {
	    "server":"http://localhost:8091",
	    "username":"Administrator",
	    "password":"password",
	    "bucket":"db1",
	    "num_index_replicas":0,
    "sync": `function(doc, oldDoc) {
        channel(doc.channels);
	access(doc.user, doc.access);
	role(doc.user, doc.role);
        for (val in doc) {
          console.log(val);
	}
    }`,
    "users":{
      "GUEST": {
        "disabled": false
      }
     }
    }
  }
}
```

```
$ go build && ./sync_gateway cnf_dbg.json
2021-08-13T09:40:10.118+01:00 ==== Couchbase Sync Gateway/() CE ====
2021-08-13T09:40:10.118+01:00 [INF] Loading content from [cnf_dbg.json] ...
2021-08-13T09:40:10.119+01:00 [INF] Found unknown fields in startup config. Attempting automatic config upgrade.
2021-08-13T09:40:10.119+01:00 [INF] Loading content from [cnf_dbg.json] ...
2021-08-13T09:40:37.192+01:00 [WRN] RetryLoop for Cluster Bootstrap giving up after 28 attempts -- base.RetryLoopCtx() at util.go:441
2021-08-13T09:40:37.192+01:00 [ERR] Couldn't start Sync Gateway: couchbase://host:8091 not supported for couchbase:// scheme. Use couchbase://host -- rest.ServerMain() at main.go:25
```

rest/config_legacy.go
```diff
@@ -97,11 +100,43 @@ func (lc *LegacyServerConfig) ToStartupConfig() (*StartupConfig, DbConfigMap, er
...
+
+		if connSpec.Scheme == "http" {
+			*server = "couchbase://"
+			for _, addr := range connSpec.Addresses {
+				*server = fmt.Sprintf("%s%s,", *server, addr.Host)
+			}
+			// Trailing , is valid but removed to make field neater
+			*server = strings.TrimSuffix(*server, ",")
+		}
```
bbrks: This whole block feels like it should be a separate function that can take a plain `gocbv1`/`http://` connection string and return a `gocbv2`/`couchbase://` compatible one.

It's almost the opposite of `CouchbaseURIToHttpURL` in `base/util.go`

Should make it easier to test more permutations of it too.

rest/config_legacy.go
```diff
@@ -97,11 +100,43 @@ func (lc *LegacyServerConfig) ToStartupConfig() (*StartupConfig, DbConfigMap, er
...
 		}
+
+		connSpec, err := gocbconnstr.Parse(*server)
+		if err != nil {
+			return nil, DbConfigMap{}, err
+		}
+
+		// support for username and password if not specified in username and password fields
+		splitServers := strings.Split(*server, ",")
+		// Only first URL is supported to have username and pass or else gocbconnstr.Parse does not work correctly
```
bbrks: This seems unexpected to me, but might just be undocumented behaviour of how it (or SG) used to work.

Would be worth double checking if SG 2.8 supports running with a username/password in a host that isn't first.

Did you mean `url.Parse` fails for comma-separated hosts, rather than `gocbconnstr.Parse`?

rest/config_legacy.go
```diff
@@ -4,9 +4,12 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"net/url"
 	"strings"
 	"time"
 
+	"gopkg.in/couchbaselabs/gocbconnstr.v1"
```
bbrks: Both libraries are (currently) the same, but `github.com/couchbase/gocbcore/v10/connstr` is the newer version/path for gocbv2, which I think we should be using instead.

```suggestion
	"github.com/couchbase/gocbcore/v10/connstr"
```

rest/config_legacy_test.go
```diff
@@ -71,6 +71,18 @@ func TestLegacyConfigToStartupConfig(t *testing.T) {
 			input:    LegacyServerConfig{},
 			expected: StartupConfig{API: APIConfig{AdminInterfaceAuthentication: base.BoolPtr(true)}},
 		},
+		{
+			name:     "http:// to couchbase://",
+			base:     StartupConfig{},
+			input:    LegacyServerConfig{Databases: DbConfigMap{"db": &DbConfig{BucketConfig: BucketConfig{Server: base.StringPtr("http://http.couchbase.com:2929,host2,host1:22,[2001:db8::8811],[2001:db8::8822]:888")}}}},
+			expected: StartupConfig{Bootstrap: BootstrapConfig{Server: "couchbase://http.couchbase.com,host2,host1,[2001:db8::8811],[2001:db8::8822]"}},
```
bbrks, line 78: Plenty of combinations across these 2 tests which is good to see, although I think we should also be covering these separate cases, to ensure we're not changing already-compatible connection strings.

- `http://foo,bar?foo=fooval&bar=barval` -> `couchbase://foo,bar?foo=fooval&bar=barval`
  - ensure all query parameters are preserved in the new URL
- `couchbase://foo:21210` -> `couchbase://foo:21210`
  - ensure custom ports for `couchbase(s)://` schemes are retained

bbrks (changes requested):
General approach looks good. A couple of comments for cleanup/test improvements, and one related to basic auth which is unexpected.

rest/config_legacy.go
IsaacLambat: `gocbconnstr.Parse` does not work as intended when the username and password is not in the first server address (as it causes the username to become a address and other unexpected behaviour).

After testing with SGW 2.8, the username and password only works if it is in the first server URL and then this username and password is used for the fallback URLs as well.

rest/config_legacy.go
```diff
@@ -236,6 +251,59 @@ func (lc *LegacyServerConfig) ToStartupConfig() (*StartupConfig, DbConfigMap, er
...
+	for _, addr := range connSpec.Addresses {
+		var port string
+		if addr.Port != 8091 && addr.Port != -1 {
+			if connSpec.Scheme != "http" {
+				port = ":" + strconv.Itoa(addr.Port)
+			} else {
+				return "", "", "", fmt.Errorf("automatic connection string conversion does not support non-default host ports. " +
+					"Please change the server field to use the couchbase(s):// scheme")
+			}
+		}
```
bbrks: Related to the `couchbase://...:8091` change suggested in the tests. I think this covers and simplifies a bit.

```suggestion
		if connSpec.Scheme == "http" && addr.Port != 8091 {
			return "", "", "", fmt.Errorf("automatic connection string conversion does not support non-default host ports. " +
				"Please change the server field to use the couchbase(s):// scheme")
		}
		if addr.Port != -1 {
			port = ":" + strconv.Itoa(addr.Port)
		}
```

bbrks (changes requested):
looks good apart from below comments

rest/config_legacy.go
```diff
@@ -274,14 +274,14 @@ func gocbv1tov2AddressUpgrade(server string) (newServer, username, password stri
 
 	var hosts string
 	for _, addr := range connSpec.Addresses {
+		if addr.Port != -1 && addr.Port != 8091 && connSpec.Scheme == "http" {
```
bbrks: Unset ports with `http://` are an implicit port `80`, which I think is safe to assume is a non-default Couchbase port, and should error, because we can't make the assumption that the memcached port is default.

rest/config_legacy.go
```diff
@@ -274,14 +274,14 @@ func gocbv1tov2AddressUpgrade(server string) (newServer, username, password stri
...
+
 		var port string
-		if addr.Port != 8091 && addr.Port != -1 {
-			if connSpec.Scheme != "http" {
-				port = ":" + strconv.Itoa(addr.Port)
-			} else {
-				return "", "", "", fmt.Errorf("automatic connection string conversion does not support non-default host ports. " +
-					"Please change the server field to use the couchbase(s):// scheme")
-			}
+		if addr.Port != -1 && connSpec.Scheme != "http" {
```
bbrks: If we remove the `!= -1` check above as suggested, we don't need the additional http scheme check here either

bbrks (changes requested):

rest/config_legacy.go
```diff
@@ -236,6 +251,62 @@ func (lc *LegacyServerConfig) ToStartupConfig() (*StartupConfig, DbConfigMap, er
...
 }
 
+func gocbv1tov2AddressUpgrade(server string) (newServer, username, password string, err error) {
+	connSpec, err := connstr.Parse(server)
+	if err != nil {
+		return "", "", "", err
+	}
+
+	splitServers := strings.Split(server, ",")
+	// Only first host is supported to have username and pass or else gocbconnstr.Parse does not work correctly
```
adamcfraser: The regex used by gocb for part splitting and host splitting doesn't look like it's changed between gocb v1 and v2.  Was gocb v1 doing pre-processing on username/password before using gocbconnstr?

rest/config_legacy.go
```diff
@@ -236,6 +251,62 @@ func (lc *LegacyServerConfig) ToStartupConfig() (*StartupConfig, DbConfigMap, er
...
+		}
+
+		hosts = hosts + addr.Host + port + ","
+	}
+
+	if connSpec.Scheme == "http" {
+		connSpec.Scheme = "couchbase"
+	}
+
+	server = connSpec.Scheme + "://" + hosts
```
adamcfraser: connstr.ConnSpec looks like it's already got a String() function - what's the reason to build up our own server string, instead of building a ConnSpec and using the existing String()?

rest/config_legacy.go
```diff
@@ -236,6 +251,62 @@ func (lc *LegacyServerConfig) ToStartupConfig() (*StartupConfig, DbConfigMap, er
...
+			username = urlUsername
+			password = urlPassword
+		}
+	}
+
+	var hosts string
+	for _, addr := range connSpec.Addresses {
+		if connSpec.Scheme == "http" {
+			if addr.Port != 8091 {
+				return "", "", "", fmt.Errorf("automatic connection string conversion does not support non-default host ports. " +
```
adamcfraser: Users probably won't know what triggered 'automatic connection string conversion' - maybe rephrase as "automatic migration connection string from http:// to couchbase:// scheme doesn't support non-default ports"

rest/config_legacy.go
```diff
@@ -236,6 +251,62 @@ func (lc *LegacyServerConfig) ToStartupConfig() (*StartupConfig, DbConfigMap, er
 	return &sc, dbs, nil
 }
 
+func gocbv1tov2AddressUpgrade(server string) (newServer, username, password string, err error) {
```
adamcfraser: I think this is better described as legacyServerAddressUpgrade, with a comment added about what work it's doing.

adamcfraser (changes requested):

rest/config_legacy.go
IsaacLambat: Had a look at what Sync Gateway 2.8.2 does and it uses GoLang Parse then fills the dbConfig username and password (if not set) if the user is set in the URL. Will implement this in to here for auto upgrade.

bbrks (changes requested):
Really minor defensive-only change, but otherwise looks good.

bbrks (dismissed):
lgtm

rest/config_legacy.go
```diff
@@ -93,17 +96,28 @@ type UnsupportedServerConfigLegacy struct {
...
 	// find a database's credentials for bootstrap (this isn't the first database config entry due to map iteration)
 	bsc := &BootstrapConfig{}
 	for _, dbConfig := range lc.Databases {
 		if dbConfig.Server == nil || *dbConfig.Server == "" {
 			continue
 		}
+
+		server, username, password, err := legacyServerAddressUpgrade(*dbConfig.Server)
+		if err != nil {
+			return nil, nil, err
```
adamcfraser: Is it sufficient just to log the error here, and then set `server = *dbConfig.Server`?

rest/config_legacy.go
```diff
@@ -4,9 +4,12 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"net/url"
 	"strings"
 	"time"
 
+	"github.com/couchbase/gocbcore/v10/connstr"
+
```
adamcfraser: whitespace can be removed

rest/config_legacy.go
```diff
@@ -246,6 +260,41 @@ func (dbc *DbConfig) ToDatabaseConfig() *DatabaseConfig {
...
+		if urlUsername != "" && urlPasswordSet {
+			username = urlUsername
+			password = urlPassword
+		}
+	}
+
+	if connSpec.Scheme == "http" {
+		connSpec.Scheme = "couchbase"
+		for i, addr := range connSpec.Addresses {
+			if addr.Port != 8091 {
```
adamcfraser: I thought we previously supported using http:// without specifying a port.  Does connstr.Parse fill in the default port if it's not specified?

adamcfraser (changes requested):

rest/config_legacy.go
IsaacLambat: No, connstr.Parse just sets the port as -1 however the latest stages of Gocb handle if it's -1 then it uses 8091

adamcfraser (approved):

PR #5170, opened 2021-08-11
Store docID in AllowedAttachment to support CBMobile_2 retrieval

adamcfraser:
Supports V2 attachment retrieval when clients aren't specifying docID in getAttachment request.


sarathkumarsivan (commented):
It looks good to me.

sarathkumarsivan (approved):
Approving the request.

PR #5174, opened 2021-08-16
CBG-1404 Automatic removal of Lithium attachments when they are removed from the document or the document is tombstoned or purged

sarathkumarsivan:
Handling for removal of Lithium attachments when they are removed from the document or the document is tombstoned or purged.

### DIFF
--- a/db/attachment.go
+++ b/db/attachment.go
@@ -130,9 +130,9 @@ func (db *Database) storeAttachments(doc *Document, newAttachmentsMeta Attachmen
 	return newAttachmentData, nil
 }
 
-// retrieveAttachments returns the list of keys from the attachment metadata that can be used for identifying obsolete
-// attachments and triggering subsequent removal of those attachments to reclaim the storage.
-func retrieveAttachments(docID string, docAttachments AttachmentsMeta) (attachments map[string]struct{}, err error) {
+// retrieveV2AttachmentKeys returns the list of V2 attachment keys from the attachment metadata that can be used for
+// identifying obsolete attachments and triggering subsequent removal of those attachments to reclaim the storage.
+func retrieveV2AttachmentKeys(docID string, docAttachments AttachmentsMeta) (attachments map[string]struct{}, err error) {
 	attachments = make(map[string]struct{})
 	for _, value := range docAttachments {
 		meta, ok := value.(map[string]interface{})
--- a/db/attachment.go
+++ b/db/attachment.go
@@ -26,8 +26,13 @@ const (
 	AttVersion2 int = 2
 )
 
-// ErrAttachmentVersion is thrown in case of any error in parsing version from the attachment meta.
-var ErrAttachmentVersion = base.HTTPErrorf(http.StatusBadRequest, "invalid version found in attachment meta")
+var (
+	// ErrAttachmentVersion is thrown in case of any error in parsing version from the attachment meta.
+	ErrAttachmentVersion = base.HTTPErrorf(http.StatusBadRequest, "invalid version found in attachment meta")
+
+	// ErrAttachmentMeta returned when the document contains invalid _attachments metadata properties.
+	ErrAttachmentMeta = base.HTTPErrorf(http.StatusBadRequest, "Invalid _attachments")
+)
 
 // AttachmentData holds the attachment key and value bytes.
 type AttachmentData map[string][]byte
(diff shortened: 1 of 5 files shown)


db/attachment.go
```diff
@@ -125,6 +130,53 @@ func (db *Database) storeAttachments(doc *Document, newAttachmentsMeta Attachmen
 	return newAttachmentData, nil
 }
 
+// retrieveObsoleteAttachments determines the list of attachments that are no longer being referenced from any document
+// and returns a slice of attachments keys for subsequent garbage collection.
+func (db *Database) retrieveObsoleteAttachments(doc *Document, docAttachments AttachmentsMeta, parentRev string, docHistory []string) (obsoleteAttachments []string, err error) {
```
adamcfraser: This is comparing attachments on the current rev with attachments on the current rev's ancestor.  I don't think this accounts for attachments associated with conflicting revisions (which will be stored as _attachment in the conflicting revision body, see storeOldBodyInRevTreeAndUpdateCurrent).  I think we want to compare the set (current doc attachments + all conflicting rev body attachments) with (new doc attachments + all conflicting rev body attachments).  Otherwise we risk deleting an attachment referenced on a non-winning revision, and not have a way to recreate that attachment if that revision becomes the active rev.

Similarly, I don't think parentRev should be used, as that's not necessarily the parent of the active revision (e.g. the case where someone updates a conflicting branch, but the new revision is still on the losing branch).   There are quite a few permutations there that should get unit test coverage.

adamcfraser (changes requested):

db/attachment.go
sarathkumarsivan: Obsolete attachment retrieval logic has been refactored and the retrieveObsoleteAttachments can be completely removed. The retrieveActiveAttachments has been remaned to retrieveAttachments and included the versions check while marking obsolete attachments.

db/crud.go
```diff
@@ -1909,6 +1928,24 @@ func (db *Database) updateAndReturnDoc(docid string, allowImport bool, expiry ui
 	// Now that the document has successfully been stored, we can make other db changes:
 	base.DebugfCtx(db.Ctx, base.KeyCRUD, "Stored doc %q / %q as #%v", base.UD(docid), newRevID, doc.Sequence)
 
+	// Delete obsolete attachments from the bucket.
+	if !inConflict {
```
adamcfraser: I filed CBG-1627 to add support for conflicted documents - can you add a TODO here referencing that?

rest/api_test.go
```diff
@@ -7918,3 +7918,722 @@ func TestDocChannelSetPruning(t *testing.T) {
...
+		rt.purgeDoc(docID)
+
+		// Check whether both attachments are removed from the underlying storage.
+		requireAttachmentNotFound(docID, att1Name)
+		rt.requireDocNotFound(att1Key)
+		requireAttachmentNotFound(docID, att2Name)
+		rt.requireDocNotFound(att2Key)
+	})
+
+	t.Run("single inline attachment removal upon document update", func(t *testing.T) {
```
adamcfraser, line 8409: Is there a difference between the inline and non-inline case, once the document has been created, other than the first case has one revision and the second has two?

rest/api_test.go
```diff
@@ -7918,3 +7918,722 @@ func TestDocChannelSetPruning(t *testing.T) {
...
+		// Retrieve new attachment key used for internal attachment storage and retrieval.
+		attKeyNew := retrieveAttachmentKey(docID, attName)
+		require.Equal(t, "_sync:att2:IfWNJ/gn0pX/zYYMZQRWheO68a1FBsqgFAETsxZkdTQ=:sha1-0naD6SgfLVDr+zakP8RkNlBYORw=", attKeyOld)
+		require.NotEqual(t, attKeyOld, attKeyNew)
+
+		// Perform cleanup after the test ends.
+		rt.purgeDoc(docID)
+	})
+
+	t.Run("multiple attachments removal upon document update", func(t *testing.T) {
```
adamcfraser, line 8165: Do you have test coverage for 'document with multiple attachments, and only one is removed'?

adamcfraser (changes requested):
Generally looks good, a few suggestions around test cases, and cleanup on the import handling.

db/crud.go
sarathkumarsivan: Thank you. Added the comments TODO comments referencing that ticket.

rest/api_test.go
sarathkumarsivan: There is no difference between the inline and non-inline case, once the document has been created, other than the first case has one revision and the second has two. However, this unit test has been included as an end-to-end coverage for inline flow.

rest/api_test.go
sarathkumarsivan: Included a new test "doc with multiple attachments and removal of a single one upon document update" that covers the case of document with multiple attachments, and only one is removed.

adamcfraser (approved):

PR #5176, opened 2021-08-17
CBG-1452: Persistent Config optimistic concurrency control

JRascagneres:
CBG-1452

Adds opt-in concurrency control for persistent config by utilizing rev IDs for configs which are supplied as ETags.

- On GET /db/_config the ETag is supplied
- On a PUT /db/_config the previous provided ETag  can be either emitted which will do an upsert, or can be provided (in the If-Match header) at which point we will first check that the current db config matches the provided ETag.
- Changed `fetchAndLoadDatabase` to split out the fetch work into `fetchDatabase`. Allowing is to obtain the current database config to validate the version.
- Renamed `adminRequest` to `bootstrapAdminRequest` to make its use-case more clear.
- In `_applyConfig` strip out Version

## Dependencies
- [x] PR #5155 

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1003/
- [x] `xattrs=false` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1004/
Failures look unrelated to this change

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -390,7 +390,7 @@ func (h *handler) handlePutDbConfig() (err error) {
 
 				headerVersion := h.rq.Header.Get("If-Match")
 				if headerVersion != "" && headerVersion != bucketDbConfig.Version {
-					return nil, base.HTTPErrorf(http.StatusConflict, "Provided If-Match header does not match current config version")
+					return nil, base.HTTPErrorf(http.StatusPreconditionFailed, "Provided If-Match header does not match current config version")
 				}
 
 				if err := base.ConfigMerge(&bucketDbConfig, dbConfig); err != nil {
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -44,6 +44,11 @@ func (h *handler) handleCreateDB() error {
 			return base.HTTPErrorf(http.StatusBadRequest, err.Error())
 		}
 
+		config.Version, err = GenerateDatabaseConfigVersionID("", config)
+		if err != nil {
+			return err
+		}
+
 		config.cas, err = h.server.bootstrapContext.connection.InsertConfig(*config.Bucket, h.server.config.Bootstrap.ConfigGroupID, config)
 		if err != nil {
 			// remove database if we can't persist to avoid inconsistent cluster state
@@ -135,6 +140,19 @@ func (h *handler) handleGetDbConfig() error {
 	includeJavascript, _ := h.getOptBoolQuery("include_javascript", true)
 	_ = includeJavascript
 
+	if h.server.bootstrapContext.connection != nil {
+		found, dbConfig, err := h.server.fetchDatabase(h.db.Name)
+		if err != nil {
+			return err
+		}
+
+		h.response.Header().Set("ETag", dbConfig.Version)
+
+		if !found {
+			return base.HTTPErrorf(http.StatusNotFound, "database config not found")
+		}
+	}
+
 	redact, _ := h.getOptBoolQuery("redact", true)
 	if redact {
 		cfg, err := h.server.GetDatabaseConfig(h.db.Name).Redacted()
@@ -145,6 +163,7 @@ func (h *handler) handleGetDbConfig() error {
 	} else {
 		h.writeJSON(h.server.GetDatabaseConfig(h.db.Name))
 	}
+
 	return nil
 }
 
@@ -369,7 +388,11 @@ func (h *handler) handlePutDbConfig() (err error) {
 					return nil, err
 				}
 
-				// TODO: CBG-1452 Optimistic Concurrency Control/RevID check
+				headerVersion := h.rq.Header.Get("If-Match")
+				if headerVersion != "" && headerVersion != bucketDbConfig.Version {
+					return nil, base.HTTPErrorf(http.StatusConflict, "Provided If-Match header does not match current config version")
+				}
+
 				if err := base.ConfigMerge(&bucketDbConfig, dbConfig); err != nil {
 					return nil, err
 				}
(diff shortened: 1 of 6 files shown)


rest/admin_api.go
```diff
@@ -365,7 +384,11 @@ func (h *handler) handlePutDbConfig() (err error) {
 					return nil, err
 				}
 
-				// TODO: CBG-1452 Optimistic Concurrency Control/RevID check
+				headerVersion := h.rq.Header.Get("If-Match")
+				if headerVersion != "" && headerVersion != bucketDbConfig.Version {
+					return nil, base.HTTPErrorf(http.StatusConflict, "Provided If-Match header does not match current config version")
```
bbrks: `http.StatusConflict` is used throughout SG for the same type of situation on docs, but RFC7232 suggests that `http.StatusPreconditionFailed` is the appropriate status code when a precondition header like `If-Match` is not satisfied.

rest/bootstrap_test.go
```diff
@@ -165,6 +165,23 @@ func adminRequest(t *testing.T, method, path, body string) *http.Response {
 	return resp
 }
 
+func adminRequestWithHeaders(t *testing.T, method, path, body string, headers map[string]string) *http.Response {
```
bbrks: I named `adminRequest` above quite poorly, it's a bit too easy to misuse outside of the context of a bootstrap test. Would suggest this to avoid accidental usage (the `4985+bootstrapTestPortOffset` thing is very much specific to one small set of tests)

```suggestion
func bootstrapAdminRequestWithHeaders(t *testing.T, method, path, body string, headers map[string]string) *http.Response {
```

rest/config.go
```diff
@@ -965,9 +965,19 @@ func (sc *ServerContext) fetchAndLoadConfigs() (count int, err error) {
 // fetchAndLoadDatabase will attempt to find the given database name first in a matching bucket name,
 // but then fall back to searching through configs in each bucket to try and find a config.
 func (sc *ServerContext) fetchAndLoadDatabase(dbName string) (found bool, err error) {
+	found, dbConfig, err := sc.fetchDatabase(dbName)
+	if err != nil || !found {
+		return false, err
+	}
+	sc.applyConfigs([]DatabaseConfig{*dbConfig})
+
+	return true, err
```
bbrks:
```suggestion
	return true, nil
```

rest/admin_api_test.go
```diff
@@ -3370,3 +3370,47 @@ func TestIncludeRuntimeStartupConfig(t *testing.T) {
...
+	resp := adminRequest(t, http.MethodPut, "/db/",
+		`{"bucket": "`+tb.GetName()+`", "num_index_replicas": 0}`,
+	)
+	assert.Equal(t, http.StatusCreated, resp.StatusCode)
+
+	// Get config
+	resp = adminRequest(t, "GET", "/db/_config?redact=false", "")
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+
+	eTag := resp.Header.Get("ETag")
```
bbrks, line 3406: Might want to check this, as I'm not sure the assertion below will fail if etag is an empty string (it will just always do an unsafe PUT)

```suggestion
	eTag := resp.Header.Get("ETag")
	assert.NotEqual(t, "", eTag)
```

bbrks (changes requested):
Approach looks good to me. A couple of comments below.

rest/admin_api_test.go
JRascagneres: Good point

rest/bootstrap_test.go
JRascagneres: Also opted to rename the existing adminRequest. I know it doesn't make the PR quite as clean to review but as you say, the naming can be improved so decided to do that here.

rest/admin_api.go
```diff
@@ -375,6 +398,11 @@ func (h *handler) handlePutDbConfig() (err error) {
 					return nil, err
 				}
 
+				bucketDbConfig.Version, err = GenerateDatabaseConfigVersionID(bucketDbConfig.Version, &bucketDbConfig)
```
bbrks, line 413: Similarly, don't we need to prevent users from specfiying `Version` in the REST API body?

bbrks (changes requested):
I think we're exposing `Version` in the config via the REST API, instead of only headers.

rest/admin_api.go
```diff
@@ -131,6 +136,19 @@ func (h *handler) handleGetDbConfig() error {
...
+		if err != nil {
+			return err
+		}
+
+		h.response.Header().Set("ETag", dbConfig.Version)
+
+		if !found {
+			return base.HTTPErrorf(http.StatusNotFound, "database config not found")
+		}
+	}
```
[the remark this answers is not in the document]
JRascagneres: Don't think so. We're not returning the same config that we're fetching here. dbConfig that is being used for version isn't the one that we're returning in the body.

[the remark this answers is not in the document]
bbrks: Oops yeah. Ok, make sure `GetDatabaseConfig` isn't populating Version then :D

rest/admin_api.go
JRascagneres: Yeah granted, this should be done

rest/admin_api.go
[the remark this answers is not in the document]
JRascagneres: Yeah true, did some testing I think it is. I've made `_applyConfig` strip it out. We have no use for version locally so I think this is probably nicer than stripping it out in places. Given the whole point of it requires accessing the current bucket state

bbrks (dismissed):
LGTM. We can revisit the DatabaseConfig struct to evaluate what needs storing but stripping from responses.

PR #5178, opened 2021-08-18
CBG-1603 CBG-1604 CBG-1605: Implement new /{db}/_config options (include_runtime, include_javascript)

bbrks:
CBG-1603 CBG-1604 CBG-1605

Implement new /{db}/_config options:
- `include_runtime` determines whether the config returned is from the bucket, or from the loaded in-memory config
- `include_javascript` determines whether config fields containing javascript are included in the response

Utilises existing dbconfig bucket load from CBG-1452 for optimistic concurrency control ETag header to reduce number of bucket ops to 1 for all scenarios.

- [x] PR #5155
  - [x] PR #5176

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -159,6 +159,8 @@ func (h *handler) handleGetDbConfig() error {
 
 	// refresh_config=true forces the config loaded out of the bucket to be applied on the node
 	if h.getBoolQuery("refresh_config") && h.server.bootstrapContext.connection != nil {
+		// set cas=0 to force a refresh
+		bucketDbConfig.cas = 0
 		h.server.applyConfigs([]DatabaseConfig{*bucketDbConfig})
 	}
 
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -136,21 +136,16 @@ func (h *handler) handleDbOffline() error {
 
 // Get admin database info
 func (h *handler) handleGetDbConfig() error {
-	if h.getBoolQuery("refresh_config") && h.server.bootstrapContext.connection != nil {
-		_, err := h.server.fetchAndLoadDatabase(h.db.Name)
-		if err != nil {
-			return err
-		}
-	}
 
-	// TODO: STUBS
-	includeRuntime, _ := h.getOptBoolQuery("include_runtime", false)
-	_ = includeRuntime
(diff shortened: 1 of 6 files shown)


bbrks:
@adamcfraser would like a review on this if possible - even though it's stacked up behind two PRs. The diff is fairly small and isolated when based on top of those.

rest/config.go
```diff
@@ -711,6 +711,13 @@ func (dbConfig *DbConfig) Redacted() (*DbConfig, error) {
 		return nil, err
 	}
 
+	err = config.redactInPlace()
+	return &config, err
+}
+
+// redactInPlace modifies the given config to redact the fields inside it.
+func (config *DbConfig) redactInPlace() error {
+
```
adamcfraser, line 720: Not strictly related to this PR, but should we only be redacting the password when it's non-empty?

rest/admin_api.go
```diff
@@ -150,20 +145,47 @@ func (h *handler) handleGetDbConfig() error {
 			return base.HTTPErrorf(http.StatusNotFound, "database config not found")
 		}
 
-		h.response.Header().Set("ETag", dbConfig.Version)
+		h.response.Header().Set("ETag", bucketDbConfig.Version)
+	}
+
+	// refresh_config=true forces the config loaded out of the bucket to be applied on the node
```
adamcfraser, line 160: Does refresh_config force reload even if the versions already match?  Or is that handled lower down.

adamcfraser (changes requested):
Generally looks good, a couple of questions.

rest/admin_api.go
bbrks: Nope, it will skip applying the config if the CAS from the bucket matches the one we have already loaded in-memory. This is done down inside `_applyConfig` with some debug logging.

```
[DBG] Database %q bucket %q config has not changed since last update.
```

I've added an extra case (which currently isn't exercised) where when config.cas=0 the config update is always applied, to force a reload.

rest/config.go
bbrks: sure can do

rest/admin_api.go
bbrks: As discussed yesterday, pushed a commit to make `refresh_config` force a reload even if the in-memory database is up to date.

adamcfraser (approved):

PR #5182, opened 2021-08-19
CBG-1561: Add permission checks to _config

JRascagneres:
CBG-1561
PR aims to support fine grained permissions on `PUT` `/db/_config`. 
We do this by first unmarshalling into a map so we can check the keys set on the payload then based on this and the perms we have we can allow various scenarios.

Required changes on the permissions side as there was an issue there. Code should be commented fairly well so probably not worth repeating here.

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -9,8 +9,11 @@
 package rest
 
 import (
+	"bytes"
 	"fmt"
+	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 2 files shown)


bbrks (approved):
Config permission checks look good, but want Adam to take a look through the updated `checkAdminAuth` logic.

adamcfraser (approved):

PR #5184, opened 2021-08-19
CBG-1606 CBG-1607: Implement sync and import_filter endpoints

bbrks:
CBG-1606 CBG-1607

Allows independent read and write of sync function and import_filter for database configs.

Both PUT endpoint handlers share a lot of code with `handlePutDbConfig` which could potentially be factored out into shared functions.

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -535,17 +535,88 @@ func (h *handler) handlePutDbConfigSync() error {
 // GET database config import filter function
 func (h *handler) handleGetDbConfigImportFilter() error {
 	h.assertAdminOnly()
-	// TODO: STUB
-	h.writeJSONStatus(http.StatusOK, `function(doc) {
-	// this is a stub import filter function
-}`)
+	var (
+		etagVersion          string
+		importFilterFunction string
+	)
+
+	if h.server.bootstrapContext.connection != nil {
+		found, dbConfig, err := h.server.fetchDatabase(h.db.Name)
+		if err != nil {
+			return err
+		}
+		if !found {
+			return base.HTTPErrorf(http.StatusNotFound, "database config not found")
+		}
+		etagVersion = dbConfig.Version
+		if dbConfig.ImportFilter != nil {
+			importFilterFunction = *dbConfig.ImportFilter
+		}
+	}
+
+	h.response.Header().Set("ETag", etagVersion)
+	h.writeJavascript(importFilterFunction)
 	return nil
 }
 
 // PUT a new database config import filter function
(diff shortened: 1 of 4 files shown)


rest/admin_api.go
```diff
@@ -445,34 +445,178 @@ func (h *handler) handlePutDbConfig() (err error) {
 // GET database config sync function
 func (h *handler) handleGetDbConfigSync() error {
 	h.assertAdminOnly()
-	// TODO: STUB
-	h.writeJSONStatus(http.StatusOK, `function(doc, oldDoc) {
-	// this is a stub sync function
-}`)
+	var (
```
adamcfraser, line 448: Do we also need to support DELETE for the sync and import_filter endpoints?

rest/handler.go
```diff
@@ -638,6 +638,23 @@ func (h *handler) readSanitizeJSON(val interface{}) error {
 	return err
 }
 
+// readJavascript reads a javascript function from a request body.
+func (h *handler) readJavascript() (string, error) {
+	// Performs the Content-Type validation and Content-Encoding check.
+	input, err := processContentEncoding(h.rq.Header, h.requestBody, "text/javascript")
```
adamcfraser: I thought text/javascript was deprecated in favour of application/javascript.  
https://www.iana.org/assignments/media-types/media-types.xhtml#application
https://www.iana.org/assignments/media-types/media-types.xhtml#text

I might be missing a subtlety there, though - let me know.  Could also consider accepting both, if that's appropriate.

rest/handler.go
```diff
@@ -783,6 +800,19 @@ func (h *handler) writeJSON(value interface{}) {
 	h.writeJSONStatus(http.StatusOK, value)
 }
 
+func (h *handler) writeJavascript(js string) {
+	if !h.requestAccepts("text/plain") {
```
adamcfraser: We're checking for text/plain, but then sending as text/javascript.  Is that intentional?  

Also same question as above about application/javascript instead of text/javascript.

adamcfraser (changes requested):
A few questions/comments.

rest/handler.go
bbrks: Oops, copy-paste error. I've pulled this stuff out into a `writeWithMimetypeStatus` to avoid the duplication of mime types.

rest/admin_api.go
bbrks: Sure. I still need to find a way of reducing code duplication between all of these endpoints modifying a single field, but copy-pasted for now to get them working for the beta.

adamcfraser (approved):

PR #5190, opened 2021-08-20
CBG-1560 - New persistent example configs

IsaacLambat:
CBG-1560

- Added example configs
- Fixed failing unit test TestLoadServerConfigExamples


examples/legacy/serviceconfig.json
```diff
@@ -1,4 +1,5 @@
 {
+  "disable_persistent_config": true,
```
adamcfraser: The service scripts expect to find serviceconfig.json in /examples, so I think that moving this is going to break the service scripts.  For the time being I suggest just leaving this file in place, and not moving to legacy.

examples/README.md
```diff
@@ -1,5 +1,52 @@
+## Example Startup Configurations
+Here are a list of example Sync Gateway Startup configurations found in `examples/startup_configs`:
```
adamcfraser: It's a bit of a nit, but I think 'examples/startup_config' is better than 'examples/startup_configs' (same for database_configs), and for consistency I think we should use 'legacy_config'.

examples/startup_configs/https/README.md
```diff
@@ -1,15 +1,15 @@
-# examples/ssl
+# examples/startup_configs/tls
```
adamcfraser: It would be preferable to leave this in examples/ssl I think - it's harder to find in startup_configs/https.  You also name it startup_configs/tls here, but startup_configs/https in the path name.  For now let's just leave it where it is, and just update ssl.json to have what you've currently got in startup_configs/https/https.json.

adamcfraser (changes requested):

adamcfraser (approved):

PR #5195, opened 2021-08-23
CBG-1632 - Fixed registerLegacyFlags always filling adminInterface and publicInterface with default address if flag not set

IsaacLambat:
CBG-1632

- registerLegacyFlags only sets startup config fields `PublicInterface, AdminInterface, and Pretty` if the flag is set to something different then the default value. This way the default value does not override user set values.


bbrks (approved):
LGTM

PR #5202, opened 2021-08-24
CBG-1615: Don't flush to disabled log files

JRascagneres:
CBG-1615

Adds a basic enabled check to `FlushBufferToLog`. 
Previously even if a log level was not enabled we would flush the initial in-memory log lines to that level. 

This has been tested with an additional unit test.
Integration test should not be necessary for this as log file changes can be adequately tested with basic unit tests.

### DIFF
--- a/base/logger_file.go
+++ b/base/logger_file.go
@@ -117,10 +117,12 @@ func (l *FileLogger) FlushBufferToLog() {
 	// Need to clear hanging new line to avoid empty line
 	logString := strings.TrimSuffix(l.buffer.String(), "\n")
 
+	// Clear buffer as we no longer need it
+	l.buffer.Reset()
+
 	if l.Enabled.IsTrue() {
 		l.logf(logString)
 	}
-
 }
 
 // Rotate will rotate the active log file.
--- a/base/logger_file.go
+++ b/base/logger_file.go
@@ -116,7 +116,11 @@ func NewFileLogger(config *FileLoggerConfig, level LogLevel, name string, logFil
 func (l *FileLogger) FlushBufferToLog() {
 	// Need to clear hanging new line to avoid empty line
 	logString := strings.TrimSuffix(l.buffer.String(), "\n")
-	l.logf(logString)
+
+	if l.Enabled.IsTrue() {
+		l.logf(logString)
+	}
+
 }
 
 // Rotate will rotate the active log file.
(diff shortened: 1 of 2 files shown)


base/logger_file.go
```diff
@@ -116,7 +116,11 @@ func NewFileLogger(config *FileLoggerConfig, level LogLevel, name string, logFil
 func (l *FileLogger) FlushBufferToLog() {
```
bbrks, line 116: Super-nit but slightly earlier return opportunity

```go
func (l *FileLogger) FlushBufferToLog() {
	if !l.Enabled.IsTrue() {
		return
	}

	// Need to clear hanging new line to avoid empty line
	logString := strings.TrimSuffix(l.buffer.String(), "\n")
	l.logf(logString)
}
```

rest/server_context_test.go
```diff
@@ -522,3 +524,113 @@ func TestUseTLSServer(t *testing.T) {
...
+
+			// Setup logging
+			err = config.SetupAndValidateLogging()
+			assert.NoError(t, err)
+
+			// Flush memory loggers
+			base.FlushLoggerBuffers()
+
+			// Add a minor sleep to allow the files to complete writing
+			time.Sleep(1 * time.Second)
```
bbrks: Do we have to have a fixed sleep if we're checking in a retry loop? Maybe just increase the sleeper func's iterations or wait time.

bbrks (changes requested):
looks good, just nitpicks

base/logger_file.go
JRascagneres: Don't think this is worth based on discussed changes to add reset.

rest/server_context_test.go
JRascagneres: I think this should still be kept to stay safe.

Imagine in the default case we're waiting for 4 log files. If we hit the sleeper iteration and for whatever reason 'trace' was about to write but didn't quite get there (which I did see once in my tests) then we'd pass. 

Sleep gives an extra second for those writes to give us a little extra confidence any log files have written. Probably means we don't really need to have the sleeper too but will leave.

bbrks: I'd have expected the FlushLoggerBuffers call to be synchronous as long as we don't have collation buffers enabled (which we don't appear to for the memory loggers) but maybe that assumption is wrong

JRascagneres: I think it may well be collation buffers, I can turn them off. Why do you think we don't have them enabled?

bbrks (approved):

PR #5203, opened 2021-08-25
CBG-1635 Close gocbv2 collections per test

adamcfraser:
Have testBucketPool use a single cluster instance per collection, instead of sharing across test runs, as gocbv2 only closes bucket/collection pipelines on cluster close.
- [ ] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/view/Build/job/sync-gateway-integration/1044/

### DIFF
--- a/base/collection.go
+++ b/base/collection.go
@@ -130,7 +130,9 @@ func (c *Collection) UUID() (string, error) {
 }
 
 func (c *Collection) Close() {
-	// No close handling for collection
+	if c.cluster != nil {
+		c.cluster.Close(nil)
+	}
 	return
 }
 
--- a/base/collection.go
+++ b/base/collection.go
@@ -131,7 +131,9 @@ func (c *Collection) UUID() (string, error) {
 
 func (c *Collection) Close() {
 	if c.cluster != nil {
-		c.cluster.Close(nil)
+		if err := c.cluster.Close(nil); err != nil {
+			Warnf("Error closing collection cluster: %v", err)
+		}
 	}
 	return
 }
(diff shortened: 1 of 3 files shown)


base/collection.go
```diff
@@ -130,7 +130,9 @@ func (c *Collection) UUID() (string, error) {
 }
 
 func (c *Collection) Close() {
-	// No close handling for collection
+	if c.cluster != nil {
+		c.cluster.Close(nil)
```
bbrks:
```suggestion
		if err := c.cluster.Close(nil); err != nil {
			Warnf("Error closing collection cluster: %v", err)
		}
```

base/main_test_cluster.go
```diff
@@ -202,7 +202,11 @@ func initV2Cluster(server string) *gocb.Cluster {
 }
 
 func (c *tbpClusterV2) getBucketNames() ([]string, error) {
-	manager := c.cluster.Buckets()
+
+	cluster := getCluster(c.server)
+	defer cluster.Close(nil)
```
bbrks: I was thinking of a `closeCluster` helper here (and in the other defer cases below) to keep errcheck happy without the awkward inline function pattern.

```go
func (c *tbpClusterV2) closeCluster(cluster *gocb.Cluster) {
	if err := cluster.Close(nil); err != nil {
		c.logger(context.Background(), "Couldn't close cluster connection: %v", err)
	}
}
```

```suggestion
	defer c.closeCluster(cluster)
```

bbrks (changes requested):
Looks good. Ran several times locally and it's managed to address the leak without any actual runtime cost.

Couple of comments below to keep errcheck happy and log if we hit them.

bbrks (approved):

PR #5204, opened 2021-08-25
CBG-1638: Added parsing for legacy flags

IsaacLambat:
CBG-1638

- Added parsing for the legacy flags with appropriate deprecation/removal notices.
- Added testing for legacy flags

Legacy flags overwrite persistent config flags if both are set.

## Deprecation messages
If deprecated legacy flag used: `[WRN] Flag "-url" is deprecated. Please use "-bootstrap.server" in future.`
`Flag "-verbose" is deprecated. Please use "-logging.console.log_level info" in future.`

If `verbose` flag is used with logging.console.log_level: `Cannot use deprecated flag "-verbose" with flag "-logging.console.log_level". To set Sync Gateway to be verbose, please use flag "-logging.console.log_level info". Ignoring flag...`

If flag `configServer` is used: `flag "-configServer" is no longer supported and has been removed`

If flag `dbname` or `deploymentID` is used: `Flag "-deploymentID" is no longer supported and has been removed`

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1047
- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1050
- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1171 (flakey test fail)

## Dependencies

- [x] https://github.com/couchbase/sync_gateway/pull/5203

## Merge post-beta
- [x] Merge post-beta

### DIFF
--- a/rest/main.go
+++ b/rest/main.go
@@ -347,7 +347,7 @@ func parseFlags(args []string) (flagStartupConfig *StartupConfig, fs *flag.FlagS
 	fs = flag.NewFlagSet(args[0], flag.ContinueOnError)
 
 	// used by service scripts as a way to specify a per-distro defaultLogFilePath
-	defaultLogFilePath = *fs.String("defaultLogFilePath", "", "Path to log files, if not overridden by --logFilePath, or the config")
+	defaultLogFilePathFlag := fs.String("defaultLogFilePath", "", "Path to log files, if not overridden by --logFilePath, or the config")
 
 	disablePersistentConfig = fs.Bool("disable_persistent_config", false, "Can be set to false to disable persistent config handling, and read all configuration from a legacy config file.")
 
@@ -376,6 +376,8 @@ func parseFlags(args []string) (flagStartupConfig *StartupConfig, fs *flag.FlagS
 	if err != nil {
 		return nil, nil, nil, err
 	}
+
+	defaultLogFilePath = *defaultLogFilePathFlag
 
 	return &legacyStartupConfig, fs, disablePersistentConfig, nil
 }
--- a/rest/main.go
+++ b/rest/main.go
@@ -34,20 +34,8 @@ func serverMain(ctx context.Context, osArgs []string) error {
 	base.InitializeMemoryLoggers()
 	base.LogSyncGatewayVersion()
 
-	fs := flag.NewFlagSet(osArgs[0], flag.ContinueOnError)
-
-	// used by service scripts as a way to specify a per-distro defaultLogFilePath
-	defaultLogFilePathFlag := fs.String("defaultLogFilePath", "", "Path to log files, if not overridden by --logFilePath, or the config")
-
-	disablePersistentConfigFlag := fs.Bool("disable_persistent_config", false, "Can be set to false to disable persistent config handling, and read all configuration from a legacy config file.")
-
-	// register config property flags
-	flagStartupConfig := NewEmptyStartupConfig()
-
-	configFlags := registerConfigFlags(&flagStartupConfig, fs)
-	legacyConfigFlags := registerLegacyFlags(&flagStartupConfig, fs)
-
-	if err := fs.Parse(osArgs[1:]); err != nil {
+	flagStartupConfig, fs, disablePersistentConfig, err := parseFlags(osArgs)
+	if err != nil {
 		// Return nil for ErrHelp so the shell exit code is 0
 		if err == flag.ErrHelp {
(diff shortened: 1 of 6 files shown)


rest/main_legacy.go
```diff
@@ -58,67 +64,67 @@ func legacyServerMain(osArgs []string, flagStartupConfig *StartupConfig) error {
...
-	configServer := fs.String("configServer", "", "URL of server that can return database configs")
-	if *configServer != "" {
-		//
-	}
-	deploymentID := fs.String("deploymentID", "", "Customer/project identifier for stats reporting")
-	if *deploymentID != "" {
-		//
+func registerLegacyFlags(config *StartupConfig, fs *flag.FlagSet) map[string]legacyConfigFlag {
+	return map[string]legacyConfigFlag{
+		"interface":        {&config.API.PublicInterface, "api.public_interface", fs.String("interface", DefaultPublicInterface, "Address to bind to")},
```
bbrks: I think we should include some kind of `"DEPRECATED: "` prefix in front of the descriptions of these flags. It's difficult to tell you shouldn't use them ahead of time via `-help`.

rest/main_legacy_test.go
```diff
@@ -0,0 +1,68 @@
...
+	"flag"
+	"testing"
+
+	"github.com/couchbase/sync_gateway/base"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Test legacy flags using valid values
+func TestLegacyFlagsValid(t *testing.T) {
```
bbrks, line 13: Is it possible to write a test that ensures legacy flags do not override non-legacy flags?
Especially around the special handling for `-verbose`/`-logging.console.log_level` and "normal" handling for something like `-url`/`-bootstrap.server`

rest/main.go
```diff
@@ -60,9 +55,14 @@ func serverMain(ctx context.Context, osArgs []string) error {
...
 	}
 
-	err = fillConfigWithFlags(fs, configFlags)
+	err := fillConfigWithFlags(fs, configFlags)
 	if err != nil {
-		return err
+		return fmt.Errorf("error merging flags on to config: %w", err)
+	}
+
+	err = fillConfigWithLegacyFlags(legacyConfigFlags, fs, flagStartupConfig.Logging.Console.LogLevel != nil)
```
bbrks: I think filling legacy flags AFTER the normal flags will make them override non-legacy right?

That doesn't sound like something we want to happen.

bbrks (changes requested):
Some usability concerns below

rest/main.go
IsaacLambat: Yes, I can do that. The reason I made it not do that is so that it can check if `Logging.Console.LogLevel` has been set by the user, so then a warning can be issued if `verbose` is set.

Will swap the order, and use mergo to make sure persistent config flags take priority.

IsaacLambat:
Todo: 

Check if ParseCommandLine function needs the legacy flags inside of it.

bbrks:
> Todo:
> 
> Check if ParseCommandLine function needs the legacy flags inside of it.

@IsaacLambat Is this still to do, or is it ready for review (after resolving conflict)?

rest/main_test.go
```diff
@@ -32,3 +35,81 @@ func TestMain(m *testing.M) {
...
+		},
+	}
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			_, _, disablePersistentConfig, err := parseFlags(append(osArgsPrefix, test.osArgs...))
+			if test.expectError {
+				assert.Error(t, err)
+				if test.expectedError != nil {
+					assert.EqualError(t, test.expectedError, err.Error())
+				}
```
bbrks: Would prefer something like this - to comprehensively cover error cases, even if that means adding an expected error for the unknown flag case.

Prior to this suggestion, the test would pass if we weren't expecting an error but actually got one.

```suggestion
			if test.expectedError != nil {
				require.Error(t, err)
				assert.Contains(t, err.Error(), test.expectedError.Error())
			} else {
				assert.NoError(t, err)
```

bbrks (changes requested):
Just a small tweak to a test to cover error cases - other than that, this is OK to merge.

bbrks (approved):

PR #5210, opened 2021-08-27
CBG-1650 - added runtime config and database configs to sgcollect

IsaacLambat:
CBG-1650

- Renamed "Running server config" to "Collect server config"
- Added task to collect runtime config
- Added task to loop over database configs and collect them

Tested with single DB and multiple, as well as partial redaction. Partial redaction caused UD tags to be replaced with hashes as intended. This was verified through the username config option for both the server configs, and the database configs. These are not hashed when redaction is "none" or default.

### DIFF
--- a/tools/sgcollect_info
+++ b/tools/sgcollect_info
@@ -440,7 +440,7 @@ def make_config_tasks(zip_dir, sg_config_path, sg_url, should_redact):
                                  log_file="sync_gateway.log",
                                  content_postprocessors=server_config_postprocessors)
     collect_config_tasks.append(runtime_config_task)
-
+# Get persisted dbconfigs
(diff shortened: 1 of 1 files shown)


JRascagneres (dismissed):
Looks fine to me. 
Will leave to @adamcfraser to review too.

JRascagneres:
Actually can we verify what the redaction behaviour is? Wondering if we need to call the ?redacted endpoints when sgcollect is in redacted mode. You say you tested but no mention of results.

I think redacted is default but what if redacted is turned off?

adamcfraser (changes requested):

adamcfraser (approved):

PR #5212, opened 2021-08-27
CBG-1660: Strip out credentials obtained through GET db config

JRascagneres:
CBG-1660

Strip out credentials from db config during the GET db config operation. 
Tested with a unit test.


sarathkumarsivan (commented):
Looks good to me. Left a minor comment on using assert.Empty instead of assert.Equal.

adamcfraser (approved):

PR #5216, opened 2021-09-01
CBG-1649: Non-existent db should return 401 / 403

JRascagneres:
CBG-1649

Handles the case where a db that is requested as part of a endpoint doesn't exist.
If admin auth is enabled rather than returning a 404, it'll do an authorization check and return 401 if no user or 403 if user is authorized but no database is found.
This matches the behaviour of the flow when a database does exist but a user doesn't have access to it.

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1082/

### DIFF
--- a/rest/handler.go
+++ b/rest/handler.go
@@ -183,10 +183,30 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
 		h.setHeader("Server", base.ProductNameString)
 	}
 
+	// If an Admin Request and admin auth enabled or a metrics request with metrics auth enabled we need to check the
+	// user credentials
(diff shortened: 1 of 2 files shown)


rest/handler.go
```diff
@@ -437,6 +453,33 @@ func (h *handler) checkAuth(context *db.DatabaseContext) (err error) {
 	return nil
 }
 
+// checkAdminAuthorizationOnly simply checks whether a username / password combination is authorized pulling the
+// credentials from the handler
+func (h *handler) checkAdminAuthorizationOnly() (bool, error) {
```
JRascagneres: Rename this. Its actually authentication not authorization

adamcfraser (commented):
Everything looks good, aside from the rename from authorization to authentication that you've already identified.

JRascagneres:
> Everything looks good, aside from the rename from authorization to authentication that you've already identified.

Thanks will get that done. Just was waiting on doing it unless you had any other comments.

adamcfraser (approved):

PR #5218, opened 2021-09-01
CBG-1663: Create new temporary connection for Bootstrap ops

JRascagneres:
CBG-1663

- Modify `CouchbaseCluster` to store server URL and clusterOptions rather than a cluster connection.
- For each operation start up a connection, perform the operation, and close the connection
- Additional fix to close connection in `GetCouchbaseCollection` on error 

Has a test which ensures that we continue to get the error on a non-existent db. Existing bootstrap tests should ensure this change doesn't affect the other operations.

## [Integration Tests](http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/build?delay=0sec)
- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1070/
Couple unrelated failures

PR #5221, opened 2021-09-02
CBG-1670: Improve error message for PUT db failure

JRascagneres:
CBG-1670

 - Change getBucket in bootstrap to use the fail fast retry - meaning an auth error will immediately return rather than doing the 10 second wait.
 - If the error returned from creating bucket connection is auth error return `ErrAuthError`
 - Check for this error when `InsertConfig` is run in `handleCreateDB` and if we get this return 403

Returned error looks like:
```
{
    "error": "Forbidden",
    "reason": "auth failure accessing provided bucket dbx"
}
```

Tested by modifying existing test to check for `403` instead of `500` and also assert on the error message itself.

## Dependencies
- [x] Merge https://github.com/couchbase/sync_gateway/pull/5218

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1141/

### DIFF
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -227,11 +227,16 @@ func (cc *CouchbaseCluster) getBucket(bucketName string) (b *gocb.Bucket, teardo
 	b = connection.Bucket(bucketName)
 	err = b.WaitUntilReady(time.Second*10, &gocb.WaitUntilReadyOptions{
 		DesiredState:  gocb.ClusterStateOnline,
-		RetryStrategy: gocb.NewBestEffortRetryStrategy(nil),
+		RetryStrategy: &goCBv2FailFastRetryStrategy{},
 		ServiceTypes:  []gocb.ServiceType{gocb.ServiceTypeKeyValue},
 	})
 	if err != nil {
 		_ = connection.Close(&gocb.ClusterCloseOptions{})
+
+		if errors.Is(err, gocb.ErrAuthenticationFailure) {
+			return nil, nil, ErrAuthError
+		}
+
 		return nil, nil, err
 	}
 
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -63,7 +63,7 @@ func (h *handler) handleCreateDB() error {
 		persistedConfig.cas, err = h.server.bootstrapContext.connection.InsertConfig(bucket, h.server.config.Bootstrap.ConfigGroupID, persistedConfig)
 		if err != nil {
 			if errors.Cause(err) == base.ErrAuthError {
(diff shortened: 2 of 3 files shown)


bbrks (dismissed):
lgtm

adamcfraser:
@JRascagneres Requires rebase.

rest/admin_api.go
```diff
@@ -62,6 +62,9 @@ func (h *handler) handleCreateDB() error {
 
 		persistedConfig.cas, err = h.server.bootstrapContext.connection.InsertConfig(bucket, h.server.config.Bootstrap.ConfigGroupID, persistedConfig)
 		if err != nil {
+			if errors.Cause(err) == base.ErrAuthError {
+				return base.HTTPErrorf(http.StatusForbidden, "auth failure accessing provided bucket %s", bucket)
```
adamcfraser: Should we indicate here that the access failure occurs using the bootstrap credentials (as opposed to the credentials used to auth the admin API REST call)?  

I'm also curious whether 'auth' was left intentionally vague - if this is authorization (and not authentication), I think using authorization failure is a better choice.

base/bootstrap.go
```diff
@@ -227,11 +227,16 @@ func (cc *CouchbaseCluster) getBucket(bucketName string) (b *gocb.Bucket, teardo
...
 	err = b.WaitUntilReady(time.Second*10, &gocb.WaitUntilReadyOptions{
 		DesiredState:  gocb.ClusterStateOnline,
-		RetryStrategy: gocb.NewBestEffortRetryStrategy(nil),
+		RetryStrategy: &goCBv2FailFastRetryStrategy{},
 		ServiceTypes:  []gocb.ServiceType{gocb.ServiceTypeKeyValue},
 	})
 	if err != nil {
 		_ = connection.Close(&gocb.ClusterCloseOptions{})
+
+		if errors.Is(err, gocb.ErrAuthenticationFailure) {
```
adamcfraser, line 236: Does gocb differentiate between authentication and authorization failure, or do we get ErrAuthenticationFailure in both cases?

adamcfraser (changes requested):

base/bootstrap.go
JRascagneres: I don't think it does. Tried with with a bootstrap details that could only access a couple of buckets and got this failure.

rest/admin_api.go
JRascagneres: Can add info about bootstrap creds.
Yes auth was left vague. As with the above comment, we don't really know which it is.

adamcfraser (approved):

PR #5222, opened 2021-09-03
Remove attachment version from REST API response

sarathkumarsivan:
From Lithium on, each attachment reference should include an additional version in the document metadata to indicate that the attachment reference is made through the new reference scheme — one-to-one mapping between documents and attachments. The version property has a value of “2” and it remains internal to the system; it is not necessary to expose the version over the REST API.


bbrks (approved):
LGTM

PR #5225, opened 2021-09-03
Add consistent 10sec/best effort options to all bootstrap ops

bbrks:
Consistently apply timeouts/retry to all bootstrap connection ops done for config.

If no timeout or retry strategy are chosen, gocb v2 uses best effort for 75 seconds, which exceeds a lot of our 30 second test timeouts.

### DIFF
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -78,7 +78,10 @@ func (cc *CouchbaseCluster) GetConfigBuckets() ([]string, error) {
 		return nil, errors.New("nil CouchbaseCluster")
 	}
 
-	buckets, err := cc.c.Buckets().GetAllBuckets(nil)
+	buckets, err := cc.c.Buckets().GetAllBuckets(&gocb.GetAllBucketsOptions{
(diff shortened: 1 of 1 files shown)


base/bootstrap.go
```diff
@@ -78,7 +78,10 @@ func (cc *CouchbaseCluster) GetConfigBuckets() ([]string, error) {
 		return nil, errors.New("nil CouchbaseCluster")
 	}
 
-	buckets, err := cc.c.Buckets().GetAllBuckets(nil)
+	buckets, err := cc.c.Buckets().GetAllBuckets(&gocb.GetAllBucketsOptions{
+		Timeout:       time.Second * 10,
```
bbrks, line 82: Make a cluster-wide timeout instead of per-op.

bbrks: Doing this depends on the changes in #5218 - closing PR and opening Jira ticket instead.

https://issues.couchbase.com/browse/CBG-1679

PR #5228, opened 2021-09-03
CBG-1666: Fix all_docs with limit when using GSI

JRascagneres:
CBG-1666

Calling `results.Close()` on GSI results in gocb v2 requires the results to be exhausted. In the event where we have a limit we are stopping when we hit this limit and not exhausting the rows, resulting in the `the result must be closed before accessing the meta-data` error. 

Fixed by just adding a for loop which iterates through any remaining results and just no-opping on each. This operation is performed when results.Close() is called.

This is already tested by `TestAllDocsAccessControl`, however we only hit this error in the GSI case which is not currently part of our standard integration tests, due to underperforming box.

Additionally made a fix to that existing test just to require on the length. Previously this test would panic due to this problem. 

**Considerations**
1. Considered making all_docs take limit, however, filtering needs to be done locally on SGW for channel access so this isn't possible
2. Considered adding pagination work to again, avoid ending up with un-used results, however, this is involved as it requires using start / end sequences to avoid duplicate results.

In the end the approach in this PR is preferable to the above. Although 2 may work well the work to refactor allDocs to handle this is significant and is not worth it given that this endpoint is being deprecated.

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1121/

GSI:
- [ ] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1136/

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -834,7 +834,6 @@ func (db *Database) ForEachDocID(callback ForEachDocIDFunc, resultsOpts ForEachD
 // Iterate over the results of an AllDocs query, performing ForEachDocID handling for each row
 func (db *Database) processForEachDocIDResults(callback ForEachDocIDFunc, limit uint64, results sgbucket.QueryResultIterator) error {
 
-	endOfResultsReached := false
 	count := uint64(0)
 	for {
 		var queryRow AllDocsIndexQueryRow
@@ -867,7 +866,6 @@ func (db *Database) processForEachDocIDResults(callback ForEachDocIDFunc, limit 
 			}
 		}
 		if !found {
-			endOfResultsReached = true
 			break
 		}
 
@@ -883,15 +881,6 @@ func (db *Database) processForEachDocIDResults(callback ForEachDocIDFunc, limit 
 		}
 
 	}
-
-	// Have to iterate over any remaining results to clear the reader
-	// Otherwise we get "the result must be closed before accessing the meta-data" on close details on CBG-1666
-	if !endOfResultsReached && !db.UseViews() {
-		for results.Next(nil) {
-			// noop to drain results
-		}
-	}
-
 	return nil
 }
 
--- a/db/database.go
+++ b/db/database.go
@@ -834,6 +834,7 @@ func (db *Database) ForEachDocID(callback ForEachDocIDFunc, resultsOpts ForEachD
 // Iterate over the results of an AllDocs query, performing ForEachDocID handling for each row
 func (db *Database) processForEachDocIDResults(callback ForEachDocIDFunc, limit uint64, results sgbucket.QueryResultIterator) error {
 
+	endOfResultsReached := false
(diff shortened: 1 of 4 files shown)


db/database.go
```diff
@@ -881,6 +883,18 @@ func (db *Database) processForEachDocIDResults(callback ForEachDocIDFunc, limit
...
+
+	// Have to iterate over any remaining results to clear the reader
+	// Otherwise we get "the result must be closed before accessing the meta-data" on close details on CBG-1666
+	if !endOfResultsReached && !db.UseViews() {
+		for {
+			found := results.Next(nil)
+			if !found {
+				break
+			}
+		}
```
bbrks: Not sure how much the compiler will optimise away in your case, but this still feels cleaner even if it makes no performance difference.

```suggestion
		for r.Next(nil) {
			// noop to drain results
		}
```

db/database.go
```diff
@@ -834,6 +834,7 @@ func (db *Database) ForEachDocID(callback ForEachDocIDFunc, resultsOpts ForEachD
 // Iterate over the results of an AllDocs query, performing ForEachDocID handling for each row
 func (db *Database) processForEachDocIDResults(callback ForEachDocIDFunc, limit uint64, results sgbucket.QueryResultIterator) error {
```
bbrks: What do you think about making the `gocbRawIterator.Close()` implementation drain any remaining results prior to closing?

That covers use-cases outside of this `all_docs` one, and transparently does what the SDK needs to do in order to close a result iterator to keep the API consistent with the others (no special cases for GSI and gocbv1/v2 needed).

bbrks (changes requested):

db/database.go
JRascagneres: Agree this would probably be best. However at the moment we can't do this.
See this which has a description of the issue: https://issues.couchbase.com/browse/GOCBC-1169

JRascagneres:
Postponing based on GOCBC-1169

db/database.go
adamcfraser: @JRascagneres What's our use case that 'has multiple places where we call NextBytes()', that you mention on that ticket?

JRascagneres: Well just anywhere we call `Next()` on a `sgbucket.QueryResultIterator` we ultimately call `NextBytes()`.

 The issue we were hitting with all docs is that we weren't exhausting all of the results before calling close. This PR added that fix to exhaust the results on an all_docs request, however, as Ben pointed out this would be better done as part of our `Close()` wrapper to ensure this works for all uses and future uses of iterating on a query set.

That part was me trying to explain that we iterate over results in multiple places so I might have to end up doing this check to exhaust results in more than one place to be safe so it would be better to have this fix so I can implement the check on `Close()`

adamcfraser (approved):

PR #5232, opened 2021-09-06
CBG-1686: Cleanup perms and Read Only fix

JRascagneres:
CBG-1686

Removed some unnecessary response permissions as these were left from changed perms
 - Note this doesn't affect functionality anything but might as well clean up

Fixed a Read / Write issue. We explicitly checked for PUT / POST previously, however, this didn't include DELETE. Flipped this around to check for GET / HEAD too as this seems safer.

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1081/

### DIFF
--- a/rest/handler.go
+++ b/rest/handler.go
@@ -477,10 +477,10 @@ func checkAdminAuth(bucketName, basicAuthUsername, basicAuthPassword string, att
 	if bucketName != "" {
 		requestRoles = BucketScopedEndpointRoles
 	} else {
-		if attemptedHTTPOperation == http.MethodPut || attemptedHTTPOperation == http.MethodPost {
-			requestRoles = ClusterScopedEndpointRolesWrite
-		} else {
+		if attemptedHTTPOperation == http.MethodGet || attemptedHTTPOperation == http.MethodHead {
 			requestRoles = ClusterScopedEndpointRolesRead
+		} else {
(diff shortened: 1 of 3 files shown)


rest/handler.go
```diff
@@ -477,10 +477,10 @@ func checkAdminAuth(bucketName, basicAuthUsername, basicAuthPassword string, att
 	if bucketName != "" {
 		requestRoles = BucketScopedEndpointRoles
 	} else {
-		if attemptedHTTPOperation == http.MethodPut || attemptedHTTPOperation == http.MethodPost {
-			requestRoles = ClusterScopedEndpointRolesWrite
-		} else {
+		if attemptedHTTPOperation == http.MethodGet || attemptedHTTPOperation == http.MethodHead {
```
bbrks: Might want `http.MethodOptions` to be considered a read-only op. It's used (or allowed) in a couple of places in SG, but not explicitly in the routing handler list.

bbrks (changes requested):
Looks good, one suggestion for the `OPTIONS` method.

rest/handler.go
JRascagneres: Added

bbrks (approved):

adamcfraser (approved):

PR #5236, opened 2021-09-08
CBG-1624: Allow per-database credential overrides from startup config

bbrks:
CBG-1624

Allows per-database credentials to be set that override the bootstrap credentials.

### Config example:
```
{
  "bootstrap": {
    "username": "bob",
    "password": "secret"
  },
  "database_credentials": {
    "db1": {
      "username": "alice",
      "password": "hunter2"
    },
    "db2": {
      "x509_cert_path": "/tmp/foo",
      "x509_key_path": "/tmp/bar"
    }
  }
}
```

In this case, bootstrap to retrieve dbconfigs happens as bob, but then whenever we create or open a database matching the name in the config, we will use those credentials instead of the bootstrap ones.

## REST API Changes
- `GET /_config` now returns the `database_credentials` section of the config to reflect the config file on disk.
- `GET /_config?include_runtime=true` now shows per-database credentials inside each running database config to reflect the running state of each database.
- `GET /{db}/_config?include_runtime=false` has no change (the dbconfig stored in the bucket remains unchanged with per-db overrides)
- `GET /{db}/_config?include_runtime=true` now shows per-database credentials when set
- Redaction is applied to passwords on all of the above scenarios.

- [ ] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1094/

### DIFF
--- a/rest/config_startup.go
+++ b/rest/config_startup.go
@@ -67,6 +67,8 @@ type StartupConfig struct {
 	Replicator  ReplicatorConfig   `json:"replicator,omitempty"`
 	Unsupported UnsupportedConfig  `json:"unsupported,omitempty"`
 
+	DatabaseCredentials map[string]*DatabaseCredentialsConfig `json:"database_credentials,omitempty" help:"A map of database name to credentials, that can be used instead of the bootstrap ones."`
+
 	MaxFileDescriptors uint64 `json:"max_file_descriptors,omitempty" help:"Max # of open file descriptors (RLIMIT_NOFILE)"`
 
 	DeprecatedConfig *DeprecatedConfig `json:"-,omitempty" help:"Deprecated options that can be set from a legacy config upgrade, but cannot be set from a 3.0 config."`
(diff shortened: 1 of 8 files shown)


rest/config_startup.go
```diff
@@ -143,6 +145,15 @@ type HTTP2Config struct {
 	Enabled *bool `json:"enabled,omitempty" help:"Whether HTTP2 support is enabled"`
 }
 
+type PerDatabaseCredentialsConfig map[string]*DatabaseCredentialsConfig
+
+type DatabaseCredentialsConfig struct {
+	Username     string `json:"username,omitempty"       help:"Username for authenticating to the bucket"`
+	Password     string `json:"password,omitempty"       help:"Password for authenticating to the bucket"`
+	X509CertPath string `json:"x509_cert_path,omitempty" help:"Cert path (public key) for X.509 bucket auth"`
+	X509KeyPath  string `json:"x509_key_path,omitempty"  help:"Key path (private key) for X.509 bucket auth"`
```
adamcfraser, line 154: Just for clarification, on the omission of cacertpath in this config:
1. Are database configs always going to inherit cacertpath from the bootstrap config?
2. Does that apply even if some dbs are using x509 creds, and others are using basic auth?  (I think that's required for TLS support, just verifying)

adamcfraser (changes requested):

rest/config_startup.go
bbrks: TLDR: Yes

1. The Couchbase Server root certificate is cluster-scoped, so even when two databases have separate X.509 certs, they would still have to be signed by the same root cert associated with the bootstrap cluster.
2. If using mixed X.509/basic auth, same applies - X.509 certs will have to be signed by the root which is also used for TLS connection.

adamcfraser (approved):

PR #5239, opened 2021-09-10
CBG-1619: Wait for successful db startup before persisting config on db create

bbrks:
CBG-1619

Ensures that the supplied per-database credentials in the startup config are valid by loading the database first, before persisting to the cluster.
- Unloads the database on a failed write to prevent cluster inconsistencies.
- Promoted locking up a level in handleCreateDB to avoid races between background config loop and a failed database creation/removal.

We don't need to validate per-database credentials on update because we know the database has already started up successfully, and must be already using the per-database credentials, if they're set.

## Dependencies
- [x] #5236 

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1254/

### DIFF
--- a/base/bucket.go
+++ b/base/bucket.go
@@ -369,10 +369,6 @@ func GetBucket(spec BucketSpec) (bucket Bucket, err error) {
 		}
 
 		if err != nil {
-			if pkgerrors.Cause(err) == ErrAuthError {
-				Warnf("Unable to authenticate as user %q: %v", UD(username), err)
-				return nil, ErrFatalBucketConnection
-			}
 			return nil, err
 		}
 
--- a/base/collection.go
+++ b/base/collection.go
@@ -96,9 +96,15 @@ func GetCollectionFromCluster(cluster *gocb.Cluster, spec BucketSpec, waitUntilR
 
 	// Connect to bucket
 	bucket := cluster.Bucket(spec.BucketName)
-	err := bucket.WaitUntilReady(time.Duration(waitUntilReadySeconds)*time.Second, nil)
+	err := bucket.WaitUntilReady(time.Duration(waitUntilReadySeconds)*time.Second, &gocb.WaitUntilReadyOptions{
+		DesiredState:  gocb.ClusterStateOnline,
+		RetryStrategy: &goCBv2FailFastRetryStrategy{},
+	})
 	if err != nil {
 		_ = cluster.Close(&gocb.ClusterCloseOptions{})
+		if errors.Is(err, gocb.ErrAuthenticationFailure) {
+			return nil, ErrAuthError
+		}
 		Warnf("Error waiting for bucket to be ready: %v", err)
 		return nil, err
 	}
--- a/base/collection.go
+++ b/base/collection.go
@@ -97,7 +97,6 @@ func GetCollectionFromCluster(cluster *gocb.Cluster, spec BucketSpec, waitUntilR
 	// Connect to bucket
 	bucket := cluster.Bucket(spec.BucketName)
 	err := bucket.WaitUntilReady(time.Duration(waitUntilReadySeconds)*time.Second, &gocb.WaitUntilReadyOptions{
-		DesiredState:  gocb.ClusterStateOnline,
 		RetryStrategy: &goCBv2FailFastRetryStrategy{},
 	})
 	if err != nil {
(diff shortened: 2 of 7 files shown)


adamcfraser (dismissed):

adamcfraser:
Looks good, pending merge of CBG-1624

bbrks:
I'm running into complexities around unwanted retry handling after these operations are swapped around. Still working through those before this PR is ready for review.

base/bucket.go
```diff
@@ -369,10 +369,6 @@ func GetBucket(spec BucketSpec) (bucket Bucket, err error) {
 		}
 
 		if err != nil {
-			if pkgerrors.Cause(err) == ErrAuthError {
-				Warnf("Unable to authenticate as user %q: %v", UD(username), err)
-				return nil, ErrFatalBucketConnection
-			}
```
bbrks, line 375: This `ErrAuthError`->`ErrFatalBucketConnection` handling was effectively pushed up a level to `ConnectToBucket`, to separately start treating auth errors as abortable in the retry loop. This means we can preserve the original error to use instead of losing all information about it.

rest/config.go
```diff
@@ -1051,14 +1051,14 @@ func setupServerContext(config *StartupConfig, persistentConfig bool) (*ServerCo
 // fetchAndLoadConfigs retrieves all database configs from the ServerContext's bootstrapConnection, and loads them into the ServerContext.
 // It will remove any databases currently running that are not found in the bucket.
 func (sc *ServerContext) fetchAndLoadConfigs() (count int, err error) {
+	sc.lock.Lock()
+	defer sc.lock.Unlock()
+
```
bbrks, line 1056: Lock moved up to avoid the race condition where SG adds a running database before persisting, and the resulting background config poller sees the database was absent from the cluster before it's had a chance to be added, meaning the database would be removed.

This race was triggered reliably inside `TestConfigsIncludeDefaults` prior to the fix.

bbrks (commented):
A couple of inline commentary-style comments describing changes.

base/collection.go
```diff
@@ -91,9 +91,15 @@ func GetCouchbaseCollection(spec BucketSpec) (*Collection, error) {
 	// Connect to bucket
 	bucket := cluster.Bucket(spec.BucketName)
 	// TODO: identify required services and add to WaitUntilReadyOptions
-	err = bucket.WaitUntilReady(30*time.Second, nil)
+	err = bucket.WaitUntilReady(30*time.Second, &gocb.WaitUntilReadyOptions{
+		DesiredState:  gocb.ClusterStateOnline,
```
adamcfraser: I think this is just a copy-paste from other places that we're using goCBv2FailFastRetryStrategy, but I don't think we need to specify the DesiredState here - the default is ClusterStateOnline.

rest/admin_api.go
```diff
@@ -58,25 +59,43 @@ func (h *handler) handleCreateDB() error {
...
+				return base.HTTPErrorf(http.StatusForbidden, "auth failure accessing provided bucket: %s", bucket)
+			}
 			return base.HTTPErrorf(http.StatusInternalServerError, "couldn't load database: %v", err)
 		}
+
+		// now we've started the db successfully, we can persist it to the cluster
+		cas, err := h.server.bootstrapContext.connection.InsertConfig(bucket, h.server.config.Bootstrap.ConfigGroupID, persistedConfig)
+		if err != nil {
+			// unload the database to prevent the cluster being in an inconsistent state
+			h.server._removeDatabase(dbName)
```
adamcfraser, line 91: What happens in the case where the database already existed?  It looks like it would be removed from this node but continue to be running on other nodes, until the next time the config was updated.  Should we attempt to reload the database locally using the previously persisted config in that case?

adamcfraser (changes requested):
Posted a question about the handling when we fail to persist a database that already existed.

rest/admin_api.go
bbrks: I could start an async load to avoid the potential wait time until next load, but I still think it makes sense to unload the user's database, and return the conflict/already exists error back to the user who initiated the request, given the "winning" persisted config is not theirs.

bbrks: I've pushed up a commit to address this case if you want to take a look at the proposed solution.

adamcfraser (approved):

PR #5241, opened 2021-09-13
CBG-1427 - Full Body replication is sent when delta source is tombstone

IsaacLambat:
CBG-1427

- Added new error `ErrFromRevIsTombstone`
- When getting the delta, if the from revision is marked as deleted then this error is returned. Then sendRevAsDelta falls back to full body replication so then a successful resurrection is done.

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1113

### DIFF
--- a/base/error.go
+++ b/base/error.go
@@ -54,9 +54,9 @@ var (
 	// ErrEmptyDocument is returned when trying to insert a document with a null body.
 	ErrEmptyDocument = &sgError{"Document body is empty"}
 
-	// ErrFromRevIsTombstone is returned to indicate delta sync should do a full body replication due to the
+	// ErrDeltaSourceIsTombstone is returned to indicate delta sync should do a full body replication due to the
 	// delta source being a tombstone (therefore having an empty body)
-	ErrFromRevIsTombstone = &sgError{"From rev is a tombstone"}
+	ErrDeltaSourceIsTombstone = &sgError{"From rev is a tombstone"}
 )
 
 func (e *sgError) Error() string {
--- a/base/error.go
+++ b/base/error.go
@@ -53,6 +53,10 @@ var (
 
 	// ErrEmptyDocument is returned when trying to insert a document with a null body.
 	ErrEmptyDocument = &sgError{"Document body is empty"}
+
+	// ErrFromRevIsTombstone is returned to indicate delta sync should do a full body replication due to the
+	// delta source being a tombstone (therefore having an empty body)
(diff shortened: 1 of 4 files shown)


base/error.go
```diff
@@ -53,6 +53,10 @@ var (
 
 	// ErrEmptyDocument is returned when trying to insert a document with a null body.
 	ErrEmptyDocument = &sgError{"Document body is empty"}
+
+	// ErrFromRevIsTombstone is returned to indicate delta sync should do a full body replication due to the
+	// delta source being a tombstone (therefore having an empty body)
+	ErrFromRevIsTombstone = &sgError{"From rev is a tombstone"}
```
JRascagneres: I'd be more clear that this is delta sync.
`ErrDeltaSourceIsTombstone` would be better imo.

db/blip_handler.go
```diff
@@ -683,6 +683,9 @@ func (bsc *BlipSyncContext) sendRevAsDelta(sender *blip.Sender, docID, revID, de
 		// Something went wrong in the diffing library. We want to know about this!
 		base.WarnfCtx(bsc.loggingCtx, "Falling back to full body replication. Error generating delta from %s to %s for key %s - err: %v", deltaSrcRevID, revID, base.UD(docID), err)
 		return bsc.sendRevision(sender, docID, revID, seq, knownRevs, maxHistory, handleChangesResponseDb)
+	} else if err == base.ErrFromRevIsTombstone {
+		base.TracefCtx(bsc.loggingCtx, base.KeySync, "Falling back to full body replication. Delta source %s is tombstone and delta %s is resurrection for key %s", deltaSrcRevID, revID, base.UD(docID))
```
JRascagneres:
```suggestion
		base.TracefCtx(bsc.loggingCtx, base.KeySync, "Falling back to full body replication.  Delta source %s is tombstone. Unable to generate delta to %s for key %s", deltaSrcRevID, revID, base.UD(docID))
```

I think that reads better. What do you think?

JRascagneres (changes requested):
Couple minor comments

db/blip_handler.go
IsaacLambat: Much better

JRascagneres (approved):
This looks good to me. Just assigning to @adamcfraser for a final look.

adamcfraser (approved):

PR #5244, opened 2021-09-15
CBG-1698 - Fixed defaultLogFilePath not being set by flag when using persistent config

IsaacLambat:
CBG-1698

The problem was the flag was being dereferenced into defaultLogFilePath before the flag had been parsed. The same fix has been applied to CBG-1638 as that had the same issue (albeit in a different function).

- `defaultLogFilePath` now gets set after the flags have been parsed instead of before


JRascagneres (changes requested):
Gave it a test and seems to work well. I would avoid making the same change to multiple PRs though. I'd remove this change from the other PR and add this PR being merged as a dependency. 

Also do we still need to process this flag again in `ParseCommandLine`?

adamcfraser (approved):

PR #5245, opened 2021-09-15
CBG-1595: Integration tests do not work with a couchbases:// scheme

IsaacLambat:
CBG-1595

- All tests now work with the couchbases:// protocol with no TLS verification.
- New environment variable SG_TEST_TLS_SKIP_VERIFY

## TODO:
- [x] Add `SG_TEST_TLS_SKIP_VERIFY` to Jenkins (perhaps automatically when using couchbases:// ?)
- [x] Run new integration tests with couchbases:// and couchbase://

- [x] `couchbase` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1223/
- [x] `couchbases` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1224/

### DIFF
--- a/rest/admin_api_auth_test.go
+++ b/rest/admin_api_auth_test.go
@@ -14,9 +14,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func MakeUser(t *testing.T, serverURL, username, password string, roles []string) {
-	httpClient := http.DefaultClient
-
+func MakeUser(t *testing.T, httpClient *http.Client, serverURL, username, password string, roles []string) {
 	form := url.Values{}
 	form.Add("password", password)
 	form.Add("roles", strings.Join(roles, ","))
@@ -39,13 +37,12 @@ func MakeUser(t *testing.T, serverURL, username, password string, roles []string
 	require.Equal(t, http.StatusOK, resp.StatusCode)
 }
 
-func DeleteUser(t *testing.T, serverURL, username string) {
+func DeleteUser(t *testing.T, httpClient *http.Client, serverURL, username string) {
 	req, err := http.NewRequest("DELETE", fmt.Sprintf("%s/settings/rbac/users/local/%s", serverURL, username), nil)
 	require.NoError(t, err)
 
 	req.SetBasicAuth(base.TestClusterUsername(), base.TestClusterPassword())
 
-	httpClient := http.DefaultClient
 	resp, err := httpClient.Do(req)
 	require.NoError(t, err)
 
@@ -156,8 +153,8 @@ func TestCheckPermissions(t *testing.T) {
 	for _, testCase := range testCases {
 		t.Run(testCase.Name, func(t *testing.T) {
 			if testCase.CreateUser != "" {
-				MakeUser(t, eps[0], testCase.CreateUser, testCase.CreatePassword, testCase.CreateRoles)
-				defer DeleteUser(t, eps[0], testCase.CreateUser)
+				MakeUser(t, httpClient, eps[0], testCase.CreateUser, testCase.CreatePassword, testCase.CreateRoles)
+				defer DeleteUser(t, httpClient, eps[0], testCase.CreateUser)
 			}
 
 			statusCode, permResults, err := CheckPermissions(httpClient, eps, "", testCase.Username, testCase.Password, testCase.RequestPermissions, testCase.ResponsePermissions)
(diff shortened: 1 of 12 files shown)


rest/admin_api_auth_test.go
```diff
@@ -15,7 +16,14 @@ import (
 )
 
 func MakeUser(t *testing.T, serverURL, username, password string, roles []string) {
-	httpClient := http.DefaultClient
+	// TODO: Make work with management HTTP client on Jenkins
+	httpClient := &http.Client{
```
JRascagneres: Just to humor me can you try to use the management HTTP client locally and breakpoint it. See if it has the InsecureSkipVerify set?

rest/api_test.go
```diff
@@ -3121,6 +3121,8 @@ func TestAddingAttachment(t *testing.T) {
 	rt := NewRestTester(t, nil)
 	defer rt.Close()
 	defer func() { walrus.MaxDocSize = 0 }()
+	// Generate RT config for skip TLS when running with couchbases:// protocol
+	rt.Bucket()
```
JRascagneres: Why is this now required? This should get called when a bucket is requested by the test? This is implicitly done in this case when createDoc is called. Which triggers an admin request.

rest/api_test.go
```diff
@@ -3179,6 +3181,8 @@ func TestAddingLargeDoc(t *testing.T) {
 	rt := NewRestTester(t, nil)
 	defer rt.Close()
 	defer func() { walrus.MaxDocSize = 0 }()
+	// Generate RT config to skip TLS config (for testing iwth couchbases://)
+	rt.Bucket()
```
JRascagneres: Same as above

rest/bootstrap_test.go
```diff
@@ -139,6 +137,8 @@ func bootstrapStartupConfigForTest(t *testing.T) StartupConfig {
 	config.Bootstrap.Server = base.UnitTestUrl()
 	config.Bootstrap.Username = base.TestClusterUsername()
 	config.Bootstrap.Password = base.TestClusterPassword()
+	config.Bootstrap.ServerTLSSkipVerify = base.BoolPtr(true)
```
JRascagneres: Could probably do a similar to thing as below. I think. Avoids setting TLS Skip Verify if we're not using TLS which is probably preferable to always setting it

rest/changes_api_test.go
```diff
@@ -194,6 +194,8 @@ func TestPostChanges(t *testing.T) {
 
 	rt := NewRestTester(t, &RestTesterConfig{SyncFn: `function(doc) {channel(doc.channel);}`})
 	defer rt.Close()
+	// Generate RT config for skip TLS when running with couchbases:// protocol
+	rt.Bucket()
```
JRascagneres: Earlier comment

rest/config_test.go
```diff
@@ -1198,6 +1201,7 @@ func TestSetupServerContext(t *testing.T) {
 		config := DefaultStartupConfig("")
 		config.Bootstrap.Server = base.UnitTestUrl() // Valid config requires server to be explicitly defined
 		config.Bootstrap.UseTLSServer = base.BoolPtr(base.ServerIsTLS(base.UnitTestUrl()))
+		config.Bootstrap.ServerTLSSkipVerify = base.BoolPtr(true)
```
JRascagneres: Again, can maybe use same logic as above to we're only setting when using TLS.

rest/server_context_test.go
```diff
@@ -112,7 +112,7 @@ func TestAllDatabaseNames(t *testing.T) {
 	defer tb2.Close()
 
 	serverConfig := &StartupConfig{
-		Bootstrap: BootstrapConfig{UseTLSServer: base.BoolPtr(false)},
+		Bootstrap: BootstrapConfig{UseTLSServer: base.BoolPtr(base.ServerIsTLS(base.UnitTestUrl())), ServerTLSSkipVerify: base.BoolPtr(true)},
```
JRascagneres: Same as earlier, only set true if using TLS

JRascagneres (dismissed):
Generally fine although just had a thought after writing comments...

Thinking if we should have a new test environment variable for this stuff... Basically so we can choose to turn on / off skip verify. Thoughts are we could have a bug where we're not properly validating the ca cert and we want to catch this in our testing without everything just skipping verification. I know we don't regularly run x509 / TLS tests at the moment but it'd be something to think about. What do you think?

rest/admin_api_auth_test.go
IsaacLambat: Had a look at this. It does not set `InsecureSkipVerify` and it does fail locally as well when this is not set. 

Looks like gocbcore would need to add this on their end on creation as you can't set it on an existing HTTP client.

IsaacLambat: Fixed through commit b43cb76

bbrks:
Let's have another two full test runs for `couchbase` and `couchbases`

bbrks (approved):

PR #5246, opened 2021-09-15
Fix TestHasAttachmentsFlag

JRascagneres:
PR fixes TestHasAttachmentsFlag. https://github.com/couchbase/sync_gateway/pull/5222 removed the version field from the `Get1xBody` response, however, we were still asserting on it.
Switched it to get the data from alternate source so that we can still verify the version check.

Shouldn't require a full integration run as its fairly minimal change - however I have verified that this works on Couchbase Server locally.

### DIFF
--- a/db/crud_test.go
+++ b/db/crud_test.go
@@ -131,11 +131,16 @@ func TestHasAttachmentsFlag(t *testing.T) {
 
 	// Retrieve the document:
 	log.Printf("Retrieve doc 2-a...")
-	gotbody, err := db.Get1xBody("doc1")
-	assert.NoError(t, err, "Couldn't get document")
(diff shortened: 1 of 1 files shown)


adamcfraser (dismissed):

JRascagneres:
Note: Review was dismissed when I pushed up new version.
New version has no code changes. Reason was basically the bootstrap step failed to cache the boostrapped files properly so subsequent checks failed. However, it did create the cache - just empty, meaning that re-running the CI just resulted in re-using that failed cache. 
Had to push up new version to force the cache change as the hash is part of the cache key.

adamcfraser (approved):

PR #5249, opened 2021-09-16
CBG-1672 - Return 422 status for unprocessible deltas instead of 404 to use non-delta retry handling

IsaacLambat:
CBG-1672

- HTTP 422 is returned in handleRev when delta sync errors occur
- A test validates that 422 is returned by delta sync in an error case (manually verified by logs)

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1132


bbrks:
As discussed, let's handle the "404" status case still for older clients in `sendRevisionWithProperties` and let's add a debug log when we're falling back to use the full body replication.

bbrks (approved):

PR #5252, opened 2021-09-20
CBG-1627/CBG-1647: Support conflicts for attachment deletion

JRascagneres:
CBG-1627/CBG-1647

- Modifies the obsolete attachment deletion work to ensure that deletions do not occur incorrectly when there are conflicting revisions
- Adds `getAttachmentIDsForLeafRevisions` function as a helper

Functionality is tested with an additional unit test along with any existing tests that ensure that this works fine with no conflicts.

- [ ] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1157/
- [ ] `xattrs=false` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1158/

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -1743,7 +1743,7 @@ func (db *Database) updateAndReturnDoc(docid string, allowImport bool, expiry ui
 	var unusedSequences []uint64                                 // Must be scoped outside callback, used over multiple iterations
 	var oldBodyJSON string                                       // Stores previous revision body for use by DocumentChangeEvent
 	var createNewRevIDSkipped bool
-	var previousAttachments, currentAttachments map[string]struct{}
+	var previousAttachments map[string]struct{}
 
 	// Update the document
 	inConflict := false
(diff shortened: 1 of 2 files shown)


db/crud.go
```diff
@@ -1989,6 +1980,51 @@ func (db *Database) updateAndReturnDoc(docid string, allowImport bool, expiry ui
...
+	}
+
+	documentLeafRevisions := doc.History.GetLeaves()
+	if len(documentLeafRevisions) > 1 {
+		for _, leafRevision := range documentLeafRevisions {
+			if leafRevision == newRevID {
+				continue
+			}
+
+			revInfo, err := doc.History.getInfo(leafRevision)
```
adamcfraser: You could use GetLeavesFiltered to only return leaves where `revInfo.HasAttachments && leafRevision != newRevID` - I think that would simplify this code a bit.

Alternately you could use revtree.forEachLeaf to combine the two operations you're doing here (GetLeaves() and getInfo()), and avoid the additional lookup in getInfo.

adamcfraser (changes requested):
Approach looks fine - left a comment on a possible simplification of getAttachmentIDsForLeafRevisions.  Have a look and see what you think.

db/crud.go
JRascagneres: Yeah thats a good plan, implementing 1 seemed to have cleaned it up well

adamcfraser (approved):

PR #5258, opened 2021-09-22
CBG-1406: Implement compaction API with generic handling

JRascagneres:
CBG-1406

- Aims to implement the new compaction work for attachment cleanup
- While doing this work we are also re-writing the compact endpoint to allow for asynchronous work for both tombstones and this newly implemented attachment feature.
- Aiming to make a generic interface to be used for this type of work for:
	- Resync
	- Tombstone Compaction
	- Attachment Compaction
	- (More possibly in the future)

// TODO: Cleanup any unnecessary prints / debugs
// TODO: Write proper tests
// TODO: Integration tests

Raising as a PR now as a draft to get a review on the current state.


db/background_status_mgr.go
```diff
@@ -0,0 +1,183 @@
...
+
+const (
+	BackgroundProcessActionStart BackgroundProcessAction = "start"
+	BackgroundProcessActionStop  BackgroundProcessAction = "stop"
+)
+
+type BackgroundManagerInterface interface {
+	ResetStatus()
+	GetJSONStatus() ([]byte, error)
+}
```
adamcfraser: With the code as it stands today, there isn't any need for this to be an interface, I don't think, because there aren't any callers that are trying to work with the interface generically, as opposed to a specific implementation.  If the goal is just to share the implementation of state management, I think all that's really needed is the BackgroundManager type, which can be embedded in the various implementations.

However, if BackgroundManager needs to be able to invoke implementation-specific functionality (like calling GetJSONStatus() from Stop(), like resyncManager used to do), then I'd probably turn this around and have callers all instantiate a BackgroundManager, and pass in an implementation of an interface (probably BackgroundProcessInterface instead of BackgroundManagerInterface).  So you'd have something like:

```
type BackgroundManager struct {
   ...existing state management...
   process BackgroundProcessInterface
}
```
Going this route I think you'd actually want BackgroundProcessInterface to include a method to actually kick off the process, like Run(), so that you're just working generically with the BackgroundManager from the REST handler. 

We'd probably also want to replace the separate implementations (like ResyncManager) on databaseContext with something like 
```
BackgroundProcesses    map[string]BackgroundManager
```
that could have items added to it at runtime.

adamcfraser (commented):
Left a comment about two possible ways you could go to share the BackgroundProcess code between resync and compact.  Either one would probably be fine, it feels like the current PR is somewhere in between the two.

db/background_status_mgr.go
JRascagneres: I'll probably have to go with the former. I kind of need each individual 'manager' to have its own functionality, like in the resync case `UpdateProcessedChanged` and I think in the compact case there is going to be more like that. Thats going to be awkward and involve casting I think if we were to go with #2.

I'll have a think though. I do want to eventually refactor it to have a `Run()` like you say but wanted to avoid growing this ticket too large to begin with. 

I suppose if we do have a map with static names for things like 'Resync' then I could do the casting I didn't like to sound of in the first paragraph there...

adamcfraser: It's expected that each 'manager' is going to have it's own functionality - that functionality would be internal to the manager implemention, though, and not need to be part of the BackgroundProcessInterface.  For something like UpdateProcessedChanged - I don't think that's needed outside of the 'run' call (which is UpdateAllDocChannels for resync).

JRascagneres: Yeah I just mean that if I have it as a generic 'BackgroundManager' on the context rather than a typed version like 'ResyncManager' accessing those specific methods might be trickier but granted if I have run and have it encapsulated in that it might work better then I never really need to access specific implementations from outside of the type.
That involved more refactoring than I was hoping to do though but I guess its probably worth spending the time to do it properly now, just trying to avoid scope creeping a little too.

JRascagneres: I'm going to actually opt for the first option. Spent quite a bit of time working with both and the second one just didn't work quite as I wanted it to and added unnecessary complexity. I think I might have just pulled `interface` out of thin air and decided to run with it whereas it doesn't really add anything.

PR #5259, opened 2021-09-24
CBG-1705: Release query ops on Close()

JRascagneres:
CBG-1705
 - Remove uses of `releaseViewOp` in the gocb v2 use cases
 - Switch to setting the view op channel on the `gocbRawIterator`
 - Above allows us to release the query op only once the query connection has been closed
 - As part of test framework we now check to ensure all view ops are released again

Fixes / Changes:
 - Fixed tests where we were not closing connections
 - Fixed Explain query to close connection
 - Fixed waitForIndex to close connection
 - Fixed emptyAllDocsIndex to close connection

Added temporary disables to two tests when running GSI:
- TestMetricsHandler --> This test fails as we get a prometheus panic due to duplicate metrics
- TestNewDatabaseContextWithOIDCProviderOptionErrors --> Disable the prometheus part of this as we get a prometheus panic due to duplicate metrics

Guess for the above Prometheus metrics is somehow there is another test that doesn't close properly that only runs under GSI. This needs investigation though. Tracked as CBG-1715

- [x] `xattrs=true` GSI http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1193/
- [x] `xattrs=true` Views http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1195/
- [x] `xattrs=false` Views http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1197/

Each have TestDbConfigDoesNotIncludeCredentials failed.
Xattrs True views also has a couple others but are unrelated.

### DIFF
--- a/base/collection.go
+++ b/base/collection.go
@@ -100,10 +100,10 @@ func GetCouchbaseCollection(spec BucketSpec) (*Collection, error) {
 
 	viewOpsQueue := make(chan struct{}, MaxConcurrentQueryOps)
 	collection := &Collection{
-		Collection: bucket.DefaultCollection(),
-		Spec:       spec,
-		cluster:    cluster,
-		viewOps:    viewOpsQueue,
+		Collection:   bucket.DefaultCollection(),
+		Spec:         spec,
+		cluster:      cluster,
+		viewQueryOps: viewOpsQueue,
 	}
 
 	return collection, nil
@@ -113,7 +113,7 @@ type Collection struct {
 	*gocb.Collection               // underlying gocb Collection
 	Spec             BucketSpec    // keep a copy of the BucketSpec for DCP usage
 	cluster          *gocb.Cluster // Associated cluster - required for N1QL operations
-	viewOps          chan struct{} // Manages max concurrent view ops (per kv node)
+	viewQueryOps     chan struct{} // Manages max concurrent view ops (per kv node)
 }
 
 // DataStore
--- a/base/collection.go
+++ b/base/collection.go
@@ -98,12 +98,12 @@ func GetCouchbaseCollection(spec BucketSpec) (*Collection, error) {
 		return nil, err
 	}
 
-	viewOpsQueue := make(chan struct{}, MaxConcurrentQueryOps)
+	queryOpsQueue := make(chan struct{}, MaxConcurrentQueryOps)
 	collection := &Collection{
-		Collection:   bucket.DefaultCollection(),
-		Spec:         spec,
-		cluster:      cluster,
-		viewQueryOps: viewOpsQueue,
+		Collection: bucket.DefaultCollection(),
+		Spec:       spec,
+		cluster:    cluster,
+		queryOps:   queryOpsQueue,
 	}
 
 	return collection, nil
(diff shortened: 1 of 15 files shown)


base/collection.go
```diff
@@ -100,10 +100,10 @@ func GetCouchbaseCollection(spec BucketSpec) (*Collection, error) {
...
 	viewOpsQueue := make(chan struct{}, MaxConcurrentQueryOps)
 	collection := &Collection{
-		Collection: bucket.DefaultCollection(),
-		Spec:       spec,
-		cluster:    cluster,
-		viewOps:    viewOpsQueue,
+		Collection:   bucket.DefaultCollection(),
+		Spec:         spec,
+		cluster:      cluster,
+		viewQueryOps: viewOpsQueue,
```
adamcfraser: What do you think about renaming this to queryOps?  Leaving 'view' in there gives the initial impression it's only for view queries.  Same applies to waitForAvailViewOp and releaseViewOp.

base/collection_view.go
```diff
@@ -242,8 +242,10 @@ func (c *Collection) executeViewQuery(ddoc, name string, params map[string]inter
 
 	// On timeout, return an typed error
 	if isGoCBQueryTimeoutError(err) {
+		c.releaseViewOp()
```
adamcfraser: Looks like we need to either releaseViewOp() on optsErr (line 238), or move the waitForAvailViewOp down to line 240.

base/collection_view.go
```diff
@@ -171,7 +171,8 @@ func (c *Collection) View(ddoc, name string, params map[string]interface{}) (sgb
 
 	if gocbViewResult != nil {
 		viewResultIterator := &gocbRawIterator{
-			rawResult: gocbViewResult,
+			rawResult:                  gocbViewResult,
```
adamcfraser, line 174: Need to releaseViewOp if we exit on error on line 169 I think, as well as if gocbViewResult is nil.  Since we process the entire result here, might be better to skip the gocbRawIterator approach here, and just defer close after calling executeViewQuery.

For the View case, it might be cleaner to just move the `c.waitForAvailViewOp()` up to View and ViewQuery, instead of in executeViewQuery, to reduce the code distance between acquiring and releasing the op.

db/indexes.go
```diff
@@ -363,18 +363,26 @@ func waitForIndex(bucket base.N1QLStore, indexName string, queryStatement string
 	retrySleeper := base.CreateMaxDoublingSleeperFunc(180, 100, 5000)
 	retryCount := 0
 	for {
-		_, err := bucket.Query(queryStatement, nil, base.RequestPlus, true)
-		if err == nil {
+		resultSet, resultsError := bucket.Query(queryStatement, nil, base.RequestPlus, true)
+
+		// Immediately close results. We don't need these
+		resultSetCloseError := resultSet.Close()
```
adamcfraser: Do we need a nil check on resultSet here?  collection.Query looks like it's going to return nil in some error cases, at least.

adamcfraser (changes requested):
Flagged a number of places where it looks like we need to release the view op.  

I understand why this approach is necessary when streaming results, but it's definitely a lot easier to make a coding error and neglect to release the query op.  I don't see any obvious alternatives, though, so am fine with relying on the checkForViewOpsQueueEmptied to catch this kind of error in future.

base/collection_view.go
JRascagneres: Don't think this is required? If there is an error in `executeViewQuery` we already release the view op in there.

JRascagneres:
> I understand why this approach is necessary when streaming results, but it's definitely a lot easier to make a coding error and neglect to release the query op. I don't see any obvious alternatives, though, so am fine with relying on the checkForViewOpsQueueEmptied to catch this kind of error in future.

Agreed its not ideal but like you say not really sure how to get around it. I'm fairly confident that that check will catch any future mistakes.

adamcfraser (approved):

PR #5262, opened 2021-09-29
CBG-1406: Generic API for background processes

JRascagneres:
CBG-1406

- Written generic background process API which manages state
- Currently implemented for tombstone compaction and resync - Allows for easy addition of future APIs like upcoming attachment compaction

Tested with additional tombstone compaction testing and enhancements to resync testing

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1232/
- [x] `xattrs=false` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1233/

Couple failures but seem unrelated.


db/background_mgr.go
```diff
@@ -0,0 +1,200 @@
...
+
+type BackgroundProcessAction string
+
+const (
+	BackgroundProcessActionStart BackgroundProcessAction = "start"
+	BackgroundProcessActionStop  BackgroundProcessAction = "stop"
+)
+
+// BackgroundManagerI this is an interface satisfied by BackgroundManager
+// This is technically not needed, however, it allows easily visibility into exposed methods
```
adamcfraser: I feel like the normal definition of public methods already does a sufficient job of providing visibility to exposed methods, and we should remove this interface if it's not used.

adamcfraser: Also it's not accurate, since SetRunState/GetRunState/SetError are public.

rest/api.go
```diff
@@ -156,35 +156,31 @@ func (h *handler) handleFlush() error {
...
-				defer atomic.CompareAndSwapUint32(&h.db.State, db.DBResyncing, db.DBOffline)
-				defer h.db.ResyncManager.SetRunStatus(db.ResyncStateStopped)
-				_, err := h.db.UpdateAllDocChannels(regenerateSequences)
-				if err != nil {
-					base.Errorf("Error occurred running resync operation: %v", err)
-					h.db.ResyncManager.SetError(err)
-				}
-			}()
+			h.db.ResyncManager.ResetStatus()
+			h.db.ResyncManager.SetRunState(db.BackgroundProcessStateRunning)
```
adamcfraser: I don't think we should require external actors to set the run state - it would be preferablce for BackgroundManager to manage that internally in response to the Start/Stop actions.

In this particular case, it looks to me like we could set the status to 'running' inside 'Start' prior to starting the async goroutine, and perform the GetStatus/writeRawJSON after Start returns.

db/database.go
```diff
@@ -1161,147 +1162,151 @@ func (db *Database) UpdateAllDocChannels(regenerateSequences bool) (int, error)
 
 		var importRow QueryIdRow
 		for results.Next(&importRow) {
-			if db.ResyncManager.ShouldStop() {
+			select {
```
adamcfraser, line 1183: Instead of nesting all the processing inside the 'default', I might just treat this as a check on the terminator channel prior to continuing, like:
```
select {
    case <-terminator:
         clean up and exit
    default:
}
[we now continue our regular programming]
```

db/background_mgr.go
```diff
@@ -0,0 +1,200 @@
...
+func (b *BackgroundManager) SetError(err error) {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
+	b.LastError = err
+	b.State = BackgroundProcessStateError
+}
+
+// ==============================================
+// Resync Implementation of Background Manager
```
```suggestion
// Resync Implementation of Background Manager Process
```

db/background_mgr.go
```diff
@@ -0,0 +1,200 @@
...
+		status = []byte("{}")
+	}
+	b.lock.Unlock()
+
+	processPairs := b.Process.GetProcessStatus()
+
+	retStatus, err := base.InjectJSONProperties(status, processPairs...)
+	if err != nil {
+		return nil
+	}
```
adamcfraser, line 77: I think I'd prefer for GetProcessStatus to pass BackgroundManagerStatus down to the process, and let the process build the marshalled JSON in the usual way.  That would allow the process to omit State or LastError (if that made sense for a given use case), but more importantly would use standard json struct annotations, avoiding the need for 'test only' definitions of those.

adamcfraser (approved):

PR #5280, opened 2021-10-05
CBG-1716: norev uses "sequence" property name except for protocol version 2 and ISGR

IsaacLambat:
CBG-1716

norev uses the "seq" property name to store or read the sequence number on the norev message if the client type is ISGR, and the active subprotocol on BLIP is version 2 to maintain backwards compatibility. Otherwise "sequence" is used as the property name.

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1247/

### DIFF
--- a/db/blip_sync_context.go
+++ b/db/blip_sync_context.go
@@ -518,13 +518,16 @@ func (bsc *BlipSyncContext) sendBLIPMessage(sender *blip.Sender, msg *blip.Messa
 }
 
 func (bsc *BlipSyncContext) sendNoRev(sender *blip.Sender, docID, revID string, seq SequenceID, err error) error {
-
 	base.DebugfCtx(bsc.loggingCtx, base.KeySync, "Sending norev %q %s due to unavailable revision: %v", base.UD(docID), revID, err)
 
 	noRevRq := NewNoRevMessage()
 	noRevRq.SetId(docID)
 	noRevRq.SetRev(revID)
-	noRevRq.SetSeq(seq)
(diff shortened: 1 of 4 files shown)


db/blip_sync_context.go
```diff
@@ -518,13 +518,16 @@ func (bsc *BlipSyncContext) sendBLIPMessage(sender *blip.Sender, msg *blip.Messa
...
 
 func (bsc *BlipSyncContext) sendNoRev(sender *blip.Sender, docID, revID string, seq SequenceID, err error) error {
-
 	base.DebugfCtx(bsc.loggingCtx, base.KeySync, "Sending norev %q %s due to unavailable revision: %v", base.UD(docID), revID, err)
 
 	noRevRq := NewNoRevMessage()
 	noRevRq.SetId(docID)
 	noRevRq.SetRev(revID)
-	noRevRq.SetSeq(seq)
+	if bsc.blipContext.ActiveSubprotocol() == BlipCBMobileReplicationV2 && bsc.clientType == BLIPClientTypeSGR2 {
```
bbrks, line 526: I'd expect an equivalent change to be made in `handleNoRev` (the receiving side). This is only covering the send use-case at the minute.

bbrks (changes requested):

bbrks (dismissed):
LGTM pending integration test

bbrks (approved):

bbrks (approved):

PR #5287, opened 2021-10-07
CBG-1468: Implement Mark phase of attachment compaction

JRascagneres:
CBG-1468

- PR for Mark phase of attachment compaction work
- Adds a SetXattr function allowing only an xattr to be updated without the additional work of adding body hash and cas.

## Dependencies
- [x] https://github.com/couchbase/sg-bucket/pull/66
- [x] https://github.com/couchbaselabs/walrus/pull/59
- [ ] https://github.com/couchbase/sync_gateway/pull/5293

### DIFF
--- a/base/bucket_xattr.go
+++ b/base/bucket_xattr.go
@@ -35,6 +35,10 @@ func (bucket *CouchbaseBucketGoCB) WriteUpdateWithXattr(k string, xattrKey strin
 	return WriteUpdateWithXattr(bucket, k, xattrKey, userXattrKey, exp, previous, callback)
 }
 
+func (bucket *CouchbaseBucketGoCB) SetXattr(k string, xattrKey string, xv []byte) (casOut uint64, err error) {
+	return SetXattr(bucket, k, xattrKey, xv)
+}
+
 func (bucket *CouchbaseBucketGoCB) UpdateXattr(k string, xattrKey string, exp uint32, cas uint64, xv interface{}, deleteBody bool, isDelete bool) (casOut uint64, err error) {
 	return UpdateTombstoneXattr(bucket, k, xattrKey, exp, cas, xv, deleteBody)
 }
@@ -334,6 +338,18 @@ func (bucket *CouchbaseBucketGoCB) SubdocUpdateBodyAndXattr(k string, xattrKey s
 		return 0, err
 	}
 	return uint64(docFragment.Cas()), nil
+}
+
+// SubdocSetXattr performs a set of the given xattr. Does a straight set with no cas.
+func (bucket *CouchbaseBucketGoCB) SubdocSetXattr(k string, xattrKey string, xv interface{}) (casOut uint64, err error) {
+	mutateInBuilder := bucket.Bucket.MutateInEx(k, gocb.SubdocDocFlagAccessDeleted, 0, 0).
+		UpsertEx(xattrKey, xv, gocb.SubdocFlagXattr)
+	docFragment, mutateErr := mutateInBuilder.Execute()
+	if mutateErr == nil || mutateErr == gocbcore.ErrSubDocSuccessDeleted {
+		return uint64(docFragment.Cas()), nil
+	}
+
+	return uint64(0), mutateErr
 }
 
 // SubdocUpdateithXattrOnly upserts an xattr, does not modify body
--- a/db/attachment_compaction.go
+++ b/db/attachment_compaction.go
@@ -1,0 +1,132 @@
+package db
+
+import (
+	"fmt"
+	"strings"
+
+	sgbucket "github.com/couchbase/sg-bucket"
+	"github.com/couchbase/sync_gateway/base"
+)
+
+const CompactionIDKey = "compactID"
+
+func Mark(db *Database, compactionID string, terminator chan bool) (int, error) {
+	base.InfofCtx(db.Ctx, base.KeyAll, "Starting first phase of attachment compaction (mark phase) with compactionID: %q", compactionID)
+	compactionLoggingID := "compaction: " + compactionID
+
+	var markProcessFailureErr error
+	var attachmentsMarked int
+
+	// failProcess used when a failure is deemed as 'un-recoverable' and we need to abort the compaction process.
+	failProcess := func(err error, format string, args ...interface{}) bool {
+		markProcessFailureErr = err
+		close(terminator)
+		base.WarnfCtx(db.Ctx, format, args)
+		return false
+	}
+
+	callback := func(event sgbucket.FeedEvent) bool {
+		// We've had an error previously so no point doing work for any remaining items
+		if markProcessFailureErr != nil {
+			return false
+		}
+
+		// We only want to process full docs. Not any sync docs.
+		if strings.HasPrefix(string(event.Key), base.SyncPrefix) {
+			return true
+		}
+
+		// Attempt to unmarshal the feed data into a document
+		doc, err := UnmarshalDocumentFromFeed(string(event.Key), event.Cas, event.Value, event.DataType, "")
+		if err != nil {
+			return failProcess(err, "[%s] Failed to unmarshal doc %s from feed. Err: %v", compactionLoggingID, base.UD(string(event.Key)), err)
+		}
+
+		// Any doc written by SGW should have this value set. Not having this means we can skip this doc as it has not
+		// got sync data
+		if doc.Cas == uint64(0) {
+			return true
+		}
+
+		// We need to mark attachments in every leaf revision of the current doc
+		// We will build up a list of attachment names which map to attachment doc IDs. Avoids doing multiple KV ops
+		// when marking if multiple leaves are referencing the same attachment.
+		attachmentKeys := make(map[string]string)
+		for _, leafRevision := range doc.History.GetLeaves() {
+			_, _, attachmentMeta, err := db.getRevision(doc, leafRevision)
+			if err != nil {
+				return failProcess(err, "[%s] Failed to get doc %s revision %s. Err: %v", compactionLoggingID, base.UD(string(event.Key)), base.UD(leafRevision), err)
+			}
+
+			// Iterate over the attachments
+			for attID, attMeta := range attachmentMeta {
+				attMetaMap, ok := attMeta.(map[string]interface{})
+				if !ok {
+					continue
+				}
+
+				attVer, ok := GetAttachmentVersion(attMetaMap)
+				if !ok {
+					continue
+				}
+
+				if attVer != AttVersion1 {
+					continue
+				}
+
+				digest, ok := attMetaMap["digest"]
+				if !ok {
+					continue
+				}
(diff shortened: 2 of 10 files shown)


base/bucket_xattr.go
```diff
@@ -336,6 +340,18 @@ func (bucket *CouchbaseBucketGoCB) SubdocUpdateBodyAndXattr(k string, xattrKey s
 	return uint64(docFragment.Cas()), nil
 }
 
+// SubdocSetXattr performs a set of the given xattr in a cas safe way.
+func (bucket *CouchbaseBucketGoCB) SubdocSetXattr(k string, xattrKey string, exp uint32, cas uint64, xv interface{}) (casOut uint64, err error) {
```
adamcfraser: Generally 'Set' is used to describe non-cas-safe operations, and 'Replace' or 'WriteCas' for the type of operation you're doing here.

db/document.go
```diff
@@ -416,47 +416,67 @@ func UnmarshalDocumentSyncData(data []byte, needHistory bool) (*SyncData, error)
...
+	return false, body, nil, nil, nil
+}
+
 // Unmarshals sync metadata for a document arriving via DCP.  Includes handling for xattr content
 // being included in data.  If not present in either xattr or document body, returns nil but no error.
 // Returns the raw body, in case it's needed for import.
 
 // TODO: Using a pool of unmarshal workers may help prevent memory spikes under load
 func UnmarshalDocumentSyncDataFromFeed(data []byte, dataType uint8, userXattrKey string, needHistory bool) (result *SyncData, rawBody []byte, rawXattr []byte, rawUserXattr []byte, err error) {
+	hasXattr, body, syncXattr, userXattr, err := SplitOutComponentsFromFeed(data, dataType, userXattrKey)
```
adamcfraser: I find this change makes UnmarshalDocumentSyncDataFromFeed quite a bit harder to follow (error handling is obscured, 'hasXattr' is vague when there are sync, user, and any other xattrs possible).  More generally I'm not convinced that `SplitOutComponentsFromFeed` adds significant value to what parseXattrStreamData already provides, and having callers check the datatype doesn't seem like that much extra work.

db/document.go
```diff
@@ -416,47 +416,67 @@ func UnmarshalDocumentSyncData(data []byte, needHistory bool) (*SyncData, error)
...
-		body = data
+		return result, body, syncXattr, userXattr, nil
 	}
 
 	// Non-xattr data, or sync xattr not present.  Attempt to retrieve sync metadata from document body
 	result, err = UnmarshalDocumentSyncData(body, needHistory)
 	return result, body, rawUserXattr, nil, err
 }
 
+func UnmarshalDocumentFromFeed(docid string, cas uint64, data []byte, dataType uint8, userXattrKey string) (doc *Document, err error) {
```
adamcfraser, line 460: I'm fine with UnmarshalDocumentFromFeed as a helper function, but as mentioned above, I think it's clearer if you just do the dataType check here.

db/document.go
```diff
@@ -416,47 +416,67 @@ func UnmarshalDocumentSyncData(data []byte, needHistory bool) (*SyncData, error)
...
+		return nil, err
+	}
+
+	if hasXattr {
+		doc, err = unmarshalDocumentWithXattr(docid, body, syncXattr, userXattr, cas, DocUnmarshalAll)
+	} else {
+		doc, err = unmarshalDocument(docid, data)
+	}
+
+	if err != nil {
```
adamcfraser: I usually skip the error check if the only thing we're doing afterward is return.  Can just switch to returning on line 467 and 469.

db/attachment_compaction.go
```diff
@@ -0,0 +1,114 @@
...
+	var markProcessFailureErr error
+	var attachmentsMarked int
+	callback := func(event sgbucket.FeedEvent) bool {
+		// We've had an error previously so no point doing work for any remaining items
+		if markProcessFailureErr != nil {
+			return false
+		}
+
+		// We only want to process full docs. Not any sync docs.
+		if strings.HasPrefix(string(event.Key), "_") {
```
adamcfraser: Our criteria for importing docs is `strings.HasPrefix(key, base.SyncPrefix)` - I think we should be using the same here.

We also want to return `true` in this case - we can consider this document as processed, and allow the DCP client to update the stream metadata accordingly.

db/attachment_compaction.go
```diff
@@ -0,0 +1,114 @@
...
+			return false
+		}
+
+		// Attempt to unmarshal the feed data into a document
+		doc, err := UnmarshalDocumentFromFeed(string(event.Key), event.Cas, event.Value, event.DataType, "")
+		if err != nil {
+			return false
```
adamcfraser: Should this log a warning, or possibly have the process track the number of such documents encountered and report in its status?  

Here (and in all other cases we're returning false), review whether we actually want to avoid persisting a DCP checkpoint based on this mutation.  I think the answer is generally no, and we should return true.

db/attachment_compaction.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+
+		// We need to mark attachments in every leaf revision
+		for _, leafRevision := range doc.History.GetLeaves() {
+			_, _, attachmentMeta, err := db.getRevision(doc, leafRevision)
+			if err != nil {
+				return false
```
adamcfraser: If we hit an error with a single leaf, should we continue and process the other leaves? 

On the other hand, if you think this should be treated as a fatal error for the compaction process, you need a way to stop the process - returning false alone isn't going to do that.

db/attachment_compaction.go
```diff
@@ -0,0 +1,114 @@
...
+
+				if attVer != AttVersion1 {
+					continue
+				}
+
+				digest := attMetaMap["digest"]
+				attKey := MakeAttachmentKey(AttVersion1, string(event.Key), digest.(string))
+
+				// Stamp the current compaction ID into the attachment xattr. This is performing the actual marking
+				_, err = db.Bucket.WriteXattrUpdate(attKey, "_sync", func(xattr []byte) (updatedXattr []byte, err error) {
```
adamcfraser: It seems like it would be common for conflicting leaf revisions to contain references to the same attachment.  I feel like you should be building a slice of all V1 attachments associated with the document, then iterating over that set once to stamp the compaction ID.

db/attachment_compaction.go
```diff
@@ -0,0 +1,114 @@
...
+
+					return xattr, nil
+				})
+
+				// If an error occurs while stamping in that ID we need to fail this process and then the entire compaction
+				// process. Otherwise an attachment could end up getting erroneously deleted in the later sweep phase.
+				if err != nil {
+					markProcessFailureErr = err
+					terminator <- true
+					base.DebugfCtx(db.Ctx, base.KeyAll, "[%s] Failed to mark attachment %s from doc %s rev %s with attachment docID %s. Err: %v", compactionLoggingID, attID, event.Key, leafRevision, attKey, err)
```
adamcfraser: If we're killing the process, I assume we can also exit the callback here (definitely returning false this time, so that the compaction process could be retried in the case of transient server issues).

db/attachment_compaction.go
```diff
@@ -0,0 +1,114 @@
...
+				}
+
+				base.DebugfCtx(db.Ctx, base.KeyAll, "[%s] Marked attachment %s from doc %s rev %s with attachment docID %s", compactionLoggingID, attID, event.Key, leafRevision, attKey)
+				attachmentsMarked++
+			}
+		}
+		return false
```
adamcfraser: Why are we returning false here?

adamcfraser (changes requested):
General approach looks ok, some questions about the implementation details.

base/bucket_xattr.go
JRascagneres: Swapping it out for an actual set now anyway without cas - no need for this with the setting of the xattr on the attachment.

db/attachment_compaction.go
JRascagneres: To be honest I wasn't entirely sure what this was doing and just copied false from the test you had written. I think I understand that now so will go over it.

db/document.go
JRascagneres: Not sure I agree with this. Perhaps we should rename the `hasXattr` but otherwise I think this makes it easier to read to be honest. It adds some duplication.

Its not too much duplication or change though so I'm okay to be convinced otherwise. I'll switch it though if you think it makes it hard to read.

db/attachment_compaction.go
```diff
@@ -0,0 +1,129 @@
...
+		return 0, err
+	}
+	defer dcpClient.Close()
+
+	doneChan, err := dcpClient.Start()
+	if err != nil {
+		return 0, err
+	}
+
+	<-doneChan
```
adamcfraser: Should this be in a select to be able to react to a user stopping the compact via REST API?  Or is that work still pending?

JRascagneres: Guessing I would've noticed this once I got around to the REST API work but makes sense to do this now :P

manifest/default.xml
```diff
@@ -40,7 +40,7 @@ licenses/APL2.txt.
 
   <project name="sync_gateway_admin_ui" path="godeps/src/github.com/couchbaselabs/sync_gateway_admin_ui" revision="93c74bac9ddc2979ab895a37087c225c998b03bf" remote="couchbaselabs"/>
 
-  <project name="walrus" path="godeps/src/github.com/couchbaselabs/walrus" remote="couchbaselabs" revision="2594ed1ddc60c95a30b7e3fc76cfeda2cc74b1fd"/>
+  <project name="walrus" path="godeps/src/github.com/couchbaselabs/walrus" remote="couchbaselabs" revision="CBG-1468"/>
```
adamcfraser:
```suggestion
  <project name="walrus" path="godeps/src/github.com/couchbaselabs/walrus" remote="couchbaselabs" revision="a5872ff542ba9cf0452528bdbe7ec3f7e90e73d0"/>
```

manifest/default.xml
```diff
@@ -59,7 +59,7 @@ licenses/APL2.txt.
 
   <project name="gomemcached" path="godeps/src/github.com/couchbase/gomemcached" remote="couchbase" revision="4201f50df22f08f75a0491451fa6507387d41304"/>
 
-  <project name="sg-bucket" path="godeps/src/github.com/couchbase/sg-bucket" remote="couchbase" revision="08f9bc2aa1cb5cb87b9a9bc0f803ddefbc73c5f8"/>
+  <project name="sg-bucket" path="godeps/src/github.com/couchbase/sg-bucket" remote="couchbase" revision="CBG-1468"/>
```
adamcfraser:
```suggestion
  <project name="sg-bucket" path="godeps/src/github.com/couchbase/sg-bucket" remote="couchbase" revision="1972e5eab603e22c8be5e75cffc6f78de4a8b70c"/>
```

adamcfraser (dismissed):

adamcfraser (approved):

PR #5290, opened 2021-10-12
CBG-1469: Attachment compaction Sweep

JRascagneres:
CBG-1469

Implements the sweep phase of the attachment cleanup, deleting any attachments which are not marked with the same ID as the current attachment cleanup phase.

## Dependencies
- [x] https://github.com/couchbase/sync_gateway/pull/5287

### DIFF
--- a/db/attachment_compaction.go
+++ b/db/attachment_compaction.go
@@ -154,7 +154,7 @@ func Sweep(db *Database, compactionID string, terminator chan bool) (int, error)
 
 		// If the data contains an xattr then the attachment likely has a compaction ID, need to check this value
 		if event.DataType&base.MemcachedDataTypeXattr != 0 {
-			_, xattr, _, err := parseXattrStreamData(base.SyncXattrName, "", event.Value)
+			_, xattr, _, err := parseXattrStreamData(base.AttachmentCompactionXattrName, "", event.Value)
 			if err != nil && !errors.Is(err, base.ErrXattrNotFound) {
 				base.WarnfCtx(db.Ctx, "[%s] Unexpected error occurred attempting to parse attachment xattr: %v", compactionLoggingID, err)
 				return true
--- a/db/attachment_compaction.go
+++ b/db/attachment_compaction.go
@@ -215,8 +215,12 @@ func Sweep(db *Database, compactionID string, terminator chan bool) (int, error)
 		return 0, err
 	}
 
-	<-doneChan
-	base.InfofCtx(db.Ctx, base.KeyAll, "[%s] Sweep phase of attachment compaction completed. Deleted %d attachments", compactionLoggingID, attachmentsDeleted)
+	select {
+	case <-doneChan:
+		base.InfofCtx(db.Ctx, base.KeyAll, "[%s] Sweep phase of attachment compaction completed. Deleted %d attachments", compactionLoggingID, attachmentsDeleted)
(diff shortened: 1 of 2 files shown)


db/attachment_compaction.go
```diff
@@ -127,3 +128,88 @@ func Mark(db *Database, compactionID string, terminator chan bool) (int, error)
...
+	// be deleted.
+	callback := func(event sgbucket.FeedEvent) bool {
+		// We only want to look over v1 attachment docs, skip otherwise
+		if !strings.HasPrefix(string(event.Key), base.AttPrefix) {
+			return true
+		}
+
+		// If the data contains an xattr then the attachment likely has a compaction ID, need to check this value
+		if event.DataType&base.MemcachedDataTypeXattr != 0 {
+			_, xattr, _, err := parseXattrStreamData(base.SyncXattrName, "", event.Value)
```
adamcfraser: I think we should use something other than _sync for the xattr name.

db/attachment_compaction.go
```diff
@@ -127,3 +128,88 @@ func Mark(db *Database, compactionID string, terminator chan bool) (int, error)
...
+				err = base.JSONUnmarshal(xattr, &syncData)
+				if err != nil {
+					base.WarnfCtx(db.Ctx, "[%s] Failed to unmarshal xattr data: %v", compactionLoggingID, err)
+					return true
+				}
+
+				docCompactID, ok := syncData[CompactionIDKey]
+				if ok && docCompactID == compactionID {
+					return true
+				}
```
adamcfraser, line 176: Since we're doing a set in the mark phase, what do we want to do if there's a mismatch in compactionID?  Deleting the attachment doesn't seem appropriate if another compaction has somehow been started and overwritten the previous ID.

db/attachment_compaction.go
```diff
@@ -127,3 +128,88 @@ func Mark(db *Database, compactionID string, terminator chan bool) (int, error)
...
+			}
+		}
+
+		// If we've reached this point the current v1 attachment being processed either:
+		// - Has no compactionID set in its xattr
+		// - Has a compactionID set in its xattr but it is from a previous run and therefore is not equal to the passed
+		// in compactionID
+		// Therefore, we want to purge the doc
+		err := db.Bucket.Delete(string(event.Key))
```
adamcfraser: I don't think we have a use case for concurrent update, but doing a CAS-safe delete still seems prudent here.

adamcfraser (changes requested):

db/attachment_compaction.go
JRascagneres: Yeah that seems fair, can swap this out for `Remove`

db/attachment_compaction.go
JRascagneres: If there is a mismatch in compactionID pretty sure we do want to still delete it. What happens if there has been a previous run that marked it but now it has been removed from the doc it was attached to? Its not going to lose that xattr with the previous mark ID. 
I think we just need to be careful about not allowing two runs at the same time. Storing a compaction ID / status in the bucket for persistence should avoid this scenario.

db/attachment_compaction.go
JRascagneres: _sync:compact

adamcfraser (approved):

PR #5291, opened 2021-10-13
CBG-1714: Unsupported and logger config is hidden if not used from config endpoint and migrated startup config

IsaacLambat:
CBG-1714

Unsupported and logger configs in startup config are now pointers and nil checked around the code. 

- [ ] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1258/

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -281,7 +281,9 @@ func (h *handler) handleGetConfig() error {
 			}
 		}
 
-		cfg.Logging = *base.BuildLoggingConfigFromLoggers(h.server.config.Logging.RedactionLevel, h.server.config.Logging.LogFilePath)
+		if h.server.config.Logging != nil {
+			cfg.Logging = base.BuildLoggingConfigFromLoggers(h.server.config.Logging.RedactionLevel, h.server.config.Logging.LogFilePath)
+		}
 		cfg.Databases = databaseMap
 
 		h.writeJSON(cfg)
@@ -1072,7 +1074,10 @@ func (h *handler) handleSGCollect() error {
 
 	zipFilename := sgcollectFilename()
 
-	logFilePath := h.server.config.Logging.LogFilePath
+	logFilePath := ""
+	if h.server.config.Logging != nil {
+		logFilePath = h.server.config.Logging.LogFilePath
+	}
 
 	if err := sgcollectInstance.Start(logFilePath, h.serialNumber, zipFilename, params); err != nil {
 		return base.HTTPErrorf(http.StatusInternalServerError, "Error running sgcollect_info: %v", err)
--- a/rest/admin_api_test.go
+++ b/rest/admin_api_test.go
@@ -3314,9 +3314,11 @@ func TestIncludeRuntimeStartupConfig(t *testing.T) {
 	require.NoError(t, err)
 
 	// Check that db revs limit is there now and error logging config
+	require.NotNil(t, runtimeServerConfigResponse.Databases)
 	assert.Contains(t, runtimeServerConfigResponse.Databases, "db")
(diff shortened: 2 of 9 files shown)


rest/server_context.go
```diff
@@ -1044,7 +1044,7 @@ type statsWrapper struct {
 
 func (sc *ServerContext) startStatsLogger() {
 
-	if sc.config.Unsupported.StatsLogFrequency == nil || sc.config.Unsupported.StatsLogFrequency.Value() == 0 {
+	if sc.config.Unsupported == nil || sc.config.Unsupported.StatsLogFrequency == nil || sc.config.Unsupported.StatsLogFrequency.Value() == 0 {
```
bbrks, line 1047: Do both of these checks need to be `&&` `!= nil` for the comment below to make sense?

rest/admin_api_test.go
```diff
@@ -3314,9 +3314,11 @@ func TestIncludeRuntimeStartupConfig(t *testing.T) {
 	require.NoError(t, err)
 
 	// Check that db revs limit is there now and error logging config
+	require.NotNil(t, runtimeServerConfigResponse.Databases)
 	assert.Contains(t, runtimeServerConfigResponse.Databases, "db")
 	assert.Equal(t, base.Uint32Ptr(100), runtimeServerConfigResponse.Databases["db"].RevsLimit)
 
+	require.NotNil(t, runtimeServerConfigResponse.Logging)
 	assert.NotNil(t, runtimeServerConfigResponse.Logging.Error)
 	assert.Equal(t, "debug", runtimeServerConfigResponse.Logging.Console.LogLevel.String())
```
bbrks, line 3323: Should we add an assertion for a Nil `unsupported` field?

rest/admin_api.go
```diff
@@ -281,7 +281,9 @@ func (h *handler) handleGetConfig() error {
 			}
 		}
 
-		cfg.Logging = *base.BuildLoggingConfigFromLoggers(h.server.config.Logging.RedactionLevel, h.server.config.Logging.LogFilePath)
+		if h.server.config.Logging != nil {
```
bbrks, line 284: I'm not sure this nil check is nessesary or correct.

`BuildLoggingConfigFromLoggers` is building a config from current runtime logger settings (that can be changed via the REST API, and won't nessisarily be the same as the initial values in `h.server.config`)

rest/admin_api.go
```diff
@@ -1072,7 +1074,10 @@ func (h *handler) handleSGCollect() error {
 
 	zipFilename := sgcollectFilename()
 
-	logFilePath := h.server.config.Logging.LogFilePath
+	logFilePath := ""
+	if h.server.config.Logging != nil {
```
bbrks, line 1078: It looks like `h.server.config.Logging.LogFilePath` is set to `DefaultLogFilePath` on startup. Is the nil check nessesary, and if logging is nil, does it make sense to pass down an empty path? I'm not sure the sgcollect code can handle that.

bbrks (changes requested):
As you can see below, the code we've got in this PR isn't really addressing the original needs of the ticket (omit the empty logging and unsupported fields), and is just adding more complexity for little gain.

Screenshot 2021-10-14 at 13 25 22

I think we're still running up against the flag config code populating non-nil empty structs, which at this point seems nearly unavoidable without further work in that area.

I think we should strongly consider abandoning the ticket, including the `unsupported` part until we can get a more comprehensive approach to turning config into runtime options.

bbrks:
Decided to postpone this requirement until we're able to get a fully implemented `ServerContextOptions` thing that avoids mixing config and runtime options.

PR #5292, opened 2021-10-13
CBG-1719: Fixed flakey integration test for TestUnprocessibleDelta

IsaacLambat:
CBG-1719

Replication is used continuously instead of one shots, along with using a defer to stop the replicator at the end of the function. 

- [ ] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1266/

### DIFF
--- a/rest/replicator_test.go
+++ b/rest/replicator_test.go
@@ -5814,10 +5814,19 @@ func TestUnprocessableDeltas(t *testing.T) {
 		ReplicationStatsMap: base.SyncGatewayStats.NewDBStats(t.Name(), true, false, false).DBReplicatorStats(t.Name()),
 	})
 	assert.Equal(t, "", ar.GetStatus().LastSeqPush)
+
 	assert.NoError(t, ar.Start())
-	defer func() { assert.NoError(t, ar.Stop()) }()
 
 	err = passiveRT.waitForRev("test", revID)
+	require.NoError(t, err)
+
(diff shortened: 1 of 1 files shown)


rest/replicator_test.go
```diff
@@ -5833,10 +5827,14 @@ func TestUnprocessibleDeltas(t *testing.T) {
 	rev.BodyBytes = []byte("{invalid}")
 	passiveRT.GetDatabase().GetRevisionCacheForTest().Upsert(rev)
 
-	assert.NoError(t, ar.Start())
-	// Wait for active to replicate to passive
-	activeRT.waitForReplicationStatus(t.Name(), db.ReplicationStateStopped)
-	// Check if it managed to replicate
+	// Make 2nd revision
+	resp = activeRT.SendAdminRequest(http.MethodPut, "/db/test?rev="+revID, `{"field1":"f1_2","field2":"f2_2"}`)
```
bbrks: Making the 2nd revision after we've modified rev1's body in the revcache means this test isn't going to be triggering the 422 error case any more, because it's attempting to pull a valid rev2 - unless I'm mistaken?

bbrks (changes requested):
I think we've unintentionally changed the test to not cover the original error.

Could you maybe change the error code back to InternalServerError that went in along with this test originally to see if the test now fails?

bbrks (approved):

PR #5296, opened 2021-10-14
Ensure query ops config option is set for GoCBV2

JRascagneres:
Looks like somehow the query ops limiting config option wasn't getting applied for gocb v2 and would just always use the default of 1000. This PR fixes that.

### DIFF
--- a/base/collection.go
+++ b/base/collection.go
@@ -110,7 +110,11 @@ func GetCollectionFromCluster(cluster *gocb.Cluster, spec BucketSpec, waitUntilR
 	}
 
 	// Set limits for concurrent query and kv ops
-	collection.queryOps = make(chan struct{}, MaxConcurrentQueryOps)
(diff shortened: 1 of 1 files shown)


base/collection.go
```diff
@@ -110,7 +110,11 @@ func GetCollectionFromCluster(cluster *gocb.Cluster, spec BucketSpec, waitUntilR
 	}
 
 	// Set limits for concurrent query and kv ops
-	collection.queryOps = make(chan struct{}, MaxConcurrentQueryOps)
+	maxConcurrentQueryOps := MaxConcurrentQueryOps
```
bbrks, line 113: Maybe rename the const while were here, so it's clear how it's used.

```suggestion
	maxConcurrentQueryOps := DefaultMaxConcurrentQueryOps
```

bbrks (changes requested):

PR #5305, opened 2021-10-20
Increase litecore test verbosity

bbrks:
Would like trace logs to investigate raw blip frames as part of CBG-1739

- [x] https://github.com/couchbaselabs/sg_dev_tools/pull/28


JRascagneres (approved):
lgtm

PR #5306, opened 2021-10-21
CBG-1756: Improve mark phase unmarshal efficiency

JRascagneres:
CBG-1756

PR aims to improve the efficiency of the attachment compaction mark phase. Mainly:
- Avoid marshaling body if not required
- Only look over body map / body key map if a conflict is present
- If the above is deems we need to look over body map / body key only marshal the required data from these
- If the document sync data does not contain an attachment we will scan the body too for pre-2.5 attachments.

Added to an existing test to ensure items in body map / body key map are looked at. 

It may be valid to remove the work to look in body map / body key depending on resolution of CBG-1757 
- If we have an issue where these weren't being backed up there is no need to do this work to look through them.

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1294/
(Couple unrelated looking failures)
- [ ] `xattrs=false` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration//
TODO

### DIFF
--- a/db/attachment_compaction.go
+++ b/db/attachment_compaction.go
@@ -60,11 +60,6 @@ func Mark(db *Database, compactionID string, terminator chan struct{}, markedAtt
 
 		// If we're in a conflict state we need to go and check and mark attachments from other leaves, not just winning
 		if attachmentData.Flags&channels.Conflict != 0 {
-
-			type AttachmentsMetaMap struct {
-				Attachments map[string]AttachmentsMeta `json:"_attachments"`
-			}
-
 			// Iterate over body map
 			// These are strings containing conflicting bodies, need to scan these for attachments
 			for _, bodyMap := range attachmentData.History.BodyMap {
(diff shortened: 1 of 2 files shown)


db/attachment_compaction.go
```diff
@@ -165,34 +165,65 @@ type AttachmentCompactionData struct {
...
-	if err != nil {
-		return nil, err
+
+	return attachmentData, nil
+}
+
+// checkForInlineAttachments will scan a body for "_attachments" for pre-2.5 attachments and will return any attachments
+// found
+func checkForInlineAttachments(body []byte) (*AttachmentsMetaMap, error) {
+	if strings.Contains(string(body), BodyAttachments) {
```
adamcfraser: Should use bytes.Contains here, string(body) will unnecessarily allocate a new string.

adamcfraser (changes requested):
Generally looks good - minor change suggested for Contains check.

adamcfraser (approved):

PR #5307, opened 2021-10-21
CBG-1761: Fixed panic and added meaningful error in sanitizeDbConfigs when no server address is provided

IsaacLambat:
CBG-1761

- Added nil check when dereferencing bucket server
- Added error when no server is specified in any of the dbConfigs
- Unit test

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1​29​0

### DIFF
--- a/rest/main.go
+++ b/rest/main.go
@@ -263,17 +263,20 @@ func automaticConfigUpgrade(configPath string) (sc *StartupConfig, disablePersis
 // validate / sanitize db configs
 // - remove fields no longer valid for persisted db configs
 // - ensure servers are the same
+// - ensure a server is provided in at least one db config
 func sanitizeDbConfigs(configMap DbConfigMap) (DbConfigMap, error) {
 	var databaseServerAddress string
 
 	for dbName, dbConfig := range configMap {
-		if databaseServerAddress == "" {
-			databaseServerAddress = *dbConfig.Server
-		}
+		if dbConfig.Server != nil {
+			if databaseServerAddress == "" {
+				databaseServerAddress = *dbConfig.Server
+			}
 
-		if *dbConfig.Server != databaseServerAddress {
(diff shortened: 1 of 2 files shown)


rest/main.go
```diff
@@ -294,6 +297,11 @@ func sanitizeDbConfigs(configMap DbConfigMap) (DbConfigMap, error) {
 		// Make sure any updates are written back to the config
 		configMap[dbName] = dbConfig
 	}
+
+	if databaseServerAddress == "" {
+		return nil, fmt.Errorf("automatic upgrade to persistent config requires at least 1 database config to have a " +
+			"server address specified in the 2.x config")
```
bbrks: I'm not exactly sure "at least 1" is true. I think _all_ databases must have one set, and _all_ must match.
We don't really have a way of doing persistent config and auto-upgrade in any other case.

Otherwise we'll be attempting to persist at least one database config to a server that would've originally been backed by walrus.

In short, erroring out whilst iterating as soon as we see a nil dbConfig.Server is probably the right thing to do.

bbrks (changes requested):

rest/main.go
```diff
@@ -263,10 +263,16 @@ func automaticConfigUpgrade(configPath string) (sc *StartupConfig, disablePersis
 // validate / sanitize db configs
 // - remove fields no longer valid for persisted db configs
 // - ensure servers are the same
+// - ensure a server is provided in at least one db config
```
bbrks:
```suggestion
// - ensure matching servers are provided in all db configs
```

bbrks (dismissed):
minor change to comment to describe updated behaviour, but otherwise looks good after test finishes

bbrks (approved):

PR #5309, opened 2021-10-25
CBG-1755: Handle multiple compaction IDs

JRascagneres:
CBG-1755

- Switch to storing multiple compaction IDs in an xattr instead of just one --> Have a map with compactionID - timestamp
- Mark adds a compactionID with subdoc op
- Sweep reads the compactionIDs and does a basic check if the current running compactionID is present in the map
- Adds an additional phase which does a 'cleanup':
	- Deletes the current running compactionID from the map
	- If any other runs are more than 30 days old we also delete those (can occur if an earlier cleanup phase fails)

- Adds RemoveXattr which does a cas safe subdoc delete operation
- Adds DeleteXattr which does a straight delete operation with no cas check

## Dependencies
- [x] https://github.com/couchbaselabs/walrus/pull/60
- [x] https://github.com/couchbase/sg-bucket/pull/67

## Integration Tests
- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1319/
- [x] `xattrs=false` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1313/

### DIFF
--- a/base/bucket_xattr.go
+++ b/base/bucket_xattr.go
@@ -39,6 +39,14 @@ func (bucket *CouchbaseBucketGoCB) SetXattr(k string, xattrKey string, xv []byte
 	return SetXattr(bucket, k, xattrKey, xv)
 }
 
+func (bucket *CouchbaseBucketGoCB) RemoveXattr(k string, xattrKey string, cas uint64) (err error) {
+	return RemoveXattr(bucket, k, xattrKey, cas)
+}
+
+func (bucket *CouchbaseBucketGoCB) DeleteXattr(k string, xattrKey string) (err error) {
+	return DeleteXattr(bucket, k, xattrKey)
+}
+
 func (bucket *CouchbaseBucketGoCB) UpdateXattr(k string, xattrKey string, exp uint32, cas uint64, xv interface{}, deleteBody bool, isDelete bool) (casOut uint64, err error) {
 	return UpdateTombstoneXattr(bucket, k, xattrKey, exp, cas, xv, deleteBody)
 }
--- a/base/bucket_xattr.go
+++ b/base/bucket_xattr.go
@@ -43,8 +43,8 @@ func (bucket *CouchbaseBucketGoCB) RemoveXattr(k string, xattrKey string, cas ui
 	return RemoveXattr(bucket, k, xattrKey, cas)
 }
 
-func (bucket *CouchbaseBucketGoCB) DeleteXattr(k string, xattrKey string) (err error) {
-	return DeleteXattr(bucket, k, xattrKey)
+func (bucket *CouchbaseBucketGoCB) DeleteXattrs(k string, xattrKeys ...string) (err error) {
+	return DeleteXattrs(bucket, k, xattrKeys...)
 }
 
 func (bucket *CouchbaseBucketGoCB) UpdateXattr(k string, xattrKey string, exp uint32, cas uint64, xv interface{}, deleteBody bool, isDelete bool) (casOut uint64, err error) {
@@ -207,6 +207,24 @@ func (bucket *CouchbaseBucketGoCB) SubdocDeleteXattr(k string, xattrKey string, 
 		return nil
 	}
 	return mutateErrDeleteXattr
+}
+
+// SubdocDeleteXattrs will delete the supplied xattr keys from a document. Not a cas safe operation.
+func (bucket *CouchbaseBucketGoCB) SubdocDeleteXattrs(k string, xattrKeys ...string) error {
+	bucket.singleOps <- struct{}{}
+	defer func() {
+		<-bucket.singleOps
+	}()
+
+	builder := bucket.Bucket.MutateIn(k, 0, 0)
+
+	for _, xattrKey := range xattrKeys {
+		builder.RemoveEx(xattrKey, gocb.SubdocFlagXattr)
+	}
+
+	_, mutateErr := builder.Execute()
+
+	return mutateErr
 }
 
 // SubdocDeleteBodyAndXattr removes the body and specified xattr for a document.  Used
--- a/base/collection_xattr_common.go
+++ b/base/collection_xattr_common.go
@@ -1,10 +1,9 @@
 package base
 
 import (
-	"errors"
 	"fmt"
+	"strings"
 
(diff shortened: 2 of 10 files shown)


base/collection_xattr_common.go
```diff
@@ -275,6 +277,56 @@ func SetXattr(store SubdocXattrStore, k string, xattrKey string, xv []byte) (cas
...
 }
 
+func RemoveXattr(store SubdocXattrStore, k string, xattrKey string, cas uint64) error {
+	worker := func() (shouldRetry bool, err error, value interface{}) {
+		writeErr := store.SubdocDeleteXattr(k, xattrKey, cas)
+		if writeErr == nil {
+			return false, nil, nil
+		}
+
+		if errors.Is(err, gocb.ErrCasMismatch) {
```
adamcfraser: The common wrappers in collection_xattr_common shouldn't have any gocb version-specific code, including error handling.  Could use IsCasMismatch here, but since we're just returning the error anyway, I think we can just omit this check and have RemoveXattr return the underlying error, so that callers could check for IsCasMismatch.

If you really think we need to return a typed error here, I think it should be a new ErrCasMismatch error.  ErrCasFailureShouldRetry has specific meaning for the WriteUpdate operations.  And any new error should be added to IsCasMismatch to ensure it continues working as expected.

db/attachment_compaction.go
```diff
@@ -344,3 +345,129 @@ func Sweep(db *Database, compactionID string, terminator chan struct{}, purgedAt
...
+
+	callback := func(event sgbucket.FeedEvent) bool {
+
+		docID := string(event.Key)
+
+		if !strings.HasPrefix(docID, base.AttPrefix) {
+			return true
+		}
+
+		if event.DataType&base.MemcachedDataTypeXattr != 0 {
```
adamcfraser: Maybe exit early in for non-xattr data types to reduce nesting?

db/attachment_compaction.go
```diff
@@ -344,3 +345,129 @@ func Sweep(db *Database, compactionID string, terminator chan struct{}, purgedAt
...
+			_, xattr, _, err := parseXattrStreamData(base.AttachmentCompactionXattrName, "", event.Value)
+			if err != nil && !errors.Is(err, base.ErrXattrNotFound) {
+				base.WarnfCtx(db.Ctx, "[%s] Unexpected error occurred attempting to parse attachment xattr: %v", compactionLoggingID, err)
+				return true
+			}
+
+			if xattr != nil {
+				var syncData map[string]interface{}
```
adamcfraser: syncData is a confusing name for this variable, since it's not sync metadata.

db/attachment_compaction.go
```diff
@@ -344,3 +345,129 @@ func Sweep(db *Database, compactionID string, terminator chan struct{}, purgedAt
...
+				err = base.JSONUnmarshal(xattr, &syncData)
+				if err != nil {
+					base.WarnfCtx(db.Ctx, "[%s] Failed to unmarshal xattr data: %v", compactionLoggingID, err)
```
adamcfraser:
```suggestion
					base.WarnfCtx(db.Ctx, "[%s] Failed to unmarshal attachment comppaction xattr: %v", compactionLoggingID, err)
```

db/attachment_compaction.go
```diff
@@ -344,3 +345,129 @@ func Sweep(db *Database, compactionID string, terminator chan struct{}, purgedAt
...
+
+				// Get compactID map containing all compactIDs on the document, if one is not present for some reason we
+				// can skip this
+				compactIDSync, compactIDSyncPresent := syncData[CompactionIDKey]
+				if !compactIDSyncPresent {
+					return true
+				}
+
+				// Attempt to type assert compaction ID a map, so we can iterate over the compact IDs
+				compactIDSyncMap, ok := compactIDSync.(map[string]interface{})
```
adamcfraser: Can we define a struct for the attachment compaction metadata and unmarshal to that in the first place?  I think it would simplify this code and also make it easier to see what's going on.

db/attachment_compaction.go
```diff
@@ -344,3 +345,129 @@ func Sweep(db *Database, compactionID string, terminator chan struct{}, purgedAt
...
+					if err != nil && !errors.Is(err, base.ErrCasFailureShouldRetry) {
+						base.WarnfCtx(db.Ctx, "[%s] Failed to remove compaction ID xattr for doc %s: %v", compactionLoggingID, base.UD(docID), err)
+						return true
+					}
+
+				}
+
+				// If we only want to remove select compact IDs iterate over them and delete each one
+				for compactID, _ := range toDelete {
+					err = db.Bucket.DeleteXattr(docID, getCompactionIDSubDocPath(compactID))
```
adamcfraser: Can we avoid multiple kv ops when deleting multiple compaction IDs?

adamcfraser (changes requested):

db/attachment_compaction.go
JRascagneres: Yeah thats valid, have swapped DeleteXattr to DeleteXattrs and added an additional underlying subdoc op which allows multiple paths to be provided.

base/bucket_xattr.go
```diff
@@ -39,6 +39,14 @@ func (bucket *CouchbaseBucketGoCB) SetXattr(k string, xattrKey string, xv []byte
 	return SetXattr(bucket, k, xattrKey, xv)
 }
 
+func (bucket *CouchbaseBucketGoCB) RemoveXattr(k string, xattrKey string, cas uint64) (err error) {
+	return RemoveXattr(bucket, k, xattrKey, cas)
+}
+
+func (bucket *CouchbaseBucketGoCB) DeleteXattrs(k string, xattrKeys ...string) (err error) {
```
adamcfraser, line 46: I think we may want to revisit the naming of these functions (as well as some existing functions) if we're passing a full path as 'xattrKey'.  The previous subdoc operations treated xattrKey as the name of the xattr, but the underlying implementation may support passing in a full path in some cases.  Not critical for this PR, will file an enhancement ticket.

adamcfraser (commented):
Looks good - just need to update manifest.

bbrks (approved):

PR #5310, opened 2021-10-26
Use base.TestsDisableGSI value in TestAllDatabaseNames

bbrks:
This test was running GSI stuff, due to the the default values, even though `TestsDisableGSI` was set.


JRascagneres (approved):
LGTM

PR #5317, opened 2021-11-02
CBG-1750: Implement bucket persistence of attachment compaction

JRascagneres:
CBG-1750

For the attachment compaction work we want to support two new things:
 - Only allow the process to run one instance of compaction at a time
 - Allow GET to be called from other nodes so that all nodes can access the current run status
 - Make the process resumable

We want to make this as generic as possible so that this can be implemented for other features. In the resumable case this is attachment compaction specific but the ability to store status in the cluster as well as maintaining heartbeat docs to limit the process to one running instance can easily be applied to other processes, such as resync.

---
**Changes**
- New state -- Completed -- Signifies that a process has completed. Previous 'stopped' now signifies that the process either crashed or was manually stopped prior to completion. 
 - When attempting to mark start - create a heartbeat document
 		- If document already exists, fail out and error 
 		- If document is successfully created kick off a heartbeat go routine which will maintain that document for the lifetime of the process
 - Check if a process status document exists in the bucket
 		- If one exists in a 'stopped' state we can just continue - new compactID
 		- If one exists but is not in a 'stopped' state then we need to resume the previous run - use old compactID and start at correct phase
 - Kick off a goroutine which updates the status
 		- This polls the 'local' status for the process and updates the cluster status
 - GetStatus will now return the cluster status document if it exists. If the cluster status document exists and a heartbeat doesn't then we need to do a check. Basically if the cluster status state isn't 'stopped' / 'errored' / 'aborted' then this processed crashed, but it'd be weird to return the status to the user as 'running'. Therefore if no heartbeat and it isn't in one of the aforementioned states we set it to 'aborted' as we know it must have failed.
 - Once process completes:
 		- Stop goroutines
 		- Delete heartbeat doc
 		- Set status doc to stopped
---

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1382/
- [x] `xattrs=false` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1385/

### DIFF
--- a/db/background_mgr.go
+++ b/db/background_mgr.go
@@ -86,12 +86,12 @@ type BackgroundManagerStatus struct {
 // Examples of this: ReSync, Compaction
 type BackgroundManagerProcessI interface {
 	Init(options map[string]interface{}, clusterStatus []byte) error
-	Run(options map[string]interface{}, persistClusterStatusCallback updateStatusClusterAwareCallback, terminator chan struct{}) error
+	Run(options map[string]interface{}, persistClusterStatusCallback updateStatusCallbackFunc, terminator chan struct{}) error
 	GetProcessStatus(status BackgroundManagerStatus) ([]byte, error)
 	ResetStatus()
 }
 
-type updateStatusClusterAwareCallback func() error
+type updateStatusCallbackFunc func() error
 
 func (b *BackgroundManager) Start(options map[string]interface{}) error {
 	err := b.markStart()
@@ -100,7 +100,7 @@ func (b *BackgroundManager) Start(options map[string]interface{}) error {
 	}
 
 	var processClusterStatus []byte
-	if b.clusterAwareOptions != nil {
+	if b.isClusterAware() {
 		processClusterStatus, _, err = b.clusterAwareOptions.bucket.GetRaw(b.clusterAwareOptions.StatusDocID())
 		if err != nil && !base.IsDocNotFoundError(err) {
 			return pkgerrors.Wrap(err, "Failed to get current process status")
@@ -114,7 +114,7 @@ func (b *BackgroundManager) Start(options map[string]interface{}) error {
 		return err
 	}
 
-	if b.clusterAwareOptions != nil {
+	if b.isClusterAware() {
 		go func() {
 			ticker := time.NewTicker(BackgroundManagerStatusUpdateIntervalSecs * time.Second)
 			for {
@@ -134,10 +134,7 @@ func (b *BackgroundManager) Start(options map[string]interface{}) error {
 
 	go func() {
 		updateStatusClusterAwareCallback := func() error {
-			if b.clusterAwareOptions != nil {
-				return b.UpdateHeartbeatDocClusterAware()
-			}
-			return nil
+			return b.UpdateStatusClusterAware()
 		}
 		err := b.Process.Run(options, updateStatusClusterAwareCallback, b.terminator)
 		if err != nil {
@@ -157,7 +154,7 @@ func (b *BackgroundManager) Start(options map[string]interface{}) error {
 
 		// Once our background process run has completed we should update the completed status and delete the heartbeat
 		// doc
-		if b.clusterAwareOptions != nil {
+		if b.isClusterAware() {
 			err = b.UpdateStatusClusterAware()
 			if err != nil {
 				base.Warnf("Failed to update background manager status: %v", err)
@@ -169,7 +166,7 @@ func (b *BackgroundManager) Start(options map[string]interface{}) error {
 		}
 	}()
 
-	if b.clusterAwareOptions != nil {
+	if b.isClusterAware() {
 		err = b.UpdateStatusClusterAware()
 		if err != nil {
 			base.Errorf("Failed to update background manager status: %v", err)
@@ -186,7 +183,7 @@ func (b *BackgroundManager) markStart() error {
 	processAlreadyRunningErr := base.HTTPErrorf(http.StatusServiceUnavailable, "Process already running")
 
 	// If we're running in cluster aware 'mode' base the check off of a heartbeat doc
-	if b.clusterAwareOptions != nil {
+	if b.isClusterAware() {
 		_, err := b.clusterAwareOptions.bucket.WriteCas(b.clusterAwareOptions.HeartbeatDocID(), 0, BackgroundManagerHeartbeatExpirySecs, 0, []byte("{}"), sgbucket.Raw)
 		if base.IsCasMismatch(err) {
 			return processAlreadyRunningErr
@@ -235,7 +232,7 @@ func (b *BackgroundManager) markStart() error {
 }
 
 func (b *BackgroundManager) GetStatus() ([]byte, error) {
-	if b.clusterAwareOptions != nil {
+	if b.isClusterAware() {
 		status, err := b.getStatusFromCluster()
 		if err != nil {
 			return nil, err
@@ -345,7 +342,7 @@ func (b *BackgroundManager) markStop() error {
 
 	processAlreadyStoppedErr := base.HTTPErrorf(http.StatusServiceUnavailable, "Process already stopped")
 
-	if b.clusterAwareOptions != nil {
+	if b.isClusterAware() {
 		_, _, err := b.clusterAwareOptions.bucket.GetRaw(b.clusterAwareOptions.HeartbeatDocID())
 		if err != nil {
 			if base.IsDocNotFoundError(err) {
@@ -405,6 +402,9 @@ func (b *BackgroundManager) SetError(err error) {
 // UpdateStatusClusterAware gets the current local status from the running process and updates the status document in
 // the bucket. Implements a retry. Used for Cluster Aware operations
 func (b *BackgroundManager) UpdateStatusClusterAware() error {
+	if b.clusterAwareOptions == nil {
+		return nil
+	}
 	err, _ := base.RetryLoop("UpdateStatusClusterAware", func() (shouldRetry bool, err error, value interface{}) {
 		status, err := b.getStatusLocal()
 		if err != nil {
@@ -463,6 +463,10 @@ func (b *BackgroundManager) UpdateHeartbeatDocClusterAware() error {
 	return nil
 }
 
+func (b *BackgroundManager) isClusterAware() bool {
+	return b.clusterAwareOptions != nil
+}
+
 // ======================================================
 // Resync Implementation of Background Manager Process
 // ======================================================
@@ -486,7 +490,7 @@ func (r *ResyncManager) Init(options map[string]interface{}, clusterStatus []byt
 	return nil
 }
 
-func (r *ResyncManager) Run(options map[string]interface{}, persistClusterStatusCallback updateStatusClusterAwareCallback, terminator chan struct{}) error {
+func (r *ResyncManager) Run(options map[string]interface{}, persistClusterStatusCallback updateStatusCallbackFunc, terminator chan struct{}) error {
 	database := options["database"].(*Database)
 	regenerateSequences := options["regenerateSequences"].(bool)
 
@@ -554,7 +558,7 @@ func (t *TombstoneCompactionManager) Init(options map[string]interface{}, cluste
 	return nil
 }
 
-func (t *TombstoneCompactionManager) Run(options map[string]interface{}, persistClusterStatusCallback updateStatusClusterAwareCallback, terminator chan struct{}) error {
+func (t *TombstoneCompactionManager) Run(options map[string]interface{}, persistClusterStatusCallback updateStatusCallbackFunc, terminator chan struct{}) error {
 	database := options["database"].(*Database)
 
 	defer atomic.CompareAndSwapUint32(&database.CompactState, DBCompactRunning, DBCompactNotRunning)
@@ -655,7 +659,7 @@ func (a *AttachmentCompactionManager) Init(options map[string]interface{}, clust
 	return newRunInit()
 }
 
-func (a *AttachmentCompactionManager) Run(options map[string]interface{}, persistClusterStatusCallback updateStatusClusterAwareCallback, terminator chan struct{}) error {
+func (a *AttachmentCompactionManager) Run(options map[string]interface{}, persistClusterStatusCallback updateStatusCallbackFunc, terminator chan struct{}) error {
 	database := options["database"].(*Database)
 
 	persistClusterStatus := func() {
--- a/db/background_mgr.go
+++ b/db/background_mgr.go
@@ -141,11 +141,13 @@ func (b *BackgroundManager) Start(options map[string]interface{}) error {
 
 		b.Terminate()
 
-		if b.GetRunState() == BackgroundProcessStateStopping {
-			b.setRunState(BackgroundProcessStateStopped)
-		} else if b.GetRunState() != BackgroundProcessStateError {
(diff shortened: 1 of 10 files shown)


db/background_mgr.go
```diff
@@ -12,17 +12,26 @@ import (
...
 	"github.com/google/uuid"
 )
 
 type BackgroundProcessState string
 
+// These states are used for background tasks
+// Running = The process is currently doing work
+// Stopped = The process stopped following completion of its task
+// Aborting = A user has requested that the process be stopped and therefore will stop shortly (usually after completing its current 'iteration')
+// Aborted = The process has stopped either by user request or the process 'crashed' midway through --> Essentially means the process is not running and the previous run had not completed
```
adamcfraser: If it's necessary to differentiate between a requested stop and completion, I'd prefer to leave stopped and stopping as-is, and have a 'completed' state for processes that stopped following completion of its task.

db/background_mgr.go
```diff
@@ -124,23 +298,32 @@ func (b *BackgroundManager) Stop() error {
...
 
-	b.terminator <- struct{}{}
+	b.Terminate()
 	return nil
 }
 
+// Terminate stops the process via terminator channel
+// Only to be used internally to this file and by tests.
+func (b *BackgroundManager) Terminate() {
+	if !b.terminatorClosed {
```
adamcfraser: Since there's no synchronization here, this doesn't look like it's going to prevent the channel being closed twice (if that's the intention).

Can we avoid the extra terminatorClosed by setting the status (to stopping) here w/ a compareAndSwap, and only closing the terminator if the state isn't already stopping?

db/background_mgr.go
```diff
@@ -52,47 +76,160 @@ type BackgroundManagerStatus struct {
...
 	err := b.markStart()
 	if err != nil {
 		return err
 	}
 
+	var processClusterStatus []byte
+	if b.clusterAwareOptions != nil {
+		processClusterStatus, _, err = b.clusterAwareOptions.bucket.GetRaw(b.clusterAwareOptions.StatusDocID())
+		if err != nil && !base.IsDocNotFoundError(err) {
+			return err
```
adamcfraser: If GetRaw returns an error, should we wrap the error to make it more meaningful to the caller of Start?  Want to make sure the error is meaningful in case it's getting passed all the way back to the REST API response.

db/background_mgr.go
```diff
@@ -52,47 +76,160 @@ type BackgroundManagerStatus struct {
...
 // Examples of this: ReSync, Compaction
 type BackgroundManagerProcessI interface {
+	Init(clusterStatus []byte) error
 	Run(options map[string]interface{}, terminator chan struct{}) error
 	GetProcessStatus(status BackgroundManagerStatus) ([]byte, error)
 	ResetStatus()
 }
 
 func (b *BackgroundManager) Start(options map[string]interface{}) error {
+	b.terminator = make(chan struct{}, 1)
```
adamcfraser: Start seems to be called based on a db-scoped BackgroundManager instance in some cases (db.AttachmentCompactionManager, db.ResyncManager).  How are we guarding against it being called more than once?

It looks like db.TombstoneCompactionManager handles that externally (with db.CompactState), but it would make more sense for the BackgroundManagers to handle, I think?  BackgroundProcessState seems like it might be a reasonable way to manage that.

db/background_mgr.go
```diff
@@ -52,47 +76,160 @@ type BackgroundManagerStatus struct {
...
+		processClusterStatus, _, err = b.clusterAwareOptions.bucket.GetRaw(b.clusterAwareOptions.StatusDocID())
+		if err != nil && !base.IsDocNotFoundError(err) {
+			return err
+		}
+	}
+
 	b.resetStatus()
+
+	err = b.Process.Init(processClusterStatus)
+	if err != nil {
```
adamcfraser, line 113: Init returns error if it fails to unmarshal the processClusterStatus.  In the event of malformed processClusterStatus, how does the user recover?

db/background_mgr.go
```diff
@@ -52,47 +76,160 @@ type BackgroundManagerStatus struct {
...
+		go func() {
+			ticker := time.NewTicker(time.Second)
+			for {
+				select {
+				case <-ticker.C:
+					// Implement a retry on here. It's not the end of the world if we miss a couple of these, but we
+					// want to give it a best attempt
+					err = b.UpdateStatusClusterAware()
+					if err != nil {
+						base.Errorf("Failed to update background manager status: %v", err)
```
adamcfraser: Errorf is usually reserved for severe errors that the user should act on.  This doesn't seem like the case here - warning seems sufficient, as the only loss of functionality here is the ability to resume.

db/background_mgr.go
```diff
@@ -52,47 +76,160 @@ type BackgroundManagerStatus struct {
...
+		} else if b.GetRunState() != BackgroundProcessStateError {
+			b.setRunState(BackgroundProcessStateStopped)
+		}
+
+		// Once our background process run has completed we should update the completed status and delete the heartbeat
+		// doc
+		if b.clusterAwareOptions != nil {
+			err = b.UpdateStatusClusterAware()
+			if err != nil {
+				base.Errorf("Failed to update background manager status: %v", err)
```
adamcfraser: Same as above on Warn vs Error

db/background_mgr.go
```diff
@@ -52,47 +76,160 @@ type BackgroundManagerStatus struct {
...
+			}
+
+			b.Terminate()
```
adamcfraser: Should Terminate be inside the clusterAwareOptions check like this?

db/background_mgr.go
```diff
@@ -164,6 +347,40 @@ func (b *BackgroundManager) SetError(err error) {
...
+
+		return false, nil, nil
+	}, base.CreateSleeperFunc(5, 100))
+	return err
+}
+
+// UpdateHeartbeatDocClusterAware simply performs a touch operation on the heartbeat document to update its expiry.
+// Implements a retry. Used for Cluster Aware operations
+func (b *BackgroundManager) UpdateHeartbeatDocClusterAware() error {
+	err, _ := base.RetryLoop("", func() (shouldRetry bool, err error, value interface{}) {
```
adamcfraser: I don't think we'd want retry when touching the heartbeat - waiting for the next interval is usually sufficient.

db/background_mgr.go
```diff
@@ -164,6 +347,40 @@ func (b *BackgroundManager) SetError(err error) {
...
+		_, err = b.clusterAwareOptions.bucket.Touch(b.clusterAwareOptions.HeartbeatDocID(), 30)
```
adamcfraser: 30 seconds seems like a long time to detect node failure, given that we're attempting to perform a touch every second.  What's the rationale here?

adamcfraser (changes requested):
The lack of synchronization around the background manager state leaves me feeling like there are potential issues for concurrent requests.  Have a look at the comments, and if I'm missing something let me know.

db/background_mgr.go
JRascagneres: Wouldn't having a completed state be confusing if for example attachment compaction has a 'completed' state but resync just said 'stopped'. Or do you mean have a separate 'completed' field?

db/background_mgr.go
JRascagneres: TombstoneCompactionManager does handle it externally but that can be removed now really. The markStart() function handles this. Its safe due a lock it holds. Need to  move terminator tho.

db/background_mgr.go
JRascagneres: Okay although its fairly important to know if that doc is missing at any point as if that doc disappears for whatever reason it would allow another instance of the process to start up. Not sure how we can avoid that if we just wait for the next interval.

db/background_mgr.go
JRascagneres: It was for the case where we allow a failure to be ignored until the next interval but with the retry this can be lowered.

db/background_mgr.go
JRascagneres: Yeah thats true. Can probably just act as if there was no cluster status and restart rather than resume the process.

db/background_mgr.go
```diff
@@ -337,12 +337,31 @@ func (b *BackgroundManager) markStop() error {
...
+	if b.clusterAwareOptions != nil {
+		_, _, err := b.clusterAwareOptions.bucket.GetRaw(b.clusterAwareOptions.HeartbeatDocID())
+		if err != nil {
+			if base.IsDocNotFoundError(err) {
+				return processAlreadyStoppedErr
+			}
+			return base.HTTPErrorf(http.StatusInternalServerError, "Unable to verify whether a process is running: %v", err)
+		}
+
+		err = b.clusterAwareOptions.bucket.Set(b.clusterAwareOptions.HeartbeatDocID(), BackgroundManagerHeartbeatExpirySecs, HeartbeatDoc{ShouldStop: true})
```
adamcfraser, line 354: Are we setting the expiry here to handle race conditions where the node running compact completes and removes the heartbeat doc before this heartbeat is written?

db/background_mgr.go
```diff
@@ -406,13 +429,23 @@ func (b *BackgroundManager) UpdateHeartbeatDocClusterAware() error {
...
 
-	b.clusterAwareOptions.lastSuccessfulHeartbeat = time.Now()
+	var status HeartbeatDoc
+	err = base.JSONUnmarshal(statusRaw, &status)
+	if err != nil {
+		return err
+	}
+
+	if status.ShouldStop {
+		_ = b.Stop()
```
adamcfraser: Why is it safe to ignore error here?

db/background_mgr.go
```diff
@@ -290,43 +586,103 @@ func (t *TombstoneCompactionManager) ResetStatus() {
...
 
-func (a *AttachmentCompactionManager) Run(options map[string]interface{}, terminator chan struct{}) error {
-	database := options["database"].(*Database)
+func (a *AttachmentCompactionManager) Init(clusterStatus []byte) error {
+	if clusterStatus != nil {
+		var attachmentResponseStatus AttachmentManagerResponse
+		err := base.JSONUnmarshal(clusterStatus, &attachmentResponseStatus)
+
+		// If the previous run completed, or there was an error during unmarshalling the status we will start the
+		// process from scratch with a new compaction ID. Otherwise, we should resume with the compact ID and phase
```
adamcfraser, line 645: Do we have a way to force a restart from scratch if either the compact was manually stopped, or the node failed?

db/background_mgr.go
```diff
@@ -290,43 +586,103 @@ func (t *TombstoneCompactionManager) ResetStatus() {
...
+	switch a.Phase {
+	case "mark", "":
+		a.SetPhase("mark")
+		_, err := Mark(database, a.CompactID, terminator, &a.MarkedAttachments)
+		if err != nil {
+			return err
+		}
+		fallthrough
+	case "sweep":
+		a.SetPhase("sweep")
```
adamcfraser, line 684: Do we need to force status persistence when we SetPhase, in case sweep does some work then fails before the next status update? (same question for cleanup below)

db/background_mgr.go
```diff
@@ -63,42 +97,163 @@ func (b *BackgroundManager) Start(options map[string]interface{}) error {
...
+		return err
+	}
+
+	if b.clusterAwareOptions != nil {
+		go func() {
+			ticker := time.NewTicker(BackgroundManagerStatusUpdateIntervalSecs * time.Second)
+			for {
+				select {
+				case <-ticker.C:
+					// Implement a retry on here. It's not the end of the world if we miss a couple of these, but we
```
adamcfraser: We now have retry down in UpdateStatucClusterAware - can we remove this comment?

adamcfraser (changes requested):

db/background_mgr.go
JRascagneres: Yeah, I'd missed that. Added it in now with some logging so we can tell whether it was force restarted / started / resumed.

db/background_mgr.go
adamcfraser: Can just add logging here

db/background_mgr.go
JRascagneres: Yes. As discussed but basically: Want to be super safe anywhere we write this doc as this doc sticking around would end up blocking any other process from running, so need to make sure this doc gets deleted if the process fails / stops / crashes etc.

db/background_mgr.go
JRascagneres: Yeah. As discussed added callback to `Run` which will allow the process to force cluster status persistence

db/background_mgr.go
```diff
@@ -52,53 +85,181 @@ type BackgroundManagerStatus struct {
...
+			}
+		}()
+	}
+
 	go func() {
-		defer b.setRunState(BackgroundProcessStateStopped)
-		err := b.Process.Run(options, b.terminator)
+		updateStatusClusterAwareCallback := func() error {
+			if b.clusterAwareOptions != nil {
+				return b.UpdateHeartbeatDocClusterAware()
```
adamcfraser: Shouldn't this be UpdateStatusClusterAware()?  Assuming yes, can we move the b.clusterAwareOptions check down into that function, and just pass it directly instead of having to wrap with this callback?

db/background_mgr.go
```diff
@@ -52,53 +85,181 @@ type BackgroundManagerStatus struct {
...
 // Examples of this: ReSync, Compaction
 type BackgroundManagerProcessI interface {
-	Run(options map[string]interface{}, terminator chan struct{}) error
+	Init(options map[string]interface{}, clusterStatus []byte) error
+	Run(options map[string]interface{}, persistClusterStatusCallback updateStatusClusterAwareCallback, terminator chan struct{}) error
 	GetProcessStatus(status BackgroundManagerStatus) ([]byte, error)
 	ResetStatus()
 }
 
+type updateStatusClusterAwareCallback func() error
```
adamcfraser: What does 'ClusterAware' signify in this naming?  Any reason it can't just be updateStatusCallbackFunc?

(elsewhere I think we follow the naming pattern of the Func suffix for function signature alias types)

db/background_mgr.go
```diff
@@ -52,53 +85,181 @@ type BackgroundManagerStatus struct {
...
+
 func (b *BackgroundManager) Start(options map[string]interface{}) error {
 	err := b.markStart()
 	if err != nil {
 		return err
 	}
 
+	var processClusterStatus []byte
+	if b.clusterAwareOptions != nil {
```
adamcfraser: We're checking if `b.clusterAwareOptions != nil' in a lot of places.  It wasn't immediately obvious to me that this is equivalent to checking whether the background manager is cluster aware or is just local-only.  

I suppose it's a nitpick, but either a b.isClusterAware() function that was is checking the Options value under the hood, or a b.isClusterAware bool on the struct would make this more clear.

adamcfraser (changes requested):

db/background_mgr.go
JRascagneres: Naming was just because it doesn't actually do anything in the non-cluster aware 'mode'. But I can drop that then I guess theoretically it could have other stuff added to it in the future if required.

db/background_mgr.go
JRascagneres: Added function to check those options.

adamcfraser (approved):

PR #5318, opened 2021-11-02
CBG-1764: Creating a new legacy database does not inherit credentials from a static database

IsaacLambat:
CBG-1764

- Credentials no longer inherited from any static databases
- unit test to test this behaviour
- cluster is closed on error to stop constant warning logs due to authentication failure
- adding database fails fast

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1​344

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -105,7 +105,7 @@ func (h *handler) handleCreateDB() error {
 		// store the cas in the loaded config after a successful insert
 		h.server.dbConfigs[dbName].cas = cas
 	} else {
-		// Intentionally do not pass bootstrap config in so it's credentials or server is not inherited (CBG-1764)
+		// Intentionally pass in an empty BootstrapConfig to avoid inheriting any credentials or server when running with a legacy config (CBG-1764)
 		if err := config.setup(dbName, BootstrapConfig{}, nil); err != nil {
 			return err
 		}
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -40,11 +40,6 @@ func (h *handler) handleCreateDB() error {
 	}
 	config.Name = dbName
 
-	bucket := dbName
-	if config.Bucket != nil {
-		bucket = *config.Bucket
-	}
-
 	if h.server.persistentConfig {
 		if err := config.validatePersistentDbConfig(); err != nil {
 			return base.HTTPErrorf(http.StatusBadRequest, err.Error())
@@ -57,6 +52,11 @@ func (h *handler) handleCreateDB() error {
 		version, err := GenerateDatabaseConfigVersionID("", config)
 		if err != nil {
 			return err
+		}
+
+		bucket := dbName
+		if config.Bucket != nil {
+			bucket = *config.Bucket
 		}
 
 		// copy config before setup to persist the raw config the user supplied
@@ -112,7 +112,7 @@ func (h *handler) handleCreateDB() error {
 		// load database in-memory for non-persistent nodes
 		if _, err := h.server.AddDatabaseFromConfigFailFast(DatabaseConfig{DbConfig: *config}); err != nil {
 			if errors.Is(err, base.ErrAuthError) {
-				return base.HTTPErrorf(http.StatusForbidden, "auth failure accessing provided bucket using database credentials: %s", base.Metadata(bucket))
+				return base.HTTPErrorf(http.StatusForbidden, "auth failure using provided bucket credentials for database %s", base.Metadata(config.Name))
 			}
 			return err
 		}
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -105,10 +105,6 @@ func (h *handler) handleCreateDB() error {
 		// store the cas in the loaded config after a successful insert
 		h.server.dbConfigs[dbName].cas = cas
 	} else {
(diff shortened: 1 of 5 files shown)


rest/admin_api.go
```diff
@@ -40,6 +40,11 @@ func (h *handler) handleCreateDB() error {
 	}
 	config.Name = dbName
 
+	bucket := dbName
+	if config.Bucket != nil {
+		bucket = *config.Bucket
+	}
+
```
JRascagneres: Whats the reason for moving this? 
It also seems a little misleading as we're not necessarily connecting to the `bucket` var in the non-persistent case

rest/admin_api.go
```diff
@@ -105,12 +105,15 @@ func (h *handler) handleCreateDB() error {
...
-		if err := config.setup(dbName, h.server.config.Bootstrap, nil); err != nil {
+		if err := config.setup(dbName, BootstrapConfig{}, nil); err != nil {
 			return err
 		}
 
 		// load database in-memory for non-persistent nodes
-		if _, err := h.server.AddDatabaseFromConfig(DatabaseConfig{DbConfig: *config}); err != nil {
+		if _, err := h.server.AddDatabaseFromConfigFailFast(DatabaseConfig{DbConfig: *config}); err != nil {
+			if errors.Is(err, base.ErrAuthError) {
+				return base.HTTPErrorf(http.StatusForbidden, "auth failure accessing provided bucket using database credentials: %s", base.Metadata(bucket))
```
JRascagneres: As discussed don't think we want `bucket` here given the possible change to collections in the future. Might be better to just tie wording to database. 

Even without that might want to move where the bucket name is. The error would be:
`auth failure accessing provided bucket using database credentials: bucketName` which doesn't make sense. At minimum something like:
`auth failure accessing bucket %s using provided credentials`.

But again, try to tie it to database I think

rest/admin_api_test.go
```diff
@@ -3705,3 +3705,42 @@ func TestConfigsIncludeDefaults(t *testing.T) {
 	runtimeServerConfigDatabase = runtimeServerConfigResponse.Databases["db2"]
 	assert.True(t, runtimeServerConfigDatabase.Unsupported.DisableCleanSkippedQuery)
 }
+
+func TestLegacyCredentialInheritance(t *testing.T) {
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("This test only works against Couchbase Server")
+	}
+	defer base.SetUpTestLogging(base.LevelInfo, base.KeyHTTP)()
```
JRascagneres, line 3714: Can probably move this now you have test working? Presume this was for debugging

JRascagneres (changes requested):
Mostly looks good. Couple minor comments

JRascagneres:
Just seen a bootstrap related failure on the integration tests so make sure that gets fixed too. I doubt its a one-off given it sounds related.

IsaacLambat:
> Just seen a bootstrap related failure on the integration tests so make sure that gets fixed too. I doubt its a one-off given it sounds related.

Can't seem to reproduce locally nor does any of the lines I have changed get hit, so will run new integration test and see if it works

EDIT: Was due to server context not being closed. Fixed now.

rest/admin_api_test.go
IsaacLambat: Seems like this is needed so logging can be teardown due to setupServerContext setting up logging

rest/admin_api.go
```diff
@@ -112,7 +112,7 @@ func (h *handler) handleCreateDB() error {
 		// load database in-memory for non-persistent nodes
 		if _, err := h.server.AddDatabaseFromConfigFailFast(DatabaseConfig{DbConfig: *config}); err != nil {
 			if errors.Is(err, base.ErrAuthError) {
-				return base.HTTPErrorf(http.StatusForbidden, "auth failure accessing provided bucket using database credentials: %s", base.Metadata(bucket))
+				return base.HTTPErrorf(http.StatusForbidden, "auth failure using provided bucket credentials for database %s", base.Metadata(config.Name))
```
JRascagneres: This should be base.MD() not Metadata. We don't use Metadata directly and always use MD(). It probably works in this case as its a string but MD() is used to transform variables which may not be strings.

JRascagneres (changes requested):
One v small change and then LGTM

rest/admin_api.go
IsaacLambat: Oops, didn't spot that. Thanks

JRascagneres (dismissed):
LGTM now

rest/admin_api.go
```diff
@@ -105,12 +105,15 @@ func (h *handler) handleCreateDB() error {
 		// store the cas in the loaded config after a successful insert
 		h.server.dbConfigs[dbName].cas = cas
 	} else {
-		if err := config.setup(dbName, h.server.config.Bootstrap, nil); err != nil {
+		if err := config.setup(dbName, BootstrapConfig{}, nil); err != nil {
```
adamcfraser, line 109: Can we add a comment here to clarify that we are intentionally not inheriting from the bootstrap when processing a legacy config.

JRascagneres (approved):

PR #5319, opened 2021-11-02
CBG-1766: Limit max concurrent query ops to be inline with max idle conns per host

JRascagneres:
CBG-1766

Changed max concurrent query ops to be limited to query node count * DefaultHttpMaxIdleConnsPerHost.
This ensures that we can properly handle a large number of concurrent query ops without having to constantly open new connections. 

Opted to limit node count to 1 if we either can't get the count or if the count is 0. Means we can still handle queries if for whatever there is an issue getting the node count or if they are still in startup for example. 

Given the low level operations here it's difficult to write a test for. However, I did manual testing for both collections and buckets and ensured that the setting was correctly limiting the max concurrent query ops in line with the number of nodes. Tested with 1, 2, 3 and 4 nodes.

### DIFF
--- a/base/bucket_gocb.go
+++ b/base/bucket_gocb.go
@@ -186,7 +186,7 @@ func GetCouchbaseBucketGoCBFromAuthenticatedCluster(cluster *gocb.Cluster, spec 
 
 	if maxConcurrentQueryOps > DefaultHttpMaxIdleConnsPerHost*queryNodeCount {
 		maxConcurrentQueryOps = DefaultHttpMaxIdleConnsPerHost * queryNodeCount
-		Warnf("Limited max concurrent query ops to %d to be in line with number of query nodes", maxConcurrentQueryOps)
+		Infof(KeyAll, "Setting max_concurrent_query_ops to %d based on query node count (%d)", maxConcurrentQueryOps, queryNodeCount)
 	}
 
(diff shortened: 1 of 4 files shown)


base/bucket_gocb.go
```diff
@@ -179,6 +179,16 @@ func GetCouchbaseBucketGoCBFromAuthenticatedCluster(cluster *gocb.Cluster, spec
...
 	}
 
+	queryNodeCount := len(goCBBucket.IoRouter().N1qlEps())
+	if queryNodeCount == 0 {
+		queryNodeCount = 1
+	}
+
+	if maxConcurrentQueryOps > DefaultHttpMaxIdleConnsPerHost*queryNodeCount {
+		maxConcurrentQueryOps = DefaultHttpMaxIdleConnsPerHost * queryNodeCount
+		Warnf("Limited max concurrent query ops to %d to be in line with number of query nodes", maxConcurrentQueryOps)
```
adamcfraser: Info logging is appropriate here, I think - and I might rephrase as 'Setting maxConcurrentQueryOps to %d based on query node count (%d)'

adamcfraser (commented):
Looks good - just a tweak to the log message when value is set.

adamcfraser (approved):

PR #5320, opened 2021-11-02
CBG-1779: Warning logged if root pool cannot be retrieved instead of erroring

IsaacLambat:
CBG-1779

- Error changes to warning when system root pool cannot be retrieved. 
- Log text changes from `Error getting root CAs` to `Could not retrieve root CAs`


JRascagneres (approved):
Seems reasonable to me.

Full warning log would be `Could not retrieve root CAs: crypto/x509: system root pool is not available on Windows`

PR #5321, opened 2021-11-03
CBG-1773: Fix race condition in attachment compaction stat callback

JRascagneres:
CBG-1773

Prior to this change there was a race in the attachment compaction stat callback. We were just sending through a pointer to the stat integer from the Mark / Sweep functions. Due to the fact that the DCP workers run in parallel there was a clear race in mutating this value.

- Added an AtomicInt type for this purpose
- Switched SgwIntStat to embed AtomicInt 

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1346/

### DIFF
--- a/db/attachment_compaction.go
+++ b/db/attachment_compaction.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	sgbucket "github.com/couchbase/sg-bucket"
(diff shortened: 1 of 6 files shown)


db/attachment_compaction.go
```diff
@@ -15,12 +16,11 @@ import (
 
 const CompactionIDKey = "compactID"
 
-func Mark(db *Database, compactionID string, terminator chan struct{}, markedAttachmentCallback func(markedAttachments *int)) (int, error) {
+func Mark(db *Database, compactionID string, terminator chan struct{}, markedAttachmentCountPtr *uint64) (uint64, error) {
```
adamcfraser: On the blip side we're using SgwIntStat to encapsulate synchronized stat handling, even when those stats aren't wired up to the metrics endpoint.  What do you think about following that approach?  It would be nice to have a standard way to share stats between goroutines that we use across the product.

adamcfraser (changes requested):

db/attachment_compaction.go
JRascagneres: I'm not sure if I'm a fan of that really. SgwIntStat is built around its use as a metric on the endpoint. Really you should initialize that with `NewIntStat` which has items we don't want in this case. I think it could be used by just initializing as `SgwIntStat{}` but I think that has the possibility to misuse it a little.
I accept that maybe passing *int into this function could be obscuring its use a little though, but I tried to minimize that in the parameter naming.

JRascagneres: Add an AtomicInt type, maybe embed in stats implementation.

adamcfraser (approved):

PR #5323, opened 2021-11-05
CBG-1740: Fix create db admin auth when user has access to only provided bucket

JRascagneres:
CBG-1740

- Make a new `makeHandler` - `makeHandlerSpecificAuthScope` which takes a callback function. This callback will be provided with the payload body and should return the authScope.
- This allows in `handleCreateDB` to be given a function which will open up the payload and provide the supplied bucket as the auth scope.

Currently when reading the payload body this clears the body that the handler may need and therefore we repopulate this with a new `io.ReadCloser`. A future enhancement may be to read this earlier (for all payloads) and save a bytes body on the handler for future use, however, to minimize the scope of this change I opted not to do that. 

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1355/

### DIFF
--- a/rest/admin_api_auth_test.go
+++ b/rest/admin_api_auth_test.go
@@ -9,7 +9,6 @@ import (
 	"strconv"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/couchbase/sync_gateway/base"
 	"github.com/stretchr/testify/assert"
@@ -1541,8 +1540,6 @@ func TestCreateDBSpecificBucketPerm(t *testing.T) {
 	mobileSyncGateway := "mobile_sync_gateway"
 	MakeUser(t, httpClient, eps[0], mobileSyncGateway, "password", []string{fmt.Sprintf("%s[%s]", mobileSyncGateway, tb.GetName())})
 	defer DeleteUser(t, httpClient, eps[0], mobileSyncGateway)
-
-	time.Sleep(2 * time.Second)
 
 	resp := rt.SendAdminRequestWithAuth("PUT", "/db2/", `{"bucket": "`+tb.GetName()+`", "username": "`+base.TestClusterUsername()+`", "password": "`+base.TestClusterPassword()+`", "num_index_replicas": 0, "use_views": `+strconv.FormatBool(base.TestsDisableGSI())+`}`, mobileSyncGateway, "password")
 	assertStatus(t, resp, http.StatusCreated)
--- a/rest/admin_api_auth_test.go
+++ b/rest/admin_api_auth_test.go
@@ -6,8 +6,10 @@ import (
 	"net/http"
 	"net/url"
 	"reflect"
+	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/couchbase/sync_gateway/base"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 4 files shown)


rest/admin_api_auth_test.go
```diff
@@ -1515,3 +1517,33 @@ func TestNewlyCreateSGWPermissions(t *testing.T) {
...
+	defer tb.Close()
+
+	eps, httpClient, err := rt.ServerContext().ObtainManagementEndpointsAndHTTPClient()
+	require.NoError(t, err)
+
+	mobileSyncGateway := "mobile_sync_gateway"
+	MakeUser(t, httpClient, eps[0], mobileSyncGateway, "password", []string{fmt.Sprintf("%s[%s]", mobileSyncGateway, tb.GetName())})
+	defer DeleteUser(t, httpClient, eps[0], mobileSyncGateway)
+
+	time.Sleep(2 * time.Second)
```
JRascagneres: remove this

rest/handler.go
```diff
@@ -134,6 +135,20 @@ func makeOfflineHandler(server *ServerContext, privs handlerPrivs, accessPermiss
 	})
 }
 
+// Create a http.Handler that will run a handler with the given method. It also takes in a callback function which when
+// given the endpoint payload returns an auth scope. Currently, used for create db where we need to read bucket from the
```
adamcfraser: I think you can remove the 'Currently' comment - it's straightforward to find usage.

I think it's probably clearer to just call this makeAuthScopeHandler as well.

rest/handler.go
```diff
@@ -99,6 +99,7 @@ type handler struct {
 	runOffline            bool
 	queryValues           url.Values // Copy of results of rq.URL.Query()
 	permissionsResults    map[string]bool
+	dbAuthStringFunc      func(bodyJSON []byte) (string, error)
```
adamcfraser: I feel like this would be better as just 'authScopeFunc' - omit the db, and make it clear it's expected to return an authScope string.

I also prefer to define an alias for for function signature - it doesn't make that much of a difference here, but it's a lot clearer to read when being used as a parameter (as below).

rest/handler.go
```diff
@@ -112,7 +113,7 @@ const (
 
 type handlerMethod func(*handler) error
 
-// Creates an http.Handler that will run a handler with the given method
+// Creates a http.Handler that will run a handler with the given method
```
adamcfraser: I'm going to say that whoever writes the comment in the first place gets to set the terms of the 'aitch vs. haitch' debate, so these should stay as 'an'.

rest/handler.go
```diff
@@ -275,6 +290,23 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
...
+			}
+			// The above readBody() will end up clearing the body which the later handler will require. Re-populate this
+			// for the later handler.
+			h.requestBody = ioutil.NopCloser(bytes.NewReader(body))
+			authScope, err = h.dbAuthStringFunc(body)
+			if err != nil {
+				return base.HTTPErrorf(http.StatusInternalServerError, "Unable to read body: %v", err)
+			}
+			if authScope == "" {
+				return base.HTTPErrorf(http.StatusBadRequest, "Must supply bucket in database config")
```
adamcfraser: This error message looks like it should be more generic.

rest/routing.go
```diff
@@ -293,7 +295,18 @@ func CreateAdminRouter(sc *ServerContext) *mux.Router {
 	// The routes below are part of the CouchDB REST API but should only be available to admins,
 	// so the handlers are moved to the admin port.
 	r.Handle("/{newdb:"+dbRegex+"}/",
-		makeHandler(sc, adminPrivs, []Permission{PermCreateDb}, nil, (*handler).handleCreateDB)).Methods("PUT")
+		makeHandlerSpecificAuthScope(sc, adminPrivs, []Permission{PermCreateDb}, nil, (*handler).handleCreateDB, func(bodyJSON []byte) (string, error) {
```
adamcfraser: I think I'd prefer to define the auth scope function elsewhere (in admin_api beside handleCreateDB, maybe?), to keep router.go more readable.

rest/routing.go
```diff
@@ -293,7 +295,18 @@ func CreateAdminRouter(sc *ServerContext) *mux.Router {
...
+			var body db.Body
```
adamcfraser: Any reason to use db.Body here, as opposed to a new struct that just has "bucket" defined?

adamcfraser (changes requested):
Approach looks good, several style nitpicks.

rest/handler.go
JRascagneres: Okay 😛 Goland said they were all wrong so just went with tweaking it

rest/routing.go
JRascagneres: Was just quick and dirty. Was going to change that on Monday after approval of basic idea.

adamcfraser (approved):

PR #5324, opened 2021-11-09
CBG-1785 - Return specific error when multiple databases target same bucket

bbrks:
CBG-1785

When creating databases, we should check that the bucket is not used by any other databases first, and return a specific error when we encounter this unsupported scenario, instead of a generic/misleading error message.

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1374/

### DIFF
--- a/rest/bootstrap_test.go
+++ b/rest/bootstrap_test.go
@@ -47,7 +47,10 @@ func TestBootstrapRESTAPISetup(t *testing.T) {
 		tb.Close()
 	}()
 	resp := bootstrapAdminRequest(t, http.MethodPut, "/db1/",
-		`{"bucket": "`+tb.GetName()+`", "num_index_replicas": 0}`,
+		fmt.Sprintf(
+			`{"bucket": "%s", "num_index_replicas": 0, "enable_shared_bucket_access": %t, "use_views": %t}`,
+			tb.GetName(), base.TestUseXattrs(), base.TestsDisableGSI(),
+		),
 	)
 	assert.Equal(t, http.StatusCreated, resp.StatusCode)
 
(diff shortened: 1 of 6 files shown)


rest/bootstrap_test.go
```diff
@@ -131,6 +131,46 @@ func TestBootstrapRESTAPISetup(t *testing.T) {
...
+	go func() {
+		serverErr <- startServer(&config, sc)
+	}()
+	require.NoError(t, sc.waitForRESTAPIs())
+
+	// Get a test bucket, and use it to create the database.
+	tb := base.GetTestBucket(t)
+	defer func() { tb.Close() }()
+	resp := bootstrapAdminRequest(t, http.MethodPut, "/db1/",
+		`{"bucket": "`+tb.GetName()+`", "num_index_replicas": 0}`,
```
JRascagneres: Would be nice to add the use views / xattrs config options based on the test vars

rest/bootstrap_test.go
```diff
@@ -131,6 +131,46 @@ func TestBootstrapRESTAPISetup(t *testing.T) {
...
+	)
+	assert.Equal(t, http.StatusCreated, resp.StatusCode)
+
+	// Create db2 using the same bucket and expect it to fail
+	resp = bootstrapAdminRequest(t, http.MethodPut, "/db2/",
+		`{"bucket": "`+tb.GetName()+`", "num_index_replicas": 0}`,
```
JRascagneres: as above

JRascagneres (changes requested):
Minor test comment but besides that LGTM pending integration test passing

JRascagneres (approved):

PR #5325, opened 2021-11-10
CBG-1763: Namespaced SG cluster metadata by group ID

IsaacLambat:
CBG-1763

- If Group ID is set, it is used to namespace anything that uses `CfgSG` to store config settings (such as SG Replicate), namespace the sync function, namespace the heartbeat documents, and namespaced DCP Checkpoints for the import feed
- If group ID is not set, then the old document names are used (IE. not putting `default` in the doc names)
- Modified tests to account for these changed

## Dependencies 
- [x] https://github.com/couchbase/sg-bucket/pull/68
- [x] Add commit hash to manifest

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1460

### DIFF
--- a/base/constants.go
+++ b/base/constants.go
@@ -198,7 +198,7 @@ func DCPCheckpointPrefixWithGroupID(groupID string) string {
 
 func SGCfgPrefixWithGroupID(groupID string) string {
 	if groupID != "" {
-		return SyncPrefix + groupID + ":cfg:"
+		return SGCfgPrefix + ":" + groupID + ":"
 	}
 	return SGCfgPrefix
 }
--- a/base/constants.go
+++ b/base/constants.go
@@ -114,14 +114,12 @@ const (
 	Att2Prefix             = SyncPrefix + "att2:"
 	BackfillCompletePrefix = SyncPrefix + "backfill:complete:"
 	BackfillPendingPrefix  = SyncPrefix + "backfill:pending:"
-	DCPCheckpointPrefix    = SyncPrefix + "dcp_ck:"
 	RepairBackup           = SyncPrefix + "repair:backup:"
 	RepairDryRun           = SyncPrefix + "repair:dryrun:"
 	RevBodyPrefix          = SyncPrefix + "rb:"
 	RevPrefix              = SyncPrefix + "rev:"
 	RolePrefix             = SyncPrefix + "role:"
 	SessionPrefix          = SyncPrefix + "session:"
-	SGCfgPrefix            = SyncPrefix + "cfg"
 	SyncSeqPrefix          = SyncPrefix + "seq:"
 	UserEmailPrefix        = SyncPrefix + "useremail:"
 	UserPrefix             = SyncPrefix + "user:"
@@ -129,7 +127,6 @@ const (
 	UnusedSeqRangePrefix   = SyncPrefix + "unusedSeqs:"
 
 	DCPBackfillSeqKey = SyncPrefix + "dcp_backfill"
-	SyncDataKey       = SyncPrefix + "syncdata"
 	SyncSeqKey        = SyncPrefix + "seq"
 
 	PersistentConfigPrefix = SyncPrefix + "dbconfig:"
@@ -188,6 +185,27 @@ var (
 	// ErrUnknownField is marked as the cause of the error when trying to decode a JSON snippet with unknown fields
 	ErrUnknownField = errors.New("unrecognized JSON field")
 )
+
+func DCPCheckpointPrefix(groupID string) string {
+	if groupID != "" {
+		return SyncPrefix + groupID + ":dcp_ck:"
+	}
+	return SyncPrefix + "dcp_ck:"
+}
+
+func SGCfgPrefix(groupID string) string {
+	if groupID != "" {
+		return SyncPrefix + groupID + ":cfg:"
+	}
+	return SyncPrefix + "cfg"
+}
+
+func SyncDataKey(groupID string) string {
+	if groupID != "" {
+		return SyncPrefix + groupID + ":syncdata"
+	}
+	return SyncPrefix + "syncdata"
+}
 
 // UnitTestUrl returns the configured test URL.
 func UnitTestUrl() string {
--- a/base/constants.go
+++ b/base/constants.go
@@ -188,7 +188,7 @@ var (
 
 func DCPCheckpointPrefix(groupID string) string {
 	if groupID != "" {
-		return SyncPrefix + groupID + ":dcp_ck:"
+		return SyncPrefix + "dcp_ck:" + groupID + ":"
 	}
 	return SyncPrefix + "dcp_ck:"
 }
--- a/base/constants.go
+++ b/base/constants.go
@@ -121,6 +121,7 @@ const (
 	RevPrefix              = SyncPrefix + "rev:"
 	RolePrefix             = SyncPrefix + "role:"
 	SessionPrefix          = SyncPrefix + "session:"
+	SGCfgPrefix            = SyncPrefix + "cfg"
 	SyncSeqPrefix          = SyncPrefix + "seq:"
 	UserEmailPrefix        = SyncPrefix + "useremail:"
 	UserPrefix             = SyncPrefix + "user:"
@@ -195,11 +196,11 @@ func DCPCheckpointPrefixWithGroupID(groupID string) string {
 	return DCPCheckpointPrefix
 }
 
-func SGCfgPrefix(groupID string) string {
+func SGCfgPrefixWithGroupID(groupID string) string {
 	if groupID != "" {
 		return SyncPrefix + groupID + ":cfg:"
 	}
-	return SyncPrefix + "cfg"
+	return SGCfgPrefix
 }
 
 func SyncDataKeyWithGroupID(groupID string) string {
--- a/base/constants.go
+++ b/base/constants.go
@@ -114,6 +114,7 @@ const (
 	Att2Prefix             = SyncPrefix + "att2:"
 	BackfillCompletePrefix = SyncPrefix + "backfill:complete:"
 	BackfillPendingPrefix  = SyncPrefix + "backfill:pending:"
+	DCPCheckpointPrefix    = SyncPrefix + "dcp_ck:"
 	RepairBackup           = SyncPrefix + "repair:backup:"
 	RepairDryRun           = SyncPrefix + "repair:dryrun:"
 	RevBodyPrefix          = SyncPrefix + "rb:"
@@ -127,6 +128,7 @@ const (
 	UnusedSeqRangePrefix   = SyncPrefix + "unusedSeqs:"
 
 	DCPBackfillSeqKey = SyncPrefix + "dcp_backfill"
+	SyncDataKey       = SyncPrefix + "syncdata"
 	SyncSeqKey        = SyncPrefix + "seq"
 
 	PersistentConfigPrefix = SyncPrefix + "dbconfig:"
@@ -186,11 +188,11 @@ var (
 	ErrUnknownField = errors.New("unrecognized JSON field")
 )
 
-func DCPCheckpointPrefix(groupID string) string {
+func DCPCheckpointPrefixWithGroupID(groupID string) string {
 	if groupID != "" {
-		return SyncPrefix + "dcp_ck:" + groupID + ":"
+		return DCPCheckpointPrefix + groupID + ":"
 	}
-	return SyncPrefix + "dcp_ck:"
+	return DCPCheckpointPrefix
 }
 
 func SGCfgPrefix(groupID string) string {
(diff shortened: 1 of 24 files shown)


base/heartbeat.go
```diff
@@ -65,8 +65,9 @@ type HeartbeatListener interface {
 //       e.g  Default for a 4 node cluster: 12 ops/second
 type couchbaseHeartBeater struct {
 	bucket                  Bucket
-	nodeUUID                string
 	keyPrefix               string
+	groupID                 *string
```
adamcfraser: *string is a relatively uncommon pattern in go, and I'm not sure it completely makes sense here.  We don't actually want to support cases where the string is changed once it's been passed to the heartbeater.  I don't think we support a configGroupID of empty string (do we?), so I'd prefer to just work with `string`.

base/heartbeat.go
```diff
@@ -83,12 +84,13 @@ type couchbaseHeartBeater struct {
 // the keyPrefix which will be prepended to the heartbeat doc keys,
 // and the nodeUUID, which is an opaque identifier for the "thing" that is using this
 // library.  nodeUUID will be passed to listeners on stale node detection.
-func NewCouchbaseHeartbeater(bucket Bucket, keyPrefix, nodeUUID string) (heartbeater *couchbaseHeartBeater, err error) {
+func NewCouchbaseHeartbeater(bucket Bucket, keyPrefix, nodeUUID string, groupID *string) (heartbeater *couchbaseHeartBeater, err error) {
```
adamcfraser: Looking at use cases like this one - where the functionality already accepts a key prefix - has me thinking about the relative merits of _sync:[groupID]:[functionalPrefix]:[value] versus _sync:[functionalPrefix]:[groupID]:[value], in particular for cases such as filtering a DCP stream.  The former would make it easier to broadly filter by config group, but a bit more difficult to filter by functional prefix.  Was there any discussion of the relative merits of each approach?

base/sg_cluster_cfg.go
```diff
@@ -51,19 +52,23 @@ func NewCfgSG(bucket Bucket) (*CfgSG, error) {
...
 		subscriptions: make(map[string][]chan<- cbgt.CfgEvent),
+		groupID:       groupID,
 	}
 	return c, nil
 }
 
-func sgCfgBucketKey(cfgKey string) string {
+func (c *CfgSG) sgCfgBucketKey(cfgKey string) string {
+	if c.groupID != nil {
+		return SGCfgPrefix + ":" + *c.groupID + ":" + cfgKey
```
adamcfraser: Don't we have a potential collision here between the `(groupID, cfgKey)` pairs `(nil, "a:b")` and `("a","b")`?  I don't think have any restrictions on cfgKey that would prevent `a:b` as a value.

base/sg_cluster_cfg.go
```diff
@@ -128,7 +133,10 @@ func (c *CfgSG) Subscribe(cfgKey string, ch chan cbgt.CfgEvent) error {
 }
 
 func (c *CfgSG) FireEvent(docID string, cas uint64, err error) {
-
+	if c.groupID != nil {
+		// Remove first instance of group ID
+		docID = strings.Replace(docID, *c.groupID, "", 1)
```
adamcfraser: It would be preferable to define the combined prefix at CfgSG initialization time and just use that with strings.TrimPrefix below, instead of doing multiple string operations.

adamcfraser (changes requested):
Posted a few comments.  In addition, this doesn't yet handle import feed checkpoints - those are persisted using DCPCheckpointPrefix, and don't use CfgSG.

base/heartbeat.go
IsaacLambat: There was a discussion on this, and the reason it was decided that it would be best to include the group ID after the functional prefix is because of the `DocChanged` function for the DCP filtering prefix checks.

However, due to the naming collision mention in the other comment, the group ID will be used before the functional prefix if a group ID is being used (or else the old legacy doc name will be used).

base/sg_cluster_cfg.go
IsaacLambat: This is a potential collision. @bbrks and I discussed this and decided the best solution that keeps backwards compatibility is to use a different constant prefix (that's created during runtime using the group ID) if a group ID is used. 

For example: (groupID, cfgKey) pairs ("a","b") would be `_sync:a:cfg:b` for CfgSG. Pairs (nil, "a:b") would be `_sync:cfga:b` which is the same as it was in legacy, while avoiding a collision. 

This will also be used in other places where namespacing is being added such as the sync function and heartbeater.

base/constants.go
```diff
@@ -189,6 +186,27 @@ var (
...
+	}
+	return SyncPrefix + "cfg"
+}
+
+func SyncDataKey(groupID string) string {
+	if groupID != "" {
+		return SyncPrefix + groupID + ":syncdata"
+	}
+	return SyncPrefix + "syncdata"
+}
```
bbrks: I think `_sync:dcp_ck:` can remain as a constant given group ID is appended as a suffix.

And `_sync:syncdata` could be as well if we moved the groupID portion to the end of they key.
I don't think there's a requirement for that key to be prefixed rather than suffixed.

base/constants.go
```diff
@@ -189,6 +186,27 @@ var (
 	ErrUnknownField = errors.New("unrecognized JSON field")
 )
 
+func DCPCheckpointPrefix(groupID string) string {
```
bbrks: Make constant `DCPCheckpointPrefix` and rename these functions to be GroupID specific.

bbrks (dismissed):

base/dcp_common.go
```diff
@@ -97,6 +98,7 @@ func NewDCPCommon(callback sgbucket.FeedEventCallbackFunc, bucket Bucket, maxVbN
 		lastCheckpointTime:     make([]time.Time, maxVbNo),
 		backfill:               &newBackfillStatus,
 		feedID:                 feedID,
+		groupID:                groupID,
```
adamcfraser: I'd prefer to store the computed prefix `DCPCheckpointPrefixWithGroupID(c.groupID)` here, both to avoid the string concatenation cost every time we read/write checkpoints, but also to make it explicit what DCPCommon is using groupID for.

base/constants.go
```diff
@@ -189,6 +189,27 @@ var (
...
+func DCPCheckpointPrefixWithGroupID(groupID string) string {
+	if groupID != "" {
+		return DCPCheckpointPrefix + groupID + ":"
+	}
+	return DCPCheckpointPrefix
+}
+
+func SGCfgPrefixWithGroupID(groupID string) string {
+	if groupID != "" {
+		return SyncPrefix + groupID + ":cfg:"
```
adamcfraser: Previously we didn't have a trailing colon on SGCfgPrefix.  This doesn't look backward compatible if the groupID is empty?

base/heartbeat.go
```diff
@@ -65,8 +65,9 @@ type HeartbeatListener interface {
 //       e.g  Default for a 4 node cluster: 12 ops/second
 type couchbaseHeartBeater struct {
 	bucket                  Bucket
-	nodeUUID                string
 	keyPrefix               string
+	groupID                 string
+	nodeUUID                string
```
adamcfraser: I'm not sure why nodeUUID is moving around here and below - better to avoid that kind of unnecessary diff.

base/sg_cluster_cfg.go
```diff
@@ -51,19 +53,20 @@ func NewCfgSG(bucket Bucket) (*CfgSG, error) {
 		bucket:        bucket,
 		loggingCtx:    loggingCtx,
 		subscriptions: make(map[string][]chan<- cbgt.CfgEvent),
+		groupID:       groupID,
```
adamcfraser: As in the previous case, would prefer to define the computed prefix here instead of groupID, unless there's another reason to store groupID.  In this case we're paying for the concatenation with cfgKey anyway, but it's still better to make it clear what groupID is needed for.

db/change_listener.go
```diff
@@ -59,6 +61,7 @@ func (listener *changeListener) Start(bucket base.Bucket, dbStats *expvar.Map) e
 		Backfill:   sgbucket.FeedNoBackfill,
 		Terminator: listener.terminator,
 		DoneChan:   make(chan struct{}),
+		GroupID:    listener.groupID,
```
adamcfraser: Same as above, regarding prefix storage vs. groupID.  DCP processing is particularly performance sensitive, definitely want to avoid the concatenation per DCP mutation if we can help it.

adamcfraser: I also don't think GroupID belongs in FeedArgs - it can just be a property on the listener and avoid the sgbucket change.

adamcfraser (changes requested):
A few places where I think we should try to compute the prefix once instead of per-op, and some other minor nits.

base/constants.go
IsaacLambat: If group ID is empty, the old `SGCfgPrefix` is returned, which is now a separate constant (see return below).

base/heartbeat.go
IsaacLambat: Moved it as thought it might be best to have it in the order that it is in the key. Will change back

db/change_listener.go
IsaacLambat: Passing down the prefixes to `StartDCPFeed` from the listener would require a sgbucket change to modify the KVStore interface `StartDCPFeed` function to add the extra parameters. This would cause a lot of refactoring for other parts of the code that implement this (including to the walrus repo) that don't use the group ID and could cause unintended problems.

I think it would be best to use the FeedArguments (as the prefixes are being used internally for DCP) but instead of group ID like it is currently, to use the pre calculated prefixes (like suggested on https://github.com/couchbase/sg-bucket/pull/68). The prefixes would the SG Cfg prefix (used for DCP key filtering) and DCP checkpoint prefix.

base/constants.go
adamcfraser: This format - `_sync:groupID:cfg:` doesn't match the discussion on the JIRA issue, which is: 

> Scope CBGT config by group ID as you have - (i.e. _sync:cfg:group1:foobar and fall back to _sync:cfgfoobar for default group)

If we followed the approach on the ticket, we wouldn't need to be passing the SGCfg prefix down to DCPCommon, and I think that means we could also remove it from FeedArgs.  If there was a reason to use the format you've got here, can you provide the details?

adamcfraser (changes requested):

db/change_listener.go
```diff
@@ -36,16 +36,20 @@ type changeListener struct {
...
 type DocChangedFunc func(event sgbucket.FeedEvent)
 
-func (listener *changeListener) Init(name string) {
+func (listener *changeListener) Init(name string, groupID string) {
 	listener.bucketName = name
 	listener.counter = 1
 	listener.terminateCheckCounter = 0
 	listener.keyCounts = map[string]uint64{}
 	listener.tapNotifier = sync.NewCond(&sync.Mutex{})
+	listener.checkpointPrefix = base.DCPCheckpointPrefixWithGroupID(groupID)
```
adamcfraser: changeListener doesn't do any checkpointing, so this shouldn't actually be necessary to set (here or in the FeedArgs for changeListener).  I think I'd prefer to take it out, to reduce complexity and avoid giving the impression that changeListener is checkpointing.

adamcfraser (changes requested):

db/sg_replicate_cfg_test.go
```diff
@@ -626,3 +626,85 @@ func TestIsCfgChanged(t *testing.T) {
...
+	require.NoError(t, err)
+	err = managerGroupB.AddReplication(&ReplicationCfg{
+		ReplicationConfig: ReplicationConfig{
+			ID:           "replicatorGroupB",
+			InitialState: ReplicationStateStopped,
+		},
+	})
+	require.NoError(t, err)
+
+	// Check replications are assigned to correct nodes
```
adamcfraser, line 689: Do we want colliding replicator names in different config groups to ensure that the cfg definitions are isolated?

rest/admin_api_test.go
```diff
@@ -4119,3 +4120,108 @@ func TestEmptyStringJavascriptFunctions(t *testing.T) {
...
+				passiveBucket.GetName(), base.TestsDisableGSI(), channels.DefaultSyncFunction,
+			),
+		)
+		assert.Equal(t, http.StatusCreated, resp.StatusCode)
+	}
+
+	// Start replicators
+	for i, group := range groupIDs {
+		channel := "chan" + group
+		replicationID := "repl" + group
```
adamcfraser: Similar to the previous test - since we don't have any collision on replication IDs, it seems like this test might have passed prior to the namespacing enhancements in this PR?

adamcfraser (changes requested):

adamcfraser (approved):

PR #5330, opened 2021-11-12
CBG-1769: Import filter and sync function can be set and retrieved while db is offline

IsaacLambat:
CBG-1769
- Made `/{db}/_config/import_filter` and `/{db}/_config/sync` endpoints available while the database is offline for `PUT`, `GET`, and `DELETE` methods
- Made tests to test all these methods on the endpoint while the database is offline

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1​38​7

### DIFF
--- a/rest/admin_api_test.go
+++ b/rest/admin_api_test.go
@@ -3797,14 +3797,6 @@ func TestDbOfflineConfigLegacy(t *testing.T) {
 	require.Equal(t, http.StatusOK, resp.Code)
 	dbConfigBeforeOffline := string(resp.BodyBytes())
 
-	resp = rt.SendAdminRequest("GET", "/db/_config/import_filter", "")
-	require.Equal(t, http.StatusOK, resp.Code)
-	importFilterBeforeOffline := string(resp.BodyBytes())
-
-	resp = rt.SendAdminRequest("GET", "/db/_config/sync", "")
-	require.Equal(t, http.StatusOK, resp.Code)
-	syncFuncBeforeOffline := string(resp.BodyBytes())
-
 	// Take DB offline
 	resp = rt.SendAdminRequest("POST", "/db/_offline", "")
 	require.Equal(t, http.StatusOK, resp.Code)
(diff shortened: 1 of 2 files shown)


rest/admin_api_test.go
```diff
@@ -3781,28 +3781,49 @@ func TestDatabaseOfflineConfigLegacy(t *testing.T) {
...
 	require.Equal(t, http.StatusCreated, resp.Code)
 
+	// Get config values before taking db offline
 	resp = rt.SendAdminRequest("GET", "/db/_config", "")
 	require.Equal(t, http.StatusOK, resp.Code)
 	dbConfigBeforeOffline := string(resp.BodyBytes())
 
+	resp = rt.SendAdminRequest("GET", "/db/_config/import_filter", "")
+	require.Equal(t, http.StatusOK, resp.Code)
+	importFilterBeforeOffline := string(resp.BodyBytes())
```
bbrks: It would be nice to assert that this (and sync below) is equal to the one specified in the dbconfig.

I'm actually not sure the handler works in non-persistent config mode, even though this test is passing - becuase it's essentially asserting `"" == ""`

rest/admin_api_test.go
```diff
@@ -3831,28 +3852,219 @@ func TestDatabaseOfflineConfigPersistent(t *testing.T) {
...
+
+	// Check configs match
+	resp = bootstrapAdminRequest(t, http.MethodGet, "/db/_config/import_filter", "")
+	require.Equal(t, http.StatusOK, resp.StatusCode)
+	respBuf := new(bytes.Buffer)
+	_, err = respBuf.ReadFrom(resp.Body)
+	require.NoError(t, err)
+	importFilterAfterOnline := respBuf.String()
+	require.NoError(t, resp.Body.Close())
+	assert.Equal(t, "", importFilterAfterOnline)
```
bbrks: Would be nice to check the functional behaviour of the `import_filter` and `sync` function reverting to defaults after going offline, deleting, and bringing back online by writing a few docs into the bucket and have import run them through the default import filter and sync functions.

In order to test this, your sync and import filter functions would have to be set to disallow the document being written at the end of the test to ensure delete does revert back to default to allow the doc to be written/imported.

bbrks (changes requested):
Would like to extend test to functional behaivour rather than just checking the config endpoints.

In addition, I think the current non-persistent test works but not for the reason we'd expect.

rest/admin_api_test.go
IsaacLambat: Yes, the `import_filter` and `sync` endpoints do not work in non-persistent config mode. I have removed them from the test, and instead am just going to assert both offline and online `GET /db/_config` endpoints bodies match which includes the `import_filter` and `sync` functions as the test was originally.

rest/admin_api_test.go
```diff
@@ -3831,28 +3836,236 @@ func TestDatabaseOfflineConfigPersistent(t *testing.T) {
...
 	require.NoError(t, base.JSONDecoder(resp.Body).Decode(&dbConfigBeforeOffline))
 	require.NoError(t, resp.Body.Close())
 
+	resp = bootstrapAdminRequest(t, http.MethodGet, "/db/_config/import_filter", "")
+	require.Equal(t, http.StatusOK, resp.StatusCode)
+	respBuf := new(bytes.Buffer)
+	_, err = respBuf.ReadFrom(resp.Body)
+	require.NoError(t, err)
+	importFilterBeforeOffline := respBuf.String()
+	require.NoError(t, resp.Body.Close())
```
bbrks: Use `assertResp` helper here

bbrks (changes requested):

bbrks (approved):

PR #5331, opened 2021-11-15
CBG-1788: Only allow any bucket to be referenced once

JRascagneres:
CBG-1788

Adds a simple check to avoid running a persistent config where any bucket is referenced by more than one database.
In the event that this it the case

Logging example:
```
2021-11-15T15:21:29.600Z [ERR] Couldn't start Sync Gateway: automatic upgrade to persistent config failed. Only one database can target any given bucket. x1 used by db2 and db -- rest.ServerMain() at main.go:26
```

Modified existing test to add test case and modify the way that the error string is checked.

Integration test not required. Direct testing of `sanitizeDbConfigs` should be sufficient. Have validated with manual testing to verify.

### DIFF
--- a/rest/main.go
+++ b/rest/main.go
@@ -273,7 +273,7 @@ func automaticConfigUpgrade(configPath string) (sc *StartupConfig, disablePersis
 func sanitizeDbConfigs(configMap DbConfigMap) (DbConfigMap, error) {
 	var databaseServerAddress string
 
-	processedBucketNames := make(map[string]struct{}, len(configMap))
+	processedBucketDbNames := make(map[string]string, len(configMap))
 
 	for dbName, dbConfig := range configMap {
 		if dbConfig.Server == nil || *dbConfig.Server == "" {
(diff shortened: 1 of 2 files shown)


rest/main.go
```diff
@@ -305,6 +307,14 @@ func sanitizeDbConfigs(configMap DbConfigMap) (DbConfigMap, error) {
 
 		// Make sure any updates are written back to the config
 		configMap[dbName] = dbConfig
+
+		if _, ok := processedBucketNames[*dbConfig.Bucket]; ok {
+			return nil, fmt.Errorf("automatic upgrade to persistent config failed. Only one database can " +
+				"target any given bucket")
+		}
+
+		processedBucketNames[*dbConfig.Bucket] = struct{}{}
```
bbrks: I'm guessing we don't want to modify the given `configMap` before we've done the bucket check, even if the caller is going to be using the error to stop processing.

```suggestion
		if _, ok := processedBucketNames[*dbConfig.Bucket]; ok {
			return nil, fmt.Errorf("automatic upgrade to persistent config failed. Only one database can " +
				"target any given bucket")
		}
		processedBucketNames[*dbConfig.Bucket] = struct{}{}

		// Make sure any updates are written back to the config
		configMap[dbName] = dbConfig
```

rest/main_test.go
```diff
@@ -109,51 +109,66 @@ func TestParseFlags(t *testing.T) {
...
+				"db2": &DbConfig{
+					BucketConfig: BucketConfig{Server: base.StringPtr("1.2.3.4"), Bucket: base.StringPtr("bucket")},
+				},
+			},
+			expectedError: "automatic upgrade to persistent config failed. Only one database can target any given bucket",
 		},
 	}
 	for _, test := range testCases {
 		t.Run(test.name, func(t *testing.T) {
 			dbConfigMap, err := sanitizeDbConfigs(test.input)
```
bbrks, line 161: As an aside, the function signature returning `configMap` makes it unclear whether the function is supposed to be returning a new map, or modifying the input.

The expected `// strip now disallowed options` sanitising behaviour isn't covered by this test either, opening this up to an accidental breaking change in the future.

Could we ensure at least `"server"` is being nilled in both: the returned `dbConfigMap` and `test.input` to assert current behaviour.

bbrks (changes requested):

rest/main.go
JRascagneres: Figured it didn't really make a difference but sure.

bbrks: The only reason it does is that this function is modifying the input map, rather than returning a copy.

We can't guarantee the caller is always going to be checking the error.

bbrks (approved):

PR #5334, opened 2021-11-16
CBG-1790: Panic handling added to NewImportPIndexImpl and unit test to repro bug

IsaacLambat:
CBG-1790

- Added unit test to reproduce panic that happens when deleting a database that targets the same bucket as another when import_docs is true in legacy. This is skipped for now.
- Added panic handling function so the panic caught and logged before SG exits

### DIFF
--- a/rest/admin_api_test.go
+++ b/rest/admin_api_test.go
@@ -3937,3 +3937,53 @@ func TestEmptyStringJavascriptFunctions(t *testing.T) {
 	assert.Contains(t, string(respBody), "sync function cannot be empty string")
 	assert.Contains(t, string(respBody), "import filter function cannot be empty string")
 }
+
+// CBG-1790: Deleting a database that targets the same bucket as another causes a panic in legacy
+func TestDeleteDatabasePointingAtSameBucket(t *testing.T) {
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("This test only works against Couchbase Server")
+	}
+	defer base.SetUpTestLogging(base.LevelInfo, base.KeyHTTP)()
+	// Start SG with no databases in bucket(s)
+	config := bootstrapStartupConfigForTest(t)
+	sc, err := setupServerContext(&config, false)
+	require.NoError(t, err)
+	defer sc.Close()
+	serverErr := make(chan error, 0)
+	go func() {
+		serverErr <- startServer(&config, sc)
+	}()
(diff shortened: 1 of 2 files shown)


rest/admin_api_test.go
```diff
@@ -3937,3 +3937,56 @@ func TestEmptyStringJavascriptFunctions(t *testing.T) {
...
+	assert.Equal(t, http.StatusCreated, resp.StatusCode)
+
+	resp = bootstrapAdminRequest(t, http.MethodDelete, "/db2/", "")
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+
+	resp = bootstrapAdminRequest(t, http.MethodGet, "/db2/_config", "")
+	assert.Equal(t, http.StatusNotFound, resp.StatusCode)
+
+	// Make another database that uses import in-order to trigger the panic instantly instead of having to time.Sleep
+	resp = bootstrapAdminRequest(t, http.MethodPut, "/db3/", fmt.Sprintf(dbConfig, "db3"))
```
bbrks, line 3990: I think this test can be simplified a bit, as it's not nessesary to use the bootstrap style of test to reproduce the panic.

You can get away with a RestTester and a TestBucket, using the fact that there's a built-in database to remove first.

```suggestion
	if base.UnitTestUrlIsWalrus() || !base.TestUseXattrs() {
		t.Skip("This test only works against Couchbase Server with xattrs")
	}

	defer base.SetUpTestLogging(base.LevelInfo, base.KeyHTTP)()

	tb := base.GetTestBucket(t)
	rt := NewRestTester(t, &RestTesterConfig{TestBucket: tb})
	defer rt.Close()

	resp := rt.SendAdminRequest(http.MethodDelete, "/db/", "")
	assertStatus(t, resp, http.StatusOK)

	// Make another database that uses import in-order to trigger the panic instantly instead of having to time.Sleep
	resp = rt.SendAdminRequest(http.MethodPut, "/db1/", fmt.Sprintf(`{
  "bucket": "%s",
  "username": "%s", 
  "password": "%s",
  "use_views": %t,
  "num_index_replicas": 0
}`, tb.GetName(), base.TestClusterUsername(), base.TestClusterPassword(), base.TestsDisableGSI()))
```

bbrks (changes requested):
Minor comment to simplify the disabled test, but feel free to copy/paste if you agree with it

bbrks:
Closing PR as need to revisit for fix

PR #5342, opened 2021-11-24
CBG-1547: Fixed leaking continuous sub changes feed when not sending changes

IsaacLambat:
CBG-1547

- Added terminator check at the start of the database changes fetch to terminate the sub-changes feed regardless of if there are any changes to be sent
- `TestMultipleOutstandingChangesSubscriptions` is moved to `blip_api_test.go` so it can be race tested
- Changed test to assert stats and applied test fixes from branch [`cbg1547-prereq`](https://github.com/couchbase/sync_gateway/compare/CBG-1547...cbg1547-prereq)

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1415 (flaky test failures)

### DIFF
--- a/rest/blip_api_test.go
+++ b/rest/blip_api_test.go
@@ -3672,7 +3672,6 @@ func TestMultipleOutstandingChangesSubscriptions(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, "", errorCode, "resp: %s", respBody)
 
-	pullStats = bt.restTester.GetDatabase().DbStats.CBLReplicationPull()
 	assert.EqualValues(t, 1, pullStats.NumPullReplActiveOneShot.Value())
 
 	// Send continous subChanges to subscribe to changes, which will cause the "changes" profile handler above to be called back
@@ -3690,8 +3689,7 @@ func TestMultipleOutstandingChangesSubscriptions(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, "", errorCode, "resp: %s", respBody)
 
-	pullStats = bt.restTester.GetDatabase().DbStats.CBLReplicationPull()
-	// Might need to expect 0 after CBG-1824
+	// Might need to expect 0 after CBG-1824 depending on fix implementation
 	assert.EqualValues(t, 1, pullStats.NumPullReplActiveContinuous.Value())
 
 	// Send a second continuous subchanges request, expect an error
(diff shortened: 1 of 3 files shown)


rest/blip_api_test.go
```diff
@@ -3628,3 +3628,113 @@ func TestMinRevPosWorkToAvoidUnnecessaryProveAttachment(t *testing.T) {
...
+	subChangesResponse = subChangesRequest.Response()
+	goassert.Equals(t, subChangesResponse.SerialNumber(), subChangesRequest.SerialNumber())
+	errorCode = subChangesResponse.Properties["Error-Code"]
+	log.Printf("errorCode: %v", errorCode)
+	respBody, err = subChangesResponse.Body()
+	require.NoError(t, err)
+	assert.Equal(t, "", errorCode, "resp: %s", respBody)
+
+	pullStats = bt.restTester.GetDatabase().DbStats.CBLReplicationPull()
+	// Might need to expect 0 after CBG-1824
```
bbrks: Not here we don't - I guess this comment was meant to be put up at line 3676 ?

rest/blip_api_test.go
```diff
@@ -3628,3 +3628,113 @@ func TestMinRevPosWorkToAvoidUnnecessaryProveAttachment(t *testing.T) {
...
+	goassert.True(t, sent)
+	subChangesResponse := subChangesRequest.Response()
+	goassert.Equals(t, subChangesResponse.SerialNumber(), subChangesRequest.SerialNumber())
+	errorCode := subChangesResponse.Properties["Error-Code"]
+	log.Printf("errorCode: %v", errorCode)
+	respBody, err := subChangesResponse.Body()
+	require.NoError(t, err)
+	assert.Equal(t, "", errorCode, "resp: %s", respBody)
+
+	pullStats = bt.restTester.GetDatabase().DbStats.CBLReplicationPull()
```
bbrks: We shouldn't need to keep retrieving this. It's a pointer, so can be set once at the start of the test, and values read from it throughout.

bbrks (changes requested):
Just a couple of nitpicks on the test

rest/blip_api_test.go
IsaacLambat: Put that there as well as thought if the CBG-1824 fix is to decrement the stat after the changes have been sent, then they will be decremented before the assert gets called.

bbrks (approved):

PR #5344, opened 2021-11-24
CBG-1808: convert empty sync function to nil

cbbruno:
CBG-1808

- converts sync func to nil if it is empty or blanks
- same for import filter
- done on validateVersion to work legacy or v3 configs
- added check in existing `TestInvalidJavascriptFunctions`

## Integration Test
http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1416/

### DIFF
--- a/rest/persistent_config_test.go
+++ b/rest/persistent_config_test.go
@@ -319,12 +319,12 @@ func TestImportFilterEndpoint(t *testing.T) {
 			tb.GetName(), base.TestsDisableGSI(),
 		),
 	)
-	_ = resp.Body.Close()
+	assert.NoError(t, resp.Body.Close())
 	assert.Equal(t, http.StatusCreated, resp.StatusCode)
 
 	// Ensure we won't fail with an empty import filter
 	resp = bootstrapAdminRequest(t, http.MethodPut, "/db1/_config/import_filter", "")
-	_ = resp.Body.Close()
+	assert.NoError(t, resp.Body.Close())
 	assert.Equal(t, http.StatusOK, resp.StatusCode)
 
 	// Add a document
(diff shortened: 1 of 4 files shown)


bbrks (dismissed):

rest/persistent_config_test.go
```diff
@@ -319,25 +319,26 @@ func TestImportFilterEndpoint(t *testing.T) {
...
 	assert.Equal(t, http.StatusCreated, resp.StatusCode)
 
-	// Ensure we cannot set an empty import filter
+	// Ensure we won't fail with an empty import filter
 	resp = bootstrapAdminRequest(t, http.MethodPut, "/db1/_config/import_filter", "")
-	assert.Equal(t, http.StatusBadRequest, resp.StatusCode)
-	responseBody, err := ioutil.ReadAll(resp.Body)
-	assert.NoError(t, err)
-	assert.Contains(t, string(responseBody), "import filter function cannot be empty string")
+	_ = resp.Body.Close()
```
adamcfraser: We have a lot of rest tester code that issues REST API requests (via SendRequest and SendAdminRequest) where the tests aren't explicitly closing the response body.  Is that actually required whenever bootstrapAdminRequest is invoked?  It would be nice to avoid this kind of noise in every test unless it's strictly needed.

adamcfraser (changes requested):
Changes look fine, had a question about the need to close the response body.

rest/persistent_config_test.go
JRascagneres: Yeah this is required with the `bootstrapAdminRequest` method. @bbrks work to improve the testing framework to work nicely with persistent config should mean this isn't required in the future. But as it stands with persistent config testing, this is required.

adamcfraser: Why is it the close required in this usage and not the case where we're using Handler.ServeHTTP for REST API testing?

JRascagneres: I was under the impression the underlying TestResponse would read the body and close the request for us. Perhaps I'm wrong about that.

adamcfraser (approved):

PR #5349, opened 2021-11-30
CBG-1559: Fix TestHandleDBConfig unclosed bucket

cbbruno:
CBG-1559

- Re-enabled test
- Used same test bucket config in Rest Tester
- Used same db name as Rest Tester (db), instead of bucket name, to avoid 2 DBs pointing to the same bucket

## Integration Tests
- [x] http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1438/

### DIFF
--- a/rest/admin_api_test.go
+++ b/rest/admin_api_test.go
@@ -2270,16 +2270,13 @@ func TestHandlePutDbConfigWithBackticks(t *testing.T) {
 }
 
 func TestHandleDBConfig(t *testing.T) {
-	t.Skip("disabled pending CBG-1566")
 	if base.GTestBucketPool.NumUsableBuckets() < 2 {
 		t.Skipf("test requires at least 2 usable test buckets")
 	}
 
-	rt := NewRestTester(t, nil)
-
 	tb := base.GetTestBucket(t)
 
-	defer tb.Close()
+	rt := NewRestTester(t, &RestTesterConfig{TestBucket: tb})
 	defer rt.Close()
 
 	bucket := tb.GetName()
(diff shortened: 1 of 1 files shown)


bbrks (approved):

rest/admin_api_test.go
```diff
@@ -2270,45 +2270,40 @@ func TestHandlePutDbConfigWithBackticks(t *testing.T) {
 }
 
 func TestHandleDBConfig(t *testing.T) {
-	t.Skip("disabled pending CBG-1566")
 	if base.GTestBucketPool.NumUsableBuckets() < 2 {
 		t.Skipf("test requires at least 2 usable test buckets")
 	}
```
bbrks: I think we can remove this now the test is only using one bucket.

rest/admin_api_test.go
```diff
@@ -2270,45 +2270,40 @@ func TestHandlePutDbConfigWithBackticks(t *testing.T) {
...
 
 	// change cache size so we can see the update being reflected in the API response
 	dbConfig := &DbConfig{
+		BucketConfig: BucketConfig{Bucket: &bucket},
 		CacheConfig: &CacheConfig{
 			RevCacheConfig: &RevCacheConfig{
 				Size: base.Uint32Ptr(1337), ShardCount: base.Uint16Ptr(7),
 			},
 		},
+		NumIndexReplicas:   base.UintPtr(0),
```
bbrks, line 2302: We can also set these, primarily to avoid the test creating indexes when we don't have them enabled.

```suggestion
		NumIndexReplicas:   base.UintPtr(0),
		EnableXattrs:       base.BoolPtr(base.TestUseXattrs()),
		UseViews:           base.BoolPtr(base.TestsDisableGSI()),
```

bbrks (changes requested):
Actually a couple of nitpicks :D

bbrks (approved):

PR #5363, opened 2021-12-07
CBG-1720: added mutex, concurrent calls in test

cbbruno:
CBG-1720

- added Mutex so concurrent calls to FlushLogBuffers do not cause data race or wait group reuse issues.
- added concurrent calls to FlushLogBuffers in the test.

- [ ] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1​46​6/

### DIFF
--- a/rest/server_context_test.go
+++ b/rest/server_context_test.go
@@ -634,18 +634,16 @@ func TestLogFlush(t *testing.T) {
 			base.FlushLoggerBuffers()
 
 			// Concurrent calls to FlushLogBuffers should not cause data race or wait group reuse issues
-			// Wait for concurrent calls so they don't cause issues with SetupAndValidateLogging of next t.Run
+			// Wait for concurrent calls so they don't cause issues with SetupAndValidateLogging from next t.Run
 			var flushCallsWg = sync.WaitGroup{}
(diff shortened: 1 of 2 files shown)


rest/server_context_test.go
```diff
@@ -632,8 +633,20 @@ func TestLogFlush(t *testing.T) {
...
+				flushCallsWg.Add(1)
+				go func() {
+					base.FlushLogBuffers()
+					flushCallsWg.Done()
+				}()
+			}
+
 			// Flush collation buffers to ensure the files that will be built do get written
 			base.FlushLogBuffers()
+			flushCallsWg.Wait()
```
adamcfraser, line 647: Is the additional flush (on line 648) still needed now that we're flushing as part of the concurrency checking?

adamcfraser (changes requested):
Change looks good - minor question about the test.

rest/server_context_test.go
cbbruno: True... also bumped number of concurrent calls to make sure we hit wait re-use issue more easily, after more testing locally.

adamcfraser (approved):

PR #5364, opened 2021-12-08
CBG-1790 Define cbgt index types per config group, use indexParams to retrieve dest implementations

adamcfraser:
CBG-1790

Previously SG defined an cbgt index type definition per db, and the index type definition callbacks referenced that database's import listener directly. This introduces race conditions in cbgt, which expects type definitions to be registered in memory once at process start, and not removed.

An index type for the config group is now created on server context initialization, and databases register indexes of that type. In order for the config group's index callbacks to properly target specific database contexts (specifically the cbgt.Dest implementation that will be working the feed), databases register their dests into a synchronized global map on startup (and remove from the map on database stop). The dest map key is stored with the per-databases cbgt index registration (stored in cbgt's bucket cfg per database).

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1469/

### DIFF
--- a/base/dcp_sharded.go
+++ b/base/dcp_sharded.go
@@ -98,6 +98,9 @@ func createCBGTIndex(c *CbgtContext, dbName string, configGroupID string, bucket
 	}
 
 	indexParams, err := cbgtIndexParams(ImportDestKey(dbName))
+	if err != nil {
+		return err
+	}
 
 	vbNo, err := bucket.GetMaxVbno()
 	if err != nil {
@@ -560,6 +563,12 @@ var cbgtDestFactoriesLock sync.Mutex
 
 func StoreDestFactory(destKey string, dest CbgtDestFactoryFunc) {
 	cbgtDestFactoriesLock.Lock()
+	_, ok := cbgtDestFactories[destKey]
+
+	// We don't expect duplicate destKey registration - log a warning if it already exists
+	if ok {
+		Warnf("destKey %s already exists in cbgtDestFactories - new value will replace the existing dest")
+	}
 	cbgtDestFactories[destKey] = dest
 	cbgtDestFactoriesLock.Unlock()
 }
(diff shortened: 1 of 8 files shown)


base/dcp_sharded.go
```diff
@@ -97,6 +97,8 @@ func createCBGTIndex(c *CbgtContext, dbName string, bucket Bucket, spec BucketSp
 		return err
 	}
 
+	indexParams, err := cbgtIndexParams(ImportDestKey(dbName))
```
bbrks, line 100: unhandled error if `cbgtIndexParams` fails to marshal

base/dcp_sharded.go
```diff
@@ -544,3 +549,33 @@ func (l *importHeartbeatListener) GetNodes() ([]string, error) {
...
+// global map to avoid races between db creation and db addition to the server context database set
+
+type CbgtDestFactoryFunc = func() (cbgt.Dest, error)
+
+var cbgtDestFactories = make(map[string]CbgtDestFactoryFunc)
+var cbgtDestFactoriesLock sync.Mutex
+
+func StoreDestFactory(destKey string, dest CbgtDestFactoryFunc) {
+	cbgtDestFactoriesLock.Lock()
+	cbgtDestFactories[destKey] = dest
```
bbrks, line 572: Given it's an unexpected scenario that `destKey` already exists in the map, is it worth checking and returning an error from this function so it can be logged?

db/import_listener.go
```diff
@@ -55,8 +55,11 @@ func (il *importListener) StartImportFeed(bucket base.Bucket, dbStats *base.DbSt
 
 	importFeedStatsMap := dbContext.DbStats.Database().ImportFeedMapStats
 
-	// Register cbgt PIndex to support sharded import.
-	il.RegisterImportPindexImpl()
+	// Store the listener in global map for dbname-based retrieval by cbgt prior to index registration
+	base.StoreDestFactory(base.ImportDestKey(il.database.Name), il.NewImportDest)
+
+	// Register cbgt PIndex to support sharded import, if not already defined
+	//il.RegisterImportPindexImpl()
```
bbrks: Looks like this is handled elsewhere now via the globally scoped `RegisterImportPindexImpl` function.

bbrks (changes requested):
Approach seems good, a couple of comments.

bbrks (approved):

PR #5367, opened 2021-12-08
Increase log level in flaky TestDocEtag (CBG 1734)

cbbruno:
CBG-1734

- failure seems rare and there is not enough info to try to repro.
- added logging setup to collect more data on next failure.
- set debug level to reduce logging noise.

- [ ] `xattrs=true` (http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1478/)

### DIFF
--- a/rest/api_test.go
+++ b/rest/api_test.go
@@ -117,6 +117,8 @@ func TestDocLifecycle(t *testing.T) {
 
 //Validate that Etag header value is surrounded with double quotes, see issue #808
 func TestDocEtag(t *testing.T) {
+	defer base.SetUpTestLogging(base.LevelDebug, base.KeyAccess)()
+
 	rt := NewRestTester(t, &RestTesterConfig{guestEnabled: true})
 	defer rt.Close()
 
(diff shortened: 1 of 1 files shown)


rest/api_test.go
```diff
@@ -117,6 +117,8 @@ func TestDocLifecycle(t *testing.T) {
 
 //Validate that Etag header value is surrounded with double quotes, see issue #808
 func TestDocEtag(t *testing.T) {
+	defer base.SetUpTestLogging(base.LevelDebug, base.KeyAccess)()
```
bbrks: `KeyAccess` is used to log information about user access grant changes (i.e. new channel access, role changes, etc.) - given we're not sure what causes this flaky test, I'd suggest `KeyAll` to enable all types of logging.

```suggestion
	defer base.SetUpTestLogging(base.LevelDebug, base.KeyAll)()
```

bbrks (changes requested):
Just a comment about which log key to set

https://docs.couchbase.com/sync-gateway/current/logging.html#lbl-log-keys

bbrks (approved):

PR #5371, opened 2021-12-10
CBG-1502 Checkpoint persistence for one-shot DCP feeds

adamcfraser:
CBG-1502

Switches one-shot DCP feed to persist checkpoint metadata to the bucket, to support resumption after node failure.  Metadata is persisted per DCP worker, to avoid contention on in-memory metadata at persistence time.  Workers schedule persistence for their metadata based on their metaPersistFrequency.  This defaults to 1 minute, but can be adjusted for unit testing.

On successful one-shot feed completion, checkpoints are purged from the bucket.

- [ ] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1488/

### DIFF
--- a/base/dcp_client_test.go
+++ b/base/dcp_client_test.go
@@ -260,8 +260,7 @@ func TestResumeStoppedFeed(t *testing.T) {
 	mutationCount := uint64(0)
 	counterCallback := func(event sgbucket.FeedEvent) bool {
 		if bytes.HasPrefix(event.Key, []byte(t.Name())) {
-			atomic.AddUint64(&mutationCount, 1)
-			count := atomic.LoadUint64(&mutationCount)
+			count := atomic.AddUint64(&mutationCount, 1)
 			if count > 5000 {
 				err := dcpClient.Close()
 				assert.NoError(t, err)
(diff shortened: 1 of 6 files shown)


base/dcp_client_test.go
```diff
@@ -243,8 +238,108 @@ func TestDCPClientMultiFeedConsistency(t *testing.T) {
...
+	defer bucket.Close()
+
+	var dcpClient *DCPClient
+
+	// create callback
+	mutationCount := uint64(0)
+	counterCallback := func(event sgbucket.FeedEvent) bool {
+		if bytes.HasPrefix(event.Key, []byte(t.Name())) {
+			atomic.AddUint64(&mutationCount, 1)
+			count := atomic.LoadUint64(&mutationCount)
```
bbrks: I think this might be racy (and call Close twice)

You can use the count returned by Add instead of the additional atomic Load.

bbrks (changes requested):
Just a small test-only race. Feel free to fix in a subsequent PR

bbrks (approved):

PR #5372, opened 2021-12-10
CBG-1846: Load database on /db/_config update prior to writing config

bbrks:
CBG-1846

- On a failed load during `/db/_config` update, the database is rolled back to the version in the bucket.
- Add more upfront `dbConfig` validation to avoid rollback churn in most cases.
- Add `RestTesterCluster` and use it in `TestPersistentDbConfigWithInvalidUpsert` to reproduce issue.

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1487/
  - `TestAttachmentCompactionStopImmediateStart` Race assertion fixed in #5371

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -24,7 +24,6 @@ import (
 	"os/signal"
 	"strings"
 	"syscall"
-	"time"
 
 	"golang.org/x/crypto/bcrypt"
 
@@ -470,6 +469,7 @@ func (dbConfig *DbConfig) AutoImportEnabled() (bool, error) {
 
 const dbConfigFieldNotAllowedErrorMsg = "Persisted database config does not support customization of the %q field"
 
+// validatePersistentDbConfig checks for fields that are only allowed in non-persistent mode.
 func (dbConfig *DbConfig) validatePersistentDbConfig() (errorMessages error) {
 	if dbConfig.Server != nil {
 		errorMessages = multierror.Append(errorMessages, fmt.Errorf(dbConfigFieldNotAllowedErrorMsg, "server"))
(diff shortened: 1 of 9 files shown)


rest/config.go
```diff
@@ -1082,19 +1067,31 @@ func (sc *ServerContext) fetchAndLoadConfigs() (count int, err error) {
...
 	}
-	sc.applyConfigs(map[string]DatabaseConfig{dbName: *dbConfig})
+	sc._applyConfigs(map[string]DatabaseConfig{dbName: *dbConfig})
 
 	return true, nil
 }
 
 func (sc *ServerContext) fetchDatabase(dbName string) (found bool, dbConfig *DatabaseConfig, err error) {
+	select {
+	case <-sc.bootstrapContext.terminator:
```
adamcfraser: Do we need to worry about the bootstrapContext being terminated inside fetchDatabase after this initial check?

rest/server_context_test.go
```diff
@@ -418,7 +418,7 @@ func TestTLSSkipVerifyCombinations(t *testing.T) {
 				},
 			}
 
-			err := startupConfig.validate()
+			err := startupConfig.validate(false)
```
adamcfraser: Should this use base.IsEnterpriseEdition()?

rest/config_test.go
```diff
@@ -1322,7 +1323,7 @@ func TestClientTLSMissing(t *testing.T) {
 			if test.tlsCert {
 				config.API.HTTPS.TLSCertPath = "test.cert"
 			}
-			err := config.validate()
+			err := config.validate(false)
```
adamcfraser: Same question as above, about whether we can use base.IsEnterpriseEdition here.

adamcfraser (changes requested):
Looks good - one question on bootstrap terminator check, and minor test nits about using IsEnterpriseEdition when calling config.validate.

rest/config.go
bbrks: I think this was added as a fix for when `handleCreateDB` was doing an async on-demand load and racing during test teardown. I don't think the check is required at all now that is synchronous.

adamcfraser (approved):

PR #5379, opened 2021-12-15
CBG-1046: Added ability to specify user for active peer in ISGR

IsaacLambat:
CBG-1046

- Added new replicator config option `run_as` which will run the replication as the user defined. If left blank, the replication is ran as an admin user
- Added test to test sync function is being ran by correct user, and only the docs the user has access to is being replicated
- Deprecated replication config options `username` and `password` which have been replaced by `remote_username` and `remote_password`. This will log a deprecation warning every time the fields are modified
- Added a test to test that `username` and `password` is mapped on to `remote_username` and `remote_password` respectively and ensures the password redaction is working correctly

# Dependencies
- [x] Merge #5383 in to fix race condition in test

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1503

### DIFF
--- a/db/sg_replicate_cfg.go
+++ b/db/sg_replicate_cfg.go
@@ -250,18 +250,14 @@ func (rc *ReplicationConfig) Upsert(c *ReplicationUpsertConfig) {
 		rc.Remote = *c.Remote
 	}
 
-	if c.Username != nil || c.Password != nil {
-		base.Warnf(`Deprecation notice: replication config fields "username" and "password" are now "remote_username" and "remote_password" respectively`)
-		if c.Username != nil {
-			rc.RemoteUsername = *c.Username
-		}
-
-		if c.Username != nil { // Causes bug CBG-1858
-			rc.RemotePassword = *c.Password
-		}
+	if c.Username != nil {
+		rc.Username = *c.Username
 	}
 
-	// Remote credential fields take priority over deprecated credential fields
+	if c.Username != nil {
+		rc.Password = *c.Password
+	}
+
 	if c.RemoteUsername != nil {
 		rc.RemoteUsername = *c.RemoteUsername
 	}
@@ -355,6 +351,9 @@ func (rc *ReplicationConfig) Equals(compareToCfg *ReplicationConfig) (bool, erro
 // both replication config and remote URL, i.e., any password will be replaced with xxxxx.
 func (rc *ReplicationConfig) Redacted() *ReplicationConfig {
 	config := *rc
+	if config.Password != "" {
+		config.Password = base.RedactedStr
+	}
 	if config.RemotePassword != "" {
 		config.RemotePassword = base.RedactedStr
 	}
@@ -575,8 +574,17 @@ func (m *sgReplicateManager) NewActiveReplicatorConfig(config *ReplicationCfg) (
 
 	// If auth credentials are explicitly defined in the replication configuration,
 	// enrich remote database URL connection string with the supplied auth credentials.
+	username := config.Username
+	password := config.Password
+	if username != "" || password != "" {
+		base.Warnf(`Deprecation notice: replication config fields "username" and "password" are now "remote_username" and "remote_password" respectively`)
+	}
 	if config.RemoteUsername != "" {
-		rc.RemoteDBURL.User = url.UserPassword(config.RemoteUsername, config.RemotePassword)
+		username = config.RemoteUsername
+		password = config.RemotePassword
+	}
+	if username != "" {
+		rc.RemoteDBURL.User = url.UserPassword(username, password)
 	}
 
 	rc.WebsocketPingInterval = m.dbContext.Options.SGReplicateOptions.WebsocketPingInterval
--- a/db/sg_replicate_cfg.go
+++ b/db/sg_replicate_cfg.go
@@ -107,6 +107,7 @@ type ReplicationConfig struct {
 	Cancel                 bool                      `json:"cancel,omitempty"`
 	Adhoc                  bool                      `json:"adhoc,omitempty"`
 	BatchSize              int                       `json:"batch_size,omitempty"`
+	RunAsAdminUser         bool                      `json:"run_as_admin_user,omitempty"`
 }
 
 func DefaultReplicationConfig() ReplicationConfig {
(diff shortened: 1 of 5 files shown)


db/sg_replicate_cfg.go
```diff
@@ -248,7 +251,7 @@ func (rc *ReplicationConfig) Upsert(c *ReplicationUpsertConfig) {
 		rc.Username = *c.Username
 	}
 
-	if c.Username != nil {
+	if c.Username != nil && c.Password != nil {
```
bbrks: This change doesn't look right?

Upsert merges the given `c` into `rc` - which isn't going to happen if you try to just change a password without also re-stating the username.

db/sg_replicate_cfg.go
```diff
@@ -107,6 +107,7 @@ type ReplicationConfig struct {
 	Cancel                 bool                      `json:"cancel,omitempty"`
 	Adhoc                  bool                      `json:"adhoc,omitempty"`
 	BatchSize              int                       `json:"batch_size,omitempty"`
+	RunAsAdminUser         bool                      `json:"run_as_admin_user,omitempty"`
```
bbrks: Was there a decision made on changing this to be `local_username`? I think it has been discussed several times

db/sg_replicate_cfg.go
```diff
@@ -482,10 +489,19 @@ func (m *sgReplicateManager) NewActiveReplicatorConfig(config *ReplicationCfg) (
...
+			return nil, err
+		}
+		activeDB.SetUser(user)
+	}
+
 	rc = &ActiveReplicatorConfig{
 		ID:                 config.ID,
 		Continuous:         config.Continuous,
-		ActiveDB:           &Database{DatabaseContext: m.dbContext}, // sg-replicate interacts with local as admin
+		ActiveDB:           activeDB, // sg-replicate interacts with local as admin
```
bbrks: Not any more

```suggestion
		ActiveDB:           activeDB,
```

bbrks (changes requested):

db/sg_replicate_cfg.go
IsaacLambat: Yes we discussed it, but iirc, we where going to keep it unseparated, and separate it in the future if it gets confusing.

bbrks: I had the impression of the reverse coming away from those discussions :) It was already confusing having them mixed, and to do it correctly now rather than later and then have to worry about backwards compatibility.

bbrks: If you're unsure, let's leave it for now and decide on a call

db/sg_replicate_cfg.go
```diff
@@ -244,12 +250,24 @@ func (rc *ReplicationConfig) Upsert(c *ReplicationUpsertConfig) {
 		rc.Remote = *c.Remote
 	}
 
-	if c.Username != nil {
-		rc.Username = *c.Username
+	if c.Username != nil || c.Password != nil {
+		base.Warnf(`Deprecation notice: replication config fields "username" and "password" are now "remote_username" and "remote_password" respectively`)
+		if c.Username != nil {
+			rc.RemoteUsername = *c.Username
+		}
```
bbrks: Doing this mapping here in Upsert from the old to the new option breaks backwards compatbility when just loading existing replication configurations.

We don't have test coverage to catch that case given we're always writing configs through Upsert the REST API, so we'll need to think of a way to test that.

bbrks (changes requested):
We'll need to change how we're dealing with the deprecated options, as it doesn't work in a backwards compatible way.

db/sg_replicate_cfg.go
```diff
@@ -107,6 +109,7 @@ type ReplicationConfig struct {
 	Cancel                 bool                      `json:"cancel,omitempty"`
 	Adhoc                  bool                      `json:"adhoc,omitempty"`
 	BatchSize              int                       `json:"batch_size,omitempty"`
+	RunAs                  string                    `json:"run_as,omitempty"`
```
bbrks, line 112: Review whether it still makes sense to deprecate `username` and move to `remote_username` given this is now called `run_as`.

bbrks (commented):
.

db/sg_replicate_cfg.go
```diff
@@ -206,7 +212,7 @@ func (rc *ReplicationConfig) ValidateReplication(fromConfig bool) (err error) {
 		return base.HTTPErrorf(http.StatusBadRequest, "Replication remote URL is invalid")
 	}
 
-	if (remoteURL != nil && remoteURL.User.Username() != "") && rc.Username != "" {
+	if (remoteURL != nil && remoteURL.User.Username() != "") && rc.RemoteUsername != "" {
```
bbrks: It looks like this needs to be an _additional_ case. We want to limit the users to only providing credentials in one place. With this change it becomes possible to set in the deprecated `username` field and one of the above fields.

db/sg_replicate_cfg.go
```diff
@@ -148,6 +153,7 @@ type ReplicationUpsertConfig struct {
 	Cancel                 *bool       `json:"cancel,omitempty"`
 	Adhoc                  *bool       `json:"adhoc,omitempty"`
 	BatchSize              *int        `json:"batch_size,omitempty"`
+	RunAs                  *string     `json:"run_as,omitempty"`
```
bbrks, line 156: I think `RunAs` needs to be included in the `ActiveReplicatorConfig.CheckpointHash()` method - as it affects what data is replicated - and we cannot rely on the correctness of checkpoints if that value changes.

bbrks (changes requested):

db/sg_replicate_cfg.go
IsaacLambat: Fixed that. Can also limit it so the `username` and `remote_username` fields cannot be used at the same time if needed. 
```suggestion
	if rc.RemoteUsername != "" && rc.Username != "" {
		return base.HTTPErrorf(http.StatusBadRequest,
			"Cannot set remote_username and username config options. Please only set the remote_username and remote_password config options. ")
	}

	if (remoteURL != nil && remoteURL.User.Username() != "") && (rc.RemoteUsername != "" || rc.Username != "") {
```

bbrks (approved):
This looks good now, thanks!

PR #5383, opened 2021-12-21
CBG-1864: Fixed race condition caused when getting user roles

IsaacLambat:
CBG-1864

GetRoles is now preloaded when creating a new blip context or refreshing the user of the database

Tested locally with race detection for `TestSpecifyUserDocsToReplicate` from CBG-1046 https://github.com/couchbase/sync_gateway/pull/5379

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1549 failing test `TestBlipPusherUpdateDatabase` is skipped on master rebase

### DIFF
--- a/auth/principal.go
+++ b/auth/principal.go
@@ -123,6 +123,8 @@ type User interface {
 
 	RoleHistory() TimedSetHistory
 
+	PreloadRoles()
+
 	RevokedChannels(since uint64, lowSeq uint64, triggeredBy uint64) RevokedChannels
 
 	// Obtains the period over which the user had access to the given channel. Either directly or via a role.
--- a/auth/principal.go
+++ b/auth/principal.go
@@ -123,7 +123,7 @@ type User interface {
 
 	RoleHistory() TimedSetHistory
 
-	PreloadRoles()
+	InitializeRoles()
 
 	RevokedChannels(since uint64, lowSeq uint64, triggeredBy uint64) RevokedChannels
 
--- a/auth/user.go
+++ b/auth/user.go
@@ -29,6 +29,9 @@ type userImpl struct {
 	auth  *Authenticator
 	roles []Role
 
+	// getRolesLock ensures there is no race condition when checking if the roles have been set, and when setting
+	getRolesLock sync.RWMutex
+
 	// warnChanThresholdOnce ensures that the check for channels
 	// per user threshold is only performed exactly once.
 	warnChanThresholdOnce sync.Once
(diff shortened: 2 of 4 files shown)


auth/user.go
```diff
@@ -455,18 +458,28 @@ func (user *userImpl) SetPassword(password string) {
...
+				}
 			}
+			user.roles = roles
 		}
-		user.roles = roles
+		user.getRolesLock.Unlock()
+	} else {
+		user.getRolesLock.RUnlock()
 	}
 	return user.roles
```
bbrks, line 471: We're still reading `user.roles` outside of a lock here - so the race isn't fully fixed.

I think you'll have to read it into a variable whilst holding a lock, and return that variable. That would prevent a writer from changing to a new slice on line 478 underneath a read on 484

bbrks (changes requested):
I don't think the locks are fully covering the `user.roles` slice for a simultaneous read and write.

bbrks:
Assigned to @adamcfraser to review approach - given the potential cost of RLock being acquired for each run through the sync function. It's possible there's a lockless way to do this but couldn't come up with one.

adamcfraser:
I'd like to review whether we actually benefit from lazy-loading roles in the user context (i.e. what are the use cases where we won't ever need roles), and whether there's a better locking approach possible.

auth/user.go
```diff
@@ -471,6 +471,10 @@ func (user *userImpl) GetRoles() []Role {
...
 }
 
+func (user *userImpl) PreloadRoles() {
+	_ = user.GetRoles()
```
adamcfraser, line 475: Why is it safe to ignore error here?

auth/principal.go
```diff
@@ -123,6 +123,8 @@ type User interface {
 
 	RoleHistory() TimedSetHistory
 
+	PreloadRoles()
```
adamcfraser: Is there a reason to not just add GetRoles to the user interface?

adamcfraser (dismissed):

auth/user.go
IsaacLambat: That returns the roles that the user has which is not needed for preloading the roles as it is added on `userImpl` in GetRoles()

auth/principal.go
IsaacLambat: As discussed, changing to `InitializeRoles()`

bbrks (approved):

PR #5384, opened 2021-12-22
CBG-1581: Unit test to verify final checkpoint is done on replicator stopped

IsaacLambat:
CBG-1581

Added test to make sure the checkpoint document is wrote with the correct status when a replicator is stopped. Checkpointing at an interval is disabled.

- [x] `xattrs=true test=TestReplicatorCheckpointOnStop count=5` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1513

### DIFF
--- a/rest/admin_api_test.go
+++ b/rest/admin_api_test.go
@@ -4314,7 +4314,11 @@ func TestReplicatorCheckpointOnStop(t *testing.T) {
 	err := activeRT.GetDatabase().SGReplicateMgr.StartReplications()
 	require.NoError(t, err)
 
-	rev := activeRT.createDoc(t, "test")
+	database, err := db.CreateDatabase(activeRT.GetDatabase())
+	require.NoError(t, err)
+	rev, doc, err := database.Put("test", db.Body{})
+	require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


rest/admin_api_test.go
```diff
@@ -4297,3 +4297,58 @@ func TestDeleteDatabasePointingAtSameBucketPersistent(t *testing.T) {
...
+	val, _, err := activeRT.Bucket().GetRaw(expectedCheckpointName)
+	require.NoError(t, err)
+	var config struct {
+		Status struct {
+			Status string
+		}
+	}
+	err = json.Unmarshal(val, &config)
+	require.NoError(t, err)
+	assert.Equal(t, "stopped", config.Status.Status)
```
bbrks: Would it be better checking if the checkpoint doc contains a sequence that matches the doc sent over the replication rather than a status? That's the primary motivation for a checkpoint - the status being in there is an implementation detail of how we store status.

bbrks (changes requested):
Test looks OK although I do think we should be checking the sequence contents of the checkpoint instead.

bbrks (approved):

PR #5398, opened 2022-01-06
CBG-1875: Improve redaction for base.Set

JRascagneres:
CBG-1875

Added redaction for base.Set. Added a new struct `RedactorSet` which stores the set itself and also the 'type' of redaction we want to perform (UD, MD, SD).  Then when either Redact() or String() is called the set can be processed and either returned as a regular string, or as a redacted string. This requires only 1 iteration of the set at redact time.

This is covered by expanding unit tests.

### DIFF
--- a/base/redactor_userdata.go
+++ b/base/redactor_userdata.go
@@ -56,6 +56,10 @@ func UD(i interface{}) RedactorFunc {
 		return func() Redactor {
 			return UserData(v)
 		}
+	case Set:
+		return func() Redactor {
+			return v.buildRedactorSet(UD)
+		}
 	case fmt.Stringer:
 		return func() Redactor {
 			return UserData(v.String())
--- a/base/redactor.go
+++ b/base/redactor.go
@@ -76,6 +76,44 @@ func (redactorSlice RedactorSlice) String() string {
 	return "[ " + string(tmp) + "]"
 }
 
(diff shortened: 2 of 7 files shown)


base/redactor_userdata.go
```diff
@@ -56,6 +56,10 @@ func UD(i interface{}) RedactorFunc {
 		return func() Redactor {
 			return UserData(v)
 		}
+	case Set:
+		return func() Redactor {
+			return UD(v.ToArray())
```
adamcfraser: This approach looks like it's going to mean three iterations over the dataset (once in ToArray, once in the buildRedactorFuncSlice, and once in RedactorSlice.Redact).  It also has whatever overhead is associated with allocation of the intermediate slices, as well as the otherwise unnecessary reflection work.   Given that, and that set logging may end up on performance-critical paths, is there an option to have Set implement Redactor?

adamcfraser (changes requested):

base/redactor_userdata.go
JRascagneres: Have modified the approach to build a RedactorSlice using the new function set.buildRedactor(). This works in a similar way to the slice work but cuts out some of the work. It removes one iteration and removes the reflection work. It still has two iterations but I think this is the best way forward. Has an interation to build the redactor and then another when the item is either redacted or not redacted.

adamcfraser (approved):

PR #5406, opened 2022-01-18
CBG-1880: Listed all routes in API documentation

IsaacLambat:
CBG-1880

Made YAML file containing all the endpoints that Sync Gateway supports. Does not have titles or descriptions. All titles are a placeholder of `a`.

### DIFF
--- a/docs/api/api_docs.yaml
+++ b/docs/api/api_docs.yaml
@@ -1,0 +1,1189 @@
+openapi: 3.0.0
+info:
+  title: Sync Gateway API
+  version: '3.0'
+  license:
+    name: Business Source License 1.1 (BSL)
+    url: 'https://github.com/couchbase/sync_gateway/blob/master/LICENSE'
+  description: ''
+servers:
+  - url: 'http://localhost:4984'
+    description: Public API
+  - url: 'http://localhost:4985'
+    description: Admin API
+  - url: 'http://localhost:4986'
+    description: Metric API
+paths:
+  /:
+    get:
+      summary: Abc
+      responses: {}
+      operationId: get
+      x-internal: false
+      description: ''
+    head:
+      summary: ''
+      operationId: head
+      responses:
+        '200':
+          description: OK
+  '/{db}/':
+    parameters:
+      - schema:
+          type: string
+        name: db
+        in: path
+        required: true
+    get:
+      summary: aaa
+      responses: {}
+      operationId: get-db
+    post:
+      summary: a
+      responses: {}
+      operationId: post-db
+    delete:
+      summary: a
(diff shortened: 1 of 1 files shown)


docs/api/api_docs.yaml
```diff
@@ -0,0 +1,1189 @@
...
+    description: Public API
+  - url: 'http://localhost:4985'
+    description: Admin API
+  - url: 'http://localhost:4986'
+    description: Metric API
+paths:
+  /:
+    get:
+      summary: Abc
+      responses: {}
```
bbrks: Just fill in a default value containing the required fields to get the spec validating.

```suggestion
      responses: 
        default: 
          description: TODO
```

docs/api/api_docs.yaml
```diff
@@ -0,0 +1,1189 @@
...
+      operationId: get
+      x-internal: false
+      description: ''
```
bbrks: All but `responses` are [optional fields](https://swagger.io/specification/#path-item-object ). Seems worth removing to simplify until we have stuff to put in them.

docs/api/api_docs.yaml
```diff
@@ -0,0 +1,1189 @@
...
+      summary: ''
+      operationId: head
+      responses:
+        '200':
+          description: OK
+  '/{db}/':
+    parameters:
+      - schema:
+          type: string
+        name: db
```
bbrks: Nitpick/inventing convention - `name` first.

```suggestion
        - name: db
        schema:
          type: string
```

docs/api/api_docs.yaml
```diff
@@ -0,0 +1,1189 @@
...
+      operationId: get-_metrics
+  /_expvar:
+    get:
+      summary: a
+      tags: []
+      responses: {}
+      operationId: get-_expvar
+components:
+  schemas: {}
+  securitySchemes: {}
```
```suggestion
```

bbrks (changes requested):
Mostly comments around simplifying and making sure the spec is valid:

This doesn't pass the [Swagger Editor](https://editor.swagger.io/) validation due to some missing required fields - but it also seems there are a lot of auto-generated fields that we don't need either (e.g. `summary`, `operationId`, `x-internal`, ...) - all of which are optional.

Could we simplify but also meet the minimum requirements for a valid spec? Even if the descriptions are just left as "TODO" (which is easier to find/replace than `a`)

bbrks (approved):
This looks like a great start!

Would at some point still like to investigate how to group or separate the Public/Admin/Metric APIs differently so we're still able to group by functional area, even if that means maintaining 3 specs with references to shared resources - but we can always do that later.

PR #5408, opened 2022-01-20
CBG-1881: Added users and roles

IsaacLambat:
CBG-1881

Added user and role endpoints to the API documentation

### Endpoints
```
GET HEAD POST /{db}/_user/
GET HEAD PUT DELETE /{db}/_user/{name}
GET HEAD POST /{db}/_role/
GET HEAD PUT DELETE /{db}/_role/{name}
```

### DIFF
--- a/docs/api/api_docs.yaml
+++ b/docs/api/api_docs.yaml
@@ -520,53 +520,97 @@ paths:
     get:
       responses:
         '200':
-          description: OK
+          description: Users retrieved successfully
+          content:
+            application/json:
+              schema:
+                type: array
+                minItems: 0
+                uniqueItems: true
+                description: List of all user names
+                items:
+                  type: string
+              examples:
+                Example:
+                  value:
+                    - Alice
+                    - Bob
+        '404':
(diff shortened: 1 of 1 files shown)


docs/api/api_docs.yaml
```diff
@@ -520,53 +520,97 @@ paths:
...
+                Example:
+                  value:
+                    - Alice
+                    - Bob
+        '404':
+          $ref: '#/components/responses/Not-found'
+      tags:
+        - Admin
+      summary: Get all the names of the users
+      description: Retrieves all the names of the users that are in the database. The result length is limited by the query pagination limit that's set in the database config.
```
bbrks: Are you sure this is the case? It looks like the _code_ is iterating over paginated queries to build a full set of results, but users of the API don't need to know that, and the pagination is not exposed via the API.

bbrks (changes requested):
Just a nitpick about the pagination description - otherwise looks great.

bbrks (approved):

PR #5410, opened 2022-01-26
CBG-1883: Added docs for document endpoints

IsaacLambat:
CBG-1883

Added documentation for endpoints:
```
POST /{db}/
GET HEAD /{db}/{docid}
PUT /{db}/{docid}
DELETE /{db}/{docid}

GET HEAD /{db}/_raw/{docid}
POST /{db}/_purge

GET HEAD /{db}/_local/{docid}
PUT /{db}/_local/{docid}
DELETE /{db}/_local/{docid}

GET POST HEAD /{db}/_all_docs
POST /{db}/_bulk_docs
POST /{db}/_bulk_get
GET POST HEAD /{db}/_changes

GET HEAD /{db}/_design/{ddoc}
PUT /{db}/_design/{ddoc}
DELETE /{db}/_design/{ddoc}
GET /{db}/_design/{ddoc}/_view/{view}

GET /{db}/_revtree/{docid}
```

PR #5413, opened 2022-01-27
CBG-1884: Documented attachment endpoints

IsaacLambat:
CBG-1884

Endpoints documented:
`GET HEAD PUT /{db}/{docid}/{attach}`

## Dependencies
- [x] Merge after #5410

### DIFF
--- a/docs/api/api_docs.yaml
+++ b/docs/api/api_docs.yaml
@@ -1018,7 +1018,8 @@ paths:
           schema:
             type: string
           in: header
-          description: 'RFC-2616 bytes range header. For example, bytes=123-456.'
+          description: 'RFC-2616 bytes range header.'
+          example: 'bytes=123-456'
         - name: meta
           schema:
             type: boolean
--- a/docs/api/api_docs.yaml
+++ b/docs/api/api_docs.yaml
@@ -982,6 +982,10 @@ paths:
         '200':
           description: Found attachment successfully.
           headers:
+            Content-Length:
+              schema:
+                type: number
+                description: Length of the attachment in bytes
             Etag:
               schema:
                 type: string
@@ -999,17 +1003,28 @@ paths:
       description: |-
         This request retrieves a file attachment associated with the document. 
 
-        The raw data of the associated attachment is returned (just as if you were accessing a static file). The `Content-Type` response header is the same content type set when the document attachment was added to the database. The `Content-Disposition` response header will be set instead if the content type is vulnerable to a phishing attack, causing the browser to download the file instead of display it (unless overriden by by database config option `serve_insecure_attachment_types`).
+        The raw data of the associated attachment is returned (just as if you were accessing a static file). The `Content-Type` response header is the same content type set when the document attachment was added to the database. The `Content-Disposition` response header will be set if the content type is considered unsafe to display in a browser (unless overriden by by database config option `serve_insecure_attachment_types`) which will force the attachment to be downloaded.
 
         If the `meta` query parameter is set then the response will be in JSON with the additional metadata tags.
       parameters:
         - $ref: '#/components/parameters/rev'
-        - schema:
+        - name: content_encoding
+          schema:
+            type: boolean
+            default: 'true'
+          in: query
+          description: 'Set to false to disable the `Content-Encoding` response header.'
+        - name: Range
+          schema:
+            type: string
+          in: header
+          description: 'RFC-2616 bytes range header. For example, bytes=123-456.'
+        - name: meta
+          schema:
             type: boolean
             default: 'false'
           in: query
-          name: meta
-          description: Include the metadata of the attachment in the response body. Will cause `Content-Type=application/json` in the response header.
+          description: Return only the metadata of the attachment in the response body.
     put:
       responses:
         '201':
@@ -1034,11 +1049,11 @@ paths:
       description: |-
         This request adds or updates an attachment associated with the document. If the document does not exist, it will be created and the attachment will be added to it.
 
-        The maximum content size of an attachment is 20MB. The `Content-Type` header of the request should be what the content type of the attachment is otherwise this will default to `application/octet-stream`.
+        If the attachment already exists, the data of the existing attachment will be replaced in the new revision.
 
-        If the attachment already exists, the data of the existing attachment will be replaced in the new revision. The revisions are for tracking document changes, not attachment changes.
+        The maximum content size of an attachment is 20MB. The `Content-Type` header of the request specifies the content type of the attachment.
 
-        To remove an attachment from a document, simply update the `_attachments` dictionary of the document in the `PUT /{db}/{docid}` request.
+        To remove an attachment from a document, omit the attachment in the `_attachments` object of the document in the `PUT /{db}/{docid}` request.
       parameters:
         - name: Content-Type
           schema:
@@ -1050,12 +1065,12 @@ paths:
           schema:
             type: string
           in: query
-          description: The document revision to add/modify the attachment on. Required only when modifying an existing document. This falls back on the `If-Match` header if not set.
+          description: The existing document revision ID to modify. Required only when modifying an existing document.
         - name: If-Match
           schema:
             type: string
           in: header
-          description: Used as the `rev` query parameter if it is not set.
+          description: An alternative way of specifying the document revision ID.
       requestBody:
         content:
           Attachment content type:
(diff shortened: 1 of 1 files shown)


docs/api/api_docs.yaml
```diff
@@ -976,27 +976,111 @@ paths:
...
+        '416':
+          description: Requested range exceeds content length
       tags:
         - Admin
         - Public
+      summary: Get an attachment from a document
+      description: |-
+        This request retrieves a file attachment associated with the document. 
+
+        The raw data of the associated attachment is returned (just as if you were accessing a static file). The `Content-Type` response header is the same content type set when the document attachment was added to the database. The `Content-Disposition` response header will be set instead if the content type is vulnerable to a phishing attack, causing the browser to download the file instead of display it (unless overriden by by database config option `serve_insecure_attachment_types`).
```
```suggestion
        The raw data of the associated attachment is returned (just as if you were accessing a static file). The `Content-Type` response header is the same content type set when the document attachment was added to the database. The `Content-Disposition` response header to force attachment downloads will be set if the content type is considered unsafe to display in a browser (unless overriden by by database config option `serve_insecure_attachment_types`).
```

docs/api/api_docs.yaml
```diff
@@ -976,27 +976,111 @@ paths:
...
+
+        If the `meta` query parameter is set then the response will be in JSON with the additional metadata tags.
+      parameters:
+        - $ref: '#/components/parameters/rev'
+        - schema:
+            type: boolean
+            default: 'false'
+          in: query
+          name: meta
+          description: Include the metadata of the attachment in the response body. Will cause `Content-Type=application/json` in the response header.
```
```suggestion
          description: Return only the metadata of the attachment in the response body.
```

docs/api/api_docs.yaml
```diff
@@ -976,27 +976,111 @@ paths:
...
         - Public
+      summary: Create or update an attachment on a document
+      description: |-
+        This request adds or updates an attachment associated with the document. If the document does not exist, it will be created and the attachment will be added to it.
+
+        The maximum content size of an attachment is 20MB. The `Content-Type` header of the request should be what the content type of the attachment is otherwise this will default to `application/octet-stream`.
+
+        If the attachment already exists, the data of the existing attachment will be replaced in the new revision. The revisions are for tracking document changes, not attachment changes.
+
+        To remove an attachment from a document, simply update the `_attachments` dictionary of the document in the `PUT /{db}/{docid}` request.
```
```suggestion
        To remove an attachment from a document, omit the attachment in the `_attachments` object of the document in the `PUT /{db}/{docid}` request.
```

docs/api/api_docs.yaml
```diff
@@ -976,27 +976,111 @@ paths:
...
       tags:
         - Admin
         - Public
+      summary: Create or update an attachment on a document
+      description: |-
+        This request adds or updates an attachment associated with the document. If the document does not exist, it will be created and the attachment will be added to it.
+
+        The maximum content size of an attachment is 20MB. The `Content-Type` header of the request should be what the content type of the attachment is otherwise this will default to `application/octet-stream`.
+
+        If the attachment already exists, the data of the existing attachment will be replaced in the new revision. The revisions are for tracking document changes, not attachment changes.
```
bbrks: Swap these two paragraps around to keep the high-level description at the top. Specifics around headers/size limits further down.

bbrks: > The revisions are for tracking document changes, not attachment changes.

I'm not really sure what this means. Attachment changes are included as part of a document change.

docs/api/api_docs.yaml
```diff
@@ -976,27 +976,111 @@ paths:
...
+        '409':
+          $ref: '#/components/responses/Conflict'
       tags:
         - Admin
         - Public
+      summary: Create or update an attachment on a document
+      description: |-
+        This request adds or updates an attachment associated with the document. If the document does not exist, it will be created and the attachment will be added to it.
+
+        The maximum content size of an attachment is 20MB. The `Content-Type` header of the request should be what the content type of the attachment is otherwise this will default to `application/octet-stream`.
```
```suggestion
        The maximum content size of an attachment is 20MB. The `Content-Type` header of the request specifies the content type of the attachment.
```

Reword and default values are described already in the spec. I don't think there's a need to repeat them in a description.

docs/api/api_docs.yaml
```diff
@@ -976,27 +976,111 @@ paths:
...
+        - name: rev
+          schema:
+            type: string
+          in: query
+          description: The document revision to add/modify the attachment on. Required only when modifying an existing document. This falls back on the `If-Match` header if not set.
+        - name: If-Match
+          schema:
+            type: string
+          in: header
+          description: Used as the `rev` query parameter if it is not set.
```
bbrks: Wording of this is ambiguous and exposes some unnecessary implementation detail.

```suggestion
          description: The existing document revision ID to modify. Required only when modifying an existing document.
        - name: If-Match
          schema:
            type: string
          in: header
          description: An alternative way of specifying the document revision ID.
```

docs/api/api_docs.yaml
```diff
@@ -976,27 +976,111 @@ paths:
           type: string
         in: path
         required: true
+        description: 'The attachment name. This value must be URL encoded. For example, if the attachment name is `blob_/avatar`, the path component passed to the URL should be `blob_%2Favatar` (tested with [URLEncoder](https://www.urlencoder.org/)).'
     get:
       responses:
         '200':
-          description: OK
+          description: Found attachment successfully.
+          headers:
```
bbrks, line 984:
```suggestion
          headers:
            Content-Length:
              schema:
                type: string
                description: length of attachment in bytes
```

docs/api/api_docs.yaml
```diff
@@ -976,27 +976,111 @@ paths:
...
         - Public
+      summary: Get an attachment from a document
+      description: |-
+        This request retrieves a file attachment associated with the document. 
+
+        The raw data of the associated attachment is returned (just as if you were accessing a static file). The `Content-Type` response header is the same content type set when the document attachment was added to the database. The `Content-Disposition` response header will be set instead if the content type is vulnerable to a phishing attack, causing the browser to download the file instead of display it (unless overriden by by database config option `serve_insecure_attachment_types`).
+
+        If the `meta` query parameter is set then the response will be in JSON with the additional metadata tags.
+      parameters:
+        - $ref: '#/components/parameters/rev'
```
bbrks, line 1010: Missing request parameters

```suggestion
        - $ref: '#/components/parameters/rev'
        - schema:
            type: boolean
            default: 'true'
          in: query
          name: content_encoding
          description: Set to false to disable the `Content-Encoding` response header.
        - schema:
            type: string
          in: header
          name: Range
          description: 'RFC-2616 bytes range header. E.g: bytes=123-456'
```

docs/api/api_docs.yaml
```diff
@@ -976,27 +976,111 @@ paths:
...
+            schema:
+              description: The content to store in the body
+              type: string
+        description: The attachment data
     head:
       responses:
         '200':
-          description: OK
+          description: The document exists and the attachment exists on the document.
+          headers:
```
bbrks, line 1086: Content-Length

bbrks (changes requested):

docs/api/api_docs.yaml
IsaacLambat: I was trying to say that attachments don't get their own revision ID but rather the document does. Will remove this as it seems unnecessary.

docs/api/api_docs.yaml
```diff
@@ -982,6 +982,10 @@ paths:
         '200':
           description: Found attachment successfully.
           headers:
+            Content-Length:
+              schema:
+                type: number
+                description: Length of the attachment in bytes
```
bbrks: whoops, my suggestion was incorrectly indented

```suggestion
              description: Length of the attachment in bytes
```

docs/api/api_docs.yaml
```diff
@@ -999,17 +1003,28 @@ paths:
...
+          schema:
+            type: boolean
+            default: 'true'
+          in: query
+          description: 'Set to false to disable the `Content-Encoding` response header.'
+        - name: Range
+          schema:
+            type: string
+          in: header
+          description: 'RFC-2616 bytes range header. For example, bytes=123-456.'
```
bbrks: Again my bad w/ suggestion - I've seen the correct way of specifying an example:

```suggestion
          description: 'RFC-2616 bytes range header.'
          example: 'bytes=123-456'
```

bbrks (changes requested):
Feel free to just apply these suggestions in GitHub - they were my incorrect suggestions to begin with 😄

bbrks (approved):

PR #5416, opened 2022-01-28
CBG-1905: Include log context for many user-triggered processes

bbrks:
CBG-1905

- Earlier creation of HTTP request context, which is now created lazily via `h.ctx()`
	- Passed down via `Database.Ctx` for user-scoped operations that fall outside of a handler's scope
	- `Database.Ctx` is used to supply context to standalone components:
	  - Authenticator
	  - Changes feeds
	  - ChannelComputer/Views/Queries
	  - CRUD ops
	  - On-demand import
 - Exported `base.TestCtx(testing.TB)` to generate test-specific contexts for the above

## TODO/Future work
- Remove non-context log functions (CBG-1925)
- Pass context into Bucket APIs? gocb can take a context for per-op cancellation
  - Decided to postpone changes here until bucket API is revisited

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1562/

### DIFF
--- a/db/revtree.go
+++ b/db/revtree.go
@@ -10,6 +10,7 @@ package db
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"math"
@@ -125,7 +126,7 @@ func (tree *RevTree) UnmarshalJSON(inputjson []byte) (err error) {
 		return
 	}
 
-	//validate revTreeList revs, parents and channels lists are of equal length
+	// validate revTreeList revs, parents and channels lists are of equal length
 	if !(len(rep.Revs) == len(rep.Parents) && len(rep.Revs) == len(rep.Channels)) {
 		return errors.New("revtreelist data is invalid, revs/parents/channels counts are inconsistent")
 	}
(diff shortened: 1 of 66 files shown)


db/revtree.go
```diff
@@ -783,7 +784,7 @@ func encodeRevisions(revs []string) Revisions {
 		if i == 0 {
 			start = gen
 		} else if gen != start-i {
-			base.Warnf("Found gap in revision list. Expecting gen %v but got %v in %v", start-i, gen, revs)
+			base.WarnfCtx(ctx, "Found gap in revision list. Expecting gen %v but got %v in %v", start-i, gen, revs)
```
adamcfraser: I haven't done a complete trace, but it seems like there are several places we're forced to pass context because we end up here.  This particular warning I think would benefit more from docID than user context, since it's primarily informational and on its own doesn't indicate a particular issue.

adamcfraser (commented):
Generally looks good.  One question about usage in encodeRevisions, but otherwise this is a good first step.  

In some lower-level utility usages I wonder if it's preferable to move the logging up (and include the information in a returned error), but I think we can still start with this, and refactor things up as we touch that code.

db/revtree.go
bbrks: Agreed. Will swap context for docID and cleanup obsolete context in callers.

adamcfraser (approved):

PR #5418, opened 2022-01-31
CBG-1885: Added replication endpoints

IsaacLambat:
CBG-1885

Added replication endpoints to docs:
```
GET /{db}/_blipsync 

GET HEAD POST /{db}/_replication/
GET HEAD PUT DELETE /{db}/_replication/{replicationid}

GET HEAD /_replicationStatus/
GET HEAD PUT /_replicationStatus/{replicationid}
```

### DIFF
--- a/docs/api/api_docs.yaml
+++ b/docs/api/api_docs.yaml
@@ -1617,7 +1617,7 @@ paths:
               - cbl2
               - sgr2
           in: query
-          description: This is the client that is making the BLIP Sync request.
+          description: This is the client type that is making the BLIP Sync request. Used to control client-type specific replication behaviour.
   '/{targetdb}/':
     parameters:
       - name: targetdb
@@ -2043,7 +2043,9 @@ paths:
     get:
       responses:
         '200':
-          description: Retrieved replication configurations successfully
+          description: |-
+            Retrieved replication configurations successfully.
+            The `assigned_node` fields will end with `(local)` or `(non-local)` depending on if the replication is running on this Sync Gateway node.
           content:
             application/json:
               schema:
@@ -2054,9 +2056,7 @@ paths:
         - Admin
       summary: Get all replication configurations
       description: |-
-        This will retrieve all the defined replication configs as a key map.
-
-        The `assigned_node` fields will end with `(local)` or `(non-local)` depending on if the replication is local or not.
+        This will retrieve all database replication definitions.
     post:
       responses:
         '200':
@@ -2064,7 +2064,7 @@ paths:
         '201':
           $ref: '#/components/responses/Replicator-created'
         '400':
-          description: Bad Request. Likely cause is that replicator was not stopped before attempting update.
+          $ref: '#/components/responses/request-problem'
         '404':
           $ref: '#/components/responses/Not-found'
       tags:
@@ -2128,7 +2128,7 @@ paths:
         '200':
           description: Replication successfully deleted
         '400':
-          description: Bad Request
+          $ref: '#/components/responses/request-problem'
         '404':
           $ref: '#/components/responses/Not-found'
       tags:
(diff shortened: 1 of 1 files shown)


docs/api/api_docs.yaml
```diff
@@ -1597,11 +1597,27 @@ paths:
...
+      parameters:
+        - name: client
+          schema:
+            type: string
+            default: cbl2
+            enum:
+              - cbl2
+              - sgr2
+          in: query
+          description: This is the client that is making the BLIP Sync request.
```
```suggestion
          description: This is the client type that is making the BLIP Sync request. Used to control client-type specific replication behaviour.
```

docs/api/api_docs.yaml
```diff
@@ -2027,91 +2043,220 @@ paths:
...
+                $ref: '#/components/schemas/All-replications'
+        '404':
+          $ref: '#/components/responses/Not-found'
       tags:
         - Admin
+      summary: Get all replication configurations
+      description: |-
+        This will retrieve all the defined replication configs as a key map.
+
+        The `assigned_node` fields will end with `(local)` or `(non-local)` depending on if the replication is local or not.
```
bbrks: Defining local with local doesn't clarify much 😄  but also shouldn't this really be in the response schema description instead of the endpoint description?

```suggestion
        The `assigned_node` fields will end with `(local)` or `(non-local)` depending on if the replication is running on this sync gateway node.
```

docs/api/api_docs.yaml
```diff
@@ -2027,91 +2043,220 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/All-replications'
+        '404':
+          $ref: '#/components/responses/Not-found'
       tags:
         - Admin
+      summary: Get all replication configurations
+      description: |-
+        This will retrieve all the defined replication configs as a key map.
```
bbrks: Not just configs - returns things like "state". A map is also just a regular JSON object. I'm not sure it's worth calling out specifically.

```suggestion
        This will retrieve all database replication definitions.
```

docs/api/api_docs.yaml
```diff
@@ -2027,91 +2043,220 @@ paths:
...
+        The `assigned_node` fields will end with `(local)` or `(non-local)` depending on if the replication is local or not.
     post:
       responses:
         '200':
-          description: OK
+          $ref: '#/components/responses/Replicator-updated'
+        '201':
+          $ref: '#/components/responses/Replicator-created'
+        '400':
+          description: Bad Request. Likely cause is that replicator was not stopped before attempting update.
```
bbrks: Let's not speculate - and leave the reason for the message contained in the response body (which is already defined) as a response type.

```suggestion
          $ref: '#/components/responses/request-problem'
```

docs/api/api_docs.yaml
```diff
@@ -2027,91 +2043,220 @@ paths:
...
+        If an existing replication is being updated, that replication must be stopped first and, if the `replication_id` is specified in the request body, it must match the replication ID in the URI.
+      requestBody:
+        $ref: '#/components/requestBodies/Replication-upsert'
     delete:
       responses:
         '200':
-          description: OK
+          description: Replication successfully deleted
+        '400':
+          description: Bad Request
```
```suggestion
          $ref: '#/components/responses/request-problem'
```

docs/api/api_docs.yaml
```diff
@@ -2027,91 +2043,220 @@ paths:
...
+          $ref: '#/components/responses/Not-found'
       tags:
         - Admin
+      description: |-
+        Change the replication state to target. 
+
+        This is done through the action query parameter, which has 3 valid values:
+        * `start` - starts a stopped replication
+        * `stop` - stops an active replication
+        * `reset` - resets the replication checkpoint to 0. For bidirection replication, both push and pull checpoints are reset to 0. The replication must be stopped to use this.
```
```suggestion
        * `reset` - resets the replication checkpoint to 0. For bidirectional replication, both push and pull checkpoints are reset to 0. The replication must be stopped to use this.
```

docs/api/api_docs.yaml
```diff
@@ -2027,91 +2043,220 @@ paths:
...
+      parameters:
+        - name: action
+          schema:
+            type: string
+            enum:
+              - start
+              - stop
+              - reset
+          in: query
+          description: The target state to put the replicator in too.
```
```suggestion
          description: The target state to put the replicator into.
```

docs/api/api_docs.yaml
```diff
@@ -2027,91 +2043,220 @@ paths:
...
+          required: true
     head:
       responses:
         '200':
-          description: OK
+          description: Replication exists
+        '400':
+          description: Bad Request
```
```suggestion
          $ref: '#/components/responses/request-problem'
```

bbrks (changes requested):
Mostly looks good - lots of detail in the schemas!

bbrks (approved):

PR #5423, opened 2022-02-05
CBG-1910 Avoid retrieval of temporary revision backups during ancestor lookup

adamcfraser:
When attempting to retrieve an ancestor for use as oldDoc in the sync function, we shouldn't check for temporary revision body backups.  These aren't intended for use by the sync function (as they would provide different results depending on whether they have expired), and the check adds unnecessary performance overhead at write time when the ancestor is unavailable or not the immediate parent.

CBG-1910

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1574/

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -605,29 +605,14 @@ func extractInlineAttachments(bodyBytes []byte) (attachments AttachmentsMeta, cl
 	return attsMap, bodyBytes, body, nil
 }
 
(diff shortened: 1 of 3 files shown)


adamcfraser:
I can't reproduce any of the integration test failures locally - assuming they are unrelated.

bbrks (approved):
LGTM

PR #5424, opened 2022-02-07
Reduce RestTester bcrypt cost

bbrks:
Minor (~100ms per test) speedup of `RestTester` tests by reducing the cost of bcrypt hashing

### DIFF
--- a/rest/utilities_testing.go
+++ b/rest/utilities_testing.go
@@ -159,7 +159,7 @@ func (rt *RestTester) Bucket() base.Bucket {
 		panic("invalid RestTester StartupConfig: " + err.Error())
 	}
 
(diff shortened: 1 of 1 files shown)


rest/utilities_testing.go
```diff
@@ -158,6 +159,9 @@ func (rt *RestTester) Bucket() base.Bucket {
 		panic("invalid RestTester StartupConfig: " + err.Error())
 	}
 
+	// Post-validation, we can lower the bcrypt cost to reduce test runtime.
```
bbrks:
```suggestion
	// Post-validation, we can lower the bcrypt cost beyond SG limits to reduce test runtime.
```

adamcfraser (approved):

PR #5429, opened 2022-02-10
CBG-1874: Deprecated 'allow-conflicts=true'

IsaacLambat:
CBG-1874

Deprecated the `allow-conflicts=true` database option. Tested manually.

- [x] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1585

### DIFF
--- a/docs/api/api_docs.yaml
+++ b/docs/api/api_docs.yaml
@@ -5646,6 +5646,7 @@ components:
         allow_conflicts:
           type: boolean
           default: true
+          deprecated: true
           description: This controls whether to allow conflicting document revisions.
         num_index_replicas:
           type: number
--- a/rest/config.go
+++ b/rest/config.go
@@ -694,10 +694,6 @@ func (dbConfig *DbConfig) validateVersion(isEnterpriseEdition bool) error {
 		}
 	}
 
-	if dbConfig.AllowConflicts != nil && *dbConfig.AllowConflicts {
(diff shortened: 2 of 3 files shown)


docs/api/api_docs.yaml
```diff
@@ -5646,6 +5646,7 @@ components:
         allow_conflicts:
           type: boolean
           default: true
+          deprecated: true
```
bbrks, line 5649: This is **_so_** nice to see being used like this already!

bbrks (approved):

rest/config.go
```diff
@@ -694,6 +694,10 @@ func (dbConfig *DbConfig) validateVersion(isEnterpriseEdition bool) error {
 		}
 	}
 
+	if dbConfig.AllowConflicts != nil && *dbConfig.AllowConflicts {
+		base.Warnf(`Deprecation notice: Setting the database configuration option "allow_conflicts" to true is due to be removed. In the future, conflicts will not be allowed.`)
+	}
+
```
bbrks: On 2nd thought - maybe we should be logging this warning on database startup instead (maybe in `dbcOptionsFromConfig`?), rather than when the option is set. Which is presumably only going to be once, and only on a single node.

That way customers get ongoing notices of the deprecated option in use.

bbrks (changes requested):
Had a change of mind - see inline comment

bbrks (approved):

PR #5432, opened 2022-02-14
Pipeline job tweaked for new test infrastructure

JRascagneres:
This job modifies the Jenkinsfile to work with our new testing infrastructure, main aspects just brings dependencies into Jenkins rather than having them installed on the box in prefined locations. The tweaks now mean Go is installed through jenkins and that the SSH keys are stored through jenkins.


bbrks (dismissed):
LGTM

PR #5434, opened 2022-02-14
CBG-1869: Flaky TestLogFlush: expected 4 files, but got 5

cbbruno:
CBG-1869

- Not able to replicate.
- added more info to log content, test directory name, to indicate what test run it might be coming from.
- added dump of the log file content in case the file count check fails.

- [ ] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-integration/1591/

### DIFF
--- a/base/util_testing.go
+++ b/base/util_testing.go
@@ -430,7 +430,7 @@ func setTestLogging(logLevel LogLevel, caller string, logKeys ...LogKey) (teardo
 		return func() {
 			if caller != "" {
 				Infof(KeyAll, "%v: Reset logging", caller)
-				FlushLogBuffers()
+				FlushLoggerBuffers()
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


base/util_testing.go
```diff
@@ -430,6 +430,7 @@ func setTestLogging(logLevel LogLevel, caller string, logKeys ...LogKey) (teardo
 		return func() {
 			if caller != "" {
 				Infof(KeyAll, "%v: Reset logging", caller)
+				FlushLoggerBuffers()
```
bbrks: AFAIK this `FlushLoggerBuffers` function is a special-case only for the buffered/memory loggers (which aren't initialized by tests, with the exception of `TestLogFlush`) - seems overkill to run it for all tests just to cover one case.

bbrks (approved):

PR #5446, opened 2022-02-18
Write a final test profile when the suite finishes

bbrks:
Uses same pattern to write a final test profile even if the profile frequency interval was not hit

https://jenkins.sgwdev.com/job/SyncGateway-Integration/111/


JRascagneres (approved):
Seems reasonable

PR #5447, opened 2022-02-18
CBG-1951: Automated master integration test and Jenkins script in repo

IsaacLambat:
CBG-1951

- Added the script that will be used for the Jenkins integration testing that will be ran when a commit is pushed to master, or when a manually integration test is started.
- The script takes the flag `-m` where a parameterless build can be started (used for the automatic master testing).
- Added Slack notifications when a master build fails or is unstable (failed test).
- If not using Go Modules, a folder will be made called `sgw_int_testing` which the bootstrapping process will take place in


jenkins-integration.sh
```diff
@@ -0,0 +1,162 @@
...
+set -e # Abort on errors
+set -x # Output all executed shell commands
+
+# Use Git SSH and define private repos
+git config --global --add url."git@github.com:".insteadOf "https://github.com/"
+export GOPRIVATE=github.com/couchbaselabs/go-fleecedelta
+
+# Build Sync Gateway with the right edition
+export SG_EDITION="${SG_EDITION}"
+
```
bbrks: Nitpick: Doesn't need exporting as the variable is only used within this bash script

```suggestion
```

jenkins-integration.sh
```diff
@@ -0,0 +1,162 @@
+#!/usr/bin/env bash
```
bbrks: If we add this script, we'll have both:

- `jenkins-integration-build.sh`
- `jenkins-integration.sh`

Which are both doing almost identical things. Is it possible to tweak that script to get the behaviour we need for parameterless master builds?

jenkins-integration.sh
```diff
@@ -0,0 +1,162 @@
...
+
+# Build Sync Gateway with the right edition
+export SG_EDITION="${SG_EDITION}"
+
+if [ ${USE_GO_MODULES} == "true" ]; then
+  cd sgw
+else
+  rm -rf sgw
+  mkdir sgw 
+  cd sgw
```
bbrks: I appriciate this means job changes in Jenkins to fix, so maybe do as a follow-up PR instead:

I'm not a fan of this (cloning SG into a sub-directory in Jenkins), because it means the script cannot be used locally, and it doesn't give us any benefits on the Jenkins side either, by just having everything nested one level deeper.

jenkins-integration.sh
```diff
@@ -0,0 +1,162 @@
...
+go get -u -v github.com/tebeka/go2xunit
+go install -v github.com/tebeka/go2xunit
+go get -v -u github.com/axw/gocov/...
+go install -v github.com/axw/gocov/...
+go get -v -u github.com/AlekSi/gocov-xml
+go install -v github.com/AlekSi/gocov-xml
+
+# Set environment vars
+GO_TEST_FLAGS="-v -p 1 -count=${RUN_COUNT}"
+export INT_LOG_FILE_NAME="int_logs_${SG_EDITION}_${SG_COMMIT}_${TARGET_TEST}"
```
bbrks: Nitpick:

```suggestion
INT_LOG_FILE_NAME="int_logs_${SG_EDITION}_${SG_COMMIT}_${TARGET_TEST}"
```

jenkins-integration.sh
```diff
@@ -0,0 +1,162 @@
...
+
+if [ -d "godeps" ]; then
+  export GOPATH=`pwd`/godeps
+fi
+export PATH=$PATH:`go env GOPATH`/bin
+echo "PATH: $PATH"
+
+if [ ${TEST_DEBUG} == "true" ]; then
```
bbrks: This doesn't seem to work. I'm guessing you need to quote TEST_DEBUG?

```10:43:34 ./sgw/jenkins-integration.sh: line 69: [: ==: unary operator expected```

File: jenkins-integration.sh [hunk 6]
```diff
@@ -0,0 +1,162 @@
+#!/usr/bin/env bash
+DEFAULT_PACKAGE_TIMEOUT="45m"
+
+if [ $1 == "-m" ]; then
+    echo "Running in automated master integration mode"
+    # Set automated setting parameters
+    SG_COMMIT="master"
+    TARGET_PACKAGE="..."
+    TARGET_TEST="ALL"
+    RUN_WARLUS="true"
```
bbrks: typo causing the jobs not to run the walrus tests

```suggestion
    RUN_WALRUS="true"
```

jenkins-integration.sh
```diff
@@ -0,0 +1,162 @@
...
+if [ "${SG_EDITION}" == "EE" ]; then
+    GO_TEST_FLAGS="${GO_TEST_FLAGS} -tags cb_sg_enterprise"
+fi
+
+go test ${GO_TEST_FLAGS} -coverprofile=coverage_int.out -coverpkg=github.com/couchbase/sync_gateway/${TARGET_PACKAGE} github.com/couchbase/sync_gateway/${TARGET_PACKAGE} 2>&1 | tee "${INT_LOG_FILE_NAME}.txt" | grep -E '(--- (FAIL|PASS|SKIP):|github.com/couchbase/sync_gateway(/.+)?\t|TEST: |panic: )'
+if [ "${PIPESTATUS[0]}" -ne "0" ]; then # If test exit code is not 0 (failed)
+  echo "Go test failed! Parsing logs to find cause..."
+  TEST_FAILED=true
+fi
+
```
bbrks: Missing in this script vs. what we used to run for integration tests:

```
# Strip non-printable characters
LC_CTYPE=C tr -dc [:print:][:space:] < verbose_unit_ee.out.raw > verbose_unit_ee.out
LC_CTYPE=C tr -dc [:print:][:space:] < verbose_unit_ce.out.raw > verbose_unit_ce.out
LC_CTYPE=C tr -dc [:print:][:space:] < verbose_int.out.raw > verbose_int.out
```

This was used because some of our test logging can in some cases emit non-printable/non-valid UTF-8 characters from logging binary attachment data, which caused issues with creating valid XML.

jenkins-integration.sh
```diff
@@ -0,0 +1,162 @@
...
+# Generate xunit test report that can be parsed by the JUnit Plugin
+~/go/bin/go2xunit -input "${INT_LOG_FILE_NAME}.txt" -output "${INT_LOG_FILE_NAME}.xml"
```
bbrks: What about the test report for the walrus runs?

jenkins-integration.sh
```diff
@@ -0,0 +1,162 @@
...
+if [ "${TEST_FAILED}" = true ] ; then
+  # If output contained `FAIL:`
+  if grep -q 'FAIL:' "${INT_LOG_FILE_NAME}.txt"; then
+    # Test failure, so mark as unstable
+    exit 50
+  else
+    # Test setup failure or unrecovered panic, so mark as failed
+    exit 1
+  fi
+fi
```
bbrks: Nitpick: eof newline

```suggestion
fi

```

bbrks (changes requested):

jenkins-integration-build.sh
```diff
@@ -1,179 +1,164 @@
...
+set -x # Output all executed shell commands
+
+# Use Git SSH and define private repos
+git config --global --add url."git@github.com:".insteadOf "https://github.com/"
+export GOPRIVATE=github.com/couchbaselabs/go-fleecedelta
+
+# Use Go modules (3.1 and above) or bootstrap for legacy Sync Gateway versions (3.0 and below)
+if [ ${USE_GO_MODULES} == "false" ]; then
+  rm -rf *
+  wget https://raw.githubusercontent.com/couchbase/sync_gateway/$SG_COMMIT/bootstrap.sh
```
bbrks: Are these 2 steps really nessesary? They still basically prevent the script from being run locally in an existing sg repo, and AFAIK aren't required in Jeknins - because the job will already have a cloned copy of the repo at `$SG_COMMIT`

bbrks (changes requested):
I have a question about how the script bootstraps below

bbrks (approved):

PR #5461, opened 2022-03-08
CBG-1969: Override revpos when client erroneously doesn't update it

JRascagneres:
CBG-1969

Prior versions of CBL can end up in a situation where, if an attachment is updated, the revpos is not incremented, meaning that it will remain the same as the prior version of the attachment. This means that Sync Gateway never gets or knows about the updated attachment and therefore the data is not stored on CBS or replicated to other clients.

This PR aims to work around this issue by checking for the erroneous revpos and overriding it. 
- Obtains current bucket doc to compare against (if available)
- Checks revpos and digest for the current and incoming document
- If revpos is same but digest is different update revpos to latest revpos so that we know to request the new data

(Also moved the len(history) > 1 check inside the attachment section as its the only place it was used anyway.

This is tested by a unit test. It creates an attachment on SGW and pulls the attachment down to the client. It then updates this attachment and pushes up purposfully incorrect attachment meta --> An updated digest but with the old name and with the old revpos. A GET to SGW is then issued to ensure that it does indeed have the updated attachment.

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -906,26 +906,87 @@ func (bh *blipHandler) handleRev(rq *blip.Message) (err error) {
 
 	var rawBucketDoc *sgbucket.BucketDocument
 
-	// Look at attachments with revpos > the last common ancestor's
-	minRevpos := 1
-	if len(history) > 0 {
-		currentDoc, rawDoc, err := bh.db.GetDocumentWithRaw(bh.loggingCtx, docID, DocUnmarshalSync)
-		// If we're able to obtain current doc data then we should use the common ancestor generation++ for min revpos
-		// as we will already have any attachments on the common ancestor so don't need to ask for them.
-		// Otherwise we'll have to go as far back as we can in the doc history and choose the last entry in there.
-		if err == nil {
-			commonAncestor := currentDoc.History.findAncestorFromSet(currentDoc.CurrentRev, history)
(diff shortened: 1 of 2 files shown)


db/blip_handler.go
```diff
@@ -906,27 +906,88 @@ func (bh *blipHandler) handleRev(rq *blip.Message) (err error) {
...
+
+				currentAttachmentRevpos, ok := base.ToInt64(currentAttachmentMeta["revpos"])
+				if !ok {
+					return base.HTTPErrorf(http.StatusInternalServerError, "Current attachment data is invalid")
+				}
+
+				// Compare the revpos and attachment digest. If revpos is same but attachment digest is different we
+				// need to override the revpos and set it to the current revision as the incoming revpos must be invalid
+				// and we need to request it.
+				if currentAttachmentRevpos == incomingAttachmentRevpos && currentAttachmentDigest != incomingAttachmentDigest {
```
adamcfraser: I think we can do incomingAttachmentRevpos <= minRevPos here instead of the first condition (although should double-check whether it's <= or <)

adamcfraser (changes requested):

db/blip_handler.go
JRascagneres: Switched out. Pretty sure we want the inverse of the existing check so yeah <= works.

adamcfraser (approved):

PR #5462, opened 2022-03-09
CBG-1971: Added additional logging and upped logging level

IsaacLambat:
CBG-1971

- Upped logging level to trace
- Fixed log line reporting wrong variable
- Added additional logging

### DIFF
--- a/rest/replication_api_test.go
+++ b/rest/replication_api_test.go
@@ -940,7 +940,7 @@ func TestPullOneshotReplicationAPI(t *testing.T) {
 func TestReplicationConcurrentPush(t *testing.T) {
 
 	base.RequireNumTestBuckets(t, 2)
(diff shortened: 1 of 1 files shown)


JRascagneres (commented):
I'm fine with the logging fix and additional logging but is trace necessary?
Does it get us any more info for replications that'd be helpful? Just a bit warey of whacking that up

IsaacLambat:
> I'm fine with the logging fix and additional logging but is trace necessary? Does it get us any more info for replications that'd be helpful? Just a bit warey of whacking that up

I understand that it makes a lot more logs (as a lot of lines about the checkpointer get printed). The reason I bumped it up is that it logs when it logs lots of replication checkpoint information which could be linked to the `Replicate+: couldn't find existing local checkpoint for ID "sgr2cp:push:rep_DEF"` debug errors seen in the logs which might be related to the flakiness (as this does not appear when the test is successful). 

Can change back to debug or even info straight after the test fails the same way unexpectedly or can change back to debug now but would be worried helpful information might be missed.

adamcfraser (approved):

PR #5464, opened 2022-03-09
Fixed description for "marked_attachment" on the Rest API specifications

IsaacLambat:
Modified the compact endpoint `marked_attachments` description to show that it does not purge marked attachments.

### DIFF
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -1524,7 +1524,7 @@ Compact-status:
       description: |-
         **Applicable to attachment compaction only**
 
-        This is the amount of attachments that have been marked to be kept.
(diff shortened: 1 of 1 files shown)


docs/api/components/schemas.yaml
```diff
@@ -1524,7 +1524,7 @@ Compact-status:
       description: |-
         **Applicable to attachment compaction only**
 
-        This is the amount of attachments that have been marked to be purged.
+        This is the amount of attachments that have been marked to be kept.
```
JRascagneres: Still not really accurate

```suggestion
        This is the number of references there are to legacy attachments. 
```

JRascagneres (changes requested):

JRascagneres (approved):

PR #5465, opened 2022-03-10
CBG-1963: TestConcurrentRefreshUser sends all revs before starting changes timeout timer

IsaacLambat:
CBG-1963

The timeout seemed to be hit due to the `bt.SendRev` call calling `func (request *Message) Response() *Message` in the `go-blip` package taking a while to return. 
So instead of starting the changes timeout timer before all the revs have been sent, it now does it after by making it single-threaded.

- Removed goroutine use meaning the changes timeout starts counting when all the documents revs have been sent
- Dropped timeout back to 30 seconds
- Added log information for how long it took to receive all changes

- [x] `xattrs=true count=20 test=^TestConcurrentRefreshUser` https://jenkins.sgwdev.com/job/SyncGateway-Integration/149/

### DIFF
--- a/rest/blip_api_test.go
+++ b/rest/blip_api_test.go
@@ -1005,7 +1005,7 @@ function(doc, oldDoc) {
 }
 
 // Start subChanges w/ continuous=true, batchsize=20
-// Start goroutine sending rev messages for documents that grant access to themselves for the active replication's user
+// Start sending rev messages for documents that grant access to themselves for the active replication's user
 
 func TestConcurrentRefreshUser(t *testing.T) {
 	defer base.SetUpTestLogging(base.LevelInfo, base.KeyHTTP, base.KeySync, base.KeySyncMsg, base.KeyChanges, base.KeyCache)()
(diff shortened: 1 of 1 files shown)


rest/blip_api_test.go
```diff
@@ -1005,7 +1005,7 @@ function(doc, oldDoc) {
 }
 
 // Start subChanges w/ continuous=true, batchsize=20
-// Start goroutine sending rev messages for documents that grant access to themselves for the active replication's user
```
bbrks, line 1008: The change to make the test single threaded seems to go against the whole point of the test (issuing concurrent grants to detect the race fixed by CBG-628) and I'm not sure solves the underlying timeout problem.

Is this the view performance issue you're currently investigating?

bbrks (dismissed):

rest/blip_api_test.go
IsaacLambat: I don't see how this change goes against the point of the test as a Goroutine is not being started for every `SendRev` and instead only wraps the loop that creates the revs in a goroutine. All this change does is mean that the timeout to receive the changes gets started once all the revs have been sent (as opposed to at the same time the loop starts to send the revs).

rest/blip_api_test.go
```diff
@@ -1114,31 +1114,29 @@ function(doc, oldDoc) {
 	subChangesResponse := subChangesRequest.Response()
 	goassert.Equals(t, subChangesResponse.SerialNumber(), subChangesRequest.SerialNumber())
 
-	// Start goroutine to simulate sending docs from the client
-
+	// Start goroutines to simulate sending docs from the client
```
IsaacLambat: Remove `goroutines` in comment

rest/blip_api_test.go
```diff
@@ -1005,7 +1005,7 @@ function(doc, oldDoc) {
 }
 
 // Start subChanges w/ continuous=true, batchsize=20
-// Start goroutine sending rev messages for documents that grant access to themselves for the active replication's user
+// Start sending rev messages for documents that grant access to themselves for the active replication's user
 
 func TestConcurrentRefreshUser(t *testing.T) {
```
IsaacLambat, line 1009: Add findings about performance.

bbrks (approved):

PR #5467, opened 2022-03-18
CBG-1959: Unset default console log level to restore dynamic level/key enabling

bbrks:
CBG-1959

- Unset default console log level to restore dynamic log level/key enabling
- Modify test to cover behaviour

## [Integration Tests](https://jenkins.sgwdev.com/job/SyncGateway-Integration/build?delay=0sec)
- n/a

PR #5468, opened 2022-03-18
Set console logging ColorEnabled value even when logger is disabled

bbrks:
Allows color to be set, even if the output is disabled. This enables `Consolef(...)` (which bypasses the enabled checks) to use color.

### Before
Screenshot 2022-03-18 at 14 05 06

### After
Screenshot 2022-03-18 at 14 05 09

## Dependencies
- [x] #5467 

- n/a


IsaacLambat (approved):
LGTM and works well

PR #5469, opened 2022-03-18
Add GitHub action to validate OpenAPI specs

bbrks:
Validates the OpenAPI specs when modified. Caught some existing failures, will leave those to be fixed separately.

- The configuration specified in `.redocly.yaml` determines what rules are run and which API specs exist, and is automatically found by the tools below:
  - GH Actions runs the Redocly OpenAPI lint tool whenever a file changes in `docs/api/**`
  - You can also run locally with `openapi lint` installable here: https://github.com/Redocly/openapi-cli#node

Screenshot 2022-03-18 at 16 44 39

## Dependencies
- [x] https://github.com/mhiew/redoc-lint-github-action/pull/2 

- n/a

### DIFF
--- a/.redocly.yaml
+++ b/.redocly.yaml
@@ -1,0 +1,26 @@
+# Copyright 2022-Present Couchbase, Inc.
+#
+# Use of this software is governed by the Business Source License included in
+# the file licenses/BSL-Couchbase.txt.  As of the Change Date specified in that
+# file, in accordance with the Business Source License, use of this software
+# will be governed by the Apache License, Version 2.0, included in the file
+# licenses/APL2.txt.
+
+# This configuration file is read by the Redocly `openapi` CLI tool to validate/lint our specs (from GitHub Actions)
+
+apiDefinitions:
+  admin: ./docs/api/admin.yaml
+  public: ./docs/api/public.yaml
+  metric: ./docs/api/metric.yaml
+
+lint:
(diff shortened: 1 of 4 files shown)


.redocly.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+apiDefinitions:
+  admin: ./docs/api/admin.yaml
+  public: ./docs/api/public.yaml
+  metric: ./docs/api/metric.yaml
+
+lint:
+  extends:
+    - minimal
+  rules:
```
IsaacLambat, line 19: Would be good to allow examples to have additional properties that are not defined in the schema to stop some of the warnings (due to doc bodies having user defined properties): https://redocly.com/docs/cli/resources/built-in-rules/#no-invalid-schema-examples

IsaacLambat (changes requested):

.redocly.yaml
bbrks: When I briefly looked a couple of weeks ago, I think I saw an in-spec way to describe additional user-defined fields in the schema, which seems like the more correct approach.

IsaacLambat (approved):
LGTM

.redocly.yaml
bbrks: IIRC it was using the `additionalProperties` field

PR #5475, opened 2022-03-21
CBG-1981: Removed password change from TestReplicatorRevocationsMultipleAlternateAccess

IsaacLambat:
CBG-1981

- ~~Added trace logging to `TestReplicatorRevocationsMultipleAlternateAccess`~~ Changed to info due to reproing with trace
- Removed unnecessary password change happening in test

### DIFF
--- a/rest/replicator_test.go
+++ b/rest/replicator_test.go
@@ -5126,6 +5126,7 @@ func TestReplicatorRevocationsNoRevButAlternateAccess(t *testing.T) {
 }
 
 func TestReplicatorRevocationsMultipleAlternateAccess(t *testing.T) {
+	defer base.SetUpTestLogging(base.LevelTrace, base.KeyAll)()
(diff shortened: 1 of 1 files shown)


JRascagneres:
Interesting, guessing the password change thing was the cause?

JRascagneres (approved):

bbrks:
Was this test not _creating_ a user on `rt2` for the replicator to use?

IsaacLambat:
> Was this test not _creating_ a user on `rt2` for the replicator to use?

The test creates the user the replicator uses in the `initScenario` function

bbrks (changes requested):
Test seems to be failing more in GH actions with this change

IsaacLambat:
Closing PR due to more time needed to investigate the root cause of the problem for whoever gets the ticket and to investigate why removing the password change causes the test to fail more often.

PR #5488, opened 2022-03-30
CBG-1995: Added support for top-level properties with an underscore prefix

IsaacLambat:
CBG-1995

- Documents can now be created with properties that begin with an underscore.
- using the `_purged` property will continue to still return an error
- Modified existing tests to assert with the new behaviour
- Added 2 unit tests - one to test the Rest API and ISGR, and one to test importing a document that using an underscore prefix
- Rev ID that are generated now include user-set underscore parameters.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/166

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -868,8 +868,8 @@ func (db *Database) Put(docid string, body Body) (newRevID string, doc *Document
 		}
 
 		// Make up a new _rev, and add it to the history:
-		bodyWithoutSpecialProps, wasStripped := stripSpecialProperties(body)
-		canonicalBytesForRevID, err := base.JSONMarshalCanonical(bodyWithoutSpecialProps)
+		bodyWithoutInternalProps, wasStripped := stripInternalProperties(body)
+		canonicalBytesForRevID, err := base.JSONMarshalCanonical(bodyWithoutInternalProps)
 		if err != nil {
 			return nil, nil, false, nil, err
 		}
--- a/db/crud.go
+++ b/db/crud.go
@@ -1316,9 +1316,9 @@ func validateNewBody(body Body) error {
 		return base.HTTPErrorf(http.StatusNotFound, "Document revision is not accessible")
 	}
 
-	// Reject bodies containing user special properties for compatibility with CouchDB
-	if containsUserSpecialProperties(body) {
-		return base.HTTPErrorf(400, "user defined top level properties beginning with '_' are not allowed in document body")
+	// Reject bodies that contains the "_purged" property.
+	if body[BodyPurged] != nil {
+		return base.HTTPErrorf(http.StatusBadRequest, "user defined top level property '_purged' is not allowed in document body")
 	}
 	return nil
 }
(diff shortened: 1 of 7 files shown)


db/database_test.go
```diff
@@ -1561,27 +1544,25 @@ func TestPostWithUserSpecialProperty(t *testing.T) {
...
+	rev2id, _, err := db.Put(docid, body)
+	assert.NoError(t, err)
 
-	// Test retrieval gets rev1
+	// Test retrieval gets rev2
 	doc, err = db.GetDocument(base.TestCtx(t), docid, DocUnmarshalAll)
-	goassert.True(t, doc != nil)
-	goassert.True(t, doc.CurrentRev == rev1id)
+	require.NotNil(t, doc)
+	assert.Equal(t, rev2id, doc.CurrentRev)
```
bbrks, line 1564: Is it worth adding a check that `_special` is in the retrieved doc body to ensure we're not stripping it instead of rejecting?

db/crud.go
```diff
@@ -1316,9 +1316,9 @@ func validateNewBody(body Body) error {
 		return base.HTTPErrorf(http.StatusNotFound, "Document revision is not accessible")
 	}
 
-	// Reject bodies containing user special properties for compatibility with CouchDB
-	if containsUserSpecialProperties(body) {
-		return base.HTTPErrorf(400, "user defined top level properties beginning with '_' are not allowed in document body")
+	// Reject bodies that contains the "_purged" property.
+	if body[BodyPurged] != nil {
```
adamcfraser, line 1320: As I recall, _purged is just used by Sync Gateway to identify documents that shouldn't be imported (when we need to create/delete a document to trigger tombstone compaction from indexes).  I see that BodyPurged wasn't included in the set of allowed properties in containsUserSpecialProperties, but I'm not sure that was intentional.  Are you thinking that the omission from containsUserSpecialProperties was providing an accidental guard against BodyPurged that needs to be maintained?

adamcfraser (commented):
Generally looks good - one clarification on the _purged handling.

db/crud.go
IsaacLambat: I did think that it was providing a guard against `BodyPurged` which is why I made the validation case. However since it was accidentally omitted from the `containsUserSpecialProperties` and it is only needed for import, I have removed the validation case. 
Import does read the `BodyPurged` and reject the document before it gets to the validation function.

bbrks: I am not sure allowing `_purged` to be a user-definable property makes sense, if SG is doing any special handling on it (even if that's just import).

Having it _allowed_ to be set by a user sounds error-prone if they use that field in one of their documents for another purpose and expects it to be imported.

bbrks (changes requested):
I don't think removing the validation of `_purged` is safe. See inline comment.

bbrks (dismissed):
LGTM - will leave it for @adamcfraser to review the updates on the `_purged` handling.

bbrks (approved):

PR #5495, opened 2022-04-04
CBG-1914: Improved handling for some dbconfig updates during a rolling upgrade 

bbrks:
May be easier to review by commit.

CBG-1914 - Prevents issues in mixed-version deployments where a config with new fields from a new SG version are applied to an older SG node and fails to do so.

- Minor refactor of version information pulled from build scripts (nessesary in order to get patch version information)
- Added a `ComparableVersion` type, which can be parsed from a `major.minor.patch` string (along with finer-grained information) that provides methods to compare versions.
- Persisted database configs store SG node version information on write
- When a database config is being loaded from the bucket as part of the background config update process, SG will warn and skip the update if the SG version on the config is newer than the node that is trying to load it.
  - E.g: `[WRN] Cannot apply config update from server for db "foo", this SG version is older than config's SG version (3.0.0 < 3.1.0-EE)`
- If the database config is being loaded from the bucket on initiali startup, no version check is applied to support downgrade use-cases.
- Older SG nodes are still allowed to overwrite a config file from a newer SG version to further support downgrade.

## Depends on
- [x] #5545 

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/258/

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -1286,6 +1286,24 @@ func (sc *ServerContext) applyConfigs(dbNameConfigs map[string]DatabaseConfig) (
 
 // _applyConfig loads the given database, failFast=true will not attempt to retry connecting/loading
 func (sc *ServerContext) _applyConfig(cnf DatabaseConfig, failFast bool) (applied bool, err error) {
+	// 3.0.0 doesn't write a SGVersion, but everything else will
+	configSGVersionStr := "3.0.0"
+	if cnf.SGVersion != "" {
+		configSGVersionStr = cnf.SGVersion
+	}
+
+	configSGVersion, err := base.NewComparableVersionFromString(configSGVersionStr)
+	if err != nil {
+		return false, err
+	}
+
+	// Skip applying if the config is from a newer SG version than this node
+	nodeSGVersion := base.ProductVersion
+	if nodeSGVersion.Less(configSGVersion) {
+		base.WarnfCtx(context.TODO(), "Cannot apply config update from server for db %q, this SG version is older than config's SG version (%s < %s)", cnf.Name, nodeSGVersion.String(), configSGVersion.String())
+		return false, nil
+	}
+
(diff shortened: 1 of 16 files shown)


rest/config.go
```diff
@@ -1225,6 +1225,24 @@ func (sc *ServerContext) applyConfigs(dbNameConfigs map[string]DatabaseConfig) (
...
+	}
+
+	configSGVersion, err := base.NewComparableVersionFromString(configSGVersionStr)
+	if err != nil {
+		return false, err
+	}
+
+	// Skip applying if the config is from a newer SG version than this node
+	nodeSGVersion := base.ProductVersion
+	if nodeSGVersion.Less(configSGVersion) {
```
adamcfraser: To confirm, this is not just the update case, but also means that databases updated by a newer version of Sync Gateway will not be loaded at all if the older Sync Gateway node is restarted, correct?

This seems like it prevents downgrade scenarios (customer hits an issue in 3.1.1 after updating the db config, can't roll back SG to 3.1.0).  Any thoughts on how customers would deal with that situation?

adamcfraser (changes requested):
Generally looks good, one question about downgrading Sync Gateway.

rest/config.go
bbrks: Ugh, you're right... This case was intended to just avoid the situation where a mixed-version cluster is running and receving newer config updates from server without falling over...

But yeah, on a restart/downgrade the node would still end up using this code-path to pull their config from the server on startup.

I think we can differentiate between initial startup and the continuous config updates, and try to apply the version to the node regardless of version checking in the initial startup case.

bbrks: Had to rebase, but the last commit contains all of the changes made to support this downgrade use-case (along with test coverage)

adamcfraser (dismissed):

adamcfraser (approved):

PR #5498, opened 2022-04-06
Fixed Go test failure annotations

IsaacLambat:
Fixed `golang-test-annotations` to annotate go test failures on Github, even if their was logging prior to the failure.


bbrks (approved):
LGTM - We can keep an eye on the upstream PR to see if/when it gets merged and swap the action back

PR #5499, opened 2022-04-06
Test fix: Leaked goroutine in TestDeleteDatabasePointingAtSameBucketPersistent

bbrks:
Not reading from an unbuffered `serverErr` means the send goroutine locks and leaks.


IsaacLambat (approved):
LGTM

PR #5500, opened 2022-04-07
GH Actions: Sort test logs by package, timestamp

bbrks:
Copes `go test`'s behaviour of buffering and printing ordered package logs, even for parallel tests.
Avoids interleaved logs from paralell test packages.


IsaacLambat (approved):
LGTM. Logs look completely sorted, and timestamps all seem to be in ascending order per package as expected.

PR #5502, opened 2022-04-12
CBG-2026: Add option to disable basic auth for the public API

markspolakovs:
CBG-2026

Adds a new config option, `disable_password_auth`, which enforces that only OIDC or guest authentication is used on the public interface.

Things to look out for when reviewing:
* Are there any other paths where authentication is performed that don't pass through `(*handler).checkAuth()`?
  * Cursory check suggests it should be all of them, as even things like `_blipsync` pass through `makeHandler`
* Have I updated all the documentation I should?
* Are there other things worth testing?

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/180/

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -123,32 +123,32 @@ type DatabaseContext struct {
 }
 
 type DatabaseContextOptions struct {
-	CacheOptions              *CacheOptions
-	RevisionCacheOptions      *RevisionCacheOptions
-	OldRevExpirySeconds       uint32
-	AdminInterface            *string
-	UnsupportedOptions        *UnsupportedOptions
-	OIDCOptions               *auth.OIDCOptions
-	DBOnlineCallback          DBOnlineCallback // Callback function to take the DB back online
-	ImportOptions             ImportOptions
-	EnableXattr               bool             // Use xattr for _sync
-	LocalDocExpirySecs        uint32           // The _local doc expiry time in seconds
-	SecureCookieOverride      bool             // Pass-through DBConfig.SecureCookieOverride
-	SessionCookieName         string           // Pass-through DbConfig.SessionCookieName
-	SessionCookieHttpOnly     bool             // Pass-through DbConfig.SessionCookieHTTPOnly
-	AllowConflicts            *bool            // False forbids creating conflicts
-	SendWWWAuthenticateHeader *bool            // False disables setting of 'WWW-Authenticate' header
-	DisablePublicBasicAuth    bool             // True enforces OIDC/guest only
-	UseViews                  bool             // Force use of views
-	DeltaSyncOptions          DeltaSyncOptions // Delta Sync Options
-	CompactInterval           uint32           // Interval in seconds between compaction is automatically ran - 0 means don't run
-	SGReplicateOptions        SGReplicateOptions
-	SlowQueryWarningThreshold time.Duration
-	QueryPaginationLimit      int    // Limit used for pagination of queries. If not set defaults to DefaultQueryPaginationLimit
-	UserXattrKey              string // Key of user xattr that will be accessible from the Sync Function. If empty the feature will be disabled.
-	ClientPartitionWindow     time.Duration
-	BcryptCost                int
-	GroupID                   string
+	CacheOptions                  *CacheOptions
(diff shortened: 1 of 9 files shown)


markspolakovs:
~~Integration test failure looks like https://issues.couchbase.com/browse/CBG-2015~~

db/database.go
```diff
@@ -138,6 +138,7 @@ type DatabaseContextOptions struct {
 	SessionCookieHttpOnly     bool             // Pass-through DbConfig.SessionCookieHTTPOnly
 	AllowConflicts            *bool            // False forbids creating conflicts
 	SendWWWAuthenticateHeader *bool            // False disables setting of 'WWW-Authenticate' header
+	DisablePublicBasicAuth    bool             // True enforces OIDC/guest only
```
bbrks: Have reviewed and looks good if the answers to the questions in Jira are both "No"

But will wait for clarification around guest/sessions first, as I suspect we'll need to cover those additional cases.

rest/oidc_test_provider_test.go
```diff
@@ -376,6 +376,106 @@ func TestOpenIDConnectTestProviderWithRealWorldToken(t *testing.T) {
...
+	response, err = http.DefaultClient.Do(request)
+	require.NoError(t, err, "Error sending request with bearer token")
+	checkGoodAuthResponse(t, response, "foo_noah")
+
+	// Now verify that we can perform a trivial request
+	request, err = http.NewRequest(http.MethodGet, mockSyncGatewayURL+"/"+restTester.DatabaseConfig.Name, nil)
+	require.NoError(t, err, "Error creating new request")
+	response, err = client.Do(request)
+	require.NoError(t, err, "Error sending request")
+	require.Equal(t, http.StatusOK, response.StatusCode)
```
bbrks, line 468: I think we need to close the response body (all above usages are being closed either in the test or in the assertion util functions)

```suggestion
	require.Equal(t, http.StatusOK, response.StatusCode)
	require.NoError(t, response.Body.Close(), "Error closing response body")
```

rest/oidc_test_provider_test.go
```diff
@@ -376,6 +376,106 @@ func TestOpenIDConnectTestProviderWithRealWorldToken(t *testing.T) {
...
+
+	// Finally, forget the session and check that we're back to getting a 401, but with no WWW-Authenticate
+	client.Jar = nil
+	request, err = http.NewRequest(http.MethodGet, mockSyncGatewayURL+"/"+restTester.DatabaseConfig.Name, nil)
+	require.NoError(t, err, "Error creating new request")
+	response, err = client.Do(request)
+	require.NoError(t, err, "Error sending request")
+	require.Equal(t, http.StatusUnauthorized, response.StatusCode)
```
bbrks, line 477: ditto

```suggestion
	require.Equal(t, http.StatusUnauthorized, response.StatusCode)
	require.NoError(t, response.Body.Close(), "Error closing response body")

```

rest/config.go
```diff
@@ -695,6 +696,13 @@ func (dbConfig *DbConfig) validateVersion(isEnterpriseEdition bool) error {
 		}
 	}
 
+	// Ensure that, if password auth is disabled, there's another way to authenticate
+	if dbConfig.DisablePasswordAuth {
+		if dbConfig.OIDCConfig == nil && (dbConfig.Guest == nil || base.BoolDefault(dbConfig.Guest.Disabled, true)) {
+			multiError = multiError.Append(fmt.Errorf("password authentication is disabled, but there is no alternative auth configuration (OIDC or guest access)"))
+		}
+	}
```
bbrks: There is one more authentication mechanism customers may use outside of OIDC or guest which means we can't do this check.
Admin-created sessions for users (i.e. `:4985/db/_session` -> `createUserSession()`). This might be hooked up to some external non-OIDC auth service that doles out sessions for users.

bbrks (changes requested):
Functionally looks good, although I think we need to relax the config validation - see inline comment, and just a couple of test tweaks.

bbrks (approved):
LGTM!

PR #5503, opened 2022-04-13
CBG-2023: Internal properties are now prevented from use

IsaacLambat:
CBG-2023

- Internal properties are prevented depending on if using blip, import, or the REST API.
- Intentionally left todo in code for future use

`_deleted` is not rejected from import.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/219/ `1 failed test now fixed`
- [x] `just run failed test that has been fixed` https://jenkins.sgwdev.com/job/SyncGateway-Integration/220/

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -768,17 +768,6 @@ type removalDocument struct {
 	Removed bool `json:"_removed"`
 }
 
-func validateBlipBody(body Body) error {
-	// Prevent disallowed internal properties from being used
-	disallowed := []string{base.SyncPropertyName, BodyId, BodyRev, BodyDeleted, BodyRevisions}
-	for _, prop := range disallowed {
-		if body[prop] != nil {
-			return base.HTTPErrorf(http.StatusBadRequest, "top-level property '"+prop+"' is a reserved internal property")
-		}
-	}
-	return nil
-}
-
 // Received a "rev" request, i.e. client is pushing a revision body
 func (bh *blipHandler) handleRev(rq *blip.Message) (err error) {
 	startTime := time.Now()
@@ -835,10 +824,6 @@ func (bh *blipHandler) handleRev(rq *blip.Message) (err error) {
 		RevID: revID,
 	}
 	newDoc.UpdateBodyBytes(bodyBytes)
-	err = validateBlipBody(newDoc.Body())
-	if err != nil {
-		return err
-	}
 
 	injectedAttachmentsForDelta := false
 	if deltaSrcRevID, isDelta := revMessage.DeltaSrc(); isDelta {
@@ -889,6 +874,11 @@ func (bh *blipHandler) handleRev(rq *blip.Message) (err error) {
 		newDoc.UpdateBody(deltaSrcMap)
 		base.TracefCtx(bh.loggingCtx, base.KeySync, "docID: %s - body after patching: %v", base.UD(docID), base.UD(deltaSrcMap))
 		bh.replicationStats.HandleRevDeltaRecvCount.Add(1)
+	}
+
+	err = validateBlipBody(bodyBytes, newDoc)
+	if err != nil {
+		return err
 	}
 
 	// Handle and pull out expiry
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -768,6 +768,25 @@ type removalDocument struct {
 	Removed bool `json:"_removed"`
 }
 
+func validateBlipBody(body Body) error {
+	if body[base.SyncPropertyName] != nil {
+		return base.HTTPErrorf(http.StatusBadRequest, "top level property '_sync' is a reserved internal property")
+	}
+	if body[BodyId] != nil {
+		return base.HTTPErrorf(http.StatusBadRequest, "top level property '_id' is a reserved internal property")
+	}
+	if body[BodyRev] != nil {
+		return base.HTTPErrorf(http.StatusBadRequest, "top level property '_rev' is a reserved internal property")
+	}
+	if body[BodyDeleted] != nil {
+		return base.HTTPErrorf(http.StatusBadRequest, "top level property '_deleted' is a reserved internal property")
+	}
+	if body[BodyRevisions] != nil {
+		return base.HTTPErrorf(http.StatusBadRequest, "top level property '_rev' is a reserved internal property")
+	}
+	return nil
+}
+
 // Received a "rev" request, i.e. client is pushing a revision body
 func (bh *blipHandler) handleRev(rq *blip.Message) (err error) {
 	startTime := time.Now()
@@ -824,6 +843,10 @@ func (bh *blipHandler) handleRev(rq *blip.Message) (err error) {
 		RevID: revID,
 	}
 	newDoc.UpdateBodyBytes(bodyBytes)
+	err = validateBlipBody(newDoc.Body())
+	if err != nil {
+		return err
+	}
 
 	injectedAttachmentsForDelta := false
 	if deltaSrcRevID, isDelta := revMessage.DeltaSrc(); isDelta {
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -773,7 +773,7 @@ func validateBlipBody(body Body) error {
 	disallowed := []string{base.SyncPropertyName, BodyId, BodyRev, BodyDeleted, BodyRevisions}
 	for _, prop := range disallowed {
 		if body[prop] != nil {
-			return base.HTTPErrorf(http.StatusBadRequest, "top level property '"+prop+"' is a reserved internal property")
+			return base.HTTPErrorf(http.StatusBadRequest, "top-level property '"+prop+"' is a reserved internal property")
 		}
 	}
 	return nil
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -769,20 +769,12 @@ type removalDocument struct {
 }
 
 func validateBlipBody(body Body) error {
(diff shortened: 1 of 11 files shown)


db/blip_handler.go
```diff
@@ -824,6 +835,10 @@ func (bh *blipHandler) handleRev(rq *blip.Message) (err error) {
 		RevID: revID,
 	}
 	newDoc.UpdateBodyBytes(bodyBytes)
+	err = validateBlipBody(newDoc.Body())
```
bbrks: This is going to be adding extra unmarshalling with `bodyBytes` via `newDoc.Body()` in all BLIP document writes, whether they contain an underscore property or not.

db/crud.go
```diff
@@ -1050,6 +1060,13 @@ func (db *Database) PutExistingRevWithBody(docid string, body Body, docHistory [
 
 }
 
+func validateDocUpdate(body Body) error {
```
bbrks: It's not clear from the function name why this is only chekcing `_sync` and not `_id`, `_rev`, etc.
Can you clarify by renaming the function to be specific to REST API body handling and state why those properties are allowed through in this case?

db/crud.go
```diff
@@ -1621,19 +1645,18 @@ func (db *Database) documentUpdateFunc(docExists bool, doc *Document, allowImpor
 		return
 	}
 
-	// Marshal raw user xattrs for use in Sync Fn. If this fails we can bail out so we should do early as possible.
-	metaMap, err := doc.GetMetaMap(db.Options.UserXattrKey)
+	err = validateNewBody(newDoc.Body())
```
bbrks: It's not obvious to me that newDoc.Body is already unmarshalled at this point.

It might still end up being more efficient to call GetDeepMutableBody and do the validation on that, since we know we have to do that work anyway for a valid doc. Doing it that way means it would only be 1 unmarshal instead of potentially 2.

db/import.go
```diff
@@ -443,6 +443,27 @@ func (db *Database) backupPreImportRevision(docid, revid string) error {
...
+
+	if isPurged, ok := body[BodyPurged].(bool); ok && isPurged {
+		return base.ErrImportCancelledPurged
+	}
+
+	// TODO: Validate attachment data to ensure user is not setting invalid attachments
+
+	if body == nil {
+		return base.ErrEmptyDocument
+	}
```
bbrks: Can't this be done at the very top as an early return?

db/blip_handler.go
```diff
@@ -768,6 +768,17 @@ type removalDocument struct {
...
+func validateBlipBody(body Body) error {
+	// Prevent disallowed internal properties from being used
+	disallowed := []string{base.SyncPropertyName, BodyId, BodyRev, BodyDeleted, BodyRevisions}
+	for _, prop := range disallowed {
+		if body[prop] != nil {
+			return base.HTTPErrorf(http.StatusBadRequest, "top-level property '"+prop+"' is a reserved internal property")
+		}
+	}
+	return nil
+}
```
bbrks: I'm not keen on having these validation functions spread throughout the codebase.
It makes it a little difficult to compare what properties are allowed in the 3 write code-paths. Could we move the functions to a common place?

db/crud.go
```diff
@@ -1040,6 +1045,11 @@ func (db *Database) PutExistingRevWithBody(docid string, body Body, docHistory [
 	delete(body, BodyAttachments)
 	newDoc.UpdateBody(body)
 
+	err = validateDocUpdate(body)
+	if err != nil {
+		return nil, "", err
+	}
```
bbrks: earlier exit

```suggestion
	err = validateDocUpdate(body)
	if err != nil {
		return nil, "", err
	}

	newDoc.UpdateBody(body)
```

rest/admin_api_test.go
```diff
@@ -4570,3 +4570,150 @@ func TestReplicatorCheckpointOnStop(t *testing.T) {
...
+			docBody:     `{"_id": "id_mismatch"}`,
+			docID:       "completely_different_id",
+			expectError: true,
+		},
+		{
+			name:        "Rev match",
+			docBody:     `{"_rev": "1-ca9ad22802b66f662ff171f226211d5c", "rand": "439870"}`,
+			rev:         "1-ca9ad22802b66f662ff171f226211d5c",
+			expectError: false,
+		},
```
bbrks: Is it worth adding test cases for either/or body/parameter? Or is there is sufficient coverage elsewhere for those cases?

bbrks (changes requested):
I have some concerns about unintentionally doing extra unmarshalling work with these changes. See inline comments

db/crud.go
IsaacLambat: Yep. It seems that it would have been unmarshalled at this point already in the REST API case, but not in the blip case (causing extra unmarshalling). Will change it back.

rest/admin_api_test.go
IsaacLambat: Didn't add these test cases as they are tested throughout the code however these are not what other tests are intending to test, so will add these test cases just to keep it specific to this change.

db/validation.go
```diff
@@ -0,0 +1,85 @@
...
+			return base.HTTPErrorf(http.StatusBadRequest, "top-level property '"+prop+"' is a reserved internal property therefore cannot be imported")
+		}
+	}
+	// TODO: Validate attachment data to ensure user is not setting invalid attachments
+
+	return nil
+}
+
+// validateBlipBody validates incoming blip rev bodies
+func validateBlipBody(rawBody []byte, doc *Document) error {
```
IsaacLambat, line 67: Takes in `rawBody` to save a call to `doc.BodyBytes()`

db/validation.go
```diff
@@ -0,0 +1,85 @@
...
+	// Prevent disallowed internal properties from being used
+	disallowed := []string{base.SyncPropertyName, BodyId, BodyRev, BodyDeleted, BodyRevisions}
+	for _, prop := range disallowed {
+		// Only unmarshal if raw body contains the disallowed property
+		if bytes.Contains(rawBody, []byte(prop)) {
+			if doc.Body()[prop] != nil {
```
bbrks: This would allow `"_rev": null`, which doesn't seem right. I'd suggest doing this instead so it doesn't matter what the value is.

```suggestion
			if _, ok := doc.Body()[prop]; ok {
```

db/validation.go
bbrks: Can you put this in the godoc comment so it's clear why the function takes both?

db/validation.go
```diff
@@ -0,0 +1,85 @@
...
+	return nil
+}
+
+// validateBlipBody validates incoming blip rev bodies
+func validateBlipBody(rawBody []byte, doc *Document) error {
+	// Prevent disallowed internal properties from being used
+	disallowed := []string{base.SyncPropertyName, BodyId, BodyRev, BodyDeleted, BodyRevisions}
+	for _, prop := range disallowed {
+		// Only unmarshal if raw body contains the disallowed property
+		if bytes.Contains(rawBody, []byte(prop)) {
```
bbrks: Would be nice to see if this can be a tiny bit more targeted to a JSON property to avoid false-positives. I think the best we can do here is wrapping the property in double quotes though, as the following colon could include whitespace.

For example, a string that happens to contain `_deleted_` (e.g. markdown) will match the Contains case but then fail the subsequent body check.

```suggestion
		if bytes.Contains(rawBody, []byte(`"`+prop+`"`)) {
```

db/validation.go
```diff
@@ -0,0 +1,85 @@
...
+		return base.HTTPErrorf(http.StatusNotFound, "Document revision is not accessible")
+	}
+
+	// Reject bodies that contains the "_purged" property.
+	if body[BodyPurged] != nil {
+		return base.HTTPErrorf(http.StatusBadRequest, "user defined top-level property '_purged' is not allowed in document body")
+	}
+
+	for key := range body {
+		if strings.HasPrefix(key, BodyInternalPrefix) {
```
bbrks, line 25: I think this is relevant even for existing code doing this byte/string checking upfront - but these checks are case-sensitive, and the Go JSON implementation is case _insensitive._

That difference might be the cause of subtle bugs, if a user is able to bypass the validation by writing properties in uppercase.

rest/blip_api_test.go
```diff
@@ -3975,3 +3975,131 @@ func TestAttachmentWithErroneousRevPos(t *testing.T) {
...
+			expectReject: false,
+		},
+		{
+			name:         "Invalid _sync",
+			inputBody:    map[string]interface{}{"_sync": true},
+			expectReject: true,
+		},
+		{
+			name:         "Valid _id",
+			inputBody:    map[string]interface{}{"_id": "documentid"},
```
bbrks, line 4004: As mentioned previously, our validation checks are mostly case sensitive. Would be interesting to include some uppercase tests for coverage.

bbrks (changes requested):
I've thought of a potential edge case in the existing and new code that is doing byte/string Contains checks prior to unmarshalling. See inline comments. I think we'll need to figure out a way of addressing that.

db/validation.go
IsaacLambat: Will do as part of [CBG-2051](https://issues.couchbase.com/browse/CBG-2051)

rest/blip_api_test.go
IsaacLambat: Will add as part of [CBG-2051](https://issues.couchbase.com/browse/CBG-2051)

bbrks (approved):

PR #5506, opened 2022-04-14
CBG-2016: Relax principal name restrictions

markspolakovs:
CBG-2016

Especially with OIDC, customers may want to e.g. namespace their usernames. Up until now we only permitted a select few characters (alphanumeric and any of "_-+.@") in principal names. Extend this to any character except for a few problematic cases:

- Invalid UTF-8
- Strings consisting only of punctuation (i.e. no alpha-numerics)
- Slashes, colons, backticks, and commas

- [x] `xattrs=true,gsi=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/203/
- [x] `xattrs=true,gsi=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/204/

### DIFF
--- a/auth/role.go
+++ b/auth/role.go
@@ -12,10 +12,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"regexp"
 	"strconv"
 	"strings"
 	"time"
+	"unicode"
 	"unicode/utf8"
 
 	"github.com/couchbase/sync_gateway/base"
@@ -94,15 +94,6 @@ func (pair *GrantHistorySequencePair) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
-// Reasons for forbidding each of these:
-// * colons: basic authentication uses them to separate usernames from passwords
-// * commas: fails channels.IsValidChannel, which channels.compileAccessMap uses via SetFromArray
-// * slashes: would need to make many (possibly breaking) changes to routing
-var (
-	kValidNameRegexp        = regexp.MustCompile(`^[^:,/]*$`)
-	atLeastOneAlphanumRegex = regexp.MustCompile(`[a-zA-Z0-9]`)
-)
-
 func (role *roleImpl) initRole(name string, channels base.Set) error {
 	channels = ch.ExpandingStar(channels)
 	role.Name_ = name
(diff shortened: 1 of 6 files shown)


auth/role.go
```diff
@@ -102,9 +110,16 @@ func (role *roleImpl) initRole(name string, channels base.Set) error {
...
+// at least one alphanumeric (except for the guest user), and no colons, commas, or slashes.
 func IsValidPrincipalName(name string) bool {
-	return kValidNameRegexp.Copy().MatchString(name)
+	if !utf8.ValidString(name) {
+		return false
+	}
+	if len(name) > 0 && !atLeastOneAlphanumRegex.MatchString(name) {
+		return false
+	}
+	return kValidNameRegexp.MatchString(name)
```
bbrks: I think It'll be a lot cheaper to do this (along with a utf8 rune check) in a single-pass of `name` given the type of checks were making.

I think it's worth doing given how often this is called - but feel free to benchmark to see if it's a valid assumption.

db/database_test.go
```diff
@@ -1364,20 +1364,12 @@ func TestAccessFunctionValidation(t *testing.T) {
 	_, _, err = db.Put("doc2", body)
 	assert.NoError(t, err, "")
 
-	body = Body{"users": []string{"bad username"}, "userChannels": []string{"BBC1"}}
-	_, _, err = db.Put("doc3", body)
-	assertHTTPError(t, err, 500)
-
-	body = Body{"users": []string{"role:bad rolename"}, "userChannels": []string{"BBC1"}}
-	_, _, err = db.Put("doc4", body)
-	assertHTTPError(t, err, 500)
```
bbrks: It's probably worth keeping these bad user/role checks but we'll have the modify them to fit the new definition of "bad", otherwise we're really only testing the happy-path case for a channel mapper's validation.

bbrks (changes requested):

auth/role.go
markspolakovs: Indeed, good call:

```
$ benchstat orig.txt new-validString.txt
name                     old time/op  new time/op  delta
IsValidPrincipalName-12  1.07µs ± 0%  0.07µs ± 1%  -93.57%  (p=0.016 n=4+5)
```

bbrks (approved):

PR #5508, opened 2022-04-14
Implement Connected Client BLIP API

snej:
CBG-2038

The Connected Client API runs over the same BLIP connection as the replicator, but uses a few new request types and adds a new parameter to subChanges. [It's documented here](https://github.com/couchbase/couchbase-lite-core/blob/feature/connected-client/docs/overview/ConnectedClientProtocol.md).

This PR:
- Adds a `future` parameter to `subChanges`. If set to `true`, it effectively sets `since` to the latest sequence ID, so only future changes will be sent.
- Adds `getRev` which requests the current revision of a single document.
- Adds `putRev` which stores a single document. (This is basically identical to `rev`, but made a separate request for reasons explained in the spec.)

Not implemented yet:
- `unsubChanges` — I think I will need help with this, because the whole machinery of the BLIP changes feed is complex enough that couldn't figure out how to stop one.

Potential issues:
- I implemented `handlePutRev` by simply calling `handleRev`. This works, but `handleRev` does some logging and stats-incrementing that might not be appropriate outside a replication; I'll leave that for reviewers to decide.

### DIFF
--- a/db/blip_connected_client.go
+++ b/db/blip_connected_client.go
@@ -1,0 +1,66 @@
+/*
+Copyright 2022-Present Couchbase, Inc.
+
+Use of this software is governed by the Business Source License included in
+the file licenses/BSL-Couchbase.txt.  As of the Change Date specified in that
+file, in accordance with the Business Source License, use of this software will
+be governed by the Apache License, Version 2.0, included in the file
+licenses/APL2.txt.
+*/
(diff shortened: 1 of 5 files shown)


bbrks (approved):
LGTM - will get issues filed for the TODOs/FIXME

PR #5509, opened 2022-04-19
CBG-2014: Check attachment size before sending to DB

markspolakovs:
CBG-2014

Previously we would rely on CBS to tell SGW when an attachment is too big (>20MB). However, if an attachment exceeds 30MB, CBS will abruptly close the connection instead, which would manifest as a 500 error.

Instead, check the attachment size before we try sending the attachment to the server.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/194/

### DIFF
--- a/db/attachment.go
+++ b/db/attachment.go
@@ -14,9 +14,11 @@ import (
 	"crypto/sha1"
 	"crypto/sha256"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"net/http"
 
+	"github.com/couchbase/gomemcached"
 	"github.com/couchbase/sync_gateway/base"
 )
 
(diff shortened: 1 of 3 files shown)


db/attachment.go
```diff
@@ -246,6 +251,9 @@ func (db *Database) setAttachment(key string, value []byte) error {
 func (db *Database) setAttachments(attachments AttachmentData) error {
 	for key, data := range attachments {
 		attachmentSize := int64(len(data))
+		if attachmentSize > int64(gomemcached.MaxBodyLen) {
```
bbrks: Can we define our own constant for doc size? `gomemcached` is virtually unused (directly) by SG and I'm not sure it makes sense for that library to determine what we do or don't send via GoCB.

db/attachment_test.go
```diff
@@ -1548,3 +1549,34 @@ func TestGetAttVersion(t *testing.T) {
...
+	context, err := NewDatabaseContext("db", base.GetTestBucket(t), false, DatabaseContextOptions{})
+	assert.NoError(t, err, "Couldn't create context for database 'db'")
+	defer context.Close()
+	db, err := CreateDatabase(context)
+	assert.NoError(t, err, "Couldn't create database 'db'")
+
+	normalAttachment := make([]byte, 10*1024*1024)
+	hugeAttachment := make([]byte, 30*1024*1024)
+	_, _ = rand.Read(normalAttachment)
+	_, _ = rand.Read(hugeAttachment)
```
bbrks: Could we have sizes that don't border a size limit (to avoid unintentionally hitting any off-by-one / `<` vs `<=` cases along various parts of the stack?)

e.g:
- 15MB // ok
- 25MB // covers old CB doc too large case
- 35MB // covers the EOF case

bbrks (changes requested):
Just a couple of small comments but otherwise looks ok

bbrks (approved):
LGTM 👍

PR #5511, opened 2022-04-20
CBG-1736: Namespace ISGR checkpoints by Cluster UUID

bbrks:
CBG-1736

- Generates a Cluster UUID which is stored in the SGRCluster doc.
- For newly created replications, prepends the cluster UUID into the checkpoint prefix.
  - Avoids edge replications with identical names from having conflicting remote checkpoints.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/200/

### DIFF
--- a/db/sg_replicate_cfg.go
+++ b/db/sg_replicate_cfg.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/couchbase/cbgt"
 	"github.com/couchbase/sync_gateway/base"
+	"github.com/google/uuid"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


bbrks:
@adamcfraser Can I get a review on this? I'll need to open a 3.0.1 backport for it, and want to get feedback soon

db/sg_replicate_cfg.go
```diff
@@ -972,6 +987,9 @@ func (m *sgReplicateManager) AddReplication(replication *ReplicationCfg) error {
 		if exists {
 			return false, base.ErrAlreadyExists
 		}
+		if replication.ClusterUUID == "" {
```
adamcfraser, line 990: Is it possible to add a test to verify correct behaviour on CAS retry on concurrent update of the cluster (i.e. we don't end up with two replications with different cluster UUIDs)

adamcfraser (changes requested):

db/sg_replicate_cfg.go
adamcfraser: As discussed, this isn't a blocker to merge this, as replications existing with different cluster UUIDs doesn't break the underlying functionality.

adamcfraser (approved):

PR #5513, opened 2022-04-21
CBG-2036: Validate OIDC issuer is reachable when validating config

markspolakovs:
CBG-2036

To ensure that the OIDC provider is reachable, we want to validate it when we're updating the database config. Note that we only do this when creating/updating a database, _not_ at startup (even with legacy config) - this is to ensure that SGW doesn't fail to start.

Users can opt out of this by passing `?disable_oidc_validation=true` in the PUT request.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/205/


rest/config.go
```diff
@@ -696,6 +696,15 @@ func (dbConfig *DbConfig) validateVersion(isEnterpriseEdition bool) error {
 		}
 	}
 
+	if validateOIDCConfig && dbConfig.OIDCConfig != nil {
+		for name, provider := range dbConfig.OIDCConfig.Providers {
+			_, _, err := provider.DiscoverConfig(context.TODO())
+			if err != nil {
+				multiError = multiError.Append(fmt.Errorf("failed to validate OIDC configuration for %s: %w", name, err))
```
markspolakovs, line 703: Does this need &lt;ud&gt;? I'd expect not, since this is returned directly to the user rather than being logged?

rest/config.go
```diff
@@ -696,6 +696,15 @@ func (dbConfig *DbConfig) validateVersion(isEnterpriseEdition bool) error {
 		}
 	}
 
+	if validateOIDCConfig && dbConfig.OIDCConfig != nil {
+		for name, provider := range dbConfig.OIDCConfig.Providers {
+			_, _, err := provider.DiscoverConfig(context.TODO())
```
markspolakovs: Is there a better context to use here? `validateVersion` doesn't take in a context, and other functions in `config.go` also use `context.TODO()`.

bbrks: Grabbing the context off the handler from the HTTP request (`h.rq.Context()`) and passing it down into validate would probably be worth doing, given the amount of work being done and logged inside `DiscoverConfig`.

rest/config.go
bbrks: It will get logged. We have a `RedactableError` type but I'm not sure it plays nicely with `MultiError` OR `%w` :/

In similar cases we've just tagged and redacted the value inside the error, even if it's going to be displayed to the user of the API (e.g. in `CouchbaseBucketGoCB) Touch`)

```
2022-04-22T15:18:04.387+01:00 [INF] HTTP: c:#001 POST /db2/_config (as <ud>Administrator</ud> as ADMIN)
2022-04-22T15:18:05.435+01:00 [TRC] Config+: merging upserted config into bucket config -- rest.(*handler).handlePutDbConfig.func1() at admin_api.go:529
2022-04-22T15:18:08.576+01:00 [WRN] Couldn't update config for database - rolling back: 400 1 errors:
failed to validate OIDC configuration for fooprovider: Get "https://foo.bbrks.me/.well-known/openid-configuration": x509: “TRAEFIK DEFAULT CERT” certificate is not trusted -- rest.(*handler).handlePutDbConfig() at admin_api.go:571
2022-04-22T15:18:10.712+01:00 [TRC] Config+: Got config for bucket "b2" with cas 1650637004435881984 -- rest.(*ServerContext).fetchDatabase() at config.go:1154
2022-04-22T15:18:10.712+01:00 [DBG] Config+: Database "db2" bucket "b2" config has not changed since last update
2022-04-22T15:18:10.712+01:00 [INF] HTTP: c:#001 #001:     --> 400 1 errors:
failed to validate OIDC configuration for fooprovider: Get "https://foo.bbrks.me/.well-known/openid-configuration": x509: “TRAEFIK DEFAULT CERT” certificate is not trusted  (6327.9 ms)
```

bbrks (changes requested):

rest/config.go
markspolakovs: Discussed this on Slack, and even if we redact the provider name (`%s`), we can't (easily) redact the wrapped error (`%w`) which can contain the issuer as in the above example. However, this is config provided by the SG administrator (so would be in a sgcollect), so not strictly user data, so not redacting is likely fine here.

rest/admin_api.go
```diff
@@ -40,12 +42,14 @@ func (h *handler) handleCreateDB() error {
...
 
+	validateOIDC := !h.getBoolQuery(paramDisableOIDCValidation)
+
 	if h.server.persistentConfig {
 		if err := config.validatePersistentDbConfig(); err != nil {
 			return base.HTTPErrorf(http.StatusBadRequest, err.Error())
 		}
 
-		if err := config.validate(); err != nil {
+		if err := config.validate(h.rqCtx, validateOIDC); err != nil {
```
bbrks: Think this will be safer if the context has not been previously initialised

```suggestion
		if err := config.validate(h.ctx(), validateOIDC); err != nil {
```

rest/admin_api.go
```diff
@@ -488,9 +492,11 @@ func (h *handler) handlePutDbConfig() (err error) {
 		dbConfig.Name = dbName
 	}
 
+	validateOIDC := !h.getBoolQuery(paramDisableOIDCValidation)
+
 	if !h.server.persistentConfig {
 		updatedDbConfig := &DatabaseConfig{DbConfig: *dbConfig}
-		err = updatedDbConfig.validate()
+		err = updatedDbConfig.validate(h.rqCtx, validateOIDC)
```
```suggestion
		err = updatedDbConfig.validate(h.ctx(), validateOIDC)
```

rest/admin_api.go
```diff
@@ -532,7 +538,7 @@ func (h *handler) handlePutDbConfig() (err error) {
 			if err := dbConfig.validatePersistentDbConfig(); err != nil {
 				return nil, base.HTTPErrorf(http.StatusBadRequest, err.Error())
 			}
-			if err := bucketDbConfig.validate(); err != nil {
+			if err := bucketDbConfig.validate(h.rqCtx, validateOIDC); err != nil {
```
```suggestion
			if err := bucketDbConfig.validate(h.ctx(), validateOIDC); err != nil {
```

rest/admin_api.go
```diff
@@ -690,7 +696,7 @@ func (h *handler) handlePutDbConfigSync() error {
 
 			bucketDbConfig.Sync = &js
 
-			if err := bucketDbConfig.validate(); err != nil {
+			if err := bucketDbConfig.validate(h.rqCtx, !h.getBoolQuery(paramDisableOIDCValidation)); err != nil {
```
```suggestion
			if err := bucketDbConfig.validate(h.ctx(), !h.getBoolQuery(paramDisableOIDCValidation)); err != nil {
```

rest/admin_api.go
```diff
@@ -838,7 +844,7 @@ func (h *handler) handlePutDbConfigImportFilter() error {
 
 			bucketDbConfig.ImportFilter = &js
 
-			if err := bucketDbConfig.validate(); err != nil {
+			if err := bucketDbConfig.validate(h.rqCtx, !h.getBoolQuery(paramDisableOIDCValidation)); err != nil {
```
```suggestion
			if err := bucketDbConfig.validate(h.ctx(), !h.getBoolQuery(paramDisableOIDCValidation)); err != nil {
```

bbrks (changes requested):
Looks good, just a minor change to how we're fetching the handler context

bbrks (approved):

PR #5514, opened 2022-04-22
CBG-1980: Fixed incorrect attachment stats being reported

IsaacLambat:
CBG-1980

- All attachment compaction phases wait for the DCP client doneChannel to close before returning out the function causing a correct count to be returned that will not increment after the function has exited
- The DCP doneChannel is only closed after all the DCP workers have terminated
- Unit test to repro incorrect stats

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/207/

### DIFF
--- a/base/dcp_client_worker.go
+++ b/base/dcp_client_worker.go
@@ -85,8 +85,8 @@ func (w *DCPWorker) Send(event streamEvent) {
 }
 
 func (w *DCPWorker) Start(wg *sync.WaitGroup) {
+	wg.Add(1)
 	go func() {
-		wg.Add(1)
 		defer wg.Done()
 		for {
 			select {
--- a/base/dcp_client_worker.go
+++ b/base/dcp_client_worker.go
@@ -3,6 +3,7 @@ package base
 import (
 	"bytes"
 	"context"
+	"sync"
 	"time"
 
 	sgbucket "github.com/couchbase/sg-bucket"
@@ -83,9 +84,10 @@ func (w *DCPWorker) Send(event streamEvent) {
 	}
 }
 
-func (w *DCPWorker) Start() {
-
+func (w *DCPWorker) Start(wg *sync.WaitGroup) {
 	go func() {
+		wg.Add(1)
+		defer wg.Done()
 		for {
 			select {
 			case event := <-w.eventFeed:
--- a/db/attachment_compaction_test.go
+++ b/db/attachment_compaction_test.go
@@ -813,7 +813,7 @@ func TestAttachmentCompactIncorrectStat(t *testing.T) {
 
 	testDb := setupTestDB(t)
 	defer testDb.Close()
-	// Create the docs that will be marked and swept
+	// Create the docs that will be marked and not swept
 	body := map[string]interface{}{"foo": "bar"}
 	for i := 0; i < docsToCreate; i++ {
 		key := fmt.Sprintf("%s_%d", t.Name(), i)
@@ -821,7 +821,7 @@ func TestAttachmentCompactIncorrectStat(t *testing.T) {
 		assert.NoError(t, err)
 	}
 
-	attKeys := make([]string, 0, 11)
+	attKeys := make([]string, 0, docsToCreate)
 	for i := 0; i < docsToCreate; i++ {
 		docID := fmt.Sprintf("testDoc-%d", i)
 		attKey := fmt.Sprintf("att-%d", i)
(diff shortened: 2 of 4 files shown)


base/dcp_client_worker.go
```diff
@@ -83,9 +84,10 @@ func (w *DCPWorker) Send(event streamEvent) {
 	}
 }
 
-func (w *DCPWorker) Start() {
-
+func (w *DCPWorker) Start(wg *sync.WaitGroup) {
 	go func() {
+		wg.Add(1)
```
bbrks: It's not safe to call Add inside the goroutine decrementing it because `wg.Wait()` might hit the case where the WaitGroup is 0 before these goroutines have started.

From the Go docs:
> Note that calls with a positive delta that occur when the counter is zero must happen before a Wait. Calls with a negative delta, or calls with a positive delta that start when the counter is greater than zero, may happen at any time. Typically this means the calls to Add should execute before the statement creating the goroutine or other event to be waited for.

db/attachment_compaction_test.go
```diff
@@ -802,3 +802,84 @@ func createDocWithInBodyAttachment(t *testing.T, docID string, docBody []byte, a
...
+	for i := 0; i < docsToCreate; i++ {
+		docID := fmt.Sprintf("testDoc-%d", i)
+		attKey := fmt.Sprintf("att-%d", i)
+		attBody := map[string]interface{}{"value": strconv.Itoa(i)}
+		attJSONBody, err := base.JSONMarshal(attBody)
+		assert.NoError(t, err)
+		attKeys = append(attKeys, CreateLegacyAttachmentDoc(t, testDb, docID, []byte("{}"), attKey, attJSONBody))
+	}
+
+	defer base.SetUpTestLogging(base.LevelInfo, base.KeyAll)()
```
bbrks: Uptake SetUpTestLogging changes from `master`

bbrks (dismissed):
Haven't done a proper review (I'll leave that to @JRascagneres who is more familiar with this/the task manager code) but wanted to note the wg usage:

db/attachment_compaction_test.go
```diff
@@ -802,3 +802,84 @@ func createDocWithInBodyAttachment(t *testing.T, docID string, docBody []byte, a
...
+	defer testDb.Close()
+	// Create the docs that will be marked and swept
+	body := map[string]interface{}{"foo": "bar"}
+	for i := 0; i < docsToCreate; i++ {
+		key := fmt.Sprintf("%s_%d", t.Name(), i)
+		_, _, err := testDb.Put(key, body)
+		assert.NoError(t, err)
+	}
+
+	attKeys := make([]string, 0, 11)
```
JRascagneres:
```suggestion
	attKeys := make([]string, 0, docsToCreate)
```

db/attachment_compaction_test.go
```diff
@@ -802,3 +802,84 @@ func createDocWithInBodyAttachment(t *testing.T, docID string, docBody []byte, a
...
+		}
+		return false, nil, nil
+	}
+
+	err, _ = base.RetryLoop("wait for marking to start", retryFunc, base.CreateSleeperFunc(1000, 1))
+	require.NoError(t, err)
+
+	terminator.Close() // Terminate mark function
+	// Allow time for timing issue to be hit where stat increments when it shouldn't
+	time.Sleep(time.Second * 1)
```
JRascagneres, line 866: Keep the sleep here for the test (unfortunately can't think of a better way)
but we should wait for the mark phase to return here.

db/attachment_compaction_test.go
```diff
@@ -802,3 +802,84 @@ func createDocWithInBodyAttachment(t *testing.T, docID string, docBody []byte, a
...
+
+	// Start sweeping with different compact ID so all documents get sweeped
+	stat = &base.AtomicInt{}
+	terminator = base.NewSafeTerminator()
+	go func() {
+		count, err = attachmentCompactSweepPhase(testDb, "sweep", nil, false, terminator, stat)
+		require.NoError(t, err)
+	}()
+
+	err, _ = base.RetryLoop("wait for marking to start", retryFunc, base.CreateSleeperFunc(1000, 1))
```
JRascagneres: change to sweep

JRascagneres (changes requested):
Couple test changes but mostly fine

JRascagneres (approved):

PR #5518, opened 2022-04-25
CBG-860: Replace defer SetUpTestLogging() with tb.Cleanup()

markspolakovs:
CBG-860

Replaces test functions of the kind `defer Foo()()` with `Foo(t)`, and makes them take advantage of `t.Cleanup()` to ensure they get run after the test. This ensures that it's impossible to forget the `defer` or the second `()`.

Because this changes almost every test, I've split the PR into two commits for ease of reviewing, one implementing the new signature and another updating them all to use it. This does mean that the first commit won't compile on its own.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/213/

### DIFF
--- a/base/util_test.go
+++ b/base/util_test.go
@@ -637,7 +637,7 @@ func (m *mockTB) Cleanup(fn func()) {
 	m.cleanupFn = fn
 }
 
-func TestSetUpTestLogging(t *testing.T) {
+func TestSetTestLogging(t *testing.T) {
 	if GlobalTestLoggingSet.IsTrue() {
 		t.Skip("Test does not work when a global test log level is set")
 	}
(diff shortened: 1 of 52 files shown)


markspolakovs:
Double-checked that the global log level is applied correctly:

```
❯ SG_TEST_LOG_LEVEL=debug go test -v ./auth | head
2022-04-25T12:32:51.935+01:00 [INF] auth.TestMain: Setup logging: level: debug - keys: *
=== RUN   TestValidateGuestUser
2022-04-25T12:32:51.935+01:00 [INF] Access: Recomputed channels for "<ud></ud>": <ud>!:1</ud>
2022-04-25T12:32:51.935+01:00 [INF] Access: Computed roles for "<ud></ud>": <ud></ud>
--- PASS: TestValidateGuestUser (0.00s)
=== RUN   TestValidateUser
2022-04-25T12:32:51.935+01:00 [INF] Access: Recomputed channels for "<ud>ValidName</ud>": <ud>!:1</ud>
2022-04-25T12:32:51.935+01:00 [INF] Access: Computed roles for "<ud>ValidName</ud>": <ud></ud>
2022-04-25T12:32:51.935+01:00 [INF] Access: Recomputed channels for "<ud>ValidName</ud>": <ud>!:1</ud>
2022-04-25T12:32:51.935+01:00 [INF] Access: Computed roles for "<ud>ValidName</ud>": <ud></ud>
`

File: base/util_testing.go [hunk 1]
```diff
@@ -464,49 +466,45 @@ func SetUpGlobalTestLogging(m *testing.M) (teardownFn func()) {
...
 			consoleLogger.logger.SetOutput(consoleLogger.output)
 		} else {
 			consoleLogger.logger.SetOutput(os.Stderr)
 		}
 		teardownFnOrig()
-	}
+	})
 }
 
 func setTestLogging(logLevel LogLevel, caller string, logKeys ...LogKey) (teardownFn func()) {
```
bbrks, line 505: Would it be possible to replace the `panic` with a `tb.Fatal` now?

base/util_testing.go
```diff
@@ -464,49 +466,45 @@ func SetUpGlobalTestLogging(m *testing.M) (teardownFn func()) {
...
-// To set multiple log keys, append as variadic arguments
-// E.g. KeyCache,KeyDCP,KeySync
-//
-// Usage:
-//     teardownFn := SetUpTestLogging(LevelDebug, KeyCache,KeyDCP,KeySync)
-//     defer teardownFn()
+// canCleanup is a subset of testing.TB used for testing the test logging.
+type canCleanup interface {
+	Cleanup(func())
+}
```
bbrks: With the mock comment above addressed, I think the functions can just take a `testing.TB` interface instead.

base/util_test.go
```diff
@@ -638,31 +646,35 @@ func TestSetUpTestLogging(t *testing.T) {
 	require.Equal(t, LevelInfo, *consoleLogger.LogLevel)
 	require.Equal(t, *logKeyMask(KeyHTTP), *consoleLogger.LogKeyMask)
 
-	teardownFn := SetUpTestLogging(LevelDebug, KeyDCP, KeySync)
+	mock := &mockTB{}
```
bbrks: I can see why this was required for testing `SetUpTestLogging`, but I think tweaking the test to be running `setTestLogging` and executing the returned function would be sufficient.

bbrks (changes requested):
Mostly looks good. Have a few specific questions that came up from `TestSetUpTestLogging`

base/util_testing.go
markspolakovs: `setTestLogging` can't be made to take a `testing.TB` (otherwise it wouldn't need to return a `teardownFn` at all) because it can also be called from `SetUpGlobalTestLogging`, which runs before any tests have started and so doesn't have a `testing.T`.

base/util_test.go
```diff
@@ -629,7 +629,15 @@ func TestRedactBasicAuthURL(t *testing.T) {
...
-func TestSetUpTestLogging(t *testing.T) {
+type mockTB struct {
+	cleanupFn func()
+}
+
+func (m *mockTB) Cleanup(fn func()) {
+	m.cleanupFn = fn
+}
+
+func TestSetTestLogging(t *testing.T) {
```
bbrks: now unused

```suggestion
func TestSetTestLogging(t *testing.T) {
```

bbrks (approved):

PR #5521, opened 2022-04-26
CBG-2004: Prove attachment unit tested, and slight cleanup of prove attachment code

IsaacLambat:
CBG-2004

- Added unit test for using testing the prove attachment case when a `ErrAttachmentNotFound` is returned over Blip
- Slight clean-up of prove attachment code

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/218 - known flaky test fail which passes locally

### DIFF
--- a/rest/blip_api_test.go
+++ b/rest/blip_api_test.go
@@ -4102,3 +4102,57 @@ func TestBlipInternalPropertiesHandling(t *testing.T) {
 		})
 	}
 }
+
+// CBG-2004: Test that prove attachment over Blip works correctly when receiving a ErrAttachmentNotFound
+func TestProveAttachmentNotFound(t *testing.T) {
+	rt := NewRestTester(t, &RestTesterConfig{
+		guestEnabled: true,
+	})
+	defer rt.Close()
+
+	bt, err := NewBlipTesterFromSpecWithRT(t, nil, rt)
+	assert.NoError(t, err, "Error creating BlipTester")
+	defer bt.Close()
+
+	attachmentData := []byte("attachmentA")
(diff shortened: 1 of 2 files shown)


rest/blip_api_test.go
```diff
@@ -3975,3 +3975,57 @@ func TestAttachmentWithErroneousRevPos(t *testing.T) {
...
+
+	err = rt.WaitForPendingChanges()
+	require.NoError(t, err)
+
+	// Should log:
+	// "Peer sent prove attachment error 404 attachment not found, falling back to getAttachment for proof in doc <ud>doc1</ud> (digest sha1-wzp8ZyykdEuZ9GuqmxQ7XDrY7Co=)"
+	base.SetUpTestLogging(t, base.LevelInfo, base.KeyAll)
+
+	// Use different attachment name to bypass digest check in ForEachStubAttachment() which skips prove attachment code
+	// Set attachment to V2 so it can be retrieved by RT successfully
```
IsaacLambat, line 4145: Attachment cannot be retrieved from the bucket if the version is not included like so.

The GET attachment handler for `/{db}/{doc}/{attach}` assumes that the attachment version is version 1 (even though it's stored in the bucket as a v2 attachment) so it can't find it due to the name being incorrect. This is due to the attachment metadata not having a version tag which `GetAttachmentVersion` parses.

Possible bug, or could be Blip tester behaviour only.

IsaacLambat: Still fails if not including attachment version when not hitting Prove Attachment code. If the version is included it also gets stripped out so assuming it is a blip tester problem or else this problem would have already come up a lot earlier in the past.

bbrks (approved):

PR #5523, opened 2022-05-02
CBG-2027 Enhancements to GET /db/_user/

adamcfraser:
Adds the ability to include email and disabled values when retrieving all users via GET /db/_user/ by setting name_only=false.  Also supports specifying a limit on that REST API when name_only=false is being used.

Views support is not required (and not included, as it would require an update to view definitions).  For 3.1.0 we plan to support this functionality more generally with new indexes on the system collection, but putting this in master for the time being.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/222/ - full run with views
- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/223/ - new test with GSI

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -1339,8 +1339,12 @@ func (h *handler) getUsers() error {
 		}
 		bytes, marshalErr = base.JSONMarshal(users)
 	}
+
+	if marshalErr != nil {
+		return marshalErr
+	}
 	h.writeRawJSON(bytes)
-	return marshalErr
+	return nil
 }
 
 func (h *handler) getRoles() error {
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -30,6 +30,10 @@ const kDefaultDBOnlineDelay = 0
 
 const paramDisableOIDCValidation = "disable_oidc_validation"
 
+// GetUsers  - GET /{db}/_user/
+const paramNameOnly = "name_only"
+const paramLimit = "limit"
+
 // ////// DATABASE MAINTENANCE:
 
 // "Create" a database (actually just register an existing bucket)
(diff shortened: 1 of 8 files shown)


rest/admin_api.go
```diff
@@ -1309,13 +1313,34 @@ func (h *handler) getRoleInfo() error {
...
+		users, err := h.db.GetUsers(h.db.Ctx, int(limit))
+		if err != nil {
+			return err
+		}
+		bytes, marshalErr = base.JSONMarshal(users)
 	}
-	bytes, err := base.JSONMarshal(users)
 	h.writeRawJSON(bytes)
-	return err
+	return marshalErr
```
bbrks: This (including the existing error handling code) isn't safe as `writeRawJSON` writes a 200 status code, and the subsequent error cannot override the status code because it's already been sent to the client.

```
$ curl -v http://Administrator:password@localhost:4985/db2/_user/
*   Trying 127.0.0.1:4985...
* Connected to localhost (127.0.0.1) port 4985 (#0)
* Server auth using Basic with user 'Administrator'
> GET /db2/_user/ HTTP/1.1
> Host: localhost:4985
> Authorization: Basic QWRtaW5pc3RyYXRvcjpwYXNzd29yZA==
> User-Agent: curl/7.79.1
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Content-Length: 0
< Content-Type: application/json
< Server: Couchbase Sync Gateway/3.1 branch/ commit/ CE
< Date: Tue, 03 May 2022 13:47:46 GMT
<
* Connection #0 to host localhost left intact
```

```
2022-05-03T14:47:46.845+01:00 [INF] HTTP: c:#001 GET /db2/_user/ (as <ud>Administrator</ud> as ADMIN)
2022/05/03 14:47:46 http: superfluous response.WriteHeader call from github.com/couchbase/sync_gateway/rest.(*handler).writeStatus (handler.go:1077)
2022-05-03T14:47:46.891+01:00 [ERR] c:#001 #001: foo -- rest.(*handler).writeError() at handler.go:1046
2022-05-03T14:47:46.891+01:00 [INF] HTTP: c:#001 #001:     --> 500 Internal error: foo  (3470.3 ms)
```

rest/admin_api.go
```diff
@@ -30,6 +30,10 @@ const kDefaultDBOnlineDelay = 0
 
 const paramDisableOIDCValidation = "disable_oidc_validation"
 
+// GetUsers  - GET /{db}/_user/
+const paramNameOnly = "name_only"
+const paramLimit = "limit"
```
bbrks, line 35: Can these parameters be included in the API docs?

You can add define them in the `docs/api/components/parameters.yaml` file

```yaml
usersNameOnly:
  name: name_only
  in: query
  required: false
  schema:
    type: boolean
    default: true
  description: Whether to return user names only, or more detailed information for each user.
usersLimit:
  name: limit
  in: query
  required: false
  schema:
    type: integer
  description: How many results to return.
```

and then reference them from the `docs/api/paths/admin/{db}~_user~.yaml` file on the GET method:

```yaml
get:
  parameters:
    - $ref: ../../components/parameters.yaml#/usersNameOnly
    - $ref: ../../components/parameters.yaml#/usersLimit
```

bbrks (changes requested):

rest/admin_api.go
adamcfraser: Will do - thanks for typing that up.

bbrks (approved):

PR #5530, opened 2022-05-06
CBG-2053: Added stat tracking for connected client putRev messages

IsaacLambat:
CBG-2053

- Added additional stats for `handlePutRev`, and modified existing stats
- Moved `handleRev` code to shared function `processRev` which both `handleRev` and `handlePutRev` now use
- Added stats mapping for `processRev` so `handleRev` and `handlePutRev` both can have their own independant stats
- Added test to make sure stats for `handleRev` still work correctly

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/240/

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -790,7 +790,8 @@ type processRevStats struct {
 }
 
 // Processes a "rev" request, i.e. client is pushing a revision body
-func (bh *blipHandler) processRev(rq *blip.Message, stats processRevStats) (err error) {
+// stats must always be provided, along with all the fields filled with valid pointers
+func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err error) {
 	startTime := time.Now()
 	defer func() {
 		stats.processingTime.Add(time.Since(startTime).Nanoseconds())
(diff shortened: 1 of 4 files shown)


db/blip_handler.go
```diff
@@ -780,15 +780,24 @@ type removalDocument struct {
...
+	count           *base.SgwIntStat // Increments when rev processed successfully
+	errorCount      *base.SgwIntStat
+	deltaRecvCount  *base.SgwIntStat
+	bytes           *base.SgwIntStat
+	processingTime  *base.SgwIntStat
+	docsPurgedCount *base.SgwIntStat
+}
+
+// Processes a "rev" request, i.e. client is pushing a revision body
+func (bh *blipHandler) processRev(rq *blip.Message, stats processRevStats) (err error) {
```
bbrks: Nitpick: Would feel clearer to pass a pointer to the struct here instead of copying the struct.

Doesn't strictly matter, since the struct itself contains pointers, but would communicate the intended usage of `processRev` modifying items inside the struct a bit better.

bbrks (dismissed):
This looks good, I have one _tiny_ nitpick about the new stats parameter of `processRev` but if you disagree that's fine

bbrks (approved):

PR #5531, opened 2022-05-06
CBG-2058: Always pass a terminator to Compact

markspolakovs:
CBG-2058

If SGW was left running long enough that the compaction background task runs (so only on nodes with import enabled), and there are tombstones present older than the metadata purge interval, the task would check that a nil terminator is complete and thus panic. This PR ensures we always pass down a terminator to Compact, as well as adding a regression test.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/249/

### DIFF
--- a/db/database_test.go
+++ b/db/database_test.go
@@ -2505,6 +2505,40 @@ func TestGetAllUsers(t *testing.T) {
 	assert.Equal(t, 2, len(limitedUsers))
 }
 
+// Regression test for CBG-2058.
+func TestImportCompactPanic(t *testing.T) {
+	if !base.TestUseXattrs() {
+		t.Skip("requires xattrs")
+	}
+
+	// Set the compaction and purge interval unrealistically low to reproduce faster
+	options := DatabaseContextOptions{
+		CompactInterval: 1,
+	}
+	AddOptionsFromEnvironmentVariables(&options)
+	bucket := base.GetTestBucket(t)
+	dbc, err := NewDatabaseContext("db", bucket, true, options)
+	require.NoError(t, err, "Couldn't create context for database 'db'")
+	defer dbc.Close()
+
+	dbc.PurgeInterval = time.Millisecond
+	db, err := CreateDatabase(dbc)
+	require.NoError(t, err, "Couldn't create database 'db'")
+
+	// Create a document, then delete it, to create a tombstone
+	rev, doc, err := db.Put("test", Body{})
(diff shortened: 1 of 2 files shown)


db/database_test.go
```diff
@@ -66,6 +66,11 @@ func setupTestDBForBucketWithOptions(t testing.TB, tBucket base.Bucket, dbcOptio
 
 func setupTestDBWithOptionsAndImport(t testing.TB, dbcOptions DatabaseContextOptions) *Database {
 	AddOptionsFromEnvironmentVariables(&dbcOptions)
+	if dbcOptions.GroupID == "" && base.IsEnterpriseEdition() {
+		// In a real SG this would be set in DefaultStartupConfig.
+		dbcOptions.GroupID = "default"
+		RegisterImportPindexImpl("default")
```
bbrks: This seems odd that it's not already done as part of `NewDatabaseContext` in the real SG code.
Having it done inside the function that generates an options struct (`dbcOptionsFromConfig`) feels a bit like a side-effect.

db/database_test.go
```diff
@@ -66,6 +66,11 @@ func setupTestDBForBucketWithOptions(t testing.TB, tBucket base.Bucket, dbcOptio
 
 func setupTestDBWithOptionsAndImport(t testing.TB, dbcOptions DatabaseContextOptions) *Database {
 	AddOptionsFromEnvironmentVariables(&dbcOptions)
+	if dbcOptions.GroupID == "" && base.IsEnterpriseEdition() {
+		// In a real SG this would be set in DefaultStartupConfig.
+		dbcOptions.GroupID = "default"
```
bbrks: `"default"` is actually not the default beyond the SG config file, it's an empty string - but it not working with empty strings implies there might be a larger problem here...

bbrks (changes requested):
Have a few questions about the group ID stuff - maybe we can walk through the changes in a call to discuss instead of back-and-forth

db/database_test.go
markspolakovs: Discussed offline - this should indeed be moved into `NewDatabaseContext`, but that's slightly outside the scope of this PR (and it's risky for a PR that's getting backported to 3.0.1), so I'll leave this as is with a TODO (to get the tests to pass) and file a separate one to move this out.

bbrks (approved):
LGTM

PR #5534, opened 2022-05-10
CBG-2070: Refactor oidc_api_test factories

markspolakovs:
CBG-2070

The names of some of the test helper functions were getting pretty long (e.g. I was about to create `mockProviderWithRegisterWithUsernameClaimWithUserPrefixWithRolesClaim` which is far too long), so break them out into a `mockProviderWith` method that takes some varargs that each set an option on the mock provider - that way it can always be the same function but with different arguments.

As part of this, I noticed that some of them were setting `UserPrefix` when arguably they shouldn't have been. In most cases the fix is to explicitly add `mockProviderUserPrefix`, but there are a few tests (namely *"successful registered user auth when neither username_claim nor user_prefix are set"* and its register counterpart) that test the result of applying `UserPrefix`, which were incorrect: per the docs, if neither `UserPrefix` nor `UsernameClaim` are set, the username should be `[issuer][subject]`, where the issuer is a URL, but the tests were checking for the provider `Name` (not the issuer). I've updated these tests to correctly reference the issuer.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/252/

### DIFF
--- a/rest/oidc_api_test.go
+++ b/rest/oidc_api_test.go
@@ -455,137 +455,55 @@ func mockProvider(name string) *auth.OIDCProvider {
 	return &auth.OIDCProvider{
 		Name:          name,
 		ClientID:      "baz",
-		UserPrefix:    name,
 		ValidationKey: base.StringPtr("qux"),
 	}
 }
 
-// mockProviderWithRegister returns an auto registration enabled provider.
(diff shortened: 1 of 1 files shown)


bbrks (approved):
Looks good, thanks!

PR #5535, opened 2022-05-10
CBG-2071 ensure CRC32 hashes are always 8 characters

torcolvin:
CBG-2071 server and SGW hashes would some times mismatch since SGW would not prepend leading zeros to provide a hash value.

### DIFF
--- a/base/util.go
+++ b/base/util.go
@@ -1057,6 +1057,7 @@ func Crc32cHash(input []byte) uint32 {
 	return crc32.Checksum(input, table)
 }
 
+// Crc32cHashString returns a zero padded version of a crc32 hash to always be hexidecimal prefixed 8 character string
 func Crc32cHashString(input []byte) string {
 	return fmt.Sprintf("0x%08x", Crc32cHash(input))
 }
--- a/base/util.go
+++ b/base/util.go
@@ -1058,7 +1058,7 @@ func Crc32cHash(input []byte) uint32 {
 }
 
 func Crc32cHashString(input []byte) string {
-	return fmt.Sprintf("0x%x", Crc32cHash(input))
+	return fmt.Sprintf("0x%08x", Crc32cHash(input))
 }
 
 func SplitHostPort(hostport string) (string, string, error) {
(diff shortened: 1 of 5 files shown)


base/util.go
```diff
@@ -1058,7 +1058,7 @@ func Crc32cHash(input []byte) uint32 {
 }
 
 func Crc32cHashString(input []byte) string {
-	return fmt.Sprintf("0x%x", Crc32cHash(input))
+	return fmt.Sprintf("0x%08x", Crc32cHash(input))
```
bbrks, line 1062: This function is also used by `GenerateIndexName` in `base/dcp_sharded.go` - changing the padding in this use-case will have implications for pre-existing CBGT indexes.
Given a new index name, import processing will not be able to resume from the last checkpoint, would reprocess documents from sequence zero, and leave old non-padded index around.

My preferred approach would be to have `GenerateIndexName` do its own non-padding formatting on `Crc32cHash` to maintain the existing behaviour, and include a godoc comment on this `Crc32cHashString` function to clarify that it now returns a padded 8 char hex string.

rest/import_test.go
```diff
@@ -2363,3 +2363,60 @@ func TestImportInternalPropertiesHandling(t *testing.T) {
...
+	var rawInsertResponse RawResponse
+	err = base.JSONUnmarshal(response.Body.Bytes(), &rawInsertResponse)
+	assert.NoError(t, err, "Unable to unmarshal raw response")
+	initialRev := rawInsertResponse.Sync.Rev
+
+	// 2. Test import behaviour after SDK touch
+	updatedBody := make(map[string]interface{})
+	updatedBody["test"] = "TestImportDelete"
+	updatedBody["channels"] = "HBO"
+
```
bbrks: not used given we're only doing a touch

```suggestion
```

bbrks (dismissed):
Looks good, left one comment around other usages of `Crc32cHashString` that we might want to maintain existing non-padded behaviour for.

adamcfraser (approved):

PR #5536, opened 2022-05-11
CBG-2064: Allow mapping OIDC claims to roles/channels

markspolakovs:
CBG-2064

This PR adds two new configuration properties for OIDC providers, roles_claim and channels_claim, both empty by default. When set, SGW will look for a claim matching that and add its value to the user's role/channel sets.

* This adds three new properties to the persisted User - oidc_issuer, oidc_roles and oidc_channels.
* The new sets are unioned with the explicit roles/channels and any added through the sync function.
* We also store the issuer of the provider on the User doc - if, when they sign in, that provider has been removed from the DB configuration, we revoke the OIDC roles/channels.

An open question is whether the OIDC roles/channels need to be revoked if the user signs in through something other than OIDC. Right now this is not done (i.e. they will keep their OIDC roles/channels even if signing in through basic auth / admin _session) - if this needs changing this will likely be best done as a separate ticket as it requires changes to the non-OIDC login flow.

## Dependencies

- [x] #5546 

- [x]  `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/313/

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -788,7 +788,7 @@ func (auth *Authenticator) authenticateOIDCIdentity(identity *Identity, provider
 	// Check if we need to grant or revoke access to any OIDC roles/channels
 	if user != nil {
 		if !user.OIDCRoles().Equals(oidcRoles) || !user.OIDCChannels().Equals(oidcChannels) {
-			if err := auth.UpdateUserOIDCRolesChannels(user, oidcRoles, oidcChannels); err != nil {
+			if err := auth.UpdateUserOIDCRolesChannels(user, provider.Issuer, oidcRoles, oidcChannels); err != nil {
 				err = fmt.Errorf("failed to save updated OIDC roles/channels for %v: %w", base.UD(user.Name()).Redact(), err)
 				base.WarnfCtx(auth.LogCtx, "Unable to update OIDC user: %v", err)
 				return nil, time.Time{}, err
@@ -838,7 +838,7 @@ func (auth *Authenticator) RegisterNewUser(username, email string) (User, error)
 
 // UpdateUserOIDCRolesChannels saves the user's current OIDC roles/channels, and resulting implicit roles/channels,
 // to the database.
-func (auth *Authenticator) UpdateUserOIDCRolesChannels(user_ User, newOIDCRoles, newOIDCChannels base.Set) error {
+func (auth *Authenticator) UpdateUserOIDCRolesChannels(user_ User, issuer string, newOIDCRoles, newOIDCChannels base.Set) error {
 	updateRolesChannelsCallback := func(currentPrincipal Principal) (updatedPrincipal Principal, err error) {
 		user, ok := currentPrincipal.(User)
 		if !ok {
@@ -850,6 +850,11 @@ func (auth *Authenticator) UpdateUserOIDCRolesChannels(user_ User, newOIDCRoles,
 
 		// TODO: these calls to rebuildRoles/rebuildChannels are somewhat expensive (require access/role-access queries)
 		// and we could hit this more than once if we're inside a CAS-loop.
+
+		if user.OIDCIssuer() != issuer {
+			user.SetOIDCIssuer(issuer)
+		}
+		// newOIDCRoles/newOIDCChannels will have come from the new issuer, so no need to revoke and re-add
 
 		oidcChannels := user.OIDCChannels()
 		if oidcChannels == nil {
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -628,7 +628,7 @@ func (auth *Authenticator) AuthenticateUser(username string, password string) (U
 // Used to authenticate a JWT token coming from an insecure source (e.g. client request)
 // If the token is validated but the user for the username defined in the subject claim doesn't exist,
 // creates the user when autoRegister=true.
-func (auth *Authenticator) AuthenticateUntrustedJWT(token string, providers OIDCProviderMap, callbackURLFunc OIDCCallbackURLFunc) (User, error) {
+func (auth *Authenticator) AuthenticateUntrustedJWT(token string, providers OIDCProviderMap, callbackURLFunc OIDCCallbackURLFunc) (User, PrincipalConfig, error) {
 
 	base.DebugfCtx(auth.LogCtx, base.KeyAuth, "AuthenticateUntrustedJWT called with token: %s", base.UD(token))
 	var provider *OIDCProvider
@@ -641,7 +641,7 @@ func (auth *Authenticator) AuthenticateUntrustedJWT(token string, providers OIDC
 		jwt, err := jwt.ParseSigned(token)
 		if err != nil {
 			base.DebugfCtx(auth.LogCtx, base.KeyAuth, "Error parsing JWT in AuthenticateUntrustedJWT: %v", err)
-			return nil, err
+			return nil, PrincipalConfig{}, err
 		}
 
 		// Extract issuer and audience(s) from JSON Web Token.
@@ -650,7 +650,7 @@ func (auth *Authenticator) AuthenticateUntrustedJWT(token string, providers OIDC
 		base.DebugfCtx(auth.LogCtx, base.KeyAuth, "JWT issuer: %v, audiences: %v", base.UD(issuer), base.UD(audiences))
 		if err != nil {
 			base.DebugfCtx(auth.LogCtx, base.KeyAuth, "Error getting issuer and audience from token: %v", err)
-			return nil, err
+			return nil, PrincipalConfig{}, err
 		}
 
 		base.DebugfCtx(auth.LogCtx, base.KeyAuth, "Call GetProviderForIssuer w/ providers: %+v", base.UD(providers))
@@ -659,15 +659,15 @@ func (auth *Authenticator) AuthenticateUntrustedJWT(token string, providers OIDC
 	}
 
 	if provider == nil {
-		return nil, base.RedactErrorf("No provider found for issuer %v", base.UD(issuer))
+		return nil, PrincipalConfig{}, base.RedactErrorf("No provider found for issuer %v", base.UD(issuer))
 	}
 
 	identity, verifyErr := verifyToken(auth.LogCtx, token, provider, callbackURLFunc)
 	if verifyErr != nil {
-		return nil, verifyErr
+		return nil, PrincipalConfig{}, verifyErr
 	}
-	user, _, err := auth.authenticateOIDCIdentity(identity, provider)
-	return user, err
+	user, updates, _, err := auth.authenticateOIDCIdentity(identity, provider)
+	return user, updates, err
 }
 
 // verifyToken verifies claims and signature on the token; ensure that it's been signed by the provider.
@@ -702,7 +702,7 @@ func verifyToken(ctx context.Context, token string, provider *OIDCProvider, call
 // If the token is validated but the user for the username defined in the subject claim doesn't exist,
 // creates the user when autoRegister=true.
 func (auth *Authenticator) AuthenticateTrustedJWT(token string, provider *OIDCProvider, callbackURLFunc OIDCCallbackURLFunc) (user User,
-	tokenExpiry time.Time, err error) {
+	updates PrincipalConfig, tokenExpiry time.Time, err error) {
 	base.DebugfCtx(auth.LogCtx, base.KeyAuth, "AuthenticateTrustedJWT called with token: %s", base.UD(token))
 
 	var identity *Identity
@@ -711,14 +711,14 @@ func (auth *Authenticator) AuthenticateTrustedJWT(token string, provider *OIDCPr
 		identity, err = VerifyClaims(token, provider.ClientID, provider.Issuer)
 		if err != nil {
 			base.DebugfCtx(auth.LogCtx, base.KeyAuth, "Error verifying raw token in AuthenticateTrustedJWT: %v", err)
-			return nil, time.Time{}, err
+			return nil, PrincipalConfig{}, time.Time{}, err
 		}
 	} else {
 		// Verify claims and signature on the JWT.
 		var verifyErr error
 		identity, verifyErr = verifyToken(auth.LogCtx, token, provider, callbackURLFunc)
 		if verifyErr != nil {
-			return nil, time.Time{}, verifyErr
+			return nil, PrincipalConfig{}, time.Time{}, verifyErr
 		}
 	}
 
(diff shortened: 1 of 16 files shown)


auth/auth.go
```diff
@@ -781,3 +831,49 @@ func (auth *Authenticator) RegisterNewUser(username, email string) (User, error)
...
+			return nil, base.ErrUpdateCancel
+		}
+
+		nextSeq := user.Sequence() + 1
+		user.SetSequence(nextSeq)
+
+		// TODO: these calls to rebuildRoles/rebuildChannels are somewhat expensive (require access/role-access queries)
+		// and we could hit this more than once if we're inside a CAS-loop.
+
+		newChans := user.OIDCChannels()
```
adamcfraser: This is named newChans, but looks like it's storing the previous/existing user channels, while the 'roles' and 'channels' parameters the newly updated roles/channels.  If I've got that right, can we name these 'existingChans'?

auth/auth.go
```diff
@@ -781,3 +831,49 @@ func (auth *Authenticator) RegisterNewUser(username, email string) (User, error)
...
+		}
+		if !newRoles.Equals(roles) {
+			newRoles.UpdateAtSequence(roles, nextSeq)
+			user.SetOIDCRoles(newRoles, nextSeq)
+			if err := auth.rebuildChannels(user); err != nil {
+				return nil, fmt.Errorf("failed to rebuild channels for %v: %w", base.UD(user.Name()).Redact(), err)
+			}
+		}
+
+		base.DebugfCtx(auth.LogCtx, base.KeyAuth, "Updating user %s OIDC roles/channels to: %v, %v", base.UD(user.Name()), base.UD(user.OIDCRoles()), base.UD(user.OIDCChannels()).Redact())
```
adamcfraser: base.UD(user.OIDCChannels()).Redact() looks like a relatively expensive operation that's going to be performed here even if debug logging isn't enabled.  

Is there a reason to not just have base.UD(user.OIDCChannels()) here?

auth/oidc.go
```diff
@@ -159,6 +159,9 @@ type OIDCProvider struct {
 	// Sync Gateway and the underlying OIDC library.
 	UsernameClaim string `json:"username_claim"`
 
+	RolesClaim    string `json:"roles_claim"`
```
adamcfraser, line 165: Can remove the whitespace on line 161.

adamcfraser (changes requested):
Generally looks good - a few questions/comments.

auth/auth.go
markspolakovs: It does start out as the user's previous OIDC channels, but then we add/remove the incoming channel set (to preserve the sequences of the pre-existing ones that don't get revoked). You're right on the naming though, it is a bit confusing, let me see if I can think of a better name.

auth/oidc.go
markspolakovs: Huh, there's none there for me?
image

rest/handler.go
```diff
@@ -269,6 +269,25 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
 		}
 	}
 
+	// If the user has OIDC roles/channels configured, we need to check if the OIDC issuer they came from is still valid.
```
adamcfraser, line 277: As discussed, would be nice to keep invoke smaller/cleaner. Either move to a helper function, push down to checkAuth and deal with early returns via defer or refactoring, etc.

auth/auth.go
```diff
@@ -723,6 +737,24 @@ func (auth *Authenticator) authenticateOIDCIdentity(identity *Identity, provider
 	}
 	base.DebugfCtx(auth.LogCtx, base.KeyAuth, "OIDCUsername: %v", base.UD(username))
 
+	var oidcRoles, oidcChannels base.Set
+	if provider.RolesClaim != "" {
+		oidcRoles, err = getJWTClaimAsSet(identity, provider.RolesClaim)
+		if err != nil {
+			return nil, time.Time{}, fmt.Errorf("failed to find OIDC roles: %w", err)
```
adamcfraser: In the scenario where a configured claim can't be resolved as a set, I'm wondering whether authentication should fail altogether, or if the oidcRoles/oidcChannels should just be emptied.  Any background on why you went the 'fail altogether' route here?

auth/auth.go
```diff
@@ -749,6 +781,24 @@ func (auth *Authenticator) authenticateOIDCIdentity(identity *Identity, provider
 		}
 	}
 
+	// Check if we need to grant or revoke access to any OIDC roles/channels
+	if user != nil {
+		if !user.OIDCRoles().Equals(oidcRoles) || !user.OIDCChannels().Equals(oidcChannels) {
+			if err := auth.UpdateUserOIDCRolesChannels(user, provider.Issuer, oidcRoles, oidcChannels); err != nil {
```
adamcfraser: This looks like a single auth could update the user document multiple times (for email above, and then again here).  It would be preferable to avoid this.

auth/auth.go
```diff
@@ -749,6 +781,24 @@ func (auth *Authenticator) authenticateOIDCIdentity(identity *Identity, provider
...
+				err = fmt.Errorf("failed to save updated OIDC roles/channels for %v: %w", base.UD(user.Name()).Redact(), err)
+				base.WarnfCtx(auth.LogCtx, "Unable to update OIDC user: %v", err)
+				return nil, time.Time{}, err
+			}
+			// Fetch the up-to-date roles/channels
+			user, err = auth.GetUser(user.Name())
```
adamcfraser: Do we need to re-fetch the user here, or can we apply the changes to the in-memory version we already loaded?

auth/auth.go
```diff
@@ -781,3 +831,58 @@ func (auth *Authenticator) RegisterNewUser(username, email string) (User, error)
...
+		if !ok {
+			return nil, base.ErrUpdateCancel
+		}
+
+		if user.OIDCIssuer() == issuer && user.OIDCChannels().Equals(newOIDCChannels) && user.OIDCRoles().Equals(newOIDCRoles) {
+			return nil, base.ErrUpdateCancel
+		}
+
+		nextSeq := user.Sequence() + 1
+		user.SetSequence(nextSeq)
```
adamcfraser: User sequences are based on the global SG sequence, not incremented per user.  UpdatePrincipal includes the correct way to update the sequence, can use that as a reference.

auth/auth.go
```diff
@@ -781,3 +831,58 @@ func (auth *Authenticator) RegisterNewUser(username, email string) (User, error)
...
+
+		// TODO: these calls to rebuildRoles/rebuildChannels are somewhat expensive (require access/role-access queries)
```
adamcfraser: We would be performing rebuildRoles/rebuildChannels as part of JWT authentication normally (if roles/channels had been invalidated).  Are we doing it twice in the case that OIDCChannels/OIDCRoles have changed?

auth/auth.go
```diff
@@ -781,3 +831,58 @@ func (auth *Authenticator) RegisterNewUser(username, email string) (User, error)
...
 	return user, err
 }
+
+// UpdateUserOIDCRolesChannels saves the user's current OIDC roles/channels, and resulting implicit roles/channels,
+// to the database.
+func (auth *Authenticator) UpdateUserOIDCRolesChannels(user_ User, issuer string, newOIDCRoles, newOIDCChannels base.Set) error {
+	updateRolesChannelsCallback := func(currentPrincipal Principal) (updatedPrincipal Principal, err error) {
+		user, ok := currentPrincipal.(User)
+		if !ok {
+			return nil, base.ErrUpdateCancel
```
adamcfraser: Should this actually be ErrUpdateCancel?  If this somehow isn't a user should this return a non-recoverable error?

auth/auth.go
```diff
@@ -781,3 +831,58 @@ func (auth *Authenticator) RegisterNewUser(username, email string) (User, error)
...
+		// newOIDCRoles/newOIDCChannels will have come from the new issuer, so no need to revoke and re-add
+
+		oidcChannels := user.OIDCChannels()
+		if oidcChannels == nil {
+			oidcChannels = ch.TimedSet{}
+		}
+		if !oidcChannels.Equals(newOIDCChannels) {
+			oidcChannels.UpdateAtSequence(newOIDCChannels, nextSeq)
+			user.SetOIDCChannels(oidcChannels, nextSeq)
+			if err := auth.rebuildRoles(user); err != nil {
```
adamcfraser: Should this be rebuildChannels?

auth/auth.go
```diff
@@ -781,3 +831,58 @@ func (auth *Authenticator) RegisterNewUser(username, email string) (User, error)
...
+		}
+
+		oidcRoles := user.OIDCRoles()
+		if oidcRoles == nil {
+			oidcRoles = ch.TimedSet{}
+		}
+		if !oidcRoles.Equals(newOIDCRoles) {
+			oidcRoles.UpdateAtSequence(newOIDCRoles, nextSeq)
+			user.SetOIDCRoles(oidcRoles, nextSeq)
+			if err := auth.rebuildChannels(user); err != nil {
```
adamcfraser: Should this be rebuildRoles?

adamcfraser (changes requested):
Caught a few more things on this pass - have a look.

auth/auth.go
markspolakovs: getJWTClaimAsSet only returns an error if the claim is present, but has a type other than `string` or `[]string` (if it's simply absent it returns the empty set), which to me would indicate a misconfiguration on the provider side - in my head it'd be confusing why authentication succeeds but the user doesn't get the channels they're supposed to, so having it fail would highlight it faster (though I could definitely see the case for succeeding but logging the error).

auth/auth.go
markspolakovs: UpdateUserEmail does the same, so I used that as a reference - may be worth changing both then?

markspolakovs:
I know 56b44545871337830acede4c07e92c5dbbcbe057 is a bit big - let me know if you'd prefer it to be split into a separate PR.

db/users.go
```diff
@@ -26,46 +26,31 @@ import (
...
+// Note that read-only fields are omitted.
+func (p PrincipalConfig) AsPrincipalUpdates() auth.PrincipalUpdates {
+	chans := base.SetFromArray(p.ExplicitRoleNames)
+	return auth.PrincipalUpdates{
+		Name:              *p.Name,
+		ExplicitChannels:  &p.ExplicitChannels,
+		Email:             base.StringPtr(p.Email),
+		Disabled:          p.Disabled,
+		Password:          p.Password,
+		ExplicitRoleNames: &chans,
```
markspolakovs:
```suggestion
	roles := base.SetFromArray(p.ExplicitRoleNames)
	return auth.PrincipalUpdates{
		Name:              *p.Name,
		ExplicitChannels:  &p.ExplicitChannels,
		Email:             base.StringPtr(p.Email),
		Disabled:          p.Disabled,
		Password:          p.Password,
		ExplicitRoleNames: &roles,
```

auth/user.go
```diff
@@ -42,6 +42,9 @@ type userImplBody struct {
 	PasswordHash_    []byte          `json:"passwordhash_bcrypt,omitempty"`
 	OldPasswordHash_ interface{}     `json:"passwordhash,omitempty"` // For pre-beta compatibility
 	ExplicitRoles_   ch.TimedSet     `json:"explicit_roles,omitempty"`
+	OIDCRoles_       ch.TimedSet     `json:"oidc_roles,omitempty"`
+	OIDCChannels_    ch.TimedSet     `json:"oidc_channels,omitempty"`
+	OIDCIssuer_      string          `json:"oidc_issuer,omitempty"`
```
markspolakovs, line 48: As discussed, need to store the timestamp the OIDC roles/channels were last modified, so admins can know how stale they are.

rest/admin_api.go
```diff
@@ -1187,6 +1187,9 @@ func marshalPrincipal(princ auth.Principal, includeDynamicGrantInfo bool) ([]byt
 		if includeDynamicGrantInfo {
 			info.Channels = user.InheritedChannels().AsSet()
 			info.RoleNames = user.RoleNames().AllKeys()
+			info.OIDCIssuer = user.OIDCIssuer()
+			info.OIDCRoles = user.OIDCRoles().AllKeys()
+			info.OIDCChannels = user.OIDCChannels().AllKeys()
```
markspolakovs: As discussed, need to omit these if the issuer is no longer configured.

auth/auth.go
adamcfraser: I've thought about it further and think I agree (about fail fast).  In the scenario where claims are incorrectly defined on the provider side, it would be preferable for auth to fail, as opposed to replications moving forward with gaps in data because of (possibly temporary) issues with claim-based channel assignment.

adamcfraser:
I think I agree that it would be preferable to split https://github.com/couchbase/sync_gateway/commit/56b44545871337830acede4c07e92c5dbbcbe057 out into its own PR - there's enough in there that I'd like to review it alone and make sure I'm not overlooking anything.  Thanks!

auth/oidc.go
```diff
@@ -452,6 +458,35 @@ func formatUsername(value interface{}) (string, error) {
...
 
+// getJWTClaimAsSet looks up the given claim in the identity's claims, and transforms it to a set of strings.
+// The claim can be either a string or a slice of strings; other types will result in an error.
+// If the claim is missing from the identity, the empty set (and nil error) will be returned.
+func getJWTClaimAsSet(identity *Identity, claim string) (base.Set, error) {
+	val, ok := identity.Claims[claim]
+	if !ok {
+		return base.Set{}, nil
+	}
+
```
adamcfraser, line 469: Can we use the existing base.ValueToStringArray here, and then call base.SetFromArray on the response?  It seems like it would avoid some duplication on the underlying functionality, and it doesn't look like the approach below would add significant efficiency.

auth/auth.go
```diff
@@ -749,11 +773,23 @@ func (auth *Authenticator) authenticateOIDCIdentity(identity *Identity, provider
...
+
+	if user != nil {
+		now := time.Now()
+		updates = PrincipalConfig{
+			Name:            base.StringPtr(user.Name()),
+			Email:           &identity.Email,
+			OIDCIssuer:      &provider.Issuer,
+			OIDCRoles:       oidcRoles,
+			OIDCChannels:    oidcChannels,
+			OIDCLastUpdated: &now,
```
adamcfraser, line 788: Does OIDCLastUpdated have a functional purpose in the context of this PR, or is it in place for a future enhancement?

base/util.go
```diff
@@ -1704,3 +1704,14 @@ func CoalesceBools(a, b *bool) *bool {
 	}
 	return nil
 }
+
+// CoalesceTimes returns the first non-nil argument, or nil if both are nil.
```
adamcfraser, line 1708: I couldn't find a usage for this function in this PR?  Is it still useful/needed?

db/users.go
```diff
@@ -145,6 +140,31 @@ func (dbc *DatabaseContext) UpdatePrincipal(ctx context.Context, updates *auth.P
 			if updates.ExplicitRoleNames != nil && !updatedExplicitRoles.Equals(updates.ExplicitRoleNames) {
 				changed = true
 			}
+
+			if updates.OIDCIssuer != nil && *updates.OIDCIssuer != user.OIDCIssuer() {
```
adamcfraser, line 149: If the issuer changes, but updates.OIDCRoles (or updates.OIDCChannels) is nil, should we be setting those values to nil?

adamcfraser (changes requested):

auth/oidc.go
markspolakovs: Will do in the next commit. The only functional difference I can see is that we'd get more detailed error messages (currently `getJWTClaimAsSet` will tell you exactly which index was a non-string, whereas all we get from `ValueToStringArray` is the values of the non-strings).

auth/auth.go
markspolakovs: It's presented through the admin API (`/<db>/_user/<user>`) - it was the compromise we decided to avoid confusing admins when they see users with stale OIDC roels/channels.

base/util.go
markspolakovs: Good catch - should have been used in `PrincipalConfig.Merge` but looks like it got accidentally removed in a rebase.

markspolakovs: (This actually exposed a gap in the test coverage, as it looks like `Merge` wasn't exercised. I'll add an extra test that should hit this path.)

db/users.go
markspolakovs: Not sure I follow - `updates.OIDCRoles` being nil means "don't change this", am I missing something?

adamcfraser: As discussed, we just want a check that the OIDCIssuer/OIDCRoles/OIDCChannels are all or none on the incoming updates.

adamcfraser (changes requested):

adamcfraser (approved):

PR #5537, opened 2022-05-11
CBG-2050 Add option for read-only anonymous access

adamcfraser:
Adds an internal (unsupported) option to make anonymous access (via GUEST user) read-only.  Includes low-level enforcement immediately prior to sync function execution, but also performs an earlier check for 2.x and REST API calls.  REST API calls are identifying read-only calls based on the admin WriteAppData permission - this overloads that functionality somewhat, and would be a candidate for a wider refactoring of privileges for the REST API in future.

Also adds a missing userBlipHandler for MessageGetRev and MessagePutRev.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/259/

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -63,6 +63,10 @@ const (
 	maxHistoryEntriesPerGrant    = 10          // Ceiling of history entries count to ensure there is no more than 10 entries
 )
 
+const (
+	GuestUserReadOnly = "Anonymous access is read-only"
+)
+
 // Creates a new Authenticator that stores user info in the given Bucket.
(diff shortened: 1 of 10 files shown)


bbrks (approved):
LGTM

Should we extend this to on-demand import also? Or is that acceptable to do as guest, given the write is being run in the context of an admin?

adamcfraser:
Shouldn't be needed for on-demand import - as you say, that's not a new write, it's just addressing the timing gap between feed-based import and the GET request.

PR #5540, opened 2022-05-13
[3.0.1] CBG-2033 Option for xattr storage of persistent config

adamcfraser:
Adds a new unsupported config flag for system xattr storage of persistent database config.  Abstracts config storage using a new ConfigPersistence interface.  The existing document body storage is implemented by DocumentBootstrapPersistence, the new xattr storage by XattrBootstrapPersistence.  

The new implementation differentiates between the document cas (used for cas-safe updates) and the config cas (used to detect whether a config has changed).  The latter has been renamed cfgCas.  When using DocumentBootstrapPersistence, these values are the same.  When using xattr storage, cas is stamped into the xattr when the config changes via macro expansion, and this value is used as cfgCas. 

If a config document is deleted directly in the bucket when using XattrBootstrapPersistence, it will be recreated on next load (with preserved system xattr).  

Does not support backwards compatibility with non-xattr storage.

- [ ] `xattrs=true` http://uberjenkins.sc.couchbase.com:8080/view/Build/job/sync-gateway-integration/1597/
- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/263

### DIFF
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/couchbase/gocb/v2"
-	"github.com/couchbase/gocbcore/v10/memd"
 	"github.com/imdario/mergo"
 )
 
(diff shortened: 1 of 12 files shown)


bbrks (dismissed):
LGTM

PR #5546, opened 2022-05-19
Refactor db.UpdatePrincipal to take a struct from the auth package

markspolakovs:
In #5536, we need to be able to make changes to a user as a side-effect of OIDC signins. This would normally be done through `db.UpdatePrincipal`, but its parameter is a `db.PrincipalConfig` which we can't return from `auth` due to import cycles. Change its argument to be the new `auth.PrincipalUpdates` and update existing code accordingly.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/266/

### DIFF
--- a/auth/principal.go
+++ b/auth/principal.go
@@ -145,3 +145,68 @@ type User interface {
 
 	setRolesSince(ch.TimedSet)
 }
+
+// PrincipalUpdates represents a set of changes to a principal. It is meant for use with db.UpdatePrincipal, but is in
+// the auth package to avoid import cycles.
+// Name must always be specified, the other fields can be omitted to keep their current value.
+type PrincipalUpdates struct {
+	Name             string
+	ExplicitChannels *base.Set
+	// Users only
+	Email             *string
+	OIDCIssuer        *string
+	Password          *string
+	Disabled          *bool
+	ExplicitRoleNames *base.Set
+}
+
+// Merge returns a new PrincipalUpdates that represents the combination of both this and other's changes.
+// If any changes conflict, those of the other take precedence.
+func (u PrincipalUpdates) Merge(other PrincipalUpdates) PrincipalUpdates {
+	name := u.Name
+	if other.Name != "" {
+		name = other.Name
+	}
+	return PrincipalUpdates{
+		Name:              name,
+		ExplicitChannels:  base.CoalesceSets(other.ExplicitChannels, u.ExplicitChannels),
+		Email:             base.CoalesceStrings(other.Email, u.Email),
+		OIDCIssuer:        base.CoalesceStrings(other.OIDCIssuer, u.OIDCIssuer),
+		Password:          base.CoalesceStrings(other.Password, u.Password),
+		Disabled:          base.CoalesceBools(other.Disabled, u.Disabled),
+		ExplicitRoleNames: base.CoalesceSets(other.ExplicitRoleNames, u.ExplicitRoleNames),
+	}
+}
+
+// IsPasswordValid checks if the passwords in this PrincipalUpdates is valid.  Only allows
+// empty passwords if allowEmptyPass is true.
+func (u PrincipalUpdates) IsPasswordValid(allowEmptyPass bool) (isValid bool, reason string) {
+	// if it's an anon user, they should not have a password
+	if u.Name == "" {
+		if u.Password != nil {
+			return false, "Anonymous users should not have a password"
+		} else {
+			return true, ""
+		}
+	}
+
+	/*
+		if allowEmptyPass && ( u.Password == nil || len(*u.Password) == 0) {
+			return true, ""
+		}
+
+		if u.Password == nil || (u.Password != nil && len(*u.Password) < 3) {
+			return false, "Passwords must be at least three 3 characters"
+		}
+	*/
+
+	if u.Password == nil || len(*u.Password) == 0 {
+		if !allowEmptyPass {
+			return false, "Empty passwords are not allowed "
+		}
+	} else if len(*u.Password) < 3 {
+		return false, "Passwords must be at least three 3 characters"
+	}
+
+	return true, ""
+}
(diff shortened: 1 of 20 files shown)


db/users.go
```diff
@@ -20,52 +20,33 @@ import (
...
 	ExplicitChannels base.Set `json:"admin_channels,omitempty"`
-	Channels         base.Set `json:"all_channels,omitempty"`
 	// Fields below only apply to Users, not Roles:
 	Email             string   `json:"email,omitempty"`
 	Disabled          *bool    `json:"disabled,omitempty"`
 	Password          *string  `json:"password,omitempty"`
 	ExplicitRoleNames []string `json:"admin_roles,omitempty"`
 	RoleNames         []string `json:"roles,omitempty"`
+	// Fields below are read-only
+	Channels base.Set `json:"all_channels,omitempty"`
```
adamcfraser: RoleNames is also read-only.

db/users.go
```diff
@@ -20,52 +20,33 @@ import (
 	ch "github.com/couchbase/sync_gateway/channels"
 )
 
-// Struct that configures settings of a User/Role, for UpdatePrincipal.
-// Also used in the rest package as a JSON object that defines a User/Role within a DbConfig
-// and structures the request/response body in the admin REST API for /db/_user/*
+// PrincipalConfig represents a user/role as a JSON object.
+// Used to define a user/role within DbConfig, and structures the request/response body in the admin REST API
+// for /db/_user/*
```
adamcfraser: There seems to be some overlap between PrincipalConfig and PrincipalUpdates.  PrincipalUpdates doesn't include the read-only fields used for REST API responses, but aside from that there's a lot of commonality.  Did you consider more of a straight migration of PrincipalConfig from db to auth?  If so, what were the obstacles to that approach?

auth/principal.go
```diff
@@ -145,3 +145,68 @@ type User interface {
 
 	setRolesSince(ch.TimedSet)
 }
+
+// PrincipalUpdates represents a set of changes to a principal. It is meant for use with db.UpdatePrincipal, but is in
+// the auth package to avoid import cycles.
+// Name must always be specified, the other fields can be omitted to keep their current value.
+type PrincipalUpdates struct {
+	Name             string
+	ExplicitChannels *base.Set
```
adamcfraser: base.Set is already a map, so already supports nil values, and we should be able to differentiate between undefined and empty set without defining as a pointer to base.Set.

adamcfraser (changes requested):
Some questions - I feel like there might be an opportunity to use a single shared struct for PrincipalConfig and PrincipalUpdates - let me know what you think.

db/users.go
markspolakovs: It's an interesting thought, and I've pushed dcf8a24d101e88b9b13736893e7bcc7b0d0af8d3 which makes that change (and we can roll back if we decide it's not desirable). The downsides that I can see are:
* the read-only fields, as you called out (I had to update a couple tests that were incorrectly using `Channels` instead of `ExplicitChannels`)
* its structure becomes part of the admin API so we have less freedom to change it - having two separate structs maintains a separation between the public API and our implementation details
* it's a fairly noisy diff (though hopefully a one-off)

I don't feel strongly either way.

auth/principal.go
```diff
@@ -145,3 +145,69 @@ type User interface {
...
+
+// Merge returns a new PrincipalConfig that represents the combination of both this and other's changes.
+// If any changes conflict, those of the other take precedence.
+func (u PrincipalConfig) Merge(other PrincipalConfig) PrincipalConfig {
+	name := u.Name
+	if other.Name != nil {
+		name = other.Name
+	}
+	return PrincipalConfig{
+		Name:              name,
```
bbrks: Is there a reason why this is the only one not using the Coalesce functions? Looks functionally the same to me.

```suggestion
	return PrincipalConfig{
		Name:              base.CoalesceStrings(other.Name, u.Name),
``

Comment (markspolakovs) [inline auth/principal.go]:
```suggestion
	return PrincipalConfig{
		Name:              base.CoalesceStrings(other.Name, u.Name),
```

auth/principal.go
```diff
@@ -145,3 +145,69 @@ type User interface {
...
+// Used to define a user/role within DbConfig, and structures the request/response body in the admin REST API
+// for /db/_user/*
+type PrincipalConfig struct {
+	Name             *string  `json:"name,omitempty"`
+	ExplicitChannels base.Set `json:"admin_channels,omitempty"`
+	// Fields below only apply to Users, not Roles:
+	Email             *string  `json:"email,omitempty"`
+	Disabled          *bool    `json:"disabled,omitempty"`
+	Password          *string  `json:"password,omitempty"`
+	ExplicitRoleNames base.Set `json:"admin_roles,omitempty"`
```
bbrks, line 159: I see this field has changed from `[]string` to a map/set. What was the reason for that?

How are we handling backwards compatibility for the db config, user/role REST APIs or when loading existing data from the bucket?

bbrks (changes requested):

auth/principal.go
bbrks: I see you just left the exact same suggestion to yourself 😆

auth/principal.go
markspolakovs: I'm not sure, but I think the reason was consistency with `ExplicitChannels`. Can revert it, though if I'm not mistaken there shouldn't be a BWC issue since `base.Set` has a `json.Marshaler`/`json.Unmarshaler` that serialises it as an array anyway (and roles get stored as a set in `User` so having duplicates has always been semantically invalid).

bbrks: That's fine, didn't realise the unmarshaller worked that way. Thx

bbrks (approved):
This looks OK to me with that last clarification.

adamcfraser (approved):

PR #5552, opened 2022-05-23
CBG-2051: Fixed case sensitive internal property checking for removals and improve unit testing

IsaacLambat:
CBG-2051

- `processRev` purge on removal code is now case sensetive when checking for `_removed`
- Internal property tests now have a new upper case test case for testing internal properties
- Internal property tests now check the fields returned by the REST API 

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/277/ `1 flaky test fail`

### DIFF
--- a/rest/admin_api_test.go
+++ b/rest/admin_api_test.go
@@ -4556,61 +4556,57 @@ func TestApiInternalPropertiesHandling(t *testing.T) {
 	testCases := []struct {
 		name                        string
 		inputBody                   map[string]interface{}
-		expectReject                bool
+		expectedErrorStatus         *int // If nil, will check for 201 Status Created
 		skipDocContentsVerification *bool
 	}{
 		{
-			name:         "Valid document with special prop",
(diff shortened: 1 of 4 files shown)


rest/admin_api_test.go
```diff
@@ -4625,18 +4634,20 @@ func TestApiInternalPropertiesHandling(t *testing.T) {
 
 			resp := rt.SendAdminRequest("PUT", "/db/"+docID, string(rawBody))
 			if test.expectReject {
-				assert.NotEqual(t, 201, resp.Code)
+				assert.NotEqual(t, http.StatusOK, resp.Code)
```
bbrks: Wouldn't this still be a not created status?

But further, can't we assert on the actual status returned? E.g could `expectedErrorStatus` be a test case parameter, instead of `expectReject`?

bbrks (changes requested):
Change looks good - one nitpick in the existing tests, if you agree with it.

bbrks (approved):

PR #5554, opened 2022-05-23
CBG-2080: Stub new scopes/collections database config options

bbrks:
CBG-2080

Allows the new `scopes` and `collections` database config options to be used. Currently not wired up to anything.

Basic config validation is implemented:
- zero or one scope
- at least one collection under a scope
- valid js functions if set

Changes:
- Defined new `scopes` and `collections` config fields in database config struct.
- Extracted js function validation logic into function and unit tested.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/278
  - Flake: `--- FAIL: TestLogFlush/Default (20.14s)`

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -710,32 +710,33 @@ func (dbConfig *DbConfig) validateVersion(ctx context.Context, isEnterpriseEditi
 	// scopes and collections validation
 	if len(dbConfig.Scopes) > 1 {
 		multiError = multiError.Append(fmt.Errorf("only one named scope is supported, but had %d (%v)", len(dbConfig.Scopes), dbConfig.Scopes))
-	}
-	for scopeName, scopeConfig := range dbConfig.Scopes {
-		if len(scopeConfig.Collections) == 0 {
-			multiError = multiError.Append(fmt.Errorf("must specify at least one collection in scope %v", scopeName))
-			continue
-		}
-
-		if dbConfig.Sync != nil {
-			multiError = multiError.Append(errors.New("cannot specify a database-level sync function with named scopes and collections"))
-		}
-		if dbConfig.ImportFilter != nil {
-			multiError = multiError.Append(errors.New("cannot specify a database-level import filter with named scopes and collections"))
-		}
-
-		// validate each collection's config
-		for collectionName, collectionConfig := range scopeConfig.Collections {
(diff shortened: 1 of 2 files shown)


rest/config.go
```diff
@@ -705,6 +707,43 @@ func (dbConfig *DbConfig) validateVersion(ctx context.Context, isEnterpriseEditi
 		}
 	}
 
+	// scopes and collections validation
+	numScopes := 0
+	for scopeName, scopeConfig := range dbConfig.Scopes {
+		numScopes++
+		if numScopes > 1 {
+			multiError = multiError.Append(fmt.Errorf("only one named scope is supported, but had %d (%v)", len(dbConfig.Scopes), dbConfig.Scopes))
```
markspolakovs: How big do we expect `dbConfig.Scopes` to be? Would the `%v` make the error message too long to (easily) read?

rest/config.go
```diff
@@ -705,6 +707,43 @@ func (dbConfig *DbConfig) validateVersion(ctx context.Context, isEnterpriseEditi
 		}
 	}
 
+	// scopes and collections validation
+	numScopes := 0
+	for scopeName, scopeConfig := range dbConfig.Scopes {
+		numScopes++
+		if numScopes > 1 {
```
markspolakovs: Nitpicking: could this be `if len(dbConfig.Scopes) > 1` which removes the need for the `numScopes` variable?

markspolakovs (dismissed):
LGTM, just a couple questions (but happy for this to go in as is if Isaac is too)

rest/config.go
markspolakovs: Should the check be outside the loop, to only run it once?

rest/config.go
IsaacLambat: This would cause a lot of error spam if there are a few scopes defined. Would be good to get it so this only counts as one error regardless of the amount of scopes defined.

IsaacLambat (changes requested):
Would be good to not repeat the same error message multiple times when multiple scopes are defined while still bundling all the errors from the first scope as well (to save the user from restarting as you have done currently).

rest/config.go
bbrks: Don't expect it to grow much more than 1 scope. It's explicitly defined by the database config, so not going to expand to all available scopes or anything.

IsaacLambat (approved):
LGTM!

PR #5556, opened 2022-05-24
User N1QL queries, GraphQL queries, JS functions

snej:
CBG-2106

Full implementation of User Queries (my name; I don't know if there is an official name for the whole feature.)
- Named N1QL queries with parameters
- GraphQL queries
- Named JavaScript functions with parameters

These have both REST and BLIP (Connected Client) APIs.

- [Spec](https://docs.google.com/document/d/1ggxUJhotM0-Ww8CnM69HKBcKbboQyw5VeWLUkdp8sUA/edit?usp=sharing)
- [Developer documentation draft](https://docs.google.com/document/d/1sskhijMOlviqhY_AJE7_e7W1ue1J32BCfKLt_PKnzhU/edit?usp=sharing)

(these docs are still restricted; ask Wayne if you want access.)


db/blip_connected_client.go
```diff
@@ -68,3 +76,68 @@ func (bh *blipHandler) handlePutRev(rq *blip.Message) error {
...
+		}
+	}
+
+	bh.logEndpointEntry(rq.Profile(), fmt.Sprintf("name: %s", name))
+
+	// In the N1QL string, replace "BUCKET" with the actual bucket name, wrapped in back-quotes in
+	// case it's not a valid N1QL identifier (e.g. `travel-sample`.)
+	// TODO: This is a workaround for the inability to use back-quotes for this in an SG config
+	// file, because those are already reserved for quoting long literals like the sync function.
+	statement := strings.ReplaceAll(query.Statement, "BUCKET", "`"+bh.db.Name+"`")
```
bbrks: Arbitrary keyspaces in queries feels like a footgun waiting to happen, especially given the stronger restrictions around data access with multi-tenancy and cloud.

Per-doc/channel authentication I acknowledge is still a remaining task, but this one feels like a bigger issue.

What is the approach going to be here to ensure the query being run is operating on the bucket or scope associated with the database?

db/blip_connected_client.go
```diff
@@ -68,3 +76,68 @@ func (bh *blipHandler) handlePutRev(rq *blip.Message) error {
...
+
+	// Run the query:
+	results, err := bh.db.N1QLQueryWithStats(bh.db.Ctx, QueryTypeConnectedClientPrefix+name, statement, params,
```
bbrks: This will be already doing a `$_keyspace` replacement for the bucket/collection associated with the database, so I don't think there's a requirement to do it above.

db/blip_connected_client.go
```diff
@@ -68,3 +76,68 @@ func (bh *blipHandler) handlePutRev(rq *blip.Message) error {
...
+			return base.HTTPErrorf(http.StatusInternalServerError, "Query error: %s", qe.Errors[0].Message)
+		} else {
+			base.WarnfCtx(bh.loggingCtx, "Unknown error running query %q: %T %#v", name, err, err)
+			return base.HTTPErrorf(http.StatusInternalServerError, "Unknown error running query")
+		}
+	}
+
+	// Write the results to the response:
+	var out bytes.Buffer
+	enc := json.NewEncoder(&out)
```
bbrks: Util func that wraps different JSON libraries for CE/EE

```suggestion
	enc := base.JSONEncoder(&out)
```

db/blip_connected_client.go
```diff
@@ -68,3 +76,68 @@ func (bh *blipHandler) handlePutRev(rq *blip.Message) error {
...
+	var row interface{}
+	for results.Next(&row) {
+		enc.Encode(row) // always ends with a newline
+	}
```
bbrks, line 187: The query result needs closing after iteration to free up the `concurrentQueryOpLimitChan` queue and grab any errors from the result iteration.

```suggestion
	}
	if err = results.Close(); err != nil {
		return err
	}
```

bbrks (changes requested):
A few small comments, and a question about how we're going to be dealing with multi-tenancy/cloud with customer-defined queries.

That seems like a more fundamental topic than the filtering of doc/channels mentioned in the PR description.

db/blip_connected_client.go
snej: So should the query in the config file just use `$$_keyspace` as a placeholder for the bucket?

snej: Hm, I tried doing the above and got N1QL parse errors at runtime, so I guess that's not it.

snej: Ah, the real problem is that the substitution doesn't escape the bucket name, which in my case is `travel-sample`. Adding a fix for that as a separate commit.

snej:
I just pushed another commit that implements "Authorization with User Security Context" as per the spec.

db/blip_connected_client.go
adamcfraser: For multi-tenancy the we require db-scoped RBAC, and rely on the mobile_sync_gateway role (or similar query service roles) only being scoped to the target bucket.  That's required here, where we don't have an additional layer of protection at the SG level (since query operations are performed at the cluster level), and it's possible for an administrator to define a query that's not using `$_keyspace`.

adamcfraser:
@snej Given that we've added REST endpoint support, can we add a feature/unsupported flag for this functionality?  With that in place we'll be able to merge this.

rest/routing.go
```diff
@@ -109,6 +109,13 @@ func createCommonRouter(sc *ServerContext, privs handlerPrivs) (root, db, keyspa
 
 	dbr.Handle("/_blipsync", makeHandler(sc, privs, []Permission{PermWriteAppData}, nil, (*handler).handleBLIPSync)).Methods("GET")
 
+	// User queries & functions
+	if sc.config.Unsupported.UserQueries != nil && *sc.config.Unsupported.UserQueries {
+		dbr.Handle("/_function/{name}", makeHandler(sc, privs, []Permission{PermReadAppData}, nil, (*handler).handleUserFunction)).Methods("GET", "POST")
```
adamcfraser, line 114: Do we support functions that can mutate documents?  If so, this should be PermWriteAppData.

adamcfraser: Looking at handleUserFunction, I see we're only supporting writes on POST.  I think we should split this out into two handlers here, with PermWriteAppData on POST.

db/user_query_n1ql.go
```diff
@@ -0,0 +1,82 @@
...
+		args["user"] = map[string]interface{}{}
+	}
+
+	// Check that the user is authorized:
+	if err := query.Allow.authorize(db.user, args, "query", name); err != nil {
+		return nil, err
+	}
+
+	// Run the query:
+	iter, err := db.N1QLQueryWithStats(db.Ctx, QueryTypeUserPrefix+name, query.Statement, args,
```
adamcfraser, line 67: Are we wanting to record stats for user-defined queries?  If yes, I think there's some work involved to register queries for stats tracking (initQueryStats), in particular for prometheus.

However, I'm on the fence about whether we should even include user queries in those stats at this point.  I can see where stats on users query performance would be useful, but need to think about the potential bloating of stats logging, and whether user queries should have their own place in stats logging.  The closest parallel I think we have today is ISGR, where we log stats for each user-created replication.  Again, there's some extra work needed for stats that come and go (when the query is created/deleted) to avoid breaking things on the Prometheus side.

rest/query_api.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return err
+	}
+	// Run the query:
+	return h.db.WithTimeout(db.UserQueryTimeout, func() error {
+		results, err := h.db.UserN1QLQuery(queryName, queryParams)
+		if err != nil {
+			return err
+		}
+		if err = h.db.CheckTimeout(); err != nil {
```
adamcfraser: It looks like we checked the timeout as we exited UserN1QLQuery - is this check just defensive in case of future changes?

rest/query_api.go
```diff
@@ -0,0 +1,178 @@
...
+			}
+			if err = h.addJSON(row); err != nil {
+				return err
+			}
+			// The iterator streams results as the query engine produces them, so this loop may take most of the query's time; check for timeout after each iteration:
+			if err = h.db.CheckTimeout(); err != nil {
+				return err
+			}
+		}
+		if err = results.Close(); err != nil {
```
adamcfraser: We need to call results.Close() in all cases - otherwise we leave an open connection for the query and an entry in the channel we use to throttle concurrent query ops. So probably want to stick this in a defer, or don't return early on the errors inside results iteration.

db/database.go
```diff
@@ -1738,3 +1753,31 @@ func (context *DatabaseContext) IsGuestReadOnly() bool {
 func (context *DatabaseContext) ForceAPIForbiddenErrors() bool {
 	return context.Options.UnsupportedOptions != nil && context.Options.UnsupportedOptions.ForceAPIForbiddenErrors
 }
+
+//////// TIMEOUTS
+
+// Calls a function, synchronously, while imposing a timeout on the Database's Context. Any call to CheckTimeout while the function is running will return an error if the timeout has expired.
+// The function will *not* be aborted automatically! Its code must check for timeouts by calling CheckTimeout periodically, returning once that produces an error.
+func (db *Database) WithTimeout(timeout time.Duration, operation func() error) error {
```
adamcfraser, line 1845: I think that even in the case of a connected websocket client, we're creating separate *Database instances per handler.  Given that, is it necessary to restore the previous context on exit, or can we simplify this to either (a) just swap the context to add a timeout, and avoid having to invoke operation as a callback, or (b) specify the timeout when we create the db for the blipHandlers and rest handlers that should have their own timeouts.

adamcfraser: Alternatively, what do you think about passing a completely separate context as a parameter to CallUserFunction, UserN1QLQuery, etc.  That seems like a clean way to have those functions manage function-specific timeouts, without having to worry about a potentially shared *Database instance.

db/user_query.go
```diff
@@ -0,0 +1,184 @@
...
+		for _, rolePattern := range allow.Roles {
+			if role, err := allow.expandPattern(rolePattern, args, user); err != nil {
+				return err
+			} else if userRoles.Contains(role) {
+				return nil // User has one of the allowed roles
+			}
+		}
+		// Check if the user has access to one of the given channels.
+		for _, channelPattern := range allow.Channels {
+			if channelPattern == channels.AllChannelWildcard {
```
adamcfraser, line 108: Should specifying wildcard in allow.Channels be supported?  Based on this implementation it looks like it's functionally equivalent to omitting the entire allow section.

db/user_query.go
```diff
@@ -0,0 +1,184 @@
...
+		return base.HTTPErrorf(http.StatusNotFound, "no such %s %q", what, name)
+	} else {
+		return user.UnauthError(fmt.Sprintf("you are not allowed to call %s %q", what, name))
+	}
+}
+
+// Expands patterns of the form `$param` or `$(param)` in `pattern`, looking up each such
+// `param` in the `args` map and substituting its value.
+// (`$$` is replaced with `$`.)
+// It is an error if any `param` has no value, or if its value is not a string or integer.
```
adamcfraser, line 134: Out of curiosity, is there an underlying reason to not support float parameters?

db/user_query_graphql.go
```diff
@@ -0,0 +1,213 @@
...
+	ctx = context.WithValue(ctx, mutAllowedKey, mutationAllowed)
+	result := graphql.Do(graphql.Params{
+		Schema:         gq.schema,
+		RequestString:  query,
+		VariableValues: variables,
+		OperationName:  operationName,
+		Context:        ctx,
+	})
+	if len(result.Errors) > 0 {
+		// ??? Is this worth logging?
```
adamcfraser: If it's not definitely needed, should bump down to info level logging to avoid triggering alerts based on warnings.

db/user_function.go
```diff
@@ -0,0 +1,90 @@
...
+		return nil, err
+	}
+
+	// Compile and run the function:
+	compiled := config.compiled
+	if compiled == nil {
+		compiled = newUserFunctionJSServer(name, "user function", "args, context", config.SourceCode)
+		// This is a race condition if two threads find the script uncompiled and both compile it.
+		// However, the effect is simply that two identical UserFunction instances are created and
+		// one of them (the first one stored to config.compiled) is thrown away; harmless.
```
adamcfraser: Even if it's functionally harmless, this sounds like it's going to make go's race detector unhappy (which will make unit tests unhappy under race).  Since there aren't any error cases to worry about in newUserFunctionJSServer, could we use sync.Once to handle concurrency here?  Alternately just drop the lazy loading altogether and compile all functions on dbContext creation?

db/user_function.go
```diff
@@ -0,0 +1,90 @@
...
+func newUserFunctionJSServer(name string, what string, argList string, sourceCode string) *sgbucket.JSServer {
+	js := fmt.Sprintf(kJavaScriptWrapper, argList, sourceCode)
+	return sgbucket.NewJSServer(js, 0, kUserFunctionCacheSize,
+		func(fnSource string, timeout time.Duration) (sgbucket.JSServerTask, error) {
+			return newUserJavaScriptRunner(name, what, fnSource)
+		})
+}
+
+// Number of Otto contexts to cache per function, i.e. the number of goroutines that can be simultaneously running each function.
+const kUserFunctionCacheSize = 2
```
adamcfraser: What's the behaviour if more than two goroutines try to invoke the function at once?  If I'm remembering an issue we had with this for the sync function correctly (CBG-236), on cache overflow it's going to initialize a new runner (as opposed to blocking waiting for an available runner in the cache).
It may be that this is a more general problem we need to solve around being able to throttle the number of concurrently executing js functions.  If you agree, can you file a ticket for that enhancement to make sure it doesn't get overlooked?

db/user_js_runner.go
```diff
@@ -0,0 +1,384 @@
...
+			// Not-found is not an error; just return null.
+			return nil, nil
+		}
+		return nil, err
+	}
+	body, err := rev.Body()
+	if err != nil {
+		return nil, err
+	}
+	if docType != nil && body["type"] != *docType {
```
adamcfraser, line 223: Are we relying on the fixed property name "type" to identify doc types?  This seems like something that's less likely to be present when using collections.
Looking closer I see we're never actually calling this function with a docType - is this code that should be removed, or is there a pending enhancement planned?

rest/query_api.go
```diff
@@ -0,0 +1,178 @@
...
+	if h.rq.Method == "POST" {
+		// POST: Params come from the request body in JSON format:
+		err = h.readJSONInto(&params)
+	} else {
+		// GET: Params come from the URL queries (`?key=value`):
+		for key, values := range h.getQueryValues() {
+			// `values` is an array of strings, one per instance of the key in the URL
+			if len(values) > 1 {
+				err = base.HTTPErrorf(http.StatusBadRequest, "Duplicate parameter '%s'", key)
+				return
```
adamcfraser: We generally try to avoid naked returns in functions that are more than a few lines, to improve code readability.

rest/routing.go
```diff
@@ -294,6 +301,32 @@ func CreateAdminRouter(sc *ServerContext) *mux.Router {
...
+
+		dbr.Handle("/_config/queries",
+			makeOfflineHandler(sc, adminPrivs, []Permission{PermUpdateDb}, nil, (*handler).handleGetDbConfigQueries)).Methods("GET")
+		dbr.Handle("/_config/queries/{query}",
+			makeOfflineHandler(sc, adminPrivs, []Permission{PermUpdateDb}, nil, (*handler).handleGetDbConfigQuery)).Methods("GET")
+		dbr.Handle("/_config/queries",
+			makeOfflineHandler(sc, adminPrivs, []Permission{PermUpdateDb}, nil, (*handler).handlePutDbConfigQueries)).Methods("PUT", "DELETE")
+		dbr.Handle("/_config/queries/{query}",
+			makeOfflineHandler(sc, adminPrivs, []Permission{PermUpdateDb}, nil, (*handler).handlePutDbConfigQuery)).Methods("PUT", "DELETE")
+	}
```
adamcfraser, line 329: The API spec (under docs/api) should be updated with these new endpoints, including the supported parameters for each.  Not a blocker for this PR (and could be delegated), but please file a ticket to track that work.

rest/query_api.go
```diff
@@ -0,0 +1,178 @@
...
+				return err
+			}
+			queryString = string(query)
+		} else {
+			// POST JSON: Get the "query", "operationName" and "variables" properties:
+			body, err := h.readJSON()
+			if err != nil {
+				return err
+			}
+			queryString = body["query"].(string)
```
adamcfraser: It's not necessarily consistent across the codebase, but in general it's preferable to define these parameter name strings as constants to simplify maintenance across code and tests.  Can just define at the top of this file, there isn't a centralized set of definitions.

rest/handler_config_database.go
```diff
@@ -0,0 +1,131 @@
...
+	} else {
+		// Update in-memory config read from JSON file:
+		databaseConfig := *h.server.GetDatabaseConfig(dbName)
+
+		if h.headerDoesNotMatchEtag("If-Match", databaseConfig.Version) {
+			return base.HTTPErrorf(http.StatusPreconditionFailed, "Provided If-Match header does not match current config version")
+		}
+
+		// Now call the mutator function:
+		if err := mutator(&databaseConfig.DbConfig); err != nil {
```
adamcfraser: This looks like a potential race modifying serverContext.dbConfigs[] for concurrent invocation of the config mutators.  Even leaving the data race aside it doesn't look like the etag handling would be providing the expected guarantees.

rest/config_legacy.go
```diff
@@ -98,6 +98,7 @@ type UnsupportedServerConfigLegacy struct {
 	Http2Config           *HTTP2Config `json:"http2,omitempty"`               // Config settings for HTTP2
 	StatsLogFrequencySecs *uint        `json:"stats_log_freq_secs,omitempty"` // How often should stats be written to stats logs
 	UseStdlibJSON         *bool        `json:"use_stdlib_json,omitempty"`     // Bypass the jsoniter package and use Go's stdlib instead
+	UserQueries           *bool        `json:"user_queries,omitempty"`        // Enable user N1QL/GraphQL/JS queries
```
adamcfraser: Do we want to support this when using legacy config?  The current intent of legacy config is to support customer upgrade cases.  Given the extensive API for administering query/function definitions at runtime (which wouldn't be persisted through restart for legacy config), I'm thinking we should force people to use persistent configs with this feature.

db/user_js_runner.go
```diff
@@ -0,0 +1,384 @@
...
+//////// DATABASE CALLBACK FUNCTION IMPLEMENTATIONS:
+
+// Implementation of JS `app.func(name, params)` function
+func (runner *userJSRunner) do_func(funcName string, params map[string]interface{}) (interface{}, error) {
+	return runner.currentDB.CallUserFunction(funcName, params, runner.mutationAllowed)
+}
+
+// Implementation of JS `app.get(docID, docType)` function
+func (runner *userJSRunner) do_get(docID string, docType *string) (interface{}, error) {
+	if err := runner.currentDB.CheckTimeout(); err != nil {
```
adamcfraser, line 207: For my own understanding - when app.get() returns a timeout error, the calling (customer-defined) javascript function has the option to catch that error and not necessarily return the timeout to the client, is that correct?

db/user_js_runner.go
```diff
@@ -0,0 +1,384 @@
...
+	if !runner.mutationAllowed {
+		return base.HTTPErrorf(http.StatusForbidden, "a read-only request is not allowed to mutate the database")
+	}
+
+	// TODO: Currently this is "last writer wins": get the current revision if any, and pass it
+	// to Put so that the save always succeeds.
+	rev, err := runner.currentDB.GetRev(docID, "", false, []string{})
+	if err == nil {
+		body["_rev"] = rev.RevID
+	}
```
adamcfraser: As written this still has the possibility that someone else updates the rev in between Get and Put.  Are we ok returning 409 to the caller of do_save in this case?

adamcfraser (changes requested):
Overall implementation looks great, added comments for a handful of implementation details that may need a tweak.

db/database.go
snej: I'm OK with not restoring the old Context on the way out, if a Database instance isn't going to outlive the operation that sets a timeout.

Having a separate Context seems error-prone since the Database already has a Context, and one might forget which is the right one to use.

db/user_query.go
snej: No, if the Allow is missing then only admins have access. (See line 94.)

db/user_query.go
snej: It didn't seem likely someone would have a channel name pattern that used decimals, but I suppose it's possible, and it's easy enough to add a check for `CanFloat`.

db/user_query_graphql.go
snej: I think I'll take it out.

db/user_function.go
snej: I think early compilation is best. That way if a function has a syntax error you'll find out when you start the server / update the config, not just the first time someone tries to call that function.

db/user_js_runner.go
snej: `docType` is an optional parameter to the JS function:
- `get(id)` just gets the document with that docID.
- `get(id, docType)` gets the document but also verifies that its `type` property has the expected value.
I found this useful in implementing the travel-sample schema; for instance the implementation of the resolver for `airport(id:ID)` is `{return context.app.get(args.id, 'airport');}`. That way if the caller passes the ID of a document that isn't an airport, the resolver fails instead of violating type-safety.

db/user_js_runner.go
snej: You're right. Is there a method that will just slam a new revision without MVCC, or do I need to add a loop that retries when it gets a conflict?

adamcfraser: No I don't think we've got any other use cases for this type of LWW - probably easiest to just add retry loop.

db/user_js_runner.go
snej: I hadn't considered that. Hmm. 

In principle, it doesn't matter because any native function it calls after that will keep throwing a timeout error, and even Otto's Interrupt handler will panic the next time it's called (and that panic does _not_ get intercepted by JS; it just flies straight out of Otto into our code.)

However, the Context timeout mechanism works by reading a value from a channel. Does that mean that you only get one chance to find out about the timeout, and if you keep going you won't receive another value so the timeout checks won't trigger anymore?

db/user_js_runner.go
adamcfraser: I can see how this would be useful, was just wondering about decoupling it from the from the hardcoded "type" property for compatibility with existing datasets that use a different property to define type.  If this is primarily used for testing at the moment, can leave that as a future (possible) enhancement?

db/user_js_runner.go
adamcfraser: WithTimeout is going to close the context's Done channel, so that's safe for multiple reads.

db/user_query.go
adamcfraser: Ah I'd missed that - so wildcard on channels is actually the only way to define that it's available to anyone on the public port?  That feels a little bit unintuitive, as technically users don't need to be granted any channels, so the wildcard is matching the empty set.  I don't think `*` is a valid username, so we could consider using the wildcard in allow.Users instead for this case?

db/database.go
adamcfraser: That sounds reasonable - let's just leave it as-is on the way out.

rest/config_legacy.go
snej: OK, I've removed it.

rest/query_api.go
snej: This has been fixed.

rest/query_api.go
snej: This has been fixed.

rest/routing.go
snej: This has been fixed.

db/user_query.go
snej: We'll need to include Wayne, since this was his design.

db/user_query.go
snej: This has been added.

db/user_query_graphql.go
snej: This has been fixed.

db/user_function.go
snej: This has been fixed.

db/user_function.go
snej: I filed CBG-2358.

db/user_js_runner.go
snej: I was under the impression that using `type` was very common ... if it's not, I can take this feature out.

rest/config_legacy.go
snej: On the other hand, this change seems to make testing the admin API a lot harder. At least, I've been banging my head against how to create a RestTester with persistent config ... I can create it, but can't figure out how to successfully add a database to it.

rest/handler_config_database.go
snej: Then the `PUT /db/` handler (`handler.handlePutDbConfig()`) has the same bug, because I copied the code from there. (Except it doesn't support Etags at all; I added that.)

I tried fixing the race condition by locking `h.server.lock`, but that causes deadlocks in the handler because it's not a recursive mutex.

db/user_js_runner.go
snej: Fixed. (And added unit tests for this.)

db/user_query.go
adamcfraser: Fine with leaving this as-is for now, can revisit as needed.

db/user_js_runner.go
adamcfraser: We can leave "but I use 'docType' instead of 'type'" as a problem to be addressed with a future enhancement as needed.

rest/routing.go
adamcfraser: Filed CBG-2366 to track.

rest/handler_config_database.go
adamcfraser: It looks like this is handled for the persistent config case (UpdateConfig's cas handling is going to ensure the etag checking is valid).  I don't think the potential race in the non-persistent config case (which should be test-only?) requires any additional attention.

adamcfraser (dismissed):

adamcfraser:
@snej This looks good - errcheck flagged a couple of missed error checks.  Let me know when those are fixed and it's rebased and we can merge.

snej:
The failed sgw-pipeline-errcheck is from the previous push -- the timestamp on it is yesterday (9/6). Why wasn't it re-run?

PR #5557, opened 2022-05-27
CBG-2055 Do not evaluate revpos during storeAttachments

adamcfraser:
CBG-2055
revpos shouldn't be used as a criteria for attachment persistence for incoming writes. In particular during storeAttachments, a variety of circumstances can result in changes to the incoming revpos, and we don't want to fail to persist or remove attachments based on revpos.

In addition to removing the revpos check in storeAttachments, adds some additional handling in blip_handler for different flavors of revpos being sent by CBL to ensure it's being set to the incoming revID when the attachment doesn't exist.

- [x] http://uberjenkins.sc.couchbase.com:8080/view/Build/job/sync-gateway-integration/1599/

### DIFF
--- a/db/attachment.go
+++ b/db/attachment.go
@@ -290,12 +290,18 @@ func (db *Database) ForEachStubAttachment(body Body, minRevpos int, docID string
 			if err != nil && !base.IsDocNotFoundError(err) {
 				return err
 			}
-			if newData, err := callback(name, digest, data, meta); err != nil {
+			newData, err := callback(name, digest, data, meta)
+			if err != nil {
 				return err
-			} else if newData != nil {
+			}
+			if newData != nil {
 				meta["data"] = newData
 				delete(meta, "stub")
 				delete(meta, "follows")
+			} else {
+				// Update version in the case where this is a new attachment on the doc sharing a V2 digest with
+				// an existing attachment
+				meta["ver"] = AttVersion2
 			}
 		}
 	}
--- a/db/attachment.go
+++ b/db/attachment.go
@@ -104,23 +104,18 @@ func (db *Database) storeAttachments(doc *Document, newAttachmentsMeta Attachmen
 			if meta["stub"] != true {
 				return nil, base.HTTPErrorf(400, "Missing data of attachment %q", name)
 			}
-
-			revpos, ok := base.ToInt64(meta["revpos"])
-			if !ok || revpos < 1 {
-				return nil, base.HTTPErrorf(400, "Missing/invalid revpos in stub attachment %q", name)
-			}
 			// Try to look up the attachment in ancestor attachments
 			if parentAttachments == nil {
 				parentAttachments = db.retrieveAncestorAttachments(doc, parentRev, docHistory)
 			}
 
+			// Note: in a non-conflict CAS retry, parentAttachments may be nil, because the attachment
+			//  data was persisted prior to the CAS failure writing the doc.  In this scenario the
+			//  incoming doc attachment metadata has already been updated to stub=true to avoid attempting to
+			//  persist the attachment again, even though there is not an attachment on an ancestor.
 			if parentAttachments != nil {
 				if parentAttachment := parentAttachments[name]; parentAttachment != nil {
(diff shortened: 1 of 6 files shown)


rest/blip_api_test.go
```diff
@@ -3351,16 +3351,76 @@ func TestCBLRevposHandling(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	rev1ID := "1-ca9ad22802b66f662ff171f226211d5c"
+	_, ok := btc.WaitForRev("doc1", rev1ID)
+	require.True(t, ok)
+
+	// force attachment into test client's store to validate it's fetched
+	attachmentAData := base64.StdEncoding.EncodeToString([]byte("attachmentA"))
+	contentType := "text/plain"
+	length, digest, err := btc.saveAttachment(contentType, attachmentAData)
```
bbrks, line 3409:
```suggestion
	length, digest, err := btc.saveAttachment(contentType, attachmentAData)
	require.NoError(t, err)
```

bbrks (commented):
Looks OK to me but I would like @IsaacLambat or @JRascagneres to take a closer look/testing given they've been in this area more recently than me.

db/attachment.go
```diff
@@ -104,23 +104,18 @@ func (db *Database) storeAttachments(doc *Document, newAttachmentsMeta Attachmen
...
 			if parentAttachments == nil {
 				parentAttachments = db.retrieveAncestorAttachments(doc, parentRev, docHistory)
 			}
 
+			// Note: in a non-conflict CAS retry, parentAttachments may be nil, because the attachment
+			//  data was persisted prior to the CAS failure writing the doc.  In this scenario the
+			//  incoming doc attachment metadata has already been updated to stub=true to avoid attempting to
+			//  persist the attachment again, even though there is not an attachment on an ancestor.
 			if parentAttachments != nil {
 				if parentAttachment := parentAttachments[name]; parentAttachment != nil {
```
IsaacLambat, line 117: If an attachment name is changed in the document, this check fails. This causes the version to not be stamped in, and then the attachment gets deleted from the bucket due to the obsolete attachment handling (as it can't find the attachment due to the incorrect version).

IsaacLambat (changes requested):
PR looks good, and change has solved the problem caused by the incorrect revpos. However, changing an attachment name via Blip (regardless of revpos) causes the attachment to be deleted from the bucket still.

Will create a new branch based on this branch with a unit test (and possibly an attempted fix) for the attachment name change scenario.

db/attachment.go
adamcfraser: What do you mean by 'attachment name changed in the bucket'?  What's the scenario where that would happen?

IsaacLambat: Sorry my mistake. Corrected initial comment now.

adamcfraser: Sounds good.  Have pushed up a tweak to cover that case.

rest/api_test.go
```diff
@@ -8034,57 +8034,8 @@ func TestBasicAttachmentRemoval(t *testing.T) {
...
-            "digest": "sha1-witfkXg0JglCjW9RssWvTAveakI="
-         }
-      },
-      "time_saved": "2021-09-01T17:33:03.054227821Z"
-   },
-  "key": "value"
-}`)
-	}
-
 	createDocWithLegacyAttachment := func(docID string, rawDoc []byte, attKey string, attBody []byte) {
```
IsaacLambat, line 8037: Nitpick: Can the calls to this inline function be removed and calls to the inline one be replaced with the new one (by passing the testing and rt variables)?

IsaacLambat (dismissed):
Minor nitpick but can be ignored.
LGTM

rest/api_test.go
adamcfraser: Yeah, when this gets ported to master I think that would be reasonable.  Omitted for this version to reduce the changes made on 3.0.3.

adamcfraser:
Reviewed integration test results, failures are unrelated to these changes.

PR #5559, opened 2022-05-30
CBG-1692: Make SGCollect load config from config file if SGW is inaccessible

markspolakovs:
CBG-1692

sgcollect tries to determine the log file path from the SG config, which it normally gets through /_config. If SG isn't running, make it also try loading it from the JSON config file on disk - even in 3.x, where it's only bootstrap config, it'll still contain the logging details for it.

Tested it locally against a Helium and 2.8.3 SG, but not further - no automated testing for sgcollect (cf. [CBG-1889](https://issues.couchbase.com/browse/CBG-1889)).

### DIFF
--- a/tools/sgcollect_info
+++ b/tools/sgcollect_info
@@ -229,7 +229,7 @@ def urlopen_with_basic_auth(url, username, password):
     else:
         return urllib.request.urlopen(url)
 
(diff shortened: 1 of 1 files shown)


JRascagneres (approved):
Seems to make sense

PR #5561, opened 2022-06-03
CBG-2100 allow x509 tests to use under jenkins

torcolvin:
The bulk of these changes allow copy of CA certs and configuration into
a docker container, which is used by jenkins.

- Add jenkins output files to gitignore
- Make sure three ways of running X509 are supposed: couchbase server on
  docker, on localhost (OS X only), ssh to remote couchbase
- Some changes to jenkins script
	- allow set -u to tell what variables are missing. Use bash {:-}
	  syntax for potentially unset variables
	- run through shfmt
	- change git config --add to --replace-all to not write infinite
	  options to .gitconfig
- make sure all files produced by integration tests are marked in
  .gitignore

### DIFF
--- a/jenkins-integration-build.sh
+++ b/jenkins-integration-build.sh
@@ -1,7 +1,9 @@
 #!/usr/bin/env bash
 DEFAULT_PACKAGE_TIMEOUT="45m"
 
-if [ "$1" == "-m" ]; then
+set -u
+
+if [ "${1:-}" == "-m" ]; then
     echo "Running in automated master integration mode"
     # Set automated setting parameters
     SG_COMMIT="master"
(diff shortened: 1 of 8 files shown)


jenkins-integration-build.sh
```diff
@@ -14,10 +16,11 @@ if [ "$1" == "-m" ]; then
     XATTRS="true"
     RUN_COUNT="1"
     # CBS server settings
-    COUCHBASE_SERVER_PROTOCOL="couchbase"
+    COUCHBASE_SERVER_PROTOCOL="couchbases"
```
adamcfraser: Remove these for now (until we're turning this on for master run)

adamcfraser (changes requested):

adamcfraser (approved):

PR #5562, opened 2022-06-08
CBG-2119: Fix resetting disable_password_auth to false

markspolakovs:
CBG-2119

`DbConfig.DisablePasswordAuth` was declared as `bool` rather than `*bool`, meaning that when an API caller POSTed `{"disable_password_auth": true}` it was set correctly, but when they POSTed `{"disable_password_auth": false}` Go treated it as the zero value because of `omitempty` and thus did not update the config. So you could enable password auth, but never disable it.

This PR changes its type to `*bool` ensuring that `false` is distinguished from absence. Our tests exercised the feature itself, but not the config manipulation - this PR adds a regression test.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/304/

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -138,7 +138,7 @@ type DatabaseContextOptions struct {
 	SessionCookieHttpOnly         bool             // Pass-through DbConfig.SessionCookieHTTPOnly
 	AllowConflicts                *bool            // False forbids creating conflicts
 	SendWWWAuthenticateHeader     *bool            // False disables setting of 'WWW-Authenticate' header
-	DisablePasswordAuthentication *bool            // True enforces OIDC/guest only
+	DisablePasswordAuthentication bool             // True enforces OIDC/guest only
 	UseViews                      bool             // Force use of views
 	DeltaSyncOptions              DeltaSyncOptions // Delta Sync Options
 	CompactInterval               uint32           // Interval in seconds between compaction is automatically ran - 0 means don't run
--- a/db/database.go
+++ b/db/database.go
@@ -138,7 +138,7 @@ type DatabaseContextOptions struct {
 	SessionCookieHttpOnly         bool             // Pass-through DbConfig.SessionCookieHTTPOnly
 	AllowConflicts                *bool            // False forbids creating conflicts
 	SendWWWAuthenticateHeader     *bool            // False disables setting of 'WWW-Authenticate' header
-	DisablePasswordAuthentication bool             // True enforces OIDC/guest only
+	DisablePasswordAuthentication *bool            // True enforces OIDC/guest only
 	UseViews                      bool             // Force use of views
 	DeltaSyncOptions              DeltaSyncOptions // Delta Sync Options
 	CompactInterval               uint32           // Interval in seconds between compaction is automatically ran - 0 means don't run
--- a/rest/admin_api_test.go
+++ b/rest/admin_api_test.go
@@ -4132,11 +4132,11 @@ func TestConfigResetBooleanFields(t *testing.T) {
 
 	res := rt.SendAdminRequest(http.MethodPost, "/db/_config", `{"bucket":"`+rt.Bucket().GetName()+`","num_index_replicas":0,"disable_password_auth": true}`)
 	assertStatus(t, res, http.StatusCreated)
-	assert.True(t, base.BoolDefault(rt.GetDatabase().Options.DisablePasswordAuthentication, false), "disable_password_auth was not true")
+	assert.True(t, base.BoolDefault(rt.GetDatabase().Options.DisablePasswordAuthentication, false))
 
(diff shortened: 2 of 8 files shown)


rest/admin_api_test.go
```diff
@@ -4125,6 +4125,33 @@ func TestEmptyStringJavascriptFunctions(t *testing.T) {
...
+	res := rt.SendAdminRequest(http.MethodPut, "/db/_config", string(newConfigJSON))
+	assertStatus(t, res, http.StatusCreated)
+	assert.True(t, base.BoolDefault(rt.GetDatabase().Options.DisablePasswordAuthentication, false), "disable_password_auth was not true")
+
+	newConfig["disable_password_auth"] = false
+	newConfigJSON, err = base.JSONMarshal(newConfig)
+	require.NoError(t, err, "failed to marshal new config")
+	res = rt.SendAdminRequest(http.MethodPut, "/db/_config", string(newConfigJSON))
+	assertStatus(t, res, http.StatusCreated)
+	assert.False(t, base.BoolDefault(rt.GetDatabase().Options.DisablePasswordAuthentication, false), "disable_password_auth was not false")
```
markspolakovs:
```suggestion
	assert.True(t, base.BoolDefault(rt.GetDatabase().Options.DisablePasswordAuthentication, false))

	newConfig["disable_password_auth"] = false
	newConfigJSON, err = base.JSONMarshal(newConfig)
	require.NoError(t, err, "failed to marshal new config")
	res = rt.SendAdminRequest(http.MethodPut, "/db/_config", string(newConfigJSON))
	assertStatus(t, res, http.StatusCreated)
	assert.False(t, base.BoolDefault(rt.GetDatabase().Options.DisablePasswordAuthentication, false))
```

rest/admin_api_test.go
```diff
@@ -4125,6 +4125,33 @@ func TestEmptyStringJavascriptFunctions(t *testing.T) {
...
+		"bucket":                      rt.Bucket().GetName(),
+		"num_index_replicas":          0,
+		"enable_shared_bucket_access": base.TestUseXattrs(),
+		"use_views":                   base.TestsDisableGSI(),
+		"disable_password_auth":       true,
+	}
+
+	newConfigJSON, err := base.JSONMarshal(newConfig)
+	require.NoError(t, err, "failed to marshal new config")
+	res := rt.SendAdminRequest(http.MethodPut, "/db/_config", string(newConfigJSON))
```
IsaacLambat: It would be better to test the POST endpoint to confirm it upserts the `disable_password_auth` field correctly rather than replacing the whole config with the correct value which does not properly unit test this new change.

IsaacLambat (changes requested):
You just need to figure out how to make the unit test work with a POST request and then it looks good.

rest/admin_api_test.go
```diff
@@ -4125,6 +4125,20 @@ func TestEmptyStringJavascriptFunctions(t *testing.T) {
 	assert.Equal(t, http.StatusCreated, resp.StatusCode)
 }
 
+// Regression test for CBG-2119 - ensure that bool fields are handled correctly both when set as true and as false
+func TestConfigResetBooleanFields(t *testing.T) {
```
markspolakovs:
```suggestion
// Regression test for CBG-2119 - ensure that the disable_password_auth bool field is handled correctly both when set as true and as false
func TestDisablePasswordAuthThroughAdminAPI(t *testing.T) {
```

IsaacLambat (dismissed):
LGTM

db/database.go
```diff
@@ -138,7 +138,7 @@ type DatabaseContextOptions struct {
 	SessionCookieHttpOnly         bool             // Pass-through DbConfig.SessionCookieHTTPOnly
 	AllowConflicts                *bool            // False forbids creating conflicts
 	SendWWWAuthenticateHeader     *bool            // False disables setting of 'WWW-Authenticate' header
-	DisablePasswordAuthentication bool             // True enforces OIDC/guest only
+	DisablePasswordAuthentication *bool            // True enforces OIDC/guest only
```
adamcfraser: I don't think this should be changed to *bool - I expected it would be config.DisablePasswordAuth that should change.  That would then allow us to perform the nil check handling only when the config is updated, and avoid doing it every time auth occurs.

Is there a reason that approach wouldn't work?

adamcfraser (changes requested):

db/database.go
markspolakovs: We unmarshal the user's PUT request directly into a `DbConfig` structure, so it'd have to be a `*bool` there - but if I'm reading it right, the structure used for runtime checks is not `DbConfig` but `DatabaseContextOptions`, so I think we can have the field there be a `bool`.

(Ed: only after writing that I realised that's exactly what you wrote 😄)

adamcfraser (approved):

PR #5564, opened 2022-06-09
CBG-1866 Route failover information through worker for metadata update

torcolvin:
Avoids metadata races without adding mutex to metadata since the only metadata update information happens through a 
single select statement.

Fix additional dataraces in test code found by couchbases protocol with the race detector:

- use atomics to avoid data races
- change assert to require when required for test semantics
- make sure actually are testing something with attachment compaction by resetting count to zero
- allow longer time to start DCP workers in the case of TLS protocol.
- change how doneChannel works, so it will return error (or nil), followed by channel closed. This makes sure in the case of close error the workers are still stopped.
- close DCPClient even when it fails to start in test code.

Some ideas on changing semantics for more robust/obvious behavior:

- When DCPClient.Start() fails, don't force the caller to close the client
- Don't close DCPClient after the doneChannel is received (our API should assert that it is closed). https://cs.github.com/couchbase/sync_gateway/blob/6e0cb44aa1f61264160f5b4f21332d8b0119c38d/db/attachment_compaction.go?q=dcpclient.close#L138
- Consider DCPClient.Close or separate method to be synchronous, since DCP feed will still receive events while close is happening.
- Consider not returning doneChannel as part of DCPClient.Start since it could be already done when it returns (in the case of an error or very quick oneShot) and a select statement on a doneChannel would not detect that.

### DIFF
--- a/base/dcp_client.go
+++ b/base/dcp_client.go
@@ -116,7 +116,6 @@ func NewDCPClient(ID string, callback sgbucket.FeedEventCallbackFunc, options DC
 }
 
 func (dc *DCPClient) Start() (doneChan chan error, err error) {
-
 	err = dc.initAgent(dc.spec)
 	if err != nil {
 		return nil, err
@@ -126,14 +125,17 @@ func (dc *DCPClient) Start() (doneChan chan error, err error) {
 	for i := uint16(0); i < dc.numVbuckets; i++ {
 		openErr := dc.openStream(i)
 		if openErr != nil {
+			// If this error condition is hit, the caller must call DCPClient.Close(). If we close here, should we make it blocking?
 			return nil, fmt.Errorf("Unable to start DCP client, error opening stream for vb %d: %w", i, openErr)
 		}
 	}
+	// It is possible, but unlikely for doneChannel to be closed when returning this if the DCPWorkers fail to start or finish very quickly.
 	return dc.doneChannel, nil
 }
 
 // Close is used externally to stop the DCP client. If the client was already closed due to error, returns that error
 func (dc *DCPClient) Close() error {
+	// Should this close be synchronous?
 	dc.close()
 	return dc.getCloseError()
 }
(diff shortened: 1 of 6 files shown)


base/dcp_client.go
```diff
@@ -116,24 +116,26 @@ func NewDCPClient(ID string, callback sgbucket.FeedEventCallbackFunc, options DC
...
-		return nil, err
+		return dc.doneChannel, err
 	}
 	dc.startWorkers()
 
 	for i := uint16(0); i < dc.numVbuckets; i++ {
 		openErr := dc.openStream(i)
 		if openErr != nil {
-			return nil, fmt.Errorf("Unable to start DCP client, error opening stream for vb %d: %w", i, openErr)
+			// If this error condition is hit, the caller must call DCPClient.Close(). If we close here, should we make it blocking?
```
adamcfraser: Can you add a comment for Start() to clarify that this is the contract?  (that callers must always call close on DCPClient, even in error cases).

base/dcp_client.go
```diff
@@ -116,24 +116,26 @@ func NewDCPClient(ID string, callback sgbucket.FeedEventCallbackFunc, options DC
...
+			return dc.doneChannel, fmt.Errorf("Unable to start DCP client, error opening stream for vb %d: %w", i, openErr)
 		}
 	}
+	// It is possible, but unlikely for doneChannel to be closed when returning this if the DCPWorkers fail to start or finish very quickly.
 	return dc.doneChannel, nil
 }
 
 // Close is used externally to stop the DCP client. If the client was already closed due to error, returns that error
 func (dc *DCPClient) Close() error {
+	// Should this close be synchronous?
```
adamcfraser: Possibly as a future enhancement, but don't need to flag this now.

adamcfraser (dismissed):

adamcfraser (approved):

PR #5565, opened 2022-06-10
CBG-1389: Sync Gateway will automatically restart on Windows if it fails

IsaacLambat:
CBG-1389

This will cause the Windows MSI installer to make the Sync Gateway service restart instantly if it fails (ie. a panic or fatal error occurs). This is done via the Windows Service recovery settings.

This is what the recovery settings on Windows show when the MSI has installed the service:

If SGW exits due to a user error (such as misconfiguration in the startup config), then Windows will be trying to start SGW constantly with no break causing log spam which is the same as what currently happens on via the Linux.

- [ ] Get review from the build team

### DIFF
--- a/build/windows/wix_installer/sync-gateway.wxs
+++ b/build/windows/wix_installer/sync-gateway.wxs
@@ -99,8 +99,8 @@
                    Vital="yes">
             <util:ServiceConfig
                      FirstFailureActionType="restart"
-                     SecondFailureActionType="none"
-                     ThirdFailureActionType="none" />
+                     SecondFailureActionType="restart"
+                     ThirdFailureActionType="restart" />
           </ServiceInstall>
 
         <ServiceControl Id="sc_SGWService" Start="install" Stop="both" Remove="uninstall" Name="SyncGateway" Wait="yes">
--- a/build/windows/wix_installer/sync-gateway.wxs
+++ b/build/windows/wix_installer/sync-gateway.wxs
@@ -97,6 +97,10 @@
                    Type="ownProcess"
                    Arguments='start "[INSTALLDIR]$(var.WinSGWEXE)" "[INSTALLDIR]$(var.WinServiceConfig)" "[INSTALLDIR]var\lib\couchbase\logs\sync_gateway_error.log" '
                    Vital="yes">
+            <util:ServiceConfig
+                     FirstFailureActionType="restart"
+                     SecondFailureActionType="none"
+                     ThirdFailureActionType="none" />
           </ServiceInstall>
 
         <ServiceControl Id="sc_SGWService" Start="install" Stop="both" Remove="uninstall" Name="SyncGateway" Wait="yes">


bbrks (commented):
Change looks reasonable to me, will leave approval for build.

Can you describe the level of testing that has been done for this change?
- Is it all non-clean SG exits, or any exits?
- How were non-clean SG exits tested?
- Does the Windows service wrapper that we ship return SG's exit code, or does it have its own?

build/windows/wix_installer/sync-gateway.wxs
```diff
@@ -97,6 +97,10 @@
                    Type="ownProcess"
                    Arguments='start "[INSTALLDIR]$(var.WinSGWEXE)" "[INSTALLDIR]$(var.WinServiceConfig)" "[INSTALLDIR]var\lib\couchbase\logs\sync_gateway_error.log" '
                    Vital="yes">
+            <util:ServiceConfig
+                     FirstFailureActionType="restart"
+                     SecondFailureActionType="restart"
+                     ThirdFailureActionType="restart" />
```
bbrks, line 103: With the last change to make all 3 of these `"restart"` - does that mean the service indefinitely restarts if misconfigured, or just 3 times?
Does that match the Linux services? I had a feeling they stopped after a certain number of attempts.

IsaacLambat: It means the service will indefinitely restart. The service was already indefinitely restarting when only the first failure was set to restart and the rest were set to not take action on failure because the reset fail count was set to 0.

It seems the Linux services also don't stop after a certain number of attempts however if Sync Gateway fails too quickly (EG. incorrect JSON in the startup config), Systemd will not try to restart it. 

Windows does not implement this though so it just constantly retries, even if Sync Gateway stops instantly. 
This could be combatted in 2 ways:
- Have SGW only restart the first and second time when they fail, then reset daily (set on the Windows Service settings).
- Use the service wrapper to handle SGW fails, then add logic in to that that copies Systemd's logic of not restarting the service if it fails too fast.

I am leaning towards the second option to prevent spam, and give us more control on what to do if SGW fails however that means the user has less control.

IsaacLambat:
> Change looks reasonable to me, will leave approval for build.
> 
> Can you describe the level of testing that has been done for this change?
> 
>     * Is it all non-clean SG exits, or any exits?
> 
>     * How were non-clean SG exits tested?
> 
>     * Does the Windows service wrapper that we ship return SG's exit code, or does it have its own?

The testing I did on this was to create a MSI package to install a package then make Sync Gateway intentionally fail multiple times. This was done using a custom SG version that paniced when a certain endpoint was called, and by using invalid JSON in the startup config.

All exits of SG are not clean, and therefore it would restart in all cases SG exits. The only way to stop the service is going to the Windows services, and manually stopping the `Couchbase Sync Gateway` service or by uninstalling SG.

The service wrapper logs the SG exit code then returns it's own (in the form of a Go panic).

build/windows/wix_installer/sync-gateway.wxs
bbrks: >It seems the Linux services also don't stop after a certain number of attempts however if Sync Gateway fails too quickly (EG. incorrect JSON in the startup config), Systemd will not try to restart it.

Ahh ok, yeah that burst behaviour is definitely what I was thinking of!

In that case, I think indefinite but configurable retries on Windows are an acceptable compromise. It should be immediately obvious that SG has not started correctly when somebody has misconfigured something.

I'd rather not add any more stuff in the windows service wrapper if it's non-standard behaviour on that platform.

mhocouchbase (approved):
Per conversation with Isaac, this follows the same practice that linux does.  End user is expected to either solve the problem or stop the service from control panel if the service restarts indefinitely.

PR #5568, opened 2022-06-15
CBG-2059: More accurately tag redactable fields in HTTP logs

markspolakovs:
CBG-2059

Previously when processing a HTTP log line in `tagPathVars`, we would redact the first instance of all the path variables that need redaction (e.g. `{docid}`). However, given the database `testificate` and document `test`, we would expect to log `/test/<ud>testificate</ud>`, but we'd actually log `/<ud>test</ud>ificate/test` - not only missing out the document name, but also redacting part of the DB name.

This patch changes the tagging to operate from the end of the string, as in nearly all circumstances the redactable path segments follow the non-redactable ones (a list is in the Jira ticket). As far as I'm aware the only exception is `/db/_user/{name}/_session/{sessionid}` (`name` is redactable but `sessionid` is not) - I'd expect the likelihood of a username being an exact substring of a session ID to be fairly limited. 

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/323/

### DIFF
--- a/base/logging.go
+++ b/base/logging.go
@@ -9,6 +9,7 @@
 package base
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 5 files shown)


bbrks (approved):
I think this is a reasonable approach. Thanks!

PR #5570, opened 2022-06-15
CBG-2100 Allow rest tests to use couchbases://

torcolvin:
CBG-2100

Skip test so Jenkins can run with `couchbases://` protocol without errors.


rest/admin_api_auth_test.go
```diff
@@ -345,7 +345,10 @@ func TestAdminAuth(t *testing.T) {
 	if base.UnitTestUrlIsWalrus() {
 		t.Skip("Test requires Couchbase Server")
 	}
-
+	serverURL := base.UnitTestUrl()
+	if base.ServerIsTLS(serverURL) {
+		t.Skipf("URI %s needs to start with couchbase://", serverURL)
+	}
 	rt := NewRestTester(t, nil)
```
bbrks: This test can be made to work with either scheme. Preferable to skipping.

Potentially similar fix for below tests, although have not looked to see why those are failing.

```suggestion
	rt := NewRestTester(t, &RestTesterConfig{useTLSServer: base.ServerIsTLS(base.UnitTestUrl())})
```

bbrks (changes requested):

rest/admin_api_auth_test.go
torcolvin: there were so many circumstances where RestTesterConfig is initialized, so I moved the logic to set TLS in NewRestTester. All integration tests seem to pass.

bbrks (dismissed):

bbrks (approved):

PR #5574, opened 2022-06-16
CBG-2081: OpenAPI specs/REST API stubs for collections

bbrks:
CBG-2081

- update api specs for keyspace/collections changes
- db-> keyspace changes in REST API code
- stub per-collection sequences in in `GET /db/` response

- [ ] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/341/

### DIFF
--- a/docs/api/admin.yaml
+++ b/docs/api/admin.yaml
@@ -92,8 +92,8 @@ paths:
     $ref: './paths/admin/{db}~_config~sync.yaml'
   '/{db}/_config/import_filter':
     $ref: './paths/admin/{db}~_config~import_filter.yaml'
-  '/{db}/_resync':
-    $ref: './paths/admin/{db}~_resync.yaml'
+  '/{keyspace}/_resync':
+    $ref: './paths/admin/{keyspace}~_resync.yaml'
   '/{keyspace}/_purge':
     $ref: './paths/admin/{keyspace}~_purge.yaml'
   '/{db}/_flush':
--- a/docs/api/admin.yaml
+++ b/docs/api/admin.yaml
@@ -24,10 +24,10 @@ paths:
     $ref: './paths/admin/{db}~_session.yaml'
   '/{db}/_session/{sessionid}':
     $ref: './paths/admin/{db}~_session~{sessionid}.yaml'
-  '/{db}/_raw/{docid}':
-    $ref: './paths/admin/{db}~_raw~{docid}.yaml'
-  '/{db}/_revtree/{docid}':
-    $ref: './paths/admin/{db}~_revtree~{docid}.yaml'
+  '/{keyspace}/_raw/{docid}':
+    $ref: './paths/admin/{keyspace}~_raw~{docid}.yaml'
+  '/{keyspace}/_revtree/{docid}':
+    $ref: './paths/admin/{keyspace}~_revtree~{docid}.yaml'
   '/{db}/_user/':
     $ref: './paths/admin/{db}~_user~.yaml'
   '/{db}/_user/{name}':
@@ -94,8 +94,8 @@ paths:
     $ref: './paths/admin/{db}~_config~import_filter.yaml'
   '/{db}/_resync':
     $ref: './paths/admin/{db}~_resync.yaml'
-  '/{db}/_purge':
-    $ref: './paths/admin/{db}~_purge.yaml'
+  '/{keyspace}/_purge':
+    $ref: './paths/admin/{keyspace}~_purge.yaml'
   '/{db}/_flush':
     $ref: './paths/admin/{db}~_flush.yaml'
   '/{db}/_online':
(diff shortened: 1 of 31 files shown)


IsaacLambat (dismissed):
LGTM

docs/api/admin.yaml
```diff
@@ -94,8 +94,8 @@ paths:
     $ref: './paths/admin/{db}~_config~import_filter.yaml'
   '/{db}/_resync':
```
adamcfraser: I think we're defining sync functions per collection.  If so, running resync per collection seems like something we'll want to do?

docs/api/admin.yaml
```diff
@@ -112,20 +112,22 @@ paths:
...
     $ref: './paths/admin/{db}~.yaml'
+  '/{keyspace}/':
+    $ref: './paths/admin/{keyspace}~.yaml'
   /_expvar:
     $ref: ./paths/admin/_expvar.yaml
   /:
     $ref: ./paths/admin/~.yaml
   '/{db}/_all_docs':
     $ref: './paths/admin/{db}~_all_docs.yaml'
-  '/{db}/_bulk_docs':
```
adamcfraser, line 125: Is the plan to have 1.x replication only supported against a single collection?

rest/api.go
```diff
@@ -358,15 +358,24 @@ func (h *handler) instanceStartTimeMicro() int64 {
...
-	SequenceNumber                uint64 `json:"update_seq"`
-	CommittedUpdateSequenceNumber uint64 `json:"committed_update_seq"` // Used by perf tests, shouldn't be removed
-	InstanceStartTimeMicro        int64  `json:"instance_start_time"`  // microseconds since epoch
-	CompactRunning                bool   `json:"compact_running"`
-	PurgeSequenceNumber           uint64 `json:"purge_seq"`
-	DiskFormatVersion             uint64 `json:"disk_format_version"`
-	State                         string `json:"state"`
-	ServerUUID                    string `json:"server_uuid,omitempty"`
+	DBName                        string                       `json:"db_name"`
+	SequenceNumber                uint64                       `json:"update_seq"`           // The last sequence written to the _default collection
```
adamcfraser: Does this make sense to keep, given that the _default collection may not be included in the database?  At minimum it could be switched to omitempty.

rest/api.go
```diff
@@ -392,6 +401,15 @@ func (h *handler) handleGetDB() error {
 		DiskFormatVersion:             0, // Probably meaningless, but add for compatibility
 		State:                         runState,
 		ServerUUID:                    h.db.DatabaseContext.GetServerUUID(),
+		// TODO: Get scope/collection sequences
```
adamcfraser: Should be removed until we have real data to populate with - potentially could hardcode _default._default for now if needed for testing.

rest/handler.go
```diff
@@ -174,6 +174,44 @@ func (h *handler) ctx() context.Context {
...
+	case 3:
+		dbField = &parts[0]
+		scopeField = &parts[1]
+		collectionField = &parts[2]
+	default:
+		return "", "", "", fmt.Errorf("unknown keyspace format: %q - expected 1-3 fields", ks)
+	}
+
+	// make sure all declared fields have stuff in them
+	if dbField != nil && *dbField == "" ||
```
adamcfraser: I don't think there's a case where dbField is nil in the current implementation

rest/handler.go
```diff
@@ -174,6 +174,44 @@ func (h *handler) ctx() context.Context {
...
+	)
+
+	// FIXME: This will be quicker/zero alloc if we use base.safeCut...
+	parts := strings.Split(ks, base.ScopeCollectionSeparator)
+	switch len(parts) {
+	case 1:
+		dbField = &parts[0]
+	case 2:
+		dbField = &parts[0]
+		collectionField = &parts[1]
```
adamcfraser: Stuck a comment in the design doc on this.  Using the default scope when it's omitted feels a bit off as long as the database is only associated with a single scope (that might be the default scope).  I'd expect [db.myCollection] to be targeting the scope associated with db.  I don't think that changes future support for multiple scopes per DB - in that case we would require the fully qualified keyspace.

rest/handler.go
```diff
@@ -174,6 +174,44 @@ func (h *handler) ctx() context.Context {
...
+// parseKeyspace will return a fully qualified db, scope and collection for a given '.' separated keyspace string.
+// Returns "_default" for scope and collection if the given keyspace is not otherwise fully qualified.
+func parseKeyspace(ks string) (db, scope, collection string, err error) {
+	var (
+		dbField,
+		scopeField,
+		collectionField *string
+	)
+
+	// FIXME: This will be quicker/zero alloc if we use base.safeCut...
```
adamcfraser: Any reason not to fix this in this PR?

docs/api/admin.yaml
```diff
@@ -112,20 +112,22 @@ paths:
...
-    $ref: './paths/admin/{db}~_bulk_docs.yaml'
-  '/{db}/_bulk_get':
-    $ref: './paths/admin/{db}~_bulk_get.yaml'
+  '/{keyspace}/_bulk_docs':
+    $ref: './paths/admin/{keyspace}~_bulk_docs.yaml'
+  '/{keyspace}/_bulk_get':
+    $ref: './paths/admin/{keyspace}~_bulk_get.yaml'
   '/{db}/_changes':
```
adamcfraser, line 131: It feels inconsistent from an API perspective to have _changes stay on db, but other multi-doc operations move to keyspace.  I understand that _changes will allow collection selection via parameters, but still seems a bit trickier to reason about as an API.

adamcfraser (changes requested):
General approach looks good, some questions and suggestions.

adamcfraser (approved):

PR #5581, opened 2022-06-21
Run API specs through openapi-format

bbrks:
Noisy diff, but just ran all the api docs through `openapi-format` (https://www.npmjs.com/package/openapi-format)

```sh
$ find docs/api -name \*.yaml -exec openapi-format {} -o {} \;
...
$ openapi lint
validating /docs/api/admin.yaml...
/docs/api/admin.yaml: validated in 257ms

validating /docs/api/public.yaml...
/docs/api/public.yaml: validated in 78ms

validating /docs/api/metric.yaml...
/docs/api/metric.yaml: validated in 11ms

Woohoo! Your OpenAPI definitions are valid. 🎉
```

### DIFF
--- a/docs/api/admin.yaml
+++ b/docs/api/admin.yaml
@@ -20,154 +20,151 @@ servers:
         description: The hostname to use
         default: localhost
 paths:
-  # Admin only paths
-  /{db}/_session:
+  '/{db}/_session':
     $ref: './paths/admin/{db}~_session.yaml'
-  /{db}/_session/{sessionid}:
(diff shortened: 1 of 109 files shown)


IsaacLambat (approved):
Great improvements and even found some tags in the wrong place as well. Happy for it to be merged.

NITPICK: Would be good if the validation specification could make it mandatory to use quotes around descriptions and summary's (to avoid using quotes in some places and no quotes in others like what currently happens due to commas, etc being used).

PR #5586, opened 2022-06-22
CBG-2140: Config parsing and validation for Local JWT Auth

markspolakovs:
CBG-2140

Sets up the configuration fields for local JWT authentication.

This is a slightly noisy diff - the majority of the changes are from moving some of the OIDC provider config's fields into a shared JWTConfigCommon structure and updating all the manual instantiations of it. The rest of the noise stems from changing `ClientID` from a `string` to a `*string` (per local JWT spec) - this won't functionally affect OIDC, because the OIDC validation (in `NewDatabaseContext`) verifies that it's non-empty. The most interesting files are likely `auth/jwt.go` and `rest/config+config_test.go`.

This doesn't implement all the validation logic, as one more check (duplicate OIDC/JWT issuers) will be broken out into a follow-up PR that refactors some of the OIDC validation logic.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/332/

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -708,7 +708,7 @@ func (auth *Authenticator) AuthenticateTrustedJWT(token string, provider *OIDCPr
 	var identity *Identity
 	if provider.AllowUnsignedProviderTokens {
 		// Verify claims - ensures that the token we received from the provider is valid for Sync Gateway
-		identity, err = VerifyClaims(token, provider.ClientID, provider.Issuer)
+		identity, err = VerifyClaims(token, base.StringDefault(provider.ClientID, ""), provider.Issuer)
 		if err != nil {
 			base.DebugfCtx(auth.LogCtx, base.KeyAuth, "Error verifying raw token in AuthenticateTrustedJWT: %v", err)
 			return nil, PrincipalConfig{}, time.Time{}, err
--- a/auth/jwt.go
+++ b/auth/jwt.go
@@ -1,0 +1,53 @@
+package auth
+
+import (
+	"gopkg.in/square/go-jose.v2"
+)
+
+// SupportedAlgorithms is list of signing algorithms explicitly supported
+// by github.com/coreos/go-oidc package. If a provider supports other algorithms,
+// such as HS256 or none, those values won't be passed to the IDTokenVerifier.
+var SupportedAlgorithms = map[jose.SignatureAlgorithm]bool{
+	jose.RS256: true,
+	jose.RS384: true,
+	jose.RS512: true,
+	jose.ES256: true,
+	jose.ES384: true,
+	jose.ES512: true,
+	jose.PS256: true,
+	jose.PS384: true,
+	jose.PS512: true,
+}
+
+// JWTConfigCommon groups together configuration options common to both OIDC and local JWT authentication.
+type JWTConfigCommon struct {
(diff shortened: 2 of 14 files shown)


auth/jwt.go
```diff
@@ -0,0 +1,53 @@
...
+	jose.PS256: true,
+	jose.PS384: true,
+	jose.PS512: true,
+}
+
+// JWTConfigCommon groups together configuration options common to both OIDC and local JWT authentication.
+type JWTConfigCommon struct {
+	Issuer         string  `json:"issuer"`                    // OIDC Issuer
+	Register       bool    `json:"register"`                  // If true, server will register new user accounts
+	ClientID       *string `json:"client_id,omitempty"`       // Client ID
```
markspolakovs, line 26: This may end up being renamed to `audience` for local-JWT (still discussing with PM)

base/util.go
```diff
@@ -845,6 +845,13 @@ func BoolDefault(b *bool, ifNil bool) bool {
 	return ifNil
 }
 
+func StringDefault(s *string, ifNil string) string {
```
bbrks, line 848: I added this exact function a couple of commits ago 😄 might need to rebase on latest master to avoid issues when merging.

auth/jwt.go
```diff
@@ -0,0 +1,53 @@
...
+	JWTAlgorithm string
+	JWTAlgList   []JWTAlgorithm
+)
+
+type LocalJWTAuthProvider struct {
+	JWTConfigCommon
+
+	Algorithms JWTAlgList        `json:"algorithms"`
+	Keys       []jose.JSONWebKey `json:"keys"`
+}
```
markspolakovs, line 51: Needs OpenAPI documentation

auth/auth.go
```diff
@@ -708,7 +708,7 @@ func (auth *Authenticator) AuthenticateTrustedJWT(token string, provider *OIDCPr
 	var identity *Identity
 	if provider.AllowUnsignedProviderTokens {
 		// Verify claims - ensures that the token we received from the provider is valid for Sync Gateway
-		identity, err = VerifyClaims(token, provider.ClientID, provider.Issuer)
+		identity, err = VerifyClaims(token, base.StringDefault(provider.ClientID, ""), provider.Issuer)
```
bbrks, line 711: Given that it looks like everywhere that uses `provider.ClientID` uses the `StringDefault` util to fall back to empty string - what's the reasoning behind the change to a `*string` ?

base/util.go
markspolakovs: Was wondering where it came from! :D

auth/auth.go
markspolakovs: For local JWT auth, the `client_id` field is required, but can be set to an empty string to disable validation ([spec](https://docs.google.com/document/d/1jxz8fnDfPQ-BXxBGOGEgf2NoTRh00A4v30wPt1NA8hM/edit#heading=h.gyjd8mnd5uke)) - the change is to distinguish the absent case from the empty case.

For OIDC there's no functional change, hence all the defaults, but it'll become meaningful when local JWT authentication is implemented (#5593).

IsaacLambat (dismissed):
LGTM

IsaacLambat (approved):

PR #5588, opened 2022-06-22
Up document count for test

torcolvin:
This test relies on having enough documents so that terminate will be called before all documents are marked. In jekins this test has failed, so increase document count.


torcolvin:
https://jenkins.sgwdev.com/job/SyncGateway-Integration/327/testReport/junit/github/com_couchbase_sync_gateway_db/TestAttachmentCompactIncorrectStat/ for example failure.

adamcfraser (approved):

PR #5590, opened 2022-06-22
CBG-2052: Implement Blip Unsub changes for Connected Client protocol

IsaacLambat:
CBG-2052

-  Replaced changes feed terminator option with cancellable context
- Unsub changes cancels the context if it exists, otherwise returns an error
- Unsub changes uses a lock to make sure that the context is not created by subChanges while unsubChanges is checking if the context exists
- Added a lock for when closing the blip handler
- Added testing for this

- Changed changes options `Ctx` to be called `LoggingCtx`

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/391

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -178,11 +178,8 @@ func (bh *blipHandler) handleSetCheckpoint(rq *blip.Message) error {
 
 // Received a "subChanges" subscription request
 func (bh *blipHandler) handleSubChanges(rq *blip.Message) error {
-
-	bh.lock.Lock()
-	defer bh.lock.Unlock()
-
-	bh.gotSubChanges = true
+	bh.changesLock.Lock()
+	defer bh.changesLock.Unlock()
 
 	defaultSince := bh.db.CreateZeroSinceValue()
 	latestSeq := func() (SequenceID, error) {
@@ -231,12 +228,11 @@ func (bh *blipHandler) handleSubChanges(rq *blip.Message) error {
 	}
 
 	continuous := subChangesParams.continuous()
-	// used for stats tracking
-	bh.continuous = continuous
+
 	// Start asynchronous changes goroutine
 	go func() {
 		// Pull replication stats by type
-		if bh.continuous {
+		if continuous {
 			bh.replicationStats.SubChangesContinuousActive.Add(1)
 			defer bh.replicationStats.SubChangesContinuousActive.Add(-1)
 			bh.replicationStats.SubChangesContinuousTotal.Add(1)
@@ -247,6 +243,7 @@ func (bh *blipHandler) handleSubChanges(rq *blip.Message) error {
 		}
 
 		defer func() {
+			bh.changesCtxCancel()
 			bh.activeSubChanges.Set(false)
 		}()
 		// sendChanges runs until blip context closes, or fails due to error
@@ -269,8 +266,8 @@ func (bh *blipHandler) handleSubChanges(rq *blip.Message) error {
 }
 
 func (bh *blipHandler) handleUnsubChanges(rq *blip.Message) error {
-	bh.lock.Lock()
-	defer bh.lock.Unlock()
+	bh.changesLock.Lock()
+	defer bh.changesLock.Unlock()
 
 	if !bh.activeSubChanges.IsTrue() {
 		return base.HTTPErrorf(http.StatusBadRequest, "No subChanges subscription active to unsubscribe from")
@@ -1289,8 +1286,8 @@ func (bsc *BlipSyncContext) addAllowedAttachments(docID string, attMeta []Attach
 		return
 	}
 
-	bsc.lock.Lock()
-	defer bsc.lock.Unlock()
+	bsc.allowedAttachmentsLock.Lock()
+	defer bsc.allowedAttachmentsLock.Unlock()
 
 	if bsc.allowedAttachments == nil {
 		bsc.allowedAttachments = make(map[string]AllowedAttachment, 100)
@@ -1320,8 +1317,8 @@ func (bsc *BlipSyncContext) removeAllowedAttachments(docID string, attMeta []Att
 		return
 	}
 
-	bsc.lock.Lock()
-	defer bsc.lock.Unlock()
+	bsc.allowedAttachmentsLock.Lock()
+	defer bsc.allowedAttachmentsLock.Unlock()
 
 	for _, attachment := range attMeta {
 		key := allowedAttachmentKey(docID, attachment.digest, activeSubprotocol)
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -184,10 +184,6 @@ func (bh *blipHandler) handleSubChanges(rq *blip.Message) error {
 
 	bh.gotSubChanges = true
 
-	// Context to cancel subchanges from being sent
-	changesCtx := context.Background()
-	bh.changesCtx, bh.changesCtxCancel = context.WithCancel(changesCtx)
-
 	defaultSince := bh.db.CreateZeroSinceValue()
 	latestSeq := func() (SequenceID, error) {
 		seq, err := bh.db.LastSequence()
@@ -201,6 +197,11 @@ func (bh *blipHandler) handleSubChanges(rq *blip.Message) error {
 	// Ensure that only _one_ subChanges subscription can be open on this blip connection at any given time.  SG #3222.
 	if !bh.activeSubChanges.CASRetry(false, true) {
 		return fmt.Errorf("blipHandler already has an outstanding continous subChanges.  Cannot open another one")
+	}
+
+	// Create ctx if it does not exist or has been closed
+	if bh.changesCtx == nil || bh.changesCtx.Err() != nil {
+		bh.changesCtx, bh.changesCtxCancel = context.WithCancel(context.Background())
 	}
 
 	if len(subChangesParams.docIDs()) > 0 && subChangesParams.continuous() {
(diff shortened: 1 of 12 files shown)


db/blip_handler.go
```diff
@@ -263,8 +267,8 @@ func (bh *blipHandler) handleSubChanges(rq *blip.Message) error {
 }
 
 func (bh *blipHandler) handleUnsubChanges(rq *blip.Message) error {
-	return base.HTTPErrorf(http.StatusNotImplemented, "unsubChanges not implemented yet")
-	// TODO: Implement unsubChanges
+	bh.changesCtxCancel()
```
bbrks, line 272: What happens if a client calls `unsubChanges` before a `subChanges` has been issued? Does it panic? Can we handle that better if so?

rest/blip_api_test.go
```diff
@@ -4492,3 +4492,43 @@ func TestBlipLegacyAttachDocUpdate(t *testing.T) {
...
+	unsubChangesRequest := blip.NewRequest()
+	unsubChangesRequest.SetProfile(db.MessageUnsubChanges)
+	err = btc.pullReplication.sendMsg(unsubChangesRequest)
+	require.NoError(t, err)
+	unsubChangesResponse := unsubChangesRequest.Response()
+	_, err = unsubChangesResponse.Body()
+	require.NoError(t, err)
+
+	// Confirm no more changes being sent
+	resp = rt.updateDoc("doc2", "", `{"key":"val1"}`)
```
bbrks: This test might be racy given it's not a synchronous operation to call `unsubChanges`. If the changes feed takes a while to actually exit, doc2 might end up being sent.

I'm not exactly sure of the fix here - the replication protocol wording implies unsub is async, but it's possible we can block inside the test on stats instead to ensure timing.

bbrks (changes requested):
A couple of comments related to edge cases/potential race condition in the test.

db/changes.go
```diff
@@ -1305,7 +1306,12 @@ func generateBlipSyncChanges(database *Database, inChannels base.Set, options Ch
 
 	// Store one-shot here to protect
 	isOneShot := !options.Continuous
-	err, forceClose = GenerateChanges(context.Background(), database, inChannels, options, docIDFilter, send)
+	// Use the context if provided in the options
+	ctx := options.ChangesCtx
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	err, forceClose = GenerateChanges(ctx, database, inChannels, options, docIDFilter, send)
```
adamcfraser: I don't think closing the context being passed to GenerateChanges is sufficient to terminate all goroutines associated with a changes feed.  Closing the changes terminator channel is the only thing that does that today - you can search for usages of that terminator to see where it's used - only one of those also does a check for the context being cancelled.  We could refactor changes to use a context throughout instead of the terminator channel, but otherwise I think we need to continue using the terminator to stop changes.

adamcfraser (changes requested):

rest/blip_api_test.go
```diff
@@ -4494,3 +4494,49 @@ func TestBlipLegacyAttachDocUpdate(t *testing.T) {
...
+	require.EqualValues(t, 1, activeReplStat.Value())
+
+	// Unsub changes
+	unsubChangesRequest := blip.NewRequest()
+	unsubChangesRequest.SetProfile(db.MessageUnsubChanges)
+	err = btc.pullReplication.sendMsg(unsubChangesRequest)
+	require.NoError(t, err)
+	unsubChangesResponse := unsubChangesRequest.Response()
+	_, err = unsubChangesResponse.Body()
+	require.NoError(t, err)
```
bbrks: Doing this doesn't really match how the `BlipTesterClient` works - which is a higher level API that hides a lot of the BLIP/replication messaging.

You could just move this code into a method on `BlipTesterClient`, so you can call `require.NoError(t, btc.UnsubChanges())`, in the same way you call `StartPull()` to start them.

rest/blip_api_test.go
```diff
@@ -4494,3 +4494,49 @@ func TestBlipLegacyAttachDocUpdate(t *testing.T) {
...
+func TestUnsubChanges(t *testing.T) {
+	base.SetUpTestLogging(t, base.LevelInfo, base.KeyAll)
+	rt := NewRestTester(t, &RestTesterConfig{guestEnabled: true})
+	defer rt.Close()
+
+	btc, err := NewBlipTesterClientOptsWithRT(t, rt, nil)
+	require.NoError(t, err)
+	defer btc.Close()
+
+	// Sub changes
```
bbrks, line 4627: I think I mentioned in previous comment about making sure it works - but would be nice to bake this into a test assertion:

Can we run an `UnsubChanges` before the first `SubChanges` to make sure this doesn't panic/cause other issues?

db/blip_handler.go
```diff
@@ -263,8 +268,14 @@ func (bh *blipHandler) handleSubChanges(rq *blip.Message) error {
 }
 
 func (bh *blipHandler) handleUnsubChanges(rq *blip.Message) error {
-	return base.HTTPErrorf(http.StatusNotImplemented, "unsubChanges not implemented yet")
-	// TODO: Implement unsubChanges
+	bh.lock.Lock()
+	defer bh.lock.Unlock()
```
bbrks: Note to self: Take a closer look at what this lock is guarding

db/blip_handler.go
```diff
@@ -198,6 +198,11 @@ func (bh *blipHandler) handleSubChanges(rq *blip.Message) error {
 		return fmt.Errorf("blipHandler already has an outstanding continous subChanges.  Cannot open another one")
 	}
 
+	// Create ctx if it does not exist or has been closed
+	if bh.changesCtx == nil || bh.changesCtx.Err() != nil {
```
bbrks: I think it might actually be safer for us _not_ to handle `nil` contexts - it's not recommended to pass around nil contexts, and we want to find the code paths that are doing so.

As far as I can tell, this should never be nil, which would save a fair amount of awkward nil checking for `changesCtx`/`changesCtxCancel`.

db/changes.go
```diff
@@ -524,8 +524,11 @@ func (db *Database) MultiChangesFeed(chans base.Set, options ChangesOptions) (<-
 		return nil, nil
 	}
 
-	if (options.Continuous || options.Wait) && options.Terminator == nil {
-		base.WarnfCtx(db.Ctx, "MultiChangesFeed: Terminator missing for Continuous/Wait mode")
+	if options.ChangesCtx == nil {
+		if options.Continuous || options.Wait {
+			base.WarnfCtx(db.Ctx, "MultiChangesFeed: Changes Context missing for Continuous/Wait mode")
+		}
+		options.ChangesCtx = context.Background()
```
bbrks: I'm wondering if this is a situation where we _want_ to find out about this (at dev time) by returning an error/panicing, rather than warning and carrying on with no way to kill the changes goroutines that are started up.

Looks like this is ancient code [if you step back far enough](https://github.com/couchbase/sync_gateway/commit/6d381ffea63ffdbf2efbfe0337ac68de95255334) and I've neither seen this warning anywhere before, nor do I think it's a risk for us to start to be stricter about this.

bbrks (changes requested):
A few comments but I do still need to take a closer look at the locking on `blipHandler` to figure out what its intended use is.

rest/blip_api_test.go
IsaacLambat: I did try this, and it did panic initially but it is fixed now. The problem with having it on a test assertion is that Blip recovers from the panic and just logs it instead, so not sure how I could assert on this unless I assert on the logs containing the panic text.

Will try to come up with something.

db/changes.go
IsaacLambat: The main point of adding this context is because lots of tests call the changes feed directly, bypassing the context initialisation. I will remove this changes context check, and change lots of tests to have the context added in the changes options.

rest/blip_api_test.go
bbrks: BLIP should return an error code and a message in the body if the handler paniced

db/changes.go
bbrks: Understood. I think it makes sense to force the tests to pass in a context, otherwise we'd potentially never stop changes feeds as we have a non-cancellable background context in use.

db/blip_handler.go
bbrks: As discussed - rename this existing lock to indicate it's guarding `allowedAttachments` - and create a new `changesCtxLock` for your usage.

db/changes.go
```diff
@@ -400,7 +400,7 @@ func (db *Database) changesFeed(singleChannelCache SingleChannelCache, options C
 				paginationOptions.Limit = base.MinInt(remainingLimit, queryLimit)
 			}
 
-			// TODO: pass db.Ctx down to changeCache?
+			// TODO: pass db.LoggingCtx down to changeCache?
```
bbrks: unintentional change

```suggestion
			// TODO: pass db.Ctx down to changeCache?
```

bbrks (changes requested):

db/blip_handler.go
```diff
@@ -264,8 +266,14 @@ func (bh *blipHandler) handleSubChanges(rq *blip.Message) error {
...
 
 func (bh *blipHandler) handleUnsubChanges(rq *blip.Message) error {
-	return base.HTTPErrorf(http.StatusNotImplemented, "unsubChanges not implemented yet")
-	// TODO: Implement unsubChanges
+	bh.changesCtxLock.Lock()
+	defer bh.changesCtxLock.Unlock()
+
+	if !bh.activeSubChanges.IsTrue() {
+		return base.HTTPErrorf(http.StatusBadRequest, "No subChanges subscription active to unsubscribe from")
+	}
```
bbrks: I think we can probably get away without doing this check here. We may still want to try and cancel the context if we have one, even if activeSubChanges is reporting false for whatever reason.

```suggestion
```

bbrks (changes requested):
One final small change on UnsubChanges then I think this is good to go

bbrks (approved):
LGTM - pending tests

adamcfraser (approved):

PR #5587, opened 2022-06-22
CBG-2140: Merge OIDC and local-JWT config validation

markspolakovs:
CBG-2140

Move all the OIDC validation logic to `DbConfig.validate`, to colocate it with local JWT validation, and implement the remaining validations from the spec.

## [Integration Tests](https://jenkins.sgwdev.com/job/SyncGateway-Integration/build?delay=0sec)
- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/340/

PR #5593, opened 2022-06-24
CBG-2141: Implement local JWT authentication

markspolakovs:
CBG-2141

Implements local JWT authentication. Change the signature of `AuthenticateUntrustedJWT` to check both OIDC and local-JWT providers, and delegate the token validation to either (through a `jwtAuthenticator` interface), then continue into `authenticateJWTIdentity` (fka `authenticateOIDCIdentity`) to find the user etc. (`AuthenticateTrustedJWT` remains untouched because there's no notion of a "trusted JWT" for local-JWT like there is for OIDC authorization code flow.)

There are two primary tests of the new functionality - `TestJWTVerifyToken` tests the raw JWT validation logic, while `TestLocalJWTAuthenticationE2E` runs a matrix of tests of the `/_session` flow. `TestLocalJWTAuthenticationEdgeCases` runs E2E tests of some extra cases that the previous doesn't cover, while `TestLocalJWTRolesChannels` checks that the role/channel mapping functionality also works.

This might be easier to review commit-by-commit, in particular with the two large rename commits towards the end.

Open questions:
- [x] I've tried to minimise code duplication, but there's still a fair amount of room for refactoring - perhaps someday we can reimplement OIDC on top of the local-JWT framework rather than side by side (though likely outside the scope of this ticket)
  - Filed CBG-2170
- [x] Many things still refer to "OIDC" when really they're applicable to both - the most egregious likely being `OIDCRoles/OIDCChannels/OIDCIssuer/OIDCLastUpdated` (which are exposed to the user through the admin API) - there may be merit in renaming these
  - Done.

## Dependencies

- [x] #5586 
- [x] #5587 

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/359/


auth/auth.go
```diff
@@ -623,78 +623,66 @@ func (auth *Authenticator) AuthenticateUser(username string, password string) (U
...
-	if provider == nil {
-		return nil, PrincipalConfig{}, base.RedactErrorf("No provider found for issuer %v", base.UD(issuer))
+	issuer, audiences, err := getIssuerWithAudience(token)
+	if err != nil {
+		base.DebugfCtx(auth.LogCtx, base.KeyAuth, "Error extracting issuer/audiences in AuthenticateUntrustedJWT: %v", err)
+		return nil, PrincipalConfig{}, err
+	}
+
+	var (
+		authenticatorName string
```
adamcfraser, line 640: Would it be cleaner to add GetName() to the jwtAuthenticator interface?

auth/jwt.go
```diff
@@ -38,6 +46,25 @@ type JWTConfigCommon struct {
...
 }
 
+// ValidFor returns whether the issuer matches, and one of the audiences matches
+func (j JWTConfigCommon) ValidFor(issuer string, audiences audience) bool {
+	if j.Issuer != issuer {
+		return false
+	}
+	// Validation enforces that ClientID is non-nil for OIDC.
+	if j.ClientID == nil || *j.ClientID == "" {
+		return true
```
adamcfraser, line 59: Should this be false?  If validation is supposed to enforce that clientID shouldn't be empty, we're in an unexpected situation if we hit this, and wouldn't expect that we should return valid=true.

auth/auth.go
```diff
@@ -623,78 +623,66 @@ func (auth *Authenticator) AuthenticateUser(username string, password string) (U
...
+		authenticator     jwtAuthenticator
+	)
+	// can't do `authenticator, ok = getProviderWhenSingle()` because it'll return a non-nil pointer to a nil OIDCProvider
```
adamcfraser: Can remove this comment, I think, if it's just highlighting golang's "an interface holding a nil value isn't nil".

auth/auth.go
```diff
@@ -623,78 +623,66 @@ func (auth *Authenticator) AuthenticateUser(username string, password string) (U
...
+	if single, ok := oidcProviders.getProviderWhenSingle(); ok {
```
adamcfraser: If we've got a single OIDC provider defined, but it's not a match for the issuer/audience, this looks like it's not going to successfully fall through to checking the token against localJWT config.

auth/jwt.go
```diff
@@ -50,4 +77,123 @@ type LocalJWTAuthProvider struct {
 	Keys       []jose.JSONWebKey `json:"keys"`
 }
 
+func (l *LocalJWTAuthProvider) verifyToken(ctx context.Context, token string, _ OIDCCallbackURLFunc) (*Identity, error) {
```
adamcfraser: This looks like it clones some functionality from coreos/go-oidc/verify.go.  What functionality is different for local JWT handling that would prevent us from creating an IDTokenVerifier and just using that code directly?
I'm concerned both about the potential for bugs introduced when cloning their logic, but also additional maintenance work to uptake fixes made in the library.

auth/jwt.go
```diff
@@ -50,4 +77,123 @@ type LocalJWTAuthProvider struct {
...
+	// If the prefix contains forward slash or underscore, it's not valid as-is for a username: forward slash
+	// breaks the REST API, underscore breaks uniqueness of "[prefix]_[sub]".  URL encode the prefix to cover
+	// this scenario
+	l.UserPrefix = url.QueryEscape(l.UserPrefix)
+}
+
 type LocalJWTConfig map[string]*LocalJWTAuthProvider
+
+type jwtAuthenticator interface {
+	verifyToken(ctx context.Context, token string, callbackURLFunc OIDCCallbackURLFunc) (*Identity, error)
```
adamcfraser, line 192: Having something OIDC-specific on the common interface suggests that the interface might not be at exactly the right level.  In this case I think it's to support lazy-loading of the OIDC provider?  I wonder if that could be handled prior to the verifyToken invocation for the OIDC flows.

adamcfraser (changes requested):
Overall looks good - some questions on the details.

auth/jwt.go
markspolakovs: That's only the case for OIDC (I can see how the comment is misleading though, will update) - for local-JWT, an empty (but non-nil) `ClientID` disables `aud` verification ([spec](https://docs.google.com/document/d/1jxz8fnDfPQ-BXxBGOGEgf2NoTRh00A4v30wPt1NA8hM/edit#heading=h.gyjd8mnd5uke)).

auth/jwt.go
markspolakovs: It is a bit of a smell, I agree. The problem is that `buildCallbackURLFunc` is required for `OIDCProvider.GetClient`(which then uses it in `initOIDCClient` and creates `OIDCClient` based on it) and removing that requirement would require extensive changes to `OIDCClient` which seemed beyond the scope of this PR. My gut feeling is that this would be made redundant by CBG-2170 - though I'm happy to reconsider if you feel strongly.

auth/auth.go
markspolakovs: The problem is that `LocalJWTAuthProvider` doesn't have a `name` field like `OIDCProvider` does - the latter gets filled in in `NewDatabaseContext` but I didn't want to introduce that extra complexity, so the closest thing we have to a name is the map key.

auth/jwt.go
adamcfraser: The updated comment still sounds like j.ClientID==nil should return false, while j.ClientID="" should return true.  I think this is more than a nitpick, as we want to avoid bugs that fail to set j.ClientID and result in audience validation being skipped.

adamcfraser (changes requested):
The nil clientID handling still doesn't look quite right.

auth/jwt.go
markspolakovs: Makes sense. I've changed it to return false, i.e. fail closed rather than open.

auth/auth.go
markspolakovs: This gets improved in #5604 - it separates out the provider config from the "runtime" provider, which means we can initialise UserPrefix when the first becomes the second and remove the need for this altogether.

adamcfraser (approved):

PR #5597, opened 2022-06-29
CBG-2173: Stop processing mutations after DCPClient closes

markspolakovs:
CBG-2173

While deflaking `base.TestResumeStoppedFeed` I uncovered a scenario where, after a `DCPClient` is closed, we could occasionally skip sequences. This PR makes the client ignore any mutations that it receives after it starts closing down - this means that their seqnos won't get persisted, and the next time the client starts they'll get processed properly.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/354/

### DIFF
--- a/base/dcp_client.go
+++ b/base/dcp_client.go
@@ -151,7 +151,9 @@ func (dc *DCPClient) GetMetadata() []DCPMetadata {
 // closes that channel.
 func (dc *DCPClient) close() {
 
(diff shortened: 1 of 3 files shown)


markspolakovs:
This is a sporadic failure so passing unit tests don't necessarily mean all is well, but I've been running the test with this change for the past ~4 hours without fail so far.

adamcfraser (approved):

PR #5598, opened 2022-06-29
fix linter issues prior to CBG-2154

torcolvin:
- Ignore nestif for high cyclomatic complexity, these are hard to refactor and obvious on looking.

- [ ] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/354/

### DIFF
--- a/db/blip.go
+++ b/db/blip.go
@@ -27,18 +27,18 @@ const (
 )
 
 var (
-	// kCompressedTypes are MIME types that explicitly indicate they're compressed:
+	// compressedTypes are MIME types that explicitly indicate they're compressed:
 	compressedTypes = regexp.MustCompile(`(?i)\bg?zip\b`)
 
-	// kGoodTypes are MIME types that are compressible:
+	// goodTypes are MIME types that are compressible:
 	goodTypes = regexp.MustCompile(`(?i)(^text)|(xml\b)|(\b(html|json|yaml)\b)`)
 
-	// kBadTypes are MIME types that are generally incompressible:
+	// badTypes are MIME types that are generally incompressible:
 	badTypes = regexp.MustCompile(`(?i)^(audio|image|video)/`)
 	// An interesting type is SVG (image/svg+xml) which matches _both_! (It's compressible.)
 	// See <http://www.iana.org/assignments/media-types/media-types.xhtml>
 
-	// kBadFilenames are filename extensions of incompressible types:
+	// badFilenames are filename extensions of incompressible types:
 	badFilenames = regexp.MustCompile(`(?i)\.(zip|t?gz|rar|7z|jpe?g|png|gif|svgz|mp3|m4a|ogg|wav|aiff|mp4|mov|avi|theora)$`)
 )
 
--- a/db/blip.go
+++ b/db/blip.go
@@ -12,6 +12,7 @@ package db
 
 import (
 	"context"
+	"regexp"
 	"strings"
 
 	"github.com/couchbase/go-blip"
@@ -23,6 +24,22 @@ const (
 	// sub protocol.  One must match identically with one provided by the peer (CBLite / ISGR)
 	BlipCBMobileReplicationV2 = "CBMobile_2"
 	BlipCBMobileReplicationV3 = "CBMobile_3"
+)
+
+var (
+	// kCompressedTypes are MIME types that explicitly indicate they're compressed:
+	compressedTypes = regexp.MustCompile(`(?i)\bg?zip\b`)
+
+	// kGoodTypes are MIME types that are compressible:
+	goodTypes = regexp.MustCompile(`(?i)(^text)|(xml\b)|(\b(html|json|yaml)\b)`)
+
+	// kBadTypes are MIME types that are generally incompressible:
+	badTypes = regexp.MustCompile(`(?i)^(audio|image|video)/`)
+	// An interesting type is SVG (image/svg+xml) which matches _both_! (It's compressible.)
+	// See <http://www.iana.org/assignments/media-types/media-types.xhtml>
+
+	// kBadFilenames are filename extensions of incompressible types:
+	badFilenames = regexp.MustCompile(`(?i)\.(zip|t?gz|rar|7z|jpe?g|png|gif|svgz|mp3|m4a|ogg|wav|aiff|mp4|mov|avi|theora)$`)
 )
 
 // NewSGBlipContext returns a go-blip context with the given ID, initialized for use in Sync Gateway.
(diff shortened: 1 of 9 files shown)


db/blip_messages.go
```diff
@@ -20,7 +20,8 @@ import (
 	"github.com/couchbase/sync_gateway/base"
 )
 
-type BLIPMessageSender interface {
+// blipMessageSender validates specific reuest types
```
bbrks:
```suggestion
// blipMessageSender validates specific request types
```

db/blip.go
```diff
@@ -25,6 +26,22 @@ const (
...
+	// kGoodTypes are MIME types that are compressible:
+	goodTypes = regexp.MustCompile(`(?i)(^text)|(xml\b)|(\b(html|json|yaml)\b)`)
+
+	// kBadTypes are MIME types that are generally incompressible:
+	badTypes = regexp.MustCompile(`(?i)^(audio|image|video)/`)
+	// An interesting type is SVG (image/svg+xml) which matches _both_! (It's compressible.)
+	// See <http://www.iana.org/assignments/media-types/media-types.xhtml>
+
+	// kBadFilenames are filename extensions of incompressible types:
+	badFilenames = regexp.MustCompile(`(?i)\.(zip|t?gz|rar|7z|jpe?g|png|gif|svgz|mp3|m4a|ogg|wav|aiff|mp4|mov|avi|theora)$`)
```
markspolakovs: Comment prefixes may need updating - still have the `k` prefix

db/blip_sync_context.go
```diff
@@ -191,11 +174,9 @@ func (bsc *BlipSyncContext) register(profile string, handlerFn func(*blipHandler
 				response.SetError("HTTP", status, msg)
 			}
 			base.InfofCtx(bsc.loggingCtx, base.KeySyncMsg, "#%d: Type:%s   --> %d %s Time:%v", handler.serialNumber, profile, status, msg, time.Since(startTime))
-		} else {
 			// Log the fact that the handler has finished, except for the "subChanges" special case which does it's own termination related logging
-			if profile != "subChanges" {
-				base.DebugfCtx(bsc.loggingCtx, base.KeySyncMsg, "#%d: Type:%s   --> OK Time:%v", handler.serialNumber, profile, time.Since(startTime))
-			}
+		} else if profile != "subChanges" {
+			base.DebugfCtx(bsc.loggingCtx, base.KeySyncMsg, "#%d: Type:%s   --> OK Time:%v", handler.serialNumber, profile, time.Since(startTime))
```
bbrks, line 179: move comment down from above

```suggestion
			// Log the fact that the handler has finished, except for the "subChanges" special case which does it's own termination related logging
			base.DebugfCtx(bsc.loggingCtx, base.KeySyncMsg, "#%d: Type:%s   --> OK Time:%v", handler.serialNumber, profile, time.Since(startTime))
```

db/blip_handler.go
```diff
@@ -343,7 +343,7 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 	_, forceClose := generateBlipSyncChanges(changesDb, channelSet, options, opts.docIDs, func(changes []*ChangeEntry) error {
 		base.DebugfCtx(bh.loggingCtx, base.KeySync, "    Sending %d changes", len(changes))
 		for _, change := range changes {
-			if !strings.HasPrefix(change.ID, "_") {
+			if !strings.HasPrefix(change.ID, "_") { //nolint: nestif
```
bbrks: Do you think `nestif` is overly sensitive with the default settings? We could consider tweaking or disabling that linter entirely.

It seems to have a problem with a lot of code that doesn't warrant being flagged IMO, and would reduce the `//nolint` noise.

db/blip_handler.go
```diff
@@ -715,7 +715,7 @@ func (bsc *BlipSyncContext) sendRevAsDelta(sender *blip.Sender, docID, revID, de
 	bsc.replicationStats.SendRevDeltaRequestedCount.Add(1)
 
 	revDelta, redactedRev, err := handleChangesResponseDb.GetDelta(docID, deltaSrcRevID, revID)
-	if err == ErrForbidden {
+	if err == ErrForbidden { //nolint: gocritic // can't convert if/else if to switch since base.IsFleeceDeltaError is not switchable
```
bbrks, line 718: Could maybe do a conditionless switch, but I'm indifferent and it doesn't feel any cleaner...

```go
	switch {
	case err == ErrForbidden:
		return err
	case base.IsFleeceDeltaError(err):
		base.WarnfCtx(bsc.loggingCtx, "Falling back to full body replication. Error generating delta from %s to %s for key %s - err: %v", deltaSrcRevID, revID, base.UD(docID), err)
		return bsc.sendRevision(sender, docID, revID, seq, knownRevs, maxHistory, handleChangesResponseDb)
	case err == base.ErrDeltaSourceIsTombstone:
		base.TracefCtx(bsc.loggingCtx, base.KeySync, "Falling back to full body replication. Delta source %s is tombstone. Unable to generate delta to %s for key %s", deltaSrcRevID, revID, base.UD(docID))
		return bsc.sendRevision(sender, docID, revID, seq, knownRevs, maxHistory, handleChangesResponseDb)
	case err != nil:
		base.DebugfCtx(bsc.loggingCtx, base.KeySync, "Falling back to full body replication. Couldn't get delta from %s to %s for key %s - err: %v", deltaSrcRevID, revID, base.UD(docID), err)
		return bsc.sendRevision(sender, docID, revID, seq, knownRevs, maxHistory, handleChangesResponseDb)
	}
```

bbrks (changes requested):
Mostly looks sensible.

We can maybe consider tweaking/disabling `nestif` though.
It seems overly sensitive, to the point where adding `//nolint` directives or refactoring would make the code harder to read IMO...

db/blip_handler.go
torcolvin: We could up the amount of cyclomatic complexity for this but I think generally anywhere it finds nestif problems it _probably_ should be a separate function. This is a personal bias.

db/blip_handler.go
torcolvin: I'd be tempted to ignore the linter on this issue, because I think switch statements are great when condition is the same but less useful here.

db/blip_handler.go
torcolvin: Since this doesn't seem configurable in .golangci.yml I'm just going to skip nestif linter. We already disable `gocyclo` which does similar things and it's visibly obvious when a function is too big/complicated and should be spotted in a code review.

bbrks: Yeah, I agree that it's obvious during review if things need to be split more. Don't need to rely on the linter for this.

bbrks (approved):

PR #5599, opened 2022-06-30
fix lint issues

torcolvin:
- don't lint on test constants that are in test code (utilities_test.go)
- make waitForChanges private since it returns private struct
- create a common change handler in utilities_testing.go


torcolvin:
A question would be what to call the common change handler which just was found with `dupl` linter.

bbrks (approved):
Looks good to me 👍

PR #5605, opened 2022-07-01
CBG-2134: Public channel is now added to the default GUEST user config

IsaacLambat:
CBG-2134

- The GUEST users channels are rebuilt on first auth
- The rest tester no longer sets admin party if the database config GUEST user config is filled
- Added testing for this change

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/366/

### DIFF
--- a/auth/user.go
+++ b/auth/user.go
@@ -75,7 +75,6 @@ func (auth *Authenticator) defaultGuestUser() User {
 		auth: auth,
 	}
 	user.Channels_ = user.ExplicitChannels_.Copy()
-	user.Channels_.AddChannel(ch.DocumentStarChannel, 1)
 	return user
 }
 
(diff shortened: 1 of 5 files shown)


auth/user.go
```diff
@@ -75,6 +75,7 @@ func (auth *Authenticator) defaultGuestUser() User {
 		auth: auth,
 	}
 	user.Channels_ = user.ExplicitChannels_.Copy()
+	user.Channels_.AddChannel(ch.DocumentStarChannel, 1)
```
bbrks: Feels cleaner to have `defaultGuestUser` not contain _any_ channel information by default, and have it be calculated on first auth, like how other users get access to the star channel.

That allows the `ch.DocumentStarChannel` assignment to remain in one place (`(Authenticator) rebuildChannels()`).

bbrks (changes requested):

bbrks (approved):
LGTM

PR #5606, opened 2022-07-06
CBG-2157: Allow SG databases to be created for a single named collection

bbrks:
CBG-2157

Allows a SG database to start up aginst a single named scope/collection and use the REST API to read and write documents in that keyspace instead of the default scope and collection.

- Validates that REST API keyspace usage is valid and matches running configuration.
- Exposes `DatabaseContext.Scopes[].Collections[]` for access in handlers (e.g. `getCollections`)
- Supports KV operations on the named collection.
- Persistent config still operates on default scope/collection.

Does not support:
- Query/Index/Views (CBG-2159 / CBG-2158)
- DCP (CBG-2182)

- [ ] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/369/

### DIFF
--- a/base/collection.go
+++ b/base/collection.go
@@ -116,8 +116,17 @@ func GetCollectionFromCluster(cluster *gocb.Cluster, spec BucketSpec, waitUntilR
 	// Safe to get first node as there will always be at least one node in the list and cluster compat is uniform across all nodes.
 	clusterCompatMajor, clusterCompatMinor := decodeClusterVersion(nodesMetadata[0].ClusterCompatibility)
 
+	c := bucket.DefaultCollection()
+	if spec.Collection != nil {
+		if spec.Scope != nil {
+			c = bucket.Scope(*spec.Scope).Collection(*spec.Collection)
+		} else {
+			c = bucket.Collection(*spec.Collection)
+		}
+	}
+
 	collection := &Collection{
-		Collection:                bucket.DefaultCollection(),
+		Collection:                c,
 		Spec:                      spec,
 		cluster:                   cluster,
 		clusterCompatMajorVersion: uint64(clusterCompatMajor),
@@ -698,7 +707,7 @@ func (c *Collection) BucketItemCount() (itemCount int, err error) {
 
 	// TODO: implement APIBucketItemCount for collections as part of CouchbaseStore refactoring.  Until then, give flush a moment to finish
 	time.Sleep(1 * time.Second)
-	//itemCount, err = bucket.APIBucketItemCount()
+	// itemCount, err = bucket.APIBucketItemCount()
 	return 0, err
 }
 
--- a/base/collection.go
+++ b/base/collection.go
@@ -93,6 +93,18 @@ func GetCouchbaseCollection(spec BucketSpec) (*Collection, error) {
 
 }
 
+// bucketSpecScopeAndCollection returns a scope and collection for the given bucket spec.
+func bucketSpecScopeAndCollection(spec BucketSpec) (scope string, collection string) {
+	scope, collection = DefaultScope, DefaultCollection
+	if spec.Scope != nil {
+		scope = *spec.Scope
+	}
+	if spec.Collection != nil {
+		collection = *spec.Collection
+	}
+	return scope, collection
+}
+
 func GetCollectionFromCluster(cluster *gocb.Cluster, spec BucketSpec, waitUntilReadySeconds int) (*Collection, error) {
 
 	// Connect to bucket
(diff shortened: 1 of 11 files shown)


rest/handler.go
```diff
@@ -283,6 +278,33 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
 		}
 	}
 
+	// Named collections handling
+	if dbc := h.server.GetDbConfig(keyspaceDb); dbc != nil && len(dbc.Scopes) > 0 {
```
adamcfraser: Why do we need to look up the set of scopes from the dbConfig, as opposed to checking the scopes defined on the dbContext?

rest/handler.go
```diff
@@ -283,6 +278,33 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
...
+		// Allow an empty scope to refer to the one SG is running with, rather than falling back to _default
+		if keyspaceScope == nil {
+			keyspaceScope = dbContext.BucketSpec.Scope
```
adamcfraser: Similar question here - why BucketSpec.Scope as opposed to checking in dbContext.Scopes?  If the intention is to design with support for multiple Scopes on the dbContext, then should add a way to specify the 'default' scope (not to be confused with _default scope) for the database.

rest/handler.go
```diff
@@ -283,6 +278,33 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
...
+		}
+		if keyspaceCollection == nil {
+			if len(dbc.Scopes[*keyspaceScope].Collections) > 1 {
```
adamcfraser: You're not doing an existence check here (like you are below with foundScope) - can probably just do this once.

db/blip_sync_context.go
```diff
@@ -76,8 +76,8 @@ type BlipSyncContext struct {
 	blipContextDb                    *Database       // 'master' database instance for the replication, used as source when creating handler-specific databases
 	loggingCtx                       context.Context // logging context for connection
 	dbUserLock                       sync.RWMutex    // Must be held when refreshing the db user
-	gotSubChanges                    bool            //nolint: structcheck // false structcheck positive due to https://github.com/golangci/golangci-lint/issues/826
-	continuous                       bool            //nolint: structcheck // false structcheck positive due to https://github.com/golangci/golangci-lint/issues/826
+	gotSubChanges                    bool            // nolint: structcheck // false structcheck positive due to https://github.com/golangci/golangci-lint/issues/826
```
adamcfraser, line 79: Are these comments autogenerated by the linter?  They feel a bit messy.

db/database.go
```diff
@@ -119,6 +119,15 @@ type DatabaseContext struct {
...
 	ServerContextHasStarted      chan struct{}            // Closed via PostStartup once the server has fully started
+	Scopes                       map[string]Scope         // A map keyed by scope name containing a set of scopes/collections.
+}
+
+type Scope struct {
+	Collections map[string]Collection
+}
+
+type Collection struct {
+	Datastore base.Bucket // Where data is stored for the collection - This should be changed to allow multiple collections to share the same underlying SDK connection once support for >1 collection is added.
```
adamcfraser: Is this being used anywhere, or is databaseContext.Bucket still being used for SDK connections?

base/collection.go
```diff
@@ -116,8 +116,17 @@ func GetCollectionFromCluster(cluster *gocb.Cluster, spec BucketSpec, waitUntilR
 	// Safe to get first node as there will always be at least one node in the list and cluster compat is uniform across all nodes.
 	clusterCompatMajor, clusterCompatMinor := decodeClusterVersion(nodesMetadata[0].ClusterCompatibility)
 
+	c := bucket.DefaultCollection()
```
adamcfraser: I don't think bucket.DefaultCollection() is particularly lightweight - we'd be better off only allocating when we're actually using the default collection.
Given that bucket.DefaultCollection is just a helper function for bucket.Scope("_default").Collection("_default), maybe just write a helper function for bucketSpec to return scope and collection name that will do the appropriate substitution of default values?

adamcfraser (changes requested):
Generally looks fine - a few comments.

db/blip_sync_context.go
bbrks: These two fields are due to be removed by @IsaacLambat via some refactoring in #5590 - turns out those linter complaints were semi-legitimate.

db/database.go
bbrks: `databaseContext.Bucket` is still being used, not this - I think this API needs to change to return something that is able to invoke things like `GetDocument()`, etc.

For now I'll make it store the one DatabaseContext reference before I move those APIs to something lower level.

rest/handler.go
bbrks: I've changed it to `dbContext.Scopes` and use the one scope that is running - but I don't think we should add the "default_scope" config option unless/until we actually support >1 scope

adamcfraser (approved):

PR #5607, opened 2022-07-06
CBG-2151: A noRev is only sent on ErrNotFound

IsaacLambat:
CBG-2151

- Added error returning for `GetRaw` for the LeakyBucket
- Added testing to make sure a `noRev` is sent on `ErrNotFound` only.
- An error gets returned from `sendRevision` if it is not `ErrNotFound` otherwise a `noRev` gets sent

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/372/

### DIFF
--- a/db/blip_sync_context.go
+++ b/db/blip_sync_context.go
@@ -541,8 +541,10 @@ func (bsc *BlipSyncContext) sendNoRev(sender *blip.Sender, docID, revID string, 
 // Pushes a revision body to the client
 func (bsc *BlipSyncContext) sendRevision(sender *blip.Sender, docID, revID string, seq SequenceID, knownRevs map[string]bool, maxHistory int, handleChangesResponseDb *Database) error {
 	rev, err := handleChangesResponseDb.GetRev(docID, revID, true, nil)
-	if err != nil {
+	if base.IsDocNotFoundError(err) {
 		return bsc.sendNoRev(sender, docID, revID, seq, err)
+	} else if err != nil {
+		return err
 	}
 
 	base.TracefCtx(bsc.loggingCtx, base.KeySync, "sendRevision, rev attachments for %s/%s are %v", base.UD(docID), revID, base.UD(rev.Attachments))
--- a/db/blip_sync_context.go
+++ b/db/blip_sync_context.go
@@ -544,7 +544,7 @@ func (bsc *BlipSyncContext) sendRevision(sender *blip.Sender, docID, revID strin
 	if base.IsDocNotFoundError(err) {
 		return bsc.sendNoRev(sender, docID, revID, seq, err)
 	} else if err != nil {
-		return err
+		return fmt.Errorf("failed to GetRev for doc %s with rev %s: %w", base.UD(docID).Redact(), base.MD(revID).Redact(), err)
 	}
 
 	base.TracefCtx(bsc.loggingCtx, base.KeySync, "sendRevision, rev attachments for %s/%s are %v", base.UD(docID), revID, base.UD(rev.Attachments))
--- a/rest/api_test.go
+++ b/rest/api_test.go
@@ -7817,7 +7817,7 @@ func TestRevocationUserHasDocAccessDocNotFound(t *testing.T) {
 	require.True(t, ok)
 
 	leakyBucket.SetGetRawCallback(func(s string) error {
-		assert.NoError(t, leakyBucket.Delete("doc"))
+		require.NoError(t, leakyBucket.Delete("doc"))
 		return nil
 	})
 
--- a/rest/api_test.go
+++ b/rest/api_test.go
@@ -7816,8 +7816,9 @@ func TestRevocationUserHasDocAccessDocNotFound(t *testing.T) {
 	leakyBucket, ok := base.AsLeakyBucket(rt.Bucket())
 	require.True(t, ok)
 
-	leakyBucket.SetGetRawCallback(func(s string) {
+	leakyBucket.SetGetRawCallback(func(s string) error {
 		assert.NoError(t, leakyBucket.Delete("doc"))
+		return nil
 	})
 
 	changes = revocationTester.getChanges(changes.Last_Seq, 1)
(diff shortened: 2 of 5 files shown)


rest/blip_api_test.go
```diff
@@ -4494,3 +4494,82 @@ func TestBlipLegacyAttachDocUpdate(t *testing.T) {
...
+			expectNoRev: true,
+		},
+		{
+			name:        "overload",
+			error:       gocb.ErrOverload,
+			expectNoRev: false,
+		},
+	}
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
```
torcolvin: Rather than making an explicit test name, I'd do `fmt.Sprintf("%s", test.error)`

rest/blip_api_test.go
```diff
@@ -4494,3 +4494,82 @@ func TestBlipLegacyAttachDocUpdate(t *testing.T) {
...
+			// Make the LeakyBucket return an error
+			leakyBucket.SetGetRawCallback(func(key string) error {
+				return test.error
+			})
+
+			// Flush cache so document has to be retrieved from the leaky bucket
+			rt.GetDatabase().FlushRevisionCacheForTest()
+
+			err = btc.StartPull()
+			assert.NoError(t, err)
```
torcolvin: Use require here, if this fails, the subsequent data will be invalid.

rest/blip_api_test.go
```diff
@@ -4494,3 +4494,82 @@ func TestBlipLegacyAttachDocUpdate(t *testing.T) {
...
+
+			// Wait 3 seconds for noRev to be received
+			select {
+			case msg := <-recievedNoRevs:
+				if test.expectNoRev {
+					assert.Equal(t, test.name, msg.Properties["id"])
```
torcolvin: I'd change this to require, because the subsequent assertion on 4572 might end up being true if NoRev happens.

rest/blip_api_test.go
```diff
@@ -4494,3 +4494,82 @@ func TestBlipLegacyAttachDocUpdate(t *testing.T) {
...
+			require.True(t, ok)
+
+			btc, err := NewBlipTesterClientOptsWithRT(t, rt, nil)
+			require.NoError(t, err)
+			defer btc.Close()
+
+			// Change noRev handler so it's known when a noRev is received
+			recievedNoRevs := make(chan *blip.Message)
+			btc.pullReplication.bt.blipContext.HandlerForProfile[db.MessageNoRev] = func(msg *blip.Message) {
+				fmt.Println("Received noRev")
```
torcolvin: drop print statement?

rest/blip_api_test.go
```diff
@@ -4494,3 +4494,82 @@ func TestBlipLegacyAttachDocUpdate(t *testing.T) {
...
+				recievedNoRevs <- msg
+			}
+
+			resp := rt.SendAdminRequest(http.MethodPut, "/db/"+test.name, `{"foo":"bar"}`)
+			assertStatus(t, resp, http.StatusCreated)
```
torcolvin: out of scope for this review but I'd be tempted to do a `s/assertStatus/requireStatus/` since this statement is always a require

rest/api_test.go
```diff
@@ -7816,8 +7816,9 @@ func TestRevocationUserHasDocAccessDocNotFound(t *testing.T) {
 	leakyBucket, ok := base.AsLeakyBucket(rt.Bucket())
 	require.True(t, ok)
 
-	leakyBucket.SetGetRawCallback(func(s string) {
+	leakyBucket.SetGetRawCallback(func(s string) error {
 		assert.NoError(t, leakyBucket.Delete("doc"))
```
torcolvin: I'd be tempted to change this to require

db/blip_sync_context.go
```diff
@@ -541,8 +541,10 @@ func (bsc *BlipSyncContext) sendNoRev(sender *blip.Sender, docID, revID string,
 // Pushes a revision body to the client
 func (bsc *BlipSyncContext) sendRevision(sender *blip.Sender, docID, revID string, seq SequenceID, knownRevs map[string]bool, maxHistory int, handleChangesResponseDb *Database) error {
 	rev, err := handleChangesResponseDb.GetRev(docID, revID, true, nil)
-	if err != nil {
+	if base.IsDocNotFoundError(err) {
 		return bsc.sendNoRev(sender, docID, revID, seq, err)
+	} else if err != nil {
+		return err
```
torcolvin: Is there anything we should add to this error message? What conditions trigger this error message (MessageOverload below)? In this case should we assert that the error message returned has enough information so that we know what is going on or should we log here. In this case, we aren't going to hit the trace logging.

rest/blip_api_test.go
IsaacLambat: `assertStatus` actually does a require in the function (for some reason). This should hopefully be fixed in ticket CBG-1984

rest/blip_api_test.go
IsaacLambat: Nice catch

rest/blip_api_test.go
IsaacLambat: Left that in as might be helpful if the test ever needs debugging. Have added extra information (the message properties) to it.

rest/blip_api_test.go
IsaacLambat: Left that as an assert because if it receives a noRev for a different document (which should never happen), then it would be interesting to know if the actual rev does get successfully replicated (which it shouldn't). Does this sound ok?

db/blip_sync_context.go
IsaacLambat: I have added additional information to the error returned. A warning gets logged when this returns.

rest/blip_api_test.go
IsaacLambat: Thanks for PR. Have updated to use new version `requireStatus`

torcolvin: I think it's hard to address general issues as in CBG-1984 but it's easier to fix one offs with a quick sed statement.

bbrks (approved):
LGTM

PR #5609, opened 2022-07-06
CBG-2171 Create User: Key length exceeds 251

cbbruno:
CBG-2171

- Adds check for User/Role name length to admin router.

- [ ] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/373/

### DIFF
--- a/auth/role.go
+++ b/auth/role.go
@@ -105,37 +105,34 @@ func (role *roleImpl) initRole(name string, channels base.Set) error {
 // IsValidPrincipalName checks if the given user/role name would be valid. Valid names must be valid UTF-8, containing
 // at least one alphanumeric (except for the guest user), and no colons, commas, backticks, or slashes.
 func IsValidPrincipalName(name string) bool {
-	/*
-		if len(name) == 0 {
-			return true // guest user
-		}
-		if !utf8.ValidString(name) {
+	if len(name) == 0 {
+		return true // guest user
+	}
+	if !utf8.ValidString(name) {
+		return false
+	}
+	seenAnAlphanum := false
+	for _, char := range name {
+		// Reasons for forbidding each of these:
+		// colons: basic authentication uses them to separate usernames from passwords
+		// commas: fails channels.IsValidChannel, which channels.compileAccessMap uses via SetFromArray
+		// slashes: would need to make many (possibly breaking) changes to routing
+		// backticks: MB-50619
+		if char == '/' || char == ':' || char == ',' || char == '`' {
 			return false
 		}
-		seenAnAlphanum := false
-		for _, char := range name {
-			// Reasons for forbidding each of these:
-			// colons: basic authentication uses them to separate usernames from passwords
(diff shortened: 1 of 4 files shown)


auth/role.go
```diff
@@ -127,6 +128,77 @@ func IsValidPrincipalName(name string) bool {
 	return seenAnAlphanum
 }
 
+// ValidatePrincipalName performs the same checks as IsValidPrincipalName, but adds length check and retuns a more
+// verbose error message.  This function is slower than IsValidPrincipalName, and should be used only for user
+// and role creation.  Names should have a max length of 239 chars, to account for SG prefixes.  If failFast is
+// true, the function will return at the first requirement not satisfied.  If false, it will gather all validation
+// errors and return them as one concatenated error message.
+func ValidatePrincipalName(name string, failFast bool) error {
```
adamcfraser: Since we're only intending to use this in the creation flow, I'm not sure failFast is still needed.  Thoughts?

base/constants.go
```diff
@@ -164,6 +164,9 @@ const (
 
 	// RedactedStr can be substituted in place of any sensitive data being returned by an API. The 'xxxxx' pattern is the same used by Go's url.Redacted() method.
 	RedactedStr = "xxxxx"
+
+	// MaxPrincipalNameLen is the maximum lenght for user or role names (250 minus UserPrefix / RolePrefix), and is used to validate CRUD
```
adamcfraser:
```suggestion
	// MaxPrincipalNameLen is the maximum length for user or role names (250 minus UserPrefix / RolePrefix), and is used to validate CRUD
```

base/constants.go
```diff
@@ -164,6 +164,9 @@ const (
...
+	MaxPrincipalNameLen = 239
```
adamcfraser: How about defining this as `MaxPrincipalNameLen = 250 - len(UserPrefix)` in case we ever change UserPrefix.

auth/role.go
```diff
@@ -127,6 +128,77 @@ func IsValidPrincipalName(name string) bool {
 	return seenAnAlphanum
 }
 
+// ValidatePrincipalName performs the same checks as IsValidPrincipalName, but adds length check and retuns a more
+// verbose error message.  This function is slower than IsValidPrincipalName, and should be used only for user
+// and role creation.  Names should have a max length of 239 chars, to account for SG prefixes.  If failFast is
+// true, the function will return at the first requirement not satisfied.  If false, it will gather all validation
+// errors and return them as one concatenated error message.
```
adamcfraser: The RegisterNewUser function also does user creation.  The name length check should happen there as well.  Given the current approach, calling ValidatePrincipalName there would mean duplicate validation (one call to ValidatePrincipalName, one to IsValidPrincipalName) until we refactor to only validate the name at creation time.
Instead, I think we should just add the length check to IsValidPrincipalName as well for the time being.

adamcfraser (changes requested):
A few comments/suggestions.

adamcfraser (approved):

PR #5611, opened 2022-07-07
CBG-2183: Fix replication panic w/ non-existent role in user role history

### LINKED ISSUE #5610: Panic due to null reference after non-existing role is removed from user
### Sync Gateway version
    {"couchdb":"Welcome","vendor":{"name":"Couchbase Sync Gateway","version":"3.0"},"version":"Couchbase Sync Gateway/3.0.3(1;8187d9a) CE"}

### Operating system
Azure AKS Kubernetes running container `couchbase/sync-gateway:3.0.3-community` - seems to be CentOS

### Config file
```
{
    "bootstrap": {
        "server": "couchbase://***",
        "use_tls_server": false,
        "username": "****",
        "password": "****"
    },
    "api": {
      "public_interface": ":4984",
      "admin_interface": ":4985",
      "metrics_interface": ":4986",
      "https": {}
    },
    "logging": {
        "log_file_path": "/var/tmp/sglogs",
        "console": {
            "enabled": true,
            "log_level": "info",
            "log_keys": [
                "*"
            ]
        }
    }
}
```

### Log output
Full output not available at the moment.  Null reference in go on [`user.go:380`](https://github.com/couchbase/sync_gateway/blob/3.0.3/auth/user.go#L380) when client tries to sync causes panic.  Repeated attempts cause crash loop & denial of service.

### Steps to reproduce
1.  Create user with `admin_roles` including two roles not previously defined (a1 and a2, for example)
2.  (unclear if required to trigger) reference role in sync function
3.  Update the user to not reference one of the roles (update to `['a1']`, for example)
4.  Try to sync from .net client
5.  Crash occurs

### Workaround
Create the non-existing role with request to the `PUT /{bucket}/_role/{role}` endpoint on admin API

### Expected behavior
Should not crash

### Actual behavior
It crashed

markspolakovs:
CBG-2183

`getPrincipal` returns a nil principal and nil error if the principal doesn't exist, but `ChannelGrantedPeriods` was missing a nil check. Add one with an appropriate log message, and add a regression test.

Fixes #5610.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/374/

### DIFF
--- a/rest/blip_api_test.go
+++ b/rest/blip_api_test.go
@@ -4495,6 +4495,7 @@ func TestBlipLegacyAttachDocUpdate(t *testing.T) {
 	}
 }
 
+// Regression test for CBG-2183.
 func TestBlipRevokeNonExistentRole(t *testing.T) {
 	rt := NewRestTester(t, &RestTesterConfig{
 		guestEnabled: false,
--- a/rest/blip_api_test.go
+++ b/rest/blip_api_test.go
@@ -4494,3 +4494,40 @@ func TestBlipLegacyAttachDocUpdate(t *testing.T) {
 		require.Equal(t, keyValueErr.DocumentID, v2Key)
 	}
 }
+
+func TestBlipRevokeNonExistentRole(t *testing.T) {
+	rt := NewRestTester(t, &RestTesterConfig{
+		guestEnabled: false,
+	})
+	defer rt.Close()
+
+	base.SetUpTestLogging(t, base.LevelTrace, base.KeyAll)
+
+	// 1. Create user with admin_roles including two roles not previously defined (a1 and a2, for example)
+	const testUsername = "bilbo"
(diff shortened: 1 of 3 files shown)


rest/blip_api_test.go
```diff
@@ -4494,3 +4494,40 @@ func TestBlipLegacyAttachDocUpdate(t *testing.T) {
 		require.Equal(t, keyValueErr.DocumentID, v2Key)
 	}
 }
+
+func TestBlipRevokeNonExistentRole(t *testing.T) {
```
markspolakovs, line 4499:
```suggestion
// Regression test for CBG-2183.
func TestBlipRevokeNonExistentRole(t *testing.T) {
```

bbrks (approved):
All seems reasonable - will let Adam have a look prior to merging.

adamcfraser (approved):

PR #5612, opened 2022-07-07
rename assert -> require

torcolvin:
The function does a require under the hood, this just matches naming.

### DIFF
--- a/rest/utilities_testing.go
+++ b/rest/utilities_testing.go
@@ -629,7 +629,7 @@ func (rt *RestTester) WaitForViewAvailable(viewURLPath string) (err error) {
 func (rt *RestTester) GetDBState() string {
 	var body db.Body
 	resp := rt.SendAdminRequest("GET", "/db/", "")
-	assertStatus(rt.tb, resp, 200)
+	requireStatus(rt.tb, resp, 200)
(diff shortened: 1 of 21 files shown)


bbrks (approved):

rest/utilities_testing.go
```diff
@@ -756,7 +756,7 @@ func requestByUser(method, resource, body, username string) *http.Request {
 	return r
 }
 
-func assertStatus(t testing.TB, response *TestResponse, expectedStatus int) {
+func requireStatus(t testing.TB, response *TestResponse, expectedStatus int) {
```
IsaacLambat, line 759: Would be good to have a `assertStatus` function added as well that can be used for future tests that need to assert the status

IsaacLambat (changes requested):

rest/utilities_testing.go
torcolvin: Since we don't have one right now, I'd be inclined to wait until it was needed, generally preferring require to assert anyway.

IsaacLambat: Yeah, makes sense. Will add it if I ever need it in the future

IsaacLambat (approved):

PR #5613, opened 2022-07-07
CBG-2185: Permit (but warn) duplicate OIDC/JWT issuers

markspolakovs:
CBG-2185

We added a check for duplicate OIDC/JWT issuers because it causes Implicit Grant flow to be nondeterministic, but this broke existing configs that were valid in lower versions. Allow this but log a warning.

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -700,7 +700,7 @@ func (dbConfig *DbConfig) validateVersion(ctx context.Context, isEnterpriseEditi
 		}
 	}
 
-	seenIssuers := base.Set{}
+	seenIssuers := make(map[string]int)
 	if dbConfig.OIDCConfig != nil {
 		validProviders := len(dbConfig.OIDCConfig.Providers)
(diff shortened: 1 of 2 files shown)


rest/config.go
```diff
@@ -767,10 +761,20 @@ func (dbConfig *DbConfig) validateVersion(ctx context.Context, isEnterpriseEditi
...
 			}
 		}
-		if _, ok := seenIssuers[local.Issuer]; ok {
-			multiError = multiError.Append(fmt.Errorf("duplicate OIDC/JWT issuer: %s", local.Issuer))
-		} else {
-			seenIssuers.Add(local.Issuer)
+		seenIssuers[local.Issuer]++
+	}
+
+	if len(seenIssuers) > 0 {
```
adamcfraser: I think logging a separate warning per duplicated issuer is reasonable, so I propose simplifying this code to just iterate over seenIssues and log when count > 1.

adamcfraser (dismissed):

bbrks (approved):

PR #5618, opened 2022-07-08
CBG-2158: Collections support for index/queries

bbrks:
CBG-2158

Add collection support to SG's index/query code
- All queries now specify an alias for use in `META()`
  - e.g. `FROM $_keyspace AS queryKeyspaceAlias` 
  - Avoids: `Ambiguous reference to field 'b1'` in queries using `META($_keyspace)` (where $_keyspace ~= `b1`.`sc`.`col`)
- Also drop collections in bucket readier when flushing - Flush does not remove collections...
- Adds (skipped/failing) test for collections DCP support
- Adds test for collections bucket API query support
  - End-to-end SG query test somewhat dependent on DCP support

## Dependencies
- [x] #5622 

- [x] `gsi=true` ran locally - no _new_ failures

### DIFF
--- a/base/bucket_n1ql.go
+++ b/base/bucket_n1ql.go
@@ -22,11 +22,12 @@ import (
 	"gopkg.in/couchbase/gocb.v1"
 )
 
-const KeyspaceQueryToken = "$_keyspace" // Token used for bucket name replacement in query statements
-const MaxQueryRetries = 30              // Maximum query retries on indexer error
-const IndexStateOnline = "online"       // bucket state value, as returned by SELECT FROM system:indexes.  Index has been created and built.
-const IndexStateDeferred = "deferred"   // bucket state value, as returned by SELECT FROM system:indexes.  Index has been created but not built.
-const IndexStatePending = "pending"     // bucket state value, as returned by SELECT FROM system:indexes.  Index has been created, build is in progress
+const KeyspaceQueryToken = "$_keyspace"    // Token used for bucket name replacement in query statements
+const KeyspaceAlias = "queryKeyspaceAlias" // Keyspace alias set for the keyspace in FROM statements in queries
+const MaxQueryRetries = 30                 // Maximum query retries on indexer error
+const IndexStateOnline = "online"          // bucket state value, as returned by SELECT FROM system:indexes.  Index has been created and built.
+const IndexStateDeferred = "deferred"      // bucket state value, as returned by SELECT FROM system:indexes.  Index has been created but not built.
+const IndexStatePending = "pending"        // bucket state value, as returned by SELECT FROM system:indexes.  Index has been created, build is in progress
 const PrimaryIndexName = "#primary"
 
 // IndexOptions used to build the 'with' clause
--- a/base/bucket_n1ql.go
+++ b/base/bucket_n1ql.go
@@ -40,7 +40,7 @@ type N1qlIndexOptions struct {
 var _ N1QLStore = &CouchbaseBucketGoCB{}
 
 // Keyspace for a bucket in the default scope and collection can just be a bucket name
-func (bucket *CouchbaseBucketGoCB) EscapedFullyQualifiedKeyspace() string {
+func (bucket *CouchbaseBucketGoCB) EscapedKeyspace() string {
 	return fmt.Sprintf("`%s`", bucket.GetName())
 }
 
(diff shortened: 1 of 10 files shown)


base/bucket_n1ql.go
```diff
@@ -21,11 +22,12 @@ import (
 	"gopkg.in/couchbase/gocb.v1"
 )
 
-const KeyspaceQueryToken = "$_keyspace" // Token used for bucket name replacement in query statements
-const MaxQueryRetries = 30              // Maximum query retries on indexer error
-const IndexStateOnline = "online"       // bucket state value, as returned by SELECT FROM system:indexes.  Index has been created and built.
-const IndexStateDeferred = "deferred"   // bucket state value, as returned by SELECT FROM system:indexes.  Index has been created but not built.
-const IndexStatePending = "pending"     // bucket state value, as returned by SELECT FROM system:indexes.  Index has been created, build is in progress
+const KeyspaceQueryToken = "$_keyspace"    // Token used for bucket name replacement in query statements
```
bbrks:
```suggestion
const KeyspaceQueryToken = "$_keyspace"    // Token used for keyspace name replacement in query statement. The replacement will be an escaped keyspace.
```

base/bucket_n1ql.go
```diff
@@ -37,8 +39,23 @@ type N1qlIndexOptions struct {
 
 var _ N1QLStore = &CouchbaseBucketGoCB{}
 
-// Keyspace for a bucket is bucket name
-func (bucket *CouchbaseBucketGoCB) Keyspace() string {
+// Keyspace for a bucket is just a bucket name
```
bbrks:
```suggestion
// Keyspace for a bucket in the default scope and collection can just be a bucket name
```

db/indexes.go
```diff
@@ -118,17 +118,17 @@ var (
 	// Queries used to check readiness on startup.  Only required for critical indexes.
 	readinessQueries = map[SGIndexType]string{
 		IndexAccess: "SELECT $sync.access.foo as val " +
-			"FROM `%s` " +
+			"FROM %s AS %s " +
```
bbrks, line 121: @adamcfraser Question for you:

Technically we're changing the indicies when running on the `_default` collection.
Correct me if I'm wrong, but these changes should be safe not to bump the index version for a rebuild since all we're doing is specifying a keyspace alias, which is required when `$_keyspace` is a fully qualified collection.

I don't think there are any upgrade/mixed SG version concerns either (3.1 index/3.0 query and vice versa) but I do need to manually verify this.

adamcfraser: You would need to validate that the updated queries (that include `from %s as %s`) work against the 3.0 index definition.  If that works, I agree there shouldn't be any mixed concerns.

base/bucket_n1ql.go
```diff
@@ -21,11 +22,12 @@ import (
 	"gopkg.in/couchbase/gocb.v1"
 )
 
-const KeyspaceQueryToken = "$_keyspace" // Token used for bucket name replacement in query statements
-const MaxQueryRetries = 30              // Maximum query retries on indexer error
-const IndexStateOnline = "online"       // bucket state value, as returned by SELECT FROM system:indexes.  Index has been created and built.
-const IndexStateDeferred = "deferred"   // bucket state value, as returned by SELECT FROM system:indexes.  Index has been created but not built.
-const IndexStatePending = "pending"     // bucket state value, as returned by SELECT FROM system:indexes.  Index has been created, build is in progress
+const KeyspaceQueryToken = "$_keyspace"    // Token used for keyspace name replacement in query statement. The replacement will be an escaped keyspace.
+const KeyspaceAlias = "queryKeyspaceAlias" // Keyspace alias set for the keyspace in FROM statements in queries
```
adamcfraser: Given this name will appear in bucket index definitions as well as logged query statements, I wonder if it's a bit too generic.  Even sgKeyspaceAlias might be better, while still being descriptive.

adamcfraser (changes requested):
Looks good - would like to verify compatibility of new queries with legacy indexes before merging, though.

db/indexes.go
bbrks: It works as expected in mixed versions (these aliased queries can run on a 3.0 index, and the other way around if we needed to support downgrade)

bbrks:
I found an edge case when an index is explicitly created against a `_default._default` (the `system:indexes` entries have different fields set for that particular case), so I'll push up an additional commit to fix that.

adamcfraser (approved):

PR #5620, opened 2022-07-08
CBG-2154 add GetCollections blip handler

torcolvin:
There is one test that fails that I do not think should, and has to do with getting a document on a collection that should only be in the default collection but appears to be somewhat in the special collection. Additionally, are you actually allowed two collections, `_default._default` and the additional one you specify? I'll leave this test to fail until further discussion.


rest/blip_api_test.go
```diff
@@ -4532,3 +4532,109 @@ func TestBlipRevokeNonExistentRole(t *testing.T) {
...
+
+	btc, err := NewBlipTesterClientOptsWithRT(t, rt, nil)
+	require.NoError(t, err)
+	defer btc.Close()
+
+	checkpointBody := db.Body{}
+	checkpointID1 := "checkpoint1"
+	dbInstance := db.Database{DatabaseContext: rt.GetDatabase()}
+	//TODO, use db.checkpointDocIDPrefix exported ?
+	revID, err := dbInstance.PutSpecial(db.DocTypeLocal, "checkpoint/"+checkpointID1, checkpointBody)
```
bbrks: Agreed - although we should maybe move the constant out of the ISGR-specific (ActiveReplicator) file and into maybe `blip_sync_messages.go`

rest/blip_api_test.go
```diff
@@ -4532,3 +4532,109 @@ func TestBlipRevokeNonExistentRole(t *testing.T) {
...
+			errorCode:  "",
+		},
+		{
+			requestBody: db.GetCollectionsRequestBody{
+				CheckpointIDs: []string{"id", "id2"},
+				Collections:   []string{"_default._default"},
+			},
+			resultBody: []db.Body{nil},
+			errorCode:  fmt.Sprintf("%d", http.StatusBadRequest),
+		},
```
bbrks: It might be helpful to include a short test name that describes what's being tested for each case (which can also be passed into `t.Run(...)`)

This one took a second to figure out that the expected error was due to the mismatched array lengths.

rest/blip_api_test.go
```diff
@@ -4532,3 +4532,109 @@ func TestBlipRevokeNonExistentRole(t *testing.T) {
...
+		{
+			requestBody: db.GetCollectionsRequestBody{
+				CheckpointIDs: []string{checkpointID1},
+				Collections:   []string{"_default._default"},
+			},
+			resultBody: []db.Body{db.Body{}},
```
bbrks: It seems unexpected that this returns the checkpoint if `PutSpecial` was run on the database that is running in `fooScope.fooCollection`

rest/blip_api_test.go
```diff
@@ -4532,3 +4532,109 @@ func TestBlipRevokeNonExistentRole(t *testing.T) {
...
+			errorCode:  fmt.Sprintf("%d", http.StatusBadRequest),
+		},
+		{
+			requestBody: db.GetCollectionsRequestBody{
+				CheckpointIDs: []string{checkpointID1},
+				Collections:   []string{"fooScope.fooCollection"},
+			},
+			// FIXME: this test should return a nil but actually returns empty body
+			resultBody: []db.Body{nil},
+			errorCode:  "",
```
bbrks: This behaviour seems correct to me? (You wrote the checkpoint to the database being run on `fooScope.fooCollection`)

db/blip_handler_collections.go
```diff
@@ -0,0 +1,94 @@
...
+func (bh *blipHandler) handleGetCollections(rq *blip.Message) error {
+	var requestBody GetCollectionsRequestBody
+	err := rq.ReadJSONBody(&requestBody)
+	if err != nil {
+		return fmt.Errorf("Error deserializing json Sync err: %w body: %s", err, requestBody)
+	}
+	bh.logEndpointEntry(rq.Profile(), requestBody.String())
+
+	if len(requestBody.Collections) != len(requestBody.CheckpointIDs) {
+		// TODO: can I send the information back to the client, is this leaking data? Should I log?
```
bbrks: You can tell them what was wrong with their request (mismatched array lengths) without leaking data - I don't think there's much more that needs to be sent.

db/blip_handler_collections.go
```diff
@@ -0,0 +1,94 @@
...
+	checkpoints := make([]Body, len(requestBody.Collections))
+	for i, collection := range requestBody.Collections {
+		scope, collectionName, err := parseScopeAndCollection(collection)
+		if err != nil {
+			return base.HTTPErrorf(http.StatusBadRequest, "Invalid specification for collection: %s", err)
+		}
+
+		var db *Database
+		if isDefaultScopeAndCollection(*scope, *collectionName) {
+			db = bh.db
```
bbrks: Because we currently only support one collection, bh.db refers to the one namaed collection that was defined in the db config.

It's feasible to treat `_default._default` like any other named collection, and pass it into `bh.db.Scopes[...]` below.

db/blip_handler_collections.go
```diff
@@ -0,0 +1,94 @@
...
+			}
+			db = &Database{DatabaseContext: collection.CollectionCtx}
+		}
+
+		value, err := db.GetSpecial(DocTypeLocal, checkpointDocIDPrefix+requestBody.CheckpointIDs[i])
+		if err != nil {
+			checkpoints[i] = nil
+			continue
+		}
+		delete(value, BodyRev)
```
bbrks: `_rev` needs to be left in the checkpoint body for CBL to be able to use it when writing back future checkpoints.
Unlike `getCheckpoint`, which deletes it and returns it in the response property, we can't as we can have >1 rev ID to return.

db/blip_handler_collections.go
```diff
@@ -0,0 +1,94 @@
...
+		delete(value, BodyId)
+		checkpoints[i] = value
+	}
+	response := rq.Response()
+	if response == nil {
+		// TODO: give better err message
+		return fmt.Errorf("Could not read response")
+	}
+	// TODO: Marshaling here when we could use raw bytes all the way from the bucket, but I think this comment was true with GetCheckpoint but not GetCollections
```
bbrks: 👍

db/blip_handler_collections_test.go
```diff
@@ -0,0 +1,81 @@
...
+	testCases := []struct {
+		scope      string
+		collection string
+		isDefault  bool
+	}{
+		{
+			scope:      "foo.bar",
+			collection: "foo.bar",
+			isDefault:  false,
+		},
```
bbrks, line 74: May as well throw in a positive test case

```suggestion
		},
		{
			scope:      "_default"
			collection: "bar",
			isDefault:  false,
		},
		{
			scope:      "_default"
			collection: "_default",
			isDefault:  true,
		},
```

db/blip_handler_collections_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+func TestIsDefaultCollection(t *testing.T) {
+	testCases := []struct {
+		scope      string
+		collection string
+		isDefault  bool
+	}{
+		{
+			scope:      "foo.bar",
+			collection: "foo.bar",
```
bbrks: Should this be:
```suggestion
			scope:      "foo",
			collection: "bar",
```
?

bbrks (dismissed):
We can walk through this on the call

db/blip_handler_collections.go
torcolvin: So, throw a 404 if you pass `[_default/]._default` but it's not present?

bbrks: yup!

torcolvin: This won't be a 404 becuase we pass nil instead of 404s

db/blip_handler_collections.go
bbrks: Unresolving this - `BodyRev` needs to be kept inside the checkpoint, not deleted.

db/blip_handler_collections.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+		collection, ok := scopeOnDb.Collections[*collectionName]
+		if !ok {
+			checkpoints[i] = nil
+			continue
+		}
+		db = &Database{DatabaseContext: collection.CollectionCtx}
+
+		value, err := db.GetSpecial(DocTypeLocal, CheckpointDocIDPrefix+requestBody.CheckpointIDs[i])
+		if err != nil {
```
bbrks, line 75: Do we know if GetSpecial returns a "Not Found" error if the checkpoint doesn't exist? Or does it return an empty checkpoint value?

db/blip_handler_collections.go
```diff
@@ -0,0 +1,87 @@
...
+			checkpoints[i] = nil
+			continue
```
bbrks: We're ok to do this on an unexpected error - but see comment above about "Not Found" handling.

I don't think we have test coverage for that.

rest/blip_api_test.go
```diff
@@ -4661,3 +4661,111 @@ func TestUnsubChanges(t *testing.T) {
...
+		},
+		{
+			name: "inNonDefaultCollection",
+			requestBody: db.GetCollectionsRequestBody{
+				CheckpointIDs: []string{checkpointID1},
+				Collections:   []string{"fooScope.fooCollection"},
+			},
+			resultBody: []db.Body{db.Body{}},
+			errorCode:  "",
+		},
```
bbrks: Need a test case for the correct collection name, but an unseen checkpoint ID.

In this case we need to let CBL know they can use this collection but don't have a checkpoint to start from.

```suggestion
		{
			name: "unseenCheckpointNonDefaultCollection",
			requestBody: db.GetCollectionsRequestBody{
				CheckpointIDs: []string{"id"},
				Collections:   []string{"fooScope.fooCollection"},
			},
			resultBody: []db.Body{db.Body{}},
			errorCode:  "",
		},
```

rest/blip_api_test.go
```diff
@@ -4661,3 +4661,111 @@ func TestUnsubChanges(t *testing.T) {
...
+			resultBody: []db.Body{nil},
+			errorCode:  fmt.Sprintf("%d", http.StatusBadRequest),
+		},
+		{
+			name: "inNonDefaultCollection",
+			requestBody: db.GetCollectionsRequestBody{
+				CheckpointIDs: []string{checkpointID1},
+				Collections:   []string{"fooScope.fooCollection"},
+			},
+			resultBody: []db.Body{db.Body{}},
```
bbrks: If we set a value in the checkpoint body at the start of the test, we can differentiate between correctly retrieving the checkpoint, and just telling CBL they're able to use the collection without a checkpoint.

bbrks (changes requested):
I think the case where we've got a correct collection but an new checkpoint ID isn't handled, or at least not unit tested.

We need this handling for clients that haven't got a prior checkpoint to retrieve, but need to know the collection is valid to use.

db/blip_handler_collections.go
torcolvin: Do we need BodyID and BodyRev here?

db/blip_handler_collections.go
torcolvin: We'll return Body{} here

db/blip_handler_collections.go
```diff
@@ -73,10 +72,12 @@ func (bh *blipHandler) handleGetCollections(rq *blip.Message) error {
 
 		value, err := db.GetSpecial(DocTypeLocal, CheckpointDocIDPrefix+requestBody.CheckpointIDs[i])
 		if err != nil {
-			checkpoints[i] = nil
+			// handle key not found here specially
```
bbrks: We still need to do this (see `base.IsDocNotFoundError`) - and not just all errors

If `err` is not nil and is anything else (e.g. network error) I think we have to return a 503 to let the client know they can't carry on with this replication.

bbrks (changes requested):
Everything else looks good but still need special error handling from GetSpecial

db/blip_handler_collections.go
```diff
@@ -70,11 +70,15 @@ func (bh *blipHandler) handleGetCollections(rq *blip.Message) error {
...
+		value, err := db.GetSpecial(DocTypeLocal, key)
 		if err != nil {
-			// handle key not found here specially
-			// this will return key "_sync:local:checkpoint/id" missing if collection exists but not present
-			checkpoints[i] = Body{}
+			status, _ := base.ErrorAsHTTPStatus(err)
+			if status == http.StatusNotFound {
+				checkpoints[i] = Body{}
+			} else {
+				checkpoints[i] = nil
```
bbrks, line 82: As discussed, let's keep this PR moving but revisit error handling here in CBG-2203 

Could add the TODO/Warning log until then.

```suggestion
				// TODO: CBG-2203 - should we return an error such that the client disconnects here?
				base.WarnfCtx(bh.loggingCtx, "Unable to fetch client checkpoint %q for collection %s.%s", key, *scope, *collectionName)
				checkpoints[i] = nil
```

bbrks (changes requested):

bbrks (approved):

PR #5622, opened 2022-07-11
Fix scope/collections cleanup in test teardown

bbrks:
When flushing a bucket, the scopes and collections are not dropped within it.

This causes issues related to 'scope already exists'-type errors when test teardown only flushes a bucket, and keeps scopes and collections around (despite the documents being removed from them)

This becomes an issue as soon as we have >1 tests using scopes and collections, so will unblock:
- #5620
- #5618

### DIFF
--- a/base/collection.go
+++ b/base/collection.go
@@ -657,6 +657,40 @@ func (c *Collection) isRecoverableWriteError(err error) bool {
 	return false
 }
 
+// dropAllScopesAndCollections attempts to drop *all* non-_default scopes and collections from the bucket associated with the collection.  Intended for test usage only.
+func (c *Collection) dropAllScopesAndCollections() error {
+	cm := c.Bucket().Collections()
+	scopes, err := cm.GetAllScopes(nil)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


base/collection.go
```diff
@@ -657,6 +657,40 @@ func (c *Collection) isRecoverableWriteError(err error) bool {
...
+				return err
+			}
+			continue
+		}
+
+		// can't delete _default scope - but we can delete the non-_default collections within it
+		for _, collection := range scope.Collections {
+			if collection.Name != DefaultCollection {
+				if err := cm.DropCollection(collection, nil); err != nil {
+					WarnfCtx(context.TODO(), "Error dropping collection %s in scope %s on bucket %s: %v  Will retry.", MD(collection.Name).Redact(), MD(scope).Redact(), MD(c.Bucket().Name()).Redact(), err)
```
torcolvin, line 684: Do we know why this would error, or is it just experience from server? are there certain situations where we should retry?

Should we program with no retries and add them when we determine what they are?

torcolvin (approved):

base/collection.go
bbrks: as discussed - temp network errors, etc.

PR #5624, opened 2022-07-11
CBG-2155 add collection namespacing

torcolvin:
First pass at adding collection namespacing, no tests.

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -43,7 +43,7 @@ var handlersByProfile = map[string]blipHandlerFunc{
 	MessageGetRev:          userBlipHandler(collectionBlipHandler((*blipHandler).handleGetRev)),
 	MessagePutRev:          userBlipHandler(collectionBlipHandler((*blipHandler).handlePutRev)),
 
-	MessageGetCollections: userBlipHandler(collectionBlipHandler((*blipHandler).handleGetCollections)),
+	MessageGetCollections: userBlipHandler((*blipHandler).handleGetCollections),
 }
 
 // maxInFlightChangesBatches is the maximum number of in-flight changes batches a client is allowed to send without being throttled.
@@ -135,19 +135,22 @@ func collectionBlipHandler(next blipHandlerFunc) blipHandlerFunc {
 			bh.collection = bh.db
 			return next(bh, bm)
 		}
+		if len(bh.collectionMapping) == 0 {
+			return base.HTTPErrorf(http.StatusBadRequest, "Passing collection requires calling GetCollections first")
+		}
+
 		collectionIndex, err := strconv.Atoi(collectionIndexStr)
 		if err != nil {
 			return base.HTTPErrorf(http.StatusBadRequest, "collection property needs to be an int, was %q", collectionIndexStr)
 		}
 
-		if len(bh.collectionMapping) == 0 {
-			return base.HTTPErrorf(http.StatusBadRequest, "Passing collection requires calling GetCollections first")
-		}
-
 		if len(bh.collectionMapping) <= collectionIndex {
-			return base.HTTPErrorf(http.StatusBadRequest, "Collection index %d does not match a collection set by GetCollections", collectionIndex)
+			return base.HTTPErrorf(http.StatusBadRequest, "Collection index %d is outside indexes set by GetCollections", collectionIndex)
 		}
 		bh.collection = bh.collectionMapping[collectionIndex]
+		if bh.collection == nil {
+			return base.HTTPErrorf(http.StatusBadRequest, "Collection index %d does not match a valid collection from GetCollections", collectionIndex)
+		}
 		// Call down to the underlying handler and return it's value
 		return next(bh, bm)
 	}
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -30,20 +30,20 @@ import (
 
 // handlersByProfile defines the routes for each message profile (verb) of an incoming request to the function that handles it.
 var handlersByProfile = map[string]blipHandlerFunc{
-	MessageGetCheckpoint:   (*blipHandler).handleGetCheckpoint,
-	MessageSetCheckpoint:   (*blipHandler).handleSetCheckpoint,
-	MessageSubChanges:      userBlipHandler((*blipHandler).handleSubChanges),
-	MessageUnsubChanges:    userBlipHandler((*blipHandler).handleUnsubChanges),
-	MessageChanges:         userBlipHandler((*blipHandler).handleChanges),
-	MessageRev:             userBlipHandler((*blipHandler).handleRev),
-	MessageNoRev:           (*blipHandler).handleNoRev,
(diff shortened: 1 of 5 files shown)


db/blip_handler.go
```diff
@@ -124,11 +126,37 @@ func (bh *blipHandler) refreshUser() error {
...
+			return nil
+		}
+		collectionIndex, err := strconv.Atoi(collectionIndexStr)
+		if err != nil {
+			return base.HTTPErrorf(http.StatusBadRequest, "collection property needs to be an int, was %q", collectionIndexStr)
+		}
+
+		if len(bh.collectionMapping) > collectionIndex {
+			return base.HTTPErrorf(http.StatusBadRequest, "Passing collection requires calling GetCollections first")
+		}
```
bbrks: Should this be `if len(bh.collectionMapping) == 0` ?

If you have 1 collection, the index could be `0` and we'd still fall into this case with the current code.

db/blip_handler_collections.go
```diff
@@ -50,10 +50,11 @@ func (bh *blipHandler) handleGetCollections(rq *blip.Message) error {
 		// TODO: can I send the information back to the client, is this leaking data? Should I log?
 		return base.HTTPErrorf(http.StatusBadRequest, "Length of collections must match length of checkpoint ids. Request body: %v", requestBody)
 	}
+	bh.collectionMapping = requestBody.Collections
```
bbrks: Might be nice to be defensive and assert that we don't already have `bh.collectionMapping` set to ensure client's aren't sending multiple `getCollections` messages on a single connection.

db/blip_handler_collections.go
```diff
@@ -90,3 +91,24 @@ func (bh *blipHandler) handleGetCollections(rq *blip.Message) error {
...
+	if err != nil {
+		return nil, base.HTTPErrorf(http.StatusBadRequest, "Invalid specification for collection %q %s", scopeAndCollection, err)
+	}
+
+	if isDefaultScopeAndCollection(*scope, *collectionName) {
+		return bh.db, nil
+	}
+	scopeOnDb, ok := bh.db.Scopes[*scope]
+	if !ok {
+		// should we return 404 here, to not leak information
```
bbrks:
```suggestion
```

SG users will be authorised for the database (a set of collections), so it should be ok to report 404 back here.

bbrks (dismissed):
just a couple of things

db/blip_handler.go
```diff
@@ -30,29 +30,32 @@ import (
...
+	MessageChanges:         userBlipHandler(collectionBlipHandler((*blipHandler).handleChanges)),
+	MessageRev:             userBlipHandler(collectionBlipHandler((*blipHandler).handleRev)),
+	MessageNoRev:           collectionBlipHandler((*blipHandler).handleNoRev),
+	MessageGetAttachment:   userBlipHandler(collectionBlipHandler((*blipHandler).handleGetAttachment)),
+	MessageProveAttachment: userBlipHandler(collectionBlipHandler((*blipHandler).handleProveAttachment)),
+	MessageProposeChanges:  collectionBlipHandler((*blipHandler).handleProposeChanges),
+	MessageGetRev:          userBlipHandler(collectionBlipHandler((*blipHandler).handleGetRev)),
+	MessagePutRev:          userBlipHandler(collectionBlipHandler((*blipHandler).handlePutRev)),
+
+	MessageGetCollections: userBlipHandler(collectionBlipHandler((*blipHandler).handleGetCollections)),
```
bbrks: This is the one message type that shouldn't have the collection wrapper on it, since this is the message that initializes the collection map.

db/blip_handler_collections.go
```diff
@@ -85,10 +85,32 @@ func (bh *blipHandler) handleGetCollections(rq *blip.Message) error {
...
+	scopeOnDb, ok := bh.db.Scopes[*scope]
+	if !ok {
+		return nil, base.HTTPErrorf(http.StatusNotFound, "%q does not exist for a scope", *scope)
+	}
+	collection, ok := scopeOnDb.Collections[*collectionName]
+	if !ok {
+		return nil, base.HTTPErrorf(http.StatusNotFound, "%q does not exist for a collection", scopeAndCollection)
+	}
+	return &Database{DatabaseContext: collection.CollectionCtx, user: bh.blipContextDb.User(), Ctx: bh.db.Ctx}, nil
+}
```
torcolvin: This code is now dead because I moved it up to an earlier function

db/blip_handler_collections.go
```diff
@@ -85,10 +85,32 @@ func (bh *blipHandler) handleGetCollections(rq *blip.Message) error {
 		}
 		delete(value, BodyId)
 		checkpoints[i] = value
+		// Should we add this if connection is nil?
+		bh.collectionMapping = append(bh.collectionMapping, &Database{
+			DatabaseContext: collection.CollectionCtx,
+			user:            bh.blipContextDb.User(), Ctx: bh.db.Ctx,
+		})
```
torcolvin: I will need to assert bh.collectionMapping is the same length as checkpoints returned, which means I need to append the bh.collectionMapping in the nil to empty body cases.

db/blip_handler.go
```diff
@@ -124,6 +127,32 @@ func (bh *blipHandler) refreshUser() error {
...
+		}
+
+		if len(bh.collectionMapping) == 0 {
+			return base.HTTPErrorf(http.StatusBadRequest, "Passing collection requires calling GetCollections first")
+		}
+
+		if len(bh.collectionMapping) <= collectionIndex {
+			return base.HTTPErrorf(http.StatusBadRequest, "Collection index %d does not match a collection set by GetCollections", collectionIndex)
+		}
+		bh.collection = bh.collectionMapping[collectionIndex]
```
bbrks, line 150: as discussed, check if the mapped collection is nil here and return an appropriate error (404?) to avoid panicing inside a handler trying to use an invalid collection

db/blip_handler.go
```diff
@@ -124,6 +127,32 @@ func (bh *blipHandler) refreshUser() error {
...
+			return next(bh, bm)
+		}
+		collectionIndex, err := strconv.Atoi(collectionIndexStr)
+		if err != nil {
+			return base.HTTPErrorf(http.StatusBadRequest, "collection property needs to be an int, was %q", collectionIndexStr)
+		}
+
+		if len(bh.collectionMapping) == 0 {
+			return base.HTTPErrorf(http.StatusBadRequest, "Passing collection requires calling GetCollections first")
+		}
```
bbrks: I guess we could move this up to the top to avoid some unnessesary parsing work if we're going to bail out anyway.

bbrks (changes requested):

db/blip_handler_collections.go
torcolvin: I can't easily do this until CBG-2204, so deferring this.

bbrks (approved):
This all looks good. Thanks!

PR #5625, opened 2022-07-12
CBG-2197: Avoid registering CBGT PIndex type for emptyAllDocsIndex codepath

bbrks:
CBG-2197

Adds a `skipRegisterImportPindex` parameter to `NewDatabaseContext` if we know we don't need to register CBGT PIndexes

Avoids concurrent map writes to the CBGT PIndex map coming from the TestBucketPool async bucket readier function calling `emptyAllDocsIndex` and creating a temporary DatabaseContext...

### DIFF
--- a/db/attachment_test.go
+++ b/db/attachment_test.go
@@ -47,7 +47,7 @@ func TestBackupOldRevisionWithAttachments(t *testing.T) {
 	xattrsEnabled := base.TestUseXattrs()
 
 	bucket := base.GetTestBucket(t)
-	dbCtx, err := NewDatabaseContext("db", bucket, false, false, DatabaseContextOptions{
+	dbCtx, err := NewDatabaseContext("db", bucket, false, DatabaseContextOptions{
 		EnableXattr: xattrsEnabled,
 		DeltaSyncOptions: DeltaSyncOptions{
 			Enabled:          deltasEnabled,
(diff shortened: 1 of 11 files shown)


bbrks:
Assigning you @markspolakovs as you touched this last when you moved it into `NewDatabaseContext` - which _is_ where it belongs, but it caused this.

markspolakovs (dismissed):
The change makes sense, but seems like a bit of a smell - is there still a risk of concurrent map writes if NewDatabaseContext happens to be called concurrently by two callers that set it to `false` / how likely is that? (Also, adding an extra parameter to a commonly used method that's `false` 90% of the time doesn't seem quite right to me :D).

Approving nonetheless because it's not a strong objection, but something doesn't "feel right".

bbrks:
I agree, but couldn't come up with alternatives other than shoving the bool in DatabaseContextOptions, but I suppose that would be a lot cleaner, so I might do that instead.

I don't think there's a risk otherwise - as anything that calls NewDatabaseContext is holding a ServerContext lock, or is a unit test running sequentially.

adamcfraser (changes requested):
My initial thought was that we should be making this change in the bucket pool utility (emptyAllDocsIndex) to run without a database context, instead of instantiating a database context.  The CBGT index creation isn't the only unnecessary work that's done during database context creation.  

Based on that, I was going to suggest writing something to execute the required query manually.

However, now I'm wondering how this is intended to behave, since queryAllDocs doesn't return documents that don't have sync metadata.  Do we want to ensure those are purged as well?

bbrks:
The `N1QLBucketEmptierFunc` called prior to `emptyAllDocsIndex` does a `DELETE FROM $_keyspace` using a primary index to ensure the bucket is empty. The purpose of `emptyAllDocsIndex` is to compact/purge tombstones.

Changing that to work without a database is more than I want to do at this point. I just want to fix the concurrent map race...

adamcfraser (approved):

PR #5626, opened 2022-07-12
CBG-2198: TestPullOneshotReplicationAPI fix: Wait for pending changes before starting the one-shot replication

bbrks:
CBG-2198

This test doesn't wait for all writes to be cached before running a one-shot replication.
This means if >0 write wasn't cached and the replicator starts quick enough, it can miss documents.

- [x] `TestPullOneshotReplicationAPI` https://jenkins.sgwdev.com/job/SyncGateway-Integration/392/

### DIFF
--- a/rest/replication_api_test.go
+++ b/rest/replication_api_test.go
@@ -900,6 +900,8 @@ func TestPullOneshotReplicationAPI(t *testing.T) {
 		docIDs[i] = docID
 	}
 
+	require.NoError(t, remoteRT.WaitForPendingChanges())
+
 	// Create oneshot replication, verify running
 	replicationID := t.Name()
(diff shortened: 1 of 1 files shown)


rest/replication_api_test.go
```diff
@@ -900,6 +900,8 @@ func TestPullOneshotReplicationAPI(t *testing.T) {
 		docIDs[i] = docID
 	}
 
+	require.NoError(t, remoteRT.WaitForPendingChanges())
```
markspolakovs, line 903: Perhaps worth a brief comment why this line is important?

markspolakovs (approved):
LGTM, makes sense

rest/replication_api_test.go
bbrks: It's a relatively common pattern and to be honest I can't be bothered running another commit through CI for it 😆

PR #5627, opened 2022-07-12
CBG-2143: Added unsupported option to force forbidden errors from REST API

IsaacLambat:
CBG-2143

* Refactored HTTP error responses to be in same place
* Added unsupported DB config option to force forbidden errors via REST API
* Added prepareSyncFn function to output all the variables needed to put in to runSyncFn.
* Added additional logging for API Errors

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/483
- [x] `xattrs=true test=TestForceAPIForbiddenErrors` https://jenkins.sgwdev.com/job/SyncGateway-Integration/512

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -33,6 +33,9 @@ const (
 // this is different from a client specifically requesting a revision they know about, which are treated as a _removal.
 var ErrForbidden = base.HTTPErrorf(403, "forbidden")
 
+var ErrMissing = base.HTTPErrorf(404, "missing")
+var ErrDeleted = base.HTTPErrorf(404, "deleted")
+
 // ////// READING DOCUMENTS:
 
 func realDocID(docid string) string {
@@ -269,7 +272,11 @@ func (db *Database) getRev(docid, revid string, maxHistory int, historyFrom []st
 	}
 
 	if revision.BodyBytes == nil {
-		return DocumentRevision{}, base.HTTPErrorf(404, "missing")
+		if db.ForceAPIForbiddenErrors() {
+			base.InfofCtx(db.Ctx, base.KeyCRUD, "Doc: %s %s is missing", docid, revid)
+			return DocumentRevision{}, ErrForbidden
+		}
+		return DocumentRevision{}, ErrMissing
 	}
 
 	// RequestedHistory is the _revisions returned in the body.  Avoids mutating revision.History, in case it's needed
@@ -284,6 +291,10 @@ func (db *Database) getRev(docid, revid string, maxHistory int, historyFrom []st
 
 	isAuthorized, redactedRev := db.authorizeUserForChannels(docid, revision.RevID, revision.Channels, revision.Deleted, requestedHistory)
 	if !isAuthorized {
+		if db.ForceAPIForbiddenErrors() {
+			base.InfofCtx(db.Ctx, base.KeyCRUD, "Not authorized to view doc: %s %s", docid, revid)
+			return DocumentRevision{}, ErrForbidden
+		}
 		if revid == "" {
 			return DocumentRevision{}, ErrForbidden
 		}
@@ -293,11 +304,11 @@ func (db *Database) getRev(docid, revid string, maxHistory int, historyFrom []st
 	// If the revision is a removal cache entry (no body), but the user has access to that removal, then just
 	// return 404 missing to indicate that the body of the revision is no longer available.
 	if revision.Removed {
-		return DocumentRevision{}, base.HTTPErrorf(404, "missing")
+		return DocumentRevision{}, ErrMissing
 	}
 
 	if revision.Deleted && revid == "" {
-		return DocumentRevision{}, base.HTTPErrorf(404, "deleted")
+		return DocumentRevision{}, ErrDeleted
 	}
 
 	return revision, nil
@@ -317,7 +328,7 @@ func (db *Database) GetDelta(docID, fromRevID, toRevID string) (delta *RevisionD
 	// return 404 missing to indicate that the body of the revision is no longer available.
 	// Delta can't be generated if we don't have the fromRevision body.
 	if fromRevision.Removed {
-		return nil, nil, base.HTTPErrorf(404, "missing")
+		return nil, nil, ErrMissing
 	}
 
 	// If the fromRevision was a tombstone, then return error to tell delta sync to send full body replication
@@ -366,7 +377,7 @@ func (db *Database) GetDelta(docID, fromRevID, toRevID string) (delta *RevisionD
 		}
 
 		if toRevision.Removed {
-			return nil, nil, base.HTTPErrorf(404, "missing")
+			return nil, nil, ErrMissing
 		}
 
 		// If the revision we're generating a delta to is a tombstone, mark it as such and don't bother generating a delta
@@ -495,7 +506,7 @@ func (db *DatabaseContext) getRevision(ctx context.Context, doc *Document, revid
 	// No inline body, so look for separate doc:
 	if bodyBytes == nil {
 		if !doc.History.contains(revid) {
-			return nil, nil, nil, base.HTTPErrorf(404, "missing")
+			return nil, nil, nil, ErrMissing
 		}
 
 		bodyBytes, err = db.getOldRevisionJSON(ctx, doc.ID, revid)
@@ -642,7 +653,7 @@ func (db *Database) get1xRevFromDoc(doc *Document, revid string, listRevisions b
 		if revid == "" {
 			revid = doc.CurrentRev
 			if doc.History[revid].Deleted == true {
-				return nil, false, base.HTTPErrorf(404, "deleted")
+				return nil, false, ErrDeleted
 			}
 		}
 		if bodyBytes, _, attachments, err = db.getRevision(db.Ctx, doc, revid); err != nil {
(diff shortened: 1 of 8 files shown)


db/database.go
```diff
@@ -196,6 +196,7 @@ type UnsupportedOptions struct {
 	SgrTlsSkipVerify          bool                     `json:"sgr_tls_skip_verify,omitempty"`           // Config option to enable self-signed certs for SG-Replicate testing.
 	RemoteConfigTlsSkipVerify bool                     `json:"remote_config_tls_skip_verify,omitempty"` // Config option to enable self signed certificates for external JavaScript load.
 	GuestReadOnly             bool                     `json:"guest_read_only,omitempty"`               // Config option to restrict GUEST document access to read-only
+	AvoidLeakingDataExistence bool                     `json:"avoid_leaking_data_existence,omitempty"`  // Config option to avoid leaking if a documents exist or not to users without access to them
```
bbrks: ~How about: `ForceForbiddenErrors`? Although I'm open to more ideas~

outdated comment

db/crud.go
```diff
@@ -269,7 +272,11 @@ func (db *Database) getRev(docid, revid string, maxHistory int, historyFrom []st
 	}
 
 	if revision.BodyBytes == nil {
-		return DocumentRevision{}, base.HTTPErrorf(404, "missing")
+		if db.ForceAPIForbiddenErrors() {
+			base.InfofCtx(db.Ctx, base.KeyCRUD, "Doc: %s %s is missing", docid, revid)
```
```suggestion
			base.InfofCtx(db.Ctx, base.KeyCRUD, "Doc: %s %s is missing", base.UD(docid), revid)
```

db/crud.go
```diff
@@ -284,6 +291,10 @@ func (db *Database) getRev(docid, revid string, maxHistory int, historyFrom []st
 
 	isAuthorized, redactedRev := db.authorizeUserForChannels(docid, revision.RevID, revision.Channels, revision.Deleted, requestedHistory)
 	if !isAuthorized {
+		if db.ForceAPIForbiddenErrors() {
+			base.InfofCtx(db.Ctx, base.KeyCRUD, "Not authorized to view doc: %s %s", docid, revid)
```
```suggestion
			base.InfofCtx(db.Ctx, base.KeyCRUD, "Not authorized to view doc: %s %s", base.UD(docid), revid)
```

db/crud.go
```diff
@@ -895,6 +905,48 @@ func (db *Database) Put(docid string, body Body) (newRevID string, doc *Document
...
+				// Make sure the user has permission to modify the document before confirming doc existence
+				mutableBody, metaMap, newRevID, err := db.prepareSyncFn(doc, newDoc)
+				if err != nil {
+					return nil, nil, false, nil, err
+				}
+
+				_, _, _, _, _, err = db.runSyncFn(doc, mutableBody, metaMap, newRevID)
+				if err != nil {
+					err = ErrForbidden
+					base.DebugfCtx(db.Ctx, base.KeyCRUD, "Could not modify doc %q due to document exists conflict and sync func rejection: %v", base.UD(doc.ID), err)
```
bbrks: We're losing the underlying error here before we log it. These two lines need to be flipped around I think (or just return `ErrForbidden` directly)

db/crud.go
```diff
@@ -895,6 +905,48 @@ func (db *Database) Put(docid string, body Body) (newRevID string, doc *Document
 			newDoc._rawBody = canonicalBytesForRevID
 		}
 
+		// Handle telling the user if there is a conflict
+		if docExistsConflict {
+			if db.ForceAPIForbiddenErrors() {
+				// Make sure the user has permission to modify the document before confirming doc existence
+				mutableBody, metaMap, newRevID, err := db.prepareSyncFn(doc, newDoc)
+				if err != nil {
+					return nil, nil, false, nil, err
```
bbrks: Is this error also one we care about obscuring?

db/crud.go
```diff
@@ -895,6 +905,48 @@ func (db *Database) Put(docid string, body Body) (newRevID string, doc *Document
...
+				}
+			}
+			return nil, nil, false, nil, base.HTTPErrorf(http.StatusConflict, "Document exists")
+		}
+		if docRevConflict {
+			if db.ForceAPIForbiddenErrors() {
+				// Make sure the user has permission to modify the document before confirming doc existence
+				mutableBody, metaMap, newRevID, err := db.prepareSyncFn(doc, newDoc)
+				if err != nil {
+					return nil, nil, false, nil, err
```
bbrks: ditto about whether we care about this error

db/crud.go
```diff
@@ -895,6 +905,48 @@ func (db *Database) Put(docid string, body Body) (newRevID string, doc *Document
...
+				}
+
+				_, _, _, _, _, err = db.runSyncFn(doc, mutableBody, metaMap, newRevID)
+				if err != nil {
+					err = ErrForbidden
+					base.DebugfCtx(db.Ctx, base.KeyCRUD, "Could not modify doc %q due to document exists conflict and sync func rejection: %v", base.UD(doc.ID), err)
+					return nil, nil, false, nil, err
+				}
```
bbrks: I'm a bit confused about the case  where `docExistsConflict` and `db.ForceAPIForbiddenErrors()` are true and we _don't_ get an error from `runSyncFn` - is that possible? It sounds like we shouldn't return the conflict error, but that's what's going to happen if we fall through.

db/crud.go
```diff
@@ -895,6 +905,48 @@ func (db *Database) Put(docid string, body Body) (newRevID string, doc *Document
...
+		}
+		if docRevConflict {
+			if db.ForceAPIForbiddenErrors() {
+				// Make sure the user has permission to modify the document before confirming doc existence
+				mutableBody, metaMap, newRevID, err := db.prepareSyncFn(doc, newDoc)
+				if err != nil {
+					return nil, nil, false, nil, err
+				}
+
+				_, _, _, _, _, err = db.runSyncFn(doc, mutableBody, metaMap, newRevID)
```
bbrks, line 916: Are these two bits of code for `docRevConflict` and `docExistsConflict` identical aside from some logging? Could they share it?

db/crud.go
```diff
@@ -895,6 +905,48 @@ func (db *Database) Put(docid string, body Body) (newRevID string, doc *Document
...
+				if err != nil {
+					base.DebugfCtx(db.Ctx, base.KeyCRUD, "Could not modify doc %q due to document revision conflict and sync func rejection: %v", base.UD(doc.ID), err)
+					return nil, nil, false, nil, ErrForbidden
+				}
```
bbrks, line 920: same comment about falling through to the conflict error below

bbrks (changes requested):
some initial comments

db/crud.go
IsaacLambat: Not sure I fully understand. This is intentional and, from I understand, wanted behaviour. If the user does have access to write to the document, then they will get the conflict error proving the document exists.

If `forbidden` should be returned on conflict and if `db.ForceAPIForbiddenErrors()` is true, then the sync function does not need to be ran at all but then this means less information for the user that has write access to that document.

bbrks: OK I think I misunderstood where the error we're trying to force as forbidden was. That makes sense.

db/crud.go
```diff
@@ -895,6 +904,32 @@ func (db *Database) Put(docid string, body Body) (newRevID string, doc *Document
...
+				if err != nil {
+					base.DebugfCtx(db.Ctx, base.KeyCRUD, "Could not modify doc %q due to %s and sync func rejection: %v", base.UD(doc.ID), conflictErr, err)
+					return nil, nil, false, nil, ErrForbidden
+				}
+			}
+			return nil, nil, false, nil, conflictErr
+		}
+
+		// Process the attachments, and populate _sync with metadata. This alters 'body' so it has to
+		// be done before calling CreateRevID (the ID is based on the digest of the body.)
```
bbrks, line 926: This isn't the case any more - but it sounds like it should be!

Can you look into this a bit further to see if there's anything we can do about it?

bbrks (changes requested):
I must've forgotten to post this pending comment - my bad.

rest/api_test.go
```diff
@@ -9456,6 +9456,181 @@ func TestAttachmentDeleteOnExpiry(t *testing.T) {
...
+			if (!doc.doNotSync) {
+				access("NoPerms", "chan2");
+				access("Perms", "chan2");
+				requireAccess("chan");
+				channel(doc.channels);
+			}
+		}`,
+		DatabaseConfig: &DatabaseConfig{DbConfig: DbConfig{
+			Unsupported: &db.UnsupportedOptions{
+				ForceAPIForbiddenErrors: true,
```
bbrks: Do you think it's worth making this a subtest that runs the same set of requests against both `true`/`false` and asserts on the status codes appropriate for either?

That way at least we've got one test that should cover both configurations, rather than relying on existing tests to be covering `false` status codes

bbrks (changes requested):
All the changes look good, I have a question about whether the test can cover status codes for both config option values.

I think it'll be a smallish change to if/else around the expected status for each request we make (could use a helper/anonymous function to reduce the repetitiveness of that too).

rest/api_test.go
IsaacLambat: I planned to just rely on the existing tests, but I can make this a subtest and assert on the proper status codes for not forcing API forbidden errors.

bbrks (approved):

PR #5628, opened 2022-07-12
Integration Test: Parameterise CBS docker volume root dir

bbrks:
Allow Couchbase Server to run on a parameterised root volume directory with env var `CBS_ROOT_DIR`

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/395/


IsaacLambat (approved):
Awesome!

Guessing master integration tests will be changed to use the RAM disk by default when this has been proven successful on manual integration runs?

bbrks:
> Awesome!
> 
> Guessing master integration tests will be changed to use the RAM disk by default when this has been proven successful on manual integration runs?

That should be set from the thing running it (i.e. the Jenkins job) - rather than in master section of the script, because it requires specific setup which may not be true for everybody running the script.

PR #5629, opened 2022-07-13
CBG-1436: Switch sharded import DCP feed to gocb v2

markspolakovs:
CBG-1436

Switches the import DCP feed to GoCB v2.

**NB**: This will only be used for EE - CE will likely be addressed as part of CBG-1143 as it uses the same code.

# Dependencies

- [x] cbgt changes (https://review.couchbase.org/c/cbgt/+/177563, https://review.couchbase.org/c/cbgt/+/177568/2, https://review.couchbase.org/c/cbgt/+/177572/2)
- [x] Bump cbgt in go.mod once those are merged

- [x] `server=7.0.3,protocol=couchbase` https://jenkins.sgwdev.com/job/SyncGateway-Integration/465/
- [x] `server=6.6.5,protocol=couchbase` https://jenkins.sgwdev.com/job/SyncGateway-Integration/466/


base/bucket.go
```diff
@@ -210,12 +210,16 @@ func (spec *BucketSpec) GetGoCBConnString() (string, error) {
...
+		spec.KvPoolSize = 1
 	} else {
-		spec.KvPoolSize, _ = strconv.Atoi(poolSizeFromConnStr)
+		poolSizeFromConnStr := asValues.Get("kv_pool_size")
+		if poolSizeFromConnStr == "" {
+			asValues.Set("kv_pool_size", DefaultGocbKvPoolSize)
+			spec.KvPoolSize, _ = strconv.Atoi(DefaultGocbKvPoolSize)
+		} else {
+			spec.KvPoolSize, _ = strconv.Atoi(poolSizeFromConnStr)
+		}
```
markspolakovs: Can be reverted once https://review.couchbase.org/c/cbgt/+/177568/2 goes in.

base/dcp_sharded.go
```diff
@@ -306,6 +307,8 @@ func initCBGTManager(bucket Bucket, spec BucketSpec, cfgSG cbgt.Cfg, dbUUID stri
 	options := make(map[string]string)
 	options[cbgt.FeedAllotmentOption] = cbgt.FeedAllotmentOnePerPIndex
 	options["managerLoadDataDir"] = "false"
+	// Ensure we always use TLS if configured - cbgt defaults to non-TLS on initial connection
+	options["feedInitialBootstrapNonTLS"] = strconv.FormatBool(!spec.IsTLS())
```
markspolakovs, line 312: Needs https://review.couchbase.org/c/cbgt/+/177563 to go in first

base/bucket.go
```diff
@@ -197,7 +197,7 @@ func (spec BucketSpec) UseClientCert() bool {
 }
 
 // GetGoCBConnString builds a gocb (v1 or v2 depending on the BucketSpec.CouchbaseDriver) connection string based on BucketSpec.Server.
-func (spec *BucketSpec) GetGoCBConnString() (string, error) {
+func (spec *BucketSpec) GetGoCBConnString(dcp bool) (string, error) {
```
adamcfraser: Instead of passing a dcp bool to this function, I think it would be better to pass an options struct with bool flags for values we want to be configurable (kv pool size now, and I anticipate the memcached read cache buffer in the near future).

base/bucket.go
```diff
@@ -210,12 +210,16 @@ func (spec *BucketSpec) GetGoCBConnString() (string, error) {
 	asValues := url.Values(connSpec.Options)
 
 	// Add kv_pool_size as used in both GoCB versions
-	poolSizeFromConnStr := asValues.Get("kv_pool_size")
-	if poolSizeFromConnStr == "" {
-		asValues.Set("kv_pool_size", DefaultGocbKvPoolSize)
-		spec.KvPoolSize, _ = strconv.Atoi(DefaultGocbKvPoolSize)
+	if dcp {
+		spec.KvPoolSize = 1
```
adamcfraser: Do we actually want to update the spec here?  If that's the case, can we skip the process described above and just define a custom bucket spec for use by DCP?

base/dcp_feed_type.go
```diff
@@ -131,6 +145,39 @@ func SGFeedSourceUUIDLookUp(sourceName, sourceParams, serverIn string, options m
...
+func SGGoCBFeedStartDCPFeed(mgr *cbgt.Manager, feedName, indexName, indexUUID,
+	sourceType, sourceName, bucketUUID, params string,
+	dests map[string]cbgt.Dest) error {
+
+	paramsWithAuth := addCbgtAuthToDCPParams(params)
+	return cbgt.StartGocbcoreDCPFeed(mgr, feedName, indexName, indexUUID, sourceType, sourceName, bucketUUID,
+		paramsWithAuth, dests)
+}
+
+func SGGoCBFeedPartitions(sourceType, sourceName, sourceUUID, sourceParams,
```
adamcfraser, line 245: If any of these handlers have identical behaviour regardless of DCP feed type, we can just use the shared definition.  Is that the case here?

base/dcp_sharded.go
```diff
@@ -88,7 +89,8 @@ func GenerateLegacyIndexName(dbName string) string {
 // will receive PIndexImpl callbacks (New, Open) for assigned PIndex to initiate DCP processing.
 func createCBGTIndex(c *CbgtContext, dbName string, configGroupID string, bucket Bucket, spec BucketSpec, numPartitions uint16) error {
 
-	sourceType := SOURCE_GOCOUCHBASE_DCP_SG
+	// TODO(CBG-2196): only use gocb on 7.0+
+	sourceType := SOURCE_GOCB_DCP_SG
```
adamcfraser, line 92: I know it's not part of this ticket, but would like to be sure that we're tracking the upgrade work when moving from a gocouchbase to a gocb feed.  Feed definitions end up stored in the bucket and have an association with checkpoints - need to validate that we're resuming from the previous feed's checkpoints when upgrading.

base/dcp_test.go
```diff
@@ -180,15 +180,16 @@ func TestCBGTIndexCreation(t *testing.T) {
...
-					"couchbase",      // sourceType
-					bucket.GetName(), // sourceName
-					bucketUUID,       // sourceUUID
-					sourceParams,     // sourceParams
-					indexType,        // indexType
-					legacyIndexName,  // indexName
-					indexParams,      // indexParams
-					planParams,       // planParams
-					"",               // prevIndexUUID
+					// TODO: this should be cbgt.SOURCE_GOCOUCHBASE but this causes test setup to fail
```
adamcfraser: Should we be keeping the test for the old approach around, and only run with older CBS versions?  Even as a disabled, dev-only test.

adamcfraser: This is maybe a more general question about our integration testing - with all the changes for collections, I suspect we should think about having an option to run the integration tests against a 6.5 server to catch regressions.

adamcfraser (commented):
General approach looks good, some comments on the details.

base/dcp_sharded.go
markspolakovs: Ack, filed https://issues.couchbase.com/browse/CBG-2213

base/bucket.go
markspolakovs: Not quite sure I follow - the doc comment on `KvPoolSize` is

```go
// gocb kv_pool_size - number of pipelines per node. Initialized on GetGoCBConnString
```

so it seems right to set it in here?

base/dcp_feed_type.go
markspolakovs: It's not - the first set of functions use `StartDCPFeed`/`CouchbasePartitions`/`CouchbasePartitionSeqs`/`CouchbaseStats`, the second use `StartGocbcoreDCPFeed`/`CBPartitions`/`CBPartitionSeqs`/`CBStats`.

base/dcp_test.go
markspolakovs: We already have the capability - I've filed CBG-2216 to look into running it on a regular basis.

base/dcp_dest.go
```diff
@@ -48,7 +48,7 @@ func init() {
 	for i := 0; i < len(vbucketIdStrings); i++ {
 		vbucketIdStrings[i] = fmt.Sprintf("%d", i)
 	}
-	feedType = cbgtFeedType_cbdatasource
+	feedType = cbgtFeedType_gocb // TODO(CBG-2196): only use gocb on v7.0+
```
bbrks: I guess this is not the case any more with the gocb DCP 6.x compat - but if it were, would it mean an if/else around a capability and just setting `feedType` to one or the other?

base/dcp_feed_type.go
```diff
@@ -12,14 +16,86 @@ import (
...
 var cbgtCredentials map[string]cbgtCreds
+
+// cbgtBucketToDBName maps bucket names to DB names for DBs currently registered with CBGT. Necessary because in some
+// contexts we only have access to the bucket name.
+var cbgtBucketToDBName map[string]string
+
+// cbgtRootCertPools is a map of bucket UUIDs to cert pools for its CA certs. The documentation comment of
+// cbgtRootCAsProvider describes the behaviour of different values.
+var cbgtRootCertPools map[string]*x509.CertPool
 var cbgtCredentialsLock sync.Mutex
```
bbrks: I guess this can be renamed to something like `cbgtGlobalsLock` now?

base/dcp_feed_type.go
```diff
@@ -12,14 +16,86 @@ import (
...
+	SOURCE_GOCOUCHBASE_DCP_SG = "couchbase-dcp-sg"
+	SOURCE_GOCB_DCP_SG        = "gocb-dcp-sg"
+)
+
+// cbgtRootCAsProvider implements cbgt.RootCAsProvider. It returns a x509.CertPool factory with the root certificates
+// for the given bucket. Edge cases:
+// * If it returns a function that returns nil, TLS is used but certificate validation is disabled.
+// * If it returns a nil function, TLS is disabled altogether.
+func cbgtRootCAsProvider(bucketName, bucketUUID string) func() *x509.CertPool {
+	if pool, ok := cbgtRootCertPools[bucketUUID]; ok {
```
bbrks: We lock when we write to this map, but not when we read from it here. Should we be doing?

base/dcp_feed_type.go
```diff
@@ -12,14 +16,86 @@ import (
 // Cannot use the serverContext to retrieve this information, as the cbgt manager for a database is initialized
 // before the database is added to the server context's set of databases
 var cbgtCredentials map[string]cbgtCreds
+
+// cbgtBucketToDBName maps bucket names to DB names for DBs currently registered with CBGT. Necessary because in some
+// contexts we only have access to the bucket name.
+var cbgtBucketToDBName map[string]string
```
bbrks, line 22: What context do we only have a bucket name in?

This may not work if we have >1 database on a bucket and they're using different credentials/scopes.

The case I'm thinking of in particular is this:
- `db1` `bucket1` `scope1` `[col1,...]`
- `db2` `bucket1` `scope2` `[col1,...]`

Which could be a supported configuration - there's still only a single scope per db.

base/dcp_feed_type.go
```diff
@@ -51,6 +143,12 @@ type SGFeedSourceParams struct {
 
 	// Used to pass the SG database name to SGFeed* shims
 	DbName string `json:"sg_dbname,omitempty"`
+
+	// Scope within the bucket to stream data from.
+	Scope string `json:"scope,omitempty"`
```
bbrks, line 151: Just for my own benefit - does GoCB/CBGT not allow streaming >1 scope?

bbrks (changes requested):
Some initial comments

base/dcp_dest.go
markspolakovs: Effectively - though given this appears to work just fine on 6.x that comment can likely be removed.

base/dcp_feed_type.go
markspolakovs: Good catch

base/dcp_feed_type.go
markspolakovs: `cbgt.GetPoolsDefaultForBucket` has signature `func(server, bucket string, scopes bool) ([]byte, error)`.

base/dcp_feed_type.go
markspolakovs: I believe so:

https://github.com/couchbase/cbgt/blob/91cc2647a6b9cb775e476319ab52b5baf79d85a0/feed_dcp.go#L84-L88

base/dcp_feed_type.go
markspolakovs: As discussed, I'll start the conversation with the cbgt maintainers about changing this, but let's not block this PR on that.

base/dcp_sharded.go
```diff
@@ -328,7 +339,25 @@ func initCBGTManager(bucket Bucket, spec BucketSpec, cfgSG cbgt.Cfg, dbUUID stri
...
+			certs, err := loadCertificatesFromFile(spec.CACertPath)
+			if err != nil {
+				return nil, err
+			}
+			setCbgtRootCertsForBucket(bucketUUID, certs)
+		} else if spec.TLSSkipVerify {
+			setCbgtRootCertsForBucket(bucketUUID, nil)
+		} else {
+			setCbgtRootCertsForBucket(bucketUUID, getSystemCertPool())
+		}
```
bbrks: Have you already seen `getRootCAs()` ? It looks like it's mostly doing the same thing as these load/system cert pool utils.

base/x509_utils.go
```diff
@@ -0,0 +1,42 @@
...
+// systemCertPool is used to lazy-load the system cert pool in getSystemCertPool
+var systemCertPool atomic.Value
+
+// getSystemCertPool returns a CertPool for the system root store.
+func getSystemCertPool() *x509.CertPool {
+	val := systemCertPool.Load()
+	if val != nil {
+		return val.(*x509.CertPool)
+	}
+	pool, err := x509.SystemCertPool()
```
bbrks: SystemCertPool support for Windows was only added in 1.18, which I think means we may have to handle this slightly more gracefully?

In the other x509 util, we create an empty cert pool and let it fail when connecting if we error.

bbrks (changes requested):
Another couple of comments but I think this PR now seems mostly there

bbrks (approved):

PR #5630, opened 2022-07-13
Changed API specifications to be up-to-date and pull request template

IsaacLambat:
- Changed API specifications to match current database config options
- Added summary to endpoint without one so it would not show the description as the title
- Updated redocly config to use `apis` instead of `apiDefinitions` which is deprecated
- Updated PR template

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,14 +3,14 @@ CBG-0000
 Describe your PR here...
 - Use bullet points if there's more than one thing changed
 
+## Pre-review checklist
+- [ ] Removed debug logging (`fmt.Print`, `log.Print`, ...)
+- [ ] Logging sensitive data? Make sure it's tagged (e.g. `base.UD(docID)`, `base.MD(dbName)`)
+- [ ] Updated relevant information in the API specifications (such as endpoint descriptions, schemas, ...) in `docs/api`
+
 ## Dependencies (if applicable)
 - [ ] Link upstream PRs
 - [ ] Update Go module dependencies when merged
 
 ## [Integration Tests](https://jenkins.sgwdev.com/job/SyncGateway-Integration/build?delay=0sec)
 - [ ] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/000/
-
-## Reviewer checklist
-- [ ] Check all debug logging (`fmt.Print`, `log.Print`, ...) is removed
-- [ ] Check logged sensitive data is tagged (e.g. `base.UD(docID)`, `base.MD(dbName)`)
-- [ ] Check all relevant information in the API specifications (such as endpoint descriptions, schemas, ...) has been updated in `docs/api`
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,14 +3,14 @@ CBG-0000
 Describe your PR here...
 - Use bullet points if there's more than one thing changed
 
-## Pre-review checklist (remove once done)
-- Removed debug logging (`fmt.Print`, `log.Print`, ...)
-- Logging sensitive data? Make sure it's tagged (e.g. `base.UD(docID)`, `base.MD(dbName)`)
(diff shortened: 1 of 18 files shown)


docs/api/components/schemas.yaml
```diff
@@ -1002,11 +1025,17 @@ Database:
       description: The maximum amount of query operations that can be running at any one point.
       type: integer
       default: 1000
+    scopes:
+      description: Scope and collection specific config.
+      type: object
+      properties:
+        scope_name:
+          $ref: '#/Scopes'
```
```suggestion
      additionalProperties:
        $ref: '#/Scopes'
```

docs/api/components/schemas.yaml
```diff
@@ -963,6 +963,29 @@ Replication-status:
   required:
     - replication_id
   title: Replication-status
+Scopes:
+  description: A map of all the collections with their corrosponding configs for this scope
+  type: object
+  properties:
+    collection_name:
+      $ref: '#/CollectionConfig'
```
bbrks: `collection_name` isn't a literal, should be a user-definable property that has a CollectionsConfig type

```suggestion
    collections:
      description: The set of collections on the scope
      type: object
      additionalProperties:
        $ref: '#/CollectionConfig'
```

docs/api/components/schemas.yaml
```diff
@@ -963,6 +963,29 @@ Replication-status:
...
+  title: Scopes
+CollectionConfig:
+  description: The configuration for the individual collection
+  type: object
+  properties:
+    SyncFn:
+      description: The Javascript function that newly created documents in this collection are ran through.
+      type: string
+      default: 'function(doc){channel(doc.channels);}'
```
bbrks: This won't actually be the default sync function for a collection in its final implementation. Just keep it as an example.

```suggestion
      example: 'function(doc){channel(doc.channels);}'
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,14 +3,14 @@ CBG-0000
 Describe your PR here...
 - Use bullet points if there's more than one thing changed
 
-## Pre-review checklist (remove once done)
```
bbrks, line 6: I think we should leave it here but just remove the `(remove once done)` part, and make each item a checkbox.

It really should be the developers responsibility to make sure this is done before it's considered ready for review, but can't harm to leave the list to remind the reviewer to check.

docs/api/components/schemas.yaml
```diff
@@ -963,6 +963,29 @@ Replication-status:
   required:
     - replication_id
   title: Replication-status
+Scopes:
+  description: A map of all the collections with their corrosponding configs for this scope
```
```suggestion
  description: A map of all the collections with their corresponding configs for this scope
```

bbrks (changes requested):
1. I don't agree about absolving the developer of doing the checklist. But it should be left up as a reminder for reviewers to check these.

2. The scopes/collections parts need changing to reflect the API instead of static `collection_name` properties.

Once the `additionalProperties` is used instead, the specs become clearer that it's a user-definable property name:
 
Screenshot 2022-07-13 at 18 31 41

Screenshot 2022-07-13 at 18 31 27

docs/api/components/schemas.yaml
IsaacLambat: Makes sense. Forgot about that property. Saw a couple other places which didn't use this when they should have so will change them as well.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,14 +3,14 @@ CBG-0000
...
-- Removed debug logging (`fmt.Print`, `log.Print`, ...)
-- Logging sensitive data? Make sure it's tagged (e.g. `base.UD(docID)`, `base.MD(dbName)`)
-- Updated relevant information in the API specifications (such as endpoint descriptions, schemas, ...) in `docs/api`
-
 ## Dependencies (if applicable)
 - [ ] Link upstream PRs
-- [ ] Bump manifest once merged
+- [ ] Update Go module dependencies when merged
```
markspolakovs, line 13: ~~Is this step necessary? Manifests need manual work but if I've understood it correctly `go.mod` should be dealt with automatically by the Go tooling.~~

Never mind, realised you meant updating the go.mod once upstream PRs are merged - ignore me :D

bbrks (approved):

PR #5636, opened 2022-07-15
Make the bootstrapAdminRequest helpers less error-prone by closing resp.Body internally

bbrks:
The `bootstrapAdminRequest` utility functions do not use the typical `rest.TestResponse`, as they're not intercepting the `RestTester` handlers, and are actually making HTTP requests to a running server.

This changes the utility function to read and close the response body, to avoid cases where this wasn't done in tests, and includes assertion helper methods for easier use.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/477/

### DIFF
--- a/rest/bootstrap_test.go
+++ b/rest/bootstrap_test.go
@@ -65,8 +65,8 @@ func TestBootstrapRESTAPISetup(t *testing.T) {
 	resp = bootstrapAdminRequest(t, http.MethodGet, "/db1/", ``)
 	assert.Equal(t, http.StatusOK, resp.StatusCode)
 	var dbRootResp DatabaseRoot
-	require.NoError(t, base.JSONDecoder(resp.Body).Decode(&dbRootResp))
-	require.NoError(t, resp.Body.Close())
+	resp.requireStatus(http.StatusOK)
+	require.NoError(t, base.JSONUnmarshal([]byte(resp.Body), &dbRootResp))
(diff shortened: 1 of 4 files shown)


rest/bootstrap_test.go
```diff
@@ -65,17 +65,16 @@ func TestBootstrapRESTAPISetup(t *testing.T) {
 	resp = bootstrapAdminRequest(t, http.MethodGet, "/db1/", ``)
 	assert.Equal(t, http.StatusOK, resp.StatusCode)
 	var dbRootResp DatabaseRoot
-	require.NoError(t, base.JSONDecoder(resp.Body).Decode(&dbRootResp))
-	require.NoError(t, resp.Body.Close())
+	resp.requireStatus(http.StatusOK)
```
IsaacLambat, line 66: Is there a reason line 57 and 63 and other similar cases did not get changed to `resp.RequireStatus`?

It wouldn't cause any issues if the assert failed and the test continues, however if the initial PUT fails then the whole test is definitely going to fail.

IsaacLambat (dismissed):
LGTM

rest/bootstrap_test.go
bbrks: Just missed them in the find/replace

All of them should be caught now

IsaacLambat (approved):
LGTM

PR #5638, opened 2022-07-18
ActiveReplicator push collection panic fix/test coverage

bbrks:
Two fixes for `ActiveReplicator` push panicing due to nil collection on `sendChanges` `NotifyTerminatedChanges`, either of which fix the panic:
- Initialise `bh.collection` on an `ActivePushReplicator` to match `collectionBlipHandler` behaviour
- Keep `sendChanges` `NotifyTerminatedChanges` on database instead of collection (because users are/will be db-scoped, not collection-scoped)

### Misc fix:
- Connected client: make `GetRev` operate on the `collection` instead of `db`

### Test enhancements:
- Make `ActiveReplicator` report unexpected handler panics on `Stop` - but opt-out for the ISGR codepath (i.e. used for test-only assertions)

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/437/

### DIFF
--- a/base/stats.go
+++ b/base/stats.go
@@ -272,7 +272,7 @@ type CBLReplicationPullStats struct {
 	AttachmentPullCount *SgwIntStat `json:"attachment_pull_count"`
 	// The high watermark for the number of documents buffered during feed processing, waiting on a missing earlier sequence.
 	MaxPending *SgwIntStat `json:"max_pending"`
-	/// The total number of active replications. This metric only counts continuous pull replications.
+	// The total number of active replications. This metric only counts continuous pull replications.
 	NumReplicationsActive *SgwIntStat `json:"num_replications_active"`
 	// The total number of continuous pull replications in the active state.
 	NumPullReplActiveContinuous *SgwIntStat `json:"num_pull_repl_active_continuous"`
@@ -532,6 +532,9 @@ type DbReplicatorStats struct {
 	ConflictResolvedRemoteCount *SgwIntStat `json:"sgr_conflict_resolved_remote_count"`
 	// The total number of conflicting documents that were resolved successfully by a merge action (by the active replicator)
 	ConflictResolvedMergedCount *SgwIntStat `json:"sgr_conflict_resolved_merge_count"`
+
+	// The number of times a handler panicked and didn't know how to recover from it.
+	NumHandlersPanicked *SgwIntStat `json:"-"`
 }
 
 type SecurityStats struct {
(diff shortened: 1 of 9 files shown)


base/stats.go
```diff
@@ -272,7 +272,7 @@ type CBLReplicationPullStats struct {
 	AttachmentPullCount *SgwIntStat `json:"attachment_pull_count"`
 	// The high watermark for the number of documents buffered during feed processing, waiting on a missing earlier sequence.
 	MaxPending *SgwIntStat `json:"max_pending"`
-	/// The total number of active replications. This metric only counts continuous pull replications.
+	// The total number of active replications. This metric only counts continuous pull replications.
```
torcolvin, line 275: does this comment actually belong on the line below? outside scope of review though

db/blip_handler.go
```diff
@@ -440,10 +441,10 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 	// On forceClose, send notify to trigger immediate exit from change waiter
 	if forceClose {
 		user := ""
-		if bh.collection.User() != nil {
-			user = bh.collection.User().Name()
+		if bh.db.User() != nil {
+			user = bh.db.User().Name()
 		}
-		bh.collection.DatabaseContext.NotifyTerminatedChanges(user)
+		bh.db.DatabaseContext.NotifyTerminatedChanges(user)
```
torcolvin, line 447: It's not clear to me why this matters. We copy the DB context after the user has been set on it.

The only change I could see is that we might actually have to call 		bh.db.DatabaseContext.NotifyTerminatedChanges(user) on both collection and db? I'm not sure what NotifyTerminatedChanges does though.

torcolvin (commented):
Which is the part that panics?

db/blip_handler.go
bbrks: For now it doesn't make a difference (db == collection) - but when we have more than one collection, users are stored at the database level and can read/write to multiple collections.

Making this change just covers us when we change `db` to be something other than a customer's data collection.

bbrks:
Forgot to answer your main question:
> Which is the part that panics?

```
if bh.collection.User() != nil {
```
because `collection` was nil for an ISGR push replication.

```
2022-07-20T14:42:04.120Z [WRN] c:replication-push [replication-push] PANIC sending changes: runtime error: invalid memory address or nil pointer dereference
goroutine 125508 [running]:
runtime/debug.Stack()
	/home/ec2-user/tools/org.jenkinsci.plugins.golang.GolangInstallation/1.17.11/src/runtime/debug/stack.go:24 +0x72
github.com/couchbase/sync_gateway/db.(*blipHandler).sendChanges.func1()
	/home/ec2-user/workspace/Sync_Gateway_Pipeline_master/db/blip_handler.go:346 +0x8a
panic({0x19c9c20, 0x29ac2c0})
	/home/ec2-user/tools/org.jenkinsci.plugins.golang.GolangInstallation/1.17.11/src/runtime/panic.go:1047 +0x266
github.com/couchbase/sync_gateway/db.(*Database).User(...)
	/home/ec2-user/workspace/Sync_Gateway_Pipeline_master/db/database.go:840
github.com/couchbase/sync_gateway/db.(*blipHandler).sendChanges(0xc001489d40, 0xc001c85c20, 0xc0010dbf58)
	/home/ec2-user/workspace/Sync_Gateway_Pipeline_master/db/blip_handler.go:443 +0x644
github.com/couchbase/sync_gateway/db.(*ActivePushReplicator)._connect.func2(0x138bbe0)
	/home/ec2-user/workspace/Sync_Gateway_Pipeline_master/db/active_replicator_push.go:124 +0x379
created by github.com/couchbase/sync_gateway/db.(*ActivePushReplicator)._connect
	/home/ec2-user/workspace/Sync_Gateway_Pipeline_master/db/active_replicator_push.go:122 +0xabf
 -- db.(*blipHandler).sendChanges.func1() at blip_handler.go:346
```

base/stats.go
bbrks: I don't think so - the stat is just not accounting for all of the cases it should be doing. The below stat is more specific (and correct)

torcolvin (approved):
LGTM

PR #5641, opened 2022-07-19
CBG-2215: Integration test fixes for 6.x and collections

markspolakovs:
CBG-2215

The test bucket framework always tried to drop scopes and collections, which would fail on Server pre-7, eventually leading to bucket flush timeouts.

## [Integration Tests](https://jenkins.sgwdev.com/job/SyncGateway-Integration/build?delay=0sec)
- [x] `server 7.0.3` https://jenkins.sgwdev.com/job/SyncGateway-Integration/445/
- [ ] `server 6.6.5` https://jenkins.sgwdev.com/job/SyncGateway-Integration/448/

PR #5642, opened 2022-07-19
Don't run collections tests against CBS versions that don't support them

markspolakovs:
## Dependencies

- [x] #5641 

- [x] `Server 6.6.5, package rest` https://jenkins.sgwdev.com/job/SyncGateway-Integration/456/
  * [TestReplicationConcurrentPush flake](https://issues.couchbase.com/browse/CBG-1971)

### DIFF
--- a/base/main_test_cluster.go
+++ b/base/main_test_cluster.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"time"
 
 	"github.com/couchbase/gocb/v2"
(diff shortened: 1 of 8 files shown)


base/util_testing.go
```diff
@@ -670,3 +670,33 @@ func ForAllDataStores(t *testing.T, testCallback func(*testing.T, sgbucket.DataS
 		})
 	}
 }
+
+// clusterCompatibility = MajorVersion * 65536 + MinorVersion
+const minCompatVersionForCollections = 7 * 65536
```
bbrks: Might be worth seeing if we can use `decodeClusterVersion()`/`encodeClusterVersion()` instead (in `base/util.go`)

base/main_test_cluster.go
```diff
@@ -271,3 +272,20 @@ func (c *tbpClusterV2) closeCluster(cluster *gocb.Cluster) {
...
+	defer c.closeCluster(cluster)
+
+	nodesMeta, err := cluster.Internal().GetNodesMetadata(nil)
+	if err != nil {
+		FatalfCtx(context.Background(), "TEST: failed to fetch nodes metadata: %v", err)
+	}
+	minVal := math.MaxInt
+	for _, node := range nodesMeta {
+		if node.ClusterCompatibility < minVal {
+			minVal = node.ClusterCompatibility
```
bbrks: Isn't `ClusterCompatibility` already `min(version)` across all nodes in the cluster? Do we have to check each node?

In other areas we've always just got the first node's cluster compat version for this reason

base/util_testing.go
```diff
@@ -670,3 +670,33 @@ func ForAllDataStores(t *testing.T, testCallback func(*testing.T, sgbucket.DataS
...
+
+// map of server names to cluster compat versions (string->int)
+var serverClusterCompat sync.Map
```
bbrks: I'm not sure I understand the intention of the `c.server` lookup? There's only one server behind a test bucket pool

bbrks (changes requested):

base/main_test_cluster.go
markspolakovs: Ah, didn't know that!

bbrks (approved):
LGTM pending rebase/conflict fix

Issue #3558, opened 2018-05-11
Initialize cache after DCP start


adamcfraser:
We should retrieve lastSequence() after starting the DCP feed, to guarantee that there's no race between cache initialization and DCP startup.

adamcfraser:
First step should be to write a unit test to repro - might require leaky bucket.

tleyden:
@adamcfraser here is my understanding of the issue -- can you confirm this?  (annotation below)


```
┌────┐                   ┌────┐                     ┌─────────┐                        ┌─────────┐
│Sg1 │                   │Sg2 │                     │  CB KV  │                        │ CB DCP  │
└────┘                   └────┘                     └─────────┘                        └─────────┘
                                                                                                  
                                                                                                  
                    Get Latest Sequence                                                           
  ────────────────────────────────────────────────────────▶                                       
                                                                                                  
  ◀────────────────────────────────────────────────────────                                       
                           100                                                                    
                                                                                                  
                                                                                                  
                               Write Doc w/ Seq 101                                               
                           ───────────────────────────────▶                                       
                                                                                                  
                                        Mutation notification Doc w/ Seq 101                      
                           ◀───────────────────────────────────────────────────────────────────   
                                                                                                  
                                                                                                  
                                                                                                  
                                             Start DCP Feed                                       
  ────────────────────────────────────────────────────────────────────────────────────────────▶   
                                                                                                  
                                                                                                  
                                                                                                  
      Write Doc w/ Seq 102                                                                        
  ───────────────────────────────▶                                                                
                                                                                                  
                                                                                                  
                                   Mutation notification Doc w/ Seq 102                           
  ◀────────────────────────────────────────────────────────────────────────────────────────────   
```

- SG1 starts up while SG2 has already been running for some time
- SG1 gets the last seq of 100
- After SG1 gets the last seq of 100, but before SG1 starts the DCP feed, SG2 writes a doc with seq 101
- SG1 starts the DCP feed, but it's cache is in an invalid state because the cache thinks it's is valid from seq 100, however it's missing the doc with seq 101, and so really it's only valid from seq 102 (which happens later)

adamcfraser:
That's the general idea, but there are some additional considerations:
- For nodes doing import/migrate (or accel), 'Start the DCP Feed' resumes from where it left off.  In other scenarios, 'Start the DCP feed' retrieves the vbucket high seqnos and starts from there. 
[...]

PR #5646, opened 2022-07-21
CBG-1143 Switch caching DCP feed to use gocb2

torcolvin:
- swap the lock system in change_cache for a channel, since
  DCPClient.Start() will start processing mutations before it is fully
  started. Add a Started channel to handle this in sgbucket

- https://github.com/couchbase/sg-bucket/pull/76

## [Integration Tests]()
- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/515/


base/dcp_client_metadata.go
```diff
@@ -8,6 +8,13 @@ import (
 	"github.com/couchbase/gocbcore/v10"
 )
 
+const (
+	// DCPMetadataDB uses database backed metadata storage
+	DCPMetadataDB = iota
```
bbrks: nitpick: `database` isn't typical Couchbase Server terminology but we can probably just get away with saying it's anything implementing the `CouchbaseStore` interface to avoid bucket/collection semantics.

db/change_cache.go
```diff
@@ -184,29 +183,34 @@ func (c *changeCache) Init(dbcontext *DatabaseContext, notifyChange func(base.Se
...
+		if dcpStartedErr != nil {
+			// error is handled in StartGOCB2DCPFeed
+			return
+		}
+		// background tasks that perform housekeeping duties on the cache
+		bgt, err := NewBackgroundTask("InsertPendingEntries", c.context.Name, c.InsertPendingEntries, c.options.CachePendingSeqMaxWait/2, c.terminator)
+		// TODO: how to test this error condition?
+		if err != nil {
+			base.ErrorfCtx(context.TODO(), "Failed to initialize task %w", err)
+			dbcontext.Close()
```
bbrks: Does this need to return, to stop a nil `bgt` being added to the `c.backgroundTasks` slice? It seems we're in a state that we can't continue from, but we are continuing regardless.

base/dcp_receiver.go
```diff
@@ -202,9 +204,90 @@ type NoPasswordAuthHandler struct {
...
+	if args.Started != nil {
+		args.Started <- nil
+		close(args.Started)
+	}
+	InfofCtx(loggingCtx, KeyDCP, "Started DCP Feed [%s]", MD(feedName))
+	go func() {
+		select {
+		case <-doneChan:
+			TracefCtx(loggingCtx, KeyDCP, "Closed DCP Feed [%s-%s].", MD(feedName))
+			err = <-doneChan
```
bbrks: The 2nd read of `doneChan` is going to get a different value than the one read by the `case` clause

base/dcp_receiver.go
```diff
@@ -202,9 +204,90 @@ type NoPasswordAuthHandler struct {
...
+			InfofCtx(loggingCtx, KeyDCP, "Closing DCP Feed [%s] based on termination notification", MD(feedName))
+			dcpClient.Close()
+			err = <-doneChan
+			break
+		}
+		if args.DoneChan != nil {
+			close(args.DoneChan)
+		}
+	}()
+	return err
```
bbrks: `err` is only ever assigned to in the above goroutine, which will have likely not started/run to completion before this function returns `err`

I'm not sure how we're handling the errors that come from closing the DCP client in this implementation.

base/dcp_receiver.go
```diff
@@ -202,9 +204,90 @@ type NoPasswordAuthHandler struct {
...
+	}
+
+	_, highSeqNos, statsErr := store.GetStatsVbSeqno(numVbuckets, true)
+	if statsErr != nil {
+		return nil, fmt.Errorf("Unable to obtain high seqnos for one-shot DCP feed: %w", statsErr)
+	}
+
+	metadata := make([]DCPMetadata, numVbuckets)
+	for vbNo := uint16(0); vbNo < numVbuckets; vbNo++ {
+		// TODO: should we/can we initialize VbUUID or other fields?
```
bbrks: I think so. Should be able to use the slice returned by `GetSTatsVbSeqno` with this index - although I have no idea about the failover entries...

base/dcp_receiver.go
```diff
@@ -202,9 +204,90 @@ type NoPasswordAuthHandler struct {
...
+
+// StartGOCB2DCPFeed starts a DCP Feed.
+func StartGOCB2DCPFeed(bucket Bucket, spec BucketSpec, args sgbucket.FeedArguments, callback sgbucket.FeedEventCallbackFunc, dbStats *expvar.Map) error {
+	metadata, err := getHighSeqMetadata(bucket)
+	if err != nil {
+		return err
+	}
+	bucketName := spec.BucketName
+	// TODO: does this need to be more unique, if two sync gateways connect to same server, will this reject the second DCP feed?
+	feedName := fmt.Sprintf("%s-%s:%s", args.ID, ProductAPIVersion, bucketName)
```
bbrks: Yes, feed IDs must be unique or kv will terminate the connection of the existing client

We use `GenerateDcpStreamName` elsewhere to include a UUID in the name

db/change_cache.go
```diff
@@ -184,29 +183,34 @@ func (c *changeCache) Init(dbcontext *DatabaseContext, notifyChange func(base.Se
...
-	}
-	c.backgroundTasks = append(c.backgroundTasks, bgt)
-
-	bgt, err = NewBackgroundTask("CleanSkippedSequenceQueue", c.context.Name, c.CleanSkippedSequenceQueue, c.options.CacheSkippedSeqMaxWait/2, c.terminator)
-	if err != nil {
-		return err
-	}
-	c.backgroundTasks = append(c.backgroundTasks, bgt)
+	go func() {
+		dcpStartedErr := <-dcpStarted
```
bbrks: As mentioned, let's discuss this change, as I'm not clear on what problem it's solving yet that the original lock isn't.

db/change_cache.go
```diff
@@ -184,29 +183,34 @@ func (c *changeCache) Init(dbcontext *DatabaseContext, notifyChange func(base.Se
...
 	return nil
 }
 
 func (c *changeCache) Start(initialSequence uint64) error {
-
-	// Unlock the cache after this function returns.
-	defer c.lock.Unlock()
-
 	// Set initial sequence for sequence buffering
 	c._setInitialSequence(initialSequence)
```
bbrks: The initial sequence is still set and read from places that are acquiring the lock (`DocChanged`, `changeCache.Clear`, `processEntry`, ...)

I'm not convinced this lock is safe to remove given the description of #3558

db/change_cache_test.go
```diff
@@ -1986,7 +1986,7 @@ func BenchmarkProcessEntry(b *testing.B) {
 			defer context.Close()
 
 			changeCache := &changeCache{}
-			if err := changeCache.Init(context, nil, nil); err != nil {
+			if err := changeCache.Init(context, nil, nil, nil); err != nil {
```
bbrks: Passing in a `nil` channel here means the goroutine in Init will block forever trying to read from it, which doesn't seem correct.

base/dcp_client.go
```diff
@@ -49,6 +49,7 @@ type DCPClientOptions struct {
 	FailOnRollback             bool           // When true, the DCP client will terminate on DCP rollback
 	InitialMetadata            []DCPMetadata  // When set, will be used as initial metadata for the DCP feed.  Will override any persisted metadata
 	CheckpointPersistFrequency *time.Duration // Overrides metadata persistence frequency - intended for test use
+	MetadataType               int            // Overrides metadata persistence frequency - intended for test use
```
bbrks: - might be nice to have a type alias like `cbgtFeedType`
- comment needs updating

bbrks (dismissed):
first pass comments

base/dcp_receiver.go
markspolakovs: OTOH, if this is for the caching feed / CE import feed, will we expect to see multiple SGWs connecting?

torcolvin: Changed to use common function.

db/change_cache.go
torcolvin: This resolves the problem of mutation events occurring while `DCPClient.Start()` is happening, because that's blocking in `DatabaseContext.NewDatabaseContext`. Due to locking in the `changeCache`, we can deadlock here in `DCPClient.Start` which will send blocking events which are locked in `changeCache.processEntry`. We don't want housekeeping tasks to also lock in `changeCache.processEntry` while we are opening the database context, so we wait the database context is actually created.

This is a summary of an voice chat.

db/change_cache_test.go
torcolvin: Check for `nil` channel here, for tests.

db/change_cache.go
torcolvin: Added back locking, since it will trigger the race detector in a test.

db/change_cache.go
torcolvin: Added return for clarity.

base/dcp_receiver.go
torcolvin: Stopped shadowing this variable.

base/dcp_receiver.go
torcolvin: This is probably a question for @adamcfraser - these will be populated when the DCPClient starts, is my understanding.

base/dcp_client.go
torcolvin: Updated comments.

bbrks:
I'm using this branch locally and I'm running into an issue where the caching feed does not start up correctly when I run SG, but also doesn't bubble any error or warning up to SG - resulting in a database running without being able to recieve any DCP mutations.

It seems to be reliably hitting this for whatever reason, but I'm not exactly sure why. Can you check if you see the same thing? I've had it happen in `TestCollectionsDCP` on a temporary branch with both yours and Marks' PRs merged in and the last changes feed DCP assertion uncommented, so I think it's reproducible and testable.

It looks like from Wireshark the stream is being opened with no vbucket UUID, the error happens, a rollback occurs and we re-open but with a To sequence of 0 so the stream immediately finishes.

[couchbase-gocbv2-dcp-range-error.pcapng.gz](https://github.com/couchbase/sync_gateway/files/9180860/couchbase-gocbv2-dcp-range-error.pcapng.gz)

Screenshot 2022-07-25 at 12 52 54
Screenshot 2022-07-25 at 12 53 02
Screenshot 2022-07-25 at 12 53 05
Screenshot 2022-07-25 at 12 55 37



https://github.com/couchbase/kv_engine/blob/master/docs/dcp/documentation/commands/stream-request.md#error-codes
> **PROTOCOL_BINARY_RESPONSE_ERANGE (0x22)**
> 
> This error code may be returned for one of the following reason. Check the server logs for more information on which case occurred.
>
> - If the start and end sequence numbers are specified incorrectly. For example the start sequence number cannot be bigger than the current high sequence number in the VBucket. The start sequence number must also be less than the end sequence number.
> - If the snapshot start or snapshot end seqno are not specified correctly. For all stream requests the snapshot start seqno must be less than or equal to the start seqno and the start seqno must be less than or equal to the snapshot end seqno.

For one of my buckets (`b2` below) I see all 1024 vbuckets fail and the entire stream close with a trace log.
For another bucket, I see only a handful of vbuckets fail, and get a partial stream running.

```
2022-07-25T12:10:29.384+01:00 [INF] DCP: Starting mutation feed on bucket b2 due to either channel cache mode or doc tracking (auto-import)
2022-07-25T12:10:29.384+01:00 [INF] DCP: Using DCP feed for bucket: "b2" (based on feed_type specified in config file)
2022-07-25T12:10:29.468+01:00 [INF] DCP: Open stream for vbID 0 failed due to rollback or range error, will roll back metadata and retry: requested value is outside range | {"status_code":34,"bucket":"b2","error_name":"ERANGE","error_description":"Invalid range requested","opaque":14,"last_dispatched_to":"192.168.1.21:11210","last_dispatched_from":"192.168.1.19:64857","last_connection_id":"7c5415b81bd4e40d/db4b04e75ad7318f"}
2022-07-25T12:10:29.470+01:00 [DBG] DCP+: Stream (vb:0) closed, all items streamed
2022-07-25T12:10:29.472+01:00 [INF] DCP: Open stream for vbID 1 failed due to rollback or range error, will roll back metadata and retry: requested value is outside range | {"status_code":34,"bucket":"b2","error_name":"ERANGE","error_description":"Invalid range requested","opaque":16,"last_dispatched_to":"192.168.1.21:11210","last_dispatched_from":"192.168.1.19:64857","last_connection_id":"7c5415b81bd4e40d/db4b04e75ad7318f"}
2022-07-25T12:10:29.473+01:00 [DBG] DCP+: Stream (vb:1) closed, all items streamed
2022-07-25T12:10:29.474+01:00 [INF] DCP: Open stream for vbID 2 failed due to rollback or range error, will roll back metadata and retry: requested value is outside range | {"status_code":34,"bucket":"b2","error_name":"ERANGE","error_description":"Invalid range requested","opaque":18,"last_dispatched_to":"192.168.1.21:11210","last_dispatched_from":"192.168.1.19:64857","last_connection_id":"7c5415b81bd4e40d/db4b04e75ad7318f"}
2022-07-25T12:10:29.475+01:00 [DBG] DCP+: Stream (vb:2) closed, all items streamed

...

2022-07-25T12:10:31.792+01:00 [INF] DCP: Open stream for vbID 1023 failed due to rollback or range error, will roll back metadata and retry: requested value is outside range | {"status_code":34,"bucket":"b2","error_name":"ERANGE","error_description":"Invalid range requested","opaque":2060,"last_dispatched_to":"192.168.1.21:11210","last_dispatched_from":"192.168.1.19:64857","last_connection_id":"7c5415b81bd4e40d/db4b04e75ad7318f"}
2022-07-25T12:10:31.793+01:00 [DBG] DCP+: Stream (vb:1023) closed, all items streamed
2022-07-25T12:10:31.793+01:00 [INF] DCP: Started caching DCP Feed "SG-v-3.1-commit--uuid-6485b66e-0c0a-11ed-a616-acde48001122" for bucket "b2"
2022-07-25T12:10:31.793+01:00 [INF] Created background task: "InsertPendingEntries" with interval 2.5s
2022-07-25T12:10:31.793+01:00 [INF] Created background task: "CleanSkippedSequenceQueue" with interval 30m0s
2022-07-25T12:10:31.793+01:00 [TRC] DCP+: Closed caching DCP Feed "SG-v-3.1-commit--uuid-6485b66e-0c0a-11ed-a616-acde48001122" for "b2" -- base.StartGOCB2DCPFeed.func1() at dcp_receiver.go:279
```

torcolvin:
I've removed the locks that I added (I can remove the unused channel from walrus / sg-bucket. I'm not fully convinced that they don't exist but I can't reproduce this week and neither can jenkins.

base/bucket_gocb.go
```diff
@@ -1500,7 +1500,7 @@ func (bucket *CouchbaseBucketGoCB) StartTapFeed(args sgbucket.FeedArguments, dbS
 func (bucket *CouchbaseBucketGoCB) StartDCPFeed(args sgbucket.FeedArguments, callback sgbucket.FeedEventCallbackFunc, dbStats *expvar.Map) error {
 
 	// TODO: Evaluate whether to use cbgt for non-sharded caching feed, as a way to push concurrency upstream
-	return StartDCPFeed(bucket, bucket.Spec, args, callback, dbStats)
+	return StartGOCB2DCPFeed(bucket, bucket.Spec, args, callback, dbStats)
```
adamcfraser: Can we rename to maybe StartGocbDCPFeed?  I think it's a bit less of a wall of letters and numbers, and there's no gocb1 DCP in the mix.

base/dcp_client.go
```diff
@@ -87,8 +88,14 @@ func NewDCPClient(ID string, callback sgbucket.FeedEventCallbackFunc, options DC
...
 	checkpointPrefix := fmt.Sprintf("%s:%v", client.checkpointPrefix, ID)
-	client.metadata = NewDCPMetadataCS(bucket, numVbuckets, numWorkers, checkpointPrefix)
-
+	switch options.MetadataStoreType {
+	case DCPMetadataDB:
+		client.metadata = NewDCPMetadataCS(bucket, numVbuckets, numWorkers, checkpointPrefix)
+	case DCPMetadataInMemory:
+		client.metadata = NewDCPMetadataMem(numVbuckets)
+	default:
+		panic(fmt.Sprintf("Unknown Metadatatype: %d", options.MetadataStoreType))
```
adamcfraser: Return error here instead of panicking - that's the handling for config errors elsewhere in this function.

base/dcp_mutation_feed.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	numVbuckets, err := store.GetMaxVbno()
+	if err != nil {
+		return nil, fmt.Errorf("Unable to determine maxVbNo when creating DCP client: %w", err)
+	}
+
+	vbUUIDs, highSeqNos, statsErr := store.GetStatsVbSeqno(numVbuckets, true)
+	if statsErr != nil {
+		return nil, fmt.Errorf("Unable to obtain high seqnos for one-shot DCP feed: %w", statsErr)
```
adamcfraser: Error message should be updated (isn't just one-shot).

base/dcp_mutation_feed.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	metadata := make([]DCPMetadata, numVbuckets)
+	for vbNo := uint16(0); vbNo < numVbuckets; vbNo++ {
+		metadata[vbNo].VbUUID = gocbcore.VbUUID(vbUUIDs[vbNo])
+		metadata[vbNo].FailoverEntries = make([]gocbcore.FailoverEntry, 0)
```
adamcfraser: Should we be initializing the FailoverEntries per vbucket, in the same way we're doing in makeVbucketMetadata?  (i.e. setting the VbUUID)

base/dcp_mutation_feed.go
```diff
@@ -0,0 +1,103 @@
...
+		metadata[vbNo].StartSeqNo = gocbcore.SeqNo(highSeqNos[vbNo])
+		metadata[vbNo].EndSeqNo = gocbcore.SeqNo(uint64(0xFFFFFFFFFFFFFFFF))
+		metadata[vbNo].SnapStartSeqNo = gocbcore.SeqNo(highSeqNos[vbNo])
+		metadata[vbNo].SnapEndSeqNo = gocbcore.SeqNo(highSeqNos[vbNo])
+	}
+	return metadata, nil
+}
+
+// StartGOCB2DCPFeed starts a DCP Feed.
+func StartGOCB2DCPFeed(bucket Bucket, spec BucketSpec, args sgbucket.FeedArguments, callback sgbucket.FeedEventCallbackFunc, dbStats *expvar.Map) error {
```
adamcfraser: It doesn't look like we're actually using BucketSpec here (could get spec.BucketName from bucket.Name()).  

If possible I think it would be preferable to pass in the parameters we actually need (including the collection set in future) instead of passing spec around.  In general I think the use case for BucketSpec should be 'I need to establish a new connection to the server from scratch'.

adamcfraser: It looks like dbStats aren't being used.  At minimum I think we need to support the dcp_rollback_count stat, which I think is managed by the DCPClient today.  dcp_backfill_expected/dcp_backfill_completed are not needed.

base/dcp_mutation_feed.go
```diff
@@ -0,0 +1,103 @@
...
+
+	doneChan, err := dcpClient.Start()
+	loggingCtx := context.TODO()
+	if err != nil {
+		ErrorfCtx(loggingCtx, "Failed to start caching DCP Feed %q for bucket %q: %w", feedName, MD(bucketName), err)
+		closeErr := dcpClient.Close()
+		ErrorfCtx(loggingCtx, "Close error from caching DCP Feed %q for bucket %q: %w", feedName, MD(bucketName), closeErr)
+		return err
+	}
+	InfofCtx(loggingCtx, KeyDCP, "Started caching DCP Feed %q for bucket %q", feedName, MD(bucketName))
```
adamcfraser: In the current DCP client we're only starting this goroutine if the terminator is non-nil.  Is the change to always run the goroutine intentional?

base/dcp_mutation_feed.go
```diff
@@ -0,0 +1,103 @@
...
+	go func() {
+		select {
+		case dcpCloseError := <-doneChan:
+			// This is a close because DCP client closed on its own
+			InfofCtx(loggingCtx, KeyDCP, "Forced closed caching DCP Feed %q for %q", feedName, MD(bucketName))
+			// wait for channel close
+			<-doneChan
```
adamcfraser: I don't think this is common practice (select on a channel, then block on the channel again inside the select).  It feels like a good way to end up in a deadlock that can't be escaped by closing the terminator.

One approach would be to just select on the channel, and then behave differently based on whether the error is nil.  The other would be to enhance the DCP client to not overload the done channel for both errors and done.

base/dcp_mutation_feed.go
```diff
@@ -0,0 +1,103 @@
...
+			if dcpCloseError != nil {
+				WarnfCtx(loggingCtx, "Error on closing caching DCP Feed %q for %q: %w", feedName, MD(bucketName), dcpCloseError)
+			}
+			// FIXME: close dbContext here
```
adamcfraser: Does the gocb v2 DCP client attempt to reconnect on error?  The cbdatasource DCP client does, and so SG doesn't need to build any of that handling.  If that's not the case for the new client, we'll need to at that either inside the DCP client, or here.

base/dcp_mutation_feed.go
```diff
@@ -0,0 +1,103 @@
...
+			if dcpCloseErr != nil {
+				WarnfCtx(loggingCtx, "Error on closing caching DCP Feed %q for %q: %w", feedName, MD(bucketName), dcpCloseErr)
+			}
+			dcpCloseErr = <-doneChan
+			if dcpCloseErr != nil {
+				WarnfCtx(loggingCtx, "Error on closing caching DCP Feed %q for %q: %w", feedName, MD(bucketName), dcpCloseErr)
+			}
+			break
+		}
+		if args.DoneChan != nil {
```
adamcfraser: Does it make sense to have two separate channels (args.DoneChan and the feed's doneChan)?

base/dcp_mutation_feed.go
```diff
@@ -0,0 +1,103 @@
...
+			}
+			// FIXME: close dbContext here
+			break
+		case <-args.Terminator:
+			InfofCtx(loggingCtx, KeyDCP, "Closing caching DCP Feed %q for bucket %q based on termination notification", feedName, MD(bucketName))
+			dcpCloseErr := dcpClient.Close()
+			if dcpCloseErr != nil {
+				WarnfCtx(loggingCtx, "Error on closing caching DCP Feed %q for %q: %w", feedName, MD(bucketName), dcpCloseErr)
+			}
+			dcpCloseErr = <-doneChan
```
adamcfraser: I feel like it would be preferable to have dcpClient.Close() block on doneChan (or at least have an option to do that) - the double error handling here seems clunky.

adamcfraser (dismissed):
A few questions/comments.

base/bucket_gocb.go
torcolvin: I also moved and changed the messages here, because it's used as import feed for CE.

base/dcp_mutation_feed.go
torcolvin: I haven't found a case that's no-nil - I'd rather make it required if I could? We definitely terminate to shut the database down and the integration tests all pass.

base/dcp_mutation_feed.go
torcolvin: This is a somewhat challenging problem to address right now as per offline discussion with @bbrks. We need the `Bucket` type in `DCPClient` if we are using persistent metadata. Moreover `Bucket` is actually a `Collection` which is conceptually wrong for a `DCPClient`, since it will be acting for multiple collections in a future code review, and we need actually pass something that can persist metadata for multiple collections, which oddly `Bucket` can not do.

I think we _do_ want to change this API, but not right now.

base/dcp_mutation_feed.go
torcolvin: I seeded this with VbUUID = 0, SeqNo = highSeqNo as per  https://github.com/couchbase/kv_engine/blob/master/docs/dcp/documentation/failure-scenarios.md#failover-log-behavior . The old behavior was to seed with SeqNo = 0

base/dcp_mutation_feed.go
torcolvin: Will fix in CBG-2234 soon after this PR goes through.

base/dcp_mutation_feed.go
torcolvin: Will fix in CBG-2234 soon after this PR goes through.

base/dcp_mutation_feed.go
torcolvin: Will fix in CBG-2234 soon after this PR goes through.

base/dcp_mutation_feed.go
torcolvin: Added.

base/dcp_mutation_feed.go
torcolvin: Fixed this so that it reconnects indefinitely if this is not a one shot DCP feed after the initial connection (`DCPClient.Start`), which is retried for 5 minutes.

torcolvin (commented):
I filed future tickets for issues I didn't address in this code review. This has some intermittent failures in jenkins but I'm not sure that they are new or different from what we see.

base/collection.go
```diff
@@ -532,7 +532,11 @@ func (c *Collection) Incr(k string, amt, def uint64, exp uint32) (uint64, error)
 }
 
 func (c *Collection) StartDCPFeed(args sgbucket.FeedArguments, callback sgbucket.FeedEventCallbackFunc, dbStats *expvar.Map) error {
-	return StartDCPFeed(c, c.Spec, args, callback, dbStats)
+	if !IsEnterpriseEdition() {
+		// use new DCP feed in CBG-2232
+		return StartDCPFeed(c, c.Spec, args, callback, dbStats)
+	}
```
bbrks: This does not feel like the right place to decide whether to use cbdatasource or gocb...

We _always_ want to use gocb for the mutation/changes feed, enterprise or not.

Can `StartImportFeed` skip calling `collection.StartDCPFeed` and instead call `Start(Cbdatasource)DCPFeed` directly? We have everything available up one level (including the EE/CE check), and import feels like the special case here.

Screenshot 2022-08-02 at 14 09 52

bbrks (dismissed):
I'm ok with defferring the close cleanup work but how we choose whether to use cbdatasource/gocb needs fixing. See inline comment.

base/collection.go
markspolakovs: ~~In fact, I can't think of a reason why we'd need to use cbdatasource at all - community edition will still support collections (right?), so we have to use gocb for its import feed too. The decision of which feed to use for import should be "sharded or not sharded", not "gocb or cbdatasource" - could that help simplify this code?~~

Ignore me, just saw CBG-2232

torcolvin: Moved this up a level.

bbrks (approved):

PR #5650, opened 2022-07-25
[POC] Unsupported flag for CB <6.5 compatibility

markspolakovs:
This is a proof-of-concept of an unsupported flag to address CBG-2217/CBG-2218 by disabling all code that uses GoCB v2 (since it relies on GCCCP which is only introduced in 6.5). This flag is only available in legacy config mode because persistent config uses gocb v2 for config polling.


markspolakovs:
Closing. Can reopen if we deem it necessary.

PR #5655, opened 2022-07-26
Remove stage parallelism from Jenkins pipeline

bbrks:
First attempt at fixing the `go get .../errcheck permission denied`

I think it might be a race between `Bootstrap` and `Go Tools` stages, so force them to run sequentially like we do on `master`.


bbrks:
Merged into #5654

PR #5656, opened 2022-07-26
Comment out the problematic test case for CBS < 7.1 (MB-50619)

bbrks:
Due to MB-50619, this particular test case fails when running on servers before 7.1.0 - which we rolled back to.

- [x] `^Test(Role)?AccessQuery$ gsi=true xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/485/


markspolakovs (approved):
LGTM - the names would fail `base.IsValidPrincipalName` so shouldn't make it to the access / role access queries anyway

PR #5658, opened 2022-07-28
CBG-2182 make DCP aware of collections

torcolvin:
CBG-2182

I copied the implementation to find collection IDs from cbgt - this is actually more or less shared code between these repos, but I've duplicated it here.

This code is hit by `TestCollectionsDCP` but I haven't made any effort to test `gocbcore` code and I'm not sure if it's worthwhile.

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/595


base/collection_view.go
```diff
@@ -379,3 +380,9 @@ func asViewConsistency(value interface{}) gocb.ViewScanConsistency {
 	}
 
 }
+
+// GetGoCBAgent returns the underlying agent from gocbcore
+func (c *Collection) GetGoCBAgent() (*gocbcore.Agent, error) {
```
adamcfraser: I think it would be preferable to put the functionality that uses the agent on the Collection API (i.e. add getCollectionID here), to avoid leaking dependencies on the underlying gocb implementation.

base/dcp_client.go
```diff
@@ -329,15 +336,38 @@ func (dc *DCPClient) rollback(vbID uint16) (err error) {
 	return nil
 }
 
+func (dc *DCPClient) getOpenStreamOptions() (*gocbcore.OpenStreamOptions, error) {
```
adamcfraser: I don't think we expect to change the set of collection IDs for a DCP client post-initialization.  Given that, can we just accept the set of collection IDs as an input parameter?  Or at least, compute them in NewDCPClient, instead of adding the bucket reference to DCPClient?  That would let us avoid problems like the bucket getting invalidated underneath the client.

base/gocb_utils.go
```diff
@@ -153,3 +158,52 @@ func getRootCAs(caCertPath string) (*x509.CertPool, error) {
...
+		if op != nil {
+			op.Cancel()
+			// wait for confirmation after canceling the PendingOp
+			<-signal
+		}
+		return gocbcore.ErrTimeout
+	}
+}
+
+func getCollectionIDs(scope, collection string, agent *gocbcore.Agent) ([]uint32, error) {
```
adamcfraser: I think you could design this now to work against a slice of collections, even if we're only invoking it for a single collection today.

base/gocb_utils.go
```diff
@@ -153,3 +158,52 @@ func getRootCAs(caCertPath string) (*x509.CertPool, error) {
 	}
 	return rootCAs, nil
 }
+
+// waitForGoCB response waits for a response from a gocb.Agent
+func waitForGoCBResponse(signal <-chan error, op gocbcore.PendingOp, timeout time.Duration) error {
```
adamcfraser: We've got some existing code (like Collection.GetStatsVbSeqno) that does the gocb callback handling in what I think is a bit simpler/cleaner implementation.  Have a look and let me know if you think that approach can be followed here.

adamcfraser (changes requested):

base/collection_view.go
torcolvin: Moved code here, but it'll get moved to a database level call when we run on multiple collections.

base/dcp_client.go
torcolvin: Moved to collections API.

base/gocb_utils.go
torcolvin: This is unfortunately not possible due to GOCBC-1332.

base/collection_view.go
```diff
@@ -379,3 +381,62 @@ func asViewConsistency(value interface{}) gocb.ViewScanConsistency {
...
+
+	err = waitForGocbPendingOp(finished, pendingOp, 5*time.Minute)
+	if err != nil {
+		return 0, fmt.Errorf("GetCollectionID for %s.%s, err: %w", scope, collection, err)
+	}
+	return collectionID, nil
+}
+
+// this is not needed if gocb query supports deadline https://issues.couchbase.com/browse/GOCBC-1332
+func waitForGocbPendingOp(signal <-chan error, op gocbcore.PendingOp, timeout time.Duration) error {
```
adamcfraser: It looks like the fix for GOCBC-1332 is imminent, I'd suggest removing this handling from this PR and filing a ticket to uptake GetCollectionID deadline when it's available.

base/collection.go
```diff
@@ -31,6 +31,11 @@ var ErrCollectionsUnsupported = errors.New("collections not supported")
 var _ sgbucket.KVStore = &Collection{}
 var _ CouchbaseStore = &Collection{}
 
+type Keyspace struct {
```
adamcfraser: Keyspace is probably more correct as bucketname/scope name/collection name.  It doesn't look like this is used in this PR, maybe remove until it's used, so that there's a better context to decide the best structure?

base/collection_view.go
```diff
@@ -379,3 +381,62 @@ func asViewConsistency(value interface{}) gocb.ViewScanConsistency {
...
 }
+
+// GetGoCBAgent returns the underlying agent from gocbcore
+func (c *Collection) getGoCBAgent() (*gocbcore.Agent, error) {
+	return c.Bucket().Internal().IORouter()
+
+}
+
+// getCollectionID returns the gocbcore CollectionID for the current collection
+func (c *Collection) getCollectionID() (uint32, error) {
```
adamcfraser: Is it correct that collectionID is immutable on server?  If so it might make sense to store the collectionID on Collection after we retrieve it, in case we have future use cases where we end up calling getCollection more than once for a Collection instance.

base/gocb_dcp_feed.go
```diff
@@ -45,6 +45,22 @@ func getHighSeqMetadata(bucket Bucket) ([]DCPMetadata, error) {
 	return metadata, nil
 }
 
+func getCollectionIDs(bucket Bucket, scope, collection *string) ([]uint32, error) {
+	c, ok := bucket.(*Collection)
+	if !ok {
```
adamcfraser: It looks like getCollectionIDs (and by extension, StartGocbDCPFeed) only work if bucket is actually a *Collection.  Is that correct?  If so, should we change the bucket parameter on StartGocbDCPFeed to be *Collection, and have callers responsible for casting as needed?

base/gocb_dcp_feed.go
```diff
@@ -45,6 +45,22 @@ func getHighSeqMetadata(bucket Bucket) ([]DCPMetadata, error) {
...
+		return []uint32{}, fmt.Errorf("bucket is not a collection")
+	}
+	var collectionIDs []uint32
+	if scope != nil && collection != nil {
```
adamcfraser: Similarly, I'd probably move this check up to StartGocbDCPFeed, and just not call this function in the case where we're targeting the default scope + collection, since otherwise we don't really need these parameters here.

adamcfraser (changes requested):

base/collection_view.go
torcolvin: I believe this is immutable, but @bbrks should confirm.

base/collection.go
torcolvin: This is now unused, removed.

base/collection_view.go
torcolvin: It is immutable, especially for a connection, so cached the value.

base/collection_view.go
torcolvin: Updated gocbcore/gocb to get this fix.

base/gocb_dcp_feed.go
torcolvin: Migrated a bunch of code to Collection.

base/gocb_dcp_feed.go
torcolvin: Moved this code up, which makes sense because we have to preserve server < 7 which don't support collections at all.

base/collection.go
```diff
@@ -546,14 +547,14 @@ func (c *Collection) Dump() {
...
-	agent, agentErr := c.Bucket().Internal().IORouter()
+	agent, agentErr := c.getGoCBAgent()
 	if agentErr != nil {
 		return nil, nil, agentErr
 	}
 
 	statsOptions := gocbcore.StatsOptions{
 		Key:      "vbucket-seqno",
-		Deadline: time.Now().Add(5 * time.Minute),
+		Deadline: getDCPGoCBDeadline(),
```
adamcfraser: Should this be using a deadline that's different than getBucketOpDeadline?  It seems functionally equivalent, and might be preferable to have it hooked into the existing configurability.

base/collection.go
```diff
@@ -985,3 +986,71 @@ func (t *SGRawTranscoder) Encode(value interface{}) ([]byte, uint32, error) {
...
+	if configOpTimeout != nil {
+		opTimeout = *configOpTimeout
+	}
+	return time.Now().Add(opTimeout)
+}
+
+// getCollectionID returns the gocbcore CollectionID for the current collection
+func (c *Collection) getCollectionID() (uint32, error) {
+	// return cached value if present
+	if c.collectionID != nil {
```
adamcfraser: Do we have any scenarios where we'd have concurrent invocation of getCollectionID that would merit using atomics to read/write c.collectionID?

base/collection.go
```diff
@@ -985,3 +986,71 @@ func (t *SGRawTranscoder) Encode(value interface{}) ([]byte, uint32, error) {
...
+			callbackErr = fmt.Errorf("getCollectionID not retrieved for %s.%s", scope, collection)
+			return
+		}
+
+		collectionID = res.CollectionID
+	}
+	_, err = agent.GetCollectionID(scope,
+		collection,
+		gocbcore.GetCollectionIDOptions{
+			Deadline: getDCPGoCBDeadline(),
```
adamcfraser: Same question as above - seems like getBucketOpDeadline might make sense here?

base/dcp_client_metadata.go
```diff
@@ -255,9 +255,9 @@ func (m *DCPMetadataCS) Persist(workerID int, vbIDs []uint16) {
 
 func (m *DCPMetadataCS) load(workerID int) {
 	var meta WorkerMetadata
-	_, err := m.bucket.Get(m.getMetadataKey(workerID), &meta)
+	_, err := m.collection.Get(m.getMetadataKey(workerID), &meta)
 	if err != nil {
-		if IsKeyNotFoundError(m.bucket, err) {
+		if IsDocNotFoundError(err) {
```
adamcfraser: I don't think the behaviour of IsKeyNotFoundError and IsDocNotFoundError are equivalent.  Is this an intentional switch to use IsDocNotFoundError?

base/dcp_client_metadata.go
```diff
@@ -159,19 +159,19 @@ func BuildDCPMetadataSliceFromVBUUIDs(vbUUIDS []uint64) []DCPMetadata {
 // DCPMetadataCS stores DCP metadata in the specified CouchbaseStore.  It does not require that the store is the
 // same one being streamed over DCP.
 type DCPMetadataCS struct {
-	bucket    Bucket
-	keyPrefix string
-	metadata  []DCPMetadata
-	endSeqNos []gocbcore.SeqNo
+	collection *Collection
```
adamcfraser: Does DCPMetadataCS actually require a *Collection?  I don't see anywhere we were previously casting to *Collection in the implementation.  If it actually just requires Bucket, that makes it more easily unit testable (walrus).

adamcfraser (changes requested):

base/collection.go
torcolvin: Talked this over with @bbrks and the idea is that if something can't start up a DCP stream, we should give it a longer timeout but use the user configurable timeout for document gets.

base/dcp_client_metadata.go
torcolvin: This is another question for @bbrks I suppose - `IsKeyNotFoundError` is not implemented for collections.

base/dcp_client_metadata.go
torcolvin: It puts data into a couchbase store, so I suppose we can leave it as a bucket. It's not tested under walrus though otherwise this code would fail.

base/dcp_client_metadata.go
torcolvin: When using a Bucket, I keep this behavior.

base/collection.go
torcolvin: Decided to use standard bucket op timeout.

base/collection.go
torcolvin: We definitely could so I made this an Atomic.Value

base/collection.go
torcolvin: Done.

base/collection.go
```diff
@@ -182,6 +183,7 @@ type Collection struct {
 	cluster          *gocb.Cluster // Associated cluster - required for N1QL operations
 	queryOps         chan struct{} // Manages max concurrent query ops
 	kvOps            chan struct{} // Manages max concurrent kv ops
+	collectionID     atomic.Value  // cached copy of collectionID
```
adamcfraser, line 186: I think this can just be *uint32, and use atomic.LoadUint32 and atomic.StoreUint32 for access.  The underlying implementation is faster, I think, and it also removes the type casting you're doing elsewhere.

db/import_listener.go
```diff
@@ -85,7 +85,11 @@ func (il *importListener) StartImportFeed(bucket base.Bucket, dbStats *base.DbSt
 	}
 	if !base.IsEnterpriseEdition() {
 		groupID := ""
-		return base.StartGocbDCPFeed(bucket, bucket.GetName(), feedArgs, il.ProcessFeedEvent, importFeedStatsMap.Map, base.DCPMetadataStoreCS, groupID)
+		collection, ok := bucket.(*base.Collection)
```
adamcfraser: Should this also be AsCollection()?

adamcfraser (changes requested):

base/collection.go
torcolvin: We decided to drop this - we don't want to use `sync.Once` since this operation might fail in the case of a network outage and we want to be able to retry getting the collection ID. A maybe more idiomatic API is to make this a struct with a mutex, but not going to pursue it for this code review.

The reason to use atomic.Value is that we need to check if `collectionID` is `nil` before you can call `atomic.LoadUint32` - and checking for nil obviates the use of atomic to prevent the race detection.

db/import_listener.go
torcolvin: Fixed to use AsCollection function.

adamcfraser (approved):

PR #5662, opened 2022-07-29
CBG-2226: PurgeInterval refreshed with at tombstone compaction time

IsaacLambat:
CBG-2226

- Refresh tombstone purge interval with the server metadata purge interval when compaction is started. 
- The purge interval will not be updated if it is 0 in the db which is only applicable in test cases
- Use existing purge interval if SG fails to retrieve server metadata purge interval
- Added testing with helper function

## Pre-review checklist

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/505/
- [x] `xattrs=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/511/ `known failures`

### DIFF
--- a/rest/admin_api_test.go
+++ b/rest/admin_api_test.go
@@ -4883,3 +4883,78 @@ func TestPublicChanGuestAccess(t *testing.T) {
 	resp = rt.SendRequest(http.MethodGet, "/db/docNoAccess", "")
 	requireStatus(t, resp, http.StatusForbidden)
 }
+
+func setServerPurgeInterval(t *testing.T, rt *RestTester, newPurgeInterval string) {
+	eps, httpClient, err := rt.ServerContext().ObtainManagementEndpointsAndHTTPClient()
+	require.NoError(t, err)
+
+	settings := url.Values{}
+	settings.Add("purgeInterval", newPurgeInterval)
+	settings.Add("autoCompactionDefined", "true")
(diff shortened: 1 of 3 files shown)


rest/admin_api_test.go
```diff
@@ -4883,3 +4883,78 @@ func TestPublicChanGuestAccess(t *testing.T) {
...
+			dbc.PurgeInterval = test.dbPurgeInterval
+			setServerPurgeInterval(t, rt, test.newServerInterval)
+
+			// Start compact to modify purge interval
+			database, _ := db.GetDatabase(dbc, nil)
+			_, err = database.Compact(false, func(purgedDocCount *int) {}, base.NewSafeTerminator())
+			require.NoError(t, err)
+
+			// Check purge interval is as expected
+			assert.EqualValues(t, test.expectedPurgeIntervalAfterCompact, dbc.PurgeInterval)
```
bbrks: This assertion only works when xattrs=true because Compact noops/exits early in the case of xattrs=false

Either:
- Skip the test entirely with xattrs=false
- Ensure this value doesn't change with xattrs=false

bbrks (changes requested):
Fix test for xattrs=false (and kick off an integration run with that set)

bbrks (approved):

PR #5664, opened 2022-08-01
NewDatabaseContext: consistently use logging context

markspolakovs:
Replace `context.TODO()` in `change_cache.go` with a log context with a correlation ID of the database name. This ensures that log lines for different DBs can be differentiated.

Before:
```
2022-08-01T14:21:01.222+01:00 [DBG] DCP+: Received #8 after   2ms ("<ud>cornflakes</ud>" / "1-6c09ee187cece3ab2da390476ac6b48f")
```

After:
```
2022-08-01T14:16:51.192+01:00 [DBG] DCP+: c:db:testificate Received #6 after   1ms ("<ud>porcupine</ud>" / "1-6c09ee187cece3ab2da390476ac6b48f")
```

## Pre-review checklist
- [n/a] Removed debug logging (`fmt.Print`, `log.Print`, ...)
- [n/a] Updated relevant information in the API specifications (such as endpoint descriptions, schemas, ...) in `docs/api`

N/A (logging-only change)

### DIFF
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -157,11 +157,9 @@ func DefaultCacheOptions() CacheOptions {
 // notifyChange is an optional function that will be called to notify of channel changes.
 // After calling Init(), you must call .Start() to start useing the cache, otherwise it will be in a locked state
 // and callers will block on trying to obtain the lock.
-func (c *changeCache) Init(dbcontext *DatabaseContext, notifyChange func(base.Set), options *CacheOptions) error {
+func (c *changeCache) Init(logCtx context.Context, dbcontext *DatabaseContext, notifyChange func(base.Set), options *CacheOptions) error {
 	c.context = dbcontext
-	c.logCtx = context.WithValue(context.Background(), base.LogContextKey{}, base.LogContext{
-		CorrelationID: "changeCache:" + base.MD(dbcontext.Name).Redact(),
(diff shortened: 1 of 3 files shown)


db/change_cache.go
```diff
@@ -158,6 +159,9 @@ func DefaultCacheOptions() CacheOptions {
 // and callers will block on trying to obtain the lock.
 func (c *changeCache) Init(dbcontext *DatabaseContext, notifyChange func(base.Set), options *CacheOptions) error {
```
bbrks: The caller of this (`NewDatabaseContext`) defines `logCtx := context.TODO()` (with a load of places actually using that variable in a reasonable way)

So I'm wondering if we can un-TODO it one layer up then just pass it in as a context parameter and get even more db log context coverage.

bbrks (changes requested):
I'd be ok merging this as is - but I think we can also cover way more by moving the context creation up 1 more level.

db/change_cache.go
markspolakovs: It's slightly against the intended design of the context package:

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it.

but I think it makes sense for the way we're using it.

bbrks (approved):

PR #5669, opened 2022-08-02
Skip LiteCore SyncServer tests because they're segfaulting

bbrks:
Filed CBG-2237 to track re-enabling or getting more information to the litecore team for a fix


IsaacLambat (approved):
LGTM

PR #5673, opened 2022-08-03
CBG-2213: Better handle CBGT feed rolling upgrades

markspolakovs:
CBG-2213

Change the behaviour of the sharded DCP feed to accommodate rolling online upgrades from pre-Helium to Helium nodes.

* Rather than registering a new feed type for the gocbcore-backed feed, reuse the existing feed type - when only streaming the default collection, there is no functional difference. This means that there is no need to delete and re-create the index upon upgrade, which simplifies behaviour.
  * This also means we can simplify the connection string logic, as there is no need to replace `couchbase[s]://` with `http[s]://` for gocbcore (as we used to for cbdatasource).
* Add an integration test for the above behaviour. The upgrade itself will likely need functional testing (and/or CBG-2273).

## Pre-review checklist

- [ ] `server=7.0.3` https://jenkins.sgwdev.com/job/SyncGateway-Integration/530/
- [ ] `server=6.6.5`https://jenkins.sgwdev.com/job/SyncGateway-Integration/531/
- Bucket flush failures on both. Re-ran with only the `db` package in https://jenkins.sgwdev.com/job/SyncGateway-Integration/585/ and https://jenkins.sgwdev.com/job/SyncGateway-Integration/586/ respectively.

### DIFF
--- a/base/dcp_sharded.go
+++ b/base/dcp_sharded.go
@@ -14,7 +14,6 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
-	"math"
 	"strconv"
 	"strings"
 	"sync"
@@ -30,17 +29,18 @@ import (
 const CBGTIndexTypeSyncGatewayImport = "syncGateway-import-"
 const DefaultImportPartitions = 16
 
-const (
-	compatVersionHelium      = 1
-	currentNodeCompatVersion = compatVersionHelium
-)
+// firstVersionToSupportCollections represents the earliest Sync Gateway release that supports collections (codename "Helium").
+var firstVersionToSupportCollections = &ComparableVersion{
+	epoch: 0,
+	major: 3,
+	minor: 1,
+	patch: 0,
+}
 
 // nodeExtras is the contents of the JSON value of the cbgt.NodeDef.Extras field as used by Sync Gateway.
 type nodeExtras struct {
-	// CompatVersion is the node's "compatibility version", an integer that is incremented each time there is a cluster-wide
-	// behaviour change that nodes need to handle differently depending on if older nodes are present. For example, see
-	// CBG-2213.
-	CompatVersion uint `json:"v"`
+	// Version is the node's version.
+	Version *ComparableVersion `json:"v"`
 }
 
 // CbgtContext holds the two handles we have for CBGT-related functionality.
@@ -250,7 +250,7 @@ func getCBGTIndexUUID(manager *cbgt.Manager, indexName string) (exists bool, pre
 func initCBGTManager(bucket Bucket, spec BucketSpec, cfgSG cbgt.Cfg, dbUUID string, dbName string) (*CbgtContext, error) {
 	// Check if there are pre-Helium nodes, and if so, set the LithiumCompat flag to ensure we don't try to start a
 	// collections-enabled feed when some nodes won't support it.
-	minVersion, err := getMinNodeCompatVersion(cfgSG)
+	minVersion, err := getMinNodeVersion(cfgSG)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get minimum node version in cluster: %w", err)
 	}
@@ -282,9 +282,9 @@ func initCBGTManager(bucket Bucket, spec BucketSpec, cfgSG cbgt.Cfg, dbUUID stri
 	// container: Used by cbgt to determine node hierarchy.  Not needed by Sync Gateway
 	container := ""
 
-	// extras: Can be used to pass Sync Gateway specific node information to callbacks. Used to detect node upgrades -
-	// currently pre-Helium to Helium.
-	extras, err := JSONMarshal(&nodeExtras{CompatVersion: currentNodeCompatVersion})
+	// extras: Can be used to pass Sync Gateway specific node information to callbacks. Used to store the node version
+	// from Helium (3.1.0) onwards, empty for older versions.
+	extras, err := JSONMarshal(&nodeExtras{Version: ProductVersion})
 	if err != nil {
 		return nil, err
 	}
@@ -349,7 +349,7 @@ func initCBGTManager(bucket Bucket, spec BucketSpec, cfgSG cbgt.Cfg, dbUUID stri
 		),
 		Manager:       mgr,
 		Cfg:           cfgSG,
-		LithiumCompat: NewAtomicBool(minVersion < compatVersionHelium),
+		LithiumCompat: NewAtomicBool(minVersion.Less(firstVersionToSupportCollections)),
 	}
 	if cbgtContext.LithiumCompat.IsTrue() {
 		WarnfCtx(cbgtContext.loggingCtx, "Found old (3.0.x or below) Sync Gateway nodes in cluster - entering compatibility mode."+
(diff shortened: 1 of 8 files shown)


base/dcp_sharded.go
```diff
@@ -29,13 +30,27 @@ import (
...
+	compatVersionHelium      = 1
+	currentNodeCompatVersion = compatVersionHelium
+)
+
+// nodeExtras is the contents of the JSON value of the cbgt.NodeDef.Extras field as used by Sync Gateway.
+type nodeExtras struct {
+	// CompatVersion is the node's "compatibility version", an integer that is incremented each time there is a cluster-wide
+	// behaviour change that nodes need to handle differently depending on if older nodes are present. For example, see
+	// CBG-2213.
+	CompatVersion uint `json:"v"`
```
adamcfraser: Can we use the Sync Gateway version here to provide the same functionality, while also providing some useful diagnostic information for supportability?

base/dcp_sharded.go
```diff
@@ -29,13 +30,27 @@ import (
...
+}
+
 // CbgtContext holds the two handles we have for CBGT-related functionality.
 type CbgtContext struct {
 	Manager           *cbgt.Manager            // Manager is main entry point for initialization, registering indexes
 	Cfg               cbgt.Cfg                 // Cfg manages storage of the current pindex set and node assignment
 	heartbeater       Heartbeater              // Heartbeater used for failed node detection
 	heartbeatListener *importHeartbeatListener // Listener subscribed to failed node alerts from heartbeater
 	loggingCtx        context.Context          // Context for cbgt logging
+	LithiumCompat     *AtomicBool              // set on startup if there are pre-Helium nodes in the cluster, later reset once all pre-Helium nodes are gone
```
adamcfraser: I think it's preferable to avoid using release code names in code - after a few releases it's easy to forget the actual SG version it's associated with.  In this case, since this flag just has a single purpose ("supports collections"), can we just name it based on that?

db/dcp_sharded_upgrade_test.go
```diff
@@ -0,0 +1,283 @@
...
+			return true, nil, nil
+		}
+		return db.ImportListener.cbgtContext.LithiumCompat.IsTrue(), nil, nil
+		// will take around ~10 seconds at least
+	}, base.CreateSleeperFunc(200, 10))
+	assert.NoError(t, err, "LithiumCompat never became false")
+
+	//pIndexes, _, err = cbgt.CfgGetPlanPIndexes(db.CfgSG)
+	//require.NoError(t, err)
+	// TODO: sometimes cbgt adds and immediately removes a PIndex causing this to fail
```
adamcfraser: Should just remove these TODOs and asserts if they are unreliable.

db/dcp_sharded_upgrade_test.go
```diff
@@ -0,0 +1,283 @@
...
+	// Write a doc before starting the dbContext to check that import works
+	const (
+		testDoc1 = "testDoc1"
+		testDoc2 = "testDoc2"
+	)
+	require.NoError(t, tb.SetRaw(testDoc1, 0, nil, []byte(`{}`)))
+
+	t.Log("TEST: setup complete")
+
+	// Start this retry loop asynchronously because the change may happen before NewDatabaseContext returns
```
adamcfraser: Is this comment still accurate?  I don't see what retry loop is being run asynchronously here.

db/dcp_sharded_upgrade_test.go
```diff
@@ -0,0 +1,283 @@
...
+	require.NoError(t, tb.SetRaw(base.SyncPrefix+"cfgplanPIndexes", 0, nil, []byte(planPIndexesJSON)))
+
+	// Write a doc before starting the dbContext to check that import works
+	const (
+		testDoc1 = "testDoc1"
+		testDoc2 = "testDoc2"
+	)
+	require.NoError(t, tb.SetRaw(testDoc1, 0, nil, []byte(`{}`)))
+
+	t.Log("TEST: setup complete")
```
adamcfraser: This seems like dev-time logging that could be removed.

db/dcp_sharded_upgrade_test.go
```diff
@@ -0,0 +1,283 @@
...
+	}, base.CreateSleeperFunc(200, 10))
+	assert.NoError(t, err, "LithiumCompat never became false")
+
+	//pIndexes, _, err = cbgt.CfgGetPlanPIndexes(db.CfgSG)
+	//require.NoError(t, err)
+	// TODO: sometimes cbgt adds and immediately removes a PIndex causing this to fail
+	//assert.Len(t, pIndexes.PlanPIndexes, numPartitions, "number of planPIndexes should match partitions after LithiumCompat change")
+
+	// And write another doc to check that import still works
+	require.NoError(t, tb.SetRaw(testDoc2, 0, nil, []byte(`{}`)))
```
adamcfraser, line 256: Are we sure that this isn't just a doc that's written to a pindex that happens to be assigned to this node?

base/dcp_sharded.go
```diff
@@ -46,11 +61,11 @@ func StartShardedDCPFeed(dbName string, configGroup string, uuid string, heartbe
 		return nil, err
 	}
 
-	dcpContextID := MD(spec.BucketName).Redact() + "-" + DCPImportFeedID
-	cbgtContext.loggingCtx = context.WithValue(context.Background(), LogContextKey{},
-		LogContext{CorrelationID: dcpContextID},
-	)
-
+	if cbgtContext.LithiumCompat.IsTrue() {
+		if spec.Scope != nil || spec.Collection != nil {
```
adamcfraser: I don't think StartShardedDCPFeed is running in any kind of retry loop, is it?  Given that, I don't see the value in persisting the LithiumCompat to the cbgtContext, if we're only ever going to error out right away - could just return as an error from initCBGTManager?

base/dcp_sharded.go
```diff
@@ -536,6 +592,12 @@ func (l *importHeartbeatListener) subscribeNodeChanges() error {
 					}
 				}
 
+				if l.ctx.LithiumCompat.IsTrue() && minCompatVersion >= compatVersionHelium {
```
adamcfraser: Similar question to the above - I don't think we're even registering the importHeartbeatListener if LithiumCompat.IsTrue(), so does this code get used outside of the unit test?

adamcfraser (changes requested):
Some minor naming questions, but I'm also unclear on whether the monitoring of other nodes has value if we're failing to start the database in non-compatibility scenarios.

db/dcp_sharded_upgrade_test.go
markspolakovs: Yes - initially all the PIndexes are assigned to a node that doesn't exist (unless the node under test generates the UUID `680cce1d31e72b69`, which has a chance of 1 in `16*(2^16)`)

base/dcp_sharded.go
markspolakovs: That's a fair point - in theory we can do the `getMinNodeCompatVersion` check before starting the manager as it can be done with just a `cbgt.Cfg`. In theory there's the risk that this becomes a TOCTOU (i.e. a node that doesn't support collections joins the cluster in between the `getMinNodeCompatVersion` check and the manager startup), but IMHO this is a rare enough case that it's not worth worrying about.

base/dcp_sharded.go
markspolakovs: In theory yes, we can use `ComparableVersion` to check against the product version - we can assume that `3.0.x` is the last version to _not_ support collections, so we can use a Less check against `3.1.0` (even if Helium becomes eg `3.5.0` that'll still pass). Will update.

markspolakovs:
@markspolakovs to manually test as well

markspolakovs:
Verified manually:

1. Started a SG 3.0.0
2. Ran a continuous write load against the bucket, and verified that the documents were getting imported
3. Started a SG built on this branch, verified that both were continuously importing (the new SG got 8 partitions assigned)
4. Killed the old SG, verified that after ~30 seconds the new SG took over its partitions and continued import without starting from zero.

Adapted the unit test to check this as well, though testing the exact sequence handling is difficult. Assigning to @adamcfraser for final code review.

db/dcp_sharded_upgrade_test.go
adamcfraser: My question was more "are we sure that all PIndexes are getting reassigned to the new node"?  If only half are being assigned, and this doc happens to be in a vbucket in that half, the test would still pass.

adamcfraser (commented):
New changes look good - just a final (minor) question on whether the test is validating PIndex assignment comprehensively.

db/dcp_sharded_upgrade_test.go
markspolakovs: Oh, I see what you mean - added an extra check that all PIndexes are assigned to this node.

markspolakovs:
Windows CI issue is #5695, rebased to pick up fix.

adamcfraser (dismissed):

adamcfraser:
@markspolakovs Changes look good, but it looks like the CI ee-unit-tests are failing on the new TestShardedDCPUpgrade test, waiting stale nodes to be cleaned up.  Have you already looked at that?

markspolakovs:
Test failure only occurred against Walrus - skipped it there.

PR #5675, opened 2022-08-04
CBG-2243: Upgrade Go to 1.18

markspolakovs:
CBG-2243

Go 1.17 is now EOL, so switch to building with the latest 1.18 release.

While we're at it, replace some utility functions in the `base` package with generic implementations (kept in separate commits, so if we decide we don't want them we can easily revert just those).

## Pre-review checklist

- [x] https://jenkins.sgwdev.com/job/SyncGateway-Integration/570/

**Note:** do not merge until 2022-08-08, to ensure we can get a perf test to compare 1.17/1.18


base/util.go
```diff
@@ -1687,10 +1681,7 @@ func TerminateAndWaitForClose(terminator chan struct{}, done chan struct{}, time
 	return nil
 }
 
-// TODO: replace these with a generic implementation once we upgrade to Go 1.18
-
-// CoalesceStrings returns the first non-nil argument, or nil if both are nil.
-func CoalesceStrings(a, b *string) *string {
+func Coalesce[T any](a, b *T) *T {
```
IsaacLambat, line 1685:
```suggestion
// Coalesce returns the first non-nil argument, or nil if both are nil.
func Coalesce[T any](a, b *T) *T {
```

IsaacLambat (changes requested):
Minor change and then ready to go after test passes

IsaacLambat (dismissed):
LGTM. May need to get it benchmarked before merging

IsaacLambat (approved):
LGTM

PR #5677, opened 2022-08-04
CBG-2187: do not list deleted roles

cbbruno:
CBG-2187

Describe your PR here...
- Change GET roles endpoint to not return roles marked as deleted by default.
- Add optional parameter to the endpoint to include deleted roles if needed.

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/582/


db/query.go
```diff
@@ -170,6 +171,28 @@ var QueryPrincipals = SGQuery{
...
+			"WHERE META(%s).id LIKE '%s' "+
+			"AND (META(%s).id LIKE '%s' "+
+			"OR META(%s).id LIKE '%s') "+
+			"AND (%s.deleted IS MISSING OR %s.deleted = false) "+
+			"AND META(%s).id >= $%s "+ // Uses >= for inclusive startKey
+			"ORDER BY META(%s).id",
+		base.KeyspaceQueryAlias,
+		base.KeyspaceQueryToken, base.KeyspaceQueryAlias,
+		base.KeyspaceQueryAlias, SyncDocWildcard,
+		base.KeyspaceQueryAlias, `\\_sync:user:%`,
```
adamcfraser: Given that our only use case is for role retrieval, and we're having to build a new query anyway, I'd suggest omitting users from this query.  Deployments with a very large number of users and a very small number of roles are common, and it would be nice to avoid the query fetch on each user, since this isn't a covering index.

adamcfraser (changes requested):

db/query.go
cbbruno: the usage below seems to require both, users and roles... but I guess I could revert and leave `AllPrincipalIDs` and `QueryPrincipals` as they are (retrieve all), and create new functions just for active roles, something like `GetActiveRoleIDs`, that is only called by the API ?

https://github.com/couchbase/sync_gateway/blob/b630ddf8be5714cb357aa79aed96dfa9678e726c/db/database.go#L1547

(and maybe here)
https://github.com/couchbase/sync_gateway/blob/b630ddf8be5714cb357aa79aed96dfa9678e726c/db/database.go#L1498

also, looking back at the usage above, think we should be able to call `AllPrincipalIDs` only once, `if docsChanged > 0 || regenerateSequences {...}` .

adamcfraser (commented):
Changes look good.  Have you run tests under GSI?  The linked integration test was run with views.

cbbruno:
> Details

I did not... I only run the tests I changed with GSI locally ... I tried now but it seems there are other issues with our jenkins.

adamcfraser (approved):

PR #5679, opened 2022-08-05
CBG-2201: Added LeakyBucket support to the rest tester under integration tests

IsaacLambat:
CBG-2201

- Added LeakyBucket support to the Rest Tester for running with integration tests when using config option
- Added collection support to the rest tester when using leaky buckets
- A package will now panic if a bucket is left unclosed instead of just logging a warning

## Pre-review checklist

- [ ] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/675/

### DIFF
--- a/base/main_test_bucket_pool.go
+++ b/base/main_test_bucket_pool.go
@@ -400,10 +400,15 @@ func (tbp *TestBucketPool) printStats() {
 	tbp.Logf(ctx, "==========================")
 
 	tbp.unclosedBucketsLock.Lock()
+	unclosedBucketWarnings := ""
 	for testName, buckets := range tbp.unclosedBuckets {
 		for bucketName := range buckets {
 			tbp.Logf(ctx, "WARNING: %s left %s bucket unclosed!", testName, bucketName)
+			unclosedBucketWarnings += fmt.Sprintf("%s left %s bucket unclosed!\n", testName, bucketName)
 		}
+	}
+	if unclosedBucketWarnings != "" {
+		panic(unclosedBucketWarnings)
 	}
 	tbp.unclosedBucketsLock.Unlock()
 
--- a/base/util_testing.go
+++ b/base/util_testing.go
@@ -64,6 +64,25 @@ func (tb *TestBucket) GetUnderlyingBucket() Bucket {
 	return tb.Bucket
 }
 
+// AsCollectionStore gets the underlying collection store from a bucket for test purposes
+func AsCollectionStore(t testing.TB, bucket Bucket) (*Collection, bool) {
+	var underlyingBucket Bucket
+	switch typedBucket := bucket.(type) {
+	case *Collection:
+		return bucket.(*Collection), true
+	case *LeakyBucket:
+		underlyingBucket = typedBucket.GetUnderlyingBucket()
+	case *LoggingBucket:
+		underlyingBucket = typedBucket.GetUnderlyingBucket()
+	case *TestBucket:
+		underlyingBucket = typedBucket.GetUnderlyingBucket()
+	default:
+		// bail out for unrecognised/unsupported buckets
+		return nil, false
+	}
+	return AsCollectionStore(t, underlyingBucket)
+}
+
 // LeakyBucketClone wraps the underlying bucket on the TestBucket with a LeakyBucket and returns a new TestBucket handle.
 func (tb *TestBucket) LeakyBucketClone(c LeakyBucketConfig) *TestBucket {
 	return &TestBucket{
--- a/base/util_testing.go
+++ b/base/util_testing.go
@@ -64,25 +64,6 @@ func (tb *TestBucket) GetUnderlyingBucket() Bucket {
 	return tb.Bucket
 }
 
-// AsCollectionStore gets the underlying collection store from a bucket for test purposes
-func AsCollectionStore(t testing.TB, bucket Bucket) (*Collection, bool) {
-	var underlyingBucket Bucket
-	switch typedBucket := bucket.(type) {
-	case *Collection:
-		return bucket.(*Collection), true
-	case *LeakyBucket:
-		underlyingBucket = typedBucket.GetUnderlyingBucket()
-	case *LoggingBucket:
-		underlyingBucket = typedBucket.GetUnderlyingBucket()
-	case *TestBucket:
-		underlyingBucket = typedBucket.GetUnderlyingBucket()
-	default:
-		// bail out for unrecognised/unsupported buckets
-		return nil, false
-	}
-	return AsCollectionStore(t, underlyingBucket)
-}
-
 // LeakyBucketClone wraps the underlying bucket on the TestBucket with a LeakyBucket and returns a new TestBucket handle.
 func (tb *TestBucket) LeakyBucketClone(c LeakyBucketConfig) *TestBucket {
 	return &TestBucket{
--- a/db/database.go
+++ b/db/database.go
@@ -261,15 +261,15 @@ func connectToBucketErrorHandling(spec base.BucketSpec, gotErr error) (fatalErro
 
 type OpenBucketFn func(spec base.BucketSpec) (base.Bucket, error)
 
-// ConnectToBucketFailFast opens a Couchbase connect and return a specific bucket without retrying on failure.
-func ConnectToBucketFailFast(spec base.BucketSpec) (bucket base.Bucket, err error) {
+// connectToBucketFailFast opens a Couchbase connect and return a specific bucket without retrying on failure.
+func connectToBucketFailFast(spec base.BucketSpec) (bucket base.Bucket, err error) {
 	bucket, err = base.GetBucket(spec)
 	_, err = connectToBucketErrorHandling(spec, err)
 	return bucket, err
 }
 
-// ConnectToBucket opens a Couchbase connection and return a specific bucket.
-func ConnectToBucket(spec base.BucketSpec) (base.Bucket, error) {
+// connectToBucket opens a Couchbase connection and return a specific bucket.
+func connectToBucket(spec base.BucketSpec) (base.Bucket, error) {
 
 	// start a retry loop to connect to the bucket backing off double the delay each time
 	worker := func() (bool, error, interface{}) {
@@ -299,9 +299,9 @@ func ConnectToBucket(spec base.BucketSpec) (base.Bucket, error) {
 // GetConnectToBucketFn returns a different OpenBucketFn to connect to the bucket depending on the value of failFast
 func GetConnectToBucketFn(failFast bool) OpenBucketFn {
 	if failFast {
-		return ConnectToBucketFailFast
+		return connectToBucketFailFast
 	}
-	return ConnectToBucket
+	return connectToBucket
 }
 
 // Function type for something that calls NewDatabaseContext and wants a callback when the DB is detected
--- a/db/database.go
+++ b/db/database.go
@@ -259,6 +259,8 @@ func connectToBucketErrorHandling(spec base.BucketSpec, gotErr error) (fatalErro
 	return false, nil
 }
 
+type OpenBucketFn func(spec base.BucketSpec) (base.Bucket, error)
+
 // ConnectToBucketFailFast opens a Couchbase connect and return a specific bucket without retrying on failure.
 func ConnectToBucketFailFast(spec base.BucketSpec) (bucket base.Bucket, err error) {
 	bucket, err = base.GetBucket(spec)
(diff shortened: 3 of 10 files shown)


rest/server_context.go
```diff
@@ -367,7 +367,8 @@ func GetBucketSpec(config *DatabaseConfig, serverConfig *StartupConfig) (spec ba
 // Adds a database to the ServerContext.  Attempts a read after it gets the write
 // lock to see if it's already been added by another process. If so, returns either the
 // existing DatabaseContext or an error based on the useExisting flag.
-func (sc *ServerContext) _getOrAddDatabaseFromConfig(config DatabaseConfig, useExisting, failFast bool) (*db.DatabaseContext, error) {
+// Pass in a connectFn to replace the default ConnectToBucket function. This will cause the failFast argument to be ignored
```
bbrks: Maybe rename this function for clarity - `connectFn` implies you need to actually set up a connection.

```suggestion
// Pass in a bucketFromBucketSpecFn to replace the default ConnectToBucket function. This will cause the failFast argument to be ignored
```

rest/utilities_testing.go
```diff
@@ -228,7 +228,24 @@ func (rt *RestTester) Bucket() base.Bucket {
...
+		if rt.useLeakyTBForDB {
+			// If using a leaky bucket, make sure it ignores attempted closures to avoid double closures panic
+			// due to closing both RestTesterServerContext and testBucket in rt.Close()
+			leakyBucket, isLeaky := base.AsLeakyBucket(testBucket)
+			if isLeaky {
+				leakyBucket.SetIgnoreClose(true)
+			} else {
+				panic("Using option useLeakyTBForDB but testBucket is not a leaky bucket")
+			}
+
```
bbrks: We can simplify to make the if only deal with the error case

```suggestion
			if !isLeaky {
				panic("Using option useLeakyTBForDB but testBucket is not a leaky bucket")
			}

			leakyBucket.SetIgnoreClose(true)
```

rest/utilities_testing.go
```diff
@@ -228,7 +228,24 @@ func (rt *RestTester) Bucket() base.Bucket {
...
 
-		_, err = rt.RestTesterServerContext.AddDatabaseFromConfig(*rt.DatabaseConfig)
+		if rt.useLeakyTBForDB {
+			// If using a leaky bucket, make sure it ignores attempted closures to avoid double closures panic
+			// due to closing both RestTesterServerContext and testBucket in rt.Close()
+			leakyBucket, isLeaky := base.AsLeakyBucket(testBucket)
+			if isLeaky {
+				leakyBucket.SetIgnoreClose(true)
+			} else {
+				panic("Using option useLeakyTBForDB but testBucket is not a leaky bucket")
```
bbrks: Some other areas of this function use `rt.tb.Fatalf` to abort the affected test, instead of failing the entire test suite - which seems preferable.

rest/utilities_testing.go
```diff
@@ -62,6 +62,7 @@ type RestTesterConfig struct {
 	persistentConfig                bool
 	groupID                         *string
 	createScopesAndCollections      bool // If true, will automatically create any defined scopes and collections on startup.
+	useLeakyTBForDB                 bool // If true, the leaky test bucket will be used on the database. Useful for using a LeakyBucket on the DB context for integration tests.
```
bbrks: Can this be inferred if `RestTesterConfig.TestBucket` is non-nil? (is there a situation where you _wouldn't_ want to use the leaky test bucket if you were passing one in?)

bbrks (changes requested):

rest/utilities_testing.go
IsaacLambat: Did try that way at first, however some tests actually rely on the DB generating the bucket and therefore failing. For example, `TestCollectionsBasicIndexQuery` which relies on the index not being created on the default collection however it does get created on the default collection when TB is passed in to the database context.

rest/server_context.go
IsaacLambat: Changed `AddDatabaseFromConfigWithConnectFn` to `AddDatabaseFromConfigWithBucket` and made it take a bucket now instead then pass that down using a `bucketFromBucketSpecFn` function.

rest/server_context.go
```diff
@@ -72,7 +72,7 @@ type bootstrapContext struct {
 func (sc *ServerContext) CreateLocalDatabase(dbs DbConfigMap) error {
 	for _, dbConfig := range dbs {
 		dbc := dbConfig.ToDatabaseConfig()
-		_, err := sc._getOrAddDatabaseFromConfig(*dbc, false, false)
+		_, err := sc._getOrAddDatabaseFromConfig(*dbc, false, db.ConnectToBucket)
```
bbrks: Might be nicer to continue using the `db.GetConnectToBucketFn` helper but just with a static `false` parameter - fewer places to change if `ConnectToBucket` needs to change/move.

db/database.go
```diff
@@ -293,6 +295,14 @@ func ConnectToBucket(spec base.BucketSpec) (base.Bucket, error) {
 	return ibucket.(base.Bucket), nil
 }
 
+// GetConnectToBucketFn returns a different OpenBucketFn to connect to the bucket depending on the value of failFast
+func GetConnectToBucketFn(failFast bool) OpenBucketFn {
+	if failFast {
+		return ConnectToBucketFailFast
+	}
+	return ConnectToBucket
+}
```
bbrks: I think we should make these `ConnectToBucket(FailFast)?` functions private so their use is restricted - encourage people to use this `GetConnectToBucketFn`

rest/utilities_testing.go
```diff
@@ -231,7 +230,39 @@ func (rt *RestTester) Bucket() base.Bucket {
...
 		rt.DatabaseConfig.SGReplicateEnabled = base.BoolPtr(rt.RestTesterConfig.sgReplicateEnabled)
 
-		_, err = rt.RestTesterServerContext.AddDatabaseFromConfig(*rt.DatabaseConfig)
+		if rt.RestTesterConfig.TestBucket != nil {
+			// If using a leaky bucket, make sure it ignores attempted closures to avoid double closures panic
+			// due to closing both RestTesterServerContext and testBucket in rt.Close()
+			leakyBucket, isLeaky := base.AsLeakyBucket(testBucket)
+			if isLeaky {
+				leakyBucket.SetIgnoreClose(true)
+			}
```
bbrks: Reduce scope of this to avoid accidental usage of `leakyBucket` even if `isLeaky` is false

(this suggestion makes `leakyBucket` only accessible inside the if block)

```suggestion
			if leakyBucket, isLeaky := base.AsLeakyBucket(testBucket); isLeaky {
				leakyBucket.SetIgnoreClose(true)
			}
```

base/util_testing.go
```diff
@@ -64,6 +64,25 @@ func (tb *TestBucket) GetUnderlyingBucket() Bucket {
 	return tb.Bucket
 }
 
+// AsCollectionStore gets the underlying collection store from a bucket for test purposes
+func AsCollectionStore(t testing.TB, bucket Bucket) (*Collection, bool) {
```
bbrks: `CollectionStore` is not what we're returning from this function (other cases return something implementing a *Store interface)

I don't think we need this to be test-specific either - @torcolvin had mentioned needing to write this function.

```suggestion
// AsCollection gets the underlying Collection from a bucket
func AsCollection(bucket Bucket) (*Collection, bool) {
```

rest/utilities_testing.go
```diff
@@ -231,7 +230,39 @@ func (rt *RestTester) Bucket() base.Bucket {
...
+
+			// Set scopes and collections on bucket if set in DB Config
```
bbrks: The comment from `MakeBucketSpec` is clearer about this being a temporary thing

```suggestion
			// WIP: Collections Phase 1 - Grab just one scope/collection from the defined set.
			// Phase 2 (multi collection) means DatabaseContext needs a set of BucketSpec/Collections, not just one...
```

rest/utilities_testing.go
```diff
@@ -231,7 +230,39 @@ func (rt *RestTester) Bucket() base.Bucket {
...
+				if !isCollection {
+					rt.tb.Fatalf("Could not get collection from bucket with type %T: %v", testBucket.Bucket, err)
+				}
+
+				collectionBucket.Spec.Scope = scope
+				collectionBucket.Spec.Collection = collection
+				collectionBucket.Collection = collectionBucket.Collection.Bucket().Scope(*scope).Collection(*collection)
+			}
+
+			_, err = rt.RestTesterServerContext.AddDatabaseFromConfigWithBucket(rt.tb, *rt.DatabaseConfig, testBucket.Bucket)
```
bbrks, line 266: should this be `collectionBucket` based on setting spec/collection above? Or doesn't it matter?

```suggestion
			_, err = rt.RestTesterServerContext.AddDatabaseFromConfigWithBucket(rt.tb, *rt.DatabaseConfig, collectionBucket)
```

bbrks (changes requested):

rest/utilities_testing.go
IsaacLambat: This would cause the leaky bucket to be lost and therefore not fire the callbacks

bbrks:
@IsaacLambat 

Can you take a look at the build failure? Probably a break after rebasing or something

```
[2022-08-11T01:08:51.702Z] # github.com/couchbase/sync_gateway/base
[2022-08-11T01:08:51.702Z] base/collection.go:1085:6: AsCollection redeclared in this block
[2022-08-11T01:08:51.702Z] 	base/collection.go:181:6: other declaration of AsCollection
[2022-08-11T01:08:51.702Z] base/collection.go:1100:9: cannot use AsCollection(underlyingBucket) (value of type bool) as type error in return statement:
[2022-08-11T01:08:51.702Z] 	bool does not implement error (missing Error method)
```

base/main_test_bucket_pool.go
```diff
@@ -400,11 +400,16 @@ func (tbp *TestBucketPool) printStats() {
...
 	tbp.unclosedBucketsLock.Lock()
+	unclosedBucketWarnings := ""
 	for testName, buckets := range tbp.unclosedBuckets {
 		for bucketName := range buckets {
 			tbp.Logf(ctx, "WARNING: %s left %s bucket unclosed!", testName, bucketName)
+			unclosedBucketWarnings += fmt.Sprintf("%s left %s bucket unclosed!\n", testName, bucketName)
 		}
 	}
+	if unclosedBucketWarnings != "" {
+		panic(unclosedBucketWarnings)
```
adamcfraser, line 411: I don't recall the background for this change - is this to better catch this scenario pre-merge?

rest/utilities_testing.go
```diff
@@ -232,7 +240,36 @@ func (rt *RestTester) Bucket() base.Bucket {
 
 		rt.DatabaseConfig.SGReplicateEnabled = base.BoolPtr(rt.RestTesterConfig.sgReplicateEnabled)
 
-		_, err = rt.RestTesterServerContext.AddDatabaseFromConfig(*rt.DatabaseConfig)
+		if rt.leakyBucketConfig != nil {
+			// Scopes and collections have to be set on the bucket being passed in for the db to use.
+			if rt.createScopesAndCollections {
```
adamcfraser: I'm not sure why this code is inside a createScopesAndCollections, unless 'createScopesAndCollections' is now equivalent to 'useScopesAndCollections' for the RestTester?

rest/utilities_testing.go
```diff
@@ -232,7 +240,36 @@ func (rt *RestTester) Bucket() base.Bucket {
...
+				}
+				if scope != nil && collection != nil {
+					collectionBucket, err := base.AsCollection(testBucket.Bucket)
+					if err != nil {
+						rt.tb.Fatalf("Could not get collection from bucket with type %T: %v", testBucket.Bucket, err)
+					}
+
+					collectionBucket.Spec.Scope = scope
+					collectionBucket.Spec.Collection = collection
+					collectionBucket.Collection = collectionBucket.Collection.Bucket().Scope(*scope).Collection(*collection)
```
adamcfraser: What exactly is going on here?  testBucket.Bucket is initialized as a gocb Collection targeting the default collection, and if we have a named collection we're using the associated Bucket() to swap to the underlying collection to the named collection?  
Are there any allocations associated with the default collection that need to be cleaned up?  I know Collection.Close() doesn't really do any work, but it feels a bit unusual to assign collectionBucket.Collection twice in this setup.

adamcfraser (changes requested):
A few questions on how we're switching from default to named collection.  I'm also wondering if this approach will be compatible with future changes to support multiple collections in a single RestTester context?

base/main_test_bucket_pool.go
IsaacLambat: Originally, this was put in when a Leaky TestBucket was to be passed in to the RT in-order to make sure the bucket was closed manually at the end of the test and can't be missed. I left this change in (even though a TB is no longer passed in) because it would be useful in general as it is quite easy to miss a bucket being left unclosed when not going through all the test logs.

rest/utilities_testing.go
IsaacLambat: There are no extra allocations done when changing the scope and collection to a named scope and collection. This is because the GoCB `Scope` and `Collection` functions return the provided bucket wrapped in a GoCB `Scope` and `Collection` struct.

IsaacLambat:
> A few questions on how we're switching from default to named collection. I'm also wondering if this approach will be compatible with future changes to support multiple collections in a single RestTester context?

The change to add scopes and collections when using a leaky bucket is a temporary measure that can be ripped out once tests are backed by collections.

adamcfraser (approved):

PR #5680, opened 2022-08-05
Fix golangci-lint in actions

torcolvin:
Increase timeout for occasional slow runs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,8 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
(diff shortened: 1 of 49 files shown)


markspolakovs (commented):
Could you expand a bit more on "make comments forward compatible with go 1.19" - where's the spacing requirement coming from? This seems like a relatively noisy diff and I'm not fully sure why it's necessary.

markspolakovs:
Discussed on Slack, this is coming from https://tip.golang.org/doc/go1.19#go-doc:

>As part of this change, [gofmt](https://tip.golang.org/cmd/gofmt) now reformats doc comments to make their rendered meaning clearer.

adamcfraser (approved):

PR #5682, opened 2022-08-05
CBG-2234 make DCPClient.Close blocking

torcolvin:
This makes jenkins tests hang in rest sometimes, but this looks like it doesn't change behavior. It will pass if you run all the tests separately.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1017/
- [x] `GSI=false,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1019/

### DIFF
--- a/base/dcp_client.go
+++ b/base/dcp_client.go
@@ -127,10 +127,11 @@ func NewDCPClient(ID string, callback sgbucket.FeedEventCallbackFunc, options DC
 	return client, nil
 }
 
-// Start returns an error and a channel to indicate when the DCPClient is done. If Start returns an error, DCPClient.Close() needs to be called.
+// Start returns an error and a channel to indicate when the DCPClient is done.
 func (dc *DCPClient) Start() (doneChan chan error, err error) {
 	err = dc.initAgent(dc.spec)
 	if err != nil {
+		_ = dc.Close()
 		return dc.doneChannel, err
 	}
 	dc.startWorkers()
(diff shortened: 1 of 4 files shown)


base/dcp_client.go
```diff
@@ -127,26 +127,29 @@ func NewDCPClient(ID string, callback sgbucket.FeedEventCallbackFunc, options DC
...
 		}
 	}
 	return dc.doneChannel, nil
 }
 
-// Close is used externally to stop the DCP client. If the client was already closed due to error, returns that error
+// Close is used externally to stop the DCP client. If the client was already closed due to error, returns that error.
 func (dc *DCPClient) Close() error {
 	dc.close()
+	<-dc.doneChannel
```
adamcfraser, line 152: The goroutine in dc.close() is going to write dc.getCloseError() to the channel - can we read that instead of calling again on line 153?

Also, it would be more defensive to have a select with a timeout option on this channel read, what do you think?

base/dcp_client_test.go
```diff
@@ -333,8 +327,10 @@ func TestResumeStoppedFeed(t *testing.T) {
 		if bytes.HasPrefix(event.Key, []byte(t.Name())) {
 			count := atomic.AddUint64(&mutationCount, 1)
 			if count > 5000 {
-				err := dcpClient.Close()
-				assert.NoError(t, err)
+				// use go routine to not deadlock, since this callback will prevent a close from happening
```
adamcfraser, line 330: If DCP clients can't invoke Close() within the callback without spawning a separate goroutine, I'm not sure that a blocking external Close() is the pattern we want?  Seems like an easy way to introduce deadlocks.

adamcfraser (changes requested):

base/dcp_client.go
torcolvin: The problem is that if it is already closed when it hits close(), we won't get this error, it's just a closed channel. I don't feel strongly about having a timeout, although as you say we can hit deadlocks below.

Because the dcp feed can be closed at almost any time, there's a real chance of missing the close error when it does error when you select and wait for `DCPClient.close`. The only pattern that seems OK is to wait on doneChan and not have it return errors, but check the errors after.

base/dcp_client_test.go
torcolvin: See the above comment, but in all cases here we are doing a synchronous close so I wonder if a public helper function is worthwhile.

torcolvin:
Closing this PR to say that this isn't going to be done anytime soon, but leaving CBG-2234 branch alive in case we revisit it.

PR #5684, opened 2022-08-08
CBG-2204: Add collections support to BlipTesterClient

JRascagneres:
CBG-2204

 - Adds support for collections to BlipTesterClient
 - Introduces new `BlipTesterCollectionClient` struct which represents work for given collection. Existing client has map of these indexed by collection
 - To support backwards compatibility -- If no collections are specified a `BlipTesterCollectionClient` is created with just collection index of `""`. Existing functions work and just direct calls to that empty 'dummy' collection
 - Example of how to work with a collection client is `TestCollectionsReplication`

FIXMEs are implemented to make tests work awaiting CBG-2264

## Pre-review checklist
- [X] Removed debug logging (`fmt.Print`, `log.Print`, ...)
- [X] Logging sensitive data? Make sure it's tagged (e.g. `base.UD(docID)`, `base.MD(dbName)`)
- [X] Updated relevant information in the API specifications (such as endpoint descriptions, schemas, ...) in `docs/api`

- [X] Link upstream PRs
- [X] Update Go module dependencies when merged

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/584/

(one failure because it requires GSI, now skipped it)


rest/blip_api_collections_test.go
```diff
@@ -228,3 +229,83 @@ func TestBlipGetCollectionsAndSetCheckpoint(t *testing.T) {
...
+	resp := rt.SendAdminRequest(http.MethodPut, "/db."+scopeAndCollectionKey+"/doc1", "{}")
+	requireStatus(t, resp, http.StatusCreated)
+
+	require.NoError(t, rt.WaitForPendingChanges())
+
+	btcCollection, err := btc.GetCollection(scopeAndCollectionKey)
+	assert.NoError(t, err)
+
+	err = btcCollection.StartOneshotPull()
+	assert.NoError(t, err)
```
torcolvin: I'd turn this into require

rest/blip_api_collections_test.go
```diff
@@ -228,3 +229,83 @@ func TestBlipGetCollectionsAndSetCheckpoint(t *testing.T) {
...
+	require.NoError(t, err)
+	defer btc.Close()
+
+	resp := rt.SendAdminRequest(http.MethodPut, "/db."+scopeAndCollectionKey+"/doc1", "{}")
+	requireStatus(t, resp, http.StatusCreated)
+
+	require.NoError(t, rt.WaitForPendingChanges())
+
+	btcCollection, err := btc.GetCollection(scopeAndCollectionKey)
+	assert.NoError(t, err)
```
torcolvin: I'd turn this into require as well

rest/blip_client_test.go
```diff
@@ -102,7 +119,14 @@ func (btr *BlipTesterReplicator) initHandlers(btc *BlipTesterClient) {
 			panic("no digest sent with proveAttachment")
 		}
 
-		attData, err := btc.getAttachment(digest)
+		collection := msg.Properties[db.BlipCollection]
+
+		// FIXME: Remove when we have CBG-2264 implemented
+		collection = btc.Collections[0]
+
+		btcr := btc.CollectionClients[collection]
```
torcolvin: Can we reuse code from https://cs.github.com/couchbase/sync_gateway/blob/465adbdb0d3ce7062f53dc4c1bcbe122f7cde4a4/db/blip_handler.go?q=blipcollection#L132

or turn this into a function since it is duplicated 5x and we know we have to fix?

rest/blip_client_test.go
```diff
@@ -484,23 +531,85 @@ func createBlipTesterClientOpts(tb testing.TB, rt *RestTester, opts *BlipTesterC
...
 		return nil, err
 	}
 
+	btc.CollectionClients = make(map[string]*BlipTesterCollectionClient)
+
+	if len(opts.Collections) != 0 {
+		getCollectionsRequest := blip.NewRequest()
+		getCollectionsRequest.SetProfile(db.MessageGetCollections)
+
+		type CollectionsRequest struct {
```
torcolvin: dupe of https://cs.github.com/couchbase/sync_gateway/blob/5efe9e8dcbb3836091c7b93e0d85acccbdb43cad/db/blip_sync_messages.go?q=collectionsrequest#L491

rest/blip_client_test.go
```diff
@@ -484,23 +531,85 @@ func createBlipTesterClientOpts(tb testing.TB, rt *RestTester, opts *BlipTesterC
...
+			Collections:   btc.Collections,
+			CheckpointIDs: checkpointIDs,
+		}
+		body, err := base.JSONMarshal(requestBody)
+		if err != nil {
+			return nil, err
+		}
+
+		getCollectionsRequest.SetBody(body)
+		if err := btc.pullReplication.sendMsg(getCollectionsRequest); err != nil {
```
torcolvin, line 559: I think `sendMsg` is asynchronous so reading the Body() below might trigger race detector.

rest/blip_client_test.go
```diff
@@ -484,23 +531,85 @@ func createBlipTesterClientOpts(tb testing.TB, rt *RestTester, opts *BlipTesterC
...
+			return nil, err
+		}
+
+		type CollectionsResponseEntry struct {
+			LastSequence *int    `json:"last_sequence"`
+			Rev          *string `json:"rev"`
+		}
```
torcolvin, line 566: If this is the response, I wonder if we should change blip_api_collections_test.go to use this instead of `db.Body` which is `map[string]interface{}`

rest/blip_client_test.go
JRascagneres: Response() has built in wait

rest/blip_client_test.go
JRascagneres: Eh, maybe worth turning into a function but it'll be removed in a ticket this sprint and the FIXME's make it obvious

rest/blip_client_test.go
```diff
@@ -34,16 +34,31 @@ type BlipTesterClientOpts struct {
 	Channels               []string
 	SendRevocations        bool
 	SupportedBLIPProtocols []string
+	Collections            []string
 
 	// a deltaSrc rev ID for which to reject a delta
-	rejectDeltasForSrcRev string //nolint: structcheck // false structcheck positive due to https://github.com/golangci/golangci-lint/issues/826
+	rejectDeltasForSrcRev string // nolint: structcheck // false structcheck positive due to https://github.com/golangci/golangci-lint/issues/826
```
markspolakovs: That comment shouldn't have the space before `nolint` - from https://golangci-lint.run/usage/false-positives/#nolint-directive:

> Use //nolint instead of // nolint because machine-readable comments should have no space by Go convention.

rest/blip_client_test.go
```diff
@@ -514,30 +626,63 @@ func NewBlipTesterClientOptsWithRT(tb testing.TB, rt *RestTester, opts *BlipTest
...
+	btcReplicator.collection = collection
+	btcReplicator.collectionIdx = collectionIdx
+
+	btc.CollectionClients[collection] = btcReplicator
+	return nil
+}
+
+// DummyCollection() just returns the BlipTesterCollectionClient with empty key. This is used when not running in a
+// collection mode
+func (btc *BlipTesterClient) DummyCollection() *BlipTesterCollectionClient {
```
bbrks: dummy -> default
remove implementation detail 

```suggestion
// DefaultCollection returns the client's default collection
func (btc *BlipTesterClient) DefaultCollection() *BlipTesterCollectionClient {
```

rest/blip_client_test.go
```diff
@@ -872,13 +1027,13 @@ func (btr *BlipTesterReplicator) storeMessage(msg *blip.Message) {
...
 
-func (btc *BlipTesterClient) WaitForBlipRevMessage(docID, revID string) (msg *blip.Message, found bool) {
+func (btc *BlipTesterCollectionClient) WaitForBlipRevMessage(docID, revID string) (msg *blip.Message, found bool) {
 	ticker := time.NewTicker(50 * time.Millisecond)
 	timeout := time.After(10 * time.Second)
 	for {
 		select {
 		case <-timeout:
-			btc.rt.tb.Fatalf("BlipTesterClient timed out waiting for BLIP message docID: %v, revID: %v", docID, revID)
+			// btc.rt.tb.Fatalf("BlipTesterClient timed out waiting for BLIP message docID: %v, revID: %v", docID, revID)
```
bbrks: ?

rest/blip_client_test.go
```diff
@@ -514,30 +626,63 @@ func NewBlipTesterClientOptsWithRT(tb testing.TB, rt *RestTester, opts *BlipTest
...
+	return nil
+}
+
+// DummyCollection() just returns the BlipTesterCollectionClient with empty key. This is used when not running in a
+// collection mode
+func (btc *BlipTesterClient) DummyCollection() *BlipTesterCollectionClient {
+	return btc.CollectionClients[""]
+}
+
+func (btc *BlipTesterClient) GetCollection(collection string) (*BlipTesterCollectionClient, error) {
```
bbrks: nitpick - but following gocb's naming conventions

```suggestion
func (btc *BlipTesterClient) Collection(collection string) (*BlipTesterCollectionClient, error) {
```

rest/blip_client_test.go
```diff
@@ -661,6 +812,8 @@ func (btc *BlipTesterClient) PushRevWithHistory(docID, parentRev string, body []
 	proposeChangesRequest := blip.NewRequest()
 	proposeChangesRequest.SetProfile(db.MessageProposeChanges)
 	proposeChangesRequest.SetBody([]byte(fmt.Sprintf(`[["%s","%s","%s"]]`, docID, newRevID, parentRev)))
+	btc.addCollectionProperty(proposeChangesRequest)
```
bbrks: Does it make sense to do this inside `sendMsg` (with a special case for not doing it for `getCollections`)?

Having to specify this in every method feels error-prone or easy to miss it for new message types.

rest/blip_client_test.go
```diff
@@ -901,3 +1056,71 @@ func (btc *BlipTesterClient) GetBlipRevMessage(docID, revID string) (msg *blip.M
...
+}
+
+func (btc *BlipTesterClient) PullReplication() *BlipTesterReplicator {
+	return btc.pullReplication
+}
+
+func (btc *BlipTesterClient) PushReplication() *BlipTesterReplicator {
+	return btc.pushReplication
+}
+
```
bbrks: I'm not sure these changes were required and would reduce a lot of the diffs in tests if reverted

```suggestion
```

rest/blip_client_test.go
```diff
@@ -34,16 +34,31 @@ type BlipTesterClientOpts struct {
...
+	pushReplication *BlipTesterReplicator // CBL -> SG replications
+
+	CollectionClients map[string]*BlipTesterCollectionClient
+}
+
+type BlipTesterCollectionClient struct {
+	*BlipTesterClientOpts                       // Copy of parents
+	pullReplication       *BlipTesterReplicator // Pointer to parents
+	pushReplication       *BlipTesterReplicator // Pointer to parents
+	rt                    *RestTester           // Pointer to parents to allow test failures
```
bbrks: Can these not all be accessed via a pointer to the parent, instead of pointers to everything in the parent?

```suggestion
	parent *BlipTesterClient
```

bbrks (changes requested):

rest/blip_client_test.go
JRascagneres: Hrm pretty sure I didn't change that... I think it might be goimports... Wonder if I can add an exception to avoid it doing that somehow

rest/blip_client_test.go
JRascagneres: @bbrks I had Default originally but changed it. I didn't want it to be confused with the actual `_default` collection.... Are you fine with this?

rest/blip_client_test.go
JRascagneres: Yeah thats fair, originally didn't have so many items pointing to parent.

rest/blip_client_test.go
JRascagneres: Made more sense when the replicators were on the collection clients

rest/blip_client_test.go
JRascagneres: Not sure how I missed this 🤦

rest/blip_client_test.go
JRascagneres: I renamed it back but kinda apprehensive. It does kinda use _default collection but doesn't actually use any of the getCollection stuff so kinda weird.

rest/blip_client_test.go
JRascagneres: I've had a go at this and couldn't really figure out a neat way of doing it. Besides requiring a user to pass in a collection to `sendMsg`. Do you think thats preferable?

rest/blip_client_test.go
JRascagneres: Right okay. Changed to //nolint:structcheck as discussed in slack.

rest/blip_client_test.go
```diff
@@ -117,6 +135,15 @@ func (btr *BlipTesterReplicator) initHandlers(btc *BlipTesterClient) {
...
 		btr.storeMessage(msg)
 
+		collection := msg.Properties[db.BlipCollection]
+
+		// FIXME: Remove when we have CBG-2264 implemented
+		if len(btc.Collections) > 0 {
+			collection = btc.Collections[0]
+		}
+
+		btcr := btc.CollectionClients[collection]
```
markspolakovs, line 141: This doesn't seem right - [per spec](https://docs.google.com/document/d/1kbPUuWgtnAgV9Oj55_soWmIPA8f9v3RBrjMXVvbn4p4/edit#heading=h.6zpikbtvy0cn), the `collection` property is the index into the original `getCollections` body array, but it looks like `btc.CollectionClients` is indexed by the collection name. Am I missing something?

rest/blip_client_test.go
```diff
@@ -594,11 +733,12 @@ func (btc *BlipTesterClient) Close() {
 	btc.attachments = make(map[string][]byte, 0)
 	btc.attachmentsLock.Unlock()
 
-	btc.pullReplication.Close()
-	btc.pushReplication.Close()
+	btc.parent.pullReplication.Close()
+	btc.parent.pushReplication.Close()
```
bbrks: Wouldn't this cause the parents to be closed multiple times when >1 collection?

I think the replicators can continue to be closed inside `BlipTesterClient.Close()` just once (preferably before the collections are closed to avoid in-flight things reading/writing to the above fields)

rest/blip_client_test.go
bbrks: Cleanest way I can think of is `sendPushMsg`/`sendPullMsg` wrappers on a `BlipTesterCollectionClient` that does the job of adding the collection property and then invoking the parent's replicator's `sendMsg` method

rest/blip_client_test.go
bbrks: +1 - something looks off here about how we're accessing vs. how `initCollectionReplication` is adding items to the map

bbrks (changes requested):

rest/blip_client_test.go
JRascagneres: Yeah fair enough, missed that, talked to Marks about this, this morning. Going to keep it internally indexed by name though and convert it. I think this is easier for debugging (as its a test client) and think it makes it easier to follow.

rest/blip_client_test.go
```diff
@@ -34,16 +34,28 @@ type BlipTesterClientOpts struct {
...
 // BlipTesterClient is a fully fledged client to emulate CBL behaviour on both push and pull replications through methods on this type.
 type BlipTesterClient struct {
 	BlipTesterClientOpts
 
-	rt *RestTester
+	rt              *RestTester
+	pullReplication *BlipTesterReplicator // SG -> CBL replications
+	pushReplication *BlipTesterReplicator // CBL -> SG replications
+
+	CollectionClients map[string]*BlipTesterCollectionClient
```
markspolakovs, line 51:
```suggestion
	// Per-collection replications, keyed by collection name (scope.collection)
	CollectionClients map[string]*BlipTesterCollectionClient
```

rest/blip_client_test.go
```diff
@@ -484,23 +521,80 @@ func createBlipTesterClientOpts(tb testing.TB, rt *RestTester, opts *BlipTesterC
...
+		}
+
+		for i, perCollectionResponse := range collectionResponse {
+			if perCollectionResponse == nil {
+				return nil, fmt.Errorf("collection doesn't exist on peer %s", opts.Collections[i])
+			}
+		}
+	}
+
+	if len(opts.Collections) == 0 {
```
markspolakovs, line 586: NITPICK: can this if statement be combined with the one above?

markspolakovs (commented):
Seems fine to me, but I'm not too familiar with this code so will leave final approval to Ben or Adam.

bbrks:
@JRascagneres this has picked up a conflict

bbrks (approved):

PR #5685, opened 2022-08-08
run golangci-lint --fix to address whitespace linting

torcolvin:
Single code review so these don't show up in each PR, since they are automatically fixed by goland.


bbrks (approved):
Do you want this merging?

PR #5686, opened 2022-08-09
CBG-2274: Use sync function defined on collection

markspolakovs:
CBG-2274

Update sync function handling for named collections - if one is defined there, use it instead of the DB's, and change the default to match PRD (collection name rather than `doc.channels`).

## Pre-review checklist

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/583/


markspolakovs:
As discussed, this is code that will need redoing from scratch anyway when it comes time to add multi-collection support, so it's perhaps too soon to do this now.

PR #5689, opened 2022-08-10
CBG-788 set agent priority to medium for caching feed/CE import feed

torcolvin:
## Pre-review checklist

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/CBG-788/

### DIFF
--- a/base/dcp_client.go
+++ b/base/dcp_client.go
@@ -46,17 +46,19 @@ type DCPClient struct {
 	checkpointPrefix           string                         // DCP checkpoint key prefix
 	checkpointPersistFrequency *time.Duration                 // Used to override the default checkpoint persistence frequency
 	dbStats                    *expvar.Map                    // Stats for database
+	agentPriority              gocbcore.DcpAgentPriority      // agentPriority specifies the priority level for a dcp stream
 }
 
 type DCPClientOptions struct {
 	NumWorkers                 int
 	OneShot                    bool
-	FailOnRollback             bool                 // When true, the DCP client will terminate on DCP rollback
-	InitialMetadata            []DCPMetadata        // When set, will be used as initial metadata for the DCP feed.  Will override any persisted metadata
-	CheckpointPersistFrequency *time.Duration       // Overrides metadata persistence frequency - intended for test use
-	MetadataStoreType          DCPMetadataStoreType // define storage type for DCPMetadata
-	GroupID                    string               // specify GroupID, only used when MetadataStoreType is DCPMetadataCS
-	DbStats                    *expvar.Map          // Optional stats
+	FailOnRollback             bool                      // When true, the DCP client will terminate on DCP rollback
+	InitialMetadata            []DCPMetadata             // When set, will be used as initial metadata for the DCP feed.  Will override any persisted metadata
+	CheckpointPersistFrequency *time.Duration            // Overrides metadata persistence frequency - intended for test use
+	MetadataStoreType          DCPMetadataStoreType      // define storage type for DCPMetadata
+	GroupID                    string                    // specify GroupID, only used when MetadataStoreType is DCPMetadataCS
+	DbStats                    *expvar.Map               // Optional stats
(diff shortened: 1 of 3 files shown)


base/dcp_client.go
```diff
@@ -76,6 +78,9 @@ func NewDCPClient(ID string, callback sgbucket.FeedEventCallbackFunc, options DC
 		return nil, fmt.Errorf("Unable to determine maxVbNo when creating DCP client: %w", err)
 	}
 
+	if options.AgentPriority == gocbcore.DcpAgentPriorityHigh {
+		return nil, fmt.Errorf("sync gateway should not set high proirity to DCP feeds")
```
adamcfraser:
```suggestion
		return nil, fmt.Errorf("sync gateway should not set high priority for DCP feeds")
```

adamcfraser (dismissed):

base/dcp_client_test.go
```diff
@@ -385,3 +385,26 @@ func TestResumeStoppedFeed(t *testing.T) {
 	}
 
 }
+
+// TestBadAgentPriority makes sure we can not specify agent priority as high
+func TestBadAgentPriority(t *testing.T) {
+	if UnitTestUrlIsWalrus() {
+		t.Skip("This test only works against Couchbase Server, since DCPClient requires collections")
```
bbrks: There's nothing collection specific here (`AsCollection` means we're using gocbv2 against Couchbase Server, not necessarily that we're using a named collection)

```suggestion
		t.Skip("This test only works against Couchbase Server, since DCPClient requires a Couchbase DCP feed")
```

base/dcp_client_test.go
```diff
@@ -385,3 +385,26 @@ func TestResumeStoppedFeed(t *testing.T) {
...
+		panic("Should not hit this callback")
+	}
+	dcpClientOpts := DCPClientOptions{
+		AgentPriority: gocbcore.DcpAgentPriorityHigh,
+	}
+	collection, err := AsCollection(bucket)
+	require.NoError(t, err)
+	dcpClient, err := NewDCPClient(feedID, panicCallback, dcpClientOpts, collection)
+	require.Error(t, err)
+	require.Nil(t, dcpClient)
```
bbrks, line 410: If we don't get an error here - do we want the the test to close the `DCPClient` we managed to erroneously open? I think we'd leak it otherwise.

base/dcp_client_test.go
```diff
@@ -385,3 +385,26 @@ func TestResumeStoppedFeed(t *testing.T) {
...
+	if UnitTestUrlIsWalrus() {
+		t.Skip("This test only works against Couchbase Server, since DCPClient requires collections")
+	}
+
+	bucket := GetTestBucket(t)
+	defer bucket.Close()
+
+	feedID := "fakeID"
+	panicCallback := func(event sgbucket.FeedEvent) bool {
+		panic("Should not hit this callback")
```
bbrks: if we panic we'd bail out of the entire test suite - rather than failing this one test.

```suggestion
		t.Errorf("Should not hit this callback")
```

bbrks (changes requested):
Change looks fine - just some comments on the test to better handle it failing

base/dcp_client_test.go
torcolvin: calling NewDCPClient doesn't actually open the dcp client, it just sets it up! Start() is what actually opens it.

base/dcp_client_test.go
torcolvin: DCPClient requires type `base.Collection`, which is not a named collection. Is there a better way to phrase this? i.e. it can't work with walrus because it can't be a cbdatasource.

base/dcp_client_test.go
torcolvin: done

base/dcp_client_test.go
bbrks: As discussed maybe just `base.Collection` for less ambiguity

bbrks (approved):

PR #5692, opened 2022-08-11
Consistently clean up test scopes/collections

markspolakovs:
Tests that use `FlushBucketEmptierFunc` (`auth` and `base` packages) would run `dropAllScopesAndCollections` when cleaning up a test bucket, but tests that use `ViewsAndGSIBucketReadier` would not, leading to test failures because of dangling scopes/collections. This PR makes the latter also use the (newly exported) `DropAllScopesAndCollections`.

## Pre-review checklist

- [ ] `xattrs=true,gsi=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/606/

### DIFF
--- a/db/util_testing.go
+++ b/db/util_testing.go
@@ -230,12 +230,6 @@ WHERE META(ks).xattrs._sync.sequence >= 0
 
 // ViewsAndGSIBucketReadier empties the bucket, initializes Views, and waits until GSI indexes are empty. It is run asynchronously as soon as a test is finished with a bucket.
 var ViewsAndGSIBucketReadier base.TBPBucketReadierFunc = func(ctx context.Context, b base.Bucket, tbp *base.TestBucketPool) error {
-	if c, ok := b.(*base.Collection); ok {
-		if err := c.DropAllScopesAndCollections(); err != nil && !errors.Is(err, base.ErrCollectionsUnsupported) {
-			return err
-		}
-	}
-
 	if base.TestsDisableGSI() {
 		tbp.Logf(ctx, "flushing bucket and readying views")
 		if err := base.FlushBucketEmptierFunc(ctx, b, tbp); err != nil {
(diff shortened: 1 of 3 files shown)


db/util_testing.go
```diff
@@ -230,6 +230,11 @@ WHERE META(ks).xattrs._sync.sequence >= 0
 
 // ViewsAndGSIBucketReadier empties the bucket, initializes Views, and waits until GSI indexes are empty. It is run asynchronously as soon as a test is finished with a bucket.
 var ViewsAndGSIBucketReadier base.TBPBucketReadierFunc = func(ctx context.Context, b base.Bucket, tbp *base.TestBucketPool) error {
+	if c, ok := b.(*base.Collection); ok {
+		if err := c.DropAllScopesAndCollections(); err != nil && !errors.Is(err, base.ErrCollectionsUnsupported) {
```
IsaacLambat: It seems that running in the GSI disabled case, `DropAllScopesAndCollections` would run 3 times: here, in `FlushBucketEmptierFunc` and in `collection.Flush` which could potentially cause test tear down to be slower. Shouldn't be a big issue because this teardown is done in the background, but could this cause some issues?

Unrelated to this change but for current behaviour, it seems every test tear down that uses the `FlushBucketEmptierFunc` runs `DropAllScopesAndCollections` twice: in `FlushBucketEmptierFunc` and then in `collection.Flush` (which `FlushBucketEmptierFunc` calls). Is this worth fixing in this PR too?

IsaacLambat (changes requested):
Change seems good to me as long as it won't cause any test issues with `DropAllScopesAndCollections` being called a multiple times each tear down

db/util_testing.go
markspolakovs: Good catch - I reckon that moving this call to be after the `base.TestsDisableGSI()` check should get rid of one of those.

markspolakovs: Also removed the call in `Collection.Flush()` per Slack discussion.

IsaacLambat (dismissed):
LGTM. Can merge after integration test passes

bbrks (dismissed):

markspolakovs:
Rebased to pick up #5695 unit test fix.

IsaacLambat (approved):

PR #5694, opened 2022-08-12
CBG-2281 skip test that is failing in GH since go1.18 upgrade

torcolvin:
This is just a stopgap solution, since it causes more windows failures than not in GH.

## Pre-review checklist


JRascagneres:
Looks like JWT failure too which I've seen a few times...
If we decide to skip these for the time being make sure we've filed tickets to revisit this.

bbrks (approved):

PR #5697, opened 2022-08-12
Update timeout and assert response error on sync fn and import filter timeout tests

IsaacLambat:
- Update timeouts for `TestSyncFnTimeout` and `TestImportFilterTimeout` to 15 seconds instead of 5
- Assert response error text for `TestSyncFnTimeout` to make sure the error is the error expected

## Pre-review checklist

### DIFF
--- a/rest/api_test.go
+++ b/rest/api_test.go
@@ -9719,10 +9719,10 @@ func TestSyncFnTimeout(t *testing.T) {
 	syncFnFinishedWG.Add(1)
 	go func() {
 		response := rt.SendAdminRequest("PUT", "/db/doc", `{"foo": "bar"}`)
-		requireStatus(t, response, 500) // TODO: Change to assertStatus when CBG-2143 merged maybe with response matching if response is changed
+		assertHTTPErrorReason(t, response, 500, "JS sync function timed out")
 		syncFnFinishedWG.Done()
 	}()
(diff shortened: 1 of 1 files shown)


rest/api_test.go
```diff
@@ -9719,10 +9719,10 @@ func TestSyncFnTimeout(t *testing.T) {
 	syncFnFinishedWG.Add(1)
 	go func() {
 		response := rt.SendAdminRequest("PUT", "/db/doc", `{"foo": "bar"}`)
-		requireStatus(t, response, 500) // TODO: Change to assertStatus when CBG-2143 merged maybe with response matching if response is changed
+		assertHTTPErrorReason(t, response, 500, "JS sync function timed out")
```
torcolvin, line 9722: I don't think this needs to hit this code review but I'd rather see `requireHTTPErrorReason`

torcolvin (approved):

rest/api_test.go
IsaacLambat: Having a require in the Go routine causes the test to carry on hanging (due to the Go routine being halted) until the timeout is hit on line 9725. I wanted this assert so if the test does fail, it does not hang until the timeout is hit, but rather does `syncFnFinishedWG.Done()` so the test can finish instantly.

PR #5708, opened 2022-08-18
CBG-2257: Per bucket credentials can now be used for fetching configs

IsaacLambat:
CBG-2257

- Fetching configs now only grabs buckets from the `bucket_credentials` ~~that don't already have a db associated with them~~ when running in serverless mode.
- `CouchbaseCluster.getBucket` can now use bucket credentials if set in the startup config `bucket_credentials` field. This will be force used when running in serverless mode.
	- This also means that `InsertConfig` and `DeleteConfig` use the per bucket authentication details which is useful for CBG-2258

## Pre-review checklist

- [ ] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/660/

### DIFF
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -37,7 +37,7 @@ var _ BootstrapConnection = &CouchbaseCluster{}
 // NewCouchbaseCluster creates and opens a Couchbase Server cluster connection.
 func NewCouchbaseCluster(server, username, password,
 	x509CertPath, x509KeyPath, caCertPath string,
-	forcePerBucketAuth bool, perBucketAuth map[string]*gocb.Authenticator,
+	forcePerBucketAuth bool, perBucketCreds PerBucketCredentialsConfig,
 	tlsSkipVerify *bool) (*CouchbaseCluster, error) {
 
 	securityConfig, err := GoCBv2SecurityConfig(tlsSkipVerify, caCertPath)
@@ -45,7 +45,7 @@ func NewCouchbaseCluster(server, username, password,
 		return nil, err
 	}
 
-	authenticatorConfig, err := GoCBv2Authenticator(
+	clusterAuthConfig, err := GoCBv2Authenticator(
 		username, password,
 		x509CertPath, x509KeyPath,
 	)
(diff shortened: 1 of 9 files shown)


rest/main.go
```diff
@@ -359,9 +360,22 @@ func backupCurrentConfigFile(sourcePath string) (string, error) {
...
+		)
+		if err != nil {
+			return nil, err
+		}
+		perBucketAuth[bucket] = &authenticator
+	}
+
+	cluster, err := base.NewCouchbaseCluster(config.Bootstrap.Server, config.Bootstrap.Username, config.Bootstrap.Password,
+		config.Bootstrap.X509CertPath, config.Bootstrap.X509KeyPath, config.Bootstrap.CACertPath,
+		base.BoolDefault(config.Unsupported.Serverless, false), perBucketAuth, config.Bootstrap.ServerTLSSkipVerify)
```
adamcfraser: I think an accessor method like config.IsServerless might be a good idea.

rest/main.go
```diff
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/couchbase/gocb/v2"
```
adamcfraser: Generally we're trying to avoid having gocb-specific information in anywhere but the base package, to facilitate future changes to SDK versions or Bucket implementations. In this case I think it's preferable to pass the config.BucketCredentials down to NewCouchbaseCluster.

rest/config_test.go
```diff
@@ -2529,3 +2529,85 @@ func TestBucketCredentialsValidation(t *testing.T) {
 		})
 	}
 }
+
+// Tests behaviour of CBG-2257 to poll only buckets in BucketCredentials that don't currently have a database
+func TestServerlessPollBuckets(t *testing.T) {
```
adamcfraser: Should the 'poll only buckets that don't have a database' part of this wait until CBG-2280 is completed?  Or at least be based on a flag that's enabled when CBG-2280 is done?  Otherwise running in serverless mode isn't going to pick up changes to database configurations.

rest/config_test.go
```diff
@@ -2529,3 +2529,85 @@ func TestBucketCredentialsValidation(t *testing.T) {
 		})
 	}
 }
+
```
adamcfraser: I think it would be a good idea to start putting the serverless tests in their own file(s), so that we can maintain a better picture of the suite of tests we have for serverless.

base/bootstrap.go
```diff
@@ -78,6 +88,25 @@ func (cc *CouchbaseCluster) connect() (*gocb.Cluster, error) {
 	return cluster, nil
 }
 
+// connect attempts to open a gocb.Cluster connection. Callers will be responsible for closing the connection.
+func (cc *CouchbaseCluster) connect() (*gocb.Cluster, error) {
```
adamcfraser: We don't have that many places that call connect().  Did you consider just adding authenticator to the connect() function.  I think that might make it more obvious what callers are doing (i.e. they are either passing in the cluster credentials, or bucket credentials).

adamcfraser (changes requested):

rest/config_test.go
IsaacLambat: Commented out the code that does this so it can be enabled as part of CBG-2280

adamcfraser (approved):

PR #5710, opened 2022-08-19
CBG-2279 validate views is incompatible with collections

torcolvin:
- Validate in config parser
- Throw error in tests if you call InitializeViews. As a result, `UseViews=false` on Collection specific tests.

## Pre-review checklist

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/676


markspolakovs (commented):
Is it also worth adding a `!TestsDisableGSI()` skip condition to `base.TestRequiresCollections()`? Otherwise those tests will fail when (accidentally?) run in integration with views.

torcolvin:
> Is it also worth adding a `!TestsDisableGSI()` skip condition to `base.TestRequiresCollections()`? Otherwise those tests will fail when (accidentally?) run in integration with views. 

They won't fail directly, only if they call `InitializeViews` - this is a stop gap solution until we can enable a full suite as collections backed. The main concern about requiring GSI now is that we don't regularly run with Jenkins, but I imagine when this switches to the default way of running in jenkins, we can add collections requires GSI.

rest/blip_api_collections_test.go
```diff
@@ -28,6 +28,7 @@ func TestBlipGetCollections(t *testing.T) {
 						},
 					},
 				},
+				UseViews: base.BoolPtr(false),
```
adamcfraser: Given the current test structure (where the suite gets run with views or GSI), should these tests be doing a skip based on TestsDisableGSI(), like we do for the N1QL tests, instead of trying to force views=false?  I feel like this approach might break with a test bucket pool that's initializing for views.

adamcfraser (changes requested):

rest/blip_api_collections_test.go
torcolvin: I would say yes in general, but as per thread with Marks, until we have GSI tests running at all, these are the only tests of collections. I'd probably do this when running tests with GSI becomes default under Jenkins.

torcolvin: This is too confusing even if it happens to work, and will be moot once we move to testing GSI regularly with jenkins. In the meantime, someone working on collections should remember to run with `SG_TEST_USE_GSI=true go test -run "TestCollection"`

adamcfraser (dismissed):

adamcfraser (approved):
LGTM.

PR #5713, opened 2022-08-23
CBG-2289: Use more descriptive DCP stream names for attachment compaction

gregns1:
CBG-2289

Small change to add more descriptive DCP stream names for attachment compaction. For example new streams will look like this sg-3.1:att_compaction:TestAttachmentSweep_sweep where 'TestAttachmentSweep' will be a UUID for the stream. 

## Pre-review checklist

- [ ] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/677/

### DIFF
--- a/db/attachment_compaction.go
+++ b/db/attachment_compaction.go
@@ -3,6 +3,7 @@ package db
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


db/attachment_compaction.go
```diff
@@ -551,3 +559,12 @@ func getCompactionDCPClientOptions(collection *base.Collection, groupID string)
 	return clientOptions, nil
 
 }
+
+func generateCompactionDCPStreamName(compactionID, compactionAction string) string {
```
markspolakovs, line 563: Have you seen / considered using `base.GenerateDcpStreamName`? Even if it needs changing to work with attachment compaction, it might still be worthwhile instead of a new function that does something very similar.

IsaacLambat: This would require big changes to that function, and would mean making two code paths as well as arguments needing to be passed in that may not be used depending on the code path and would get quite messy. I think it would be best to have this change separate to keep the code clean, and the attachment compaction functions all together.

JRascagneres: Agreed, trying to reuse that function wouldn't be clean. Even stuff like the UUID generation can't be used as we need to pass in the compaction ID. You'd basically be passing everything down at that point.

JRascagneres (approved):
Approved pending integration passing

IsaacLambat (approved):

PR #5714, opened 2022-08-23
Make test bucket pool wait until query service is ready before use

bbrks:
Make the test bucket pool wait for a query service to be ready before attempting to use it to clean out the prepared statements.

In scripted constrained environments like integration test jobs it's easier to hit a case where we try running this query and the query service has not started up yet. Especially in older CB Server versions, where startup is generally slower.

## Pre-review checklist

- [x] `6.6.5 xattrs=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/680/

### DIFF
--- a/base/bucket_n1ql.go
+++ b/base/bucket_n1ql.go
@@ -124,6 +124,19 @@ func (bucket *CouchbaseBucketGoCB) BuildDeferredIndexes(indexSet []string) error
 	return BuildDeferredIndexes(bucket, indexSet)
 }
 
+func (bucket *CouchbaseBucketGoCB) waitUntilQueryServiceReady(_ time.Duration) error {
+	pr, err := bucket.Bucket.Ping([]gocb.ServiceType{gocb.N1qlService})
+	if err != nil {
+		return err
+	}
+	for _, service := range pr.Services {
+		if service.Service == gocb.N1qlService && service.Success {
+			return nil
+		}
+	}
+	return fmt.Errorf("Couldn't ping N1QL service: %v", pr)
+}
+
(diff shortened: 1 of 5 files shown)


base/bucket_n1ql.go
```diff
@@ -124,6 +124,19 @@ func (bucket *CouchbaseBucketGoCB) BuildDeferredIndexes(indexSet []string) error
 	return BuildDeferredIndexes(bucket, indexSet)
 }
 
+func (bucket *CouchbaseBucketGoCB) waitUntilQueryServiceReady(_ time.Duration) error {
```
torcolvin, line 127: Is my understanding that only under `gocb/v1` that we use this code?

torcolvin (approved):

base/bucket_n1ql.go
bbrks: Yep

torcolvin: I'm about to pull out the https://cs.github.com/couchbase/sync_gateway/blob/e94af5ecfb6f85e9e0b4b5c91a7675560a343f99/base/util_testing.go#L663 because it won't work with collections - and this code then won't need to be modified. I tried to pull the whole file (bucket_n1ql.go) but there are some constants that are used by new files and it was a mess that I will defer.

You could remove that right now and then we will leave base/bucket_n1ql.go / base/bucket.go mostly unused, or we can preserve this for future.

bbrks: Is there a ticket to track removal of gocbv1 code? If not, filing that and getting the information in there would be good.

torcolvin: https://issues.couchbase.com/browse/CBG-2339

PR #5716, opened 2022-08-24
Handle stdlib wrapped errors in ErrorAsHTTPStatus

markspolakovs:
`base.ErrorAsHTTPStatus` today uses `pkgerrors.Cause` to get the underlying HTTP error, which works as expected when code uses `pkgerrors.Wrap()`, but not the more idiomatic (since Go 1.13) `fmt.Errorf()`. Replace `pkgerrors.Cause` with an `unwrap`per that properly handles errors wrapped per the Go 1.13 idiom.

`pkgerrors` also implements the Go 1.13 interfaces (https://github.com/pkg/errors/blob/5dd12d0cfe7f152f80558d591504ce685299311e/errors.go#L247-L248), so there should be no concerns for old code that still uses the `pkgerrors` style.

## Pre-review checklist

- [ ] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/697/

### DIFF
--- a/base/error.go
+++ b/base/error.go
@@ -169,7 +169,7 @@ func ErrorAsHTTPStatus(err error) (int, string) {
 	case *RetryTimeoutError:
 		return http.StatusGatewayTimeout, unwrappedErr.Error()
 	}
-	return http.StatusInternalServerError, fmt.Sprintf("Internal error: %v", unwrappedErr)
+	return http.StatusInternalServerError, fmt.Sprintf("Internal error: %v", err)
 }
 
 // Returns the standard CouchDB error string for an HTTP error status.
--- a/base/error.go
+++ b/base/error.go
@@ -81,6 +81,18 @@ func HTTPErrorf(status int, format string, args ...interface{}) *HTTPError {
 	return &HTTPError{status, fmt.Sprintf(format, args...)}
 }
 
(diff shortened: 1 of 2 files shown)


bbrks:
You mention this changing an error here #5717 

> Starting a SG against a CB 6.6.5 instance, when issuing a PUT /collections/ with a scope/collection defined, I get
> 
>     {
>       "error": "Internal Server Error",
>       "reason": "couldn't load database: couldn't reload database: Collections specified in database config, but not supported by connected Couchbase Server."
>     }
> 
> (nb: this won't be an "Internal Server Error" once https://github.com/couchbase/sync_gateway/pull/5716 is merged)

What will the value of `error` be after this change, in that case?
Do we need to worry about other unintended changes of responses due to this change?

markspolakovs:
With this change (and another change that I've just spotted that I'll push up in a moment), the error will become

```json
{
  "error": "Bad Request",
  "reason": "Named collections specified in database config, but not supported by connected Couchbase Server."
}
```

There is a risk of changes of responses, though I'd hope that the integration tests will check for that, since they assert on specific status codes. Happy to close if you feel it's too risky though.

markspolakovs:
Closing for now as not actually necessary.

PR #5717, opened 2022-08-24
CBG-2286: Check on startup if the bucket supports collections

markspolakovs:
CBG-2286

Refuse to create the DB if collections are configured, but the datastore doesn't support it.

## Pre-review checklist

- [ ] `package rest, CB 7.0.3` https://jenkins.sgwdev.com/job/SyncGateway-Integration/700/
- [ ] `package rest, CB 6.6.5` https://jenkins.sgwdev.com/job/SyncGateway-Integration/701/

## Manual verification

Starting a SG against a CB 6.6.5 instance, when issuing a `PUT /collections/` with a scope/collection defined, I get

```json
{
  "error": "Bad Request",
  "reason": "Named collections specified in database config, but not supported by connected Couchbase Server."
}
```

### DIFF
--- a/rest/server_context.go
+++ b/rest/server_context.go
@@ -40,6 +40,8 @@ const kStatsReportInterval = time.Hour
 const kDefaultSlowQueryWarningThreshold = 500 // ms
 const KDefaultNumShards = 16
 
+var errCollectionsUnsupported = base.HTTPErrorf(http.StatusBadRequest, "Named collections specified in database config, but not supported by connected Couchbase Server.")
+
 // Shared context of HTTP handlers: primarily a registry of databases by name. It also stores
 // the configuration settings so handlers can refer to them.
 // This struct is accessed from HTTP handlers running on multiple goroutines, so it needs to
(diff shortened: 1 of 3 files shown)


rest/server_context.go
```diff
@@ -408,6 +408,12 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(config DatabaseConfig, useE
 		return nil, err
 	}
 
+	if spec.Scope != nil && spec.Collection != nil {
```
bbrks: `Scope` and `Collection` on a `BucketSpec` are temporary and go away as soon as we need to support more than one collection.

We should be checking `len(DatabaseConfig.Scopes) > 0` to determine whether we're running with any named scope/collection, even if they happen to be `_default`.`_default`

bbrks (changes requested):

rest/server_context.go
markspolakovs: Will do - the only downside is that that means this check will happen after DbContext setup and index build, which means that you'll be waiting a while before finding out that your configuration is unsupported. Not a huge deal though.

bbrks: I dunno how we're able to build indexes on a collection if the server doesn't support collections

`config DatabaseConfig` in this function contains the nessesary info to determine whether the database is going ot be using a named collection, without creating a DatabaseContext first

markspolakovs: The index build would fail, but it'd be good if we gave a descriptive error than some N1QL internals. This is all moot though, because I misread `DatabaseConfig` as `DatabaseContext`...

bbrks (approved):

PR #5718, opened 2022-08-24
Convert some asserts in import_test.go into requires

markspolakovs:
Convert some asserts immediately before unconditional slice accesses (`[0]`) into requires, to ensure the test fails cleanly rather than panicking.

## Pre-review checklist

n/a


IsaacLambat (approved):
LGTM.
A few parts could also be turned to `requires` such as when unmarshalling JSON and writing to the bucket but it's not critical and would not cause any panics if it failed.

PR #5719, opened 2022-08-24
CBG-2344: Implement remaining DCPClient callbacks and always use collections

markspolakovs:
CBG-2344

Implement the DCPClient callbacks that we need to be fully Collections-aware:
* Expiration - handled similarly to Deletion, except as a separate stream event type
* SeqNoAdvanced - increments the VB seqno but otherwise no special handling

Also enables collections on DCPClient by default, even when only streaming the default collection - partly as a workaround for MB-53448 (which only affects "legacy" DCP clients).

Note that we don't handle the CollectionID on the DCP messages - we likely will need to when we move to multi-collection support.

## Pre-review checklist

- [x] https://github.com/couchbase/sg-bucket/pull/79

- [ ] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/699/
  * timed out

### DIFF
--- a/base/dcp_client.go
+++ b/base/dcp_client.go
@@ -31,6 +31,7 @@ type DCPClient struct {
 	workers                    []*DCPWorker                   // Workers for concurrent processing of incoming mutations and callback.  vbuckets are partitioned across workers
 	workersWg                  sync.WaitGroup                 // Active workers WG - used for signaling when the DCPClient workers have all stopped so the doneChannel can be closed
 	spec                       BucketSpec                     // Bucket spec for the target data store
+	supportsCollections        bool                           // Whether the target data store supports collections
 	numVbuckets                uint16                         // number of vbuckets on target data store
 	terminator                 chan bool                      // Used to close worker goroutines spawned by the DCPClient
 	doneChannel                chan error                     // Returns nil on successful completion of one-shot feed or external close of feed, error otherwise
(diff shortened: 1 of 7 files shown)


base/dcp_client.go
```diff
@@ -362,8 +364,14 @@ func (dc *DCPClient) openStreamRequest(vbID uint16) error {
 	vbMeta := dc.metadata.GetMeta(vbID)
 
 	options := gocbcore.OpenStreamOptions{}
-	if len(dc.collectionIDs) != 0 {
-		options.FilterOptions = &gocbcore.OpenStreamFilterOptions{CollectionIDs: dc.collectionIDs}
+	// Always use a collection-aware feed if supported
+	if dc.supportsCollections {
+		// If no collection IDs specified, filter to the default collection
+		collIds := dc.collectionIDs
+		if len(collIds) == 0 {
```
torcolvin, line 371: I don't like the assumption that this will always be 0, I think we should move this logic to `Collection.GetCollectionId` and pass `_default._default` or mark as unsupported.

torcolvin (commented):
I'd be tempted to change the callers here https://cs.github.com/couchbase/sync_gateway/blob/c33de1bf00e6b2ae12d81c4e62176a1f7f064e29/db/attachment_compaction.go?q=repo%3Acouchbase%2Fsync_gateway+collectionids#L543

https://cs.github.com/couchbase/sync_gateway/blob/841ecdf02dc6ed1e05df26c0e69f8fd130a55148/base/gocb_dcp_feed.go?q=repo%3Acouchbase%2Fsync_gateway+collectionids#L52

The key thing to remember is that these are probably about to become []*Collection.

It's not clear to me that this is better, but we seem like we are doing the logic twice - always pass collections in even if they are the default makes more sense if there are multiple collections.

There thing that Collection parameter is used for is two things which we'll probably need to split.

(1) metadata storage for checkpoints
(2) collection.Spec for TLS parameters and bucketname when creating `DCPAgentConfig`.

markspolakovs:
@torcolvin 

> The key thing to remember is that these are probably about to become []*Collection.

I'm not quite sure I follow - are you suggesting changing the `CollectionIDs []uint32` option to `Collections []*Collection`? That might make sense to do at some point, but I'd argue it's a bit outside the scope of this PR - happy to be corrected though. In particular, checkpoint storage seems more in the realm of [CBG-2266](https://issues.couchbase.com/browse/CBG-2266).

torcolvin:
> I'm not quite sure I follow - are you suggesting changing the `CollectionIDs []uint32` option to `Collections []*Collection`? That might make sense to do at some point, but I'd argue it's a bit outside the scope of this PR - happy to be corrected though. In particular, checkpoint storage seems more in the realm of [CBG-2266](https://issues.couchbase.com/browse/CBG-2266).

I'm just thinking in terms of making the API obviously forward compatible to when we pass `[]*Collection`, i.e. let the caller pass it in and don't worry in DCPClient whether it is supported or not.

markspolakovs:
> don't worry in DCPClient whether it is supported or not

We'll still need a check there, because this DCPClient code will need to work with server 6.6 (for the time being at least), so I'm not sure that we can safely assume collections support.

torcolvin:
> > don't worry in DCPClient whether it is supported or not
> 
> We'll still need a check there, because this DCPClient code will need to work with server 6.6 (for the time being at least), so I'm not sure that we can safely assume collections support.

I was suggesting we always add `_default._default` in the place where we pass in `collectionIDs` and if `collectionIDs` are zero, we keep existing behavior.

base/dcp_client_stream_observer.go
```diff
@@ -85,7 +85,14 @@ func (dc *DCPClient) End(end gocbcore.DcpStreamEnd, err error) {
 }
 
 func (dc *DCPClient) Expiration(expiration gocbcore.DcpExpiration) {
```
markspolakovs, line 89: Note to self: do we need to opt-in to expirations?

adamcfraser (approved):

PR #5722, opened 2022-08-26
CBG-2305: Implemented functions for suspending and unsuspending databases

IsaacLambat:
CBG-2305

- Added new functions on the ServerContext for suspending and unsuspending a database
- Changed to `getDatabase`to try to unsuspend the database first, before attempting to find the config in the buckets (which is cached when running in serverless).
- Added new function to fetch configs from the buckets on CBS and will cache them. By default, this will refresh the dbConfigs if it's older than a second when called. This is customisable through the config option `FetchConfigsTTL`.
- Added testing for this
- Fixed persistent config testing by setting default group ID
- Added db level option `suspendable` to indicate if a database can be suspended or not. Defaults to true if running in serverless mode

## Pre-review checklist

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/831/

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -63,7 +63,7 @@ const (
 
 	DefaultUseTLSServer = true
 
-	DefaultFetchConfigsTTL = time.Second
+	DefaultMinConfigFetchInterval = time.Second
 )
 
 // Bucket configuration elements - used by db, index
@@ -1342,6 +1342,18 @@ func (sc *ServerContext) fetchAndLoadConfigs(ctx context.Context, isInitialStart
 	return sc._applyConfigs(ctx, fetchedConfigs, isInitialStartup), nil
 }
 
+// fetchAndLoadDatabaseSince refreshes all dbConfigs if they where last fetched past the refreshInterval. It then returns found if
+// the fetched configs contain the dbName.
+func (sc *ServerContext) fetchAndLoadDatabaseSince(ctx context.Context, dbName string, refreshInterval *base.ConfigDuration) (found bool, err error) {
+	configs, err := sc.fetchConfigsSince(ctx, sc.config.Unsupported.Serverless.MinConfigFetchInterval)
+	if err != nil {
+		return false, err
+	}
+
+	found = configs[dbName] != nil
+	return found, nil
+}
+
 func (sc *ServerContext) fetchAndLoadDatabase(ctx context.Context, dbName string) (found bool, err error) {
 	sc.lock.Lock()
 	defer sc.lock.Unlock()
@@ -1426,16 +1438,16 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
 	return false, nil, nil
 }
 
-// fetchConfigsWithTTL returns database configs from the server context. These configs are refreshed before returning if
-// they are older than the duration set in the FetchConfigsTTL startup config option.
-func (sc *ServerContext) fetchConfigsWithTTL() (dbNameConfigs map[string]*DatabaseConfig, err error) {
-	ttl := DefaultFetchConfigsTTL
-	if sc.config.Unsupported.Serverless.FetchConfigsTTL != nil {
-		ttl = sc.config.Unsupported.Serverless.FetchConfigsTTL.Value()
+// fetchConfigsSince returns database configs from the server context. These configs are refreshed before returning if
+// they are older than the refreshInterval. The refreshInterval defaults to DefaultMinConfigFetchInterval if nil.
+func (sc *ServerContext) fetchConfigsSince(ctx context.Context, refreshInterval *base.ConfigDuration) (dbNameConfigs map[string]*DatabaseConfig, err error) {
+	minInterval := DefaultMinConfigFetchInterval
+	if refreshInterval != nil {
+		minInterval = refreshInterval.Value()
 	}
 
-	if sc.fetchConfigsLastUpdate.IsZero() || time.Now().Sub(sc.fetchConfigsLastUpdate) > ttl {
-		_, err = sc.fetchAndLoadConfigs(false)
+	if sc.fetchConfigsLastUpdate.IsZero() || time.Since(sc.fetchConfigsLastUpdate) > minInterval {
+		_, err = sc.fetchAndLoadConfigs(ctx, false)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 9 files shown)


adamcfraser (commented):
Looking at the changes here, I'm wondering if it would be a better approach to leverage the existing dbConfigs, instead of introducing an additional collection of DatabaseConfig on the server context - can discuss further offline.

rest/config.go
```diff
@@ -1427,6 +1431,25 @@ func (sc *ServerContext) fetchDatabase(dbName string) (found bool, dbConfig *Dat
...
 
+// fetchConfigsWithTTL returns database configs from the server context. These configs are refreshed before returning if
+// they are older than the duration set in the FetchConfigsTTL startup config option.
+func (sc *ServerContext) fetchConfigsWithTTL() (dbNameConfigs map[string]*DatabaseConfig, err error) {
+	ttl := DefaultFetchConfigsTTL
+	if sc.config.Unsupported.Serverless.FetchConfigsTTL != nil {
+		ttl = sc.config.Unsupported.Serverless.FetchConfigsTTL.Value()
+	}
+
+	if sc.fetchConfigsLastUpdate.IsZero() || time.Now().Sub(sc.fetchConfigsLastUpdate) > ttl {
```
adamcfraser: I think using `time.Since(sc.fetchConfigsLastUpdate) > ttl)` would simplify this, and also use the common case optimization for time.Now() inside Since.

rest/config.go
```diff
@@ -61,6 +62,8 @@ const (
 	DefaultNumIndexReplicas = uint(1)
 
 	DefaultUseTLSServer = true
+
+	DefaultFetchConfigsTTL = time.Second
```
adamcfraser: I don't think TTL is the right terminology for this - I specifically associate time-to-live with expiry functionality, and don't think it's equivalent to event frequency.  I might call this DefaultMaxConfigFetchFrequency or DefaultMinConfigFetchInterval.

rest/server_context.go
```diff
@@ -203,19 +206,29 @@ func (sc *ServerContext) GetDatabase(name string) (*db.DatabaseContext, error) {
...
+	}
+
+	// database not loaded, fallback to fetching it from cluster
 	if sc.bootstrapContext.connection != nil {
-		sc.lock.Lock()
-		defer sc.lock.Unlock()
-		// database not loaded, go look for it in the cluster
-		found, err := sc._fetchAndLoadDatabase(name)
+		if sc.config.IsServerless() {
+			_, err = sc.fetchConfigsWithTTL()
```
adamcfraser: The existing fallback on line 221 is only loading a single db by name.  Can we do the same in the serverless case?

rest/server_context.go
```diff
@@ -1036,6 +1049,88 @@ func (sc *ServerContext) _removeDatabase(dbName string) bool {
...
+}
+func (sc *ServerContext) _unsuspendDatabase(dbName string) (*db.DatabaseContext, error) {
+	dbCtx := sc.databases_[dbName]
+	if dbCtx != nil {
+		return dbCtx, nil
+	}
+
+	// Check if database is in dbConfigs so no need to search through buckets
+	if dbConfig, ok := sc.dbConfigs[dbName]; ok {
+		if !base.BoolDefault(dbConfig.Suspendable, sc.config.IsServerless()) {
```
adamcfraser, line 1097: I don't think we should be doing a check on suspendable when unsuspending?  Seems like we should be using this flag to prevent it getting into a suspended state, but if it somehow ends up in that state we shouldn't block it coming online.

rest/server_context.go
adamcfraser: Assuming yes, we should restore the 'found' handling to avoid locking if fetch wasn't successful.

rest/server_context.go
```diff
@@ -1036,6 +1049,88 @@ func (sc *ServerContext) _removeDatabase(dbName string) bool {
...
+		return ErrSuspendingDisallowed
+	}
+
+	bucket := dbCtx.Bucket.GetName()
+	base.InfofCtx(context.TODO(), base.KeyAll, "Suspending db %q (bucket %q)", base.MD(dbName), base.MD(bucket))
+
+	if !sc._unloadDatabase(dbName) {
+		return base.ErrNotFound
+	}
+	delete(sc.bucketDbName, bucket)
```
adamcfraser: Should unloadDatabase be removing the entry from sc.bucketDbName?  I'm wary of having a set of parallel storage that callers need to remember to clean up (and that need to get refactored when we have collections support)

rest/server_context.go
```diff
@@ -1036,6 +1049,88 @@ func (sc *ServerContext) _removeDatabase(dbName string) bool {
...
+	return nil
+}
+
+func (sc *ServerContext) unsuspendDatabase(dbName string) (*db.DatabaseContext, error) {
+	sc.lock.Lock()
+	defer sc.lock.Unlock()
+
+	return sc._unsuspendDatabase(dbName)
+}
+func (sc *ServerContext) _unsuspendDatabase(dbName string) (*db.DatabaseContext, error) {
```
adamcfraser: It looks like the only callers of _unsuspendDatabase and _suspendDatabase are the locking versions.  If the non-locking versions aren't needed otherwise, we can remove.

adamcfraser (changes requested):

rest/server_context.go
IsaacLambat: As discussed, I no longer remove this entry in the map so duplicate databases cannot be made in the bucket while a database is suspended. Plus the changes that will happen due to collections.

rest/config.go
```diff
@@ -1423,6 +1439,25 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
...
 
+// fetchConfigsSince returns database configs from the server context. These configs are refreshed before returning if
+// they are older than the refreshInterval. The refreshInterval defaults to DefaultMinConfigFetchInterval if nil.
+func (sc *ServerContext) fetchConfigsSince(ctx context.Context, refreshInterval *base.ConfigDuration) (dbNameConfigs map[string]*DatabaseConfig, err error) {
+	minInterval := DefaultMinConfigFetchInterval
+	if refreshInterval != nil {
+		minInterval = refreshInterval.Value()
+	}
+
+	if sc.fetchConfigsLastUpdate.IsZero() || time.Since(sc.fetchConfigsLastUpdate) > minInterval {
```
adamcfraser: I don't think you need the IsZero check if you're using time.Since?

rest/server_context.go
IsaacLambat: ~~These are used in CBG-2306~~ No longer used in CBG-2306

adamcfraser (commented):
One small suggestion - otherwise looks good.

adamcfraser (approved):

PR #5723, opened 2022-08-26
CBG-2306: Add REST API support for unsuspending a database

IsaacLambat:
CBG-2306

- If running in serverless, when requesting a database context
	- Check if database is active
	- If not, check authentication
	- If it's not loaded, an attempt to unsuspend it is done
	- If it's not suspended, then an attempt to load it from the buckets is done
- Added testing
- Split GetDatabase up

## Pre-review checklist

## Dependancies
- [x] #5722 

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/906

### DIFF
--- a/base/error.go
+++ b/base/error.go
@@ -217,6 +217,14 @@ func IsDocNotFoundError(err error) bool {
 	}
 }
 
+// IsHTTPErrorStatusNotFound returns true if the error is a HTTP Error with the status 404 Not Found
+func IsHTTPErrorStatusNotFound(err error) bool {
+	if httpError, ok := err.(*HTTPError); ok && httpError.Status == http.StatusNotFound {
+		return true
+	}
+	return false
+}
+
 // MultiError manages a set of errors.  Callers must use ErrorOrNil when returning MultiError to callers
 // in order to properly handle nil checks on the returned MultiError
 //  Sample usage:
--- a/base/error.go
+++ b/base/error.go
@@ -217,14 +217,6 @@ func IsDocNotFoundError(err error) bool {
 	}
 }
 
-// IsHTTPErrorStatusNotFound returns true if the error is a HTTP Error with the status 404 Not Found
-func IsHTTPErrorStatusNotFound(err error) bool {
-	if httpError, ok := err.(*HTTPError); ok && httpError.Status == http.StatusNotFound {
-		return true
-	}
-	return false
-}
-
 // MultiError manages a set of errors.  Callers must use ErrorOrNil when returning MultiError to callers
 // in order to properly handle nil checks on the returned MultiError
 //  Sample usage:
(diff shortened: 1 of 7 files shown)


rest/handler.go
```diff
@@ -266,24 +266,25 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
 	// look up the database context:
 	var dbContext *db.DatabaseContext
 	if keyspaceDb != "" {
+		if shouldCheckAdminAuth {
+			// Confirm admin port is authorized before attempting to load db
+			authorized, err := h.checkAdminAuthenticationOnly()
```
adamcfraser: This looks like it's going to make two calls to the permissions endpoint for all admin requests (here and on line 426), which we want to avoid. (the previous usage of checkAdminAuthenticationOnly was only in the case where the database didn't exist and we were determining whether to return 404 or 403).

adamcfraser (changes requested):

base/error.go
```diff
@@ -217,6 +217,14 @@ func IsDocNotFoundError(err error) bool {
 	}
 }
 
+// IsHTTPErrorStatusNotFound returns true if the error is a HTTP Error with the status 404 Not Found
+func IsHTTPErrorStatusNotFound(err error) bool {
```
adamcfraser: It looks like we're only using this in the context of GetActiveDatabase returning not found.  I don't think it makes sense to format the error as an HTTP error, and then have to parse that error again to see if it represents 'not found'.  Can we just have GetActiveDatabase return a non-HTTP error like ErrNotFound, or have it return a flag for found/not found?  Our rest handlers call ErrorAsHTTPStatus when returning errors, so I don't think we need to prematurely convert to an HTTP error.

rest/handler.go
```diff
@@ -267,24 +267,30 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
...
 					}
-
-					if authorized {
+					if !authorized {
+						return base.HTTPErrorf(http.StatusUnauthorized, "")
+					}
+				}
+				dbContext, err = h.server.GetInactiveDatabase(h.ctx(), keyspaceDb)
+				if err != nil {
+					base.InfofCtx(h.ctx(), base.KeyHTTP, "Error trying to get inactive db %s: %v", base.MD(keyspaceDb), err)
```
adamcfraser: If this is a not found error, I think that's an expected error case that doesn't require info-level logging.   I think we should only log in the case of an unexpected error.

rest/serverless_test.go
```diff
@@ -22,6 +22,7 @@ import (
 
 // Tests behaviour of CBG-2257 to poll only buckets in BucketCredentials that don't currently have a database
 func TestServerlessPollBuckets(t *testing.T) {
+	base.LongRunningTest(t)
```
adamcfraser: I'm not sure why these need to be added (and in one case below it's being added redundantly).  What aspect of these tests are inherently long-running?

adamcfraser (changes requested):
Overall approach looks good, but a question about how HTTPError is being used.

rest/serverless_test.go
IsaacLambat: Originally, I added this when each of the serverless tests where taking over 10 seconds, however now seem to be faster locally. 

Don't know what's caused it to be faster (whether it was a laptop issue or the conversion to using the rest tester for all the tests or general test framework improvements) however this is great as the tests are much faster now.

adamcfraser (approved):

PR #5724, opened 2022-08-30
CBG-2314: Per-DB Logging: Main, ServerContext, and DatabaseContext

cbbruno:
CBG-2314

Describe your PR here...
- Added context in APIs for main, handler, ServerContext, DatabaseContext, and respective Tests.
- Removed Database.Ctx, replaced by ctx parameter.
- Removed CbgtContext.loggingCtx, replaced by ctx parameter.

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/792/

### DIFF
--- a/base/dcp_sharded.go
+++ b/base/dcp_sharded.go
@@ -644,13 +644,13 @@ type CbgtDestFactoryFunc = func() (cbgt.Dest, error)
 var cbgtDestFactories = make(map[string]CbgtDestFactoryFunc)
 var cbgtDestFactoriesLock sync.Mutex
 
-func StoreDestFactory(destKey string, dest CbgtDestFactoryFunc) {
+func StoreDestFactory(ctx context.Context, destKey string, dest CbgtDestFactoryFunc) {
 	cbgtDestFactoriesLock.Lock()
 	_, ok := cbgtDestFactories[destKey]
 
 	// We don't expect duplicate destKey registration - log a warning if it already exists
 	if ok {
-		WarnfCtx(context.Background(), "destKey %s already exists in cbgtDestFactories - new value will replace the existing dest", destKey)
+		WarnfCtx(ctx, "destKey %s already exists in cbgtDestFactories - new value will replace the existing dest", destKey)
 	}
 	cbgtDestFactories[destKey] = dest
 	cbgtDestFactoriesLock.Unlock()
--- a/base/dcp_sharded.go
+++ b/base/dcp_sharded.go
@@ -54,8 +54,8 @@ type CbgtContext struct {
 
 // StartShardedDCPFeed initializes and starts a CBGT Manager targeting the provided bucket.
 // dbName is used to define a unique path name for local file storage of pindex files
-func StartShardedDCPFeed(dbName string, configGroup string, uuid string, heartbeater Heartbeater, bucket Bucket, spec BucketSpec, numPartitions uint16, cfg cbgt.Cfg) (*CbgtContext, error) {
-	cbgtContext, err := initCBGTManager(bucket, spec, cfg, uuid, dbName)
+func StartShardedDCPFeed(ctx context.Context, dbName string, configGroup string, uuid string, heartbeater Heartbeater, bucket Bucket, spec BucketSpec, numPartitions uint16, cfg cbgt.Cfg) (*CbgtContext, error) {
+	cbgtContext, err := initCBGTManager(ctx, bucket, spec, cfg, uuid, dbName)
 	if err != nil {
 		return nil, err
 	}
@@ -241,7 +241,7 @@ func getCBGTIndexUUID(manager *cbgt.Manager, indexName string) (exists bool, pre
 // createCBGTManager creates a new manager for a given bucket and bucketSpec
 // Inline comments below provide additional detail on how cbgt uses each manager
 // parameter, and the implications for SG
-func initCBGTManager(bucket Bucket, spec BucketSpec, cfgSG cbgt.Cfg, dbUUID string, dbName string) (*CbgtContext, error) {
+func initCBGTManager(ctx context.Context, bucket Bucket, spec BucketSpec, cfgSG cbgt.Cfg, dbUUID string, dbName string) (*CbgtContext, error) {
 	// Check if there are pre-Helium nodes, and if so, set the LithiumCompat flag to ensure we don't try to start a
 	// collections-enabled feed when some nodes won't support it.
 	minVersion, err := getMinNodeVersion(cfgSG)
@@ -343,7 +343,7 @@ func initCBGTManager(bucket Bucket, spec BucketSpec, cfgSG cbgt.Cfg, dbUUID stri
 		options)
 
 	cbgtContext := &CbgtContext{
-		loggingCtx: LogContextWith(context.Background(), &LogContext{CorrelationID: MD(spec.BucketName).Redact() + "-" + DCPImportFeedID}),
+		loggingCtx: LogContextWith(ctx, &LogContext{CorrelationID: MD(spec.BucketName).Redact() + "-" + DCPImportFeedID}),
 		Manager:    mgr,
 		Cfg:        cfgSG,
 	}
--- a/base/dcp_sharded.go
+++ b/base/dcp_sharded.go
@@ -49,7 +49,7 @@ type CbgtContext struct {
 	Cfg               cbgt.Cfg                 // Cfg manages storage of the current pindex set and node assignment
 	heartbeater       Heartbeater              // Heartbeater used for failed node detection
 	heartbeatListener *importHeartbeatListener // Listener subscribed to failed node alerts from heartbeater
-	loggingCtx        context.Context          // Context for cbgt logging
+	// loggingCtx        context.Context          // Context for cbgt logging
 }
 
 // StartShardedDCPFeed initializes and starts a CBGT Manager targeting the provided bucket.
@@ -61,7 +61,7 @@ func StartShardedDCPFeed(ctx context.Context, dbName string, configGroup string,
 	}
 
 	// Start Manager.  Registers this node in the cfg
-	err = cbgtContext.StartManager(dbName, configGroup, bucket, spec, numPartitions)
+	err = cbgtContext.StartManager(ctx, dbName, configGroup, bucket, spec, numPartitions)
 	if err != nil {
 		return nil, err
 	}
@@ -96,7 +96,7 @@ func GenerateLegacyIndexName(dbName string) string {
 // to the manager's cbgt cfg.  Nodes that have registered for this indexType with the manager via
 // RegisterPIndexImplType (see importListener.RegisterImportPindexImpl)
 // will receive PIndexImpl callbacks (New, Open) for assigned PIndex to initiate DCP processing.
-func createCBGTIndex(c *CbgtContext, dbName string, configGroupID string, bucket Bucket, spec BucketSpec, numPartitions uint16) error {
+func createCBGTIndex(ctx context.Context, c *CbgtContext, dbName string, configGroupID string, bucket Bucket, spec BucketSpec, numPartitions uint16) error {
 	sourceType := SOURCE_DCP_SG
 
 	bucketUUID, err := bucket.UUID()
@@ -146,8 +146,8 @@ func createCBGTIndex(c *CbgtContext, dbName string, configGroupID string, bucket
 	}
 
 	// Determine index name and UUID
-	indexName, previousIndexUUID := dcpSafeIndexName(c, dbName)
-	InfofCtx(c.loggingCtx, KeyDCP, "Creating cbgt index %q for db %q", indexName, MD(dbName))
+	indexName, previousIndexUUID := dcpSafeIndexName(ctx, c, dbName)
+	InfofCtx(ctx, KeyDCP, "Creating cbgt index %q for db %q", indexName, MD(dbName))
 
 	// Register bucketDataSource callback for new index if we need to configure TLS
 	cbgt.RegisterBucketDataSourceOptionsCallback(indexName, c.Manager.UUID(), func(options *cbdatasource.BucketDataSourceOptions) *cbdatasource.BucketDataSourceOptions {
@@ -158,12 +158,12 @@ func createCBGTIndex(c *CbgtContext, dbName string, configGroupID string, bucket
 		}
 
 		networkType := getNetworkTypeFromConnSpec(connSpec)
-		InfofCtx(c.loggingCtx, KeyDCP, "Using network type: %s", networkType)
+		InfofCtx(ctx, KeyDCP, "Using network type: %s", networkType)
 
 		// default (aka internal) networking is handled by cbdatasource, so we can avoid the shims altogether in this case, for all other cases we need shims to remap hosts.
 		if networkType != clusterNetworkDefault {
 			// A lookup of host dest to external alternate address hostnames
-			options.ConnectBucket, options.Connect, options.ConnectTLS = alternateAddressShims(c.loggingCtx, spec.IsTLS(), connSpec.Addresses, networkType)
+			options.ConnectBucket, options.Connect, options.ConnectTLS = alternateAddressShims(ctx, spec.IsTLS(), connSpec.Addresses, networkType)
 		}
 
 		return options
@@ -185,7 +185,7 @@ func createCBGTIndex(c *CbgtContext, dbName string, configGroupID string, bucket
 	)
 	c.Manager.Kick("NewIndexesCreated")
 
-	InfofCtx(c.loggingCtx, KeyDCP, "Initialized sharded DCP feed %s with %d partitions.", indexName, numPartitions)
+	InfofCtx(ctx, KeyDCP, "Initialized sharded DCP feed %s with %d partitions.", indexName, numPartitions)
 	return err
 
 }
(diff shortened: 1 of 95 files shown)


rest/admin_api_auth_test.go
```diff
@@ -193,25 +193,27 @@ func TestCheckPermissionsWithX509(t *testing.T) {
...
+	svrctx := NewServerContext(ctx, &StartupConfig{
 		Bootstrap: BootstrapConfig{
 			Server:       serverURL,
 			X509CertPath: certPath,
 			X509KeyPath:  keyPath,
 			CACertPath:   caCertPath,
 		},
 	}, false)
-	defer ctx.Close()
+	ctx = svrctx.AddServerLogContext(ctx)
```
adamcfraser: I think tests should only call AddServerLogContext when they need to (i.e. they are running with multiple server contexts and would like to differentiate the logging).

base/logging_context.go
```diff
@@ -115,17 +115,14 @@ func LogContextWith(parent context.Context, adder ContextAdder) context.Context
...
 
 func (c *ServerLogContext) getContextKey() LogContextKey {
 	return serverLogContextKey
 }
 
 func (c *ServerLogContext) addContext(format string) string {
-	if c != nil && c.ConfigGroupID != "" {
-		format = "g:" + c.ConfigGroupID + " " + format
-	}
+	// nothing to log for now
```
adamcfraser: Right now I think the only use case for ServerLogContext is for tests that run multiple server contexts (e.g. ISGR testing) and would like to differentiate the logging.  For that use case, maybe we have an optional property on ServerLogContext that they can set, and if that property isn't set then we don't log?

rest/handler.go
```diff
@@ -442,11 +456,10 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
 	// Now set the request's Database (i.e. context + user)
 	if dbContext != nil {
 		h.keyspaceScope, h.keyspaceCollection = *keyspaceScope, *keyspaceCollection
-		h.db, err = db.GetDatabase(dbContext, h.user)
+		h.db, err = db.GetDatabase(h.ctx(), dbContext, h.user)
```
adamcfraser: db.GetDatabase is going to call dbCtx.AddDatabaseLogContext(ctx), but this has already been done above on line 271.  I think we want to avoid wrapping the context multiple times for the same logging context, do you agree?

db/database.go
```diff
@@ -834,12 +841,14 @@ func (context *DatabaseContext) Authenticator(ctx context.Context) *auth.Authent
 }
 
 // Makes a Database object given its name and bucket.
-func GetDatabase(context *DatabaseContext, user auth.User) (*Database, error) {
-	return &Database{DatabaseContext: context, user: user}, nil
+func GetDatabase(ctx context.Context, dbCtx *DatabaseContext, user auth.User) (*Database, error) {
+	ctx = dbCtx.AddDatabaseLogContext(ctx) // add db log info before passing ctx down
+	return &Database{Ctx: ctx, DatabaseContext: dbCtx, user: user}, nil
```
adamcfraser: Do you think we'll be removing Database.Ctx in a future PR, and instead have Database functions accept a context?  Or is this a case where it makes sense to store a context on the struct.

db/blip_sync_context.go
```diff
@@ -225,7 +225,7 @@ func (bsc *BlipSyncContext) copyContextDatabase() *Database {
 }
 
 func (bsc *BlipSyncContext) _copyContextDatabase() *Database {
-	databaseCopy, _ := GetDatabase(bsc.blipContextDb.DatabaseContext, bsc.blipContextDb.User())
+	databaseCopy, _ := GetDatabase(context.TODO(), bsc.blipContextDb.DatabaseContext, bsc.blipContextDb.User())
```
adamcfraser: _copyContextDatabase gets called a lot (once per user-scoped blip handler, I think?).  I don't think we'd want to create a new context and add db log context to it (in GetDatabase) every time.  This might go back to the question of whether Database should still store a ctx in its struct, but I thought we'd want to pass bsc.loggingCtx here.

adamcfraser (changes requested):

base/logging_context.go
cbbruno: Sure, what would you suggest?  
- I think you mentioned in previous meeting, "ServerName"?
- How to pass that name/id/string ?  Currently we don't explicitly create a ServerLogContext, we just call ServerContext.AddServerLogContext.  
- Or can be a new property in RestTester and set only by RestTester.Context() ?
- Or maybe we can use ConfigGroupID if is diff from "default" ?

db/blip_sync_context.go
cbbruno: well, I just flagged it for now, as we talked for changing APIs, and eval what we can pass here later, when we get to /db package (and eval how bsc.loggingCtx is being set)... but I can change it now if you're sure of it, let me know.

rest/admin_api_auth_test.go
cbbruno: agreed , I'll do another pass to remove these.

rest/handler.go
cbbruno: agreed, let me see if I can remove LogContext from GetDatabase, thanks.

db/database.go
cbbruno: hmm... don't see this comment when browsing changed files (deleted?)... anyway, think we're using that Database struct to handle blip sync requests (as we use handler for API), so makes sense to me keep it... seems to be a bigger change (than logging) with blip sync setup if we remove that ctx.

base/logging_context.go
adamcfraser: Add a LogContextID to the ServerContext that can optionally be set by tests, and isn't logged when empty.

rest/blip_sync.go
```diff
@@ -37,16 +37,16 @@ func (h *handler) handleBLIPSync() error {
...
 		return err
 	}
 
 	// Overwrite the existing logging context with the blip context ID
-	h.db.Ctx = base.LogContextWith(h.db.Ctx, &base.LogContext{CorrelationID: base.FormatBlipContextID(blipContext.ID)})
+	h.rqCtx = base.LogContextWith(h.ctx(), &base.LogContext{CorrelationID: base.FormatBlipContextID(blipContext.ID)})
 
 	// Create a new BlipSyncContext attached to the given blipContext.
-	ctx := db.NewBlipSyncContext(blipContext, h.db, h.formatSerialNumber(), db.BlipSyncStatsForCBL(h.db.DbStats))
+	ctx := db.NewBlipSyncContext(h.ctx(), blipContext, h.db, h.formatSerialNumber(), db.BlipSyncStatsForCBL(h.db.DbStats))
```
adamcfraser: Should this be passing h.rqCtx?

rest/blip_sync.go
```diff
@@ -37,16 +37,16 @@ func (h *handler) handleBLIPSync() error {
 	}
 
 	// Create a BLIP context:
-	blipContext, err := db.NewSGBlipContext(h.db.Ctx, "")
+	blipContext, err := db.NewSGBlipContext(h.ctx(), "")
```
adamcfraser, line 40: Just to confirm, the existing code doesn't include the blipContextID when logging blip messages?  (i.e. doesn't include the log context added on line 46?). Just want to ensure you've checked that nothing is changing for the KeyWebSocket/KeyWebSocketFrame logging done by defaultBlipLogger.

rest/utilities_testing.go
```diff
@@ -801,6 +805,17 @@ func (rt *RestTester) ReplacePerBucketCredentials(config base.PerBucketCredentia
 	rt.ServerContext().bootstrapContext.connection = couchbaseCluster
 }
 
+func (rt *RestTester) Context() context.Context {
```
adamcfraser, line 820: This is going to initialize a new context every time it's called - is that intentional?

rest/utilities_testing.go
```diff
@@ -426,7 +430,7 @@ func (rt *RestTester) Close() {
 	}
 	rt.closed = true
 	if rt.RestTesterServerContext != nil {
-		rt.RestTesterServerContext.Close()
+		rt.RestTesterServerContext.Close(base.TestCtx(rt.tb))
```
adamcfraser: What's the rationale to use base.TestCtx(rt.tb) here instead of rt.Context()?

adamcfraser (changes requested):

rest/blip_sync.go
cbbruno: current code uses `h.db.Ctx`, when it creates NewBlipSyncContext, in line 49.

rest/blip_sync.go
cbbruno: yeah, to save the call, will do.

rest/utilities_testing.go
cbbruno: existing tests access RT ServerContext and DatabaseContext by calling respective getters, which run Bucket(), which seems to set some properties, so I think it's safer to recreate ctx to capture any changes.

rest/utilities_testing.go
cbbruno: yes, should be rt.Context(), will do.

rest/blip_sync.go
adamcfraser: Understood, wanted to make sure it wasn't passing a reference that was subsequently getting updated.  Have you been able to complete any testing (automated or manual) to confirm the behaviour?

base/dcp_sharded.go
```diff
@@ -343,9 +343,9 @@ func initCBGTManager(bucket Bucket, spec BucketSpec, cfgSG cbgt.Cfg, dbUUID stri
 		options)
 
 	cbgtContext := &CbgtContext{
-		loggingCtx: LogContextWith(context.Background(), &LogContext{CorrelationID: MD(spec.BucketName).Redact() + "-" + DCPImportFeedID}),
-		Manager:    mgr,
-		Cfg:        cfgSG,
+		// loggingCtx: LogContextWith(ctx, &LogContext{CorrelationID: MD(spec.BucketName).Redact() + "-" + DCPImportFeedID}),
```
adamcfraser: We were defining a logging context for DCP that included bucket name and feed ID.  I don't see that being done for the context that's passed into initCBGTManager - where is this log context now being added?

base/dcp_test.go
```diff
@@ -150,8 +150,9 @@ func TestCBGTIndexCreation(t *testing.T) {
 			spec := bucket.BucketSpec
 
 			// Use an in-memory cfg, set up cbgt manager
+			ctx := LogContextWith(TestCtx(t), &DatabaseLogContext{DatabaseName: tc.dbName})
```
adamcfraser, line 153: Why does the test need to add the DatabaseLogContext here, but not in other tests (like TestCBGTIndexCreationSafeLegacyName)?  And should we even have a DatabaseLogContext in the base package?

db/attachment_compaction_test.go
```diff
@@ -20,12 +21,13 @@ func TestAttachmentMark(t *testing.T) {
 	}
 
 	testDb := setupTestDB(t)
-	defer testDb.Close()
+	ctx := testDb.AddDatabaseLogContext(base.TestCtx(t))
```
adamcfraser: Assuming that most tests are going to want a context, what do you think about modifying setupTestDB to return (testDB, ctx)?  It would be nice if we didn't have to include this line for every test.

db/attachment_compaction_test.go
```diff
@@ -816,12 +827,13 @@ func TestAttachmentCompactIncorrectStat(t *testing.T) {
 	}
 
 	testDb := setupTestDB(t)
-	defer testDb.Close()
+	ctx1 := testDb.AddDatabaseLogContext(base.TestCtx(t))
```
adamcfraser: Why is this ctx1 and not ctx?

db/blip_connected_client.go
```diff
@@ -89,8 +90,8 @@ func (bh *blipHandler) handleFunction(rq *blip.Message) error {
 		return err
 	}
 	bh.logEndpointEntry(rq.Profile(), fmt.Sprintf("name: %s", name))
-	return bh.db.WithTimeout(UserQueryTimeout, func() error {
-		results, err := bh.db.CallUserFunction(name, requestParams, true)
+	return bh.db.WithTimeout(bh.changesCtx, UserQueryTimeout, func(ctx context.Context) error {
```
adamcfraser: I thought changesCtx was only used for cancellation - should this be bh.loggingCtx (here and below)

db/import_listener.go
```diff
@@ -28,6 +28,7 @@ type importListener struct {
 	stats            *base.DatabaseStats // Database stats group
 	cbgtContext      *base.CbgtContext   // Handle to cbgt manager,cfg
 	checkpointPrefix string              // DCP checkpoint key prefix
+	loggingCtx       context.Context     // ctx for logging, since function signatures without ctx are required by external dependency, for ex.: sgbucket.FeedEventCallbackFunc
```
adamcfraser: sgbucket is owned by us, so isn't an external dependency and could be changed.  If there are other cases (like cbgt), maybe mention those here instead.

db/database.go
```diff
@@ -660,7 +661,7 @@ func (context *DatabaseContext) SetOnChangeCallback(callback DocChangedFunc) {
 	context.mutationListener.OnDocChanged = callback
 }
 
-func (dbCtx *DatabaseContext) GetServerUUID() string {
+func (dbCtx *DatabaseContext) GetServerUUID(ctx context.Context) string {
```
adamcfraser, line 665: Do we need to do `ctx = context.AddDatabaseLogContext(ctx)` here?  I'm not clear on why that's needed in Close and NewDatabaseContext but wouldn't be needed for other public methods on *DatabaseContext.

adamcfraser: If it varies by usage, how does a developer adding a new method to *DatabaseContext know whether they should be adding the database log context or not?

db/sg_replicate_cfg.go
```diff
@@ -422,9 +422,9 @@ func (ar *ActiveReplicator) alignState(targetState string) error {
 
 }
 
-func (dbc *DatabaseContext) StartReplications() {
+func (dbc *DatabaseContext) StartReplications(ctx context.Context) {
 	if dbc.Options.SGReplicateOptions.Enabled {
-		base.DebugfCtx(dbc.SGReplicateMgr.loggingCtx, base.KeyReplicate, "Will start Inter-Sync Gateway Replications for database %q", dbc.Name)
+		base.DebugfCtx(ctx, base.KeyReplicate, "Will start Inter-Sync Gateway Replications for database %q", dbc.Name)
```
adamcfraser: Until we determine whether we're removing it, this should continue to use the SGReplicateMgr loggingCtx, as it's got custom CorrelationID configured.

rest/handler.go
```diff
@@ -164,6 +164,10 @@ func newHandler(server *ServerContext, privs handlerPrivs, r http.ResponseWriter
 		startTime:    time.Now(),
 		runOffline:   runOffline,
 	}
+
+	h.rqCtx = base.LogContextWith(h.rq.Context(), &base.LogContext{CorrelationID: h.formatSerialNumber()})
```
adamcfraser: Do we need to set this here if we call h.ctx() consistently elsewhere?

adamcfraser (changes requested):
Overall approach is looking good - a few questions on the details.

base/dcp_sharded.go
cbbruno: my bad, I totally missed finishing this and removing the comments, thanks... I will added logging ctx after init, after actually creating the cbgt ctx, since error on init might make sense to be on db context.

base/dcp_test.go
cbbruno: - added here bc dbname varies in each test case (unlike TestCBGTIndexCreationSafeLegacyName) 
- think we talked before about location of these custom log contexts.  Since addPrefixes is in base and need to know about these custom ctx keys (we don't have a self registering mechanism now), if I move it to /db I end up with circular dependency.  It would also be nice to have a /logging to separate interfaces, key registration, and help to avoid dependency, but that would cause more changes.

db/attachment_compaction_test.go
cbbruno: makes a lot of sense now looking back, will do.

db/attachment_compaction_test.go
cbbruno: sure, will do.

db/blip_connected_client.go
cbbruno: good question... think now that we're using the same initial ctx everywhere, we might want unify loggingCtx and changesCtx?  I will end up using the passed ctx for both... see NewBlipSyncContext and where it is invoked:

- https://github.com/couchbase/sync_gateway/blob/1c09dfcc4f7191087df606fd0b5bb3d5519ef4c6/db/blip_sync_context.go#L35
- https://github.com/couchbase/sync_gateway/blob/f141467f4a04891943284abe523e230c35e065cf/db/active_replicator.go#L216
- https://github.com/couchbase/sync_gateway/blob/117e565ab5de68dea18c5bc880c806b7bbd0e8a0/rest/blip_sync.go#L49

db/database.go
cbbruno: honestly, think the AddDatabaseLogContext on Close is unnecessary now, leftover from my initial changes 2 weeks ago, removing... and we should be adding db ctx only on NewDatabaseContext and tests.

db/import_listener.go
cbbruno: oh ok, not sure how/if cbgt is used here yet, but I'll note this is for event callbacks, thanks.

db/sg_replicate_cfg.go
cbbruno: ok, will do.

rest/blip_sync.go
cbbruno: compared to master as suggested in chat.

rest/handler.go
cbbruno: sure, will do.

db/blip_connected_client.go
adamcfraser: I think we need to keep them separate.  changesCtx is specifically for cancellation, and only affects a subset of blip functionality.  bh.loggingCtx should continue to be used for logging context.  If we want to combine cancellation and logging context in future (across the entire product) we can do so, but I'd prefer to keep these separate until then.

adamcfraser (approved):

PR #5705, opened 2022-08-17
Allow DropIndex to work on collections

torcolvin:
This prevents errors on DropIndexes in tests in db packages. This is not exercised until we do tests on named collections, but it works for non named collections.

## Pre-review checklist
- [x] Removed debug logging (`fmt.Print`, `log.Print`, ...)
- [x] Logging sensitive data? Make sure it's tagged (e.g. `base.UD(docID)`, `base.MD(dbName)`)
- [x] Updated relevant information in the API specifications (such as endpoint descriptions, schemas, ...) in `docs/api`

## [Integration Tests](https://jenkins.sgwdev.com/job/SyncGateway-Integration/build?delay=0sec)
- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/654/

PR #5726, opened 2022-08-30
Remove gocb/v1 tests

torcolvin:
This is a side effect of #5705 because I would have need to change `CouchbaseBucketGoCB.EscapedKeyspace()` like I did for `Collection.EscapedKeyspace()`.

`CouchbaseBucketGoCB` can not be used in collections, and I do not believe it is currently used. This test prevents GSI=true from working right now.
 
## Pre-review checklist

- [x] Link upstream PRs

I don't have an integration test build because of other issues, but it passes the GSI base/db tests.

### DIFF
--- a/base/util_testing.go
+++ b/base/util_testing.go
@@ -660,11 +660,6 @@ func ForAllDataStores(t *testing.T, testCallback func(*testing.T, sgbucket.DataS
 		})
 	}
 
-	dataStores = append(dataStores, dataStore{
-		name:   "gocb.v1",
-		driver: GoCBCustomSGTranscoder,
-	})
-
 	for _, dataStore := range dataStores {
 		t.Run(dataStore.name, func(t *testing.T) {
 			bucket := GetTestBucketForDriver(t, dataStore.driver)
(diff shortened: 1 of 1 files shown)


base/util_testing.go
```diff
@@ -660,11 +660,6 @@ func ForAllDataStores(t *testing.T, testCallback func(*testing.T, sgbucket.DataS
 		})
 	}
 
-	dataStores = append(dataStores, dataStore{
-		name:   "gocb.v1",
-		driver: GoCBCustomSGTranscoder,
-	})
```
bbrks: This change has unintended side-effects when running in `Walrus`, because `GetTestBucketForDriver` is doing this:

```go
	// If walrus, use bucket as-is
	if !TestUseCouchbaseServer() {
		return &TestBucket{
			Bucket:     bucket,
			BucketSpec: spec,
			closeFn:    closeFn,
		}
	}
```

Which is now never going to be run because of the mutually exclusive `TestUseCouchbaseServer` check around gocbv2.
I think I'd be OK with removing `TestUseCouchbaseServer` check above, although it still feels a bit like another footgun waiting to happen and don't have a better fix off the top of my head.

bbrks (changes requested):

base/util_testing.go
torcolvin: Run these tests with walrus, sort of. 

I'd propose a code cleanup to remove {{ForAllDataStores}} as a future commit.

bbrks (approved):

PR #5731, opened 2022-08-31
CBG-2330: Collection-aware import listener

markspolakovs:
CBG-2330

Add collection awareness to the ImportListener.

* Moves the `dbContext.Scopes` initialization from `_getOrAddDatabaseFromConfig` into `NewDatabaseContext` because it needs to already be set when the import listeners are started.
* Currently uses the dbContext's bucket as the metadata store (pending CBG-2266)

## Pre-review checklist

- [ ] https://github.com/couchbase/sg-bucket/pull/81

- [ ] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/740/


base/collection.go
```diff
@@ -1068,6 +1069,72 @@ func (c *Collection) GetCollectionID() (uint32, error) {
...
+	return 0, false
+}
+
+func (c *Collection) GetCollectionManifest() (CollectionsManifest, error) {
+	agent, err := c.Collection.Bucket().Internal().IORouter()
+	if err != nil {
+		return CollectionsManifest{}, fmt.Errorf("failed to get gocbcore agent: %w", err)
+	}
+	result := make(chan any) // either a CollectionsManifest or error
+	_, err = agent.GetCollectionManifest(gocbcore.GetCollectionManifestOptions{}, func(res *gocbcore.GetCollectionManifestResult, err error) {
```
adamcfraser: I think we'll want to set Deadline in the GetCollectionManifestOptions (to c.getBucketOpDeadline()).

base/collection.go
```diff
@@ -1068,6 +1069,72 @@ func (c *Collection) GetCollectionID() (uint32, error) {
...
+	UID  string `json:"uid"`
+}
+
+type ScopeManifest struct {
+	Name        string               `json:"name"`
+	UID         string               `json:"uid"`
+	Collections []CollectionManifest `json:"collections"`
+}
+
+type CollectionsManifest struct {
```
adamcfraser: It looks like gocbcore has a public definition of the GetCollectionsManifest response format already (Manifest/ManifestScope/ManifestCollection).  Can we use that, or is there a reason to define our own?
(it comes with an UnmarshalJSON to do the UID string->int conversions, which is a bonus)

base/dcp_dest.go
```diff
@@ -105,7 +106,7 @@ func (d *DCPDest) DataUpdate(partition string, key []byte, seq uint64,
 	if !dcpKeyFilter(key) {
 		return nil
 	}
-	event := makeFeedEventForDest(key, val, cas, partitionToVbNo(partition), 0, 0, sgbucket.FeedOpMutation)
+	event := makeFeedEventForDest(key, val, cas, partitionToVbNo(partition), collectionIDFromExtras(extras), 0, 0, sgbucket.FeedOpMutation)
```
adamcfraser, line 109: I don't think we should be getting any DataUpdate calls from cbgt (since we implement DestEx).  Is there a scenario where DataUpdate will be getting called where we need to use collectionIDFromExtras, or is this just defensive?

base/dcp_receiver.go
```diff
@@ -94,7 +94,7 @@ func (r *DCPReceiver) DataDelete(vbucketId uint16, key []byte, seq uint64,
 
 // Make a feed event for a gomemcached request.  Extracts expiry from extras
 func makeFeedEventForMCRequest(rq *gomemcached.MCRequest, opcode sgbucket.FeedOpcode) sgbucket.FeedEvent {
-	return makeFeedEvent(rq.Key, rq.Body, rq.DataType, rq.Cas, ExtractExpiryFromDCPMutation(rq), rq.VBucket, opcode)
+	return makeFeedEvent(rq.Key, rq.Body, rq.DataType, rq.Cas, ExtractExpiryFromDCPMutation(rq), rq.VBucket, 0, opcode)
```
adamcfraser, line 97: Should we be filing a followup ticket to rip out the cbgt gocouchbase support?  I think this could be removed at that point.

base/dcp_sharded.go
```diff
@@ -29,7 +29,7 @@ import (
 const CBGTIndexTypeSyncGatewayImport = "syncGateway-import-"
 const DefaultImportPartitions = 16
 
-// firstVersionToSupportCollections represents the earliest Sync Gateway release that supports collections (codename "Helium").
+// firstVersionToSupportCollections represents the earliest Sync Gateway release that supports collections (codename "Elixir").
```
adamcfraser: I think I'd just remove the codename from the comment altogether

db/import_listener.go
```diff
@@ -42,37 +44,60 @@ func NewImportListener(groupID string) *importListener {
 // Writes DCP stats into the StatKeyImportDcpStats map
 func (il *importListener) StartImportFeed(bucket base.Bucket, dbStats *base.DbStats, dbContext *DatabaseContext) (err error) {
 	il.bucketName = bucket.GetName()
-	il.database = Database{DatabaseContext: dbContext, user: nil}
+	il.dbCtx = dbContext
```
adamcfraser: It looks like the only thing we're using dbCtx for is dbName and stats (in import_pindex), now that we're not using bucket, and we already have databaseStats (although not import stats) on importListener.  

Does it make sense to just include dbName and the stats references on the importListener struct, and remove dbCtx?  It might minimize future cases of unreleased databaseContexts.

db/import_listener.go
```diff
@@ -42,37 +44,60 @@ func NewImportListener(groupID string) *importListener {
...
+			collectionNamesByScope[scopeName] = append(collectionNamesByScope[scopeName], collName)
+
+			collID, ok := collectionManifest.GetIDForCollection(scopeName, collName)
+			if !ok {
+				return fmt.Errorf("failed to find ID for collection %s.%s", base.MD(scopeName).Redact(), base.MD(collName).Redact())
+			}
+			il.collections[collID] = Database{DatabaseContext: collCtx.CollectionCtx, user: nil}
+		}
+	} else {
+		il.collections[0x0] = Database{DatabaseContext: dbContext, user: nil}
```
adamcfraser: Is `0x0` needed here as opposed to just `0`?

adamcfraser (changes requested):
Overall looks good, some questions on the details.

base/dcp_dest.go
markspolakovs: It's defensive - I went back and forth on whether to implement it, given as you say cbgt should only ever be calling DataUpdateEx.

base/dcp_receiver.go
markspolakovs: Filed CBG-2402

base/collection.go
markspolakovs: Good catch - I'll switch to using it, though still define our own alias for it so we have the `GetIDForCollection` utility method.

base/collection.go
markspolakovs: Should probably also `waitForAvailableKvOp`/`releaseKvOp`.

adamcfraser (approved):

PR #5733, opened 2022-09-01
CBG-2247: Etags should be quoted

gregns1:
CBG-2247

Added functionality to quote ETags when setting them. Then also Unquoting If-Match headers when they are used. 
- Changed admin_api.go to set quoted ETags then Unquote them when getting If-Match
- doc_api.go changed to unquote If-Match headers 
- admin_api_test.go has been updated to reflect these changes along side api_test.go 

## Pre-review checklist

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/870/

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -230,12 +230,7 @@ func (h *handler) handleGetDbConfig() error {
 			return base.HTTPErrorf(http.StatusNotFound, "database config not found")
 		}
 
-		quoteETag, err := h.setEtag(dbConfig.Version)
-		if err != nil {
-			return err
-		}
-		h.response.Header().Set("ETag", quoteETag)
-
+		h.setEtag(dbConfig.Version)
 		// refresh_config=true forces the config loaded out of the bucket to be applied on the node
 		if h.getBoolQuery("refresh_config") && h.server.bootstrapContext.connection != nil {
 			// set cas=0 to force a refresh
@@ -536,13 +531,7 @@ func (h *handler) handlePutDbConfig() (err error) {
 				return nil, err
 			}
 
-			headerVersion := h.rq.Header.Get("If-Match")
-			headerVersion, err = h.readETag(headerVersion)
-			if err != nil {
-				return nil, err
-			}
-
-			if headerVersion != "" && headerVersion != bucketDbConfig.Version {
+			if h.headerDoesNotMatchEtag("If-Match", bucketDbConfig.Version) {
 				return nil, base.HTTPErrorf(http.StatusPreconditionFailed, "Provided If-Match header does not match current config version")
 			}
 
@@ -603,11 +592,8 @@ func (h *handler) handlePutDbConfig() (err error) {
 	}
 	// store the cas in the loaded config after a successful update
 	h.server.dbConfigs[dbName].cas = cas
-	quoteETag, err := h.setEtag(updatedDbConfig.Version)
-	if err != nil {
-		return err
-	}
-	h.response.Header().Set("ETag", quoteETag)
+	h.setEtag(updatedDbConfig.Version)
+
 	return base.HTTPErrorf(http.StatusCreated, "updated")
 
 }
@@ -636,11 +622,7 @@ func (h *handler) handleGetDbConfigSync() error {
 		}
 	}
 
-	quoteETag, err := h.setEtag(etagVersion)
-	if err != nil {
-		return err
-	}
-	h.response.Header().Set("ETag", quoteETag)
+	h.setEtag(etagVersion)
 	h.writeJavascript(syncFunction)
 	return nil
 }
@@ -664,13 +646,7 @@ func (h *handler) handleDeleteDbConfigSync() error {
 				return nil, err
 			}
 
-			headerVersion := h.rq.Header.Get("If-Match")
-			headerVersion, err = h.readETag(headerVersion)
-			if err != nil {
-				return nil, err
-			}
-
-			if headerVersion != "" && headerVersion != bucketDbConfig.Version {
+			if h.headerDoesNotMatchEtag("If-Match", bucketDbConfig.Version) {
 				return nil, base.HTTPErrorf(http.StatusPreconditionFailed, "Provided If-Match header does not match current config version")
 			}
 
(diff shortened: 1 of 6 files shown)


rest/admin_api.go
```diff
@@ -230,7 +231,8 @@ func (h *handler) handleGetDbConfig() error {
 			return base.HTTPErrorf(http.StatusNotFound, "database config not found")
 		}
 
-		h.response.Header().Set("ETag", dbConfig.Version)
+		quoteETag := strconv.Quote(dbConfig.Version)
```
bbrks: `strconv.Quote/Unquote` does more than wrap the given string in double quotes - which may not be quite what we need for Etag headers.

rest/admin_api.go
```diff
@@ -621,7 +628,8 @@ func (h *handler) handleGetDbConfigSync() error {
 		}
 	}
 
-	h.response.Header().Set("ETag", etagVersion)
+	quoteETag := strconv.Quote(etagVersion)
+	h.response.Header().Set("ETag", quoteETag)
```
bbrks: It may make sense to pull the logic out into a common `h.setEtag(etagVersion)` function so all handlers don't have to know the intricacies of quoting Etag values.

rest/admin_api.go
```diff
@@ -646,6 +654,10 @@ func (h *handler) handleDeleteDbConfigSync() error {
 			}
 
 			headerVersion := h.rq.Header.Get("If-Match")
+			if headerVersion != "" {
+				headerVersion, _ = strconv.Unquote(headerVersion)
+			}
```
bbrks: Similarly, a common method on `handler` to read an If-Match header/etag version

bbrks (changes requested):
Some initial feedback, but I would actually hold off on doing anything more on this until we know what's happening with the Etag handling changes in #5556

rest/admin_api.go
```diff
@@ -230,7 +230,11 @@ func (h *handler) handleGetDbConfig() error {
 			return base.HTTPErrorf(http.StatusNotFound, "database config not found")
 		}
 
-		h.response.Header().Set("ETag", dbConfig.Version)
+		quoteETag, err := h.setEtag(dbConfig.Version)
+		if err != nil {
+			return err
+		}
+		h.response.Header().Set("ETag", quoteETag)
```
bbrks: I expected `setEtag` to set the response header, not return the value for each handler to be manually setting it, given they're methods on a handler.

rest/handler.go
```diff
@@ -824,6 +824,26 @@ func (h *handler) userAgentIs(agent string) bool {
 	return len(userAgent) > len(agent) && userAgent[len(agent)] == '/' && strings.HasPrefix(userAgent, agent)
 }
 
+func (h *handler) setEtag(eTag string) (string, error) {
+	// Checks to see if the string has already been quoted. If so return an error message
+	if len(eTag) > 0 && strings.Contains(eTag, `"`) {
+		return "", fmt.Errorf("ETag value is already quoted when trying to set it")
```
bbrks: Seems more accurate 

```suggestion
		return "", fmt.Errorf("ETag values must not contain a quote")
```

rest/admin_api.go
```diff
@@ -533,6 +537,11 @@ func (h *handler) handlePutDbConfig() (err error) {
 			}
 
 			headerVersion := h.rq.Header.Get("If-Match")
+			headerVersion, err = h.readETag(headerVersion)
```
bbrks: Similarly, I think `readETag` should be getting the header value, rather than making every handler have to know how to use `If-Match`

rest/admin_api_test.go
```diff
@@ -3350,7 +3351,8 @@ func TestPersistentConfigConcurrency(t *testing.T) {
 	getETag := resp.Header.Get("ETag")
 	assert.Equal(t, putETag, getETag)
 
-	resp = bootstrapAdminRequestWithHeaders(t, http.MethodPost, "/db/_config", "{}", map[string]string{"If-Match": "x"})
+	quotedStr := `"` + "x" + `"`
```
bbrks: this is equivalent

```suggestion
	quotedStr := `"x"`
```

bbrks (changes requested):

rest/admin_api.go
```diff
@@ -532,8 +531,7 @@ func (h *handler) handlePutDbConfig() (err error) {
 				return nil, err
 			}
 
-			headerVersion := h.rq.Header.Get("If-Match")
-			if headerVersion != "" && headerVersion != bucketDbConfig.Version {
+			if h.headerDoesNotMatchEtag("If-Match", bucketDbConfig.Version) {
```
bbrks: I know this is existing code - but I wonder if we should move `"If-Match"` into the function, as I don't see a case where we'd want the header name to be anything different.

rest/doc_api.go
```diff
@@ -393,7 +397,10 @@ func (h *handler) handlePutDoc() error {
 		bodyRev := body[db.BodyRev]
 		if oldRev := h.getQuery("rev"); oldRev != "" {
 			body[db.BodyRev] = oldRev
-		} else if ifMatch := h.rq.Header.Get("If-Match"); ifMatch != "" {
+		} else if ifMatch, err := h.getEtag("If-Match"); ifMatch != "" {
+			if err != nil {
```
bbrks: `getEtag` always returns an `""` value with the error, so this error handling would never be hit.

In this particular case, given it's a rev ID, we know it should have a specific format to match, so SG is going to fail further down if the provided value is quoted or otherwise wrong anyway, so I think we can just ignore the error.

rest/doc_api.go
```diff
@@ -451,7 +458,10 @@ func (h *handler) handlePutDocReplicator2(docid string, roundTrip bool) (err err
 	var parentRev string
 	if oldRev := h.getQuery("rev"); oldRev != "" {
 		parentRev = oldRev
-	} else if ifMatch := h.rq.Header.Get("If-Match"); ifMatch != "" {
+	} else if ifMatch, err := h.getEtag("If-Match"); ifMatch != "" {
+		if err != nil {
```
bbrks: same comment here about error handling

bbrks (changes requested):
Some minor comments but looks fine to merge once those are done.

bbrks (approved):

PR #5736, opened 2022-09-02
CBG-2315: Disallow changing scope after DB creation

markspolakovs:
CBG-2315

Validate that the `scopes` property of the DB config cannot be changed after initial creation.

## Pre-review checklist

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/755/
  * Likely flake.
- [x] `.*Collections.*` https://jenkins.sgwdev.com/job/SyncGateway-Integration/766/

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -533,7 +533,7 @@ func (dbConfig *DbConfig) validatePersistentDbConfig() (errorMessages error) {
 }
 
 // validateConfigUpdate combines the results of validate and validateChanges.
-func (dbConfig *DbConfig) validateConfigUpdate(ctx context.Context, old *db.DatabaseContext, validateOIDCConfig bool) error {
+func (dbConfig *DbConfig) validateConfigUpdate(ctx context.Context, old DbConfig, validateOIDCConfig bool) error {
 	err := dbConfig.validate(ctx, validateOIDCConfig)
 	var multiErr *base.MultiError
 	if !errors.As(err, &multiErr) {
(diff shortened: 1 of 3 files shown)


rest/config.go
```diff
@@ -532,6 +532,39 @@ func (dbConfig *DbConfig) validatePersistentDbConfig() (errorMessages error) {
...
+	if !errors.As(err, &multiErr) {
+		multiErr = multiErr.Append(err)
+	}
+	multiErr = multiErr.Append(dbConfig.validateChanges(ctx, old))
+	return multiErr.ErrorOrNil()
+}
+
+// validateChanges compares the current DbConfig with the "old" config, and returns an error if any disallowed changes
+// are attempted.
+func (dbConfig *DbConfig) validateChanges(ctx context.Context, old *db.DatabaseContext) error {
```
markspolakovs: I'm not sure about this accepting a `*DatabaseContext` - ideally it'd keep around a reference to the old `DbConfig` but that may require a more extensive refactor.

rest/config.go
```diff
@@ -532,6 +532,39 @@ func (dbConfig *DbConfig) validatePersistentDbConfig() (errorMessages error) {
...
+			newScope = name
+		}
+		for name := range old.Scopes {
+			oldScope = name
+		}
+		if newScope != oldScope {
+			return fmt.Errorf("cannot change scopes after database creation")
+		}
+	}
+	return nil
```
bbrks: Would be nice if we can use logic that doesn't bake-in assumptions about what's a valid scope count. If we have two sets of different scopes but the last items in each happen to match, then this logic would incorrectly "pass".

bbrks (changes requested):

bbrks (approved):

PR #5739, opened 2022-09-02
CBG-2242: fix errors.As in tests

gregns1:
CBG-2242

Small changes to config_test.go to fix issues with Errors.As in new version of go. Using testify package seems to fix the issue. 

## Pre-review checklist

- [ ] N/A

### DIFF
--- a/rest/config_test.go
+++ b/rest/config_test.go
@@ -28,6 +28,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 
@@ -1656,7 +1657,16 @@ func TestLoadJavaScript(t *testing.T) {
 			}()
 			js, err := loadJavaScript(inputJavaScriptOrPath, test.insecureSkipVerify)
 			if test.errExpected != nil {
-				require.True(t, assert.ErrorAs(t, err, &test.errExpected))
+				errx509Type := x509.UnknownAuthorityError{}
+				if test.errExpected == errx509Type {
+					expectedErrorString := test.errExpected.Error()
+					if runtime.GOOS == "darwin" {
+						expectedErrorString = "certificate is not trusted"
+					}
+					require.ErrorContains(t, err, expectedErrorString)
+				} else {
+					require.ErrorContains(t, err, test.errExpected.Error())
+				}
 			}
 			assert.Equal(t, test.jsExpected, js)
 		})
@@ -1812,16 +1822,18 @@ func TestSetupDbConfigWithSyncFunction(t *testing.T) {
 					RemoteConfigTlsSkipVerify: true,
 				}
 			}
-			if test.errExpected != nil {
-				test.errExpected = &JavaScriptLoadError{
-					JSLoadType: SyncFunction,
-					Path:       sync,
-					Err:        test.errExpected,
-				}
-			}
 			err := dbConfig.setup(dbConfig.Name, BootstrapConfig{}, nil, nil, false)
 			if test.errExpected != nil {
-				require.True(t, assert.ErrorAs(t, err, &test.errExpected))
+				errx509Type := x509.UnknownAuthorityError{}
+				if test.errExpected == errx509Type {
+					expectedErrorString := test.errExpected.Error()
+					if runtime.GOOS == "darwin" {
+						expectedErrorString = "certificate is not trusted"
+					}
+					require.ErrorContains(t, err, expectedErrorString)
+				} else {
+					require.ErrorContains(t, err, test.errExpected.Error())
+				}
 			} else {
 				assert.Equal(t, test.jsSyncFnExpected, *dbConfig.Sync)
 			}
@@ -1912,16 +1924,18 @@ func TestSetupDbConfigWithImportFilterFunction(t *testing.T) {
 					RemoteConfigTlsSkipVerify: true,
 				}
 			}
-			if test.errExpected != nil {
-				test.errExpected = &JavaScriptLoadError{
-					JSLoadType: ImportFilter,
-					Path:       importFilter,
-					Err:        test.errExpected,
-				}
-			}
 			err := dbConfig.setup(dbConfig.Name, BootstrapConfig{}, nil, nil, false)
 			if test.errExpected != nil {
-				require.True(t, assert.ErrorAs(t, err, &test.errExpected))
+				errx509Type := x509.UnknownAuthorityError{}
+				if test.errExpected == errx509Type {
+					expectedErrorString := test.errExpected.Error()
+					if runtime.GOOS == "darwin" {
+						expectedErrorString = "certificate is not trusted"
+					}
+					require.ErrorContains(t, err, expectedErrorString)
+				} else {
+					require.ErrorContains(t, err, test.errExpected.Error())
+				}
 			} else {
 				assert.Equal(t, test.jsImportFilterExpected, *dbConfig.ImportFilter)
 			}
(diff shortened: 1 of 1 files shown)


rest/config_test.go
```diff
@@ -1656,7 +1656,7 @@ func TestLoadJavaScript(t *testing.T) {
 			}()
 			js, err := loadJavaScript(inputJavaScriptOrPath, test.insecureSkipVerify)
 			if test.errExpected != nil {
-				require.True(t, errors.As(err, &test.errExpected)) // nolint: govet // CBG-2242
+				require.True(t, assert.ErrorAs(t, err, &test.errExpected))
```
bbrks: This fixes the Go vet error but doesn't actually fix the underlying complaint that it's an incorrect usage of `errors.As` (where err is being assigned to `test.errExpected`)

Switching this to `require.ErrorIs` does seem to be causing other issues, but that's primarily because the error types being specified (e.g. `x509.UnknownAuthorityError`/`JavaScriptLoadError`) don't implement the `Unwrap` or `Is` interfaces, so cannot actually assert on the underlying error type in this way.

I would advise reworking the test such that it's doing a direct string comparison on the error message, instead of trying to use error type checking. It's simpler and works if we're able to determine the error message output in the test cases.

```go
require.EqualError(t, err, test.errExpected.Error())
```

or 

```go
require.ErrorContains(t, err, test.errExpectedString)
```

bbrks (changes requested):
This one was a deeper rabbit hole than anticipated, but I left a suggestion inline.

rest/config_test.go
```diff
@@ -1656,7 +1657,16 @@ func TestLoadJavaScript(t *testing.T) {
...
 			js, err := loadJavaScript(inputJavaScriptOrPath, test.insecureSkipVerify)
 			if test.errExpected != nil {
-				require.True(t, errors.As(err, &test.errExpected)) // nolint: govet // CBG-2242
+				errx509Type := x509.UnknownAuthorityError{}
+				if test.errExpected == errx509Type {
+					expectedErrorString := test.errExpected.Error()
+					if runtime.GOOS == "darwin" {
+						expectedErrorString = "certificate is not trusted"
+					}
+					require.ErrorContains(t, err, expectedErrorString)
```
bbrks: Can we put this into a test helper function so it's shared for all the tests? e.g. `assertX509UnknownAuthority(err)`

Still _some_ amount of duplication to do the type switching on the expected error, but that's ok.

bbrks (changes requested):

rest/config_test.go
```diff
@@ -1656,7 +1657,12 @@ func TestLoadJavaScript(t *testing.T) {
...
 			if test.errExpected != nil {
-				require.True(t, errors.As(err, &test.errExpected)) // nolint: govet // CBG-2242
+				errx509Type := x509.UnknownAuthorityError{}
+				if test.errExpected == errx509Type {
+					require.ErrorContains(t, err, assertX509UnknownAuthority(test.errExpected))
+				} else {
+					require.ErrorContains(t, err, test.errExpected.Error())
+				}
 			}
 			assert.Equal(t, test.jsExpected, js)
```
bbrks: It looks like this test never made sure that we get no error when `test.ErrExpected == nil`

Maybe an extra `else` to cover this? Similar applies for all of the other tests we're touching. below

```suggestion
			} else {
				require.NoError(t, err)
				assert.Equal(t, test.jsExpected, js)
			}
```

rest/config_test.go
```diff
@@ -1812,16 +1818,14 @@ func TestSetupDbConfigWithSyncFunction(t *testing.T) {
...
-			}
 			err := dbConfig.setup(dbConfig.Name, BootstrapConfig{}, nil, nil, false)
 			if test.errExpected != nil {
-				require.True(t, errors.As(err, &test.errExpected)) // nolint: govet // CBG-2242
+				errx509Type := x509.UnknownAuthorityError{}
+				if test.errExpected == errx509Type {
+					require.ErrorContains(t, err, assertX509UnknownAuthority(test.errExpected))
+				} else {
+					require.ErrorContains(t, err, test.errExpected.Error())
+				}
```
bbrks: It still feels odd to have the x509 type checking stuff duplicated in every test.

How about extending `assertX509UnknownAuthority` to cover a non-x509 error but still takes the x509 error type into consideration?

```suggestion
				requireErrorWithX509UnknownAuthority(t, err, test.errExpected.Error())
```

```go
func requireErrorWithX509UnknownAuthority(t testing.TB, actual, expected error) {
	expectedErrorString := expected.Error()

	switch typedError := expected.(type) {
	case x509.UnknownAuthorityError:
		expectedErrorString = typedError.Error()
		if runtime.GOOS == "darwin" {
			expectedErrorString = "certificate is not trusted"
		}
	}

	require.ErrorContains(t, actual, expectedErrorString)
}
```

bbrks (changes requested):
One more suggestion that I can can tidy up the x509 error handling code a bit more that I'd missed as an option in my previous comment. But otherwise looks good.

rest/config_test.go
gregns1: This is actually something similar to what I originally wrote for the last commit but wasn't sure if you wanted me to take the `require.ErrorContains(t, actual, expectedErrorString)` out of the main body of the test into the helper function. I agree this way is much better I'll make the changes now and get the PR updated.

rest/config_test.go
```diff
@@ -2128,18 +2110,11 @@ func TestWebhookFilterFunctionLoad(t *testing.T) {
...
-					Err:        test.errExpected,
-				}
-			}
 			ctx := &db.DatabaseContext{EventMgr: db.NewEventManager()}
 			sc := &ServerContext{}
 			err := sc.initEventHandlers(ctx, &dbConfig)
 			if test.errExpected != nil {
-				require.True(t, errors.As(err, &test.errExpected)) // nolint: govet // CBG-2242
+				assertX509UnknownAuthority(t, err, test.errExpected)
 			}
```
bbrks, line 2125: else no error

rest/config_test.go
```diff
@@ -1812,16 +1815,9 @@ func TestSetupDbConfigWithSyncFunction(t *testing.T) {
...
-					Path:       sync,
-					Err:        test.errExpected,
-				}
-			}
 			err := dbConfig.setup(dbConfig.Name, BootstrapConfig{}, nil, nil, false)
 			if test.errExpected != nil {
-				require.True(t, errors.As(err, &test.errExpected)) // nolint: govet // CBG-2242
+				assertX509UnknownAuthority(t, err, test.errExpected)
 			} else {
 				assert.Equal(t, test.jsSyncFnExpected, *dbConfig.Sync)
```
bbrks, line 1823: as mentioned in previous comment, do the `require.NoError` in these other else blocks

bbrks (changes requested):
Just the other "else NoError" things to add then good to merge

rest/config_test.go
```diff
@@ -2619,3 +2599,15 @@ func TestCollectionsValidation(t *testing.T) {
 		})
 	}
 }
+
+func assertX509UnknownAuthority(t testing.TB, actual, expected error) {
```
bbrks: I'd rename to `require` to match the underlying behaviour (for context: `require.` stops test execution at the failure point, `assert.` continues running the test code but will still record it as a failed test)

```suggestion
func requireX509UnknownAuthority(t testing.TB, actual, expected error) {
```

bbrks: I'd also tweak the name to make it account for the non-x509 error handling we're doing, and leave a comment describing how it covers both (and why we even need this function in the first place, due to the OS-specific errors)

```suggestion
// requireErrorWithX509UnknownAuthority ... TODO
func requireErrorWithX509UnknownAuthority(t testing.TB, actual, expected error) {
```

bbrks (changes requested):
Last comment to tweak naming and add a comment, then good to merge.

Let me or any of the others know if you need a hand with resolving the conflict.

bbrks (approved):

PR #5747, opened 2022-09-07
Disable test with xattrs

torcolvin:
This test doesn't write `_sync` docs inline, but they can be tested with named collections when all tests are using them.

### DIFF
--- a/rest/api_collections_test.go
+++ b/rest/api_collections_test.go
@@ -107,7 +107,9 @@ func TestCollectionsPutDocInKeyspace(t *testing.T) {
 
 func TestCollectionsDCP(t *testing.T) {
 	base.TestRequiresCollections(t)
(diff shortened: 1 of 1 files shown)


rest/api_collections_test.go
```diff
@@ -107,7 +107,9 @@ func TestCollectionsPutDocInKeyspace(t *testing.T) {
 
 func TestCollectionsDCP(t *testing.T) {
 	base.TestRequiresCollections(t)
-	base.SetUpTestLogging(t, base.LevelDebug, base.KeyHTTP, base.KeyDCP, base.KeyImport)
+	if !base.TestUseXattrs() {
+		t.Skip("Test does not provide inline _sync data for non xattrs")
```
bbrks:
```suggestion
		t.Skip("Test relies on import - needs xattrs")
```

bbrks (approved):

PR #5758, opened 2022-09-14
Bump log level/keys in TestReplicatorDoNotSendDeltaWhenSrcIsTombstone

bbrks:
This test fails semi-regularly but we do not have verbose enough detail in the logs to determine if a replication race condition is causing the failure condition.

- [ ] n/a

### DIFF
--- a/rest/replicator_test.go
+++ b/rest/replicator_test.go
@@ -5750,7 +5750,7 @@ func TestReplicatorDoNotSendDeltaWhenSrcIsTombstone(t *testing.T) {
 	base.RequireNumTestBuckets(t, 2)
 
 	defer db.SuspendSequenceBatching()()
-	base.SetUpTestLogging(t, base.LevelDebug, base.KeyHTTP, base.KeyHTTPResp, base.KeyChanges, base.KeySync, base.KeySyncMsg, base.KeyWebSocket, base.KeyWebSocketFrame, base.KeyDCP)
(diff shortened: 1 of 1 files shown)


rest/replicator_test.go
```diff
@@ -5750,7 +5750,7 @@ func TestReplicatorDoNotSendDeltaWhenSrcIsTombstone(t *testing.T) {
 	base.RequireNumTestBuckets(t, 2)
 
 	defer db.SuspendSequenceBatching()()
-	base.SetUpTestLogging(t, base.LevelInfo, base.KeyAll)
+	base.SetUpTestLogging(t, base.LevelDebug, base.KeyHTTP, base.KeyHTTPResp, base.KeyChanges, base.KeySync, base.KeySyncMsg, base.KeyWebSocket, base.KeyWebSocketFrame, base.KeyDCP)
```
IsaacLambat: Would it be useful to also have `KeyReplicate` here to see logs about conflicts and resolutions of them conflicts?

IsaacLambat (changes requested):

rest/replicator_test.go
bbrks: At this point there's not much I'm filtering out, so I'll just replace with All

```suggestion
	base.SetUpTestLogging(t, base.LevelDebug, base.KeyAll)
```

IsaacLambat (approved):
LGTM

PR #5767, opened 2022-09-16
CBG-2400 split some packages out of rest

torcolvin:
CBG-2400

- Created two new packages in rest/attachmentcompactiontest and adminapitest.
- Moved test code that was in test only parts of rest to utility packages `rest/utilities_testing_*.go`

## Pre-review checklist

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/835/

### DIFF
--- a/rest/attachmentcompactiontest/main_test.go
+++ b/rest/attachmentcompactiontest/main_test.go
@@ -1,0 +1,22 @@
+/*
+Copyright 2020-Present Couchbase, Inc.
+
+Use of this software is governed by the Business Source License included in
+the file licenses/BSL-Couchbase.txt.  As of the Change Date specified in that
+file, in accordance with the Business Source License, use of this software will
+be governed by the Apache License, Version 2.0, included in the file
+licenses/APL2.txt.
+*/
+
+package attachmentcompactiontest
+
+import (
+	"testing"
+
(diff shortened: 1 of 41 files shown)


rest/attachmentcompactiontest/main_test.go
```diff
@@ -0,0 +1,22 @@
...
+package attachmentcompactiontest
+
+import (
+	"testing"
+
+	"github.com/couchbase/sync_gateway/db"
+)
+
+func TestMain(m *testing.M) {
+	memWatermarkThresholdMB := uint64(2048)
```
adamcfraser, line 20: How were the thresholds for the new test packages selected?

adamcfraser (commented):
Changes look fine, just one point of clarification about how the new test packages interact with memWatermarkThreshold.

rest/attachmentcompactiontest/main_test.go
torcolvin: I started at 256 and doubled them until they passed with a buffer.

The watermark actually probably should be defined by the number of buckets available to the tests - so at 3 buckets these takes take a little more than 1GB. In fact, the watermark is totally different when run with walrus vs integration and I didn't want the tests to fail for this reason.

bbrks (approved):

PR #5769, opened 2022-09-19
CBG-2400 Switch bucket pool to create a non default collection

torcolvin:
- Use SG_TEST_USE_COLLECTION_POOL to switch between using non default collections by default. If this flag isn't set, tests requiring named collections will not run. This only applies when using GSI.
- Create a named collection on a bucket as well as the default collection. After storing the type of collection received by `base.GetTestBucket` and its variants, clean that collection by using primary index to drop all items.
- Some tests require using default collections (e.g. tests that read persistent config). There may be some tests that are writing to the wrong collection by accident.
- After slightly changing tbpClusterV2 interface, remove tbpClusterV1 which is only compatible with gocb/v1

This is intertwined with #5767, because GSI tests will not work robustly without that commit.

## Pre-review checklist

- [x] Link upstream PRs

- [x] `xattrs=true,gsi=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/853/
- [x] `xattrs=true,gsi=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/859/ (known issue https://issues.couchbase.com/browse/CBG-2433, https://issues.couchbase.com/browse/CBG-2435)
- [x] `xattrs=false,gsi=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/858/  (known issue https://issues.couchbase.com/browse/CBG-2433, https://issues.couchbase.com/browse/CBG-2435)
- [x] `xattrs=false,gsi=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/856/

### DIFF
--- a/rest/adminapitest/admin_api_test.go
+++ b/rest/adminapitest/admin_api_test.go
@@ -4015,7 +4015,7 @@ func TestDeleteFunctionsWhileDbOffline(t *testing.T) {
 	}()
 
 	// Get a test bucket, and use it to create the database.
-	// FIXME: Does this make sense it would require the default collection?
+	// FIXME: CBG-2266 this test reads in persistent config
 	tb := base.GetTestBucketDefaultCollection(t)
 	defer func() { tb.Close() }()
 
@@ -4225,7 +4225,7 @@ func TestGroupIDReplications(t *testing.T) {
 
 	base.SetUpTestLogging(t, base.LevelInfo, base.KeyAll)
 
-	// FIXME: Is requiring a default collection normal for this test?
+	// FIXME: CBG-2266 this test reads in persistent config
 
 	// Create test buckets to replicate between
 	passiveBucket := base.GetTestBucketDefaultCollection(t)
--- a/rest/adminapitest/admin_api_test.go
+++ b/rest/adminapitest/admin_api_test.go
@@ -3288,7 +3288,8 @@ func TestPersistentConfigConcurrency(t *testing.T) {
 	resp = rest.BootstrapAdminRequest(t, http.MethodGet, "/db/_config", "")
 	resp.RequireStatus(http.StatusOK)
 	eTag := resp.Header.Get("ETag")
-	assert.NotEqual(t, "", eTag)
+	unquoteETag := strings.Trim(eTag, `"`)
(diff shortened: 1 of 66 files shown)


rest/adminapitest/admin_api_test.go
```diff
@@ -4015,7 +4015,8 @@ func TestDeleteFunctionsWhileDbOffline(t *testing.T) {
 	}()
 
 	// Get a test bucket, and use it to create the database.
-	tb := base.GetTestBucket(t)
+	// FIXME: Does this make sense it would require the default collection?
+	tb := base.GetTestBucketDefaultCollection(t)
```
bbrks: These and the other FIXMEs I'm not immediately sure of the answer to - but is probably something we need to figure out?

- It's either something the test is doing that doesn't support collections, in which case we can remove the fixme
- Something the bucket/collection pool approach doesn't solve, in which case we'll have to figure out where that gap is and how to close it
- or some other weirdness that needs investigation (in which case we could file tickets?)

FIXMEs are almost certainly never going to be revisited without a means of tracking/reminding.

rest/blip_api_collections_test.go
```diff
@@ -16,24 +16,32 @@ func TestBlipGetCollections(t *testing.T) {
 	// FIXME as part of CBG-2203 to enable subtest checkpointExistsWithErrorInNonDefaultCollection
 	base.TestRequiresCollections(t)
```
bbrks, line 17: Unrelated but CBG-2203 (and the blocking LeakyBucket enhancements) I believe are done but there's nothing tracking the work to enable the subtest and it wasn't revisited as part of those tickets.

Do we need another ticket for test coverage of this case now it's possible to write a test for?

bbrks (changes requested):

rest/adminapitest/admin_api_test.go
torcolvin: These all rely subtly on persistent config and I'm going to change these to be changed in CBG-2266.

rest/blip_api_collections_test.go
torcolvin: I'm going to look at this now.

bbrks (approved):

PR #5770, opened 2022-09-19
Function+GraphQL API v0.2

snej:
- Moved the code into a new package, 'functions'. The 'db' package calls it via interfaces defined in 'db/functions.go'
- Combined functions and queries, per spec. Now just called 'functions'.
- JS API changes per spec
- JSON config changes per spec
- Implemented recursion depth limit (20)
- Moved the JavaScript code into a separate '.js' file


db/functions.go
```diff
@@ -0,0 +1,109 @@
...
+const UserFunctionTimeout = 60 * time.Second
+
+//////// USER FUNCTIONS
+
+// A map from names to user functions.
+type UserFunctions = map[string]UserFunction
+
+// A JavaScript function or N1QL query that can be invoked by a client.
+// (Created by functions.CompileUserFunction or functions.CompileUserFunctions)
+type UserFunction = interface {
```
adamcfraser:
```suggestion
type UserFunction interface {
```

functions/function.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+func (fn *functionImpl) AllowByDefault(allow bool) {
+	fn.allowByDefault = allow
+}
+
+// Creates an Invocation of a UserFunction.
+func (fn *functionImpl) Invoke(db *db.Database, args map[string]any, mutationAllowed bool, ctx context.Context) (db.UserFunctionInvocation, error) {
+	if ctx == nil {
+		panic("missing context to UserFunction.Invoke")
```
adamcfraser: I know this isn't new code in this PR, but missed this in previous review.  There are numerous uses of panic inside the functions code.  Where are these being recovered?  Failed function invocation shouldn't halt Sync Gateway, so would like to understand the model being used for panic recovery within functions.

functions/function.go
```diff
@@ -0,0 +1,289 @@
...
+Copyright 2022-Present Couchbase, Inc.
+
+Use of this software is governed by the Business Source License included in
+the file licenses/BSL-Couchbase.txt.  As of the Change Date specified in that
+file, in accordance with the Business Source License, use of this software will
+be governed by the Apache License, Version 2.0, included in the file
+licenses/APL2.txt.
+*/
+
+package functions
```
adamcfraser: I think the main motivation for a separate functions package is code organization/isolation?  If that's accurate, what's your opinion on defining this package as db/functions instead of a root level functions?  I don't think we anticipate any functions that operate outside of the context of a database, so it seems better from an organizational perspective.

It also seems like the main source of import recursion (necessitating the db/functions.go interfaces) is in the db package's blip code - having db/functions might lead us toward a db/blipSync in future that would allow us to break that recursion and remove that interface code.

adamcfraser (changes requested):
Generally looks good - question on panic handling and package organization.

functions/function.go
snej: That sounds like a good idea. I'll move the package.

functions/function.go
snej: This panic is not recoverable; it's an assertion. It just pre-empts the null-pointer panic you'd get soon thereafter if ctx is nil, and is easier to understand and diagnose.

There are two other similar users of `panic` in js_runner.go. If you have an existing assertion function, let me know and I'll use that instead.

The other uses of `panic` are in js_callbacks.go. They trigger JavaScript exceptions, in Otto callbacks.

adamcfraser (approved):

functions/function.go
adamcfraser: Agree with this based on the existing panics in js_runner.go.  Can consider replacing this with a non-fatal assert in future.

PR #5774, opened 2022-09-20
CBG-2362: Identify whether SG is running in persistent config mode (or not) via REST API

gregns1:
CBG-2362

Simple change to root response struct to include persistent cofig boolean to indicate whether it is running in persistent config mode or not. 

## Pre-review checklist

- [ ] N/A

### DIFF
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -2145,3 +2145,45 @@ Status:
         Blank if `api.hide_product_version=true` in the startup configuration.
       type: string
   title: Status
+NodeInfo:
+  type: object
+  properties:
+    ADMIN:
+      description: '`true` if the request is from the Admin API - otherwise omitted.'
+      type: boolean
+      example: true
+    couchdb:
+      description: CouchDB welcome
+      type: string
+      example: Welcome
+    vendor:
+      description: Product vendor
+      type: object
+      properties:
+        name:
+          description: Product name
+          type: string
+          example: Couchbase Sync Gateway
+        version:
+          description: |-
+            API version.
+            Omitted if `api.hide_product_version=true`
+          type: string
+          example: 3.1
+      required:
+        - name
+    version:
+      description: |-
+        Product version, including the build number and edition (i.e. `EE` or `CE`)
+        Omitted if `api.hide_product_version=true`
+      type: string
+      example: Couchbase Sync Gateway/3.1.0(1;a765231) EE
+    PersistentConfig:
+      description: |-
+        Indication for whether sync gateway is running in persistent config mode or legacy config mode.
+        `true` if the sync gateway node is running in persistent config mode.
+      type: boolean
+      example: true
+  required:
+    - couchdb
+    - vendor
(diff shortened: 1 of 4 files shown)


rest/api.go
```diff
@@ -54,6 +55,7 @@ func (h *handler) handleRoot() error {
 		Vendor: vendor{
 			Name: base.ProductNameString,
 		},
+		PersistentConfig: h.server.persistentConfig,
```
bbrks, line 58: Can we update the API documentation in `docs/api/paths/admin/~.yaml` and `docs/api/paths/public/~.yaml` to reflect this change?

The response schema seems to be copied across both files for some reason. So if you want to do a bit of a clean up job, you can move the response schema into a separate file (`docs/api/components/schemas.yaml`) and reference it from both `~.yaml` files like most of the other response bodies.

bbrks (changes requested):
Can we update the API documentation along with this change? Left details inline.

rest/api.go
```diff
@@ -35,10 +35,11 @@ const (
...
 type rootResponse struct {
-	Admin   bool   `json:"ADMIN,omitempty"`
-	CouchDB string `json:"couchdb,omitempty"` // TODO: Lithium - remove couchdb welcome
-	Vendor  vendor `json:"vendor,omitempty"`
-	Version string `json:"version,omitempty"`
+	Admin            bool   `json:"ADMIN,omitempty"`
+	CouchDB          string `json:"couchdb,omitempty"` // TODO: Lithium - remove couchdb welcome
+	Vendor           vendor `json:"vendor,omitempty"`
+	Version          string `json:"version,omitempty"`
+	PersistentConfig bool   `json:"persistentConfig,omitempty"`
```
bbrks: We typically use `snake_case` for JSON properties in Sync Gateway.

```suggestion
	PersistentConfig bool   `json:"persistent_config,omitempty"`
```

docs/api/components/schemas.yaml
```diff
@@ -2145,3 +2145,45 @@ Status:
...
+          example: 3.1
+      required:
+        - name
+    version:
+      description: |-
+        Product version, including the build number and edition (i.e. `EE` or `CE`)
+        Omitted if `api.hide_product_version=true`
+      type: string
+      example: Couchbase Sync Gateway/3.1.0(1;a765231) EE
+    PersistentConfig:
```
```suggestion
    persistent_config:
```

bbrks (changes requested):
Minor comment about `camelCase` -> `snake_case` on JSON property then is good to merge

rest/api.go
```diff
@@ -39,7 +39,7 @@ type rootResponse struct {
 	CouchDB          string `json:"couchdb,omitempty"` // TODO: Lithium - remove couchdb welcome
 	Vendor           vendor `json:"vendor,omitempty"`
 	Version          string `json:"version,omitempty"`
-	PersistentConfig bool   `json:"persistentConfig,omitempty"`
+	PersistentConfig bool   `json:"persistent_config,omitempty"`
```
bbrks: I missed one thing in my last review - `omitempty` on a bool will remove the field in the response if the value is `false`, which ideally we'd not want to do here.

```suggestion
	PersistentConfig bool   `json:"persistent_config"`
```

bbrks (changes requested):

bbrks (approved):

PR #5778, opened 2022-09-21
CBG-2356 use a test name for GroupID

torcolvin:
This allows for parametrized tests that might share a test bucket to use unique credentials, such as `TestServerlessDBSetupForceCreds`. 

## Pre-review checklist

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/877/ (fails due to #5782)

### DIFF
--- a/rest/utilities_testing.go
+++ b/rest/utilities_testing.go
@@ -181,7 +181,7 @@ func (rt *RestTester) Bucket() base.Bucket {
 		// If running in persistent config mode, the database has to be manually created. If the db name is the same as a
 		// past tests db name, a db already exists error could happen if the past tests bucket is still flushing. Prevent this
 		// by setting the group ID as the current test bucket name by default.
-		sc.Bootstrap.ConfigGroupID = testBucket.GetName()
+		sc.Bootstrap.ConfigGroupID = rt.TB.Name()
 	}
 
 	sc.Unsupported.UserQueries = base.BoolPtr(rt.EnableUserQueries)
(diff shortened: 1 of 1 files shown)


rest/utilities_testing.go
```diff
@@ -181,7 +181,7 @@ func (rt *RestTester) Bucket() base.Bucket {
 		// If running in persistent config mode, the database has to be manually created. If the db name is the same as a
 		// past tests db name, a db already exists error could happen if the past tests bucket is still flushing. Prevent this
 		// by setting the group ID as the current test bucket name by default.
```
IsaacLambat: Comment needs updating to reflect it now sets the group ID to the test name.

rest/utilities_testing.go
```diff
@@ -181,7 +181,7 @@ func (rt *RestTester) Bucket() base.Bucket {
...
-		sc.Bootstrap.ConfigGroupID = testBucket.GetName()
+		sc.Bootstrap.ConfigGroupID = rt.TB.Name()
```
IsaacLambat: The max group ID length in `(sc *StartupConfig) Validate` (done on line 194) is set to 100. This might cause problems if a subtest has a long name then tests will start failing. 

100 characters is a fair few but it could cause problems in the future. Could there be a way to prevent the name being over 100 characters if it is (such as using a hash maybe)?

IsaacLambat (changes requested):

rest/utilities_testing.go
torcolvin: The good news is that I tested that it errors here, but I agree I should change it.

IsaacLambat (approved):
LGTM. Would be good to get full, successful Jenkins run before merging as well as final review from Adam

PR #5783, opened 2022-09-23
Rewrite sgcollect in go

markspolakovs:
Rewrite sgcollect in go.

* Will almost certainly break the build, as it'll expect Python sgcollect to be present.
* The sgcollect code intentionally doesn't import modules from the existing SG packages (eg. utilities in `base` or type definitions from `rest`). This is to keep the binary size down, as all those packages' dependencies will be imported even if they're not used.

## Pre-review checklist


torcolvin:
Closing this PR since we aren't planning to do this anytime soon - https://jira.issues.couchbase.com/browse/CBG-4526 will track this.

PR #5784, opened 2022-09-23
CBG-1893: DELETE attachment endpoint

gregns1:
CBG-1893

- Added a new DELETE endpoint to give the functionality to delete an attachment from a document without going through the PUT attachment endpoint
- Added some more functionality to the existing test for testing document attachments
- Updated the api documentation to reflect the changes being made to the endpoints

## Pre-review checklist

- [ ] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/886/

### DIFF
--- a/docs/api/paths/public/{keyspace}~{docid}~{attach}.yaml
+++ b/docs/api/paths/public/{keyspace}~{docid}~{attach}.yaml
@@ -152,7 +152,7 @@ delete:
       schema:
         type: string
   responses:
-    '201':
+    '200':
       description: Attachment removed from the document successfully
       headers:
         Etag:
--- a/docs/api/paths/public/{keyspace}~{docid}~{attach}.yaml
+++ b/docs/api/paths/public/{keyspace}~{docid}~{attach}.yaml
@@ -141,12 +141,6 @@ delete:
 
     If the attachment exists, the attachment will be removed from the document.
   parameters:
-    - name: Content-Type
-      in: header
-      description: The content type of the attachment.
-      schema:
-        type: string
-        default: application/octet-stream
     - name: rev
       in: query
       description: The existing document revision ID to modify.
--- a/docs/api/paths/public/{keyspace}~{docid}~{attach}.yaml
+++ b/docs/api/paths/public/{keyspace}~{docid}~{attach}.yaml
@@ -71,8 +71,6 @@ put:
     If the attachment already exists, the data of the existing attachment will be replaced in the new revision.
 
     The maximum content size of an attachment is 20MB. The `Content-Type` header of the request specifies the content type of the attachment.
-
-    To remove an attachment from a document, omit the attachment in the `_attachments` object of the document in the `PUT /{db}/{docid}` request.
   parameters:
     - name: Content-Type
       in: header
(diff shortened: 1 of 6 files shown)


docs/api/paths/public/{keyspace}~{docid}~{attach}.yaml
```diff
@@ -128,3 +126,44 @@ head:
...
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/schemas.yaml#/New-revision
+    '404':
+      $ref: ../../components/responses.yaml#/Not-found
+    '409':
+      $ref: ../../components/responses.yaml#/Conflict
+  tags:
+    - Document
```
IsaacLambat: Add a blank line at the end of the file to match with the rest of the docs

rest/doc_api.go
```diff
@@ -356,6 +356,62 @@ func (h *handler) handlePutAttachment() error {
...
 }
 
+func (h *handler) handleDeleteAttachment() error {
+	if h.isReadOnlyGuest() {
+		return base.HTTPErrorf(http.StatusForbidden, auth.GuestUserReadOnly)
+	}
+
+	docid := h.PathVar("docid")
+	attachmentName := h.PathVar("attach")
+	attachmentContentType := h.rq.Header.Get("Content-Type")
```
IsaacLambat: Content type is not used or needed for a delete so this can be removed, as well as from the docs

IsaacLambat (changes requested):

rest/doc_api.go
gregns1: Yeah good point, I will get this updated in the morning.

rest/api_test.go
```diff
@@ -253,6 +255,18 @@ func TestDocAttachment(t *testing.T) {
...
+	response = rt.SendRequest("DELETE", "/db/doc/attach2?rev="+revid, "")
+	RequireStatus(t, response, 404)
+
+	// delete the attachment calling the delete attachment endpoint
+	response = rt.SendRequest("DELETE", "/db/doc/attach1?rev="+revid, "")
+	RequireStatus(t, response, 201)
+
+	// attempt to access deleted attachment (should return error)
+	response = rt.SendRequest("GET", "/db/doc/attach1", "")
+	RequireStatus(t, response, 404)
```
adamcfraser, line 277: I think a test to 'attempt to delete an attachment with an incorrect revid' should be included, to verify it's returning a 409.

rest/doc_api.go
```diff
@@ -356,6 +356,58 @@ func (h *handler) handlePutAttachment() error {
...
+	delete(attachments, attachmentName)
+	body[db.BodyAttachments] = attachments
+
+	newRev, _, err := h.db.Put(h.ctx(), docid, body)
+	if err != nil {
+		return err
+	}
+	h.setEtag(newRev)
+
+	h.writeRawJSONStatus(http.StatusCreated, []byte(`{"id":`+base.ConvertToJSONString(docid)+`,"ok":true,"rev":"`+newRev+`"}`))
```
adamcfraser: I think a DELETE operation should return a 200 (http.StatusOK), and not a 201.

rest/doc_api.go
```diff
@@ -356,6 +356,58 @@ func (h *handler) handlePutAttachment() error {
...
+			return base.HTTPErrorf(http.StatusConflict, "Cannot modify attachments without a specific rev ID")
+		}
+	}
+
+	// get document attachments and check if attachment exists
+	attachments := db.GetBodyAttachments(body)
+	if _, ok := attachments[attachmentName]; !ok {
+		return base.HTTPErrorf(http.StatusNotFound, "Attachment %s is not found", attachmentName)
+	}
+	// delete specified attachment from the map
```
adamcfraser, line 400: There's quite a bit of shared code between this and PUT attachment - it looks like the next two lines are the only thing that differs.  Did you evaluate whether it made sense to move some of the shared functionality to a shared function?  (it might be valid to keep it separate, just want to understand the reasoning)

adamcfraser (changes requested):

rest/doc_api.go
gregns1: I think you raise a good point. My reasoning for not having a shared helper function was that some of the shared code areas have some small differences. For example the code at the top is not pulling content type header as its not needed. The code after fetching the body, when getting document not found, checks if the error is relating to incorrect revid or not and returns the appropriate status code (an issue I spotted after your comment about adding test case for specifying incorrect revid for delete endpoint). Then the code to check if the attachment exists on the document or not. 
I am probably wrong with my thinking on that though, do you think I should separate the lines 360 to 393 into a separate shared function and get a way of executing the slightly different logic based off if the function is being called from PUT or DELETE endpoints?

adamcfraser: I think there are enough differences there that it's worth just leaving them without shared code - thanks for the additional information.

adamcfraser (approved):

PR #5791, opened 2022-09-29
CBG-2417: Investigate usages of prometheus.MustRegister

gregns1:
CBG-2417

Added some error handling for prometheus when swapping use of prometheus.MustRegister for prometheus.Register and passing errors up to be handled. Changes include a few changes to the way stats are set up in the tests due to adding the errors in. 

## Pre-review checklist

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/913/

### DIFF
--- a/base/stats.go
+++ b/base/stats.go
@@ -13,6 +13,7 @@ package base
 import (
 	"context"
 	"expvar"
+	"log"
 	"math"
 	"strconv"
 	"strings"
@@ -65,7 +66,7 @@ type SgwStats struct {
 	dbStatsMapMutex sync.Mutex
 }
 
-var SyncGatewayStats SgwStats
+var SyncGatewayStats *SgwStats
 
 // SkipPrometheusStatsRegistration used to avoid registering a stat with prometheus
 // Defaults to false - Only intended for test use
@@ -73,31 +74,44 @@ var SyncGatewayStats SgwStats
 // global stats are unaffected
 var SkipPrometheusStatsRegistration bool
 
-func NewSyncGatewayStats() *SgwStats {
+func NewSyncGatewayStats() (*SgwStats, error) {
 	sgwStats := SgwStats{
 		GlobalStats: &GlobalStat{},
 		DbStats:     map[string]*DbStats{},
 	}
 
-	sgwStats.GlobalStats.initResourceUtilizationStats()
-	sgwStats.initReplicationStats()
+	err := sgwStats.GlobalStats.initResourceUtilizationStats()
+	if err != nil {
+		return nil, err
+	}
+	err = sgwStats.initReplicationStats()
+	if err != nil {
+		return nil, err
+	}
 
 	// This provides a stat for sgw_up where the value will be fixed to one. This is to allow backwards compatibility
 	// where the standalone exporter would export a value of 1 if it has contact with SGW.
-	NewIntStat("", "up", nil, nil, prometheus.GaugeValue, 1)
+	_, err = NewIntStat("", "up", nil, nil, prometheus.GaugeValue, 1)
+	if err != nil {
+		return nil, err
+	}
 
-	return &sgwStats
+	return &sgwStats, nil
 }
 
 func init() {
 	// Initialize Sync Gateway Stats
 
 	// All stats will be stored as part of this struct. Global variable accessible everywhere. To add stats see stats.go
-	SyncGatewayStats = *NewSyncGatewayStats()
+	sgwStats, err := NewSyncGatewayStats()
+	if err != nil {
+		log.Fatalf("Fatal error: %v", err)
+	}
+	SyncGatewayStats = sgwStats
 
 	// Publish our stats to expvars. This will run String method on SyncGatewayStats ( type SgwStats ) which will
 	// marshal the stats to JSON
-	expvar.Publish(StatsGroupKeySyncGateway, &SyncGatewayStats)
+	expvar.Publish(StatsGroupKeySyncGateway, SyncGatewayStats)
 }
 
 // This String() is to satisfy the expvar.Var interface which is used to produce the expvar endpoint output.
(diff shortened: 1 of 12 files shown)


base/stats.go
```diff
@@ -65,39 +66,52 @@ type SgwStats struct {
...
 func init() {
 	// Initialize Sync Gateway Stats
 
 	// All stats will be stored as part of this struct. Global variable accessible everywhere. To add stats see stats.go
-	SyncGatewayStats = *NewSyncGatewayStats()
+	sgwStats, err := NewSyncGatewayStats()
+	if err != nil {
+		log.Fatalf("Fatal error: %v", err)
+	}
+	SyncGatewayStats = sgwStats
```
JRascagneres: any reason to do that. Can we just directly set SyncGatewayStats?

base/stats.go
```diff
@@ -116,40 +130,109 @@ type GlobalStat struct {
...
+func (s *SgwStats) initReplicationStats() error {
 	s.ReplicatorStats = &ReplicatorStats{
 		new(expvar.Map).Init(),
 	}
-	prometheus.MustRegister(s.ReplicatorStats)
+	err := prometheus.Register(s.ReplicatorStats)
+	if err != nil {
+		return err
+	}
+	return nil
```
JRascagneres: Can probably just do `return prometheus.Register(s.ReplicatorStats)`

base/stats.go
```diff
@@ -748,41 +840,68 @@ type QueryStat struct {
...
 	} else {
-		s.DbStats[name].InitQueryStats(
+		err := s.DbStats[name].InitQueryStats(
 			false,
 			queryNames...,
 		)
+		if err != nil {
+			return nil, err
+		}
 	}
```
JRascagneres: Can maybe avoid an error check here. Inside `ifs` rather than doing `err :=` do `err =` and then can error check outside of the if else

db/channel_cache_single_test.go
```diff
@@ -32,7 +32,14 @@ func TestDuplicateDocID(t *testing.T) {
...
 
-	cache := newSingleChannelCache(context, "Test1", 0, (base.NewSyncGatewayStats()).NewDBStats("", false, false, false).Cache())
+	stats, err := base.NewSyncGatewayStats()
+	require.NoError(t, err)
+
+	dbstats, err := stats.NewDBStats("", false, false, false)
+	require.NoError(t, err)
+
+	cache := newSingleChannelCache(context, "Test1", 0, dbstats.Cache())
+	assert.True(t, cache != nil)
```
JRascagneres: Think there is a not nil check. Not sure if we need to do it but better than `assert.True`

db/channel_cache_single_test.go
```diff
@@ -80,7 +87,14 @@ func TestLateArrivingSequence(t *testing.T) {
...
```
JRascagneres: Ditto

db/channel_cache_single_test.go
```diff
@@ -114,7 +128,14 @@ func TestLateSequenceAsFirst(t *testing.T) {
...
```
JRascagneres: Ditto

db/channel_cache_single_test.go
```diff
@@ -148,7 +169,14 @@ func TestDuplicateLateArrivingSequence(t *testing.T) {
...
```
JRascagneres: Ditto

db/channel_cache_single_test.go
```diff
@@ -224,7 +252,15 @@ func TestPrependChanges(t *testing.T) {
...
 	// 1. Test prepend to empty cache
-	cache := newSingleChannelCache(dbCtx, "PrependEmptyCache", 0, (base.NewSyncGatewayStats()).NewDBStats("", false, false, false).Cache())
+	stats, err := base.NewSyncGatewayStats()
+	require.NoError(t, err)
+
+	dbstats, err := stats.NewDBStats("", false, false, false)
+	require.NoError(t, err)
+
+	cache := newSingleChannelCache(dbCtx, "PrependEmptyCache", 0, dbstats.Cache())
+	assert.True(t, cache != nil)
```
JRascagneres: Ditto

db/database_test.go
```diff
@@ -548,7 +548,8 @@ func TestDeltaSyncWhenFromRevIsChannelRemoval(t *testing.T) {
 	require.NoError(t, err, "Error creating user")
 
 	db.user = user
-	db.DbStats.InitDeltaSyncStats()
+	err = db.DbStats.InitDeltaSyncStats()
+	require.NoError(t, err)
```
JRascagneres: For this and the below ones can be simplified. `requre.NoError(t, db.DbStats.InitDeltaSyncStats())`

base/stats.go
```diff
@@ -65,39 +66,52 @@ type SgwStats struct {
...
 }
 
 func init() {
 	// Initialize Sync Gateway Stats
 
 	// All stats will be stored as part of this struct. Global variable accessible everywhere. To add stats see stats.go
-	SyncGatewayStats = *NewSyncGatewayStats()
+	sgwStats, err := NewSyncGatewayStats()
+	if err != nil {
+		log.Fatalf("Fatal error: %v", err)
```
JRascagneres: Add a little more context. Maybe there is more context lower but nice to have. 
`Fatal error attempting to instantiate sync gateway stats: %v`. Something like that?

JRascagneres (changes requested):
Mostly nickpicks but a few comments

db/channel_cache_single_test.go
```diff
@@ -240,7 +276,7 @@ func TestPrependChanges(t *testing.T) {
 	require.Len(t, cachedChanges, 3)
 
 	// 2. Test prepend to populated cache, with overlap and duplicates
-	cache = newSingleChannelCache(dbCtx, "PrependPopulatedCache", 0, (base.NewSyncGatewayStats()).NewDBStats("", false, false, false).Cache())
+	cache = newSingleChannelCache(dbCtx, "PrependPopulatedCache", 0, dbstats.Cache())
```
JRascagneres, line 283: I think this (and the below) are fine... Just note that I think the behaviour is different, previously we were creating new stats, now re-using. If test passes tho its fine.

JRascagneres (approved):

PR #5800, opened 2022-10-03
CBG-2433: db/indexes_test.go fail without views enabled

gregns1:
CBG-2433

Changes to index test code to clear all index and reinitialize them at the start of each test. This avoids having the wrong indexes available to the test. 

## Pre-review checklist

- [ ] `GSI=true,xattrs=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/927/
- [ ] `GSI=true,xattrs=true` (for collections) https://jenkins.sgwdev.com/job/SyncGateway-Integration/928/

### DIFF
--- a/base/collection.go
+++ b/base/collection.go
@@ -204,6 +204,7 @@ func (c *Collection) UUID() (string, error) {
 	return config.BucketUUID(), nil
 }
 
+// GetCluster returns an open cluster object
 func (c *Collection) GetCluster() *gocb.Cluster {
 	return c.cluster
 }
--- a/base/collection.go
+++ b/base/collection.go
@@ -204,6 +204,10 @@ func (c *Collection) UUID() (string, error) {
 	return config.BucketUUID(), nil
 }
 
+func (c *Collection) GetCluster() *gocb.Cluster {
+	return c.cluster
+}
+
 func (c *Collection) Close() {
 	if c.cluster != nil {
 		if err := c.cluster.Close(nil); err != nil {
--- a/db/indexes_test.go
+++ b/db/indexes_test.go
@@ -71,30 +71,46 @@ func TestInitializeIndexes(t *testing.T) {
 func validateAllIndexesOnline(bucket base.Bucket, xattrs bool) error {
 
 	col, err := base.AsCollection(bucket)
+	if err != nil {
+		return err
+	}
 
 	cluster := col.GetCluster()
 	mgr := cluster.QueryIndexes()
 
 	watchOption := gocb.WatchQueryIndexOptions{
 		WatchPrimary:   true,
-		ScopeName:      "sg_test_1",
-		CollectionName: "sg_test_1",
+		ScopeName:      col.ScopeName(),
+		CollectionName: col.Name(),
 	}
 
 	// Watch and wait some time for indexes to come online
 	if xattrs {
-		err = mgr.WatchIndexes(bucket.GetName(), []string{"#primary", "sg_access_x1", "sg_allDocs_x1", "sg_channels_x1", "sg_roleAccess_x1", "sg_syncDocs_x1", "sg_tombstones_x1"}, 10*time.Second, &watchOption)
+		err = mgr.WatchIndexes(bucket.GetName(), listIndexes(xattrs), 10*time.Second, &watchOption)
 		if err != nil {
 			return err
 		}
 	} else {
-		err = mgr.WatchIndexes(bucket.GetName(), []string{"#primary", "sg_access_1", "sg_allDocs_1", "sg_channels_1", "sg_roleAccess_1", "sg_syncDocs_1"}, 10*time.Second, &watchOption)
+		err = mgr.WatchIndexes(bucket.GetName(), listIndexes(xattrs), 10*time.Second, &watchOption)
 		if err != nil {
 			return err
 		}
 	}
 
 	return nil
+}
+
+func listIndexes(xattrs bool) []string {
+	allSGIndexes := make([]string, 0)
+
+	for _, sgIndex := range sgIndexes {
+		fullIndexName := sgIndex.fullIndexName(xattrs)
+		if fullIndexName == "sg_tombstones_1" {
+			continue
+		}
+		allSGIndexes = append(allSGIndexes, fullIndexName)
+	}
+	return allSGIndexes
 }
 
 func TestPostUpgradeIndexesSimple(t *testing.T) {
--- a/db/indexes_test.go
+++ b/db/indexes_test.go
@@ -16,7 +16,9 @@ import (
 	"fmt"
 	"log"
 	"testing"
+	"time"
 
+	"github.com/couchbase/gocb/v2"
 	sgbucket "github.com/couchbase/sg-bucket"
 	"github.com/couchbase/sync_gateway/base"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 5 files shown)


db/indexes_test.go
```diff
@@ -57,53 +59,42 @@ func TestInitializeIndexes(t *testing.T) {
...
+	cluster := col.GetCluster()
+	mgr := cluster.QueryIndexes()
 
-	var indexRow struct {
-		Name  string
-		State string
+	watchOption := gocb.WatchQueryIndexOptions{
+		WatchPrimary:   true,
+		ScopeName:      "sg_test_1",
+		CollectionName: "sg_test_1",
```
torcolvin: you want to do to something like:

```
collection, err := base.AsCollection(bucket)
require.NoError(err)
```

then use `ScopeName: collection.ScopeName()` and `CollectionName: collection.Name()`

db/indexes_test.go
```diff
@@ -57,53 +59,42 @@ func TestInitializeIndexes(t *testing.T) {
...
 	}
 
-	for results.Next(&indexRow) {
-		if indexRow.State != base.IndexStateOnline {
-			return fmt.Errorf("Index %s is not online", indexRow.Name)
-		} else {
-			log.Printf("Validated index %s is %s", indexRow.Name, indexRow.State)
+	// Watch and wait some time for indexes to come online
+	if xattrs {
+		err = mgr.WatchIndexes(bucket.GetName(), []string{"#primary", "sg_access_x1", "sg_allDocs_x1", "sg_channels_x1", "sg_roleAccess_x1", "sg_syncDocs_x1", "sg_tombstones_x1"}, 10*time.Second, &watchOption)
```
torcolvin: Construct these like:


https://cs.github.com/couchbase/sync_gateway/blob/7e9bc42ecbf2198af01bda918448265772e231f6/db/indexes.go?q=path%3Aindexes.go#L292

just create a function to return all indexes names since it is used twice

torcolvin (commented):
In general, it still doesn't make sense that you'd need to do anything other than wait for indexes in one test, so having this in multiple tests is where you'd need to drop indexes is bad, because it puts load on the server. 

I'm definitely thinking about removing all these tests because we implicitly test the system in our integration test harness, and presumable QE tests index creation otherwise.

db/indexes_test.go
```diff
@@ -57,53 +59,58 @@ func TestInitializeIndexes(t *testing.T) {
...
-			log.Printf("Validated index %s is %s", indexRow.Name, indexRow.State)
+	return nil
+}
+
+func listIndexes(xattrs bool) []string {
+	allSGIndexes := make([]string, 0)
+
+	for _, sgIndex := range sgIndexes {
+		fullIndexName := sgIndex.fullIndexName(xattrs)
+		if fullIndexName == "sg_tombstones_1" {
```
torcolvin: use `sgInex.isXattrOnly` for this - you want to wait for this

db/indexes_test.go
```diff
@@ -119,6 +126,8 @@ func TestPostUpgradeIndexesSimple(t *testing.T) {
 
 	n1qlStore, ok := base.AsN1QLStore(db.Bucket)
 	assert.True(t, ok)
+	err := dropAndInitializeIndexes(base.TestCtx(t), n1qlStore, db.Bucket, db.UseXattrs())
```
torcolvin: can we drop this?

db/indexes_test.go
```diff
@@ -153,6 +162,9 @@ func TestPostUpgradeIndexesSimple(t *testing.T) {
 	// Restore indexes after test
 	err = InitializeIndexes(n1qlStore, db.UseXattrs(), 0, false)
 	assert.NoError(t, err)
+
+	validateErr := validateAllIndexesOnline(db.Bucket, db.UseXattrs())
```
torcolvin: can we drop this?

db/indexes_test.go
```diff
@@ -166,6 +178,8 @@ func TestPostUpgradeIndexesVersionChange(t *testing.T) {
 	require.True(t, db.Bucket.IsSupported(sgbucket.DataStoreFeatureN1ql))
 	n1qlStore, ok := base.AsN1QLStore(db.Bucket)
 	assert.True(t, ok)
+	err := dropAndInitializeIndexes(base.TestCtx(t), n1qlStore, db.Bucket, db.UseXattrs())
```
torcolvin: can we drop this?

db/indexes_test.go
```diff
@@ -311,3 +326,29 @@ func TestIsIndexerError(t *testing.T) {
 	err = errors.New("err:[5000]  MCResponse status=KEY_ENOENT, opcode=0x89, opaque=0")
 	assert.True(t, isIndexerError(err))
 }
+
+// Drop and reinitialize all indexes
```
torcolvin: traditional go style is // dropAndInitializeIndexes drops all sync gateway indexes

the style is to start with the function name first to get godoc to pick up

go.mod
```diff
@@ -9,8 +9,8 @@ require (
 	github.com/couchbase/clog v0.1.0
 	github.com/couchbase/go-blip v0.0.0-20220810102707-e9964d569180
 	github.com/couchbase/go-couchbase v0.1.1
-	github.com/couchbase/gocb/v2 v2.5.3-0.20220803131303-46b466983d0f
-	github.com/couchbase/gocbcore/v10 v10.1.5-0.20220809160836-bf53e9527651
+	github.com/couchbase/gocb/v2 v2.5.3
+	github.com/couchbase/gocbcore/v10 v10.1.5
```
torcolvin: I don't see why this needs to happy, does this go away with `go.mod` ? Otherwise I'd just set `git checkout go.mod go.sum`

db/indexes_test.go
```diff
@@ -57,53 +59,58 @@ func TestInitializeIndexes(t *testing.T) {
...
-				initErr := InitializeIndexes(n1qlStore, test.xattrs, 0, true)
-				require.NoError(t, initErr, "Error initializing all indexes")
-
-				// Recreate the primary index required by the test bucket pooling framework
-				err := n1qlStore.CreatePrimaryIndex(base.PrimaryIndexName, nil)
-				require.NoError(t, err)
-
-				validateErr := validateAllIndexesOnline(b)
-				require.NoError(t, validateErr, "Error validating indexes online")
+				err := dropAndInitializeIndexes(base.TestCtx(t), n1qlStore, b, test.xattrs)
```
bbrks, line 62: To summarise our conversation earlier this week:

- There's still value in having this test attempt to create our indexes with all combinations of xattrs and collections, as it will find index definition breakages _much_ earlier than the weekly Jenkins job, or QE (i.e. at dev time pre-merge)
- General approach should be to make this problematic test reset state back to the original (i.e. call `dropAndInitializeIndexes` at the end of each subtest, but use `base.TestUseXattrs()` instead of the overridden `test.xattrs` value)
- Pulling this test out into its own package for more isolation could be an option too, but that seems overkill and maybe only an absolute last resort.

base/collection.go
```diff
@@ -204,6 +204,10 @@ func (c *Collection) UUID() (string, error) {
 	return config.BucketUUID(), nil
 }
 
+func (c *Collection) GetCluster() *gocb.Cluster {
```
torcolvin, line 208: nit: I'd always document public functions on public objects 

```
// GetCluster returns an open cluster object

File: db/indexes_test.go [hunk 11]
```diff
@@ -57,53 +59,61 @@ func TestInitializeIndexes(t *testing.T) {
...
+		if err != nil {
+			return err
+		}
+	} else {
+		err = mgr.WatchIndexes(bucket.GetName(), listIndexes(xattrs), 10*time.Second, &watchOption)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
```
torcolvin: Both branches are the same here, linting catches this issue - gocritic / dupBranchBody. It might be worth asking folks who use goland if there's something that would tell them this.

db/indexes_test.go
```diff
@@ -57,53 +59,63 @@ func TestInitializeIndexes(t *testing.T) {
...
+}
+
+func listIndexes(xattrs bool) []string {
```
torcolvin: nit: I'd probably name this indexNames

// indexNames returns all the names of indexes for a collection

db/indexes_test.go
```diff
@@ -57,53 +59,63 @@ func TestInitializeIndexes(t *testing.T) {
...
-	var indexRow struct {
-		Name  string
-		State string
+	cluster := col.GetCluster()
+	mgr := cluster.QueryIndexes()
+
+	watchOption := gocb.WatchQueryIndexOptions{
+		WatchPrimary:   true,
+		ScopeName:      col.ScopeName(),
+		CollectionName: col.Name(),
```
torcolvin, line 89: This happily passes with CBS 6.6, which surprised me since it doesn't use collections. If you are adding code that adds gocb parameters for collections, it's worth trying to remember to pass through jenkins with couchbase server 6.6. We do run this weekly.

torcolvin (dismissed):
These changes look good, just some small naming changes suggested.

The one improvement I'd like to make is using this index waiting code in production, because we right now do an exponential backoff. This will become much easier to use when we no longer have gocb v1 code to support (https://issues.couchbase.com/browse/CBG-2339), so https://issues.couchbase.com/browse/CBG-2470 will push the implementation of waiting for indexes to gocb.

torcolvin (approved):

PR #5804, opened 2022-10-06
Configurable limits on function/graphQL count, size, etc.

snej:
Implemented changes in [API spec 0.3](https://docs.google.com/document/d/1ggxUJhotM0-Ww8CnM69HKBcKbboQyw5VeWLUkdp8sUA/edit?usp=sharing):
- Functions config is now a struct not a map
  - The map is now its Definitions field
  - Added optional limit properties to the struct
- Added limit properties to GraphQLConfig
- Initialization code checks the limits, aside from `max_request_size`
- REST and BLIP clients check the `max_request_size` limit

## Pre-review checklist

- [x] Link upstream PRs

- [ ] `GSI=true,xattrs=true` (for collections) https://jenkins.sgwdev.com/job/SyncGateway-Integration/000/


snej:
It looks like the one EE unit test that failed is a pre-existing failure? It's not related to anything I changed.

adamcfraser:
Kicked off integration tests here:
https://jenkins.sgwdev.com/job/SyncGateway-Integration/985/

rest/functions_api.go
```diff
@@ -56,15 +60,26 @@ func (h *handler) handleFunctionCall() error {
...
 	if h.rq.Method == "POST" {
 		// POST: Args come from the request body in JSON format:
+		if h.rq.ContentLength >= 0 {
+			if err := db.CheckRequestSize(h.rq.ContentLength, maxSize); err != nil {
+				return "", nil, err
+			}
+		}
 		err = h.readJSONInto(&args)
+		if err == nil && h.rq.ContentLength < 0 && maxSize != nil {
+			err = db.CheckRequestSize(int64(db.EstimateSizeOfJSON(args)), maxSize)
```
adamcfraser: It feels relatively cumbersome to estimate the size of JSON by iterating over elements in map[string]interface{}{}, when we just unmarshalled the raw bytes for that JSON a few lines earlier. 

json.Decoder has an InputOffset property that can be read post-Decode - could you use that to add a version of readJSONInto that returned bytes read?

adamcfraser (changes requested):
Changes generally look good - one question about whether we can avoid EstimateSizeOfJSON.

db/functions.go
```diff
@@ -90,10 +98,52 @@ func (db *Database) CallUserFunction(name string, args map[string]interface{}, m
...
+	}
+}
+
+// Utility that returns a rough estimate of the original size of the JSON a value was parsed from.
+func EstimateSizeOfJSON(args any) int {
+	switch arg := args.(type) {
+	case bool:
+		return 4
+	case int64:
+		return 4
```
bbrks, line 127: 8 bytes?

db/functions.go
```diff
@@ -90,10 +98,52 @@ func (db *Database) CallUserFunction(name string, args map[string]interface{}, m
...
+	case string:
+		return len(arg) + 2
+	case []any:
+		size := 2
+		for _, item := range arg {
+			size += EstimateSizeOfJSON(item) + 1
+		}
+		return size
+	case map[string]any:
+		size := 0
```
bbrks: Wouldn't an empty object be 2 bytes to start with? (`{}`)

db/functions.go
```diff
@@ -90,10 +98,52 @@ func (db *Database) CallUserFunction(name string, args map[string]interface{}, m
...
 		return nil, base.HTTPErrorf(http.StatusServiceUnavailable, "GraphQL is not configured")
+	} else {
+		return graphql.Query(db, query, operationName, variables, mutationAllowed, ctx)
+	}
+}
+
+//////// UTILITIES
+
+// Returns an HTTP 413 error if `maxSize` is non-nil and less than `actualSize`.
+func CheckRequestSize[T int | int64](actualSize T, maxSize *int) error {
```
bbrks, line 111: Not sure why generics are needed here, especially since we're always doing a type conversion to `int64` inside

db/functions.go
```diff
@@ -90,10 +98,52 @@ func (db *Database) CallUserFunction(name string, args map[string]interface{}, m
...
+			size += EstimateSizeOfJSON(item) + 1
+		}
+		return size
+	case map[string]any:
+		size := 0
+		for key, item := range arg {
+			size += len(key) + EstimateSizeOfJSON(item) + 4
+		}
+		return size
+	default:
```
bbrks, line 146:
```suggestion
	case nil:
		return 4
	default:
```

rest/functions_api.go
snej: OK. It means using json.Decoder directly instead of going through those functions, but the other stuff they do is mostly not relevant, like checking for unknown fields. I did have to add the Content-Type check.

Q: The existing code uses base.JSONDecoder(), which returns an interface base.JSONDecoderI that does not support the InputOffset method. It looks like this is done to support pluggable JSON implementations?

db/functions.go
snej: Nope; this function roughly estimates the length of the JSON encoding, not the size of a binary value. The string `true` is 4 bytes and `false` is 5; I simplified to just return 4. An integer is variable length, but rather than having to do logarithms or build a string, I'm just handwaving 4 as being a conservative average length.

This function is admittedly sloppy (it's just for detecting cases where the size goes crazy, likely due to abuse), but it needs to be fast because it's going to be run before every function call.

db/functions.go
snej: Good point.

rest/functions_api.go
adamcfraser: That's a good point - you may need to switch to using the native library directly, as the other implementation of JSONDecoderI doesn't have support for InputOffset (despite a pending PR https://github.com/json-iterator/go/pull/493).

db/functions.go
snej: For the convenience of the caller. Go is too damn picky to promote an int to an int64 😝 so if I make the parameter `int64` I have to add casts to three call sites; if I make it `int` I have to add five casts.

db/functions.go
snej: Oh, so that's how a null value is encoded? Thanks. But I'll put it at the beginning.

rest/functions_api.go
snej: Yup; see my changes next time you review the PR.

snej:
The windows-latest test failed because of "TestReplicationConcurrentPush", which is nothing to do with me, maybe just a flaky test?
Anyway, ready to re-review.

adamcfraser (dismissed):

adamcfraser (approved):

PR #5805, opened 2022-10-07
CBG-2453: Friendly error message when updating a config with a non-existing collection

gregns1:
CBG-2453

Simple change to return a more user friendly error message if a non existent scope/collection is specified in databse config update. 

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` (for collections) https://jenkins.sgwdev.com/job/SyncGateway-Integration/933/

### DIFF
--- a/base/util_testing.go
+++ b/base/util_testing.go
@@ -760,18 +760,6 @@ func TestRequiresCollections(t *testing.T) {
 	}
 }
 
-func WaitUntilScopeAndCollectionExists(collection *gocb.Collection) error {
-	err, _ := RetryLoop("wait for scope and collection to exist", func() (shouldRetry bool, err error, value interface{}) {
-		_, err = collection.Exists("waitUntilScopeAndCollectionExists", nil)
-		if err != nil {
-			WarnfCtx(context.TODO(), "Error checking if collection exists: %v", err)
-			return true, err, nil
-		}
-		return false, nil, nil
-	}, CreateMaxDoublingSleeperFunc(30, 10, 1000))
-	return err
-}
-
 // CreateBucketScopesAndCollections will create the given scopes and collections within the given BucketSpec.
 func CreateBucketScopesAndCollections(ctx context.Context, bucketSpec BucketSpec, scopes map[string][]string) error {
 	atLeastOneScope := false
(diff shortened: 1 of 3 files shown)


rest/server_context.go
```diff
@@ -447,6 +447,16 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
 		if !bucket.IsSupported(sgbucket.DataStoreFeatureCollections) {
 			return nil, errCollectionsUnsupported
 		}
+		// TODO: This needs changing when implementing Phase 2 collections: support for multiple collections
+		collection, err := base.AsCollection(bucket)
+		if err != nil {
+			return nil, err
+		}
+		// Check if scope/collection specified exists
+		_, err = collection.Exists("waitUntilScopeAndCollectionExists", nil)
```
bbrks: I wonder whether we want to apply the same retry logic to non-existing collections as we do for non-existing buckets, to allow scripted startup/collection creation.

See inside `connectToBucket` for a reference as to what we do regarding retry when opening buckets.

It may be that we make the existing util function a non-test one so we can use it here.

bbrks (changes requested):

base/util_testing.go
```diff
@@ -760,7 +760,7 @@ func TestRequiresCollections(t *testing.T) {
 	}
 }
 
-func waitUntilScopeAndCollectionExists(collection *gocb.Collection) error {
+func WaitUntilScopeAndCollectionExists(collection *gocb.Collection) error {
```
bbrks: Nitpick: This is no longer a testing utility - can it be moved to `base/collection.go`?

Might be nice to have a comment describing why we have to do the Exists check in the first place too.

```suggestion
// WaitUntilScopeAndCollectionExists will try to perform an `Exists` operation on an arbitrary doc ID in the given scope and collection until it can succeed in telling us whether the doc exists or not.
// There's no WaitForReady operation in GoCB for collections, so attempting to use the collection in this way this seems like our best option.
func WaitUntilScopeAndCollectionExists(collection *gocb.Collection) error {
```

bbrks (changes requested):
This should be good to merge after moving the function/adding the godoc comment.
Apologies for the review delays.

bbrks (approved):

PR #5807, opened 2022-10-07
CBG-2282: Added timeout and change retry handling for GoCB v2 ops

IsaacLambat:
CBG-2282

### `initializeCouchbaseServerConnections`
`initializeGoCBAgent` has been left to fail fast due to SG handling the retry loop of it due to CBS and SG race conditioning happening when launching at the same time.

 The cluster connection now uses a best effort retry strategy so any operations that use the GoCB clusster connection will have best effort retry handling instead of failing fast. This is set up in `NewCouchbaseCluster`.

The best effort retry handling is not done when getting the configs (both for the initial launch and for the incremental config fetching) so that this can have an auth error instantly. If this fails fast for any other reason, it is not an issue as it will try again when doing the incremental fetch configs (as defined by `config_update_frequency`) which will load/reload the database depending on if it loaded successfully on startup or not. This is done in `(cc *CouchbaseCluster) connect(auth *gocb.Authenticator) (*gocb.Cluster, error)`.

The `WaitUntilReady` in `(cc *CouchbaseCluster) connectToBucket(bucketName string) (b *gocb.Bucket, teardownFn func(), err error)` did not get changed to use best effort retry handling so it can also report authentication failures instantly. Even though `(cc *CouchbaseCluster) connect(auth *gocb.Authenticator) (*gocb.Cluster, error)` will catch authentication errors, it does not try to authenticate with the bucket directly (only the cluster) so therefore the WaitUntilReady is left as a fail fast so it can be confirmed the user has access to that bucket (as RBAC may prevent this). This is useful for making sure per bucket authentication is successful without getting stuck in a retry loop unnecessarily.

I have added a timeout to the cluster options which will use the GoCB defaults. These defaults are 10 seconds for bucket ops, and 7.5 seconds for a view query op.

Testing was done to make sure authentication failures fail fast using: incorrect per bucket user details, per bucket user details that are valid but can't access the bucket (RBAC), and incorrect cluster details.

### `connectToBucket(ctx context.Context, spec base.BucketSpec) (base.Bucket, error)`
This function gets used when loading/adding a database to Sync Gateway. The cluster options here have been changed to use a best effort retry loop. 

This function implements it's own retry loop around connecting to the bucket so GoCB does not need to attempt to retry so the `WaitUntilReady` in `GetCouchbaseCollection(spec BucketSpec) (*Collection, error)` has been left as a fail fast. This means that authentication errors can be handled that where not handled by `initializeCouchbaseServerConnections` (such as because a database was created that used invalid per db credentials). Using a best effort retry in this `WaitUntilReady` would cause hanging on the PUT db endpoint, and then an unambiguous timeout error which is not helpful.
 
An appropriate timeout is already used for the cluster options which uses user defined options, or 10 seconds by default.

Testing was done using invalid per db credentials and then either creating the database with the same name, or having it already exist when SG is started. This failed fast as expected.

### `GetCollectionFromCluster`
I added a fast fail to the `WaitUntilReady` here so when an invalid bucket name is used, the operation will fast fail and the `PUT /{db}/` endpoint will not just timeout with an unknown error.

Tested using unit test `TestCreateDbOnNonExistentBucket`

## Pre-review checklist

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/945/
- [x] `xattrs=true gsi=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/951

### DIFF
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -102,7 +102,8 @@ func NewCouchbaseCluster(server, username, password,
 	clusterOptions := gocb.ClusterOptions{
 		Authenticator:  clusterAuthConfig,
 		SecurityConfig: securityConfig,
-		RetryStrategy:  &goCBv2FailFastRetryStrategy{},
+		TimeoutsConfig: GoCBv2TimeoutsConfig(nil, nil),
+		RetryStrategy:  gocb.NewBestEffortRetryStrategy(nil),
 	}
 
 	cbCluster := &CouchbaseCluster{
--- a/base/collection.go
+++ b/base/collection.go
@@ -126,7 +126,9 @@ func GetCollectionFromCluster(cluster *gocb.Cluster, spec BucketSpec, waitUntilR
 
 	// Connect to bucket
 	bucket := cluster.Bucket(spec.BucketName)
-	err := bucket.WaitUntilReady(time.Duration(waitUntilReadySeconds)*time.Second, nil)
+	err := bucket.WaitUntilReady(time.Duration(waitUntilReadySeconds)*time.Second, &gocb.WaitUntilReadyOptions{
+		RetryStrategy: &goCBv2FailFastRetryStrategy{},
+	})
 	if err != nil {
 		_ = cluster.Close(&gocb.ClusterCloseOptions{})
 		if errors.Is(err, gocb.ErrAuthenticationFailure) {
--- a/base/collection.go
+++ b/base/collection.go
@@ -67,7 +67,7 @@ func GetCouchbaseCollection(spec BucketSpec) (*Collection, error) {
 		Authenticator:  authenticator,
 		SecurityConfig: securityConfig,
 		TimeoutsConfig: timeoutsConfig,
-		RetryStrategy:  &goCBv2FailFastRetryStrategy{},
+		RetryStrategy:  gocb.NewBestEffortRetryStrategy(nil),
 	}
 
 	if spec.KvPoolSize > 0 {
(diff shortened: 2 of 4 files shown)


bbrks (dismissed):
Seems reasonable to me - would like @torcolvin to have a look as the original issue filer.

base/bootstrap.go
```diff
@@ -102,7 +102,8 @@ func NewCouchbaseCluster(server, username, password,
 	clusterOptions := gocb.ClusterOptions{
 		Authenticator:  clusterAuthConfig,
 		SecurityConfig: securityConfig,
-		RetryStrategy:  &goCBv2FailFastRetryStrategy{},
+		TimeoutsConfig: GoCBv2TimeoutsConfig(nil, nil),
+		RetryStrategy:  gocb.NewBestEffortRetryStrategy(nil),
```
torcolvin, line 106: The only potential concern I have about this, is that this is now applying a timeout and potential retry to query ops (7.5s), if gocb thinks the operation is retryable. The relevant parameters are gocb.TimeoutsConfig.ViewTimeout and gocb.TimeoutsConfig.QueryTimeout. However, the timeout is so small and my guess is that retry is unlikely, so this will not pose a problem.

torcolvin (approved):

base/bootstrap.go
adamcfraser: I'd like to understand the interaction between the gocb query timeout (deadline) and the retry timeout, particularly in the case where the deadline is longer than the retry timeout.

IsaacLambat: Talked to Charles about this. The timeout is for the entire operation (which can have multiple retries per operation) so the request can be retried until the timeout is reached.

PR #5810, opened 2022-10-10
CBG-2411: Removed potential double decrement of wait group

IsaacLambat:
CBG-2411

- Removed wait group decrementing on callback error

This has been tested manually by making `callbackFunc` return an error through `callbackErr`. Could not be tested through a unit test.

## Pre-review checklist

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/946/

### DIFF
--- a/base/collection.go
+++ b/base/collection.go
@@ -1041,7 +1041,6 @@ func (c *Collection) GetCollectionID() (uint32, error) {
 	}
 	wg.Wait()
 	if callbackErr != nil {
-		wg.Done()
 		return 0, fmt.Errorf("GetCollectionID for %s.%s, err: %w", scope, collection, callbackErr)
 	}
 	// cache value for future use


base/collection.go
```diff
@@ -1041,7 +1041,6 @@ func (c *Collection) GetCollectionID() (uint32, error) {
 	}
 	wg.Wait()
```
bbrks, line 1042: Given there's only 1 path through this code (via the callback) that's async, do we need to `Add`, `Done` and `Wait` all on the same goroutine? Doesn't make much sense to me.

I.e. Can we rewrite this bit of code so that we only add to the WaitGroup when we pass the `err != nil` check - and then do the `Add`/`Wait` only for the callback's `Done`?

/cc @torcolvin

bbrks (changes requested):

base/collection.go
torcolvin: If we add the waitgroup before `err != nil` check, there's a chance we might end up with negative wait group I think.

The other pattern we could use is the signal pattern, but I'm not sure which is more readable. https://github.com/couchbase/sync_gateway/blob/335e13d55d6b6ec64d74ecee440a66842ce526f5/base/collection.go#L1074

bbrks: as discussed - let's just move forward with the current fix

bbrks (approved):

PR #5815, opened 2022-10-13
CBG-2329 make mutation feed collection aware

torcolvin:
- Added a `channels.ID` type to go with a `channels.Set` type. At this time, `base.Set` and `timed.Set` are still used for user/role channel markings. I think some of TimedSet will have to take a channel ID.
- Renamed `channels.SetOf` which required taking a testing object to match behavior of `base.SetOf`, but returns an error condition.
- Assume that there's one collectionID -> Database, this will have to be changed. Normally we can get the collections from the user, but really this is used for a few things - what happens when user or role documents change? In a user specification, we have a list of channels, but we don't know what collection they are in. I expect this code to be removed in a future review. This also requires a bunch of functions to return an error condition.

There's one step left to do which is modify how the `*` channel is expanded, and write a test for this because this isn't covered by existing unit tests
 
## Pre-review checklist

- [x] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/995/
- [x] `GSI=true,xattrs=true` (for collections) https://jenkins.sgwdev.com/job/SyncGateway-Integration/994/ (known test flake, reran failing tests again https://jenkins.sgwdev.com/job/SyncGateway-Integration/999/  with pass

### DIFF
--- a/auth/user.go
+++ b/auth/user.go
@@ -20,6 +20,7 @@ import (
 	"golang.org/x/crypto/bcrypt"
 
 	"github.com/couchbase/sync_gateway/base"
+	"github.com/couchbase/sync_gateway/channels"
 	ch "github.com/couchbase/sync_gateway/channels"
 )
 
@@ -588,15 +589,15 @@ func (user *userImpl) ExpandWildCardChannel(channels base.Set) base.Set {
 	return channels
 }
 
-func (user *userImpl) FilterToAvailableChannels(channels base.Set) (filtered ch.TimedSet, removed []string) {
+func (user *userImpl) FilterToAvailableChannels(channels channels.Set) (filtered ch.TimedSet, removed []string) {
 	filtered = ch.TimedSet{}
 	for channel := range channels {
-		if channel == ch.AllChannelWildcard {
+		if channel.Name == ch.AllChannelWildcard {
 			return user.InheritedChannels().Copy(), nil
 		}
-		added := filtered.AddChannel(channel, user.CanSeeChannelSince(channel))
+		added := filtered.AddChannel(channel.Name, user.CanSeeChannelSince(channel.Name))
 		if !added {
-			removed = append(removed, channel)
+			removed = append(removed, channel.Name)
 		}
 	}
 	return filtered, removed
--- a/auth/user.go
+++ b/auth/user.go
@@ -20,7 +20,6 @@ import (
 	"golang.org/x/crypto/bcrypt"
 
 	"github.com/couchbase/sync_gateway/base"
-	"github.com/couchbase/sync_gateway/channels"
 	ch "github.com/couchbase/sync_gateway/channels"
 )
 
@@ -589,7 +588,7 @@ func (user *userImpl) ExpandWildCardChannel(channels base.Set) base.Set {
 	return channels
 }
 
-func (user *userImpl) FilterToAvailableChannels(channels channels.Set) (filtered ch.TimedSet, removed []string) {
+func (user *userImpl) FilterToAvailableChannels(channels ch.Set) (filtered ch.TimedSet, removed []string) {
 	filtered = ch.TimedSet{}
 	for channel := range channels {
 		if channel.Name == ch.AllChannelWildcard {
--- a/channels/active_channels.go
+++ b/channels/active_channels.go
@@ -17,6 +17,7 @@ import (
 	"github.com/couchbase/sync_gateway/base"
 )
 
+// ChanngedChannels contains each modified channel with true if active, and false if inactive
 type ChangedChannels map[ID]bool
 
 // activeChannels is a concurrency-safe map of active replications per channel, modified via
@@ -101,7 +102,7 @@ func (ac *ActiveChannels) _incr(channel ID) {
 func (ac *ActiveChannels) _decr(channel ID) {
 	current, ok := ac.channelCounts[channel]
 	if !ok {
-		base.WarnfCtx(context.Background(), "Attempt made to decrement inactive channel %s - will be ignored", base.UD(channel.String()))
+		base.WarnfCtx(context.Background(), "Attempt made to decrement inactive channel %s - will be ignored", base.UD(channel))
 		return
 	}
 	if current <= 1 {
--- a/channels/active_channels.go
+++ b/channels/active_channels.go
@@ -56,19 +56,23 @@ func (ac *ActiveChannels) UpdateChanged(changedChannels ChangedChannels) {
 }
 
 // Active channel counts track channels being replicated by an active changes request.
-func (ac *ActiveChannels) IncrChannels(chans Set) {
+func (ac *ActiveChannels) IncrChannels(timedSetByCollection TimedSetByCollectionID) {
 	ac.lock.Lock()
 	defer ac.lock.Unlock()
-	for channelName, _ := range chans {
-		ac._incr(channelName)
+	for collectionID, timedSetByChannel := range timedSetByCollection {
+		for channelName, _ := range timedSetByChannel {
+			ac._incr(NewID(channelName, collectionID))
+		}
 	}
 }
 
-func (ac *ActiveChannels) DecrChannels(chans Set) {
+func (ac *ActiveChannels) DecrChannels(timedSetByCollection TimedSetByCollectionID) {
 	ac.lock.Lock()
 	defer ac.lock.Unlock()
-	for channel, _ := range chans {
-		ac._decr(channel)
+	for collectionID, timedSetByChannel := range timedSetByCollection {
+		for channelName, _ := range timedSetByChannel {
+			ac._decr(NewID(channelName, collectionID))
+		}
 	}
 }
 
--- a/channels/active_channels.go
+++ b/channels/active_channels.go
@@ -17,6 +17,8 @@ import (
 	"github.com/couchbase/sync_gateway/base"
 )
 
+type ChangedChannels map[ID]bool
+
 // activeChannels is a concurrency-safe map of active replications per channel, modified via
 // incr/decr operations.
 // Incrementing a channel not already in the map adds it to the map.
@@ -25,27 +27,27 @@ import (
 // Note: private properties shouldn't be accessed directly, to support potential
 // refactoring of activeChannels to use a sharded map as needed.
 type ActiveChannels struct {
-	channelCounts map[string]uint64 // Count of active pull replications (changes) per channel
-	lock          sync.RWMutex      // Mutex for channelCounts map access
-	countStat     *base.SgwIntStat  // Channel count stat
+	channelCounts map[ID]uint64    // Count of active pull replications (changes) per channel
+	lock          sync.RWMutex     // Mutex for channelCounts map access
+	countStat     *base.SgwIntStat // Channel count stat
 }
 
 func NewActiveChannels(activeChannelCountStat *base.SgwIntStat) *ActiveChannels {
 	return &ActiveChannels{
-		channelCounts: make(map[string]uint64),
+		channelCounts: make(map[ID]uint64),
 		countStat:     activeChannelCountStat,
 	}
 }
 
 // Update changed increments/decrements active channel counts based on a set of changed channels.  Triggered
 // when the set of channels being replicated by a given replication changes.
-func (ac *ActiveChannels) UpdateChanged(changedChannels ChangedKeys) {
+func (ac *ActiveChannels) UpdateChanged(changedChannels ChangedChannels) {
 	ac.lock.Lock()
-	for channelName, isIncrement := range changedChannels {
+	for channel, isIncrement := range changedChannels {
 		if isIncrement {
-			ac._incr(channelName)
+			ac._incr(channel)
 		} else {
-			ac._decr(channelName)
+			ac._decr(channel)
 		}
 	}
 
@@ -53,59 +55,59 @@ func (ac *ActiveChannels) UpdateChanged(changedChannels ChangedKeys) {
 }
 
 // Active channel counts track channels being replicated by an active changes request.
-func (ac *ActiveChannels) IncrChannels(chans TimedSet) {
+func (ac *ActiveChannels) IncrChannels(chans Set) {
 	ac.lock.Lock()
+	defer ac.lock.Unlock()
 	for channelName, _ := range chans {
 		ac._incr(channelName)
 	}
-	ac.lock.Unlock()
 }
 
-func (ac *ActiveChannels) DecrChannels(chans TimedSet) {
+func (ac *ActiveChannels) DecrChannels(chans Set) {
 	ac.lock.Lock()
-	for channelName, _ := range chans {
-		ac._decr(channelName)
+	defer ac.lock.Unlock()
+	for channel, _ := range chans {
+		ac._decr(channel)
 	}
-	ac.lock.Unlock()
 }
 
-func (ac *ActiveChannels) IsActive(channelName string) bool {
+func (ac *ActiveChannels) IsActive(channel ID) bool {
 	ac.lock.RLock()
-	_, ok := ac.channelCounts[channelName]
+	_, ok := ac.channelCounts[channel]
 	ac.lock.RUnlock()
 	return ok
 }
 
-func (ac *ActiveChannels) IncrChannel(channelName string) {
+func (ac *ActiveChannels) IncrChannel(channel ID) {
 	ac.lock.Lock()
-	ac._incr(channelName)
+	ac._incr(channel)
 	ac.lock.Unlock()
 }
 
-func (ac *ActiveChannels) DecrChannel(channelName string) {
+func (ac *ActiveChannels) DecrChannel(channel ID) {
 	ac.lock.Lock()
-	ac._decr(channelName)
+	ac._decr(channel)
 	ac.lock.Unlock()
 }
 
-func (ac *ActiveChannels) _incr(channelName string) {
-	current, ok := ac.channelCounts[channelName]
+func (ac *ActiveChannels) _incr(channel ID) {
+	current, ok := ac.channelCounts[channel]
 	if !ok {
 		ac.countStat.Add(1)
 	}
-	ac.channelCounts[channelName] = current + 1
+	ac.channelCounts[channel] = current + 1
 }
 
-func (ac *ActiveChannels) _decr(channelName string) {
-	current, ok := ac.channelCounts[channelName]
+func (ac *ActiveChannels) _decr(channel ID) {
+	current, ok := ac.channelCounts[channel]
 	if !ok {
-		base.WarnfCtx(context.Background(), "Attempt made to decrement inactive channel %s - will be ignored", base.UD(channelName))
+		base.WarnfCtx(context.Background(), "Attempt made to decrement inactive channel %s - will be ignored", base.UD(channel.String()))
 		return
 	}
 	if current <= 1 {
-		delete(ac.channelCounts, channelName)
+		delete(ac.channelCounts, channel)
 		ac.countStat.Add(-1)
 	} else {
-		ac.channelCounts[channelName] = current - 1
+		ac.channelCounts[channel] = current - 1
 	}
 }
--- a/channels/active_channels_test.go
+++ b/channels/active_channels_test.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/couchbase/sync_gateway/base"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestActiveChannelsConcurrency(t *testing.T) {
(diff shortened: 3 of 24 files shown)


channels/set.go
```diff
@@ -27,6 +27,15 @@ const UserStarChannel = "*"     // user channel for "can access all docs"
 const DocumentStarChannel = "!" // doc channel for "visible to all users"
 const AllChannelWildcard = "*"  // wildcard for 'all channels'
 
+type ID struct {
```
adamcfraser, line 34: Although it's in the channels package, I still wonder if ChannelID would be more self-documenting and clear.

channels/set.go
```diff
@@ -27,6 +27,15 @@ const UserStarChannel = "*"     // user channel for "can access all docs"
...
+	Name         string // name of channel
+	CollectionID uint32 // collection it belongs to
+}
+
+type Set map[ID]present
```
adamcfraser, line 53: How efficient is go at doing map looking based on struct keys?  I don't think I've tested, might be interesting to see a simple benchmark of string vs ID lookups for different sized maps.

db/changes.go
```diff
@@ -533,12 +533,17 @@ func (db *Database) MultiChangesFeed(ctx context.Context, chans base.Set, option
...
+	// FIXME, put collection IDs in a set and expandwildcardchannel on each collectionID
+	//if db.user != nil {
+	//	waitChans = db.user.ExpandWildCardChannel(chans)
+	//}
+	collectionID, err := db.getCollectionID()
+	if err != nil {
+		return nil, err
 	}
-	return db.mutationListener.NewWaiterWithChannels(waitChans, db.user)
+	return db.mutationListener.NewWaiterWithChannels(waitChans, db.user, collectionID), nil
```
adamcfraser: ChangeWaiter is really just doing string matching, and I think it probably makes sense to have it continue to do that?  Adding a stringify to channels.ID might be the simplest way to do this - I don't think there are any collisions if you just do basic concatenation.

db/changes.go
```diff
@@ -587,7 +592,11 @@ func (db *Database) checkForUserUpdates(ctx context.Context, userChangeCount uin
 
 			changedRoles := db.user.RoleNames().CompareKeys(previousRoles)
 			if len(changedRoles) > 0 {
-				changeWaiter.RefreshUserKeys(db.User())
+				collectionID, err := db.getCollectionID()
+				if err != nil {
+					return false, 0, nil, err
+				}
+				changeWaiter.RefreshUserKeys(db.User(), collectionID)
```
adamcfraser: This reinforces the idea that changeWaiter should just be working with strings - there isn't any need for a collection when talking about user/role keys in the changeWaiter - these are still just db-scoped (not collection-scoped) users and roles that need to be notified when their user/role documents change.

db/changes.go
adamcfraser: You could still change NewWaiterWithChannels to take channels.Set as a parameter (and do the string generation in that method), but things like Notify could be left as-is.  Adding a new NotifyChannels might be reasonable if you want to pass in a channels.Set to avoid converting a channels.Set to a base.Set everywhere.

adamcfraser (changes requested):
channels.ID makes sense, but I don't think we need to modify changeWaiter extensively.  Details in comments.

channels/set.go
torcolvin: It looks like about 70%, this is the code I used to test https://gist.github.com/torcolvin/00b958ae15f74f133fb612bb350c4b6a

auth/user.go
```diff
@@ -20,6 +20,7 @@ import (
 	"golang.org/x/crypto/bcrypt"
 
 	"github.com/couchbase/sync_gateway/base"
+	"github.com/couchbase/sync_gateway/channels"
```
adamcfraser: Would be nice to avoid the duplicated import here if possible, could stick to using ch alias throughout to avoid refactoring locally scoped `channels` variables.

channels/active_channels.go
```diff
@@ -17,6 +17,8 @@ import (
 	"github.com/couchbase/sync_gateway/base"
 )
 
+type ChangedChannels map[ID]bool
```
adamcfraser, line 21: The previous version of this (ChangedKeys) is defined with the associated struct function (TimedSet.CompareKeys) and includes a comment describing format/usage.  Does it make sense to do the same for this?

Looking further this seems like a bit of a placeholder until TimedSet is updated to be collection-aware, is that right?   If so, maybe stick this in TimedSet with a TODO?

db/changes.go
```diff
@@ -649,25 +666,26 @@ func (db *Database) SimpleMultiChangesFeed(ctx context.Context, chans base.Set,
...
+			channelsSince[singleCollectionID] = chansSince
 		} else {
-			channelsSince = channels.AtSequence(chans, 0)
+			channelsSince = channels.AtSequenceByCollection(chans, 0)
 		}
 
 		// Mark channel set as active, schedule defer
-		db.activeChannels.IncrChannels(channelsSince)
-		defer db.activeChannels.DecrChannels(channelsSince)
+		activeChannels := channelsSince.GetChannels()
```
adamcfraser: We generally want to avoid this kind of duplicate iteration and allocation - once to convert CollectionByTimedSet to channels.Set, and then a second to iterate over the channels.Set in IncrChannels.  Since we're already modifying IncrChannels in this PR, I think it makes more sense to change it to take CollectionByTimedSet as input.

channels/timed_set.go
```diff
@@ -390,3 +390,31 @@ func TimedSetFromString(encoded string) TimedSet {
 	}
 	return set
 }
+
+type CollectionByTimedSet map[uint32]TimedSet
```
adamcfraser: I don't think the 'By' makes sense here - that implies to me "a set of collections, indexed by TimedSet", where we've got the opposite here.  Maybe TimedSetByCollectionID, or possibly CollectionTimedSetMap?  If we're only ever going to index a TimedSet map by collection then a simple 'TimedSetMap' might be sufficient.  In either case I'd also want to alias a `type CollectionID uint32`, or at least include a comment to document what the uint32 map key represents.

channels/active_channels_test.go
```diff
@@ -16,47 +16,58 @@ import (
...
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestActiveChannelsConcurrency(t *testing.T) {
 	activeChannelStat := &base.SgwIntStat{}
 	ac := NewActiveChannels(activeChannelStat)
 	var wg sync.WaitGroup
 
+	ABCChan := ID{Name: "ABC"}
```
adamcfraser: Does collectionID=0 represent the default collection?  Even if it does, it seems like it would be a good idea to enforce that ID has a collectionID populated, to avoid potential bugs in future.  To simplify tests, could potentially add a helper constructor like DefaultID(channelName) that populated collection ID with the default collection ID.

channels/set.go
adamcfraser: Thanks for putting that together.  I tested a few other permutations using that code - using map keys of concatentated strings, or strconv(int) + string, and the map lookup for a struct was faster than performing that concatenation (per lookup) and then doing a string lookup:
struct map speed keyCount: 1000000, Time: 139ms
string map speed keyCount: 1000000, Time: 76ms
concat map speed keyCount: 1000000, Time: 140ms
Based on that I agree that the struct map is the best approach for the use case for now.  (In cases where we can do the concatenation outside of a high contention mutex, and then the map lookup while holding the mutex, it still might be preferable to use a concatenated string key.  But I'm fine leaving that as a performance optimization in future).

channels/set.go
```diff
@@ -66,3 +93,73 @@ func ExpandingStar(set base.Set) base.Set {
...
+	return s
+}
+
+// Contains returns true if the set includes the channel.
+func (s Set) Contains(ch ID) bool {
+	_, exists := s[ch]
+	return exists
+}
+
+// ToSerializedStrings returns a base set with the serialized form of channel IDs.
```
adamcfraser: Is there a use case for ToSerializedStrings()?  It seems like the kind of thing we would generally want to avoid (i.e. code should either use a channels.Set or a base.Set, and not a base.Set of stringified channel IDs.

channels/timed_set.go
```diff
@@ -390,3 +390,31 @@ func TimedSetFromString(encoded string) TimedSet {
...
+
+// AtSequenceByCollection creates a map of collectionID to a map of channel IDs to sequences.
```
adamcfraser, line 397: These both look like relatively expensive conversions - are these intended to be test-only, or are there runtime usages?

db/change_cache.go
```diff
@@ -298,13 +298,21 @@ func (c *changeCache) InsertPendingEntries(ctx context.Context) error {
...
 	if c.notifyChange != nil && len(changedChannels) > 0 {
-		c.notifyChange(changedChannels)
+		c.notifyChannels(changedChannels)
 	}
 	c.lock.Unlock()
 
 	return nil
 }
 
+func (c *changeCache) notifyChannels(chans channels.Set) {
```
adamcfraser: This looks like another case of redundant iterations - where notifyChange is already doing an iteration, and it would be preferable to avoid duplicate iteration (once to convert to base.Set, and once to do the actual Notify.  

You could do this in one of two ways:
- add a NotifyChannels(ids channels.Set) on changeListener that mimics Notify(base.Set), and does the ID.String() inside its existing iteration over the keys.  
- modify _addPendingLogs to return the already stringified keys

I think option 1 is probably preferable, but will let you evaluate.

db/change_listener.go
```diff
@@ -293,14 +293,14 @@ func (listener *changeListener) NewWaiter(keys []string) *ChangeWaiter {
...
+func (listener *changeListener) NewWaiterWithChannels(chans channels.Set, user auth.User) *ChangeWaiter {
 	waitKeys := make([]string, 0, 5)
 	for channel := range chans {
-		waitKeys = append(waitKeys, channel)
+		waitKeys = append(waitKeys, channel.String())
 	}
 	var userKeys []string
 	if user != nil {
-		userKeys = []string{base.UserPrefix + user.Name()}
+		userKeys = append(userKeys, base.UserPrefix+user.Name())
```
adamcfraser: Is there an advantage to using append here as opposed to the existing approach?

rest/bulk_api.go
```diff
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/couchbase/sync_gateway/base"
+	"github.com/couchbase/sync_gateway/channels"
```
adamcfraser: As above, preferable to avoid duplicate imports.

db/database.go
```diff
@@ -1901,3 +1901,15 @@ func (dbCtx *DatabaseContext) AddDatabaseLogContext(ctx context.Context) context
 	}
 	return ctx
 }
+
+// GetSingleCollectionID returns a collectionID. This is a shim for single collections.
+func (dbCtx *DatabaseContext) GetSingleCollectionID() (uint32, error) {
```
adamcfraser, line 1906: Is this temporary and will be removed?  Might be good to indicate in a comment if so.

db/channel_cache.go
```diff
@@ -371,9 +372,9 @@ func (c *channelCacheImpl) addChannelCache(channelName string) (*singleChannelCa
 	return singleChannelCache, true
 }
 
-func (c *channelCacheImpl) getActiveChannelCache(channelName string) (*singleChannelCacheImpl, bool) {
+func (c *channelCacheImpl) getActiveChannelCache(channel channels.ID) (*singleChannelCacheImpl, bool) {
 
-	cacheValue, found := c.channelCaches.Get(channelName)
+	cacheValue, found := c.channelCaches.Get(channel.String())
```
adamcfraser: Is there a reason to not index c.channelCaches by ID instead of string?  I suppose that could be done as a follow-on enhancement if the changes to RangeSafeCollection are significant.  Is the performance overhead of doing channel.String() repeatedly acceptable because they are happening outside of the RangeSafeCollection mutex?

At minimum I wonder whether an ID can be considered immutable, so that the string returned by channel.String() could be computed once instead of per-call.

db/channel_cache_single.go
```diff
@@ -101,7 +101,7 @@ type SingleChannelCache interface {
 }
 
 type singleChannelCacheImpl struct {
-	channelName      string               // The channel name, duh
+	channel          channels.ID          // The channel
```
adamcfraser: I think channelID is better here, or just ID.

db/channel_cache_single.go
```diff
@@ -413,7 +413,8 @@ func (c *singleChannelCacheImpl) GetChanges(options ChangesOptions) ([]*LogEntry
 	// overlap, which helps confirm that we've got everything.
 	c.cacheStats.ChannelCacheMisses.Add(1)
 	endSeq := cacheValidFrom
-	resultFromQuery, err := c.queryHandler.getChangesInChannelFromQuery(options.LoggingCtx, c.channelName, startSeq, endSeq, options.Limit, options.ActiveOnly)
+	// TOR: make sure queryhandler knows what collection it needs
```
adamcfraser: At the moment, the queryHandler is associated with the database's single collection?

db/changes.go
```diff
@@ -682,9 +700,9 @@ func (db *Database) SimpleMultiChangesFeed(ctx context.Context, chans base.Set,
 		for {
 			// Updates the ChangeWaiter to the current set of available channels
 			if changeWaiter != nil {
-				changeWaiter.UpdateChannels(channelsSince)
+				changeWaiter.UpdateChannels(activeChannels)
 			}
-			base.DebugfCtx(ctx, base.KeyChanges, "MultiChangesFeed: channels expand to %#v ... %s", base.UD(channelsSince.String()), base.UD(to))
+			base.DebugfCtx(ctx, base.KeyChanges, "MultiChangesFeed: channels expand to %#v ... %s", base.UD(activeChannels.ToSerializedStrings()), base.UD(to))
```
adamcfraser: If I recall correctly, the work to build activeChannels.ToSerializedStrings() is going to occur even if we don't have debug logging enabled.  Would be preferable to avoid this, but it would mean an enhancement to UD to support channel.Set in a similar way that we're supporting base.Set with RedactorSet.  Some work would be needed to avoid a circular dependency there (probably by implementing a redaction interface in base that both base.Set and channels.Set can interface, then type switch on that interface).  
I'm fine leaving that as a follow-on enhancement, but can you file a ticket for that to make sure it's done?

db/changes.go
```diff
@@ -713,116 +731,121 @@ func (db *Database) SimpleMultiChangesFeed(ctx context.Context, chans base.Set,
...
 
-				// Check whether requires backfill based on changedChannels in this _changes feed
-				isNewChannel := false
-				if changedChannels != nil {
-					isNewChannel, _ = changedChannels[name]
-				}
+					// Check whether requires backfill based on changedChannels in this _changes feed
+					isNewChannel := false
+					if changedChannels != nil {
+						// FIXME: doesn't seem right
```
adamcfraser: Does this still need review?

db/changes.go
```diff
@@ -713,116 +731,121 @@ func (db *Database) SimpleMultiChangesFeed(ctx context.Context, chans base.Set,
...
+							if revokedSeq > options.Since.TriggeredBy {
+								revocationSinceSeq = options.Since.TriggeredBy
+								revokeFrom = 0
+							}
 						}
-					}
 
-					feed := db.buildRevokedFeed(ctx, channel, options, revokedSeq, revocationSinceSeq, revokeFrom, to)
-					feeds = append(feeds, feed)
+						// FIXME: multiple collections
```
adamcfraser: This doesn't seem like a FIXME any more.

adamcfraser (changes requested):
Overall the approach and implementation look good - comments on a few of the implementation details that I think are worth looking at now instead of deferring for the next phase of work.

channels/active_channels_test.go
torcolvin: collectionID=0 is `_default._default`, which is an interesting distinction to make, because it's not `scopeFoo._default`.

db/channel_cache_single.go
torcolvin: I removed this comment, it will get fixed in [CBG-2327](https://issues.couchbase.com/browse/CBG-2327), and the tests very obviously fail.

db/changes.go
torcolvin: I implemented this with the `GoString` method which will do lazy evaluation of it and removes `ToSerializedStrings`

channels/set.go
torcolvin: It was used for logging but I replaced it with `GoString()`.

db/change_listener.go
torcolvin: No, this was just part of code that I refactored and undid, returning to original way.

db/channel_cache.go
torcolvin: This is actually performance sensitive and I decided to make `RangeSafeCollection` not generic in order to support this. I think that I would just have to change the iteration before to make this work on `interface{}` and this might actually be a use for go generics.

channels/set.go
torcolvin: The sole use case is for https://github.com/couchbase/sync_gateway/blob/master/db/changes.go#L687

channels/timed_set.go
torcolvin: I changed the name - I wasn't thrilled by the type alias of `CollectionID` which I like in the abstract but it not used by gocb/gocbcore.

channels/timed_set.go
torcolvin: `AtSequenceByCollection` is the constructor creating sequence IDs for each channel.

channels/active_channels.go
torcolvin: I think this might be a placeholder until user/roles are collection aware, because then the `TimedSet` returned from auth package are aware.

db/change_cache.go
torcolvin: I tried option 1, but left it as a separate commit for you to look at.

db/changes.go
torcolvin: I did this, but I feel like we are still iterating, we just move where we do it. I am not sure I can see the 100% picture without the principal changes. I left this as a separate commit in case you are indecisive also.

channels/active_channels.go
adamcfraser: What are you thinking this changes to at that point?  Are you thinking that once principals are channel aware, we go back to using ChangedKeys, and have a separate ChangedKeys per collection?

db/changes.go
adamcfraser: Aren't we just iterating over collections and channels once now?  Previously we were iterating over collections and channels to build a slice of ID, then iterating again over that slice of IDs to Incr/Decr channels.

channels/timed_set.go
adamcfraser: I only see test usages today, so I don't think it's a generic constructor?  I'd like to avoid accidentally using this in production (in general I think it would make more sense to build the TimedSetByCollectionID in the first place, instead of building a channels.Set then converting to TimedSetByCollectionID), unless you think there's an actual use case I'm overlooking.

channels/active_channels_test.go
adamcfraser: I don't think named scopes have a _default collection, do they?

adamcfraser (commented):
The changes look ok, want to make sure I understand some of the resolved comments.

db/change_listener.go
```diff
@@ -181,22 +181,33 @@ func (listener changeListener) TapFeed() base.TapFeed {
...
+		listener.keyCounts[key.String()] = listener.counter
 	}
 	base.DebugfCtx(context.TODO(), base.KeyChanges, "Notifying that %q changed (keys=%q) count=%d",
 		base.MD(listener.bucketName), base.UD(keys), listener.counter)
 	listener.tapNotifier.Broadcast()
 	listener.tapNotifier.L.Unlock()
 }
 
+// Changes the counter, notifying waiting clients. Only use for single update is updating user, which might be collection scoped in the future.
+func (listener *changeListener) notifyUser(key string) {
```
adamcfraser: I like the idea of having a separate Notify for a single key, since that's currently how principals are notified.  I don't think there's anything in here that's specific to user/role, though.  Might be better to just switch to NotifyKey() and maybe NotifyChannels for the version that accepts channels.Set?

adamcfraser (changes requested):
One API tweak requested for the notify changes.

db/changes.go
torcolvin: We do iterate over these to build the TimedSetByCollection, and there is one fewer iteration than before.

channels/active_channels.go
torcolvin: To step back a bit, `ActiveChannels` is used to track replications for a database. It's basically a fancy counter, so we could use serialized formats, but at a potential cost. The question is whether we want to actually know active channels per replication by channel, or we are happy with a database level value.

Ideas:

* have ChangedChannels be a `ChangedKeys` again, because the keys could be a principal that doesn't have a channel. I could have converted everything to string and use the existing datastructures. `ActiveChannels` is really just a fancy counter.
* Otherwise we could use `ChangedKeys` per collection if we want to track channels per collection on a replication.
* We could create a datastructure that is actually `ActiveChannels` but it is collection scoped.

channels/timed_set.go
torcolvin: We use this in `db/changes.go`.

db/change_listener.go
torcolvin: Fixed.

adamcfraser (dismissed):

adamcfraser (approved):

PR #5819, opened 2022-10-15
CBG-2277 Use current revision when performing 3.0 revocation access check

adamcfraser:
CBG-2277

Modifies the handling in blip_handler.go to check per changes entry, instead of per rev per changes entry (since result will now be the same in both cases).

Switches UserHasDocAccess to call user.AuthorizeAnyChannel directly, since it's not using the change entry that authorizeUserForChannels builds.

- [ ] `GSI=true,xattrs=true` (for collections) https://jenkins.sgwdev.com/job/SyncGateway-Integration/970/

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -391,37 +391,35 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 		base.DebugfCtx(bh.loggingCtx, base.KeySync, "    Sending %d changes", len(changes))
 		for _, change := range changes {
 			if !strings.HasPrefix(change.ID, "_") {
-				for _, item := range change.Changes {
-					changeRow := bh.buildChangesRow(change, item["rev"])
-
-					// If change is a removal and we're running with protocol V3 and change change is not a tombstone
-					// fall into 3.0 removal handling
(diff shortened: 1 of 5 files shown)


bbrks:
The test we just filed looks a bit suspect for `UserHasDocAccess` given the code is changing this function. Maybe double check this.

db/blip_handler.go
```diff
@@ -391,37 +391,35 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 		base.DebugfCtx(bh.loggingCtx, base.KeySync, "    Sending %d changes", len(changes))
 		for _, change := range changes {
 			if !strings.HasPrefix(change.ID, "_") {
-				for _, item := range change.Changes {
-					changeRow := bh.buildChangesRow(change, item["rev"])
+				// If change is a removal and we're running with protocol V3 and change change is not a tombstone
```
JRascagneres, line 394: Just a quick mistake in the comment:
change change

JRascagneres (dismissed):

JRascagneres:
> The test we just filed looks a bit suspect for `UserHasDocAccess` given the code is changing this function. Maybe double check this.

+1 on this

adamcfraser:
Updated integration tests with the test fix are clean:
https://jenkins.sgwdev.com/job/SyncGateway-Integration/973/

PR #5821, opened 2022-10-17
CBG-2450: Do not send revisions that have a doc id starting with a null byte

IsaacLambat:
CBG-2450

- Do not include docs in the changes feed that have a doc id prefixed with a null byte (therefore not sent over BLIP)
- Added testing to make sure a document with a id prefixed with a null byte is not received through a replication

Caveat - The doc will not be sent over the changes feed when requested through `GET /db/_changes`

## Pre-review checklist

- [ ] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/985/


rest/replication_api_test.go
```diff
@@ -1764,3 +1764,123 @@ func TestDBReplicationStatsTeardown(t *testing.T) {
 	require.NoError(t, err)
 
 }
+
+// Repros CBG-2450 - Make sure doc that have an ID prefixed with a null byte don't get sent over BLIP
+func TestReplicateDocIDNullCharPrefix(t *testing.T) {
```
bbrks: I'm wondering whether we can change (or add) a changes feed unit-test (in the db package)

It's more targeted, and avoids the overhead of 2 buckets/RestTester/ISGR and could potentially give us easier coverage of the more risky areas of this change like around limits/sequence numbering if we decide to go for the extra coverage.

bbrks (dismissed):
The comment I left is optional, depending on how quickly we need this change in... but I feel like the 2 bucket/RestTester/ISGR style test is a bit heavy-weight and less flexible for us to test this change instead of a targeted changes feed unit test.

adamcfraser (approved):

PR #5823, opened 2022-10-18
feat: added new sync indexes

nirav24:
adding syncUser, syncRole, syncSession indexes to be used instead of syncDoc index.

CBG-0000

Describe your PR here...

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` (for collections) https://jenkins.sgwdev.com/job/SyncGateway-Integration/000/

### DIFF
--- a/db/indexes.go
+++ b/db/indexes.go
@@ -83,9 +83,9 @@ var (
 		IndexAllDocs:    "allDocs",
 		IndexTombstones: "tombstones",
 		IndexSyncDocs:   "syncDocs",
-		IndexUser:       "syncUser",
-		IndexSession:    "syncSession",
-		IndexRole:       "syncRole",
+		IndexUser:       "users",
+		IndexSession:    "sessions",
+		IndexRole:       "roles",
 	}
 
 	// Index versions - must be incremented when index definition changes
--- a/db/indexes.go
+++ b/db/indexes.go
@@ -56,7 +56,6 @@ const (
 	IndexUser
 	IndexSession
 	IndexRole
-	IndexRoleExcludeDeleted
 	indexTypeCount // Used for iteration
 )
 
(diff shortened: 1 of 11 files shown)


db/indexes.go
```diff
@@ -98,14 +117,22 @@ var (
...
-		IndexAllDocs:    "$sync.sequence, $sync.rev, $sync.flags, $sync.deleted",
-		IndexTombstones: "$sync.tombstoned_at",
-		IndexSyncDocs:   "META().id",
+		IndexAllDocs:                "$sync.sequence, $sync.rev, $sync.flags, $sync.deleted",
+		IndexTombstones:             "$sync.tombstoned_at",
+		IndexSyncDocs:               "META().id",
+		IndexSyncUser:               "META().id, name, email, disabled",
+		IndexSyncSession:            "META().id, username",
+		IndexSyncRole:               "META().id",
+		IndexSyncRoleExcludeDeleted: "META().id, deleted",
```
adamcfraser: We should just include the deleted property in IndexSyncRole and just use a single index for all role use cases.

db/indexes.go
```diff
@@ -50,6 +53,10 @@ const (
 	IndexAllDocs
 	IndexTombstones
 	IndexSyncDocs
+	IndexSyncUser
```
adamcfraser: IndexUsers, IndexRoles, IndexSessions are better naming for these (as there's no concept of non-sync users/roles/sessions)

db/query.go
```diff
@@ -500,6 +553,9 @@ func (context *DatabaseContext) QueryPrincipals(ctx context.Context, startKey st
 	}
 
 	queryStatement := replaceIndexTokensQuery(QueryPrincipals.statement, sgIndexes[IndexSyncDocs], context.UseXattrs())
+	if context.Options.Serverless {
```
adamcfraser: Better to use if/else here to avoid paying for token replacement for a statement we're not going to use.

db/indexes.go
```diff
@@ -63,32 +70,44 @@ const (
 var (
 	// Simple index names - input to indexNameFormat
 	indexNames = map[SGIndexType]string{
-		IndexAccess:     "access",
-		IndexRoleAccess: "roleAccess",
-		IndexChannels:   "channels",
-		IndexAllDocs:    "allDocs",
-		IndexTombstones: "tombstones",
-		IndexSyncDocs:   "syncDocs",
+		IndexAccess:                 "access",
```
adamcfraser: We want to create either IndexSyncDocs or the new set of indexes for a given environment - need to discuss the best approach to do this.

adamcfraser (changes requested):

db/indexes.go
nirav24: ~I tried that but with `IndexSyncRole` index, the query performs fetch on user. I might be doing something wrong. I will try to verify it again with `IndexSyncRole`~

I was using disabled instead of deleted 🤦‍♂️ fixed it.

db/indexes.go
nirav24: make sense. updated

db/query.go
nirav24: addressed

db/indexes.go
nirav24: @adamcfraser do you want to update `indexNames` for these to not include `sync` as well? I chose to keep sync in it since we use `\\_sync:role:%` 

the change will be something like below:

```golang
indexNames = map[SGIndexType]string{
// other indexes
	IndexUser:       "users",
	IndexSession:    "sessions",
	IndexRole:       "roles",
}
```

nirav24:
moved to https://github.com/couchbase/sync_gateway/pull/5835

PR #5830, opened 2022-10-20
CBG-2472: Create db should return error when dbname on path doesn't match body

mohammed-madi:
CBG-2472

- When creating a DB, the path name was used by default, even if the given JSON body contained a different name.
- Now if the JSON doesn't contain a name, it defaults to the name in the path. Otherwise it compares the path name and JSON name and if they're not the same it throws an HTTP Error 400, eg. ` 400 When providing a name in the JSON body (dummy), ensure it matches the name in the path (db1). `
- Added one unit test with 3 subtests.
- Exported persistentConfig in RestTesterConfig.

- [x] 'xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/996/

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -48,13 +48,11 @@ func (h *handler) handleCreateDB() error {
 
 	validateOIDC := !h.getBoolQuery(paramDisableOIDCValidation)
 
-	if config.Name == "" {
-		config.Name = dbName
+	if dbName != config.Name && config.Name != "" {
+		return base.HTTPErrorf(http.StatusBadRequest, "Name in path (%s) is not the same name in the JSON body (%s).\nDatabase name in path (%s) must match name in body (%s) when provided", dbName, config.Name, dbName, config.Name)
 	}
 
-	if dbName != config.Name {
-		return base.HTTPErrorf(http.StatusBadRequest, "Name in path (%s) is not the same name in the JSON body (%s)", dbName, config.Name)
-	}
+	config.Name = dbName
 
 	if h.server.persistentConfig {
 		if err := config.validatePersistentDbConfig(); err != nil {
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -48,6 +48,10 @@ func (h *handler) handleCreateDB() error {
 
 	validateOIDC := !h.getBoolQuery(paramDisableOIDCValidation)
 
+	if config.Name == "" {
+		config.Name = dbName
+	}
+
 	if dbName != config.Name {
 		return base.HTTPErrorf(http.StatusBadRequest, "Name in path (%s) is not the same name in the JSON body (%s)", dbName, config.Name)
 	}
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -45,9 +45,12 @@ func (h *handler) handleCreateDB() error {
 	if err != nil {
 		return err
 	}
-	config.Name = dbName
 
 	validateOIDC := !h.getBoolQuery(paramDisableOIDCValidation)
+
+	if dbName != config.Name {
+		return base.HTTPErrorf(http.StatusBadRequest, "Name in path (%s) is not the same name in the JSON body (%s)", dbName, config.Name)
+	}
 
 	if h.server.persistentConfig {
 		if err := config.validatePersistentDbConfig(); err != nil {
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -49,7 +49,7 @@ func (h *handler) handleCreateDB() error {
 	validateOIDC := !h.getBoolQuery(paramDisableOIDCValidation)
 
 	if dbName != config.Name && config.Name != "" {
-		return base.HTTPErrorf(http.StatusBadRequest, "Name in path (%s) is not the same name in the JSON body (%s).\nDatabase name in path (%s) must match name in body (%s) when provided", dbName, config.Name, dbName, config.Name)
+		return base.HTTPErrorf(http.StatusBadRequest, "When providing a name in the JSON body (%s), ensure it matches the name in the path (%s).", config.Name, dbName)
 	}
 
 	config.Name = dbName
--- a/rest/adminapitest/admin_api_test.go
+++ b/rest/adminapitest/admin_api_test.go
@@ -2141,6 +2141,46 @@ func TestHandleCreateDB(t *testing.T) {
 	rest.RequireStatus(t, resp, http.StatusBadRequest)
 }
 
+func TestHandleCreateDBNoJsonName(t *testing.T) {
+	rt := rest.NewRestTester(t, nil)
+	defer rt.Close()
+
+	bucket := "albums"
+	kvTLSPort := 11207
+	resource := fmt.Sprintf("/%s/", bucket)
+
+	bucketConfig := rest.BucketConfig{Bucket: &bucket, KvTLSPort: kvTLSPort}
+	dbConfig := &rest.DbConfig{BucketConfig: bucketConfig, SGReplicateEnabled: base.BoolPtr(false)}
+
+	reqBody, err := base.JSONMarshal(dbConfig)
+	assert.NoError(t, err, "Error unmarshalling changes response")
+
+	resp := rt.SendAdminRequest(http.MethodPut, resource, string(reqBody))
+	rest.RequireStatus(t, resp, http.StatusCreated)
+	resp = rt.SendAdminRequest(http.MethodGet, resource, string(reqBody))
+	assert.Contains(t, resp.Body.String(), "\"db_name\":\"albums\"")
+}
+
+func TestHandleCreateDBDifferentJsonName(t *testing.T) {
+	rt := rest.NewRestTester(t, nil)
+	defer rt.Close()
+
+	bucket := "albums"
+	kvTLSPort := 11207
+	resource := fmt.Sprintf("/%s/", bucket)
+
+	bucketConfig := rest.BucketConfig{Bucket: &bucket, KvTLSPort: kvTLSPort}
+	dbConfig := &rest.DbConfig{BucketConfig: bucketConfig, Name: "dummy", SGReplicateEnabled: base.BoolPtr(false)}
+
+	reqBody, err := base.JSONMarshal(dbConfig)
+	assert.NoError(t, err, "Error unmarshalling changes response")
+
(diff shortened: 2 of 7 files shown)


rest/adminapitest/admin_api_test.go
```diff
@@ -2141,6 +2141,46 @@ func TestHandleCreateDB(t *testing.T) {
 	rest.RequireStatus(t, resp, http.StatusBadRequest)
 }
 
+func TestHandleCreateDBNoJsonName(t *testing.T) {
```
IsaacLambat: These tests could be in the same test and then use sub tests to run multiple scenarios such as using the same path and name in the body, a different path and name, and a path but no name.

See `TestPutIDRevMatchBody` for an example of using sub tests.

rest/adminapitest/admin_api_test.go
```diff
@@ -2141,6 +2141,46 @@ func TestHandleCreateDB(t *testing.T) {
...
+	bucketConfig := rest.BucketConfig{Bucket: &bucket, KvTLSPort: kvTLSPort}
+	dbConfig := &rest.DbConfig{BucketConfig: bucketConfig, SGReplicateEnabled: base.BoolPtr(false)}
+
+	reqBody, err := base.JSONMarshal(dbConfig)
+	assert.NoError(t, err, "Error unmarshalling changes response")
+
+	resp := rt.SendAdminRequest(http.MethodPut, resource, string(reqBody))
+	rest.RequireStatus(t, resp, http.StatusCreated)
+	resp = rt.SendAdminRequest(http.MethodGet, resource, string(reqBody))
+	assert.Contains(t, resp.Body.String(), "\"db_name\":\"albums\"")
```
IsaacLambat: Have a look at `TestServerlessFetchConfigsLimited` to see how to make a database on the RestTester through the rest API with ease. You should not need to provide a kv port, and bucket config.

By default, the RestTester creates a database called `db`, so it would be good to run the rest tester in persistent config mode (an option in the RT config) as this makes the rest tester not automatically create a database. The test will need to be skipped when running with Walrus (as walrus does not support persistent config).

rest/admin_api.go
```diff
@@ -45,10 +45,17 @@ func (h *handler) handleCreateDB() error {
 	if err != nil {
 		return err
 	}
-	config.Name = dbName
 
 	validateOIDC := !h.getBoolQuery(paramDisableOIDCValidation)
 
+	if config.Name == "" {
```
IsaacLambat: This if statement is not needed as the config name can be set after the validation is done after line 55-57

IsaacLambat (changes requested):

rest/admin_api.go
```diff
@@ -45,10 +45,17 @@ func (h *handler) handleCreateDB() error {
...
+		config.Name = dbName
+	}
+
+	if dbName != config.Name {
+		return base.HTTPErrorf(http.StatusBadRequest, "Name in path (%s) is not the same name in the JSON body (%s)", dbName, config.Name)
```
IsaacLambat: Might be good to indicate that the name in the JSON body does not need to be provided such as something along the lines of:
`Database name in path (%s) must match name in body (%s) when provided`

What do you think?

IsaacLambat (changes requested):

rest/admin_api.go
mohammed-madi: All requested changes added in commit https://github.com/couchbase/sync_gateway/pull/5830/commits/5ef3a73bd47ad104a918d9e02ec05469f3de48df

rest/admin_api.go
```diff
@@ -45,10 +45,15 @@ func (h *handler) handleCreateDB() error {
 	if err != nil {
 		return err
 	}
-	config.Name = dbName
 
 	validateOIDC := !h.getBoolQuery(paramDisableOIDCValidation)
 
+	if dbName != config.Name && config.Name != "" {
+		return base.HTTPErrorf(http.StatusBadRequest, "Name in path (%s) is not the same name in the JSON body (%s).\nDatabase name in path (%s) must match name in body (%s) when provided", dbName, config.Name, dbName, config.Name)
```
IsaacLambat: It would be best not to use a `\n` in a HTTP JSON error response as it will probably not be parsed by whatever is reading the JSON.

An example of this is when using Postman: 
image

Also, a bit of a nit pick, but the error could be made more concise rather than 2 sentences

rest/adminapitest/admin_api_test.go
```diff
@@ -2141,6 +2141,73 @@ func TestHandleCreateDB(t *testing.T) {
...
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			tb := base.GetTestBucket(t)
+			defer tb.Close()
+
+			rt := rest.NewRestTester(t, &rest.RestTesterConfig{
+				CustomTestBucket: tb,
+				PersistentConfig: true,
+				MutateStartupConfig: func(config *rest.StartupConfig) {
+					config.Bootstrap.ConfigUpdateFrequency = base.NewConfigDuration(0)
```
IsaacLambat: Config update frequency controls the interval that the bucket/s are checked for any database updates and databases not in Sync Gateway.

This option (along with the `MutateStartupConfig` option on the RT) is not needed for this test.

rest/adminapitest/admin_api_test.go
```diff
@@ -2141,6 +2141,73 @@ func TestHandleCreateDB(t *testing.T) {
...
+				)
+			} else {
+				resp = rt.SendAdminRequest(http.MethodPut, "/db1/",
+					fmt.Sprintf(
+						`{"bucket": "%s", "num_index_replicas": 0, "enable_shared_bucket_access": %t, "use_views": %t}`,
+						tb.GetName(), base.TestUseXattrs(), base.TestsDisableGSI(),
+					),
+				)
+			}
+			if test.name == "Name mismatch" {
```
IsaacLambat: It would be better to instead use the if statement to check `test.expectError` instead of using the test name to decide the assert.

rest/adminapitest/admin_api_test.go
```diff
@@ -2141,6 +2141,73 @@ func TestHandleCreateDB(t *testing.T) {
...
+				CustomTestBucket: tb,
+				PersistentConfig: true,
+				MutateStartupConfig: func(config *rest.StartupConfig) {
+					config.Bootstrap.ConfigUpdateFrequency = base.NewConfigDuration(0)
+				},
+			})
+			defer rt.Close()
+
+			var resp *rest.TestResponse
+			if test.name != "No JSON Name" {
```
IsaacLambat: Using the test name to decide what action to take seems a bit wrong to me as it means changing the test in the future may cause unexpected issues. It would be better here to instead use the if statement to check if `test.JSONname` is not empty and then continue with the logic to insert that name in the database config.

rest/adminapitest/admin_api_test.go
```diff
@@ -2141,6 +2141,73 @@ func TestHandleCreateDB(t *testing.T) {
...
+				resp = rt.SendAdminRequest(http.MethodPut, "/db1/",
+					fmt.Sprintf(
+						`{"bucket": "%s", "name": "%s", "num_index_replicas": 0, "enable_shared_bucket_access": %t, "use_views": %t}`,
```
IsaacLambat: Instead of having nearly the same code here as L2195, a variable could be set here like`DbConfigJson = ``"name":"`` + test.JSONname + ``"``, ` (double backticks should be single here) which can then be inserted in to the database config on L2195. 

This way, if the variable is not blank, it will be inserted in to the database config, otherwise it will not change the database config.

IsaacLambat (changes requested):
Lots of improvements since last time. Just a few easy changes to the test are needed, and a change to the error text.

rest/adminapitest/admin_api_test.go
mohammed-madi: Added in commit https://github.com/couchbase/sync_gateway/pull/5830/commits/a3d2ce57cdc84f9eab786c4923e28cc0a0add5a5 
As well as above requested changes

rest/adminapitest/admin_api_test.go
```diff
@@ -2141,6 +2141,66 @@ func TestHandleCreateDB(t *testing.T) {
...
+					`{"bucket": "%s", %s "num_index_replicas": 0, "enable_shared_bucket_access": %t, "use_views": %t}`,
+					tb.GetName(), DbConfigJson, base.TestUseXattrs(), base.TestsDisableGSI(),
+				),
+			)
+			if test.expectError {
+				rest.RequireStatus(t, resp, http.StatusBadRequest)
+			} else {
+				rest.RequireStatus(t, resp, http.StatusCreated)
+				resp := rt.SendAdminRequest(http.MethodGet, "/db1/", "")
+				assert.Contains(t, resp.Body.String(), "\"db_name\":\"db1\"")
```
IsaacLambat: It might be good to unmarshal the JSON and check it instead of a string comparison just in-case the JSON format of that page is changed in the future but it's still valid (such as a space added between the colon and value).
```suggestion

				resp = rt.SendAdminRequest(http.MethodGet, "/db1/", "")
				rest.RequireStatus(t, resp, http.StatusOK)
				
				var dbStatus map[string]any
				err := json.Unmarshal(resp.BodyBytes(), &dbStatus)
				require.NoError(t, err)
				assert.EqualValues(t, dbStatus["db_name"], "db1")
```

IsaacLambat (changes requested):

rest/adminapitest/admin_api_test.go
mohammed-madi: Change added in https://github.com/couchbase/sync_gateway/pull/5830/commits/521c33fc8ab5595a0fa3459bfe9b858a5b703f94

mohammed-madi:
More recent jenkins run at https://jenkins.sgwdev.com/job/SyncGateway-Integration/996/

IsaacLambat (approved):
LGTM. PR description will need updating with the new change details and the latest Jenkins run 

Will assign to Adam for a final check

mohammed-madi:
TestReplicatorCheckpointOnStop failed in prior Jenkins run
https://jenkins.sgwdev.com/blue/organizations/jenkins/Sync%20Gateway%20Pipeline/detail/PR-5830/7/tests/

bbrks (approved):
LGTM

PR #5832, opened 2022-10-21
CBG-2266 Split Collection and Gocbv2Bucket and interfaces and introduce MetadataCollection

bbrks:
CBG-2266

## Integration tests
- [x] `gsi=true xattrs=true default_collection=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1186/
- [x] `gsi=true xattrs=true default_collection=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1185/

PR #5833, opened 2022-10-22
Support ad-hoc unit testing against existing Couchbase Server bucket

adamcfraser:
When running unit tests at dev time, there are scenarios where it's more efficient to test against an already provisioned bucket, instead of depending on the TestBucketPool to provision/flush the bucket.  To simplify this type of testing, this adds SG_TEST_USE_EXISTING_BUCKET environment variable which will bypass the bucket readier code, and instead attempt to establish a connection to the named bucket.

### DIFF
--- a/base/main_test_bucket_pool.go
+++ b/base/main_test_bucket_pool.go
@@ -113,6 +113,7 @@ type TestBucketPool struct {
 	unclosedBucketsLock sync.Mutex
 
 	useNamedCollections bool
+	useExistingBucket   bool
 }
 
 // shouldUseNamedCollections returns true if cluster will be created with named collections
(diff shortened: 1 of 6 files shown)


torcolvin:
I like this idea, but I might wait for https://github.com/couchbase/sync_gateway/pull/5832 to land first.

bbrks:
@torcolvin I'd rather get this in and deal with the conflict on #5832 as I've not yet fully completed the test bucket pool changes there

base/main_test_bucket_pool.go
```diff
@@ -318,6 +322,44 @@ func (tbp *TestBucketPool) GetWalrusTestBucket(t testing.TB, url string) (b Buck
 	}
 }
 
+// GetExistingBucket opens a bucket conection to an existing bucket
+func (tbp *TestBucketPool) GetExistingBucket(t testing.TB) (b Bucket, s BucketSpec, teardown func()) {
+	testCtx := TestCtx(t)
+
+	bucketCluster := initV2Cluster(UnitTestUrl())
+
+	var useNamedCollections bool
```
torcolvin: Is this `shouldUseNamedCollections` ?

base/main_test_bucket_pool.go
```diff
@@ -318,6 +322,44 @@ func (tbp *TestBucketPool) GetWalrusTestBucket(t testing.TB, url string) (b Buck
...
+		if !TestsDisableGSI() {
+			useNamedCollections = true
+		}
+	}
+
+	bucketName := tbpBucketName(TestUseExistingBucketName())
+	bucketSpec := getBucketSpec(bucketName, useNamedCollections)
+
+	bucket := bucketCluster.Bucket(bucketSpec.BucketName)
+	err := bucket.WaitUntilReady(10*time.Second, nil)
```
torcolvin: Can we turn this into a test constant https://cs.github.com/couchbase/sync_gateway/blob/d1ce9ca698e6e3a4474e74886fb41a5bd4648e6a/base/main_test_cluster.go?q=waituntilready#L149 ? I notice different times on server 6.6 and I'd rather modify in one spot.

torcolvin (commented):
LGTM, just some comments about code duplication.

base/main_test_bucket_pool.go
bbrks: I think I'll just set everything to use the `waitForReadyBucketTimeout = time.Minute` const - will take slightly longer in the failure case but happy-path shouldn't be affected by the increased timeout.

torcolvin (approved):

PR #5835, opened 2022-10-24
CBG-2523: added new sync indexes

nirav24:
CBG-2523

Describe your PR here...
- Add new indexes when running in serverless mode

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1030/
- [x] `GSI=true,xattrs=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1034/


db/indexes.go
```diff
@@ -166,14 +209,15 @@ func init() {
...
-	flags            SGIndexFlags // Additional index options
+	simpleName       string            // Simplified index name (used to build fullIndexName)
+	expression       string            // Expression used to create index
+	filterExpression string            // (Optional) Filter expression used to create index
+	version          int               // Index version.  Must be incremented any time the index definition changes
+	previousVersions []int             // Previous versions of the index that will be removed during post_upgrade cleanup
+	required         bool              // Whether SG blocks on startup until this index is ready
+	readinessQuery   string            // Query used to determine view readiness
+	flags            SGIndexFlags      // Additional index options
+	CreationMode     IndexCreationMode // Signal when to create indexes
```
adamcfraser: I don't think this needs to be public, does it?

db/indexes.go
```diff
@@ -101,11 +124,17 @@ var (
 		IndexAllDocs:    "$sync.sequence, $sync.rev, $sync.flags, $sync.deleted",
 		IndexTombstones: "$sync.tombstoned_at",
 		IndexSyncDocs:   "META().id",
+		IndexUser:       "META().id, name, email, disabled",
+		IndexSession:    "META().id, username",
+		IndexRole:       "META().id, deleted",
```
adamcfraser: Would be preferable to include 'name' on IndexRole to avoid the key-based parsing and better support future pagination options.

db/query.go
```diff
@@ -186,7 +204,26 @@ var QueryRolesExcludeDeleted = SGQuery{
...
+		base.KeyspaceQueryAlias),
+	adhoc: false,
+}
+
+var QueryRolesExcludeDeletedUsingRoleIdx = SGQuery{
+	name: QueryTypeRolesExcludeDeleted,
+	statement: fmt.Sprintf(
+		"SELECT META(%s).id "+
+			"FROM %s AS %s "+
+			"USE INDEX($idx) "+
```
adamcfraser: It looks like we're defining a separate query here because QueryRolesExcludeDeleted includes the SyncDocWildcard check in the WHERE clause, is that accurate?  I'm curious why the existing QueryUsers does not have that in the WHERE clause - any idea?  If it's not actually necessary in the roles query, it would allow us to use the same query against either index.

db/indextest/util_testing.go
```diff
@@ -0,0 +1,81 @@
...
+		if err != nil && strings.Contains(err.Error(), "Index not exist") {
+			return err
+		}
+	}
+	return nil
+}
+
+// If certain environment variables are set, for example to turn on XATTR support, then update
+// the DatabaseContextOptions accordingly
+func AddOptionsFromEnvironmentVariables(dbcOptions *db.DatabaseContextOptions) {
```
adamcfraser: You can move this function into util_testing.go in the db package to avoid having to clone it here.

adamcfraser (changes requested):

db/indexes.go
nirav24: this seems weird. `CreationMode` was made private in https://github.com/couchbase/sync_gateway/pull/5835/commits/eaf55335223d6e32423772270da3696e884edf3d. although, I need to make `IndexCreationMode` private too. will do that

db/query.go
nirav24: ~I will need to test it to see why QueryRoles need SyncDocWildCard in where clause.~

seems like QueryRolesExcludeDeleted work without using SyncDocWildCard

db/database.go
```diff
@@ -998,9 +1002,122 @@ func (db *Database) processForEachDocIDResults(callback ForEachDocIDFunc, limit
 	return nil
 }
 
-// Returns the IDs of all users and roles
+// Returns the IDs of all users and roles, including deleted Roles
 func (db *DatabaseContext) AllPrincipalIDs(ctx context.Context) (users, roles []string, err error) {
 
+	if !db.IsServerless() || db.Options.UseViews {
+		return db.getAllPrincipalIDs(ctx)
```
adamcfraser: Let's rename this to getAllPrincipalIDsSyncDocs to avoid naming confusion between AllPrincipalIDs and getAllPrincipalIDs.

db/query.go
```diff
@@ -551,7 +594,42 @@ func (context *DatabaseContext) QueryRoles(ctx context.Context, startKey string,
...
+	params[QueryParamStartKey] = startKey
+
+	if limit > 0 {
+		queryStatement = fmt.Sprintf("%s LIMIT %d", queryStatement, limit)
+	}
+
+	return queryStatement, params
+}
+
+// Retrieves role ids using the roles index
```
adamcfraser: Can you update the comment to explicitly mention that the query result includes deleted roles?  It took me a minute to work out the difference between QueryRoles and QueryAllRoles.

db/query.go
```diff
@@ -551,7 +594,42 @@ func (context *DatabaseContext) QueryRoles(ctx context.Context, startKey string,
...
+func (context *DatabaseContext) QueryAllRoles(ctx context.Context, startKey string, limit int) (sgbucket.QueryResultIterator, error) {
+
+	// View Query
+	if context.Options.UseViews {
+		return nil, fmt.Errorf("QueryAllRoles doesn't support views")
```
adamcfraser, line 629: I'm not sure why ViewPrincipals doesn't support querying for all roles - can you remind me of those details?

db/indextest/util_testing.go
```diff
@@ -0,0 +1,69 @@
...
+Copyright 2020-Present Couchbase, Inc.
+
+Use of this software is governed by the Business Source License included in
+the file licenses/BSL-Couchbase.txt.  As of the Change Date specified in that
+file, in accordance with the Business Source License, use of this software will
+be governed by the Apache License, Version 2.0, included in the file
+licenses/APL2.txt.
+*/
+
+package indextest
```
adamcfraser: util_testing.go is generally intended for test functionality that needs to be made available cross-package.  Shared test functionality that's intended for use only within the package can just be defined in one of the test files.

db/indextest/database_test.go
```diff
@@ -0,0 +1,175 @@
+/*
```
adamcfraser: It feels like these tests could also live in indextest_test.go - any reason you think they need their own file?

db/indextest/util_testing.go
```diff
@@ -0,0 +1,69 @@
...
+func getDatabaseContextOptions(isServerless bool) db.DatabaseContextOptions {
+	defaultCacheOptions := db.DefaultCacheOptions()
+
+	return db.DatabaseContextOptions{
+		CacheOptions: &defaultCacheOptions,
+		Serverless:   isServerless,
+	}
+}
+
+func setupTestDBForBucketWithOptions(t testing.TB, dbcOptions db.DatabaseContextOptions) (*db.Database, context.Context) {
```
adamcfraser: For this in particular, I'd create a public version in db/util_testing.go for shared use by the db and db/indextest packages.

adamcfraser (changes requested):
Some organization comments, but also would like to revisit/discuss the limitations around role queries and views.

db/query.go
nirav24: `ViewPrincipals` contains both Users and Roles. This query is only meant for Roles. Also, as we discussed, views are deprecated and aren't supported in serverless mode.

When UseViews is enabled, the caller of this function `AllPrincipalIDs` fall backs to `QueryPrincipals` which uses `ViewPrincipals`.

db/indextest/database_test.go
nirav24: I created `db/indextest/database_test.go` just because these test lived in `db/database_test.go`. I don't have any strong reason other than this.

db/indextest/util_testing.go
nirav24: I see. I thought it would be a place to have common function defined that are used for testing. I will move these functions into test file.

rest/admin_api.go
```diff
@@ -1403,14 +1409,16 @@ func (h *handler) getUsers() error {
...
 
 func (h *handler) getRoles() error {
+	includeDeleted, _ := h.getOptBoolQuery(paramDeleted, false)
+
 	var roles []string
 	var err error
 
-	includeDeleted, _ := h.getOptBoolQuery(paramDeleted, false)
-	if includeDeleted {
+	// GetRoleIDs supports Views only when includeDeleted=false
```
adamcfraser: As discussed, move the conditional logic down to the db package so that it can have ownership over which view/index to choose to use for a given request.

adamcfraser (changes requested):

adamcfraser (approved):

PR #5836, opened 2022-10-26
CBG-2497 Support xattr storage of persistent config (forward port)

adamcfraser:
CBG-2497

Forward port of CBG-2033.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1002/

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -1289,6 +1289,9 @@ func SetupServerContext(ctx context.Context, config *StartupConfig, persistentCo
 // fetchAndLoadConfigs retrieves all database configs from the ServerContext's bootstrapConnection, and loads them into the ServerContext.
 // It will remove any databases currently running that are not found in the bucket.
 func (sc *ServerContext) fetchAndLoadConfigs(ctx context.Context, isInitialStartup bool) (count int, err error) {
+	sc.lock.Lock()
+	defer sc.lock.Unlock()
+
 	fetchedConfigs, err := sc.FetchConfigs(ctx, isInitialStartup)
 	if err != nil {
 		return 0, err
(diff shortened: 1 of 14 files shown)


rest/config.go
```diff
@@ -1289,6 +1289,9 @@ func SetupServerContext(ctx context.Context, config *StartupConfig, persistentCo
 // fetchAndLoadConfigs retrieves all database configs from the ServerContext's bootstrapConnection, and loads them into the ServerContext.
 // It will remove any databases currently running that are not found in the bucket.
 func (sc *ServerContext) fetchAndLoadConfigs(ctx context.Context, isInitialStartup bool) (count int, err error) {
+	sc.lock.Lock()
+	defer sc.lock.Unlock()
```
bbrks: I'm not sure this addition looks right - we're trying to get a read lock and/or a write lock later on in this function which will deadlock.

The 3.0.1 version of `fetchAndLoadConfigs` looks significantly different and does all of its locking like this in the entire function (which blocked on loading configs over the network). Since then it's changed to be more granular and only to be when applying config.

[#5600 CBG-2174: Narrow the lock in fetchAndLoadConfigs to avoid high response latency](https://github.com/couchbase/sync_gateway/pull/5600)

bbrks (changes requested):
Left a question around the `sc.lock` usage in `fetchAndLoadConfigs` which I think came from an incorrect merge, or a previous version of the function that did locking differently.

[#5600 CBG-2174: Narrow the lock in fetchAndLoadConfigs to avoid high response latency](https://github.com/couchbase/sync_gateway/pull/5600)

bbrks (approved):

PR #5840, opened 2022-10-27
add n1ql user queries execution auth tests

moshaad7:


toad-sage (approved):
LGTM

toad-sage:
I think here we have also covered negative cases

PR #5852, opened 2022-10-31
Added missing check and unit test case for MaxCodeSize

jineshparakh:
CBG-0000

Describe your PR here...
- Added missing check for MaxCodeSize and corresponding unit test case

## Pre-review checklist

- [ ] Link upstream PRs

### DIFF
--- a/db/functions/graphql.go
+++ b/db/functions/graphql.go
@@ -131,6 +131,9 @@ ResolverLoop:
 		for fieldName, fnConfig := range resolver {
 			if fnConfig.Args != nil {
 				err = fmt.Errorf("'args' is not valid in a GraphQL resolver config")
+			} else if config.MaxCodeSize != nil && len(fnConfig.Code) > *config.MaxCodeSize {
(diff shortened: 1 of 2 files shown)


db/functions/graphql.go
```diff
@@ -150,6 +150,11 @@ ResolverLoop:
 				multiError = multiError.Append(err)
 				break ResolverLoop // stop
 			}
+			if config.MaxCodeSize != nil && len(fnConfig.Code) > *config.MaxCodeSize {
```
snej: This block should go earlier, before the resolver gets compiled, so we don't try to compile a huge chunk of code. At line 134 would be good (just after validating `fnConfig.args`.
Also, don't break out of ResolverLoop; that prevents discovering any more errors. (The MaxResolverCount check does break out of the loop because once there are too many resolvers there's no point looking at any more.)

snej (changes requested):
Looks great aside from two things I noted inline...

db/functions/graphql.go
jineshparakh: Will change this

snej (approved):

snej:
Thanks!

PR #5856, opened 2022-10-31
CBG-2470: de duplicate wait for index code

gregns1:
CBG-2470

Gocb v2 has WatchIndexes function that will watch all index given to it till they come online. This PR will remove some of the duplicated code associated with waiting for indexes and replace it with the gocb v2 code available. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1055/

### DIFF
--- a/base/bucket_n1ql_test.go
+++ b/base/bucket_n1ql_test.go
@@ -35,7 +35,6 @@ func TestN1qlQuery(t *testing.T) {
 
 	bucket := GetTestBucket(t)
 	defer bucket.Close()
-	col, err := AsCollection(bucket)
 
 	n1qlStore, ok := AsN1QLStore(bucket)
 	if !ok {
@@ -54,13 +53,13 @@ func TestN1qlQuery(t *testing.T) {
 	}
 
 	indexExpression := "val"
-	err = n1qlStore.CreateIndex("testIndex_value", indexExpression, "", testN1qlOptions)
+	err := n1qlStore.CreateIndex("testIndex_value", indexExpression, "", testN1qlOptions)
 	if err != nil && err != ErrAlreadyExists {
 		t.Errorf("Error creating index: %s", err)
 	}
 
 	// Wait for index readiness
-	onlineErr := col.WaitForIndexesOnline([]string{"testIndex_value"}, false)
+	onlineErr := n1qlStore.WaitForIndexesOnline([]string{"testIndex_value"}, false)
 	if onlineErr != nil {
 		t.Fatalf("Error waiting for index to come online: %v", err)
 	}
@@ -81,7 +80,7 @@ func TestN1qlQuery(t *testing.T) {
 		}
 	}()
 
-	readyErr := col.WaitForIndexesOnline([]string{"testIndex_value"}, false)
+	readyErr := n1qlStore.WaitForIndexesOnline([]string{"testIndex_value"}, false)
 	require.NoError(t, readyErr, "Error validating index online")
 
 	// Query the index
@@ -144,8 +143,6 @@ func TestN1qlFilterExpression(t *testing.T) {
 
 	bucket := GetTestBucket(t)
 	defer bucket.Close()
-	col, err := AsCollection(bucket)
-	require.NoError(t, err)
 	n1qlStore, ok := AsN1QLStore(bucket)
 	if !ok {
 		t.Fatalf("Requires bucket to be N1QLStore")
@@ -164,13 +161,13 @@ func TestN1qlFilterExpression(t *testing.T) {
 
 	indexExpression := "val"
 	filterExpression := "val < 3"
-	err = n1qlStore.CreateIndex("testIndex_filtered_value", indexExpression, filterExpression, testN1qlOptions)
+	err := n1qlStore.CreateIndex("testIndex_filtered_value", indexExpression, filterExpression, testN1qlOptions)
 	if err != nil {
 		t.Fatalf("Error creating index: %s", err)
 	}
 
 	// Wait for index readiness
-	readyErr := col.WaitForIndexesOnline([]string{"testIndex_filtered_value"}, false)
+	readyErr := n1qlStore.WaitForIndexesOnline([]string{"testIndex_filtered_value"}, false)
 	require.NoError(t, readyErr, "Error validating index online")
 
 	// Defer index teardown
@@ -223,8 +220,6 @@ func TestIndexMeta(t *testing.T) {
 
 	bucket := GetTestBucket(t)
 	defer bucket.Close()
-	col, err := AsCollection(bucket)
-	require.NoError(t, err)
 	n1qlStore, ok := AsN1QLStore(bucket)
 	if !ok {
 		t.Fatalf("Requires bucket to be N1QLStore")
@@ -241,7 +236,7 @@ func TestIndexMeta(t *testing.T) {
 		t.Fatalf("Error creating index: %s", err)
 	}
 
-	readyErr := col.WaitForIndexesOnline([]string{"testIndex_value"}, false)
+	readyErr := n1qlStore.WaitForIndexesOnline([]string{"testIndex_value"}, false)
 	require.NoError(t, readyErr, "Error validating index online")
 
 	// Defer index teardown
@@ -269,8 +264,6 @@ func TestMalformedN1qlQuery(t *testing.T) {
 
 	bucket := GetTestBucket(t)
 	defer bucket.Close()
-	col, err := AsCollection(bucket)
-	require.NoError(t, err)
 	n1qlStore, ok := AsN1QLStore(bucket)
 	if !ok {
 		t.Fatalf("Requires bucket to be N1QLStore")
@@ -288,12 +281,12 @@ func TestMalformedN1qlQuery(t *testing.T) {
 	}
 
 	indexExpression := "val"
-	err = n1qlStore.CreateIndex("testIndex_value_malformed", indexExpression, "", testN1qlOptions)
+	err := n1qlStore.CreateIndex("testIndex_value_malformed", indexExpression, "", testN1qlOptions)
 	if err != nil {
 		t.Fatalf("Error creating index: %s", err)
 	}
 
-	readyErr := col.WaitForIndexesOnline([]string{"testIndex_value_malformed"}, false)
+	readyErr := n1qlStore.WaitForIndexesOnline([]string{"testIndex_value_malformed"}, false)
 	assert.NoError(t, readyErr, "Error validating index online")
 
 	// Defer index teardown
@@ -344,18 +337,16 @@ func TestCreateAndDropIndex(t *testing.T) {
 	bucket := GetTestBucket(t)
 	defer bucket.Close()
 	n1qlStore, ok := AsN1QLStore(bucket)
-	col, err := AsCollection(bucket)
-	require.NoError(t, err)
 	if !ok {
 		t.Fatalf("Requires bucket to be N1QLStore")
 	}
 
 	createExpression := SyncPropertyName + ".sequence"
-	err = n1qlStore.CreateIndex("testIndex_sequence", createExpression, "", testN1qlOptions)
+	err := n1qlStore.CreateIndex("testIndex_sequence", createExpression, "", testN1qlOptions)
 	if err != nil {
 		t.Fatalf("Error creating index: %s", err)
 	}
-	readyErr := col.WaitForIndexesOnline([]string{"testIndex_sequence"}, false)
+	readyErr := n1qlStore.WaitForIndexesOnline([]string{"testIndex_sequence"}, false)
 	assert.NoError(t, readyErr, "Error validating index online")
 
 	// Drop the index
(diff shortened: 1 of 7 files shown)


base/collection_n1ql.go
```diff
@@ -119,8 +122,44 @@ func (c *Collection) CreatePrimaryIndex(indexName string, options *N1qlIndexOpti
 	return CreatePrimaryIndex(c, indexName, options)
 }
 
-func (c *Collection) WaitForIndexOnline(indexName string) error {
-	return WaitForIndexOnline(c, indexName)
+// WaitForIndexOnline takes set of indexes and watches them till they're online.
+func (c *Collection) WaitForIndexOnline(indexNames []string, watchPrimary bool) error {
```
torcolvin: I think you can change this name to be `WaitForIndexes` or `WaitForIndexesOnline` and just have it take a slice of strings. We can look in the slice of strings to see if `PrimaryIndexName` is in it.

torcolvin: In fact we don't have to do this, because gocb does this for us:

https://github.com/couchbase/gocb/blob/v2.6.0/cluster_queryindexes.go#L727

base/collection_n1ql.go
```diff
@@ -119,8 +122,44 @@ func (c *Collection) CreatePrimaryIndex(indexName string, options *N1qlIndexOpti
...
+	mgr := cluster.QueryIndexes()
+
+	watchOption := gocb.WatchQueryIndexOptions{
+		WatchPrimary:   watchPrimary,
+		RetryStrategy:  gocb.NewBestEffortRetryStrategy(nil),
+		ScopeName:      c.ScopeName(),
+		CollectionName: c.Name(),
+	}
+
+	go func() {
```
torcolvin: Rather than putting this function in a goroutnine, I think we can replace `db.waitForIndex` - we'll still call this `cluster.QueryIndexes.WatchIndexes` under the hood, but we want to call it for a really long time. 

I think we could do something like:

```
resultSet, resultsError := bucket.Query(queryStatement, nil, base.RequestPlus, true)

		// Immediately close results. We don't need these
		if resultSet != nil {
			resultSetCloseError := resultSet.Close()
			if resultSetCloseError != nil {
				base.InfofCtx(logCtx, base.KeyAll, "Failed to close query results when verifying index %q availability for bucket %q", base.MD(indexName), base.MD(bucket.GetName()))
			}
		}

		if resultsError == nil {
			return nil
		}
```

with `WatchIndexes` where `RetryStrategy` is `goCBv2FailFastRetryStrategy`

base/collection_n1ql_common.go
```diff
@@ -241,7 +240,7 @@ func buildIndexes(s N1QLStore, indexNames []string) error {
 		InfofCtx(context.TODO(), KeyQuery, "Indexer error creating index - waiting for background build.  Error:%v", err)
 		// Wait for bucket to be created in background before returning
 		for _, indexName := range indexNames {
-			waitErr := s.WaitForIndexOnline(indexName)
+			waitErr := WaitForIndexOnline(s, indexName)
```
torcolvin: move this to waitforallindexes

torcolvin (commented):
I think the code might go up one level of removal. Because `WatchForIndexes` can work on multiple indexes at once and we always wait in a group - we could wait in a group.

It might be better to call `cluster.QueryIndexes().GetAllIndexes()` and then we can check `QueryIndex.State == "online"` like gocb does internally and log immediately as a particular index comes online. 

I think this might be out of the scope of this PR, but I was thinking you can entirely drop `GetIndexMeta` because this can be done by `QueryIndex` now.

base/collection_n1ql.go
```diff
@@ -119,8 +121,50 @@ func (c *Collection) CreatePrimaryIndex(indexName string, options *N1qlIndexOpti
 	return CreatePrimaryIndex(c, indexName, options)
 }
 
-func (c *Collection) WaitForIndexOnline(indexName string) error {
-	return WaitForIndexOnline(c, indexName)
+// WaitForIndexOnline takes set of indexes and watches them till they're online.
+func (c *Collection) WaitForIndexesOnline(indexNames []string, watchPrimary bool) error {
+	logCtx := context.TODO()
+	mgr := c.cluster.QueryIndexes()
+	maxNumAttempts := 250
```
torcolvin: we are going to have to increase this len(indexNames)

base/collection_n1ql.go
```diff
@@ -119,8 +121,50 @@ func (c *Collection) CreatePrimaryIndex(indexName string, options *N1qlIndexOpti
...
+		ScopeName:      c.ScopeName(),
+		CollectionName: c.Name(),
+	}
+	indexOption := gocb.GetAllQueryIndexesOptions{
+		ScopeName:      c.ScopeName(),
+		CollectionName: c.Name(),
+	}
+
+	for {
+		waitErr = mgr.WatchIndexes(c.BucketName(), indexNames, waitTime, &watchOption)
```
torcolvin: I'd just drop this and check each index in `GetAllIndexes` is online

base/collection_n1ql_common.go
```diff
@@ -218,33 +217,35 @@ func BuildDeferredIndexes(s N1QLStore, indexSet []string) error {
...
-	buildErr := buildIndexes(s, deferredIndexes)
+	buildErr := buildIndexes(bucket, s, deferredIndexes)
 	return buildErr
 }
 
 // BuildIndexes executes a BUILD INDEX statement in the current bucket, using the form:
 //
 //	BUILD INDEX ON `bucket.Name`(`index1`, `index2`, ...)
-func buildIndexes(s N1QLStore, indexNames []string) error {
+func buildIndexes(bucket Bucket, s N1QLStore, indexNames []string) error {
```
torcolvin: I think we want to keep N1QLStore if possible

db/indexes_test.go
```diff
@@ -117,6 +117,7 @@ func TestPostUpgradeIndexesSimple(t *testing.T) {
 	if base.TestsDisableGSI() {
 		t.Skip("This test only works with Couchbase Server and UseViews=false")
 	}
+	base.SetUpTestLogging(t, base.LevelInfo, base.KeyAll)
```
torcolvin: Delete before final commit

base/collection_n1ql.go
gregns1: I think we need this here as this is how we ensure that the indexes we specify/need are there and online where as the get all indexes is purely for the logging purposes. For example, i tried doing this changed the validate index code in the index_test.go to point to this function and tests failed due to a bucket having non xattr indexes when it was expecting them.

base/collection_n1ql_common.go
gregns1: I ensured I left the n1ql store in the function parameters so its still used. We need the bucket to be able to call the new WatchForIndexesOnline code.

base/collection_n1ql.go
torcolvin: I was suggesting basically keeping a list of onlineIndexes, and once onlineIndexes == indexNames we'll break out of this for loop

base/collection_n1ql_common.go
torcolvin: Can we add WatchForIndexes to N1QLStore code?

gregns1: Good point 🤦🏼‍♂️ that should work

base/collection_n1ql.go
gregns1: I think I need some thought into this. I tried increasing this * len(indexNames) but when in a situation when an index is not coming online as expected (I simulated this in a test) we end being stuck inside this loop for a VERY long time with over a 1000 retries and up to 5 seconds wait in between each retry (as defined by the sleeper function) whilst each retry it logs the currently online indexes so I'm wary its also logging a lot. Would this be okay? This I may need to make a change to only log each index as it comes online once thinking about it?

base/collection_n1ql.go
gregns1: Updated the PR with this approach.

base/collection_n1ql_common.go
```diff
@@ -174,7 +174,7 @@ func waitForIndexExistence(store N1QLStore, indexName string, shouldExist bool)
 }
 
 // BuildDeferredIndexes issues a build command for any deferred sync gateway indexes associated with the bucket.
-func BuildDeferredIndexes(s N1QLStore, indexSet []string) error {
+func BuildDeferredIndexes(bucket Bucket, s N1QLStore, indexSet []string) error {
```
torcolvin: now bucket is unnecessary

base/leaky_bucket.go
```diff
@@ -282,6 +282,15 @@ func (b *LeakyBucket) GetWithXattr(k string, xattr string, userXattrKey string,
 	return b.bucket.GetWithXattr(k, xattr, userXattrKey, rv, xv, uxv)
 }
 
+func (b *LeakyBucket) WaitForIndexesOnline(indexNames []string, watchPrimary bool) error {
+	bucket := b.GetUnderlyingBucket()
+	col, err := AsCollection(bucket)
```
torcolvin: Can we do it like we used to to unwrap it to a `N1QLStore` now that you put this back on N1QLStore?

base/collection_n1ql.go
```diff
@@ -119,8 +121,57 @@ func (c *Collection) CreatePrimaryIndex(indexName string, options *N1qlIndexOpti
...
+			}
+		}
+
+		for _, listVal := range indexNames {
+			if logged[listVal] == true {
+				onlineIndexCount++
+			}
+		}
+
+		if onlineIndexCount == len(indexNames) {
```
torcolvin: if you rename `logged` to `onlineIndexes` you actually know onlineIndexCount with `len(onlineIndexes)`

base/collection_n1ql.go
```diff
@@ -119,8 +121,57 @@ func (c *Collection) CreatePrimaryIndex(indexName string, options *N1qlIndexOpti
...
+	for {
+		onlineIndexCount := 0
+		currIndexes, err := mgr.GetAllIndexes(c.BucketName(), &indexOption)
+		if err != nil {
+			return err
+		}
+
+		for i := 0; i < len(currIndexes); i++ {
+			if currIndexes[i].State == IndexStateOnline {
+				if logged[currIndexes[i].Name] == false {
```
torcolvin: nit: go idiomatic is `!logged[currIndexes[i].Name]` instead of `logged[currIndexes[i].name] == false`

db/indexes.go
```diff
@@ -283,7 +282,7 @@ func (i *SGIndex) createIfNeeded(bucket base.N1QLStore, useXattrs bool, numRepli
 }
 
 // Initializes Sync Gateway indexes for bucket.  Creates required indexes if not found, then waits for index readiness.
-func InitializeIndexes(bucket base.N1QLStore, useXattrs bool, numReplicas uint, failFast bool) error {
```
torcolvin: I would not have this take `bucket base.Bucket` parameter. This is about to change with CBG-2266,  but `base.Bucket` is an interface that is shared with `N1QLStore` (also an interface) so the really would be the same object and it's weird to pass them both in the same function. Also I think bucket is no longer needed now.

If I would have to pass both, I'd consider whether I could pass the interface that was the superset (base.Bucket) in this case.

base/bucket_n1ql_test.go
```diff
@@ -530,6 +547,8 @@ func TestCreateAndDropIndexErrors(t *testing.T) {
 
 	bucket := GetTestBucket(t)
 	defer bucket.Close()
+	//col, err := AsCollection(bucket)
```
torcolvin: remove dead code

torcolvin (commented):
I think we're almost there, the one change I'd like to see in this PR is now we can use these functions to replace the code in `db/indexes_test.go` to have one copy of the code.

- `validateAllIndexesOnline` --> `InitializeIndexes`

base/collection_n1ql.go
```diff
@@ -119,8 +119,57 @@ func (c *Collection) CreatePrimaryIndex(indexName string, options *N1qlIndexOpti
...
+		maxNumAttempts = 1
+	}
+	retrySleeper := CreateMaxDoublingSleeperFunc(maxNumAttempts, 100, 5000)
+	retryCount := 0
+
+	onlineIndexes := make(map[string]bool)
+
+	indexOption := gocb.GetAllQueryIndexesOptions{
+		ScopeName:      c.ScopeName(),
+		CollectionName: c.Name(),
```
torcolvin, line 137: one last change I think to add noretries to this strategy becuase we'll handle retries on our own retrysleeper

base/collection_n1ql_common.go
```diff
@@ -240,36 +240,16 @@ func buildIndexes(s N1QLStore, indexNames []string) error {
 	if IsIndexerRetryBuildError(err) {
 		InfofCtx(context.TODO(), KeyQuery, "Indexer error creating index - waiting for background build.  Error:%v", err)
 		// Wait for bucket to be created in background before returning
-		for _, indexName := range indexNames {
-			waitErr := s.WaitForIndexOnline(indexName)
-			if waitErr != nil {
-				return waitErr
-			}
+		waitErr := s.WaitForIndexesOnline(indexNames, false)
```
torcolvin: quick nit:

`return s.WaitForIndexesOnline(indexNames, false)`

base/collection_n1ql.go
gregns1: I have added the fail fast parameter here already to allow for it to only execute once.

torcolvin: The failfast parameter is different than the timeout.

All kv ops use timeouts and retry strategy based on what they inherit when you create the cluster. As per https://github.com/couchbase/sync_gateway/commit/46b733faf1180db7c055bd97a421f5c4e1e9ea0d

we retry operations up to `DefaultGocbV2OperationTimeout = 10 * time.Second` or optionally overriden by `BucketOpTimeoutMs` in db config.

gregns1: Ah i get you, think I again misunderstood what you meant sorry

torcolvin (approved):

PR #5867, opened 2022-10-31
Functional and Unit Tests for graphQL and User Functions

jineshparakh:
CBG-0000

Describe your PR here...
- Refactored functions_api_tests.go and moved it to the new functionsapitest package.
- Added functional test cases (both positive and negative) for user-functions and graphQL.
- Added missing unit test cases for graphQL. Improved coverage from ~50% to >90%

## Pre-review checklist

- [ ] Link upstream PRs


rest/functionsapitest/common_test.go
```diff
@@ -0,0 +1,272 @@
...
+				},
+			},
+		},
+	},
+}}
+
+// Asserts that running testFunc in 100 concurrent goroutines is no more than 10% slower
+// than running it 100 times in succession. A low bar indeed, but can detect some serious
+// bottlenecks, or of course deadlocks.
+func testConcurrently(t *testing.T, rt *rest.RestTester, testFunc func() bool) bool {
```
snej: This is a utility function that should probably go in `utilities_testing_functions_api.go`

rest/functionsapitest/common_test.go
```diff
@@ -0,0 +1,272 @@
...
+	wg.Wait()
+	concurrentDuration := time.Since(startTime)
+
+	log.Printf("---- %d sequential took %v, concurrent took %v ... speedup is %f",
+		numTasks, sequentialDuration, concurrentDuration,
+		float64(sequentialDuration)/float64(concurrentDuration))
+	return assert.LessOrEqual(t, concurrentDuration, 1.1*numTasks*sequentialDuration)
+}
+
+func TestFunctions(t *testing.T) {
```
snej: This is a GraphQL test; shouldn't it go in graphql_test.go?

rest/functionsapitest/common_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+	t.Run("Function", func(t *testing.T) {
+		testConcurrently(t, rt, func() bool {
+			response := rt.SendRequest("GET", "/db/_function/square?n=13", "")
+			return assert.Equal(t, 200, response.Result().StatusCode) &&
+				assert.Equal(t, "169", string(response.BodyBytes()))
+		})
+	})
+
+	t.Run("GraphQL", func(t *testing.T) {
```
snej: This is a GraphQL test; shouldn't it go in graphql_test.go?

snej (changes requested):
It's very hard to tell what's changed: everything's been moved to new source files, so the diff just shows everything as new. I know this isn't your fault, you were told to move the tests. In hindsight I realize it would have been better to do this in two parts, like moving the existing tests first, getting that reviewed, and then adding the new tests. I'll do my best to try and find all the new code in here.

You've combined the admin (config) API tests in the same source files with the other tests. I think they should stay separate, as they were. Having four source files would be fine -- functions admin, graphql admin, functions queries, graphql queries.

~~Also, I don't see any new tests of functions here, just the two simple calls in the pre-existing concurrent test. Are those TBD?~~ Never mind -- I just didn't see them because Github wasn't showing that source file.

rest/functionsapitest/user_functions_test.go
```diff
@@ -0,0 +1,1463 @@
...
+
+		assert.Nil(t, rt.GetDatabase().Options.UserFunctions.Definitions["square"])
+		assert.Equal(t, 1, len(rt.GetDatabase().Options.UserFunctions.Definitions))
+
+		response = rt.SendAdminRequest("GET", "/db/_function/square?n=13", "")
+		assert.Equal(t, 404, response.Result().StatusCode)
+	})
+}
+
+func TestUserFunctionsMaxFunctionCount(t *testing.T) {
```
snej: This is already tested in the functions package, and testing it here is redundant.

rest/functionsapitest/user_functions_test.go
```diff
@@ -0,0 +1,1463 @@
...
+		request, err := json.Marshal(testFunctionConfig)
+		assert.NoError(t, err)
+
+		response := rt.SendAdminRequest("PUT", "/db/_config/functions", string(request))
+		assert.Equal(t, 400, response.Result().StatusCode)
+		assert.Contains(t, string(response.BodyBytes()), "too many functions declared")
+	})
+}
+
+func TestUserFunctionsMaxCodeSize(t *testing.T) {
```
snej: This is already tested in the functions package, and testing it here is redundant.

rest/functionsapitest/user_functions_test.go
```diff
@@ -0,0 +1,1463 @@
...
+			assert.NoError(t, err)
+			assert.Equal(t, 201, resp.Result().StatusCode)
+		})
+
+		// negative cases:
+		t.Run("using single dollar", func(t *testing.T) {
+			resp, err := createDB(t, userFuncConfigSingleDollar)
+			assert.NoError(t, err)
+			assert.Equal(t, 500, resp.Result().StatusCode)
+			assert.Contains(t, string(resp.BodyBytes()), "undefined environment variable")
```
snej: I don't think this is a safe test; it literally depends on whether there is an environment variable named `user` in the process running the test. If there happened to be one, it would break the test.

I don't think you should be testing this. It's basically a bug, not a feature. (Same goes for the next subtest.)

rest/functionsapitest/user_functions_test.go
```diff
@@ -0,0 +1,1463 @@
...
+
+	t.Run("Delete a non-existing function and check HTTP Status Code", func(t *testing.T) {
+		response := rt.SendAdminRequest("DELETE", "/db/_config/functions/foo", "")
+		assert.Equal(t, 404, response.Result().StatusCode)
+	})
+}
+
+/// ILLEGAL SYNTAX TESTS
+
+func TestIllegalSyntax(t *testing.T) {
```
snej: This is not testing the REST package. All it does is send a config to the functions package and let that return an error, which is already tested in the functions package.

rest/functionsapitest/user_functions_test.go
```diff
@@ -0,0 +1,1463 @@
...
+		assert.Contains(t, string(response.BodyBytes()), "Internal Server Error")
+		assert.Contains(t, string(response.BodyBytes()), "syntax error")
+	})
+}
+
+//////// FUNCTIONS EXECUTION API TESTS
+
+/// AUTH TESTS
+
+func createUserAlice(t *testing.T, rt *rest.RestTester) (string, string) {
```
snej: All of this from here to line 1200 or so is just copied-and-pasted, with adaptations, from the functions package. Testing it again here is redundant.

snej (changes requested):

rest/functionsapitest/graphql_test.go
```diff
@@ -0,0 +1,863 @@
...
+
+const kDummyGraphQLSchema = `
+    type Query {
+        square(n: Int!) : Int!
+    }`
+
+var allowAll = &functions.Allow{Channels: []string{"*"}}
+
+// The GraphQL schema:
+var kTestGraphQLSchema = `
```
snej: Much of this source file is copied and pasted from the functions package unit tests. That is not a good idea:
1. Testing the same functionality (just called through a different API) is redundant and doesn't improve coverage.
2. Even if you're using the config to test something different (like parameter marshaling), the use of the exact same config and logic in the functions guarantees that the functions package isn't doing anything different than in the unit tests.

snej (changes requested):

snej:
These tests should be checking the functionality of the `rest` package, i.e. how an incoming request gets sent to the `functions` package. That means
1. Updating and retrieving the configuration
2. Routing requests, i.e. mapping URL paths to the right handlers for functions or GraphQL
3. Extracting the function name, GraphQL query, and arguments and passing them to the call
4. Returning results and errors

You should **not** be testing functionality of the actual functions or GraphQL, the stuff that's already tested in the `functions` unit tests. You just need to test that everything gets passed to `functions` and returned from it correctly.

moshaad7:
@snej  
Temporarily,  we have copied the value of  memWatermarkThresholdMB from other files, 
we will need the value of memWatermarkThresholdMB for functionsapitests

`func TestMain(m *testing.M) {
	memWatermarkThresholdMB := uint64(8192)
	db.TestBucketPoolWithIndexes(m, memWatermarkThresholdMB)
}`

rest/functionsapitest/user_functions_test.go
moshaad7: instead of an using an environment variable named `user` , 
what if I keep on generating a random variable until it is not present in the list of environment variables. 

Can we use that to test this ? 
I am asking this because, I feel like we should test if $$variables are properly parsed and passed to functions package as expected.

toad-sage:
> These tests should be checking the functionality of the `rest` package, i.e. how an incoming request gets sent to the `functions` package. That means
> 
> 1. Updating and retrieving the configuration
> 2. Routing requests, i.e. mapping URL paths to the right handlers for functions or GraphQL
> 3. Extracting the function name, GraphQL query, and arguments and passing them to the call
> 4. Returning results and errors
> 
> You should **not** be testing functionality of the actual functions or GraphQL, the stuff that's already tested in the `functions` unit tests. You just need to test that everything gets passed to `functions` and returned from it correctly.

@snej 
in 2nd point , how can we know whether the request is routing the correct handler or not ?
in 3rd point , 
how can we find the arguments which is being passed in GraphQL query ?
the response that we are receiving when we hit this endpoint (/db/_config/graphql) contains schemas and resolvers in form
of text. (In case of functions we do receive the "args" field which contains the arguments required , but don't know anything which provides with args in case of GraphQL).

snej:
@toad-sage Sorry, I was unclear. The bullet points describe the job of the `rest` package _itself_. The job of your tests is to check that it is doing those 4 things correctly.

snej:
> but don't know anything which provides with args in case of GraphQL

GraphQL calls them "variables". They're in an optional key of the request JSON.

snej:
> @snej Temporarily, we have copied the value of memWatermarkThresholdMB from other files, we will need the value of memWatermarkThresholdMB for functionsapitests

Sorry, I don't know anything about that parameter; you should ask one of the regular SG engineers about that.

rest/functionsapitest/user_functions_admin_test.go
```diff
@@ -462,6 +462,77 @@ func TestFunctionsConfigAsPartOfDBConfig(t *testing.T) {
...
+		testFunctionConfig.MaxRequestSize = base.IntPtr(5)
+
+		request, err := json.Marshal(testFunctionConfig)
+		assert.NoError(t, err)
+
+		response := rt.SendAdminRequest("PUT", "/db/_config/functions", string(request))
+		assert.Equal(t, 200, response.Result().StatusCode)
+
+		t.Run("js function", func(t *testing.T) {
+			response = rt.SendAdminRequest("POST", "/db/_function/multiply", `{"first":1,"second":2,"third":3,"fourth":4}`)
```
snej: Check GET requests too.

(And I don't think it's necessary to test N1QL as well as JS. The `rest` package is what checks the request size, and it doesn't know the difference between the two.)

rest/utilities_testing_functions_api_test.go
```diff
@@ -21,6 +19,8 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+//////// Concurrency Tests
```
snej, line 22: This comment looks like a leftover and should be deleted.

rest/functionsapitest/graphql_queries_test.go
```diff
@@ -0,0 +1,260 @@
...
+	defer rt.Close()
+
+	t.Run("AsAdmin - getUser", func(t *testing.T) {
+		response := rt.SendAdminRequest("POST", "/db/_graphql", `{"query": "query($id:ID!){ getUser(id:$id) { id , name } }" , "variables": {"id": 1}}`)
+		assert.Equal(t, 200, response.Result().StatusCode)
+		assert.Equal(t, `{"data":{"getUser":{"id":"1","name":"user1"}}}`, string(response.BodyBytes()))
+
+		queryParam := `query($id:ID!){ getUser(id:$id) { id , name } }`
+		variableParam := `{"id": 1}`
+		getRequestUrl := fmt.Sprintf("/db/_graphql?query=%s&variables=%s", queryParam, variableParam)
```
snej: This does not produce a valid URL -- it contains lots of illegal characters like spaces. The `queryParam` and `variableParam` need to be URL-encoded.

(I'm surprised the test passes -- that may be because we're not really sending an HTTP request but creating a Request object dynamically, and that may bypass some of the URL parsing.)

rest/functionsapitest/graphql_queries_test.go
```diff
@@ -0,0 +1,260 @@
...
+		response = rt.SendAdminRequest("GET", getRequestUrl, "")
+		assert.Equal(t, 200, response.Result().StatusCode)
+		assert.Equal(t, `{"data":{"getUser":{"id":"1","name":"user1"}}}`, string(response.BodyBytes()))
+	})
```
snej, line 61: There is a third way to query that also needs to be tested: the body of the POST is in GraphQL syntax and the Content-Type is `application/graphql`. (See rest/functions_api.go:168.)

snej (changes requested):

rest/utilities_testing_functions_api_test.go
jineshparakh: With respect to the last review comments, I've moved the concurrency tests to this new file.
Below this comment, there is the config and all the concurrency tests written.
Should I make some changes to it?

rest/functionsapitest/graphql_queries_test.go
jineshparakh: Will change this

rest/functionsapitest/graphql_queries_test.go
jineshparakh: Added the test.

snej (approved):

PR #5870, opened 2022-11-01
CBG-1891: use dcp stream to resync docs

nirav24:
CBG-1891

Describe your PR here...
- Using one shot DCP stream for Resync Operation.

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1249/

### DIFF
--- a/db/background_mgr.go
+++ b/db/background_mgr.go
@@ -47,6 +47,7 @@ const (
 // BackgroundManager this is the over-arching type which is exposed in DatabaseContext
 type BackgroundManager struct {
 	BackgroundManagerStatus
+	name                                   string
 	lastError                              error
 	terminator                             *base.SafeTerminator
 	backgroundManagerStatusUpdateWaitGroup sync.WaitGroup
@@ -94,6 +95,11 @@ type BackgroundManagerProcessI interface {
 }
 
 type updateStatusCallbackFunc func() error
+
+// GetName returns name of the background manager
+func (b *BackgroundManager) GetName() string {
+	return b.name
+}
 
 func (b *BackgroundManager) Start(ctx context.Context, options map[string]interface{}) error {
 	err := b.markStart()
--- a/db/background_mgr.go
+++ b/db/background_mgr.go
@@ -390,8 +390,7 @@ func (b *BackgroundManager) setRunState(state BackgroundProcessState) {
 	b.State = state
 }
 
-// Currently only test
-func (b *BackgroundManager) GetRunState(t testing.TB) BackgroundProcessState {
+func (b *BackgroundManager) GetRunState() BackgroundProcessState {
 	b.lock.Lock()
 	defer b.lock.Unlock()
 	return b.State
--- a/db/background_mgr_resync_dcp.go
+++ b/db/background_mgr_resync_dcp.go
@@ -137,8 +137,11 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]interface
 	if err != nil {
 		return err
 	}
-	collectionID := base.DefaultCollectionID
-	collectionIDs := []uint32{collectionID}
+
+	collectionIDs := make([]uint32, 0, len(db.CollectionByID))
+	for collectionID := range db.CollectionByID {
+		collectionIDs = append(collectionIDs, collectionID)
+	}
 
 	clientOptions, err := getReSyncDCPClientOptions(collectionIDs, db.Options.GroupID)
 	if err != nil {
--- a/db/background_mgr_resync_dcp.go
+++ b/db/background_mgr_resync_dcp.go
@@ -1,0 +1,284 @@
+//  Copyright 2012-Present Couchbase, Inc.
+//
+//  Use of this software is governed by the Business Source License included
+//  in the file licenses/BSL-Couchbase.txt.  As of the Change Date specified
+//  in that file, in accordance with the Business Source License, use of this
+//  software will be governed by the Apache License, Version 2.0, included in
+//  the file licenses/APL2.txt.
+
+package db
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+	"sync/atomic"
+
+	sgbucket "github.com/couchbase/sg-bucket"
+	"github.com/couchbase/sync_gateway/base"
+	"github.com/google/uuid"
+)
+
+// =====================================================================
+// Resync Implementation of Background Manager Process using DCP stream
+// =====================================================================
+
+type ResyncManagerDCP struct {
+	DocsProcessed base.AtomicInt
+	DocsChanged   base.AtomicInt
+	ResyncID      string
+	VBUUIDs       []uint64
+	lock          sync.Mutex
+}
+
+var _ BackgroundManagerProcessI = &ResyncManagerDCP{}
+
+func NewResyncManagerDCP(metadataStore base.DataStore) *BackgroundManager {
+	return &BackgroundManager{
+		Process: &ResyncManagerDCP{},
+		clusterAwareOptions: &ClusterAwareBackgroundManagerOptions{
+			metadataStore: metadataStore,
+			processSuffix: "resync",
+		},
+		terminator: base.NewSafeTerminator(),
+	}
+}
+
+func (r *ResyncManagerDCP) Init(ctx context.Context, options map[string]interface{}, clusterStatus []byte) error {
+	newRunInit := func() error {
+		uniqueUUID, err := uuid.NewRandom()
+		if err != nil {
+			return err
+		}
+
+		r.ResyncID = uniqueUUID.String()
+		base.InfofCtx(ctx, base.KeyAll, "Resync: Starting new resync run with resync ID: %q", r.ResyncID)
+		return nil
+	}
+
+	if clusterStatus != nil {
+		var statusDoc ResyncManagerStatusDocDCP
+		err := base.JSONUnmarshal(clusterStatus, &statusDoc)
+
+		reset, ok := options["reset"].(bool)
+		if reset && ok {
+			base.InfofCtx(ctx, base.KeyAll, "Resync: Resetting resync process. Will not  resume any "+
+				"partially completed process")
+		}
+
+		// If the previous run completed, or there was an error during unmarshalling the status we will start the
+		// process from scratch with a new resync ID. Otherwise, we should resume with the resync ID, stats specified in the doc.
+		if statusDoc.State == BackgroundProcessStateCompleted || err != nil || (reset && ok) {
+			return newRunInit()
+		} else {
+			r.ResyncID = statusDoc.ResyncID
+			r.DocsChanged.Set(int64(statusDoc.DocsChanged))
+			r.DocsProcessed.Set(int64(statusDoc.DocsProcessed))
+			r.VBUUIDs = statusDoc.VBUUIDs
+
+			base.InfofCtx(ctx, base.KeyAll, "Resync: Attempting to resume resync with resycn ID: %q", r.ResyncID)
+		}
+
+		return nil
+	}
+
+	return newRunInit()
+}
+
+func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]interface{}, persistClusterStatusCallback updateStatusCallbackFunc, terminator *base.SafeTerminator) error {
+	db := options["database"].(*Database)
+	regenerateSequences := options["regenerateSequences"].(bool)
+
+	resyncLoggingID := "Resync: " + r.ResyncID
+
+	persistClusterStatus := func() {
+		err := persistClusterStatusCallback()
+		r.SetStats(int(r.DocsProcessed.Value()), int(r.DocsChanged.Value()))
+		if err != nil {
+			base.WarnfCtx(ctx, "[%s] Failed to persist cluster status on-demand for resync operation: %v", resyncLoggingID, err)
+		}
+	}
+	defer persistClusterStatus()
+
+	defer atomic.CompareAndSwapUint32(&db.State, DBResyncing, DBOffline)
+
+	callback := func(event sgbucket.FeedEvent) bool {
+		var err error
+		docID := string(event.Key)
+		key := realDocID(docID)
+		base.TracefCtx(ctx, base.KeyAll, "[%s] Received DCP event %d for doc %v", resyncLoggingID, event.Opcode, base.UD(docID))
+		// Don't want to process raw binary docs
+		// The binary check should suffice but for additional safety also check for empty bodies
+		if event.DataType == base.MemcachedDataTypeRaw || len(event.Value) == 0 {
+			return true
+		}
+
+		// We only want to process full docs. Not any sync docs.
+		if strings.HasPrefix(docID, base.SyncDocPrefix) {
+			return true
+		}
+
+		r.DocsProcessed.Add(1)
+
+		databaseCollection := db.CollectionByID[event.CollectionID]
+		_, unusedSequences, err := (&DatabaseCollectionWithUser{
+			DatabaseCollection: databaseCollection,
+		}).updateDocument(ctx, docID, key, regenerateSequences, []uint64{})
+
+		databaseCollection.releaseSequences(ctx, unusedSequences)
+
+		if err == nil {
+			r.DocsChanged.Add(1)
+		} else if err != base.ErrUpdateCancel {
+			base.WarnfCtx(ctx, "[%s] Error updating doc %q: %v", resyncLoggingID, base.UD(docID), err)
+		}
+		return true
+	}
+
+	collection, err := base.AsGocbV2Bucket(db.Bucket)
+	if err != nil {
+		return err
+	}
+	collectionID := base.DefaultCollectionID
+	collectionIDs := []uint32{collectionID}
+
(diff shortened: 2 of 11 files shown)


base/stats.go
```diff
@@ -461,6 +461,9 @@ type DatabaseStats struct {
 	// The total time spent evaluating the sync_function.
 	SyncFunctionTime *SgwIntStat `json:"sync_function_time"`
 
+	// The resync_start_time.
+	ResyncStartTime *SgwIntStat `json:"resync_start_time"`
```
adamcfraser: Given that we already have a background process for resync that doesn't use start time, I'm not sure we need a new stat if we're just changing the underlying implementation.

db/background_mgr_resync_dcp.go
```diff
@@ -0,0 +1,282 @@
...
+		base.InfofCtx(ctx, base.KeyAll, "[%s] Finished re-running sync function. %d/%d docs changed", resyncLoggingID, r.DocsChanged.Value(), r.DocsProcessed.Value())
+		err = dcpClient.Close()
+		if err != nil {
+			base.WarnfCtx(ctx, "[%s] Failed to close resync DCP client! %v", resyncLoggingID, err)
+			return err
+		}
+
+		db.releaseSequences(ctx, unusedSequences)
+		if regenerateSequences {
+			if err := db.updateAllPrincipalsSequences(ctx); err != nil {
```
adamcfraser: The user/role documents we want to update are part of the DCP feed (we're omitting them on line 124).  Would be preferable to process as part of DCP feed instead of as a second step.

base/stats.go
nirav24: sure

db/background_mgr_resync_dcp.go
nirav24: I was thinking of processing user/role from DCP feed too but If I understood correctly, current implementation only processes user/role after it completes processing all docs. If the process was interrupted in the middle, it will not release sequences, update user/role and invalidate user cache as per https://github.com/couchbase/sync_gateway/blob/6a13939608760c221fbd436ee595a6b56f3a9b7e/db/database.go#L1740-L1790

adamcfraser: As discussed, I think it's reasonable to leave principal sequence handling the way it is (processed after documents).

db/database.go
```diff
@@ -644,7 +644,7 @@ func NewDatabaseContext(ctx context.Context, dbName string, bucket base.Bucket,
 		// No cleanup necessary, stop heartbeater above will take care of it
 	}
 
-	dbContext.ResyncManager = NewResyncManager(bucket)
+	dbContext.ResyncManager = NewResyncManagerDCP(bucket)
```
nirav24: @adamcfraser we potentially need to discuss if we are going to toggle between `NewResyncManager` and `NewResyncManagerDCP`. I remember you wanted to do a performance test. We should discuss that

db/background_mgr.go
```diff
@@ -390,8 +390,7 @@ func (b *BackgroundManager) setRunState(state BackgroundProcessState) {
 	b.State = state
 }
 
-// Currently only test
-func (b *BackgroundManager) GetRunState(t testing.TB) BackgroundProcessState {
+func (b *BackgroundManager) GetRunState() BackgroundProcessState {
```
adamcfraser, line 399: This doesn't look like it needs to be made non-test for this PR.

db/background_mgr_resync_dcp.go
```diff
@@ -0,0 +1,298 @@
...
+
+		// If the previous run completed, or there was an error during unmarshalling the status we will start the
+		// process from scratch with a new resync ID. Otherwise, we should resume with the resync ID, stats specified in the doc.
+		if statusDoc.State == BackgroundProcessStateCompleted || err != nil || (reset && ok) {
+			return newRunInit()
+		} else {
+			r.ResyncID = statusDoc.ResyncID
+			r.SetStatus(statusDoc.DocsChanged, statusDoc.DocsProcessed)
+
+			base.InfofCtx(ctx, base.KeyAll, "Resync: Attempting to resume resync with resycn ID: %s", r.ResyncID)
```
adamcfraser:
```suggestion
			base.InfofCtx(ctx, base.KeyAll, "Resync: Attempting to resume resync with resync ID: %s", r.ResyncID)
```

db/background_mgr_resync_dcp.go
```diff
@@ -0,0 +1,298 @@
...
+		return nil
+	}
+
+	if clusterStatus != nil {
+		var statusDoc ResyncManagerStatusDocDCP
+		err := base.JSONUnmarshal(clusterStatus, &statusDoc)
+
+		reset, ok := options["reset"].(bool)
+		if reset && ok {
+			base.InfofCtx(ctx, base.KeyAll, "Resync: Resetting resync process. Will not  resume any "+
```
adamcfraser:
```suggestion
			base.InfofCtx(ctx, base.KeyAll, "Resync: Resetting resync process. Will not resume any "+
```

db/background_mgr_resync_dcp.go
```diff
@@ -0,0 +1,298 @@
...
+	VBUUIDs []uint64 `json:"vbuuids"`
+}
+
+type ResyncManagerStatusDocDCP struct {
+	ResyncManagerResponseDCP `json:"status"`
+	ResyncManagerMeta        `json:"meta"`
+}
+
+// getReSyncDCPClientOptions returns the default set of DCPClientOptions suitable for resync
+func getReSyncDCPClientOptions(collection *base.Collection, groupID string) (*base.DCPClientOptions, error) {
```
adamcfraser:
```suggestion
func getResyncDCPClientOptions(collection *base.Collection, groupID string) (*base.DCPClientOptions, error) {
```

db/background_mgr_resync_dcp_test.go
```diff
@@ -0,0 +1,410 @@
...
+	require.NoError(t, err)
+
+	stats := getResyncStats(resyncMrg.Process)
+	assert.Less(t, stats.DocsProcessed, int64(docsToCreate), "DocsProcessed is equal to docs created. Consider setting docsToCreate > %d.", docsToCreate)
+	assert.Less(t, stats.DocsChanged, int64(docsToCreate))
+
+	assert.Less(t, db.DbStats.Database().SyncFunctionCount.Value(), int64(docsToCreate))
+}
+
+func TestResycnManagerDCPStart(t *testing.T) {
```
adamcfraser:
```suggestion
func TestResyncManagerDCPStart(t *testing.T) {
```

db/background_mgr_resync_dcp_test.go
```diff
@@ -0,0 +1,410 @@
...
+	defer resyncMrg.resetStatus()
+
+	options := make(map[string]interface{})
+	options["database"] = db
+	options["regenerateSequences"] = false
+
+	err := resyncMrg.Start(ctx, options)
+	require.NoError(t, err)
+	go func() {
+		time.Sleep(2 * time.Second)
```
adamcfraser: We don't want to rely on sleep timing in tests like this - in different environments the feed could either not start, or have already completed.

db/background_mgr_resync_dcp_test.go
```diff
@@ -0,0 +1,410 @@
...
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("Test requires Couchbase Server")
+	}
+	base.LongRunningTest(t)
+
+	docsToCreate := 10_000
+	db, ctx := setupTestDBForResyncWithDocs(t, docsToCreate, true)
+	defer db.Close(ctx)
+
+	resyncMrg := NewResyncManagerDCP(db.Bucket)
```
adamcfraser:
```suggestion
	resyncMgr := NewResyncManagerDCP(db.Bucket)
```

db/background_mgr_resync_dcp_test.go
```diff
@@ -0,0 +1,410 @@
...
+	assert.Less(t, db.DbStats.Database().SyncFunctionCount.Value(), int64(docsToCreate))
+}
+
+func TestResycnManagerDCPStart(t *testing.T) {
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("Test requires Couchbase Server")
+	}
+	base.LongRunningTest(t)
+
+	t.Run("Resync without updating resync function", func(t *testing.T) {
```
adamcfraser:
```suggestion
	t.Run("Resync without updating sync function", func(t *testing.T) {
```

db/background_mgr_resync_dcp_test.go
```diff
@@ -0,0 +1,410 @@
...
+		require.NoError(t, err)
+
+		stats := getResyncStats(resyncMrg.Process)
+		assert.Equal(t, int64(docsToCreate), stats.DocsProcessed)
+		assert.Equal(t, int64(0), stats.DocsChanged)
+
+		assert.Equal(t, int(db.DbStats.Database().SyncFunctionCount.Value()), docsToCreate)
+	})
+
+	t.Run("Resync with updated resync function", func(t *testing.T) {
```
adamcfraser:
```suggestion
	t.Run("Resync with updated sync function", func(t *testing.T) {
```

db/background_mgr_resync_dcp_test.go
```diff
@@ -0,0 +1,410 @@
...
+	require.NoError(t, err)
+
+	stats = getResyncStats(resyncMrg.Process)
+	assert.GreaterOrEqual(t, stats.DocsProcessed, int64(docsToCreate))
+	assert.Equal(t, int64(docsToCreate), stats.DocsChanged)
+
+	assert.GreaterOrEqual(t, db.DbStats.Database().SyncFunctionCount.Value(), int64(docsToCreate))
+}
+
+// helper function to insert documents equals to docsToCreate, and update resync function if updateResyncFuncAfterDocsAdded set to true
```
adamcfraser:
```suggestion
// helper function to insert documents equals to docsToCreate, and update sync function if updateResyncFuncAfterDocsAdded set to true
```

db/background_mgr_resync_dcp_test.go
```diff
@@ -0,0 +1,410 @@
...
+func setupTestDBForResyncWithDocs(t *testing.T, docsToCreate int, updateResyncFuncAfterDocsAdded bool) (*Database, context.Context) {
+	db, ctx := setupTestDB(t)
+	db.Options.QueryPaginationLimit = 100
+	syncFn := `
+function sync(doc, oldDoc){
+	channel("channel." + "ABC");
```
adamcfraser:
```suggestion
	channel("channel.ABC");
```

db/background_mgr_resync_dcp_test.go
```diff
@@ -0,0 +1,410 @@
...
+		require.NoError(t, err)
+	}
+
+	assert.Equal(t, docsToCreate, int(db.DbStats.Database().SyncFunctionCount.Value()))
+	db.DbStats.Database().SyncFunctionCount.Set(0)
+
+	if updateResyncFuncAfterDocsAdded {
+		syncFn = `
+function sync(doc, oldDoc){
+	channel("channel." + "ABC123");
```
adamcfraser:
```suggestion
	channel("channel.ABC123");
```

db/database.go
```diff
@@ -712,13 +716,79 @@ func (context *DatabaseContext) Close(ctx context.Context) {
...
+func waitForBackgroundManagersToStop(ctx context.Context, waitTimeMax time.Duration, bgManagers []*BackgroundManager) {
+	timeout := time.NewTicker(waitTimeMax)
+	interval := time.NewTicker(1 * time.Second)
+	defer timeout.Stop()
+	defer interval.Stop()
+
+	for {
+		select {
+		case <-timeout.C:
+			base.WarnfCtx(ctx, "Timeout after %v of waiting for Background Managers to terminate", waitTimeMax)
```
adamcfraser: It would be useful to know which background managers failed to stop.

db/database.go
```diff
@@ -1719,72 +1670,213 @@ func (db *DatabaseCollectionWithUser) UpdateAllDocChannels(ctx context.Context,
...
 		}
-
-		for _, role := range roles {
-			role, err := authr.GetRole(role)
+	})
+	shouldUpdate = changed > 0 || forceUpdate
+	return doc, shouldUpdate, updatedExpiry, doc.Sequence, updatedUnusedSequences, nil
+}
+
+func (db *DatabaseCollectionWithUser) updateDocument(ctx context.Context, docid, key string, regenerateSequences bool, unusedSequences []uint64) (updatedHighSeq uint64, updatedUnusedSequences []uint64, err error) {
```
adamcfraser: I feel like updateDocument is too generic of a name in this context.  resyncDocument would be better.

db/database_test.go
```diff
@@ -2592,6 +2592,249 @@ func TestGetRoleIDs(t *testing.T) {
 	assert.ElementsMatch(t, []string{role1.Name()}, roles)
 }
 
+func Test_updateAllPrincipalsSequences(t *testing.T) {
```
adamcfraser, line 2620: The rest of the repo follows the naming pattern TestUpdateAllPrincipalsSequences (no underscore) - it's preferable to follow that convention for new tests.

adamcfraser (changes requested):

db/background_mgr.go
nirav24: `GetRunState` is used in `stopBackgroundManagers` and `waitForBackgroundManagersToStop` to confirm BackgroundManager has stopped, which prompted to convert GetRunState to non-test. I couldn't think of a way to implement same functionality without changing this method to be non-test. would love to know if you have any other approach in your mind.

db/database_test.go
nirav24: `updateAllPrincipalSequnces` is private function hence I put `_` suffix. There is similar test for `validateJavascriptFunction` at https://github.com/couchbase/sync_gateway/blob/7e43eff49390972eea361c38117810b90efbd335/rest/config_test.go#L2355

let me know if you do want to change the name.

db/background_mgr_resync_dcp_test.go
nirav24: updated to check stats frequently and stop DCP feed after processing certain docs

db/database.go
nirav24: I agree too. However, the problem is that there is no way in current implementation to know what BackgroundManager is (is it AttachmentCompaction or ResyncManager or something else). In order to log which BackgroundManager failed to stop within given time limit, we would have to change `BackgroundManagerProcessI` to add `Name` method so that we can leverage that in logging. 

Another solution could be what I did in https://github.com/couchbase/sync_gateway/pull/5870/commits/49ec67bc0bc7b4e77dc278f89ef96e9b8f6801ab#diff-d1f45819145818652adec732708bb3dc6640a0f966b4287696eeb45ab9c901eaR733 where `stopBackgroundProcess` was taking in `processName` to log this kind of messages (signalled to stop, failed to stop, stopped, etc). 

Personally, I'd prefer to not change `BackgroundManagerProcessI` interface just for logging purpose. Let me know what are your thoughts on this.

db/database_test.go
adamcfraser: Ah ok - that's fine.

db/database.go
adamcfraser: As discussed, let's add a name to BackgroundManager struct and API for this purpose.

nirav24: addressed

db/database.go
adamcfraser: Yes, let's add an unsupported database level config option to run with the old, non-DCP version.

db/background_mgr_resync_dcp.go
```diff
@@ -0,0 +1,297 @@
...
+
+		if err == nil {
+			r.DocsChanged.Add(1)
+		} else if err != base.ErrUpdateCancel {
+			base.WarnfCtx(ctx, "[%s] Error updating doc %q: %v", resyncLoggingID, base.UD(docID), err)
+		}
+		return true
+	}
+
+	collection, err := base.AsGocbV2Bucket(db.Bucket)
```
adamcfraser, line 136: walrus has support for collections and DCP - is there anything about this implementation that prevents compatibility with walrus for simplified testing?

db/background_mgr_resync_dcp.go
```diff
@@ -0,0 +1,297 @@
...
+	VBUUIDs []uint64 `json:"vbuuids"`
+}
+
+type ResyncManagerStatusDocDCP struct {
+	ResyncManagerResponseDCP `json:"status"`
+	ResyncManagerMeta        `json:"meta"`
+}
+
+// getReSyncDCPClientOptions returns the default set of DCPClientOptions suitable for resync
+func getReSyncDCPClientOptions(collectionIDs []uint32, groupID string) (*base.DCPClientOptions, error) {
```
adamcfraser, line 280:
```suggestion
func getResyncDCPClientOptions(collectionIDs []uint32, groupID string) (*base.DCPClientOptions, error) {
```

db/background_mgr_resync_dcp.go
adamcfraser: This is true for any use of NewDCPClient, not just resync. Filed CBG-2661 for the more general enhancement.

adamcfraser (approved):

PR #5871, opened 2022-11-01
Add DatabaseCollection type CBG-2327

torcolvin:
CBG-2327

I'm not attached to naming on this PR. I'd love feedback on the new class names and potential abbreviations for receivers.

- Created `DatabaseCollection` and `DatabaseCollectionWithUser` types. There is `DatabaseContext.CollectionByID` that returns a map of `DatabaseCollections`. In general, I did a 1:1 rename on `DatabaseContext` -> `DatabaseCollection` and `Database` -> `DatabaseCollectionWithUser`.
- I left the receiver names on the methods the same to reduce the code changes. I would change this in a subsequent review.
- I left views code on the `DatabaseContext` object.
- Stub functions are created on `Database` indicating that we have to actually choose a collection in the future: `GetSingleDatabaseCollection`, `GetSingleDatabaseCollectionWithUser`. These would be addressed in a future code review. Some of this isn't really possible until CBG-2266 happens and a `base.Bucket` isn't actually a `base.Collection`.
- Left passthrough from `DatabaseCollection` to `DatabaseContext`. I tried a different approach of passing the options through, but in testing we frequently modify ptrs on `DatabaseContext` objects to have a different values and while this approach would work OK for real code, it breaks a lot of tests. I created new accessors on `DatabaseCollection` to allow refactoring in the future.
- I found having an API with an optional user attr somewhat awkward, and might have chosen to pass them in separately

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1064/
- [x] `GSI=false,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1062/

### DIFF
--- a/base/collection.go
+++ b/base/collection.go
@@ -30,7 +30,7 @@ import (
 var ErrCollectionsUnsupported = errors.New("collections not supported")
 
 // DefaultCollectionID represents _default._default collection
-const DefaultCollectionID = uint32(0x0)
+const DefaultCollectionID = uint32(0)
 
 var _ sgbucket.KVStore = &Collection{}
 var _ CouchbaseStore = &Collection{}
--- a/base/collection.go
+++ b/base/collection.go
@@ -30,7 +30,7 @@ import (
 var ErrCollectionsUnsupported = errors.New("collections not supported")
 
 // DefaultCollectionID represents _default._default collection
-const DefaultCollectionID = 0x0
+const DefaultCollectionID = uint32(0x0)
 
 var _ sgbucket.KVStore = &Collection{}
 var _ CouchbaseStore = &Collection{}
--- a/db/active_replicator_checkpointer.go
+++ b/db/active_replicator_checkpointer.go
@@ -462,9 +462,10 @@ func (c *Checkpointer) _setCheckpoints(seq string, status *ReplicationStatus) (e
 func (c *Checkpointer) getLocalCheckpoint() (checkpoint *replicationCheckpoint, err error) {
 	base.TracefCtx(c.ctx, base.KeyReplicate, "getLocalCheckpoint")
 
-	checkpointBytes, err := c.activeDB.GetSpecialBytes(DocTypeLocal, CheckpointDocIDPrefix+c.clientID)
+	collection := c.activeDB.GetSingleDatabaseCollection()
+	checkpointBytes, err := collection.GetSpecialBytes(DocTypeLocal, CheckpointDocIDPrefix+c.clientID)
 	if err != nil {
-		if !base.IsKeyNotFoundError(c.activeDB.Bucket, err) {
+		if !base.IsKeyNotFoundError(collection.Bucket, err) {
 			return &replicationCheckpoint{}, err
 		}
 		base.DebugfCtx(c.ctx, base.KeyReplicate, "couldn't find existing local checkpoint for client %q", c.clientID)
@@ -476,7 +477,8 @@ func (c *Checkpointer) getLocalCheckpoint() (checkpoint *replicationCheckpoint, 
 }
 
 func (c *Checkpointer) setLocalCheckpoint(checkpoint *replicationCheckpoint) (newRev string, err error) {
-	newRev, err = c.activeDB.putSpecial(DocTypeLocal, CheckpointDocIDPrefix+c.clientID, checkpoint.Rev, checkpoint.AsBody())
+	collection := c.activeDB.GetSingleDatabaseCollection()
+	newRev, err = collection.putSpecial(DocTypeLocal, CheckpointDocIDPrefix+c.clientID, checkpoint.Rev, checkpoint.AsBody())
 	if err != nil {
 		base.TracefCtx(c.ctx, base.KeyReplicate, "Error setting local checkpoint(%v): %v", checkpoint, err)
 		return "", err
@@ -662,9 +664,10 @@ func (c *Checkpointer) setLocalCheckpointStatus(status string, errorMessage stri
 func getLocalCheckpoint(db *DatabaseContext, clientID string) (*replicationCheckpoint, error) {
 	base.TracefCtx(context.TODO(), base.KeyReplicate, "getLocalCheckpoint for %s", clientID)
 
-	checkpointBytes, err := db.GetSpecialBytes(DocTypeLocal, CheckpointDocIDPrefix+clientID)
+	collection := db.GetSingleDatabaseCollection()
+	checkpointBytes, err := collection.GetSpecialBytes(DocTypeLocal, CheckpointDocIDPrefix+clientID)
 	if err != nil {
-		if !base.IsKeyNotFoundError(db.Bucket, err) {
+		if !base.IsKeyNotFoundError(collection.Bucket, err) {
 			return nil, err
 		}
 		base.DebugfCtx(context.TODO(), base.KeyReplicate, "couldn't find existing local checkpoint for ID %q", clientID)
@@ -696,7 +699,8 @@ func setLocalCheckpointStatus(ctx context.Context, db *Database, clientID string
 	checkpoint.Status.Status = status
 	checkpoint.Status.ErrorMessage = errorMessage
 	base.TracefCtx(ctx, base.KeyReplicate, "setLocalCheckpoint(%v)", checkpoint)
-	newRev, putErr := db.putSpecial(DocTypeLocal, CheckpointDocIDPrefix+clientID, checkpoint.Rev, checkpoint.AsBody())
+	collection := db.GetSingleDatabaseCollection()
+	newRev, putErr := collection.putSpecial(DocTypeLocal, CheckpointDocIDPrefix+clientID, checkpoint.Rev, checkpoint.AsBody())
 	if putErr != nil {
 		base.WarnfCtx(ctx, "Unable to persist status in local checkpoint for %s, status not updated: %v", clientID, putErr)
 	} else {
--- a/db/active_replicator_push.go
+++ b/db/active_replicator_push.go
@@ -92,9 +92,8 @@ func (apr *ActivePushReplicator) _connect() error {
 	bh := blipHandler{
 		BlipSyncContext: apr.blipSyncContext,
 		db:              apr.config.ActiveDB,
-		// FIXME: this is probably an array of collections
-		collection:   singleCollection,
-		serialNumber: apr.blipSyncContext.incrementSerialNumber(),
+		collection:      singleCollection,
+		serialNumber:    apr.blipSyncContext.incrementSerialNumber(),
 	}
 
 	seq, err := ParseSequenceID(apr.Checkpointer.lastCheckpointSeq)
--- a/db/active_replicator_push.go
+++ b/db/active_replicator_push.go
@@ -97,7 +97,7 @@ func (apr *ActivePushReplicator) _connect() error {
 		serialNumber: apr.blipSyncContext.incrementSerialNumber(),
 	}
 
-	seq, err := singleCollection.ParseSequenceID(apr.Checkpointer.lastCheckpointSeq)
+	seq, err := ParseSequenceID(apr.Checkpointer.lastCheckpointSeq)
 	if err != nil {
 		base.WarnfCtx(apr.ctx, "couldn't parse checkpointed sequence ID, starting push from seq:0")
 	}
--- a/db/active_replicator_push.go
+++ b/db/active_replicator_push.go
@@ -85,14 +85,19 @@ func (apr *ActivePushReplicator) _connect() error {
 		return err
 	}
 
+	singleCollection := &DatabaseCollectionWithUser{
+		DatabaseCollection: apr.config.ActiveDB.GetSingleDatabaseCollection(),
+		user:               apr.config.ActiveDB.user,
+	}
 	bh := blipHandler{
 		BlipSyncContext: apr.blipSyncContext,
 		db:              apr.config.ActiveDB,
-		collection:      apr.config.ActiveDB,
-		serialNumber:    apr.blipSyncContext.incrementSerialNumber(),
+		// FIXME: this is probably an array of collections
+		collection:   singleCollection,
+		serialNumber: apr.blipSyncContext.incrementSerialNumber(),
 	}
 
-	seq, err := apr.config.ActiveDB.ParseSequenceID(apr.Checkpointer.lastCheckpointSeq)
+	seq, err := singleCollection.ParseSequenceID(apr.Checkpointer.lastCheckpointSeq)
 	if err != nil {
 		base.WarnfCtx(apr.ctx, "couldn't parse checkpointed sequence ID, starting push from seq:0")
 	}
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -56,9 +56,9 @@ const maxInFlightChangesBatches = 2
 
 type blipHandler struct {
 	*BlipSyncContext
-	db            *Database // Handler-specific copy of the BlipSyncContext's blipContextDb
-	collection    *Database // Handler-specific copy of the BlipSyncContext's collection specific DB
-	collectionIdx *int      // index into BlipSyncContext.collectionMapping for the collection
+	db            *Database                   // Handler-specific copy of the BlipSyncContext's blipContextDb
+	collection    *DatabaseCollectionWithUser // Handler-specific copy of the BlipSyncContext's collection specific DB
+	collectionIdx *int                        // index into BlipSyncContext.collectionMapping for the collection
 
 	serialNumber uint64 // This blip handler's serial number to differentiate logs w/ other handlers
 }
@@ -104,6 +104,18 @@ func userBlipHandler(next blipHandlerFunc) blipHandlerFunc {
 	}
 }
 
+func (bh *blipHandler) updateUser() {
+	// refresh the handler's database with the new BlipSyncContext database
+	bh.db = bh._copyContextDatabase()
+	for _, col := range bh.collectionMapping {
+		col = &DatabaseCollectionWithUser{
+			DatabaseCollection: col.DatabaseCollection,
+			user:               bh.db.user,
+		}
+	}
+
+}
+
 func (bh *blipHandler) refreshUser() error {
 
 	bc := bh.BlipSyncContext
@@ -125,7 +137,7 @@ func (bh *blipHandler) refreshUser() error {
 			bc.blipContextDb.SetUser(newUser)
 
 			// refresh the handler's database with the new BlipSyncContext database
-			bh.db = bh._copyContextDatabase()
+			bh.updateUser()
 		}
 	}
 	return nil
@@ -136,7 +148,11 @@ func collectionBlipHandler(next blipHandlerFunc) blipHandlerFunc {
 	return func(bh *blipHandler, bm *blip.Message) error {
 		collectionIndexStr, ok := bm.Properties[BlipCollection]
 		if !ok {
-			bh.collection = bh.db
+			bh.collection = &DatabaseCollectionWithUser{
+				DatabaseCollection: bh.db.GetSingleDatabaseCollection(),
+				user:               bh.db.user,
+			}
+
 			return next(bh, bm)
 		}
 		if len(bh.collectionMapping) == 0 {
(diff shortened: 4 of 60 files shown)


base/collection.go
```diff
@@ -30,7 +30,7 @@ import (
 var ErrCollectionsUnsupported = errors.New("collections not supported")
 
 // DefaultCollectionID represents _default._default collection
-const DefaultCollectionID = 0x0
+const DefaultCollectionID = uint32(0x0)
```
adamcfraser: I think we use uint32(0) elsewhere in the code base.

db/active_replicator_checkpointer.go
```diff
@@ -462,9 +462,10 @@ func (c *Checkpointer) _setCheckpoints(seq string, status *ReplicationStatus) (e
 func (c *Checkpointer) getLocalCheckpoint() (checkpoint *replicationCheckpoint, err error) {
 	base.TracefCtx(c.ctx, base.KeyReplicate, "getLocalCheckpoint")
 
-	checkpointBytes, err := c.activeDB.GetSpecialBytes(DocTypeLocal, CheckpointDocIDPrefix+c.clientID)
+	collection := c.activeDB.GetSingleDatabaseCollection()
```
adamcfraser, line 465: As discussed, for the purposes of review I'm treating all instances of GetSingleDatabaseCollection() as something that will be removed in subsequent phases of the collections implementation.  (mentioning this here as I'm unsure whether these should target a named collection or the metadata collection)

db/active_replicator_push.go
```diff
@@ -85,14 +85,19 @@ func (apr *ActivePushReplicator) _connect() error {
...
 		db:              apr.config.ActiveDB,
-		collection:      apr.config.ActiveDB,
-		serialNumber:    apr.blipSyncContext.incrementSerialNumber(),
+		// FIXME: this is probably an array of collections
+		collection:   singleCollection,
+		serialNumber: apr.blipSyncContext.incrementSerialNumber(),
 	}
 
-	seq, err := apr.config.ActiveDB.ParseSequenceID(apr.Checkpointer.lastCheckpointSeq)
+	seq, err := singleCollection.ParseSequenceID(apr.Checkpointer.lastCheckpointSeq)
```
adamcfraser: I understand not wanting to blow up the scope of this PR, but this is probably a reasonable time to move methods off DatabaseContext/DatabaseCollection if they are clearly no longer using the context (like ParseSequenceID), since we're updating the method signature anyway.

db/blip_handler.go
```diff
@@ -385,8 +401,9 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 
 	// Create a distinct database instance for changes, to avoid races between reloadUser invocation in changes.go
 	// and BlipSyncContext user access.
-	changesDb := bh.copyContextDatabase()
-	_, forceClose := generateBlipSyncChanges(bh.loggingCtx, changesDb, channelSet, options, opts.docIDs, func(changes []*ChangeEntry) error {
+	bh.updateUser()
```
adamcfraser: I don't think we want to update the user on the handler here - the idea is that we're creating a new DB instance unrelated to the handler that we're using for changes.

adamcfraser: If you agree, I suggest moving the contents of updateUser back into refreshUser, since it would be the only usage.

db/blip_handler.go
```diff
@@ -136,7 +148,11 @@ func collectionBlipHandler(next blipHandlerFunc) blipHandlerFunc {
 	return func(bh *blipHandler, bm *blip.Message) error {
 		collectionIndexStr, ok := bm.Properties[BlipCollection]
 		if !ok {
-			bh.collection = bh.db
+			bh.collection = &DatabaseCollectionWithUser{
```
adamcfraser: If this is a collection-scoped handler (collectionBlipHandler), should we always expect a collection property?  Otherwise I'm not sure what specific collection we'd be falling back to here.

adamcfraser (changes requested):
Publishing a partial review, as I've only made it through a subset of files thus far.

db/active_replicator_push.go
torcolvin: I looked at `^func.*\Data.*)` to spot free functions by eye but there's a chance I missed some.

db/active_replicator_checkpointer.go
torcolvin: Agreed, I was going to wait for https://github.com/couchbase/sync_gateway/pull/5832

db/blip_handler.go
torcolvin: We set `bh.collection` on all handlers - and that's what is used by all handlers since most blip handler target a single request. Then previously in a PR I did `bh.db` -> `bh.collection`.

db/active_replicator_push.go
```diff
@@ -85,14 +85,19 @@ func (apr *ActivePushReplicator) _connect() error {
...
+	singleCollection := &DatabaseCollectionWithUser{
+		DatabaseCollection: apr.config.ActiveDB.GetSingleDatabaseCollection(),
+		user:               apr.config.ActiveDB.user,
+	}
 	bh := blipHandler{
 		BlipSyncContext: apr.blipSyncContext,
 		db:              apr.config.ActiveDB,
-		collection:      apr.config.ActiveDB,
-		serialNumber:    apr.blipSyncContext.incrementSerialNumber(),
+		// FIXME: this is probably an array of collections
```
adamcfraser: The blipHandler here is specifically being used to send a subChanges message.  The plan is to send one subChanges per collection, so think the eventual change (with CBG-2491) will be multiple handlers with one collection each.  So if you want a FIXME here, could describe that instead.

db/blip_sync_context.go
```diff
@@ -125,7 +125,7 @@ type BlipSyncContext struct {
 	// when readOnly is true, handleRev requests are rejected
 	readOnly bool
 
-	collectionMapping []*Database // Mapping of array id to collection mapping
+	collectionMapping []*DatabaseCollectionWithUser // Mapping of array id to collection mapping
```
adamcfraser: It looks to me like things would be simplified if this was just []*DatabaseCollection.  collectionBlipHandler could do the work to combine with the user to make a  DatabaseCollectionWithUser for use by a specific blipHandler.

That would simplify the updateUser handling quite a bit, and we can just use the (immutable) DatabaseCollection slice when doing things like looking up indexes or collectionIDs.

db/blip_handler.go
```diff
@@ -104,6 +104,23 @@ func userBlipHandler(next blipHandlerFunc) blipHandlerFunc {
 	}
 }
 
+func (bh *blipHandler) updateUser() {
+	// refresh the handler's database with the new BlipSyncContext database
+	bh.db = bh._copyContextDatabase()
+	for _, col := range bh.collectionMapping {
```
adamcfraser: Commented below in blip_sync_context about turning collectionMapping into []DatabaseCollection to avoid this work.

db/blip_handler.go
adamcfraser: Actually, should we be falling back to the default collection if the collection property is not specified?

db/blip_handler.go
```diff
@@ -400,7 +420,7 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 					}
 
 					// If the user has access to the doc through another channel don't send change
-					userHasAccessToDoc, err := UserHasDocAccess(bh.loggingCtx, bh.collection, change.ID)
+					userHasAccessToDoc, err := UserHasDocAccess(bh.loggingCtx, changesDb.GetSingleDatabaseCollectionWithUser(), change.ID)
```
adamcfraser, line 419: Since CBL is going to be sending a separate subChanges per collection, I think we can do this collection lookup when we create the changesDb above and store as a locally scoped variable?

db/change_cache.go
```diff
@@ -472,7 +472,8 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent) {
 	// If not using xattrs and no sync metadata found, check whether we're mid-upgrade and attempting to read a doc w/ metadata stored in xattr
 	// before ignoring the mutation.
 	if !c.context.UseXattrs() && !syncData.HasValidSyncData() {
-		migratedDoc, _ := c.context.checkForUpgrade(docID, DocUnmarshalNoHistory)
+		dbCollection := c.context.CollectionByID[event.CollectionID]
```
adamcfraser: Shouldn't this also stay as GetSingleDatabaseCollection() for now, so that it's updated consistently with line 373 when multi-collection handling is in place for changeCache?

db/changes.go
```diff
@@ -56,6 +56,7 @@ type ChangeEntry struct {
 	backfill     backfillFlag // Flag used to identify non-client entries used for backfill synchronization (di only)
 	principalDoc bool         // Used to indicate _user/_role docs
 	Revoked      bool         `json:"revoked,omitempty"`
+	CollectionID uint32       `json:"-"`
```
adamcfraser: Can this just be private if it's not intended to be serialized as json?

db/changes.go
```diff
@@ -95,6 +96,11 @@ func (db *Database) addDocToChangeEntry(ctx context.Context, entry *ChangeEntry,
 		return
 	}
 
+	dbCollection, ok := db.CollectionByID[entry.CollectionID]
+	if !ok {
+		base.FatalfCtx(ctx, "Changes feed: could not determine collection from change entry")
```
adamcfraser: Fatalf should be used with care.  It looks like we're using Warnf and returning in cases of unrecoverable errors below, I think this should follow that pattern.

db/changes.go
```diff
@@ -136,11 +142,12 @@ func (db *Database) addDocToChangeEntry(ctx context.Context, entry *ChangeEntry,
 }
 
 func (db *Database) AddDocToChangeEntryUsingRevCache(ctx context.Context, entry *ChangeEntry, revID string) (err error) {
-	rev, err := db.getRev(ctx, entry.ID, revID, 0, nil, RevCacheIncludeBody)
+	dbCollectionWithUser := db.GetSingleDatabaseCollectionWithUser()
```
adamcfraser, line 145: It looks like all the AddDocToChangeEntry variations should get migrated from Database to DatabaseCollectionForUser.  I'm assuming they aren't part of this initial PR because they have consumers that depend on other, pending work - is that accurate?

db/changes_view.go
```diff
@@ -207,7 +212,11 @@ func (dbc *DatabaseContext) getChangesForSequences(ctx context.Context, sequence
 		if usingViews {
 			entry, found = nextChannelViewEntry(queryResults)
 		} else {
-			entry, found = nextChannelQueryEntry(queryResults)
+			collectionID, err := dbc.GetSingleCollectionID()
```
adamcfraser: I know this is temporary code, but I think we should still fetch the collectionID outside the loop instead of for each entry.

db/crud.go
```diff
@@ -1345,7 +1345,7 @@ func (db *Database) tombstoneActiveRevision(ctx context.Context, doc *Document,
 	// Backup previous revision body, then remove the current body from the doc
 	bodyBytes, err := doc.BodyBytes()
 	if err == nil {
-		_ = db.setOldRevisionJSON(ctx, doc.ID, revID, bodyBytes, db.Options.OldRevExpirySeconds)
+		_ = db.setOldRevisionJSON(ctx, doc.ID, revID, bodyBytes, db.oldRevExpirySeconds())
```
adamcfraser, line 1348: Do you know whether these types of accessors get inlined by the compiler?  If we're making an unnecessary alloc/copy of the value every time we're invoking in this way, I'm concerned about the cumulative impact of this approach.

db/database.go
```diff
@@ -82,55 +82,55 @@ const BGTCompletionMaxWait = 30 * time.Second
...
+	State                       uint32                  // The runtime state of the DB from a service perspective
+	ResyncManager               *BackgroundManager
+	TombstoneCompactionManager  *BackgroundManager
+	AttachmentCompactionManager *BackgroundManager
+	ExitChanges                 chan struct{}        // Active _changes feeds on the DB will close when this channel is closed
+	OIDCProviders               auth.OIDCProviderMap // OIDC clients
+	LocalJWTProviders           auth.LocalJWTProviderMap
+	PurgeInterval               time.Duration // Metadata purge interval
+	serverUUID                  string        // UUID of the server, if available
+	revisionCache               RevisionCache // Cache of recently-accessed doc revisions
```
adamcfraser: Some things got moved around here, can you revert to minimize the diffs and clean up unintended whitespace changes?

db/database.go
```diff
@@ -1661,39 +1667,41 @@ func (db *Database) UpdateAllDocChannels(ctx context.Context, regenerateSequence
...
 		for _, name := range roles {
-			db.invalRoleChannels(ctx, name, endSeq)
+			collection.invalRoleChannels(ctx, name, endSeq)
 		}
 	}
 	return docsChanged, nil
 }
 
-func (db *Database) invalUserRoles(ctx context.Context, username string, invalSeq uint64) {
+func (db *DatabaseCollection) invalUserRoles(ctx context.Context, username string, invalSeq uint64) {
```
adamcfraser, line 1814: I guess will specifically be 'invalidate user roles for a given collection', and that's why this moves to DatabaseCollection (even though users and roles are Database-level concepts)?

db/database.go
```diff
@@ -1749,11 +1757,12 @@ func (context *DatabaseContext) UseViews() bool {
 	return context.Options.UseViews
 }
 
+// TOR: This function is duplicated on DatabaseCollection
 func (context *DatabaseContext) DeltaSyncEnabled() bool {
 	return context.Options.DeltaSyncOptions.Enabled
 }
 
-func (context *DatabaseContext) AllowExternalRevBodyStorage() bool {
+func (context *DatabaseCollection) AllowExternalRevBodyStorage() bool {
```
adamcfraser, line 1895: I don't think this should move to DatabaseCollection - UseXattrs is always going to be a Database-level configuration.

db/database.go
```diff
@@ -1749,11 +1757,12 @@ func (context *DatabaseContext) UseViews() bool {
 	return context.Options.UseViews
 }
 
+// TOR: This function is duplicated on DatabaseCollection
```
adamcfraser: It sounds like we want a consistent pattern for how to access DatabaseContext configuration from a DatabaseCollection.  Need to decide whether it's going to be a duplicated helper function on DatabaseCollection, or whether callers will need to do databaseCollection.dbCtx.DeltaSyncEnabled().  The former makes the code less wordy and means less refactoring, but the latter makes the code a self-documenting that these properties aren't collection-specific.  I'm not convinced one way or the other, but we should pick an approach and apply it consistently.

db/database.go
```diff
@@ -1834,10 +1832,11 @@ func initDatabaseStats(dbName string, autoImport bool, options DatabaseContextOp
 }
 
 // For test usage
-func (context *DatabaseContext) GetRevisionCacheForTest() RevisionCache {
+func (context *DatabaseCollection) GetRevisionCacheForTest() RevisionCache {
 	return context.revisionCache
 }
 
+// Duplicated on DatabaseCollection
```
adamcfraser: Can clean up these comments based on the decision discussed above.

db/database_collection.go
```diff
@@ -0,0 +1,167 @@
...
+
+	"github.com/couchbase/sync_gateway/auth"
+	"github.com/couchbase/sync_gateway/base"
+	"github.com/couchbase/sync_gateway/channels"
+)
+
+// DatabaseCollection provides a representation of a single collection of a database.
+type DatabaseCollection struct {
+	Bucket        base.Bucket      // Storage
+	revisionCache RevisionCache    // Cache of recently-accessed doc revisions
```
adamcfraser, line 22: At what point can we remove DatabaseContext.revisionCache?  I'm unsure about adding it here before it's removed there.

db/database_collection.go
```diff
@@ -0,0 +1,167 @@
...
+
+}
+
+// ForceAPIForbiddenErrors returns true if we return 403 vs empty docs. This is controlled at the database level.
+func (c *DatabaseCollection) ForceAPIForbiddenErrors() bool {
+	return c.dbCtx.Options.UnsupportedOptions != nil && c.dbCtx.Options.UnsupportedOptions.ForceAPIForbiddenErrors
+}
+
+// importFilter returns the sync function.
+// FIXME: collections - this needs to be moved to per-collection
```
adamcfraser: For all these FIXME, would be preferable to include a link to the associated CBG ticket (and create if needed).

db/functions/function.go
```diff
@@ -129,7 +129,7 @@ func (fn *functionImpl) AllowByDefault(allow bool) {
 }
 
 // Creates an Invocation of a UserFunction.
-func (fn *functionImpl) Invoke(db *db.Database, args map[string]any, mutationAllowed bool, ctx context.Context) (db.UserFunctionInvocation, error) {
+func (fn *functionImpl) Invoke(database *db.Database, args map[string]any, mutationAllowed bool, ctx context.Context) (db.UserFunctionInvocation, error) {
```
adamcfraser, line 132: This change is fine, but not sure why it's needed.

db/functions/js_callbacks.go
```diff
@@ -91,15 +92,16 @@ func (runner *jsRunner) do_func(funcName string, params map[string]any, sudo boo
...
+	if err := db.CheckTimeout(runner.ctx); err != nil {
 		return nil, err
 	}
 	if sudo {
 		user := runner.currentDB.User()
 		runner.currentDB.SetUser(nil)
 		defer func() { runner.currentDB.SetUser(user) }()
 	}
-	rev, err := runner.currentDB.GetRev(runner.ctx, docID, "", false, nil)
+	collection := runner.currentDB.GetSingleDatabaseCollectionWithUser()
```
adamcfraser: It would be better to define currentCollection on jsRunner, and initialize in jsRunner.CallWithDB, as opposed to doing it every callback.

db/revision.go
```diff
@@ -323,7 +323,7 @@ func (db *Database) refreshPreviousRevisionBackup(ctx context.Context, docid str
 }
 
 // Currently only used by unit tests - deletes an archived old revision from the database
-func (db *Database) PurgeOldRevisionJSON(ctx context.Context, docid string, revid string) error {
+func (db *DatabaseCollection) PurgeOldRevisionJSON(ctx context.Context, docid string, revid string) error {
```
adamcfraser, line 326: I was going to say this should be DatabaseCollectionWithUser, but looking at the usage I think it's fine the way it is.

rest/api_collections_test.go
```diff
@@ -194,6 +194,7 @@ func TestMultiCollectionDCP(t *testing.T) {
 		t.Skip("Test relies on import - needs xattrs")
 	}
 
+	t.Skip("Skip until CBG-2266 is implemented")
```
adamcfraser, line 197: Maybe stick a note on CBG-2266 about this test to ensure this doesn't get missed.

adamcfraser (changes requested):
Overall approach makes sense, a few spots for review/clarification/cleanup.

db/blip_handler.go
torcolvin: Yes, for pre helium clients.

db/database.go
torcolvin: I agree it's a database property, but the only caller is in crud.go where we don't have the database instance. If we don't keep an original pointer to `DatabaseContext.Options`, some options are going to be needed for `crud.go`

db/functions/function.go
torcolvin: `db` variable shadows `db` package.

db/database.go
torcolvin: Fixed.

db/database.go
torcolvin: The reason behind this is that `MarkPrincipalsChanged` is called by crud opts. It's a database level operation only called by functions that are collection scoped, because it's based off non principal documents.

db/database_collection.go
torcolvin: I removed the FIXMEs, becuase I'm not sure where these will actually got modified and they do have associated jira tickets.

db/blip_sync_context.go
torcolvin: Made this change.

db/active_replicator_push.go
torcolvin: This comment was from before we decided this.

db/blip_handler.go
torcolvin: Implemented.

db/changes.go
torcolvin: whoops, fixed.

db/changes.go
torcolvin: This can only be programmer error (forgetting to add collectionID to `ChangeEntry` or `LogEntry`, so I thought Fatalf might be appropriate (I even thought about panicking)

db/crud.go
torcolvin: They get inlined. https://godbolt.org/z/5Mh96ad4Y

db/database_collection.go
torcolvin: Yes, removed.

db/functions/js_callbacks.go
adamcfraser: Agree to leave this for now, needs review of user mutability in js_callbacks.

adamcfraser: I think this should be default collection for the time being, until collection support gets added to the functions/graphQL API.

db/changes.go
torcolvin: Resolved with Warnf

adamcfraser (approved):

PR #5887, opened 2022-11-07
Use require so failing tests don't panic

torcolvin:
From

https://jenkins.sgwdev.com/job/weekly-matrix/BACKING_STORE=enterprise-6.6.5,QUERY_PROVIDER=gsi,SYNC_STORE=xattrs,label=sgw-integration-ec2/21/testReport/junit/github/com_couchbase_sync_gateway_db_indextest/TestRoleQuery_TestRoleQuery_in_Serverless_false/

### DIFF
--- a/db/indextest/indextest_test.go
+++ b/db/indextest/indextest_test.go
@@ -19,7 +19,7 @@ import (
 	"github.com/couchbase/sync_gateway/channels"
 	"github.com/couchbase/sync_gateway/db"
 	"github.com/google/uuid"
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestRoleQuery(t *testing.T) {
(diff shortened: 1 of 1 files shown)


db/indextest/indextest_test.go
```diff
@@ -46,38 +46,40 @@ func TestRoleQuery(t *testing.T) {
...
 			err = authenticator.DeleteRole(role1, false, 0)
-			assert.NoError(t, err)
+			require.NoError(t, err)
 
 			// Standard query
 			results, queryErr := database.QueryRoles(ctx, "", 0)
-			assert.NoError(t, queryErr, "Query error")
+			require.NoError(t, queryErr, "Query error")
+			defer func() {
+				require.NoError(t, results.Close())
```
adamcfraser, line 75: How does a require within a defer behave, with respect to other pending defer functions?  Do they still fire?

db/indextest/indextest_test.go
```diff
@@ -46,38 +46,40 @@ func TestRoleQuery(t *testing.T) {
...
+			}()
 			var row map[string]interface{}
 			rowCount := 0
 			for results.Next(&row) {
 				rowCount++
 			}
-			assert.Equal(t, 4, rowCount)
-			assert.NoError(t, results.Close())
+			require.Equal(t, 4, rowCount)
```
adamcfraser, line 82: This looks like it should still be an assert.

adamcfraser (changes requested):

db/indextest/indextest_test.go
torcolvin: It's now the last line, and if something was added to the test I wouldn't want it to continue. I don't have strong feelings though.

db/indextest/indextest_test.go
torcolvin: This will show two errors if there was require that was hit first:

- require from "normal" test behavior
- require from a defer
- require from a non defer (db.close()) here.

It might be better practice to always put asserts in defers because future code would not execute if you added another line to this function.

adamcfraser (approved):

PR #5894, opened 2022-11-09
CBG-2420: Admin API cannot fix a db with a broken saved config

mohammed-madi:
CBG-2420

Describe your PR here...
- Changed handleDeleteDB to loop over buckets if GetDatabase does not find the requested database context, which is what happens when an invalid DB config is persisted on the bucket.
- A database stuck in the state described in CBG-2420 can now be deleted through the REST API
- Wrote unit test that puts a db in that state and deletes it with a REST API request

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1324/


rest/admin_api.go
```diff
@@ -915,19 +915,44 @@ func (h *handler) handlePutDbConfigImportFilter() error {
...
+			var config map[string]interface{}
+			_, err = h.server.BootstrapContext.Connection.GetConfig(s, h.server.Config.Bootstrap.ConfigGroupID, &config)
+			if err != nil && err != base.ErrNotFound {
+				return err
+			}
+			if config["name"] == dbName {
+				bucket = s
+				break
+			}
+		}
```
bbrks: We have a function called `ServerContext.fetchDatabase(..., dbName)` which does approximately the same thing, with some optimisations and handling for per-bucket credentials.

It might be worth seeing if we're able to re-use that via `h.server`

rest/persistent_config_test.go
```diff
@@ -161,6 +161,55 @@ func TestAutomaticConfigUpgradeError(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	resp = rt.SendAdminRequest(http.MethodGet, "/newdb/", "{}")
+	RequireStatus(t, resp, http.StatusNotFound)
+	resp = rt.SendAdminRequest(http.MethodDelete, "/newdb/", fmt.Sprintf(
+		`{"bucket": "%s", "num_index_replicas": 0, "enable_shared_bucket_access": %t, "use_views": %t}`,
+		tb.GetName(), base.TestUseXattrs(), base.TestsDisableGSI(),
+	))
+	RequireStatus(t, resp, http.StatusOK)
+	_, _, err = tb.GetRaw("_sync:dbconfig:default")
```
bbrks: We're not doing anything with this `err`. I assume you wanted to check that you got an error back when as the dbconfig was removed?

rest/persistent_config_test.go
```diff
@@ -161,6 +161,55 @@ func TestAutomaticConfigUpgradeError(t *testing.T) {
...
+	require.NoError(t, err)
+	_, err = rt.ServerContext().fetchAndLoadConfigs(rt.Context(), false)
+	assert.NoError(t, err)
+
+	resp = rt.SendAdminRequest(http.MethodGet, "/newdb/", "{}")
+	RequireStatus(t, resp, http.StatusNotFound)
+	resp = rt.SendAdminRequest(http.MethodDelete, "/newdb/", fmt.Sprintf(
+		`{"bucket": "%s", "num_index_replicas": 0, "enable_shared_bucket_access": %t, "use_views": %t}`,
+		tb.GetName(), base.TestUseXattrs(), base.TestsDisableGSI(),
+	))
```
bbrks: Delete doesn't take a body

```suggestion
	resp = rt.SendAdminRequest(http.MethodDelete, "/newdb/", "")
```

rest/persistent_config_test.go
```diff
@@ -161,6 +161,55 @@ func TestAutomaticConfigUpgradeError(t *testing.T) {
...
+	_, err = rt.ServerContext().BootstrapContext.Connection.UpdateConfig(tb.GetName(), rt.ServerContext().Config.Bootstrap.ConfigGroupID,
+		func(rawBucketConfig []byte, rawBucketConfigCas uint64) (newConfig []byte, err error) {
+			bytes, err := json.Marshal(cnf)
+			return bytes, err
+		})
+	require.NoError(t, err)
+	_, err = rt.ServerContext().fetchAndLoadConfigs(rt.Context(), false)
+	assert.NoError(t, err)
+
+	resp = rt.SendAdminRequest(http.MethodGet, "/newdb/", "{}")
```
bbrks: Get doesn't take a body

```suggestion
	resp = rt.SendAdminRequest(http.MethodGet, "/newdb/", "")
```

rest/persistent_config_test.go
```diff
@@ -161,6 +161,55 @@ func TestAutomaticConfigUpgradeError(t *testing.T) {
...
+	_, err := rt.ServerContext().BootstrapContext.Connection.GetConfig(tb.GetName(), rt.ServerContext().Config.Bootstrap.ConfigGroupID, &cnf)
+	require.NoError(t, err)
+
+	// Add invalid json fields to the config
+	cnf["num_index_replicas"] = "0"
+	marshaled, err := json.Marshal(cnf)
+
+	// Set the config to _sync:dbconfig:default
+	err = tb.SetRaw("_sync:dbconfig:default", 0, nil, marshaled)
+	require.NoError(t, err)
```
bbrks: I'm not 100% sure, but this SetRaw step might not be needed if your `UpdateConfig` callback is returning the broken marshalled version below to do the update.

bbrks (changes requested):

rest/persistent_config_test.go
mohammed-madi: Was used for debugging but i somehow missed it

mohammed-madi:
Comments on test addressed in https://github.com/couchbase/sync_gateway/pull/5894/commits/544f693fbfd045091cae3a341ae7346b6817565a

rest/admin_api.go
bbrks: Do you have any details on why we can't use fetchDatabase here? We will need to re-implement the same logic for per-bucket credentials at the very least, if we're unable to reuse.

mohammed-madi: FetchDatabase reused in commit https://github.com/couchbase/sync_gateway/pull/5894/commits/251ccc647cf1287e5c150048b2eed9e5efb03f75

rest/admin_api.go
```diff
@@ -915,19 +915,37 @@ func (h *handler) handlePutDbConfigImportFilter() error {
...
+	dbContext, _ := h.server.GetDatabase(h.ctx(), dbName)
+
+	if dbContext != nil {
+		bucket = dbContext.Bucket.GetName()
+	} else if h.server.persistentConfig {
+		_, cnf, err := h.server.fetchDatabase(h.ctx(), dbName)
+		if err != nil {
+			return err
+		}
+		bucket = *cnf.Bucket
```
bbrks: Maybe do a defensive nil check here on cnf and cnf.Bucket to avoid panicing if we haven't got either.

rest/config.go
```diff
@@ -1399,6 +1400,11 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
 			continue
 		}
 		if err != nil {
+			if err.(*json.UnmarshalTypeError) != nil {
+				base.WarnfCtx(ctx, "unable to fetch config in group %q from bucket %q: %v", sc.Config.Bootstrap.ConfigGroupID, bucket, err)
+				cnf.Bucket = &bucket
+				return true, &cnf, nil
+			}
```
bbrks: This seems overly specific to one particular type of broken config error, and only handling one json library in particular. EE uses a non-stdlib json package, which returns different error types.

I'm thinking we'll have to refactor to split this function into two, so you're able to use the pieces you need from it, without having to unmarshal into a full DatabaseConfig struct.

My initial thoughts are to write a function that does the existing loop logic and invoke a callback.
The callback has the ability to either break the loop or continue and is given `bucket` as a parameter, so it's still able to call stuff like `GetConfig(bucket, ...)`

Then two callbacks:
- One that uses the existing inner-loop code which is used by fetchDatabase
- Another specific to your use case: finding the bucket name that matches a given db name, if one exists...

For this we don't need to unmarshal into a full DatabaseConfig. We can unmarshal into an inline struct to only get what we need (and ignore any of the other invalid properties we may have:

```go
func (sc *ServerContext) foreachDbConfig(ctx, callback func(bucket string) (exit bool)) {
	// existing for each bucket/looping code and callback
}

// functionally same as before
func (sc *ServerContext) fetchDatabase(ctx, dbName) {
	// sc.foreachDbConfig ...
	// callback implementing existing logic
}


func (sc *ServerContext) bucketNameFromDbName(dbName string) (bucketName string, found bool) {
	// sc.foreachDbConfig ...
	// callback implementing your new logic
		var cfgDbName struct {
				Name string `json:"name"`
		}
		cas, err := sc.BootstrapContext.Connection.GetConfig(bucket, sc.Config.Bootstrap.ConfigGroupID, &cfgDbName)
		...
		if dbName == cfgDbName.Name {
			return bucket
		}
}
```

bbrks (changes requested):
Not sure I agree with the implementation using `fetchDatabase` - but I left a detailed comment suggesting how to refactor to support both use-cases.

rest/config.go
mohammed-madi: [d13063f](https://github.com/couchbase/sync_gateway/pull/5894/commits/d13063f8f0e9cb90de6f167b6ead9199b55c4d5b) addresses this comment, let me know if further changes are required

rest/config.go
```diff
@@ -1430,10 +1438,38 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
...
 		}
 		base.TracefCtx(ctx, base.KeyConfig, "Got config for bucket %q with cas %d", bucket, cas)
-		return true, &cnf, nil
+		return true, nil
 	}
 
-	return false, nil, nil
+	err = sc.foreachDbConfig(callback)
+	if err != nil {
+		return false, nil, nil
```
bbrks: Should we be returning the error here?

```suggestion
		return false, nil, err
```

rest/config.go
```diff
@@ -1430,10 +1438,38 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
...
+	}
+
+	return true, &cnf, nil
+}
+
+func (sc *ServerContext) bucketNameFromDbName(dbName string) (bucketName string, found bool) {
+	// Only used in handleDeleteDB in config.go when GetDatabase does not find the db (CBG-2420)
```
bbrks: I usually find these types of comments get quickly out of date unless there's a very specific reason why this function should ONLY be used in one particular codepath. Probably best to just remove it. We can also trace it back to this PR with `git blame`/annotations

It might be worth adding a comment though, to explain why we're not using the `DatabaseConfig` struct and instead are creating a smaller one.

```suggestion
	// Minimal representation of config struct to be tolerant of invalid database configurations where we still need to find a database name
```

rest/admin_api.go
```diff
@@ -915,19 +915,33 @@ func (h *handler) handlePutDbConfigImportFilter() error {
...
 		})
 		if err != nil {
-			return base.HTTPErrorf(http.StatusInternalServerError, "couldn't remove database %q from bucket %q: %s", base.MD(h.db.Name), base.MD(bucket), err.Error())
+			return base.HTTPErrorf(http.StatusInternalServerError, "couldn't remove database %q from bucket %q: %s", base.MD(dbName), base.MD(bucket), err.Error())
 		}
 	}
-
-	if !h.server.RemoveDatabase(h.ctx(), h.db.Name) {
-		return base.HTTPErrorf(http.StatusNotFound, "missing")
+	if !h.server.RemoveDatabase(h.ctx(), dbName) && bucket == "" {
```
bbrks: If `bucket == ""` we shouldn't reach this line of code, as we exited on line 932, unless I'm misreading?

rest/admin_api.go
```diff
@@ -915,19 +915,33 @@ func (h *handler) handlePutDbConfigImportFilter() error {
...
+
+	var bucket string
+
+	dbContext, _ := h.server.GetDatabase(h.ctx(), dbName)
+
+	if dbContext != nil {
+		bucket = dbContext.Bucket.GetName()
+	} else if h.server.persistentConfig {
+		bucket, _ = h.server.bucketNameFromDbName(dbName)
+		fmt.Println("BUCKET", bucket)
```
IsaacLambat: Left in print line

rest/config.go
```diff
@@ -1370,7 +1370,8 @@ func (sc *ServerContext) _fetchAndLoadDatabase(ctx context.Context, dbName strin
 	return true, nil
 }
 
-func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (found bool, dbConfig *DatabaseConfig, err error) {
+func (sc *ServerContext) foreachDbConfig(callback func(bucket string) (exit bool, err error)) (err error) {
```
IsaacLambat: Nitpick: function name not camel case. Should be `forEachDbConfig`

IsaacLambat (dismissed):

bbrks (changes requested):
Looks good. I left a few comments on specific things

rest/admin_api.go
mohammed-madi: Ive removed the previous `bucket == "" `check, it was previously failing testHandleDeleteDB, will see how it behaves on jenkins

mohammed-madi: jenkins build failed on TestServerlessBucketCredentialsFetchDatabases

mohammed-madi: `	// Limit SG to buckets defined on BucketCredentials map
	rt.ReplacePerBucketCredentials(map[string]*base.CredentialsConfig{})
	// Make sure fetch fails as it cannot see all buckets in cluster
	found, _, err = rt.ServerContext().fetchDatabase(ctx, "db")
	assert.NoError(t, err)
	assert.False(t, found)`

Doesnt make sense to not expect an error if fetch is expected to not find the db? Ill change it to `	assert.Error(t, err)`

rest/config.go
```diff
@@ -1380,27 +1381,34 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
...
 		}
 	}
 
-	// move bucket matching dbName to the front so it's searched first
-	for i, bucket := range buckets {
-		if bucket == dbName {
-			buckets = append(buckets[i:], buckets[:i]...)
+	for _, bucket := range buckets {
+		exit, err := callback(bucket)
+		if err != nil && err != base.ErrNotFound {
```
bbrks: I don't think it's nessesary to special case this error - the callbacks should expect the loop to stop for any error or any `true` exit return value - and the NotFound error is already filtered out in the callback as a non-critical error

rest/config.go
```diff
@@ -1380,27 +1381,34 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
...
+			return err
+		}
+		if exit {
+			return nil
 		}
 	}
+	return base.ErrNotFound
```
bbrks, line 1375: This function now returns a not found error when it used to return nil if the loop runs to completion without an exit or error from a callback.

I think it's expected that iterating over 0 or many buckets without the callback erroring or exiting is an expected scenario and shouldn't be returning an error.

bbrks (changes requested):
This looks mostly good - I have some questions around `ErrNotFound` handling in the foreach loop as I don't think it quite makes sense to error in this way if the loop completed

rest/config.go
mohammed-madi: addressed in https://github.com/couchbase/sync_gateway/pull/5894/commits/eabaadd62b38564ece35fc81aa4538a3167cd0b6

rest/config.go
mohammed-madi: addressed in https://github.com/couchbase/sync_gateway/pull/5894/commits/b47150822734e0747e4fcad18a548f6ffb3edf4e

rest/admin_api.go
```diff
@@ -915,19 +915,32 @@ func (h *handler) handlePutDbConfigImportFilter() error {
...
+
+	var bucket string
+
+	dbContext, _ := h.server.GetDatabase(h.ctx(), dbName)
+
+	if dbContext != nil {
+		bucket = dbContext.Bucket.GetName()
+	} else if h.server.persistentConfig {
+		_, cnf, err := h.server.fetchDatabase(h.ctx(), dbName)
+		bucket = *cnf.Bucket
```
bbrks: If we got an error `cnf` will probably be nil and this will panic.

I think it's just a case of moving error handling up to avoid?

rest/config.go
```diff
@@ -1398,7 +1399,7 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
 			base.DebugfCtx(ctx, base.KeyConfig, "%q did not contain config in group %q", bucket, sc.Config.Bootstrap.ConfigGroupID)
 			continue
 		}
-		if err != nil {
+		if err != nil && err.(*json.UnmarshalTypeError) == nil {
```
bbrks: If error is not a json error this panics.

```
panic: interface conversion: error is *errors.errorString, not *json.UnmarshalTypeError
```

bbrks: I'm not sure what this is trying to do - check for non-JSON errors? Do we need to think about the EE jsoniter library returning other error types?

rest/config.go
```diff
@@ -1429,6 +1430,13 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
 			cnf.CertPath = sc.Config.Bootstrap.X509CertPath
 			cnf.KeyPath = sc.Config.Bootstrap.X509KeyPath
 		}
+		if err != nil {
+			if err.(*json.UnmarshalTypeError) != nil {
```
bbrks: same comment applies here as above

bbrks (changes requested):

rest/admin_api.go
mohammed-madi: Addressed in https://github.com/couchbase/sync_gateway/pull/5894/commits/e73f4d4d4f01b3a447bab6ebc4695715fd207418

rest/config.go
bbrks: This comment is still applicable. Returning an error after iterating over zero buckets (or iterating without a callback exiting) seems unexpected.

It successfully iterated, and wasn't looking for anything in particular to return not found for.

rest/admin_api.go
```diff
@@ -915,19 +915,28 @@ func (h *handler) handlePutDbConfigImportFilter() error {
...
+
+	var bucket string
+
+	dbContext, _ := h.server.GetDatabase(h.ctx(), dbName)
+
+	if dbContext != nil {
+		bucket = dbContext.Bucket.GetName()
+	} else if h.server.persistentConfig {
+		bucket, _ = h.server.bucketNameFromDbName(dbName)
+	}
```
bbrks: I'm not sure why we have to get bucket name for non-persistent config codepath (and do the `bucket == ""` check on line 938)? Is it nessesary?

bbrks (changes requested):
I'm OK with where this code at, with the exception of two places:

- I'm not asking for changes for this one, but I want to understand why bucket name retrival for non-persistent config codepath is required along with the `bucket == ""` check at the bottom of `handleDeleteDB`
- The `ErrNotFound` at the end of `forEachDbConfig`. If the loop finishes without any callbacks exiting the loop, I'm not sure why this function needs to return an error? The function did the job of iterating successfully.

rest/admin_api.go
bbrks: And to be specific, I was thinking of something like this:

- `handleDeleteDB` treats the persistent config path as a special case inside the existing if block, rather than introducing extra variables (like `bucket`/`dbContext`) that aren't set for some code-paths.
- `bucketNameFromDbName` can avoid a bunch of work looking up db config in buckets by first looking to see if we have a matching database loaded - which is basically what `handleDeleteDB` was doing anyway

```diff
diff --git a/rest/admin_api.go b/rest/admin_api.go
index ea00a21a..9d6f0411 100644
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -962,27 +962,28 @@ func (h *handler) handleDeleteDB() error {
        h.assertAdminOnly()
        dbName := h.PathVar("olddb")

-       var bucket string
-
-       dbContext, _ := h.server.GetDatabase(h.ctx(), dbName)
-
-       if dbContext != nil {
-               bucket = dbContext.Bucket.GetName()
-       } else if h.server.persistentConfig {
-               bucket, _ = h.server.bucketNameFromDbName(dbName)
-       }
-
        if h.server.persistentConfig {
-               _, err := h.server.BootstrapContext.Connection.UpdateConfig(bucket, h.server.Config.Bootstrap.ConfigGroupID, func(rawBucketConfig []byte, rawBucketConfigCas uint64) (updatedConfig []byte, err error) {
-                       return nil, nil
-               })
-               if err != nil {
+               bucket, ok := h.server.bucketNameFromDbName(dbName)
+               if !ok {
+                       return base.HTTPErrorf(http.StatusNotFound, "no such database %q", dbName)
+               }
+
+               if _, err := h.server.BootstrapContext.Connection.UpdateConfig(bucket, h.server.Config.Bootstrap.ConfigGroupID,
+                       func(rawBucketConfig []byte, rawBucketConfigCas uint64) (updatedConfig []byte, err error) {
+                               return nil, nil
+                       },
+               ); err != nil {
                        return base.HTTPErrorf(http.StatusInternalServerError, "couldn't remove database %q from bucket %q: %s", base.MD(dbName), base.MD(bucket), err.Error())
                }
+
+               // we don't care whether the database is actually loaded or not when deleted - we report the delete either way...
+               _ = h.server.RemoveDatabase(h.ctx(), dbName)
+       } else {
+               if !h.server.RemoveDatabase(h.ctx(), dbName) {
+                       return base.HTTPErrorf(http.StatusNotFound, "qwerty no such database %q", dbName)
+               }
        }
-       if !h.server.RemoveDatabase(h.ctx(), dbName) && bucket == "" {
-               return base.HTTPErrorf(http.StatusNotFound, "no such database %q", dbName)
-       }
+
        _, _ = h.response.Write([]byte("{}"))
        return nil
 }
diff --git a/rest/config.go b/rest/config.go
index 7324f8a5..7f1dd52e 100644
--- a/rest/config.go
+++ b/rest/config.go
@@ -1428,6 +1428,14 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
 }

 func (sc *ServerContext) bucketNameFromDbName(dbName string) (bucketName string, found bool) {
+       // see if we find the database in-memory first, otherwise fall back to scanning buckets for db configs
+       sc.lock.RLock()
+       dbc, ok := sc.databases_[dbName]
+       sc.lock.RUnlock()
+       if ok {
+               return dbc.Bucket.GetName(), true
+       }
+
        // Minimal representation of config struct to be tolerant of invalid database configurations where we still need to find a database name
        var cfgDbName struct {
                Name string `json:"name"`
```

rest/config.go
bbrks: Changing this to `nil` does indeed break tests. But we should get to the root of why, and not just make this return the right error to make them work. We can revisit this tomorow if you need help.

rest/admin_api.go
mohammed-madi: addressed in [f05a008](https://github.com/couchbase/sync_gateway/pull/5894/commits/f05a00887713e080a574e9f35fd52cf4b2e19014)

bbrks (approved):

PR #5896, opened 2022-11-10
CBG-2312: Fix for race on resync status found by test failiure

gregns1:
CBG-2312

A race has been found where getting the status of the resync. https://github.com/couchbase/sync_gateway/blob/9d551b5346bf034021d05a76297e9d61946028fa/db/background_mgr.go#L278 <- Inside this function we get the status doc at the top and in this test we have a case where sometimes the status is changing before we get the heartbeat doc. For example, we get the status 'running' and it completes in the time we get to get heartbeat doc and the heart beat doc is no longer there as it has completed. Then it sets the status to stopped causing the test to fail. In this PR I have moved the get status doc and the get heartbeat doc closer together to lessen the chance of this race happening. I have filed a ticket for a more permanent fix: https://issues.couchbase.com/browse/CBG-2535

Also added a wait for pending changes call as I was getting some failure cases when running against server where docs weren't ready. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1316/

### DIFF
--- a/db/background_mgr.go
+++ b/db/background_mgr.go
@@ -289,7 +289,6 @@ func (b *BackgroundManager) getStatusFromCluster() ([]byte, error) {
 		}
 		return nil, err
 	}
-	_, _, heartbeatErr := b.clusterAwareOptions.bucket.GetRaw(b.clusterAwareOptions.HeartbeatDocID())
 
 	var clusterStatus map[string]interface{}
 	err = base.JSONUnmarshal(status, &clusterStatus)
--- a/db/background_mgr.go
+++ b/db/background_mgr.go
@@ -471,7 +471,6 @@ func (b *BackgroundManager) UpdateHeartbeatDocClusterAware() error {
 		// goroutine which intermittently runs this but this snuck in before it was stopped. This may result in the doc
 		// being deleted before this runs. We can ignore that error is that is the case.
 		if base.IsDocNotFoundError(err) && b.terminator.IsClosed() {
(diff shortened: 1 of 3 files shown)


adamcfraser (changes requested):
I think the permanent fix (for CBG-2535) should be relatively straightforward (see notes on ticket), so I'd prefer we went ahead with that change instead of moving up the heartbeat retrieval (particularly since we don't need attempt to retrieve the heartbeat if the replication is already stopped/completed).

gregns1:
Updated and tested against server and seems to work fine. This solution doesn't work against walrus though. The update status operation even when forcing a mismatch in CAS value seems to not return an error. Looking into the Walrus code I think I have found where the issue is but not sure if that is an actual bug in walrus or if its intentional? https://github.com/couchbaselabs/walrus/blob/6f7d5a15211637ae0c59011c949dfdebc6a53930/crud.go#L422 This function is called to write the status when running against Walrus. The function doesn't seem to use the CAS value that is passed into it and instead performs a get on the doc and uses the CAS returned by that to write to the doc with. In this context we need to use the CAS passed into the function to trigger the error on the write to perform the get on the status again when we have mismatched CAS values due to the status being updated during execution of getStatusFromCluster(). 

I'm not sure if to mark the test as not available when running with Walrus or if this is a bug in Walrus code should I put a PR up for Walrus to fix it?

adamcfraser (approved):

PR #5901, opened 2022-11-14
CBG-2550 Principal Collection Access

adamcfraser:
CBG-2550

Principal Collection Access

Adds storage for non-default collection channel access to principals. Access for the default collection uses the existing root-level storage for backward compatibility, named collections are stored in nested map by scope, collection.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1099/

### DIFF
--- a/auth/collection_access_test.go
+++ b/auth/collection_access_test.go
@@ -1,3 +1,11 @@
+// Copyright 2022-Present Couchbase, Inc.
+//
+// Use of this software is governed by the Business Source License included
+// in the file licenses/BSL-Couchbase.txt.  As of the Change Date specified
+// in that file, in accordance with the Business Source License, use of this
+// software will be governed by the Apache License, Version 2.0, included in
+// the file licenses/APL2.txt.
+
 package auth
 
 import (
--- a/auth/collection_access_test.go
+++ b/auth/collection_access_test.go
@@ -2,12 +2,13 @@ package auth
 
 import (
 	"bytes"
+	"log"
+	"testing"
+
 	"github.com/couchbase/sync_gateway/base"
 	ch "github.com/couchbase/sync_gateway/channels"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"log"
-	"testing"
 )
 
 func canSeeAllCollectionChannels(scope, collection string, princ Principal, channels base.Set) bool {
--- a/auth/collection_access_test.go
+++ b/auth/collection_access_test.go
@@ -1,0 +1,192 @@
+package auth
+
+import (
+	"bytes"
+	"github.com/couchbase/sync_gateway/base"
+	ch "github.com/couchbase/sync_gateway/channels"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"log"
+	"testing"
+)
+
+func canSeeAllCollectionChannels(scope, collection string, princ Principal, channels base.Set) bool {
+	for channel := range channels {
+		if !princ.CanSeeCollectionChannel(scope, collection, channel) {
+			return false
+		}
+	}
+	return true
+}
+
+func TestUserCollectionAccess(t *testing.T) {
+
+	// User with no access:
+	bucket := base.GetTestBucket(t)
+	defer bucket.Close()
+	auth := NewAuthenticator(bucket, nil, DefaultAuthenticatorOptions())
+	user, _ := auth.NewUser("foo", "password", nil)
+	scope := "scope1"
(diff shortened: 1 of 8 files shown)


auth/role_collection_access.go
```diff
@@ -0,0 +1,225 @@
...
+				}
+				forbidden = append(forbidden, channel)
+			}
+		}
+		if forbidden != nil {
+			return role.UnauthError(fmt.Sprintf("You are not allowed to see channels %v", forbidden))
+		}
+		return nil
+	}
+	return role.UnauthError(fmt.Sprintf("Unauthorized to see channels %v", channels))
```
torcolvin, line 204: Do you want this error message or say no such scope and collection? is that leaking information?

auth/collection_access_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"log"
+	"testing"
+
+	"github.com/couchbase/sync_gateway/base"
+	ch "github.com/couchbase/sync_gateway/channels"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func canSeeAllCollectionChannels(scope, collection string, princ Principal, channels base.Set) bool {
```
torcolvin, line 22: nice to have: pass t in here, and assert on type, so you know which channel you can't see is

auth/role.go
```diff
@@ -25,13 +25,14 @@ import (
...
-	Name_             string          `json:"name,omitempty"`
-	ExplicitChannels_ ch.TimedSet     `json:"admin_channels,omitempty"`
-	Channels_         ch.TimedSet     `json:"all_channels"`
-	Sequence_         uint64          `json:"sequence"`
-	ChannelHistory_   TimedSetHistory `json:"channel_history,omitempty"`   // Added to when a previously granted channel is revoked. Calculated inside of rebuildChannels.
-	ChannelInvalSeq   uint64          `json:"channel_inval_seq,omitempty"` // Sequence at which the channels were invalidated. Data remains in Channels_ for history calculation.
-	Deleted           bool            `json:"deleted,omitempty"`
+	Name_             string                                  `json:"name,omitempty"`
+	ExplicitChannels_ ch.TimedSet                             `json:"admin_channels,omitempty"`
+	Channels_         ch.TimedSet                             `json:"all_channels"`
```
torcolvin, line 30: Should this become omit-empty? What do we do if there are only non default collections enabled?

auth/role_collection_access.go
```diff
@@ -0,0 +1,225 @@
...
+			for channel := range channels {
+				if ca.CanSeeChannel(channel) {
+					return nil
+				}
+			}
+		} else if ca.Channels().Contains(ch.UserStarChannel) {
+			return nil
+		}
+	}
+	return role.UnauthError("You are not allowed to see this")
```
torcolvin, line 224: Do you want to share the scope and collection here, or not?

torcolvin (commented):
I think looks fine.

Some things I wondered about:

- Does it make sense to use kv CollectionID in some places? Obviously when it gets deserialized it will be named, but when you actually call we work off CollectionIDs, e.g. `FilterToAvailableChannels`. I think the question is where does it make sense to do the conversion? 
- Are there cases where we'll ask for all channels, and in those cases are we going to get multi-collection answers in the case where `foo.bar` and `_default._default` are configured?
- Do certain fields need to become `omitempty` when there are only non default collections specified?

auth/role.go
adamcfraser: I left this as-is for now, because of the way we set the value to nil for channel invalidation.  The full implementation of invalidation for scopes and collections is still pending, could revisit at that time.

auth/role_collection_access.go
adamcfraser: I didn't think it was necessary as the caller would have that information, and this code only has test usage for the time being.

adamcfraser:
To the specific questions:
1. Does it make sense to use kv CollectionID in some places? 
I intentionally avoided using CollectionID in the auth package for data that's persisted to user/role documents - I think that should be our general approach for persisted data.  
2. Are there cases where we'll ask for all channels, and in those cases are we going to get multi-collection answers in the case where foo.bar and _default._default are configured?
There shouldn't be cases where we're asking for all channels (cross-collection) - I think we'll always be doing access requests in the context of a given document or collection.  
3. Do certain fields need to become omitempty when there are only non default collections specified?
Possibly - but as noted I think it makes sense to review that in the context of CBG-2554 (invalidation)

torcolvin (approved):

PR #5911, opened 2022-11-15
CBG-2547: Tuning the KV pool size for serverless mode

gregns1:
CBG-2547

Tuning the KV pool size for when running sync gateway in serverless mode. Added a serverless boolean to the bucket spec struct to be able to access this information. Populated this boolean from the flag set in startup config. Also added some test cases to an existing test to ensure the connection string when running in serverless mode is correct. If the KV pool size is specified in the connection string it will not be overridden by the default. The KV pool size for the cbgt is set here https://github.com/couchbase/sync_gateway/blob/21b449fd31bdb706aef28284239cdd47cf413977/base/dcp_sharded.go#L303 and is already 1 by default so haven't changed this logic. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1101/

### DIFF
--- a/base/bucket.go
+++ b/base/bucket.go
@@ -100,6 +100,7 @@ type CouchbaseBucketType int
 
 // Full specification of how to connect to a bucket
 type BucketSpec struct {
+	Serverless                    bool
 	Server, BucketName, FeedType  string
 	Auth                          AuthHandler
 	Certpath, Keypath, CACertPath string         // X.509 auth parameters
@@ -145,8 +146,10 @@ type GoCBConnStringParams struct {
 }
 
 // FillDefaults replaces any unset fields in this GoCBConnStringParams with their default values.
-func (p *GoCBConnStringParams) FillDefaults() {
-	if p.KVPoolSize == 0 {
+func (p *GoCBConnStringParams) FillDefaults(serverless bool) {
+	if p.KVPoolSize == 0 && serverless {
+		p.KVPoolSize = DefaultGocbKvPoolSizeServerless
+	} else if p.KVPoolSize == 0 {
 		p.KVPoolSize = DefaultGocbKvPoolSize
 	}
 }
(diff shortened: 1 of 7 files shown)


rest/config.go
```diff
@@ -132,6 +133,7 @@ func (bucketConfig *BucketConfig) GetCredentials() (username string, password st
 // DbConfig defines a database configuration used in a config file or the REST API.
 type DbConfig struct {
 	BucketConfig
+	StartupConfig
```
bbrks: This changes the API/storage representation of a database config to also include all of the startup config options - which is not something we want to do.

base/bucket.go
```diff
@@ -100,6 +100,7 @@ type CouchbaseBucketType int
 
 // Full specification of how to connect to a bucket
 type BucketSpec struct {
+	Serverless                    bool
```
bbrks: I don't think the serverless flag should be inside `BucketSpec` - We're adding SG logic into here by doing this.
We want to keep `BucketSpec` purely about defining how to connect to a bucket.

base/bucket.go
```diff
@@ -156,7 +159,7 @@ func (spec *BucketSpec) GetGoCBConnString(params *GoCBConnStringParams) (string,
 	if params == nil {
 		params = &GoCBConnStringParams{}
 	}
-	params.FillDefaults()
+	params.FillDefaults(spec.Serverless)
```
bbrks: It looks like `params.KVPoolSize` is used as the default/fallback when a connection string doesn't explicitly have one (lines 176-179 below) - so maybe we just set that value based on whether we're running in serverless or not further up in the stack, to avoid adding serverless flags this far down (in non-SG-related code)

bbrks (changes requested):
I'm not a fan of passing the serverless flag all the way down into BucketSpec to achieve this. We might need to think about how we can set a kvpoolsize value earlier on in the code based on serverless without it trickling down this far.

bbrks (dismissed):
LGTM

bbrks (approved):

PR #5913, opened 2022-11-16
CBG-2506: Tuning import partitions for when in serverless mode

gregns1:
CBG-2506

Small PR to tune the import partitions for when running sync gateway in serverless mode. Also added a unit test to test the functionality with one case proving the default is set in serverless mode and another proving the value is not overwritten when import partitions are specified in the config. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1102/

### DIFF
--- a/rest/server_context.go
+++ b/rest/server_context.go
@@ -655,10 +655,8 @@ func dbcOptionsFromConfig(ctx context.Context, sc *ServerContext, config *DbConf
 	}
 	importOptions.BackupOldRev = base.BoolDefault(config.ImportBackupOldRev, false)
 
-	if config.ImportPartitions == nil && sc.Config.IsServerless() {
-		importOptions.ImportPartitions = base.DefaultImportPartitionsServerless
-	} else if config.ImportPartitions == nil {
-		importOptions.ImportPartitions = base.DefaultImportPartitions
+	if config.ImportPartitions == nil {
+		importOptions.ImportPartitions = base.GetDefaultImportPartitions(sc.Config.IsServerless())
 	} else {
 		importOptions.ImportPartitions = *config.ImportPartitions
 	}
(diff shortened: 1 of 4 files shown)


rest/server_context.go
```diff
@@ -655,7 +655,9 @@ func dbcOptionsFromConfig(ctx context.Context, sc *ServerContext, config *DbConf
 	}
 	importOptions.BackupOldRev = base.BoolDefault(config.ImportBackupOldRev, false)
 
-	if config.ImportPartitions == nil {
+	if config.ImportPartitions == nil && sc.Config.IsServerless() {
```
adamcfraser: It's a bit of a style nit, but I think I would implement a `DefaultImportPartitions(serverless bool) uint16` function in base to encapsulate the handling here and above, particularly given the complexity it's adding to this particular if block.

adamcfraser (changes requested):
Minor style comment, otherwise looks good.

rest/serverless_test.go
```diff
@@ -579,3 +579,43 @@ func TestServerlessUnsuspendAdminAuth(t *testing.T) {
...
+	}{
+		{
+			name:               "serverless partitions",
+			expectedPartitions: base.Uint16Ptr(6),
+		},
+		{
+			name:               "serverless partitions with import_partition specified",
+			importPartition:    base.Uint16Ptr(8),
+			expectedPartitions: base.Uint16Ptr(8),
+		},
```
bbrks: Is it worth adding the test case for `serverless: false` == `16` or do we have sufficient coverage elsewhere?

bbrks (dismissed):
LGTM but see what you think about the comment

rest/serverless_test.go
gregns1: Good point! Updated the PR to add this test case in.

adamcfraser (approved):

PR #5915, opened 2022-11-16
CBG-2332 move changes feed calls down to collection level

torcolvin:
Starting with `SimpleMultiChangesFeed`, moved all callers to `DatabaseCollection`.

- Changed `N1QLQueryWithStats` to be a free function so it can work on `DatabaseCollection` and Database for principal objects.

Future work didn't push this into this PR because it's already quite long:

-  `ChangeEntry` doesn't have to track collectionID, and then probably be able to remove `TimedSetByCollectionID`. This might depend on principal changes.
-  Force blip to require collection property (this breaks both ISGR tests and blip unit tests).

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1103/
- [x] `GSI=false,xattrs=true`https://jenkins.sgwdev.com/job/SyncGateway-Integration/1104/

### DIFF
--- a/db/background_mgr_resync.go
+++ b/db/background_mgr_resync.go
@@ -61,9 +61,13 @@ func (r *ResyncManager) Run(ctx context.Context, options map[string]interface{},
 		persistClusterStatus()
 	}
 
-	_, err := database.UpdateAllDocChannels(ctx, regenerateSequences, callback, terminator)
-	if err != nil {
-		return err
+	for _, collection := range database.CollectionByID {
+		_, err := (&DatabaseCollectionWithUser{
+			DatabaseCollection: collection,
+		}).UpdateAllDocChannels(ctx, regenerateSequences, callback, terminator)
+		if err != nil {
+			return err
+		}
 	}
 
 	return nil
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -178,7 +178,7 @@ func (c *changeCache) Init(logCtx context.Context, collection *DatabaseCollectio
 
 	c.channelCache = channelCache
 
-	base.InfofCtx(c.logCtx, base.KeyCache, "Initializing changes cache for %s.%s.%s with options %+v", base.UD(collection.ScopeName()), base.UD(collection.Name()), c.options)
+	base.InfofCtx(c.logCtx, base.KeyCache, "Initializing changes cache for %s.%s.%s with options %+v", base.UD(collection.bucketName()), base.UD(collection.ScopeName()), base.UD(collection.Name()), c.options)
 
 	heap.Init(&c.pendingLogs)
 
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -583,8 +583,8 @@ type cachePrincipal struct {
 	Sequence uint64 `json:"sequence"`
 }
 
-func (c *changeCache) Remove(docIDs []string, startTime time.Time) (count int) {
-	return c.channelCache.Remove(docIDs, startTime)
+func (c *changeCache) Remove(collectionID uint32, docIDs []string, startTime time.Time) (count int) {
+	return c.channelCache.Remove(collectionID, docIDs, startTime)
 }
 
 // Principals unmarshalled during caching don't need to instantiate a real principal - we're just using name and seq from the document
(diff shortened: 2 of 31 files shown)


db/background_mgr_resync.go
```diff
@@ -61,9 +61,13 @@ func (r *ResyncManager) Run(ctx context.Context, options map[string]interface{},
 		persistClusterStatus()
 	}
 
-	_, err := database.UpdateAllDocChannels(ctx, regenerateSequences, callback, terminator)
-	if err != nil {
-		return err
+	for _, collection := range database.CollectionByID {
+		_, err := (&DatabaseCollectionWithUser{
```
adamcfraser, line 65: Should we be populating the DatabaseCollectionWithUser with database's user?
If not, should UpdateAllDocChannels be moved from DatabaseCollectionWithUser to DatabaseCollection?

I suspect it was defined on Database to simply calls to other Database methods, even if it always runs as admin.  If that's not worth refactoring right now, it feels like we should still respect any user that is potentially present on Database.

db/change_cache.go
```diff
@@ -167,7 +167,7 @@ func (c *changeCache) Init(logCtx context.Context, dbcontext *DatabaseContext, n
 	c.initTime = time.Now()
 	c.skippedSeqs = NewSkippedSequenceList()
 	c.lastAddPendingTime = time.Now().UnixNano()
-	c.sgCfgPrefix = base.SGCfgPrefixWithGroupID(dbcontext.Options.GroupID)
+	c.sgCfgPrefix = base.SGCfgPrefixWithGroupID(groupID)
```
adamcfraser: In this case I might prefer adding a DatabaseCollection.groupID() helper function that traverses back through the dbContext to get the groupID, what do you think?

db/change_cache.go
```diff
@@ -176,24 +176,20 @@ func (c *changeCache) Init(logCtx context.Context, dbcontext *DatabaseContext, n
...
 	}
 
-	channelCache, err := NewChannelCacheForContext(c.options.ChannelCacheOptions, c.context)
-	if err != nil {
-		return err
-	}
 	c.channelCache = channelCache
 
-	base.InfofCtx(c.logCtx, base.KeyCache, "Initializing changes cache for database %s with options %+v", base.UD(dbcontext.Name), c.options)
+	base.InfofCtx(c.logCtx, base.KeyCache, "Initializing changes cache for collection %s.%s with options %+v", base.UD(collection.ScopeName()), base.UD(collection.Name()), c.options)
```
adamcfraser: I think we still want database name in this style of logging (along with scope and collection), as in theory we could have two databases targeting 'bucket1.scope1.collection1' and 'bucket2.scope1.collection1' and we'd want to be able to differentiate the logging.

db/database_collection.go
```diff
@@ -38,27 +41,42 @@ func (c *DatabaseCollection) AllowConflicts() bool {
...
 func (c *DatabaseCollection) backupOldRev() bool {
 	return c.dbCtx.Options.ImportOptions.BackupOldRev
 
 }
 
-// changeCache returns the set of channels in memory. This is currently at the database level.
-func (c *DatabaseCollection) changeCache() *changeCache {
-	return c.dbCtx.changeCache
+// ChangeCache returns the set of channels in memory. This is currently at the database level.
+func (c *DatabaseCollection) ChangeCache() *changeCache {
```
adamcfraser: Is this public just to avoid the naming collision with c.changeCache?  If so, do we really need the accessor function?

If there's a good reason for it, I think we should still avoid making this public, either by renaming to GetChangeCache(), or rename the struct object changeCache_.  (neither are great, to be honest, would prefer omitting unless there's a good reason.)

adamcfraser (changes requested):
Generally looks ok, but I'm not 100% clear on the motivation to change to using accessors instead of struct properties in some cases (e.g. db.changeCache to dbCollection.ChangeCache()).  This just seems like an additional layer on the stack that's not providing any value.

db/background_mgr_resync.go
torcolvin: I don't see anywhere else I missed this, but in general yeah. I thought about making accessor methods for this, but as this is the only caller I'm not sure it's worthwhile.

db/change_cache.go
torcolvin: I like this idea, because you don't pass empty string in tests and we already have the value in another attr we pass in.

db/database_collection.go
torcolvin: Actually removing this, for its single use outside of DB exposed a bug where docIDs where removed from all channel caches. This is still hard to write unit tests for until we can have multiple collections. I wrote `DatabaseCollection.RemoveFromChangeCache`.

adamcfraser (approved):

PR #5930, opened 2022-11-23
CBG-2502: Revocation feed errors cause MultiChangesFeed to close

mohammed-madi:
CBG-2502

Describe your PR here...
- GetDocSyncData used to cause the MultiChangesFeed to close, it now skips the document and there is more logging for different types of failures
- Added GetWithXattr callback to Leaky bucket
- Added TestRevocationGetSyncDataError

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1266/


db/changes.go
```diff
@@ -241,9 +241,20 @@ func (db *DatabaseCollectionWithUser) buildRevokedFeed(ctx context.Context, ch c
...
 					if err != nil {
+						base.InfofCtx(ctx, base.KeyChanges, "Error getting sync data, seq: %v in channel %s, skipping document. Error: %v", seqID, base.UD(singleChannelCache.ChannelID()), err)
+						continue
+					}
+					requiresRevocation, err := db.wasDocInChannelPriorToRevocation(ctx, syncData, logEntry.DocID, singleChannelCache.ChannelID().Name, revocationSinceSeq)
+					if err != nil {
+						if syncData.History == nil {
+							base.TracefCtx(ctx, base.KeyChanges, "Document history empty: %s, no revocation sent", base.UD(logEntry.DocID))
+							continue
+						}
```
bbrks: This seems like it could be done before calling `db.wasDocInChannelPriorToRevocation`?

bbrks (changes requested):
Needs a rebase to resolve recent conflicts - left comment about error handling case.

db/changes.go
mohammed-madi: fixed in https://github.com/couchbase/sync_gateway/pull/5930/commits/144083e56a7d20c7a31ac30273957ea46cfc5145

db/changes.go
```diff
@@ -250,8 +250,19 @@ func (db *DatabaseCollectionWithUser) buildRevokedFeed(ctx context.Context, ch c
...
+					// Get doc sync data so we can verify the docs grant history
+					syncData, err := db.GetDocSyncData(ctx, logEntry.DocID)
+					if err != nil {
+						base.InfofCtx(ctx, base.KeyChanges, "Error getting sync data, seq: %v in channel %s, skipping document. Error: %v", seqID, base.UD(singleChannelCache.ChannelID()), err)
+						continue
+					}
+					if syncData.History == nil {
+						base.TracefCtx(ctx, base.KeyChanges, "Document history empty: %s, no revocation sent", base.UD(logEntry.DocID))
+						continue
+					}
```
bbrks: Last suggestion before merge: I think we can drop this check - we don't write `syncData` with nil history, but even if we did, all of the following code should work as-is using only `ChannelSet`/`ChannelSetHistory`

```suggestion
```

bbrks (changes requested):

db/changes.go
mohammed-madi: Addressed in https://github.com/couchbase/sync_gateway/pull/5930/commits/347eb319699a001848ede4f74b5231e1e7f3fd36

db/changes.go
```diff
@@ -250,8 +250,15 @@ func (db *DatabaseCollectionWithUser) buildRevokedFeed(ctx context.Context, ch c
 				// If its less than we can send a standard revocation with Sequence ID as above.
 				// Otherwise: we need to determine whether a previous revision of the document was in the channel prior
 				// to the since value, and only send a revocation if that was the case
+
 				if logEntry.Sequence > sinceVal {
-					requiresRevocation, err := db.wasDocInChannelPriorToRevocation(ctx, logEntry.DocID, singleChannelCache.ChannelID().Name, revocationSinceSeq)
+					// Get doc sync data so we can verify the docs grant history
+					syncData, err := db.GetDocSyncData(ctx, logEntry.DocID)
+					if err != nil {
+						base.InfofCtx(ctx, base.KeyChanges, "Error getting sync data, seq: %v in channel %s, skipping document. Error: %v", seqID, base.UD(singleChannelCache.ChannelID()), err)
```
bbrks: Minor tweak to log message:
- Remove "error" word (it raises too many false alarms from searches)
- Remove sequence value (not important/useful)
- Add DocID (this is useful, as we can go and check why that document had no sync data)

```suggestion
						base.InfofCtx(ctx, base.KeyChanges, "Couldn't get history for doc %q for channel %s, skipping revocation checks: %v", base.UD(logEntry.DocID), base.UD(singleChannelCache.ChannelID()), err)
```

bbrks (changes requested):
Will merge once this log message is tweaked as per suggestion

PR #5932, opened 2022-11-23
CBG-2507: reduce DCP + KV buffer size when in serverless

gregns1:
CBG-2507

PR aims to update the DCP buffer size and the KV buffer size to 1MB when running in server less mode. The PR includes a tests to test functionality where you can directly set the dcp buffer size and kv buffer size through the /db/_config API. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1397/


gregns1:
Screenshot 2022-12-19 at 14 39 59

Screenshot above is from a run I did earlier today. Seems to have worked all the allocation to the 20MB buffer is from test initialization, setting up rest tester and closing the bucket. One thing I want to bring up I put a serverless flag in the bucket spec as I needed to access the serverless value in line 251 on collection.go I may need some advice on how to avoid this and where I could possibly pull the flag from instead?

rest/config.go
```diff
@@ -1358,6 +1358,10 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
 		for bucket, _ := range sc.Config.BucketCredentials {
 			buckets = append(buckets, bucket)
 		}
+		err = sc.BootstrapContext.Connection.SetConnectionStringServerless()
```
adamcfraser: I had thought the connection string should be getting set when BootstrapContext.Connection is initialized (in CreateCouchbaseClusterFromStartupConfig), and should only be using Serverless values when the SG node is running in serverless mode.

base/bootstrap.go
```diff
@@ -32,6 +35,8 @@ type BootstrapConnection interface {
 	UpdateConfig(bucket, groupID string, updateCallback func(rawBucketConfig []byte, rawBucketConfigCas uint64) (updatedConfig []byte, err error)) (newCAS uint64, err error)
 	// Close releases any long-lived connections
 	Close()
+
+	SetConnectionStringServerless() error
```
adamcfraser: This seems like it could be set in NewCouchbaseCluster, and wouldn't need to be modified post-initialization.  If that's accurate, does it need to be on the BootstrapConnection interface?

base/bucket.go
```diff
@@ -132,6 +132,9 @@ type BucketSpec struct {
 	MaxConcurrentQueryOps         *int           // maximum number of concurrent query operations (default: DefaultMaxConcurrentQueryOps)
 	BucketOpTimeout               *time.Duration // How long bucket ops should block returning "operation timed out". If nil, uses GoCB default.  GoCB buckets only.
 	KvPoolSize                    int            // gocb kv_pool_size - number of pipelines per node. Initialized on GetGoCBConnString
+	KvBufferSize                  int
```
adamcfraser: Add comments to document these properties, similar to KvPoolSize.

base/bucket.go
```diff
@@ -132,6 +132,9 @@ type BucketSpec struct {
...
+	DcpBuffer                     int
+	Serverless                    bool
```
adamcfraser: I don't feel like 'Serverless' should be part of the bucket spec.  It looks like it's only here so that it can be used to set the DcpBuffer, but we've added DcpBuffer here too.

base/dcp_feed_type.go
```diff
@@ -171,6 +174,10 @@ func cbgtFeedParams(spec BucketSpec, scope string, collections []string, dbName
 		feedParams.Collections = collections
 	}
 
+	if spec.Serverless == true {
+		feedParams.DCPBuffer = 1000000
```
adamcfraser: This is hardcoded, but we have a value on spec we could be using?

base/bucket.go
```diff
@@ -159,15 +162,24 @@ func (spec BucketSpec) UseClientCert() bool {
 }
 
 type GoCBConnStringParams struct {
-	// The KV pool size, as passed down to gocbcore. Defaults to DefaultGocbKvPoolSize.
-	KVPoolSize int
+	// The KV pool size, KV buffer size and DCP buffer size are passed down to gocbcore.
+	// Defaults to different values based on being in serverless or not.
+	KVPoolSize   int
+	KVBufferSize int
+	DCPBuffer    int
```
adamcfraser: Should this be named DCPBufferSize for consistency?

base/bucket.go
```diff
@@ -188,12 +200,32 @@ func (spec *BucketSpec) GetGoCBConnString(params *GoCBConnStringParams) (string,
...
 	} else {
-		spec.KvPoolSize, _ = strconv.Atoi(poolSizeFromConnStr)
+		spec.KvPoolSize, _ = strconv.Atoi(fromConnStr)
+	}
+
+	fromConnStr = asValues.Get("kv_buffer_size")
+	if fromConnStr == "" && spec.KvBufferSize != 0 {
+		asValues.Set("kv_buffer_size", strconv.Itoa(spec.KvBufferSize))
+	} else if fromConnStr == "" {
+		asValues.Set("kv_buffer_size", strconv.Itoa(params.KVBufferSize))
```
adamcfraser, line 212: What's the scenario where spec.KvBufferSize and params.KvBufferSize are different values?

base/bucket.go
```diff
@@ -188,12 +200,32 @@ func (spec *BucketSpec) GetGoCBConnString(params *GoCBConnStringParams) (string,
...
+	fromConnStr := asValues.Get("kv_pool_size")
+	if fromConnStr == "" {
 		asValues.Set("kv_pool_size", strconv.Itoa(params.KVPoolSize))
 		spec.KvPoolSize = params.KVPoolSize
 	} else {
-		spec.KvPoolSize, _ = strconv.Atoi(poolSizeFromConnStr)
+		spec.KvPoolSize, _ = strconv.Atoi(fromConnStr)
+	}
+
+	fromConnStr = asValues.Get("kv_buffer_size")
```
adamcfraser: Reusing the fromConnStr variable isn't helpful here I don't think - it would be clearer if the previous naming convention was used (poolSizeFromConnStr, kvBufferSizeFromConnStr, etc)

base/constants.go
```diff
@@ -117,6 +117,12 @@ const (
 	// Number of kv connections (pipelines) per Couchbase Server node when running in serverless
 	DefaultGocbKvPoolSizeServerless = 1
 
+	// kv connections buffer size per Couchbase Server node
+	DefaultGoCBKvBufferSize = 20 * 1024 * 1024
```
adamcfraser: If we aren't overriding this value, I don't think we want to specify it at all (i.e. let goCB pick the default value themselves), for compatibility with any future change gocb makes.

base/dcp_client.go
```diff
@@ -28,6 +28,10 @@ const openStreamTimeout = 30 * time.Second
 const openRetryCount = uint32(10)
 const defaultNumWorkers = 8
 
+// DCP buffer size values based off if we are running in serverless or not
+const DefaultDCPBufferServerless = 1 * 1024 * 1024
+const DefaultDCPBuffer = 20 * 1024 * 1024
```
adamcfraser: Similar to the above - we don't want to explicitly set the value unless we're changing it.

base/dcp_client.go
```diff
@@ -189,9 +196,17 @@ func (dc *DCPClient) close() {
 }
 
 func (dc *DCPClient) initAgent(spec BucketSpec) error {
-	connStr, err := spec.GetGoCBConnString(&GoCBConnStringParams{
-		KVPoolSize: GoCBPoolSizeDCP,
-	})
+	defaultValues := &GoCBConnStringParams{}
+	if dc.Serverless {
```
adamcfraser: Similar to the previous case, I don't think we need a 'Serverless' property on the DCPClient.  We just need to set these values appropriately in the spec and retrieve them from there.

base/dcp_sharded.go
```diff
@@ -298,10 +298,18 @@ func initCBGTManager(ctx context.Context, bucket Bucket, spec BucketSpec, cfgSG
 	//   		https://github.com/couchbaselabs/cbgt/issues/1
 	//   		https://github.com/couchbaselabs/cbgt/issues/25
 	bindHttp := uuid
+	defaultValues := &GoCBConnStringParams{}
+	if serverless {
```
adamcfraser: See above about using spec instead of serverless param.

adamcfraser (changes requested):
A few suggestions about standardizing how we set values, and avoiding redundant settings.

base/dcp_feed_type.go
gregns1: As far as I am aware they are in different units for some reason so I was setting manually to 1000000 bytes whereas the gocb params take 1 * 1024 * 1024 for 1MB.

base/bucket.go
gregns1: If you set the values through unsupported config on the database, the params values here above will still be set to the default serverless ones. If we set these values through the unsuppotrted options, the spec values on the bucket will be set immediately and should be different to the default serverless values. So ultimately this check here is to see if we have already set the buffers through the database config.

rest/config.go
```diff
@@ -1454,6 +1458,10 @@ func (sc *ServerContext) FetchConfigs(ctx context.Context, isInitialStartup bool
 		//		i++
 		//	}
 		// }
+		err := sc.BootstrapContext.Connection.SetConnectionStringServerless()
+		if err != nil {
+			return nil, err
+		}
```
gregns1, line 1464: When running through on the dubugger there was times when we hit the function below `FetchDatabases` and the connection string was simple just couchbase://localhost with no serverless connection options so that initial cluster connection was using the normal buffer. This was my way around that. I saw your comment about adding this in `NewClusterConnection ` but I wasn't sure where I would get a serverless flag from there to ensure we only set it when runnning in serverless.

rest/config.go
gregns1: When running through on the dubugger there was times when we hit the function below FetchDatabases and the connection string was simple just couchbase://localhost with no serverless connection options so that initial cluster connection was using the normal buffer. This was my way around that. I saw your comment about adding this in NewClusterConnection but I wasn't sure where I would get a serverless flag from there to ensure we only set it when runnning in serverless. Here there is already a check for if we are running in serverless so thought it was a good place to call it.

adamcfraser: I don't think I agree - I still feel like it would be better to properly initialize the values when the BootstrapContext is created, especially since you're setting it as a global, BootstrapContext-wide value, and it doesn't just apply to the next GetConfig call.  Seems like it would be easy for a future scenario to try to specify the value on the context, and have a fetchDatabase call unexpectedly overwrite that value.

If there's times the connection string is just simple couchbase://localhost, I think the right thing to do is work out where that's being set in those cases.

base/bucket.go
adamcfraser: I still don't understand why we have two different ways to specify this information (spec and params).  Are there other use cases that require the use of params?

adamcfraser (changes requested):

base/bucket.go
gregns1: So for supporting changing the buffer sizes through the unsupported options api, inside my changes in server context I do a check to pull these values from the unsupported option if specified. If it is specified I add it to the bucket spec there. But after that function is called `GetBucketSpec` we call the `GetGoCBConnString` where we pass the default serverless values in. So I added these checks here to check if the spec values had already been set. If they have already been set its due to the unsupported options config and we don't want to override this with the default values being set in params. In contrast if no unsupported options are specified we need to set the string to the default serverless values. 
Is this the right approach or do you think I should just append these values onto the `server` string on the spec in `GetBucketSpec` and then In the `GetGoCBConnString` these values wouldn't be overwritten as they have been set already?

adamcfraser: I think it would be preferable to make the 'unsupported vs default' evaluation in _getOrAddDatabaseFromConfig, and use that to set the values of GoCBConnStringParams there.  That will make the behaviour of this function more understandable and consistent with what we do for kv pool size.

adamcfraser (changes requested):

adamcfraser (approved):

PR #5933, opened 2022-11-26
CBG-2554 Channel computation and invalidation for collections

adamcfraser:
CBG-2554

Extends the Principal and User interface to add collection-scoped API.

Adds Collections property to Authenticator config to allow callers to specify the set of collections that should be evaluated when channels are rebuilt.

Defines a new PrincipalCollectionAccess interface as a common interface for interaction with a collection's channel information that can be used for the default collection (stored on the principal root) or a named collection (stored in collection_access).

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1193/ (use default=false)
- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1194/ (use default=true)

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -22,7 +22,8 @@ import (
 	"gopkg.in/square/go-jose.v2/jwt"
 )
 
-/** Manages user authentication for a database. */
+// Authenticator manages user authentication for a database.
+
 type Authenticator struct {
 	datastore       base.DataStore
 	channelComputer ChannelComputer
@@ -36,13 +37,22 @@ type AuthenticatorOptions struct {
 	SessionCookieName        string
 	BcryptCost               int
 	LogCtx                   context.Context
+
+	// Collections defines the set of collections used by the authenticator when rebuilding channels.
+	// Channels are only recomputed for collections included in this set.
+	// This can be used to limit (re-)computation of channel access to only those collections relevant to
+	// a given operation. If not specified, is set to _default._default in NewAuthenticator constructor.
+	Collections map[string]map[string]struct{}
 }
 
 // Interface for deriving the set of channels and roles a User/Role has access to.
 // The instantiator of an Authenticator must provide an implementation.
 type ChannelComputer interface {
-	ComputeChannelsForPrincipal(context.Context, Principal) (ch.TimedSet, error)
-	ComputeRolesForUser(context.Context, User) (ch.TimedSet, error)
+	// ComputeChannelsForPrincipal returns the set of channels granted to the principal by documents in scope.collection
+	ComputeChannelsForPrincipal(ctx context.Context, p Principal, scope string, collection string) (ch.TimedSet, error)
+
+	// ComputeRolesForUser returns the set of roles granted to the user by documents in all collections
+	ComputeRolesForUser(ctx context.Context, u User) (ch.TimedSet, error)
 }
 
 type userByEmailInfo struct {
@@ -67,8 +77,15 @@ const (
 	GuestUserReadOnly = "Anonymous access is read-only"
 )
 
+var defaultCollectionMap = map[string]map[string]struct{}{base.DefaultScope: {base.DefaultCollection: struct{}{}}}
+
 // Creates a new Authenticator that stores user info in the given Bucket.
+
 func NewAuthenticator(datastore base.DataStore, channelComputer ChannelComputer, options AuthenticatorOptions) *Authenticator {
+
+	if len(options.Collections) == 0 {
+		options.Collections = defaultCollectionMap
+	}
 	return &Authenticator{
 		datastore:            datastore,
 		channelComputer:      channelComputer,
@@ -147,13 +164,16 @@ func (auth *Authenticator) getPrincipal(docID string, factory func() Principal) 
 			return nil, nil, false, pkgerrors.WithStack(base.RedactErrorf("base.JSONUnmarshal() error for doc ID: %s in getPrincipal().  Error: %v", base.UD(docID), err))
 		}
 		changed := false
-		if princ.Channels() == nil && !princ.IsDeleted() {
-			// Channel list has been invalidated by a doc update -- rebuild it:
-			if err := auth.rebuildChannels(princ); err != nil {
+		if !princ.IsDeleted() {
+			// Check whether any collection's channel list has been invalidated by a doc update -- if so, rebuild it:
+			channelsChanged, err := auth.rebuildChannels(princ)
+			if err != nil {
 				base.WarnfCtx(auth.LogCtx, "RebuildChannels returned error: %v", err)
 				return nil, nil, false, err
 			}
-			changed = true
+			if channelsChanged {
+				changed = true
+			}
 		}
 		if user, ok := princ.(User); ok {
 			if user.RoleNames() == nil {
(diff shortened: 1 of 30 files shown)


auth/auth.go
```diff
@@ -37,13 +38,22 @@ type AuthenticatorOptions struct {
 	SessionCookieName        string
 	BcryptCost               int
 	LogCtx                   context.Context
+
+	// Collections defines the set of collections used by the authenticator when rebuilding channels.
```
torcolvin, line 41: I might add something about how this is keyed on `map[scope]map[collection]`

auth/auth.go
```diff
@@ -23,7 +23,8 @@ import (
 	"gopkg.in/square/go-jose.v2/jwt"
 )
 
-/** Manages user authentication for a database. */
+// Authenticator user authentication for a database.  When running
```
torcolvin: docstring unfinished?

auth/auth.go
```diff
@@ -391,9 +446,14 @@ func (auth *Authenticator) InvalidateChannels(name string, isUser bool, invalSeq
...
+	subdocPath := "channel_inval_seq"
+	if scope != base.DefaultScope || collection != base.DefaultCollection {
+		subdocPath = "collection_access." + scope + "." + collection + "." + subdocPath
+	}
+
 	if auth.bucket.IsSupported(sgbucket.DataStoreFeatureSubdocOperations) {
-		err := auth.bucket.SubdocInsert(docID, "channel_inval_seq", 0, invalSeq)
-		if err != nil && err != base.ErrNotFound && err != base.ErrAlreadyExists {
+		err := auth.bucket.SubdocInsert(docID, subdocPath, 0, invalSeq)
+		if err != nil && err != base.ErrNotFound && err != base.ErrAlreadyExists && err != base.ErrPathNotFound {
```
torcolvin, line 456: How do we get a new error from this? Is this because sometimes we are trying to insert into collection or scopes that don't exist?

I don't feel strongly about this, but in two months I'm not going to remember what this means.

I might consider flattening

```
type ScopeAndCollection {
    Scope string
    Collection string
} 
```

which might make this more readability and flattens out the map. I didn't do a perf analysis on this.

auth/auth_test.go
```diff
@@ -290,18 +290,51 @@ func TestSaveRoles(t *testing.T) {
 }
 
 type mockComputer struct {
-	channels     ch.TimedSet
+	channels     map[string]map[string]ch.TimedSet
```
torcolvin, line 316: I think this might be worth a type alias and maybe a constructor?

auth/auth_test.go
```diff
@@ -311,39 +344,94 @@ func (self *mockComputer) ComputeRolesForUser(context.Context, User) (ch.TimedSe
...
+	options := DefaultAuthenticatorOptions()
+	options.Collections = map[string]map[string]struct{}{
+		base.DefaultScope: {base.DefaultCollection: struct{}{}},
+		"scope1":          {"collection1": struct{}{}},
+	}
+	auth := NewAuthenticator(bucket, &computer, options)
+	user, _ := auth.NewUser("testUser", "password", ch.BaseSetOf(t, "explicit1"))
+	user.SetCollectionExplicitChannels("scope1", "collection1", ch.AtSequence(ch.BaseSetOf(t, "explicit2"), 1), 0)
+	err := auth.Save(user)
+	assert.NoError(t, err)
```
torcolvin, line 407: i think this should be require, if this fails, subsequent tests will fail

auth/auth_test.go
```diff
@@ -311,39 +344,94 @@ func (self *mockComputer) ComputeRolesForUser(context.Context, User) (ch.TimedSe
...
+	bucket := base.GetTestBucket(t)
+	defer bucket.Close()
+	computer := mockComputer{}
+	computer.AddChannelsForCollection(base.DefaultScope, base.DefaultCollection, ch.AtSequence(ch.BaseSetOf(t, "derived1", "derived2"), 1))
+	computer.AddChannelsForCollection("scope1", "collection1", ch.AtSequence(ch.BaseSetOf(t, "derived3", "derived4"), 1))
+
+	options := DefaultAuthenticatorOptions()
+	options.Collections = map[string]map[string]struct{}{
+		base.DefaultScope: {base.DefaultCollection: struct{}{}},
+		"scope1":          {"collection1": struct{}{}},
```
torcolvin, line 401: if you want to explicitly use an invalid scope, I'd label it something as such so we know it doesn't exist "nonExistentScope" or something

auth/role_collection_access.go
```diff
@@ -58,10 +58,9 @@ func (role *roleImpl) CollectionChannels(scope, collection string) ch.TimedSet {
 			return nil
 		}
 		return cc.Channels_
+	} else {
```
torcolvin: super nit: might remove else here for clarity since this is end of function

base/error.go
```diff
@@ -46,6 +46,7 @@ var (
 	ErrChannelFeed           = &sgError{"Error while building channel feed"}
 	ErrXattrNotFound         = &sgError{"Xattr Not Found"}
 	ErrTimeout               = &sgError{"Operation timed out"}
+	ErrPathNotFound          = &sgError{"Subdoc path not found"}
```
torcolvin: should this be ErrSubDocPathNotFound or change the message?

db/crud.go
```diff
@@ -2311,9 +2312,9 @@ func isAccessError(err error) bool {
 	return base.ContainsString(base.SyncFnAccessErrors, err.Error())
 }
 
-// Recomputes the set of channels a User/Role has been granted access to by sync() functions.
+// Recomputes the set of channels a User/Role has been granted access to by sync() functions for the default collection
```
torcolvin, line 2315: Is this intentional to only work on default collection, or a TODO?

rest/access_test.go
```diff
@@ -1047,3 +1047,200 @@ func TestAccessOnTombstone(t *testing.T) {
...
+	base.SetUpTestLogging(t, base.LevelInfo, base.KeyAccess)
+
+	rtConfig := RestTesterConfig{SyncFn: `function(doc) {if(doc.type == "setaccess") {channel(doc.channel); access(doc.owner, doc.channel);} else { channel(doc.channel)}}`}
+	rt := NewRestTester(t, &rtConfig)
+	defer rt.Close()
+
+	ctx := rt.Context()
+	a := rt.ServerContext().Database(ctx, "db").Authenticator(ctx)
+	user, err := a.GetUser("")
+	assert.NoError(t, err)
```
torcolvin, line 1062: i'd drop requires on most of these, if the user is invalid the test continuing isn't helpful

rest/access_test.go
```diff
@@ -1047,3 +1047,200 @@ func TestAccessOnTombstone(t *testing.T) {
 	}
 
 }
+
+func TestDynamicChannelGrant(t *testing.T) {
+
+	base.SetUpTestLogging(t, base.LevelInfo, base.KeyAccess)
```
torcolvin, line 1053: Is marking test logging setup intentional for the test or temporary?

torcolvin (commented):
I put a lot of comments, but most of them are things you don't have to address, or are questions about the way the code is written. It looks reasonable enough to me.

What do we do if there's collection information in the traditional location and also in the new location? Do we want something to invalidate legacy channels on a doc if collections is enabled, or a future ticket to do this? Or do we need this for a rolling helium update?

auth/auth.go
adamcfraser: Right, subdocPath isn't a root level property when invalidating channels for a collection, and if it's trying to invalidate a collection that doesn't exist, it will get the PathNotFound error.  In this case we don't have anything that needs to be invalidated, so the error can be ignored.

rest/access_test.go
adamcfraser: This was intentional - this should be the standard pattern for enabling logging for our tests, unless I'm missing something.

base/error.go
adamcfraser: Updated the message.

bbrks (approved):
This looks good to me 👍

PR #5934, opened 2022-11-28
CBG-2505: New approach after mistake with previous PR on the same issue

gregns1:
CBG-2505

New approach to this flaking test issue after last approach filled disk with logs. *Still want to test more on this to ensure there is nothing i have missed on this.

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1135/

### DIFF
--- a/db/change_cache_test.go
+++ b/db/change_cache_test.go
@@ -904,7 +904,6 @@ func TestLowSequenceHandlingAcrossChannels(t *testing.T) {
 	_, err = verifySequencesInFeed(feed, []uint64{9})
 	assert.True(t, err == nil)
 
-	//changesCtxCancel()
 }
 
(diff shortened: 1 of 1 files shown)


torcolvin:
This data race doesn't occur since we made changes collection scoped.

I wanted to fix this code because I think the changes feed should be closable, but it's in fact not closable via context cancellation + waiting for feed to be closed. The `SimpleMultiChangesFeed` relies on `collection.mutationListener().tapNotifier` which is a `sync.Cond` and in order to have the feed exit via context, you actually need to send a `Broadcast` to wake this up.  @adamcfraser and I felt it would be of low value to fix a theoretical test only bug. The feed will get cleaned up when you close the database.

PR #5935, opened 2022-11-29
Support GetCollectionID on CBG-2266

torcolvin:
The crux of the problem is that for `base.Collection` types `GetCollectionID` would always return 0 for any named collection. Changed `NamedDataStore` to return an optional error (implemented for gocb only), so we can report if we can't get the collectionID from kv.

The problem with deferring `GetCollectionID` to happen later (in a synced operation, so it would only happen) is that we have to handle the error in a variety of unpleasant places, which is why I changed this in https://github.com/couchbase/sync_gateway/commit/0838041a2d4dd957f9bbfdbf7ad3440906ee5e56 . This has practical as well as ergonomic consequences.

This turns `NamedDataStore`  into a less concrete type than it used to be, which is also not entirely how it is meant.

https://github.com/couchbase/sg-bucket/pull/85
https://github.com/couchbaselabs/walrus/pull/73

This could be a stop-gap solution until we fix it.
## Pre-review checklist

- [x] Link upstream PRs


base/bucket.go
```diff
@@ -528,9 +528,10 @@ func AsSubdocStore(ds DataStore) (sgbucket.SubdocStore, bool) {
...
 // There's no WaitForReady operation in GoCB for collections, only Buckets, so attempting to use Exists in this way this seems like our best option to check for availability.
-func WaitUntilDataStoreExists(ds DataStore) error {
+func WaitUntilCollectionExists(collection *gocb.Collection) error {
 	return WaitForNoError(func() error {
-		_, err := ds.Exists("WaitUntilDataStoreExists")
+		_, err := collection.Exists("WaitUntilCollectionExists", nil)
 		return err
 	})
+
 }
```
bbrks: I'm not sure why this change is nessesary. It prevents usage for Walrus

base/collection_gocb.go
```diff
@@ -40,11 +40,23 @@ var (
 func AsCollection(dataStore DataStore) (*Collection, error) {
 	collection, ok := dataStore.(*Collection)
 	if !ok {
-		return nil, errors.New("dataStore is not a *Collection")
+		return nil, fmt.Errorf("dataStore is not a %T", dataStore)
```
bbrks: This change doesn't look correct? Maybe this:

```suggestion
		return nil, fmt.Errorf("dataStore is not a *Collection (got %T)", dataStore)
```

rest/doc_api.go
```diff
@@ -448,6 +448,8 @@ func (h *handler) handlePutDoc() error {
 	var newRev string
 	var doc *db.Document
 
+	fmt.Printf("HONK collection=%+v\n", h.collection)
+
```
bbrks:
```suggestion
```

base/collection_gocb.go
```diff
@@ -40,11 +40,23 @@ var (
...
 	}
 	return collection, nil
 }
 
+// NewCollection creates a new collection object, caching some kv ops.
+func NewCollection(bucket *GocbV2Bucket, collection *gocb.Collection) (*Collection, error) {
+	c := &Collection{
+		Bucket:     bucket,
+		Collection: collection}
+	err := c.setCollectionID()
```
bbrks, line 53: This was the key piece I was missing, but the way this is still continuing to create collections on `NamedDataStore()` access means we're not actually caching this value beyond a single request.

As mentioned in Slack I think we may need to rethink how we store/fetch collections in general and have ideas about how to do this. But for now if we're able to set a collection ID for a collection it should be sufficient.

rest/server_context.go
```diff
@@ -484,11 +484,9 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
...
 			for collectionName, _ := range scopeConfig.Collections {
-				dataStore := bucket.NamedDataStore(base.ScopeAndCollectionName{Scope: scopeName, Collection: collectionName})
-
-				// Check if scope/collection specified exists. Will enter retry loop if connection unsuccessful
-				if err := base.WaitUntilDataStoreExists(dataStore); err != nil {
-					return nil, fmt.Errorf("attempting to create/update database with a scope/collection that is not found")
+				dataStore, err := bucket.NamedDataStore(base.ScopeAndCollectionName{Scope: scopeName, Collection: collectionName})
+				if err != nil {
+					return nil, err
 				}
```
bbrks: I'm not convinced we should be pushing the `WaitUntilDataStoreExists` logic down into the Collection constructor for a couple of reasons:

- It changes the sg-bucket API to be more clunky, by now returning an error
- It makes accessing a collection a _lot_ more costly (it's actually doing a KV op under the hood for each invocation now for the exists check, in addition to the agent op to fetch kvCollectionID)

bbrks (changes requested):
These were old PR comments before discussing on Slack but to reiterate: I think we can get away with calling `setCollectionID()` in `NamedDataStore` for now - until we have a better way of keeping collection state on a bucket for cheaper access.

base/bucket.go
torcolvin: I changed this to only occur with a cast, but I could do a type cast in this function, see what you think in the next PR.

rest/server_context.go
torcolvin: We could cache this information on `Bucket` but there are two problems with this approach:

1. User creates collection `foo` and it comes back with CollectionID 2
2. User drops collection `foo` while SG is running but database is offline and it has CollectionID 3

So we don't know when these get invalidated. If we always get a new bucket connection when we re-create the dbcontext, or take dbContext on and offline, we could cache with a call when we connect to bucket by moving `base.Collection.GetCollectionManifest` to `Gocbv2Bucket` and store all manifests on Bucket, if we are not worried about invalidation problem. If we don't find the collectionID in this manifest (newly created collection), we could get the manifest again, and this makes getting `NamedDataStore` faster again for 99% of cases, but it could still return an error if KV is no available.

We'd store a map of _all_ collections on a bucket even though we don't use them all, which is a probably a small memory cost.

bbrks (approved):
LGTM - I've implemented `sgbucket.DataStoreName` on `WalrusCollection` and pulled into my branch to make the failing `TestGetDatabaseCollectionWithUserDefaultCollection` pass with Walrus.

PR #5952, opened 2022-12-06
CBG-2266 cleanup collections

torcolvin:
Without explicitly deleting these scopes, the bucket emptier function will fail because it tries to drop the scope with a primary index and once this test is run, we'll lose access to this bucket.

An example, using 1 bucket with this test and the subsequent test will show the failure:

```
env SG_TEST_COLLECTION_POOL_SIZE=2 SG_TEST_BUCKET_POOL_SIZE=1 SG_TEST_BUCKET_POOL_PRESERVE=true SG_TEST_BUCKET_POOL_DEBUG=true SG_TEST_USE_GSI=true SG_TEST_BACKING_STORE=couchbase go test -tags cb_sg_enterprise ./rest -run '(TestCollectionsChangeConfigScope|TestRoot)'  -v -failfast
```

will produce

```
2022-12-06T11:10:05.630-05:00 TEST: b:sg_int_0_1670342845169039014 Running bucket through readier function
2022-12-06T11:10:05.630-05:00 TEST: b:sg_int_0_1670342845169039014 emptying bucket via N1QL, readying views and indexes
2022-12-06T11:10:05.641-05:00 TEST: b:sg_int_0_1670342845169039014 Couldn't ready bucket, got error: bucket does not have primary index, so can't empty bucket using N1QL - Retrying
```

and not ready the bucket ever.

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1199/

### DIFF
--- a/rest/api_collections_test.go
+++ b/rest/api_collections_test.go
@@ -432,7 +432,7 @@ func TestCollectionsChangeConfigScope(t *testing.T) {
 	require.NoError(t, err)
 	defer func() {
 		collection, err := base.AsCollection(tb.DefaultDataStore())
-		assert.NoError(t, err)
+		require.NoError(t, err)
 		cm := collection.Collection.Bucket().Collections()
 		for scope := range scopesAndCollections {
 			assert.NoError(t, cm.DropScope(scope, nil))
--- a/rest/api_collections_test.go
+++ b/rest/api_collections_test.go
@@ -419,15 +419,26 @@ func TestCollectionsChangeConfigScope(t *testing.T) {
 	tb := base.GetTestBucket(t)
 	defer tb.Close()
 	ctx := base.TestCtx(t)
-	err := base.CreateBucketScopesAndCollections(ctx, tb.BucketSpec, map[string][]string{
+
+	scopesAndCollections := map[string][]string{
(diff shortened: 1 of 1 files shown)


rest/api_collections_test.go
```diff
@@ -419,15 +419,26 @@ func TestCollectionsChangeConfigScope(t *testing.T) {
...
 		"quxScope": {
 			"quux",
 		},
-	})
+	}
+	err := base.CreateBucketScopesAndCollections(ctx, tb.BucketSpec, scopesAndCollections)
 	require.NoError(t, err)
+	defer func() {
+		collection, err := base.AsCollection(tb.DefaultDataStore())
+		assert.NoError(t, err)
```
bbrks: nitpick but possibly avoid panic if we get here without it being a collection?

```suggestion
		require.NoError(t, err)
```

bbrks (dismissed):

rest/api_collections_test.go
torcolvin: you usually don't want to do `require` inside a defer, but this might make sense here

bbrks (approved):

PR #5955, opened 2022-12-07
CBG-2266 make rest tests run with non default collections by default

torcolvin:
To facilitate testing collections across the rest packages, several variants of `NewRestTester` have been created:

* `NewRestTester` itself has modified behavior. Under `SG_TEST_USE_DEFAULT_COLLECTION=false` it will now create a Database with a single named collection on it with walrus or GSI. With `SG_TEST_USE_DEFAULT_COLLECTION=true` it will maintain the existing behavior, of database backed by default collection.
* `NewRestTesterDefaultCollection` will always use a default collection. These are tagged with jira tickets where we want to modify the behavior to work with a named collection (e.g. replication or channel access).
* `NewRestTesterMultipleCollections` creates a rest tester with 1 database and N named collections. This is limited to GSI/walrus.

Subsequent changes that occur are modifying use of `DefaultDataStore` to the datastore off a collection.

I also modified `TestBucket.GetNamedDataStore` to return datastores in particular order. It was ordered under walrus but map iteration under GSI. This is test only code.

## Pre-review checklist

- [ ] `GSI=true,xattrs=true,single named collection` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1221/
- [x] `GSI=true,xattrs=true,default colleciton only collection` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1213/

### DIFF
--- a/rest/changesttest/changes_collection_test.go
+++ b/rest/changesttest/changes_collection_test.go
@@ -25,22 +25,12 @@ import (
 func TestMultiCollectionChangesAdmin(t *testing.T) {
 
 	base.SetUpTestLogging(t, base.LevelDebug, base.KeyHTTP, base.KeyChanges, base.KeyCache)
-	testBucket := base.GetTestBucket(t)
-	scopesConfig, keyspaces := getMultiCollectionConfig(t, testBucket, 2)
-	rtConfig := &rest.RestTesterConfig{
-		CustomTestBucket: testBucket,
-		DatabaseConfig: &rest.DatabaseConfig{
-			DbConfig: rest.DbConfig{
-				Scopes: scopesConfig,
-			},
-		},
-	}
+
+	rt, keyspaces := rest.NewRestTesterMultipleCollections(t, nil, 2)
+	defer rt.Close()
 
 	c1Keyspace := keyspaces[0]
 	c2Keyspace := keyspaces[1]
-
-	rt := rest.NewRestTester(t, rtConfig)
-	defer rt.Close()
 
 	// Put several documents, will be retrieved via query
 	response := rt.SendAdminRequest("PUT", "/"+c1Keyspace+"/pbs1", `{"value":1, "channels":["PBS"]}`)
@@ -91,22 +81,12 @@ func TestMultiCollectionChangesAdmin(t *testing.T) {
 func TestMultiCollectionChangesAdminSameChannelName(t *testing.T) {
 
 	base.SetUpTestLogging(t, base.LevelDebug, base.KeyHTTP, base.KeyChanges, base.KeyCache)
-	testBucket := base.GetTestBucket(t)
-	scopesConfig, keyspaces := getMultiCollectionConfig(t, testBucket, 2)
-	rtConfig := &rest.RestTesterConfig{
-		CustomTestBucket: testBucket,
-		DatabaseConfig: &rest.DatabaseConfig{
-			DbConfig: rest.DbConfig{
-				Scopes: scopesConfig,
-			},
-		},
-	}
+
+	rt, keyspaces := rest.NewRestTesterMultipleCollections(t, nil, 2)
+	defer rt.Close()
 
 	c1Keyspace := keyspaces[0]
 	c2Keyspace := keyspaces[1]
-
-	rt := rest.NewRestTester(t, rtConfig)
-	defer rt.Close()
 
 	// Put several documents, will be retrieved via query
 	response := rt.SendAdminRequest("PUT", "/"+c1Keyspace+"/pbs1_c1", `{"value":1, "channels":["PBS"]}`)
@@ -156,22 +136,12 @@ func TestMultiCollectionChangesAdminSameChannelName(t *testing.T) {
 func TestMultiCollectionChangesUser(t *testing.T) {
 
 	base.SetUpTestLogging(t, base.LevelDebug, base.KeyHTTP, base.KeyChanges, base.KeyCache, base.KeyCRUD)
-	testBucket := base.GetTestBucket(t)
-	scopesConfig, keyspaces := getMultiCollectionConfig(t, testBucket, 2)
-	rtConfig := &rest.RestTesterConfig{
-		CustomTestBucket: testBucket,
-		DatabaseConfig: &rest.DatabaseConfig{
-			DbConfig: rest.DbConfig{
-				Scopes: scopesConfig,
-			},
-		},
-	}
+
+	rt, keyspaces := rest.NewRestTesterMultipleCollections(t, nil, 2)
+	defer rt.Close()
 
 	c1Keyspace := keyspaces[0]
 	c2Keyspace := keyspaces[1]
-
-	rt := rest.NewRestTester(t, rtConfig)
-	defer rt.Close()
 
 	// Create user with access to channel PBS in both collections
 	ctx := rt.Context()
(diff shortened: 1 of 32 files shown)


rest/utilities_testing.go
```diff
@@ -1016,6 +1089,9 @@ type BlipTesterSpec struct {
 
 	// Supported blipProtocols for the client to use in order of preference
 	blipProtocols []string
+
+	// If true, only run the default collection
+	onlyDefaultCollection bool
```
adamcfraser: Is this actually used anywhere?  I don't see any reads of this property.

rest/utilities_testing.go
```diff
@@ -1968,3 +2067,18 @@ func (sc *ServerContext) suspendDatabase(t *testing.T, ctx context.Context, dbNa
 
 	return sc._suspendDatabase(ctx, dbName)
 }
+
+// getRESTkeyspace returns a keyspace for REST URIs
+func getRESTKeyspace(_ testing.TB, dbName string, collection *db.DatabaseCollection) string {
+	return strings.Join([]string{dbName, collection.ScopeName(), collection.Name()}, base.ScopeCollectionSeparator)
```
adamcfraser, line 2077: Could this use collection.dbCtx to get the dbName, to avoid having to pass it in?

rest/utilities_testing.go
```diff
@@ -236,10 +268,16 @@ func (rt *RestTester) Bucket() base.Bucket {
 			// If no db config was passed in, create one
 			rt.DatabaseConfig = &DatabaseConfig{}
 		}
-
 		if rt.DatabaseConfig.UseViews == nil {
 			rt.DatabaseConfig.UseViews = base.BoolPtr(base.TestsDisableGSI())
 		}
+		if rt.collectionConfig != useSingleCollectionDefaultOnly && (!*rt.DatabaseConfig.UseViews || base.UnitTestUrlIsWalrus()) {
```
adamcfraser: This condition doesn't handle the case where rt.collectionConfig = useSingleCollection and SG_TEST_USE_DEFAULT_COLLECTION=true.  I'm not clear why that isn't needed.

rest/utilities_testing.go
```diff
@@ -1062,11 +1138,17 @@ func (bt BlipTester) DatabaseContext() *db.DatabaseContext {
...
+}
+
+// NewBlipTesterFromSpecWithRT creates a blip tester from an existing rest tester
 func NewBlipTesterFromSpecWithRT(tb testing.TB, spec *BlipTesterSpec, rt *RestTester) (blipTester *BlipTester, err error) {
 	blipTesterSpec := spec
 	if spec == nil {
-		// Default spec
-		blipTesterSpec = &BlipTesterSpec{}
+		spec := getDefaultBlipTesterSpec()
+		blipTesterSpec = &spec
```
adamcfraser: It looks like this didn't previously set GuestEnabled?  I'm not even sure why that would be the default, but we should at minimum not change previous behaviour here.

rest/utilities_testing.go
```diff
@@ -381,6 +434,26 @@ func (rt *RestTester) GetDatabase() *db.DatabaseContext {
...
+	database := rt.GetDatabase()
+	require.Equal(rt.TB, 1, len(database.CollectionByID))
+	for _, collection := range database.CollectionByID {
+		return collection
+	}
+	return nil
+}
+
+// GetSingleTestDataStore will return a datastore if there is only one collection configured on the RestTester database.
+func (rt *RestTester) GetSingleTestDataStore() base.DataStore {
```
adamcfraser: Maybe a nitpick, but what do you think about calling this GetSingleDataStore()?  

The current name had me thinking there was a struct TestDataStore (similar to TestBucket), and since this is a method on RestTester I think the 'Test' is redundant here.

rest/utilities_testing.go
```diff
@@ -81,7 +91,13 @@ type RestTester struct {
 	closed                  bool
 }
 
+// NewRestTester returns a rest tester backed by a single database and a single collection. This collection may be named or default collection based on global test configuration.
```
adamcfraser, line 95: When using NewRestTester(), how are the tests that use /db/ in paths resolving to a single named collection when running with SG_TEST_USE_DEFAULT_COLLECTION=false?  Are they still relying on the handler.go code to pick a single collection that needs to be removed?

adamcfraser (changes requested):

rest/utilities_testing.go
torcolvin: This is a mistake, I meant to enable this!

rest/utilities_testing.go
torcolvin: fixed

rest/utilities_testing.go
torcolvin: Without setting `GuestEnabled`, blip tester will fail to run correctly - the server fails with a handshake error. I don't think this code is executing, but it's out of scope to modify here.

rest/utilities_testing.go
torcolvin: I feel OK about this, but I used this to clarify that these were OK to use because it is explicit test usage where I've made the change. I think we have to evaluate if we are using it in a non test case with a grep, and I think this is fine for `RestTester` because the bugs would be test only.

rest/utilities_testing.go
torcolvin: This is out of scope for this PR, but this works because if there's a single collection `foo.bar` the following are equivalent as per part of PRD. We discussed changing this:

- /db
- /db.bar
- /db.foo.bar

But I think `/db` should only be `/db._default._default` to avoid confusion, but we'll have to add new functions that get keyspaces and this code review was already really long.

adamcfraser: Yes, I think it makes sense to have this PR deliver "Run RestTester against a single named collection when SG_TEST_USE_DEFAULT_COLLECTION=false", and have the "fix RestTester tests that depend on /db/ mapping to /db.foo.bar/" be a separate ticket.

torcolvin: Filed https://issues.couchbase.com/browse/CBG-2622 for this

rest/changesttest/changes_collection_test.go
```diff
@@ -389,11 +390,11 @@ func TestMultiCollectionChangesCustomSyncFunctions(t *testing.T) {
 		},
 	}
 
-	rt, keyspaces := rest.NewRestTesterMultipleCollections(t, rtConfig, numCollections)
+	rt, _ := rest.NewRestTesterMultipleCollections(t, rtConfig, numCollections)
 	defer rt.Close()
 
-	c1Keyspace := keyspaces[0]
-	c2Keyspace := keyspaces[1]
+	c1Keyspace := strings.Join([]string{rt.GetDatabase().Name, dataStoreNames[0].ScopeName(), dataStoreNames[0].CollectionName()}, base.ScopeCollectionSeparator)
```
adamcfraser: We definitely don't want to have to be writing this for every test that wants to use keyspaces - that was the entire motivation to return the keyspace as a simple string. What required this change?

adamcfraser (changes requested):

rest/changesttest/changes_collection_test.go
torcolvin: Refactoring mishap.

adamcfraser (approved):

PR #5957, opened 2022-12-07
CBG-2556: Make ISGR checkpointer callbacks deal with pre-parsed SequenceID

bbrks:
CBG-2556: Make ISGR checkpointer callbacks deal with pre-parsed `SequenceID`s
- The checkpoint format has not changed. That remains a plain non-JSON-encoded string.

This has two benefits:
1. efficiency gains due to not re-parsing sequence numbers whenever we run a checkpoint
2. consistency for sequence number formatting across checkpointer expected/processed lists

@adamcfraser I would like a review especially around testing of this (I added 4 internal checkpointer stats to track lengths of sequence sets, but I'm not sure this is ideal)

- [x] `xattrs=true,gsi=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1211/

### DIFF
--- a/db/active_replicator_checkpointer.go
+++ b/db/active_replicator_checkpointer.go
@@ -34,12 +34,12 @@ type Checkpointer struct {
 	// lock guards the expectedSeqs slice, and processedSeqs map
 	lock sync.Mutex
 	// expectedSeqs is an ordered list of sequence IDs we expect to process revs for
-	expectedSeqs []string
+	expectedSeqs []SequenceID
 	// processedSeqs is a map of sequence IDs we've processed revs for
-	processedSeqs map[string]struct{}
+	processedSeqs map[SequenceID]struct{}
 	// idAndRevLookup is a temporary map of DocID/RevID pair to sequence number,
 	// to handle cases where we receive a message that doesn't contain a sequence.
-	idAndRevLookup map[IDAndRev]string
+	idAndRevLookup map[IDAndRev]SequenceID
 	// ctx is used to stop the checkpointer goroutine
 	ctx context.Context
 	// lastRemoteCheckpointRevID is the last known remote checkpoint RevID.
@@ -47,7 +47,7 @@ type Checkpointer struct {
 	// lastLocalCheckpointRevID is the last known local checkpoint RevID.
 	lastLocalCheckpointRevID string
 	// lastCheckpointSeq is the last checkpointed sequence
-	lastCheckpointSeq string
+	lastCheckpointSeq SequenceID
 
 	stats CheckpointerStats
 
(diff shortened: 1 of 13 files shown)


db/blip_handler.go
```diff
@@ -667,7 +673,12 @@ func (bh *blipHandler) handleChanges(rq *blip.Message) error {
 
 			// skip parsing seqno if we're not going to use it (no callback defined)
 			if bh.sgr2PullAddExpectedSeqsCallback != nil {
-				expectedSeqs[IDAndRev{DocID: docID, RevID: revID}] = seqStr(change[0])
+				seq, err := ParseJSONSequenceID(seqStr(change[0]))
+				if err != nil {
+					// We've already asked for the doc/rev for the sequence so assume we're going to receive it... Just log this and carry on
+					base.WarnfCtx(bh.loggingCtx, "Unable to parse expected sequence %q for %q/%q: %v", change[0], base.UD(docID), revID, err)
+				}
+				expectedSeqs[IDAndRev{DocID: docID, RevID: revID}] = seq
```
bbrks: Going back and re-reading this, it means we'll be storing `{Seq:0}` as an expected sequence for a given doc/rev which sounds like a bad idea 🤔

It's possible to move this sequence parsing up above the callbacks, so that we don't request the revision at all in that case, which I think is preferable.

bbrks: @adamcfraser based on discussion I think we can leave the warning as-is, but don't populate expectedSeqs (meaning invalid sequences remain in the processed map when recieved, as we weren't expecting it)

db/active_replicator_checkpointer.go
```diff
@@ -58,12 +58,16 @@ type Checkpointer struct {
...
 type CheckpointerStats struct {
-	ExpectedSequenceCount     int64
-	ProcessedSequenceCount    int64
-	AlreadyKnownSequenceCount int64
-	SetCheckpointCount        int64
-	GetCheckpointHitCount     int64
-	GetCheckpointMissCount    int64
+	ExpectedSequenceCount           int64
+	ExpectedSequenceLen             int
+	ExpectedSequenceLenPostCleanup  int
```
adamcfraser, line 63: I went back and forth on it, but I'm fine with the extra stats, even just for test usage.  They are inexpensive and add decent value from a testability perspective.

db/active_replicator_checkpointer.go
```diff
@@ -207,17 +212,17 @@ func (c *Checkpointer) CheckpointNow() {
...
 
 	seq := c._updateCheckpointLists()
-	if seq == "" {
+	if seq == nil {
 		return
 	}
 
 	base.InfofCtx(c.ctx, base.KeyReplicate, "checkpointer: calculated seq: %v", seq)
-	err := c._setCheckpoints(seq, status)
+	err := c._setCheckpoints(*seq, status)
```
adamcfraser: Are we intentionally passing by value here, or can we switch to avoid a copy?

db/active_replicator_checkpointer.go
```diff
@@ -422,17 +430,22 @@ func (c *Checkpointer) fetchCheckpoints() (*ReplicationStatus, error) {
 		c.stats.GetCheckpointHitCount++
 	}
 
-	c.lastCheckpointSeq = checkpointSeq
+	parsedCheckpointSeq, err := ParsePlainSequenceID(checkpointSeq)
+	if err != nil {
```
adamcfraser: On a failed parse, do you think we should be setting the sequence to zero for rollback, like we do above?  Otherwise, is this situation recoverable?

adamcfraser (changes requested):
Generally looks good, a few small comments.

db/active_replicator_checkpointer.go
bbrks: could pass the pointer, updated

db/active_replicator_checkpointer.go
bbrks: We're able to warn and use seq:0 so I've done that and consider it a checkpoint miss for stats

adamcfraser (approved):

PR #5958, opened 2022-12-08
CBG-2333 Support multi-collection changesCache and _changes

adamcfraser:
CBG-2333

Send all DCP events to all change caches for sequence buffering, only add to channel cache when collection matches.  Switches changeListener's OnDocChanged to a subscription model, so that each collection's changeCache can be notified.

Also reworks to handler.invoke to only perform default scope/collection check in handler for keyspace paths, and to perform that check after auth.

Fix to keyspace routing meant two additional routes had to be switched to keyspace:
 - all_docs to support include_docs=true
 - root document POST

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1215/

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -682,8 +682,8 @@ func (context *DatabaseContext) GetOIDCProvider(providerName string) (*auth.OIDC
 }
 
 // Registers an on change callback.  Each change cache (per collection) will register here
-func (context *DatabaseContext) RegisterOnChangeCallback(collectionID uint32, callback DocChangedFunc) {
-	context.mutationListener.RegisterOnChangeCallback(collectionID, callback)
+func (context *DatabaseContext) RegisterOnChangeCallback(collectionID uint32, callback DocChangedFunc) error {
+	return context.mutationListener.RegisterOnChangeCallback(collectionID, callback)
 }
 
 func (dbCtx *DatabaseContext) GetServerUUID(ctx context.Context) string {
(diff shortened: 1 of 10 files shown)


rest/handler.go
```diff
@@ -468,9 +470,11 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
 		if err != nil {
 			return err
 		}
-		h.collection, err = h.db.GetDatabaseCollectionWithUser(*keyspaceScope, *keyspaceCollection)
-		if err != nil {
-			return err
+		if ks != "" {
+			h.collection, err = h.db.GetDatabaseCollectionWithUser(*keyspaceScope, *keyspaceCollection)
```
torcolvin, line 474: I'm a bit worried it is dangerous to not always set a a collection, because we never check later to see if `h.collection` is nil or not. We also never check to see if `h.db` is present so this is probably fine.

rest/utilities_testing.go
```diff
@@ -798,6 +802,40 @@ func (rt *RestTester) PutDocumentWithRevID(docID string, newRevID string, parent
...
+	newConfigBytes, _ := base.JSONMarshal(currentConfig)
+
+	userResponse = rt.SendAdminRequest("PUT", "/"+dbName+"/_user/"+username, string(newConfigBytes))
+	if userResponse.Code != 200 {
+		return fmt.Errorf("User update failed: %s", userResponse.Body.Bytes())
+	}
+	return nil
+}
+
+func SplitKeyspace(keyspace string) (dbName, scopeName, channelName string) {
```
torcolvin: This is a dupe of `rest.parse.Keyspace`

db/database.go
```diff
@@ -2133,7 +2133,10 @@ func newDatabaseCollection(ctx context.Context, dbContext *DatabaseContext, data
 		return nil, err
 	}
 	// Set the DB Context notifyChange callback to call back the changecache DocChanged callback
-	dbContext.RegisterOnChangeCallback(dbCollection.GetCollectionID(), dbCollection.changeCache.DocChanged)
+	err = dbContext.RegisterOnChangeCallback(dbCollection.GetCollectionID(), dbCollection.changeCache.DocChanged)
+	if err != nil {
```
torcolvin, line 2137: should this be fatal? I don't think the database is really set up in this case

torcolvin (dismissed):

torcolvin (approved):

PR #5960, opened 2022-12-09
CBG-2774 Run collection-specific sync functions

adamcfraser:
CBG-2774 

Run collection-specific sync functions

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1219/

### DIFF
--- a/base/constants_syncdocs.go
+++ b/base/constants_syncdocs.go
@@ -10,7 +10,10 @@ licenses/APL2.txt.
 
 package base
 
-import "errors"
+import (
+	"errors"
+	"fmt"
+)
 
 const SyncDocPrefix = "_sync:" // SyncDocPrefix is a common prefix for all Sync Gateway metadata documents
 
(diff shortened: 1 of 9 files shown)


base/constants_syncdocs.go
```diff
@@ -37,11 +40,12 @@ const (
...
-	PersistentConfigPrefixWithoutGroupID = SyncDocPrefix + "dbconfig:" // PersistentConfigPrefixWithoutGroupID stores a database config
-	SyncFunctionKeyWithoutGroupID        = SyncDocPrefix + "syncdata"  // SyncFunctionKeyWithoutGroupID stores a copy of the Sync Function
+	DCPCheckpointPrefixWithoutGroupID       = SyncDocPrefix + "dcp_ck:"             // DCPCheckpointPrefixWithoutGroupID stores a DCP checkpoint
+	SGCfgPrefixWithoutGroupID               = SyncDocPrefix + "cfg"                 // SGCfgPrefixWithoutGroupID stores CfgSG/cbgt data
+	HeartbeaterPrefixWithoutGroupID         = SyncDocPrefix                         // HeartbeaterPrefixWithoutGroupID stores a SG node heartbeat document
+	PersistentConfigPrefixWithoutGroupID    = SyncDocPrefix + "dbconfig:"           // PersistentConfigPrefixWithoutGroupID stores a database config
+	SyncFunctionKeyWithoutGroupID           = SyncDocPrefix + "syncdata"            // SyncFunctionKeyWithoutGroupID stores a copy of the Sync Function
+	CollectionSyncFunctionKeyWithoutGroupID = SyncDocPrefix + "syncdata_collection" // CollectionSyncFunctionKeyWithoutGroupID stores a copy of the Sync Function
```
bbrks, line 48: I've been keeping `key` and `prefix` distinct in these to avoid confusion/accidental misuse and I think we should do the same here

```suggestion
	CollectionSyncFunctionPrefixWithoutGroupID = SyncDocPrefix + "syncdata_collection" // CollectionSyncFunctionPrefixWithoutGroupID stores a copy of the Sync Function
```

db/database_collection.go
```diff
@@ -256,3 +262,50 @@ func (c *DatabaseCollectionWithUser) User() auth.User {
...
+
+// ////// SYNC FUNCTION:
+
+// Sets the database context's sync function based on the JS code from config.
+// Returns a boolean indicating whether the function is different from the saved one.
+// If multiple gateway instances try to update the function at the same time (to the same new
+// value) only one of them will get a changed=true result.
+func (dc *DatabaseCollection) UpdateSyncFun(ctx context.Context, syncFun string) (changed bool, err error) {
+	if syncFun == "" {
+		dc.ChannelMapper = nil
```
torcolvin, line 274: Should this actually point back to parent sync function, or is this reinitializing with the default?

bbrks: I don't think we want inheritence on the sync function, but the current RestTester workaround is sort of providing that until we're able to drop that behaviour.

bbrks (changes requested):
Looks good. One nitpick that you can ignore

bbrks (approved):

PR #5961, opened 2022-12-09
CBG-2620: Add star channel handling and unit test

mohammed-madi:
CBG-2620

Describe your PR here...
- Add star channel handling to revocation processing in changes.go
- Add unit test that removes star channel access from a user, and ensures docs are purged after.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1220/


db/changes.go
```diff
@@ -347,7 +347,7 @@ func (db *DatabaseCollectionWithUser) wasDocInChannelPriorToRevocation(ctx conte
 	// Iterate over the channel history information on the document and find any periods where the doc was in the
 	// channel and the channel was accessible by the user
 	for _, docHistoryEntry := range append(syncData.ChannelSet, syncData.ChannelSetHistory...) {
-		if docHistoryEntry.Name != chanName {
+		if docHistoryEntry.Name != chanName && "*" != chanName {
```
bbrks: I think this could be the `channels.UserStarChannel` const

bbrks: To save evaluating `"*" != chanName` for each iteration it might be worth doing something like this:

```go
isStarChan = chanName == channels.UserStarChannel
for _, ... range ... {
    if !isStarChan && docHistoryEntry.Name != chanName {
```

(the ordering of the statements in the if is also significant, as it gives an early exit opportunity without evaluating `docHistoryEntry.Name != chanName` each loop

bbrks (changes requested):
Minor comment about const usage and micro-optimisation.

db/changes.go
mohammed-madi: Addressed in [43121ce](https://github.com/couchbase/sync_gateway/pull/5961/commits/43121ce71726f4b6e6f8e21b059c40fbac8ebf1b)

mohammed-madi: Addressed in [43121ce](https://github.com/couchbase/sync_gateway/pull/5961/commits/43121ce71726f4b6e6f8e21b059c40fbac8ebf1b)

bbrks (approved):
LGTM - @adamcfraser can you review the approach?

adamcfraser:
It looks like the change only affects documents that have been mutated post-revocation (i.e. where since < doc.seq) - I would have expected us to also need to update the handling for older documents, based on the ticket description.  Need to review the behaviour of the unit test pre-fix to understand why this is sufficient.

adamcfraser (approved):

PR #5966, opened 2022-12-12
CBG-2275 support collection specific import filters

torcolvin:
Use collection specific import filter. This will use the database level import filter if a collection level is not specified.

I wonder if I should have written these tests using the SDK and `db` package rather than in the `rest` package, but it was easier to follow the pattern of changes tests.

This also includes https://issues.couchbase.com/browse/CBG-2575.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1232/


db/database.go
```diff
@@ -178,8 +178,8 @@ type ScopeOptions struct {
 }
 
 type CollectionOptions struct {
-	Sync *string // Collection sync function
-
+	Sync          *string        // Collection sync function
+	ImportOptions *ImportOptions // Collection import filter and options
```
adamcfraser: I think it makes sense to just define this as the import filter (*string), since ImportOptions.ImportPartitions and BackupOldRev are database-wide settings.

adamcfraser (dismissed):

db/database.go
torcolvin: Fixed. It's the compiled string here I think.

bbrks (approved):
LGTM

PR #5971, opened 2022-12-13
GraphQL and Functions, stage 3

snej:
Updates GraphQL and JS-functions support to version 0.3 of [the spec](https://docs.google.com/document/d/1ggxUJhotM0-Ww8CnM69HKBcKbboQyw5VeWLUkdp8sUA/edit#heading=h.jlb29iydw5r0).

- Functions now support a boolean ‘mutating’ property.  At runtime, mutations via ‘context.user’ are only allowed if ‘mutating’ is determined to be true.
- GraphQL resolvers now enforce mutability across Queries and Mutations.  At runtime, mutations via ‘context.user’ are only allowed if calling from w/in a resolver chain that originated from a Mutation.
- Change save(id, doc) to save(doc, id). id is optional. If id is null and doc contains `_id`, use that as the id. if id and `_id` are null, use a UUID for the id. If id=null and `_id`=null and `_rev`!=null, then match the REST service behavior. If id!=null and `_id`!=null and id!=_id, match the REST service behavior.
- Add `delete(doc: object): boolean`.
- Reorder graphql resolver args to be inline w/ standard: `function(parent, args, context, info)`
- Change variable-expansion syntax in Allows properties to `${name}` (was `$name` or `$(name)`)

Note: The resource-limits features in v0.3 were already implemented by the previous PR that's already been merged.

Also adds a number of new unit tests written by new hires, which I've already reviewed (e.g. all the stuff in `rest/`.)

## Pre-review checklist

- [x] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1431/


snej:
The windows-latest CI failure looks unrelated; something to do with replication:
```
2022-12-14T23:04:20.4351063Z     replicator_test.go:392: 
2022-12-14T23:04:20.4351311Z         	Error Trace:	replicator_test.go:392
2022-12-14T23:04:20.4351583Z         	Error:      	Received unexpected error:
2022-12-14T23:04:20.4352208Z         	            	RetryLoop for Wait for condition options giving up after 201 attempts
2022-12-14T23:04:20.4352435Z         	Test:       	TestAdhocReplicationStatus
...
2022-12-14T23:04:20.4355644Z --- FAIL: TestAdhocReplicationStatus (21.88s)
```

db/functions/function.go
```diff
@@ -164,6 +168,17 @@ func (fn *functionImpl) Invoke(database *db.Database, args map[string]any, mutat
 		}
 	}
 
+	if (!fn.Mutating || !mutationAllowed) && ctx.Value(readOnlyKey) == nil {
```
adamcfraser, line 171: Bit of a nitpick, but I feel like this code would be easier to read as:
```
//  If context isn't already marked read-only, check whether it should be
if ctx.Value(readOnlyKey) == nil {
   if !fn.Mutating {
      ctx = context.WithValue(ctx, readOnlyKey, fmt.Sprintf("%s %q", fn.typeName, fn.name))
   } else if !mutationAllowed {
      ctx = context.WithValue(ctx, readOnlyKey, "a read-only API call")
   }
 }

File: db/functions/function.go [hunk 2]
```diff
@@ -208,11 +222,45 @@ func (fn *functionImpl) checkArguments(args map[string]any) error {
...
+				if err != nil && err.Status >= http.StatusInternalServerError {
+					return err
+				}
+			}
+			return nil
+		}
+		// Test each role and channel string:
+		for _, str := range fn.Allow.Roles {
+			if err := checkPattern(str); err != nil {
+				return err
```
adamcfraser, line 247: Does it make sense for this error to include the Role/Channel name that hit the error?  Or is that obvious from the error message returned by expandPattern?

db/functions/function.go
```diff
@@ -243,57 +291,128 @@ func (fn *functionImpl) authorize(user auth.User, args map[string]any) error {
...
+			// Convert `rval` to a string:
+			for rval.Kind() == reflect.Interface {
+				rval = rval.Elem()
+			}
+			if rval.Kind() == reflect.String {
+				return rval.String()
+			} else if rval.CanInt() || rval.CanUint() || rval.CanFloat() {
+				return fmt.Sprintf("%v", rval)
+			} else {
+				err = base.HTTPErrorf(http.StatusBadRequest, "argument %q must be a string or number, not %v", arg, rval.Kind())
```
adamcfraser: I'm not sure why we don't set badConfig=true here - is this a recoverable error?

db/functions/function.go
```diff
@@ -243,57 +291,128 @@ func (fn *functionImpl) authorize(user auth.User, args map[string]any) error {
...
+
+			// Convert `rval` to a string:
+			for rval.Kind() == reflect.Interface {
+				rval = rval.Elem()
+			}
+			if rval.Kind() == reflect.String {
+				return rval.String()
+			} else if rval.CanInt() || rval.CanUint() || rval.CanFloat() {
+				return fmt.Sprintf("%v", rval)
+			} else {
```
adamcfraser, line 330: Would it make sense to handle reflect.Bool here as well?

db/functions/js_callbacks.go
```diff
@@ -23,9 +24,16 @@ import (
...
 	// Implementation of the 'delete(docID)' callback:
 	runner.DefineNativeFunction("_delete", func(call otto.FunctionCall) otto.Value {
-		docID := ottoStringParam(call, 0, "user.delete")
-		doc := ottoObjectParam(call, 1, true, "user.delete")
-		sudo := ottoBoolParam(call, 2)
+		var docID string
+		var doc map[string]any
+		if arg0 := call.Argument(0); arg0.IsString() {
+			docID = arg0.String()
+		} else if arg0.IsObject() {
```
adamcfraser, line 31: What's the object handling here?  Is this delete(doc), with the id being retrieved from doc._id?

rest/functionsapitest/main_test.go
```diff
@@ -0,0 +1,22 @@
...
+package functionsapitest
+
+import (
+	"testing"
+
+	"github.com/couchbase/sync_gateway/db"
+)
+
+func TestMain(m *testing.M) {
+	memWatermarkThresholdMB := uint64(18192)
```
adamcfraser, line 20: Is this a typo or intentional?  Most other packages manage with at most 8192 as the watermark.

adamcfraser (changes requested):
Generally looks fine, a few minor nits/questions.

Should also run integration tests once conflicts have been resolved.

db/functions/function.go
snej: The calling function (compileFunction) amends the message to include the function name, which so far has been good enough (for my & Wayne's use.) But we can of course improve errors that we find people are getting confused by.

db/functions/function.go
snej: Well, my rule of thumb is to do cheap tests (`fn.Mutating`, `mutationAllowed`) before expensive ones (`ctx.Value()`). Do you think the readability is worth changing that?

db/functions/function.go
snej: It's an error that happens at runtime when a client calls the API with an arg of the wrong type; not an error in the configuration.

The `badConfig` variable here is weird -- setting it doesn't really have any effect. I think it's a leftover from an earlier version of that function. I'll take it out.

db/functions/function.go
snej: Good idea. I'll add that.

db/functions/js_callbacks.go
snej: We allow the parameter to be either the docID string, or the document object. In the latter case we get the ID from the `_id` property. This function is just unpacking the Otto arguments before calling `runner.do_delete()`, so it sets one of two variables (`docID` or `doc`) and passes them both on.

rest/functionsapitest/main_test.go
snej: Huh, I'm not sure because I didn't write that -- it was one of the new hires in India. It sure could be a typo. 

I don't even know what that parameter controls; is it a memory limit?

adamcfraser:
Changes all look good, kicked off integration tests (https://jenkins.sgwdev.com/job/SyncGateway-Integration/1431/), will approve/merge once those run clean.

adamcfraser (approved):

PR #5974, opened 2022-12-14
do not create allDocs index when starChannel is disabled

nirav24:
CBG-0000

Describe your PR here...
- When starChannel is not enable, we do not create allDocs index anymore
-  Returns error message when _all_docs endpoint is called but allDocs index doesn't exist because starChannel was disabled

## Pre-review checklist

- [ ] Link upstream PRs

### DIFF
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -39,7 +39,7 @@ var SkippedSeqCleanViewBatch = 50 // Max number of sequences checked per query d
 
 // Enable keeping a channel-log for the "*" channel (channel.UserStarChannel). The only time this channel is needed is if
 // someone has access to "*" (e.g. admin-party) and tracks its changes feed.
-var EnableStarChannelLog = false
+var EnableStarChannelLog = true
 
 // Manages a cache of the recent change history of all channels.
 //
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -155,7 +155,7 @@ func DefaultCacheOptions() CacheOptions {
 // Initializes a new changeCache.
 // lastSequence is the last known database sequence assigned.
 // notifyChange is an optional function that will be called to notify of channel changes.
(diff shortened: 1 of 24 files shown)


db/change_cache.go
```diff
@@ -41,6 +41,9 @@ var SkippedSeqCleanViewBatch = 50 // Max number of sequences checked per query d
 // someone has access to "*" (e.g. admin-party) and tracks its changes feed.
 var EnableStarChannelLog = true
 
+// IsAllDocsIndexCreated keeps track of AllDocs index existence.
+var IsAllDocsIndexCreated = false
```
adamcfraser: Will need to be a per-database property.

db/indexes.go
```diff
@@ -293,6 +298,15 @@ func (i *SGIndex) createIfNeeded(bucket base.N1QLStore, useXattrs bool, numRepli
 		return false, nil
 	}
 
+	// only create allDocs Index if EnableStarChannel is set to true
+	if i.simpleName == sgIndexes[IndexAllDocs].simpleName {
+		if !EnableStarChannelLog {
```
adamcfraser: I see the benefit of doing this check here (can set IsAllDocsIndexCreated), but calling createIfNeeded when we don't actually want to create the index feels counter-intuitive.  I think it would be preferable to have separate functionality to perform the check if the index exists for indexes we don't need to create.

rest/bulk_api.go
```diff
@@ -27,6 +27,11 @@ import (
 
 // HTTP handler for _all_docs
 func (h *handler) handleAllDocs() error {
+
+	if !db.IsAllDocsIndexCreated {
```
adamcfraser: May want to just base this on enableStarChannel, to avoid scenarios where the index gets deleted by user after we've done the startup check for existence.

adamcfraser (changes requested):

PR #5977, opened 2022-12-16
CBG-2622 do not allow /db to make to single default collection, force using keyspace

torcolvin:
To facilitate this change:

- Change `NewRestTester` to return a keyspace, and use the keyspace in the tests to allow them to in a test environment. Waffled a bit on whether to return keyspace from `NewRester` as a convenience method. The convenience of having the keyspace comes at a cost, which is that you used to be able to modify `RestTester` configuration after you called `NewRestTester` but before you actually used it.
- As a consequence of above, occasionally moved the `RestTesterConfig` into `NewRestTester` constructor.
- It doesn't seem to make sense to return keyspaces in persistent config mode, because usually (but not always?) there are no databases yet.

Currently evaluating whether Jenkins is having the flakes or real issues.
## Pre-review checklist

- [ ] `GSI=true,xattrs=true,default_collection=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1280/
- [x] `GSI=true,xattrs=true,default_collection=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1276/
- [x] `GSI=true,xattrs=false,default_collection=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1275/

### DIFF
--- a/rest/access_test.go
+++ b/rest/access_test.go
@@ -276,7 +276,7 @@ func TestNumAccessErrors(t *testing.T) {
 		SyncFn: `function(doc, oldDoc){if (doc.channels.indexOf("foo") > -1){requireRole("foobar")}}`,
 	}
 
-	rt := NewRestTester(t, &rtConfig)
+	rt, keyspace := NewRestTester(t, &rtConfig)
 	defer rt.Close()
 
 	ctx := rt.Context()
@@ -287,13 +287,13 @@ func TestNumAccessErrors(t *testing.T) {
 	assert.NoError(t, err)
 	assert.NoError(t, a.Save(user))
 
-	response := rt.Send(RequestByUser("PUT", "/db/doc", `{"prop":true, "channels":["foo"]}`, "user"))
+	response := rt.Send(RequestByUser("PUT", fmt.Sprintf("/%s/doc", keyspace), `{"prop":true, "channels":["foo"]}`, "user"))
 	RequireStatus(t, response, 403)
 
 	base.WaitForStat(func() int64 { return rt.GetDatabase().DbStats.SecurityStats.NumAccessErrors.Value() }, 1)
 }
 func TestUserHasDocAccessDocNotFound(t *testing.T) {
-	rt := NewRestTester(t, &RestTesterConfig{
+	rt, keyspace := NewRestTester(t, &RestTesterConfig{
 		DatabaseConfig: &DatabaseConfig{DbConfig: DbConfig{
 			QueryPaginationLimit: base.IntPtr(2),
 			CacheConfig: &CacheConfig{
@@ -309,7 +309,7 @@ func TestUserHasDocAccessDocNotFound(t *testing.T) {
 	defer rt.Close()
 	ctx := rt.Context()
 
-	resp := rt.SendAdminRequest("PUT", "/db/doc", `{"channels": ["A"]}`)
+	resp := rt.SendAdminRequest("PUT", fmt.Sprintf("/%s/doc", keyspace), `{"channels": ["A"]}`)
 	RequireStatus(t, resp, http.StatusCreated)
 
 	database, err := db.CreateDatabase(rt.GetDatabase())
@@ -986,7 +986,7 @@ func TestAccessOnTombstone(t *testing.T) {
 			 }
 			 channel(doc.channel)
 		 }`}
-	rt := NewRestTester(t, &rtConfig)
+	rt, keyspace := NewRestTester(t, &rtConfig)
 	defer rt.Close()
 
 	ctx := rt.Context()
@@ -1002,7 +1002,7 @@ func TestAccessOnTombstone(t *testing.T) {
 	assert.NoError(t, a.Save(bernard))
 
 	// Create doc that gives user access to its channel
-	response := rt.SendAdminRequest("PUT", "/db/alpha", `{"owner":"bernard", "channel":"PBS"}`)
+	response := rt.SendAdminRequest("PUT", fmt.Sprintf("/%s/alpha", keyspace), `{"owner":"bernard", "channel":"PBS"}`)
 	RequireStatus(t, response, 201)
 	var body db.Body
 	assert.NoError(t, base.JSONUnmarshal(response.Body.Bytes(), &body))
@@ -1016,7 +1016,7 @@ func TestAccessOnTombstone(t *testing.T) {
 	var changes struct {
 		Results []db.ChangeEntry
 	}
-	response = rt.Send(RequestByUser("GET", "/db/_changes", "", "bernard"))
+	response = rt.Send(RequestByUser("GET", fmt.Sprintf("/%s/_changes", keyspace), "", "bernard"))
 	log.Printf("_changes looks like: %s", response.Body.Bytes())
 	err = base.JSONUnmarshal(response.Body.Bytes(), &changes)
 	assert.NoError(t, err)
@@ -1026,11 +1026,11 @@ func TestAccessOnTombstone(t *testing.T) {
 	}
 
 	// Delete the document
-	response = rt.SendAdminRequest("DELETE", fmt.Sprintf("/db/alpha?rev=%s", revId), "")
+	response = rt.SendAdminRequest("DELETE", fmt.Sprintf("/%s/alpha?rev=%s", keyspace, revId), "")
 	RequireStatus(t, response, 200)
 
 	// Make sure it actually was deleted
-	response = rt.SendAdminRequest("GET", "/db/alpha", "")
+	response = rt.SendAdminRequest("GET", fmt.Sprintf("/%s/alpha", keyspace), "")
 	RequireStatus(t, response, 404)
 
 	// Wait for change caching to complete
@@ -1038,7 +1038,7 @@ func TestAccessOnTombstone(t *testing.T) {
 
 	// Check user access again:
 	changes.Results = nil
-	response = rt.Send(RequestByUser("GET", "/db/_changes", "", "bernard"))
+	response = rt.Send(RequestByUser("GET", fmt.Sprintf("/%s/_changes", keyspace), "", "bernard"))
 	log.Printf("_changes looks like: %s", response.Body.Bytes())
 	assert.NoError(t, base.JSONUnmarshal(response.Body.Bytes(), &changes))
 	require.Len(t, changes.Results, 1)
@@ -1054,7 +1054,7 @@ func TestDynamicChannelGrant(t *testing.T) {
 	base.SetUpTestLogging(t, base.LevelInfo, base.KeyAccess)
 
 	rtConfig := RestTesterConfig{SyncFn: `function(doc) {if(doc.type == "setaccess") {channel(doc.channel); access(doc.owner, doc.channel);} else { channel(doc.channel)}}`}
-	rt := NewRestTester(t, &rtConfig)
+	rt, keyspace := NewRestTester(t, &rtConfig)
 	defer rt.Close()
 
 	ctx := rt.Context()
@@ -1069,9 +1069,6 @@ func TestDynamicChannelGrant(t *testing.T) {
 	user, err = a.NewUser("user1", "letmein", nil)
 	require.NoError(t, err)
 	require.NoError(t, a.Save(user))
-
-	collection := rt.GetDatabase().GetSingleDatabaseCollection()
-	keyspace := fmt.Sprintf("%s.%s.%s", "db", collection.ScopeName(), collection.Name())
 
 	// Create a document in channel chan1
 	response := rt.Send(RequestByUser("PUT", "/"+keyspace+"/doc1", `{"channel":"chan1", "greeting":"hello"}`, "user1"))
@@ -1114,7 +1111,7 @@ func TestRoleChannelGrantInheritance(t *testing.T) {
 	base.SetUpTestLogging(t, base.LevelInfo, base.KeyAccess)
 
 	rtConfig := RestTesterConfig{SyncFn: `function(doc) {if(doc.type == "setaccess") {channel(doc.channel); access(doc.owner, doc.channel);} else { channel(doc.channel)}}`}
-	rt := NewRestTester(t, &rtConfig)
+	rt, keyspace := NewRestTester(t, &rtConfig)
 	defer rt.Close()
 
 	ctx := rt.Context()
@@ -1123,7 +1120,6 @@ func TestRoleChannelGrantInheritance(t *testing.T) {
 	collection := rt.GetDatabase().GetSingleDatabaseCollection()
 	scopeName := collection.ScopeName()
 	collectionName := collection.Name()
-	keyspace := fmt.Sprintf("%s.%s.%s", "db", scopeName, collectionName)
 
 	user, err := a.GetUser("")
 	assert.NoError(t, err)
@@ -1212,7 +1208,7 @@ func TestPublicChannel(t *testing.T) {
                  channel(doc.channel)
               }
            }`}
-	rt := NewRestTester(t, &rtConfig)
+	rt, keyspace := NewRestTester(t, &rtConfig)
 	defer rt.Close()
 
 	ctx := rt.Context()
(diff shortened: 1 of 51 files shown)


rest/handler.go
```diff
@@ -440,11 +442,7 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
 					// _default doesn't exist for a non-default scope - so make it a required element if it's ambiguous
 					return base.HTTPErrorf(http.StatusBadRequest, "Ambiguous keyspace: %s.%s", base.MD(keyspaceDb), base.MD(base.StringDefault(keyspaceScope, "")))
 				}
-				// one collection only - use that one
-				for collectionName := range scope.Collections {
-					keyspaceCollection = &collectionName
-					break
-				}
+				keyspaceScope = base.StringPtr(base.DefaultCollection)
```
bbrks: I think this was supposed to be collection?

```suggestion
				keyspaceCollection = base.StringPtr(base.DefaultCollection)
```

bbrks (dismissed):
not a full review but spotted this

rest/access_test.go
```diff
@@ -111,20 +111,20 @@ func TestStarAccess(t *testing.T) {
 	assert.NoError(t, a.Save(bernard))
 
 	// GET /db/docid - basic test for channel user has
-	response := rt.Send(RequestByUser("GET", "/db/doc1", "", "bernard"))
+	response := rt.SendUserRequest("GET", "/db/doc1", "", "bernard")
```
adamcfraser, line 114: Does `/db/doc1` resolve the same way as `/{{.db}}/doc1`, or `/{{.keyspace}}/doc1`?  (and is it a planned task to switch over remaining usages of `/db/...`?)

rest/admin_api_auth_test.go
```diff
@@ -496,120 +496,127 @@ func TestAdminAPIAuth(t *testing.T) {
...
+	const (
+		noPrefix = iota
+		dbPrefix
+		collectionPrefix
+	)
+
 	endPoints := []struct {
 		Method          string
-		DBScoped        bool
+		urlPrefix       urlPrefixType
```
adamcfraser: I don't think urlPrefix is as descriptive as 'DBScoped' in describing the test cases.  I'm again cursing server for choosing the name 'scope' for a group of collections, as I think 'RequestScope' with options 'None/Db/Collection' would have been ideal here.  

Maybe a better alternative is just to drop DBScoped altogether, and include {{.db}} or {{.keyspace}} in endpoint. Endpoint?

rest/adminapitest/admin_api_test.go
```diff
@@ -2113,17 +2118,17 @@ func TestSwitchDbConfigCollectionName(t *testing.T) {
...
+		dataStore1Name.CollectionName())
 	resp := rest.BootstrapAdminRequest(t, http.MethodPut, "/db/", fmt.Sprintf(
 		`{"bucket": "%s", "scopes": {"%s": {"collections": {"%s": {}}}}, "num_index_replicas": 0, "enable_shared_bucket_access": true, "use_views": false}`,
 		tb.GetName(), dataStore1Name.ScopeName(), dataStore1Name.CollectionName(),
 	))
 	resp.RequireStatus(http.StatusCreated)
 
 	// put a doc in db
-	resp = rest.BootstrapAdminRequest(t, http.MethodPut, fmt.Sprintf("/%s/10001", keyspace1), `{"type":"test_doc"}`)
+	resp = rest.BootstrapAdminRequest(t, http.MethodPut, "/"+keyspace1+"/10001", `{"type":"test_doc"}`)
```
adamcfraser, line 2131: Is this `{{.keyspace1}}`?

rest/adminapitest/admin_api_test.go
```diff
@@ -2134,7 +2139,7 @@ func TestSwitchDbConfigCollectionName(t *testing.T) {
 	resp.RequireStatus(http.StatusCreated)
 
 	// put doc in new collection
-	resp = rest.BootstrapAdminRequest(t, http.MethodPut, fmt.Sprintf("/%s/10001", keyspace2), `{"type":"test_doc1"}`)
+	resp = rest.BootstrapAdminRequest(t, http.MethodPut, fmt.Sprintf("/db.%s.%s/10001", dataStore2Name.ScopeName(), dataStore2Name.CollectionName()), `{"type":"test_doc1"}`)
```
adamcfraser, line 2142: Is this `{{.keyspace2}}`?

rest/api_benchmark_test.go
```diff
@@ -143,7 +143,7 @@ func BenchmarkReadOps_GetRevCacheMisses(b *testing.B) {
 
 	// Create user
 	username := "user1"
-	rt.SendAdminRequest("PUT", fmt.Sprintf("/db/_user/%s", username), fmt.Sprintf(`{"name":"%s", "password":"letmein", "admin_channels":["channel_1"]}`, username))
+	rt.SendAdminRequest("PUT", "/{{.keyspace}}/_user/"+username, fmt.Sprintf(`{"name":"%s", "password":"letmein", "admin_channels":["channel_1"]}`, username))
```
adamcfraser: I think this should be `{{.db}}`?

rest/api_test.go
```diff
@@ -112,7 +112,7 @@ func TestDisablePublicBasicAuth(t *testing.T) {
 	assert.NoError(t, err)
 	assert.NoError(t, a.Save(user))
 
-	response = rt.Send(RequestByUser(http.MethodGet, "/db/", "", "user1"))
+	response = rt.SendUserRequest(http.MethodGet, "/db/", "", "user1")
```
adamcfraser: Can this be `{{.db}}`?

rest/api_test.go
```diff
@@ -122,7 +122,7 @@ func TestDisablePublicBasicAuth(t *testing.T) {
 
 	// As a sanity check, ensure it does work when the setting is disabled
 	rt.ServerContext().Database(ctx, "db").Options.DisablePasswordAuthentication = false
-	response = rt.Send(RequestByUser(http.MethodGet, "/db/", "", "user1"))
+	response = rt.SendUserRequest(http.MethodGet, "/db/", "", "user1")
```
adamcfraser: Can this be `{{.db}}`?

rest/adminapitest/collections_admin_api_test.go
```diff
@@ -75,12 +75,6 @@ func TestCollectionsSyncImportFunctions(t *testing.T) {
 	resp := rt.SendAdminRequest(http.MethodPut, "/db/", bucketConfig)
 	rest.RequireStatus(t, resp, http.StatusCreated)
 
-	resp = rt.SendAdminRequest(http.MethodGet, "/db/_config/import_filter", "")
-	rest.RequireStatus(t, resp, http.StatusBadRequest)
-
-	resp = rt.SendAdminRequest(http.MethodGet, "/db/_config/sync", "")
-	rest.RequireStatus(t, resp, http.StatusBadRequest)
-
```
adamcfraser, line 83: I'm not sure why these are being removed.

rest/blip_api_delta_sync_test.go
```diff
@@ -118,7 +118,7 @@ func TestBlipDeltaSyncPushPullNewAttachment(t *testing.T) {
 
 	// Create doc1 rev 1-77d9041e49931ceef58a1eef5fd032e8 on SG with an attachment
 	bodyText := `{"greetings":[{"hi": "alice"}],"_attachments":{"hello.txt":{"data":"aGVsbG8gd29ybGQ="}}}`
-	response := rt.SendAdminRequest(http.MethodPut, "/db/"+docID, bodyText)
+	response := rt.SendAdminRequest(http.MethodPut, "/{{.keyspace}}/"+docID, bodyText)
```
adamcfraser, line 121: I'm not clear why some blip tests are restricted to default collection (using NewRestTesterDefaultCollection and marked with CBG-2619 in this PR), but others like this one work as-is.  What's the difference?

rest/functions_api_test.go
```diff
@@ -92,13 +93,12 @@ func testConcurrently(t *testing.T, rt *RestTester, testFunc func() bool) bool {
...
 func TestFunctions(t *testing.T) {
-	rt := NewRestTester(t, &RestTesterConfig{GuestEnabled: true, EnableUserQueries: true})
+	rt := NewRestTester(t, &RestTesterConfig{GuestEnabled: true, EnableUserQueries: true, DatabaseConfig: kGraphQLTestConfig})
 	defer rt.Close()
-	rt.DatabaseConfig = kGraphQLTestConfig
 
 	t.Run("GraphQL with variables", func(t *testing.T) {
 		testConcurrently(t, rt, func() bool {
-			response := rt.SendRequest("POST", "/db/_graphql",
+			response := rt.SendRequest("POST", fmt.Sprintf("/%s/_graphql", rt.GetDatabase().Name),
```
adamcfraser: Can this be `{{.db}}` (and other usages in functions_api_test.go)?

rest/handler.go
```diff
@@ -423,11 +423,13 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
...
-				for scopeName := range dbContext.Scopes {
-					keyspaceScope = &scopeName
-					break
+				if len(dbContext.Scopes) == 1 {
+					for scopeName, _ := range dbContext.Scopes {
+						keyspaceScope = base.StringPtr(scopeName)
+					}
+
+				} else {
+					return base.HTTPErrorf(http.StatusBadRequest, "Ambiguous keyspace: %s.%s", base.MD(keyspaceDb), base.MD(base.StringDefault(keyspaceScope, "")))
```
adamcfraser, line 432: It looks like the previous code would return 'keyspace not found' instead of 'ambiguous keyspace' if dbContext.Scopes was non-nil but length zero.  Do you think ambiguous is more correct for this case?

rest/handler.go
```diff
@@ -440,11 +442,7 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
 					// _default doesn't exist for a non-default scope - so make it a required element if it's ambiguous
 					return base.HTTPErrorf(http.StatusBadRequest, "Ambiguous keyspace: %s.%s", base.MD(keyspaceDb), base.MD(base.StringDefault(keyspaceScope, "")))
 				}
-				// one collection only - use that one
-				for collectionName := range scope.Collections {
-					keyspaceCollection = &collectionName
-					break
-				}
+				keyspaceCollection = base.StringPtr(base.DefaultCollection)
```
adamcfraser, line 445: There isn't a default collection on a non-default scope, and I don't see where we're setting keyspaceScope to base.DefaultScope in this path.

rest/handler_test.go
```diff
@@ -104,28 +109,28 @@ func TestHTTPLoggingRedaction(t *testing.T) {
 		{
 			name:        "raw-docid",
 			method:      http.MethodGet,
-			path:        "/db/_raw/test",
-			expectedLog: "/db/_raw/<ud>test</ud>",
+			path:        fmt.Sprintf("/%s/_raw/test", keyspace),
```
adamcfraser, line 112: Can we use {{.keyspace}} here?

rest/oidc_api_test.go
```diff
@@ -416,7 +416,7 @@ func TestGetOIDCCallbackURL(t *testing.T) {
 			providers := auth.OIDCProviderMap{"foo": mockProvider("foo"), "bar": mockProvider("bar")}
 			openIDConnectOptions := auth.OIDCOptions{Providers: providers, DefaultProvider: base.StringPtr("foo")}
 			rtConfig := RestTesterConfig{DatabaseConfig: &DatabaseConfig{DbConfig: DbConfig{OIDCConfig: &openIDConnectOptions}}}
-			rt := NewRestTester(t, &rtConfig)
+			rt := NewRestTesterDefaultCollection(t, &rtConfig)
```
adamcfraser: What pending ticket requires this to be default only?

rest/utilities_testing_resttester.go
```diff
@@ -30,14 +30,14 @@ type PutDocResponse struct {
 }
 
 func (rt *RestTester) GetDoc(docID string) (body db.Body) {
-	rawResponse := rt.SendAdminRequest("GET", "/db/"+docID, "")
+	rawResponse := rt.SendAdminRequest("GET", fmt.Sprintf("/%s/%s", rt.GetSingleKeyspace(), docID), "")
```
adamcfraser: Can we use {{.keyspace}} here?

rest/utilities_testing_attachment.go
```diff
@@ -22,7 +23,7 @@ import (
 func (rt *RestTester) WaitForAttachmentCompactionStatus(t *testing.T, state db.BackgroundProcessState) db.AttachmentManagerResponse {
 	var response db.AttachmentManagerResponse
 	err := rt.WaitForConditionWithOptions(func() bool {
-		resp := rt.SendAdminRequest("GET", "/db/_compact?type=attachment", "")
+		resp := rt.SendAdminRequest("GET", fmt.Sprintf("/%s/_compact?type=attachment", rt.GetSingleKeyspace()), "")
```
adamcfraser: Can we use {{.keyspace}} here?

adamcfraser (changes requested):
Overall approach is great, some questions on specific test handling.

rest/access_test.go
torcolvin: I didn't switch any tests by eye and since this test uses the default collection anyway until https://github.com/couchbase/sync_gateway/pull/5990 lands, I was going to let the person who fixes the test. "/db" is correct here because the keyspace is `db._default._default`

rest/adminapitest/admin_api_test.go
torcolvin: This isn't, because this test doesn't use the rest tester. I thought about allowing publically or switching this test to use the rest tester, but it was starting to creep up in scope.

rest/adminapitest/admin_api_test.go
torcolvin: See above, because this isn't using `RestTester`

rest/admin_api_auth_test.go
torcolvin: done.

rest/api_test.go
torcolvin: done.

rest/api_benchmark_test.go
torcolvin: done, although I left this for https://github.com/couchbase/sync_gateway/pull/5990 to fix the channel access.

rest/adminapitest/collections_admin_api_test.go
torcolvin: There's no longer a /db/ becomes ambiguous in a multiple collection context. This ends up returning alternately 404 or 400, I am going to fix this in a followup ticket.

rest/blip_api_delta_sync_test.go
torcolvin: As long as the blip message contains a `Collection` property name the tests work. In some tests, we are constructing blip messages in a way that forget the `Collection` property.

rest/functions_api_test.go
torcolvin: done.

rest/handler_test.go
torcolvin: yes for path but not for expectedLog

rest/adminapitest/collections_admin_api_test.go
torcolvin: Furthermore, this is confusing in the case where we are using the default collection because it is actually a different case than the non default collection. I didn't think it was worth testing this use case.

rest/handler.go
torcolvin: https://issues.couchbase.com/browse/CBG-2654 for fixing this.

rest/handler.go
torcolvin: I'm going to catch this in https://issues.couchbase.com/browse/CBG-2654

rest/adminapitest/admin_api_test.go
adamcfraser: I agree that we should go back and enhance BootstrapAdminRequest to do template substitution in a future patch.

rest/adminapitest/collections_admin_api_test.go
adamcfraser: ah ok, I hadn't noticed that we already have keyspace1/keyspace2 tests for these endpoints just below.  I think this makes sense.

rest/oidc_api_test.go
torcolvin: This was unnecessary, removed.

adamcfraser (approved):

PR #5980, opened 2022-12-16
Use V8 instead of Otto for JavaScript

snej:
CBG-2949
Replaces Otto with V8 for all uses of JavaScript.

The new `js` package is a general purpose thread-safe library for creating "services" that call a JavaScript function in an isolated runtime context.

## Pre-review checklist

- [x] Link upstream PRs -- https://github.com/couchbase/sg-bucket/pull/91


snej:
On Ubuntu, the only test failure is `TestReplicatorRevocationsMultipleAlternateAccess`, which gets several unexpected errors of the form `RetryLoop for Wait for condition options giving up after 201 attempts`. No idea what this is about.

On Windows the build fails with `ld.exe: cannot find -lv8`. This is a known issue with v8go: its CI doesn't currently provide a prebuilt Windows binary of V8. We're going to need to build our own.

snej:
**ci / test-race** is hitting JS fn timeouts during `TestGraphQLConcurrently`. Partly that's because I left JS logging at debug level so there are zillions of log lines (oops) but also maybe the race detector slows down concurrent code a lot? I'll turn down the logging and increase the timeouts.

**continuous-integration/jenkins/pr-head** is failing with `go build github.com/snej/v8go: g++: exec: "g++": executable file not found in $PATH` ... is the builder not configured with a C++ compiler? It'll need one to build the v8go package.

bbrks:
@snej No our CI images don't have any C/C++ toolchain set up. We've never needed it up until this point.

We try to install/manage our Go build dependency via Jenkins so that we're able to run the job on any runner and not presume what's installed, but it looks like it's not that feasible for g++/clang so I'm not sure what to do given it's going to be a set of platform/distro specific install steps.

snej:
@bbrks Oh! I didn't realize none of our package dependencies have used Cgo until now.

bbrks:
We've avoided it where possible (without too much effort), but it seems unavoidable given the v8 work

Hopefully a case of just adding some more steps to the worker node setup script inside Jenkins, and crossing fingers that Amazon Linux has a usable version, but we'll need to find a bit of time after collections is wrapped to get this done.

snej:
**ci/addlicense** has three remaining nags:
```
js/underscore-umd-min.js
js/underscore-umd.js
db/functions/engine/dist/main.js
```
- The first two are a 3rd party JS library (included in Otto, so we need to keep including it for backward compatibility if nothing else.) We can't put our license on these. Should they go someplace else, like a `vendor` dir?
- `main.js` is the compiled form of the TypeScript source code in db/functions/engine/src. It's checked in as well to avoid making the TypeScript compiler (and node.js) a required part of the build. I can add something to its build script to prefix the license comment to it.

snej:
I've added the license comment to `main.js`. It was surprisingly difficult to get WebPack to do this.

I don't know a good way to deal with the `underscore.js` files, except by adding `-ignore 'js/underscore*.js'` to the `addlicense` command. (I tried this locally and it worked.)

snej:
I've filed a [Jira ticket](https://issues.couchbase.com/browse/CBG-2741) covering the build changes needed for SG to use v8go.

channels/channelmapper.go
```diff
@@ -60,79 +113,149 @@ func GetDefaultSyncFunction(scopeName, collectionName string) string {
...
-func (mapper *ChannelMapper) MapToChannelsAndAccess(body map[string]interface{}, oldBodyJSON string, metaMap map[string]interface{}, userCtx map[string]interface{}) (*ChannelMapperOutput, error) {
-	numberFixBody := ConvertJSONNumbers(body)
-	numberFixMetaMap := ConvertJSONNumbers(metaMap)
+func (mapper *ChannelMapper) Function() string {
+	return mapper.fnSource
+}
+
+// This function is DEPRECATED. It's currently used in some tests that can't easily be changed.
+// Its current implementation is a kludge, and it shouldn't be used in production.
+func (mapper *ChannelMapper) SetFunction(fnSource string) error {
```
torcolvin, line 122: This is called from `NewDatabaseContext` and is used as one of the primary places in production.

db/functions/engine/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+1. `cd db/functions/engine`
+2. `npm install`
+
+### Building It
+
+On a clean build, or after any changes in the `engine` source tree:
+
+1. `cd db/functions/engine`
+2. `npm run build`
```
torcolvin, line 32: Is it worth writing some CI to make sure that we run the tests here and that if running npm run build changes, it fails if we don't update the file?

channels/channelmapper.go
```diff
@@ -60,79 +113,149 @@ func GetDefaultSyncFunction(scopeName, collectionName string) string {
...
+	host := mapper.service.Host()
+	mapper.fnSource = fnSource
+	mapper.service = js.NewService(host, "channelMapper", wrappedFuncSource(fnSource, host))
+	return nil
+}
+
+func wrappedFuncSource(funcSource string, host js.ServiceHost) string {
+	// Use different wrapper scripts for modern vs older JS.
```
torcolvin, line 130: Is the difference here going to be otto vs v8, or is this within versions of v8?

db/database_test.go
```diff
@@ -28,14 +28,9 @@ import (
 	"github.com/couchbase/sync_gateway/auth"
 	"github.com/couchbase/sync_gateway/base"
 	"github.com/couchbase/sync_gateway/channels"
-	"github.com/robertkrimen/otto/underscore"
 	"github.com/stretchr/testify/assert"
 )
 
-func init() {
-	underscore.Disable() // It really slows down unit tests (by making otto.New take a lot longer)
-}
```
torcolvin, line 37: we are still running some of these tests with otto - that said, I'd be fine with putting this into a separate PR and moving this to `main_test.go` for relevant tests - and we can remove when we remove otto.

db/crud.go
```diff
@@ -1476,63 +1470,50 @@ func (db *DatabaseCollectionWithUser) storeOldBodyInRevTreeAndUpdateCurrent(ctx
...
-	if err != nil {
-		return
-	}
-
-	if curBody != nil {
+	if curBodyBytes != nil && string(curBodyBytes) != "null" {
 		base.DebugfCtx(ctx, base.KeyCRUD, "updateDoc(%q): Rev %q causes %q to become current again",
 			base.UD(doc.ID), newRevID, doc.CurrentRev)
-		channelSet, access, roles, syncExpiry, oldBodyJSON, err = db.getChannelsAndAccess(ctx, doc, curBody, metaMap, doc.CurrentRev)
+		channelSet, access, roles, syncExpiry, oldBodyJSON, err = db.getChannelsAndAccess(ctx, doc, curBodyBytes, metaMap, doc.CurrentRev, false /*TODO: Is false correct??*/)
```
torcolvin: address this TODO?

db/crud.go
```diff
@@ -1476,63 +1470,50 @@ func (db *DatabaseCollectionWithUser) storeOldBodyInRevTreeAndUpdateCurrent(ctx
...
-	}
-
 	return
 }
 
 // Run the sync function on the given document and body. Need to inject the document ID and rev ID temporarily to run
 // the sync function.
-func (db *DatabaseCollectionWithUser) runSyncFn(ctx context.Context, doc *Document, body Body, metaMap map[string]interface{}, newRevId string) (*uint32, string, base.Set, channels.AccessMap, channels.AccessMap, error) {
-	channelSet, access, roles, syncExpiry, oldBody, err := db.getChannelsAndAccess(ctx, doc, body, metaMap, newRevId)
+func (db *DatabaseCollectionWithUser) runSyncFn(ctx context.Context, doc *Document, body []byte, metaMap channels.MetaMap, newRevId string, deleted bool) (*uint32, string, base.Set, channels.AccessMap, channels.AccessMap, error) {
```
torcolvin, line 1496: why is it passing body separately helpful? Does it make sense to expose `Document._rawBody` here as a more clear method, or is there a mutability problem?

db/crud.go
```diff
@@ -2277,11 +2258,20 @@ func (col *DatabaseCollectionWithUser) getChannelsAndAccess(ctx context.Context,
...
+			RevID:   revID,
+			Deleted: deleted,
+			Body:    string(body),
+			OldBody: oldJson,
+			Meta:    metaMap,
+			UserCtx: MakeUserCtx(col.user, col.ScopeName, col.Name),
+		}
 		var output *channels.ChannelMapperOutput
 
+		base.DebugfCtx(ctx, base.KeyCRUD, "Calling sync fn with _id=%q, _rev=%q, _del=%v, body %s , oldBody %s", doc.ID, revID, doc.Deleted, body, oldJson) //TEMP
```
torcolvin: I'd drop this or put in trace logging

channels/channelmapper.go
```diff
@@ -60,79 +113,149 @@ func GetDefaultSyncFunction(scopeName, collectionName string) string {
...
-				return value
-			}
-			return asInt
-		} else {
-			numErr, _ := err.(*strconv.NumError)
-			if numErr.Err == strconv.ErrRange {
-				return value
-			}
+// Runs the sync function. Thread-safe.
+func (mapper *ChannelMapper) MapToChannelsAndAccess2(input ChannelMapperInput) (*ChannelMapperOutput, error) {
```
torcolvin: change the name of this function to `runSync` or `RunSync`?

channels/channelmapper.go
```diff
@@ -60,79 +113,149 @@ func GetDefaultSyncFunction(scopeName, collectionName string) string {
...
+	result, err := mapper.service.WithRunner(func(runner js.Runner) (any, error) {
+		ctx := context.Background()
+		if mapper.timeout > 0 {
+			var cancelFn context.CancelFunc
+			ctx, cancelFn = context.WithTimeout(ctx, mapper.timeout)
+			defer cancelFn()
```
torcolvin, line 152: otto doesn't use context cancellation, and I haven't found where this is set correctly, but is it a problem that we cancel the context when we are still using otto?

channels/channelmapper.go
```diff
@@ -60,79 +113,149 @@ func GetDefaultSyncFunction(scopeName, collectionName string) string {
...
+	Access           map[string][]string
+	Roles            map[string][]string
+	Expiry           any
+}
+
+func callSyncFn(runner js.Runner, in ChannelMapperInput) (*ChannelMapperOutput, error) {
+	// Call the sync fn:
+	jsResult, err := runner.Run(in.DocID, in.RevID, in.Deleted, js.JSONString(in.Body), in.OldBody, in.Meta.Key, string(in.Meta.JSONValue), in.UserCtx)
+	if err != nil {
+		return nil, fmt.Errorf("unexpected error calling sync fn: %+w", err)
```
torcolvin:
```suggestion
		return nil, fmt.Errorf("unexpected error calling sync fn: %q", err)
```

channels/channelmapper.go
```diff
@@ -60,79 +113,149 @@ func GetDefaultSyncFunction(scopeName, collectionName string) string {
...
+	}
+	var result syncFnResult
+	if err := json.Unmarshal([]byte(jsResult.(string)), &result); err != nil {
+		return nil, fmt.Errorf("unparseable output from sync-fn wrapper: %w", err)
```
torcolvin, line 181:
```suggestion
		return nil, fmt.Errorf("unparseable output from sync-fn wrapper: %q", err)
```

channels/channelmapper.go
```diff
@@ -60,79 +113,149 @@ func GetDefaultSyncFunction(scopeName, collectionName string) string {
...
+			return nil, err
 		}
-	case []interface{}:
-		for i, v := range value {
-			value[i] = ConvertJSONNumbers(v)
+	}
+	if output.Access, err = compileJSAccessMap(result.Access, ""); err != nil {
+		return nil, err
+	}
+	if output.Roles, err = compileJSAccessMap(result.Roles, "role:"); err != nil {
```
torcolvin:
```suggestion
	if output.Roles, err = compileJSAccessMap(result.Roles, RoleAccessPrefix); err != nil {
```

channels/channelmapper_test.go
```diff
@@ -9,595 +9,650 @@
...
-	"github.com/robertkrimen/otto/underscore"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	// Docs: https://pkg.go.dev/github.com/snej/v8go
 )
 
-func init() {
-	underscore.Disable() // It really slows down unit tests (by making otto.New take a lot longer)
-}
-
```
torcolvin, line 25: we don't want to drop this when we still support otto, move to `main_test.go` is fine

channels/channelmapper_test.go
```diff
@@ -9,595 +9,650 @@
...
-	mapper := NewChannelMapper(`function(doc) {channel(doc.x.concat(doc.y));}`, 0)
-	res, err := mapper.MapToChannelsAndAccess(parse(`{"x":["abc"],"y":["xyz"]}`), `{}`, emptyMetaMap(), noUser)
-	assert.NoError(t, err, "MapToChannelsAndAccess failed")
-	assert.Equal(t, BaseSetOf(t, "abc", "xyz"), res.Channels)
+	js.TestWithVMs(t, func(t *testing.T, vm js.VM) {
+		mapper := NewChannelMapper(vm, `function(doc) {channel(doc.x.concat(doc.y));}`, 0)
+		res, err := mapper.MapToChannelsAndAccess(parse(`{"x":["abc"],"y":["xyz"]}`), `{}`, emptyMetaMap(), noUser)
+		if assert.NoError(t, err, "MapToChannelsAndAccess failed") {
+			assert.Equal(t, BaseSetOf(t, "abc", "xyz"), res.Channels)
+		}
```
torcolvin:
```suggestion
		require.NoError(t, err)
		assert.Equal(t, BaseSetOf(t, "abc", "xyz"), res.Channels)
```

channels/channelmapper_test.go
```diff
@@ -9,595 +9,650 @@
...
+// Verify the sync fn cannot access the internal JS variables used in the wrapper.
+func TestHiddenVariables(t *testing.T) {
+	js.TestWithVMs(t, func(t *testing.T, vm js.VM) {
+		for _, hidden := range []string{"userCtx", "shouldValidate", "result", "eval", "Function"} {
+			mapper := NewChannelMapper(vm, `function(doc) {return `+hidden+`;}`, 0)
+			_, err := mapper.MapToChannelsAndAccess(parse(`{}`), `{}`, emptyMetaMap(), noUser)
+			if assert.Error(t, err, "Was able to access %q", hidden) {
+				assert.ErrorContains(t, err, "ReferenceError:")
+				assert.ErrorContains(t, err, " is not defined")
+			}
```
torcolvin:
```suggestion
			require.Error(t, err, "Was able to access %q", hidden)
			assert.ErrorContains(t, err, "ReferenceError:")
			assert.ErrorContains(t, err, " is not defined")
```

channels/channelmapper_test.go
```diff
@@ -9,595 +9,650 @@
...
-	assert.Equal(t, base.HTTPErrorf(403, base.SyncFnErrorMissingRole), res.Rejection)
-
-	res, err = mapper.MapToChannelsAndAccess(parse(`{"role": "girl"}`), `{}`, emptyMetaMap(), nil)
-	assert.NoError(t, err, "MapToChannelsAndAccess failed")
-	assert.Equal(t, nil, res.Rejection)
+		var sally = map[string]interface{}{"name": "sally", "roles": map[string]int{"girl": 1, "5yo": 1}}
+		res, err := mapper.MapToChannelsAndAccess(parse(`{"role": "girl"}`), `{}`, emptyMetaMap(), sally)
+		if assert.NoError(t, err, "MapToChannelsAndAccess failed") {
+			assert.Equal(t, nil, res.Rejection)
+		}
```
torcolvin, line 379:
```suggestion
		require.NoError(t, err, "MapToChannelsAndAccess failed")
		assert.Equal(t, nil, res.Rejection)
```

db/database.go
```diff
@@ -418,8 +423,20 @@ func NewDatabaseContext(ctx context.Context, dbName string, bucket base.Bucket,
...
+	if options.JavaScriptEngine != nil {
+		jsEngineName = *options.JavaScriptEngine
+	}
+	if engine := js.EngineNamed(jsEngineName); engine != nil {
+		dbContext.JS.Init(engine, MaxJavaScriptVMs)
+	} else {
+		return nil, fmt.Errorf("%q is not an available JavaScript engine", jsEngineName)
+	}
+
 	cleanupFunctions = append(cleanupFunctions, func() {
```
torcolvin: you need to move the 

		base.SyncGatewayStats.ClearDBStats(dbName)

to above initializing js engine because we have the unlikely potential to require an error.

db/functions/callbacks.go
```diff
@@ -0,0 +1,244 @@
...
+	if err := db.CheckTimeout(ctx); err != nil {
+		return err
+	}
+	d.db = dbc
+	d.ctx = ctx
+	if dbUser := dbc.User(); dbUser != nil {
+		d.user = &userCredentials{
+			Name:     dbUser.Name(),
+			Roles:    dbUser.RoleNames().AllKeys(),
+			Channels: dbUser.Channels().AllKeys(),
```
torcolvin, line 54: There's something weird here we can potentially address in the future users and roles are collection scoped, but a principal object can see all of them.

The graphql is not collection aware so this isn't immediately problematic.

db/functions/callbacks.go
```diff
@@ -0,0 +1,244 @@
...
+				body["_rev"] = rev.RevID
+			}
+
+			_, _, err = collection.Put(d.ctx, docID, body)
+			if err == nil {
+				break // success!
+			} else if status, _ := base.ErrorAsHTTPStatus(err); status != http.StatusConflict {
+				return false, err
+			}
+			// on conflict (race condition), retry...
```
torcolvin, line 145: is this  an important todo?

db/import.go
```diff
@@ -443,68 +442,32 @@ func (db *DatabaseCollectionWithUser) backupPreImportRevision(ctx context.Contex
...
-	}
-}
-
-// Calls a jsEventFunction returning an interface{}
-func (i *ImportFilterFunction) EvaluateFunction(ctx context.Context, doc Body) (bool, error) {
+	// Return a function that calls the function on a document:
+	return func(ctx context.Context, doc Body) (bool, error) {
+		if timeout > 0 {
+			var cancelFn context.CancelFunc
+			ctx, cancelFn = context.WithTimeout(ctx, timeout)
```
torcolvin, line 457: otto doesn't use context cancellation, is this being obeyed in otto?

db/import.go
```diff
@@ -443,68 +442,32 @@ func (db *DatabaseCollectionWithUser) backupPreImportRevision(ctx context.Contex
...
 		}
-		return boolResult, nil
-	default:
-		base.WarnfCtx(ctx, "Import filter function returned non-boolean result %v Type: %T", result, result)
-		return false, errors.New("Import filter function returned non-boolean value.")
+		switch result := result.(type) {
+		case bool:
+			return result, nil
+		case string:
+			return strconv.ParseBool(result)
```
torcolvin, line 469: What's the justification for dropping this error handling?

db/import.go
```diff
@@ -443,68 +442,32 @@ func (db *DatabaseCollectionWithUser) backupPreImportRevision(ctx context.Contex
...
+		}
+		return false, fmt.Errorf("import filter function returned non-boolean type %T", result)
```
torcolvin:
```suggestion
		return false, fmt.Errorf("import filter function returned non-boolean type %T: %+v", result, result)
```

rest/config.go
```diff
@@ -859,13 +877,13 @@ func (dbConfig *DbConfig) validateVersion(ctx context.Context, isEnterpriseEditi
...
 			for collectionName, collectionConfig := range scopeConfig.Collections {
-				if isEmpty, err := validateJavascriptFunction(collectionConfig.SyncFn); err != nil {
+				if isEmpty, err := validateJavascriptFunction(jsvm, collectionConfig.SyncFn, 1, 3); err != nil {
 					multiError = multiError.Append(fmt.Errorf("collection %q sync function error: %w", collectionName, err))
 				} else if isEmpty {
 					collectionConfig.SyncFn = nil
 				}
 
-				if isEmpty, err := validateJavascriptFunction(collectionConfig.ImportFilter); err != nil {
+				if isEmpty, err := validateJavascriptFunction(jsvm, collectionConfig.ImportFilter, 1, 99); err != nil {
```
torcolvin: you allow 999 args above, and this makes me think you want to put some constants for minJSArgs, maxImportFilterArguments

rest/persistent_config_test.go
```diff
@@ -389,7 +389,7 @@ func TestImportFilterEndpoint(t *testing.T) {
 	resp.RequireStatus(http.StatusOK)
 
 	// Modify the import filter to always reject import
-	resp = BootstrapAdminRequest(t, http.MethodPut, "/db1/_config/import_filter", `function(){return false}`)
+	resp = BootstrapAdminRequest(t, http.MethodPut, "/db1/_config/import_filter", `function(doc){return false}`)
```
torcolvin, line 392: is it breaking backward compatibility to require 1 argument? presumably we previously let 0 argument functions through?

A great feature of JS I suppose.

We _definitely_ support setting only one or two args for sync functions for backward compatibility and I think the the validation in `rest` package breaks that.

torcolvin (commented):
I didn't review all of this code by any means.

Things I didn't review due to time constraints:

- any JS/typescript code
- understanding the `GetDeepMutableBody` and the `ShallowCopy` in the crud code.

db/crud.go
```diff
@@ -1163,20 +1163,14 @@ func (db *DatabaseCollectionWithUser) resolveConflict(ctx context.Context, local
 	// TODO: Make doc expiry (_exp) available over replication.
 	// remoteExpiry := remoteDoc.Expiry
 
-	localDocBody, err := localDoc.GetDeepMutableBody()
-	if err != nil {
-		return "", nil, err
-	}
+	localDocBody := localDoc.Body().ShallowCopy()
```
torcolvin, line 1166: This needs to be `GetDeepMutableBody()` for otto and probably `Body().ShallowCopy()` for V8 - this is enough that I'd be tempted to write a shim to get the right part of the body?

I think not interacting with otto as go objects directly has the potential to slow down otto, which I don't think we want to do.

channels/channelmapper.go
snej: NewDatabaseContext doesn't actually call this! It calls UpdateSyncFun, but that method only calls SetFunction if there's already a ChannelMapper, which isn't true when the database is being created.

The only other callers of UpdateSyncFun are tests. As an experiment I just changed SetFunction to panic, and three unit tests failed, all because they called UpdateSyncFun directly after the db was open: TestSetFunction (obviously),  Test_resyncDocument, TestChannelAccessChanges.

db/functions/engine/README.md
snej: That could be good, but I myself am not good at CI stuff so I don't trust myself to write it.

channels/channelmapper.go
snej: Otto vs V8. Otto is stuck at a very old version of JS.

db/database_test.go
snej: I took that line out because it doesn't make a significant difference anymore, i.e. the tests don't really take longer with it removed. I remember adding that back in 2013 or so, and either MacBooks have gotten a lot faster or the Go runtime has (probably more of the latter.)

db/crud.go
snej: Ah, right. `false` is correct, so I'll remove the comment.

db/crud.go
snej: It's already being passed separately; the change is to pass it as unparsed JSON instead of a map.
If you mean "why can't the method just get it from `doc._rawBody`", that's because they're not always the same. (The reasons are complicated and I don't recall the details right now; look at the calling code.) That's why body is already being passed separately.

db/crud.go
snej: Changed it to Trace.

channels/channelmapper.go
snej: Good idea, how about `Run`?
I'll also move the old `MapToChannelsAndAccess` into `channelmapper_tests.go` since it's only used by tests.

channels/channelmapper.go
snej: `%w` has the nice side effect of wrapping the original error inside the one being created so it can be examined later.
But using `%+w` seems wrong; the docs don't say it supports the `+` modifier. I'll remove it.

channels/channelmapper.go
snej: See previous comment.

channels/channelmapper.go
snej: Good catch.

channels/channelmapper_test.go
snej: See earlier comment on the other instance of this; tl;dr I took it out because it's not slow anymore.

channels/channelmapper_test.go
snej: ok

channels/channelmapper_test.go
snej: ok

channels/channelmapper_test.go
snej: ok

channels/channelmapper.go
snej: js.OttoRunner [handles that](https://github.com/couchbase/sg-bucket/blob/30840f2aaf67deca57617ed0c9f942a85536ac2b/js/otto_runner.go#L101-L127).

db/database.go
snej: yup.

db/functions/callbacks.go
snej: It's not a todo, it's a description of what happens. This line is reached when there's a conflict, and since it's at the end of a `for` block it will go back up and retry.

db/import.go
snej: Yup, see the other comment above (it's handled in js.OttoRunner.)

db/import.go
snej: I don't think I dropped anything; [the equivalent code in the master branch](https://github.com/couchbase/sync_gateway/blob/3e36d801b8c6a659622d3d148dfcedd772c0b5ca/db/import.go#L500-L505) does the same thing, only with some extra no-op lines that unpack and repack the result.

db/import.go
snej: good idea.

rest/config.go
snej: It looks like an import filter only takes one arg, so I'll change the 999 and the 99 to 1.

rest/persistent_config_test.go
snej: A zero-arg import filter would be useless, wouldn't it? (If you always want to reject or approve import, you don't need a filter at all.)

channels/channelmapper.go
torcolvin: It might only get called under tests that run under GSI/CBS, but we do call it here https://github.com/couchbase/sync_gateway/blob/3e36d801b8c6a659622d3d148dfcedd772c0b5ca/db/database.go#L511 to determine if we need to do a resync.

rest/persistent_config_test.go
torcolvin: True! This test looks right and you took care of the validation for one argument which I think is correct.

channels/channelmapper.go
snej: That call is also in NewDatabaseContext, so it's only initializing the sync fn, not changing it, and SetFunction is not called. `dbCollection` was only instantiated 5 lines earlier, and newDatabaseCollection doesn't initialize c.ChannelMapper, so it's still nil.

The _only_ case where ChannelMapper.SetFunction is called is when a unit test calls UpdateSyncFun directly on an already-open database. That seems like a bad thing for a test to do since it's an otherwise-impossible change to the database.

snej:
Hm, got a single test failure: in "test v8 (ubuntu) (pull_request)" the test TestReconnectReplicator failed:
```
2023-06-07T23:42:31.2727087Z     utilities_testing_resttester.go:199: 
2023-06-07T23:42:31.2727596Z         	Error Trace:	/home/runner/work/sync_gateway/sync_gateway/rest/utilities_testing_resttester.go:199
2023-06-07T23:42:31.2728496Z         	            				/home/runner/work/sync_gateway/sync_gateway/rest/utilities_testing_resttester.go:203
2023-06-07T23:42:31.2729429Z         	            				/home/runner/work/sync_gateway/sync_gateway/rest/replicatortest/replicator_test.go:2451
2023-06-07T23:42:31.2729672Z         	Error:      	Received unexpected error:
2023-06-07T23:42:31.2730325Z         	            	RetryLoop for Wait for condition options giving up after 201 attempts
2023-06-07T23:42:31.2730658Z         	Test:       	TestReconnectReplicator/max_backoff_not_specified
2023-06-07T23:42:31.2731010Z         	Messages:   	Expected status: reconnecting, actual status: running
```

torcolvin:
> Hm, got a single test failure: in "test v8 (ubuntu) (pull_request)" the test TestReconnectReplicator failed:
> 
> ```
> 2023-06-07T23:42:31.2727087Z     utilities_testing_resttester.go:199: 
> 2023-06-07T23:42:31.2727596Z         	Error Trace:	/home/runner/work/sync_gateway/sync_gateway/rest/utilities_testing_resttester.go:199
> 2023-06-07T23:42:31.2728496Z         	            				/home/runner/work/sync_gateway/sync_gateway/rest/utilities_testing_resttester.go:203
> 2023-06-07T23:42:31.2729429Z         	            				/home/runner/work/sync_gateway/sync_gateway/rest/replicatortest/replicator_test.go:2451
> 2023-06-07T23:42:31.2729672Z         	Error:      	Received unexpected error:
> 2023-06-07T23:42:31.2730325Z         	            	RetryLoop for Wait for condition options giving up after 201 attempts
> 2023-06-07T23:42:31.2730658Z         	Test:       	TestReconnectReplicator/max_backoff_not_specified
> 2023-06-07T23:42:31.2731010Z         	Messages:   	Expected status: reconnecting, actual status: running
> ```

This is in a new test, @gregns1 can you have a look?

gregns1:
> > Hm, got a single test failure: in "test v8 (ubuntu) (pull_request)" the test TestReconnectReplicator failed:
> 
> > 
> 
> > ```
> 
> > 2023-06-07T23:42:31.2727087Z     utilities_testing_resttester.go:199: 
> 
> > 2023-06-07T23:42:31.2727596Z         	Error Trace:	/home/runner/work/sync_gateway/sync_gateway/rest/utilities_testing_resttester.go:199
> 
> > 2023-06-07T23:42:31.2728496Z         	            				/home/runner/work/sync_gateway/sync_gateway/rest/utilities_testing_resttester.go:203
> 
> > 2023-06-07T23:42:31.2729429Z         	            				/home/runner/work/sync_gateway/sync_gateway/rest/replicatortest/replicator_test.go:2451
> 
> > 2023-06-07T23:42:31.2729672Z         	Error:      	Received unexpected error:
> 
> > 2023-06-07T23:42:31.2730325Z         	            	RetryLoop for Wait for condition options giving up after 201 attempts
> 
> > 2023-06-07T23:42:31.2730658Z         	Test:       	TestReconnectReplicator/max_backoff_not_specified
> 
> > 2023-06-07T23:42:31.2731010Z         	Messages:   	Expected status: reconnecting, actual status: running
> 
> > ```
> 
> 
> 
> This is in a new test, @gregns1 can you have a look?

This looks like a race in the test. I'll have a look tomorrow to get a fix for it, you can ignore in the context of this PR.

snej:
I can't even find that test! I've got feature/v8+subgraph checked out, it's up-to-date with Github, but the string `TestReconnectReplicator` doesn't exist in the source tree and definitely not at `rest/replicatortest/replicator_test.go:2451`. Am I losing my mind?

snej:
The test log says `HEAD is now at e326d6b Merge 0891f2eda0e468ff4354c2e23090e4a7674f2d17 into 3e36d801b8c6a659622d3d148dfcedd772c0b5ca` -- that means it's merged master into the PR branch. I had no idea CI was testing these hypothetical merges ... I'm used to it testing the exact branch of the PR. Huh.

torcolvin:
Closing as inactive, this PR is linked to ticket if we want to go back to it.

PR #5981, opened 2022-12-19
CBG-2533 make _post_upgrade multi-collection aware

torcolvin:
- Didn't change the behavior of this function. This function could actually take a long time, dropping index is going to be synchronous for up to 2 minutes and might take longer. In practice, I think this would end up in a gateway timeout from the caller and this should probably be asynchronous cleanup. 
- Didn't add new tests for error behavior or tests in rest package because adding indexing tests stresses the GSI test harness and we can't do with walrus.
- Allow multi collection tests with default `db` tester which uses `DatabaseContextOptionsDirectly`.

I do feel like I short changed this PR by not writing tests for error conditions on the upgrade, but I'm not sure how to simulate index errors, and not sure if doing this is going to cause more stress on the test harness than we would benefit via test coverage.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1261/

### DIFF
--- a/db/indexes_test.go
+++ b/db/indexes_test.go
@@ -162,9 +162,6 @@ func TestPostUpgradeIndexesVersionChange(t *testing.T) {
 	if base.TestsDisableGSI() {
 		t.Skip("This test only works with Couchbase Server and UseViews=false")
 	}
-	if !base.TestUseXattrs() {
-		t.Skip("This test has expectation of starting with xattrs=true")
-	}
 
 	db, ctx := setupTestDB(t)
 	defer db.Close(ctx)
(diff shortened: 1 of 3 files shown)


db/indexes_test.go
```diff
@@ -162,6 +162,9 @@ func TestPostUpgradeIndexesVersionChange(t *testing.T) {
 	if base.TestsDisableGSI() {
 		t.Skip("This test only works with Couchbase Server and UseViews=false")
 	}
+	if !base.TestUseXattrs() {
```
adamcfraser: What's changed to make this test require xattrs?

adamcfraser (commented):
Enhancement looks good, one question about test change.

db/indexes_test.go
torcolvin: It relies on the non xattr indexes to be the ones that are "obsolete". I could have change the code to register the opposite of the tests (i.e. create xattr indexes when `SG_TEST_USE_XATTRS=false` and have them be the opposite).

This doesn't seem to mimic a real situation and since xattrs are usually true, it was just to simplify the test reading. I can set this variable dynamically like we do above.

adamcfraser (approved):

PR #5982, opened 2022-12-19
CBG-2534 make Database.Compact collection aware. 

torcolvin:
Move `Database.Compact` to iterate over collections.

There is one assertion in the modified test that is stumping me when `TestCompact(db.QueryTombstoneBatch + 20)`, I don't get the expected number of queries.

## Pre-review checklist

- [] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1281/


db/database.go
```diff
@@ -1451,77 +1451,80 @@ func (db *Database) Compact(ctx context.Context, skipRunningStateCheck bool, cal
...
+					}
+					return purgedDocCount, nil
+				default:
 				}
 
-				// At this point, the doc is not in a usable state for mobile
-				// so mark it to be removed from cache, even if the subsequent delete fails
-				purgedDocs = append(purgedDocs, tombstonesRow.Id)
+				resultCount++
+				base.DebugfCtx(ctx, base.KeyCRUD, "\tDeleting %q", tombstonesRow.Id)
```
adamcfraser, line 1483: This (and logging below) looks like it should be updated to include collection information, since ctx is at the database level.

adamcfraser (commented):
Changes look good, one proposed enhancement to the logging.

db/database.go
torcolvin: I wondered if this should be rolled into (https://issues.couchbase.com/browse/CBG-2621, and I thought about implementing CBG-2651 stubs for this.

db/database.go
```diff
@@ -1451,77 +1451,81 @@ func (db *Database) Compact(ctx context.Context, skipRunningStateCheck bool, cal
...
-	dataStore := db.Bucket.DefaultDataStore()
-
 	purgeBody := Body{"_purged": true}
-	for {
-		purgedDocs := make([]string, 0)
-		results, err := db.singleCollection.QueryTombstones(ctx, purgeOlderThan, QueryTombstoneBatch)
+	for _, c := range db.CollectionByID {
+		// create admin collection interface
+		collection, err := db.GetDatabaseCollectionWithUser(c.ScopeName(), c.Name())
+		ctx := base.CollectionNameCtx(ctx, collection.Name())
```
bbrks: If we got an error `collection.Name()` is going to panic here

db/database.go
torcolvin: I wrote a context for this,  but I'm not attached to it.

db/database.go
```diff
@@ -1455,6 +1455,7 @@ func (db *Database) Compact(ctx context.Context, skipRunningStateCheck bool, cal
 	for _, c := range db.CollectionByID {
 		// create admin collection interface
 		collection, err := db.GetDatabaseCollectionWithUser(c.ScopeName(), c.Name())
+		ctx := base.CollectionNameCtx(ctx, collection.Name())
```
adamcfraser: Shouldn't this be getting added more generically to the context in handler.go in the case where it's a keyspace endpoint, and passed through with the ctx parameter?

adamcfraser (changes requested):

db/database.go
torcolvin: I'm not sure I understand this question. 

Compact may be executed from an endpoint or through a Background task. If accessed through an endpoint it's stamped with a db (since this request is db scoped) and a request id.

I override the context here only inside the loop which becomes collection scoped.

bbrks (commented):
posting old comment I forgot about

db/database.go
adamcfraser: Makes sense, and agree with overriding it here given the usage.

adamcfraser (approved):

PR #5983, opened 2022-12-19
CBG-2645: Compact intermediate sequences in ISGR checkpointer expected lists

bbrks:
CBG-2645

- Iterates backwards over expected sequences and removes those which we know we're able to checkpoint past, if the list of expected sequences grows beyond 100 (`c.expectedSeqCompactionThreshold`)
- Logging improvements for `idAndRevLookup` failure and `norev` handling

Would like a keen eye on the iteration/cleanup logic in `_updateCheckpointLists` and the test case.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1277/

### DIFF
--- a/db/active_replicator_checkpointer.go
+++ b/db/active_replicator_checkpointer.go
@@ -280,17 +280,15 @@ func (c *Checkpointer) _updateCheckpointLists() (safeSeq *SequenceID) {
 	// if we have many remaining expectedSeqs, see if we can shrink the lists even more
 	// compact contiguous blocks of sequences by keeping only the last processed sequence in both lists
 	if len(c.expectedSeqs) > c.expectedSeqCompactionThreshold {
-		// compact processed sequences up to but not including the last one to ensure
-		// we're not keeping large amounts of intermediate sequence numbers in memory
+		// start at the one before the end of the list (since we know we need to retain that one anyway, if it's processed)
 		for i := len(c.expectedSeqs) - 2; i >= 0; i-- {
 			current := c.expectedSeqs[i]
 			next := c.expectedSeqs[i+1]
 			_, processedCurrent := c.processedSeqs[current]
 			_, processedNext := c.processedSeqs[next]
 			if processedCurrent && processedNext {
-				// remove the current sequence from both sets, since we know we've also processed the next sequence
+				// remove the current sequence from both sets, since we know we've also processed the next sequence and are able to checkpoint that
 				delete(c.processedSeqs, current)
-				// remove i from expectedSeqs
 				c.expectedSeqs = append(c.expectedSeqs[:i], c.expectedSeqs[i+1:]...)
 			}
 		}
--- a/db/active_replicator_checkpointer.go
+++ b/db/active_replicator_checkpointer.go
@@ -22,6 +22,8 @@ import (
 	"github.com/couchbase/sync_gateway/base"
 )
 
+const defaultExpectedSeqCompactionThreshold = 100
+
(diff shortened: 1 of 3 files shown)


db/active_replicator_checkpointer_test.go
```diff
@@ -29,90 +54,133 @@ func TestCheckpointerSafeSeq(t *testing.T) {
...
+				expectedSeqs:                   genExpectedForTest(t, "1", "2", "3", "4", "5", "6"),
+				processedSeqs:                  genProcessedForTest(t, "1", "3", "4", "5"),
+				expectedSeqCompactionThreshold: 3, // this many expected seqs to trigger compaction
+			},
+			expectedSafeSeq:         &SequenceID{Seq: 1},
+			expectedExpectedSeqsIdx: 0,
+			expectedExpectedSeqs:    genExpectedForTest(t, "2", "5", "6"),
+			expectedProcessedSeqs:   genProcessedForTest(t, "5"),
+		},
+		{
```
adamcfraser, line 173: I feel like adding a similar test that has non-sequential expected sequences would be a good idea.

db/active_replicator_checkpointer.go
```diff
@@ -72,17 +76,18 @@ type CheckpointerStats struct {
...
+		blipSender:                     blipSender,
+		activeDB:                       replicatorConfig.ActiveDB,
+		expectedSeqs:                   make([]SequenceID, 0),
+		processedSeqs:                  make(map[SequenceID]struct{}),
+		idAndRevLookup:                 make(map[IDAndRev]SequenceID),
+		checkpointInterval:             replicatorConfig.CheckpointInterval,
+		ctx:                            ctx,
+		stats:                          CheckpointerStats{},
+		statusCallback:                 statusCallback,
+		expectedSeqCompactionThreshold: 100,
```
adamcfraser: nit: making this a constant would be nice

db/active_replicator_checkpointer.go
```diff
@@ -265,9 +274,26 @@ func (c *Checkpointer) _updateCheckpointLists() (safeSeq *SequenceID) {
...
+		// start at the one before the end of the list (since we know we need to retain that one anyway, if it's processed)
+		for i := len(c.expectedSeqs) - 2; i >= 0; i-- {
+			current := c.expectedSeqs[i]
+			next := c.expectedSeqs[i+1]
+			_, processedCurrent := c.processedSeqs[current]
+			_, processedNext := c.processedSeqs[next]
+			if processedCurrent && processedNext {
+				// remove the current sequence from both sets, since we know we've also processed the next sequence and are able to checkpoint that
+				delete(c.processedSeqs, current)
+				c.expectedSeqs = append(c.expectedSeqs[:i], c.expectedSeqs[i+1:]...)
```
adamcfraser, line 294: If we had a large block of processed sequences, it would be nice to do this append just once, but I suppose the performance improvement doesn't merit the additional complexity/potential for bugs in this case?

adamcfraser (changes requested):
Overall looks good, just a request for an additional test case and a comment/question.

db/active_replicator_checkpointer_test.go
bbrks: Added non-sequential (2, 4, 6, 8, etc...) and also shuffled to ensure initial sorting is still being done ahead of compaction

db/active_replicator_checkpointer.go
bbrks: As discussed, this feels too complex to implement across multiple ranges for the benefits it'd give (given we're not growing the underlying array, it shouldn't be allocating/copying)

adamcfraser (approved):

PR #5990, opened 2022-12-21
CBG-2618: Update tests to use collection access for channels

gregns1:
CBG-2618

Changes to make all tests that use channels collection aware though using the "collection_access" api field. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1402/
- [x]  `GSI=true. xattrs=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1361/
- [x]  `GSI=false, xattrs=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1359/
- [x]  `GSI=false, xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1403/

### DIFF
--- a/rest/access_test.go
+++ b/rest/access_test.go
@@ -151,7 +151,7 @@ func TestStarAccess(t *testing.T) {
 	require.NoError(t, err)
 
 	// GET /db/_changes
-	response = rt.SendUserRequest("GET", "/{{.keyspace}}/_changes", "", "bernard")
+	response = rt.SendUserRequest("GET", "/{{.keyspace}}/_changes?since=0", "", "bernard")
 	RequireStatus(t, response, 200)
 	log.Printf("_changes looks like: %s", response.Body.Bytes())
 	err = base.JSONUnmarshal(response.Body.Bytes(), &changes)
--- a/rest/access_test.go
+++ b/rest/access_test.go
@@ -120,6 +120,7 @@ func TestStarAccess(t *testing.T) {
 	response := rt.SendAdminRequest("GET", "/{{.db}}/_user/bernard", ``)
 	RequireStatus(t, response, 200)
 	fmt.Println(response.Body)
+	fmt.Println("after user")
 
 	// GET /db/docid - basic test for channel user has
 	response = rt.SendUserRequest("GET", "/{{.keyspace}}/doc1", "", "bernard")
--- a/rest/access_test.go
+++ b/rest/access_test.go
@@ -397,11 +397,7 @@ func TestForceAPIForbiddenErrors(t *testing.T) {
 			c := collection.Name()
 			s := collection.ScopeName()
 
-			pass := "password"
-			userConfig := auth.PrincipalConfig{
-				Password: &pass,
-			}
-			payload, err := AdminChannelGrant(userConfig, collection, []string{"chan"})
+			payload, err := GetUserPayload("Perms", "password", "", collection, []string{"chan"}, nil)
 			require.NoError(t, err)
 			resp := rt.SendAdminRequest(http.MethodPut, "/{{.db}}/_user/Perms", payload)
 			RequireStatus(t, resp, http.StatusOK)
--- a/rest/access_test.go
+++ b/rest/access_test.go
@@ -11,6 +11,7 @@ package rest
 import (
 	"bytes"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"log"
 	"net/http"
@@ -114,12 +115,9 @@ func TestStarAccess(t *testing.T) {
 	//
 	// Part 1 - Tests for user with single channel access:
 	//
-	//response := rt.SendAdminRequest("PUT", "/db/_user/bernard", `{"password":"letmein",`+AdminChannelGrant(s, c, `"admin_channels":["books"]`)+`}`)
-	//RequireStatus(t, response, 201)
 	bernard, _ := a.NewUser("bernard", "letmein", channels.BaseSetOf(t, "books"))
 	assert.NoError(t, a.Save(bernard))
-	response := rt.SendAdminRequest("GET", "/db/_user/bernard", ``)
-	fmt.Println(response.Body)
+	response := rt.SendAdminRequest("GET", "/{{.db}}/_user/bernard", ``)
 
 	// GET /db/docid - basic test for channel user has
 	response = rt.SendUserRequest("GET", "/{{.keyspace}}/doc1", "", "bernard")
@@ -193,7 +191,8 @@ func TestStarAccess(t *testing.T) {
 	//
 
 	// Create a user:
-	fran, _ := a.NewUser("fran", "letmein", channels.BaseSetOf(t, "*"))
+	fran, err := a.NewUser("fran", "letmein", channels.BaseSetOf(t, "*"))
+	assert.NoError(t, err)
 	assert.NoError(t, a.Save(fran))
 
 	// GET /db/docid - basic test for doc that has channel
@@ -395,7 +394,7 @@ func TestForceAPIForbiddenErrors(t *testing.T) {
 			c := collection.Name()
 			s := collection.ScopeName()
 
-			resp := rt.SendAdminRequest(http.MethodPut, "/db/_user/Perms", `{"password": "password", `+AdminChannelGrant(s, c, `"admin_channels": ["chan"]`)+`}`)
+			resp := rt.SendAdminRequest(http.MethodPut, "/{{.db}}/_user/Perms", `{"password": "password", `+AdminChannelGrant(collection, `"admin_channels": ["chan"]`)+`}`)
 			RequireStatus(t, resp, http.StatusOK)
 
 			// Create the initial document
@@ -483,16 +482,31 @@ func TestForceAPIForbiddenErrors(t *testing.T) {
 
 			user, err := a.GetUser("NoPerms")
 			assert.NoError(t, err)
-			collAccess := user.GetCollectionsAccess()[s][c]
-			assert.NotContains(t, collAccess.Channels_.String(), "chan2")
+			var allChannels struct {
+				Channels []string `json:"all_channels"`
+			}
+			if base.TestsDisableGSI() {
+				err := json.Unmarshal(resp.BodyBytes(), &allChannels)
+				require.NoError(t, err)
+				assert.NotContains(t, allChannels.Channels, "chan2")
+			} else {
+				collAccess := user.GetCollectionsAccess()[s][c]
+				assert.NotContains(t, collAccess.Channels_.String(), "chan2")
+			}
 
 			resp = rt.SendAdminRequest(http.MethodGet, "/db/_user/Perms", ``)
 			RequireStatus(t, resp, http.StatusOK)
 
-			user, err = a.GetUser("NoPerms")
-			assert.NoError(t, err)
-			collAccess = user.GetCollectionsAccess()[s][c]
-			assert.NotContains(t, collAccess.Channels_.String(), "chan2")
+			if base.TestsDisableGSI() {
+				err = json.Unmarshal(resp.BodyBytes(), &allChannels)
+				require.NoError(t, err)
+				assert.NotContains(t, allChannels.Channels, "chan2")
+			} else {
+				user, err = a.GetUser("NoPerms")
+				assert.NoError(t, err)
+				collAccess := user.GetCollectionsAccess()[s][c]
+				assert.NotContains(t, collAccess.Channels_.String(), "chan2")
+			}
 
 			// Successful PUT which will grant access grants
 			resp = rt.SendUserRequestWithHeaders(http.MethodPut, "/{{.keyspace}}/doc?rev="+rev, `{"channels": "chan"}`, nil, "Perms", "password")
@@ -501,11 +515,16 @@ func TestForceAPIForbiddenErrors(t *testing.T) {
 			// Make sure channel access grant was successful
 			resp = rt.SendAdminRequest(http.MethodGet, "/db/_user/Perms", ``)
 			RequireStatus(t, resp, http.StatusOK)
-
-			user, err = a.GetUser("Perms")
-			assert.NoError(t, err)
-			collAccess = user.GetCollectionsAccess()[s][c]
-			assert.Contains(t, collAccess.Channels_.String(), "chan2")
+			if base.TestsDisableGSI() {
+				err = json.Unmarshal(resp.BodyBytes(), &allChannels)
+				require.NoError(t, err)
+				assert.Contains(t, allChannels.Channels, "chan2")
+			} else {
+				user, err = a.GetUser("Perms")
+				assert.NoError(t, err)
+				collAccess := user.GetCollectionsAccess()[s][c]
+				assert.Contains(t, collAccess.Channels_.String(), "chan2")
+			}
 
 			// DELETE requests
 			// Attempt to delete document with no permissions
@@ -850,14 +869,14 @@ func TestChannelAccessChanges(t *testing.T) {
 	assert.Equal(t, true, body["ok"])
 	alphaRevID := body["rev"].(string)
 
-	RequireStatus(t, rt.Send(Request("PUT", "/db."+s+"."+c+"/beta", `{"owner":"boadecia"}`)), 201)
-	RequireStatus(t, rt.Send(Request("PUT", "/db."+s+"."+c+"/delta", `{"owner":"alice"}`)), 201)
-	RequireStatus(t, rt.Send(Request("PUT", "/db."+s+"."+c+"/gamma", `{"owner":"zegpold"}`)), 201)
+	RequireStatus(t, rt.SendRequest("PUT", "/{{.keyspace}}/beta", `{"owner":"boadecia"}`), 201)
+	RequireStatus(t, rt.SendRequest("PUT", "/{{.keyspace}}/delta", `{"owner":"alice"}`), 201)
+	RequireStatus(t, rt.SendRequest("PUT", "/{{.keyspace}}/gamma", `{"owner":"zegpold"}`), 201)
 
-	RequireStatus(t, rt.Send(Request("PUT", "/db."+s+"."+c+"/a1", `{"channel":"alpha"}`)), 201)
-	RequireStatus(t, rt.Send(Request("PUT", "/db."+s+"."+c+"/b1", `{"channel":"beta"}`)), 201)
-	RequireStatus(t, rt.Send(Request("PUT", "/db."+s+"."+c+"/d1", `{"channel":"delta"}`)), 201)
-	RequireStatus(t, rt.Send(Request("PUT", "/db."+s+"."+c+"/g1", `{"channel":"gamma"}`)), 201)
+	RequireStatus(t, rt.SendRequest("PUT", "/{{.keyspace}}/a1", `{"channel":"alpha"}`), 201)
+	RequireStatus(t, rt.SendRequest("PUT", "/{{.keyspace}}/b1", `{"channel":"beta"}`), 201)
+	RequireStatus(t, rt.SendRequest("PUT", "/{{.keyspace}}/d1", `{"channel":"delta"}`), 201)
+	RequireStatus(t, rt.SendRequest("PUT", "/{{.keyspace}}/g1", `{"channel":"gamma"}`), 201)
 
 	rt.MustWaitForDoc("g1", t)
 
--- a/rest/access_test.go
+++ b/rest/access_test.go
@@ -49,14 +49,11 @@ func TestPublicChanGuestAccess(t *testing.T) {
 
 	resp = rt.SendAdminRequest(http.MethodGet, "/db/_user/GUEST", ``)
 	RequireStatus(t, resp, http.StatusOK)
+	user := auth.PrincipalConfig{}
 	fmt.Println("GUEST user:", resp.Body.String())
-	if base.IsDefaultCollection(s, c) {
-		assert.EqualValues(t, []interface{}{"!"}, resp.GetRestDocument()["all_channels"])
-	} else {
-		expected := fmt.Sprintf(`map[%s:map[%s:map[all_channels:[!]]]]`, s, c)
-		collectionAccess := resp.GetRestDocument()["collection_access"]
-		assert.EqualValues(t, expected, fmt.Sprint(collectionAccess))
-	}
+	err := json.Unmarshal(resp.BodyBytes(), &user)
+	require.NoError(t, err)
+	assert.EqualValues(t, []string{"!"}, user.GetChannels(s, c).ToArray())
 
 	// Confirm guest user cannot access other channels it has no access too
 	resp = rt.SendAdminRequest(http.MethodPut, "/{{.keyspace}}/docNoAccess", `{"channels": ["cookie"], "foo": "bar"}`)
@@ -393,13 +390,17 @@ func TestForceAPIForbiddenErrors(t *testing.T) {
 					}},
 				})
 			defer rt.Close()
-			ctx := base.TestCtx(t)
-			a := rt.ServerContext().Database(ctx, "db").Authenticator(ctx)
 			collection := rt.GetSingleTestDatabaseCollection()
 			c := collection.Name()
 			s := collection.ScopeName()
 
-			resp := rt.SendAdminRequest(http.MethodPut, "/{{.db}}/_user/Perms", `{"password": "password", `+AdminChannelGrant(collection, `"admin_channels": ["chan"]`)+`}`)
+			pass := "password"
+			userConfig := auth.PrincipalConfig{
+				Password: &pass,
+			}
+			payload, err := AdminChannelGrant(userConfig, collection, []string{"chan"})
+			require.NoError(t, err)
+			resp := rt.SendAdminRequest(http.MethodPut, "/{{.db}}/_user/Perms", payload)
 			RequireStatus(t, resp, http.StatusOK)
 
 			// Create the initial document
@@ -485,33 +486,17 @@ func TestForceAPIForbiddenErrors(t *testing.T) {
 			resp = rt.SendAdminRequest(http.MethodGet, "/db/_user/NoPerms", ``)
 			RequireStatus(t, resp, http.StatusOK)
 
-			user, err := a.GetUser("NoPerms")
-			assert.NoError(t, err)
-			var allChannels struct {
-				Channels []string `json:"all_channels"`
-			}
-			if base.IsDefaultCollection(collection.Name(), collection.ScopeName()) {
-				err := json.Unmarshal(resp.BodyBytes(), &allChannels)
-				require.NoError(t, err)
-				assert.NotContains(t, allChannels.Channels, "chan2")
-			} else {
-				collAccess := user.GetCollectionsAccess()[s][c]
-				assert.NotContains(t, collAccess.Channels_.String(), "chan2")
-			}
+			user := auth.PrincipalConfig{}
+			err = json.Unmarshal(resp.BodyBytes(), &user)
+			require.NoError(t, err)
+			assert.NotContains(t, user.GetChannels(s, c).ToArray(), "chan2")
 
 			resp = rt.SendAdminRequest(http.MethodGet, "/db/_user/Perms", ``)
 			RequireStatus(t, resp, http.StatusOK)
 
-			if base.IsDefaultCollection(collection.Name(), collection.ScopeName()) {
-				err = json.Unmarshal(resp.BodyBytes(), &allChannels)
-				require.NoError(t, err)
-				assert.NotContains(t, allChannels.Channels, "chan2")
-			} else {
-				user, err = a.GetUser("NoPerms")
-				assert.NoError(t, err)
-				collAccess := user.GetCollectionsAccess()[s][c]
-				assert.NotContains(t, collAccess.Channels_.String(), "chan2")
-			}
+			err = json.Unmarshal(resp.BodyBytes(), &user)
+			require.NoError(t, err)
+			assert.NotContains(t, user.GetChannels(s, c).ToArray(), "chan2")
 
 			// Successful PUT which will grant access grants
 			resp = rt.SendUserRequestWithHeaders(http.MethodPut, "/{{.keyspace}}/doc?rev="+rev, `{"channels": "chan"}`, nil, "Perms", "password")
@@ -520,16 +505,10 @@ func TestForceAPIForbiddenErrors(t *testing.T) {
 			// Make sure channel access grant was successful
 			resp = rt.SendAdminRequest(http.MethodGet, "/db/_user/Perms", ``)
 			RequireStatus(t, resp, http.StatusOK)
-			if base.IsDefaultCollection(collection.Name(), collection.ScopeName()) {
-				err = json.Unmarshal(resp.BodyBytes(), &allChannels)
-				require.NoError(t, err)
-				assert.Contains(t, allChannels.Channels, "chan2")
-			} else {
-				user, err = a.GetUser("Perms")
-				assert.NoError(t, err)
-				collAccess := user.GetCollectionsAccess()[s][c]
-				assert.Contains(t, collAccess.Channels_.String(), "chan2")
-			}
+
+			err = json.Unmarshal(resp.BodyBytes(), &user)
+			require.NoError(t, err)
+			assert.Contains(t, user.GetChannels(s, c).ToArray(), "chan2")
 
 			// DELETE requests
 			// Attempt to delete document with no permissions
@@ -605,9 +584,6 @@ func TestBulkDocsChangeToAccess(t *testing.T) {
 func TestAllDocsAccessControl(t *testing.T) {
 	rt := NewRestTester(t, nil)
 	defer rt.Close()
-	collection := rt.GetSingleTestDatabaseCollection()
-	c := collection.Name()
-	s := collection.ScopeName()
 
 	type allDocsRow struct {
 		ID    string `json:"id"`
@@ -806,7 +782,6 @@ func TestAllDocsAccessControl(t *testing.T) {
 	assert.Equal(t, []string{"Cinemax"}, allDocsResult.Rows[0].Value.Channels)
 
 	// Check _all_docs as admin:
-	response = rt.SendAdminRequest("GET", "/db."+s+"."+c+"/_all_docs", "")
 	response = rt.SendAdminRequest(http.MethodGet, "/{{.keyspace}}/_all_docs", "")
 	RequireStatus(t, response, 200)
 
@@ -980,12 +955,12 @@ func TestChannelAccessChanges(t *testing.T) {
 	dbc := rt.ServerContext().Database(ctx, "db")
 	database, _ := db.GetDatabase(dbc, nil)
 
-	collectionUser := database.GetSingleDatabaseCollectionWithUser()
+	collectionWithUser := database.GetSingleDatabaseCollectionWithUser()
 
 	changed, err := database.UpdateSyncFun(ctx, `function(doc) {access("alice", "beta");channel("beta");}`)
 	assert.NoError(t, err)
 	assert.True(t, changed)
-	changeCount, err := collectionUser.UpdateAllDocChannels(ctx, false, func(docsProcessed, docsChanged *int) {}, base.NewSafeTerminator())
+	changeCount, err := collectionWithUser.UpdateAllDocChannels(ctx, false, func(docsProcessed, docsChanged *int) {}, base.NewSafeTerminator())
 	assert.NoError(t, err)
 	assert.Equal(t, 9, changeCount)
 
--- a/rest/access_test.go
+++ b/rest/access_test.go
@@ -9,7 +9,6 @@
 package rest
 
 import (
-	"bytes"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
@@ -119,7 +118,8 @@ func TestStarAccess(t *testing.T) {
 	//
 	// Part 1 - Tests for user with single channel access:
 	//
-	bernard, _ := a.NewUser("bernard", "letmein", channels.BaseSetOf(t, "books"))
+	bernard, err := a.NewUser("bernard", "letmein", channels.BaseSetOf(t, "books"))
+	assert.NoError(t, err)
 	assert.NoError(t, a.Save(bernard))
 
 	// GET /db/docid - basic test for channel user has
@@ -652,21 +652,15 @@ func TestAllDocsAccessControl(t *testing.T) {
 	assert.NoError(t, a.Save(alice))
 
 	// Get a single doc the user has access to:
-	request, _ := http.NewRequest("GET", "/db."+s+"."+c+"/doc3", nil)
-	request.SetBasicAuth("alice", "letmein")
-	response := rt.Send(request)
+	response := rt.SendUserRequest(http.MethodGet, "/{{.keyspace}}/doc3", "", "alice")
 	RequireStatus(t, response, 200)
 
 	// Get a single doc the user doesn't have access to:
-	request, _ = http.NewRequest("GET", "/db."+s+"."+c+"/doc2", nil)
-	request.SetBasicAuth("alice", "letmein")
-	response = rt.Send(request)
+	response = rt.SendUserRequest(http.MethodGet, "/{{.keyspace}}/doc2", "", "alice")
 	RequireStatus(t, response, 403)
 
 	// Check that _all_docs only returns the docs the user has access to:
-	request, _ = http.NewRequest("GET", "/db."+s+"."+c+"/_all_docs?channels=true", nil)
-	request.SetBasicAuth("alice", "letmein")
-	response = rt.Send(request)
+	response = rt.SendUserRequest(http.MethodGet, "/{{.keyspace}}/_all_docs?channels=true", "", "alice")
 	RequireStatus(t, response, 200)
 
 	allDocsResult := allDocsResponse{}
@@ -682,9 +676,7 @@ func TestAllDocsAccessControl(t *testing.T) {
 	assert.Equal(t, []string{"Cinemax"}, allDocsResult.Rows[2].Value.Channels)
 
 	// Check all docs limit option
-	request, _ = http.NewRequest("GET", "/db."+s+"."+c+"/_all_docs?limit=1&channels=true", nil)
-	request.SetBasicAuth("alice", "letmein")
-	response = rt.Send(request)
+	response = rt.SendUserRequest(http.MethodGet, "/{{.keyspace}}/_all_docs?limit=1&channels=true", "", "alice")
 	RequireStatus(t, response, 200)
 
 	log.Printf("Response = %s", response.Body.Bytes())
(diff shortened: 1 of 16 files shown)


rest/api_test_no_race_test.go
```diff
@@ -60,16 +63,17 @@ func TestChangesAccessNotifyInteger(t *testing.T) {
 			Last_Seq db.SequenceID
 		}
 		changesJSON := `{"style":"all_docs", "heartbeat":300000, "feed":"longpoll", "limit":50, "since":"0"}`
-		changesResponse := rt.Send(RequestByUser("POST", "/db/_changes", changesJSON, "bernard"))
-		err = base.JSONUnmarshal(changesResponse.Body.Bytes(), &changes)
+		changesResponse := rt.Send(RequestByUser("POST", "/db."+s+"."+c+"/_changes", changesJSON, "bernard"))
```
torcolvin: this gets changed to:

```
changesResponse := rt.SendUserRequest("POST", "/{{.keyspace}}/_changes", changesJSON, "bernard"))
```

rest/utilities_testing.go
```diff
@@ -1386,6 +1386,15 @@ func (bt *BlipTester) SendRev(docId, docRev string, body []byte, properties blip
 
 }
 
+func AdminChannelGrant(scopeName, collectionName, adminChannels string) (collectionAdminChannels string) {
```
torcolvin: This function is great!  A couple suggestions:

(1) Write a docstring
(2) When possible, avoid passing three identical objects in a row lest you pass them in the wrong order.

My suggestion would be 

`AdminChannelGrant(rt.GetSingleDatabaseCollection(), adminChannels)` 

so it's one line in code and `* db.DatabaseCollection` in the func signature

rest/user_api_test.go
```diff
@@ -941,7 +965,7 @@ function(doc, oldDoc) {
 		input = input + `]}`
 
 		log.Printf("Sending 2nd round of _bulk_docs")
-		response = rt.Send(RequestByUser("POST", "/db/_bulk_docs", input, "bernard"))
+		response = rt.Send(RequestByUser("POST", "/db."+s+"."+c+"/_bulk_docs", input, "bernard"))
```
torcolvin: swap to rt.RequestByUser

rest/user_api_test.go
```diff
@@ -896,7 +920,7 @@ function(doc, oldDoc) {
 			// Timeout allows us to read continuous changes after processing is complete.  Needs to be long enough to
 			// ensure it doesn't terminate before the first change is sent.
 			log.Printf("Invoking _changes?feed=continuous&since=%s&timeout=2000", since)
-			changesResponse := rt.Send(RequestByUser("GET", fmt.Sprintf("/db/_changes?feed=continuous&since=%s&timeout=2000", since), "", "bernard"))
+			changesResponse := rt.Send(RequestByUser("GET", fmt.Sprintf("/db."+s+"."+c+"/_changes?feed=continuous&since=%s&timeout=2000", since), "", "bernard"))
```
torcolvin: swap to rt.RequestByUser

rest/user_api_test.go
```diff
@@ -869,7 +893,7 @@ function(doc, oldDoc) {
 		input = input + fmt.Sprintf(`{"_id":"%s", "type":"list"}`, docId)
 	}
 	input = input + `]}`
-	response = rt.SendAdminRequest("POST", "/db/_bulk_docs", input)
+	response = rt.SendAdminRequest("POST", "/db."+s+"."+c+"/_bulk_docs", input)
```
torcolvin: this should be `{{.keyspace}}` here

rest/user_api_test.go
```diff
@@ -847,11 +868,14 @@ function(doc, oldDoc) {
...
-	rt := NewRestTesterDefaultCollection(t, // CBG-2618: fix collection channel access
+	rt := NewRestTester(t,
 		&rtConfig)
 	defer rt.Close()
+	collection := rt.GetSingleTestDatabaseCollection()
+	c := collection.Name()
+	s := collection.ScopeName()
 
-	response := rt.SendAdminRequest("PUT", "/db/_user/bernard", `{"name":"bernard", "password":"letmein", "admin_channels":["profile-bernard"]}`)
+	response := rt.SendAdminRequest("PUT", "/db/_user/bernard", `{"name":"bernard", "password":"letmein",`+AdminChannelGrant(s, c, `"admin_channels":["profile-bernard"]`)+`}`)
```
torcolvin: when we are changing stuff, i'd swap 

`/db/` to `/{{.db}/` but it's not a requirement

rest/user_api_test.go
```diff
@@ -318,26 +329,27 @@ func TestUserAPI(t *testing.T) {
...
 
 	// GET the user and verify that it shows the channels inherited from the role
 	response = rt.SendAdminRequest("GET", "/db/_user/snej", "")
 	RequireStatus(t, response, 200)
+	user, _ = rt.ServerContext().Database(ctx, "db").Authenticator(ctx).GetUser("snej")
 	body = nil
 	require.NoError(t, base.JSONUnmarshal(response.Body.Bytes(), &body))
 	assert.Equal(t, []interface{}{"hipster"}, body["admin_roles"])
-	assert.Equal(t, []interface{}{"!", "bar", "fedoras", "fixies", "foo"}, body["all_channels"])
+	assert.Equal(t, "!:1,bar:3,foo:3", user.CollectionChannels(s, c).String())
```
torcolvin: this seems weird that you aren't getting the same channels back at all. Also, I'd expect the original output for `SG_TEST_USE_DEFAULT_COLLECTION=true` and something different for the new output.

rest/api_test_no_race_test.go
```diff
@@ -126,7 +133,7 @@ func TestChangesNotifyChannelFilter(t *testing.T) {
 					 "filter":"` + base.ByChannelFilter + `",
 					 "channels":"ABC,PBS"}`
 	sinceZeroJSON := fmt.Sprintf(changesJSON, "0")
-	changesResponse := rt.Send(RequestByUser("POST", "/db/_changes", sinceZeroJSON, "bernard"))
+	changesResponse := rt.Send(RequestByUser("POST", "/db."+s+"."+c+"/_changes", sinceZeroJSON, "bernard"))
```
torcolvin: swap this to:

```
changesResponse := rt.SendUserRequest("POST", "/{{.keyspace}}/_changes", sinceZeroJSON, "bernard")
```

rest/api_test_no_race_test.go
```diff
@@ -144,7 +151,7 @@ func TestChangesNotifyChannelFilter(t *testing.T) {
 			Last_Seq db.SequenceID
 		}
 		sinceLastJSON := fmt.Sprintf(changesJSON, lastSeq)
-		changesResponse := rt.Send(RequestByUser("POST", "/db/_changes", sinceLastJSON, "bernard"))
+		changesResponse := rt.Send(RequestByUser("POST", "/db."+s+"."+c+"/_changes", sinceLastJSON, "bernard"))
```
torcolvin: swap to:

```
changesResponse := rt.SendUserRequest("POST", "/{{.keyspace}}/_changes", sinceZeroJSON, "bernard")
```

rest/changes_test.go
```diff
@@ -103,41 +106,41 @@ func TestUserJoiningPopulatedChannel(t *testing.T) {
...
 	// Create user1
-	response := rt.SendAdminRequest("PUT", "/db/_user/user1", `{"email":"user1@couchbase.com", "password":"letmein", "admin_channels":["alpha"]}`)
+	response := rt.SendAdminRequest("PUT", "/db/_user/user1", `{"email":"user1@couchbase.com", "password":"letmein",`+AdminChannelGrant(s, c, `"admin_channels":["alpha"]`)+`}`)
 	RequireStatus(t, response, 201)
 
 	// Create 100 docs
 	for i := 0; i < 100; i++ {
-		docpath := fmt.Sprintf("/db/doc%d", i)
+		docpath := fmt.Sprintf("/db."+s+"."+c+"/doc%d", i)
 		RequireStatus(t, rt.Send(Request("PUT", docpath, `{"foo": "bar", "channels":["alpha"]}`)), 201)
```
torcolvin: If you change this to `rt.SendRequest`, you can modify 114 to be:

docpath := fmt.Sprintf("/{{.keyspace}}/doc%d", i)

rest/changestest/changes_api_test.go
```diff
@@ -636,7 +645,7 @@ func TestPostChangesAdminChannelGrantRemoval(t *testing.T) {
 	}
 
 	// Update the user doc to grant access to PBS, HBO in addition to ABC
-	response := rt.SendAdminRequest(http.MethodPut, "/db/_user/bernard", `{"admin_channels":["ABC", "PBS", "HBO"]}`)
+	response := rt.SendAdminRequest(http.MethodPut, "/db/_user/bernard", `{`+rest.AdminChannelGrant(s, c, `"admin_channels":["ABC", "PBS", "HBO"]`)+`}`)
```
torcolvin: swap to `{{.db}}`

rest/changestest/changes_api_test.go
```diff
@@ -736,7 +748,7 @@ func TestPostChangesAdminChannelGrantRemovalWithLimit(t *testing.T) {
 	cacheWaiter.AddAndWait(1)
 
 	// Grant user access to channel PBS
-	userResponse := rt.SendAdminRequest("PUT", "/db/_user/bernard", `{"admin_channels":["ABC","PBS"]}`)
+	userResponse := rt.SendAdminRequest("PUT", "/db/_user/bernard", `{`+rest.AdminChannelGrant(s, c, `"admin_channels":["ABC","PBS"]`)+`}`)
```
torcolvin: swap to `{{.db}}`

rest/access_test.go
```diff
@@ -93,38 +98,44 @@ func TestStarAccess(t *testing.T) {
...
+	guest, err = a.GetUser("")
+	assert.NoError(t, err)
 	guest.SetDisabled(true)
 	err = a.Save(guest)
 	assert.NoError(t, err)
 	//
 	// Part 1 - Tests for user with single channel access:
 	//
-	bernard, err := a.NewUser("bernard", "letmein", channels.BaseSetOf(t, "books"))
+	//response := rt.SendAdminRequest("PUT", "/db/_user/bernard", `{"password":"letmein",`+AdminChannelGrant(s, c, `"admin_channels":["books"]`)+`}`)
```
torcolvin: delete dead code

rest/access_test.go
```diff
@@ -93,38 +98,44 @@ func TestStarAccess(t *testing.T) {
...
+	//RequireStatus(t, response, 201)
+	bernard, _ := a.NewUser("bernard", "letmein", channels.BaseSetOf(t, "books"))
 	assert.NoError(t, a.Save(bernard))
+	response := rt.SendAdminRequest("GET", "/db/_user/bernard", ``)
```
torcolvin: this can be `{{.db}}`

rest/access_test.go
```diff
@@ -93,38 +98,44 @@ func TestStarAccess(t *testing.T) {
...
+	fmt.Println(response.Body)
```
torcolvin: remove print

rest/access_test.go
```diff
@@ -182,20 +193,20 @@ func TestStarAccess(t *testing.T) {
 	//
 
 	// Create a user:
-	fran, err := a.NewUser("fran", "letmein", channels.BaseSetOf(t, "*"))
+	fran, _ := a.NewUser("fran", "letmein", channels.BaseSetOf(t, "*"))
```
torcolvin: It's great that you caught this was unused! but let's add `require.NoError(t, err)` since everything after this test will fail if new user doesn't work

rest/access_test.go
```diff
@@ -375,151 +386,161 @@ func TestForceAPIForbiddenErrors(t *testing.T) {
...
 					}},
 				})
 			defer rt.Close()
+			ctx := base.TestCtx(t)
+			a := rt.ServerContext().Database(ctx, "db").Authenticator(ctx)
+			collection := rt.GetSingleTestDatabaseCollection()
+			c := collection.Name()
+			s := collection.ScopeName()
+
+			resp := rt.SendAdminRequest(http.MethodPut, "/db/_user/Perms", `{"password": "password", `+AdminChannelGrant(s, c, `"admin_channels": ["chan"]`)+`}`)
```
torcolvin: use `{{.db}}` here

rest/access_test.go
```diff
@@ -603,19 +631,19 @@ func TestAllDocsAccessControl(t *testing.T) {
 	assert.NoError(t, a.Save(alice))
 
 	// Get a single doc the user has access to:
-	request, _ := http.NewRequest("GET", "/db/doc3", nil)
+	request, _ := http.NewRequest("GET", "/db."+s+"."+c+"/doc3", nil)
 	request.SetBasicAuth("alice", "letmein")
 	response := rt.Send(request)
 	RequireStatus(t, response, 200)
```
torcolvin: you can change this to:

```
response := rt.SendUserRequest(http.MethodGet, "/{{.keyspace}}/doc3", nil, "alice")
rest.RequireStatus(response, http.StatusOK)
```

This was presumably written before we had a rest helper function

rest/access_test.go
```diff
@@ -603,19 +631,19 @@ func TestAllDocsAccessControl(t *testing.T) {
...
 
 	// Get a single doc the user doesn't have access to:
-	request, _ = http.NewRequest("GET", "/db/doc2", nil)
+	request, _ = http.NewRequest("GET", "/db."+s+"."+c+"/doc2", nil)
 	request.SetBasicAuth("alice", "letmein")
 	response = rt.Send(request)
 	RequireStatus(t, response, 403)
```
torcolvin: switch to SendUserRequest

rest/access_test.go
```diff
@@ -603,19 +631,19 @@ func TestAllDocsAccessControl(t *testing.T) {
...
 
 	// Check that _all_docs only returns the docs the user has access to:
-	request, _ = http.NewRequest("GET", "/db/_all_docs?channels=true", nil)
+	request, _ = http.NewRequest("GET", "/db."+s+"."+c+"/_all_docs?channels=true", nil)
 	request.SetBasicAuth("alice", "letmein")
 	response = rt.Send(request)
 	RequireStatus(t, response, 200)
```
torcolvin: switch to SendUserRequest

rest/access_test.go
```diff
@@ -633,7 +661,7 @@ func TestAllDocsAccessControl(t *testing.T) {
 	assert.Equal(t, []string{"Cinemax"}, allDocsResult.Rows[2].Value.Channels)
 
 	// Check all docs limit option
-	request, _ = http.NewRequest("GET", "/db/_all_docs?limit=1&channels=true", nil)
+	request, _ = http.NewRequest("GET", "/db."+s+"."+c+"/_all_docs?limit=1&channels=true", nil)
```
torcolvin: switch to SendUserRequest

rest/access_test.go
```diff
@@ -647,7 +675,7 @@ func TestAllDocsAccessControl(t *testing.T) {
 	assert.Equal(t, []string{"Cinemax"}, allDocsResult.Rows[0].Value.Channels)
 
 	// Check all docs startkey option
-	request, _ = http.NewRequest("GET", "/db/_all_docs?startkey=doc5&channels=true", nil)
+	request, _ = http.NewRequest("GET", "/db."+s+"."+c+"/_all_docs?startkey=doc5&channels=true", nil)
```
torcolvin: switch to SendUserRequest

rest/access_test.go
```diff
@@ -661,7 +689,7 @@ func TestAllDocsAccessControl(t *testing.T) {
 	assert.Equal(t, []string{"Cinemax"}, allDocsResult.Rows[0].Value.Channels)
 
 	// Check all docs startkey option with double quote
-	request, _ = http.NewRequest("GET", `/db/_all_docs?startkey="doc5"&channels=true`, nil)
+	request, _ = http.NewRequest("GET", `/db.`+s+`.`+c+`/_all_docs?startkey="doc5"&channels=true`, nil)
```
torcolvin: switch to SendUserRequest

rest/access_test.go
```diff
@@ -675,7 +703,7 @@ func TestAllDocsAccessControl(t *testing.T) {
 	assert.Equal(t, []string{"Cinemax"}, allDocsResult.Rows[0].Value.Channels)
 
 	// Check all docs endkey option
-	request, _ = http.NewRequest("GET", "/db/_all_docs?endkey=doc3&channels=true", nil)
+	request, _ = http.NewRequest("GET", "/db."+s+"."+c+"/_all_docs?endkey=doc3&channels=true", nil)
 	request.SetBasicAuth("alice", "letmein")
 	response = rt.Send(request)
 	RequireStatus(t, response, 200)
```
torcolvin: switch to SendUserRequest

rest/access_test.go
```diff
@@ -689,7 +717,7 @@ func TestAllDocsAccessControl(t *testing.T) {
 	assert.Equal(t, []string{"Cinemax"}, allDocsResult.Rows[0].Value.Channels)
 
 	// Check all docs endkey option
-	request, _ = http.NewRequest("GET", `/db/_all_docs?endkey="doc3"&channels=true`, nil)
+	request, _ = http.NewRequest("GET", `/db.`+s+`.`+c+`/_all_docs?endkey="doc3"&channels=true`, nil)
 	request.SetBasicAuth("alice", "letmein")
 	response = rt.Send(request)
 	RequireStatus(t, response, 200)
```
torcolvin: switch to SendUserRequest

rest/access_test.go
```diff
@@ -703,7 +731,7 @@ func TestAllDocsAccessControl(t *testing.T) {
 	assert.Equal(t, []string{"Cinemax"}, allDocsResult.Rows[0].Value.Channels)
 
 	// Check _all_docs with include_docs option:
-	request, _ = http.NewRequest("GET", "/db/_all_docs?include_docs=true", nil)
+	request, _ = http.NewRequest("GET", "/db."+s+"."+c+"/_all_docs?include_docs=true", nil)
```
torcolvin: switch to SendUserRequest

rest/access_test.go
```diff
@@ -719,7 +747,7 @@ func TestAllDocsAccessControl(t *testing.T) {
 
 	// Check POST to _all_docs:
 	body := `{"keys": ["doc4", "doc1", "doc3", "b0gus"]}`
-	request, _ = http.NewRequest("POST", "/db/_all_docs?channels=true", bytes.NewBufferString(body))
+	request, _ = http.NewRequest("POST", "/db."+s+"."+c+"/_all_docs?channels=true", bytes.NewBufferString(body))
```
torcolvin: switch to SendUserRequest

rest/access_test.go
```diff
@@ -744,7 +772,7 @@ func TestAllDocsAccessControl(t *testing.T) {
 	assert.Equal(t, "", allDocsResult.Rows[3].Value.Rev)
 
 	// Check GET to _all_docs with keys parameter:
-	request, _ = http.NewRequest("GET", `/db/_all_docs?channels=true&keys=%5B%22doc4%22%2C%22doc1%22%2C%22doc3%22%2C%22b0gus%22%5D`, nil)
+	request, _ = http.NewRequest("GET", `/db.`+s+`.`+c+`/_all_docs?channels=true&keys=%5B%22doc4%22%2C%22doc1%22%2C%22doc3%22%2C%22b0gus%22%5D`, nil)
```
torcolvin: switch to SendUserRequest

rest/access_test.go
```diff
@@ -766,7 +794,7 @@ func TestAllDocsAccessControl(t *testing.T) {
 
 	// Check POST to _all_docs with limit option:
 	body = `{"keys": ["doc4", "doc1", "doc3", "b0gus"]}`
-	request, _ = http.NewRequest("POST", "/db/_all_docs?limit=1&channels=true", bytes.NewBufferString(body))
+	request, _ = http.NewRequest("POST", "/db."+s+"."+c+"/_all_docs?limit=1&channels=true", bytes.NewBufferString(body))
```
torcolvin: switch to SendUserRequest

rest/access_test.go
```diff
@@ -781,7 +809,7 @@ func TestAllDocsAccessControl(t *testing.T) {
 	assert.Equal(t, []string{"Cinemax"}, allDocsResult.Rows[0].Value.Channels)
 
 	// Check _all_docs as admin:
-	response = rt.SendAdminRequest("GET", "/db/_all_docs", "")
+	response = rt.SendAdminRequest("GET", "/db."+s+"."+c+"/_all_docs", "")
```
torcolvin: this should be `{{.keyspace}}`

rest/access_test.go
```diff
@@ -815,26 +846,26 @@ func TestChannelAccessChanges(t *testing.T) {
 	assert.NoError(t, a.Save(zegpold))
 
 	// Create some docs that give users access:
-	response := rt.Send(Request("PUT", "/db/alpha", `{"owner":"alice"}`))
+	response := rt.Send(Request("PUT", "/db."+s+"."+c+"/alpha", `{"owner":"alice"}`))
```
torcolvin: This can be:

```
response := rt.SendRequest("PUT", "/{{.keyspace}}/alpha", `{"owner":"alice"}`))
```

rest/access_test.go
```diff
@@ -815,26 +846,26 @@ func TestChannelAccessChanges(t *testing.T) {
...
+	RequireStatus(t, rt.Send(Request("PUT", "/db."+s+"."+c+"/gamma", `{"owner":"zegpold"}`)), 201)
 
-	RequireStatus(t, rt.Send(Request("PUT", "/db/a1", `{"channel":"alpha"}`)), 201)
-	RequireStatus(t, rt.Send(Request("PUT", "/db/b1", `{"channel":"beta"}`)), 201)
-	RequireStatus(t, rt.Send(Request("PUT", "/db/d1", `{"channel":"delta"}`)), 201)
-	RequireStatus(t, rt.Send(Request("PUT", "/db/g1", `{"channel":"gamma"}`)), 201)
+	RequireStatus(t, rt.Send(Request("PUT", "/db."+s+"."+c+"/a1", `{"channel":"alpha"}`)), 201)
+	RequireStatus(t, rt.Send(Request("PUT", "/db."+s+"."+c+"/b1", `{"channel":"beta"}`)), 201)
+	RequireStatus(t, rt.Send(Request("PUT", "/db."+s+"."+c+"/d1", `{"channel":"delta"}`)), 201)
+	RequireStatus(t, rt.Send(Request("PUT", "/db."+s+"."+c+"/g1", `{"channel":"gamma"}`)), 201)
```
torcolvin: swap to a `RestTester` function that can use keyspace

rest/access_test.go
```diff
@@ -873,11 +904,11 @@ func TestChannelAccessChanges(t *testing.T) {
 			"!":     channels.NewVbSimpleSequence(uint64(1)),
 			"zero":  channels.NewVbSimpleSequence(uint64(1)),
 			"gamma": channels.NewVbSimpleSequence(gammaGrantDocSeq),
-		}, zegpold.Channels())
+		}, zegpold.CollectionChannels(s, c))
 
 	// Update a document to revoke access to alice and grant it to zegpold:
 	str := fmt.Sprintf(`{"owner":"zegpold", "_rev":%q}`, alphaRevID)
-	RequireStatus(t, rt.Send(Request("PUT", "/db/alpha", str)), 201)
+	RequireStatus(t, rt.Send(Request("PUT", "/db."+s+"."+c+"/alpha", str)), 201)
```
torcolvin: swap to `SendUserRequest`

rest/access_test.go
```diff
@@ -937,7 +968,7 @@ func TestChannelAccessChanges(t *testing.T) {
 	assert.Equal(t, "a1", changes.Results[0].ID)
 
 	// What happens if we call access() with a nonexistent username?
-	RequireStatus(t, rt.Send(Request("PUT", "/db/epsilon", `{"owner":"waldo"}`)), 201) // seq 10
+	RequireStatus(t, rt.Send(Request("PUT", "/db."+s+"."+c+"/epsilon", `{"owner":"waldo"}`)), 201) // seq 10
```
torcolvin: swap to `SendUserRequest`

rest/access_test.go
```diff
@@ -949,17 +980,17 @@ func TestChannelAccessChanges(t *testing.T) {
...
 	assert.NoError(t, err)
 	assert.True(t, changed)
-	changeCount, err := collection.UpdateAllDocChannels(ctx, false, func(docsProcessed, docsChanged *int) {}, base.NewSafeTerminator())
+	changeCount, err := collectionUser.UpdateAllDocChannels(ctx, false, func(docsProcessed, docsChanged *int) {}, base.NewSafeTerminator())
 	assert.NoError(t, err)
 	assert.Equal(t, 9, changeCount)
 
 	expectedIDs := []string{"beta", "delta", "gamma", "a1", "b1", "d1", "g1", "alpha", "epsilon"}
-	changes, err = rt.WaitForChanges(len(expectedIDs), "/db/_changes", "alice", false)
+	changes, err = rt.WaitForChanges(len(expectedIDs), "/db."+s+"."+c+"/_changes", "alice", false)
```
torcolvin: I think this function can take `{{.keyspace}}` and if it can't, we should make it

rest/blip_api_crud_test.go
```diff
@@ -852,7 +855,7 @@ function(doc, oldDoc) {
 	userWaiter := userDb.NewUserWaiter()
 
 	// Update the user to grant them access to ABC
-	response := rt.SendAdminRequest("PUT", "/db/_user/user1", `{"admin_channels":["ABC"]}`)
+	response := rt.SendAdminRequest("PUT", "/db/_user/user1", `{`+AdminChannelGrant(s, c, `"admin_channels":["ABC"]`)+`}`)
```
torcolvin: swap this to `{{.db}}`

rest/blip_api_crud_test.go
```diff
@@ -1419,6 +1422,9 @@ func TestAccessGrantViaAdminApi(t *testing.T) {
 		connectingPassword: "1234",
 	})
 	defer bt.Close()
+	collection := bt.DatabaseContext().GetSingleDatabaseCollection()
```
torcolvin: We want to remove `GetSingleDatabaseCollection` everywhere from the actual `DatabaseContext` code, but we allow it for rest tester. It can't be removed until some more ISGR changes go through...

That said, blip test stores a restTester so if this is needed: `collection := bt.restTester.GetSingleTestDatabaseCollection()`

rest/blip_api_crud_test.go
```diff
@@ -1429,7 +1435,7 @@ func TestAccessGrantViaAdminApi(t *testing.T) {
 	)
 
 	// Update the user doc to grant access to PBS
-	response := bt.restTester.SendAdminRequest("PUT", "/db/_user/user1", `{"admin_channels":["user1", "PBS"]}`)
+	response := bt.restTester.SendAdminRequest("PUT", "/db/_user/user1", `{`+AdminChannelGrant(s, c, `"admin_channels":["user1", "PBS"]`)+`}`)
```
torcolvin: swap to `{{.db}}`

rest/blip_api_crud_test.go
```diff
@@ -1949,8 +1955,11 @@ func TestRemovedMessageWithAlternateAccess(t *testing.T) {
 
 	rt := NewRestTesterDefaultCollection(t, nil) // CBG-2619: make collection aware
 	defer rt.Close()
+	collection := rt.GetSingleTestDatabaseCollection()
+	c := collection.Name()
+	s := collection.ScopeName()
 
-	resp := rt.SendAdminRequest("PUT", "/db/_user/user", `{"admin_channels": ["A", "B"], "password": "test"}`)
+	resp := rt.SendAdminRequest("PUT", "/db/_user/user", `{`+AdminChannelGrant(s, c, `"admin_channels": ["A", "B"]`)+`, "password": "test"}`)
```
torcolvin: swap to `{{.db}}`

rest/jwt_auth_test.go
```diff
@@ -25,6 +25,7 @@ import (
 
 	"github.com/couchbase/sync_gateway/auth"
 	"github.com/couchbase/sync_gateway/base"
+	ch "github.com/couchbase/sync_gateway/channels"
```
torcolvin: in auth, we use ch to mean channels package, but in rest tester we typically don't, so let's use regular package name.

channels is a very unfortunate naming choice for a concept in go and a package name.

rest/revocation_test.go
```diff
@@ -1212,13 +1223,16 @@ func TestRevocationUserHasDocAccessDocNotFound(t *testing.T) {
...
-	leakyDataStore.SetGetRawCallback(func(s string) error {
-		require.NoError(t, leakyDataStore.Delete("doc"))
-		return nil
-	})
+	dataStores, err := rt.Bucket().ListDataStores()
+	require.NoError(t, err)
+	for _, dataStoreName := range dataStores {
+		if dataStoreName.ScopeName() == s && dataStoreName.CollectionName() == c {
+			fmt.Println(dataStoreName.CollectionName())
+			fmt.Println(dataStoreName)
```
torcolvin: I'd drop print statements here

rest/revocation_test.go
```diff
@@ -1212,13 +1223,16 @@ func TestRevocationUserHasDocAccessDocNotFound(t *testing.T) {
...
-	require.True(t, ok)
-
-	leakyDataStore.SetGetRawCallback(func(s string) error {
-		require.NoError(t, leakyDataStore.Delete("doc"))
-		return nil
-	})
+	dataStores, err := rt.Bucket().ListDataStores()
+	require.NoError(t, err)
+	for _, dataStoreName := range dataStores {
+		if dataStoreName.ScopeName() == s && dataStoreName.CollectionName() == c {
```
torcolvin: prefer `base.IsDefaultCollection` function for name checks

rest/revocation_test.go
```diff
@@ -1441,24 +1458,32 @@ func TestRevocationWithUserXattrs(t *testing.T) {
...
+	collection := rt.GetSingleTestDatabaseCollection()
+	c := collection.Name()
+	s := collection.ScopeName()
+
+	var data sgbucket.DataStore
+	dataStores, _ := rt.Bucket().ListDataStores()
+	for _, dataStoreName := range dataStores {
+		if dataStoreName.ScopeName() == s && dataStoreName.CollectionName() == c {
+			data = rt.Bucket().NamedDataStore(dataStoreName)
+		}
```
torcolvin: if you need the data store, we should think about allowing this a test only function. I'm reluctant to make `DatabaseCollection.dataStore` public but that might be an option also.

rest/revocation_test.go
```diff
@@ -2311,12 +2348,15 @@ func TestBlipRevokeNonExistentRole(t *testing.T) {
...
+	collection := rt.GetSingleTestDatabaseCollection()
+	c := collection.Name()
+	s := collection.ScopeName()
 
 	base.SetUpTestLogging(t, base.LevelTrace, base.KeyAll)
 
 	// 1. Create user with admin_roles including two roles not previously defined (a1 and a2, for example)
 	const testUsername = "bilbo"
-	res := rt.SendAdminRequest(http.MethodPut, fmt.Sprintf("/%s/_user/%s", rt.GetDatabase().Name, testUsername), fmt.Sprintf(`{"name": %q, "password": "test", "admin_roles": ["a1", "a2"], "admin_channels": ["c1"]}`, testUsername))
+	res := rt.SendAdminRequest(http.MethodPut, fmt.Sprintf("/%s/_user/%s", rt.GetDatabase().Name, testUsername), fmt.Sprintf(`{"name": %q, "password": "test", "admin_roles": ["a1", "a2"], `+AdminChannelGrant(s, c, `"admin_channels": ["c1"]`)+`}`, testUsername))
```
torcolvin: you can use `{{.db}}` to avoid a Sprintf here

torcolvin (commented):
I left a lot of comments but most of them are just swap values around and I thought it would be easier to hit them if I commented on them. I am sure that I missed some.

The biggest point is that some of the testing is not covered by a single integration test but actually 4 different runs.

- SG_TEST_USE_DEFAULT_COLLECTION=false,SG_TEST_USE_XATTRS=true (this is the default jenkins run)
- SG_TEST_USE_DEFAULT_COLLECTION=false,SG_TEST_USE_XATTRS=false
- SG_TEST_USE_DEFAULT_COLLECTION=true,SG_TEST_USE_XATTRS=true
- SG_TEST_USE_DEFAULT_COLLECTION=true,SG_TEST_USE_XATTRS=false

It's unfortunate that `WriteDirect` is not covered with xattrs testing, but it's because no one was motivated enough to come up with a way to write metadata previously. I think this would be a great enhancement and I have been bitten by this myself.

I think the only things that are not trivial are the fact the http output should be identical if the default collection is set (SG_TEST_USE_DEFAULT_COLLECTION=true) and different otherwise. I'm especially concerned when it changes names of channels, which is marked slightly. I think you'll have to dig into this a bit more with integration test runs.

The actual default of `SG_TEST_USE_DEFAULT_COLLECTION` is "auto" in jenkins, which means true if `SG_TEST_USE_GSI=true` and false otherwise.

rest/access_test.go
```diff
@@ -35,26 +34,35 @@ func TestPublicChanGuestAccess(t *testing.T) {
...
 
 	resp = rt.SendAdminRequest(http.MethodGet, "/db/_user/GUEST", ``)
 	RequireStatus(t, resp, http.StatusOK)
 	fmt.Println("GUEST user:", resp.Body.String())
-	assert.EqualValues(t, []interface{}{"!"}, resp.GetRestDocument()["all_channels"])
+	if base.IsDefaultCollection(s, c) {
+		assert.EqualValues(t, []interface{}{"!"}, resp.GetRestDocument()["all_channels"])
+	} else {
+		expected := fmt.Sprintf(`map[%s:map[%s:map[all_channels:[!]]]]`, s, c)
+		collectionAccess := resp.GetRestDocument()["collection_access"]
```
adamcfraser: I don't think GetRestDocument, and asserting a match on the stringified property, is the best approach here.  Would be less brittle to unmarshal the response into it's actual struct (auth.PrincipalConfig) and use that to validate the contents.

adamcfraser: With that approach, you could also add a GetExplicitChannels(s, c) helper function on PrincipalConfig, similar to the existing SetExplicitChannels, and let that sort out default/non-default collection handling.

rest/access_test.go
```diff
@@ -375,151 +390,181 @@ func TestForceAPIForbiddenErrors(t *testing.T) {
...
-			err = json.Unmarshal(resp.BodyBytes(), &allChannels)
-			require.NoError(t, err)
-			assert.NotContains(t, allChannels.Channels, "chan2")
+
+			if base.IsDefaultCollection(collection.Name(), collection.ScopeName()) {
+				err = json.Unmarshal(resp.BodyBytes(), &allChannels)
+				require.NoError(t, err)
+				assert.NotContains(t, allChannels.Channels, "chan2")
+			} else {
+				user, err = a.GetUser("NoPerms")
```
adamcfraser: Evaluating the /db/_user response in the default collection case isn't the same as fetching the user directly from the authenticator (which you're doing for the non-default case).  As above, I'd recommend unmarshalling the response into PrincipalConfig, and then add a  GetChannels(s, c) helper method to PrincipalConfig that you can use for the assert, so that every test doesn't need to have the default/non-default collection handling.   There's already a similar method on the role object (CollectionChannels) you could use as an example of the default/non-default handling.

adamcfraser: In general requiring the test to have base.IsDefaultCollection should be a hint that a helper function might be a good idea, can apply that to other tests in this PR.

rest/access_test.go
```diff
@@ -781,7 +809,8 @@ func TestAllDocsAccessControl(t *testing.T) {
 	assert.Equal(t, []string{"Cinemax"}, allDocsResult.Rows[0].Value.Channels)
 
 	// Check _all_docs as admin:
-	response = rt.SendAdminRequest("GET", "/db/_all_docs", "")
+	response = rt.SendAdminRequest("GET", "/db."+s+"."+c+"/_all_docs", "")
```
adamcfraser: Still need to remove line 812.

rest/access_test.go
```diff
@@ -949,17 +983,17 @@ func TestChannelAccessChanges(t *testing.T) {
 	dbc := rt.ServerContext().Database(ctx, "db")
 	database, _ := db.GetDatabase(dbc, nil)
 
-	collection := database.GetSingleDatabaseCollectionWithUser()
+	collectionUser := database.GetSingleDatabaseCollectionWithUser()
```
adamcfraser: A bit of a nit, but 'collectionUser' sounds like this is a user object, when it's actually a DatabaseCollectionWithUser.  Either 'collection' or 'collectionWithUser' would be better.

rest/utilities_testing.go
```diff
@@ -1386,6 +1386,31 @@ func (bt *BlipTester) SendRev(docId, docRev string, body []byte, properties blip
 
 }
 
+// Expects adminChannels of the form `admin_channels":["alpha"]`
+// If scope and collection are non-zero, builds collection access string for the collection
+func AdminChannelGrant(collection *db.DatabaseCollection, adminChannels string) (collectionAdminChannels string) {
```
adamcfraser: I like this helper function, but I think it would be preferable to have this return a marshalled PrincipalConfig, instead of building the string manually, if the motivation is to build a payload for the _user or _role API.  That will better support future modifications to PrincipalConfig.

rest/blip_api_crud_test.go
```diff
@@ -1949,8 +1951,9 @@ func TestRemovedMessageWithAlternateAccess(t *testing.T) {
 
 	rt := NewRestTesterDefaultCollection(t, nil) // CBG-2619: make collection aware
 	defer rt.Close()
+	collection := rt.GetSingleTestDatabaseCollection()
 
-	resp := rt.SendAdminRequest("PUT", "/db/_user/user", `{"admin_channels": ["A", "B"], "password": "test"}`)
+	resp := rt.SendAdminRequest("PUT", "/db/_user/user", `{`+AdminChannelGrant(collection, `"admin_channels": ["A", "B"]`)+`, "password": "test"}`)
```
adamcfraser: Modifying AdminChannelGrant to set password would be more robust than the payload concatenation here, I think.

rest/utilities_testing.go
adamcfraser: I'd also suggest taking the input channel parameter as just []string, to simplify usage.

rest/changestest/changes_api_test.go
```diff
@@ -3979,8 +3985,12 @@ func WriteDirectWithKey(testDb *db.DatabaseContext, key string, channelArray []s
 	syncData.TimeSaved = time.Now()
 	// syncData := fmt.Sprintf(`{"rev":"%s", "sequence":%d, "channels":%s, "TimeSaved":"%s"}`, rev, sequence, chanMap, time.Now())
 
-	_, err := testDb.Bucket.DefaultDataStore().Add(key, 0, db.Body{base.SyncPropertyName: syncData, "key": key})
+	dataStores, _ := testDb.Bucket.ListDataStores()
```
adamcfraser: I think collection.dataStore is the data store you're looking for here.  I think it's reasonable to add a public accessor to make that available, so that WriteDirect can just take DatabaseCollection instead of DatabaseContext.

rest/changestest/changes_api_test.go
```diff
@@ -2063,9 +2062,14 @@ func TestChangesIncludeDocs(t *testing.T) {
 	// Flush the rev cache, and issue changes again to ensure successful handling for rev cache misses
 	testDB.GetSingleDatabaseCollection().FlushRevisionCacheForTest()
 	// Also nuke temporary revision backup of doc_pruned.  Validates that the body for the pruned revision is generated correctly when no longer resident in the rev cache
-	assert.NoError(t, testDB.Bucket.DefaultDataStore().Delete(base.RevPrefix+"doc_pruned:34:2-5afcb73bd3eb50615470e3ba54b80f00"))
+	dataStores, _ := testDB.Bucket.ListDataStores()
```
adamcfraser: As mentioned below, getting the datastore from the DatabaseCollection would be simpler here, instead of trying to find it in the bucket.

rest/utilities_testing.go
```diff
@@ -1386,6 +1386,31 @@ func (bt *BlipTester) SendRev(docId, docRev string, body []byte, properties blip
...
+	collectionName := collection.Name()
+	scopeName := collection.ScopeName()
+	if base.IsDefaultCollection(scopeName, collectionName) {
+		return adminChannels
+	}
+
+	return fmt.Sprintf(`"collection_access":{%q: {%q: {%s}}}`, scopeName, collectionName, adminChannels)
+}
+
+func GetNamedDatastore(dataStores []sgbucket.DataStoreName, collection *db.DatabaseCollection) sgbucket.DataStoreName {
```
adamcfraser: If this is needed, I feel like collection should just implement the sgbucket.DataStoreName interface.

rest/revocation_test.go
```diff
@@ -1442,24 +1453,28 @@ func TestRevocationWithUserXattrs(t *testing.T) {
 	})
 
 	defer rt.Close()
+	collection := rt.GetSingleTestDatabaseCollection()
 
-	collection, err := base.AsCollection(rt.Bucket().DefaultDataStore())
-	if err != nil {
-		t.Skip("Test requires Couchbase Bucket")
-	}
+	var data sgbucket.DataStore
```
adamcfraser: As elsewhere, just get dataStore from the collection.

rest/user_api_test.go
```diff
@@ -262,23 +264,28 @@ func TestUsersAPIDetailsWithLimit(t *testing.T) {
...
 	// GET the user and make sure the result is OK
 	response = rt.SendAdminRequest("GET", "/db/_user/snej", "")
 	RequireStatus(t, response, 200)
 	var body db.Body
 	require.NoError(t, base.JSONUnmarshal(response.Body.Bytes(), &body))
 	assert.Equal(t, "snej", body["name"])
 	assert.Equal(t, "jens@couchbase.com", body["email"])
-	assert.Equal(t, []interface{}{"bar", "foo"}, body["admin_channels"])
-	assert.Equal(t, []interface{}{"!", "bar", "foo"}, body["all_channels"])
+	assert.Contains(t, "bar:1,foo:1", user.CollectionExplicitChannels(s, c).String())
```
adamcfraser: What required the change to Contains and string matching here?  It feels less robust.

adamcfraser (dismissed):
Generally looks good, a few suggestions to simplify the work tests are doing.

rest/utilities_testing.go
torcolvin: I think let's make this a test function that is like

`func GetUserPayload(username, password string, channels []string) string` 

and 

`func GetRolePayload(rolename string, channels []string) string` 

if they share common functionality, which I'm not sure they do, we can do a private function like

`func addChannelsToPrincipal(config *PrincipalConfig, channels []string)`

that does some common work but then returns the marshalled output of the entire payload.

torcolvin (dismissed):
LGTM, can you run through jenkins again just in case?

torcolvin (approved):

PR #5994, opened 2022-12-24
Turn off TCP 'NODELAY' option for WebSockets

snej:
Go's 'net' package sneakily(?) turns on the TCP 'NODELAY' option by default, because it improves throughput in sockets that issue large writes.

However, NODELAY isn't so good with our replicator protocol because there are a lot of small messages, and it causes each one to be sent as a separate packet. This is especially noticeable with the replies to BLIP "rev" messages, which are numerous and tiny.

This commit calls TCPConn.SetNoDelay(false) on sockets used for WebSockets. It's a bit awkward to do because the TCP sockets are opened for us by the 'net/http' package, but fortunately there are hooks to get access to the underlying TCPConn.

## Pre-review checklist

- [x] Link upstream PRs

### DIFF
--- a/base/http_listener.go
+++ b/base/http_listener.go
@@ -54,6 +54,14 @@ func ListenAndServeHTTP(addr string, connLimit uint, certFile, keyFile string, h
 			return nil, nil, err
 		}
 	}
+
+	// Callback that turns off TCP NODELAY option when a client transitions to a WebSocket:
+	connStateFunc := func(clientConn net.Conn, state http.ConnState) {
+		if state == http.StateHijacked {
+			turnOffNoDelay(context.Background(), clientConn)
+		}
+	}
+
 	listener, err := ThrottledListen("tcp", addr, connLimit)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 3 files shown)


rest/config.go
```diff
@@ -453,7 +453,7 @@ func readFromPath(path string, insecureSkipVerify bool) (rc io.ReadCloser, err e
 	messageFormat := "Loading content from [%s] ..."
 	if strings.HasPrefix(path, "http://") || strings.HasPrefix(path, "https://") {
 		base.InfofCtx(context.Background(), base.KeyAll, messageFormat, path)
-		client := base.GetHttpClient(insecureSkipVerify)
+		client := base.GetHttpClientForWebSocket(insecureSkipVerify)
```
adamcfraser: I'm not sure this is needed here (this http client is just used to fetch resources like remote config files).  The usage of GetHttpClient in db/active_replicator.go (for ISGR) looks like it should be updated, though.

adamcfraser (changes requested):

rest/config.go
snej: OK, reverted this one & changed the other.

adamcfraser (approved):

PR #5995, opened 2022-12-28
CBG-2643 Run resync for specific collections

nirav24:
[CBG-2643](https://issues.couchbase.com/browse/CBG-2643)

Describe your PR here...
- run Resync for specific collection rather than all collections
  - collections can be specified in request body
  ```json
  {
    "scopes": {
      "scopeName": ["collection1", "collection2"]
    },
    "regenerate_sequences": false
  }
  ```
  - when body isn't provided, resync will be run for all collections

## Pre-review checklist

- [x] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1381/
- [x] Latest run `GSI=true,xattrs=true https://jenkins.sgwdev.com/job/SyncGateway-Integration/1416/


db/background_mgr_resync_dcp.go
```diff
@@ -32,6 +33,9 @@ type ResyncManagerDCP struct {
 	lock          sync.RWMutex
 }
 
+// resyncPostReqBody contains map of scope names with collection names against which resync needs to run
+type resyncPostReqBody map[string][]string
```
adamcfraser: It would be preferable to define the resync post body with a top level 'scopes' property containing this map, to support future extensibility of the payload.

adamcfraser: In fact, it makes sense to support setting regenerate_sequences=true inside the payload (in addition to being able to set as a query param).

db/background_mgr_resync_dcp.go
```diff
@@ -217,6 +228,41 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]interface
...
+		}
+	} else {
+		hasAllCollections = false
+		collectionManifest, err := bucket.GetCollectionManifest()
+		if err != nil {
+			return nil, hasAllCollections, fmt.Errorf("failed to load collection manifest: %w", err)
+		}
+
+		for scopeName, collNames := range collectionsMap {
+			for _, collName := range collNames {
```
adamcfraser: Should be verifying that the requested collection is part of the database.  Using db.GetDatabaseCollection to do this check will also let you use GetCollectionID() with the response to get the collectionID, instead of fetching the manifest again.

adamcfraser (changes requested):

db/background_mgr_resync_dcp.go
nirav24: Which one will take higher precedence, the query param or one in the request body? or we regenerate sequences if either one is true?

Edit: I made changes to regenerate sequences if either (query param/request body) is set to true. let me know if that is incorrect and i will make necessary changes

docs/api/admin.yaml
```diff
@@ -100,8 +100,8 @@ paths:
     $ref: './paths/admin/{keyspace}~_config~sync.yaml'
   '/{keyspace}/_config/import_filter':
     $ref: './paths/admin/{keyspace}~_config~import_filter.yaml'
-  '/{keyspace}/_resync':
-    $ref: './paths/admin/{keyspace}~_resync.yaml'
+  '/{db}/_resync':
+    $ref: './paths/admin/{db}~_resync.yaml'
```
adamcfraser, line 104: Can you update _resync.yaml to include the POST payload used to specify collections?

rest/api.go
```diff
@@ -274,8 +274,13 @@ func (h *handler) handleGetResync() error {
 }
 
 func (h *handler) handlePostResync() error {
+
 	action := h.getQuery("action")
 	regenerateSequences, _ := h.getOptBoolQuery("regenerate_sequences", false)
+	body, err := h.readBody()
```
adamcfraser, line 287: I'd prefer to do all the payload parsing/query parameter handling here in rest, instead of passing the full payload body through as 'collections' to ResyncManager.Start(), to keep all the REST API-specific handling in the rest package.

adamcfraser (changes requested):
Generally looks good, comments on one implementation detail and request for docs update.

rest/api.go
```diff
@@ -290,6 +303,7 @@ func (h *handler) handlePostResync() error {
 			err := h.db.ResyncManager.Start(h.ctx(), map[string]interface{}{
 				"database":            h.db,
 				"regenerateSequences": regenerateSequences,
+				"collections":         resyncPostReqBody,
```
adamcfraser: I think I wasn't clear in my previous review, but I meant that 'collections' should be getting passed down to ResyncManagerStart as just a set of collections, as opposed to the REST payload.  i.e. the ResyncManager.Start should define the options used, and should be independent of this particular usage from the REST API.  So define ResyncPostReqBody here in the REST package, and define the struct for collections only in db.  It's fine if that ends up looking like:
```
in db:
type ResyncCollections map[string][]string

in rest:
type ResyncPostReqBody struct {
	Scope               db.ResyncCollections `json:"scopes,omitempty"`
	RegenerateSequences bool                `json:"regenerate_sequences,omitempty"`
}
```
The main point is just to avoid defining the REST payload down in the db package, to better support other use cases in future (as well as avoid the redundant definition of regenerateSequences).

adamcfraser (changes requested):

rest/api.go
nirav24: oh I see. Fixed it

adamcfraser (commented):
Changes look good, fine to merge once CI failures are sorted out

adamcfraser (approved):

PR #5996, opened 2022-12-30
Make rt.GetSingleKeyspace return only db name when default collection for regression testing

bbrks:
Not strictly nessesary, as SG should handle both `/db/` and `/db._default._default/` equally, but doing this covers backwards compatibility/regression for utils using `rt.GetSingleKeyspace` (e.g. `rt.PutDoc`)

The only additional coverage we could do on top of this is to make this return _both_ formats for the default case.

- n/a (testing only)

### DIFF
--- a/rest/utilities_testing.go
+++ b/rest/utilities_testing.go
@@ -2103,6 +2103,12 @@ func (sc *ServerContext) suspendDatabase(t *testing.T, ctx context.Context, dbNa
 
 // getRESTkeyspace returns a keyspace for REST URIs
 func getRESTKeyspace(_ testing.TB, dbName string, collection *db.DatabaseCollection) string {
+	if base.IsDefaultCollection(collection.ScopeName(), collection.Name()) {
+		// for backwards compatibility (and user-friendliness),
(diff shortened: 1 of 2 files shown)


rest/utilities_testing.go
```diff
@@ -2117,13 +2117,19 @@ func (rt *RestTester) GetKeyspaces() []string {
...
 
-// getSingleKeyspace the name of the keyspace if there is only one test collection on one database.
+// GetSingleKeyspace the name of the keyspace if there is only one test collection on one database.
 func (rt *RestTester) GetSingleKeyspace() string {
 	db := rt.GetDatabase()
-	require.Equal(rt.TB, 1, len(db.CollectionByID), "Database is configured with more collection")
-	for _, collection := range db.CollectionByID {
+	require.Equal(rt.TB, 1, len(db.CollectionByID), "Database must be configured with only one collection to use this function")
+	for id, collection := range db.CollectionByID {
+		if id == base.DefaultCollectionID {
```
adamcfraser: Is it worth pushing this handling down into getRESTKeyspace so it gets used by GetKeyspaces as well?

torcolvin (approved):

PR #6009, opened 2023-01-09
CBG-2635: Add multi collection channel grant test

mohammed-madi:
CBG-2635

Describe your PR here...

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1351/

### DIFF
--- a/rest/api_collections_test.go
+++ b/rest/api_collections_test.go
@@ -269,6 +269,78 @@ func TestMultiCollectionDCP(t *testing.T) {
 	// require.NoError(t, rt.WaitForDoc(docID))
 }
 
+func TestMultiCollectionDynamicChannelAccess(t *testing.T) {
+	base.TestRequiresCollections(t)
+	//base.SetUpTestLogging(t, base.LevelDebug, base.KeyAll)
+
+	if !base.TestUseXattrs() {
+		t.Skip("Test relies on import - needs xattrs")
+	}
+
+	tb := base.GetTestBucket(t)
+	defer tb.Close()
+
+	scopesConfig := GetCollectionsConfig(t, tb, 2)
+	dataStoreNames := GetDataStoreNamesFromScopesConfig(scopesConfig)
+	c1SyncFunction := `function(doc) {access(doc.username, [doc.grant1,doc.grant2]);
+                  channel(doc.chan);
+            }`
+
+	scopesConfig[dataStoreNames[0].ScopeName()].Collections[dataStoreNames[0].CollectionName()] = CollectionConfig{SyncFn: &c1SyncFunction}
+	scopesConfig[dataStoreNames[1].ScopeName()].Collections[dataStoreNames[1].CollectionName()] = CollectionConfig{SyncFn: &c1SyncFunction}
+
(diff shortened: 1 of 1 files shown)


rest/api_collections_test.go
```diff
@@ -269,6 +269,78 @@ func TestMultiCollectionDCP(t *testing.T) {
 	// require.NoError(t, rt.WaitForDoc(docID))
 }
 
+func TestMultiCollectionDynamicChannelAccess(t *testing.T) {
+	base.TestRequiresCollections(t)
+	//base.SetUpTestLogging(t, base.LevelDebug, base.KeyAll)
+
+	if !base.TestUseXattrs() {
+		t.Skip("Test relies on import - needs xattrs")
+	}
```
bbrks: I can't see anything in the test writing directly to the bucket/collection/datastore. It all goes through SG, which makes me think this isn't nessesary.

rest/api_collections_test.go
```diff
@@ -269,6 +269,78 @@ func TestMultiCollectionDCP(t *testing.T) {
...
+	resp = rt.SendUserRequestWithHeaders(http.MethodGet, "/{{.keyspace1}}/testDocBarA", "", nil, "userB", "letmein")
+	RequireStatus(t, resp, http.StatusOK)
+	resp = rt.SendUserRequestWithHeaders(http.MethodGet, "/{{.keyspace1}}/testDocBarA", "", nil, "userAB", "letmein")
+	RequireStatus(t, resp, http.StatusForbidden)
+	resp = rt.SendUserRequestWithHeaders(http.MethodGet, "/{{.keyspace2}}/testDocBazA", "", nil, "userA", "letmein")
+	RequireStatus(t, resp, http.StatusOK)
+	resp = rt.SendUserRequestWithHeaders(http.MethodGet, "/{{.keyspace2}}/testDocBazA", "", nil, "userB", "letmein")
+	RequireStatus(t, resp, http.StatusForbidden)
+
+}
```
bbrks, line 341: It's kind of confusing having test data like this where the user names don't match what channels they are being granted (e.g. `userAB` is not being granted any channels, rather than A and B)

My preference would be to either change users to actual "people"/actors, like `alice`, `bob`, `charlie`... or `user1`, `user2`,... etc. or to make the names match their grants (but that itself might change over time in more complicated tests)

bbrks (changes requested):
minor change to naming of users to make it less confusing but otherwise seems fine

rest/api_collections_test.go
mohammed-madi: [5b2f320](https://github.com/couchbase/sync_gateway/pull/6009/commits/5b2f3202cb1f09cd0f83b94dd7186af73eebdb63)

rest/api_collections_test.go
mohammed-madi: [5b2f320](https://github.com/couchbase/sync_gateway/pull/6009/commits/5b2f3202cb1f09cd0f83b94dd7186af73eebdb63)

bbrks (approved):

PR #6010, opened 2023-01-10
CBG-2619: make blip tests in rest package single named collection aware

gregns1:
CBG-2619

Changes to Blip tests to make them collection aware. Includes some helper functions to do this, one of which is in my collection channels PR so will conflict depending on which is merged first. All tests that were labelled as part fpo this ticket have changed. `SetupSGRPeers` was part of this ticket to make collection aware but there is still work to do on [CBG-2491](https://issues.couchbase.com/browse/CBG-2491) to make this possible. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1421/
- [x]  `GSI=true. xattrs=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1370/
- [x]  `GSI=false, xattrs=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1371/
- [x]  `GSI=false, xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1422/

### DIFF
--- a/rest/attachment_test.go
+++ b/rest/attachment_test.go
@@ -2627,7 +2627,7 @@ func TestAttachmentDeleteOnExpiry(t *testing.T) {
 
 }
 func TestUpdateExistingAttachment(t *testing.T) {
-	rt := NewRestTester(t, &RestTesterConfig{ // CBG-2619
+	rt := NewRestTester(t, &RestTesterConfig{
 		GuestEnabled: true,
 	})
 	defer rt.Close()
@@ -2698,7 +2698,7 @@ func TestUpdateExistingAttachment(t *testing.T) {
 // TestPushUnknownAttachmentAsStub sets revpos to an older generation, for an attachment that doesn't exist on the server.
 // Verifies that getAttachment is triggered, and attachment is properly persisted.
 func TestPushUnknownAttachmentAsStub(t *testing.T) {
-	rt := NewRestTester(t, &RestTesterConfig{ // CBG-2619
+	rt := NewRestTester(t, &RestTesterConfig{
 		GuestEnabled: true,
 	})
 	defer rt.Close()
@@ -2967,7 +2967,7 @@ func TestPutInvalidAttachment(t *testing.T) {
 // validates that proveAttachment isn't being invoked when the attachment is already present and the
 // digest doesn't change, regardless of revpos.
 func TestCBLRevposHandling(t *testing.T) {
-	rt := NewRestTester(t, &RestTesterConfig{ // CBG-2619
+	rt := NewRestTester(t, &RestTesterConfig{
 		GuestEnabled: true,
 	})
 	defer rt.Close()
--- a/rest/attachment_test.go
+++ b/rest/attachment_test.go
@@ -2632,11 +2632,9 @@ func TestUpdateExistingAttachment(t *testing.T) {
 	})
 	defer rt.Close()
 
-	btc, btcCollection, err := BlipClientInitialization(t, rt, nil)
+	btc, err := NewBlipTesterClientOptsWithRT(t, rt, nil)
 	assert.NoError(t, err)
 	defer btc.Close()
-	_, err = btc.pushReplication.bt.BlipCollectionSetup(rt)
-	require.NoError(t, err)
 
 	var doc1Body db.Body
 	var doc2Body db.Body
@@ -2656,19 +2654,19 @@ func TestUpdateExistingAttachment(t *testing.T) {
 	err = json.Unmarshal(doc2Bytes, &doc2Body)
 	assert.NoError(t, err)
 
-	err = btcCollection.StartOneshotPull()
+	err = btc.StartOneshotPull()
 	assert.NoError(t, err)
-	_, ok := btcCollection.WaitForRev("doc1", "1-ca9ad22802b66f662ff171f226211d5c")
+	_, ok := btc.WaitForRev("doc1", "1-ca9ad22802b66f662ff171f226211d5c")
 	require.True(t, ok)
-	_, ok = btcCollection.WaitForRev("doc2", "1-ca9ad22802b66f662ff171f226211d5c")
+	_, ok = btc.WaitForRev("doc2", "1-ca9ad22802b66f662ff171f226211d5c")
 	require.True(t, ok)
 
 	attachmentAData := base64.StdEncoding.EncodeToString([]byte("attachmentA"))
 	attachmentBData := base64.StdEncoding.EncodeToString([]byte("attachmentB"))
 
-	revIDDoc1, err := btcCollection.PushRev("doc1", doc1Body["rev"].(string), []byte(`{"key": "val", "_attachments": {"attachment": {"data": "`+attachmentAData+`"}}}`))
+	revIDDoc1, err := btc.PushRev("doc1", doc1Body["rev"].(string), []byte(`{"key": "val", "_attachments": {"attachment": {"data": "`+attachmentAData+`"}}}`))
 	require.NoError(t, err)
-	revIDDoc2, err := btcCollection.PushRev("doc2", doc2Body["rev"].(string), []byte(`{"key": "val", "_attachments": {"attachment": {"data": "`+attachmentBData+`"}}}`))
+	revIDDoc2, err := btc.PushRev("doc2", doc2Body["rev"].(string), []byte(`{"key": "val", "_attachments": {"attachment": {"data": "`+attachmentBData+`"}}}`))
 	require.NoError(t, err)
 
 	err = rt.WaitForRev("doc1", revIDDoc1)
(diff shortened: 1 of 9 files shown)


rest/attachment_test.go
```diff
@@ -2990,20 +2997,19 @@ func TestCBLRevposHandling(t *testing.T) {
 	err = json.Unmarshal(doc2Bytes, &doc2Body)
 	assert.NoError(t, err)
 
-	err = btc.StartOneshotPull()
```
bbrks, line 2993: I wonder if we should also remove these methods on the client (and force future tests to use `btc.DefaultCollection().StartOneShotPull()` if that's what they're actually intending to do)

rest/blip_client_test.go
```diff
@@ -1135,3 +1143,60 @@ func (btc *BlipTesterCollectionClient) sendPushMsg(msg *blip.Message) error {
...
 }
+
+func (btc *BlipTesterClient) BlipClientCollectionSetup(collection *db.DatabaseCollection) (*BlipTesterCollectionClient, error) {
+	var btcCollection *BlipTesterCollectionClient
+	scopeAndCollectionKey := strings.Join([]string{collection.ScopeName(), collection.Name()}, base.ScopeCollectionSeparator)
+	if !base.IsDefaultCollection(collection.ScopeName(), collection.Name()) {
+		checkpointID1 := "checkpoint1"
+		getCollectionsRequest, err := db.NewGetCollectionsMessage(db.GetCollectionsRequestBody{
+			CheckpointIDs: []string{checkpointID1},
+			Collections:   []string{fmt.Sprintf("%s.%s", collection.ScopeName(), collection.Name())},
```
bbrks: Can this use `scopeAndCollectionKey` created above ?

rest/blip_client_test.go
```diff
@@ -1135,3 +1143,60 @@ func (btc *BlipTesterCollectionClient) sendPushMsg(msg *blip.Message) error {
...
+			return nil, err
+		}
+		err = btc.pushReplication.sendMsg(getCollectionsRequest)
+		if err != nil {
+			return nil, err
+		}
+		resp := getCollectionsRequest.Response()
+		if resp == nil {
+			return nil, fmt.Errorf("error in getCollections response")
+		}
```
bbrks: We should read the response body to make sure the other side of the replication agreed to replicate with the given collection.

There's detail about the response here or here:
- https://github.com/couchbase/couchbase-lite-core/blob/master/modules/docs/pages/replication-protocol.adoc#getCollections
- https://docs.google.com/document/d/1kbPUuWgtnAgV9Oj55_soWmIPA8f9v3RBrjMXVvbn4p4/edit#heading=h.v2cz3nwptjc4

rest/blip_client_test.go
```diff
@@ -1135,3 +1143,60 @@ func (btc *BlipTesterCollectionClient) sendPushMsg(msg *blip.Message) error {
...
+		btcCollection, err = btc.Collection(scopeAndCollectionKey)
+		if err != nil {
+			return nil, err
+		}
+		_, found := btc.pushReplication.WaitForMessage(getCollectionsRequest.SerialNumber())
+		if !found {
+			return nil, fmt.Errorf("error waiting for getCollections serial number to be stored by the replicator")
+		}
```
bbrks: I don't think we need to do this, as BLIP blocks `.Response()` until recieved.

rest/utilities_testing.go
```diff
@@ -1386,6 +1369,41 @@ func (bt *BlipTester) SendRev(docId, docRev string, body []byte, properties blip
...
+	collectionName := collection.Name()
+	scopeName := collection.ScopeName()
+	if base.IsDefaultCollection(scopeName, collectionName) {
+		return adminChannels
+	}
+
+	return fmt.Sprintf(`"collection_access":{%q: {%q: {%s}}}`, scopeName, collectionName, adminChannels)
+}
+
+func (bt *BlipTester) BlipCollectionSetup(collection *db.DatabaseCollection) (blip.Properties, error) {
```
bbrks: This appears to be pretty much the same as `BlipTesterClient.BlipClientCollectionSetup` so I'm wondering if we can share the code

rest/blip_client_test.go
```diff
@@ -1135,3 +1143,60 @@ func (btc *BlipTesterCollectionClient) sendPushMsg(msg *blip.Message) error {
...
+	} else {
+		btcCollection = btc.DefaultCollection()
+	}
+	return btcCollection, nil
+}
+
+func BlipClientInitialization(t *testing.T, rt *RestTester, collection *db.DatabaseCollection, options *BlipTesterClientOpts) (*BlipTesterClient, error) {
```
bbrks: It looks like all tests require a similar set of steps to get a BLIP test able to work with a single collection, which is going to be the common case.

1. `NewRestTester`
2. `rt.GetSingleTestDatabaseCollection`
3. `BlipClientInitialization(t, rt, collection, nil)`
4. `client.BlipClientCollectionSetup(collection)`

Given that most (all?) tests require these steps and in this order, I'm wondering if it makes sense to at least merge the last two or three like this:

1. `rt := NewRestTester`
2. `client, btcCollection, err := NewBlipClientWithSingleTestCollection(rt)`

So you could leave RestTester creation to the test (given there are different configs required), and then a utility function to give you a blip client and single collection from a RestTester that is passed in, which does the job of `BlipClientInitialization` and `BlipClientCollectionSetup` in one go.

I think maybe this falls apart when you need to use `BlipTesterSpec` but spec could be an optional parameter passed into the constructor.

rest/utilities_testing.go
```diff
@@ -1386,6 +1369,41 @@ func (bt *BlipTester) SendRev(docId, docRev string, body []byte, properties blip
...
 
+func AdminChannelGrant(collection *db.DatabaseCollection, adminChannels string) (collectionAdminChannels string) {
+	collectionName := collection.Name()
+	scopeName := collection.ScopeName()
+	if base.IsDefaultCollection(scopeName, collectionName) {
+		return adminChannels
+	}
+
+	return fmt.Sprintf(`"collection_access":{%q: {%q: {%s}}}`, scopeName, collectionName, adminChannels)
+}
```
bbrks: Might as well point you to this comment on the other PR :) 

https://github.com/couchbase/sync_gateway/pull/5990#discussion_r1067379377

bbrks (dismissed):
This mostly looks good. I have left some suggestions about how to simplify setup for the common use-case (single test collection)

rest/utilities_testing.go
gregns1: I agree I'd much prefer that. Having trouble getting my head around how to approach that with passing BlipTester or BliptesterClient type into a singular function/method? 
Could declare the function to take both types as parameters (`BlipTesterCollectionSetup(*BlipTester, *BlipTesterClient, collection)` ) then just nil BlipTester or BlipClient respectively when using it in different tests? As far as I know there is no way to get BlipTester type from BlipTesterClient or vice versa?

bbrks: There's two `BlipTesters` inside a `BlipTesterClient` which you might be able to use (`BlipTesterClient.pushReplication.bt`) - you're already chosing `pushReplication` inside the other method to send the getCollections request so in both cases I think you have a `blip.Sender` type.

rest/blip_api_attachment_test.go
```diff
@@ -345,6 +347,9 @@ func TestPutAttachmentViaBlipGetViaRest(t *testing.T) {
 	getAttachmentRequest.SetProfile(db.MessageGetAttachment)
 	getAttachmentRequest.Properties[db.GetAttachmentDigest] = input.attachmentDigest
 	getAttachmentRequest.Properties[db.GetAttachmentID] = input.docId
+	if !base.IsDefaultCollection(collection.ScopeName(), collection.Name()) {
```
torcolvin: use `bt.addCollectionProperty(getAttachmentRequest)` here so we can make changes in a central location

rest/blip_api_delta_sync_test.go
```diff
@@ -48,13 +48,15 @@ func TestBlipDeltaSyncPushAttachment(t *testing.T) {
 	btc, err := NewBlipTesterClientOptsWithRT(t, rt, nil)
 	require.NoError(t, err)
 	defer btc.Close()
+	require.NoError(t, err)
```
torcolvin: remove this duplicate line

rest/blip_api_delta_sync_test.go
```diff
@@ -820,15 +816,18 @@ func TestBlipDeltaSyncPush(t *testing.T) {
 		}},
 		GuestEnabled: true,
 	}
-	rt := NewRestTesterDefaultCollection(t, // CBG-2619: make collection aware
+	rt := NewRestTester(t,
 		&rtConfig)
 	defer rt.Close()
+	collection := rt.GetSingleTestDatabaseCollection()
+	var msg *blip.Message
+	var ok bool
```
torcolvin: are these unused?

rest/blip_api_delta_sync_test.go
```diff
@@ -932,9 +931,12 @@ func TestBlipNonDeltaSyncPush(t *testing.T) {
...
```
torcolvin: are these needed?

torcolvin (approved):

PR #6018, opened 2023-01-13
CBG-2642: Fix NamedDatastore panic when the specified scope/collection don't exist

mohammed-madi:
CBG-2642

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1404/


base/collection.go
```diff
@@ -621,14 +624,19 @@ func (b *GocbV2Bucket) DefaultDataStore() sgbucket.DataStore {
...
-func (b *GocbV2Bucket) NamedDataStore(name sgbucket.DataStoreName) sgbucket.DataStore {
+func (b *GocbV2Bucket) NamedDataStore(name sgbucket.DataStoreName) (sgbucket.DataStore, error) {
+	err := WaitForNoError(func() error {
+		_, err := NewCollection(
+			b, b.bucket.Scope(name.ScopeName()).Collection(name.CollectionName()))
+		return err
+	})
+	if err != nil {
+		return nil, fmt.Errorf("attempting to create/update database with a scope/collection that is not found: %w", ErrAuthError)
+	}
```
bbrks: I am not sure why this function needs to call `NewCollection` twice?

Also not sure the error retry is best suited here specifically. I'd expected it up in `_getOrAddDatabaseFromConfig` so that NamedDataStore itself is retried before calling `WaitUntilDataStoreExists`

Maybe those two can be combined into a new function if it's a common pattern to do?

bbrks (changes requested):
Just a question about the retry stuff inside `(b *GocbV2Bucket) NamedDataStore`, otherwise makes sense.

base/collection.go
mohammed-madi: If the retry loop is done before the `WaitUntilDataStoreExists` in `_getOrAddDatabaseFromConfig`, is it not better to call `NamedDataStore` after `WaitUntilDataStoreExists` since the retry loop would be doing the same thing as `WaitUntilDataStoreExists`?

mohammed-madi: https://github.com/couchbase/sync_gateway/pull/6018/commits/6e9cef6ef7089a499dd1e20509f45af572390485

base/collection.go
```diff
@@ -621,14 +624,14 @@ func (b *GocbV2Bucket) DefaultDataStore() sgbucket.DataStore {
...
 
 // NamedDataStore returns a collection on a bucket within the given scope and collection.
-func (b *GocbV2Bucket) NamedDataStore(name sgbucket.DataStoreName) sgbucket.DataStore {
+func (b *GocbV2Bucket) NamedDataStore(name sgbucket.DataStoreName) (sgbucket.DataStore, error) {
 	c, err := NewCollection(
 		b,
 		b.bucket.Scope(name.ScopeName()).Collection(name.CollectionName()))
 	if err != nil {
-		panic(err)
+		return nil, fmt.Errorf("attempting to create/update database with a scope/collection that is not found: %w", ErrAuthError)
```
bbrks: This method may not always be called in the context of creating/updating a database.
I think we should keep the message specific to collection not found and the caller of this method can decide what extra context to add to the message if necessary.

We're also losing the actual contents of the underlying error `err`, here because we wanted the base error to return the ErrAuthError type for uniform error handling for non-existent/inaccessible bucket.

What happens if `err` is non-nil but not because the collection was not found? Should we be handling the not-found case specifically to return the AuthError and return `err` for everything else?

rest/server_context.go
```diff
@@ -483,9 +483,15 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
 		}
 
 		for scopeName, scopeConfig := range config.Scopes {
+			var dataStore sgbucket.DataStore
 			for collectionName, _ := range scopeConfig.Collections {
```
bbrks: Scoping this outside of the loop feels error-prone and unnessesary. It would allow someone to use `dataStore` on the last iterated collection outside of the loop.

```suggestion
			for collectionName, _ := range scopeConfig.Collections {
				var dataStore sgbucket.DataStore
```

bbrks (changes requested):

rest/api_collections_test.go
```diff
@@ -464,6 +464,28 @@ func TestCollectionsSGIndexQuery(t *testing.T) {
...
 	require.NoError(t, err)
 }
+func TestCollectionsPutDBInexistentCollection(t *testing.T) {
+	base.TestRequiresCollections(t)
+	base.SetUpTestLogging(t, base.LevelDebug, base.KeyAll)
+
+	if !base.TestUseXattrs() {
+		t.Skip("Test relies on import - needs xattrs")
+	}
+
```
bbrks: Are these (I noticed the same in other PRs) coming from copy/paste of existing tests or somewhere else? We should try to minimise and simplify tests where possible.

This test doesn't write to the bucket/use import. The vast majority of tests won't use import as they do their document writes through Sync Gateway, and not Couchbase Server's bucket.

```suggestion
```

bbrks (changes requested):
Looks fine, but test is being skipped with non-xattrs but it doesn't use import at all. Seems like an unnecessary restriction.

rest/api_collections_test.go
mohammed-madi: https://github.com/couchbase/sync_gateway/pull/6018/commits/19ea492a3a4bafab3e78c85a81fffe8f2e122287

bbrks (approved):

PR #6020, opened 2023-01-13
CBG-2675 make sure no replication occurs if there is no default db

torcolvin:
Prevent replication if there is no default db. The real fix removing `bh.collection = bh.db.GetSingleDatabaseCollectionWithUser()` which is used for ISGR at this time.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1383/


rest/blip_client_test.go
```diff
@@ -1105,25 +1060,29 @@ func (btc *BlipTesterCollectionClient) addCollectionProperty(msg *blip.Message)
...
+	return msg
+}
 
-	idx, err := strconv.Atoi(collectionIdx)
-	if err != nil {
-		return "", err
+func (btc *BlipTesterClient) getCollectionClientFromMessage(msg *blip.Message) *BlipTesterCollectionClient {
+	collectionIdx, exists := msg.Properties[db.BlipCollection]
+	if !exists {
+		require.NotNil(btc.rt.TB, btc.nonCollectionAwareClient)
```
adamcfraser, line 1092: I'm on the fence about using require this deep into a testing utility, but if it significantly simplifies usage then I could be convinced.

rest/utilities_testing.go
```diff
@@ -2133,3 +2197,17 @@ func (rt *RestTester) GetSingleKeyspace() string {
...
 	return ""
 }
+
+// getCollectionsForBLIP
+func (rt *RestTester) getCollectionsForBLIP() []string {
+	db := rt.GetDatabase()
+	var collections []string
+	for _, collection := range db.CollectionByID {
+		if base.IsDefaultCollection(collection.ScopeName(), collection.Name()) {
+			continue
```
adamcfraser: Should we always be skipping the default collection here?

torcolvin: Not necessarily, we should probably only skip the default collection if it's the only collection.

rest/blip_client_test.go
torcolvin: Many of it's callers panic when they call errors here, so I'd rather have a require.

db/blip_handler.go
```diff
@@ -141,21 +141,14 @@ func collectionBlipHandler(next blipHandlerFunc) blipHandlerFunc {
 	return func(bh *blipHandler, bm *blip.Message) error {
 		collectionIndexStr, ok := bm.Properties[BlipCollection]
 		if !ok {
+			if !bh.db.hasDefaultCollection() {
+				return base.HTTPErrorf(http.StatusBadRequest, "Method requires passing a collection property and a prior GetCollections message")
+			}
```
adamcfraser, line 146: Should it be using the default collection and not bh.db.singleCollection below?  Or is that part of a future PR, and today "hasDefaultCollection" always means singleCollection==defaultCollection?

db/database.go
```diff
@@ -2239,7 +2245,7 @@ func (dbc *Database) GetDatabaseCollectionWithUser(scopeName, collectionName str
 
 // GetDatabaseCollection returns a collection if one exists, otherwise error.
 func (dbc *DatabaseContext) GetDatabaseCollection(scopeName, collectionName string) (*DatabaseCollection, error) {
-	if base.IsDefaultCollection(scopeName, collectionName) && dbc.onlyDefaultCollection() {
+	if base.IsDefaultCollection(scopeName, collectionName) && dbc.OnlyDefaultCollection() {
```
adamcfraser: Why do we require OnlyDefaultCollection here?

db/database_stats.go
```diff
@@ -21,7 +21,7 @@ package db
 
 // Update database-specific stats that are more efficiently calculated at stats collection time
 func (db *DatabaseContext) UpdateCalculatedStats() {
-	if !db.onlyDefaultCollection() {
+	if !db.OnlyDefaultCollection() {
```
adamcfraser: Is this needed?  Seems like hasDefaultCollection may be sufficient.

rest/utilities_testing.go
```diff
@@ -1320,17 +1326,77 @@ func createBlipTesterWithSpec(tb testing.TB, spec BlipTesterSpec, rt *RestTester
 		return nil, err
 	}
 
+	collections := bt.restTester.getCollectionsForBLIP()
+	if !spec.skipCollectionsInitialization && len(collections) > 0 {
+		bt.initializeCollections(collections)
+	}
```
adamcfraser, line 1332: The interaction between skipCollectionsInitialization and getCollectionsForBLIP doesn't seem like it covers all the cases.  It seems like we'd need the ability to test:

1. named collections without getCollections message (negative test)
2. named collections with getCollections message (positive test)
3. default collection only without getCollections message (positive test)
4. default collection only with getCollectionsMessage (positive test)
It looks to me like number 4 isn't supported with this approach?

adamcfraser (changes requested):

db/blip_handler.go
torcolvin: It will be part of the ISGR PR forthcoming after this one.

db/database.go
torcolvin: legacy code, i only changed this because I wanted to make `OnlyDefaultCollection` public for `rest`.

db/database_stats.go
torcolvin: legacy code, i only changed this because I wanted to make OnlyDefaultCollection public for rest.

db/blip_handler.go
torcolvin: This is going to be fixed with ISGR PR later this week.

rest/utilities_testing.go
torcolvin: In general we only test 1 and 4 in a specific test wit h`skipInitialization` in `TestBlipGetCollections` where we `skipCollectionInitialization`.

adamcfraser (approved):

PR #6028, opened 2023-01-19
CBG-2685: change context on database operations

gregns1:
CBG-2685

Simple change to pass a background context into functions that start carrying out actions to create/update a database.
This will prevent the request scoped context cancellation cancelling any operations further on down the stack that use the context off the database. 
I have done this through creating a struct and a helper function to initialize the struct. This is to make it explicit what is required for these functions so future enhancements in the area can be sure to use the same context for functions on the database. 

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1414/

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -10,7 +10,6 @@ package rest
 
 import (
 	"bytes"
-	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -38,13 +37,9 @@ const paramDeleted = "deleted"
 
 // ////// DATABASE MAINTENANCE:
 
-type nonCancellableContext struct {
-	ctx context.Context
-}
-
 // "Create" a database (actually just register an existing bucket)
 func (h *handler) handleCreateDB() error {
-	contextNoCancel := createNonCancelCtx()
+	contextNoCancel := base.NewNonCancelCtx()
 	h.assertAdminOnly()
 	dbName := h.PathVar("newdb")
 	config, err := h.readSanitizeDbConfigJSON()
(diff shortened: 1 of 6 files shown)


rest/admin_api.go
```diff
@@ -97,7 +103,7 @@ func (h *handler) handleCreateDB() error {
 				"Duplicate database name %q", dbName)
 		}
 
-		_, err = h.server._applyConfig(h.ctx(), loadedConfig, true, false)
+		_, err = h.server._applyConfig(contextNoCancel.ctx, loadedConfig, true, false)
```
bbrks: Would be good to change these functions to take an actual `nonCancellableContext` type instead of a `context.Context` - at least then we're enforcing the use of a non-cancellable context.

rest/admin_api.go
```diff
@@ -59,7 +65,7 @@ func (h *handler) handleCreateDB() error {
 			return base.HTTPErrorf(http.StatusBadRequest, err.Error())
 		}
 
-		if err := config.validate(h.ctx(), validateOIDC); err != nil {
+		if err := config.validate(contextNoCancel.ctx, validateOIDC); err != nil {
```
bbrks: This one could remain as `h.ctx()` since it's not setting anything up, might be useful to have the logging information from the request in here.

rest/admin_api.go
```diff
@@ -37,8 +38,13 @@ const paramDeleted = "deleted"
 
 // ////// DATABASE MAINTENANCE:
 
+type nonCancellableContext struct {
+	ctx context.Context
+}
```
bbrks: I'd put this type in `base` and have it exported (along with a description of why it needs to exist).
Seems like something that may be useful outside of just the `rest` package.

rest/admin_api.go
```diff
@@ -149,6 +155,13 @@ func (h *handler) handleCreateDB() error {
 	return base.HTTPErrorf(http.StatusCreated, "created")
 }
 
+func createNonCancelCtx() nonCancellableContext {
```
bbrks: It's typical to use `New_____()` for a "constructor" function names.

https://google.github.io/styleguide/go/decisions#package-vs-exported-symbol-name 
or
https://go.dev/doc/effective_go#composite_literals

bbrks (changes requested):

bbrks (approved):

PR #6030, opened 2023-01-20
CBG- 2638: Add dynamic multi-collection coverage for import

mohammed-madi:
CBG-2638

Describe your PR here...
- TestMultiCollectionImportFilter now also tests adding and removing a collection

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1420/

### DIFF
--- a/rest/adminapitest/admin_api_test.go
+++ b/rest/adminapitest/admin_api_test.go
@@ -1015,11 +1015,10 @@ func TestResyncErrorScenariosUsingDCPStream(t *testing.T) {
 	}`
 
 	testBucket := base.GetTestBucket(t)
-	defer testBucket.Close()
 	rt := rest.NewRestTester(t,
 		&rest.RestTesterConfig{
 			SyncFn:           syncFn,
-			CustomTestBucket: testBucket.NoCloseClone(),
+			CustomTestBucket: testBucket,
 		},
 	)
 	defer rt.Close()
--- a/rest/adminapitest/admin_api_test.go
+++ b/rest/adminapitest/admin_api_test.go
@@ -1015,6 +1015,7 @@ func TestResyncErrorScenariosUsingDCPStream(t *testing.T) {
 	}`
 
 	testBucket := base.GetTestBucket(t)
+
 	rt := rest.NewRestTester(t,
 		&rest.RestTesterConfig{
 			SyncFn:           syncFn,
--- a/rest/adminapitest/admin_api_test.go
+++ b/rest/adminapitest/admin_api_test.go
@@ -1015,11 +1015,11 @@ func TestResyncErrorScenariosUsingDCPStream(t *testing.T) {
 	}`
 
 	testBucket := base.GetTestBucket(t)
-
+	defer testBucket.Close()
 	rt := rest.NewRestTester(t,
 		&rest.RestTesterConfig{
 			SyncFn:           syncFn,
-			CustomTestBucket: testBucket,
+			CustomTestBucket: testBucket.NoCloseClone(),
 		},
 	)
 	defer rt.Close()
--- a/rest/attachmentcompactiontest/attachment_compaction_api_test.go
+++ b/rest/attachmentcompactiontest/attachment_compaction_api_test.go
@@ -162,8 +162,10 @@ func TestAttachmentCompactionPersistence(t *testing.T) {
 	}
 
 	tb := base.GetTestBucket(t)
+	noCloseTB := tb.NoCloseClone()
+
 	rt1 := rest.NewRestTester(t, &rest.RestTesterConfig{
-		CustomTestBucket: tb,
+		CustomTestBucket: noCloseTB,
 	})
 	rt2 := rest.NewRestTester(t, &rest.RestTesterConfig{
 		CustomTestBucket: tb,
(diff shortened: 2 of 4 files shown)


rest/importtest/collections_import_test.go
```diff
@@ -136,4 +146,100 @@ func TestMultiCollectionImportFilter(t *testing.T) {
...
+	scopesConfig[dataStoreNames[2].ScopeName()].Collections[dataStoreNames[2].CollectionName()] = rest.CollectionConfig{ImportFilter: &importFilter3}
+
+	scopesConfigString, err := json.Marshal(scopesConfig)
+	require.NoError(t, err)
+
+	response = rt.SendAdminRequest("PUT", "/db/_config", fmt.Sprintf(
+		`{"bucket": "%s", "num_index_replicas": 0, "enable_shared_bucket_access": %t, "scopes":%s}`,
+		testBucket.GetName(), base.TestUseXattrs(), string(scopesConfigString)))
+	rest.RequireStatus(t, response, http.StatusCreated)
+	rt2 := rest.NewRestTesterMultipleCollections(t, rtConfig, 3)
```
torcolvin: Initializing another RestTester in general means this test requires two buckets to run, but it doesn't here because `RestTester` is constructed with a non closeable bucket. This is not quite the configuration we want to necessarily be testing because like two SG running that are targeting the same bucket (the first hasn't shut down yet).

In Jenkins, this doesn't matter, but in dev running tests under GSI, I use `SG_TEST_BUCKET_POOL_SIZE=1` and `SG_TEST_COLLECTION_POOL_SIZE=1` to speed up the test harness. This only works when you have set this up. I'd match `base.RequireNumTestDataStores(t, 3)` as a constant with `numCollections` you create. I'd rather matching maxNumCollections here.

rest/importtest/collections_import_test.go
```diff
@@ -136,4 +146,100 @@ func TestMultiCollectionImportFilter(t *testing.T) {
...
+	defer rt2.Close()
+
+	_ = rt2.Bucket()
+	dataStore1, err = rt2.TestBucket.GetNamedDataStore(0)
+	require.NoError(t, err)
+	dataStore2, err = rt2.TestBucket.GetNamedDataStore(1)
+	require.NoError(t, err)
+	dataStore3, err := rt2.TestBucket.GetNamedDataStore(2)
+	require.NoError(t, err)
```
torcolvin: As above, these are actually the same bucket as bucket 1, which makes for a somewhat confusing test, or at least duplicated code.

torcolvin (commented):
Tests look good but this test is now long and convoluted and I think it's worth refactoring to make it clear what separate things it is doing. I think this might mean splitting into 3 tests for each time you are retesting the RestTester. I've put some more detailed comments below.

rest/importtest/collections_import_test.go
mohammed-madi: I ran into an issue where gocb would throw `collections are not enabled` on the `dataStore.Add` call if `GetNamedDataStore` was not called again after PUTting a new db config and `GetNamedDataStore` was throwing `connection shut down` if it was called with the old restTester (rt)
Now that I think of it I should've tested this manually to see if theres a bug.
I could leave comments clarifying these points or I could make another test that minimises the assertions done in this test, and focuses on the scenarios I added (if you check the ticket, bullet points 1-3 are tested in this test)

rest/importtest/collections_import_test.go
mohammed-madi: [1df3251](https://github.com/couchbase/sync_gateway/pull/6030/commits/1df32518eb2600aa65b659ff1fb1f5c07c334d2e)

rest/importtest/collections_import_test.go
mohammed-madi: [1df3251](https://github.com/couchbase/sync_gateway/pull/6030/commits/1df32518eb2600aa65b659ff1fb1f5c07c334d2e)

rest/adminapitest/admin_api_test.go
```diff
@@ -1015,7 +1015,6 @@ func TestResyncErrorScenariosUsingDCPStream(t *testing.T) {
 	}`
 
 	testBucket := base.GetTestBucket(t)
-
```
torcolvin: we can drop changes to this file

rest/attachmentcompactiontest/attachment_compaction_api_test.go
```diff
@@ -162,10 +162,8 @@ func TestAttachmentCompactionPersistence(t *testing.T) {
 	}
 
 	tb := base.GetTestBucket(t)
-	noCloseTB := tb.NoCloseClone()
```
torcolvin: I seems suspect that this _does_ need a noCloseClone because what happens in this test (see defers) is that rt1 tears down and doesn't close the bucket, then rt2 will actually close the bucket.

Anyway, it's out side scope of this PR so i'd just put it back.

mohammed-madi: [Undo TestAttachmentCompactionPersistence change](https://github.com/couchbase/sync_gateway/pull/6030/commits/d9e17a6d3cfcf27dfa36302335b45befc372e946)

rest/adminapitest/admin_api_test.go
mohammed-madi: [undo TestResyncErrorScenariosUsingDCPStream change](https://github.com/couchbase/sync_gateway/pull/6030/commits/4dcb5a500b415749889b960d89d45c0d2228ea1a)

torcolvin (approved):

PR #6031, opened 2023-01-20
CBG-2489: Collection stats

bbrks:
CBG-2489

- Refactor NewDBStats to take slices of query and collections to initialise
- Add collection stats for sync and import
- NumDocsRead/NumDocsWrite and associated Bytes stats
- Unit/integration test coverage 

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1434/

### DIFF
--- a/base/stats.go
+++ b/base/stats.go
@@ -456,9 +456,9 @@ type DatabaseStats struct {
 	WarnGrantsPerDocCount *SgwIntStat `json:"warn_grants_per_doc_count"`
 	// The total number of warnings relating to the xattr sync data being larger than a configured threshold.
 	WarnXattrSizeCount *SgwIntStat `json:"warn_xattr_size_count"`
-	// The total number of times that the sync_function is evaluated.
+	// The total number of times that a sync function was evaluated for the database (across all collections).
 	SyncFunctionCount *SgwIntStat `json:"sync_function_count"`
-	// The total time spent evaluating the sync_function.
+	// The total time spent evaluating a sync function (across all collections).
 	SyncFunctionTime *SgwIntStat `json:"sync_function_time"`
 
 	// These can be cleaned up in future versions of SGW, implemented as maps to reduce amount of potential risk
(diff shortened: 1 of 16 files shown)


db/crud.go
```diff
@@ -2220,14 +2220,18 @@ func (db *DatabaseCollectionWithUser) getChannelsAndAccess(ctx context.Context,
 
 	if db.channelMapper() != nil {
 		// Call the ChannelMapper:
-		startTime := time.Now()
 		db.dbStats().Database().SyncFunctionCount.Add(1)
+		db.dbStats().CollectionStat(db.ScopeName(), db.Name()).SyncFunctionCount.Add(1)
```
adamcfraser: What do you think about storing a reference to the CollectionStats on DatabaseCollection when it's initialized, to avoid the map lookup every time we want to write a stat?  (that's especially true if the CollectionStat call ends up needing a mutex)

base/stats.go
```diff
@@ -1402,6 +1431,19 @@ func (d *DbStats) unregisterDatabaseStats() {
 	prometheus.Unregister(d.DatabaseStats.SyncFunctionTime)
 }
 
+func (d *DbStats) CollectionStat(scopeName, collectionName string) *CollectionStats {
+	scopeAndCollectionName := scopeName + "." + collectionName
+	if _, ok := d.CollectionStats[scopeAndCollectionName]; !ok {
+		// DbStats was not initialised with this collection upfront in NewDBStats for some reason (_default._default for example, since we pass named collections only ...)
```
adamcfraser: This looks like a potential data race, and locking would be expensive.  I feel like we should rely on NewDBStats to do the right thing and just log an error if the requested scope and collection is not found.

adamcfraser (changes requested):

adamcfraser (approved):

PR #6042, opened 2023-01-25
CBG-2672 remove querying skipped sequences

torcolvin:
Removing querying skipped sequences in preparation for making the CBG-2672 change to use a single changeCache for all collections. 

We think this is probably not worth the time that it costs to do the query, and if we hit this code, that means there's a bug in SG processing DCP information, and it would be better to fix that than running an expensive query. It's unknown how often customers might hit this scenario.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1435/


db/change_cache.go
```diff
@@ -317,68 +315,8 @@ func (c *changeCache) CleanSkippedSequenceQueue(ctx context.Context) error {
...
-		changedChannels := c.processEntry(entry)
-		changedChannelsCombined = changedChannelsCombined.Update(changedChannels)
-	}
-
-	// Since the calls to processEntry() above may unblock pending sequences, if there were any changed channels we need
-	// to notify any change listeners that are working changes feeds for these channels
-	if c.notifyChange != nil && len(changedChannelsCombined) > 0 {
-		c.notifyChange(changedChannelsCombined)
 	}
+	oldSkippedSequences = nil
```
adamcfraser: We need to remove the condition on line 316, I believe, in order to properly populate pendingRemovals.

adamcfraser (changes requested):

adamcfraser (approved):

PR #6071, opened 2023-02-01
CBG-2690 loop over all servers when determining high seqno

torcolvin:
Remember to look at this PR with whitespace changed turned off.

- Apparently the operation succeeds even if the server node doesn't have kv enabled, so this is a safe call. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1460/


adamcfraser (approved):
As discussed, I think there are some failover/rebalance cases that we'll better handle with the pending SDK enhancement, but this covers the functionality until we get that.

PR #6073, opened 2023-02-02
CBG-2633: TestMultiCollectionChangesMultipleChannels

mohammed-madi:
CBG-2633

Describe your PR here...
- TestMultiCollectionChangesMultipleChannels adds coverage for changes feeds (one shot, continuous) with multiple collections

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1496/

I ran the test 10 times on jenkins with -race to see if its potentially flakey due to the continuous feed goroutine.

### DIFF
--- a/rest/changestest/changes_collection_test.go
+++ b/rest/changestest/changes_collection_test.go
@@ -292,9 +292,9 @@ func TestMultiCollectionChangesMultiChannelOneShot(t *testing.T) {
 	changesResponse = rt.GetChangesOneShot(t, "keyspace2", 0, "BRNchr", 3) // 2 docs in BRN, 1 doc in CHR in collection 2
 	logChangesResponse(t, changesResponse.Body.Bytes())
 
-	rt.AssertChangesFeedMultiCollection(t, "bernard", "continuous", 2, []int{4, 2}, 500)
-	rt.AssertChangesFeedMultiCollection(t, "charlie", "continuous", 2, []int{3, 1}, 500)
-	rt.AssertChangesFeedMultiCollection(t, "BRNchr", "continuous", 2, []int{7, 3}, 1000)
+	rt.RequireContinuousFeedChangesCount(t, "bernard", "continuous", 2, []int{4, 2}, 500)
+	rt.RequireContinuousFeedChangesCount(t, "charlie", "continuous", 2, []int{3, 1}, 500)
+	rt.RequireContinuousFeedChangesCount(t, "BRNchr", "continuous", 2, []int{7, 3}, 1000)
 }
 
 // TestMultiCollectionChangesUserDynamicGrant tests a dynamic channel grant when that channel is not already resident
--- a/rest/changestest/changes_collection_test.go
+++ b/rest/changestest/changes_collection_test.go
@@ -283,7 +283,7 @@ func TestMultiCollectionChangesMultipleChannels(t *testing.T) {
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		changesResponse = rt.SendUserRequest("GET", "/{{.keyspace1}}/_changes?feed=continuous&timeout=5000", "", "bernard")
+		changesResponse = rt.SendUserRequest("GET", "/{{.keyspace1}}/_changes?feed=continuous&timeout=2000", "", "bernard")
 		rest.RequireStatus(t, changesResponse, 200)
 		contChanges, err := rt.ReadContinuousChanges(changesResponse)
 		assert.NoError(t, err)
@@ -301,6 +301,8 @@ func TestMultiCollectionChangesMultipleChannels(t *testing.T) {
 	rest.RequireStatus(t, response, 201)
 	response = rt.SendAdminRequest("PUT", "/{{.keyspace1}}/chr4_c1", `{"value":1, "channels":["CHR"]}`)
 	rest.RequireStatus(t, response, 201)
+	_ = rt.WaitForPendingChanges()
+
 	wg.Wait()
 }
 
--- a/rest/changestest/changes_collection_test.go
+++ b/rest/changestest/changes_collection_test.go
@@ -282,7 +282,7 @@ func TestMultiCollectionChangesMultiChannelOneShot(t *testing.T) {
 	// Changes that shouldn't be caught by the continuous changes feed
 	response = rt.SendAdminRequest("PUT", "/{{.keyspace2}}/brn4_c1", `{"value":1, "channels":["BRN"]}`)
 	rest.RequireStatus(t, response, 201)
-	response = rt.SendAdminRequest("PUT", "/{{.keyspace1}}/chr4_c1", `{"value":1, "channels":["CHR"]}`)
+	response = rt.SendAdminRequest("PUT", "/{{.keyspace1}}/chr3_c1", `{"value":1, "channels":["CHR"]}`)
 	rest.RequireStatus(t, response, 201)
 	_ = rt.WaitForPendingChanges()
 
@@ -292,6 +292,9 @@ func TestMultiCollectionChangesMultiChannelOneShot(t *testing.T) {
 	changesResponse = rt.GetChangesOneShot(t, "keyspace2", 0, "BRNchr", 3) // 2 docs in BRN, 1 doc in CHR in collection 2
 	logChangesResponse(t, changesResponse.Body.Bytes())
 
+	rt.AssertChangesFeedMultiCollection(t, "bernard", "continuous", 2, []int{4, 2}, 500) // 2 docs in BRN, 1 doc in CHR in collection 2
+	rt.AssertChangesFeedMultiCollection(t, "charlie", "continuous", 2, []int{3, 1}, 500) // 2 docs in BRN, 1 doc in CHR in collection 2
+	rt.AssertChangesFeedMultiCollection(t, "BRNchr", "continuous", 2, []int{7, 3}, 1000) // 2 docs in BRN, 1 doc in CHR in collection 2
 }
 
 // TestMultiCollectionChangesUserDynamicGrant tests a dynamic channel grant when that channel is not already resident
--- a/rest/changestest/changes_collection_test.go
+++ b/rest/changestest/changes_collection_test.go
@@ -264,6 +264,8 @@ func TestMultiCollectionChangesMultipleChannels(t *testing.T) {
 
 	response = rt.SendAdminRequest("PUT", "/{{.keyspace1}}/brn3_c1", `{"value":1, "channels":["BRN"]}`)
 	rest.RequireStatus(t, response, 201)
+	_ = rt.WaitForPendingChanges()
+
 	changesResponse = rt.SendUserRequest("GET", fmt.Sprintf("/{{.keyspace1}}/_changes?since=%s", lastSeqBRN), "", "bernard")
 	err = base.JSONUnmarshal(changesResponse.Body.Bytes(), &changes)
 	assert.NoError(t, err, "Error unmarshalling changes response")
@@ -281,7 +283,7 @@ func TestMultiCollectionChangesMultipleChannels(t *testing.T) {
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		changesResponse = rt.SendUserRequest("GET", "/{{.keyspace1}}/_changes?feed=continuous&timeout=2000", "", "bernard")
+		changesResponse = rt.SendUserRequest("GET", "/{{.keyspace1}}/_changes?feed=continuous&timeout=5000", "", "bernard")
 		rest.RequireStatus(t, changesResponse, 200)
 		contChanges, err := rt.ReadContinuousChanges(changesResponse)
 		assert.NoError(t, err)
--- a/rest/changestest/changes_collection_test.go
+++ b/rest/changestest/changes_collection_test.go
@@ -9,7 +9,9 @@
 package changestest
 
 import (
+	"fmt"
 	"log"
+	"sync"
 	"testing"
 
 	"github.com/couchbase/sync_gateway/channels"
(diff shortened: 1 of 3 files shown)


rest/changestest/changes_collection_test.go
```diff
@@ -191,6 +194,130 @@ func TestMultiCollectionChangesUser(t *testing.T) {
...
+	assert.NoError(t, a.Save(brndan))
+
+	// Put several documents, will be retrieved via query
+	response := rt.SendAdminRequest("PUT", "/{{.keyspace1}}/brn1_c1", `{"channels":["BRN"]}`)
+	rest.RequireStatus(t, response, 201)
+
+	// Keep rev to delete doc later
+	var body db.Body
+	require.NoError(t, base.JSONUnmarshal(response.Body.Bytes(), &body))
+	rev := body["rev"].(string)
```
torcolvin: I'd be inclined to name this `brn1_c1Rev` for test readability.

rest/changestest/changes_collection_test.go
```diff
@@ -191,6 +194,130 @@ func TestMultiCollectionChangesUser(t *testing.T) {
 	logChangesResponse(t, changesResponse.Body.Bytes())
 }
 
+func TestMultiCollectionChangesMultipleChannels(t *testing.T) {
```
torcolvin: I'd add OneShot to this

rest/changestest/changes_collection_test.go
```diff
@@ -191,6 +194,130 @@ func TestMultiCollectionChangesUser(t *testing.T) {
...
+	bernard, err := a.NewUser("bernard", "letmein", channels.BaseSetOf(t, "BRN"))
+	assert.NoError(t, err)
+	assert.NoError(t, a.Save(bernard))
+	charlie, err := a.NewUser("charlie", "letmein", channels.BaseSetOf(t, "CHR"))
+	assert.NoError(t, err)
+	assert.NoError(t, a.Save(charlie))
+	brndan, err := a.NewUser("BRNchr", "letmein", channels.BaseSetOf(t, "CHR", "BRN"))
+	assert.NoError(t, err)
+	assert.NoError(t, a.Save(brndan))
+
```
torcolvin, line 215: you can start 2 continuous changes feed here for keyspace1 and keyspace2, and everytime you look for a one shot feed you should see changes or no changes. I'd always check one shot feed first because I think it's going to be more straightforward to debug.

The other approach is to write two different tests, but I think testing these in line is also clear.

rest/changestest/changes_collection_test.go
```diff
@@ -191,6 +194,130 @@ func TestMultiCollectionChangesUser(t *testing.T) {
...
+	rest.RequireStatus(t, response, 201)
+	response = rt.SendAdminRequest("PUT", "/{{.keyspace1}}/chr2_c1", `{"value":1, "channels":["CHR"]}`)
+	rest.RequireStatus(t, response, 201)
+	response = rt.SendAdminRequest("PUT", "/{{.keyspace2}}/brn2_c2", `{"value":1, "channels":["BRN"]}`)
+	rest.RequireStatus(t, response, 201)
+	response = rt.SendAdminRequest("PUT", "/{{.keyspace2}}/chr2_c2", `{"value":1, "channels":["CHR"]}`)
+	rest.RequireStatus(t, response, 201)
+	_ = rt.WaitForPendingChanges()
+
+	changesResponse = rt.SendUserRequest("GET", "/{{.keyspace1}}/_changes?since=0", "", "bernard")
```
torcolvin: each time we read from single shot changes feed, we should assert that we see the same results come over the continous changes feed. (or not)

We might have to add a `rt.WaitForCondition` to reading continuous feed because the single shot changes feed and the single shot feed are running in separate go routines. I'd be inclined to always verify the single shot feed first because it'll ease debugging in the future.

rest/changestest/changes_collection_test.go
```diff
@@ -191,6 +194,130 @@ func TestMultiCollectionChangesUser(t *testing.T) {
...
+	assert.NoError(t, err, "Error unmarshalling changes response")
+	require.Len(t, changes.Results, 1)
+	logChangesResponse(t, changesResponse.Body.Bytes())
+
+	changesResponse = rt.SendUserRequest("GET", "/{{.keyspace2}}/_changes?since=0", "", "bernard")
+	err = base.JSONUnmarshal(changesResponse.Body.Bytes(), &changes)
+	assert.NoError(t, err, "Error unmarshalling changes response")
+	require.Len(t, changes.Results, 0)
+	logChangesResponse(t, changesResponse.Body.Bytes())
+
```
torcolvin, line 235: add asserts for continous feeds here

rest/changestest/changes_collection_test.go
```diff
@@ -191,6 +194,130 @@ func TestMultiCollectionChangesUser(t *testing.T) {
...
+	change, err = rest.GetNextContinuousChange(reader)
+	require.Error(t, io.EOF)
+
+	// Changes that shouldn't be caught by the continuous changes feed
+	response = rt.SendAdminRequest("PUT", "/{{.keyspace2}}/brn4_c1", `{"value":1, "channels":["BRN"]}`)
+	rest.RequireStatus(t, response, 201)
+	response = rt.SendAdminRequest("PUT", "/{{.keyspace1}}/chr4_c1", `{"value":1, "channels":["CHR"]}`)
+	rest.RequireStatus(t, response, 201)
+	_ = rt.WaitForPendingChanges()
+
```
torcolvin, line 266: assert no changes over continuous feed

rest/changestest/changes_collection_test.go
```diff
@@ -191,6 +194,130 @@ func TestMultiCollectionChangesUser(t *testing.T) {
...
+	assert.NoError(t, err, "Error unmarshalling changes response")
+	require.Len(t, changes.Results, 2)
+	logChangesResponse(t, changesResponse.Body.Bytes())
+	lastSeqBRN := changes.Last_Seq
+
+	response = rt.SendAdminRequest("PUT", "/{{.keyspace1}}/brn3_c1", `{"value":1, "channels":["BRN"]}`)
+	rest.RequireStatus(t, response, 201)
+	_ = rt.WaitForPendingChanges()
+
+	changesResponse = rt.SendUserRequest("GET", fmt.Sprintf("/{{.keyspace1}}/_changes?since=%s", lastSeqBRN), "", "bernard")
```
torcolvin: we want to assert there are also no docs over changes feed keyspace2 - I think you can write a helper functionfunction to verify number of results like

`func getChangesOneShot(t testing *T, keyspace, since, username string, changesCount int)` 

which would improve readability of this test.

rest/changestest/changes_collection_test.go
```diff
@@ -191,6 +194,130 @@ func TestMultiCollectionChangesUser(t *testing.T) {
...
+	err = base.JSONUnmarshal(changesResponse.Body.Bytes(), &changes)
+	assert.NoError(t, err, "Error unmarshalling changes response")
+	require.Len(t, changes.Results, 1)
+	logChangesResponse(t, changesResponse.Body.Bytes())
+
+	changesResponse = rt.SendUserRequest("GET", "/{{.keyspace1}}/_changes?since=0", "", "charlie")
+	err = base.JSONUnmarshal(changesResponse.Body.Bytes(), &changes)
+	assert.NoError(t, err, "Error unmarshalling changes response")
+	require.Len(t, changes.Results, 2)
+	logChangesResponse(t, changesResponse.Body.Bytes())
```
torcolvin: we test elsewhere the since values work, so I'd be inclined to just check keyspace1 and keyspace2 don't collide and leave it at that.

rest/changestest/changes_collection_test.go
```diff
@@ -191,6 +194,130 @@ func TestMultiCollectionChangesUser(t *testing.T) {
...
+
+	// Start a continuous changes feed in its own goroutine
+	continuousResponse := rt.SendUserRequest("GET", "/{{.keyspace1}}/_changes?feed=continuous&timeout=5000", "", "bernard")
+	reader := bufio.NewReader(continuousResponse.Body)
```
torcolvin: we want to make the timeout really big here to make sure that CBS in integration tests is going to be able to pick up these changes, I'd start at 30s

rest/changestest/changes_collection_test.go
```diff
@@ -191,6 +194,130 @@ func TestMultiCollectionChangesUser(t *testing.T) {
...
+
+	// change 1
+	change, err := rest.GetNextContinuousChange(reader)
+	require.NoError(t, err)
+	require.Equal(t, uint64(1), change.Seq.Seq)
+	require.Equal(t, "brn1_c1", change.ID)
```
torcolvin: I'd write the asserts to match the singleShot changes feed, so we could either just check number, or check sequence number and docID.

rest/changestest/changes_collection_test.go
torcolvin: We aren't going to pursue this.

rest/changestest/changes_collection_test.go
torcolvin: We aren't going to do this.

rest/changestest/changes_collection_test.go
torcolvin: I'd be inclined to shorten the timeout here and move this code to the end of the test, where we run a matrix of tests

| user        | keyspace1 | keyspace2 |
| --------  | ---------- | ----------- | 
| bernand |                    |                    |
| charlie    |                    |                    |
| BRNchr  |                    |                    | 

You should be able to run a function like 

```
response := rt.SendUserRequest(http.MethodGet, fmt.Sprintf("/%s/_changes?feed=continuous&since=%s&timeout=2000", keyspace since), "", "bernard")
changes, err := readContinuousChanges(changesResponse)
// Assert number of changes
```

I'd see if we can shorten the 2000 timeout, but we might have to do something fancier if this fails under CBS. The goal is to make the changes feed run as fast as possible without actually failing the test. The changes will be in memory since we have just run the changes feed with one shot, so I think we can drop to 500/1000. In the event that this doesn't work, we'd write an assertion where we re-run the request with increasingly long timeouts until it passes (or fails). But I'm not sure we have to go there yet.

rest/changestest/changes_collection_test.go
torcolvin: We can drop this for the matrix testing approach at the end of the test.

rest/changestest/changes_collection_test.go
mohammed-madi: https://github.com/couchbase/sync_gateway/pull/6073/commits/2d3706629b8aa3313fac3191cdacb93da1dc08c6
Im not sure if this makes the continuous changes checks a t l.260 and after redundant, but let me know if i should remove anything

rest/utilities_testing.go
```diff
@@ -2337,11 +2337,23 @@ func GetNextContinuousChange(reader *bufio.Reader) (*db.ChangeEntry, error) {
...
-		Results []db.ChangeEntry
-		LastSeq db.SequenceID
+var changes struct {
+	Results []db.ChangeEntry
+	LastSeq db.SequenceID
+}
+
+// AssertChangesFeedMultiCollection Calls a changes feed on every collection and asserts that the nth expected change is
+// the number of changes for the nth collection.
+func (rt *RestTester) AssertChangesFeedMultiCollection(t testing.TB, username string, feedMode string, numKeyspaces int, expectedChanges []int, timeout int) {
```
torcolvin: This LGTM but it's confusing to read so I'd just changes to

```
RequireContinousFeedChangesCount(tb testing.TB, username, keyspace, expectedChanges int) 
```

This function only works with continuous feed anyway, and then you can map keyspace name with number of changes.

rest/utilities_testing.go
```diff
@@ -2337,11 +2337,23 @@ func GetNextContinuousChange(reader *bufio.Reader) (*db.ChangeEntry, error) {
 	}
 }
 
-func (rt *RestTester) GetChangesOneShot(t testing.TB, keyspace string, since int, username string, changesCount int) *TestResponse {
-	var changes struct {
-		Results []db.ChangeEntry
-		LastSeq db.SequenceID
+var changes struct {
+	Results []db.ChangeEntry
+	LastSeq db.SequenceID
```
torcolvin: This is duplicated in a ton of places and is canonically `rest.ChangesResults`.

rest/changestest/changes_collection_test.go
torcolvin: I think this will make the continuous changes checks redundant and we can remove a bunch of code.

rest/changestest/changes_collection_test.go
torcolvin: I think changing this test would be really helpful.

rest/changestest/changes_collection_test.go
torcolvin: We're going to go with an approach where we just check at the end.

rest/utilities_testing.go
```diff
@@ -2285,3 +2287,72 @@ func (rt *RestTester) getCollectionsForBLIP() []string {
...
+			}
+			log.Printf("Got change ==> %v", change)
+			return change, nil
+		}
+	}
+}
+
+// RequireContinuousFeedChangesCount Calls a changes feed on every collection and asserts that the nth expected change is
+// the number of changes for the nth collection.
+func (rt *RestTester) RequireContinuousFeedChangesCount(t testing.TB, username string, numKeyspaces int, expectedChanges []int, timeout int) {
```
torcolvin: I think I didn't make this as clear, but I actually think that writing this assertion multiple times for:

`RequireContinousFeedChangesCount(t, username, keyspace string, expected changes)`

rest/changestest/changes_collection_test.go
torcolvin: Let's remove all this code with `GetNextContinuousChange` and the callers and focus on the `RequireContinuousFeedChangesCount` and that's it for this review.

mohammed-madi: https://github.com/couchbase/sync_gateway/pull/6073/commits/b93638b51cdc0a6d1dc39af955ca321ecd0be877

rest/utilities_testing.go
mohammed-madi: https://github.com/couchbase/sync_gateway/pull/6073/commits/40ad8fa145d1006e7227564a278b1b60341e5654

torcolvin (approved):

PR #6075, opened 2023-02-03
CBG-2683 Validate metadatastore name

torcolvin:
- I'm not sure what the redaction should be here, we would use base.MD(dbName), but this only occurs on an update of the database where they presumably already have access to this information.
- Should we retry a KV op, or should I explicitly check reason as KV_COLLECTION_OUTDATED? I'm not sure if that's always going to show up.
- There's not an obvious way for me to test the negative version of this.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1473/

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -14,7 +14,6 @@ import (
 	"fmt"
 	"net/http"
 	"regexp"
-	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -359,11 +358,6 @@ func NewDatabaseContext(ctx context.Context, dbName string, bucket base.Bucket, 
 		metadataStore = bucket.DefaultDataStore()
 	}
 
-	err := validateMetadataStore(ctx, metadataStore)
-	if err != nil {
-		return nil, err
-	}
-
 	// Register the cbgt pindex type for the configGroup
 	RegisterImportPindexImpl(ctx, options.GroupID)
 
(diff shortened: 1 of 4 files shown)


db/database.go
```diff
@@ -2331,6 +2336,22 @@ func (dbc *DatabaseContext) GetSingleDatabaseCollection() *DatabaseCollection {
 	return dbc.singleCollection
 }
 
+func validateMetadataStore(ctx context.Context, metadataStore base.DataStore) error {
+	_, err := metadataStore.Exists("fakedoc") // no need to pass a doc and check if it exists, this op checks if the datastore is real
```
adamcfraser: What's the retry strategy we use for the Exists call?  We don't want to report the metadataStore as missing due to a transient server connectivity issue.  

Would requiring a successful GetCollectionID() be a more reliable alternative?

adamcfraser (changes requested):

db/database.go
torcolvin: There's no retry on `GetCollectionID` either which sort of runs into the same problem where we do not retry and expect the connection to be open.

`NamedDataStore` calls `Collection.setCollectionID` in `newCollection` constructor. For probably inconsistent reasons, we construct `Collection{}` literally because `collectionID` on default collection is zero.

I think this might expose a different problem where we have transient errors in `NewDatabaseContext` from `bucket.NamedDataStore` since we don't retry KV ops there.

It might be well to put all the retry logic into the `newCollection` constructor which would handle both these circumstances. What retry time is reasonable?

torcolvin: In further investigation, this is done for non metadatastore in https://github.com/couchbase/sync_gateway/blob/master/rest/server_context.go#L507

It seems like I should move this logic to `server_context.go` so it works the same for each collection. It returns an `ErrAuthNotFound`.

I would be tempted to have `DefaultDataStore` call `NamedDataStore` but the latter returns and error and the former does not and would need to change sg-bucket/walrus.

adamcfraser (approved):

PR #6083, opened 2023-02-07
CBG-2687 allow docsProcessed to account for server tombstoned docs that come over DCP

torcolvin:
We are processing all docs that come over DCP, but when running tests outside of isolation, and reusing the same bucket, we can see documents over DCP in a different test. (Diagnosed this by printing the docsProcessed/docsChanged).

Looking at these extra docs, they come from a previous tests and we'll see a mutation event (and the doc body is not found in bucket, so they are not processed), followed by a deletion event. I added code to skip the deletion event processing.

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1481/

### DIFF
--- a/db/background_mgr_resync_dcp.go
+++ b/db/background_mgr_resync_dcp.go
@@ -107,8 +107,9 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]interface
 		docID := string(event.Key)
 		key := realDocID(docID)
 		base.TracefCtx(ctx, base.KeyAll, "[%s] Received DCP event %d for doc %v", resyncLoggingID, event.Opcode, base.UD(docID))
-		// Ignore deletion events
-		if event.Opcode == sgbucket.FeedOpDeletion {
+
+		// Ignore documents without Xattrs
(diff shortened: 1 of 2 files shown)


db/background_mgr_resync_dcp.go
```diff
@@ -105,10 +104,13 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]interface
 	defer atomic.CompareAndSwapUint32(&db.State, DBResyncing, DBOffline)
 
 	callback := func(event sgbucket.FeedEvent) bool {
-		var err error
 		docID := string(event.Key)
 		key := realDocID(docID)
 		base.TracefCtx(ctx, base.KeyAll, "[%s] Received DCP event %d for doc %v", resyncLoggingID, event.Opcode, base.UD(docID))
+		// Ignore deletion events
+		if event.Opcode == sgbucket.FeedOpDeletion {
```
adamcfraser: Only ignore deletes if there isn't a mobile xattr on the document.

adamcfraser (changes requested):

adamcfraser (approved):

PR #6086, opened 2023-02-08
CBG-2636: Add EE checks to admin api auth tests that use mobile sgw role

mohammed-madi:
CBG-2636
Add rt.IsServerEnterprise to check if the connected CBS instance is EE or CE

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1489/

### DIFF
--- a/rest/utilities_testing.go
+++ b/rest/utilities_testing.go
@@ -29,8 +29,6 @@ import (
 	"text/template"
 	"time"
 
-	"github.com/couchbase/gocb/v2"
-
 	"github.com/couchbase/go-blip"
 	sgbucket "github.com/couchbase/sg-bucket"
 	"github.com/couchbase/sync_gateway/auth"
@@ -2289,19 +2287,4 @@ func (rt *RestTester) getCollectionsForBLIP() []string {
 			strings.Join([]string{collection.ScopeName, collection.Name}, base.ScopeCollectionSeparator))
 	}
 	return collections
-}
-
-// IsServerEnterprise returns true if the connected couchbase server instance is Enterprise edition
-// And false for Community edition
-func (rt *RestTester) IsServerEnterprise(t testing.TB) bool {
-	gocbBucket, err := base.AsGocbV2Bucket(rt.Bucket())
-	require.NoError(t, err)
-
-	metadata, err := gocbBucket.GetCluster().Internal().GetNodesMetadata(&gocb.GetNodesMetadataOptions{})
-	require.NoError(t, err)
-
-	if strings.Contains("enterprise", metadata[0].Version) {
-		return true
-	}
-	return false
 }
--- a/rest/utilities_testing.go
+++ b/rest/utilities_testing.go
@@ -16,7 +16,6 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
-	"github.com/couchbase/gocb/v2"
 	"log"
 	"net"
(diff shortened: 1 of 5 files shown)


rest/utilities_testing.go
```diff
@@ -2288,3 +2290,18 @@ func (rt *RestTester) getCollectionsForBLIP() []string {
...
+	require.NoError(t, err)
+
+	metadata, err := gocbBucket.GetCluster().Internal().GetNodesMetadata(&gocb.GetNodesMetadataOptions{})
+	require.NoError(t, err)
+
+	if strings.Contains("enterprise", metadata[0].Version) {
+		return true
+	}
+	return false
+}
```
torcolvin: this makes a kv op when you call it, but we know as soon as you connect to the cluster, let's move this to when we make a connection for our test cluster and put this in the base package:

https://github.com/couchbase/sync_gateway/blob/master/base/main_test_cluster.go#L98

And add a helper like we do for `TestsUseCollections` to be `TestsUseServerCE`

I think this is worthwhile since we know we'll need this in the DB package. I'd even be tempted to refactor this function to something testable in base package and return a "enum" value 

```
type ServerType uint8

const (
     ServerCommunityEdition
     ServerEnterpriseEdition
)
 
func getServerEdition(cluster) (ServerType, error)
```

The thing I'm trying to protect against is that we are using an "internal" API and there's a chance CBS doesn't specify those versions and it's a coincidence this API works, so having a test that specifically tests this API would be great.

torcolvin (commented):
LGTM, though I'd prefer a refactor to move test function into a testable function and call it once per CBS connection.

One thing we might want to consider for future tests is how this variable will interact with walrus, which won't have any server edition. But I think it's out of scope for this PR.

rest/utilities_testing_test.go
```diff
@@ -99,3 +103,22 @@ func TestAttachmentRoundTrip(t *testing.T) {
...
+		form := url.Values{}
+		form.Add("password", "password")
+		form.Add("roles", "[mobile_sync_Gateway]")
+		eps, _, err := rt.ServerContext().ObtainManagementEndpointsAndHTTPClient()
+		require.NoError(t, err)
+
+		req, err := http.NewRequest("PUT", fmt.Sprintf("%s/settings/rbac/users/local/%s", eps[0], "username"), strings.NewReader(form.Encode()))
+		require.Error(t, err)
+		require.Equal(t, req, http.StatusBadRequest)
+	}
```
torcolvin:
```suggestion
	if base.TestsUseServerCE() {
		t.Skip("test only runs with CE server")
	}
	rt := NewRestTester(t, nil)
	defer rt.Close()
	form := url.Values{}
	form.Add("password", "password")
	form.Add("roles", "[mobile_sync_Gateway]")
	eps, _, err := rt.ServerContext().ObtainManagementEndpointsAndHTTPClient()
	require.NoError(t, err)

	req, err := http.NewRequest("PUT", fmt.Sprintf("%s/settings/rbac/users/local/%s", eps[0], "username"), strings.NewReader(form.Encode()))
	require.Error(t, err)
	require.Equal(t, req, http.StatusBadRequest)
	
```

in idiomatic go, prefer the early return for readability.

torcolvin (dismissed):

torcolvin (approved):

PR #6088, opened 2023-02-08
CBG-2699 add clusterUUID to sync metadata

torcolvin:
-renamed receiver methods from `db` -> `col` for clarity
- intentionally left UUID blank for walrus, but don't know if this is a good plan or not

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1486/

### DIFF
--- a/base/collection.go
+++ b/base/collection.go
@@ -140,13 +140,7 @@ func GetGocbV2BucketFromCluster(cluster *gocb.Cluster, spec BucketSpec, waitUnti
 		clusterCompatMajorVersion: uint64(clusterCompatMajor),
 		clusterCompatMinorVersion: uint64(clusterCompatMinor),
 	}
-	uuid, err := getServerUUID(gocbv2Bucket)
-	if err != nil {
-		_ = cluster.Close(&gocb.ClusterCloseOptions{})
-		return nil, err
-	}
 
-	gocbv2Bucket.ClusterUUID = uuid
 	// Set limits for concurrent query and kv ops
 	maxConcurrentQueryOps := MaxConcurrentQueryOps
 	if spec.MaxConcurrentQueryOps != nil {
@@ -189,7 +183,6 @@ type GocbV2Bucket struct {
 	queryOps                                             chan struct{} // Manages max concurrent query ops
 	kvOps                                                chan struct{} // Manages max concurrent kv ops
 	clusterCompatMajorVersion, clusterCompatMinorVersion uint64        // E.g: 6 and 0 for 6.0.3
-	ClusterUUID                                          string        // uuid of cluster
 }
 
 var (
@@ -450,10 +443,6 @@ func (b *GocbV2Bucket) QueryEpsCount() (int, error) {
 // found, retrieves the cluster-wide value.
 func (b *GocbV2Bucket) MetadataPurgeInterval() (time.Duration, error) {
 	return getMetadataPurgeInterval(b)
-}
-
-func (b *GocbV2Bucket) ServerUUID() (uuid string, err error) {
-	return getServerUUID(b)
 }
 
 func (b *GocbV2Bucket) MaxTTL() (int, error) {
--- a/base/collection.go
+++ b/base/collection.go
@@ -130,7 +130,7 @@ func GetGocbV2BucketFromCluster(cluster *gocb.Cluster, spec BucketSpec, waitUnti
 	clusterCompatMajor, clusterCompatMinor, err := getClusterVersion(cluster)
 	if err != nil {
 		_ = cluster.Close(&gocb.ClusterCloseOptions{})
-		return nil, fmt.Errorf("%s", err)
+		return nil, err
 	}
 
 	gocbv2Bucket := &GocbV2Bucket{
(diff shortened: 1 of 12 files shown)


base/collection.go
```diff
@@ -140,7 +140,13 @@ func GetGocbV2BucketFromCluster(cluster *gocb.Cluster, spec BucketSpec, waitUnti
 		clusterCompatMajorVersion: uint64(clusterCompatMajor),
 		clusterCompatMinorVersion: uint64(clusterCompatMinor),
 	}
+	uuid, err := getServerUUID(gocbv2Bucket)
+	if err != nil {
+		_ = cluster.Close(&gocb.ClusterCloseOptions{})
+		return nil, fmt.Errorf("%s", err)
```
bbrks: Avoid error wrapping without `%w` or just return err so callers can continue to do errors.Is/As handling 

```suggestion
		return nil, fmt.Errorf("Couldn't get server UUID: %w", err)
```

or

```suggestion
		return nil, err
```

db/database_test.go
```diff
@@ -3084,6 +3084,20 @@ func TestGetDatabaseCollectionWithUserDefaultCollection(t *testing.T) {
...
 
+func TestClusterUUID(t *testing.T) {
+	bucket := base.GetTestBucket(t)
+	defer bucket.Close()
+
+	db, err := NewDatabaseContext(base.TestCtx(t), "db", bucket, false, DatabaseContextOptions{})
+	require.NoError(t, err)
+
+	if base.TestUseCouchbaseServer() {
+		require.Equal(t, 32, len(db.clusterUUID)) // no dashes in UUID
```
bbrks: this type of assertion will print out the actual string rather than something like `32 != 36` if it fails

```suggestion
		require.Len(t, db.clusterUUID, 32) // no dashes in UUID
```

db/document.go
```diff
@@ -91,6 +91,7 @@ type SyncData struct {
 
 	addedRevisionBodies     []string          // revIDs of non-winning revision bodies that have been added (and so require persistence)
 	removedRevisionBodyKeys map[string]string // keys of non-winning revisions that have been removed (and so may require deletion), indexed by revID
+	ClusterUUID             string            `json:"cluster_uuid,omitempty"`
```
bbrks: We should move this up above at least `History` in this struct - see comment explaining why.

Sidenote/thought: Is it possible to get test coverage for this "History should be marshalled last" to catch future cases? I'll file a ticket for it.

bbrks (changes requested):

base/collection.go
```diff
@@ -183,6 +189,7 @@ type GocbV2Bucket struct {
 	queryOps                                             chan struct{} // Manages max concurrent query ops
 	kvOps                                                chan struct{} // Manages max concurrent kv ops
 	clusterCompatMajorVersion, clusterCompatMinorVersion uint64        // E.g: 6 and 0 for 6.0.3
+	ClusterUUID                                          string        // uuid of cluster
```
adamcfraser: What's the benefit of defining ClusterUUID on GocbV2Bucket, instead using the version we already have on DatabaseContext (the lazy loading done in GetServerUUID())

base/collection.go
```diff
@@ -140,7 +140,13 @@ func GetGocbV2BucketFromCluster(cluster *gocb.Cluster, spec BucketSpec, waitUnti
 		clusterCompatMajorVersion: uint64(clusterCompatMajor),
 		clusterCompatMinorVersion: uint64(clusterCompatMinor),
 	}
+	uuid, err := getServerUUID(gocbv2Bucket)
+	if err != nil {
```
adamcfraser: I don't think existing usages of getServerUUID are fatal on failed retrieval.  I'm not sure it needs to be in this case, but if it does, I suspect getServerUUID needs to be made more robust (to handle transient failures).

adamcfraser (changes requested):

base/collection.go
torcolvin: The benefit here was that I would do it directly after making a call for a connection to a bucket so it would be less inclined toward failure because you just wouldn't a bucket if you couldn't immediately make a gocb connection and hit this end point.

torcolvin: Moved to NewDatabaseContext with a somewhat arbitrary timeout, used elsewhere in the file, but it's actually used for bootstrap connection and not the NewDatabaseContext. If you have better ideas on numbers to choose, I'm all ears.

base/collection.go
torcolvin: Added a retry loop and we now only call it in one place.

db/database.go
```diff
@@ -129,6 +129,7 @@ type DatabaseContext struct {
 	singleCollection             *DatabaseCollection            // Temporary collection
 	CollectionByID               map[uint32]*DatabaseCollection // A map keyed by collection ID to Collection
 	CollectionNames              map[string]map[string]struct{} // Map of scope, collection names
+	ClusterUUID                  string                         // uuid of cluster
```
adamcfraser: Do we have some duplication between ClusterUUID/getClusterUUID and serverUUID/GetServerUUID, or am I misunderstanding the difference?

adamcfraser (changes requested):

db/database.go
torcolvin: just a mistake, settled on ServerUUID as we use it elsewhere.

adamcfraser (approved):

PR #6091, opened 2023-02-10
CBG-2688 Modify sync function for collections

torcolvin:
- rename `DefaultSyncFunction` to `DocChannelsSyncFunction` which isn't a great name
- remove NewDefaultChannelMapper since it is not collection aware
- updated tests to use `UpdateSyncFn or pass SyncFn via rest tester.
- Removed ChannelMapper from `DatabaseContext` to always be collection scoped. This also means constructing `NewDatabaseContext` that `Scopes` is now a required field - this actually simplifies the code a good bit. It does move _some_ validation into `NewDatabaseContext` when we set the import filter which was formerly done in `ServerContext` but I think it makes it more possible to test sync and import in the `db` package.
- Changed the name of the sync function key to not store kv collection id, since in a DR unidirectional XDCR this value will be wrong, and we shouldn't write any docs with the kv collection id in them. Use the same synchronization field for `_default._default` for backward compatibility.
- Preserve no js passthrough behavior in `crud.go` if `DatabaseCollection.ChannelMapper = nil`

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2884/

### DIFF
--- a/base/dcp_sharded.go
+++ b/base/dcp_sharded.go
@@ -61,7 +61,7 @@ func StartShardedDCPFeed(ctx context.Context, dbName string, configGroup string,
 		return nil, fmt.Errorf("failed to get minimum node version in cluster: %w", err)
 	}
 	if minVersion.Less(firstVersionToSupportCollections) {
-		if scope != "" || len(collections) > 0 {
+		if scope != DefaultScope || len(collections) > 0 {
 			return nil, fmt.Errorf("cannot start DCP feed on non-default collection with legacy nodes present in the cluster")
 		}
 	}
--- a/base/dcp_sharded.go
+++ b/base/dcp_sharded.go
@@ -62,7 +62,7 @@ func StartShardedDCPFeed(ctx context.Context, dbName string, configGroup string,
 	}
 	if minVersion.Less(firstVersionToSupportCollections) {
 		// DefaultScope is allowed by older versions of CBGT as long as no collections are specified.
-		if scope != DefaultScope || len(collections) > 0 {
+		if len(collections) > 0 {
 			return nil, fmt.Errorf("cannot start DCP feed on non-default collection with legacy nodes present in the cluster")
 		}
 	}
--- a/base/dcp_sharded.go
+++ b/base/dcp_sharded.go
@@ -61,6 +61,7 @@ func StartShardedDCPFeed(ctx context.Context, dbName string, configGroup string,
 		return nil, fmt.Errorf("failed to get minimum node version in cluster: %w", err)
 	}
 	if minVersion.Less(firstVersionToSupportCollections) {
+		// DefaultScope is allowed by older versions of CBGT as long as no collections are specified.
 		if scope != DefaultScope || len(collections) > 0 {
 			return nil, fmt.Errorf("cannot start DCP feed on non-default collection with legacy nodes present in the cluster")
 		}
--- a/db/attachment_test.go
+++ b/db/attachment_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 
 	"github.com/couchbase/sync_gateway/base"
+	"github.com/couchbase/sync_gateway/channels"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -214,15 +215,15 @@ func TestAttachmentForRejectedDocument(t *testing.T) {
 	defer db.Close(ctx)
 
 	collection := GetSingleDatabaseCollectionWithUser(t, db)
-	_, err := collection.UpdateSyncFun(ctx, `function(doc, oldDoc) {
+	collection.ChannelMapper = channels.NewChannelMapper(
+		`function(doc, oldDoc) {
 		throw({forbidden: "None shall pass!"});
-	}`)
-	require.NoError(t, err)
+	}`, db.Options.JavascriptTimeout)
 
 	docBody := `{"_attachments": {"hello.txt": {"data":"aGVsbG8gd29ybGQ="}}}`
 	var body Body
 	require.NoError(t, base.JSONUnmarshal([]byte(docBody), &body))
-	_, _, err = collection.Put(ctx, "doc1", unjson(docBody))
+	_, _, err := collection.Put(ctx, "doc1", unjson(docBody))
 	require.Error(t, err)
 
 	// Attempt to retrieve the attachment doc
--- a/db/attachment_test.go
+++ b/db/attachment_test.go
@@ -22,7 +22,6 @@ import (
 	"testing"
 
 	"github.com/couchbase/sync_gateway/base"
-	"github.com/couchbase/sync_gateway/channels"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -55,13 +54,7 @@ func TestBackupOldRevisionWithAttachments(t *testing.T) {
 			Enabled:          deltasEnabled,
 			RevMaxAgeSeconds: DefaultDeltaSyncRevMaxAge,
 		},
-		Scopes: map[string]ScopeOptions{
-			base.DefaultScope: ScopeOptions{
-				Collections: map[string]CollectionOptions{
-					base.DefaultCollection: {},
-				},
-			},
-		},
+		Scopes: GetScopesOptions(t, bucket, 1),
 	})
 	assert.NoError(t, err, "Couldn't create context for database 'db'")
 	defer dbCtx.Close(ctx)
@@ -230,14 +223,15 @@ func TestAttachmentForRejectedDocument(t *testing.T) {
 	defer db.Close(ctx)
 
 	collection := GetSingleDatabaseCollectionWithUser(t, db)
-	collection.ChannelMapper = channels.NewChannelMapper(`function(doc, oldDoc) {
+	_, err := collection.UpdateSyncFun(ctx, `function(doc, oldDoc) {
 		throw({forbidden: "None shall pass!"});
-	}`, 0)
+	}`)
+	require.NoError(t, err)
 
 	docBody := `{"_attachments": {"hello.txt": {"data":"aGVsbG8gd29ybGQ="}}}`
 	var body Body
 	require.NoError(t, base.JSONUnmarshal([]byte(docBody), &body))
-	_, _, err := collection.Put(ctx, "doc1", unjson(docBody))
+	_, _, err = collection.Put(ctx, "doc1", unjson(docBody))
 	require.Error(t, err)
 
 	// Attempt to retrieve the attachment doc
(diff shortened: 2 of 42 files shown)


base/dcp_sharded.go
```diff
@@ -61,7 +61,7 @@ func StartShardedDCPFeed(ctx context.Context, dbName string, configGroup string,
 		return nil, fmt.Errorf("failed to get minimum node version in cluster: %w", err)
 	}
 	if minVersion.Less(firstVersionToSupportCollections) {
-		if scope != "" || len(collections) > 0 {
+		if scope != DefaultScope || len(collections) > 0 {
```
adamcfraser: It feels inconsistent that we're requiring non-empty scope name here, but not including `_default` in collections.  If the default scopename has to be provided, should we not always include the default collection name as well?

db/database.go
```diff
@@ -438,67 +436,62 @@ func NewDatabaseContext(ctx context.Context, dbName string, bucket base.Bucket,
...
-					syncFn = *collOpts.Sync
+				if fnChanged {
+					syncFunctionsChanged = true
 				}
 
-				if collOpts.ImportFilter != nil {
-					dbCollection.importFilterFunction = collOpts.ImportFilter
-				}
+			} else {
+				defaultSyncFunction := channels.GetDefaultSyncFunction(scopeName, collName)
```
adamcfraser, line 498: This looks like it's going to force javascript execution for the default sync function for collections - is that intended?

db/attachment_test.go
```diff
@@ -55,6 +54,7 @@ func TestBackupOldRevisionWithAttachments(t *testing.T) {
 			Enabled:          deltasEnabled,
 			RevMaxAgeSeconds: DefaultDeltaSyncRevMaxAge,
 		},
+		Scopes: GetScopesOptions(t, bucket, 1),
```
adamcfraser: This looks like it forces this test to not be compatible when running with just the default collection?  If that's not the case, why is it necessary to explicitly define this in the DatabaseContextOptions?

db/database.go
```diff
@@ -438,67 +436,62 @@ func NewDatabaseContext(ctx context.Context, dbName string, bucket base.Bucket,
 	// Initialize the tap Listener for notify handling
 	dbContext.mutationListener.Init(bucket.GetName(), options.GroupID)
 
-	if len(options.Scopes) > 0 {
-		dbContext.Scopes = make(map[string]Scope, len(options.Scopes))
-		dbContext.CollectionNames = make(map[string]map[string]struct{}, len(options.Scopes))
-		for scopeName, scope := range options.Scopes {
-			dbContext.Scopes[scopeName] = Scope{
-				Collections: make(map[string]*DatabaseCollection, len(scope.Collections)),
+	if len(options.Scopes) == 0 {
```
adamcfraser, line 462: Is it possible to initialize dbContext.Scopes to {_default._default} if it's not specified in options.Scopes, instead of requiring all callers to initialize in the default case?

db/database_collection.go
```diff
@@ -191,6 +181,11 @@ func (c *DatabaseCollection) IsClosed() bool {
 	return c.dataStore == nil
 }
 
+// IsDefaultCollection returns true if collection is _default._default.
+func (c *DatabaseCollection) IsDefaultCollection() bool {
+	return c.GetCollectionID() == base.DefaultCollectionID
```
adamcfraser: The fact that base.DefaultCollectionID is the zero value for this property makes me a bit nervous - this check relies on setCollectionID() being called, which only happens if the collection was created using the NewCollection constructor.  (this is always the case today, but there's the potential for future regressions)

I'd expect this to only be used in performance-critical paths (where checking base.IsDefaultCollection(scopeName, collectionName) would be too slow).  Is that the case here?

db/database_stats.go
```diff
@@ -21,10 +21,6 @@ package db
 
 // Update database-specific stats that are more efficiently calculated at stats collection time
 func (db *DatabaseContext) UpdateCalculatedStats() {
```
adamcfraser: I'm unclear how this change is related to the rest of the PR - is this catching an unrelated bug?

db/dcp_sharded_upgrade_test.go
```diff
@@ -224,6 +224,7 @@ func TestShardedDCPUpgrade(t *testing.T) {
 		ImportOptions: ImportOptions{
 			ImportPartitions: numPartitions,
 		},
+		Scopes: GetScopesOptionsDefaultCollectionOnly(t),
```
adamcfraser, line 228: As discussed elsewhere, seems like we could set the default in NewDatabaseContext if DatabaseContextOptions.Scopes isn't populated.

db/query_test.go
```diff
@@ -268,9 +271,10 @@ func TestAccessQuery(t *testing.T) {
 	defer db.Close(ctx)
 	collection := GetSingleDatabaseCollectionWithUser(t, db)
 
-	db.ChannelMapper = channels.NewChannelMapper(`function(doc, oldDoc) {
+	_, err := collection.UpdateSyncFun(ctx, `function(doc, oldDoc) {
```
adamcfraser: UpdateSyncFun is going to write the sync function document to the database.  Setting the ChannelMapper (as the test was previously doing) is a bit more lightweight for test purposes - any reason we shouldn't just modify collection.ChannelMapper directly here?

db/sg_replicate_cfg_test.go
```diff
@@ -649,16 +649,17 @@ func TestReplicateGroupIDAssignedNodes(t *testing.T) {
 	defer tb.Close()
 	ctx := base.TestCtx(t)
 
+	const numCollections = 1
 	// Set up databases
-	dbDefault, err := NewDatabaseContext(ctx, "default", tb.NoCloseClone(), false, DatabaseContextOptions{GroupID: ""})
+	dbDefault, err := NewDatabaseContext(ctx, "default", tb.NoCloseClone(), false, DatabaseContextOptions{GroupID: "", Scopes: GetScopesOptions(t, tb, 1)})
```
adamcfraser: Did this test work when running with named collections previously?  If yes, was this databaseContext being set up to use `_default._default`?

I agree it's preferable to have the test use a named collection (via GetScopesOptions) if that's how the suite is running, but it also feels like the amount of boilerplate required to just call NewDatabaseContext (with the addition of GetScopesOptions) is getting a bit cumbersome.

db/util_testing.go
```diff
@@ -525,8 +511,14 @@ func SetupTestDBForDataStoreWithOptions(t testing.TB, tBucket *base.TestBucket,
 	return db, ctx
 }
 
-// getScopesOptions sets up a ScopesOptions from a TestBucket for use with non default collections to pass in DatabaseContextOptions.
-func getScopesOptions(t testing.TB, testBucket *base.TestBucket, numCollections int) ScopesOptions {
+// GetScopesOptions sets up a ScopesOptions from a TestBucket for use with non default collections to pass in DatabaseContextOptions.
```
adamcfraser: I think the 'non-default collections' part of this comment should now be removed?

rest/utilities_testing.go
```diff
@@ -280,7 +280,11 @@ func (rt *RestTester) Bucket() base.Bucket {
 			// If scopes is already set, assume the caller has a plan
 			if rt.DatabaseConfig.Scopes == nil {
 				// Configure non default collections by default
-				rt.DatabaseConfig.Scopes = GetCollectionsConfig(rt.TB, testBucket, rt.numCollections)
+				syncFn := base.StringPtr(rt.SyncFn)
```
adamcfraser: I feel like this would be more readable as:
```
var syncFn *string
if rt.SyncFn != nil {
   syncFn = base.StringPtr(rt.SyncFn)
}
```
As written it makes me wonder "why are we calling StringPtr when we know we don't care about empty values?"

adamcfraser (changes requested):
Overall approach looks good, some questions around some details.

db/database.go
torcolvin: This doesn't set the sync function, it just logs it for the user on info level.

db/attachment_test.go
torcolvin: I was thinking in general it's better to have it run against all collections if possible.

db/database_stats.go
torcolvin: Different PR, I put it in https://github.com/couchbase/sync_gateway/pull/6097

base/dcp_sharded.go
torcolvin: Added a comment here.

db/database_collection.go
torcolvin: I think it was an overoptimization, it's a convenience function.

db/sg_replicate_cfg_test.go
torcolvin: Made this a constant here, hopefully to add clarity.

db/dcp_sharded_upgrade_test.go
torcolvin: Commented here why this is explicitly necessary.

db/attachment_test.go
torcolvin: Switched to use a convenience function.

db/database.go
torcolvin: I don't want to do this, because I don't think that we often want to test with the default collection explicitly. There are a few circumstances (testing legacy attachment behavior, upgrade behavior) but mostly we want to be able to have our databases backed by collections determined by the test harness.

This is a summary of what was discussed offline.

base/dcp_sharded.go
adamcfraser: That doesn't look like a cbgt thing - is it just the collections > 0 check we're doing in cbgtFeedParams?  Assuming that the case, should we just check collections > 0 here as well?

db/dcp_sharded_upgrade_test.go
adamcfraser: Makes sense.

db/query_test.go
adamcfraser: @torcolvin This question is still open - would like to know for future reference whether we can just continue to modify the ChannelMapper directly.

adamcfraser (changes requested):

base/dcp_sharded.go
torcolvin: Yeah that's fine.

db/query_test.go
adamcfraser: Discussed offline, for future reference we can continue to take either approach, UpdateSyncFun or direct ChannelMapper updates.

adamcfraser (approved):

adamcfraser:
The CI failure on TestConcurrentRefreshUser doesn't look directly related to these changes - it looks like channel access is being lost in between sending the changes message and handling the corresponding rev request, and I'm not able to reproduce the failure locally - kicked off a re-run.

PR #6092, opened 2023-02-13
Maintain multiple activeSubChanges, one per collection

torcolvin:
- maintain a mapping of CollectionContexts rather than collections
- move shutdown of subChanges go routines to `stopAndDisconnect` to avoid data race, since we have the potential to call `apr.blipSyncContext = nil` before we are actually done waiting.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1501/

### DIFF
--- a/db/active_replicator_common.go
+++ b/db/active_replicator_common.go
@@ -215,6 +215,7 @@ func (a *activeReplicatorCommon) stopSubChanges() {
 	t := time.NewTicker(retryInterval)
 	defer t.Stop()
 
+	collectionCtxs := a.blipSyncContext.collections.getAll()
 retry:
 	for {
 		select {
@@ -222,15 +223,11 @@ retry:
 			break retry
 		case <-t.C:
 			// wait until all activeSubChanges have stopped
-			for _, collectionCtx := range a.blipSyncContext.collectionContexts {
+			for _, collectionCtx := range collectionCtxs {
 				if collectionCtx.activeSubChanges.IsTrue() {
 					continue retry
 				}
 			}
-			if a.blipSyncContext.nonCollectionAwareContext.activeSubChanges.IsTrue() {
-				continue retry
-			}
-			break retry
 		}
 	}
 }
--- a/db/active_replicator_common.go
+++ b/db/active_replicator_common.go
@@ -210,7 +210,7 @@ func (a *activeReplicatorCommon) stopSubChanges() {
 		maxWaitTime   = 10 * time.Second
 		retryInterval = 10 * time.Millisecond
 	)
-	ctx, cancel := context.WithDeadline(a.ctx, time.Now().Add(maxWaitTime))
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(maxWaitTime))
 	defer cancel()
 	t := time.NewTicker(retryInterval)
 	defer t.Stop()
--- a/db/active_replicator_common.go
+++ b/db/active_replicator_common.go
@@ -194,12 +194,45 @@ func (a *activeReplicatorCommon) _disconnect() error {
 	}
 
 	if a.blipSyncContext != nil {
+		a.stopSubChanges()
 		base.TracefCtx(a.ctx, base.KeyReplicate, "closing blip sync context")
 		a.blipSyncContext.Close()
 		a.blipSyncContext = nil
 	}
 
 	return nil
+}
+
+// stopSubChanges stops any active subChanges requests
+func (a *activeReplicatorCommon) stopSubChanges() {
+	base.TracefCtx(a.ctx, base.KeyReplicate, "closing active subChanges")
+	const (
+		maxWaitTime   = 10 * time.Second
+		retryInterval = 10 * time.Millisecond
+	)
+	ctx, cancel := context.WithDeadline(a.ctx, time.Now().Add(maxWaitTime))
+	defer cancel()
+	t := time.NewTicker(retryInterval)
+	defer t.Stop()
+
+retry:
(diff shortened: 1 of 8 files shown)


db/blip_handler.go
```diff
@@ -149,9 +149,12 @@ func collectionBlipHandler(next blipHandlerFunc) blipHandlerFunc {
 				DatabaseCollection: bh.db.singleCollection,
 				user:               bh.db.user,
 			}
+			if bh.nonCollectionAwareContext == nil {
+				bh.nonCollectionAwareContext = &BlipSyncCollectionContext{dbCollection: bh.collection.DatabaseCollection}
```
bbrks: This seems to be racing in Jenkins. I think because two BLIP messages mean this handler can be called simultaneously.

Not sure what the correct course of action is here without a deeper look.

bbrks (changes requested):
Approach seems reasonable - including the changed stop/disconnect.

Think the `nonCollectionAwareContext` initialization needs some synchronization to keep the race detector happy though.

db/blip_handler.go
torcolvin: Wow, I didn't see that at all, and it didn't repro for me at all, thanks for pointing it out.

db/active_replicator_common.go
```diff
@@ -202,6 +203,35 @@ func (a *activeReplicatorCommon) _disconnect() error {
...
 }
 
+// stopSubChanges stops any active subChanges requests
+func (a *activeReplicatorCommon) stopSubChanges() {
+	base.TracefCtx(a.ctx, base.KeyReplicate, "closing active subChanges")
+	const (
+		maxWaitTime   = 10 * time.Second
+		retryInterval = 10 * time.Millisecond
+	)
+	ctx, cancel := context.WithDeadline(a.ctx, time.Now().Add(maxWaitTime))
```
bbrks: Using `a.ctx` here as the parent I think means it'll be cancelled from `a._stop()` before disconnect/this function is called, resulting in an immediate timeout for the retry loop instead of waiting for the deadline.

bbrks (changes requested):
I missed this on the first pass but I think use of context means the retry loop doesn't wait properly

db/active_replicator_common.go
torcolvin: That's probably how I saw the data race in the first place :)

bbrks:
@torcolvin is the timeout related? I see unsubChanges code in the stacktrace

I can't tell whether this change has caused the tests to take longer to complete given the ~20minute GH Actions test runs. I thought they were much quicker.

torcolvin:
> @torcolvin is the timeout related? I see unsubChanges code in the stacktrace
> 
> I can't tell whether this change has caused the tests to take longer to complete given the ~20minute GH Actions test runs. I thought they were much quicker.

After further investigation into this, I realised this code never worked since its inception in https://github.com/couchbase/sync_gateway/commit/35f70c043114cc0feb697c505e44ca7d628f3fd6. I think the intention of that commit fixed a potential race in the pull replicator, but in the push replicator, the `apr.blipSyncContext` was always nil, so those requests don't get swapped to close. Adding in a working version of the context meant that it would just wait 10s per pull replication to close. I reverted (removed) the code to the functional format, just close `blipSyncContext` and don't wait for any subsequent changes.

db/blip_handler.go
```diff
@@ -304,7 +305,11 @@ func (bh *blipHandler) handleSubChanges(rq *blip.Message) error {
 
 		defer func() {
 			bh.changesCtxCancel()
-			bh.activeSubChanges.Set(false)
+			collectionCtx, err := bh.collections.get(bh.collectionIdx)
+			if err != nil {
+				base.WarnfCtx(bh.loggingCtx, "Error getting collectionContext on stopping subChanges request: %w", err)
+			}
```
bbrks: Nit: Is it not possible to use `collectionCtx` or `collectionCtx.activeSubChanges` from the outer scope (line 252) to avoid having to get twice and deal with potential error handling?

db/blip_collection_context.go
```diff
@@ -0,0 +1,98 @@
...
+
+// set adds a set of collections to this contexts struct.
+func (b *blipCollections) set(collectionCtxs []*blipSyncCollectionContext) {
+	b.Lock()
+	defer b.Unlock()
+	b.collectionContexts = collectionCtxs
+}
+
+// getCollectionContext returns a collection matching the blip collection idx set by the initial GetCollections handshake. If collectionIdx is nil, assume that the messages are not collection aware.
+func (b *blipCollections) get(collectionIdx *int) (*blipSyncCollectionContext, error) {
```
bbrks, line 48: I'm cautious about the amount of locking we're having to do each time we call one of these accessors, but changing things to pass around context references instead of indexes into the slice seems like a decent chunk of refactoring, so until it shows up as a problem I'll just 👍 it

bbrks (dismissed):
I only have nitpicks or theoretical comments, so I think this can go in if you're happy

db/blip_handler.go
torcolvin: This is already defined in this function outside this goroutine, removed duplicate.

db/blip_collection_context.go
torcolvin: This is a real mess when dealing with how it's handled in `sendChanges` from ISGR and I didn't feel like I could test it well enough to be sure I was doing it right without being able to test multiple collections in ISGR.

bbrks (approved):

PR #6093, opened 2023-02-14
API Spec: Tweak wording for admin /_config descriptions

bbrks:
The endpoint returns the initial startup config by default, and only the actively running config if `include_runtime` is set.


IsaacLambat (approved):
LGTM

PR #6095, opened 2023-02-14
CBG-760 hash user to invalidate sessions

torcolvin:
- Hash password in order to invalidate sessions.
- Kept `QuerySessions` for `DELETE /db/_user/username/session` 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1504/

### DIFF
--- a/auth/principal.go
+++ b/auth/principal.go
@@ -88,7 +88,7 @@ type User interface {
 	// GetSessionUUID returns the UUID that a session to match to be a valid session.
 	GetSessionUUID() string
 
-	// UpdateSessionUUID creates a new session UUID
+	// UpdateSessionUUID will invalidate all earlier sessions associated with this user.
 	UpdateSessionUUID()
 
 	// Changes the user's password.
--- a/auth/principal.go
+++ b/auth/principal.go
@@ -86,7 +86,7 @@ type User interface {
 	Authenticate(password string) bool
 
 	// GetSessionUUID returns the UUID that a session to match to be a valid session.
-	GetSessionUUID() []byte
+	GetSessionUUID() string
 
 	// UpdateSessionUUID creates a new session UUID
 	UpdateSessionUUID()
--- a/auth/principal.go
+++ b/auth/principal.go
@@ -88,7 +88,7 @@ type User interface {
 	// GetSessionUUID returns the UUID that a session to match to be a valid session.
 	GetSessionUUID() string
 
-	// UpdateSessionUUID will invalidate all earlier sessions associated with this user.
+	// UpdateSessionUUID creates a new session UUID
 	UpdateSessionUUID()
 
 	// Changes the user's password.
--- a/auth/principal.go
+++ b/auth/principal.go
@@ -85,6 +85,9 @@ type User interface {
 	// Authenticates the user's password.
 	Authenticate(password string) bool
 
+	// GetPasswordHash returns the hashed password.
+	GetPasswordHash() []byte
+
 	// Changes the user's password.
 	SetPassword(password string) error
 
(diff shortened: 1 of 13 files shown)


adamcfraser:
@torcolvin Assigning back to you based on our discussion about avoiding the query even in the REST API case.

torcolvin:
I didn't remove the obsolete index in the upgrade pathway, because I was unsure about prior art for doing so. As of right now, we only remove xattr / non xattr index. I also wasn't sure if we wanted to remove it if a user needs to upgrade to helium and then downgrade.

auth/principal.go
```diff
@@ -85,6 +85,12 @@ type User interface {
 	// Authenticates the user's password.
 	Authenticate(password string) bool
 
+	// GetSessionUUID returns the UUID that a session to match to be a valid session.
+	GetSessionUUID() []byte
```
adamcfraser: Is there a reason to store this as []byte instead of string?

auth/session.go
```diff
@@ -67,6 +69,13 @@ func (auth *Authenticator) AuthenticateCookie(rq *http.Request, response http.Re
 	user, err := auth.GetUser(session.Username)
 	if user != nil && user.Disabled() {
 		user = nil
+		return user, err
+	}
+	if err != nil {
+		return nil, err
+	}
+	if !bytes.Equal(session.SessionUUID, user.GetSessionUUID()) {
+		return nil, base.HTTPErrorf(http.StatusUnauthorized, "User has changed since session acquired")
```
adamcfraser: I think a better message might be "Session no longer valid for user".

auth/session.go
```diff
@@ -82,11 +91,19 @@ func (auth *Authenticator) CreateSession(username string, ttl time.Duration) (*L
 		return nil, err
 	}
 
+	user, err := auth.GetUser(username)
+	if user != nil && user.Disabled() {
+		return nil, base.HTTPErrorf(400, "User is disabled")
```
adamcfraser, line 91: Is the disabled handling required here?  I don't think we have a reason to remove the ability to create sessions for disabled users?

auth/session.go
```diff
@@ -82,11 +91,19 @@ func (auth *Authenticator) CreateSession(username string, ttl time.Duration) (*L
...
+	} else if err != nil {
+		return nil, err
```
adamcfraser, line 93: Do we need to prevent creation of sessions for non-existing users, or could we consider empty as the initial state of session UUID?  (presumably that's also needed for backward compatibility for existing users)

adamcfraser (changes requested):

auth/principal.go
torcolvin: Bad copy paste, fixed.

auth/session.go
torcolvin: fixed.

auth/session.go
torcolvin: I think I copy/pasted from a different function, they'd fail in authentication anyway so this is fine to remove.

auth/session.go
torcolvin: I am not sure I understand this question. I added a case to test where the SessionUUID is empty to make backward compatible with existing users.

adamcfraser: Previously we didn't require the user to exist in order to create a session.  I don't think that needs to change, if we treat "empty" as the initial session UUID value.

torcolvin: Decided to not worry about this since the session API prevents you from creating a session before a user. I did a follow up commit so that we don't read the user data twice when creating a session as an enhancement.

adamcfraser (approved):

PR #6098, opened 2023-02-16
CBG-2676: Expose internal ISGR checkpointer list length stats

mohammed-madi:
CBG-2676

Describe your PR here...
- dbStats can now be passed when creating a checkpointer, and it will store expected/processed sequence lists lengths, pre and post cleanup.
- Tests that assert on those stats in the checkpointer stats were changes to assert on them in the dbstats

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1506/

### DIFF
--- a/db/active_replicator_checkpointer.go
+++ b/db/active_replicator_checkpointer.go
@@ -59,17 +59,19 @@ type Checkpointer struct {
 
 	// closeWg waits for the time-based checkpointer goroutine to finish.
 	closeWg sync.WaitGroup
+	// For logging stats about the length of expected and processed sequence lists
+	dbStats *base.DbReplicatorStats
 }
 
 type statusFunc func(lastSeq string) *ReplicationStatus
 
 type CheckpointerStats struct {
 	ExpectedSequenceCount           int64
-	ExpectedSequenceLen             int
-	ExpectedSequenceLenPostCleanup  int
+	ExpectedSequenceLen             int64
+	ExpectedSequenceLenPostCleanup  int64
 	ProcessedSequenceCount          int64
-	ProcessedSequenceLen            int
-	ProcessedSequenceLenPostCleanup int
+	ProcessedSequenceLen            int64
+	ProcessedSequenceLenPostCleanup int64
 	AlreadyKnownSequenceCount       int64
 	SetCheckpointCount              int64
 	GetCheckpointHitCount           int64
@@ -88,6 +90,7 @@ func NewCheckpointer(ctx context.Context, clientID string, configHash string, bl
 		checkpointInterval:             replicatorConfig.CheckpointInterval,
 		ctx:                            ctx,
 		stats:                          CheckpointerStats{},
+		dbStats:                        replicatorConfig.ReplicationStatsMap,
 		statusCallback:                 statusCallback,
 		expectedSeqCompactionThreshold: defaultExpectedSeqCompactionThreshold,
 	}
@@ -258,9 +261,8 @@ func (c *Checkpointer) _updateCheckpointLists() (safeSeq *SequenceID) {
 	base.TracefCtx(c.ctx, base.KeyReplicate, "checkpointer: _updateCheckpointLists(expectedSeqs: %v, processedSeqs: %v)", c.expectedSeqs, c.processedSeqs)
 	base.TracefCtx(c.ctx, base.KeyReplicate, "Inside update checkpoint lists")
 
-	c.stats.ExpectedSequenceLen = len(c.expectedSeqs)
-	c.stats.ProcessedSequenceLen = len(c.processedSeqs)
-
+	c.stats.ExpectedSequenceLen = int64(len(c.expectedSeqs))
+	c.stats.ProcessedSequenceLen = int64(len(c.processedSeqs))
 	maxI := c._calculateSafeExpectedSeqsIdx()
 
 	if maxI > -1 {
@@ -295,8 +297,13 @@ func (c *Checkpointer) _updateCheckpointLists() (safeSeq *SequenceID) {
 		}
 	}
 
-	c.stats.ExpectedSequenceLenPostCleanup = len(c.expectedSeqs)
-	c.stats.ProcessedSequenceLenPostCleanup = len(c.processedSeqs)
+	c.stats.ExpectedSequenceLenPostCleanup = int64(len(c.expectedSeqs))
+	c.stats.ProcessedSequenceLenPostCleanup = int64(len(c.processedSeqs))
+
+	c.dbStats.ProcessedSequenceLen.Set(c.stats.ProcessedSequenceLen)
+	c.dbStats.ProcessedSequenceLen.Set(c.stats.ProcessedSequenceLenPostCleanup)
+	c.dbStats.ExpectedSequenceLen.Set(c.stats.ExpectedSequenceLen)
+	c.dbStats.ExpectedSequenceLenPostCleanup.Set(c.stats.ExpectedSequenceLenPostCleanup)
 
 	return safeSeq
 }
--- a/db/active_replicator_checkpointer.go
+++ b/db/active_replicator_checkpointer.go
@@ -300,11 +300,12 @@ func (c *Checkpointer) _updateCheckpointLists() (safeSeq *SequenceID) {
 	c.stats.ExpectedSequenceLenPostCleanup = int64(len(c.expectedSeqs))
 	c.stats.ProcessedSequenceLenPostCleanup = int64(len(c.processedSeqs))
 
-	c.dbStats.ProcessedSequenceLen.Set(c.stats.ProcessedSequenceLen)
-	c.dbStats.ProcessedSequenceLenPostCleanup.Set(c.stats.ProcessedSequenceLenPostCleanup)
-	c.dbStats.ExpectedSequenceLen.Set(c.stats.ExpectedSequenceLen)
-	c.dbStats.ExpectedSequenceLenPostCleanup.Set(c.stats.ExpectedSequenceLenPostCleanup)
-
+	if c.dbStats != nil {
+		c.dbStats.ProcessedSequenceLen.Set(c.stats.ProcessedSequenceLen)
+		c.dbStats.ProcessedSequenceLenPostCleanup.Set(c.stats.ProcessedSequenceLenPostCleanup)
+		c.dbStats.ExpectedSequenceLen.Set(c.stats.ExpectedSequenceLen)
+		c.dbStats.ExpectedSequenceLenPostCleanup.Set(c.stats.ExpectedSequenceLenPostCleanup)
+	}
 	return safeSeq
 }
 
--- a/db/active_replicator_checkpointer.go
+++ b/db/active_replicator_checkpointer.go
@@ -301,7 +301,7 @@ func (c *Checkpointer) _updateCheckpointLists() (safeSeq *SequenceID) {
 	c.stats.ProcessedSequenceLenPostCleanup = int64(len(c.processedSeqs))
 
 	c.dbStats.ProcessedSequenceLen.Set(c.stats.ProcessedSequenceLen)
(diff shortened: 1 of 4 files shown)


db/active_replicator_checkpointer.go
```diff
@@ -295,9 +297,15 @@ func (c *Checkpointer) _updateCheckpointLists() (safeSeq *SequenceID) {
...
 	}
 
-	c.stats.ExpectedSequenceLenPostCleanup = len(c.expectedSeqs)
-	c.stats.ProcessedSequenceLenPostCleanup = len(c.processedSeqs)
+	c.stats.ExpectedSequenceLenPostCleanup = int64(len(c.expectedSeqs))
+	c.stats.ProcessedSequenceLenPostCleanup = int64(len(c.processedSeqs))
 
+	if c.dbStats != nil {
+		c.dbStats.ProcessedSequenceLen.Set(c.stats.ProcessedSequenceLen)
+		c.dbStats.ProcessedSequenceLenPostCleanup.Set(c.stats.ProcessedSequenceLenPostCleanup)
```
bbrks: I don't think we need two sets of stats like this.

What I was expecting to see was the `CheckpointerStats struct` containing pointers to the underlying `SgwIntStat`s inside `DbReplicationStats` so they can be directly updated from this code.

I.e. when you initialize CheckpointerStats, you set the pointers to those in `replicatorConfig.ReplicationStatsMap` so they're linked directly.

bbrks (changes requested):

db/active_replicator_checkpointer.go
mohammed-madi: Makes sense, https://github.com/couchbase/sync_gateway/pull/6098/commits/bac1189cc8b7b646408e4c71baa3f20f27984066
Ill also rerun jenkins

db/active_replicator_checkpointer.go
```diff
@@ -78,16 +80,22 @@ type CheckpointerStats struct {
...
+		configHash:         configHash,
+		blipSender:         blipSender,
+		activeDB:           replicatorConfig.ActiveDB,
+		expectedSeqs:       make([]SequenceID, 0),
+		processedSeqs:      make(map[SequenceID]struct{}),
+		idAndRevLookup:     make(map[IDAndRev]SequenceID),
+		checkpointInterval: replicatorConfig.CheckpointInterval,
+		ctx:                ctx,
+		stats: CheckpointerStats{
+			ProcessedSequenceLen:            *replicatorConfig.ReplicationStatsMap.ProcessedSequenceLen,
```
bbrks: Dereferencing these here means the Checkpointer stats will be a copy of the original stat - any increments done on c.stats will not be reflected in the ReplicationStatsMap instances.

db/active_replicator_checkpointer.go
```diff
@@ -258,9 +266,10 @@ func (c *Checkpointer) _updateCheckpointLists() (safeSeq *SequenceID) {
 	base.TracefCtx(c.ctx, base.KeyReplicate, "checkpointer: _updateCheckpointLists(expectedSeqs: %v, processedSeqs: %v)", c.expectedSeqs, c.processedSeqs)
 	base.TracefCtx(c.ctx, base.KeyReplicate, "Inside update checkpoint lists")
 
-	c.stats.ExpectedSequenceLen = len(c.expectedSeqs)
-	c.stats.ProcessedSequenceLen = len(c.processedSeqs)
-
+	if c.dbStats != nil {
+		c.dbStats.ProcessedSequenceLen.Set(int64(len(c.processedSeqs)))
```
bbrks: Is the alternative approach of storing pointers in CheckpointerStats and not having to keep a c.dbStats reference around better? I think it simplifies the code setting the stats values as all of them will be in the same place (`c.stats`)

bbrks (changes requested):
We're not keeping references of stats correctly here, and are actually creating copies of stats, assuming the intention is to be accessing via `c.stats` rather than `c.dbStats`

The previous comment suggested the right approach ("`CheckpointerStats struct` containing pointers to the underlying `SgwIntStats` inside `DbReplicationStats`")

db/active_replicator_checkpointer.go
mohammed-madi: https://github.com/couchbase/sync_gateway/pull/6098/commits/1a96d97ac227be538d451c808622f622e65ad4aa

db/active_replicator_checkpointer.go
```diff
@@ -59,17 +59,19 @@ type Checkpointer struct {
 
 	// closeWg waits for the time-based checkpointer goroutine to finish.
 	closeWg sync.WaitGroup
+	// For logging stats about the length of expected and processed sequence lists
+	dbStats *base.DbReplicatorStats
```
bbrks:
```suggestion
```

This isn't needed because we can directly access `c.stats.ExpectedSequenceLen`, etc. to update/read.

db/active_replicator_checkpointer.go
bbrks: Now we've done that change we don't need `c.dbStats` to access... we can continue to use `c.stats.ProcessedSequenceLen.Set(...)` directly

bbrks (changes requested):
The latest update looks good but there's still unnecessary code leftover from the preivous implementations.

db/active_replicator_checkpointer.go
mohammed-madi: https://github.com/couchbase/sync_gateway/pull/6098/commits/2eb80fbd8ae9af22fd8bb4eaf2e8a02c404034bf

db/active_replicator_checkpointer_test.go
```diff
@@ -43,6 +46,12 @@ func genProcessedForTest(t testing.TB, seqs ...string) map[SequenceID]struct{} {
 
 func TestCheckpointerSafeSeq(t *testing.T) {
 
+	stats, err := base.SyncGatewayStats.NewDBStats(t.Name(), false, false, false, nil, nil)
+	replicationStats, err := stats.DBReplicatorStats(t.Name())
+	require.NoError(t, err)
+	replicatorConfig := &ActiveReplicatorConfig{}
+	replicatorConfig.ReplicationStatsMap = replicationStats
+
```
bbrks: Initialising stats this way means each subtest shares the same stats, which might make future assertions tricky and adds a lot of identical test case boilerplate code.

Would it be better to move this code into the test function and set `c.stats` there?

bbrks (changes requested):
This is basically good to go aside from cleaning up `TestCheckpointerSafeSeq`

db/active_replicator_checkpointer_test.go
mohammed-madi: https://github.com/couchbase/sync_gateway/pull/6098/commits/8046b642151d0b4a0e7dfb2ddeb343935f782201

db/active_replicator_checkpointer_test.go
```diff
@@ -249,6 +252,15 @@ func TestCheckpointerSafeSeq(t *testing.T) {
...
 		t.Run(tt.name, func(t *testing.T) {
+			stats, err := base.SyncGatewayStats.NewDBStats(t.Name(), false, false, false, nil, nil)
+			replicationStats, err := stats.DBReplicatorStats(t.Name())
+			require.NoError(t, err)
+			replicatorConfig := &ActiveReplicatorConfig{}
+			replicatorConfig.ReplicationStatsMap = replicationStats
+			tt.c.stats.ProcessedSequenceLen = replicatorConfig.ReplicationStatsMap.ProcessedSequenceLen
+			tt.c.stats.ProcessedSequenceLenPostCleanup = replicatorConfig.ReplicationStatsMap.ProcessedSequenceLenPostCleanup
+			tt.c.stats.ExpectedSequenceLen = replicatorConfig.ReplicationStatsMap.ExpectedSequenceLen
+			tt.c.stats.ExpectedSequenceLenPostCleanup = replicatorConfig.ReplicationStatsMap.ExpectedSequenceLenPostCleanup
```
bbrks: Could be simplified a bit more to this, since we don't care about Prometheus or any other NewDBStats/DBReplicatorStats stuff in this particular test.

Would still allow us to assert on these stats using something like `tt.c.stats.ExpectedSequenceLen.Value()` inside the test

```suggestion
			tt.c.stats = CheckpointerStats{
				ExpectedSequenceLen:             &base.SgwIntStat{},
				ExpectedSequenceLenPostCleanup:  &base.SgwIntStat{},
				ProcessedSequenceLen:            &base.SgwIntStat{},
				ProcessedSequenceLenPostCleanup: &base.SgwIntStat{},
			}
```

db/active_replicator_checkpointer.go
```diff
@@ -295,9 +299,8 @@ func (c *Checkpointer) _updateCheckpointLists() (safeSeq *SequenceID) {
 		}
 	}
 
-	c.stats.ExpectedSequenceLenPostCleanup = len(c.expectedSeqs)
-	c.stats.ProcessedSequenceLenPostCleanup = len(c.processedSeqs)
-
+	c.stats.ExpectedSequenceLenPostCleanup.Set(int64(len(c.processedSeqs)))
```
bbrks: Typo here causing the wrong stat to be set.

```suggestion
	c.stats.ProcessedSequenceLenPostCleanup.Set(int64(len(c.processedSeqs)))
```

This suggests we don't have sufficient test coverage of the stats inside `TestActiveReplicatorPullSkippedSequence` to catch this error. I would recommend these assertions to cover them in the more comprehensive unit test:

```diff
diff --git a/db/active_replicator_checkpointer_test.go b/db/active_replicator_checkpointer_test.go
index 915c9076..e2cb9b82 100644
--- a/db/active_replicator_checkpointer_test.go
+++ b/db/active_replicator_checkpointer_test.go
@@ -267,6 +267,9 @@ func TestCheckpointerSafeSeq(t *testing.T) {
 			})

 			t.Run("_updateCheckpointLists", func(t *testing.T) {
+				originalExpectedSeqsLen := len(tt.c.expectedSeqs)
+				originalProcessedSeqsLen := len(tt.c.processedSeqs)
+
 				actualSafeSeq := tt.c._updateCheckpointLists()
 				if tt.expectedSafeSeq == nil {
 					assert.Nil(t, actualSafeSeq)
@@ -275,6 +278,10 @@ func TestCheckpointerSafeSeq(t *testing.T) {
 				}
 				assert.Equal(t, tt.expectedExpectedSeqs, tt.c.expectedSeqs)
 				assert.Equal(t, tt.expectedProcessedSeqs, tt.c.processedSeqs)
+				assert.Equal(t, int64(originalExpectedSeqsLen), tt.c.stats.ExpectedSequenceLen.Value())
+				assert.Equal(t, int64(len(tt.expectedExpectedSeqs)), tt.c.stats.ExpectedSequenceLenPostCleanup.Value())
+				assert.Equal(t, int64(originalProcessedSeqsLen), tt.c.stats.ProcessedSequenceLen.Value())
+				assert.Equal(t, int64(len(tt.expectedProcessedSeqs)), tt.c.stats.ProcessedSequenceLenPostCleanup.Value())

 				// _updateCheckpointLists should be idempotent
 				// we'd expect no safe seq, and no further changes to either list, as long as c.processedSeqs hasn't been added to
@@ -282,6 +289,8 @@ func TestCheckpointerSafeSeq(t *testing.T) {
 				assert.Nil(t, actualSafeSeq2)
 				assert.Equal(t, tt.expectedExpectedSeqs, tt.c.expectedSeqs)
 				assert.Equal(t, tt.expectedProcessedSeqs, tt.c.processedSeqs)
+				assert.Equal(t, tt.c.stats.ExpectedSequenceLenPostCleanup.Value(), tt.c.stats.ExpectedSequenceLen.Value())
+				assert.Equal(t, tt.c.stats.ProcessedSequenceLenPostCleanup.Value(), tt.c.stats.ProcessedSequenceLen.Value())
 			})
 		})
 	}
```

bbrks (changes requested):
Incorrect stat being set (see below for correction and extra unit test coverage)

bbrks (approved):
LGTM 👍

PR #6100, opened 2023-02-17
Add HashStringKVMap util for generating a hash based on the contents of a map

bbrks:
Required for upcoming ISGR collections work, which requires identifying when the set of collections (or their mappings) change.
We use this generated hash in the replication config hash inside the checkpoint to detect changes and reset the replicator since value.
 
Tried to figure out if I could reduce number of map iterations and I'm not sure I can. If anybody has any brighter ideas I'm open to it.

```go
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkHashStringKVMap
BenchmarkHashStringKVMap-12    	 1496924	       797.8 ns/op	     192 B/op	       5 allocs/op
PASS
```

### DIFF
--- a/base/util.go
+++ b/base/util.go
@@ -1784,3 +1784,31 @@ func WaitForNoError(callback func() error) error {
 	}, CreateMaxDoublingSleeperFunc(30, 10, 1000))
 	return err
 }
+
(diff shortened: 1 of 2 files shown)


bbrks:
not needed - going in wrong direction

PR #6101, opened 2023-02-21
CBG-2343 Add database registry and associated API

adamcfraser:
Adds a new database registry document to the bucket that lists the set of databases defined for the bucket (across all config groups), and their associated collections.  The registry is used by persistent config polling to find all configs for a given bucket.

The registry is also used for cross-database config validation.  Currently the only validation is to ensure collections aren’t assigned to multiple databases (in separate config groups).

To synchronize access to the registry and ensure values are kept in step with the associated config files, changes were required to how the bootstrap manager interacts with config files.

The low-level config file operations in BootstrapConnection were made more generic so that they could be used for modifying both config and registry files. A new API (ConfigManager), implemented by bootstrapContext in REST, should be used for all mutations to config files.  It includes handling for failure cases (including but not limited to node failure between registry and database config updates).

CBG-2343

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1539/

### DIFF
--- a/rest/adminapitest/admin_api_test.go
+++ b/rest/adminapitest/admin_api_test.go
@@ -2823,7 +2823,7 @@ func TestPersistentConfigConcurrency(t *testing.T) {
 		t.Skip("This test only works against Couchbase Server")
 	}
 
-	base.SetUpTestLogging(t, base.LevelInfo, base.KeyHTTP)
+	base.SetUpTestLogging(t, base.LevelInfo, base.KeyHTTP, base.KeyDCP)
 
 	serverErr := make(chan error, 0)
 
@@ -3563,7 +3563,7 @@ func TestDbConfigPersistentSGVersions(t *testing.T) {
 	require.NoError(t, err)
 
 	// initialise with db config
-	_, err = sc.BootstrapContext.Connection.InsertConfig(tb.GetName(), t.Name(), dbConfig)
+	_, err = sc.BootstrapContext.InsertConfig(ctx, tb.GetName(), t.Name(), &dbConfig)
 	require.NoError(t, err)
 
 	assertRevsLimit := func(sc *rest.ServerContext, revsLimit uint32) {
@@ -3584,18 +3584,15 @@ func TestDbConfigPersistentSGVersions(t *testing.T) {
 	assertRevsLimit(sc, 123)
 
 	writeRevsLimitConfigWithVersion := func(sc *rest.ServerContext, version string, revsLimit uint32) error {
-		_, err = sc.BootstrapContext.Connection.UpdateConfig(tb.GetName(), t.Name(), func(rawBucketConfig []byte, rawBucketConfigCas uint64) (updatedConfig []byte, err error) {
-			var db rest.DatabaseConfig
-			if err := base.JSONUnmarshal(rawBucketConfig, &db); err != nil {
-				return nil, err
-			}
+		_, err = sc.BootstrapContext.UpdateConfig(base.TestCtx(t), tb.GetName(), "db", t.Name(), func(db *rest.DatabaseConfig) (updatedConfig *rest.DatabaseConfig, err error) {
+
 			db.SGVersion = version
 			db.DbConfig.RevsLimit = base.Uint32Ptr(revsLimit)
 			db.Version, err = rest.GenerateDatabaseConfigVersionID(db.Version, &db.DbConfig)
 			if err != nil {
 				return nil, err
 			}
-			return base.JSONMarshal(db)
+			return db, nil
 		})
 		return err
 	}
--- a/rest/adminapitest/admin_api_test.go
+++ b/rest/adminapitest/admin_api_test.go
@@ -3546,13 +3546,14 @@ func TestDbConfigPersistentSGVersions(t *testing.T) {
 	tb := base.GetTestBucket(t)
 	defer func() { tb.Close() }()
 
+	dbName := "db"
 	dbConfig := rest.DatabaseConfig{
 		SGVersion: "", // leave empty to emulate what 3.0.0 would've written to the bucket
 		DbConfig: rest.DbConfig{
 			BucketConfig: rest.BucketConfig{
 				Bucket: base.StringPtr(tb.GetName()),
 			},
-			Name:             "db",
+			Name:             dbName,
 			EnableXattrs:     base.BoolPtr(base.TestUseXattrs()),
 			UseViews:         base.BoolPtr(base.TestsDisableGSI()),
 			NumIndexReplicas: base.UintPtr(0),
@@ -3568,7 +3569,7 @@ func TestDbConfigPersistentSGVersions(t *testing.T) {
 
 	assertRevsLimit := func(sc *rest.ServerContext, revsLimit uint32) {
 		rest.WaitAndAssertCondition(t, func() bool {
-			dbc, err := sc.GetDatabase(ctx, "db")
+			dbc, err := sc.GetDatabase(ctx, dbName)
 			if err != nil {
 				t.Logf("expected database with RevsLimit=%v but got err=%v", revsLimit, err)
 				return false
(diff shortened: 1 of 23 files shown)


rest/adminapitest/admin_api_test.go
```diff
@@ -3584,18 +3584,15 @@ func TestDbConfigPersistentSGVersions(t *testing.T) {
 	assertRevsLimit(sc, 123)
 
 	writeRevsLimitConfigWithVersion := func(sc *rest.ServerContext, version string, revsLimit uint32) error {
-		_, err = sc.BootstrapContext.Connection.UpdateConfig(tb.GetName(), t.Name(), func(rawBucketConfig []byte, rawBucketConfigCas uint64) (updatedConfig []byte, err error) {
-			var db rest.DatabaseConfig
-			if err := base.JSONUnmarshal(rawBucketConfig, &db); err != nil {
-				return nil, err
-			}
+		_, err = sc.BootstrapContext.UpdateConfig(base.TestCtx(t), tb.GetName(), "db", t.Name(), func(db *rest.DatabaseConfig) (updatedConfig *rest.DatabaseConfig, err error) {
```
torcolvin: can you put `dbConfig.Name` here? if we want to change this name for readability later?

rest/config.go
```diff
@@ -1507,17 +1514,12 @@ func (sc *ServerContext) FetchConfigs(ctx context.Context, isInitialStartup bool
...
 	for _, bucket := range buckets {
-		base.TracefCtx(ctx, base.KeyConfig, "Checking for config for group %q from bucket %q", sc.Config.Bootstrap.ConfigGroupID, bucket)
-		var cnf DatabaseConfig
-		cas, err := sc.BootstrapContext.Connection.GetConfig(bucket, sc.Config.Bootstrap.ConfigGroupID, &cnf)
-		if err == base.ErrNotFound {
-			base.DebugfCtx(ctx, base.KeyConfig, "Bucket %q did not contain config for group %q", bucket, sc.Config.Bootstrap.ConfigGroupID)
-			continue
-		}
+
+		base.TracefCtx(ctx, base.KeyConfig, "Checking for configs for group %q from bucket %q", sc.Config.Bootstrap.ConfigGroupID, bucket)
```
torcolvin, line 1549: I'd do:

`ctx := base.AddBucketContext(ctx, bucketName)` and add logging for buckets like we do for test buckets.

Might be out of scope of this ticket, but it's easier to put this information in logging ctx, IMO.

rest/config.go
```diff
@@ -1318,6 +1318,10 @@ func (sc *ServerContext) fetchAndLoadConfigs(ctx context.Context, isInitialStart
 		}
 	}
 
+	base.InfofCtx(ctx, base.KeyAll, "calling applyConfigs")
+	defer func() {
+		base.InfofCtx(ctx, base.KeyAll, "finished calling applyConfigs")
+	}()
```
bbrks: Is this Debug/Trace logging?

rest/config.go
```diff
@@ -1351,6 +1355,39 @@ func (sc *ServerContext) _fetchAndLoadDatabase(nonContextStruct base.NonCancella
...
+	for _, bucketName := range buckets {
+		var dbConfig DatabaseConfig
+		legacyCas, getErr := sc.BootstrapContext.Connection.GetMetadataDocument(bucketName, PersistentConfigKey30(groupID), &dbConfig)
+		if getErr == base.ErrNotFound {
+			continue
+		} else if getErr != nil {
+			return fmt.Errorf("Error retrieving 3.0 config for bucket: %s, groupID: %s: %w", bucketName, groupID, err)
+		}
+
+		_, insertErr := sc.BootstrapContext.InsertConfig(ctx, bucketName, groupID, &dbConfig)
```
bbrks, line 1372: Should we log at Info before/when doing this?

rest/config_manager.go
```diff
@@ -0,0 +1,639 @@
...
+// GetConfig does *not* validate that config version matches registry version - operations requiring synchronization
+// with registry should use getRegistryAndDatabase, or getConfig with the required version
+func (b *bootstrapContext) GetConfig(bucketName, groupID, dbName string, config *DatabaseConfig) (cas uint64, err error) {
+	return b.Connection.GetMetadataDocument(bucketName, PersistentConfigKey(groupID, dbName), config)
+}
+
+// InsertConfig saves a new database config for a given bucket and config group ID. This is a three-step process:
+//  1. getRegistryAndDatabase to enforce synchronization pre-update
+//  2. Update the registry to add the config
+//  3. Write the config
```
bbrks, line 57: Given the multi-step/complexity in these functions I'm wondering whether it's worth logging at each step for future diagnosis, or at least in the outer retry loop.

rest/config_registry.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+	collectionConflicts := r.getCollectionConflicts(ctx, config.Name, config.Scopes)
+	if len(collectionConflicts) > 0 {
+		return nil, base.HTTPErrorf(http.StatusConflict, fmt.Sprintf("Cannot update config for database %s - collections are in use by another database: %v", base.UD(config.Name), collectionConflicts))
+	}
+
+	// For conflicts with in-flight updates, call getRegistryAndDatabase to block until those updates complete or rollback
+	previousVersionConflicts = r.getPreviousConflicts(ctx, config.Name, config.Scopes)
+	if len(previousVersionConflicts) > 0 {
+		return previousVersionConflicts, base.HTTPErrorf(http.StatusConflict, fmt.Sprintf("Cannot update config, collections are in use by another database with an update in progress"))
```
bbrks: nit:

```suggestion
		return previousVersionConflicts, base.HTTPErrorf(http.StatusConflict, "Cannot update config, collections are in use by another database with an update in progress")
```

rest/config_registry.go
```diff
@@ -0,0 +1,349 @@
...
+// If in-conflict with an in-flight update to other database(s), returns that set as previousVersionConflicts so that callers
+// can wait and retry.
+func (r *GatewayRegistry) upsertDatabaseConfig(ctx context.Context, configGroupID string, config *DatabaseConfig) (previousVersionConflicts []configGroupAndDatabase, err error) {
+
+	if config == nil {
+		return nil, fmt.Errorf("attempted to upsertDatabaseConfig with nil config")
+	}
+	collectionConflicts := r.getCollectionConflicts(ctx, config.Name, config.Scopes)
+	if len(collectionConflicts) > 0 {
+		return nil, base.HTTPErrorf(http.StatusConflict, fmt.Sprintf("Cannot update config for database %s - collections are in use by another database: %v", base.UD(config.Name), collectionConflicts))
```
bbrks: nit:

```suggestion
		return nil, base.HTTPErrorf(http.StatusConflict, "Cannot update config for database %s - collections are in use by another database: %v", base.UD(config.Name), collectionConflicts)
```

rest/rest_tester_cluster_test.go
```diff
@@ -203,7 +203,7 @@ func TestPersistentDbConfigWithInvalidUpsert(t *testing.T) {
 	assert.NotContains(t, string(resp.BodyBytes()), `"revs_limit":`)
 
 	// remove the db config directly from the bucket
-	docID, err := base.PersistentConfigKey(*rtc.config.groupID)
+	docID := PersistentConfigKey(*rtc.config.groupID, db)
 	require.NoError(t, err)
```
bbrks:
```suggestion
```

bbrks (changes requested):
I think we could use a bit more logging around the complicated parts (see inline) to aid in any future analysis but otherwise looks good.

rest/config.go
adamcfraser: Dev-time logging, I've removed.

rest/config_manager.go
adamcfraser: Agree - added this logging.

rest/config.go
adamcfraser: Added logging here.

adamcfraser:
Addressed PR feedback and rebased on master.  Kicked off new integration test run:
https://jenkins.sgwdev.com/job/SyncGateway-Integration/1539/

bbrks (approved):

PR #6102, opened 2023-02-21
enable ineffassign

torcolvin:
Add ineffassign to required linters. Nearly all changes are test changes except:

- parseSequenceInt now returns empty structs on error
- TimedSetFromString returns error if vbucket is not an int
- one graphql rest change to handle error
- handle error case in rest tester if we can not get a user

The other non test changes are just removing unused variables.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1567/


auth/oidc.go
```diff
@@ -325,12 +325,12 @@ func (op *OIDCProvider) initOIDCClient(ctx context.Context) error {
 		return err
 	}
 
-	metadata, verifier, err := op.DiscoverConfig(ctx)
+	metadata, _, err := op.DiscoverConfig(ctx)
```
adamcfraser, line 328: It doesn't look like this leaves any consumers for the verifier returned by DiscoverConfig.  I'm not convinced this was an intentional change.  I realize this PR doesn't change that, but I feel this might merit a tracking ticket to evaluate whether we should be using the verifier returned by DiscoverConfig in the case where we're using standard discovery.

torcolvin: https://issues.couchbase.com/browse/CBG-2752

adamcfraser (approved):

PR #6103, opened 2023-02-21
CBG-2595 use metadata specific indexes

torcolvin:
I added `rest/integrationtest` that doesn't initialize any indexes so that we can catch bugs like we saw before where we weren't creating indexes for metadata storage with a non default collection. This test could probably be expanded to test other scenarios but I wasn't sure which were good ideas.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1518/

### DIFF
--- a/base/logging_context.go
+++ b/base/logging_context.go
@@ -30,13 +30,13 @@ type LogContext struct {
 	TestName string
 
 	// TestBucketName is the name of a bucket used during a test
-	TestBucketName string
+	BucketName string
 
 	// TestScopeName is the name of a scope on used during a test
-	TestScopeName string
+	ScopeName string
 
 	// TestCollectionName is the name of the collection used during a test
-	TestCollectionName string
+	CollectionName string
 }
 
 // addContext returns a string format with additional log context if present.
@@ -49,16 +49,19 @@ func (lc *LogContext) addContext(format string) string {
 		format = "c:" + lc.CorrelationID + " " + format
 	}
 
-	if lc.TestBucketName != "" {
-		keyspace := "b:" + lc.TestBucketName
-		if lc.TestScopeName != "" {
-			keyspace += "." + lc.TestScopeName
+	if lc.BucketName != "" {
+		keyspace := "b:" + lc.BucketName
+		if lc.ScopeName != "" {
+			keyspace += "." + lc.ScopeName
 		}
-		if lc.TestCollectionName != "" {
-			keyspace += "." + lc.TestCollectionName
+		if lc.CollectionName != "" {
+			keyspace += "." + lc.CollectionName
 		}
 
 		format = keyspace + " " + format
+	} else if lc.CollectionName != "" {
+		format = "col:" + lc.CollectionName + " " + format
+
 	}
 
 	if lc.TestName != "" {
@@ -96,16 +99,25 @@ func bucketCtx(parent context.Context, b Bucket) context.Context {
 func bucketNameCtx(parent context.Context, bucketName string) context.Context {
 	parentLogCtx, _ := parent.Value(requestContextKey).(LogContext)
 	newCtx := LogContext{
-		TestName:       parentLogCtx.TestName,
-		TestBucketName: bucketName,
+		TestName:   parentLogCtx.TestName,
+		BucketName: bucketName,
+	}
+	return LogContextWith(parent, &newCtx)
+}
+
+// BucketCollectionCtx extends the parent context with a collection name.
+func BucketAndCollectionCtx(parent context.Context, b Bucket, collectionName string) context.Context {
+	newCtx := LogContext{
+		BucketName:     b.GetName(),
+		CollectionName: collectionName,
 	}
 	return LogContextWith(parent, &newCtx)
 }
 
 // CollectionCtx extends the parent context with a collection name.
 func CollectionCtx(parent context.Context, collectionName string) context.Context {
-	newCtx := CollectionLogContext{
-		Collection: collectionName,
+	newCtx := LogContext{
+		CollectionName: collectionName,
 	}
 	return LogContextWith(parent, &newCtx)
 }
@@ -114,10 +126,10 @@ func CollectionCtx(parent context.Context, collectionName string) context.Contex
 func testKeyspaceNameCtx(parent context.Context, bucketName, scopeName, collectionName string) context.Context {
 	parentLogCtx, _ := parent.Value(requestContextKey).(LogContext)
 	newCtx := LogContext{
-		TestName:           parentLogCtx.TestName,
-		TestBucketName:     bucketName,
-		TestScopeName:      scopeName,
-		TestCollectionName: collectionName,
+		TestName:       parentLogCtx.TestName,
+		BucketName:     bucketName,
+		ScopeName:      scopeName,
+		CollectionName: collectionName,
 	}
 	return LogContextWith(parent, &newCtx)
 }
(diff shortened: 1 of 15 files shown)


base/util_testing.go
```diff
@@ -130,7 +130,9 @@ func getTestBucket(t testing.TB, persistent bool) *TestBucket {
 func (tb *TestBucket) GetNamedDataStore(count int) (DataStore, error) {
 	dataStoreNames := tb.GetNonDefaultDatastoreNames()
 	if count > len(dataStoreNames) {
-		tb.t.Errorf("You are requesting more datastores %d than are available on this test instance %d", dataStoreNames, count)
+		tb.t.Fatalf("You are requesting more datastores %d than are available on this test instance %d", dataStoreNames, count)
+	} else if len(dataStoreNames) == 0 {
```
bbrks: Do you want this check first, otherwise any non-zero `count` would always be `> 0` and only give first error

rest/integrationtest/integration_test.go
```diff
@@ -0,0 +1,83 @@
...
+			response := rt.SendAdminRequest(http.MethodPut, "/{{.db}}/_user/"+user, rest.GetUserPayload(t, user, rest.RestTesterDefaultUserPassword, "", rt.GetSingleTestDatabaseCollection(), []string{"ChannelA"}, nil))
+			rest.RequireStatus(t, response, http.StatusCreated)
+		}
+		response := rt.SendAdminRequest(http.MethodGet, "/{{.db}}/_user/", "")
+		rest.RequireStatus(t, response, http.StatusOK)
+
+		var responseUsers []string
+		err = json.Unmarshal(response.Body.Bytes(), &responseUsers)
+		require.NoError(t, err)
+		require.Equal(t, users, responseUsers)
```
bbrks: Is there anything in these tests that we wouldn't expect to work in views (with the exception of `requireNoIndexes`)

Anything index/GSI-specific we need to assert on?

rest/integrationtest/main_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+package integrationtest
+
+import (
+	"testing"
+
+	"github.com/couchbase/sync_gateway/base"
+)
+
+func TestMain(m *testing.M) {
```
bbrks: If this package is mainly to to avoid upfront creation of indexes, perhaps copy `db/indexestest`'s naming convention

Otherwise almost all of our tests would fit the `integrationtest` criteria.

base/util_testing.go
```diff
@@ -130,7 +130,9 @@ func getTestBucket(t testing.TB, persistent bool) *TestBucket {
 func (tb *TestBucket) GetNamedDataStore(count int) (DataStore, error) {
 	dataStoreNames := tb.GetNonDefaultDatastoreNames()
 	if count > len(dataStoreNames) {
-		tb.t.Errorf("You are requesting more datastores %d than are available on this test instance %d", dataStoreNames, count)
+		tb.t.Fatalf("You are requesting more datastores %d than are available on this test instance %d", dataStoreNames, count)
```
bbrks: Parameters are wrong way around (and missing `len()` for `%d`)

```suggestion
		tb.t.Fatalf("You are requesting more datastores %d than are available on this test instance %d", count, len(dataStoreNames))
```

db/indexes.go
```diff
@@ -331,22 +344,41 @@ func (i *SGIndex) createIfNeeded(bucket base.N1QLStore, useXattrs bool, numRepli
...
+	Serverless      bool                  // if true, create indexes for serverless
+	UseXattrs       bool                  // if true, create indexes on xattrs, otherwise, use inline sync data
+}
+
 // Initializes Sync Gateway indexes for bucket.  Creates required indexes if not found, then waits for index readiness.
-func InitializeIndexes(n1QLStore base.N1QLStore, useXattrs bool, numReplicas uint, failFast bool, isServerless bool) error {
+func InitializeIndexes(ctx context.Context, n1QLStore base.N1QLStore, options InitializeIndexOptions) error {
 
-	base.InfofCtx(context.TODO(), base.KeyAll, "Initializing indexes with numReplicas: %d...", numReplicas)
+	base.InfofCtx(ctx, base.KeyAll, "Initializing indexes for with numReplicas: %d...", options.NumReplicas)
```
bbrks: missing datastore name or accidental `for `?

rest/server_context.go
```diff
@@ -514,15 +523,24 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
...
+					hasDefaultCollection = true
+					metadataIndexOption = db.IndexesAll
+				}
+				if err := initDataStore(dataStore, metadataIndexOption); err != nil {
 					return nil, err
 				}
 			}
 		}
+		if !hasDefaultCollection {
+			if err := initDataStore(bucket.DefaultDataStore(), db.IndexesMetadataOnly); err != nil {
```
bbrks, line 537: As mentioned in Slack - don't think this is fully working. The metadata-only default collection gets a full set of indexes.

bbrks (changes requested):

rest/integrationtest/integration_test.go
torcolvin: Works on walrus/views, but I added some actual index asserts to catch bug you pointed out.

base/logging_context.go
```diff
@@ -49,16 +49,19 @@ func (lc *LogContext) addContext(format string) string {
...
-			keyspace += "." + lc.TestCollectionName
+		if lc.CollectionName != "" {
+			keyspace += "." + lc.CollectionName
 		}
 
 		format = keyspace + " " + format
+	} else if lc.CollectionName != "" {
+		format = "col:" + lc.CollectionName + " " + format
+
 	}
```
bbrks: I don't want to hold up the functional changes we're aiming to get completed by Friday, so I think we should back these changes out for now and move forwards with the index changes in this PR.

It feels unrelated/scope creep to this PR and has issues with the current implementation. Doesn't seem to be consistently applied, resulting in even related logs looking like this:

```
2023-02-22T14:40:59.796Z [INF] col:b1.foo.bar Verifying index availability...
2023-02-22T14:40:59.814Z [INF] b:b1.bar Indexes not ready - retrying...
2023-02-22T14:40:59.919Z [INF] b:b1.bar Indexes not ready - retrying...
2023-02-22T14:41:00.124Z [INF] b:b1.bar Indexes not ready - retrying...
2023-02-22T14:41:00.527Z [INF] b:b1.bar Indexes not ready - retrying...
2023-02-22T14:41:01.333Z [INF] b:b1.bar Indexes not ready - retrying...
...
2023-02-22T14:41:02.943Z [INF] b:b1.bar Index sg_access_x1 is online
2023-02-22T14:41:02.943Z [INF] b:b1.bar Index sg_allDocs_x1 is online
2023-02-22T14:41:02.943Z [INF] b:b1.bar Index sg_channels_x1 is online
2023-02-22T14:41:02.943Z [INF] b:b1.bar Index sg_roleAccess_x1 is online
2023-02-22T14:41:02.943Z [INF] b:b1.bar Index sg_tombstones_x1 is online
2023-02-22T14:41:02.943Z [INF] col:b1.foo.bar Indexes ready
2023-02-22T14:41:02.943Z [INF] col:b1 Initializing indexes with numReplicas: 0...
```

Open questions around context logging which should end up being answered in a ticket before we actually make the changes:
- Is bucket (and possibly scope) already implied by `db` context logging?
- Should `col:` be fully qualified, or compact assuming `db` is present and above is true. We've hashed this out a few times in Slack and always settled on non-qualified, it seems we're changing it in this PR to now be fully qualified?

bbrks (changes requested):
Can we roll back the unrelated changes to context logging in this PR and revisit separately?

I don't want to hold up this PR as it's required for RC1, and looks basically good to merge, but there's further work to do on the context logging (see inline comment)

base/logging_context.go
torcolvin: I agree these are good questions, and should probably we elucidated. There are exceptional cases that we are finding.

In this case, we are creating indexes before we have a db, so we don't know the bucket, except via the database, implicitly. In this case the database hasn't been created.

For the DBRegistry, we are going to see the same problems of not having a bucket. Additionally, the logging is actually different for test code and not test code. I don't think we generally see the `b:` logging outside of tests, except in the situation I just described.

bbrks (approved):

PR #6105, opened 2023-02-21
CBG-2679 use 3.0 metadata for in memory implementations as well

torcolvin:
Fixes https://jenkins.sgwdev.com/job/MasterIntegration/387/testReport/junit/github/com_couchbase_sync_gateway_base/TestCBGTIndexCreationSafeLegacyName/

I could reproduce with `-count 100`

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1528/

### DIFF
--- a/base/cbgt.go
+++ b/base/cbgt.go
@@ -1,0 +1,21 @@
+package base
+
+import (
+	"fmt"
+
+	"github.com/couchbase/cbgt"
+)
+
+// NewCbgtCfgMem runs cbgt.NewCfgMem and sets the matching version number we expect for Sync Gateway.
+func NewCbgtCfgMem() (*cbgt.CfgMem, error) {
+	cfg := cbgt.NewCfgMem()
+	cas, err := cfg.Set(cbgt.VERSION_KEY, []byte(SGCbgtMetadataVersion), 0)
+	if err != nil {
+		return nil, err
+	}
+	expectedCas := uint64(1)
+	if cas != uint64(1) {
+		return nil, fmt.Errorf("Expected cas value %d, got: %d", expectedCas, cas)
+	}
+	return cfg, nil
+}
--- a/base/cbgt.go
+++ b/base/cbgt.go
@@ -1,3 +1,11 @@
+// Copyright 2023-Present Couchbase, Inc.
+//
+// Use of this software is governed by the Business Source License included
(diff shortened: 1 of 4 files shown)


base/cbgt.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+// NewCbgtCfgMem runs cbgt.NewCfgMem and sets the matching version number we expect for Sync Gateway.
+func NewCbgtCfgMem() (*cbgt.CfgMem, error) {
+	cfg := cbgt.NewCfgMem()
+	cas, err := cfg.Set(cbgt.VERSION_KEY, []byte(SGCbgtMetadataVersion), 0)
+	if err != nil {
+		return nil, err
+	}
+	expectedCas := uint64(1)
```
adamcfraser: Why do we care about the CAS value CfgMem assigns to the VERSION_KEY entry?

base/cbgt.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"fmt"
+
+	"github.com/couchbase/cbgt"
+)
+
+// NewCbgtCfgMem runs cbgt.NewCfgMem and sets the matching version number we expect for Sync Gateway.
+func NewCbgtCfgMem() (*cbgt.CfgMem, error) {
+	cfg := cbgt.NewCfgMem()
+	cas, err := cfg.Set(cbgt.VERSION_KEY, []byte(SGCbgtMetadataVersion), 0)
```
adamcfraser: I would expect cbgt to be writing the version key, in the same way it does when running against a document-backed store.  Why do we need to define it in advance in the Cfg key value store when using CfgMem?

adamcfraser (changes requested):

base/cbgt.go
torcolvin: We don't, I can remove this check.

base/cbgt.go
torcolvin: The spec of the implmentation is such that it reads the key out of the config if it's not set, so this matches our not in memory version https://github.com/couchbase/sync_gateway/blob/master/base/sg_cluster_cfg.go#L69

adamcfraser: Makes sense - thanks.

adamcfraser (approved):

PR #6109, opened 2023-02-23
Prerequisite refactoring for upcoming ISGR PR

bbrks:
- Some cleanup/refactoring to minimise upcoming ISGR PR size
- Broken down each separate refactor/cleanup by standalone commit
- There should be no functional changes present
- Added some new test/utility functions

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1534/
  - `TestRemoveObsoleteIndexOnError` failed (known issue)

### DIFF
--- a/base/collection_common.go
+++ b/base/collection_common.go
@@ -32,3 +32,19 @@ func (s ScopeAndCollectionName) ScopeName() string {
 func (s ScopeAndCollectionName) CollectionName() string {
 	return s.Collection
 }
+
+// String returns a dot-separated formatted scope and collection name.
+func (s ScopeAndCollectionName) String() string {
+	return s.Scope + ScopeCollectionSeparator + s.Collection
+}
+
+type ScopeAndCollectionNames []ScopeAndCollectionName
+
+// ScopeAndCollectionNames returns a dot-separated formatted slice of scope and collection names.
+func (s ScopeAndCollectionNames) ScopeAndCollectionNames() []string {
+	scopes := make([]string, 0, len(s))
+	for _, scopeAndCollection := range s {
+		scopes = append(scopes, scopeAndCollection.String())
+	}
+	return scopes
+}
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -551,7 +551,7 @@ func (bh *blipHandler) sendBatchOfChanges(sender *blip.Sender, changeArray [][]i
 		bh.replicationStats.SendChangesCount.Add(int64(len(changeArray)))
 		// Spawn a goroutine to await the client's response:
 		go func(bh *blipHandler, sender *blip.Sender, response *blip.Message, changeArray [][]interface{}, sendTime time.Time, dbCollection *DatabaseCollectionWithUser) {
-			if err := bh.handleChangesResponse(sender, response, changeArray, sendTime, dbCollection); err != nil {
+			if err := bh.handleChangesResponse(sender, response, changeArray, sendTime, dbCollection, bh.collectionIdx); err != nil {
 				base.WarnfCtx(bh.loggingCtx, "Error from bh.handleChangesResponse: %v", err)
 				if bh.fatalErrorCallback != nil {
 					bh.fatalErrorCallback(err)
@@ -794,12 +794,7 @@ func (bh *blipHandler) handleProposeChanges(rq *blip.Message) error {
 
 // ////// DOCUMENTS:
 
-func (bsc *BlipSyncContext) sendRevAsDelta(sender *blip.Sender, docID, revID string, deltaSrcRevID string, seq SequenceID, knownRevs map[string]bool, maxHistory int, handleChangesResponseCollection *DatabaseCollectionWithUser) error {
-	var collectionIdx *int
-	if coll, ok := bsc.collections.getIndexForDB(handleChangesResponseCollection); ok {
-		collectionIdx = &coll
-	}
-
+func (bsc *BlipSyncContext) sendRevAsDelta(sender *blip.Sender, docID, revID string, deltaSrcRevID string, seq SequenceID, knownRevs map[string]bool, maxHistory int, handleChangesResponseCollection *DatabaseCollectionWithUser, collectionIdx *int) error {
 	bsc.replicationStats.SendRevDeltaRequestedCount.Add(1)
 
 	revDelta, redactedRev, err := handleChangesResponseCollection.GetDelta(bsc.loggingCtx, docID, deltaSrcRevID, revID)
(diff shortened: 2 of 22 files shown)


base/collection_common.go
```diff
@@ -32,3 +32,19 @@ func (s ScopeAndCollectionName) ScopeName() string {
...
+
+// String returns a dot-separated formatted scope and collection name.
+func (s ScopeAndCollectionName) String() string {
+	return s.Scope + ScopeCollectionSeparator + s.Collection
+}
+
+type ScopeAndCollectionNames []ScopeAndCollectionName
+
+// ScopeAndCollectionNames returns a dot-separated formatted slice of scope and collection names.
+func (s ScopeAndCollectionNames) ScopeAndCollectionNames() []string {
```
torcolvin, line 44: This would be useful https://github.com/couchbase/sync_gateway/blob/master/rest/utilities_testing.go#L2320

db/blip_handler.go
```diff
@@ -551,7 +551,7 @@ func (bh *blipHandler) sendBatchOfChanges(sender *blip.Sender, changeArray [][]i
 		bh.replicationStats.SendChangesCount.Add(int64(len(changeArray)))
 		// Spawn a goroutine to await the client's response:
 		go func(bh *blipHandler, sender *blip.Sender, response *blip.Message, changeArray [][]interface{}, sendTime time.Time, dbCollection *DatabaseCollectionWithUser) {
-			if err := bh.handleChangesResponse(sender, response, changeArray, sendTime, dbCollection); err != nil {
+			if err := bh.handleChangesResponse(sender, response, changeArray, sendTime, dbCollection, bh.collectionIdx); err != nil {
```
torcolvin, line 554: It makes me wary, but I'm sure that there's a good reason to pass the collection index here.

My understanding was that we wanted a copy of the state of the ~database~ collection before we start replication, so at some point (not in this review?) we're going to be able to access this again. I might have done this anyway, which is why you removed `getIndexForDB`. In that case, I wonder if it makes sense to pass two collections?

I don't object to this change, just trying to understand.

db/blip_sync_context.go
```diff
@@ -576,12 +576,7 @@ func (bsc *BlipSyncContext) sendNoRev(sender *blip.Sender, docID, revID string,
 }
 
 // Pushes a revision body to the client
-func (bsc *BlipSyncContext) sendRevision(sender *blip.Sender, docID, revID string, seq SequenceID, knownRevs map[string]bool, maxHistory int, handleChangesResponseCollection *DatabaseCollectionWithUser) error {
-	var collectionIdx *int
-	if coll, ok := bsc.collections.getIndexForDB(handleChangesResponseCollection); ok {
-		collectionIdx = &coll
-	}
-
+func (bsc *BlipSyncContext) sendRevision(sender *blip.Sender, docID, revID string, seq SequenceID, knownRevs map[string]bool, maxHistory int, handleChangesResponseCollection *DatabaseCollectionWithUser, collectionIdx *int) error {
```
torcolvin, line 579: I know I wrote this code but I don't understand why we need a new collection here - presumably it's if the user was updated? I wonder if it would be more clear to have a function that calls reloadUser on `DatabaseCollectionWithUser` or just reconstruct it with:

```
collection := &DatabaseCollectionWithUser{DatabaseContext: handleChangesResponseCollection.DatabaseContext, user: bh.user}
```

I think you'd have to lock around user to do this.

db/special_docs.go
```diff
@@ -32,6 +32,10 @@ func (db *DatabaseCollection) GetSpecial(doctype string, docid string) (Body, er
 }
 
 func (db *DatabaseCollection) GetSpecialBytes(doctype string, docid string) ([]byte, error) {
```
torcolvin: at this point, i'd change the receiver type to `c` becuase it's not a db operation

db/special_docs.go
```diff
@@ -53,6 +57,10 @@ func (db *DatabaseCollection) GetSpecialBytes(doctype string, docid string) ([]b
 
 // Updates or deletes a special document.
 func (db *DatabaseCollection) putSpecial(doctype string, docid string, matchRev string, body Body) (string, error) {
```
torcolvin: change receiver name to c because this is not a database

rest/utilities_testing.go
```diff
@@ -2306,6 +2306,18 @@ func (rt *RestTester) GetKeyspaces() []string {
 	return keyspaces
 }
 
+func (rt *RestTester) GetDbCollections() []*db.DatabaseCollection {
```
torcolvin, line 2310: This is worth a comment because I get asked this, it's return all collections in lexicographic order for ease of test usage and it'll match `GetKeyspaces()` or `getCollectionsForBLIP`

torcolvin (commented):
Nothing on this review is a dealbreaker at all, and I'm happy for you to avoid any comments that you disagree with or are onerous to implement.

I didn't spot any bugs, so I'd approve it on sight.

db/blip_sync_context.go
bbrks: I have no idea, but would rather not touch it if I don't have to, to be completely honest...

db/blip_handler.go
bbrks: Passing collectionIdx directly avoids having to iterate over the set of collections to find the index matching the collection name via `getIndexForDB` - which seemed way too inefficient to do for every revision.

We know the set of collections on a replication won't change so this seems safe to me?

db/blip_sync_context.go
torcolvin: Yeah, let's leave this for later.. but I'm pretty sure this is because we have to decide when to reload user information.

base/collection_common.go
bbrks: Best I can do is this to avoid unnecessary double-iteration.

```go
collections = append(collections, base.ScopeAndCollectionName{
	Scope:      collection.ScopeName,
	Collection: collection.Name,
}.String())
```

db/special_docs.go
bbrks: Done in a single commit but it added a decent chunk of changes:

```
 10 files changed, 142 insertions(+), 142 deletions(-)
```

rest/utilities_testing.go
bbrks: `getCollectionsForBLIP` isn't sorted in the same way (its ranging over a uint-keyed map) but I've added a comment and said it matches `GetKeyspaces` at least

torcolvin (approved):

PR #6112, opened 2023-02-24
CBG-2321 Namespace sequence metadata keys

adamcfraser:
Adds infrastructure for namespacing metadata keys, and implements for database-scoped metadata documents. Defines a factory type for generating metadata keys (MetadataKeys) that is initialized when a databaseContext is created, and passed to consumers associated with that database.

CBG-2321

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1548/
- [ ] `GSI=true,xattrs=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1549/

### DIFF
--- a/base/constants_syncdocs.go
+++ b/base/constants_syncdocs.go
@@ -11,42 +11,328 @@ licenses/APL2.txt.
 package base
 
 import (
+	"errors"
 	"fmt"
+	"strconv"
 )
 
-const SyncDocPrefix = "_sync:" // SyncDocPrefix is a common prefix for all Sync Gateway metadata documents
+const SyncDocPrefix = "_sync:"                                 // Prefix for all legacy (non-namespaced) Sync Gateway metadata documents
+const MetadataIdPrefix = "m_"                                  // Prefix for metadataId, to prevent collisions between namespaced and non-namespaced metadata documents
+const SyncDocMetadataPrefix = SyncDocPrefix + MetadataIdPrefix // Prefix for all namespaced Sync Gateway metadata documents
+const DCPCheckpointPrefix = "dcp_ck:"
 
-// Sync Gateway Metadata documents
+// Sync Gateway Metadata document types
+type metadataKey int
+
 const (
-	SyncSeqKey                       = SyncDocPrefix + "seq"                           // SyncSeqKey is a counter document storing the SG sequence number of a collection
-	UnusedSeqPrefix                  = SyncDocPrefix + "unusedSeq:"                    // UnusedSeqPrefix stores a sequence that was reserved by SG but was unused
-	UnusedSeqRangePrefix             = SyncDocPrefix + "unusedSeqs:"                   // UnusedSeqRangePrefix stores a sequence range that was reserved by SG but was unused
-	RevBodyPrefix                    = SyncDocPrefix + "rb:"                           // RevBodyPrefix stores a conflicting revision's body
-	RevPrefix                        = SyncDocPrefix + "rev:"                          // RevPrefix stores an old revision's body for temporary lookup (in-flight requests or delta sync)
-	BackgroundProcessHeartbeatPrefix = SyncDocPrefix + "background_process:heartbeat:" // BackgroundProcessHeartbeatPrefix stores a background process heartbeat
-	BackgroundProcessStatusPrefix    = SyncDocPrefix + "background_process:status:"    // BackgroundProcessStatusPrefix stores a background process status
-	UserPrefix                       = SyncDocPrefix + "user:"                         // UserPrefix stores user documents keyed by user name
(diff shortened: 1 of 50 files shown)


base/util_testing.go
```diff
@@ -172,6 +172,10 @@ func (b *TestBucket) GetSingleDataStore() sgbucket.DataStore {
 	return b.Bucket.DefaultDataStore()
 }
 
+func (b *TestBucket) GetMetadataStore() sgbucket.DataStore {
```
torcolvin, line 175: should this be on a non test bucket as well?

base/constants_syncdocs.go
```diff
@@ -11,83 +11,345 @@ licenses/APL2.txt.
...
+	"background_process:status:",    // stores a background process status
+	"user:",                         // stores a user
+	"role:",                         // stores a role
+	"useremail:",                    // stores a role
+	"session:",                      // stores a session
+
+}
+
+func (m metadataKey) String() string {
+	return metadataKeyNames[m]
```
torcolvin, line 63: I don't feel strongly about this: I might have this return from a `map[string]string` defined in this function so you can't accidentally get these two lists out of sync.

base/constants_syncdocs.go
```diff
@@ -11,83 +11,345 @@ licenses/APL2.txt.
...
 	}
-	return fmt.Sprintf("%s:%s.%s", CollectionSyncFunctionKeyWithoutGroupID, scopeName, collectionName)
+	return m.heartbeaterPrefix
+}
+
+// SGCfgPrefix returns a document prefix to use for cfg documents (cbgt)
+//
+//	format: _sync:{m_$}:hb:[groupID]:   (collections)
+//	format: _sync:[groupID]:   (default)
+func (m *MetadataKeys) SGCfgPrefix(groupID string) string {
```
torcolvin, line 212: this is going to change for import cbgt and I wonder if we will need two ways of doing the same thing, because it is also used for heartbeat, etc.

db/active_replicator.go
```diff
@@ -34,11 +34,16 @@ type ActiveReplicator struct {
...
 	ar := &ActiveReplicator{
-		ID:        config.ID,
-		config:    config,
-		statusKey: replicationStatusKey(config.ID),
+		ID:     config.ID,
+		config: config,
 	}
 
+	metakeys := base.DefaultMetadataKeys
+	if config.ActiveDB != nil {
```
torcolvin, line 42: how can `ActiveDB` be nil here?

rest/changestest/changes_api_test.go
```diff
@@ -3619,12 +3619,16 @@ func TestChangesIncludeConflicts(t *testing.T) {
...
+		t.Skip("TestChangesLargeSequences doesn't support walrus - needs to customize " + base.DefaultMetadataKeys.SyncSeqKey() + " prior to db creation")
 	}
 	if !base.TestsDisableGSI() {
 		t.Skip("Requires N1QL due to CBG-361")
 
 	}
+
+	if base.TestsUseNamedCollections() {
+		t.Skip("Requires default collection to set " + base.DefaultMetadataKeys.SyncSeqKey())
+	}
```
torcolvin, line 3631: is it worthwhile making it possible to set large seqno on non default collections? for `RestTesterConfig` ?

torcolvin (commented):
This LGTM modulo the test passing.

rest/changestest/changes_api_test.go
adamcfraser: Probably worth it at some point, but I didn't think it was needed just for this test, as the test is mainly about json serialization working as intended in the REST API, and I thought this test running with the default collection would be sufficient coverage for regressions in that area.

db/active_replicator.go
adamcfraser: I was just being defensive in case tests were doing anything non-standard, but it's fair to say that the rest of the replicator code treats ActiveDB as always being non-nil.  I could drop the check.

base/constants_syncdocs.go
adamcfraser: Can you provide some more details about how this is changing for import? I think I'd assumed that import was just changing things like index definitions that are stored inside the cfg docs, but not the naming of the cfg docs themselves.

base/constants_syncdocs.go
adamcfraser: I'd originally set this up this way to do less work at String() time, although I suppose that's not strictly needed with the way metaKeys is being initialized.  I might leave as-is for now, open to changing in future.

base/util_testing.go
adamcfraser: I didn't need it for this PR, at least.

PR #6113, opened 2023-02-24
CBG-2491: Multi-collection ISGR

bbrks:
CBG-2491

- Allows ISGR replications to opt-into collections with a `collections_enabled` bool.
  - `collections_local` and `collections_remote` are reserved for restricting the set of collections and allowing remapping of collection names from local/remote.
  - Currently uses the full set of collections on a database if collections are enabled, otherwise continues to use the old replication protocol and replicate only the default collection.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1556/
  - Known/unrelated failure [CBG-2573](https://issues.couchbase.com/browse/CBG-2573?jql=text%20~%20%22TestBasicAttachmentRemoval%22)

### DIFF
--- a/db/active_replicator_checkpointer.go
+++ b/db/active_replicator_checkpointer.go
@@ -13,6 +13,7 @@ package db
 import (
 	"context"
 	"errors"
+	"fmt"
 	"sort"
 	"strings"
 	"sync"
@@ -30,7 +31,8 @@ type Checkpointer struct {
 	clientID           string
 	configHash         string
 	blipSender         *blip.Sender
-	activeDB           *Database
+	dataStore          base.DataStore // dataStore is where the checkpoints are stored
+	localDocExpirySecs int
 	checkpointInterval time.Duration
 	statusCallback     statusFunc // callback to retrieve status for associated replication
 	// lock guards the expectedSeqs slice, and processedSeqs map
@@ -50,6 +52,8 @@ type Checkpointer struct {
 	lastLocalCheckpointRevID string
 	// lastCheckpointSeq is the last checkpointed sequence
 	lastCheckpointSeq SequenceID
+	// collectionIdx is the replication index of the collection we're checkpointing for
+	collectionIdx *int
 
 	// expectedSeqCompactionThreshold is the number of expected sequences that we'll tolerate before considering compacting away already processed sequences
 	// time vs. space complexity tradeoff, since we need to iterate over the expectedSeqs slice to compact it
@@ -76,12 +80,13 @@ type CheckpointerStats struct {
 	GetCheckpointMissCount          int64
 }
 
-func NewCheckpointer(ctx context.Context, clientID string, configHash string, blipSender *blip.Sender, replicatorConfig *ActiveReplicatorConfig, statusCallback statusFunc) *Checkpointer {
+func NewCheckpointer(ctx context.Context, dataStore base.DataStore, clientID string, configHash string, blipSender *blip.Sender, replicatorConfig *ActiveReplicatorConfig, statusCallback statusFunc, collectionIdx *int) *Checkpointer {
 	return &Checkpointer{
 		clientID:           clientID,
 		configHash:         configHash,
 		blipSender:         blipSender,
-		activeDB:           replicatorConfig.ActiveDB,
+		dataStore:          dataStore,
+		localDocExpirySecs: int(replicatorConfig.ActiveDB.Options.LocalDocExpirySecs),
 		expectedSeqs:       make([]SequenceID, 0),
 		processedSeqs:      make(map[SequenceID]struct{}),
 		idAndRevLookup:     make(map[IDAndRev]SequenceID),
@@ -95,6 +100,7 @@ func NewCheckpointer(ctx context.Context, clientID string, configHash string, bl
 		},
 		statusCallback:                 statusCallback,
 		expectedSeqCompactionThreshold: defaultExpectedSeqCompactionThreshold,
+		collectionIdx:                  collectionIdx,
 	}
 }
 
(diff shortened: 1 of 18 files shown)


db/active_replicator_checkpointer.go
```diff
@@ -50,6 +51,8 @@ type Checkpointer struct {
 	lastLocalCheckpointRevID string
 	// lastCheckpointSeq is the last checkpointed sequence
 	lastCheckpointSeq SequenceID
+	// collectionIdx is the replication index of the collection we're checkpointing for
```
torcolvin: since collection idx is used in a bunch of different ways, call out that this is the blip collection idx from `GetCollections` sync message.

db/active_replicator_config.go
```diff
@@ -83,6 +83,13 @@ type ActiveReplicatorConfig struct {
 	// TotalReconnectTimeout, if non-zero, is the amount of time to wait before giving up trying to reconnect.
 	TotalReconnectTimeout time.Duration
 
+	// CollectionsEnabled can be set to replicate one or more named collections, rather than just the default collection.
+	CollectionsEnabled bool
+	// KeyspaceMap represents a set of dot-separated scope/collections that will be replicated.
+	// The values of the map can be set to map a collection on the active side to something else on the passive side.
+	// This map can be nil to replicate all scopes/collections for the database when CollectionsEnabled is set to true.
+	KeyspaceMap map[string]string // map of active "scope.collection" to passive "scope.collection"
```
torcolvin: should this be `ScopeAndCollectionName` ?

db/active_replicator_pull.go
```diff
@@ -205,31 +253,52 @@ func (apr *ActivePullReplicator) reset() error {
...
+		}
 	}
 
 	apr.lock.Lock()
-	apr.Checkpointer = nil
+	_ = apr.forEachCollection(func(c *activeReplicatorCollection) error {
+		c.Checkpointer = nil
+		return nil
+	})
 	apr.lock.Unlock()
```
torcolvin: drop this into a defer, just in case?

db/active_replicator_pull.go
```diff
@@ -148,38 +179,55 @@ func (apr *ActivePullReplicator) Complete() {
...
 
-	apr.Checkpointer = NewCheckpointer(apr.checkpointerCtx, apr.CheckpointID, checkpointHash, apr.blipSender, apr.config, apr.getPullStatus)
+	err := apr.forEachCollection(func(c *activeReplicatorCollection) error {
+		c.Checkpointer = NewCheckpointer(apr.checkpointerCtx, c.dataStore, apr.CheckpointID, checkpointHash, apr.blipSender, apr.config, apr.getPullStatus, c.collectionIdx)
 
-	var err error
-	apr.initialStatus, err = apr.Checkpointer.fetchCheckpoints()
-	base.InfofCtx(apr.ctx, base.KeyReplicate, "Initialized pull replication status: %+v", apr.initialStatus)
+		if !apr.config.CollectionsEnabled {
+			err := c.Checkpointer.fetchCheckpoints()
```
torcolvin: Since the checkpointer is scoped to the datastore, why do we not need to fetch checkpoints for other collections?

I think this is because we get the checkpoints from `GetCollections` message, but then maybe `fetchCheckpoint` should be renamed to something that indicates it is for the default collection only.

db/active_replicator_pull_collections.go
```diff
@@ -0,0 +1,52 @@
...
+		err = apr._subChanges(base.IntPtr(i), since)
+		if err != nil {
+			break
+		}
+	}
+
+	if err != nil {
+		// clean up anything we've opened so far
+		base.TracefCtx(apr.ctx, base.KeyReplicate, "cancelling the checkpointer context inside _startPullWithCollections where we send blip request")
+		apr.blipSender.Close()
```
torcolvin, line 46: Maybe this should set `apr.checkpointerCtx = nil` ?

db/active_replicator_push.go
```diff
@@ -181,37 +134,55 @@ func (apr *ActivePushReplicator) Complete() {
...
+			if highSeq.Before(safeSeq) {
+				highSeq = safeSeq
+			}
+		}
+		return nil
+	})
+	var highSeqStr string
+	if highSeq.IsNonZero() {
+		highSeqStr = highSeq.String()
 	}
```
torcolvin: Is it worth wrapping up this into common code between push and pull?

db/active_replicator_push.go
```diff
@@ -238,26 +209,35 @@ func (apr *ActivePushReplicator) reset() error {
...
 		return err
 	}
 
 	apr.lock.Lock()
-	apr.Checkpointer = nil
+	_ = apr.forEachCollection(func(c *activeReplicatorCollection) error {
+		c.Checkpointer = nil
+		return nil
+	})
 	apr.lock.Unlock()
```
torcolvin: i'd swap this to a defer unlock now that the function isn non trivial.

db/active_replicator_push.go
```diff
@@ -238,26 +209,35 @@ func (apr *ActivePushReplicator) reset() error {
...
 
 // registerCheckpointerCallbacks registers appropriate callback functions for checkpointing.
-func (apr *ActivePushReplicator) registerCheckpointerCallbacks() {
-	apr.blipSyncContext.sgr2PushAlreadyKnownSeqsCallback = apr.Checkpointer.AddAlreadyKnownSeq
-
-	apr.blipSyncContext.sgr2PushAddExpectedSeqsCallback = apr.Checkpointer.AddExpectedSeqs
+func (apr *ActivePushReplicator) registerCheckpointerCallbacks(c *activeReplicatorCollection) {
+	blipSyncContextCollection, err := apr.blipSyncContext.collections.get(c.collectionIdx)
+	if err != nil {
+		base.WarnfCtx(apr.ctx, "Unable to get blipSyncContextCollection for collection %v", c.collectionIdx)
```
torcolvin, line 220: This panics in the pull replicator, and maybe it's worth having a fatal error for this in both push and pull?

db/active_replicator_config.go
bbrks: Ended up replacing KeyspaceMap with something simpler to document in the case of no mappings:

`CollectionsLocal` and an optional `CollectionsRemote` array

db/active_replicator_pull.go
bbrks: Yep, fair point - renamed this to be default collection-specific

torcolvin (approved):

PR #6119, opened 2023-02-28
CBG-2721: add keep_zip to sgcollect options

mohammed-madi:
CBG-2721

sgcollect now deletes redact zip and normal zip files by default, unless keep_zip is set to true in the request json body. Deletes after upload, if upload is set to true, and deletes before exiting sgcollect if upload is set to false


tools/sgcollect_info
```diff
@@ -124,6 +124,8 @@ def create_option_parser():
                       help="specifies proxy for upload")
     parser.add_option("--tmp-dir", dest="tmp_dir", default=None,
                       help="set the temp dir used while processing collected data. Overrides the TMPDIR env variable if set")
+    parser.add_option("--keep-zip", dest="keep_zip", default="",
```
torcolvin: this should be `default=False` and `action: "store_true"` because this is boolean option.

This should only be possible to pass if you've also pass `--upload`. Add a test to validate that this is the case. I'd rename this to `--keep-zip-after-upload`

tools/sgcollect_info
```diff
@@ -787,15 +790,22 @@ def main():
...
-            do_upload_and_exit(zip_filename, upload_url, options.upload_proxy)
+            do_upload_and_exit(zip_filename, upload_url, options.upload_proxy, keep_zip)
 
     if options.redact_level != "none":
         print("Zipfile built: {0}".format(redact_zip_file))
 
     print("Zipfile built: {0}".format(zip_filename))
 
+    if not keep_zip:
+        os.remove(zip_filename)
```
torcolvin: I'd wrap these in a function

```
try:
     os.remove(zip_filename)
     print("Zipfile deleted: {0}".format(zip_filename))
except Exception as e:
      print("Zipfile deletion failed: {0} {1}".format(zip_filename, e))
```

and they only run when upload is in effective. Most users are going to run this from the command and _don't_ want their files deleted.

tools/sgcollect_info
```diff
@@ -787,15 +790,22 @@ def main():
     # Upload the zip to the URL to S3 if required
     if upload_url:
         if options.redact_level != "none":
-            do_upload_and_exit(redact_zip_file, upload_url, options.upload_proxy)
+            do_upload_and_exit(redact_zip_file, upload_url, options.upload_proxy, keep_zip)
         else:
-            do_upload_and_exit(zip_filename, upload_url, options.upload_proxy)
+            do_upload_and_exit(zip_filename, upload_url, options.upload_proxy, keep_zip)
```
torcolvin: Because the code is going to exit at this point - these functions call `sys.exit` and raise `SystemExit`, you probably want this to be:

```
if upload_url:
       try:
            if options.redact_level != "none":
                  do_upload_and_exit(redact_zip_file, upload_url, options.upload_proxy, keep_zip)
            else:
                  do_upload_and_exit(zip_filename, upload_url, options.upload_proxy, keep_zip)
      finally:
            remove_file(zip_file)
            remove_file(redact_zip_file)
```

see below about making sure that if we can't remove one file we don't remove another file

torcolvin (commented):
I added a test framework in https://github.com/couchbase/sync_gateway/pull/6120 and I'd like to see each of the issues hit by a unit test.

tools/sgcollect_info
```diff
@@ -73,6 +73,12 @@ USAGE = """usage: %prog [options] output_file.zip
 
 mydir = os.path.dirname(sys.argv[0])
 
+def delete_zip(filename):
+    try:
+     os.remove(filename)
+     print("Zipfile deleted: {0}".format(filename))
```
torcolvin: this has only two whitespace, but the rest of the file is 4 space?

tools/sgcollect_info
```diff
@@ -124,7 +130,7 @@ def create_option_parser():
                       help="specifies proxy for upload")
     parser.add_option("--tmp-dir", dest="tmp_dir", default=None,
                       help="set the temp dir used while processing collected data. Overrides the TMPDIR env variable if set")
-    parser.add_option("--keep-zip", dest="keep_zip", default="",
+    parser.add_option("--keep-zip", dest="keep_zip", action="store_true", default="False",
```
torcolvin: this is should be `False` and not `"False"`

tools/tasks.py
```diff
@@ -1065,7 +1065,7 @@ def disarm(self):
         self.p = None
 
 
-def do_upload_and_exit(path, url, proxy):
+def do_upload_and_exit(path, url, proxy, keepZip):
```
torcolvin: keepZip is unused here

tools/sgcollect_info
```diff
@@ -738,7 +747,7 @@ def main():
 
     # If user asked to just upload, then upload and exit
     if options.just_upload_into is not None:
-        do_upload_and_exit(args[0], options.just_upload_into, options.upload_proxy)
+        do_upload_and_exit(args[0], options.just_upload_into, options.upload_proxy, keep_zip)
```
torcolvin: I think here we could just avoid deleting it.

mohammed-madi: https://github.com/couchbase/sync_gateway/pull/6119/commits/4bb03785f8f89dda3648cfe2c45fca796c80ff2f

tools/tasks.py
mohammed-madi: https://github.com/couchbase/sync_gateway/pull/6119/commits/4bb03785f8f89dda3648cfe2c45fca796c80ff2f

torcolvin (approved):

PR #6122, opened 2023-03-01
CBG-2733 make tests pass without xattrs

torcolvin:
- Some changes to rest/indextest for account of different test names
- Changed the db/indexes_test.go to use defer to clean up failed tests, and actually drop all indexes if they actually create indexes that aren't accounted for by the db setup. I'd like to move these tests to /db/indexestest because they are costly to run and also have propensity to kill the indexing service, and at least they'd be isolated from other db tests.
- change test bucket pool to create indexes on `_default._default` as metadata only if using default collection. This holds true for `db` package, but if someone creates `NewRestTesterDefaultCollection`, then `server_context.go` will create all indexes on the default bucket anyway in test.

## Pre-review checklist

- [x] `GSI=true,xattrs=true,default_collection=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1562/
- [x] `GSI=true,xattrs=false,default_collection=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1563/
- [x] `GSI=true,xattrs=true,default_collection=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1564/
- [x] `GSI=true,xattrs=false,default_collection=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1565/

### DIFF
--- a/db/indexes_test.go
+++ b/db/indexes_test.go
@@ -15,6 +15,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"sort"
 	"testing"
 
 	sgbucket "github.com/couchbase/sg-bucket"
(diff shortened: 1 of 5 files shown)


db/indexes_test.go
```diff
@@ -58,50 +59,42 @@ func TestInitializeIndexes(t *testing.T) {
 			n1qlStore, isGoCBBucket := base.AsN1QLStore(collection.dataStore)
 			require.True(t, isGoCBBucket)
 
+			// drop and restore indexes between tests
+			defer func() {
+				options := InitializeIndexOptions{
+					FailFast:    false,
+					NumReplicas: 0,
+					Serverless:  db.IsServerless(),
+					UseXattrs:   base.TestUseXattrs(),
```
adamcfraser, line 68: The UseXattrs here doesn't match line 82.  Is that intentional?  Would it be better to move this defer below line 78 and just reuse `options`?

db/indexes_test.go
```diff
@@ -115,58 +108,68 @@ func TestPostUpgradeIndexesSimple(t *testing.T) {
...
-
 	}
+
+	var expectedRemovedIndexes []string
 	for _, sgIndex := range sgIndexes {
-		if !sgIndex.shouldCreate(options) {
-			expectedIndexes--
+		if sgIndex.shouldCreate(options) {
+			// when xattrs will be true, this index doesn't exist for removal
+			if sgIndex.simpleName != indexNames[IndexTombstones] {
```
adamcfraser: I don't think we should be looking for specific index names here - if we add another index with the same characteristics (xattrs only), this test will start failing and need additional index-specific handling.  

Why doesn't shouldCreate already evaluate whether IndexTombstones is going to be created?  It's doing a check for i.IsXattrOnly() in there.

adamcfraser: Is the underlying problem is that the initial cleanup using removeObsoleteIndexes on line 139 isn't properly removing the tombstone index?

adamcfraser (changes requested):

db/indexes_test.go
torcolvin: I'm going to change this to see if it's an xattr only index.

db/indexes_test.go
torcolvin: This will go away if I can drop indexes between tests, but basically it is restoring the state to what the bucket pool expects.

adamcfraser (approved):

PR #6136, opened 2023-03-09
CBG-2749 create helper functions to be changed as we pass through a request

torcolvin:
Created everything on `LogContext` to represent the state of logging, and helper functions to allow updating it. Calling `LogContextWith` with a `LogContext{}` as a second argument will reset it. I wonder if we should have an explicit function for that.

For consideration:

- Order the log information is outputted (see logging_context_test.go)
- Whether there are circumstances where we want to elide information in a logger.
- Places we might want to reset db/collection information, so log lines aren't that long.
 
## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1598/

### DIFF
--- a/rest/server_context.go
+++ b/rest/server_context.go
@@ -514,7 +514,11 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
 			MetadataIndexes: metadataIndexes,
 			UseXattrs:       config.UseXattrs(),
 		}
-		ctx := base.CollectionCtx(ctx, ds.GetName())
+		dsName, ok := base.AsDataStoreName(ds)
+		if !ok {
+			return fmt.Errorf("Could not get datastore name from %s", base.MD(ds.GetName()))
+		}
+		ctx := base.KeyspaceLogCtx(ctx, bucket.GetName(), dsName.ScopeName(), ds.GetName())
 		indexErr := db.InitializeIndexes(ctx, n1qlStore, options)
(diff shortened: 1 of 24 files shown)


rest/server_context.go
```diff
@@ -514,7 +514,11 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
 			MetadataIndexes: metadataIndexes,
 			UseXattrs:       config.UseXattrs(),
 		}
-		ctx := base.CollectionCtx(ctx, ds.GetName())
+		dsName, ok := base.AsDataStoreName(ds)
+		if !ok {
+			return false, fmt.Errorf("Could not get datastore name from %s", base.MD(ds.GetName()))
+		}
+		ctx := base.KeyspaceLogCtx(ctx, bucket.GetName(), dsName.ScopeName(), ds.GetName())
```
bbrks: `ds.GetName()` is fully qualified, resulting in this:

```
2023-03-16T12:36:23.042Z [INF] b:b1.foo.b1.foo.bar Initializing indexes with numReplicas: 0...
```

```suggestion
		ctx := base.KeyspaceLogCtx(ctx, bucket.GetName(), dsName.ScopeName(), ds.CollectionName())
```

bbrks (changes requested):
All looks good, aside from one thing

bbrks (approved):

PR #6142, opened 2023-03-11
CBG-2662 Require resync when a collection moves between databases

adamcfraser:
CBG-2662 
If a collection moves between databases, resync must be run with regenerateSequences=true in order to preserve sequence consistency.

When a collection is associated with a database, a syncInfo document (_sync:syncInfo) is written to the collection that stores the metadataID for the collection.

On database start, checks syncInfo for all collections.  On mismatch, starts the database in offline mode.

When in this state, a 'require_resync' property is included in the root db response and the root /_status endpoint with a slice of collections that require resync.

When resync w/ regenerate sequences is successfully run for a collection, syncInfo is updated for the collection.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1592/

### DIFF
--- a/base/constants_syncdocs.go
+++ b/base/constants_syncdocs.go
@@ -354,3 +354,45 @@ func CollectionSyncFunctionKeyWithGroupID(groupID string, scopeName, collectionN
 	}
 	return fmt.Sprintf("%s:%s.%s", CollectionSyncFunctionKeyWithoutGroupID, scopeName, collectionName)
 }
+
+// SyncInfo documents are stored in collections to identify the metadataID associated with sync metadata in that collection
+type SyncInfo struct {
+	MetadataID string `json:"metadataID"`
+}
+
+// initSyncInfo attempts to initialize syncInfo for a datastore
+//  1. If syncInfo doesn't exist, it is created for the specified metadataID
+//  2. If syncInfo exists with a matching metadataID, returns requiresResync=false
+//  3. If syncInfo exists with a non-matching metadataID, returns requiresResync=true
+func InitSyncInfo(ds DataStore, metadataID string) (requiresResync bool, err error) {
+
+	var syncInfo SyncInfo
+	_, fetchErr := ds.Get(SGSyncInfo, &syncInfo)
+	if IsKeyNotFoundError(ds, fetchErr) {
+		newSyncInfo := &SyncInfo{MetadataID: metadataID}
+		_, addErr := ds.Add(SGSyncInfo, 0, newSyncInfo)
+		if IsCasMismatch(addErr) {
+			// attempt new fetch
+			_, fetchErr = ds.Get(SGSyncInfo, &syncInfo)
+			if fetchErr != nil {
+				return true, fmt.Errorf("Error retrieving syncInfo (after failed add): %v", fetchErr)
+			}
+		} else if addErr != nil {
+			return true, fmt.Errorf("Error adding syncInfo: %v", addErr)
+		}
+		// successfully added
+		return false, nil
+	} else if fetchErr != nil {
+		return true, fmt.Errorf("Error retrieving syncInfo: %v", fetchErr)
+	}
+
+	return syncInfo.MetadataID != metadataID, nil
+}
+
+// SetSyncInfo sets syncInfo in a DataStore to the specified metadataID
+func SetSyncInfo(ds DataStore, metadataID string) error {
+	syncInfo := &SyncInfo{
+		MetadataID: metadataID,
+	}
+	return ds.Set(SGSyncInfo, 0, nil, syncInfo)
+}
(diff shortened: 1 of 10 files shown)


rest/adminapitest/collections_admin_api_test.go
```diff
@@ -162,3 +164,131 @@ func TestCollectionsSyncImportFunctions(t *testing.T) {
...
+	require.NoError(t, err)
+	rest.RequireStatus(t, resp, http.StatusCreated)
+
+	// Write documents to collection 1 and 2
+	ks_db1_c1 := db1Name + "." + scope + "." + collection1
+	ks_db1_c2 := db1Name + "." + scope + "." + collection2
+	ks_db2_c1 := db2Name + "." + scope + "." + collection1
+
+	resp = rt.SendAdminRequest("PUT", "/"+ks_db1_c1+"/testDoc1", `{"foo":"bar"}`)
+	resp = rt.SendAdminRequest("PUT", "/"+ks_db1_c2+"/testDoc2", `{"foo":"bar"}`)
```
torcolvin, line 219: we can use `{{.db1keyspace1}}` `{{.db1keyspace2}} here

base/constants_syncdocs.go
```diff
@@ -354,3 +354,45 @@ func CollectionSyncFunctionKeyWithGroupID(groupID string, scopeName, collectionN
...
+		newSyncInfo := &SyncInfo{MetadataID: metadataID}
+		_, addErr := ds.Add(SGSyncInfo, 0, newSyncInfo)
+		if IsCasMismatch(addErr) {
+			// attempt new fetch
+			_, fetchErr = ds.Get(SGSyncInfo, &syncInfo)
+			if fetchErr != nil {
+				return true, fmt.Errorf("Error retrieving syncInfo (after failed add): %v", fetchErr)
+			}
+		} else if addErr != nil {
+			return true, fmt.Errorf("Error adding syncInfo: %v", addErr)
```
torcolvin, line 381:
```suggestion
			return true, fmt.Errorf("Error adding syncInfo: %w", addErr)
```

base/constants_syncdocs.go
```diff
@@ -354,3 +354,45 @@ func CollectionSyncFunctionKeyWithGroupID(groupID string, scopeName, collectionN
...
+	var syncInfo SyncInfo
+	_, fetchErr := ds.Get(SGSyncInfo, &syncInfo)
+	if IsKeyNotFoundError(ds, fetchErr) {
+		newSyncInfo := &SyncInfo{MetadataID: metadataID}
+		_, addErr := ds.Add(SGSyncInfo, 0, newSyncInfo)
+		if IsCasMismatch(addErr) {
+			// attempt new fetch
+			_, fetchErr = ds.Get(SGSyncInfo, &syncInfo)
+			if fetchErr != nil {
+				return true, fmt.Errorf("Error retrieving syncInfo (after failed add): %v", fetchErr)
```
torcolvin, line 378:
```suggestion
				return true, fmt.Errorf("Error retrieving syncInfo (after failed add): %w", fetchErr)
```

base/constants_syncdocs.go
```diff
@@ -354,3 +354,45 @@ func CollectionSyncFunctionKeyWithGroupID(groupID string, scopeName, collectionN
...
+			}
+		} else if addErr != nil {
+			return true, fmt.Errorf("Error adding syncInfo: %v", addErr)
+		}
+		// successfully added
+		return false, nil
+	} else if fetchErr != nil {
+		return true, fmt.Errorf("Error retrieving syncInfo: %v", fetchErr)
```
torcolvin, line 386:
```suggestion
		return true, fmt.Errorf("Error retrieving syncInfo: %w", fetchErr)
```

db/background_mgr_resync_dcp.go
```diff
@@ -209,6 +209,20 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]interface
 			}
 
 		}
+
+		// If we regenerated sequences, update syncInfo for all collections affected
+		if regenerateSequences {
```
torcolvin, line 214: The behavior query and dcp resync is different. For DCP resync we update all the docs first, then if that works across all collections, then we'll set all collections resync to be correct. For query resync, we will update the syncInfo as soon as one collection sync works.

bbrks: I think the latter approach (query resync) is slightly more preferable but also not possible to do with DCP resync, as it streams over multiple collections.

If we change query resync to match the DCP behaviour, you'd end up potentially re-running resync unnecessarily for a collection, if you stopped the resync process prematurely and it had only completed a subset of collections.

DCP doesn't have this advantage because the stream is ordered differently and all collections would resync the same rate (assuming even distribution of docs across collections)

bbrks (approved):
LGTM

@torcolvin I agree with your inline suggestions, do you want to commit them or shall I?

PR #6147, opened 2023-03-14
CBG-2780 write sync docs for default collection with the expected values

torcolvin:
## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1597/

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -420,15 +420,7 @@ func NewDatabaseContext(ctx context.Context, dbName string, bucket base.Bucket, 
 	}
 
 	// Initialize metadata ID and keys
-	// TODO: apply length limit to metadataID
-	metadataID := ""
-	if !options.Scopes.onlyDefaultCollection() {
-		metadataID = dbName
-	}
-	if options.MetadataID != "" {
-		metadataID = options.MetadataID
-	}
-	metaKeys := base.NewMetadataKeys(metadataID)
+	metaKeys := base.NewMetadataKeys(options.MetadataID)
 	dbContext.MetadataKeys = metaKeys
 
 	cleanupFunctions = append(cleanupFunctions, func() {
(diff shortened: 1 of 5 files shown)


db/database.go
```diff
@@ -422,7 +422,10 @@ func NewDatabaseContext(ctx context.Context, dbName string, bucket base.Bucket,
 
 	// Initialize metadata ID and keys
 	// TODO: apply length limit to metadataID
-	metadataID := dbName
+	metadataID := ""
+	if !options.Scopes.onlyDefaultCollection() {
```
bbrks: Doesn't this mean that databases that opt-into a collection results in all existing users/metadata is lost (or in the wrong place at least)?

How do we plan on dealing with that?

bbrks (changes requested):
I have a question about how this would work for existing databases

db/database.go
torcolvin: I changed this so that the `bootstrapContext`'s metadata registry computes the persistent config. This makes this test pass, but the code has some peculiarities.

1. walrus, even in persistent config mode, doesn't have a concept of bootstrap registry. So all tests use walrus write the legacy config format.
2. if you create two databases with walrus OR legacy config in the same bucket they'll share sync metadata data and clobber each other.

I propose a two part fix to this problem. 

1. implement ability to use bootstrap config to walrus, making it easier for us to do multi database tests with walrus and use persistent config for all our tests
2. disable using collections and legacy config, because we don't even check when we _might_ have a problem.

Another idea which not one I am fond of, would be to create a bootstrap connection even when legacy config is enabled, so we have access to the config registry. This adds code complexity, and there's a code that assumes bootstrap connection == persistent config. Furthermore, it doesn't solve the walrus problem.

Regardless, I propose resolving this PR and deferring the above to a future jira ticket.

bbrks (approved):
LGTM 👍

db/database.go
torcolvin: https://issues.couchbase.com/browse/CBG-2787 for followup

PR #6148, opened 2023-03-15
CBG-2762: return error when collection_access has inexistent collections on /_user and _role endpoints

mohammed-madi:
CBG-2762

Describe your PR here...
- /_user and /_role endpoints now check if all collections in the passed collection_access map exist. if an inexistent collection is found, a 404 is returned

I'm unsure if the api docs need changing, both _user and _role specify 404 as a response but for Resource not found errors and not collection/scope not found

## [Integration Tests]
- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1600/

### DIFF
--- a/db/users.go
+++ b/db/users.go
@@ -120,7 +120,7 @@ func (dbc *DatabaseContext) UpdatePrincipal(ctx context.Context, updates *auth.P
 		if collectionAccessChanged {
 			changed = true
 		}
-		//if updates.CollectionAccess
+
 		for scopeName, collections := range updates.CollectionAccess {
 			for collectionName, _ := range collections {
 				_, err = dbc.GetDatabaseCollection(scopeName, collectionName)
--- a/db/users.go
+++ b/db/users.go
@@ -14,6 +14,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/couchbase/sync_gateway/auth"
 	"github.com/couchbase/sync_gateway/base"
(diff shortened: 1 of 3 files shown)


db/users.go
```diff
@@ -121,6 +121,15 @@ func (dbc *DatabaseContext) UpdatePrincipal(ctx context.Context, updates *auth.P
 			changed = true
 		}
 
+		for scopeName, collections := range updates.CollectionAccess {
```
torcolvin: To avoid an additional over collections, do this check in `dbc.RequiresCollectionAccessUpdate` and change that to return err as well. We can exit early there.

db/users.go
```diff
@@ -268,22 +258,27 @@ func (dbc *DatabaseContext) UpdateCollectionExplicitChannels(ctx context.Context
 }
 
 // RequiresCollectionAccessUpdate returns true if the provided map of CollectionAccessConfig requires an update to the principal
-func (dbc *DatabaseContext) RequiresCollectionAccessUpdate(ctx context.Context, princ auth.Principal, updates map[string]map[string]*auth.CollectionAccessConfig) (requiresUpdate bool) {
+func (dbc *DatabaseContext) RequiresCollectionAccessUpdate(ctx context.Context, princ auth.Principal, updates map[string]map[string]*auth.CollectionAccessConfig) (requiresUpdate bool, err error) {
```
torcolvin: I would say that it's a golang anti pattern to use named return values, it sort of just sucks about go because it makes the naming less clear, I'd drop named error codes here.

The risk here is that we refactor this function more and then `err` as below escapes its scope.

torcolvin (commented):
LGTM modulo minor changes, can you run this through jenkins?

db/users.go
```diff
@@ -116,9 +116,8 @@ func (dbc *DatabaseContext) UpdatePrincipal(ctx context.Context, updates *auth.P
 			changed = true
 		}
 
-		collectionAccessChanged := dbc.RequiresCollectionAccessUpdate(ctx, princ, updates.CollectionAccess)
-		if collectionAccessChanged {
-			changed = true
+		if changed, err = dbc.RequiresCollectionAccessUpdate(ctx, princ, updates.CollectionAccess); err != nil {
```
torcolvin: the bug here is that changed means something different than collection changed, so you need to keep `collectionAccessChanged` as a different variable than `changed`

torcolvin (approved):

PR #6150, opened 2023-03-16
CBG-2761: return bad request when writing read_only collection_access properties

mohammed-madi:
CBG-2761

Describe your PR here...
- Return 400 bad request when writing read_only collection_access properties for /_user and /_role endpoints

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1609/

### DIFF
--- a/rest/user_api_test.go
+++ b/rest/user_api_test.go
@@ -1369,9 +1369,6 @@ func TestGetUserCollectionAccess(t *testing.T) {
 	collectionPayload := fmt.Sprintf(`,"%s": {
 					"admin_channels":["foo", "bar1"]
 				}`, collection2Name)
-	rdOnlycollectionPayload := fmt.Sprintf(`,"%s": {
-					"jwt_channels":["foo", "bar1"]
-				}`, collection2Name)
 
 	// Create a user with collection metadata
 	userRolePayload := `{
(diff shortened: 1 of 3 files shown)


rest/user_api_test.go
```diff
@@ -1369,9 +1369,10 @@ func TestGetUserCollectionAccess(t *testing.T) {
 	collectionPayload := fmt.Sprintf(`,"%s": {
 					"admin_channels":["foo", "bar1"]
 				}`, collection2Name)
-	//rdOnlycollectionPayload := fmt.Sprintf(`,"%s": {
-	//				"jwt_channels":["foo", "bar1"]
-	//			}`, collection2Name)
+	rdOnlycollectionPayload := fmt.Sprintf(`,"%s": {
+					"jwt_channels":["foo", "bar1"]
+				}`, collection2Name)
```
torcolvin: can you write a loop for each unmodifable property?

rest/utilities_testing.go
```diff
@@ -1007,6 +1006,17 @@ func (rt *RestTester) SetAdminChannels(username string, keyspace string, channel
 	}
 
 	currentConfig.SetExplicitChannels(*scopeName, *collectionName, channels...)
+	// Remove read only properties returned from the user api
+	for _, scope := range currentConfig.CollectionAccess {
+		if scope != nil {
+			for _, collectionAccess := range scope {
+				collectionAccess.Channels_ = nil
```
torcolvin, line 1013: which tests are setting this property? is this just a result of marshal / unmarshalling?

mohammed-madi: TestMultiCollectionChangesUserDynamicGrant and TestMultiCollectionChangesUserDynamicGrantDCP
It's included in the user api response, which this function unmarshals, modifies then marshals and uses it in line 1022

torcolvin: I think I'm OK with this - I don't think adding the boilerplate to those tests would be helpful.

torcolvin (approved):

PR #6152, opened 2023-03-17
CBG-2493: Rollback bug causing oneshot behaviour on continuous clients

gregns1:
CBG-2493

High level below is to fix bug where if a rollback was needfed the feed would close after the rollback was done even if the client was not a oneshot client. This caused an issue in QE test where documents weren't getting streamed back to sync gateway mutation feed after docs were put into server through sync gateway. Essentially due to these docs hitting these particular vBuckets that ended rollback and there was no open stream to stream these to the mutation feed on sync gateway. 

This PR adds a function for creating initialising the endSeqNo map on the metadata on the DCP client for continuous clients. This was only being done for one shot clients before. This led to the map being initialised with 0's for all vbID's and the rollback setting the vBucket metadata end seqno to the end senqo in the map for the vBucket. Causing feeds to shutdown immediately as they end up with end seqno of 0. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1625/

### DIFF
--- a/base/dcp_client.go
+++ b/base/dcp_client.go
@@ -437,6 +437,7 @@ func (dc *DCPClient) openStream(vbID uint16, maxRetries uint32) (err error) {
 			InfofCtx(logCtx, KeyDCP, "Open stream for vbID %d failed due to rollback or range error, will roll back metadata and retry: %v", vbID, openStreamErr)
 			err := dc.rollback(vbID)
 			if err != nil {
+				DebugfCtx(logCtx, KeyDCP, "failed to rollback metadata for vb %d: error: %v", vbID, err)
 				return fmt.Errorf("metadata rollback failed for vb %d: %v", vbID, err)
 			}
 		case errors.Is(openStreamErr, gocbcore.ErrShutdown):
@@ -460,10 +461,17 @@ func (dc *DCPClient) openStream(vbID uint16, maxRetries uint32) (err error) {
 }
 
 func (dc *DCPClient) rollback(vbID uint16) (err error) {
+	logCtx := context.TODO()
 	if dc.dbStats != nil {
 		dc.dbStats.Add("dcp_rollback_count", 1)
 	}
-	dc.metadata.Rollback(vbID)
+	if dc.oneShot {
+		DebugfCtx(logCtx, KeyDCP, "initiating a oneshot rollback for vb: %d", vbID)
+		dc.metadata.OneshotRollback(vbID)
+	} else {
+		DebugfCtx(logCtx, KeyDCP, "initiating a continuous rollback for vb: %d", vbID)
+		dc.metadata.ContinuousRollback(vbID)
+	}
 	return nil
 }
 
--- a/base/dcp_client.go
+++ b/base/dcp_client.go
@@ -237,17 +237,6 @@ func (dc *DCPClient) configureOneShot() error {
 	return nil
 }
 
-func (dc *DCPClient) configureContinuous() {
-	var i uint16
-	maxSeqno := gocbcore.SeqNo(0xffffffffffffffff)
-	endSeqNos := make(map[uint16]uint64, dc.numVbuckets)
-
-	for i = 0; i < dc.numVbuckets; i++ {
-		endSeqNos[i] = uint64(maxSeqno)
-	}
-	dc.metadata.SetEndSeqNos(endSeqNos)
-}
-
 // Start returns an error and a channel to indicate when the DCPClient is done. If Start returns an error, DCPClient.Close() needs to be called.
 func (dc *DCPClient) Start() (doneChan chan error, err error) {
 	err = dc.initAgent(dc.spec)
@@ -259,8 +248,6 @@ func (dc *DCPClient) Start() (doneChan chan error, err error) {
 		if err != nil {
 			return dc.doneChannel, err
 		}
-	} else {
-		dc.configureContinuous()
 	}
 	dc.startWorkers()
 
--- a/base/dcp_client.go
+++ b/base/dcp_client.go
@@ -473,11 +473,11 @@ func (dc *DCPClient) openStream(vbID uint16, maxRetries uint32) (err error) {
 	return fmt.Errorf("openStream failed to complete after %d attempts, last error: %w", openRetryCount, openStreamErr)
 }
 
-func (dc *DCPClient) rollback(logCtx context.Context, vbID uint16) (err error) {
+func (dc *DCPClient) rollback(ctx context.Context, vbID uint16) (err error) {
 	if dc.dbStats != nil {
 		dc.dbStats.Add("dcp_rollback_count", 1)
 	}
-	dc.metadata.Rollback(logCtx, vbID)
+	dc.metadata.Rollback(ctx, vbID)
 	return nil
 }
 
--- a/base/dcp_client.go
+++ b/base/dcp_client.go
@@ -237,6 +237,17 @@ func (dc *DCPClient) configureOneShot() error {
 	return nil
 }
 
+func (dc *DCPClient) configureContinuous() {
+	var i uint16
+	maxSeqno := gocbcore.SeqNo(0xffffffffffffffff)
+	endSeqNos := make(map[uint16]uint64, dc.numVbuckets)
+
+	for i = 0; i < dc.numVbuckets; i++ {
+		endSeqNos[i] = uint64(maxSeqno)
+	}
+	dc.metadata.SetEndSeqNos(endSeqNos)
+}
+
 // Start returns an error and a channel to indicate when the DCPClient is done. If Start returns an error, DCPClient.Close() needs to be called.
 func (dc *DCPClient) Start() (doneChan chan error, err error) {
 	err = dc.initAgent(dc.spec)
@@ -248,6 +259,8 @@ func (dc *DCPClient) Start() (doneChan chan error, err error) {
 		if err != nil {
 			return dc.doneChannel, err
 		}
+	} else {
+		dc.configureContinuous()
 	}
 	dc.startWorkers()
 
@@ -435,9 +448,9 @@ func (dc *DCPClient) openStream(vbID uint16, maxRetries uint32) (err error) {
 				return fmt.Errorf("%s, failOnRollback requested", openStreamErr)
 			}
 			InfofCtx(logCtx, KeyDCP, "Open stream for vbID %d failed due to rollback or range error, will roll back metadata and retry: %v", vbID, openStreamErr)
-			err := dc.rollback(vbID)
+			err := dc.rollback(logCtx, vbID)
 			if err != nil {
-				DebugfCtx(logCtx, KeyDCP, "failed to rollback metadata for vb %d: error: %v", vbID, err)
+				WarnfCtx(logCtx, "failed to rollback metadata for vb %d: error: %v", vbID, err)
 				return fmt.Errorf("metadata rollback failed for vb %d: %v", vbID, err)
 			}
 		case errors.Is(openStreamErr, gocbcore.ErrShutdown):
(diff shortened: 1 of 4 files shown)


base/dcp_client.go
```diff
@@ -437,6 +437,7 @@ func (dc *DCPClient) openStream(vbID uint16, maxRetries uint32) (err error) {
 			InfofCtx(logCtx, KeyDCP, "Open stream for vbID %d failed due to rollback or range error, will roll back metadata and retry: %v", vbID, openStreamErr)
 			err := dc.rollback(vbID)
 			if err != nil {
+				DebugfCtx(logCtx, KeyDCP, "failed to rollback metadata for vb %d: error: %v", vbID, err)
```
torcolvin: I would actually be inclined to set this to WarnfCtx() because when it happens it's really bad and we would want to know about it on debug logging, which I believe is warning

base/dcp_client.go
```diff
@@ -460,10 +461,17 @@ func (dc *DCPClient) openStream(vbID uint16, maxRetries uint32) (err error) {
 }
 
 func (dc *DCPClient) rollback(vbID uint16) (err error) {
+	logCtx := context.TODO()
```
torcolvin: let's pass ctx into this function

base/dcp_client_metadata.go
```diff
@@ -294,3 +336,17 @@ func (m *DCPMetadataCS) getMetadataKey(workerID int) string {
 type WorkerMetadata struct {
 	DCPMeta map[uint16]DCPMetadata
 }
+
+// findRollbackStartSeqno finds the highest seqno available in the failover log and will use that as start seqno for the rollback.
+// This reduces the work needed on a rollback.
+func findRollbackStartSeqno(failoverLog []gocbcore.FailoverEntry) gocbcore.SeqNo {
+	var rollBackSeqno gocbcore.SeqNo = 0
+	if len(failoverLog) > 0 {
```
torcolvin: this seems unnecessary, the loop will just not occur if this happens

base/dcp_client_metadata.go
```diff
@@ -196,17 +218,37 @@ func NewDCPMetadataCS(store DataStore, numVbuckets uint16, numWorkers int, keyPr
...
+// after rollback is complete
+func (m *DCPMetadataCS) OneshotRollback(vbID uint16) {
 	// Preserve endSeqNo on rollback
 	endSeqNo := m.metadata[vbID].EndSeqNo
+	rollbackStartSeqno := findRollbackStartSeqno(m.metadata[vbID].FailoverEntries)
 	m.metadata[vbID] = DCPMetadata{
 		VbUUID:          0,
-		StartSeqNo:      0,
+		StartSeqNo:      rollbackStartSeqno,
 		EndSeqNo:        endSeqNo,
```
torcolvin: When I look more at this, endSeqNo should be preserved ! If you see this getting reset to not  `gocbcore.SeqNo(0xffffffffffffffff)` we are writing that in an incorrect way somewhere. Then we wouldn't need a separate function for continuous and non continous? We only call `SeqEndSeqNos` in one place, opening a one shot stream. So maybe the actual problem has not yet been identified?

My best guess would be somehow we are re-initializing metadata on failover and see `EndSeqNo` as 0? But this is just because I don't have an idea on how this could happen otherwise.

base/dcp_client_test.go
```diff
@@ -311,13 +311,114 @@ func TestDCPClientMultiFeedConsistency(t *testing.T) {
...
+		}
+		return false
+	}
+
+	feedID := t.Name()
+	gocbv2Bucket, err := AsGocbV2Bucket(bucket.Bucket)
+	require.NoError(t, err)
+
+	collection, ok := dataStore.(*Collection)
+	require.True(t, ok)
```
torcolvin: slightly prefer `AsCollection(dataStore)` here.

base/dcp_client_test.go
```diff
@@ -311,13 +311,114 @@ func TestDCPClientMultiFeedConsistency(t *testing.T) {
...
+
+	// Add documents
+	updatedBody := map[string]interface{}{"foo": "bar"}
+	for i := 0; i < 10000; i++ {
+		key := fmt.Sprintf("%s_%d", t.Name(), i)
+		err := dataStore.Set(key, 0, nil, updatedBody)
+		require.NoError(t, err)
+	}
+
+	// wait for a timeout to ensure client streams lal mutations over continuous feed
```
torcolvin:
```suggestion
	// wait for a timeout to ensure client streams all mutations over continuous feed
```

base/dcp_client_test.go
```diff
@@ -311,13 +311,114 @@ func TestDCPClientMultiFeedConsistency(t *testing.T) {
...
+
+	// function to force the rollback of some vBuckets
+	dcpClient.forceRollbackvBucket(vbUUID)
+
+	doneChan, startErr := dcpClient.Start()
+	require.NoError(t, startErr)
+
+	// Add documents
+	updatedBody := map[string]interface{}{"foo": "bar"}
+	for i := 0; i < 10000; i++ {
```
torcolvin: I'd drop numDocs as a constant here for easier manipulation

torcolvin (commented):
I wonder if you write documents and start and stop a DCP feed to see if you get real information out of what the FailoverEntry struct will look like, so we can actually test that out!

base/dcp_client.go
```diff
@@ -459,11 +473,11 @@ func (dc *DCPClient) openStream(vbID uint16, maxRetries uint32) (err error) {
 	return fmt.Errorf("openStream failed to complete after %d attempts, last error: %w", openRetryCount, openStreamErr)
 }
 
-func (dc *DCPClient) rollback(vbID uint16) (err error) {
+func (dc *DCPClient) rollback(logCtx context.Context, vbID uint16) (err error) {
```
torcolvin: I like adding contexts to functions but go idiom says they should always be `ctx` whether used for logging or not.

```suggestion
func (dc *DCPClient) rollback(ctx context.Context, vbID uint16) (err error) {
```

base/dcp_client.go
```diff
@@ -459,11 +473,11 @@ func (dc *DCPClient) openStream(vbID uint16, maxRetries uint32) (err error) {
...
 	if dc.dbStats != nil {
 		dc.dbStats.Add("dcp_rollback_count", 1)
 	}
-	dc.metadata.Rollback(vbID)
+	dc.metadata.Rollback(logCtx, vbID)
```
torcolvin:
```suggestion
	dc.metadata.Rollback(ctx, vbID)
```

base/dcp_client_metadata.go
```diff
@@ -83,8 +83,8 @@ func NewDCPMetadataMem(numVbuckets uint16) *DCPMetadataMem {
 	return m
 }
 
-// Rollback resets the metadata, preserving EndSeqNo if set
-func (m *DCPMetadataMem) Rollback(vbID uint16) {
+// Rollback resets the metadata, preserving EndSeqNo
+func (m *DCPMetadataMem) Rollback(logCtx context.Context, vbID uint16) {
```
torcolvin:
```suggestion
func (m *DCPMetadataMem) Rollback(ctx context.Context, vbID uint16) {
```

base/dcp_client_metadata.go
```diff
@@ -93,6 +93,7 @@ func (m *DCPMetadataMem) Rollback(vbID uint16) {
 		SnapEndSeqNo:    0,
 		FailoverEntries: make([]gocbcore.FailoverEntry, 0),
 	}
+	TracefCtx(logCtx, KeyDCP, "rolling back vb:%d with metadata set to %v", vbID, m.metadata[vbID])
```
torcolvin:
```suggestion
	TracefCtx(ctx, KeyDCP, "rolling back vb:%d with metadata set to %v", vbID, m.metadata[vbID])
```

base/dcp_client_metadata.go
```diff
@@ -83,8 +83,8 @@ func NewDCPMetadataMem(numVbuckets uint16) *DCPMetadataMem {
 	return m
 }
 
-// Rollback resets the metadata, preserving EndSeqNo if set
-func (m *DCPMetadataMem) Rollback(vbID uint16) {
+// Rollback resets the metadata, preserving EndSeqNo
```
torcolvin, line 84: I think we can remove `configureContinuous` and just set 

`EndSeqNo: m.metadata[vbID].endSeqNos[vbID]`

to match the behavior of `DCPMetadataCS`. 

In fact, the member `endSeqNos` of both metadata structs looks unused and I think we should just remove it from both structures.

base/dcp_client_metadata.go
```diff
@@ -196,8 +197,7 @@ func NewDCPMetadataCS(store DataStore, numVbuckets uint16, numWorkers int, keyPr
 	return m
 }
 
-func (m *DCPMetadataCS) Rollback(vbID uint16) {
-	// Preserve endSeqNo on rollback
+func (m *DCPMetadataCS) Rollback(logCtx context.Context, vbID uint16) {
```
torcolvin:
```suggestion
func (m *DCPMetadataCS) Rollback(ctx context.Context, vbID uint16) {
```

base/dcp_client_metadata.go
```diff
@@ -207,6 +207,7 @@ func (m *DCPMetadataCS) Rollback(vbID uint16) {
 		SnapEndSeqNo:    0,
 		FailoverEntries: make([]gocbcore.FailoverEntry, 0),
 	}
+	TracefCtx(logCtx, KeyDCP, "rolling back vb:%d with metadata set to %v", vbID, m.metadata[vbID])
```
torcolvin:
```suggestion
	TracefCtx(ctx, KeyDCP, "rolling back vb:%d with metadata set to %v", vbID, m.metadata[vbID])
```

base/dcp_client_test.go
```diff
@@ -311,13 +311,117 @@ func TestDCPClientMultiFeedConsistency(t *testing.T) {
...
 			}
 		})
 	}
 }
 
+func TestContinuousDCPRollback(t *testing.T) {
+	SetUpTestLogging(t, LevelDebug, KeyDCP)
+
+	if UnitTestUrlIsWalrus() {
+		t.Skip("This test only works against Couchbase Server")
```
torcolvin:
```suggestion
		t.Skip("This test requires DCP feed from gocb and therefore Couchbase Sever")
```

If we ever implement DCP in walrus.

base/dcp_client_test.go
```diff
@@ -311,13 +311,117 @@ func TestDCPClientMultiFeedConsistency(t *testing.T) {
...
+				require.Equal(t, uint16(0), uint16(numVBuckets))
 			case <-feed3Timeout:
 				t.Errorf("timeout waiting for first one-shot feed to complete")
 			}
 		})
 	}
 }
 
+func TestContinuousDCPRollback(t *testing.T) {
+	SetUpTestLogging(t, LevelDebug, KeyDCP)
```
torcolvin:
```suggestion
```
Remove debug logging when we are done with it to not make jenkins logs even bigger. You can always get more logs with `SG_TEST_LOG_LEVEL` env var too.

base/dcp_client_test.go
```diff
@@ -311,13 +311,117 @@ func TestDCPClientMultiFeedConsistency(t *testing.T) {
...
+	}
+
+	// wait for a timeout to ensure client streams all mutations over continuous feed
+	select {
+	case <-doneChan:
+		t.Fatalf("continuous client closed early")
+	case <-timeout:
+		mutationCount := atomic.LoadUint64(&mutationCount)
+		require.Equal(t, uint64(10000), mutationCount)
+	}
```
torcolvin: The way this test is written it will _always_ take 20s - this is unfortunate - we should be able to actually assert the count in the `counterCallback` and send a signal when it is done.

This is bad for impatient patient and fast computers and jenkins, which can be slow.

base/dcp_client_test.go
```diff
@@ -311,13 +311,117 @@ func TestDCPClientMultiFeedConsistency(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// function to force the rollback of some vBuckets
+	dcpClient.forceRollbackvBucket(vbUUID)
+
+	doneChan, startErr := dcpClient.Start()
+	require.NoError(t, startErr)
+
+	// Add documents
+	const numDocs = 10000
```
torcolvin: The fewer documents we can write to reproduce the issue, the better for test. We need to write enough or _know_ that that we are writing a document to a specific vbucket and that it will get mutated.

Note that linux/windows have 1024 vbuckets, but mac only has 64 of them.

torcolvin (commented):
Lots of small comments to make the code easier for us to understand, but I totally see the very subtle bug now and the test is awesome.

torcolvin (approved):

PR #6153, opened 2023-03-20
CBG-2767: fix test assertions on CheckpointHit and CheckpointMiss stats

gregns1:
CBG-2767

We were only calling `fetchDefaultCollectionCheckpoints` on default collections. Adding a new function to calculate the hit and miss count for collections on bothe the push and pull replicator. This is needed as ideally we would do this inside `_initCollections` but this is before we create a checkpointer object on each collection. So my solution is to have a function in each active pull and push replicator files to iterate trhough map of collections on the replicator and reason based on the since value being empty or not.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1623/
- [x] `GSI=true,xattrs=true` (use default collection) https://jenkins.sgwdev.com/job/SyncGateway-Integration/1622/

### DIFF
--- a/db/active_replicator_checkpointer.go
+++ b/db/active_replicator_checkpointer.go
@@ -388,14 +388,59 @@ func (r *replicationCheckpoint) Copy() *replicationCheckpoint {
 	}
 }
 
-// fetchCollectionCheckpoints sets lastCheckpointSeq for the given Checkpointer by requesting various checkpoints on the local and remote.
+// fetchNamedCollectionCheckpoints will:
+// - fetch the remote and local checkpoints off the checkpointer object
+// - increments miss count if the remote and local checkpoints are not the same
+// - increment hit count if successfully parse sequence ID
+// - increment miss count if error during parse sequence ID or if checkpoint sequence miss
+func (c *Checkpointer) fetchNamedCollectionCheckpoints() error {
+	localCheckpoint, err := c.getLocalCheckpoint()
+	if err != nil {
+		return err
+	}
+
+	base.DebugfCtx(c.ctx, base.KeyReplicate, "got local checkpoint: %v", localCheckpoint)
+	c.lastLocalCheckpointRevID = localCheckpoint.Rev
+
+	remoteCheckpoint, err := c.getRemoteCheckpoint()
+	if err != nil {
+		return err
+	}
+	base.DebugfCtx(c.ctx, base.KeyReplicate, "got remote checkpoint: %v", remoteCheckpoint)
+	c.lastRemoteCheckpointRevID = remoteCheckpoint.Rev
+
+	localSeq := localCheckpoint.LastSeq
+	remoteSeq := remoteCheckpoint.LastSeq
+
+	if localSeq != remoteSeq {
+		c.stats.GetCheckpointMissCount++
+		return nil
+	}
+
+	checkpointSeq := localSeq
+
+	if checkpointSeq != "" {
+		_, err = ParsePlainSequenceID(checkpointSeq)
+		if err == nil {
+			c.stats.GetCheckpointHitCount++
+		} else {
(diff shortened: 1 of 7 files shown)


db/active_replicator_checkpointer.go
```diff
@@ -388,6 +388,51 @@ func (r *replicationCheckpoint) Copy() *replicationCheckpoint {
...
+	}
+
+	base.DebugfCtx(c.ctx, base.KeyReplicate, "got local checkpoint: %v", localCheckpoint)
+	c.lastLocalCheckpointRevID = localCheckpoint.Rev
+
+	remoteCheckpoint, err := c.getRemoteCheckpoint()
+	if err != nil {
+		return err
+	}
+	base.DebugfCtx(c.ctx, base.KeyReplicate, "got remote checkpoint: %v", remoteCheckpoint)
```
torcolvin: This code is duplicated by the `_initCollections` code and I'm reluctant to call it twice. We have the checkpoint from `apr._initCollections`.

I'd either do this ticket two ways - 

1. in `_startPullWithCollections` and `_startPushWithCollections`, just put hit or miss stats right before we do `subChanges` based on whether `since` is empty string or not for collections code
2. Implement https://issues.couchbase.com/browse/CBG-2788 by separating how we call `getRemoteCheckpoint` as a string value that is passed into this function (it will either come from a direct call to `getRemoteCheckpoint` or from `replicationCheckpoint` from `_initCollections`, and call the same function for the default and non default.

I could see it is possible if we test for misses based on whether local checkpoint and remote checkpoint misses we won't hit this. In general, the miss is whether or not the checkpoint is present on the remote.

I'd be inclined to implement (1) as a stop gap simpler, and then address (2) as a separate ticket,  though some tests might have to _still_ be skipped.

db/active_replicator_push.go
```diff
@@ -341,3 +341,15 @@ func (apr *ActivePushReplicator) _startPushNonCollection() error {
...
+	for _, v := range apr.namedCollections {
+		if v.collectionIdx == collectionID {
+			if !since.IsNonZero() {
+				v.Checkpointer.stats.GetCheckpointMissCount++
+			} else {
+				v.Checkpointer.stats.GetCheckpointHitCount++
+			}
+		}
+	}
+}
```
bbrks: We could move this method onto `ActiveReplicatorCommon` (which is embedded inside `apr` - instead of having a separate copy for push and pull.

rest/replicatortest/replicator_test.go
```diff
@@ -2503,12 +2503,12 @@ func TestActiveReplicatorPullFromCheckpoint(t *testing.T) {
 	assert.Equal(t, int64(numRT2DocsInitial), pullCheckpointer.Stats().ExpectedSequenceCount)
 
 	// checkpoint assertions
-
-	// CBG-2767 skip assertions for GetCheckpoint stats
+	// CBG-2788 skip assertions till this ticket is implemented
 	if rt2.GetDatabase().OnlyDefaultCollection() {
 		assert.Equal(t, int64(0), pullCheckpointer.Stats().GetCheckpointHitCount)
 		assert.Equal(t, int64(1), pullCheckpointer.Stats().GetCheckpointMissCount)
 	}
```
bbrks: I was expecting these stats to work even without the work to fetch local checkpoints as well as remote, since fetching remote still should work the same, as long as there are no rollbacks/bucket flushes.

Is that not true? Do we know why?

bbrks (changes requested):
The change seems good to me (could maybe do a bit of DRY by moving into `ActiveReplicatorCommon`), although I'm not sure why all of the test assertions didn't work?

rest/replicatortest/replicator_test.go
gregns1: Ah I'm stupid. Got some of the logic wrong I think and for some reason in my head put it down to rollback cases etc. Apologies I'm updating that now.

bbrks (approved):

PR #6157, opened 2023-03-21
CBG-2759: Collection aware ISGR channel fitering

bbrks:
CBG-2759

- Replications can now optionally define a set of per-collection channels in the filter query params when using the `syncgateway/bychannel` filter.
  - Example:
    ```json
	{
      "direction": "pushAndPull",
      "remote": "http://remotesg:4985/remoteDb",
      "collections_enabled": true,
      "collections_local": ["scope1.collection1", "scope1.collection2"],
      "filter": "syncgateway/bychannel",
      "query_params": {
        "collections_channels": {
          "scope1.collection1": ["chanA", "chanB"],
          "scope1.collection2": ["chanc"]
        }
      }
    }
    ```
- The existing channel filter query parameter to can be used to apply to all collections.
  - This is useful in situations where you aren't specifying an explicit list of collections to replicate, but only want to replicate a subset of documents by channel in all collections.
  - Example:
    ```json
	{
      "direction": "pushAndPull",
      "remote": "http://remotesg:4985/remoteDb",
      "collections_enabled": true,
      "collections_local": ["scope1.collection1", "scope1.collection2"],
      "filter": "syncgateway/bychannel",
      "query_params": {
        "channels": ["chanA", "chanB"]
      }
    }
    ```

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1620/

### DIFF
--- a/db/active_replicator_common_collections.go
+++ b/db/active_replicator_common_collections.go
@@ -198,10 +198,14 @@ func (arc *activeReplicatorCommon) forEachCollection(callback func(*activeReplic
 // getFilteredChannels returns the filtered channels.
 // collectionIdx can be nil if replicating without collections enabled.
 func (config ActiveReplicatorConfig) getFilteredChannels(collectionIdx *int) []string {
-	if collectionIdx != nil {
-		if len(config.CollectionsChannelFilter)-1 >= *collectionIdx {
-			return config.CollectionsChannelFilter[*collectionIdx]
-		}
+	if collectionIdx == nil {
+		return config.FilterChannels
 	}
-	return config.FilterChannels
+
+	if len(config.CollectionsChannelFilter)-1 >= *collectionIdx {
+		return config.CollectionsChannelFilter[*collectionIdx]
+	}
+
+	// filter not present for collections, show all channels
+	return nil
 }
--- a/db/active_replicator_common_collections.go
+++ b/db/active_replicator_common_collections.go
@@ -198,14 +198,10 @@ func (arc *activeReplicatorCommon) forEachCollection(callback func(*activeReplic
 // getFilteredChannels returns the filtered channels.
 // collectionIdx can be nil if replicating without collections enabled.
 func (config ActiveReplicatorConfig) getFilteredChannels(collectionIdx *int) []string {
-	if collectionIdx == nil {
-		return config.FilterChannels
+	if collectionIdx != nil {
+		if len(config.CollectionsChannelFilter)-1 >= *collectionIdx {
+			return config.CollectionsChannelFilter[*collectionIdx]
+		}
 	}
-
-	if len(config.CollectionsChannelFilter)-1 >= *collectionIdx {
-		return config.CollectionsChannelFilter[*collectionIdx]
-	}
-
-	// filter not present for collections, show all channels
-	return nil
+	return config.FilterChannels
 }
--- a/db/active_replicator_common_collections.go
+++ b/db/active_replicator_common_collections.go
@@ -34,9 +34,17 @@ func (arc *activeReplicatorCommon) _initCollections() ([]replicationCheckpoint, 
 		getCollectionsCheckpointIDs []string
 	)
 
+	// ensure remote collection set is the same length as local collection set
 	if remoteLen := len(arc.config.CollectionsRemote); remoteLen > 0 {
 		if localLen := len(arc.config.CollectionsLocal); localLen != remoteLen {
 			return nil, fmt.Errorf("local and remote collections must be the same length... had %d and %d", localLen, remoteLen)
+		}
+	}
+
+	// ensure channel filter set is the same length as local collection set
+	if channelFilterLen := len(arc.config.CollectionsChannelFilter); channelFilterLen > 0 {
+		if localLen := len(arc.config.CollectionsLocal); localLen != channelFilterLen {
+			return nil, fmt.Errorf("local collections and channel filter set must be the same length... had %d and %d", localLen, channelFilterLen)
 		}
 	}
 
@@ -156,6 +164,21 @@ func (arc *activeReplicatorCommon) forEachCollection(callback func(*activeReplic
 	} else {
 		if err := callback(arc.defaultCollection); err != nil {
 			return err
+		}
+	}
+	return nil
+}
+
+// getFilteredChannels returns the filtered channels.
+// collectionIdx can be nil if replicating without collections enabled.
+func (config ActiveReplicatorConfig) getFilteredChannels(collectionIdx *int) []string {
+	if collectionIdx != nil {
+		if len(config.CollectionsChannelFilter) > 0 {
+			return config.CollectionsChannelFilter[*collectionIdx]
+		}
+	} else {
+		if config.FilterChannels != nil {
+			return config.FilterChannels
 		}
 	}
 	return nil
--- a/db/active_replicator_common_collections.go
+++ b/db/active_replicator_common_collections.go
@@ -12,6 +12,7 @@ import (
 	"encoding/json"
 	"fmt"
 
+	"github.com/couchbase/go-blip"
 	"github.com/couchbase/sync_gateway/base"
 )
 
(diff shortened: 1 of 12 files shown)


db/sg_replicate_cfg.go
```diff
@@ -591,7 +596,7 @@ func (m *sgReplicateManager) NewActiveReplicatorConfig(config *ReplicationCfg) (
 	// Channel filter processing
 	if config.Filter == base.ByChannelFilter {
 		rc.Filter = base.ByChannelFilter
-		rc.FilterChannels, err = ChannelsFromQueryParams(config.QueryParams)
+		err = rc.setFilterChannels(config)
```
torcolvin: nit:

```suggestion
		err := rc.setFilterChannels(config)
```

db/active_replicator_common_collections.go
```diff
@@ -34,12 +34,20 @@ func (arc *activeReplicatorCommon) _initCollections() ([]replicationCheckpoint,
...
 
+	// ensure remote collection set is the same length as local collection set
 	if remoteLen := len(arc.config.CollectionsRemote); remoteLen > 0 {
 		if localLen := len(arc.config.CollectionsLocal); localLen != remoteLen {
 			return nil, fmt.Errorf("local and remote collections must be the same length... had %d and %d", localLen, remoteLen)
 		}
 	}
 
+	// ensure channel filter set is the same length as local collection set
+	if channelFilterLen := len(arc.config.CollectionsChannelFilter); channelFilterLen > 0 {
```
torcolvin: should we make `arc.config.CollectionsChannelFilter` and `arc.config.ChannelFilter` mutually exclusive? Otherwise the filter will fall back to `ChannelFilter` as per `getFilteredChannels`

db/sg_replicate_util.go
```diff
@@ -12,36 +12,70 @@ package db
...
 	"github.com/couchbase/sync_gateway/base"
 )
 
-// QueryParams retrieves the channels associated with the byChannels a replication filter
-// from the generic queryParams interface{}.
-// The Channels may be passed as a JSON array of strings directly,
-// or embedded in a JSON object with the "channels" property and array value
-func ChannelsFromQueryParams(queryParams interface{}) (channels []string, err error) {
+// CollectionChannelsFromQueryParams returns the channels associated with the byChannel replication filter from the generic queryParams.
+// The channels may be passed in one of two ways:
```
torcolvin:
```suggestion
// The channels may be passed in one of three ways:
```

db/sg_replicate_util.go
```diff
@@ -12,36 +12,70 @@ package db
...
+//
+//  1. As a JSON array of strings directly:
+//     ["channel1", "channel2"]
+//
+//  2. As a JSON array of strings embedded in a JSON object with the "channels" property:
```
torcolvin: If passed like this, note that they are valid for all collections.

db/sg_replicate_util.go
```diff
@@ -12,36 +12,70 @@ package db
...
+		if chanarray, ok := val["channels"].([]interface{}); ok {
+			allCollectionsChannels, err = interfaceValsToStringVals(chanarray)
+			return nil, allCollectionsChannels, err
+		}
 
-	var chanarray []interface{}
-	if paramsmap, ok := queryParams.(map[string]interface{}); ok {
-		if chanarray, ok = paramsmap["channels"].([]interface{}); !ok {
-			return nil, errors.New("Replication specifies sync_gateway/bychannel filter, but query_params is missing channels property")
+		if collectionChannels, ok := val["collections_channels"].(map[string]interface{}); ok {
```
torcolvin: You could also check for both parameters here.

In order of least surprise to me would be the following behaviors. As long as we document and test the behavior I could be persuaded to use any of them.

1. `channels` is only for default collection and `collection_channels` is only for multiple collections, like the principal documents.
2. `channels` and `collection_channels` are mutually exclusive, but `channels` will imply collection1.channel1, collection2.channel1...
3. `channels` overrides `collection_channels` silently (what I think is the current behavior?)
4. `collection_channels` overrides channels silently
5. `collection_channels` takes precedence, but there's fallback behavior to channels if specified.

db/sg_replicate_util.go
```diff
@@ -12,36 +12,70 @@ package db
...
-			}
+}
+
+// interfaceValsToStringVals takes a []interface{} and returns a []string
+func interfaceValsToStringVals(interfaceVals []interface{}) ([]string, error) {
+	stringVals := make([]string, len(interfaceVals))
+	for i := range interfaceVals {
+		str, ok := interfaceVals[i].(string)
+		if !ok {
+			return nil, fmt.Errorf("Bad channel name %v (%T) in query_params for sync_gateway/bychannel filter - expecting strings", interfaceVals[i], interfaceVals[i])
```
torcolvin: do we need `base.UD` here, or is it OK because it will probably only make it back to user again?

rest/replicatortest/replicator_collection_test.go
```diff
@@ -170,6 +182,11 @@ func TestActiveReplicatorMultiCollection(t *testing.T) {
 		assert.Len(t, changes.Results, expectedNumDocs)
 
 		for j := 1; j <= numDocsPerCollection; j++ {
+			if !slices.Contains(collectionsFilter[i], channels[j%len(channels)]) {
+				// doc doesn't match channel filter
+				continue
+			}
```
torcolvin: Should we actually check if the doc does _not_ exist here are returns 404?

rest/replicatortest/replicator_test.go
```diff
@@ -1405,6 +1405,90 @@ func TestRequireReplicatorStoppedBeforeUpsert(t *testing.T) {
...
+				"` + rt1Keyspace + `": ["ChannelTwo"]
+			}
+		}
+	}`
+
+	resp = rt1.SendAdminRequest("PUT", "/{{.db}}/_replication/"+replicationID, replConfUpdate)
+	rest.RequireStatus(t, resp, http.StatusOK)
+
+	resp = rt1.SendAdminRequest("PUT", "/{{.db}}/_replicationStatus/"+replicationID+"?action=start", "")
+	rest.RequireStatus(t, resp, http.StatusOK)
```
torcolvin, line 1484: Does tearing down the database stop the replications? I have often thought about putting them in defers for testing

rest/replicatortest/replicator_test.go
```diff
@@ -1405,6 +1405,90 @@ func TestRequireReplicatorStoppedBeforeUpsert(t *testing.T) {
...
+	replConf := `
+	{
+		"replication_id": "` + replicationID + `",
+		"remote": "` + remoteURLString + `",
+		"direction": "push",
+		"continuous": true,
+		"filter":"sync_gateway/bychannel",
+		"query_params": {
+			"collections_channels": {
+				"` + rt1Keyspace + `": ["ChannelOne"]
```
torcolvin, line 1448: If this is running with the default collection, is this defined behavior? Does it matter if `_default._default` is explicitly defined?

Hand in hand with this question was wondering whether we should modify `RestTester.CreateReplicationForDB` for a named collection, that is, to pass the value of collection_channels for a keyspace if we find there's a single keyspace

torcolvin (commented):
Most comments are small, bigger question on how `channels` and `collection_channels` are going to be used together.

rest/replicatortest/replicator_test.go
bbrks: This test can run with default collection (`SG_TEST_USE_DEFAULT_COLLECTION=true`) as is - with the exception of the conditional `"collections_enabled"` value. That should always be true to allow a named `_default._default` to work.

rest/replicatortest/replicator_test.go
bbrks: Yep, via:

`DatabaseContext.Close()` -> `SGReplicateMgr.Stop()` -> `forEachReplicator` `.Stop()`

db/sg_replicate_util.go
bbrks: Decided to go with 2 - I can see it being useful.

db/active_replicator_common_collections.go
```diff
@@ -34,12 +34,23 @@ func (arc *activeReplicatorCommon) _initCollections() ([]replicationCheckpoint,
...
+	// ensure channel filter set is the same length as local collection set
+	if collectionsChannelFilterLen := len(arc.config.CollectionsChannelFilter); collectionsChannelFilterLen > 0 {
+		if localLen := len(arc.config.CollectionsLocal); localLen != collectionsChannelFilterLen {
+			return nil, fmt.Errorf("local collections and channel filter set must be the same length... had %d and %d", localLen, collectionsChannelFilterLen)
+		}
+		if channelFilterLen := len(arc.config.FilterChannels); channelFilterLen > 0 {
+			return nil, fmt.Errorf("channel filter and collection channel filter set cannot both be set")
+		}
+	}
+
```
torcolvin: nice to have because this function is getting massive, throw into a separate validation function.

db/active_replicator_common_collections.go
```diff
@@ -160,3 +171,14 @@ func (arc *activeReplicatorCommon) forEachCollection(callback func(*activeReplic
...
+// getFilteredChannels returns the filtered channels.
+// collectionIdx can be nil if replicating without collections enabled.
+func (config ActiveReplicatorConfig) getFilteredChannels(collectionIdx *int) []string {
+	if collectionIdx != nil {
+		if len(config.CollectionsChannelFilter)-1 >= *collectionIdx {
+			return config.CollectionsChannelFilter[*collectionIdx]
+		}
+	}
+	return config.FilterChannels
+}
```
torcolvin, line 207: I think this function works but it's almost by accident: I think an easier way of writing this is:

```
        // default collection (should we validate config.CollectionsEnabled when we call this?)
	if collectionIdx == nil {
             return config.FilterChannels
        }
	if len(config.CollectionsChannelFilter)-1 >= *collectionIdx {
			return config.CollectionsChannelFilter[*collectionIdx]
		}
        // filter not present for collections, show all channels
	return nil
}
```

The reason it works before is that `config.FilterChannels` was prevalidated to have empty list.

db/active_replicator_config.go
```diff
@@ -241,5 +246,21 @@ func (arc *ActiveReplicatorConfig) Equals(other *ActiveReplicatorConfig) bool {
...
+
+	if !reflect.DeepEqual(arc.CollectionsLocal, other.CollectionsLocal) {
+		return false
+	}
+
+	if !reflect.DeepEqual(arc.CollectionsRemote, other.CollectionsRemote) {
+		return false
+	}
+
+	if !reflect.DeepEqual(arc.CollectionsChannelFilter, other.CollectionsChannelFilter) {
```
torcolvin, line 261: Is it worth adding test code for this equality?

torcolvin (commented):
LGTM, small comments left.

db/active_replicator_common_collections.go
bbrks: Actually, I want it to behave like this so that `config.FilterChannels` can act across all collections (option 2 in earlier comment) as the fallback (assuming the earlier validation ensured that these two fields are mutually exclusive).

torcolvin (approved):

PR #6164, opened 2023-03-28
CBG-2789 implement channels.Set redaction

torcolvin:
This mostly does the right thing at the same cost as `base.Set` redaction. However, it looks like `base.Set` doesn't do sorting of elements? It's very hard to read without sorting but if we've gotten this long without sorting for `base.Set` then maybe it doesn't matter?

One way that we could optimize this is by storing the `channels.ID.serializationString` with `1.<ud>A</ud>` so we pay the cost at construction of disassembling the string. This would mean channelID.String() would always include the redaction tags but that's almost a benefit?

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/000/

### DIFF
--- a/channels/set.go
+++ b/channels/set.go
@@ -46,7 +46,7 @@ func NewID(channelName string, collectionID uint32) ID {
 	return ID{
 		Name:          channelName,
 		CollectionID:  collectionID,
-		serialization: strconv.FormatUint(uint64(collectionID), 10) + "." + channelName,
+		serialization: strconv.FormatUint(uint64(collectionID), 10) + "." + base.UserDataPrefix + channelName + base.UserDataSuffix,
 	}
 }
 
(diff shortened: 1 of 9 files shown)


channels/set.go
```diff
@@ -167,3 +167,35 @@ func (s Set) String() string {
...
+func (redactorSet RedactorSet) GetRedactionString(shouldRedact bool) string {
+	if !shouldRedact {
+		return redactorSet.set.String()
+	}
+	keys := make([]string, 0, len(redactorSet.set))
+	for channelID := range redactorSet.set {
+		keys = append(keys, fmt.Sprintf("%d.%s", channelID.CollectionID, redactorSet.redactorFunc(channelID.Name).Redact()))
+	}
+	sort.Strings(keys)
+	return fmt.Sprintf("{%s}", strings.Join(keys, ", "))
```
adamcfraser: If we don't need the sort (and I don't think we do, since we're not using for base.Set), I think it makes sense to optimize this function in the same way we're doing in base.RedactorSet.GetRedactionString:  avoid the use of Sprintf (which has overhead to parse the format) and avoid strings.Join (which is second iteration over the set).

adamcfraser (changes requested):

channels/set.go
torcolvin: The `base.Set` code which I copied this from is self inconsistent - `base.UD(set)` will not sort but `base.Set.String()` will sort the strings! 

I decided, arbitrarily that all channel data is de facto user data so even the standard string printing will show the redaction. This was for optimization.

adamcfraser (approved):

PR #6166, opened 2023-03-29
CBG-2766 follow up: Replication stats/status writer separate from checkpointer

bbrks:
CBG-2766 follow up

- Introduces LoggingMutex types. I'm OK with removing this from the PR if it's contentious. It was just used for debugging a deadlock.
- Introduces another background process in the replicator that is used to periodically update replication stats, outside of the regular state change updates.
- Prior to collections, this was done when the checkpointer ran, but now we have N checkpointers, we do not want to be writing N status updates each time.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1645/
- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1647/

### DIFF
--- a/db/active_replicator_common.go
+++ b/db/active_replicator_common.go
@@ -352,7 +352,7 @@ func (a *activeReplicatorCommon) getCheckpointHighSeq() string {
 
 func (a *activeReplicatorCommon) _publishStatus() {
 	status := a._getStatusCallback()
-	_, err := setLocalStatus(a.ctx, a.config.ActiveDB.MetadataStore, a.CheckpointID, status, int(a.config.ActiveDB.Options.LocalDocExpirySecs))
+	err := setLocalStatus(a.ctx, a.config.ActiveDB.MetadataStore, a.CheckpointID, status, int(a.config.ActiveDB.Options.LocalDocExpirySecs))
 	if err != nil {
 		base.WarnfCtx(a.ctx, "Couldn't set status for replication: %v", err)
 	}
@@ -379,10 +379,8 @@ func (arc *activeReplicatorCommon) startStatusReporter() error {
 	return nil
 }
 
-// getLocalStatus retrieves replication status for a given client ID from the given metadataStore
-func getLocalStatus(ctx context.Context, metadataStore base.DataStore, clientID string) (*ReplicationStatusDoc, error) {
-	base.TracefCtx(ctx, base.KeyReplicate, "getLocalStatus for %q", clientID)
-
+// getLocalStatusDoc retrieves replication status document for a given client ID from the given metadataStore
+func getLocalStatusDoc(ctx context.Context, metadataStore base.DataStore, clientID string) (*ReplicationStatusDoc, error) {
 	statusDocBytes, err := getSpecialBytes(metadataStore, DocTypeLocal, ReplicationStatusDocIDPrefix+clientID, 0)
 	if err != nil {
 		if !base.IsKeyNotFoundError(metadataStore, err) {
@@ -396,15 +394,27 @@ func getLocalStatus(ctx context.Context, metadataStore base.DataStore, clientID 
 	return statusDoc, err
 }
 
-// setLocalStatus updates replication status. Increments existing rev.
-func setLocalStatus(ctx context.Context, metadataStore base.DataStore, clientID string, status *ReplicationStatus, localDocExpirySecs int) (newRevID string, err error) {
+// getLocalStatus retrieves replication status for a given client ID from the given metadataStore
+func getLocalStatus(ctx context.Context, metadataStore base.DataStore, clientID string) (*ReplicationStatus, error) {
+	localStatusDoc, err := getLocalStatusDoc(ctx, metadataStore, clientID)
+	if err != nil {
+		return nil, err
+	}
+	if localStatusDoc != nil {
+		return localStatusDoc.Status, nil
+	}
+	return nil, nil
+}
+
+// setLocalStatus updates replication status.
+func setLocalStatus(ctx context.Context, metadataStore base.DataStore, clientID string, status *ReplicationStatus, localDocExpirySecs int) (err error) {
 	base.TracefCtx(ctx, base.KeyReplicate, "setLocalStatus for %q (%v)", clientID, status)
 
-	// getLocalStatus to obtain the current rev
-	currentStatus, err := getLocalStatus(ctx, metadataStore, clientID)
+	// obtain current rev
+	currentStatus, err := getLocalStatusDoc(ctx, metadataStore, clientID)
 	if err != nil {
 		base.WarnfCtx(ctx, "Unable to retrieve local status doc for %s, status not updated", clientID)
-		return "", nil
+		return nil
 	}
 
 	var revID string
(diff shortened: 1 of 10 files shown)


db/active_replicator_pull.go
```diff
@@ -224,16 +230,12 @@ func (apr *ActivePullReplicator) _initCheckpointer(remoteCheckpoints []replicati
...
+func (apr *ActivePullReplicator) _getStatus() *ReplicationStatus {
+	status := &ReplicationStatus{
+		ID: apr.CheckpointID,
+	}
 
-// getPullStatus is used internally, and passed as statusCallback to checkpointer
-func (apr *ActivePullReplicator) getPullStatus(lastSeqPulled string) *ReplicationStatus {
-	status := &ReplicationStatus{}
-	status.Status, status.ErrorMessage = apr.getStateWithErrorMessage()
+	status.Status, status.ErrorMessage = apr._getStateWithErrorMessage()
```
torcolvin, line 239: Why is this OK to call without a lock?

db/active_replicator_common.go
```diff
@@ -320,29 +325,121 @@ func (a *activeReplicatorCommon) GetStats() *BlipSyncStats {
...
+	if interval == 0 {
+		interval = DefaultCheckpointInterval
+	}
+	go func(ctx context.Context) {
+		ticker := time.NewTicker(interval)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ticker.C:
+				base.TracefCtx(arc.ctx, base.KeyReplicate, "writing status. context is not cancelled here")
```
torcolvin: This is going to log every 5 secs, I propose removing:

```suggestion
```

db/active_replicator_common.go
```diff
@@ -320,29 +325,121 @@ func (a *activeReplicatorCommon) GetStats() *BlipSyncStats {
...
+	if err != nil {
+		base.WarnfCtx(a.ctx, "Couldn't set status for replication: %v", err)
+	}
+}
+
+func (arc *activeReplicatorCommon) startStatusReporter() error {
+	interval := arc.config.CheckpointInterval
+	if interval == 0 {
+		interval = DefaultCheckpointInterval
+	}
```
torcolvin: Is this worth putting in the constructor or somewhere more clear? I thought about `newActiveReplicatorCommon`?

I'm fine here too.

db/active_replicator_common.go
```diff
@@ -320,29 +325,121 @@ func (a *activeReplicatorCommon) GetStats() *BlipSyncStats {
...
+		}
+		base.DebugfCtx(ctx, base.KeyReplicate, "couldn't find existing local checkpoint for ID %q", clientID)
+		return nil, nil
+	}
+	var statusDoc *ReplicationStatusDoc
+	err = base.JSONUnmarshal(statusDocBytes, &statusDoc)
+	return statusDoc, err
+}
+
+// setLocalStatus updates replication status. Increments existing rev.
```
torcolvin: Is is correct to say increment the existing rev?

This might be pedantic to try to clarify, but so I understand at least, this will update the revision stored in the metadata store to the latest version by the version set by the checkpoint.

It would be impossible to decrement probably, but it is not necessarily incrementing by 1, and it can start at empty string?

db/active_replicator_common.go
```diff
@@ -320,29 +325,121 @@ func (a *activeReplicatorCommon) GetStats() *BlipSyncStats {
...
+	}
+	if currentStatus == nil {
+		currentStatus = &ReplicationStatusDoc{
+			Status: &ReplicationStatus{
+				ID: clientID,
+			},
+		}
 	}
+
+	_, err = putSpecial(metadataStore, DocTypeLocal, ReplicationStatusDocIDPrefix+clientID, currentStatus.Rev, nil, 0)
```
torcolvin: Since we have multiple databases in a bucket, I'd expect `ReplicationStatusDocIDPrefix` to use `MetadataKeys` do be db scoped. This might not be necessarily because clientID will be unique? (or is clientID always whatever the user names???)

If it's not guaranteed to be unique, this has to be metadata spaced, and if it is it might be nice for debugging.

torcolvin (commented):
I like the idea of a thing we can substitute in that writes logging for a mutex, but even in a typical scenario of enabling trace logging, I wouldn't use it.

I might leave the code so if we needed to use it for debugging we'd use that.

In general, I'm confused why there are places we get the status without a lock, and maybe clarifying those with comments would be good. I did not comment on all them.

db/active_replicator_common.go
bbrks: Updated to remove rev ID references and return value - implementation detail.

db/active_replicator_pull.go
bbrks: The caller of `_getStatus` is either:
- Acquiring the read lock (`GetStatus`)
- Is the constructor of `apr`, ensuring nobody else can read/write before we've finished constructing and returned it for other goroutines to access.

db/active_replicator_common.go
bbrks: Agreed. I'll revisit this on Monday when I have a bit more time.

db/active_replicator_common.go
```diff
@@ -320,29 +327,124 @@ func (a *activeReplicatorCommon) GetStats() *BlipSyncStats {
...
+	go func(ctx context.Context) {
+		ticker := time.NewTicker(arc.config.CheckpointInterval)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ticker.C:
+				func() {
+					arc.lock.RLock()
+					defer arc.lock.RUnlock()
+					arc._publishStatus()
```
torcolvin, line 381: If we have a continuous replication but there's no activity, does it still make sense to update status? (I don't have a reasons to believe otherwise, just asking)

db/active_replicator_common.go
```diff
@@ -40,12 +39,14 @@ type activeReplicatorCommon struct {
 	checkpointerCtxCancel context.CancelFunc
 	CheckpointID          string // Used for checkpoint retrieval when Checkpointer isn't available
 	initialStatus         *ReplicationStatus
+	statusKey             string // key used when persisting replication status
 	state                 string
 	lastError             error
-	stateErrorLock        sync.RWMutex // state and lastError share their own mutex to support retrieval while holding the main lock
+	stateErrorLock        base.LoggingRWMutex // state and lastError share their own mutex to support retrieval while holding the main lock
```
bbrks:
```suggestion
	stateErrorLock        sync.RWMutex // state and lastError share their own mutex to support retrieval while holding the main lock
```

db/active_replicator_common.go
```diff
@@ -40,12 +39,14 @@ type activeReplicatorCommon struct {
...
 	replicationStats      *BlipSyncStats
+	_getStatusCallback    func() *ReplicationStatus
 	onReplicatorComplete  ReplicatorCompleteFunc
-	lock                  sync.RWMutex
+	lock                  base.LoggingRWMutex
```
bbrks:
```suggestion
	lock                  sync.RWMutex
```

db/active_replicator_common.go
```diff
@@ -302,14 +316,17 @@ func (a *activeReplicatorCommon) getLastError() error {
 	return a.lastError
 }
 
-func (a *activeReplicatorCommon) getStateWithErrorMessage() (state string, lastErrorMessage string) {
-	a.stateErrorLock.RLock()
-	defer a.stateErrorLock.RUnlock()
+func (a *activeReplicatorCommon) _getStateWithErrorMessage() (state string, lastErrorMessage string) {
```
torcolvin, line 320: Can we note this function is only safe to call when stateErrorLock is acquired?

I think we can push this code as is, but I think I'd be likely to break this state into its own datastructure, so you can indicate when you are calling it like

`GetStateRLock() (string, string)`
`GetStateWithoutLock() (string, string)`

db/active_replicator_push.go
```diff
@@ -180,16 +186,9 @@ func (apr *ActivePushReplicator) _initCheckpointer(remoteCheckpoints []replicati
...
 
-// GetStatus is used externally to retrieve pull replication status.  Combines current running stats with
-// initialStatus.
-func (apr *ActivePushReplicator) GetStatus() *ReplicationStatus {
-	return apr.getPushStatus(apr.getCheckpointHighSeq())
-}
-
-// getPullStatus is used internally, and passed as statusCallback to checkpointer
-func (apr *ActivePushReplicator) getPushStatus(lastSeqPushed string) *ReplicationStatus {
+func (apr *ActivePushReplicator) _getStatus() *ReplicationStatus {
```
torcolvin, line 190: Can we note here that `_getStatus()` needs to be called apr.stateLock already acquired?

torcolvin (dismissed):

db/active_replicator_common.go
bbrks: Underscore and non-underscore variants of functions have prior-art but I do agree with noting _which_ lock is required.

e.g:
- `changeListener.CurrentCount()`
- `ServerContext.fetchAndLoadDatabase()`
- `CollectionBucket.GetCollectionID()`
- `ServerContext.isDatabaseSuspended()`

torcolvin (approved):

PR #6170, opened 2023-04-03
CBG-2799 rollback to latest sequence if available

torcolvin:
Use new versions of gocbcore / cbgt for https://issues.couchbase.com/browse/GOCBC-1401.

This updates the direct dependency only on testify for reasons I don't understand from `go get github.com/couchbase/gocbcore/v10@cab6da1895aee0ee014156ecfb00ef911664476d` but it's a test only dependency and I'm not worried about it.

This does pass my non automated rollback tests for just a DCP feed, and I added a new test for rollback out of range errors.

I didn't test import feed.

This should improve the behavior of attachment compaction / dcp resync background processes, but if they get stuck with bad metadata they will need to be called with `?reset=true` to update metadata https://issues.couchbase.com/browse/CBG-2793 covers this, and it would be done by reading the doneChannel error of `DCPClient` and deleting the named metadata.

## Pre-review checklist

- [] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1653/


base/dcp_client.go
```diff
@@ -460,14 +460,59 @@ func (dc *DCPClient) openStream(vbID uint16, maxRetries uint32) (err error) {
...
+// getFailoverLogs querys for the current failover logs for a given vBucket ID.
+func (dc *DCPClient) getFailoverLogs(ctx context.Context, vbID uint16) ([]gocbcore.FailoverEntry, error) {
+	getFailoverErr := make(chan error)
+	var failoverLogs []gocbcore.FailoverEntry
+	getFailoverLogsCallback := func(f []gocbcore.FailoverEntry, err error) {
+		failoverLogs = f
+		getFailoverErr <- err
+	}
+
+	_, err := dc.agent.GetFailoverLog(vbID, getFailoverLogsCallback)
```
adamcfraser: Does GetFailoverLog have any recoverable error states (where we should retry, instead of returning error and panicking).

base/dcp_client.go
```diff
@@ -460,14 +460,59 @@ func (dc *DCPClient) openStream(vbID uint16, maxRetries uint32) (err error) {
...
+	if err != nil {
+		return err
+	}
+	dc.metadata.Rollback(ctx, vbID, vbUUID, seqNo)
 	return nil
 }
 
+// findRollbackSequence returns the vbuuid and seqno to rollback to for the given vbID
+func findRollbackVBUUIDSequence(ctx context.Context, vbID uint16, existingHighSeqNo gocbcore.SeqNo, entries []gocbcore.FailoverEntry) (gocbcore.VbUUID, gocbcore.SeqNo, error) {
+	if len(entries) == 0 {
```
adamcfraser: Is there ever a valid case for len(entries) == 0?  Zero entries isn't returned as an error case in gocbcore's dcpComponent.GetFailoverLog - should this scenario be treated as "we should actually start from zero" here?

base/dcp_client.go
```diff
@@ -437,7 +437,7 @@ func (dc *DCPClient) openStream(vbID uint16, maxRetries uint32) (err error) {
 			InfofCtx(logCtx, KeyDCP, "Open stream for vbID %d failed due to rollback or range error, will roll back metadata and retry: %v", vbID, openStreamErr)
 			err := dc.rollback(logCtx, vbID)
 			if err != nil {
-				WarnfCtx(logCtx, "failed to rollback metadata for vb %d: error: %v", vbID, err)
+				PanicfCtx(logCtx, "failed to rollback metadata for vb %d: error: %v", vbID, err)
```
adamcfraser: Is it possible to have the decision to panic left to a particular DCP stream type?  Or are there no feeds that it's better to roll back to zero?

base/dcp_client.go
```diff
@@ -460,14 +460,59 @@ func (dc *DCPClient) openStream(vbID uint16, maxRetries uint32) (err error) {
 	return fmt.Errorf("openStream failed to complete after %d attempts, last error: %w", openRetryCount, openStreamErr)
 }
 
-func (dc *DCPClient) rollback(ctx context.Context, vbID uint16) (err error) {
+// rollback changes the metadata of the client for the specified vbucket to mark the new startSeqNo and vBucket UUID.
+func (dc *DCPClient) rollback(ctx context.Context, vbID uint16) error {
```
adamcfraser: If we hit an error retrieving the failover logs or finding the rollback sequence, should we roll the metadata back to zero prior to returning error and triggering panic?  (so that in the case of persisted metadata, we're not just going to roll back again after restart)

base/dcp_client_metadata.go
```diff
@@ -82,17 +82,16 @@ func NewDCPMetadataMem(numVbuckets uint16) *DCPMetadataMem {
...
-func (m *DCPMetadataMem) Rollback(ctx context.Context, vbID uint16) {
-	endSeqNo := m.metadata[vbID].EndSeqNo
+func (m *DCPMetadataMem) Rollback(ctx context.Context, vbID uint16, vbUUID gocbcore.VbUUID, startSeqNo gocbcore.SeqNo) {
 	m.metadata[vbID] = DCPMetadata{
-		VbUUID:          0,
-		StartSeqNo:      0,
-		EndSeqNo:        endSeqNo,
-		SnapStartSeqNo:  0,
-		SnapEndSeqNo:    0,
+		VbUUID:          vbUUID,
```
adamcfraser: Is there a safety check required that startSeqNo is less than EndSeqNo?  Or is that valid and the stream is just closed immediately by the remote.

adamcfraser (changes requested):
A few areas for review/discussion.

base/dcp_client.go
torcolvin: The scenario here is if there's no active vbuckets - e.g. in a 2 node cluster you have a failover and have lost some of your vbuckets.

base/dcp_client.go
torcolvin: This should definitely get retried since it can fail for any reason for a kv op and gocbcore's code explicitly does no retries. I'm not sure how long we should retry here though..

adamcfraser: We use BucketSpec.RetrySleeper elsewhere, is the BucketSpec on DCPClient populated?

base/dcp_client.go
adamcfraser: Ok.  In that case, we'd only want to return error if existingHighSeqNo is greater than zero, do you agree?

base/dcp_client_metadata.go
```diff
@@ -81,45 +93,47 @@ func NewDCPMetadataMem(numVbuckets uint16) *DCPMetadataMem {
...
+		if failoverLog.SeqNo <= startSeqNo {
+			rollbackVbuuid = failoverLog.VbUUID
+			break
+		}
 	}
-	TracefCtx(ctx, KeyDCP, "rolling back vb:%d with metadata set to %v", vbID, m.metadata[vbID])
+	m.metadata[vbID].VbUUID = rollbackVbuuid
+	m.metadata[vbID].StartSeqNo = startSeqNo
+	m.metadata[vbID].SnapStartSeqNo = startSeqNo
+	m.metadata[vbID].SnapEndSeqNo = startSeqNo
```
adamcfraser: We're no longer setting endSeqNo here - that seems like it would still be needed for a one-shot feed?

adamcfraser (changes requested):
One question about endSeqNo handling on rollback.

base/dcp_client_metadata.go
torcolvin: We are just keeping the values we used to have here.

adamcfraser (dismissed):

adamcfraser (approved):

PR #6172, opened 2023-04-03
CBG-2798: replication rebalance fail

gregns1:
CBG-2798

Changed the behaviour where we set the replication status to target status in the event of a replication not getting assigned a node to run on. So added some logging to indicate this has happened then I changed it to error state with assigned node is empty string. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1642/


db/sg_replicate_cfg.go
```diff
@@ -1217,6 +1219,11 @@ func (c *SGRCluster) RebalanceReplications() {
 
 	// If there aren't any nodes available, there's nothing more to be done
 	if len(c.Nodes) == 0 {
+		if len(unassignedReplications) > 0 {
+			for _, v := range unassignedReplications {
+				base.WarnfCtx(c.loggingCtx, "Replication %s does not have an assigned node.", v.ID)
+			}
+		}
```
bbrks: nit: we can range over an empty (or even nil) slice/map to avoid the length check

```suggestion
		for _, v := range unassignedReplications {
			base.WarnfCtx(c.loggingCtx, "Replication %s does not have an assigned node.", v.ID)
		}
```

db/sg_replicate_cfg.go
```diff
@@ -995,7 +995,9 @@ func (m *sgReplicateManager) GetReplication(replicationID string) (*ReplicationC
 	}
 
 	// TODO: remove the local/non-local handling below when CBG-909 is completed
-	if replication.AssignedNode == m.localNodeUUID {
+	if replication.AssignedNode == "" {
+		replication.AssignedNode = ""
+	} else if replication.AssignedNode == m.localNodeUUID {
```
bbrks: I'd suggest either an early return to avoid the reassignment to the same value.

```suggestion
	// unassigned replication
	if replication.AssignedNode == "" {
		return replication, nil
	}

	if replication.AssignedNode == m.localNodeUUID {
```

db/sg_replicate_cfg.go
```diff
@@ -1396,9 +1403,16 @@ func (m *sgReplicateManager) GetReplicationStatus(replicationID string, options
 			if err != nil {
 				return nil, err
 			}
-			status = &ReplicationStatus{
-				ID:     replicationID,
-				Status: remoteCfg.TargetState,
+			if remoteCfg.AssignedNode == "" {
+				status = &ReplicationStatus{
+					ID:     replicationID,
+					Status: ReplicationStateError,
```
bbrks: I'm wondering if it's worth considering an extra state (`ReplicationStateUnassigned`) for this case?

Presumably the error is recoverable when a node running sg-replicate becomes available, which I think would differ from existing error states.

bbrks (changes requested):
Some thoughts about this - notably trying to not overload ReplicationStatusError for this scenario.

bbrks (approved):
LGTM

PR #6173, opened 2023-04-04
CBG-1971: TestReplicationConcurrentPush flaky test fix

gregns1:
CBG-1971

Test was flaking on waiting for nodes to be assigned. But also did a few runs where that wasn't included and notcied that it also failed on asserting for the number of docs pushed on the second replicator (rep_DEF). After digging into the losg I noticed that in all cases where it failed for this reaon I saw a log line:
`[INF] Replicate: t:TestReplicationConcurrentPush c:sgr-mgr- db:activedb Initializing replication <ud>rep_ABC</ud>`
But no corresponding log line for the second created replications (rep_DEF). When looking into the code I noticed this is called when refreshing the sgreplicate config. After adding new print lines everywhere I noticed on failure scenarios the reason the test wasn't actually initialising the second replication was due to `StartReplications` being called after the first replication was created and inserted into sgreplicate. This function created a new suscription and in this time the second replication config is inserted into sgreplicate. There is a race where this second replication is missed as a new addition as a new channel is created in subscribe replicate. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1648/ (for running particular tests 200 times) 
- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1650/   (full inegration test run)

### DIFF
--- a/db/sg_replicate_cfg.go
+++ b/db/sg_replicate_cfg.go
@@ -1194,6 +1194,10 @@ func (c *SGRCluster) GetReplicationIDsForNode(nodeUUID string) (replicationIDs [
 	return replicationIDs
 }
 
+func (m *sgReplicateManager) GetActiveReplicators() map[string]*ActiveReplicator {
+	return m.activeReplicators
+}
+
 // RebalanceReplications distributes the set of defined replications across the set of available nodes
 func (c *SGRCluster) RebalanceReplications() {
 
--- a/db/sg_replicate_cfg.go
+++ b/db/sg_replicate_cfg.go
@@ -802,6 +802,7 @@ func (m *sgReplicateManager) RefreshReplicationCfg(ctx context.Context) error {
 
 	// Check for replications newly assigned to this node
 	for replicationID, replicationCfg := range configReplications {
+		fmt.Println("newly assigned code", replicationID)
 		if replicationCfg.AssignedNode == m.localNodeUUID {
 			replicator, exists := m.activeReplicators[replicationID]
 			if !exists {
--- a/db/sg_replicate_cfg.go
+++ b/db/sg_replicate_cfg.go
@@ -1194,8 +1194,10 @@ func (c *SGRCluster) GetReplicationIDsForNode(nodeUUID string) (replicationIDs [
 	return replicationIDs
 }
 
-func (m *sgReplicateManager) GetActiveReplicators() map[string]*ActiveReplicator {
-	return m.activeReplicators
+func (m *sgReplicateManager) GetNumberActiveReplicators() int {
+	m.activeReplicatorsLock.Lock()
+	defer m.activeReplicatorsLock.Unlock()
+	return len(m.activeReplicators)
 }
 
 // RebalanceReplications distributes the set of defined replications across the set of available nodes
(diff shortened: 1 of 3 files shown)


db/sg_replicate_cfg.go
```diff
@@ -1191,6 +1191,10 @@ func (c *SGRCluster) GetReplicationIDsForNode(nodeUUID string) (replicationIDs [
 	return replicationIDs
 }
 
+func (m *sgReplicateManager) GetActiveReplicators() map[string]*ActiveReplicator {
+	return m.activeReplicators
```
bbrks: I think we need a lock to read values from this map (`m.activeReplicatorsLock`) and given the function returns the full map, we have no control over the locking.

We'll either need to make a full copy of the map, or given the use case, maybe a method specifically for retrieving `NumActiveReplicators()` which holds the lock whilst calling `len()`

rest/replicatortest/replicator_test.go
```diff
@@ -935,12 +935,14 @@ func TestReplicationConcurrentPush(t *testing.T) {
 
 	activeRT, remoteRT, remoteURLString, teardown := rest.SetupSGRPeers(t)
 	defer teardown()
-	// Create push replications, verify running
+	// Create push replication, wait for assigned replication (this gives time for the first replication to be created and start initializing),
+	// verify running, then create second replication, verify running and verify both replicators are initialized on sgReplicateManager
 	activeRT.CreateReplication("rep_ABC", remoteURLString, db.ActiveReplicatorTypePush, []string{"ABC"}, true, db.ConflictResolverDefault)
-	activeRT.CreateReplication("rep_DEF", remoteURLString, db.ActiveReplicatorTypePush, []string{"DEF"}, true, db.ConflictResolverDefault)
```
bbrks: Is there actually a bug here where we're not able to create two replications at the same time?

I'd expect the CAS retry in `sgReplicateManager.updateCluster` ensures that both could be created, even if concurrently, as it's something the REST API could potentially allow to happen in the real world.

bbrks (changes requested):
I'm not convinced that we're not just avoiding a real-world race condition here that we'd like to handle better (with CAS retry/update callback)

rest/replicatortest/replicator_test.go
gregns1: So my analysis of the problem (I could be wrong) was that we can create two at the same time, but the changes to sg replicate config come through on a channel `cfgEvents` and we have a race where sometimes where only the first replication is picked up and takes time to initialise. Then by the time the second replication is processed as a cfg change and goes through the refresh config/initialise replication process the test is already asserting against the stats for that replication. I think in real world for customers the replication will initialise eventually. Also think that this test isn't really a problem when running against server I think its more flaking for walrus and I put that down to the added latency when communicating with server helping in the context of the test?

db/sg_replicate_cfg.go
gregns1: Good point will change this!

rest/replicatortest/replicator_test.go
bbrks: The thing that makes me uncertain of that is that we don't see any logging associated with `rep_DEF` in the entire test (which runs for ~20 seconds)

https://mobile.jenkins.couchbase.com/job/sgw-unix-build/26219/consoleFull 

It's almost like DEF never even gets created (or at least assigned to a node)

bbrks (approved):
I'm OK with this fix after in-person discussion.

PR #6174, opened 2023-04-06
Add 3.1 release manifest

torcolvin:
Should I also add a file for 

manifest/3.1/3.1.0.xml with duplicate contents of 3.1.xml or leave that to the inevitable 3.1.1 point release?

I was going to merge this change into the release branch AND master.

### DIFF
--- a/manifest/3.1.xml
+++ b/manifest/3.1.xml
@@ -12,7 +12,7 @@ licenses/APL2.txt.
 
 <manifest>
 
-  <remote fetch="https://github.com/couchbase/" name="couchbase"/>
+  <remote fetch="https://github.com/couchbase/" name="couchbase" revision="release/3.1.0"/>
   <default remote="couchbase" revision="master"/>
 
   <!-- Build Scripts (required on CI servers) -->
--- a/manifest/3.1.xml
+++ b/manifest/3.1.xml
@@ -12,7 +12,7 @@ licenses/APL2.txt.
 
 <manifest>
 
-  <remote fetch="https://github.com/couchbase/" name="couchbase" revision="release/3.1.0"/>
+  <remote fetch="https://github.com/couchbase/" name="couchbase"/>
   <default remote="couchbase" revision="master"/>
 
   <!-- Build Scripts (required on CI servers) -->
@@ -25,6 +25,6 @@ licenses/APL2.txt.
 
 
   <!-- Sync Gateway -->
-   <project name="sync_gateway" path="sync_gateway" remote="couchbase" />
+   <project name="sync_gateway" path="sync_gateway" remote="couchbase" revision="release/3.1.0"/>
 
 </manifest>
--- a/manifest/3.1.xml
+++ b/manifest/3.1.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+Copyright 2016-Present Couchbase, Inc.
+
+Use of this software is governed by the Business Source License included in
+the file licenses/BSL-Couchbase.txt.  As of the Change Date specified in that
+file, in accordance with the Business Source License, use of this software will
+be governed by the Apache License, Version 2.0, included in the file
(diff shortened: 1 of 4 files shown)


bbrks:
- `3.1.0.xml` comes only when we have a new `3.1.x` release (otherwise you'd end up with 2 `3.1.0` manifests)
- Nothing reads the branch version of `product-config.json` - `master` is the source of truth for what versions to build. Changing this isn't nessesary.
- We can get the `3.1.0.xml` into the release branch for our weird Jenkins jobs that depend on the release branch - but the build team's jobs read the `master` version.

torcolvin:
> * Nothing reads the branch version of `product-config.json` - `master` is the source of truth for what versions to build. Changing this isn't nessesary.

Do you prefer to have release branch match master? Is it correct in theory that none of this needs to be done on the release branch?

manifest/3.1.xml
```diff
@@ -0,0 +1,30 @@
...
+  <project name="product-texts" path="product-texts" remote="couchbase"/>
+  <project name="build" path="cbbuild" remote="couchbase" revision="844e2484067053d7af88e706a814766265bf03df">
+    <annotation name="VERSION" value="3.1.0"     keep="true"/>
+    <annotation name="BLD_NUM" value="@BLD_NUM@" keep="true"/>
+    <annotation name="RELEASE" value="@RELEASE@" keep="true"/>
+  </project>
+
+
+  <!-- Sync Gateway -->
+   <project name="sync_gateway" path="sync_gateway" remote="couchbase" />
```
bbrks: We need to let them (the build team's job) know where to find 3.1.0 in SG's repo here:

```suggestion
   <project name="sync_gateway" path="sync_gateway" remote="couchbase" revision="release/3.1.0"/>
```

manifest/product-config.json
bbrks, line 1: The `manifest/default.xml` entry (starting line 4) needs updating to `"release": "3.2.0",` now

manifest/default.xml
bbrks, line 1: Also need to make this change in `base/version.go` (`ProductAPIVersionMinor = "2"`)

manifest/product-config.json
```diff
@@ -389,6 +389,15 @@
...
             "start_build": 1
         },
+        "manifest/3.1.xml": {
+            "release": "3.1.0",
+            "release_name": "Couchbase Sync Gateway 3.1.0",
+            "production": true,
+            "interval": 120,
+            "go_version": "1.19.5",
+            "trigger_blackduck": true,
+            "start_build": 1
```
bbrks: `"start_build"` needs to be `+1` of the current build numebr of 3.1.0 to ensure any subsequent builds get assigned the next build number.

At time of writing the latest build is `585`: 
http://latestbuilds.service.couchbase.com/builds/latestbuilds/sync_gateway/3.1.0/

bbrks (changes requested):
A couple of things (these are in the doc)

manifest/3.1.xml
bbrks: this wasn't fixed

manifest/3.1.xml
```diff
@@ -0,0 +1,30 @@
...
+Use of this software is governed by the Business Source License included in
+the file licenses/BSL-Couchbase.txt.  As of the Change Date specified in that
+file, in accordance with the Business Source License, use of this software will
+be governed by the Apache License, Version 2.0, included in the file
+licenses/APL2.txt.
+-->
+
+<manifest>
+
+  <remote fetch="https://github.com/couchbase/" name="couchbase" revision="release/3.1.0"/>
```
bbrks: remotes don't have revisions

```suggestion
  <remote fetch="https://github.com/couchbase/" name="couchbase"/>
```

bbrks (changes requested):
the revision was added in the wrong place in the manifest

bbrks (approved):

PR #6176, opened 2023-04-11
CBG-2711 add 3 node cluster option to jenkins

torcolvin:
create ./integration-test/start_server.sh which allows you to start a cluster in a multinode or a single node setup. The intent is that SG developers could use this to set up clusters but I think I need another PR for this to work on OS X.

When creating a 3 node cluster (MULTI_NODE on jenkins), it will create each bucket with 1 replica.

Questions:

- Can we change the AMI to include a modern version of docker? This is one of those things that won't work on OS X because this uses docker compose v1 (written in python) and modern docker uses docker compose v2. This can be put off until we update the AMI. I also add jq at runtime if `JENKINS_URL` is set.
- Why do we set `--volume ${DOCKER_CBS_ROOT_DIR}/cbs:/root --volume ${WORKSPACE_ROOT}:/workspace` ? It seems like it is easier to use docker volumes and just do a `docker exec -t couchbase /opt/couchbase/bin/cbcollect_info /workspace/cbcollect.zip && docker cp couchbase:/workspace/cbcollect.zip cbcollect.zip` to get the file. The only reason to mount the disk seems like would be if we pick up partial files from checks. I can't figure out why we bind mount the `/root` volume at all.

Ran 6.6.5,7.1.4,7.0.3 with MULTI_NODE on and off for a base DCP test.

### DIFF
--- a/base/main_test_bucket_pool_config.go
+++ b/base/main_test_bucket_pool_config.go
@@ -54,6 +54,9 @@ const (
 
 	// wait this long when requesting a test bucket from the pool before giving up and failing the test.
 	waitForReadyBucketTimeout = time.Minute
+
+	// Creates buckets with a specific number of number of replicas
+	tbpEnvBucketNumReplicas = "SG_TEST_BUCKET_NUM_REPLICAS"
 )
 
 var tbpDefaultBucketSpec = BucketSpec{
@@ -134,6 +137,19 @@ func tbpNumBuckets() int {
 		}
 	}
 	return numBuckets
+}
+
+// tbpNumReplicasreturns the number of replicas to use in each bucket.
+func tbpNumReplicas() uint32 {
+	numReplicas := os.Getenv(tbpEnvBucketPoolSize)
+	if numReplicas == "" {
+		return 0
+	}
+	replicas, err := strconv.Atoi(numReplicas)
+	if err != nil {
+		FatalfCtx(context.TODO(), "Couldn't parse %s: %v", tbpEnvBucketPoolSize, err)
+	}
+	return uint32(replicas)
 }
 
 // tbpNumCollectionsPerBucket returns the configured number of collections prepared in a bucket.
--- a/base/main_test_bucket_pool_config.go
+++ b/base/main_test_bucket_pool_config.go
@@ -141,7 +141,7 @@ func tbpNumBuckets() int {
 
 // tbpNumReplicasreturns the number of replicas to use in each bucket.
 func tbpNumReplicas() uint32 {
-	numReplicas := os.Getenv(tbpEnvBucketPoolSize)
+	numReplicas := os.Getenv(tbpEnvBucketNumReplicas)
 	if numReplicas == "" {
 		return 0
 	}
--- a/jenkins-integration-build.sh
+++ b/jenkins-integration-build.sh
@@ -62,6 +62,13 @@ else
     go install -v github.com/AlekSi/gocov-xml@latest
 fi
 
+if [[ -n "${JENKINS_URL}" ]]; then
+    # last 1.x version, when updating aws linux 2 docker, docker-compose becomes docker compose
+    sudo yum install -y jq
+    sudo curl -L "https://github.com/docker/compose/releases/download/1.29.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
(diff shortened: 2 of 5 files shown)


base/main_test_bucket_pool_config.go
```diff
@@ -136,6 +139,19 @@ func tbpNumBuckets() int {
 	return numBuckets
 }
 
+// tbpNumReplicasreturns the number of replicas to use in each bucket.
+func tbpNumReplicas() uint32 {
+	numReplicas := os.Getenv(tbpEnvBucketPoolSize)
```
bbrks: Tying `numReplicas` to bucket pool size seems odd (valid range for repliacs is [only `0-3`](https://docs.couchbase.com/server/current/learn/clusters-and-availability/replication-architecture.html))

If anything we'd want it to be equal to the number of CB nodes, or a fixed value and let CB decide how many replicas can be created based on cluster.

jenkins-integration-build.sh
```diff
@@ -153,7 +141,7 @@ if [ "${SG_EDITION}" == "EE" ]; then
     GO_TEST_FLAGS="${GO_TEST_FLAGS} -tags cb_sg_enterprise"
 fi
 
-go test ${GO_TEST_FLAGS} -coverprofile=coverage_int.out -coverpkg=github.com/couchbase/sync_gateway/... github.com/couchbase/sync_gateway/${TARGET_PACKAGE} 2>&1 | stdbuf -oL tee "${INT_LOG_FILE_NAME}.out.raw" | stdbuf -oL grep -a -E '(--- (FAIL|PASS|SKIP):|github.com/couchbase/sync_gateway(/.+)?\t|TEST: |panic: )'
+go test ${GO_TEST_FLAGS} -coverprofile=coverage_int.out -coverpkg=github.com/couchbase/sync_gateway/... github.com/couchbase/sync_gateway/${TARGET_PACKAGE} 2>&1 | stdbuf -oL tee "${INT_LOG_FILE_NAME}.out.raw" | stdbuf -oL grep -a -P '(--- (FAIL|PASS|SKIP):|github.com/couchbase/sync_gateway(/.+)?\t|TEST: |panic: )'
```
bbrks: BSD grep (used by macOS) doesn't have `-P`

I was under the impression GNU grep supported both. Was this change nessesary on your environment?

jenkins-integration-build.sh
```diff
@@ -112,34 +119,15 @@ if [ "${RUN_WALRUS}" == "true" ]; then
...
-DOCKER_CBS_ROOT_DIR="$(pwd)"
-if [ "${CBS_ROOT_DIR:-}" != "" ]; then
-    DOCKER_CBS_ROOT_DIR="${CBS_ROOT_DIR}"
+# Run CBS
+if [[ -z ${MULTI_NODE:-} ]]; then
+    # Run CBS
+    ./integration-test/start_server.sh "${COUCHBASE_SERVER_VERSION}"
+else
+    ./integration-test/start_server.sh -m "${COUCHBASE_SERVER_VERSION}"
+    export SG_TEST_BUCKET_NUM_REPLICAS=1
```
bbrks, line 128: Num replicas here should be zero, right?

bbrks (changes requested):

base/main_test_bucket_pool_config.go
torcolvin: whoops, mistake here, meant to use the env var

jenkins-integration-build.sh
torcolvin: this was just to get rid of an annoying error on linux:

```
grep: warning: stray \ before t
```

I'll pull it out of this PR

jenkins-integration-build.sh
torcolvin: I think it should be 1 because this is multi-node cluster and we want at least one replica for each bucket?

bbrks: My bad - read the if backwards.

Should we specify 2 replicas in that case, if we know there are 3 CB Data nodes?

jenkins-integration-build.sh
bbrks: Hmm, that's supposed to be matching a tab character.
Maybe single quotes need to be double quotes?

bbrks (approved):

PR #6177, opened 2023-04-11
CBG-2686: fix for dcp resync flake

gregns1:
CBG-2686

We were asseerting on docs prcessed on dcp resync. I think this was causing some flaking results. Due to nature of dcp it's hard to assert on what you see over dcp so I have changed the assertions to be at least the number of docs added in the test case and thus have to change assertion on sync function runs stats too. This will behind the `isDCPResync` flag and the original, assertion will remain there for non dcp runs. 
I have then added a check in the code to get status doc from the cluster. We were originally just setting sttaus to stopped if heartbeat isnt found. It is entirely possible that the status is completed and thus the heartbeat doc won't be there. So if the status is completed we cant it to remain this way indicating the background process has been successful. 

* Took a slightly different apprioach as it still flaked on git hub actions once. I think this now fixes the issue, I have re run jenkins github actions a few times now with no flake for this test. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1689/ (Run dcp resync 200 times)
- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1690/ (Run non dcp resync 200 times against server)

### DIFF
--- a/db/background_mgr.go
+++ b/db/background_mgr.go
@@ -307,7 +307,7 @@ func (b *BackgroundManager) getStatusFromCluster() ([]byte, error) {
 		_, _, err = b.clusterAwareOptions.metadataStore.GetRaw(b.clusterAwareOptions.HeartbeatDocID())
 		if err != nil {
 			if base.IsDocNotFoundError(err) {
-				if clusterStatus["status"] != BackgroundProcessStateCompleted {
+				if clusterState != string(BackgroundProcessStateCompleted) {
 					clusterStatus["status"] = BackgroundProcessStateStopped
 					status, err = base.JSONMarshal(clusterStatus)
 					if err != nil {
--- a/db/background_mgr.go
+++ b/db/background_mgr.go
@@ -307,7 +307,12 @@ func (b *BackgroundManager) getStatusFromCluster() ([]byte, error) {
 		_, _, err = b.clusterAwareOptions.metadataStore.GetRaw(b.clusterAwareOptions.HeartbeatDocID())
 		if err != nil {
 			if base.IsDocNotFoundError(err) {
-				if clusterState != string(BackgroundProcessStateCompleted) {
+				if clusterState == string(BackgroundProcessStateRunning) {
+					status, _, err = b.getStatusLocal()
+					if err != nil {
+						return nil, err
+					}
+				} else {
 					clusterStatus["status"] = BackgroundProcessStateStopped
 					status, err = base.JSONMarshal(clusterStatus)
 					if err != nil {
--- a/db/background_mgr.go
+++ b/db/background_mgr.go
@@ -307,10 +307,12 @@ func (b *BackgroundManager) getStatusFromCluster() ([]byte, error) {
 		_, _, err = b.clusterAwareOptions.metadataStore.GetRaw(b.clusterAwareOptions.HeartbeatDocID())
 		if err != nil {
 			if base.IsDocNotFoundError(err) {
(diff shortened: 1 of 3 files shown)


rest/sync_fn_test.go
```diff
@@ -603,19 +603,21 @@ func TestResync(t *testing.T) {
 			})
 			assert.NoError(t, err)
 
-			assert.Equal(t, testCase.expectedSyncFnRuns, int(rt.GetDatabase().DbStats.Database().SyncFunctionCount.Value()))
-
 			if !isDCPResync {
 				var queryName string
+				assert.Equal(t, testCase.expectedSyncFnRuns, int(rt.GetDatabase().DbStats.Database().SyncFunctionCount.Value()))
```
torcolvin: These assertions are all dead - it is _always_ DCP resync unless `UseQueryBasedResyncManager` is set.

rest/sync_fn_test.go
```diff
@@ -512,7 +512,10 @@ func TestDBOfflineSingleResync(t *testing.T) {
 	assert.Equal(t, int64(2000), rt.GetDatabase().DbStats.Database().SyncFunctionCount.Value())
 }
 
-func TestResync(t *testing.T) {
+func TestResyncOverDCP(t *testing.T) {
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("This test doesn't works with walrus")
```
torcolvin:
```suggestion
                t.Skip("DCP requires gocb and CBS")
```

db/background_mgr.go
```diff
@@ -307,10 +307,12 @@ func (b *BackgroundManager) getStatusFromCluster() ([]byte, error) {
 		_, _, err = b.clusterAwareOptions.metadataStore.GetRaw(b.clusterAwareOptions.HeartbeatDocID())
 		if err != nil {
 			if base.IsDocNotFoundError(err) {
-				clusterStatus["status"] = BackgroundProcessStateStopped
-				status, err = base.JSONMarshal(clusterStatus)
-				if err != nil {
-					return nil, err
+				if clusterStatus["status"] != BackgroundProcessStateCompleted {
```
torcolvin:
```suggestion
				if clusterState != BackgroundProcessStateCompleted
```

rest/sync_fn_test.go
```diff
@@ -603,21 +601,8 @@ func TestResync(t *testing.T) {
 			})
 			assert.NoError(t, err)
```
torcolvin:
```suggestion
			require.NoError(t, err)
```

torcolvin (commented):
Just a few small nitpicks for readability in the future.

torcolvin:
I am OK with this change and if it fixes GH actions run a sufficient number of times that's OK. I'd like to see a test that works solely on Background Manager where it starts/stops the goroutines quickly since this fix is BackgroundManager specific.

But if that's hard to reproduce, I'm OK with merging this.

gregns1:
> I am OK with this change and if it fixes GH actions run a sufficient number of times that's OK. I'd like to see a test that works solely on Background Manager where it starts/stops the goroutines quickly since this fix is BackgroundManager specific.
> 
> But if that's hard to reproduce, I'm OK with merging this.

Added temporary change to run the test 100 times in gh actions and no failiure. Also re ran the tests in the jenkins bit in actions and no failiures. Think this fixes the issue but will reopen the ticket if we see it again.

torcolvin (approved):

PR #5809, opened 2022-10-07
CBG-2424: Allow databases to override CORS config

snej:
CBG-2424

Allows a database to specify its own CORS configuration, overriding the server default.

- Added "cors" property to DbConfig (same schema as in server config.) If present, it overrides the server config for that database.
- Moved CORSConfig type and matchedOrigin func to auth package, to enable...
- Added a *CORSConfig to DatabaseContext struct.
- wrapRouter() fn no longer adds CORS headers to all requests.
- Instead, handler.invoke adds the headers as soon as it figures out which database (or none) the request targets.
- Added a unit test to make sure a per-db CORS config takes effect for requests for that db, but not for other requests.

## Pre-review checklist
- [x] Removed debug logging (`fmt.Print`, `log.Print`, ...)
- [x] Logging sensitive data? Make sure it's tagged (e.g. `base.UD(docID)`, `base.MD(dbName)`)
- [x] Updated relevant information in the API specifications (such as endpoint descriptions, schemas, ...) in `docs/api`

## Dependencies (if applicable)
- [ ] Link upstream PRs
- [ ] Update Go module dependencies when merged

## [Integration Tests](https://jenkins.sgwdev.com/job/SyncGateway-Integration/build?delay=0sec)
- [ ] `xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/000/
- [ ] `GSI=true,xattrs=true` (for collections) https://jenkins.sgwdev.com/job/SyncGateway-Integration/000/

PR #6179, opened 2023-04-11
CBG-2807: Allow databases to override CORS config

torcolvin:
(copied from https://github.com/couchbase/sync_gateway/pull/5809)
CBG-2424

Allows a database to specify its own CORS configuration, overriding the server default.

- Added "cors" property to DbConfig (same schema as in server config.) If present, it overrides the server config for that database.
- Moved CORSConfig type and matchedOrigin func to auth package, to enable...
- Added a *CORSConfig to DatabaseContext struct.
- wrapRouter() fn no longer adds CORS headers to all requests.
- Instead, handler.invoke adds the headers as soon as it figures out which database (or none) the request targets.
- Added a unit test to make sure a per-db CORS config takes effect for requests for that db, but not for other requests.

Possible additions for testing:

- dynamically change CORS in with an updated config
- test base CORS in bootstrap config with override in server config?
- multiple databases on a rest tester with different CORS

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration-1.19.5/5/


rest/handler.go
```diff
@@ -219,21 +219,45 @@ func ParseKeyspace(ks string) (db string, scope, collection *string, err error)
...
 
+	err := h._validateAndWriteHeaders(method, accessPermissions, responsePermissions)
+	if err != nil {
+		return err
+	}
+	return method(h) // Call the actual handler code
+}
+
+// _validateAndWriteHeaders sets up handler.db and validates the permission of the user and returns an error if there is not permission.
+func (h *handler) _validateAndWriteHeaders(method handlerMethod, accessPermissions []Permission, responsePermissions []Permission) error {
```
adamcfraser: We typically reserve the leading underscore for methods that require the caller to hold a mutex.  That doesn't seem to be the case here.

rest/handler.go
```diff
@@ -219,21 +219,45 @@ func ParseKeyspace(ks string) (db string, scope, collection *string, err error)
...
 	var isRequestLogged bool
 	defer func() {
 		if !isRequestLogged {
 			h.logRequestLine()
 		}
 	}()
 
+	var dbContext *db.DatabaseContext
+	defer func() {
+		// Now that we know the DB, add CORS headers to the response:
```
adamcfraser, line 250: Should we bypass this functionality if err is non-nil?  Wouldn't adding the CORS headers in the error case leak information about db existence if auth fails below?  (tests for this case would be recommended)

adamcfraser (changes requested):

rest/handler.go
torcolvin: nope, I think this is OK based on testing.

adamcfraser: Which test covers this case?  The case I'm thinking of is:
- database 'abc' has db-specific CORS config enabled
- database 'def' does not exist
- user bob doesn't have access to abc
- if user bob makes a request to abc, doesn't he get CORS response headers that he does not see for the same request to def?

rest/api_test.go
```diff
@@ -230,45 +230,113 @@ func TestFunkyDocIDs(t *testing.T) {
...
+		},
+	})
+	defer rt.Close()
+
+	reqHeaders := map[string]string{
+		"Origin": "http://couchbase.com",
 	}
-	response = rt.SendRequestWithHeaders("GET", "/{{.keyspace}}/", "", reqHeaders)
-	assert.Equal(t, "http://staging.example.com", response.Header().Get("Access-Control-Allow-Origin"))
+	response := rt.SendRequestWithHeaders("GET", "/db/", "", reqHeaders)
```
adamcfraser: Why is this 'db' and not {{.db}}? If there's a reason, maybe stick a comment in the test.

rest/routing.go
```diff
@@ -388,8 +384,8 @@ func wrapRouter(sc *ServerContext, privs handlerPrivs, router *mux.Router) http.
 				h.writeStatus(http.StatusNotFound, "unknown URL")
 			} else {
 				response.Header().Add("Allow", strings.Join(options, ", "))
-				if privs != adminPrivs && sc.Config.API.CORS != nil && len(originHeader) > 0 {
-					response.Header().Add("Access-Control-Max-Age", strconv.Itoa(sc.Config.API.CORS.MaxAge))
+				if privs != adminPrivs && cors != nil && len(rq.Header["Origin"]) > 0 {
+					response.Header().Add("Access-Control-Max-Age", strconv.Itoa(cors.MaxAge))
```
adamcfraser, line 389: This looks like it's always going to be the server-config level setting, and doesn't get updated later when we call AddResponseHeaders in invoke.  Is that the intended behaviour?

adamcfraser (changes requested):

rest/routing.go
torcolvin: Added `TestCorsNoMux` for this, it's intended behavior for when we don't match anything on `mux.Router`.

rest/api_test.go
torcolvin: Code was written before the templating existed, I added it here.

rest/routing.go
adamcfraser: My question was for the case where we *do* match something on mux.Router, I don't think Access-Control-Max-Age gets updated to the db-specific value.  But I see what you mean now - that code is only in the non-matching path.  Let me go sanity check if/when that gets set elsewhere.

adamcfraser (approved):

adamcfraser (changes requested):
It looks like TestCORSOriginPerDatabase is failing with the latest changes, since it attempts to make requests as the GUEST user, and GUEST is disabled (so it's falling back to the server-level CORS config).

rest/routing.go
torcolvin: I wrote some validation so you can only set Max-Age on bootstrap to clarify this.

rest/handler.go
```diff
@@ -223,25 +223,47 @@ func ParseKeyspace(ks string) (db string, scope, collection *string, err error)
...
 	var isRequestLogged bool
 	defer func() {
 		if !isRequestLogged {
 			h.logRequestLine()
 		}
 	}()
 
+	defer func() {
+		// Now that we know the DB, add CORS headers to the response:
+		if h.privs != adminPrivs {
```
adamcfraser: Should this include `&& h.privs != metricPrivs`?  Usage seems inconsistent below (existing line 369 below looks like it should be doing the same), but I think it's a no-op there because dbcontext will always be nil for metrics calls.  For this case, it feels like we shouldn't be setting CORS (even the server config CORS) on the metrics endpoint.

docs/api/components/schemas.yaml
```diff
@@ -1744,6 +1744,28 @@ Database:
...
+          description: List of allowed login origins
+          type: array
+          items:
+            type: string
+        headers:
+          description: List of allowed headers
+          type: array
+          items:
+            type: string
+        max_age:
```
adamcfraser: Can we remove this from the API definition, since we don't support it being set at the database level?

adamcfraser (changes requested):
Fixes look good - two minor questions.

adamcfraser (approved):

PR #6183, opened 2023-04-12
CBG-2842 Always return unauthorized for non authorized public API

torcolvin:
## Pre-review checklist

- [] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration-1.19.5/4/


rest/api_test.go
```diff
@@ -217,29 +217,33 @@ func TestCORSOrigin(t *testing.T) {
 	reqHeaders := map[string]string{
 		"Origin": "http://example.com",
 	}
-	response := rt.SendRequestWithHeaders("GET", "/{{.keyspace}}/", "", reqHeaders)
+	response := rt.SendRequestWithHeaders("GET", "/{{.db}}/", "", reqHeaders)
+	RequireStatus(t, response, http.StatusUnauthorized)
```
adamcfraser: Were these tests previously returning these statuses, or does that change in this PR?

adamcfraser: Were these tests previously returning these statuses, or does that change in this PR?

rest/user_api_test.go
```diff
@@ -1520,3 +1520,56 @@ func TestPutUserCollectionAccess(t *testing.T) {
...
+		PersistentConfig: true,
+	})
+	defer rt.Close()
+
+	dbConfig := rt.NewDbConfig()
+
+	dbName := "realdb"
+	rt.CreateDatabase("realdb", dbConfig)
+
+	response := rt.SendRequest(http.MethodGet, "/{{.db}}/", "")
```
adamcfraser: Does this resolve to "realdb"?  Is there an advantage to using the token here - to me it seems to obscure what's actually being tested.

rest/handler.go
```diff
@@ -688,6 +695,11 @@ func checkJWTIssuerStillValid(ctx context.Context, dbCtx *db.DatabaseContext, us
 // checkAdminAuthenticationOnly simply checks whether a username / password combination is authenticated pulling the
 // credentials from the handler
 func (h *handler) checkAdminAuthenticationOnly() (bool, error) {
+	if h.privs == adminPrivs && !*h.server.Config.API.AdminInterfaceAuthentication {
```
adamcfraser: Is this change required?  This looks like it's already being evaluated by the called (if shouldCheckAdminAuth).

adamcfraser (commented):
The PR description describes this as handling the admin case, but the associated ticket (CBG-2842) is for public access.  Which case is this fixing?

rest/api_test.go
torcolvin: If you ask `{{.keyspace}}` it will return 400, which is how keyspace is resolved than db differently. I could leave this and test that it returns 400. 

This change is only a test clarification, I wanted to assert I wasn't changing the error code, but I can leave this out of this initial PR.

I added this code to make sure nothing was changed.

torcolvin: Yes, using the same.

rest/user_api_test.go
torcolvin: yeah, fixing.

torcolvin:
> The PR description describes this as handling the admin case, but the associated ticket ([CBG-2842](https://issues.couchbase.com/browse/CBG-2842)) is for public access. Which case is this fixing?

This was just a mistake of a message, this fixes public access and doesn't regress on admin access.

rest/handler.go
torcolvin: nope

rest/api_test.go
```diff
@@ -114,11 +120,22 @@ func TestDisablePublicBasicAuth(t *testing.T) {
...
+
+	response = rt.SendUserRequest(http.MethodGet, "/notadb/", "", "user1")
+	RequireStatus(t, response, http.StatusUnauthorized)
+	require.Contains(t, response.Body.String(), ErrInvalidLogin.Message)
 	assert.NotContains(t, response.Header(), "WWW-Authenticate", "expected to not receive a WWW-Auth header when password auth is disabled")
 
 	// Also check that we can't create a session through POST /db/_session
 	response = rt.SendRequest(http.MethodPost, "/{{.db}}/_session", `{"name":"user1","password":"letmein"}`)
 	RequireStatus(t, response, http.StatusUnauthorized)
+	require.Contains(t, response.Body.String(), ErrLoginRequired.Message)
```
adamcfraser, line 134: I'm not sure whether this is testing what it wants to be testing - it looks like the "Login required" might be happening during authentication of the request (if it's trying to auth against the public port as GUEST).  I think the intention here is to test that DisablePublicBasicAuth prevents creating a session using username/password in the session body.  So we want the request to authenticate successfully, but the session creation to fail.

This looks like an existing test bug - maybe file a non-Helium ticket to correct this test?

rest/handler.go
```diff
@@ -1049,6 +1067,21 @@ func (h *handler) getBearerToken() string {
 	return ""
 }
 
+// providedAuthCredentials returns true if basic auth or session auth is enabled
+func (h *handler) providedAuthCredentionals() bool {
```
adamcfraser:
```suggestion
func (h *handler) providedAuthCredentials() bool {
```

rest/handler.go
```diff
@@ -1049,6 +1067,21 @@ func (h *handler) getBearerToken() string {
 	return ""
 }
 
+// providedAuthCredentials returns true if basic auth or session auth is enabled
```
adamcfraser, line 1070: This doesn't cover all the mechanisms by which auth could be provided (it's missing bearer JWT token).  The session cookie check is also only going to work in cases where a valid db has been found, which doesn't seem like it's necessarily the case for all callers.

rest/handler.go
```diff
@@ -648,7 +664,10 @@ func (h *handler) checkAuth(dbCtx *db.DatabaseContext) (err error) {
 		if dbCtx.Options.SendWWWAuthenticateHeader == nil || *dbCtx.Options.SendWWWAuthenticateHeader {
 			h.response.Header().Set("WWW-Authenticate", wwwAuthenticateHeader)
 		}
-		return base.HTTPErrorf(http.StatusUnauthorized, "Login required")
+		if h.providedAuthCredentionals() {
```
adamcfraser: It looks to me like we've already fallen through to GUEST access handling at this point, which I thought would imply that auth credentials haven't been provided.  Which scenario is this covering?

adamcfraser (changes requested):

rest/handler.go
torcolvin: It's not the case for all callers, which is why the check only works if it's the default cookie type if database is fine. I added bearer check though.

rest/handler.go
torcolvin: This is hit by 

```
        response = rt.SendUserRequest(http.MethodGet, "/{{.db}}/", "", "user1")
	RequireStatus(t, response, http.StatusUnauthorized)
	require.Contains(t, response.Body.String(), ErrInvalidLogin.Message)
```

from `TestDisablePublicBasicAuth`

rest/api_test.go
torcolvin: This is actually behavior I changed - 

Otherwise:

```response = rt.SendRequest(http.MethodPost, "/{{.db}}/_session", `{"name":"user1","password":"letmein"}`)```

and 

```response = rt.SendRequest(http.MethodPost, "/notadb/_session", `{"name":"user1","password":"letmein"}`)```

would return different errors. The change is in session_api.go

rest/handler.go
adamcfraser: I thought it would be preferable to do this check earlier in checkAuth (i.e. if DisablePasswordAuth is true, return ErrInvalidLogin if basic auth credentials are provided).  However, that would break the existing behaviour, where if guest is enabled AND DisablePasswordAuth is true, basic auth credentials are ignored and it falls through to GUEST.  I'm not convinced this is the desired behaviour, but it's not a security issue and we don't want to break existing handling.

adamcfraser (approved):

PR #6189, opened 2023-04-13
(3.0.6 backport) CBG-2842 Always return unauthorized for non authorized public API (#6183)

torcolvin:
CBG-2850

Backport of https://github.com/couchbase/sync_gateway/commit/e0520bcbe86c5d12a35dba8a77bc73f15e09fa6b

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration-go1.16.15/42/

### DIFF
--- a/rest/admin_api_auth_test.go
+++ b/rest/admin_api_auth_test.go
@@ -963,7 +963,6 @@ func TestAdminAPIAuth(t *testing.T) {
 			assertStatus(t, resp, http.StatusUnauthorized)
 
 			resp = rt.SendAdminRequestWithAuth(endPoint.Method, formattedEndpoint, body, "noaccess", "password")
-			require.Contains(t, resp.Body.String(), ErrLoginRequired.Message)
 			assertStatus(t, resp, http.StatusForbidden)
 
 			if !endPoint.SkipSuccessTest {
@@ -1573,7 +1572,7 @@ func TestLoginRequiredForAdmin(t *testing.T) {
 	username := "nonexistent"
 	password := "nope"
 
-	response := rt.SendAdminRequestWithAuth(http.MethodGet, "/{{.db}}/", "", username, password)
+	response := rt.SendAdminRequestWithAuth(http.MethodGet, "/db/", "", username, password)
 
 	assertStatus(t, response, http.StatusUnauthorized)
 	require.Contains(t, response.Body.String(), ErrInvalidLogin.Message)
--- a/rest/admin_api_auth_test.go
+++ b/rest/admin_api_auth_test.go
@@ -961,7 +961,7 @@ func TestAdminAPIAuth(t *testing.T) {
 		t.Run(endPoint.Method+formattedEndpoint, func(t *testing.T) {
 			resp := rt.SendAdminRequest(endPoint.Method, formattedEndpoint, body)
 			assertStatus(t, resp, http.StatusUnauthorized)
-
+			require.Contains(t, resp.Body.String(), ErrLoginRequired.Message)
 			resp = rt.SendAdminRequestWithAuth(endPoint.Method, formattedEndpoint, body, "noaccess", "password")
 			assertStatus(t, resp, http.StatusForbidden)
 
--- a/rest/admin_api_auth_test.go
+++ b/rest/admin_api_auth_test.go
@@ -508,6 +508,7 @@ func TestAdminAuthWithX509(t *testing.T) {
 	_, statusCode, err = checkAdminAuth("", "invalidUser", "invalidPassword", "", httpClient, managementEndpoints, true, []Permission{{"!admin", false}}, nil)
 	assert.NoError(t, err)
 	assert.Equal(t, http.StatusUnauthorized, statusCode)
+	require.Contains(t, err.Error(), ErrInvalidLogin.Message)
 }
 
 func TestAdminAPIAuth(t *testing.T) {
@@ -962,6 +963,7 @@ func TestAdminAPIAuth(t *testing.T) {
 			assertStatus(t, resp, http.StatusUnauthorized)
 
 			resp = rt.SendAdminRequestWithAuth(endPoint.Method, formattedEndpoint, body, "noaccess", "password")
+			require.Contains(t, resp.Body.String(), ErrLoginRequired.Message)
 			assertStatus(t, resp, http.StatusForbidden)
 
 			if !endPoint.SkipSuccessTest {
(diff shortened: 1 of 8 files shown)


rest/handler.go
```diff
@@ -209,17 +213,24 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
...
+					if !authorized || h.providedAuthCredentials() {
+						return ErrInvalidLogin
+					}
+				} else {
+					if h.privs == regularPrivs || h.privs == publicPrivs {
+						if !h.providedAuthCredentials() {
+
+							return ErrLoginRequired
+						}
+						return ErrInvalidLogin
```
bbrks: I'm not sure why this change is different than the one that went into `master` ?

https://github.com/couchbase/sync_gateway/pull/6183/files#diff-9899f6c02129d809f5816c71f72902fd77314b35b000f7484d4ba6ea2e5ca1ef

rest/session_api.go
bbrks, line 1: No `session_test.go` change for `TestSessionExtension` error message ?

rest/admin_api_auth_test.go
```diff
@@ -508,6 +508,7 @@ func TestAdminAuthWithX509(t *testing.T) {
 	_, statusCode, err = checkAdminAuth("", "invalidUser", "invalidPassword", "", httpClient, managementEndpoints, true, []Permission{{"!admin", false}}, nil)
 	assert.NoError(t, err)
 	assert.Equal(t, http.StatusUnauthorized, statusCode)
+	require.Contains(t, err.Error(), ErrInvalidLogin.Message)
 }
 
 func TestAdminAPIAuth(t *testing.T) {
```
bbrks, line 514: There a new assertion for error message in this test from the original PR in this test

rest/admin_api_test.go
```diff
@@ -3635,11 +3635,13 @@ func TestNotExistentDBRequest(t *testing.T) {
 
 	// Request to non-existent db with valid credentials
 	resp := rt.SendAdminRequestWithAuth("PUT", "/dbx/_config", "", "random", "password")
-	assertStatus(t, resp, http.StatusForbidden)
+	assertStatus(t, resp, http.StatusUnauthorized)
+	require.Contains(t, resp.Body.String(), ErrInvalidLogin.Message)
```
bbrks: This is different than the status code assertion on `master`, which is still `Forbidden`

rest/api_test.go
```diff
@@ -118,10 +124,15 @@ func TestDisablePublicBasicAuth(t *testing.T) {
...
-
+	require.Contains(t, response.Body.String(), ErrInvalidLogin.Message)
 	// Also check that we can't create a session through POST /db/_session
 	response = rt.SendRequest(http.MethodPost, "/db/_session", `{"name":"user1","password":"letmein"}`)
 	assertStatus(t, response, http.StatusUnauthorized)
+	require.Contains(t, response.Body.String(), ErrLoginRequired.Message)
+
+	response = rt.SendRequest(http.MethodPost, "/notadb/_session", `{"name":"user1","password":"letmein"}`)
+	assertStatus(t, response, http.StatusUnauthorized)
+	require.Contains(t, response.Body.String(), ErrLoginRequired.Message)
```
bbrks, line 141: Nit: This is above the assertions for`/_session` in `master`. It would probably be best to keep them consistent.

bbrks (changes requested):
This looks slightly different than the code that went into `master` - is that intentional?

rest/admin_api_test.go
torcolvin: > This looks slightly different than the code that went into `master` - is that intentional?

I wish it weren't different, but the code that handles inactiveDatabases makes the error codes different. As long as the error codes match for existing database but unable to access and non existent database, I think this is OK.

The starting state of 3.0 and 3.1 was always different https://docs.google.com/spreadsheets/d/1aXL9D7R54R0prYr-6ini-Qrk3Pj42k3gE4VQbP-lCaI/edit#gid=0 and I've preserved the error codes.

torcolvin:
> This looks slightly different than the code that went into `master` - is that intentional?

I wish it weren't different, but the code that handles inactiveDatabases makes the error codes different. As long as the error codes match for existing database but unable to access and non existent database, I think this is OK.

The starting state of 3.0 and 3.1 was always different https://docs.google.com/spreadsheets/d/1aXL9D7R54R0prYr-6ini-Qrk3Pj42k3gE4VQbP-lCaI/edit#gid=0 and I've preserved the error codes.

rest/handler.go
```diff
@@ -209,17 +213,24 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
...
+			if httpError, ok := err.(*base.HTTPError); ok && httpError.Status == http.StatusNotFound {
+				if shouldCheckAdminAuth {
 					authorized, err := h.checkAdminAuthenticationOnly()
 					if err != nil {
 						return err
 					}
 
-					if authorized {
-						return base.HTTPErrorf(http.StatusForbidden, "")
+					if !authorized || h.providedAuthCredentials() {
```
adamcfraser: On master we only use providedAuthCredentials on the public port case - it looks like we're using it in conjunction with admin auth here (inside shouldCheckAdminAuth), which seems incorrect?  Which combination is this handling?

rest/handler.go
```diff
@@ -209,17 +213,24 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
...
+						return ErrInvalidLogin
+					}
```
adamcfraser, line 225: Line 225 looks like it should have `else { return base.HTTPErrorf(http.StatusForbidden, "")}` to handle the case of an admin user that successfully authenticated requesting a db that doesn't exist.

rest/handler.go
```diff
@@ -209,17 +213,24 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
...
+				} else {
+					if h.privs == regularPrivs || h.privs == publicPrivs {
```
adamcfraser, line 228: Is this privs check required, if we're in the else for "shouldCheckAdminAuth"?

adamcfraser (dismissed):

rest/handler.go
torcolvin: This covers the case (which is maybe test-only) where h.server.config.API.AdminInterfaceAuthentication is nil

adamcfraser: I see.  These conditions are organized in a way that's a bit hard to follow, but I guess the `else` on line 226 covers both "public port" and "admin port but admin auth is disabled", and the latter falls through to returning 'not found' on line 237.

rest/handler.go
torcolvin: Seems to be unused.

rest/handler.go
torcolvin: Fixed.

adamcfraser:
This looks good to me, and integration test is clean.  @bbrks can you sanity check that your review comments are addressed before we merge this?

bbrks (approved):

PR #6190, opened 2023-04-13
Sync optimizations for changes and proposeChanges

snej:
CBG-2851

Describe your PR here...
* **Skip bucket Get when a new doc is pushed via BLIP.** When a client pushes a doc and `proposeChanges` sees that the docID doesn't exist in the bucket, the `rev` handler will write the doc to the bucket as new without trying to get an existing copy first. This will usually succeed, saving time. If it fails, it's handled like any other CAS mismatch, by retrying and getting the existing doc.
  - Added a field BlipSyncContext.pendingInsertions, a set of docIDs.
  - RevDiff returns a new status `ProposedRev_OK_IsNew`.
  - When the `proposeChanges` handler gets ProposedRev_OK_IsNew, it treats it
  like ProposedRev_OK and adds the docID to pendingInsertions.
  - The `rev` handler then checks if the docID is in pendingInsertions; if so,
  it passes an empty BucketDoc with cas=0 to PutExistingRev as the
  `existingDoc`.
  - WriteUpdateWithXattr now allows a `previous` doc with cas=0, interpreting it
  as the document not previously existing.
* **Make RevDiff & CheckProposedRev read and unmarshal less.**
  - Added a new DocUnmarshalLevel constant, DocUnmarshalHistory, which unmarshals
  just the history, current rev and CAS.
  - Added GetDocSyncDataNoImport, which is like GetDocSyncData but does not
  check whether on-demand import is needed (this lets it avoid reading the doc
  body) and takes a DocumentUnmarshalLevel parameter.
  - RevDiff and CheckProposedRev call this instead. RevDiff uses the level
  DocUnmarshalHistory, and CheckProposedRev uses DocUnmarshalRev unless it
  might need the history.

## Pre-review checklist

- [x] Link upstream PRs

### DIFF
--- a/base/collection_xattr_common.go
+++ b/base/collection_xattr_common.go
@@ -182,8 +182,14 @@ func UpdateTombstoneXattr(store SubdocXattrStore, k string, xattrKey string, exp
 	return cas, err
 }
 
-// WriteUpdateWithXattr retrieves the existing doc from the bucket, invokes the callback to update the document, then writes the new document to the bucket.  Will repeat this process on cas
-// failure.  If previousValue/xattr/cas are provided, will use those on the first iteration instead of retrieving from the bucket.
+// WriteUpdateWithXattr retrieves the existing doc from the bucket, invokes the callback to update
+// the document, then writes the new document to the bucket.  Will repeat this process on cas
+// failure.
+//
+// If previous document is provided, will use it on 1st iteration instead of retrieving from bucket.
+// A zero CAS in `previous` is interpreted as no document existing; this can be used to short-
+// circuit the initial Get when the document is unlikely to already exist.
+
 func WriteUpdateWithXattr(store SubdocXattrStore, k string, xattrKey string, userXattrKey string, exp uint32, opts *sgbucket.MutateInOptions, previous *sgbucket.BucketDocument, callback sgbucket.WriteUpdateWithXattrFunc) (casOut uint64, err error) {
 
 	var value []byte
@@ -192,19 +198,23 @@ func WriteUpdateWithXattr(store SubdocXattrStore, k string, xattrKey string, use
 	var cas uint64
 	emptyCas := uint64(0)
 
-	// If an existing value has been provided, use that as the initial value
-	if previous != nil && previous.Cas > 0 {
-		value = previous.Body
-		xattrValue = previous.Xattr
-		cas = previous.Cas
-		userXattrValue = previous.UserXattr
-	}
-
 	for {
 		var err error
-		// If no existing value has been provided, retrieve the current value from the bucket
-		if cas == 0 {
-			// Load the existing value.
+		if previous != nil {
+			// If an existing value has been provided, use that as the initial value.
+			// A zero CAS is interpreted as no document existing.
+			if previous.Cas != 0 {
+				value = previous.Body
+				xattrValue = previous.Xattr
+				cas = previous.Cas
+				userXattrValue = previous.UserXattr
+			} else if previous.Body != nil || previous.Xattr != nil {
+				panic("WriteUpdateWithXattr: previous doc has zero Cas but is not empty")
+			}
+			previous = nil // a retry will get value from bucket, as below
+		} else {
+			// If no existing value has been provided, or on a retry,
+			// retrieve the current value from the bucket
 			cas, err = store.SubdocGetBodyAndXattr(k, xattrKey, userXattrKey, &value, &xattrValue, &userXattrValue)
 
 			if err != nil {
@@ -224,7 +234,6 @@ func WriteUpdateWithXattr(store SubdocXattrStore, k string, xattrKey string, use
 
 		// If it's an ErrCasFailureShouldRetry, then retry by going back through the for loop
 		if err == ErrCasFailureShouldRetry {
-			cas = 0 // force the call to SubdocGetBodyAndXattr() to refresh
 			continue
 		}
 
@@ -258,7 +267,6 @@ func WriteUpdateWithXattr(store SubdocXattrStore, k string, xattrKey string, use
 		value = nil
 		xattrValue = nil
 		cas = 0
-
 	}
 }
 
--- a/base/collection_xattr_common.go
+++ b/base/collection_xattr_common.go
@@ -208,8 +208,6 @@ func WriteUpdateWithXattr(store SubdocXattrStore, k string, xattrKey string, use
 				xattrValue = previous.Xattr
 				cas = previous.Cas
 				userXattrValue = previous.UserXattr
-			} else if previous.Body != nil || previous.Xattr != nil {
-				panic("WriteUpdateWithXattr: previous doc has zero Cas but is not empty")
 			}
 			previous = nil // a retry will get value from bucket, as below
 		} else {
--- a/db/blip_collection_context.go
+++ b/db/blip_collection_context.go
@@ -43,6 +43,11 @@ type blipCollections struct {
 	sync.RWMutex
 }
 
+// Max number of docIDs to keep in pendingInsertions. (Normally items added to this set are
+// removed soon thereafter when the client sends the `rev` message; this limit is just to cover
+// failure cases where a client never sends the revs, to keep the set from growing w/o bound.)
+const kMaxPendingInsertions = 1000
+
 // newBlipSyncCollection constructs a context to hold all blip data for a given collection.
 func newBlipSyncCollectionContext(dbCollection *DatabaseCollection) *blipSyncCollectionContext {
 	c := &blipSyncCollectionContext{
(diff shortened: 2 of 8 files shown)


db/blip_sync_context.go
```diff
@@ -115,6 +115,8 @@ type BlipSyncContext struct {
 	readOnly bool
 
 	collections *blipCollections // all collections handled by blipSyncContext, implicit or via GetCollections
+
+	pendingInsertions map[string]struct{} // DocIDs from handleProposeChanges that aren't in the db
```
torcolvin: I think this needs to be in `blipSyncCollectionContext` because the docIDs can overlap for multiple collections.

db/blip_sync_context.go
```diff
@@ -655,3 +657,24 @@ func toHistory(revisions Revisions, knownRevs map[string]bool, maxHistory int) [
 	}
 	return history
 }
+
+// Remembers a docID that doesn't exist in the bucket at the time handleProposeChanges ran.
+func (bsc *BlipSyncContext) notePendingInsertion(docID string) {
+	bsc.allowedAttachmentsLock.Lock() // TODO: Rename this lock?
+	defer bsc.allowedAttachmentsLock.Unlock()
```
torcolvin: I don't think this has to be allowedAttachmentsLock as above, this is collection specific locking

db/blip_sync_context.go
torcolvin: we also use `base.Set` for this structure

db/crud.go
```diff
@@ -187,6 +187,31 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
...
 }
 
+// This gets *just* the Sync Metadata (_sync field) rather than the entire doc, for efficiency
+// reasons. Unlike GetDocSyncData it does not check for on-demand import; this means it does not
+// need to read the doc body from the bucket.
+func (db *DatabaseCollection) GetDocSyncDataNoImport(ctx context.Context, docid string, level DocumentUnmarshalLevel) (SyncData, error) {
+	if db.UseXattrs() {
+		var err error
+		if key := realDocID(docid); key == "" {
+			err = base.HTTPErrorf(400, "Invalid doc ID")
```
```suggestion
			err = base.HTTPErrorf(400, "Invalid doc ID: %q", base.UD(docid))
```

torcolvin (commented):
This is not a thorough review from me, just spotted something that is not multi-collection aware.

base/collection_xattr_common.go
```diff
@@ -192,19 +198,23 @@ func WriteUpdateWithXattr(store SubdocXattrStore, k string, xattrKey string, use
...
+		if previous != nil {
+			// If an existing value has been provided, use that as the initial value.
+			// A zero CAS is interpreted as no document existing.
+			if previous.Cas != 0 {
+				value = previous.Body
+				xattrValue = previous.Xattr
+				cas = previous.Cas
+				userXattrValue = previous.UserXattr
+			} else if previous.Body != nil || previous.Xattr != nil {
+				panic("WriteUpdateWithXattr: previous doc has zero Cas but is not empty")
```
torcolvin: It seems like this is an impossible condition, but it would still be better to not outright panic - 

base.FatalfCtx() will log and panic?

db/blip_collection_context.go
```diff
@@ -50,6 +52,27 @@ func newBlipSyncCollectionContext(dbCollection *DatabaseCollection) *blipSyncCol
 	return c
 }
 
+// Remembers a docID that doesn't exist in the collection at the time handleProposeChanges ran.
+func (bsc *blipSyncCollectionContext) notePendingInsertion(docID string) {
+	bsc.pendingInsertionsLock.Lock() // TODO: Rename this lock?
```
torcolvin:
```suggestion
	bsc.pendingInsertionsLock.Lock()
```

db/blip_collection_context.go
```diff
@@ -50,6 +52,27 @@ func newBlipSyncCollectionContext(dbCollection *DatabaseCollection) *blipSyncCol
...
+	defer bsc.pendingInsertionsLock.Unlock()
+	if bsc.pendingInsertions == nil {
+		bsc.pendingInsertions = base.Set{}
+	}
```
torcolvin: move this to `newBlipCollectionContext`

base/collection_xattr_common.go
adamcfraser: Can we just return error here?  panic isn't the preferred approach for low level operations as it's less clear when they are properly recovered, and we don't need a bug to bring down the SG process in this case.

db/blip_collection_context.go
```diff
@@ -50,6 +52,27 @@ func newBlipSyncCollectionContext(dbCollection *DatabaseCollection) *blipSyncCol
...
+	bsc.pendingInsertions.Add(docID)
```
adamcfraser: It looks like the size of pendingInsertions could be unbounded in the case of a client bug (or an ISGR bug) where the associated rev message is never sent.  Can we be defensive about that by limiting the maximum size of pendingInsertions, and falling back to the less efficient handling if it's filled?  (along with logging a warning on the case where the limit is reached)

db/crud.go
```diff
@@ -187,6 +187,31 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
 
 }
 
+// This gets *just* the Sync Metadata (_sync field) rather than the entire doc, for efficiency
+// reasons. Unlike GetDocSyncData it does not check for on-demand import; this means it does not
+// need to read the doc body from the bucket.
+func (db *DatabaseCollection) GetDocSyncDataNoImport(ctx context.Context, docid string, level DocumentUnmarshalLevel) (SyncData, error) {
+	if db.UseXattrs() {
+		var err error
+		if key := realDocID(docid); key == "" {
```
adamcfraser: I think this realDocID handling in the original GetDocSyncData is because we only call that function from inside a changes feed, where we may have pseudo-docIDs like `_user:bob` for things like channel backfill.  I don't think it's necessary for the usages here?  (revDiff, checkProposedRev)

db/crud.go
```diff
@@ -187,6 +187,31 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
...
+				var doc *Document
+				doc, err = unmarshalDocumentWithXattr(key, nil, xattrValue, nil, cas, level)
+				if err == nil {
+					return doc.SyncData, nil
+				}
+			}
+		}
+		return SyncData{}, err
+	} else {
+		return db.GetDocSyncData(ctx, docid)
```
adamcfraser: Is switching from the previous`GetDocument` to `GetDocSyncData` in the non-xattr case safe/equivalent?  GetDocument looks like it includes things like checkForUpgrade at least.

adamcfraser (changes requested):
Overall the enhancements look great, some questions on the details.

db/crud.go
snej: Adam pointed out I can remove the test that causes this error.

db/crud.go
snej: Hm. Then shouldn't GetDocSyncData itself check for upgrade?

snej: Looks like GetDocSyncData is only called by the changes feed; maybe that isn't supposed to upgrade?
I'll add the checkForUpgrade code (from GetDocumentWithRaw) to GetDocSyncDataNoImport.

base/collection_xattr_common.go
snej: I'll take this out -- IIRC this was there because I wasn't sure if something would legitimately pass a `previous` with a zero CAS, and if so this would let me find out when running tests.

db/blip_collection_context.go
snej: done

base/collection_xattr_common.go
snej: This case doesn't actually occur (I added the panic to flush it out if it did legitimately happen, but it was never called.)

snej:
Note that I fixed the test failure by having RevsDiff treat ErrXattrNotFound the same as NotFound ... I'm not sure that's correct. Should the ErrXattrNotFound be handled differently?

adamcfraser:
> Note that I fixed the test failure by having RevsDiff treat ErrXattrNotFound the same as NotFound ... I'm not sure that's correct. Should the ErrXattrNotFound be handled differently?

The approach you've got makes sense for this use case - GetXattr differentiates between ErrNotFound and ErrXattrNotFound for use cases that want to handle them differently.

adamcfraser:
Changes look good, but we should run integration tests before merging to make sure everything is clean when running against Couchbase Server.  There were some fixes on master specifically around integration tests.  @snej can you rebase on master and then [kick off integration tests](https://jenkins.sgwdev.com/job/SyncGateway-Integration/build?delay=0sec)?

snej:
@adamcfraser Unfortunately Jenkins says "snej is missing the Job/Build permission"...

torcolvin:
> @adamcfraser Unfortunately Jenkins says "snej is missing the Job/Build permission"...

I've granted you permission on jenkins.

snej:
I just kicked off a build, but it looks like [Adam already ran the integration tests Friday](https://jenkins.sgwdev.com/job/SyncGateway-Integration/1745/) and they passed.

adamcfraser (approved):

PR #6201, opened 2023-04-20
Feature/v8 build tag

torcolvin:
Add a build tag to enable v8 or otto only.

I'm trying to add this to jenkins automation, but when I run `go test -tags cb_sg_enterprise -tags
- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1722/

### DIFF
--- a/js/underscore-umd-min.js
+++ b/js/underscore-umd-min.js
@@ -1,3 +1,13 @@
+/**
+ * Copyright 2023-Present Couchbase, Inc.
+ *
+ * Use of this software is governed by the Business Source License included
+ * in the file licenses/BSL-Couchbase.txt.  As of the Change Date specified
+ * in that file, in accordance with the Business Source License, use of this
+ * software will be governed by the Apache License, Version 2.0, included in
+ * the file licenses/APL2.txt.
+ */
+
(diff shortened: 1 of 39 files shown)


snej:
I'm not clear on what all these code changes do. Are you trying to do the equivalent of `#ifdef`-ing all the V8-dependent code so it's possible to build without V8?

torcolvin:
> I'm not clear on what all these code changes do. Are you trying to do the equivalent of `#ifdef`-ing all the V8-dependent code so it's possible to build without V8?

Yeah, the idea is to build without V8 for beryllium and and with V8 for elixir. We also know it won't compile on windows at all, so this would allow us to merge to master.

js/underscore-umd-min.js
```diff
@@ -1,3 +1,13 @@
+/**
+ * Copyright 2023-Present Couchbase, Inc.
```
snej: We can't claim copyright on this -- it's a 3rd party JS library.

js/underscore-umd.js
```diff
@@ -1,3 +1,13 @@
+/**
+ * Copyright 2023-Present Couchbase, Inc.
```
snej: We can't claim copyright on this -- it's a 3rd party JS library.

snej (changes requested):

snej (approved):

PR #6205, opened 2023-04-21
CBG-2877 allow options to contain db-scoped CORS

torcolvin:
This allows DB scoped methods to have CORS information.

I considered better ways to solve this with CORS middleware like https://github.com/go-chi/cors (there's a gorilla mux version, but gorilla mux is dead).

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1727/

### DIFF
--- a/rest/routing.go
+++ b/rest/routing.go
@@ -9,7 +9,6 @@
 package rest
 
 import (
-	"fmt"
 	"net/http"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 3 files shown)


rest/routing.go
```diff
@@ -368,18 +369,28 @@ func wrapRouter(sc *ServerContext, privs handlerPrivs, router *mux.Router) http.
...
+			fmt.Println("dbName: ", dbName)
+			if dbName != "" {
+				db, err := h.server.GetActiveDatabase(dbName)
+				if err == nil {
+					cors = db.CORS
+				}
+			}
+			if cors != nil && privs != adminPrivs && privs != metricsPrivs {
+				cors.AddResponseHeaders(rq, response)
+			}
```
adamcfraser, line 394: Proposed changes:
- check whether VARS are populated prior to wouldMatch - if so, can use those directly.  Otherwise:
- return the matchInfo from wouldMatch, and use it to check for either {db} or {keyspace} on the matching path
- take the first match (while iterating over methods on line 374)
- use that to swap to db.CORS prior to calling AddResponseHeaders and the remaining code.

adamcfraser (approved):

PR #6207, opened 2023-04-24
CBG-2880: Cancel timed out gocbcore dcp operations

chvck:
When timeouts occur the gocbcore operations must be cancelled to flush them from the agent's system. Not doing so can result in handlers being called unexpectedly, e.g. ending up writing to a channel that is no longer being read from.

CBG-2880

Describe your PR here...

## Pre-review checklist
- [ x ] Removed debug logging (`fmt.Print`, `log.Print`, ...)
- [ x ] Logging sensitive data? Make sure it's tagged (e.g. `base.UD(docID)`, `base.MD(dbName)`)
- [ x ] Updated relevant information in the API specifications (such as endpoint descriptions, schemas, ...) in `docs/api`

- [ x ] Link upstream PRs
- [ x ] Update Go module dependencies when merged

### DIFF
--- a/base/dcp_client.go
+++ b/base/dcp_client.go
@@ -168,7 +168,9 @@ func (dc *DCPClient) getCollectionHighSeqNos(collectionID uint32) ([]uint64, err
 	// serverIdx start at 1 (-1 in gocbcore), 0 refers to the master node
 	for serverIdx := 1; serverIdx <= numServers; serverIdx++ {
 
-		highSeqNoError := make(chan error)
+		// This has to be buffered so that Cancel() below doesn't lead to blocking in the callback.
+		// (If Cancel succeeds then it will lead to directly calling the callback).
+		highSeqNoError := make(chan error, 1)
 		highSeqNoCallback := func(entries []gocbcore.VbSeqNoEntry, err error) {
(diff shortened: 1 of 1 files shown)


chvck:
I've introduced another bug here. Will push a fix shortly.

base/dcp_client.go
```diff
@@ -195,6 +197,8 @@ func (dc *DCPClient) getCollectionHighSeqNos(collectionID uint32) ([]uint64, err
 				return nil, err
 			}
 		case <-time.After(getVbSeqnoTimeout):
+			op.Cancel()
+			<-highSeqNoError
```
adamcfraser, line 201: @chvck Is this still required now that this is a buffered channel?  (same below for openStreamError)

chvck: Possibly not, no. Just ensures the value is read and everything is tidied up before returning. Could also read the error value from it, and return success if there is no error - possible if Cancel and successful response return are racing.

adamcfraser: With the additional information that op.Cancel() is always going to invoke the callback, the current approach makes sense.

adamcfraser (approved):

PR #6208, opened 2023-04-24
Implement cbgt mgrEventHandlers

adamcfraser:
OnFeedError is required when using a gocbcore feed to trigger reconnection to a feed after receiving an EOF.

CBG-2899

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1729/


adamcfraser:
TestRequireResync failure in the integration tests looks like a potential race in the test - the call to resync fails because the db isn't offline, but logs that it's closing the db immediately after that.

The test passes when run locally and in isolation on jenkins:
https://jenkins.sgwdev.com/job/SyncGateway-Integration/1730/ (single test)
https://jenkins.sgwdev.com/job/SyncGateway-Integration/1731/ (single package)

bbrks (commented):
👍

PR #6214, opened 2023-04-26
CBG-2894: Reject user auth when channel threshold is over 500 

gregns1:
CBG-2894

When channels set on a user have changed we need to be able to see how many channels this user has to warn if over warning threshold and to reject any future authorization requests for the user if they have exceeded the channel limit on the user. This code will only be called if a change in channel set has been detected removing the need for this check to be performed each time a user authenticates. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1765/

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -197,7 +197,7 @@ func (auth *Authenticator) InheritedChannels(user User, princ Principal) error {
 		}
 	})
 
-	// Error at >=500 channels if ServerlessChannelThreshold is set
+	// Error if ServerlessChannelThreshold is set and is >= than the threshold
 	if serverlessChanThreshold := auth.ServerlessChannelThreshold; serverlessChanThreshold != 0 {
 		if uint32(channelCount) >= serverlessChanThreshold {
 			base.ErrorfCtx(auth.LogCtx, "User ID: %v channel count: %d exceeds %d for channels per user threshold. Auth will be rejected until rectified",
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -284,7 +284,8 @@ func (auth *Authenticator) rebuildChannels(princ Principal) (changed bool, err e
 		}
 	}
 
-	if auth.ServerlessChannelThreshold != 0 {
+	// only warn/limit if the threshold is set and if we are in this function as a "user" not a role
+	if auth.ServerlessChannelThreshold != 0 && princUser != nil {
 		// Warning at 50 channels
 		princUser.GetWarnChanSync().Do(func() {
 			if channelsPerUserThreshold := auth.ChannelsWarningThreshold; channelsPerUserThreshold != nil {
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -224,19 +224,21 @@ func (auth *Authenticator) getPrincipal(docID string, factory func() Principal) 
 	return princ, nil
 }
 
-func (auth *Authenticator) channelCount(princ Principal, user User, roles []Role, scope, collection string) int {
+// channelCount will count the number of channels user has in a scope and collection then return this count
+func (auth *Authenticator) channelCount(princ Principal, user User, roles []Role, scope, collection string) uint32 {
 	cumulativeChannels := 0
 
-	channels := princ.CollectionChannels(scope, collection).Copy()
+	channels := princ.CollectionChannels(scope, collection)
 	for _, role := range roles {
 		roleSince := user.RoleNames()[role.Name()]
 		channels.AddAtSequence(role.CollectionChannels(scope, collection), roleSince.Sequence)
 	}
 	// We want to keep track of cumulative number of channels for limits across all collections
 	cumulativeChannels += len(channels)
-	return cumulativeChannels
+	return uint32(cumulativeChannels)
 }
 
+// getUserRoles gets all roles a user has been granted
 func (auth *Authenticator) getUserRoles(user User) ([]Role, error) {
 	roles := make([]Role, 0, len(user.RoleNames()))
 	for name := range user.RoleNames() {
@@ -257,11 +259,10 @@ func (auth *Authenticator) getUserRoles(user User) ([]Role, error) {
 func (auth *Authenticator) rebuildChannels(princ Principal) (changed bool, err error) {
 
 	changed = false
-	cumulativeChannels := 0
+	var cumulativeChannels uint32
 	var roles []Role
 	var princUser User
 	if user, ok := princ.(User); ok {
-		roles, _ = auth.getUserRoles(user)
 		princUser = user
 	}
 
@@ -276,6 +277,12 @@ func (auth *Authenticator) rebuildChannels(princ Principal) (changed bool, err e
 				changed = true
 				if auth.ServerlessChannelThreshold != 0 {
 					if user, ok := princ.(User); ok {
+						if roles == nil {
+							roles, err = auth.getUserRoles(user)
+							if err != nil {
+								return changed, err
+							}
+						}
 						channelCount := auth.channelCount(princ, user, roles, scope, collection)
 						cumulativeChannels += channelCount
 					}
@@ -289,7 +296,7 @@ func (auth *Authenticator) rebuildChannels(princ Principal) (changed bool, err e
 		// Warning at 50 channels
 		princUser.GetWarnChanSync().Do(func() {
 			if channelsPerUserThreshold := auth.ChannelsWarningThreshold; channelsPerUserThreshold != nil {
-				if uint32(cumulativeChannels) >= *channelsPerUserThreshold {
+				if cumulativeChannels >= *channelsPerUserThreshold {
 					base.WarnfCtx(auth.LogCtx, "User ID: %v channel count: %d exceeds %d for channels per user warning threshold",
 						base.UD(princUser.Name()), cumulativeChannels, *channelsPerUserThreshold)
 				}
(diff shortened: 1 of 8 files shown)


auth/auth.go
```diff
@@ -186,6 +234,12 @@ func (auth *Authenticator) getPrincipal(docID string, factory func() Principal)
 			}
 			if channelsChanged {
 				changed = true
+				if user, ok := princ.(User); ok {
+					err = auth.InheritedChannels(user, princ)
```
bbrks: I think we end up doing most of this work inside `rebuildChannels` anyway (the exception being the role channels).

I'm wondering if it's worth putting the warning/limit logic inside there (behind a parameter to limit to only this codepath?) to avoid having to iterate the full set of channels twice.

auth/auth.go
```diff
@@ -160,6 +161,53 @@ func (auth *Authenticator) GetRoleIncDeleted(name string) (Role, error) {
 	return role, err
 }
 
+func (auth *Authenticator) InheritedChannels(user User, princ Principal) error {
+	cumulativeChannels := ch.TimedSet{}
```
bbrks: We can probably save a bit of memory by counting as we're iterating, rather than storing what channels we see then grabbing length.

auth/auth.go
```diff
@@ -160,6 +161,53 @@ func (auth *Authenticator) GetRoleIncDeleted(name string) (Role, error) {
...
+		for collection, _ := range collections {
+			channels := princ.CollectionChannels(scope, collection).Copy()
+			for name := range user.RoleNames() {
+				role, err := auth.GetRole(name)
+				if err != nil {
+					return err
+				} else if role != nil {
+					roles = append(roles, role)
+				}
+			}
```
bbrks: Isn't this only required once? Roles are associated with the database, not a specific collection

bbrks (dismissed):
some initial comments

auth/auth.go
```diff
@@ -223,13 +224,47 @@ func (auth *Authenticator) getPrincipal(docID string, factory func() Principal)
 	return princ, nil
 }
 
+func (auth *Authenticator) channelCount(princ Principal, user User, roles []Role, scope, collection string) int {
```
torcolvin: Can you add a docstring for this?

auth/auth.go
```diff
@@ -239,9 +274,36 @@ func (auth *Authenticator) rebuildChannels(princ Principal) (changed bool, err e
...
+					}
+				}
+			}
+		}
+	}
+
+	// only warn/limit if the threshold is set and if we are in this function as a "user" not a role
+	if auth.ServerlessChannelThreshold != 0 && princUser != nil {
+		// Warning at 50 channels
+		princUser.GetWarnChanSync().Do(func() {
```
torcolvin: This is going to not work as intended, since this only works the version first time this code is run, I'd write a test where it is over threshold the first time and then under threshold the second time, and it should only warn once.

What happens if it high (warning), then low, then high again (error)?

auth/auth.go
```diff
@@ -223,13 +224,47 @@ func (auth *Authenticator) getPrincipal(docID string, factory func() Principal)
...
+}
+
 // Rebuild channels computes the full set of channels for all collections defined for the authenticator.
 // For each collection in Authenticator.collections:
 //   - if there is no CollectionAccess on the principal for the collection, rebuilds channels for that collection
 //   - If CollectionAccess on the principal has been invalidated, rebuilds channels for that collection
 func (auth *Authenticator) rebuildChannels(princ Principal) (changed bool, err error) {
 
 	changed = false
+	cumulativeChannels := 0
```
torcolvin: declare as uint32 so you don't have to cast it?

base/error.go
```diff
@@ -67,6 +67,9 @@ var (
 
 	// ErrConfigRegistryReloadRequired is returned when a db config fetch requires a registry reload based on version mismatch (config is newer)
 	ErrConfigRegistryReloadRequired = &sgError{"Config registry reload required"}
+
+	// ErrMaximumChannelsForUserExceeded is returned when running in serverless mode and the user has more than 500 channels granted to them
+	ErrMaximumChannelsForUserExceeded = &sgError{"User has exceeded maximum of 500 channels"}
```
torcolvin: Let's move `ServerlessChannelLimit` to base and then we can use this constant here, otherwise it runs the risk of being out of date

auth/auth_test.go
```diff
@@ -2752,6 +2752,59 @@ func TestObtainChannelsForDeletedRole(t *testing.T) {
...
+			if testCase.Collection {
+				opts.Collections = map[string]map[string]struct{}{
+					"scope1": {"collection1": struct{}{}, "collection2": struct{}{}},
+				}
+			}
+			auth := NewAuthenticator(dataStore, nil, opts)
+			user1, err := auth.NewUser("user1", "pass", ch.BaseSetOf(t, "ABC"))
+			require.NoError(t, err)
+			err = auth.Save(user1)
+			assert.NoError(t, err)
```
torcolvin:
```suggestion
			require.NoError(t, err)
```

auth/auth_test.go
```diff
@@ -2752,6 +2752,59 @@ func TestObtainChannelsForDeletedRole(t *testing.T) {
...
+			_, err = auth.AuthenticateUser("user1", "pass")
+			assert.NoError(t, err)
```
torcolvin:
```suggestion
			require.NoError(t, err)
```

auth/auth_test.go
```diff
@@ -2752,6 +2752,59 @@ func TestObtainChannelsForDeletedRole(t *testing.T) {
...
+
+			if !testCase.Collection {
+				user1.SetCollectionExplicitChannels("_default", "_default", ch.AtSequence(ch.BaseSetOf(t, "ABC", "DEF", "GHI", "JKL", "MNO", "PQR"), 1), 1)
+				err = auth.Save(user1)
+				assert.NoError(t, err)
```
torcolvin:
```suggestion
				require.NoError(t, err)
```

auth/auth_test.go
```diff
@@ -2752,6 +2752,59 @@ func TestObtainChannelsForDeletedRole(t *testing.T) {
...
+			} else {
+				user1.SetCollectionExplicitChannels("scope1", "collection1", ch.AtSequence(ch.BaseSetOf(t, "ABC", "DEF", "GHI", "JKL"), 1), 1)
+				user1.SetCollectionExplicitChannels("scope1", "collection2", ch.AtSequence(ch.BaseSetOf(t, "MNO", "PQR"), 1), 1)
+				err = auth.Save(user1)
+				assert.NoError(t, err)
```
torcolvin:
```suggestion
				require.NoError(t, err)
```

auth/auth.go
```diff
@@ -223,13 +224,47 @@ func (auth *Authenticator) getPrincipal(docID string, factory func() Principal)
 	return princ, nil
 }
 
+func (auth *Authenticator) channelCount(princ Principal, user User, roles []Role, scope, collection string) int {
+	cumulativeChannels := 0
+
+	channels := princ.CollectionChannels(scope, collection).Copy()
```
torcolvin: why does this need a copy but `role.CollectionChannels(scope, collection)` doesn't need a copy?

auth/auth.go
gregns1: I'll do this but will have to cast int to uint32 in the channelCount function

auth/auth.go
gregns1: I took this apporach as elsewhere we do this warning using the sync once. I don't think you can reset the sync.Once so I either keep as is to follow what is done elsewhere in the code or I remove it (but could end up with a lot of warnings in logs as result)

auth/auth.go
```diff
@@ -223,13 +235,75 @@ func (auth *Authenticator) getPrincipal(docID string, factory func() Principal)
 	return princ, nil
 }
 
+func (auth *Authenticator) inheritedCollectionChannels(user User, scope, collection string) (ch.TimedSet, error) {
```
torcolvin, line 239: add a dosctring for this something like return channel names for given collection

auth/auth.go
```diff
@@ -223,13 +235,75 @@ func (auth *Authenticator) getPrincipal(docID string, factory func() Principal)
...
+
+	channels := user.CollectionChannels(scope, collection)
+	for _, role := range roles {
+		roleSince := user.RoleNames()[role.Name()]
+		channels.AddAtSequence(role.CollectionChannels(scope, collection), roleSince.Sequence)
+	}
+	return channels, nil
+}
+
+func (auth *Authenticator) getInheritedChannelsLength(user User) (int, error) {
```
torcolvin, line 254:
```suggestion
// getInheritedChannelsCount returns number of channels for a user across the all collections
func (auth *Authenticator) getInheritedChannelsCount(user User) (int, error) {
```

auth/auth.go
```diff
@@ -223,13 +235,75 @@ func (auth *Authenticator) getPrincipal(docID string, factory func() Principal)
...
+	var cumulativeChannels int
+	for scope, collections := range auth.Collections {
+		for collection, _ := range collections {
```
torcolvin: nit

```suggestion
		for collection := range collections {
```

auth/auth.go
```diff
@@ -223,13 +235,75 @@ func (auth *Authenticator) getPrincipal(docID string, factory func() Principal)
...
+			if err != nil {
+				return 0, err
+			}
+			cumulativeChannels += len(channels)
+		}
+	}
+	return cumulativeChannels, nil
+}
+
+func (auth *Authenticator) checkChannelLimits(channels int, user User) error {
```
torcolvin, line 269:
```suggestion
checkChannelsLimits will log a warning when user is approaching channel limit and returnerror if the user has exceeded the channel limit.
func (auth *Authenticator) checkChannelLimits(channels int, user User) error {
```

auth/auth.go
```diff
@@ -223,13 +235,75 @@ func (auth *Authenticator) getPrincipal(docID string, factory func() Principal)
...
+			}
+		}
+	})
+
+	// Error if ServerlessChannelThreshold is set and is >= than the threshold
+	if uint32(channels) >= auth.ServerlessChannelThreshold {
+		base.ErrorfCtx(auth.LogCtx, "User ID: %v channel count: %d exceeds %d for channels per user threshold. Auth will be rejected until rectified",
+			base.UD(user.Name()), channels, auth.ServerlessChannelThreshold)
+		return base.ErrMaximumChannelsForUserExceeded
+	}
```
torcolvin: You quite possibly want to do this check before sync.Once if you are already over for the first time to avoid confusing logging.

torcolvin (commented):
Just some nitpicking comments.

torcolvin (approved):

PR #6217, opened 2023-04-27
Tweak wording of doc_ids filter for /_changes API

bbrks:
- `doc_ids=[id1]` != `doc_ids=id1`
- The endpoint will try to match a literal `[id1]` document ID


bbrks:
We should actually be handling this API endpoint with both formats for CouchDB compatibility. We have code in place so need to investigate further.

Edit: Value needs to be actually a full valid JSON array: e.g. `["id1"]`

PR #6225, opened 2023-05-01
allow rbac tests to run

torcolvin:
I'm not actually sure what feature is only supported in 7.1 in this test - I think it's user creation, but if it's not, let me know how to change this test.

There were a few changes to make to this test with database and keyspace which are expected.

Unexpected was that `/_stats` and `/_expvar/` are granted to `syncGatewayConfigurator`

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1760/

### DIFF
--- a/base/main_test_bucket_pool_config.go
+++ b/base/main_test_bucket_pool_config.go
@@ -13,6 +13,7 @@ import (
 	"errors"
 	"os"
 	"strconv"
+	"testing"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


adamcfraser (commented):
The majority of the RBAC permissions being used in that test (sync_gateway_dev_ops, etc - everything except mobile_sync_gateway) didn't exist until server 7.1.  So instead of supportsSettingRBAC, the cluster capability is more like 'includes mobile permissions'.

adamcfraser (approved):

PR #6226, opened 2023-05-02
CBG-2895: Add static replication connection limit

gregns1:
CBG-2895

Adding a configurable limit option for concurerent replication connections, this works for one shot and continuous feeds. Adding unit test for both scenarios and also have it limiting the changes endpoint requests for non normal feed requests. Unit test for this functionality is also included. I have also added a stat to track number of rejected connections. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1777/

### DIFF
--- a/base/error.go
+++ b/base/error.go
@@ -70,6 +70,9 @@ var (
 
 	// ErrMaximumChannelsForUserExceeded is returned when running in serverless mode and the user has more than 500 channels granted to them
 	ErrMaximumChannelsForUserExceeded = &sgError{fmt.Sprintf("User has exceeded maximum of %d channels", ServerlessChannelLimit)}
+
+	// ErrReplicationLimitExceeded is returned when then replication connection threshold is exceeded
+	ErrReplicationLimitExceeded = &sgError{"Replication limit exceeded. Try agin later."}
 )
 
 func (e *sgError) Error() string {
@@ -120,6 +123,8 @@ func ErrorAsHTTPStatus(err error) (int, string) {
 		return http.StatusServiceUnavailable, unwrappedErr.Error()
 	case ErrMaximumChannelsForUserExceeded:
 		return http.StatusInternalServerError, "Maximum number of channels exceeded for this user"
+	case ErrReplicationLimitExceeded:
+		return http.StatusServiceUnavailable, unwrappedErr.Error()
 	}
 
 	// gocb V2 errors
--- a/base/stats.go
+++ b/base/stats.go
@@ -494,7 +494,7 @@ type DatabaseStats struct {
 	// The total number of times that a sync function encountered an exception (across all collections).
 	SyncFunctionExceptionCount *SgwIntStat `json:"sync_function_exception_count"`
 	// The total number of times a replication connection is rejected due ot it being over the threshold
-	NumReplicationsRejected *SgwIntStat `json:"num_replications_rejected"`
+	NumReplicationsRejectedLimit *SgwIntStat `json:"num_replications_rejected_limit"`
 
 	// These can be cleaned up in future versions of SGW, implemented as maps to reduce amount of potential risk
 	// prior to Hydrogen release. These are not exported as part of prometheus and only exposed through expvars
@@ -1414,7 +1414,7 @@ func (d *DbStats) initDatabaseStats() error {
 	if err != nil {
 		return err
 	}
-	resUtil.NumReplicationsRejected, err = NewIntStat(SubsystemDatabaseKey, "num_replications_rejected", labelKeys, labelVals, prometheus.CounterValue, 0)
+	resUtil.NumReplicationsRejectedLimit, err = NewIntStat(SubsystemDatabaseKey, "num_replications_rejected_limit", labelKeys, labelVals, prometheus.CounterValue, 0)
 	if err != nil {
 		return err
 	}
@@ -1459,7 +1459,7 @@ func (d *DbStats) unregisterDatabaseStats() {
 	prometheus.Unregister(d.DatabaseStats.SyncFunctionCount)
 	prometheus.Unregister(d.DatabaseStats.SyncFunctionTime)
 	prometheus.Unregister(d.DatabaseStats.SyncFunctionExceptionCount)
-	prometheus.Unregister(d.DatabaseStats.NumReplicationsRejected)
+	prometheus.Unregister(d.DatabaseStats.NumReplicationsRejectedLimit)
 }
 
 func (d *DbStats) CollectionStat(scopeName, collectionName string) (*CollectionStats, error) {
--- a/base/stats.go
+++ b/base/stats.go
@@ -493,6 +493,8 @@ type DatabaseStats struct {
 	SyncFunctionTime *SgwIntStat `json:"sync_function_time"`
 	// The total number of times that a sync function encountered an exception (across all collections).
 	SyncFunctionExceptionCount *SgwIntStat `json:"sync_function_exception_count"`
+	// The total number of times a replication connection is rejected due ot it being over the threshold
+	NumReplicationsRejected *SgwIntStat `json:"num_replications_rejected"`
 
 	// These can be cleaned up in future versions of SGW, implemented as maps to reduce amount of potential risk
 	// prior to Hydrogen release. These are not exported as part of prometheus and only exposed through expvars
@@ -1409,6 +1411,10 @@ func (d *DbStats) initDatabaseStats() error {
 		return err
 	}
 	resUtil.SyncFunctionExceptionCount, err = NewIntStat(SubsystemDatabaseKey, "sync_function_exception_count", labelKeys, labelVals, prometheus.CounterValue, 0)
+	if err != nil {
+		return err
+	}
+	resUtil.NumReplicationsRejected, err = NewIntStat(SubsystemDatabaseKey, "num_replications_rejected", labelKeys, labelVals, prometheus.CounterValue, 0)
 	if err != nil {
 		return err
 	}
--- a/base/stats.go
+++ b/base/stats.go
@@ -1459,6 +1459,7 @@ func (d *DbStats) unregisterDatabaseStats() {
 	prometheus.Unregister(d.DatabaseStats.SyncFunctionCount)
 	prometheus.Unregister(d.DatabaseStats.SyncFunctionTime)
 	prometheus.Unregister(d.DatabaseStats.SyncFunctionExceptionCount)
+	prometheus.Unregister(d.DatabaseStats.NumReplicationsRejected)
 }
 
 func (d *DbStats) CollectionStat(scopeName, collectionName string) (*CollectionStats, error) {
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -2112,7 +2112,7 @@ Startup-config:
           type: integer
           maximum: 9
           minimum: 0
-        max_concurrent_connections:
+        max_concurrent_replications:
           description: Maximum number of concurrent replication connections allowed. If set to 0 this limit will be ignored.
           type: integer
       readOnly: true
@@ -2200,7 +2200,7 @@ Runtime-config:
           $ref: '#/File-logging-config'
         stats:
           $ref: '#/File-logging-config'
-    max_concurrent_connections:
+    max_concurrent_replications:
       description: Maximum number of concurrent replication connections allowed. If set to 0 this limit will be ignored.
       type: integer
       default: 0
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -2112,6 +2112,9 @@ Startup-config:
           type: integer
           maximum: 9
           minimum: 0
+        max_concurrent_connections:
+          description: Maximum number of concurrent replication connections allowed. If set to 0 this limit will be ignored.
+          type: integer
       readOnly: true
     unsupported:
       description: Settings that are not officially supported. It is highly recommended these are **not** used.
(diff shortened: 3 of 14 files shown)


rest/server_context.go
```diff
@@ -70,6 +70,7 @@ type ServerContext struct {
 	LogContextID                  string          // ID to differentiate log messages from different server context
 	fetchConfigsLastUpdate        time.Time       // The last time fetchConfigsWithTTL() updated dbConfigs
 	allowScopesInPersistentConfig bool            // Test only backdoor to allow scopes in persistent config, not supported for multiple databases with different collections targeting the same bucket
+	activeReplicatorCount         int             // The count of concurrent active replicators
```
torcolvin: if we make this an `atomic.Int32` is that going be better than accessing requiring a full context lock, especially that this is going to update frequently? The golang 1.19 atomic types are pretty great because the compiler will complain if you don't access them without loading.

docs/api/components/schemas.yaml
```diff
@@ -2162,6 +2165,46 @@ Startup-config:
       type: integer
       readOnly: true
   title: Startup-config
+Runtime-config:
+  type: object
+  properties:
+    logging:
```
torcolvin, line 2171: Is it correct to add all these log settings here, or as part of this review?

rest/adminapitest/admin_api_test.go
```diff
@@ -241,6 +241,39 @@ func TestLoggingKeys(t *testing.T) {
 	assert.Equal(t, map[string]interface{}{}, noLogKeys)
 }
 
+func TestServerlessChangesEndpointLimit(t *testing.T) {
+	base.RequireNumTestBuckets(t, 2)
+	base.SetUpTestLogging(t, base.LevelInfo, base.KeyReplicate, base.KeyHTTP, base.KeyHTTPResp, base.KeySync, base.KeySyncMsg)
+	rt := rest.NewRestTester(t, nil)
+	defer rt.Close()
+
+	resp := rt.SendAdminRequest(http.MethodPut, "/_config", `{"max_concurrent_connections" : 2}`)
```
torcolvin: If you do a `PUT` I'd be cautious of dropping some of the original configuration here as we discussed briefly with @IsaacLambat that I'm not sure we support doing a partial configuration update.

rest/adminapitest/admin_api_test.go
```diff
@@ -241,6 +241,39 @@ func TestLoggingKeys(t *testing.T) {
...
+	// send some changes requests in go routines to run concurrently along with test
+	go func() {
+		resp1 := rt.SendAdminRequest(http.MethodGet, "/{{.keyspace}}/_changes?feed=longpoll&since=999999&timeout=100000", "")
+		rest.RequireStatus(t, resp1, http.StatusOK)
+	}()
+
+	go func() {
+		resp2 := rt.SendAdminRequest(http.MethodGet, "/{{.keyspace}}/_changes?feed=longpoll&since=999999&timeout=100000", "")
+		rest.RequireStatus(t, resp2, http.StatusOK)
+	}()
```
torcolvin: I'm always cautious about creating long running go routines in a test, if you drop `rt.WaitForActiveReplicatorCount` - these can run the background.

I'm not sure how far I want to take this, but I'd be inclined to put these in a waitgroup so we know the assertions are actually run, and make sure they are finished at the end of the test. The downside of this is that the test will time out if it isn't asserting correctly. OTOH, these assertions may never get hit because you are asking for them to wait 1000000.

I think that tearing down a database in a defer will destroy these goroutines, but you should make sure in the test.

rest/blip_sync.go
```diff
@@ -71,3 +79,31 @@ func (h *handler) handleBLIPSync() error {
...
 }
+
+func (sc *ServerContext) limitConcurrentReplications(ctx context.Context) (func(sc *ServerContext), error) {
+	if sc.Config.Replicator.MaxConcurrentConnections == 0 {
+		return nil, nil
+	}
+	sc.lock.Lock()
+	defer sc.lock.Unlock()
+
+	capacity := sc.Config.Replicator.MaxConcurrentConnections
```
torcolvin: Is the reason to lock the whole config here rather than the atomic because we care about config value and also the counter value?

rest/blip_sync.go
```diff
@@ -71,3 +79,31 @@ func (h *handler) handleBLIPSync() error {
...
+	release := func(sc *ServerContext) {
+		sc.lock.Lock()
+		defer sc.lock.Unlock()
+		connections := sc.Config.Replicator.MaxConcurrentConnections
+		sc.activeReplicatorCount--
+		base.TracefCtx(ctx, base.KeyHTTP, "Released replication slot (remaining: %d/%d)", connections-sc.activeReplicatorCount, connections)
+	}
+
+	if count >= capacity {
+		base.InfofCtx(ctx, base.KeyHTTP, "Replication limit exceeded (active: %d)", count)
```
torcolvin: do you want to log what the limit is here?

docs/api/components/schemas.yaml
gregns1: They are part of the runtime config endpoint request schema. In order to add the fields I needed I had to create docs from the runtime endpoint insteead of sharing the docs from the boostrap config.

rest/adminapitest/admin_api_test.go
gregns1: There is only a get and puyt for ther _cofnig endpoint, just working with what we have at this stage. It's only this and the logging you can alter through this endpoint and looking at code I don't think it will alter logging levels as it checks if logging in rerquest body is nil or not.

docs/api/components/schemas.yaml
torcolvin: My fault, I was thinking this was db/_config.

rest/adminapitest/admin_api_test.go
torcolvin: My fault, I was thinking this was db/_config.

rest/server_context.go
gregns1: I did think about this but ideally i'd want to lock the server context so the max concurrent replicatiosn value doesn't get changed by an incoming request during execution so thought it would be best to lock. Although I understand there is some performance degredation when locking and unlocking frequently in comparison to using atomic.

rest/blip_sync.go
gregns1: Will add this, good spot.

base/error.go
```diff
@@ -67,6 +67,8 @@ var (
 
 	// ErrConfigRegistryReloadRequired is returned when a db config fetch requires a registry reload based on version mismatch (config is newer)
 	ErrConfigRegistryReloadRequired = &sgError{"Config registry reload required"}
+
+	ErrReplicationLimitExceeded = &sgError{"Replication limit exceeded. Try agin later."}
```
adamcfraser, line 75:
```suggestion
	ErrReplicationLimitExceeded = &sgError{"Replication limit exceeded. Try again later."}
```

base/stats.go
```diff
@@ -493,6 +493,8 @@ type DatabaseStats struct {
 	SyncFunctionTime *SgwIntStat `json:"sync_function_time"`
 	// The total number of times that a sync function encountered an exception (across all collections).
 	SyncFunctionExceptionCount *SgwIntStat `json:"sync_function_exception_count"`
+	// The total number of times a replication connection is rejected due ot it being over the threshold
+	NumReplicationsRejected *SgwIntStat `json:"num_replications_rejected"`
```
adamcfraser:
```suggestion
	NumReplicationsRejected *SgwIntStat `json:"num_replications_rejected_limit"`
```

adamcfraser: For me, the term 'rejected' is already strongly connected to documents rejected by the sync function, and so adding 'limit' at the end makes it more obvious at a glance what this stat represents.

docs/api/components/schemas.yaml
```diff
@@ -2112,6 +2112,9 @@ Startup-config:
           type: integer
           maximum: 9
           minimum: 0
+        max_concurrent_connections:
```
adamcfraser: Is a 'connection' always a replication?  If so I think max_concurrent_replications would be better.  Having it just be 'connection' makes it seems like any in-flight REST request would qualify.

rest/adminapitest/admin_api_test.go
adamcfraser: A longpoll changes request will stay connected until it finds new data to replicate.  So a common pattern for  avoiding timeous/sleeps in this scenario is to start the longpoll feed(s), wait for the condition you're trying to test (number of active replications), and then clean up by writing a document visible to the user, to trigger the longpoll feed to close. There's a 'caught up' stat that can be used to check when the longpoll feed has made it to the waiting state.

TestChangesNotifyChannelFilter is one example, I think.

rest/blip_sync.go
```diff
@@ -71,3 +79,31 @@ func (h *handler) handleBLIPSync() error {
 
 	return nil
 }
+
+func (sc *ServerContext) limitConcurrentReplications(ctx context.Context) (func(sc *ServerContext), error) {
+	if sc.Config.Replicator.MaxConcurrentConnections == 0 {
```
adamcfraser: This check requires synchronization for when we eventually want to support runtime modification of this value.  Is that intentionally being left out of scope?

rest/blip_sync.go
adamcfraser: Strongly agree that a more narrowly defined mutex should be used for this check.

rest/blip_sync.go
```diff
@@ -71,3 +79,31 @@ func (h *handler) handleBLIPSync() error {
...
+		return nil, nil
+	}
+	sc.lock.Lock()
+	defer sc.lock.Unlock()
+
+	capacity := sc.Config.Replicator.MaxConcurrentConnections
+	count := sc.activeReplicatorCount
+
+	release := func(sc *ServerContext) {
```
adamcfraser: Returning a function that locks sc.lock.Lock() seems like it would be prone to introducing deadlock error (where someone tries to call release while holding the lock, not realizing what it's doing).  Using a more narrow mutex might solve that, but I wonder if two different functions for 'add' and 'release'  would be preferable here.

rest/blip_sync.go
```diff
@@ -22,6 +23,13 @@ import (
 
 // HTTP handler for incoming BLIP sync WebSocket request (/db/_blipsync)
 func (h *handler) handleBLIPSync() error {
+	if release, err := h.server.limitConcurrentReplications(h.rqCtx); err != nil {
+		h.db.DbStats.Database().NumReplicationsRejected.Add(1)
+		return err
+	} else if release != nil {
+		defer release(h.server)
+	}
```
adamcfraser, line 34: I'm not sure the pattern of returning a teardown function is merited here - seems like it would be more straightforward in the form:
```
err := h.server.incrementConcurrentReplications()
if err != nil {
   // update stat, etc
   return err
}
defer h.server.decrementConcurrentReplication()
```

adamcfraser: If there's a case where err=nil and release=nil that I'm overlooking, that might change things, but I don't see that use case.

rest/blip_sync.go
```diff
@@ -71,3 +79,31 @@ func (h *handler) handleBLIPSync() error {
...
+		defer sc.lock.Unlock()
+		connections := sc.Config.Replicator.MaxConcurrentConnections
+		sc.activeReplicatorCount--
+		base.TracefCtx(ctx, base.KeyHTTP, "Released replication slot (remaining: %d/%d)", connections-sc.activeReplicatorCount, connections)
+	}
+
+	if count >= capacity {
+		base.InfofCtx(ctx, base.KeyHTTP, "Replication limit exceeded (active: %d)", count)
+		return nil, base.ErrReplicationLimitExceeded
+	} else if count < capacity {
```
adamcfraser: Since you're returning on the previous case, this doesn't need to be in an else.

rest/blip_sync.go
```diff
@@ -71,3 +79,31 @@ func (h *handler) handleBLIPSync() error {
...
+		sc.activeReplicatorCount++
+		base.TracefCtx(ctx, base.KeyHTTP, "Acquired replication slot (remaining: %d/%d)", capacity-sc.activeReplicatorCount, capacity)
```
adamcfraser: I'd probably log this as 'active: %d/%d' since it provides the same information and avoids evaluation of `capacity - sc.activeReplicatorCount`, which would occur even if trace logging isn't enabled.

rest/changes_api.go
```diff
@@ -248,6 +248,15 @@ func (h *handler) handleChanges() error {
 		feed = "normal"
 	}
 
+	if feed != "normal" {
+		release, err := h.server.limitConcurrentReplications(h.rqCtx)
```
adamcfraser: I'm on the fence about whether we omit one-shot ("normal") feeds - a one-shot changes feed can be relatively long-running.

adamcfraser (dismissed):

rest/admin_api.go
```diff
@@ -462,6 +463,11 @@ func (h *handler) handlePutConfig() error {
 		base.EnableStatsLogger(*config.Logging.Stats.Enabled)
 	}
 
+	if config.ReplicationLimit != nil {
+		h.server.Config.Replicator.MaxConcurrentReplications = *config.ReplicationLimit
+		h.server.ActiveReplicationsCounter.activeReplicatorLimit = *config.ReplicationLimit
```
bbrks, line 472: I think this needs the lock held

rest/config_startup.go
```diff
@@ -136,8 +136,9 @@ type AuthConfig struct {
 }
 
 type ReplicatorConfig struct {
-	MaxHeartbeat    *base.ConfigDuration `json:"max_heartbeat,omitempty"    help:"Max heartbeat value for _changes request"`
-	BLIPCompression *int                 `json:"blip_compression,omitempty" help:"BLIP data compression level (0-9)"`
+	MaxHeartbeat              *base.ConfigDuration `json:"max_heartbeat,omitempty"    help:"Max heartbeat value for _changes request"`
+	BLIPCompression           *int                 `json:"blip_compression,omitempty" help:"BLIP data compression level (0-9)"`
+	MaxConcurrentReplications int                  `json:"max_concurrent_replications,omitempty" help:"Maximum number of replication connections to the node"`
```
bbrks, line 141: Either this should be a `uint`, or we should add some validation, as I think setting this to anything negative will be problematic.

bbrks (changes requested):

bbrks (approved):

PR #6229, opened 2023-05-08
CBG-2928 add blip stats for database

torcolvin:
[CBG-2928](https://github.com/couchbase/sync_gateway/tree/CBG-2928)

This adds per database blip stats with an arbitrary time interval for reporting.

https://github.com/couchbase/go-blip/pull/61

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1773/


base/stats.go
```diff
@@ -430,6 +430,8 @@ type CollectionStats struct {
 }
 
 type DatabaseStats struct {
+	BlipBytesReceived *SgwIntStat `json:"blip_bytes_received"`
```
adamcfraser: Since stat names have some external visibility, I think it's preferable to use 'replication_bytes_received' and 'replication_bytes_sent'.

db/active_replicator.go
```diff
@@ -219,8 +219,8 @@ func connect(arc *activeReplicatorCommon, idSuffix string) (blipSender *blip.Sen
 		}
 	}
 
-	bsc = NewBlipSyncContext(arc.ctx, blipContext, arc.config.ActiveDB, blipContext.ID, arc.replicationStats)
-	bsc.loggingCtx = base.CorrelationIDLogCtx(context.Background(), arc.config.ID+idSuffix)
+	ctx := base.CorrelationIDLogCtx(context.Background(), arc.config.ID+idSuffix)
+	bsc = NewBlipSyncContext(ctx, blipContext, arc.config.ActiveDB, blipContext.ID, arc.replicationStats)
```
adamcfraser: The arc.ctx that was previously getting passed to NewBlipSyncContext looks like it may have already had the CorrelationIDLogCtx on it (set in ActivePushReplicator.Start or ActivePullReplicator.Start).  It feels like continuing to use arc.ctx would be the preferable approach, to avoid performing this idSuffix work in multiple places.  What do you think?

db/blip_sync_context.go
```diff
@@ -35,9 +36,10 @@ var ErrClosedBLIPSender = errors.New("use of closed BLIP sender")
 
 func NewBlipSyncContext(ctx context.Context, bc *blip.Context, db *Database, contextID string, replicationStats *BlipSyncStats) *BlipSyncContext {
 	bsc := &BlipSyncContext{
+
+		ctx:                     ctx,
```
adamcfraser: What's the benefit of refactoring loggingCtx to ctx in this PR?

db/blip_sync_context.go
```diff
@@ -655,3 +669,34 @@ func toHistory(revisions Revisions, knownRevs map[string]bool, maxHistory int) [
 	}
 	return history
 }
+
+// reportIfTimeElapsed will report stats if enough time has passed since the previous report.
+func (s *blipSyncStats) reportIfTimeElapsed(bsc *BlipSyncContext) {
+	currentTime := time.Now().UnixMilli()
+	if (currentTime - s.lastReportTime.Load()) > bsc.blipContextDb.BlipStatsReportingInterval {
+		s.lastReportTime.Store(currentTime)
```
adamcfraser: Should we be updating lastReportTime while holding s.lock.Lock() in report() ?

db/blip_sync_context.go
```diff
@@ -655,3 +669,34 @@ func toHistory(revisions Revisions, knownRevs map[string]bool, maxHistory int) [
...
+func (s *blipSyncStats) report(bsc *BlipSyncContext) {
+	if bsc.blipContextDb == nil || bsc.blipContext == nil {
+		return
+	}
+	dbStats := bsc.blipContextDb.DbStats.Database()
+	if dbStats == nil {
+		return
+	}
+	s.lock.Lock()
+	defer s.lock.Unlock()
```
adamcfraser: Should we recheck s.lastReportTime while holding the lock, to skip the update if we're racing with another goroutine to update in the same reporting window?

db/blip_sync_context.go
```diff
@@ -655,3 +669,34 @@ func toHistory(revisions Revisions, knownRevs map[string]bool, maxHistory int) [
 	}
 	return history
 }
+
+// reportIfTimeElapsed will report stats if enough time has passed since the previous report.
+func (s *blipSyncStats) reportIfTimeElapsed(bsc *BlipSyncContext) {
```
adamcfraser: blipSyncStats only exists as a property on BlipSyncContext, but then all methods on blipSyncStats require us to pass *BlipSyncContext as a parameter.  Would it make more sense to just define these as methods on *BlipSyncContext?

db/blip_sync_context.go
```diff
@@ -655,3 +669,34 @@ func toHistory(revisions Revisions, knownRevs map[string]bool, maxHistory int) [
...
+	currentTime := time.Now().UnixMilli()
+	if (currentTime - s.lastReportTime.Load()) > bsc.blipContextDb.BlipStatsReportingInterval {
```
adamcfraser: Should we be initializing lastReportTime to currentTime when blipSyncContext/blipSyncStats is instantiated, to avoid reportIfTimeElapsed reporting stats on the first message received?

db/database.go
```diff
@@ -132,6 +134,7 @@ type DatabaseContext struct {
 	MetadataKeys                 *base.MetadataKeys             // Factory to generate metadata document keys
 	RequireResync                base.ScopeAndCollectionNames   // Collections requiring resync before database can go online
 	CORS                         *auth.CORSConfig               // CORS configuration
+	BlipStatsReportingInterval   int64                          // interval to report blip stats in milliseconds
```
adamcfraser: This feels like it should live in DatabaseContextOptions, what do you think?

adamcfraser (changes requested):
General approach looks good, a few questions on the details.

db/blip_sync_context.go
torcolvin: When the code had a context cancellation, I felt this was more important, I reverted this now.

db/active_replicator.go
torcolvin: I believe this was done intentional, confirming this with @bbrks and I've added a comment. The idea is that we want to drop collection or request or cancellation information when starting a replication. That said, I'm not sure why we don't set this earlier in the replication and use `arc.ctx` everywhere.

base/stats.go
torcolvin: Fixed.

db/blip_sync_context.go
torcolvin: done

db/blip_sync_context.go
torcolvin: done

db/database.go
torcolvin: Added this to context options.

db/blip_sync_context.go
torcolvin: This is the only part of the PR that I'm not 100% sure I implemented correctly.

db/blip_sync_context.go
torcolvin: This is the only part of the PR that I'm not 100% sure I implemented correctly.

db/blip_sync_context.go
```diff
@@ -669,33 +671,44 @@ func toHistory(revisions Revisions, knownRevs map[string]bool, maxHistory int) [
...
 	if dbStats == nil {
 		return
 	}
-	s.lock.Lock()
-	defer s.lock.Unlock()
+	bsc.stats.lock.Lock()
+	defer bsc.stats.lock.Unlock()
+
+	// a goroutine has already updated these stats
+	if !bsc.timeElapsedForStatsReporting(currentTime) {
```
adamcfraser: Can this check just use time.Now instead of currentTime?  Consider a case where many goroutines spawn with currentTime=t and queue up to acquire the lock.  At t+BlipStatsReportingInterval, all of those goroutines are going to try to update stats, even though the stats were just updated by the previous goroutine.

It feels like reportStats actually just needs to pass in a flag like 'forceUpdate bool' that bypasses the time check (that would be used on close).  With that approach I think you could get rid of reportIfTimeElapsed and just incorporate that logic into reportStats.

db/active_replicator.go
adamcfraser: I still don't follow why we're changing the context we pass to NewBlipSyncContext here.  Is this something that's required for this PR, or just cleanup?  The functional change is relatively minor - I think it just means that the context used [here](https://github.com/couchbase/sync_gateway/pull/6229/commits/41bb530143fc01ae8c9b93981fd160265bbd91d6#diff-4fc7767cce3963e7ae66f5dbd700c6d8d089f97a5102631255f7ca9f5a1a1941L64) is going to be the logging context and not arc.ctx - but I'm not sure why we're making that change.

adamcfraser (changes requested):
Left two comments on things that don't look quite right.

adamcfraser (dismissed):
Left two comments on things that don't look quite right.

adamcfraser (approved):

PR #6231, opened 2023-05-09
Update scopes documentation

torcolvin:
Fix some API documentation.

### DIFF
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -1090,7 +1090,7 @@ CollectionConfig:
     sync:
       description: The Javascript function that newly created documents in this collection are ran through.
       type: string
-      example: 'function(doc){channel(doc.channels);}'
+      example: 'function(doc){channel("collection name");}'
     import_filter:
       description: |-
         This is the function that all imported documents in this collection are ran through in order to filter out what to import and what not to import. This allows you to control what is made available to Couchbase Mobile clients. If it is not set, then no documents are filtered when imported.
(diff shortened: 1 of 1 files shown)


docs/api/components/schemas.yaml
```diff
@@ -1165,7 +1165,9 @@ Database:
       description: The name of the database.
       type: string
     sync:
-      description: The Javascript function that newly created documents are ran through for the _default scope and collection.
+      description: The Javascript function that newly created documents are ran through for the default scope and collection.
+
+        If `scopes` parameter is set, this is ignored.
```
bbrks, line 1170: Should we add SG config validation to cover this case - to explicitly disallow it?

docs/api/components/schemas.yaml
```diff
@@ -1203,9 +1205,11 @@ Database:
       default: 16
     import_filter:
       description: |-
-        This is the function that all imported documents in the _default scope and collection are ran through in order to filter out what to import and what not to import. This allows you to control what is made available to Couchbase Mobile clients. If it is not set, then no documents are filtered when imported.
+        This is the function that all imported documents in the default scope and collection are ran through in order to filter out what to import and what not to import. This allows you to control what is made available to Couchbase Mobile clients. If it is not set, then no documents are filtered when imported.
 
         `import_docs` must be true to make this field applicable.
+
+        If `scopes` parameter is set, this is ignored.
```
bbrks, line 1212: Ditto

docs/api/components/schemas.yaml
torcolvin: Already done in https://github.com/couchbase/sync_gateway/blob/ae024a247c0ee62a5598121843e9904084ac15e1/rest/config.go#L852

docs/api/components/schemas.yaml
torcolvin: Already done in https://github.com/couchbase/sync_gateway/blob/ae024a247c0ee62a5598121843e9904084ac15e1/rest/config.go#L855

bbrks (approved):

PR #6235, opened 2023-05-11
CBG-2938: Ignore Cbgt EOF feed errors when intentionally stopped

bbrks:
CBG-2938

- Changed `CbgtContext.StopHeartbeatListener` to `CbgtContext.Stop`
- Made event handler context cancellable from `Stop`
- Made `OnFeedError` callback check context for cancellation before logging about EOF

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1776/
  - `TestResyncUsingDCPStreamForNamedCollection` flake (unrelated)

### DIFF
--- a/base/dcp_sharded.go
+++ b/base/dcp_sharded.go
@@ -53,6 +53,8 @@ type CbgtContext struct {
 	heartbeater       Heartbeater              // Heartbeater used for failed node detection
 	heartbeatListener *importHeartbeatListener // Listener subscribed to failed node alerts from heartbeater
 	eventHandlers     *sgMgrEventHandlers      // Event handler callbacks
+	ctx               context.Context          // Log context
+	dbName            string                   // Database name
 }
 
 // StartShardedDCPFeed initializes and starts a CBGT Manager targeting the provided bucket.
@@ -352,6 +354,8 @@ func initCBGTManager(ctx context.Context, bucket Bucket, spec BucketSpec, cfgSG 
 		Manager:       mgr,
 		Cfg:           cfgSG,
 		eventHandlers: eventHandlers,
+		ctx:           ctx,
+		dbName:        dbName,
 	}
 
 	if spec.Auth != nil || (spec.Certpath != "" && spec.Keypath != "") {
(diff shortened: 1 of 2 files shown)


db/import_listener.go
```diff
@@ -214,7 +214,7 @@ func (il *importListener) ImportFeedEvent(event sgbucket.FeedEvent) {
 func (il *importListener) Stop() {
 	if il != nil {
 		if il.cbgtContext != nil {
-			il.cbgtContext.StopHeartbeatListener()
+			il.cbgtContext.Stop()
```
torcolvin, line 217: does Stop() stop both?

base/dcp_sharded.go
```diff
@@ -741,8 +748,11 @@ func (meh *sgMgrEventHandlers) OnFeedError(srcType string, r cbgt.Feed, err erro
 		// the connectivity problems either during the next rebalance
 		// (new kv node after failover-recovery rebalance) or
 		// on the next janitor work cycle(ephemeral network issue to the same node).
-		if strings.Contains(err.Error(), "EOF") {
-			InfofCtx(meh.ctx, KeyDCP, "Handling EOF on cbgt feed - notifying manager to trigger reconnection to feed.  indexUUID: %v, err: %v", indexUUID, err)
+		if strings.Contains(feedErr.Error(), "EOF") {
+			// If this wasn't an intentional close, log about the EOF
+			if meh.ctx.Err() != context.Canceled {
```
torcolvin, line 769: slightly prefer errors.Is in case that error gets wrapped

torcolvin (commented):
LGTM otherwise

db/import_listener.go
bbrks: yep

base/dcp_sharded.go
bbrks: I'm ok switching this to `!errors.Is(meh.ctx.Err(), context.Canceled)` as it does still work, but errors from `context.Err()` _can't_ be wrapped.

`context` (in 1.20+) has its own way of "wrapping" that still results in `Err()` only returning one of:
- `nil`
- `context.Canceled`
- `context.DeadlineExceeded`

And `errors.Is` still doesn't really support `Err()` as you'd expect it to.

example:
https://go.dev/play/p/XE9rdrP7F9A

torcolvin: This to leave as is fine, it was just a automatic reaction to seeing an error comparison.

base/dcp_sharded.go
```diff
@@ -439,8 +442,11 @@ func getMinNodeVersion(cfg cbgt.Cfg) (*ComparableVersion, error) {
 	return minVersion, nil
 }
 
-// StopHeartbeatListener unregisters the listener from the heartbeater, and stops it.
-func (c *CbgtContext) StopHeartbeatListener() {
+// Stop unregisters the listener from the heartbeater, and stops it and associated handlers.
+func (c *CbgtContext) Stop() {
```
adamcfraser, line 450: I agree it makes sense to have a more generic CbgtContext.Stop(), but if we're doing that, the rest of the 'on stop' code from importListener.Stop should get moved there as well.

adamcfraser (changes requested):

adamcfraser (approved):

PR #6242, opened 2023-05-12
CBG-2944: Ensure proveAttachments works for v2 attachments with a v2 replication protocol

bbrks:
CBG-2944

- Ensures that v2 attachments are found for `proveAttachments` by getting doc ID
- Not handling case for v3 replication protocol as `proveAttachments` isn't used (attachments are document scoped)
- Adds BlipTesterClient enhancements to run `proveAttachments` under v2 replication for known attachments

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1778/

### DIFF
--- a/db/blip_sync_stats.go
+++ b/db/blip_sync_stats.go
@@ -141,6 +141,7 @@ func BlipSyncStatsForCBL(dbStats *base.DbStats) *BlipSyncStats {
 	blipStats.HandleChangesResponseCount = dbStats.CBLReplicationPull().RequestChangesCount
 	blipStats.HandleChangesResponseTime = dbStats.CBLReplicationPull().RequestChangesTime
 	blipStats.HandleChangesSendRevCount = dbStats.CBLReplicationPull().RevSendCount
+	blipStats.SendRevErrorTotal = dbStats.CBLReplicationPull().RevErrorCount
 	blipStats.HandleChangesSendRevLatency = dbStats.CBLReplicationPull().RevSendLatency
 	blipStats.HandleChangesSendRevTime = dbStats.CBLReplicationPull().RevProcessingTime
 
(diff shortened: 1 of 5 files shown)


db/blip_sync_stats.go
```diff
@@ -141,6 +141,7 @@ func BlipSyncStatsForCBL(dbStats *base.DbStats) *BlipSyncStats {
 	blipStats.HandleChangesResponseCount = dbStats.CBLReplicationPull().RequestChangesCount
 	blipStats.HandleChangesResponseTime = dbStats.CBLReplicationPull().RequestChangesTime
 	blipStats.HandleChangesSendRevCount = dbStats.CBLReplicationPull().RevSendCount
+	blipStats.SendRevErrorTotal = dbStats.CBLReplicationPull().RevErrorCount
```
adamcfraser: For consistency, can we move this up with the other SendRev stats (after line 131)?  I realize there's some ambiguity but it looks like the `SendRev*` stats are the ones that are set in sendRevision, and the `HandleChanges*` stats are incremented in handleChangesResponse.

adamcfraser (changes requested):
Looks good - one minor formatting nit.

One question about this comment on the PR, though:
"Not handling case for v3 replication protocol as proveAttachments isn't used (attachments are document scoped)"

Would ISGR running the v3 replication protocol still send proveAttachment for a legacy attachment?  My reading of the code is that it would just fall back to getAttachment in that case after getting an AttachmentNotFound error, but would like a sanity check on that.

adamcfraser (changes requested):
Looks good - one minor formatting nit.

One question about this comment on the PR, though:
"Not handling case for v3 replication protocol as proveAttachments isn't used (attachments are document scoped)"

Would ISGR running the v3 replication protocol still send proveAttachment for a legacy attachment?  My reading of the code is that it would just fall back to getAttachment in that case after getting an AttachmentNotFound error, but would like a sanity check on that.

adamcfraser (approved):

PR #6243, opened 2023-05-15
CBG-2853 Allow one-shot replications to wait for DCP to catch up on changes feed

adamcfraser:
CBG-2853

Adds requestPlus option for changes feeds. When set, changes feeds will loop until the cached sequence (via DCP) is greater than the database sequence at the time the changes request was issued.

requestPlus can be enabled for non-continuous changes requests in one of three ways:
- by setting request_plus=true on a REST API changes call
- by setting the requestPlus property to "true" on a subChanges message
- by setting "changes_request_plus":true in the database config (default=false)

The request setting is given priority - if not set on a request, the value will fall back to the database config value.

Required minor refactoring of how options.Wait was used in changes.go, to support use of requestPlus and longpoll together. No functional changes to longpoll if requestPlus is not set.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1779/

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -295,6 +295,19 @@ func (bh *blipHandler) handleSubChanges(rq *blip.Message) error {
 
 	continuous := subChangesParams.continuous()
 
+	requestPlusSeq := uint64(0)
+	// If non-continuous, check whether requestPlus handling is set for request or via database config
+	if continuous == false {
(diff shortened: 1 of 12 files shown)


bbrks (approved):
LGTM 👍

PR #6246, opened 2023-05-15
Use less or equal as a speculative fix

torcolvin:
It seems like the failure in https://jenkins.sgwdev.com/job/MasterIntegration/474/testReport/junit/github/com_couchbase_sync_gateway_rest/TestBlipStatsFastReport/ could be caused by this if the millisecond resolution happens very fast. But it seems like the machines are slower than my test machines.

I don't think this fix will cause any problems but it's a speculative fix.


adamcfraser (dismissed):

adamcfraser (approved):

adamcfraser:
Agree that this should be fine in normal cases where BlipStatsReportingInterval is greater than zero.  (and in the test usage, we want to fire on every invocation of reportStats anyway)

PR #6251, opened 2023-05-16
CBG-2905 remove cached connections when bucket disappear

torcolvin:
CBG-2905 

If a bucket disappears, then remove the cached connections. Within a ServerContext, `CouchbaseCluster.GetConfigBuckets` gets called 1/sec from https://github.com/couchbase/sync_gateway/blob/4fe200cf387419034ec250ee614f9e760900710a/rest/config.go#L1519

So with `import_docs=false`, to avoid CBG-2983, you'll now see this logged once and it will fail.

Because of the nature of `CouchbaseCluster`, it seemed necessary to write this in a thread safe manner, but I am not sure that it is ever called concurrently right now. The other consumers might be the metadata registry.

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1787/

### DIFF
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -53,11 +53,11 @@ type CouchbaseCluster struct {
 	clusterOptions          gocb.ClusterOptions
 	forcePerBucketAuth      bool // Forces perBucketAuth authenticators to be used to connect to the bucket
 	perBucketAuth           map[string]*gocb.Authenticator
-	bucketConnectionMode    BucketConnectionMode     // Whether to cache cluster connections
-	cachedClusterConnection *gocb.Cluster            // Cached cluster connection, should only be used by GetConfigBuckets
-	cachedBucketConnections map[string]*cachedBucket // Per-bucket cached connections
-	cachedConnectionLock    sync.Mutex               // mutex for access to cachedBucketConnections
-	configPersistence       ConfigPersistence        // ConfigPersistence mode
+	bucketConnectionMode    BucketConnectionMode    // Whether to cache cluster connections
+	cachedClusterConnection *gocb.Cluster           // Cached cluster connection, should only be used by GetConfigBuckets
+	cachedBucketConnections cachedBucketConnections // Per-bucket cached connections
+	cachedConnectionLock    sync.Mutex              // mutex for access to cachedBucketConnections
+	configPersistence       ConfigPersistence       // ConfigPersistence mode
 }
 
 type BucketConnectionMode int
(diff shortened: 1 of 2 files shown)


base/bootstrap.go
```diff
@@ -238,9 +236,20 @@ func (cc *CouchbaseCluster) GetConfigBuckets() ([]string, error) {
 		return nil, err
 	}
 
+	cachedBuckets := make(map[string]struct{})
```
adamcfraser: The name of this variable seems misleading - isn't this actually something like 'serverBuckets' or 'currentBuckets', as it's based on GetAllBuckets?

base/bootstrap.go
```diff
@@ -70,14 +70,12 @@ const (
 )
 
 type cachedBucket struct {
-	bucket     *gocb.Bucket
-	teardownFn func()
+	bucket      *gocb.Bucket
+	teardownFn  func()
+	refcount    int
+	shouldClose bool
```
adamcfraser: Can you add a comment here to describe the interaction between shouldClose, refcount and teardownFn()?  

It takes a bit of time to work through the handling just by reading the code, a summary here would be helpful in future.

base/bootstrap.go
```diff
@@ -422,25 +431,44 @@ func (cc *CouchbaseCluster) getBucket(bucketName string) (b *gocb.Bucket, teardo
...
 	}
+	closeAndTeardown := func() {
+		refcount := cc.cachedBucketConnections[bucketName].refcount
+		if refcount <= 0 {
+			newTeardownFn()
+			return
+		}
+		// refcount is not zero because someone else is using bucket, so let's close
+		cc.cachedBucketConnections[bucketName].shouldClose = true
+	}
```
adamcfraser: Having three nested teardown functions (refcountTeardown -> closeAndTeardown -> newTeardownFn) seems a bit complex for handling `if shouldClose and refcount == 0, close the bucket and remove from cache`.  The similar naming also makes it tricky to follow exactly what's going on.  (I realize some of the bad naming existed prior to this PR)

It also makes it difficult to identify whether a particular caller is already holding cc.cachedConnectionLock.Lock().  (i.e. the refcount access on line 453 needs to be holding the lock, but I can't easily tell whether it does).

I've spent a decent chunk of time looking at this code and still am not confident there isn't a deadlock somewhere.  Any ideas for how we might simplify the amount of redirection and inline functions being used?

adamcfraser (changes requested):

base/bootstrap.go
torcolvin: I tried to clarify this by using a set.

base/bootstrap.go
torcolvin: I'm not at all sure I made this better, but I gave it a shot and I wrote a test for the locking behavior.

base/bootstrap.go
torcolvin: Added comments.

base/bootstrap.go
```diff
@@ -70,13 +70,73 @@ const (
 )
 
 type cachedBucket struct {
-	bucket     *gocb.Bucket
-	teardownFn func()
+	bucket      *gocb.Bucket // underlying bucket
+	teardownFn  func()       // teardown function which will close the gocb connection
```
adamcfraser: I think that renaming this bucketCloseFn will make it easier to parse getBucket, to know when we're talking about  closing the actual bucket, or doing the refcount based teardown.

base/bootstrap.go
```diff
@@ -70,13 +70,73 @@ const (
...
+	refcount    int          // count of how many functions are using this cachedBucket
+	shouldClose bool         // mark this cachedBucket as needing to be closed
```
adamcfraser:
```suggestion
	shouldClose bool         // mark this cachedBucket as needing to be closed when refcount=0
```

base/bootstrap.go
```diff
@@ -70,13 +70,73 @@ const (
...
-// noopTeardown is returned by getBucket when using a cached bucket - these buckets are torn down
-// when CouchbaseCluster.Close is called.
-func noopTeardown() {}
+// cahedBucketConnections is a lockable map cached buckets containing refcounts
+type cachedBucketConnections struct {
+	buckets map[string]*cachedBucket
+	lock    sync.Mutex
+}
+
+// removeOutdatedBuckets removes any buckets that aren't active from the map and closes the cluster connection.
```
adamcfraser:
```suggestion
// removeOutdatedBuckets marks any buckets that aren't active for closure.
```

base/bootstrap.go
```diff
@@ -70,13 +70,73 @@ const (
...
+		_, exists := activeBuckets[bucketName]
+		if exists {
+			continue
+		}
+		bucket.shouldClose = true
+		c._teardown(bucketName)
+	}
+}
+
+// removeOutdatedBuckets removes any buckets that aren't active from the map and closes the cluster connection.
```
adamcfraser:
```suggestion
// closeAll marks all buckets for closure.
```

base/bootstrap.go
```diff
@@ -70,13 +70,73 @@ const (
...
+func (c *cachedBucketConnections) removeOutdatedBuckets(activeBuckets Set) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	for bucketName, bucket := range c.buckets {
+		_, exists := activeBuckets[bucketName]
+		if exists {
+			continue
+		}
+		bucket.shouldClose = true
+		c._teardown(bucketName)
```
adamcfraser, line 95: This call is going to decrement the refcount.  Looking at the usage of this method in GetConfigBuckets, I don't follow where the corresponding increment of the bucket happens - should we actually be decrementing here?

I suspect this decrement is supposed to offset the fact that we initializing the bucket entry on the cache with refcount:1 (in getBucket).  However, I think it's possible for removeOutdatedBuckets to be called multiple times before the bucket is actually removed from the map.  (i.e. if there's a concurrent, very slow running config update)

Should we instead be initializing the refcount to 0, and skip decrementing when called from here? (i.e. use refcounts only to track in-flight ops using the bucket, and not this kind of administration)

adamcfraser (changes requested):
I like the new organization, it's a lot easier to follow and reason about.  A few suggestions for naming/comments, but also one possible functional issue.

base/bootstrap.go
torcolvin: It's in this usage that refCount actually goes to -1, which is a little bit weird and I guess we don't technically need to decrement. It's a good spot here.

I think we have to typically use refcount 1 when we add to the map, look at the usage in `primeBucketConnectionCache` - when you call `cc.getBucket()` it returns a bucket with a refcount of 1 until teardown happens.

adamcfraser (approved):

PR #6253, opened 2023-05-17
Fix RedactableError not satisfying the Redactor interface

bbrks:
- Add `String()` method to `RedactableError` to allow it to satisfy `Redactor` again.
- Add compile-time interface checks to type to catch future regressions.

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1785/

### DIFF
--- a/base/redactable_error.go
+++ b/base/redactable_error.go
@@ -12,13 +12,17 @@ package base
 
 import "fmt"
 
-// A redactable error can be used as a drop-in replacement for a base error (as would have been created via
-// fmt.Errorf), which has the ability to redact any sensitive user data by calling redact() on all if it's
-// stored args.
+// RedactableError is an error that can be used as a drop-in replacement for an error,
+// which has the ability to redact any sensitive data by calling redact() on all of its args.
 type RedactableError struct {
 	fmt  string
 	args []interface{}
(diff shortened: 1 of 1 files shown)


base/redactable_error.go
```diff
@@ -12,14 +12,18 @@ package base
...
 	fmt  string
 	args []interface{}
 }
 
+var (
+	_ error    = &RedactableError{}
+	_ Redactor = &RedactableError{}
+)
+
 // Create a new redactable error.  Same signature as fmt.Errorf() for easy drop-in replacement.
```
torcolvin, line 27: This code seems fine, although you can't use `%w` which is always tempting to use around errors, I'd drop a comment. If you wrap this error, I also think the redaction stops working so I worry about this use.

This code is unused though so I wonder if it's worth keeping around?

torcolvin (approved):

base/redactable_error.go
bbrks: - I agree, this type predated `%w` - we should figure out what to do about that when we have time. CBG-2997
- `RedactableError` is used ... ?

Screenshot_20230517_142811

PR #6255, opened 2023-05-18
Log around config.FromConnStr to diagnose slow DNS SRV resolution

bbrks:
`config.FromConnStr` performs SRV DNS lookups, which under some circumstances can cause slow startup times for Sync Gateway.

Log around this function to diganose and log at warning when it takes longer than 10 seconds.

## Pre-review checklist

- n/a

### DIFF
--- a/rest/server_context.go
+++ b/rest/server_context.go
@@ -41,6 +41,9 @@ const kStatsReportInterval = time.Hour
 const kDefaultSlowQueryWarningThreshold = 500 // ms
 const KDefaultNumShards = 16
 
+// fromConnStrWarningThreshold determines the amount of time it should take before we warn about parsing a connstr (mostly for DNS SRV resolution)
+const fromConnStrWarningThreshold = 10 * time.Second
+
 // defaultBlipStatsReportingInterval is the default interval when to report blip stats, at the end of a message handler.
 const defaultBlipStatsReportingInterval = 30 * time.Second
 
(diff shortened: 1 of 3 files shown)


rest/server_context.go
```diff
@@ -41,6 +41,9 @@ const kStatsReportInterval = time.Hour
 const kDefaultSlowQueryWarningThreshold = 500 // ms
 const KDefaultNumShards = 16
 
+// fromConnStrWarningThreshold determines the amount of time it should take before we warn about parsing a connstr (mostly for DNS SRV resolution)
```
torcolvin: I
```suggestion
// fromConnStrWarningThreshold determines the amount of time it should take before we warn about parsing a connstr (mostly for DNS resolution)
```
I think there's other DNS resolution and the call happens to check if SRV records are there but it can do other things.

torcolvin (commented):
I think having slow DNS is relatively common.

But we actually do this in _many_ places, usually at startup or DB startup:

https://github.com/couchbase/sync_gateway/blob/5c7673ef195c781398e117086d282ab9a5a49e6c/base/dcp_client.go#L323

And anywhere a cluster connection is established (bootstrap) https://github.com/couchbase/gocb/blob/da582ed180ca768052410105b5e4d4394f80e674/client.go#L105

I think having _any_ logging for this is fine, but it will not be comprehensive.

torcolvin (dismissed):
I don't think you need to address the comments, I think this would be helpful enough for users and it's not worth putting more time in it.

torcolvin (approved):

bbrks:
Didn't see the other method used as it was on a different type - a text search instead of AST search found it :)

PR #6256, opened 2023-05-18
CBG-2983 Close cbgt indexes on feed close

adamcfraser:
Required to trigger close of cbgt memcached connections.

CBG-2983

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1789/

### DIFF
--- a/base/dcp_sharded.go
+++ b/base/dcp_sharded.go
@@ -349,6 +349,7 @@ func initCBGTManager(ctx context.Context, bucket Bucket, spec BucketSpec, cfgSG 
 		serverURL,
 		eventHandlers,
 		options)
+	eventHandlers.manager = mgr
 
 	cbgtContext := &CbgtContext{
 		Manager:       mgr,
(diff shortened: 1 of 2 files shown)


base/dcp_sharded.go
```diff
@@ -773,4 +779,21 @@ func (meh *sgMgrEventHandlers) OnFeedError(srcType string, r cbgt.Feed, feedErr
 		}
 	}
 
+	if len(indexUUID) == 0 {
```
torcolvin, line 782: I know this code was lifted from cbft and I basically understand what it is doing, but I think some comments are worth mentioning.

Example comments: `indexUUUID = 0 means the feed is closing`

When is bucketName goign to be empty here? or manager is nil? 

nits for readability, take or leave: 

- I'd drop the nesting and do early returns here.
- I'd make `deleteIndexErr` as `err` so that we don't make future mistakes in logging since `err` is defined above in he function.

torcolvin (commented):
LGTM, just thoughts on future readability.

base/dcp_sharded.go
adamcfraser: Thanks for raising this - looking into this further I'm concerned that the in the case where the database is closed but the bucket still exists, the stats connections still aren't being closed.  Needs more investigation.

adamcfraser:
Closing this - there are some fundamental issues with this approach to closing cbgt's agents - need to revisit.

PR #6258, opened 2023-05-22
CBG-2793: attachment compaction code erroneously sets failOnRollback

gregns1:
CBG-2793

As it stands attachment comapction will fail on rollback. I have added code that if a rollback is indicated on the mark phase of the process we reset the phase and retry. 
- added test to force a rollback during a mark phase of attachment compaction and assert that the process recovers and finishes correctly. 
- Added a function in compaction code to purge the dcp metadata from the bucket for the compaction process in event of a rollback error being returned from the mark phase
- changes to a test already there (for continuous rollback handling) as the test wasn't rolling back as expected, I expect it was a side effect of recent dcp work at tail end of 3.1

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1839/


base/dcp_client.go
```diff
@@ -443,6 +446,12 @@ func (dc *DCPClient) openStream(vbID uint16, maxRetries uint32) error {
 				return fmt.Errorf("%s, failOnRollback requested", openStreamErr)
 			}
 			InfofCtx(logCtx, KeyDCP, "Open stream for vbID %d failed due to rollback or range error, will roll back metadata and retry: %v", vbID, openStreamErr)
+			if dc.attachmentCompactionClient {
```
torcolvin: I think it's conceptually easier to think about this is we make this `failOnRollback` and then make a public method to purge metadata - and handle that purge in the calling code (attachment compaction).

If we need to purge while the client is still alive, I'd create an option where you failOnRollback that does failOnRollback and purge metadata, but it seems pretty niche.

You could type the RollbackError to make it easier to find from attachment compaction if that makes it easier.

db/background_mgr_attachment_compaction.go
```diff
@@ -127,8 +128,28 @@ func (a *AttachmentCompactionManager) Run(ctx context.Context, options map[strin
...
+			}
+			// if error isn't rollback then assume its not recoverable
+			return false, err, nil
+		}
+		// retry loop for handling a rollback during mark phase of compaction process
+		err, _ = base.RetryLoop("attachmentCompactMarkPhase", worker, base.CreateMaxDoublingSleeperFunc(25, 100, 10000))
 		if err != nil || terminator.IsClosed() {
+			if errors.Is(err, base.ErrAttachmentCompactionRollback) {
+				// log warning to show we hit max number of retries
+				base.WarnfCtx(ctx, "maximum retry attempts reached on mark phase: %v", err)
```
torcolvin, line 164: I think we should log when we purge the metadata, see above that this might occur in a different spot depending on how the error is handled.

db/background_mgr_attachment_compaction.go
```diff
@@ -153,6 +174,23 @@ func (a *AttachmentCompactionManager) Run(ctx context.Context, options map[strin
 	return nil
 }
 
+func (a *AttachmentCompactionManager) newRunRollback(ctx context.Context, options map[string]interface{}) error {
```
torcolvin: Can this be replaced with rerunning `Init`?

rest/attachmentcompactiontest/attachment_compaction_api_test.go
```diff
@@ -156,6 +157,113 @@ func TestAttachmentCompactionAPI(t *testing.T) {
...
+
+		resp := rt.SendAdminRequest("GET", "/{{.db}}/_compact?type=attachment", "")
+		rest.RequireStatus(t, resp, http.StatusOK)
+
+		err := base.JSONUnmarshal(resp.BodyBytes(), &response)
+		require.NoError(t, err)
+
+		return response.State == db.BackgroundProcessStateStopped
+	})
+	require.NoError(t, err)
```
torcolvin:
```suggestion
	rt.WaitForAttachmentCompactionStatus(db.BackgroundProcessStopped)`
```

rest/attachmentcompactiontest/attachment_compaction_api_test.go
```diff
@@ -156,6 +157,113 @@ func TestAttachmentCompactionAPI(t *testing.T) {
...
+
+		resp := rt.SendAdminRequest("GET", "/{{.db}}/_compact?type=attachment", "")
+		rest.RequireStatus(t, resp, http.StatusOK)
+
+		err := base.JSONUnmarshal(resp.BodyBytes(), &response)
+		require.NoError(t, err)
+
+		return response.State == db.BackgroundProcessStateRunning
+	})
+	require.NoError(t, err)
```
torcolvin:
```suggestion
	rt.WaitForAttachmentCompactionStatus(db.BackgroundProcessRunning)`
```

rest/attachmentcompactiontest/attachment_compaction_api_test.go
```diff
@@ -156,6 +157,113 @@ func TestAttachmentCompactionAPI(t *testing.T) {
...
+		time.Sleep(1 * time.Second)
+
+		resp := rt.SendAdminRequest("GET", "/{{.db}}/_compact?type=attachment", "")
+		rest.RequireStatus(t, resp, http.StatusOK)
+
+		err = base.JSONUnmarshal(resp.BodyBytes(), &response)
+		require.NoError(t, err)
+
+		return response.State == db.BackgroundProcessStateCompleted
+	})
```
torcolvin:
```suggestion
	rt.WaitForAttachmentCompactionStatus(db.BackgroundProcessStopped)
```

I'm reasonably certain there's a test race condition if this runs too fast we'll miss the `Running` status.

db/background_mgr_attachment_compaction.go
gregns1: I could be having a serious case of Monday brain but I wasn't sure how to get the "cluster status" to pass into the Init function from here?

torcolvin: I think you can pass `nil` here and it will match the behavior you wrote.

db/attachment_compaction_test.go
```diff
@@ -239,6 +239,10 @@ func TestAttachmentCleanupRollback(t *testing.T) {
 	if base.UnitTestUrlIsWalrus() {
 		t.Skip("This test only works against Couchbase Server")
 	}
+	if base.TestsUseNamedCollections() {
+		t.Skip("This test only works with default collection")
+	}
+
 	base.SetUpTestLogging(t, base.LevelInfo, base.KeyAll)
 	var garbageVBUUID gocbcore.VbUUID = 1234
 	testDb, ctx := setupTestDB(t)
```
torcolvin: change this to:

```
	dbcOptions := DatabaseContextOptions{
		Scopes:   GetScopesOptionsDefaultCollectionOnly(t),
	}
	testDB, ctx :=  SetupTestDBWithOptions(t, dbcOptions)
```

which will force use of the default collection

db/attachment_compaction_test.go
```diff
@@ -239,14 +239,14 @@ func TestAttachmentCleanupRollback(t *testing.T) {
 	if base.UnitTestUrlIsWalrus() {
 		t.Skip("This test only works against Couchbase Server")
 	}
-	if base.TestsUseNamedCollections() {
-		t.Skip("This test only works with default collection")
+	base.SetUpTestLogging(t, base.LevelInfo, base.KeyAll)
```
torcolvin, line 242: I think it does this logging by default?

gregns1: Not all logging is included. Some of the logging around rollback will be missing when you don't include this. Given the test is testing this area I did this to include this logging.

torcolvin: Cool, sounds good.

base/dcp_client_test.go
```diff
@@ -393,17 +390,36 @@ func TestContinuousDCPRollback(t *testing.T) {
...
+
 	// Assert that the number of vBuckets active are the same as the total number of vBuckets on the client.
 	// In continuous rollback the streams should not close after they're finished.
-	numVBuckets := len(dcpClient.activeVbuckets)
-	require.Equal(t, dcpClient.numVbuckets, uint16(numVBuckets))
+	numVBuckets := len(dcpClient1.activeVbuckets)
+	require.Equal(t, dcpClient1.numVbuckets, uint16(numVBuckets))
 
-	require.NoError(t, dcpClient.Close())
+	require.NoError(t, dcpClient1.Close())
```
torcolvin: I'd move this to a

```
defer func() {
    assert.NoError(t, dcpClient1.Close())
}
```

an assert because in a defer we don't want to change this

base/dcp_client_test.go
```diff
@@ -412,13 +428,11 @@ func TestContinuousDCPRollback(t *testing.T) {
 func (dc *DCPClient) forceRollbackvBucket(uuid gocbcore.VbUUID) {
 	metadata := make([]DCPMetadata, dc.numVbuckets)
 	for i := uint16(0); i < dc.numVbuckets; i++ {
-		if i%2 == 0 {
+		if i%2 == 0 && i < 100 {
```
torcolvin: add a comment to list logic here - notably this logic seems suspicious because if you only want it to hit a few vbuckets. There are only 64 vBuckets for MacOS, but 1024 on Linux.

db/background_mgr_attachment_compaction.go
```diff
@@ -100,6 +103,25 @@ func (a *AttachmentCompactionManager) Init(ctx context.Context, options map[stri
 	return newRunInit()
 }
 
+func (a *AttachmentCompactionManager) PurgeDCPMetadata(ctx context.Context, datastore base.DataStore, database *Database, phase string) error {
+	streamName := GenerateCompactionDCPStreamName(a.CompactID, phase)
+	checkpointPrefix := fmt.Sprintf("%s:%v", "_sync:dcp_ck:", streamName)
```
torcolvin: This should use https://github.com/couchbase/sync_gateway/blob/c17d66c59f565e1cb4bed3efcf745c34bd80e7ab/base/constants_syncdocs.go#L257

db/background_mgr_attachment_compaction.go
```diff
@@ -120,15 +142,42 @@ func (a *AttachmentCompactionManager) Run(ctx context.Context, options map[strin
...
-		persistClusterStatus()
-		_, a.VBUUIDs, err = attachmentCompactMarkPhase(ctx, dataStore, collectionID, database, a.CompactID, terminator, &a.MarkedAttachments)
+		worker := func() (shouldRetry bool, err error, value interface{}) {
+			persistClusterStatus()
+			_, a.VBUUIDs, err = attachmentCompactMarkPhase(ctx, dataStore, collectionID, database, a.CompactID, terminator, &a.MarkedAttachments)
+			if errors.As(err, &rollbackErr) {
+				base.InfofCtx(ctx, base.KeyDCP, "rollback indicated on mark phase of attachment, resetting the task")
+				err = a.PurgeDCPMetadata(ctx, dataStore, database, MarkPhase)
+				if err != nil {
+					base.WarnfCtx(ctx, "error occurred during purging of dcp metadata")
```
torcolvin:
```suggestion
					base.WarnfCtx(ctx, "error occurred during purging of dcp metadata: %s", err)
```

db/background_mgr_attachment_compaction.go
```diff
@@ -120,15 +142,42 @@ func (a *AttachmentCompactionManager) Run(ctx context.Context, options map[strin
...
+					return false, err, nil
+				}
+				err = a.Init(ctx, options, nil)
+				if err != nil {
+					base.WarnfCtx(ctx, "error on initialization of new run after rollback has been indicated")
```
torcolvin:
```suggestion
					base.WarnfCtx(ctx, "error on initialization of new run after rollback has been indicated: %s", err)
```

db/background_mgr_attachment_compaction.go
```diff
@@ -142,9 +191,30 @@ func (a *AttachmentCompactionManager) Run(ctx context.Context, options map[strin
 		fallthrough
 	case "cleanup":
 		a.SetPhase("cleanup")
-		persistClusterStatus()
-		err := attachmentCompactCleanupPhase(ctx, dataStore, collectionID, database, a.CompactID, a.VBUUIDs, terminator)
+		worker := func() (shouldRetry bool, err error, value interface{}) {
+			persistClusterStatus()
+			err = attachmentCompactCleanupPhase(ctx, dataStore, collectionID, database, a.CompactID, a.VBUUIDs, terminator)
+			if errors.As(err, &rollbackErr) || errors.Is(err, base.ErrVbUUIDMismatch) {
```
torcolvin: why does this check 

```
if errors.As(err, &rollbackErr) || errors.Is(err, base.ErrVbUUIDMismatch)
```

in cleanup but not in mark or sweep?

db/attachment_compaction_test.go
```diff
@@ -234,6 +235,90 @@ func TestAttachmentCleanup(t *testing.T) {
 
 }
 
+func TestAttachmentCleanupRollback(t *testing.T) {
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("This test only works against Couchbase Server")
```
torcolvin:
```suggestion
		t.Skip("This test only works against Couchbase Server since it requires DCP")
```

torcolvin (commented):
Don't we need to change something in https://github.com/couchbase/sync_gateway/blob/c17d66c59f565e1cb4bed3efcf745c34bd80e7ab/db/attachment_compaction.go#L556

db/background_mgr_attachment_compaction.go
gregns1: Added it to the mark phase. We don't have rollback handling on the sweep phase, if rollback is hit here user needs to manually reset the process.

db/background_mgr_attachment_compaction.go
```diff
@@ -100,6 +103,25 @@ func (a *AttachmentCompactionManager) Init(ctx context.Context, options map[stri
 	return newRunInit()
 }
 
+func (a *AttachmentCompactionManager) PurgeDCPMetadata(ctx context.Context, datastore base.DataStore, database *Database, phase string) error {
```
torcolvin: can we have this just take a DCPClient which you'd return from `attachmentCompactMarkPhase` etc?

I'm trying to make this easier to read. If this doesn't work, that's okay, but we could call a function without having to worry about reconstructing the prefix that seems preferable.

db/background_mgr_attachment_compaction.go
torcolvin: I'd be inclined to put the error handling into a function because we reuse for mark and cleanup, like:

handleAttachmentCompactionRollbackError(ctx, DCPClient, err) (shouldRetry, err)

torcolvin (commented):
It looks good, I'm trying to see if I can make the readability any better but if my comments don't make sense or don't work, we can leave as is.

torcolvin (commented):
It looks good, I'm trying to see if I can make the readability any better but if my comments don't make sense or don't work, we can leave as is.

torcolvin (commented):
It looks good, I'm trying to see if I can make the readability any better but if my comments don't make sense or don't work, we can leave as is.

db/background_mgr_attachment_compaction.go
gregns1: I've addressed this issue by extracting the key prefix from the client in the code when the client is created. Then returning that in the phases that we need to handle rollback on and using that to pass into this function.

torcolvin (approved):

PR #6260, opened 2023-05-23
CBG-2977 allow DELETE on a broken DB config

torcolvin:
CBG-2977

- This only works with a bootstrap connection, because otherwise you'd just modify the file on disk. This means that this _only_ has test coverage where we use persistent config enabled. Also, persistent config isn't enabled for walrus.
- Added some boilerplate to allow `LeakyBucket` and `TestBucket` to support `DropDataStore` and `CreateDataStore` on the assumption this test would work with walrus. It does not though, because it relies on the bootstrap connection.
- Trying to delete a config without a bootstrap connection will result in a 404 since you wouldn't be able start up sync gateway with a bad config.
- the only non test changes are in `rest/config.go` and `rest/handler.go`. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1814/

### DIFF
--- a/rest/handler.go
+++ b/rest/handler.go
@@ -321,13 +321,14 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 
 	var dbContext *db.DatabaseContext
 
+	var bucketName string
+
 	// look up the database context:
 	if keyspaceDb != "" {
 		h.addDatabaseLogContext(keyspaceDb)
 		var err error
 		if dbContext, err = h.server.GetActiveDatabase(keyspaceDb); err != nil {
 			if err == base.ErrNotFound {
-
 				if shouldCheckAdminAuth {
 					// Check if authenticated before attempting to get inactive database
 					authorized, err := h.checkAdminAuthenticationOnly()
@@ -342,7 +343,7 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 				dbContext, dbConfigFound, err = h.server.GetInactiveDatabase(h.ctx(), keyspaceDb)
 				if err != nil {
 					if httpError, ok := err.(*base.HTTPError); ok && httpError.Status == http.StatusNotFound {
-						if shouldCheckAdminAuth {
+						if shouldCheckAdminAuth && (!h.allowNilDBContext || !dbConfigFound) {
 							return base.HTTPErrorf(http.StatusForbidden, "")
 						} else if h.privs == regularPrivs || h.privs == publicPrivs {
 							if !h.providedAuthCredentials() {
@@ -356,6 +357,7 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 						base.InfofCtx(h.ctx(), base.KeyHTTP, "Error trying to get db %s: %v", base.MD(keyspaceDb), err)
 						return err
 					}
+					bucketName, _ = h.server.bucketNameFromDbName(keyspaceDb)
 				}
 			} else {
 				return err
@@ -422,7 +424,6 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 			}
 		}
 	}
-
 	if shouldCheckAdminAuth {
 		// If server is walrus but auth is enabled we should just kick the user out as invalid as we have nothing to
 		// validate credentials against
@@ -449,13 +450,12 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 			authScope = dbContext.Bucket.GetName()
 		} else {
 			managementEndpoints, httpClient, err = h.server.ObtainManagementEndpointsAndHTTPClient()
-			authScope = ""
+			authScope = bucketName
 		}
 		if err != nil {
 			base.WarnfCtx(h.ctx(), "An error occurred whilst obtaining management endpoints: %v", err)
 			return base.HTTPErrorf(http.StatusInternalServerError, "")
 		}
-
 		if h.authScopeFunc != nil {
 			body, err := h.readBody()
 			if err != nil {
--- a/rest/handler.go
+++ b/rest/handler.go
@@ -1,4 +1,4 @@
-//  Copyright 2012-Present Couchbase, Inc.
+///  Copyright 2012-Present Couchbase, Inc.
 //
 //  Use of this software is governed by the Business Source License included
 //  in the file licenses/BSL-Couchbase.txt.  As of the Change Date specified
@@ -101,7 +101,7 @@ type handler struct {
 	formattedSerialNumber string
 	loggedDuration        bool
 	runOffline            bool       // allows running on an offline database
-	allowMetadataDBOnly   bool       // allow acceess to a database based only on name, looking up in metadata registry
+	allowNilDBContext     bool       // allow acceess to a database based only on name, looking up in metadata registry
 	queryValues           url.Values // Copy of results of rq.URL.Query()
 	permissionsResults    map[string]bool
 	authScopeFunc         authScopeFunc
@@ -148,8 +148,8 @@ func makeOfflineHandler(server *ServerContext, privs handlerPrivs, accessPermiss
 func makeMetadataDBOfflineHandler(server *ServerContext, privs handlerPrivs, accessPermissions []Permission, responsePermissions []Permission, method handlerMethod) http.Handler {
 	return http.HandlerFunc(func(r http.ResponseWriter, rq *http.Request) {
 		options := handlerOptions{
-			runOffline:          true,
-			allowMetadataDBOnly: true,
+			runOffline:        true,
+			allowNilDBContext: true,
 		}
 		h := newHandler(server, privs, r, rq, options)
 		err := h.invoke(method, accessPermissions, responsePermissions)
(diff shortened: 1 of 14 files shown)


rest/utilities_testing.go
```diff
@@ -2457,6 +2457,8 @@ func (rt *RestTester) GetChangesOneShot(t testing.TB, keyspace string, since int
 }
 
 func (rt *RestTester) NewDbConfig() DbConfig {
+	// make sure bucket has been initialized
+	_ = rt.Bucket()
```
adamcfraser: Isn't this going to happen on line 2464 already?

rest/handler.go
```diff
@@ -330,8 +330,12 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 							return ErrInvalidLogin
 						}
 					}
-					base.InfofCtx(h.ctx(), base.KeyHTTP, "Error trying to get db %s: %v", base.MD(keyspaceDb), err)
-					return err
+					// look for db in config registry
```
adamcfraser: I'm assuming this is here so that we don't error here before we can get to the custom handling in deleteDB - is that correct?  Is this safe for all other calls?

Looking at the handling immediately below, for example - it looks like we're not going to perform the correct offline checks for a db-scoped request that's in this situation.  Further down, checkAuth will no-op if the dbContext is nil.  Where do we guard that non-deleteDB requests (a) don't eventually panic on a nil db ctx, (b) don't leak information via inconsistent error codes, or (c) succeed when they shouldn't.

adamcfraser (changes requested):

rest/handler.go
torcolvin: The answer is that I'm not sure about this.

Here's what I do know:
- jenkins integration tests continue to pass, but they largely don't use a bootstrap connection
- we almost always check `h.db == nil` on spot knowledge.

I think this is the right long term fix, but we don't have good test coverage for it.

We could limit this allowance in this handler whether to the request is admin and only allow this for deleting a db (and maybe upserting a config). We do check admin auth here https://github.com/couchbase/sync_gateway/blob/master/rest/handler.go#L310

rest/handler.go
```diff
@@ -146,25 +162,30 @@ func makeOfflineHandler(server *ServerContext, privs handlerPrivs, accessPermiss
...
+		h := newHandler(server, privs, r, rq, handlerOptions{})
 		h.authScopeFunc = dbAuthStringFunc
 		err := h.invoke(method, accessPermissions, responsePermissions)
 		h.writeError(err)
 		h.logDuration(true)
 	})
 }
 
-func newHandler(server *ServerContext, privs handlerPrivs, r http.ResponseWriter, rq *http.Request, runOffline bool) *handler {
+type handlerOptions struct {
```
adamcfraser, line 173: I really like this approach to managing options for future expansion, good choice.

rest/handler.go
```diff
@@ -330,8 +351,17 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 							return ErrInvalidLogin
 						}
 					}
-					base.InfofCtx(h.ctx(), base.KeyHTTP, "Error trying to get db %s: %v", base.MD(keyspaceDb), err)
-					return err
+					if h.allowMetadataDBOnly {
```
adamcfraser: It looks like the code won't make it this far when admin auth is enabled (which will be most use cases) - it will return error on line 345.  This seems like an gap with this approach in general - we need the dbcontext down on line 453 to properly set the authScope for the permission check of whether the user has the PermDeleteDb permission for the bucket.
We handle a similar case for createDb using authScopeFunc - it seems like we could do something similar here.  For that matter, deleteDb doesn't seem to actually need the database context - do we want to follow the same pattern as createDb and use a different path name to bypass the dbcontext initialization in the first place?

rest/upgradetest/remove_collection_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/couchbase/sync_gateway/base"
+	"github.com/couchbase/sync_gateway/rest"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestRemoveCollection tests when a collection has been removed from CBS, and the server is restarted. We should be able to modify or delete the database.
+func TestRemoveCollection(t *testing.T) {
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("test relies on boostrap connection and needs CBS")
```
adamcfraser:
```suggestion
		t.Skip("test relies on bootstrap connection and needs CBS")
```

rest/handler.go
```diff
@@ -330,8 +351,17 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
...
+						// look for db in config registry
+						_, ok := h.server.bucketNameFromDbName(keyspaceDb)
```
adamcfraser: There's some redundant work here - GetInactiveDatabase would have just done a fetchDatabase in the case where it didn't find the loaded db, and only returns 'not found' when the associated applyConfig fails, so we don't need another scan over all the buckets looking for the db config.  Seems like we could modify GetInactiveDatabase to return another flag for 'dbConfigExists' that we could check here instead.

rest/handler.go
```diff
@@ -146,25 +162,30 @@ func makeOfflineHandler(server *ServerContext, privs handlerPrivs, accessPermiss
...
 		err := h.invoke(method, accessPermissions, responsePermissions)
 		h.writeError(err)
 		h.logDuration(true)
 	})
 }
 
-func newHandler(server *ServerContext, privs handlerPrivs, r http.ResponseWriter, rq *http.Request, runOffline bool) *handler {
+type handlerOptions struct {
+	runOffline          bool // if true, allow handler to run when a database is offline
+	allowMetadataDBOnly bool // if true, allow handler to access a database without a valid dbcontext
```
adamcfraser: I think allowNilDbContext might be a better name for this parameter, as it lines up with your comment.  And then have the comment be something like "// if true, allow a db-scoped handler to be invoked with a nil dbContext in cases where the database config exists but has an error preventing dbContext initialization"

rest/server_context.go
```diff
@@ -377,18 +378,18 @@ func (sc *ServerContext) ReloadDatabaseWithConfig(nonContextStruct base.NonCance
...
-	_, err := sc._getOrAddDatabaseFromConfig(ctx, config, false, db.GetConnectToBucketFn(failFast))
+	_, err := sc._getOrAddDatabaseFromConfig(ctx, config, false, db.GetConnectToBucketFn(failFast), failFast)
 	return err
 }
 
 // Adds a database to the ServerContext.  Attempts a read after it gets the write
 // lock to see if it's already been added by another process. If so, returns either the
 // existing DatabaseContext or an error based on the useExisting flag.
-func (sc *ServerContext) getOrAddDatabaseFromConfig(ctx context.Context, config DatabaseConfig, useExisting bool, openBucketFn db.OpenBucketFn) (*db.DatabaseContext, error) {
+func (sc *ServerContext) getOrAddDatabaseFromConfig(ctx context.Context, config DatabaseConfig, useExisting bool, openBucketFn db.OpenBucketFn, failFast bool) (*db.DatabaseContext, error) {
```
adamcfraser: I think we're using openBucketFn exclusively to determine whether to fail fast on that operation, so adding a separate failFast flag seems potentially redundant. I think the motivation is the same for both cases - whether we should retry on NotFound when (a) the bucket doesn't exist, and (b) a collection doesn't exist.

We only ever invoke this with GetConnectToBucketFn for openBucketFn for non-test usages, so for those cases it would be preferable to just pass the failFast flag instead of an openBucketFn AND failFast flag.

I understand not wanting to bloat this PR, but it's a bit ugly that every non-test invocation of this function looks like `db.GetConnectToBucketFn(failFast), failFast`

adamcfraser (changes requested):
I don't think the current approach works with admin auth enabled, added some suggestions about how we might address.

rest/handler.go
torcolvin: I fixed and added tests for this, I think.

rest/server_context.go
```diff
@@ -482,11 +496,15 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
 	// Connect to bucket
 	base.InfofCtx(ctx, base.KeyAll, "Opening db /%s as bucket %q, pool %q, server <%s>",
 		base.MD(dbName), base.MD(spec.BucketName), base.SD(base.DefaultPool), base.SD(spec.Server))
-	bucket, err := openBucketFn(ctx, spec)
+
+	// the connectToBucketFn is used for testing seam
```
adamcfraser, line 500: I think it's a better to avoid mutating options in the function they are being passed to - it makes it harder to debug what the incoming value was.  
In this case, we don't have anywhere else we use options.ConnectToBucketFn, so could just have something like:
```
var bucket, err
if options.connectToBucketFn != nil {
   bucket, err = options.connectToBucketFn(...)
 } else {
    bucket, err = db.ConnectToBucket(...)
 }
 ```
 If you think there's a future use case to call ConnectToBucket again within the lifetime of the database context, I'd store the connect function on the database context and populate here with the same kind of if/else.

adamcfraser (changes requested):
One request on how we use options, otherwise looks good.

rest/server_context.go
torcolvin: I think this is a good idea, fixed.

adamcfraser (approved):

PR #6261, opened 2023-05-25
CBG-3024 Make sure CE import uses checkpoints

torcolvin:
CBG-3024

This test works with CE, and fails before. I tried to close this database and restart a new database in this test (which I've done in a RestTester persistent config with a matching groupid), so I could make sure it resumed from the correct point.

I could add an assertion to make sure the DCP checkpoint is persisted?

## Pre-review checklist

- [x] jenkins EE https://jenkins.sgwdev.com/job/SyncGateway-Integration/1804/

Failed jenkins test run: https://jenkins.sgwdev.com/job/SyncGateway-Integration/1800/ this doesn't touch any CE code.

Manual test of this test only in CE/EE works.


db/change_listener_test.go
```diff
@@ -133,3 +133,41 @@ func TestUserWaiterForRoleChange(t *testing.T) {
...
+
+	doc2 := "doc2"
+	revID2, _, err := collection.Put(base.TestCtx(t), doc2, Body{"key": "value"})
+	require.NoError(t, err)
+	_, ok := base.WaitForStat(func() int64 {
+		return db.DbStats.Database().DCPReceivedCount.Value()
+	}, 1)
+	require.True(t, ok)
+
+	_, exists := collection.revisionCache.Peek(base.TestCtx(t), doc2, revID2)
```
adamcfraser: I'm not sure that checking the rev cache is directly relevant, as it's the channel cache that gets populated by the caching feed - I'd probably leave this test to just check the DCP counts.

adamcfraser (changes requested):
Looks good - just a minor comment on the test.

db/change_listener_test.go
torcolvin: The problem with testing the stats is that you want to make sure it never reaches 2 - so we want to make sure doc2 comes over the feed but not doc1. It's hard to test that the stat doesn't reach a value - you don't know how long to wait.

adamcfraser (approved):

PR #6262, opened 2023-05-25
CBG-2929 add HTTPBytesWritten stat for public API

torcolvin:
CBG-2929

Created a custom http.ResponseWriter and log the DB scoped requests over the public and guest APIs.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1928/


base/stats.go
```diff
@@ -461,6 +461,8 @@ type DatabaseStats struct {
 	DocWritesXattrBytes *SgwIntStat `json:"doc_writes_xattr_bytes"`
 	// Highest sequence number seen on the caching DCP feed.
 	HighSeqFeed *SgwIntStat `json:"high_seq_feed"`
+	// Number of bytes written over public interface for REST api
```
adamcfraser: This looks like it's just a subset of CBG-2929 (doesn't include request bytes over REST, and doesn't include blip).  If it makes sense to split into separate PRs, creating subtickets under CBG-2929 seems like a good idea to make sure nothing gets missed.

rest/counted_response_writer.go
```diff
@@ -0,0 +1,61 @@
...
+// CountableResponseWriter is an interface that any custom http.ResponseWriter used by Sync Gateway handlers need to implement.
+type CountableResponseWriter interface {
+	http.ResponseWriter
+	GetBytesWritten() int64
+}
+
+// http.ResponseWriter wrapper that counts the number of bytes written in a response. This ignores the bytes written in headers.
+type CountedResponseWriter struct {
+	writer       http.ResponseWriter
+	bytesWritten int64
```
adamcfraser: Would it be more efficient and/or generate less GC to pass the stat into CountedResponseWriter, and just update the stat directly instead of maintaining a separate int64 on the CountedResponseWriter?

rest/handler.go
```diff
@@ -557,6 +559,20 @@ func (h *handler) logRequestLine() {
...
 
+// logBytesWritten only logs database level bytes written over the public API
+func (h *handler) logBytesWritten() {
+	if h.db == nil {
+		return
+	}
+	if h.privs != publicPrivs && h.privs != regularPrivs {
+		return
+	}
+	if h.isBlipSync() {
```
adamcfraser, line 603: How are we handling bytes associated with streaming non-blip _changes requests (0feed=continuous or feed=websocket)?

adamcfraser (changes requested):

base/stats.go
torcolvin: These are in https://github.com/couchbase/sync_gateway/commit/a31be806b47b07cb9e3d7fdfaaaa8ec024d2f0a8

rest/handler.go
torcolvin: Should be yes if they are over public API.

rest/counted_response_writer.go
torcolvin: I sort of did this and not, the problem is we have to create a the response writer before we know if the database exists or not, so we may or may not be doing anything with this value.

adamcfraser: If this implies that we're associating a CountedResponseWriter with a request even if it's not a db-scoped request, that doesn't seem ideal.  What's the reason it needs to be initialized that early, as opposed to doing it once we've set the db on the handler?

rest/server_context.go
```diff
@@ -41,8 +41,8 @@ const kStatsReportInterval = time.Hour
 const kDefaultSlowQueryWarningThreshold = 500 // ms
 const KDefaultNumShards = 16
 
-// defaultBlipStatsReportingInterval is the default interval when to report blip stats, at the end of a message handler.
-const defaultBlipStatsReportingInterval = 30 * time.Second
+// defaultBytesStatsReportingInterval is the default interval when to report blip stats, at the end of a message handler.
```
adamcfraser:
```suggestion
// defaultBytesStatsReportingInterval is the default interval when to report request bytes transferred stats
```

rest/handler.go
```diff
@@ -294,11 +303,11 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 		return base.HTTPErrorf(http.StatusUnsupportedMediaType, "Unsupported Content-Encoding; use gzip")
 	}
 
-	if h.shouldShowProductVersion() {
-		h.setHeader("Server", base.VersionString)
-	} else {
-		h.setHeader("Server", base.ProductNameString)
-	}
+	//if h.shouldShowProductVersion() {
```
adamcfraser: I'm not sure what's going on with this change - is this intentional?

rest/handler.go
```diff
@@ -582,6 +592,30 @@ func (h *handler) logRequestLine() {
 	base.InfofCtx(h.ctx(), base.KeyHTTP, "%s %s%s%s", h.rq.Method, base.SanitizeRequestURL(h.rq, &queryValues), proto, h.formattedEffectiveUserName())
 }
 
+// iupdateResponseWriter will create an updated Response Writer if we need to log db stats.
+func (h *handler) isPublicDBRequestForStats() bool {
```
adamcfraser: "isPublicDBRequestForStats" reads like "is this a request for stats over the public API".  

Is a better name for this just "shouldUpdateBytesTransferredStats"?  In the comment you could describe the criteria we're using for this, which I think is just "is a non-blip, database scoped call over the public port"

rest/api_test.go
```diff
@@ -2275,6 +2276,7 @@ func TestHandleStats(t *testing.T) {
 
 // TestHideProductInfo ensures that detailed product info is not shown on non-admin REST API responses if set.
 func TestHideProductInfo(t *testing.T) {
+	t.Skip("here")
```
adamcfraser: Unclear what's changed that's requiring this to be skipped.

rest/changes_api.go
```diff
@@ -402,14 +402,6 @@ func (h *handler) sendSimpleChanges(channels base.Set, options db.ChangesOptions
 			}
 		}
```
adamcfraser: This change seems unrelated to this enhancement, if this kind of cleanup is justified I think it should be as a standalone PR.

rest/doc_api_test.go
```diff
@@ -164,7 +164,7 @@ func TestGuestReadOnly(t *testing.T) {
 	// Attempt to access _blipsync as guest - blip sync handling for read-only GUEST is applied at replication level (to allow pull-only replications).
 	// Should succeed permission check, and only fail on websocket upgrade
 	response = rt.SendRequest("GET", "/{{.db}}/_blipsync", "")
-	RequireStatus(t, response, http.StatusUpgradeRequired)
+	RequireStatus(t, response, http.StatusInternalServerError)
```
adamcfraser: What's happening in this PR to change the error response for this endpoint?

rest/encoded_response_writer.go
```diff
@@ -66,17 +78,37 @@ func (w *EncodedResponseWriter) WriteHeader(status int) {
...
+		w.reportStats(false)
+		return n, err
+	}
+	n, err := w.ResponseWriter.Write(b)
+	w.lastBytesWritten += int64(n)
+	w.reportStats(false)
+	return n, err
+}
+
+// ReportStats reports bytes written GetLastBytesWritten returns the number of bytes written by this response writer. This is not locked, so is only safe to call while no one is calling CountedResponseWriter.Write, usually after the response has been fully written.
```
adamcfraser:
```suggestion
// ReportStats updates the stat with the number of bytes written by this response writer. This is not locked, so is only safe to call while no one is calling CountedResponseWriter.Write, usually after the response has been fully written.
```

rest/encoded_response_writer.go
```diff
@@ -66,17 +78,37 @@ func (w *EncodedResponseWriter) WriteHeader(status int) {
...
+	w.lastBytesWritten = 0
+	w.lastReportTime = currentTime
 }
 
 func (w *EncodedResponseWriter) disableCompression() {
 	w.sniffDone = true
 }
 
 func (w *EncodedResponseWriter) sniff(bytes []byte) {
+	fmt.Println(w)
```
adamcfraser: Why does this need to be added to sniff?

adamcfraser (changes requested):

rest/changes_api.go
torcolvin: It's not unrelated exactly, because I didn't implement deprecated (from go 1.7) CloseNotifier interface in the ResponseWriters.

I can pull it out into a separate commit though.

rest/encoded_response_writer.go
torcolvin: mistake.

rest/doc_api_test.go
torcolvin: Fixed with a new isHijackable method.

rest/api_test.go
torcolvin: mistake

rest/handler.go
torcolvin: mistake

base/stats.go
```diff
@@ -461,6 +461,8 @@ type DatabaseStats struct {
...
+	HTTPBytesWritten *SgwIntStat `json:"http_bytes_written"`
```
bbrks: This seems ambiguous if we're not including Admin or Metric API data in it. Should this be specifically named after the public API?

bbrks: Where are we accounting for request body sizes? (here seems like response body size only)

rest/encoded_response_writer.go
```diff
@@ -121,6 +151,11 @@ func (w *EncodedResponseWriter) Close() {
 	}
 }
 
+// isHijackable determines if the underlying writer implements hijack interface
+func (w *EncodedResponseWriter) isHijackable() bool {
+	return false
```
bbrks, line 158: What makes this non-hijackable? Isn't there a potentially hijackable ResponseWriter inside the EncodedResponseWriter?

bbrks (commented):
Left some questions inline

rest/handler.go
torcolvin: These are actually not done with `http.ResponseWriter` so potentially I don't even have to skip this here, I'm not sure what is more clear.

base/stats.go
torcolvin: Split to https://issues.couchbase.com/browse/CBG-3163

torcolvin: renamed to PublicRestBytesWritten to match NumPublicRestRequests

adamcfraser (approved):

PR #6264, opened 2023-05-25
CBG-2909-v2: Move processes into db online function

bbrks:
CBG-2909

Moves the following processes into a `StartOnlineProcesses` method that is run only when a database is brought online (either at startup, or through a REST API request)
- `changeCache` (Init and Start)
- `Heartbeater` (Import and ISGR sharding)
- `ImportListener`
- `OIDCProviders`
- `Compact` background process
- `TombstoneCompactionManager`
- `AttachmentCompactionManager`
- ISGR replications

An offline database still has these processes (`NewDatabaseContext`):
- Stats (an offline database can report stats)
- `EventManager` (it has to send offline/online events)
- `ResyncManager` (Resync requires the database be offline)
- `sequenceAllocator` (resync with regenerate sequences)
- `activeChannels` ???
- `channelCache` ???
- `NewCfgSG` ???
- `mutationListener` ???
- `SGReplicateMgr` ???

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1865/

### DIFF
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -223,6 +223,10 @@ func (c *changeCache) Start(initialSequence uint64) error {
 // Stops the cache. Clears its state and tells the housekeeping task to stop.
 func (c *changeCache) Stop() {
 
+	if c.stopped == nil {
+		// changeCache never initialized - nothing to stop
+		return
+	}
 	if !c.stopped.CompareAndSwap(false, true) {
 		return
 	}
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -55,7 +55,8 @@ type changeCache struct {
 	receivedSeqs       map[uint64]struct{}     // Set of all sequences received
 	pendingLogs        LogPriorityQueue        // Out-of-sequence entries waiting to be cached
 	notifyChange       func(channels.Set)      // Client callback that notifies of channel changes
-	stopped            *base.AtomicBool        // Set by the Stop method
+	started            base.AtomicBool         // Set by the Start method
+	stopped            base.AtomicBool         // Set by the Stop method
 	skippedSeqs        *SkippedSequenceList    // Skipped sequences still pending on the TAP feed
 	lock               sync.RWMutex            // Coordinates access to struct fields
 	options            CacheOptions            // Cache config
@@ -162,7 +163,6 @@ func (c *changeCache) Init(logCtx context.Context, dbContext *DatabaseContext, c
 	c.logCtx = logCtx
 
 	c.notifyChange = notifyChange
-	c.stopped = base.NewAtomicBool(true) // set to false by Start
 	c.receivedSeqs = make(map[uint64]struct{})
 	c.terminator = make(chan bool)
 	c.initTime = time.Now()
@@ -213,8 +213,8 @@ func (c *changeCache) Start(initialSequence uint64) error {
 	// Set initial sequence for cache (validFrom)
 	c.channelCache.Init(initialSequence)
 
-	if !c.stopped.CompareAndSwap(true, false) {
-		return fmt.Errorf("changeCache already started")
+	if !c.started.CompareAndSwap(false, true) {
+		return errors.New("changeCache already started")
 	}
 
 	return nil
@@ -223,11 +223,13 @@ func (c *changeCache) Start(initialSequence uint64) error {
 // Stops the cache. Clears its state and tells the housekeeping task to stop.
 func (c *changeCache) Stop() {
 
-	if c.stopped == nil {
-		// changeCache never initialized - nothing to stop
+	if !c.started.IsTrue() {
+		// changeCache never started - nothing to stop
 		return
 	}
+
 	if !c.stopped.CompareAndSwap(false, true) {
+		base.WarnfCtx(c.logCtx, "changeCache was already stopped")
 		return
 	}
 
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -55,7 +55,7 @@ type changeCache struct {
 	receivedSeqs       map[uint64]struct{}     // Set of all sequences received
 	pendingLogs        LogPriorityQueue        // Out-of-sequence entries waiting to be cached
 	notifyChange       func(channels.Set)      // Client callback that notifies of channel changes
-	stopped            bool                    // Set by the Stop method
+	stopped            *base.AtomicBool        // Set by the Stop method
 	skippedSeqs        *SkippedSequenceList    // Skipped sequences still pending on the TAP feed
 	lock               sync.RWMutex            // Coordinates access to struct fields
 	options            CacheOptions            // Cache config
@@ -162,6 +162,7 @@ func (c *changeCache) Init(logCtx context.Context, dbContext *DatabaseContext, c
 	c.logCtx = logCtx
 
 	c.notifyChange = notifyChange
+	c.stopped = base.NewAtomicBool(true) // set to false by Start
 	c.receivedSeqs = make(map[uint64]struct{})
 	c.terminator = make(chan bool)
 	c.initTime = time.Now()
@@ -212,13 +213,17 @@ func (c *changeCache) Start(initialSequence uint64) error {
 	// Set initial sequence for cache (validFrom)
 	c.channelCache.Init(initialSequence)
 
+	if !c.stopped.CompareAndSwap(true, false) {
+		return fmt.Errorf("changeCache already started")
+	}
+
 	return nil
 }
 
 // Stops the cache. Clears its state and tells the housekeeping task to stop.
 func (c *changeCache) Stop() {
 
-	if !c.setStopped() {
+	if !c.stopped.CompareAndSwap(false, true) {
 		return
 	}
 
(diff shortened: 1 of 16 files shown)


db/database.go
```diff
@@ -2323,3 +2103,206 @@ func (dbc *DatabaseContext) AuthenticatorOptions() auth.AuthenticatorOptions {
...
+	// If sgreplicate is enabled on this node, register this node to accept notifications
+	if sgReplicateEnabled {
+		registerNodeErr := db.SGReplicateMgr.StartLocalNode(db.UUID, db.Heartbeater)
+		if registerNodeErr != nil {
+			return registerNodeErr
+		}
+	}
+
+	// Start DCP feed
+	base.InfofCtx(ctx, base.KeyDCP, "Starting mutation feed on bucket %v due to either channel cache mode or doc tracking (auto-import)", base.MD(db.Bucket.GetName()))
```
torcolvin: I know you copy pasted this but the auto import is handled elsewhere

```suggestion
	base.InfofCtx(ctx, base.KeyDCP, "Starting mutation feed on bucket %v", base.MD(db.Bucket.GetName()))
```

db/database.go
```diff
@@ -2323,3 +2103,206 @@ func (dbc *DatabaseContext) AuthenticatorOptions() auth.AuthenticatorOptions {
...
+	cacheFeedStatsMap := db.DbStats.Database().CacheFeedMapStats
+	if err := db.mutationListener.Start(db.Bucket, cacheFeedStatsMap.Map, db.Scopes, db.MetadataStore); err != nil {
+		db.channelCache = nil
```
torcolvin, line 2190: If you set db.channelCache here, do we actually want to do something with a defer to always stop the channel cache if this startup doesn't work?

If this is necessary in this specific spot, I'd mark it, because I would have expected it to be needed if:

```
err := db.changeCache.Start(initialSequence)
```

fails also.

db/database.go
```diff
@@ -2323,3 +2103,206 @@ func (dbc *DatabaseContext) AuthenticatorOptions() auth.AuthenticatorOptions {
...
+	// sending heartbeats before registering itself to the cfg, to avoid triggering immediate removal by other active nodes.
+	if base.IsEnterpriseEdition() && (importEnabled || sgReplicateEnabled) {
+		// Create heartbeater
+		heartbeaterPrefix := db.MetadataKeys.HeartbeaterPrefix(db.Options.GroupID)
+		heartbeater, err := base.NewCouchbaseHeartbeater(db.MetadataStore, heartbeaterPrefix, db.UUID)
+		if err != nil {
+			return pkgerrors.Wrapf(err, "Error starting heartbeater for bucket %s", base.MD(db.Bucket.GetName()).Redact())
+		}
+		err = heartbeater.StartSendingHeartbeats()
+		if err != nil {
```
torcolvin, line 2164: The old code would stop the heartbeater here if there was an error below this, in:

```
			dbContext.Heartbeater.Stop()
		})
```

db/database.go
```diff
@@ -2323,3 +2103,206 @@ func (dbc *DatabaseContext) AuthenticatorOptions() auth.AuthenticatorOptions {
...
+		db.Heartbeater = heartbeater
+	}
+
+	// If sgreplicate is enabled on this node, register this node to accept notifications
+	if sgReplicateEnabled {
+		registerNodeErr := db.SGReplicateMgr.StartLocalNode(db.UUID, db.Heartbeater)
+		if registerNodeErr != nil {
+			return registerNodeErr
+		}
+	}
```
torcolvin, line 2184: The old code would stop the ReplicateMgr if it doesn't make it online later:

```
			dbContext.SGReplicateMgr.Stop()
		})
```

db/database.go
```diff
@@ -2323,3 +2103,206 @@ func (dbc *DatabaseContext) AuthenticatorOptions() auth.AuthenticatorOptions {
...
+
+	if err := db.changeCache.Start(initialSequence); err != nil {
+		return err
+	}
+
+	// If this is an xattr import node, start import feed.  Must be started after the caching DCP feed, as import cfg
+	// subscription relies on the caching feed.
+	if db.autoImport {
+		db.ImportListener = NewImportListener(ctx, db.MetadataKeys.DCPCheckpointPrefix(db.Options.GroupID), db)
+		if importFeedErr := db.ImportListener.StartImportFeed(db); importFeedErr != nil {
```
torcolvin, line 2222: ImportFeed should be stopped if later we can't get online

db/database.go
```diff
@@ -2323,3 +2103,206 @@ func (dbc *DatabaseContext) AuthenticatorOptions() auth.AuthenticatorOptions {
...
+		registerNodeErr := db.SGReplicateMgr.StartLocalNode(db.UUID, db.Heartbeater)
+		if registerNodeErr != nil {
+			return registerNodeErr
+		}
+	}
+
+	// Start DCP feed
+	base.InfofCtx(ctx, base.KeyDCP, "Starting mutation feed on bucket %v due to either channel cache mode or doc tracking (auto-import)", base.MD(db.Bucket.GetName()))
+	cacheFeedStatsMap := db.DbStats.Database().CacheFeedMapStats
+	if err := db.mutationListener.Start(db.Bucket, cacheFeedStatsMap.Map, db.Scopes, db.MetadataStore); err != nil {
```
torcolvin, line 2189: mutationlistener should be stopped if later we can't get online

torcolvin (commented):
My biggest concerns are that if the StartOnlineProcesses fail, we leave behind DCP feeds, etc that we don't clean up, see `cleanupFunctions` in `NewDatabaseContext`

torcolvin (commented):
My biggest concerns are that if the StartOnlineProcesses fail, we leave behind DCP feeds, etc that we don't clean up, see `cleanupFunctions` in `NewDatabaseContext`

db/change_cache.go
```diff
@@ -55,7 +55,7 @@ type changeCache struct {
 	receivedSeqs       map[uint64]struct{}     // Set of all sequences received
 	pendingLogs        LogPriorityQueue        // Out-of-sequence entries waiting to be cached
 	notifyChange       func(channels.Set)      // Client callback that notifies of channel changes
-	stopped            bool                    // Set by the Stop method
+	stopped            *base.AtomicBool        // Set by the Stop method
```
adamcfraser: I agree with changing stopped to an atomic bool for compareAndSwap (instead of the setStopped/isStopped helpers), but feel like adding a separate bool for `started` or `initialized` would make the code easier to follow, instead of turning `stopped` into a 3-state "boolean" where you need to remember the meaning of nil vs false.

db/change_listener.go
```diff
@@ -192,6 +192,11 @@ func (listener *changeListener) Stop() {
 	logCtx := context.TODO()
 	base.DebugfCtx(logCtx, base.KeyChanges, "changeListener.Stop() called")
 
+	if !listener.started.IsTrue() {
+		// already stopped
```
adamcfraser: I think this is 'never started' as opposed to 'already stopped' (I thought it meant Stop() had already been called when I read it the first time)

db/database.go
```diff
@@ -2323,3 +2103,206 @@ func (dbc *DatabaseContext) AuthenticatorOptions() auth.AuthenticatorOptions {
...
+		db.channelCache = nil
+		return err
+	}
+
+	// Get current value of _sync:seq
```
adamcfraser, line 2198: It looks like we've moved the fetch of initialSequence until after the DCP feed has been started in this PR.  Was this intentional?  It looks to me like this actually fixes a bug - if the feed starts after initialSequence we're not guaranteed to see everything after initialSequence on the feed (which is the contract expected when calling changeCache.Start(initialSequence).  This might have been a bug introduced in CBG-1000, but it's possible I'm overlooking something - curious whether you made this swap intentionally.

db/database.go
```diff
@@ -2323,3 +2103,206 @@ func (dbc *DatabaseContext) AuthenticatorOptions() auth.AuthenticatorOptions {
...
+				}
+				db.backgroundTasks = append(db.backgroundTasks, bgt)
+			} else {
+				base.WarnfCtx(ctx, "Automatic compaction can only be enabled on nodes running an Import process")
+			}
+		}
+
+	}
+
+	// Make sure there is no MaxTTL set on the bucket (SG #3314)
```
adamcfraser: I think we need to have this check db creation time, not online time, since we don't want expiries to be set on things like persistent config files.

adamcfraser (changes requested):
Some small suggestions, and a question about timing around cache/DCP initialization.

db/database.go
bbrks: I guess both are required, to also catch the case where a bucket TTL is added after initial db creation but before startup. Moved into function to call from both places.

db/database.go
bbrks: I moved it down purely due to the `inialSequence` value not being passed into anything other than the changeCache, so didn't make sense to me to do it any earlier - the bugfix was unintentional.

adamcfraser: Ok.  I think the point of doing it earlier was to try to get other work done in the 1.5s between initialSequenceTime and waitForReleasedSequences, but as mentioned DCP needs to start before we get initialSequence, so I'm fine leaving this as-is.

adamcfraser (approved):

PR #6265, opened 2023-05-25
CBG-2983 Close cbgt agents on database close

adamcfraser:
Manager.Stop() doesn’t close cbgt’s gocb agents - partly because the fts use case is to have one manager that exists for the lifetime of the process.  For SG’s usage, where manager lifecycle is bound to a database, we need to shut down these agents when we close the database/importListener.

CBG-2983

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1806/

### DIFF
--- a/base/dcp_sharded.go
+++ b/base/dcp_sharded.go
@@ -468,6 +468,9 @@ func (c *CbgtContext) Stop() {
 	}
 	// ClosePIndex calls are synchronous, so can stop manager once they've completed
 	c.Manager.Stop()
+	// CloseStatsClients closes the memcached connection cbgt uses for stats calls (highseqno, etc).  sourceName and
+	// sourceUUID are bucketName/bucket UUID in our usage.  cbgt has a single global stats connection per bucket,
+	// but does a refcount check before closing, so handles the case of multiple SG databases targeting the same bucket.
 	cbgt.CloseStatsClients(c.sourceName, c.sourceUUID)
 	c.RemoveFeedCredentials(c.dbName)
 }
@@ -744,7 +747,7 @@ func (meh *sgMgrEventHandlers) OnUnregisterPIndex(pindex *cbgt.PIndex) {
 func (meh *sgMgrEventHandlers) OnFeedError(srcType string, r cbgt.Feed, feedErr error) {
 
 	// cbgt always passes srcType = SOURCE_GOCBCORE, but we have a wrapped type associated with our indexes - use that instead
-	// for logging and when calling DeleteAllIndexFromSource
+	// for our logging
 	srcType = SOURCE_DCP_SG
 
 	DebugfCtx(meh.ctx, KeyDCP, "cbgt Mgr OnFeedError, srcType: %s, feed name: %s, err: %v",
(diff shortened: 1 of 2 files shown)


base/dcp_sharded.go
```diff
@@ -771,6 +778,7 @@ func (meh *sgMgrEventHandlers) OnFeedError(srcType string, r cbgt.Feed, feedErr
 			}
 			dcpFeed.NotifyMgrOnClose()
 		}
+		return
```
torcolvin, line 784: nit: this return is kind of weird because it will just return below, i'd swap to an earlier return for readability 

```
if gone {
    return
}
```

base/dcp_sharded.go
```diff
@@ -740,6 +743,10 @@ func (meh *sgMgrEventHandlers) OnUnregisterPIndex(pindex *cbgt.PIndex) {
 // This will trigger cbgt closing and then attempting to reconnect to the feed.
 func (meh *sgMgrEventHandlers) OnFeedError(srcType string, r cbgt.Feed, feedErr error) {
```
torcolvin, line 747: nit, readabilty:

```suggestion
// OnFeedError returns an invalid srcType as the second argument
func (meh *sgMgrEventHandlers) OnFeedError(_ string, r cbgt.Feed, feedErr error) {
```

base/dcp_sharded.go
```diff
@@ -54,6 +54,8 @@ type CbgtContext struct {
 	eventHandlers     *sgMgrEventHandlers      // Event handler callbacks
 	ctx               context.Context          // Log context
 	dbName            string                   // Database name
+	sourceName        string                   // cbgt source name. Store on CbgtContext for access during teardown
```
torcolvin, line 57: would it be more clear to call this bucketName and bucketUUID? they are only used as sourceName and sourceUUID in cbgt's `CloseStatsClients`

torcolvin (commented):
Everything is just naming improvements you can take or leave, but I think there's a problem.

If `CloseStatsClients` takes bucketName and bucketUUID, this will break if there a multiple databases on the same bucket, as is now allowed in 3.1

base/dcp_sharded.go
adamcfraser: I agree it's a bit weird as written, but I almost introduced a bug in the previous change when I added handling below this without the return, so thought it might be a good thing to leave in place.

base/dcp_sharded.go
adamcfraser: They are also sourceName and sourceUUID in c.Manager.CreateIndex.  You're right that for our usage they are always going to be bucketName and bucketUUID, but I thought it would be useful to keep track of them under the name/concept cbgt is using.

base/dcp_sharded.go
adamcfraser: I'm not completely convinced that this is an improvement to readability - going to leave this as-is for now.

adamcfraser (commented):
PR feedback responses.

adamcfraser:
Added a comment to indicate that cbgt does refcounting on CloseStatsClients, so it handles the case of multiple databases targeting the same bucket.

torcolvin (approved):

base/dcp_sharded.go
torcolvin: This is fine.

PR #6285, opened 2023-06-02
Make test pass if there are buckets that are non test buckets

torcolvin:
Tested with and without travel-sample bucket.


snej (approved):
Thanks!

bbrks (approved):

PR #6287, opened 2023-06-05
CBG-3022: Replicator will not reconnect when max_back_off != 0

gregns1:
CBG-3022

MaxBackOff Behaviour:

MaxBackOff = 0: 
* Current behaviour: Reconnects with exponential back off up till 5 mins and will stop trying to reconnect after 10 mins 
* Behaviour we want: same as above
* TotalReconnectTimeout = 10 mins 
* MaxReconnectInterval = 5 mins 

MaxBackOff = default value:
* Current Behaviour: Doesn’t reconnect 
* Behaviour we want: to rear exponential back off till 5 mins, but retry indefinitely
* TotalReconnectTimeout = 0
* MaxReconnectInterval = 5 mins 

MaxBackOff = something other than 0 or default:
* Current Behaviour: Doesn’t reconnect 
* Behaviour we want: to have exponential back off time till the value specified (in minutes) and then retry indefinitely
* TotalReconnectTimeout = 0 
* MaxReconnectInterval = <specified MaxBackOff value> mins

What the problem is currently that TotalReconnectTimeout = 0 will disable reconnect all together. I think we want when TotalReconnectTimeout = 0 to retry indefinitely. Changes I am making locally seem to work for this without effecting tests that are testing for missing collections. 

This change also broke CBG-2895 (replication limit handling) tests. The tests were asserting on the replicator going into an error state once being rejected by the limit. We should’ve been entering a reconnect loop (like we’d expect CBL to be doing, hence the 503 code returned), so I have also update those tests to assert on replication going into reconnect loop. Also took out assertion on stats, due to the nature these stats get incremented each time on the reconnect loop and there is no way we can assert on how many times this will happen in the test. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1851/

### DIFF
--- a/db/active_replicator.go
+++ b/db/active_replicator.go
@@ -214,7 +214,7 @@ func connect(arc *activeReplicatorCommon, idSuffix string) (blipSender *blip.Sen
 	blipContext.WebsocketPingInterval = arc.config.WebsocketPingInterval
 	blipContext.OnExitCallback = func() {
 		// fall into a reconnect loop only if the connection is unexpectedly closed.
-		if arc.ctx.Err() == nil && arc.config.TotalReconnectTimeout != 0 {
+		if arc.ctx.Err() == nil {
 			go arc.reconnectLoop()
 		}
 	}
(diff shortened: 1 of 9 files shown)


bbrks (approved):
LGTM :+1: thanks!

PR #6291, opened 2023-06-06
make jenkins tests compatible with amazon linux 2023

torcolvin:
- docker-compose was moved to jenkins init script
- detect docker-compose (v1) by looking for amazon linux 2 specifically

I plan to move the Integration Test / Master Integration / Pipeline Agent to 2023, but leave the old Integration Test runs to use aws linux 2 AMIs

### DIFF
--- a/integration-test/start_server.sh
+++ b/integration-test/start_server.sh
@@ -42,7 +42,10 @@ if [ "${CBS_ROOT_DIR:-}" != "" ]; then
     DOCKER_CBS_ROOT_DIR="${CBS_ROOT_DIR}"
 fi
 
-if [[ -n "${JENKINS_URL:-}" ]]; then
+set +e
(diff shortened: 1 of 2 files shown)


adamcfraser:
Do the service scripts also need to be updated to support Amazon Linux 2023, or does `amzn*` cover it?  https://github.com/couchbase/sync_gateway/blob/master/service/sync_gateway_service_install.sh#LL290C1-L290C6

torcolvin:
> Do the service scripts also need to be updated to support Amazon Linux 2023, or does `amzn*` cover it? https://github.com/couchbase/sync_gateway/blob/master/service/sync_gateway_service_install.sh#LL290C1-L290C6

This is fine too, but this actually is a different problem which is "is amazon linux 2023 package a supported platform for running sync gateway?"

My change only covers running tests on jenkins.

bbrks (approved):

PR #6293, opened 2023-06-07
CBG-2932: Request Count stat for REST

gregns1:
CBG-2932

SImple implementation where we count the number of public rest requests throiughh sybnc gateway. Does this through adding a function on the handler to call each time a new handler is crerated and if the handlerPrivs is public realted we increment the stat. Also has a check to exit the function before we increment the stats for blipsync requests. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1866/

### DIFF
--- a/rest/api_test.go
+++ b/rest/api_test.go
@@ -75,30 +75,34 @@ func TestPublicRESTStatCount(t *testing.T) {
 	// create a user to authenticate as for public api calls and assert the stat hasn't incremented as a result
 	resp := rt.SendAdminRequest(http.MethodPut, "/{{.db}}/_user/greg", GetUserPayload(t, "greg", "letmein", "", rt.GetSingleTestDatabaseCollection(), []string{"ABC"}, nil))
 	RequireStatus(t, resp, http.StatusCreated)
-	base.WaitForStat(func() int64 {
+	_, ok := base.WaitForStat(func() int64 {
 		return rt.GetDatabase().DbStats.DatabaseStats.NumPublicRestRequests.Value()
 	}, 0)
+	require.True(t, ok)
 
 	// use public api to put a doc through SGW then assert the stat has increased by 1
 	resp = rt.SendUserRequest(http.MethodPut, "/{{.keyspace}}/doc1", `{"foo":"bar", "channels":["ABC"]}`, "greg")
 	RequireStatus(t, resp, http.StatusCreated)
-	base.WaitForStat(func() int64 {
+	_, ok = base.WaitForStat(func() int64 {
 		return rt.GetDatabase().DbStats.DatabaseStats.NumPublicRestRequests.Value()
 	}, 1)
+	require.True(t, ok)
 
 	// send admin request assert that the public rest count doesn't increase
 	resp = rt.SendAdminRequest(http.MethodGet, "/{{.keyspace}}/doc1", "")
 	RequireStatus(t, resp, http.StatusOK)
-	base.WaitForStat(func() int64 {
+	_, ok = base.WaitForStat(func() int64 {
 		return rt.GetDatabase().DbStats.DatabaseStats.NumPublicRestRequests.Value()
 	}, 1)
+	require.True(t, ok)
 
 	// send another public request to assert the stat increases by 1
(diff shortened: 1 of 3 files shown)


rest/handler.go
```diff
@@ -128,6 +128,7 @@ func makeHandler(server *ServerContext, privs handlerPrivs, accessPermissions []
 		err := h.invoke(method, accessPermissions, responsePermissions)
 		h.writeError(err)
 		h.logDuration(true)
+		h.logRESTCount()
```
torcolvin: I wonder if the right place to put this is https://github.com/couchbase/sync_gateway/blob/master/rest/handler.go#L266

Because it will count requests before they happen - since continuous feed requests are long lived. I'm not sure how many people use them though, or if there are other long running requests. I'd consult with other team members on this.

rest/handler.go
```diff
@@ -605,6 +607,19 @@ func (h *handler) logDuration(realTime bool) {
 	)
 }
 
+func (h *handler) logRESTCount() {
```
torcolvin, line 610: can you add a docstring for this?

rest/handler.go
bbrks: Agree - would be preferable to increment stat on incoming request, rather than completed request.

bbrks (dismissed):

rest/api_test.go
```diff
@@ -68,6 +68,44 @@ func TestRoot(t *testing.T) {
 	assert.Equal(t, "GET, HEAD", response.Header().Get("Allow"))
 }
 
+func TestPublicRESTStatCount(t *testing.T) {
+	rt := NewRestTester(t, &RestTesterConfig{SyncFn: channels.DocChannelsSyncFunction})
+	defer rt.Close()
+
+	// create a user to authenticate as for public api calls and assert the stat hasn't incremented as a result
+	resp := rt.SendAdminRequest(http.MethodPut, "/{{.db}}/_user/greg", GetUserPayload(t, "greg", "letmein", "", rt.GetSingleTestDatabaseCollection(), []string{"ABC"}, nil))
```
torcolvin: I created some helper functions for this, `rt.CreateUser`

rest/api_test.go
```diff
@@ -68,6 +68,44 @@ func TestRoot(t *testing.T) {
...
+	require.True(t, ok)
+
+	// send another public request to assert the stat increases by 1
+	resp = rt.SendUserRequest(http.MethodGet, "/{{.keyspace}}/doc1", "", "greg")
+	RequireStatus(t, resp, http.StatusOK)
+	_, ok = base.WaitForStat(func() int64 {
+		return rt.GetDatabase().DbStats.DatabaseStats.NumPublicRestRequests.Value()
+	}, 2)
+	require.True(t, ok)
+
```
torcolvin, line 105: I wonder if we can do something to test a few other cases:

(1) test the metrics port
(2) test blipsync code? It is probably sufficient to hit the endpoint directly but I wonder if instantiating a blip tester and getting changes would also work because that's what a customer would do.
(3) test requests that don't access a given database

rest/handler.go
```diff
@@ -605,6 +606,20 @@ func (h *handler) logDuration(realTime bool) {
 	)
 }
 
+// logRESTCount will increment the number of public REST requests stat for public REST requests excluding blipsync requests
```
```suggestion
// logRESTCount will increment the number of public REST requests stat for public REST requests excluding _blipsync requests if they are attached to a database.
```

torcolvin (commented):
LGTM just a few comments on testing to hit all the cases to prevent regressions.

rest/api_test.go
```diff
@@ -104,6 +103,36 @@ func TestPublicRESTStatCount(t *testing.T) {
 	}, 2)
 	require.True(t, ok)
 
+	resp = rt.SendUserRequest(http.MethodGet, "/{{.db}}/_blipsync", "", "greg")
+	fmt.Println(resp.Code)
```
torcolvin: change the print to require?

gregns1: ah left that in from debugging

torcolvin (approved):

PR #6295, opened 2023-06-09
CBG-2930: Import compute stat implementation

gregns1:
CBG-2930

Import feed compute stat implementation. 

- total time of the function (ms) * memory of Event.Value in bytes 
- We have to record time in seconds and convert to milliseconds as .Millisecond() returns ineger value. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1917/
- [x] `GSI=true,xattrs=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1916/

### DIFF
--- a/base/constants.go
+++ b/base/constants.go
@@ -177,6 +177,8 @@ const (
 	EmptyDocument = `{}`
 )
 
+const MiB = 1 * 1024 * 1024
+
 var (
 	SyncFnAccessErrors = []string{
 		HTTPErrorf(403, SyncFnErrorMissingRole).Error(),
--- a/base/constants.go
+++ b/base/constants.go
@@ -177,8 +177,6 @@ const (
 	EmptyDocument = `{}`
 )
 
-const MiB = 1 * 1024 * 1024
-
 var (
 	SyncFnAccessErrors = []string{
 		HTTPErrorf(403, SyncFnErrorMissingRole).Error(),
--- a/base/stats.go
+++ b/base/stats.go
@@ -692,6 +692,8 @@ type SharedBucketImportStats struct {
 	ImportHighSeq *SgwIntStat `json:"import_high_seq"`
 	// The total number of import partitions.
 	ImportPartitions *SgwIntStat `json:"import_partitions"`
+	// Represents the compute unit
+	ImportProcessCompute *SgwFloatStat `json:"import_process_compute"`
 }
 
 type SgwStat struct {
@@ -1921,6 +1923,10 @@ func (d *DbStats) InitSharedBucketImportStats() error {
 			return err
 		}
 		resUtil.ImportPartitions, err = NewIntStat(SubsystemSharedBucketImport, "import_partitions", labelKeys, labelVals, prometheus.GaugeValue, 0)
+		if err != nil {
+			return err
+		}
+		resUtil.ImportProcessCompute, err = NewFloatStat(SubsystemSharedBucketImport, "import_process_compute", labelKeys, labelVals, prometheus.GaugeValue, 0)
 		if err != nil {
 			return err
 		}
--- a/base/stats.go
+++ b/base/stats.go
@@ -505,6 +505,8 @@ type DatabaseStats struct {
 	SyncFunctionExceptionCount *SgwIntStat `json:"sync_function_exception_count"`
 	// The total number of times a replication connection is rejected due ot it being over the threshold
 	NumReplicationsRejectedLimit *SgwIntStat `json:"num_replications_rejected_limit"`
+	// Represents the compute unit for import processes on the database
+	ImportProcessCompute *SgwIntStat `json:"import_process_compute"`
 
 	// These can be cleaned up in future versions of SGW, implemented as maps to reduce amount of potential risk
 	// prior to Hydrogen release. These are not exported as part of prometheus and only exposed through expvars
@@ -692,8 +694,6 @@ type SharedBucketImportStats struct {
 	ImportHighSeq *SgwIntStat `json:"import_high_seq"`
 	// The total number of import partitions.
 	ImportPartitions *SgwIntStat `json:"import_partitions"`
-	// Represents the compute unit
-	ImportProcessCompute *SgwIntStat `json:"import_process_compute"`
 }
 
 type SgwStat struct {
@@ -1456,6 +1456,11 @@ func (d *DbStats) initDatabaseStats() error {
 	if err != nil {
 		return err
 	}
+	resUtil.ImportProcessCompute, err = NewIntStat(SubsystemSharedBucketImport, "import_process_compute", labelKeys, labelVals, prometheus.CounterValue, 0)
+	if err != nil {
+		return err
+	}
+
 	resUtil.ImportFeedMapStats = &ExpVarMapWrapper{new(expvar.Map).Init()}
 
 	resUtil.CacheFeedMapStats = &ExpVarMapWrapper{new(expvar.Map).Init()}
@@ -1502,6 +1507,7 @@ func (d *DbStats) unregisterDatabaseStats() {
 	prometheus.Unregister(d.DatabaseStats.NumReplicationsRejectedLimit)
 	prometheus.Unregister(d.DatabaseStats.NumPublicRestRequests)
 	prometheus.Unregister(d.DatabaseStats.TotalSyncTime)
+	prometheus.Unregister(d.DatabaseStats.ImportProcessCompute)
 }
 
 func (d *DbStats) CollectionStat(scopeName, collectionName string) (*CollectionStats, error) {
@@ -1926,10 +1932,6 @@ func (d *DbStats) InitSharedBucketImportStats() error {
 		if err != nil {
 			return err
 		}
-		resUtil.ImportProcessCompute, err = NewIntStat(SubsystemSharedBucketImport, "import_process_compute", labelKeys, labelVals, prometheus.CounterValue, 0)
-		if err != nil {
-			return err
-		}
 
 		d.SharedBucketImportStats = resUtil
 	}
@@ -1943,7 +1945,6 @@ func (d *DbStats) unregisterSharedBucketImportStats() {
 	prometheus.Unregister(d.SharedBucketImportStats.ImportProcessingTime)
 	prometheus.Unregister(d.SharedBucketImportStats.ImportHighSeq)
 	prometheus.Unregister(d.SharedBucketImportStats.ImportPartitions)
-	prometheus.Unregister(d.SharedBucketImportStats.ImportProcessCompute)
 }
 
 func (d *DbStats) SharedBucketImport() *SharedBucketImportStats {
--- a/base/stats.go
+++ b/base/stats.go
@@ -693,7 +693,7 @@ type SharedBucketImportStats struct {
 	// The total number of import partitions.
 	ImportPartitions *SgwIntStat `json:"import_partitions"`
 	// Represents the compute unit
-	ImportProcessCompute *SgwFloatStat `json:"import_process_compute"`
+	ImportProcessCompute *SgwIntStat `json:"import_process_compute"`
 }
 
 type SgwStat struct {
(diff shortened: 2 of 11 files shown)


base/stats.go
```diff
@@ -1909,6 +1911,10 @@ func (d *DbStats) InitSharedBucketImportStats() error {
 		if err != nil {
 			return err
 		}
+		resUtil.ImportProcessCompute, err = NewFloatStat(SubsystemSharedBucketImport, "import_process_compute", labelKeys, labelVals, prometheus.GaugeValue, 0)
+		if err != nil {
+			return err
+		}
```
bbrks: I wonder if these types of stats should actually be in a serverless section that is only enabled/calculated when the serverless flag is enabled.

This has no benefit/makes no sense for an on-prem deployment.

db/import_listener.go
```diff
@@ -15,11 +15,14 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"time"
 
 	sgbucket "github.com/couchbase/sg-bucket"
 	"github.com/couchbase/sync_gateway/base"
 )
 
+const MiB = 1048576
```
bbrks: This seems like it might be useful to put in `base/constants.go`? There's quite a lot of instances of `1024 * 1024` that would be clearer with this unit.

bbrks (dismissed):

rest/importtest/import_test.go
```diff
@@ -1917,7 +1917,7 @@ func TestImportRevisionCopyUnavailable(t *testing.T) {
 	assert.Equal(t, 404, response.Code)
 }
 
-func TestImportComputeStatOnDemand(t *testing.T) {
+func TestImportComputeStatOnDemandGet(t *testing.T) {
 	base.SkipImportTestsIfNotEnabled(t)
```
torcolvin, line 1921: is this skip required here, especially since you set `AutoImport: false`?

Also, if you set `AutoImport: true` does this do double counting anywhere? That's the thing I'm most worried about in this PR.

rest/importtest/import_test.go
```diff
@@ -1971,6 +1971,54 @@ func TestImportComputeStatOnDemand(t *testing.T) {
...
+	dataStore := rt.GetSingleDataStore()
+
+	key := t.Name()
+	docBody := make(map[string]interface{})
+	docBody["type"] = "non-mobile"
+	docBody["channels"] = "ABC"
+
+	// assert the stat starts at 0 for a new database
+	computeStat := rt.GetDatabase().DbStats.SharedBucketImportStats.ImportProcessCompute.Value()
+	assert.Equal(t, float64(0), computeStat)
```
```suggestion
	require.Equal(t, float64(0), computeStat)
```

rest/importtest/import_test.go
```diff
@@ -1971,6 +1971,54 @@ func TestImportComputeStatOnDemand(t *testing.T) {
...
+
+	// add doc to bucket
+	_, err := dataStore.Add(key, 0, docBody)
+	assert.NoError(t, err, "Unable to insert doc TestImportDelete")
```
```suggestion
	require.NoError(t, err, "Unable to insert doc TestImportDelete")
```

rest/importtest/import_test.go
```diff
@@ -1971,6 +1971,54 @@ func TestImportComputeStatOnDemand(t *testing.T) {
...
+
+	// trigger on demand import of this doc - should be rejected
+	response := rt.SendAdminRequest("GET", "/db/_raw/"+key, "")
+	rest.RequireStatus(t, response, http.StatusNotFound)
+	assertDocProperty(t, response, "reason", "Not imported")
+
+	// assert stat still no incremented as import filter rejects the doc
+	computeStat1 := rt.GetDatabase().DbStats.SharedBucketImportStats.ImportProcessCompute.Value()
+	assert.Equal(t, float64(0), computeStat1)
```
```suggestion
	require.Equal(t, float64(0), computeStat1)
```

rest/importtest/import_test.go
```diff
@@ -1971,6 +1971,54 @@ func TestImportComputeStatOnDemand(t *testing.T) {
...
+
+	// rewrite through SG - should treat as new insert
+	docBodyString := `{"type":"SG client rewrite",
+	                 "channels": "NBC"}`
+	response = rt.SendAdminRequest("PUT", fmt.Sprintf("/db/%s", key), docBodyString)
+	assert.Equal(t, 201, response.Code)
```
```suggestion
	rest.RequireStatus(t, response, http.StatusCreated)
```

rest/importtest/import_test.go
gregns1: Skip is required. Will ensure walrus doesn't run on it etc.

db/import_listener.go
```diff
@@ -156,7 +157,24 @@ func (il *importListener) ProcessFeedEvent(event sgbucket.FeedEvent) (shouldPers
...
+
 func (il *importListener) ImportFeedEvent(event sgbucket.FeedEvent) {
+	startTime := time.Now()
+	defer func() {
+		// we must grab the time in seconds here and convert to ms as the .Milliseconds() function returns integer millisecond count
+		functionTime := time.Since(startTime).Seconds()
+		bytes := float64(len(event.Value))
+		stat := calculateImportCompute(bytes, functionTime)
+		il.importStats.ImportProcessCompute.Add(stat)
+	}()
```
torcolvin: you only want to actually calculate stats if this is going to run the import function. We have to unmarshal the document or the xattrs to see if it has metadata before we know if we need to run import.

so it has to hit `syncData == nil || !isSGWrite` block. I think this is why the import double counting occurs with on demand access. So probably set a variable for whether it is actually running an import and only report in the defer if import was run.

base/constants.go
```diff
@@ -177,6 +177,8 @@ const (
 	EmptyDocument = `{}`
 )
 
+const MiB = 1 * 1024 * 1024
+
```
```suggestion
```

rest/importtest/import_test.go
```diff
@@ -1916,6 +1917,141 @@ func TestImportRevisionCopyUnavailable(t *testing.T) {
 	assert.Equal(t, 404, response.Code)
 }
 
+func TestImportComputeStatOnDemandGet(t *testing.T) {
+	base.SkipImportTestsIfNotEnabled(t)
```
torcolvin, line 1921: This I don't think requires the import feed, so I don't know that this is correct, see also OnDemandPut, can you test this with removing this and `SG_TEST_USE_XATTRS=false`?

torcolvin (dismissed):

rest/importtest/import_test.go
gregns1: Its needed -> https://github.com/couchbase/sync_gateway/blob/8bc18bd35833deea85b6f21586f0b8de7059a0f4/db/database.go#L1911

gregns1: Test will panic without it. SharedBucketImportStats don't exist without XATTRS. We would need to move the stats to collection stats or database stats or something if this is needed.

torcolvin: This is weird, but there are just no stats with xattrs, I'm inclined to push deprecation of inline data in sync gateway. 

Thanks for explanation, that was not intuitive to me.

torcolvin:
This is also missing the stats for resync based import:

* query based resync https://github.com/couchbase/sync_gateway/blob/master/db/background_mgr_resync.go#L81 (this is unsupported as of 3.1 but flagged if DCP resync fails)
* https://github.com/couchbase/sync_gateway/blob/master/db/background_mgr_resync_dcp.go#L109

db/background_mgr_resync_dcp_test.go
```diff
@@ -336,6 +336,43 @@ func TestResyncManagerDCPRunTwice(t *testing.T) {
 	wg.Wait()
 }
 
+func TestResyncImportComputeStat(t *testing.T) {
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("Test requires Couchbase Server")
```
```suggestion
		t.Skip("Test requires Couchbase Server for DCP")
```

db/background_mgr_resync_dcp_test.go
```diff
@@ -336,6 +336,43 @@ func TestResyncManagerDCPRunTwice(t *testing.T) {
...
+	}
+	const docsToCreate = 1000
```
torcolvin: can we create fewer docs? creating a lot of docs causes them to show up in future tests DCP feeds as deleted until the server decides to reset DCP feed. We should need minimal docs for this test.

It probably won't make this test faster but it will make some other tests fast.

db/background_mgr_resync_dcp_test.go
```diff
@@ -336,6 +336,43 @@ func TestResyncManagerDCPRunTwice(t *testing.T) {
...
+	defer db.Close(ctx)
+
+	resycMgr := NewResyncManagerDCP(db.MetadataStore, base.TestUseXattrs(), db.MetadataKeys)
+	require.NotNil(t, resycMgr)
+	db.ResyncManager = resycMgr
+
+	options := map[string]interface{}{
+		"database":            db,
+		"regenerateSequences": false,
+		"collections":         ResyncCollections{},
```
torcolvin, line 354: does this test work with 

matrix of:

SG_TEST_USE_GSI=true,false
SG_TEST_USE_XATTRS=true,false
SG_TEST_USE_DEFAULT_COLLECTION=true,false

?

rest/importtest/import_test.go
```diff
@@ -1916,6 +1918,191 @@ func TestImportRevisionCopyUnavailable(t *testing.T) {
...
+
+	for i := 0; i < numDocs; i++ {
+		resp = rt.SendAdminRequest(http.MethodPut, "/{{.keyspace}}/"+fmt.Sprint(i), `{"random": "doc"}`)
+		rest.RequireStatus(t, resp, http.StatusCreated)
+	}
+
+	computeStat := rt.GetDatabase().DbStats.DatabaseStats.ImportProcessCompute.Value()
+	require.Equal(t, int64(0), computeStat)
+
+	resp = rt.SendAdminRequest(http.MethodPost, "/{{.db}}/_offline", "")
```
torcolvin, line 2082: offline is async so we need to wait for this to actually turn offline, this is copy pasted from elsewhere in the code base but you need:

```
        response = rt.SendAdminRequest("POST", "/db/_offline", "")
	RequireStatus(t, response, 200)

	rt.waitForState("online")
```

I might find the other places we do and replace this with a function like rt.takeDbOffline

rest/importtest/import_test.go
```diff
@@ -1916,6 +1918,191 @@ func TestImportRevisionCopyUnavailable(t *testing.T) {
...
+		response := rt.SendAdminRequest("GET", "/{{.db}}/_resync", "")
+		var status db.ResyncManagerResponse
+		err := json.Unmarshal(response.BodyBytes(), &status)
+		assert.NoError(t, err)
+
+		var val interface{}
+		_, err = rt.Bucket().DefaultDataStore().Get(rt.GetDatabase().ResyncManager.GetHeartbeatDocID(t), &val)
+
+		return status.State == db.BackgroundProcessStateCompleted && base.IsDocNotFoundError(err)
+	}, 200, 200)
```
torcolvin: This is duplicated for a bunch of functions in existing tests, can we write a function for this. This is a non inclusive list:

- TestResyncOverDCP
- TestDBOfflineSingleResync
- TestDBOfflinePostResync

and write helpers for taking db offline and checking completion. We can probably write a function that takes `rt.GetDatabase().ResyncManager` and does a type switch to assert the right thing for DCP and non DCP, and then make this test a parameterized test for DCP and non DCP.

db/database.go
```diff
@@ -1734,9 +1734,21 @@ func (db *DatabaseCollectionWithUser) getResyncedDocument(ctx context.Context, d
...
+		functionTime := time.Since(startTime).Milliseconds()
+		var bytes int
+		if updatedDoc != nil {
+			bytes = len(updatedDoc._rawBody)
+		} else {
+			return
+		}
+		stat := calculateImportCompute(int64(bytes), functionTime)
+		db.dbStats().DatabaseStats.ImportProcessCompute.Add(stat)
+	}()
```
torcolvin, line 1751: Question for the group is whether we should include `UpdateAllPrincipals` in this computation as well, which is done for the whole group. We could bundle up the whole resync process as one computation rather than going doc by doc.

db/background_mgr_resync_dcp_test.go
gregns1: I ran all my test locally with the full matrix and everything passed, will double check again before putting this back on you though

torcolvin (approved):

PR #6302, opened 2023-06-13
Fix error checking inside Request utility func

bbrks:
If `http.NewRequest` returns `nil, err` we don't know and panic when trying to do `request.RequestURI`


torcolvin (commented):
Is this worth writing a test for to make sure panic starts with "NewRequest failed:" or is this too hard to test?

bbrks:
It's easy enough to test if you forget to put a leading `/` in a request path with a `{{.keyspace}}`, as I found out...

added test to trigger by specifying invalid URL escape

torcolvin (approved):

PR #6304, opened 2023-06-13
Fix CE integration tests

bbrks:
Follow up for #6086 #6094 

The function `isServerEnterprise` was doing `strings.Contains` incorrectly, resulting in both CE and EE _always_ returning `false` for `isServerEnterprise`.

Corrected and flipped bool returned by `TestsUseServerCE` to accommodate fix.
This resulted in uncovering failing CE tests.

Subsequent fixes:
- `TestDisablePermissionCheck`
  - Was trying to use RBAC permissions in CE (which don't exist)
- `TestAdminAPIAuth`
  - Was trying to use a role that doesn't exist in CE
- `TestCECheck`
  - Was creating a HTTP request, expecting it not to work (when it did), and not sending it! The test was _never_ being executed...
  - Fixed test to actually send the request to CB Server

- [x] `EE` `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1881/
- [ ] `CE` `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1882/
  - FAIL: [`TestNewlyCreateSGWPermissions`](https://jenkins.sgwdev.com/job/SyncGateway-Integration/1882/testReport/junit/github/com_couchbase_sync_gateway_rest/TestNewlyCreateSGWPermissions/)
- [x] `CE` `TestNewlyCreateSGWPermissions` fix https://jenkins.sgwdev.com/job/SyncGateway-Integration/1883/

### DIFF
--- a/rest/admin_api_auth_test.go
+++ b/rest/admin_api_auth_test.go
@@ -918,9 +918,11 @@ func TestDisablePermissionCheck(t *testing.T) {
 	viewsAdminRole := RouteRole{RoleName: "views_admin", DatabaseScoped: true}
 	statsReadPermission := Permission{".stats!read", true}
 
+	serverIsCE := base.TestsUseServerCE()
+
 	rt := NewRestTester(t, nil)
 	SGWorBFArole := RouteRole{MobileSyncGatewayRole.RoleName, true}
-	if base.TestsUseServerCE() {
+	if serverIsCE {
 		SGWorBFArole = RouteRole{BucketFullAccessRole.RoleName, true}
 	}
 	rt.Close()
@@ -930,6 +932,7 @@ func TestDisablePermissionCheck(t *testing.T) {
 		CreateUser         string
 		CreateUserRole     *RouteRole
 		RequirePerms       []Permission
+		EEOnly             bool
 		DoPermissionCheck  bool
 		ExpectedStatusCode int
 	}{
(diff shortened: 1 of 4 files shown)


rest/admin_api_auth_test.go
```diff
@@ -845,8 +845,10 @@ func TestAdminAPIAuth(t *testing.T) {
 		},
 	}
 
+	serverIsCE := base.TestsUseServerCE()
+
 	SGWorBFArole := MobileSyncGatewayRole.RoleName
-	if base.TestsUseServerCE() {
+	if serverIsCE {
 		SGWorBFArole = BucketFullAccessRole.RoleName
 	}
```
torcolvin, line 853: I wonder if we can combine this into a function to get role for sync gateway admin. Is there ever a plan to have `MobileSyncGateway` on CE server or is version of server dependent?

torcolvin (dismissed):

rest/admin_api_auth_test.go
bbrks: It would be both an edition and version check if it were to ever happen. But I don't expect it to.

mohammed-madi (dismissed):

torcolvin (approved):

PR #6306, opened 2023-06-13
CBG-3101: Use Scope.Query if server supports it

torcolvin:
## Pre-review checklist

- [x] `GSI=true,xattrs=true,enterprise-7.2.0,db package only` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1889/
- [x] `GSI=true,xattrs=true,enterprise-6.6.6,db package only` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1888/

### DIFF
--- a/base/cluster_n1ql.go
+++ b/base/cluster_n1ql.go
@@ -181,11 +181,9 @@ func (cl *ClusterOnlyN1QLStore) runQuery(statement string, n1qlOptions *gocb.Que
 	if n1qlOptions == nil {
 		n1qlOptions = &gocb.QueryOptions{}
 	}
-	if !cl.supportsCollections {
-		return cl.cluster.Query(statement, n1qlOptions)
-	}
-	scope := cl.cluster.Bucket(cl.bucketName).Scope(cl.scopeName)
-	return scope.Query(statement, n1qlOptions)
+	queryResults, err := cl.cluster.Query(statement, n1qlOptions)
+
+	return queryResults, err
 }
 
 func (cl *ClusterOnlyN1QLStore) WaitForIndexesOnline(indexNames []string, failfast bool) error {
--- a/base/cluster_n1ql.go
+++ b/base/cluster_n1ql.go
@@ -186,8 +186,17 @@ func (cl *ClusterOnlyN1QLStore) runQuery(statement string, n1qlOptions *gocb.Que
 	return queryResults, err
 }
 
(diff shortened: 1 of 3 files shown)


base/cluster_n1ql.go
```diff
@@ -181,9 +181,11 @@ func (cl *ClusterOnlyN1QLStore) runQuery(statement string, n1qlOptions *gocb.Que
 	if n1qlOptions == nil {
 		n1qlOptions = &gocb.QueryOptions{}
 	}
-	queryResults, err := cl.cluster.Query(statement, n1qlOptions)
-
-	return queryResults, err
+	if !cl.supportsCollections {
+		return cl.cluster.Query(statement, n1qlOptions)
+	}
+	scope := cl.cluster.Bucket(cl.bucketName).Scope(cl.scopeName)
```
adamcfraser: I think opening a new bucket connection here defeats the purpose of the cluster-based connection, if it's opening a connection to the bucket under the hood.

adamcfraser (changes requested):

base/collection_n1ql.go
```diff
@@ -144,13 +144,21 @@ func (c *Collection) BuildDeferredIndexes(indexSet []string) error {
...
 
 	if n1qlOptions == nil {
 		n1qlOptions = &gocb.QueryOptions{}
 	}
-	queryResults, err := b.cluster.Query(statement, n1qlOptions)
+
+	var queryResults *gocb.QueryResult
+	var err error
+	if b.IsSupported(sgbucket.BucketStoreFeatureCollections) {
+		scope := b.cluster.Bucket(b.BucketName()).Scope(scopeName)
```
adamcfraser: I think we already have a handle to a bucket here (b.bucket) that we could be using instead of having the SDK build one via cluster.Bucket().

adamcfraser (changes requested):

base/cluster_n1ql.go
torcolvin: I've ignored this for now.

base/collection_n1ql.go
torcolvin: Good call.

adamcfraser (approved):
LGTM.  Some of this is going to get reworked as part of index init optimizations, but this is a good intermediary approach to unblock environments that require Scope-limited queries.

PR #6307, opened 2023-06-13
CBG-2993: allow tcp_nodelay to be set to false for tls connections

gregns1:
CBG-2993

Small change to allow tcp_nodelay to be set for tls connections. The PR essentialy checks for tls connection and then gets the underlying tcp connection to perform the SetNoDelay operation on. 
Tested manually through the debugger and seems to work fine with tls connections and TCP connections. 
This change includes a refactor of the code. I have made the function return type void and the callers don't use the returned type. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1890/

### DIFF
--- a/base/util.go
+++ b/base/util.go
@@ -1650,25 +1650,27 @@ func GetHttpClientForWebSocket(insecureSkipVerify bool) *http.Client {
 // Turns off TCP NODELAY on a TCP connection.
 // On success returns true; on failure logs a warning and returns false.
 // (There's really no reason for a caller to take note of the return value.)
-func turnOffNoDelay(ctx context.Context, conn net.Conn) bool {
+func turnOffNoDelay(ctx context.Context, conn net.Conn) {
+	var underlyingTCPConn *net.TCPConn
 	if tlsConn, ok := conn.(*tls.Conn); ok {
-		underlyingTCPConn := tlsConn.NetConn()
-		if tcp, ok := underlyingTCPConn.(*net.TCPConn); !ok {
-			WarnfCtx(ctx, "Couldn't turn off NODELAY for tls conn %v: underlying connection is not type *net.TCPConn", conn)
-		} else if err := tcp.SetNoDelay(false); err != nil {
-			WarnfCtx(ctx, "Couldn't turn off NODELAY for %v: %v", conn, err)
+		netConnection := tlsConn.NetConn()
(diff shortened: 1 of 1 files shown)


base/util.go
```diff
@@ -1650,15 +1650,27 @@ func GetHttpClientForWebSocket(insecureSkipVerify bool) *http.Client {
...
-		WarnfCtx(ctx, "Couldn't turn off NODELAY for %v: %v", conn, err)
+func turnOffNoDelay(ctx context.Context, conn net.Conn) {
+	var underlyingTCPConn *net.TCPConn
+	if tlsConn, ok := conn.(*tls.Conn); ok {
+		netConnection := tlsConn.NetConn()
+		if tcp, isTCPConn := netConnection.(*net.TCPConn); isTCPConn {
+			underlyingTCPConn = tcp
+		} else {
+			WarnfCtx(ctx, "Couldn't turn off NODELAY for conn %v: underlying connection is not type *net.TCPConn", conn)
+		}
```
bbrks: We're having to do this `TCPConn` type assertion twice due to the structure of these conditions.

I think this function could do 3 distinct steps to reduce the duplication, and reduce the nesting by using early returns. Something like this:

```go
var netConn = conn

// 1. Unwrap `net.Conn` for known wrappers (only `tls.Conn` as of now)
if tlsConn, ok := netConn.(*tls.Conn); ok {
	netConn = tlsConn.NetConn()
}

// 2. Type assert `net.TCPConn` from unwrapped `net.Conn`
tcpConn, ok := netConn.(*net.TCPConn)
if !ok {
	// 2.a Or warn/return if we still don't have a `net.TCPConn`, making sure to keep `%T` in the warning, so we know what type caused that type assertion to fail!
	WarnfCtx(ctx, "Couldn't turn off NODELAY for %v: %T is not type *net.TCPConn", conn, conn)
	return
}

// 3. `SetNoDelay`
if err := tcpConn.SetNoDelay(false); err != nil {
	WarnfCtx(ctx, "Couldn't turn off NODELAY for %v: %v", conn, err)
}
```

bbrks (changes requested):
Have left a suggestion for further cleanup

bbrks (approved):
:+1:

PR #6313, opened 2023-06-15
CBG-2927: Total Sync Time stat

gregns1:
CBG-2927

Small change to add a new stat called TotalSyncTime. This is implemneted through a background task that runs each seocnd and incremnets the stat by the NumReplicationsActive stat each time. 
NOTE: due to timing issues the unit test cannot assert on the exact value of the stat so rather just asserts that the stat increases after a wait for a few seconds. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1898/

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -2310,6 +2310,16 @@ func (db *DatabaseContext) StartOnlineProcesses(ctx context.Context) (returnedEr
 
 	}
 
+	// create a background task to keep track of the number of active replication connections the database has each second
+	bgtSyncTime, err := NewBackgroundTask(ctx, "TotalSyncTimeStat", func(ctx context.Context) error {
+		db.UpdateTotalSyncTimeStat()
+		return nil
+	}, 1*time.Second, db.terminator)
+	if err != nil {
+		return err
+	}
+	db.backgroundTasks = append(db.backgroundTasks, bgtSyncTime)
+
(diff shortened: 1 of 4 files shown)


db/database.go
```diff
@@ -2310,6 +2310,16 @@ func (db *DatabaseContext) StartOnlineProcesses(ctx context.Context) (returnedEr
 
 	}
 
+	// create a background task to keep track of the number of active replication connections the database has each second
+	bgtSyncTime, err := NewBackgroundTask(ctx, "TotalSyncTimeStat", func(ctx context.Context) error {
+		db.UpdateTotalSyncTimeStat()
+		return nil
+	}, 1*time.Second, db.terminator)
```
torcolvin, line 2317: If you make time a parameter in `DatabaseContextOptions` we can tweak it to make the test run faster.

gregns1: changed the test instead to make it faster

rest/replicatortest/replicator_test.go
```diff
@@ -2468,6 +2468,43 @@ func TestReconnectReplicator(t *testing.T) {
 
 }
 
+// TestTotalSyncTimeStat:
```
torcolvin, line 2471: Can you write a test for blip (CBL) replications too?

torcolvin (approved):

PR #6314, opened 2023-06-16
CBG-2837 Support async database initialization

adamcfraser:
When a database is started in offline mode, allows all offline REST endpoints to be used while index initialization occurs asynchronously.

When the database is taken online via config when async initialization is in progress, sets db state to 'Starting' and does not block.  When initialization completes, database state is moved to 'Online'.

Changes to the collection set while offline will cancel and restart the async initialization with the updated set.

Initialization creation status is not persisted (as we do with background processes), because we want to re-validate index existence on every database startup.  If async processing completes, any subsequent database start/reload will check for index existence in the usual way (checking system:indexes).

CBG-2837

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1927/

### DIFF
--- a/base/collection.go
+++ b/base/collection.go
@@ -32,8 +32,64 @@ import (
 // GetGoCBv2Bucket opens a connection to the Couchbase cluster and returns a *GocbV2Bucket for the specified BucketSpec.
 func GetGoCBv2Bucket(spec BucketSpec) (*GocbV2Bucket, error) {
 
-	cluster, err := GetGocbCluster(spec)
+	logCtx := context.TODO()
+	connString, err := spec.GetGoCBConnString(nil)
 	if err != nil {
+		WarnfCtx(logCtx, "Unable to parse server value: %s error: %v", SD(spec.Server), err)
+		return nil, err
+	}
+
+	securityConfig, err := GoCBv2SecurityConfig(&spec.TLSSkipVerify, spec.CACertPath)
+	if err != nil {
+		return nil, err
+	}
+
+	authenticator, err := spec.GocbAuthenticator()
+	if err != nil {
+		return nil, err
+	}
+
+	if _, ok := authenticator.(gocb.CertificateAuthenticator); ok {
+		InfofCtx(logCtx, KeyAuth, "Using cert authentication for bucket %s on %s", MD(spec.BucketName), MD(spec.Server))
+	} else {
+		InfofCtx(logCtx, KeyAuth, "Using credential authentication for bucket %s on %s", MD(spec.BucketName), MD(spec.Server))
+	}
+
+	timeoutsConfig := GoCBv2TimeoutsConfig(spec.BucketOpTimeout, StdlibDurationPtr(spec.GetViewQueryTimeout()))
+	InfofCtx(logCtx, KeyAll, "Setting query timeouts for bucket %s to %v", spec.BucketName, timeoutsConfig.QueryTimeout)
+
+	clusterOptions := gocb.ClusterOptions{
+		Authenticator:  authenticator,
+		SecurityConfig: securityConfig,
+		TimeoutsConfig: timeoutsConfig,
+		RetryStrategy:  gocb.NewBestEffortRetryStrategy(nil),
+	}
+
+	if spec.KvPoolSize > 0 {
+		// TODO: Equivalent of kvPoolSize in gocb v2?
+	}
+
+	cluster, err := gocb.Connect(connString, clusterOptions)
+	if err != nil {
+		InfofCtx(logCtx, KeyAuth, "Unable to connect to cluster: %v", err)
+		return nil, err
+	}
+
+	err = cluster.WaitUntilReady(time.Second*5, &gocb.WaitUntilReadyOptions{
+		DesiredState:  gocb.ClusterStateOnline,
+		ServiceTypes:  []gocb.ServiceType{gocb.ServiceTypeManagement},
+		RetryStrategy: &goCBv2FailFastRetryStrategy{},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	if err != nil {
+		_ = cluster.Close(nil)
+		if errors.Is(err, gocb.ErrAuthenticationFailure) {
+			return nil, ErrAuthError
+		}
+		WarnfCtx(context.TODO(), "Error waiting for cluster to be ready: %v", err)
 		return nil, err
 	}
 
@@ -122,68 +178,6 @@ func GetGocbV2BucketFromCluster(cluster *gocb.Cluster, spec BucketSpec, waitUnti
 	gocbv2Bucket.kvOps = make(chan struct{}, MaxConcurrentSingleOps*nodeCount*numPools)
 
 	return gocbv2Bucket, nil
-}
-
-// GetGocbCluster opens a connection to the Couchbase cluster and returns a *gocb.Cluster for the specified BucketSpec.
-func GetGocbCluster(spec BucketSpec) (*gocb.Cluster, error) {
-
-	logCtx := context.TODO()
-	connString, err := spec.GetGoCBConnString(nil)
-	if err != nil {
-		WarnfCtx(logCtx, "Unable to parse server value: %s error: %v", SD(spec.Server), err)
-		return nil, err
-	}
-
-	securityConfig, err := GoCBv2SecurityConfig(&spec.TLSSkipVerify, spec.CACertPath)
-	if err != nil {
-		return nil, err
-	}
-
-	authenticator, err := spec.GocbAuthenticator()
-	if err != nil {
-		return nil, err
-	}
-
-	if _, ok := authenticator.(gocb.CertificateAuthenticator); ok {
-		InfofCtx(logCtx, KeyAuth, "Using cert authentication for bucket %s on %s", MD(spec.BucketName), MD(spec.Server))
-	} else {
-		InfofCtx(logCtx, KeyAuth, "Using credential authentication for bucket %s on %s", MD(spec.BucketName), MD(spec.Server))
-	}
-
-	timeoutsConfig := GoCBv2TimeoutsConfig(spec.BucketOpTimeout, StdlibDurationPtr(spec.GetViewQueryTimeout()))
-	InfofCtx(logCtx, KeyAll, "Setting query timeouts for bucket %s to %v", spec.BucketName, timeoutsConfig.QueryTimeout)
-
-	clusterOptions := gocb.ClusterOptions{
-		Authenticator:  authenticator,
-		SecurityConfig: securityConfig,
-		TimeoutsConfig: timeoutsConfig,
-		RetryStrategy:  gocb.NewBestEffortRetryStrategy(nil),
-	}
-
-	if spec.KvPoolSize > 0 {
-		// TODO: Equivalent of kvPoolSize in gocb v2?
-	}
-
-	cluster, err := gocb.Connect(connString, clusterOptions)
-	if err != nil {
-		InfofCtx(logCtx, KeyAuth, "Unable to connect to cluster: %v", err)
-		return nil, err
-	}
-
-	err = cluster.WaitUntilReady(time.Second*5, &gocb.WaitUntilReadyOptions{
-		DesiredState:  gocb.ClusterStateOnline,
-		ServiceTypes:  []gocb.ServiceType{gocb.ServiceTypeManagement},
-		RetryStrategy: &goCBv2FailFastRetryStrategy{},
-	})
-	if err != nil {
-		_ = cluster.Close(nil)
-		if errors.Is(err, gocb.ErrAuthenticationFailure) {
-			return nil, ErrAuthError
-		}
-		WarnfCtx(context.TODO(), "Error waiting for cluster to be ready: %v", err)
-		return nil, err
-	}
-	return cluster, nil
 }
 
 type GocbV2Bucket struct {
--- a/base/collection.go
+++ b/base/collection.go
@@ -32,60 +32,8 @@ import (
 // GetGoCBv2Bucket opens a connection to the Couchbase cluster and returns a *GocbV2Bucket for the specified BucketSpec.
 func GetGoCBv2Bucket(spec BucketSpec) (*GocbV2Bucket, error) {
 
-	logCtx := context.TODO()
-	connString, err := spec.GetGoCBConnString(nil)
+	cluster, err := GetGocbCluster(spec)
 	if err != nil {
-		WarnfCtx(logCtx, "Unable to parse server value: %s error: %v", SD(spec.Server), err)
-		return nil, err
-	}
-
-	securityConfig, err := GoCBv2SecurityConfig(&spec.TLSSkipVerify, spec.CACertPath)
-	if err != nil {
-		return nil, err
-	}
-
-	authenticator, err := spec.GocbAuthenticator()
-	if err != nil {
-		return nil, err
-	}
-
-	if _, ok := authenticator.(gocb.CertificateAuthenticator); ok {
-		InfofCtx(logCtx, KeyAuth, "Using cert authentication for bucket %s on %s", MD(spec.BucketName), MD(spec.Server))
-	} else {
-		InfofCtx(logCtx, KeyAuth, "Using credential authentication for bucket %s on %s", MD(spec.BucketName), MD(spec.Server))
-	}
-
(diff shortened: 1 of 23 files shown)


rest/database_init_manager.go
```diff
@@ -0,0 +1,312 @@
...
+	// TODO: CBG-2838 refactor initialize indexes to reduce number of system:indexes calls
+	var indexErr error
+	for scName, indexSet := range w.collections {
+		// Add the index set to the common indexOptions
+		collectionIndexOptions := w.options.indexOptions
+		collectionIndexOptions.MetadataIndexes = indexSet
+
+		// TODO: CBG-2838 Refactor InitializeIndexes API to move scope, collection to parameters on system:indexes calls
+		// Set the scope and collection name on the cluster n1ql store for use by initializeIndexes
+		w.n1qlStore.SetScopeAndCollection(scName)
```
torcolvin, line 235: can you redefine ctx here?

```
ctx := base.KeyspaceLogCtx(ctx, bucket.GetName(), dsName.ScopeName(), dsName.CollectionName())
```

rest/database_init_manager.go
```diff
@@ -0,0 +1,312 @@
...
+	w.watcherLock.Lock()
+	w.lastError = indexErr
+	for _, doneChan := range w.watchers {
+		if indexErr != nil {
+			doneChan <- indexErr
+		}
+		close(doneChan)
+	}
+	w.completed = true
+	w.watcherLock.Unlock()
```
torcolvin: I don't think this will panic but is it worth adding unlock in a defer?

rest/server_context.go
```diff
@@ -518,8 +525,16 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
...
+		scopeAndCollection base.ScopeAndCollectionName
+		indexSet           db.CollectionIndexesType
+		datastore          base.DataStore
+	}
+
+	collectionsRequiringIndexes := make([]indexInitData, 0)
+
 	// initDataStore is a function to initialize Views or GSI indexes for a datastore
-	initDataStore := func(ds base.DataStore, metadataIndexes db.CollectionIndexesType) (err error) {
+	initDataStore := func(ds base.DataStore, metadataIndexes db.CollectionIndexesType, name base.ScopeAndCollectionName) (err error) {
```
torcolvin, line 537: you can derive scope and collection information from a datastore, albeit in an annoying way https://github.com/couchbase/sync_gateway/blob/b20d5152bd4979620ce0098d491c5e6e454af2db/base/bucket.go#L97

rest/server_context.go
```diff
@@ -628,11 +618,60 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
...
 		}
 	}
 
+	startOffline := base.BoolDefault(config.StartOffline, false)
+	var dbInitDoneChan chan error
+	// Initialize any required indexes
+	if len(collectionsRequiringIndexes) > 0 {
+		gsiSupported := bucket.IsSupported(sgbucket.BucketStoreFeatureN1ql)
+		if !gsiSupported {
+			return nil, errors.New("Sync Gateway was unable to connect to a query node on the provided Couchbase Server cluster.  Ensure a query node is accessible, or set 'use_views':true in Sync Gateway's database config.")
```
torcolvin, line 632: is it worth suggesting only a query node here, or checking that there's only a default collection before suggesting views?

I'd love to _not_ suggest views here.

Or should we move this to a db validation, or do we not talk to server then?

rest/server_context.go
```diff
@@ -628,11 +618,60 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
...
+		}
+
+		// If database has been requested to start offline, or there's an active async initialization, use async initialization
+		if sc.DatabaseInitManager != nil && (startOffline || sc.DatabaseInitManager.HasActiveInitialization(dbName)) {
```
torcolvin, line 637: If startOffline is requested, should we bail out if `DatabaseInitManager` is not nil?

torcolvin (commented):
There's some code that is duplicated with code I've put in 3.1 - like dropAllIndexes https://github.com/couchbase/sync_gateway/blob/master/db/indextest/main_test.go#L53

that would be cool to combine but for the most part the only part that seems duplicated is the calls to initializeIndexes synchronous and not (with channel reporting in one case and not in another case).

This was not a complete code review.

base/collection.go
```diff
@@ -32,60 +32,8 @@ import (
...
-		InfofCtx(logCtx, KeyAuth, "Unable to connect to cluster: %v", err)
-		return nil, err
-	}
-
-	err = cluster.WaitUntilReady(time.Second*5, &gocb.WaitUntilReadyOptions{
-		DesiredState:  gocb.ClusterStateOnline,
-		ServiceTypes:  []gocb.ServiceType{gocb.ServiceTypeManagement},
-		RetryStrategy: &goCBv2FailFastRetryStrategy{},
-	})
+	cluster, err := GetGocbCluster(spec)
```
torcolvin: I don't object to this refactor but it is not used.

base/collection.go
```diff
@@ -176,6 +124,68 @@ func GetGocbV2BucketFromCluster(cluster *gocb.Cluster, spec BucketSpec, waitUnti
 	return gocbv2Bucket, nil
 }
 
+// GetGocbCluster opens a connection to the Couchbase cluster and returns a *gocb.Cluster for the specified BucketSpec.
+func GetGocbCluster(spec BucketSpec) (*gocb.Cluster, error) {
```
torcolvin: possible fix: passing a context into this for multiple dbs might be nice for logging

rest/database_init_manager.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	bucketName := dbConfig.Name
+	if dbConfig.Bucket != nil {
+		bucketName = *dbConfig.Bucket
+	}
+
+	// Initialize ClusterN1QLStore for the bucket.  Scope and collection name are set
+	// per-operation
+	n1qlStore, err := couchbaseCluster.GetClusterN1QLStore(bucketName, "", "")
```
torcolvin, line 89: If these are set per operation, does it make sense to not pass scope and collection into the cluster and require it later only?

These are going to get separate agents anyway, it's almost like ClusterN1QLStore might be a misnomer until we can query system:indexes without a Scope based query.

rest/database_init_manager.go
```diff
@@ -0,0 +1,312 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	indexOptions := m.BuildIndexOptions(startupConfig, dbConfig)
```
torcolvin, line 94: can we use this function in server_context.go for synchronous build also?

rest/database_init_manager.go
```diff
@@ -0,0 +1,312 @@
...
+	worker, ok := m.workers[dbName]
+	if !ok {
+		return
+	}
+	worker.Stop()
+}
+
+// buildCollectionIndexData determines the set of indexes required for each collection in the config, including
+// the metadata collection
+func (m *DatabaseInitManager) buildCollectionIndexData(config *DatabaseConfig) CollectionInitData {
```
torcolvin, line 155: this is again duplicate code from server_context.go - i'm a bit less concerned about this vs index options.

We discussed offline about an enhancement on the main branch to make all the code use async pathway and just block on all workers being done.

rest/database_init_manager.go
```diff
@@ -0,0 +1,312 @@
...
+	w.watcherLock.Lock()
+	defer w.watcherLock.Unlock()
+	if w.completed {
+		// If the worker has completed while we acquired the lock, return any error and close channel
+		doneChan = make(chan error, 1)
+		if w.lastError != nil {
+			doneChan <- w.lastError
+		}
+		close(doneChan)
+		return doneChan, nil
```
torcolvin: Is it correct to return nil in this case, if know it failed?

rest/server_context.go
```diff
@@ -628,11 +618,60 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
...
+	startOffline := base.BoolDefault(config.StartOffline, false)
+	var dbInitDoneChan chan error
+	// Initialize any required indexes
+	if len(collectionsRequiringIndexes) > 0 {
+		gsiSupported := bucket.IsSupported(sgbucket.BucketStoreFeatureN1ql)
+		if !gsiSupported {
+			return nil, errors.New("Sync Gateway was unable to connect to a query node on the provided Couchbase Server cluster.  Ensure a query node is accessible, or set 'use_views':true in Sync Gateway's database config.")
+		}
+
+		// If database has been requested to start offline, or there's an active async initialization, use async initialization
```
torcolvin, line 635: Add something here about the nil case being true when persistent config is not used?

rest/server_context.go
```diff
@@ -839,14 +868,74 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
 		return dbcontext, nil
 	}
 
-	atomic.StoreUint32(&dbcontext.State, db.DBOnline)
-	_ = dbcontext.EventMgr.RaiseDBStateChangeEvent(dbName, "online", stateChangeMsg, &sc.Config.API.AdminInterface)
+	// If asyncOnline wasn't specified, block until db init is completed, then start online processes
+	if !options.asyncOnline {
+		base.DebugfCtx(ctx, base.KeyConfig, "Waiting for database init to complete...")
+		if dbInitDoneChan != nil {
```
torcolvin, line 875: I don't know if you can hit this pathway, but do you need to make sure we aren't using persistent config here?

rest/server_context.go
```diff
@@ -839,14 +868,74 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
...
+// db online (it may have been set to offline=true)
+func (sc *ServerContext) asyncDatabaseOnline(nonCancelCtx base.NonCancellableContext, dbc *db.DatabaseContext, doneChan chan error, version string) {
+
+	ctx := nonCancelCtx.Ctx
+	if doneChan != nil {
+		initError := <-doneChan
+		if initError != nil {
+			base.WarnfCtx(ctx, "Async database init returned error: %v", initError)
+			return
+		}
```
torcolvin, line 911: is there an error state of failed to come online? do we need to do a state transition of db?

rest/server_context.go
```diff
@@ -839,14 +868,74 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
...
+	}
+
+	// Before bringing the database online, ensure that the database hasn't been modified while we waited for initialization to complete
+	currentDbVersion := sc.GetDbVersion(dbc.Name)
+	if currentDbVersion != version {
+		base.InfofCtx(ctx, base.KeyConfig, "Database version changed while waiting for aync init - cancelling obsolete online request. Old version: %s New version: %s", version, currentDbVersion)
```
torcolvin:
```suggestion
		base.InfofCtx(ctx, base.KeyConfig, "Database version changed while waiting for async init - cancelling obsolete online request. Old version: %s New version: %s", version, currentDbVersion)
```

We possibly need to mark a state change on the db here too.

rest/server_context.go
```diff
@@ -839,14 +868,74 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
...
+		return
+	}
+
+	base.DebugfCtx(ctx, base.KeyConfig, "Async database initialization complete, starting online processes...")
+	err := dbc.StartOnlineProcesses(ctx)
+	if err != nil {
+		base.InfofCtx(ctx, base.KeyAll, "Error starting online processes after async initialization: %v", err)
```
torcolvin, line 925: Why don't we return an error here?

torcolvin (commented):
I didn't look at the moving of the installPrincipals code yet..

base/collection.go
adamcfraser: Thanks for catching - I think it makes sense to revert this since it didn't end up being needed.

base/collection.go
adamcfraser: Reverting this, so not going to add this enhancement in this PR.

rest/database_init_manager.go
adamcfraser: I was envisioning a future refactor to remove IndexMetaBucketID()/IndexMetaScopeID()/IndexMetaKeyspaceID() from the N1QLStore API, and replace those with parameters that are passed to IndexMeta operations.  That would avoid this mild hack of having to set scope and collection name on the ClusterN1QLStore per operation.

I agree cluster N1QL store might be a misnomer now because of the need to perform scope-limiting for future compatibility.  When you say 'separate agents anyway', I think that only applies per scope, right?  (gocb is going to cache agents per scope, not per collection?)

rest/database_init_manager.go
adamcfraser: Left this code duplicated for now to reduce amount of change/risk for synchronous handling, but agree this makes sense.  Mentioned this on CBG-3158 for followup.

rest/database_init_manager.go
adamcfraser: Filed CBG-3158 to track this.

rest/database_init_manager.go
adamcfraser: Makes sense.  (had to add a BucketName() method to the N1QLStore API to do this, but I think it's worthwhile)

rest/database_init_manager.go
adamcfraser: That's reasonable - made that change.

rest/database_init_manager.go
adamcfraser: I think returning nil is correct ("addWatcher" was successful, and the worker status is returned on the doneChan as expected).  Looking at the code I think that means that we never return error from this method, and should just remove the error from the function signature.

rest/server_context.go
adamcfraser: This is an existing error message that just got moved around in this PR, so will probably prefer to leave this unchanged for now.

rest/server_context.go
adamcfraser: I think that combination is valid and still works (file-based config, with offline=true in the config)

rest/server_context.go
adamcfraser: Do you mean clarify that we *are* using persistent config?  You might be right that an additional check is needed, just trying to work out which combination you're thinking of.

rest/server_context.go
adamcfraser: Good point, this needs more consideration.

base/collection.go
```diff
@@ -80,6 +80,10 @@ func GetGoCBv2Bucket(spec BucketSpec) (*GocbV2Bucket, error) {
 		ServiceTypes:  []gocb.ServiceType{gocb.ServiceTypeManagement},
 		RetryStrategy: &goCBv2FailFastRetryStrategy{},
 	})
+	if err != nil {
+		return nil, err
+	}
+
 	if err != nil {
```
bbrks: 2nd error handling block never going to be hit - I think this was added accidentally, as it's going to leave the cluster connection around?

bbrks (changes requested):
Nothing other than the inline comment jumped out at me, but I'd like to spend a bit of time using/manually testing it to get a better feel for how it works in practice.

rest/database_init_manager.go
torcolvin: It should be one agent per scope, I agree.

rest/server_context.go
torcolvin: I guess:

persistent config = true , then dbInitDoneChan should be non nil
persistent config = false, then dbInitDoneChan is nil

So maybe this looks okay.

base/collection.go
adamcfraser: Thanks for catching - this shouldn't be here.

rest/server_context.go
adamcfraser: asyncDatabaseOnline doesn't return error since it's getting run in a goroutine that wouldn't have the ability to do anything other than log the error in any case.

rest/server_context.go
adamcfraser: Updated this to take the db back offline if we get an error from index init.

bbrks (approved):
I've spent a fair bit of time playing with the feature locally and am happy with this 👍

torcolvin (approved):

PR #6317, opened 2023-06-20
CBG-3108 only log relevant SG indexes

torcolvin:
In the first commit, just log SG only indexes. I also add which indexes _aren't_ ready yet so we know which indexes are slow to create.

I think the second commit is more comprehensive about logging but might be too much for 3.1.1

## Pre-review checklist

### DIFF
--- a/base/bucket_n1ql_test.go
+++ b/base/bucket_n1ql_test.go
@@ -11,6 +11,7 @@ licenses/APL2.txt.
 package base
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"testing"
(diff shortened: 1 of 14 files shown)


rest/config.go
```diff
@@ -478,6 +478,9 @@ func readFromPath(path string, insecureSkipVerify bool) (rc io.ReadCloser, err e
 
 func (dbConfig *DbConfig) AutoImportEnabled() (bool, error) {
 	if dbConfig.AutoImport == nil {
+		if !dbConfig.UseXattrs() {
```
adamcfraser: Is this change required for this PR?  It seems unrelated.

adamcfraser (changes requested):

rest/config.go
torcolvin: This change already exists on master and if I do a `git diff origin/master` I don't see this change so I don't know what github is doing here.

adamcfraser (approved):

PR #6318, opened 2023-06-21
CBG-2933: sync compute stat 

gregns1:
CBG-2933

Adds compute stat for sync compute. 
- Addded calculation to register function on the blip handler for all incoming blip messages. 
- Added calculation to where we cann sendmessage on blip to handle outgoing blip messages. 
- Test coverage of both ISGR and CBL replications 
- Noticed a mistake in import process compute stat, it had the shared bucket import key on it and it hsould be the databse key, so added that change.
- Calculation is done in helper method on the blipsynccontext object 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1934/

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -560,7 +560,10 @@ func (bh *blipHandler) sendBatchOfChanges(sender *blip.Sender, changeArray [][]i
 			return
 		}
 		stat := CalculateComputeStat(int64(bytes), functionTime)
+		// increment associated blip stat
 		bh.replicationStats.DocWriteComputeUnit.Add(stat)
+		// increment stat on the database stats
+		bh.blipContextDb.DbStats.DatabaseStats.DocWriteComputeUnit.Add(stat)
 	}()
 
 	if len(changeArray) > 0 {
@@ -666,7 +669,10 @@ func (bh *blipHandler) handleChanges(rq *blip.Message) error {
 			return
 		}
 		stat := CalculateComputeStat(int64(bytes), functionTime)
+		// increment associated blip stat
 		bh.replicationStats.DocCheckComputeUnit.Add(stat)
+		// increment stat on the database stats
+		bh.blipContextDb.DbStats.DatabaseStats.DocCheckComputeUnit.Add(stat)
 	}()
 	bh.replicationStats.HandleChangesCount.Add(int64(len(changeList)))
 	defer func() {
@@ -806,7 +812,10 @@ func (bh *blipHandler) handleProposeChanges(rq *blip.Message) error {
 			return
 		}
 		stat := CalculateComputeStat(int64(bytes), functionTime)
+		// increment associated blip stat
 		bh.replicationStats.DocCheckComputeUnit.Add(stat)
+		// increment stat on the database stats
+		bh.blipContextDb.DbStats.DatabaseStats.DocCheckComputeUnit.Add(stat)
 	}()
 	bh.replicationStats.HandleChangesCount.Add(int64(len(changeList)))
 	defer func() {
@@ -969,7 +978,10 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 		}
 		functionTime := time.Since(startTime).Milliseconds()
 		stat := CalculateComputeStat(int64(bytes), functionTime)
+		// increment associated blip stat
 		bh.replicationStats.DocReadComputeUnit.Add(stat)
+		// increment stat on the database stats
+		bh.blipContextDb.DbStats.DatabaseStats.DocReadComputeUnit.Add(stat)
 	}()
 
 	// addRevisionParams := newAddRevisionParams(rq)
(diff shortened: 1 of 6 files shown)


db/blip_sync_context.go
```diff
@@ -707,3 +719,55 @@ func (bsc *BlipSyncContext) reportStats(updateImmediately bool) {
...
+	if bsc.blipContextDb == nil || bsc.blipContext == nil {
+		return
+	}
+	dbStats := bsc.blipContextDb.DbStats.Database()
+	if dbStats == nil {
+		return
+	}
+	blipInterval := time.Duration(bsc.blipContextDb.Options.BlipStatsReportingInterval)
+	bsc.loadStats()
+	go func() {
```
torcolvin: My understanding is that we really don't want to add a goroutine per replicator but I think we can do this in the middleware or in each function if required. I think you don't need this stat for anything except where we explicitly send and receive a blip message without going through the handler code (in sendChanges).

I think we do have to think about double counting again here, but

db/blip_sync_context.go
```diff
@@ -144,6 +144,7 @@ func (bsc *BlipSyncContext) SetClientType(clientType BLIPSyncContextClientType)
 // Includes the outer handler as a nested function.
 func (bsc *BlipSyncContext) register(profile string, handlerFn func(*blipHandler, *blip.Message) error) {
 
+	startTime := time.Now()
```
torcolvin: I believe the start of the function will need to be at line 150.

I'd probably wrap the reporting into a function 

```
func (*bsc) reportComputeStats(request *blip.Message, startTime time.Time)
```

so it can we handled even in the case of a panic, and we can reuse this function in the blipHandler since `blipSyncContext` is an embedded struct in `blipHandler`.

db/blip_handler.go
```diff
@@ -1311,6 +1327,13 @@ func (bh *blipHandler) sendGetAttachment(sender *blip.Sender, docID string, name
 	if !bh.sendBLIPMessage(sender, outrq) {
 		return nil, ErrClosedBLIPSender
 	}
+	messageCPUtime := time.Since(startTime).Milliseconds()
```
torcolvin: this comes from `processRev` if I'm ready this code correctly, and it's worth documenting how you can get here.

This will be run while `processRev` handler runs, but it the handler in `register` is running for the revision. I think this is fine.

db/blip_handler.go
```diff
@@ -1357,6 +1381,13 @@ func (bh *blipHandler) sendProveAttachment(sender *blip.Sender, docID, name, dig
 	if !bh.sendBLIPMessage(sender, outrq) {
 		return ErrClosedBLIPSender
 	}
+	messageCPUtime := time.Since(startTime).Milliseconds()
+	messBytes, err := outrq.Body()
+	if err == nil && bh.blipContextDb != nil {
+		bytes := len(messBytes)
+		stat := CalculateComputeStat(int64(bytes), messageCPUtime)
+		bh.blipContextDb.DbStats.Database().SyncProcessCompute.Add(stat)
+	}
```
torcolvin: as above, I think this comes from `processRev`, can you verify my assumption and add a comment?

db/blip_sync_context.go
```diff
@@ -585,6 +601,13 @@ func (bsc *BlipSyncContext) sendNoRev(sender *blip.Sender, docID, revID string,
 	if !bsc.sendBLIPMessage(sender, noRevRq.Message) {
 		return ErrClosedBLIPSender
 	}
+	messageCPUtime := time.Since(startTime).Milliseconds()
+	messBytes, err := noRevRq.Body()
+	if err == nil && bsc.blipContextDb != nil {
+		bytes := len(messBytes)
+		stat := CalculateComputeStat(int64(bytes), messageCPUtime)
+		bsc.blipContextDb.DbStats.Database().SyncProcessCompute.Add(stat)
+	}
```
torcolvin: This code is called from `sendBatchOfChanges` - I wonder if this whole code is going to have such a short period of execution time whether it's worth tracking this separately from sendBatchOfChanges. I'd be interested in someone else's call on this.

gregns1: I think this is called from `sendRevision` which in turn is called from `handleChangesResponse` so I feel like is needed even though I agree it will be very small.

torcolvin: SGTM, I think just write a comment about it.

db/blip_handler.go
gregns1: Good point this I guess is already being measured in the handler code, I feel like this should be removed then? But I guess its new messge being sent so should still be counted. Ill add the comments to expalin this. Good spot.

db/blip_sync_context.go
```diff
@@ -707,3 +712,13 @@ func (bsc *BlipSyncContext) reportStats(updateImmediately bool) {
 	bsc.stats.lastReportTime.Store(currentTime)
 
 }
+
+func (bsc *BlipSyncContext) reportComputeStat(rq *blip.Message, startTime time.Time) {
```
torcolvin, line 717: I'd add a comment to this.

```
reportComputeStat will report the amount of data transferred for a blip Message. This message can be a request or a response.
```

db/blip_sync_context.go
```diff
@@ -707,3 +712,13 @@ func (bsc *BlipSyncContext) reportStats(updateImmediately bool) {
...
+	messageCPUtime := time.Since(startTime).Milliseconds()
+	messBody, err := rq.Body()
+	if err == nil && bsc.blipContextDb != nil {
+		bytes := len(messBody)
```
torcolvin, line 721: if bytes is zero, let's early return here. I think you _might_ be able to have message with empty Body() and only Properties, but you can check that by constructing a message.

torcolvin (commented):
LGTM, two tiny comments.

torcolvin (approved):

PR #6319, opened 2023-06-22
Add const to allow dev-time disabling of gocb log level remapping

bbrks:
Allows easier dev-time evaluation of what GoCB logs look like if we don't remap the log levels.

Initial findings:
- Info level seems OK to keep as info, without dropping to debug, with the exception of dumping structs for "Creating new agent".
- Debug is very noisy with vbucket and cluster information every second per agent/CCCPOLL but that could possibly be moved to trace inside GoCB, allowing us more info at debug?

### DIFF
--- a/base/logger_external.go
+++ b/base/logger_external.go
@@ -21,11 +21,19 @@ import (
 	"github.com/couchbase/goutils/logging"
 )
 
+// remapGoCBLogLevels controls whether the gocb and gocbcore log levels are remapped to match the verbosity of SG's log levels.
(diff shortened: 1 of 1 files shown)


torcolvin (commented):
This seems fine. We have a lot of agents running and I think the CCCP polling is where we get penalized for this.

- persistent config polling
- import DCP feed
- mutation DCP feed
- bucket connection for database

So having those at a separable level seems like it would be best.

torcolvin (approved):

PR #6320, opened 2023-06-22
CBG-3128: Prevent unnecessary user updates when using OIDC claims to grant channel/roles

bbrks:
CBG-3128

- Add `TestOpenIDConnectImplicitFlowReuseToken` to reproduce identical OIDC claims causing user updates due to JWTLastUpdated changing prematurely.
- Only set `JWTLastUpdated` when actually updating a user as the result of access changes.
- Test util: Add `AssertTimestampGreaterThan` / `AssertTimeGreaterThan` to centralise Windows time resolution hack.
- Fix `AllOrNoneNil` handling for typed nils (and cover with unit test)

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1912/

### DIFF
--- a/db/users.go
+++ b/db/users.go
@@ -44,7 +44,7 @@ func (db *DatabaseContext) DeleteRole(ctx context.Context, name string, purge bo
 // UpdatePrincipal updates or creates a principal from a PrincipalConfig structure.
 func (dbc *DatabaseContext) UpdatePrincipal(ctx context.Context, updates *auth.PrincipalConfig, isUser bool, allowReplace bool) (replaced bool, err error) {
 	// Sanity checking
-	if !base.AllOrNoneNil(updates.JWTIssuer, updates.JWTRoles, updates.JWTChannels, updates.JWTLastUpdated) {
+	if !base.AllOrNoneNil(updates.JWTIssuer, updates.JWTRoles, updates.JWTChannels) {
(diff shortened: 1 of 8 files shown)


db/users.go
```diff
@@ -211,9 +208,7 @@ func (dbc *DatabaseContext) UpdatePrincipal(ctx context.Context, updates *auth.P
 			if updates.JWTChannels != nil && updatedJWTChannels.UpdateAtSequence(updates.JWTChannels, nextSeq) {
 				user.SetJWTChannels(updatedJWTChannels, nextSeq)
 			}
-			if updates.JWTLastUpdated != nil {
-				user.SetJWTLastUpdated(*updates.JWTLastUpdated)
-			}
+			user.SetJWTLastUpdated(time.Now())
```
adamcfraser: I don't think we want to set JWTLastUpdated if no other JWT properties have changed, do we?  i.e. we don't want to be updating this property if admin channels is the only thing being changed.

adamcfraser (changes requested):

db/users.go
bbrks: That's fair - I'll add a 2nd bool to track whether changes were the result of JWT claims.

adamcfraser (approved):

PR #6321, opened 2023-06-23
CBG-3127 call unsub changes in the unit test

torcolvin:
The test panics in walrus because the continuous changes feed is still running when the database is closed. In practice, it'd call `ReloadUser` and then the datastore would be closed and SG would log a warning and be done. However, walrus panics if you try to write to a closed bucket.

- Removed some verbose debugging to try to reduce our test log bloat.
- Remove sleep after running 100k with/without -race.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1914/

### DIFF
--- a/rest/blip_api_crud_test.go
+++ b/rest/blip_api_crud_test.go
@@ -155,19 +155,24 @@ func TestBlipPushRevisionInspectChanges(t *testing.T) {
 	assert.NoError(t, timeoutErr, "Timed out waiting")
 }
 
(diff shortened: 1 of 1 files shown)


bbrks (approved):
As discussed - happy to merge this to reduce chances of hitting CBG-3127 occurring but need a proper fix for cancellation/synchronous close of changes feeds without client intervention.

PR #6325, opened 2023-06-28
CBG-3110: Handle ErrImportCancelled case for OnDemandImportForGet

bbrks:
CBG-3110

- Prevents panic when `ErrImportCancelled` ends up causing `nil, nil` to be returned from `ImportDoc`.
- Adds test to reproduce - where import is called for a `nil` binary document.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1924/

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -77,6 +77,7 @@ func (c *DatabaseCollection) GetDocumentWithRaw(ctx context.Context, docid strin
 			if importErr != nil {
 				return nil, nil, importErr
 			}
+			// nil, nil returned when ErrImportCancelled is swallowed by importDoc switch
 			if doc == nil {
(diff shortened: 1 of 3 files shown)


db/crud.go
```diff
@@ -77,6 +77,9 @@ func (c *DatabaseCollection) GetDocumentWithRaw(ctx context.Context, docid strin
 			if importErr != nil {
 				return nil, nil, importErr
 			}
+			if doc == nil {
```
torcolvin, line 81: Can you add a comment why importError is nil but doc is nil?

torcolvin (commented):
Seems right to me.

adamcfraser (approved):

PR #6326, opened 2023-06-28
CBG-3130 notify on request plus unused sequence docs

torcolvin:
Creates a fake key for unused sequences, and if request_plus is enabled, the changeWaiter will listen for this key.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1925/

### DIFF
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -33,8 +33,8 @@ const (
 	DefaultCachePendingSeqMaxWait = 5 * time.Second  // Max time we'll wait for a pending sequence before sending to missed queue
 	DefaultSkippedSeqMaxWait      = 60 * time.Minute // Max time we'll wait for an entry in the missing before purging
 	QueryTombstoneBatch           = 250              // Max number of tombstones checked per query during Compact
-	unusedSeqKey                  = "_unusedSeqKey"
-	unusedSeqCollectionID         = 0
+	unusedSeqKey                  = "_unusedSeqKey"  // Key used by ChangeWaiter to mark unused sequences
+	unusedSeqCollectionID         = 0                // Collection ID used by ChangeWaiter to mark unused sequences
 )
 
 // Enable keeping a channel-log for the "*" channel (channel.UserStarChannel). The only time this channel is needed is if
@@ -557,6 +557,7 @@ func (c *changeCache) releaseUnusedSequence(sequence uint64, timeReceived time.T
 	if c.notifyChange != nil && len(changedChannels) > 0 {
 		c.notifyChange(changedChannels)
 	}
+	c.channelCache.AddSkippedSequence(change)
 }
 
 // Process unused sequence notification.  Extracts sequence from docID and sends to cache for buffering
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -33,6 +33,8 @@ const (
 	DefaultCachePendingSeqMaxWait = 5 * time.Second  // Max time we'll wait for a pending sequence before sending to missed queue
 	DefaultSkippedSeqMaxWait      = 60 * time.Minute // Max time we'll wait for an entry in the missing before purging
 	QueryTombstoneBatch           = 250              // Max number of tombstones checked per query during Compact
+	unusedSeqKey                  = "_unusedSeqKey"
+	unusedSeqCollectionID         = 0
 )
 
 // Enable keeping a channel-log for the "*" channel (channel.UserStarChannel). The only time this channel is needed is if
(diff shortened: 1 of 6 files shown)


db/changes.go
```diff
@@ -672,6 +672,11 @@ func (col *DatabaseCollectionWithUser) SimpleMultiChangesFeed(ctx context.Contex
 			channelsSince = channels.AtSequence(chans, 0)
 		}
 
+		if options.RequestPlusSeq > 0 {
+			// QUESTION: why can't I use 0 for AtSequence here?
+			channelsSince.Add(channels.AtSequence(base.SetOf(unusedSeqKey), currentCachedSequence))
```
adamcfraser: I don't think you actually want to add the pseudo-channel to channelsSince, as that variable is used for more than just the call to changeWaiter.UpdateChannels().  You're right that changeWaiter doesn't use the AtSequence so it could be ignored, but by adding it to channelsSince you'd be doing things we don't want/need to do, like adding a channel feed on line 721/726 below.   

To avoid allocating a completely new set just to pass it to changeWaiter.UpdateChannels, I'm fine with modifying the changeWaiter.UpdateChannels add an optional parameter that would be a list of 'pseudo-channels'.  We wouldn't need to store those differently inside the changeWaiter (yet).

db/changes.go
```diff
@@ -1037,7 +1042,7 @@ func (col *DatabaseCollectionWithUser) SimpleMultiChangesFeed(ctx context.Contex
 			}
 			if userChanged && col.user != nil {
 				newChannelsSince, _ := col.user.FilterToAvailableCollectionChannels(col.ScopeName, col.Name, chans)
-
+				// QUESTION: updating channels here again will cause test to hang
```
adamcfraser: With the change above, shouldn't need to modify the set here.

adamcfraser (changes requested):
Approach looks good to me, seems like it might just need a small tweak

db/change_cache.go
```diff
@@ -546,9 +548,16 @@ func (c *changeCache) releaseUnusedSequence(sequence uint64, timeReceived time.T
...
+	unusedSeq := channels.NewID(unusedSeqKey, unusedSeqCollectionID)
+	if changedChannels == nil {
+		changedChannels = channels.SetOfNoValidate(unusedSeq)
+	} else {
+		changedChannels.Add(unusedSeq)
+	}
 	if c.notifyChange != nil && len(changedChannels) > 0 {
 		c.notifyChange(changedChannels)
 	}
+	c.channelCache.AddSkippedSequence(change)
```
adamcfraser: I expect we need to AddSkippedSequence to the cache before we call notifyChange, to avoid feeds waking up and checking the cache high sequence before we've updated it.

rest/changes_request_plus_test.go
```diff
@@ -0,0 +1,59 @@
...
+	unusedSeq, err := db.AllocateTestSequence(rt.GetDatabase())
+	require.NoError(t, err)
+
+	requestFinished := make(chan struct{})
+	go func() {
+		resp = rt.SendUserRequest(http.MethodGet, fmt.Sprintf("/{{.keyspace}}/_changes?since=%d&request_plus=true", docSeq), "", username)
+		RequireStatus(t, resp, http.StatusOK)
+		close(requestFinished)
+	}()
+	err = db.ReleaseTestSequence(rt.GetDatabase(), unusedSeq)
```
adamcfraser, line 60: I think there's a race here, where line 53 could execute before line 49 (or after line 49, but before the changes loop goes into wait mode).

Other request plus tests use NumPullReplTotalCaughtUp to wait for the changes request to go into wait mode before releasing the sequence.  (e.g. TestRequestPlusPull)

adamcfraser (changes requested):
Couple of minor suggestions, generally looks good.

rest/changes_request_plus_test.go
torcolvin: I changed this but one thing that confused me is that this can actually be _two_ sequences ahead where I'd expect it to be one.

adamcfraser: Do you see the same if you disable sequence batching?  I think I disabled it in the other request plus tests in order to ensure deterministic test results (using `defer db.SuspendSequenceBatching()()`)

torcolvin: Disabling sequence batching does fix the non deterministic problem, but I don't think you need to `WaitForTotalCaughtUp` unless you want to make sure that the matters because `db.AllocateTestSequence` is synchronous, and runs before the changes feed.

adamcfraser (approved):

PR #6332, opened 2023-07-12
CBG-3163: BytesRead for REST

gregns1:
CBG-3163

Adding a new struct that holds the request body to be ready by the request `CountedRequestReader` and this is now stored on the handler in place of the request body. 
New function that implements the io.Read method but adds the returned bytes read count to the `CountedRequestReader` struct. Another function to load this atomically before adding to the stat. The stat is incremented inside a function thats called in the handler code. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1941/


rest/handler.go
```diff
@@ -620,6 +620,35 @@ func (h *handler) logRESTCount() {
...
+func (h *handler) logBytesRead(bodyByteArray []byte) error {
+	if h.db == nil {
+		return nil
+	}
+	if h.privs != publicPrivs && h.privs != regularPrivs {
+		return nil
+	}
+	if h.isBlipSync() {
+		return nil
+	}
```
torcolvin: we can do `if h.shouldUpdateBytesTransferredStats()`

This just got merged today.

rest/bytes_read_public_api_test.go
```diff
@@ -0,0 +1,298 @@
...
+	"testing"
+
+	"github.com/couchbase/sync_gateway/base"
+	"github.com/couchbase/sync_gateway/channels"
+	"github.com/couchbase/sync_gateway/db"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBytesReadDcoOperations(t *testing.T) {
```
torcolvin:
```suggestion
func TestBytesReadDocOperations(t *testing.T) {
```

torcolvin (dismissed):

torcolvin (commented):
`ContentLength` is only populated if the client specifies a `Content-Length` in their header, which is true for curl and some clients but not all clients.

https://www.rfc-editor.org/rfc/rfc7230#section-3.3.2 describes how this is optional.

I think the way to address this is to have a custom io.Reader that have the underlying reader as a method will count bytes when you read them. This will override {{Read}} method and count the bytes, and probably have a flag to see if we've reported the value.

At the end of the handler, we can use whether it reported stats or not, and read the body if no one else reported then, like we do https://github.com/couchbase/sync_gateway/blob/master/rest/handler.go#L130

One thing we could do is make sure that `h.requestBody` always implements the `CountedRequestReader` when we finish a request.

I'd change `h.response.reportStats` to `h.reportStats` which will report stats on both request and response.

rest/bytes_read_public_api_test.go
```diff
@@ -0,0 +1,374 @@
...
+
+	// send admin request assert that the public rest count doesn't increase
+	resp = rt.SendAdminRequest(http.MethodGet, "/{{.keyspace}}/doc1", "")
+	RequireStatus(t, resp, http.StatusOK)
+	_, ok = base.WaitForStat(func() int64 {
+		return rt.GetDatabase().DbStats.DatabaseStats.PublicRestBytesRead.Value()
+	}, int64(len(inputBytes)))
+	require.True(t, ok)
+
+	// send user request that has empty body, asser the stat doesn't increase
```
torcolvin:
```suggestion
	// send user request that has empty body, assert the stat doesn't increase
```

rest/bytes_read_public_api_test.go
```diff
@@ -0,0 +1,374 @@
...
+	// test public endpoint but one that doesn't access a db and assert that doesn't increment stat
+	resp = rt.SendUserRequest(http.MethodGet, "/", "", "greg")
+	RequireStatus(t, resp, http.StatusOK)
+	// assert the stat doesn't increment
+	_, ok = base.WaitForStat(func() int64 {
+		return rt.GetDatabase().DbStats.DatabaseStats.PublicRestBytesRead.Value()
+	}, int64(len(inputBytes)))
+	require.True(t, ok)
+
+	// send another public request (this time POST) to check stat increments, but check it increments by correct bytes value
```
torcolvin, line 95: add a test here to make sure that a stat increases for a user who doesn't have access to the db

rest/bytes_read_public_api_test.go
```diff
@@ -0,0 +1,374 @@
...
+	RequireStatus(t, resp, 200)
+
+	attachmentBody := "this is the body of attachment"
+	attachmentContentType := "content/type"
+	reqHeaders := map[string]string{
+		"Content-Type": attachmentContentType,
+	}
+	byteArrayAttachmentBody := []byte(attachmentBody)
+
+	// attach to existing document created above
```
torcolvin, line 159: add a test for the bad user case to make sure we are executing the code here

rest/bytes_read_public_api_test.go
```diff
@@ -0,0 +1,374 @@
...
+	RequireStatus(t, resp, 201)
+
+	// Now call _revs_diff as the user and take the bytes length of the endpoint input
+	input = `{"rd1": ["13-def", "12-xyz"],
+              "rd2": ["34-def"],
+              "rd9": ["1-a", "2-b", "3-c"],
+              "_design/ddoc": ["1-woo"]
+             }`
+	inputBytes := []byte(input)
+	resp = rt.SendUserRequest(http.MethodPost, "/{{.keyspace}}/_revs_diff", input, "alice")
```
torcolvin, line 205: add the same test for a bad user

rest/bytes_read_public_api_test.go
```diff
@@ -0,0 +1,374 @@
...
+	RequireStatus(t, resp, http.StatusCreated)
+	resp = rt.SendAdminRequest("PUT", "/{{.keyspace}}/doc2", `{"channels":["ABC"]}`)
+	RequireStatus(t, resp, http.StatusCreated)
+	resp = rt.SendAdminRequest("PUT", "/{{.keyspace}}/doc1", `{"channels":[]}`)
+	RequireStatus(t, resp, http.StatusCreated)
+
+	// send user request to _all_docs an take the bytes length of input for the endpoint
+	input := `{"keys": ["doc4", "doc1", "doc3", "b0gus"]}`
+	inputBytes := []byte(input)
+	resp = rt.SendUserRequest(http.MethodPost, "/{{.keyspace}}/_all_docs", input, "alice")
```
torcolvin, line 247: add same test for bad user

rest/bytes_read_public_api_test.go
```diff
@@ -0,0 +1,374 @@
...
+func TestBytesReadBulkDocs(t *testing.T) {
+	rt := NewRestTester(t, nil)
+	defer rt.Close()
+
+	rt.CreateUser("alice", []string{"ABC"})
+
+	// call bulk_docs as user. Take bytes of the input string and asser the stat has increased by that length
+	input := `{"docs": [{"_id": "bulk1", "n": 1}, {"_id": "bulk2", "n": 2}, {"_id": "_local/bulk3", "n": 3}]}`
+	inputBytes := []byte(input)
+	resp := rt.SendUserRequest("POST", "/{{.keyspace}}/_bulk_docs", input, "alice")
```
torcolvin, line 277: add same test for bad user

rest/counted_request_reader.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// Read overrides the Read method from io package that will add the number of bytes read to the CountedRequestReader struct for
+// retrieval for the stat on the database
+func (c *CountedRequestReader) Read(buf []byte) (int, error) {
+	numBytesRead, err := c.reader.Read(buf)
+
+	if numBytesRead >= 0 {
+		atomic.AddInt64(&c.numBytes, int64(numBytesRead))
+	}
```
torcolvin: It looks like https://pkg.go.dev/io#Reader can only return 0 or greater.

I think it's safe to _not_ use an atomic here, because I don't think we can call read and report simultaneously, `Read` on a request is a blocking operation.

rest/handler.go
```diff
@@ -275,6 +277,7 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 			h.logRequestLine()
 		}
 		h.logRESTCount()
+		//h.logBytesRead()
```
```suggestion
```

rest/handler.go
```diff
@@ -129,7 +128,7 @@ func makeHandler(server *ServerContext, privs handlerPrivs, accessPermissions []
 		h.writeError(err)
 		h.response.reportStats(true)
 		h.logDuration(true)
-
+		h.logBytesRead()
```
torcolvin: Can we create a function that does:

```
func (h *handler) reportDbStats() {
    	if !h.shouldUpdateBytesTransferredStats() {
		return
	}
        h.response.reportStats(true)
	// atomically load the number of bytes read on the request
	stat := h.requestBody.LoadCount()
	// as this is int64 lets protect against a situation where a negative is returned from LoadCount() function and thus decrementing the stat
	if stat >= 0 {
		h.db.DbStats.DatabaseStats.PublicRestBytesRead.Add(stat)
	}
}
```

if we bundle them together I think we are less likely to forget them if we have to implement another handler.

rest/handler.go
```diff
@@ -144,6 +143,7 @@ func makeOfflineHandler(server *ServerContext, privs handlerPrivs, accessPermiss
 		h.writeError(err)
 		h.response.reportStats(true)
 		h.logDuration(true)
+		h.logBytesRead()
```
torcolvin: Is there a test for this? I think the only public handler that would hit this is 

```
	root.Handle("/{db:"+dbRegex+"}/", makeOfflineHandler(sc, privs, []Permission{PermDevOps}, nil, (*handler).handleGetDB)).Methods("GET", "HEAD")
```

which you wouldn't ordinarily pass data to a GET for this but I think you could pass a body because I bet we don't check for this.

This is to future proof this API in case we add non admin API with this function

rest/handler.go
```diff
@@ -158,6 +158,7 @@ func makeMetadataDBOfflineHandler(server *ServerContext, privs handlerPrivs, acc
 		err := h.invoke(method, accessPermissions, responsePermissions)
 		h.writeError(err)
 		h.response.reportStats(true)
+		h.logBytesRead()
```
torcolvin: Just as a note, this won't get hit because the only caller of this function is on the admin port, but I still think it's a good idea for future proofing.

rest/handler.go
```diff
@@ -405,6 +412,11 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 	if h.privs != adminPrivs {
 		var err error
 		if err = h.checkAuth(dbContext); err != nil {
+			// ContentLength on the request can be -1 if the length is 'unknown'. Add check to protect against decrementing the stat
+			// In the eventuality auth fails on the request we need to still increment stat for the size of the body in bytes
+			if h.rq.ContentLength >= 0 {
+				dbContext.DbStats.DatabaseStats.PublicRestBytesRead.Add(h.rq.ContentLength)
+			}
```
torcolvin: I think this is not right, this should be hit by the handling in `make*Handler` functions - and `ContentLength` is not guaranteed to be populated.

torcolvin (commented):
Lots of nit comments but only because this code is confusing - mostly add tests to make sure we don't regress and I think in this case we don't need an atomic, similar to the `ResponseWriter`.

rest/handler.go
gregns1: Yeah i hadn't yet moved this

rest/bytes_read_public_api_test.go
gregns1: This is done inside TestBytesReadAuthFailed

rest/handler.go
gregns1: We don't read the body on a few GET requests. If this is needed this is further enhancemnet to this PR.

torcolvin: Since this PR is meant to track ingress, we'll want to read the body. I think the code you wrote will do this.

gregns1: The code I've written only tracks whats read. If there is no explicit read from the request body nothing is counted. 
See here as an example -> https://github.com/couchbase/sync_gateway/blob/e2a3a58ec3b0c08b6b5f519c034e3a4dcedb922e/rest/changes_api.go#L167

We don't read in case of GET request (don't typically need to and will save effort involved doing so). So If i was to pass a body to GET changes as its not read nothing is counted.

rest/counted_request_reader.go
```diff
@@ -0,0 +1,42 @@
...
+	numBytesRead, err := c.reader.Read(buf)
+
+	c.numBytes += int64(numBytesRead)
+	return numBytesRead, err
+}
+
+// LoadCount will load the number of bytes read for the request from the CountedRequestReader struct
+func (c *CountedRequestReader) LoadCount() int64 {
+	return c.numBytes
+}
```
torcolvin: I'd probably call this GetBodyBytes() and read this if we are going to need to read this - this avoids having to read bytes in this function later:

```
			// if auth fails we still need to record bytes read over the rest api for the stat, to do this we need to read
			// the body to populate the bytes count on the CountedRequestReader struct
			_, _ = h.readBody()
			dbContext.DbStats.DatabaseStats.PublicRestBytesRead.Add(h.requestBody.LoadCount())
```

I don't think there's another place we need to read it but this moves code out a really complicated function, and in the case of this call where it does hit an invalid case, it's actually going to be checking the `numBytes` twice anyway, so we might as well do a Just In Time read if we need.

Then we read the bytes from a message if we need to (public db scoped) but not in the other cases that I mentioned.

rest/counted_request_reader.go
```diff
@@ -0,0 +1,49 @@
...
+		// it's possible for reader to be nil (nil bodies can be supplied) so nil check needed to be panic safe
+		if c.reader != nil {
+			_, _ = io.ReadAll(c)
+		}
+	}
+	return c.numBytes
+}
+
+// Close to satisfy the interface
+func (c *CountedRequestReader) Close() error { return nil }
```
bbrks: Maybe this Close method can use the [interface smuggling](https://utcc.utoronto.ca/~cks/space/blog/programming/GoInterfaceSmuggling) pattern, to allow `reader` to be closed opportunistically.

Right now we lose the ability to actually close the underlying reader without this or changing `io.Reader` to an explicit `io.ReadCloser` (which may also be acceptable for our use)

```suggestion
// Close to satisfy the io.ReadCloser interface
func (c *CountedRequestReader) Close() error {
	if rc, ok := c.reader.(io.ReadCloser); ok {
		return rc.Close()
	}
	return nil
}
```

rest/handler.go
```diff
@@ -405,6 +407,10 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 	if h.privs != adminPrivs {
 		var err error
 		if err = h.checkAuth(dbContext); err != nil {
+			// if auth fails we still need to record bytes read over the rest api for the stat, to do this we need to read
+			// the body to populate the bytes count on the CountedRequestReader struct
+			_, _ = h.readBody()
```
bbrks: Don't think `h.readBody()` is required after the last commit/update - as `GetBodyBytesCount()` will read if required.

rest/handler.go
```diff
@@ -290,14 +288,18 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 		}
 	}()
 
+	h.requestBody = NewReaderCounter(h.rq.Body)
+
```
bbrks: Is this required for the `default`/ error handling case below? If so, I think we need to consider removing the `case "":` one (which makes me think switch/case is wrong and we should just do a single `if` for gzip handling)

rest/handler.go
```diff
@@ -473,7 +479,7 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 			}
 			// The above readBody() will end up clearing the body which the later handler will require. Re-populate this
 			// for the later handler.
-			h.requestBody = io.NopCloser(bytes.NewReader(body))
+			h.requestBody.reader = io.NopCloser(bytes.NewReader(body))
```
bbrks, line 486: I suspect this will cause double counting of the request body, as we're basically repopulating the counter's reader to be consumed again. Could do with writing a test (likely an integration test for RBAC) that hits this codepath to verify.

Seems like we need to replace the `CountedRequestReader` with a non-counting `io.Reader` (which has implications for the type of `h.requestBody`.

rest/handler.go
```diff
@@ -290,14 +288,18 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
...
+
 	switch h.rq.Header.Get("Content-Encoding") {
 	case "":
-		h.requestBody = h.rq.Body
+		h.requestBody = NewReaderCounter(h.rq.Body)
 	case "gzip":
+		var gzipRequestReader *gzip.Reader
 		var err error
-		if h.requestBody, err = gzip.NewReader(h.rq.Body); err != nil {
+		if gzipRequestReader, err = gzip.NewReader(h.rq.Body); err != nil {
```
bbrks: Nit: Could avoid the `var` declarations if we drop the error handling out of the assignments.

```suggestion
		gzipRequestReader, err := gzip.NewReader(h.rq.Body)
		if err != nil {
```

rest/bytes_read_public_api_test.go
```diff
@@ -0,0 +1,475 @@
...
+	// {{.keyspace}} isn't supported so use default collection
+	rq, err := http.NewRequest("PUT", "/db/doc", &buf)
+	require.NoError(t, err)
+	rq.Header.Set("Content-Encoding", "gzip")
+	resp := rt.Send(rq)
+	RequireStatus(t, resp, http.StatusCreated)
+
+	_, ok := base.WaitForStat(func() int64 {
+		return rt.GetDatabase().DbStats.DatabaseStats.PublicRestBytesRead.Value()
+	}, int64(len(inputBytes)))
```
bbrks, line 443: We're counting the decompressed request body here - which seems wrong in context of calculating actual bandwidth used.

Might (hopefully) just be a case of re-ordering the io.Reader wrappers:
E.g. `counter(gzip(req.Body)) = counts uncompressed` vs. `gzip(counter(req.Body)) = counts compressed`

rest/handler.go
```diff
@@ -654,6 +660,21 @@ func (h *handler) logRESTCount() {
...
+func (h *handler) reportDbStats() {
+	if !h.shouldUpdateBytesTransferredStats() {
+		return
+	}
+	var bytesReadStat int64
+	h.response.reportStats(true)
+	// load the number of bytes read on the request
+	bytesReadStat = h.requestBody.GetBodyBytesCount()
+	// as this is int64 lets protect against a situation where a negative is returned from GetBodyBytesCount() function and thus decrementing the stat
+	if bytesReadStat >= 0 {
```
bbrks: Nit: Avoids `.Add(0)` call on the noop case, which is probably going to be common for GET requests and work we'd like to avoid doing (atomic operations are cheaper than mutexes, but not free - performing them has CPU cache implications)

```suggestion
	if bytesReadStat > 0 {
```

bbrks (changes requested):
General approach looks fine. See inline comments for specific suggestions.

rest/handler.go
gregns1: That was a mistake should only have the `NewReaderCounter` assignment to the requestBody on the handler inside the switch case. Will take that out, good spot!

rest/counted_request_reader.go
gregns1: Changed to a ReaderCloser as seesm the original requestBody before my changes was a ReaderCloser type! Updating the close method to reflect that

rest/handler.go
torcolvin: The other way to do this might be to store a boolean on `CountedRequestReader` whether we've read already or not.

But we could also use the `CountedResponseWriter` and `NonCountedRepsonseWriter` pattern here too.

rest/bytes_read_public_api_test.go
gregns1: Going with approximation due to the underlying encoded byte array being private and the gzip utility decompressing before reading.

bbrks (approved):

PR #6341, opened 2023-07-26
CBG-3190: Made new tool to output Prometheus metric definitions to JSON file

IsaacLambat:
CBG-3190

Made a tool to output Prometheus metric definitions to a JSON file.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1944/

Example output:

```
[
	{
		"name": "sgw_resource_utilization_admin_net_bytes_recv",
		"unit": "bytes",
		"help": "The total number of bytes received (since node start-up) on the network interface to which the Sync Gateway api.admin_interface is bound. By default, that is the number of bytes received on 127.0.0.1:4985 since node start-up.",
		"added": "3.0.0",
		"stability": "committed",
		"type": "counter"
	},
	{
		"name": "sgw_resource_utilization_admin_net_bytes_sent",
		"unit": "bytes",
		"help": "The total number of bytes sent (since node start-up) on the network interface to which the Sync Gateway api.admin_interface is bound.By default, that is the number of bytes sent on 127.0.0.1:4985 since node start-up.",
		"added": "3.0.0",
		"stability": "committed",
		"type": "counter"
	},
	{
		"name": "sgw_resource_utilization_error_count",
		"help": "The total number of errors logged.",
		"added": "3.0.0",
		"stability": "committed",
		"type": "counter"
	},
	{
		"name": "sgw_resource_utilization_go_memstats_heapalloc",
		"unit": "bytes",
		"help": "HeapAlloc is bytes of allocated heap objects. \"Allocated\" heap objects include all reachable objects, as well as unreachable objects that the garbage collector has not yet freed. Specifically, HeapAlloc increases as heap objects are allocated and decreases as the heap is swept and unreachable objects are freed. Sweeping occurs incrementally between GC cycles, so these two processes occur simultaneously, and as a result HeapAlloc tends to change smoothly (in contrast with the sawtooth that istypical of stop-the-world garbage collectors).",
		"added": "3.0.0",
		"stability": "committed",
		"type": "gauge"
	},
	{
		"name": "sgw_resource_utilization_go_memstats_heapidle",
		"unit": "bytes",
		"help": "HeapIdle is bytes in idle (unused) spans. Idle spans have no objects in them. These spans could be (and may already have been) returned to the OS, or they can be reused for heap allocations, or they can be reused as stack memory. HeapIdle minus HeapReleased estimates the amount of memory that could be returned to the OS, but is being retained by the runtime so it can grow the heap without requesting more memory from the OS. If this difference is significantly larger than the heap size, it indicates there was a recent transient spike in live heap size.",
		"added": "3.0.0",
		"stability": "committed",
		"type": "gauge"
	},
	{
		"name": "sgw_resource_utilization_go_memstats_heapinuse",
		"unit": "bytes",
		"help": "HeapInuse is bytes in in-use spans. In-use spans have at least one object in them. These spans an only be used for other objects of roughly the same size. HeapInuse minus HeapAlloc estimates the amount of memory that has been dedicated to particular size classes, but is not currently being used. This is an upper bound on fragmentation, but in general this memory can be reused efficiently.",
		"added": "3.0.0",
		"stability": "committed",
		"type": "gauge"
	},
```


base/stats.go
```diff
@@ -702,11 +702,47 @@ type SharedBucketImportStats struct {
...
 type SgwStat struct {
-	statFQN       string
-	statDesc      *prometheus.Desc
-	labelValues   []string
-	statValueType prometheus.ValueType
+	fqn       string
+	help      string
+	labels    map[string]string
+	valueType prometheus.ValueType
+	desc      *prometheus.Desc
```
JRascagneres: Can we rename the changed ones back? Seems unnecessary and adds a load of changes to the PR.

base/stats.go
```diff
@@ -702,11 +702,47 @@ type SharedBucketImportStats struct {
...
+}
+
+func (s SgwStat) ValueTypeString() string {
+	switch s.valueType {
+	case prometheus.CounterValue:
+		return "counter"
+	case prometheus.GaugeValue:
+		return "gauge"
+	}
+	return ""
```
JRascagneres: consts would be nice at the top of the file - capella-esque

base/stats.go
```diff
@@ -725,21 +761,26 @@ type SgwBoolStat struct {
 	Val bool
 }
 
+func (s *SgwBoolStat) FormatString() string {
+	return "bool"
```
JRascagneres: ditto const

base/stats.go
```diff
@@ -725,21 +761,26 @@ type SgwBoolStat struct {
...
 
-	constLabels := make(prometheus.Labels)
+	labels := make(prometheus.Labels)
 	for i, labelKey := range labelKeys {
-		constLabels[labelKey] = labelVals[i]
+		labels[labelKey] = labelVals[i]
 	}
 
-	desc := prometheus.NewDesc(name, description, nil, constLabels)
+	desc := prometheus.NewDesc(name, description, nil, labels)
```
JRascagneres: Again, any need for renames?

base/stats.go
```diff
@@ -763,12 +804,16 @@ func NewIntStat(subsystem string, key string, description string, labelKeys []st
 	return stat, nil
 }
 
+func (s *SgwIntStat) FormatString() string {
+	return "int"
```
JRascagneres: const

base/stats.go
```diff
@@ -798,12 +843,16 @@ func NewFloatStat(subsystem string, key string, description string, labelKeys []
 	return stat, nil
 }
 
+func (s *SgwFloatStat) FormatString() string {
+	return "float"
```
JRascagneres: const

base/stats.go
```diff
@@ -878,12 +927,16 @@ func NewDurStat(subsystem string, key string, description string, labelKeys []st
 	return stat, nil
 }
 
+func (s *SgwDurStat) FormatString() string {
+	return "duration"
```
JRascagneres: const

tools/stats-definition-exporter/README.md
```diff
@@ -0,0 +1,12 @@
+# Sync Gateway Prometheus Stat Definition Exporter
+This tool exports all the Prometheus Sync Gateway stats definitions to a JSON file. The JSON file is put inside of a gzipped tarball archive. The JSON is prettified by using tab indents and new lines.
+
+## Output
+### To File
+By default, this tool will output to a file. The archive file outputted is called `sync_gateway_metrics_metadata.tar.gz` however if the string `@PRODUCT_VERSION@` has been replaced in the `main.go` file, the archive file will be called  `sync_gateway_metrics_metadata_PRODUCT_VERSION.tar.gz`. The JSON file inside the archive is always called `sync_gateway_metrics_metadata.json`. This is outputted in the current working directory.
```
JRascagneres: Did you discuss this with Chris? For server the build team creates the compressed file for us... Additionally do we want the product version appended?

JRascagneres (changes requested):
Few initial comments. Can we put a snippet of the JSON too on the PR for review.

tools/stats-definition-exporter/README.md
IsaacLambat: For this, I was just going off the comment in the ticket: https://issues.couchbase.com/browse/CBG-3190?focusedId=694842&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-694842

JRascagneres: Worth double checking with @malarky. On server I'm 90% sure it was done on the build side.

base/stats.go
IsaacLambat: Nope, other then I thought it looked a bit better. Will change it back.

tools/stats-definition-exporter/main_test.go
```diff
@@ -0,0 +1,63 @@
...
+func TestRegisterStats(t *testing.T) {
+	globalStats, dbStats, err := registerStats()
+
+	assert.NotNil(t, globalStats)
+	assert.NotNil(t, dbStats)
+	assert.NoError(t, err)
+}
+
+func TestFileOutput(t *testing.T) {
+	defer func() {
```
JRascagneres: I think the typical way to do testing on files is tmpfiles?

tools/stats-definition-exporter/README.md
bbrks: My opinion is to keep this tool only ever producing a single uncompressed fixed-name file and let the build team do whatever they want to it (compress it, name it with a version/build number, etc.)

It's more difficult to do this here than gzipping as part of a build step.

tools/stats-definition-exporter/main_test.go
bbrks: Will review this more closely tomorrow

malarky:
Just a quick one as I had a glance at this... We don't have `format` in the metrics reference spec, but we will need `units` for anything that's not a `counter`, thanks!

tools/stats-definition-exporter/README.md
malarky: Yeah, the file should be left uncompressed, and without a version number - let the build team archive it up 👍🏻

tools/stats-definition-exporter/README.md
```diff
@@ -0,0 +1,83 @@
...
+  }
+]
+```
+
+- `name` is the fully qualified name of the stat.
+- `unit` is what unit the stat uses such as bytes or nanoseconds. If the stat has no units, then this will be omitted.
+- `labels` is a list of label keys that Prometheus uses to uniquely distinguish between the same stat being declared multiple times. For example, `databases`, `collections` etc. This is omitted if the stat has no labels.
+- `help` contains a description of what the stat does.
+- `added` is the Sync Gateway version the stat got added.
+- `stability` is what the current stability of the stat is such as `committed`.
```
bbrks: Is this an enum? What are the other options?

base/stats.go
```diff
@@ -61,6 +61,28 @@ const (
...
+
+	StatFormatInt      = "int"
+	StatFormatFloat    = "float"
+	StatFormatDuration = "duration"
+	StatFormatBool     = "bool"
+
+	StatVersionAdded3dot1dot0 = "3.1.0"
+
+	StatStabilityNoStability = ""
+	StatStabilityCommitted   = "committed"
```
bbrks: If this stability value does turn out to be an enum, specify the other options here, even if we don't make use of them.

base/stats.go
```diff
@@ -61,6 +61,28 @@ const (
...
+	StatUnitNanoseconds   = "nanoseconds"
+	StatUnitSeconds       = "seconds"
+	StatUnitUnixTimestamp = "unix timestamp"
+
+	StatFormatInt      = "int"
+	StatFormatFloat    = "float"
+	StatFormatDuration = "duration"
+	StatFormatBool     = "bool"
+
+	StatVersionAdded3dot1dot0 = "3.1.0"
```
bbrks: Are there no `3.2.0` stats on `master`? That surprises me.

base/stats.go
```diff
@@ -1322,167 +1418,167 @@ func (d *DbStats) initDatabaseStats() error {
 	labelKeys := []string{DatabaseLabelKey}
 	labelVals := []string{d.dbName}
 
-	resUtil.ReplicationBytesReceived, err = NewIntStat(SubsystemDatabaseKey, "replication_bytes_received", ReplicationBytesReceivedDesc, labelKeys, labelVals, prometheus.CounterValue, 0)
+	resUtil.ReplicationBytesReceived, err = NewIntStat(SubsystemDatabaseKey, "replication_bytes_received", StatUnitBytes, ReplicationBytesReceivedDesc, StatVersionAdded3dot1dot0, StatStabilityNoStability, labelKeys, labelVals, prometheus.CounterValue, 0)
```
bbrks: What makes this (and the other) stats have no stability, vs. committed?

Should no stability have a meaningful name, rather than just empty? (see prior `enum` comments)

tools/stats-definition-exporter/README.md
```diff
@@ -0,0 +1,83 @@
...
+```json
+[
+  {
+    "name": "string",
+    "unit": "string",
+    "labels": ["strings"],
+    "help": "string",
+    "added": "string",
+    "stability": "string",
+    "type": "string"
```
bbrks, line 28: There's no mention of `deprecated`, which I'd have expected so we're able to communicate upcoming stat changes/removals as per the doc.

Is the linked doc/spec outdated in this area?

base/stats.go
```diff
@@ -725,7 +801,11 @@ type SgwBoolStat struct {
 	Val bool
 }
 
-func newSGWStat(subsystem string, key string, description string, labelKeys []string, labelVals []string, statValueType prometheus.ValueType) *SgwStat {
+func (s *SgwBoolStat) FormatString() string {
+	return StatFormatBool
+}
+
+func newSGWStat(subsystem, key, unit, description, versionAdded, stability string, labelKeys, labelVals []string, statValueType prometheus.ValueType) *SgwStat {
```
bbrks: Should we check/enforce that `versionAdded` and `stability` are always present?

Either in here or in the callers or a shared function (I see description checking is already duplicated, so it might make sense to move all of the parameter validation into the same place)

tools/stats-definition-exporter/main.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	if *outputConsoleOnlyFlag {
+		err = statsToConsole(logger, stats)
+	} else {
+		err = statsToFile(logger, stats, *outputFileFlag)
+	}
+	if err != nil {
+		logger.Fatalf("could not write stats: %v", err)
+	}
```
bbrks: A lot of this code (and the associated sets of two functions for console/file) seem unnessesary, as the console can be considered the same as a file if your functions take an `io.Writer`

tools/stats-definition-exporter/file_writer.go
```diff
@@ -0,0 +1,38 @@
...
+	err = os.WriteFile(fullFilePath, statsJSON, 0644)
+	if err != nil {
+		return fmt.Errorf("could not write json to file: %w", err)
+	}
+
+	return nil
+}
+
+func getJSONBytes(stats []StatDefinition) ([]byte, error) {
+	statsJSON, err := json.MarshalIndent(stats, "", "\t")
```
bbrks: You can Marshal JSON directly into an `io.Writer` using the JSON Encoder APIs to save you getting all stats into a `[]byte` in memory first

base/stats.go
```diff
@@ -1322,167 +1418,167 @@ func (d *DbStats) initDatabaseStats() error {
...
 	if err != nil {
 		return err
 	}
-	resUtil.ImportProcessCompute, err = NewIntStat(SubsystemDatabaseKey, "import_process_compute", ImportProcessComputeDesc, labelKeys, labelVals, prometheus.CounterValue, 0)
+	resUtil.ImportProcessCompute, err = NewIntStat(SubsystemDatabaseKey, "import_process_compute", StatUnitNoUnits, ImportProcessComputeDesc, StatVersionAdded3dot1dot0, StatStabilityNoStability, labelKeys, labelVals, prometheus.CounterValue, 0)
 	if err != nil {
 		return err
 	}
-	resUtil.SyncProcessCompute, err = NewIntStat(SubsystemDatabaseKey, "sync_process_compute", SyncProcessComputeDesc, labelKeys, labelVals, prometheus.CounterValue, 0)
+	resUtil.SyncProcessCompute, err = NewIntStat(SubsystemDatabaseKey, "sync_process_compute", StatUnitNoUnits, SyncProcessComputeDesc, StatVersionAdded3dot1dot0, StatStabilityNoStability, labelKeys, labelVals, prometheus.CounterValue, 0)
```
bbrks: Do we want to differentiate `StatUnitNoUnits` and `TODO`/`Unspecified`?

This stat _technically does_ have a unit (albeit a weird one and not directly relevant to customers: millisecond bytes)

base/stats.go
```diff
@@ -1322,167 +1418,167 @@ func (d *DbStats) initDatabaseStats() error {
...
 	if err != nil {
 		return err
 	}
-	resUtil.TotalSyncTime, err = NewIntStat(SubsystemDatabaseKey, "total_sync_time", TotalSyncTimeDesc, labelKeys, labelVals, prometheus.CounterValue, 0)
+	resUtil.TotalSyncTime, err = NewIntStat(SubsystemDatabaseKey, "total_sync_time", StatUnitSeconds, TotalSyncTimeDesc, StatVersionAdded3dot1dot0, StatStabilityNoStability, labelKeys, labelVals, prometheus.CounterValue, 0)
 	if err != nil {
 		return err
 	}
-	resUtil.ImportProcessCompute, err = NewIntStat(SubsystemDatabaseKey, "import_process_compute", ImportProcessComputeDesc, labelKeys, labelVals, prometheus.CounterValue, 0)
+	resUtil.ImportProcessCompute, err = NewIntStat(SubsystemDatabaseKey, "import_process_compute", StatUnitNoUnits, ImportProcessComputeDesc, StatVersionAdded3dot1dot0, StatStabilityNoStability, labelKeys, labelVals, prometheus.CounterValue, 0)
```
bbrks: Definitely not a 3.1.0 stat - 3.2.0 I think?

bbrks (changes requested):
I've not gone through all of the stats to check units/version/etc. but I did leave comments on a couple.

Who's job is it to verify these?

tools/stats-definition-exporter/README.md
JRascagneres: Yes, talked about it, he's adding them

base/stats.go
JRascagneres: As above

base/stats.go
JRascagneres: Yes, talked about it, he's adding

base/stats.go
JRascagneres: because not committed, this will be changed to one of the added enums

tools/stats-definition-exporter/README.md
JRascagneres: There is a deprecated, this will be added

base/stats.go
JRascagneres: Per spec we want to follow the units given here - https://prometheus.io/docs/practices/naming/#base-units (or very close, eg that says seconds but we can use nanoseconds) but we don't want to force a bunch of units in where it doesn't really fit. Talked with Malarky about this

bbrks (dismissed):

malarky:
Woo, looks good, thanks all!

Can't really do a full review of all the messages, but from a quick scan through they look decent (and we can iterate with little changes to the text if we need to).

Two things though...

A few messages have some weird quote/backtick/apostrophes going on, for example in `sgw_cache_chan_cache_channels_evicted_nru`:

```
"The total number of active channel cache channels evicted, based on ‘not recently used’ criteria."
```

Those seem to be a pair of U+2018 and U+2019, where we probably just want a pair of regular single quotes (and not sure if they might need escaping, like we do with the double-quotes, or if JSON is OK with that).

And some of the `HELP` entries might be a little long - there's no firm limit but from our internal spec:

> End-user understandable short description of the metric family. Generally a single sentence is appropriate, but may run to two or three sentences if really required. If more information on the metric is needed, use the notes attribute.

There are a handful that wrap to three lines on my laptop screen, which feels excessive, e.g. `sgw_resource_utilization_go_memstats_heapalloc` and friends.

Using `sgw_resource_utilization_go_memstats_heapalloc` as an example, maybe something like this instead:

```
"help": "HeapAlloc is bytes of allocated heap objects. \"Allocated\" heap objects include all reachable objects, as well as unreachable objects that the garbage collector has not yet freed.",
"notes": "Specifically, HeapAlloc increases as heap objects are allocated and decreases as the heap is swept and unreachable objects are freed. Sweeping occurs incrementally between GC cycles, so these two processes occur simultaneously, and as a result HeapAlloc tends to change smoothly (in contrast with the sawtooth that istypical of stop-the-world garbage collectors).",
```

JRascagneres (approved):

PR #6345, opened 2023-07-31
CBG-3206: In-memory struct and API for HLV

gregns1:
CBG-3206

Basic in memory struct for HLV and the basic operations on that struct. 
Note to reviewer logic on detecting conflict between two HLV's is pulled from this section of this doc -> https://docs.google.com/document/d/1KTGXrRnxWMaRI6jwI0-0hoSmBHe3S6IEL3QvprhN_Lk/edit#heading=h.x50fvn11ltma 
Includes tests on operations defined on the HLV but also various conflict detection scenarios including false conflict scenarios. Essentially if a HLV is 'dominating' they are said not to be in conflict, so test coverage is included. 

Have included a helper method for tests to create a HLV for test.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1945/

### DIFF
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -1,0 +1,97 @@
+// Copyright 2023-Present Couchbase, Inc.
+//
+// Use of this software is governed by the Business Source License included
+// in the file licenses/BSL-Couchbase.txt.  As of the Change Date specified
+// in that file, in accordance with the Business Source License, use of this
+// software will be governed by the Apache License, Version 2.0, included in
+// the file licenses/APL2.txt.
+
+package db
+
+type HybridLogicalVector struct {
+	CurrentVersionCAS uint64            // current version cas (or cvCAS) stores the current CAS at the time of replication
+	SourceID          string            // source bucket uuid of where this entry originated from
+	Version           uint64            // current cas of the current version on the version vector
+	MergeVersions     map[string]uint64 // map of merge versions for fast efficient lookup
+	PreviousVersions  map[string]uint64 // map of previous versions for fast efficient lookup
+}
+
+// CurrentVersionVector is a structure used to add a new sourceID:CAS entry to a HLV
+type CurrentVersionVector struct {
+	VersionCAS uint64
+	SourceID   string
+}
+
+// GetCurrentVersion return the current version vector from the HLV in memory
+func (hlv *HybridLogicalVector) GetCurrentVersion() (string, uint64) {
+	return hlv.SourceID, hlv.Version
+}
+
+// IsInConflict tests to see if in memory HLV is conflicting with another HLV
+func (hlv *HybridLogicalVector) IsInConflict(otherVector HybridLogicalVector) bool {
+	// test if in memory version vector is dominating or if they are equal. If so no conflict is present
+	if hlv.isDominating(otherVector) || hlv.isEqual(otherVector) {
+		return false
+	}
+	// if the version vectors aren't dominating or equal then conflict is present
+	return true
+}
+
+// AddVersion adds a version vector to the in memory representation of a HLV and moves current version vector to
+// previous versions on the HLV
+func (hlv *HybridLogicalVector) AddVersion(version CurrentVersionVector) {
+	hlv.PreviousVersions[hlv.SourceID] = hlv.Version
+	hlv.Version = version.VersionCAS
+	hlv.SourceID = version.SourceID
+}
+
+// Remove removes a vector from previous versions section of in memory HLV
+func (hlv *HybridLogicalVector) Remove(source string) {
+	delete(hlv.PreviousVersions, source)
+}
+
+// isDominating tests if in memory HLV is dominating over another
+func (hlv *HybridLogicalVector) isDominating(otherVector HybridLogicalVector) bool {
+	// in memory hlv is dominating if source ID matches other sourceID but with in memory CAS greater than other vector
+	if hlv.SourceID == otherVector.SourceID && hlv.Version > otherVector.Version {
+		return true
+	}
+	// if im memory CAS is greater than other and other sourceID:CAS pair is inside previous versions or
+	// merge versions then in memory HLV is dominating
+	if hlv.Version > otherVector.Version {
+		if hlv.PreviousVersions[otherVector.SourceID] == otherVector.Version {
+			return true
+		}
+		if hlv.MergeVersions[otherVector.SourceID] == otherVector.Version {
+			return true
+		}
+	}
+	return false
+}
+
+// isEqual tests if in memory HLV is equal to another
+func (hlv *HybridLogicalVector) isEqual(otherVector HybridLogicalVector) bool {
+	// if in memory hlv sourceID the sae as other sourceID and in memory CAS is equal to other CAS then HLV's are equal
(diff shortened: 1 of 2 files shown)


db/hybrid_logical_vector.go
```diff
@@ -0,0 +1,97 @@
...
+		return false
+	}
+	// if the version vectors aren't dominating or equal then conflict is present
+	return true
+}
+
+// AddVersion adds a version vector to the in memory representation of a HLV and moves current version vector to
+// previous versions on the HLV
+func (hlv *HybridLogicalVector) AddVersion(version CurrentVersionVector) {
+	hlv.PreviousVersions[hlv.SourceID] = hlv.Version
```
adamcfraser, line 66: I was assuming that the sourceID for the current version wouldn't have an entry in PreviousVersions - so for this operation, we'd need a check whether the incoming version.SourceID matches the existing hlv.SourceID.  Do you agree?

adamcfraser: In addition, I think we want error checking to make sure that the incoming version isn't lower than an existing version for the SourceID.

db/hybrid_logical_vector.go
```diff
@@ -0,0 +1,97 @@
...
+	hlv.Version = version.VersionCAS
+	hlv.SourceID = version.SourceID
+}
+
+// Remove removes a vector from previous versions section of in memory HLV
+func (hlv *HybridLogicalVector) Remove(source string) {
```
adamcfraser: I feel like we may want to return an indication if the source isn't found.  Could be as ErrNotFound, or as a bool.

db/hybrid_logical_vector.go
```diff
@@ -0,0 +1,97 @@
...
+	VersionCAS uint64
+	SourceID   string
+}
+
+// GetCurrentVersion return the current version vector from the HLV in memory
+func (hlv *HybridLogicalVector) GetCurrentVersion() (string, uint64) {
+	return hlv.SourceID, hlv.Version
+}
+
+// IsInConflict tests to see if in memory HLV is conflicting with another HLV
```
adamcfraser, line 44: HLV A is in conflict with HLV B if:
- A's current version dominates B's version for A's current source, OR
- B's current version dominates A's version for B's current source 

I think this is a simpler evaluation than what you're currently doing in isDominating (although there's some additional work to look up the source version across cv/pv/mv).

db/hybrid_logical_vector.go
```diff
@@ -0,0 +1,97 @@
...
+		if hlv.equalMergeVectors(otherVector) {
+			return true
+		}
+	}
+	// they aren't equal
+	return false
+}
+
+// equalMergeVectors tests if two merge vectors between HLV's are equal or not
+func (hlv *HybridLogicalVector) equalMergeVectors(otherVector HybridLogicalVector) bool {
```
adamcfraser, line 121: This doesn't check for entries in otherVector that aren't present in hlv - needs at least an "is length equal" check at the start to handle this.

db/hybrid_logical_vector.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+		if hlv.MergeVersions[otherVector.SourceID] == otherVector.Version {
+			return true
+		}
+	}
+	return false
+}
+
+// isEqual tests if in memory HLV is equal to another
+func (hlv *HybridLogicalVector) isEqual(otherVector HybridLogicalVector) bool {
```
adamcfraser, line 99: We may not need isEqual yet, but if we want to implement it should also be comparing previousVersions.

db/hybrid_logical_vector_test.go
```diff
@@ -0,0 +1,187 @@
...
+	assert.True(t, reflect.DeepEqual(hlv.PreviousVersions, pv))
+}
+
+// TestConflictDetectionDominating:
+//   - Tests two cases where one HLV's is said to be 'dominating' over another and thus not in conflict
+//   - Test case 1: where sourceID is the same between HLV's but the in memory representation has higher version CAS
+//   - Test case 2: where sourceID is different and the in memory HLV has higher CAS than other HLV but the other HLV's
+//     sourceID:CAS pair is present in the in memory HLV's previous versions
+//   - Assert that both scenarios returns false from IsInConflict method, as we have a HLV that is dominating in each case
+func TestConflictDetectionDominating(t *testing.T) {
```
adamcfraser, line 84: I think tests for conflicts should define HLVs with multiple versions, instead of just a single version per HLV.  (the latter is a valid case, but only one of many).  For testing it would be useful to define a human readable shorthand for the HLV that could be parsed into the HLV struct.  You could come up with a proposal, I'm thinking something like:
   [ver@src, ver@src, ver@src, ver@src]
This would assume the 0th entry in the list is current version, and everything else is a previous version.  Adding some type of annotation for merge versions would be needed, maybe m_ver@src?

adamcfraser (changes requested):

db/hybrid_logical_vector.go
gregns1: I agree making those changes now! Thanks for clearing that up!

db/hybrid_logical_vector.go
gregns1: I agree I have added additional error checking and additional check for equal source ID's. But in my head if sourceID different, thus this mutation has come from a different source we need to move the current sourceID version pair into previous versions still right? I have updated the code to reflect this but let me know if I am wrong here.

db/hybrid_logical_vector.go
gregns1: Given we may need it in future I will leave it in for now (unless you think its worth taking out at this stage). Have added the comparison between previoious versions too.

db/hybrid_logical_vector.go
```diff
@@ -22,76 +28,123 @@ type CurrentVersionVector struct {
...
+	// if entry is not found in previous versions we return error
+	if hlv.PreviousVersions[source] == 0 {
+		return base.ErrNotFound
+	}
 	delete(hlv.PreviousVersions, source)
+	return nil
 }
 
 // isDominating tests if in memory HLV is dominating over another
 func (hlv *HybridLogicalVector) isDominating(otherVector HybridLogicalVector) bool {
```
adamcfraser, line 83: I feel like this evaluation would be simpler if you added a method called GetVersion(sourceID) to HybridLogicalVector that would fetch the version for a given source wherever it's located in the HLV (current version, previous versions, merge version). 

With that in place, isDominating is just a single comparison of hlv.currentVersion > otherVector.GetVersion(hlv.SourceID).

adamcfraser: (I'm not 100% sure whether it's valid for the same source to appear in both the mv and cv/pv - just in case, GetVersion can return the largest version found for the source.

adamcfraser (commented):
Looks good, but a suggestion to reduce complexity of isDominating.

db/hybrid_logical_vector.go
gregns1: I agree much easier way to implement and provides a useful method to get the latest version CAS for a given sourceID in a HLV.

adamcfraser (approved):

PR #6346, opened 2023-07-31
CBG-3235: Query based resync infinite loop when querylimit <= number of docs+tombstones

mohammed-madi:
CBG-3235

Describe your PR here...
- Assign highseq before checking if doc is empty on resyncDocument, which was defaulting to 0 before causing
- 		if queryRowCount < queryLimit || highSeq >= endSeq {
- to fail as highseq is 0, leading to an infinite loop

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1946/

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -1761,10 +1761,11 @@ func (db *DatabaseCollectionWithUser) resyncDocument(ctx context.Context, docid,
 			raw []byte, rawXattr []byte, deleteDoc bool, expiry *uint32, err error) {
 			// There's no scenario where a doc should from non-deleted to deleted during UpdateAllDocChannels processing,
 			// so deleteDoc is always returned as false.
+			doc, err := unmarshalDocumentWithXattr(docid, currentValue, currentXattr, currentUserXattr, cas, DocUnmarshalAll)
+			updatedHighSeq = doc.Sequence
 			if currentValue == nil || len(currentValue) == 0 {
 				return nil, nil, deleteDoc, nil, base.ErrUpdateCancel
 			}
-			doc, err := unmarshalDocumentWithXattr(docid, currentValue, currentXattr, currentUserXattr, cas, DocUnmarshalAll)
 			if err != nil {
 				return nil, nil, deleteDoc, nil, err
 			}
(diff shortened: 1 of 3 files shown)


db/database.go
```diff
@@ -1761,13 +1761,14 @@ func (db *DatabaseCollectionWithUser) resyncDocument(ctx context.Context, docid,
...
-				return nil, nil, deleteDoc, nil, base.ErrUpdateCancel
-			}
 			doc, err := unmarshalDocumentWithXattr(docid, currentValue, currentXattr, currentUserXattr, cas, DocUnmarshalAll)
 			if err != nil {
 				return nil, nil, deleteDoc, nil, err
 			}
+			updatedHighSeq = doc.Sequence
+			if currentValue == nil || len(currentValue) == 0 {
+				return nil, nil, deleteDoc, nil, base.ErrUpdateCancel
+			}
```
bbrks: This doesn't quite seem like the right place for a fix, but I might be wrong. Why would we need to run `unmarshalDocumentWithXattr` before deciding that based on `currentValue`, we don't need to resync?

Notably: from `resyncDocument` we're now returning a "new sequence" that is actually the previous sequence!

I was thinking the loop fix would be inside _the caller_ of `resyncDocument` to appropriately handle zero-value sequences for the error conditions.

bbrks (changes requested):
I have a question about exactly what this function now returns. I'm not sure it's correct, and the loop fix is just a side-effect.

It seems that we need to actually handle zero-values along with the error, where the caller invokes `resyncDocument`.

db/database.go
mohammed-madi: https://github.com/couchbase/sync_gateway/pull/6346/commits/825508a9d99560ff2f431627d664ab234b473532 
Implemented the fix discussed in sync up yesterday, doc.sequence is now unmarhsalled from the QueryResync rows, with a check for views since the rows are different

bbrks (approved):

PR #6348, opened 2023-07-31
CBG-3286 Per-db console log settings

bbrks:
CBG-0000

Stores a reference to the database log configuration inside the database-scoped log context, so that we're able to look it up when we log to determine whether to emit the console log or not.

This allows databases to define their own console log level and log keys. Neither of the two settings takes precedence. They're additive (either node OR db log level/key are checked when writing a log line, if either match, it will be logged).

### DIFF
--- a/base/logger_external.go
+++ b/base/logger_external.go
@@ -46,7 +46,7 @@ func initExternalLoggers() {
 }
 
 func updateExternalLoggers() {
-	if consoleLogger != nil && consoleLogger.shouldLog(LevelDebug, KeyWalrus) {
+	if consoleLogger != nil && consoleLogger.shouldLog(context.Background(), LevelDebug, KeyWalrus) {
 		rosmar.SetLogLevel(rosmar.LevelTrace)
 	} else {
 		rosmar.SetLogLevel(rosmar.LevelInfo)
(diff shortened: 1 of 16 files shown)


JRascagneres (dismissed):

base/logger_external.go
```diff
@@ -46,7 +46,7 @@ func initExternalLoggers() {
 }
 
 func updateExternalLoggers() {
-	if consoleLogger != nil && consoleLogger.shouldLog(LevelDebug, KeyWalrus) {
+	if consoleLogger != nil && consoleLogger.shouldLog(context.Background(), LevelDebug, KeyWalrus) {
```
adamcfraser: Does it make sense to pass context.Background() to shouldLog when we don't have an actual context?   shouldLogDatabaseContext looks like it exits earliest if the context is nil - would it be preferable to leave as nil?

adamcfraser (commented):
Approach seems reasonable - added a question about whether background context makes sense in some cases.

IsaacLambat (changes requested):
The Rest API documentation needs updating (`docs/api/components/schemas.yaml L1129`)

IsaacLambat (approved):

PR #6349, opened 2023-07-31
CBG-3247 turn channel send into blocking call

torcolvin:
This is a speculative fix, but definitely simplifies the code. Ran {{base}} DCP tests against race detector.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1951/

### DIFF
--- a/base/dcp_client.go
+++ b/base/dcp_client.go
@@ -504,13 +504,7 @@ func (dc *DCPClient) openStreamRequest(vbID uint16) error {
 		if err == nil {
 			err = dc.verifyFailoverLog(vbID, f)
 			if err == nil {
-				e := streamOpenEvent{
-					streamEventCommon: streamEventCommon{
-						vbID: vbID,
-					},
-					failoverLogs: f,
-				}
-				dc.workerForVbno(vbID).Send(e)
+				dc.metadata.SetFailoverEntries(vbID, f)
 			}
 		}
 		openStreamError <- err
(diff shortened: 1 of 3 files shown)


base/dcp_client.go
```diff
@@ -504,13 +504,7 @@ func (dc *DCPClient) openStreamRequest(vbID uint16) error {
...
 			err = dc.verifyFailoverLog(vbID, f)
 			if err == nil {
-				e := streamOpenEvent{
-					streamEventCommon: streamEventCommon{
-						vbID: vbID,
-					},
-					failoverLogs: f,
-				}
-				dc.workerForVbno(vbID).Send(e)
+				dc.metadata.SetFailoverEntries(vbID, f)
```
adamcfraser, line 507: I don't feel 100% confident that the onStreamEnd usage of openStream (which calls openStreamRequest) doesn't have a scenario that would result in SetFailoverEntries could be called for a vbID while a worker is trying to process an event for the same entry while working the event queue.  

It doesn't look to me like we have any reads of the properties that are modified here(VbUUID and FailoverEntries) during worker processing, so it might be fine.  I'm still wondering whether we want a RW mutex specifically for those two properties for extra safety - what do you think?

adamcfraser (changes requested):

base/dcp_client.go
torcolvin: I feel pretty confident about this, because we call `GetMeta` right before, and this can only occur when the stream has ended, and only one `openStream` call can happen at once. I worry more about adding any additional locking here.

adamcfraser: I agree that this change isn't introducing any new races - thanks.

adamcfraser (approved):

PR #6358, opened 2023-08-03
CBG-3207: Implementation of persisted form of HLV.

gregns1:
CBG-3207

Converts the in memory representation of HLV in #6345 to correct format for persistence. 

Defined here -> https://docs.google.com/document/d/1S0WStgbqRGlmwbdNyHn_9qqEqvshPv9WyG8skauJDEc/edit#heading=h.x82j5i9fdonk

## Pre-review checklist

- [x] Link upstream PRs -> https://github.com/couchbase/sync_gateway/pull/6345

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1956/

### DIFF
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -9,7 +9,6 @@
 package db
 
 import (
-	"encoding/json"
 	"fmt"
 
 	"github.com/couchbase/sync_gateway/base"
@@ -30,11 +29,11 @@ type CurrentVersionVector struct {
 }
 
 type PersistedHybridLogicalVector struct {
-	CurrentVersionCAS string            `json:"cvCas"` // current version cas (or cvCAS) stores the current CAS at the time of replication
-	SourceID          string            `json:"src"`   // source bucket uuid of where this entry originated from
-	Version           string            `json:"vrs"`   // current cas of the current version on the version vector
-	MergeVersions     map[string]string `json:"mv"`    // map of merge versions for fast efficient lookup
-	PreviousVersions  map[string]string `json:"pv"`
+	CurrentVersionCAS string            `json:"cvCas,omitempty"` // current version cas (or cvCAS) stores the current CAS at the time of replication
+	SourceID          string            `json:"src,omitempty"`   // source bucket uuid of where this entry originated from
+	Version           string            `json:"vrs,omitempty"`   // current cas of the current version on the version vector
+	MergeVersions     map[string]string `json:"mv,omitempty"`    // map of merge versions for fast efficient lookup
+	PreviousVersions  map[string]string `json:"pv,omitempty"`
 }
 
 type PersistedVersionVector struct {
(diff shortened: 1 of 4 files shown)


db/hybrid_logical_vector.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+}
+
+func (hlv *HybridLogicalVector) MarshalJSON() ([]byte, error) {
+	persistedHLV := PersistedVersionVector{}
+	cvCasByteArray := base.Uint64CASToLittleEndianHex(hlv.CurrentVersionCAS)
+	var persistedSRC string
+	// check if it is a server mutation, if not we need to convert to base64 encoding and add the mobile mutation prefix
+	if hlv.SourceID[0:2] != "s_" {
+		encodedValue, err := base.HexToBase64(hlv.SourceID)
```
adamcfraser: I don't think we want to make any assumption about the format of non-server SourceIDs (i.e. that they are hex values).  Replicating peers *should* be defining a storage-optimized sourceID value, but I think it's going to be that peer's responsibility to do so, and SG shouldn't do any additional formatting.

db/hybrid_logical_vector.go
```diff
@@ -0,0 +1,138 @@
...
+		if err != nil {
+			return nil, err
+		}
+		persistedSRC = string(encodedValue)
+		// as it doesn't have 's_' at start it will be mobile mutation, so we need leading 'm_' to indicate this
+		persistedSRC = "m_" + persistedSRC
```
adamcfraser: I don't think we want/need the 'm_' prefix - as above, replicating peers should be defining their own sourceID, and I don't think the m_ prefix is required.

db/hybrid_logical_vector.go
adamcfraser: More generally, I think it makes sense for the in-memory storage of the sourceIDs to match the format used during persistence, to simplify triaging issues where we're comparing log output to persisted HLVs.

I understand this makes it more work to get from a sourceID back to a bucket UUID, but that seems like the less common comparison of the two.

adamcfraser (changes requested):

bbrks (dismissed):
Looks alright to me. Will leave @adamcfraser to final approval/merge

db/hybrid_logical_vector.go
```diff
@@ -157,3 +169,87 @@ func (hlv *HybridLogicalVector) GetVersion(sourceID string) uint64 {
...
+		return err
+	}
+	// convert the data to in memory format
+	hlv.convertPersistedHLVToInMemoryHLV(persistedJSON)
+	return nil
+}
+
+func (hlv *HybridLogicalVector) convertHLVToPersistedFormat() (*PersistedVersionVector, error) {
+	persistedHLV := PersistedVersionVector{}
+	cvCasByteArray := base.Uint64CASToLittleEndianHex(hlv.CurrentVersionCAS)
```
adamcfraser: Since cvCas is omitempty, do we want to do a zero check here first before converting to a non-empty string?

adamcfraser (changes requested):
Approach looks good - one question about omitempty handling for non-required fields.

adamcfraser (approved):

PR #6366, opened 2023-08-14
CBG-3210: Updating HLV on Put And PutExistingRev

gregns1:
CBG-3210
CBG-3208

Also includes CBG-3208 putting persisted form of hlv to sync data.

Changed to update HLV sourceID, Version and Previous Versions in the HLV when doing a Put of a document. In doing so we need to know the type of ecvent that is causing the Put to take place e.g. import event, sg write etc. So added code to track what type of event and pass that down so the logic can be computed corretly for the HLV update and mutate in options. 

Includes change to unmarshal the HLV in sync data using `unmarshalDocumentWithXattr`

Also includes some changes to existing tests that were failing due to the version vector not being present. 

## Pre-review checklist

## Upstream Dependencies
- [x] https://github.com/couchbase/sg-bucket/pull/107
- [x] https://github.com/couchbaselabs/rosmar/pull/13
- [x] https://github.com/couchbase/sync_gateway/pull/6473

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2085/

### DIFF
--- a/db/attachment_test.go
+++ b/db/attachment_test.go
@@ -807,12 +807,7 @@ func TestMigrateBodyAttachments(t *testing.T) {
   },
   "channels": {
     "ABC": null
-  },
-  "_vv":{
-   		"cvCas":"0x0",
-   		"src":"abc",
-   		"vrs":"0x0"
-	}
+  }
 }`
 
 		var bodyVal map[string]interface{}
@@ -1099,11 +1094,6 @@ func TestMigrateBodyAttachmentsMerge(t *testing.T) {
   "channels": {
     "ABC": null
   },
-  "_vv":{
-   		"cvCas":"0x0",
-   		"src":"abc",
-   		"vrs":"0x0"
-	},
   "attachments": {
     "bye.txt": {
       "digest": "sha1-l+N7VpXGnoxMm8xfvtWPbz2YvDc=",
--- a/db/attachment_test.go
+++ b/db/attachment_test.go
@@ -807,7 +807,12 @@ func TestMigrateBodyAttachments(t *testing.T) {
   },
   "channels": {
     "ABC": null
-  }
+  },
+  "_vv":{
+   		"cvCas":"0x0",
+   		"src":"abc",
+   		"vrs":"0x0"
+	}
 }`
 
 		var bodyVal map[string]interface{}
@@ -1094,6 +1099,11 @@ func TestMigrateBodyAttachmentsMerge(t *testing.T) {
   "channels": {
     "ABC": null
   },
+  "_vv":{
+   		"cvCas":"0x0",
+   		"src":"abc",
+   		"vrs":"0x0"
+	},
   "attachments": {
     "bye.txt": {
       "digest": "sha1-l+N7VpXGnoxMm8xfvtWPbz2YvDc=",
--- a/db/crud.go
+++ b/db/crud.go
@@ -869,6 +869,22 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
 	return nil
 }
 
+func (db *DatabaseCollectionWithUser) hlvLogicTemp(d *Document) (*Document, error) {
+	newVVEntry := CurrentVersionVector{}
+	bucketUUID, err := db.dbCtx.Bucket.UUID()
+	if err != nil {
+		return nil, err
+	}
+	newVVEntry.SourceID = bucketUUID
+	// add cas here
+	//newVVEntry.VersionCAS = 0 // this will become gocb.MutationMacroCAS
+	err = d.VersionVector.AddVersion(newVVEntry)
+	if err != nil {
+		return nil, err
+	}
+	return d, nil
+}
+
 // Updates or creates a document.
 // The new body's BodyRev property must match the current revision's, if any.
 func (db *DatabaseCollectionWithUser) Put(ctx context.Context, docid string, body Body) (newRevID string, doc *Document, err error) {
@@ -1908,6 +1924,11 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 				return
 			}
 			prevCurrentRev = doc.CurrentRev
+
+			doc, err = db.hlvLogicTemp(doc)
+			if err != nil {
+				return
+			}
 
 			// Check whether Sync Data originated in body
 			if currentXattr == nil && doc.Sequence > 0 {
--- a/db/crud.go
+++ b/db/crud.go
@@ -872,6 +872,9 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
 // updateHLV updates the HLV in the sync data appropriately based on what type of document update event we are encountering
 func (db *DatabaseCollectionWithUser) updateHLV(d *Document, docUpdateEvent uint32) (*Document, error) {
 
+	if d.HLV == nil {
+		d.HLV = &HybridLogicalVector{}
+	}
 	switch docUpdateEvent {
 	case BlipWriteEvent:
 		// preserve any other logic on the HLV that has been done by the client, only update to cvCAS will be needed
@@ -1935,13 +1938,6 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 			}
 			prevCurrentRev = doc.CurrentRev
 
-			doc, err = db.updateHLV(doc, docUpdateEvent)
-			if err != nil {
-				return
-			}
-			// update the mutate in options based on the above logic
-			updatedSpec = doc.SyncData.HLV.computeMacroExpansions()
-
 			// Check whether Sync Data originated in body
 			if currentXattr == nil && doc.Sequence > 0 {
 				doc.inlineSyncData = true
@@ -1965,6 +1961,14 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 				err = base.RedactErrorf("WriteUpdateWithXattr() not able to find revision (%v) in history of doc: %+v.  Cannot update doc.", doc.CurrentRev, base.UD(doc))
 				return
 			}
+
+			// update the HLV values
+			doc, err = db.updateHLV(doc, docUpdateEvent)
+			if err != nil {
+				return
+			}
+			// update the mutate in options based on the above logic
+			updatedSpec = doc.SyncData.HLV.computeMacroExpansions()
 
 			deleteDoc = currentRevFromHistory.Deleted
 
--- a/db/crud.go
+++ b/db/crud.go
@@ -26,8 +26,6 @@ import (
 
 const (
 	kMaxRecentSequences = 20 // Maximum number of sequences stored in RecentSequences before pruning is triggered
-	// hlvExpandMacroCASValue causes the field to be populated by CAS value by macro expansion
-	hlvExpandMacroCASValue = math.MaxUint64
 )
 
 // ErrForbidden is returned when the user requests a document without a revision that they do not have access to.
@@ -871,29 +869,17 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
 	return nil
 }
 
-// updateMutateInSpec returns the mutate in spec needed for the document update based off the outcome in updateHLV
-func updateMutateInSpec(hlv HybridLogicalVector) []sgbucket.MacroExpansionSpec {
-	var outputSpec []sgbucket.MacroExpansionSpec
-	if hlv.Version == math.MaxUint64 {
-		spec := sgbucket.NewMacroExpansionSpec(xattrCurrentVersionPath(base.SyncXattrName), sgbucket.MacroCas)
-		outputSpec = append(outputSpec, spec)
-	}
-	if hlv.CurrentVersionCAS == math.MaxUint64 {
-		spec := sgbucket.NewMacroExpansionSpec(xattrCurrentVersionCASPath(base.SyncXattrName), sgbucket.MacroCas)
-		outputSpec = append(outputSpec, spec)
-	}
-	return outputSpec
-}
-
 // updateHLV updates the HLV in the sync data appropriately based on what type of document update event we are encountering
-func (db *DatabaseCollectionWithUser) updateHLV(d *Document, docUpdateEvent DocUpdateEvent) (*Document, error) {
-
-	switch docUpdateEvent.eventType {
+func (db *DatabaseCollectionWithUser) updateHLV(d *Document, docUpdateEvent uint32) (*Document, error) {
+	if d.HLV == nil {
+		d.HLV = &HybridLogicalVector{}
+	}
+	switch docUpdateEvent {
 	case BlipWriteEvent:
 		// preserve any other logic on the HLV that has been done by the client, only update to cvCAS will be needed
 		d.HLV.CurrentVersionCAS = hlvExpandMacroCASValue
 	case ImportEvent:
-		// VV should remain unchanged
+		// work to be done to decide if the VV needs updating here, pending CBG-3503
 	case SGWriteEvent:
 		// add a new entry to the version vector
 		newVVEntry := CurrentVersionVector{}
@@ -903,6 +889,8 @@ func (db *DatabaseCollectionWithUser) updateHLV(d *Document, docUpdateEvent DocU
 		if err != nil {
 			return nil, err
 		}
+		// update the cvCAS on the SGWrite event too
+		d.HLV.CurrentVersionCAS = hlvExpandMacroCASValue
 	}
 	return d, nil
 }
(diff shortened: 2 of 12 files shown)


db/document.go
```diff
@@ -178,6 +180,7 @@ type Document struct {
 	Deleted        bool
 	DocExpiry      uint32
 	RevID          string
+	VersionVector  HybridLogicalVector
```
adamcfraser: This should be stored in SyncData.  These top-level properties (Deleted, DocExpiry, etc) are (I believe) technical debt used for transient transport of values, and we should avoid this approach unless it's completely necessary.

I think I understand the motivation to have this separate from Sync Data (anticipating the time when this is stored in the _vv xattr and we don't want it automatically marshalled to _sync), but we have a lot of places in the code that work with SyncData and need to be able to evaluate the revision/version history.  (e.g. usages of GetDocSyncData).  Given that I think there's value in having SyncData represent the complete document metadata from an in-memory perspective, and abstract the storage.  If you think that adds too much complexity we can discuss, but would like to understand how you'd handle cases where we want to pass around SyncData.

db/document.go
```diff
@@ -64,23 +65,24 @@ type ChannelSetEntry struct {
...
+	Cas               string                       `json:"cas"`                               // String representation of a cas value, populated via macro expansion
+	Crc32c            string                       `json:"value_crc32c"`                      // String representation of crc32c hash of doc body, populated via macro expansion
+	Crc32cUserXattr   string                       `json:"user_xattr_value_crc32c,omitempty"` // String representation of crc32c hash of user xattr
+	TombstonedAt      int64                        `json:"tombstoned_at,omitempty"`           // Time the document was tombstoned.  Used for view compaction
+	Attachments       AttachmentsMeta              `json:"attachments,omitempty"`
+	ChannelSet        []ChannelSetEntry            `json:"channel_set"`
+	ChannelSetHistory []ChannelSetEntry            `json:"channel_set_history"`
+	HLV               PersistedHybridLogicalVector `json:"_vv,omitempty"`
```
adamcfraser: I thought we'd want to store HybridLogicalVector in SyncData, and only use the Persisted form at marshal and unmarshal time.  (similar to how RevTree and revTreeList behave today).

db/hybrid_logical_vector.go
```diff
@@ -67,13 +67,19 @@ func (hlv *HybridLogicalVector) IsInConflict(otherVector HybridLogicalVector) bo
...
 	if newVersion.VersionCAS < hlv.Version {
-		return fmt.Errorf("attempting to add new verison vector entry with a CAS that is less than the current version CAS value")
+		return fmt.Errorf("attempting to add new verison vector entry with a CAS that is less than the current version CAS value. Current cas: %d new cas %d", hlv.Version, newVersion.VersionCAS)
 	}
 	// if new entry has the same source we simple just update the version
 	if newVersion.SourceID == hlv.SourceID {
 		hlv.Version = newVersion.VersionCAS
 		return nil
 	}
+	// check if this is the first time we're adding a source - version pair
```
adamcfraser, line 73: I'd probably put this before the case on line 73, just for readability.

db/crud.go
```diff
@@ -870,6 +870,30 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
 	return nil
 }
 
+func (db *DatabaseCollectionWithUser) computeHLVLogic(d *Document) (*Document, error) {
```
adamcfraser: I think 'updateHLV' is a better name for this.

db/crud.go
```diff
@@ -870,6 +870,30 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
...
+	newVVEntry.SourceID = bucketUUID
+	// add cas here this will be macro expanded
+	// need to keep same value for now or adding new version vector entry will fail with new cas < than old cas
+	newVVEntry.VersionCAS = d.VersionVector.Version // this will be macro expanded, need to keep same value for now
+	err = d.VersionVector.AddVersion(newVVEntry)
+	if err != nil {
+		return nil, err
+	}
+
+	persistedHLV, err := d.VersionVector.convertHLVToPersistedFormat()
```
adamcfraser: As mentioned elsewhere, I think we want to ensure we're only generating the persisted format at persistence time.

db/crud.go
```diff
@@ -870,6 +870,30 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
...
 
+func (db *DatabaseCollectionWithUser) computeHLVLogic(d *Document) (*Document, error) {
+	newVVEntry := CurrentVersionVector{}
+	bucketUUID, err := db.dbCtx.Bucket.UUID()
+	if err != nil {
+		return nil, err
+	}
+	newVVEntry.SourceID = bucketUUID
+	// add cas here this will be macro expanded
+	// need to keep same value for now or adding new version vector entry will fail with new cas < than old cas
```
adamcfraser: This feels error-prone - like it would be easy to introduce a bug where we either fail to update the value later, or accidentally macro expand over a version that been provided by CBL that we should be preserving.  Alternatives might be to use VersionCAS=MaxUint64 as "should macro expand", or add a flag to HybridLogicalVector that has the same meaning.

db/document.go
adamcfraser: See comment below for additional discussion.

db/crud.go
```diff
@@ -2589,11 +2618,14 @@ func (db *DatabaseCollectionWithUser) CheckProposedRev(ctx context.Context, doci
 const (
 	xattrMacroCas         = "cas"
 	xattrMacroValueCrc32c = "value_crc32c"
+	versionVectorMacro    = "._vv."
+	xattrMacroVersion     = "vrs"
```
adamcfraser: Assuming this is just a temporary path (until we move to _vv) - otherwise I think I'd recommend a single constant for `_vv.vrs`.  Even so I'd remove the `.` delimiters from versionVectorMacro and just insert them when building the path (in xattrVersionPath).

db/crud.go
```diff
@@ -2607,3 +2639,7 @@ func xattrCasPath(xattrKey string) string {
 func xattrCrc32cPath(xattrKey string) string {
 	return xattrKey + "." + xattrMacroValueCrc32c
 }
+
+func xattrVersionPath(xattrKey string) string {
```
adamcfraser: Probably better named as xattrCurrentVersionPath?

adamcfraser (changes requested):

db/crud.go
gregns1: I have changed this temporarily to MaxUint64 before adding the extra logic here. I think i'm a bit confused on the logic i'm missing here. Am I right on saying if we have a CBL mutation, they update the source version pair accordingly, we then need to preserve that right? And instead we macro expand the cvCAS on the HLV? Then the same for ISGR?

adamcfraser: That's all correct.

db/crud.go
```diff
@@ -869,6 +869,46 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
 	return nil
 }
 
+// updateMutateInSpec updates the mutate in options with the appropriate mutate in spec needed based off the outcome in updateHLV
```
bbrks: nit: clarify that we're not modifying something passed in - but returning a slice to be passed into the MutateIn options

```suggestion
// updateMutateInSpec returns the mutate in spec needed for the document update based off the outcome in updateHLV
```

db/crud.go
```diff
@@ -869,6 +869,46 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
...
+	return outputSpec
+}
+
+// updateHLV updates the HLV in the sync data appropriately based on what type of document update event we are encountering
+func (db *DatabaseCollectionWithUser) updateHLV(d *Document, docUpdateEvent DocUpdateEvent) (*Document, error) {
+
+	switch docUpdateEvent.eventType {
+	case BlipWriteEvent:
+		// preserve any other logic on the HLV that has been done by the client, only update to cvCAS will be needed
+		d.HLV.CurrentVersionCAS = math.MaxUint64 // this will be an alias to "should macro expand"
```
bbrks: Might be worth creating a new SG `const` for this value to make it clearer about the intent when used.

e.g.

```go
// hlvExpandMacroValue causes the field to be populated by expanded macro
const hlvExpandMacroValue = math.MaxUint64
```

Naming TBD, as you know the logic better than I do :)

db/crud.go
```diff
@@ -869,6 +869,46 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
...
+	case ImportEvent:
+		// VV should remain unchanged
+	case SGWriteEvent:
+		// add a new entry to the version vector
+		newVVEntry := CurrentVersionVector{}
+		bucketUUID, err := db.dbCtx.Bucket.UUID()
```
bbrks: `Bucket.UUID()` isn't cheap to do for GoCBv2, as it's not fetching a cached value. It's creating a new agent each time to fetch a snapshot then get bucket UUID from that.

Might be worth considering stamping Bucket UUID into `dbCtx`?

db/document.go
```diff
@@ -1192,6 +1193,14 @@ func (doc *Document) UnmarshalWithXattr(ctx context.Context, data []byte, xdata
 			Cas: casOnlyMeta.Cas,
 		}
 		doc._rawBody = data
+	case DocUnmarshalVV:
+		tmpData := SyncData{}
+		unmarshalErr := base.JSONUnmarshal(xdata, &tmpData)
+		if unmarshalErr != nil {
+			return pkgerrors.WithStack(base.RedactErrorf("Failed to UnmarshalWithXattr() doc with id: %s (DocUnmarshalVV).  Error: %v", base.UD(doc.ID), unmarshalErr))
```
bbrks: `pkgerrors` is kind of old now. We want to avoid using it in new code and instead use `%w` to wrap our Go errors with the stdlib functions so we can use `errors.Is`/`errors.As`

bbrks (changes requested):
A few things I've spotted inline

db/attachment_test.go
```diff
@@ -807,7 +807,12 @@ func TestMigrateBodyAttachments(t *testing.T) {
   },
   "channels": {
     "ABC": null
-  }
+  },
+  "_vv":{
+   		"cvCas":"0x0",
+   		"src":"abc",
+   		"vrs":"0x0"
+	}
```
adamcfraser: Is this required here because document.HLV is HybridLogicalVector and not *HybridLogicalVector, and so the omitempty isn't effective?

I realize the storage of _vv in sync is transient until we add _vv xattr support, but I wonder if we should still switch to *HybridLogicalVector to avoid marshalling the struct unless it's actually populated.

db/crud.go
```diff
@@ -869,6 +871,42 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
 	return nil
 }
 
+// updateMutateInSpec returns the mutate in spec needed for the document update based off the outcome in updateHLV
+func updateMutateInSpec(hlv HybridLogicalVector) []sgbucket.MacroExpansionSpec {
+	var outputSpec []sgbucket.MacroExpansionSpec
+	if hlv.Version == math.MaxUint64 {
```
adamcfraser: These should be checking against the new constant (hlvExpandMacroCASValue).

db/crud.go
```diff
@@ -26,6 +26,8 @@ import (
 
 const (
 	kMaxRecentSequences = 20 // Maximum number of sequences stored in RecentSequences before pruning is triggered
+	// hlvExpandMacroCASValue causes the field to be populated by CAS value by macro expansion
+	hlvExpandMacroCASValue = math.MaxUint64
```
adamcfraser: I think it would make sense to store this in hybrid_logical_vector.go.

db/crud.go
```diff
@@ -869,6 +871,42 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
 	return nil
 }
 
+// updateMutateInSpec returns the mutate in spec needed for the document update based off the outcome in updateHLV
+func updateMutateInSpec(hlv HybridLogicalVector) []sgbucket.MacroExpansionSpec {
```
adamcfraser: The name 'updateMutateInSpec' doesn't really make sense for this function with the updated signature.  I think you could use computeMacroExpansions.  It also feels like this could be a method on the HybridLogicalVector, instead of a standalone function.  What do you think?

db/crud.go
```diff
@@ -869,6 +871,42 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
...
+
+// updateHLV updates the HLV in the sync data appropriately based on what type of document update event we are encountering
+func (db *DatabaseCollectionWithUser) updateHLV(d *Document, docUpdateEvent DocUpdateEvent) (*Document, error) {
+
+	switch docUpdateEvent.eventType {
+	case BlipWriteEvent:
+		// preserve any other logic on the HLV that has been done by the client, only update to cvCAS will be needed
+		d.HLV.CurrentVersionCAS = hlvExpandMacroCASValue
+	case ImportEvent:
+		// VV should remain unchanged
```
adamcfraser: There are cases where import needs to update the VV (if the source of the mutation was a non-XDCR SDK write), but it's out of scope of this ticket.  Filed CBG-3503 for this work, can you mention that ticket in the comment?

db/crud.go
```diff
@@ -869,6 +871,42 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
...
+	case SGWriteEvent:
+		// add a new entry to the version vector
+		newVVEntry := CurrentVersionVector{}
+		newVVEntry.SourceID = db.dbCtx.BucketUUID
+		newVVEntry.VersionCAS = hlvExpandMacroCASValue
```
adamcfraser, line 888: I think the current XDCR doc suggests populating cvCAS on an SG write as well.

db/database.go
```diff
@@ -48,6 +48,16 @@ const (
 	DBCompactRunning
 )
 
+const (
+	ImportEvent uint32 = iota
+	SGWriteEvent
+	BlipWriteEvent
+)
+
+type DocUpdateEvent struct {
```
adamcfraser: What else do you anticipate putting in this struct, other than type?

adamcfraser (changes requested):

db/database.go
gregns1: I was probably thinking too far ahead, but was thinking we may need to be able to pass down the stack a flag to to indicate if a doc mutation/creation has arrived to sync gateway from a non version vector aware peer/client? So initially thought this could be a vehicle to add that flag into and pass down the stack to help handle that type of 'doc update event' in sync gateway.

But you raise a good point i'll update the PR to remove the struct and have it as a variable instead!

db/crud.go
gregns1: Good idea, have updated this!

db/crud.go
gregns1: Updated to reflect this, also added assertions to relevant test to reflect this change.

db/attachment_test.go
gregns1: Hit a problem implementing this. I get test failures for rosmar and not server. One test that fails is `TestXattrImportFilterOptIn` and I put this down to an on demand import getting rejected by the import filter. That rejection seems to nil the HLV on sync data and then the write for the doc fails due to mutate in not finding the hlv path in xattrs. I wasn't hitting this before as the HLV on sync data wasn't pointer thus the HLV wasn't getting nil'ed. 

I reverted the change so _vv would still appear in xattr but nothing will be in it due to the content of HybridLogicalVector being omitempty. These tests now seem to still pass without the changes.

gregns1: Have fixed the issue. So now the HLV on sync data is back to being pointer!

db/database.go
```diff
@@ -48,6 +48,14 @@ const (
...
 )
 
+const (
+	ImportEvent uint32 = iota
+	SGWriteEvent
+	BlipWriteEvent
+)
+
+var eventType uint32
+
```
adamcfraser: I think you want to use the standard pattern for an enum type.  In addition, since we've already got the concepts of events in SGW, the name should be a bit more specific.
```suggestion
const (
	ImportEvent docUpdateType = iota
	SGWriteEvent
	BlipWriteEvent
)

type docUpdateType uint32

```

db/crud.go
```diff
@@ -869,6 +869,33 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
 	return nil
 }
 
+// updateHLV updates the HLV in the sync data appropriately based on what type of document update event we are encountering
+func (db *DatabaseCollectionWithUser) updateHLV(d *Document, docUpdateEvent uint32) (*Document, error) {
```
adamcfraser: Once you've updated the enum (see other note), should make this change:
```suggestion
func (db *DatabaseCollectionWithUser) updateHLV(d *Document, docUpdateEvent docUpdateEventType) (*Document, error) {
```

db/crud.go
```diff
@@ -869,6 +869,33 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
...
+
+	if d.HLV == nil {
+		d.HLV = &HybridLogicalVector{}
+	}
+	switch docUpdateEvent {
+	case BlipWriteEvent:
```
adamcfraser: I don't think BlipWriteEvent and SGWriteEvent are the best descriptions for these options.  Today the REST API has the ability to invoke PutExistingRev, and that might be something we end up supporting for versions as well.  Given that I'm thinking it would be preferable to define these based on the functional case.  

I also think these are more like types of updates, as opposed to 'events'.  I'm open to suggestions - maybe something like `docUpdateType` for the type, with `Import`, `ExistingVersion` and `NewVersion`?

db/database.go
adamcfraser: (Made a comment elsewhere about renaming the three types as well)

adamcfraser (changes requested):

adamcfraser (approved):

adamcfraser:
LGTM - @bbrks can you verify your requested changes got made and merge?

bbrks (approved):

PR #6368, opened 2023-08-15
CBG-3203: Don't store channel history for non-leaf revisions

bbrks:
CBG-3203

- Moves channel history storage for non-winning leaf-revisions into a channelMap during marshal/unmarshal.
- Use `_sync.Channels` to build set of activeChannels returned by `IsChannelRemoval`

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2182/


db/revtree.go
```diff
@@ -81,19 +87,37 @@ func (tree RevTree) MarshalJSON() ([]byte, error) {
 				rep.BodyKeyMap[strconv.FormatInt(int64(i), 10)] = info.BodyKey
 			}
 		}
-		rep.Channels[i] = info.Channels
+
+		// for leaf revisions we'll store channel information
+		// there's some duplication here for the current rev
+		// (active channels are stored outside the revtree)
+		// but special casing this doesn't seem worthwhile
```
adamcfraser: Since the majority of documents will be non-branching, I think we should actually special case the 'leaf revision is the current rev' case (and omit ChannelsMap for that scenario), since that means we avoid storing the channel set twice in all documents.

db/revtree.go
```diff
@@ -81,19 +87,37 @@ func (tree RevTree) MarshalJSON() ([]byte, error) {
...
+		if len(info.Channels) > 0 {
```
adamcfraser, line 87: When do we (need to) populate info.Channels for non-leaf nodes?  (i.e. why can't we assume that it's a leaf when this is populated)

adamcfraser: If we can't assume that - we have channels.Branched in syncData.Flags that we can use to quickly check whether the document has any non-tombstoned leaves (and so bypass this work in the most common case).  Any way to leverage that?

db/revtree.go
```diff
@@ -149,29 +173,51 @@ func (tree *RevTree) UnmarshalJSON(inputjson []byte) (err error) {
...
+		leaves := tree.Leaves()
+		for iStr, channels := range rep.ChannelsMap {
+			i, err := strconv.ParseInt(iStr, 10, 64)
+			if err != nil {
+				return err
+			}
 			info := (*tree)[rep.Revs[i]]
-			info.Deleted = true // because tree[rep.Revs[i]].Deleted=true is a compile error
-			(*tree)[rep.Revs[i]] = info
+			if _, isLeaf := leaves[info.ID]; isLeaf {
```
adamcfraser: Do we need the leaf check here, if we're only ever marshalling ChannelsMap for leaves?

adamcfraser (changes requested):
The changes look fine, but the calls to .Leaves() in marshal and unmarshal seem like the have the potential for non-trivial performance head when marshalling/unmarshalling metadata.  Posted some questions about whether it's necessary.

db/revtree.go
bbrks: I think the only way to make this assumption valid would be to make `addRevision` strip parent channel information when you add a new revision to a tree.

I can do that instead, as it avoids having to do the `tree.Leaves()` call

db/revtree.go
bbrks: Done

db/crud.go
```diff
@@ -550,7 +550,11 @@ func (col *DatabaseCollectionWithUser) authorizeDoc(doc *Document, revid string)
 	if revid == "" {
 		revid = doc.CurrentRev
 	}
-	if rev := doc.History[revid]; rev != nil {
+
+	if revid == doc.CurrentRev {
+		ch := doc.currentChannels()
+		return col.user.AuthorizeAnyCollectionChannel(col.ScopeName, col.Name, ch)
```
bbrks: This is sub-optimal - `currentChannels` iterates over `doc.SyncData.Channels` to determine the active set of channels for an active revision.

db/revision_cache_interface.go
```diff
@@ -283,7 +283,11 @@ func revCacheLoaderForDocument(ctx context.Context, backingStore RevisionCacheBa
 		return bodyBytes, body, history, channels, removed, nil, deleted, nil, getHistoryErr
 	}
 	history = encodeRevisions(doc.ID, validatedHistory)
-	channels = doc.History[revid].Channels
+	if doc.CurrentRev == revid {
+		channels = doc.currentChannels()
```
bbrks: Similarly here, iterating over `doc.SyncData.Channels` for the winning rev

bbrks:
@adamcfraser I think this PR is finally good for a fresh review! Thanks

db/revtree.go
```diff
@@ -44,29 +44,32 @@ type RevTree map[string]*RevInfo
...
 	rep := revTreeList{
-		Revs:     make([]string, n),
-		Parents:  make([]int, n),
-		Channels: make([]base.Set, n),
+		Revs:    make([]string, n),
+		Parents: make([]int, n),
 	}
 	revIndexes := map[string]int{"": -1}
 
+	winner, _, _ := tree.winningRevision(context.TODO())
```
adamcfraser: Can we move this check to revtree.UnmarshalJSON when Channels_Old is present, such that we only populate info.Channels for non-winning revisions?  (i.e. it behaves like ChannelMap, and we can assume that info.Channels is only populated for non-winning leaf revisions.)

In addition to standardizing the handling for non-winning revisions, it would have the  benefit of only incurring the cost of winningRevision when Channels_Old is present.

db/crud.go
```diff
@@ -541,21 +541,19 @@ func (db *DatabaseCollectionWithUser) Get1xRevAndChannels(ctx context.Context, d
...
 	}
-	if revid == "" {
-		revid = doc.CurrentRev
-	}
-	if rev := doc.History[revid]; rev != nil {
-		// Authenticate against specific revision:
-		return col.user.AuthorizeAnyCollectionChannel(col.ScopeName, col.Name, rev.Channels)
-	} else {
+
+	channelsForRev, ok := doc.channelsForRev(revid)
```
adamcfraser, line 539: It looks like the effect here is that an authorizeDoc call for a existing, non-leaf rev is going to be treated as a 403 (because channelsForRev will be empty).  Is this a change in behaviour (i.e. it previously would have returned a 404)?

adamcfraser (changes requested):
I like the approach, especially how it minimizes the changes.  Couple of questions on the details.

db/crud.go
bbrks: `channelsForRev` will return `nil, true` if the rev was found in history, even if we didn't find channels for that rev, so I think the behaviour of `authorizeDoc` remains the same?

db/revtree.go
bbrks: Should be done with latest commit

db/crud.go
adamcfraser: I see what you mean.  The original comment on CBG-3203 was "Further analysis is required to identify any spots we're still attempting to reference channel assignment for non-leaf nodes - if those exist, they should be updated to use the channels associated with the active revision."  

What's your opinion on whether we should do what you've got here (return a 403) when attempting to authorize a non-leaf rev, or if we should authenticate against the channels on the active rev?

bbrks: Ah, I see where you were going with this comment now. Because we've stripped the non-leaf channels, there's an implicit change in behaviour from this function, rather than an explicit one. Prior to removal of revtree channels, clients could request an older rev and still get correct access checks from it.

This would affect _all_ of the in-flight old rev requests we handle. I don't think 403 is accurate nor suitable, so this code does need changing.

I think the options here would be to either:
- Use active rev's channels as an optimistic check (assume doc channels haven't changed).
  - Worst case, if channels have been added: Clients can pull a revision they didn't have access to, assuming they knew the rev ID (changes wouldn't have sent them the entry, since it's sequence based rather than rev ID based)
  - If doc channels have been removed: Clients get a 403 for an in-flight rev they would've had access to. Doesn't seem like a big deal, given they just lost doc access anyway.
- Return 404 (are in-flight revs considered best-effort given caching/TTL?)
  - If clients still have doc access they'll get a subsequent changes/rev anyway, and the in-flight rev would've been replaced soon.
  - If clients don't have doc access any more, they won't get an older rev they would've previously.

I'm leaning towards returning 404 for non-leaf revs, as it seems safer (in terms of not sending anything a client shouldn't have), but it does effectively remove the concept of sending in-flight revisions.

bbrks: As discussed, will return 404 in this case (all uncached in-flight rev requests)

May need to differentiate between no channels and unset channels from revcache to do this.

bbrks: Now explicitly return `ErrMissing` in this case for callers.

Test coverage via `TestUncachedOldRevisionChannel`

db/crud.go
```diff
@@ -690,14 +693,14 @@ func (db *DatabaseCollectionWithUser) getAncestorJSON(ctx context.Context, doc *
...
-	if err := db.authorizeDoc(doc, revid); err != nil {
+	if err := db.authorizeDoc(ctx, doc, revid); err != nil {
 		// As a special case, you don't need channel access to see a deletion revision,
 		// otherwise the client's replicator can't process the deletion (since deletions
 		// usually aren't on any channels at all!) But don't show the full body. (See #59)
 		// Update: this applies to non-deletions too, since the client may have lost access to
 		// the channel and gotten a "removed" entry in the _changes feed. It then needs to
 		// incorporate that tombstone and for that it needs to see the _revisions property.
-		if revid == "" || doc.History[revid] == nil {
+		if revid == "" || doc.History[revid] == nil || errors.Is(err, ErrMissing) {
```
adamcfraser, line 694: From the comment on line 700, it seems like we want to bypass the security check for tombstones or removals (so that we successfully set bodyBytes to either EmptyDocument or RemovedRedactedDocument below).  I'm wondering if the ErrMissing handling is going to break this, in particular for removals.  I went looking for a test that covers this path - TestGetRemovedAsUser seems close but I don't think it purges the removal from the rev cache before attempting to retrieve.  I think it would be good to extend TestGetRemovedAsUser for this scenario to see whether it changes before/after this PR:
- removal revision is non-leaf
- removal revision is not resident in the rev cache
- verify removal is still returned

adamcfraser: The same might be true for tombstones, although non-leaf tombstones are more of a corner case.  It would be common for removals to be non-leaf revisions.

adamcfraser (changes requested):

db/crud.go
bbrks: My interpretation of the comment was that an old rev is neither removed from a channel, nor a tombstone. Is that incorrect?

An old revision we don't have channel info for any more seems the same as somebody requesting a pruned revision, or `doc1` / `123-invalid` which would return `ErrMissing`.

adamcfraser: I was thinking of the usage of [IsChannelRemoval in revCacheLoaderForDocument ](https://github.com/couchbase/sync_gateway/blob/master/db/revision_cache_interface.go#L265-L266), where we check the channel history to identify removals for revisions that aren't in the rev tree, and return a removal instead of 404 for those.  I wasn't clear about the interaction (if any) between that handling and the changes here.

bbrks: As discussed, will follow up and get test coverage for a removal that isn't cached and ensure behaviour doesn't change with this PR.

torcolvin:
I'm closing this PR because we are not planning to implement this as a high priority, but I will leave the branch around when we want to actually do this ticket. I noted this in JIRA also.

PR #6377, opened 2023-08-23
CBG-3292: Handling for corrupt db config 

gregns1:
CBG-3292

First I have changed it so that when sync gateway on the config interval looks for database configs in the bucket, will also check that the config bucket name matches the bucket the config is found in. If it doesn't we add the db context to the server context corrupt database map to keep track of. Then the changes will remove the in memory representation of the db config and the db context off of the server context all together. 
This forces operation on the database to fail so I have added some handling for a suitable error message to be returned to the user in this eventuality informing them that they need to update config to correct the bucket name. 
Lastly we needed to be able to correct this corrupt config which is a challenge as at this stage all requests to the db would fail. But I added handling to handle a PUT/POST requests to the db to update the config to correct the corrupt bucket name. 
Test seems to panic against rosmar when trying to update bucket config so added walrus check to avoid it against rosmar. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1982/

### DIFF
--- a/base/constants.go
+++ b/base/constants.go
@@ -230,6 +230,13 @@ func UnitTestUrlIsWalrus() bool {
 	return ServerIsWalrus(UnitTestUrl())
 }
 
+func TestsRequireBootstrapConnection() bool {
+	if !TestUseCouchbaseServer() && UnitTestUrlIsWalrus() {
+		return true
+	}
+	return false
+}
+
 // ServerIsTLS returns true if the server URL is using an accepted secure protocol as it's prefix
 // Prefix checked: couchbases:
 func ServerIsTLS(server string) bool {
--- a/base/constants.go
+++ b/base/constants.go
@@ -14,6 +14,7 @@ import (
 	"errors"
 	"os"
 	"strings"
+	"testing"
 	"time"
 
 	"github.com/couchbaselabs/rosmar"
@@ -230,11 +231,10 @@ func UnitTestUrlIsWalrus() bool {
 	return ServerIsWalrus(UnitTestUrl())
 }
 
-func TestsRequireBootstrapConnection() bool {
-	if !TestUseCouchbaseServer() && UnitTestUrlIsWalrus() {
-		return true
+func TestsRequireBootstrapConnection(t *testing.T) {
+	if UnitTestUrlIsWalrus() {
+		t.Skipf("Tests requiring a bootstrap connection are not supporting using rosmar yet - CBG-3271")
 	}
-	return false
 }
 
 // ServerIsTLS returns true if the server URL is using an accepted secure protocol as it's prefix
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -522,9 +522,24 @@ func (h *handler) handlePutDbConfig() (err error) {
 		}
 	}
 
-	bucket := h.db.Bucket.GetName()
+	// if the database is a corrupted database config then we will have no bucket on the db to use (ah teh db context
+	// was previously closed), so nil check needed here to protect against panic
+	// is needed
+	var bucket string
+	if h.db.Bucket != nil {
+		bucket = h.db.Bucket.GetName()
+	}
 	if dbConfig.Bucket != nil {
 		bucket = *dbConfig.Bucket
+	}
+	if bucket == "" {
+		// check if the db config is in corrupt state
+		if h.server.isInCorruptState(h.db.Name) {
+			// database is in corrupt state must provide bucket name to fix
+			return base.HTTPErrorf(http.StatusInternalServerError, "database is in corrupt state must provide bucket name in config update")
+		}
+		// we should never get to this but worth covering off to protect from failure to update config
+		return base.HTTPErrorf(http.StatusInternalServerError, "missing bucket name")
 	}
 
 	// Set dbName based on path value (since db doesn't necessarily exist), and update in incoming config in case of insert
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -126,7 +126,12 @@ func (h *handler) handleCreateDB() error {
 			return base.HTTPErrorf(http.StatusInternalServerError, "couldn't load database: %v", err)
 		}
 
-		// now we've started the db successfully, we can persist it to the cluster
+		// now we've started the db successfully, we can persist it to the cluster first checking if this db used to be a corrupt db
+		// if it used to be corrupt we need to remove it from the invalid database map on server context and remove the old corrupt config from the bucket
+		err = h.fixCorruptDatabaseConfig(contextNoCancel.Ctx, bucket, h.server.Config.Bootstrap.ConfigGroupID, dbName)
+		if err != nil {
+			return err
+		}
 		cas, err := h.server.BootstrapContext.InsertConfig(contextNoCancel.Ctx, bucket, h.server.Config.Bootstrap.ConfigGroupID, &persistedConfig)
 		if err != nil {
 			// unload the requested database config to prevent the cluster being in an inconsistent state
@@ -522,24 +527,9 @@ func (h *handler) handlePutDbConfig() (err error) {
 		}
 	}
 
-	// if the database is a corrupted database config then we will have no bucket on the db to use (ah teh db context
-	// was previously closed), so nil check needed here to protect against panic
-	// is needed
-	var bucket string
-	if h.db.Bucket != nil {
-		bucket = h.db.Bucket.GetName()
-	}
+	bucket := h.db.Bucket.GetName()
 	if dbConfig.Bucket != nil {
 		bucket = *dbConfig.Bucket
-	}
-	if bucket == "" {
-		// check if the db config is in corrupt state
-		if h.server.isInCorruptState(h.db.Name) {
-			// database is in corrupt state must provide bucket name to fix
-			return base.HTTPErrorf(http.StatusInternalServerError, "database is in corrupt state must provide bucket name in config update")
-		}
-		// we should never get to this but worth covering off to protect from failure to update config
-		return base.HTTPErrorf(http.StatusInternalServerError, "missing bucket name")
 	}
 
 	// Set dbName based on path value (since db doesn't necessarily exist), and update in incoming config in case of insert
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -128,7 +128,7 @@ func (h *handler) handleCreateDB() error {
 
 		// now we've started the db successfully, we can persist it to the cluster first checking if this db used to be a corrupt db
 		// if it used to be corrupt we need to remove it from the invalid database map on server context and remove the old corrupt config from the bucket
-		err = h.fixCorruptDatabaseConfig(contextNoCancel.Ctx, bucket, h.server.Config.Bootstrap.ConfigGroupID, dbName)
+		err = h.removeCorruptConfigIfExists(contextNoCancel.Ctx, bucket, h.server.Config.Bootstrap.ConfigGroupID, dbName)
 		if err != nil {
 			return err
 		}
--- a/rest/adminapitest/admin_api_test.go
+++ b/rest/adminapitest/admin_api_test.go
@@ -1433,6 +1433,94 @@ func TestResyncStop(t *testing.T) {
 	assert.True(t, syncFnCount < 2000, "Expected syncFnCount < 2000 but syncFnCount=%d", syncFnCount)
 }
 
+// TestCorruptDbConfigHandling:
+//   - Create persistent config rest tester
+//   - Create a db with a non-corrupt config
+//   - Grab the persisted config from the bucket for purposes of changing the bucket name on it (this simulates a
+//     dbconfig becoming corrupt)
+//   - Update the persisted config to change the bucket name to a non-existent bucket
+//   - Assert that the db context and config are removed from server context and that operations on the database GET and
+//     DELETE fail with appropriate error message for user
+//   - Test we are able to update the config to correct the corrupted db config
+//   - asser the db returns to the server context and is removed from corrupt database tracking AND that the bucket name
+//     on the config now matches the rest tester bucket name
+func TestCorruptDbConfigHandling(t *testing.T) {
+	base.LongRunningTest(t)
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("Walrus/Rosmar does not support persistent config")
+	}
+	numCollections := 1
+	base.RequireNumTestDataStores(t, numCollections)
+
+	rt := rest.NewRestTester(t, &rest.RestTesterConfig{
+		CustomTestBucket: base.GetPersistentTestBucket(t),
+		PersistentConfig: true,
+		MutateStartupConfig: func(config *rest.StartupConfig) {
+			// configure the interval time to pick up new configs from the bucket to every 2 seconds
+			config.Bootstrap.ConfigUpdateFrequency = base.NewConfigDuration(2)
+		},
+	})
+	defer rt.Close()
+
+	ds := rt.CustomTestBucket.GetSingleDataStore()
+	name, ok := base.AsDataStoreName(ds)
+	require.True(t, ok)
+
+	// create db with correct config
+	resp := rt.SendAdminRequest(http.MethodPut, "/db1/", fmt.Sprintf(
+		`{"bucket": "%s", "num_index_replicas": 0, "enable_shared_bucket_access": %t, "scopes": {"%s": {"collections": {"%s":{}}}}}`,
+		rt.CustomTestBucket.GetName(), base.TestUseXattrs(), name.ScopeName(), name.CollectionName()))
+	rest.RequireStatus(t, resp, http.StatusCreated)
+
+	// wait for db to come online
+	require.NoError(t, rt.WaitForDBOnline())
+
+	// grab the persisted db config from the bucket
+	dbConfig := rest.DatabaseConfig{}
+	_, err := rt.ServerContext().BootstrapContext.GetConfig(rt.CustomTestBucket.GetName(), rt.ServerContext().Config.Bootstrap.ConfigGroupID, "db1", &dbConfig)
+	require.NoError(t, err)
+
+	// update the persisted config to a fake bucket name
+	newBucketName := "fakeBucket"
+	_, err = rt.UpdatePersistedBucketName(&dbConfig, &newBucketName)
+	require.NoError(t, err)
+
+	// wait for some time for interval to remove the db
+	time.Sleep(2 * time.Second)
+
+	// assert that the in memory representation of the db config on the server context is gone now we have broken the config
+	responseConfig := rt.ServerContext().GetDbConfig("db1")
+	assert.Nil(t, responseConfig)
+
+	// assert that fetching config fails with the correct error message to the user
+	resp = rt.SendAdminRequest(http.MethodGet, "/db1/_config", "")
+	rest.RequireStatus(t, resp, http.StatusNotFound)
+	assert.Contains(t, resp.Body.String(), "Database config corrupt.")
+
+	// assert trying to delete fails with the correct error message to the user
+	resp = rt.SendAdminRequest(http.MethodDelete, "/db1/", "")
+	rest.RequireStatus(t, resp, http.StatusNotFound)
+	assert.Contains(t, resp.Body.String(), "Database config corrupt.")
+
+	// correct the name through update to config
+	resp = rt.SendAdminRequest(http.MethodPut, "/db1/_config", fmt.Sprintf(
+		`{"bucket": "%s", "num_index_replicas": 0, "enable_shared_bucket_access": %t, "scopes": {"%s": {"collections": {"%s":{}}}}}`,
+		rt.CustomTestBucket.GetName(), base.TestUseXattrs(), name.ScopeName(), name.CollectionName()))
+	rest.RequireStatus(t, resp, http.StatusCreated)
+
+	// wait some time for interval to pick up change
+	time.Sleep(2 * time.Second)
+
+	// assert that the config is back in memory even after another interval update pass and asser the persisted config
+	// bucket name matches rest tester bucket name
+	dbCtx, err := rt.ServerContext().GetDatabase(base.TestCtx(t), "db1")
+	require.NoError(t, err)
+	assert.NotNil(t, dbCtx)
+	assert.Equal(t, rt.CustomTestBucket.GetName(), dbCtx.Bucket.GetName())
+	corruptMap := rt.ServerContext().GetCorruptDatabaseMap()
+	assert.Equal(t, 0, len(corruptMap))
+}
+
 func TestResyncStopUsingDCPStream(t *testing.T) {
 	if base.UnitTestUrlIsWalrus() {
 		// This test requires a gocb bucket
--- a/rest/adminapitest/admin_api_test.go
+++ b/rest/adminapitest/admin_api_test.go
@@ -1449,6 +1449,7 @@ func TestCorruptDbConfigHandling(t *testing.T) {
 	if base.TestsRequireBootstrapConnection() {
 		t.Skip("Walrus/Rosmar does not support persistent config")
 	}
+	base.SetUpTestLogging(t, base.LevelInfo, base.KeyConfig)
 
 	rt := rest.NewRestTester(t, &rest.RestTesterConfig{
 		CustomTestBucket: base.GetPersistentTestBucket(t),
--- a/rest/adminapitest/admin_api_test.go
+++ b/rest/adminapitest/admin_api_test.go
@@ -1491,17 +1491,17 @@ func TestCorruptDbConfigHandling(t *testing.T) {
 	// assert that fetching config fails with the correct error message to the user
 	resp = rt.SendAdminRequest(http.MethodGet, "/db1/_config", "")
 	rest.RequireStatus(t, resp, http.StatusNotFound)
-	assert.Contains(t, resp.Body.String(), "Must update database config immediately")
+	assert.Contains(t, resp.Body.String(), "You must update database config immediately")
 
 	// assert trying to delete fails with the correct error message to the user
 	resp = rt.SendAdminRequest(http.MethodDelete, "/db1/", "")
 	rest.RequireStatus(t, resp, http.StatusNotFound)
-	assert.Contains(t, resp.Body.String(), "Must update database config immediately")
+	assert.Contains(t, resp.Body.String(), "You must update database config immediately")
 
 	// correct the name through update to config
 	resp = rt.ReplaceDbConfig("db1", dbConfig)
 	rest.RequireStatus(t, resp, http.StatusNotFound)
-	assert.Contains(t, resp.Body.String(), "Must update database config immediately")
+	assert.Contains(t, resp.Body.String(), "You must update database config immediately")
 
 	// create db of same name with correct db config to correct the corrupt db config
 	resp = rt.CreateDatabase("db1", dbConfig)
--- a/rest/adminapitest/admin_api_test.go
+++ b/rest/adminapitest/admin_api_test.go
@@ -1440,48 +1440,28 @@ func TestResyncStop(t *testing.T) {
 //     dbconfig becoming corrupt)
 //   - Update the persisted config to change the bucket name to a non-existent bucket
 //   - Assert that the db context and config are removed from server context and that operations on the database GET and
-//     DELETE fail with appropriate error message for user
-//   - Test we are able to update the config to correct the corrupted db config
-//   - asser the db returns to the server context and is removed from corrupt database tracking AND that the bucket name
+//     DELETE and operation to update the config fail with appropriate error message for user
+//   - Test we are able to update the config to correct the corrupted db config using the /db1/ endpoint
+//   - assert the db returns to the server context and is removed from corrupt database tracking AND that the bucket name
 //     on the config now matches the rest tester bucket name
 func TestCorruptDbConfigHandling(t *testing.T) {
 	base.LongRunningTest(t)
 	if base.TestsRequireBootstrapConnection() {
 		t.Skip("Walrus/Rosmar does not support persistent config")
 	}
-	numCollections := 1
-	base.RequireNumTestDataStores(t, numCollections)
 
 	rt := rest.NewRestTester(t, &rest.RestTesterConfig{
 		CustomTestBucket: base.GetPersistentTestBucket(t),
 		PersistentConfig: true,
 		MutateStartupConfig: func(config *rest.StartupConfig) {
-			// configure the interval time to pick up new configs from the bucket to every 2 seconds
-			config.Bootstrap.ConfigUpdateFrequency = base.NewConfigDuration(2)
+			// configure the interval time to pick up new configs from the bucket to every 1 seconds
+			config.Bootstrap.ConfigUpdateFrequency = base.NewConfigDuration(1000000000)
 		},
 	})
 	defer rt.Close()
 
-	ds := rt.CustomTestBucket.GetSingleDataStore()
-	name, ok := base.AsDataStoreName(ds)
-	require.True(t, ok)
-
-	scopesMap := make(map[string]rest.ScopeConfig)
-	collectionMap := make(map[string]rest.CollectionConfig)
-	collectionMap[name.CollectionName()] = rest.CollectionConfig{}
-	scopesMap[name.ScopeName()] = rest.ScopeConfig{
-		Collections: collectionMap,
-	}
-
 	// create db with correct config
-	dbConfig := rest.DbConfig{
-		BucketConfig: rest.BucketConfig{
-			Bucket: base.StringPtr(rt.CustomTestBucket.GetName()),
-		},
-		NumIndexReplicas: base.UintPtr(0),
-		EnableXattrs:     base.BoolPtr(base.TestUseXattrs()),
-		Scopes:           scopesMap,
-	}
+	dbConfig := rt.NewDbConfig()
 	resp := rt.CreateDatabase("db1", dbConfig)
 	rest.RequireStatus(t, resp, http.StatusCreated)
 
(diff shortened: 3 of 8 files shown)


rest/config.go
```diff
@@ -1463,8 +1463,24 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
 		// inherit properties the bootstrap config
 		cnf.CACertPath = sc.Config.Bootstrap.CACertPath
 
+		// We need to check for corruption in the database config (CC. CBG-3292). If the fetched config doesn't match the
+		// bucket name we got the config from we need to maker this db context as corrupt. Then remove the context and
+		// in memory representation on the server context.
 		bucketCopy := bucket
-		cnf.Bucket = &bucketCopy
+		if bucket != *cnf.Bucket {
+			base.DebugfCtx(ctx, base.KeyConfig, "Mismatch in database config bucket name: %s and backend bucket: %s You must update database config immediately", *cnf.Bucket, bucket)
```
torcolvin: We almost certainly want to log this as a warning here so it doesn't rotate out of the logs, since warning logs have a very low log rotation period.

rest/admin_api.go
```diff
@@ -522,10 +522,25 @@ func (h *handler) handlePutDbConfig() (err error) {
...
+	}
 	if dbConfig.Bucket != nil {
 		bucket = *dbConfig.Bucket
 	}
+	if bucket == "" {
+		// check if the db config is in corrupt state
+		if h.server.isInCorruptState(h.db.Name) {
+			// database is in corrupt state must provide bucket name to fix
+			return base.HTTPErrorf(http.StatusInternalServerError, "database is in corrupt state must provide bucket name in config update")
+		}
```
torcolvin: I wonder if we can avoid storing the state of the corrupt databases, but we can do a read only when we do a put. Then we avoid having to worry about timing of when the bootstrap connection reads the config and we can avoid storing this as a global.

rest/adminapitest/admin_api_test.go
```diff
@@ -1433,6 +1433,94 @@ func TestResyncStop(t *testing.T) {
...
+//   - Assert that the db context and config are removed from server context and that operations on the database GET and
+//     DELETE fail with appropriate error message for user
+//   - Test we are able to update the config to correct the corrupted db config
+//   - asser the db returns to the server context and is removed from corrupt database tracking AND that the bucket name
+//     on the config now matches the rest tester bucket name
+func TestCorruptDbConfigHandling(t *testing.T) {
+	base.LongRunningTest(t)
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("Walrus/Rosmar does not support persistent config")
+	}
```
torcolvin: Can you create a method for skipping these tests. `base.TestsRequireBootstrapConnection()` which I'll enable in https://issues.couchbase.com/browse/CBG-3271.

rest/adminapitest/admin_api_test.go
```diff
@@ -1433,6 +1433,94 @@ func TestResyncStop(t *testing.T) {
...
+
+	ds := rt.CustomTestBucket.GetSingleDataStore()
+	name, ok := base.AsDataStoreName(ds)
+	require.True(t, ok)
+
+	// create db with correct config
+	resp := rt.SendAdminRequest(http.MethodPut, "/db1/", fmt.Sprintf(
+		`{"bucket": "%s", "num_index_replicas": 0, "enable_shared_bucket_access": %t, "scopes": {"%s": {"collections": {"%s":{}}}}}`,
+		rt.CustomTestBucket.GetName(), base.TestUseXattrs(), name.ScopeName(), name.CollectionName()))
+	rest.RequireStatus(t, resp, http.StatusCreated)
```
torcolvin: Can you use https://github.com/couchbase/sync_gateway/blob/master/rest/utilities_testing.go#L441

rest/adminapitest/admin_api_test.go
```diff
@@ -1433,6 +1433,94 @@ func TestResyncStop(t *testing.T) {
...
+	resp = rt.SendAdminRequest(http.MethodGet, "/db1/_config", "")
+	rest.RequireStatus(t, resp, http.StatusNotFound)
+	assert.Contains(t, resp.Body.String(), "Database config corrupt.")
+
+	// assert trying to delete fails with the correct error message to the user
+	resp = rt.SendAdminRequest(http.MethodDelete, "/db1/", "")
+	rest.RequireStatus(t, resp, http.StatusNotFound)
+	assert.Contains(t, resp.Body.String(), "Database config corrupt.")
+
+	// correct the name through update to config
```
torcolvin, line 1501: https://github.com/couchbase/sync_gateway/blob/master/rest/utilities_testing.go#L448 will do this

rest/config.go
```diff
@@ -1463,8 +1463,24 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
...
+			// track corrupt database context
+			ok := sc._addCorruptDatabase(ctx, dbName)
+			if !ok {
+				base.DebugfCtx(ctx, base.KeyConfig, "Failed ot add sb: %s to corrupt list", cnf.Name)
+			}
+			// don't load config + remove from server context (apart from corrupt database map)
+			sc._removeDatabase(ctx, cnf.Name)
```
torcolvin: We need to remove this if the database exists (but it might not) so we probably want two tests and code pathways:

1. if the database config is modified and corrupted, we want to log with a really clear message that we are unloading the database. I think the return value of `_removeDatabase` will tell us if there's an active database
2. we read the database config, but it's corrupt

I guess the reason to keep track of invalid databases is to avoid log spamming for this issue, but because it's so fatal, I wonder whether that's necessary. It would log every second, so that's bad, but having no logging about this is also bad. I think the optimal log strategy might be somewhere in the middle and creating a datastructure that holds databases without bucket names or without bucket names that locks separately from other server context locks:

Prototype:

```
type invalidBucketName struct {
        lastLogTime atomic.Int64
}

type invalidBucketNameDatabases struct {
      dbNames map[string]invalidBucketNameDatabase
      m sync.RWMutex
}

func (*d invalidBucketNameDatabases) addInvalidDatabase(dbname string) {
       // lock map
       // check last logged time, if past threshold log again
       // update log time
}
```

rest/config.go
```diff
@@ -1603,8 +1619,24 @@ func (sc *ServerContext) FetchConfigs(ctx context.Context, isInitialStartup bool
...
-			cnf.Bucket = &bucketCopy
+			if bucket != *cnf.Bucket {
+				base.DebugfCtx(ctx, base.KeyConfig, "Mismatch in database config bucket name: %s and backend bucket: %s You must update database config immediately", *cnf.Bucket, bucket)
+				// track corrupt database context
+				ok := sc._addCorruptDatabase(ctx, cnf.Name)
+				if !ok {
+					base.DebugfCtx(ctx, base.KeyConfig, "Failed ot add sb: %s to corrupt list", cnf.Name)
+				}
+				// don't load config + remove from server context (apart from corrupt database map)
+				sc._removeDatabase(ctx, cnf.Name)
```
torcolvin: If this validation is done in two functions, let's wrap this in common code

rest/handler.go
```diff
@@ -338,7 +338,20 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 	// look up the database context:
 	if keyspaceDb != "" {
 		var err error
-		if dbContext, err = h.server.GetActiveDatabase(keyspaceDb); err != nil {
+		dbContext, err = h.server.GetActiveDatabase(keyspaceDb)
```
torcolvin: If we don't let the database load, I don't understand why we need a dbcontext for PUT and POST - fixing this would require a PUT to override it but I don't think it should fail?

Maybe test these in separate functions?

rest/server_context.go
```diff
@@ -292,8 +294,18 @@ func (sc *ServerContext) GetInactiveDatabase(ctx context.Context, name string) (
 			}
 		}
 	}
+	// handle the correct error message being returned for a corrupt database config
+	var httpErr *base.HTTPError
+	sc.lock.RLock()
+	_, ok := sc.corruptDbContext[name]
+	sc.lock.RUnlock()
+	if !dbConfigFound && ok {
+		httpErr = base.HTTPErrorf(http.StatusNotFound, "Database config corrupt. Database config bucket name not equal to bucket it is persisted in. Must update database config immediately")
```
torcolvin: I think that we should add the names of the buckets and database names here, if this code has to stay, which I'm not sure it does.

gregns1: So this code needs to stay to return the correct error message for a corrupt db config, will add the extra info to error message though

rest/config.go
```diff
@@ -1463,8 +1505,17 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
...
 		cnf.CACertPath = sc.Config.Bootstrap.CACertPath
 
+		// We need to check for corruption in the database config (CC. CBG-3292). If the fetched config doesn't match the
+		// bucket name we got the config from we need to maker this db context as corrupt. Then remove the context and
+		// in memory representation on the server context.
 		bucketCopy := bucket
-		cnf.Bucket = &bucketCopy
+		if bucket != *cnf.Bucket {
+			sc.handleInvalidDatabaseConfig(ctx, bucket, cnf)
+			return true, fmt.Errorf("mismatch in peristed databse bucket name vs the actual bucketname. Please correct.")
```
torcolvin:
```suggestion
			return true, fmt.Errorf("mismatch in persisted database bucket name %q vs the actual bucket name %q. Please correct.", *cnf.Bucket, bucket)
```

rest/config.go
```diff
@@ -1463,8 +1505,17 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
...
+		} else {
+			// no corruption detected carry on as usual
+			cnf.Bucket = &bucketCopy
+		}
```
torcolvin:
```suggestion
		cnf.Bucket = &bucketCopy
```

we can drop the else statement here for readability (nit)

rest/config.go
```diff
@@ -1603,8 +1661,17 @@ func (sc *ServerContext) FetchConfigs(ctx context.Context, isInitialStartup bool
...
+			// in memory representation on the server context.
 			bucketCopy := bucket
-			cnf.Bucket = &bucketCopy
+			if bucket != *cnf.Bucket {
+				sc.handleInvalidDatabaseConfig(ctx, bucket, *cnf)
+				continue
+			} else {
+				// no corruption detected carry on as usual
+				cnf.Bucket = &bucketCopy
+			}
```
torcolvin:
```suggestion
			cnf.Bucket = &bucketCopy
```

we could drop the else statement here for readability (nit)

rest/handler.go
```diff
@@ -588,6 +588,24 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 	return nil
 }
 
+// fixCorruptDatabaseConfig will remove the config from the bucket and remove it from the map if it exists on the invalid database config map
+func (h *handler) fixCorruptDatabaseConfig(ctx context.Context, bucket, configGroupID, dbName string) error {
+	h.server.invalidDatabaseConfigTracking.m.Lock()
+	defer h.server.invalidDatabaseConfigTracking.m.Unlock()
+
+	_, ok := h.server.invalidDatabaseConfigTracking.dbNames[dbName]
```
torcolvin: for readability:

rather than worry locking inside datastructures, I'd do something like

`h.server.invalidDatabaseConfigTracking.exists(dbname)`

which acquires only a read lock and then

`h.server.invalidDatabaseConfigTracking.remove(dbname)`

which would acquire a writelock when you call 

```
delete(h.server.invalidDatabaseConfigTracking.dbNames, dbName)
```

Unless you are worried about this having a race condition? You can test this with simultaneous calls to create DB but I'm not sure that this will work right now anyway.

rest/utilities_testing.go
```diff
@@ -2498,6 +2498,12 @@ func DropAllTestIndexes(t *testing.T, tb *base.TestBucket) {
 	}
 }
 
+func (sc *ServerContext) GetCorruptDatabaseMap() map[string]*invalidConfigInfo {
+	sc.invalidDatabaseConfigTracking.m.RLock()
+	defer sc.invalidDatabaseConfigTracking.m.RUnlock()
+	return sc.invalidDatabaseConfigTracking.dbNames
+}
```
torcolvin: This is only used in testing, so I'm not so particular but I might do something that was like

```
func (sc *ServerContext) RequireInvalidDatabaseConfigNames(t *testing.T, dbNames []string)
```

So you don't have to worry about mutating the structure, and you can only call this function from a test environment. And then you can just fail with an appropriate message.

rest/adminapitest/admin_api_test.go
```diff
@@ -1433,6 +1433,110 @@ func TestResyncStop(t *testing.T) {
...
+	_, err := rt.ServerContext().BootstrapContext.GetConfig(rt.CustomTestBucket.GetName(), rt.ServerContext().Config.Bootstrap.ConfigGroupID, "db1", &databaseConfig)
+	require.NoError(t, err)
+
+	// update the persisted config to a fake bucket name
+	newBucketName := "fakeBucket"
+	_, err = rt.UpdatePersistedBucketName(&databaseConfig, &newBucketName)
+	require.NoError(t, err)
+
+	// wait for some time for interval to remove the db
+	time.Sleep(2 * time.Second)
```
torcolvin: rather than waiting for specific time, can we call a function with a backoff until `GetDbConfig` returns nil? That way if things happen faster we don't have to wait.

Generally a sleep in a test is a bad code smell.

rest/adminapitest/admin_api_test.go
```diff
@@ -1433,6 +1433,110 @@ func TestResyncStop(t *testing.T) {
...
+
+	// create db with correct config
+	dbConfig := rest.DbConfig{
+		BucketConfig: rest.BucketConfig{
+			Bucket: base.StringPtr(rt.CustomTestBucket.GetName()),
+		},
+		NumIndexReplicas: base.UintPtr(0),
+		EnableXattrs:     base.BoolPtr(base.TestUseXattrs()),
+		Scopes:           scopesMap,
+	}
```
torcolvin: `rt.NewDbConfig` should return this

rest/adminapitest/admin_api_test.go
```diff
@@ -1433,6 +1433,110 @@ func TestResyncStop(t *testing.T) {
...
+	rest.RequireStatus(t, resp, http.StatusCreated)
+
+	// wait some time for interval to pick up change
+	time.Sleep(2 * time.Second)
+
+	// assert that the config is back in memory even after another interval update pass and asser the persisted config
+	// bucket name matches rest tester bucket name
+	dbCtx, err := rt.ServerContext().GetDatabase(base.TestCtx(t), "db1")
+	require.NoError(t, err)
+	assert.NotNil(t, dbCtx)
```
torcolvin: again with this sleep, can we do a wait in a `RetryLoop`?

rest/adminapitest/admin_api_test.go
```diff
@@ -1433,6 +1433,110 @@ func TestResyncStop(t *testing.T) {
...
+	}
+	numCollections := 1
+	base.RequireNumTestDataStores(t, numCollections)
+
+	rt := rest.NewRestTester(t, &rest.RestTesterConfig{
+		CustomTestBucket: base.GetPersistentTestBucket(t),
+		PersistentConfig: true,
+		MutateStartupConfig: func(config *rest.StartupConfig) {
+			// configure the interval time to pick up new configs from the bucket to every 2 seconds
+			config.Bootstrap.ConfigUpdateFrequency = base.NewConfigDuration(2)
```
torcolvin: I believe this is two nanoseconds since the default is `10 * time.Second`

```
config.Bootstrap.ConfigUpdateFrequency = 2 * time.Nanosecond
```

although then I'd ask why this is not 1 second when I read the code. I think it's great to make this run as fast as possible but I'd like to make it clear about units

rest/adminapitest/admin_api_test.go
```diff
@@ -1433,6 +1433,110 @@ func TestResyncStop(t *testing.T) {
...
+//   - Test we are able to update the config to correct the corrupted db config
+//   - asser the db returns to the server context and is removed from corrupt database tracking AND that the bucket name
+//     on the config now matches the rest tester bucket name
+func TestCorruptDbConfigHandling(t *testing.T) {
+	base.LongRunningTest(t)
+	if base.TestsRequireBootstrapConnection() {
+		t.Skip("Walrus/Rosmar does not support persistent config")
+	}
+	numCollections := 1
+	base.RequireNumTestDataStores(t, numCollections)
```
torcolvin: there should always be one collection - if it requires a named collection, you should test for that - use `SG_TEST_USE_DEFAULT_COLLECTION=true` and `SG_TEST_USE_GSI=false` permutations.

torcolvin (commented):
I left comments on the code, and the code comments will hold true regardless, but there is a bigger problem, maybe.

We store the map by dbName but if you run a backup on `bucketA` with `db1` to `bucketB` and `bucketC`, there are actually _three_ buckets with a matching database name, but we index the invalid names by dbname only so fixing in `bucketB` will remove this from `bucketC`. This code may behave OK, but if that's the case, I'd prefer a test for this. I'd probably try to break that test from the other test cases, or at least parametrize the test for the different cases to make it easier to understand what the test is doing in the future.

Additionally, I'm worried about the frequency of logging here and I'd like other people's input. It's definitely possible that this isn't possible to fix quickly and I don't know the right level to log on. It might be right to log the first time on warning and then log on info for the subsequent errors. (Just throwing out ideas).

rest/handler.go
```diff
@@ -588,6 +588,24 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 	return nil
 }
 
+// fixCorruptDatabaseConfig will remove the config from the bucket and remove it from the map if it exists on the invalid database config map
+func (h *handler) fixCorruptDatabaseConfig(ctx context.Context, bucket, configGroupID, dbName string) error {
```
torcolvin: I'd rename this to `removeCorruptConfigIfExists` because it doesn't fix the configuration, it will just delete the config file.

Is this necessary because we are presumably updating it in `fetchConfig` and the CAS is changing requiring a reread?

gregns1: My test was failing without this method. Failure reason that there is already a config with same name in the bucket, so I think this is needed unless code else where isn't behaving as excepted.

rest/server_context.go
```diff
@@ -292,8 +298,18 @@ func (sc *ServerContext) GetInactiveDatabase(ctx context.Context, name string) (
 			}
 		}
 	}
+	// handle the correct error message being returned for a corrupt database config
+	var httpErr *base.HTTPError
+	sc.invalidDatabaseConfigTracking.m.RLock()
+	invalidConfig, ok := sc.invalidDatabaseConfigTracking.dbNames[name]
+	sc.invalidDatabaseConfigTracking.m.RUnlock()
```
torcolvin: add an `exists()` to this map which handles locking

torcolvin (commented):
As discussed offline, we need to worry about different cases and can write tests for them - I'd each of these as a separate test case so it's clear what they are doing.

1. start `RestTester` with no buckets and write a dbconfig to a bucket with the SDK that is an invalid bucket (this is a restore of a bucket with nothing running). Assert that the database doesn't exist or we hit the point in time where it doesn't work
2. make sure if you do a PUT with a mismatched bucket name in the config it will fail (this may already be tested)
3. in `bucketA` and `bucketB`, write two dbconfigs with bucket name as `bucketC`; then start RestTester, make sure neither of them are picked up (this is the case where we have two buckets that are backed up/restored but no working buckets)
4. create `bucketA` and `bucketB` with dbconfigs that that both list bucket name as `bucketA`, then start RestTester make sure database for bucketA is picked up and there is logging for `bucketB` having the wrong bucket name (this is the case where we did a backup/restore of bucketA as bucketB and started SG)
5. create a working dbconfig in `bucketA` in a resttester, make sure db is running. write the config into `bucketB` while the RestTester is running, make sure the database backed by `bucketA` is still running and there's appropriate logging. (this is the case where SG is running and backup/restore hits)

These roughly match the scenarios of what happens if you start a SG when you have multiple buckets with different information and also what happens if a SG is already running and a bucket appears.

Because an invalid database can never appear, I don't think we need code in `handler.go` - if there's an bucket name mismatch, we just won't load the database, so you can't query it with `/db` -> it'll just return 404 as not found because the database wasn't loaded.

The last change we want to make to the bootstrap logic is to log a warning the first time we see a mismatched bucket, but then after log a warning. This is a change from the code on a timer that I wrote.

rest/adminapitest/admin_api_test.go
```diff
@@ -1433,6 +1433,393 @@ func TestResyncStop(t *testing.T) {
...
+	defer tb1.Close()
+	tb2 := base.GetPersistentTestBucket(t)
+	defer tb2.Close()
+
+	rt := rest.NewRestTester(t, &rest.RestTesterConfig{
+		CustomTestBucket: tb1,
+		PersistentConfig: true,
+		MutateStartupConfig: func(config *rest.StartupConfig) {
+			// configure the interval time to pick up new configs from the bucket to every 1 seconds
+			config.Bootstrap.ConfigUpdateFrequency = base.NewConfigDuration(1000000000)
```
torcolvin: very small nit

```suggestion
			// configure the interval time to pick up new configs from the bucket to every 1 seconds
			config.Bootstrap.ConfigUpdateFrequency = base.NewConfigDuration(1 * time.Second)
```

base/constants.go
```diff
@@ -230,6 +230,13 @@ func UnitTestUrlIsWalrus() bool {
 	return ServerIsWalrus(UnitTestUrl())
 }
 
+func TestsRequireBootstrapConnection() bool {
+	if !TestUseCouchbaseServer() && UnitTestUrlIsWalrus() {
+		return true
+	}
+	return false
+}
```
torcolvin: nit, shortens code below, note rename to match other test functions - let's move this to `util_testing.go`.

```suggestion
func TestRequiresBootstrapConnection(t *testing.T) {
	if UnitTestUrlIsWalrus() {
		t.Skipf("Tests requiring a bootstrap connection are not supporting using rosmar yet - CBG-3271")
	}
}
```

rest/adminapitest/admin_api_test.go
```diff
@@ -1433,6 +1433,393 @@ func TestResyncStop(t *testing.T) {
...
+		t.Skip("Walrus/Rosmar does not support persistent config")
+	}
+	base.SetUpTestLogging(t, base.LevelInfo, base.KeyConfig)
+
+	rt := rest.NewRestTester(t, &rest.RestTesterConfig{
+		CustomTestBucket: base.GetPersistentTestBucket(t),
+		PersistentConfig: true,
+		MutateStartupConfig: func(config *rest.StartupConfig) {
+			// configure the interval time to pick up new configs from the bucket to every 1 seconds
+			config.Bootstrap.ConfigUpdateFrequency = base.NewConfigDuration(1000000000)
```
torcolvin: nit

```suggestion
			config.Bootstrap.ConfigUpdateFrequency = base.NewConfigDuration(1 * time.Second)
```

rest/adminapitest/admin_api_test.go
```diff
@@ -1433,6 +1433,393 @@ func TestResyncStop(t *testing.T) {
...
+//   - Assert that the db context and config are removed from server context and that operations on the database GET and
+//     DELETE and operation to update the config fail with appropriate error message for user
+//   - Test we are able to update the config to correct the corrupted db config using the /db1/ endpoint
+//   - assert the db returns to the server context and is removed from corrupt database tracking AND that the bucket name
+//     on the config now matches the rest tester bucket name
+func TestCorruptDbConfigHandling(t *testing.T) {
+	base.LongRunningTest(t)
+	if base.TestsRequireBootstrapConnection() {
+		t.Skip("Walrus/Rosmar does not support persistent config")
+	}
```
torcolvin:
```suggestion
	base.TestRequiresBootstrapConnection(t)
```

rest/adminapitest/admin_api_test.go
```diff
@@ -1433,6 +1433,393 @@ func TestResyncStop(t *testing.T) {
...
+
+// TestBadConfigInsertionToBucket:
+//   - start a rest tester
+//   - insert an invalid db config to the bucket while rest tester is running
+//   - assert that the db config is picked up as an invalid db config
+//   - assert that a call to the db endpoint will fail with correct error message
+func TestBadConfigInsertionToBucket(t *testing.T) {
+	if base.TestsRequireBootstrapConnection() {
+		t.Skip("Walrus/Rosmar does not support persistent config")
+	}
```
torcolvin:
```suggestion
	base.TestRequiresBootstrapConnection(t)
```

rest/adminapitest/admin_api_test.go
```diff
@@ -1433,6 +1433,393 @@ func TestResyncStop(t *testing.T) {
...
+	base.RequireNumTestBuckets(t, 2)
+	tb1 := base.GetPersistentTestBucket(t)
+	defer tb1.Close()
+
+	rt := rest.NewRestTester(t, &rest.RestTesterConfig{
+		CustomTestBucket: base.GetPersistentTestBucket(t),
+		PersistentConfig: true,
+		MutateStartupConfig: func(config *rest.StartupConfig) {
+			// configure the interval time to pick up new configs from the bucket to every 1 seconds
+			config.Bootstrap.ConfigUpdateFrequency = base.NewConfigDuration(1000000000)
```
torcolvin: nit

```suggestion
	base.TestRequiresBootstrapConnection(t)
```

rest/adminapitest/admin_api_test.go
```diff
@@ -1433,6 +1433,393 @@ func TestResyncStop(t *testing.T) {
...
+	rest.RequireStatus(t, resp, http.StatusNotFound)
+}
+
+// TestMultipleBucketWithBadDbConfigScenario1:
+//   - in bucketA and bucketB, write two db configs with bucket name as bucketC
+//   - Start new rest tester and ensure they aren't picked up as valid configs
+func TestMultipleBucketWithBadDbConfigScenario1(t *testing.T) {
+	if base.TestsRequireBootstrapConnection() {
+		t.Skip("Walrus/Rosmar does not support persistent config")
+	}
```
torcolvin: nit

```suggestion
	base.TestRequiresBootstrapConnection(t)
```

rest/adminapitest/admin_api_test.go
```diff
@@ -1433,6 +1433,393 @@ func TestResyncStop(t *testing.T) {
...
+	tb2 := base.GetPersistentTestBucket(t)
+	defer tb2.Close()
+	tb3 := base.GetPersistentTestBucket(t)
+	defer tb3.Close()
+
+	rt1 := rest.NewRestTester(t, &rest.RestTesterConfig{
+		CustomTestBucket: tb1,
+		PersistentConfig: true,
+		MutateStartupConfig: func(config *rest.StartupConfig) {
+			// configure same config groupID
```
torcolvin: nit

```suggestion
			// all RestTesters all this test must use the same config ID
```

pull group id into a constant.

You can also set this with `RestTesterConfig.GroupID`. https://github.com/couchbase/sync_gateway/blob/2689cf8e564c46d5cc38edc01264c0f051a8e433/rest/utilities_testing.go#L230

rest/config.go
```diff
@@ -266,6 +267,55 @@ func GetTLSVersionFromString(stringV *string) uint16 {
...
+	d.m.Lock()
+	defer d.m.Unlock()
+	if d.dbNames[dbname] == nil {
+		// db hasn't been tracked as invalid config yet so add it
+		d.dbNames[dbname] = &configInfo
+	}
+	// if we get here we already have the db logged as an invalid config, so now we need to work out iof we should log for it now
+	if !d.dbNames[dbname].logged {
+		// we need to log at warning if we get here
+		base.WarnfCtx(d.loggingCtx, "Mismatch in database config for database %s bucket name: %s and backend bucket: %s You must update database config immediately", dbname, d.dbNames[dbname].configBucketName, d.dbNames[dbname].persistedBucketName)
```
torcolvin: I think we can make this comment more clear:

"The database config for %q in bucket %q points to %q as the bucket in the configuration will. You must update database config immediately", base.MD(dbname, d.dbNames[dbname].persistedBucketName), (d.dbNames[dbname].configBucketName)

I'd prefer to update this function to only have this string once since it would be easy to find it one place and not the other. We should add base.MD() since this is technically redactable information, though I don't believe we redact this at this time. I haven't tried this but you might be able to declare this `[]fmt.Stringer{"msg", base.MD("foo"), }

We could almost put the logging function as a function pointer to deduplicate but WarnfCtx takes fewer arguments than InfofCtx

rest/config.go
```diff
@@ -266,6 +267,55 @@ func GetTLSVersionFromString(stringV *string) uint16 {
...
 
+type invalidConfigInfo struct {
+	logged              bool
+	configBucketName    string
+	persistedBucketName string
+}
+
+type invalidDatabaseConfigs struct {
+	dbNames    map[string]*invalidConfigInfo
+	loggingCtx context.Context
```
torcolvin: Storing a context on a struct is an anti-pattern https://go.dev/blog/context-and-structs so we are trying to fix this as we go, let's pass this into `addInvalidDatabase`

In addition to confusion, the biggest problem is that we can easily pass `nil` here.

rest/config.go
```diff
@@ -1463,7 +1513,15 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
 		// inherit properties the bootstrap config
 		cnf.CACertPath = sc.Config.Bootstrap.CACertPath
 
+		// We need to check for corruption in the database config (CC. CBG-3292). If the fetched config doesn't match the
+		// bucket name we got the config from we need to maker this db context as corrupt. Then remove the context and
+		// in memory representation on the server context.
 		bucketCopy := bucket
```
torcolvin: I'd move this below the if statement

rest/config.go
```diff
@@ -1463,7 +1513,15 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
...
+		if bucket != *cnf.Bucket {
+			sc.handleInvalidDatabaseConfig(ctx, bucket, cnf)
+			return true, fmt.Errorf("mismatch in persisted database bucket name %q vs the actual bucket name %q. Please correct.", cnf.Name, bucket)
```
torcolvin: Can we add dbname and group id here and add `base.MD()` around the names?

rest/config.go
```diff
@@ -1603,7 +1668,15 @@ func (sc *ServerContext) FetchConfigs(ctx context.Context, isInitialStartup bool
 			// inherit properties the bootstrap config
 			cnf.CACertPath = sc.Config.Bootstrap.CACertPath
 
+			// We need to check for corruption in the database config (CC. CBG-3292). If the fetched config doesn't match the
+			// bucket name we got the config from we need to maker this db context as corrupt. Then remove the context and
+			// in memory representation on the server context.
 			bucketCopy := bucket
```
torcolvin: can you move this below the `handleInvalidDatabaseConfig`

rest/handler.go
```diff
@@ -588,6 +588,21 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 	return nil
 }
 
+// removeCorruptConfigIfExists will remove the config from the bucket and remove it from the map if it exists on the invalid database config map
+func (h *handler) removeCorruptConfigIfExists(ctx context.Context, bucket, configGroupID, dbName string) error {
+	_, ok := h.server.invalidDatabaseConfigTracking.exists(dbName)
+	if ok {
```
torcolvin: nit, I'd prefer early returns for readability

```
if !ok {
      return true
}
```

rest/server_context.go
```diff
@@ -292,8 +298,16 @@ func (sc *ServerContext) GetInactiveDatabase(ctx context.Context, name string) (
 			}
 		}
 	}
+	// handle the correct error message being returned for a corrupt database config
+	var httpErr *base.HTTPError
+	invalidConfig, ok := sc.invalidDatabaseConfigTracking.exists(name)
+	if !dbConfigFound && ok {
+		httpErr = base.HTTPErrorf(http.StatusNotFound, "Database %s config corrupt. Database config bucket name: %s not equal to bucket it is persisted in: %s. Must update database config immediately", name, invalidConfig.configBucketName, invalidConfig.persistedBucketName)
```
torcolvin: I'd be inclined to have a single message about corruption, if the `fmt.Stringer` idea works that be great.

We want to add groupID and `base.MD`

torcolvin (commented):
Most comments are just nitpicks about readability.

There is one bigger problem which I think it's worth writing a new test or two for - what happens if you have

1. dbconfig A, groupID A backed by bucket A in bucket A
2. dbconfig B, groupID B backed by bucket A in bucket B
3. dbconfig C, groupID A backed by bucket A in bucket B

It's possible most tests need to do include groupID so make sure that we are tracking dbName, bucket, groupID because it seems like our invalid database structure drops groupID.

Generally I think groupID is used to have a duplicate configuration but avoid having certain nodes participate in import.

torcolvin:
> Most comments are just nitpicks about readability.
> 
> There is one bigger problem which I think it's worth writing a new test or two for - what happens if you have
> 
> 1. dbconfig A, groupID A backed by bucket A in bucket A
> 2. dbconfig B, groupID B backed by bucket A in bucket B
> 3. dbconfig C, groupID A backed by bucket A in bucket B
> 
> It's possible most tests need to do include groupID so make sure that we are tracking dbName, bucket, groupID because it seems like our invalid database structure drops groupID.
> 
> Generally I think groupID is used to have a duplicate configuration but avoid having certain nodes participate in import.

This is actually not an issue outside of tests because Sync Gateway only has a single `ServerContext` and it can only have one groupID

torcolvin (approved):

PR #6379, opened 2023-08-23
CBG-3289 handle in memory walrus databases as rosmar databases

torcolvin:
Handle in memory walrus databases as rosmar databases. Tests to make sure that old-style in memory databases (`walrus:data`), rosmar in memory (rosmar:/?mode=memory), and rosmar on disk all will work without being part of a bucket pool.

Removed a context.TODO()/context.Background() because there was always a clear calling context.

## Pre-review checklist

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -35,6 +35,7 @@ import (
 	"github.com/couchbase/sync_gateway/base"
 	"github.com/couchbase/sync_gateway/db"
 	"github.com/couchbase/sync_gateway/db/functions"
+	"github.com/couchbaselabs/rosmar"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


rest/config.go
```diff
@@ -88,7 +89,14 @@ func (dc *DbConfig) MakeBucketSpec() base.BucketSpec {
 	tlsPort := 11207
 
 	if bc.Server != nil {
-		server = *bc.Server
+		// walrus:data represents in memory data store, convert this to rosmar.InMemoryURL
+		// but do not match walrus:///path/to/db, which is incompatible with rosmar URLs
```
bbrks: I left a more detailed comment in Slack about this, but `walrus:data` is a persisted Walrus bucket, with a relative path to `data` (and only if that directory exists... and then it's not persisted 🙃)

bbrks (dismissed):

adamcfraser (approved):

PR #6388, opened 2023-09-04
CBG-3357: Change log levels for StartOnlineProcesses

bbrks:
CBG-3357

For logging before/after `StartOnlineProcesses`:
- Raise log levels/remove Config log key

- [x] n/a

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -411,7 +411,7 @@ func NewDatabaseContext(ctx context.Context, dbName string, bucket base.Bucket, 
 	// but in db package tests this is unlikely to be set. In this case we'll use the existing bucket connection to store metadata.
 	metadataStore := options.MetadataStore
 	if metadataStore == nil {
-		base.DebugfCtx(ctx, base.KeyConfig, "MetadataStore was nil - falling back to use existing bucket connection %q for database %q", bucket.GetName(), dbName)
+		base.DebugfCtx(ctx, base.KeyAll, "MetadataStore was nil - falling back to use existing bucket connection %q for database %q", bucket.GetName(), dbName)
 		metadataStore = bucket.DefaultDataStore()
 	}
 
@@ -2018,7 +2018,7 @@ func initDatabaseStats(dbName string, autoImport bool, options DatabaseContextOp
 
 	var collections []string
 	if len(options.Scopes) == 0 {
-		base.DebugfCtx(context.TODO(), base.KeyConfig, "No named collections defined for database %q, using default collection for stats", base.MD(dbName))
+		base.DebugfCtx(context.TODO(), base.KeyAll, "No named collections defined for database %q, using default collection for stats", base.MD(dbName))
 		collections = append(collections, base.DefaultScope+"."+base.DefaultCollection)
 	} else {
 		for scopeName, scope := range options.Scopes {
--- a/db/database.go
+++ b/db/database.go
@@ -402,7 +402,7 @@ func NewDatabaseContext(ctx context.Context, dbName string, bucket base.Bucket, 
 		return nil, err
 	}
 
-	dbStats, statsError := initDatabaseStats(dbName, autoImport, options)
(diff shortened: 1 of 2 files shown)


rest/server_context.go
```diff
@@ -924,11 +924,12 @@ func (sc *ServerContext) asyncDatabaseOnline(nonCancelCtx base.NonCancellableCon
 		return
 	}
 
-	base.DebugfCtx(ctx, base.KeyConfig, "Async database initialization complete, starting online processes...")
+	base.InfofCtx(ctx, base.KeyAll, "Async database initialization complete, starting online processes...")
 	err := dbc.StartOnlineProcesses(ctx)
 	if err != nil {
-		base.InfofCtx(ctx, base.KeyAll, "Error starting online processes after async initialization: %v", err)
+		base.WarnfCtx(ctx, "Error starting online processes after async initialization: %v", err)
```
bbrks: Arguably this should be an error instead of just a warning - the database failed to come online and it will require user-interaction to either retry or correct?

rest/server_context.go
```diff
@@ -924,11 +924,12 @@ func (sc *ServerContext) asyncDatabaseOnline(nonCancelCtx base.NonCancellableCon
...
 		atomic.CompareAndSwapUint32(&dbc.State, db.DBStarting, db.DBOffline)
+		return
```
torcolvin: This is a really bad bug - is there a way that we can hit this in a unit test?

bbrks: I think I can force an error with either:
- A Bucket level TTL (https://github.com/couchbase/sync_gateway/blob/aff7f541bee8f8a3cfd26a6e10ddaa9376937e84/db/database.go#L2392C1-L2394)
- A user with only a subset of JWT properties set (https://github.com/couchbase/sync_gateway/blob/64ab69664e07f7a7108cc965a89fa74addc79d57/db/users.go#L48)

I'll see if I can stick either of these into a test (I think the user options will be more testable with Walrus/Rosmar)

bbrks: I've bashed my head trying to get "real-world" coverage (aka `RestTester`) of this all day, and it's a lot trickier than first appears. 

- I can't use bucket TTLs, as there's another (earlier) check in NewDatabaseContext that prevents the database from being created in the first place.
- I can't use invalid user config, because that's only supported in non-persistent config mode, and async online is only supported in persistent config mode.
- I can't force an error from LeakyDataStore when LastSequence is called, as persistent config means the leaky bucket wrapper gets lost.

The only remaining option I can see for a RestTester-based test is to provide a malformed OIDC callback URL, which _does_ correctly error... BUT I now have trouble hitting the codepath that runs the async `StartOnlineProcesses`, because there's an interaction between `"offline":true` and an already running "active async initialization" that I cannot hit in the right way from a test...

The only other option I have is to have a lower-level unit test only for `asyncDatabaseOnline` to make sure the database state remains Offline and doesn't revert to Online...

bbrks: I've managed to cover this line/state transition in a new test: `TestPersistentDbConfigAsyncOnlineWithInvalidConfig`

rest/server_context.go
bbrks: Bumped to Error

bbrks:
I've split the db state transition fix into #6392 and kept this PR purely as logging changes

db/database.go
```diff
@@ -2018,7 +2018,7 @@ func initDatabaseStats(dbName string, autoImport bool, options DatabaseContextOp
 
 	var collections []string
 	if len(options.Scopes) == 0 {
-		base.DebugfCtx(context.TODO(), base.KeyConfig, "No named collections defined for database %q, using default collection for stats", base.MD(dbName))
+		base.DebugfCtx(context.TODO(), base.KeyAll, "No named collections defined for database %q, using default collection for stats", base.MD(dbName))
```
torcolvin: Can you add a context.Context here?

torcolvin (approved):

PR #6390, opened 2023-09-05
CBG-3352 remove invalid entries from revcache

torcolvin:
Change `Invalidate` function from rev cache to remove the value from the cache. Ran `-race` 10 times without issues.

[x] https://jenkins.sgwdev.com/job/SyncGateway-Integration/1987/

### DIFF
--- a/db/revision_cache_lru.go
+++ b/db/revision_cache_lru.go
@@ -75,29 +75,29 @@ func (sc *ShardedLRURevisionCache) Invalidate(ctx context.Context, docID, revID 
 
 // An LRU cache of document revision bodies, together with their channel access.
 type LRURevisionCache struct {
-	cache        map[IDAndRev]*list.Element // Fast lookup of list element by doc/rev ID
-	lruList      *list.List                 // List ordered by most recent access (Front is newest)
-	capacity     uint32                     // Max number of revisions to cache
-	backingStore RevisionCacheBackingStore  // provides the methods used by the RevisionCacheLoaderFunc
-	lock         sync.Mutex                 // For thread-safety
+	backingStore RevisionCacheBackingStore
+	cache        map[IDAndRev]*list.Element
+	lruList      *list.List
 	cacheHits    *base.SgwIntStat
 	cacheMisses  *base.SgwIntStat
+	lock         sync.Mutex
+	capacity     uint32
 }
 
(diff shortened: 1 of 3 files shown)


db/revision_cache_lru.go
```diff
@@ -147,7 +147,10 @@ func (rc *LRURevisionCache) getFromCache(ctx context.Context, docID, revID strin
 		return DocumentRevision{}, nil
 	}
 
-	if value.invalid {
+	value.lock.RLock()
```
adamcfraser: This is the check that has performance implications, particularly on systems with a large number of cores, where acquiring a read lock can require scheduling work across cores.  We've seen this in our performance testing in the past, which is how we ended up with the design that attempts to keep the value immutable.  
As discussed, the invalidation flag breaks this immutability so we're going to need to change something.  I don't think switching back to acquiring a read lock on value is the ideal approach, as invalidation is an uncommon use case, and we'd be taking on that cost on every rev cache read.

adamcfraser (changes requested):

adamcfraser (approved):

adamcfraser:
Looks good - thanks!

db/revision_cache_lru.go
```diff
@@ -273,9 +263,10 @@ func (rc *LRURevisionCache) Upsert(ctx context.Context, docRev DocumentRevision)
 
 func (rc *LRURevisionCache) Invalidate(ctx context.Context, docID, revID string) {
 	value := rc.getValue(docID, revID, false)
-	if value != nil {
-		value.setInvalidFlag()
+	if value == nil {
+		return
 	}
+	rc.removeValue(value)
```
bbrks, line 269: `getValue` does a non-zero amount of work to update the LRU, even though we're about to remove the entry, as well as aquiring the revcache lock in both get and remove.

Maybe it's better to implement a remove method that only requires key?

You could essentially do:

```go
key := IDAndRev{DocID: docID, RevID: revID}
rc.lock.Lock()
defer rc.lock.Unlock()
if element := rc.cache[key]; element != nil {
	rc.lruList.Remove(element)
}
delete(rc.cache, key)
```

bbrks (changes requested):
Suggestion for alternative method of removal

PR #6392, opened 2023-09-06
CBG-3382: Catch invalid state transition in asyncDatabaseOnline

bbrks:
Spun out of #6388 

CBG-3382

- Test to force an error from `StartOnlineProcesses` from `asyncDatabaseOnline` that causes an invalid state transition via missed `return`
- Catch and panic with `atomic.CompareAndSwap`

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1989/

### DIFF
--- a/rest/server_context.go
+++ b/rest/server_context.go
@@ -934,7 +934,7 @@ func (sc *ServerContext) asyncDatabaseOnline(nonCancelCtx base.NonCancellableCon
 
 	if !atomic.CompareAndSwapUint32(&dbc.State, db.DBStarting, db.DBOnline) {
 		// 2nd atomic might end up being Starting here if there's a legitimate race, but it's the most we can do for CAS
-		panic(fmt.Sprintf("database state wasn't Starting during asyncDatabaseOnline Online transition... now %q", db.RunStateString[atomic.LoadUint32(&dbc.State)]))
+		base.PanicfCtx(ctx, "database state wasn't Starting during asyncDatabaseOnline Online transition... now %q", db.RunStateString[atomic.LoadUint32(&dbc.State)])
 	}
 
 	stateChangeMsg := "DB loaded from config"
--- a/rest/server_context.go
+++ b/rest/server_context.go
@@ -929,12 +929,16 @@ func (sc *ServerContext) asyncDatabaseOnline(nonCancelCtx base.NonCancellableCon
 	if err != nil {
 		base.InfofCtx(ctx, base.KeyAll, "Error starting online processes after async initialization: %v", err)
 		atomic.CompareAndSwapUint32(&dbc.State, db.DBStarting, db.DBOffline)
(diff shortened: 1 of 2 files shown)


rest/server_context.go
```diff
@@ -929,12 +929,16 @@ func (sc *ServerContext) asyncDatabaseOnline(nonCancelCtx base.NonCancellableCon
 	if err != nil {
 		base.InfofCtx(ctx, base.KeyAll, "Error starting online processes after async initialization: %v", err)
 		atomic.CompareAndSwapUint32(&dbc.State, db.DBStarting, db.DBOffline)
+		return
+	}
+
+	if !atomic.CompareAndSwapUint32(&dbc.State, db.DBStarting, db.DBOnline) {
+		// 2nd atomic might end up being Starting here if there's a legitimate race, but it's the most we can do for CAS
+		panic(fmt.Sprintf("database state wasn't Starting during asyncDatabaseOnline Online transition... now %q", db.RunStateString[atomic.LoadUint32(&dbc.State)]))
```
torcolvin: Should we `PanicfCtx` and say that we can't do anything to recover from this state in the comment?

We can't take down a database because it's already coming online, so the best we can do is take down SG?

Or do you think this is unlikely to get hit and not worth panicking about?

bbrks: Hmm... This was mostly to catch it more obviously via test if we decided to forget another `return` again. I'm not sure what we should do really. I don't expect this to end up panicing in a released version, but I might expect it in development/testing.

I agree about at least using `PanicfCtx` over `panic` though. Better chance of getting that logging captured.

rest/server_context.go
```diff
@@ -934,7 +934,7 @@ func (sc *ServerContext) asyncDatabaseOnline(nonCancelCtx base.NonCancellableCon
 
 	if !atomic.CompareAndSwapUint32(&dbc.State, db.DBStarting, db.DBOnline) {
 		// 2nd atomic might end up being Starting here if there's a legitimate race, but it's the most we can do for CAS
-		panic(fmt.Sprintf("database state wasn't Starting during asyncDatabaseOnline Online transition... now %q", db.RunStateString[atomic.LoadUint32(&dbc.State)]))
+		base.PanicfCtx(ctx, "database state wasn't Starting during asyncDatabaseOnline Online transition... now %q", db.RunStateString[atomic.LoadUint32(&dbc.State)])
```
torcolvin, line 937: Does ctx here have the database name? If so, that's a cool benefit of `PanicfCtx`.

If it doesn't, can you add the database name in the error?

bbrks: It does

torcolvin: Awesome!

torcolvin (approved):

PR #6407, opened 2023-09-07
CBG-3393: Bump to latest Go version (1.21.1)

bbrks:
CBG-3393

https://go.dev/doc/devel/release#go1.21.1

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1993/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v4
         with:
-          go-version: 1.20.3
+          go-version: 1.21.1
       - name: go-build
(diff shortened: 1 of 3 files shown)


torcolvin (approved):
I like this, but I think you should bump sg-bucket and rosmar too - they won't really need a subsequent version bump because it will only be a CI change.

torcolvin:
I am definitely supportive of bumping the go version but the go versions are baked into jenkins for the standard integration test runs, so we might want to create a new Jenkins project with go 1.20 labelled 3.1 release

bbrks:
3.1 is built with `1.19.5` and we don't have a release tied to `1.20.x` so I opted to just bump the main job.

PR #6414, opened 2023-09-12
Be more specific about errors during invalid response for BLIP GetAttachment

bbrks:
Be more specific about what check failed when a peer sends invalid attachment data. Make `"length"` explicitly required, instead of failing when it's going to end up as a failure to do `ToInt64(nil)`

## Testing

Already covered via `TestPutInvalidAttachment`

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1999/


torcolvin (approved):
Should we add redacted docID in error message, or do we know from the handler?

adamcfraser (approved):

PR #6419, opened 2023-09-15
CBG-3291 Define macro expansion in MutateInOpts

adamcfraser:
CBG-3291

Uptake sg-bucket changes to pass macro expansions for a given operation to the XattrStore implementation (*Collection), instead of hardcoding in the low-level subdoc operations.

The macro expansion options are set in three places for non-test code:
- updateAndReturnDoc, handles standard writes/imports via WriteUpdateWithXattr
- migrateMetadata, performs migration via WriteWithXattr
- resyncDocument, uses WriteUpdateWithXattr (different retry than updateAndReturnDoc)

Prerequisite for the expanded use of macro expansion for HLV properties.

Also includes some cleanup of the XattrStore interface to remove internal-only operations.

- [x] rosmar https://github.com/couchbaselabs/rosmar/pull/10
- [x] sg-bucket https://github.com/couchbase/sg-bucket/pull/104

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2020/


base/collection_xattr.go
```diff
@@ -630,3 +626,27 @@ func (c *Collection) DeleteUserXattr(k string, xattrKey string) (uint64, error)
 	}
 	return uint64(result.Cas()), nil
 }
+
+// initializeMutateInSpec will append macro expansions defined in MutateInOptions to the provided
+// gocb MutateInSpec.
+func (c *Collection) appendMacroExpansions(mutateInSpec []gocb.MutateInSpec, opts *sgbucket.MutateInOptions) []gocb.MutateInSpec {
```
torcolvin: I'm loosely inclined to make this not be a method on `Collection` because it doesn't have a dependency on `Collection`

db/crud.go
```diff
@@ -2582,3 +2585,25 @@ func (db *DatabaseCollectionWithUser) CheckProposedRev(ctx context.Context, doci
 		return ProposedRev_Conflict, doc.CurrentRev
 	}
 }
+
+const (
+	xattrMacroCas         = "cas"
```
torcolvin, line 2590: should we pull this value from sg-bucket? not convinced either way

db/crud.go
```diff
@@ -1899,6 +1899,10 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 	if db.UseXattrs() || upgradeInProgress {
 		var casOut uint64
 		// Update the document, storing metadata in extended attribute
+		if opts == nil {
+			opts = &sgbucket.MutateInOptions{}
```
torcolvin, line 1903: is this `DefaultMutateInOpts()` ?

db/database.go
```diff
@@ -1819,15 +1819,17 @@ func (db *DatabaseCollectionWithUser) resyncDocument(ctx context.Context, docid,
...
 				raw, rawXattr, err = updatedDoc.MarshalWithXattr()
 				return raw, rawXattr, deleteDoc, updatedExpiry, err
 			} else {
 				return nil, nil, deleteDoc, nil, base.ErrUpdateCancel
 			}
 		}
-		_, err = db.dataStore.WriteUpdateWithXattr(key, base.SyncXattrName, db.userXattrKey(), 0, nil, nil, writeUpdateFunc)
+		opts := &sgbucket.MutateInOptions{
+			MacroExpansion: macroExpandSpec(base.SyncXattrName),
+		}
```
torcolvin, line 1831: Is this `DefaultMutateInOpts()` ?

db/crud.go
adamcfraser: I think it belongs in SGW as this defines the document property that's using the macro expansion, not the macro expansion type.

db/crud.go
adamcfraser: It's possible that the caller has already set MutateInOptions.PreserveExpiry, so just initializing here and then adding the MacroExpansion (for either case) below.

db/database.go
adamcfraser: I intended DefaultMutateInOpts() to be test-only, as I thought the non-test usages should be more intentional about what they are setting.  I think I'd prefer to leave as-is for now, and possibly revisit during future HLV work that makes the default more complex.

torcolvin (dismissed):

torcolvin (approved):

PR #6423, opened 2023-09-18
CBG-3378 Switch some Debug statements to Info

torcolvin:
This was based on an analysis of `rg Debugf` and judging if they were error states. Some cases we return the error to the user, but I'm not convinced the users are always going to check the response codes.

These choices were largely arbitrary.


auth/auth.go
```diff
@@ -787,12 +787,12 @@ func (auth *Authenticator) AuthenticateUser(username string, password string) (U
 func (auth *Authenticator) AuthenticateUntrustedJWT(rawToken string, oidcProviders OIDCProviderMap, localJWT LocalJWTProviderMap, callbackURLFunc OIDCCallbackURLFunc) (User, PrincipalConfig, error) {
 	token, err := jwt.ParseSigned(rawToken)
 	if err != nil {
-		base.DebugfCtx(auth.LogCtx, base.KeyAuth, "Error parsing JWT in AuthenticateUntrustedJWT: %v", err)
+		base.InfofCtx(auth.LogCtx, base.KeyAuth, "Error parsing JWT in AuthenticateUntrustedJWT: %v", err)
```
adamcfraser: All of this AuthenticateUntrustedJWT logging It looks like this could be logged based on malformed JWTs provided by clients.  Given that, I don't know if we want spam info logs with these messages if a bad/malicious client is spamming authentication requests.

db/import.go
```diff
@@ -245,7 +245,7 @@ func (db *DatabaseCollectionWithUser) importDoc(ctx context.Context, docid strin
 			}
 
 			if importErr != nil {
-				base.DebugfCtx(ctx, base.KeyImport, "Error returned for doc %s while evaluating import function - will not be imported.", base.UD(docid))
+				base.InfofCtx(ctx, base.KeyImport, "Error returned for doc %s while evaluating import function - will not be imported.", base.UD(docid))
```
adamcfraser: Should this be logging importErr as well?

adamcfraser (dismissed):

db/import.go
```diff
@@ -245,7 +245,7 @@ func (db *DatabaseCollectionWithUser) importDoc(ctx context.Context, docid strin
 			}
 
 			if importErr != nil {
-				base.DebugfCtx(ctx, base.KeyImport, "Error returned for doc %s while evaluating import function - will not be imported.", base.UD(docid))
+				base.InfofCtx(ctx, base.KeyImport, "Error returned for doc %s while evaluating import function - will not be imported. %s", base.UD(docid), err)
```
bbrks, line 248:
```suggestion
				base.InfofCtx(ctx, base.KeyImport, "Error returned for doc %s while evaluating import function - will not be imported. %s", base.UD(docid), importErr)
```

bbrks (changes requested):

auth/auth.go
torcolvin: I removed these because I don't really know, but I thought about putting them in so you'd know if there was a bad/malicious client and what they were doing wrong so you could block them.

db/import.go
torcolvin: Fixed this by shadowing err.

bbrks (approved):

PR #6433, opened 2023-09-19
Lower test logging verbosity for non-flaky tests

bbrks:
These tests aren't flaky (as far as I can find in Jira) - so put logging back to Info


gregns1 (approved):
LGTM

PR #6438, opened 2023-09-20
CBG-3238 perform a length check on raw bytes

torcolvin:
- ensure code doesn't panic if invalid xattr are found over DCP stream

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2025/


db/document.go
```diff
@@ -515,6 +515,9 @@ func parseXattrStreamData(xattrName string, userXattrName string, data []byte) (
 	}
 
 	xattrsLen := binary.BigEndian.Uint32(data[0:4])
+	if int(xattrsLen) > len(data) {
```
adamcfraser: This looks like it should check if int(xattrsLen+4) > len(data) to avoid all possible panics on line 518.

db/document.go
```diff
@@ -515,6 +515,9 @@ func parseXattrStreamData(xattrName string, userXattrName string, data []byte) (
...
+		return nil, nil, nil, fmt.Errorf("%w (%d) from bytes %s", base.ErrXattrInvalidLen, xattrsLen, data)
+	}
```
adamcfraser, line 520: Are we able to write a test for what happens if data[0:4] doesn't represent xattrsLen (i.e. document doesn't have xattrs, it's just the body), but the Uint32 value of data[0:4] is less than len(data)?  I'm assuming we'll return an error while trying to parse the xattr key/value pairs below, but would be good to verify there aren't any subsequent panics in that case.

db/attachment_compaction.go
```diff
@@ -29,102 +29,115 @@ const (
...
-	var markProcessFailureErr error
-
-	// failProcess used when a failure is deemed as 'un-recoverable' and we need to abort the compaction process.
-	failProcess := func(err error, format string, args ...interface{}) bool {
-		markProcessFailureErr = err
-		terminator.Close()
-		base.WarnfCtx(ctx, format, args...)
-		return false
+// processAttachmentCompactSweepCallback(ctx context.Context, dataStore base.DataStore, db *Database, compactionID, compactionLoggingID string, dryRun bool, purgedAttachmentCount *base.AtomicInt, event sgbucket.FeedEvent) {
+func processAttachmentCompactMarkCallback(ctx context.Context, dataStore base.DataStore, compactionID, compactionLoggingID string, markedAttachmentCount *base.AtomicInt, event sgbucket.FeedEvent) error {
```
adamcfraser: Can you provide some more context on the need for these attachment compaction changes?  I don't see how they are strictly related to the length check enhancement.

adamcfraser (changes requested):

db/document.go
torcolvin: Fixed.

db/document.go
torcolvin: Added a test for this.

db/attachment_compaction.go
torcolvin: As discussed offline, I removed these changes because they were bigger than was appropriate. I left a single change to treat a document with an invalid xattr as a document with xattrs.

adamcfraser (approved):

PR #6444, opened 2023-09-21
Add dependabot module updates

bbrks:
I would like to use this as a mechanism to at least give us notifications for module updates, and ideally semi-automate the update.

Dependabot will open individual PRs for any upstream module updates.
We can just choose whether to merge them or close them.

More info here: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates


gregns1 (approved):
LGTM

PR #6445, opened 2023-09-21
CBG-3394 don't panic if user is deleted with active blip replication

torcolvin:
Use common function that we use in changes feed to refresh user.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2026/

### DIFF
--- a/rest/blip_api_crud_test.go
+++ b/rest/blip_api_crud_test.go
@@ -2826,7 +2826,7 @@ func TestBlipRefreshUser(t *testing.T) {
 	require.NoError(t, err)
 
 	testResponse := unsubChangesRequest.Response()
-	require.Equal(t, strconv.Itoa(http.StatusInternalServerError), testResponse.Properties[db.BlipErrorCode])
+	require.Equal(t, strconv.Itoa(db.CBLReconnectErrorCode), testResponse.Properties[db.BlipErrorCode])
 	body, err := testResponse.Body()
 	require.NoError(t, err)
 	require.NotContains(t, string(body), "Panic:")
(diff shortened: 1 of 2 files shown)


bbrks (approved):

rest/blip_api_crud_test.go
```diff
@@ -2770,3 +2771,63 @@ func TestRequestPlusPullDbConfig(t *testing.T) {
...
+
+	_, ok = btc.GetRev(docID, "1-78211b5eedea356c9693e08bc68b93ce")
+	require.True(t, ok)
+
+	// delete user with an active blip connection
+	response = rt.SendAdminRequest(http.MethodDelete, "/{{.db}}/_user/"+username, "")
+	RequireStatus(t, response, http.StatusOK)
+
+	// further requests will 500, but shouldn't panic
+	unsubChangesRequest := blip.NewRequest()
```
bbrks, line 2821: Actually I think this is the wrong status code to be sending.

A 503 error will force CBL to reconnect, which in this scenario will try to auth again with the now deleted user and fail. I think this is the behaviour we want.

bbrks (changes requested):

rest/blip_api_crud_test.go
torcolvin: Does this break backward compatibility? Also, what's the benefit of reconnecting here?

bbrks: The benefit is the replication on the client stops when the user is deleted with a very obvious/explicit error, rather than carrying on and continously getting 500 errors for each message.

I don't expect this is a case where compatibility is guaranteed one way or another.

bbrks (approved):

PR #6451, opened 2023-09-22
CBG-3399: TestUserXattrDeleteWithRevCache flake fix

gregns1:
CBG-3399

Bump up import partitions on rest tester as just having killed the import feed when having multiple nodes sometimes. I tried having a single node with auto import on. Still failed sometimes, this seems to stop it failing.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2027/
- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2028/
- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2031/
- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2051/

### DIFF
--- a/rest/utilities_testing.go
+++ b/rest/utilities_testing.go
@@ -328,7 +328,7 @@ func (rt *RestTester) Bucket() base.Bucket {
 		autoImport, _ := rt.DatabaseConfig.AutoImportEnabled(ctx)
 		if rt.DatabaseConfig.ImportPartitions == nil && base.TestUseXattrs() && base.IsEnterpriseEdition() && autoImport {
 			// Speed up test setup - most tests don't need more than one partition given we only have one node
-			rt.DatabaseConfig.ImportPartitions = base.Uint16Ptr(2)
+			rt.DatabaseConfig.ImportPartitions = base.Uint16Ptr(1)
 		}
 
 		_, isLeaky := base.AsLeakyBucket(rt.TestBucket)
--- a/rest/utilities_testing.go
+++ b/rest/utilities_testing.go
@@ -328,7 +328,7 @@ func (rt *RestTester) Bucket() base.Bucket {
 		autoImport, _ := rt.DatabaseConfig.AutoImportEnabled(ctx)
 		if rt.DatabaseConfig.ImportPartitions == nil && base.TestUseXattrs() && base.IsEnterpriseEdition() && autoImport {
 			// Speed up test setup - most tests don't need more than one partition given we only have one node
-			rt.DatabaseConfig.ImportPartitions = base.Uint16Ptr(1)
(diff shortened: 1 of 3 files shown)


rest/utilities_testing.go
```diff
@@ -328,7 +328,7 @@ func (rt *RestTester) Bucket() base.Bucket {
 		autoImport, _ := rt.DatabaseConfig.AutoImportEnabled(ctx)
 		if rt.DatabaseConfig.ImportPartitions == nil && base.TestUseXattrs() && base.IsEnterpriseEdition() && autoImport {
 			// Speed up test setup - most tests don't need more than one partition given we only have one node
-			rt.DatabaseConfig.ImportPartitions = base.Uint16Ptr(1)
+			rt.DatabaseConfig.ImportPartitions = base.Uint16Ptr(2)
```
torcolvin: I don't know 2 is a sufficient number? If you have more than 2 rest testers backing a single bucket, this will similarly fail?

Is it worthwhile to just leave the default? On a standard test, how much does this slow down?

gregns1: Slows down a lot.
Example test run for the test in question:
Default partitions = 21 seconds 
2 partitions = 7 seconds 

We don't have many tests that have more than two rest testers, the ones i've seen where we have more than two aren't running import feeds.

torcolvin: Gross - maybe we should explicitly set `ImportPartitions` in the tests that need it - these would be tests that are testing import across multiple rest testers.

gregns1: So ran a integration test on this branch and on master branch. Only difference being import partitions being 2 on here and 1 on master. Master took 1h 1 min and this branch took 59 mins. So overall changing to 2 import partitions as default seems to not slow the integration tests down too much so i think we're safe to merge as a temp fix and i'll file tickets to carry out more investigation into this. 

 I also ran another test with auto import enabled: TestUserXattrAutoImport
 With 2 partitions - 6 seconds 
 With default 16 partitions - 18 seconds

torcolvin: This is still a 3% slowdown - maybe we just change ImportPartitions in the test that are needed and mark this line for investigation in https://issues.couchbase.com/browse/CBG-3438

gregns1: It actually isn't, this branch actually finished faster in an integration test than master did. But will update the PR with the above when I get chance.

torcolvin: Ok, I'm fine with pushing this as is, but I'd mark this line with this ticket for association.

torcolvin (approved):

PR #6460, opened 2023-09-25
CBG-2579: TestReplicatorCheckpointOnStop test fix 

gregns1:
CBG-2579

Not able to repro the failiure locally.

But the first thing that jumps out is that the test fails to fetch the persisted checkpoint, suggesting its simply not there. A comment in the test previously suggested setting the checkpoint interval to 0 is disabling the interval. This is incorrect, it will default to the default value in this context which is 5 seconds. See -> https://github.com/couchbase/sync_gateway/blob/ac740e85e06f965fff11a6f7972e8c8c56d6834b/db/active_replicator_common.go#L93C3-L93C3 I have now replaced this with `reduceTestCheckpointInterval` to increase the checkpoint frequency and 50 seconds. Hopefully will stop the test from flaking.  

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2036/

### DIFF
--- a/rest/replicatortest/replicator_test.go
+++ b/rest/replicatortest/replicator_test.go
@@ -7949,8 +7949,8 @@ func TestReplicatorCheckpointOnStop(t *testing.T) {
 	defer teardown()
 	activeCtx := activeRT.Context()
 
-	// reduce checkpointing interval temporarily to ensure there is a checkpoint to assert against later in test
-	defer reduceTestCheckpointInterval(50 * time.Millisecond)()
+	// increase checkpointing interval temporarily to ensure there is a checkpoint to assert against later
+	defer reduceTestCheckpointInterval(50 * time.Second)()
 
 	rev, doc, err := activeRT.GetSingleTestDatabaseCollectionWithUser().Put(activeCtx, "test", db.Body{})
 	require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


rest/replicatortest/replicator_test.go
```diff
@@ -7949,8 +7949,8 @@ func TestReplicatorCheckpointOnStop(t *testing.T) {
 	defer teardown()
 	activeCtx := activeRT.Context()
 
-	// Disable checkpointing at an interval
-	activeRT.GetDatabase().SGReplicateMgr.CheckpointInterval = 0
+	// reduce checkpointing interval temporarily to ensure there is a checkpoint to assert against later in test
+	defer reduceTestCheckpointInterval(50 * time.Millisecond)()
```
bbrks: If we're checkpointing at any interval, how can we be sure the replicator synchronously checkpointed on stop as opposed to the checkpointer ticker doing it?

The intention of the test is that the replicator doesn't rely on the checkpointer interval when a replicator is stopped.

Could we set the interval to something really high if `0` doesn't work?

bbrks (changes requested):

rest/replicatortest/replicator_test.go
gregns1: Ah I got the wrong end of the stick. Added your suggestion!

But I wonder if there is contention between the stop checkpointer and the checkpoint interval and thats the reason the checkpoint is missing sometimes? Its hard to fully debug without reproing it. I don't see wither of these log lines in the lastest failiure case on jenkins -> https://github.com/couchbase/sync_gateway/blob/6bf619004c1098d8a69d31fe2e7cda45d108ce6f/db/active_replicator_checkpointer.go#L245C12-L245C12 https://github.com/couchbase/sync_gateway/blob/6bf619004c1098d8a69d31fe2e7cda45d108ce6f/db/active_replicator_checkpointer.go#L248 I did think about adding a 'wait for stat' on the set checkpoint count on the active replicator before trying to assert against it but I don't think that'll help this test case.

bbrks (approved):

PR #6463, opened 2023-09-25
CBG-3386 turn off failOnRollback for resync

torcolvin:
## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2037/


bbrks:
I'd like @gregns1 to review this if possible, as I don't fully understand the implications of only switching this bool (where do we "handle" the rollback? Is it already done inside the DCP Client code?)

gregns1 (commented):
Although the follback side to this is handled in our DCP client code now, I think it would be good to have a unit test covering this. 
I think we could have a unit test similar to TestAttachmentCompactionMarkPhaseRollback where we can manual alter metadata to force a rollback during resync. Then assert that the docs process is greater than the total number of docs created on the database?  This would at least show that the client is rollback back as expected and resync process is reprocessing some docs?

gregns1 (approved):
LGTM

PR #6473, opened 2023-09-27
CBG-3448: alter WriteUpdateWithXattrFunc definitions in sync gateway

gregns1:
CBG-3448

Small PR to make sync gateway compatible with update to WriteUpdateWithXattrFunc callback. Chnages are needed for the PR https://github.com/couchbase/sync_gateway/pull/6366 to work. 

## Pre-review checklist

- [x] https://github.com/couchbase/sg-bucket/pull/107
- [x] https://github.com/couchbaselabs/rosmar/pull/13
- [x] Update Go module dependencies when above two are merged merged

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2057/

### DIFF
--- a/base/collection_xattr_common.go
+++ b/base/collection_xattr_common.go
@@ -201,7 +201,7 @@ func WriteUpdateWithXattr(ctx context.Context, store *Collection, k string, xatt
 		}
 
 		// Invoke callback to get updated value
-		updatedValue, updatedXattrValue, isDelete, callbackExpiry, err := callback(value, xattrValue, userXattrValue, cas)
+		updatedValue, updatedXattrValue, isDelete, callbackExpiry, macroOpts, err := callback(value, xattrValue, userXattrValue, cas)
 
 		// If it's an ErrCasFailureShouldRetry, then retry by going back through the for loop
 		if err == ErrCasFailureShouldRetry {
@@ -214,6 +214,9 @@ func WriteUpdateWithXattr(ctx context.Context, store *Collection, k string, xatt
 		}
 		if callbackExpiry != nil {
 			exp = *callbackExpiry
+		}
+		if len(macroOpts) != 0 {
+			opts.MacroExpansion = append(opts.MacroExpansion, macroOpts...)
 		}
 
 		// Attempt to write the updated document to the bucket.  Mark body for deletion if previous body was non-empty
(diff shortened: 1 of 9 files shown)


base/collection_xattr_common.go
```diff
@@ -215,6 +215,9 @@ func WriteUpdateWithXattr(ctx context.Context, store *Collection, k string, xatt
 		if callbackExpiry != nil {
 			exp = *callbackExpiry
 		}
+		if updatedOpts != nil {
+			opts = updatedOpts
+		}
```
adamcfraser, line 220: If we assume the callback is doing something like "append some additional macro expansion options", then I feel like this block runs the risk of ending up with duplicates if this loop runs multiple times (cas mismatch on line 260).  Any reason we can't just leave opts intact, and pass updatedOpts to WriteWithXattr?

base/collection_xattr_common.go
```diff
@@ -201,7 +201,7 @@ func WriteUpdateWithXattr(ctx context.Context, store *Collection, k string, xatt
 		}
 
 		// Invoke callback to get updated value
-		updatedValue, updatedXattrValue, isDelete, callbackExpiry, err := callback(value, xattrValue, userXattrValue, cas)
+		updatedValue, updatedXattrValue, isDelete, callbackExpiry, updatedOpts, err := callback(value, xattrValue, userXattrValue, opts, cas)
```
adamcfraser: I'm on the fence about whether the callback should take opts as an input (as opposed to just having it return []sgbucket.MacroExpansionSpec that would get appended by the caller).  Is there a use case (now or future) where you think the callback might want to reference existing opts content?

adamcfraser (dismissed):

base/collection_xattr_common.go
gregns1: This makes sense to me, will update to do this.

base/collection_xattr_common.go
gregns1: I was also on the fence about returning the macro expansion spec list or the whole mutate in options. I originally went for the mutate in spec to be returned then changed to return the whole options struct because I felt like it gave us more freedom to expand upon in future if needed. But looking at it this morning after your comments I agree, i'll update the PRs to reflect this.

bbrks (dismissed):

bbrks (approved):

PR #6480, opened 2023-09-29
CBG-3464 Delete legacy config when migrated config already exists

adamcfraser:
CBG-3464

In the case where a 3.1 version of the config exists, delete duplicate 3.0 config if found with matching database and config group during migration.

Also adds a test for migrate handling with use_xattr_config=true.

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2072/

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -1472,9 +1472,9 @@ func (sc *ServerContext) migrateV30Configs(ctx context.Context) error {
 		if insertErr != nil {
 			if insertErr == base.ErrAlreadyExists {
 				base.DebugfCtx(ctx, base.KeyConfig, "Found legacy config for database %s, but already exists in registry.", base.MD(dbConfig.Name))
-				continue
+			} else {
+				return fmt.Errorf("Error migrating v3.0 config for bucket %s groupID %s: %w", base.MD(bucketName), base.MD(groupID), insertErr)
 			}
-			return fmt.Errorf("Error migrating v3.0 config for bucket %s groupID %s: %w", base.MD(bucketName), base.MD(groupID), insertErr)
 		}
 		removeErr := sc.BootstrapContext.Connection.DeleteMetadataDocument(ctx, bucketName, PersistentConfigKey30(ctx, groupID), legacyCas)
 		if removeErr != nil {
--- a/rest/config.go
+++ b/rest/config.go
@@ -1464,7 +1464,8 @@ func (sc *ServerContext) migrateV30Configs(ctx context.Context) error {
 		if getErr == base.ErrNotFound {
 			continue
 		} else if getErr != nil {
-			return fmt.Errorf("Error retrieving 3.0 config for bucket: %s, groupID: %s: %w", bucketName, groupID, getErr)
+			base.InfofCtx(ctx, base.KeyConfig, "Unable to retrieve 3.0 config during config migration for bucket: %s, groupID: %s: %w", base.MD(bucketName), base.MD(groupID), getErr)
+			continue
 		}
 
 		base.InfofCtx(ctx, base.KeyConfig, "Found legacy persisted config for database %s - migrating to db registry.", base.MD(dbConfig.Name))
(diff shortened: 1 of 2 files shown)


rest/config.go
```diff
@@ -1461,9 +1461,9 @@ func (sc *ServerContext) migrateV30Configs(ctx context.Context) error {
 		if insertErr != nil {
 			if insertErr == base.ErrAlreadyExists {
 				base.DebugfCtx(ctx, base.KeyConfig, "Found legacy config for database %s, but already exists in registry.", base.MD(dbConfig.Name))
-				continue
+			} else {
+				return fmt.Errorf("Error migrating v3.0 config for bucket %s groupID %s: %w", base.MD(bucketName), base.MD(groupID), insertErr)
```
torcolvin:
```suggestion
				return base.ReadactErrorf("Error migrating v3.0 config for bucket %s groupID %s: %w", base.MD(bucketName), base.MD(groupID), insertErr)
```

rest/config.go
```diff
@@ -1461,9 +1461,9 @@ func (sc *ServerContext) migrateV30Configs(ctx context.Context) error {
...
 			}
-			return fmt.Errorf("Error migrating v3.0 config for bucket %s groupID %s: %w", base.MD(bucketName), base.MD(groupID), insertErr)
 		}
 		removeErr := sc.BootstrapContext.Connection.DeleteMetadataDocument(ctx, bucketName, PersistentConfigKey30(ctx, groupID), legacyCas)
```
torcolvin, line 1481: Is it worth worrying about another node hitting this delete simultaneously, or should we just not worry about the remove error?

An idea might be pass on a KeyNotFoundError, but fail on other errors (SDK timeout)?

rest/persistent_config_test.go
```diff
@@ -1138,6 +1138,95 @@ func TestMigratev30PersistentConfig(t *testing.T) {
...
+		serverErr <- StartServer(ctx, &config, sc)
+	}()
+	require.NoError(t, sc.WaitForRESTAPIs(ctx))
+
+	// Get a test bucket, and use it to create the database.
+	tb := base.GetTestBucket(t)
+	defer func() {
+		fmt.Println("closing test bucket")
+		tb.Close(ctx)
+	}()
```
torcolvin, line 1180: I think you can use `RestTester` for this?

rest/config.go
bbrks: Frustratingly `RedactableError` doesn't work with `%w`, as it uses `fmt.Sprintf` under the hood (which also doesn't support `%w`)

https://issues.couchbase.com/browse/CBG-2997

But I have just had an idea of how to make this work - will experiment and open PR if it works

adamcfraser: Good catch, there's an earlier error that should follow the same pattern, will fix both.

rest/config.go
adamcfraser: Considering this further, I don't think we should be terminating iteration through the buckets on an error for a single bucket (particularly since config migration is designed as a best effort, and doesn't block startup).  Going to rework the earlier cases that are returning error to instead log and continue.

rest/persistent_config_test.go
adamcfraser: This test is primarily doing bootstrap operations - what's the benefit of RestTester?

torcolvin: The benefit would be that there is potentially less boilerplate and consistent ways of starting a `ServerContext` and closing the server context and it would open a bucket for you.

I'm not set on this and you have to be able to pass `BootstrapOptions` into `RestTester`. I have seen this pattern before where it is used because we didn't have support in `RestTester` for the given functionality.

adamcfraser: I agree with making that change, but would prefer to minimize the changes in this PR, to manage the backport impact (the other 8 usages of this pattern in this file should also be changed).     Filed https://issues.couchbase.com/browse/CBG-3498 for this enhancement.

torcolvin (approved):

PR #6482, opened 2023-09-29
CBG-3353: Handle nil buckets in db config

bbrks:
CBG-3353

- Handles `nil` buckets in existing db configs (uses shared `dbConfig.GetBucketName()` function)
- Stamps buckets into new or updated db configs to avoid in future

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2069/

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -80,6 +80,11 @@ func (h *handler) handleCreateDB() error {
 			metadataID = h.server.BootstrapContext.standardMetadataID(config.Name)
 		}
 
+		// Before taking a copy of the "persistedDbConfig", stamp bucket if missing
+		if config.Bucket == nil {
+			config.Bucket = &dbName
+		}
+
 		// copy config before setup to persist the raw config the user supplied
 		var persistedDbConfig DbConfig
 		if err := base.DeepCopyInefficient(&persistedDbConfig, config); err != nil {
@@ -95,7 +100,8 @@ func (h *handler) handleCreateDB() error {
 		loadedConfig := DatabaseConfig{
 			Version:    version,
 			MetadataID: metadataID,
-			DbConfig:   *config}
+			DbConfig:   *config,
+		}
 
 		persistedConfig := DatabaseConfig{
 			Version:    version,
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -200,8 +200,12 @@ func getAuthScopeHandleCreateDB(ctx context.Context, h *handler) (bucketName str
 	}
 
 	if dbConfigBody.Bucket == "" {
+		dbName := h.PathVar("newdb")
+		if dbName == "" {
+			return "", base.HTTPErrorf(http.StatusBadRequest, "bucket or db name not specified in request")
+		}
 		// imply bucket name from db name in path if not in body
-		return h.PathVar("newdb"), nil
+		return dbName, nil
 	}
 
 	return dbConfigBody.Bucket, nil
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -50,7 +50,7 @@ func (h *handler) handleCreateDB() error {
 
 	validateOIDC := !h.getBoolQuery(paramDisableOIDCValidation)
 
-	if dbName != config.Name && config.Name != "" {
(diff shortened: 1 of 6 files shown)


rest/admin_api.go
```diff
@@ -173,22 +177,34 @@ func (h *handler) handleCreateDB() error {
...
+	err = DecodeAndSanitiseConfig(ctx, reader, &dbConfigBody, false)
 	if err != nil {
 		return "", err
 	}
-	if body.Bucket == "" {
-		return "", nil
+
+	if dbConfigBody.Bucket == "" {
+		// imply bucket name from db name in path if not in body
+		return h.PathVar("newdb"), nil
```
torcolvin: Is it worth checking if `newdb` is empty string here? I don't think it could but it might be better to error rather than try to figure it out later if we re-use getAuthScopeHandleCreateDB

torcolvin (commented):
Is there an advantage to the `h.ctx()`-> `ctx` changes?

One advantage of `h.ctx()` is that you can update it to include db/keyspace as soon as it is available.

rest/persistent_config_test.go
```diff
@@ -1238,3 +1238,106 @@ func makeDbConfig(bucketName string, dbName string, scopesConfig ScopesConfig) D
 	}
 	return dbConfig
 }
+
+func TestPersistentConfigNoBucketField(t *testing.T) {
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("This test only works against Couchbase Server")
+	}
```
torcolvin:
```suggestion
	 base.TestsRequireBootstrapConnection(t)
```

rest/persistent_config_test.go
```diff
@@ -1238,3 +1238,106 @@ func makeDbConfig(bucketName string, dbName string, scopesConfig ScopesConfig) D
...
+
+	base.SetUpTestLogging(t, base.LevelTrace, base.KeyConfig)
+
+	b1 := base.GetTestBucket(t)
+	defer b1.Close(base.TestCtx(t))
+	b1Name := b1.GetName()
+	t.Logf("b1: %s", b1Name)
```
torcolvin:
```suggestion
```

rest/persistent_config_test.go
```diff
@@ -1238,3 +1238,106 @@ func makeDbConfig(bucketName string, dbName string, scopesConfig ScopesConfig) D
...
+
+	// at the end of the test we'll move config from b1 into b2 to test backup/restore-type migration
+	b2 := base.GetTestBucket(t)
+	defer b2.Close(base.TestCtx(t))
+	b2Name := b2.GetName()
+	t.Logf("b2: %s", b2Name)
```
torcolvin:
```suggestion
```

rest/persistent_config_test.go
```diff
@@ -1238,3 +1238,106 @@ func makeDbConfig(bucketName string, dbName string, scopesConfig ScopesConfig) D
 	}
 	return dbConfig
 }
+
+func TestPersistentConfigNoBucketField(t *testing.T) {
+	if base.UnitTestUrlIsWalrus() {
```
torcolvin: Can you mark this as requiring as 2 test buckets?

rest/config.go
```diff
@@ -364,6 +364,8 @@ func (dbConfig *DbConfig) setDatabaseCredentials(credentials base.CredentialsCon
 // setup populates fields in the dbConfig
 func (dbConfig *DbConfig) setup(ctx context.Context, dbName string, bootstrapConfig BootstrapConfig, dbCredentials, bucketCredentials *base.CredentialsConfig, forcePerBucketAuth bool) error {
 	dbConfig.Name = dbName
+
+	// use db name as bucket if absent from config (handling for non-stamped configs)
```
torcolvin: To clarify for future: Non stamped configs are configurations that are explicitly added with a REST api without `name` and also legacy configurations that don't have bucket in them?

bbrks: It's basically just handling for pre-existing configs that were written at the time without a bucket field. Don't expect newer configs to have an empty bucket field any more.

torcolvin (approved):

PR #6486, opened 2023-10-02
CBG-3445 prevent N-1 downgrades by stamping db version into gateway registry

torcolvin:
- update gateway registry on database startup to include new version numbers
- ensure that a database is not loaded if is more than a minor version upgrade

Improvements to ComparableVersion

- make copyable so that you can't pass a nil version by mistake
- change String() method to be a non pointer method to avoid mishaps if you don't print a pointer.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2076/


base/version_comparable.go
```diff
@@ -49,31 +47,35 @@ func NewComparableVersionFromString(version string) (*ComparableVersion, error)
...
 		epoch:   epoch,
 		major:   major,
 		minor:   minor,
 		patch:   patch,
 		other:   other,
 		build:   build,
 		edition: edition,
-	}, nil
+	}
+	v.str = v.formatComparableVersion()
```
adamcfraser, line 62: Are you thinking this is safer than v.str=version?  If so, should version != v.formatComparableVersion() return error?

base/version_comparable_test.go
```diff
@@ -115,6 +115,71 @@ func TestInvalidComparableVersion(t *testing.T) {
...
+func TestAtLeastMinorDowngradeVersion(t *testing.T) {
+	testCases := []struct {
+		versionA       string
+		versionB       string
+		minorDowngrade bool
+	}{
+		{
+			versionA:       "1.0.0",
+			versionB:       "1.0.0",
+			minorDowngrade: false,
```
adamcfraser, line 146: I think versionA: `1.0.0.2` versionB: `1.0.0.1` would be a good case to add to be sure we avoid regressions.

rest/config.go
```diff
@@ -1439,6 +1439,21 @@ func (sc *ServerContext) _fetchAndLoadDatabase(nonContextStruct base.NonCancella
 	return true, nil
 }
 
+// stampSGVersion looks at each of the GatewayRegistry and updates the Sync Gateway version to start them on
```
adamcfraser: I think the comment needs editing, since it's currently only doing one bucket/registry at a time.

rest/config.go
```diff
@@ -1439,6 +1439,21 @@ func (sc *ServerContext) _fetchAndLoadDatabase(nonContextStruct base.NonCancella
...
+	if !registry.SGVersion.Less(&sc.BootstrapContext.sgVersion) {
+		return nil
+	}
+	registry.SGVersion = sc.BootstrapContext.sgVersion
+	err := sc.BootstrapContext.setGatewayRegistry(ctx, bucketName, registry)
+	if err != nil {
+		base.WarnfCtx(ctx, "Error setting gateway registry in bucket %q: %q", base.MD(bucketName), err)
+		return err
+	}
+	base.InfofCtx(ctx, base.KeyConfig, "Updated Sync Gateway version number in bucket %q from %s to %s", base.MD(bucketName), registry.SGVersion, sc.BootstrapContext.sgVersion)
```
adamcfraser: Looks like you need to make a copy of registry.SGVersion prior to line 1447 for logging purposes.

rest/config.go
```diff
@@ -1751,6 +1766,21 @@ func (sc *ServerContext) applyConfigs(ctx context.Context, dbNameConfigs map[str
 
 // _applyConfig loads the given database, failFast=true will not attempt to retry connecting/loading
 func (sc *ServerContext) _applyConfig(nonContextStruct base.NonCancellableContext, cnf DatabaseConfig, failFast, isInitialStartup bool) (applied bool, err error) {
+	ctx := nonContextStruct.Ctx
+
+	registry, err := sc.BootstrapContext.getGatewayRegistry(ctx, *cnf.Bucket)
```
adamcfraser: The idea behind the ConfigManager interface was to try to abstract/hide the underlying registry implementation from ServerContext.  I'm realizing now I already broke that contract in handleGetClusterInfo, but I think we'd benefit from keeping the _applyConfig code as simple as possible, and turn the two blocks of new code here into:
```
  ConfigManager.VerifyVersion(ctx, bucketName, currentVersion) error
  ConfigManager.SetVersion(ctx. bucketName, currentVersion)
```  
What do you think?

adamcfraser (changes requested):

base/version_comparable.go
torcolvin: I think that validation is good to do, I was pretty much preserving the behavior after removing the `sync.Once`

torcolvin: You can end in situations where these aren't the same "1.0.0.0" will resolve to "1.0.0" for example. I don't think this is likely in practice.

rest/config.go
torcolvin: I like this idea, but I named the functions a little differently.

base/version_comparable.go
```diff
@@ -28,18 +26,19 @@ type ComparableVersion struct {
...
+func zeroComparableVersion() *ComparableVersion {
+	return &ComparableVersion{
+		epoch:   0,
+		major:   0,
+		minor:   0,
+		patch:   0,
+		other:   0,
+		build:   0,
+		edition: "",
+	}
```
adamcfraser, line 41: Should this set `str`, on the off-chance it gets logged somewhere?

base/version_comparable.go
```diff
@@ -129,10 +135,21 @@ func (a *ComparableVersion) Less(b *ComparableVersion) bool {
 	return false
 }
 
-func (pv *ComparableVersion) String() string {
-	pv.strOnce.Do(func() {
-		pv.str = pv.formatComparableVersion()
-	})
+// AtLeastMinor returns true there is a major or minor downgrade.
```
adamcfraser:
```suggestion
// AtLeastMinor returns true there is a major or minor downgrade from b to a.
```

base/version_comparable.go
```diff
@@ -129,10 +135,21 @@ func (a *ComparableVersion) Less(b *ComparableVersion) bool {
...
+	if a.epoch > b.epoch {
+		return true
+	}
+	if a.major > b.major {
+		return true
+	}
+	if a.minor > b.minor {
+		return true
+	}
+	return false
```
adamcfraser: I think there's a bug here - this would return true for 3.2.0 -> 4.1.0, wouldn't it?

adamcfraser (changes requested):

base/version_comparable.go
torcolvin: Actually, this is a case where it would return `0.0.0` so I changed it to maintain compatibility since this can be used in the sharded import feed.

base/version_comparable.go
torcolvin: whoops

base/version_comparable.go
```diff
@@ -129,10 +137,25 @@ func (a *ComparableVersion) Less(b *ComparableVersion) bool {
 	return false
 }
 
-func (pv *ComparableVersion) String() string {
-	pv.strOnce.Do(func() {
-		pv.str = pv.formatComparableVersion()
-	})
+// AtLeastMinorDowngrade returns true there is a major or minor downgrade from a to b.
+func (a *ComparableVersion) AtLeastMinorDowngrade(b *ComparableVersion) bool {
```
adamcfraser, line 141: Possibly a nitpick, but I thought this would be easier to follow in the form:
```
if a.epoch != b.epoch {
    return a.epoch > b.epoch
}
etc
```
It feels a bit more transparent that if the values are equal, we fall through to the next check.  If you prefer the way it is I can live with that.

base/version_comparable_test.go
```diff
@@ -115,6 +115,115 @@ func TestInvalidComparableVersion(t *testing.T) {
...
+		},
+		{
+			versionA:       "2.2.0",
+			versionB:       "1.1.0",
+			minorDowngrade: true,
+		},
+		{
+			versionA:       "1.1.0",
+			versionB:       "2.2.0",
+			minorDowngrade: false,
```
adamcfraser, line 211: It looks like we still don't have a test case where a.major < b.major with a.minor > b.minor (e.g. 1.2.0 to 2.1.0), and vice versa (2.1.0 to 1.2.0).  Would be good to include for completeness.

adamcfraser (changes requested):

adamcfraser (dismissed):
LGTM

adamcfraser (approved):

PR #6490, opened 2023-10-03
[3.0.9] CBG-3458 don't load config if db registry is present

torcolvin:
A db registry will be present in > 3.0 Sync Gateway. Only added the check when persistent config is enabled, since this is primarily targeting capella and it simplifies how to get a temporary connection to a bucket before we load a database.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration-go1.16.15/78/

### DIFF
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -17,7 +17,10 @@ type BootstrapConnection interface {
 	GetConfigBuckets() ([]string, error)
 	// GetConfig fetches a database config for a given bucket and config group ID, along with the CAS of the config document.
 	GetConfig(bucket, groupID string, valuePtr interface{}) (cas uint64, err error)
+	// HasPost30Config returns true if the bucket has a database registry from 3.1 or greater.
+	HasPost30Config(bucket string) (bool, error)
 	// InsertConfig saves a new database config for a given bucket and config group ID.
+
 	InsertConfig(bucket, groupID string, value interface{}) (newCAS uint64, err error)
 	// UpdateConfig updates an existing database config for a given bucket and config group ID. updateCallback can return nil to remove the config.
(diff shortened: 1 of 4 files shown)


base/bootstrap.go
```diff
@@ -191,6 +194,24 @@ func (cc *CouchbaseCluster) GetConfig(location, groupID string, valuePtr interfa
...
+	b, teardown, err := cc.getBucket(bucket)
+
+	if err != nil {
+		return false, err
+	}
+
+	defer teardown()
+
+	var valuePtr interface{}
+	_, err = cc.configPersistence.loadConfig(b.DefaultCollection(), "_sync:registry", valuePtr)
```
adamcfraser: Would be helpful to define

```
	SGRegistryKey                        = SyncPrefix + "registry"          // Registry of all SG databases defined for the bucket (for all group IDs)
```
in base/constants.go.  If I'm looking for interaction with the registry in future I might look there first.

rest/config.go
```diff
@@ -1307,6 +1307,17 @@ func (sc *ServerContext) _applyConfig(cnf DatabaseConfig, failFast bool) (applie
 		}
 	}
 
+	if sc.bootstrapContext.connection != nil {
```
adamcfraser, line 1310: When is bootstrapContext.connection == nil?  Are we skipping this check when running with file-based config?

adamcfraser (changes requested):

rest/config.go
torcolvin: I decided to skip this in file based config because unlike in 3.1, we don't have a bootstrap connection in file based config, so I'd have to make and close a bucket specific connection. Since the primary consumer of this feature is capella, I thought that I'd avoid adding code complexity for a minor gain.

It is checked in 3.1 for file based config.

adamcfraser (approved):

adamcfraser:
Failing windows test is TestWebhookBasic - unrelated to these changes.

PR #6491, opened 2023-10-04
CBG-3209: Add cv index and retrieval for revision cache

gregns1:
CBG-3209

- Added CV to document revision struct
- Added loader function for CV that returns revid for puprose of populating revID lookup map 
- Added relevant interface definitions to add support for lookup by CV 
- Two ways to remove from cache, one from CV and another by revid, removes from both lookup maps when requesting to remove from cache
- Added getValueByCV which will retrieve value by CV, then if not found and create is true then will create this value and add to the cache, updating both revID lookup and CV lookup maps
- Altered upsert behaviour to work with CV too 
- Loader functiosn return revid/cv wherever appropriate for purpose fo populating opposite lookup map 
- Two functions `addToCVLookupPostRevLoad` `addToRevLookupPostCVLoad` to add to opposite lookup map 
- Chnaged a test that was running resync with xatrrs disabled. This caused my PR to panic as we assume xattrs are there as they aren't supported for this release. 

All in all the changes to add support for CV lookup, i've ensured that this is alongside alongside the RevID lookup in case we get legacy clients requesting for revs via the revID. Thus need both ways to lookup until backwards compatibility is removed. Likewise  if we communicate with a non HLV aware client and we receive a new doc/doc update over replicator we need a way to update rev cache with that rev supplied.  I have split some function out into separate functions to keep revID lookups and CV lookups separate. 

In addition in the loader code for CV `revCacheLoaderForDocumentCV` we need to have the concept of channel removal by CV for the checking if a doc was a channel removal.

Added tests to test the various operations on the cache and test that eviction also woks as expected.

`currentChannels` function implementation is taken from, Ben's PR for CBG-3203

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2130/ (known flake)

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -558,28 +558,6 @@ func (col *DatabaseCollectionWithUser) authorizeDoc(doc *Document, revid string)
 	}
 }
 
-func (c *DatabaseCollection) getCurrentVersion(ctx context.Context, doc *Document) (bodyBytes []byte, body Body, attachments AttachmentsMeta, err error) {
-	bodyBytes, err = doc.BodyBytes(ctx)
-	if err != nil {
-		base.WarnfCtx(ctx, "Marshal error when retrieving active current version body: %v", err)
-		return nil, nil, nil, err
-	}
-
-	body = doc._body
-	attachments = doc.Attachments
-
-	// handle backup revision inline attachments, or pre-2.5 meta
-	if inlineAtts, cleanBodyBytes, cleanBody, err := extractInlineAttachments(bodyBytes); err != nil {
-		return nil, nil, nil, err
-	} else if len(inlineAtts) > 0 {
-		// we found some inline attachments, so merge them with attachments, and update the bodies
-		attachments = mergeAttachments(inlineAtts, attachments)
-		bodyBytes = cleanBodyBytes
-		body = cleanBody
-	}
-	return bodyBytes, body, attachments, err
-}
-
 // Gets a revision of a document. If it's obsolete it will be loaded from the database if possible.
 // inline "_attachments" properties in the body will be extracted and returned separately if present (pre-2.5 metadata, or backup revisions)
 func (c *DatabaseCollection) getRevision(ctx context.Context, doc *Document, revid string) (bodyBytes []byte, body Body, attachments AttachmentsMeta, err error) {
@@ -2026,6 +2004,7 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 			}
 		} else if doc != nil {
 			doc.Cas = casOut
+			doc.HLV.Version = casOut
 		}
 	}
 
@@ -2083,7 +2062,7 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 			Expiry:           doc.Expiry,
 			Deleted:          doc.History[newRevID].Deleted,
 			_shallowCopyBody: storedDoc.Body(ctx),
-			CV:               &CurrentVersionVector{VersionCAS: doc.Cas, SourceID: doc.HLV.SourceID},
+			CV:               &CurrentVersionVector{VersionCAS: doc.HLV.Version, SourceID: doc.HLV.SourceID},
 		}
 
 		if createNewRevIDSkipped {
--- a/db/crud.go
+++ b/db/crud.go
@@ -2083,7 +2083,7 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 			Expiry:           doc.Expiry,
 			Deleted:          doc.History[newRevID].Deleted,
 			_shallowCopyBody: storedDoc.Body(ctx),
-			CV:               nil, // currently nil this as work to add update to HLV is on another branch
+			CV:               &CurrentVersionVector{VersionCAS: doc.Cas, SourceID: doc.HLV.SourceID},
 		}
 
 		if createNewRevIDSkipped {
--- a/db/crud.go
+++ b/db/crud.go
@@ -881,7 +881,7 @@ func (db *DatabaseCollectionWithUser) updateHLV(d *Document, docUpdateEvent DocU
 		d.HLV.CurrentVersionCAS = hlvExpandMacroCASValue
 	case Import:
 		// work to be done to decide if the VV needs updating here, pending CBG-3503
-	case NewVersion:
+	case NewVersion, InternalExistingVersion:
 		// add a new entry to the version vector
 		newVVEntry := CurrentVersionVector{}
 		newVVEntry.SourceID = db.dbCtx.BucketUUID
@@ -1045,8 +1045,8 @@ func (db *DatabaseCollectionWithUser) Put(ctx context.Context, docid string, bod
 }
 
 // Adds an existing revision to a document along with its history (list of rev IDs.)
-func (db *DatabaseCollectionWithUser) PutExistingRev(ctx context.Context, newDoc *Document, docHistory []string, noConflicts bool, forceAllConflicts bool, existingDoc *sgbucket.BucketDocument) (doc *Document, newRevID string, err error) {
-	return db.PutExistingRevWithConflictResolution(ctx, newDoc, docHistory, noConflicts, nil, forceAllConflicts, existingDoc)
+func (db *DatabaseCollectionWithUser) PutExistingRev(ctx context.Context, newDoc *Document, docHistory []string, noConflicts bool, forceAllConflicts bool, existingDoc *sgbucket.BucketDocument, docUpdateEvent DocUpdateType) (doc *Document, newRevID string, err error) {
+	return db.PutExistingRevWithConflictResolution(ctx, newDoc, docHistory, noConflicts, nil, forceAllConflicts, existingDoc, docUpdateEvent)
 }
 
 // PutExistingRevWithConflictResolution Adds an existing revision to a document along with its history (list of rev IDs.)
@@ -1054,14 +1054,14 @@ func (db *DatabaseCollectionWithUser) PutExistingRev(ctx context.Context, newDoc
 //  1. If noConflicts == false, the revision will be added to the rev tree as a conflict
 //  2. If noConflicts == true and a conflictResolverFunc is not provided, a 409 conflict error will be returned
 //  3. If noConflicts == true and a conflictResolverFunc is provided, conflicts will be resolved and the result added to the document.
-func (db *DatabaseCollectionWithUser) PutExistingRevWithConflictResolution(ctx context.Context, newDoc *Document, docHistory []string, noConflicts bool, conflictResolver *ConflictResolver, forceAllowConflictingTombstone bool, existingDoc *sgbucket.BucketDocument) (doc *Document, newRevID string, err error) {
+func (db *DatabaseCollectionWithUser) PutExistingRevWithConflictResolution(ctx context.Context, newDoc *Document, docHistory []string, noConflicts bool, conflictResolver *ConflictResolver, forceAllowConflictingTombstone bool, existingDoc *sgbucket.BucketDocument, docUpdateEvent DocUpdateType) (doc *Document, newRevID string, err error) {
 	newRev := docHistory[0]
 	generation, _ := ParseRevID(ctx, newRev)
 	if generation < 0 {
 		return nil, "", base.HTTPErrorf(http.StatusBadRequest, "Invalid revision ID")
 	}
 
-	docUpdateEvent := ExistingVersion
+	//docUpdateEvent := ExistingVersion
 	allowImport := db.UseXattrs()
 	doc, _, err = db.updateAndReturnDoc(ctx, newDoc.ID, allowImport, newDoc.DocExpiry, nil, docUpdateEvent, existingDoc, func(doc *Document) (resultDoc *Document, resultAttachmentData AttachmentData, createNewRevIDSkipped bool, updatedExpiry *uint32, resultErr error) {
 		// (Be careful: this block can be invoked multiple times if there are races!)
@@ -1162,7 +1162,7 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithConflictResolution(ctx c
 	return doc, newRev, err
 }
 
-func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context, docid string, body Body, docHistory []string, noConflicts bool) (doc *Document, newRev string, err error) {
+func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context, docid string, body Body, docHistory []string, noConflicts bool, docUpdateEvent DocUpdateType) (doc *Document, newRev string, err error) {
 	err = validateAPIDocUpdate(body)
 	if err != nil {
 		return nil, "", err
@@ -1187,7 +1187,7 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
 
 	newDoc.UpdateBody(body)
 
-	doc, newRevID, putExistingRevErr := db.PutExistingRev(ctx, newDoc, docHistory, noConflicts, false, nil)
+	doc, newRevID, putExistingRevErr := db.PutExistingRev(ctx, newDoc, docHistory, noConflicts, false, nil, docUpdateEvent)
 
 	if putExistingRevErr != nil {
 		return nil, "", putExistingRevErr
@@ -2004,7 +2004,8 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 			}
 		} else if doc != nil {
 			doc.Cas = casOut
-			doc.HLV.Version = casOut
+			// update the doc's HLV defined post macro expansion
+			doc = postWriteUpdateHLV(doc, casOut)
 		}
 	}
 
@@ -2123,6 +2124,27 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 	// Mark affected users/roles as needing to recompute their channel access:
 	db.MarkPrincipalsChanged(ctx, docid, newRevID, changedAccessPrincipals, changedRoleAccessUsers, doc.Sequence)
 	return doc, newRevID, nil
+}
+
+func postWriteUpdateHLV(doc *Document, casOut uint64) *Document {
+	if doc.HLV == nil {
+		return doc
+	}
+	if doc.HLV.Version == hlvExpandMacroCASValue {
+		doc.HLV.Version = casOut
+	}
+	if doc.HLV.CurrentVersionCAS == hlvExpandMacroCASValue {
+		doc.HLV.CurrentVersionCAS = casOut
+	}
+	// this is a temporary workaround. To be removed pending CBG-3503. This is to account for test that are relying on
+	// updates to arrive over import. As it stands the logic is still to be implemented on import doc event type leading
+	// to a doc when arriving over import, having outdated or no HLV defined, leading to issues with rev cache having
+	// appropriate doc updates
+	// TODO: Remove the below block when working on CBG-3503
+	if doc.Cas != doc.HLV.Version {
+		doc.HLV.Version = casOut
+	}
+	return doc
 }
 
 func getAttachmentIDsForLeafRevisions(ctx context.Context, db *DatabaseCollectionWithUser, doc *Document, newRevID string) (map[string]struct{}, error) {
--- a/db/crud.go
+++ b/db/crud.go
@@ -881,7 +881,7 @@ func (db *DatabaseCollectionWithUser) updateHLV(d *Document, docUpdateEvent DocU
 		d.HLV.CurrentVersionCAS = hlvExpandMacroCASValue
 	case Import:
 		// work to be done to decide if the VV needs updating here, pending CBG-3503
-	case NewVersion, InternalExistingVersion:
+	case NewVersion, ExistingVersionWithUpdateToHLV:
 		// add a new entry to the version vector
 		newVVEntry := CurrentVersionVector{}
 		newVVEntry.SourceID = db.dbCtx.BucketUUID
@@ -2134,14 +2134,6 @@ func postWriteUpdateHLV(doc *Document, casOut uint64) *Document {
 	}
 	if doc.HLV.CurrentVersionCAS == hlvExpandMacroCASValue {
 		doc.HLV.CurrentVersionCAS = casOut
-	}
-	// this is a temporary workaround. To be removed pending CBG-3503. This is to account for test that are relying on
-	// updates to arrive over import. As it stands the logic is still to be implemented on import doc event type leading
-	// to a doc when arriving over import, having outdated or no HLV defined, leading to issues with rev cache having
-	// appropriate doc updates
-	// TODO: Remove the below block when working on CBG-3503
-	if doc.Cas != doc.HLV.Version {
-		doc.HLV.Version = casOut
 	}
 	return doc
 }
--- a/db/crud.go
+++ b/db/crud.go
@@ -1061,7 +1061,6 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithConflictResolution(ctx c
 		return nil, "", base.HTTPErrorf(http.StatusBadRequest, "Invalid revision ID")
 	}
 
-	//docUpdateEvent := ExistingVersion
 	allowImport := db.UseXattrs()
 	doc, _, err = db.updateAndReturnDoc(ctx, newDoc.ID, allowImport, newDoc.DocExpiry, nil, docUpdateEvent, existingDoc, func(doc *Document) (resultDoc *Document, resultAttachmentData AttachmentData, createNewRevIDSkipped bool, updatedExpiry *uint32, resultErr error) {
 		// (Be careful: this block can be invoked multiple times if there are races!)
--- a/db/crud.go
+++ b/db/crud.go
@@ -317,7 +317,7 @@ func (db *DatabaseCollectionWithUser) getRev(ctx context.Context, docid, revid s
 	if revid != "" {
 		// Get a specific revision body and history from the revision cache
 		// (which will load them if necessary, by calling revCacheLoader, above)
-		revision, err = db.revisionCache.Get(ctx, docid, revid, nil, includeBody, RevCacheOmitDelta)
+		revision, err = db.revisionCache.GetWithRev(ctx, docid, revid, includeBody, RevCacheOmitDelta)
 	} else {
 		// No rev ID given, so load active revision
 		revision, err = db.revisionCache.GetActive(ctx, docid, includeBody)
@@ -381,7 +381,7 @@ func (db *DatabaseCollectionWithUser) GetDelta(ctx context.Context, docID, fromR
 		return nil, nil, nil
 	}
 
-	fromRevision, err := db.revisionCache.Get(ctx, docID, fromRevID, nil, RevCacheOmitBody, RevCacheIncludeDelta)
+	fromRevision, err := db.revisionCache.GetWithRev(ctx, docID, fromRevID, RevCacheOmitBody, RevCacheIncludeDelta)
 
 	// If the fromRevision is a removal cache entry (no body), but the user has access to that removal, then just
 	// return 404 missing to indicate that the body of the revision is no longer available.
@@ -424,7 +424,7 @@ func (db *DatabaseCollectionWithUser) GetDelta(ctx context.Context, docID, fromR
 
 		// db.DbStats.StatsDeltaSync().Add(base.StatKeyDeltaCacheMisses, 1)
 		db.dbStats().DeltaSync().DeltaCacheMiss.Add(1)
-		toRevision, err := db.revisionCache.Get(ctx, docID, toRevID, nil, RevCacheOmitBody, RevCacheIncludeDelta)
+		toRevision, err := db.revisionCache.GetWithRev(ctx, docID, toRevID, RevCacheOmitBody, RevCacheIncludeDelta)
 		if err != nil {
 			return nil, nil, err
 		}
@@ -442,7 +442,7 @@ func (db *DatabaseCollectionWithUser) GetDelta(ctx context.Context, docID, fromR
 		// If the revision we're generating a delta to is a tombstone, mark it as such and don't bother generating a delta
 		if deleted {
 			revCacheDelta := newRevCacheDelta([]byte(base.EmptyDocument), fromRevID, toRevision, deleted, nil)
-			db.revisionCache.UpdateDelta(ctx, docID, fromRevID, nil, revCacheDelta)
+			db.revisionCache.UpdateDelta(ctx, docID, fromRevID, revCacheDelta)
 			return &revCacheDelta, nil, nil
 		}
 
@@ -482,7 +482,7 @@ func (db *DatabaseCollectionWithUser) GetDelta(ctx context.Context, docID, fromR
 		revCacheDelta := newRevCacheDelta(deltaBytes, fromRevID, toRevision, deleted, toRevAttStorageMeta)
 
 		// Write the newly calculated delta back into the cache before returning
-		db.revisionCache.UpdateDelta(ctx, docID, fromRevID, nil, revCacheDelta)
+		db.revisionCache.UpdateDelta(ctx, docID, fromRevID, revCacheDelta)
 		return &revCacheDelta, nil, nil
 	}
 
@@ -561,7 +561,7 @@ func (col *DatabaseCollectionWithUser) authorizeDoc(doc *Document, revid string)
 func (c *DatabaseCollection) getCurrentVersion(ctx context.Context, doc *Document) (bodyBytes []byte, body Body, attachments AttachmentsMeta, err error) {
 	bodyBytes, err = doc.BodyBytes(ctx)
 	if err != nil {
-		base.WarnfCtx(ctx, "Marshal error when retrieving active current revision body: %v", err)
+		base.WarnfCtx(ctx, "Marshal error when retrieving active current version body: %v", err)
 		return nil, nil, nil, err
 	}
 
@@ -2012,7 +2012,7 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 
 			// Prior to saving doc, remove the revision in cache
 			if createNewRevIDSkipped {
-				db.revisionCache.Remove(doc.ID, doc.CurrentRev, nil)
+				db.revisionCache.RemoveWithRev(doc.ID, doc.CurrentRev)
 			}
 
 			base.DebugfCtx(ctx, base.KeyCRUD, "Saving doc (seq: #%d, id: %v rev: %v)", doc.Sequence, base.UD(doc.ID), doc.CurrentRev)
--- a/db/crud.go
+++ b/db/crud.go
@@ -317,7 +317,7 @@ func (db *DatabaseCollectionWithUser) getRev(ctx context.Context, docid, revid s
 	if revid != "" {
 		// Get a specific revision body and history from the revision cache
 		// (which will load them if necessary, by calling revCacheLoader, above)
-		revision, err = db.revisionCache.Get(ctx, docid, revid, includeBody, RevCacheOmitDelta)
+		revision, err = db.revisionCache.Get(ctx, docid, revid, nil, includeBody, RevCacheOmitDelta)
 	} else {
 		// No rev ID given, so load active revision
 		revision, err = db.revisionCache.GetActive(ctx, docid, includeBody)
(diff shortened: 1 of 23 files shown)


db/revision_cache_interface.go
```diff
@@ -31,14 +31,14 @@ type RevisionCache interface {
 	// Get returns the given revision, and stores if not already cached.
 	// When includeBody=true, the returned DocumentRevision will include a mutable shallow copy of the marshaled body.
 	// When includeDelta=true, the returned DocumentRevision will include delta - requires additional locking during retrieval.
-	Get(ctx context.Context, docID, revID string, includeBody bool, includeDelta bool) (DocumentRevision, error)
+	Get(ctx context.Context, docID, revID string, cv *CurrentVersionVector, includeBody, includeDelta bool) (DocumentRevision, error)
```
adamcfraser: I think it would be preferable to define separate APIs for rev- and version-based retrieval from the revision cache, for two reasons:
- it avoids issues around priority (or error handling) if a caller populates both parameters
- it makes it clearer to identify whether the calling code is working with revisions or versions (which will simplify removal in future)

db/revision_cache_interface.go
```diff
@@ -47,10 +47,10 @@ type RevisionCache interface {
 	Upsert(ctx context.Context, docRev DocumentRevision)
 
 	// Remove eliminates a revision in the cache.
-	Remove(docID, revID string)
+	Remove(docID, revID string, cv *CurrentVersionVector)
 
 	// UpdateDelta stores the given toDelta value in the given rev if cached
-	UpdateDelta(ctx context.Context, docID, revID string, toDelta RevisionDelta)
+	UpdateDelta(ctx context.Context, docID, revID string, cv *CurrentVersionVector, toDelta RevisionDelta)
```
adamcfraser: I didn't think we should be modifying this API until we support version-based Delta (which would include the required changes to RevisionDelta in this file).

db/revision_cache_interface.go
```diff
@@ -47,10 +47,10 @@ type RevisionCache interface {
 	Upsert(ctx context.Context, docRev DocumentRevision)
 
 	// Remove eliminates a revision in the cache.
-	Remove(docID, revID string)
+	Remove(docID, revID string, cv *CurrentVersionVector)
```
adamcfraser: Similar to Get - I think a version-specific API is preferred.

db/revision_cache_interface.go
```diff
@@ -31,14 +31,14 @@ type RevisionCache interface {
...
-	Get(ctx context.Context, docID, revID string, includeBody bool, includeDelta bool) (DocumentRevision, error)
+	Get(ctx context.Context, docID, revID string, cv *CurrentVersionVector, includeBody, includeDelta bool) (DocumentRevision, error)
 
 	// GetActive returns the current revision for the given doc ID, and stores if not already cached.
 	// When includeBody=true, the returned DocumentRevision will include a mutable shallow copy of the marshaled body.
 	GetActive(ctx context.Context, docID string, includeBody bool) (docRev DocumentRevision, err error)
 
 	// Peek returns the given revision if present in the cache
-	Peek(ctx context.Context, docID, revID string) (docRev DocumentRevision, found bool)
+	Peek(ctx context.Context, docID, revID string, cv *CurrentVersionVector) (docRev DocumentRevision, found bool)
```
adamcfraser: Do we need a version-aware Peek?  This seems like it should wait until (if) we decide to implement a version-aware version of backupPreImportRevision.

db/crud.go
```diff
@@ -558,6 +558,28 @@ func (col *DatabaseCollectionWithUser) authorizeDoc(doc *Document, revid string)
 	}
 }
 
+func (c *DatabaseCollection) getCurrentVersion(ctx context.Context, doc *Document) (bodyBytes []byte, body Body, attachments AttachmentsMeta, err error) {
+	bodyBytes, err = doc.BodyBytes(ctx)
+	if err != nil {
+		base.WarnfCtx(ctx, "Marshal error when retrieving active current revision body: %v", err)
```
```suggestion
		base.WarnfCtx(ctx, "Marshal error when retrieving active current version body: %v", err)
```

db/revision_cache_interface.go
```diff
@@ -259,6 +267,22 @@ func revCacheLoader(ctx context.Context, backingStore RevisionCacheBackingStore,
 	return revCacheLoaderForDocument(ctx, backingStore, doc, id.RevID)
 }
 
+func revCacheLoaderForCv(ctx context.Context, backingStore RevisionCacheBackingStore, id IDandCV, unmarshalBody bool) (bodyBytes []byte, body Body, history Revisions, channels base.Set, removed bool, attachments AttachmentsMeta, deleted bool, expiry *time.Time, err error) {
```
adamcfraser: Where are we checking that the current version of the document that we get from backingStore matches cv?

adamcfraser: A unit test for this case would be a good idea, as this seems like an easy place for subtle bugs to get introduced.

db/revision_cache_lru.go
```diff
@@ -234,12 +255,28 @@ func (rc *LRURevisionCache) Put(ctx context.Context, docRev DocumentRevision) {
...
 
 // Upsert a revision in the cache.
 func (rc *LRURevisionCache) Upsert(ctx context.Context, docRev DocumentRevision) {
+	// update the cv cache, then the revid cache too for backwards compatability
+	if docRev.CV != nil {
+		rc.upsertHLVCache(ctx, docRev)
+		// we have also updated the revID lookup cache too so return early here
+		return
+	}
+	rc.upsertRevIDCache(ctx, docRev)
```
adamcfraser: Can we stick to a single Upsert, and have the conditional handling around docRev.CV==nil inside the Upsert?

db/revision_cache_lru.go
```diff
@@ -274,7 +334,33 @@ func (rc *LRURevisionCache) getValue(docID, revID string, create bool) (value *r
...
+	}
+
+	key := IDandCV{DocID: docID, Source: cv.SourceID, Version: cv.VersionCAS}
+	rc.lock.Lock()
+	if elem := rc.hlvCache[key]; elem != nil {
+		rc.lruList.MoveToFront(elem)
+		value = elem.Value.(*revCacheValue)
+	} else if create {
+		// we need to keep this new element in the CV lookup cache and the revid cache for backwards compatability reasons
+		legacyKey := IDAndRev{DocID: docID, RevID: revID}
```
adamcfraser: This looks like it will always require callers to know both cv and revID, to ensure that both maps are properly updated.  This wouldn't typically be the case for callers of getFromCache that end up here, I wouldn't think. Let me know if I'm overlooking something, or if you've got a unit test that successfully loads the cache using a cv-based Get, and then fetches from the cache using rev.  (or vice versa)

We probably need to discuss this handling in a bit of depth.

adamcfraser (changes requested):

db/revision_cache_lru.go
gregns1: I agree makes sense to me. Good spot that was an oversight from me.

db/revision_cache_interface.go
gregns1: Good point I've added this check and appropriate error message to be returned in that fail case. Also added a unit test to cover this scenario entitled `TestLoaderMismatchInCV`

db/crud.go
```diff
@@ -2083,7 +2083,7 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 			Expiry:           doc.Expiry,
 			Deleted:          doc.History[newRevID].Deleted,
 			_shallowCopyBody: storedDoc.Body(ctx),
-			CV:               nil, // currently nil this as work to add update to HLV is on another branch
+			CV:               &CurrentVersionVector{VersionCAS: doc.Cas, SourceID: doc.HLV.SourceID},
```
adamcfraser: Shouldn't we use doc.HLV.Version instead of doc.Cas here?

db/revision_cache_bypass.go
```diff
@@ -45,10 +46,12 @@ func (rc *BypassRevisionCache) GetWithRev(ctx context.Context, docID, revID stri
 	docRev = DocumentRevision{
 		RevID: revID,
 	}
-	docRev.BodyBytes, docRev._shallowCopyBody, docRev.History, docRev.Channels, docRev.Removed, docRev.Attachments, docRev.Deleted, docRev.Expiry, err = revCacheLoaderForDocument(ctx, rc.backingStore, doc, revID)
+	docRev.BodyBytes, docRev._shallowCopyBody, docRev.History, docRev.Channels, docRev.Removed, docRev.Attachments, docRev.Deleted, docRev.Expiry, fetchedCV, err = revCacheLoaderForDocument(ctx, rc.backingStore, doc, revID)
 	if err != nil {
 		return DocumentRevision{}, err
 	}
+	// populate doc rev with the CV
+	docRev.CV = &fetchedCV
```
adamcfraser: If this is being done on a separate line just to convert to pointer, should revCacheLoaderForDocument return *CurrentVersionVector instead?

db/revision_cache_bypass.go
```diff
@@ -62,14 +65,20 @@ func (rc *BypassRevisionCache) GetWithCV(ctx context.Context, docID string, cv *
...
 	if err != nil {
 		return DocumentRevision{}, err
 	}
-	docRev.BodyBytes, docRev._shallowCopyBody, docRev.History, docRev.Channels, docRev.Removed, docRev.Attachments, docRev.Deleted, docRev.Expiry, err = revCacheLoaderForDocumentCV(ctx, rc.backingStore, doc, *cv)
+
+	docRev.BodyBytes, docRev._shallowCopyBody, docRev.History, docRev.Channels, docRev.Removed, docRev.Attachments, docRev.Deleted, docRev.Expiry, _, err = revCacheLoaderForDocumentCV(ctx, rc.backingStore, doc, *cv)
 	if err != nil {
 		return DocumentRevision{}, err
 	}
+	docRev.RevID = doc.RevID
```
adamcfraser: Can we just set this on line 77 with the returned revID?

db/revision_cache_bypass.go
```diff
@@ -92,10 +102,12 @@ func (rc *BypassRevisionCache) GetActive(ctx context.Context, docID string, incl
 		RevID: doc.CurrentRev,
 	}
 
-	docRev.BodyBytes, docRev._shallowCopyBody, docRev.History, docRev.Channels, docRev.Removed, docRev.Attachments, docRev.Deleted, docRev.Expiry, err = revCacheLoaderForDocument(ctx, rc.backingStore, doc, doc.SyncData.CurrentRev)
+	docRev.BodyBytes, docRev._shallowCopyBody, docRev.History, docRev.Channels, docRev.Removed, docRev.Attachments, docRev.Deleted, docRev.Expiry, fetchedCV, err = revCacheLoaderForDocument(ctx, rc.backingStore, doc, doc.SyncData.CurrentRev)
```
adamcfraser: As mentioned elsewhere, having revCacheLoaderForDocument return *CurrentVersionVector simplifies this code.

db/revision_cache_interface.go
```diff
@@ -46,8 +51,11 @@ type RevisionCache interface {
 	// Update will remove existing value and re-create new one
 	Upsert(ctx context.Context, docRev DocumentRevision)
 
-	// Remove eliminates a revision in the cache.
-	Remove(docID, revID string)
+	// RemoveWithRev eliminates a revision in the cache using its revID.
```
adamcfraser:
```suggestion
	// RemoveWithRev evicts a revision from the cache using its revID.
```

db/revision_cache_interface.go
```diff
@@ -46,8 +51,11 @@ type RevisionCache interface {
...
+	RemoveWithRev(docID, revID string)
+
+	// RemoveWithCV eliminates a revision in the cache using its current version.
```
adamcfraser:
```suggestion
	// RemoveWithCV evicts a revision from the cache using its current version.
```

db/revision_cache_interface.go
```diff
@@ -28,10 +28,15 @@ const (
...
 // RevisionCache is an interface that can be used to fetch a DocumentRevision for a Doc ID and Rev ID pair.
 type RevisionCache interface {
-	// Get returns the given revision, and stores if not already cached.
+	// GetWithRev returns the given revision, and stores if not already cached.
 	// When includeBody=true, the returned DocumentRevision will include a mutable shallow copy of the marshaled body.
 	// When includeDelta=true, the returned DocumentRevision will include delta - requires additional locking during retrieval.
-	Get(ctx context.Context, docID, revID string, includeBody bool, includeDelta bool) (DocumentRevision, error)
+	GetWithRev(ctx context.Context, docID, revID string, includeBody, includeDelta bool) (DocumentRevision, error)
+
+	// GetWithCV returns the given revision by the CV, and stores if not already cached.
```
adamcfraser:
```suggestion
	// GetWithCV returns the given revision by CV, and stores if not already cached.
```

db/revision_cache_interface.go
```diff
@@ -246,44 +262,88 @@ func newRevCacheDelta(deltaBytes []byte, fromRevID string, toRevision DocumentRe
...
+	if doc.HLV != nil {
+		fetchedCV = CurrentVersionVector{SourceID: doc.HLV.SourceID, VersionCAS: doc.HLV.Version}
+	}
+
+	return bodyBytes, body, history, channels, removed, attachments, deleted, doc.Expiry, fetchedCV, err
+}
+
+// revCacheLoaderForDocumentCV used either during cache miss (from revCacheLoaderForCv), or used directly when getting current active CV from cache
+func revCacheLoaderForDocumentCV(ctx context.Context, backingStore RevisionCacheBackingStore, doc *Document, cv CurrentVersionVector) (bodyBytes []byte, body Body, history Revisions, channels base.Set, removed bool, attachments AttachmentsMeta, deleted bool, expiry *time.Time, revid string, err error) {
+	if bodyBytes, body, attachments, err = backingStore.getCurrentVersion(ctx, doc); err != nil {
```
adamcfraser, line 331: revCacheLoaderForDocumentCV needs to validate that the current version on doc matches the requested version (cv).  The revID version of this (revCacheLoaderForDocument) is doing this in backingStore.getRevision, but we'll want to do this here.

db/crud.go
```diff
@@ -558,6 +558,28 @@ func (col *DatabaseCollectionWithUser) authorizeDoc(doc *Document, revid string)
 	}
 }
 
+func (c *DatabaseCollection) getCurrentVersion(ctx context.Context, doc *Document) (bodyBytes []byte, body Body, attachments AttachmentsMeta, err error) {
```
adamcfraser: This doesn't look like it needs to be defined on a data store in the same way getRevision does, because we're not making any fallback calls to the datastore to look for things like temporary revision backups.  Until we have a reason to do that, I think we should just defined getCurrentVersion inside revision_cache_interface.

db/database_test.go
```diff
@@ -2849,72 +2849,62 @@ func Test_invalidateAllPrincipalsCache(t *testing.T) {
 }
 
 func Test_resyncDocument(t *testing.T) {
-	testCases := []struct {
-		useXattr bool
-	}{
-		{useXattr: true},
-		{useXattr: false},
+	if !base.TestUseXattrs() {
```
adamcfraser, line 2852: Why did the structure of this test need to be changed in this PR?

db/document.go
```diff
@@ -1242,3 +1242,14 @@ func (doc *Document) MarshalWithXattr() (data []byte, xdata []byte, err error) {
 
 	return data, xdata, nil
 }
+
+// Returns a set of the current (winning revision's) channels for the document.
+func (doc *Document) currentChannels() base.Set {
```
adamcfraser: Is there potential reuse of the work @bbrks is doing for CBG-3203 here?  (computing active channels at Document unmarshal time?)

db/revision_cache_interface.go
```diff
@@ -246,44 +262,88 @@ func newRevCacheDelta(deltaBytes []byte, fromRevID string, toRevision DocumentRe
...
+	}
+	var doc *Document
+	unmarshalLevel := DocUnmarshalSync
+	if unmarshalBody {
+		unmarshalLevel = DocUnmarshalAll
+	}
+	if doc, err = backingStore.GetDocument(ctx, id.DocID, unmarshalLevel); doc == nil {
+		return bodyBytes, body, history, channels, removed, attachments, deleted, expiry, revid, err
+	}
+	// nil check to be panic safe
```
adamcfraser:
```suggestion
```

db/revision_cache_interface.go
```diff
@@ -246,44 +262,88 @@ func newRevCacheDelta(deltaBytes []byte, fromRevID string, toRevision DocumentRe
...
+	if doc.HLV != nil {
+		// fetch the current version for the loaded doc and compare against the CV specified in the IDandCV key
+		fetchedDocSource, fetchedDocVersion := doc.HLV.GetCurrentVersion()
+		if fetchedDocSource != cv.SourceID || fetchedDocVersion != cv.VersionCAS {
+			return bodyBytes, body, history, channels, removed, attachments, deleted, expiry, revid, base.RedactErrorf("mismatch between specified current version and fetched document current version for doc %s", base.UD(id.DocID))
+		}
+	} else {
+		return bodyBytes, body, history, channels, removed, attachments, deleted, expiry, revid, base.RedactErrorf("mismatch between specified current version and fetched document current version for doc %s", base.UD(id.DocID))
```
adamcfraser: Duplicating this error message (including the string) seems a bit messt.  I might add a helper function like doc.HasCurrentVersion(src, version) that could include the nil check as well as the current version check to simplify this code.

db/revision_cache_interface.go
adamcfraser: I see now that you've got this logic in revCacheLoaderForCv.  I think you'll want to move it down here to cover other consumers of revCacheLoaderForDocumentCv (like BypassRevisionCache).

db/revision_cache_lru.go
```diff
@@ -140,18 +157,38 @@ func (rc *LRURevisionCache) UpdateDelta(ctx context.Context, docID, revID string
 	}
 }
 
-func (rc *LRURevisionCache) getFromCache(ctx context.Context, docID, revID string, loadOnCacheMiss bool, includeBody bool, includeDelta bool) (DocumentRevision, error) {
-	value := rc.getValue(docID, revID, loadOnCacheMiss)
-	if value == nil {
-		return DocumentRevision{}, nil
+func (rc *LRURevisionCache) getFromCache(ctx context.Context, docID, revID string, cv *CurrentVersionVector, loadOnCacheMiss, includeBody, includeDelta bool) (DocumentRevision, error) {
```
adamcfraser: It doesn't seem to me like there's enough common code that we shouldn't just define getFromCacheByRev and getFromCacheByCV to reduce the code complexity of each - what do you think?

db/revision_cache_lru.go
```diff
@@ -389,9 +622,15 @@ func (value *revCacheValue) updateBody(ctx context.Context) (err error) {
 // asDocumentRevision copies the rev cache value into a DocumentRevision.  Should only be called for non-empty
 // revCacheValues - copies all immutable revCacheValue properties, and adds the provided body/delta.
 func (value *revCacheValue) asDocumentRevision(body Body, delta *RevisionDelta) (DocumentRevision, error) {
+	var docID string
+	if value.cvLoad {
```
adamcfraser: Instead of using cvLoad here, can we use value.key if it's non-empty, and then fall back to value.hlvKey?

db/revision_cache_lru.go
```diff
@@ -354,7 +579,15 @@ func (value *revCacheValue) load(ctx context.Context, backingStore RevisionCache
 		}
 	} else {
 		cacheHit = false
-		value.bodyBytes, value.body, value.history, value.channels, value.removed, value.attachments, value.deleted, value.expiry, value.err = revCacheLoader(ctx, backingStore, value.key, includeBody)
+		if value.cvLoad {
```
adamcfraser: Can we make this decision based on the presence of value.key instead of cvLoad?

db/revision_cache_lru.go
```diff
@@ -86,25 +95,29 @@ type LRURevisionCache struct {
...
-	removed     bool
+	err                 error
+	history             Revisions
+	channels            base.Set
+	expiry              *time.Time
+	attachments         AttachmentsMeta
+	delta               *RevisionDelta
+	body                Body
+	key                 IDAndRev
+	hlvKey              IDandCV
```
adamcfraser: This looks like we're storing docID twice on the revCacheValue.  (IDAndRev and IDandCV).  We should avoid this to conserve cache space.  I think IDAndRev/IDAndCV can still be used as keys in the cache maps, but stored as ID/Rev/CV in the revCacheValue?

db/revision_cache_lru.go
```diff
@@ -86,25 +95,29 @@ type LRURevisionCache struct {
...
+	cvLoad              bool
```
adamcfraser: since cvLoad is just a load-time artifact, I don't think we should store permanently in the cache.  Made comments elsewhere about replacing this with checks for presence/absence of rev and cv.

db/revision_cache_lru.go
```diff
@@ -86,25 +95,29 @@ type LRURevisionCache struct {
...
+	bodyBytes           []byte
+	lock                sync.RWMutex
+	deleted             bool
+	removed             bool
+	presentInBothLookup bool
```
adamcfraser: Same comment as cvLoad - can we infer this on the presence/absence of rev/cv instead?

db/revision_cache_lru.go
```diff
@@ -140,18 +157,38 @@ func (rc *LRURevisionCache) UpdateDelta(ctx context.Context, docID, revID string
...
 	docRev, statEvent, err := value.load(ctx, rc.backingStore, includeBody, includeDelta)
 	rc.statsRecorderFunc(statEvent)
 
 	if err != nil {
 		rc.removeValue(value) // don't keep failed loads in the cache
+		return DocumentRevision{}, err
 	}
+
+	// if cache miss we need to ensure both lookup maps are updated
+	if !statEvent {
```
adamcfraser: Can we rename statEvent to cacheHit on line 175?  It would make this code more self-documenting, and I don't think 'statEvent' was a particularly good name in the first place.

db/revision_cache_lru.go
```diff
@@ -274,40 +327,210 @@ func (rc *LRURevisionCache) getValue(docID, revID string, create bool) (value *r
...
+}
+
+// addToRevLookupPostCVLoad will generate and entry in the Rev lookup map for a new document entering the cache
+func (rc *LRURevisionCache) addToRevLookupPostCVLoad(docID, revID string, cv *CurrentVersionVector) {
+	legacyKey := IDAndRev{DocID: docID, RevID: revID}
+	key := IDandCV{DocID: docID, Source: cv.SourceID, Version: cv.VersionCAS}
+
+	rc.lock.Lock()
+	defer rc.lock.Unlock()
+	// check for duplicate value in rev cache from concurrent fetch by rev ID at the same time
```
```suggestion
	// check for existing value in rev cache map (due to concurrent fetch by rev ID)
```

db/revision_cache_lru.go
```diff
@@ -274,40 +327,210 @@ func (rc *LRURevisionCache) getValue(docID, revID string, create bool) (value *r
...
+		//	 - Process A continues on and updates teh rev map now to point to its own value, process A then release the lock
+		//	 - Process B has been waiting on the lock, now its free it acquires the rev cache lock
+		//	 - Process B looks  up key in cv lookup map, find and entry, assumes duplicate and removes the entry from teh cv lookup map,
+		//	 and removes the element from teh cache list
+		//	 - We now find ourselves in a situation where both processes have removed both respective cache loads from the cache itself
+		//	 thus no entry in cache for thus doc revision/version anymore. This needs to be avoided.
+		//	 - SOLUTION: Added a flag tyo the rev cache value struct to indicate if this particular value has been successfully updated in BOTH
+		//	 - lookup tables and thus doesn't need to be updated again. This avoids duplicate updates to both lookup maps AND avoids removing both
+		//	 rev cache values from the rev cache itself.
+		elem.Value.(*revCacheValue).lock.RLock()
```
adamcfraser: We should be trying to avoid acquiring the value lock during this processing.  We're doing a lookup in both maps in these functions anyway (lines 369 and 399 here) while holding the cache lock.  Can we address the scenario above by always retaining the entry referenced by the CV map when there are duplicate entries (both here and in addToCVLookupPostRevLoad)?  I'm thinking something like this here:
```
cvElem, cvFound := rc.hlvCache[key]
revElem, revFound := rc.cache[legacyKey]
if !cvFound {
   // value has been evicted, exit early
}
if revFound {
   if cvElem==revElem {
       // already match, return
    }
    // If both exist (concurrent load), retain the element referenced by cv map
    rc.cache[legacyKey] = cvElem
    rc.lruList.Remove(revElem)
}
```
and then this in addToCVLookupPostRevLoad:

```
cvElem, cvFound := rc.hlvCache[key]
revElem, revFound := rc.cache[legacyKey]
if !revFound {
   // value has been evicted, exit early
}
if cvFound {
   if cvElem==revElem {
       // already match, return
    }
    // If both exist (concurrent load), retain the element referenced by cv map
    rc.cache[legacyKey] = cvElem
    rc.lruList.Remove(revElem). 
}

I might be overlooking a subtlety of the case you described in the comment.  If the above approach doesn't work, let me know what I'm overlooking.

File: db/revision_cache_lru.go [hunk 31]
```diff
@@ -274,40 +327,210 @@ func (rc *LRURevisionCache) getValue(docID, revID string, create bool) (value *r
...
+	// grab the revid key from the value to enable us to remove the reference from the rev lookup map too
+	legacyKey := element.Value.(*revCacheValue).key
+	rc.lruList.Remove(element)
+	delete(rc.hlvCache, key)
+	// remove from rev lookup map too
+	delete(rc.cache, legacyKey)
+}
+
+// removeFromCacheIDRev removes an entry from rev cache by revID
+func (rc *LRURevisionCache) removeFromCacheIDRev(docID, revID string) {
```
```suggestion
func (rc *LRURevisionCache) removeFromCacheByRev(docID, revID string) {
```

db/revision_cache_lru.go
```diff
@@ -274,40 +327,210 @@ func (rc *LRURevisionCache) getValue(docID, revID string, create bool) (value *r
...
+	rc.removeFromCacheIDRev(docID, revID)
+}
+
+// RemoveWithCV removes a value from rev cache by CV reference if present
+func (rc *LRURevisionCache) RemoveWithCV(docID string, cv *CurrentVersionVector) {
+	rc.removeFromCacheCV(docID, cv)
+}
+
+// removeFromCacheCV removes an entry from rev cache by CV
+func (rc *LRURevisionCache) removeFromCacheCV(docID string, cv *CurrentVersionVector) {
```
```suggestion
func (rc *LRURevisionCache) removeFromCacheByCV(docID string, cv *CurrentVersionVector) {
```

db/revision_cache_lru.go
```diff
@@ -274,40 +327,210 @@ func (rc *LRURevisionCache) getValue(docID, revID string, create bool) (value *r
...
+		for rc.lruList.Len() > int(rc.capacity) {
+			rc.purgeOldest_()
+		}
+	}
+	// its possible the element has been evicted if we don't find the element above (high churn on rev cache)
+	// need to return doc revision to caller still but no need repopulate the cache
+}
+
+// addToCVLookupPostRevLoad will generate and entry in the CV lookup map for a new document entering the cache
+func (rc *LRURevisionCache) addToCVLookupPostRevLoad(docID, revID string, cv *CurrentVersionVector) {
```
```suggestion
func (rc *LRURevisionCache) addToHLVMapPostLoad(docID, revID string, cv *CurrentVersionVector) {
```

db/revision_cache_lru.go
```diff
@@ -274,40 +327,210 @@ func (rc *LRURevisionCache) getValue(docID, revID string, create bool) (value *r
...
+		for rc.lruList.Len() > int(rc.capacity) {
+			rc.purgeOldest_()
+		}
+	}
+	rc.lock.Unlock()
+	return
+}
+
+// addToRevLookupPostCVLoad will generate and entry in the Rev lookup map for a new document entering the cache
+func (rc *LRURevisionCache) addToRevLookupPostCVLoad(docID, revID string, cv *CurrentVersionVector) {
```
```suggestion
func (rc *LRURevisionCache) addToRevMapPostLoad(docID, revID string, cv *CurrentVersionVector) {
```

adamcfraser (changes requested):
Some suggestions, mainly around:
- minimizing the information stored in the cache
- avoiding value locks during post-load map updates
- style/naming suggestions for code simplification/readability
Have a look and let me know what you think.

db/crud.go
gregns1: So the reason I used CAS here is becuse Version at this point was just the max uint64 value as the cas was being mutated into it. I realise now I can just set the version to the the casOut where we do so for the doc.Cas on line 2028, so have added that now!

db/revision_cache_lru.go
gregns1: Makes sense and seems to work too!

db/database_test.go
gregns1: This has been exposed as an issue here as i've added CurrentVersion to the docRev inside updateAndReturnDoc. Given the HLV is only updated/initialised in xattrs mode (given we arent supporting inline xattrs for VV release IIRC?) -> https://github.com/couchbase/sync_gateway/blob/4d13bab788884ba6a63beb7bd7746f11b688a302/db/crud.go#L1966C20-L1966C20 So when trying to add the CV to doc rev here -> https://github.com/couchbase/sync_gateway/blob/4d13bab788884ba6a63beb7bd7746f11b688a302/db/crud.go#L2054C4-L2054C4 the test panics as its not defined. Adding a nil check here doesn't protect as the rev cache then panics on trying to update the cache as it expects a documnet to have a HLV defined. 

I originally thought given for the release we aren't supporting the inline xattrs I removed the test case all together on this branch. But I could be wrong to do this?

db/crud.go
```diff
@@ -2004,6 +2004,7 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 			}
 		} else if doc != nil {
 			doc.Cas = casOut
+			doc.HLV.Version = casOut
```
adamcfraser: Good catch that we need to populate casOut on the doc for the in-memory doc for all properties that are doing cas macro expansion.  However, updateAndReturnDoc will be used for updates that don't set Version to CAS, so I don't think it's safe to set version in this way.  I think you want a helper function similar to db.updateHLV that takes casOut and the DocUpdateType, and uses that to populate the properties that were using macro expansion.  

It might be even better to have updateHLV return a list of the properties that were marked for macro expansion, to minimize bugs when updateHLV is changed in future, but I'd be ok with just adding something like postWriteUpdateHLV(*Document, DocUpdateType, casOut) to do this work for now.

db/revision_cache_lru.go
```diff
@@ -272,42 +329,212 @@ func (rc *LRURevisionCache) getValue(docID, revID string, create bool) (value *r
...
+		if cvElem == revElem {
+			// already match, return
+			return
+		}
+		rc.cache[legacyKey] = cvElem
+		rc.lruList.Remove(revElem)
+	}
+
+	// grab elem again using the hlv key in case the element in question has been evicted, in this case we don't want to update cache
+	if elem := rc.hlvCache[key]; elem != nil {
```
adamcfraser: I don't think this block (from line 402 to 414) is necessary after making the above changes, is it?

db/revision_cache_lru.go
```diff
@@ -272,42 +329,212 @@ func (rc *LRURevisionCache) getValue(docID, revID string, create bool) (value *r
...
+	defer rc.lock.Unlock()
+	// check for existing value in rev cache map (due to concurrent fetch by rev ID)
+	cvElem, cvFound := rc.hlvCache[key]
+	revElem, revFound := rc.cache[legacyKey]
+	if !cvFound {
+		// its possible the element has been evicted if we don't find the element above (high churn on rev cache)
+		// need to return doc revision to caller still but no need repopulate the cache
+		return
+	}
+	// Check if the element in the other lookup cache points to a value that has been updated in both lookup maps already
```
adamcfraser: I think this comment is unnecessarily verbose, as the updated code is fairly self-documenting.  I think it could be reduced to `Check if another goroutine has already updated the rev map` here.

db/revision_cache_lru.go
```diff
@@ -272,42 +329,212 @@ func (rc *LRURevisionCache) getValue(docID, revID string, create bool) (value *r
...
+	//	 - Process B looks up key in cv lookup map, find and entry, assumes duplicate and removes the entry from the cv lookup map,
+	//	 and removes the element from teh cache list
+	//	 - We now find ourselves in a situation where both processes have removed both respective cache loads from the cache itself
+	//	 thus no entry in cache for this doc revision/version anymore. This needs to be avoided.
+	if revFound {
+		if cvElem == revElem {
+			// already match, return
+			return
+		}
+		rc.cache[legacyKey] = cvElem
```
adamcfraser, line 387: Could add a comment like "if CV map and rev map are targeting different list elements, update to have both use the cv map"

db/revision_cache_lru.go
```diff
@@ -272,42 +329,212 @@ func (rc *LRURevisionCache) getValue(docID, revID string, create bool) (value *r
...
+		value.lock.Unlock()
+		rc.cache[legacyKey] = elem
+		for rc.lruList.Len() > int(rc.capacity) {
+			rc.purgeOldest_()
+		}
+	}
+}
+
+// addToHLVMapPostLoad will generate and entry in the CV lookup map for a new document entering the cache
+func (rc *LRURevisionCache) addToHLVMapPostLoad(docID, revID string, cv *CurrentVersionVector) {
```
adamcfraser, line 396: Same comments as addToRevMapPostLoad about less verbose comments, and not requiring the lower block of code.

adamcfraser (changes requested):

db/revision_cache_lru.go
gregns1: I think we do, If we get to the point in the code on line 393 and we revFound is false we still need to add it to the rev lookup map, as the item hasn't been found in that map. Upon reading with fresher eyes this morning I realise I can refactor to be part of the if statement to make it better though. I did sanity check my thinking by removing the code altogether and running test but a lot of the test fail without it for example on a test where i'm putting a doc with cv defined in the cache, then lookup by revid it fails to find it.

gregns1: Extra context: Essentially gets exposed when doing a PUT on rev cache. As I have written the code to always insert using the CV (the getValueByCV code path) then subsequently call addToRevMapPostLoad to update the rev lookup. When it gets to this code on line 393 as the code has so far only added to the cv lookup map it doesn't have an entry in the rev lookup. So this code here is needed to add the reference to rev lookup.

db/revision_cache_lru.go
gregns1: Good spot, I agree its not needed with this function. But the other function addToRevMapPostLoad still needs it in (see above comment for the context).

db/crud.go
gregns1: Very good point, totally didn't even consider that. I made a change locally to do this, I essentially made a helper function to look at the docs's defined HLV at that point, any values defined with the max uint64 value we swap out for the casOut value. However this has exposed a further issue. A couple of the rev cache test now fail. Test TestPutExistingRevRevisionCacheAttachmentProperty being one of them. It looks like because it calls PutExistingRevWithBody to update a rev. In doing so the HLV logic is to leave the version source in tact and populate the cvCAS. With this in mind the update to the doc will allocate a new revid to it, but as we leave the hlv that is defined in tact this means the original CV allocated to this doc is used on the rev cache. So when we call PUT then subsequently getValueByCV it finds an entry for that key in the CV lookup map. Thus not creating new value for the update in the cache. Then we hit addToRevMapPostLoad and it create a new key and adds to the rev lookup fine (as we have a new rev id defined) but then points that new entry in the map to the CV entry which is not the new update to the doc, its the old doc version. Essentially the doc update gets lost the cache here. 

Overall means the changes for CBG-3211 PutExistingRev for CV need to go in before this part fully works. I'm not sure if to leave it as it is, populating the version with casOut for now to allow test to work? But leave comment to say it needs updating when that ticket goes in? The PR up currently is mine so I could just add it to that when this one goes in?

db/revision_cache_lru.go
gregns1: On Friday we discussed about removing the rev id assignment here. I have no refactored to check if we have the rev id defined on the cvElem and if not we assign rev id. This is because the rev id assignment here is needed for the PUT code path. So to ensure we only acquire the lock when we need it I have refactored to check if the revid is empty on the elem in the list, if so we define it else we leave it just adding to the rev lookup map.

db/crud.go
gregns1: I have now implemented this. We have added a new doc update type called `InternalExistingVersion` which will follow the NewVersion code path when updating the HLV. So then when using `PutExistingRevWithBody` in our test framework we now generate a new HLV and the rev cache works as expected. This throws another issue though, as import logic hasn't yet been implemented, HLV is currently left as is when import event is triggered. This leads to similar issue we had with `PutExistingRevWithBody` where when a test adds an update to a doc, then relies upon import to bring this change to sync gateway, the new version of this doc is lost at the cache as it finds the old version in the cache. 

I found this issue in test `TestAttachmentDeleteOnExpiry` the test puts a doc with expiry, waits for the doc to expire then uses on demand import to import the tombstone. The issue is we get to import the tombstone and as the HLV is not updated currently on import, the old version is found in the cache and thus we don't create new entry at the cache for the tombstone. 

To fix this temperately I have added a check to postWriteUpdateHLV function that should be removed when work on CBG-3503 has been worked on. The check simply compares the doc cas and the version in the HLV, if not the same then sets the version to casOut. Although this logic is incorrect it allows the failing tests to pass for now. I feel like the only other way around this would be to either roll the import logic into this PR or get the import logic implemented before this goes in.

adamcfraser (approved):

PR #6495, opened 2023-10-05
Remove walrus restriction from tests that now work with rosmar

torcolvin:


db/attachment_test.go
```diff
@@ -1160,12 +1160,10 @@ func TestMigrateBodyAttachmentsMerge(t *testing.T) {
 // TestMigrateBodyAttachmentsMergeConflicting will set up a document with the same attachment name in both pre-2.5 and post-2.5 metadata, making sure that the metadata with the most recent revpos is chosen.
 func TestMigrateBodyAttachmentsMergeConflicting(t *testing.T) {
 
-	if base.TestUseXattrs() && base.UnitTestUrlIsWalrus() {
+	if base.TestUseXattrs() {
 		t.Skip("Test requires Couchbase Server bucket when using xattrs")
 	}
```
bbrks: Looks like this test is able to work with and without xattrs

```suggestion
```

bbrks (changes requested):
Looks good. I think the one still limiting xattrs=true doens't need to be

bbrks (approved):

PR #6504, opened 2023-10-07
CBG-3384 Avoid triggering unnecessary resync when working with default collection

adamcfraser:
CBG-3384

- Don’t persist syncInfo for empty metadataID (legacy config)
- Modify requirements for assignment of default metadata ID to a database:
  1. Remove requirement for presence of _sync:seq in default collection
  2. Allow assignment when default collection’s syncInfo is using defaultMetadataID

These avoid triggering resync for databases using only the default collection, in legacy and non-legacy mode.

Also ensures databases share metadataIDs across config groups (CBG-3527)

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2111/

### DIFF
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -42,8 +42,12 @@ type BootstrapConnection interface {
 	// TouchMetadataDocument sets the specified property in a bootstrap metadata document for a given bucket and key.  Used to
 	// trigger CAS update on the document, to block any racing updates. Does not retry on CAS failure.
 	TouchMetadataDocument(ctx context.Context, bucket, key string, property string, value string, cas uint64) (casOut uint64, err error)
-	// KeyExists checks whether the specified key exists
+	// KeyExists checks whether the specified key exists in the bucket's default collection
 	KeyExists(ctx context.Context, bucket, key string) (exists bool, err error)
+	// GetDocument retrieves the document with the specified key from the bucket's default collection.
+	// Returns exists=false if key is not found, returns error for any other error.
+	GetDocument(ctx context.Context, bucket, docID string, rv interface{}) (exists bool, err error)
+
 	// Returns the bootstrap connection's cluster connection as N1QLStore for the specified bucket/scope/collection.
 	// Does NOT establish a bucket connection, the bucketName/scopeName/collectionName is for query scoping only
 	GetClusterN1QLStore(bucketName, scopeName, collectionName string) (*ClusterOnlyN1QLStore, error)
@@ -447,6 +451,7 @@ func (cc *CouchbaseCluster) UpdateMetadataDocument(ctx context.Context, location
 
 }
 
+// KeyExists checks whether a key exists in the default collection for the specified bucket
 func (cc *CouchbaseCluster) KeyExists(ctx context.Context, location, docID string) (exists bool, err error) {
 	if cc == nil {
 		return false, errors.New("nil CouchbaseCluster")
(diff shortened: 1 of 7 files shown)


base/bootstrap.go
```diff
@@ -42,8 +42,12 @@ type BootstrapConnection interface {
 	// TouchMetadataDocument sets the specified property in a bootstrap metadata document for a given bucket and key.  Used to
 	// trigger CAS update on the document, to block any racing updates. Does not retry on CAS failure.
 	TouchMetadataDocument(ctx context.Context, bucket, key string, property string, value string, cas uint64) (casOut uint64, err error)
-	// KeyExists checks whether the specified key exists
+	// KeyExists checks whether the specified key exists in the bucket's default collection
```
torcolvin, line 45: should we say metadata collection here and below for when we migrate this to a separate collection?

Right now default collection = metadata collection but it may not always.

base/bootstrap.go
```diff
@@ -463,6 +468,34 @@ func (cc *CouchbaseCluster) KeyExists(ctx context.Context, location, docID strin
...
+		Transcoder: NewSGJSONTranscoder(),
+	}
+	getResult, err := b.DefaultCollection().Get(docID, getOptions)
+	if err != nil {
+		if errors.Is(err, gocb.ErrDocumentNotFound) {
+			return false, nil
+		}
+		return false, err
+	}
+	err = getResult.Content(rv)
```
torcolvin, line 495: fix ineffassign

base/constants_syncdocs.go
```diff
@@ -397,6 +400,9 @@ func InitSyncInfo(ds DataStore, metadataID string) (requiresResync bool, err err
 
 // SetSyncInfo sets syncInfo in a DataStore to the specified metadataID
 func SetSyncInfo(ds DataStore, metadataID string) error {
+	if metadataID == "" {
```
torcolvin, line 405: Can you drop a comment for why this happens here, for future reference?

rest/config_manager.go
```diff
@@ -753,21 +752,19 @@ func (b *bootstrapContext) computeMetadataID(ctx context.Context, registry *Gate
...
 	bucketName := config.GetBucketName()
-	exists, err := b.Connection.KeyExists(ctx, bucketName, base.SGSyncInfo)
+	var syncInfo base.SyncInfo
+	exists, err := b.Connection.GetDocument(ctx, bucketName, base.SGSyncInfo, &syncInfo)
 	if err != nil {
-		base.WarnfCtx(ctx, "Error checking whether metadataID is already defined for default collection - using standard metadataID.  Error: %v", err)
-		return standardMetadataID
-	}
-	if exists {
+		base.WarnfCtx(ctx, "Error checking syncInfo metadataID in default collection - using standard metadataID.  Error: %v", err)
```
torcolvin, line 770: I haven't thoroughly thought this through, but is this going to be called only when there's a cas change, or is this going to log at bootstrap pooling interval?

torcolvin (commented):
LGTM

base/bootstrap.go
adamcfraser: I intentionally left this comment as default collection because the bootstrap doesn't currently have the concept of a metadata store, and will require some refactoring when we want to add system collection support.

rest/config_manager.go
adamcfraser: This is only going to be called once per db creation - once the metadataID is assigned (to standardMetadataID in this case), it won't fire again.

rest/upgradetest/upgrade_registry_test.go
```diff
@@ -160,6 +162,171 @@ func TestUpgradeDatabasePreHelium(t *testing.T) {
...
+	assert.Equal(t, http.StatusCreated, resp.Code)
+
+	// Get the legacy config for upgrade test below
+	resp = legacyRT.SendAdminRequest("GET", "/_config?include_runtime=true", "")
+	legacyConfigBytes := resp.BodyBytes()
+
+	var legacyConfig rest.LegacyServerConfig
+	err := base.JSONUnmarshal(legacyConfigBytes, &legacyConfig)
+	assert.NoError(t, err)
+	legacyRT.Close()
```
torcolvin, line 285: move this up to defer at the top of test to protect other tests?

torcolvin (approved):

rest/upgradetest/upgrade_registry_test.go
adamcfraser: This needs to be closed before opening the non-legacy RTs, so leaving as-is.

PR #6507, opened 2023-10-09
makes golangci-lint more strict

torcolvin:
Add unconvert / unused to stricter linting. Skip code in `rest/debug.go` which is largely unused.


bbrks:
What level of analysis has been done to determine whether the things being removed are unintentionally unused? (just so I know how much digging to do as part of the review)

torcolvin:
> What level of analysis has been done to determine whether the things being removed are unintentionally unused? (just so I know how much digging to do as part of the review)

I didn't do a huge amount of analysis about unused code, except around new code. If it was unused, I assumed we had accidentally written around it, like I did with [0069df9](https://github.com/couchbase/sync_gateway/pull/6507/commits/0069df9275060b54ae16c973e221c0a6c8e73d27)

base/dcp_dest.go
```diff
@@ -251,10 +244,6 @@ func partitionToVbNo(ctx context.Context, partition string) uint16 {
 	return uint16(vbNo)
 }
 
-func vbNoToPartition(vbNo uint16) string {
-	return vbucketIdStrings[vbNo]
```
bbrks, line 256: This array is initialized with 1024 vbucket IDs but also no longer read once this function has gone.
We could do more and rip out the declaration and init whilst we're here.

base/dcp_dest.go
```diff
@@ -29,12 +29,6 @@ import (
 // (Atoi is faster than map lookup when going in the other direction)
 var vbucketIdStrings [1024]string
 
-type cbgtFeedType int8
-
-const (
-	cbgtFeedType_gocb cbgtFeedType = iota
-)
-
 type destFeedType int8
```
bbrks: This also seems unused?

db/event.go
```diff
@@ -108,7 +108,6 @@ const (
 // A compiled JavaScript event function.
 type jsEventTask struct {
 	sgbucket.JSRunner
-	responseType ResponseType
```
bbrks, line 111: `type ResponseType` and the const values could be dropped as well if this goes

bbrks (dismissed):
Everything looks good and intentionally removed - couple of places where we'll get new unused code as soon as we merge so might as well go rip out the other bits.

base/dcp_dest.go
torcolvin: Ripped this out with the test.

bbrks (approved):

PR #6511, opened 2023-10-10
CBG-3498 create function NewRestTesterPersistentConfig

torcolvin:
`NewRestTesterPersistentConfig` will create a `RestTester` with persistent config and start up a database. This differs from setting `RestTesterConfig{Persistent: true}` in that the database is created for you automatically. I didn't allow passing configuration options, since I am not sure all of them work with `Persistent: true` and creating the database. There are legitimate reasons you might want to not create a "default" database with custom options. That said, you almost want an enum:

1. CreateLegacyDatabase
2. CreatePersistentDatabase
3. CreateNoDatabase

to replace the `Persistent` option of `RestTester`.

This PR migrates many uses of creating ServerContext directly to RestTester which makes the tests collection-aware and often runnable under rosmar. Some uses were NOT migrated if they interacted with the `ServerContext` configuration options, or primarily just called bootstrap functions directly. 

Additionally missing from the migration is some of the `TestResync.*` since they are actually inactive because they skip if a query based resync is not set up. This seemed out of scope for this ticket, since some of the tests don't pass as is.

- https://github.com/couchbase/sync_gateway/blob/master/rest/adminapitest/admin_api_test.go#L596
- https://github.com/couchbase/sync_gateway/blob/master/rest/adminapitest/admin_api_test.go#L667
- https://github.com/couchbase/sync_gateway/blob/master/rest/adminapitest/admin_api_test.go#L981 (new collections test that never ran)
-  https://github.com/couchbase/sync_gateway/blob/master/rest/adminapitest/admin_api_test.go#L1246
-  https://github.com/couchbase/sync_gateway/blob/master/rest/adminapitest/admin_api_test.go#L1433

There are two parts to this review:

### Evaluate the new functions and ergnomics in general. 

1. Do you like `NewRestTesterPersistentConfig` ? The nice thing about this API is that it will make it easier to update tests to use rosmar when we are ready, I tried to mark the tests that actually require a bootstrap connection.
2. Do you prefer code that is one line like: 

```
RequireStatus(t, rt.SendRequestWithHeaders(http.MethodPost, "/{{.db}}/_session", "{}", map[string]string{"Authorization": BearerToken + " " + jwt}), http.StatusUnauthorized)
```

or two lines like:

```
resp := rt.SendRequestWithHeaders(http.MethodPost, "/{{.db}}/_session", "{}", map[string]string{"Authorization": BearerToken + " " + jwt})
RequireStatus(t, resp, http.StatusUnauthorized)
```

3. `rest.RequireStatus` only verifies status code, and there was an api like `resp.RequireResponse(http.StatusOK, importFilter)` which could verify text and status code in one. There's not a comparable API, but 

### Do a review that the tests test the same code that they did before.

I think only one person needs to do this evaluation.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2121/


rest/utilities_testing.go
```diff
@@ -2500,7 +2519,19 @@ func (rt *RestTester) NewDbConfig() DbConfig {
...
 			syncFn = base.StringPtr(rt.SyncFn)
 		}
-		config.Scopes = GetCollectionsConfigWithSyncFn(rt.TB, rt.TestBucket, syncFn, rt.numCollections)
+		var importFilter *string
+		if rt.ImportFilter != "" {
+			importFilter = base.StringPtr(rt.ImportFilter)
+		}
+
+		config.Scopes = GetCollectionsConfigWithFiltering(rt.TB, rt.TestBucket, rt.numCollections, syncFn, importFilter)
+	} else {
```
bbrks, line 2523: The `if` condition was bad enough, but adding an `else` to it makes parsing this a bit of a head scratcher prior to coffee ...

Is there any chance of some intermediate variables to help reduce the thinking required when reading this?

rest/utilities_testing.go
```diff
@@ -631,15 +650,15 @@ func (rt *RestTester) templateResource(resource string) (string, error) {
...
-					} else {
-						data[fmt.Sprintf("db%dkeyspace%d", i+1, j+1)] = keyspace
-					}
+				continue
+			}
+			for j, keyspace := range getKeyspaces(rt.TB, database) {
+				if !multipleDatabases {
+					data[fmt.Sprintf("keyspace%d", j+1)] = keyspace
+				} else {
+					data[fmt.Sprintf("db%dkeyspace%d", i+1, j+1)] = keyspace
```
bbrks, line 664: Is there a combination of multiple database, single collection? Would that have to use `db1keyspace1` rather than `db1keyspace` ?

bbrks (dismissed):
LGTM - Two nits, give or take.

bbrks (approved):

PR #6514, opened 2023-10-10
CBG-3465 [3.1.2 backport] Delete legacy config when migrated config already exists

adamcfraser:
CBG-3465

Backports CBG-3464 and accompanying test fix.


torcolvin (approved):

adamcfraser:
Failure is known test race.

PR #6515, opened 2023-10-11
CBG-3211: Add PutExistingRev for HLV

gregns1:
CBG-3211

Adds a new function entitled PutExistingCurrentVersion which carries much the same functionality for PutExistingRev but carries out logic on the HLV.  
- Check for conflict between the two HLV's  
- If conflict is found we cancel the rest of the update to the doc. Here we need to add merge version and send the doc back to the client.  
- If not in conflict we need to add all newer source/version pairs from the incoming HLV to the local HLV for that doc. I have added a function called AddNewerVersions that does this, first by iterating through the Previous Versions and adding any newer version pair, then the incoming HLV's current version pair. 
- If not in conflict the revtree entry still needs to be computed and added to the doc update. 
- Added tests to ensure that the CV of the incoming HLV is preserved and that the cvCAS is updated still, then the PV of the HLV on the doc is updated correctly. 
- Also included a test where a conflict is found and assert we cancel the doc update, then fetch the persisted sync data to ensure it has remained unchanged. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2149/

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -2084,14 +2084,6 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 			// update the mutate in options based on the above logic
 			updatedSpec = doc.SyncData.HLV.computeMacroExpansions()
 
-			// update the HLV values
-			doc, err = db.updateHLV(doc, docUpdateEvent)
-			if err != nil {
-				return
-			}
-			// update the mutate in options based on the above logic
-			updatedSpec = doc.SyncData.HLV.computeMacroExpansions()
-
 			deleteDoc = currentRevFromHistory.Deleted
 
 			// Return the new raw document value for the bucket to store.
--- a/db/crud.go
+++ b/db/crud.go
@@ -1144,7 +1144,6 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 		}
 
 		newDoc.RevID = newRev
-		newDoc.Deleted = newDoc.Deleted
 
 		return newDoc, newAttachments, false, nil, nil
 	})
--- a/db/crud.go
+++ b/db/crud.go
@@ -1062,6 +1062,105 @@ func (db *DatabaseCollectionWithUser) Put(ctx context.Context, docid string, bod
 	return newRevID, doc, err
 }
 
+func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Context, newDoc *Document, docHistory []string, docHLV HybridLogicalVector, existingDoc *sgbucket.BucketDocument) (doc *Document, cv *CurrentVersionVector, newRevID string, err error) {
+	newRev := docHistory[0]
+	generation, _ := ParseRevID(ctx, newRev)
+	if generation < 0 {
+		return nil, nil, "", base.HTTPErrorf(http.StatusBadRequest, "Invalid revision ID")
+	}
+
+	docUpdateEvent := ExistingVersion
+	allowImport := db.UseXattrs()
+	doc, _, err = db.updateAndReturnDoc(ctx, newDoc.ID, allowImport, newDoc.DocExpiry, nil, docUpdateEvent, existingDoc, func(doc *Document) (resultDoc *Document, resultAttachmentData AttachmentData, createNewRevIDSkipped bool, updatedExpiry *uint32, resultErr error) {
+		// (Be careful: this block can be invoked multiple times if there are races!)
+
+		var isSgWrite bool
+		var crc32Match bool
+
+		// Is this doc an sgWrite?
+		if doc != nil {
+			isSgWrite, crc32Match, _ = doc.IsSGWrite(ctx, nil)
+			if crc32Match {
+				db.dbStats().Database().Crc32MatchCount.Add(1)
+			}
+		}
+
+		// If the existing doc isn't an SG write, import prior to updating
+		if doc != nil && !isSgWrite && db.UseXattrs() {
+			err := db.OnDemandImportForWrite(ctx, newDoc.ID, doc, newDoc.Deleted)
+			if err != nil {
+				return nil, nil, false, nil, err
+			}
+		}
+
+		// Find the point where this doc's history branches from the current rev:
+		currentRevIndex := len(docHistory)
+		parent := ""
+		for i, revid := range docHistory {
+			if doc.History.contains(revid) {
+				currentRevIndex = i
+				parent = revid
+				break
+			}
+		}
+		if currentRevIndex == 0 {
+			base.DebugfCtx(ctx, base.KeyCRUD, "PutExistingRevWithBody(%q): No new revisions to add", base.UD(newDoc.ID))
+			newDoc.RevID = newRev
+			return nil, nil, false, nil, base.ErrUpdateCancel // No new revisions to add
+		}
+
+		// Conflict check here
+		if doc.HLV == nil {
+			doc.HLV = &HybridLogicalVector{}
+		}
+		if !docHLV.IsInConflict(*doc.HLV) {
+			// update hlv for all newer incoming source version pairs
+			doc.HLV.AddNewerVersions(docHLV)
+		} else {
+			base.InfofCtx(ctx, base.KeyCRUD, "conflict detected between the two HLV's for doc %s", base.UD(doc.ID))
+			// cancel rest of update, HLV needs to be sent back to client with merge versions populated
+			return nil, nil, false, nil, base.HTTPErrorf(http.StatusConflict, "Document revision conflict")
+		}
+
+		// Add all the new-to-me revisions to the rev tree:
+		for i := currentRevIndex - 1; i >= 0; i-- {
+			err := doc.History.addRevision(newDoc.ID,
+				RevInfo{
+					ID:      docHistory[i],
+					Parent:  parent,
+					Deleted: i == 0 && newDoc.Deleted})
+
+			if err != nil {
+				return nil, nil, false, nil, err
+			}
+			parent = docHistory[i]
+		}
+
+		parentRevID := doc.History[newRev].Parent
+		// Process the attachments, replacing bodies with digests.
+		newAttachments, err := db.storeAttachments(ctx, doc, newDoc.DocAttachments, generation, parentRevID, docHistory)
+		if err != nil {
+			return nil, nil, false, nil, err
+		}
+
+		newDoc.RevID = newRev
+		newDoc.Deleted = newDoc.Deleted
+
+		return newDoc, newAttachments, false, nil, nil
+	})
+
+	if doc != nil && doc.HLV != nil {
+		if cv == nil {
+			cv = &CurrentVersionVector{}
+		}
+		source, version := doc.HLV.GetCurrentVersion()
+		cv.SourceID = source
+		cv.VersionCAS = version
+	}
+
+	return doc, cv, newRev, err
+}
+
 // Adds an existing revision to a document along with its history (list of rev IDs.)
 func (db *DatabaseCollectionWithUser) PutExistingRev(ctx context.Context, newDoc *Document, docHistory []string, noConflicts bool, forceAllConflicts bool, existingDoc *sgbucket.BucketDocument, docUpdateEvent DocUpdateType) (doc *Document, newRevID string, err error) {
 	return db.PutExistingRevWithConflictResolution(ctx, newDoc, docHistory, noConflicts, nil, forceAllConflicts, existingDoc, docUpdateEvent)
--- a/db/crud.go
+++ b/db/crud.go
@@ -1062,7 +1062,7 @@ func (db *DatabaseCollectionWithUser) Put(ctx context.Context, docid string, bod
 	return newRevID, doc, err
 }
 
-func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Context, newDoc *Document, docHLV HybridLogicalVector, existingDoc *sgbucket.BucketDocument) (doc *Document, cv *CurrentVersionVector, newRevID string, err error) {
+func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Context, newDoc *Document, docHLV HybridLogicalVector, existingDoc *sgbucket.BucketDocument) (doc *Document, cv *SourceAndVersion, newRevID string, err error) {
 	var matchRev string
 	if existingDoc != nil {
 		doc, unmarshalErr := unmarshalDocumentWithXattr(ctx, newDoc.ID, existingDoc.Body, existingDoc.Xattr, existingDoc.UserXattr, existingDoc.Cas, DocUnmarshalRev)
@@ -1149,11 +1149,11 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 
 	if doc != nil && doc.HLV != nil {
 		if cv == nil {
-			cv = &CurrentVersionVector{}
+			cv = &SourceAndVersion{}
 		}
 		source, version := doc.HLV.GetCurrentVersion()
 		cv.SourceID = source
-		cv.VersionCAS = version
+		cv.Version = version
 	}
 
 	return doc, cv, newRevID, err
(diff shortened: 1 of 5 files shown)


db/hybrid_logical_vector.go
```diff
@@ -283,3 +304,11 @@ func (hlv *HybridLogicalVector) computeMacroExpansions() []sgbucket.MacroExpansi
 	}
 	return outputSpec
 }
+
+// addToPreviousVersions will take a source/version pair and add it to the HLV previous versions map
+func (hlv *HybridLogicalVector) addToPreviousVersions(source string, version uint64) {
+	if hlv.MergeVersions == nil {
+		hlv.MergeVersions = make(map[string]uint64)
+	}
+	hlv.MergeVersions[source] = version
```
adamcfraser: This looks like it's incorrectly targeting hlv.MergeVersions (instead of hlv.PreviousVersions)?

db/hybrid_logical_vector.go
```diff
@@ -283,3 +304,11 @@ func (hlv *HybridLogicalVector) computeMacroExpansions() []sgbucket.MacroExpansi
 	}
 	return outputSpec
 }
+
+// addToPreviousVersions will take a source/version pair and add it to the HLV previous versions map
+func (hlv *HybridLogicalVector) addToPreviousVersions(source string, version uint64) {
```
adamcfraser: I'd name this 'setPreviousVersion()' or possibly 'upsertPreviousVersion()' - add usually implies the operation should fail if a value already exists for the key.

db/hybrid_logical_vector.go
```diff
@@ -180,6 +183,24 @@ func (hlv *HybridLogicalVector) GetVersion(sourceID string) uint64 {
...
 }
 
+// AddNewerVersions will take a hlv and add any newer source/version pairs found across CV and PV found in the other HLV taken as parameter
+func (hlv *HybridLogicalVector) AddNewerVersions(otherVector HybridLogicalVector) {
+	if otherVector.PreviousVersions != nil || len(otherVector.PreviousVersions) != 0 {
+		_, localVersion := hlv.GetCurrentVersion()
+		// Iterate through incoming vector previous versions, compare to current version cas of local HLV
+		// Any more recent versions in incoming vectors previous versions need to be added to local HLV previous versions
+		for i, v := range otherVector.PreviousVersions {
+			if v > localVersion {
```
adamcfraser: We want to update with the `version` from otherVector for `source` if hlv's version for that `source` is lower.  There isn't a requirement that this value is greater than currentVersion.

db/crud.go
```diff
@@ -1044,6 +1044,104 @@ func (db *DatabaseCollectionWithUser) Put(ctx context.Context, docid string, bod
 	return newRevID, doc, err
 }
 
+func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Context, newDoc *Document, docHistory []string, docHLV HybridLogicalVector, existingDoc *sgbucket.BucketDocument) (doc *Document, cv *CurrentVersionVector, newRevID string, err error) {
```
adamcfraser: We don't expect callers of PutExistingCurrentVersion to be providing a revision history as well (docHistory). 

As mentioned on the ticket, this function should be computing a new revID for the mutation, similar to Put.

adamcfraser (changes requested):
A few areas that need a second look.

db/hybrid_logical_vector.go
```diff
@@ -180,6 +183,32 @@ func (hlv *HybridLogicalVector) GetVersion(sourceID string) uint64 {
 	return latestVersion
 }
 
+// AddNewerVersions will take a hlv and add any newer source/version pairs found across CV and PV found in the other HLV taken as parameter
+func (hlv *HybridLogicalVector) AddNewerVersions(otherVector HybridLogicalVector) {
```
adamcfraser: Can we add some unit tests for AddNewerVersions to ensure we're covering all permutations?

In particular, I'm unsure that the current code handles cases like:
```
hlv:
   sourceID: abc,
   version: 15
otherVector:
   sourceID: def
   version: 25,
   PreviousVersions: "abc":20
expectedHLV:
   sourceID: def,
   version: 25,
   PreviousVersions: "abc":20
```
and
```
hlv:
   sourceID: abc,
   version: 15
   PreviousVersions:"def":30 
otherVector:
   sourceID: def
   version: 35,
   PreviousVersions: "abc":15
expectedHLV:
   sourceID: def,
   version: 35,
   PreviousVersions: "abc":15
```

db/hybrid_logical_vector.go
```diff
@@ -180,6 +183,32 @@ func (hlv *HybridLogicalVector) GetVersion(sourceID string) uint64 {
...
+			if hlv.PreviousVersions[i] < v {
+				hlv.setPreviousVersion(i, v)
+			}
+		}
+	}
+
+	// create current version for incoming vector and attempt to add it to the local HLV, AddVersion will handle if attempting to add older
+	// version than local HLVs CV pair
+	otherVectorCV := CurrentVersionVector{SourceID: otherVector.SourceID, VersionCAS: otherVector.Version}
+	err := hlv.AddVersion(otherVectorCV)
```
adamcfraser: I think AddVersion makes the assumption that a given source won't appear in both cv (sourceID) and also pv, but that scenario is possible here based on the iteration above.

adamcfraser (changes requested):

db/hybrid_logical_vector.go
gregns1: Good spot. I have added some code in AddNewerVersions that accounts for this.

db/hybrid_logical_vector.go
gregns1: Good spot, exposed some issue with my logic there. Added these test cases plus couple extra. 

There is some logic I have implemented in AddNewerVersions that accounts for otherVector having a lower CV. In this case AddVersion will error so added some code to add the CV from otherVector to PV in local HLV. Now in this specific use case I don't think this should ever be the case so I don't believe this code is needed but left it in as I feel it could be useful for some other scenarios where you may want to almost 'merge' two HLV structures? But if you disagree i'm happy to remove these lines of code. For reference I'm talking about lines 212 to 226 in AddNewerVersions.

db/hybrid_logical_vector_test.go
```diff
@@ -229,3 +229,56 @@ func TestHybridLogicalVectorPersistence(t *testing.T) {
 	assert.Equal(t, inMemoryHLV.PreviousVersions, hlvFromPersistance.PreviousVersions)
 	assert.Equal(t, inMemoryHLV.MergeVersions, hlvFromPersistance.MergeVersions)
 }
+
+func TestAddNewerVerionsBetweenTwoVectors(t *testing.T) {
```
adamcfraser: This seems like it would be an excellent candidate for subtests.  Defining a struct with 'localHLV', 'incomingHLV' and 'expectedHLV' as properties would make these tests very easy to quickly read and verify they are testing the right thing.  e.g.
```
{
 localHLV: []string{"abc@15"},
 incomingHLV: []string{"def@25", "abc@20"},
 expectedHLV: []string{"def@25", "abc@20"},
}, ...
```

adamcfraser:
```suggestion
func TestAddNewerVersionsBetweenTwoVectors(t *testing.T) {
```

db/util_testing.go
```diff
@@ -644,3 +644,25 @@ func DefaultMutateInOpts() *sgbucket.MutateInOptions {
 		MacroExpansion: macroExpandSpec(base.SyncXattrName),
 	}
 }
+
+func constructDocumentFromBody(docID string, body Body) (newDoc *Document) {
```
adamcfraser: This is doing a lot of work that's not actually required by the new test usage (which doesn't include any of these reserved properties in the body).    In particular this looks like it's designed to handle document body properties in the way the 1.x REST API does, but we don't currently implement REST API support for HLVs. 

Any reason to use this approach, as opposed to a helper function like createTestDocument(docID string, revID string, body Body, deleted bool, ...)?

db/hybrid_logical_vector.go
```diff
@@ -85,7 +85,17 @@ func (hlv *HybridLogicalVector) AddVersion(newVersion CurrentVersionVector) erro
 	if hlv.PreviousVersions == nil {
 		hlv.PreviousVersions = make(map[string]uint64)
 	}
-	hlv.PreviousVersions[hlv.SourceID] = hlv.Version
+	// we need to check if source ID already exists in PV, if so we need to ensure we are only updating with the
```
adamcfraser, line 97: I feel like the check on line 70 should only be used when newVersion.SourceID == hlv.sourceID.  Do you agree?

db/hybrid_logical_vector.go
```diff
@@ -85,7 +85,17 @@ func (hlv *HybridLogicalVector) AddVersion(newVersion CurrentVersionVector) erro
...
+	// sourceID-version pair if incoming version is greater than version already there
+	if currPVVersion := hlv.PreviousVersions[hlv.SourceID]; currPVVersion != 0 {
```
adamcfraser: I think using the existence check for a map like `if currPVVersion, ok := hlv.PreviousVersions[hlv.SourceID]; ok {` is more idiomatic go for this type of check.

db/hybrid_logical_vector.go
```diff
@@ -85,7 +85,17 @@ func (hlv *HybridLogicalVector) AddVersion(newVersion CurrentVersionVector) erro
...
+		// if we get here source ID exists in PV, only replace version if it is less than the incoming version
+		if currPVVersion < hlv.Version {
+			hlv.PreviousVersions[hlv.SourceID] = hlv.Version
+		}
```
adamcfraser, line 105: Should we return error in an else here?

db/hybrid_logical_vector.go
```diff
@@ -167,6 +177,9 @@ func (hlv *HybridLogicalVector) equalPreviousVectors(otherVector HybridLogicalVe
 
 // GetVersion returns the latest CAS value in the HLV for a given sourceID, if the sourceID is not present in the HLV it will return 0 CAS value
 func (hlv *HybridLogicalVector) GetVersion(sourceID string) uint64 {
+	if sourceID == "" {
+		return 0
```
adamcfraser: Returning zero here gives the impression that there's an entry in the HLV for `sourceID:0`.  What's the benefit of returning zero, instead of GetVersion returning (uint64, bool)?

db/hybrid_logical_vector.go
```diff
@@ -180,6 +193,43 @@ func (hlv *HybridLogicalVector) GetVersion(sourceID string) uint64 {
...
+				hlv.setPreviousVersion(otherVector.SourceID, otherVector.Version)
+			} else if hlv.PreviousVersions[otherVector.SourceID] != 0 && hlv.PreviousVersions[otherVectorCV.SourceID] < otherVector.Version {
+				// source exists but has version less than incoing version associated with it so replace this value in PV
+				// with incoming pair
+				hlv.setPreviousVersion(otherVector.SourceID, otherVector.Version)
+			}
+		}
+	}
+	// if current source exists in PV, delete it.
+	if hlv.PreviousVersions[hlv.SourceID] != 0 {
```
adamcfraser: Same comment here about using existence check instead of ascribing meaning to version=0.

adamcfraser (changes requested):

db/hybrid_logical_vector.go
gregns1: Yes thinking about it that makes sense I think. So I have added this but now have had to move the call to AddVersion above the iteration through the previous version inside AddNewerVersions. Take example 1 of your test cases above you provided. We can end up with adding "abc" and 20 to PV then call add version and hit the error as we have value in PV higher than the current version. I don't think this is an issue, tests seem fine moving it and trying to think through it I don't think there is an logical issues with this.

db/hybrid_logical_vector.go
gregns1: Yes I agree I think? I'm not sure I fully understand though, my worry was about protecting against us trying to add a new version from different source id with version less than current version. 

E.g. local HLV CV = src: bucket1, vrs: 123 then call this function with the above change attempting to add CV src: bucket2, vrs: 12 would success in adding the incoming CV pair to local CV then moving the bucket1, 123 pair to PV? Given that the incoming pair is less than current pair in terms of version this seems incorrect? Unless you're suggesting tis isn't needed as thus function shouldn't be called unless this incoming pair is a newer CV pair?

gregns1: An example of the above is testcase4 in my test (i've had to change the expected output on it due to the above change) TestAddNewerVersionsBetweenTwoVectors, In my mind having def:15 as CV and abc:20 in PV on expected output seems wrong as abc's version is greater than def's?

adamcfraser: We don't have a requirement for sources to have clock synchronization - we expect clock skew, particularly for non-server peers.

db/hybrid_logical_vector_test.go
```diff
@@ -229,3 +229,51 @@ func TestHybridLogicalVectorPersistence(t *testing.T) {
 	assert.Equal(t, inMemoryHLV.PreviousVersions, hlvFromPersistance.PreviousVersions)
 	assert.Equal(t, inMemoryHLV.MergeVersions, hlvFromPersistance.MergeVersions)
 }
+
+func TestAddNewerVersionsBetweenTwoVectors(t *testing.T) {
+	testCases := []struct {
+		name          string
+		localInput    []string
+		incomingInput []string
+		expectedInput []string
```
adamcfraser: I'm not sure why this is named 'Input' - maybe just 'expected'?

db/hybrid_logical_vector_test.go
```diff
@@ -229,3 +229,51 @@ func TestHybridLogicalVectorPersistence(t *testing.T) {
...
+			name:          "testcase3",
+			localInput:    []string{"abc@17", "def@30"},
+			incomingInput: []string{"def@35", "abc@15"},
+			expectedInput: []string{"def@35", "abc@17"},
+		},
+		{
+			name:          "testcase4",
+			localInput:    []string{"abc@20", "ghi@9"},
+			incomingInput: []string{"def@15", "abc@17"},
+			expectedInput: []string{"def@15", "ghi@9", "abc@20"},
```
adamcfraser: The local and incoming are conflicting HLVs here, as neither HLV dominates the other, so it's not really an expected use case for AddNewerVersions.  It might not need to be part of this PR, but I think a similar unit test that runs both IsInConflict and then AddNewerVersions (when !IsInConflict) would be a good idea.

db/hybrid_logical_vector.go
```diff
@@ -177,7 +193,48 @@ func (hlv *HybridLogicalVector) GetVersion(sourceID string) uint64 {
...
+		}
+	}
+	// if current source exists in PV, delete it.
+	if _, ok := hlv.PreviousVersions[hlv.SourceID]; ok {
+		delete(hlv.PreviousVersions, hlv.SourceID)
+	}
+	return nil
+}
+
+func (hlv *HybridLogicalVector) SetMergeVectors(otherVector HybridLogicalVector) {
```
adamcfraser: I'd prefer to omit MergeVersion handling from this PR - I don't think there's enough test coverage to verify we're doing it properly (e.g. we need handling for existing MVs, we need pruning of MV on subsequent updated).

db/hybrid_logical_vector_test.go
```diff
@@ -229,3 +229,66 @@ func TestHybridLogicalVectorPersistence(t *testing.T) {
...
+			expected:      []string{"def@25", "abc@20"},
+		},
+		{
+			name:          "testcase2",
+			localInput:    []string{"abc@15", "def@30"},
+			incomingInput: []string{"def@35", "abc@15"},
+			expected:      []string{"def@35", "abc@15"},
+		},
+		{
+			name:          "testcase3",
```
adamcfraser: This test case is also a conflict (incoming HLV doesn't dominate the local cv), so I think should be omitted from this test.

adamcfraser (changes requested):
Let's remove the merge handling until we're ready to implement that, and remove the invalid test case.  With that removed this should be good to merge.

db/hybrid_logical_vector_test.go
gregns1: Okay I agree I didn't notice that, I have removed this case. Does the function IsInConflict need changing then? When running localHLV.IsInConflict(incomingHLV) no conflict is detected because the function checks if either one is dominating over the other. Should this be changed to only check one way?

adamcfraser (approved):

PR #6517, opened 2023-10-11
Fix data race in AssertLogContains and functional race in TestDocChangedLogging

bbrks:
- Ensures `TestDocChangedLogging` waits for a marker doc to be recieved, as we can't reliably wait for the no-op of non-metadata doc write to happen to assert on logs.
- Forces console log buffer flush before checking contents, instead of retry loop.
- Avoids multiple buffer reads during log writes
  - Prevents race if the log write causes buffer to grow during read inside retry loop
  - Prevents functional issues if buffer read was in the middle of the log line causing failed assertions (log lines split over multiple buffer reads)

Reliably tested before/after fix with:

```
$ go test -run='^TestDocChangedLogging$' -race -count=1000 -v ./rest/changestest
```

- n/a

### DIFF
--- a/base/logging.go
+++ b/base/logging.go
@@ -388,22 +388,14 @@ func LogTraceEnabled(logKey LogKey) bool {
 // AssertLogContains asserts that the logs produced by function f contain string s.
 func AssertLogContains(t *testing.T, s string, f func()) {
 	// Temporarily override logger output
-	b := bytes.Buffer{}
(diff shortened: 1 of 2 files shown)


bbrks:
... this made the test fail way more reliably without a data race 🤔

There's something else going on here that I'd need to investigate

torcolvin (approved):

PR #6518, opened 2023-10-11
CBG-3522 add collection to import feed logic

torcolvin:
## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2098/

### DIFF
--- a/db/import_listener.go
+++ b/db/import_listener.go
@@ -130,35 +130,43 @@ func (il *importListener) StartImportFeed(dbContext *DatabaseContext) (err error
 // internal documents based on key, then checks sync metadata to determine whether document needs to be imported
 func (il *importListener) ProcessFeedEvent(event sgbucket.FeedEvent) (shouldPersistCheckpoint bool) {
 
+	ctx := il.loggingCtx
 	// Ignore non-mutation/deletion events
 	if event.Opcode != sgbucket.FeedOpMutation && event.Opcode != sgbucket.FeedOpDeletion {
 		return true
 	}
-	key := string(event.Key)
+	docID := string(event.Key)
+
+	collection, ok := il.collections[event.CollectionID]
+	if !ok {
+		base.WarnfCtx(ctx, "Received import event for unrecognised collection 0x%x", event.CollectionID)
+		return
+	}
+	ctx = base.CollectionLogCtx(ctx, collection.Name)
 
 	// Ignore internal documents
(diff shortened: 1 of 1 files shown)


db/import_listener.go
```diff
@@ -130,35 +130,43 @@ func (il *importListener) StartImportFeed(dbContext *DatabaseContext) (err error
...
 		return true
 	}
 
-	il.ImportFeedEvent(event)
+	il.ImportFeedEvent(ctx, collection, event)
 	return true
 }
 
-func (il *importListener) ImportFeedEvent(event sgbucket.FeedEvent) {
+func (il *importListener) ImportFeedEvent(ctx context.Context, collection DatabaseCollectionWithUser, event sgbucket.FeedEvent) {
```
bbrks: As you've noted, I think this does need to be a pointer to allow `ReloadUser` to change the user inside this struct - I'm not sure whether this function is called from here or not.

torcolvin: The part that is weird to me is that this was _never_ a pointer and I didn't change this when migrating to collections https://github.com/couchbase/sync_gateway/commit/dd9e90cd2cfdfed0f4568143d9641d751126ef83

torcolvin: I think user doesn't matter because there's no user associated with import feed.

db/import_listener.go
```diff
@@ -130,35 +130,43 @@ func (il *importListener) StartImportFeed(dbContext *DatabaseContext) (err error
...
 	if event.Opcode != sgbucket.FeedOpMutation && event.Opcode != sgbucket.FeedOpDeletion {
 		return true
 	}
-	key := string(event.Key)
+	docID := string(event.Key)
+
+	collection, ok := il.collections[event.CollectionID]
+	if !ok {
+		base.WarnfCtx(ctx, "Received import event for unrecognised collection 0x%x", event.CollectionID)
+		return
```
adamcfraser:
```suggestion
		return true
```
Looks like this should return true to preserve the previous handling.

db/import_listener.go
adamcfraser: Should this be a pointer in any case, to avoid any unwanted copying?

adamcfraser (changes requested):

db/import_listener.go
torcolvin: Made a ptr.

adamcfraser (approved):

PR #6535, opened 2023-10-17
CBG-3268 Create DocVersion to replace RevID in test to make HLV aware

torcolvin:
Creates `DocVersion` to represent a document version or revision. This data structure would store version information. Right now, it is only compromised of RevID, but in the future it would be constructed either with HLV or RevID but not both, enforced at constructor time.

Modified functions to return `DocVersion` instead of custom response or `RevID`. Legacy versions of the functions are left behind to avoid blowing up the scope of this PR, but would be expected to be removed before this ticket is resolved:
- `GetDoc`
- `PutDoc`
- `UpdateDoc` (arguably this function should be `UpsertDoc`)
- `DeleteDoc`

Functions in `RestTester` and `BlipTestClient` are stubbed to work with versions vectors. These currently don't have different behavior and rely on revision IDs:

- `RestTester.WaitForRev` -> `RestTester.WaitForVersion`
- `BlipTesterClient.WaitForRev` -> `BlipTesterClient.WaitForRev` 

Additional modifications:

- modify `RestTester.CreateDoc` -> `RestTester.CreateTestDoc`.
  - use `RestTester.CreateTestDoc` to create a new document when the contents don't matter
  - use `RestTester.PutDoc` to create a new document with specified contents
  
 Future PRs (some sooner than others), which would blow up scope of this PR:
 
 - remove use of the following:
   - `UpdateDocRev`
   - `DeleteDocRev`
   - `WaitForRev`
- evaluate use of rev query parameter and explicit assert on `RevID`. Some cases can't be changed yet (e.g. `atts_since=`) and it should probably be done in parts.
- Make BlipTesterClient more HLV aware and remove other public `Rev` assertions

[] https://jenkins.sgwdev.com/job/SyncGateway-Integration/2120/


rest/utilities_testing_resttester.go
```diff
@@ -162,6 +175,11 @@ func (rt *RestTester) WaitForPullBlipSenderInitialisation(name string) {
 	require.NoError(rt.TB, rt.WaitForCondition(successFunc), "blip sender on active replicator not initialized")
 }
 
+// UsingHLV returns true if using HLV vs revision ID
+func (rt *RestTester) UsingHLV() bool {
```
bbrks: Using HLV is not one or the other - it's RevID or RevID+HLV (we're retaining rev IDs for backwards compatibility and want to continue asserting on those in addition to HLVs when enabled)

bbrks (dismissed):
I'm fine with most of the changes (which are mostly renaming), but the general concept of HLV OR RevID is flawed.

We'll have either RevID only, or RevID+HLV to be asserting on.

rest/utilities_testing.go
```diff
@@ -2310,13 +2308,50 @@ func WaitAndAssertBackgroundManagerExpiredHeartbeat(t testing.TB, bm *db.Backgro
 
 // RespRevID returns a rev ID from the given response, or fails the given test if a rev ID was not found.
 func RespRevID(t *testing.T, response *TestResponse) (revID string) {
+	revision := DocVersionFromPutResponse(t, response)
+	return revision.RevID
+}
+
+// DocVersion represents a specific version of a document in an revID/HLV agnostic manner.
+type DocVersion struct {
+	DocID string
```
adamcfraser: My first instinct would have been to omit DocID from this, and have it just include RevID and (in future) currentVersion.  My feeling is that this actually obfuscates what's going on in a helper function like DocUpdate, and it's better to keep docID and DocVersion as separate parameters to that kind of method.

What do you think the benefits are of including and requiring docID in this struct?

adamcfraser (changes requested):

rest/utilities_testing.go
torcolvin: The benefits were because this is often used together and it makes sense to store it conceptually.

```
doc1Version := rt.PutDoc("doc1", `{"a" : true}`)
rt.WaitForVersion(doc1Version)
doc1Version = rt.UpdateDoc(doc1Version, `{"a" : false}"`)
rt.WaitForVersion(doc1Version)
rt.DeleteDoc(doc1Version)
```

vs:

```
doc1ID := "doc1"
doc1Version := rt.PutDoc(doc1ID, `{"a" : true}`)
rt.WaitForVersion(doc1ID, doc1Version)
doc1Version = rt.UpdateDoc(doc1ID, doc1Version, `{"a" : false}"`)
rt.WaitForVersion(doc1ID, doc1Version)
rt.DeleteDoc(doc1ID, doc1Version)
```

I'm not settled on this way of doing things.

adamcfraser: I prefer the latter.  We need to define docID anyway (for PutDoc in the first example), and I think it keeps the kv API (PutDoc, UpdateDoc, etc) more consistent to always take docID and then an optional version.

rest/attachment_test.go
```diff
@@ -575,7 +575,7 @@ func TestAttachmentsNoCrossTalk(t *testing.T) {
 	}
 
 	// attach to existing document with correct rev (should succeed)
-	response := rt.SendAdminRequestWithHeaders("PUT", "/{{.keyspace}}/doc1/attach1?rev="+doc1revId, attachmentBody, reqHeaders)
+	response := rt.SendAdminRequestWithHeaders("PUT", "/{{.keyspace}}/doc1/attach1?rev="+doc1Version.RevID, attachmentBody, reqHeaders)
```
adamcfraser, line 578: Just to confirm - is an update to the handling on line 583 planned to be included in a follow-up PR?

rest/attachment_test.go
```diff
@@ -2106,13 +2083,12 @@ func TestBasicAttachmentRemoval(t *testing.T) {
...
 		docID := "foo12"
-		revID := rt.CreateDoc(t, docID)
-		require.Equal(t, "1-45ca73d819d5b1c9b8eea95290e79004", revID)
+		version := rt.CreateTestDoc(docID)
 
 		// Add an attachment to the document.
 		att1Name := "alice.txt"
 		att1Body := "this is the body of attachment alice.txt"
-		revID = storeAttachment(t, docID, revID, att1Name, att1Body)
+		revID := storeAttachment(t, docID, version.RevID, att1Name, att1Body)
```
adamcfraser, line 2091: Seems like storeAttachment should return DocVersion, and then use UpdateDoc() on line 2152 below.

rest/blip_api_attachment_test.go
```diff
@@ -590,7 +590,7 @@ func TestBlipLegacyAttachNameChange(t *testing.T) {
 	CreateDocWithLegacyAttachment(t, rt, docID, rawDoc, attKey, attBody)
 
 	// Get the document and grab the revID.
-	responseBody := rt.GetDoc(docID)
+	responseBody := rt.GetDocBody(docID)
```
adamcfraser: Can this be changed to GetDoc, and have a NotEmpty for DocVersion?

rest/blip_api_crud_test.go
```diff
@@ -651,23 +651,21 @@ func TestProposedChangesIncludeConflictingRev(t *testing.T) {
 	// Write existing docs to server directly (not via blip)
 	rt := bt.restTester
 	resp := rt.PutDoc("conflictingInsert", `{"version":1}`)
-	conflictingInsertRev := resp.Rev
+	conflictingInsertRev := resp.RevID
```
adamcfraser, line 654: To be changed to DocVersion in a future PR?

rest/changestest/changes_api_test.go
```diff
@@ -576,48 +577,54 @@ func TestPostChangesAdminChannelGrantRemoval(t *testing.T) {
...
+	abc2Version := rt.PutDoc(abc2ID, `{"channel":["ABC"]}`)
+	abc3Version := rt.PutDoc(abc3ID, `{"channel":["ABC"]}`)
 	cacheWaiter.AddAndWait(13)
 
 	// Update some docs to remove channel
-	_ = rt.PutDoc("pbs-2", fmt.Sprintf(`{"_rev":%q}`, pbs2.Rev))
-	_ = rt.PutDoc("mix-1", fmt.Sprintf(`{"_rev":%q, "channel":["PBS"]}`, mix1.Rev))
-	_ = rt.PutDoc("mix-2", fmt.Sprintf(`{"_rev":%q, "channel":["HBO"]}`, mix2.Rev))
-	_ = rt.PutDoc("mix-4", fmt.Sprintf(`{"_rev":%q}`, mix4.Rev))
+	_ = rt.PutDoc("pbs-2", fmt.Sprintf(`{"_rev":%q}`, pbs2.RevID))
```
adamcfraser: This (and cases below) look like it should change to UpdateDoc and omit _rev in the body.

adamcfraser (changes requested):

rest/attachment_test.go
torcolvin: I'm going to  make further changes to the attachment tests in a separate PR, there are a lot.

rest/attachment_test.go
torcolvin: Agreed, I'm going to change storeAttachment in a separate PR.

rest/blip_api_attachment_test.go
torcolvin: I think this call is not necessary because of the implementation of:

```
func (rt *RestTester) GetDoc(docID string) (body db.Body) {
	rawResponse := rt.SendAdminRequest("GET", "/{{.keyspace}}/"+docID, "")
	RequireStatus(rt.TB, rawResponse, 200)
	require.NoError(rt.TB, base.JSONUnmarshal(rawResponse.Body.Bytes(), &body))
	return body
}
```

I think by definition if this returns 200, it will always have a RevID.

rest/blip_api_crud_test.go
torcolvin: yep, although this is the code that about rev conflicts so it won't be able to be changed fully.

adamcfraser (approved):

PR #6536, opened 2023-10-17
Fixes test without xattrs

torcolvin:
- use RestTester.NewDbConfig() to produce a db config that works with SG_TEST_USE_XATTRS=false
- explicitly use NewRestTesterDefaultCollection so the test is going to in the default setting and actually run regularly

From https://jenkins.sgwdev.com/job/weekly-matrix/BACKING_STORE=enterprise-7.2.0,QUERY_PROVIDER=views,SYNC_STORE=inline,label=sgw-amazon-linux2023/89/

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2121/
- [x] `GSI=false,xattrs=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2118/

### DIFF
--- a/rest/config_database.go
+++ b/rest/config_database.go
@@ -184,7 +184,10 @@ func DefaultDbConfig(sc *StartupConfig, useXattrs bool) *DbConfig {
 			dbConfig.ImportPartitions = base.Uint16Ptr(base.GetDefaultImportPartitions(sc.IsServerless()))
 		}
 
+	} else {
+		dbConfig.AutoImport = base.BoolPtr(false)
 	}
+
 	revsLimit := db.DefaultRevsLimitNoConflicts
 	if dbConfig.AllowConflicts != nil && *dbConfig.AllowConflicts {
 		revsLimit = db.DefaultRevsLimitConflicts
--- a/rest/config_database.go
+++ b/rest/config_database.go
@@ -102,7 +102,7 @@ func DefaultPerDBLogging(bootstrapLoggingCnf base.LoggingConfig) *DbLoggingConfi
 // MergeDatabaseConfigWithDefaults merges the passed in config onto a DefaultDbConfig which results in returned value
 // being populated with defaults when not set
 func MergeDatabaseConfigWithDefaults(sc *StartupConfig, dbConfig *DbConfig) (*DbConfig, error) {
-	defaultDbConfig := DefaultDbConfig(sc)
+	defaultDbConfig := DefaultDbConfig(sc, dbConfig.UseXattrs())
 
 	err := base.ConfigMerge(defaultDbConfig, dbConfig)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


rest/upgradetest/upgrade_registry_test.go
```diff
@@ -190,26 +185,14 @@ func TestLegacyMetadataID(t *testing.T) {
...
-	require.True(t, ok)
-
-	// Need to sanitize the db config, but can't use sanitizeDbConfigs because it assumes non-empty server address
-	dbConfig.Username = ""
-	dbConfig.Password = ""
-	dbConfigBytes, err := base.JSONMarshal(dbConfig)
-
-	require.NoError(t, err)
-	resp = persistentRT.SendAdminRequest("PUT", "/db/", string(dbConfigBytes))
+	resp = persistentRT.CreateDatabase("db", persistentRT.NewDbConfig())
```
adamcfraser: I think the use of ToStartupConfig might be intentional in this test, as it's trying to cover metadataID handling a legacy config is being upgraded.  It feels like using NewDbConfig() might hide potential bugs in the upgrade process?

adamcfraser (changes requested):

rest/config_database.go
```diff
@@ -184,6 +178,13 @@ func DefaultDbConfig(sc *StartupConfig) *DbConfig {
 		Logging:                          DefaultPerDBLogging(sc.Logging),
 	}
 
+	if useXattrs {
+		dbConfig.AutoImport = base.BoolPtr(base.DefaultAutoImport)
+		if base.IsEnterpriseEdition() {
+			dbConfig.ImportPartitions = base.Uint16Ptr(base.GetDefaultImportPartitions(sc.IsServerless()))
```
bbrks, line 184: Setting these conditionally means they'll not be shown with their default values for `include_runtime=true` for a database without xattrs.

This seems wrong? If we want `include_runtime` to show the behaviour of `AutoImportEnabled()` maybe include an `else`

bbrks (changes requested):

rest/config_database.go
torcolvin: The problem that this is trying to fix is that if you do `/_config?include_runtime=true` with a database without xattrs is that it will not be roundtripable because of the way the configs merge:

Before this change:

Default
AutoImport = true
ImportPartitions = 16

Non Xattrs database
AutoImport = nil
ImportPartitions = nil

When you merge those configs it becomes:

AutoImport = true
ImportPartitions = 16

If you try to PUT a new database, you'll trip over https://github.com/couchbase/sync_gateway/blob/ee619160735d0109c568def3ffc8641a606acfbc/rest/config.go#L765

The other way I thought to fix this problem was to explicitly the values when we have a DbConfig before we add it to server context

AutoImport = nil :
  AutoImport = false if non xattrs
  AutoImport = true if xattrs
ImportPartitions = nil
  ImportPartitions = 0 if non xattrs
  ImportPartions = 16 if xattrs
  
If I change this, there's actually a difference between `ImportPartitions = 0` and `ImportPartitions = nil` and I'd have to change https://github.com/couchbase/sync_gateway/blob/ee619160735d0109c568def3ffc8641a606acfbc/rest/config.go#L768

adamcfraser: I agree with Tor's approach here, because it's not valid for those properties to be set when running with xattrs disabled, so the default values shown in include_runtime=true should in fact be nil/omitted.

adamcfraser (dismissed):

rest/config_database.go
bbrks: I think `import_docs` should still default to false with xattrs disabled - it's a valid config combination to set by default, but I'm OK with omitting partitions given the error when set.

torcolvin: Changed this to false.

bbrks (approved):

PR #6543, opened 2023-10-19
CBG-3268 modify attachment_test.go to remove rev

torcolvin:
- Create BlipTester `PushVersion`
- Fix `SendAdminRequest` to call `FixQuotedSlashes` by using `Request` function.
- Removed asserts of `rev` after a simple doc PUT for `PutDoc` or `UpdateDoc`.

Using a search for `[Rr]ev` in `attachment_test.go modify the code to use common functions. This is not complete, or necessarily ideal.

- Etags tests should be parameterized to accept all types of Revisions.
- I wrote helper functions `storeAttachment`, `storeAttachmentWithHeaders` and `storeAttachmentWithIfMatch`, but some helper functions maybe are still necessary:
  - DELETE attachment with a rev ID (wasn't sure what the API will be)
  - storeAttachment where an error condition is expected
- bulk_put / bulk_get where it has `rev` set explicitly, tests that start with `TestBulkGet`
- tests with conflict handling are never going to be supported with HLV, maybe we should move them to a separate file?
- no atts_since REST api yet
- no api for `PUT` doc with `new_edits=false` with HLV, I'm not sure if this is a valid use case. This is pretty much only used in tests with rev conflicts.

My proposal is figure out what future tasks we want to do and put them into separate tickets. The goal of this PR should be to make sure I didn't remove any assertions that are useful. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2129/

### DIFF
--- a/rest/blip_client_test.go
+++ b/rest/blip_client_test.go
@@ -784,15 +784,9 @@ func (btr *BlipTesterReplicator) sendMsg(msg *blip.Message) (err error) {
 
 // PushRev creates a revision on the client, and immediately sends a changes request for it.
 // The rev ID is always: "N-abc", where N is rev generation for predictability.
-func (btc *BlipTesterCollectionClient) PushRev(docID, parentRev string, body []byte) (revID string, err error) {
-	return btc.PushRevWithHistory(docID, parentRev, body, 1, 0)
-}
-
-// PushVersion creates a revision on the client, and immediately sends a changes request for it.
-// The rev ID is always: "N-abc", where N is rev generation for predictability.
-func (btc *BlipTesterCollectionClient) PushVersion(docID string, version DocVersion, body []byte) (DocVersion, error) {
-	revID, err := btc.PushRev(docID, version.RevID, body)
-	return DocVersion{RevID: revID}, err
+func (btc *BlipTesterCollectionClient) PushRev(docID string, parentVersion DocVersion, body []byte) (DocVersion, error) {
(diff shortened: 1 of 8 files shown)


rest/blip_client_test.go
```diff
@@ -788,6 +788,13 @@ func (btc *BlipTesterCollectionClient) PushRev(docID, parentRev string, body []b
 	return btc.PushRevWithHistory(docID, parentRev, body, 1, 0)
 }
 
+// PushVersion creates a revision on the client, and immediately sends a changes request for it.
+// The rev ID is always: "N-abc", where N is rev generation for predictability.
+func (btc *BlipTesterCollectionClient) PushVersion(docID string, version DocVersion, body []byte) (DocVersion, error) {
```
adamcfraser: I think the 'Rev' in PushRev refers to the 'rev' message from the replication protocol, which isn't going to be changing.  So I think a better change would be to update the signature of btc.PushRev to take DocVersion instead of parentRev, but leave the naming intact.

adamcfraser (commented):
One proposed change on PushVersion.

On the new_edits=false question - that's the REST API used for a client to specify the revID (instead of having SG generate it).  The same concept exists for HLV versions, but you're correct that this may only be supported for 2.x replication clients.

adamcfraser (approved):

PR #6546, opened 2023-10-20
CBG-3553 add bucket.Close() when opening the database

torcolvin:
- add dbcontext.Close if there's a failure after dbContext is open
- modify non persistent config PUT /db/ config code to return 500 to match error codes on error for persistent config (previously returned 400)

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2132/ (minus failure fixed in https://github.com/couchbase/sync_gateway/pull/6546/commits/760ec92fd791c86938e3012196d794855a624268)

### DIFF
--- a/db/sequence_allocator.go
+++ b/db/sequence_allocator.go
@@ -71,11 +71,14 @@ func newSequenceAllocator(ctx context.Context, datastore base.DataStore, dbStats
 
 	// The reserveNotify channel manages communication between the releaseSequenceMonitor goroutine and _reserveSequenceRange invocations.
 	s.reserveNotify = make(chan struct{}, 1)
+	_, err := s.lastSequence(ctx) // just reads latest sequence from bucket
+	if err != nil {
+		return nil, err
+	}
 	go func() {
 		defer base.FatalPanicHandler()
 		s.releaseSequenceMonitor(ctx)
 	}()
-	_, err := s.lastSequence(ctx) // just reads latest sequence from bucket
 	return s, err
 }
 
--- a/db/sequence_allocator.go
+++ b/db/sequence_allocator.go
@@ -155,7 +155,7 @@ func (s *sequenceAllocator) lastSequence(ctx context.Context) (uint64, error) {
 	s.dbStats.SequenceGetCount.Add(1)
 	last, err := s.getSequence()
 	if err != nil {
-		base.WarnfCtx(ctx, "Error from Get in getSequence(): %v", err)
+		return 0, fmt.Errorf("Couldn't get sequence from bucket: %w", err)
 	}
 	return last, err
 }
--- a/rest/server_context.go
+++ b/rest/server_context.go
@@ -512,13 +512,18 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
 		dbName = spec.BucketName
 	}
 	defer func() {
-		if returnedError != nil {
-			_, dbRegistered := sc.databases_[dbName]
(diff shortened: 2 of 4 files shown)


db/sequence_allocator.go
bbrks, line 1: It'd be nice if we didn't lose the error's context from `sequenceAllocator.lastSequence()`

e.g.

```diff
diff --git a/db/sequence_allocator.go b/db/sequence_allocator.go
index b85ef326..1da14c4e 100644
--- a/db/sequence_allocator.go
+++ b/db/sequence_allocator.go
@@ -155,9 +155,9 @@ func (s *sequenceAllocator) lastSequence(ctx context.Context) (uint64, error) {
 	s.dbStats.SequenceGetCount.Add(1)
 	last, err := s.getSequence()
 	if err != nil {
-		base.WarnfCtx(ctx, "Error from Get in getSequence(): %v", err)
+		return 0, fmt.Errorf("Couldn't get sequence from bucket: %w", err)
 	}
-	return last, err
+	return last, nil
 }

 // Returns the next available sequence.
```

Give or take dropping the warning log... if the returned error contains the info I'm not sure it's necessary.

Before

```
2023-10-23T17:11:36.496+01:00 [WRN] db:testdb Error from Get in getSequence(): json: cannot unmarshal string into Go value of type uint64 -- db.(*sequenceAllocator).lastSequence() at sequence_allocator.go:158
2023-10-23T17:11:36.496+01:00 [ERR] c:#002 #002: 500 couldn't load database: json: cannot unmarshal string into Go value of type uint64 -- rest.(*handler).writeError() at handler.go:1466
2023-10-23T17:11:36.496+01:00 [INF] HTTP: c:#002 #002:     --> 500 couldn't load database: json: cannot unmarshal string into Go value of type uint64  (4.6 ms)
```

After

```
2023-10-23T17:06:58.933+01:00 [ERR] c:#002 #002: 500 couldn't load database: Couldn't get sequence from bucket: json: cannot unmarshal string into Go value of type uint64 -- rest.(*handler).writeError() at handler.go:1466
2023-10-23T17:06:58.933+01:00 [INF] HTTP: c:#002 #002:     --> 500 couldn't load database: Couldn't get sequence from bucket: json: cannot unmarshal string into Go value of type uint64  (3.6 ms)
```

bbrks (changes requested):
I'm good with these changes.
Maybe one more that would've sped up the CBSE investigation?

db/sequence_allocator.go
torcolvin: I'd leave it because this will happen from config polling and it is our opportunity to see this. I tested with PUT but the circumstances that it is a problem is when it is loaded every 10s by config polling.

rest/server_context.go
```diff
@@ -509,6 +511,17 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
 	if dbName == "" {
 		dbName = spec.BucketName
 	}
+	defer func() {
+		if returnedError != nil {
+			_, dbRegistered := sc.databases_[dbName]
+			if !dbRegistered && dbcontext != nil {
+				dbcontext.Close(ctx) // will close bucket
+			} else if bucket != nil {
+				bucket.Close(ctx)
```
adamcfraser: In the case where dbRegistered==true, I don't think it's safe to close the bucket either, is it?

adamcfraser (changes requested):

adamcfraser (dismissed):

PR #6554, opened 2023-10-24
CBG-3268 replace WaitForRev with WaitForVersion

torcolvin:
Replaces `WaitForRev` with `WaitForVersion` to remove rev id assertions. When reviewing this:

- Look for dropped conceptual assertions
- Look for unintentional or unwaited drops in whether the request is user or admin scoped (`PutDoc`, `UpdateDoc`, `DeleteDoc` are admin scoped.
- Make sure there are no dropped `WaitForPendingChanges` or other things that can cause race conditions.

Fails lint CI due to https://github.com/couchbase/sync_gateway/pull/6555

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2138/

### DIFF
--- a/rest/blip_api_collections_test.go
+++ b/rest/blip_api_collections_test.go
@@ -243,9 +243,8 @@ func TestCollectionsReplication(t *testing.T) {
 	require.NoError(t, err)
 	defer btc.Close()
 
-	resp := rt.SendAdminRequest(http.MethodPut, "/{{.keyspace}}/doc1", "{}")
-	RequireStatus(t, resp, http.StatusCreated)
-
+	const docID = "doc1"
+	version := rt.PutDoc(docID, "{}")
 	require.NoError(t, rt.WaitForPendingChanges())
 
 	btcCollection := btc.SingleCollection()
@@ -253,7 +252,7 @@ func TestCollectionsReplication(t *testing.T) {
 	err = btcCollection.StartOneshotPull()
 	require.NoError(t, err)
 
-	_, ok := btcCollection.WaitForRev("doc1", "1-ca9ad22802b66f662ff171f226211d5c")
+	_, ok := btcCollection.WaitForVersion(docID, version)
 	require.True(t, ok)
 }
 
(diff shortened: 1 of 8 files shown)


rest/blip_api_collections_test.go
```diff
@@ -267,28 +266,26 @@ func TestBlipReplicationMultipleCollections(t *testing.T) {
...
-	for _, collection := range rt.GetDatabase().CollectionByID {
-		collectionWithAdmin := db.DatabaseCollectionWithUser{DatabaseCollection: collection}
-		revID, _, err := collectionWithAdmin.Put(base.TestCtx(t), docName, body)
-		require.NoError(t, err)
-		require.Equal(t, docRevID, revID)
+	body := `{"foo":"bar"}`
+	for _, keyspace := range rt.GetKeyspaces() {
+		resp := rt.SendAdminRequest(http.MethodPut, "/"+keyspace+"/"+docName, `{"foo":"bar"}`)
+		RequireStatus(t, resp, http.StatusCreated)
+		version = DocVersionFromPutResponse(t, resp)
```
bbrks: Think this could potentially begin to break with version vectors, since it's not nessisarily the case that each HLV would be the same for the same doc body - if a component of the version was a timestamp.

We're currently relying on the general idea that `version=hash(body)` in the test for line 286

rest/blip_api_crud_test.go
```diff
@@ -1847,25 +1847,24 @@ func TestBlipPullRevMessageHistory(t *testing.T) {
...
 
-	data, ok = client.WaitForRev("doc1", "2-26359894b20d89c97638e71c40482f28")
+	data, ok = client.WaitForVersion(docID, version2)
 	assert.True(t, ok)
 	assert.Equal(t, `{"greetings":[{"hello":"world!"},{"hi":"alice"},{"howdy":12345678901234567890}]}`, string(data))
 
 	msg, ok := client.pullReplication.WaitForMessage(5)
 	assert.True(t, ok)
-	assert.Equal(t, "1-0335a345b6ffed05707ccc4cbc1b67f4", msg.Properties[db.RevMessageHistory])
+	assert.Equal(t, version1.RevID, msg.Properties[db.RevMessageHistory]) // CBG-3268 update to use version
```
bbrks, line 1867: Would need blip test client support for running in a VV-only mode and determining replication protocol changes to see if there's an equivalent.

Blocked behind much later work - can't reasonably be done in this ticket.

rest/revocation_test.go
```diff
@@ -238,11 +226,11 @@ func TestRevocationScenario1(t *testing.T) {
 	assert.Equal(t, "25", changes.Last_Seq)
 
 	revocationTester.fillToSeq(44)
-	revocationTester.removeRole("user", "foo")
+	revocationTester.removeFooRole()
 	revocationTester.fillToSeq(54)
 	revocationTester.removeRoleChannel("foo", "ch1")
```
bbrks: This seems like only a half-complete refactor/simplification of revocationTester, and makes the API of it more inconsistent than it was.

E.g:
- `add`/`removeRoleChannel("foo", "ch1")` remains unchanged
- `addRole("foo")` vs. `removeFooRole()`

Is this nessesary for the HLV work?

bbrks (changes requested):
Mostly ok - some inline comments

rest/revocation_test.go
torcolvin: Ripped this out.

rest/blip_api_crud_test.go
torcolvin: There's some work that wasn't done here, presumably this will fail in test now - but some tests should be extended. The thing about this test is that it's only testing for the presence of `RevMessageHistory` and this might not be present in VV-only mode.

rest/blip_api_collections_test.go
torcolvin: Updated to fix this to look for specific versions.

bbrks (approved):

PR #6556, opened 2023-10-25
CBG-1966 enable staticcheck

torcolvin:
- remove kvPoolSize empty check (setup in GetGoCBConnString
- replace math/rand with crypto/rand since math/rand.Seed is deprecated since 1.20 for crypto/rand.Seed
- remove random seed on init since go 1.20, which is now builtin, rather than Seed(1)
- remove potential nil checks after we acquire locks in ActiveReplicator
- remove explicit set to benchmark
- check error before defer foo.Close()

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2141/

### DIFF
--- a/auth/oidc_test.go
+++ b/auth/oidc_test.go
@@ -1255,7 +1255,6 @@ func TestJWTRolesChannels(t *testing.T) {
 				require.Equal(t, base.SetFromArray(login.expectedChannels), user.Channels().AsSet())
 
 				require.Greater(t, user.JWTLastUpdated(), lastUpdateTime)
-				lastUpdateTime = user.JWTLastUpdated()
 			}
 		})
 	}
--- a/auth/oidc_test.go
+++ b/auth/oidc_test.go
@@ -1205,9 +1205,8 @@ func TestJWTRolesChannels(t *testing.T) {
 
 			for i, login := range tc.logins {
 				var (
-					user           User
-					err            error
-					lastUpdateTime time.Time
+					user User
+					err  error
 				)
 				if i == 0 {
 					user, err = auth.NewUser(testUserPrefix+"_"+testSubject, "test", base.SetFromArray(login.explicitChannels))
@@ -1254,7 +1253,6 @@ func TestJWTRolesChannels(t *testing.T) {
 				}
 				require.Equal(t, base.SetFromArray(login.expectedChannels), user.Channels().AsSet())
 
-				require.Greater(t, user.JWTLastUpdated(), lastUpdateTime)
 			}
 		})
 	}
--- a/auth/password_hash_test.go
+++ b/auth/password_hash_test.go
@@ -35,7 +35,6 @@ func BenchmarkBcryptCostTimes(b *testing.B) {
 
 	for i := minCostToTest; i < maxCostToTest; i++ {
 		b.Run(fmt.Sprintf("cost%d", i), func(bn *testing.B) {
-			bn.N = 1
 			_, err := bcrypt.GenerateFromPassword([]byte("hunter2"), i)
 			assert.NoError(bn, err)
 		})
--- a/auth/password_hash_test.go
+++ b/auth/password_hash_test.go
@@ -41,20 +41,24 @@ func BenchmarkBcryptCostTimes(b *testing.B) {
 	}
 }
 
-// TestBcryptDefaultCostTime will ensure that the default bcrypt cost takes at least a 'reasonable' amount of time
-// If this test fails, it suggests maybe we need to think about increasing the default cost...
-func TestBcryptDefaultCostTime(t *testing.T) {
-	// Modest 2.2GHz macbook i7 takes ~80ms at cost 10
-	// Assume server CPUs are ~2x faster
-	minimumDuration := 40 * time.Millisecond
+// TestBcryptCostTimes will output the time it takes to hash a password with each bcrypt cost value
+func TestBcryptCostTimes(t *testing.T) {
+	// Little value in running this regularly. Might be useful for one-off informational purposes
+	t.Skip("Test disabled")
 
-	startTime := time.Now()
-	_, err := bcrypt.GenerateFromPassword([]byte("hunter2"), DefaultBcryptCost)
-	duration := time.Since(startTime)
+	minCostToTest := bcrypt.DefaultCost
+	maxCostToTest := bcrypt.DefaultCost + 5
 
-	t.Logf("bcrypt.GenerateFromPassword with cost %d took: %v", DefaultBcryptCost, duration)
-	assert.NoError(t, err)
-	assert.True(t, minimumDuration < duration)
+	for i := minCostToTest; i < maxCostToTest; i++ {
+		t.Run(fmt.Sprintf("cost%d", i), func(t *testing.T) {
+			startTime := time.Now()
+			_, err := bcrypt.GenerateFromPassword([]byte("hunter2"), i)
+			duration := time.Since(startTime)
+
+			t.Logf("bcrypt.GenerateFromPassword with cost %d took: %v", i, duration)
+			assert.NoError(t, err)
+		})
+	}
 }
 
 func TestSetBcryptCost(t *testing.T) {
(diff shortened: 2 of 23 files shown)


auth/oidc_test.go
```diff
@@ -1255,7 +1255,6 @@ func TestJWTRolesChannels(t *testing.T) {
 				require.Equal(t, base.SetFromArray(login.expectedChannels), user.Channels().AsSet())
 
 				require.Greater(t, user.JWTLastUpdated(), lastUpdateTime)
-				lastUpdateTime = user.JWTLastUpdated()
```
bbrks, line 1258: This doesn't quite look right - comparing greater than `0` seems unintentional.

auth/password_hash_test.go
```diff
@@ -35,7 +35,6 @@ func BenchmarkBcryptCostTimes(b *testing.B) {
 
 	for i := minCostToTest; i < maxCostToTest; i++ {
 		b.Run(fmt.Sprintf("cost%d", i), func(bn *testing.B) {
```
bbrks, line 37: Benchmark looks wrong as-is anyway (no b.N loop)

Maybe we should just swap this to a test that logs the values, as we get the same one-off benefit even without assertions.

```go
// TestBcryptCostTimes will output the time it takes to hash a password with each bcrypt cost value
func TestBcryptCostTimes(t *testing.T) {
	// Little value in running this regularly. Might be useful for one-off informational purposes
	t.Skip("Test disabled")

	minCostToTest := bcrypt.DefaultCost
	maxCostToTest := bcrypt.DefaultCost + 5

	for i := minCostToTest; i < maxCostToTest; i++ {
		t.Run(fmt.Sprintf("cost%d", i), func(t *testing.T) {
			startTime := time.Now()
			_, err := bcrypt.GenerateFromPassword([]byte("hunter2"), i)
			duration := time.Since(startTime)

			t.Logf("bcrypt.GenerateFromPassword with cost %d took: %v", i, duration)
			assert.NoError(t, err)
		})
	}
}
```

```
=== RUN   TestBcryptCostTimes
--- PASS: TestBcryptCostTimes (2.11s)
=== RUN   TestBcryptCostTimes/cost10
    password_hash_test.go:42: bcrypt.GenerateFromPassword with cost 10 took: 68.808833ms
    --- PASS: TestBcryptCostTimes/cost10 (0.07s)
=== RUN   TestBcryptCostTimes/cost11
    password_hash_test.go:42: bcrypt.GenerateFromPassword with cost 11 took: 136.465541ms
    --- PASS: TestBcryptCostTimes/cost11 (0.14s)
=== RUN   TestBcryptCostTimes/cost12
    password_hash_test.go:42: bcrypt.GenerateFromPassword with cost 12 took: 271.3435ms
    --- PASS: TestBcryptCostTimes/cost12 (0.27s)
=== RUN   TestBcryptCostTimes/cost13
    password_hash_test.go:42: bcrypt.GenerateFromPassword with cost 13 took: 546.19325ms
    --- PASS: TestBcryptCostTimes/cost13 (0.55s)
=== RUN   TestBcryptCostTimes/cost14
    password_hash_test.go:42: bcrypt.GenerateFromPassword with cost 14 took: 1.089720583s
    --- PASS: TestBcryptCostTimes/cost14 (1.09s)
PASS
```

base/collection.go
```diff
@@ -63,10 +63,6 @@ func GetGoCBv2Bucket(ctx context.Context, spec BucketSpec) (*GocbV2Bucket, error
 		RetryStrategy:  gocb.NewBestEffortRetryStrategy(nil),
 	}
 
-	if spec.KvPoolSize > 0 {
-		// TODO: Equivalent of kvPoolSize in gocb v2?
-	}
```
bbrks, line 68: ~Are we actually OK removing this? Or should we file a ticket if we think it's not implemented?~

I read the PR description highlighting `GetGoCBConnString`. All good.

db/active_replicator_pull.go
```diff
@@ -40,10 +40,6 @@ func (apr *ActivePullReplicator) Start(ctx context.Context) error {
 	apr.lock.Lock()
 	defer apr.lock.Unlock()
 
-	if apr == nil {
-		return fmt.Errorf("nil ActivePullReplicator, can't start")
-	}
```
bbrks, line 45: Wouldn't a more appropriate fix be to move this to above the lock, or is it guaranteed that `apr` can't be nil?

db/active_replicator_pull.go
```diff
@@ -158,10 +154,6 @@ func (apr *ActivePullReplicator) _subChanges(collectionIdx *int, since string) e
 func (apr *ActivePullReplicator) Complete() {
 	base.TracefCtx(apr.ctx, base.KeyReplicate, "ActivePullReplicator.Complete()")
 	apr.lock.Lock()
-	if apr == nil {
-		apr.lock.Unlock()
-		return
-	}
```
bbrks, line 164: Similarly, I'm not sure of the original motivation for this guard? Maybe being overly defensive?

db/attachment_test.go
```diff
@@ -10,12 +10,12 @@ package db
 
 import (
 	"context"
+	"crypto/rand"
```
bbrks: This file/`TestLargeAttachments` doesn't seem to _need_ cryptographic randomness? What's the motivation for the change here?

db/crud.go
```diff
@@ -413,10 +413,10 @@ func (db *DatabaseCollectionWithUser) GetDelta(ctx context.Context, docID, fromR
...
 			db.dbStats().DeltaSync().DeltaCacheHit.Add(1)
 			return fromRevision.Delta, nil, nil
-		} else {
-			// TODO: Recurse and merge deltas when gen(revCacheDelta.toRevID) < gen(toRevId)
-			// until then, fall through to generating delta for given rev pair
-		}
+		} // else {
+		// TODO: Recurse and merge deltas when gen(revCacheDelta.toRevID) < gen(toRevId)
+		// until then, fall through to generating delta for given rev pair
+		//	}
```
bbrks: I'd be OK removing this completely. Storage/memory vs. CPU tradeoff.

rest/multipart_test.go
```diff
@@ -163,7 +164,8 @@ func TestWriteJSONPart(t *testing.T) {
 	// a body larger than 300 bytes to test writeJSONPart with compression=true and compression=false
 	mockFakeBody := func() db.Body {
 		bytes := make([]byte, 139)
-		rand.Read(bytes)
+		_, err := rand.Read(bytes)
+		require.NoError(t, err)
```
bbrks: Doesn't need crypto rand

bbrks (changes requested):
This one has some questionable changes/lint warnings. See inline for details.

db/active_replicator_pull.go
torcolvin: I think because the code works as is, that this was just overly defensive programming and broken in a refactor.

db/active_replicator_pull.go
torcolvin: I think this was overly defensive programming, and is not hit currently.

db/attachment_test.go
torcolvin: https://github.com/golang/go/issues/20661 officially deprecates `math/rand` for `crypto/rand`. The answer seems to be that people are worried that you'll use the wrong one by mistake.

I don't know that we need cryptographic randomness anywhere, but I could see making this mistake. Most tools (goimports, etc) will add cryptographic rand anyway for new code.

We could make an exception for this because I don't think that we currently need cryptographic randomness anywhere, but I don't think this is necessary.

auth/oidc_test.go
torcolvin: I think this is right, because it is checking that this value got updated - are you suggesting I remove this check, or move `lastUpdated` time lower than `user.SetJWTLastUpdated(time.Now())` and use `AssertTimestampGreaterThan`?

torcolvin: Nevermind, this doesn't get updated.

db/attachment_test.go
bbrks: Looks like the intention is to switch to a locally scoped `math.Rand.Read()` if you don't need cryptographically secure randomness. We should do that here

bbrks (approved):

PR #6558, opened 2023-10-26
Drop changeCache.updateStats() lock to read lock

bbrks:
The method doesn't write to any fields on the struct. Doesn't need to hold the full lock.

Avoids contention with the read locks in `getNextSequence()`/`getInitialSequence()`.

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2143/
- [ ] `-race` `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2144/

### DIFF
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -81,8 +81,8 @@ type changeCacheStats struct {
 
 func (c *changeCache) updateStats(ctx context.Context) {
 
-	c.lock.Lock()
-	defer c.lock.Unlock()
+	c.lock.RLock()
+	defer c.lock.RUnlock()
 	if c.db == nil {
 		return
 	}


db/change_cache.go
```diff
@@ -81,8 +81,8 @@ type changeCacheStats struct {
 
 func (c *changeCache) updateStats(ctx context.Context) {
 
-	c.lock.Lock()
-	defer c.lock.Unlock()
+	c.lock.RLock()
+	defer c.lock.RUnlock()
```
torcolvin, line 85: Does `c._getMaxStableSequence()` require a lock because it touches `nextSequence`?

bbrks: It does, but read only

torcolvin: Ah, brain mishap, looks good.

torcolvin (approved):

PR #6564, opened 2023-11-03
CBG-3585 log bucket and groupID during config search

torcolvin:
This makes it a bit more clear which bucket a config error might come from. I added groupID to these logs as well, since they are logged by some lines in these functions but not all.

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -1467,11 +1467,11 @@ func (sc *ServerContext) migrateV30Configs(ctx context.Context) error {
 			continue
 		}
 
-		base.InfofCtx(ctx, base.KeyConfig, "Found legacy persisted config for database %s - migrating to db registry.", base.MD(dbConfig.Name))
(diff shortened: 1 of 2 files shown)


bbrks (approved):

adamcfraser:
Generally speaking logging groupID is redundant information, since a given node (and it's logs) is only associated with a single groupID per startup, and the groupID is logged at startup.  Given the particular relevance of groupID to the registry I'm fine leaving it here for ease of use, but as a rule I still think we should omit.

torcolvin:
> Generally speaking logging groupID is redundant information, since a given node (and it's logs) is only associated with a single groupID per startup, and the groupID is logged at startup. Given the particular relevance of groupID to the registry I'm fine leaving it here for ease of use, but as a rule I still think we should omit.

I alway felt groupID was redundant but it was present in other log lines in the same function. The important thing that was left out was the bucket. OTOH, if you are trying to debug issues here having groupID again is helpful.

PR #6567, opened 2023-11-07
CBG-3576: BlipTestClient support for HLV and rev tree modes

gregns1:
CBG-3576

- Added a new struct called BlipTestClientRunner that will be used to run the blip client subtests
- A Method Run() on this type that will run each tests code twice, one for rev tree handling, then (when support is added) for version vector protocol. It does this by passing down the subprotocol version into the test code to be passed into the BlipTesterClient creation. Some tests specifically test certain subprotocol versions, so in these cases the passed down subprotocol version from the Run() function is ignored. 
- Moved the method NewBlipTesterClientWithRT method to run on the BlipTestClientRunner type. So the user cannot call this function without first initialising a BlipTestClientRunner type.
- On BlipTestClientRunner there is a flag to indicate that the Run() method is being used, this allows us to force people to only use this method to create a BlipTesterClient from inside the run code. If a user tries to create a BlipTesterClient outside the Run() method the test will fail immediately.
- Altered method that originally acted upon the BlipTesterClient directly (such as PushRev) to act upon the BlipTestClientRunner type. This along side the the fact you cannot call NewBlipTesterClientWithRT from outside the Run() method, add a lot of protection to force anyone writing BlipTestClient tests to use the Run() method. 
- On the above point, we have tests that use multiple BlipTestClients, so in order to move the functions such as PushRev to act upon BlipTestClientRunner I needed a way to add all these clients to the new type. Thus on BlipTestClientRunner I have a map of clients that are keyed by a unique id, that is assigned to the BlipTesterClient during creation inside NewBlipTesterClientWithRT, this client id is then passed into functions that act upon the BlipTesterClient. Again methods such as PushRev. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2166/

### DIFF
--- a/rest/attachment_test.go
+++ b/rest/attachment_test.go
@@ -1060,6 +1060,7 @@ func TestAttachmentContentType(t *testing.T) {
 }
 
 func TestBasicAttachmentRemoval(t *testing.T) {
+	t.Skip("Disabled pending CBG-3503")
 	rt := NewRestTester(t, &RestTesterConfig{GuestEnabled: true})
 	defer rt.Close()
 
@@ -2223,6 +2224,7 @@ func TestAttachmentDeleteOnPurge(t *testing.T) {
 }
 
 func TestAttachmentDeleteOnExpiry(t *testing.T) {
+	t.Skip("Disabled pending CBG-3503")
 
 	rt := NewRestTester(t, nil)
 	defer rt.Close()
--- a/rest/attachment_test.go
+++ b/rest/attachment_test.go
@@ -2262,184 +2262,188 @@ func TestAttachmentDeleteOnExpiry(t *testing.T) {
 
 }
 func TestUpdateExistingAttachment(t *testing.T) {
-	rt := NewRestTester(t, &RestTesterConfig{
+	rtConfig := &RestTesterConfig{
 		GuestEnabled: true,
-	})
-	defer rt.Close()
+	}
 
-	btc, err := NewBlipTesterClientOptsWithRT(t, rt, nil)
-	require.NoError(t, err)
+	btc := NewBlipTesterClientOptsWithRT(nil)
 	defer btc.Close()
 
 	const (
 		doc1ID = "doc1"
 		doc2ID = "doc2"
 	)
-	doc1Version := rt.PutDoc(doc1ID, `{}`)
-	doc2Version := rt.PutDoc(doc2ID, `{}`)
 
-	require.NoError(t, rt.WaitForPendingChanges())
+	btc.Run(func(t *testing.T) {
+		// Add doc1 and doc2
+		doc1Version := rt.PutDoc(doc1ID, `{}`)
+		doc2Version := rt.PutDoc(doc2ID, `{}`)
 
-	err = btc.StartOneshotPull()
-	assert.NoError(t, err)
-	_, ok := btc.WaitForVersion(doc1ID, doc1Version)
-	require.True(t, ok)
-	_, ok = btc.WaitForVersion(doc2ID, doc2Version)
-	require.True(t, ok)
+		require.NoError(t, btc.rt.WaitForPendingChanges())
 
-	attachmentAData := base64.StdEncoding.EncodeToString([]byte("attachmentA"))
-	attachmentBData := base64.StdEncoding.EncodeToString([]byte("attachmentB"))
+		err := btc.StartOneshotPull()
+		assert.NoError(t, err)
+		_, ok := btc.WaitForVersion(doc1ID, doc1Version)
+		require.True(t, ok)
+		_, ok = btc.WaitForVersion(doc2ID, doc2Version)
+		require.True(t, ok)
 
-	doc1Version, err = btc.PushRev(doc1ID, doc1Version, []byte(`{"key": "val", "_attachments": {"attachment": {"data": "`+attachmentAData+`"}}}`))
-	require.NoError(t, err)
-	doc2Version, err = btc.PushRev(doc2ID, doc2Version, []byte(`{"key": "val", "_attachments": {"attachment": {"data": "`+attachmentBData+`"}}}`))
-	require.NoError(t, err)
+		attachmentAData := base64.StdEncoding.EncodeToString([]byte("attachmentA"))
+		attachmentBData := base64.StdEncoding.EncodeToString([]byte("attachmentB"))
 
-	assert.NoError(t, rt.WaitForVersion(doc1ID, doc1Version))
-	assert.NoError(t, rt.WaitForVersion(doc2ID, doc2Version))
+		doc1Version, err = btc.PushRev("doc1", doc1Version, []byte(`{"key": "val", "_attachments": {"attachment": {"data": "`+attachmentAData+`"}}}`))
+		require.NoError(t, err)
+		doc2Version, err = btc.PushRev("doc2", doc2Version, []byte(`{"key": "val", "_attachments": {"attachment": {"data": "`+attachmentBData+`"}}}`))
+		require.NoError(t, err)
 
-	_, err = rt.GetSingleTestDatabaseCollection().GetDocument(base.TestCtx(t), "doc1", db.DocUnmarshalAll)
-	require.NoError(t, err)
-	_, err = rt.GetSingleTestDatabaseCollection().GetDocument(base.TestCtx(t), "doc2", db.DocUnmarshalAll)
-	require.NoError(t, err)
+		assert.NoError(t, rt.WaitForVersion(doc1ID, doc1Version))
+		assert.NoError(t, rt.WaitForVersion(doc2ID, doc2Version))
 
-	doc1Version, err = btc.PushRev(doc1ID, doc1Version, []byte(`{"key": "val", "_attachments":{"attachment":{"digest":"sha1-SKk0IV40XSHW37d3H0xpv2+z9Ck=","length":11,"content_type":"","stub":true,"revpos":3}}}`))
-	require.NoError(t, err)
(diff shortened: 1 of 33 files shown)


rest/blip_client_test.go
```diff
@@ -571,74 +573,194 @@ func getCollectionsForBLIP(_ testing.TB, rt *RestTester) []string {
...
+		btc.nonCollectionAwareClient.Close()
+	}
+}
+
+func (btc *BlipTesterClient) RunDualClients(btc2 *BlipTesterClient, test func(t *testing.T), t *testing.T, rtConfig *RestTesterConfig) {
+	rt := NewRestTester(t, rtConfig)
+	btc.rt = rt
+	btc2.rt = rt
+	parentT := btc.rt.TB.(*testing.T)
+	parentT.Run("revTree", func(t *testing.T) {
```
torcolvin: You can use `rt.Run` here and other places.

rest/blip_client_test.go
```diff
@@ -571,74 +573,194 @@ func getCollectionsForBLIP(_ testing.TB, rt *RestTester) []string {
...
+	for _, collectionClient := range btc.collectionClients {
+		collectionClient.Close()
+	}
+	if btc.nonCollectionAwareClient != nil {
+		btc.nonCollectionAwareClient.Close()
+	}
+}
+
+func (btc *BlipTesterClient) RunDualClients(btc2 *BlipTesterClient, test func(t *testing.T), t *testing.T, rtConfig *RestTesterConfig) {
+	rt := NewRestTester(t, rtConfig)
```
torcolvin: I'd name this `RunHLVAndRev`

gregns1: I don't think that names makes sense here. This particular method is for test that run dual BlipTestClients such as TestBlipDeltaSyncPullTombstonedStarChan. All these methods run BlipTesterClients Tests in rev and VV mode unless SkipVersionVectorInitialization is set.

bbrks: This method doesn't do HLV + Rev testing.

rest/attachment_test.go
```diff
@@ -2262,184 +2262,188 @@ func TestAttachmentDeleteOnExpiry(t *testing.T) {
...
-	doc1Version, err = btc.PushRev(doc1ID, doc1Version, []byte(`{"key": "val", "_attachments": {"attachment": {"data": "`+attachmentAData+`"}}}`))
-	require.NoError(t, err)
-	doc2Version, err = btc.PushRev(doc2ID, doc2Version, []byte(`{"key": "val", "_attachments": {"attachment": {"data": "`+attachmentBData+`"}}}`))
-	require.NoError(t, err)
+		attachmentAData := base64.StdEncoding.EncodeToString([]byte("attachmentA"))
+		attachmentBData := base64.StdEncoding.EncodeToString([]byte("attachmentB"))
 
-	assert.NoError(t, rt.WaitForVersion(doc1ID, doc1Version))
-	assert.NoError(t, rt.WaitForVersion(doc2ID, doc2Version))
+		doc1Version, err = btc.PushRev("doc1", doc1Version, []byte(`{"key": "val", "_attachments": {"attachment": {"data": "`+attachmentAData+`"}}}`))
```
bbrks:
```suggestion
		doc1Version, err = btc.PushRev(doc1ID, doc1Version, []byte(`{"key": "val", "_attachments": {"attachment": {"data": "`+attachmentAData+`"}}}`))
```

rest/attachment_test.go
```diff
@@ -2262,184 +2262,188 @@ func TestAttachmentDeleteOnExpiry(t *testing.T) {
...
+		require.NoError(t, err)
+		doc2Version, err = btc.PushRev("doc2", doc2Version, []byte(`{"key": "val", "_attachments": {"attachment": {"data": "`+attachmentBData+`"}}}`))
```
bbrks:
```suggestion
		doc2Version, err = btc.PushRev(doc2ID, doc2Version, []byte(`{"key": "val", "_attachments": {"attachment": {"data": "`+attachmentBData+`"}}}`))
```

rest/attachment_test.go
```diff
@@ -2262,184 +2262,188 @@ func TestAttachmentDeleteOnExpiry(t *testing.T) {
...
+		assert.NoError(t, btc.rt.WaitForVersion(doc2ID, doc2Version))
 
-	doc1Version, err = btc.PushRev(doc1ID, doc1Version, []byte(`{"key": "val", "_attachments":{"attachment":{"digest":"sha1-SKk0IV40XSHW37d3H0xpv2+z9Ck=","length":11,"content_type":"","stub":true,"revpos":3}}}`))
-	require.NoError(t, err)
+		_, err = btc.rt.GetSingleTestDatabaseCollection().GetDocument(base.TestCtx(t), "doc1", db.DocUnmarshalAll)
+		require.NoError(t, err)
+		_, err = btc.rt.GetSingleTestDatabaseCollection().GetDocument(base.TestCtx(t), "doc2", db.DocUnmarshalAll)
+		require.NoError(t, err)
+
+		doc1Version, err = btc.PushRev("doc1", doc1Version, []byte(`{"key": "val", "_attachments":{"attachment":{"digest":"sha1-SKk0IV40XSHW37d3H0xpv2+z9Ck=","length":11,"content_type":"","stub":true,"revpos":3}}}`))
```
bbrks:
```suggestion
		doc1Version, err = btc.PushRev(doc1ID, doc1Version, []byte(`{"key": "val", "_attachments":{"attachment":{"digest":"sha1-SKk0IV40XSHW37d3H0xpv2+z9Ck=","length":11,"content_type":"","stub":true,"revpos":3}}}`))
```

rest/blip_api_collections_test.go
```diff
@@ -28,322 +28,321 @@ func TestBlipGetCollections(t *testing.T) {
...
+	rtConfig := &RestTesterConfig{
 		GuestEnabled: true,
-	})
-	defer rt.Close()
+	}
 
-	btc, err := NewBlipTesterClientOptsWithRT(t, rt, nil)
-	require.NoError(t, err)
+	btc := NewBlipTesterClientOpts(&BlipTesterClientOpts{
+		SkipVersionVectorInitialization: true, // no need for version vector test here
```
bbrks: Is this just because the test doesn't do any replication?

rest/blip_client_test.go
```diff
@@ -571,74 +573,194 @@ func getCollectionsForBLIP(_ testing.TB, rt *RestTester) []string {
...
 		BlipTesterClientOpts: *opts,
-		rt:                   rt,
 	}
 
+	return client
+}
+
+func (btc *BlipTesterClient) Close() {
+	btc.pullReplication.Close()
+	btc.pushReplication.Close()
```
bbrks:
```suggestion
	btc.tearDownBlipClientReplications()
```

rest/blip_client_test.go
```diff
@@ -571,74 +573,194 @@ func getCollectionsForBLIP(_ testing.TB, rt *RestTester) []string {
...
+	parentT.Run("versionVector", func(t *testing.T) {
+		// bump sub protocol version here before replication init
+		err := btc.createBlipTesterReplications()
+		require.NoError(t, err)
+		err = btc2.createBlipTesterReplications()
+		require.NoError(t, err)
+
+		btc.rt.TB = t
+		defer func() { btc.rt.TB = parentT }()
+		test(t)
```
bbrks: There's a lot of repetition in this function and in the other Run functions which I think can largely be reduced down into a common run function that take various parameters

rest/blip_client_test.go
```diff
@@ -571,74 +573,194 @@ func getCollectionsForBLIP(_ testing.TB, rt *RestTester) []string {
...
+	btc.pushReplication.Close()
+	for _, collectionClient := range btc.collectionClients {
+		collectionClient.Close()
+	}
+	if btc.nonCollectionAwareClient != nil {
+		btc.nonCollectionAwareClient.Close()
+	}
+}
+
+func (btc *BlipTesterClient) RunDualClients(btc2 *BlipTesterClient, test func(t *testing.T), t *testing.T, rtConfig *RestTesterConfig) {
```
bbrks: I think we might be able to take this refactoring a little bit further to simplify testing.

Right now, there's nothing forcing a dev writing a test to use the Run methods other than convention (which is easily forgotten).

It would be handy to refactor this more to force us to put test code inside the function called by Run.
- I'm thinking of something like moving the methods like `PutRev`, `StartOneshotPull`, `WaitForVersion`, etc. onto a type that is only accessible from the function passed into Run.
  - (This type might actually be called `BlipTesterClient`, and the "test runner" called something else)
  - Right now it's easy to forget to use `Run()`, since autocomplete will suggest you can do things like `btc.PutRev` outside of the context of a `btc.Run(...)`
- By doing this, it might allow you to do something clever like creating your clients in the test function itself, rather than having to write something like `RunDualClients` (what happens when we test 3 clients with HLVs? :) )

bbrks (dismissed):
See my comment on `RunDualClients` - I'd like to see if we can take this refactor a little bit further to enforce the usage of `Run` and steer us towards a more common test pattern.

The other comments are just small things I noticed before thinking about larger changes.

rest/blip_api_collections_test.go
gregns1: I think I had this set for too many test, have removed the ones that can still run run version vector mode. I think the only tests I now set tis to true is TestSendRevisionNoRevHandling (due to no rev handling being different for VV) TestBlipPushPullV2AttachmentV2Client, TestBlipProveAttachmentV2, TestBlipProveAttachmentV2Push (all because they specifically test V2 protocol).

torcolvin (approved):

PR #6569, opened 2023-11-08
Bump Go version to 1.21.4

bbrks:
https://github.com/golang/go/issues?q=milestone%3AGo1.21.4+label%3ACherryPickApproved

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v4
         with:
-          go-version: 1.21.3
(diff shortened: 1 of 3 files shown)


torcolvin:
LGTM to me, I wonder if there's a way that we can also make weekly-jenkins use go version from a file so we do not have to remember to update it.

bbrks:
It's possible, but we'll probably need to roll our own Go toolchain install code instead of relying on the Jenkins Golang plugin.

We used to do this in the distant past with `gvm`. There are now better/builtin ways for doing this (https://go.dev/doc/manage-install) so might be worth revisiting at some point.

torcolvin (approved):

PR #6571, opened 2023-11-10
CBG-3355: Add current version to channel cache

gregns1:
CBG-3355

- Add CV to log entries struct that we populate channel cache with. I have separated CV into source id and version as separate entries on log entries as the CV struct we have defined is in DB package and causes circular dependencies when used for channel cache population 
- Added test to ensure that the struct is correctly populated when adding log entry to the cache
- Added test that writes a doc and waits for its population on the channel cache. Once there it will fetch this entry and assert that CV is correctly populated from the HLV defined on the doc. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2167/

### DIFF
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -499,6 +499,13 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent) {
 	if len(rawUserXattr) > 0 {
 		collection.revisionCache.RemoveWithRev(docID, syncData.CurrentRev)
 	}
+	var docSource string
+	var docVersion uint64
+	if syncData.HLV != nil {
+		docSource = syncData.HLV.SourceID
+		docVersion = syncData.HLV.Version
+	}
+
 	change := &LogEntry{
 		Sequence:     syncData.Sequence,
 		DocID:        docID,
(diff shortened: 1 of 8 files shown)


db/change_cache.go
```diff
@@ -499,6 +499,13 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent) {
 	if len(rawUserXattr) > 0 {
 		collection.revisionCache.RemoveWithRev(docID, syncData.CurrentRev)
 	}
+	var docSource string
```
adamcfraser: Instead of defining temporary variables `docSource` and `docVersion`, could just move lines 504-507 down to line 521.

rest/changes_test.go
```diff
@@ -273,3 +274,50 @@ func TestWebhookWinningRevChangedEvent(t *testing.T) {
 	assert.Equal(t, 5, int(atomic.LoadUint32(&WinningRevChangedCount)))
 	assert.Equal(t, 6, int(atomic.LoadUint32(&DocumentChangedCount)))
 }
+
+// TestCurrentVersionPopulationOnChannelCache:
```
adamcfraser: This test should be moved to the db package - I don't see anything here that merits making the channelCache accessible from outside the db package.  In general that kind of requirement makes me think I've got the test in the wrong location, unless there's something in the rest package that also needs to be exercised.  That doesn't seem like the case here.
Have a look at the tests in db/changes_test.go (like TestFilterToAvailableChannels) for an example of a test that's just trying to write a doc and check it arriving over DCP.

adamcfraser (changes requested):

adamcfraser (approved):

PR #6575, opened 2023-11-14
CBG-3356: Add current version to ChangeEntry

gregns1:
CBG-3356

- Add CV to change entry 
- Populated when constructing the change entry from log entries retrieved from the channel cache 
- Test that checks changes are correctly populated with CV
- NOTE: each test has to check for changes before adding doc, this has the intention of 'activating' channel cache. Needed as backfill for channel cache is not yet implemented for CV  
- Some tests asserting on change entries needed changing. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2214/

### DIFF
--- a/db/changes.go
+++ b/db/changes.go
@@ -1290,8 +1290,7 @@ func createChangesEntry(ctx context.Context, docid string, db *DatabaseCollectio
 
 	if populatedDoc.HLV != nil {
 		cv := SourceAndVersion{}
-		cv.SourceID = populatedDoc.HLV.SourceID
-		cv.Version = populatedDoc.HLV.Version
+		cv.SourceID, cv.Version = populatedDoc.HLV.GetCurrentVersion()
 		row.CurrentVersion = &cv
 	}
 
--- a/db/changes.go
+++ b/db/changes.go
@@ -483,6 +483,8 @@ func makeChangeEntry(logEntry *LogEntry, seqID SequenceID, channel channels.ID) 
 		collectionID: logEntry.CollectionID,
 	}
 	// populate CurrentVersion entry if log entry has sourceID and Version populated
+	// This allows current version to be nil in event of CV not being populated on log entry
+	// allowing omitempty to work as expected
 	if logEntry.SourceID != "" && logEntry.Version != 0 {
 		change.CurrentVersion = &SourceAndVersion{SourceID: logEntry.SourceID, Version: logEntry.Version}
 	}
--- a/db/changes.go
+++ b/db/changes.go
@@ -57,7 +57,7 @@ type ChangeEntry struct {
 	principalDoc   bool         // Used to indicate _user/_role docs
 	Revoked        bool         `json:"revoked,omitempty"`
 	collectionID   uint32
-	CurrentVersion *CurrentVersionVector `json:"current_version,omitempty"` // the current version of the change entry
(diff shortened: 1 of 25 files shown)


db/crud.go
```diff
@@ -1059,7 +1059,11 @@ func (db *DatabaseCollectionWithUser) Put(ctx context.Context, docid string, bod
 		return newDoc, newAttachments, false, nil, nil
 	})
 
-	return newRevID, doc, err
+	if err == nil {
+		currentVersion = SourceAndVersion{SourceID: doc.HLV.SourceID, Version: doc.HLV.Version}
+	}
+
+	return newRevID, currentVersion, doc, err
```
bbrks: If this value is computable from everything in `doc` do we need a new return value from this method?

Places that care about `currentVersion` could call a method like `doc.CurrentVersion()` that just returns `SourceAndVersion{SourceID: doc.HLV.SourceID, Version: doc.HLV.Version}` ?

db/changes.go
```diff
@@ -1281,6 +1288,13 @@ func createChangesEntry(ctx context.Context, docid string, db *DatabaseCollectio
 	row.Seq = SequenceID{Seq: populatedDoc.Sequence}
 	row.SetBranched((populatedDoc.Flags & channels.Branched) != 0)
 
+	if populatedDoc.HLV != nil {
+		cv := SourceAndVersion{}
+		cv.SourceID = populatedDoc.HLV.SourceID
+		cv.Version = populatedDoc.HLV.Version
+		row.CurrentVersion = &cv
+	}
```
bbrks: The method on `Document` I mentioned in previous comment would also be of use here.

bbrks (changes requested):
A couple of questions about reducing amount of changes due to the function signature change to `Put()` - I think we might be able to get away without it

rest/changes_test.go
```diff
@@ -273,3 +273,65 @@ func TestWebhookWinningRevChangedEvent(t *testing.T) {
 	assert.Equal(t, 5, int(atomic.LoadUint32(&WinningRevChangedCount)))
 	assert.Equal(t, 6, int(atomic.LoadUint32(&DocumentChangedCount)))
 }
+
+func TestCVPopulationOnChangesViaAPI(t *testing.T) {
```
bbrks, line 277: Could we get tests to cover CV on the doc ID changes feed and revocation feeds? I see the code is changed for those as well (`makeChangeEntry` for normal/revoked and `createChangesEntry` for doc IDs) but no coverage.

bbrks (dismissed):
I don't know if the CV on the changes feed needs formatting or if JSON is fine, the original issue doesn't specify - so will just approve with the assumption it's correct.

rest/changes_test.go
gregns1: I have updated the test I have for doc ID changes feed to actually use this feed. I made a mistake and it wasn't actually using this code path. Should be fixed now in TestCVPopulationOnDocIDChanges. Really good spot! 

With regards to the revocation feed codepath, until Adam's changes for the channel indexes are merged I can't test that codepath. Revocation feed bypass the channel cache -> https://github.com/couchbase/sync_gateway/blob/be6b7047ac1b6d8ee953c3ff8b6de595cead7194/db/changes.go#L199 and force the query, at which we don't populate the CV off the query yet.

bbrks (approved):

PR #6580, opened 2023-11-16
gitignore: planPIndexes directory from cbgt (MB-59477)

bbrks:
Awaiting cbgt-side fix (MB-59477) but until then, ignore before somebody (me) accidentally commits it.


torcolvin (approved):
Thanks so much!!!!

PR #6584, opened 2023-11-27
CBG-3271 Implement bootstrap connection for rosmar

torcolvin:
- implement and start a BootstrapConnection in RestTester
- create a new bucket object each time since it's a cheap copy option in rosmar

some test fixup:

- use NewDbConfig for rosmar config, which uses a peculiar combination of views and collections, normally disallowed
- make some config polling faster to speed up testing
- change some error reporting to use testify to report trace

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2191/

### DIFF
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -355,6 +355,9 @@ func (cc *CouchbaseCluster) WriteMetadataDocument(ctx context.Context, location,
 	if cc == nil {
 		return 0, errors.New("nil CouchbaseCluster")
 	}
+	if cas == 0 {
+		return 0, RedactErrorf("CAS for %q in bucket %q must be non-zero", docID, location)
+	}
 
 	b, teardown, err := cc.getBucket(ctx, location)
 	if err != nil {
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -48,6 +48,9 @@ type BootstrapConnection interface {
 	// Returns exists=false if key is not found, returns error for any other error.
 	GetDocument(ctx context.Context, bucket, docID string, rv interface{}) (exists bool, err error)
 
+	// SetConnectionStringServerless sets the connection string to use serverless mode, needs to be called before any connection occurs.
+	SetConnectionStringServerless() error
+
 	// Returns the bootstrap connection's cluster connection as N1QLStore for the specified bucket/scope/collection.
 	// Does NOT establish a bucket connection, the bucketName/scopeName/collectionName is for query scoping only
 	GetClusterN1QLStore(bucketName, scopeName, collectionName string) (*ClusterOnlyN1QLStore, error)
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -356,7 +356,7 @@ func (cc *CouchbaseCluster) WriteMetadataDocument(ctx context.Context, location,
 		return 0, errors.New("nil CouchbaseCluster")
 	}
 	if cas == 0 {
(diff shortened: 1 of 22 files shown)


rest/server_context.go
```diff
@@ -369,10 +368,15 @@ func (sc *ServerContext) AllDatabases() map[string]*db.DatabaseContext {
 	return databases
 }
 
-func (sc *ServerContext) AllInvalidDatabases() map[string]*invalidConfigInfo {
+// AllInvalidDatabaseNames returns the names of all the databases that have invalid configs
+func (sc *ServerContext) AllInvalidDatabaseNames(t *testing.T) []string {
```
adamcfraser: As discussed, can move to utilities_testing.go

rest/server_context.go
```diff
@@ -2021,21 +2013,22 @@ func (sc *ServerContext) Database(ctx context.Context, name string) *db.Database
 func (sc *ServerContext) initializeCouchbaseServerConnections(ctx context.Context, failFast bool) error {
```
adamcfraser: It looks like initializeCouchbaseServerConnections does two things:
1. Set up gocb connections for GoCBAgent admin auth (GoCBAgent, NoX509HTTPClient)
2. Set up the bootstrap connection
Now that the second one of these isn't Couchbase Server-specific, I feel like it might be cleaner to split into two functions (InitializeAdminAuthConnections, InitializeBootstrap), so that we can keep the 'ServerIsWalrus' check up in SetupServerContext.  This also avoids the cognitive dissonance of calling 'initializeCouchbaseServerConnections' to set up a rosmar bootstrap.

rest/utilities_testing.go
```diff
@@ -2584,6 +2579,17 @@ func (sc *ServerContext) RequireInvalidDatabaseConfigNames(t *testing.T, expecte
 	require.EqualValues(t, expectedDbNames, dbNames)
 }
 
+// AllInvalidDatabaseNames returns the names of all the databases that have invalid configs. Testing only since this locks the database context.
+func (sc *ServerContext) AllInvalidDatabaseNames(_ *testing.T) []string {
```
adamcfraser, line 2582: It looks like every test use of AllInvalidDatabaseNames is actually only interested in the length of invalidDatabaseConfigTracking.dbNames.  Would a helper function for length only make more sense?

base/bootstrap.go
```diff
@@ -352,6 +355,9 @@ func (cc *CouchbaseCluster) WriteMetadataDocument(ctx context.Context, location,
 	if cc == nil {
 		return 0, errors.New("nil CouchbaseCluster")
 	}
+	if cas == 0 {
+		return 0, RedactErrorf("CAS for %q in bucket %q must be non-zero", UD(docID), MD(location))
```
adamcfraser: Including something in the error message about how the caller should be using InsertMetadataDocument if they are trying to insert might be helpful.

adamcfraser (changes requested):
Overall looks good, a few comments on implementation details.

rest/utilities_testing.go
torcolvin: I'm inclined to leave this - the places this is called is for checking the length, but we can't directly do:

`require.Len(t, sc.AllInvalidDatabaseNames(), 1)`

because these are done in retry loops. The advantage of having the list of names is that testify's `require.Len` will report something like: ["dbA", "dbB"] is not of len 1 - so you see the names in the error message.

adamcfraser (approved):

PR #6588, opened 2023-11-29
CBG-3634: Fully escape JSON strings when using InjectJSONProperties

bbrks:
CBG-3634

Similar to #4451, to ensure safe JSON string output, we need to marshal it - which is almost as cheap as `strconv.Quote()`, whilst covering escape of control characters 

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2192/

### DIFF
--- a/base/util.go
+++ b/base/util.go
@@ -1309,9 +1309,9 @@ func InjectJSONProperties(b []byte, kvPairs ...KVPair) (new []byte, err error) {
 			valBytes = []byte(strconv.FormatUint(v, 10))
 		case bool:
 			valBytes = []byte(strconv.FormatBool(v))
-		case string:
-			// it's not safe to use strings without marshalling
-			valBytes = []byte(ConvertToJSONString(v))
+		// case string:
+		// it's not safe to use strings without marshalling
+		// fall through to default below
 		default:
 			valBytes, err = JSONMarshal(kv.Val)
 		}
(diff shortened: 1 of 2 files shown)


base/util.go
```diff
@@ -1307,10 +1307,11 @@ func InjectJSONProperties(b []byte, kvPairs ...KVPair) (new []byte, err error) {
 			valBytes = []byte(strconv.FormatUint(uint64(v), 10))
 		case uint64:
 			valBytes = []byte(strconv.FormatUint(v, 10))
-		case string:
-			valBytes = []byte(`"` + v + `"`)
 		case bool:
 			valBytes = []byte(strconv.FormatBool(v))
+		case string:
+			// it's not safe to use strings without marshalling
+			valBytes = []byte(ConvertToJSONString(v))
```
torcolvin: I can't think of a circumstance where this would happen where `ConvertToJSONString` would fail in practice unless there were non utf-8 strings. In this case, I don't think this function would be called because it should only be called on non binary (already json according to CBS) documents.

To be pedantic, you could include string in the default case because then `JSONMarshal` error value is checked.

bbrks: Updated, just falling through to `default` implicitly with a comment why.

torcolvin (approved):

PR #6590, opened 2023-11-29
CBG-3607: disable the ability to set shared_bucket_access to false. I…

gregns1:
CBG-3607

Disable the ability to set shared_bucket_access to false. In future we will probably look to fully remove the config parameter but for now this will protect against panics in beryllium branch that relate to xattrs not being enabled.

This PR does the following: 
- Returning error inside `dbcOptionsFromConfig` when a false value is detected for `EnableXattrs`
- Disallowing the test environment variable `SG_TEST_USE_XATTRS` to be false
- Couple of test to test the functionality 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2194/
- [x] `GSI=true,xattrs=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2202/ (should fail to start)

### DIFF
--- a/rest/server_context.go
+++ b/rest/server_context.go
@@ -1085,6 +1085,11 @@ func dbcOptionsFromConfig(ctx context.Context, sc *ServerContext, config *DbConf
 		}
 	}
 
+	// In sync gateway version 4.0+ we do not support the disabling of use of xattrs
+	if !config.UseXattrs() {
+		return db.DatabaseContextOptions{}, fmt.Errorf("sync gateway requires shared_bucket_access to be enabled")
+	}
+
 	// Create a callback function that will be invoked if the database goes offline and comes
 	// back online again
 	dbOnlineCallback := func(dbContext *db.DatabaseContext) {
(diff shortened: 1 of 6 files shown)


rest/server_context.go
```diff
@@ -1093,6 +1093,11 @@ func dbcOptionsFromConfig(ctx context.Context, sc *ServerContext, config *DbConf
 		}
 	}
 
+	// In sync gateway version 4.0+ we do not support the disabling of use of xattrs
+	if !config.UseXattrs() {
+		return db.DatabaseContextOptions{}, fmt.Errorf("sync gateway requires shared_bucket_access to be enabled")
```
bbrks: This is an annoyingly named config option but I think this is a less confusing error message:

```suggestion
		return db.DatabaseContextOptions{}, fmt.Errorf("sync gateway requires enable_shared_bucket_access=true")
```

bbrks (changes requested):
I'm OK with this intermediate step before we start ripping things out but could do with a tweak to the error message that people are going to see.

bbrks (approved):

PR #6595, opened 2023-11-30
CBG-3616 add node cpu stat

torcolvin:
I've duplicated the descriptions in openapi and go, and I don't immediately have a way to fix this. I didn't mark the field as deprecated, because I believe the couchbase docs do not generate `deprecated` output, but I wrote deprecated in the description.

If we don't want to duplicate the descriptions, or they are documented elsewhere there were a few mistakes I'll keep:

- process_cpu_percent_utilization was marked as an integer and not a float
- uptime was not listed

## Pre-review checklist


base/stats_descriptions.go
```diff
@@ -51,9 +51,11 @@ const (
 
 	NumGoroutinesDesc = "The total number of goroutines."
 
-	ProcessCPUPercentUtilDesc = "The CPU's utilization as percentage value. The CPU usage calculation is performed based on user and system CPU time, but it does not include components such as iowait. The derivation means that the values of " +
+	ProcessCPUPercentUtilDesc = "The CPU's utilization as percentage value * 100. The extra 100 is a mistake left for backwards compability. Please consider using node_cpu_percent_utilization. The CPU usage calculation is performed based on user and system CPU time, but it does not include components such as iowait. The derivation means that the values of " +
```
bbrks: 10x not 100x

bbrks (changes requested):
Fine with these changes but 10x not 100x

bbrks (approved):

PR #6597, opened 2023-11-30
CBG-3612 Allow import to override existing expiry when set by Sync Function

torcolvin:
This is definitely the right thing to do because as is the `gocb.MutateInOptions` and `gocb.UpsertOptions` will be invalid and it allows this type of code to pass where:

1. write a document with TTL via SDK
2. have a sync function that calls `expiry`

However, there's a legacy bug around this code which is harder to fix:

1. write a document with TTL via SDK
2. call `expiry(0)`
3. (bug) document will have TTL set in doc 1 and not remove the TTL 

Should I try to fix the second bug as part of this review, or file a separate ticket for this? Or is this something not likely to actually occur in practice? I worry that this is going to be a complicated fix because I think some interfaces will have to pass `exp` as a pointer to differentiate between explicit 0 and nil.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2199/

### DIFF
--- a/base/collection_gocb_utils.go
+++ b/base/collection_gocb_utils.go
@@ -18,9 +18,7 @@ func fillUpsertOptions(goCBUpsertOptions *gocb.UpsertOptions, upsertOptions *sgb
 	if upsertOptions == nil {
 		return
 	}
-	if goCBUpsertOptions.Expiry == 0 {
-		goCBUpsertOptions.PreserveExpiry = upsertOptions.PreserveExpiry
-	}
+	goCBUpsertOptions.PreserveExpiry = upsertOptions.PreserveExpiry
 }
 
 // Fills in the GoCB MutateInOptions based on the passed in MutateInOptions
@@ -28,7 +26,5 @@ func fillMutateInOptions(goCBMutateInOptions *gocb.MutateInOptions, mutateInOpti
 	if mutateInOptions == nil {
 		return
 	}
-	if goCBMutateInOptions.Expiry == 0 {
-		goCBMutateInOptions.PreserveExpiry = mutateInOptions.PreserveExpiry
-	}
+	goCBMutateInOptions.PreserveExpiry = mutateInOptions.PreserveExpiry
 }
--- a/base/collection_gocb_utils.go
+++ b/base/collection_gocb_utils.go
@@ -21,7 +21,7 @@ func fillUpsertOptions(ctx context.Context, goCBUpsertOptions *gocb.UpsertOption
 		return
 	}
 	goCBUpsertOptions.PreserveExpiry = upsertOptions.PreserveExpiry
-	if goCBUpsertOptions.Expiry != 0 && !upsertOptions.PreserveExpiry {
+	if goCBUpsertOptions.Expiry != 0 && upsertOptions.PreserveExpiry {
 		InfofCtx(ctx, KeyCRUD, "Expiry set on gocb.UpsertOptions, but sgbucket.UpsertOptions.PreserveExpiry is false. Force setting PreserveExpiry to false to allow write to proceed.")
 		goCBUpsertOptions.PreserveExpiry = false
 
@@ -34,7 +34,7 @@ func fillMutateInOptions(ctx context.Context, goCBMutateInOptions *gocb.MutateIn
 		return
 	}
 	goCBMutateInOptions.PreserveExpiry = mutateInOptions.PreserveExpiry
-	if goCBMutateInOptions.Expiry != 0 && !mutateInOptions.PreserveExpiry {
+	if goCBMutateInOptions.Expiry != 0 && mutateInOptions.PreserveExpiry {
 		InfofCtx(ctx, KeyCRUD, "Expiry set on gocb.MutateInOptions, but sgbucket.MutateInOptions.PreserveExpiry is false. Force setting PreserveExpiry to false to allow write to proceed.")
 		goCBMutateInOptions.PreserveExpiry = false
 
--- a/base/collection_gocb_utils.go
+++ b/base/collection_gocb_utils.go
@@ -9,22 +9,35 @@
 package base
 
 import (
+	"context"
+
 	"github.com/couchbase/gocb/v2"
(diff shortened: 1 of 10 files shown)


db/crud.go
```diff
@@ -1922,6 +1922,10 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 			if err != nil {
 				return
 			}
+			// If importing and the sync function has modified the expiry, allow sgbucket.MutateInOptions to modify the expiry
+			if allowImport && syncFuncExpiry != nil && *syncFuncExpiry != expiry {
```
adamcfraser: What case does the allowImport check here protect us from?  (allowImport isn't the same as "this is an import mutation")

db/import.go
```diff
@@ -124,20 +124,7 @@ func (db *DatabaseCollectionWithUser) importDoc(ctx context.Context, docid strin
...
-		if expiry == nil {
-			getExpiry, getExpiryErr := db.dataStore.GetExpiry(ctx, docid)
-			if getExpiryErr != nil {
-				return nil, getExpiryErr
-			}
-			expiry = &getExpiry
-		}
-		existingDoc.Expiry = *expiry
-	}
+	mutationOptions := &sgbucket.MutateInOptions{PreserveExpiry: true}
```
adamcfraser: I think we still need to fetch the current expiry from the bucket if the Couchbase Server version we're using doesn't support PreserveExpiry?

adamcfraser (changes requested):

db/import.go
torcolvin: All versions of server 3.1.2 supports (CBS 7.0) support PreserveExpiry but I can leave this code in place, it was mostly for clarity here.

base/collection_gocb_utils.go
```diff
@@ -9,22 +9,35 @@
...
 )
 
 // Fills in the GoCB UpsertOptions based on the passed in SG Bucket UpsertOptions
-func fillUpsertOptions(goCBUpsertOptions *gocb.UpsertOptions, upsertOptions *sgbucket.UpsertOptions) {
+func fillUpsertOptions(ctx context.Context, goCBUpsertOptions *gocb.UpsertOptions, upsertOptions *sgbucket.UpsertOptions) {
 	if upsertOptions == nil {
 		return
 	}
 	goCBUpsertOptions.PreserveExpiry = upsertOptions.PreserveExpiry
+	if goCBUpsertOptions.Expiry != 0 && !upsertOptions.PreserveExpiry {
```
adamcfraser: Shouldn't this be checking for preserveExpiry == true?

base/collection_gocb_utils.go
```diff
@@ -9,22 +9,35 @@
...
 }
 
 // Fills in the GoCB MutateInOptions based on the passed in MutateInOptions
-func fillMutateInOptions(goCBMutateInOptions *gocb.MutateInOptions, mutateInOptions *sgbucket.MutateInOptions) {
+func fillMutateInOptions(ctx context.Context, goCBMutateInOptions *gocb.MutateInOptions, mutateInOptions *sgbucket.MutateInOptions) {
 	if mutateInOptions == nil {
 		return
 	}
 	goCBMutateInOptions.PreserveExpiry = mutateInOptions.PreserveExpiry
+	if goCBMutateInOptions.Expiry != 0 && !mutateInOptions.PreserveExpiry {
```
adamcfraser: (same as above about preserveExpiry == true)

base/collection_gocb_utils.go
adamcfraser: I think the Info logging also needs to be updated.

adamcfraser (changes requested):

db/crud.go
torcolvin: removed, I misunderstood what this does (allowImport == useXattrs)

base/collection_gocb_utils.go
torcolvin: Whoops, fixed.

base/collection_gocb_utils.go
torcolvin: Whoops, fixed.

db/crud.go
```diff
@@ -1669,16 +1669,21 @@ func (db *DatabaseContext) assignSequence(ctx context.Context, docSequence uint6
...
 	} else if updatedExpiry != nil {
 		finalExp = updatedExpiry
-	} else {
-		finalExp = &expiry
+	} else if expiry != nil {
+		finalExp = expiry
 	}
 
-	doc.UpdateExpiry(*finalExp)
+	if finalExp != nil {
```
adamcfraser, line 1681: I'm unclear why we needed to change expiry to *uint32 here - it looks like this function returns the same result for expiry=0 and expiry=nil (if syncExpiry and updatedExpiry are both nil).

db/crud.go
```diff
@@ -1856,7 +1860,11 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 
 	if !db.UseXattrs() {
 		// Update the document, storing metadata in _sync property
-		_, err = db.dataStore.Update(key, expiry, func(currentValue []byte) (raw []byte, syncFuncExpiry *uint32, isDelete bool, err error) {
+		var initialExpiry uint32 // expiry before Update callback happens
+		if expiry != nil {
+			initialExpiry = *expiry
+		}
```
adamcfraser, line 1866: Same comment here  - expiry==nil has the same result as expiry==0 (initialExpiry=0), so I think I'm missing the code that requires us to change expiry to a pointer in updateAndReturnDoc.  Can you let me know where exactly that is?

adamcfraser (changes requested):

db/crud.go
adamcfraser: Discussed this - it's required for the check in updateAndReturnDoc to handle the legacy (non-PreserveExpiry) case.

adamcfraser (approved):

PR #6601, opened 2023-12-06
CBG-3255: CBL Push replication for V4 subprotocol

gregns1:
CBG-3255

Changes:
- handleProposeChanges has CheckProposedVersion implementation, which does much the same as CheckProposedRev but for HLV 
- Process Rev has some extra logic instead for V4 protocol and above. Given this function can expect revids or version vectors currently.  
- Some tests to show process rev adding the document version correctly to the bucket 
- Tests for CheckProposedVersion behaviour given certain version vectors passed into it  
- Test for extracting a Version vector struct from a string representation in the Blip string form  
- There was a bug where merge versions that were populated where not getting persisted to bucket through PutExistingCurrentVersion. Hence the change to the function there too. 

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2251/

### DIFF
--- a/db/active_replicator.go
+++ b/db/active_replicator.go
@@ -208,7 +208,13 @@ func connect(arc *activeReplicatorCommon, idSuffix string) (blipSender *blip.Sen
 	arc.replicationStats.NumConnectAttempts.Add(1)
 
 	var originPatterns []string // no origin headers for ISGR
-	blipContext, err := NewSGBlipContext(arc.ctx, arc.config.ID+idSuffix, originPatterns)
+
+	// Commented out following line pending work to get ISGR workiong woth Version Vectors.
+	//blipContext, err := NewSGBlipContext(arc.ctx, arc.config.ID+idSuffix, originPatterns)
+
+	// force use of V3 and below subprotocol versions
+	protocols := []string{"CBMobile_3", "CBMobile_2"}
+	blipContext, err := NewSGBlipContextWithProtocols(arc.ctx, arc.config.ID+idSuffix, originPatterns, protocols)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/db/active_replicator.go
+++ b/db/active_replicator.go
@@ -209,11 +209,12 @@ func connect(arc *activeReplicatorCommon, idSuffix string) (blipSender *blip.Sen
 
 	var originPatterns []string // no origin headers for ISGR
 
-	// Commented out following line pending work to get ISGR workiong woth Version Vectors.
-	//blipContext, err := NewSGBlipContext(arc.ctx, arc.config.ID+idSuffix, originPatterns)
+	// TODO: CBG-3661 ActiveReplicator subprotocol versions
+	// - make this configurable for testing mixed-version replications
+	// - if unspecified, default to v2 and v3 until VV is supported with ISGR, then also include v4
+	protocols := []string{CBMobileReplicationV2.SubprotocolString(), CBMobileReplicationV3.SubprotocolString()}
 
-	// force use of V3 and below subprotocol versions
-	protocols := []string{"CBMobile_3", "CBMobile_2"}
+	//blipContext, err := NewSGBlipContext(arc.ctx, arc.config.ID+idSuffix, originPatterns)
 	blipContext, err := NewSGBlipContextWithProtocols(arc.ctx, arc.config.ID+idSuffix, originPatterns, protocols)
 	if err != nil {
 		return nil, nil, err
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -992,7 +992,7 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 
 	var history []string
 	if bh.activeCBMobileSubprotocol < CBMobileReplicationV4 {
-		// include current rev propery in history
+		// include current rev property in history
 		history = append(history, rev)
 	}
 	var versionVectorStr string
@@ -1028,7 +1028,7 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 
 		//  TODO: Doing a GetRevCopy here duplicates some rev cache retrieval effort, since deltaRevSrc is always
 		//        going to be the current rev (no conflicts), and PutExistingRev will need to retrieve the
-		//        current rev over again.  Should push this handling down PutExistingRev and use the rev
+		//        current rev over again.  Should push this handling down PutExistingRev and use the version
 		//        returned via callback in WriteUpdate, but blocked by moving attachment metadata to a rev property first
 		//        (otherwise we don't have information needed to do downloadOrVerifyAttachments below prior to PutExistingRev)
 
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -1183,9 +1183,9 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 	// bh.conflictResolver != nil represents an active SGR2 and BLIPClientTypeSGR2 represents a passive SGR2
 	forceAllowConflictingTombstone := newDoc.Deleted && (bh.conflictResolver != nil || bh.clientType == BLIPClientTypeSGR2)
 	if bh.conflictResolver != nil {
-		_, _, err = bh.collection.PutExistingRevWithConflictResolution(bh.loggingCtx, newDoc, history, true, bh.conflictResolver, forceAllowConflictingTombstone, rawBucketDoc)
+		_, _, err = bh.collection.PutExistingRevWithConflictResolution(bh.loggingCtx, newDoc, history, true, bh.conflictResolver, forceAllowConflictingTombstone, rawBucketDoc, ExistingVersionWithUpdateToHLV)
 	} else {
-		_, _, err = bh.collection.PutExistingRev(bh.loggingCtx, newDoc, history, revNoConflicts, forceAllowConflictingTombstone, rawBucketDoc)
+		_, _, err = bh.collection.PutExistingRev(bh.loggingCtx, newDoc, history, revNoConflicts, forceAllowConflictingTombstone, rawBucketDoc, ExistingVersionWithUpdateToHLV)
 	}
 	if err != nil {
 		return err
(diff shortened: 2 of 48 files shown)


db/active_replicator.go
```diff
@@ -207,7 +207,12 @@ func (ar *ActiveReplicator) GetStatus(ctx context.Context) *ReplicationStatus {
 func connect(arc *activeReplicatorCommon, idSuffix string) (blipSender *blip.Sender, bsc *BlipSyncContext, err error) {
 	arc.replicationStats.NumConnectAttempts.Add(1)
 
-	blipContext, err := NewSGBlipContext(arc.ctx, arc.config.ID+idSuffix)
+	// Commented out following line pending work to get ISGR workiong woth Version Vectors.
+	//blipContext, err := NewSGBlipContext(arc.ctx, arc.config.ID+idSuffix)
+
+	// force use of V3 and below subprotocol versions
+	protocols := []string{"CBMobile_3", "CBMobile_2"}
```
bbrks: Filed CBG-3661 as a follow-up, to allow configurable subprotocol verisons on an ActiveReplicator (for testing purposes).

We should try and get that into `master` and then allow Beryllium to add support for v4 ISGR once that dev milestone comes along.

db/blip_handler.go
```diff
@@ -785,12 +785,18 @@ func (bh *blipHandler) handleProposeChanges(rq *blip.Message) error {
...
+		version := change[1].(string)
 		parentRevID := ""
 		if len(change) > 2 {
 			parentRevID = change[2].(string)
 		}
-		status, currentRev := bh.collection.CheckProposedRev(bh.loggingCtx, docID, revID, parentRevID)
+		var status ProposedRevStatus
+		var currentRev string
+		if bh.activeCBMobileSubprotocol >= CBMobileReplicationV4 {
+			status, _ = bh.collection.CheckProposedVersion(bh.loggingCtx, docID, version)
```
bbrks: Do we need to set `currentRev` to `localDocCV` that `CheckProposedVersion()` returns here? It seems like we should

db/blip_handler.go
```diff
@@ -930,18 +936,22 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
...
+		return base.HTTPErrorf(http.StatusNotImplemented, "conflicts resolver handling not yet implemented for v4 protocol")
+	}
+
 	// addRevisionParams := newAddRevisionParams(rq)
 	revMessage := RevMessage{Message: rq}
 
 	// Doc metadata comes from the BLIP message metadata, not magic document properties:
 	docID, found := revMessage.ID()
-	revID, rfound := revMessage.Rev()
+	version, rfound := revMessage.Rev()
```
bbrks: We should have a quick discussion with the team to align terminology as some of it is ambiguous. I'm on the fence about leaving this as revID.

We're already starting to use `version` to mean HLV in some areas of the SG codebase, and the replication protocol doc still calls these RevIDs, even if they could be either rev tree IDs or version vectors.

db/revision_cache_interface.go
```diff
@@ -327,6 +327,7 @@ func revCacheLoaderForDocument(ctx context.Context, backingStore RevisionCacheBa
 }
 
 // revCacheLoaderForDocumentCV used either during cache miss (from revCacheLoaderForCv), or used directly when getting current active CV from cache
+// nolint:staticcheck
```
bbrks: Can you include some information on the end of the nolint comment about why we don't want to lint this?

db/hybrid_logical_vector_test.go
```diff
@@ -358,3 +358,153 @@ func (h *HLVAgent) insertWithHLV(ctx context.Context, key string) (casOut uint64
...
+		})
+	}
+}
+
+func BenchmarkExtractHLVFromBlipMessage(b *testing.B) {
+	for _, bm := range extractHLVFromBlipMsgBMarkCases {
+		b.Run(bm.name, func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				_, err := extractHLVFromBlipMessage(bm.hlvString)
+				require.NoError(b, err)
```
bbrks:
```suggestion
				_, _ = extractHLVFromBlipMessage(bm.hlvString)
```

assertions can be very expensive in the context of a benchmark due to the way they work. Better to ignore values entirely if you have test/assertion coverage elsewhere.

The explicit ignored error `_, _ =` should avoid linters flagging it up.

db/hybrid_logical_vector_test.go
```diff
@@ -358,3 +358,153 @@ func (h *HLVAgent) insertWithHLV(ctx context.Context, key string) (casOut uint64
...
+		previousVersions: true,
+	},
+	{
+		name:        "test case 5",
+		hlvString:   "24@def",
+		expectedHLV: []string{"def@24"},
+	},
+}
+
+func TestBenchmarkHLVMessageCases(t *testing.T) {
```
bbrks: Is the only difference between this test and `TestExtractHLVFromChangesMessage` the set of test cases?

Should we just have one test that encompasses both sets?

db/hybrid_logical_vector.go
```diff
@@ -356,3 +358,87 @@ func (hlv *HybridLogicalVector) setPreviousVersion(source string, version uint64
 	}
 	hlv.PreviousVersions[source] = version
 }
+
+// extractHLVFromBlipMessage extracts the full HLV a string in the format seen over Blip
+func extractHLVFromBlipMessage(versionVectorStr string) (HybridLogicalVector, error) {
```
bbrks, line 429: As we've already discussed, we're ok with this initial implementation to get replication working.

Filed CBG-3662 as a follow-up to optimise.

db/crud.go
```diff
@@ -1106,13 +1106,25 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 			if addNewerVersionsErr != nil {
 				return nil, nil, false, nil, addNewerVersionsErr
 			}
+			if docHLV.MergeVersions != nil {
+				if doc.HLV.MergeVersions == nil {
+					doc.HLV.MergeVersions = make(map[string]uint64)
+				}
```
bbrks: I don't think this is nessesary if we're always overwriting the map in its entirety with `docHLV.MergeVersions` ?

```suggestion
```

db/crud.go
```diff
@@ -1106,13 +1106,25 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
...
+				doc.HLV.MergeVersions = docHLV.MergeVersions
```
bbrks: Why isn't `doc.HLV` initialized with this set if we always do it, or
why doesn't `doc.HLV.AddNewerVersions(docHLV)` do this (if `MergeVersions` are not considered newer that'd be a good reason)

db/crud.go
```diff
@@ -1106,13 +1106,25 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
...
 				addNewerVersionsErr := doc.HLV.AddNewerVersions(docHLV)
 				if addNewerVersionsErr != nil {
 					return nil, nil, false, nil, addNewerVersionsErr
 				}
+				if docHLV.MergeVersions != nil {
+					if doc.HLV.MergeVersions == nil {
+						doc.HLV.MergeVersions = make(map[string]uint64)
+					}
+					doc.HLV.MergeVersions = docHLV.MergeVersions
+				}
```
bbrks: Same code in _both_ the if and else case seems like it's in the wrong place.

Why was this nessesary, and could it not be outside of the if/else if we always do the same thing.

db/hybrid_logical_vector_test.go
```diff
@@ -358,3 +358,153 @@ func (h *HLVAgent) insertWithHLV(ctx context.Context, key string) (casOut uint64
...
+		{
+			name:             "test case 2",
+			hlvString:        "25@def; 20@abc,18@hij",
+			expectedHLV:      []string{"def@25", "abc@20", "hij@18"},
+			previousVersions: true,
+		},
+		{
+			name:             "test case 3",
+			hlvString:        "25@def; 22@def,21@eff; 20@abc,18@hij",
+			expectedHLV:      []string{"def@25", "abc@20", "hij@18", "m_def@22", "m_eff@21"},
```
bbrks: Does this mean the string representation of a HLV is misaligned between CBL and SG/XDCR?

E.g. `25@def` == `def@25`

That seems like a major annoyance and possibly has performance implications (we'd need to parse/reformat rather than passing things straight through)

bbrks (changes requested):

db/hybrid_logical_vector_test.go
gregns1: So I can change this if you think it will be easier? The expected HLV list you see there is input for a test only function i created a while back (called createHLVForTest), rather annoyingly created before I read any CBL replication docs and I opted for the reverse order in the format when creating this function.  

Should I align it with CBL's representation?

db/blip_handler.go
gregns1: Good spot. I have added helper function to convert from SourceAndVersion struct to the string CBL form to be assigned to currentRev in that function.

db/crud.go
gregns1: My thinking was that they aren't necessarily newer versions. I have stopped the repeated code as you pointed out in comment above (moved outside the if statement)

db/hybrid_logical_vector_test.go
bbrks: Ahh I see. I think we should reverse/change it, if this is just our test-representation.

db/document.go
```diff
@@ -1174,6 +1188,14 @@ func (doc *Document) UnmarshalWithXattr(ctx context.Context, data []byte, xdata
 			Cas: casOnlyMeta.Cas,
 		}
 		doc._rawBody = data
+	case DocUnmarshalVV:
+		tmpData := SyncData{}
```
adamcfraser: The motivation for these custom unmarshallers is to minimize the unmarshalling work.  In this case we're doing the same amount of work as DocUnmarshalNoHistory, so I think you could just use that.

adamcfraser (changes requested):

adamcfraser:
Work has been merged separately.

PR #6610, opened 2023-12-12
CBG-3652 check Origin header on websocket

torcolvin:
Depends on:

-  https://github.com/couchbase/go-blip/pull/70
-  https://github.com/couchbase/go-blip/pull/71

The weirdest parts of this review are:

- `hostOnlyCORS` since websocket library expects originPatterns without the scheme. I specifically handle `*` and expect everything else to be a URL. I haven't validated that this will cover all cases, the implementation of matching in websocket is `filepath.Match(strings.ToLower(pattern), strings.ToLower(s))` where `s` is the `Host` portion of the URL.
- Should I move `hostOnlyCORS` to database creation where we can do validation? I don't like adding more variables to `DatabaseContext` or `DatabaseOptions` but I don't like the fact that the user can't do anything if this fails because I forgot to handle a case in converting from the typical matching in http to the matching in the websocket library.

I waffled considerably on whether to change `createBlipTesterWithSpec` or duplicate the logic in `cors_test.go` and I'm happy to change this. There's considerable weirdness about `useHostOrigin` parameter because I didn't want to add a `net.Listener` to the `BlipTesterSpec` because you have to worry about closing it and it seemed more cumbersome in tests.

## Pre-review checklist

- [x] Link upstream PRs


rest/config.go
```diff
@@ -970,6 +970,10 @@ func (dbConfig *DbConfig) validateVersion(ctx context.Context, isEnterpriseEditi
 		}
 	}
 
+	if dbConfig.CORS != nil {
+		_, err := hostOnlyCORS(dbConfig.CORS.Origin)
+		base.WarnfCtx(ctx, "The cors.origin contains values that will be ignored: %s", err)
```
adamcfraser: Technically we're still going to try to string match against these when constructing Access-Control-Allow-Origin for non-ws requests, so the error is a bit misleading.  We could indicate what we're expecting (origin is not a well-formed URL)", and maybe change the 'will' to 'may' (to indicate that we don't know for sure that clients aren't going to send a malformed 'Origin' header for non-ws requests).

adamcfraser (changes requested):

adamcfraser (approved):

PR #6614, opened 2023-12-15
Beryllium: Rename `SourceAndVersion` to `Version` / Improve HLV comments.

bbrks:
- Rename `SourceAndVersion` to just `Version`
- Rename `SourceAndVersion.Version`to`Version.Value`
- Add/Improve comments
- Add `Version.String()` method to return CBL-like string representation of a version.

### DIFF
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -47,9 +47,9 @@ func CreateVersion(source string, version uint64) Version {
 
 // String returns a Couchbase Lite-compatible string representation of the version.
 func (v Version) String() string {
-	timestamp := base.Uint64CASToLittleEndianHex(v.Value)
+	timestamp := string(base.Uint64CASToLittleEndianHex(v.Value))
 	source := base64.StdEncoding.EncodeToString([]byte(v.SourceID))
-	return fmt.Sprintf("%s@%s", timestamp, source)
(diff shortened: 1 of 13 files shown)


db/hybrid_logical_vector.go
```diff
@@ -28,19 +30,30 @@ type HybridLogicalVector struct {
...
-		Version:  version,
+		Value:    version,
 	}
 }
 
+// String returns a Couchbase Lite-compatible string representation of the version.
+func (v Version) String() string {
+	timestamp := base.Uint64CASToLittleEndianHex(v.Value)
+	source := base64.StdEncoding.EncodeToString([]byte(v.SourceID))
+	return fmt.Sprintf("%s@%s", timestamp, source)
```
adamcfraser: Should we do concatenation here instead of Sprintf?

adamcfraser (dismissed):
Looks good.  Minor suggestion for String() efficiency.

db/hybrid_logical_vector.go
bbrks:
```suggestion
	timestamp := string(base.Uint64CASToLittleEndianHex(v.Value))
	source := base64.StdEncoding.EncodeToString([]byte(v.SourceID))
	return timestamp + "@" + source
```

adamcfraser (approved):

PR #6620, opened 2023-12-19
Remove logging due to potential Start/Stop race condition

torcolvin:
I don't love this fix for https://mobile.jenkins.couchbase.com/job/sgw-unix-build/30553/consoleFull - 

I would consider just dropping this log line like we do in push, especially since it is trace only and `stopAndDisconnect` already has a lock.

### DIFF
--- a/db/active_replicator_pull.go
+++ b/db/active_replicator_pull.go
@@ -299,7 +299,11 @@ func (apr *ActivePullReplicator) registerCheckpointerCallbacks(c *activeReplicat
 
 // Stop stops the pull replication and waits for the sub changes goroutine to finish.
 func (apr *ActivePullReplicator) Stop() error {
+	// lock around apr.ctx which can be updated during Start method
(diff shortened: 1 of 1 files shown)


bbrks (changes requested):
I'd rather just remove the log. We definitely shouldn't do this.

_Maybe_ add a context-free trace log inside `stopAndDisconnect` before acquiring lock to find deadlocks? I think that was the original motivation for this line

torcolvin:
> I'd rather just remove the log. We definitely shouldn't do this.
> 
> _Maybe_ add a context-free trace log inside `stopAndDisconnect` before acquiring lock to find deadlocks? I think that was the original motivation for this line

As soon as I wrote this I hated it and I'm not sure why I even put it up for review as is. Removed the logging, since I think running with trace is exceptionally rare.

bbrks (approved):

PR #6289, opened 2023-06-06
CBG-2898: Update stat comment

gregns1:
CBG-2898

 Small PR to just update comment on already existing stat in place.

## Pre-review checklist
- [x] Removed debug logging (`fmt.Print`, `log.Print`, ...)
- [x] Logging sensitive data? Make sure it's tagged (e.g. `base.UD(docID)`, `base.MD(dbName)`)
- [x] Updated relevant information in the API specifications (such as endpoint descriptions, schemas, ...) in `docs/api`

## [Integration Tests](https://jenkins.sgwdev.com/job/SyncGateway-Integration/build?delay=0sec)
- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/1852/

PR #6621, opened 2023-12-20
Correct NumReplicationsActive Prometheus stat description

bbrks:
This was already fixed for the expvar description of the stat in #6289, but wasn't updated in the Prometheus stat description added in #6334

- n/a

### DIFF
--- a/base/stats_descriptions.go
+++ b/base/stats_descriptions.go
@@ -247,7 +247,7 @@ const (
 
 	NumDocWritesDesc = "The total number of documents written by any means (replication, rest API interaction or imports) since Sync Gateway node startup."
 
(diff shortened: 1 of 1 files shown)


torcolvin (approved):
This seems good. Some of the non database stats are up to date in https://github.com/couchbase/sync_gateway/blob/master/docs/api/components/schemas.yaml#L171 but there are not descriptions for these.

I'm OK not updating the apidocs, because another place to get out of sync, but just thought I'd ask.

bbrks:
This is a problem that needs solving at some point in the future since we now have 3 (or more) copies of stat descriptions without any way to keep them in sync. It's going to become more of an issue the more stats we add/modify.

PR #6622, opened 2023-12-20
CBG-3638 make BootstrapStartupConfigForTest parallel safe

torcolvin:
Previous to this commit, tests that used BootstrapStartupConfigForTest would start a real web server that listened on fixed ports: 14984, 14985, 14986. This worked fine for integration test runs with Couchbase Server, since they have to be run with `go test -p 1` because of limitations of the amount of churning the bucket pool does. This limitation does not apply to rosmar. If two packages were running tests simultaneously, this would error.

BootstrapStartupConfigForTest now creates servers on :0, and there is a test API for looking up the address of the listener.

Allow BootstrapStartupConfigForTest to work for CE edition without GroupID when it is not necessary, so the tests work on github. In the case of a bucket pool backed by CBS, these require a unique groupID since even a tombstoned config will be read. In rosmar, the buckets are dropped between each test so this isn't a concern. The motivation for this is to run more tests in GH actions.

Contains https://github.com/couchbase/sync_gateway/pull/6619 which should be done first.

Moved `TestPutDBConfigOIDC` out of `rest/adminapittest` so I could use `ServerContext.getServerAddr` without making more of the test parts of `ServerContext` public.

In some ways, the bootstrap tester is more heavyweight than a `RestTester` because it actually starts up listeners on external ports. I have considered whether it makes sense to re-write some of these tests using RestTester and some probably could be. I think they were written when it was less easy to work with PersistentConfig in RestTester so this was the default way to make persistent config work.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2239/


rest/adminapitest/admin_api_test.go
```diff
@@ -981,137 +981,6 @@ func TestResyncUsingDCPStreamReset(t *testing.T) {
 
 }
 
-func TestResyncForNamedCollection(t *testing.T) {
```
bbrks: Is this test already covered by a different one? Can't see this being moved anywhere in this PR.

bbrks (changes requested):
I'm good with these changes, just one question about why `TestResyncForNamedCollection` is being removed

rest/adminapitest/admin_api_test.go
torcolvin: This test never worked but I actually removed it in https://github.com/couchbase/sync_gateway/commit/0b60724d38365885c471a3bd30a82a4cf25fa223

because it made the skip:

```
	_, ok := (rt.GetDatabase().ResyncManager.Process).(*db.ResyncManager)
	if !ok {
		t.Skip("This test only works when ResyncManager is used")
	}
```

and this coverage for this has always been in `TestResyncUsingDCPStreamForNamedCollection`.

bbrks (approved):

PR #6628, opened 2024-01-03
CBG-3695 implement Couchbase Server bucket to bucket XDCR

torcolvin:
- allow making requests to management API public
- create xdcr as a separate package because I could, it could be a subpackage of base, or in base.
- create a basic Replication interface to be matched in rosmar
- change integration test script so I can more easily use a toy build

If you try to flush a bucket that has an active replication occuring, the flush will fail. If you delete a bucket that has an active replication, the entire XDCR API (REST/command line) will throw 500s for a bit until it catches up and you can delete the replication. Therefore, we always have to `Stop` the replication before deleting it.

The idea of a cluster in Couchbase Server XDCR is global. You need a cluster to map from A <-> B, before you can map buckets. Only one name can have `A<->B` mapping. This means that running XDCR with CBS will require `-p 1`, but we already require that for our integration tests.

Tested with toy build with some mobile XDCR support + 7.2.3.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2244/

### Setup for toy build usage:

Put `Dockerfile` and `run.sh` below in a directory. Call `run.sh` to start a server against a particular build.

**Dockerfile**
```
FROM couchbase/server

ARG DEB
COPY $DEB /tmp/$DEB
RUN apt-get install -y /tmp/$DEB
```

**run.sh**
```
#!/bin/bash

set -eux

SCRIPT_DIR=$(dirname $0)

cd $SCRIPT_DIR
FULL_PATH_DEB=http://172.23.126.166/builds/latestbuilds/couchbase-server/toybuilds/18191/couchbase-server-enterprise_7.6.0-18191-ubuntu20.04_amd64.deb
DEB=$(basename $FULL_PATH_DEB)

if [ -e $DEB ]; then
    echo "$DEB exists"
else
    http -d $FULL_PATH_DEB
fi

DOCKER_IMAGE=cbsxdcr
docker build --tag $DOCKER_IMAGE --build-arg DEB=$DEB --platform linux/amd64 .

export DOCKER_DEFAULT_PLATFORM=linux/amd64
~/repos/sync_gateway/integration-test/start_server.sh --non-dockerhub $DOCKER_IMAGE:latest
```

### DIFF
--- a/xdcr/cbs_xdcr_test.go
+++ b/xdcr/cbs_xdcr_test.go
@@ -1,0 +1,65 @@
+// Copyright 2024-Present Couchbase, Inc.
+//
+// Use of this software is governed by the Business Source License included
+// in the file licenses/BSL-Couchbase.txt.  As of the Change Date specified
+// in that file, in accordance with the Business Source License, use of this
+// software will be governed by the Apache License, Version 2.0, included in
+// the file licenses/APL2.txt.
+
+package xdcr
+
+import (
+	"testing"
+	"time"
+
+	"github.com/couchbase/sync_gateway/base"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCBSXDCR(t *testing.T) {
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("This test is testing Couchbase Server XDCR")
+	}
+	ctx := base.TestCtx(t)
+	bucket1 := base.GetTestBucket(t)
+	bucket2 := base.GetTestBucket(t)
+	defer bucket1.Close(ctx)
+	defer bucket2.Close(ctx)
+
+	fromBucket, err := base.AsGocbV2Bucket(bucket1)
+	require.NoError(t, err)
+	toBucket, err := base.AsGocbV2Bucket(bucket2)
+	require.NoError(t, err)
+
+	xdcr, err := NewCouchbaseServerXDCR(ctx, fromBucket, toBucket)
+	require.NoError(t, err)
+	err = xdcr.Start(ctx)
+	require.NoError(t, err)
+	defer func() {
+		assert.NoError(t, xdcr.Stop(ctx))
+	}()
+	const (
+		syncDoc       = "_sync:doc1doc2"
+		attachmentDoc = "_sync:att2:foo"
+		normalDoc     = "doc2"
+		exp           = 0
+		body          = `{"key":"value"}`
+	)
+	for _, doc := range []string{syncDoc, attachmentDoc, normalDoc} {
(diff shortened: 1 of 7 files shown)


xdcr/cbs_xdcr_test.go
```diff
@@ -0,0 +1,65 @@
...
+		require.NoError(t, err)
+	}
+	// make sure attachments are copied
+	for _, doc := range []string{normalDoc, attachmentDoc} {
+		require.EventuallyWithT(t, func(c *assert.CollectT) {
+			var value string
+			_, err = bucket2.DefaultDataStore().Get(doc, &value)
+			assert.NoError(c, err, "Could not get doc %s", doc)
+			assert.Equal(c, body, value)
+		}, time.Second*5, time.Millisecond*100)
```
bbrks, line 60: I didn't know this was a thing, but it seems better than our normal `WaitForCondition` with the `CollectT`/assertion behaviour. Neat.

xdcr/replication.go
```diff
@@ -0,0 +1,22 @@
...
+package xdcr
+
+import "context"
+
+// Replication represents a bucket to bucket replication.
+type Replication interface {
+	// Start starts the replication.
+	Start(context.Context) error
+	// Stop terminates the replication.
+	Stop(context.Context) error
```
bbrks, line 19: Pause/Resume will be useful for our testing when we need to create conflicts.

Similarly, some way of introspecting a replication (status, stats)

Not sure if you want to wait until there's a basic Rosmar implementation before adding those - it might make it easier to abstract if there are two implementations.

bbrks (approved):
Seems OK to me 👍

xdcr/replication.go
torcolvin: I was going to do a minimal implementation so I get something up for rosmar - I agree that there's stats needed.

I'm happy to merge as is or wait for a more complete implementation. Do you think this should go into beryllium only branch or master branch?

bbrks: Yeah, that's fine by me.

I reckon we merge this into `master` since it doesn't really depend on any HLV code/features to be usable.

torcolvin: `Start` and `Stop` actually can be called repeatedly so I'm not sure `Pause` and `Resume` are necessary. It depends on how this integrates with stats.

bbrks: I'm not sure what the actual difference is between pause/resume and create/delete.

I just saw it in the XDCR REST API docs and assumed we'd be following their naming/operations/semantics for this interface and make Rosmar follow suit.

https://docs.couchbase.com/server/current/rest-api/rest-xdcr-pause-resume.html

PR #6631, opened 2024-01-08
CBG_3645 Prereq: don't override http.DefaultClient

torcolvin:
Setting http.DefaultClient.Jar would cause anything that ran after it
to authenticate with this session token. In practice, this happens for
blip tests becuase they use guest access and the flow is:

- check basic auth
- check session
- check guest access

To be safe, I switched places that use http.DefaultClient as a global in
anything other than read-only to use http.Client{} explicitly.

At initial look, this allows {{go test -shuffle on}} to work.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2247/

### DIFF
--- a/base/util.go
+++ b/base/util.go
@@ -1538,7 +1538,7 @@ func GetHttpClient(insecureSkipVerify bool) *http.Client {
 		transport.TLSClientConfig.InsecureSkipVerify = true
 		return &http.Client{Transport: transport}
 	}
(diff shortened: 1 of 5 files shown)


bbrks (approved):
Nice. This is a really subtle/unfortunate side-effect using `http.DefaultClient`!

PR #6633, opened 2024-01-08
CBG-3700 fix race conditions

torcolvin:
Addresses a race in `go test -tags cb_sg_enterprise ./rest/replicatortest -race -run TestGroupIDReplications`

- each ServerContext will reset the global logger. Provide a way to avoid the logging setup, assuming that the test logging framework will already be configured. The logger can get reset while a different server (database) is logging, such as writing checkpoint documents.
- dbReplicatorStats map can be accessed simultaneously by SGReplicateMgr and when it is explicitly called in the test.

I don't really care for the way I avoid re-initialization of the global logger, but I don't have a better alternative in mind.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2450/

### DIFF
--- a/base/stats.go
+++ b/base/stats.go
@@ -359,7 +359,7 @@ type DbStats struct {
 	SecurityStats           *SecurityStats                `json:"security,omitempty"`
 	SharedBucketImportStats *SharedBucketImportStats      `json:"shared_bucket_import,omitempty"`
 	CollectionStats         map[string]*CollectionStats   `json:"per_collection,omitempty"`
-	replicatorMutex         sync.Mutex
+	dbReplicatorStatsMutex         sync.Mutex
 }
 
 type CacheStats struct {
--- a/base/stats.go
+++ b/base/stats.go
@@ -359,7 +359,7 @@ type DbStats struct {
 	SecurityStats           *SecurityStats                `json:"security,omitempty"`
 	SharedBucketImportStats *SharedBucketImportStats      `json:"shared_bucket_import,omitempty"`
 	CollectionStats         map[string]*CollectionStats   `json:"per_collection,omitempty"`
-	dbReplicatorStatsMutex         sync.Mutex
+	dbReplicatorStatsMutex  sync.Mutex
 }
 
 type CacheStats struct {
@@ -1960,8 +1960,8 @@ func (d *DbStats) InitCollectionStats(scopeAndCollectionNames ...string) error {
 }
 
 func (d *DbStats) DBReplicatorStats(replicationID string) (*DbReplicatorStats, error) {
-	d.replicatorMutex.Lock()
-	defer d.replicatorMutex.Unlock()
+	d.dbReplicatorStatsMutex.Lock()
+	defer d.dbReplicatorStatsMutex.Unlock()
 
 	if _, ok := d.DbReplicatorStats[replicationID]; ok {
 		return d.DbReplicatorStats[replicationID], nil
--- a/base/stats.go
+++ b/base/stats.go
@@ -359,6 +359,7 @@ type DbStats struct {
 	SecurityStats           *SecurityStats                `json:"security,omitempty"`
 	SharedBucketImportStats *SharedBucketImportStats      `json:"shared_bucket_import,omitempty"`
 	CollectionStats         map[string]*CollectionStats   `json:"per_collection,omitempty"`
+	replicatorMutex         sync.Mutex
 }
 
 type CacheStats struct {
@@ -1096,7 +1097,8 @@ func (s *SgwStats) NewDBStats(name string, deltaSyncEnabled bool, importEnabled 
 	s.dbStatsMapMutex.Lock()
 	defer s.dbStatsMapMutex.Unlock()
 	dbStats := &DbStats{
-		dbName: name,
+		dbName:            name,
+		DbReplicatorStats: make(map[string]*DbReplicatorStats),
 	}
 
 	// These have a pretty good chance of being used so we'll initialise these for every database stat struct created
@@ -1823,6 +1825,7 @@ func (d *DbStats) initSecurityStats() error {
 }
 
 func (d *DbStats) unregisterReplicationStats(replicationID string) {
+
 	if d.DbReplicatorStats[replicationID] == nil {
 		return
 	}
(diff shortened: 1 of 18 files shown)


base/stats.go
```diff
@@ -359,6 +359,7 @@ type DbStats struct {
 	SecurityStats           *SecurityStats                `json:"security,omitempty"`
 	SharedBucketImportStats *SharedBucketImportStats      `json:"shared_bucket_import,omitempty"`
 	CollectionStats         map[string]*CollectionStats   `json:"per_collection,omitempty"`
+	replicatorMutex         sync.Mutex
```
adamcfraser:
```suggestion
	dbReplicatorStatsMutex         sync.Mutex
```

rest/config_startup.go
```diff
@@ -85,6 +85,8 @@ type StartupConfig struct {
 	CouchbaseKeepaliveInterval *int   `json:"couchbase_keepalive_interval,omitempty" help:"TCP keep-alive interval between SG and Couchbase server"`
 
 	DeprecatedConfig *DeprecatedConfig `json:"-,omitempty" help:"Deprecated options that can be set from a legacy config upgrade, but cannot be set from a 3.0 config."`
+
+	avoidLoggingSetup bool `json:"-"` // Used to avoid logging setup so as to not modify globals. This only used for testing multiple ServerContexts simultaneously. This will use the same logging configuration as is already set up.
```
adamcfraser: Do you think we could build this into initLogging, so that callers don't need to worry about setting this flag?  (i.e. have InitLogging run within a mutex and set a global flag when complete, and check that flag when run)

adamcfraser (changes requested):

rest/config_startup.go
torcolvin: There's a single test `TestConfigsIncludeDefaults` that requires initialization of a logger, to test whether it will pass the `DCP` base key to the database level. I'm not convinced this part of this test needs the workaround `defer RunBootstrapLoggerInitialization(t)()`.

I don't know if I like this solution better because it relies on a global but I think it makes the part of the code to be modified very small, which I like.

rest/config.go
```diff
@@ -1344,16 +1346,18 @@ func (sc *StartupConfig) Validate(ctx context.Context, isEnterpriseEdition bool)
...
-	// Logging config will now have been loaded from command line
-	// or from a sync_gateway config file so we can validate the
-	// configuration and setup logging now
-	if err := config.SetupAndValidateLogging(ctx); err != nil {
-		// If we didn't set up logging correctly, we *probably* can't log via normal means...
-		// as a best-effort, last-ditch attempt, we'll log to stderr as well.
-		log.Printf("[ERR] Error setting up logging: %v", err)
-		return nil, fmt.Errorf("error setting up logging: %v", err)
+	// logger will be initialized only when running tests
+	if !loggerInitialized {
```
adamcfraser: This doesn't feel like quite the right place to be using this variable - wouldn't we'd want to encapsulate it in the logging handling (inside SetupAndValidateLogging)?

We'd also still need some sort of synchronization to avoid races.

rest/config_startup.go
adamcfraser: From the previous commit, I had the impression that we should only ever be calling SetupAndValidateLogging once per process.  (and that unit tests with multiple server contexts should reuse the same logging config). Is that accurate?  If yes, I think we should be encapsulate the 'run once' logic inside SetupAndValidateLogging.

I looked at TestConfigsIncludeDefaults and I think it can just use SetUpTestLogging to initialize the DCP log key - I don't think it needs a custom call to SetupAndValidateLogging.

torcolvin: I needed to add `config.Logging.Console.Enabled = base.BoolPtr(true)` to this test, which is normally enabled as a side effect of `InitLogging` but not normally needed in a JSON for server config. I think this is fine for this test, which keeps the guts of the inheritance.

adamcfraser (approved):

rest/config.go
```diff
@@ -1344,26 +1348,31 @@ func (sc *StartupConfig) Validate(ctx context.Context, isEnterpriseEdition bool)
...
-	// configuration and setup logging now
-	if err := config.SetupAndValidateLogging(ctx); err != nil {
-		// If we didn't set up logging correctly, we *probably* can't log via normal means...
-		// as a best-effort, last-ditch attempt, we'll log to stderr as well.
-		log.Printf("[ERR] Error setting up logging: %v", err)
-		return nil, fmt.Errorf("error setting up logging: %v", err)
+	// If SetupServerContext is called while any other go routines that might use logging are running, it will
+	// cause a data race, therefore only initialize logging and other globals on the first call. From a main
+	// program, there is only one ServerContext.
+	if serverContextGlobalsInitialized.CompareAndSwap(false, true) {
```
bbrks, line 1354: I'm not sure how this is intended to work for a suite of tests that only invoke `SetupServerContext` once per test.

Is it expected that only the first test gets to set global config, and all others can't?

I'd have expected this flag to be reset between tests maybe (`t.Cleanup`?), if the intention is to only allow this to be run once _per test_ to avoid races

rest/config.go
```diff
@@ -1344,26 +1348,31 @@ func (sc *StartupConfig) Validate(ctx context.Context, isEnterpriseEdition bool)
...
-	// Logging config will now have been loaded from command line
-	// or from a sync_gateway config file so we can validate the
-	// configuration and setup logging now
-	if err := config.SetupAndValidateLogging(ctx); err != nil {
-		// If we didn't set up logging correctly, we *probably* can't log via normal means...
-		// as a best-effort, last-ditch attempt, we'll log to stderr as well.
-		log.Printf("[ERR] Error setting up logging: %v", err)
-		return nil, fmt.Errorf("error setting up logging: %v", err)
+	// If SetupServerContext is called while any other go routines that might use logging are running, it will
+	// cause a data race, therefore only initialize logging and other globals on the first call. From a main
```
bbrks, line 1352: Doesn't protect against tests that have a mix of RestTester and full Bootstrap/ServerContext (e.g. `TestGroupIDReplications`)

PR #6636, opened 2024-01-10
Linting: Add ruleguard rule for logging errors with `%w` (suggest `%s` or `%v`)

bbrks:
- gocritic now reads ruleguard rules from `ruleguard` directory.
- Created ruleguard rule for Sync Gateway logging format containing Go error wrapping verb (%w)
- Strict lint config enables `go-critic` with only one enabled linter: `ruleguard`

Example:
```go
			base.WarnfCtx(ctx, "CheckProposedRev(%q) --> %T %w", base.UD(docid), err, err)
```

```sh
$ ruleguard -rules ruleguard/rules-*.go ./...
db/crud.go:2577:4: logwrappederr: cannot use error wrapping verb %w outside of fmt.Errorf() - use %s or %v instead? (rules-logwrappederr.go:64)

$ golangci-lint --config .golangci.yml run | rg --after-context=2 'error wrapping verb'
db/crud.go:2577:4: ruleguard: cannot use error wrapping verb %w outside of fmt.Errorf() - use %s or %v instead? (gocritic)
			base.WarnfCtx(ctx, "CheckProposedRev(%q) --> %T %w", base.UD(docid), err, err)
			^
```


ruleguard/logwrappederr.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/quasilyte/go-ruleguard/dsl"
+)
+
+// logwrappederr finds instances of Sync Gateway logging where we are attempting to use %w on an error. Should use %s or %v.
+func logwrappederr(m dsl.Matcher) {
+	m.
+		Match(
+			// this is horrible but the variadics after $err don't match with the optional wildcards $*_
+			// also no way to regex match to provide non-package qualified versions
+			`base.PanicfCtx($ctx, $format, $err)`,
```
torcolvin: I think you can do `base.PanicfCtx($ctx, $format, $*args)` and go vet will test if you use %w without an err field

ruleguard/go.mod
```diff
@@ -0,0 +1,5 @@
+module ruleguard
+
+go 1.21
```
torcolvin, line 3: Out of scope but I wonder if it's worth thinking about adding `toolchain 1.21` to our code

torcolvin (commented):
I get the following errors, in addition to the real rules:

```
rest/admin_api.go:9:1: ruleguard: execution error: used Run() with an empty rule set; forgot to call Load() first? (gocritic)
package rest
^
rest/admin_api_auth_routing_permissions.go:9:1: ruleguard: execution error: used Run() with an empty rule set; forgot to call Load() first? (gocritic)
package rest
^
rest/admin_functions_api.go:9:1: ruleguard: execution error: used Run() with an empty rule set; forgot to call Load() first? (gocritic)
package rest
```

from running `golangci-lint run --config=.golangci-strict.yml rest`

ruleguard/go.mod
bbrks: What does that give us?

torcolvin: The compiler will download a specific version of go toolchain (which is possible since go 1.21), which would make it possible to manage the versions of go used in only one place `go.mod` (vs having to manage in github workflows, Jenkinsfile, jenkins itself.

It also would be useful if we update go on a different branch, since local development would pick up that version.

torcolvin:
> I get the following errors, in addition to the real rules:
> 
> ```
> rest/admin_api.go:9:1: ruleguard: execution error: used Run() with an empty rule set; forgot to call Load() first? (gocritic)
> package rest
> ^
> rest/admin_api_auth_routing_permissions.go:9:1: ruleguard: execution error: used Run() with an empty rule set; forgot to call Load() first? (gocritic)
> package rest
> ^
> rest/admin_functions_api.go:9:1: ruleguard: execution error: used Run() with an empty rule set; forgot to call Load() first? (gocritic)
> package rest
> ```
> 
> from running `golangci-lint run --config=.golangci-strict.yml rest`

I can't reproduce this today, which makes sense. However, the `ruleguard` directory itself produces a golangci-lint error itself.

ruleguard/go.mod
bbrks: Ok, this sounds interesting but not in the context of this PR. Feel free to file a ticket.

ruleguard/logwrappederr.go
bbrks: Reworked this rule to be a bit less strict and simpler (it'll detect ALL usages of `%w` no matter the type of args)

bbrks:
> I can't reproduce this today, which makes sense. However, the `ruleguard` directory itself produces a golangci-lint error itself.

Fixed lint errors inside ruleguard directory and also included an example/test file.

ruleguard/logwrappederr.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"github.com/quasilyte/go-ruleguard/dsl"
+)
+
+// logwrappederr finds instances of Sync Gateway logging where we are attempting to use %w on an error. Should use %s or %v.
+func logwrappederr(m dsl.Matcher) {
+	m.
+		Match(
+			// this is horrible but the variadics after $err don't match with the optional wildcards $*_
+			// also no way to regex match to provide non-package qualified versions
```
torcolvin: I think this comment is dead?

ruleguard/logwrappederr.go
```diff
@@ -0,0 +1,45 @@
...
+			`base.WarnfCtx($ctx, $format, $*_)`,
+			`WarnfCtx($ctx, $format, $*_)`,
+			`base.InfofCtx($ctx, $logkey, $format, $*_)`,
+			`InfofCtx($ctx, $logkey, $format, $*_)`,
+			`base.DebugfCtx($ctx, $logkey, $format, $*_)`,
+			`DebugfCtx($ctx, $logkey, $format, $*_)`,
+			`base.TracefCtx($ctx, $logkey, $format, $*_)`,
+			`TracefCtx($ctx, $logkey, $format, $*_)`,
+			`fmt.Printf($format, $*_)`,
+			`log.Printf($format, $*_)`,
```
torcolvin, line 38: does `t.Logf` belong here?

torcolvin (dismissed):

torcolvin (approved):

PR #6640, opened 2024-01-12
CBG-3254: CBL pull replication for v4 protocol

gregns1:
CBG-3254

Breakdown of changes:

- Changes to sendChanges: if in v4 protocol replication we populate the rev field on changes row with the current version off the change entry. For all other protocol versions we leave it as is. 
- handleChangesResponse: added code to temporarily send full revision for v4 replication
Inside sendRevision, I firstly changed revid -> rev given this can be a string of a revid or current version. Added code to fetch the doc by CV if running in v4 replication and above. Also added function toHistoryForHLV which will take a full HLV and convert it to the CBL string format needed for the replicator. The result of this is passed into blipRevMessageProperties as it was before. 
- I needed to change the documentRevision struct returned by the rev cache. Before we just had CV on this struct but given we need the full HLV to send in a blip message inside sendRevision I have removed CV added a string representation of history to the stuct.
- I also enabled the versionVector sub test on the btcRunner. In doing this we needed a way to assert on the current version as well as the revid on the DocVersion struct created by Tor. So I have had to add temp functions to perform operations directly on the db (not through rest api) so it can be assigned to this struct. In doing this, inside GetVersion which is used to assert on in the blip tests I have added a lookup by CV 
- I have added a test assertion function to assert on blip history entitled AssertOnBlipHistory which has to check what version the protocol is in as its very possible a lot of the time the history section of the blip message in our current test suite is empty given we are updating docs on the same source. 
- Added a test entitled TestHLVMapToCBLString that test the format of the conversion of the HLV maps to CBL string format. For the test case that has both PV and MV defined I have had to split the resulting string and assert on substring as the resulting string can have Merge versions in a random order given iteration through a map is random. 
- Added a test entitled TestPullReplicationUpdateOnOtherHLVAwarePeer which purpose is to see if blip history property is correctly populated in HLV aware replication. It makes use of Adams HLV agent to simulate a write from a different peer. In using this I had to move the HLV agent code into a different file as it was inaccessible in its old location
- I also refactored some of the btcRunner code from a conversion with Ben. We need the ability to skip sub tests in the btcRunner object. For example, the test I added TestPullReplicationUpdateOnOtherHLVAwarePeer needs to be HLV only replication. I have done this by passing in a map of the sub test you want to skip, this is getting used to skip attachment test and delta sync tests for CBL replication too (given we haven't yet implemented that). This allows us the freedom to add further sub tests to the runner in future (can imagine a world we may want a separate sub test for delta sync?) 

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2269/

### DIFF
--- a/db/active_replicator.go
+++ b/db/active_replicator.go
@@ -211,7 +211,7 @@ func connect(arc *activeReplicatorCommon, idSuffix string) (blipSender *blip.Sen
 	// TODO: CBG-3661 ActiveReplicator subprotocol versions
 	// - make this configurable for testing mixed-version replications
 	// - if unspecified, default to v2 and v3 until VV is supported with ISGR, then also include v4
-	protocols := []string{CBMobileReplicationV3.SubprotocolString(), CBMobileReplicationV2.SubprotocolString()} //blipContext, err := NewSGBlipContext(arc.ctx, arc.config.ID+idSuffix, originPatterns)
+	protocols := []string{CBMobileReplicationV3.SubprotocolString(), CBMobileReplicationV2.SubprotocolString()}
 	blipContext, err := NewSGBlipContextWithProtocols(arc.ctx, arc.config.ID+idSuffix, originPatterns, protocols)
 	if err != nil {
 		return nil, nil, err
--- a/db/active_replicator.go
+++ b/db/active_replicator.go
@@ -211,8 +211,7 @@ func connect(arc *activeReplicatorCommon, idSuffix string) (blipSender *blip.Sen
 	// TODO: CBG-3661 ActiveReplicator subprotocol versions
 	// - make this configurable for testing mixed-version replications
 	// - if unspecified, default to v2 and v3 until VV is supported with ISGR, then also include v4
-	protocols := []string{CBMobileReplicationV3.SubprotocolString()}
-	//blipContext, err := NewSGBlipContext(arc.ctx, arc.config.ID+idSuffix, originPatterns)
+	protocols := []string{CBMobileReplicationV3.SubprotocolString(), CBMobileReplicationV2.SubprotocolString()} //blipContext, err := NewSGBlipContext(arc.ctx, arc.config.ID+idSuffix, originPatterns)
 	blipContext, err := NewSGBlipContextWithProtocols(arc.ctx, arc.config.ID+idSuffix, originPatterns, protocols)
 	if err != nil {
 		return nil, nil, err
--- a/db/active_replicator.go
+++ b/db/active_replicator.go
@@ -208,7 +208,12 @@ func connect(arc *activeReplicatorCommon, idSuffix string) (blipSender *blip.Sen
 	arc.replicationStats.NumConnectAttempts.Add(1)
 
 	var originPatterns []string // no origin headers for ISGR
-	blipContext, err := NewSGBlipContext(arc.ctx, arc.config.ID+idSuffix, originPatterns)
+	// TODO: CBG-3661 ActiveReplicator subprotocol versions
+	// - make this configurable for testing mixed-version replications
+	// - if unspecified, default to v2 and v3 until VV is supported with ISGR, then also include v4
+	protocols := []string{CBMobileReplicationV3.SubprotocolString()}
+	//blipContext, err := NewSGBlipContext(arc.ctx, arc.config.ID+idSuffix, originPatterns)
+	blipContext, err := NewSGBlipContextWithProtocols(arc.ctx, arc.config.ID+idSuffix, originPatterns, protocols)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/db/blip.go
+++ b/db/blip.go
@@ -73,7 +73,7 @@ func defaultBlipLogger(ctx context.Context) blip.LogFn {
 }
 
 // blipRevMessageProperties returns a set of BLIP message properties for the given parameters.
-func (bsc *BlipSyncContext) blipRevMessageProperties(revisionHistory []string, deleted bool, seq SequenceID) blip.Properties {
+func blipRevMessageProperties(revisionHistory []string, deleted bool, seq SequenceID) blip.Properties {
 	properties := make(blip.Properties)
 
 	// TODO: Assert? db.SequenceID.MarshalJSON can never error
@@ -81,12 +81,7 @@ func (bsc *BlipSyncContext) blipRevMessageProperties(revisionHistory []string, d
 	properties[RevMessageSequence] = string(seqJSON)
 
 	if len(revisionHistory) > 0 {
-		if bsc.activeCBMobileSubprotocol <= CBMobileReplicationV3 {
-			properties[RevMessageHistory] = strings.Join(revisionHistory, ",")
-		} else {
-			// v4 and above will already be formatted
-			properties[RevMessageHistory] = revisionHistory[0]
-		}
+		properties[RevMessageHistory] = strings.Join(revisionHistory, ",")
 	}
 
 	if deleted {
--- a/db/blip.go
+++ b/db/blip.go
@@ -73,7 +73,7 @@ func defaultBlipLogger(ctx context.Context) blip.LogFn {
 }
 
 // blipRevMessageProperties returns a set of BLIP message properties for the given parameters.
-func blipRevMessageProperties(revisionHistory []string, deleted bool, seq SequenceID) blip.Properties {
+func (bsc *BlipSyncContext) blipRevMessageProperties(revisionHistory []string, deleted bool, seq SequenceID) blip.Properties {
 	properties := make(blip.Properties)
 
 	// TODO: Assert? db.SequenceID.MarshalJSON can never error
@@ -81,7 +81,12 @@ func blipRevMessageProperties(revisionHistory []string, deleted bool, seq Sequen
 	properties[RevMessageSequence] = string(seqJSON)
 
 	if len(revisionHistory) > 0 {
-		properties[RevMessageHistory] = strings.Join(revisionHistory, ",")
+		if bsc.activeCBMobileSubprotocol <= CBMobileReplicationV3 {
+			properties[RevMessageHistory] = strings.Join(revisionHistory, ",")
+		} else {
+			// v4 and above will already be formatted
+			properties[RevMessageHistory] = revisionHistory[0]
+		}
 	}
 
 	if deleted {
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -478,7 +478,7 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 					}
 
 				}
-				// tidy thius up + I think we need to send full VV here? Need help here I think
+				// if V3 and below populate change row with rev id
 				if bh.activeCBMobileSubprotocol <= CBMobileReplicationV3 {
 					for _, item := range change.Changes {
 						changeRow := bh.buildChangesRow(change, item["rev"])
@@ -488,8 +488,7 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 						}
 					}
 				} else {
-					// maybe create version of CreateVersionFromString for this in HLV file
-					//str := change.CurrentVersion.String()
+					// populate change row with current version
 					changeRow := bh.buildChangesRow(change, change.CurrentVersion.String())
 					pendingChanges = append(pendingChanges, changeRow)
 					if err := sendPendingChangesAt(opts.batchSize); err != nil {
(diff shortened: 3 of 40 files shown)


db/blip.go
```diff
@@ -73,15 +73,20 @@ func defaultBlipLogger(ctx context.Context) blip.LogFn {
...
+func (bsc *BlipSyncContext) blipRevMessageProperties(revisionHistory []string, deleted bool, seq SequenceID) blip.Properties {
 	properties := make(blip.Properties)
 
 	// TODO: Assert? db.SequenceID.MarshalJSON can never error
 	seqJSON, _ := base.JSONMarshal(seq)
 	properties[RevMessageSequence] = string(seqJSON)
 
 	if len(revisionHistory) > 0 {
-		properties[RevMessageHistory] = strings.Join(revisionHistory, ",")
+		if bsc.activeCBMobileSubprotocol <= CBMobileReplicationV3 {
```
adamcfraser: It looks like having v4 and above pass in a history of length 1 will already handle this without the additional subprotocol check here (strings.Join already includes streamlined handling for length 0 and length 1 slices).  Given that, do you think we need the additional Subprotocol check here?

db/crud.go
```diff
@@ -913,15 +913,15 @@ func (db *DatabaseCollectionWithUser) updateHLV(d *Document, docUpdateEvent DocU
 
 // Updates or creates a document.
 // The new body's BodyRev property must match the current revision's, if any.
-func (db *DatabaseCollectionWithUser) Put(ctx context.Context, docid string, body Body) (newRevID string, doc *Document, err error) {
+func (db *DatabaseCollectionWithUser) Put(ctx context.Context, docid string, body Body) (newRevID string, cv Version, doc *Document, err error) {
```
adamcfraser: Given that this returns the updated *Document, is there a reason callers couldn't just retrieve cv from the HLV in the Document?

rest/doc_api.go
```diff
@@ -605,9 +606,9 @@ func (h *handler) handleDeleteDoc() error {
 			return err
 		}
 	}
-	newRev, err := h.collection.DeleteDoc(h.ctx(), docid, revid)
+	newRev, cv, err := h.collection.DeleteDoc(h.ctx(), docid, revid)
 	if err == nil {
-		h.writeRawJSONStatus(http.StatusOK, []byte(`{"id":`+base.ConvertToJSONString(docid)+`,"ok":true,"rev":"`+newRev+`"}`))
+		h.writeRawJSONStatus(http.StatusOK, []byte(fmt.Sprintf(`{"id":`+base.ConvertToJSONString(docid)+`,"ok":true,"rev":"`+newRev+`", "current_version": {"source_id":"`+cv.SourceID+`", "version": %d}}`, cv.Value)))
```
adamcfraser: I didn't expect REST API changes to be part of the pull replication work - what's the reason this is required now?  (I don't think we've decided on things like the format of cv for REST API consumers)

db/blip_sync_context.go
```diff
@@ -356,9 +357,13 @@ func (bsc *BlipSyncContext) handleChangesResponse(sender *blip.Sender, response
 
 			var err error
 			if deltaSrcRevID != "" {
-				err = bsc.sendRevAsDelta(sender, docID, revID, deltaSrcRevID, seq, knownRevs, maxHistory, handleChangesResponseDbCollection, collectionIdx)
+				// error here for v4 protocol, delta sync not yet implemented for v4
+				if bsc.activeCBMobileSubprotocol > CBMobileReplicationV3 {
```
adamcfraser: As discussed, I think we prefer to just fall back to sendRevision here?

db/blip_sync_context.go
```diff
@@ -606,52 +611,68 @@ func (bsc *BlipSyncContext) sendNoRev(sender *blip.Sender, docID, revID string,
...
-func (bsc *BlipSyncContext) sendRevision(sender *blip.Sender, docID, revID string, seq SequenceID, knownRevs map[string]bool, maxHistory int, handleChangesResponseCollection *DatabaseCollectionWithUser, collectionIdx *int) error {
-	rev, err := handleChangesResponseCollection.GetRev(bsc.loggingCtx, docID, revID, true, nil)
+func (bsc *BlipSyncContext) sendRevision(sender *blip.Sender, docID, rev string, seq SequenceID, knownRevs map[string]bool, maxHistory int, handleChangesResponseCollection *DatabaseCollectionWithUser, collectionIdx *int) error {
+	var err error
+	var docRev DocumentRevision
+	if bsc.activeCBMobileSubprotocol <= CBMobileReplicationV3 {
+		docRev, err = handleChangesResponseCollection.GetRev(bsc.loggingCtx, docID, rev, true, nil)
+	} else {
+		// extract cv string rev representation
+		version, vrsErr := CreateVersionFromString(rev)
```
adamcfraser, line 617: CreateVersionFromString is a relatively expensive operation (base64 decoding and hex cas conversion).  It's particularly true here, since these are coming from changeArray (which we just constructed when we sent a batch of changes).  
More broadly, I feel like we should discuss whether we can have the majority of our code work with the formatted (base64 and string cas) in memory, to avoid constant conversion back and forth.

db/blip_sync_context.go
```diff
@@ -677,6 +698,42 @@ func toHistory(revisions Revisions, knownRevs map[string]bool, maxHistory int) [
 	return history
 }
 
+// toHistoryForHLV formats blip History property for V4 replication and above
+func toHistoryForHLV(hlv *HybridLogicalVector) string {
```
adamcfraser: This definitely feels like it should be a method on HLV.  And similar to the case above, this will be a common enough operation that we want to optimize it and preferably store it in this format in the HLV.  (i.e. if there are 10K users all replicating the same revision from the rev cache, we don't want to have to compute this uniquely for each user)

adamcfraser (changes requested):

rest/doc_api.go
gregns1: I also agree on this. I went ahead and did it to allow us to get the cv as well as the revid as this is needed for our blip tests to assert on both revid and cv when running the different subprotocol sub tests. See GetVersion that is called from WaitForVersion. 
An alternative is for me to remove these changes and then add a call to the _raw endpoint inside PutDoc which is then can be passed into DocVersionFromPutResponse to be unmarshaled?

db/blip.go
gregns1: Yes good point, didn't think of that! Made that change.

db/crud.go
gregns1: I originally tried this, the issue I had was with test that intentionally try to update conflicting doc. In this situation PutExistingRev returned a nil document struct, thus these test panic.

db/blip_sync_context.go
gregns1: I have made some changes (now to store on docRev the history in string format etc). I haven't yet removed this CreateVersionFromString() here. After the update to the storage format discussed on MB-60281 I started to change the way we represent the HLV in memory but stopped as it will grow this PR by a lot. I propose filing ticket and getting into next sprint to refactor to support the updated storage format and only convert to unit64 etc when doing conflict checks.

db/active_replicator.go
```diff
@@ -208,7 +208,12 @@ func connect(arc *activeReplicatorCommon, idSuffix string) (blipSender *blip.Sen
 	arc.replicationStats.NumConnectAttempts.Add(1)
 
 	var originPatterns []string // no origin headers for ISGR
-	blipContext, err := NewSGBlipContext(arc.ctx, arc.config.ID+idSuffix, originPatterns)
+	// TODO: CBG-3661 ActiveReplicator subprotocol versions
+	// - make this configurable for testing mixed-version replications
+	// - if unspecified, default to v2 and v3 until VV is supported with ISGR, then also include v4
+	protocols := []string{CBMobileReplicationV3.SubprotocolString()}
```
bbrks: Missing v2 (which would allow ISGR to continue working between 2.x and 4.0 nodes)

```suggestion
	protocols := []string{CBMobileReplicationV3.SubprotocolString(), CBMobileReplicationV2.SubprotocolString()}
```

db/blip_handler.go
```diff
@@ -478,8 +478,18 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 					}
 
 				}
-				for _, item := range change.Changes {
-					changeRow := bh.buildChangesRow(change, item["rev"])
+				// if V3 and below populate change row with rev id
```
bbrks, line 481: Should move the common bit down out of the if so we're not duplicating the same work for different protocol versions.

```go
				// if V3 and below populate change row with rev id
				if bh.activeCBMobileSubprotocol <= CBMobileReplicationV3 {
						pendingChanges = append(pendingChanges, changeRow)
					}
				} else {
					changeRow := bh.buildChangesRow(change, change.CurrentVersion.String())
					pendingChanges = append(pendingChanges, changeRow)
				}

				if err := sendPendingChangesAt(opts.batchSize); err != nil {
					return err
				}
```

db/hybrid_logical_vector.go
```diff
@@ -399,3 +406,39 @@ func (hlv *HybridLogicalVector) setPreviousVersion(source string, version uint64
...
 }
+
+// toHistoryForHLV formats blip History property for V4 replication and above
+func (hlv *HybridLogicalVector) toHistoryForHLV() string {
+	// take pv and mv from hlv if defined and add to history
+	var pvString, mvString string
+	var s strings.Builder
+	// Merge versions must be defined first if they exist
+	if hlv.MergeVersions != nil {
+		mvString = extractStringFromMap(hlv.MergeVersions)
```
bbrks: Pushing this down into its own function that creates a new string is going to be allocation heavy.

We can either iterate over the map here and use the string builder directly, or pass the string builder down into a function if you want to keep it separate.

rest/blip_client_test.go
```diff
@@ -1028,11 +1035,26 @@ func (btc *BlipTesterCollectionClient) GetVersion(docID string, docVersion DocVe
...
+		if data, ok := rev[docVersion.CV.String()]; ok && data != nil {
+			return data.body, true
+		}
 	}
 
 	return nil, false
 }
 
+func (btc *BlipTesterClient) AssertOnBlipHistory(t *testing.T, msg *blip.Message, docVersion DocVersion) {
+	if btc.SupportedBLIPProtocols[0] == db.CBMobileReplicationV4.SubprotocolString() {
```
bbrks: This check will break when we bump subprotocol again.

Need to make `btc.SupportedBLIPProtocols` numerical so we can do `>=` or `<=` operations on it.

bbrks (dismissed):

db/active_replicator.go
```diff
@@ -208,7 +208,11 @@ func connect(arc *activeReplicatorCommon, idSuffix string) (blipSender *blip.Sen
 	arc.replicationStats.NumConnectAttempts.Add(1)
 
 	var originPatterns []string // no origin headers for ISGR
-	blipContext, err := NewSGBlipContext(arc.ctx, arc.config.ID+idSuffix, originPatterns)
+	// TODO: CBG-3661 ActiveReplicator subprotocol versions
+	// - make this configurable for testing mixed-version replications
+	// - if unspecified, default to v2 and v3 until VV is supported with ISGR, then also include v4
+	protocols := []string{CBMobileReplicationV3.SubprotocolString(), CBMobileReplicationV2.SubprotocolString()} //blipContext, err := NewSGBlipContext(arc.ctx, arc.config.ID+idSuffix, originPatterns)
```
adamcfraser: Can cleanup the trailing comment here?

db/hybrid_logical_vector.go
```diff
@@ -399,3 +406,36 @@ func (hlv *HybridLogicalVector) setPreviousVersion(source string, version uint64
...
+	// Merge versions must be defined first if they exist
+	if hlv.MergeVersions != nil {
+		var pairList []string
+		for key, value := range hlv.MergeVersions {
+			vrs := Version{SourceID: key, Value: value}
+			pairList = append(pairList, vrs.String())
+		}
+		if len(pairList) > 0 {
+			separatorNeeded = true
+			s.WriteString(strings.Join(pairList, ","))
```
adamcfraser: strings.Join is an efficient way to iterate over a set and do WriteString of separator and elements, but in this case we're already iterating over the set, and already have our own strings.Builder.  Why not just add elements to s as we iterate over MergeVersions and PreviousVersions, instead of iterating twice?

adamcfraser (changes requested):
Overall LGTM, two minor tweaks requested

db/hybrid_logical_vector.go
gregns1: I agree in hindsight this makes more sense. I originally went for this approach to avoid adding a trailing ',' at the end of the string as this would break the design CBL expect. I have got round that now with a item counter so we can keep track of the elements in the map we are iterating over.

adamcfraser (approved):

PR #6642, opened 2024-01-12
CBG-3686 add idle kv ops stat

torcolvin:
Setup up two SG nodes in two modes and compared stat on server for idle ops:

- ISGR between two nodes with different couchbase servers
- two nodes serving same database (to shard DCP)

`CfgSg` operations will only show up if there's multiple SGs.

I had uncertainty about what level to put the calls, especially `GetMetadataDocument`. In this case, some idle operations are going to get called. I could move this to the bootstrap polling when we call `GetDatabaseConfigs`. On the other hand, I could move this into `config_persistentence.go` and cover xattr and non xattr case.

For heartbeater, I put the Get cases, but not the updating the node lists, which presumably happens on a rebalance. I think this is most correct.

For `CfgSg`, I added for Get operations, since again `Set` will happen for ISGR or cbgt DCP rebalance.

The concerns I have are as follows:

- It is really easy to regress on this behavior because I did manual testing only and if we change pathways of code, this will break.
- I think overcounting is OK when we are doing other operations (like updating a configuration file), because then it won't register as idle anyway, there will be more CRUD operations.
- We are going to call this stat function a few times a second - is this updating the stat too frequently? I think it's fine because it is an atomic.

## Pre-review checklist

### DIFF
--- a/base/heartbeat.go
+++ b/base/heartbeat.go
@@ -280,9 +280,9 @@ func (h *couchbaseHeartBeater) checkStaleHeartbeats(ctx context.Context) error {
 			continue
 		}
 
-		SyncGatewayStats.GlobalStats.ResourceUtilizationStats().NumIdleKvOps.Add(1)
 		timeoutDocID := heartbeatTimeoutDocID(heartbeatNodeUUID, h.keyPrefix)
 		_, _, err := h.datastore.GetRaw(timeoutDocID)
+		SyncGatewayStats.GlobalStats.ResourceUtilizationStats().NumIdleKvOps.Add(1)
 		if err != nil {
 			if !IsKeyNotFoundError(h.datastore, err) {
 				// unexpected error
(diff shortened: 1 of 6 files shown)


base/heartbeat.go
```diff
@@ -280,6 +280,7 @@ func (h *couchbaseHeartBeater) checkStaleHeartbeats(ctx context.Context) error {
 			continue
 		}
 
+		SyncGatewayStats.GlobalStats.ResourceUtilizationStats().NumIdleKvOps.Add(1)
```
adamcfraser: I think we generally want to increment the stat after the operation is performed, not before.  Doing it before the error check is fine if that's what you were trying to avoid.

base/sg_cluster_cfg.go
```diff
@@ -95,8 +95,8 @@ func (c *CfgSG) Set(cfgKey string, val []byte, cas uint64) (uint64, error) {
 		return 0, fmt.Errorf("cfg_sg: key cannot start with a colon")
 	}
 
+	SyncGatewayStats.GlobalStats.ResourceUtilizationStats().NumIdleKvOps.Add(1)
```
adamcfraser: The PR description said you added this stat for CfgSG.Get, but you've got it on Set here.

Having said that - what Cfg operations are actually being performed by cbgt when the DCP feed is idle?  I had thought it was notified about cfg changes via Subscribe(), and didn't do polling.

adamcfraser (changes requested):
Question about the cbgt ops, and suggestion to move the stat incr to after the op completes.

adamcfraser (approved):

PR #6646, opened 2024-01-17
implement body.Close checker

torcolvin:
There were no production code that was missing body.Close() but there was a lot missing in tests. This is a small resource leak.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2263/

### DIFF
--- a/rest/oidc_api_test.go
+++ b/rest/oidc_api_test.go
@@ -2287,6 +2287,7 @@ func TestEventuallyReachableOIDCClient(t *testing.T) {
 			request := createOIDCRequest(t, sessionEndpoint, token)
 			response, err := http.DefaultClient.Do(request)
 			require.NoError(t, err, "Error sending request with bearer token")
+			defer func() { assert.NoError(t, response.Body.Close()) }()
 			assert.Equal(t, http.StatusUnauthorized, response.StatusCode) // Status code when unreachable
 
 			// Now reachable - success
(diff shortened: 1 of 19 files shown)


adamcfraser (dismissed):

rest/utilities_testing_user.go
```diff
@@ -37,6 +37,7 @@ func MakeUser(t *testing.T, httpClient *http.Client, serverURL, username, passwo
 		if err != nil {
 			return true, err, resp
 		}
+		assert.NoError(t, resp.Body.Close())
```
bbrks: Is it possible for the code outside the retry loop to read the body after it has already been closed? (47-51)

I think we can also drop the outer Body.Close() if we ensure the unclosed body doesn't leak out of the retry loop.

rest/oidc_api_test.go
```diff
@@ -2287,6 +2287,7 @@ func TestEventuallyReachableOIDCClient(t *testing.T) {
 			request := createOIDCRequest(t, sessionEndpoint, token)
 			response, err := http.DefaultClient.Do(request)
 			require.NoError(t, err, "Error sending request with bearer token")
+			defer func() { assert.NoError(t, response.Body.Close()) }()
```
bbrks, line 2294: Doesn't each response body need closing? The defer will only get one of the 3 here

bbrks (changes requested):
I'm OK with adding this, just a couple of specific comments inline.

bbrks (approved):

PR #6652, opened 2024-01-24
Allow go mod tidy

torcolvin:
If you run `go.mod tidy` after https://github.com/couchbase/sync_gateway/commit/49f66ef7190f3f0ec241dbe0458eb3afaaa84dec, the go-ruleguard package would be deleted.

I declared this as an indirect dependency following best practice outlined here: https://github.com/golang/go/issues/25922#issuecomment-413898264


bbrks (approved):
Thanks!

PR #6654, opened 2024-01-24
Namespace document between revtree and versionvector tests

torcolvin:
When run with version vectors enabled, there would be two docs with `doc1Marker` existing that exist, once from the `revtree` test and once from `versionVector` test as per https://github.com/couchbase/sync_gateway/blob/beryllium/rest/blip_client_test.go#L635


torcolvin:
This isn't necessary, the tests use a separate RestTester.

PR #6655, opened 2024-01-26
CBG-3719: convert in memory format of HLV to match XDCR/CBL format

gregns1:
CBG-3719

- Changed the in memory structs to have string representation of version instead of uint64 
- Changes to tests to correctly asert on the new version values 
- Removal of json marshal and unmarshal function in Hybridlogicalvector.go as they're no longer needed
- Added interface to hybridlogicalvector.go for ease of adding methods to act upon both HybridLogicalVector and DecodedHybridLogicalVector

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2275/

### DIFF
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -19,10 +19,10 @@ import (
 )
 
 // hlvExpandMacroCASValue causes the field to be populated by CAS value by macro expansion
-const hlvExpandMacroCASValue = math.MaxUint64
+const hlvExpandMacroCASValue = "expand"
 
-// HybridLogicalVector (HLV) is a type that represents a vector of Hybrid Logical Clocks.
-type HybridLogicalVector struct {
+// DecodedHybridLogicalVector (HLV) is a type that represents a decoded vector of Hybrid Logical Clocks.
+type DecodedHybridLogicalVector struct {
 	CurrentVersionCAS uint64            // current version cas (or cvCAS) stores the current CAS at the time of replication
 	ImportCAS         uint64            // Set when an import modifies the document CAS but preserves the HLV (import of a version replicated by XDCR)
 	SourceID          string            // source bucket uuid of where this entry originated from
@@ -36,10 +36,25 @@ type Version struct {
 	// SourceID is an ID representing the source of the value (e.g. Couchbase Lite ID)
 	SourceID string `json:"source_id"`
 	// Value is a Hybrid Logical Clock value (In Couchbase Server, CAS is a HLC)
+	Value string `json:"version"`
+}
+
+// ConflictFormatVersion is a sourceID and version pair in string/uint64 format for use in conflict detection
+type DecodedVersion struct {
+	// SourceID is an ID representing the source of the value (e.g. Couchbase Lite ID)
+	SourceID string `json:"source_id"`
+	// Value is a Hybrid Logical Clock value (In Couchbase Server, CAS is a HLC)
 	Value uint64 `json:"version"`
 }
 
-func CreateVersion(source string, version uint64) Version {
+func CreateDecodedVersion(source string, version uint64) DecodedVersion {
+	return DecodedVersion{
+		SourceID: source,
+		Value:    version,
+	}
+}
+
+func CreateVersion(source, version string) Version {
 	return Version{
 		SourceID: source,
 		Value:    version,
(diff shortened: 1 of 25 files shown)


db/hybrid_logical_vector.go
```diff
@@ -122,8 +148,16 @@ func (hlv *HybridLogicalVector) IsInConflict(otherVector HybridLogicalVector) bo
...
+	var newVersionCAS uint64
+	if newVersion.Value == hlvExpandMacroCASValue {
+		// if we have macro expand value then we need to add max uint64 here for below comparison
+		newVersionCAS = math.MaxUint64
+	} else {
+		newVersionCAS = base.HexCasToUint64(newVersion.Value)
+	}
+	hlvVersionCAS := base.HexCasToUint64(hlv.Version)
+	if newVersionCAS < hlvVersionCAS {
+		return fmt.Errorf("attempting to add new version vector entry with a CAS that is less than the current version CAS value. Current cas: %s new cas %s", hlv.Version, newVersion.Value)
```
bbrks: Could possibly do this to prevent future accidental misuse of `newVersionCAS = math.MaxUint64`

```suggestion
	hlvVersionCAS := base.HexCasToUint64(hlv.Version)
	if newVersion.Value != hlvExpandMacroCASValue {
		newVersionCAS = base.HexCasToUint64(newVersion.Value)
		if newVersionCAS < hlvVersionCAS {
			return fmt.Errorf("attempting to add new version vector entry with a CAS that is less than the current version CAS value. Current cas: %s new cas %s", hlv.Version, newVersion.Value)
```

db/hybrid_logical_vector.go
```diff
@@ -272,7 +335,13 @@ func (hlv *HybridLogicalVector) AddNewerVersions(otherVector HybridLogicalVector
 		// Iterate through incoming vector previous versions, update with the version from other vector
 		// for source if the local version for that source is lower
 		for i, v := range otherVector.PreviousVersions {
-			if hlv.PreviousVersions[i] == 0 || hlv.PreviousVersions[i] < v {
+			if hlv.PreviousVersions[i] == "" {
+				hlv.setPreviousVersion(i, v)
+			}
+			// if we get here then there is entry for this source in PV dso we must check if its newer or not
```
bbrks:
```suggestion
			// if we get here then there is entry for this source in PV so we must check if its newer or not
```

db/hybrid_logical_vector.go
```diff
@@ -272,7 +335,13 @@ func (hlv *HybridLogicalVector) AddNewerVersions(otherVector HybridLogicalVector
 		// Iterate through incoming vector previous versions, update with the version from other vector
 		// for source if the local version for that source is lower
 		for i, v := range otherVector.PreviousVersions {
-			if hlv.PreviousVersions[i] == 0 || hlv.PreviousVersions[i] < v {
+			if hlv.PreviousVersions[i] == "" {
+				hlv.setPreviousVersion(i, v)
+			}
```
bbrks: Think we probably want to do this here?

```suggestion
			} else {
```

db/hybrid_logical_vector.go
```diff
@@ -443,6 +414,36 @@ func (hlv *HybridLogicalVector) toHistoryForHLV() string {
...
+	if hlv.CurrentVersionCAS != "" {
+		decodedCVCAS = base.HexCasToUint64(hlv.CurrentVersionCAS)
+	}
+	decodedHLV := DecodedHybridLogicalVector{
+		CurrentVersionCAS: decodedCVCAS,
+		Version:           decodedVersion,
+		ImportCAS:         decodedImportCAS,
+		SourceID:          hlv.SourceID,
+		PreviousVersions:  make(map[string]uint64),
+		MergeVersions:     make(map[string]uint64),
```
bbrks: Could pre-allocate size for these maps

```suggestion
		PreviousVersions:  make(map[string]uint64, len(hlv.PreviousVersions)),
		MergeVersions:     make(map[string]uint64, len(hlv.MergeVersions)),
```

bbrks (changes requested):
Think this looks good, just a couple of inline comments

db/hybrid_logical_vector.go
gregns1: Good spot!

db/hybrid_logical_vector.go
gregns1: I agree, made those changes.

bbrks (approved):
Looks good to me 👍

PR #6657, opened 2024-01-29
CBG-3724 ensure import rollback works correctly

torcolvin:
Major fix: pass a function to call cbgt.JANITOR_ROLLBACK_PINDEX when creating a pindex. This is necessary to ensure cbgt.Mgr will actually generate the index.

- fix the TestImportRollback test which was giving a false pass before because the cbgt.Mgr was getting a kick op when the nodes were rebalanced. In order for rollback to work, the cbgt.Dest RollbackEx function has to force cgbt.Mgr to actually update the PIndex. A cbgt.JANITOR_ROLLBACK_PINDEX op will force the recreation of a PIndex. This test was broken before adding code to it.
- Enhance the TestImportRollback to detect a rollback with invalid FailoverLogs. This simulates a situation where the checkpoints have been migrated from a separate bucket.

- modify TestContinuousRollback to also test with bad vbuuid FailoverEntry. This test actually would pass, because Sync Gateway, unlike cbgt, did not pass DcpStreamAddFlagStrictVBUUID. DcpStreamAddFlagStrictVBUUID forces a rollback in the case where StartSeqNo=0 vbuuid=invalid whereas without the flag it is treated like StartSeqNo=0 vbuuid=0. Changed the code to match cbgt behavior adding DcpStreamAddFlagStrictVBUUID.
- Added a separate flag in DCPClient to detect active vBuckets, which are the number of vBuckets that actually have a connection, and are not stuck in a connecting state. This differs from the existing activeVbuckets which is set from the time of NewDCPClient to when the streams end. In a continuous feed, the streams should never end.

Cleanup:

- move RollbackEx function from DCPDest to DCPCommon. DCPCommon is an embedded struct in DCPDest.
- panic if DCPCommon.Rollback is called. This should never happen given the design of cbgt.Dest, but if it did, the rollback would not work. This is different from the old behavior where it would not rollback and potentially enter a rollback loop if this function was called.
- remove maxVbNo from NewDCPDest in effort to shorten the arguments since the flag can be inferred from base.Bucket.
- Create some public structs to make tests easier:
	- ShardedImportDCPMetadata
	- CBGCfgIndexDefs, CBGTIndexNodeDefsKnown, CBGTIndexNodeDefsWanted, CBGTCfgPlanPIndexes

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2278/

### DIFF
--- a/base/dcp_client.go
+++ b/base/dcp_client.go
@@ -52,8 +52,8 @@ type DCPClient struct {
 	terminator                 chan bool                      // Used to close worker goroutines spawned by the DCPClient
 	doneChannel                chan error                     // Returns nil on successful completion of one-shot feed or external close of feed, error otherwise
 	metadata                   DCPMetadataStore               // Implementation of DCPMetadataStore for metadata persistence
-	activeVbuckets             *vBucketSet                    // vbuckets that have an open stream
-	connectedVbuckets          *vBucketSet                    // vbuckets that have a connected stream
+	activeVbuckets             map[uint16]struct{}            // vbuckets that have an open stream
+	activeVbucketLock          sync.Mutex                     // Synchronization for activeVbuckets
 	oneShot                    bool                           // Whether DCP feed should be one-shot
 	closing                    AtomicBool                     // Set when the client is closing (either due to internal or external request)
 	closeError                 error                          // Will be set to a non-nil value for unexpected error
@@ -116,8 +116,12 @@ func NewDCPClient(ctx context.Context, ID string, callback sgbucket.FeedEventCal
 		dbStats:             options.DbStats,
 		agentPriority:       options.AgentPriority,
 		collectionIDs:       options.CollectionIDs,
-		activeVbuckets:      newVBucketSet(numVbuckets),
-		connectedVbuckets:   newVBucketSet(0),
+	}
+
+	// Initialize active vbuckets
+	client.activeVbuckets = make(map[uint16]struct{})
+	for vbNo := uint16(0); vbNo < numVbuckets; vbNo++ {
+		client.activeVbuckets[vbNo] = struct{}{}
 	}
 
 	checkpointPrefix := fmt.Sprintf("%s:%v", client.checkpointPrefix, ID)
@@ -436,7 +440,6 @@ func (dc *DCPClient) openStream(vbID uint16, maxRetries uint32) error {
 
 		openStreamErr = dc.openStreamRequest(vbID)
 		if openStreamErr == nil {
-			dc.connectedVbuckets.add(vbID)
 			return nil
 		}
 
@@ -509,7 +512,7 @@ func (dc *DCPClient) openStreamRequest(vbID uint16) error {
 	}
 
 	op, openErr := dc.agent.OpenStream(vbID,
-		memd.DcpStreamAddFlagActiveOnly|memd.DcpStreamAddFlagStrictVBUUID,
+		memd.DcpStreamAddFlagActiveOnly,
 		vbMeta.VbUUID,
 		vbMeta.StartSeqNo,
 		vbMeta.EndSeqNo,
@@ -555,8 +558,11 @@ func (dc *DCPClient) verifyFailoverLog(vbID uint16, f []gocbcore.FailoverEntry) 
 }
 
 func (dc *DCPClient) deactivateVbucket(vbID uint16) {
-	numActive := dc.activeVbuckets.remove(vbID)
-	if numActive == 0 {
+	dc.activeVbucketLock.Lock()
+	delete(dc.activeVbuckets, vbID)
+	activeCount := len(dc.activeVbuckets)
+	dc.activeVbucketLock.Unlock()
+	if activeCount == 0 {
 		dc.close()
 		// On successful one-shot feed completion, purge persisted checkpoints
 		if dc.oneShot {
@@ -566,7 +572,6 @@ func (dc *DCPClient) deactivateVbucket(vbID uint16) {
 }
 
 func (dc *DCPClient) onStreamEnd(e endStreamEvent) {
-	dc.connectedVbuckets.remove(e.vbID)
 	if e.err == nil {
 		DebugfCtx(dc.ctx, KeyDCP, "Stream (vb:%d) closed, all items streamed", e.vbID)
 		dc.deactivateVbucket(e.vbID)
@@ -649,43 +654,4 @@ func getLatestVbUUID(failoverLog []gocbcore.FailoverEntry) (vbUUID gocbcore.VbUU
 
 func (dc *DCPClient) GetMetadataKeyPrefix() string {
 	return dc.metadata.GetKeyPrefix()
-}
-
-// vBucketSet is a thread-safe set of vbucket IDs
-type vBucketSet struct {
-	vbuckets map[uint16]struct{}
-	lock     sync.RWMutex
-}
-
-// newVBucketSet returns a new vBucketSet and turns on values specified up to value. Typical use would be newVBucketSet(1024) to initialize 0-1023 as values.
-func newVBucketSet(numVBuckets uint16) *vBucketSet {
-	v := vBucketSet{
-		vbuckets: make(map[uint16]struct{}),
-	}
-	for i := uint16(0); i < numVBuckets; i++ {
-		v.vbuckets[i] = struct{}{}
-	}
-	return &v
-}
-
-// add a vbucket ID from the set
-func (v *vBucketSet) add(vbID uint16) {
-	v.lock.Lock()
-	defer v.lock.Unlock()
-	v.vbuckets[vbID] = struct{}{}
-}
-
-// remove a vbucket ID from the set and returns number of vbuckets in the set.
-func (v *vBucketSet) remove(vbID uint16) int {
-	v.lock.Lock()
-	defer v.lock.Unlock()
-	delete(v.vbuckets, vbID)
-	return len(v.vbuckets)
-}
-
-// count returns the number of items in the set
-func (v *vBucketSet) count() int {
-	v.lock.RLock()
-	defer v.lock.RUnlock()
-	return len(v.vbuckets)
 }
--- a/base/dcp_client.go
+++ b/base/dcp_client.go
@@ -555,8 +555,8 @@ func (dc *DCPClient) verifyFailoverLog(vbID uint16, f []gocbcore.FailoverEntry) 
 }
 
 func (dc *DCPClient) deactivateVbucket(vbID uint16) {
-	dc.activeVbuckets.remove(vbID)
-	if dc.activeVbuckets.count() == 0 {
+	numActive := dc.activeVbuckets.remove(vbID)
+	if numActive == 0 {
 		dc.close()
 		// On successful one-shot feed completion, purge persisted checkpoints
 		if dc.oneShot {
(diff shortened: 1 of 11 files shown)


base/dcp_client.go
```diff
@@ -558,11 +555,8 @@ func (dc *DCPClient) verifyFailoverLog(vbID uint16, f []gocbcore.FailoverEntry)
 }
 
 func (dc *DCPClient) deactivateVbucket(vbID uint16) {
-	dc.activeVbucketLock.Lock()
-	delete(dc.activeVbuckets, vbID)
-	activeCount := len(dc.activeVbuckets)
-	dc.activeVbucketLock.Unlock()
-	if activeCount == 0 {
+	dc.activeVbuckets.remove(vbID)
+	if dc.activeVbuckets.count() == 0 {
```
adamcfraser: This is a change in locking - previously we were acquiring the count while holding the lock, now we're reacquiring the lock to get the count.  This introduces the possibility that multiple goroutines could execute the count=0 code, which I don't think we want.  Possibly modify remove() to return the decremented count to preserve the previous handling?

base/dcp_client.go
```diff
@@ -512,7 +509,7 @@ func (dc *DCPClient) openStreamRequest(vbID uint16) error {
 	}
 
 	op, openErr := dc.agent.OpenStream(vbID,
-		memd.DcpStreamAddFlagActiveOnly,
+		memd.DcpStreamAddFlagActiveOnly|memd.DcpStreamAddFlagStrictVBUUID,
```
adamcfraser: Do we actually want this?  It sounds like the behaviour when it's not set (ignore the vbuuid if startSeq is 0) is what we prefer.  I don't know if we need to have our DCPClient match cbgt handling here?

base/dcp_dest.go
```diff
@@ -37,26 +37,25 @@ type SGDest interface {
...
+func NewDCPDest(ctx context.Context, callback sgbucket.FeedEventCallbackFunc, bucket Bucket, persistCheckpoints bool,
+	dcpStats *expvar.Map, feedID string, importPartitionStat *SgwIntStat, checkpointPrefix string, metaKeys *MetadataKeys, rollback func()) (SGDest, context.Context, error) {
 
 	// TODO: Metadata store?
 	metadataStore := bucket.DefaultDataStore()
-	dcpCommon, err := NewDCPCommon(ctx, callback, bucket, metadataStore, maxVbNo, persistCheckpoints, dcpStats, feedID, checkpointPrefix, metaKeys)
+	dcpCommon, err := NewDCPCommon(ctx, callback, bucket, metadataStore, persistCheckpoints, dcpStats, feedID, checkpointPrefix, metaKeys, rollback)
```
adamcfraser: Are we moving where maxVbNo is computed in this PR for any reason?

base/dcp_receiver.go
```diff
@@ -29,8 +29,8 @@ func makeFeedEventForMCRequest(rq *gomemcached.MCRequest, opcode sgbucket.FeedOp
 	return makeFeedEvent(rq.Key, rq.Body, rq.DataType, rq.Cas, ExtractExpiryFromDCPMutation(rq), rq.VBucket, 0, opcode)
 }
 
-// vBucketMetaData is an internal struct that is exposed to enable json marshaling.
-type vBucketMetaData struct {
+// ShardedImportDCPMetadata is an internal struct that is exposed to enable json marshaling, used by sharded import feed. It differs from DCPMetadata for reasons of backward compatibility.
```
adamcfraser: It's not just backward compatibility - this specifically needs to match cbgt's (non-exported) metadata format, https://github.com/couchbase/cbgt/blob/master/feed_dcp_gocbcore.go#L1134.  I think that's the main reason for the difference between this and DCPMetadata.

db/import_pindex.go
```diff
@@ -75,40 +75,37 @@ func getNewPIndexImplType(ctx context.Context) func(indexType, indexParams, path
...
 		if err != nil {
 			base.ErrorfCtx(ctx, "Error creating NewImportDest during NewImportPIndexImpl: %v", err)
 		}
 		return nil, importDest, err
 	}
 	return newImportPIndexImpl
 }
-func OpenImportPIndexImpl(indexType, path string, restart func()) (cbgt.PIndexImpl, cbgt.Dest, error) {
+
+// openImportPIndexImpl is required to have an implementation from cbgt.PIndexImplType.Open. When this function fails, PIndexImplType will fall back to using PIndexImplType.OpenUsing
```
adamcfraser: Is this correct?  (that Open falls back to OpenUsing). I don't see that happening in the current cbgt - I don't think cbgt uses Open at all any more (OpenPIndexImpl is only called by tests)

base/dcp_common.go
```diff
@@ -58,10 +58,12 @@ type DCPCommon struct {
 	feedID                 string                         // Unique feed ID, used for logging
 	loggingCtx             context.Context                // Logging context, prefixes feedID
 	checkpointPrefix       string                         // DCP checkpoint key prefix
+	janitorRollback        func()                         // This function will trigger a janitor_pindex_rollback
```
adamcfraser: I thought this should be stored on DCPDest, since it's a cbgt-specific artifact, but I guess DCPDest is actually the only usage of DCPCommon now that we've switched the other feeds over to DCPClient?

adamcfraser (changes requested):

db/import_pindex.go
torcolvin: I would argue this is a bug in cbgt, but it is required by https://github.com/couchbaselabs/cbgt/blob/master/manager_planner.go#L294

It should check `OpenUsing` instead of `Open` here.

base/dcp_dest.go
torcolvin: The reason was to make this call stack less complicated because we can and should infer it from `Bucket`. It's not necessary in this PR but I thought when I'm adding argument 11, I might want to see if they are necessary.

base/dcp_receiver.go
torcolvin: Updated this comment, thanks.

base/dcp_client.go
torcolvin: Done.

base/dcp_common.go
```diff
@@ -160,23 +161,19 @@ func (c *DCPCommon) getMetaData(vbucketId uint16) (
...
-// RollbackEx should be called by cbdatasource - Rollback required to maintain the interface.  In the event
-// it's called, logs warning and does a hard reset on metadata for the vbucket.  Returns error if metadata
-// persistence fails
-func (c *DCPCommon) rollback(vbucketId uint16, rollbackSeq uint64) error {
-	WarnfCtx(c.loggingCtx, "DCP Rollback request.  Expected RollbackEx call - resetting vbucket %d to 0.", vbucketId)
-	c.dbStatsExpvars.Add("dcp_rollback_count", 1)
-	c.updateSeq(vbucketId, 0, false)
-	err := c.setMetaData(vbucketId, nil, true)
-	return err
+func (c *DCPCommon) Rollback(partition string, rollbackSeq uint64) error {
```
adamcfraser: Generally I don't think we want to be implementing the cbgt interfaces (Dest, Reciever) in DCPCommon - the idea with DCPCommon is to define the common handling of these interfaces.  

It looks like we're actually still implementing Rollback in dcp_dest.go - was there other motivation for defining this here?

base/dcp_common.go
```diff
@@ -160,23 +161,19 @@ func (c *DCPCommon) getMetaData(vbucketId uint16) (
...
 }
 
-// RollbackEx includes the vbucketUUID needed to reset the metadata correctly
-func (c *DCPCommon) rollbackEx(vbucketId uint16, vbucketUUID uint64, rollbackSeq uint64, rollbackMetaData []byte) error {
+// rollbackEx is called when a DCP stream issues a rollback. The metadata persisted for a given uuid and sequence number and then cbgt.Mgr JANITOR_ROLLBACK_PINDEX is issued via janitorRollback function.
+func (c *DCPCommon) rollbackEx(vbucketId uint16, vbucketUUID uint64, rollbackSeq uint64, rollbackMetaData []byte, janitorRollback func()) error {
 	WarnfCtx(c.loggingCtx, "DCP RollbackEx request - rolling back DCP feed for: vbucketId: %d, rollbackSeq: %x.", vbucketId, rollbackSeq)
 	c.dbStatsExpvars.Add("dcp_rollback_count", 1)
 	c.updateSeq(vbucketId, rollbackSeq, false)
+	// if we fail to set the metadata, and don't retry, the
```
adamcfraser: This comment looks like it got truncated.

base/dcp_common.go
```diff
@@ -60,8 +60,9 @@ type DCPCommon struct {
...
 }
 
-func NewDCPCommon(ctx context.Context, callback sgbucket.FeedEventCallbackFunc, bucket Bucket, metaStore DataStore,
-	maxVbNo uint16, persistCheckpoints bool, dbStats *expvar.Map, feedID, checkpointPrefix string, metaKeys *MetadataKeys) (*DCPCommon, error) {
+// NewDCPCommon creates a new DCPCommon which manages updates coming from a cbgt-based DCP feed. The callback function will receive events from a DCP feed. The bucket is the gocb bucket to stream events from. It stores checkpoints in the metaStore collection prefixes from metaKeys + checkpointPrefix. The feed name will start with feedID and DCPCommon will add unique string. Specific stats for DCP are stored in expvars rather than SgwStats. The janitorRollback function is supplied by the global cbgt.PIndexImplType.New function, for initial opening of a partition index, and cbgt.PIndexImplType.OpenUsing for reopening of a partition index. The rollback function provides a way to pass cbgt.JANITOR_ROLLBACK_PINDEX to cbgt.Mgr and is supplied.
+func NewDCPCommon(ctx context.Context, callback sgbucket.FeedEventCallbackFunc, bucket Bucket, maxVbNo uint16, metaStore DataStore,
```
adamcfraser: It looks like parameter ordering got (unintentionally?) swapped here (maxVbNo, metaStore).

base/dcp_dest.go
```diff
@@ -188,13 +191,19 @@ func (d *DCPDest) OpaqueSet(partition string, value []byte) error {
 	return nil
 }
 
+// Rollback is required by cbgt.Dest interface but will not work when called by Sync Gateway as we need additional information to perform a rollback. Due to the design of cbgt.Dest this will not be called without a programming error.
 func (d *DCPDest) Rollback(partition string, rollbackSeq uint64) error {
-	return d.rollback(partitionToVbNo(d.loggingCtx, partition), rollbackSeq)
+	panic("Only RollbackEx should be called, this function is required to be implmented by cbgt.Dest interface. This function does not provide Sync Gateway with enough information to rollback.")
```
adamcfraser: I'd prefer to avoid panicking and continue to log a warning (or possibly error, if you think it's that severe).

adamcfraser (changes requested):

base/dcp_common.go
torcolvin: This was just a mistake I left behind.

base/dcp_dest.go
torcolvin: This error will result in the feed not running, turned this into a warning.

adamcfraser (approved):

PR #6659, opened 2024-02-02
Include BLIP context ID in _blipsync response with X-Correlation-ID…

bbrks:
CBG-3747

- Include BLIP context ID in _blipsync response with `X-Correlation-ID` header

Screenshot 2024-02-02 at 09 17 29

- n/a

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -77,6 +77,8 @@ func newBlipHandler(ctx context.Context, bc *BlipSyncContext, db *Database, seri
 type BLIPSyncContextClientType string
 
 const (
+	BLIPCorrelationIDResponseHeader = "X-Correlation-ID"
+
 	BLIPSyncClientTypeQueryParam = "client"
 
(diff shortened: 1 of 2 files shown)


torcolvin (commented):
Is it worth writing a test for this? 

Is it worth naming the header so it's clear it is the blip correlation id and not the http correlation id? It is somewhat regrettable that these both look similar in SG logs?

bbrks:
`X-Correlation-ID` is a de facto header name (an alternative being `X-Request-ID`)

I'm not even sure why we have BLIP-specific context IDs, given we can always link them back to an originating HTTP request ID... but the time to correct that has long gone.

bbrks:
Also tried to write a test but couldn't because:
- The HTTP request needs to be upgradable to a websocket/BLIP connection.
- `websocket.Dial` does not return the HTTP response headers.

torcolvin (approved):

PR #6660, opened 2024-02-02
CBG-3741: Throttle concurrent revs per-replication

bbrks:
CBG-3741

- Throttles the maximum number of in-flight/parallel `rev` messages for a given client for push and pull.
  - Limits memory usage and other work associated with document retrieval, even if cached.
  - Should still allow clients to be fed a continous stream of revisions assuming replications were bandwidth bottlenecked.
  - Clients can send `revs` without being throttled, but the handling of messages, and any repsonses will be throttled.
  - Configurable. Default 5 (min 5, max 200) via startup config for tuning based on SG node sizing.
  - New stats to monitor rev throttling (`WriteThrottledCount` and `WriteThrottledTime`)
- Made changes batch throttle configurable in the same way.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2299/

### DIFF
--- a/base/stats.go
+++ b/base/stats.go
@@ -483,6 +483,10 @@ type CBLReplicationPushStats struct {
 	ProposeChangeTime *SgwIntStat `json:"propose_change_time"`
 	// Total time spent processing writes. Measures complete request-to-response time for a write.
 	WriteProcessingTime *SgwIntStat `json:"write_processing_time"`
+	// WriteThrottledCount is the cumulative number of writes that were throttled.
+	WriteThrottledCount *SgwIntStat `json:"write_throttled_count"`
+	// WriteThrottledTime is the cumulative time spent throttling writes.
+	WriteThrottledTime *SgwIntStat `json:"write_throttled_time"`
 }
 
 // CollectionStats are stats that are tracked on a per-collection basis.
@@ -1497,6 +1501,14 @@ func (d *DbStats) initCBLReplicationPushStats() error {
 	if err != nil {
 		return err
 	}
+	resUtil.WriteThrottledCount, err = NewIntStat(SubsystemReplicationPush, "write_throttled_count", StatUnitNoUnits, WriteThrottledCountDesc, StatAddedVersion3dot2dot0, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, labelKeys, labelVals, prometheus.CounterValue, 0)
+	if err != nil {
+		return err
+	}
+	resUtil.WriteThrottledTime, err = NewIntStat(SubsystemReplicationPush, "write_throttled_time", StatUnitNanoseconds, WriteThrottledTimeDesc, StatAddedVersion3dot2dot0, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, labelKeys, labelVals, prometheus.CounterValue, 0)
+	if err != nil {
+		return err
+	}
 
 	d.CBLReplicationPushStats = resUtil
 	return nil
@@ -1510,6 +1522,8 @@ func (d *DbStats) unregisterCBLReplicationPushStats() {
 	prometheus.Unregister(d.CBLReplicationPushStats.ProposeChangeCount)
 	prometheus.Unregister(d.CBLReplicationPushStats.ProposeChangeTime)
 	prometheus.Unregister(d.CBLReplicationPushStats.WriteProcessingTime)
+	prometheus.Unregister(d.CBLReplicationPushStats.WriteThrottledCount)
+	prometheus.Unregister(d.CBLReplicationPushStats.WriteThrottledTime)
 }
 
 func (d *DbStats) CBLReplicationPush() *CBLReplicationPushStats {
--- a/base/stats.go
+++ b/base/stats.go
@@ -83,6 +83,7 @@ const (
 	StatAddedVersion3dot1dot0     = "3.1.0"
 	StatAddedVersion3dot1dot2     = "3.1.2"
 	StatAddedVersion3dot1dot3dot1 = "3.1.3.1"
+	StatAddedVersion3dot1dot4     = "3.1.4"
 	StatAddedVersion3dot2dot0     = "3.2.0"
 
 	StatDeprecatedVersionNotDeprecated = ""
@@ -456,8 +457,12 @@ type CBLReplicationPullStats struct {
 	// The total amount of time processing rev messages (revisions) during pull revision.
 	RevProcessingTime *SgwIntStat `json:"rev_processing_time"`
 	// The total number of rev messages processed during replication.
-	RevSendCount  *SgwIntStat `json:"rev_send_count"`
-	RevErrorCount *SgwIntStat `json:"rev_error_count"`
+	RevSendCount *SgwIntStat `json:"rev_send_count"`
+	// The total number of rev messages that were throttled.
+	ReadThrottledCount *SgwIntStat `json:"read_throttled_count"`
+	// The total time spent throttling rev messages.
+	ReadThrottledTime *SgwIntStat `json:"read_throttled_time"`
+	RevErrorCount     *SgwIntStat `json:"rev_error_count"`
 	// The total amount of time between Sync Gateway receiving a request for a revision and that revision being sent.
 	//
 	// In a pull replication, Sync Gateway sends a /_changes request to the client and the client responds with the list of revisions it wants to receive.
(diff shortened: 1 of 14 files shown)


base/stats.go
```diff
@@ -1497,6 +1501,14 @@ func (d *DbStats) initCBLReplicationPushStats() error {
 	if err != nil {
 		return err
 	}
+	resUtil.WriteThrottledCount, err = NewIntStat(SubsystemReplicationPush, "write_throttled_count", StatUnitNoUnits, WriteThrottledCountDesc, StatAddedVersion3dot2dot0, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, labelKeys, labelVals, prometheus.CounterValue, 0)
```
torcolvin: If this is going into 3.1.4 then this is should have added version as such.

base/stats.go
```diff
@@ -1497,6 +1501,14 @@ func (d *DbStats) initCBLReplicationPushStats() error {
...
+	if err != nil {
+		return err
+	}
+	resUtil.WriteThrottledTime, err = NewIntStat(SubsystemReplicationPush, "write_throttled_time", StatUnitNanoseconds, WriteThrottledTimeDesc, StatAddedVersion3dot2dot0, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, labelKeys, labelVals, prometheus.CounterValue, 0)
```
torcolvin: If this is going into 3.1.4 then this is should have added version as such.

base/stats_descriptions.go
```diff
@@ -205,6 +205,9 @@ const (
...
 		"(b). Assessing the benefit of adding additional Sync Gateway nodes, as it can point to Sync Gateway being a bottleneck (c). Troubleshooting slow push replication, in which case it ought to be considered in conjunction with sync_function_time"
 
+	WriteThrottledCountDesc = "The total number of times a revision was throttled during push replication from clients. The write_throttled_count stat can be useful to to determine an appropriate limit of concurrent revisions for each client. There's a direct tradeoff with memory and CPU usage for replicating clients and large amounts of concurrent revisions."
+	WriteThrottledTimeDesc  = "The total time waiting for an available slot to handle a pushed revision after being throttled. The write_throttled_time stat can be useful to determine whether clients are waiting too long for an available slot to push a revision. There's a direct tradeoff with memory and CPU usage for replicating clients and large amounts of concurrent revisions."
```
torcolvin: Note that this time is nanoseconds? Are all our times nanoseconds?

rest/config_startup.go
```diff
@@ -138,9 +143,11 @@ type AuthConfig struct {
 }
 
 type ReplicatorConfig struct {
-	MaxHeartbeat              *base.ConfigDuration `json:"max_heartbeat,omitempty"    help:"Max heartbeat value for _changes request"`
-	BLIPCompression           *int                 `json:"blip_compression,omitempty" help:"BLIP data compression level (0-9)"`
-	MaxConcurrentReplications int                  `json:"max_concurrent_replications,omitempty" help:"Maximum number of replication connections to the node"`
+	MaxHeartbeat                *base.ConfigDuration `json:"max_heartbeat,omitempty"    help:"Max heartbeat value for _changes request"`
+	BLIPCompression             *int                 `json:"blip_compression,omitempty" help:"BLIP data compression level (0-9)"`
+	MaxConcurrentReplications   int                  `json:"max_concurrent_replications,omitempty"    help:"Maximum number of replication connections to the node"`
+	MaxConcurrentChangesBatches int                  `json:"max_concurrent_changes_batches,omitempty" help:"Maximum number of changes batches to process concurrently per replication (1-5)"`
```
torcolvin: It would be nice to extend a test in `config_startup_test.go` just in case we refactor parsing.

And get a test for the config validation in `server_context.go`.

Is there a reason to not make this `*int` to represent that this is optional? Should we state the default in the help, or that this is optional?

db/blip_sync_context.go
```diff
@@ -354,12 +368,18 @@ func (bsc *BlipSyncContext) handleChangesResponse(sender *blip.Sender, response
 				}
 			}
 
-			var err error
-			if deltaSrcRevID != "" {
-				err = bsc.sendRevAsDelta(sender, docID, revID, deltaSrcRevID, seq, knownRevs, maxHistory, handleChangesResponseDbCollection, collectionIdx)
-			} else {
-				err = bsc.sendRevision(sender, docID, revID, seq, knownRevs, maxHistory, handleChangesResponseDbCollection, collectionIdx)
-			}
+			err := func() error {
```
torcolvin: Given that the below callbacks are only called when ISGR is in place, does it make sense to drop this function here and put the defer until the end of `handleChangesResponse` including the possible callbacks?

The rest of the time spent in `handleChangesResponse` is not in sending a message back so it won't overwhelm the network here so I see a good reason not to include the callbacks in the throttling here, but I worry about a potential refactor meaning that we are throttling too little.

db/blip_handler.go
```diff
@@ -932,6 +944,19 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 		}
 	}()
 
+	// throttle concurrent revs
+	if cap(bh.inFlightRevsThrottle) > 0 {
```
torcolvin, line 948: Is the `if cap(bh.inFlightRevsThrottle) > 0 {` necessary? We always specify a non zero value and if you did have a zero value it would not hit the default case anyway?

base/stats_descriptions.go
bbrks: Will include in description 👍

Yes generally, unless they specify otherwise.

db/blip_handler.go
bbrks: If `cap == 0` we end up deadlocking because there's nothing to release (read) the throttle channel.

This is more a defensive measure against:
- us forgetting to initialize the channel with a buffer
- or somehow somebody bypasses the config validation and sets size to `0`

rest/config_startup.go
bbrks: > It would be nice to extend a test in config_startup_test.go just in case we refactor parsing.

What in particular were you thinking I test in there? I'm not sure what parsing refactor means.

> And get a test for the config validation in server_context.go.

👍 

> Is there a reason to not make this *int to represent that this is optional? Should we state the default in the help, or that this is optional?

I don't think there's a reason I did this. Will change for consistency.

db/blip_sync_context.go
bbrks: Dropped these changes entirely given the single-threadedness of pull replication changes batches. No need for the throttle.

rest/config_startup.go
torcolvin: For the test, I think I commented on the wrong line, but I was thinking to make sure that command line flags `replicator.max_concurrent_changes_batches` etc will make it the startup config.

db/blip_handler.go
torcolvin: Thanks, makes sense.

rest/blip_api_crud_test.go
```diff
@@ -1198,6 +1198,48 @@ func TestBlipSendAndGetRev(t *testing.T) {
 	assert.True(t, deletedValue)
 }
 
+// Sends many revisions concurrently and ensures that SG limits the processing on the server-side with MaxConcurrentRevs
+func TestBlipSendConcurrentRevs(t *testing.T) {
+
+	base.SetUpTestLogging(t, base.LevelTrace, base.KeyHTTP, base.KeySync, base.KeySyncMsg)
```
torcolvin: should we drop logging from trace here?

rest/config_startup.go
```diff
@@ -138,9 +143,11 @@ type AuthConfig struct {
 }
 
 type ReplicatorConfig struct {
-	MaxHeartbeat              *base.ConfigDuration `json:"max_heartbeat,omitempty"    help:"Max heartbeat value for _changes request"`
-	BLIPCompression           *int                 `json:"blip_compression,omitempty" help:"BLIP data compression level (0-9)"`
-	MaxConcurrentReplications int                  `json:"max_concurrent_replications,omitempty" help:"Maximum number of replication connections to the node"`
+	MaxHeartbeat                *base.ConfigDuration `json:"max_heartbeat,omitempty"    help:"Max heartbeat value for _changes request"`
+	BLIPCompression             *int                 `json:"blip_compression,omitempty" help:"BLIP data compression level (0-9)"`
+	MaxConcurrentReplications   int                  `json:"max_concurrent_replications,omitempty"    help:"Maximum number of replication connections to the node"`
```
torcolvin, line 148: Is it worth fixing this to be `*int` while you are in the neighborhood? Could take or leave this.

bbrks: I think I'll leave this - the difference there is `0` is the default value for this option.

torcolvin (approved):

PR #6680, opened 2024-02-12
CBG-3688 make sure kv_pool_size is overriden correctly for DCP

torcolvin:
- create two functions for getting the connection string
	- GetGoCBConnString
	- GetGoCBConnStringForDCP

Simplifications to API:

- Fix a confusing invariant where GetGoCBConnString would initialize KvPoolSize on BucketSpec by requiring a constructor that does string parsing of server value to pull out kv_pool_size. This will fail if an invalid kv_pool_size value is created.

- Move the creation of modified connection string for the bootstrap connection to the caller (rest/main.go). This means that both serverless and regular callers will have values set on the connection string for bootstrap connection. Before these values were only set for serverless bootstrap:
  	- max_perhost_idle_http_connections
	- max_idle_http_connections
	- idle_http_connection_timeout

Questions:

- We only set `ca_cert_path` for Bucket connections and not for the bootstrap connection? The only way that I imagine that this works is if the ca cert is in the system already or embedded in the pem file. I tried to use our x509 tests but they are broken because they don't run automatically. I tried to fix them but it is somewhat non trivial.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2303/

### DIFF
--- a/base/bucket.go
+++ b/base/bucket.go
@@ -215,10 +215,6 @@ func (spec *BucketSpec) getGoCBConnString(forceKvPoolSize *int) (string, error) 
 		asValues.Set(kvPoolSizeKey, strconv.Itoa(spec.KvPoolSize))
 	}
 
-	if spec.CACertPath != "" {
-		asValues.Set("ca_cert_path", spec.CACertPath)
-	}
-
 	connSpec.Options = asValues
 	return connSpec.String(), nil
 }
--- a/base/bucket.go
+++ b/base/bucket.go
@@ -118,11 +118,6 @@ type CouchbaseBucketType int
 
 // Full specification of how to connect to a bucket
 type BucketSpec struct {
-	BucketSpecOptions
-	KvPoolSize int // gocb kv_pool_size - number of pipelines per node. Initialized on GetGoCBConnString
-}
-
-type BucketSpecOptions struct {
 	Server, BucketName, FeedType  string
 	Auth                          AuthHandler
 	Certpath, Keypath, CACertPath string         // X.509 auth parameters
@@ -140,38 +135,6 @@ const defaultInitialRetryMS = 5
 // Create a RetrySleeper based on the default properties.  Used to retry bucket operations after transient errors.
 func DefaultRetrySleeper() RetrySleeper {
 	return CreateDoublingSleeperFunc(defaultNumRetries, defaultInitialRetryMS)
-}
-
-// NewBucketSpec creates a new BucketSpec from server and other settings.
-func NewBucketSpec(options BucketSpecOptions) (*BucketSpec, error) {
-	return newBucketSpec(options, false)
-}
-
-// NewBucketSpec creates a new BucketSpec from options with serverless settings for kv_pool_size.
-func NewBucketSpecServerless(options BucketSpecOptions) (*BucketSpec, error) {
-	return newBucketSpec(options, true)
-}
-
-// newBucketSpec creates a new BucketSpec from options. Will parse connection string for kv_pool_size to set value, if a gocb bucket.
-func newBucketSpec(options BucketSpecOptions, isServerless bool) (*BucketSpec, error) {
-	spec := BucketSpec{
-		BucketSpecOptions: options,
-	}
-	if spec.IsWalrusBucket() {
-		return &spec, nil
-	}
-	kvPoolSize, err := getKvPoolSize(spec.Server)
-	if err != nil {
-		return nil, err
-	}
-	if kvPoolSize != nil {
-		spec.KvPoolSize = *kvPoolSize
-	} else if isServerless {
-		spec.KvPoolSize = DefaultGocbKvPoolSizeServerless
-	} else {
-		spec.KvPoolSize = DefaultGocbKvPoolSize
-	}
-	return &spec, nil
 }
 
 func (spec BucketSpec) IsWalrusBucket() bool {
@@ -211,8 +174,8 @@ func (spec *BucketSpec) getGoCBConnString(forceKvPoolSize *int) (string, error) 
 
 	if forceKvPoolSize != nil {
 		asValues.Set(kvPoolSizeKey, strconv.Itoa(*forceKvPoolSize))
-	} else {
-		asValues.Set(kvPoolSizeKey, strconv.Itoa(spec.KvPoolSize))
+	} else if asValues.Get(kvPoolSizeKey) == "" {
+		asValues.Set(kvPoolSizeKey, strconv.Itoa(DefaultGocbKvPoolSize))
 	}
 
 	connSpec.Options = asValues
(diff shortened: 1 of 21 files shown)


base/gocb_connection_string.go
```diff
@@ -0,0 +1,116 @@
...
+func DefaultGoCBConnStringParams() *GoCBConnStringParams {
+	return &GoCBConnStringParams{
+		KvPoolSize:    DefaultGocbKvPoolSize,
+		KvBufferSize:  0,
+		DcpBufferSize: 0,
+	}
+}
+
+// ServerlessGoCBConnStringParams returns a GoCBConnStringParams with the default values for serverless deployments.
+func ServerlessGoCBConnStringParams() *GoCBConnStringParams {
```
adamcfraser: Should this be DefaultServerlessGoCBConnStringParams?

base/gocb_connection_string.go
```diff
@@ -0,0 +1,116 @@
...
+	return &GoCBConnStringParams{
+		KvPoolSize:    DefaultGocbKvPoolSizeServerless,
+		KvBufferSize:  DefaultKvBufferSizeServerless,
+		DcpBufferSize: DefaultDCPBufferServerless,
+	}
+}
+
+// getGoCBConnSpec returns a gocb connection spec based on the server string. If defaults is not nil, it will override the default values in the connection string.
```
adamcfraser:
```suggestion
// getGoCBConnSpec returns a gocb connection spec based on the server string. The provided defaults will be used only when the corresponding property is not set in the connection string.
```

base/gocb_connection_string.go
```diff
@@ -0,0 +1,116 @@
...
+import (
+	"fmt"
+	"net/url"
+	"strconv"
+
+	"github.com/couchbaselabs/gocbconnstr"
+)
+
+const (
+	kvPoolSizeKey = "kv_pool_size"
```
adamcfraser: I like the idea of making this a constant, I think it makes sense to do the same for kv_buffer_size and dcp_buffer_size?

base/bucket.go
```diff
@@ -119,6 +118,11 @@ type CouchbaseBucketType int
 
 // Full specification of how to connect to a bucket
 type BucketSpec struct {
+	BucketSpecOptions
```
adamcfraser: This refactor feels arbitrary - I don't think KvPoolSize is any more or less of an 'option' than things like BucketOpTimeout.  What's the problem that this refactor fixes?

adamcfraser (changes requested):
A few comments/questions.

base/bucket.go
torcolvin: KvPoolSize isn't an input parameter, it is only ever read from the connection string. It is parsed from the connection string and then stored when creating a bucket to determine how to throttle kvops in the below code.

```
	numPools := 1
	if spec.KvPoolSize > 0 {
		numPools = spec.KvPoolSize
	}
	gocbv2Bucket.kvOps = make(chan struct{}, MaxConcurrentSingleOps*nodeCount*numPools)
```

A different way of addressing the immutability of the struct would be to never set this value and avoid the use of a constructor for this struct. Then, inside `GetGocbV2BucketFromCluster` you could call `getKvPoolSize(spec.Server)` which would reparse the connection string. The reason I didn't do it like this was to only parse the connection string once but this is such a cheap operation (string parsing) compared to bucket opening that it would be fine.

Do you think this would be more readable?

torcolvin: I am not set on https://github.com/couchbase/sync_gateway/pull/6680/commits/2f038b936796ca9afd7557675a86ad396328f332 but it simplifies the PR.

This will behave identically to the original PR but in order for serverless values to work, you have to modify the connection string on the `BucketSpec` to have `kv_pool_size=1`. This is done in `GetBucketSpec` anyway to add `dcp_buffer_size` and `kv_buffer_size` custom values so I think this is probably a reasonable refactor.

The downsides (which aren't bad):

- the setting of an implicit `kv_pool_size` is set as `DefaultGocbKvPoolSize` inside some functions - `BucketSpec.getGoCBConnString` and `BucketSpec.GetKvPoolSize` and they have the potential to get out of sync. 
- the place that we will validate that we haven't passed `kv_pool_size` twice in a connection string will take place when opening a bucket and not loading a database. This is unlikely behavior we didn't test for anyway.
- we parse the connection string twice when creating a `Bucket`, once to get the connection string for a cluster and another for figuring out the `kv_pool_size`. This is a cheap string operation vs a kv op for a rare condition.

base/bucket.go
```diff
@@ -156,86 +152,36 @@ func (spec BucketSpec) UseClientCert() bool {
...
 
-// GetGoCBConnString builds a gocb connection string based on BucketSpec.Server.
-func (spec *BucketSpec) GetGoCBConnString(params *GoCBConnStringParams) (string, error) {
-	if params == nil {
-		params = &GoCBConnStringParams{}
-	}
-	params.FillDefaults()
-	connSpec, err := gocbconnstr.Parse(spec.Server)
+// getGoCBConnString builds a gocb connection string based on BucketSpec.server values. This is used for bucket connections. KvPoolSize can be forced despite the values of the connection values.
+func (spec *BucketSpec) getGoCBConnString(forceKvPoolSize *int) (string, error) {
```
adamcfraser, line 166: Did you consider an approach that would allow us to override other values in future?  (e.g. kv_buffer_size, dcp_buffer_size), or do you think that's unlikely to be necessary?  I was thinking of something like `connStringOverrides map[string]int`, but I can't decide whether that's over-engineering the solution.

base/collection.go
```diff
@@ -165,11 +165,14 @@ func GetGocbV2BucketFromCluster(ctx context.Context, cluster *gocb.Cluster, spec
 	if mgmtEpsErr != nil && len(mgmtEps) > 0 {
 		nodeCount = len(mgmtEps)
 	}
-	numPools := 1
-	if spec.KvPoolSize > 0 {
-		numPools = spec.KvPoolSize
+
+	numPools, err := spec.GetKvPoolSize()
```
adamcfraser: It feels like GetKvPoolSize should be a helper function that takes a connString, and doesn't use the spec?  That way we can be certain that the value here matches what's actually getting sent to gocb (on the conn string), and we don't need to have any logic inside GetKvPoolSize around default values, etc.

That would also let us get rid of the test-only getKVConnectionPol [sic], and just use the new connString based function for test verification.

base/dcp_sharded.go
```diff
@@ -296,7 +291,7 @@ func initCBGTManager(ctx context.Context, bucket Bucket, spec BucketSpec, cfgSG
 
 	// Since cbgt initializes a buffer per CBS node per partition in most cases (vbuckets in partitions can't be grouped by CBS node),
 	// setting the small buffer size used in cbgt 1.3.2.  (see CBG-3341 for potential optimization of this value)
-	options["kvConnectionBufferSize"] = "16384"
+	options["kvConnectionBufferSize"] = "16384" // This value will be overridden by the BucketSpec connection string.
```
adamcfraser: I think it might be appropriate to log something at warn in the case where the connection string is setting a larger value for this, to help us triage possible memory issues that would result.  The message would be along the lines of "dcp_buffer_size is set to %d on the connection string - this will result in increased memory usage"

adamcfraser (changes requested):
Overall this makes sense.  Added a suggestion to go further with removing KvPoolSize from the bucket spec and just extract it from a computed connstring as needed.

base/bucket.go
torcolvin: I think it isn't necessary for now, unless we decide that we need to for dcp_buffer_size in cbgt. I think when we need to expand it, we'd should cross the bridge when we come to it.

base/dcp_sharded.go
torcolvin: For clarity, this is actually `kv_buffer_size`: https://sourcegraph.com/github.com/couchbase/gocbcore@8c8c79204421aa47127a042ac2ac560d87920645/-/blob/agent_config.go?L411

base/collection.go
torcolvin: I changed this to a generic helper function so I could use it for `kv_pool_size` and `kv_buffer_size`.

base/dcp_sharded.go
torcolvin: Implemented this idea.

rest/server_context.go
```diff
@@ -455,9 +455,33 @@ func (sc *ServerContext) getOrAddDatabaseFromConfig(ctx context.Context, config
 	return sc._getOrAddDatabaseFromConfig(ctx, config, options)
 }
 
-func GetBucketSpec(ctx context.Context, config *DatabaseConfig, serverConfig *StartupConfig) (spec base.BucketSpec, err error) {
+// GetBucketSpec returns a BucketSpec from a given DatabaseConfig and StartupConfig.
+func GetBucketSpec(ctx context.Context, config *DatabaseConfig, serverConfig *StartupConfig) (*base.BucketSpec, error) {
```
adamcfraser: What's the benefit in changing to a pointer return for BucketSpec here?

adamcfraser (commented):
Generally looks good, wanted one bit of clarification.

rest/server_context.go
torcolvin: I had changed it to a pointer in a previous iteration because the value that it returns is effectively nil and you wouldn't want to use it. It doesn't matter in this case because we check the error value of the function and performing a construction/copy on BucketSpec is rare and cheap.

I restored the previous behavior.

adamcfraser (approved):

PR #6682, opened 2024-02-13
Speed up `auth` package tests by setting lower test-only bcrypt cost

bbrks:
Initialize most `auth` tests with `NewTestAuthenticator`, which lowers the bcrypt cost.

## Comparison
```diff
- ok  	github.com/couchbase/sync_gateway/auth	8.656s
+ ok  	github.com/couchbase/sync_gateway/auth	2.361s
```

- n/a (unit test covered)

### DIFF
--- a/auth/auth_test.go
+++ b/auth/auth_test.go
@@ -27,6 +27,11 @@ import (
 	"gopkg.in/square/go-jose.v2/jwt"
 )
 
+func NewTestAuthenticator(t testing.TB, dataStore sgbucket.DataStore, channelComputer ChannelComputer, opts AuthenticatorOptions) *Authenticator {
+	opts.BcryptCost = bcrypt.MinCost // lower cost for testing speedup
+	return NewAuthenticator(dataStore, channelComputer, opts)
+}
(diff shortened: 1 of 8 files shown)


auth/auth_test.go
```diff
@@ -27,6 +27,11 @@ import (
 	"gopkg.in/square/go-jose.v2/jwt"
 )
 
+func NewTestAuthenticator(t testing.TB, dataStore sgbucket.DataStore, channelComputer ChannelComputer, opts AuthenticatorOptions) *Authenticator {
```
gregns1, line 30: Nitpick, but do we need t passed in here? Doesn't look to be used

gregns1 (approved):
LGTM, one nitpick but happy to merge as is

auth/auth_test.go
bbrks: It's not used but added to prevent this function accidentally being moved and used in non-test code.

gregns1: Makes sense! Good shout! 😄

PR #6687, opened 2024-02-15
CBG-3641: Ignore read-only fields if values unchanged in User API

gregns1:
CBG-3641

- Added helper method checkUserAPIReadOnlyFields to check read only fields, if they exist in the request ensure they are unchanged. If they are unchanged then nil the value on the request so UpdatePrinciple ignores it. If change detected then error. 
- Add test perfoming a GET on a user that will have read only fields populated, the using that request response in a PUT on the user, ensuring there is no error

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2302/

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -1348,9 +1348,15 @@ func (h *handler) updatePrincipal(name string, isUser bool) error {
 		}
 	}
 
-	// NB: other read-only properties are ignored but no error is returned for backwards-compatibility
-	if newInfo.JWTIssuer != nil || len(newInfo.JWTRoles) > 0 || len(newInfo.JWTChannels) > 0 {
-		return base.HTTPErrorf(http.StatusBadRequest, "Can't change read-only properties")
+	// Check read only fields in request against existing read only fields on the user, if the request attempts to
+	// change them, return error.
+	var unchanged bool
+	user, _ := h.db.Authenticator(h.ctx()).GetUser(internalUserName(name))
+	if user != nil {
+		newInfo, unchanged = checkUserAPIReadOnlyFields(newInfo, user)
+		if !unchanged {
+			return base.HTTPErrorf(http.StatusBadRequest, "Can't change read-only properties")
+		}
 	}
 
 	internalName := internalUserName(*newInfo.Name)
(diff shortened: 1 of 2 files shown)


rest/admin_api.go
```diff
@@ -1348,9 +1348,15 @@ func (h *handler) updatePrincipal(name string, isUser bool) error {
...
+	var unchanged bool
+	user, _ := h.db.Authenticator(h.ctx()).GetUser(internalUserName(name))
+	if user != nil {
+		newInfo, unchanged = checkUserAPIReadOnlyFields(newInfo, user)
+		if !unchanged {
+			return base.HTTPErrorf(http.StatusBadRequest, "Can't change read-only properties")
+		}
 	}
 
 	internalName := internalUserName(*newInfo.Name)
```
bbrks: Maybe we should move this up so we can use `internalName` instead of calling `internalUserName` twice?

It looks safe enough to do the name validation before the read-only check.

rest/admin_api.go
```diff
@@ -1770,3 +1776,29 @@ func databaseLoadErrorAsHTTPError(err error) error {
 	}
 	return base.HTTPErrorf(http.StatusInternalServerError, "couldn't load database: %v", err)
 }
+
+// checkUserAPIReadOnlyFields will return true if read only user fields are unchanged, else false. If a read only field
+// is unchanged it will nil the field so the update to principle ignores it.
```
bbrks, line 1781: > If a read only field is unchanged it will nil the field so the update to principle ignores it.

There is already logic inside `UpdatePrincipal` to skip if the non-nil values are the same, but there's no harm in being defensive here.

bbrks (changes requested):
Minor nit to reuse `internalName` but otherwise looks good

rest/admin_api.go
gregns1: My thinking here is due to this -> https://github.com/couchbase/sync_gateway/blob/9b782554ec47ef47bd2c19ed2f4965d662469a7e/db/users.go#L47 when I ran some test locally where a user is only granted channels, and thus only has issuer and channels defined on the PUT, it fails here. So thought it best to just nil them all in defense of that.

bbrks (approved):

PR #6691, opened 2024-02-16
CBG-3610: Fix regression on `PUT /_user/...` endpoint when sending `null` `admin_channels` / `admin_roles`

bbrks:
CBG-3610

- Test to repro issue (default collection passing on 3.0.x, failing on 3.1.x without this fix)
- Fix regression on `PUT /_user/...` endpoint when setting `null` `admin_channels` / `admin_roles`

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2310/
- [ ] `default collection,GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2311/

### DIFF
--- a/rest/user_api_test.go
+++ b/rest/user_api_test.go
@@ -1275,6 +1275,48 @@ func TestGetUserCollectionAccess(t *testing.T) {
 	RequireStatus(t, putResponse, 404)
 }
 
+// TestPutUserUnsetAdminChannels ensures that a PUT on the /_user/... endpoint with `null` admin_channels can be used to unset channels.
+// Repros CBG-3610, a regression introduced in 3.1.0
+func TestPutUserUnsetAdminChannels(t *testing.T) {
+	rt := NewRestTester(t, nil)
+	defer rt.Close()
+
+	defaultCollection := rt.GetSingleTestDatabaseCollection().IsDefaultCollection()
+
+	principalExplicitChannels := `"admin_channels": ["foo", "bar"]`
+	if !defaultCollection {
+		principalExplicitChannels = `"collection_access": {"` + rt.GetDbCollections()[0].ScopeName + `": {"` + rt.GetDbCollections()[0].Name + `": {"admin_channels": ["foo", "bar"]}}}`
+	}
+
+	// Create a user with some admin channels
+	response := rt.SendAdminRequest(http.MethodPut, "/db/_user/demo", `{"password": "password1", `+principalExplicitChannels+`}`)
+	RequireStatus(t, response, http.StatusCreated)
+
+	nilExplicitChannels := `"admin_channels": null`
+	if !defaultCollection {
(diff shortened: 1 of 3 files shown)


rest/user_api_test.go
```diff
@@ -1275,6 +1275,48 @@ func TestGetUserCollectionAccess(t *testing.T) {
 	RequireStatus(t, putResponse, 404)
 }
 
+// TestPutUserUnsetAdminChannels ensures that a PUT on the /_user/... endpoint with `null` admin_channels can be used to unset channels.
+// Repros CBG-3610, a regression introduced in 3.1.0
+func TestPutUserUnsetAdminChannels(t *testing.T) {
+	rt := NewRestTester(t, nil)
```
torcolvin, line 1281: I would tend to write tests with `NewRestTesterPersistentConfig()`

rest/user_api_test.go
```diff
@@ -1275,6 +1275,48 @@ func TestGetUserCollectionAccess(t *testing.T) {
...
+	defer rt.Close()
+
+	defaultCollection := rt.GetSingleTestDatabaseCollection().IsDefaultCollection()
+
+	principalExplicitChannels := `"admin_channels": ["foo", "bar"]`
```
torcolvin: Is this worth parameterizing so that `admin_roles` gets checked as well?

rest/user_api_test.go
bbrks: Is there much/any difference?

I'm seeing relatively few `NewRestTesterPersistentConfig` in comparison to `NewRestTester` so want to understand why I should not use it.

rest/user_api_test.go
bbrks: Think I can use `GetUserPayload` throughout this test, which handily already exists

rest/user_api_test.go
bbrks: Or rather, should `NewRestTester` use persistent config, and we have a non-persistent variant if we wish to test that?

torcolvin: I think we should change `NewRestTester` but I created a helper function to make sure that we use it in new tests. It is a bigger refactoring to make all `RestTester` use persistent config and there are two variants we care about that make it hard to refactor.

1. use persistent config, and have RestTester create the database for you. This is what non persistent config `NewRestTester` does now.
2. places that actually explicitly specify `RestTesterConfig{PersistentConfig: true}` and then manually create a databases. These can't be modified.

I created this helper function recently to make it easier for new tests to use persistent config without undergoing this refactoring. For this test it doesn't matter at all, because there's no modifications to database config but I thought I'd use this opportunity to try to start a new pattern. When we change the defaults for `RestTester` it might be easier change the code that we already know works.

We don't need to change this location as is.

torcolvin (approved):

PR #6700, opened 2024-02-21
CBG-3255  Replication protocol support for HLV - push replication

adamcfraser:
CBG-3255

Adds push replication support for HLV clients. Delta sync and attachments are not yet supported (pending CBG-3736, CBG-3797).

On proposeChanges, checks whether the incoming CV and parent version represent a new document, known version, valid update, or conflict. Uses the same handling as revTreeID (conflict if parent version isn’t the server’s current version), with the additional non-conflict case where the incoming CV and server CV share the same source and the incoming CV is a newer version.

For the incoming rev, detects conflict based on the incoming cv (based on the implicit hierarchy in an HLV, where cv > pv > mv).

Includes some test helpers to support writing tests with simplified versions (e.g. 1@abc) while still asserting for encoded source and version.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2317/

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -792,12 +792,18 @@ func (bh *blipHandler) handleProposeChanges(rq *blip.Message) error {
 
 	for i, change := range changeList {
 		docID := change[0].(string)
-		revID := change[1].(string)
+		rev := change[1].(string) // rev can represent a RevTree ID or HLV current version
 		parentRevID := ""
 		if len(change) > 2 {
 			parentRevID = change[2].(string)
 		}
-		status, currentRev := bh.collection.CheckProposedRev(bh.loggingCtx, docID, revID, parentRevID)
+		var status ProposedRevStatus
+		var currentRev string
+		if bh.activeCBMobileSubprotocol >= CBMobileReplicationV4 {
+			status, currentRev = bh.collection.CheckProposedVersion(bh.loggingCtx, docID, rev, parentRevID)
+		} else {
+			status, currentRev = bh.collection.CheckProposedRev(bh.loggingCtx, docID, rev, parentRevID)
+		}
 		if status == ProposedRev_OK_IsNew {
 			// Remember that the doc doesn't exist locally, in order to optimize the upcoming Put:
 			bh.collectionCtx.notePendingInsertion(docID)
@@ -937,18 +943,21 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 		}
 	}()
 
-	// addRevisionParams := newAddRevisionParams(rq)
+	if bh.activeCBMobileSubprotocol >= CBMobileReplicationV4 && bh.conflictResolver != nil {
+		return base.HTTPErrorf(http.StatusNotImplemented, "conflict resolver handling (ISGR) not yet implemented for v4 protocol")
+	}
+
 	revMessage := RevMessage{Message: rq}
 
 	// Doc metadata comes from the BLIP message metadata, not magic document properties:
 	docID, found := revMessage.ID()
-	revID, rfound := revMessage.Rev()
+	rev, rfound := revMessage.Rev()
 	if !found || !rfound {
-		return base.HTTPErrorf(http.StatusBadRequest, "Missing docID or revID")
+		return base.HTTPErrorf(http.StatusBadRequest, "Missing docID or rrevIDev")
 	}
 
 	if bh.readOnly {
-		return base.HTTPErrorf(http.StatusForbidden, "Replication context is read-only, docID: %s, revID:%s", docID, revID)
+		return base.HTTPErrorf(http.StatusForbidden, "Replication context is read-only, docID: %s, rev:%s", docID, rev)
 	}
 
 	base.DebugfCtx(bh.loggingCtx, base.KeySyncMsg, "#%d: Type:%s %s", bh.serialNumber, rq.Profile(), revMessage.String())
@@ -968,7 +977,7 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 			return err
 		}
 		if removed, ok := body[BodyRemoved].(bool); ok && removed {
-			base.InfofCtx(bh.loggingCtx, base.KeySync, "Purging doc %v - removed at rev %v", base.UD(docID), revID)
+			base.InfofCtx(bh.loggingCtx, base.KeySync, "Purging doc %v - removed at rev %v", base.UD(docID), rev)
 			if err := bh.collection.Purge(bh.loggingCtx, docID); err != nil {
 				return err
 			}
@@ -980,7 +989,7 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 				if err != nil {
 					base.WarnfCtx(bh.loggingCtx, "Unable to parse sequence %q from rev message: %v - not tracking for checkpointing", seqStr, err)
 				} else {
-					bh.collectionCtx.sgr2PullProcessedSeqCallback(&seq, IDAndRev{DocID: docID, RevID: revID})
+					bh.collectionCtx.sgr2PullProcessedSeqCallback(&seq, IDAndRev{DocID: docID, RevID: rev})
 				}
 			}
 			return nil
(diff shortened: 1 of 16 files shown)


db/blip_handler.go
```diff
@@ -937,18 +943,21 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
...
+
 	revMessage := RevMessage{Message: rq}
 
 	// Doc metadata comes from the BLIP message metadata, not magic document properties:
 	docID, found := revMessage.ID()
-	revID, rfound := revMessage.Rev()
+	rev, rfound := revMessage.Rev()
 	if !found || !rfound {
-		return base.HTTPErrorf(http.StatusBadRequest, "Missing docID or revID")
+		return base.HTTPErrorf(http.StatusBadRequest, "Missing docID or rrevIDev")
```
bbrks:
```suggestion
		return base.HTTPErrorf(http.StatusBadRequest, "Missing docID or rev")
```

db/blip_handler.go
```diff
@@ -980,17 +989,38 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
...
+			history = append(history, strings.Split(historyStr, ",")...)
+		}
+	} else {
+		versionVectorStr := rev
+		if historyStr := rq.Properties[RevMessageHistory]; historyStr != "" {
+			versionVectorStr += ";" + historyStr
+		}
+		incomingHLV, err = extractHLVFromBlipMessage(versionVectorStr)
+		if err != nil {
+			return base.HTTPErrorf(http.StatusUnprocessableEntity, "error extracting hlv from blip message")
```
bbrks, line 1020: Do we care about the contents of the error? (log or wrap)

db/blip_handler.go
```diff
@@ -1009,7 +1039,15 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 		//       while retrieving deltaSrcRevID.  Couchbase Lite replication guarantees client has access to deltaSrcRevID,
 		//       due to no-conflict write restriction, but we still need to enforce security here to prevent leaking data about previous
 		//       revisions to malicious actors (in the scenario where that user has write but not read access).
-		deltaSrcRev, err := bh.collection.GetRev(bh.loggingCtx, docID, deltaSrcRevID, false, nil)
+		var deltaSrcRev DocumentRevision
+		if bh.activeCBMobileSubprotocol >= CBMobileReplicationV4 {
+			cv := Version{}
+			cv.SourceID, cv.Value = incomingHLV.GetCurrentVersion()
+			// swap for GetCV once merged (CBG-3212)
+			deltaSrcRev, err = bh.collection.revisionCache.GetWithCV(bh.loggingCtx, docID, &cv, RevCacheOmitBody, RevCacheOmitDelta)
```
bbrks: This is in now, which should perform correct access checks (via `documentRevisionForRequest`)

```suggestion
			deltaSrcRev, err = bh.collection.GetCV(bh.loggingCtx, docID, &cv, RevCacheOmitBody)
```

db/blip_handler.go
```diff
@@ -1189,7 +1226,9 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 	// If the doc is a tombstone we want to allow conflicts when running SGR2
 	// bh.conflictResolver != nil represents an active SGR2 and BLIPClientTypeSGR2 represents a passive SGR2
 	forceAllowConflictingTombstone := newDoc.Deleted && (bh.conflictResolver != nil || bh.clientType == BLIPClientTypeSGR2)
-	if bh.conflictResolver != nil {
+	if bh.activeCBMobileSubprotocol >= CBMobileReplicationV4 {
+		_, _, _, err = bh.collection.PutExistingCurrentVersion(bh.loggingCtx, newDoc, incomingHLV, rawBucketDoc)
+	} else if bh.conflictResolver != nil {
```
bbrks, line 1231: I'm not sure this is correct if we want to support conflict resolution for ISGR on 4.0 - which granted is a future milestone

We'll need 4 versions of `PutExisting...` but it's feeling like this needs to be pushed down into a new function to do that work.

db/crud.go
```diff
@@ -1129,20 +1116,28 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
...
+			doc.HLV = &newHLV
+			addNewerVersionsErr := doc.HLV.AddNewerVersions(newDocHLV)
 			if addNewerVersionsErr != nil {
 				return nil, nil, false, nil, addNewerVersionsErr
 			}
 		} else {
-			incomingDecodedHLV := docHLV.ToDecodedHybridLogicalVector()
-			localDecodedHLV := doc.HLV.ToDecodedHybridLogicalVector()
-			if !incomingDecodedHLV.IsInConflict(localDecodedHLV) {
+			if (*doc.HLV).isDominating(newDocHLV) {
```
bbrks: nit: method call will auto-deref the pointer

```suggestion
			if doc.HLV.isDominating(newDocHLV) {
```

db/crud_test.go
```diff
@@ -1761,6 +1766,8 @@ func TestPutExistingCurrentVersion(t *testing.T) {
 	pv[bucketUUID] = docUpdateVersion
 	assert.True(t, reflect.DeepEqual(syncData.HLV.PreviousVersions, pv))
 	assert.Equal(t, "3-60b024c44c283b369116c2c2570e8088", syncData.CurrentRev)
+
+	// TODO: Test the case where server version dominates incoming
```
bbrks: Is this TODO dependent on anything in particular? (XDCR testing?)

db/database_test.go
```diff
@@ -307,6 +307,157 @@ func TestDatabase(t *testing.T) {
...
+			require.NoError(t, err)
+			err = hlv.AddVersion(Version{"cluster2", hlv.CurrentVersionCAS + 4})
+			require.NoError(t, err)
+			// TODO: Get full BLIP formatted HLV
+			hlvString := hlv.GetCurrentVersionString()
+			status, _ := collection.CheckProposedVersion(ctx, "doc1", hlvString)
+			assert.Equal(t, ProposedRev_Conflict, status)
+		})
+	*/
+
```
bbrks: This is now covered by above test table.

```suggestion
```

db/revision_cache_interface.go
```diff
@@ -331,8 +331,9 @@ func revCacheLoaderForDocument(ctx context.Context, backingStore RevisionCacheBa
 // nolint:staticcheck
 func revCacheLoaderForDocumentCV(ctx context.Context, backingStore RevisionCacheBackingStore, doc *Document, cv Version) (bodyBytes []byte, body Body, history Revisions, channels base.Set, removed bool, attachments AttachmentsMeta, deleted bool, expiry *time.Time, revid string, hlv *HybridLogicalVector, err error) {
 	if bodyBytes, body, attachments, err = backingStore.getCurrentVersion(ctx, doc); err != nil {
-		// TODO: pending CBG-3213 support of channel removal for CV
-		// we need implementation of IsChannelRemoval for CV here.
+		// need implementation of IsChannelRemoval for CV
+		// TODO: CBG-3213 - pending support of channel removal for CV
+		base.ErrorfCtx(ctx, "pending CBG-3213 support of channel removal for CV: %v", err)
```
bbrks: I think we may need a new ticket for this.

bbrks (changes requested):

db/blip_handler.go
adamcfraser: Given that PutExistingRev is just a wrapper for PutExistingRevWithConflictResolution (that passes a nil conflict resolver), I think we could simplify this in future by just calling PutExistingRevWithConflictResolution directly here for the non-CV case.  I was going to put a TODO, but I think this will be obvious when someone comes along and adds conflict resolution for CBMobileReplicationV4 in future.

db/revision_cache_interface.go
adamcfraser: Filed https://issues.couchbase.com/browse/CBG-3814 for this.  There's some interaction between this and removing channel history from the revtree (CBG-3203), I think it makes sense to wait until that work is done.

db/crud_test.go
adamcfraser: Updated the test to cover this case.

db/blip_handler.go
adamcfraser: Good catch - added logging.

bbrks (approved):

PR #6702, opened 2024-02-24
CBG-3797 Attachment handling for HLV push replication

adamcfraser:
CBG-3797

HLV clients don't consider revpos, and evaluate whether they need to request an attachment based on the existing set of attachments on the document.

SGW still needs to persist revpos into _attachments to support revtree clients.  For new attachments added by HLV client, revpos is set to the generation of SGW's computed revTreeID for the incoming revision.

- [ ] After CBG-3255 is merged, rebase on beryllium before merging

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2320/

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -799,7 +799,7 @@ func (bh *blipHandler) handleProposeChanges(rq *blip.Message) error {
 		}
 		var status ProposedRevStatus
 		var currentRev string
-		if bh.activeCBMobileSubprotocol >= CBMobileReplicationV4 {
+		if bh.useHLV() {
 			status, currentRev = bh.collection.CheckProposedVersion(bh.loggingCtx, docID, rev, parentRevID)
 		} else {
 			status, currentRev = bh.collection.CheckProposedRev(bh.loggingCtx, docID, rev, parentRevID)
@@ -943,7 +943,7 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 		}
 	}()
 
-	if bh.activeCBMobileSubprotocol >= CBMobileReplicationV4 && bh.conflictResolver != nil {
+	if bh.useHLV() && bh.conflictResolver != nil {
 		return base.HTTPErrorf(http.StatusNotImplemented, "conflict resolver handling (ISGR) not yet implemented for v4 protocol")
 	}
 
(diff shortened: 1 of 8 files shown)


db/blip_handler.go
```diff
@@ -1568,3 +1572,7 @@ func allowedAttachmentKey(docID, digest string, activeCBMobileSubprotocol CBMobi
 func (bh *blipHandler) logEndpointEntry(profile, endpoint string) {
 	base.InfofCtx(bh.loggingCtx, base.KeySyncMsg, "#%d: Type:%s %s", bh.serialNumber, profile, endpoint)
 }
+
+func (bh *blipHandler) useHLV() bool {
+	return bh.activeCBMobileSubprotocol >= CBMobileReplicationV4
+}
```
bbrks, line 1578: Thanks! Already got tired of reading the subprotocol check! 👍

rest/blip_api_attachment_test.go
```diff
@@ -312,52 +312,60 @@ func TestBlipPushPullNewAttachmentCommonAncestor(t *testing.T) {
...
 
 		// Wait for the documents to be replicated at SG
 		_, ok := btc.pushReplication.WaitForMessage(2)
 		assert.True(t, ok)
 
-		resp := btc.rt.SendAdminRequest(http.MethodGet, "/{{.keyspace}}/"+docID+"?rev="+revId, "")
-		assert.Equal(t, http.StatusOK, resp.Code)
+		collection := rt.GetSingleTestDatabaseCollection()
+		doc, _, err := collection.GetDocWithXattr(ctx, docID, db.DocUnmarshalNoHistory)
+		require.NoError(t, err)
```
bbrks: Do we need to limit this test to xattr only? Or should we switch this (and other usages) to `collection.GetDocument()` ?

```suggestion
		doc, err := collection.GetDocument(ctx, docID, db.DocUnmarshalNoHistory)
		require.NoError(t, err)
```

bbrks (changes requested):

rest/blip_api_attachment_test.go
adamcfraser: Good catch. GetDocument is a reasonable change for these tests.  For the next milestone we might want to disable VersionVectorSubtest inside BlipTestClientRunner if xattrs=false (as that's not a supported config when we start writing the HLV to its own xattr), but we don't need that yet.

bbrks (dismissed):
LGTM pending rebase

adamcfraser:
Rebased on beryllium.

bbrks (approved):

PR #6703, opened 2024-02-26
CBG-3791: change getAuthScopeHandleCreateDB method to not expand environment variables

gregns1:
CBG-3791

- Removed the use of DecodeAndSanitiseConfig as it expands environment variables 
- As we only need bucket name from this function just kept much the functionality the same but with no expansion of environment variables.  
- Kept the use of Decode over Unmarshal as I feel it's more memory efficient and given we unmarshal the config later in readSanitizeDbConfigJSON I didn't want this done twice. 
- Evaluated the use of readSanitizeDbConfigJSON in place of what I have done in this PR but found it awkward putting the bytes back into a reader for use later. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2327/

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -171,7 +171,7 @@ func getAuthScopeHandleCreateDB(ctx context.Context, h *handler) (bucketName str
 	// grab a copy of the request body and restore body buffer for later handlers
 	bodyJSON, readErr := h.readBody()
 	if readErr != nil {
-		return "", base.HTTPErrorf(http.StatusInternalServerError, "Unable to read body: %v", err)
+		return "", base.HTTPErrorf(http.StatusInternalServerError, "Unable to read body: %v", readErr)
 	}
 	// mark the body as already read to avoid double-counting bytes for stats once it gets read again
 	h.requestBody.bodyRead = true
@@ -187,6 +187,16 @@ func getAuthScopeHandleCreateDB(ctx context.Context, h *handler) (bucketName str
 	if err != nil {
 		return "", err
 	}
+
+	// Expand environment variables if needed
+	if base.BoolDefault(h.server.Config.Unsupported.AllowDbConfigEnvVars, true) {
+		b, err = expandEnv(h.ctx(), b)
+		if err != nil {
+			return "", err
+		}
+	}
+	b = base.ConvertBackQuotedStrings(b)
+
 	d := base.JSONDecoder(bytes.NewBuffer(b))
 	err = d.Decode(&dbConfigBody)
 	if err != nil {
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -169,8 +169,8 @@ func (h *handler) handleCreateDB() error {
 func getAuthScopeHandleCreateDB(ctx context.Context, h *handler) (bucketName string, err error) {
 
 	// grab a copy of the request body and restore body buffer for later handlers
-	bodyJSON, err := h.readBody()
-	if err != nil {
+	bodyJSON, readErr := h.readBody()
+	if readErr != nil {
 		return "", base.HTTPErrorf(http.StatusInternalServerError, "Unable to read body: %v", err)
 	}
 	// mark the body as already read to avoid double-counting bytes for stats once it gets read again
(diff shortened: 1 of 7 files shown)


rest/admin_api.go
```diff
@@ -169,8 +169,8 @@ func (h *handler) handleCreateDB() error {
 func getAuthScopeHandleCreateDB(ctx context.Context, h *handler) (bucketName string, err error) {
 
 	// grab a copy of the request body and restore body buffer for later handlers
-	bodyJSON, err := h.readBody()
-	if err != nil {
+	bodyJSON, readErr := h.readBody()
+	if readErr != nil {
 		return "", base.HTTPErrorf(http.StatusInternalServerError, "Unable to read body: %v", err)
```
```suggestion
		return "", base.HTTPErrorf(http.StatusInternalServerError, "Unable to read body: %v", readErr)
```

rest/admin_api.go
```diff
@@ -181,7 +181,14 @@ func getAuthScopeHandleCreateDB(ctx context.Context, h *handler) (bucketName str
 		Bucket string `json:"bucket"`
 	}
 	reader := bytes.NewReader(bodyJSON)
-	err = DecodeAndSanitiseConfig(ctx, reader, &dbConfigBody, false)
+
+	// read and decode json to pull bucket name from the config
```
bbrks: I think we should evaluate env vars based on the `AllowDbConfigEnvVars` setting.

It's still possible that customers have a `$BUCKET` env var they're using when creating databases. We just want to apply those settings consistently.

torcolvin: I think you can call `readSantizedJson` here.

The thing that might happen is that you need to call `ConvertBackQuotedStrings` to make it valid json. The test should include a config with backquotes

bbrks (changes requested):

rest/adminapitest/admin_api_test.go
```diff
@@ -4015,3 +4015,61 @@ func TestDatabaseCreationErrorCode(t *testing.T) {
...
+				"collections": {
+					"%s": {
+        					"sync": "%s"
+					}
+				}
+			}
+		  },
+		  "num_index_replicas": 0,
+		  "enable_shared_bucket_access": true,
+		  "use_views": false}`,
```
torcolvin: this might fail if SG_TEST_USE_XATTRS=false is set, can you check?

gregns1: Just checked and no failure.

rest/admin_api.go
```diff
@@ -181,7 +181,24 @@ func getAuthScopeHandleCreateDB(ctx context.Context, h *handler) (bucketName str
...
+	b, err := io.ReadAll(reader)
+	if err != nil {
+		return "", err
+	}
+
+	// Expand environment variables if needed
+	if base.BoolDefault(h.server.Config.Unsupported.AllowDbConfigEnvVars, true) {
+		b, err = expandEnv(h.ctx(), b)
```
bbrks: We're doing bytes->reader->bytes for no reason here with NewReader->ReadAll 

We should be able to pass `bodyJSON` straight into `expandEnv` and `ConvertBackQuotedStrings`, since they don't modify the slice, they return a modified copy of it.

It does feels like we need to push those two pieces (`expandEnv`/`ConvertBackQuotedStrings`) into a shared function called by this, `DecodeAndSanitiseConfig`, and `readSanitizeJSON`.

We can pass a bool in to that function control env var behaviour and just have it take bytes and return bytes.

bbrks (changes requested):

rest/adminapitest/admin_api_test.go
```diff
@@ -4015,3 +4015,112 @@ func TestDatabaseCreationErrorCode(t *testing.T) {
 		rest.RequireStatus(t, resp, http.StatusPreconditionFailed)
 	}
 }
+
+// TestDatabaseCreationWithEnvVariable:
+//   - Create rest tester that enables admin auth and disallows db config env vars
+//   - Create CBS user to authenticate with over admin port to force auth scope callback call
+//   - Create db with sync function that calls env variable
+//   - Assert that db is created
+func TestDatabaseCreationWithEnvVariable(t *testing.T) {
```
bbrks, line 4024: This test doesn't work for `SG_TEST_USE_GSI=false` or `SG_TEST_USE_DEFAULT_COLLECTION=true`

Will need to build the db config to accommodate those modes.

Some existing functions to consider using or adapting:
- `dbConfigForTestBucket`
- `makeDbConfig`

Should be able to include string values like `"${VAR}"` for sync function before marshalling it to pass into the request.

bbrks (changes requested):
Changes look fine. We need to make sure the new tests work with the different combinations of env vars we can use though.

rest/adminapitest/admin_api_test.go
gregns1: Oh wow didn't even think about that, on it fixing it

bbrks (approved):

PR #6704, opened 2024-02-28
CBG-3702: setup 7.6.1 buckets for integration tests

gregns1:
CBG-3702

- When creating a bucket object, query the bucket setting for presence of enableCrossClusterVersioning true, if present set flag on the bucket object that bucket supportsHLV 
- Change to `IsSupported` for `BucketStoreFeatureMobileXDCR` which will just return the flag value on the bucket object 
- When creating test bucket pool check that the server version is enterprise and > 7.6.1. This takes some string manipulation but is test only code so should be fine. `getClusterVersion` doesn't return enough version points we need. 
- If we don't skip mobile XDCR bucket setup we will query the bucket settings to set enableCrossClusterVersioning to true
- Changes to xdcr code to allow us to set the mobile flag in replication configuration. 
- New test TestMobileXDCRNoSyncDataCopied that setups mobile XDCR replication and asserts the correct documents are replicated, in addition checks the version vector is populated on destination bucket. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2329/


base/collection.go
```diff
@@ -246,11 +253,35 @@ func (b *GocbV2Bucket) IsSupported(feature sgbucket.BucketStoreFeature) bool {
...
 		return false
 	}
 }
 
+// queryHLVBucketSetting sends request to server to check for enableCrossClusterVersioning bucket setting
+func (b *GocbV2Bucket) queryHLVBucketSetting(ctx context.Context) error {
+	url := fmt.Sprintf("/pools/default/buckets/%s", b.GetName())
+	output, statusCode, err := mgmtRequest(ctx, b, http.MethodGet, url, nil)
+	if err != nil || statusCode != http.StatusOK {
+		return fmt.Errorf("error executing query for mobile XDCR bucket setting, status code: %d error: %v", statusCode, err)
```
torcolvin: can you print the output here in the failure cases?

base/collection.go
```diff
@@ -246,11 +253,35 @@ func (b *GocbV2Bucket) IsSupported(feature sgbucket.BucketStoreFeature) bool {
...
+		SupportsHLV *bool `json:"enableCrossClusterVersioning,omitempty"`
+	}
+	var bucketSettings bucket
+	err = JSONUnmarshal(output, &bucketSettings)
+	if err != nil {
+		return err
+	}
+	if bucketSettings.SupportsHLV != nil {
+		b.supportsHLV = *bucketSettings.SupportsHLV
+	}
```
torcolvin, line 282: might be worth adding a comment here:

In Server < 7.6.1 (not sure if this is actually Server 7.6.1) this field will not be present, but if it is not configured on the bucket, it will return false.

base/main_test_bucket_pool.go
```diff
@@ -433,6 +445,44 @@ func (tbp *TestBucketPool) emptyPreparedStatements(ctx context.Context, b Bucket
...
+	store, ok := AsCouchbaseBucketStore(bucket)
+	if !ok {
+		tbp.Fatalf(ctx, "unable to get server management endpoints. Underlying bucket type was not GoCBBucket")
+	}
+
+	posts := url.Values{}
+	posts.Add("enableCrossClusterVersioning", "true")
+
+	url := fmt.Sprintf("/pools/default/buckets/%s", store.GetName())
+	_, statusCode, err := mgmtRequest(ctx, store, http.MethodPost, url, strings.NewReader(posts.Encode()))
```
torcolvin: can we print output for this if this fails?

base/main_test_bucket_pool.go
```diff
@@ -433,6 +445,44 @@ func (tbp *TestBucketPool) emptyPreparedStatements(ctx context.Context, b Bucket
...
+	if err != nil {
+		tbp.Fatalf(ctx, "request to mobile XDCR bucket setting failed, error: %v", err)
+	}
+	if statusCode != http.StatusOK {
+		tbp.Fatalf(ctx, "request to mobile XDCR bucket setting failed with status code, %d", statusCode)
+	}
+}
+
+// mgmtRequest sends REST API request to server
+func mgmtRequest(ctx context.Context, bucket CouchbaseBucketStore, method, url string, body io.Reader) ([]byte, int, error) {
```
torcolvin: This is duplicated in xdcr package - I am tempted to make `MgmtRequest` API use this because it looks like all the callers are immediately reading output. This would mean that we don't have to worry about closing the body as the callers.

I can't think of a time right now where the management API would return a request that we'd want to handle streaming.

base/main_test_cluster.go
```diff
@@ -203,3 +214,48 @@ func (c *tbpClusterV2) supportsMobileRBAC() (bool, error) {
 	}
 	return major >= 7 && minor >= 1, nil
 }
+
+// mobileXDCRCompatible checks if a cluster is mobile XDCR compatible, a cluster must be enterprise edition AND > 7.6.1
+func (c *tbpClusterV2) mobileXDCRCompatible() (bool, error) {
```
torcolvin: Can we move this logic to the constructor for `tbpClusterV2` and just set a variable here, so we don't call this each time we open a bucket? I think that would make the code logic easier.

base/main_test_bucket_pool_config.go
```diff
@@ -81,6 +81,25 @@ func TestsRequireMobileRBAC(t *testing.T) {
 	}
 }
 
+// canUseMobileXDCR checks if cluster supports mobile XDCR
+func (tbp *TestBucketPool) canUseMobileXDCR(ctx context.Context) (bool, error) {
+	var clusterSupport bool
+	if tbp.cluster != nil {
```
torcolvin: When is this nil?

base/main_test_cluster.go
```diff
@@ -203,3 +214,48 @@ func (c *tbpClusterV2) supportsMobileRBAC() (bool, error) {
...
+	if !enterprise {
+		return false, nil
+	}
+
+	metadata, err := c.cluster.Internal().GetNodesMetadata(&gocb.GetNodesMetadataOptions{})
+	if err != nil {
+		return false, err
+	}
+
+	// take server version
```
torcolvin: I think it might be helpful here to write the possible strings for this

base/main_test_cluster.go
```diff
@@ -203,3 +214,48 @@ func (c *tbpClusterV2) supportsMobileRBAC() (bool, error) {
...
+	}
+	minor, err := strconv.Atoi(strList[1])
+	if err != nil {
+		return false, err
+	}
+	patch, err := strconv.Atoi(strList[2])
+	if err != nil {
+		return false, err
+	}
+	return major >= 7 && minor >= 6 && patch >= 1, nil
```
torcolvin: This would fail for server 8.0.0

go.mod
```diff
@@ -9,8 +9,8 @@ require (
 	github.com/couchbase/cbgt v1.3.9
 	github.com/couchbase/clog v0.1.0
 	github.com/couchbase/go-blip v0.0.0-20231212195435-3490e96d30e3
-	github.com/couchbase/gocb/v2 v2.7.1
-	github.com/couchbase/gocbcore/v10 v10.3.1
+	github.com/couchbase/gocb/v2 v2.7.2
```
torcolvin: Do we need to update gocb in this PR?

xdcr/cbs_xdcr.go
```diff
@@ -27,11 +28,17 @@ const (
 	totalDocsWrittenStat      = "xdcr_docs_written_total"
 )
 
+const (
+	MobileOff    = "Off"
+	MobileActive = "Active"
+)
```
torcolvin, line 40: If you use

```
type serverMobileSetting uint8

const (
      serverMobileOff = iota
      serverMobileOn
)
```

you can avoid some type checks

xdcr/cbs_xdcr.go
```diff
@@ -147,8 +158,13 @@ func (x *CouchbaseServerXDCR) Start(ctx context.Context) error {
...
 	body.Add("replicationType", "continuous")
-	// filters all sync docs, except binary docs (attachments)
-	body.Add("filterExpression", fmt.Sprintf("NOT REGEXP_CONTAINS(META().id, \"^%s\") OR REGEXP_CONTAINS(META().id, \"^%s\")", base.SyncDocPrefix, base.Att2Prefix))
+	// if mobile setting is active, set the setting on thw replication
+	if x.MobileSetting != MobileOff {
+		body.Add("mobile", x.MobileSetting)
+	} else {
+		// filters all sync docs, except binary docs (attachments)
+		body.Add("filterExpression", fmt.Sprintf("NOT REGEXP_CONTAINS(META().id, \"^%s\") OR REGEXP_CONTAINS(META().id, \"^%s\")", base.SyncDocPrefix, base.Att2Prefix))
+	}
```
torcolvin: This was a hack when I couldn't use mobile, I think xdcr without {{mobile}} is not something we need to test, or we might want a custom filter expression if we were testing unidirectional XDCR for backup/restore purposes.

xdcr/cbs_xdcr_test.go
```diff
@@ -74,3 +74,88 @@ func TestCBSXDCR(t *testing.T) {
...
 
 }
+
+// TestMobileXDCRNoSyncDataCopied:
+//   - Setup XDCR with mobile flag set to active
+//   - Put doc, attachment doc and sync doc
+//   - Assert that doc and attachment doc are replicated, sync doc is not
+//   - Assert that the version vector is written on destination bucket for each replicated doc
+func TestMobileXDCRNoSyncDataCopied(t *testing.T) {
+	if base.UnitTestUrlIsWalrus() {
```
torcolvin, line 27: Let's write enable in https://issues.couchbase.com/browse/CBG-3703

xdcr/cbs_xdcr.go
gregns1: Will refactor to allow a custom filter expression to be supplied instead.

base/main_test_bucket_pool_config.go
gregns1: Don't think it ever should be but did this to be panic safe

base/collection.go
```diff
@@ -468,19 +500,19 @@ func (b *GocbV2Bucket) BucketName() string {
 	return b.GetName()
 }
 
-func (b *GocbV2Bucket) MgmtRequest(ctx context.Context, method, uri, contentType string, body io.Reader) (*http.Response, error) {
+func (b *GocbV2Bucket) MgmtRequest(ctx context.Context, method, uri, contentType string, body io.Reader) ([]byte, int, error) {
```
torcolvin, line 505:
```suggestion
MgmtRequest makes a request to the http couchbase management api. The uri is the non host part of the URL, such as /pools/default/buckets. This function will read the entire contents of the response and return the output bytes, the status code, and an error.
func (b *GocbV2Bucket) MgmtRequest(ctx context.Context, method, uri, contentType string, body io.Reader) ([]byte, int, error) {
```

base/main_test_bucket_pool.go
```diff
@@ -433,6 +444,30 @@ func (tbp *TestBucketPool) emptyPreparedStatements(ctx context.Context, b Bucket
 	}
 }
 
+// setXDCRBucketSetting sets the bucket setting "enableCrossClusterVersioning" for mobile XDCR
+func (tbp *TestBucketPool) setXDCRBucketSetting(ctx context.Context, bucket Bucket) {
+	if tbp.skipMobileXDCR {
+		tbp.Logf(ctx, "skipping mobile XDCR bucket setting")
```
torcolvin:
```suggestion
		tbp.Logf(ctx, "Not setting crossClusterVersioningEnabled")
```

base/main_test_bucket_pool.go
```diff
@@ -433,6 +444,30 @@ func (tbp *TestBucketPool) emptyPreparedStatements(ctx context.Context, b Bucket
...
+	}
+
```
torcolvin, line 451:
```suggestion
		tbp.Logf(ctx, "Setting crossClusterVersioningEnabled=true")
```

base/main_test_bucket_pool.go
```diff
@@ -503,6 +539,8 @@ func (tbp *TestBucketPool) createTestBuckets(numBuckets, bucketQuotaMB int, buck
 
 			tbp.emptyPreparedStatements(ctx, bucket)
 
+			tbp.setXDCRBucketSetting(ctx, bucket)
```
torcolvin: I think that I'd be inclined to put the logic here to test if we should set `enableCrossClusterVersioningEnabled`.

If this is rosmar, we are going to skip this value and I don't think this is worth logging.

xdcr/cbs_xdcr_test.go
```diff
@@ -32,7 +33,78 @@ func TestCBSXDCR(t *testing.T) {
 	toBucket, err := base.AsGocbV2Bucket(bucket2)
 	require.NoError(t, err)
 
-	xdcr, err := NewCouchbaseServerXDCR(ctx, fromBucket, toBucket)
+	xdcr, err := NewCouchbaseServerXDCR(ctx, fromBucket, toBucket, serverMobileOff)
+	require.NoError(t, err)
+	// set filter for sync docs as mobile xdcr is not on
+	xdcr.filter = fmt.Sprintf("NOT REGEXP_CONTAINS(META().id, \"^%s\") OR REGEXP_CONTAINS(META().id, \"^%s\")", base.SyncDocPrefix, base.Att2Prefix)
```
torcolvin: should this be based on what version of couchbase server we are running? I don't think we should test this feature with a filter expression at all.

I would test this based on versions of server whether the _sync docs are going to get copied.

If we set the filter option, we aren't going to see the results of {{-mobile}} setting.

gregns1: I think we either remove the test altogether or leave as is until we start running integration tests against 7.6.1. The idea of me doing this was so we can have a test that will keep testing the XDCR functionality until we can properly run tests against 7.6.1. I'm filing a ticket to get 7.6.1 in Jenkins so can link this test to that so once we can run tests in Jenkins we can remove it.

torcolvin: I think just change this test to make sure XDCR is working differently in different server version? We should be able to expose `GTestBucketPool.supportsMobileXDCR` ? 

Maybe even have it assert on the existence of sync docs with < 7.6.1, and not otherwise.

I see the xdcr package as mostly general still which we could use to test a unidirectional DR XDCR issue.

torcolvin (approved):

PR #6711, opened 2024-03-04
CBG-3278: move serverless stats behind serverless flag

gregns1:
CBG-3278

- Moved stats added for serverless into its own struct on stats, will be initiated on db stats if in serverless mode
- Went through tests that asserted on these stats to change them to serverless tests, this meant making them persistent config too
- Some test were testing the stats on ISGR, so created a separate helper to make SGR peers in serverless and persistent config mode, chose this over passing logic into the existing helper method to make it tidier. 
- Added helper `NewRestTesterPersistentConfigServerless` to create serverless rest tester
- NOTE: All tests that assert on these stats will now be skipped for rosmar due to needing bucket specific creds for persistent config 

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2330/

### DIFF
--- a/base/stats.go
+++ b/base/stats.go
@@ -1112,12 +1112,26 @@ type QueryStat struct {
 	QueryTime       *SgwIntStat
 }
 
-func (s *SgwStats) NewDBStats(name string, deltaSyncEnabled bool, importEnabled bool, viewsEnabled bool, serverlessEnabled bool, queryNames []string, collections []string) (*DbStats, error) {
+// DbStatsOptions contains the configuration options for initialising db stats
+type DbStatsOptions struct {
+	DeltaSyncEnabled  bool
+	ImportEnabled     bool
+	ViewsEnabled      bool
+	ServerlessEnabled bool
+	QueryNames        []string
+	Collections       []string
+}
+
+func (s *SgwStats) NewDBStats(name string, dbStatsOpts *DbStatsOptions) (*DbStats, error) {
 	s.dbStatsMapMutex.Lock()
 	defer s.dbStatsMapMutex.Unlock()
 	dbStats := &DbStats{
 		dbName:            name,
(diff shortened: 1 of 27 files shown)


base/stats.go
```diff
@@ -1106,7 +1112,7 @@ type QueryStat struct {
 	QueryTime       *SgwIntStat
 }
 
-func (s *SgwStats) NewDBStats(name string, deltaSyncEnabled bool, importEnabled bool, viewsEnabled bool, queryNames []string, collections []string) (*DbStats, error) {
+func (s *SgwStats) NewDBStats(name string, deltaSyncEnabled bool, importEnabled bool, viewsEnabled bool, serverlessEnabled bool, queryNames []string, collections []string) (*DbStats, error) {
```
torcolvin: Can we switch the parameters of this to an options struct `DBStatsOptions` so it is easier to read the intent behind the callers and we don't have to modify it everywhere the next time we call it?

rest/utilities_testing.go
```diff
@@ -121,6 +121,18 @@ func NewRestTesterPersistentConfig(tb testing.TB) *RestTester {
 	return rt
 }
 
+func NewRestTesterPersistentConfigServerless(tb *testing.T) *RestTester {
+	RequireBucketSpecificCredentials(tb)
+	config := &RestTesterConfig{
+		PersistentConfig: true,
+		Serverless:       true,
+		SyncFn:           channels.DocChannelsSyncFunction,
```
torcolvin: I would not be inclined to set a non default SyncFn here if we don't need to.

I would be inclined to document this as:

```
NewRestTesterPersistentConfigServerless is a convenience function to set up RestTester in persistent config with serverless mode.```

torcolvin (approved):

PR #6715, opened 2024-03-05
CBG-3721: [Diagnostic API] Import filter and sync function dry run endpoints

mohammed-madi:
CBG-3721

Describe your PR here...
- Added 2 endpoints to the diagnostic API, sync function and import filter dry runs
- Added API docs
- Added TestGetDocDryRuns

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2348/


rest/routing.go
```diff
@@ -373,6 +373,9 @@ func createDiagnosticRouter(sc *ServerContext) *mux.Router {
 	keyspace := r.PathPrefix("/{keyspace:" + dbRegex + "}/").Subrouter()
 	keyspace.StrictSlash(true)
 	keyspace.Handle("/{docid:"+docRegex+"}/_all_channels", makeHandler(sc, adminPrivs, []Permission{PermReadAppData}, nil, (*handler).handleGetDocChannels)).Methods("GET")
+	keyspace.Handle("/sync/{docid:"+docRegex+"}", makeHandler(sc, adminPrivs, []Permission{PermReadAppData}, nil, (*handler).handleSyncFnDryRun)).Methods("GET")
```
adamcfraser: Our standard approach is use leading underscores for the first entry of keyspaced paths to avoid routing collisions.  I realize there isn't a generic `{docid:"+docRegex+"}` route on the diagnostic router (yet), but I think it would be good to stick to that approach for consistency and future-proofing.

db/crud.go
```diff
@@ -1165,6 +1165,26 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
 
 }
 
+func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, body Body, docID string) (*uint32, string, base.Set, channels.AccessMap, channels.AccessMap, error) {
+	doc := &Document{
+		ID:    docID,
+		_body: body,
+	}
+	// Make sure the user has permission to modify the document before confirming doc existence
```
adamcfraser: This comment isn't accurate in this usage.

rest/diagnostic_doc_api.go
```diff
@@ -12,9 +12,23 @@ package rest
...
 	"github.com/couchbase/sync_gateway/auth"
+	"github.com/couchbase/sync_gateway/base"
+	"github.com/couchbase/sync_gateway/channels"
 	"github.com/couchbase/sync_gateway/db"
 )
 
+type SyncFnDryRun struct {
+	Channels  base.Set           `json:"Channels"`
+	Access    channels.AccessMap `json:"Access"`
+	Roles     channels.AccessMap `json:"Roles"`
```
adamcfraser: Sync Function execution also returns 'expiry' - I think that would be useful information for the dry run.

rest/diagnostic_doc_api.go
```diff
@@ -39,3 +53,45 @@ func (h *handler) handleGetDocChannels() error {
 	h.writeJSON(resp)
 	return nil
 }
+
+// HTTP handler for a GET of a document
```
adamcfraser: Comment needs to be updated.

db/crud.go
```diff
@@ -1165,6 +1165,26 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
...
+	mutableBody, metaMap, newRevID, err := db.prepareSyncFn(doc, doc)
```
adamcfraser: Since we're getting 'body' and docID from the incoming request, I expected us to use the existing version of the document (from the bucket) as oldDoc, when it exists.   Otherwise I don't think there's any reason for callers to provide docID?

rest/routing.go
```diff
@@ -373,6 +373,9 @@ func createDiagnosticRouter(sc *ServerContext) *mux.Router {
...
+	keyspace.Handle("/import_filter/{docid:"+docRegex+"}", makeHandler(sc, adminPrivs, []Permission{PermReadAppData}, nil, (*handler).handleImportFilterDryRun)).Methods("GET")
```
adamcfraser: It doesn't look like we actually use the docid that's included on this path.  Do we need it?

rest/diagnostic_doc_api.go
```diff
@@ -39,3 +53,45 @@ func (h *handler) handleGetDocChannels() error {
...
+		access,
+		roles,
+		errorMsg,
+	}
+
+	h.writeJSON(resp)
+	return nil
+}
+
+// HTTP handler for a GET of a document
```
adamcfraser: Comment needs to be updated.

rest/diagnostic_doc_api.go
```diff
@@ -39,3 +53,45 @@ func (h *handler) handleGetDocChannels() error {
...
 }
+
+// HTTP handler for a GET of a document
+func (h *handler) handleSyncFnDryRun() error {
+	docid := h.PathVar("docid")
+	body, err := h.readDocument()
+	if err != nil {
+		return err
+	}
+	_, _, channelSet, access, roles, err := h.collection.SyncFnDryrun(h.ctx(), body, docid)
```
adamcfraser: Since this appears to be the only caller of SyncFnDryRun, should unused omitted fields just be removed from the method?  (I think we actually want to use expiry, but SyncFnDryrun doesn't need to return oldBody.

db/crud.go
```diff
@@ -1165,6 +1165,26 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
...
+	if err != nil {
+		base.InfofCtx(ctx, base.KeyCRUD, "Failed to prepare to run sync function: %v", err)
```
adamcfraser: Can we change the log key here to a new diagnostic endpoint log key?  It will help us differentiate between diagnostic and non-diagnostic logging (particularly in cases like this where the log message is just cut-and-paste from non-diagnostic code).

db/crud.go
```diff
@@ -1165,6 +1165,26 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
...
+		return nil, "", nil, nil, nil, err
+	}
+
+	syncExpiry, oldBody, channelSet, access, roles, err := db.runSyncFn(ctx, doc, mutableBody, metaMap, newRevID)
```
adamcfraser: runSyncFn increments several stats (in getChannelsAndAccess and checkDocChannelsAndGrantsLimits). I suspect we don't want those stats to be updated for dry runs.  This might require a 'dry run' flag to be added to the underlying code - I think that's more straightforward than messing around with the database context used for diagnostic operations.

docs/api/paths/diagnostic/keyspace-sync.yaml
```diff
@@ -0,0 +1,61 @@
...
+# software will be governed by the Apache License, Version 2.0, included in
+# the file licenses/APL2.txt.
+parameters:
+  - $ref: ../../components/parameters.yaml#/keyspace
+get:
+  summary: Run a doc body through the sync function and return sync data.
+  description: |-
+    Run a document body through the sync function and return document sync data.
+    * Sync Gateway Application Read Only
+  requestBody:
```
adamcfraser, line 15: I think this should include a 'parameters' section to define the supported query string parameters for this endpoint (docid, rev).

rest/routing.go
```diff
@@ -373,6 +373,9 @@ func createDiagnosticRouter(sc *ServerContext) *mux.Router {
 	keyspace := r.PathPrefix("/{keyspace:" + dbRegex + "}/").Subrouter()
 	keyspace.StrictSlash(true)
 	keyspace.Handle("/{docid:"+docRegex+"}/_all_channels", makeHandler(sc, adminPrivs, []Permission{PermReadAppData}, nil, (*handler).handleGetDocChannels)).Methods("GET")
+	keyspace.Handle("/_sync", makeHandler(sc, adminPrivs, []Permission{PermReadAppData}, nil, (*handler).handleSyncFnDryRun)).Methods("GET")
+	keyspace.Handle("/_import_filter", makeHandler(sc, adminPrivs, []Permission{PermReadAppData}, nil, (*handler).handleImportFilterDryRun)).Methods("GET")
```
adamcfraser, line 377: I went back to the spec (after looking at your changes from the last review), and it looks like this endpoint is supposed to support **either** a docID on the path, or providing the body directly.  I think that still makes sense - did you decide to change that intentionally?

adamcfraser: The same applies to _sync - the spec looks like we want to support either docID or a provided body. In this case it might be the spec that needs to be updated, but what do you think the ideal behaviour is for this one?

db/crud.go
```diff
@@ -2350,8 +2432,10 @@ func (col *DatabaseCollectionWithUser) getChannelsAndAccess(ctx context.Context,
 				err = base.HTTPErrorf(500, "JS sync function timed out")
 			} else {
 				err = base.HTTPErrorf(500, "Exception in JS sync function")
-				col.collectionStats.SyncFunctionExceptionCount.Add(1)
-				col.dbStats().Database().SyncFunctionExceptionCount.Add(1)
+				if !dryRun {
```
adamcfraser: I don't think we want to log a warning (as above) in the case of a dry run either.  The Info and Debug logging above should probably also include some indication that they are associated with a dry run.

rest/diagnostic_doc_api.go
```diff
@@ -39,3 +54,53 @@ func (h *handler) handleGetDocChannels() error {
 	h.writeJSON(resp)
 	return nil
 }
+
+// HTTP handler for running a document through the sync function and returning the results
+func (h *handler) handleSyncFnDryRun() error {
+	docid := h.getQuery("docid")
+	revID := h.getQuery("rev")
```
adamcfraser: Requiring callers to provide a revID for a dry run seems like unnecessary complexity.  The main purpose of exposing revID in the rest of the REST API generally is as a mechanism for optimistic concurrency, which we don't care about for this endpoint.  

I understand that we inject revID into doc and oldDoc when running the sync function, so it may make sense to populate those for proper sync function evaluation.  Can we populate revID values ourselves?  (i.e. treat as an insert if the document doesn't exist, create as a child if it does).  This will simplify future enhancements to work with VV-based revIDs as well.

rest/diagnostic_doc_api.go
```diff
@@ -39,3 +54,53 @@ func (h *handler) handleGetDocChannels() error {
 	h.writeJSON(resp)
 	return nil
 }
+
+// HTTP handler for running a document through the sync function and returning the results
+func (h *handler) handleSyncFnDryRun() error {
+	docid := h.getQuery("docid")
```
adamcfraser: The spec has docid on the path, and not as a query string parameter.  I guess this is an intentional change to support running without specifying a docid?  Should the spec get updated?  
I'm not exactly clear on the intended behaviour in the following combinations of input - can we document those in the comment for this function:
- doc id only, no rev, no body
- body only 
- doc id and body
- doc id, rev, body
- rev only?

rest/diagnostic_doc_api.go
```diff
@@ -39,3 +54,53 @@ func (h *handler) handleGetDocChannels() error {
...
+	revID := h.getQuery("rev")
+
+	body, err := h.readDocument()
```
adamcfraser, line 69: Is it required to specify a body for this call?  If yes, should we return error if it's empty?  If no, does line 64 succeed if not present?

rest/diagnostic_doc_api_test.go
```diff
@@ -57,3 +61,70 @@ func TestGetAlldocChannels(t *testing.T) {
 	assert.Equal(t, len(channelMap["CHAN3"]), db.DocumentHistoryMaxEntriesPerChannel+1)
 
 }
+
+func TestGetDocDryRuns(t *testing.T) {
```
adamcfraser, line 65: This looks like it needs negative tests for the error cases (no body, no docid, etc)

adamcfraser (changes requested):

rest/routing.go
mohammed-madi: I was going to update the spec once this is merged but if the body is provided I dont think theres a need for a doc id, and the docid can be an optional query for sync functions that use oldDoc

rest/diagnostic_doc_api.go
mohammed-madi: Fixed

rest/diagnostic_doc_api.go
mohammed-madi: if the body is empty and no doc id is not provided it will error, if the body is not provided but a doc id is provided, the doc in the bucket will be used as the body ( will return an error if the doc isnt there).
Ive added test cases for both scenarios

rest/diagnostic_doc_api.go
```diff
@@ -39,3 +56,64 @@ func (h *handler) handleGetDocChannels() error {
...
+// body only provided, the sync function will run with no oldDoc provided
+// body and doc ID provided, the sync function will run using the current revision in the bucket as oldDoc
+// docid only provided, if the doc exists in the bucket, sync fn results will be returned, otherwise it will error
+func (h *handler) handleSyncFnDryRun() error {
+	docid := h.getQuery("doc_id")
+
+	body, err := h.readDocument()
+	if err != nil {
+		if docid == "" {
+			return err
```
adamcfraser: We should return a more user-friendly message here - it's currently just returning "Bad JSON: EOF",

rest/diagnostic_doc_api_test.go
```diff
@@ -57,3 +61,110 @@ func TestGetAlldocChannels(t *testing.T) {
...
+	response = rt.SendDiagnosticRequest("GET", "/{{.keyspace}}/_import_filter", `{"accessUser": "user"}`)
+	RequireStatus(t, response, http.StatusOK)
+
+	var respMap2 ImportFilterDryRun
+	err = json.Unmarshal(response.BodyBytes(), &respMap2)
+	assert.NoError(t, err)
+	assert.Equal(t, respMap2.Error, "TypeError: Cannot access member 'num' of undefined")
+	assert.False(t, respMap2.ShouldImport)
+
+	// Import filter import=true and no error
```
adamcfraser, line 151: It looks like we need a test for the case where the import filter rejects the document but doesn't throw an error (shouldImport:false, error:"")

rest/diagnostic_doc_api_test.go
```diff
@@ -57,3 +61,110 @@ func TestGetAlldocChannels(t *testing.T) {
...
+
+	_ = rt.PutDoc("doc", `{"user":{"num":123, "name":"user1"}, "channel":"channel1"}`)
+
+	response = rt.SendDiagnosticRequest("GET", "/{{.keyspace}}/_sync", `{"user":{"num":23}}`)
+	RequireStatus(t, response, http.StatusOK)
+
+	err = json.Unmarshal(response.BodyBytes(), &respMap)
+	assert.NoError(t, err)
+	assert.Equal(t, respMap.Exception, "403 user num too low")
+
```
adamcfraser, line 108: We don't have a sync function test for an unexpected javascript error, as opposed to an error intentionally thrown by the Sync Function (e.g. attempting to access undefined property, etc).  Would be good to add one and make sure the error is clearly returned.

rest/diagnostic_doc_api.go
```diff
@@ -39,3 +56,64 @@ func (h *handler) handleGetDocChannels() error {
...
+		}
+	}
+	expiryPtr, channelSet, access, roles, err := h.collection.SyncFnDryrun(h.ctx(), body, docid)
```
adamcfraser: We can't currently differentiate between an error thrown during sync function execution and an error that occurs during dry run processing (e.g. specified docID not found).  I think those require different handling - I don't think we should be embedding the error in the SyncFnDryRun struct and returning status 200 in the second case.
SyncFnDryRun.Exception should only be used for sync function exceptions - other errors should be returned as errors by the handler.

docs/api/components/parameters.yaml
```diff
@@ -114,6 +114,14 @@ docid:
     type: string
   example: doc1
   description: The document ID to run the operation against.
+doc_id:
```
adamcfraser, line 117: Should we just use the existing docid (above) for consistency?

rest/diagnostic_doc_api.go
```diff
@@ -39,3 +56,64 @@ func (h *handler) handleGetDocChannels() error {
 	h.writeJSON(resp)
 	return nil
 }
+
+// HTTP handler for running a document through the sync function and returning the results
+// body only provided, the sync function will run with no oldDoc provided
+// body and doc ID provided, the sync function will run using the current revision in the bucket as oldDoc
+// docid only provided, if the doc exists in the bucket, sync fn results will be returned, otherwise it will error
```
adamcfraser:
```suggestion
// docid only provided, the sync function will run using the current revision in the bucket as doc
// If docid is specified and the document does not exist in the bucket, it will return error
```

rest/diagnostic_doc_api.go
```diff
@@ -39,3 +56,64 @@ func (h *handler) handleGetDocChannels() error {
...
+		if docid == "" {
+			return err
+		}
+	}
+	expiryPtr, channelSet, access, roles, err := h.collection.SyncFnDryrun(h.ctx(), body, docid)
+	errorMsg := ""
+	if err != nil {
+		errorMsg = err.Error()
+	}
+	expiry := uint32(0)
```
adamcfraser: We can't differentiate between Sync Functions that explicitly set expiry to zero and Sync Functions that don't set expiry with this handling.  I think Expiry needs to be *uint32 in SyncFnDryRun, potentially with omitempty?

db/crud.go
```diff
@@ -2332,14 +2421,18 @@ func (col *DatabaseCollectionWithUser) getChannelsAndAccess(ctx context.Context,
...
-				base.InfofCtx(ctx, base.KeyAll, "Sync fn rejected doc %q / %q --> %s", base.UD(doc.ID), base.UD(doc.NewestRev), err)
-				base.DebugfCtx(ctx, base.KeyAll, "    rejected doc %q / %q : new=%+v  old=%s", base.UD(doc.ID), base.UD(doc.NewestRev), base.UD(body), base.UD(oldJson))
-				col.dbStats().Security().NumDocsRejected.Add(1)
-				col.collectionStats.SyncFunctionRejectCount.Add(1)
-				if isAccessError(err) {
-					col.dbStats().Security().NumAccessErrors.Add(1)
-					col.collectionStats.SyncFunctionRejectAccessCount.Add(1)
+				dryRunMsg := ""
+				if !dryRun {
+					dryRunMsg = "[DRY RUN]"
```
adamcfraser: This looks reversed - you're setting dryRunMsg inside `!dryRun`.

db/crud.go
```diff
@@ -2332,14 +2421,18 @@ func (col *DatabaseCollectionWithUser) getChannelsAndAccess(ctx context.Context,
...
+					col.dbStats().Security().NumDocsRejected.Add(1)
+					col.collectionStats.SyncFunctionRejectCount.Add(1)
+					if isAccessError(err) {
+						col.dbStats().Security().NumAccessErrors.Add(1)
+						col.collectionStats.SyncFunctionRejectAccessCount.Add(1)
+					}
 				}
+				base.InfofCtx(ctx, base.KeyAll, "%s Sync fn rejected doc %q / %q --> %s", dryRunMsg, base.UD(doc.ID), base.UD(doc.NewestRev), err)
```
```suggestion
				base.InfofCtx(ctx, base.KeyAll, "%sSync fn rejected doc %q / %q --> %s", dryRunMsg, base.UD(doc.ID), base.UD(doc.NewestRev), err)
```
(to avoid whitespace logging change in the non-dry-run case)

db/crud.go
adamcfraser: We're still logging a warning here (line 2441) in the dry run case, which we don't want to do.  You can use the warn_count stat to write a unit test to make sure the diagnostic endpoints aren't triggering these warnings.

db/crud.go
```diff
@@ -1763,8 +1850,9 @@ func (col *DatabaseCollectionWithUser) documentUpdateFunc(ctx context.Context, d
 	doc.updateWinningRevAndSetDocFlags(ctx)
 	newDocHasAttachments := len(newAttachments) > 0
 	col.storeOldBodyInRevTreeAndUpdateCurrent(ctx, doc, prevCurrentRev, newRevID, newDoc, newDocHasAttachments)
+	base.DebugfCtx(ctx, base.KeyDiagnostic, "NEWREVID in documentUpdateFunc: %v", newDoc.RevID)
```
adamcfraser: Should this logging be wrapped in dry run check?

db/crud.go
```diff
@@ -1555,7 +1642,7 @@ func (db *DatabaseCollectionWithUser) recalculateSyncFnForActiveRev(ctx context.
 	if curBody != nil {
 		base.DebugfCtx(ctx, base.KeyCRUD, "updateDoc(%q): Rev %q causes %q to become current again",
 			base.UD(doc.ID), newRevID, doc.CurrentRev)
-		channelSet, access, roles, syncExpiry, oldBodyJSON, err = db.getChannelsAndAccess(ctx, doc, curBody, metaMap, doc.CurrentRev)
+		channelSet, access, roles, syncExpiry, oldBodyJSON, err = db.getChannelsAndAccess(ctx, doc, curBody, metaMap, doc.CurrentRev, false)
```
adamcfraser: Should be passing dryRun instead of false here.

db/import.go
```diff
@@ -500,3 +500,20 @@ func (i *ImportFilterFunction) EvaluateFunction(ctx context.Context, doc Body) (
...
+		docInBucket, err := db.GetDocument(ctx, docid, DocUnmarshalAll)
+		if err == nil {
+			if doc == nil {
+				doc = docInBucket._body
+			}
+		} else {
+			return false, err
+		}
+	}
+	shouldImport, err := importFilter.EvaluateFunction(ctx, doc)
```
adamcfraser: importFilter.EvaluateFunction logs warnings - looks like we need to pass a dryRun flag to it as well.

db/import.go
```diff
@@ -500,3 +500,20 @@ func (i *ImportFilterFunction) EvaluateFunction(ctx context.Context, doc Body) (
...
 	}
 }
+func (db *DatabaseCollectionWithUser) ImportFilterDryRun(ctx context.Context, doc Body, docid string) (bool, error) {
+
+	importFilter := db.importFilter()
+	if docid != "" {
+		docInBucket, err := db.GetDocument(ctx, docid, DocUnmarshalAll)
+		if err == nil {
+			if doc == nil {
+				doc = docInBucket._body
```
adamcfraser: Should use docInBucket.Body() to access the unmarshalled body.

db/crud.go
```diff
@@ -1165,6 +1165,93 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
...
+	doc := &Document{
+		ID:    docID,
+		_body: body,
+	}
+	oldDoc := doc
+	if docID != "" {
+		if docInBucket, err := db.GetDocument(ctx, docID, DocUnmarshalAll); err == nil {
+			oldDoc = docInBucket
+			if doc._body == nil {
+				body = oldDoc._body
```
adamcfraser: Should use Body(ctx) instead of _body to access document body.

db/crud.go
```diff
@@ -1165,6 +1165,93 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
 
 }
 
+func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, body Body, docID string) (*uint32, base.Set, channels.AccessMap, channels.AccessMap, error) {
+	doc := &Document{
+		ID:    docID,
+		_body: body,
+	}
+	oldDoc := doc
```
adamcfraser, line 1154: This looks like if a body is passed in and docID is not found, we're using the body as both doc and oldDoc.  I don't think this is what users will want to do - if they are only passing body, I think it should behave as an insert (oldDoc=null)

db/crud.go
```diff
@@ -1165,6 +1165,93 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
...
+	delete(body, BodyId)
+
+	// Get the revision ID to match, and the new generation number:
+	matchRev, _ := body[BodyRev].(string)
+	generation, _ := ParseRevID(ctx, matchRev)
+	if generation < 0 {
+		return nil, nil, nil, nil, base.HTTPErrorf(http.StatusBadRequest, "Invalid revision ID")
+	}
+	generation++
+	oldDoc.RevID = body.ExtractRev()
```
adamcfraser: What body is this extracting the rev from?  It looks like it's the body that was passed in - I expected it to be docInBucket.SyncData.CurrentRev.  Do we have unit tests that validate that rev is being properly populated into doc and oldDoc when doing a dry run of the sync function?

db/crud.go
```diff
@@ -1165,6 +1165,93 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
...
+	newRev := CreateRevIDWithBytes(generation, matchRev, canonicalBytesForRevID)
+	newDoc.RevID = newRev
+	mutableBody, metaMap, newRevID, err := db.prepareSyncFn(oldDoc, newDoc)
+	if err != nil {
+		base.InfofCtx(ctx, base.KeyDiagnostic, "Failed to prepare to run sync function: %v", err)
+		return nil, nil, nil, nil, err
+	}
+	if oldDoc.History != nil {
+		if err := oldDoc.History.addRevision(newDoc.ID, RevInfo{ID: newRev, Parent: matchRev, Deleted: isDeleted}); err != nil {
+			base.InfofCtx(ctx, base.KeyCRUD, "Failed to add revision ID: %s, for doc: %s, error: %v", newRev, base.UD(doc.ID), err)
```
adamcfraser: We shouldn't log for a dry run here.

db/crud.go
```diff
@@ -1165,6 +1165,93 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
...
+			return nil, nil, nil, nil, base.ErrRevTreeAddRevFailure
+		}
+	}
+
+	syncExpiry, _, channelSet, access, roles, err := db.runSyncFn(ctx, oldDoc, mutableBody, metaMap, newRevID, true)
```
adamcfraser: Looking at this again, it seems like a lot of the work you're having to do with the dryRun flag could be avoided if this function was calling the lower level DatabaseCollectionWithUser.ChannelMapper.MapToChannelsAndAccess directly (which actually does the sync function execution).  Do we need any of the other handling in runSyncFn or getChannelsAndAccess for the dry run?

I can see how checkDocChannelsAndGrantsLimits information would actually be useful for a dry run, but you've got that disabled at the moment anyway (since we don't want to warn/log).  Maybe call DatabaseCollectionWithUser.ChannelMapper.MapToChannelsAndAccess  and pass the results to a refactored checkDocChannelsAndGrantsLimits that will allow us to return the warnings in the SyncFnDryRun response?

adamcfraser (changes requested):

docs/api/components/parameters.yaml
mohammed-madi: that one is required in the path, for this endpoint its optional as a query param

db/crud.go
mohammed-madi: Its treated as an insert if there is no docID, but olddoc is used in prepareSyncFn. If doc id is not found it returns 404

db/crud.go
mohammed-madi: If no doc id is given, the rev id is not needed anywhere. It's only used for getting the parent rev if a doc id is given. I verified it with debug logging and I would probably have to return the rev id in the dry run response if we wanted to write unit tests to check it

rest/diagnostic_doc_api_test.go
mohammed-madi: Added one that throws TypeError: Cannot access member 'num' of undefined

rest/diagnostic_doc_api_test.go
mohammed-madi: Added

rest/diagnostic_doc_api.go
```diff
@@ -11,11 +11,29 @@ licenses/APL2.txt.
...
 )
 
-// HTTP handler for a GET of a document
+type SyncFnDryRun struct {
+	Channels  base.Set           `json:"Channels"`
+	Access    channels.AccessMap `json:"Access"`
+	Roles     channels.AccessMap `json:"Roles"`
+	Exception string             `json:"Exception"`
+	Expiry    *uint32            `json:"Expiry,omitempty"`
+}
```
adamcfraser, line 29: json: names should be lower-case for consistency with the rest of (or at least the majority of) our APIs.

db/crud.go
adamcfraser: If this is an insert, we need to ensure we're not populating oldDoc body in the sync function (people trying to test sync functions that have different handling for insert/update aren't going to work).  If oldDoc needs to be non-nil, I expect you at least need to set oldDoc._body=nil?  Or is this handled in another way that I'm overlooking?

db/crud.go
adamcfraser: My concern was whether oldDoc._rev is being set for use by the sync function.  Looking at TestSyncFnOldDocBodyProperties it seems like we don't populate it in any case.  

It's still be incorrect to set oldDoc.RevID to the value specified in body though (which is doc.rev, not oldDoc.rev).  Does something break if you omit setting this property, if you don't think it's used?

db/crud.go
```diff
@@ -1165,6 +1165,102 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
...
+	}
+
+	newRev := CreateRevIDWithBytes(generation, matchRev, canonicalBytesForRevID)
+	newDoc.RevID = newRev
+	mutableBody, metaMap, newRevID, err := db.prepareSyncFn(oldDoc, newDoc)
+	if err != nil {
+		base.InfofCtx(ctx, base.KeyDiagnostic, "Failed to prepare to run sync function: %v", err)
+		return nil, err, nil
+	}
+	if oldDoc.History != nil {
```
adamcfraser: I think this should be doc instead of oldDoc (along with lines 1242)?  The new rev should only be added to the new doc.
This might not be necessary at all, if you were just doing it to make line 1248 work.

db/crud.go
```diff
@@ -1165,6 +1165,102 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
...
+			return nil, base.ErrRevTreeAddRevFailure, nil
+		}
+	}
+	// Get the parent revision, to pass to the sync function:
+	var oldJsonBytes []byte
+	if oldJsonBytes, err = db.getAncestorJSON(ctx, oldDoc, newRevID); err != nil {
+		return nil, err, nil
+	}
+	oldJson := string(oldJsonBytes)
+	base.InfofCtx(ctx, base.KeyDiagnostic, "old json: %s", oldJson)
```
adamcfraser: This looks like dev-time logging?  Or do you think it has value in production?

db/crud.go
```diff
@@ -1165,6 +1165,102 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
...
+	output, err := db.ChannelMapper.MapToChannelsAndAccess(ctx, mutableBody, oldJson, metaMap,
+		MakeUserCtx(db.user, db.ScopeName, db.Name))
+
+	//syncExpiry, _, channelSet, access, roles, err := db.runSyncFn(ctx, oldDoc, mutableBody, metaMap, newRevID, true)
```
adamcfraser: Can remove the commented code.

db/crud.go
```diff
@@ -1165,6 +1165,102 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
...
+		return nil, err, nil
+	}
+	if oldDoc.History != nil {
+		if err := oldDoc.History.addRevision(newDoc.ID, RevInfo{ID: newRev, Parent: matchRev, Deleted: isDeleted}); err != nil {
+			return nil, base.ErrRevTreeAddRevFailure, nil
+		}
+	}
+	// Get the parent revision, to pass to the sync function:
+	var oldJsonBytes []byte
+	if oldJsonBytes, err = db.getAncestorJSON(ctx, oldDoc, newRevID); err != nil {
```
adamcfraser: I don't think we need to look up the ancestor JSON.  We've already populated oldDoc if we want to populate it  (based on how the dry run is being executed) - we just need to marshal its body, don't we?

adamcfraser (changes requested):

db/crud.go
mohammed-madi: removed

db/crud.go
mohammed-madi: removed

db/crud.go
mohammed-madi: removed, was copied over from getChannelsAndAccess while trying to fix oldDoc handling

rest/diagnostic_doc_api_test.go
```diff
@@ -57,3 +61,150 @@ func TestGetAlldocChannels(t *testing.T) {
...
+	response = rt.SendDiagnosticRequest("GET", fmt.Sprintf("/{{.keyspace}}/_sync?doc_id=doc"), `{"user":{"num":120, "name":["user2"]}, "channel":"channel2"}`)
+	RequireStatus(t, response, http.StatusOK)
+
+	err = json.Unmarshal(response.BodyBytes(), &respMap)
+	assert.NoError(t, err)
+	assert.Equal(t, respMap.Access, channels.AccessMap{"user1": channels.BaseSetOf(t, "channel2")})
+
+	// get doc from bucket with no body provided
+	response = rt.SendDiagnosticRequest("GET", fmt.Sprintf("/{{.keyspace}}/_sync?doc_id=doc"), ``)
```
adamcfraser, line 124: This case is using the document from the bucket as *both* doc and oldDoc in the sync function.  Is this intentional?    Using it as doc only seems like a more natural use case to me, but I'm open to other points of view.

adamcfraser (changes requested):
Everything looks good, except for one question about whether setting both doc and oldDoc is intentional when only docID is specified.

rest/diagnostic_doc_api_test.go
mohammed-madi: I remember we discussed allowing this in sync up, either we use it as doc only or we attempt to get the previous rev as oldDoc?

mohammed-madi: Ive made the change to use it as doc only, let me know if we need to pull the old rev as old doc

adamcfraser (approved):

PR #6713, opened 2024-03-04
Improve registry logging errors

torcolvin:
PREVIEW ONLY of potential feature/issues.

## Pre-review checklist
- [ ] Removed debug logging (`fmt.Print`, `log.Print`, ...)
- [ ] Logging sensitive data? Make sure it's tagged (e.g. `base.UD(docID)`, `base.MD(dbName)`)
- [ ] Updated relevant information in the API specifications (such as endpoint descriptions, schemas, ...) in `docs/api`

## Dependencies (if applicable)
- [ ] Link upstream PRs
- [ ] Update Go module dependencies when merged

## [Integration Tests](https://jenkins.sgwdev.com/job/SyncGateway-Integration/build?delay=0sec)
- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/000/

PR #6720, opened 2024-03-07
CBG-3745: Bail out of config updates when the registry write fails after exceeding retry attempts

bbrks:
CBG-3745

Partly taken from #6713 

- Ensures config/registry operations abort when the number of retry attempts exceeds the limit
- Logging for `GetDatabaseConfigs` retry attempts
- Use `RedactErrorf` for errors that contained metadata

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2339/

### DIFF
--- a/rest/config_manager.go
+++ b/rest/config_manager.go
@@ -67,11 +67,8 @@ func (b *bootstrapContext) InsertConfig(ctx context.Context, bucketName, groupID
 	dbName := config.Name
 	ctx = b.addDatabaseLogContext(ctx, &config.DbConfig)
 
+	registryUpdated := false
 	for attempt := 1; attempt <= configUpdateMaxRetryAttempts; attempt++ {
-		if attempt == configUpdateMaxRetryAttempts {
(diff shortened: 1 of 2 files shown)


bbrks:
Requires 3.1.4 backport once merged (CBG-3749)

rest/config_manager.go
```diff
@@ -263,21 +276,29 @@ outer:
...
 			return writeErr
 		}
 
+		base.InfofCtx(ctx, base.KeyConfig, "DeleteConfig encountered cas mismatch error while persisting updated registry: %v", writeErr)
+
 		// Check for context cancel before retrying
 		select {
 		case <-ctx.Done():
-			break outer
+			break
```
torcolvin: Should this return here? I'm not sure if this should be an error or not.

adamcfraser: I think you're right - updated that to match insert and update.

torcolvin (approved):

torcolvin (approved):

PR #6724, opened 2024-03-12
CBG-3823 log warning if released unused sequence count > 1 million

torcolvin:
Tested on `TestAssignSequenceReleaseLoop` with a lowered threshold to make sure we'll see a warning message.

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -1684,6 +1684,11 @@ func (db *DatabaseContext) assignSequence(ctx context.Context, docSequence uint6
 			if err != nil {
 				return unusedSequences, err
 			}
+			numSkippedSequences := docSequence - doc.Sequence
+			if numSkippedSequences > oversizeSkippedSequenceWarning {
+				base.WarnfCtx(ctx, "doc %s / previous rev: %s /  new rev: %s be allocated sequence %d. This is significantly ahead of the previous sequence this document had %d.", base.UD(doc.ID), doc.CurrentRev, newRevID, docSequence, doc.Sequence)
+			}
+
 		}
 	}
 
--- a/db/crud.go
+++ b/db/crud.go
@@ -26,7 +26,7 @@ import (
 
 const (
 	kMaxRecentSequences            = 20    // Maximum number of sequences stored in RecentSequences before pruning is triggered
-	oversizeSkippedSequenceWarning = 10000 // Number of skipped sequences to issue a warning
+	unusedSequenceWarningThreshold = 10000 // Warn when releasing more than this many sequences due to existing sequence on the document
 )
 
 // ErrForbidden is returned when the user requests a document without a revision that they do not have access to.
@@ -1648,15 +1648,15 @@ func (db *DatabaseCollectionWithUser) addAttachments(ctx context.Context, newAtt
 }
 
 // assignSequence assigns a global sequence number from database.
-func (c *DatabaseCollectionWithUser) assignSequence(ctx context.Context, docSequence uint64, doc *Document, newRevID string, unusedSequences []uint64) ([]uint64, error) {
-	return c.dbCtx.assignSequence(ctx, docSequence, doc, newRevID, unusedSequences)
+func (c *DatabaseCollectionWithUser) assignSequence(ctx context.Context, docSequence uint64, doc *Document, unusedSequences []uint64) ([]uint64, error) {
+	return c.dbCtx.assignSequence(ctx, docSequence, doc, unusedSequences)
 }
 
 // Sequence processing :
 // Assigns provided sequence to the document
 // Update unusedSequences in the event that there is a conflict and we have to provide a new sequence number
 // Update and prune RecentSequences
-func (db *DatabaseContext) assignSequence(ctx context.Context, docSequence uint64, doc *Document, newRevID string, unusedSequences []uint64) ([]uint64, error) {
+func (db *DatabaseContext) assignSequence(ctx context.Context, docSequence uint64, doc *Document, unusedSequences []uint64) ([]uint64, error) {
 
 	// Assign the next sequence number, for _changes feed.
 	// Be careful not to request a second sequence # on a retry if we don't need one.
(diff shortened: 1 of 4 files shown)


db/crud.go
```diff
@@ -1617,6 +1618,10 @@ func (db *DatabaseContext) assignSequence(ctx context.Context, docSequence uint6
 			if err = db.sequences.releaseSequence(ctx, docSequence); err != nil {
 				base.WarnfCtx(ctx, "Error returned when releasing sequence %d. Falling back to skipped sequence handling.  Error:%v", docSequence, err)
 			}
+			numSkippedSequences := doc.Sequence - docSequence
```
adamcfraser: I think we should move this after the call to nextSequenceGreaterThan (i.e. base it on the sequence that's actually going to be assigned to the document).  This has the additional benefit of catching any potential issues in how nextSequenceGreaterThan is computing the sequence.

db/crud.go
```diff
@@ -1617,6 +1618,10 @@ func (db *DatabaseContext) assignSequence(ctx context.Context, docSequence uint6
...
+			if numSkippedSequences > oversizeSkippedSequenceWarning {
+				base.WarnfCtx(ctx, "doc %s / previous rev: %s new rev: %s has sequence %d which is significantly ahead of the known highest sequence %d", base.UD(doc.ID), doc.CurrentRev, newRevID, doc.Sequence, docSequence)
```
adamcfraser: It's not immediately obvious to me what 'known highest sequence' means here.  Can we change this to 'next allocated sequence'?

db/crud.go
```diff
@@ -25,7 +25,8 @@ import (
 )
 
 const (
-	kMaxRecentSequences = 20 // Maximum number of sequences stored in RecentSequences before pruning is triggered
+	kMaxRecentSequences            = 20      // Maximum number of sequences stored in RecentSequences before pruning is triggered
+	oversizeSkippedSequenceWarning = 1000000 // Number of skipped sequences to issue a warning for
```
adamcfraser: I think this number could be a lot smaller.  Any case where a document is allocated a new sequence that's smaller than its previous sequence is unexpected, even at high throughput.  So I think something like 10K would be reasonable.

adamcfraser: This comment hasn't been addressed yet.

db/crud.go
```diff
@@ -1621,6 +1622,11 @@ func (db *DatabaseContext) assignSequence(ctx context.Context, docSequence uint6
 			if err != nil {
 				return unusedSequences, err
 			}
+			numSkippedSequences := docSequence - doc.Sequence
```
adamcfraser: What we want to evaluate here is the difference between the sequence we attempted to assign on line 1611 (which was where the sequence allocator was at that point in time) and the sequence we get back from nextSequenceGreaterThan on line 1621.  
The warning should include all three sequences (line 1611, doc.Sequence, and line 1621)

adamcfraser (changes requested):

db/crud.go
torcolvin: I think I addressed this, but I'm a bit confused and the warning messages could probably read better. I'd appreciate feedback here.

db/crud.go
```diff
@@ -25,7 +25,8 @@ import (
 )
 
 const (
-	kMaxRecentSequences = 20 // Maximum number of sequences stored in RecentSequences before pruning is triggered
+	kMaxRecentSequences            = 20    // Maximum number of sequences stored in RecentSequences before pruning is triggered
+	oversizeSkippedSequenceWarning = 10000 // Number of skipped sequences to issue a warning
```
```suggestion
	unusedSequenceWarningThreshold = 10000 // Warn when releasing more than this many sequences due to existing sequence on the document
```

db/crud.go
```diff
@@ -1684,6 +1686,11 @@ func (db *DatabaseContext) assignSequence(ctx context.Context, docSequence uint6
 			if err != nil {
 				return unusedSequences, err
 			}
+			numSkippedSequences := docSequence - doc.Sequence
```
```suggestion
			numReleasedSequences := docSequence - firstAllocatedSequence
```
We use the 'skipped' terminology when talking about sequences being processed by the caching DCP feed, and 'unused' or 'released' sequences when talking about sequence allocation at write time.

What we're trying to report here is the jump in _sync:seq, so we want to compare docSequence against firstAllocatedSequence and not doc.Sequence.
However, it would be even better to update nextSequenceGreaterThan to return the exact of released sequences, and report that. I think we should take that approach, instead of guessing about what the sequence allocator is doing.

db/crud.go
```diff
@@ -1684,6 +1686,11 @@ func (db *DatabaseContext) assignSequence(ctx context.Context, docSequence uint6
...
+			if numSkippedSequences > oversizeSkippedSequenceWarning {
+				base.WarnfCtx(ctx, "doc %s / previous rev: %s /  new rev: %s be allocated sequence %d but standard next sequence allocation would be %d. This is significantly ahead of the previous sequence this document had %d.", base.UD(doc.ID), doc.CurrentRev, newRevID, docSequence, firstAllocatedSequence, doc.Sequence)
```
```suggestion
				base.WarnfCtx(ctx, "Doc %s / %s had an existing sequence %d that is higher than the next db sequence value %d, resulting in the release of %d unused sequences. This may indicate documents being migrated between databases by an external process", base.UD(doc.ID), doc.CurrentRev, doc.Sequence, firstAllocatedSequence, unusedSequenceRange)
```

adamcfraser (changes requested):

db/sequence_allocator.go
```diff
@@ -251,23 +253,26 @@ func (s *sequenceAllocator) nextSequenceGreaterThan(ctx context.Context, existin
 	s.dbStats.SequenceReservedCount.Add(int64(numberToRelease + numberToAllocate))
 	s.dbStats.SequenceAssignedCount.Add(1)
 
-	// Release previously allocated sequences (c), if any
-	err = s.releaseSequenceRange(ctx, prevAllocReleaseFrom, prevAllocReleaseTo)
-	if err != nil {
-		base.WarnfCtx(ctx, "Error returned when releasing sequence range [%d-%d] for previously allocated sequences. Will be handled by skipped sequence handling.  Error:%v", prevAllocReleaseFrom, prevAllocReleaseTo, err)
+	if prevAllocReleaseFrom < prevAllocReleaseTo {
```
adamcfraser: Shouldn't this be prevAllocReleaseFrom <= prevAllocReleaseTo?  Beyond that, I think this condition is already handled inside releaseSequenceRange.  Given that this is a logging enhancement it would be good to avoid changes that could introduce new regressions, unless you spotted a scenario that was previously being mishandled.

adamcfraser (changes requested):

db/sequence_allocator.go
torcolvin: Good catch, this was from earlier when I realized that `prevAllocReleaseTo` could be -1.

adamcfraser (approved):

PR #6739, opened 2024-03-20
CBG-3764 create multi-xattr subdoc APIs

torcolvin:
Update multi xattr subdoc APIs to support multiple xattrs. Where the APIs previously accepted two xattrs (user xattr and sync xattr), continue to allow two xattrs with CBS 7.6 with a CAS check in a loop. If the API previously did not accept two APIs, Couchbase Server 7.2 will return an invalid argument error if multiple xattrs are not specified.

Possible enhancements:

- enhance rosmar to be able to toggle on `BucketStoreFeatureMultiXattrSubdocOperations` to be able to test without running two versions of couchbase server
- Writing a check in all xattr APIs to make sure there are only 1 or 2 supported xattrs if using CBS < 7.6 which provides a bit easier message than one.
- Write a check in all xattr APIs to make sure that `xattrKeys` are never empty string. This was a late fix in integration tests from a somewhat opaque gocb error message. Alternatively, error in rosmar if any `xattrKeys` are empty.
- Move `ErrXattrNotFound` and `ErrXattrPartialFound` to sg-bucket and return in rosmar?

Things to look for in the PR:

- Did I remove any operations that should be retried to work a single time?
- Should I rename any functions to make it more clear which is the function that will do a cas based retry and which does the actual operation?
- Double checking any cases where I change logging for redaction OR error wrapping.

https://github.com/couchbaselabs/rosmar/pull/25
https://github.com/couchbase/sg-bucket/pull/112

## Pre-review checklist

- [x] Link upstream PRs

- [x] `GSI=true,xattrs=true,CBS-7.2.2` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2366/
- [x] `GSI=true,xattrs=true,CBS-7.6.1-3122` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2364/


base/collection_xattr.go
```diff
@@ -293,17 +293,18 @@ func (c *Collection) subdocGetBodyAndXattrs(ctx context.Context, k string, xattr
 
 		c.Bucket.waitForAvailKvOp()
 		defer c.Bucket.releaseKvOp()
-
 		// First, attempt to get the document and xattr in one shot.
 		ops := make([]gocb.LookupInSpec, 0, len(xattrKeys)+1)
 		for _, xattrKey := range xattrKeys {
+			if xattrKey == "" {
```
adamcfraser: Do we care that the mutateIn functions that iterate over xattrKeys don't have similar defensive checks for empty keys?  (it's fine if the answer is 'no we don't' - just wanted to get your thoughts on it).

db/database.go
```diff
@@ -1826,7 +1826,11 @@ func (db *DatabaseCollectionWithUser) resyncDocument(ctx context.Context, docid,
 		opts := &sgbucket.MutateInOptions{
 			MacroExpansion: macroExpandSpec(base.SyncXattrName),
 		}
-		_, err = db.dataStore.WriteUpdateWithXattrs(ctx, key, []string{base.SyncXattrName, db.userXattrKey()}, 0, nil, opts, writeUpdateFunc)
+		xattrKeys := []string{base.SyncXattrName}
+		if db.userXattrKey() != "" {
+			xattrKeys = append(xattrKeys, db.userXattrKey())
+		}
```
adamcfraser: A helper function to build the xattr keys wouldn't be a bad idea, since we're going to need to enhance it to add _vv in the near future.

base/collection_xattr.go
```diff
@@ -191,53 +188,99 @@ func (c *Collection) SubdocWrite(ctx context.Context, k string, subdocKey string
 	return casOut, err
 }
 
-// SubdocGetBodyAndXattr retrieves the document body and xattr in a single LookupIn subdoc operation.  Does not require both to exist.
-func (c *Collection) SubdocGetBodyAndXattr(ctx context.Context, k string, xattrKey string, userXattrKey string, rv interface{}, xv interface{}, uxv interface{}) (cas uint64, err error) {
+// subdocGetBodyAndXattr retrieves the document body and xattrs in a single LookupIn subdoc operation.  Does not require both to exist.
+func (c *Collection) subdocGetBodyAndXattrs(ctx context.Context, k string, xattrKeys []string, fetchBody bool) (rawBody []byte, xattrs map[string][]byte, cas uint64, err error) {
+	xattrKey2 := ""
+	if !c.IsSupported(sgbucket.BucketStoreFeatureMultiXattrSubdocOperations) {
+		if len(xattrKeys) > 2 {
```
adamcfraser, line 196: I think a comment might be merited here to explain that this is backward compatibility for "one system xattr and one user xattr" support.

For that matter, should we return error in this case if xattrKey2 is a system xattr (underscore prefix), to prevent potential misuse?

base/collection_xattr.go
```diff
@@ -247,42 +290,43 @@ func (c *Collection) SubdocGetBodyAndXattr(ctx context.Context, k string, xattrK
 			shouldRetry = c.isRecoverableReadError(lookupErr)
 			return shouldRetry, lookupErr, uint64(0)
 		}
-
-		// TODO: We may be able to improve in the future by having this secondary op as part of the first. At present
-		// there is no support to obtain more than one xattr in a single operation however MB-28041 is filed for this.
-		if userXattrKey != "" {
-			userXattrCas, userXattrErr := c.SubdocGetXattr(ctx, k, userXattrKey, uxv)
-			switch pkgerrors.Cause(userXattrErr) {
+		// If Couchbase Server < 7.6, do a second get for the second xattr.
```
adamcfraser: It's not immediately obvious that "Couchbase Server < 7.6" corresponds to `c.IsSupported(sgbucket.BucketStoreFeatureMultiXattrSubdocOperations` - maybe just change this comment to `If BucketStoreFeatureMultiXattrSubdocOperations not supported...`

adamcfraser (changes requested):
A few minor comments/questions but otherwise looks good to me.

base/collection_xattr.go
torcolvin: I think I'm inclined to drop this defensive check, the error message is fine from gocb `invalid xattr fetch with no path`. I had added this during debug.

When I went to look at adding this elsewhere, it seemed more confusing than not to add. The enhancement I think would be worth doing is making sure that we can't pass an empty xattr to rosmar https://issues.couchbase.com/browse/CBG-3858

base/collection_xattr.go
torcolvin: I was worried about ordering here for these arguments, but I'm open to the idea. The other idea would be to create a non sg-bucket function but inside Sync Gateway for getting a system xattr and a non system xattr, so the caller would be aware that they are taking a performance penalty.

The order of the arguments is trivially checkable, and to me it is an implementation detail that the xattrs we are adding will be system xattrs but the performance cost is present for any time there is more than one xattr.

adamcfraser (dismissed):

adamcfraser:
LGTM - ready to merge once dependencies are merged and go.mod updated.

adamcfraser (approved):

PR #6747, opened 2024-03-28
CBG-3852: Implementation of single sequence entry for skipped sequence slice

gregns1:
CBG-3852

Initial implementation of skipped sequence interfcae and addition of a single skipped sequence entry.
There are some incomplete functions that will require ranges to be implemented so have left comments on these functions/tests. 
Also includes Benchmarks for each operation.  

## Pre-review checklist

### DIFF
--- a/db/skipped_sequence.go
+++ b/db/skipped_sequence.go
@@ -1,0 +1,211 @@
+/*
+Copyright 2016-Present Couchbase, Inc.
+
+Use of this software is governed by the Business Source License included in
+the file licenses/BSL-Couchbase.txt.  As of the Change Date specified in that
+file, in accordance with the Business Source License, use of this software will
+be governed by the Apache License, Version 2.0, included in the file
+licenses/APL2.txt.
+*/
+
+package db
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"sync"
+	"time"
+
+	"github.com/couchbase/sync_gateway/base"
+)
+
+const (
+	ClipCapacityHeadroom = 100
+)
+
+type SkippedSequenceListInterface interface {
+	getTimestamp() int64
+	getLastSeq() uint64
+	setLastSeq(seq uint64)
+	getStartSeq() uint64
+	isRange() bool
+}
+
+// SkippedSequenceSlice stores the set of skipped sequences as an ordered slice of single skipped sequences
+// or skipped sequence ranges
+type SkippedSequenceSlice struct {
+	list []SkippedSequenceListInterface
+	lock sync.RWMutex
+}
+
+var _ SkippedSequenceListInterface = &SingleSkippedSequence{}
+
+// SingleSkippedSequence contains a single skipped sequence value + unix timestamp of the time it's created
+type SingleSkippedSequence struct {
+	seq       uint64
+	timestamp int64
+}
+
+func NewSkippedSequenceSlice() *SkippedSequenceSlice {
+	return &SkippedSequenceSlice{
+		list: []SkippedSequenceListInterface{},
+	}
+}
+
+// NewSingleSkippedSequenceEntry returns a SingleSkippedSequence with the specified sequence and the current
+// time in unix time
+func NewSingleSkippedSequenceEntry(sequence uint64) *SingleSkippedSequence {
+	return &SingleSkippedSequence{
+		seq:       sequence,
+		timestamp: time.Now().Unix(),
+	}
+}
+
+// getTimestamp returns the timestamp of the entry
+func (s *SingleSkippedSequence) getTimestamp() int64 {
+	return s.timestamp
+}
+
+// getLastSeq gets the last sequence in the range on the entry, for single items the sequence will be returned
+func (s *SingleSkippedSequence) getLastSeq() uint64 {
+	return s.seq
+}
+
+// setLastSeq sets the last sequence in the range on the entry, for single items the single sequence will be set
+func (s *SingleSkippedSequence) setLastSeq(seq uint64) {
+	diff := seq - s.seq
+	s.seq = s.seq + diff
+}
+
+// getStartSeq gets the start sequence in the range on the entry, for single items the sequence will be returned
+func (s *SingleSkippedSequence) getStartSeq() uint64 {
+	return s.seq
+}
+
+// isRange returns true if the entry is a sequence range entry, false if not
+func (s *SingleSkippedSequence) isRange() bool {
+	return false
+}
+
+// IsSkipped returns true if a given sequence exists in the skipped sequence slice
+func (s *SkippedSequenceSlice) IsSkipped(x uint64) bool {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+	_, found := s.findSequence(x)
+	return found
+}
+
+// SkippedSequenceCompact will compact the entries with timestamp old enough. It will also clip
+// the capacity of the slice to length + 100 if the current capacity is 2.5x the length
+func (s *SkippedSequenceSlice) SkippedSequenceCompact(ctx context.Context, maxWait int64) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	timeNow := time.Now().Unix()
+	indexToDelete := -1
+	for _, v := range s.list {
+		timeStamp := v.getTimestamp()
+		if (timeNow - timeStamp) >= maxWait {
+			indexToDelete++
+		} else {
+			// exit early to avoid iterating through whole slice
+			break
+		}
+	}
+	// if we have incremented the index counter, we need to the index range from the slice
+	if indexToDelete > -1 {
+		s.list = slices.Delete(s.list, 0, indexToDelete+1)
+	}
+	// resize slice to reclaim memory if we need to
+	s.clip(ctx)
+}
+
+// clip will clip the capacity of the slice to a constant value if the current capacity of the slice in 2.5x the length
+func (s *SkippedSequenceSlice) clip(ctx context.Context) {
+	// threshold is 2.5x the current length of the slice
+	threshold := 2.5 * float64(len(s.list))
+
+	if cap(s.list) > int(threshold) {
+		// check if we can safely clip without an out of bound errors
+		if (len(s.list) + ClipCapacityHeadroom) < cap(s.list) {
+			base.DebugfCtx(ctx, base.KeyCache, "clipping skipped list capacity")
+			s.list = s.list[:len(s.list):ClipCapacityHeadroom]
+		}
+	}
+}
+
+// findSequence will use binary search to search the elements in the slice for a given sequence
+func (s *SkippedSequenceSlice) findSequence(x uint64) (int, bool) {
+	i, found := slices.BinarySearchFunc(s.list, x, func(a SkippedSequenceListInterface, seq uint64) int {
+		singleSeq, ok := a.(*SingleSkippedSequence)
+		if ok {
+			if singleSeq.seq > seq {
+				return 1
+			}
+			if singleSeq.seq == seq {
+				return 0
+			}
+			return -1
+		}
+		// should never get here as it stands, will have extra handling here pending CBG-3853
+		return 1
+	})
+	return i, found
+}
+
+// removeSeq will remove a given sequence from the slice if it exists
+func (s *SkippedSequenceSlice) removeSeq(x uint64) error {
+	s.lock.Lock()
+	defer s.lock.Unlock()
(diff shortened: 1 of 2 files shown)


db/skipped_sequence.go
```diff
@@ -0,0 +1,211 @@
...
+	"fmt"
+	"slices"
+	"sync"
+	"time"
+
+	"github.com/couchbase/sync_gateway/base"
+)
+
+const (
+	ClipCapacityHeadroom = 100
```
adamcfraser: For testing/tuning, do we want to define this as a property on SkippedSequenceSlice (and maybe change this constant to DefaultClipCapacityHeadroom).

db/skipped_sequence.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	// if we have incremented the index counter, we need to the index range from the slice
+	if indexToDelete > -1 {
+		s.list = slices.Delete(s.list, 0, indexToDelete+1)
+	}
+	// resize slice to reclaim memory if we need to
+	s.clip(ctx)
+}
+
+// clip will clip the capacity of the slice to a constant value if the current capacity of the slice in 2.5x the length
```
adamcfraser: I think this comment could clarify that the 'constant value' for the updated capacity will be the current length plus the headroom.

db/skipped_sequence.go
adamcfraser: Also, this value seems quite small to me, since we care about this most when the size of the slice is very large.  Thinking of a case where the length of the slice grows to 2.5M, then shrinks to 1M: the new capacity will be 1000100.  If only 101 sequences are added to the list, go is going to increase the capacity to 2000202.  I think we want a bit more buffer than that to minimize churn when the overall trend in the slice size is shrinking, but there's still some load.

Was there a specific reason for this value?

db/skipped_sequence.go
```diff
@@ -0,0 +1,211 @@
...
+	"time"
+
+	"github.com/couchbase/sync_gateway/base"
+)
+
+const (
+	ClipCapacityHeadroom = 100
+)
+
+type SkippedSequenceListInterface interface {
```
adamcfraser:
```suggestion
type SkippedSequenceListEntryInterface interface {
```
The previous name gave the impression it was an interface for the list itself, not an entry on the list.  I'm on the fence about including 'Interface' in the name of an interface, but it's not unreasonable.

db/skipped_sequence.go
```diff
@@ -0,0 +1,211 @@
...
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	s.list = append(s.list[:index+1], s.list[index:]...)
+	s.list[index] = entry
+}
+
+// PushSkippedSequenceEntry will append a new skipped sequence entry to the end of the slice, if adding a contiguous
+// sequence function will expand the last entry of the slice to reflect this
+// nolint:staticcheck
+func (s *SkippedSequenceSlice) PushSkippedSequenceEntry(entry SkippedSequenceListInterface) {
```
adamcfraser: Until we support ranges, I think defining this as 
```suggestion
func (s *SkippedSequenceSlice) PushSkippedSequenceEntry(entry *SingleSkippedSequence) {
```
makes more sense.  Then we can make the decision in CBG-3853 about whether it makes sense to use this function when adding ranges, or if ranges should get a separate function like PushSkippedSequenceRange

db/skipped_sequence.go
```diff
@@ -0,0 +1,211 @@
...
+	defer s.lock.Unlock()
+
+	if len(s.list) == 0 {
+		s.list = append(s.list, entry)
+		return
+	}
+	// get index of last entry + last seq of entry
+	index := len(s.list) - 1
+	lastEntryLastSeq := s.list[index].getLastSeq()
+	if (lastEntryLastSeq + 1) == entry.getStartSeq() {
```
adamcfraser: If we're adding ranges in CBG-3853, I think we just leave all this out for now

db/skipped_sequence.go
```diff
@@ -0,0 +1,211 @@
...
+
+// removeSeq will remove a given sequence from the slice if it exists
+func (s *SkippedSequenceSlice) removeSeq(x uint64) error {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	index, found := s.findSequence(x)
+	if !found {
+		return fmt.Errorf("sequence %d not found in the skipped list", x)
+	}
+	// handle boarder cases where x is equal to start or end sequence on range (or just sequence for single entries)
```
adamcfraser:
```suggestion
	// handle border cases where x is equal to start or end sequence on range (or just sequence for single entries)
```

db/skipped_sequence.go
```diff
@@ -0,0 +1,211 @@
...
+	if !s.list[index].isRange() {
+		// if not a range, we can just remove the single entry from the slice
+		if s.list[index].getLastSeq() == x {
```
adamcfraser: Why do we need to check lastSeq again here?  Can findSequence return a value where lastSeq != x?

db/skipped_sequence.go
```diff
@@ -0,0 +1,211 @@
...
+	// more range handling here CBG-3853, temporarily error as we shouldn't get to this point at this time
+	return fmt.Errorf("entered range handling code")
+}
+
+// insert will insert element in middle of slice maintaining order of rest of slice
+func (s *SkippedSequenceSlice) insert(index int, entry SkippedSequenceListInterface) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	s.list = append(s.list[:index+1], s.list[index:]...)
+	s.list[index] = entry
```
adamcfraser, line 195: SliceTricks recommends something a little bit different for insert:
```
s = append(s, 0 /* use the zero value of the element type */)
copy(s[i+1:], s[i:])
s[i] = x
```
Is there an advantage to the approach you've got?

db/skipped_sequence.go
```diff
@@ -0,0 +1,211 @@
...
+		if (len(s.list) + ClipCapacityHeadroom) < cap(s.list) {
+			base.DebugfCtx(ctx, base.KeyCache, "clipping skipped list capacity")
+			s.list = s.list[:len(s.list):ClipCapacityHeadroom]
+		}
+	}
+}
+
+// findSequence will use binary search to search the elements in the slice for a given sequence
+func (s *SkippedSequenceSlice) findSequence(x uint64) (int, bool) {
+	i, found := slices.BinarySearchFunc(s.list, x, func(a SkippedSequenceListInterface, seq uint64) int {
```
adamcfraser: Is there any difference from a compiler perspective if the function is defined as a standalone function, as opposed to inline as you've got it?

db/skipped_sequence.go
```diff
@@ -0,0 +1,211 @@
...
+	getStartSeq() uint64
+	isRange() bool
+}
+
+// SkippedSequenceSlice stores the set of skipped sequences as an ordered slice of single skipped sequences
+// or skipped sequence ranges
+type SkippedSequenceSlice struct {
+	list []SkippedSequenceListInterface
+	lock sync.RWMutex
+}
```
adamcfraser, line 42: Can we implement the following methods (that are defined for SkippedSequenceList today) on SkippedSequenceSlice in this PR, so that we can ensure the implementation is valid for these?
  - getOldest
  - Contains
  - getOlderThan

adamcfraser (changes requested):

db/skipped_sequence.go
gregns1: Slight difference in the benchmarking. Below is my approach currently:
```
BenchmarkInsertSkippedItem
BenchmarkInsertSkippedItem/single_entries_large_slice
BenchmarkInsertSkippedItem/single_entries_large_slice-10         	  309714	      3615 ns/op	     108 B/op	       1 allocs/op
BenchmarkInsertSkippedItem/single_entries_small_slice
BenchmarkInsertSkippedItem/single_entries_small_slice-10         	 8871403	       136.3 ns/op	     109 B/op	       1 allocs/op
```
Then with the above suggested approach:
```
BenchmarkInsertSkippedItem
BenchmarkInsertSkippedItem/single_entries_large_slice
BenchmarkInsertSkippedItem/single_entries_large_slice-10         	  311587	      3610 ns/op	     123 B/op	       2 allocs/op
BenchmarkInsertSkippedItem/single_entries_small_slice
BenchmarkInsertSkippedItem/single_entries_small_slice-10         	 7984885	       140.0 ns/op	     114 B/op	       2 allocs/op
```
Seems that the SliceTricks approach has extra allocation per op and slightly more bytes per op used (but really not much difference).

db/skipped_sequence.go
gregns1: Updated the headroom to be configurable. Also changed the default to be 1000. My thinking behind this was after looking at a few customer environments (that didn't have issues with skipped sequences) and the list never seemed to grown more than 30-40 in any environment I looked at. So thought I should have this capacity headroom fairly low. But you raise a good point, if we get a big jump in slice length this doesn't give much headroom for further jumps. 
Another option is to have two headroom values, one for when the list isn't getting used much < around 100 items say, and one for when slice grows > 10,000 for example.

db/skipped_sequence.go
gregns1: Have moved the definition of the search func into it's own function. All seems fine with that approach 😄

db/skipped_sequence.go
gregns1: I have changed the IsSkipped method to be named Contains, implemented getOldest with a test. With regards to getOlderThan. My implementation of SkippedSequenceCompact will cover this, my thinking was when we integrate this into the cache we can just call SkippedSequenceCompact from CleanSkippedSequenceQueue. Although getOlderThan returns the sequences older than a threshold, these sequences are just used to actually go ahead and remove them from the list a little further down in CleanSkippedSequenceQueue (which we already do in SkippedSequenceCompact). So overall my thinking it would just be best to call SkippedSequenceCompact and have it return how many its removing for logging/stats purposes in CleanSkippedSequenceQueue. It will also simplify CleanSkippedSequenceQueue method.

I have updated the test `TestCompactSkippedList` to reflect this change.

db/skipped_sequence_test.go
```diff
@@ -0,0 +1,411 @@
...
+			if bm.largeSlice {
+				for i := 0; i < 10000; i++ {
+					skippedSlice.PushSkippedSequenceEntry(NewSingleSkippedSequenceEntry(uint64(i * 2)))
+				}
+			} else {
+				for i := 0; i < 100; i++ {
+					skippedSlice.PushSkippedSequenceEntry(NewSingleSkippedSequenceEntry(uint64(i * 2)))
+				}
+			}
+			sequenceNum := 40000
```
adamcfraser: Lines 246-256 look like setup for the benchmark that should be outside of b.Run, since what we're trying to benchmark here is the insert operation.

db/skipped_sequence.go
adamcfraser: I don't think the SliceTricks has an extra allocation, if it's implemented as:
```
	s.list = append(s.list, nil)
	copy(s.list[index+1:], s.list[index:])
	s.list[index] = entry
```
When running locally the two approaches seemed equivalent from a timing and allocation perspective.  I couldn't find anything definitive in the go docs/slice tricks about the relative benefits of:
```
	s= append(s, nil)
	copy(s[i+1:], s[i:])
versus
         s = append(s[:i+1], s.list[i:]...)
```
Now that I've spent this much time on it, though, I'm wondering whether we even expect a use case for insert, since this is an append-only slice?

db/skipped_sequence.go
adamcfraser: I agree that in most cases the list is small, but the memory associated with a minimum allocation of 1000 (or 10000) when the list is otherwise empty should be negligible, so I think a single value is fine.

db/skipped_sequence.go
gregns1: Ah I see where I went wrong now, thank you for putting me right! We need insert for when we have to split ranges when sequences arrive at the cache. So if we have range 10-20 and sequence 15 arrives, we split range into 10-14 and insert new range 16-20. I was thinking this is needed for single entries too as if we have range 10-20 and 19 arrives we can alter range to 10-18 but then replace the range and index+1 with a single entry for sequence 20.

db/skipped_sequence_test.go
gregns1: Have moved the benchmark set ups into their own functions 😄

adamcfraser (approved):

PR #6756, opened 2024-04-05
CBG-3835 catch error in case there is a valid json but not a JSON object

torcolvin:
The code has now been reverted to the state before https://github.com/couchbase/sync_gateway/commit/2bde1ea79a5591e0a15f8fa6d93b45f13ff01bee with two changes:

log debug for event.DataType == base.MemcachedDataTypeRaw, Info logging in 3.1.3/3.1.4 but no logging before
increment ImportErrorCount in the case of valid _sync xattr and valid json but not a json object
Wrote tests for the expected behavior, which is to silently ignore documents with invalid _sync metadata, where it is valid JSON but not a valid JSON object.

## Pre-review checklist

- [] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2380/

### DIFF
--- a/db/document.go
+++ b/db/document.go
@@ -419,7 +419,7 @@ func UnmarshalDocumentSyncData(data []byte, needHistory bool) (*SyncData, error)
 		root.SyncData = &SyncData{History: make(RevTree)}
 	}
 	if err := base.JSONUnmarshal(data, &root); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("Could not unmarshal _sync out of document body: %w", err)
 	}
 	if root.SyncData != nil && root.SyncData.Deleted_OLD {
 		root.SyncData.Deleted_OLD = false
@@ -454,7 +454,7 @@ func UnmarshalDocumentSyncDataFromFeed(data []byte, dataType uint8, userXattrKey
 			}
 			err = base.JSONUnmarshal(syncXattr, result)
 			if err != nil {
-				return nil, nil, nil, nil, err
+				return nil, nil, nil, nil, fmt.Errorf("Found _sync xattr (%q), but could not unmarshal: %w", string(syncXattr), err)
 			}
 			return result, body, syncXattr, rawUserXattr, nil
 		}
--- a/db/import_listener.go
+++ b/db/import_listener.go
@@ -158,7 +158,7 @@ func (il *importListener) ProcessFeedEvent(event sgbucket.FeedEvent) (shouldPers
 
 	// If this is a binary document we can ignore, but update checkpoint to avoid reprocessing upon restart
 	if event.DataType == base.MemcachedDataTypeRaw {
-		base.InfofCtx(ctx, base.KeyImport, "Ignoring binary mutation event for %s.", base.UD(docID))
+		base.DebugfCtx(ctx, base.KeyImport, "Ignoring binary mutation event for %s.", base.UD(docID))
 		return true
 	}
 
@@ -171,10 +171,10 @@ func (il *importListener) ImportFeedEvent(ctx context.Context, collection *Datab
 	if err != nil {
 		if err == base.ErrEmptyMetadata {
 			base.WarnfCtx(ctx, "Unexpected empty metadata when processing feed event.  docid: %s opcode: %v datatype:%v", base.UD(event.Key), event.Opcode, event.DataType)
-		} else {
-			base.WarnfCtx(ctx, "Found sync metadata, but unable to unmarshal for feed document %q.  Will not be imported.  Error: %v", base.UD(event.Key), err)
+			il.importStats.ImportErrorCount.Add(1)
+			return
 		}
-		il.importStats.ImportErrorCount.Add(1)
+		base.DebugfCtx(ctx, base.KeyImport, "Found sync metadata, but unable to unmarshal for feed document %q. Will not be imported. Error: %v", base.UD(event.Key), err)
 		return
 	}
 
--- a/db/import_listener.go
+++ b/db/import_listener.go
@@ -174,7 +174,7 @@ func (il *importListener) ImportFeedEvent(ctx context.Context, collection *Datab
 			il.importStats.ImportErrorCount.Add(1)
 			return
 		}
-		base.DebugfCtx(ctx, base.KeyImport, "%s not able to be imported. Error: %v", base.UD(event.Key), err)
+		base.DebugfCtx(ctx, base.KeyImport, "%s will not be imported: %v", base.UD(event.Key), err)
 		return
 	}
 
--- a/db/import_listener.go
+++ b/db/import_listener.go
@@ -174,7 +174,7 @@ func (il *importListener) ImportFeedEvent(ctx context.Context, collection *Datab
 			il.importStats.ImportErrorCount.Add(1)
 			return
 		}
-		base.DebugfCtx(ctx, base.KeyImport, "Found sync metadata, but unable to unmarshal for feed document %q. Will not be imported. Error: %v", base.UD(event.Key), err)
+		base.DebugfCtx(ctx, base.KeyImport, "%s not able to be imported. Error: %v", base.UD(event.Key), err)
 		return
 	}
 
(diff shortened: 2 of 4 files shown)


db/document.go
```diff
@@ -419,6 +420,10 @@ func UnmarshalDocumentSyncData(data []byte, needHistory bool) (*SyncData, error)
 		root.SyncData = &SyncData{History: make(RevTree)}
 	}
 	if err := base.JSONUnmarshal(data, &root); err != nil {
+		// note jsoniter validate will not validate number, string, boolean, array types
+		if json.Valid(data) {
```
adamcfraser: The overhead of json.Valid (another document scan) is relatively high for this function.  Under what situations will this be triggered?
1. Initial import of a standard bucket document (JSON object)?
2. Any attempt to import a non-JSON object?
3. Import of a standard bucket document with non-standard _sync property?

If it's triggered for either 1 or 2, I think the potential impact to import throughput is too high, as there are realistic deployments with high volumes of both of these.

adamcfraser: Also, it doesn't feel like this is the right place for us to be checking whether the document body is a JSON object.  (i.e. as part of the "backward compatibility check for legacy sync metadata in the document body").

adamcfraser (changes requested):
I have concerns about the overhead of the json.Valid call.  It also feels like these changes are too heavyweight when considering the original enhancement we're trying to make here (CBG-3332), particularly considering this would be a backport candidate.

db/document.go
torcolvin: I think it's not necessary or helpful so I've removed this. This code would be triggered in import if there were no sync xattr so it actually would happen quite frequently. In this case we are looking for inline sync data to possibly migrate, so we do a partial unmarshal of the document to find if we should do that.

Independently, this is probably an area that could be optimized because this means for most SDK-only writes we are unmarshaling the document _twice_, once to find just the `SyncData` and another time when we decide to import it (or not) later.

db/document.go
```diff
@@ -411,15 +411,15 @@ func unmarshalDocumentWithXattr(ctx context.Context, docid string, data []byte,
...
+// Unmarshals just a document's sync metadata from JSON data. Errors if there is no _sync attribute in the data.
 // (This is somewhat faster, if all you need is the sync data without the doc body.)
 func UnmarshalDocumentSyncData(data []byte, needHistory bool) (*SyncData, error) {
 	var root documentRoot
 	if needHistory {
 		root.SyncData = &SyncData{History: make(RevTree)}
 	}
 	if err := base.JSONUnmarshal(data, &root); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("%w: %w", base.ErrNoInlineSyncData, err)
```
adamcfraser: The error name here (and the comment describing the function) doesn't look correct to me.  It's only going to return error here if `data` is not a JSON object.   For cases where it's a JSON object but doesn't include _sync, or even if _sync is present and it's not SyncData, I don't think we'll return error here.

adamcfraser: We don't have those cases in the unit test, might want to add those and ensure they are being handled (and logged) as expected.

adamcfraser (changes requested):

db/document.go
torcolvin: I think this is still correct behavior demonstrated in the test, but let me know if this is otherwise. Since this unmarshals into `documentRoot`, you'll see a marshal error in the case of no `_sync` attribute.

The case that is ambiguously worried is where `_sync` is present but is not `SyncData` it will return `ErrNoInlineSyncData`, which is sort of misleading.

I added additional cases to at least regression test what happens if we have `_sync` xattr and also `_sync` inline (we ignore inline).

torcolvin:
The code has now been reverted to the state before https://github.com/couchbase/sync_gateway/commit/2bde1ea79a5591e0a15f8fa6d93b45f13ff01bee with two changes:

1. log debug for `event.DataType == base.MemcachedDataTypeRaw`, Info logging in 3.1.3/3.1.4 but no logging before
2. increment ImportErrorCount in the case of valid `_sync` xattr and valid json but not a json object

Wrote tests for the expected behavior, which is to silently ignore documents with invalid _sync metadata, where it is valid JSON but not a valid JSON object.

db/import_test.go
```diff
@@ -575,13 +575,124 @@ func TestImportInvalidMetadata(t *testing.T) {
...
 
-	// write a document with inline sync metadata that is unmarshalable, triggering an import error
-	// can't write a document with invalid sync metadata as an xattr, so rely on legacy behavior
-	_, err := bucket.GetSingleDataStore().Add("doc1", 0, `{"foo" : "bar", "_sync" : 1 }`)
+	// docs named so they will both be on vBucket 1 in both 64 and 1024 vbuckets
+	const (
+		doc1 = "bookstand"
+		doc2 = "chipchop"
+	)
+	// write a document with inline sync metadata that is unmarshalable. This document will be ignored.
```
adamcfraser:
```suggestion
	// write a document with inline sync metadata that is not unmarshalable into SyncData. This document will be ignored.
```

db/import_test.go
```diff
@@ -575,13 +575,124 @@ func TestImportInvalidMetadata(t *testing.T) {
...
 	require.Equal(t, int64(0), db.DbStats.SharedBucketImport().ImportCount.Value())
+	require.Equal(t, int64(0), db.DbStats.SharedBucketImport().ImportErrorCount.Value())
+
+	// docs named so they will both be on vBucket 1 in both 64 and 1024 vbuckets
+	const (
+		doc1 = "bookstand"
+		doc2 = "chipchop"
+	)
+
+	// this is silently ignored because the sync metadata is invalid
```
adamcfraser: Since 'silently ignored' is what we're trying to fix here, maybe change this comment to say it only logs as debug?

db/import_test.go
```diff
@@ -575,13 +575,124 @@ func TestImportInvalidMetadata(t *testing.T) {
...
+	_, err = bucket.GetSingleDataStore().WriteWithXattrs(ctx, doc2, 0, 0, []byte(`{"foo" : "bar"}`), map[string][]byte{base.SyncXattrName: []byte(`{}`)}, nil)
+	require.NoError(t, err)
+
+	base.RequireWaitForStat(t, func() int64 {
+		return db.DbStats.SharedBucketImport().ImportCount.Value()
+	}, 1)
+	require.Equal(t, int64(0), db.DbStats.SharedBucketImport().ImportErrorCount.Value())
+}
+
+func TestImportFeedNonJSONNewDoc(t *testing.T) {
```
adamcfraser, line 633: This test is logging:
```
Found sync metadata, but unable to unmarshal for feed document "<ud>bookstand</ud>". Will not be imported. Error: readObjectStart: expect { or n, but found 1, error found in #1 byte of ...|1|..., bigger context ...|1|...
```
which is incorrect, because it didn't find sync metadata.  I think we've decided we can't differentiate between this case and the ones above without additional overhead, so I think we need to change the error message to be more generic.

adamcfraser (changes requested):

db/import_test.go
torcolvin: Changed the error messages and wrote them inline with the tests.

db/import_listener.go
```diff
@@ -171,10 +171,10 @@ func (il *importListener) ImportFeedEvent(ctx context.Context, collection *Datab
 	if err != nil {
 		if err == base.ErrEmptyMetadata {
 			base.WarnfCtx(ctx, "Unexpected empty metadata when processing feed event.  docid: %s opcode: %v datatype:%v", base.UD(event.Key), event.Opcode, event.DataType)
-		} else {
-			base.WarnfCtx(ctx, "Found sync metadata, but unable to unmarshal for feed document %q.  Will not be imported.  Error: %v", base.UD(event.Key), err)
+			il.importStats.ImportErrorCount.Add(1)
+			return
 		}
-		il.importStats.ImportErrorCount.Add(1)
+		base.DebugfCtx(ctx, base.KeyImport, "%s not able to be imported. Error: %v", base.UD(event.Key), err)
```
adamcfraser:
```suggestion
		base.DebugfCtx(ctx, base.KeyImport, "%s will not be imported: %v", base.UD(event.Key), err)
```
I think we should tweak this to make it more clear this is intended behaviour, not an unexpected error.

adamcfraser (changes requested):
One minor tweak for the debug logging, otherwise looks good.

adamcfraser (approved):

PR #6764, opened 2024-04-10
CBG-3853: Implementation of range sequence entry for skipped sequence slice

gregns1:
CBG-3853

- Addition of the range entry handling of the skipped slice
- Addition of test cases for range entries in the slice 
- Careful testing of edge case scenarios (especially when it comes to removing sequences from the slice). Test comments describe these edge cases.
- Updates to benchmarks
- Changed compact method top return int64 instead for ease of adding this output to a stat
- Refactored binarySearchFunc for simplicity 
- Removed mutex from insert as its unnecessary (its always called form method already acquiring the mutex)

## Pre-review checklist

### DIFF
--- a/db/skipped_sequence.go
+++ b/db/skipped_sequence.go
@@ -185,7 +185,7 @@ func (s *SkippedSequenceSlice) SkippedSequenceCompact(ctx context.Context, maxWa
 	s.lock.Lock()
 	defer s.lock.Unlock()
 
-	timeNow := time.Now().Unix()
+	timeNow := getCurrentUnixTime()
 	indexToDelete := -1
 	for _, v := range s.list {
 		timeStamp := v.getTimestamp()
@@ -349,7 +349,7 @@ func (s *SkippedSequenceSlice) PushSkippedSequenceEntry(entry SkippedSequenceLis
 				s.list[index] = entry
 			} else {
 				endSeq := entry.getLastSeq()
-				s.list[index] = NewSkippedSequenceRangeEntry(lastEntryStartSeq, endSeq, time.Now().Unix())
+				s.list[index] = NewSkippedSequenceRangeEntry(lastEntryStartSeq, endSeq, getCurrentUnixTime())
 			}
 		}
 	} else {
@@ -367,4 +367,9 @@ func (s *SkippedSequenceSlice) getOldest() uint64 {
 	}
 	// grab fist element in slice and take the start seq of that range/single sequence
 	return s.list[0].getStartSeq()
+}
+
+// getCurrentUnixTime will return the current time in unix time format
+func getCurrentUnixTime() int64 {
+	return time.Now().Unix()
 }
--- a/db/skipped_sequence.go
+++ b/db/skipped_sequence.go
@@ -26,13 +26,13 @@ const (
 
 type SkippedSequenceListEntry interface {
 	getTimestamp() int64
-	setTimestamp(time int64)
 	getLastSeq() uint64
 	setStartSeq(seq uint64)
 	setLastSeq(seq uint64)
 	getStartSeq() uint64
 	isRange() bool
 	getNumSequencesInEntry() int64
+	setRangeForIncomingContiguousEntry(lastSeq uint64, timeStamp int64)
 }
 
 // SkippedSequenceSlice stores the set of skipped sequences as an ordered slice of single skipped sequences
@@ -68,31 +68,34 @@ func NewSkippedSequenceSlice(clipHeadroom int) *SkippedSequenceSlice {
 
 // NewSingleSkippedSequenceEntry returns a SingleSkippedSequence with the specified sequence and the current
 // time in unix time
-func NewSingleSkippedSequenceEntry(sequence uint64) *SingleSkippedSequence {
+func NewSingleSkippedSequenceEntry(sequence uint64, timeStamp int64) *SingleSkippedSequence {
+	// if no timestamp provided, take current unix time
+	if timeStamp == 0 {
+		timeStamp = time.Now().Unix()
+	}
 	return &SingleSkippedSequence{
 		seq:       sequence,
-		timestamp: time.Now().Unix(),
+		timestamp: timeStamp,
 	}
 }
 
 // NewSkippedSequenceRangeEntry returns a SkippedSequenceRange with the specified sequence range and the current
 // time in unix time
-func NewSkippedSequenceRangeEntry(start, end uint64) *SkippedSequenceRange {
+func NewSkippedSequenceRangeEntry(start, end uint64, timeStamp int64) *SkippedSequenceRange {
+	// if no timestamp provided, take current unix time
+	if timeStamp == 0 {
+		timeStamp = time.Now().Unix()
+	}
 	return &SkippedSequenceRange{
 		start:     start,
 		end:       end,
-		timestamp: time.Now().Unix(),
+		timestamp: timeStamp,
 	}
 }
 
 // getTimestamp returns the timestamp of the entry
 func (s *SingleSkippedSequence) getTimestamp() int64 {
 	return s.timestamp
-}
-
-// setTimestamp sets the timestamp of the skipped sequence list entry
-func (s *SingleSkippedSequence) setTimestamp(time int64) {
-	s.timestamp = time
 }
 
 // getLastSeq gets the last sequence in the range on the entry, for single items the sequence will be returned
@@ -126,14 +129,15 @@ func (s *SingleSkippedSequence) getNumSequencesInEntry() int64 {
 	return 1
 }
 
+// setRangeForIncomingContiguousEntry will set the range last seq to the incoming contiguous entry's last seq
+// + set the timestamp.
+func (s *SingleSkippedSequence) setRangeForIncomingContiguousEntry(lastSeq uint64, timeStamp int64) {
+	// no-op
+}
+
 // getTimestamp returns the timestamp of the entry
 func (s *SkippedSequenceRange) getTimestamp() int64 {
 	return s.timestamp
-}
-
-// setTimestamp sets the timestamp of the skipped sequence list entry
-func (s *SkippedSequenceRange) setTimestamp(time int64) {
-	s.timestamp = time
 }
 
 // getStartSeq gets the start sequence in the range on the entry, for single items the sequence will be returned
@@ -166,6 +170,13 @@ func (s *SkippedSequenceRange) setStartSeq(seq uint64) {
 func (s *SkippedSequenceRange) getNumSequencesInEntry() int64 {
 	numSequences := (s.end - s.start) + 1
 	return int64(numSequences)
+}
+
+// setRangeForIncomingContiguousEntry will set the range last seq to the incoming contiguous entry's last seq
+// + set the timestamp.
+func (s *SkippedSequenceRange) setRangeForIncomingContiguousEntry(lastSeq uint64, timeStamp int64) {
+	s.timestamp = timeStamp
+	s.setLastSeq(lastSeq)
 }
 
 // Contains returns true if a given sequence exists in the skipped sequence slice
@@ -230,16 +241,14 @@ func binarySearchFunc(a SkippedSequenceListEntry, seq uint64) int {
 	if !a.isRange() {
 		if a.getStartSeq() > seq {
 			return 1
-		}
-		if a.getStartSeq() == seq {
+		} else if a.getStartSeq() == seq {
 			return 0
 		}
 		return -1
 	} else {
 		if a.getStartSeq() > seq {
 			return 1
-		}
-		if a.getStartSeq() <= seq && a.getLastSeq() >= seq {
+		} else if a.getLastSeq() >= seq {
 			return 0
 		}
 		return -1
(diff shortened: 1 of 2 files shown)


db/skipped_sequence.go
```diff
@@ -26,11 +26,13 @@ const (
 
 type SkippedSequenceListEntry interface {
 	getTimestamp() int64
+	setTimestamp(time int64)
```
adamcfraser: For properties that are intended to be immutable (which I think timestamp is?), I'd prefer to set those properties in the constructor, instead of exposing a set method on the interface.  The latter approach has the potential for race conditions being introduced in future.

db/skipped_sequence.go
```diff
@@ -153,19 +225,25 @@ func (s *SkippedSequenceSlice) findSequence(x uint64) (int, bool) {
...
+		if a.getStartSeq() > seq {
 			return 1
 		}
-		if singleSeq.seq == seq {
+		if a.getStartSeq() == seq {
+			return 0
+		}
+		return -1
+	} else {
+		if a.getStartSeq() > seq {
```
adamcfraser: It looks like you could get away with one fewer comparison here, with something like:
```
if a.getStartSeq() > seq {
   return 1 
} else if a.getLastSeq() >= seq {
   return 0
}
return -1
```

db/skipped_sequence.go
```diff
@@ -181,34 +259,94 @@ func (s *SkippedSequenceSlice) removeSeq(x uint64) error {
 		// if not a range, we can just remove the single entry from the slice
 		s.list = slices.Delete(s.list, index, index+1)
 		return nil
+	} else {
+		// range entry handling
+		// if x-1 == range start seq we need to replace elem at index-1 with single entry, then alter range start seq to x+1
```
adamcfraser: This doesn't look like it covers the case where the range has two elements correctly.  It looks to me like we've got the following possibilities:
```
1. numSequencesInEntry == 1 
   a. Simple delete case.  (Not used today, but see comments below about reducing allocations)
2. numSequencesInEntry == 2
   a. startSeq = x.  Modify range to startSeq=endSeq (to reduce allocations/GC)
   b. endSeq = x. Modify range to startSeq=endSeq (to reduce allocations/GC)
3. numSequencesInEntry == 3
   a. startSeq = x. Modify range
   b. endSeq = x. Modify range
   c. x is middle entry. Modify range (startSeq=endSeq), insert single 
 4. numSequencesInEntry > 3
   a. startSeq = x.  Modify range endpoints
   b. endSeq = x. Modify range endpoints
   c. startSeq = x-1. Insert single, modify range
   d. endSeq = x+1. Modify range, insert single.
   e. startSeq + 1 < x < endSeq - 1. Modify range, insert range.
```
Given that, I think this could be made more readable by ordering the code as:
```
- check for len=1
- check for x=startSeq or x=endSeq (which covers len==2 as well)
- special handling for len=3 (startSeq<x<endSeq only)
- general handling
   - x=startSeq+1
   - x=endSeq - 1
   - general, non-border case startSeq+1<x<endSeq-1
```

db/skipped_sequence.go
```diff
@@ -181,34 +259,94 @@ func (s *SkippedSequenceSlice) removeSeq(x uint64) error {
...
+	if (lastEntryLastSeq + 1) == entry.getStartSeq() {
+		// adding contiguous sequence
+		if s.list[index].isRange() {
+			// set last seq in the range to the new arriving sequence
+			s.list[index].setLastSeq(entry.getLastSeq())
+		} else {
+			// take the sequence from the single entry and create a new range entry in its place
+			startSeq := s.list[index].getStartSeq()
+			endSeq := entry.getLastSeq()
+			s.list[index] = NewSkippedSequenceRangeEntry(startSeq, endSeq)
```
adamcfraser: In the case where the incoming `entry` is already a range, can we use that instead of allocating a new one?

db/skipped_sequence.go
```diff
@@ -181,34 +259,94 @@ func (s *SkippedSequenceSlice) removeSeq(x uint64) error {
...
 
 	if len(s.list) == 0 {
 		s.list = append(s.list, entry)
 		return
 	}
-	// adding contiguous sequence handling here, pending CBG-3853
-	s.list = append(s.list, entry)
+	// get index of last entry + last seq of entry
+	index := len(s.list) - 1
+	lastEntryLastSeq := s.list[index].getLastSeq()
```
adamcfraser, line 271: Have you thought about what we want to do with timestamps when the incoming entry is combined with the existing?  There are impacts to SkippedSequenceSlice eviction in either case, but I think using the incoming entry's timestamp is probably more correct on the whole.  Would be interested in your thoughts.

adamcfraser (changes requested):

db/skipped_sequence.go
gregns1: I agree upon reflection. I have added a new method to do this on the interface.

db/skipped_sequence.go
```diff
@@ -368,3 +368,8 @@ func (s *SkippedSequenceSlice) getOldest() uint64 {
 	// grab fist element in slice and take the start seq of that range/single sequence
 	return s.list[0].getStartSeq()
 }
+
+// getCurrentUnixTime will return the current time in unix time format
```
adamcfraser: I don't feel like getCurrentUnixTime() offers much improvement over just writing time.Now().Unix(), and the latter is more self-documenting.

db/skipped_sequence.go
```diff
@@ -58,10 +68,20 @@ func NewSkippedSequenceSlice(clipHeadroom int) *SkippedSequenceSlice {
 
 // NewSingleSkippedSequenceEntry returns a SingleSkippedSequence with the specified sequence and the current
 // time in unix time
-func NewSingleSkippedSequenceEntry(sequence uint64) *SingleSkippedSequence {
+func NewSingleSkippedSequenceEntry(sequence uint64, timeStamp int64) *SingleSkippedSequence {
```
adamcfraser: It looks like all non-test cases are using time.Now().Unix() for the timestamp.  It might be an appropriate place for two constructors, something like NewSingleSkippedSequenceEntry(seq) and NewSingleSkippedSequenceEntryAt(seq, timestamp), to unclutter the non-test usage?

db/skipped_sequence.go
```diff
@@ -90,12 +115,62 @@ func (s *SingleSkippedSequence) isRange() bool {
...
+// getNumSequencesInEntry returns the number of sequences an entry in skipped slice holds,
+// for single entries it will return just 1
+func (s *SkippedSequenceRange) getNumSequencesInEntry() int64 {
+	numSequences := (s.end - s.start) + 1
+	return int64(numSequences)
+}
+
+// setRangeForIncomingContiguousEntry will set the range last seq to the incoming contiguous entry's last seq
+// + set the timestamp.
+func (s *SkippedSequenceRange) setRangeForIncomingContiguousEntry(lastSeq uint64, timeStamp int64) {
```
adamcfraser: This naming doesn't quite describe the functionality for me.  What do you think about something simpler like 'extendRange'?

db/skipped_sequence.go
```diff
@@ -181,34 +260,105 @@ func (s *SkippedSequenceSlice) removeSeq(x uint64) error {
 		// if not a range, we can just remove the single entry from the slice
 		s.list = slices.Delete(s.list, index, index+1)
 		return nil
+	} else {
+		// range entry handling
+		rangeElem := s.list[index]
+		if rangeElem.getNumSequencesInEntry() == 1 {
```
adamcfraser: getNumSequencesInEntry is cheap but not free - maybe define numSequences := getNumSequencesInEntry() for use here and on line 282?

db/skipped_sequence_test.go
```diff
@@ -133,35 +228,146 @@ func TestRemoveSeqFromSkipped(t *testing.T) {
 			remove:       100,
 			errorRemove:  150,
 		},
+		{
+			name:                 "list_full_range_items",
+			inputList:            []uint64{5, 15, 25, 35, 45},
+			expectedStartSeqList: []uint64{5, 15, 25, 28, 35, 45},
+			expectedEndSeqList:   []uint64{10, 20, 26, 30, 40, 50},
```
adamcfraser: What do you think about making these test definitions a bit more readable, so that it's more immediately obvious what the input ranges are and what the expected results are?  In particular the fact that we're creating ranges of [input, input+5] isn't really obvious, and it's also hard to eyeball what the expected end ranges are.

I was thinking of something like:
```
input: [][]uint64{{5,10}, {15,20}, {25,30}, {35,40}, {45,50}}
expected: [][]uint64{{5,10}, {15,20}, {25,26}, {28,30}, {35,40}, {45,50}
```
With that approach you could use {25} for cases where we expect a single sequence, and {25,25} where we expect a range of length 1.  You could also include the range len=2 and range len=3 cases in this test a bit more easily.  

It might be personal preference, if you disagree you can ignore.

adamcfraser (commented):
Generally looks fine, a few minor suggestions to help readability and to ensure we have complete test coverage.

db/skipped_sequence_test.go
adamcfraser: However, want to ensure we've got coverage for all permutations of removals, including:
- length 1 range
- length 2 ranges
- length 3 ranges, for removal of first/second/third (I think test below just removes the middle element)

gregns1: I agree, have made changes to have input and expected be 2-D array of uint64[][]. 

Added test cases to the test `TestRemoveSeqFromSkipped` for length 1 range seq removal, and length 2 seq removal. Have extended test `TestRemoveSeqFromThreeSequenceRange` to include a test for removing the start/last seq on a three sequence range.

db/skipped_sequence.go
```diff
@@ -181,34 +279,106 @@ func (s *SkippedSequenceSlice) removeSeq(x uint64) error {
...
 	}
-	// more range handling here CBG-3853, temporarily error as we shouldn't get to this point at this time
-	return fmt.Errorf("entered range handling code")
+	return nil
 }
 
 // insert will insert element in middle of slice maintaining order of rest of slice
 func (s *SkippedSequenceSlice) insert(index int, entry SkippedSequenceListEntry) {
-	s.lock.Lock()
-	defer s.lock.Unlock()
```
bbrks, line 192: This change would've been clearer if the method name had a leading underscore prefix to signal the expectation that the caller is the one acquiring the lock.

Ditto for `clip` and `findSequence`

db/skipped_sequence.go
```diff
@@ -181,34 +279,106 @@ func (s *SkippedSequenceSlice) removeSeq(x uint64) error {
...
+
+		// if x == startSeq set startSeq+1, if x == lastSeq set lastSeq-1
+		if rangeElem.getStartSeq() == x {
+			rangeElem.setStartSeq(x + 1)
+			return nil
+		}
+		if rangeElem.getLastSeq() == x {
+			rangeElem.setLastSeq(x - 1)
+			return nil
+		}
```
bbrks: If `numSequences` is 2 for these cases, does that mean this range becomes a single sequence without changing the type?

Does that impact anything the next time it's run through code that does `isRange()`?

db/skipped_sequence.go
```diff
@@ -42,20 +44,37 @@ type SkippedSequenceSlice struct {
 }
 
 var _ SkippedSequenceListEntry = &SingleSkippedSequence{}
+var _ SkippedSequenceListEntry = &SkippedSequenceRange{}
```
bbrks: I'm curious what the overhead of the interface will be vs. a single concrete type that can handle single sequences and ranges. ( https://syslog.ravelin.com/go-interfaces-but-at-what-cost-961e0f58a07b )

Most of the difference in behaviour is already in `isRange` checks, and it'd save all of the no-op method implementations.

Having said that, this would be closer to a micro-optimisation than the planned refactor of skipped sequences, so I'm OK waiting for all of that work to be done first, since the refactor to change the interface to a single type wouldn't be too intrusive.

db/skipped_sequence.go
```diff
@@ -153,19 +247,23 @@ func (s *SkippedSequenceSlice) findSequence(x uint64) (int, bool) {
...
+	if !a.isRange() {
+		if a.getStartSeq() > seq {
 			return 1
+		} else if a.getStartSeq() == seq {
+			return 0
 		}
-		if singleSeq.seq == seq {
+		return -1
+	} else {
+		if a.getStartSeq() > seq {
```
bbrks: The same `a.getStartSeq() > seq` operation is being applied to both range and non-range versions before anything else, so could be simplified further (the `isRange ... else` can also be implied) 

```go
	if a.getStartSeq() > seq {
		return 1
	}

	if !a.isRange() {
		if a.getStartSeq() == seq {
			return 0
		}
		return -1
	}

	if a.getLastSeq() >= seq {
		return 0
	}
	return -1
```

bbrks (dismissed):

db/skipped_sequence.go
gregns1: So the original approach was if the above code lines makes a range of 1, we were going to delete this and insert a new single entry in its place. Adam suggested we could avoid extra allocations by just altering the range to be one sequence. It seems to work well! 
With regards to it running through `isRange()`, it would still return true for being a range entry but I think this has no functional impact. Where `isRange()` is used in binary search function, it will still find the sequence if the range is length 1. Also for `removeSeq` we have a check for 1 sequence (above this code) that will just delete the entry.  The only other way this is used is for adding new entries in `PushSkippedSequenceEntry` which I feel doesn't have functional impact.

db/skipped_sequence.go
adamcfraser: I agree with this idea, and am fine doing it in a later optimization.  However, it does make me wonder whether anything that's a no-op belongs on the interface.  (setStartSeq/setLastSeq/extendRange).  Looking at the code, this goes back to the fact that we're using isRange instead of a type switch.  Was this intentional for performance reasons?

adamcfraser:
Overall looks good, but I'm also wondering about the slightly non-standard use of interfaces here (with the isRange/no-ops).

db/skipped_sequence.go
```diff
@@ -24,89 +24,127 @@ const (
...
 
-var _ SkippedSequenceListEntry = &SingleSkippedSequence{}
-
-// SingleSkippedSequence contains a single skipped sequence value + unix timestamp of the time it's created
-type SingleSkippedSequence struct {
-	seq       uint64
-	timestamp int64
+// SkippedSequenceListEntry contains start + end sequence for a range of skipped sequences +
+// a timestamp at which the entry was created in unix format. If an entry is a singular skipped sequence the end
+// sequence will not be 0
```
adamcfraser:
```suggestion
// sequence will be 0
```

db/skipped_sequence.go
```diff
@@ -24,89 +24,127 @@ const (
...
+type SkippedSequenceListEntry struct {
+	start     uint64 // start sequence of a range
+	end       uint64 // end sequence of the range (0 if a singular skipped sequence)
```
adamcfraser, line 40: It feels a bit off that the following are functionally equivalent, but have different handling based on isRange().
- start=n, end=0
- start=n, end=n
Are there benefits to the first approach, versus just using the second one every time we need a single skipped sequence?  It seems like we could avoid isRange checks in a number of places - maybe everywhere?

adamcfraser (dismissed):

db/skipped_sequence.go
gregns1: Good point, think I tried to think of it in the same way as the previous design, will update to reflect your comment above.

torcolvin (approved):

torcolvin (approved):

PR #6782, opened 2024-04-16
CBG-3768 Update only metadata during import

adamcfraser:
During document import, perform an xattr-only update of the document.

CBG-3768

- [x] https://github.com/couchbaselabs/rosmar/pull/33

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2387/

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -886,7 +886,7 @@ func (db *DatabaseCollectionWithUser) Put(ctx context.Context, docid string, bod
 	}
 
 	allowImport := db.UseXattrs()
-	doc, newRevID, err = db.updateAndReturnDoc(ctx, newDoc.ID, allowImport, &expiry, nil, nil, func(doc *Document) (resultDoc *Document, resultAttachmentData AttachmentData, createNewRevIDSkipped bool, updatedExpiry *uint32, resultErr error) {
+	doc, newRevID, err = db.updateAndReturnDoc(ctx, newDoc.ID, allowImport, &expiry, nil, nil, false, func(doc *Document) (resultDoc *Document, resultAttachmentData AttachmentData, createNewRevIDSkipped bool, updatedExpiry *uint32, resultErr error) {
 		var isSgWrite bool
 		var crc32Match bool
 
@@ -1011,7 +1011,7 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithConflictResolution(ctx c
 	}
 
 	allowImport := db.UseXattrs()
-	doc, _, err = db.updateAndReturnDoc(ctx, newDoc.ID, allowImport, &newDoc.DocExpiry, nil, existingDoc, func(doc *Document) (resultDoc *Document, resultAttachmentData AttachmentData, createNewRevIDSkipped bool, updatedExpiry *uint32, resultErr error) {
+	doc, _, err = db.updateAndReturnDoc(ctx, newDoc.ID, allowImport, &newDoc.DocExpiry, nil, existingDoc, false, func(doc *Document) (resultDoc *Document, resultAttachmentData AttachmentData, createNewRevIDSkipped bool, updatedExpiry *uint32, resultErr error) {
 		// (Be careful: this block can be invoked multiple times if there are races!)
 
 		var isSgWrite bool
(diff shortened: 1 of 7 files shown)


db/import.go
```diff
@@ -147,7 +147,7 @@ func (db *DatabaseCollectionWithUser) importDoc(ctx context.Context, docid strin
 		existingDoc.Expiry = *expiry
 	}
 
-	docOut, _, err = db.updateAndReturnDoc(ctx, newDoc.ID, true, expiry, mutationOptions, existingDoc, func(doc *Document) (resultDocument *Document, resultAttachmentData AttachmentData, createNewRevIDSkipped bool, updatedExpiry *uint32, resultErr error) {
+	docOut, _, err = db.updateAndReturnDoc(ctx, newDoc.ID, true, expiry, mutationOptions, existingDoc, true, func(doc *Document) (resultDocument *Document, resultAttachmentData AttachmentData, createNewRevIDSkipped bool, updatedExpiry *uint32, resultErr error) {
```
torcolvin, line 150: I am about to make a change to do something similar but add `metadataOnly` as a return value to the callback function.

This code is wrong here because would mark a document that has inline sync metadata as a metadata only update.

adamcfraser: I didn't think the callback function was the right scope, because of the requirement for the additional interaction with documentUpdateFunc, and it felt like that obfuscated the decision a bit too much.  Metadata migration for inline sync data is performed in a separate kv op, so I don't think it's a consideration here.

rest/importtest/import_test.go
```diff
@@ -35,6 +35,45 @@ func assertExpiry(t testing.TB, expected uint32, actual uint32) {
 	assert.True(t, base.DiffUint32(expected, actual) < 10, fmt.Sprintf("Unexpected difference between expected: %v actual %v", expected, actual))
 }
 
+// TestImportFeed validates basic feed-based import
+func TestImportFeed(t *testing.T) {
+
+	base.SkipImportTestsIfNotEnabled(t)
+
+	rtConfig := rest.RestTesterConfig{
+		SyncFn: `function(doc, oldDoc) { channel(doc.channels) }`,
```
torcolvin: I assume you are defining a syncfn because it will undergo js sync function and not skip? I'm not sure if that's needed for this test, I'd document why this is.

Autoimport should be enabled by default as far as I can tell by the `RestTester`so you could use the default config.

rest/importtest/import_test.go
```diff
@@ -35,6 +35,45 @@ func assertExpiry(t testing.TB, expected uint32, actual uint32) {
...
+	dataStore := rt.GetSingleDataStore()
+
+	base.SetUpTestLogging(t, base.LevelDebug, base.KeyImport, base.KeyCRUD)
+
+	// Create doc via the SDK
+	mobileKey := t.Name()
+	mobileBody := make(map[string]interface{})
+	mobileBody["channels"] = "ABC"
+	_, err := dataStore.Add(mobileKey, 0, mobileBody)
+	assert.NoError(t, err, "Error writing SDK doc")
```
torcolvin, line 60:
```suggestion
	require.NoError(t, err, "Error writing SDK doc")
```

db/import.go
torcolvin: I now understand why this is different from metadata only updates and I'll have to test this when I implement a `_mou` xattr. I didn't understand that `importDoc` will call `migrateMetadata` and it will update the body from what was originally passed in.

db/crud.go
```diff
@@ -1899,7 +1899,8 @@ type updateAndReturnDocCallback func(*Document) (resultDoc *Document, resultAtta
 //  1. Receive the updated document body in the response
 //  2. Specify the existing document body/xattr/cas, to avoid initial retrieval of the doc in cases that the current contents are already known (e.g. import).
 //     On cas failure, the document will still be reloaded from the bucket as usual.
-func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, docid string, allowImport bool, expiry *uint32, opts *sgbucket.MutateInOptions, existingDoc *sgbucket.BucketDocument, callback updateAndReturnDocCallback) (doc *Document, newRevID string, err error) {
+//  3. If isImport=true, document body will not be updated - only metadata xattr(s)
+func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, docid string, allowImport bool, expiry *uint32, opts *sgbucket.MutateInOptions, existingDoc *sgbucket.BucketDocument, isImport bool, callback updateAndReturnDocCallback) (doc *Document, newRevID string, err error) {
```
torcolvin, line 1903: Take or leave this comment as maybe out of scope.

I noticed that `allowImport` and `isImport` are both args to this function and I dug deeper in the code to figure out why because these variables seem very similar and it was not clear why they would both need to be set and are different. As far as I can tell, `allowImport` could be removed because https://github.com/couchbase/sync_gateway/blob/c4ba903c5d3096088a418bf440ee2b73b8321938/db/crud.go#L1809 could say:

```
err = validateExistingDoc(doc, col.UseXattrs(), docExists)
```

and this would avoid the need to pass in `allowImport` at all.

torcolvin (changes requested):
I'll wait for an update as to why [TestXattrOnDemandImportPreservesExpiry](https://jenkins.sgwdev.com/job/SyncGateway-Integration/2385/testReport/junit/github/com_couchbase_sync_gateway_rest_importtest/TestXattrOnDemandImportPreservesExpiry/) is failing.

db/crud.go
adamcfraser: You're correct that today allowImport is equivalent to "is shared bucket access enabled", which could be checked by UseXattrs().  I don't think that's related to this PR, though, and is different than 'isImport', which mean 'is the current operation being performed an import'

rest/importtest/import_test.go
adamcfraser: No, this was just copy/paste of the setup code from another test.  I agree it can be simplified.

db/crud.go
torcolvin: Understood, I was commenting on the fact that these have very similar argument names and it was a bit confusing to me.

It is different than this PR, only related in that I look at the function signature and saw two bool arguments with import in the name and wondered how they were different.

rest/importtest/import_test.go
adamcfraser: It turns out autoimport is only enabled by default for EE, so still need the explicit setting here for the time being.

torcolvin (approved):

PR #6789, opened 2024-04-23
CBG-3854: Integration of the skipped sequence slice into the cache

gregns1:
CBG-3854

- Integrates new skipped sequence slice into the cache 
- Adds some additional stats as included in the design doc 
- Adds some test testing both functionality of the slice but also the new stats 
- Initial profiling locally shows a big improvement in terms of memory usage with a unit test that causes a jump in 80 million sequences. Not perfect given there are more areas to optimize yet but much better

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2388/

### DIFF
--- a/base/stats.go
+++ b/base/stats.go
@@ -421,7 +421,7 @@ type CacheStats struct {
 	NonMobileIgnoredCount *SgwIntStat `json:"non_mobile_ignored_count"`
 	// The total number of active channels.
 	NumActiveChannels *SgwIntStat `json:"num_active_channels"`
-	// The total number of skipped sequences.
+	// The total number of skipped sequences. This is a cumulative value.
 	NumSkippedSeqs *SgwIntStat `json:"num_skipped_seqs"`
 	// The total number of pending sequences. These are out-of-sequence entries waiting to be cached.
 	PendingSeqLen *SgwIntStat `json:"pending_seq_len"`
@@ -431,8 +431,12 @@ type CacheStats struct {
 	RevisionCacheHits *SgwIntStat `json:"rev_cache_hits"`
 	// The total number of revision cache misses.
 	RevisionCacheMisses *SgwIntStat `json:"rev_cache_misses"`
-	// The current length of the pending skipped sequence queue.
+	// The current length of the pending skipped sequence slice.
 	SkippedSeqLen *SgwIntStat `json:"skipped_seq_len"`
+	// The current capacity of the skipped sequence slice
+	SkippedSeqCap *SgwIntStat `json:"skipped_seq_cap"`
+	// The number of sequences currently in the skipped sequence slice
+	NumCurrentSeqsSkipped *SgwIntStat `json:"num_seqs_skipped"`
 	// The total view_queries.
 	ViewQueries *SgwIntStat `json:"view_queries"`
 }
@@ -1327,6 +1331,14 @@ func (d *DbStats) initCacheStats() error {
 	if err != nil {
 		return err
 	}
+	resUtil.SkippedSeqCap, err = NewIntStat(SubsystemCacheKey, "skipped_seq_cap", StatUnitNoUnits, SkippedSeqCapDesc, StatAddedVersion3dot2dot0, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, labelKeys, labelVals, prometheus.GaugeValue, 0)
+	if err != nil {
+		return err
+	}
+	resUtil.NumCurrentSeqsSkipped, err = NewIntStat(SubsystemCacheKey, "num_seqs_skipped", StatUnitNoUnits, NumCurrentSkippedSeq, StatAddedVersion3dot2dot0, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, labelKeys, labelVals, prometheus.CounterValue, 0)
+	if err != nil {
+		return err
+	}
 	resUtil.ViewQueries, err = NewIntStat(SubsystemCacheKey, "view_queries", StatUnitNoUnits, ViewQueriesDesc, StatAddedVersion3dot0dot0, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, labelKeys, labelVals, prometheus.CounterValue, 0)
 	if err != nil {
 		return err
@@ -1357,6 +1369,8 @@ func (d *DbStats) unregisterCacheStats() {
 	prometheus.Unregister(d.CacheStats.NonMobileIgnoredCount)
 	prometheus.Unregister(d.CacheStats.NumActiveChannels)
 	prometheus.Unregister(d.CacheStats.NumSkippedSeqs)
+	prometheus.Unregister(d.CacheStats.SkippedSeqCap)
+	prometheus.Unregister(d.CacheStats.NumCurrentSeqsSkipped)
 	prometheus.Unregister(d.CacheStats.PendingSeqLen)
 	prometheus.Unregister(d.CacheStats.RevisionCacheBypass)
 	prometheus.Unregister(d.CacheStats.RevisionCacheHits)
--- a/base/stats.go
+++ b/base/stats.go
@@ -422,7 +422,7 @@ type CacheStats struct {
 	// The total number of active channels.
 	NumActiveChannels *SgwIntStat `json:"num_active_channels"`
 	// The total number of skipped sequences. This is a cumulative value.
-	NumSkippedSeqs *SgwIntStat `json:"num_skipped_seqs"`
+	NumSkippedSeqs *SgwIntStat `json:"cumulative_num_skipped_seqs"`
 	// The total number of pending sequences. These are out-of-sequence entries waiting to be cached.
 	PendingSeqLen *SgwIntStat `json:"pending_seq_len"`
 	// The total number of revision cache bypass operations performed.
@@ -436,7 +436,7 @@ type CacheStats struct {
 	// The current capacity of the skipped sequence slice
 	SkippedSeqCap *SgwIntStat `json:"skipped_seq_cap"`
 	// The number of sequences currently in the skipped sequence slice
-	NumCurrentSeqsSkipped *SgwIntStat `json:"num_seqs_skipped"`
+	NumCurrentSeqsSkipped *SgwIntStat `json:"current_skipped_seq_count"`
 	// The total view_queries.
 	ViewQueries *SgwIntStat `json:"view_queries"`
 }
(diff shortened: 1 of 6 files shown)


base/stats.go
```diff
@@ -431,8 +431,12 @@ type CacheStats struct {
 	RevisionCacheHits *SgwIntStat `json:"rev_cache_hits"`
 	// The total number of revision cache misses.
 	RevisionCacheMisses *SgwIntStat `json:"rev_cache_misses"`
-	// The current length of the pending skipped sequence queue.
+	// The current length of the pending skipped sequence slice.
 	SkippedSeqLen *SgwIntStat `json:"skipped_seq_len"`
+	// The current capacity of the skipped sequence slice
+	SkippedSeqCap *SgwIntStat `json:"skipped_seq_cap"`
+	// The number of sequences currently in the skipped sequence slice
+	NumCurrentSeqsSkipped *SgwIntStat `json:"num_seqs_skipped"`
```
adamcfraser: I think the naming similarities between the existing `num_skipped_seqs` and `num_seqs_skipped` will just cause problems/confusion.  Maybe `skipped_seq_count`, to make the naming consistent with the other "current state, non-cumulative" stats?

db/change_cache.go
```diff
@@ -300,23 +294,26 @@ func (c *changeCache) InsertPendingEntries(ctx context.Context) error {
...
 
-	oldSkippedSequences := c.GetSkippedSequencesOlderThanMaxWait()
-	if len(oldSkippedSequences) == 0 {
+	base.InfofCtx(ctx, base.KeyCache, "Starting CleanSkippedSequenceQueue for database %s", base.MD(c.db.Name))
+
+	maxWait := int64(c.options.CacheSkippedSeqMaxWait.Seconds())
+	if maxWait < 1 {
+		maxWait = 1
+	}
+	numOldSkippedSequences := c.skippedSeqs.SkippedSequenceCompact(ctx, maxWait)
```
adamcfraser:
```suggestion
	compactedSequences := c.skippedSeqs.SkippedSequenceCompact(ctx, maxWait)
```
I think numCompactedSequences or maybe numEvictedSequences would be a clearer name here.

db/change_cache.go
```diff
@@ -878,33 +874,32 @@ func (h *LogPriorityQueue) Pop() interface{} {
 // ////// SKIPPED SEQUENCE QUEUE
 
 func (c *changeCache) RemoveSkipped(x uint64) error {
-	err := c.skippedSeqs.Remove(x)
-	c.db.DbStats.Cache().SkippedSeqLen.Set(int64(c.skippedSeqs.skippedList.Len()))
+	err := c.skippedSeqs.removeSeq(x)
+	c.db.DbStats.Cache().SkippedSeqLen.Set(int64(len(c.skippedSeqs.list)))
+	c.db.DbStats.Cache().SkippedSeqCap.Set(int64(cap(c.skippedSeqs.list)))
+	if err == nil {
+		c.db.DbStats.Cache().NumCurrentSeqsSkipped.Add(-1)
```
adamcfraser: It feels like c.skippedSeqs should keep track of the current value of NumCurrentSeqsSkipped internally, instead of having the changeCache try to keep track of it.  What do you think?

db/change_cache.go
```diff
@@ -878,33 +874,32 @@ func (h *LogPriorityQueue) Pop() interface{} {
...
-}
-
-func (c *changeCache) GetSkippedSequencesOlderThanMaxWait() (oldSequences []uint64) {
-	return c.skippedSeqs.getOlderThan(c.options.CacheSkippedSeqMaxWait)
+func (c *changeCache) PushSkipped(ctx context.Context, startSeq uint64, endSeq uint64) {
+	newEntry := NewSkippedSequenceRangeEntry(startSeq, endSeq)
+	numSequences := newEntry.getNumSequencesInEntry()
+	c.skippedSeqs.PushSkippedSequenceEntry(newEntry)
+	c.db.DbStats.Cache().SkippedSeqLen.Set(int64(len(c.skippedSeqs.list)))
+	c.db.DbStats.Cache().SkippedSeqCap.Set(int64(cap(c.skippedSeqs.list)))
```
adamcfraser: Updating four stats every time we PushSkipped feels like potentially a lot of overhead since they are all atomic writes.  This is particularly true for SkippedSeqCap which we expect to change infrequently, but even for the others I don't know if we need this kind of granularity.  

One option would be to only update these stats in changeCache.updateStats (which runs once per minute).  That would be fine for sgcollect/stats.log - the question is whether we'd expect a customer to want finer granularity than that, and what the performance overhead would be of updating all these in real time.

adamcfraser (changes requested):
A few naming nits, but also a question about perf overhead of stats computation per skipped sequence, and whether they should move to changeCache.updateStats.

db/change_cache.go
gregns1: Completely forgot `changeCache.updateStats` was a thing. Updated to use this!

db/change_cache.go
gregns1: Have added this and the cumulative value to the skipped sequence struct to be tracked there. I have had to make the NumCurrentSeqsSkipped value on skipped sequences struct an atomic value though due to the concurrency around sequences being added/removed from the skipped list and the background task CleanSkippedSequenceQueue running. I was on fence about having it be an atomic value handled inside the skipped sequence slice code, then have that written to the stat in `updateStats()` or just having the value written direct inside the change cache each time a sequence is added/removed? On one hand it feels tidier and better to handle it in skipped sequence code (like the current state of the PR) but it feels like we're adding an extra atomic operation to update the stat of this value.

gregns1: Looking at this with slightly fresher eyes after our meetings, I think I have solution to avoid the race between `updateStats()` and the skipped sequence code writing to them. I will push up the changes now and let me know what you think 😄

base/stats.go
```diff
@@ -421,8 +421,8 @@ type CacheStats struct {
 	NonMobileIgnoredCount *SgwIntStat `json:"non_mobile_ignored_count"`
 	// The total number of active channels.
 	NumActiveChannels *SgwIntStat `json:"num_active_channels"`
-	// The total number of skipped sequences.
-	NumSkippedSeqs *SgwIntStat `json:"num_skipped_seqs"`
+	// The total number of skipped sequences. This is a cumulative value.
+	NumSkippedSeqs *SgwIntStat `json:"cumulative_num_skipped_seqs"`
```
adamcfraser: We want to avoid renaming existing stats, so we don't break anything users have built on the existing names.

db/change_cache.go
```diff
@@ -90,6 +89,10 @@ func (c *changeCache) updateStats(ctx context.Context) {
 	c.db.DbStats.Cache().PendingSeqLen.Set(int64(c.internalStats.pendingSeqLen))
 	c.db.DbStats.CBLReplicationPull().MaxPending.SetIfMax(int64(c.internalStats.maxPending))
 	c.db.DbStats.Cache().HighSeqStable.Set(int64(c._getMaxStableCached(ctx)))
+	c.db.DbStats.Cache().SkippedSeqLen.Set(c.skippedSeqs.getSliceLength())
```
adamcfraser: Is there something non-standard going on with the whitespace here?  (spaces instead of tab, maybe?)

db/change_cache.go
```diff
@@ -300,23 +298,24 @@ func (c *changeCache) InsertPendingEntries(ctx context.Context) error {
...
-// and subsequent removal (RemoveSkipped).
+// than CacheSkippedSeqMaxWait from the slice.
 func (c *changeCache) CleanSkippedSequenceQueue(ctx context.Context) error {
 
-	oldSkippedSequences := c.GetSkippedSequencesOlderThanMaxWait()
-	if len(oldSkippedSequences) == 0 {
+	base.InfofCtx(ctx, base.KeyCache, "Starting CleanSkippedSequenceQueue for database %s", base.MD(c.db.Name))
+
+	maxWait := int64(c.options.CacheSkippedSeqMaxWait.Seconds())
+	if maxWait < 1 {
```
adamcfraser: This feels like the wrong place for this check - we'd want to do it (once) when CacheSkippedSeqMaxWait is set, instead of every time we trigger Clean.  Did you have a test scenario that required this check?

db/change_cache.go
```diff
@@ -787,9 +786,8 @@ func (c *changeCache) _addPendingLogs(ctx context.Context) channels.Set {
 			heap.Pop(&c.pendingLogs)
 			changedChannels = changedChannels.UpdateWithSlice(c._addToCache(ctx, change))
 		} else if len(c.pendingLogs) > c.options.CachePendingSeqMaxNum || time.Since(c.pendingLogs[0].TimeReceived) >= c.options.CachePendingSeqMaxWait {
-			c.db.DbStats.Cache().NumSkippedSeqs.Add(1)
-			c.PushSkipped(ctx, c.nextSequence)
-			c.nextSequence++
+			c.PushSkipped(ctx, c.nextSequence, change.Sequence-1)
```
adamcfraser: This is a nice change (to push everything from nextSequence to the first pending sequence at once, instead of in multiple iterations of the loop).  It took me a minute to work out what exactly this was doing.  What do you think about the following, for readability:
- rename 'change' to 'oldestPending'
- Add a comment above line 789 like `//  Skip all sequences up to the oldest Pending`

db/skipped_sequence.go
```diff
@@ -289,3 +301,31 @@ func (s *SkippedSequenceSlice) getOldest() uint64 {
 	// grab fist element in slice and take the start seq of that range/single sequence
 	return s.list[0].getStartSeq()
 }
+
+// getSliceLength retrieves the current skipped sequence slice length
```
adamcfraser: Since we only have a single caller of these functions (updateStats), I think it would make sense to define a SkippedSequenceSlice.getStats() function that returned all four of these in one lock.

adamcfraser (changes requested):

db/change_cache.go
gregns1: Upon reading this again I don't think its needed. If we end up with a customer setting that low of an interval it will just compact pretty much all entries as the maxWait will be 0. I don't think that is too much different to today if the customer was to set that setting that low.

db/change_cache.go
gregns1: I don't think so? I have updated to reflect the changes to have a getStats() call for the skipped slice. I think the spacing in Github could be weird, it looks fine in goland to me.

adamcfraser (approved):

PR #6790, opened 2024-04-24
CBG-3905 scope message about searching for database config

torcolvin:
On startup, this will log as follows, which is confusing at best for what it is doing.

```
2024-04-19T17:57:22.896+02:00 [INF] Config: GetDatabaseConfigs starting (attempt 1/5)
2024-04-19T17:57:32.894+02:00 [INF] Config: GetDatabaseConfigs starting (attempt 1/5)
2024-04-19T17:57:32.896+02:00 [INF] Config: GetDatabaseConfigs starting (attempt 1/5)
2024-04-19T17:57:32.897+02:00 [INF] Config: GetDatabaseConfigs starting (attempt 1/5)
```

I'm not sure if INFO logging is already too aggressive here, this was added in https://github.com/couchbase/sync_gateway/pull/6720 and will log every 10s.

Change the code to log the bucket name _everywhere_ in this function by context and drop the logging for this message to trace if this is the first attempt, and info on attempt > 1. I think warn is too much for logging attempts 2/5, because on a simultaneous update to the config the bootstrap config could fail, and we will log warning after attempt 5 fails.

I think the best thing a reviewer could do is to run Sync Gateway with and without some buckets and see if the logging verbosity makes sense.

### DIFF
--- a/rest/config_manager.go
+++ b/rest/config_manager.go
@@ -345,7 +345,12 @@ func (b *bootstrapContext) GetDatabaseConfigs(ctx context.Context, bucketName, g
 
 	var attempt int
 	for attempt = 1; attempt <= configFetchMaxRetryAttempts; attempt++ {
-		base.InfofCtx(ctx, base.KeyConfig, "Checking for database config (attempt %d/%d)", attempt, configFetchMaxRetryAttempts)
+		msg := fmt.Sprintf("Checking for database config (attempt %d/%d)", attempt, configFetchMaxRetryAttempts)
+		if attempt == 1 {
+			base.TracefCtx(ctx, base.KeyConfig, msg)
+		} else {
+			base.InfofCtx(ctx, base.KeyConfig, msg)
+		}
 		registry, err := b.getGatewayRegistry(ctx, bucketName)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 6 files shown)


rest/config_manager.go
```diff
@@ -345,7 +345,12 @@ func (b *bootstrapContext) GetDatabaseConfigs(ctx context.Context, bucketName, g
 
 	var attempt int
 	for attempt = 1; attempt <= configFetchMaxRetryAttempts; attempt++ {
-		base.InfofCtx(ctx, base.KeyConfig, "GetDatabaseConfigs starting (attempt %d/%d)", attempt, configFetchMaxRetryAttempts)
+		msg := fmt.Sprintf("Checking for database config (attempt %d/%d)", attempt, configFetchMaxRetryAttempts)
+		if attempt == 1 {
+			base.TracefCtx(ctx, base.KeyConfig, msg)
```
bbrks: I think this might be better at debug.

We know it only happens once (per bucket), so isn't super verbose, and it's useful when finding out what buckets were being looked in under normal circumstances (no retries) - that can diagnose things like incorrect RBAC permissions limiting bucket visbility.

```suggestion
			base.DebugfCtx(ctx, base.KeyConfig, msg)
```

bbrks (changes requested):
I'm good with these changes but would like to see the trace log moved to debug.

bbrks (approved):

PR #6791, opened 2024-04-25
CBG-3895 allow xattrs to be deleted at the same time as they are updated

torcolvin:
https://github.com/couchbase/sg-bucket/pull/120
https://github.com/couchbaselabs/rosmar/pull/34

Changed APIs to support xattrs that are being upserted to include xattrs to be deleted. See sg-bucket API.

- validate before gocb that we are never deleting that we are never deleting an xattr on an insert (this will fail on CBS with invalid flags because StoreSemantic is Insert). The failure mode was confusing enough that I thought it would be good to match rosmar handling.
- validate that deleted xattrs and updated xattrs are distinct sets

Update StoreSemantic which can only be Replace if xattrs are being deleted, it has to be Insert if the xattrs are being updated on a tombstone.

## Pre-review checklist

- [x] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2499/ (CBS 7.6.1)
- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2398/ (CBS 7.2)


base/collection_xattr.go
```diff
@@ -189,12 +190,12 @@ func (c *Collection) SubdocWrite(ctx context.Context, k string, subdocKey string
 }
 
 // subdocGetBodyAndXattr retrieves the document body and xattrs in a single LookupIn subdoc operation.  Does not require both to exist.
-func (c *Collection) subdocGetBodyAndXattrs(ctx context.Context, k string, xattrKeys []string, fetchBody bool) (rawBody []byte, xattrs map[string][]byte, cas uint64, err error) {
+func (c *Collection) subdocGetBodyAndXattrs(ctx context.Context, k string, xattrKeys []string, fetchBody bool) (isTombstone bool, rawBody []byte, xattrs map[string][]byte, cas uint64, err error) {
```
adamcfraser, line 193: What's the scenario where isTombstone is required?  (i.e. where the combination of fetchBody=true, rawBody=nil, and err=nil isn't sufficient for the caller to determine isTombstone).  
One concern I have is that returning isTombstone=false may be misleading if it's actually not being determined (because fetchBody=false).

base/collection_xattr.go
```diff
@@ -360,9 +362,31 @@ func (c *Collection) insertBodyAndXattrs(_ context.Context, k string, exp uint32
...
+		StoreSemantic: gocb.StoreSemanticsInsert,
+	}
+	result, mutateErr := c.Collection.MutateIn(k, mutateOps, options)
+	if mutateErr != nil {
+		return 0, mutateErr
+	}
+	return uint64(result.Cas()), nil
+}
+
+// ressurectWithBodyAndXattrs inserts a document and associated xattrs in a single mutateIn operation.
```
adamcfraser:
```suggestion
// resurrectWithBodyAndXattrs inserts a document and associated xattrs in a single mutateIn operation.
```

adamcfraser: The resurrectWithBodyAndXattrs code ended up identical to insertBodyAndXattrs, unless I'm missing something.  Can we remove it?

base/collection_xattr.go
```diff
@@ -428,68 +452,105 @@ func (c *Collection) SubdocSetXattrs(k string, xvs map[string][]byte) (casOut ui
...
 		Cas:           gocb.Cas(cas),
 	}
+
 	options.Internal.DocFlags = gocb.SubdocDocFlagAccessDeleted
 	fillMutateInOptions(ctx, options, opts)
 
 	result, mutateErr := c.Collection.MutateIn(k, mutateOps, options)
 	if mutateErr != nil {
+		if isKVError(mutateErr, memd.StatusSubDocBadMulti) {
+			mutateErr = fmt.Errorf("%w: %w", ErrXattrNotFound, mutateErr)
```
adamcfraser: Is ErrXattrNotFound actually the only case we get StatusSubDocBadMulti? 

Does the 'xattr not found' only apply to attempted deletes, or does it also include xattrs we're updating?  (i.e. does this function support inserting new xattrs?)

base/collection_xattr.go
```diff
@@ -428,68 +452,105 @@ func (c *Collection) SubdocSetXattrs(k string, xvs map[string][]byte) (casOut ui
...
+		mutateOps = append(mutateOps, gocb.RemoveSpec(xattrKey, RemoveSpecXattr))
+
 	}
 	mutateOps = append(mutateOps, gocb.ReplaceSpec("", bytesToRawMessage(v), nil))
 	mutateOps = appendMacroExpansions(mutateOps, opts)
 
 	options := &gocb.MutateInOptions{
 		Expiry:        CbsExpiryToDuration(exp),
-		StoreSemantic: gocb.StoreSemanticsUpsert,
+		StoreSemantic: gocb.StoreSemanticsReplace,
```
adamcfraser, line 496: Same question as above - do the xattrs have to already exist to use StoreSemanticsReplace?

base/collection_xattr.go
```diff
@@ -428,68 +452,105 @@ func (c *Collection) SubdocSetXattrs(k string, xvs map[string][]byte) (casOut ui
...
 		Cas:           gocb.Cas(cas),
 	}
 	fillMutateInOptions(ctx, options, opts)
 	result, mutateErr := c.Collection.MutateIn(k, mutateOps, options)
 	if mutateErr != nil {
+		if errors.Is(mutateErr, gocb.ErrDocumentNotFound) {
+			casMismatchErr := sgbucket.CasMismatchErr{Actual: 0, Expected: cas}
```
adamcfraser: Converting a 'not found' to a 'cas mismatch' seems like it's hiding some information from callers.  What's the reason to do this conversion here?

base/collection_xattr.go
```diff
@@ -428,68 +452,105 @@ func (c *Collection) SubdocSetXattrs(k string, xvs map[string][]byte) (casOut ui
...
-	mutateOps := make([]gocb.MutateInSpec, 0, len(xattrs)+1)
-	for xattrKey, xattrVal := range xattrs {
-		mutateOps = append(mutateOps, gocb.UpsertSpec(xattrKey, bytesToRawMessage(xattrVal), UpsertSpecXattr))
+	mutateOps, err := getUpsertSpecs(xattrs)
+	if err != nil {
+		return 0, err
+	}
+	if cas == 0 && len(xattrsToDelete) > 0 {
+		return 0, sgbucket.ErrDeleteXattrOnDocumentInsert
+	}
```
adamcfraser, line 518: We don't do this check in other methods that attempt to delete xattrs (like updateXattrs).  Is it needed there also?

base/collection_xattr_common.go
```diff
@@ -204,21 +239,33 @@ func (c *Collection) WriteUpdateWithXattrs(ctx context.Context, k string, xattrK
...
+				casOut, writeErr = c.WriteWithXattrs(ctx, k, exp, cas, updatedDoc.Doc, updatedDoc.Xattrs, updatedDoc.XattrsToDelete, opts)
+			}
 		}
 
 		if writeErr == nil {
 			return casOut, nil
 		}
 
-		if IsCasMismatch(writeErr) {
+		if previousLoopCas != nil && *previousLoopCas == cas {
```
adamcfraser: What scenario has the potential to cause this loop, or is this strictly defensive?

base/collection_xattr_common.go
```diff
@@ -204,21 +239,33 @@ func (c *Collection) WriteUpdateWithXattrs(ctx context.Context, k string, xattrK
...
+			err := RedactErrorf("Stopping an infinite loop trying to update doc with xattr for key=%s, xattrKeys=%s", UD(k), UD(xattrKeys))
```
adamcfraser: I think this error would be better written as something like "CAS retry triggered, but no change in document CAS detected"

adamcfraser (changes requested):
Some initial comments questions on the non-test code changes, will review the test code in more detail in a subsequent pass.

base/collection_xattr.go
torcolvin: Replace implies that the _body_ has to exist, it does not care about the presence of the xattr.  This is covered by the test `TestWriteWithXattrs` but you should make sure to confirm.

base/collection_xattr.go
torcolvin: I don't actually know how we would hit this error, so I've removed this.

This comes from `deleteBodyAndXattrs` so I'm not sure if it was a copy paste mistake.

base/collection_xattr.go
torcolvin: I think this had to do with how I was doing an assert. This raises a CasMismatchErr in rosmar because it is expecting to be able to do an insert where there was no body. But I added the assertion that is `requireDocNotFoundOrCasMismatch` so I've removed this wrapped.

That said, I think in the future wrapping these types of errors would be a good idea, so we don't have to use `IsCasMismatch`

base/collection_xattr.go
torcolvin: fixed, I kept the outer function as separate so it was clear when you were doing an insertion or a resurrection conceptually.

It is unfortunate that we can't cas check on ressurection.

base/collection_xattr.go
torcolvin: I thought that this was required if the body was `nil` or `{}` because they unmarshal into nil but the only place this was used was where we did a raw comparison so I removed this.

None of the tests in the base package fail but I think https://jenkins.sgwdev.com/job/SyncGateway-Integration/2448/ [TestOnDemandWriteImportReplacingNilDoc](https://jenkins.sgwdev.com/job/SyncGateway-Integration/2448/testReport/junit/github/com_couchbase_sync_gateway_rest_importtest/TestOnDemandWriteImportReplacingNilDoc/) might be hitting this issue.

base/collection_xattr_common.go
torcolvin: This was added defensively and allowed me to check error conditions.

We talked about this offline, and it was a good idea to do in the tests.

base/collection_xattr_common.go
torcolvin: fixed

base/collection_xattr.go
```diff
@@ -677,3 +708,15 @@ func gocbMutationMacro(meType sgbucket.MacroExpansionType) gocb.MutationMacro {
 		return gocb.MutationMacroCAS
 	}
 }
+
+// getUpsertSpecs returns a slice of gocb.MutateInSpec for the given xattrs, or returns an error if any values are nil.
+func getUpsertSpecs(xattrs map[string][]byte) ([]gocb.MutateInSpec, error) {
```
adamcfraser:
```suggestion
func getUpsertSpecsForXattrs(xattrs map[string][]byte) ([]gocb.MutateInSpec, error) {
```
This might look like nitpicking but I found the generic "getUpsertSpecs" to be very ambiguous.

base/collection_xattr.go
adamcfraser: I think this question is still unresolved - any details on this?

base/collection_xattr_common.go
```diff
@@ -39,14 +41,17 @@ func (c *Collection) WriteWithXattrs(ctx context.Context, k string, exp uint32,
...
-			casOut, err = c.updateBodyAndXattrs(ctx, k, exp, cas, opts, v, xattrs)
+			casOut, err = c.updateBodyAndXattrs(ctx, k, exp, cas, opts, v, xattrs, xattrsToDelete)
 			if err != nil {
 				shouldRetry = c.isRecoverableWriteError(err)
 				return shouldRetry, err, uint64(0)
 			}
 		} else {
-			// Update xattr only
-			casOut, err = c.UpdateXattrs(ctx, k, exp, cas, xattrs, opts)
+			if len(xattrs) == 0 {
```
adamcfraser, line 50: Is there a valid case where all we're trying to do is delete xattrs (so v and xattrs are both empty)?  I guess you're assuming the caller is going to call RemoveXattrs for this case, but is there a reason we don't want to also handle that possibility in this function?

base/collection_xattr_common.go
adamcfraser: It feels unusual to check this *after* attempting the write a second time with the unchanged cas.  Should we be performing this check before we invoke the callback and attempt the write (i.e. maybe after the fetch on line 205)

base/collection_xattr_test.go
```diff
@@ -787,13 +783,71 @@ func TestWriteTombstoneWithXattrs(t *testing.T) {
...
-				cas:        fakeCas,
-				deleteBody: false,
-				// this seems wrong, no error??
-				finalBody: []byte("{}"),
+				cas:            fakeCas,
+				deleteBody:     false,
+				writeErrorFunc: RequireDocNotFoundError,
+			},
+			{
+				name:        "previousDoc=nil,xattrsToUpdate=_xattr1,xattrsToDelete=_xattr2,cas=0,deleteBody=false",
```
adamcfraser, line 791: We don't need to make this change in this PR, but I think these tests would be more readable and concise with the following changes:
- define an enum for the set of possible values for 'previousDoc', 'updatedXattrs' and 'finalXattrs' (e.g. for previousDoc, "none, "bodyOnly", "bodyWithXattr1", ...) 
- use cas=incorrect in the naming, instead of cas=1 or fakeCas

base/collection_xattr_test.go
```diff
@@ -1040,6 +1133,195 @@ func TestWriteUpdateWithXattrs(t *testing.T) {
 	}
 }
 
+func TestWriteWithXattrs(t *testing.T) {
+	ctx := TestCtx(t)
+	bucket := GetTestBucket(t)
+	defer bucket.Close(ctx)
+
+	col := bucket.DefaultDataStore()
```
adamcfraser: Should this be GetSingleDataStore?

base/collection_xattr_test.go
```diff
@@ -1040,6 +1133,195 @@ func TestWriteUpdateWithXattrs(t *testing.T) {
...
+				return
+			}
+			require.NoError(t, err)
+			require.NotEqual(t, uint64(0), cas)
+
+			xattrKeys := make([]string, 0, len(test.xattrs))
+			for k := range test.xattrs {
+				xattrKeys = append(xattrKeys, k)
+			}
+			xattrKeys = append(xattrKeys, test.xattrsToDelete...)
```
adamcfraser: I'm not sure why we're expecting to find xattrsToDelete here.  Running the test locally we don't have any subtests that reach this point where xattrsToDelete is non-nil.  Are we missing defining a successful case where WriteWithXattrs successfully deletes xattr(s)?

adamcfraser (changes requested):
The code changes look good, a few questions on tests and naming.

base/collection_xattr_test.go
torcolvin: It was a subtlety of earlier testing and copy/paste from a different test. We do successfully delete xattrs, but they all have to be written first, and so they are defined in `test.xattrs`

base/collection_xattr.go
torcolvin: `updateXattrs` is a private function and this is checked in the callers to this function in `WriteTombstoneWithXattrs` and `WriteWithXattrs`. `UpdateXattrs` doesn't allow you to delete xattrs since that's not yet needed. I think this has the potential to go wrong if the code is refactored and I like the idea to put this check in `updateXattrs` as well, though it will not presently be hit.

base/collection_xattr_common.go
torcolvin: I think it simplifies the number of things that this function is doing and forces the caller to make it more explicit. I don't feel strongly about this but the code is really complicated already.

base/collection_xattr_test.go
torcolvin: I like the idea of the enum but it becomes pretty confusing quickly because the contents of xattr1 could change values. I'm going to leave it out of this PR.

adamcfraser (dismissed):
LGTM

adamcfraser (approved):

PR #6794, opened 2024-04-30
Add `SkipTestMain` utility function for clearer output

bbrks:
This change makes it much more obvious that the package's tests are being completely skipped.

Formatting details can be negotiated. I tried to match the regular test output format for any parsers that may be interpreting this, however that may be more trouble than it's worth.

There's also no special exit code that we can return other than `0` for `TestMain` that would show `skipped` instead of `ok`. Any non-zero code shows `FAIL`.

## Before

```sh
$ go test -v -count=1 -run '^TestAutoImportUserXattrNoSyncData$' ./rest/importuserxattrtest
ok  	github.com/couchbase/sync_gateway/rest/importuserxattrtest	0.574s
```

## After

```sh
$ go test -v -count=1 -run '^TestAutoImportUserXattrNoSyncData$' ./rest/importuserxattrtest
=== RUN   TestMain
    main_test.go:24: user xattr tests require xattrs and EE
--- SKIP: TestMain (0.00s)
ok  	github.com/couchbase/sync_gateway/rest/importuserxattrtest	0.582s
```

### DIFF
--- a/base/main_test_bucket_pool.go
+++ b/base/main_test_bucket_pool.go
@@ -729,9 +729,9 @@ func TestBucketPoolMain(ctx context.Context, m *testing.M, bucketReadierFunc TBP
 	teardownFuncs = append(teardownFuncs, SetUpTestGoroutineDump(m))
 
 	if options.RequireXDCR && GTestBucketPool.cluster != nil && GTestBucketPool.cluster.majorVersion < 7 {
-		fmt.Println("Test requires XDCR, but the cluster version is less than 7. Skipping test.")
-		return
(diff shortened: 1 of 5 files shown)


torcolvin (commented):
This seems great, I've definitely been confused by this.

https://github.com/couchbase/sync_gateway/blob/ca9728dc86a5df3fe86372546cf383985fd78b11/db/indextest/main_test.go#L26
https://github.com/couchbase/sync_gateway/blob/ca9728dc86a5df3fe86372546cf383985fd78b11/rest/importtest/main_test.go#L22

should get added too.

torcolvin (approved):

PR #6801, opened 2024-05-01
CBG-3855: Sequence range removal support

gregns1:
CBG-3855

- Addition of support for removal of a given sequence range from the skipped sequence slice 
- Note: `RemoveSequences` function this is hooked up to doesn't actually get called yet, this will be properly hooked up to cache with the work in CBG-3850 

## Pre-review checklist

### DIFF
--- a/db/skipped_sequence.go
+++ b/db/skipped_sequence.go
@@ -218,26 +218,18 @@ func (s *SkippedSequenceSlice) removeSeqRange(startSeq, endSeq uint64) error {
 	defer s.lock.Unlock()
 
 	startIndex, found := s._findSequence(startSeq)
-	if !found {
-		return fmt.Errorf("sequence %d not found in the skipped list", startSeq)
-	}
-	endIndex, found := s._findSequence(endSeq)
-	if !found {
-		return fmt.Errorf("sequence %d not found in the skipped list", startSeq)
-	}
-
 	// if both sequence x and y aren't in the same element, the range contains sequences that are not present in
 	// skipped sequence list. This is due to any contiguous sequences in the list will be in the same element. So
 	// if startSeq and endSeq are in different elements there is at least one sequence between these values not in the list
-	if startIndex != endIndex {
+	rangeElem := s.list[startIndex]
+	if !found || endSeq > rangeElem.getLastSeq() {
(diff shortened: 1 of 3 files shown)


db/skipped_sequence.go
```diff
@@ -212,6 +212,56 @@ func binarySearchFunc(a *SkippedSequenceListEntry, seq uint64) int {
...
+// removeSeqRange will remove sequence between x and y from the skipped sequence slice
+func (s *SkippedSequenceSlice) removeSeqRange(startSeq, endSeq uint64) error {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	startIndex, found := s._findSequence(startSeq)
+	if !found {
+		return fmt.Errorf("sequence %d not found in the skipped list", startSeq)
+	}
+	endIndex, found := s._findSequence(endSeq)
```
adamcfraser: Since we're going to error out right away if the range isn't within a single element, I think it will be more efficient to just perform a single findSequence (on startSeq), and then examine the element to verify that the entire range is within that element.

db/skipped_sequence.go
```diff
@@ -212,6 +212,56 @@ func binarySearchFunc(a *SkippedSequenceListEntry, seq uint64) int {
...
+	// skipped sequence list. This is due to any contiguous sequences in the list will be in the same element. So
+	// if startSeq and endSeq are in different elements there is at least one sequence between these values not in the list
+	if startIndex != endIndex {
+		return fmt.Errorf("sequence range %d to %d specified has sequences in that are not present in skipped list", startSeq, endSeq)
+	}
+
+	// handle sequence range removal
+	rangeElem := s.list[startIndex]
+	// update number of sequences currently in slice stat
+	rangeRemoved := SkippedSequenceListEntry{start: startSeq, end: endSeq}
```
adamcfraser: Do we need to use a SkippedSequenceListEntry to compute the number of sequences being removed, or can we just do endSeq-startSeq+1?

adamcfraser (changes requested):
Generally looks good - one optimization suggestion and one question about stat computation.

adamcfraser (approved):

Issue #3774, opened 2018-10-01
Importing Key/Value pair with null value causes SG to panic


Binary-Ape:
Hello!

CB Version: `5.1.0` and `5.1.1`
SG Version: `Couchbase Sync Gateway/2.0.0(832;2d8a6c0)`
OS: `CentOS 7` and `OEL 6.10`

I have a sync gateway with its config below where `enable_shared_bucket_acccess` is true and `import` is set to continuous. If someone saves a Key/Value to Couchbase where the value is just `null` then the import process will cause the SG to crash and restart. It will continue to crash and restart until it either doesn't import for some reason or gets stuck trying to start up.

Here is a simple config file that I used to test the issue on:

```javascript
    "adminInterface": "0.0.0.0:4985",
        "interface": "0.0.0.0:4984",
        "log": ["CRUD+", "HTTP+", "Access", "Changes+", "Events+", "Import+"],
        "logging":{
                "default": {
                        "logFilePath": "/home/sync_gateway/logs/log.log",
                        "rotation": {
                                "maxsize": 1,
                                "maxage": 14,
                                "maxbackups": 15
                        }
                }
        },
        "databases": {
                "db": {
                        "server": "http://localhost:8091",
                        "bucket": "Test_Bucket",
                        "username": "Test_User_App",
                        "password": "couchbase",
                        "users": {
                                "GUEST": {"disabled": false, "admin_channels": ["*"] },
                                "TestSGUser": {"password":"couchbase" ,"disabled": false, "admin_channels":["*"]}
                        },
                        "enable_shared_bucket_access": true,
                        "import_docs": "continuous",
                        "sync": `
                             function(doc, oldDoc){
                                      if(doc.Name){
                                          channel(doc.Name)
                                     }
                            }`
                }
        }
```
### Log output
https://gist.githubusercontent.com/Binary-Ape/5eb2c5224df8cce792c34ca3b9a7dbbd/raw/825bda1e790e6942e0eaec1cbf94905d0141b794/SyncGateway_Null_Issue.log

https://gist.githubusercontent.com/Binary-Ape/5eb2c5224df8cce792c34ca3b9a7dbbd/raw/cc70139a8b39e5ff2c5117e9c7e3f64a38a3788f/sync_gateway_error.log

### Expected behavior
Sync Gateway should of imported the document with its null body.

### Actual behavior
Sync Gateway Crashed

### Steps to reproduce
1.  Create a bucket with shared bucket access and importing continuous
[...]

PR #6807, opened 2024-05-06
CBG-3877 Persist HLV to _vv xattr

adamcfraser:
Modifies document marshal and unmarshal to support a set of xattrs (_sync, _vv), and does the same for parsing DCP stream events (including user xattr). 

Bumps rosmar commit to pick up the already merged enhancement for multi-xattr support.

CBG-3877

Depends on:
- [x] https://github.com/couchbase/sg-bucket/pull/125

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2427/

### DIFF
--- a/base/constants.go
+++ b/base/constants.go
@@ -136,6 +136,7 @@ const (
 	SyncPropertyName = "_sync"
 	// SyncXattrName is used when storing sync data in a document's xattrs.
 	SyncXattrName = "_sync"
+	VvXattrName   = "_vv"
 
 	// Intended to be used in Meta Map and related tests
 	MetaMapXattrsKey = "xattrs"
--- a/db/crud.go
+++ b/db/crud.go
@@ -2255,6 +2255,9 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 
 			var rawSyncXattr, rawVvXattr, rawDocBody []byte
 			rawDocBody, rawSyncXattr, rawVvXattr, err = doc.MarshalWithXattrs()
+			if err != nil {
+				return updatedDoc, err
+			}
 			if !isImport {
 				updatedDoc.Doc = rawDocBody
 				docBytes = len(updatedDoc.Doc)
--- a/db/crud.go
+++ b/db/crud.go
@@ -60,7 +60,7 @@ func (c *DatabaseCollection) GetDocumentWithRaw(ctx context.Context, docid strin
 		return nil, nil, base.HTTPErrorf(400, "Invalid doc ID")
 	}
 	if c.UseXattrs() {
-		doc, rawBucketDoc, err = c.GetDocWithXattr(ctx, key, unmarshalLevel)
+		doc, rawBucketDoc, err = c.GetDocWithXattrs(ctx, key, unmarshalLevel)
 		if err != nil {
 			return nil, nil, err
 		}
@@ -73,7 +73,7 @@ func (c *DatabaseCollection) GetDocumentWithRaw(ctx context.Context, docid strin
 		// If existing doc wasn't an SG Write, import the doc.
 		if !isSgWrite {
 			var importErr error
-			doc, importErr = c.OnDemandImportForGet(ctx, docid, rawBucketDoc.Body, rawBucketDoc.Xattrs[base.SyncXattrName], rawBucketDoc.Xattrs[c.userXattrKey()], rawBucketDoc.Cas)
+			doc, importErr = c.OnDemandImportForGet(ctx, docid, rawBucketDoc.Body, rawBucketDoc.Xattrs, rawBucketDoc.Cas)
 			if importErr != nil {
 				return nil, nil, importErr
 			}
@@ -114,7 +114,7 @@ func (c *DatabaseCollection) GetDocumentWithRaw(ctx context.Context, docid strin
 	return doc, rawBucketDoc, nil
 }
 
-func (c *DatabaseCollection) GetDocWithXattr(ctx context.Context, key string, unmarshalLevel DocumentUnmarshalLevel) (doc *Document, rawBucketDoc *sgbucket.BucketDocument, err error) {
+func (c *DatabaseCollection) GetDocWithXattrs(ctx context.Context, key string, unmarshalLevel DocumentUnmarshalLevel) (doc *Document, rawBucketDoc *sgbucket.BucketDocument, err error) {
 	rawBucketDoc = &sgbucket.BucketDocument{}
 	var getErr error
 	rawBucketDoc.Body, rawBucketDoc.Xattrs, rawBucketDoc.Cas, getErr = c.dataStore.GetWithXattrs(ctx, key, c.syncAndUserXattrKeys())
@@ -123,7 +123,7 @@ func (c *DatabaseCollection) GetDocWithXattr(ctx context.Context, key string, un
 	}
 
 	var unmarshalErr error
-	doc, unmarshalErr = unmarshalDocumentWithXattr(ctx, key, rawBucketDoc.Body, rawBucketDoc.Xattrs[base.SyncXattrName], rawBucketDoc.Xattrs[c.userXattrKey()], rawBucketDoc.Cas, unmarshalLevel)
+	doc, unmarshalErr = c.unmarshalDocumentWithXattrs(ctx, key, rawBucketDoc.Body, rawBucketDoc.Xattrs, rawBucketDoc.Cas, unmarshalLevel)
 	if unmarshalErr != nil {
 		return nil, nil, unmarshalErr
 	}
@@ -148,11 +148,8 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
 			return emptySyncData, getErr
 		}
 
-		rawXattr := xattrs[base.SyncXattrName]
-		rawUserXattr := xattrs[c.userXattrKey()]
-
 		// Unmarshal xattr only
-		doc, unmarshalErr := unmarshalDocumentWithXattr(ctx, docid, nil, rawXattr, rawUserXattr, cas, DocUnmarshalSync)
+		doc, unmarshalErr := c.unmarshalDocumentWithXattrs(ctx, docid, nil, xattrs, cas, DocUnmarshalSync)
 		if unmarshalErr != nil {
 			return emptySyncData, unmarshalErr
 		}
@@ -166,7 +163,7 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
 		if !isSgWrite {
 			var importErr error
 
-			doc, importErr = c.OnDemandImportForGet(ctx, docid, rawDoc, rawXattr, rawUserXattr, cas)
+			doc, importErr = c.OnDemandImportForGet(ctx, docid, rawDoc, xattrs, cas)
 			if importErr != nil {
 				return emptySyncData, importErr
 			}
@@ -193,6 +190,12 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
 
 }
 
+// unmarshalDocumentWithXattrs populates individual xattrs on unmarshalDocumentWithXattrs from a provided xattrs map
+func (db *DatabaseCollection) unmarshalDocumentWithXattrs(ctx context.Context, docid string, data []byte, xattrs map[string][]byte, cas uint64, unmarshalLevel DocumentUnmarshalLevel) (doc *Document, err error) {
+	return unmarshalDocumentWithXattrs(ctx, docid, data, xattrs[base.SyncXattrName], xattrs[base.VvXattrName], xattrs[db.userXattrKey()], cas, unmarshalLevel)
+
+}
+
 // This gets *just* the Sync Metadata (_sync field) rather than the entire doc, for efficiency
 // reasons. Unlike GetDocSyncData it does not check for on-demand import; this means it does not
 // need to read the doc body from the bucket.
@@ -200,10 +203,10 @@ func (db *DatabaseCollection) GetDocSyncDataNoImport(ctx context.Context, docid 
 	if db.UseXattrs() {
 		var xattrs map[string][]byte
 		var cas uint64
-		xattrs, cas, err = db.dataStore.GetXattrs(ctx, docid, []string{base.SyncXattrName})
+		xattrs, cas, err = db.dataStore.GetXattrs(ctx, docid, []string{base.SyncXattrName, base.VvXattrName})
 		if err == nil {
 			var doc *Document
-			doc, err = unmarshalDocumentWithXattr(ctx, docid, nil, xattrs[base.SyncXattrName], nil, cas, level)
+			doc, err = db.unmarshalDocumentWithXattrs(ctx, docid, nil, xattrs, cas, level)
 			if err == nil {
 				syncData = doc.SyncData
 			}
@@ -235,14 +238,14 @@ func (db *DatabaseCollection) GetDocSyncDataNoImport(ctx context.Context, docid 
 	return
 }
 
-// OnDemandImportForGet.  Attempts to import the doc based on the provided id, contents and cas.  ImportDocRaw does cas retry handling
+// OnDemandImportForGet. Attempts to import the doc based on the provided id, contents and cas. ImportDocRaw does cas retry handling
 // if the document gets updated after the initial retrieval attempt that triggered this.
-func (c *DatabaseCollection) OnDemandImportForGet(ctx context.Context, docid string, rawDoc []byte, rawXattr []byte, rawUserXattr []byte, cas uint64) (docOut *Document, err error) {
+func (c *DatabaseCollection) OnDemandImportForGet(ctx context.Context, docid string, rawDoc []byte, xattrs map[string][]byte, cas uint64) (docOut *Document, err error) {
 	isDelete := rawDoc == nil
 	importDb := DatabaseCollectionWithUser{DatabaseCollection: c, user: nil}
 	var importErr error
 
-	docOut, importErr = importDb.ImportDocRaw(ctx, docid, rawDoc, rawXattr, rawUserXattr, isDelete, cas, nil, ImportOnDemand)
+	docOut, importErr = importDb.ImportDocRaw(ctx, docid, rawDoc, xattrs, isDelete, cas, nil, ImportOnDemand)
 	if importErr == base.ErrImportCancelledFilter {
 		// If the import was cancelled due to filter, treat as not found
 		return nil, base.HTTPErrorf(404, "Not imported")
@@ -1076,7 +1079,7 @@ func (db *DatabaseCollectionWithUser) Put(ctx context.Context, docid string, bod
 func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Context, newDoc *Document, newDocHLV HybridLogicalVector, existingDoc *sgbucket.BucketDocument) (doc *Document, cv *Version, newRevID string, err error) {
 	var matchRev string
 	if existingDoc != nil {
-		doc, unmarshalErr := unmarshalDocumentWithXattr(ctx, newDoc.ID, existingDoc.Body, existingDoc.Xattrs[base.SyncXattrName], existingDoc.Xattrs[db.userXattrKey()], existingDoc.Cas, DocUnmarshalRev)
+		doc, unmarshalErr := db.unmarshalDocumentWithXattrs(ctx, newDoc.ID, existingDoc.Body, existingDoc.Xattrs, existingDoc.Cas, DocUnmarshalRev)
 		if unmarshalErr != nil {
 			return nil, nil, "", base.HTTPErrorf(http.StatusBadRequest, "Error unmarshaling existing doc")
 		}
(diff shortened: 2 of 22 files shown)


db/crud.go
```diff
@@ -192,17 +189,23 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
 
 }
 
+// Helper function to populate individual xattrs on unmarshalDocumentWithXattr from existing xattrs map
```
```suggestion
// unmarshalDocumentWithXattrs returns a copy of a document, sync gateway xattrs, and user xattrs.
```

db/import.go
```diff
@@ -91,7 +86,7 @@ func (db *DatabaseCollectionWithUser) ImportDoc(ctx context.Context, docid strin
 		if existingDoc.Deleted {
 			existingBucketDoc.Xattrs[base.SyncXattrName], err = base.JSONMarshal(existingDoc.SyncData)
 		} else {
-			existingBucketDoc.Body, existingBucketDoc.Xattrs[base.SyncXattrName], err = existingDoc.MarshalWithXattr()
+			existingBucketDoc.Body, existingBucketDoc.Xattrs[base.SyncXattrName], existingBucketDoc.Xattrs[base.VvXattrName], err = existingDoc.MarshalWithXattr()
```
torcolvin: I think it might make sense to use

```
MarshalWithXattr() (data []byte, sgXattrs map[string][]byte, err error)
```
because I think in all use cases we immediately assign this to a `sgbucket.UpdatedDoc` or a dictionary anyway.

We are going to need this for `_mou` when there will be another value.

The one circumstance I worry about a direct assignment is the case where we do a resurrection. If there are non SG xattrs, they have to preserved for resurrection. However, I don't think Sync Gateway does this, and even if it did, this would be some kind of merging of two maps.

db/import.go
```diff
@@ -388,15 +383,19 @@ func (db *DatabaseCollectionWithUser) migrateMetadata(ctx context.Context, docid
 	}
 
 	// Persist the document in xattr format
-	value, xattrValue, marshalErr := doc.MarshalWithXattr()
+	value, syncXattr, vvXattr, marshalErr := doc.MarshalWithXattr()
```
torcolvin: When migrating an existing document, wouldn't you want to assign a _vv, or are you assuming this is done later in the import?

Regardless, I think it probably makes sense to write a unit test for `migrateMetadata` that handles  `_vv` but no `_sync` xattr. This case is sufficiently weird that it might be out of sync of its _sync xattrs.

db/import_test.go
```diff
@@ -418,8 +418,10 @@ func TestImportNullDocRaw(t *testing.T) {
 
 	// Feed import of null doc
 	exp := uint32(0)
-
-	importedDoc, err := collection.ImportDocRaw(ctx, "TestImportNullDoc", []byte("null"), []byte("{}"), nil, false, 1, &exp, ImportFromFeed)
+	xattrs := map[string][]byte{
+		base.SyncXattrName: []byte("{}"),
+	}
```
torcolvin, line 487: Possible tangent: I have looked at this test before recently and it's weird. I think this is testing that the body is literally `null`. However, this is now testing something odd with `_sync` present and no `_vv`. This might be useful to test, or not.

I think it might be done this way so import doesn't hit `isSGWrite`, but I don't remember.

db/utilities_hlv_testing.go
```diff
@@ -48,8 +48,7 @@ func (h *HLVAgent) InsertWithHLV(ctx context.Context, key string) (casOut uint64
 	require.NoError(h.t, err)
 	hlv.CurrentVersionCAS = hlvExpandMacroCASValue
 
-	syncData := &SyncData{HLV: hlv}
-	syncDataBytes, err := base.JSONMarshal(syncData)
+	vvDataBytes, err := base.JSONMarshal(hlv)
 	require.NoError(h.t, err)
```
torcolvin: nit:

```suggestion
	vvDataBytes, err := base.MustJSONMarshal(h.t, hlv)
```

I might add something that this is is an SDK level insert, but perhaps this is clear.

go.sum
```diff
@@ -74,6 +74,8 @@ github.com/couchbaselabs/gocbconnstr/v2 v2.0.0-20230515165046-68b522a21131 h1:2E
 github.com/couchbaselabs/gocbconnstr/v2 v2.0.0-20230515165046-68b522a21131/go.mod h1:o7T431UOfFVHDNvMBUmUxpHnhivwv7BziUao/nMl81E=
 github.com/couchbaselabs/rosmar v0.0.0-20240404180245-795e6df684f3 h1:AUvojYsPc2WgiO9xRalRQLyXzooRRWemdEkiGl+PZno=
 github.com/couchbaselabs/rosmar v0.0.0-20240404180245-795e6df684f3/go.mod h1:SM0w4YHwXFMIyfqUbkpXZNWwAQKLwsUH91fsKUooMqw=
+github.com/couchbaselabs/rosmar v0.0.0-20240424002439-5fd321c3b01b h1:aP8cbbI/CiRyL0mOGA4HJ9jjhcOgx1M9V78i2tAAdJ4=
```
torcolvin, line 77: I'd run a `go mod tidy` here.

base/constants.go
```diff
@@ -136,6 +136,7 @@ const (
 	SyncPropertyName = "_sync"
 	// SyncXattrName is used when storing sync data in a document's xattrs.
 	SyncXattrName = "_sync"
+	VvXattrName   = "_vv"
```
torcolvin, line 139: nit:

```suggestion
	// VvXattrName stores the hybrid logical vector of a document.
	VvXattrName   = "_vv"
```

db/crud.go
```diff
@@ -236,12 +239,12 @@ func (db *DatabaseCollection) GetDocSyncDataNoImport(ctx context.Context, docid
 
 // OnDemandImportForGet.  Attempts to import the doc based on the provided id, contents and cas.  ImportDocRaw does cas retry handling
```
```suggestion
// OnDemandImportForGet. Attempts to import the doc based on the provided id, contents, xattrs, and cas.  ImportDocRaw does cas retry handling
```

db/crud.go
```diff
@@ -2192,15 +2195,14 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 		}
 		casOut, err = db.dataStore.WriteUpdateWithXattrs(ctx, key, db.syncAndUserXattrKeys(), initialExpiry, existingDoc, opts, func(currentValue []byte, currentXattrs map[string][]byte, cas uint64) (updatedDoc sgbucket.UpdatedDoc, err error) {
 			// Be careful: this block can be invoked multiple times if there are races!
-			currentXattr := currentXattrs[base.SyncXattrName]
-			currentUserXattr := currentXattrs[db.userXattrKey()]
-			if doc, err = unmarshalDocumentWithXattr(ctx, docid, currentValue, currentXattr, currentUserXattr, cas, DocUnmarshalAll); err != nil {
+			currentSyncXattr := currentXattrs[base.SyncXattrName]
```
torcolvin: move this line to:

```
if currentSyncXattr == nil && doc.Sequence > 0 {
```

Could it matter if this value is not present vs empty byte array? I am not sure if I distinguish between that, or if it is the same between rosmar/CBS? It was at one point, perhaps still, unintentionally different between CBS/rosmar.

db/document.go
```diff
@@ -402,14 +402,17 @@ func unmarshalDocument(docid string, data []byte) (*Document, error) {
 	return doc, nil
 }
 
-func unmarshalDocumentWithXattr(ctx context.Context, docid string, data []byte, xattrData []byte, userXattrData []byte, cas uint64, unmarshalLevel DocumentUnmarshalLevel) (doc *Document, err error) {
+func unmarshalDocumentWithXattr(ctx context.Context, docid string, data []byte, syncXattrData []byte, hlvXattrData []byte, userXattrData []byte, cas uint64, unmarshalLevel DocumentUnmarshalLevel) (doc *Document, err error) {
 
-	if xattrData == nil || len(xattrData) == 0 {
+	if (syncXattrData == nil || len(syncXattrData) == 0) && hlvXattrData == nil {
```
torcolvin: This should be able to be simplified to

```suggestion
	if len(syncXattrData) == 0 && hlvXattrData == nil {
```

db/document.go
```diff
@@ -402,14 +402,17 @@ func unmarshalDocument(docid string, data []byte) (*Document, error) {
...
 		// If no xattr data, unmarshal as standard doc
 		doc, err = unmarshalDocument(docid, data)
 	} else {
 		doc = NewDocument(docid)
-		err = doc.UnmarshalWithXattr(ctx, data, xattrData, unmarshalLevel)
+		err = doc.UnmarshalWithXattr(ctx, data, syncXattrData, hlvXattrData, unmarshalLevel)
+		if err != nil {
+			return nil, err
+		}
```
torcolvin: I'd be tempted to scope err locally in this block, or letting it fall out of the block like the first half of the if statement if there are no xattrs

db/document.go
```diff
@@ -520,21 +549,23 @@ func UnmarshalDocumentFromFeed(ctx context.Context, docid string, cas uint64, da
 	The 0x00 byte after the key saves us from storing a key length, and the trailing 0x00 is just for convenience to allow us to use string functions to search in them.
 */
 
-func parseXattrStreamData(xattrName string, userXattrName string, data []byte) (body []byte, xattr []byte, userXattr []byte, err error) {
+func parseXattrsStreamData(xattrNames []string, data []byte) (body []byte, xattrs map[string][]byte, err error) {
```
torcolvin: This function is a duplicate of https://github.com/couchbase/sg-bucket/blob/2a857aeebbde1fe836ea10b159fe2ebbffc0f07a/tap.go#L148 which is used by rosmar.

I'd be lovely to include this comment in the sg-bucket implementation.

db/document.go
```diff
@@ -1129,19 +1179,27 @@ func (doc *Document) MarshalJSON() (data []byte, err error) {
...
 // lazy unmarshalling as needed.
-func (doc *Document) UnmarshalWithXattr(ctx context.Context, data []byte, xdata []byte, unmarshalLevel DocumentUnmarshalLevel) error {
+func (doc *Document) UnmarshalWithXattr(ctx context.Context, data, syncXattrData, hlvXattrData []byte, unmarshalLevel DocumentUnmarshalLevel) error {
 	if doc.ID == "" {
 		base.WarnfCtx(ctx, "Attempted to unmarshal document without ID set")
 		return errors.New("Document was unmarshalled without ID set")
 	}
 
 	switch unmarshalLevel {
 	case DocUnmarshalAll, DocUnmarshalSync:
```
torcolvin, line 1074: It might make sense to update `DocUnmarshalAll` and `DocUnmarshalSync` docstrings explaining that they are going to include HLV.

db/document.go
```diff
@@ -1129,19 +1179,27 @@ func (doc *Document) MarshalJSON() (data []byte, err error) {
 // (unmarshalLevel) specifies how much of the provided document/xattr needs to be initially unmarshalled.  If
 // unmarshalLevel is anything less than the full document + metadata, the raw data is retained for subsequent
 // lazy unmarshalling as needed.
-func (doc *Document) UnmarshalWithXattr(ctx context.Context, data []byte, xdata []byte, unmarshalLevel DocumentUnmarshalLevel) error {
+func (doc *Document) UnmarshalWithXattr(ctx context.Context, data, syncXattrData, hlvXattrData []byte, unmarshalLevel DocumentUnmarshalLevel) error {
```
torcolvin: Update name of function with s/xattr/xattrs/ and state that both xattrs are optional?

db/document.go
```diff
@@ -1151,14 +1209,20 @@ func (doc *Document) UnmarshalWithXattr(ctx context.Context, data []byte, xdata
...
 		doc.SyncData = SyncData{}
-		unmarshalErr := base.JSONUnmarshal(xdata, &doc.SyncData)
+		unmarshalErr := base.JSONUnmarshal(syncXattrData, &doc.SyncData)
 		if unmarshalErr != nil {
 			return pkgerrors.WithStack(base.RedactErrorf("Failed to UnmarshalWithXattr() doc with id: %s (DocUnmarshalNoHistory).  Error: %v", base.UD(doc.ID), unmarshalErr))
 		}
+		if hlvXattrData != nil {
+			err := base.JSONUnmarshal(hlvXattrData, &doc.SyncData.HLV)
+			if err != nil {
+				return pkgerrors.WithStack(base.RedactErrorf("Failed to unmarshal HLV during UnmarshalWithXattr() doc with id: %s (DocUnmarshalAll/Sync).  Error: %v", base.UD(doc.ID), err))
```
```suggestion
				return pkgerrors.WithStack(base.RedactErrorf("Failed to unmarshal HLV during UnmarshalWithXattr() doc with id: %s (DocUnmarshalNoHistory).  Error: %v", base.UD(doc.ID), err))
```

I'd be tempted to do something like define:

```
type DocumentUnmarshalLevel uint8

func (l DocumentUnmarshalLevel) String() {
	switch l {
	case DocUnmarshalLevel:
		return "DocUnmarshalAll"     
	case DocUnmarshalSync:
		return "DocUnmarshalSync"
	case DocUnmarshalNoHistory:
		return "DocUnmarshalNoHistory"
	case DocUnmarshalHistory:
		return "DocUnmarshalHistory"
	case DocUnmarshalRev:
		return "DocUnmarshalRev"
	case DocUnmarshalCAS:
		return "DocUnmarshalCAS"
	case DocUnmarshalNone:
		return "DocUmarshalNone"
	}
	return "DocumentUnmarshalLevel %d", l)
)

File: db/document.go [hunk 16]
```diff
@@ -1194,15 +1258,15 @@ func (doc *Document) UnmarshalWithXattr(ctx context.Context, data []byte, xdata
 	}
 
 	// If there's no body, but there is an xattr, set deleted flag and initialize an empty body
-	if len(data) == 0 && len(xdata) > 0 {
+	if len(data) == 0 && len(syncXattrData) > 0 {
```
torcolvin, line 1160: Can we run into a case where there's a tombstone with an hlv but not a sync xattr? If that's the case, do we want to preserve the hlv?

I don't know if any other products will write tombstones, but I could imagine an SDK write of a tombstone ending up here.

Either way, I'd probably write a test asserting this behavior if we do decide to make a change.

db/document.go
```diff
@@ -1194,15 +1258,15 @@ func (doc *Document) UnmarshalWithXattr(ctx context.Context, data []byte, xdata
...
 		doc._body = Body{}
 		doc._rawBody = []byte(base.EmptyDocument)
 		doc.Deleted = true
 	}
 	return nil
 }
 
-func (doc *Document) MarshalWithXattr() (data []byte, xdata []byte, err error) {
+func (doc *Document) MarshalWithXattr() (data []byte, syncXattr []byte, vvXattr []byte, err error) {
```
```suggestion
// MarshalWithXattrs returns the byte arrays of the  body and the xattrs of a document.
func (doc *Document) MarshalWithXattrs() (data []byte, syncXattr []byte, vvXattr []byte, err error) {
```

db/crud.go
```diff
@@ -192,17 +189,23 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
 
 }
 
+// Helper function to populate individual xattrs on unmarshalDocumentWithXattr from existing xattrs map
+func (db *DatabaseCollection) unmarshalDocumentWithXattrs(ctx context.Context, docid string, data []byte, xattrs map[string][]byte, cas uint64, unmarshalLevel DocumentUnmarshalLevel) (doc *Document, err error) {
+	return unmarshalDocumentWithXattr(ctx, docid, data, xattrs[base.SyncXattrName], xattrs[base.VvXattrName], xattrs[db.userXattrKey()], cas, unmarshalLevel)
```
torcolvin: I think we should make the xattr vs xattrs name consistent here.

db/import.go
```diff
@@ -388,15 +383,19 @@ func (db *DatabaseCollectionWithUser) migrateMetadata(ctx context.Context, docid
...
 	if marshalErr != nil {
 		return nil, false, marshalErr
 	}
 
 	// Use WriteWithXattr to handle both normal migration and tombstone migration (xattr creation, body delete)
```
torcolvin:
```suggestion
```

This comment is out of date because we use different functions for tombstone vs non tombstone.

torcolvin (commented):
Some tests that might be missing: 

- an explicit test for _vv but no _sync is worthwhile for the import code (on demand and feed).
- coverage if _vv is out of sync from cas? (maybe this is too paranoid)

db/import.go
adamcfraser: In the migrate case, inline _sync here would represent metadata for a version that shouldn't be accounted for in the HLV (as it the mutation was created prior to 4.0), so I the migrate should be a simple migrate.  Adding an explicit test for this case might be reasonable, though.

db/import.go
adamcfraser: I considered this, but ended up preferring the clarity around exactly what MarshalWithXattr was generating in the signature.

db/import_test.go
adamcfraser: Based on the original issue, I think `null` is intentional here.  (https://github.com/couchbase/sync_gateway/issues/3774)

db/utilities_hlv_testing.go
adamcfraser: Good suggestion on MustJSONMarshal.  I think the test description covers the intention, but I can tweak it to make it more clear that "a different HLV-aware peer" is a non-SG peer.

db/document.go
adamcfraser: That's a reasonable idea, would probably want to do that in a separate PR.

db/document.go
adamcfraser: SGW doesn't care about tombstones that don't have a _sync xattr on them (it means they were never imported as a non-tombstone, and so the delete doesn't need to be replicated anywhere by SGW).

db/import.go
torcolvin: Does this change if we add `_mou`? Should have a struct that specifically contains the SG xattrs? I am not set on this but I am worried about returning three byte arrays next to each other.

adamcfraser: Even with _mou I expect I'm going to prefer the named returns, but can revisit at that time.

db/import.go
adamcfraser: Added that test, see TestMigrateMetadataWithHLV.

base/constants.go
adamcfraser: The previous comment (for SyncXattrName) is there to differentiate between the usage of SyncPropertyName and SyncXattrName.  I think VvXattrName is self-explanatory?

db/document.go
adamcfraser: It seems incorrect to be checking for zero length on one xattr and nil on the other?

db/document.go
adamcfraser: I agree, but I think it's out of the scope of this PR.

db/document.go
adamcfraser: There's probably some ambiguity here about whether 'sync metadata' refers to both _sync and _vv xattrs (as it's the combination that makes up SGW's metadata).  I've updated the comments to just say 'metadata' to avoid confusion.

adamcfraser:
Force pushed the latest branch after rebasing on anemone, kicking off new integration tests.

db/crud.go
```diff
@@ -2249,18 +2252,19 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 
 			// Return the new raw document value for the bucket to store.
 			doc.SetCrc32cUserXattrHash()
-			var rawXattr, rawDocBody []byte
-			rawDocBody, rawXattr, err = doc.MarshalWithXattr()
+
+			var rawSyncXattr, rawVvXattr, rawDocBody []byte
+			rawDocBody, rawSyncXattr, rawVvXattr, err = doc.MarshalWithXattrs()
```
torcolvin, line 2257: Take or leave this because. the rest of the code seems fine and this is existing behavior. Conceptually, I'd expect an immediate return here.

db/document.go
torcolvin: I'd probably just change this to len checks for both in case someone passes a literal `[]byte{}` in an argument and len(nil) is valid.

db/document.go
```diff
@@ -1081,58 +1094,79 @@ func (doc *Document) UnmarshalWithXattr(ctx context.Context, data []byte, xdata
...
-		doc.SyncData = SyncData{
-			CurrentRev: revOnlyMeta.CurrentRev.RevTreeID,
-			Cas:        revOnlyMeta.Cas,
+		if syncXattrData != nil {
+			var revOnlyMeta revOnlySyncData
+			unmarshalErr := base.JSONUnmarshal(syncXattrData, &revOnlyMeta)
+			if unmarshalErr != nil {
+				return pkgerrors.WithStack(base.RedactErrorf("Failed to UnmarshalWithXattrs() doc with id: %s (DocUnmarshalRev).  Error: %v", base.UD(doc.ID), unmarshalErr))
+			}
+			doc.SyncData = SyncData{
```
torcolvin, line 1134: `PutExistingCurrentVersion` calls this function with `DocUnmarshalRev` and that function is used by `processRev` request.

It seems like this should deal with hlv as well. I don't know if the name `DocUnmarshalRev` here is a confusing name given that it will use rev or hlv, if I understand correctly.

adamcfraser: `PutExistingCurrentVersion` is only using this to look up the existing revTreeID, to compute the generation for the newly generated revTreeID (for backward compatibility), so shouldn't need HLV.  I'll update the comment for `DocUnmarshalRev` to make it clear it's specifically revTreeID.

torcolvin (approved):

PR #6812, opened 2024-05-08
CBG-3913 replace parseStreamWithXattr with sg-bucket function

torcolvin:
Depends on https://github.com/couchbase/sg-bucket/pull/122 and https://github.com/couchbase/sg-bucket/pull/123

- Consolidate tests into a table driven test, these tests exist in sg-bucket too with just `DecodeValueWithXattrs`. 
- Replace errors with sgbucket errors and use errors.Is

## Pre-review checklist

- [x] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2406/

### DIFF
--- a/db/attachment_compaction.go
+++ b/db/attachment_compaction.go
@@ -211,18 +211,11 @@ func getAttachmentSyncData(dataType uint8, data []byte) (*AttachmentCompactionDa
 	var documentBody []byte
 
 	if dataType&base.MemcachedDataTypeXattr != 0 {
-		body, xattrs, err := sgbucket.DecodeValueWithXattrs(data)
+		body, xattrs, err := sgbucket.DecodeValueWithXattrs([]string{base.SyncXattrName}, data)
 		if err != nil {
 			return nil, fmt.Errorf("Could not parse DCP attachment sync data: %w", err)
 		}
-		var syncXattr []byte
-		for _, xattr := range xattrs {
-			if xattr.Name == base.SyncXattrName {
-				syncXattr = xattr.Value
-			}
-		}
-
-		err = base.JSONUnmarshal(syncXattr, &attachmentData)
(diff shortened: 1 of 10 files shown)


torcolvin:
I was thinking about changing `sgbucket.DecodeValueWithXattrs` to return `map[string][]byte` for xattr values so the caller doesn't have to loop over them again.

adamcfraser (dismissed):

db/attachment_compaction.go
```diff
@@ -211,15 +211,11 @@ func getAttachmentSyncData(dataType uint8, data []byte) (*AttachmentCompactionDa
...
 
 	if dataType&base.MemcachedDataTypeXattr != 0 {
-		body, xattr, _, err := parseXattrStreamData(base.SyncXattrName, "", data)
+		body, xattrs, err := sgbucket.DecodeValueWithXattrs([]string{base.SyncXattrName}, data)
 		if err != nil {
-			if errors.Is(err, base.ErrXattrNotFound) || errors.Is(err, base.ErrXattrInvalidLen) {
-				return nil, nil
-			}
-			return nil, err
+			return nil, fmt.Errorf("Could not parse DCP attachment sync data: %w", err)
```
adamcfraser, line 219: This looks like a functional change if the xattr isn't found.  Should it be checking if value was found?

adamcfraser (changes requested):

db/attachment_compaction.go
torcolvin: This looks like a functional change but `parseXattrStreamData` never returned `ErrXattrNotFound`. That said, I should catch `ErrXattrInvalidLen` though realistically this is also an error condition where we probably could consider stopping.

adamcfraser (approved):

PR #6822, opened 2024-05-13
CBG-3157: Add opt-in send replacementRevs feature when encountering norev

bbrks:
CBG-3157

Add opt-in replacement revs feature.
- In the event a requested revision is not sent, and a client has opted into this feature, we'll fetch the active revision of the document and send that instead.
- Access checks for the replacement rev are performed by `GetRev`
- Channel fitlers are checked to ensure the replacement rev is in the set of currently replicated channels

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2460/

### DIFF
--- a/base/set.go
+++ b/base/set.go
@@ -56,10 +56,14 @@ func (set Set) copy() Set {
 	return result
 }
 
-// Returns true if the set includes the channel.
-func (set Set) Contains(ch string) bool {
-	_, exists := set[ch]
-	return exists
+// Contains returns true if the set includes any of the given channels.
+func (set Set) Contains(ch ...string) bool {
+	for _, c := range ch {
+		if _, exists := set[c]; exists {
+			return true
+		}
+	}
+	return false
 }
 
 func (set Set) Equals(other Set) bool {
--- a/base/set.go
+++ b/base/set.go
@@ -66,6 +66,21 @@ func (set Set) Contains(ch ...string) bool {
 	return false
 }
 
+// NumMatches returns the number of matching elements between both sets.
+func (a Set) NumMatches(b Set) (count int) {
+	shortest, longest := a, b
+	// iterate over shortest
+	if len(longest) < len(shortest) {
+		shortest, longest = longest, shortest
+	}
+	for name := range shortest {
+		if _, exists := longest[name]; exists {
+			count++
+		}
+	}
+	return count
+}
+
 func (set Set) Equals(other Set) bool {
 	if len(other) != len(set) {
 		return false
--- a/base/set.go
+++ b/base/set.go
@@ -66,6 +66,21 @@ func (set Set) Contains(ch ...string) bool {
 	return false
 }
 
+// HasMatch returns the true if there is at least one matching element between both sets.
+func (a Set) HasMatch(b Set) bool {
+	shortest, longest := a, b
+	// iterate over shortest
+	if len(longest) < len(shortest) {
+		shortest, longest = longest, shortest
+	}
+	for name := range shortest {
+		if _, exists := longest[name]; exists {
+			return true
+		}
+	}
+	return false
+}
+
 // NumMatches returns the number of matching elements between both sets.
 func (a Set) NumMatches(b Set) (count int) {
 	shortest, longest := a, b
--- a/db/blip_collection_context.go
+++ b/db/blip_collection_context.go
@@ -36,6 +36,8 @@ type blipSyncCollectionContext struct {
 
 	// if we encounter a situation where a requested revision is not available, should we try sending an alternative as a replacement revision?
 	sendReplacementRevs bool
+	// set if the client has requested a filtered set of channels
+	channels base.Set
 }
 
 // blipCollections is a container for all collections blip is aware of.
--- a/db/blip_collection_context.go
+++ b/db/blip_collection_context.go
@@ -34,6 +34,8 @@ type blipSyncCollectionContext struct {
 	sgr2PushAlreadyKnownSeqsCallback func(alreadyKnownSeqs ...SequenceID)           // sgr2PushAlreadyKnownSeqsCallback is called to mark the sequence as being immediately processed
 	emptyChangesMessageCallback      func()                                         // emptyChangesMessageCallback is called when an empty changes message is received
 
+	// if we encounter a situation where a requested revision is not available, should we try sending an alternative as a replacement revision?
+	sendReplacementRevs bool
 }
 
 // blipCollections is a container for all collections blip is aware of.
--- a/db/blip_collection_context.go
+++ b/db/blip_collection_context.go
@@ -34,7 +34,7 @@ type blipSyncCollectionContext struct {
 	sgr2PushAlreadyKnownSeqsCallback func(alreadyKnownSeqs ...SequenceID)           // sgr2PushAlreadyKnownSeqsCallback is called to mark the sequence as being immediately processed
 	emptyChangesMessageCallback      func()                                         // emptyChangesMessageCallback is called when an empty changes message is received
 
-	// if we encounter a situation where a requested revision is not available, should we try sending an alternative as a replacement revision?
+	// when a requested revision is not available, whether we should send the active revision as a replacement revision (depending on access)
 	sendReplacementRevs bool
 	// set if the client has requested a filtered set of channels
 	channels base.Set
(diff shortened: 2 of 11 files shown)


db/blip_sync_context.go
```diff
@@ -617,8 +620,26 @@ func (bsc *BlipSyncContext) sendNoRev(sender *blip.Sender, docID, revID string,
...
+	}
+
+	if collectionCtx.sendReplacementRevs && base.IsDocNotFoundError(err) {
+		base.DebugfCtx(bsc.loggingCtx, base.KeySync, "Unavailable revision for %q %s - finding replacement: %v", base.UD(docID), revID, err)
+
+		// try the active rev instead as a replacement
+		replacementRev, replacementRevErr := handleChangesResponseCollection.GetRev(bsc.loggingCtx, docID, "", true, nil)
+		// set replacement and continue as normal
+		replacedRevID = revID
+		revID = replacementRev.RevID
```
adamcfraser, line 650: If replacementRevErr is 'not found', replacementRev could be nil?  (or at least RevID won't be set)

db/blip_sync_context.go
```diff
@@ -617,8 +620,26 @@ func (bsc *BlipSyncContext) sendNoRev(sender *blip.Sender, docID, revID string,
...
+		rev, err = replacementRev, replacementRevErr
+	}
 	if base.IsDocNotFoundError(err) {
-		return bsc.sendNoRev(sender, docID, revID, collectionIdx, seq, err)
+		return bsc.sendNoRev(sender, docID, revID, replacedRevID, collectionIdx, seq, err)
```
adamcfraser: I'm not sure under what scenarios where we want to send replacedRevID on a noRev (are there any?), but IsDocNotFound seems like a clear case where we don't have a replacedRevID to send, but we could fall into this after also matching the previous sendReplacementRevs check (if line 635 returns not found).

adamcfraser (changes requested):

db/blip_sync_context.go
bbrks: Agreed. This is very poorly structured - I tried to reuse existing flow as much as possible but I think it actually complicates it more. Will redo replacement rev sending to be more explicit.

base/set.go
```diff
@@ -56,10 +56,29 @@ func (set Set) copy() Set {
...
-	return exists
+// Contains returns true if the set includes any of the given channels.
+func (set Set) Contains(ch ...string) bool {
+	for _, c := range ch {
+		if _, exists := set[c]; exists {
+			return true
+		}
+	}
+	return false
+}
```
bbrks, line 67: This change was used for a previous iteration but I think it's still a useful non-breaking change to make for existing callers. Can remove if this is contentious.

db/blip_sync_context.go
bbrks: Updated to be explicit and align with CBL behaviour for norev/error cases.

db/blip_collection_context.go
```diff
@@ -34,6 +34,10 @@ type blipSyncCollectionContext struct {
 	sgr2PushAlreadyKnownSeqsCallback func(alreadyKnownSeqs ...SequenceID)           // sgr2PushAlreadyKnownSeqsCallback is called to mark the sequence as being immediately processed
 	emptyChangesMessageCallback      func()                                         // emptyChangesMessageCallback is called when an empty changes message is received
 
+	// if we encounter a situation where a requested revision is not available, should we try sending an alternative as a replacement revision?
```
adamcfraser:
```suggestion
	// when a requested revision is not available, whether we should send the active revision as a replacement revision (depending on access)
```
If this change to the comment is accurate, I think it's more descriptive.

db/blip_sync_context.go
```diff
@@ -579,7 +579,7 @@ func (bsc *BlipSyncContext) sendBLIPMessage(sender *blip.Sender, msg *blip.Messa
 }
 
 func (bsc *BlipSyncContext) sendNoRev(sender *blip.Sender, docID, revID string, collectionIdx *int, seq SequenceID, err error) error {
-	base.DebugfCtx(bsc.loggingCtx, base.KeySync, "Sending norev %q %s due to unavailable revision: %v", base.UD(docID), revID, err)
+	base.TracefCtx(bsc.loggingCtx, base.KeySync, "Sending norev %q %s due to unavailable revision: %v", base.UD(docID), revID, err)
```
adamcfraser: Do we already log this information elsewhere, which is why we want to lower to Trace here?  I thought norevs were generally unexpected and merited the debug logging.

db/blip_sync_context.go
```diff
@@ -616,11 +616,42 @@ func (bsc *BlipSyncContext) sendNoRev(sender *blip.Sender, docID, revID string,
...
-	if base.IsDocNotFoundError(err) {
-		return bsc.sendNoRev(sender, docID, revID, collectionIdx, seq, err)
-	} else if err != nil {
-		return fmt.Errorf("failed to GetRev for doc %s with rev %s: %w", base.UD(docID).Redact(), base.MD(revID).Redact(), err)
+	rev, originalErr := handleChangesResponseCollection.GetRev(bsc.loggingCtx, docID, revID, true, nil)
+
+	// set if we find an alternative revision to send in the event the originally requested rev is unavailable
+	var replacedRevID string
+
+	collectionCtx, collectionErr := bsc.collections.get(collectionIdx)
```
adamcfraser: Should we move this inside the IsDocNotFoundError?

db/blip_sync_context.go
```diff
@@ -616,11 +616,42 @@ func (bsc *BlipSyncContext) sendNoRev(sender *blip.Sender, docID, revID string,
...
+
+	if base.IsDocNotFoundError(originalErr) {
+		if !collectionCtx.sendReplacementRevs {
+			return bsc.sendNoRev(sender, docID, revID, collectionIdx, seq, originalErr)
+		}
+
+		base.DebugfCtx(bsc.loggingCtx, base.KeySync, "Unavailable revision for %q %s - finding replacement: %v", base.UD(docID), revID, originalErr)
+
+		// try the active rev instead as a replacement
+		replacementRev, replacementRevErr := handleChangesResponseCollection.GetRev(bsc.loggingCtx, docID, "", true, nil)
```
adamcfraser, line 636: There's some inefficiency here, as we would have fetched the current version of the doc from the bucket when the revCacheLoader attempted to fetch the original (not found) rev.  

Since the frequency of replacementRev is expected to be low I think we can live with this, but could consider filing a backlog enhancement to enhance revCacheLoader to have the ability to return activeRev during the initial fetch (or at least stick it in the rev cache).

db/blip_sync_context.go
```diff
@@ -616,11 +616,42 @@ func (bsc *BlipSyncContext) sendNoRev(sender *blip.Sender, docID, revID string,
...
+		if replacementRevErr != nil {
+			base.DebugfCtx(bsc.loggingCtx, base.KeySync, "Active revision unavailable for %q - sending original norev for %q: %v", base.UD(docID), revID, replacementRevErr)
+			return bsc.sendNoRev(sender, docID, revID, collectionIdx, seq, originalErr)
+		}
+
+		// if this is a filtered replication, ensure the replacement rev is in one of the filtered channels
+		// normal channel access checks are already applied in GetRev above
+		if collectionCtx.channels != nil && replacementRev.Channels.NumMatches(collectionCtx.channels) == 0 {
```
adamcfraser: Did you consider a 'HasMatch' function that will exit as soon as it finds a match?  Might be useful to stop counting as soon as we see a match, in cases where the document or replication has a very large number of channels.

adamcfraser (commented):
Overall looks fine, a few questions on implementation details.

db/blip_sync_context.go
bbrks: Good catch. I had intended to move the debug logging up a level so we could get more contextual information about _why_ we're sending the norev.

Have covered all `sendNoRev` cases now with appropriate log messages.

db/blip_sync_context.go
bbrks: I'm not sure it's worth the complexity to cover this one case, and this is the _only_ case we'd need that behaviour.

Are there stats we should add now (`SendNoRevCount` and `SendReplacementRevCount`?) to determine the frequency of this codepath to justify doing this work in future?

adamcfraser: The stats do seem like they might be generally useful - both for understanding frequency and understanding what's going on in an environment.   I'll leave it to you whether to include in this PR or file a followup ticket.

adamcfraser (approved):

db/blip_sync_context.go
bbrks: Will do as an immediate follow-up/PR (CBG-3960)

PR #6831, opened 2024-05-16
CBG-3850: Optimise releaseUnusedSequenceRange 

gregns1:
CBG-3850

- Removed unused entries in log entry 
- Added sequence range element to log entry, only populated when entry is a sequence range
- releaseUnusedSequenceRange to handle sequence ranges in the way discussed in skipped sequence discussions 
- In addition to that refactor had to handle single range entries, to do this I have just run the sequence through processEntry and returned early
- _addPendingLogs has some extra handling in it for popping off pending a sequence range + disallowing the decrease of the next sequence expected at the cache
- PushSkipped has handling to disallow adding entries where startSeq > endSeq (I hit this in a few bug areas when testing) 
- Bunch of low level tests forcing cache in lots of different scenarios and calling releaseUnusedSequenceRange directly on certain ranges 
- removeSeqRange and removeSeq range needed checks for empty lists. I found during testing that if you call the binary search function on empty list, whist found is false it returns a non 0 index. This caused a index out of bound panic in few areas. 
- High level tests TestJumpInSequencesAtAllocatorSkippedSequenceFill + TestJumpInSequencesAtAllocatorRangeInPending which have some low level counter parts TestReleasedSequenceRangeHandlingEdgeCase1 + TestReleasedSequenceRangeHandlingEdgeCase2
- New functions _removeSubsetOfRangeFromSkipped + _pushRangeToPending to handle duplicate sequences between unused range and pending/skipped lists 

Doc for reviewers contains:
- Refresher on outcome of original meeting on this 
- Two scenarios I was stuck on for a while
- At bottom find the solution we went with to solve
https://docs.google.com/document/d/1SsQ2pjA_1nOZe-ZdvBpRzwJsY7ic0XnlQ_m2bMQ4jtc/edit?usp=sharing

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2459/

### DIFF
--- a/channels/log_entry.go
+++ b/channels/log_entry.go
@@ -17,17 +17,6 @@ import (
 	"time"
 )
 
-// LogEntry
-type LogEntryType uint8
-
-const (
-	LogEntryDocument = LogEntryType(iota)
-	LogEntryPrincipal
-	LogEntryCheckpoint
-	LogEntryRollback
-	LogEntryPurge
-)
-
 // Bits in LogEntry.Flags
 const (
 	Deleted  = 1 << iota // This rev is a deletion
@@ -40,30 +29,32 @@ const (
 )
 
 type LogEntry struct {
-	Sequence     uint64       // Sequence number
-	DocID        string       // Document ID
-	RevID        string       // Revision ID
-	Flags        uint8        // Deleted/Removed/Hidden flags
-	VbNo         uint16       // vbucket number
-	TimeSaved    time.Time    // Time doc revision was saved (just used for perf metrics)
-	TimeReceived time.Time    // Time received from tap feed
-	Channels     ChannelMap   // Channels this entry is in or was removed from
-	Skipped      bool         // Late arriving entry
-	Type         LogEntryType // Log entry type
-	Value        []byte       // Snapshot metadata (when Type=LogEntryCheckpoint)
-	PrevSequence uint64       // Sequence of previous active revision
-	IsPrincipal  bool         // Whether the log-entry is a tracking entry for a principal doc
-	CollectionID uint32       // Collection ID
+	Sequence      uint64               // Sequence number
+	DocID         string               // Document ID
+	RevID         string               // Revision ID
+	Flags         uint8                // Deleted/Removed/Hidden flags
+	TimeSaved     time.Time            // Time doc revision was saved (just used for perf metrics)
+	TimeReceived  time.Time            // Time received from tap feed
+	Channels      ChannelMap           // Channels this entry is in or was removed from
+	Skipped       bool                 // Late arriving entry
+	PrevSequence  uint64               // Sequence of previous active revision
+	IsPrincipal   bool                 // Whether the log-entry is a tracking entry for a principal doc
+	CollectionID  uint32               // Collection ID
+	SequenceRange *UnusedSequenceRange // Range of sequence that have been released by the sequence allocator
+}
+
+// UnusedSequenceRange will hold a range of sequences that have been released by the sequence allocator in log entry
+type UnusedSequenceRange struct {
+	StartSeq uint64
+	EndSeq   uint64
 }
 
 func (l LogEntry) String() string {
 	return fmt.Sprintf(
-		"seq: %d docid: %s revid: %s vbno: %d type: %v collectionID: %d",
+		"seq: %d docid: %s revid: %s collectionID: %d",
 		l.Sequence,
 		l.DocID,
 		l.RevID,
-		l.VbNo,
-		l.Type,
 		l.CollectionID,
 	)
 }
--- a/channels/log_entry.go
+++ b/channels/log_entry.go
@@ -29,24 +29,18 @@ const (
 )
 
 type LogEntry struct {
-	Sequence      uint64               // Sequence number
-	DocID         string               // Document ID
-	RevID         string               // Revision ID
-	Flags         uint8                // Deleted/Removed/Hidden flags
-	TimeSaved     time.Time            // Time doc revision was saved (just used for perf metrics)
-	TimeReceived  time.Time            // Time received from tap feed
-	Channels      ChannelMap           // Channels this entry is in or was removed from
-	Skipped       bool                 // Late arriving entry
-	PrevSequence  uint64               // Sequence of previous active revision
-	IsPrincipal   bool                 // Whether the log-entry is a tracking entry for a principal doc
-	CollectionID  uint32               // Collection ID
-	SequenceRange *UnusedSequenceRange // Range of sequence that have been released by the sequence allocator
-}
-
-// UnusedSequenceRange will hold a range of sequences that have been released by the sequence allocator in log entry
-type UnusedSequenceRange struct {
-	StartSeq uint64
-	EndSeq   uint64
+	Sequence     uint64     // Sequence number
+	EndSequence  uint64     // End sequence on range of sequences that have been released by the sequence allocator (0 if entry is single sequence)
+	DocID        string     // Document ID
+	RevID        string     // Revision ID
+	Flags        uint8      // Deleted/Removed/Hidden flags
+	TimeSaved    time.Time  // Time doc revision was saved (just used for perf metrics)
+	TimeReceived time.Time  // Time received from tap feed
+	Channels     ChannelMap // Channels this entry is in or was removed from
+	Skipped      bool       // Late arriving entry
+	PrevSequence uint64     // Sequence of previous active revision
+	IsPrincipal  bool       // Whether the log-entry is a tracking entry for a principal doc
+	CollectionID uint32     // Collection ID
 }
 
 func (l LogEntry) String() string {
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -612,71 +612,81 @@ func (c *changeCache) releaseUnusedSequenceRange(ctx context.Context, fromSequen
 		if err != nil {
 			base.DebugfCtx(ctx, base.KeyCache, err.Error())
 		}
-		// remove end seq on range, protecting against duplicate sequence scenarios
-		err = c.RemoveSkipped(toSequence)
-		if err != nil {
-			// end seq on range not present in skipped, so duplicate sequence
-			base.DebugfCtx(ctx, base.KeyCache, "  Ignoring duplicate of #%d (unusedSequence)", toSequence)
-		}
-	} else if fromSequence < c.nextSequence && toSequence > c.nextSequence {
-		// mixed scenario, remove from skipped and push to rest pending
-		err := c.RemoveSkippedSequences(ctx, fromSequence, c.nextSequence-1)
-		if err != nil {
-			base.DebugfCtx(ctx, base.KeyCache, err.Error())
-		}
-		// push next seq and above to pending
+		// we have sequences (possibly duplicate sequences) in the unused range that don't exist in skipped list
 		seqRange := &channels.UnusedSequenceRange{
-			StartSeq: c.nextSequence,
+			StartSeq: fromSequence,
 			EndSeq:   toSequence,
 		}
-		change := &LogEntry{
-			TimeReceived:  timeReceived,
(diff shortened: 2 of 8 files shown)


channels/log_entry.go
```diff
@@ -40,30 +29,32 @@ const (
...
+	RevID         string               // Revision ID
+	Flags         uint8                // Deleted/Removed/Hidden flags
+	TimeSaved     time.Time            // Time doc revision was saved (just used for perf metrics)
+	TimeReceived  time.Time            // Time received from tap feed
+	Channels      ChannelMap           // Channels this entry is in or was removed from
+	Skipped       bool                 // Late arriving entry
+	PrevSequence  uint64               // Sequence of previous active revision
+	IsPrincipal   bool                 // Whether the log-entry is a tracking entry for a principal doc
+	CollectionID  uint32               // Collection ID
+	SequenceRange *UnusedSequenceRange // Range of sequence that have been released by the sequence allocator
```
adamcfraser: It looks like SequenceRange.StartSeq always going to equal LogEntry.Sequence.  Instead of allocating memory twice for this, can we just populate an 'EndSeq' value on this struct to indicate ranges?

db/change_cache.go
```diff
@@ -577,31 +577,118 @@ func (c *changeCache) releaseUnusedSequence(ctx context.Context, sequence uint64
...
 		c.channelCache.AddUnusedSequence(change)
+		if c.notifyChange != nil {
+			c.notifyChange(ctx, allChangedChannels)
+		}
+		return
 	}
 
+	// acquire mutex on cache for reading/writing to pending + reading next sequence value
+	c.lock.Lock()
+	defer c.lock.Unlock()
```
adamcfraser, line 620: Are we holding the lock longer than we need to here?  At minimum I don't think we need to hold it while calling addUnusedSequence and notifyChange.  Maybe it would be clearer if this functionality was moved into something like a 'processUnusedRange()' function that covers the locking boundaries (i.e. the equivalent of processEntry for ranges).

db/skipped_sequence.go
```diff
@@ -364,3 +371,51 @@ func (s *SkippedSequenceSlice) getStats() SkippedSequenceStats {
 		ListLengthStat:                    int64(len(s.list)),
 	}
 }
+
+// _removeSubsetOfRangeFromSkipped will be called if we get error removing the unused sequence range indicating we have duplicate sequence in the unused sequence range
+// in the unused sequence range. This function will iterate through each skipped list element and remove any sequences within the unused sequence range
+func (s *SkippedSequenceSlice) _removeSubsetOfRangeFromSkipped(ctx context.Context, seqRange *channels.UnusedSequenceRange) error {
```
adamcfraser: The leading underscore implies that callers should already hold the lock, but this method actually does the locking.

db/change_cache.go
```diff
@@ -577,31 +577,118 @@ func (c *changeCache) releaseUnusedSequence(ctx context.Context, sequence uint64
...
+
+	if toSequence < c.nextSequence {
+		// batch remove from skipped
+		err := c.RemoveSkippedSequences(ctx, fromSequence, toSequence)
+		if err != nil {
```
adamcfraser: More than one type of error is returned by RemoveSkippedSequences - we should identify that this is actually the "range has sequences that aren't present in skipped" error.

db/change_cache_test.go
```diff
@@ -2300,3 +2300,723 @@ func TestSkippedSequenceCompact(t *testing.T) {
...
+	defer dbContext.Close(ctx)
+
+	err = dbContext.StartOnlineProcesses(ctx)
+	require.NoError(t, err)
+
+	ctx = dbContext.AddDatabaseLogContext(ctx)
+	testChangeCache := &changeCache{}
+	if err := testChangeCache.Init(ctx, dbContext, dbContext.channelCache, nil, &CacheOptions{
+		CachePendingSeqMaxWait: 2 * time.Minute,
+		CacheSkippedSeqMaxWait: 2 * time.Minute,
```
adamcfraser: If the test runs very slow (> 2 minute) for some reason, these values will break some of the assertions below (as things will start being ejected from pending and skipped.  Is there a reason to not bump these up to a larger value?

db/change_cache_test.go
```diff
@@ -2300,3 +2300,723 @@ func TestSkippedSequenceCompact(t *testing.T) {
...
+	}
+
+	if err := testChangeCache.Start(0); err != nil {
+		log.Printf("Start error for testChangeCache: %v", err)
+		t.Fail()
+	}
+	defer testChangeCache.Stop(ctx)
+	require.NoError(t, err)
+
+	// push two entries that will be pushed to pending and subsequently skipped
```
adamcfraser: These entries won't be skipped, will they?  The sequences preceding them will be skipped, and these will be cached.  Should the comment be updated?

db/change_cache.go
```diff
@@ -577,31 +577,118 @@ func (c *changeCache) releaseUnusedSequence(ctx context.Context, sequence uint64
...
+		seqRange := &channels.UnusedSequenceRange{
+			StartSeq: fromSequence,
+			EndSeq:   toSequence,
+		}
+		c._pushRangeToPending(ctx, seqRange, timeReceived)
+		// unblock any pending sequences we can after new range(s) have been pushed to pending
+		changedChannels := c._addPendingLogs(ctx)
+		allChangedChannels = allChangedChannels.Update(changedChannels)
+		c.internalStats.pendingSeqLen = len(c.pendingLogs)
+	} else {
```
adamcfraser, line 632: I think we want some context here to describe why this indicates a duplicate sequence.
```suggestion
	} else {
	   // An unused sequence range than includes c.nextSequence in the middle of the range
	   // isn't possible under normal processing - unused sequence ranges will normally be moved
	   // from pending to skipped in their entirety, as it's the processing of the pending sequence 
	   // *after* the range that triggers the range to be skipped.  A partial range in skipped means 
	   // an duplicate entry with a sequence within the bounds of the range was previously present 
	   // in pending.  
```

db/change_cache.go
```diff
@@ -577,31 +577,118 @@ func (c *changeCache) releaseUnusedSequence(ctx context.Context, sequence uint64
...
+		base.DebugfCtx(ctx, base.KeyCache, "unused sequence range of #%d to %d contains duplicate sequences", fromSequence, toSequence)
```
adamcfraser: Is there a reason we shouldn't split the unused sequence range at c.nextSequence, and send the first part through the skipped handling and the second part through the pending handling?

If we can't do that, I think the log message should be higher (warn or info), and should indicate that this sequence range won't be removed from skipped until they are abandoned.

db/change_cache.go
```diff
@@ -577,31 +577,118 @@ func (c *changeCache) releaseUnusedSequence(ctx context.Context, sequence uint64
...
+		err := c.RemoveSkippedSequences(ctx, fromSequence, toSequence)
+		if err != nil {
+			base.DebugfCtx(ctx, base.KeyCache, err.Error())
+			// we have sequences (possibly duplicate sequences) in the unused range that don't exist in skipped list
+			seqRange := &channels.UnusedSequenceRange{
+				StartSeq: fromSequence,
+				EndSeq:   toSequence,
+			}
+			// handle any potential duplicate sequences between the unused range and the skipped list
+			err = c.skippedSeqs._removeSubsetOfRangeFromSkipped(ctx, seqRange)
```
adamcfraser: Do we have to acquire two different locks on the skipped sequence queue to do this.  (one for the original RemoveSkippedSequences, and another for _removeSubsetOfRangeFromSkipped).  We want to avoid that performance overhead if possible, and also don't want to have to handle with cases where the skipped set changed between the two operations.

db/change_cache.go
```diff
@@ -577,31 +577,118 @@ func (c *changeCache) releaseUnusedSequence(ctx context.Context, sequence uint64
...
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	if toSequence < c.nextSequence {
+		// batch remove from skipped
+		err := c.RemoveSkippedSequences(ctx, fromSequence, toSequence)
+		if err != nil {
+			base.DebugfCtx(ctx, base.KeyCache, err.Error())
+			// we have sequences (possibly duplicate sequences) in the unused range that don't exist in skipped list
+			seqRange := &channels.UnusedSequenceRange{
```
adamcfraser: Is there benefit in building an UnusedSequenceRange, as opposed to just having startSeq and endSeq parameters on removeSubsetOfRangeFromSkipped?  It feels like we're cluttering up the code building these structs without any functional/performance benefit.

db/change_cache.go
```diff
@@ -577,31 +577,118 @@ func (c *changeCache) releaseUnusedSequence(ctx context.Context, sequence uint64
...
+			TimeReceived:  timeReceived,
+			Sequence:      seqRange.StartSeq,
+			SequenceRange: seqRange,
+		}
+		heap.Push(&c.pendingLogs, entry)
+		return
+	}
+
+	// check for duplicate sequences between range and pending logs
+	for _, v := range c.pendingLogs {
```
adamcfraser: c.pendingLogs can be large, and so this is an expensive way to check for intersections.  Since c.pendingLogs is sorted, can we instead use a more efficient search (e.g. sort.Find) to look for the first entry in pending that's greater than startSeq, and then check whether that value is less than endSeq?

adamcfraser (changes requested):
Have a look at the comments, we can discuss if I'm overlooking anything.

db/change_cache.go
gregns1: I have added this change to use sort.Find. Very good suggestion. I was trying to figure a way to do this same approach for skipped sequences in `_removeSubsetOfRangeFromSkipped` but I think this is difficult. But it may not be as much of a problem given we now have ranges in skipped list.

db/change_cache.go
gregns1: Addressed this, moving the skipped list processing for unused range into function processUnusedSequenceRangeAtSkipped that will acquire the lock, then call directly _removeSeqRange (and _removeSubsetOfRangeFromSkipped if necessary)

db/change_cache.go
gregns1: Its definitely possible to do so, my code before was doing that but removed it as I thought we cam to decision it shouldn't be possible for that scenario to present itself. Should we handle it anyway? 

I have raised the logging to warn for now but happy to make changes to split the ranges if you think its worth doing still?

db/change_cache_test.go
gregns1: Missed this, thanks!

db/change_cache_test.go
gregns1: Upped the time for all 2 minute waits to 20 minutes.

db/skipped_sequence.go
gregns1: Oversight from me, I re-wrote a number of these functions and must've missed this. 

I have now made changes that avoids acquiring the lock twice on skipped list (as pointed out in separate comment below) so now this function is correctly being called by another acquiring the lock.

db/change_cache.go
gregns1: I agree, oversight on my part. Have moved the locking code into separate function.

channels/log_entry.go
gregns1: Very good point, updated to reflect this. Added end sequence to log entry and if its 0 we can assume the log entry itself is single sequence.

db/change_cache.go
gregns1: I have now added a similar approach in `_removeSubsetOfRangeFromSkipped` using sort.Find to find minimum index in skipped list that has overlap in the unused sequence range, then iterate from that point on (avoiding iterating through more of the skipped list than we need to). Let me know what you think, it doesn't seem a clean solution to me but its what I could think of at this time.

adamcfraser (approved):

PR #6834, opened 2024-05-17
CBG-3899 Write _mou on import, resync

adamcfraser:
Writes _mou on import and resync, populating cas and pCas.  If an imported or resync'd document already has _mou where doc.cas=_mou.cas, updates _mou.cas but preserves existing pCas.

Stores MetadataOnlyUpdate struct in document but not SyncData, to avoid unmarshalling work when mou isn't needed.

CBG-3899
- [x] `GSI=true,xattrs=true` 7.6.1 https://jenkins.sgwdev.com/job/SyncGateway-Integration/2445/
- [x] `GSI=true,xattrs=true` 7.2.0 https://jenkins.sgwdev.com/job/SyncGateway-Integration/2452/

### DIFF
--- a/db/background_mgr_resync_dcp_test.go
+++ b/db/background_mgr_resync_dcp_test.go
@@ -17,6 +17,7 @@ import (
 	"testing"
 	"time"
 
+	sgbucket "github.com/couchbase/sg-bucket"
 	"github.com/couchbase/sync_gateway/base"
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
@@ -453,6 +454,11 @@ func TestResyncMou(t *testing.T) {
 	base.SetUpTestLogging(t, base.LevelInfo, base.KeyMigrate, base.KeyImport)
 	db, ctx := setupTestDBWithOptionsAndImport(t, nil, DatabaseContextOptions{})
 	defer db.Close(ctx)
+
+	if !db.Bucket.IsSupported(sgbucket.BucketStoreFeatureMultiXattrSubdocOperations) {
+		t.Skip("Test requires multi-xattr subdoc operations, CBS 7.6 or higher")
+	}
+
 	db.Options.QueryPaginationLimit = 100 // Required for principal ID query to not deadlock
 
 	initialImportCount := db.DbStats.SharedBucketImport().ImportCount.Value()
--- a/db/background_mgr_resync_dcp_test.go
+++ b/db/background_mgr_resync_dcp_test.go
@@ -509,13 +509,13 @@ function sync(doc, oldDoc){
 	err = resyncMgr.Start(ctx, options)
 	require.NoError(t, err)
 
-	err = WaitForConditionWithOptions(t, func() bool {
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
 		var status BackgroundManagerStatus
-		rawStatus, _ := resyncMgr.GetStatus(ctx)
-		_ = json.Unmarshal(rawStatus, &status)
-		return status.State == BackgroundProcessStateCompleted
-	}, 200, 200)
-	require.NoError(t, err)
+		rawStatus, err := resyncMgr.GetStatus(ctx)
+		assert.NoError(c, err)
+		assert.NoError(c, json.Unmarshal(rawStatus, &status))
+		assert.Equal(c, BackgroundProcessStateCompleted, status.State)
+	}, 40*time.Second, 200*time.Millisecond)
 
 	stats := getResyncStats(resyncMgr.Process)
 	assert.Equal(t, int64(2), stats.DocsChanged)
--- a/db/background_mgr_resync_dcp_test.go
+++ b/db/background_mgr_resync_dcp_test.go
@@ -444,6 +444,90 @@ function sync(doc, oldDoc){
 	return db, ctx
 }
 
+// TestResyncMou ensures that resync updates create mou, and preserve pcas in mou in the case where resync is reprocessing an import
+func TestResyncMou(t *testing.T) {
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("Test requires Couchbase Server")
+	}
+
+	base.SetUpTestLogging(t, base.LevelInfo, base.KeyMigrate, base.KeyImport)
+	db, ctx := setupTestDBWithOptionsAndImport(t, nil, DatabaseContextOptions{})
+	defer db.Close(ctx)
+	db.Options.QueryPaginationLimit = 100 // Required for principal ID query to not deadlock
+
+	initialImportCount := db.DbStats.SharedBucketImport().ImportCount.Value()
+	collection := GetSingleDatabaseCollectionWithUser(t, db)
+	docBody := Body{"foo": "bar"}
+
+	// Create a document via SGW.  mou should not be updated
+	_, doc, err := collection.Put(ctx, "sgWrite", docBody)
+	require.NoError(t, err)
+	sgWriteCas := doc.Cas
+
+	syncData, mou, _ := getSyncAndMou(t, collection, "sgWrite")
+	require.NotNil(t, syncData)
+	require.Nil(t, mou)
+
+	// 2. Create via the SDK
+	_, err = collection.dataStore.WriteCas("sdkWrite", 0, 0, docBody, 0)
+	require.NoError(t, err)
+
+	base.RequireWaitForStat(t, func() int64 {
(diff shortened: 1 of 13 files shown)


bbrks (dismissed):
LGTM - will let @torcolvin have a pass

db/background_mgr_resync_dcp_test.go
```diff
@@ -444,6 +445,93 @@ function sync(doc, oldDoc){
...
+
+	base.SetUpTestLogging(t, base.LevelInfo, base.KeyMigrate, base.KeyImport)
+	db, ctx := setupTestDBWithOptionsAndImport(t, nil, DatabaseContextOptions{})
+	defer db.Close(ctx)
+
+	if !db.Bucket.IsSupported(sgbucket.BucketStoreFeatureMultiXattrSubdocOperations) {
+		t.Skip("Test requires multi-xattr subdoc operations, CBS 7.6 or higher")
+	}
+
+	db.Options.QueryPaginationLimit = 100 // Required for principal ID query to not deadlock
```
torcolvin, line 462: I this from copypasta? I don't even see what would do query, so I imagine it might be from query based resync?

db/background_mgr_resync_dcp_test.go
```diff
@@ -444,6 +445,93 @@ function sync(doc, oldDoc){
...
+
+	err = resyncMgr.Start(ctx, options)
+	require.NoError(t, err)
+
+	err = WaitForConditionWithOptions(t, func() bool {
+		var status BackgroundManagerStatus
+		rawStatus, _ := resyncMgr.GetStatus(ctx)
+		_ = json.Unmarshal(rawStatus, &status)
+		return status.State == BackgroundProcessStateCompleted
+	}, 200, 200)
```
torcolvin: just to make this easier to debug for the future, so it will print the status if it does fail:

```suggestion
	require.EventuallyWithT(t, func(c *assert.CollectT) {
		var status BackgroundManagerStatus
		rawStatus, err := resyncMgr.GetStatus(ctx)
		assert.NoError(t, err)
		assert.Equal(t, json.Unmarshal(rawStatus, &status)
		assert.Equal(t, BackgroundProcessStateComplete,  status.State)
	}, 40 * time.Second, 200 * time.Millisecond)
```

db/background_mgr_resync_dcp_test.go
```diff
@@ -444,6 +445,93 @@ function sync(doc, oldDoc){
 	return db, ctx
 }
 
+// TestResyncMou ensures that resync updates create mou, and preserve pcas in mou in the case where resync is reprocessing an import
```
torcolvin, line 448: This would need a separate test and I'm not sure that this is common, but if you have resync running without auto import, this might be worth testing an inline {{_sync}} attribute.

db/change_cache.go
```diff
@@ -401,6 +401,7 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent) {
 	}
 
 	// If using xattrs and this isn't an SG write, we shouldn't attempt to cache.
+	rawUserXattr := rawXattrs[collection.userXattrKey()]
```
torcolvin, line 404: nit:

I'd drop this assignment into this if statement since it isn't needed outside this block

db/import_test.go
```diff
@@ -24,6 +24,55 @@ import (
...
+	xattrs, importCas, err := collection.dataStore.GetXattrs(ctx, key, []string{base.SyncXattrName, base.MouXattrName})
+	require.NoError(t, err)
+	syncXattr, ok := xattrs[base.SyncXattrName]
+	require.True(t, ok)
+	require.NoError(t, base.JSONUnmarshal(syncXattr, &syncData))
+	require.NotZero(t, syncData.Sequence, "Sequence should not be zero for imported doc")
+
+	// verify mou contents
+	if db.Bucket.IsSupported(sgbucket.BucketStoreFeatureMultiXattrSubdocOperations) {
+		mouXattr, ok := xattrs[base.MouXattrName]
```
torcolvin: Is it worth making sure this xattr is _not_ present?

db/document.go
```diff
@@ -389,19 +395,32 @@ func unmarshalDocument(docid string, data []byte) (*Document, error) {
...
-		err = doc.UnmarshalWithXattr(ctx, data, xattrData, unmarshalLevel)
+		err = doc.UnmarshalWithXattr(ctx, data, syncXattrData, unmarshalLevel)
 	}
 	if err != nil {
 		return nil, err
 	}
 
+	if len(mouXattrData) > 0 {
+		if err := base.JSONUnmarshal(mouXattrData, &doc.metadataOnlyUpdate); err != nil {
+			if err != nil {
```
torcolvin: this is an extra err check here?

db/document.go
```diff
@@ -389,19 +395,32 @@ func unmarshalDocument(docid string, data []byte) (*Document, error) {
...
+				base.WarnfCtx(ctx, "Failed to unmarshal mouXattr for key %v, will be ignored. Err: %v mou:%s", base.UD(docid), err, mouXattrData)
```
torcolvin: I'd clarify that only the `_mou` is ignored and not unmarshalling the document

db/document.go
```diff
@@ -432,36 +451,35 @@ func UnmarshalDocumentSyncData(data []byte, needHistory bool) (*SyncData, error)
...
-			return nil, nil, nil, nil, err
+			return nil, nil, nil, err
 		}
-		rawSyncXattr = xattrs[base.SyncXattrName]
-		rawUserXattr = xattrs[userXattrKey]
 
+		rawSyncXattr, ok := xattrValues[base.SyncXattrName]
 		// If the sync xattr is present, use that to build SyncData
-		if len(rawSyncXattr) > 0 {
+		if ok && len(rawSyncXattr) > 0 {
```
torcolvin: I don't see how this could be !ok and `len(rawSyncXattr) > 0` , is there something I'm missing?

db/import_test.go
```diff
@@ -24,6 +24,55 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestFeedImport(t *testing.T) {
```
torcolvin, line 27: I think this test should test for inline sync metadata, which I think should come with `_mou`, from the mou from after migrateMetadata.

I think without this that eventing will do a single echo, once after migrate metadata, and once after sync comes in. 

My best guess as to how these occur in a real world is a backup/restore.

I also think that potentially there should be a test that only does on demand imports. This might not be a big deal, but we see on demand imports when there are rapid number of writes. It probably won't echo much since there will be an immediate write with different body immediately after.

db/document.go
```diff
@@ -1133,16 +1151,40 @@ func (doc *Document) MarshalWithXattr() (data []byte, xdata []byte, err error) {
...
+		}
+	}
+
+	return data, syncXattr, mouXattr, nil
+}
+
+// computeMetadataOnlyUpdate computes a new metadataOnlyUpdate based on the existing document's CAS and metadataOnlyUpdate
+func computeMetadataOnlyUpdate(currentCas uint64, currentMou *MetadataOnlyUpdate) *MetadataOnlyUpdate {
+	var prevCas string
+	currentCasString := string(base.Uint64CASToLittleEndianHex(currentCas))
```
```suggestion
	currentCasString := base.CasToString(currentCas)
```

torcolvin (commented):
Apologizes if I didn't reason correctly about on demand imports, but I thought it was worth testing at least.

db/background_mgr_resync_dcp_test.go
adamcfraser: It's like the comment says - otherwise the principal ID query will deadlock when QueryPaginationLimit is zero, and we normally initialize this outside the db package.  Possibly worth cleanup, but not in this PR I don't think.

db/change_cache.go
adamcfraser: It's used below (line 497).

db/document.go
adamcfraser: I think it was just my instinct to be more idiomatic and check the ok, but you're right that the length check is sufficient, and any performance difference is negligible.

db/import_test.go
adamcfraser: Since migrate isn't a metadata-only update (it modifies the document body) and doesn't modify _mou, I didn't think additional test cases for specifically for migrate (inline sync metadata) were necessary.

I agree about the on-demand import case, I'll add that.

torcolvin (approved):

PR #6835, opened 2024-05-20
CBG-3925: Add FileLogger RotationInterval

bbrks:
CBG-3925

- Add `FileLogger` `RotationInterval` option to force a periodic log rotation.
- Interval is opt-in and can be set anywhere between 15 minutes and 7 days.
- Logs will rotate even if no log lines have been written, producing an empty file. This makes it very obvious that log coverage is present and empty, rather than just missing.

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2465/

### DIFF
--- a/base/logger_file.go
+++ b/base/logger_file.go
@@ -226,7 +226,7 @@ func (lfc *FileLoggerConfig) init(ctx context.Context, level LogLevel, name stri
 		rotationTickerCh = rotationTicker.C
 	}
 
-	logDeletionTicker := time.NewTicker(defaultLogDeletionInterval)
+	logDeletionTicker := time.NewTicker(rotatedLogDeletionInterval)
 	go func() {
 		defer func() {
 			if panicked := recover(); panicked != nil {
--- a/base/logger_file.go
+++ b/base/logger_file.go
@@ -76,7 +76,7 @@ type logRotationConfig struct {
 	LocalTime            *bool           `json:"localtime,omitempty"`               // If true, it uses the computer's local time to format the backup timestamp.
 	RotatedLogsSizeLimit *int            `json:"rotated_logs_size_limit,omitempty"` // Max Size of log files before deletion
 	RotationInterval     *ConfigDuration `json:"rotation_interval,omitempty"`       // Interval at which logs are rotated
-	Compress             *bool           `json:"-"`                                 // Enable log compression
+	Compress             *bool           `json:"-"`                                 // Enable log compression - not exposed in config
 }
 
 // NewFileLogger returns a new FileLogger from a config.
--- a/base/logger_file.go
+++ b/base/logger_file.go
@@ -111,7 +111,7 @@ func NewFileLogger(ctx context.Context, config *FileLoggerConfig, level LogLevel
 		logger.collateBufferWg = &sync.WaitGroup{}
 
 		// Start up a single worker to consume messages from the buffer
-		go logCollationWorker(logger.collateBuffer, logger.flushChan, logger.collateBufferWg, logger.logger, *config.CollationBufferSize, fileLoggerCollateFlushTimeout)
+		go logCollationWorker(ctx, logger.collateBuffer, logger.flushChan, logger.collateBufferWg, logger.logger, *config.CollationBufferSize, fileLoggerCollateFlushTimeout)
 	}
 
 	return logger, nil
@@ -140,6 +140,13 @@ func (l *FileLogger) Rotate() error {
 	}
 
 	return errors.New("can't rotate non-lumberjack log output")
+}
+
+func (l *FileLogger) Close() error {
+	if c, ok := l.output.(io.Closer); ok {
+		return c.Close()
+	}
+	return nil
 }
 
 func (l *FileLogger) String() string {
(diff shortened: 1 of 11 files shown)


bbrks:
Might need additional synchronization for Windows to allow the temp directory to be removed. On Linux you can remove an in-use file without error.

```
2024-05-23T15:26:28.3933228Z     testing.go:1231: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\TestLogRotationInterval2971432524\001\sg_test.log: The process cannot access the file because it is being used by another process.
2024-05-23T15:26:28.3933403Z --- FAIL: TestLogRotationInterval (2.24s)
```

base/logger_file.go
```diff
@@ -205,20 +219,39 @@ func (lfc *FileLoggerConfig) init(ctx context.Context, level LogLevel, name stri
...
 			select {
-			case <-ticker.C:
+			case <-ctx.Done():
+				return
+			case <-logDeletionTicker.C:
 				err := runLogDeletion(ctx, logFilePath, level.String(), int(float64(*lfc.Rotation.RotatedLogsSizeLimit)*rotatedLogsLowWatermarkMultiplier), *lfc.Rotation.RotatedLogsSizeLimit)
 				if err != nil {
 					WarnfCtx(ctx, "%s", err)
 				}
+			case <-rotationTickerCh:
```
torcolvin, line 249: nit: I think this reads just a bit easier to see `rotationTicker.C` since I had to look a back to see if this was a normal ticker.

docs/api/components/schemas.yaml
```diff
@@ -2317,6 +2317,10 @@ Log-rotation-config-readonly:
       description: Max Size (in mb) of log files before deletion
       type: integer
       default: 1024
+    rotation_interval:
+      description: If set, the interval at which log files are rotated, even if max_size is not reached.
```
torcolvin: I would add that these are units of hours?

rest/config_flags.go
```diff
@@ -72,6 +72,7 @@ func registerConfigFlags(config *StartupConfig, fs *flag.FlagSet) map[string]con
 		"logging.console.rotation.max_age":                 {&config.Logging.Console.Rotation.MaxAge, fs.Int("logging.console.rotation.max_age", 0, "")},
 		"logging.console.rotation.localtime":               {&config.Logging.Console.Rotation.LocalTime, fs.Bool("logging.console.rotation.localtime", false, "")},
 		"logging.console.rotation.rotated_logs_size_limit": {&config.Logging.Console.Rotation.RotatedLogsSizeLimit, fs.Int("logging.console.rotation.rotated_logs_size_limit", 0, "")},
+		"logging.console.rotation.rotation_interval":       {&config.Logging.Console.Rotation.RotationInterval, fs.String("logging.console.rotation.rotation_interval", "", "")},
```
torcolvin, line 75: Is it worth adding some tests in `TestFillConfigWithFlagsValidVals` that the default will end up being 0 and this value is correctly populated?

base/logging_test.go
```diff
@@ -57,6 +59,71 @@ func Benchmark_LoggingPerformance(b *testing.B) {
...
+	// On Windows, cleanup of t.TempDir() fails due to open log file handle from Lumberjack. Cannot be fixed from SG.
+	// https://github.com/natefinch/lumberjack/issues/185
+	var logPath string
+	if runtime.GOOS == "windows" {
+		var err error
+		logPath, err = os.MkdirTemp("", t.Name())
+		require.NoError(t, err)
+		t.Cleanup(func() {
+			if err := os.RemoveAll(logPath); err != nil {
+				// log instead of error because it's likely this is going to fail on Windows for this test.
```
torcolvin, line 89:
```suggestion
				// log instead of error because it's likely this is going to fail on Windows for this test if lumberjack has not finished rotation. The `time.Sleep` at the end of the tests tries to mitigate this.
```

I wonder if the windows build machines are set up to clear temporary files, this could end up filing up the disks on the machine. We could consider writing to cwd if the builds do a fresh checkpoint of sync_gateway or do a `git clean -fdx`

Any reason here to use `os.MkdirTemp` instead of `t.TempDir` ?

torcolvin (commented):
LGTM, just minor comments.

base/logger_file.go
bbrks: I ended up writing it this way because `rotationTicker` can be nil if you haven't enabled a rotation interval. The select panics trying to access `C` on a nil ticker.

You _can_ however, safely select over a nil channel, and this avoids having to write a separate conditional select if you have log rotation enabled.

docs/api/components/schemas.yaml
bbrks: It's a duration string so can take any Go duration unit. I'll copy/paste the units description from `min_config_fetch_interval`

rest/config_flags.go
bbrks: Default will be `nil`, but yes I can add some unset values to assert on in there.

base/logging_test.go
bbrks: `t.TempDir` fails because it asserts there was no error from `RemoveAll`. I originally wrote it like that.

bbrks: I battled with this for a day and ended up on this solution. The sleep doesn't prevent this error, because Lumberjack keeps the current file handle open for some reason, even after calling `Close()`

On *nix this doesn't matter because the OS will still let us remove the file/directory.

torcolvin (approved):

PR #6837, opened 2024-05-21
CBG-3839: handle rollback of the _sync:seq document in the bucket

gregns1:
CBG-3839

- After incoming the sync:seq doc in incrementSequence function we can check if the result is less then we are expecting, thus determine the _sync:sequence must’ve been rolled back. 
- Added function _fixSyncSeqRollback for this to handle this
- Need the minimum correction stop be the sequence batch size as when having multiple nodes you can have overlapping sequence batches
- Have testing for single and multi node handling

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2519/

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -542,6 +542,11 @@ func NewDatabaseContext(ctx context.Context, dbName string, bucket base.Bucket, 
 	if err != nil {
 		return nil, err
 	}
+	clust, err := dbContext.SGReplicateMgr.GetSGRCluster()
+	if err != nil {
+		return nil, err
+	}
+	dbContext.sequences.nodes = clust.Nodes
 
 	if dbContext.UseQueryBasedResyncManager() {
 		dbContext.ResyncManager = NewResyncManager(metadataStore, metaKeys)
--- a/db/database.go
+++ b/db/database.go
@@ -542,11 +542,6 @@ func NewDatabaseContext(ctx context.Context, dbName string, bucket base.Bucket, 
 	if err != nil {
 		return nil, err
 	}
-	clust, err := dbContext.SGReplicateMgr.GetSGRCluster()
-	if err != nil {
-		return nil, err
-	}
-	dbContext.sequences.nodes = clust.Nodes
 
 	if dbContext.UseQueryBasedResyncManager() {
 		dbContext.ResyncManager = NewResyncManager(metadataStore, metaKeys)
--- a/db/sequence_allocator.go
+++ b/db/sequence_allocator.go
@@ -36,7 +36,7 @@ const (
 	idleBatchSize = 1
 
 	// minimum number that _sync:seq will be corrected by in event of a rollback
-	// assumes a SG cluster size of 50 nodes each allocating a full batch size of 10
+	// Accounts for a maximum SG cluster size of 50 nodes, each allocating a full batch size of 10
 	// if this value is too low and this correction has potential to allocate sequences that other nodes have already reserved a batch for
 	syncSeqCorrectionValue = 500
 )
@@ -249,13 +249,10 @@ func (s *sequenceAllocator) nextSequenceGreaterThan(ctx context.Context, existin
 	}
 
 	// check for rollback of _sync:seq before continuing
-	expectedValue := incrVal + prevAllocReleaseTo
-	if allocatedToSeq < expectedValue {
+	minimumExpectedValue := incrVal + prevAllocReleaseTo
+	if allocatedToSeq < minimumExpectedValue {
 		// rollback of _sync:seq detected
-		base.ErrorfCtx(ctx, "rollback of _sync:seq document detected. Allocated to %d but expected value of at least %d", allocatedToSeq, expectedValue)
-		// find diff between current _sync:seq value and what we expected it to be + correction value
-		correctionIncrValue := (expectedValue - allocatedToSeq) + syncSeqCorrectionValue
-		allocatedToSeq, err = s._fixSyncSeqRollback(ctx, correctionIncrValue, expectedValue)
+		allocatedToSeq, err = s._fixSyncSeqRollback(ctx, allocatedToSeq, minimumExpectedValue)
 		if err != nil {
 			s.mutex.Unlock()
 			return 0, 0, err
@@ -327,14 +324,10 @@ func (s *sequenceAllocator) _reserveSequenceBatch(ctx context.Context) error {
 	}
 
 	// check for rollback of _sync:seq document
-	expectedValue := s.max + s.sequenceBatchSize
-	if max < expectedValue {
+	minimumExpectedValue := s.max + s.sequenceBatchSize
+	if max < minimumExpectedValue {
 		// rollback of _sync:seq detected
-		base.ErrorfCtx(ctx, "rollback of _sync:seq document detected. Allocated to %d but expected value of at least %d", max, expectedValue)
-		// find diff between current _sync:seq value and what we expected it to be + correction value
-		correctionIncrValue := (expectedValue - max) + syncSeqCorrectionValue
-		// we must have correction value of at least the sequence batch size else we risk overlapping batches between nodes
-		max, err = s._fixSyncSeqRollback(ctx, correctionIncrValue, expectedValue)
+		max, err = s._fixSyncSeqRollback(ctx, max, minimumExpectedValue)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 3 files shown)


db/sequence_allocator.go
```diff
@@ -235,13 +235,26 @@ func (s *sequenceAllocator) nextSequenceGreaterThan(ctx context.Context, existin
...
 	if err != nil {
 		base.WarnfCtx(ctx, "Error from incrementSequence in nextSequenceGreaterThan(%d): %v", existingSequence, err)
 		s.mutex.Unlock()
 		return 0, 0, err
 	}
 
+	// check for rollback of _sync:seq before continuing
+	if allocatedToSeq < (incrVal + prevAllocReleaseTo) {
+		// rollback of _sync:seq detected
+		base.WarnfCtx(ctx, "rollback of _sync:seq document detected")
```
bbrks: Is it going to be useful to know the values we passed in and got back?

db/sequence_allocator.go
```diff
@@ -235,13 +235,26 @@ func (s *sequenceAllocator) nextSequenceGreaterThan(ctx context.Context, existin
...
+		// find diff between current _sync:seq value and what we expected it to be
+		correctionIncrValue := (incrVal + prevAllocReleaseTo) - allocatedToSeq
+		allocatedToSeq, err = s._fixSyncSeqRollback(ctx, correctionIncrValue)
+		if err != nil {
+			return 0, 0, err
```
bbrks, line 258: I think this case needs to unlock the mutex like above

db/sequence_allocator.go
```diff
@@ -306,6 +319,19 @@ func (s *sequenceAllocator) _reserveSequenceBatch(ctx context.Context) error {
 		return err
 	}
 
+	// check for rollback of _sync:seq document
+	if max < (s.max + s.sequenceBatchSize) {
+		// rollback of _sync:seq detected
+		base.WarnfCtx(ctx, "rollback of _sync:seq document detected")
+		// find diff between current _sync:seq value and what we expected it to be
+		correctionIncrValue := (s.max + s.sequenceBatchSize) - max
```
bbrks: Does the logic inside `_fixSyncSeqRollback` do this already?

bbrks (changes requested):

db/sequence_allocator.go
```diff
@@ -381,3 +407,20 @@ func (s *sequenceAllocator) waitForReleasedSequences(ctx context.Context, startT
 	time.Sleep(requiredWait)
 	return requiredWait
 }
+
+// _fixSyncSeqRollback will correct a rolled back _sync:seq document in the bucket
+func (s *sequenceAllocator) _fixSyncSeqRollback(ctx context.Context, incrValue uint64) (allocatedToSeq uint64, err error) {
+	// we must have correction value of at least the sequence batch size else we risk overlapping batches between nodes
+	if incrValue < s.sequenceBatchSize {
+		// take correction value up to minimum value of the sequence batch
+		incrValue += s.sequenceBatchSize - incrValue
```
adamcfraser: This would be simpler as incrValue = s.sequenceBatchSize?

db/sequence_allocator.go
```diff
@@ -306,6 +319,19 @@ func (s *sequenceAllocator) _reserveSequenceBatch(ctx context.Context) error {
 		return err
 	}
 
+	// check for rollback of _sync:seq document
+	if max < (s.max + s.sequenceBatchSize) {
+		// rollback of _sync:seq detected
+		base.WarnfCtx(ctx, "rollback of _sync:seq document detected")
```
adamcfraser: As with Ben's comment earlier, we should include the sequence values (max, s.max, batchSize) in the warning message.

db/sequence_allocator.go
```diff
@@ -381,3 +407,20 @@ func (s *sequenceAllocator) waitForReleasedSequences(ctx context.Context, startT
 	time.Sleep(requiredWait)
 	return requiredWait
 }
+
+// _fixSyncSeqRollback will correct a rolled back _sync:seq document in the bucket
+func (s *sequenceAllocator) _fixSyncSeqRollback(ctx context.Context, incrValue uint64) (allocatedToSeq uint64, err error) {
+	// we must have correction value of at least the sequence batch size else we risk overlapping batches between nodes
```
adamcfraser: I don't think there's a situation where we would end up with overlapping batches (and batch size isn't intended to handle that, batch size should just be a performance consideration on this node).  Can you clarify what's meant here?

adamcfraser (changes requested):

db/sequence_allocator.go
```diff
@@ -34,6 +34,9 @@ const (
 
 	// Idle batch size.  Initial batch size if SG is in an idle state (with respect to writes)
 	idleBatchSize = 1
+
+	//minimum number that _sync:seq will be corrected by in event of a rollback
```
bbrks: Might be worth adding the commentary about how we settled on this number for future reference.

```suggestion
	// minimum number that _sync:seq will be corrected by in event of a rollback
	// assumes a SG cluster size of 50 nodes each allocating a full batch size of 10
	// if this value is too low and this correction has potential to allocate sequences that other nodes have already reserved a batch for
```

db/database.go
```diff
@@ -540,6 +540,11 @@ func NewDatabaseContext(ctx context.Context, dbName string, bucket base.Bucket,
 	if err != nil {
 		return nil, err
 	}
+	clust, err := dbContext.SGReplicateMgr.GetSGRCluster()
+	if err != nil {
+		return nil, err
+	}
+	dbContext.sequences.nodes = clust.Nodes
```
bbrks: Don't think we need this any more?

db/sequence_allocator.go
```diff
@@ -381,3 +414,69 @@ func (s *sequenceAllocator) waitForReleasedSequences(ctx context.Context, startT
...
+			return 0, err
+		}
+	}
+
+	return allocatedToSeq, err
+}
+
+// shouldRetrySyncSeqCorrection returns true if value of _sync:seq is less than the expected value, else returns false
+func shouldRetrySyncSeqCorrection(expectedVal, actualVal uint64) bool {
+	return actualVal <= expectedVal
```
bbrks: This seems like it should be `<` but I think I also need a closer look at the code inside `_fixSyncSeqRollback` as well - don't have time for a complete review of that today.

bbrks (changes requested):
Still need to review logic inside `_fixSyncSeqRollback` but the initial things I saw

db/database.go
gregns1: Apologies missed removing this

db/sequence_allocator.go
```diff
@@ -34,6 +34,11 @@ const (
 
 	// Idle batch size.  Initial batch size if SG is in an idle state (with respect to writes)
 	idleBatchSize = 1
+
+	// minimum number that _sync:seq will be corrected by in event of a rollback
+	// assumes a SG cluster size of 50 nodes each allocating a full batch size of 10
```
adamcfraser:
```suggestion
	// Accounts for a maximum SG cluster size of 50 nodes, each allocating a full batch size of 10
```

db/sequence_allocator.go
```diff
@@ -235,13 +240,28 @@ func (s *sequenceAllocator) nextSequenceGreaterThan(ctx context.Context, existin
...
 		base.WarnfCtx(ctx, "Error from incrementSequence in nextSequenceGreaterThan(%d): %v", existingSequence, err)
 		s.mutex.Unlock()
 		return 0, 0, err
 	}
 
+	// check for rollback of _sync:seq before continuing
+	expectedValue := incrVal + prevAllocReleaseTo
+	if allocatedToSeq < expectedValue {
+		// rollback of _sync:seq detected
+		base.ErrorfCtx(ctx, "rollback of _sync:seq document detected. Allocated to %d but expected value of at least %d", allocatedToSeq, expectedValue)
```
adamcfraser: This should be logged at warn - error should be reserved for non-recoverable scenarios.

db/sequence_allocator.go
```diff
@@ -306,6 +326,20 @@ func (s *sequenceAllocator) _reserveSequenceBatch(ctx context.Context) error {
 		return err
 	}
 
+	// check for rollback of _sync:seq document
+	expectedValue := s.max + s.sequenceBatchSize
```
adamcfraser:
```suggestion
	minimumExpectedValue := s.max + s.sequenceBatchSize
```

db/sequence_allocator.go
```diff
@@ -306,6 +326,20 @@ func (s *sequenceAllocator) _reserveSequenceBatch(ctx context.Context) error {
...
+	if max < expectedValue {
+		// rollback of _sync:seq detected
+		base.ErrorfCtx(ctx, "rollback of _sync:seq document detected. Allocated to %d but expected value of at least %d", max, expectedValue)
+		// find diff between current _sync:seq value and what we expected it to be + correction value
+		correctionIncrValue := (expectedValue - max) + syncSeqCorrectionValue
```
adamcfraser: It feels like these four lines (332-335) should be moved into fixSyncSeqRollback (and the same for other uses).

db/sequence_allocator.go
```diff
@@ -381,3 +415,69 @@ func (s *sequenceAllocator) waitForReleasedSequences(ctx context.Context, startT
 	time.Sleep(requiredWait)
 	return requiredWait
 }
+
+// _fixSyncSeqRollback will correct a rolled back _sync:seq document in the bucket
+func (s *sequenceAllocator) _fixSyncSeqRollback(ctx context.Context, incrValue, expectedValue uint64) (allocatedToSeq uint64, err error) {
+	// grab _sync:seq value + its current cas value
+	var result uint64
+	cas, err := s.datastore.Get(s.metaKeys.SyncSeqKey(), &result)
```
adamcfraser: It looks like the retry flow here is:
1. Get _sync:seq value + cas
2. If value < targetValue, WriteCas to update
  2.1. On cas failure, return to 1
3. On success, incr(batchsize)

Can we move initial Get inside the worker to simplify this function, and to ensure we only do 3 in one place?

db/sequence_allocator.go
```diff
@@ -381,3 +415,69 @@ func (s *sequenceAllocator) waitForReleasedSequences(ctx context.Context, startT
...
+		if err != nil {
+			return 0, err
+		}
+	}
+
+	return allocatedToSeq, err
+}
+
+// shouldRetrySyncSeqCorrection returns true if value of _sync:seq is less than the expected value, else returns false
+func shouldRetrySyncSeqCorrection(expectedVal, actualVal uint64) bool {
```
adamcfraser: I think a separate function for "actual < expected" is overkill and makes the calling code harder to read.

db/sequence_allocator.go
```diff
@@ -381,3 +415,69 @@ func (s *sequenceAllocator) waitForReleasedSequences(ctx context.Context, startT
...
+	if result < expectedValue {
+		err, currSeq := base.RetryLoop(ctx, "Fix _sync:seq Value", worker, base.CreateDoublingSleeperFunc(1000, 5))
+		if err != nil {
+			return 0, err
+		}
+		allocatedToSeq = currSeq.(uint64)
+	} else {
+		// if _sync:seq has been fixed just increment by batch size to gte new batch for this node
+		base.DebugfCtx(ctx, base.KeyCRUD, "_sync:seq document fixed by another process, continuing with sequence batch allocation")
+		allocatedToSeq, err = s.incrementSequence(s.sequenceBatchSize)
```
adamcfraser: As mentioned above, seems like this could be moved inside the loop to avoid duplication with line 454.

adamcfraser (changes requested):
I think there is some simplification of fixSyncSeqRollback possible.

adamcfraser (approved):

PR #6839, opened 2024-05-22
CBG-3843: include collection names in status of resync 

gregns1:
CBG-3843

- Refactored function that gets collection id's for resync to also return the collection names
- Add the list of collection names to the manager object to then be used in the status function
- Change has been made to both DCP resync and query resync 
- Had to make the collection list on status struct omitempty as when calling endpoint to start resync the status is returned but at this point the Run() function hasn't started to calculate the collections its running on. To avoid the initial POST call returning null for collections I have made it omit empty

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2454/

### DIFF
--- a/rest/utilities_testing.go
+++ b/rest/utilities_testing.go
@@ -2670,3 +2670,18 @@ func RequireGocbDCPResync(t *testing.T) {
 		t.Skip("This test only works against Couchbase Server since rosmar has no support for DCP resync")
 	}
 }
+
+// RequireCollections Iterates through two lists of collections and asserts they both contain the same collections
+func RequireCollections(t testing.TB, collectionNamesFromStatus, collectionSpecified []string) {
+	require.Equal(t, len(collectionSpecified), len(collectionNamesFromStatus))
+	for _, collName := range collectionNamesFromStatus {
+		var found bool
+		for _, name := range collectionSpecified {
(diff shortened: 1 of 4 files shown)


rest/utilities_testing.go
```diff
@@ -2670,3 +2670,18 @@ func RequireGocbDCPResync(t *testing.T) {
 		t.Skip("This test only works against Couchbase Server since rosmar has no support for DCP resync")
 	}
 }
+
+// RequireCollections Iterates through two lists of collections and asserts they both contain the same collections
+func RequireCollections(t testing.TB, collectionNamesFromStatus, collectionSpecified []string) {
```
bbrks: You should be able to use `(assert|require).ElementsMatch` for an equivalent to this

https://pkg.go.dev/github.com/stretchr/testify/assert#ElementsMatch

bbrks (changes requested):
Looks good, just a suggestion on using an alternative assertion facility

rest/utilities_testing.go
gregns1: Had no idea that existed! Updated to use that assertion.

bbrks (approved):

PR #6845, opened 2024-05-29
CBG-3950 filter principal docs for default collection

torcolvin:
When using non default metadata id, the /_user/ and /_role/ endpoints need to filter any docs with _sync:user:db:alice from docs that look like _sync:user:alice

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2462/

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -1046,7 +1046,10 @@ func (db *DatabaseContext) GetUserNames(ctx context.Context) (users []string, er
 	startKey := dbUserPrefix
 	limit := db.Options.QueryPaginationLimit
 
-	userRe := regexp.MustCompile(`^` + dbUserPrefix + `[^:]*$`)
+	userRe, err := regexp.Compile(`^` + dbUserPrefix + `[^:]*$`)
+	if err != nil {
+		return nil, err
+	}
 	users = []string{}
 
 outerLoop:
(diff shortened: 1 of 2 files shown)


rest/user_api_test.go
```diff
@@ -1507,3 +1507,68 @@ func TestUnauthorizedAccessForDB(t *testing.T) {
...
+	dbConfig.Scopes = nil
+	RequireStatus(t, rt.CreateDatabase(defaultCollectionDB, dbConfig), http.StatusCreated)
+
+	// create a database with a named collection
+	dbConfig = rt.NewDbConfig()
+	dbConfig.Scopes = GetCollectionsConfig(t, rt.TestBucket, 1)
+	RequireStatus(t, rt.CreateDatabase(namedCollectionDB, dbConfig), http.StatusCreated)
+
+	// create matching named user and roles in both databases
+	for _, dbName := range []string{defaultCollectionDB, namedCollectionDB} {
```
adamcfraser, line 1536: I feel like it would be useful to have more interleaved user IDs (alphabetically), to make sure there isn't anything unexpected going on related to ordering.  e.g.
_sync:user:alice
_sync:user:namedcollectiondb:alice
_sync:user:nancy
etc.

db/database.go
```diff
@@ -1153,11 +1172,15 @@ outerLoop:
 				startKey = queryRow.Name
 			}
 			if len(rowID) < lenDbUserPrefix && len(rowID) < lenDbRolePrefix {
+				// make sure to increment the resultCount to not skip the next row if the results are
+				// _sync:user:alice and we are looking for _sync:user:verylongdbname:alice
+				if resultCount == 0 && anyPrincipalRe.MatchString(rowID) {
+					resultCount++
```
adamcfraser, line 1178: I'm not clear why we need to increment resultCount here and not on line 1185.

adamcfraser (changes requested):

db/database.go
torcolvin: I kept this code the same, I'm not clear on this either. 

The reason this logic is different is that we only skip the values if it is _not_ a principal document, and I'm preserving that. In the legacy logic:

For a user/rowID like: `_sync:user:alice`:

`len(rowID) < lenDbUserPrefix && len(rowID) < lenDbRolePrefix` -> false
`if !isDbUser && !isDbRole` -> false

For a case where we have:

`_sync:user:alice`
`_sync:user:namedcollectiondb:alice`

`len(rowID) < lenDbUserPrefix && len(rowID) < lenDbRolePrefix` -> true
`if !isDbUser && !isDbRole` -> false

adamcfraser (approved):

PR #6846, opened 2024-05-30
CBG-3640: change serviceconfig.json to use new bootstrap config

gregns1:
CBG-3640

- Change service config file to use up to date bootstrap config

## Pre-review checklist


examples/serviceconfig.json
```diff
@@ -1,15 +1,16 @@
...
-        "GUEST": {"disabled": false, "admin_channels": ["*"] }
-      },
-      "allow_conflicts": false,
-      "revs_limit": 20
+  "bootstrap": {
+    "server": "rosmar:/?mode=memory"
+  },
+  "api": {
+    "public_interface": ":4984",
+    "admin_interface": ":4985"
```
bbrks: We should definitely keep the existing interface setting which locked down the admin API by default.

I'm OK with dropping `0.0.0.0` for the public interface though.

bbrks (changes requested):

examples/serviceconfig.json
gregns1: Will add that now! 

Just thought should I be adding admin_interface_authentication: false as this is rosmar config?

examples/serviceconfig.json
```diff
@@ -1,15 +1,18 @@
...
-        "GUEST": {"disabled": false, "admin_channels": ["*"] }
-      },
-      "allow_conflicts": false,
-      "revs_limit": 20
+  "bootstrap": {
+    "server": "rosmar:/?mode=memory"
+  },
+  "api": {
+    "public_interface": "0.0.0.0:4984",
+    "admin_interface": ":4985",
```
bbrks: restrict admin API to localhost

```suggestion
    "public_interface": ":4984",
    "admin_interface": "127.0.0.1:4985",
```

bbrks (dismissed):

bbrks:
@torcolvin this seem correct to you? The `admin_interface_authentication` options are there to allow in-memory database creation with rosmar.

My expectation is that these would be removed when people switch to Couchbase Server.

torcolvin:
> @torcolvin this seem correct to you? The `admin_interface_authentication` options are there to allow in-memory database creation with rosmar.
> 
> My expectation is that these would be removed when people switch to Couchbase Server.

This seems correct. I wish that json allowed comments to indicate you do Not Want to copy this. I wonder if what we should do instead is force this value to be false the boostrap server is rosmar, and then we don't need to add this to the configuration.

gregns1 (commented):
I had the default admin interface wrong in my locally config meaning that the admin interface and metrics interface auth weren't turned off for rosmar boostrap. I have fixed the config now and works as is.

examples/serviceconfig.json
```diff
@@ -1,15 +1,16 @@
...
+  "api": {
+    "public_interface": ":4984",
+    "admin_interface": "127.0.0.1:4985"
+  },
+  "logging": {
+    "console": {
+      "enabled": true,
+      "log_level": "info",
+      "log_keys": ["*"]
     }
```
torcolvin: I think I'd be inclined to not have console logging at info level, this is just going to fill up systemd logs and the log files are still written to the log file directory.

I use console logging like this when I run SGW myself because it's nice to have it write to stdout but I don't think that's how it is used when it is controlled by systemd

torcolvin (approved):

PR #6848, opened 2024-05-30
CBG-3923: Add audit types and placeholder event

bbrks:
CBG-3923

Merging into audit log feature branch (so it can be movable between 3.2 and 3.3)

- Add types to be able to describe audit events
- Add placeholder audit event
- Add (test) code to generate [kv_engine `auditd`-compatible JSON module descriptor](https://github.com/couchbase/kv_engine/blob/master/auditd/README.md)
	- Output: `{"version":2,"module":"sync_gateway","events":[{"id":54000,"name":"Placeholder audit event","description":"This is a placeholder.","mandatory_fields":{"context":{"provider":"","username":""}},"optional_fields":{"isSomething":true,"operationID":1}}]}`


base/audit/auditd_descriptor.go
```diff
@@ -0,0 +1,34 @@
...
+
+const (
+	// moduleDescriptorName is the name of the module. Must match the name used in the module descriptor file.
+	moduleDescriptorName = "sync_gateway"
+	// auditdFormatVersion is the version of the auditd format to be used. Only version 2 is supported.
+	auditdFormatVersion = 2
+)
+
+// generateAuditdModuleDescriptor returns an auditd-compatible module descriptor for the given events.
+func generateAuditdModuleDescriptor(e events) ([]byte, error) {
```
torcolvin, line 23: I will just say that I was confused a bit about not passing a ptr for events, but it is an array so that is fine.

base/audit/auditd_types.go
```diff
@@ -0,0 +1,82 @@
...
+		MandatoryFields:    toAuditdFieldType(e.mandatoryFields),
+		OptionalFields:     toAuditdFieldType(e.optionalFields),
+	}
+}
+
+// auditd values that represent each type of field
+const (
+	auditFieldTypeNumber = 1
+	auditFieldTypeString = ""
+	auditFieldTypeBool   = true
```
torcolvin: I think I'd rename these types because it was not obvious to me why some were default golang values and some are not.

This matches the spec for couchbase server and I wonder if it's worth linking this spec.

base/audit/events.go
```diff
@@ -0,0 +1,35 @@
...
+// software will be governed by the Apache License, Version 2.0, included in
+// the file licenses/APL2.txt.
+
+package audit
+
+const (
+	// auditdSyncGatewayStartID is the start of an ID range allocated for Sync Gateway by auditd
+	auditdSyncGatewayStartID ID = 53248
+
+	IDPlaceholder ID = 54000
```
torcolvin, line 15: Are we going to define events up from this number? I wonder if it is worth doing for self documenting

```
	// auditdSyncGatewayStartID is the start of an ID range allocated for Sync Gateway by auditd
	auditdSyncGatewayStartID ID = iota + 53248

    // IDPlaceholderID is the last expected event and used for testing
	IDPlaceholder ID = 54000
```

base/audit/types.go
```diff
@@ -0,0 +1,39 @@
...
+	// name is a short textual name of the event
+	name string
+	// description is a longer name / description of the event
+	description string
+	// enabledByDefault indicates whether the event should be enabled by default
+	enabledByDefault bool
+	// filteringPermitted indicates whether the event can be filtered or not
+	filteringPermitted bool
+	// mandatoryFields describe field(s) required for a valid instance of the event
+	mandatoryFields map[string]any
```
torcolvin, line 28: Should we define some mandatory fields like server does? I think at minimum we should do `timestamp` -

Other ideas include `db` or `keyspace` or `username`

torcolvin (commented):
Looks good, just some questions about implementation.

base/audit/auditd_types.go
bbrks: What should I rename them to? I figured auditFieldType was enough?

I'll definitely link the spec because it's weird.

base/audit/events.go
bbrks: Not sure yet. That'll come later once the event table is finalised.

I wanted to avoid iota so we had searchable IDs in the codebase, rather than rely on the auto-incrementing iota value

base/audit/types.go
bbrks: I'll do it when we define the list of events and know what fields we have.

base/audit/auditd_types.go
torcolvin: I guess I was thinking `auditFieldBoolDefault`, etc

base/audit/events.go
torcolvin: Makes sense, and prevents accidental reordering.

torcolvin (approved):

PR #6849, opened 2024-05-30
CBG-3957 add reason in all_dbs for why db is offline

torcolvin:
- fix issue where DatabaseContext.RequiresResync was only updated when
  creating a Database. Update this value after resync completes

I think I'd rather put `requires_resync` field to match `GET /db/` output, but I thought there was a reason that we didn't do that.

Made the fields public only to assert on them in `rest.adminapitest` package.

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2467/

### DIFF
--- a/db/background_mgr_resync_dcp.go
+++ b/db/background_mgr_resync_dcp.go
@@ -242,6 +242,22 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]interface
 		base.InfofCtx(ctx, base.KeyAll, "[%s] resync was terminated. Docs changed: %d Docs Processed: %d", resyncLoggingID, r.DocsChanged.Value(), r.DocsProcessed.Value())
 	}
 
+	collectionsRequiringResync := make([]base.ScopeAndCollectionName, 0)
+	for _, collection := range db.CollectionByID {
+		dataStore := collection.dataStore
+		resyncRequired, err := base.InitSyncInfo(dataStore, db.DatabaseContext.Options.MetadataID)
+		// if we can't write this document but made it here, the resync completed fine, and DatabaseContext.RequireResync will be reset when the database is recreated when it goes online.
+		if err != nil {
+			base.WarnfCtx(ctx, "[%s] error updated sync info for collection %v", resyncLoggingID, err)
+			return nil
+		}
+		if !resyncRequired {
+			continue
+		}
+		dsName := base.ScopeAndCollectionName{Scope: dataStore.ScopeName(), Collection: dataStore.CollectionName()}
+		collectionsRequiringResync = append(collectionsRequiringResync, dsName)
+	}
+	db.RequireResync = collectionsRequiringResync
 	return nil
 }
 
--- a/db/background_mgr_resync_dcp.go
+++ b/db/background_mgr_resync_dcp.go
@@ -215,6 +215,7 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]interface
 
 		// If we regenerated sequences, update syncInfo for all collections affected
 		if regenerateSequences {
+			updatedDsNames := make(map[base.ScopeAndCollectionName]struct{}, len(collectionIDs))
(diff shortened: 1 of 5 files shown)


db/background_mgr_resync_dcp.go
```diff
@@ -242,6 +242,22 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]interface
 		base.InfofCtx(ctx, base.KeyAll, "[%s] resync was terminated. Docs changed: %d Docs Processed: %d", resyncLoggingID, r.DocsChanged.Value(), r.DocsProcessed.Value())
 	}
 
+	collectionsRequiringResync := make([]base.ScopeAndCollectionName, 0)
```
adamcfraser: We should only do this work if we successfully updated SyncInfo for any collections (on line 223) - otherwise can skip this (which would include any resync done with regenerate_sequences=false)

adamcfraser: Actually, can we just remove any collections that we successfully updated on line 223 from db.RequireResync, instead of recomputing the set?

db/background_mgr_resync_dcp.go
```diff
@@ -242,6 +242,22 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]interface
...
+	for _, collection := range db.CollectionByID {
+		dataStore := collection.dataStore
+		resyncRequired, err := base.InitSyncInfo(dataStore, db.DatabaseContext.Options.MetadataID)
```
adamcfraser: I feel like we shouldn't be using InitSyncInfo here - I don't think there's any situation we'd want to be creating SyncInfo if for whatever reason it doesn't exist. But see comment above about maybe just updating db.RequireResync based on line 223.

rest/api.go
```diff
@@ -406,10 +406,11 @@ type DatabaseRoot struct {
...
 
-type dbSummary struct {
-	DBName string `json:"db_name"`
-	Bucket string `json:"bucket"`
-	State  string `json:"state"`
+type DbSummary struct {
+	DBName        string `json:"db_name"`
+	Bucket        string `json:"bucket"`
+	State         string `json:"state"`
+	OfflineReason string `json:"offline_reason,omitempty"`
```
adamcfraser: We're only in one state at a time, so I think a simple 'reason' property is preferable.  (e.g. in future if we want to provide a reason for state=Resyncing, we can reuse).  What do you think?

adamcfraser (changes requested):

rest/api.go
torcolvin: I think I'd prefer to do create a parameter that is `RequireResync` to match the /db/ endpoint, and there isn't ambiguity.

I think the status field offers a lot of flexibility, but it comes at some downsides, which is that it is not always populated. The reason that `reason` it might not be populated is opaque, so if the user calls `/db/_offline` or   sets `start_offline: true` should the reason be offline in that case?

Capella doesn't use the /db/ `require_resync` field, and implementing this take the place of that field.

db/background_mgr_resync_dcp.go
torcolvin: Addressed with a refactor, I didn't like this either.

db/background_mgr_resync_dcp.go
torcolvin: This is much easier to reason through, addressed.

adamcfraser (approved):

rest/api.go
adamcfraser: Since require_resync can potentially be very large, I don't think it's necessarily appropriate for a list REST endpoint.  I think it makes sense to keep the _all_dbs response relatively lightweight (to support frequent polling), and keep `require_resync` available on the /db/ endpoint.  That gives consumers the option to fetch the data based on the reason in the list view, but not be committed to it on every polling interval.

adamcfraser (approved):

PR #6850, opened 2024-05-30
CBG-3849 Remove inline cas handling for xattr config persistence

adamcfraser:
CBG-3849

Remove inline cas usage for xattr config persistence - use standard document CAS semantics.  

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2468/

Ran additional integration test with forced use of XattrBootstrapPersistence for all tests:
https://jenkins.sgwdev.com/job/SyncGateway-Integration/2463/
Triaged the three failures from that run:
- base.TestXattrConfigPersistence
  - Fixed in [5dc8f01](https://github.com/couchbase/sync_gateway/pull/6850/commits/5dc8f01d4650216ff18cc67a4e163d55d55d1125)
- rest.TestPersistentConfigNoBucketField
- rest.TestPersistentDbConfigWithInvalidUpsert 
  - These two tests rely on modifying the config manually in the bucket, and so aren't compatible with XattrBootstrapPersistence.  Non-trivial to modify or skip these tests, so leaving as-is.

### DIFF
--- a/base/config_persistence.go
+++ b/base/config_persistence.go
@@ -190,10 +190,11 @@ func (xbp *XattrBootstrapPersistence) loadConfig(ctx context.Context, c *gocb.Co
 		var body map[string]interface{}
 		bodyErr := res.ContentAt(1, &body)
 		if bodyErr != nil {
-			var restoreErr error
-			casOut, restoreErr = xbp.restoreDocumentBody(c, key, valuePtr, res.Cas())
+			restoreCas, restoreErr := xbp.restoreDocumentBody(c, key, valuePtr)
 			if restoreErr != nil {
 				WarnfCtx(ctx, "Error attempting to restore unexpected deletion of config: %v", restoreErr)
+			} else {
+				casOut = restoreCas
 			}
 		}
 		return uint64(casOut), nil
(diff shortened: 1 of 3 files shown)


adamcfraser:
CI failure on test-race appears to be unrelated issue.

base/config_persistence.go
```diff
@@ -187,26 +184,19 @@ func (xbp *XattrBootstrapPersistence) loadConfig(ctx context.Context, c *gocb.Co
...
+		casOut := res.Cas()
 
 		// deleted document check - if deleted, restore
 		var body map[string]interface{}
-		bodyErr := res.ContentAt(2, &body)
+		bodyErr := res.ContentAt(1, &body)
 		if bodyErr != nil {
-			restoreErr := xbp.restoreDocumentBody(c, key, valuePtr, strCas)
+			var restoreErr error
+			casOut, restoreErr = xbp.restoreDocumentBody(c, key, valuePtr, res.Cas())
```
torcolvin: If there's an error here, I think `casOut` will be updated to 0, which might not be correct? The tombstone document will have a cas.

OTOH, the cas isn't super useful

base/config_persistence.go
```diff
@@ -215,24 +205,23 @@ func (xbp *XattrBootstrapPersistence) loadConfig(ctx context.Context, c *gocb.Co
 	}
 }
 
-// Restore a deleted document's body.  Rewrites metadata, but preserves previous cfgCas
-func (xbp *XattrBootstrapPersistence) restoreDocumentBody(c *gocb.Collection, key string, value interface{}, cfgCas string) error {
+// Restore a deleted document's body.  Rewrites metadata
+func (xbp *XattrBootstrapPersistence) restoreDocumentBody(c *gocb.Collection, key string, value interface{}, cas gocb.Cas) (casOut gocb.Cas, err error) {
```
torcolvin: We aren't using a cas check here because there aren't cas checks on insert. I think you can remove `cas` from an argument here.

 We are sure that there is no document existing because on `gocb.StoreSemanticsInsert`

base/config_persistence.go
torcolvin: I can also see how if there's no document, then it should have a zero cas. I'm not sure what to do here.

adamcfraser: I think you're correct - if the restore fails we should return the cas of the current version of the document (the tombstone), since we're returning the tombstone's xattr.

torcolvin (approved):

PR #6857, opened 2024-05-31
[3.1.7 backport] CBG-3879 Recovery from cas mismatch on metadata documents when using xattrConfig

adamcfraser:
CBG-3879

Preserves write of cas to body for rollback compatibility, but switches to document cas for all optimistic concurrency checks.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2480/
- [x] `sanity test forcing xattrConfig=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2481/

### DIFF
--- a/rest/persistent_config_test.go
+++ b/rest/persistent_config_test.go
@@ -526,75 +526,79 @@ func TestPersistentConfigRegistryRollbackAfterDbConfigRollback(t *testing.T) {
 	base.TestRequiresCollections(t)
 	base.SetUpTestLogging(t, base.LevelDebug, base.KeyHTTP, base.KeyConfig)
 
-	sc, closeFn := startBootstrapServerWithoutConfigPolling(t)
-	defer closeFn()
+	for _, test := range persistentConfigTestCases() {
+		t.Run(test.name, func(t *testing.T) {
+			sc, closeFn := startBootstrapServerWithoutConfigPolling(t, test.xattrConfig)
+			defer closeFn()
 
-	ctx := base.TestCtx(t)
-	tb := base.GetTestBucket(t)
-	defer tb.Close(ctx)
+			ctx := base.TestCtx(t)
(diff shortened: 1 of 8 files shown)


torcolvin (commented):
Can you explain why you don't have to update https://github.com/couchbase/sync_gateway/blob/24844ac239a05def4d7a7bf3f5e239cfce9a7f00/base/config_persistence.go#L270
https://github.com/couchbase/sync_gateway/blob/24844ac239a05def4d7a7bf3f5e239cfce9a7f00/base/config_persistence.go#L313

to do `gocb.UpsertSpec(cfgXattrCasPath, gocb.MutationMacroCAS, UpsertSpecXattr),` ?

Is this becuase those pathways are not written by metadata documents that do a cas check? It feels like failing to update this here means that we would still run into a backward compatibility problem if we downgrade 3.1.7 -> 3.1.6

adamcfraser:
Those are part of the DocumentBootstrapPersistence implementation, which has never used inline CAS.

rest/persistent_config_test.go
```diff
@@ -1670,9 +1635,38 @@ func TestPersistentConfigNoBucketField(t *testing.T) {
...
 }
+
+type persistentConfigTestCase struct {
+	name        string
+	xattrConfig bool
+}
+
+func persistentConfigTestCases() []persistentConfigTestCase {
+
+	if base.UnitTestUrlIsWalrus() {
```
torcolvin, line 1653: rosmar doesn't exist on 3.1 at all and we don't have bootstrap support? I'd probably remove this from here, or indicate it here.

This is fine because all of these tests are already skipped without Server.

torcolvin (dismissed):

torcolvin (approved):

PR #6882, opened 2024-06-08
CBG-3988 Ensure documents are imported when collection added to a db 

adamcfraser:
CBG-3988

Adds an import version to the database config that's incremented when a collection is added to a db. Removing a collection from a db does not increment version. When non-zero, the version is included in the import feed's DCP checkpoint prefix.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2496/

### DIFF
--- a/base/constants_syncdocs.go
+++ b/base/constants_syncdocs.go
@@ -261,6 +261,20 @@ func (m *MetadataKeys) DCPCheckpointPrefix(groupID string) string {
 	return m.dcpCheckpoint
 }
 
+// DCPVersionedCheckpointPrefix returns the prefix used to store versioned DCP checkpoints.
+//
+//	format: _sync:dcp_ck:{m_$}:{groupID:}{version:}
+func (m *MetadataKeys) DCPVersionedCheckpointPrefix(groupID string, version uint64) string {
+	checkpointPrefix := m.dcpCheckpoint
+	if groupID != "" {
+		checkpointPrefix = checkpointPrefix + groupID + ":"
+	}
+	if version != 0 {
+		checkpointPrefix = checkpointPrefix + strconv.FormatUint(version, 10) + ":"
+	}
+	return checkpointPrefix
+}
+
 // DCPBackfillKey returns the key used to store DCP backfill statistics.
 //
 //	format: _sync:{m_$}:dcp_backfill
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -588,6 +588,7 @@ func (h *handler) handlePutDbConfig() (err error) {
 			return nil, base.HTTPErrorf(http.StatusPreconditionFailed, "Provided If-Match header does not match current config version")
 		}
 		oldBucketDbConfig := bucketDbConfig.DbConfig
+		previousCollectionMap := bucketDbConfig.Scopes.CollectionMap()
 
 		if h.rq.Method == http.MethodPost {
 			base.TracefCtx(h.ctx(), base.KeyConfig, "merging upserted config into bucket config")
@@ -597,6 +598,12 @@ func (h *handler) handlePutDbConfig() (err error) {
 		} else {
 			base.TracefCtx(h.ctx(), base.KeyConfig, "using config as-is without merge")
 			bucketDbConfig.DbConfig = *dbConfig
+		}
+
+		// Update ImportVersion if new collections are being added to the database
+		if bucketDbConfig.Scopes.HasNewCollection(previousCollectionMap) {
+			bucketDbConfig.ImportVersion++
+			base.InfofCtx(h.ctx(), base.KeyConfig, "Import version incremented to %d, new collections detected", bucketDbConfig.ImportVersion)
 		}
 
 		if err := dbConfig.validatePersistentDbConfig(); err != nil {
--- a/rest/config_test.go
+++ b/rest/config_test.go
@@ -2768,3 +2768,143 @@ func TestDatabaseConfigDropScopes(t *testing.T) {
 	require.Contains(t, resp.Body.String(), "cannot change scope")
 
 }
+
+func TestScopesConfigCollectionMap(t *testing.T) {
+
+	testCases := []struct {
+		name               string
+		scopeName          string
+		initialCollections []string
+		updatedCollections []string
+		expectedHasNew     bool
+	}{
+		{
+			name:               "add collection",
+			scopeName:          "s1",
+			initialCollections: []string{"sg_test_0"},
+			updatedCollections: []string{"sg_test_0", "sg_test_1"},
+			expectedHasNew:     true,
+		},
+		{
+			name:               "replace collection",
+			scopeName:          "s1",
+			initialCollections: []string{"sg_test_0"},
+			updatedCollections: []string{"sg_test_1"},
+			expectedHasNew:     true,
+		},
+		{
+			name:               "same collection",
+			scopeName:          "s1",
+			initialCollections: []string{"sg_test_0"},
+			updatedCollections: []string{"sg_test_0"},
+			expectedHasNew:     false,
+		},
+		{
+			name:               "remove collection",
+			scopeName:          "s1",
+			initialCollections: []string{"sg_test_0", "sg_test_1"},
+			updatedCollections: []string{"sg_test_0"},
+			expectedHasNew:     false,
+		},
+		{
+			name:               "remove all and add collection",
+			scopeName:          "s1",
+			initialCollections: []string{"sg_test_0", "sg_test_1"},
+			updatedCollections: []string{"sg_test_2"},
+			expectedHasNew:     true,
+		},
+		{
+			name:               "default to named",
+			scopeName:          base.DefaultScope,
+			initialCollections: []string{"_default"},
+			updatedCollections: []string{"sg_test_0"},
+			expectedHasNew:     true,
+		},
+		{
+			name:               "default add named",
+			scopeName:          base.DefaultScope,
+			initialCollections: []string{"_default"},
+			updatedCollections: []string{"_default", "sg_test_0"},
+			expectedHasNew:     true,
+		},
+		{
+			name:               "named to default",
+			scopeName:          base.DefaultScope,
+			initialCollections: []string{"sg_test_0"},
+			updatedCollections: []string{"_default"},
+			expectedHasNew:     true,
+		},
+		{
+			name:               "default remove named",
+			scopeName:          base.DefaultScope,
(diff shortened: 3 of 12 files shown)


rest/config_test.go
```diff
@@ -2768,3 +2768,143 @@ func TestDatabaseConfigDropScopes(t *testing.T) {
...
+// makeScopesConfig creates a scopes config, with additional handling for explicitly defining the default collection in ScopesConfig
+func makeScopesConfigWithDefault(scopeName string, collections []string) *ScopesConfig {
+
+	// handling for _default._default - treat as no scopes defined
+	if len(collections) == 1 && collections[0] == base.DefaultCollection {
+		return nil
+	}
+
+	// handling for explicit default - ScopesConfig with _default._default only defined
+	if len(collections) == 1 && collections[0] == "_scopesdefault" {
```
torcolvin, line 2903: I would create a constant for this as a marker even though it is only for test. I had to look up if you could start a collection with `_`

rest/config_test.go
```diff
@@ -2768,3 +2768,143 @@ func TestDatabaseConfigDropScopes(t *testing.T) {
...
+			require.Equal(t, test.expectedHasNew, updatedScopesConfig.HasNewCollection(initialCollectionsMap))
+		})
+	}
+}
+
+// makeScopesConfig creates a scopes config, with additional handling for explicitly defining the default collection in ScopesConfig
+func makeScopesConfigWithDefault(scopeName string, collections []string) *ScopesConfig {
+
+	// handling for _default._default - treat as no scopes defined
+	if len(collections) == 1 && collections[0] == base.DefaultCollection {
```
torcolvin, line 2898: I have made this mistake when writing tests which is to accidentally use `scope: foo`, `collection: _default`, so it would be good to check whether `scopeName == base.DefaultScope`

rest/admin_api.go
```diff
@@ -599,6 +600,12 @@ func (h *handler) handlePutDbConfig() (err error) {
 			bucketDbConfig.DbConfig = *dbConfig
 		}
 
+		// Update ImportVersion if new collections are being added to the database
+		if bucketDbConfig.Scopes.HasNewCollection(previousCollectionMap) {
+			bucketDbConfig.ImportVersion++
+			base.InfofCtx(h.ctx(), base.KeyConfig, "Import version incremented to %d, new collections detected", bucketDbConfig.ImportVersion)
```
torcolvin, line 606: I don't believe h.ctx() here will include the database name, which I think would be important to understanding this message. You'd be able to determine this information from looking at the Correlation ID and the URL but this seems challenging.

rest/importtest/collections_import_test.go
```diff
@@ -410,6 +429,100 @@ func TestMultiCollectionImportRemoveCollection(t *testing.T) {
...
+	response = rt.SendAdminRequest(http.MethodPost, "/{{.db}}/_config", fmt.Sprintf(collectionsDbConfigUpsertScopes, string(oneCollectionConfigJSON)))
+	rest.RequireStatus(t, response, http.StatusCreated)
+	importVersion = GetImportVersion(t, rt, dbName)
+	require.Equal(t, uint64(2), importVersion)
+
+}
+
+func GetImportVersion(t *testing.T, rt *rest.RestTester, dbName string) uint64 {
+	var databaseConfig rest.DatabaseConfig
+	_, err := rt.ServerContext().BootstrapContext.GetConfig(rt.Context(), rt.CustomTestBucket.GetName(), rt.ServerContext().Config.Bootstrap.ConfigGroupID, dbName, &databaseConfig)
```
torcolvin, line 523: nit:

```suggestion
	_, err := rt.ServerContext().BootstrapContext.GetConfig(rt.Context(), rt.Bucket().GetName(), rt.ServerContext().Config.Bootstrap.ConfigGroupID, dbName, &databaseConfig)
```

In case we are able to not use a CustomTestBucket in the future.

rest/importtest/collections_import_test.go
```diff
@@ -410,6 +429,100 @@ func TestMultiCollectionImportRemoveCollection(t *testing.T) {
...
+
+	// Update with POST, remove collection, version should not change
+	response = rt.SendAdminRequest(http.MethodPost, "/{{.db}}/_config", fmt.Sprintf(collectionsDbConfigUpsertScopes, string(oneCollectionConfigJSON)))
+	rest.RequireStatus(t, response, http.StatusCreated)
+	importVersion = GetImportVersion(t, rt, dbName)
+	require.Equal(t, uint64(2), importVersion)
+
+}
+
+func GetImportVersion(t *testing.T, rt *rest.RestTester, dbName string) uint64 {
```
torcolvin, line 521: nit:

because rest tester already has rt.TB on it, you could drop the testing.T as an argument here to make this simpler

rest/importtest/collections_import_test.go
```diff
@@ -410,6 +429,100 @@ func TestMultiCollectionImportRemoveCollection(t *testing.T) {
...
+		PersistentConfig: true,
+	}
+
+	rt := rest.NewRestTester(t, rtConfig)
+	defer rt.Close()
+
+	// Set up collection payloads for 1, 2, 3 collections
+	oneCollectionConfig := rest.GetCollectionsConfig(t, testBucket, 1)
+	oneCollectionConfigJSON, err := json.Marshal(oneCollectionConfig)
+	require.NoError(t, err)
```
torcolvin, line 461: fyi about testing helper functions for readability

```suggestion
	oneCollectionConfigJSON := base.MustJSONMarshal(oneCollectionConfig)
```

and we can avoid the extra variable, or even convert this just as we are ready to make a rest call.

rest/importtest/collections_import_test.go
```diff
@@ -410,6 +429,100 @@ func TestMultiCollectionImportRemoveCollection(t *testing.T) {
...
+	twoCollectionConfigJSON, err := json.Marshal(twoCollectionConfig)
+	require.NoError(t, err)
+	threeCollectionConfig := rest.GetCollectionsConfig(t, testBucket, 3)
+	threeCollectionConfigJSON, err := json.Marshal(threeCollectionConfig)
+	require.NoError(t, err)
+
+	// Insert with PUT and single collection
+	dbName := "db"
+	response := rt.SendAdminRequest(http.MethodPut, "/"+dbName+"/", fmt.Sprintf(collectionsDbConfig, testBucket.GetName(), string(oneCollectionConfigJSON)))
+	rest.RequireStatus(t, response, http.StatusCreated)
```
torcolvin, line 472: you could construct this differently with helper functions: which might be more readable:

```suggestion
	dbconfig := rt.NewDbConfig()
	dbconfig.Scopes = rest.GetCollectionsConfig(t, testBucket, 1)
	rest.RequireStatus(rt.CreateDatabase("db"), dbconfig), http.StatusCreated)
```

rest/importtest/collections_import_test.go
```diff
@@ -410,6 +429,100 @@ func TestMultiCollectionImportRemoveCollection(t *testing.T) {
...
+
+	importVersion := GetImportVersion(t, rt, dbName)
+	require.Equal(t, uint64(0), importVersion)
+
+	// Update with PUT, change collection set, version should change
+	response = rt.SendAdminRequest(http.MethodPut, "/{{.db}}/_config", fmt.Sprintf(collectionsDbConfig, testBucket.GetName(), string(twoCollectionConfigJSON)))
```
torcolvin, line 478: In line with above:

we do support a matching replace dbconfig:

```
	dbconfig := rt.NewDbConfig()
	dbconfig.Scopes = rest.GetCollectionsConfig(t, testBucket, 2)
	rest.RequireStatus(rt.ReplaceDbConfig("db"), dbconfig), http.StatusCreated)``
```

base/constants_syncdocs.go
```diff
@@ -261,6 +261,20 @@ func (m *MetadataKeys) DCPCheckpointPrefix(groupID string) string {
 	return m.dcpCheckpoint
 }
 
+// DCPVersionedCheckpointPrefix returns the prefix used to store versioned DCP checkpoints.
```
torcolvin, line 264: Is this worth titling DCPImportCheckpointPrefix? I had to look up we kept DCPCheckpointPrefix (resync, attachment compaction). I'm not sure we'd have version numbers for anything else.

rest/importtest/collections_import_test.go
```diff
@@ -410,6 +429,100 @@ func TestMultiCollectionImportRemoveCollection(t *testing.T) {
...
+	require.Equal(t, uint64(1), importVersion)
+
+	// Upsert with POST, do not include or change collection set, version should not change
+	response = rt.SendAdminRequest(http.MethodPost, "/{{.db}}/_config", collectionsDbConfigUpsertRevsLimit)
+	rest.RequireStatus(t, response, http.StatusCreated)
+	importVersion = GetImportVersion(t, rt, dbName)
+	require.Equal(t, uint64(1), importVersion)
+
+	// Upsert with POST, include scopes but do not change collection set, version should not change
+	response = rt.SendAdminRequest(http.MethodPost, "/{{.db}}/_config", fmt.Sprintf(collectionsDbConfigUpsertScopes, string(twoCollectionConfigJSON)))
```
torcolvin, line 496: The equivalent test call is `rt.UpsertDbConfig`

torcolvin (dismissed):
I think a test might be helpful that explicitly uses the only checkpoint names as a str: _sync:dcp_chk:.. from 3.1.0->3.1.6 might be worthwhile to make sure that they match because our code is relying on this matching to not cause any imports to reset.

The rest of the comments are about improving test readability which I'd encourage except we are on a deadline to get this done and so I can approve this code as well and it's just information for the future.

base/constants_syncdocs.go
adamcfraser: I prefer to keep this generic, it doesn't seem unreasonable that we'll have other cases where we persist checkpoints and may want the ability to reset with versions.

rest/admin_api.go
adamcfraser: I went back and confirmed that db is already included here:
`2024-06-08T09:03:42.130-07:00 [INF] Config: c:#003 db:db Import version incremented to 1, new collections detected`

rest/importtest/collections_import_test.go
adamcfraser: For all of these, I feel that using the helper functions makes it hard to verify that we're testing all the supported PUT/POST permutations for db config update, which is what I wanted to ensure coverage of in this test.

torcolvin (approved):

PR #6889, opened 2024-06-11
CBG-3941 use infinite waiting for database init manager

torcolvin:
## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2509/


base/collection_n1ql_common.go
```diff
@@ -262,13 +274,10 @@ func buildIndexes(ctx context.Context, s N1QLStore, indexNames []string) error {
 	buildStatement := fmt.Sprintf("BUILD INDEX ON %s(%s)", s.EscapedKeyspace(), indexNameList)
 	err := s.executeStatement(buildStatement)
 
-	// If indexer reports build will be completed in the background, wait to validate build actually happens.
 	if IsIndexerRetryBuildError(err) {
-		InfofCtx(ctx, KeyQuery, "Indexer error creating index - waiting for background build.  Error:%v", err)
-		// Wait for bucket to be created in background before returning
-		return s.WaitForIndexesOnline(ctx, indexNames, false)
+		InfofCtx(ctx, KeyQuery, "Indexer had a retryable error creating index, assuming it will continue in background. Error:%v", err)
```
adamcfraser:
```suggestion
		InfofCtx(ctx, KeyQuery, "Indexer returned error that will be automatically retried by the index service - waiting for that to complete. Error:%v", err)
```

base/util.go
```diff
@@ -444,6 +444,7 @@ func RetryLoop(ctx context.Context, description string, worker RetryWorker, slee
 
 	numAttempts := 1
 
+	startTime := time.Now()
```
adamcfraser: time.Now() isn't free, and we use RetryLoop in a lot of places, some that may be on performance critical paths.  I'm not sure the additional retry duration information on error is that useful generically.  If it's particularly useful for index initiatlization, maybe move the timer outside of the retry loop call?

adamcfraser (changes requested):
Minor logging change suggested, and a question about the timer addition to RetryLoop.

base/util.go
torcolvin: I think it's somewhat helpful for index initialization but it's trivially calculatable looking at the logs. The trickiest part looking at the logs is making sure that you look at the right collection for index initialization. Without the timeout on capella, I don't think this will be terribly useful, especially in the case cancellation.

adamcfraser (approved):

PR #6890, opened 2024-06-12
CBG-3992 make sure doneChan2 is initialized

torcolvin:
Using print debugging, sometimes `doneChan2` is not quite created yet, so I did the assertion inside `databaseCompleteCallback` to make sure that it is present.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2512/ 40x without failure, it would fail in 2-5 times before

### DIFF
--- a/rest/database_init_manager_test.go
+++ b/rest/database_init_manager_test.go
@@ -11,6 +11,7 @@ package rest
 import (
 	"fmt"
 	"log"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


rest/database_init_manager_test.go
```diff
@@ -425,16 +426,18 @@ func TestDatabaseInitTeardownTiming(t *testing.T) {
...
 	databaseCompleteCount := int64(0)
 	initMgr.databaseCompleteCallback = func(dbName string) {
 		// On first completion, invoke InitializeDatabase with the same collection set post-completion
 		currentCount := atomic.LoadInt64(&databaseCompleteCount)
 		if currentCount == 0 {
+			defer wg.Done()
 			log.Printf("invoking InitializeDatabase again during teardown")
-			var err error
-			doneChan2, err = initMgr.InitializeDatabase(ctx, sc.Config, dbConfig.ToDatabaseConfig())
+			doneChan2, err := initMgr.InitializeDatabase(ctx, sc.Config, dbConfig.ToDatabaseConfig())
```
bbrks, line 438: Does the same race exist in `_getOrAddDatabaseFromConfig`? It seems like it may, although less likely to actually be a problem since there's a lot more going on in that function before the wait.

It feels like the API of `InitializeDatabase` forces an anti-pattern of waiting on an uninitialized or still empty channel

torcolvin: The problem in this test is that `InitializeDatabase` is called in a goroutine here, and then we wait outside the goroutine.

If you are referring to https://github.com/couchbase/sync_gateway/blob/e2c737d8a760562af8d11c258bde68cfc6351ab3/rest/server_context.go#L706, this is synchronous in this function.

adamcfraser (approved):

PR #6891, opened 2024-06-12
CBG-3994: fix for race between config poll and db config update

gregns1:
CBG-3994

- This issue has arisen due to changes made in CBG-3817
- Add back cas line inside UpdateConfig callback in handlePutDbConfig
- This avoid the background task of fetching db configs from the bucket replacing the runtime config in memory during a db config update. This was found by QE where they would end up with mismatch between runtime collection config and persisted config leading to keyspace not found errors.
- Have added come comments to be sure this isn’t removed in future. Added a new function to remove the cfgCas from runtime cfg to force a reload from the bucket when we need to rollback the config
- New test that reproduced the race condition before the fix

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2516/

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -116,7 +116,7 @@ func (h *handler) handleCreateDB() error {
 				"Duplicate database name %q", dbName)
 		}
 
-		_, err = h.server._applyConfig(contextNoCancel, loadedConfig, true, false, false)
+		_, err = h.server._applyConfig(contextNoCancel, loadedConfig, true, false)
 		if err != nil {
 			return databaseLoadErrorAsHTTPError(err)
 		}
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -627,6 +627,9 @@ func (h *handler) handlePutDbConfig() (err error) {
 			return nil, err
 		}
 
+		// we need to set the new dbConfig cfgCas to the old version to avoid the background task polling server for config changes from reloading a
+		// previous persisted config during the update dbConfig process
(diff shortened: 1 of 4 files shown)


rest/admin_api.go
```diff
@@ -642,6 +645,8 @@ func (h *handler) handlePutDbConfig() (err error) {
 	})
 	if err != nil {
 		base.WarnfCtx(h.ctx(), "Couldn't update config for database - rolling back: %v", err)
+		// set runtime cfgCas to 0 to force a reload of db config from the bucket
+		h.server.forceRuntimeConfigReload(contextNoCancel, dbName)
```
torcolvin: I think I'm OK with this fix, but I think there's a race that exists here and I think we want to pass this value to `fetchAndLoadDatabase`.

`forceRuntimeConfigReload` -> acquires `sc.Lock`, changes `sc.dbconfig`, then `fetchAndLoadDatabase` reacquires the lock.

I think it would be _really_ unlikely for someone to come in between these two steps but it could be avoided by adding a parameter to `fetchAndLoadDatabase`. I also think that would be more conceptually clear what is going on, though I do see it would have to be passed down a few levels of functions.

gregns1: I originally wanted to take this approach but didn't like having to pass down a flag a few levels. Thinking it through more I don't think there is too much choice though, will make the change to pass down flag to rest the CAS is wanting to force reload.

bbrks (approved):
I'm OK with this `forceReload` param approach 👍

torcolvin (approved):

PR #6895, opened 2024-06-13
CBG-3867 ban empty replications or replication id on REST calls

torcolvin:
allow them for legacy config reading, but print a warning.

This is a follow on to https://github.com/couchbase/sync_gateway/commit/84997a0d39241fafcc926c86771750f2ad60febd

I could potentially remove the call in `PutReplications` because that will now be hit earlier `dbConfig.validate`

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2523/ (unrelated failure)

### DIFF
--- a/db/sg_replicate_cfg.go
+++ b/db/sg_replicate_cfg.go
@@ -1048,7 +1048,7 @@ func (m *sgReplicateManager) validateReplications(ctx context.Context, replicati
 			base.WarnfCtx(ctx, "An ISGR replication with an empty key exists. To fix this, this replication should be removed from database config and directly from edit the database config and the %q document. %s",
 				m.dbContext.MetadataKeys.SGCfgPrefix(m.dbContext.Options.GroupID), resetCheckpointMsg)
 		} else if replication.ID == "" {
-			base.WarnfCtx(ctx, "An ISGR replication with an empty `id` exists. To fix this, update the replication using PUT /%s/_replication/%s. %s", m.dbContext.Name, replicationID, resetCheckpointMsg)
+			base.WarnfCtx(ctx, "An ISGR replication with an empty 'replication_id' exists. To fix this, update the replication using PUT /%s/_replication/%s. %s", m.dbContext.Name, replicationID, resetCheckpointMsg)
 		}
 	}
 }
--- a/db/sg_replicate_cfg.go
+++ b/db/sg_replicate_cfg.go
@@ -1040,6 +1040,7 @@ func (m *sgReplicateManager) AddReplication(replication *ReplicationCfg) error {
 
 // validateReplications checks the replication configs and provides warning messages.
 func (m *sgReplicateManager) validateReplications(ctx context.Context, replications map[string]*ReplicationCfg) {
+	// the replications may exist in a way that is missing ID or name, validate and warn the user.
 	resetCheckpointMsg := "This will reset the ISGR checkpoints but prevent multiple simultaneous replications from overwriting status and checkpoint documents"
 	for replicationID, replication := range replications {
 		if replicationID == "" {
--- a/db/sg_replicate_cfg.go
+++ b/db/sg_replicate_cfg.go
@@ -527,6 +527,7 @@ func (m *sgReplicateManager) StartReplications(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	m.validateReplications(ctx, replications)
 	for replicationID, replicationCfg := range replications {
 		base.DebugfCtx(m.loggingCtx, base.KeyCluster, "Replication %s is assigned to node %s (local node is %s) on start up", replicationID, replicationCfg.AssignedNode, m.localNodeUUID)
 		if replicationCfg.AssignedNode == m.localNodeUUID {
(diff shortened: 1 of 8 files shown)


rest/admin_api.go
```diff
@@ -60,7 +60,8 @@ func (h *handler) handleCreateDB() error {
 			return base.HTTPErrorf(http.StatusBadRequest, err.Error())
 		}
 
-		if err := config.validate(h.ctx(), validateOIDC); err != nil {
+		isUpsert := true
```
adamcfraser: Create db shouldn't be considered an upsert - are we using the wrong terminology for the legacy case?

rest/config.go
```diff
@@ -676,7 +676,8 @@ func (dbConfig *DbConfig) validatePersistentDbConfig() (errorMessages error) {
 
 // validateConfigUpdate combines the results of validate and validateChanges.
 func (dbConfig *DbConfig) validateConfigUpdate(ctx context.Context, old DbConfig, validateOIDCConfig bool) error {
-	err := dbConfig.validate(ctx, validateOIDCConfig)
+	isUpsert := true
+	err := dbConfig.validate(ctx, validateOIDCConfig, isUpsert)
```
adamcfraser: At this point for a POST _config we've already performed a merge between the incoming request and the existing db config.  It seems like if there's an existing replication in the config with id="", and we're making an upsert unrelated to replications, we're going to reject that update here.  Given that, do we think this is the right place to be performing the empty replication ID check?  Or is there something in the handling for existing replications I'm overlooking.

rest/config.go
```diff
@@ -711,11 +712,12 @@ func (dbConfig *DbConfig) validateChanges(ctx context.Context, old DbConfig) err
 	return nil
 }
 
-func (dbConfig *DbConfig) validate(ctx context.Context, validateOIDCConfig bool) error {
-	return dbConfig.validateVersion(ctx, base.IsEnterpriseEdition(), validateOIDCConfig)
+// validate checks the DbConfig for any invalid or unsupported values and return a http error. If isUpsert is true, add any checks that might be upposrte
+func (dbConfig *DbConfig) validate(ctx context.Context, validateOIDCConfig bool, isUpsert bool) error {
```
adamcfraser: I don't like adding the conditional 'isUpsert' parameter here - previously this function's usage was clear - pass a fully populated dbconfig, and find out if it's valid.  Conditional handling based on the type of update makes that quite a bit harder to reason about for me.

If the more correct name for the parameter is 'isLegacyConfig' that might be an improvement?

adamcfraser (changes requested):

rest/config.go
torcolvin: I changed the name of this to be `validateReplications` to indicate whether we want to validateReplications. It does make it narrower but hopefully easier to understand when we want to do it.

torcolvin:
Added a feature to not overwrite the replications from a dbConfig, since overwriting them would mean that the replications in CfgSG get updated every time the database gets reloaded, despite updates to /db/_replication/.

I added warning code to `sgReplicateManager` to find existing replications that are affected and provided instructions for fixing. The "easy" case of just missing ID will be correctable using REST api, but the case of missing keys has to be updated by editing the cluster doc directly. That's why I moved the validation to `sgReplicateManager` since this where we know the name of the document.

db/sg_replicate_cfg.go
```diff
@@ -1037,8 +1038,23 @@ func (m *sgReplicateManager) AddReplication(replication *ReplicationCfg) error {
...
+func (m *sgReplicateManager) validateReplications(ctx context.Context, replications map[string]*ReplicationCfg) {
+	// the replications may exist in a way that is missing ID or name, validate and warn the user.
+	resetCheckpointMsg := "This will reset the ISGR checkpoints but prevent multiple simultaneous replications from overwriting status and checkpoint documents"
+	for replicationID, replication := range replications {
+		if replicationID == "" {
+			// we don't know the name
+			base.WarnfCtx(ctx, "An ISGR replication with an empty key exists. To fix this, this replication should be removed from database config and directly from edit the database config and the %q document. %s",
+				m.dbContext.MetadataKeys.SGCfgPrefix(m.dbContext.Options.GroupID), resetCheckpointMsg)
+		} else if replication.ID == "" {
+			base.WarnfCtx(ctx, "An ISGR replication with an empty `id` exists. To fix this, update the replication using PUT /%s/_replication/%s. %s", m.dbContext.Name, replicationID, resetCheckpointMsg)
```
adamcfraser: I think avoiding back-ticks in log lines would be preferred, and we decided at one point to try to standardize on ID.
```suggestion
			base.WarnfCtx(ctx, "An ISGR replication with an empty ID exists. To fix this, update the replication using PUT /%s/_replication/%s. %s", m.dbContext.Name, replicationID, resetCheckpointMsg)
```

adamcfraser (commented):
Looks good - one nit on the logging.

adamcfraser (approved):

PR #6900, opened 2024-06-17
CBG-3783 do not print output of lsb_release

torcolvin:
When running service/sync_gateway_service_install.sh script with --service-cmd, the output of the command is used. This is used to print value here:

https://github.com/couchbase/sync_gateway/blob/905c896df90dcba8bf43e95abe6c03d4c092b67d/build/postinst.tmpl#L69

### DIFF
--- a/service/sync_gateway_service_install.sh
+++ b/service/sync_gateway_service_install.sh
@@ -47,7 +47,8 @@ usage() {
 }
 
 ostype() {
-  if command -v lsb_release; then
+  lsb_release 2>&1 > /dev/null
+  if [ $? -eq 0 ]; then
     OS=$(lsb_release -si)
     VER=$(lsb_release -sr)
   elif [ -f /etc/os-release ]; then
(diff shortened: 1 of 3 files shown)


bbrks:
I'm not sure how the PR description relates to the changed lines?

Is this change to support shells that don't have a `command` builtin? Which shells?

torcolvin:
> I'm not sure how the PR description relates to the changed lines?
> 
> Is this change to support shells that don't have a `command` builtin? Which shells?

Running `command -v lsb_release` will output `/usr/bin/lsb_release` if the command is found. Since the output of `service/sync_gateway_service_install.sh --service-cmd` is supposed to just be `systemctl start sync_gateway` but if `lsb_release` exists on the system it is:

```
/usr/bin/lsb_release
systemctl start sync_gateway
``` 

The output of this command is used in the text that is printed when the package is installed.

bbrks (approved):

PR #6901, opened 2024-06-18
CBG-4011 don't add XattrsToDelete on resurrection

torcolvin:
resurrections in datastore API check that we are not trying to delete xattrs

Technically the correct if statement is "if resurrection" where docExists and the new doc will not be a tombstone. If we had a case where it goes from tombstone -> tombstone and delete _mou, that would not be supported by `updateAndReturnDoc` but that doesn't happen.

This case is different from others in that _mou is written. The error it hit is https://github.com/couchbase/sync_gateway/blob/630014e5ec77b7c6e66f177841aff5263991f73e/base/collection_xattr_common.go#L31

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2521/

### DIFF
--- a/db/import_test.go
+++ b/db/import_test.go
@@ -924,7 +924,7 @@ func TestImportResurrectionMou(t *testing.T) {
 	}
 	require.NotNil(t, syncData)
 
-	// Delete via SDK, expect mou to be removed
+	// Delete via SDK, the mou will be updated by the import process
 	require.NoError(t, collection.dataStore.Delete(docID))
 	base.RequireWaitForStat(t, func() int64 {
 		return db.DbStats.SharedBucketImport().ImportCount.Value()
--- a/db/import_test.go
+++ b/db/import_test.go
@@ -893,6 +893,59 @@ func TestMetadataOnlyUpdate(t *testing.T) {
 
 }
 
+func TestImportResurrectionMou(t *testing.T) {
+	base.SetUpTestLogging(t, base.LevelInfo, base.KeyMigrate, base.KeyImport, base.KeyCRUD)
+	db, ctx := setupTestDBWithOptionsAndImport(t, nil, DatabaseContextOptions{})
+	defer db.Close(ctx)
+
+	collection := GetSingleDatabaseCollectionWithUser(t, db)
+
+	docID := "mouResurrection"
+
+	firstBody := Body{"foo": "bar"}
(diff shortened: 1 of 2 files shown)


db/import_test.go
```diff
@@ -893,6 +893,59 @@ func TestMetadataOnlyUpdate(t *testing.T) {
...
+	}, 1)
+	syncData, mou, _ = getSyncAndMou(t, collection, docID)
+	if db.Bucket.IsSupported(sgbucket.BucketStoreFeatureMultiXattrSubdocOperations) {
+		require.NotNil(t, mou)
+	} else {
+		require.Nil(t, mou)
+	}
+	require.NotNil(t, syncData)
+
+	// Delete via SDK, expect mou to be removed
```
adamcfraser: This comment doesn't look right - we don't expect an SDK delete to remove the mou.

We also expect the import of the delete to be a metadata-only update, so _mou should be present post-import.

db/import_test.go
```diff
@@ -893,6 +893,59 @@ func TestMetadataOnlyUpdate(t *testing.T) {
...
+	require.Nil(t, mou)
+
+	// Update via SDK, expect mou to be created
+	err = collection.dataStore.Set(docID, 0, nil, []byte(`{"foo": "baz"}`))
+	require.NoError(t, err)
+	base.RequireWaitForStat(t, func() int64 {
+		return db.DbStats.SharedBucketImport().ImportCount.Value()
+	}, 1)
+	syncData, mou, _ = getSyncAndMou(t, collection, docID)
+	if db.Bucket.IsSupported(sgbucket.BucketStoreFeatureMultiXattrSubdocOperations) {
```
adamcfraser, line 920: Should we just put a useMou() skip check at the start of this test?  It looks like the test is specifically checking _mou functionality.

adamcfraser (changes requested):

db/import_test.go
torcolvin: There is no code that tests this code pathway as is where we do an import -> ressurrection. It was not just that the mou wasn't deleted, it actually returned an `sgbucket.ErrDeleteXattrOnTombstone`. I'm inclined to leave it to ensure this pathway is successful in that case was well since it mirrors a QE test.

adamcfraser (approved):

PR #6902, opened 2024-06-20
CBG-3783 don't output error if command doesn't exist by using variable

torcolvin:
The previous command wasn't right because it would print out an error. I think I made a last minute change without testing to https://github.com/couchbase/sync_gateway/commit/80d09c9fa758357afc2ae6b5569df2ccbdd3fbf3

### DIFF
--- a/service/sync_gateway_service_install.sh
+++ b/service/sync_gateway_service_install.sh
@@ -47,7 +47,7 @@ usage() {
 }
 
 ostype() {
-  lsb_release 2>&1 > /dev/null
+  _lsb_release_exists=$(command -v lsb_release)
   if [ $? -eq 0 ]; then
     OS=$(lsb_release -si)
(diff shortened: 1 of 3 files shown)


service/sync_gateway_service_install.sh
```diff
@@ -47,7 +47,7 @@ usage() {
 }
 
 ostype() {
-  lsb_release 2>&1 > /dev/null
+  _lsb_release_exists=$(command -v lsb_release)
   if [ $? -eq 0 ]; then
```
bbrks: I think this is as portable and a more common way of writing this, with the benefit of also catching the odd case of `lsb_release` not being executable

```suggestion
  if [ -x "$(command -v lsb_release)" ]; then
```

bbrks (approved):

PR #6904, opened 2024-06-20
CBG-3563 Implement automatic memory profiling

torcolvin:
This should write memory profiles like `pprof_heap_high_{timestamp}.pb.gz` to the log file directory, so sgcollect can pick them up.

- Collect the memory profiles a maximum of every 5 minutes.
- Keep only 10 memory profiles.
- `api.heap_profile_collection_threshold` is configurable but allow disabling with `api.disable_heap_profile_collection`

https://docs.google.com/document/d/1Z8pqW-CEGpAdxSSQCakBf4tSKbmzZJGz2Ojlm2JmQ8o

Before merging this code, I want to test it in capella using their cgroup setup. After implementing this, this configuration should be documented (via DOC ticket).

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2524/

### DIFF
--- a/rest/config_flags.go
+++ b/rest/config_flags.go
@@ -64,9 +64,6 @@ func registerConfigFlags(config *StartupConfig, fs *flag.FlagSet) map[string]con
 		"api.cors.headers":      {&config.API.CORS.Headers, fs.String("api.cors.headers", "", "List of comma separated allowed headers")},
 		"api.cors.max_age":      {&config.API.CORS.MaxAge, fs.Int("api.cors.max_age", 0, "Maximum age of the CORS Options request")},
 
-		"api.heap_profile_collection_threshold": {&config.API.HeapProfileCollectionThreshold, fs.Uint64("api.heap_profile_collection_threshold", 0, "Threshold in bytes for collecting heap profiles automatically. If set, Sync Gateway will collect a memory profile when it exceeds this value. The default value will be set to 85% of the lesser of cgroup or system memory.")},
-		"api.heap_profile_disable_collection":   {&config.API.HeapProfileDisableCollection, fs.Bool("api.heap_profile_disable_collection", false, "Disables automatic heap profile collection.")},
-
 		"logging.log_file_path":   {&config.Logging.LogFilePath, fs.String("logging.log_file_path", "", "Absolute or relative path on the filesystem to the log file directory. A relative path is from the directory that contains the Sync Gateway executable file")},
 		"logging.redaction_level": {&config.Logging.RedactionLevel, fs.String("logging.redaction_level", "", "Redaction level to apply to log output. Options: none, partial, full, unset")},
 
@@ -137,6 +134,9 @@ func registerConfigFlags(config *StartupConfig, fs *flag.FlagSet) map[string]con
 		"replicator.max_concurrent_replications":    {&config.Replicator.MaxConcurrentReplications, fs.Int("replicator.max_concurrent_replications", 0, "Maximum number of replication connections to the node")},
 		"replicator.max_concurrent_changes_batches": {&config.Replicator.MaxConcurrentChangesBatches, fs.Int("replicator.max_concurrent_changes_batches", 0, "Maximum number of changes batches to process concurrently per replication")},
 		"replicator.max_concurrent_revs":            {&config.Replicator.MaxConcurrentRevs, fs.Int("replicator.max_concurrent_revs", 0, "Maximum number of revs to process concurrently per replication")},
+
+		"heap_profile_collection_threshold": {&config.HeapProfileCollectionThreshold, fs.Uint64("heap_profile_collection_threshold", 0, "Threshold in bytes for collecting heap profiles automatically. If set, Sync Gateway will collect a memory profile when it exceeds this value. The default value will be set to 85% of the lesser of cgroup or system memory.")},
+		"heap_profile_disable_collection":   {&config.HeapProfileDisableCollection, fs.Bool("heap_profile_disable_collection", false, "Disables automatic heap profile collection.")},
 
 		"unsupported.diagnostic_interface":                 {&config.Unsupported.DiagnosticInterface, fs.String("unsupported.diagnostic_interface", "", "Network interface to bind diagnostic API to")},
 		"unsupported.stats_log_frequency":                  {&config.Unsupported.StatsLogFrequency, fs.String("unsupported.stats_log_frequency", "", "How often should stats be written to stats logs")},
--- a/rest/config_flags.go
+++ b/rest/config_flags.go
@@ -64,7 +64,7 @@ func registerConfigFlags(config *StartupConfig, fs *flag.FlagSet) map[string]con
 		"api.cors.headers":      {&config.API.CORS.Headers, fs.String("api.cors.headers", "", "List of comma separated allowed headers")},
 		"api.cors.max_age":      {&config.API.CORS.MaxAge, fs.Int("api.cors.max_age", 0, "Maximum age of the CORS Options request")},
 
-		"api.heap_profile_collection_threshold": {&config.API.HeapProfileCollectionThreshold, fs.Uint64("api.heap_profile_collection_threshold", 0, "Threshold in bytes for collecting heap profiles automatically. If set, Sync Gateway will collect a memory profile when it exceeds this value. The default value will be set to 85% to the lesser of cgroup or system memory.")},
+		"api.heap_profile_collection_threshold": {&config.API.HeapProfileCollectionThreshold, fs.Uint64("api.heap_profile_collection_threshold", 0, "Threshold in bytes for collecting heap profiles automatically. If set, Sync Gateway will collect a memory profile when it exceeds this value. The default value will be set to 85% of the lesser of cgroup or system memory.")},
 		"api.heap_profile_disable_collection":   {&config.API.HeapProfileDisableCollection, fs.Bool("api.heap_profile_disable_collection", false, "Disables automatic heap profile collection.")},
 
 		"logging.log_file_path":   {&config.Logging.LogFilePath, fs.String("logging.log_file_path", "", "Absolute or relative path on the filesystem to the log file directory. A relative path is from the directory that contains the Sync Gateway executable file")},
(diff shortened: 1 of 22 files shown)


rest/config_startup.go
```diff
@@ -73,6 +76,20 @@ func DefaultStartupConfig(defaultLogFilePath string) StartupConfig {
 		},
 		MaxFileDescriptors: DefaultMaxFileDescriptors,
 	}
+
+	memoryTotal, err := memlimit.FromCgroup()
+	if err != nil {
+		base.TracefCtx(ctx, base.KeyAll, "Did not detec a cgroup for a memory limit")
```
gregns1:
```suggestion
		base.TracefCtx(ctx, base.KeyAll, "Did not detect a cgroup for a memory limit")
```

rest/stats_context_test.go
```diff
@@ -62,3 +66,41 @@ func TestDescriptionPopulation(t *testing.T) {
...
+	for i := 2; i < 12; i++ {
+		// reset heap profile time time to ensure we create new heap profiles
+		stats.lastHeapProfile = time.Time{}
+		expectedFilenames = append(expectedFilenames, fmt.Sprintf("pprof_heap_high_%02d.pb.gz", i))
+		require.NoError(t, stats.collectMemoryProfile(ctx, outputDir, fmt.Sprintf("%02d", i)))
+	}
+	slices.Sort(expectedFilenames)
+	require.Equal(t, expectedFilenames, getFilenames(t, outputDir))
+
+	// ask for another profile, this should not be collected because lastHeapProfile was not set.
```
gregns1: Nitpick: I think more context could be added here. First time I read this test I had to go back to `collectMemoryProfile` to check the logic to see why we expect this profile to not be collected.

rest/stats_context_test.go
```diff
@@ -62,3 +66,41 @@ func TestDescriptionPopulation(t *testing.T) {
...
+	// collect enough profiles to trigger rotation
+	expectedFilenames := make([]string, 0, 10)
+	for i := 2; i < 12; i++ {
+		// reset heap profile time time to ensure we create new heap profiles
+		stats.lastHeapProfile = time.Time{}
+		expectedFilenames = append(expectedFilenames, fmt.Sprintf("pprof_heap_high_%02d.pb.gz", i))
+		require.NoError(t, stats.collectMemoryProfile(ctx, outputDir, fmt.Sprintf("%02d", i)))
+	}
+	slices.Sort(expectedFilenames)
+	require.Equal(t, expectedFilenames, getFilenames(t, outputDir))
```
gregns1: For future you could use require.ElementsMatch() instead of having to sort the slice.

rest/config_flags.go
```diff
@@ -64,6 +64,9 @@ func registerConfigFlags(config *StartupConfig, fs *flag.FlagSet) map[string]con
 		"api.cors.headers":      {&config.API.CORS.Headers, fs.String("api.cors.headers", "", "List of comma separated allowed headers")},
 		"api.cors.max_age":      {&config.API.CORS.MaxAge, fs.Int("api.cors.max_age", 0, "Maximum age of the CORS Options request")},
 
+		"api.heap_profile_collection_threshold": {&config.API.HeapProfileCollectionThreshold, fs.Uint64("api.heap_profile_collection_threshold", 0, "Threshold in bytes for collecting heap profiles automatically. If set, Sync Gateway will collect a memory profile when it exceeds this value. The default value will be set to 85% to the lesser of cgroup or system memory.")},
```
gregns1:
```suggestion
		"api.heap_profile_collection_threshold": {&config.API.HeapProfileCollectionThreshold, fs.Uint64("api.heap_profile_collection_threshold", 0, "Threshold in bytes for collecting heap profiles automatically. If set, Sync Gateway will collect a memory profile when it exceeds this value. The default value will be set to 85% of the lesser of cgroup or system memory.")},
```

rest/config_startup.go
```diff
@@ -128,8 +145,10 @@ type APIConfig struct {
 	CompressResponses  *bool `json:"compress_responses,omitempty"   help:"If false, disables compression of HTTP responses"`
 	HideProductVersion *bool `json:"hide_product_version,omitempty" help:"Whether product versions removed from Server headers and REST API responses"`
 
-	HTTPS HTTPSConfig      `json:"https,omitempty"`
-	CORS  *auth.CORSConfig `json:"cors,omitempty"`
+	HTTPS                          HTTPSConfig      `json:"https,omitempty"`
+	CORS                           *auth.CORSConfig `json:"cors,omitempty"`
+	HeapProfileCollectionThreshold *uint64          `json:"heap_profile_collection_threshold,omitempty" help:"Threshold in bytes for collecting heap profiles automatically. If set, Sync Gateway will collect a memory profile when it exceeds this value. The default value will be set to 85% to the lesser of cgroup or system memory."`
```
gregns1:
```suggestion
	HeapProfileCollectionThreshold *uint64          `json:"heap_profile_collection_threshold,omitempty" help:"Threshold in bytes for collecting heap profiles automatically. If set, Sync Gateway will collect a memory profile when it exceeds this value. The default value will be set to 85% of the lesser of cgroup or system memory."`
```

gregns1 (changes requested):
Generally looks good, just few small things I noticed. I tested locally by running the branch and seems to work well!

rest/config_startup.go
```diff
@@ -73,6 +76,20 @@ func DefaultStartupConfig(defaultLogFilePath string) StartupConfig {
 		},
 		MaxFileDescriptors: DefaultMaxFileDescriptors,
 	}
+
```
adamcfraser, line 76: I feel like we shouldn't be doing the memlimit/mem calls when populating the config, and insterad this should instead be done in NewServerContext when the server context is created (in the scenario where config.API.HeapProfileCollectionThreshold is not specified).  That also avoids logging here when logging isn't initialized, and also avoids this work altogether if the user's bootstrap config has explicitly set HeapProfileCollectionThreshold.

rest/config_startup.go
```diff
@@ -128,8 +145,10 @@ type APIConfig struct {
 	CompressResponses  *bool `json:"compress_responses,omitempty"   help:"If false, disables compression of HTTP responses"`
 	HideProductVersion *bool `json:"hide_product_version,omitempty" help:"Whether product versions removed from Server headers and REST API responses"`
 
-	HTTPS HTTPSConfig      `json:"https,omitempty"`
-	CORS  *auth.CORSConfig `json:"cors,omitempty"`
+	HTTPS                          HTTPSConfig      `json:"https,omitempty"`
+	CORS                           *auth.CORSConfig `json:"cors,omitempty"`
+	HeapProfileCollectionThreshold *uint64          `json:"heap_profile_collection_threshold,omitempty" help:"Threshold in bytes for collecting heap profiles automatically. If set, Sync Gateway will collect a memory profile when it exceeds this value. The default value will be set to 85% of the lesser of cgroup or system memory."`
```
adamcfraser: I'm not sure these belong in APIConfig (which has to do with the REST APIs).  Was there a reason to put it here, as opposed to (say) BootstrapConfig?

rest/server_context.go
```diff
@@ -1666,7 +1667,18 @@ func (sc *ServerContext) logStats(ctx context.Context) error {
 	// Marshal expvar map w/ timestamp to string and write to logs
 	base.RecordStats(string(marshalled))
 
-	return nil
+	if sc.Config.API.HeapProfileDisableCollection {
```
adamcfraser: The early returns here (1671 and 1677) feel out of place - it would be easy for someone to need to add something to logStats in future and just stick it at the end of the function.  Maybe move all of this into a sc.collectMemoryProfile() function to encapsulate this?

adamcfraser (changes requested):
A few minor suggestions.

rest/config_startup.go
torcolvin: I think it doesn't belong in the bootstrap config because it doesn't have to do with couchbase server. I moved it to the toplevel.

I don't love that either, but I agree it doesn't below in API.

rest/server_context.go
```diff
@@ -178,6 +181,18 @@ func NewServerContext(ctx context.Context, config *StartupConfig, persistentConf
 		sc.DatabaseInitManager = &DatabaseInitManager{}
 	}
 
+	if config.HeapProfileCollectionThreshold != nil {
+		sc.statsContext.heapProfileCollectionThreshold = int64(*config.HeapProfileCollectionThreshold)
```
adamcfraser: It feels like we should either define this as int64 or uint64 in both places, to avoid the risk of an unsafe cast when a user specifies a very large value in the config.

adamcfraser (changes requested):

rest/server_context.go
torcolvin: I like using uint64 because that matches the values returned from gosputil or memlimit. The reason for not using it everywhere is that our stats are int64, so `GoMemStatsHeapInUse` is int64.

rest/server_context.go
```diff
@@ -1666,7 +1682,18 @@ func (sc *ServerContext) logStats(ctx context.Context) error {
...
+	}
+
+	currentMemory := base.SyncGatewayStats.GlobalStats.ResourceUtilizationStats().GoMemstatsHeapInUse.Value()
+	profileCollectionThreshold := int64(*sc.Config.HeapProfileCollectionThreshold)
+	if currentMemory <= profileCollectionThreshold {
+		return nil
+	}
+	base.InfofCtx(ctx, base.KeyAll, "Memory usage %d exceeds threshold %d, collecting memory profile", currentMemory, profileCollectionThreshold)
+
+	return sc.statsContext.collectMemoryProfile(ctx, sc.Config.Logging.LogFilePath, timestamp)
```
gregns1, line 1685: Looks like these lines (lines 1685 to 1695) can be removed? As these checks are done in `collectMemoryProfile`

gregns1 (changes requested):
Mostly looks good just one thing I noticed then should be ready to go

gregns1 (approved):

adamcfraser (approved):

PR #6907, opened 2024-06-21
Add automatic formatting to sgcollect_info, like other python files

torcolvin:
This was missed by https://github.com/couchbase/sync_gateway/pull/6772 because sgcollect_info doesn't end in py.


gregns1 (approved):
LGTM

PR #4511, opened 2020-02-26
CBG-711 Avoid long sleep in FlushLogBuffers

JRascagneres:

PR #6908, opened 2024-06-21
CBG-3926: Initial implementation of Audit()

bbrks:
CBG-3926

- Enable devmode flag in CI/Integration tests
- Changes required to produce 3 audit events (ReadDatabase, UserAuthenticated, UserAuthenticationFailed)
- `Audit()` function to create/log audit event
- Currently uses `slog` but plan to replace with more basic JSON->FileLogger

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2526/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,7 +86,7 @@ jobs:
       - name: Build
         run: go build -v "./..."
       - name: Run Tests
-        run: go test -shuffle=on -timeout=30m -count=1 -json -v "./..." | tee test.json | jq -s -jr 'sort_by(.Package,.Time) | .[].Output | select (. != null )'
+        run: go test -tags cb_sg_devmode -shuffle=on -timeout=30m -count=1 -json -v "./..." | tee test.json | jq -s -jr 'sort_by(.Package,.Time) | .[].Output | select (. != null )'
         shell: bash
       - name: Annotate Failures
         if: always()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,7 @@ jobs:
         with:
           go-version: 1.22.2
       - name: Run Tests
-        run: go test -race -shuffle=on -timeout=30m -count=1 -json -v "./..." | tee test.json | jq -s -jr 'sort_by(.Package,.Time) | .[].Output | select (. != null )'
+        run: go test -tags cb_sg_devmode -race -shuffle=on -timeout=30m -count=1 -json -v "./..." | tee test.json | jq -s -jr 'sort_by(.Package,.Time) | .[].Output | select (. != null )'
         shell: bash
       - name: Annotate Failures
         if: always()
--- a/base/audit_events.go
+++ b/base/audit_events.go
@@ -19,27 +19,12 @@ const (
 )
 
 const (
-	AuditIDAuditEnabled  AuditID = 53248 // audit enabled
-	AuditIDAuditDisabled AuditID = 53249 // audit disabled
-	AuditIDAuditChanged  AuditID = 53250 // audit configuration changed
+	AuditIDAuditEnabled AuditID = 53248
 
-	AuditIDUserAuthenticated        AuditID = 53260 // successful login
-	AuditIDUserAuthenticationFailed AuditID = 53261 // incorrect login
-	AuditIDUserAuthorizationFailed  AuditID = 53262 // correct login but incorrect permissions
+	AuditIDPublicUserAuthenticated        AuditID = 53260
+	AuditIDPublicUserAuthenticationFailed AuditID = 53261
 
-	AuditIDCreateDatabase       AuditID = 53300 // create database
-	AuditIDReadDatabase         AuditID = 53301 // view database information
-	AuditIDReadDatabaseConfig   AuditID = 53302 // view database config
-	AuditIDUpdateDatabaseConfig AuditID = 53303 // update database config
-	AuditIDDeleteDatabase       AuditID = 53304 // delete database
-
-	AuditIDDocumentCreate AuditID = 54300 // create document
-	AuditIDDocumentRead   AuditID = 54301 // read document
-	AuditIDDocumentUpdate AuditID = 54302 // update document
-	AuditIDDocumentDelete AuditID = 54303 // delete document
-
-	AuditIDReplicationConnect    AuditID = 55300 // new replication connection
-	AuditIDReplicationDisconnect AuditID = 55301 // replication connection closed
+	AuditIDReadDatabase AuditID = 53301
 )
 
 // AuditEvents is a table of audit events created by Sync Gateway.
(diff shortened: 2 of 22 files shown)


.github/workflows/ci.yml
```diff
@@ -86,7 +86,7 @@ jobs:
       - name: Build
         run: go build -v "./..."
       - name: Run Tests
-        run: go test -shuffle=on -timeout=30m -count=1 -json -v "./..." | tee test.json | jq -s -jr 'sort_by(.Package,.Time) | .[].Output | select (. != null )'
+        run: go test -tags cb_sg_devmode -shuffle=on -timeout=30m -count=1 -json -v "./..." | tee test.json | jq -s -jr 'sort_by(.Package,.Time) | .[].Output | select (. != null )'
```
torcolvin, line 89: do you want to add this to -race?

base/audit_types.go
```diff
@@ -32,15 +36,78 @@ type EventDescriptor struct {
...
+	if !ok {
+		return fmt.Errorf("unknown audit event ID %d", i)
+	}
+	return mandatoryFieldsPresent(f, event.MandatoryFields)
+}
+
+func mandatoryFieldsPresent(fields, mandatoryFields AuditFields) error {
+	me := &MultiError{}
+	for k, v := range mandatoryFields {
+		// recuse if map
```
torcolvin:
```suggestion
		// recurse if map
```

base/logger_audit.go
```diff
@@ -0,0 +1,195 @@
...
+	logger.Enabled.Set(*config.Enabled)
+
+	if buffer != nil {
+		logger.buffer = *buffer
+	}
+
+	// Only create the collateBuffer channel and worker if required.
+	if *config.CollationBufferSize > 1 {
+		logger.collateBuffer = make(chan string, *config.CollationBufferSize)
+		logger.flushChan = make(chan struct{}, 1)
```
torcolvin: I see this is the same in other code, but I don't understand why this is a buffered channel.

base/logging_context.go
```diff
@@ -51,6 +67,11 @@ type DbConsoleLogConfig struct {
 	LogKeys  *LogKeyMask
 }
 
+// DbAuditLogConfig can be used to customise the audit logging for events associated with this database.
+type DbAuditLogConfig struct {
+	EnabledEvents map[AuditID]struct{}
```
torcolvin, line 72: How does this intersect with default enabled events? Or is default enabled events something we are going to implement later?

rest/handler.go
```diff
@@ -725,26 +725,35 @@ func (h *handler) checkAuth(dbCtx *db.DatabaseContext) (err error) {
 		return nil
 	}
 
+	auditFields := make(base.AuditFields)
```
torcolvin: As discussed offline, I'm worried about the note at the top of this function: `// NOTE: checkAuth is not used for the admin interface.` and might want to see this in common code.

rest/handler.go
```diff
@@ -725,26 +725,35 @@ func (h *handler) checkAuth(dbCtx *db.DatabaseContext) (err error) {
...
+
 	// Record Auth stats
 	defer func(t time.Time) {
 		delta := time.Since(t).Nanoseconds()
 		dbCtx.DbStats.Security().TotalAuthTime.Add(delta)
 		if err != nil {
 			dbCtx.DbStats.Security().AuthFailedCount.Add(1)
+			if errors.Is(err, ErrInvalidLogin) {
```
torcolvin, line 737: do we want to log on other errors here? It seems like this would be important.

I'm not sure what the other errors would be - we could also add another field for error type

base/logging_context.go
bbrks: The set of default enabled events will make its way down to this field once wired up in [CBG-3975](https://issues.couchbase.com/browse/CBG-3975)

rest/handler.go
bbrks: I think other errors fall under operational logging and metrics more than auditing known events, but I could maybe be convinced otherwise?

torcolvin: https://github.com/couchbase/sync_gateway/blob/9737474c05067d3eeee6ed42f528b0f2160e7748/rest/handler.go#L818 seems like a niche error - disabled guest user? I'm actually not sure if it can fall into this pathway.

The argument for putting into the defer is that `checkAuth` is really hard to read and it's not clear if we've always wrapped the error correctly.

I think if you pass in an invalid cookie, you'll get an error that is not `ErrInvalidLogin` https://github.com/couchbase/sync_gateway/blob/9737474c05067d3eeee6ed42f528b0f2160e7748/auth/session.go#L42 but is also a 401 error.

I didn't look more exhaustively through the function.

rest/handler.go
bbrks: Acknowledged. I'll see if I can generalise/clean up the code around these two events.

base/logger_audit.go
bbrks: I don't see a good reason for it being `1` vs. `0`, nor can I can't see any info about that from the original PR #4511

This has reminded me that I wanted to pull this code into a shared fucntion though. Will test with an unbuffered channel and see if anything breaks.

rest/handler.go
bbrks: I partially solved this by choosing to have separate audit events for public API users and admin API users, which makes more sense anyway.

I'd prematurely implemented these events to put that demo together without thinking through the actual auditing use-cases.

base/audit_events.go
```diff
@@ -22,20 +34,42 @@ const (
...
 var AuditEvents = events{
-	AuditIDPlaceholder: {
-		Name:        "Placeholder audit event",
-		Description: "This is a placeholder.",
-		MandatoryFields: map[string]any{
-			"context": map[string]any{
-				"provider": "example provider",
-				"username": "alice",
-			},
+	AuditIDAuditEnabled: {}, // TODO: unreferenced event - somehow find this mistake via test or lint!
```
torcolvin, line 37: Do you want to file a ticket for this?

base/audit_events.go
```diff
@@ -61,7 +46,7 @@ var AuditEvents = events{
 		FilteringPermitted: false,
 		EventType:          eventTypeUser,
 	},
-	AuditIDUserAuthenticated: {
+	AuditIDPublicUserAuthenticated: {
 		Name:        "User authenticated",
```
torcolvin, line 50: Do we want to do something to indicate that this is over the public port? in the name or description? See also below.

torcolvin (approved):

base/audit_events.go
bbrks: Will tweak name and description in the follow-up full list of event definitions PR

base/audit_events.go
bbrks: https://issues.couchbase.com/browse/CBG-4037

PR #6909, opened 2024-06-21
CBG-4008 Support all config updates during async initialization

adamcfraser:
When using persistent config, all config updates (local or remote) should support async initialization, and not block on index readiness.

CBG-4008

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2535/

### DIFF
--- a/rest/indextest/index_test.go
+++ b/rest/indextest/index_test.go
@@ -799,6 +799,7 @@ func TestAsyncInitRemoteConfigUpdates(t *testing.T) {
 	// Update the db config from a remote node, verify change is picked up while in starting state
 	_, err = sc.BootstrapContext.UpdateConfig(ctx, bucketName, groupID, dbName, func(bucketDbConfig *rest.DatabaseConfig) (updatedConfig *rest.DatabaseConfig, err error) {
 		_, scopeName, collectionName, err := rest.ParseKeyspace(keyspace)
+		require.NoError(t, err)
 		if scopeName == nil || collectionName == nil {
 			bucketDbConfig.ImportFilter = nil
 		} else {
--- a/rest/indextest/index_test.go
+++ b/rest/indextest/index_test.go
@@ -261,10 +261,9 @@ func TestAsyncInitWithResync(t *testing.T) {
 	resp = rest.BootstrapAdminRequest(t, sc, http.MethodDelete, "/"+dbName+"/", "")
 	resp.RequireStatus(http.StatusOK)
 
-	rest.DropAllTestIndexes(t, tb)
+	rest.DropAllTestIndexesIncludingPrimary(t, tb)
 
 	// Set testing callbacks for async initialization
-
 	collectionCount := int64(0)
 	initStarted := make(chan error)
 	unblockInit := make(chan error)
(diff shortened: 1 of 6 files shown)


adamcfraser:
Integration test failure (TestResyncWithoutIndexes) is unrelated to this change.

torcolvin:
> Integration test failure (TestResyncWithoutIndexes) is unrelated to this change.

Until I realized this test was not even in this branch, I thought it could be related to the idea that the test doesn't delete the indexes correctly. Then I remembered that unlike `db/indextest`, `rest/indextest` uses flush to remove bucket and subsequent indexes. I think that's because we found it was flakey in `db/indextest` and not in `rest/indextest`. Here's a revived PR that puts together the code to use DCP purge and deletion of all indexes in case it is necessary for one of your PRs https://github.com/couchbase/sync_gateway/pull/6911

rest/indextest/index_test.go
```diff
@@ -596,6 +596,226 @@ func TestSyncOnline(t *testing.T) {
...
+	dbName := "db"
+	dbConfig := makeDbConfig(t, tb, syncFunc, importFilter)
+	dbConfig.Name = dbName
+	dbConfig.StartOffline = base.BoolPtr(true)
+
+	keyspace := dbName
+	if len(dbConfig.Scopes) > 0 {
+		keyspaces := getRESTKeyspaces(dbName, dbConfig.Scopes)
+		keyspace = keyspaces[0]
+	}
```
torcolvin, line 758:
```suggestion
		keyspace := getRESTKeyspaces(dbName, dbConfig.Scopes)[0]
```

This is always > 0, in fact it is always 3.

rest/indextest/index_test.go
```diff
@@ -596,6 +596,226 @@ func TestSyncOnline(t *testing.T) {
...
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("This test only works against Couchbase Server")
+	}
+	base.TestRequiresCollections(t)
+	base.SetUpTestLogging(t, base.LevelDebug, base.KeyHTTP)
+
+	sc, closeFn := rest.StartBootstrapServer(t)
+	defer closeFn()
+
+	// Set testing callbacks for async initialization
```
torcolvin, line 611:
```suggestion
	// Set testing callbacks for async initialization to pause after a single collection's indexes have been created.
```

We could actually shorten this test timeout to only create a single named collection and speed up these tests, since it will loop once for named collection and the default collection. This can make a pretty significant speedup over a bunch of tests, see https://github.com/couchbase/sync_gateway/pull/6911

rest/indextest/index_test.go
```diff
@@ -596,6 +596,226 @@ func TestSyncOnline(t *testing.T) {
...
+	require.NoError(t, err)
+	waitAndRequireDBState(t, sc, dbName, db.DBStarting)
+
+	// Attempt to update import filter while in starting mode
+	importFilter = "function(doc){ return false; }"
+	resp := rest.BootstrapAdminRequest(t, sc, http.MethodPut, "/"+keyspace+"/_config/import_filter", importFilter)
+	resp.RequireStatus(http.StatusOK)
+
+	resp = rest.BootstrapAdminRequest(t, sc, http.MethodGet, "/"+keyspace+"/_config/import_filter", "")
+	resp.RequireResponse(http.StatusOK, importFilter)
```
torcolvin, line 797: I'd prefer to make sure we don't miss steps here, and put the test cases for sync and import_filter and make sure Put/Get/Delete/Get workflow works. This could be pulled out from the code above.

torcolvin (dismissed):

rest/indextest/index_test.go
adamcfraser: What steps are you thinking we're missing here?  This test (TestAsyncInitRemoteConfigUpdates) is ensuring that async works correctly when the db is updated on a remote nod, while the previous test (TestAsyncInitConfigUpdates) validates all the different config update options locally.  I included the import filter change here to make sure the database wasn't in a state that would reject changes after processing a remote update.  Beyond that, I'd expect the coverage in TestAsyncInitRemoteConfigUpdates to suffice.

rest/indextest/index_test.go
adamcfraser: Do you think this merits this being handled differently than the rest of our test suite (i.e. using SG_TEST_COLLECTION_POOL_SIZE)?

rest/indextest/index_test.go
adamcfraser: Do you mean this is always 1 when base.TestRequiresCollections(t) passes?  I don't think it's ever three in my local testing.  I expect this code (copied from other tests) is just defensive.

torcolvin: I think it is be three because `dbConfig := makeDbConfig(t, tb, syncFunc, importFilter)` creates 3 collections always.

rest/indextest/index_test.go
torcolvin: I think we can drop this, but I thought we'd just be complete to keep import and sync. I do think that the combination of tests covers all cases.

torcolvin (approved):

PR #6914, opened 2024-06-24
CBG-4023: removal of unmarshalled body on the rev cache

gregns1:
CBG-4023

- Removed the body from revCacheValue and document revision structs 
- If body is needed it will be unmarshalled from the byte array on the rev cache now 
- Added new function `Inject1xBodyProperties` which is called from the AddDocToChangeEntryUsingRevCache code path and avoid unnecessary marshal work 

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2530/

### DIFF
--- a/db/revision_cache_interface.go
+++ b/db/revision_cache_interface.go
@@ -118,13 +118,13 @@ func newCollectionRevisionCache(revCache *RevisionCache, collectionID uint32) co
 }
 
 // Get is for per collection access to Get method
-func (c *collectionRevisionCache) Get(ctx context.Context, docID, revID string, includeBody, includeDelta bool) (DocumentRevision, error) {
-	return (*c.revCache).Get(ctx, docID, revID, c.collectionID, includeBody, includeDelta)
+func (c *collectionRevisionCache) Get(ctx context.Context, docID, revID string, includeDelta bool) (DocumentRevision, error) {
+	return (*c.revCache).Get(ctx, docID, revID, c.collectionID, includeDelta)
 }
 
 // GetActive is for per collection access to GetActive method
-func (c *collectionRevisionCache) GetActive(ctx context.Context, docID string, includeBody bool) (DocumentRevision, error) {
-	return (*c.revCache).GetActive(ctx, docID, c.collectionID, includeBody)
+func (c *collectionRevisionCache) GetActive(ctx context.Context, docID string) (DocumentRevision, error) {
+	return (*c.revCache).GetActive(ctx, docID, c.collectionID)
 }
 
 // Peek is for per collection access to Peek method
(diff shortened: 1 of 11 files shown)


db/revision_cache_interface.go
```diff
@@ -137,21 +132,64 @@ func (rev *DocumentRevision) MutableBody() (b Body, err error) {
...
 		return nil, err
 	}
 
-	// store a copy of the unmarshalled body for next time we need it
-	rev._shallowCopyBody = b
-
 	return b, nil
 }
 
+// Inject1xBodyProperties will inject special properties (_rev etc) into document body avoiding unnecessary marshal work
```
torcolvin, line 192: Do you think that there is sufficient test coverage for this? This looks like new logic to me.

I expect it would be mostly covered by 1x rest tests, but I am not always sure we actually test for all properties in each test due to need to change `_rev`.

db/revision_cache_interface.go
```diff
@@ -246,44 +283,40 @@ func newRevCacheDelta(deltaBytes []byte, fromRevID string, toRevision DocumentRe
...
 	}
 
 	return revCacheLoaderForDocument(ctx, backingStore, doc, id.RevID)
 }
 
 // Common revCacheLoader functionality used either during a cache miss (from revCacheLoader), or directly when retrieving current rev from cache
-func revCacheLoaderForDocument(ctx context.Context, backingStore RevisionCacheBackingStore, doc *Document, revid string) (bodyBytes []byte, body Body, history Revisions, channels base.Set, removed bool, attachments AttachmentsMeta, deleted bool, expiry *time.Time, err error) {
-	if bodyBytes, body, attachments, err = backingStore.getRevision(ctx, doc, revid); err != nil {
+func revCacheLoaderForDocument(ctx context.Context, backingStore RevisionCacheBackingStore, doc *Document, revid string) (bodyBytes []byte, history Revisions, channels base.Set, removed bool, attachments AttachmentsMeta, deleted bool, expiry *time.Time, err error) {
+	if bodyBytes, _, attachments, err = backingStore.getRevision(ctx, doc, revid); err != nil {
```
torcolvin: I think that this was the last place that body is used in getRevision, can we remove this as a return value?

db/revision_cache_interface.go
gregns1: `TestChangesIncludeDocs` seems to test it pretty well, but will get a test that directly calls the functions and ensuring all code paths in function work as expected. The logic was copied form what https://github.com/couchbase/sync_gateway/blob/59ff006805437ce8e6bf727704c47263aaac2784/db/revision_cache_interface.go#L210 does. Implementing this way avoids the extra unmarshal and marshal work. 

Added `TestInject1xBodyProperties` for this.

PR #6918, opened 2024-06-25
CBG-3975: Configuration wiring for audit logging and /db/_config APIs

bbrks:
CBG-3975

Configuration wiring for audit logging.

- Allows per-database enabled and events via /db/_config API.
- `GET /db/_config/audit` shows correct 'enabled' state for each available event. Read-only until [CBG-3981](https://issues.couchbase.com/browse/CBG-3981)

## Dependencies
- [x] #6908 

### DIFF
--- a/base/logger_audit.go
+++ b/base/logger_audit.go
@@ -16,7 +16,8 @@ import (
 )
 
 const (
-	auditLogName = "audit"
+	auditLogName        = "audit"
+	defaultAuditEnabled = false
 )
 
 // commonly used fields for audit events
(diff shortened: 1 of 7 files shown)


base/logger_audit.go
```diff
@@ -131,12 +132,28 @@ type AuditLogger struct {
 	config AuditLoggerConfig
 }
 
+func (l *AuditLogger) getAuditLoggerConfig() *AuditLoggerConfig {
+	c := AuditLoggerConfig{}
+	if l != nil {
+		// Copy config struct to avoid mutating running config
```
torcolvin, line 138: Does this help a lot given that all the member of AuditLoggerConfig are pointers and this is a shallow copy?

bbrks: Doesn't seem nessesary since there's no runtime configuration beyond the embedded FileLoggerConfig (yet), but I think I'd rather just keep this here in case we do add runtime-configurable options to `auditLogger`. Follows the same pattern as console and file loggers.

rest/config_database.go
```diff
@@ -90,17 +90,22 @@ func GenerateDatabaseConfigVersionID(ctx context.Context, previousRevID string,
...
+				LogKeys:  bootstrapLoggingCnf.Console.LogKeys,
 			}
 		}
 	}
-	return &DbLoggingConfig{}
+	if bootstrapLoggingCnf.Audit != nil {
+		dblc.Audit = &DbAuditLoggingConfig{
+			Enabled:       base.BoolPtr(false),
+			EnabledEvents: base.DefaultAuditEventIDs,
+		}
```
torcolvin, line 106: I would have expected these to inherit the value from the bootstrap and override if appropriate?

Is this change specific to capella, since audit logging should be enabled globally, but potentially turned off per database?

bbrks: It's anticipated that users will enable audit logging (globally), and then opt-into databases one at a time where required. Enabling globally then disabling subsets of databases creates unnessesary audit work/noise for the sake of one extra API call. It's expected that users will want to customise their enabled audit events anyway.

torcolvin (approved):

PR #6919, opened 2024-06-26
CBG-4010 make FileLogger.Close shut down its own goroutines

torcolvin:
- close the rotation gorountines when closing the file logger

This failure happens reasonably often in github and is reproduce on mac with high -count numbers after ~20 or so tries. This makes closing the logger slower, but we are only doing this in test so I think the wait time is OK.


bbrks (approved):

bbrks:
Thank you!

PR #6921, opened 2024-06-27
CBG-3976 allow injecting global fields into logger

torcolvin:
I tried to imagine if there was a more performanant way to accomplish this, but I didn't come up with anything obvious - the other alternative was to marshal the "real" output and then reconstruct the string with the new marshalled field. With 1 or 2 fields in the marshalled struct this wasn't faster in my experiments.

## Pre-review checklist

### DIFF
--- a/base/logger_audit.go
+++ b/base/logger_audit.go
@@ -111,11 +111,6 @@ func (f *AuditFields) merge(ctx context.Context, overwrites AuditFields) {
 
 // Audit creates and logs an audit event for the given ID and a set of additional data associated with the request.
 func Audit(ctx context.Context, id AuditID, additionalData AuditFields) {
-	audit(ctx, id, additionalData)
-}
-
-// audit creates and logs an audit event for the given ID and a set of additional data associated with the request, but allows passing a devmode field.
-func audit(ctx context.Context, id AuditID, additionalData AuditFields) {
 	var fields AuditFields
 
 	if IsDevMode() {
--- a/base/logger_audit.go
+++ b/base/logger_audit.go
@@ -35,7 +35,7 @@ const (
 )
 
 // expandFields populates data with information from the id, context and additionalData.
-func expandFields(id AuditID, ctx context.Context, globalFields AuditFields, additionalData AuditFields, isDevMode bool) AuditFields {
+func expandFields(id AuditID, ctx context.Context, globalFields AuditFields, additionalData AuditFields) AuditFields {
 	var fields AuditFields
 	if additionalData != nil {
 		fields = additionalData
@@ -87,14 +87,14 @@ func expandFields(id AuditID, ctx context.Context, globalFields AuditFields, add
 
 	fields[auditFieldTimestamp] = time.Now()
 
-	fields.merge(ctx, globalFields, isDevMode)
+	fields.merge(ctx, globalFields)
 	// TODO: CBG-3977 - Inject and merge data from request header
 
 	return fields
 }
 
 // Merge will perform a shallow overwrite of the fields in the AuditFields. If there are conflicts, do not overwrite but log a warning. This will panic in dev mode.
-func (f *AuditFields) merge(ctx context.Context, overwrites AuditFields, isDevMode bool) {
+func (f *AuditFields) merge(ctx context.Context, overwrites AuditFields) {
 	var duplicateFields []string
 	for k, v := range overwrites {
 		_, ok := (*f)[k]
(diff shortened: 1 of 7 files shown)


base/logger_audit.go
```diff
@@ -162,3 +166,15 @@ func (al *AuditLogger) shouldLog(id AuditID, ctx context.Context) bool {
...
+
+// baseFields returns the fields that are common to all audit events.
+func (al *AuditLogger) baseFields() AuditFields {
+	if al == nil {
+		return make(AuditFields)
+	}
+	fields := make(AuditFields, len(al.globalFields))
+	for k, v := range al.globalFields {
+		fields[k] = v
+	}
```
bbrks: This needs to be a "deep-merge", such that it won't overwrite all of `v` if it's an object.
Capella can be injecting just one field in a nested object to overwrite (e.g. `real_userid.username`

I was thinking we can write a function that recursively (for every `map[string]any`) adds or replaces any leaf properties in the object.

base/logger_audit.go
```diff
@@ -34,10 +34,12 @@ const (
...
 // expandFields populates data with information from the id, context and additionalData.
-func expandFields(id AuditID, ctx context.Context, additionalData AuditFields) AuditFields {
-	var fields AuditFields
+func expandFields(id AuditID, ctx context.Context, globalFields AuditFields, additionalData AuditFields) AuditFields {
+	fields := auditLogger.baseFields()
 	if additionalData != nil {
-		fields = additionalData
+		for k, v := range additionalData {
+			fields[k] = v
+		}
```
bbrks: Order of precedence here is backwards. Sorry, I said I'd update the Jira ticket with details and never got around to it.

We need SG's fields to be overwritten by the injected global fields, and then overwritten by the request-specific fields.

bbrks (changes requested):
Left some clarifying comments about how this merge behaviour should work. Let me know if you have any further questions

base/logger_audit.go
```diff
@@ -86,20 +86,45 @@ func expandFields(id AuditID, ctx context.Context, additionalData AuditFields) A
...
+		if isDevMode {
+			panic(msg)
+		}
+		WarnfCtx(ctx, msg)
+	}
+}
+
 // Audit creates and logs an audit event for the given ID and a set of additional data associated with the request.
 func Audit(ctx context.Context, id AuditID, additionalData AuditFields) {
+	audit(ctx, id, additionalData, IsDevMode())
```
bbrks: Do these new layers of indirection for `isDevMode` retain the benefit of the compiler removing devmode branches when not enabled (see #6898)? That was one of my largest concerns when adding it - make evaluation of it completely free for non-dev mode.

As far as I can tell the only benefit to parameterising `isDevMode` is to get test coverage of the warning? Is that really necessary? It clutters up the API quite a bit and introduces the above concern.

bbrks (commented):
I'm OK with this but not sure I see the reason why `isDevMode` is being parameterised and passed around everywhere now.

base/logger_audit.go
torcolvin: I think I should just drop panics because they are unlikely to happen in tests and there's more benefit to actually testing the production behavior than a panic to stop the behavior that isn't really going to be seen without capella's usage of it. This might be different for request handling vs global handling though.

base/logger_audit.go
```diff
@@ -87,20 +87,41 @@ func expandFields(id AuditID, ctx context.Context, additionalData AuditFields) A
...
+	}
+}
+
 // Audit creates and logs an audit event for the given ID and a set of additional data associated with the request.
 func Audit(ctx context.Context, id AuditID, additionalData AuditFields) {
+	audit(ctx, id, additionalData)
+}
+
+// audit creates and logs an audit event for the given ID and a set of additional data associated with the request, but allows passing a devmode field.
+func audit(ctx context.Context, id AuditID, additionalData AuditFields) {
```
bbrks:
```suggestion
```

base/logger_audit_test.go
```diff
@@ -0,0 +1,215 @@
...
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			ctx := TestCtx(t)
+			var err error
+			auditLogger, err = NewAuditLogger(ctx, &AuditLoggerConfig{FileLoggerConfig: FileLoggerConfig{Enabled: BoolPtr(true)}}, tmpdir, 0, nil, testCase.globalFields)
+			require.NoError(t, err)
+			startWarnCount := SyncGatewayStats.GlobalStats.ResourceUtilizationStats().WarnCount.Value()
+			output := AuditLogContents(t, func() {
+				// Test basic audit event
+				audit(ctx, AuditIDPublicUserAuthenticated, map[string]any{"method": "basic"})
```
bbrks:
```suggestion
				Audit(ctx, AuditIDPublicUserAuthenticated, map[string]any{"method": "basic"})
```

bbrks (dismissed):
Looks good, I think this `Audit->audit` thing is leftover from the devmode parameter

bbrks (approved):

PR #6929, opened 2024-07-01
CBG-4055: update api docs for all_docs endpoint

gregns1:
CBG-4055

- Small PR to update the query parameter types for all_docs

### DIFF
--- a/docs/api/components/parameters.yaml
+++ b/docs/api/components/parameters.yaml
@@ -25,7 +25,7 @@ Include-channels:
   in: query
   required: false
   schema:
-    type: string
+    type: boolean
   description: Include the channels each document is part of that the calling user also has access too.
 atts_since:
   name: atts_since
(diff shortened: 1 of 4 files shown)


torcolvin:
This looks right but I was having a look at the rest of the API and I have some questions about potential clarifications.

I'd probably make the description for POST and GET match because the first thing I looked at was to see how they were different.

I didn't test the following behavior, but I think it For POST I don't think keys is actually required - I think you can POST with an empty body and it will return the full set of docs. I think you can also post `keys: []` and it will return full set of docs.

There's a typo in this file: `databased` should be `database`.

gregns1:
Updated description. Did testing, you cannot run POST with empty body, you also can't run {"keys": []} to return all docs. You get empty result here.

torcolvin (approved):

PR #6930, opened 2024-07-01
﻿CBG-3977: Add request scoped context information.

torcolvin:
Change code to include original context information on non cancellable contexts. The one place we _might_ want to drop capella based context information is in ISGR, since the control plane user isn't the user who takes the action. Otherwise, I think all cases we want to preserve the information.

This PR has numerous fixme, which is for HTTP events that are not logged, and we don't have a way to test for.

The `LoggingContext` struct has stubs for  
```
	// TODO: CBG-3973 - Add request data (user, ips, etc.) - to be used by auditFieldsFromContext
	//// Username is the name of the authenticated user
	//Username string
	//// UserDomain can be syncgateway or couchbase depending on whether the authenticated user is a sync gateway user or a couchbase RBAC user
	//UserDomain string
	//// RequestPort is the HTTP port of the request associated with this log.
	//RequestPort string
	//// RemoteAddr is the IP and port of the remote client making the request associated with this log
	//RemoteAddr string
```

but I'm unsure if we want to support unmarshalling to these structs at the time of the request and populating the context concretely here. This means any change that app services makes will have to be modified by capella. OTOH, limiting the fields means we can do more validation.

## Pre-review checklist

- [ ] Link upstream PRs

### DIFF
--- a/base/logging_context.go
+++ b/base/logging_context.go
@@ -44,8 +44,8 @@ type LogContext struct {
 	// TestName can be a unit test name (see TestCtx)
 	TestName string
 
-	// AuditFields is a map of fields to be included in audit logs
-	AuditFields map[string]any
+	// RequestAdditionalAuditFields is a map of fields to be included in audit logs
+	RequestAdditionalAuditFields map[string]any
 	// TODO: CBG-3973 - Add request data (user, ips, etc.) - to be used by auditFieldsFromContext
 	//// Username is the name of the authenticated user
 	//Username string
@@ -121,14 +121,14 @@ func (lc *LogContext) getContextKey() LogContextKey {
 
 func (lc *LogContext) getCopy() LogContext {
 	return LogContext{
-		CorrelationID: lc.CorrelationID,
-		Database:      lc.Database,
-		DbLogConfig:   lc.DbLogConfig,
-		Bucket:        lc.Bucket,
-		Scope:         lc.Scope,
-		Collection:    lc.Collection,
-		TestName:      lc.TestName,
-		AuditFields:   lc.AuditFields,
+		CorrelationID:                lc.CorrelationID,
+		Database:                     lc.Database,
+		DbLogConfig:                  lc.DbLogConfig,
+		Bucket:                       lc.Bucket,
+		Scope:                        lc.Scope,
+		Collection:                   lc.Collection,
+		TestName:                     lc.TestName,
+		RequestAdditionalAuditFields: lc.RequestAdditionalAuditFields,
 	}
 }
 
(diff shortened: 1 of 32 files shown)


base/util.go
```diff
@@ -47,25 +47,16 @@ const (
...
-// NonCancellableContext is here to stroe a context that is not cancellable. Used to explicitly state when a change from
+// NonCancellableContext is here to store a context that is not cancellable. Used to explicitly state when a change from
 // a cancellable context to a context withoutr contex is required
 type NonCancellableContext struct {
 	Ctx context.Context
 }
 
 // NewNonCancelCtx creates a new background context struct for operations that require a fresh context
-func NewNonCancelCtx() NonCancellableContext {
+func NewNonCancelCtx(ctx context.Context) NonCancellableContext {
```
bbrks: I'm trying to think through the implications of passing the original request audit context down into this and I'm not sure we actually want it?

I agree with your ISGR comment where we don't want the original creator of the replication to be associated with all of the operations for that given replication, and I think that extends to other usages of this.

I think given this context is kept around for a long time (far longer than the original request), and am worried that we'd be putting that data into events that aren't from the request nor should be associated.

torcolvin: I would say that for ISGR we do want to reset this information (but we already do by resetting correlation ID).

For the requests on admin API to change db config I think that having the context information is helpful. Any error from `_applyConfig` synchronous API would be helpful to know if it came from config polling or a request.

Situations that are ambiguous - `DatabaseInitManager`. I think conceptually having this be associated with a request might make sense, it is first in to do an initialization.

Inside `_applyConfig` the code that we don't want to have a context associated with a request are:

- NewImportFilterFunction
- newEventHandlers

asyncOnline already resets its context https://github.com/couchbase/sync_gateway/blob/296de2ec955c18cc766ac76c0acbb978d7c0c614/rest/server_context.go#L1002

Inside `NewDatabaseContext` anything that returns logging for an error we'd want to have the logging on the request but some things do require a context reset.

- `newSequenceAllocator` spawns a goroutine that uses a context
- `NewChannelCacheForContext` starts Background tasks
- `NewCfgSg` will pick up context but reset correlation ID
- `UpdateSyncFn` will use a context for the `JSRunner` 
- `NewSgReplicateMgr` sets up a context as member of `sgReplicateManager resetting the correlation ID.

I think as a possible compromise, resetting the ctx here https://github.com/couchbase/sync_gateway/blob/296de2ec955c18cc766ac76c0acbb978d7c0c614/rest/server_context.go#L814 and also in ActiveReplicator.

I think we have to assume that a context will be cancelled once the request is ended. It's going to be really hard to spot either case if it is "wrong" because it might not be the trivial actions that trigger logging, e.g. enabling ISGR or import filter functions and you have to spot the issue carefully. It's probably not a huge deal for database config operations to lose request context but there can always be the question of whether this action was triggered by a POST/PUT or config polling.

bbrks: I am really not following how a context from a user-request to create a database needs to make it this deep into the actual instance of a database for auditing purposes, long afer the original create request has finished? Maybe we need to sync up on this so I can understand.

bbrks: Just seen your follow-up commits to back out of this change. I'd still like to understand what that would've given us, but we can revisit another time if you want.

base/logging_context.go
```diff
@@ -44,6 +44,8 @@ type LogContext struct {
 	// TestName can be a unit test name (see TestCtx)
 	TestName string
 
+	// AuditFields is a map of fields to be included in audit logs
+	AuditFields map[string]any
```
bbrks: I think we should rename this to `RequestAdditionalAuditFields` or similar, so it's obvious it's not the full set of fields (and to remind ourselves that this is injected data that cannot be used to overwrite existing fields).

bbrks (dismissed):
I think this change is good, with a nit about naming. I don't think this conflicts with my net/http request context changes.

bbrks (approved):

PR #6932, opened 2024-07-02
Add missing heap profile docs

torcolvin:


docs/api/components/schemas.yaml
```diff
@@ -2248,6 +2248,16 @@ Startup-config:
       description: TCP keep-alive interval between SG and Couchbase server
       type: integer
       readOnly: true
+    heap_profile_collection_threshold:
+      description: Threshold in bytes for collecting heap profiles automatically. If set, Sync Gateway will collect a memory profile when it exceeds this value. The default value will be set to 85% of the lesser of cgroup or system memory.
```
adamcfraser: "If set, Sync Gateway will collect a memory profile when it exceeds this value." sounds like profile collection won't happen if it's not set.  What about:
```suggestion
      description: Threshold in bytes for automatic collection of heap profiles. If not specified, defaults to 85% of the lesser of cgroup or system memory.
```

adamcfraser (changes requested):

adamcfraser (approved):

PR #6937, opened 2024-07-04
CBG-4015 Refresh _sync:seq when releasing sequences in nextSequenceGreaterThan

adamcfraser:
CBG-4015

Accounts for expected jumps in sequence values under variable allocator throughput before triggering “greater than” handling.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2547/

### DIFF
--- a/db/sequence_allocator.go
+++ b/db/sequence_allocator.go
@@ -187,6 +187,19 @@ func (s *sequenceAllocator) nextSequence(ctx context.Context) (sequence uint64, 
 	return sequence, nil
 }
 
+// _releaseCurrentBatch releases any unused sequences currently held by the allocator.
+// Writes unused sequence document while holding s.lock, shouldn't be used by high-throughput operations.
+func (s *sequenceAllocator) _releaseCurrentBatch(ctx context.Context) (numReleased uint64, err error) {
+	if s.max > s.last {
+		numReleased, err = s.releaseSequenceRange(ctx, s.last+1, s.max)
+		if err != nil {
+			return 0, err
+		}
+		s.last = s.max
+	}
+	return numReleased, nil
+}
+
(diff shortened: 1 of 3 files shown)


db/sequence_allocator.go
```diff
@@ -226,39 +239,58 @@ func (s *sequenceAllocator) nextSequenceGreaterThan(ctx context.Context, existin
...
-	// We can perform (a) and (b) as a single increment operation, but (c) and (d) aren't necessarily contiguous blocks and must be released
-	// separately
-
-	prevAllocReleaseFrom := s.last + 1
-	prevAllocReleaseTo := s.max
+	// At this point we need to allocate a sequence that's larger than what's in our current batch, so we first need to release the current batch.
+	var numReleasedBatch uint64
+	numReleasedBatch, err = s._releaseCurrentBatch(ctx)
+	if err != nil {
+		base.WarnfCtx(ctx, "Error returned when releasing sequence range [%d-%d] for existing batch. Will be handled by skipped sequence handling.  Error:%v", err)
```
gregns1:
```suggestion
		base.WarnfCtx(ctx, "Error returned when releasing sequence range [%d-%d] for existing batch. Will be handled by skipped sequence handling.  Error:%v", s.last+1, s.max, err)

```

gregns1 (changes requested):
Generally LGTM. Just noticed one log line tweak.

db/sequence_allocator.go
adamcfraser: Good catch, and this highlighted a case where we should be returning error instead of just logging.  Will fix both.

gregns1 (approved):

PR #6938, opened 2024-07-05
CBG-4003 Revocation handling for deleted roles

adamcfraser:
CBG-4003

Deleted roles were not triggering revocation of channels associated with those roles, for two reasons:
- deleted roles weren't being included in the role set used by RevokedCollectionChannels to identify revoked channels
- channel history wasn't being updated on deleted roles for non-default collections

Deleted roles are now populated on the user object, and a new API added (GetRolesIncDeleted) that returns the union of deleted and non-deleted roles.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2548/

### DIFF
--- a/rest/user_api_test.go
+++ b/rest/user_api_test.go
@@ -1580,3 +1580,66 @@ func TestUserMultipleDBs(t *testing.T) {
 		})
 	}
 }
+
+func TestDeletedRoleChanHistory(t *testing.T) {
+	tests := []struct {
+		defaultCollection bool
+	}{
+		{defaultCollection: true},
+		{defaultCollection: false},
+	}
+
+	for _, test := range tests {
+		t.Run(fmt.Sprintf("defaultCollection:%v", test.defaultCollection), func(t *testing.T) {
+			ctx := base.TestCtx(t)
+
+			var rt *RestTester
+			if test.defaultCollection {
+				rt = NewRestTesterDefaultCollection(t, nil)
+			} else {
+				rt = NewRestTesterMultipleCollections(t, nil, 1)
+			}
+
+			collection := rt.GetSingleTestDatabaseCollection()
+			defer rt.Close()
+
+			const roleName = "role"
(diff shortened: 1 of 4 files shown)


rest/user_api_test.go
```diff
@@ -1580,3 +1580,66 @@ func TestUserMultipleDBs(t *testing.T) {
 		})
 	}
 }
+
+func TestDeletedRoleChanHistory(t *testing.T) {
```
torcolvin, line 1584: Will this test work with GSI=false and thus collectiosn are disabled?

rest/user_api_test.go
```diff
@@ -1580,3 +1580,66 @@ func TestUserMultipleDBs(t *testing.T) {
...
+			// Get role - ensure accessible
+			role, err := a.GetRole(roleName)
+			assert.NoError(t, err)
+			assert.NotNil(t, role)
+			if collection.IsDefaultCollection() {
+				assert.Len(t, role.Channels(), 2)
+				assert.True(t, role.Channels().Contains("channel"))
+			} else {
+				assert.Len(t, role.GetCollectionsAccess()[collection.ScopeName][collection.Name].Channels(), 2)
+				assert.True(t, role.GetCollectionsAccess()[collection.ScopeName][collection.Name].Channels().Contains("channel"))
```
torcolvin: You can use `role.CollectionChannels(collection.ScopeName, collection.Name)` to avoid this if statement. There is also `role.CollectionChannelHistory` for the second call.

I thought about if we can use this in the `DeleteRole` API but I don't think so easily. I have thought about whether removing `Channels()` from user/role object function directory would have made this bug harder to come by. The json objects themselves would maintain the same structure but the accessor functions could require collection specification.

torcolvin (commented):
LGTM, some testing comments

rest/user_api_test.go
adamcfraser: NewRestTesterMultipleCollections has a test skip for that scenario.

gregns1 (approved):
LGTM

PR #6941, opened 2024-07-05
CBG-3973: Pass HTTP request information into context for audit log data

bbrks:
CBG-3973

- Pass HTTP request information into context for audit log data
  - request host (API addr)
  - remote addr (client IP/port)
  - `real_userid` populatd by auth'd user (public or admin)
- Converted existing audit test to more generic sub-test style to allow future coverage

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2582/

### DIFF
--- a/base/logging_context.go
+++ b/base/logging_context.go
@@ -227,6 +227,7 @@ type userIDDomain string
 const (
 	UserDomainSyncGateway userIDDomain = "sgw"
 	UserDomainCBServer    userIDDomain = "cbs"
+	UserDomainBuiltin                  = "builtin" // internal users (e.g. SG bootstrap user)
 )
 
 func UserLogCtx(parent context.Context, username string, domain userIDDomain) context.Context {
--- a/base/logging_context.go
+++ b/base/logging_context.go
@@ -238,12 +238,14 @@ func UserLogCtx(parent context.Context, username string, domain userIDDomain) co
 }
 
 type RequestData struct {
+	CorrelationID     string
 	RequestHost       string
 	RequestRemoteAddr string
 }
 
 func RequestLogCtx(parent context.Context, d RequestData) context.Context {
 	newCtx := getLogCtx(parent)
+	newCtx.CorrelationID = d.CorrelationID
 	newCtx.RequestHost = d.RequestHost
 	newCtx.RequestRemoteAddr = d.RequestRemoteAddr
 	return LogContextWith(parent, &newCtx)
--- a/base/logging_context.go
+++ b/base/logging_context.go
@@ -222,6 +222,13 @@ func KeyspaceLogCtx(parent context.Context, bucketName, scopeName, collectionNam
 	return LogContextWith(parent, &newCtx)
 }
 
+func UserLogCtx(parent context.Context, user, domain string) context.Context {
+	newCtx := getLogCtx(parent)
+	newCtx.Username = user
+	newCtx.UserDomain = domain
+	return LogContextWith(parent, &newCtx)
+}
+
 type RequestData struct {
 	RequestHost       string
 	RequestRemoteAddr string
--- a/base/logging_context.go
+++ b/base/logging_context.go
@@ -49,7 +49,7 @@ type LogContext struct {
 	// Username is the name of the authenticated user
 	Username string
 	// UserDomain can be syncgateway or couchbase depending on whether the authenticated user is a sync gateway user or a couchbase RBAC user
-	UserDomain string
+	UserDomain userIDDomain
 	// RequestHost is the HTTP Host of the request associated with this log.
 	RequestHost string
 	// RequestRemoteAddr is the IP and port of the remote client making the request associated with this log
@@ -222,9 +222,16 @@ func KeyspaceLogCtx(parent context.Context, bucketName, scopeName, collectionNam
 	return LogContextWith(parent, &newCtx)
 }
 
-func UserLogCtx(parent context.Context, user, domain string) context.Context {
+type userIDDomain string
+
+const (
+	UserDomainSyncGateway userIDDomain = "sgw"
+	UserDomainCBServer    userIDDomain = "cbs"
+)
+
+func UserLogCtx(parent context.Context, username string, domain userIDDomain) context.Context {
 	newCtx := getLogCtx(parent)
-	newCtx.Username = user
+	newCtx.Username = username
 	newCtx.UserDomain = domain
 	return LogContextWith(parent, &newCtx)
 }
--- a/base/logging_context.go
+++ b/base/logging_context.go
@@ -46,15 +46,14 @@ type LogContext struct {
 
 	// RequestAdditionalAuditFields is a map of fields to be included in audit logs
 	RequestAdditionalAuditFields map[string]any
-	// TODO: CBG-3973 - Add request data (user, ips, etc.) - to be used by auditFieldsFromContext
-	//// Username is the name of the authenticated user
-	//Username string
-	//// UserDomain can be syncgateway or couchbase depending on whether the authenticated user is a sync gateway user or a couchbase RBAC user
-	//UserDomain string
-	//// RequestPort is the HTTP port of the request associated with this log.
-	//RequestPort string
-	//// RemoteAddr is the IP and port of the remote client making the request associated with this log
-	//RemoteAddr string
+	// Username is the name of the authenticated user
+	Username string
+	// UserDomain can be syncgateway or couchbase depending on whether the authenticated user is a sync gateway user or a couchbase RBAC user
+	UserDomain string
+	// RequestHost is the HTTP Host of the request associated with this log.
+	RequestHost string
+	// RequestRemoteAddr is the IP and port of the remote client making the request associated with this log
+	RequestRemoteAddr string
 
 	// implicitDefaultCollection is set to true when the context represents the default collection, but we want to omit that value from logging to prevent verbosity.
 	implicitDefaultCollection bool
@@ -132,6 +131,10 @@ func (lc *LogContext) getCopy() LogContext {
 		Collection:                   lc.Collection,
 		TestName:                     lc.TestName,
 		RequestAdditionalAuditFields: lc.RequestAdditionalAuditFields,
+		Username:                     lc.Username,
+		UserDomain:                   lc.UserDomain,
+		RequestHost:                  lc.RequestHost,
+		RequestRemoteAddr:            lc.RequestRemoteAddr,
 	}
 }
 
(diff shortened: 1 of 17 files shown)


base/logging_context.go
```diff
@@ -206,6 +209,33 @@ func KeyspaceLogCtx(parent context.Context, bucketName, scopeName, collectionNam
 	return LogContextWith(parent, &newCtx)
 }
 
+type userIDDomain string
+
+const (
+	UserDomainSyncGateway userIDDomain = "sgw"
+	UserDomainCBServer    userIDDomain = "cbs"
+	UserDomainBuiltin                  = "builtin" // internal users (e.g. SG bootstrap user)
```
torcolvin, line 230: I don't think this will matter, but:

```suggestion
	UserDomainBuiltin   userIDDomain  = "builtin" // internal users (e.g. SG bootstrap user)
```

rest/audit_test.go
```diff
@@ -46,58 +52,163 @@ func TestAuditInjectableHeader(t *testing.T) {
...
+			name: "guest request",
+			auditableAction: func(t testing.TB) {
+				RequireStatus(t, rt.SendRequest(http.MethodGet, "/db/_session", ""), http.StatusOK)
+			},
+			expectedAuditEventFields: map[base.AuditID]base.AuditFields{
+				base.AuditIDPublicUserAuthenticated: {
+					"cid":         auditFieldValueIgnored,
+					"auth_method": "guest",
+				},
+				// TODO: GUEST view session event?
```
torcolvin: Should there be a ticket for this?

rest/audit_test.go
```diff
@@ -46,58 +52,163 @@ func TestAuditInjectableHeader(t *testing.T) {
...
 		{
-			name: "admin request",
-			fn:   rt.SendAdminRequestWithHeaders,
+			name: "guest request",
+			auditableAction: func(t testing.TB) {
+				RequireStatus(t, rt.SendRequest(http.MethodGet, "/db/_session", ""), http.StatusOK)
+			},
+			expectedAuditEventFields: map[base.AuditID]base.AuditFields{
+				base.AuditIDPublicUserAuthenticated: {
+					"cid":         auditFieldValueIgnored,
```
torcolvin: I've been trending toward using constants for this and making the field names public from the base package.

rest/audit_test.go
```diff
@@ -46,58 +52,163 @@ func TestAuditInjectableHeader(t *testing.T) {
...
+			},
 		},
-		*/
 	}
 	for _, testCase := range testCases {
-		headers := map[string]string{
-			headerName:      `{"extra":"field"}`,
-			"Authorization": getBasicAuthHeader("alice", RestTesterDefaultUserPassword),
-		}
 		t.Run(testCase.name, func(t *testing.T) {
```
torcolvin: I made this mistake -

```suggestion
		rt.Run(testCase.name, func(t *testing.T) {
```

It might also be that we want not pass in `t` to auditableAction and just use rt.TB()

rest/audit_test.go
```diff
@@ -46,58 +52,163 @@ func TestAuditInjectableHeader(t *testing.T) {
...
-				RequireStatus(t, testCase.fn(http.MethodGet, "/db/", "", headers), http.StatusOK)
-			})
-			events := bytes.Split(output, []byte("\n"))
-			foundEvent := false
-			for _, rawEvent := range events {
-				if bytes.TrimSpace(rawEvent) == nil {
+			output := base.AuditLogContents(t, testCase.auditableAction)
+			events := jsonLines(t, output)
+
+			// for each event, check the fields match what we expected
```
torcolvin, line 204: do we want to assert that specific events occur only once? I am not sure.

rest/server_context.go
```diff
@@ -254,9 +280,11 @@ func (sc *ServerContext) Close(ctx context.Context) {
 	sc.invalidDatabaseConfigTracking.dbNames = nil
 
 	for _, s := range sc._httpServers {
-		base.InfofCtx(ctx, base.KeyHTTP, "Closing HTTP Server: %v", s.addr)
-		if err := s.server.Close(); err != nil {
-			base.WarnfCtx(ctx, "Error closing HTTP server %q: %v", s.addr, err)
+		if s.server != nil {
```
torcolvin, line 283: I'm not opposed to doing a nil check here, but I don't see how this would end up nil, and I'm worried we don't do the nil checks in the other places we need.

I assume that this is from the rest tester usage.

rest/utilities_testing.go
```diff
@@ -500,7 +504,12 @@ func (rt *RestTester) GetDatabase() *db.DatabaseContext {
 
 // CreateUser creates a user with the default password and channels scoped to a single test collection.
 func (rt *RestTester) CreateUser(username string, channels []string) {
-	response := rt.SendAdminRequest(http.MethodPut, "/{{.db}}/_user/"+username, GetUserPayload(rt.TB(), "", RestTesterDefaultUserPassword, "", rt.GetSingleTestDatabaseCollection(), channels, nil))
+	var response *TestResponse
+	if rt.AdminInterfaceAuthentication {
+		response = rt.SendAdminRequestWithAuth(http.MethodPut, "/{{.db}}/_user/"+username, GetUserPayload(rt.TB(), "", RestTesterDefaultUserPassword, "", rt.GetSingleTestDatabaseCollection(), channels, nil), base.TestClusterUsername(), base.TestClusterPassword())
```
torcolvin, line 509: I think this code is fine but I've hit this type of bug before. I wonder (for future) if we should do something so that we always call `SendAdminRequestWithAuth` and then it returns empty string if the admin interface is configured without authentication.

rest/utilities_testing.go
```diff
@@ -1198,6 +1211,7 @@ func Request(method, resource, body string) *http.Request {
 	if err != nil {
 		panic(fmt.Sprintf("http.NewRequest failed: %v", err))
 	}
+	request.RemoteAddr = ":0"     // usually populated by actual HTTP requests going into a HTTP Server
```
torcolvin: Take or leave:

Does this make sense to use a fake number that couldn't be a real port number? I think it would have to be an int, maybe 65536? We could use a string if the int isn't parsed.

rest/handler.go
```diff
@@ -203,11 +203,24 @@ func newHandler(server *ServerContext, privs handlerPrivs, serverType serverType
 // ctx returns the request-scoped context for logging/cancellation.
 func (h *handler) ctx() context.Context {
 	if h.rqCtx == nil {
-		h.rqCtx = base.CorrelationIDLogCtx(h.rq.Context(), h.formatSerialNumber())
+		ctx := base.CorrelationIDLogCtx(h.rq.Context(), h.formatSerialNumber())
```
torcolvin: Reconstructing contexts is cheap but not free since it is does a copy. Should we change `RequestLogCtx` to set a correlation id also, so we only have to construct this once?

rest/handler.go
```diff
@@ -203,11 +203,24 @@ func newHandler(server *ServerContext, privs handlerPrivs, serverType serverType
...
+		serverAddr, err := h.getServerAddr()
+		if err != nil {
```
torcolvin, line 207: If this errors in prod, this will be empty string, I wonder if it makes sense to say "unable to determine" or something.

rest/handler.go
```diff
@@ -755,6 +769,12 @@ func (h *handler) checkPublicAuth(dbCtx *db.DatabaseContext) (err error) {
 			}
 		} else {
 			dbCtx.DbStats.Security().AuthSuccessCount.Add(1)
+
+			if h.isGuest() {
```
torcolvin, line 792: Can you pass an CB Server admin user on the public port and have it get authorized?

rest/handler.go
```diff
@@ -755,6 +769,12 @@ func (h *handler) checkPublicAuth(dbCtx *db.DatabaseContext) (err error) {
...
+				h.rqCtx = base.UserLogCtx(h.ctx(), base.GuestUsername, base.UserDomainSyncGateway)
+			} else if h.user != nil {
```
torcolvin: What's the else case here? Is for a handler that doesn't need a user? Should this be logged?

rest/audit_test.go
bbrks: Will swap to `rt.Run` but will leave `t` in `auditableAction`. It's too easy to forget and rely on the parent `t` which is already in scope otherwise.

rest/audit_test.go
bbrks: I'll do that also as a guard to ensure future subtest writers don't include multiple auditable actions in one `AuditLogContents`

rest/server_context.go
bbrks: Yeah, `RestTester` doesn't start up an actual HTTP server, so falling into this code now results in a panic because we populate a "fake" list of servers on ServerContext for the reason of grabbing `serverInfo` for stamping `s.addr`

rest/utilities_testing.go
bbrks: Yeah I wasn't quite sure how far to push that change down. I can see legitimate uses for sending unauthorized requests to ensure the negative cases are handled, but that does seem like a more special case (e.g. `SendAdminRequestNoAuth`)

rest/utilities_testing.go
bbrks: I'll put `test.client.addr:99999` to make it even more obvious it's a fake RemoteAddr value. `SplitHostPort` doesn't do any validation of port numbers so it's safe.

rest/handler.go
bbrks: In the actual value? I'm not sure it's worth the special case (audit logging code will parse this value to split host and port)

The scenario would be that we start up an API server, don't include it on the `ServerContext`'s `_httpServers` map for some reason.

I think we could probably get test coverage to ensure this is working correctly.

rest/handler.go
bbrks: no

rest/handler.go
bbrks: I think this is leftovers from when this code was in a different place. Reasonably confident I can remove the if

rest/handler.go
torcolvin: This is fine to skip, I forgot that audit logging code also parses this and we handle there.

rest/audit_test.go
bbrks: As it turns out, there's nothing session specific about `GET /_session` so I think it makes sense to forget about this.

torcolvin (dismissed):

torcolvin (approved):

PR #6942, opened 2024-07-06
CBG-4025 Invalidate user roles after resync

adamcfraser:
CBG-4025

As part of this change, modified access invalidation functions to perform invalidation in a single update to the principal document, instead of once per collection (and one additional time for roles).  To support this:
- Switched existing invalidation functions (invalRoleChannels, invalUserChannels, etc) to have *DatabaseContext receiver instead of `*DatabaseCollection`, and added a `ScopeAndCollectionNames` parameter to specify the set of collections that should have access invalidated.
- For ease of use, maintained the existing invalidation functions on `*DatabaseCollection` - they now just call in to the `*DatabaseContext` functions with their single collection
- Added a new invalUserRolesAndChannels to invalidate a user’s roles and channels in a single user doc update.  Only currently used by resync

Query based resync still processes a single collection’s updates at a time - it’s structured a bit differently and didn’t seem to be worth refactoring at this point.  It has been updated to properly invalidate user roles.

The new test TestResyncInvalidatePrincipals covers the fix - have verified it with `SG_TEST_USE_DEFAULT_COLLECTION=true/false`.  Also made a test utility change to remove the password parameter from GetRolePayload since roles don’t have passwords.

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2551/

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -441,11 +441,11 @@ func (auth *Authenticator) UpdateSequenceNumber(p Principal, seq uint64) error {
 }
 
 func (auth *Authenticator) InvalidateDefaultChannels(name string, isUser bool, invalSeq uint64) error {
-	return auth.InvalidateChannels(name, isUser, base.DefaultScope, base.DefaultCollection, invalSeq)
+	return auth.InvalidateChannels(name, isUser, base.ScopeAndCollectionNames{base.DefaultScopeAndCollectionName()}, invalSeq)
 }
 
 // Invalidates the channel list of a user/role by setting the ChannelInvalSeq to a non-zero value
-func (auth *Authenticator) InvalidateChannels(name string, isUser bool, scope string, collection string, invalSeq uint64) error {
+func (auth *Authenticator) InvalidateChannels(name string, isUser bool, collections base.ScopeAndCollectionNames, invalSeq uint64) error {
 	var princ Principal
 	var docID string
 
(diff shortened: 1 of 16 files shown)


auth/auth.go
```diff
@@ -465,17 +465,23 @@ func (auth *Authenticator) InvalidateChannels(name string, isUser bool, scope st
...
-		err := subdocStore.SubdocInsert(auth.LogCtx, docID, subdocPath, 0, invalSeq)
-		if err != nil && err != base.ErrAlreadyExists && err != base.ErrPathExists && err != base.ErrPathNotFound && !base.IsDocNotFoundError(err) {
-			return err
+		subdocPath := "channel_inval_seq"
+		if scope != base.DefaultScope || collection != base.DefaultCollection {
+			subdocPath = "collection_access." + scope + "." + collection + "." + subdocPath
+		}
+
+		if subdocStore, ok := base.AsSubdocStore(auth.datastore); ok {
+			err := subdocStore.SubdocInsert(auth.LogCtx, docID, subdocPath, 0, invalSeq)
```
torcolvin, line 479: `sgbucket.SubDocStore` is part of `sgbucket.DataStore` so we don't need to do this cast.

auth/auth.go
```diff
@@ -465,17 +465,23 @@ func (auth *Authenticator) InvalidateChannels(name string, isUser bool, scope st
 
 	base.InfofCtx(auth.LogCtx, base.KeyAccess, "Invalidate access of %q", base.UD(name))
 
-	subdocPath := "channel_inval_seq"
-	if scope != base.DefaultScope || collection != base.DefaultCollection {
-		subdocPath = "collection_access." + scope + "." + collection + "." + subdocPath
-	}
+	// Attempt to use subdoc if we're only modifying a single collection
+	if len(collections) == 1 {
```
torcolvin, line 469: This is more of an optimization than might be necessary, but can we change the API of SubdocInsert to be map[string]any so we can do multiple subdoc operations at once?

Later we are going to run through the logic of updating the document, though not the write operation.

adamcfraser: I considered adding that functionality in this PR, but decided against it due to the increase in scope/risk.  Since the more common usage of InvalidateChannels (at write time) is always single collection, I thought it was reasonable to omit.

auth/auth.go
adamcfraser: AsSubdocStore is also doing a check on ds.IsSupported(sgbucket.BucketStoreFeatureSubdocOperations) (which I believe is still relevant for walrus support for backports to 3.1 that support walrus).  I think I'd prefer to leave this code as it was in this PR since we expect a backport.

gregns1 (approved):
LGTM

PR #6949, opened 2024-07-09
CBG-4048 add audit events for replication

torcolvin:
- For events that are on the admin api, should they have `eventTypeAdmin`, or if they are triggered by an admin user are they going to have `eventTypeUser`? For requests that are actually over the admin OR user port, such as `_blipsync`, how should these get typed? It is my understanding that this is only going to be used for user documentation.
- ~correlation ID is overloaded for us, it can mean blip correlation ID or http correlation id. I tried to create a separate field but its name is really really long. Given that the logs are going to be compressed, are long field names a problem?~ We are going to determine whether it is blip or http based on whether it is `[id]` or `#id`
- ~If we are creating replications as a result of creating a database, should these get logged? This might indicate we should put the logging code inside `sg_replicate_cfg.go`. The caution I have about doing this is that this is tricky because some of the functions to create local maps of replications occur because another SG in the same cluster created the replication and this SG is just picking it up for sharding. In this case, we don't know which user caused this action so I was thinking we would only log that on the SG node that had the REST call.~ Decided offline to only log admin actions initiated directly as part of REST API.

There are a few pathways to change the state on ISGR replications:

PUT /db/_replication -> this will create a replication and start it, unless `initialState` is not `running`. Therefore, it seems like a this should generally be both a Create and Start audit event unless `initialState` is specified to something other than running.
POST /db/_replicationStatus/replicationID?action= or POST /db/_replication/replicationID with `initialState` -> this can update a replication.

Due to these complexities, I think it is worth adding some tests for the cases above to make sure we get multiple events if we want them. I am holding off doing this based on whether you think this logic belongs in the rest package or the db package? I thought about not returning the auditEventIDs back to the rest call, but you have to be very careful not to log if there is any error and that seemed easiest to do from the rest handlers since they check for error before returning.

## Pre-review checklist

### DIFF
--- a/base/audit_events.go
+++ b/base/audit_events.go
@@ -157,9 +157,13 @@ var AuditEvents = events{
 		EnabledByDefault:   true,
 		FilteringPermitted: false,
 		MandatoryFields: AuditFields{
-			"db":          "database name",
-			"audit_scope": "global or db",
-			"new_config":  "JSON representation of new db audit config",
+			"audit_scope":     "global or db",
+			AuditFieldPayload: "JSON representation of new db audit config",
+		},
+		mandatoryFieldGroups: []fieldGroup{
+			fieldGroupDatabase,
+			fieldGroupRequest,
+			// fieldGroupAuthenticated, // FIXME: CBG-3973
 		},
 		EventType: eventTypeAdmin,
 	},
@@ -355,8 +359,8 @@ var AuditEvents = events{
 		Name:        "Create database",
 		Description: "A new database was created",
 		MandatoryFields: AuditFields{
-			"db":     "database name",
-			"config": "JSON representation of db config",
+			"db":              "database name",
+			AuditFieldPayload: "JSON representation of db config",
 		},
 		EnabledByDefault:   true,
 		FilteringPermitted: false,
@@ -397,8 +401,8 @@ var AuditEvents = events{
 		Name:        "Update database config",
 		Description: "Database configuration was updated",
 		MandatoryFields: AuditFields{
-			"db":     "database name",
-			"config": "JSON representation of new db config",
+			"db":              "database name",
+			AuditFieldPayload: "payload",
 		},
 		EnabledByDefault:   true,
 		FilteringPermitted: false,
@@ -636,7 +640,6 @@ var AuditEvents = events{
 		Name:        "Replication connect",
 		Description: "A replication client connected",
 		MandatoryFields: AuditFields{
-			"cid":         "context id", // this is blip context id
 			"client_type": "isgr, cbl, other",
 		},
 		OptionalFields: AuditFields{
@@ -655,7 +658,6 @@ var AuditEvents = events{
 		Name:        "Replication disconnect",
 		Description: "A replication client disconnected",
 		MandatoryFields: AuditFields{
-			"cid":         "context id", // this is blip context id
 			"client_type": "isgr, cbl, other",
 		},
 		mandatoryFieldGroups: []fieldGroup{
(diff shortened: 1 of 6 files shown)


base/audit_events.go
```diff
@@ -25,6 +25,21 @@ const (
 	AuditIDPublicUserAuthenticationFailed AuditID = 53261
 
 	AuditIDReadDatabase AuditID = 53301
+
+	// BLIP Replication events
+	AuditIDReplicationConnect    AuditID = 54300
```
bbrks: Sorry, I somehow didn't realise I'd not committed the list of events that I had previously shared! I've just opened a PR #6950 to do that.

If we can get that merged, I'll be able to review your changes vs. the original event definitions.

base/audit_events.go
```diff
@@ -636,13 +636,17 @@ var AuditEvents = events{
 		Name:        "Replication connect",
 		Description: "A replication client connected",
 		MandatoryFields: AuditFields{
-			"db":  "database name",
-			"cid": "context id",
+			"cid":         "context id", // this is blip context id
```
adamcfraser: Is this different than the correlation ID (present in fieldGroupRequest)

base/audit_events.go
```diff
@@ -685,9 +702,13 @@ var AuditEvents = events{
 		Name:        "Update Inter-Sync Gateway Replication",
 		Description: "Inter-Sync Gateway Replication was updated",
 		MandatoryFields: AuditFields{
-			"db":             "database name",
-			"replication_id": "replication id",
-			"config":         "JSON representation of new replication config",
+			AuditFieldReplicationID: "replication id",
+			"config":                "JSON representation of new replication config",
```
adamcfraser: I found "new" to be a bit ambiguous here: since this endpoint supports upsert, "new" sounded like it could be the result of the merge (when it's actually, correctly, the incoming payload).   Should we use something like "payload" consistently for all events when referring to the body of the incoming request?

db/sg_replicate_cfg.go
```diff
@@ -1123,11 +1125,21 @@ func (m *sgReplicateManager) UpsertReplication(ctx context.Context, replication
...
+		switch cluster.Replications[replication.ID].TargetState {
+		case ReplicationStateRunning:
+			// by default a replication will be created running, or an existing replication can go from stopped to running
+			auditEvents = append(auditEvents, base.AuditIDISGRStart)
+		case ReplicationStateStopped:
+			// if initial creation, then there is no starting
+		case ReplicationStateResetting:
+			// a replication can start in a resetting state, or go from stop->reset
+			auditEvents = append(auditEvents, base.AuditIDISGRReset)
+		}
```
adamcfraser: I'm not sure we actually want multiple events fired if someone upserts a config (i.e AuditIDISGRUpdate as well as AuditIDISGRStart).  I think it would be clearer to reserve AuditIDISGRStart/Stop/Reset for the PUT _replicationStatus endpoint only.

adamcfraser (changes requested):
I think the other questions in the PR description have been answered offline, but if not, raise them there and we can discuss again.

base/audit_events.go
torcolvin: This is not different, this is actually a duplicate. As discussed offline, we are going to rely on the difference in the format:

blip `"[" + contextID + "]"` and HTTP `#%03d`. I will remove this.

base/audit_events.go
torcolvin: Agreed, I think I want to change this everywhere, but let me know if this doesn't make sense.

db/sg_replicate_cfg.go
torcolvin: Agreed, the payload can determine this.

adamcfraser (approved):

adamcfraser:
LGTM

PR #6950, opened 2024-07-09
Add the initial working set of audit events

bbrks:
I forgot to open this PR when I shared the spreadsheet of audit events.

Given there's not too much that's controversial with the events themselves (other than "EnabledByDefault" and "Filterable" values) I propose we get this list merged before we start duplicating work to define events.

### DIFF
--- a/base/audit_events.go
+++ b/base/audit_events.go
@@ -18,13 +18,105 @@ const (
 	auditdIDBlockSize = 0xFFF
 )
 
+// Audit IDs for Sync Gateway.
+// The IDs and grouping is arbitrary, but are spaced to allow for some additions in each area.
 const (
-	AuditIDAuditEnabled AuditID = 53248
+	// Audit events
+	AuditIDAuditEnabled       AuditID = 53248
+	AuditIDAuditDisabled      AuditID = 53249
+	AuditIDAuditConfigChanged AuditID = 53250
 
-	AuditIDPublicUserAuthenticated        AuditID = 53260
-	AuditIDPublicUserAuthenticationFailed AuditID = 53261
+	// SG events
+	AuditIDSyncGatewayStartup AuditID = 53260
 
-	AuditIDReadDatabase AuditID = 53301
+	// API events
+	AuditIDPublicHTTPAPIRequest  AuditID = 53270
+	AuditIDAdminHTTPAPIRequest   AuditID = 53271
+	AuditIDMetricsHTTPAPIRequest AuditID = 53272
+	//AuditIDBLIPMessage           AuditID = 0 // TODO: Is this required? What info should the event contain?
+
+	// Auth (public) events
+	AuditIDPublicUserAuthenticated        AuditID = 53280
+	AuditIDPublicUserAuthenticationFailed AuditID = 53281
+	AuditIDPublicUserSessionCreated       AuditID = 53282
+	AuditIDPublicUserSessionDeleted       AuditID = 53283
+	// Auth (admin) events
+	AuditIDAdminUserAuthenticated        AuditID = 53290
+	AuditIDAdminUserAuthenticationFailed AuditID = 53291
+	AuditIDAdminUserAuthorizationFailed  AuditID = 53292
+
+	// SG node events
+	AuditIDSyncGatewayCollectInfoStatus AuditID = 53300
+	AuditIDSyncGatewayCollectInfoStart  AuditID = 53301
+	AuditIDSyncGatewayCollectInfoStop   AuditID = 53302
+	AuditIDSyncGatewayStats             AuditID = 53303
+	AuditIDSyncGatewayProfiling         AuditID = 53304
+
+	// SG cluster events
+	AuditIDClusterInfoRead AuditID = 53350
+
+	// Database events
+	AuditIDCreateDatabase  AuditID = 54000
+	AuditIDReadDatabase    AuditID = 54001
+	AuditIDDeleteDatabase  AuditID = 54002
+	AuditIDDatabaseAllRead AuditID = 54003
+	// Database config events
+	AuditIDReadDatabaseConfig   AuditID = 54010
+	AuditIDUpdateDatabaseConfig AuditID = 54011
+	// Database operation events
+	AuditIDDatabaseOffline       AuditID = 54020
+	AuditIDDatabaseOnline        AuditID = 54021
+	AuditIDDatabaseCompactStatus AuditID = 54030
+	AuditIDDatabaseCompactStart  AuditID = 54031
+	AuditIDDatabaseCompactStop   AuditID = 54032
+	AuditIDDatabaseResyncStatus  AuditID = 54040
+	AuditIDDatabaseResyncStart   AuditID = 54041
+	AuditIDDatabaseResyncStop    AuditID = 54042
+	AuditIDDatabasePostUpgrade   AuditID = 54043
+	AuditIDDatabaseRepair        AuditID = 54044
+	AuditIDDatabaseFlush         AuditID = 54045
+
+	// User principal events
+	AuditIDUserCreate AuditID = 54100
+	AuditIDUserRead   AuditID = 54101
+	AuditIDUserUpdate AuditID = 54102
+	AuditIDUserDelete AuditID = 54103
+	// Role principal events
+	AuditIDRoleCreate AuditID = 54110
+	AuditIDRoleRead   AuditID = 54111
+	AuditIDRoleUpdate AuditID = 54112
+	AuditIDRoleDelete AuditID = 54113
+
+	// Changes feeds events
+	AuditIDChangesFeedStarted AuditID = 54200
+
+	// BLIP Replication events
+	AuditIDReplicationConnect    AuditID = 54300
+	AuditIDReplicationDisconnect AuditID = 54301
+
+	// ISGR events
+	AuditIDISGRCreate    AuditID = 54400
+	AuditIDISGRRead      AuditID = 54401
+	AuditIDISGRUpdate    AuditID = 54402
+	AuditIDISGRDelete    AuditID = 54403
+	AuditIDISGRStatus    AuditID = 54410
+	AuditIDISGRStart     AuditID = 54411
+	AuditIDISGRStop      AuditID = 54412
+	AuditIDISGRReset     AuditID = 54413
+	AuditIDISGRAllStatus AuditID = 54420
+	AuditIDISGRAllRead   AuditID = 54421
+
+	// Documents events
+	AuditIDDocumentCreate       AuditID = 55000
+	AuditIDDocumentRead         AuditID = 55001
+	AuditIDDocumentUpdate       AuditID = 55002
+	AuditIDDocumentMetadataRead AuditID = 55004
+	// Document attachments events
+	AuditIDDocumentDelete   AuditID = 55003
+	AuditIDAttachmentCreate AuditID = 55010
+	AuditIDAttachmentRead   AuditID = 55011
+	AuditIDAttachmentUpdate AuditID = 55012
+	AuditIDAttachmentDelete AuditID = 55013
 )
 
 // AuditEvents is a table of audit events created by Sync Gateway.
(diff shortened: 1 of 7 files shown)


base/audit_events.go
```diff
@@ -18,13 +18,105 @@ const (
...
+	AuditIDISGRAllStatus AuditID = 54420
+	AuditIDISGRAllRead   AuditID = 54421
+
+	// Documents events
+	AuditIDDocumentCreate       AuditID = 55000
+	AuditIDDocumentRead         AuditID = 55001
+	AuditIDDocumentUpdate       AuditID = 55002
+	AuditIDDocumentMetadataRead AuditID = 55004
+	// Document attachments events
+	AuditIDDocumentDelete   AuditID = 55003
```
torcolvin: switch order for numeric consistency and comment association

```suggestion
	AuditIDDocumentDelete   AuditID = 55003
	AuditIDDocumentMetadataRead AuditID = 55004
	
	// Document attachments events
```

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		EnabledByDefault:   true,
+		FilteringPermitted: false,
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDDatabaseResyncStart: {
+		Name:        "Database resync start",
+		Description: "Database resync was started",
+		MandatoryFields: AuditFields{
+			"db":                   "database name",
+			"collections":          []string{"list", "of", "collections"},
```
torcolvin: I'd probably list these as `map[string][]string{"scopeName": []string{"collectionA", "collectionB"}}` like they are in the request object. If this isn't specified, should we figure out what this is and write it here, or leave this blank?

I am leaning toward filling this value in.

base/audit_events.go
```diff
@@ -34,23 +126,99 @@ const (
...
+		EventType:          eventTypeAdmin,
+	},
 	AuditIDPublicUserAuthenticated: {
-		Name:        "User authenticated",
-		Description: "User successfully authenticated",
+		Name:        "Public API user authenticated",
+		Description: "Public API user successfully authenticated",
 		MandatoryFields: AuditFields{
-			"method": "auth_method", // e.g. "basic", "oidc", "cookie", ...
+			"method": "basic, oidc, cookie, etc.",
```
torcolvin: I'd be tempted to differentiate the keys `method`. In AuditIDPublicHTTPAPIRequest `method` means GET/POST/PUT but here is means the auth type.

I'd be tempted to not use `method` at all and always use `http_method` or `auth_method`.

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
 		FilteringPermitted: false,
 		EventType:          eventTypeUser,
 	},
+	AuditIDPublicUserSessionCreated: {
+		Name:        "Public API user session created",
+		Description: "Public API user session was created",
+		MandatoryFields: AuditFields{
+			"session_id": "session_id",
```
torcolvin, line 249: I'd be tempted to include the method of auth here, password / oidc, etc.

The create session APIs are on db endpoints, so does it make sense to include db in them?

Do we want to know if the request was initiated by the same user or by an admin user acting on behalf of someone else?

For the public /db/_session it looks like you can create a user without auth_method but looking at credentials from the body only.

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		},
+		EnabledByDefault:   true,
+		FilteringPermitted: true,
+		EventType:          eventTypeUser,
+	},
+	AuditIDPublicUserSessionDeleted: {
+		Name:        "Public API user session deleted",
+		Description: "Public API user session was deleted",
+		MandatoryFields: AuditFields{
```
torcolvin, line 258: From above, this is db scoped. This could be done by admin user or non admin user, on different endpoints:

DELETE /db/_session/sessionid and DELETE /db/_user/name/_session/_sessionid

I think there's a missing event for dropping all user sessions (admin only)

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+			"session_id": "session_id",
+		},
+		EnabledByDefault:   true,
+		FilteringPermitted: true,
+		EventType:          eventTypeUser,
+	},
+	AuditIDAdminUserAuthenticated: {
```
torcolvin, line 265: for the authentication, I think it might be useful to state what type of authentication is used if we have that? Should we include guest user true/false?

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		EnabledByDefault:   true,
+		FilteringPermitted: false,
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDSyncGatewayCollectInfoStatus: {
+		Name:               "sgcollect_info status",
+		Description:        "sgcollect_info status was viewed",
+		EnabledByDefault:   true,
+		FilteringPermitted: false,
+		MandatoryFields:    AuditFields{},
```
torcolvin, line 298: We don't include this message in the status output, but I wonder if we actually want to include the same fields as sgcollect start as optional fields so we know which sgcollect it was gathering information about?

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+			"output_dir":   "output_directory",
+			"upload_host":  "upload_host",
+			"customer":     "customer",
+			"ticket":       "ticket",
+			"keep_zip":     true,
+			"zip_filename": "zip_filename",
+		},
+		EventType: eventTypeAdmin,
+	},
+	AuditIDSyncGatewayCollectInfoStop: {
```
torcolvin, line 316: Is this an event for just STOP on the http endpoint? or when sgcollect finishes?

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		Name:               "sgcollect_info stop",
+		Description:        "sgcollect_info was stopped",
+		EnabledByDefault:   true,
+		FilteringPermitted: false,
+		MandatoryFields:    AuditFields{},
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDSyncGatewayStats: {
```
torcolvin, line 324: should this include whether this is expvar or prometheus?

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		Description:        "stats were requested",
+		EnabledByDefault:   true,
+		FilteringPermitted: false,
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDSyncGatewayProfiling: {
+		Name:        "profiling requested",
+		Description: "profiling was requested",
+		MandatoryFields: AuditFields{
+			"profile_type": "cpu, memory, etc.",
```
torcolvin, line 338: we can specify an optional filename here

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		},
+		EnabledByDefault:   true,
+		FilteringPermitted: false,
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDClusterInfoRead: {
+		Name:               "Sync Gateway cluster info read",
+		Description:        "Sync Gateway cluster info was viewed",
+		EnabledByDefault:   true,
```
torcolvin, line 350: should this include a field whether it is reading from server, or from legacy config (local only)?

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDRoleUpdate: {
+		Name:        "Update role",
+		Description: "Role was updated",
+
+		MandatoryFields: AuditFields{
+			"role":           "role_name",
+			"db":             "database name",
+			"admin_channels": []string{"list", "of", "channels"},
```
torcolvin, line 600: How do we differentiate between admin channels per collection?

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		FilteringPermitted: false,
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDRoleCreate: {
+		Name:        "Create role",
+		Description: "A new role was created",
+		MandatoryFields: AuditFields{
+			"role":           "role_name",
+			"db":             "database name",
+			"admin_channels": []string{"list", "of", "channels"},
```
torcolvin, line 575: These are collection scoped

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		FilteringPermitted: false,
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDCreateDatabase: {
+		Name:        "Create database",
+		Description: "A new database was created",
+		MandatoryFields: AuditFields{
+			"db":     "database name",
+			"config": "JSON representation of db config",
+		},
```
torcolvin, line 360: Should we log whether this is persistent config or not?

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		Name:        "Database offline",
+		Description: "Database was taken offline",
+		MandatoryFields: AuditFields{
+			"db": "database name",
+		},
+		EnabledByDefault:   true,
+		FilteringPermitted: false,
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDDatabaseOnline: {
```
torcolvin, line 417: For this event, do we want to log when asyncOnline completes? This is out of scope for this PR either way.

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		},
+		EnabledByDefault:   true,
+		FilteringPermitted: false,
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDDatabaseCompactStart: {
+		Name:        "Database compaction start",
+		Description: "Database compaction was started",
+		MandatoryFields: AuditFields{
+			"db": "database name",
```
torcolvin, line 441: This should say tombstone or compaction and dryrun parameters.

We can also reset a compaction, maybe this should be a separate event?

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		MandatoryFields: AuditFields{
+			"db":                   "database name",
+			"collections":          []string{"list", "of", "collections"},
+			"regenerate_sequences": true,
+		},
+		EnabledByDefault:   true,
+		FilteringPermitted: false,
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDDatabaseResyncStop: {
```
torcolvin, line 480: There should be a resync reset, which drops the checkpoints. This is probably just a field that goes with start that says reset.

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		},
+		EnabledByDefault:   true,
+		FilteringPermitted: false,
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDDatabasePostUpgrade: {
+		Name:        "Database post-upgrade",
+		Description: "Database post-upgrade was run",
+		MandatoryFields: AuditFields{
+			"db": "database name",
```
torcolvin: This has preview (dryrun) parameter.

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDUserCreate: {
+		Name:        "Create user",
+		Description: "A new user was created",
+		MandatoryFields: AuditFields{
+			"username": "username",
+			"db":       "database name",
+			"roles":    []string{"list", "of", "roles"},
+			"channels": []string{"list", "of", "channels"},
```
torcolvin, line 528: These are collection scoped.

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		},
+		EnabledByDefault:   true,
+		FilteringPermitted: false,
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDUserRead: {
+		Name:        "Read user",
+		Description: "Information about this user was viewed",
+		MandatoryFields: AuditFields{
+			"username": "username",
```
torcolvin, line 538: Is this going to be real_userid instead?

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDUserUpdate: {
+		Name:        "Update user",
+		Description: "User was updated",
+		MandatoryFields: AuditFields{
+			"username": "username",
+			"db":       "database name",
+			"roles":    []string{"list", "of", "roles"},
+			"channels": []string{"list", "of", "channels"},
```
torcolvin, line 552: these are collection scoped

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		},
+		EnabledByDefault:   true,
+		FilteringPermitted: false,
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDUserRead: {
```
torcolvin, line 534: We can read all users, should this be a separate event?

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		MandatoryFields: AuditFields{
+			"role":           "role_name",
+			"db":             "database name",
+			"admin_channels": []string{"list", "of", "channels"},
+		},
+		EnabledByDefault:   true,
+		FilteringPermitted: false,
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDRoleRead: {
```
torcolvin, line 581: We can read all users, should this be a separate event?

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+	},
+	AuditIDChangesFeedStarted: {
+		Name:        "Changes feed started",
+		Description: "Changes feed was started",
+		MandatoryFields: AuditFields{
+			"db":    "database name",
+			"ks":    "keyspace",
+			"since": "since",
+		},
+		OptionalFields: AuditFields{
```
torcolvin, line 625: This is missing parameters:

- revocations
- active_only
- include_docs
- limit
- request_plus

See https://github.com/couchbase/sync_gateway/blob/430234fbee5bd06259ac2c920c7602f707a65be1/rest/changes_api.go#L576

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+			"db":   "database name",
+		},
+		EnabledByDefault:   true,
+		FilteringPermitted: false,
+		EventType:          eventTypeAdmin,
+	},
+	AuditIDChangesFeedStarted: {
+		Name:        "Changes feed started",
+		Description: "Changes feed was started",
+		MandatoryFields: AuditFields{
```
torcolvin, line 620: Should we include whether this is over the admin port or not?

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+			"channels":  []string{"list", "of", "channels"},
+			"feed_type": "continuous, normal, longpoll, websocket, etc.",
+		},
+		EnabledByDefault:   true,
+		FilteringPermitted: true,
+		EventType:          eventTypeData,
+	},
+	AuditIDReplicationConnect: {
+		Name:        "Replication connect",
+		Description: "A client connected using WebSocket/BLIP for replication",
```
torcolvin: I was tempted to use a friendly name of Couchbase Lite/Inter-Sync Gateway here.

base/audit_events.go
```diff
@@ -72,6 +240,638 @@ var AuditEvents = events{
...
+		},
+		EnabledByDefault:   false,
+		FilteringPermitted: true,
+	},
+	AuditIDISGRStart: {
+		Name:        "Inter-Sync Gateway Replication start",
+		Description: "Inter-Sync Gateway Replication was started",
+		MandatoryFields: AuditFields{
+			"db":             "database name",
+			"replication_id": "replication id",
```
torcolvin, line 721: This will have a `cid` from the http request and also different `cid` from the replication itself.

torcolvin (commented):
I've skipped looking through the crud events, there is a lot of subtlety there, but there's enough stuff to work through in this PR already. I expect a lot of my comments can be easily addressed or dismissed.

As discussed offline, making fields standard for `real_userid` would be helpful to validate we haven't dropped this information by mistake. `cid` is a candidate for this. I think implementing these fields as "opt-out" in the config would be more clear, but this will involve refactoring the validation of `MandatoryFields` to assemble these fields.

base/audit_events.go
torcolvin: We can reset only an attachment compaction, this means we'll start from fresh checkpoints.

Tombstone compaction is necessarily query based it always starts from scratch after it has been stopped.

We can do a dry run for attachment compaction? I wonder if we should have separate events for tombstone and attachment compaction and make more fields mandatory?

base/audit_events.go
bbrks: It is the user-initiated action, that is what we are auditing. The `DELETE /_sgcollect_info` request

base/audit_events.go
bbrks: both, will add field to determine which flavour of stats were requested

base/audit_events.go
bbrks: There aren't any parameters available on the status request. It just returns whether `sgcollect_info` is running or not, and we don't have more information in this request scope to populate anything else.

base/audit_events.go
bbrks: Not sure it matters, think we can treat both equally.

Legacy doesn't return any info but if we treat it like it did then there's no advantage in knowing.

base/audit_events.go
bbrks: Don't think it matters. Either way a database was created and the config of that database is available to see.

base/audit_events.go
bbrks: Deferring to implementation.

base/audit_events.go
bbrks: I agree, separate events for attachment compaction and database compaction.

base/audit_events.go
bbrks: That's an option on start, I'll add it as a bool field like we do for `regenerate_sequences`.

base/audit_events.go
bbrks: `real_userid` would be the admin user who made the request, but that request might be viewing a different user (by username)

base/audit_events.go
bbrks: `/db/_users` only returns names, but yes I can treat that as a new event, and list which usernames were in that response

base/audit_events.go
bbrks: made this more general and can be more specific with the optional client_type and client_version fields.

base/audit_events.go
bbrks: Given the replication isn't necessarily going to be running on the node that the started request came in on, we cannot determine blip context id.

This is another one of those cases where we want to audit the action the user successfully performed and not the eventual outcome of the action, if that makes sense?

torcolvin (approved):

PR #6953, opened 2024-07-10
CBG-4050 add auditing for some database events

torcolvin:
NOTE: `EnabledByDefault` and `FilteringPermitted` are not up to date, they are going to be hit in a separate pass by a PM.

Proposals not in the PR:

- rename "reset" parameter for resync/attachment compaction to something else? It's tricky because it is true (wipe checkpoints) or false (resume checkpoints if applicable). I wonder if this should be renamed.
- should we validate that fields that are not injected by globally by the merge function are defined in optional fields?
- if we have a type that is a nested map type like `collections: map[string][]string`, should we validate the correct type? I prefer this syntax since it is scalable to multiple scopes, but I'm not set on this type.

I kept audit events at the end of the functions in the REST apis.

When doing the PR:

- Are there parameters that are missing?
- Are the descriptions of the event parameters correct?
- Are there any fields that should be mandatory (validated by `-tags cb_sg_devmode` which will cause panic)?
- Are there optional fields that should be documented?
- Did I add all the events that are present?

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2561/

### DIFF
--- a/base/audit_events.go
+++ b/base/audit_events.go
@@ -65,14 +65,11 @@ const (
 	AuditIDReadDatabaseConfig   AuditID = 54010
 	AuditIDUpdateDatabaseConfig AuditID = 54011
 	// Database operation events
-	AuditIDDatabaseOffline                 AuditID = 54020
-	AuditIDDatabaseOnline                  AuditID = 54021
-	AuditIDDatabaseAttachmentCompactStatus AuditID = 54030
-	AuditIDDatabaseAttachmentCompactStart  AuditID = 54031
-	AuditIDDatabaseAttachmentCompactStop   AuditID = 54032
-	AuditIDDatabaseTombstoneCompactStatus  AuditID = 54033
-	AuditIDDatabaseTombstoneCompactStart   AuditID = 54034
-	AuditIDDatabaseTombstoneCompactStop    AuditID = 54035
+	AuditIDDatabaseOffline       AuditID = 54020
+	AuditIDDatabaseOnline        AuditID = 54021
+	AuditIDDatabaseCompactStatus AuditID = 54030
+	AuditIDDatabaseCompactStart  AuditID = 54031
+	AuditIDDatabaseCompactStop   AuditID = 54032
 
 	AuditIDDatabaseResyncStatus AuditID = 54040
 	AuditIDDatabaseResyncStart  AuditID = 54041
@@ -411,8 +408,10 @@ var AuditEvents = events{
 	AuditIDDatabaseOffline: {
 		Name:        "Database offline",
 		Description: "Database was taken offline",
-		MandatoryFields: AuditFields{
-			"db": "database name",
+		mandatoryFieldGroups: []fieldGroup{
+			fieldGroupDatabase,
+			fieldGroupRequest,
+			//fieldGroupAuthenticated, // FIXME: CBG-3973
 		},
 		EnabledByDefault:   true,
 		FilteringPermitted: false,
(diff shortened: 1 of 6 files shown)


base/audit_events.go
```diff
@@ -64,17 +65,20 @@ const (
...
-	AuditIDDatabaseRepair        AuditID = 54044
-	AuditIDDatabaseFlush         AuditID = 54045
+	AuditIDDatabaseOffline                 AuditID = 54020
+	AuditIDDatabaseOnline                  AuditID = 54021
+	AuditIDDatabaseAttachmentCompactStatus AuditID = 54030
+	AuditIDDatabaseAttachmentCompactStart  AuditID = 54031
+	AuditIDDatabaseAttachmentCompactStop   AuditID = 54032
+	AuditIDDatabaseTombstoneCompactStatus  AuditID = 54033
+	AuditIDDatabaseTombstoneCompactStart   AuditID = 54034
+	AuditIDDatabaseTombstoneCompactStop    AuditID = 54035
```
adamcfraser: It doesn't feel to me like we need this granularity between Compact events (i.e. it's reasonable to say users will either want to disable or enable all or none for compact event auditing).

Given that, I think a better approach would be a 'type' Mandatory field for the compact events, and make the type-specific parameters (reset, dry_run, etc) non-mandatory fields.

rest/admin_api.go
```diff
@@ -237,19 +237,19 @@ func (h *handler) handleDbOnline() error {
...
 
 	return nil
 }
 
 // Take a DB offline
 func (h *handler) handleDbOffline() error {
 	h.assertAdminOnly()
-	var err error
-	if err = h.db.TakeDbOffline(base.NewNonCancelCtx(), "ADMIN Request"); err != nil {
+	if err := h.db.TakeDbOffline(base.NewNonCancelCtx(), "ADMIN Request"); err != nil {
```
adamcfraser, line 248: It looks like you're missing `return err` inside this if.

rest/api.go
```diff
@@ -183,6 +190,10 @@ func (h *handler) handleCompact() error {
 				return err
 			}
 			h.writeRawJSON(status)
+			base.Audit(h.ctx(), base.AuditIDDatabaseAttachmentCompactStart, base.AuditFields{
+				"dry_run": h.getBoolQuery("dry_run"),
```
adamcfraser: It feels like the names of Mandatory (and non-Mandatory) audit fields should be defined in `base/audit_events` as constants to avoid typos (and to make it easy to find usage).  Has that already been considered?

rest/api.go
```diff
@@ -260,11 +273,13 @@ func (h *handler) handleFlush() error {
 		h.server.RemoveDatabase(h.ctx(), name)
 		err := bucket.CloseAndDelete(h.ctx())
 		_, err2 := h.server.AddDatabaseFromConfig(h.ctx(), config.DatabaseConfig)
-		if err == nil {
-			err = err2
+		if err2 != nil {
```
adamcfraser: This looks like it's reversed the previous handling (where err had priority over err2 as the return value).  Is that intentional?

rest/bulk_api.go
```diff
@@ -262,7 +262,7 @@ func (h *handler) handleDump() error {
 
 // HTTP handler for _repair
 func (h *handler) handleRepair() error {
-
+	base.Audit(h.ctx(), base.AuditIDDatabaseRepair, nil)
```
adamcfraser: I don't believe this should be emitting audit logs (since this endpoint is unused and always returns error).  If the goal is to have it in place so it's not missed if the endpoint gets enabled in future, maybe switch to something like:
```
err := errors.New("_repair endpoint disabled")
if err != nil {
  base.Audit(h.ctx(), base.AuditIDDatabaseRepair, nil)
}
return err
```

base/audit_events.go
```diff
@@ -487,12 +523,11 @@ var AuditEvents = events{
...
-	AuditIDDatabasePostUpgrade: {
-		Name:        "Database post-upgrade",
-		Description: "Database post-upgrade was run",
+	AuditIDPostUpgrade: {
+		Name:        "Post-upgrade",
+		Description: "Post-upgrade was run for Sync Gateway databases",
 		MandatoryFields: AuditFields{
-			"db":      "database name",
-			"preview": true,
+			"preview": false,
```
adamcfraser: What does the value in the MandatoryFields map represent?  Default value?

rest/server_context.go
```diff
@@ -408,6 +408,7 @@ type PostUpgradeResult map[string]PostUpgradeDatabaseResult
 type PostUpgradeDatabaseResult struct {
 	RemovedDDocs   []string `json:"removed_design_docs"`
 	RemovedIndexes []string `json:"removed_indexes"`
+	err            error    // this is not exported, to be used for additional information
```
adamcfraser: I don't see this being used anywhere - is it still required?  I don't think the audit event needs to include whether removal of design docs or indexes occurred.

adamcfraser (commented):
Left a few comments/questions. 

For the comments/questions in the PR:
- Renaming API parameters would require a major release, I don't think the reset on compaction is so far off that we need to worry about it
- I like the idea of OptionalFields or NonMandatoryFields in the event definitions to prevent collisions
- I'm not sure how the `collections: map[string][]string` question applies to this PR

base/audit_events.go
torcolvin: I waffled on this, because it's easy to drop fields if they aren't mandatory and this was the way it was written originally. I don't think any of these parameters are super important though.

The sticking point to me was that tombstone is necessarily query based and doesn't support checkpoints but attachment compaction does support resume.

Do you have feelings on `type` as a generic field, or my original idea of `compaction_type`? I worry about what to name this constant and the likelihood we might accidentally use `type` to mean something else.

We had a near mishap with overlapping field names before: `method` referred to `auth_method` (basic/oidc) and `http_method` (GET/POST/PUT). The implementation as is will take fields from `Audit(ctx, id, fields)` and have the context fields overwrite the passed in fields. I do not think there's ever a reason we'd implement this.

That said, I am not sure it's a big deal to have `type` mean different things in different contexts.

An alternate way of handling this would be to log query string for this event type, like we do for the payload.

base/audit_events.go
torcolvin: This doesn't represent anything in fact, the type of the field is used "bool" to write an example field, and it will always show `true` in the example docs.

rest/server_context.go
torcolvin: I made this event more complicated than it needed to be in a previous review, I moved post upgrade from being db specific.

rest/api.go
torcolvin: nope.

base/audit_events.go
torcolvin: Decided to go with separating fields from offline conversation.

adamcfraser (approved):

PR #6954, opened 2024-07-10
CBG-2206: TestReplicatorDoNotSendDeltaWhenSrcIsTombstone flake

gregns1:
CBG-2206

- Test was asserting on 404 missing on fetching the specific rev via rest api. This is incorrect as you will still get 200 code but with _deleted body. Changed test to reflect this + had a check that fetching the doc (and not specifying rev id returned 404) 
- In addition I noticed that on destination when the delete delta was being replicated over, we were patching the empty delta to the revision on the destination. Meaning we would have _deleted in the body but also the old body fields too. The example from the test looked like so:
```
{"_deleted":true,"_id":"test","_rev":"2-26ddda6a2110ff1a408ebb8d36674b17","field1":"f1_1","field2":"f2_1"}
```
- To fix I added statement to only patch if the incoming delta is not a delete. 

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2562/

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -1013,7 +1013,7 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 	newDoc.UpdateBodyBytes(bodyBytes)
 
 	injectedAttachmentsForDelta := false
-	if deltaSrcRevID, isDelta := revMessage.DeltaSrc(); isDelta {
+	if deltaSrcRevID, isDelta := revMessage.DeltaSrc(); isDelta && !revMessage.Deleted() {
 		if !bh.sgCanUseDeltas {
 			return base.HTTPErrorf(http.StatusBadRequest, "Deltas are disabled for this peer")
 		}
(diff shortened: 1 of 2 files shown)


db/blip_handler.go
```diff
@@ -1049,17 +1049,19 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
...
 
-		deltaSrcMap := map[string]interface{}(deltaSrcBody)
-		err = base.Patch(&deltaSrcMap, newDoc.Body(bh.loggingCtx))
-		// err should only ever be a FleeceDeltaError here - but to be defensive, handle other errors too (e.g. somehow reaching this code in a CE build)
-		if err != nil {
-			// Something went wrong in the diffing library. We want to know about this!
-			base.WarnfCtx(bh.loggingCtx, "Error patching deltaSrc %s with %s for doc %s with delta - err: %v", deltaSrcRevID, revID, base.UD(docID), err)
-			return base.HTTPErrorf(http.StatusUnprocessableEntity, "Error patching deltaSrc with delta: %s", err)
-		}
+		if !revMessage.Deleted() {
```
adamcfraser: We don't intend to support deltas for deletes (i.e. if `deleted` and `deltaSrc` properties are both specified on an incoming rev message, we should ignore deltaSrc, and perform this check up on line 1016).

adamcfraser (approved):

PR #6956, opened 2024-07-11
CBG-4051 audit database creation/read/update/delete

torcolvin:
Reconstruct fake dbconfig as a payload for updating sync function and import filters.

## Pre-review checklist

### DIFF
--- a/base/audit_events.go
+++ b/base/audit_events.go
@@ -357,10 +357,14 @@ var AuditEvents = events{
 	},
 	AuditIDCreateDatabase: {
 		Name:        "Create database",
-		Description: "A new database was created",
+		Description: "A new database was created from REST API",
 		MandatoryFields: AuditFields{
-			"db":              "database name",
 			AuditFieldPayload: "JSON representation of db config",
+		},
+		mandatoryFieldGroups: []fieldGroup{
+			fieldGroupDatabase,
+			fieldGroupRequest,
+			// fieldGroupAuthenticated, // FIXME: CBG-3973
 		},
 		EnabledByDefault:   true,
 		FilteringPermitted: false,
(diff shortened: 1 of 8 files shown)


base/audit_events.go
```diff
@@ -357,53 +357,78 @@ var AuditEvents = events{
 	},
 	AuditIDCreateDatabase: {
 		Name:        "Create database",
-		Description: "A new database was created",
+		Description: "A new database was created from REST API",
```
adamcfraser: What's the non-"from REST API" case you feel is worth calling out here and below?  If it's only legacy bootstrap config I don't think that merits the extra text.

rest/admin_api.go
```diff
@@ -162,6 +162,12 @@ func (h *handler) handleCreateDB() error {
 		}
 	}
 
+	base.Audit(h.ctx(), base.AuditIDCreateDatabase,
+		base.AuditFields{
+			base.AuditFieldDatabase: dbName,
+			base.AuditFieldPayload:  rawBytes,
```
adamcfraser: Should we be performing redaction on the payload via `DbConfig.Redacted()` before logging it, or is that done elsewhere in the audit logging framework?

rest/admin_api.go
```diff
@@ -961,8 +981,10 @@ func (h *handler) handlePutCollectionConfigSync() error {
 			config := bucketDbConfig.Scopes[h.collection.ScopeName].Collections[h.collection.Name]
 			config.SyncFn = &js
 			bucketDbConfig.Scopes[h.collection.ScopeName].Collections[h.collection.Name] = config
+			updatedConfigStr = fmt.Sprintf(`{"scopes": {"%s": {"collections": { "%s": { "sync": "%s"}}}}}`, h.collection.ScopeName, h.collection.Name, js)
```
adamcfraser, line 997: This seems similar to the case discussed elsewhere - since the keyspace will already be included in the audit logging, I feel like we can omit the complexity of the nested structs for auditing purposes?  (same applies to other instances of this)

adamcfraser (changes requested):
A question about redaction, and suggestion about the nested scope/collection payloads.

rest/admin_api.go
torcolvin: The keyspace isn't typically included the audit logging unless we ask for it, and it would be an optional field for `base.AuditIDUpdateDatabaseConfig` since if you had an update from `/db/_config` you wouldn't have a keyspace.

base/audit_events.go
torcolvin: I am OK dropping these, but I think there's some potential for confusion around creation or deletion in the sense of appearing or disappearing via the bootstrap API.

I think nearly all cases of removing a database via bootstrap are going to happen because another node called DELETE, but we have seen cases where a database is removed because of a bad cluster connection. I doubt this will be confusing.

I was confused here because does "a new database was created" mean we call `NewDatabaseContext` and the database now exists on this node, or does it mean that a REST call occurred?

I think this is a point of confusion that is not likely to occur to non Sync Gateway developers.

rest/admin_api.go
torcolvin: Added a redaction that works on the string because of the possibility that we nil a field and this wouldn't show up in Redacted() and deserialization.

None of this code is performance sensitive because we are loading a new database but I'm thinking we might want a call in other audit events of `IsAuditLoggingEnabled` before performing new operations.

rest/admin_api.go
adamcfraser: Based on that, it sounds like an optional `auditFieldKeyspace` on this event would be the right choice for the keyspaced endpoints that can modify config, as opposed to crafting an AuditFieldPayload property that doesn't match the request payload?  What do you think?

adamcfraser:
Changes for redaction look good, one question/suggestion about using auditFieldKeyspace for consistency for the keyspaced config endpoints.

rest/admin_api.go
torcolvin: I think conceptually from the user this is the action that they are doing as this config upsert and that is more

I guess we could have a freeform field:

"ks": "db.scope.collection"
"sync": nil <-- delete the custom sync function

and then make "ks" and "payload" optional? We can't log a payload for a DELETE operation and DELETE on a sync or import_filter doesn't reset to empty string, it resets to nil.

adamcfraser (approved):

PR #6957, opened 2024-07-11
CBG-4027 Treat on-demand import for GET errors as not found 

adamcfraser:
CBG-4027

Ensures that these are correctly handled as norev messages when encountered during replication, as they always represent a case where the requested revision was not available.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2572/

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -240,11 +240,14 @@ func (c *DatabaseCollection) OnDemandImportForGet(ctx context.Context, docid str
 	var importErr error
 
 	docOut, importErr = importDb.ImportDocRaw(ctx, docid, rawDoc, xattrs, isDelete, cas, nil, ImportOnDemand)
+
 	if importErr == base.ErrImportCancelledFilter {
-		// If the import was cancelled due to filter, treat as not found
-		return nil, base.HTTPErrorf(404, "Not imported")
+		// If the import was cancelled due to filter, treat as 404 not imported
+		return nil, base.HTTPErrorf(http.StatusNotFound, "Not imported")
 	} else if importErr != nil {
-		return nil, importErr
+		// Treat any other failure to perform an on-demand import as not found
+		base.DebugfCtx(ctx, base.KeyImport, "Unable to import doc %q during on demand import for get - will be treated as not found.  Reason: %v", base.UD(docid), err)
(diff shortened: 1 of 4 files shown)


db/crud.go
```diff
@@ -240,11 +240,14 @@ func (c *DatabaseCollection) OnDemandImportForGet(ctx context.Context, docid str
...
+
 	if importErr == base.ErrImportCancelledFilter {
-		// If the import was cancelled due to filter, treat as not found
-		return nil, base.HTTPErrorf(404, "Not imported")
+		// If the import was cancelled due to filter, treat as 404 not imported
+		return nil, base.HTTPErrorf(http.StatusNotFound, "Not imported")
 	} else if importErr != nil {
-		return nil, importErr
+		// Treat any other failure to perform an on-demand import as not found
+		base.DebugfCtx(ctx, base.KeyImport, "Unable to import doc %q during on demand import for get - will be treated as not found.  Reason: %v", base.UD(docid), err)
```
torcolvin: If there's an error that would be 503, like gocb.OverloadError, do we want to skip this doc to not be found again?

Or is the assumption that this would be caught by the traditional import feed and then updated after?

https://github.com/couchbase/sync_gateway/blob/c6b3b3cb1a8fbbfb05ea8a52f032f5969c4e20c7/db/import.go#L142
or 
https://github.com/couchbase/sync_gateway/blob/main/db/crud.go#L1938 should be able to trigger this behavior if there's a leaky datastore that can return 503s?

torcolvin (dismissed):

db/crud.go
```suggestion
		base.DebugfCtx(ctx, base.KeyImport, "Unable to import doc %q during on demand import for get - will be treated as not found.  Reason: %v", base.UD(docid), importErr)
```

bbrks (dismissed):

db/crud.go
adamcfraser: That's correct, if there's a failure to write the doc for an on-demand import, this will fall back to the feed-based import.  In this case clients who had requested the previous rev will get a norev, and will get the current rev when it is imported.

gregns1 (approved):

PR #6961, opened 2024-07-12
bugfix: load default audit events into ctx when explicit events are not defined

bbrks:
- Allows for the `shouldLog` call to handle default sets of events.
- Print numerical audit event ID instead of a UTF-8 codepoint in error message


gregns1 (approved):
LGTM

PR #6940, opened 2024-07-05
CBG-3981: Wire up PUT/POST /db/_config/audit APIs

bbrks:
CBG-3981

- Wire up PUT/POST /db/_config/audit APIs
- Test coverage of API

## [Integration Tests](https://jenkins.sgwdev.com/job/SyncGateway-Integration/build?delay=0sec)
- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2557/

PR #6963, opened 2024-07-15
OpenAPI Specs: Update /db/_config/audit endpoint body format

bbrks:
Update `/db/_config/audit` format to match changes made in #6940 (events nested under `"events"` property, `"enabled"` exposed)

Screenshot 2024-07-15 at 13 56 42

- n/a

### DIFF
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -1878,18 +1878,28 @@ Database-audit:
   title: Database-audit
   description: A map of audit events and whether they are enabled or not.
   type: object
-  additionalProperties:
-    x-additionalPropertiesName: audit_id
-    description: The audit event ID and whether it is enabled or not.
-    type: boolean
+  properties:
+    enabled:
+      type: boolean
+    events:
+      type: object
+      additionalProperties:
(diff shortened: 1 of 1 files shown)


docs/api/components/schemas.yaml
```diff
@@ -1878,18 +1878,28 @@ Database-audit:
...
-    type: boolean
+  properties:
+    enabled:
+      type: boolean
+    events:
+      type: object
+      additionalProperties:
+        x-additionalPropertiesName: audit_id
+        description: The audit event ID and whether it is enabled or not.
+        type: boolean
```
torcolvin, line 1889: Is this worth adding an example for the format? I am somewhat confused by why this is `map[string]any` and then it is actually map[string]bool?

bbrks: The `PUT`/`POST` endpoint accepts both verbose and non-verbose formats (without specifying the `?verbose=true` query param)

This is displayed correctly for both cases in the Redocly UI with a dropdown to select the type of body format.

Screenshot 2024-07-16 at 19 21 29
Screenshot 2024-07-16 at 19 21 32

torcolvin (approved):

PR #6965, opened 2024-07-15
CBG-4046 add scope name to context for audit logging

torcolvin:
audit log marks `ks` as `db.scope.collection`. Therefore, I thought I needed the full keyspace. 

These are the differences of audit logging and regular logging for the REST API. The regular logging remains the same under this PR.

| REST API    | Audit Log                                           | Regular Log |
| ----------- | ---------------------------------- | ------------ |
| `/dbA/doc` | `{"ks": "dbA._default._default" }`  | `db:dbA`     |
| `/dbA._default/doc` | `{"ks": "dbA._default._default" }`  | `db:dbA col:_default`     |
| `/dbA._default._default/doc` | `{"ks": "dbA._default._default" }`  | `db:dbA col:_default`     |

The reason I wanted to have a way to separate an implicit default collection for logging vs audit logging is that I actually found several places that were subtly dropping the collection from a context.

1. In ISGR, the push and pull replicators were dropping the database when making a blip connection in `_connect` methods.
2. In ISGR, creating a push replicator creates a replicator with `newBlipHandler`. Normally the collection gets set on a handler object via `collectionBlipHandler` but the handler functions are not called for push replications.
3. When responding to a changes request, the `blipHandler.sendBatchOfChanges` has the correct `blipHandler.loggingCtx` at the top of the function. However, to actually send the changes, it calls `bh.handleChangesResponse` which is actually `BlipSyncContext`. `BlipSyncContext` has a db scoped log context, since the `BlipSyncContext` responds to multiple collections. Therefore, I changed code below `handleChangesResponse` to use a context that is passed in rather than the one on `BlipSyncContext`.

To avoid some of this peril in the future, though I'm not suggesting it would be better if `BlipSyncContext` wasn't an embedded struct of `blipHandler`. That would probably cause a good number of code changes. Having `blipHandler.loggingCtx` shadow `BlipSyncContext.loggingCtx` combined with the embedded struct causes extra confusion. I do not think this is the right time to fix this.

I found all these issues by actually adding auditing for CRUD events and fixing any panics when the context didn't contain the keyspace information (database + scope + colleciton).

- [x] https://jenkins.sgwdev.com/job/SyncGateway-Integration/2578/

### DIFF
--- a/base/logging_context.go
+++ b/base/logging_context.go
@@ -168,9 +168,10 @@ func BucketNameCtx(parent context.Context, bucketName string) context.Context {
 	return LogContextWith(parent, &newCtx)
 }
 
-// CollectionCtx extends the parent context with a collection name. Used when bucket and scope are implicit (e.g. in the context of a SG database)
-func CollectionLogCtx(parent context.Context, collectionName string) context.Context {
+// CollectionCtx extends the parent context with a collection name.
+func CollectionLogCtx(parent context.Context, scopeName, collectionName string) context.Context {
 	newCtx := getLogCtx(parent)
+	newCtx.Scope = scopeName
 	newCtx.Collection = collectionName
 	return LogContextWith(parent, &newCtx)
 }
--- a/base/logging_context.go
+++ b/base/logging_context.go
@@ -56,6 +56,7 @@ type LogContext struct {
 	//// RemoteAddr is the IP and port of the remote client making the request associated with this log
 	//RemoteAddr string
 
+	// implicitDefaultCollection is set to true when the context represents the default collection, but we want to omit that value from logging to prevent verbosity.
 	implicitDefaultCollection bool
 }
 
--- a/base/logging_context.go
+++ b/base/logging_context.go
@@ -55,6 +55,8 @@ type LogContext struct {
 	//RequestPort string
 	//// RemoteAddr is the IP and port of the remote client making the request associated with this log
 	//RemoteAddr string
+
+	implicitDefaultCollection bool
 }
 
 // DbLogConfig can be used to customise the logging for logs associated with this database.
@@ -83,7 +85,7 @@ func (lc *LogContext) addContext(format string) string {
 
 	if lc.Bucket != "" {
 		if lc.Database != "" {
-			if lc.Collection != "" {
+			if !lc.implicitDefaultCollection && lc.Collection != "" {
 				format = "col:" + lc.Collection + " " + format
 			}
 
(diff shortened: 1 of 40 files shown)


db/changes.go
```diff
@@ -385,6 +385,7 @@ func (db *DatabaseCollectionWithUser) changesFeed(ctx context.Context, singleCha
 
 	feed := make(chan *ChangeEntry, 1)
 
+	fmt.Printf("collection=%+v\n", db.DatabaseCollection)
```
adamcfraser: Dev-time logging cleanup.

base/logging_context.go
```diff
@@ -55,6 +55,8 @@ type LogContext struct {
 	//RequestPort string
 	//// RemoteAddr is the IP and port of the remote client making the request associated with this log
 	//RemoteAddr string
+
+	implicitDefaultCollection bool
```
adamcfraser, line 60: Adding a comment for this property would be nice - it's unclear from name alone what it indicates.

rest/handler.go
```diff
@@ -572,6 +580,11 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 			keyspaceScope = base.StringPtr(base.DefaultScope)
 			keyspaceCollection = base.StringPtr(base.DefaultCollection)
 		}
+		if explicitCollectionLogging {
```
adamcfraser, line 584: I think a comment here as well describing why/when we're choosing explicit/implicit CollectionLogging would be helpful.

db/active_replicator_pull.go
```diff
@@ -45,8 +45,7 @@ func (apr *ActivePullReplicator) Start(ctx context.Context) error {
 	}
 
 	apr.setState(ReplicationStateStarting)
-	// intentionally reset the context from having db information on it?
-	logCtx := base.LogContextWith(ctx, &base.LogContext{CorrelationID: apr.config.ID + "-" + string(ActiveReplicatorTypePull)})
+	logCtx := base.CorrelationIDLogCtx(ctx, apr.config.ID+"-"+string(ActiveReplicatorTypePull))
```
adamcfraser, line 48: Did you sanity check that this wasn't intentional to do something like removing cancellation context?

adamcfraser (changes requested):
Changes look good to me - just a few spots where comments would be helpful, and a sanity check on the active replicator context change.

db/active_replicator_pull.go
torcolvin: `LogContextWith` keeps the context and just resets the entirety of `base.LogContext`.
To drop a cancelation, you have to do `base.NonCancellableContext` or go added in 1.21 `context.Context.WithoutCancel`

adamcfraser (approved):

PR #6967, opened 2024-07-16
CBG-4071 Avoid duplicate persisted events in db config

adamcfraser:
CBG-4071


bbrks (approved):
Gah, thanks for catching.

PR #6969, opened 2024-07-16
CBG-4049 Implement audit event set

adamcfraser:
Includes implementation for following event categories:
- audit config changes
- startup
- REST API requests
- Node-level events

CBG-4049

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2592/


base/logger_audit.go
```diff
@@ -142,6 +122,11 @@ func Audit(ctx context.Context, id AuditID, additionalData AuditFields) {
 	auditLogger.logf(string(fieldsJSON))
 }
 
+// IsAuditEnabled checks if auditing is enabled for the SG node
+func IsAuditEnabled() bool {
+	return auditLogger.FileLogger.shouldLog(LevelNone)
```
torcolvin, line 142: Do we need to protect around `auditLogger == nil` when accessing an embedded struct?

base/logger_audit.go
```diff
@@ -184,6 +169,10 @@ func NewAuditLogger(ctx context.Context, config *AuditLoggerConfig, logFilePath
 		globalFields: globalFields,
 	}
 
+	if *config.FileLoggerConfig.Enabled {
+		Audit(ctx, AuditIDAuditEnabled, AuditFields{"audit_scope": "global"})
```
torcolvin, line 188: Do you want to add constants for AuditFieldAuditScope and AuditFieldScopeDb or AuditFieldScopeGlobal?

base/logging_config.go
```diff
@@ -223,10 +223,17 @@ func EnableStatsLogger(enabled bool) {
...
-func EnableAuditLogger(enabled bool) {
+func EnableAuditLogger(ctx context.Context, enabled bool) {
 	if auditLogger != nil {
+		if !enabled {
+			Audit(ctx, AuditIDAuditDisabled, AuditFields{"audit_scope": "global"})
+		}
 		auditLogger.Enabled.Set(enabled)
+		if enabled {
+			Audit(ctx, AuditIDAuditEnabled, AuditFields{"audit_scope": "global"})
+		}
```
torcolvin: take or leave:

`auditLogger.Enabled` can never panic but the construction made me wonder about doing the if/else next to each other.

```
		auditLogger.Enabled.Set(enabled)
		if enabled {
			Audit(ctx, AuditIDAuditEnabled, AuditFields{"audit_scope": "global"})
		} else {
			Audit(ctx, AuditIDAuditDisabled, AuditFields{"audit_scope": "global"})
		}
```

rest/admin_api.go
```diff
@@ -840,6 +841,13 @@ func (h *handler) handlePutDbAuditConfig() error {
 		}
 		return nil
 	})
+	if err != nil {
+		base.Audit(h.ctx(), base.AuditIDAuditConfigChanged, base.AuditFields{
+			"audit_scope":          "db",
+			base.AuditFieldPayload: body,
+		})
+	}
+	return err
```
torcolvin: This construction is backward:

```suggestion
	if err != nil {
		return err
	}
	base.Audit(h.ctx(), base.AuditIDAuditConfigChanged, base.AuditFields{
		"audit_scope":          "db",
		base.AuditFieldPayload: body,
	})
	return nil
```

rest/api.go
```diff
@@ -520,6 +520,7 @@ func (h *handler) handleProfiling() error {
 			base.InfofCtx(h.ctx(), base.KeyAll, "... ending CPU profile")
 			pprof.StopCPUProfile()
 			h.server.CloseCpuPprofFile(h.ctx())
+			base.Audit(h.ctx(), base.AuditIDSyncGatewayProfiling, base.AuditFields{"profile_type": "cpu"})
```
torcolvin: I think it is probably worth knowing which filename this is for this audit event. I think it is tempting to make filename mandatory but some of the profiles are displayed to the browser.

I understand the idea of writing this as just "profile_type" : "cpu" because this represents when the profile is collected, to match the other profile types. But CPU profiling is collected over time, so I wonder if it makes sense to have an `"action": "start"`, `"action": "write"` or `"action": "stop"` - basically so we know the difference between:

```
base.Audit(h.ctx(), base.AuditIDSyncGatewayProfiling, base.AuditFields{"profile_type": "cpu (start)", "filename": params.File})
```

and this call.

rest/api.go
```diff
@@ -583,32 +587,38 @@ func (h *handler) handleHeapProfiling() error {
 
 func (h *handler) handlePprofGoroutine() error {
 	httpprof.Handler("goroutine").ServeHTTP(h.response, h.rq)
+	base.Audit(h.ctx(), base.AuditIDSyncGatewayProfiling, base.AuditFields{"profile_type": "goroutine"})
```
torcolvin: For the profiles we send to `net/http/pprof` directly, should we log query parameters? You can do a performance based attack via calling the endpoints with long sampling times. But this is not a compliance issue.

rest/api.go
```diff
@@ -583,32 +587,38 @@ func (h *handler) handleHeapProfiling() error {
...
 
 func (h *handler) handlePprofHeap() error {
 	httpprof.Handler("heap").ServeHTTP(h.response, h.rq)
+	base.Audit(h.ctx(), base.AuditIDSyncGatewayProfiling, base.AuditFields{"profile_type": "heap"})
 	return nil
 }
 
 func (h *handler) handlePprofProfile() error {
 	httpprof.Profile(h.response, h.rq)
+	base.Audit(h.ctx(), base.AuditIDSyncGatewayProfiling, base.AuditFields{"profile_type": "cpu"})
```
torcolvin: I'd be tempted to put `profile_type` and profile type constants the same so we can match our endpoints that use `net/http/pprof` directly like this and the ones we are redirecting to a file.

rest/api.go
```diff
@@ -679,13 +694,14 @@ func sleep(rq *http.Request, d time.Duration) {
...
 	h.writeJSON(st)
+
+	base.Audit(h.ctx(), base.AuditIDSyncGatewayStats, base.AuditFields{"stats_format": "memstats"})
 	return nil
 }
 
 func (h *handler) handleMetrics() error {
 	promhttp.Handler().ServeHTTP(h.response, h.rq)
-
+	base.Audit(h.ctx(), base.AuditIDSyncGatewayStats, base.AuditFields{"stats_format": "prometheus"})
```
torcolvin: These work over the silent router where we are typically not logging other requests, do we want to not have these log?

This may be contradictory to other things I've said. OTOH, having the ends and they are always off.

If they are over the silent router, we should check if they will should authentication information.

This field should have `mandatoryFieldGroups: []fieldGroup{fieldGroupAuthenticated}` I believe.

rest/admin_api.go
```diff
@@ -1439,6 +1450,16 @@ func (h *handler) handleSGCollect() error {
...
 
+	auditFields := base.AuditFields{
+		"output_dir":   params.OutputDirectory,
+		"upload_host":  params.UploadHost,
+		"customer":     params.Customer,
+		"ticket":       params.Ticket,
+		"keep_zip":     params.KeepZip,
+		"zip_filename": zipFilename,
+	}
+	base.Audit(h.ctx(), base.AuditIDSyncGatewayCollectInfoStop, auditFields)
```
torcolvin, line 1531: This events should have `mandatoryFieldGroups: []fieldGroup{fieldGroupRequest, fieldGroupAuthenticated}`

rest/api.go
torcolvin: I think this case should have a file listed since this is writing a file. I thought maybe we should make file a mandatory but the `net/http/pprof` endpoints write to a response, so that doesn't make sense.

I do think there is something to consider the difference between cpu profile (stop) ("collect") vs the handler below:

```
base.Audit(h.ctx(), base.AuditIDSyncGatewayProfiling, base.AuditFields{"profile_type": "cpu (start)", "filename": params.File})
```

I thought about whether there should be an optional action that is "start" and "write", and how to reconcile those with the pprof handler that uses query parameters to take a profile over time.

rest/api.go
torcolvin: is it worth marking `profile_type` a constant and marking the profile types as constants, since they are duplicated between our handlers that write to files and the `net/http/pprof` handlers.

rest/api.go
```diff
@@ -679,13 +694,14 @@ func sleep(rq *http.Request, d time.Duration) {
 func (h *handler) handleStats() error {
 	st := stats{}
 	runtime.ReadMemStats(&st.MemStats)
-
 	h.writeJSON(st)
+
+	base.Audit(h.ctx(), base.AuditIDSyncGatewayStats, base.AuditFields{"stats_format": "memstats"})
```
torcolvin: I think this should have:

```
		mandatoryFieldGroups: []fieldGroup{
			fieldGroupRequest,
			fieldGroupAuthenticated,
		},
```

I also think (but did not research) that there is a behavior about how the auth information is stored on the context for a silent handler.

rest/handler.go
```diff
@@ -266,6 +266,19 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
...
+	auditFields := base.AuditFields{
+		"http_method": h.rq.Method,
+		"http_path":   h.rq.URL.Path,
+	}
+	switch h.serverType {
+	case publicServer:
+		base.Audit(h.ctx(), base.AuditIDPublicHTTPAPIRequest, auditFields)
+	case adminServer:
+		base.Audit(h.ctx(), base.AuditIDAdminHTTPAPIRequest, auditFields)
+	case metricsServer:
```
torcolvin: is this missing an audit event for diagnostics api?

rest/debug.go
```diff
@@ -95,5 +95,7 @@ func (h *handler) handleExpvar() error {
 	base.DebugfCtx(h.ctx(), base.KeyHTTP, "Recording snapshot of current debug variables.")
 	h.rq.URL.Path = strings.Replace(h.rq.URL.Path, kDebugURLPathPrefix, "/debug/vars", 1)
 	http.DefaultServeMux.ServeHTTP(h.response, h.rq)
+
+	base.Audit(h.ctx(), base.AuditIDSyncGatewayStats, base.AuditFields{"stats_format": "expvar"})
```
torcolvin: is it worth having `stats_format` as a constant?

torcolvin (commented):
I am going to come back to this review to check about whether we have all the right mandatory field groups, and whether all profiling endpoints were hit.

base/logger_audit.go
adamcfraser: I assumed this was safe as it's following the same pattern used in the Audit() function.

base/logging_config.go
adamcfraser: In the case where we're disabling audit logging, we need to log the 'disabled' audit log message before we turn off audit logging (i.e. before we call auditLogger.Enabled.Set())

rest/api.go
adamcfraser: Given that this is on the (internal) admin API only I don't think the performance attack is a particular concern.

rest/handler.go
adamcfraser: Diagnostics API is internal-only/unsupported in 3.2.

rest/api.go
adamcfraser: Agree about `profile_type`, left the `net/http/pprof` profile descriptions as-is.

rest/api.go
adamcfraser: Agree about the filename.  As you noted, we're currently handling the collection over time aspect for cpu with the different profile types `cpu (start)` and `cpu`.   Adding a separate `action` felt like overkill.

torcolvin (dismissed):

torcolvin (approved):

PR #6982, opened 2024-07-19
CBG-3947 Do not remove loaded database on transient fetch error

adamcfraser:
For transient fetch errors (recoverable read errors), leave the database intact to avoid database unload/reload churn during transient connectivity issues with server.

Required changing the pre-delete validation check on the config  from _fetchDatabase (which scans all buckets) to GetConfig (which targets the known bucket), as _fetchDatabase can't bubble up the transient error for a particular bucket.

Added a test to ensure the actual delete behaviour still works as expected (TestConfigPollingRemoveDatabase), and performed some manual testing to simulate the transient error (see Force timeouts notes in that test) - this will be required until we implement "leaky bootstrap" test support.

CBG-3947

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2591/

### DIFF
--- a/base/error.go
+++ b/base/error.go
@@ -248,6 +248,25 @@ func IsXattrNotFoundError(err error) bool {
 	return false
 }
 
+func IsRecoverableReadError(err error) bool {
+
+	if err == nil {
+		return false
+	}
+	// SG timeout error
+	if errors.Is(err, ErrTimeout) {
+		return true
+	}
+	// gocb timeouts or transient errors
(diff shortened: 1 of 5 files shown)


adamcfraser:
This should be reworked to use the per-bucket functionality in _fetchDatabase (to handle things like bucket mismatch), instead of a simple config fetch.  Closing for now, retargeted for a future release.

PR #6987, opened 2024-07-19
Audit assertion improvements: Type check mandatory fields

bbrks:
Example output:
```
panic: audit event "54000" (Create database) invalid:
4 errors:
missing mandatory field remote.ip
missing mandatory field remote.port
field value for id must be of type int but had string
field value for local must be of type map[string]interface {} but had string
```

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2594/

### DIFF
--- a/base/audit_types.go
+++ b/base/audit_types.go
@@ -10,6 +10,7 @@ package base
 
 import (
 	"fmt"
+	"reflect"
 	"strconv"
 )
 
@@ -133,7 +134,7 @@ func (f AuditFields) expandMandatoryFieldGroups(groups []fieldGroup) {
 
 func (i AuditID) MustValidateFields(f AuditFields) {
 	if err := i.ValidateFields(f); err != nil {
-		panic(fmt.Errorf("audit event %s(%s) invalid:\n%v", AuditEvents[i].Name, i, err))
+		panic(fmt.Errorf("audit event %q (%s) invalid:\n%v", i, AuditEvents[i].Name, err))
 	}
 }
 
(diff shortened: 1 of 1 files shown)


base/audit_types.go
```diff
@@ -133,7 +134,7 @@ func (f AuditFields) expandMandatoryFieldGroups(groups []fieldGroup) {
 
 func (i AuditID) MustValidateFields(f AuditFields) {
 	if err := i.ValidateFields(f); err != nil {
-		panic(fmt.Errorf("audit event %s(%s) invalid:\n%v", AuditEvents[i].Name, i, err))
+		panic(fmt.Errorf("audit event %q (%s) invalid:\n%v", i, AuditEvents[i].Name, err))
```
bbrks, line 137: I changed this just to align with the existing assertion messages in `TestAuditLoggingFields`

base/audit_types.go
```diff
@@ -151,15 +152,25 @@ func (i AuditID) ValidateFields(f AuditFields) error {
...
-		if _, ok := fields[k]; !ok {
-			me = me.Append(fmt.Errorf("missing mandatory field %s", baseName+k))
-		}
 	}
 	return me.ErrorOrNil()
 }
+
+// matchingTypes returns true if the types of a and b are the same.
+func matchingTypes(a, b any) bool {
+	return reflect.TypeOf(a).Kind() == reflect.TypeOf(b).Kind()
```
bbrks, line 175: This is too strict for fields we rely on JSON marshalling for... Need to rethink with some spare time.

```
field value for timestamp must be of type string but had time.Time
field value for id must be of type int but had uint64
```

bbrks:
will revisit after code complete

PR #6988, opened 2024-07-19
CBG-4046 add attachment and create, update, delete events

torcolvin:
I added tests for REST api and didn't add tests for blip code.

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2613/


db/blip_handler.go
```diff
@@ -1328,6 +1328,11 @@ func (bh *blipHandler) handleGetAttachment(rq *blip.Message) error {
 	response.SetCompressed(rq.Properties[BlipCompress] == trueProperty)
 	bh.replicationStats.HandleGetAttachment.Add(1)
 	bh.replicationStats.HandleGetAttachmentBytes.Add(int64(len(attachment)))
+	base.Audit(bh.loggingCtx, base.AuditIDAttachmentRead, base.AuditFields{
+		base.AuditFieldDocID:        docID,
+		base.AuditFieldDocVersion:   allowedAttachment.docVersion,
+		base.AuditFieldAttachmentID: attachmentKey,
```
adamcfraser: This isn't attachment name (which is how AuditFieldAttachmentID is being used elsewhere).  We don't currently hang on to attachment name after the rev message is sent - blip clients request attachments by digest.  

To support auditing the name I think we'll need to add the name to AttachmentStorageMeta when we build it from AttachmentsMeta, and then from there set it on AllowedAttachment when that gets built.  There are some corner cases though - I believe we allow multiple attachments on the same document with the same digest and different names, and they only get stored/replicated once (by digest).

db/crud.go
```diff
@@ -1902,8 +1918,15 @@ func (col *DatabaseCollectionWithUser) documentUpdateFunc(ctx context.Context, d
 	return updatedExpiry, newRevID, newDoc, oldBodyJSON, unusedSequences, changedAccessPrincipals, changedRoleAccessUsers, createNewRevIDSkipped, err
 }
 
+// updatedAttachments is returned after processing attachments from an inline body
+type updatedAttachments struct {
+	data         AttachmentData // stores the AttachmentData from the changed attachments
+	createdNames []string       // names of the attachments that were created, for logging later
+	updatedNames []string       // names of the attachments that were updated, for logging later
+}
+
```
adamcfraser: I assume you modified the struct in this way (instead of changing the entries in AttachmentData to structs with a created/updated flag) to minimize refactoring/risk - that seems reasonable for 3.2, but maybe file a ticket to refactor post-3.2 to reduce the duplicated storage of attachment names.

db/crud.go
```diff
@@ -2172,13 +2195,23 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
...
 		var obsoleteAttachments []string
-		for previousAttachmentID := range previousAttachments {
+		for previousAttachmentID, previousAttachmentName := range previousAttachments {
 			if _, found := leafAttachments[previousAttachmentID]; !found {
 				err = db.dataStore.Delete(previousAttachmentID)
 				if err != nil {
 					base.ErrorfCtx(ctx, "Error deleting obsolete attachment %q of doc %q, Error: %v", previousAttachmentID, base.UD(doc.ID), err)
 				} else {
 					obsoleteAttachments = append(obsoleteAttachments, previousAttachmentID)
+					if !isImport {
```
adamcfraser, line 2213: There isn't currently a way to change the set of attachments on a document during import, but this is a reasonable safety check.

adamcfraser (changes requested):
Overall approach looks good, some questions about attachment name.  Might merit unit tests to verify auditing of name is consistent across create/update/delete and via different pathways.

db/crud.go
torcolvin: This was just an oversight, addressed.

db/crud.go
torcolvin: This did hit this code pathway in a test. I didn't investigate it but I wonder if it is from having inline attachments? I'm inclined to not investigate it because it seems like the right behavior.

db/attachment.go
```diff
@@ -37,8 +37,15 @@ var (
...
-// AttachmentData holds the attachment key and value bytes.
-type AttachmentData map[string][]byte
+// updatedAttachment represents an attachment that has been modified by a create or update operation.
+type updatedAttachment struct {
+	body    []byte // The attachment data in bytes
+	updated bool   // If this attachment was an update of an existing attachments
+	created bool   // If this attachment was new on the body
+}
+
+// updatedAttachments holds the attachment key and contents of the attachment.
```
adamcfraser: What is 'the attachment key' here?  If we know whether this is attachment name, attachment digest, attachment document key, it would be helpful to mention that here.

db/attachment.go
```diff
@@ -130,14 +143,15 @@ func (db *DatabaseCollectionWithUser) storeAttachments(ctx context.Context, doc
 			}
 		}
 	}
-	return newAttachmentData, nil
+
+	return newAttachments, nil
 }
 
-// retrieveV2AttachmentKeys returns the list of V2 attachment keys from the attachment metadata that can be used for
+// retrieveV2AttachmentIDs returns the list of V2 attachment keys from the attachment metadata that can be used for
```
adamcfraser: What does IDs represent here?   Is it meant to be name, or Attachment document key?

db/attachment.go
```diff
@@ -37,8 +37,15 @@ var (
 	ErrAttachmentMeta = base.HTTPErrorf(http.StatusBadRequest, "Invalid _attachments")
 )
 
-// AttachmentData holds the attachment key and value bytes.
-type AttachmentData map[string][]byte
+// updatedAttachment represents an attachment that has been modified by a create or update operation.
+type updatedAttachment struct {
+	body    []byte // The attachment data in bytes
+	updated bool   // If this attachment was an update of an existing attachments
+	created bool   // If this attachment was new on the body
```
adamcfraser: Given that it's always only one of `updated` or `created`, it feels like two flags aren't necessary?

adamcfraser (dismissed):
Generally looks fine.  Some comment and code style comments, but can leave those to come back to in future.

db/crud.go
```diff
@@ -2100,6 +2115,24 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 		return nil, "", err
 	}
 
+	if doc.IsDeleted() {
```
adamcfraser: I don't believe we want to audit imports as create/delete/update (they are just updating the metadata).  Move all this be inside an `if !isImport { ... }` block?

db/crud.go
```diff
@@ -2100,6 +2115,24 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
...
+		base.Audit(ctx, base.AuditIDDocumentDelete, base.AuditFields{
+			base.AuditFieldDocID:      docid,
+			base.AuditFieldDocVersion: newRevID,
```
adamcfraser: Tombstones can have channels - feels like we should be including base.AuditFieldChannels here too.

base/audit_events.go
```diff
@@ -1088,17 +1087,16 @@ var AuditEvents = events{
 		},
 		mandatoryFieldGroups: []fieldGroup{
 			fieldGroupAuthenticated,
-			fieldGroupDatabase,
 			fieldGroupKeyspace,
-			fieldGroupRequest,
 		},
-
+		optionalFieldGroups: []fieldGroup{
+			fieldGroupRequest, // this will be present everywhere except tests,
```
adamcfraser, line 1130: I expect this comment should match the one for Delete, since exceptions would be the same?

base/audit_events.go
```diff
@@ -1071,9 +1068,11 @@ var AuditEvents = events{
 		},
 		mandatoryFieldGroups: []fieldGroup{
 			fieldGroupAuthenticated,
-			fieldGroupDatabase,
 			fieldGroupKeyspace,
 		},
+		optionalFieldGroups: []fieldGroup{
+			fieldGroupRequest, // this is not present on ISGR or import,
```
adamcfraser: Should this be included on AuditIDDocumentUpdate and AuditIDDocumentCreate (or omitted here)?

db/database.go
```diff
@@ -2129,6 +2129,16 @@ func (dbCtx *DatabaseContext) AddDatabaseLogContext(ctx context.Context) context
 	return ctx
 }
 
+func (dbCtx *DatabaseContext) AddBucketUserLogContext(ctx context.Context) context.Context {
+	spec := dbCtx.BucketSpec
+	if spec.Server == "" || spec.IsWalrusBucket() {
+		return base.UserLogCtx(ctx, "rosmar_noauth", base.UserDomainBuiltin, nil)
+	}
+	username, _, _ := dbCtx.BucketSpec.Auth.GetCredentials()
+	return base.UserLogCtx(ctx, username, base.UserDomainBuiltin, nil) // FIXME, should I pick up roles?
```
adamcfraser: @bbrks what's your take on this question?

db/database.go
```diff
@@ -2336,6 +2346,7 @@ func (db *DatabaseContext) StartOnlineProcesses(ctx context.Context) (returnedEr
 	// If this is an xattr import node, start import feed.  Must be started after the caching DCP feed, as import cfg
 	// subscription relies on the caching feed.
 	if db.autoImport {
+		ctx := db.AddBucketUserLogContext(ctx)
```
adamcfraser, line 2357: I'm not sure why this is only being set for import - it seems like it would apply to everything in StartOnlineProcesses?

db/util_testing.go
```diff
@@ -676,8 +676,12 @@ func SetupTestDBForDataStoreWithOptions(t testing.TB, tBucket *base.TestBucket,
...
 	require.NoError(t, err, "Couldn't create database 'db'")
 
-	ctx = db.AddDatabaseLogContext(ctx)
-	return db, ctx
+	return db, addDatabaseAndTestUserContext(ctx, db)
+}
+
+// addDatabaseAndTestUserContext adds a fake user to the context
+func addDatabaseAndTestUserContext(ctx context.Context, db *Database) context.Context {
+	return db.AddDatabaseLogContext(base.UserLogCtx(ctx, "gotest", base.UserDomainBuiltin, nil))
```
adamcfraser, line 685: Why is this not handled by the AddBucketUserLogContext?

rest/utilities_testing.go
```diff
@@ -536,7 +536,8 @@ func (rt *RestTester) GetUserAdminAPI(username string) auth.PrincipalConfig {
 // GetSingleTestDatabaseCollection will return a DatabaseCollection if there is only one. Depending on test environment configuration, it may or may not be the default collection.
 func (rt *RestTester) GetSingleTestDatabaseCollection() (*db.DatabaseCollection, context.Context) {
 	c := db.GetSingleDatabaseCollection(rt.TB(), rt.GetDatabase())
-	return c, c.AddCollectionContext(rt.Context())
+	ctx := base.UserLogCtx(c.AddCollectionContext(rt.Context()), "gotest", base.UserDomainBuiltin, nil)
```
adamcfraser, line 538: The UserLogCtx feels like it is getting set arbitrarily in a lot of places.  Can you help me understand the general approach we're using to decide whether to set "gotest"?  I would have thought a DatabaseCollection shouldn't have a user context (since it's not a DatabaseCollectionWithUser).

rest/doc_api.go
```diff
@@ -644,6 +648,10 @@ func (h *handler) handleGetLocalDoc() error {
 	}
 	value[db.BodyId] = "_local/" + docid
 	h.writeJSON(value)
+	base.Audit(h.ctx(), base.AuditIDDocumentRead, base.AuditFields{
+		base.AuditFieldDocID:      docid,
+		base.AuditFieldDocVersion: value[db.BodyRev], // this value might not be populated if this is a doc that starts with _sync
```
adamcfraser: I agree, I don't think BodyRev is a required field for local docs.  Can we just remove this?

rest/audit_test.go
```diff
@@ -974,6 +1120,22 @@ func requireDocumentReadEvents(rt *RestTester, output []byte, docID string, docV
 	require.Equal(rt.TB(), docVersions, docVersionsFound)
 }
 
+// requireAttachmentEvents validates that create attachment events
+func requireAttachmentEvents(rt *RestTester, eventID base.AuditID, output []byte, docID, docVersion string, count int) {
+	events := jsonLines(rt.TB(), output)
+	countFound := 0
+	for _, event := range events {
+		// skip events that are not document read events
```
adamcfraser:
```suggestion
		// skip events that are not the target eventID
```

rest/audit_test.go
```diff
@@ -974,6 +1120,22 @@ func requireDocumentReadEvents(rt *RestTester, output []byte, docID string, docV
 	require.Equal(rt.TB(), docVersions, docVersionsFound)
 }
 
+// requireAttachmentEvents validates that create attachment events
```
adamcfraser: Is the idea that this expects that eventID is one of the attachment crud event IDs?

adamcfraser (changes requested):

db/database.go
bbrks: Not needed. Roles are passed for filtering only. In the case of `builtin` users, people can use the username based filtering if they don't want these audit events.

db/database.go
torcolvin: I wanted to be sure that we weren't passing this anywhere we shouldn't be, but I think this is probably true.

db/util_testing.go
torcolvin: This is for code that works directly in a test: `collection.Put(ctx, ..)` needs a user and this provides a user. I wanted to make it clear that this wasn't a real user. This is only used by the db package.

rest/utilities_testing.go
torcolvin: This is for code that works directly in a test: collection.Put(ctx, ..) needs a user and this provides a user. I wanted to make it clear that this wasn't a real user. This is for cases like import tests where we directly call functions outside of REST calls.

-run
bbrks, line 1: accidental files included

rest/bulk_api.go
```diff
@@ -458,6 +458,19 @@ func (h *handler) handleBulkGet() error {
...
+					base.WarnfCtx(h.ctx(), "Unexpected format of attachments in the body %+v", atts)
+				}
+				for attachment := range attsMap {
+					base.Audit(h.ctx(), base.AuditIDAttachmentRead, base.AuditFields{
+						base.AuditFieldDocID:        docid,
+						base.AuditFieldDocVersion:   revid,
+						base.AuditFieldAttachmentID: attachment,
+					})
+				}
+			}
```
bbrks: This is only reading metadata of the attachment, unless `includeAttachments` is set.

Does this need to be conditional?

db/raw_doc_with_inline_sync.json
```diff
@@ -0,0 +1,14 @@
+{
```
adamcfraser: I think I'd rather be more deliberate about switching to the use of embed and standalone json files for testing, instead of just adding this single standalone file - would like to think about organization, packaging, etc.  Any reason this can't be inline with the test file, like we do in other spots?

db/crud.go
```diff
@@ -2355,7 +2399,7 @@ func (db *DatabaseCollectionWithUser) DeleteDoc(ctx context.Context, docid strin
 }
 
 // Purges a document from the bucket (no tombstone)
-func (db *DatabaseCollectionWithUser) Purge(ctx context.Context, key string) error {
+func (db *DatabaseCollectionWithUser) Purge(ctx context.Context, key string, needsAudit bool) error {
```
adamcfraser, line 2402: I'm not seeing any cases that call this with needsAudit=false.  Are there such cases?

adamcfraser (changes requested):

db/crud.go
torcolvin: The single case is https://github.com/couchbase/sync_gateway/blob/1fff8bb30eb7df5d2523987fef250dce740d7175/db/database.go#L1497

adamcfraser (dismissed):

bbrks (approved):

PR #6997, opened 2024-07-22
CBG-4090 Enable fieldGroupAuthenticated

adamcfraser:
CBG-4090

For HTTPRequest audit events, makes AuditFieldRealUserID an optional field, to handle non-authenticated endpoints and use cases.  Adds status as a required field to those endpoints to reliably identify authenticated vs. non-authenticated calls.

Standardizes on "user" for property name in real and effective user ID, switches to event field constants for this use cases.

Adds a hardcoded real userid for cases where auth is disabled on admin or metrics port - in that case uses (user,domain) of ("admin_noauth", "sgw_admin").  Open to naming suggestions if there's something more appropriate.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2602/

### DIFF
--- a/base/logging_context.go
+++ b/base/logging_context.go
@@ -267,10 +267,13 @@ type EffectiveUserPair struct {
 type UserIDDomain string
 
 const (
-	UserDomainSyncGateway UserIDDomain = "sgw"
-	UserDomainCBServer    UserIDDomain = "cbs"
-	UserDomainBuiltin                  = "builtin" // internal users (e.g. SG bootstrap user)
+	UserDomainSyncGateway      UserIDDomain = "sgw"
+	UserDomainCBServer         UserIDDomain = "cbs"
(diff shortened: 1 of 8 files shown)


base/logging_context.go
```diff
@@ -267,11 +267,14 @@ type EffectiveUserPair struct {
 type UserIDDomain string
 
 const (
-	UserDomainSyncGateway UserIDDomain = "sgw"
-	UserDomainCBServer    UserIDDomain = "cbs"
-	UserDomainBuiltin                  = "builtin" // internal users (e.g. SG bootstrap user)
+	UserDomainSyncGateway      UserIDDomain = "sgw"
+	UserDomainCBServer         UserIDDomain = "cbs"
+	UserDomainSyncGatewayAdmin UserIDDomain = "sgw_admin" // domain for SGW admin API when admin auth disabled
```
torcolvin, line 272: I think I'd probably add `sgw_admin_noauth` here, but I don't feel strongly about this.

torcolvin (approved):

PR #7010, opened 2024-07-23
CBG-4101 add AuditIDChangesFeedStarted from blip

torcolvin:


base/audit_events.go
```diff
@@ -814,6 +813,9 @@ var AuditEvents = events{
 			AuditFieldFeedType:    "continuous, normal, longpoll, websocket, etc.",
 			AuditFieldIncludeDocs: true,
 		},
+		optionalFieldGroups: []fieldGroup{
+			fieldGroupRequest, // active replicator push doesn't have local and remote
```
adamcfraser: Do we want to audit usage by the active push replicator?

db/blip_handler.go
```diff
@@ -457,6 +457,18 @@ func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions
 		return false
 
 	}
+	auditFields := base.AuditFields{
+		base.AuditFieldSince:    options.Since.String(),
+		base.AuditFieldChannels: channelSet,
```
adamcfraser: AuditFieldChannels and AuditFieldDocIDs ares optional, should only be populated when non-empty.

If we're trying to align with the REST-based events, we should populated AuditFieldFilter based on whether channels or docIDs are empty.

adamcfraser (changes requested):

db/blip_handler.go
torcolvin: Agreed, I thought erroneously that the way this marshaled it would omit empty fields. I've added testing for this.

base/audit_events.go
torcolvin: We decided not to do this.

db/blip_handler.go
```diff
@@ -365,6 +365,29 @@ func (bh *blipHandler) handleSubChanges(rq *blip.Message) error {
...
+	var filters []string
+	if len(subChangesParams.docIDs()) > 0 {
+		auditFields[base.AuditFieldDocIDs] = subChangesParams.docIDs()
+		filters = append(filters, base.DocIDsFilter)
+	}
+	if subChangesParams.filter() != "" {
+		filters = append(filters, subChangesParams.filter())
+	}
+	if len(filters) > 0 {
+		auditFields[base.AuditFieldFilter] = strings.Join(filters, ",")
```
adamcfraser: We don't support multiple filters for a changes request.  We can filter either by channels (base.ByChannelsFilter) or by doc ids (base.DocIDsFilter).  If it's the former, we also set the channels property - if it's the latter we set docIDs.

adamcfraser (changes requested):

db/blip_handler.go
torcolvin: We discussed a bit offline and I worry that it is not actually enforced by CBL protocol - 

https://github.com/couchbase/couchbase-lite-core/blob/master/modules/docs/pages/replication-protocol.adoc#subchanges

The cbl code also doesn't enforce this (at this level): https://github.com/couchbase/couchbase-lite-core/blob/bc77f743d13fb6de86e56167425cdb3252ae7c71/Replicator/Puller.cc#L82

It can only set filter to be `sync_gateway/bychannel` but it looks like it independently passes docids here. For REST, these are exclusive.

In practice, lite may never access this code pathway, or it might be validated elsewhere.

adamcfraser: In GenerateChanges you can see that if docIDs are specified, it's going to use DocIDChangesFeed.  And DocIDChangesFeed doesn't do anything with channels, even if CBL decided to go rogue and send both.  

So from an auditing perspective, I would audit the filter as DocIDs (only) if DocIDs are present, as that's going to be the Sync Gateway behaviour.

adamcfraser (approved):

PR #7014, opened 2024-07-25
CBG-4088: If we error in removeCorruptConfigIfExists we don't unload/remove database

gregns1:
CBG-4088

- add not found check in `removeCorruptConfigIfExists`, if doc not found error, we should continue on remove the invalid db config and carry on with db creation 
- in error scenarios (not doc not found errors) we need to unload the database before exiting create db function

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2618/

### DIFF
--- a/rest/config_test.go
+++ b/rest/config_test.go
@@ -2997,3 +2997,68 @@ func TestInvalidDbConfigNoLongerPresentInBucket(t *testing.T) {
 	resp = rt.CreateDatabase(dbName, dbConfig)
 	RequireStatus(t, resp, http.StatusCreated)
 }
+
+// TestNotFoundOnInvalidDatabase:
+//   - Create rest tester with large config polling interval
+//   - Insert a bad dbConfig into the bucket
+//   - Manually fetch and load db from buckets
+//   - Assert that the bad config is tracked as invalid config
+//   - Delete the bad config manually and attempt to correct the db config through create db endpoint
+//   - Assert db is removed form invalid db's and is now a running database on server context
+func TestNotFoundOnInvalidDatabase(t *testing.T) {
+	rt := NewRestTester(t, &RestTesterConfig{
+		CustomTestBucket: base.GetTestBucket(t),
+		PersistentConfig: true,
+		MutateStartupConfig: func(config *StartupConfig) {
+			// configure the interval time to not run
+			config.Bootstrap.ConfigUpdateFrequency = base.NewConfigDuration(100 * time.Second)
+		},
+		DatabaseConfig: nil,
+	})
+	defer rt.Close()
+	realBucketName := rt.CustomTestBucket.GetName()
+
+	// create a new invalid db config and persist to bucket
+	badName := "badBucketName"
+	dbConfig := rt.NewDbConfig()
+	dbConfig.Name = "db1"
+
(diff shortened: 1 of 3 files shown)


rest/config_test.go
```diff
@@ -2928,3 +2929,68 @@ func makeScopesConfigWithDefault(scopeName string, collections []string) *Scopes
...
+		MutateStartupConfig: func(config *StartupConfig) {
+			// configure the interval time to not run
+			config.Bootstrap.ConfigUpdateFrequency = base.NewConfigDuration(100 * time.Second)
+		},
+		DatabaseConfig: nil,
+	})
+	defer rt.Close()
+	realBucketName := rt.CustomTestBucket.GetName()
+
+	// create a new invalid db config and persist to bucket
```
torcolvin, line 3021: I'd be inclined to parameterize this test and have it work in two ways, just to be sure this works.

1. Have the initial creation be invalid, mimicking what happens on a sync gateway startup when it sees an invalid config.
2. Set a valid config first, then have config polling pick up an invalid config. I _think_ this should go through the same code path but I'm not 100% sure. This would be a running sync gateway and then having an restore/xdcr to a bucket.

rest/config_test.go
```diff
@@ -2928,3 +2929,68 @@ func makeScopesConfigWithDefault(scopeName string, collections []string) *Scopes
...
+	// manually fetch and load db configs from bucket
+	_, err = rt.ServerContext().fetchAndLoadConfigs(rt.Context(), false)
+	require.NoError(t, err)
+
+	// assert the config is picked as invalid db config
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		invalidDatabases := rt.ServerContext().AllInvalidDatabaseNames(t)
+		assert.Equal(c, 1, len(invalidDatabases))
+	}, time.Second*10, time.Millisecond*100)
+
```
torcolvin, line 3050: I think it's worth adding an assertion here to make sure that `h.server.invalidDatabaseConfigTracking` is correctly populated, so I think this means that `GET /db/` will return:

```
		httpErr = base.HTTPErrorf(http.StatusNotFound, "Mismatch in database config for database %s bucket name: %s and backend bucket: %s groupID: %s You must update database config immediately", base.MD(name), base.MD(invalidConfig.configBucketName), base.MD(invalidConfig.persistedBucketName), base.MD(sc.Config.Bootstrap.ConfigGroupID))
```

torcolvin (commented):
LGTM, I think adding some additional testing would make me even more confident.

rest/config_test.go
gregns1: `AllInvalidDatabaseNames` grabs db names from the invalidDatabaseConfigTracking map so I think this tests this, but will add a call to get db status too.

rest/config_test.go
gregns1: 1. I feel like this test already does this? I insert directly to the bucket a bad config and call fetch and load configs to actually load the db. I feel this mimics startup of a node and a config already being there in bucket? 
2. This is covered by the test written for the original PR for this work, TestCorruptDbConfigHandling is an example

torcolvin (approved):

PR #7016, opened 2024-07-26
CBG-4067: do not release doc sequence upon timeout error

gregns1:
CBG-4067

- Added check to only release doc sequences when a write error is **not** a timeout error. 
- Added some functionality to leaky bucket to force return a timeout error upon a successful write (open to ideas of better forcing a timeout here) 
- Test also covers us correctly releasing a doc sequence upon a different et error at write time by casuing conflict uszing leaky bucket functionality 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2617/

### DIFF
--- a/base/error.go
+++ b/base/error.go
@@ -237,7 +237,7 @@ func IsDocNotFoundError(err error) bool {
 	}
 }
 
-func IsTimeoutError(err error) bool {
+func IsTemporaryKvError(err error) bool {
 	if err == nil {
 		return false
 	}
@@ -250,6 +250,16 @@ func IsTimeoutError(err error) bool {
 	if errors.Is(err, gocb.ErrUnambiguousTimeout) {
 		return true
 	}
+	if errors.Is(err, gocb.ErrOverload) {
+		return true
+	}
+	if errors.Is(err, gocb.ErrTemporaryFailure) {
+		return true
+	}
+	if errors.Is(err, gocb.ErrCircuitBreakerOpen) {
+		return true
+	}
+
 	return false
 }
 
--- a/base/error.go
+++ b/base/error.go
@@ -237,27 +237,21 @@ func IsDocNotFoundError(err error) bool {
 	}
 }
 
+// IsTemporaryKvError returns true if a kv operation has an error that is likely to be ephemeral. This represents
+// situations where Couchbase Server is under load and would be expected to return a success or failure in a future call.
 func IsTemporaryKvError(err error) bool {
 	if err == nil {
 		return false
 	}
-	if errors.Is(err, ErrTimeout) {
-		return true
-	}
-	if errors.Is(err, gocb.ErrAmbiguousTimeout) {
-		return true
-	}
-	if errors.Is(err, gocb.ErrUnambiguousTimeout) {
-		return true
-	}
-	if errors.Is(err, gocb.ErrOverload) {
-		return true
-	}
-	if errors.Is(err, gocb.ErrTemporaryFailure) {
-		return true
-	}
-	if errors.Is(err, gocb.ErrCircuitBreakerOpen) {
-		return true
+	// define list of temporary errors
+	temporaryKVError := []error{ErrTimeout, gocb.ErrAmbiguousTimeout, gocb.ErrUnambiguousTimeout,
+		gocb.ErrOverload, gocb.ErrTemporaryFailure, gocb.ErrCircuitBreakerOpen}
+
+	// iterate through to check incoming error is one of them
+	for _, tempKVErr := range temporaryKVError {
+		if errors.Is(err, tempKVErr) {
+			return true
+		}
 	}
 
 	return false
--- a/base/error.go
+++ b/base/error.go
@@ -237,6 +237,22 @@ func IsDocNotFoundError(err error) bool {
 	}
 }
 
+func IsTimeoutError(err error) bool {
+	if err == nil {
+		return false
+	}
+	if errors.Is(err, ErrTimeout) {
+		return true
+	}
+	if errors.Is(err, gocb.ErrAmbiguousTimeout) {
+		return true
+	}
+	if errors.Is(err, gocb.ErrUnambiguousTimeout) {
+		return true
+	}
+	return false
+}
+
 func IsXattrNotFoundError(err error) bool {
 	if unwrappedErr := pkgerrors.Cause(err); unwrappedErr == nil {
 		return false
--- a/base/leaky_bucket.go
+++ b/base/leaky_bucket.go
@@ -121,6 +121,8 @@ type LeakyBucketConfig struct {
 
 	ForceErrorSetRawKeys []string // Issuing a SetRaw call with a specified key will return an error
 
+	ForceTimeoutErrorOnUpdateKeys []string // Specified keys will return timeout error AFTER write is sent to server
+
 	// Returns a partial error the first time ViewCustom is called
 	FirstTimeViewCustomPartialError bool
 	PostQueryCallback               func(ddoc, viewName string, params map[string]interface{}) // Issues callback after issuing query when bucket.ViewQuery is called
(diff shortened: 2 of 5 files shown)


base/error.go
```diff
@@ -237,6 +237,22 @@ func IsDocNotFoundError(err error) bool {
 	}
 }
 
+func IsTimeoutError(err error) bool {
```
torcolvin: I think we want to include `gocb.ErrOverload` and `gocb.ErrTemporaryFailure` and `gocb.ErrCircuitBreakerOpen`. This might make the function `IsTemporaryKvError()` 

I wonder if we want to scope this message to just kv write error errors. I'd be curious if `ErrRateLimitedFailure` or `ErrQuotaLimitedFailure` could occur from kv.

db/crud_test.go
```diff
@@ -1680,3 +1680,87 @@ func TestAssignSequenceReleaseLoop(t *testing.T) {
...
 	assert.Equal(t, int64(expectedReleasedSequenceCount), releasedSequenceCount)
 }
+
+// TestReleaseSequenceOnDocWrite:
+//   - Define a leaky bucket callback for a conflicting write + define key to return a timeout error for
+//   - Setup db with leaky bucket config
+//   - Init a channel cache by calling changes
+//   - Write a doc that will return timeout error but will successfully persist
+//   - Wait for it to arrive at change cache
+//   - Assert we don;t release a sequence for it + we hav eit in changes
```
torcolvin:
```suggestion
//   - Assert we don't release a sequence for it
```

I'm not sure what "we have it in changes" means

db/crud_test.go
```diff
@@ -1680,3 +1680,87 @@ func TestAssignSequenceReleaseLoop(t *testing.T) {
...
+//   - Write new doc with conflict error
+//   - Assert we release a sequence for this
+func TestReleaseSequenceOnDocWriteFailure(t *testing.T) {
+	defer SuspendSequenceBatching()()
+	base.SetUpTestLogging(t, base.LevelDebug, base.KeyAll)
+
+	var ctx context.Context
+	var db *Database
+	var enable bool
```
torcolvin: I think having this called `forceDocConflict` would be good.

db/crud_test.go
```diff
@@ -1680,3 +1680,87 @@ func TestAssignSequenceReleaseLoop(t *testing.T) {
...
+
+	const (
+		conflictDoc = "doc1"
+		timeoutDoc  = "doc"
+	)
+
+	// call back to create a conflict mid write and force a non timeout error upon write attempt
+	writeUpdateCallback := func(key string) {
+		if key == "doc1" && enable {
```
torcolvin:
```suggestion
		if key == conflictDoc && enable {
```

db/crud_test.go
```diff
@@ -1680,3 +1680,87 @@ func TestAssignSequenceReleaseLoop(t *testing.T) {
...
+			enable = false
+			body := Body{"test": "doc"}
+			collection, ctx := GetSingleDatabaseCollectionWithUser(ctx, t, db)
+			_, _, err := collection.Put(ctx, conflictDoc, body)
+			assert.NoError(t, err)
```
torcolvin:
```suggestion
			require.NoError(t, err)
```

I think this should stop the test because anything after this would not be good.

db/crud_test.go
```diff
@@ -1680,3 +1680,87 @@ func TestAssignSequenceReleaseLoop(t *testing.T) {
...
+		ForceTimeoutErrorOnUpdateKeys: []string{timeoutDoc},
+	}
+
+	db, ctx = setupTestLeakyDBWithCacheOptions(t, DefaultCacheOptions(), callbackConfig)
+	defer db.Close(ctx)
+	collection, ctx := GetSingleDatabaseCollectionWithUser(ctx, t, db)
+
+	// init channel cache
+	_, err := collection.GetChanges(ctx, base.SetOf("*"), getChangesOptionsWithZeroSeq(t))
+	require.NoError(t, err)
```
torcolvin: Why is it required to initialize channel cache? I just don't understand why.

base/error.go
```diff
@@ -237,6 +237,32 @@ func IsDocNotFoundError(err error) bool {
 	}
 }
 
+func IsTemporaryKvError(err error) bool {
```
torcolvin, line 242:
```suggestion
// IsTemporaryKvError returns true if a kv operation has an error that is likely to be ephemeral. This represents situations where Couchbase Server is under load and would be expected to return a success or failure in a future call.
func IsTemporaryKvError(err error) bool {
```

base/error.go
```diff
@@ -237,6 +237,32 @@ func IsDocNotFoundError(err error) bool {
...
+	}
+	if errors.Is(err, gocb.ErrOverload) {
+		return true
+	}
+	if errors.Is(err, gocb.ErrTemporaryFailure) {
+		return true
+	}
+	if errors.Is(err, gocb.ErrCircuitBreakerOpen) {
+		return true
+	}
```
torcolvin: nit style, maybe this is easier to read because you know these will all flag as true:

```suggestion
	for _, tempKVErr := range []error{ErrTimeout, gocb.ErrAmbiguousTimeout, gocb.ErrUnambiguousTimeout, gocb.ErrOverload, gocb.ErrTemporaryFailure, gocb.ErrCircuitBreakerOpen}) {
	    if errors.Is(err, tempKVErr) {
			return true
		}
	}
```

base/leaky_bucket.go
```diff
@@ -121,6 +121,8 @@ type LeakyBucketConfig struct {
 
 	ForceErrorSetRawKeys []string // Issuing a SetRaw call with a specified key will return an error
 
+	ForceTimeoutErrorOnUpdateKeys []string // Specified keys will return timeout error AFTER write is sent to server
```
torcolvin, line 124: This is interesting because this definitely means the write occurred. However, in the case of this ticket, we don't know whether the write went through or not? Is there a reason to actually have this error after it writes and not before?

base/leaky_datastore.go
```diff
@@ -153,7 +153,13 @@ func (lds *LeakyDataStore) Update(k string, exp uint32, callback sgbucket.Update
 			lds.config.UpdateCallback(k)
 			return updated, expiry, isDelete, err
 		}
-		return lds.dataStore.Update(k, exp, wrapperCallback)
+		casOut, err = lds.dataStore.Update(k, exp, wrapperCallback)
+		for _, errorKey := range lds.config.ForceTimeoutErrorOnUpdateKeys {
```
torcolvin: nit: you can do

`slices.Contains(err, lds.config.ForceTimeoutErrorOnUpdateKeys)`

torcolvin (commented):
This LGTM, just some questions about the tests, and some code nits you can take or leave.

db/crud_test.go
gregns1: I have added more to the comment. Its needed as I want to force the changes call made after the timeout doc is written to fetch from the cache. You'll notice if you remove the error handling I added in `updateAndReturnDoc` and run this test it will fail on that changes call, because the timeout doc is ignored as a duplicate at the cache after an unused doc is written with the same sequence. This is the issue the customer hit when I spotted it in the logs.

base/leaky_bucket.go
gregns1: I want this behaviour as this is what the customer saw. See ticket linked to this for more info but they had write, server under load and operation times out, but write eventually succeeds but we've timeout and release the document sequence. Then then the successful write is ignored at cache due to duplicate sequence with the released sequence. 

It certainly seems odd to do it this way but its the only way I could think of mocking/forcing this situation reliably in a test. We cannot be sure when we hit timeout that the write will eventually go through or not but the changes this PR makes will protect against docs being missed in replications if the write does succeed.

torcolvin (approved):

PR #7020, opened 2024-07-26
CBG-4087: remove configs from invalid tracking that are not found in config poll

gregns1:
CBG-4087

- added `removeNonExistingConfigs` that will iterate through invalid db's and check they are present in config map that I build for **all** configs fetched for each bucket. The `fetchedConfigs` map will never include invalid configs from the bucket so couldn't use that.
- Call this method in `FetchConfigs` which is used by config polling process

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2621/

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -399,6 +399,18 @@ func (d *invalidDatabaseConfigs) remove(dbname string) {
 	delete(d.dbNames, dbname)
 }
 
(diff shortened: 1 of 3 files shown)


adamcfraser (approved):
LGTM.  Had a concern about duplicate db names across buckets, but I think that's potentially a different issue/enhancement.

PR #7021, opened 2024-07-28
CBG-4114 Fixes for db scoped audit enabled/disabled events

adamcfraser:
CBG-4114

Only emit audit events when the audit enabled/disabled state changes for a db.

This changes handling for non-filterable events in logger_audit to skip subsequent enabled checking.  Required to emit db-level audit change events even if db auditing is disabled.  

Also added the ability to disable the 1.5s sequence wait on db restart for tests that don’t require it.

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2622/

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -71,6 +71,10 @@ var (
 	DefaultQueryPaginationLimit = 5000
 )
 
+var (
+	BypassReleasedSequenceWait atomic.Bool
+)
+
 const (
 	CompactIntervalMinDays = float32(0.04) // ~1 Hour in days
 	CompactIntervalMaxDays = float32(60)   // 60 Days in days
@@ -2340,7 +2344,7 @@ func (db *DatabaseContext) StartOnlineProcesses(ctx context.Context) (returnedEr
 
 	// Unlock change cache.  Validate that any allocated sequences on other nodes have either been assigned or released
 	// before starting
-	if initialSequence > 0 {
+	if initialSequence > 0 && !BypassReleasedSequenceWait.Load() {
 		_ = db.sequences.waitForReleasedSequences(ctx, initialSequenceTime)
 	}
 
--- a/db/database.go
+++ b/db/database.go
@@ -72,7 +72,7 @@ var (
 )
 
 var (
-	BypassReleasedSequenceWait atomic.Bool
+	BypassReleasedSequenceWait atomic.Bool // Used to optimize single-node testing, see DisableSequenceWaitOnDbRestart
 )
 
 const (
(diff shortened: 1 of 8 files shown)


db/util_testing.go
```diff
@@ -787,3 +787,13 @@ func RawDocWithInlineSyncData(_ testing.TB) string {
 }
 `
 }
+
+// DisableSequenceWaitOnDbStart disables the release sequence wait on db start.  Appropriate for tests
+// that make changes to database config after first startup, and don't assert/require on sequence correctness
+func DisableSequenceWaitOnDbRestart(tb testing.TB) {
```
torcolvin, line 793: This is just to speed up this test, right?

I wonder if a better fix is to disable this _everywhere_ via `NewRestTester` unless it is declared it is needed. This would speed up tests by a bit. Do you think this is worth filing a separate ticket to address for other tests that are affected?

db/database.go
```diff
@@ -71,6 +71,10 @@ var (
 	DefaultQueryPaginationLimit = 5000
 )
 
+var (
+	BypassReleasedSequenceWait atomic.Bool
```
torcolvin: I would drop a comment that this global is here for testing.

rest/server_context.go
```diff
@@ -843,18 +844,22 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
...
-
-	// If audit logging is enabled for the node, log the db setting.  This may not represent a change in
-	// auditing since the last time the db was running on this node, but we don't have any insight into the
-	// previous state.
-	if base.IsAuditEnabled() {
-		if contextOptions.LoggingConfig.DbAuditEnabled() {
-			base.Audit(ctx, base.AuditIDAuditEnabled, base.AuditFields{base.AuditFieldAuditScope: "db", base.AuditFieldDatabase: dbName})
-		} else {
-			base.Audit(ctx, base.AuditIDAuditDisabled, base.AuditFields{base.AuditFieldAuditScope: "db", base.AuditFieldDatabase: dbName})
+	/*
```
torcolvin: This code is now commented out, which doesn't seem right.

db/util_testing.go
adamcfraser: In general this doesn't slow down most tests because the wait only happens if initialSequence > 0 (i.e. only tests where the database gets restarted after some sequence allocation work has been done, which is relatively rare), so I didn't want to immediately add it to the default RestTester.  

However, I agree it should get evaluated further - filed [CBG-4124](https://issues.couchbase.com/browse/CBG-4124)

rest/server_context.go
adamcfraser: As discussed, removed this as it's being replaced by the db logging on config change.

torcolvin (approved):

PR #7024, opened 2024-07-29
CBG-4111: Differentiate between `nil` set and empty set of enabled db audit events

bbrks:
CBG-4111

Closes up a bug where we didn't correctly apply a change on top of the default set of events because we couldn't determine whether `nil` meant the default set, or none set.

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2624/

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -746,8 +746,10 @@ func (h *handler) handleGetDbAuditConfig() error {
 		// grab runtime version of config, so that we can see what events would be enabled
 		if runtimeConfig.Logging != nil && runtimeConfig.Logging.Audit != nil {
 			dbAuditEnabled = base.BoolDefault(runtimeConfig.Logging.Audit.Enabled, false)
-			for _, event := range runtimeConfig.Logging.Audit.EnabledEvents {
-				enabledEvents[base.AuditID(event)] = struct{}{}
+			if runtimeConfig.Logging.Audit.EnabledEvents != nil {
+				for _, event := range *runtimeConfig.Logging.Audit.EnabledEvents {
+					enabledEvents[base.AuditID(event)] = struct{}{}
+				}
 			}
 			dbAuditDisabledUsers = runtimeConfig.Logging.Audit.DisabledUsers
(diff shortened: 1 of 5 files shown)


rest/admin_api.go
```diff
@@ -877,14 +879,14 @@ func mutateConfigFromDbAuditConfigBody(isReplace bool, existingAuditConfig *DbAu
 		existingAuditConfig.DisabledRoles = requestAuditConfig.DisabledRoles
 
 		// we don't need to do anything to "disable" events, other than not enable them
-		existingAuditConfig.EnabledEvents = func() []uint {
+		existingAuditConfig.EnabledEvents = func() *[]uint {
```
torcolvin, line 882: I would add a comment as to why you are using `* []uint` which seems a bit weird since `[]` is typically a pointer.

rest/config.go
```diff
@@ -1068,12 +1068,14 @@ func (dbConfig *DbConfig) validateVersion(ctx context.Context, isEnterpriseEditi
...
-				id := base.AuditID(id)
-				if e, ok := base.AuditEvents[id]; !ok {
-					multiError = multiError.Append(fmt.Errorf("unknown audit event ID %q", id))
-				} else if e.IsGlobalEvent {
-					multiError = multiError.Append(fmt.Errorf("event %q is not configurable at the database level", id))
+			if dbConfig.Logging.Audit.EnabledEvents != nil {
+				for _, id := range *dbConfig.Logging.Audit.EnabledEvents {
+					id := base.AuditID(id)
+					if e, ok := base.AuditEvents[id]; !ok {
+						multiError = multiError.Append(fmt.Errorf("unknown audit event ID %q", id))
```
torcolvin, line 1075: I feel like it is helpful in error message to provide the name of the event, but I do not care too much.

torcolvin (approved):

rest/config.go
bbrks: If it's an unknown ID we don't have a name :)

PR #7025, opened 2024-07-29
CBG-3713 use atomic pointers for loggers

torcolvin:
This has https://github.com/couchbase/sync_gateway/pull/7022 in the same changes.

`ResetGlobalTestLogging` didn't in fact reset global logging because it used pointers to refer to the original loggers. Additionally, there are so many pointers in `FileLogger` that even if we make a copy of the struct, it is not a deep copy. Addressed this by reinitializing the state to the start of the bucket state.

Moved tests that were hidden behind `-race` flags to non protected states. I used `SG_TEST_LOG_LEVEL=trace` which will log in the bucket pool while a test is being torn down even under rosmar to flush out any race conditions. The only new code in rest package is adding `base.ResetGlobalTestLogging(t)` to `TestDefaultLogging` to reset the global logger state. This is not done if `SG_TEST_LOG_LEVEL` is set.

This should address https://issues.couchbase.com/browse/CBG-4109 and https://issues.couchbase.com/browse/CBG-3740

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2625/


base/logger_external.go
```diff
@@ -47,7 +47,8 @@ func initExternalLoggers() {
 
 func updateExternalLoggers() {
 	// use context.Background() since this is called from init or to reset test logging
-	if consoleLogger != nil && consoleLogger.shouldLog(context.Background(), LevelDebug, KeyWalrus) {
+	logger := consoleLogger.Load()
+	if logger.shouldLog(context.Background(), LevelDebug, KeyWalrus) {
```
bbrks, line 51: Can `logger` still be nil here? I can't see an obvious way for that to happen, but you kept that handling in `CBGoUtilsLogger) Level()` below so wanted to check.

base/logging.go
```diff
@@ -124,7 +122,8 @@ func init() {
 	// initializing a logging config, and when running under a test scenario.
 	initialCollationBufferSize := 0
 
-	consoleLogger = mustInitConsoleLogger(context.Background(), &ConsoleLoggerConfig{LogKeys: []string{logKeyNames[KeyHTTP]}, FileLoggerConfig: FileLoggerConfig{Enabled: BoolPtr(true), CollationBufferSize: &initialCollationBufferSize}})
+	consoleLogger.Store(mustInitConsoleLogger(context.Background(), &ConsoleLoggerConfig{LogKeys: []string{logKeyNames[KeyHTTP]}, FileLoggerConfig: FileLoggerConfig{Enabled: BoolPtr(true), CollationBufferSize: &initialCollationBufferSize}}))
+	auditLogger.Store(&AuditLogger{})
```
bbrks: Is this required because https://github.com/couchbase/sync_gateway/tree/CBG-4096-auditLogger isn't merged yet?

Think we may have forgotten about this commit

base/logging.go
```diff
@@ -187,8 +186,8 @@ func LogLevelCtx(ctx context.Context, logLevel LogLevel, logKey LogKey, format s
 // RecordStats writes the given stats JSON content to a stats log file, if enabled.
 // The content passed in is expected to be a JSON dictionary.
 func RecordStats(statsJson string) {
-	if statsLogger != nil {
-		statsLogger.logf(statsJson)
+	if statsLogger.Load() != nil {
+		statsLogger.Load().logf(statsJson)
```
bbrks: avoid double load

```suggestion
	if sl := statsLogger.Load(); sl != nil {
		sl.logf(statsJson)
```

base/util_testing.go
```diff
@@ -568,7 +568,7 @@ var GlobalTestLoggingSet = AtomicBool{}
 
 // SetUpGlobalTestLogging sets a global log level at runtime by using the SG_TEST_LOG_LEVEL environment variable.
 // This global level overrides any tests that specify their own test log level with SetUpTestLogging.
-func SetUpGlobalTestLogging(ctx context.Context, m *testing.M) (teardownFn func()) {
+func SetUpGlobalTestLogging(ctx context.Context) (teardownFn func()) {
```
bbrks, line 571: Would prefer to keep `_ *testing.M` in the signature so it's not accidentally misused.

bbrks (changes requested):
Looks mostly good. A few comments inline

base/logger_external.go
torcolvin: It certainly couldn't be in a test environment, but I am not sure so I'll add it back.

It doesn't matter since `shouldLog` also checks for nil pointer though.

base/util_testing.go
torcolvin: The reason I removed this is that I call this function from `ResetGlobalTestLogging` which only has `* testing.T`. I could keep testing.M and pass nil in `ResetGlobalTestLogging` but I can't think of a great a way to convey this otherwise.

base/logging.go
torcolvin: I've added this back.

bbrks (approved):

PR #7033, opened 2024-07-31
CBG-3818 Create and use GetChanges consistently (part 1)

torcolvin:
Create helper functions for managing _changes requests. This has a few advantages:

1.  Allows recognition of when a changes feed is called so we can call `rt.WaitForPendingChanges`.
2. Removes boilerplate from declaring and unmarshaling changes. This is done in a way where the `ChangesResult` structure was reused and `Last_Seq` was not always cleared. In practice, this didn't cause a problem but having a separate structure is less error prone.

There's a secondary PR handles the changes to `WaitForPendingChanges` that will go on top of this one. It was intentionally omitted to make it easier to read. I recommend merging part 2 into part 1 after reviewing this.


adamcfraser (approved):
LGTM

PR #7034, opened 2024-07-31
CBG-3818 Create and use GetChanges consistently (part 2)

torcolvin:
- changes `WaitForPendingChanges` to do its own assertions to make it easier to use. Not all places actually stopped if changes did not come.
- change `StartPull*` blip tester code to not do external assertions, and do its assertions inside the function

Analyze all `GetChanges` and `StartPull` code to see if they needed `rt.WaitForPendingChanges`. 

There are many places where `rt.WaitForPendingChanges`:

1. `WaitForChanges` will wait for that number of changes by doing a backoff. This could be generally be replaced `WaitForChanges` followed by a single changes request.
2. `WaitForSequence` will wait for a particular sequence number.

This change is on top of https://github.com/couchbase/sync_gateway/pull/7033 and this PR should be merged into part 1's branch and then both PRs could go in as a single commit to main.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2631/

### DIFF
--- a/rest/importuserxattrtest/revcache_test.go
+++ b/rest/importuserxattrtest/revcache_test.go
@@ -73,7 +73,6 @@ func TestUserXattrRevCache(t *testing.T) {
 
 	resp := rt.SendAdminRequest("PUT", "/{{.keyspace}}/"+docKey, `{}`)
 	rest.RequireStatus(t, resp, http.StatusCreated)
-	require.NoError(t, rt.WaitForPendingChanges())
 
 	cas, err := rt.GetSingleDataStore().Get(docKey, nil)
 	require.NoError(t, err)
@@ -159,7 +158,6 @@ func TestUserXattrDeleteWithRevCache(t *testing.T) {
 
 	resp := rt.SendAdminRequest("PUT", "/{{.keyspace}}/"+docKey, `{}`)
 	rest.RequireStatus(t, resp, http.StatusCreated)
-	require.NoError(t, rt.WaitForPendingChanges())
 
 	cas, err := rt.GetSingleDataStore().Get(docKey, nil)
 	require.NoError(t, err)
--- a/rest/importuserxattrtest/revcache_test.go
+++ b/rest/importuserxattrtest/revcache_test.go
@@ -73,6 +73,7 @@ func TestUserXattrRevCache(t *testing.T) {
 
 	resp := rt.SendAdminRequest("PUT", "/{{.keyspace}}/"+docKey, `{}`)
 	rest.RequireStatus(t, resp, http.StatusCreated)
+	rt.WaitForPendingChanges()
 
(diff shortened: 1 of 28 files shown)


rest/revocation_test.go
```diff
@@ -2350,26 +2340,22 @@ func TestRevocationNoRev(t *testing.T) {
...
-		firstOneShotSinceSeq := rt.GetDocumentSequence("doc")
-
 		// OneShot pull to grab doc
-		err := btcRunner.StartOneshotPull(btc.id)
-		assert.NoError(t, err)
+		rt.WaitForPendingChanges()
+		btcRunner.StartOneshotPull(btc.id)
 
 		_ = btcRunner.WaitForVersion(btc.id, docID, version)
+		firstOneShotSinceSeq := rt.GetDocumentSequence("doc")
```
adamcfraser: There are a few spots where you're moving the firstOneShotSinceSeq sequence lookup to happen after StartOneShotPull?  It seems innocuous (the pull shouldn't be modifying the doc sequence on the server) - just curious whether this was required/intentional.

rest/importuserxattrtest/revcache_test.go
```diff
@@ -73,7 +73,6 @@ func TestUserXattrRevCache(t *testing.T) {
 
 	resp := rt.SendAdminRequest("PUT", "/{{.keyspace}}/"+docKey, `{}`)
 	rest.RequireStatus(t, resp, http.StatusCreated)
-	require.NoError(t, rt.WaitForPendingChanges())
```
adamcfraser, line 76: For this (and a few other places where WaitForPendingChanges is being removed) - we're sure this isn't needed to ensure the document hits the channel cache before making the xattr update below?  It's possible this was originally here from an (unnecessary) copy/paste, but I can't tell for sure just looking at it.

adamcfraser (commented):
Generally looks fine - a few questions on some of the specifics just to clarify.

rest/revocation_test.go
torcolvin: This isn't intentional and I'll see if I can back out this move. `GetDocumentSequence` uses `/ks/_raw/doc` endpoint and doesn't need this but I think when reading the code sometimes the sequence was used with `WaitForChanges`  and I wanted to change this to be more clear. I think I want to back out any changes that aren't absolutely necessary.

rest/importuserxattrtest/revcache_test.go
torcolvin: Since this is testing the revcache explicitly, it probably should be here even though the test will pass without it. I'll take a pass and put this back to the way it was.

adamcfraser (approved):

PR #7035, opened 2024-08-02
CBG-3947 Do not remove loaded database on transient fetch error

adamcfraser:
For transient fetch errors (recoverable read errors), leave the database intact to avoid database unload/reload churn during transient connectivity issues with server.

Required changing the pre-delete validation check on the config  from _fetchDatabase (which scans all buckets) to GetConfig (which targets the known bucket), as _fetchDatabase can't bubble up the transient error for a particular bucket.

Added a test to ensure the actual delete behaviour still works as expected (TestConfigPollingRemoveDatabase), and performed some manual testing to simulate the transient error (see Force timeouts notes in that test) - this will be required until we implement "leaky bootstrap" test support.

CBG-3947

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2643/

### DIFF
--- a/base/error.go
+++ b/base/error.go
@@ -244,8 +244,12 @@ func IsTemporaryKvError(err error) bool {
 		return false
 	}
 	// define list of temporary errors
-	temporaryKVError := []error{ErrTimeout, gocb.ErrAmbiguousTimeout, gocb.ErrUnambiguousTimeout,
-		gocb.ErrOverload, gocb.ErrTemporaryFailure, gocb.ErrCircuitBreakerOpen}
+	temporaryKVError := []error{
+		ErrTimeout,                 // Sync Gateway client-side timeout
+		gocb.ErrTimeout,            // SDK op timeout.  Wrapped by gocb.ErrAmbiguousTimeout, gocb.ErrUnambiguousTimeout,
+		gocb.ErrOverload,           // SDK client-side pipeline queue full, request was not submitted to server
+		gocb.ErrTemporaryFailure,   // Couchbase Server returned temporary failure error
+		gocb.ErrCircuitBreakerOpen} // SDK client-side circuit breaker blocked request
 
 	// iterate through to check incoming error is one of them
 	for _, tempKVErr := range temporaryKVError {
@@ -263,25 +267,6 @@ func IsXattrNotFoundError(err error) bool {
 	} else if unwrappedErr == ErrXattrNotFound {
 		return true
 	} else if _, ok := unwrappedErr.(sgbucket.XattrMissingError); ok {
-		return true
-	}
-	return false
-}
-
-func IsRecoverableReadError(err error) bool {
-
-	if err == nil {
-		return false
-	}
-	// SG timeout error
-	if errors.Is(err, ErrTimeout) {
-		return true
-	}
-	// gocb timeouts or transient errors
-	if errors.Is(err, gocb.ErrTemporaryFailure) ||
-		errors.Is(err, gocb.ErrOverload) ||
-		errors.Is(err, gocb.ErrTimeout) ||
-		errors.Is(err, gocb.ErrCircuitBreakerOpen) {
 		return true
 	}
 	return false
--- a/base/error.go
+++ b/base/error.go
@@ -268,6 +268,25 @@ func IsXattrNotFoundError(err error) bool {
 	return false
 }
 
+func IsRecoverableReadError(err error) bool {
+
+	if err == nil {
+		return false
+	}
+	// SG timeout error
+	if errors.Is(err, ErrTimeout) {
+		return true
+	}
+	// gocb timeouts or transient errors
+	if errors.Is(err, gocb.ErrTemporaryFailure) ||
+		errors.Is(err, gocb.ErrOverload) ||
+		errors.Is(err, gocb.ErrTimeout) ||
+		errors.Is(err, gocb.ErrCircuitBreakerOpen) {
+		return true
+	}
+	return false
+}
+
 // MultiError manages a set of errors.  Callers must use ErrorOrNil when returning MultiError to callers
 // in order to properly handle nil checks on the returned MultiError
 //  Sample usage:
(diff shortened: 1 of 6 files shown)


base/error.go
```diff
@@ -268,6 +268,25 @@ func IsXattrNotFoundError(err error) bool {
 	return false
 }
 
+func IsRecoverableReadError(err error) bool {
```
torcolvin: This is a duplicate of `IsTemporaryKvError` - in the case that it is used otherwise it is actually covering a possible write error as well.

There's also `collection.isRecoverableReadError`.

rest/config.go
```diff
@@ -1629,17 +1629,26 @@ func (sc *ServerContext) fetchAndLoadConfigs(ctx context.Context, isInitialStart
...
+		if found && getConfigErr == nil {
+			base.DebugfCtx(ctx, base.KeyConfig, "Found config for database %q after acquiring write lock - not removing database", base.MD(dbName))
+			continue
 		}
+		if base.IsRecoverableReadError(getConfigErr) {
+			base.InfofCtx(ctx, base.KeyConfig, "Transient error fetching config for database %q to check whether we need to remove it, will not be removed: %v", base.MD(dbName), getConfigErr)
+			continue
+		}
+
 		if !found {
```
torcolvin, line 1649: Are you intentionally relying on the fact that `getConfigErr` could be not a recoverable read error that it will also be `found=false` and thus we remove the database?

I think that it is probably the right thing to fall into this code to remove the database, but I'd be tempted to log at least info.

rest/config.go
```diff
@@ -1747,56 +1756,59 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
...
+		cnf.CertPath = sc.Config.Bootstrap.X509CertPath
+		cnf.KeyPath = sc.Config.Bootstrap.X509KeyPath
+	}
+	base.TracefCtx(ctx, base.KeyConfig, "Got database config %s for bucket %q with cas %d and groupID %q", base.MD(dbName), base.MD(bucket), cas, base.MD(sc.Config.Bootstrap.ConfigGroupID))
+	return true, cnf, nil
+}
+
+func (sc *ServerContext) _fetchDatabase(ctx context.Context, dbName string) (found bool, dbConfig *DatabaseConfig, err error) {
+	var cnf DatabaseConfig
+	callback := func(bucket string) (exit bool, err error) {
```
torcolvin: nit: there's some wacky variable shadowing `_fetchDatabase`. 

`found` is declared as a return parameter for `_fetchDatabase` but we never used named returns for `fetchDatabase`. There are also named returns (unused) on `callback`. In this case `err` is actually scoped inside in the function but `found` (unused) is outside the function.

It'd be my preference to document return values and avoid named returns in this code but if this doesn't bother you, drop this comment.

rest/config.go
```diff
@@ -1747,56 +1756,59 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
...
-			base.DebugfCtx(ctx, base.KeyConfig, "unable to fetch config in group %q from bucket %q: %v", sc.Config.Bootstrap.ConfigGroupID, bucket, err)
-			return false, err
-		}
+func (sc *ServerContext) _fetchDatabaseFromBucket(ctx context.Context, bucket string, dbName string) (found bool, cnf DatabaseConfig, err error) {
 
-		if cnf.Name == "" {
-			cnf.Name = bucket
-		}
+	cas, err := sc.BootstrapContext.GetConfig(ctx, bucket, sc.Config.Bootstrap.ConfigGroupID, dbName, &cnf)
+	if err == base.ErrNotFound {
```
torcolvin: This is out of scope, but this is an error waiting to bite us, IMO.
```suggestion
	if errors.Is(err, base.ErrNotFound) {
```

rest/config.go
```diff
@@ -1747,56 +1756,59 @@ func (sc *ServerContext) fetchDatabase(ctx context.Context, dbName string) (foun
...
+		base.DebugfCtx(ctx, base.KeyConfig, "%q did not contain config in group %q", bucket, sc.Config.Bootstrap.ConfigGroupID)
+		return false, cnf, err
```
torcolvin, line 1764: This code is so wacky to me that this is basically an invalid config but we are still returning a config. 

As far as I can tell, the value of cnf is not used when `found=false` or `err != nil` value isn't used.



```mermaid
stateDiagram-v2
	handleGetDb : handleGetDb (error on not found or err)
	handleGetDbAuditConfig : handleGetDbAuditConfig (error on not found or err)
	handleGetCollectionConfigSync : handleGetCollectionConfigSync (error on not found or err)
	handleGetCollectionConfigImportFilter : handleGetCollectionConfigImportFilter (error on not found or err)
	fetchAndLoadConfigs : fetchAndLoadConfigs (cnf unused)
	_fetchAndLoadDatabase : _fetchAndLoadDatabase (cnf ignored on not found)
	
    handleGetDb --> fetchDatabase
    handleGetDbAuditConfig-->fetchDatabase
    handleGetCollectionConfigSync-->fetchDatabase
    handleGetCollectionConfingImportFilter-->fetchDatabase
    fetchDatabase-->_fetchDatabase
    fetchAndLoadConfigs-->_fetchDatabase
    _fetchAndLoadDatabase-->_fetchDatabase
```

torcolvin (commented):
Do you think we should put in test code that forces a timeout even in a clumsy way? We could have this enabled only with `-tag cb_sg_devmode`? I worry anything with manual testing is fragile.

base/error.go
adamcfraser: Good catch, replaced with IsTemporaryKvError (with some modifications)

rest/config.go
adamcfraser: Yes, we're only changing the behaviour when the error is (what we consider to be) transient. Other errors would be expected to still be present the next time the operation is performed, and indicate a situation where it's appropriate to remove the database.

Although _fetchDatabaseFromBucket is already logging the error details at debug, I agree that it would be useful to identify the specific error at info level - added that to the existing log message.

rest/config.go
adamcfraser: I stuck with the named returns as I prefer that method for self-documenting return values, but updated the variables in the callback to avoid shadowing.

rest/config.go
adamcfraser: Isn't this fairly standard for non-pointer return values - to return an empty value when there's an error?

torcolvin: I missed that this was a non pointer type, whoops.

torcolvin (approved):

PR #7038, opened 2024-08-05
CBG-4136: OpenAPI Specs: Limit db scopes to 1 and provide example

bbrks:
CBG-4136

Limit db scopes to 1 and provide example

Screenshot 2024-08-05 at 12 35 44
Screenshot 2024-08-05 at 12 35 48

- n/a

### DIFF
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -1203,7 +1203,7 @@ Database:
         $ref: '#/Scopes'
       maxProperties: 1
       example:
-        scopename1:
+        scopename:
           collections:
             collectionname1:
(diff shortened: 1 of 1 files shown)


docs/api/components/schemas.yaml
```diff
@@ -1201,6 +1201,16 @@ Database:
       additionalProperties:
         x-additionalPropertiesName: scopename
         $ref: '#/Scopes'
+      maxProperties: 1
+      example:
+        scopename1:
```
torcolvin: I might have this just be "scopename" since a part of confusion is that there can only be one.

```suggestion
        scopename:
```

torcolvin (approved):

PR #7040, opened 2024-08-05
CBG-4032: memory based rev cache implementation

gregns1:
CBG-4032

Implementation of memory bounded rev cache. 
Summary of changes: 
- Added stats for overall rev cache size in bytes 
- This is passed into the rev cache and atomically incremented/decremented as needed 
- Added a function on DocumentRev to calculate the document size based on what we agreed to be measured, channels, revisions and body in bytes.
- This is called when adding to rev cache for new docs, and after a load from the bucket 
- There is also a version of this for revision deltas 
- Added a representation of a documents size in DocumentRevision and rev cache value, for purposes of passing up and down stack. 
- Create a config value in `RevisionCacheOptions` to link up to RevCache code for the overall memory limit to be configured on. Temporarily this way, CBG-4134 will finish this side of the work. 
- Left number based eviction where it is. Added a new function for memory based eviction that is called separately after the value is populated with its size + incremented the stat to reflect this.
- Tests covering each operation on rev cache, eviction and delta caching. In addition have some test coverage of loading from the bucket. 
- For sharded revision cache memory limit, I have copied what we already fo for the number capacity (see `NewShardedLRURevisionCache`

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2664/

### DIFF
--- a/base/stats.go
+++ b/base/stats.go
@@ -468,6 +468,8 @@ type CacheStats struct {
 	RevisionCacheHits *SgwIntStat `json:"rev_cache_hits"`
 	// The total number of revision cache misses.
 	RevisionCacheMisses *SgwIntStat `json:"rev_cache_misses"`
+	// Total memory used by the rev cache
+	RevisionCacheTotalMemory *SgwIntStat `json:"revision_cache_total_memory"`
 	// The current length of the pending skipped sequence slice.
 	SkippedSeqLen *SgwIntStat `json:"skipped_seq_len"`
 	// The current capacity of the skipped sequence slice
@@ -1373,6 +1375,10 @@ func (d *DbStats) initCacheStats() error {
 	if err != nil {
 		return err
 	}
+	resUtil.RevisionCacheTotalMemory, err = NewIntStat(SubsystemCacheKey, "revision_cache_total_memory", StatUnitNoUnits, RevCacheMemoryDesc, StatAddedVersion3dot2dot1, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, labelKeys, labelVals, prometheus.CounterValue, 0)
+	if err != nil {
+		return err
+	}
 	resUtil.SkippedSeqLen, err = NewIntStat(SubsystemCacheKey, "skipped_seq_len", StatUnitNoUnits, SkippedSeqLengthDesc, StatAddedVersion3dot0dot0, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, labelKeys, labelVals, prometheus.GaugeValue, 0)
 	if err != nil {
 		return err
@@ -1422,6 +1428,7 @@ func (d *DbStats) unregisterCacheStats() {
 	prometheus.Unregister(d.CacheStats.RevisionCacheBypass)
 	prometheus.Unregister(d.CacheStats.RevisionCacheHits)
 	prometheus.Unregister(d.CacheStats.RevisionCacheMisses)
+	prometheus.Unregister(d.CacheStats.RevisionCacheTotalMemory)
 	prometheus.Unregister(d.CacheStats.SkippedSeqLen)
 	prometheus.Unregister(d.CacheStats.ViewQueries)
 }
--- a/base/stats.go
+++ b/base/stats.go
@@ -1375,7 +1375,7 @@ func (d *DbStats) initCacheStats() error {
 	if err != nil {
 		return err
 	}
-	resUtil.RevisionCacheTotalMemory, err = NewIntStat(SubsystemCacheKey, "revision_cache_total_memory", StatUnitNoUnits, RevCacheMemoryDesc, StatAddedVersion3dot2dot1, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, labelKeys, labelVals, prometheus.CounterValue, 0)
+	resUtil.RevisionCacheTotalMemory, err = NewIntStat(SubsystemCacheKey, "revision_cache_total_memory", StatUnitNoUnits, RevCacheMemoryDesc, StatAddedVersion3dot2dot1, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, labelKeys, labelVals, prometheus.GaugeValue, 0)
 	if err != nil {
 		return err
 	}
--- a/db/crud.go
+++ b/db/crud.go
@@ -722,7 +722,7 @@ func (db *DatabaseCollectionWithUser) get1xRevFromDoc(ctx context.Context, doc *
 	}
 
 	if listRevisions {
-		validatedHistory, getHistoryErr := doc.History.getHistory(revid)
+		validatedHistory, _, getHistoryErr := doc.History.getHistory(revid)
 		if getHistoryErr != nil {
 			return nil, removed, getHistoryErr
 		}
@@ -2143,7 +2143,7 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 
 	if doc.History[newRevID] != nil {
 		// Store the new revision in the cache
-		history, getHistoryErr := doc.History.getHistory(newRevID)
+		history, historyBytes, getHistoryErr := doc.History.getHistory(newRevID)
 		if getHistoryErr != nil {
 			return nil, "", getHistoryErr
 		}
@@ -2155,6 +2155,9 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 		}
 
 		revChannels := doc.History[newRevID].Channels
+
+		totalBytes := CalculateDocRevBytes(historyBytes, len(storedDocBytes), revChannels)
+
 		documentRevision := DocumentRevision{
 			DocID:       docid,
 			RevID:       newRevID,
@@ -2164,6 +2167,7 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 			Attachments: doc.Attachments,
 			Expiry:      doc.Expiry,
 			Deleted:     doc.History[newRevID].Deleted,
+			MemoryBytes: totalBytes,
 		}
 
 		if createNewRevIDSkipped {
--- a/db/crud.go
+++ b/db/crud.go
@@ -722,7 +722,7 @@ func (db *DatabaseCollectionWithUser) get1xRevFromDoc(ctx context.Context, doc *
 	}
 
 	if listRevisions {
-		validatedHistory, _, getHistoryErr := doc.History.getHistory(revid)
+		validatedHistory, getHistoryErr := doc.History.getHistory(revid)
 		if getHistoryErr != nil {
 			return nil, removed, getHistoryErr
 		}
@@ -2143,7 +2143,7 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 
 	if doc.History[newRevID] != nil {
 		// Store the new revision in the cache
-		history, historyBytes, getHistoryErr := doc.History.getHistory(newRevID)
+		history, getHistoryErr := doc.History.getHistory(newRevID)
 		if getHistoryErr != nil {
 			return nil, "", getHistoryErr
 		}
@@ -2155,9 +2155,6 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 		}
 
 		revChannels := doc.History[newRevID].Channels
-
-		totalBytes := CalculateDocRevBytes(historyBytes, len(storedDocBytes), revChannels)
-
 		documentRevision := DocumentRevision{
 			DocID:       docid,
 			RevID:       newRevID,
@@ -2167,7 +2164,6 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 			Attachments: doc.Attachments,
 			Expiry:      doc.Expiry,
 			Deleted:     doc.History[newRevID].Deleted,
-			MemoryBytes: totalBytes,
 		}
 
 		if createNewRevIDSkipped {
--- a/db/revision.go
+++ b/db/revision.go
@@ -153,14 +153,12 @@ func (revisions Revisions) parseAncestorRevisions(toAncestorRevID string) []stri
 		return nil
 	}
 	result := make([]string, 0)
-	var revHistoryBytes int
 
 	// Start at the parent, end at toAncestorRevID
 	start = start - 1
 	for i := 1; i < len(ids); i++ {
 		revID := fmt.Sprintf("%d-%s", start, ids[i])
 		result = append(result, revID)
-		revHistoryBytes += len([]byte(revID))
 		if revID == toAncestorRevID {
 			break
 		}
--- a/db/revision.go
+++ b/db/revision.go
@@ -147,11 +147,10 @@ func (revisions Revisions) ParseRevisions() []string {
 }
 
 // Returns revisions as a slice of ancestor revids, from the parent to the target ancestor.
-// Also returns measurement of revisions in bytes.
-func (revisions Revisions) parseAncestorRevisions(toAncestorRevID string) ([]string, int) {
+func (revisions Revisions) parseAncestorRevisions(toAncestorRevID string) []string {
 	start, ids := splitRevisionList(revisions)
 	if ids == nil || len(ids) < 2 {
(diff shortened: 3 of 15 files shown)


base/stats.go
```diff
@@ -431,6 +432,8 @@ type CacheStats struct {
 	RevisionCacheHits *SgwIntStat `json:"rev_cache_hits"`
 	// The total number of revision cache misses.
 	RevisionCacheMisses *SgwIntStat `json:"rev_cache_misses"`
+	// Total memory used by the rev cache
+	RevisionCacheTotalMemory *SgwIntStat `json:"revision_cache_total_memory"`
```
adamcfraser, line 472: I think a total count stat would still be useful, even when moving to a memory bounded cache.  Maybe file a separate ticket for that enhancement?

db/revtree.go
```diff
@@ -721,22 +721,26 @@ func (tree RevTree) RenderGraphvizDot() string {
 
 // Returns the history of a revid as an array of revids in reverse chronological order.
 // Returns error if detects cycle(s) in rev tree
-func (tree RevTree) getHistory(revid string) ([]string, error) {
+// Also returns measurement of revisions in bytes
```
adamcfraser:
```suggestion
// Also returns measurement of revision ID lengths in bytes
```

db/revtree.go
```diff
@@ -721,22 +721,26 @@ func (tree RevTree) RenderGraphvizDot() string {
...
 	maxHistory := len(tree)
 
+	var totalBytesForHistory int
 	history := make([]string, 0, 5)
 	for revid != "" {
 		info, err := tree.getInfo(revid)
 		if err != nil {
 			break
 		}
+		revBytes := len([]byte(revid))
```
adamcfraser, line 734: Is the []byte conversion needed?

db/crud.go
```diff
@@ -2142,7 +2142,7 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 
 	if doc.History[newRevID] != nil {
 		// Store the new revision in the cache
-		history, getHistoryErr := doc.History.getHistory(newRevID)
+		history, historyBytes, getHistoryErr := doc.History.getHistory(newRevID)
```
adamcfraser: Would it be more correct for encodeRevisions to compute the size of History, since the encoded revisions don't include rev generation (only digest), and that's what's stored in the rev cache History?

db/revision_cache_lru.go
```diff
@@ -403,12 +424,53 @@ func (value *revCacheValue) store(docRev DocumentRevision) {
...
+func CalculateDocRevBytes(historyBytes, bodyBytes int, channels base.Set) int64 {
+	var totalBytes int
+	for v := range channels {
+		bytes := len([]byte(v))
+		totalBytes += bytes
+	}
+	totalBytes += historyBytes
+	totalBytes += bodyBytes
+	// for the int64 size storage
+	totalBytes += 8
```
adamcfraser: Here I think we're accounting for DocumentRevision.MemoryBytes, but we don't account for several other DocumentRevision properties (docID, revID, expiry, attachments, deleted flag, removed flag).  Why do we do one and not the others?

db/revision_cache_lru.go
```diff
@@ -403,12 +424,53 @@ func (value *revCacheValue) store(docRev DocumentRevision) {
...
+		deltaBytes = value.delta.numDeltaBytes
+	}
 	value.delta = &toDelta
 	value.lock.Unlock()
+	return deltaBytes
+}
+
+// CalculateDocRevBytes will calculate the bytes from revisions in the document, body and channels on the document + a headroom
+// of 8 bytes for the storage of int64 for bytes measurement
+func CalculateDocRevBytes(historyBytes, bodyBytes int, channels base.Set) int64 {
```
adamcfraser: It feels unusual that this function has no connection to DocumentRevision (which is what's actually being stored in the rev cache).  I think this is partly because you're waiting to avoid a second iteration over history, is that right?  I agree with that motivation, but it does result in the memory computation being done even when not using the LRU cache, and also means the actual calculation is being performed in multiple places (here and getHistory), so it feels more likely that one of the spots could be missed by callers.

What do you think about adding a DocumentRevision.CalculateBytes() function that instead computes the history bytes as 32 x len(history[RevisionsIds]), since that's the standard rev digest size?  We've had a few cases in the past where CBL computed a shorter rev digest, but I think that's the upper limit.

db/revision_cache_lru.go
```diff
@@ -403,12 +424,53 @@ func (value *revCacheValue) store(docRev DocumentRevision) {
 		value.attachments = docRev.Attachments.ShallowCopy() // Don't store attachments the caller might later mutate
 		value.deleted = docRev.Deleted
 		value.err = nil
+		value.itemBytes = docRev.MemoryBytes
 	}
 	value.lock.Unlock()
 }
 
-func (value *revCacheValue) updateDelta(toDelta RevisionDelta) {
+func (value *revCacheValue) updateDelta(toDelta RevisionDelta) (deltaBytes int64) {
```
```suggestion
func (value *revCacheValue) updateDelta(toDelta RevisionDelta) (previousDeltaBytes int64) {
```

db/revision_cache_interface.go
```diff
@@ -165,6 +167,8 @@ type DocumentRevision struct {
 	Delta       *RevisionDelta
 	Deleted     bool
 	Removed     bool // True if the revision is a removal.
+
```
adamcfraser: Remove whitespace on line 170?

db/revision_cache_interface.go
```diff
@@ -165,6 +167,8 @@ type DocumentRevision struct {
...
+	MemoryBytes int64 // storage of the doc rev bytes measurement
```
adamcfraser: Does MemoryBytes include Delta.numDeltaBytes, when present?  It would be good to include that information in the comment, it's a bit cumbersome to try to trace through the code to find the answer.

db/revision_cache_interface.go
```diff
@@ -318,55 +322,62 @@ type RevisionDelta struct {
 	ToChannels            base.Set                // Full list of channels for the to revision
 	RevisionHistory       []string                // Revision history from parent of ToRevID to source revID, in descending order
 	ToDeleted             bool                    // Flag if ToRevID is a tombstone
+
```
adamcfraser: Remove whitespace, include comment describing numDeltaBytes, particularly how it differs from len(DeltaBytes)

db/revision_cache_interface.go
```diff
@@ -318,55 +322,62 @@ type RevisionDelta struct {
...
+	numDeltaBytes int64
```
adamcfraser: I think totalDeltaBytes might be a better name to differentiate from len(DeltaBytes).

db/revision_cache_lru.go
```diff
@@ -403,12 +424,53 @@ func (value *revCacheValue) store(docRev DocumentRevision) {
...
+	}
+	totalBytes += historyBytes
+	totalBytes += bodyBytes
+	// for the int64 size storage
+	totalBytes += 8
+	return int64(totalBytes)
+}
+
+// revCacheEvictionWithLock checks for rev cache eviction but first acquiring rev cache lock
+func (rc *LRURevisionCache) revCacheEvictionWithLock() {
```
adamcfraser: Having to acquire rc.Lock twice for every rev cache update feels like a non-trivial amount of additional contention.  Since rc.memoryBytes is atomic and has it's own locking, and rc.memoryCapacity should be immutable, it feels like we could do the rc.memoryBytes.Value() > rc.memoryCapacity without acquiring the lock, and only grab the lock when we actually need to evict.

db/revision_cache_lru.go
```diff
@@ -403,12 +424,53 @@ func (value *revCacheValue) store(docRev DocumentRevision) {
...
+	rc.lock.Lock()
+	defer rc.lock.Unlock()
+	rc._revCacheEviction()
+}
+
+// _revCacheEviction will check the current number capacity and memory capacity against maximum and memory capacity
+func (rc *LRURevisionCache) _revCacheEviction() {
```
adamcfraser: Given that there are different locking and timing requirements for capacity-based eviction and memory-based eviction, it feels like these should be left as two separate functions that can be called independently.

db/revision_cache_lru.go
```diff
@@ -286,6 +287,16 @@ func (rc *LRURevisionCache) Remove(docID, revID string, collectionID uint32) {
 		return
 	}
 	rc.lruList.Remove(element)
+	// decrement the overall memory bytes count
+	var totalBytesToRemove int64
+	revItem := element.Value.(*revCacheValue)
+	revItem.lock.RLock()
```
adamcfraser: Everything inside the lock here feels like it should be a standalone function on revCacheValue, to allow revCacheValue to manage it's own mutex.

adamcfraser (changes requested):
Overall approach seems ok, a few questions/suggestions.

base/stats.go
gregns1: I filed [CBG-4135](https://issues.couchbase.com/browse/CBG-4135) for that (started work on it yesterday, should have PR up today for it)

db/revision_cache_lru.go
gregns1: That's a good point i'll remove that. I think I was just accounting for the fact that I was inflating each rev cache item by 8 bytes but as we originally agreed we should be measuring channel,revisions and body it doesn't make sense.

db/revision_cache_lru.go
gregns1: That makes sense. This part of the PR was the but I had an issue with it did seem wasteful to acquire lock again to just check for eviction. To fix this i'll move the code that check for eviction by number of items back to where it originally was, then have eviction functions here to just check for memory based eviction and as you say only acquire lock if we need to evict here.

db/revision_cache_lru.go
gregns1: That works I've gone with your suggestion! You're right I was going for an approach to try piggy back off any iterations were already do to build revisions/channels to avoid perf overhead but this is a good approach.

db/revision_cache_lru.go
gregns1: Moved into its own function, this part of the code had changed a bit now. I made changes to have the memory count on the main body of the rev value include the delta size if its present for simplicity when it comes to removing items.

db/revision.go
```diff
@@ -153,12 +153,14 @@ func (revisions Revisions) parseAncestorRevisions(toAncestorRevID string) []stri
...
 	}
 	result := make([]string, 0)
+	var revHistoryBytes int
 
 	// Start at the parent, end at toAncestorRevID
 	start = start - 1
 	for i := 1; i < len(ids); i++ {
 		revID := fmt.Sprintf("%d-%s", start, ids[i])
 		result = append(result, revID)
+		revHistoryBytes += len([]byte(revID))
```
adamcfraser: It doesn't look like we're using this value?

db/revision_cache_interface.go
```diff
@@ -70,30 +70,32 @@ func NewRevisionCache(cacheOptions *RevisionCacheOptions, backingStores map[uint
...
 
-	return NewLRURevisionCache(cacheOptions.Size, backingStores, cacheHitStat, cacheMissStat)
+	return NewLRURevisionCache(cacheOptions.SizeByItems, cacheOptions.SizeInBytes, backingStores, cacheHitStat, cacheMissStat, memoryStat)
 }
 
 type RevisionCacheOptions struct {
-	Size       uint32
-	ShardCount uint16
+	SizeByItems uint32
+	SizeInBytes int64
```
adamcfraser: I think MaxCount and MaxBytes would be more intuitive naming for this.  (SizeByItems and SizeInBytes sounds like current size, not maximum)

db/revision_cache_interface.go
```diff
@@ -70,30 +70,32 @@ func NewRevisionCache(cacheOptions *RevisionCacheOptions, backingStores map[uint
...
 	cacheMissStat := cacheStats.RevisionCacheMisses
+	memoryStat := cacheStats.RevisionCacheTotalMemory
 
 	if cacheOptions.ShardCount > 1 {
-		return NewShardedLRURevisionCache(cacheOptions.ShardCount, cacheOptions.Size, backingStores, cacheHitStat, cacheMissStat)
+		return NewShardedLRURevisionCache(cacheOptions.ShardCount, cacheOptions.SizeByItems, cacheOptions.SizeInBytes, backingStores, cacheHitStat, cacheMissStat, memoryStat)
 	}
 
-	return NewLRURevisionCache(cacheOptions.Size, backingStores, cacheHitStat, cacheMissStat)
+	return NewLRURevisionCache(cacheOptions.SizeByItems, cacheOptions.SizeInBytes, backingStores, cacheHitStat, cacheMissStat, memoryStat)
```
adamcfraser: Refactoring to just pass cacheOptions here and on line 83 seems like it might be preferable.

db/revision_cache_lru.go
```diff
@@ -403,12 +438,82 @@ func (value *revCacheValue) store(docRev DocumentRevision) {
...
+	totalBytes += len(delta.DeltaBytes)
+
+	delta.totalDeltaBytes = int64(totalBytes)
+}
+
+// revCacheMemoryBasedEviction checks for rev cache eviction, if required calls performEviction which will acquire lock to evict
+func (rc *LRURevisionCache) revCacheMemoryBasedEviction() {
+	// if memory capacity is not set, don't check for eviction this way
+	if rc.memoryCapacity > 0 && rc.memoryBytes.Value() > rc.memoryCapacity {
+		rc.performEviction()
```
adamcfraser, line 533: This is only going to evict a single value, and there's no guarantee that the evicted value is large enough to offset the incoming value - could end up in a situation where 100 20MB documents trigger the eviction of 100 1k documents, and leave the cache significantly above the memory target.

db/revision_cache_lru.go
```diff
@@ -359,6 +388,7 @@ func (value *revCacheValue) asDocumentRevision(delta *RevisionDelta) (DocumentRe
 		Attachments: value.attachments.ShallowCopy(), // Avoid caller mutating the stored attachments
 		Deleted:     value.deleted,
 		Removed:     value.removed,
+		MemoryBytes: value.itemBytes,
```
adamcfraser: asDocumentRevision is only intended to copy immutable properties (and so doesn't require callers to hold value.lock.Lock).  With the updateDelta handling I don't think itemBytes can be treated as immutable, but it's also important from a performance perspective to not hold the lock.

db/revision_cache_lru.go
```diff
@@ -239,17 +239,32 @@ func (rc *LRURevisionCache) Upsert(ctx context.Context, docRev DocumentRevision,
 	rc.lock.Lock()
 	// If element exists remove from lrulist
 	if elem := rc.cache[key]; elem != nil {
+		revItem := elem.Value.(*revCacheValue)
+		// decrement item bytes by the removed item
+		rc.memoryBytes.Add(-revItem.itemBytes)
```
adamcfraser: I think this read of itemBytes would need to be synchronized.

db/revision_cache_lru.go
adamcfraser: One option might be to separate the delta and non-delta byte storage on revCacheValue, and compute the sum for DocumentRevision.MemoryBytes here.

adamcfraser (changes requested):

db/revision_cache_lru.go
gregns1: I'm not sure I follow. This function is to determine if we need eviction based upon if we have a max size set & the current memory count is above the threshold. 
If true it will call `performEviction()` which acquires the lock and runs a loop for eviction till the memory stat is below the threshold.

db/revision_cache_lru.go
gregns1: Good spot. I have ended up removing this all together as I can work around this by uisng the `getItemBytes()` function that acquired read lock to incr/decr the stat wherever necessary.

db/revision_cache_lru.go
adamcfraser: This was my mistake - misread performEviction as 'if' instead of 'for'.

adamcfraser (approved):
I still feel like there is a lot of unnecessary locking of itemBytes in the case where delta sync is not being used (and so itemBytes is immutable), but I'm fine coming back to that based on analysis of the perf results.

PR #7045, opened 2024-08-06
CBG-4132 wait for sequence id, not number of sequences

torcolvin:
This is from a recently re-enabled test and addresses CBG-4141.


gregns1 (approved):
LGTM

PR #7051, opened 2024-08-09
CBG-4161 log the source of metadata ID

torcolvin:
I implemented this with a subfunction rather than declaring `var logMessage` at the top of the function and using a defer because I thought it would prevent us from missing a place that we log by requiring a specific return. The only weird part is that we'll double log, once on WRN and once on DBG, if we have an error other than `IsDocNotFound` getting `_sync:syncInfo`.

I am not worried about this function getting called on each config because it is only called when creating a new db, or when main or initial bootstrap polling finds a legacy config.

```mermaid
flowchart TD
    InsertConfig --> computeMetadataID
    ComputeMetadataIDForConfig --> computeMetadataID
    handleCreateDb --> ComputeMetadataIDForConfig
    main --> automaticConfigUpgrade
    automaticConfigUpgrade --> InsertConfig
    handleCreateDb --> computeMetadataID
    migrateV30Configs --> InsertConfig
    initializeBootstrapConnection --> migrateV30Configs
```

## Pre-review checklist

### DIFF
--- a/rest/config_manager.go
+++ b/rest/config_manager.go
@@ -761,56 +761,56 @@ func (b *bootstrapContext) ComputeMetadataIDForDbConfig(ctx context.Context, con
 //  4. The _default._default collection isn't already associated with a different metadata ID (syncInfo document is not present, or has a value of defaultMetadataID)
 func (b *bootstrapContext) computeMetadataID(ctx context.Context, registry *GatewayRegistry, config *DbConfig) string {
 
-	standardMetadataID := b.standardMetadataID(config.Name)
+	metadataID, source := func() (string, string) {
+		standardMetadataID := b.standardMetadataID(config.Name)
 
-	// If there's already a metadataID assigned to this database in the registry (including other config groups), use that
-	defaultMetadataIDInUse := false
-	for _, cg := range registry.ConfigGroups {
-		for dbName, db := range cg.Databases {
-			if dbName == config.Name {
-				return db.MetadataID
-			}
-			if db.MetadataID == defaultMetadataID {
-				defaultMetadataIDInUse = true
-			}
-		}
-	}
+		// If there's already a metadataID assigned to this database in the registry (including other config groups), use that
+		for _, cg := range registry.ConfigGroups {
+			for dbName, db := range cg.Databases {
+				if dbName == config.Name {
+					return db.MetadataID, fmt.Sprintf("Using metadata ID=%q from registry for db %q", db.MetadataID, dbName)
+				}
+				if db.MetadataID == defaultMetadataID {
+					// If the default metadata ID is already in use in the registry by a different database, use standard ID.
+					return standardMetadataID, fmt.Sprintf("Using metadata ID %q for db %q since db %q is using the default metadata ID", standardMetadataID, config.Name, dbName)
 
-	// If the default metadata ID is already in use in the registry by a different database, use standard ID.
-	if defaultMetadataIDInUse {
-		return standardMetadataID
-	}
-
-	// If the database config doesn't include _default._default, use standard ID
-	if config.Scopes != nil {
-		defaultFound := false
-		for scopeName, scope := range config.Scopes {
-			for collectionName := range scope.Collections {
-				if base.IsDefaultCollection(scopeName, collectionName) {
-					defaultFound = true
 				}
 			}
 		}
-		if !defaultFound {
-			return standardMetadataID
+
+		// If the database config doesn't include _default._default, use standard ID
+		if config.Scopes != nil {
+			defaultFound := false
+			for scopeName, scope := range config.Scopes {
+				for collectionName := range scope.Collections {
+					if base.IsDefaultCollection(scopeName, collectionName) {
+						defaultFound = true
+					}
+				}
+			}
+			if !defaultFound {
+				return standardMetadataID, fmt.Sprintf("Using metadata ID %q for db %q since _default._default collection is not a collection targeted by this db", standardMetadataID, config.Name)
+			}
 		}
-	}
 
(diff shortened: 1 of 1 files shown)


rest/config_manager.go
```diff
@@ -761,56 +761,56 @@ func (b *bootstrapContext) ComputeMetadataIDForDbConfig(ctx context.Context, con
...
+
+		// If there's already a metadataID assigned to this database in the registry (including other config groups), use that
+		for _, cg := range registry.ConfigGroups {
+			for dbName, db := range cg.Databases {
+				if dbName == config.Name {
+					return db.MetadataID, fmt.Sprintf("Using metadata ID=%q from registry for db %q", db.MetadataID, dbName)
+				}
+				if db.MetadataID == defaultMetadataID {
+					// If the default metadata ID is already in use in the registry by a different database, use standard ID.
+					return standardMetadataID, fmt.Sprintf("Using metadata ID %q for db %q since db %q is using the default metadata ID", standardMetadataID, config.Name, dbName)
```
adamcfraser: This changes the previous behaviour in the case where:
- the default metadata ID is in use by a different db
- there is already a metadataID assigned to this database, and it's not the standardMetadataID

This is the reason for the previous defaultMetadataIDInUse approach, and I don't think this PR should change that functional behaviour.

rest/config_manager.go
```diff
@@ -761,56 +761,56 @@ func (b *bootstrapContext) ComputeMetadataIDForDbConfig(ctx context.Context, con
...
+		}
 
-	return defaultMetadataID
+		if exists && syncInfo.MetadataID != defaultMetadataID {
+			return standardMetadataID, fmt.Sprintf("Using metadata ID %q for db %q since base.SGSyncInfo in _default._default specifies metadata ID %q", standardMetadataID, config.Name, syncInfo.MetadataID)
+		}
 
+		return defaultMetadataID, fmt.Sprintf("Using default metadata ID %q for db %q", defaultMetadataID, config.Name)
+	}()
+	base.DebugfCtx(ctx, base.KeyConfig, source)
```
adamcfraser: Given the infrequency of this execution (as you mention in the PR description), logging at Info seems appropriate.

rest/config_manager.go
```diff
@@ -761,56 +761,56 @@ func (b *bootstrapContext) ComputeMetadataIDForDbConfig(ctx context.Context, con
...
-	// If there's already a metadataID assigned to this database in the registry (including other config groups), use that
-	defaultMetadataIDInUse := false
-	for _, cg := range registry.ConfigGroups {
-		for dbName, db := range cg.Databases {
-			if dbName == config.Name {
-				return db.MetadataID
-			}
-			if db.MetadataID == defaultMetadataID {
-				defaultMetadataIDInUse = true
+	metadataID, source := func() (string, string) {
```
adamcfraser: We generally should be using base.MD() redaction for dbnames (and should probably for metadataID as well).  What about changing this function to return `metadataID, dbName, sourceMsg`, and then change the log line (812) to something like
```
base.InfofCtx(ctx, "Using metadata ID %v for db %v, reason: %v, base.MD(metadataID), base.MD(dbName), msg)
```
We would lose the ability to log additional metadataID or dbNames with this approach, which we do in two places, though.   In general this makes me wonder whether the anonymous function is really the right approach, or whether we should just do the logging everywhere we return and be diligent in future when changing this function.

rest/config_manager.go
```diff
@@ -761,56 +761,56 @@ func (b *bootstrapContext) ComputeMetadataIDForDbConfig(ctx context.Context, con
...
+		exists, err := b.Connection.GetDocument(ctx, bucketName, base.SGSyncInfo, &syncInfo)
+		if err != nil {
+			msg := fmt.Sprintf("Error checking syncInfo metadataID in default collection - using standard metadataID.  Error: %v", err)
+			base.WarnfCtx(ctx, msg)
+			return standardMetadataID, msg
+		}
 
-	return defaultMetadataID
+		if exists && syncInfo.MetadataID != defaultMetadataID {
+			return standardMetadataID, fmt.Sprintf("Using metadata ID %q for db %q since base.SGSyncInfo in _default._default specifies metadata ID %q", standardMetadataID, config.Name, syncInfo.MetadataID)
```
adamcfraser:
```suggestion
			return standardMetadataID, fmt.Sprintf("Using metadata ID %q for db %q because db uses the default collection, and _sync:syncInfo in the default collection specifies the non-default metadata ID %q", standardMetadataID, config.Name, syncInfo.MetadataID)
```

adamcfraser (dismissed):

rest/config_manager.go
```diff
@@ -780,7 +782,6 @@ func (b *bootstrapContext) computeMetadataID(ctx context.Context, registry *Gate
 	if defaultMetadataIDInUse {
 		return standardMetadataID
```
bbrks, line 784: Missing log for the commented case above.

rest/config_manager.go
```diff
@@ -768,9 +768,11 @@ func (b *bootstrapContext) computeMetadataID(ctx context.Context, registry *Gate
 	for _, cg := range registry.ConfigGroups {
 		for dbName, db := range cg.Databases {
 			if dbName == config.Name {
+				base.InfofCtx(ctx, base.KeyConfig, "Using metadata ID=%q from registry for db %q", base.MD(db.MetadataID), base.MD(dbName))
 				return db.MetadataID
 			}
 			if db.MetadataID == defaultMetadataID {
+				// do not return standardMetadataID here in case there was a different metadata ID assigned to this database
```
bbrks, line 775: Might be worth a (debug?) log here even though we're not returning, so we can log _which_ database has the default metadata ID.

bbrks (approved):

PR #7052, opened 2024-08-09
CBG-4160: Replace base.WaitForStat with Assert/Require and fix missing stat incr

bbrks:
CBG-4160

- Replace `base.WaitForStat` with `require.EventuallyWithT` variants due to misuse and silently failing tests.
- Add missing `HandleRevDocsPurgedCount` stat increment in non-rev removal case (caught by one of the silently failing tests)
- Fix silently failing assertion in `TestActiveReplicatorPushAndPullConflict` which didn't take into account conflict count, and caused flakyness in the test

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2662/

### DIFF
--- a/base/util_testing.go
+++ b/base/util_testing.go
@@ -736,22 +736,33 @@ func DirExists(filename string) bool {
 	return info.IsDir()
 }
 
-// WaitForStat will retry for up to 20 seconds until the result of getStatFunc is equal to the expected value.
-func WaitForStat(t testing.TB, getStatFunc func() int64, expected int64) (int64, bool) {
+// _waitForStat is the internal implementation of AssertWaitForStat and RequireWaitForStat. Use those instead.
+func _waitForStat(t testing.TB, getStatFunc func() int64, expected int64) (int64, bool) {
 	workerFunc := func() (shouldRetry bool, err error, val interface{}) {
 		val = getStatFunc()
 		return val != expected, nil, val
 	}
 	// wait for up to 20 seconds for the stat to meet the expected value
 	err, val := RetryLoop(TestCtx(t), "waitForStat retry loop", workerFunc, CreateSleeperFunc(200, 100))
-	valInt64, ok := val.(int64)
+	require.NoError(t, err)
 
-	return valInt64, err == nil && ok
+	valInt64, ok := val.(int64)
+	require.True(t, ok)
+
+	return valInt64, expected == valInt64
+}
+
+// AssertWaitForStat will retry for up to 20 seconds until the result of getStatFunc is equal to the expected value.
+func AssertWaitForStat(t testing.TB, getStatFunc func() int64, expected int64) (val int64) {
+	val, ok := _waitForStat(t, getStatFunc, expected)
+	require.True(t, ok)
+	assert.Equal(t, expected, val)
+	return val
 }
 
 // RequireWaitForStat will retry for up to 20 seconds until the result of getStatFunc is equal to the expected value.
 func RequireWaitForStat(t testing.TB, getStatFunc func() int64, expected int64) {
-	val, ok := WaitForStat(t, getStatFunc, expected)
+	val, ok := _waitForStat(t, getStatFunc, expected)
 	require.True(t, ok)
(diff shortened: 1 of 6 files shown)


base/util_testing.go
```diff
@@ -736,22 +736,33 @@ func DirExists(filename string) bool {
...
 
-	return valInt64, err == nil && ok
+	return valInt64, expected == valInt64
+}
+
+// AssertWaitForStat will retry for up to 20 seconds until the result of getStatFunc is equal to the expected value.
+func AssertWaitForStat(t testing.TB, getStatFunc func() int64, expected int64) (val int64) {
+	val, ok := _waitForStat(t, getStatFunc, expected)
+	require.True(t, ok)
+	assert.Equal(t, expected, val)
```
torcolvin: `require.True(t, ok)` is going to fail if `expected != val`, so by putting this statement here it both actions as a require and doesn't print the expected vs actual values.

I think this should drop require.True(t, ok) and _waitForStat should just return the last value of the stat after 20s

base/util_testing.go
```diff
@@ -736,22 +736,33 @@ func DirExists(filename string) bool {
...
+	return val
 }
 
 // RequireWaitForStat will retry for up to 20 seconds until the result of getStatFunc is equal to the expected value.
 func RequireWaitForStat(t testing.TB, getStatFunc func() int64, expected int64) {
-	val, ok := WaitForStat(t, getStatFunc, expected)
+	val, ok := _waitForStat(t, getStatFunc, expected)
 	require.True(t, ok)
```
torcolvin: As above, I think this should drop the `ok` and just `require.Equal(t, expected, val)` so that we get the two possible values

torcolvin (commented):
I wonder if in generally this would easier to read if these were replaced with `require.EventuallyWithT` so you could easily see the code that does the assertion when it fails.

IMO some of these are only loosely stats, like `val := ar.GetStatus(ctx1).DocsWritten + ar.GetStatus(ctx1).DocWriteConflict`

rest/persistent_config_test.go
```diff
@@ -1429,7 +1429,7 @@ func TestPersistentConfigNoBucketField(t *testing.T) {
 	count, err = rt.ServerContext().fetchAndLoadConfigs(base.TestCtx(t), false)
 	require.NoError(t, err)
 	assert.Equal(t, 0, count)
-	dbBucketMismatch, _ = base.WaitForStat(t, base.SyncGatewayStats.GlobalStats.ConfigStat.DatabaseBucketMismatches.Value, dbBucketMismatch+1)
+	dbBucketMismatch = base.AssertWaitForStat(t, base.SyncGatewayStats.GlobalStats.ConfigStat.DatabaseBucketMismatches.Value, dbBucketMismatch+1)
```
torcolvin: I think this code should actually probably use RequireWaitForStat because once it is in a bad state here, continuing the test is just going to produce meaningless errors

rest/replicatortest/replicator_test.go
```diff
@@ -4501,8 +4499,9 @@ func TestActiveReplicatorPushAndPullConflict(t *testing.T) {
 			// Start the replicator (implicit connect)
 			assert.NoError(t, ar.Start(ctx1))
 			// wait for the document originally written to rt2 to arrive at rt1.  Should end up as winner under default conflict resolution
-			base.WaitForStat(t, func() int64 {
-				return ar.GetStatus(ctx1).DocsWritten
+			base.RequireWaitForStat(t, func() int64 {
+				val := ar.GetStatus(ctx1).DocsWritten + ar.GetStatus(ctx1).DocWriteConflict
```
torcolvin: Did you mean to change the return statement here to add `DocWriteConflict`?

bbrks: Yes, although I've improved it with the latest commit. It's a source of flakyness in this test where `DocsWritten` isn't enough to ensure the replication has completed.

torcolvin (dismissed):

torcolvin (approved):

PR #7058, opened 2024-08-12
CBG-3933 allow console logs to be rotated

torcolvin:
- create `initLumberjackLogger` to initialize a lumberjack logging for files
- fix a latent bug where we leaked `logCollationWorker` since we would never exit the go routine. In order to close this gap, I added `FileLogger.closed` channel, since the following events have to happen in order:
	1. exit log rotation worker goroutine
	2. wait for log rotation worker to be done
	3. close the file handle
	4. exit logCollationWorker goroutine
- fixup docs for logging config. I noticed that https://docs.couchbase.com/sync-gateway/current/configuration-schema-bootstrap.html is missing a lot of the default values, so I added them in. Use `allOf` to allow for composition of data structures to override the max age parameters.

Notes:

- I don't know that `max_age=6` is right for the stats logger. It probably doesn't matter since we never hit it.
- I kept the console logger `max_age=0` because I wanted to conform to the API. That said, as above our API isn't documented. I don't know what it worse, losing console logs or filing up disks.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2648/


base/logger_console.go
```diff
@@ -67,23 +69,27 @@ func NewConsoleLogger(ctx context.Context, shouldLogLocation bool, config *Conso
...
+			closed:           make(chan struct{}),
 		},
 		isStderr: config.FileOutput == "" && *config.Enabled,
 		config:   *config,
 	}
 	logger.Enabled.Set(*config.Enabled)
 
 	// Only create the collateBuffer channel and worker if required.
-	if *config.CollationBufferSize > 1 {
+	if *config.CollationBufferSize > 1 && !logger.isStderr {
```
bbrks: Why are we now disabling log collation for stderr output? Doesn't seem intentional, given we set a collation size of 10 for a non-file console logger.

base/logger_file.go
```diff
@@ -244,9 +245,25 @@ func (lfc *FileLoggerConfig) init(ctx context.Context, level LogLevel, name stri
...
+// initLumberjack will create a new Lumberjack logger from the given config settings. Returns a doneChan which fires when the log rotation is stopped.
+func (lfc *FileLoggerConfig) initLumberjack(ctx context.Context, name string, lumberjackFilename string) chan struct{} {
+	rotationDoneChan := make(chan struct{})
+	dir, path := filepath.Split(lumberjackFilename)
+	prefix := path[0:strings.Index(path, ".")]
+	rotateableLogger := &lumberjack.Logger{
+		Filename: lumberjackFilename,
+		MaxSize:  *lfc.Rotation.MaxSize,
+		MaxAge:   *lfc.Rotation.MaxAge,
+		Compress: BoolDefault(lfc.Rotation.Compress, true),
```
bbrks: This is a change in behaviour for the console logger, which disabled compression.

We should probably explicitly set `lfc.Rotation.Compress` in `ConsoleLoggerConfig.init` to keep that behaviour.

base/logger_file.go
```diff
@@ -244,9 +245,25 @@ func (lfc *FileLoggerConfig) init(ctx context.Context, level LogLevel, name stri
...
 	}
 
+	return rotationDoneChan, nil
+}
+
+// initLumberjack will create a new Lumberjack logger from the given config settings. Returns a doneChan which fires when the log rotation is stopped.
+func (lfc *FileLoggerConfig) initLumberjack(ctx context.Context, name string, lumberjackFilename string) chan struct{} {
+	rotationDoneChan := make(chan struct{})
+	dir, path := filepath.Split(lumberjackFilename)
+	prefix := path[0:strings.Index(path, ".")]
```
bbrks: This assumes your console log filename always has a dot in it, which isn't necessarily true.

I'm not actually sure this prefix stuff is nessesary for the cleanup logic to work. It should work OK without as far as I can tell.

```json
      "file_output": "/tmp/sglogs-2024-08-13/console/sglog",
```

```go
2024-08-13T13:52:37.767+01:00 [ERR] Unexpected panic: runtime error: slice bounds out of range [:-1] - stopping process
goroutine 1 [running]:
runtime/debug.Stack()
	/opt/homebrew/Cellar/go/1.22.4/libexec/src/runtime/debug/stack.go:24 +0x64
github.com/couchbase/sync_gateway/base.FatalPanicHandler()
	/Users/benbrooks/dev/cb/sg/base/util.go:1458 +0x34
panic({0x1017ef980?, 0x140005ec4e0?})
	/opt/homebrew/Cellar/go/1.22.4/libexec/src/runtime/panic.go:770 +0x124
github.com/couchbase/sync_gateway/base.(*FileLoggerConfig).initLumberjack(0x14000690510, {0x1018d20d8, 0x14000797900}, {0x100f42b3f, 0x7}, {0x140006f9080, 0x24})
	/Users/benbrooks/dev/cb/sg/base/logger_file.go:255 +0x2d4
github.com/couchbase/sync_gateway/base.(*ConsoleLoggerConfig).init(0x14000690510, {0x1018d20d8, 0x14000797900})
	/Users/benbrooks/dev/cb/sg/base/logger_console.go:206 +0xe0
github.com/couchbase/sync_gateway/base.NewConsoleLogger({0x1018d2068, 0x102479540}, 0x1, 0x14000580008?)
	/Users/benbrooks/dev/cb/sg/base/logger_console.go:59 +0x7c
github.com/couchbase/sync_gateway/base.InitLogging({0x1018d2068, 0x102479540}, {0x1400055f218, 0x16}, 0x1024117c0?, 0x14000796690, 0x140007966e0, 0x14000796730, 0x14000796af0, 0x140007967d0, ...)
	/Users/benbrooks/dev/cb/sg/base/logging_config.go:66 +0x54
github.com/couchbase/sync_gateway/rest.(*StartupConfig).SetupAndValidateLogging(0x1400069e788, {0x1018d2068, 0x102479540})
	/Users/benbrooks/dev/cb/sg/rest/config.go:1387 +0x1f8
github.com/couchbase/sync_gateway/rest.SetupServerContext({0x1018d2068, 0x102479540}, 0x1400069e788, 0x1)
	/Users/benbrooks/dev/cb/sg/rest/config.go:1541 +0xac
github.com/couchbase/sync_gateway/rest.serverMainPersistentConfig({0x1018d2068, 0x102479540}, 0x140007924d0, 0x1400069e288)
	/Users/benbrooks/dev/cb/sg/rest/main.go:146 +0x404
github.com/couchbase/sync_gateway/rest.serverMain({0x1018d2068, 0x102479540}, {0x1400014a000, 0x2, 0x2})
	/Users/benbrooks/dev/cb/sg/rest/main.go:59 +0x118
github.com/couchbase/sync_gateway/rest.ServerMain()
	/Users/benbrooks/dev/cb/sg/rest/main.go:33 +0x44
main.main()
	/Users/benbrooks/dev/cb/sg/main.go:17 +0x1c
 -- base.FatalfCtx() at logging.go:151
```

base/logger_file.go
```diff
@@ -352,7 +359,7 @@ func runLogDeletion(ctx context.Context, logDirectory string, logLevel string, s
 	willDelete := false
 	for i := len(files) - 1; i >= 0; i-- {
 		file := files[i]
-		if strings.HasPrefix(file.Name(), logFilePrefix+logLevel) && strings.HasSuffix(file.Name(), ".log.gz") {
+		if strings.HasPrefix(file.Name(), logPrefix) && strings.HasSuffix(file.Name(), ".gz") {
```
bbrks: This code still assumes all rotated files get compressed, which wasn't true for console logs.

I think we'll be better off making the `.gz` extention optional for this entire function, because it's also possible that customers have extracted the files on-disk at some point, and we'd probably want to clean those up whilst we're here.

base/logger_file.go
```diff
@@ -244,9 +245,25 @@ func (lfc *FileLoggerConfig) init(ctx context.Context, level LogLevel, name stri
 		lfc.CollationBufferSize = &bufferSize
 	}
 
+	return rotationDoneChan, nil
+}
+
+// initLumberjack will create a new Lumberjack logger from the given config settings. Returns a doneChan which fires when the log rotation is stopped.
+func (lfc *FileLoggerConfig) initLumberjack(ctx context.Context, name string, lumberjackFilename string) chan struct{} {
+	rotationDoneChan := make(chan struct{})
+	dir, path := filepath.Split(lumberjackFilename)
```
bbrks:
```suggestion
	dir, file := filepath.Split(lumberjackFilename)
```

bbrks (changes requested):
Haven't reviewed the API doc changes yet, but have done a full pass on the code changes.

base/logger_console.go
torcolvin: I was thinking collation is less valuable if we weren't writing to an on disk file. I changed this back.

In an intermediate version of this PR, running the `logCollationWorker` without log rotation would hang, due to the channel interactions but I've addressed that problem.

base/logger_file.go
torcolvin: I basically agree with this - but I thought of a bunch of edge cases which I added tests for. This become really hard to test for without regexp, tests are `TestConsoleIrregularLogPaths`.

Maybe this is just me, but I worry about naming something `log` for temporary debugging, and then in the old way this code worked, anything starting with `log` in the directory would get deleted, which wouldn't necessarily be the way you'd want it. Now it should delete things:

example: `/path/to/foo.log`

1. foo.log
2. foo-2016-11-04T18-30-00.000.log
3. foo-2016-11-04T18-30-00.000.log.gz

The other edge case if you name the file: `.consoleLog` with a leading `.` and no other extension. This was what tipped the point to using a regex since you could no longer rely on a prefix match since there was none for rotated logs.

base/logger_file.go
torcolvin: See below on why I changed this match to a regex.

base/logger_file.go
```diff
@@ -384,3 +396,27 @@ func runLogDeletion(ctx context.Context, logDirectory string, logLevel string, s
...
+
+// getDeletionDirAndRegexp will return the directory and a regexp matching log file and rotated patterns.
+func getDeletionDirAndRegexp(filename string) (string, *regexp.Regexp) {
+	dir, path := filepath.Split(filename)
+	lastDot := strings.LastIndex(path, ".")
+	if lastDot == -1 {
+		// foo
+		// foo-2019-01-01T00-00-00.000
+		// foo-2019-01-01T00-00-00.000.gz
+		return dir, regexp.MustCompile(fmt.Sprintf(`(%s|%s%s(?:\.gz))?$`, regexp.QuoteMeta(path), regexp.QuoteMeta(path), lumberjackRotationMidfix))
```
bbrks: Typo, the other pattern is correct.

```suggestion
		return dir, regexp.MustCompile(fmt.Sprintf(`(%s|%s%s(?:\.gz)?)$`, regexp.QuoteMeta(path), regexp.QuoteMeta(path), lumberjackRotationMidfix))
```

base/logger_file.go
```diff
@@ -384,3 +396,27 @@ func runLogDeletion(ctx context.Context, logDirectory string, logLevel string, s
 
 	return nil
 }
+
+// getDeletionDirAndRegexp will return the directory and a regexp matching log file and rotated patterns.
+func getDeletionDirAndRegexp(filename string) (string, *regexp.Regexp) {
+	dir, path := filepath.Split(filename)
+	lastDot := strings.LastIndex(path, ".")
```
bbrks: `path` and `filename` are the wrong way around from what I'd expect

```suggestion
func getDeletionDirAndRegexp(path string) (string, *regexp.Regexp) {
	dir, name := filepath.Split(path)
	lastDot := strings.LastIndex(name, ".")
```

base/logger_console_test.go
```diff
@@ -287,3 +291,92 @@ func TestConsoleLogDefaults(t *testing.T) {
...
+			var filenames []string
+			require.EventuallyWithT(t, func(c *assert.CollectT) {
+				filenames = getDirFiles(t, tempdir)
+				assert.Contains(c, filenames, test.logPath)
+				assert.Greater(c, len(filenames), 2)
+			}, time.Second, 10*time.Millisecond)
+			_, pattern := getDeletionDirAndRegexp(filepath.Join(tempdir, test.logPath))
+			for _, filename := range filenames {
+				require.Regexp(t, pattern, filename)
+			}
```
bbrks: negative testing for patterns that are too greedy

```suggestion
			require.EventuallyWithT(t, func(c *assert.CollectT) {
				filenames := getDirFiles(t, tempdir)
				assert.Contains(c, filenames, test.logPath)
				assert.Greater(c, len(filenames), 2)
			}, time.Second, 10*time.Millisecond)

			// add a few non-matching files to the directory for negative testing
			nonMatchingFileNames := []string{
				"console.log",
				"consoellog",
				"console.log.txt",
				".console",
				"consolelog-2021-01-01T00-00-00.000",
				"console-2021-01-01T00-00-00.000.log",
			}
			for _, name := range nonMatchingFileNames {
				require.NoError(t, makeTestFile(1, name, tempdir))
			}

			_, pattern := getDeletionDirAndRegexp(filepath.Join(tempdir, test.logPath))
			for _, filename := range getDirFiles(t, tempdir) {
				if slices.Contains(nonMatchingFileNames, filename) {
					require.NotRegexp(t, pattern, filename)
				} else {
					require.Regexp(t, pattern, filename)
				}
			}
```

base/logger_file.go
```diff
@@ -384,3 +396,27 @@ func runLogDeletion(ctx context.Context, logDirectory string, logLevel string, s
...
+	// foo.log
+	// foo-2019-01-01T00-00-00.000.log
+	// foo-2019-01-01T00-00-00.000.log.gz
+	//
+	// or
+	//
+	// foo.bar.log
+	// foo.bar-2019-01-01T00-00-00.000.log
+	// foo.bar-2019-01-01T00-00-00.000.log.gz
+	return dir, regexp.MustCompile(fmt.Sprintf(`(%s|%s%s%s(?:\.gz)?)$`, regexp.QuoteMeta(path), regexp.QuoteMeta(prefix), lumberjackRotationMidfix, regexp.QuoteMeta(suffix)))
```
bbrks: I'm wondering whether we really need the non-capturing group, given the caller doesn't care about subgroups at all.

It just makes the pattern harder to read and makes me wonder whether groups are relevant for the caller.

I think we could also include `^` to make it more obvious that this pattern is only for the file name, and not a full file path.

```suggestion
	return dir, regexp.MustCompile(fmt.Sprintf(`^(%s|%s%s%s(\.gz)?)$`, regexp.QuoteMeta(path), regexp.QuoteMeta(prefix), lumberjackRotationMidfix, regexp.QuoteMeta(suffix)))
```

bbrks: Took me a while to separate the Sprintf and regexp syntax and I think it would be slightly more readable without the mix, especially if we do `regexp.QuoteMeta` outside of the pattern building, and build up the rotated pattern before passing it into the larger pattern.

Doing this allows us to use the same pattern for both formats, by having `rotatedPattern` built for the LastIndex case

```suggestion
	filenamePattern := regexp.QuoteMeta(filename)
	rotatedPattern := regexp.QuoteMeta(prefix) + lumberjackRotationMidfix + regexp.QuoteMeta(suffix) + `(\.gz)?`
	return dir, regexp.MustCompile(`^(` + filenamePattern + ` | ` + rotatedPattern + `)$`)
```

base/logger_file.go
```diff
@@ -384,3 +396,27 @@ func runLogDeletion(ctx context.Context, logDirectory string, logLevel string, s
...
+}
```
bbrks: Summarising prior comments into single suggestion

```suggestion
// getDeletionDirAndRegexp will return the directory and a regexp matching log file and rotated patterns.
func getDeletionDirAndRegexp(path string) (string, *regexp.Regexp) {
	dir, filename := filepath.Split(path)

	// foo
	// foo-2019-01-01T00-00-00.000
	// foo-2019-01-01T00-00-00.000.gz
	filenamePattern := regexp.QuoteMeta(filename)
	rotatedPattern := filenamePattern + lumberjackRotationMidfix + optionalCompressSuffix

	if lastDot := strings.LastIndex(filename, "."); lastDot != -1 {
		// foo.log
		// foo-2019-01-01T00-00-00.000.log
		// foo-2019-01-01T00-00-00.000.log.gz
		//
		// or
		//
		// foo.bar.log
		// foo.bar-2019-01-01T00-00-00.000.log
		// foo.bar-2019-01-01T00-00-00.000.log.gz
		prefix := filename[:lastDot]
		suffix := filename[lastDot:]
		rotatedPattern = regexp.QuoteMeta(prefix) + lumberjackRotationMidfix + regexp.QuoteMeta(suffix) + optionalCompressSuffix
	}

	return dir, regexp.MustCompile(`^(` + filenamePattern + `|` + rotatedPattern + `)$`)
}
```

base/logger_file.go
```diff
@@ -43,6 +44,8 @@ var (
 
 	belowMinValueFmt = "%s for %v was set to %v which is below the minimum of %v"
 	aboveMaxValueFmt = "%s for %v was set to %v which is above the maximum of %v"
+
+	lumberjackRotationMidfix = `-\d{4}-\d{2}-\d{2}T\d{2}-\d{2}-\d{2}.\d{3}`
```
bbrks, line 49: related to suggestion in `getDeletionDirAndRegexp`

```suggestion
	// lumberjack backupTimeFormat = "2006-01-02T15-04-05.000"
	lumberjackRotationMidfix = `-\d{4}-\d{2}-\d{2}T\d{2}-\d{2}-\d{2}.\d{3}`
	// lumberjack compressSuffix = ".gz"
	optionalCompressSuffix = `(\.gz)?`
```

bbrks (changes requested):
I'm OK with this approach but just have a few comments around `getDeletionDirAndRegexp`

base/logger_file.go
torcolvin: I thought you needed to put a non capturing group inside a capturing group, but this seems to a rule I've made up that has no basis in reality.

torcolvin: This looks great, thanks!!

bbrks (approved):

PR #7061, opened 2024-08-13
[3.1.10 backport] CBG-4127 use a non cancellable context for OIDC metadata

torcolvin:
Backports https://github.com/couchbase/sync_gateway/commit/5db7dbed861ff0b10d0d7ac803fe5518dfeaf9da

This was not a clean cherry-pick, I split the difference on `NewNonCancelCtxForDatabase` to pull some of the dbname from the context but otherwise use nil db console logging config in auth package since it not on the `LoggingContext` in 3.1 branch.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2650/

### DIFF
--- a/auth/oidc.go
+++ b/auth/oidc.go
@@ -692,8 +692,7 @@ func (op *OIDCProvider) startDiscoverySync(ctx context.Context, discoveryURL str
 		return err
 	}
 	go func() {
-		// use DbConsoleLogConfig as nil in 3.1 branch only. This is stored on database in >= 3.2
-		ctx := base.NewNonCancelCtxForDatabase(ctx, nil).Ctx
+		ctx := base.NewNonCancelCtxForDatabase(ctx).Ctx
 		for {
 			select {
 			case <-time.After(duration):
--- a/auth/oidc.go
+++ b/auth/oidc.go
@@ -692,6 +692,8 @@ func (op *OIDCProvider) startDiscoverySync(ctx context.Context, discoveryURL str
 		return err
 	}
 	go func() {
+		// use DbConsoleLogConfig as nil in 3.1 branch only. This is stored on database in >= 3.2
+		ctx := base.NewNonCancelCtxForDatabase(ctx, nil).Ctx
 		for {
 			select {
 			case <-time.After(duration):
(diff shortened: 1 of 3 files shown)


auth/oidc.go
```diff
@@ -692,6 +692,8 @@ func (op *OIDCProvider) startDiscoverySync(ctx context.Context, discoveryURL str
 		return err
 	}
 	go func() {
+		// use DbConsoleLogConfig as nil in 3.1 branch only. This is stored on database in >= 3.2
+		ctx := base.NewNonCancelCtxForDatabase(ctx, nil).Ctx
```
bbrks: I don't think we need to make this function diverge across versions. There's no reason why this signature can't be the same, only taking in `ctx`, and the actual implementation sets `ctx.DbConsoleLogConfig` in 3.1

This is nessesary for log streaming using per-database log levels/log keys, and passing `nil` in here potentially breaks that for logging coming from this goroutine.

base/util.go
```diff
@@ -64,12 +64,14 @@ func NewNonCancelCtx() NonCancellableContext {
 }
 
 // NewNonCancelCtx creates a new background context struct for operations that require a fresh context, with database logging context added
-func NewNonCancelCtxForDatabase(dbName string, dbConsoleLogConfig *DbConsoleLogConfig) NonCancellableContext {
-	dbLogContext := DatabaseLogCtx(context.Background(), dbName, dbConsoleLogConfig)
-	ctxStruct := NonCancellableContext{
+func NewNonCancelCtxForDatabase(parentCtx context.Context, dbConsoleLogConfig *DbConsoleLogConfig) NonCancellableContext {
+	ctx := getLogCtx(parentCtx)
+
+	dbLogContext := DatabaseLogCtx(context.Background(), ctx.Database, dbConsoleLogConfig)
```
bbrks: This change keeps the same function signature as 3.2 and doesn't lose db console log settings in the OIDC case above. Keeping the behaviour much closer to the 3.2 branch

```suggestion
func NewNonCancelCtxForDatabase(parentCtx context.Context) NonCancellableContext {
	ctx := getLogCtx(parentCtx)

	dbLogContext := DatabaseLogCtx(context.Background(), ctx.Database, ctx.DbConsoleLogConfig)
```

bbrks (changes requested):

base/util.go
torcolvin: Agreed, I didn't see it because it had a different name than it does on main.

bbrks (approved):

PR #7064, opened 2024-08-13
CBG-4121 improve logging for jwts

torcolvin:
- log client name if there is a validation error instead of %v
- don't log token in the case of error
- log identity with UD(%+v) which shows the values of the keys

Verified by eye in unit tests.

### DIFF
--- a/auth/oidc.go
+++ b/auth/oidc.go
@@ -418,7 +418,7 @@ func getJWTUsername(provider JWTConfigCommon, identity *Identity) (username stri
 	if provider.UsernameClaim != "" {
 		value, ok := identity.Claims[provider.UsernameClaim]
 		if !ok {
-			return "", base.RedactErrorf("jwt: specified claim %q not found in id_token, identity: %+v", base.UD(provider.UsernameClaim), base.UD(identity))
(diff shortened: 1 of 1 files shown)


auth/oidc.go
```diff
@@ -418,7 +418,7 @@ func getJWTUsername(provider JWTConfigCommon, identity *Identity) (username stri
 	if provider.UsernameClaim != "" {
 		value, ok := identity.Claims[provider.UsernameClaim]
 		if !ok {
-			return "", fmt.Errorf("jwt: specified claim %q not found in id_token, identity: %v", provider.UsernameClaim, identity)
+			return "", base.RedactErrorf("jwt: specified claim %q not found in id_token, identity: %+v", base.UD(provider.UsernameClaim), base.UD(identity))
```
bbrks: `provider.UsernameClaim` isn't user data, it's a key name defined in OIDC provider config, rather than the value sent by the client.

```suggestion
			return "", base.RedactErrorf("jwt: specified claim %q not found in id_token, identity: %+v", provider.UsernameClaim, base.UD(identity))
```

bbrks (changes requested):
One nit, otherwise looks good

bbrks (approved):

PR #7076, opened 2024-08-14
CBG-3917: pass revNo from cbgt into feed event

gregns1:
CBG-3917

- Passes rev no from cbgt into feed event 
- No test, can't find a nice way to assert on this like Tor did in his PR on the dcp client given this is managed by cbgt. I ran import unit test and stuck break point on the feed event and visually verified that the revNo is populated. 

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2660/

### DIFF
--- a/base/dcp_common.go
+++ b/base/dcp_common.go
@@ -522,12 +522,13 @@ func dcpKeyFilter(key []byte, metaKeys *MetadataKeys) bool {
 }
 
 // Makes a feedEvent that can be passed to a FeedEventCallbackFunc implementation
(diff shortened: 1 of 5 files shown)


torcolvin (commented):
Can you run go.mod tidy?

I'd be OK upgrading cbgt on main too so these aren't out of sync, but I don't have a strong preference on that.

torcolvin (approved):

PR #7079, opened 2024-08-15
CBG-4179 check early for context cancellation

torcolvin:
The changes for the ticket were to add check for context cancellation in `buildRevokedFeed` in the case the changes hit `requiresRevocation`. This would require a check for cancellation at the top of `for _, logEntry := range changes` section. I added another check for cancellation before we run an addition N1QL query, but this could be an overoptimization. A single N1QL query on a cancelled request is probably not a big deal.

* In general, I changed a single `select` with `<-ctx.Done()` to check for `ctx.Err() != nil` to be more clear.
* Everywhere we send information to a channel, I've added a check for context cancellation to avoid leaking a stuck change waiter. This is probably not super likely to happen since the output has a buffer size of 50, and there are other exit points.

The change I feel least good about  is the following:

```
				if options.ChangesCtx.Err() != nil {
					return
				}
				select {
				case <-options.ChangesCtx.Done():
					return
				case output <- minEntry:
				}
```				
Checking for the `Done()` channel is necessary in the case that the output receiver no longer exists, so the function will return. However, when entering the select statement if there are changes and the output channel can still receive messages if it hasn't reached the 50 buffer size, we still want to exit. It's non deterministic if you'll hit the send to output or the Done check, so I do an explicit check before the select loop to exit early.

I did not add a check for context cancellation in `DocIDChangesFeed` since the construction is somewhat weird and synchronous anyway, it populates the entire set of changes into a buffered channel sized with the number of changes being sent. That channel is already closed by the time it exits the function.

## Pre-review checklist


db/changes.go
```diff
@@ -241,6 +246,11 @@ func (db *DatabaseCollectionWithUser) buildRevokedFeed(ctx context.Context, ch c
 
 			sentChanges := 0
 			for _, logEntry := range changes {
+				if options.ChangesCtx.Err() != nil {
```
adamcfraser, line 249: If we're checking here for each loop iteration, I don't think we need to check again on line 307 - what do you think?

db/changes.go
```diff
@@ -758,7 +776,10 @@ func (col *DatabaseCollectionWithUser) SimpleMultiChangesFeed(ctx context.Contex
 				if err != nil {
 					base.WarnfCtx(ctx, "Unable to obtain channel cache for %s, terminating feed", base.UD(chanName))
 					change := makeErrorEntry("Channel cache unavailable, terminating feed")
-					output <- &change
+					select {
+					case output <- &change:
+					case <-options.ChangesCtx.Done():
+					}
```
adamcfraser, line 782: We're already terminating the changes feed here, so I don't think we want to check the ChangesCtx, and we definitely don't want to potentially miss sending the error entry to output if it doesn't get picked by the select.

db/changes.go
```diff
@@ -409,6 +419,10 @@ func (db *DatabaseCollectionWithUser) changesFeed(ctx context.Context, singleCha
 		//   2. A limit is specified on the incoming ChangesOptions, and that limit is reached
 		//   3. An error is returned when calling singleChannelCache.GetChanges
 		for {
+			if options.ChangesCtx.Err() != nil {
+				base.DebugfCtx(ctx, base.KeyChanges, "Terminating channel feed %s", base.UD(to))
+				return
```
adamcfraser, line 424: Are you intentionally not sending an error entry to the feed in this case, because the outer consumer of the feed is also terminating based ChangesCtx?

db/changes.go
```diff
@@ -769,7 +790,7 @@ func (col *DatabaseCollectionWithUser) SimpleMultiChangesFeed(ctx context.Contex
 				if useLateSequenceFeeds {
 					lateSequenceFeedHandler := lateSequenceFeeds[chanID]
 					if lateSequenceFeedHandler != nil {
-						latefeed, err := col.getLateFeed(lateSequenceFeedHandler, singleChannelCache)
+						latefeed, err := col.getLateFeed(ctx, lateSequenceFeedHandler, singleChannelCache)
```
adamcfraser: Shouldn't this be passing options.ChangesCtx?

db/changes.go
```diff
@@ -1029,20 +1060,10 @@ func (col *DatabaseCollectionWithUser) SimpleMultiChangesFeed(ctx context.Contex
 
 				if waitResponse == WaiterClosed {
 					break outer
+				} else if options.ChangesCtx.Err() != nil {
```
adamcfraser: I think it would be reasonable to move this to a standalone if check prior to line 1061 - I don't think we actually want to continue processing via `break outer`, even though that's how the previous code behaved, as we're just going to eventually check options.ChangesCtx above and error out anyway.

db/changes.go
```diff
@@ -1054,7 +1075,10 @@ func (col *DatabaseCollectionWithUser) SimpleMultiChangesFeed(ctx context.Contex
 			if err != nil {
 				change := makeErrorEntry("User not found during reload - terminating changes feed")
 				base.DebugfCtx(ctx, base.KeyChanges, "User not found during reload - terminating changes feed with entry %+v", base.UD(change))
-				output <- &change
+				select {
```
adamcfraser, line 1078: Same as elsewhere, I think checking ctx.Err() is preferable to a select.

db/changes.go
```diff
@@ -1217,7 +1241,12 @@ func (db *DatabaseCollectionWithUser) getLateFeed(feedHandler *lateSequenceFeed,
 				Seq: logEntry.Sequence,
 			}
 			change := makeChangeEntry(logEntry, seqID, singleChannelCache.ChannelID())
-			feed <- &change
+			select {
+			case <-ctx.Done():
+				return
+
+			case feed <- &change:
+			}
```
adamcfraser, line 1227: As elsewhere, ctx.Err() instead?

adamcfraser (changes requested):

db/changes.go
adamcfraser: I just saw your PR comment "Checking for the Done() channel is necessary in the case that the output receiver no longer exists".  Is that an actual possibility?  I don't think adding non-deterministic behaviour is needed if it's being defensive for a scenario that can't happen.

db/changes.go
torcolvin: The check on 307 is around sending a over a channel, and any sends need to account for the fact that the consumer might have already exited.

db/changes.go
torcolvin: That is correct, there are outside consumers that will exit the context has been canceled.

- https://github.com/couchbase/sync_gateway/blob/cb97fc90623afda7527556a80cebe251edc63f56/db/changes.go#L1510
- https://github.com/couchbase/sync_gateway/blob/cb97fc90623afda7527556a80cebe251edc63f56/rest/changes_api.go#L434 (this case weird because the context cancelation is not `ChangesOptions.ChangesCtx` directly, and a case for removal of the context as on `ChangesOptions`

In both cases `forceClose` is set if the client terminates the connection.

db/changes.go
torcolvin: It definitely can happen because the consumer in `GetChanges` will exit when the `optinos.ChangesCtx`. I've added code to remove `cancelCtx` from `GenerateChanges` since these were always the same.

As discussed offline, I would like to make the context only a parameter that is passed into the changes feed, but I think we should defer from this PR.

db/changes.go
torcolvin: We have to use a select in the case that output consumer has exited. The options for things that can happen:

1. send to the feed works, the channel is unblocked
2. the send to the feed blocks, because the buffer is full. The buffer will empty because the consumer is slow to catch up
3. the send to the feed blocks, because the buffer is full. The context hasn't cancelled yet, but it will while the buffer is full.

Without this code, this go routine might not exit. In practice, this would be unlikely to happen with an `context.Err() != nil` check - the context would have to alive and then cancelled when the output is sending. This is either an extremely likely data race, or a circumstance where the output buffer is already full and the context is just slow to cancel. Given the buffer size is 50, it's not super likely that it is full.

I still think this is the right behavior to include a `<- context.Done()` when we do a send in order to not to hang a goroutine.

db/changes.go
torcolvin: See above on filing up the buffer and cancelling while the buffer is full.

db/changes.go
```diff
@@ -1001,7 +1028,11 @@ func (col *DatabaseCollectionWithUser) SimpleMultiChangesFeed(ctx context.Contex
 			// If nothing found, and in wait mode: wait for the db to change, then run again.
 			// First notify the reader that we're waiting by sending a nil.
 			base.DebugfCtx(ctx, base.KeyChanges, "MultiChangesFeed waiting... %s", base.UD(to))
-			output <- nil
+			select {
+			case <-ctx.Done():
```
adamcfraser: As above, should this be options.ChangesCtx.Done()?

db/changes.go
```diff
@@ -889,7 +910,10 @@ func (col *DatabaseCollectionWithUser) SimpleMultiChangesFeed(ctx context.Contex
 							// On feed error, send the error and exit changes processing
 							if current[i].Err == base.ErrChannelFeed {
 								base.WarnfCtx(ctx, "MultiChangesFeed got error reading changes feed: %v", current[i].Err)
-								output <- current[i]
+								select {
+								case <-ctx.Done():
```
adamcfraser: Why ctx.Done() and not options.ChangesCtx.Done()?

db/changes.go
```diff
@@ -1054,7 +1075,10 @@ func (col *DatabaseCollectionWithUser) SimpleMultiChangesFeed(ctx context.Contex
 			if err != nil {
 				change := makeErrorEntry("User not found during reload - terminating changes feed")
 				base.DebugfCtx(ctx, base.KeyChanges, "User not found during reload - terminating changes feed with entry %+v", base.UD(change))
-				output <- &change
+				select {
+				case <-ctx.Done():
```
adamcfraser: Same question as above about whether this should be options.ChangesCtx.Done()

adamcfraser (changes requested):

torcolvin:
The last revision is from test failures:

do not expect an error message ChangesEntry.Err if context is cancelled

- ChangesFeed.Err is checked in a few places
    - GenerateChanges checks entry.Err and only serves to exit the function, it won't return that entry. If there is no end entry, it will hit `<-options.ChangesCtx.Done()` and set `forceClose`.
    - `TestChannelQueryCancellation` checked the error but this was unnecessary since the assertions are that the changes feed exits  with only one `ViewQuery`
- drop error return from `generateBlipSyncChanges` since it was unused and whether or not an error is returned 
- move DatabaseContext.GetChanges test only function to test only code since it has particular use cases.
     - require that there is no error so the caller doesn't have to check since it was only used for `TestChannelQueryCancellation`
- slight behavior change to set `forceClose` in `GenerateChanges`/`sendSimpleChanges` to check for context cancellation in the case that the `ChangesEntry.Err` was hit first, so this marks this as `forceClose`. `forceClose` is used to call
  `DatabaseContext.NotifyTerminatedChanges` to wake other changes feeds.

adamcfraser (approved):

PR #7082, opened 2024-08-16
CBG-3693: Improved logging for no-user/no auto-register case in JWT auth

bbrks:
CBG-3693

- Logs when a valid token is used with an incorrect or unknown username in the claims, and we don't have auto-register set for the provider.
- Covered by existing test `TestOpenIDConnectImplicitFlowEdgeCases/unsuccessful user auth when username_claim is set with different value in token`

```
2024-08-19T17:58:59.182+01:00 [INF] Auth: c:#003 db:db Fetching provider config from explicitly defined discovery endpoint: <ud>http://127.0.0.1:55719/foo/.well-known/openid-configuration</ud>
2024-08-19T17:58:59.184+01:00 [INF] Auth: c:#003 db:db No change in discovery config detected at this time, next sync will be after 24h0m0s
2024-08-19T17:58:59.185+01:00 [INF] Auth: c:#003 db:db User "<ud>bar_alice</ud>" not found and provider "foo" does not have Register enabled
2024-08-19T17:58:59.185+01:00 [INF] HTTP: c:#003 db:db POST /db/_session
2024-08-19T17:58:59.185+01:00 [INF] HTTP: c:#003 db:db #003:     --> 401 Invalid login  (3.3 ms)
```

- n/a

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -873,7 +873,7 @@ func (auth *Authenticator) authenticateJWTIdentity(identity *Identity, provider 
 		base.InfofCtx(auth.LogCtx, base.KeyAuth, "Error retrieving OIDCUsername: %v", err)
 		return nil, PrincipalConfig{}, time.Time{}, err
 	}
-	base.DebugfCtx(auth.LogCtx, base.KeyAuth, "OIDCUsername: %v", base.UD(username))
+	base.DebugfCtx(auth.LogCtx, base.KeyAuth, "Got username %q from JWT", base.UD(username))
 
 	var jwtRoles, jwtChannels base.Set
 	if provider.RolesClaim != "" {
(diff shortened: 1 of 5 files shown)


auth/auth.go
```diff
@@ -873,7 +873,7 @@ func (auth *Authenticator) authenticateJWTIdentity(identity *Identity, provider
 		base.InfofCtx(auth.LogCtx, base.KeyAuth, "Error retrieving OIDCUsername: %v", err)
 		return nil, PrincipalConfig{}, time.Time{}, err
 	}
-	base.DebugfCtx(auth.LogCtx, base.KeyAuth, "OIDCUsername: %v", base.UD(username))
+	base.DebugfCtx(auth.LogCtx, base.KeyAuth, "Got username %q from JWT", base.UD(username))
```
torcolvin: It seems uncommon to have multiple providers, but should we be logging provider here?

I do not have a preference.

auth/auth.go
```diff
@@ -899,29 +899,32 @@ func (auth *Authenticator) authenticateJWTIdentity(identity *Identity, provider
...
 	// Auto-registration. This will normally be done when token is originally returned
 	// to client by oidc callback, but also needed here to handle clients obtaining their own tokens.
-	if user == nil && provider.Register {
+	if provider.Register {
 		base.DebugfCtx(auth.LogCtx, base.KeyAuth, "Registering new user: %v with email: %v", base.UD(username), base.UD(identity.Email))
-		var err error
-		user, err = auth.RegisterNewUser(username, identity.Email)
+		user, err := auth.RegisterNewUser(username, identity.Email)
 		if err != nil && !base.IsCasMismatch(err) {
 			base.InfofCtx(auth.LogCtx, base.KeyAuth, "Error registering new user: %v", err)
```
torcolvin, line 921: Is this error going to contain the username? I think in practice this isn't going to error looking at some pathways through `RegisterNewUser`. The best idea for this to fail is if `auth.Save` fails with a write error.

I did notice that if this user has a bad email address we still let the user register. That's outside of this PR, but I wonder if that's the right behavior.

torcolvin (commented):
Looks good, just some questions.

auth/auth.go
bbrks: Included username in log message.

~Are you sure about that email comment though? It's validated in `p.validate()` during save and returns an error.~

Edit: I see the warning from `SetEmail` now that will allow an empty email through upon register. I am not going to worry about it.

bbrks: https://github.com/couchbase/sync_gateway/pull/4217#discussion_r318173541

torcolvin (approved):

PR #7084, opened 2024-08-21
CBG-4134: link rev cache memory limit config option to rev cache

gregns1:
CBG-4134

- links db config option to rev cache 	
- Added warning for when someone tries to configure it in community edition 
- Probably should hold off merging till the memory based rev cache work is in main

## Pre-review checklist

- [x]  #7040

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2673/

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -246,9 +246,9 @@ type DeprecatedCacheConfig struct {
 }
 
 type RevCacheConfig struct {
-	MaxItemCount     *uint32 `json:"max_item_count,omitempty"` // Maximum number of revisions to store in the revision cache
-	MaxMemoryCountMB *uint32 `json:"max_memory_count_mb"`      // Maximum amount of memory the rev cache should consume in MB
-	ShardCount       *uint16 `json:"shard_count,omitempty"`    // Number of shards the rev cache should be split into
+	MaxItemCount     *uint32 `json:"size,omitempty"`                // Maximum number of revisions to store in the revision cache
+	MaxMemoryCountMB *uint32 `json:"max_memory_count_mb,omitempty"` // Maximum amount of memory the rev cache should consume in MB, when configured it will work in tandem with max items
+	ShardCount       *uint16 `json:"shard_count,omitempty"`         // Number of shards the rev cache should be split into
 }
 
 type ChannelCacheConfig struct {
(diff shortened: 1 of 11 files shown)


rest/config.go
```diff
@@ -246,8 +246,9 @@ type DeprecatedCacheConfig struct {
 }
 
 type RevCacheConfig struct {
-	Size       *uint32 `json:"size,omitempty"`        // Maximum number of revisions to store in the revision cache
```
bbrks, line 249: We can't remove config options like this, it would be a breaking change for customers with existing configuration.

If we want to keep the `size` JSON field name, and rename the internal struct field to `MaxItemCount` to better represent the behaviour in the codebase, we can do that. The alternatve would be adding a new property, deprecating `Size` but still allowing it as a fallback.

Both of these options come with downsides.

rest/config.go
```diff
@@ -246,8 +246,9 @@ type DeprecatedCacheConfig struct {
...
-	ShardCount *uint16 `json:"shard_count,omitempty"` // Number of shards the rev cache should be split into
+	MaxItemCount     *uint32 `json:"max_item_count,omitempty"` // Maximum number of revisions to store in the revision cache
+	MaxMemoryCountMB *uint32 `json:"max_memory_count_mb"`      // Maximum amount of memory the rev cache should consume in MB
```
bbrks: What's the behaviour when both of these are set? The lower of the two?

rest/config.go
```diff
@@ -246,8 +246,9 @@ type DeprecatedCacheConfig struct {
 }
 
 type RevCacheConfig struct {
```
bbrks, line 248: This should have a corresponding change in the API specs for db config.

bbrks (changes requested):

rest/config.go
gregns1: Totally didn't think about this, thanks!

rest/config.go
gregns1: Correct, both will work in tandem with one another when memory limit is set. I have added a comment to `MaxMemoryCountMB` to reflect this info, let me know if it needs work.

bbrks (approved):

PR #7087, opened 2024-08-21
Update docs for 3.2.1

torcolvin:
Created with:

- git checkout origin/main .redocly.yaml docs
- manually edit 3.2 -> 3.3
- git cherry-pick CBG-3933-docs-fix

If I create this branch without creating a manifest, no builds will be done for this, which seems right. On the other hand if there are any more commits in 3.2.0, release/3.2.1 existing means that backports will have to be done to 3.2.0 and 3.2.1. I do not expect any more changes.

### DIFF
--- a/.redocly.yaml
+++ b/.redocly.yaml
@@ -15,24 +15,55 @@ apis:
       remove-x-internal: on
   admin-internal:
     root: "./docs/api/admin.yaml"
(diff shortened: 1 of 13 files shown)


torcolvin:
I'm going to do individual backports into 3.2.1, even before we set up manifests.

PR #7096, opened 2024-08-29
CBG-3909: use deltas for pv and mv when persisting to the bucket

gregns1:
CBG-3909

- Adds functionality to have previous versions and merge version in the persisted hlv stored as deltas.
- Added functionality to read the persisted form of hlv and convert the pv and mv lists back to full values from deltas and back into map form in memory 
- Functionality to take in memory hlv and convert the pv and mv maps into delta lists and persist to the bucket in that form
- Currently I have left the fact we have to convert between cas string and uin64 cas a bit back and forth for converting deltas for work in separate ticket (if we decide to change the in memory format back to uint674 for cas)

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2739/


base/util.go
```diff
@@ -48,6 +48,8 @@ const (
 	kMaxDeltaTtlDuration = 60 * 60 * 24 * 30 * time.Second
 )
 
+var MaxDecodedLength = len(Uint64CASToLittleEndianHex(math.MaxUint64))
```
adamcfraser: Can you add a comment saying what this actual value is, just for reference?

base/util.go
```diff
@@ -1018,6 +1020,43 @@ func HexCasToUint64(cas string) uint64 {
...
+		evenHexLen = append(evenHexLen, '0')
+		casByte = evenHexLen
+	}
+
+	decoded := make([]byte, MaxDecodedLength)
+	_, err := hex.Decode(decoded, casByte[2:])
+	if err != nil {
+		return 0, err
+	}
+	res := binary.LittleEndian.Uint64(decoded)
```
adamcfraser, line 1047: Doesn't binary.LittleEndian.Uint64 expect a length 8 []byte as input?  Here it looks like we're setting decoded to MaxDecodedLength (18) - can you help me understand the case where casByte[2:] is longer than 8 bytes but we need to support this and are intentionally ignoring trailing bytes?

I thought this was just handling the case where it's less than 8 bytes because we've trimmed trailing zeros.

base/util.go
```diff
@@ -1018,6 +1020,43 @@ func HexCasToUint64(cas string) uint64 {
...
+	return res, nil
+}
+
+// Uint64ToLittleEndianHexAndStripZeros will convert a uin64 type to little endian hex, stripping any zeros off the end
```
adamcfraser:
```suggestion
// Uint64ToLittleEndianHexAndStripZeros will convert a uint64 type to little endian hex, stripping any zeros off the end
```

db/document.go
```diff
@@ -1238,6 +1244,59 @@ func (doc *Document) MarshalWithXattrs() (data []byte, syncXattr, vvXattr, mouXa
...
+	hlvData, err := base.JSONMarshal(&persistedHLV)
+	if err != nil {
+		return nil, err
+	}
+	return hlvData, nil
+}
+
+// ParseHLVFields will parse raw bytes of the hlv xattr from persisted version to in memory version, converting the
+// deltas in pv and mv back into in memory format
+func ParseHLVFields(xattr []byte) (*HybridLogicalVector, error) {
```
adamcfraser: If an HLV ends up corrupted deltas, does the document just become unreadable (i.e. when this function returns error).  I feel like we need a way to repair/rewrite the HLV in this scenario (possibly omitting malformed entries in PV).

base/util_test.go
```diff
@@ -1735,3 +1735,23 @@ func TestCASToLittleEndianHex(t *testing.T) {
 	littleEndianHex := Uint64CASToLittleEndianHex(casValue)
 	require.Equal(t, expHexValue, string(littleEndianHex))
 }
+
+func TestUint64CASToLittleEndianHexAndStripZeros(t *testing.T) {
+	hexLE := "0x0000000000000000"
+	u64 := HexCasToUint64(hexLE)
+	hexLEStripped := Uint64ToLittleEndianHexAndStripZeros(u64)
+	u64Stripped := HexCasToUint64(hexLEStripped)
```
adamcfraser: I might be overlooking something, but I expected this line to be using HexCasToUint64ForDelta.  Am I misunderstanding how Uint64ToLittleEndianHexAndStripZeros works?

db/hybrid_logical_vector.go
```diff
@@ -92,18 +185,23 @@ func ParseDecodedVersion(versionString string) (version DecodedVersion, err erro
...
 
 // String returns a version/sourceID pair in CBL string format
-func (v Version) String() string {
+func (v Version) BlipString() string {
 	return v.Value + "@" + v.SourceID
 }
 
+// DeltaString returns a pv or mv delta format string representation of version
+func (v Version) DeltaString() string {
+	return v.SourceID + "@" + v.Value
```
adamcfraser: I don't think we want two different formats (value@source and source@value) - that's going to lead to confusion.  Can you talk to XDCR about standardizing on value@source?

adamcfraser (changes requested):

base/util_test.go
gregns1: Good spot, test didn't have sufficient test cases so wasn't failing. Good catch, added a new test case and fixed this.

db/document.go
gregns1: Good shout I have not returned error in the case and continued to build HLV without the entries in this case. I chose to log at warn as if this happens its probably something the customer/we should know about? Happy to change that log level tif you have an opinion though. Added test for this pathway too `TestParseVersionVectorCorruptDelta`

base/util.go
gregns1: Good point. I don't think the max decoded length line is needed, I have removed it.

adamcfraser: Well, hex.Decode is going to panic if casByte[2:] overflows decoded, so it's possible the previous code was defensive.  https://go.dev/play/p/_LEScaOT1FG
Our HexCasToUint64 has an explicit length check to protect against this.  Possibly we just want to do the same (but also supporting the case here that the length can be less than 8): doing something like return error `if len(casByte) > 10` at the top of the function?

adamcfraser (changes requested):

base/util.go
gregns1: Added some safety checks around this area, let me know what you think

db/hybrid_logical_vector.go
```diff
@@ -122,6 +203,16 @@ type HybridLogicalVector struct {
 	PreviousVersions  map[string]string `json:"pv,omitempty"`        // map of previous versions for fast efficient lookup
 }
 
+// PersistedHLV is the version of the version vector that is persisted to the bucket, pv and mv will be lists of source version pairs in the bucket
```
adamcfraser: To avoid a new struct and the conversion between PersistedHLV and HybridLogicalVector, I'm wondering if we can take the same approach for PreviousVersions and MergeVersions that we do for RevTree inside SyncData - define a custom UnmarshalJSON/MarshalJSON that does the transformation from deltas to map (and vice versa).  That would properly isolate the in-bucket storage format to the persistence layer, and avoid the extra allocations/copy between PersistedHLV and HybridLogicalVector.  So basically:
- define a new type HLVVersions that's map[string]string, and use that for MergeVersions and PreviousVersions in HybridLogicalVector
- define MarshalJSON and UnmarshalJSON for HLVVersions that use PersistedDeltasToMap and VersionsToDeltas

I think this change would really help to hide the persistence details from users of HybridLogicalVector and minimize the impact of this change.

adamcfraser (changes requested):
Left a suggestion around implementation - have a look and let me know what you think.

db/hybrid_logical_vector.go
```diff
@@ -111,22 +198,88 @@ func (hlv *HybridLogicalVector) ExtractCurrentVersionFromHLV() *Version {
...
+	CurrentVersionCAS string `json:"cvCas,omitempty"`     // current version cas (or cvCAS) stores the current CAS in little endian hex format at the time of replication
+	ImportCAS         string `json:"importCAS,omitempty"` // Set when an import modifies the document CAS but preserves the HLV (import of a version replicated by XDCR)
+	SourceID          string `json:"src"`                 // source bucket uuid in (base64 encoded format) of where this entry originated from
+	Version           string `json:"ver"`                 // current cas in little endian hex format of the current version on the version vector
+	PrevVersions             // map of merge versions for fast efficient lookup
+	MVersions                // map of previous versions for fast efficient lookup
+}
+
+// PersistedHLV is the version of the version vector that is persisted to the bucket, pv and mv will be lists of source version pairs in the bucket
+type PersistedHLV struct {
```
adamcfraser: With the updated JSONMarshal approach, I was anticipating we'd be able to remove PersistedHLV and just use HybridLogicalVector.  I also don't think the aliasing (HLVAlias) is needed with this approach.
Have a look at https://go.dev/play/p/aYbiN9q-aXQ for the basic approach I was imagining, and let me know if there's something you don't think will work with that approach.

adamcfraser (changes requested):

base/util.go
```diff
@@ -1018,6 +1018,53 @@ func HexCasToUint64(cas string) uint64 {
...
+	if len(casByte) <= 2 {
+		return 0, fmt.Errorf("hex value is too short")
+	}
+
+	// as we strip any zeros off the end of the hex value for deltas, the input delta could be odd length
+	if len(casByte)%2 != 0 {
+		evenHexLen := make([]byte, len(casByte), len(casByte)+1)
+		copy(evenHexLen, casByte)
+		evenHexLen = append(evenHexLen, '0')
+		casByte = evenHexLen
```
adamcfraser: Given than len(casByte) is odd, I think that `casByte=append(casByte, '0')` is not going to require golang to increase the capacity of the slice (based on default slice capacities).   If that's correct I think we can avoid the GC work of the new allocation you've got here.

base/util_test.go
```diff
@@ -1735,3 +1735,33 @@ func TestCASToLittleEndianHex(t *testing.T) {
 	littleEndianHex := Uint64CASToLittleEndianHex(casValue)
 	require.Equal(t, expHexValue, string(littleEndianHex))
 }
+
+func TestUint64CASToLittleEndianHexAndStripZeros(t *testing.T) {
+	hexLE := "0x0000000000000000"
+	u64 := HexCasToUint64(hexLE)
+	hexLEStripped := Uint64ToLittleEndianHexAndStripZeros(u64)
```
adamcfraser, line 1742: Running this test locally, the value of hexLEStripped is "000".  Is this what we intend?
It looks like any hex value with only two non-zero digits retains an extra zero.  (e.g. `0xa500000000000000` results in `a50` when I would have expected `a5`.  Is there a requirement for a minimum length of three?

db/hybrid_logical_vector.go
```diff
@@ -30,6 +31,85 @@ type Version struct {
...
+
+func (vde VersionsDeltas) Len() int { return len(vde) }
+
+func (vde VersionsDeltas) Swap(i, j int) {
+	vde[i], vde[j] = vde[j], vde[i]
+}
+
+func (vde VersionsDeltas) Less(i, j int) bool {
+	if vde[i].Value == vde[j].Value {
+		return vde[i].SourceID < vde[j].SourceID
```
adamcfraser: For the purpose of computing deltas, does the ordering of sources with matching values matter?  i.e. do we actually need SourceID to be acting as a tiebreaker, or can we just stick to `vde[i].Value < vde[j].Value` below and have equality return Less() = false?

adamcfraser: Also, do we have unit test coverage for matching Values?

db/hybrid_logical_vector.go
```diff
@@ -30,6 +31,85 @@ type Version struct {
...
+func VersionDeltas(versions map[string]uint64) VersionsDeltas {
+	if versions == nil {
+		return nil
+	}
+
+	vdm := make(VersionsDeltas, 0, len(versions))
+	for src, vrs := range versions {
+		vdm = append(vdm, CreateVersion(src, vrs))
+	}
+
```
adamcfraser, line 60: Should we return early here when len(versions) == 1?  And do we have test coverage for that scenario?

db/hybrid_logical_vector.go
```diff
@@ -30,6 +31,85 @@ type Version struct {
...
+		return nil
+	}
+
+	var vrsList []string
+	deltas := VersionDeltas(m)
+	for _, delta := range deltas {
+		key := delta.SourceID
+		val := delta.Value
+		encodedVal := base.Uint64ToLittleEndianHexAndStripZeros(val)
+		listItem := encodedVal + "@" + key
```
adamcfraser: It's a bit subtle that this format is different than delta.String().  I see now that these are two different representations of the version (simple hex used by blip, LittleEndian with strippedzeros here), but I think that's easy to miss.  What do you think about a Version.StringForVersionDelta() function with a comment explaining usage, that does lines 82 and 83?

db/hybrid_logical_vector.go
```diff
@@ -447,91 +528,70 @@ func CreateEncodedSourceID(bucketUUID, clusterUUID string) (string, error) {
...
 
 func (hlv HybridLogicalVector) MarshalJSON() ([]byte, error) {
-	var cvCasByteArray []byte
-	var importCASBytes []byte
-	var vrsCasByteArray []byte
+	var cvCas string
+	var importCAS string
+	var vrsCas string
+
+	bucketHLV := make(map[string]interface{})
```
adamcfraser: Is there a reason to use an untyped map here (as opposed to BucketVector)?

adamcfraser (changes requested):

base/util.go
gregns1: Running all my tests, capacity for input to this function is always the same as the length. But the only test I run that crosses into this path of execution is the first call to `HexCasToUint64ForDelta` inside test `TestUint64CASToLittleEndianHexAndStripZeros` for a zero cas value. Removing this if statement makes the test fail with error `encoding/hex: odd length hex string`

base/util_test.go
gregns1: Good spot, fixed this so now that delta is just 0, also added new test case for this

db/hybrid_logical_vector.go
gregns1: Good idea, yes we have a scenario where we have single entry in PV in TestPvDeltaReadAndWrite

db/hybrid_logical_vector.go
gregns1: I think I did this to align with xdcr approach on it. I have removed it because the more I think about it we don't need to align with them on this part. I have returned false in that scenario. 

Added a new test case to `TestVersionDeltaCalculation` for pv map with two entries that have the same value, leading to 0 deltas being computed.

base/util.go
adamcfraser: I wasn't suggesting to remove the if - just to replace
```
evenHexLen := make([]byte, len(casByte), len(casByte)+1)
copy(evenHexLen, casByte)
evenHexLen = append(evenHexLen, '0')
casByte = evenHexLen
```
with 
```
casByte = append(casByte, '0')
```
as that shouldn't trigger any new allocations when casByte starts out with an odd length.

adamcfraser (changes requested):
Everything looks good except for one clarification on my previous comment (reducing allocations when converting casByte to an even length)

base/util.go
gregns1: Oh apologies, completely got wrong end of the stick sorry. Have updated to make this change.

adamcfraser (approved):

PR #7099, opened 2024-09-02
CBG-3715: populate pRev on mou

gregns1:
CBG-3715

- Takes rev no from dcp event and passed down into import code for mou population 
- Added revNo into the `computeMetadataOnlyUpdate` function 
- Left the on demand import code rev no 0 for now. This work will be completed in [CBG-4203](https://jira.issues.couchbase.com/browse/CBG-4203)
- New test for pRev population on import feed `TestPrevRevNoPopulationImportFeed` 
- Added revno pass down in resync code too from the dcp event. Again left the query based resync as 0 for now. Added test assertions on this in `TestResyncMou`
- Also filed ticket for implementing virtual xattr in rosmar so we can enabled tests for rosmar [CBG-4233](https://jira.issues.couchbase.com/browse/CBG-4233) 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2681/

### DIFF
--- a/base/constants.go
+++ b/base/constants.go
@@ -144,7 +144,10 @@ const (
 	// Intended to be used in Meta Map and related tests
 	MetaMapXattrsKey = "xattrs"
 
-	DocumentXattrKey = "$document"
+	// VirtualXattrRevSeqNo is used to fetch rev seq no from documents virtual xattr
+	VirtualXattrRevSeqNo = "$document.revid"
+	// VirtualXattrCrc32c is used to fetch crc32c from documents virtual xattr
+	VirtualXattrCrc32c = "$document.value_crc32c"
 
 	// Prefix for transaction metadata documents
 	TxnPrefix = "_txn:"
--- a/base/constants.go
+++ b/base/constants.go
@@ -144,6 +144,8 @@ const (
 	// Intended to be used in Meta Map and related tests
 	MetaMapXattrsKey = "xattrs"
 
+	DocumentXattrKey = "$document"
+
 	// Prefix for transaction metadata documents
 	TxnPrefix = "_txn:"
 
--- a/base/constants.go
+++ b/base/constants.go
@@ -146,8 +146,8 @@ const (
 
 	// VirtualXattrRevSeqNo is used to fetch rev seq no from documents virtual xattr
 	VirtualXattrRevSeqNo = "$document.revid"
-	// VirtualXattrCrc32c is used to fetch crc32c from documents virtual xattr
-	VirtualXattrCrc32c = "$document.value_crc32c"
+	// VirtualDocumentXattr is used to fetch the documents virtual xattr
+	VirtualDocumentXattr = "$document"
 
 	// Prefix for transaction metadata documents
 	TxnPrefix = "_txn:"
--- a/db/background_mgr_resync_dcp_test.go
+++ b/db/background_mgr_resync_dcp_test.go
@@ -521,16 +521,13 @@ function sync(doc, oldDoc){
 	stats := getResyncStats(resyncMgr.Process)
 	assert.Equal(t, int64(2), stats.DocsChanged)
 
-	_, xattrs, _, err := collection.dataStore.GetWithXattrs(ctx, "sgWrite", []string{base.DocumentXattrKey})
+	_, xattrs, _, err := collection.dataStore.GetWithXattrs(ctx, "sgWrite", []string{base.VirtualXattrRevSeqNo})
 	require.NoError(t, err)
 
-	var retrievedVxattr map[string]interface{}
-	require.NoError(t, json.Unmarshal(xattrs[base.DocumentXattrKey], &retrievedVxattr))
+	var retrievedVRevIDxattr string
+	require.NoError(t, json.Unmarshal(xattrs[base.VirtualXattrRevSeqNo], &retrievedVRevIDxattr))
 
-	rev, ok := retrievedVxattr["revid"].(string)
-	require.True(t, ok, "Unable to obtain revSeqNo as string")
-
-	revNo, err := strconv.ParseUint(rev, 10, 64)
+	revNo, err := strconv.ParseUint(retrievedVRevIDxattr, 10, 64)
 	require.NoError(t, err)
 
 	syncData, mou, _ = getSyncAndMou(t, collection, "sgWrite")
@@ -540,15 +537,12 @@ function sync(doc, oldDoc){
 	require.Equal(t, revNo-1, mou.PreviousRevSeqNo)
 
 	// reset retrieved xattr
-	retrievedVxattr = nil
-	_, xattrs, _, err = collection.dataStore.GetWithXattrs(ctx, "sdkWrite", []string{base.DocumentXattrKey})
+	retrievedVRevIDxattr = ""
+	_, xattrs, _, err = collection.dataStore.GetWithXattrs(ctx, "sdkWrite", []string{base.VirtualXattrRevSeqNo})
 	require.NoError(t, err)
-	require.NoError(t, json.Unmarshal(xattrs[base.DocumentXattrKey], &retrievedVxattr))
+	require.NoError(t, json.Unmarshal(xattrs[base.VirtualXattrRevSeqNo], &retrievedVRevIDxattr))
 
-	rev, ok = retrievedVxattr["revid"].(string)
-	require.True(t, ok, "Unable to obtain revSeqNo as string")
-
-	revNo, err = strconv.ParseUint(rev, 10, 64)
+	revNo, err = strconv.ParseUint(retrievedVRevIDxattr, 10, 64)
 	require.NoError(t, err)
 
 	syncData, mou, _ = getSyncAndMou(t, collection, "sdkWrite")
--- a/db/background_mgr_resync_dcp_test.go
+++ b/db/background_mgr_resync_dcp_test.go
@@ -524,12 +524,14 @@ function sync(doc, oldDoc){
 	require.NotNil(t, syncData)
 	require.NotNil(t, mou)
 	require.Equal(t, base.CasToString(sgWriteCas), mou.PreviousCAS)
+	require.Equal(t, uint64(1), mou.PreviousRevID)
 
 	syncData, mou, _ = getSyncAndMou(t, collection, "sdkWrite")
 	require.NotNil(t, syncData)
 	require.NotNil(t, mou)
 	require.Equal(t, initialSDKMou.PreviousCAS, mou.PreviousCAS)
 	require.NotEqual(t, initialSDKMou.CAS, mou.CAS)
+	require.Equal(t, uint64(2), mou.PreviousRevID)
 }
 
 // helper function to Unmarshal BackgroundProcess state into ResyncManagerResponseDCP
--- a/db/background_mgr_resync_dcp_test.go
+++ b/db/background_mgr_resync_dcp_test.go
@@ -13,6 +13,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"strconv"
 	"sync"
 	"testing"
 	"time"
@@ -520,18 +521,41 @@ function sync(doc, oldDoc){
 	stats := getResyncStats(resyncMgr.Process)
 	assert.Equal(t, int64(2), stats.DocsChanged)
 
+	_, xattrs, _, err := collection.dataStore.GetWithXattrs(ctx, "sgWrite", []string{"$document"})
+	require.NoError(t, err)
+
+	var retrievedVxattr map[string]interface{}
+	require.NoError(t, json.Unmarshal(xattrs["$document"], &retrievedVxattr))
+
+	rev, ok := retrievedVxattr["revid"].(string)
+	assert.True(t, ok, "Unable to retrieve virtual xattr crc32c as string")
+	fmt.Println(rev)
+
+	revNo, err := strconv.ParseUint(rev, 10, 64)
+	require.NoError(t, err)
+
 	syncData, mou, _ = getSyncAndMou(t, collection, "sgWrite")
 	require.NotNil(t, syncData)
 	require.NotNil(t, mou)
 	require.Equal(t, base.CasToString(sgWriteCas), mou.PreviousCAS)
-	require.Equal(t, uint64(1), mou.PreviousRevID)
+	require.Equal(t, revNo-1, mou.PreviousRevID)
+
+	_, xattrs, _, err = collection.dataStore.GetWithXattrs(ctx, "sdkWrite", []string{"$document"})
+	require.NoError(t, err)
+	require.NoError(t, json.Unmarshal(xattrs["$document"], &retrievedVxattr))
+
+	rev, ok = retrievedVxattr["revid"].(string)
+	assert.True(t, ok, "Unable to retrieve virtual xattr crc32c as string")
+
+	revNo, err = strconv.ParseUint(rev, 10, 64)
+	require.NoError(t, err)
 
 	syncData, mou, _ = getSyncAndMou(t, collection, "sdkWrite")
 	require.NotNil(t, syncData)
 	require.NotNil(t, mou)
 	require.Equal(t, initialSDKMou.PreviousCAS, mou.PreviousCAS)
 	require.NotEqual(t, initialSDKMou.CAS, mou.CAS)
-	require.Equal(t, uint64(2), mou.PreviousRevID)
+	require.Equal(t, revNo-1, mou.PreviousRevID)
 }
 
 // helper function to Unmarshal BackgroundProcess state into ResyncManagerResponseDCP
--- a/db/background_mgr_resync_dcp_test.go
+++ b/db/background_mgr_resync_dcp_test.go
@@ -521,15 +521,14 @@ function sync(doc, oldDoc){
 	stats := getResyncStats(resyncMgr.Process)
 	assert.Equal(t, int64(2), stats.DocsChanged)
 
-	_, xattrs, _, err := collection.dataStore.GetWithXattrs(ctx, "sgWrite", []string{"$document"})
+	_, xattrs, _, err := collection.dataStore.GetWithXattrs(ctx, "sgWrite", []string{base.DocumentXattrKey})
 	require.NoError(t, err)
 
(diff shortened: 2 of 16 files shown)


db/background_mgr_resync_dcp_test.go
```diff
@@ -520,16 +521,41 @@ function sync(doc, oldDoc){
...
 	assert.Equal(t, int64(2), stats.DocsChanged)
 
+	_, xattrs, _, err := collection.dataStore.GetWithXattrs(ctx, "sgWrite", []string{"$document"})
+	require.NoError(t, err)
+
+	var retrievedVxattr map[string]interface{}
+	require.NoError(t, json.Unmarshal(xattrs["$document"], &retrievedVxattr))
+
+	rev, ok := retrievedVxattr["revid"].(string)
+	assert.True(t, ok, "Unable to retrieve virtual xattr crc32c as string")
```
```suggestion
	require.True(t, ok, "Unable to retrieve virtual xattr crc32c as string")
```

since the code will probably panic after.

What does crc32c mean here?

db/background_mgr_resync_dcp_test.go
```diff
@@ -520,16 +521,41 @@ function sync(doc, oldDoc){
...
 	require.NotNil(t, mou)
 	require.Equal(t, base.CasToString(sgWriteCas), mou.PreviousCAS)
+	require.Equal(t, revNo-1, mou.PreviousRevID)
+
+	_, xattrs, _, err = collection.dataStore.GetWithXattrs(ctx, "sdkWrite", []string{"$document"})
+	require.NoError(t, err)
+	require.NoError(t, json.Unmarshal(xattrs["$document"], &retrievedVxattr))
+
+	rev, ok = retrievedVxattr["revid"].(string)
+	assert.True(t, ok, "Unable to retrieve virtual xattr crc32c as string")
```
```suggestion
	require.True(t, ok, "Unable to retrieve virtual xattr crc32c as string")
```

db/background_mgr_resync_dcp_test.go
```diff
@@ -520,16 +521,41 @@ function sync(doc, oldDoc){
...
+
 	syncData, mou, _ = getSyncAndMou(t, collection, "sgWrite")
 	require.NotNil(t, syncData)
 	require.NotNil(t, mou)
 	require.Equal(t, base.CasToString(sgWriteCas), mou.PreviousCAS)
+	require.Equal(t, revNo-1, mou.PreviousRevID)
+
+	_, xattrs, _, err = collection.dataStore.GetWithXattrs(ctx, "sdkWrite", []string{"$document"})
+	require.NoError(t, err)
+	require.NoError(t, json.Unmarshal(xattrs["$document"], &retrievedVxattr))
```
torcolvin: in the way this test is written, if there isn't `revid` on `sdkWrite` then this test will pass because it reuses `retrievedXattr` from `sgWrite`.

In preference, I might resolve by:

- using a loop to test for `sdkWrite` and `sgWrite`
- creating a function to `GetWithXattrs` + unmarshal and return as uint
- reset `retrievedXattr` to nil between instances.

db/crud.go
```diff
@@ -245,7 +245,7 @@ func (c *DatabaseCollection) OnDemandImportForGet(ctx context.Context, docid str
 	importDb := DatabaseCollectionWithUser{DatabaseCollection: c, user: nil}
 	var importErr error
 
-	docOut, importErr = importDb.ImportDocRaw(ctx, docid, rawDoc, xattrs, isDelete, cas, nil, ImportOnDemand)
+	docOut, importErr = importDb.ImportDocRaw(ctx, docid, rawDoc, xattrs, isDelete, cas, nil, 0, ImportOnDemand)
```
torcolvin: I think this needs to have revSeqNo populated here for correctness. I wonder if you can add a test specifically for this?

db/database.go
```diff
@@ -1826,7 +1826,7 @@ func (db *DatabaseCollectionWithUser) getResyncedDocument(ctx context.Context, d
 	return doc, shouldUpdate, updatedExpiry, doc.Sequence, updatedUnusedSequences, nil
 }
 
-func (db *DatabaseCollectionWithUser) resyncDocument(ctx context.Context, docid, key string, regenerateSequences bool, unusedSequences []uint64) (updatedHighSeq uint64, updatedUnusedSequences []uint64, err error) {
+func (db *DatabaseCollectionWithUser) resyncDocument(ctx context.Context, docid, key string, regenerateSequences bool, unusedSequences []uint64, revNo uint64) (updatedHighSeq uint64, updatedUnusedSequences []uint64, err error) {
```
torcolvin: It seems weird to pass revNo here, don't we want to get revNo from `WriteUpdateWithXattrs` because this document could get modified between when this revSeqNo was passed in and when this is called? This might need an enhancement in sg-bucket.

db/document.go
```diff
@@ -63,8 +63,9 @@ type ChannelSetEntry struct {
 }
 
 type MetadataOnlyUpdate struct {
-	CAS         string `json:"cas,omitempty"`
-	PreviousCAS string `json:"pCas,omitempty"`
+	CAS           string `json:"cas,omitempty"`
+	PreviousCAS   string `json:"pCas,omitempty"`
+	PreviousRevID uint64 `json:"pRev,omitempty"`
```
torcolvin: Let's label this PreviousRevSeqNo and document this as a server value. I think that we should use `revSeqNo`. It's particularly terrible that `$document.revid` is used.

rest/importtest/import_test.go
```diff
@@ -2378,3 +2379,75 @@ func TestImportUpdateExpiry(t *testing.T) {
...
+	mobileBody := make(map[string]interface{})
+	mobileBody["channels"] = "ABC"
+	_, err := dataStore.Add(mobileKey, 0, mobileBody)
+	assert.NoError(t, err, "Error writing SDK doc")
+
+	// Wait for import
+	err = rt.WaitForCondition(func() bool {
+		return rt.GetDatabase().DbStats.SharedBucketImportStats.ImportCount.Value() == 1
+	})
+	require.NoError(t, err)
```
```suggestion
	base.RequireWaitForStat(t, func() int64 {
		return rt.GetDatabase().DbStats.SharedBucketImportStats.ImportCount.Value()
	}, 0)
```

This will print the value if it doesn't match.

rest/importtest/import_test.go
```diff
@@ -2378,3 +2379,75 @@ func TestImportUpdateExpiry(t *testing.T) {
...
+
+	xattrs, _, err := dataStore.GetXattrs(ctx, mobileKey, []string{base.MouXattrName})
+	require.NoError(t, err)
+	if rt.GetDatabase().UseMou() {
+		var mou *db.MetadataOnlyUpdate
+		mouXattr, ok := xattrs[base.MouXattrName]
+		require.True(t, ok)
+		require.NoError(t, base.JSONUnmarshal(mouXattr, &mou))
+		// curr rev no should be 2, so prev rev is 1
+		assert.Equal(t, uint64(1), mou.PreviousRevID)
```
torcolvin: I am really surprised the value of this is 1 at all, I would have expected that the value would be a higher number because my understanding was that this represents all documents in a bucket, and we have all the config documents in a bucket.

Additionally, I don't think that this test will work correctly when run with other tests at the same time, becuase we do not flush the bucket between tests, so other tests will pick up the revseqno first?

I think it might be correct to write the document _before_ the database is created, and get xattrs from `$document.revid`?

rest/importtest/import_test.go
```diff
@@ -2378,3 +2379,75 @@ func TestImportUpdateExpiry(t *testing.T) {
...
+		require.Error(t, err)
+	}
+
+	err = dataStore.Set(mobileKey, 0, nil, []byte(`{"test":"update"}`))
+	require.NoError(t, err)
+
+	err = rt.WaitForCondition(func() bool {
+		return rt.GetDatabase().DbStats.SharedBucketImportStats.ImportCount.Value() == 2
+	})
+	require.NoError(t, err)
```
```suggestion
	base.RequireWaitForStat(t, func() int64 {
		return rt.GetDatabase().DbStats.SharedBucketImportStats.ImportCount.Value()
	}, 2)
	require.NoError(t, err)
```

db/import_test.go
```diff
@@ -709,7 +709,7 @@ func TestImportStampClusterUUID(t *testing.T) {
 	require.NoError(t, err)
 	existingDoc := &sgbucket.BucketDocument{Body: bodyBytes, Cas: cas}
 
-	importedDoc, err := collection.importDoc(ctx, key, body, nil, false, existingDoc, ImportOnDemand)
+	importedDoc, err := collection.importDoc(ctx, key, body, nil, false, 0, existingDoc, ImportOnDemand)
```
torcolvin: I think it'd be good to use the real revSeqNo here.

db/import_test.go
```diff
@@ -609,7 +609,7 @@ func TestImportNullDocRaw(t *testing.T) {
 	xattrs := map[string][]byte{
 		base.SyncXattrName: []byte("{}"),
 	}
-	importedDoc, err := collection.ImportDocRaw(ctx, "TestImportNullDoc", []byte("null"), xattrs, false, 1, &exp, ImportFromFeed)
+	importedDoc, err := collection.ImportDocRaw(ctx, "TestImportNullDoc", []byte("null"), xattrs, false, 1, &exp, 0, ImportFromFeed)
```
torcolvin: This test I think doesn't have a revSeqNo, because the null doc never really exists? But maybe I'm reading this wrong.

db/import.go
```diff
@@ -32,7 +32,7 @@ const (
 )
 
 // Imports a document that was written by someone other than sync gateway, given the existing state of the doc in raw bytes
-func (db *DatabaseCollectionWithUser) ImportDocRaw(ctx context.Context, docid string, value []byte, xattrs map[string][]byte, isDelete bool, cas uint64, expiry *uint32, mode ImportMode) (docOut *Document, err error) {
+func (db *DatabaseCollectionWithUser) ImportDocRaw(ctx context.Context, docid string, value []byte, xattrs map[string][]byte, isDelete bool, cas uint64, expiry *uint32, revNo uint64, mode ImportMode) (docOut *Document, err error) {
```
torcolvin: Idea, you can take or leave:

create `importDocOptions` that has `expiry`, `revSeqNo`,  `mode`, `isDelete` which shortens the number of args by 3 for `ImportDocRaw` and `importDoc`. It sort of galls me that the arguments are not in the same order between the two functions.

torcolvin (commented):
I think there's a fundamental race in `importDoc` - if the document is updated in the meantime, we have to reset the expiry https://github.com/couchbase/sync_gateway/blob/51c00da4c789fed54e31ad9d512659420dee73f1/db/import.go#L158

Some comments here are on code style, but others are on this fundamental problem of not necessarily having the latest cas (in resync or import on demand, which get the latest version of the doc).

I haven't thought deeply enough whether the sgbucket apis will need modification for this, but I think if they do we could consider populating this value in rosmar to make the testing easier. I think in rosmar for now that revseqno could be equal to cas because there's no concept of vbuckets.

base/constants.go
```diff
@@ -144,6 +144,8 @@ const (
 	// Intended to be used in Meta Map and related tests
 	MetaMapXattrsKey = "xattrs"
 
+	DocumentXattrKey = "$document"
```
```suggestion
	VirtualXattrKey = "$document"
```

But I wonder if we can just pull `RevSeqNoXattrKey` and use `$document.revid` here, and scope this more narrowly.

db/database_collection.go
```diff
@@ -249,7 +249,7 @@ func (c *DatabaseCollection) syncAndUserXattrKeys() []string {
 
 // syncMouAndUserXattrKeys returns the xattr keys for the user, mou and sync xattrs.
 func (c *DatabaseCollection) syncMouAndUserXattrKeys() []string {
```
torcolvin: The name of this function is no longer correct.

db/database_collection.go
```diff
@@ -249,7 +249,7 @@ func (c *DatabaseCollection) syncAndUserXattrKeys() []string {
...
-	xattrKeys := []string{base.SyncXattrName, base.VvXattrName}
+	xattrKeys := []string{base.SyncXattrName, base.VvXattrName, base.DocumentXattrKey}
 	if c.useMou() {
 		xattrKeys = append(xattrKeys, base.MouXattrName)
 	}
```
torcolvin: I think the intent behind this if statement was to have _only_ two xattrs in pre 7.6 when we can do multi subdoc operations https://github.com/couchbase/sync_gateway/blob/2f5dcd6585140f9dfb6d5fae8fc15d772f2e6ff6/base/collection_xattr.go#L196

It reads weirdly to me now. I think on anemone we require 7.6. If we don't check for 7.6 on startup I think that we should (Out of scope for this PR) and then we can drop the check inside this function.

db/document.go
```diff
@@ -1114,6 +1121,20 @@ func (doc *Document) UnmarshalWithXattrs(ctx context.Context, data, syncXattrDat
 				return pkgerrors.WithStack(base.RedactErrorf("Failed to unmarshal HLV during UnmarshalWithXattrs() doc with id: %s (DocUnmarshalAll/Sync).  Error: %v", base.UD(doc.ID), err))
 			}
 		}
+		if documentXattr != nil {
```
torcolvin, line 1120: I wonder if we can/should make this required. You'd have to implement this in rosmar but it feels like the code isn't going to behave correctly without this present.

db/document.go
```diff
@@ -1114,6 +1121,20 @@ func (doc *Document) UnmarshalWithXattrs(ctx context.Context, data, syncXattrDat
...
+			var docXattr ServerRevSeqNoXattr
+			err := base.JSONUnmarshal(documentXattr, &docXattr)
+			if err != nil {
+				return pkgerrors.WithStack(base.RedactErrorf("Failed to unmarshal doc xattr during UnmarshalWithXattrs() doc with id: %s (DocUnmarshalAll/Sync).  Error: %v", base.UD(doc.ID), err))
+			}
+			if docXattr.RevSeqNo != "" {
+				revNo, err := strconv.ParseUint(docXattr.RevSeqNo, 10, 64)
+				if err != nil {
+					return pkgerrors.WithStack(base.RedactErrorf("Failed convert rev seq number during UnmarshalWithXattrs() doc with id: %s (DocUnmarshalAll/Sync).  Error: %v", base.UD(doc.ID), err))
```
```suggestion
					return pkgerrors.WithStack(base.RedactErrorf("Failed convert rev seq number %q during UnmarshalWithXattrs() doc with id: %s (DocUnmarshalAll/Sync).  Error: %v", docXattr.RevSeqNo, base.UD(doc.ID), err))
```

db/document.go
```diff
@@ -1092,7 +1099,7 @@ func (doc *Document) MarshalJSON() (data []byte, err error) {
 // unmarshalLevel is anything less than the full document + metadata, the raw data is retained for subsequent
 // lazy unmarshalling as needed.
 // Must handle cases where document body and hlvXattrData are present without syncXattrData for all DocumentUnmarshalLevel
-func (doc *Document) UnmarshalWithXattrs(ctx context.Context, data, syncXattrData, hlvXattrData []byte, unmarshalLevel DocumentUnmarshalLevel) error {
+func (doc *Document) UnmarshalWithXattrs(ctx context.Context, data, syncXattrData, hlvXattrData, documentXattr []byte, unmarshalLevel DocumentUnmarshalLevel) error {
```
torcolvin, line 1098: Do we need to add `RevSeqNo` for other sections of `DocumentUnmarshalLevel` - I think maybe in `DocUnmarshalRev` for `PutExistingCurrentVersion` if `existingDoc == nil`

rest/importtest/import_test.go
```diff
@@ -2378,3 +2379,79 @@ func TestImportUpdateExpiry(t *testing.T) {
...
+	err = dataStore.Set(mobileKey, 0, nil, []byte(`{"test":"update"}`))
+	require.NoError(t, err)
+
+	base.RequireWaitForStat(t, func() int64 {
+		return rt.GetDatabase().DbStats.SharedBucketImportStats.ImportCount.Value()
+	}, 2)
+
+	xattrs, _, err = dataStore.GetXattrs(ctx, mobileKey, []string{base.MouXattrName, base.DocumentXattrKey})
+	require.NoError(t, err)
+	if rt.GetDatabase().UseMou() {
```
torcolvin: I thought we require 7.6 for anemone, see above?

db/database_collection.go
gregns1: Just for correctness I have moved the virtual xattr inside this check.

db/document.go
gregns1: I'm not sure we do? Currently we only set the prev rev seq no inside `computeMetadataOnlyUpdate` which is only called from `importDoc` and `resyncDocument`. I know on demand import for wrote is called inside the callback potentially if import is needed but that would fall into the on demand import ticket

db/document.go
gregns1: I'll leave this to be done as part of the rosmar ticket. I think I was just following pattern in the function currently for other xattrs.

rest/importtest/import_test.go
```diff
@@ -2378,3 +2379,71 @@ func TestImportUpdateExpiry(t *testing.T) {
 		})
 	}
 }
+
+func TestPrevRevNoPopulationImportFeed(t *testing.T) {
+	base.SkipImportTestsIfNotEnabled(t)
+	if base.UnitTestUrlIsWalrus() {
+		t.Skipf("test requires CBS for previous rev no assertion")
```
```suggestion
		t.Skipf("test requires CBS for previous rev no assertion, CBG-4233")
```

db/util_testing.go
```diff
@@ -877,3 +878,18 @@ func (c *DatabaseCollection) GetDocumentCurrentVersion(t testing.TB, key string)
 	}
 	return doc.HLV.SourceID, doc.HLV.Version
 }
+
+// retrieveDocRevSeNo will take the $document xattr and return the revSeqNo defined in that xattr
+func RetrieveDocRevSeqNo(t *testing.T, docxattr []byte) uint64 {
+	// virtual xattr not implemented for rosmar
```
torcolvin:
```suggestion
	// virtual xattr not implemented for rosmar CBG-4233
```

torcolvin (dismissed):
LGTM.

torcolvin (approved):

PR #7103, opened 2024-09-03
CBG-4226 remove allOf references

torcolvin:
This is a temporary backstop to generate docs for 3.2, pending a resolution of https://jira.issues.couchbase.com/browse/DOC-12510. That does not sound like it will occur imminently.

In addition to unraveling code, I fixed `enabled` defaults for `error`, `warn`, `info`, `stats`. There's not really a reason this can't stay long term.

To look at the code:

`npx @redocly/cli preview-docs admin` 

and make sure the endpoints are correct for db config "logging" sections, db/_config/audit.

### DIFF
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -1847,20 +1847,42 @@ Database:
               example: ["CRUD", "HTTP", "Query"]
         audit:
           description: Audit logging configuration.
-          allOf:
-            - type: object
-              properties:
-                enabled:
-                  description: Whether audit logging is enabled.
-                  type: boolean
-                  default: false
-                enabled_events:
-                  description: List of enabled audit events for this database.
-                  type: array
-                  items:
-                    type: number
-                  example: [1234, 5678]
-            - $ref: '#/Disabled-users-and-roles'
+          type: object
+          properties:
+            enabled:
+              description: Whether audit logging is enabled.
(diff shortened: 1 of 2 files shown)


docs/api/components/schemas.yaml
```diff
@@ -1884,10 +1884,36 @@ Database:
                     description: The name of the role for which audit logging is disabled.
                     type: string
   title: Database-config
+Disabled-users-and-roles:
```
adamcfraser, line 1897: I don't see where this group is being referenced.

docs/api/components/schemas.yaml
```diff
@@ -1858,21 +1858,21 @@ Database:
...
             disabled_users:
-              description: List of users for which audit logging is disabled
+              description: List of users for which audit logging is disabled.
               type: array
               items:
                 type: object
                 properties:
                   domain:
-                    description: The domain of the user for which audit logging is disabled.
+                    description: The domain of the user for which audit logging is disabled. Either cbs or sgw.
```
adamcfraser: Should the cbs/sgw be documented as enum or is that overkill/unnecessary?  I also wonder whether end users will know what 'cbs' and sgw' mean.

adamcfraser (changes requested):

adamcfraser (approved):

PR #7111, opened 2024-09-06
CBG-4211: Background task dcp client support for rosmar 

gregns1:
CBG-4211

- New function NewBackgroundManagerDcpClient that will either create a rosmar dcp client or a server dcp client based on the bucket type you pass into the function 
- Enabled all resync tests
- Attachment compaction is not supported at this stage, needs work in [CBG-4232](https://jira.issues.couchbase.com/browse/CBG-4232) 
- Needed some changes to fix race flag test failures. 1. we were setting vbUUIDs on dcp resync manager without a lock, and 2. we were getting a race between setting collection context on ctx variable in callback and the log lines in the run function reading this value. To fix this I moved setting vb uuids into separate function with lock and renames collection ctx variable in callback to not clash with ctx variable used outside the callback.
- Another race condition found in test TestResyncManagerDCPStopInMidWay, test was setting wait for condition error to variable err inside a goroutine, and then rest of test was also using same variable err. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2680/

[CBG-4232]: https://couchbasecloud.atlassian.net/browse/CBG-4232?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/db/background_mgr_dcp_client.go
+++ b/db/background_mgr_dcp_client.go
@@ -19,15 +19,15 @@ import (
 )
 
 type backgroundMgrDcpClient interface {
+	// Start returns an error and a channel to indicate when the DCPClient is done. If Start returns an error, DCPClient.Close() needs to be called.
 	Start() (doneChan chan error, err error)
 	// Close is used externally to stop the DCP client. If the client was already closed due to error, returns that error
 	Close() error
 	// GetMetadata returns metadata for all vbuckets
 	GetMetadata() []base.DCPMetadata
-	// GetMetadataKeyPrefix returns the dcp metadata key prefix
-	GetMetadataKeyPrefix() string
 }
 
+// backgroundManagerDcpClientOptions are the dcp client options needed for creation of a background task dcp client
 type backgroundManagerDcpClientOptions struct {
 	Callback sgbucket.FeedEventCallbackFunc
 	// only used by rosmar
@@ -37,21 +37,23 @@ type backgroundManagerDcpClientOptions struct {
 	FailOnRollback    bool                      // When true, the DCP client will terminate on DCP rollback
 	MetadataStoreType base.DCPMetadataStoreType // define storage type for DCPMetadata
 	GroupID           string                    // specify GroupID, only used when MetadataStoreType is DCPMetadataCS
-	CheckpointPrefix  string
-	CollectionIDs     []uint32 // CollectionIDs used by gocbcore, if empty, uses default collections
-	OneShot           bool     // Whether the dcp feed will be continuous or not
-	InitialMetadata   []base.DCPMetadata
+	CheckpointPrefix  string                    // Prefix used by
+	CollectionIDs     []uint32                  // CollectionIDs used by gocbcore, if empty, uses default collections
+	OneShot           bool                      // Whether the dcp feed will be continuous or not
 }
 
+// NewBackgroundManagerDcpClient create a new DCP client for background tasks. From the bucket you pass in it will either
+// create a rosmar dcp client or a gocb dcp client
 func NewBackgroundManagerDcpClient(ctx context.Context, bucket base.Bucket, options backgroundManagerDcpClientOptions) (backgroundMgrDcpClient, error) {
 	gocbBucket, err := base.AsGocbV2Bucket(bucket)
 	if err == nil {
 		clientOptions := &base.DCPClientOptions{
+			OneShot:           options.OneShot,
 			FailOnRollback:    options.FailOnRollback,
 			MetadataStoreType: options.MetadataStoreType,
 			GroupID:           options.GroupID,
 			CheckpointPrefix:  options.CheckpointPrefix,
-			InitialMetadata:   options.InitialMetadata,
+			CollectionIDs:     options.CollectionIDs,
 		}
 		return base.NewDCPClient(ctx, options.ID, options.Callback, *clientOptions, gocbBucket)
 	}
@@ -82,6 +84,7 @@ type rosmarDcpClient struct {
 	doneChan chan error
 }
 
+// Start returns an error and a channel to indicate when the DCPClient is done. If Start returns an error, DCPClient.Close() needs to be called.
 func (r *rosmarDcpClient) Start() (doneChan chan error, err error) {
 	err = r.bucket.StartDCPFeed(r.ctx, r.feedArgs, r.callback, nil)
 	go func() {
@@ -95,6 +98,7 @@ func (r *rosmarDcpClient) Start() (doneChan chan error, err error) {
 	return r.doneChan, err
 }
 
+// Close is used externally to stop the DCP client. If the client was already closed due to error, returns that error
 func (r *rosmarDcpClient) Close() error {
 	close(r.feedArgs.Terminator)
 	return nil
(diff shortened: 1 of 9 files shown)


db/background_mgr_dcp_client.go
```diff
@@ -0,0 +1,110 @@
...
+	"context"
+	"fmt"
+
+	"github.com/couchbase/sync_gateway/base"
+	"github.com/couchbaselabs/rosmar"
+
+	sgbucket "github.com/couchbase/sg-bucket"
+)
+
+type backgroundMgrDcpClient interface {
```
adamcfraser, line 21: Although we're using this for background mgr in the initial case, isn't this a generic interface for a DCP client (implemented by either gocb or rosmar)?  Is there something specific to background manager usage?

db/background_mgr_dcp_client.go
```diff
@@ -0,0 +1,110 @@
...
+	// Start returns an error and a channel to indicate when the DCPClient is done. If Start returns an error, DCPClient.Close() needs to be called.
+	Start() (doneChan chan error, err error)
+	// Close is used externally to stop the DCP client. If the client was already closed due to error, returns that error
+	Close() error
+	// GetMetadata returns metadata for all vbuckets
+	GetMetadata() []base.DCPMetadata
+}
+
+// backgroundManagerDcpClientOptions are the dcp client options needed for creation of a background task dcp client
+type backgroundManagerDcpClientOptions struct {
```
adamcfraser, line 31: It doesn't feel correct for the options to have two different ways to specify the set of collections to stream (one for rosmar, one for gocb).  It looks like consumers of DCPClient are already calling getCollectionIdsAndNames to build the set of IDs based on information on the database.  How do we handle usage of DatabaseContext.CollectionByID for other rosmar tests?  Is it just not used?

db/background_mgr_dcp_client.go
```diff
@@ -0,0 +1,110 @@
...
+
+// Close is used externally to stop the DCP client. If the client was already closed due to error, returns that error
+func (r *rosmarDcpClient) Close() error {
+	close(r.feedArgs.Terminator)
+	return nil
+}
+
+// GetMetadata returns metadata for all vbuckets
+func (dc *rosmarDcpClient) GetMetadata() []base.DCPMetadata {
+	return nil
```
adamcfraser, line 109: I thought GetMetadata was used as part of stop/pause/resume for the gocb DCP feed.  If that's being done through MetadataStoreType, is GetMetadata needed as part of the API?

adamcfraser (changes requested):

db/background_mgr_dcp_client.go
torcolvin: This style of implementation was mine and not Greg's.

It's a limited version of a DCPClient - I wanted to do something like make a `base.DCPClient` interface and have different backing and create the existing class `base.DCPClient` to `base.GocbDCPClient` - I ended up having a lot of trouble doing this for a few reasons. I'd aspire to make this possible, and especially to create an abstracted class that could use cbgt, gocb.DcpAgent, or rosmar's DCPAgent.

I wanted to do this by allowing the existing callers of `base.DCPClient` to be used in either case but there were definitely some things that were hard to support in a clean way:

- rosmar has its own concept of metadata which is a single document with a cas values. Correctly, rosmar doesn't really have vBuckets, but it does implement some APIs to have 16 (https://github.com/couchbaselabs/rosmar/blob/c856107e8e78d4bce8d866ed1f2dd4a583309eed/bucket_api.go#L103). We could definitely have opaque metadata for a rosmar implementation but only use it for specifically testing  a gocbcore DCP implementation.
- we have some test code that injects different metadata in place in areas that aren't testing the base implementation.
  - https://github.com/couchbase/sync_gateway/blob/d9a47e914ffa1dc7f995213339de2c4bb085bd52/db/attachment_compaction.go#L363-L365
  -https://github.com/couchbase/sync_gateway/blob/d9a47e914ffa1dc7f995213339de2c4bb085bd52/rest/adminapitest/resync_test.go#L60
- there is not a concept of in memory persistence for rosmar, other than all databases are in memory - the checkpoints of prefixes. I think we might want to address this for a general concept.
- The biggest area I spent time trying to to reconcile was around storage of `VbUUIDs` in attachment compaction https://github.com/couchbase/sync_gateway/blob/main/db/background_mgr_attachment_compaction.go#L153. This PR doesn't attempt to address this. I think that the use vbuuids here is erroneous? It uses the vbuuids that it generates from all metadata https://github.com/couchbase/sync_gateway/blob/main/db/attachment_compaction.go#L188 and then builds up a slice https://github.com/couchbase/sync_gateway/blob/main/db/attachment_compaction.go#L364 to reset the checkpoints to 0 and running over everything?
- I tried to have this use `DCPWorker` to exercise some of that code and have the implementation be closer aligned to our existing interface. I don't think this is necessary.

When I initially wrote this test code, I wanted it work for resync, attachment compaction, and other one shot DCP usages.

I think the value proposition for support attachment compaction is lower than higher testable use cases.

I still would like a common DCPClient interface (and we can cast it for test cases, if we _do_ want to be able to run something), but I wasn't at all sure that I should put more work into it at this time. I'm not sure what the answer is here.

db/background_mgr_dcp_client.go
torcolvin: I think this is perhaps an easier cleanup and came because the initial implementation of `DCPClient` and the upgrade from gocb v1 to v2 happened before collection support was added.

For `DCPClient`, the collection ids are assembled https://github.com/couchbase/sync_gateway/blob/d9a47e914ffa1dc7f995213339de2c4bb085bd52/base/gocb_dcp_feed.go#L58 but I think this code could pass in `Scopes           map[string][]string // Collection names to stream - map keys are scopes` to `DCPClientOptions` and we wouldn't have to do specify this twice and can just move the determination of collection IDs (needed for gocbcore.OpenStreamFilterOptions.CollectionIDs).

db/background_mgr_dcp_client.go
torcolvin: It's used https://github.com/couchbase/sync_gateway/blob/d9a47e914ffa1dc7f995213339de2c4bb085bd52/db/background_mgr_resync_dcp.go#L183 but I do not think this is actually _used_ and it was copied from the attachment code (where it isn't in fact used). I have some evidence of this because `ResyncManagerDCP` literally calls `AttachmentManagerMeta` https://github.com/couchbase/sync_gateway/blob/d9a47e914ffa1dc7f995213339de2c4bb085bd52/db/background_mgr_resync_dcp.go#L350 and I don't think it _consumes_ this value at any point. Resuming resync works because it has a common set of checkpoints. We can't even do anything with only vbuuids.

While I do not think this value is used, I think we might have to write this value for backward compatibility. I haven't considered this.

Yet again, this might not be the right time to look at this, but there might never be a good time.

I think there's some value to putting this into 3.3 for increased testability, but that might imply 3.2.1 code, and this change might be bigger there. I do feel OK about no-oping this for rosmar for sure.

PR #7112, opened 2024-09-09
CBG-4253 create interfaces for integration testing

torcolvin:
These tests are intended to support https://docs.google.com/document/d/1oiPwTNKkdBkWbACvwEBsc3EbVD-PCTO1s6WCNvtwY7k work for version vectors, but are general enough to support any type of testing that requires replication.

This package lives outside the rest package since it doesn't necessarily use a REST API at all (it can for 1.x replications) and this allows easy skipping of these tests. Even in a minimal form, they take 75 secs to run under Couchbase Server and 5 seconds for rosmar.

A few test bugs were fixed:

- bidirectional XDCR with couchbase server didn't work because we accidentally would delete an xdcr cluster from under a running XDCR replication, since setting up a bidirectional replication requires two calls to `controller/createReplication`. Creating two unidirectional replications would cause one to fail.
- rosmar xdcr wasn't doing backfill. This probably will be enhanced with future PRs to do checkpointing, but was needed to have the tests pass.
- made `blip_client_test.go` code public in the `rest` package to facilitate moving to a separate package. Minor modifications were needed to allow calling `NewBlipTesterClientOptsWithRT` outside of `btcRunner.Run`, since the tests are not conceptually run under a blip tester, blip testing just happens to be included. The check for `btcRunner.Run` is used to be able to run with `CBMobileReplicationV4` and `CBMobileReplicationV3` in a loop correctly. With these tests, having `CBMobileReplicationV3` would probably be a separate flag for a couchbase peer. A single test can have multiple CouchbaseLitePeers unlike standard testing so creating `AllowCreationWithoutBlipTesterClientRunner` is used only for these tests.

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2714/


topologytest/topologies.go
```diff
@@ -0,0 +1,314 @@
...
+			      v
+			  +---------+
+			  |   cbl1  |
+			  +---------+
+		*/
+		description: "CBL <-> Sync Gateway <-> CBS",
+		peers: map[string]PeerOptions{
+			"cbs1": {Type: PeerTypeCouchbaseServer, BucketID: PeerBucketID1},
+			"sg1":  {Type: PeerTypeSyncGateway, BucketID: PeerBucketID1},
+			"cbl1": {Type: PeerTypeCouchbaseLiteMock, BucketID: PeerBucketID1},
```
adamcfraser: Can/should BucketID be omitted for CBL peers?

adamcfraser: I expected this to be PeerTypeCouchbaseLite (and CouchbaseLiteMock would be one possible implementation of that peer).

topologytest/sync_gateway_peer.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func newSyncGatewayPeer(t *testing.T, name string, bucket *base.TestBucket) Peer {
+	rt := rest.NewRestTester(t, &rest.RestTesterConfig{
+		PersistentConfig: true,
+		CustomTestBucket: bucket.NoCloseClone(),
+	})
+	config := rt.NewDbConfig()
+	config.AutoImport = base.BoolPtr(true)
+	rest.RequireStatus(t, rt.CreateDatabase("db", config), http.StatusCreated)
```
adamcfraser: Should database name be `name` instead of "db", or is db required?

adamcfraser (changes requested):

adamcfraser (approved):

PR #7113, opened 2024-09-10
CBG-4234: clean up rev cache work

gregns1:
CBG-4234

Found some issues with rev cache when playing wit hit locally:
- When inserting a doc to empty rev cache that will be evicted by memory immediately, sync gateway will panic as we wern't populating the rev cache value with the memory measurement till after we call to perform memory based rev cache eviction. Meaning the iteration that the rev cache uses to remove items from cache would iterate too many times and hit empty rev cache and panic when trying to reference rev cache value to empty rev cache. 
- Secondly found that num items stat wasn't having correct after performing memory based eviction. We weren't decrementing as expected. 
- Lastly I found that we weren't re setting the num items and memory stat when a new rev cache is created. 
- I have added test coverage of the above

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2682/


bbrks (approved):
Looks good 👍

PR #7117, opened 2024-09-16
CBG-4207: Attachment metadata migration on import 

gregns1:
CBG-4207

- Added new function `MigrateAttachmentMetadata` that will move any defined attachments from sync data into the new global sync xattr and remove it from sync data
- Two new tests that test on demand import moving the attachment metadata and the import feed moving it
- Test for import feed tests a doc that needs importing moving the metadata and a doc that is deemed as not needing to be imported migrating metadata
- Define a new function for test purposes that moved global sync data attachment metadata to sync data attachment metadata 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2691/

### DIFF
--- a/db/import_listener.go
+++ b/db/import_listener.go
@@ -220,8 +220,7 @@ func (il *importListener) ImportFeedEvent(ctx context.Context, collection *Datab
 				base.DebugfCtx(ctx, base.KeyImport, "Did not import doc %q - external update will not be accessible via Sync Gateway.  Reason: %v", base.UD(docID), err)
 			}
 		}
-	}
-	if syncData != nil && syncData.Attachments != nil {
+	} else if syncData != nil && syncData.Attachments != nil {
 		base.DebugfCtx(ctx, base.KeyImport, "Attachment metadata found in sync data for doc with id %s, migrating attachment metadata", base.UD(docID))
 		// we have attachments to migrate
 		err := collection.MigrateAttachmentMetadata(ctx, docID, event.Cas, syncData)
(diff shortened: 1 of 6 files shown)


db/import_listener.go
```diff
@@ -221,6 +221,14 @@ func (il *importListener) ImportFeedEvent(ctx context.Context, collection *Datab
...
 		}
 	}
+	if syncData != nil && syncData.Attachments != nil {
+		base.DebugfCtx(ctx, base.KeyImport, "Attachment metadata found in sync data for doc with id %s, migrating attachment metadata", base.UD(docID))
+		// we have attachments to migrate
+		err := collection.MigrateAttachmentMetadata(ctx, docID, event.Cas, syncData)
+		if err != nil {
+			base.WarnfCtx(ctx, "error migrating attachment metadata from sync data to global sync for doc %s. Error: %v", base.UD(docID), err)
+		}
+	}
```
bbrks: I am curious to see if we can do this as part of the write being done by `ImportDocRaw` to save the the double write to the bucket in the actual normal import case?
Or is it already effectively done as part of doc write, and this is covering the non-import codepath? If so, should this be in an else?

gregns1: Good spot, this should be in an else. Will make the change.

bbrks: That makes sense! Thanks!

bbrks (approved):

PR #7118, opened 2024-09-19
CBG-3861 support updating vv on xdcr

torcolvin:
- this code will only update cv when appropriate, does not yet support changing pv or mv (not supported in mobile xdcr anyway)
- stubs out some kinds of tests we can write
- Creates a `SourceID` for rosmar that is just bucketUUID (clusterUUID is always empty string), but this will be unique since rosmar is only in memory
- replace `cblEncodeTestSources` with `EncodeTestHistory` since these were duplicate functions and I was about to write a third.

Parts of this PR I'm particularly interested in are whether the test code is easily readable. I gave some limited thought to how to parameterize `TestVVWriteOnDestAfterInitialReplication` and `TestVVWriteTwice` like `TestReplicateVV`, but I couldn't think of something that wasn't like:

- test.preXDCRFunc1
- run xdcr
- test.assertVV1
- test.preXDCRFunc2
- test.assertVV2

I didn't feel like this was more readable and generalizable enough. I am imagining that we have a different set of test cases for the future when we implement `pv` support that doesn't take a datastore, but rather has two `DatabaseContext` instances which would allow us to use the crud operations on `DatabaseContext` and also creating imports. When implementing this, I was going to debate if a structure like above was appropriate.

Ultimately, I just decided to stop working on this to be able to get some code in place and be able to break tasks up into smaller pieces. I do think it's worthwhile to get this in good shape and any significant improvements, I'll break into separate tickets.

### DIFF
--- a/xdcr/rosmar_xdcr.go
+++ b/xdcr/rosmar_xdcr.go
@@ -31,7 +31,7 @@ type rosmarManager struct {
 	fromBucketSourceID  string
 	toBucket            *rosmar.Bucket
 	replicationID       string
-	docsFiltered        atomic.Uint64
+	mobileDocsFiltered  atomic.Uint64
 	docsWritten         atomic.Uint64
 	errorCount          atomic.Uint64
 	targetNewerDocs     atomic.Uint64
@@ -74,7 +74,7 @@ func (r *rosmarManager) processEvent(ctx context.Context, event sgbucket.FeedEve
 		// Filter out events if we have a non XDCR filter
 		if r.filterFunc != nil && !r.filterFunc(&event) {
 			base.TracefCtx(ctx, base.KeyWalrus, "Filtering doc %s", docID)
-			r.docsFiltered.Add(1)
+			r.mobileDocsFiltered.Add(1)
 			return true
 		}
 
@@ -247,13 +247,14 @@ func opWithMeta(ctx context.Context, collection *rosmar.Collection, sourceID str
 
 // Stats returns the stats of the XDCR replication.
 func (r *rosmarManager) Stats(context.Context) (*Stats, error) {
-
-	return &Stats{
-		DocsWritten:     r.docsWritten.Load(),
-		DocsFiltered:    r.docsFiltered.Load(),
-		ErrorCount:      r.errorCount.Load(),
-		TargetNewerDocs: r.targetNewerDocs.Load(),
-	}, nil
+	stats := &Stats{
+		DocsWritten:        r.docsWritten.Load(),
+		MobileDocsFiltered: r.mobileDocsFiltered.Load(),
+		ErrorCount:         r.errorCount.Load(),
+		TargetNewerDocs:    r.targetNewerDocs.Load(),
+	}
+	stats.DocsProcessed = stats.DocsWritten + stats.MobileDocsFiltered + stats.TargetNewerDocs
+	return stats, nil
 }
 
 // xattrToBytes converts a map of xattrs of marshalled json.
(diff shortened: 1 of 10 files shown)


xdcr/xdcr_test.go
```diff
@@ -50,74 +52,299 @@ func TestMobileXDCRNoSyncDataCopied(t *testing.T) {
...
+			preXDCRFunc: func(t *testing.T, docID string) uint64 {
+				_, err := toDs.WriteCas(docID, 0, 0, []byte(`{"datastore":"toDs"}`), 0)
+				require.NoError(t, err)
+				cas, err := fromDs.WriteCas(docID, 0, 0, []byte(`{"datastore":"fromDs"}`), 0)
+				require.NoError(t, err)
+				return cas
+			},
+		},
+		{
+			name:  "dest doc newer, expect keep old version",
```
adamcfraser: This says dest doc newer, but looks like fromDs gets written to last?

xdcr/xdcr_test.go
```diff
@@ -50,74 +52,299 @@ func TestMobileXDCRNoSyncDataCopied(t *testing.T) {
...
-					var vvXattrVal map[string]any
-					require.NoError(t, base.JSONUnmarshal(vvXattrBytes, &vvXattrVal))
-					assert.NotNil(c, vvXattrVal[version])
-					assert.NotNil(c, vvXattrVal[source])
-					assert.NotNil(c, vvXattrVal[curCAS])
-
-				}, time.Second*5, time.Millisecond*100)
-			}
-		}
+func TestReplicateVV(t *testing.T) {
```
adamcfraser, line 149: This test looks like what's it's actually testing is XDCR's LWW logic?  Should it be renamed TestReplicateLWW?

xdcr/xdcr_test.go
```diff
@@ -50,74 +52,299 @@ func TestMobileXDCRNoSyncDataCopied(t *testing.T) {
...
+	require.JSONEq(t, ver1Body, string(body))
+	require.Contains(t, xattrs, base.VvXattrName)
+	casString := string(base.Uint64CASToLittleEndianHex(fromCAS))
+	var vv *db.HybridLogicalVector
+	require.NoError(t, base.JSONUnmarshal(xattrs[base.VvXattrName], &vv))
+	require.Equal(t, &db.HybridLogicalVector{
+		CurrentVersionCAS: casString,
+		SourceID:          fromBucketSourceID,
+		Version:           casString,
+	}, vv)
```
adamcfraser: This seems like commonly reused code.  Maybe add a toDs.VerifyCV(t, cvcas, sourceID, version) function?  (which could call a toDS.VerifyHLV(t, HLV) for future tests that care about pv?

xdcr/rosmar_xdcr.go
```diff
@@ -186,30 +193,55 @@ func (r *rosmarManager) Stop(_ context.Context) error {
 	return nil
 }
 
-// opWithMeta writes a document to the target datastore given a type of Deletion or Mutation event with a specific cas.
-func opWithMeta(ctx context.Context, collection *rosmar.Collection, originalCas uint64, event sgbucket.FeedEvent) error {
-	var xattrs []byte
+// opWithMeta writes a document to the target datastore given a type of Deletion or Mutation event with a specific cas. The originalXattrs will contain only the _vv and _mou xattr.
```
adamcfraser, line 196: XDCR will be preserving the _sync xattr on the target, so I had anticipated originalXattrs would include that for setWithMeta to work as expected?  Or is that pending work for a later ticket.

adamcfraser (changes requested):

xdcr/rosmar_xdcr.go
torcolvin: https://jira.issues.couchbase.com/browse/CBG-4281 for doing this, even though I think it's pretty trivial.

xdcr/xdcr_test.go
torcolvin: This test was wrong, thanks.

adamcfraser (approved):

PR #7120, opened 2024-09-20
CBG3026 prereq: More low-risk cleanup/refactoring

bbrks:
This PR contains the changes unrelated to `DatabaseInitManager` that are in the upcoming CBG-3026 PR just to completely isolate those changes given the complexity.

- Move `DatabaseLogCtx` earlier in `_getOrAddDatabaseFromConfig` so we can identify which database init-related logging is for
- `t.Run`->`rt.Run` fixes (prevents panics with failing subtests)
- Minor rename and TODO comment for upcoming work

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2694/

### DIFF
--- a/rest/server_context.go
+++ b/rest/server_context.go
@@ -588,6 +588,14 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
 		dbName = spec.BucketName
 	}
 
+	// Generate database context options from config and server context
+	contextOptions, err := dbcOptionsFromConfig(ctx, sc, &config.DbConfig, dbName)
+	if err != nil {
+		return nil, err
+	}
+	// set this early so we have dbName available in db-init related logging, before we have an actual database
+	ctx = base.DatabaseLogCtx(ctx, dbName, contextOptions.LoggingConfig)
+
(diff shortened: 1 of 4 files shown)


rest/server_context.go
```diff
@@ -588,6 +588,14 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
 		dbName = spec.BucketName
 	}
 
+	// Generate database context options from config and server context
+	contextOptions, err := dbcOptionsFromConfig(ctx, sc, &config.DbConfig, dbName)
```
torcolvin: Can you check that this is okay that this isn't going to return an error for `db.ValidateDatabaseName`?

torcolvin (dismissed):

rest/server_context.go
bbrks: Ah! I missed that. I also realised this code is being put before the defer as well, so I'll move things down a bit. There's also the duplicate check we probably want to do first to exit early.

torcolvin (approved):

PR #7121, opened 2024-09-20
CBG-3026: Use `DatabaseInitManager` in all GSI cases

bbrks:
CBG-3026

Rework database initialization code to reuse `DatabaseInitManager` in all GSI cases (including persistent config) to centralize and de-duplicate index creation code.\

Pre-req for CBG-2838 to rearrange per-collection/system:indexes iteration logic to minimise collection overhead when building/waiting for indexes.

## Depends on
- [x] #7120 

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2751

### DIFF
--- a/base/bootstrap.go
+++ b/base/bootstrap.go
@@ -45,9 +45,6 @@ type BootstrapConnection interface {
 	// Returns exists=false if key is not found, returns error for any other error.
 	GetDocument(ctx context.Context, bucket, docID string, rv interface{}) (exists bool, err error)
 
-	// Returns the bootstrap connection's cluster connection as N1QLStore for the specified bucket/scope/collection.
-	// Does NOT establish a bucket connection, the bucketName/scopeName/collectionName is for query scoping only
-	GetClusterN1QLStore(bucketName, scopeName, collectionName string) (*ClusterOnlyN1QLStore, error)
 	// Close releases any long-lived connections
 	Close()
 }
@@ -149,54 +146,23 @@ var _ BootstrapConnection = &CouchbaseCluster{}
 func NewCouchbaseCluster(ctx context.Context, server, username, password,
 	x509CertPath, x509KeyPath, caCertPath string,
 	forcePerBucketAuth bool, perBucketCreds PerBucketCredentialsConfig,
-	tlsSkipVerify *bool, useXattrConfig *bool, bucketMode BucketConnectionMode) (*CouchbaseCluster, error) {
-	opts := newCouchbaseClusterOpts{
-		server:             server,
-		username:           username,
-		password:           password,
-		x509CertPath:       x509CertPath,
-		x509KeyPath:        x509KeyPath,
-		caCertPath:         caCertPath,
-		forcePerBucketAuth: forcePerBucketAuth,
-		perBucketCreds:     perBucketCreds,
-		tlsSkipVerify:      tlsSkipVerify,
-		useXattrConfig:     useXattrConfig,
-		bucketMode:         bucketMode,
-	}
-	return NewCouchbaseClusterWithOpts(ctx, opts)
-}
-
-type newCouchbaseClusterOpts struct {
-	server             string
-	username           string
-	password           string
-	x509CertPath       string
-	x509KeyPath        string
-	caCertPath         string
-	forcePerBucketAuth bool
-	perBucketCreds     PerBucketCredentialsConfig
-	tlsSkipVerify      *bool
-	useXattrConfig     *bool
-	bucketMode         BucketConnectionMode
-}
-
-func NewCouchbaseClusterWithOpts(ctx context.Context, opts newCouchbaseClusterOpts) (*CouchbaseCluster, error) {
-	securityConfig, err := GoCBv2SecurityConfig(ctx, opts.tlsSkipVerify, opts.caCertPath)
+	tlsSkipVerify *bool, useXattrConfig bool, bucketMode BucketConnectionMode) (*CouchbaseCluster, error) {
+	securityConfig, err := GoCBv2SecurityConfig(ctx, tlsSkipVerify, caCertPath)
 	if err != nil {
 		return nil, err
 	}
 
 	clusterAuthConfig, err := GoCBv2Authenticator(
-		opts.username, opts.password,
-		opts.x509CertPath, opts.x509KeyPath,
+		username, password,
+		x509CertPath, x509KeyPath,
 	)
 	if err != nil {
 		return nil, err
 	}
 
 	// Populate individual bucket credentials
-	perBucketAuth := make(map[string]*gocb.Authenticator, len(opts.perBucketCreds))
-	for bucket, credentials := range opts.perBucketCreds {
+	perBucketAuth := make(map[string]*gocb.Authenticator, len(perBucketCreds))
+	for bucket, credentials := range perBucketCreds {
 		authenticator, err := GoCBv2Authenticator(
 			credentials.Username, credentials.Password,
 			credentials.X509CertPath, credentials.X509KeyPath,
@@ -215,19 +181,19 @@ func NewCouchbaseClusterWithOpts(ctx context.Context, opts newCouchbaseClusterOp
 	}
 
 	cbCluster := &CouchbaseCluster{
-		server:               opts.server,
-		forcePerBucketAuth:   opts.forcePerBucketAuth,
+		server:               server,
+		forcePerBucketAuth:   forcePerBucketAuth,
 		perBucketAuth:        perBucketAuth,
 		clusterOptions:       clusterOptions,
-		bucketConnectionMode: opts.bucketMode,
+		bucketConnectionMode: bucketMode,
 	}
 
-	if opts.bucketMode == CachedClusterConnections {
+	if bucketMode == CachedClusterConnections {
 		cbCluster.cachedBucketConnections = cachedBucketConnections{buckets: make(map[string]*cachedBucket)}
 	}
 
 	cbCluster.configPersistence = &DocumentBootstrapPersistence{}
-	if BoolDefault(opts.useXattrConfig, false) {
+	if useXattrConfig {
 		cbCluster.configPersistence = &XattrBootstrapPersistence{}
 	}
 
@@ -514,15 +480,6 @@ func (cc *CouchbaseCluster) Close() {
 	}
 }
 
-func (cc *CouchbaseCluster) GetClusterN1QLStore(bucketName, scopeName, collectionName string) (*ClusterOnlyN1QLStore, error) {
-	gocbCluster, err := cc.getClusterConnection()
-	if err != nil {
-		return nil, err
-	}
-
-	return NewClusterOnlyN1QLStore(gocbCluster, bucketName, scopeName, collectionName)
-}
-
 func (cc *CouchbaseCluster) getBucket(ctx context.Context, bucketName string) (b *gocb.Bucket, teardownFn func(), err error) {
 
 	if cc.bucketConnectionMode != CachedClusterConnections {
@@ -553,9 +510,7 @@ func (cc *CouchbaseCluster) getBucket(ctx context.Context, bucketName string) (b
 	return newBucket, teardownFn, nil
 }
 
-// connectToBucket establishes a new connection to a bucket, and returns the bucket after waiting for it to be ready.
-func (cc *CouchbaseCluster) connectToBucket(ctx context.Context, bucketName string) (b *gocb.Bucket, teardownFn func(), err error) {
-	var connection *gocb.Cluster
+func (cc *CouchbaseCluster) GetClusterConnectionForBucket(ctx context.Context, bucketName string) (connection *gocb.Cluster, teardownFn func(), err error) {
 	if bucketAuth, set := cc.perBucketAuth[bucketName]; set {
 		connection, err = cc.connect(bucketAuth)
 	} else if cc.forcePerBucketAuth {
@@ -563,7 +518,22 @@ func (cc *CouchbaseCluster) connectToBucket(ctx context.Context, bucketName stri
 	} else {
 		connection, err = cc.connect(nil)
 	}
+	if err != nil {
+		return nil, nil, err
+	}
 
+	teardownFn = func() {
+		err := connection.Close(&gocb.ClusterCloseOptions{})
+		if err != nil {
+			WarnfCtx(ctx, "Failed to close cluster connection: %v", err)
+		}
+	}
+	return connection, teardownFn, nil
+}
+
+// connectToBucket establishes a new connection to a bucket, and returns the bucket after waiting for it to be ready.
+func (cc *CouchbaseCluster) connectToBucket(ctx context.Context, bucketName string) (b *gocb.Bucket, teardownFn func(), err error) {
+	connection, teardownFn, err := cc.GetClusterConnectionForBucket(ctx, bucketName)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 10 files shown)


rest/database_init_manager.go
```diff
@@ -100,6 +100,7 @@ func (m *DatabaseInitManager) InitializeDatabase(ctx context.Context, startupCon
 
 	// Start a goroutine to perform the initialization
 	go func() {
+		defer func() { _ = n1qlStore.Close() }()
```
torcolvin: Is it correct to put a Close method on this given the doctstring comment for `ClusterOnlyN1qlStore`:

```
// ClusterOnlyN1qlStore implements the N1QLStore using only a cluster connection.
// Currently still intended for use for operations against a single collection, but maintains that
// information via metadata, and so supports sharing of the underlying gocb.Cluster with other
// ClusterOnlyN1QLStore instances. 
```

I think that aspect that is weird here is that:

```
couchbaseCluster, err := createBootstrapConnectionWithCredentials(ctx, startupConfig, dbConfig.Username, dbConfig.Password, base.PerUseClusterConnections)
```

has `PerUseClusterConnections`. More commentary on this below about how I think this might not be the right thing to do.

What I think we should do is that if we keep `ClusterOnlyN1qlStore` it should only close if `NewClusterOnlyN1QLStore` based on what `bucketConnectionMode` is set to.

I think test coverage of opening a connection here with both settings of `BucketConnectionMode` and making sure that the subsequent `ClusterOnlyN1qlStore` and `CouchbaseCluster` closes will shut the connection if appropriate.

rest/server_context.go
```diff
@@ -708,75 +682,52 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
...
-		gsiSupported := bucket.IsSupported(sgbucket.BucketStoreFeatureN1ql)
-		if !gsiSupported {
+	if !useViews {
+		// Initialize any required indexes
+		if gsiSupported := bucket.IsSupported(sgbucket.BucketStoreFeatureN1ql); !gsiSupported {
 			return nil, errors.New("Sync Gateway was unable to connect to a query node on the provided Couchbase Server cluster.  Ensure a query node is accessible, or set 'use_views':true in Sync Gateway's database config.")
 		}
 
+		if sc.DatabaseInitManager == nil {
+			base.AssertfCtx(ctx, "DatabaseInitManager should always be initialized")
```
torcolvin, line 722: I'd be fine to assert a panic here, but for non devmode this needs to return since 725 looks like it will panic. I think that `HasActiveInitialization` now has the wrong comments because it says:

```
func (m *DatabaseInitManager) HasActiveInitialization(dbName string) bool {
	if m == nil {
		// When not using persistent config, DatabaseInitManager will be nil
		return false
	}
```

rest/database_init_manager.go
```diff
@@ -72,9 +73,9 @@ func (m *DatabaseInitManager) InitializeDatabase(ctx context.Context, startupCon
 		delete(m.workers, dbConfig.Name)
 	}
 
-	base.InfofCtx(ctx, base.KeyAll, "Starting new async initialization for database %s ...",
+	base.InfofCtx(ctx, base.KeyAll, "Starting new initialization for database %s ...",
 		base.MD(dbConfig.Name))
-	couchbaseCluster, err := CreateBootstrapConnectionFromStartupConfig(ctx, startupConfig, base.PerUseClusterConnections)
+	couchbaseCluster, err := createBootstrapConnectionWithCredentials(ctx, startupConfig, dbConfig.Username, dbConfig.Password, base.PerUseClusterConnections)
```
torcolvin: I wonder why here that we don't just use the bootstrap connection if we have one? This might address the statement about closing a `ClusterN1QLStore` and drop an extra connection to couchbase server we don't need to maintain in this place.

A test would need to be added to ensure that we _do_ create a cluster connection if there isn't a bootstrap connection.

This might be too much of a stretch goal for this commit, but I think it might be easier. I think it's not impossible to just change the code to also just have a bootstrap connection available when running non persistent config, but do not run config pooling. Some of the reasons we didn't run a bootstrap connection used to be that we didn't need to(!) and we couldn't for walrus. I've implemented a boostrap connection for rosmar, though obviously there is no N1QL support, so this would need to type check that this is a `CouchbaseCluster` implementation of `BootstrapConnection` if this approach is used.

Another way that the code in `CouchbaseCluster` is complicated is that at some point in elixir there were `perBucketAuth` and each of these needed a separate cluster connection, so I wonder if this is why this code pathway exists here in the first place. Maybe it is still possible to use separate auth here. But, there might be an issue for non persistent config since you could specify a `server`, `username`, `password` field on the database config and those would be used here, I think.

If you add/drop a database in memory on a non persistent config cluster, you'd want to make sure this works if this code gets modified.

rest/database_init_manager.go
```diff
@@ -156,27 +157,27 @@ func (m *DatabaseInitManager) Cancel(dbName string) {
...
-				scName := base.ScopeAndCollectionName{Scope: scopeName, Collection: collectionName}
-				collectionInitData[scName] = metadataIndexOption
+func buildCollectionIndexData(config *DatabaseConfig) CollectionInitData {
+	if len(config.Scopes) == 0 {
+		return CollectionInitData{base.DefaultScopeAndCollectionName(): db.IndexesAll}
+	}
+
+	defaultScopeAndCollectionMetadataIndexes := db.IndexesMetadataOnly
+
+	collectionInitData := make(CollectionInitData)
```
torcolvin, line 181: I think this code refactor correct, but I'm actually not 100% sure. There are three basic cases we want to test for:

1. Scopes = nil ( _default._default has `db.IndexesAll`)
2. Scopes include `_default._default`  - db.IndexesAll
3. Scopes does not include `_default._default`  ( _default._default has `db.IndexesMetadataOnly` )

Since this function is now standalone, I think a parameterized test would be appropriate here.

The value of writing the test is that I wrote code in a way that was easy to break https://github.com/couchbase/sync_gateway/commit/d51ca17fffe89cdcc1b0749ee78fa55d665b1671#diff-cb34dfa4c30aa50c64ad71ec345be97702357afa3677d066e3665e8edc997adb. I don't think that it totally makes sense have a shared function that does this and move it to db package, though we could since this function could just construct datastores here, and pass the actual creation of `CollectionInitData` to db and have `db.indextest.setupN1QLStore` share code in both places. I thought recently that this would share code with `viewsAndGSIBucketInit`, but in `viewsAndGSIBucketInit` `_default._default` always has `IndexesAll` since the db/rest tests support running a database with a default collection, and we are trying to optimize for fastest test bucket pool performance.

You can take or leave this test and definitely the refactor doesn't need to happen.

rest/server_context.go
```diff
@@ -652,52 +649,29 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
...
 				dataStore, err := base.GetAndWaitUntilDataStoreReady(ctx, bucket, scName, options.failFast)
 				if err != nil {
 					return nil, fmt.Errorf("error attempting to create/update database: %w", err)
 				}
 
-				metadataIndexOption := db.IndexesWithoutMetadata
-				if base.IsDefaultCollection(scopeName, collectionName) {
-					hasDefaultCollection = true
-					metadataIndexOption = db.IndexesAll
+				if useViews {
```
torcolvin, line 672: I think this is fine to leave here, but I might note on this line that indexes will be initialized by DatabaseInitManager.

rest/server_context.go
```diff
@@ -708,75 +682,52 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
...
-		if resyncRequired {
-			collectionsRequiringResync = append(collectionsRequiringResync, scName)
-		}
-		if err := initDataStore(ds, db.IndexesAll, scName); err != nil {
-			return nil, err
+		if useViews {
+			if err := db.InitializeViews(ctx, ds); err != nil {
+				return nil, err
+			}
 		}
```
torcolvin, line 707: If there's no a default collection we do not need to initialize views.

rest/server_context.go
```diff
@@ -652,52 +649,29 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
...
-
+	hasDefaultCollection := false
 	collectionsRequiringResync := make([]base.ScopeAndCollectionName, 0)
 	if len(config.Scopes) > 0 {
 		if !bucket.IsSupported(sgbucket.BucketStoreFeatureCollections) {
 			return nil, errCollectionsUnsupported
 		}
 
-		hasDefaultCollection := false
 		for scopeName, scopeConfig := range config.Scopes {
```
torcolvin, line 659: Take or leave this, but it has been confusing to me that we loop over the datastores here for collection require resync. It doesn't quite feel like this belongs in the same function as `buildCollectionIndexData` because this code is actually attempting to read teh datastore and writing to it with `InitSyncInfo`. It could include an extra parameter that represents `hasDefaultCollectionConfigured`, though that would be technically redundant information.

I might change the code to work something more like this as separate operations:

1. do a check for `if len(config.Scopes) > 0 && if !bucket.IsSupported(sgbucket.BucketStoreFeatureCollections)` to return an error.
2. call `buildCollectionIndexData` to get the dsNames and loop over them to call `GetAndWaitUntilDataStoreReady` and call set `InitSyncInfo`. `InitSyncInfo` is skipped for the case were `_default._default` is only stored for metadata (see above for `hasDefaultCollectionConfigured` variable suggestion).

I think we want to leave `InitSyncInfo` before the `DatabaseInitManager`, because capella lets you start a resync before the indexes have been created. The resync manager will block on to wait for the principal collections to be enabled.

I find this code _really_ confusing especially since it seems to be duplicating the work in the database init manager, which was sort of what you tried to resolve in the first place.

torcolvin (commented):
There are lots of comments because I was trying to resolve the complexity. I think that there's only one bug that I can spot which is closing the connection on `ClusterN1QLStore` if `CouchbaseCluster.GetClusterN1QLStore` is called when `CouchbaseCluster.bucketConnectionMode == CachedClusterConnections`.

I gave a lot of commentary on parts of this code I find confusing that might be worth fixing or not. I am OK if you decide not to implement any of the other ideas, but I think it's worth consdering adding code level comments for why they aren't implemented, because I will ask them next time I read this code. This could be a problem because I'm very familiar with this code that I'm asking these questions, and someone not familiar would not ask these questions.

rest/database_init_manager.go
torcolvin: The interesting thing about this circumstance is that the old code didn't support per bucket credentials here:

```
couchbaseCluster, err := CreateBootstrapConnectionFromStartupConfig(ctx, startupConfig, base.PerUseClusterConnections)
```

and it would always use the credentials from the last database defined https://github.com/couchbase/sync_gateway/blob/main/rest/config_legacy.go#L156 for all databases. In practice, this code path only is used with capella where bootstrap connection is defined, so this was never an issue before. I think to be technically correct this needs to account for possibly a different server as well.

rest/server_context.go
bbrks: Since our tests and rosmar allows non-default collections and views, we must support this case to populate the prinicipal view in the metadata collection. Removing this makes `TestSyncGatewayStartupIndexes` fail.

rest/database_init_manager.go
bbrks: Agreed, this code is trivial to unit test in this way now. That has been added to cover the permutations.

I can see the value in moving `DatabaseInitManager` down into `db` and re-using it in other areas, but will leave that for now given that `setupN1QLStore` duplication is currently isolated to just 3 tests.

rest/server_context.go
bbrks: I agree that the refactor isn't going far enough to make `DatabaseInitManger` cover _everything_ a database needs (views/`SyncInfo`), and potentially move into `db` package so we can reduce the responsibility of `_getOrAddDatabaseFromConfig` and the `rest` package to keep `db` a bit more standalone, but it will blow up scope that isn't nessesary for implementing the collection-index enhancements I'm working towards.

rest/database_init_manager.go
bbrks: This should be fixed now

base/bootstrap.go
```diff
@@ -150,23 +150,53 @@ func NewCouchbaseCluster(ctx context.Context, server, username, password,
...
+	server             string
+	username           string
+	password           string
+	x509CertPath       string
+	x509KeyPath        string
+	caCertPath         string
+	forcePerBucketAuth bool
+	perBucketCreds     PerBucketCredentialsConfig
+	tlsSkipVerify      *bool
+	useXattrConfig     *bool
```
torcolvin: nit: Given that this should either be true or false here and false is the default, do you think that this should just be concrete bool, and not pointer? 

I feel the same about tlsSkipVerify but that actually would have a downstream side effect of changing behavior of `GoCBv2SecurityConfig`

rest/main.go
```diff
@@ -369,28 +369,63 @@ func backupCurrentConfigFile(sourcePath string) (string, error) {
...
+type bootstrapConnectionOpts struct {
+	server               string
+	username             string
+	password             string
+	x509CertPath         string
+	x509KeyPath          string
+	caCertPath           string
+	bucketConnectionMode base.BucketConnectionMode
+	isServerless         bool
+	bucketCredentials    base.PerBucketCredentialsConfig
```
torcolvin, line 388: This option needs to be populated with `StartupConfig.BucketCredentials`

rest/main.go
```diff
@@ -369,28 +369,63 @@ func backupCurrentConfigFile(sourcePath string) (string, error) {
...
+	tlsSkipVerify        *bool
+	useXattrConfig       *bool
```
torcolvin: This isn't used here, but I think it needs to be `StartupConfig.Unsupported.UseXattrConfig`

rest/main.go
```diff
@@ -369,28 +369,63 @@ func backupCurrentConfigFile(sourcePath string) (string, error) {
...
+	server               string
+	username             string
+	password             string
+	x509CertPath         string
+	x509KeyPath          string
+	caCertPath           string
+	bucketConnectionMode base.BucketConnectionMode
+	isServerless         bool
+	bucketCredentials    base.PerBucketCredentialsConfig
+	tlsSkipVerify        *bool
```
torcolvin, line 389: This option isn't being used, but it needs to be `config.Bootstrap.ServerTLSSkipVerify`.

rest/main.go
```diff
@@ -369,28 +369,63 @@ func backupCurrentConfigFile(sourcePath string) (string, error) {
...
+		bucketConnectionMode: bucketConnectionMode,
+		isServerless:         config.IsServerless(),
+	}
+	return createBootstrapConnectionWithOpts(ctx, opts)
 }
 
-func createBootstrapConnectionWithCredentials(ctx context.Context, config *StartupConfig, username, password string, bucketConnectionMode base.BucketConnectionMode) (base.BootstrapConnection, error) {
-	if base.ServerIsWalrus(config.Bootstrap.Server) {
-		cluster := base.NewRosmarCluster(config.Bootstrap.Server)
+type bootstrapConnectionOpts struct {
```
torcolvin, line 379: Can we just have this use `CouchbaseClusterOptions`? Or document why these need to be different structs?

torcolvin (commented):
I like the approach but I believe there are now bugs in the bootstrap connection.

Also, the thing that needs to be fixed from the original PR is whether to make  `ClusterOnlyN1QLStore.Close` a no-op. That depends on whether `CouchbaseCluster.GetClusterN1QLStore` is called from a `CouchbaseCluster` that has `CouchbaseCluster.bucketConnectionMode == PerUseClusterConnections` (It should close, otherwise `ClusterOnlyN1QLStore.Close` needs to be a no-op

rest/main.go
bbrks: Intermediate commit - I'd already realised I can get rid of `CouchbaseClusterOptions` which should resolve this

base/bootstrap.go
bbrks: Done, but left tlsSkipVerify because I want to limit code change down through to `GoCBv2SecurityConfig`

torcolvin (approved):

PR #7125, opened 2024-09-24
CBG-4210: Attachment metadata migration background job

gregns1:
CBG-4210

- New background task for attachment migration
- New function to write metaVersion to sync info upon completion of the task for each collection that the task has run on. 

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2736/

### DIFF
--- a/base/constants_syncdocs.go
+++ b/base/constants_syncdocs.go
@@ -20,6 +20,7 @@ const SyncDocPrefix = "_sync:"                                 // Prefix for all
 const MetadataIdPrefix = "m_"                                  // Prefix for metadataId, to prevent collisions between namespaced and non-namespaced metadata documents
 const SyncDocMetadataPrefix = SyncDocPrefix + MetadataIdPrefix // Prefix for all namespaced Sync Gateway metadata documents
 const DCPCheckpointPrefix = "dcp_ck:"
+const MetadataVersionNumber = 4
 
 // Sync Gateway Metadata document types
 type metadataKey int
@@ -420,16 +421,20 @@ func SetSyncInfoMetadataID(ds DataStore, metadataID string) error {
 	if metadataID == "" {
 		return nil
 	}
-	var syncInfo SyncInfo
-	_, fetchErr := ds.Get(SGSyncInfo, &syncInfo)
-	// avoid returning error if doc not found, allow upsert operation below to create it if not existent
-	if fetchErr != nil && !IsDocNotFoundError(fetchErr) {
-		return fetchErr
-	}
-	// set new metadata id, preserving the meta version if present
-	syncInfo.MetadataID = metadataID
-
-	return ds.Set(SGSyncInfo, 0, nil, syncInfo)
+	_, err := ds.Update(SGSyncInfo, 0, func(current []byte) (updated []byte, expiry *uint32, delete bool, err error) {
+		var syncInfo SyncInfo
+		if current != nil {
+			parseErr := JSONUnmarshal(current, &syncInfo)
+			if parseErr != nil {
+				return nil, nil, false, parseErr
+			}
+		}
+		// if we have a metadataID to set, set it preserving the metadata ID if present
+		syncInfo.MetadataID = metadataID
+		bytes, err := JSONMarshal(&syncInfo)
+		return bytes, nil, false, err
+	})
+	return err
 }
 
 // SetSyncInfoMetaVersion sets sync info in DataStore to specified metadata version, preserving metadataID if present
@@ -437,16 +442,20 @@ func SetSyncInfoMetaVersion(ds DataStore, metaVersion int8) error {
 	if metaVersion == 0 {
 		return nil
 	}
-	var syncInfo SyncInfo
-	_, fetchErr := ds.Get(SGSyncInfo, &syncInfo)
-	// avoid returning error if doc not found, allow upsert operation below to create it if not existent
-	if fetchErr != nil && !IsDocNotFoundError(fetchErr) {
-		return fetchErr
-	}
-	// if we have a meta version to set, set it preserving the metadata ID if present
-	syncInfo.MetaDataVersion = metaVersion
-
-	return ds.Set(SGSyncInfo, 0, nil, syncInfo)
+	_, err := ds.Update(SGSyncInfo, 0, func(current []byte) (updated []byte, expiry *uint32, delete bool, err error) {
+		var syncInfo SyncInfo
+		if current != nil {
+			parseErr := JSONUnmarshal(current, &syncInfo)
+			if parseErr != nil {
+				return nil, nil, false, parseErr
+			}
+		}
+		// if we have a meta version to set, set it preserving the metadata ID if present
+		syncInfo.MetaDataVersion = metaVersion
+		bytes, err := JSONMarshal(&syncInfo)
+		return bytes, nil, false, err
+	})
+	return err
 }
 
 // SerializeIfLonger returns name as a sha1 string if the length of the name is greater or equal to the length specificed. Otherwise, returns the original string.
--- a/base/constants_syncdocs.go
+++ b/base/constants_syncdocs.go
@@ -377,7 +377,8 @@ func CollectionSyncFunctionKeyWithGroupID(groupID string, scopeName, collectionN
 
 // SyncInfo documents are stored in collections to identify the metadataID associated with sync metadata in that collection
 type SyncInfo struct {
-	MetadataID string `json:"metadataID"`
+	MetadataID      string `json:"metadataID,omitempty"`
+	MetaDataVersion int8   `json:"metadata_version,omitempty""`
 }
 
 // initSyncInfo attempts to initialize syncInfo for a datastore
@@ -412,16 +413,39 @@ func InitSyncInfo(ds DataStore, metadataID string) (requiresResync bool, err err
 	return syncInfo.MetadataID != metadataID, nil
 }
 
-// SetSyncInfo sets syncInfo in a DataStore to the specified metadataID
-func SetSyncInfo(ds DataStore, metadataID string) error {
+// SetSyncInfoMetadataID sets syncInfo in a DataStore to the specified metadataID, preserving metadata version if present
+func SetSyncInfoMetadataID(ds DataStore, metadataID string) error {
 
 	// If the metadataID isn't defined, don't persist SyncInfo.  Defensive handling for legacy use cases.
 	if metadataID == "" {
 		return nil
 	}
-	syncInfo := &SyncInfo{
-		MetadataID: metadataID,
+	var syncInfo SyncInfo
+	_, fetchErr := ds.Get(SGSyncInfo, &syncInfo)
+	// avoid returning error if doc not found, allow upsert operation below to create it if not existent
+	if fetchErr != nil && !IsDocNotFoundError(fetchErr) {
+		return fetchErr
 	}
+	// set new metadata id, preserving the meta version if present
+	syncInfo.MetadataID = metadataID
+
+	return ds.Set(SGSyncInfo, 0, nil, syncInfo)
+}
+
+// SetSyncInfoMetaVersion sets sync info in DataStore to specified metadata version, preserving metadataID if present
+func SetSyncInfoMetaVersion(ds DataStore, metaVersion int8) error {
+	if metaVersion == 0 {
+		return nil
+	}
+	var syncInfo SyncInfo
+	_, fetchErr := ds.Get(SGSyncInfo, &syncInfo)
+	// avoid returning error if doc not found, allow upsert operation below to create it if not existent
+	if fetchErr != nil && !IsDocNotFoundError(fetchErr) {
+		return fetchErr
+	}
+	// if we have a meta version to set, set it preserving the metadata ID if present
+	syncInfo.MetaDataVersion = metaVersion
+
 	return ds.Set(SGSyncInfo, 0, nil, syncInfo)
 }
 
--- a/base/constants_syncdocs.go
+++ b/base/constants_syncdocs.go
@@ -429,7 +429,7 @@ func SetSyncInfoMetadataID(ds DataStore, metadataID string) error {
 				return nil, nil, false, parseErr
 			}
 		}
-		// if we have a metadataID to set, set it preserving the metadata ID if present
+		// if we have a metadataID to set, set it preserving the metadata version if present
 		syncInfo.MetadataID = metadataID
 		bytes, err := JSONMarshal(&syncInfo)
 		return bytes, nil, false, err
--- a/base/constants_syncdocs.go
+++ b/base/constants_syncdocs.go
@@ -20,7 +20,6 @@ const SyncDocPrefix = "_sync:"                                 // Prefix for all
 const MetadataIdPrefix = "m_"                                  // Prefix for metadataId, to prevent collisions between namespaced and non-namespaced metadata documents
 const SyncDocMetadataPrefix = SyncDocPrefix + MetadataIdPrefix // Prefix for all namespaced Sync Gateway metadata documents
 const DCPCheckpointPrefix = "dcp_ck:"
-const MetadataVersionNumber = 4
 
 // Sync Gateway Metadata document types
 type metadataKey int
@@ -379,7 +378,7 @@ func CollectionSyncFunctionKeyWithGroupID(groupID string, scopeName, collectionN
 // SyncInfo documents are stored in collections to identify the metadataID associated with sync metadata in that collection
 type SyncInfo struct {
 	MetadataID      string `json:"metadataID,omitempty"`
-	MetaDataVersion int8   `json:"metadata_version,omitempty"`
+	MetaDataVersion string `json:"metadata_version,omitempty"`
 }
 
 // initSyncInfo attempts to initialize syncInfo for a datastore
@@ -438,8 +437,8 @@ func SetSyncInfoMetadataID(ds DataStore, metadataID string) error {
 }
 
 // SetSyncInfoMetaVersion sets sync info in DataStore to specified metadata version, preserving metadataID if present
-func SetSyncInfoMetaVersion(ds DataStore, metaVersion int8) error {
-	if metaVersion == 0 {
+func SetSyncInfoMetaVersion(ds DataStore, metaVersion string) error {
+	if metaVersion == "" {
 		return nil
 	}
 	_, err := ds.Update(SGSyncInfo, 0, func(current []byte) (updated []byte, expiry *uint32, delete bool, err error) {
--- a/base/constants_syncdocs.go
+++ b/base/constants_syncdocs.go
@@ -378,7 +378,7 @@ func CollectionSyncFunctionKeyWithGroupID(groupID string, scopeName, collectionN
 // SyncInfo documents are stored in collections to identify the metadataID associated with sync metadata in that collection
 type SyncInfo struct {
 	MetadataID      string `json:"metadataID,omitempty"`
-	MetaDataVersion int8   `json:"metadata_version,omitempty""`
+	MetaDataVersion int8   `json:"metadata_version,omitempty"`
 }
 
 // initSyncInfo attempts to initialize syncInfo for a datastore
--- a/base/version.go
+++ b/base/version.go
@@ -24,6 +24,8 @@ const (
 	ProductAPIVersion      = ProductAPIVersionMajor + "." + ProductAPIVersionMinor
 )
 
+const ProductMajorVersionInteger = 4
+
 // populated via init() below
 var (
 	// ProductVersion describes the specific version information of the build.
--- a/base/version.go
+++ b/base/version.go
@@ -24,8 +24,6 @@ const (
 	ProductAPIVersion      = ProductAPIVersionMajor + "." + ProductAPIVersionMinor
 )
 
-const ProductMajorVersionInteger = 4
-
 // populated via init() below
 var (
 	// ProductVersion describes the specific version information of the build.
--- a/db/background_mgr_attachment_migration.go
+++ b/db/background_mgr_attachment_migration.go
@@ -25,15 +25,20 @@ type AttachmentMigrationManager struct {
 	DocsChanged   base.AtomicInt
 	MigrationID   string
 	CollectionIDs []uint32
+	databaseCtx   *DatabaseContext
 	lock          sync.RWMutex
 }
 
 var _ BackgroundManagerProcessI = &AttachmentMigrationManager{}
 
-func NewAttachmentMigrationManager(metadataStore base.DataStore, metaKeys *base.MetadataKeys) *BackgroundManager {
+func NewAttachmentMigrationManager(database *DatabaseContext) *BackgroundManager {
+	metadataStore := database.MetadataStore
+	metaKeys := database.MetadataKeys
 	return &BackgroundManager{
-		name:    "attachment_migration",
-		Process: &AttachmentMigrationManager{},
+		name: "attachment_migration",
+		Process: &AttachmentMigrationManager{
+			databaseCtx: database,
+		},
 		clusterAwareOptions: &ClusterAwareBackgroundManagerOptions{
 			metadataStore: metadataStore,
 			metaKeys:      metaKeys,
@@ -51,7 +56,7 @@ func (a *AttachmentMigrationManager) Init(ctx context.Context, options map[strin
 		}
 
 		a.MigrationID = uniqueUUID.String()
-		base.InfofCtx(ctx, base.KeyAll, "Attachment Migration: Starting new migration run with migration ID: %q", a.MigrationID)
+		base.InfofCtx(ctx, base.KeyAll, "Attachment Migration: Starting new migration run with migration ID: %s", a.MigrationID)
 		return nil
 	}
 
@@ -68,7 +73,7 @@ func (a *AttachmentMigrationManager) Init(ctx context.Context, options map[strin
 		a.SetStatus(statusDoc.DocsChanged, statusDoc.DocsProcessed)
 		a.SetCollectionIDs(statusDoc.CollectionIDs)
 
-		base.InfofCtx(ctx, base.KeyAll, "Attachment Migration: Attempting to resume migration with migration ID: %s", a.MigrationID)
+		base.InfofCtx(ctx, base.KeyAll, "Attachment Migration: Resuming migration with migration ID: %s, %d already processed", a.MigrationID, a.DocsProcessed.Value())
 
 		return nil
 	}
@@ -77,13 +82,13 @@ func (a *AttachmentMigrationManager) Init(ctx context.Context, options map[strin
 }
 
 func (a *AttachmentMigrationManager) Run(ctx context.Context, options map[string]interface{}, persistClusterStatusCallback updateStatusCallbackFunc, terminator *base.SafeTerminator) error {
-	db := options["database"].(*Database)
+	db := a.databaseCtx
 	migrationLoggingID := "Migration: " + a.MigrationID
 
 	persistClusterStatus := func() {
 		err := persistClusterStatusCallback(ctx)
 		if err != nil {
-			base.WarnfCtx(ctx, "[%s] Failed to persist cluster status on-demand for attachment migration operation: %v", migrationLoggingID, err)
+			base.WarnfCtx(ctx, "[%s] Failed to persist latest cluster status for attachment migration: %v", migrationLoggingID, err)
 		}
 	}
 	defer persistClusterStatus()
@@ -120,7 +125,7 @@ func (a *AttachmentMigrationManager) Run(ctx context.Context, options map[string
 		a.DocsProcessed.Add(1)
 		syncData, _, _, err := UnmarshalDocumentSyncDataFromFeed(event.Value, event.DataType, collection.userXattrKey(), false)
 		if err != nil {
-			failProcess(err, "error unmarshaling document %s: %v", base.UD(docID), err)
+			failProcess(err, "error unmarshaling document %s with migrationID %s: %v", base.UD(docID), a.MigrationID, err)
 		}
 
 		if syncData == nil || syncData.Attachments == nil {
@@ -135,10 +140,9 @@ func (a *AttachmentMigrationManager) Run(ctx context.Context, options map[string
 		// xattr migration to take place
 		err = collWithUser.MigrateAttachmentMetadata(collCtx, docID, event.Cas, syncData)
 		if err != nil {
-			failProcess(err, "error migrating document attachment metadata for doc: %s: %v", base.UD(docID), err)
-		} else {
-			a.DocsChanged.Add(1)
+			failProcess(err, "error migrating document attachment metadata for doc: %s with migrationID %s: %v", base.UD(docID), a.MigrationID, err)
 		}
+		a.DocsChanged.Add(1)
 		return true
 	}
 
@@ -151,7 +155,7 @@ func (a *AttachmentMigrationManager) Run(ctx context.Context, options map[string
 	checkpointPrefix := db.MetadataKeys.DCPCheckpointPrefix(db.Options.GroupID) + "att_migration:"
 
 	// check for mismatch in collection id's between current collections on the db and prev run
-	err = a.resetDCPMetadataIfNeeded(ctx, db.MetadataStore, db, checkpointPrefix, currCollectionIDs)
+	err = a.resetDCPMetadataIfNeeded(ctx, db, checkpointPrefix, currCollectionIDs)
 	if err != nil {
 		return err
 	}
@@ -164,7 +168,7 @@ func (a *AttachmentMigrationManager) Run(ctx context.Context, options map[string
 		base.WarnfCtx(ctx, "[%s] Failed to create attachment migration DCP client: %v", migrationLoggingID, err)
 		return err
 	}
-	base.InfofCtx(ctx, base.KeyAll, "[%s] Starting DCP feed %q for attachment migration", migrationLoggingID, dcpFeedKey)
+	base.DebugfCtx(ctx, base.KeyAll, "[%s] Starting DCP feed %q for attachment migration", migrationLoggingID, dcpFeedKey)
 
 	doneChan, err := dcpClient.Start()
 	if err != nil {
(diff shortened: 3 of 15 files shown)


base/constants_syncdocs.go
```diff
@@ -412,16 +413,39 @@ func InitSyncInfo(ds DataStore, metadataID string) (requiresResync bool, err err
...
+	var syncInfo SyncInfo
+	_, fetchErr := ds.Get(SGSyncInfo, &syncInfo)
+	// avoid returning error if doc not found, allow upsert operation below to create it if not existent
+	if fetchErr != nil && !IsDocNotFoundError(fetchErr) {
+		return fetchErr
+	}
+	// set new metadata id, preserving the meta version if present
+	syncInfo.MetadataID = metadataID
+
+	return ds.Set(SGSyncInfo, 0, nil, syncInfo)
```
adamcfraser: When are we updating MetadataID and MetaVersion?  Using Set doesn't provide any CAS safety, so a change could happen to the underlying values between lines 424 and 432.  If that's something that could happen, this should be changed to Update(), which will automatically perform CAS retry.

base/version.go
```diff
@@ -24,6 +24,8 @@ const (
 	ProductAPIVersion      = ProductAPIVersionMajor + "." + ProductAPIVersionMinor
 )
 
+const ProductMajorVersionInteger = 4
```
adamcfraser: I think it would be better to define this specifically as a metadata version number, and store in constants_syncdocs.go.

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,288 @@
...
+
+		a.DocsProcessed.Add(1)
+		syncData, rawBody, rawXattrs, err := UnmarshalDocumentSyncDataFromFeed(event.Value, event.DataType, collection.userXattrKey(), false)
+		if err != nil {
+			failProcess(err, "error unmarshaling document %s: %v", base.UD(docID), err)
+		}
+
+		if rawXattrs[base.GlobalXattrName] != nil {
+			// global xattr already defined for this document, don't process
+			return true
```
adamcfraser: Are we able to handle/repair the case where the global xattr exists but _sync also includes attachments (some sort of data migration error)?

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,288 @@
...
+		}
+
+		// We only want to process full docs. Not any sync docs.
+		if strings.HasPrefix(docID, base.SyncDocPrefix) {
+			return true
+		}
+
+		a.DocsProcessed.Add(1)
+		syncData, rawBody, rawXattrs, err := UnmarshalDocumentSyncDataFromFeed(event.Value, event.DataType, collection.userXattrKey(), false)
```
adamcfraser: I had expected UnmarshalDocumentSyncDataFromFeed to populate syncData.Attachments from _globalSync (when already present), but it doesn't look like that's the case.  I suppose that's because the caching feed doesn't need attachments so we can skip the work, but it feels like there might be unexpected usage issues in future if someone assumes it's going to be there.  If you agree can you file a ticket for followup?

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,288 @@
...
+		}
+		if event.DataType&base.MemcachedDataTypeXattr == 0 {
+			// no xattrs on dcp event we have inline sync xattr to migrate
+			opts := DefaultMutateInOpts()
+			existingDoc := &sgbucket.BucketDocument{
+				Cas:    event.Cas,
+				Body:   rawBody,
+				Expiry: event.Expiry,
+			}
+			_, _, err = collWithUser.migrateMetadata(ctx, docID, existingDoc, opts)
```
adamcfraser: Should this use collCtx instead of ctx?

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,288 @@
...
+			// no attachments to migrate
+			return true
+		}
+
+		collCtx := collection.AddCollectionContext(ctx)
+		collWithUser := &DatabaseCollectionWithUser{
+			DatabaseCollection: collection,
+		}
+		if event.DataType&base.MemcachedDataTypeXattr == 0 {
+			// no xattrs on dcp event we have inline sync xattr to migrate
```
adamcfraser: Do we actually need to handle this case (inline _sync migration) as part of this job?  We're not going to be able to do anything with the document until it's migrated from using inline sync, so I think it's fine to wait until existing processing triggers migrateMetadata.  (since the metadata is inline in the document, it's also going to be replicated by XDCR in the meantime)

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,288 @@
...
+		FailOnRollback:    false,
+		MetadataStoreType: base.DCPMetadataStoreCS,
+		GroupID:           groupID,
+		CollectionIDs:     collectionIDs,
+		CheckpointPrefix:  checkpointPrefix,
+	}
+	return clientOptions
+}
+
+func getCollectionIDs(db *Database) []uint32 {
```
adamcfraser: This looks like something that's worth defining as `func (db *Database) GetCollectionIDs() []uint32` in database.go for future reuse.

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,288 @@
...
+		return true
+	}
+
+	bucket, err := base.AsGocbV2Bucket(db.Bucket)
+	if err != nil {
+		return err
+	}
+
+	collectionIDs := getCollectionIDs(db)
+	dcpOptions := getMigrationDCPClientOptions(collectionIDs, db.Options.GroupID, db.MetadataKeys.DCPCheckpointPrefix(db.Options.GroupID))
```
adamcfraser: What happens in the case where we're resuming a previously running attachment migration job, but the set of collections on the database has changed since we last run?  We'd want to restart the DCP feed from zero, but I don't see that handled here.

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+	base.InfofCtx(ctx, base.KeyAll, "[%s] Starting DCP feed %q for attachment migration", migrationLoggingID, dcpFeedKey)
+
+	doneChan, err := dcpClient.Start()
+	if err != nil {
+		base.WarnfCtx(ctx, "[%s] Failed to start attachment migration DCP feed: %v", migrationLoggingID, err)
+		_ = dcpClient.Close()
+		return err
+	}
+	base.DebugfCtx(ctx, base.KeyAll, "[%s] DCP client started.", migrationLoggingID)
```
adamcfraser: Including something about attachment migration in this log message will avoid future confusion.

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,288 @@
...
+			return err
+		}
+		if processFailure != nil {
+			return processFailure
+		}
+		// set sync info here
+		for _, collectionID := range collectionIDs {
+			dbc, ok := db.CollectionByID[collectionID]
+			if !ok {
+				base.WarnfCtx(ctx, "[%s] Completed attachment migration, but unable to update syncInfo for collection %v (not found)", migrationLoggingID, collectionID)
```
adamcfraser: I think clarifying that 'not found' means 'collection is no longer associated with the database' would be helpful here.  I suppose you could also downgrade this to Info since there shouldn't be a functional impact (don't require migration if the collection isn't part of the database).

adamcfraser (dismissed):
Overall looks good - some questions on the specifics.

db/background_mgr_attachment_migration.go
gregns1: Good spot, this is something I missed in earlier work on this body of work, followup ticket -> CBG-4280

db/background_mgr_attachment_migration.go
gregns1: I was on the fence myself about it but included it to be safe. I agree now I think more about it we can leave this out, considering if xdcr is turned on before the existing processing migrates the metadata we won't loose the metadata in the process as xdcr will copy it over.

db/background_mgr_attachment_migration.go
gregns1: Good idea, refactored the `MigrateAttachmentMetadata` function to handle case where global xattr exists. We will now preserve it if it exists and add whatever is defined in sync data to it.

base/constants_syncdocs.go
```diff
@@ -20,6 +20,7 @@ const SyncDocPrefix = "_sync:"                                 // Prefix for all
 const MetadataIdPrefix = "m_"                                  // Prefix for metadataId, to prevent collisions between namespaced and non-namespaced metadata documents
 const SyncDocMetadataPrefix = SyncDocPrefix + MetadataIdPrefix // Prefix for all namespaced Sync Gateway metadata documents
 const DCPCheckpointPrefix = "dcp_ck:"
+const MetadataVersionNumber = 4
```
torcolvin: Can you comment the intention behind this number? Is it intended to match the sync gateway version number to facilitate major version checks, or is this just planning for the future?

Should we use https://github.com/couchbase/sync_gateway/blob/9ce7f7bdea12048db13d6a5b694c357cbc4a9e1f/rest/config_registry.go#L49 that is in the config registry? If we want per collection, then I think that we could either include the earliest SG version (since we are attempting to use version 4), and not update it, or version this in a way of just starting at 1 to indicate it doesn't have special meaning.

Apologies if this was discussed before.

I think documenting what it means and how it is used in `SyncInfo.MetaDataVersion` would be helpful too.

base/constants_syncdocs.go
```diff
@@ -412,17 +414,48 @@ func InitSyncInfo(ds DataStore, metadataID string) (requiresResync bool, err err
 	return syncInfo.MetadataID != metadataID, nil
 }
 
-// SetSyncInfo sets syncInfo in a DataStore to the specified metadataID
-func SetSyncInfo(ds DataStore, metadataID string) error {
+// SetSyncInfoMetadataID sets syncInfo in a DataStore to the specified metadataID, preserving metadata version if present
+func SetSyncInfoMetadataID(ds DataStore, metadataID string) error {
```
torcolvin, line 417: Can you add tests for `SetSyncInfoMetdataID` and `SetSyncInfoMetaVersion`?

I think the testcases would be something like:

- doc doesn't exist
- doc exists with some properties
- doc exists with all property
- doc exists with non parseable sync info

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+		base.InfofCtx(ctx, base.KeyAll, "Attachment Migration: Attempting to resume migration with migration ID: %s", a.MigrationID)
+
+		return nil
+	}
+
+	return newRunInit()
+}
+
+func (a *AttachmentMigrationManager) Run(ctx context.Context, options map[string]interface{}, persistClusterStatusCallback updateStatusCallbackFunc, terminator *base.SafeTerminator) error {
+	db := options["database"].(*Database)
```
torcolvin: I think this will panic if `database` is not a `Database`. But I think that from my reading we don't need options at all, because we could just instantiate `AttachmentMigrationManager` with a database and not use these options at all.

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+			return err
+		}
+		if processFailure != nil {
+			return processFailure
+		}
+		// set sync info here
+		for _, collectionID := range currCollectionIDs {
+			dbc, ok := db.CollectionByID[collectionID]
+			if !ok {
+				base.InfofCtx(ctx, base.KeyAll, "[%s] Completed attachment migration, but unable to update syncInfo for collection %v (collection is no longer associated with the database)", migrationLoggingID, collectionID)
```
torcolvin: I don't quite see how this can happen, the dcp client is tied to a `DatabaseContext` and will not run without it. You can't change `CollectionByID` while a `DatabaseContext` exists.

It's possible that someone will change the collections while this background job is running, but then that will issue `DatabaseContext.Close` which will close the terminators, and hit `terminator.Done()`. In the very unlikely scenario that the dcp client finishes at the same time, `db.CollectionByID` will still be present.

We definitely don't want to panic in any circumstances, so if this is possible to hit, we should handle it (and write a test). But I think that as is, that error would be not a possibility and you can just do `collection := db.CollectionByID[collectionID]`

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+				continue
+			}
+			if err := base.SetSyncInfoMetaVersion(dbc.dataStore, base.MetadataVersionNumber); err != nil {
+				base.WarnfCtx(ctx, "[%s] Completed attachment migration, but unable to update syncInfo for collection %v: %v", migrationLoggingID, collectionID, err)
+			}
+		}
+	case <-terminator.Done():
+		base.DebugfCtx(ctx, base.KeyAll, "[%s] Terminator closed. Ending Attachment Migration process.", migrationLoggingID)
```
torcolvin: Given that we log on 213, not sure that this log message is very helpful

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+	if purgeNeeded {
+		err := a.PurgeDCPMetadata(ctx, datastore, database, metadataKeyPrefix)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// PurgeDCPMetadata will purge all DCP metadata from previous run in the bucket, used to reset dcp client to 0
```
torcolvin: Can you make this a common function with `AttachmentCompactionManager.PurgeDCPMetdata` ? I think it would just take a a string parameter for logging.

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+	if err != nil {
+		return err
+	}
+	numVbuckets, err := bucket.GetMaxVbno()
+	if err != nil {
+		return err
+	}
+
+	metadata := base.NewDCPMetadataCS(ctx, datastore, numVbuckets, base.DefaultNumWorkers, metadataKeyPrefix)
+	base.InfofCtx(ctx, base.KeyDCP, "purging persisted dcp metadata for attachment migration run %s", a.MigrationID)
```
torcolvin: Since this logs at info, we should try to make this make sense to users:

```suggestion
	base.InfofCtx(ctx, base.KeyDCP, "Purging invalid checkpoints for attachment migration run %s", a.MigrationID)
```

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+	slices.Sort(collectionIDs)
+	slices.Sort(a.CollectionIDs)
+
+	var purgeNeeded bool
+	for i, v := range collectionIDs {
+		if v != a.CollectionIDs[i] {
+			purgeNeeded = true
+			break
+		}
+	}
```
torcolvin: possible tweak for readability:


```suggestion
	slices.Sort(collectionIDs)
	slices.Sort(a.CollectionIDs)

	purgeNeeded := !slices.Compare(collectionIDs, a.CollectionIDs) 
```

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+	if purgeNeeded {
```
torcolvin: nit: I'd prefer early return when possible when reading code

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+		// If the previous run completed, or there was an error during unmarshalling the status we will start the
+		// process from scratch with a new migration ID. Otherwise, we should resume with the migration ID, stats specified in the doc.
+		if statusDoc.State == BackgroundProcessStateCompleted || err != nil {
+			return newRunInit()
+		}
+		a.MigrationID = statusDoc.MigrationID
+		a.SetStatus(statusDoc.DocsChanged, statusDoc.DocsProcessed)
+		a.SetCollectionIDs(statusDoc.CollectionIDs)
+
+		base.InfofCtx(ctx, base.KeyAll, "Attachment Migration: Attempting to resume migration with migration ID: %s", a.MigrationID)
```
torcolvin:
```suggestion
		base.InfofCtx(ctx, base.KeyAll, "Attachment Migration: Resuming migration with migration ID: %s, %d already processed", a.MigrationID, a.DocsProcessed)
```

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+
+func (a *AttachmentMigrationManager) Init(ctx context.Context, options map[string]interface{}, clusterStatus []byte) error {
+	newRunInit := func() error {
+		uniqueUUID, err := uuid.NewRandom()
+		if err != nil {
+			return err
+		}
+
+		a.MigrationID = uniqueUUID.String()
+		base.InfofCtx(ctx, base.KeyAll, "Attachment Migration: Starting new migration run with migration ID: %q", a.MigrationID)
```
torcolvin:
```suggestion
		base.InfofCtx(ctx, base.KeyAll, "Attachment Migration: Starting new migration run with migration ID: %s", a.MigrationID)
```

We log this as `%s` below and I think consistency when reading logs can be helpful.

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+
+		// We only want to process full docs. Not any sync docs.
+		if strings.HasPrefix(docID, base.SyncDocPrefix) {
+			return true
+		}
+
+		a.DocsProcessed.Add(1)
+		syncData, _, _, err := UnmarshalDocumentSyncDataFromFeed(event.Value, event.DataType, collection.userXattrKey(), false)
+		if err != nil {
+			failProcess(err, "error unmarshaling document %s: %v", base.UD(docID), err)
```
torcolvin: This needs to mark the attachment ID here, or we need to add this to `failProcess`

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+		}
+
+		collCtx := collection.AddCollectionContext(ctx)
+		collWithUser := &DatabaseCollectionWithUser{
+			DatabaseCollection: collection,
+		}
+		// xattr migration to take place
+		err = collWithUser.MigrateAttachmentMetadata(collCtx, docID, event.Cas, syncData)
+		if err != nil {
+			failProcess(err, "error migrating document attachment metadata for doc: %s: %v", base.UD(docID), err)
```
torcolvin: This needs to mark attachment migration ID or this `failProcess` should do it.

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+		} else {
+			a.DocsChanged.Add(1)
+		}
```
torcolvin: nit: for readability

```suggestion
		}
		a.DocsChanged.Add(1)
```

db/background_mgr_attachment_migration.go
torcolvin: If we decide to keep this message, this should say the collection name in addition to ID, ID isn't really very useful for debugging.

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+		// set sync info here
+		for _, collectionID := range currCollectionIDs {
+			dbc, ok := db.CollectionByID[collectionID]
+			if !ok {
+				base.InfofCtx(ctx, base.KeyAll, "[%s] Completed attachment migration, but unable to update syncInfo for collection %v (collection is no longer associated with the database)", migrationLoggingID, collectionID)
+				continue
+			}
+			if err := base.SetSyncInfoMetaVersion(dbc.dataStore, base.MetadataVersionNumber); err != nil {
+				base.WarnfCtx(ctx, "[%s] Completed attachment migration, but unable to update syncInfo for collection %v: %v", migrationLoggingID, collectionID, err)
```
torcolvin: This should say the collection name here, the collection ID isn't helpful for debugging.

If this fails, does this mean that we should mark the process failed, because it will need to run again?

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+			}
+		}
+	case <-terminator.Done():
+		base.DebugfCtx(ctx, base.KeyAll, "[%s] Terminator closed. Ending Attachment Migration process.", migrationLoggingID)
+		err = dcpClient.Close()
+		if err != nil {
+			base.WarnfCtx(ctx, "[%s] Failed to close attachment migration DCP client! %v", migrationLoggingID, err)
```
torcolvin:
```suggestion
			base.WarnfCtx(ctx, "[%s] Failed to close attachment migration DCP client after attachment migration process was terminated %v", migrationLoggingID, err)
```

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+}
+
+func (a *AttachmentMigrationManager) Run(ctx context.Context, options map[string]interface{}, persistClusterStatusCallback updateStatusCallbackFunc, terminator *base.SafeTerminator) error {
+	db := options["database"].(*Database)
+	migrationLoggingID := "Migration: " + a.MigrationID
+
+	persistClusterStatus := func() {
+		err := persistClusterStatusCallback(ctx)
+		if err != nil {
+			base.WarnfCtx(ctx, "[%s] Failed to persist cluster status on-demand for attachment migration operation: %v", migrationLoggingID, err)
```
torcolvin: What does on-demand mean here? I don't know if cluster status makes sense here for logging since the cluster status doc is bark of BackgroundManager interface, I think this is really "Can't persist latest status for attachment migration"

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+	base.InfofCtx(ctx, base.KeyAll, "[%s] Starting DCP feed %q for attachment migration", migrationLoggingID, dcpFeedKey)
+
+	doneChan, err := dcpClient.Start()
+	if err != nil {
+		base.WarnfCtx(ctx, "[%s] Failed to start attachment migration DCP feed: %v", migrationLoggingID, err)
+		_ = dcpClient.Close()
+		return err
+	}
+	base.DebugfCtx(ctx, base.KeyAll, "[%s] DCP client started for Attachment Migration.", migrationLoggingID)
```
torcolvin: I'd drop this to trace logging

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+
+	select {
+	case <-doneChan:
+		base.InfofCtx(ctx, base.KeyAll, "[%s] Finished migrating attachment metadata from sync data to global sync data. %d/%d docs changed", migrationLoggingID, a.DocsChanged.Value(), a.DocsProcessed.Value())
```
torcolvin: Should this logging get moved to the bottom of the `<-doneChan` block since we can actually return and persist an error here?

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+		err = dcpClient.Close()
+		if err != nil {
+			base.WarnfCtx(ctx, "[%s] Failed to close attachment migration DCP client! %v", migrationLoggingID, err)
+			return err
```
torcolvin: I think this is a warning but I don't think this should fail the attachment process, I think that we should probably continue through this block.

(I don't think this is at all likely to fail)

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,351 @@
...
+
+	a.SetCollectionIDs(currCollectionIDs)
+	dcpOptions := getMigrationDCPClientOptions(currCollectionIDs, db.Options.GroupID, checkpointPrefix)
+	dcpFeedKey := GenerateAttachmentMigrationDCPStreamName(a.MigrationID)
+	dcpClient, err := base.NewDCPClient(ctx, dcpFeedKey, callback, *dcpOptions, bucket)
+	if err != nil {
+		base.WarnfCtx(ctx, "[%s] Failed to create attachment migration DCP client: %v", migrationLoggingID, err)
+		return err
+	}
+	base.InfofCtx(ctx, base.KeyAll, "[%s] Starting DCP feed %q for attachment migration", migrationLoggingID, dcpFeedKey)
```
torcolvin: I think should Debug at best, since we already marked that we are starting the process in `Init`, unless that resume doesn't pass through the code.

db/background_mgr_attachment_migration_test.go
```diff
@@ -0,0 +1,312 @@
...
+		t.Skip("rosmar does not support DCP client, pending CBG-4249")
+	}
+	base.LongRunningTest(t)
+
+	db, ctx := setupTestDB(t)
+	defer db.Close(ctx)
+	collection, ctx := GetSingleDatabaseCollectionWithUser(ctx, t, db)
+
+	// create some docs with attachments defined
+	for i := 0; i < 4000; i++ {
```
torcolvin, line 104: comment why we need so many docs (to be able to be able to stop the code while this running. I wonder if this should be a shared constant with other tests that are trying to resume stopped code, since this number might end up being different for CBS and rosmar (when supported).

db/background_mgr_attachment_migration_test.go
```diff
@@ -0,0 +1,312 @@
...
+		docBody := Body{
+			"value":         1234,
+			BodyAttachments: map[string]interface{}{"myatt": map[string]interface{}{"content_type": "text/plain", "data": "SGVsbG8gV29ybGQh"}},
+		}
+		key := fmt.Sprintf("%s_%d", t.Name(), i)
+		_, doc, err := collection.Put(ctx, key, docBody)
+		require.NoError(t, err)
+		assert.NotNil(t, doc.SyncData.Attachments)
```
torcolvin:
```suggestion
		require.NotNil(t, doc.SyncData.Attachments)
```

If this error is hit, the rest of the code does not work.

db/background_mgr_attachment_migration_test.go
```diff
@@ -0,0 +1,312 @@
...
+	require.NoError(t, err)
+
+	// wait for task to complete
+	err = WaitForConditionWithOptions(t, func() bool {
+		var status BackgroundManagerStatus
+		rawStatus, _ := attachMigrationMgr.GetStatus(ctx)
+		_ = base.JSONUnmarshal(rawStatus, &status)
+		return status.State == BackgroundProcessStateCompleted
+	}, 200, 200)
+	require.NoError(t, err)
```
torcolvin: This test code is duplicated. If this fails, I think it's pretty confusing, because we don't know the last status. I would replace this with:

requireBackgroundManagerState(mgr, state) function that could be reused.

I would use `require.EventuallyWithT` for this, so we get the status at the end of the failure. This should also check the the errors for `GetStatus` and `JSONUnmarshal`

db/background_mgr_attachment_migration_test.go
```diff
@@ -0,0 +1,312 @@
...
+	// assert that the two added docs above were processed but not changed
+	stats := getAttachmentMigrationStats(attachMigrationMgr.Process)
+	assert.Equal(t, int64(1), stats.DocsProcessed)
+	assert.Equal(t, int64(1), stats.DocsChanged)
+
+	// assert that the sync info metadata version doc has been written to the database collection
+	var syncInfo base.SyncInfo
+	_, err = collection.dataStore.Get(base.SGSyncInfo, &syncInfo)
+	require.NoError(t, err)
+	assert.Equal(t, int8(4), syncInfo.MetaDataVersion)
```
torcolvin:
```suggestion
	assert.Equal(t, base.MetadataVersionNumber, syncInfo.MetaDataVersion)
```

db/background_mgr_attachment_migration_test.go
```diff
@@ -0,0 +1,312 @@
...
+
+	xattrs, _, err = collection.dataStore.GetXattrs(ctx, key, []string{base.SyncXattrName, base.GlobalXattrName})
+	require.NoError(t, err)
+	require.NotNil(t, xattrs[base.GlobalXattrName])
+	require.NotNil(t, xattrs[base.SyncXattrName])
```
torcolvin: Using contains will log what xattrs are included.

```suggestion
	require.Contains(t, xattrs, base.GlobalXattrName)
	require.Contains(t, xattrs, base.SyncXattrName)
```

db/database_test.go
```diff
@@ -3667,3 +3667,36 @@ func TestInject1xBodyProperties(t *testing.T) {
...
+func TestSettingSyncInfo(t *testing.T) {
+	db, ctx := setupTestDB(t)
+	defer db.Close(ctx)
+	collection, _ := GetSingleDatabaseCollectionWithUser(ctx, t, db)
+	ds := collection.GetCollectionDatastore()
+
+	require.NoError(t, base.SetSyncInfoMetaVersion(ds, 1))
+	require.NoError(t, base.SetSyncInfoMetadataID(ds, "someID"))
+
+	// assert that after above operations meta version ius preserved after setting of metadataID
```
torcolvin:
```suggestion
	// assert that after above operations meta version is preserved after setting of metadataID
```

db/database_test.go
```diff
@@ -3667,3 +3667,36 @@ func TestInject1xBodyProperties(t *testing.T) {
 	assert.Equal(t, "value", resBody["key"])
 	assert.True(t, resBody[BodyDeleted].(bool))
 }
+
+// TestSettingSyncInfo:
+//   - Purpose of the test is to call both SetSyncInfoMetaVersion + SetSyncInfoMetadataID in different orders
+//     asserting that the operations preserve the metadataID/metaVersion
+func TestSettingSyncInfo(t *testing.T) {
```
torcolvin, line 3675: I commented about this above, but I actually think that this test will probably read better if it is a parameterized test which just contains the starting state of a document.

torcolvin (commented):
Lots of smallish comments about handling errors and logging.

I searched `fCtx` and made sure that log messages were at the appropriate level and an `Info` or `Debug` message doesn't say something about the implementation unless that is necessary.

I didn't comment on this individually and you could consider skipping for this PR, but for future reading the code, I think it's helpful to add to docstrings to the public methods. These also show up in editors to give hints as to what the function is doing when you read the code, without having to open function and read it, which I find helpful.

base/constants_syncdocs.go
gregns1: I do have tests using this in `TestSettingSyncInfo` but will add more test cases

db/background_mgr_attachment_migration.go
gregns1: I will remove, I wanted to be extra safe given this handling is also included in the resync dcp task but will remove.

db/background_mgr_attachment_migration.go
gregns1: I'm not sure I follow, how can we do this if we fail on unmarshalling the doc?

db/database.go
```diff
@@ -2567,3 +2568,31 @@ func (db *Database) DataStoreNames() base.ScopeAndCollectionNames {
...
+func (db *DatabaseContext) GetCollectionIDs() []uint32 {
+	collIDs := make([]uint32, 0)
+	for id, _ := range db.CollectionByID {
+		collIDs = append(collIDs, id)
+	}
+	return collIDs
+}
+
+// PurgeDCPMetadata will purge all DCP metadata from previous run in the bucket, used to reset dcp client to 0
+func PurgeDCPMetadata(ctx context.Context, database *DatabaseContext, metadataKeyPrefix string, taskID string) error {
```
torcolvin:
```suggestion
// PurgeDCPCheckpoints will purge all DCP checkpoints.
func PurgeDCPCheckpoints(ctx context.Context, database *DatabaseContext, checkpointPrefix string) error {
```

This function is now more generic than background tasks, so I think we should move `base.InfofCtx(ctx, base.KeyDCP, "Purging invalid checkpoints for background task run %s", taskID)` outside. I also think we should jsut checkpointPrefix since otherwise this reads more like `base.MetadataKeys`.

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,325 @@
...
+
+		// We only want to process full docs. Not any sync docs.
+		if strings.HasPrefix(docID, base.SyncDocPrefix) {
+			return true
+		}
+
+		a.DocsProcessed.Add(1)
+		syncData, _, _, err := UnmarshalDocumentSyncDataFromFeed(event.Value, event.DataType, collection.userXattrKey(), false)
+		if err != nil {
+			failProcess(err, "error unmarshaling document %s with migrationID %s: %v", base.UD(docID), a.MigrationID, err)
```
torcolvin:
```suggestion
			failProcess(err, "[%s] error unmarshaling document %s: %v, stopping attachment migration.", migrationLoggingID, base.UD(docID), err)
```

If this does fail, we want to be able to easily grep and find the error message with like messages.

db/background_mgr_attachment_migration.go
```diff
@@ -0,0 +1,325 @@
...
+		}
+
+		collCtx := collection.AddCollectionContext(ctx)
+		collWithUser := &DatabaseCollectionWithUser{
+			DatabaseCollection: collection,
+		}
+		// xattr migration to take place
+		err = collWithUser.MigrateAttachmentMetadata(collCtx, docID, event.Cas, syncData)
+		if err != nil {
+			failProcess(err, "error migrating document attachment metadata for doc: %s with migrationID %s: %v", base.UD(docID), a.MigrationID, err)
```
torcolvin:
```suggestion
			failProcess(err, "[%s] error migrating document attachment metadata for doc: %s: %v", migrationLoggingID, base.UD(docID), err)
```

db/database.go
```diff
@@ -2567,3 +2568,31 @@ func (db *Database) DataStoreNames() base.ScopeAndCollectionNames {
...
 }
+
+// GetCollectionIDs will return all collection IDs for all collections configured on the database
+func (db *DatabaseContext) GetCollectionIDs() []uint32 {
+	collIDs := make([]uint32, 0)
+	for id, _ := range db.CollectionByID {
+		collIDs = append(collIDs, id)
+	}
+	return collIDs
+}
```
torcolvin: from `exp/x/maps` package, and then a bit different in go 1.23 std library maps with iterators: `slices.Collect(maps.Keys(db.CollectionByID)`

```suggestion
	return maps.Keys(db.CollectionByID)
}
```

torcolvin (commented):
LGTM, just a few last minute comments.

torcolvin (approved):

PR #7127, opened 2024-09-24
CBG-4187: Add stat to track number of assertion failures

bbrks:
CBG-4187

Add stat to track number of assertion failures. Plus more Assertion enhancements:
- Enforced 'Assertion failure:' message prefix for searchability.
- Swapped warn to error for dev-mode, since we're asking for failures to be reported and want longer-term visibility in collected logs.
- Replace dev-mode only stat with the new global stat.
- Improve description with instruction for what to do if observed.
- Improve godoc comments for AssertfCtx to make it more obvious that it does not do control flow and is up to the caller to do so.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3001/

### DIFF
--- a/base/stats.go
+++ b/base/stats.go
@@ -89,6 +89,7 @@ const (
 	StatAddedVersion3dot2dot1     = "3.2.1"
 	StatAddedVersion3dot2dot2     = "3.2.2"
 	StatAddedVersion3dot2dot3     = "3.2.3"
+	StatAddedVersion3dot2dot4     = "3.2.4"
 	StatAddedVersion3dot3dot0     = "3.3.0"
 
 	StatDeprecatedVersionNotDeprecated = ""
@@ -301,7 +302,7 @@ func (g *GlobalStat) initResourceUtilizationStats() error {
 	if err != nil {
 		return err
 	}
-	resUtil.AssertionFailCount, err = NewIntStat(ResourceUtilizationSubsystem, "assertion_fail_count", StatUnitNoUnits, AssertionFailCountDesc, StatAddedVersion3dot2dot1, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, nil, nil, prometheus.CounterValue, 0)
+	resUtil.AssertionFailCount, err = NewIntStat(ResourceUtilizationSubsystem, "assertion_fail_count", StatUnitNoUnits, AssertionFailCountDesc, StatAddedVersion3dot2dot4, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, nil, nil, prometheus.CounterValue, 0)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


base/stats_descriptions.go
```diff
@@ -64,7 +64,8 @@ const (
 
 	SystemMemoryTotalDesc = "The total memory available on the system in bytes."
 
-	WarnCountDesc = "The total number of warnings logged."
+	WarnCountDesc          = "The total number of warnings logged."
+	AssertionFailCountDesc = "The total number of assertion failures logged."
```
torcolvin: Should we say what an assertion failure is? Something like:

Assertions count is already included in the warning stat and mark problems that are not caused by users and are a software precondition.

This message isn't good, and but for these failures there's no meaningful action a user could take. That's the difference I see between these and other warnings, which could be things that a user could act upon.

Calling `AssertfCtx` will call `WarnfCtx` directly and this value already increments the WarnCount metric. We also have no way to determine that a message was an assertion failure vs a standard warning from a log file.

I don't know how useful this stat is without being easily able to ID which are assertion failures from the log files without doing a search for the messages. 

I don't know if it makes sense to just add a prefix like "[WRN] Assertion Error: msg" or change `AssertfCtx` -> `WarnfCtx` -> `logTo` to put a new log level? I don't think that a log level change is right because I think that will break the loose log compatibility we have.

Finally, does it actually make sense to log these as errors?

I'm not opposed to the stat, I just want to have it be clear from the description how dev, support, end users are intended to use it. Should this be specially flagged from nutshell, etc as something that needs immediate dev attention? Do we expect a special alerting from capella?

base/stats.go
```diff
@@ -299,6 +299,10 @@ func (g *GlobalStat) initResourceUtilizationStats() error {
 	if err != nil {
 		return err
 	}
+	resUtil.AssertionFailCount, err = NewIntStat(ResourceUtilizationSubsystem, "assertion_fail_count", StatUnitNoUnits, AssertionFailCountDesc, StatAddedVersion3dot2dot1, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, nil, nil, prometheus.CounterValue, 0)
```
bbrks, line 302:
```suggestion
	resUtil.AssertionFailCount, err = NewIntStat(ResourceUtilizationSubsystem, "assertion_fail_count", StatUnitNoUnits, AssertionFailCountDesc, StatAddedVersion3dot3dot0, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, nil, nil, prometheus.CounterValue, 0)
```

base/stats_descriptions.go
bbrks: Rebased and pushed a commit with a series of changes to improve this:
- Enforced `Assertion failure:` message prefix for searchability.
- Swapped warn to error for non dev-mode, since we're asking for failures to be reported and want longer-term visibility in collected logs.
- Improve description with instruction for what to do if observed.
- Replace dev-mode only stat with the new global stat.
- Improve godoc comments for AssertfCtx to make it more obvious that it does not do control flow and is up to the caller to do so.

torcolvin (approved):

PR #7128, opened 2024-09-24
CBG-4236: Create SG log files on startup with 0644 permission bits set

bbrks:
CBG-4236

Creates log files on startup with `0644` permission bits set.
- This restores pre-3.2.0 file permission behaviour. Existing files will retain their existing permission bits.
- New behaviour with this PR: SG creates empty log files at startup time, rather than creating on-demand when the log file needs to be written to.
- Removes now redundant '.SG_write_check' logic that ensures a given log_file_path directory is writable. We now use the log files directly, rather than this proxy.
- 

### DIFF
--- a/base/logging_config.go
+++ b/base/logging_config.go
@@ -41,6 +41,9 @@ const (
 	fileLoggerCollateFlushTimeout    = 10 * time.Millisecond
 
 	rotatedLogDeletionInterval = time.Hour // not configurable
+
+	logFilePermission            = 0644              // rw-r--r--
+	logFileDirWriteCheckFilename = ".SG_write_check" // filename used when writing a write-check test file in the log directory
 )
 
 // ErrUnsetLogFilePath is returned when no log_file_path, or --defaultLogFilePath fallback can be used.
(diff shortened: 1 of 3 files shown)


base/logging_config.go
```diff
@@ -338,13 +304,21 @@ func validateLogFileOutput(logFileOutput string) error {
...
+	}
+
+	// Ensure LogFilePath is a directory. We'll check file permissions when it opens/creates the logfile.
+	if f, err := os.Stat(logFileOutputDirectory); err != nil {
+		return errors.Wrap(err, ErrInvalidLogFilePath.Error())
+	} else if !f.IsDir() {
+		return errors.Wrap(ErrInvalidLogFilePath, "not a directory")
 	}
 
 	// Validate given file is writeable
```
torcolvin:
```suggestion
	// Validate given file is writeable. If the file does not exist, this will open a zero sized log file intentionally to set permissions. The permissions will then be used by lumberjack. Otherwise, lumberjack will use 0600 permissions.
```

torcolvin (commented):
Do you think it's worthwhile to write a test for the following circumstances? I'm specifically concerned that the errors will be opaque if Sync Gateway is started with invalid configuration since each logger will error separately. If this is unclear, separating the directory creation from log file validation might be worthwhile.

- `logFilePath` is specified as a file instead of a directory
- `logFilePath` is not actually writeable. This test might only work on linux by changing permissions to read only.

bbrks:
Added test for 'not a directory', non-writable test already covered with `TestLogFilePathWritable`.

Real-world error at startup time:

```
2024-10-01T18:47:03.040+01:00 [ERR] Couldn't start Sync Gateway: error setting up logging: invalid log file path: mkdir /tmp/sglog/thisisafile: not a directory -- base.FatalfCtx() at logging.go:151
```

torcolvin (commented):
This is fine, but the tests need to pass on windows - the "not a directory" message is different and the file descriptor from os.Create needs to be closed so the temporary directory can be removed.

Will approve after that is done.

torcolvin (approved):

PR #7136, opened 2024-09-26
CBG-4247: refactor in memory format for hlv

gregns1:
CBG-4247

- Change in memory format for cas from string to uint64
- Remove decoded hlv 
- Updated logging in conflict check 
- removed any unused code I saw 
- Change blip format to encode version to hex encoding
- Json unmarshal/marshal functions. These may well be optimised in the delatas PR 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2734/ (test flake)

### DIFF
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -130,7 +130,7 @@ func (entry *LogEntry) SetRevAndVersion(rv channels.RevAndVersion) {
 	entry.RevID = rv.RevTreeID
 	if rv.CurrentSource != "" {
 		entry.SourceID = rv.CurrentSource
-		entry.Version = rv.CurrentVersion
+		entry.Version = base.HexCasToUint64(rv.CurrentVersion)
 	}
 }
 
@@ -494,7 +494,7 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent) {
 					change.DocID = docID
 					change.RevID = atRev.RevTreeID
 					change.SourceID = atRev.CurrentSource
-					change.Version = atRev.CurrentVersion
+					change.Version = base.HexCasToUint64(atRev.CurrentVersion)
 					change.Channels = channelRemovals
 				}
 
--- a/db/changes.go
+++ b/db/changes.go
@@ -489,7 +489,7 @@ func makeChangeEntry(logEntry *LogEntry, seqID SequenceID, channel channels.ID) 
 	// populate CurrentVersion entry if log entry has sourceID and Version populated
 	// This allows current version to be nil in event of CV not being populated on log entry
 	// allowing omitempty to work as expected
-	if logEntry.SourceID != "" && logEntry.Version != "" {
+	if logEntry.SourceID != "" && logEntry.Version != 0 {
 		change.CurrentVersion = &Version{SourceID: logEntry.SourceID, Value: logEntry.Version}
 	}
 	if logEntry.Flags&channels.Removed != 0 {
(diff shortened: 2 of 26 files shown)


db/changes.go
```diff
@@ -489,7 +489,7 @@ func makeChangeEntry(logEntry *LogEntry, seqID SequenceID, channel channels.ID)
 	// populate CurrentVersion entry if log entry has sourceID and Version populated
 	// This allows current version to be nil in event of CV not being populated on log entry
 	// allowing omitempty to work as expected
-	if logEntry.SourceID != "" && logEntry.Version != "" {
+	if logEntry.SourceID != "" && logEntry.Version != 0 {
```
adamcfraser: I think Version=0 is valid (if a bit unusual), so I think changing this check to just `SourceID != ""` is appropriate, what do you think?

db/change_cache.go
```diff
@@ -130,7 +130,7 @@ func (entry *LogEntry) SetRevAndVersion(rv channels.RevAndVersion) {
 	entry.RevID = rv.RevTreeID
 	if rv.CurrentSource != "" {
 		entry.SourceID = rv.CurrentSource
-		entry.Version = rv.CurrentVersion
+		entry.Version = base.HexCasToUint64(rv.CurrentVersion)
```
adamcfraser, line 133: It might be a good idea to stick a comment near the definition of RevAndVersion.CurrentVersion to mention that it needs to be stored as a string to support macro expansion when writing _sync.rev.

db/hybrid_logical_vector.go
```diff
@@ -78,30 +44,21 @@ func ParseVersion(versionString string) (version Version, err error) {
...
-func (v DecodedVersion) String() string {
-	timestamp := string(base.Uint64CASToLittleEndianHex(v.Value))
-	source := base64.StdEncoding.EncodeToString([]byte(v.SourceID))
-	return timestamp + "@" + source
-}
-
 // String returns a version/sourceID pair in CBL string format
 func (v Version) String() string {
-	return v.Value + "@" + v.SourceID
+	return fmt.Sprintf("%x@%s", v.Value, v.SourceID)
```
adamcfraser: If this is commonly used, I think something like `strconv.FormatUint(v.Value, 16) + "@" + v.SourceID` would be more efficient than parsing "%x@%s".

db/hybrid_logical_vector.go
```diff
@@ -114,24 +71,33 @@ func (hlv *HybridLogicalVector) ExtractCurrentVersionFromHLV() *Version {
...
-	PreviousVersions  map[string]string `json:"pv,omitempty"`        // map of previous versions for fast efficient lookup
+	CurrentVersionCAS uint64      // current version cas (or cvCAS) stores the current CAS in little endian hex format at the time of replication
+	ImportCAS         uint64      // Set when an import modifies the document CAS but preserves the HLV (import of a version replicated by XDCR)
+	SourceID          string      // source bucket uuid in (base64 encoded format) of where this entry originated from
+	Version           uint64      // current cas in little endian hex format of the current version on the version vector
+	MergeVersions     HLVVersions // map of merge versions for fast efficient lookup
+	PreviousVersions  HLVVersions // map of previous versions for fast efficient lookup
+}
+
+type BucketVector struct {
```
adamcfraser, line 82: I don't love having a separate struct just for persistence, but we can revisit in the context of the PV delta support work.

adamcfraser (commented):
Generally looks good - just a couple of minor tweaks and it should be good to merge.

db/hybrid_logical_vector.go
gregns1: Yeah will have this sorted in the deltas ticket, it wasn't immediately clear to me the best approach so thought this will do to just unblock cbl for now.

torcolvin (dismissed):

torcolvin (dismissed):

torcolvin (approved):

PR #7140, opened 2024-10-01
CBG-4282 silence non-constant format govet error

torcolvin:
This is a more strict version of govet that is coming in go 1.24 and is already present in golangci-lint 1.24. All of the things flagged were false positives.

- created `NewHTTPError` to cover when there are no arguments. I could have changed this code to `NewHTTPErrorf(status, "%s", msg)` but this seemed weird to me and like you wouldn't be sure why you had to use this code and would look this up.
- Created `FileLogger.log` function but this could easily use `logf("%s", msg)` in the places we are using it like `statsLogger` and `conditionalPrintf` since they are called very irregularly.

I don't feel strongly about these implementation choices.

### DIFF
--- a/auth/oidc.go
+++ b/auth/oidc.go
@@ -302,7 +302,7 @@ func (op *OIDCProvider) InitUserPrefix(ctx context.Context) error {
 // an initial provider metadata discovery and initiates the discovery sync in the background.
 func (op *OIDCProvider) initOIDCClient(ctx context.Context) error {
 	if op == nil {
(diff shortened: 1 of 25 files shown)


gregns1 (dismissed):
Generally looks good to me and think I agree with implementation choices instead of the use of `("%s", msg)`. Should we add more to the comments of the new `log` and `NewHTTPError` functions to add context of how they should be used when there are no arguments to be passed to format string? Happy to merge as is if you think not necessary so will approve as is.

gregns1 (approved):

PR #7146, opened 2024-10-08
CBG-4289 fix import CV value for HLV code

torcolvin:
Created new tests to cover import cases https://docs.google.com/document/d/19mC9-TptN7ynuF5XR9TQ_MckHWsSrjUfI2p71G_DD1U/edit#heading=h.xzdilbaiwy25

The specific bugs initially noticed were around `_vv.ver` getting updated to `newVVEntry.Value = expandMacroCASValueUint64` which was never correct on import. Additional logic was expanded to cover all tests.

For ease of reading this code, the existing tests were ported to a parameterized test first with the buggy logic in place, you can read the commit separately.

Code is also tested with live XDCR.

Also removes `ImportCAS` as part of CBG-4285.

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2746/

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -896,8 +896,9 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
 }
 
 // updateHLV updates the HLV in the sync data appropriately based on what type of document update event we are encountering
-func (db *DatabaseCollectionWithUser) updateHLV(d *Document, docUpdateEvent DocUpdateType) (*Document, error) {
+func (db *DatabaseCollectionWithUser) updateHLV(d *Document, previousMou *MetadataOnlyUpdate, docUpdateEvent DocUpdateType) (*Document, error) {
 
+	hasHLV := d.HLV != nil
 	if d.HLV == nil {
 		d.HLV = &HybridLogicalVector{}
 	}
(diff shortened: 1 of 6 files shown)


db/crud.go
```diff
@@ -896,34 +896,42 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
...
 		// preserve any other logic on the HLV that has been done by the client, only update to cvCAS will be needed
 		d.HLV.CurrentVersionCAS = expandMacroCASValueUint64
-		d.HLV.ImportCAS = 0 // remove importCAS for non-imports to save space
 	case Import:
-		if d.HLV.CurrentVersionCAS == d.Cas {
-			// if cvCAS = document CAS, the HLV has already been updated for this mutation by another HLV-aware peer.
-			// Set ImportCAS to the previous document CAS, but don't otherwise modify HLV
-			d.HLV.ImportCAS = d.Cas
-		} else {
+		// Update HLV if:
```
adamcfraser: I think the description as you've got it muddies the rationale behind the decision making here (simply that we want to update the HLV only if the current document cas isn't already accounted for in the HLV + mou).  
```suggestion
		// Do not update HLV if the current document version (cas) is already included in the existing HLV, 
		// as either:
		//    1. _vv.cvCAS == document.cas (current mutation is already present as cv), or
		//    2. _mou.cas == document.cas (current mutation is already present as cv, and was imported on a different cluster)
```

db/crud.go
```diff
@@ -896,34 +896,42 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
...
+		// 	- no _mou or _mou.cas != document cas
+
+		// Do not update HLV if this is an XDCR migration:
+		//
+		// - document was written by another cluster
+		// 	- _vv.cvCAS != document cas, no _mou or _mou.cas != document cas
+		// - document was written by another cluster, itself having an undergone an import
+		// 	- _vv.cvCAS == document cas and _mou.cas != document cas
+
+		if !hasHLV || (d.HLV.CurrentVersionCAS != d.Cas && (previousMou == nil || base.HexCasToUint64(previousMou.CAS) != d.Cas)) {
```
adamcfraser: If this can be rewritten to match the comment suggestion I think that would be more readable.  Even something like this maybe:
```
cvCASMatch := hasHLV && d.HLV.CurrentVersionCAS == d.Cas
mouMatch := previousMou != nil && base.HexCasToUint64(previousMou.CAS) == d.Cas)
if !cvCASMatch && !mouMatch {
```

adamcfraser (changes requested):
Functionally looks ok, some suggestions for clarity on the comments/conditions.

db/crud.go
torcolvin: There is one additional caveat, which I've express in the changed code, but I'm not sure if it is expressed correctly in the comment. There could be no existing `_vv` but an `_mou`, in which case we are going to write one. I'm not sure this is terribly likely, but I wanted to include it if a document did get corrupted or someone else is using `_mou` and not using `_vv` for some reason.

adamcfraser (approved):

PR #7149, opened 2024-10-09
Fix redocly plugins

osfameron:
These crashed when description was missing, so add defaulting/logic to fix.
Enable the excise-rbac for the other APIs for the *-capella targets.

Ping @torcolvin to review - I don't think this checklist applies to this PR?
I can create a CBG if required.

CBG-0000

Describe your PR here...

## Pre-review checklist

- [ ] Link upstream PRs

### DIFF
--- a/.redocly.yaml
+++ b/.redocly.yaml
@@ -18,6 +18,7 @@ apis:
   admin-capella:
     root: "./docs/api/admin-capella.yaml"
     decorators:
+      plugin/excise-rbac-capella: on
       filter-out:
         property: x-capella
         value: false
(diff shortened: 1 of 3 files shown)


torcolvin:
I think I know why the `redocly_preview_links` doesn't work but this only publishes not generated previews back to the PR. This is because this is a PR from a fork, and we always do PRs from branches like CBG-xxx. I granted the docs team access to be able to push branches like this. You can't merge without a PR.

We like all commits to have a jira ticket for tracking but it can be a DOC ticket, since you are doing the work, or a CBG ticket. No preference. Since this is a commit against main which is tracking 3.3, after this is merged, we typically create a backport ticket by cloning the existing ticket and marking as [3.2.1 backport]. Then once that is merged, I move the tag for docs. This process should probably be described in our meta doc docs.

Finally, this looks good except that I think `info.title` should be modified for `public-capella` since it says `Sync Gateway` instead of `App Services`. I verified with `npx @redocly/cli@latest preview-docs public-capella`.

osfameron:
Closing in favour of #7153

PR #7150, opened 2024-10-09
CBG-4293 recognize alma/rocky and other rhel variants

torcolvin:
Change `ostype` shell function to recognize `ID_LIKE` containing `rhel` as a red hat based linux.

An alternative fix here would be to add `alma* | rocky*` to `redhat* | rhel* | centos | ol)` line, but I worry that yet another rhel variant would come out and we'd forget to add it.

The following is the contents of `/etc/os-release`.

| OS | ID | ID_LIKE |
| ---| ---| -------- |
| Amazon Linux 2 | amzn | "centos rhel fedora" |
| Amazon Linux 2023 | amzn | "fedora" |
| Redhat 8, 9 | rhel | "fedora" |
| Centos 7, 8 (unsupported) | centos | "rhel fedora" |
| Alma 8, 9 | almalinux | "rhel centos fedora" |
| Rocky 8, 9 | rockylinux | "rhel centos fedora" |
| Debian 7-12 | debian | <unspecified> |
| Ubuntu 16.04-24.10 | ubuntu | "debian" |

This method preserve some support that already exists, like accounting for Oracle Linux (ol), even though this isn't a supported platform. I made this more general to support matching a future rhel fork that I haven't come up with.

# Scoping

This PR could be descoped to the alternative fix to just change `redhat* | rhel* | centos | ol)`  to ``redhat* | rhel* | centos | ol | alma* | rocky* )` 

Alternatively, this PR could be enhanced:

-  combine debian and ubuntu  and rhel sections now that every OS we support uses systemd. The only difference with debian, ubuntu and rhel variants is that the systemd file uses `/usr/bin/bash` on Debian-like systems and `/bin/bash` on rhel systems, but `/usr/bin/bash` and `/bin/bash` exists on all supported platforms.
- use a common file to define `ostype` between all three shell scripts to avoid duplication.

# Manual Testing:

I don't know how to build packages exactly (I think you need a toy build, which I don't know how to do).

For rhel, alma, rocky, ubuntu, debian, amazon (two versions) I downloaded existing 3.2.0 package in a docker container. Then I copied over `service/*sh` to `/opt/couchbase-sync-gateway/service/` and ran the scripts. The error behavior would be easily observable with the scripts exiting non zero.

# Possible automated regression testing (not implemented)

- Move `ostype` to common file like `service_utils.sh` that is source by other files.
- Create github action to use docker images for all supported platforms and run `./service/sync_gateway_service_install.sh --servicecmd` and validate that it returns non zero.

### DIFF
--- a/.github/workflows/service.yml
+++ b/.github/workflows/service.yml
@@ -42,6 +42,6 @@ jobs:
         with:
           go-version: 1.22.5
       - name: "Build Sync Gateway"
-        run: env SG_EDITON=EE ./build.sh
+        run: go build -o ./bin/sync_gateway ./...
       - name: "Run test scripts"
         run: ./integration-test/service-install-tests.sh
--- a/.github/workflows/service.yml
+++ b/.github/workflows/service.yml
@@ -42,6 +42,6 @@ jobs:
         with:
           go-version: 1.22.5
       - name: "Build Sync Gateway"
-        run: go build -o ./bin/sync_gateway ./...
+        run: mkdir -p ./bin && go build -o ./bin/sync_gateway ./...
       - name: "Run test scripts"
         run: ./integration-test/service-install-tests.sh
--- a/.github/workflows/service.yml
+++ b/.github/workflows/service.yml
@@ -42,6 +42,6 @@ jobs:
         with:
           go-version: 1.22.5
       - name: "Build Sync Gateway"
-        run: mkdir -p ./bin && go build -o ./bin/sync_gateway ./...
+        run: mkdir -p ./bin && go build -o ./bin ./...
       - name: "Run test scripts"
         run: ./integration-test/service-install-tests.sh
--- a/.github/workflows/service.yml
+++ b/.github/workflows/service.yml
@@ -1,0 +1,41 @@
+# Copyright 2022-Present Couchbase, Inc.
+#
+# Use of this software is governed by the Business Source License included in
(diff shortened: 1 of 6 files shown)


bbrks:
This seems like a big change vs. adding an extra two options for our supported platforms... How confident are you in these changes for 3.2.1?

torcolvin:
> This seems like a big change vs. adding an extra two options for our supported platforms... How confident are you in these changes for 3.2.1?

I agree with your point about too risky for 3.2.1. I irrationally fear too much the fact that someone else is going to create another rhel port name.

I did add an integration test which does catch the original problem and will help a bit with regression testing, without going through the full process of building the packages and extracting them, which I do not know how to do.

The full refactor I would prefer to do for the service scripts is here https://github.com/couchbase/sync_gateway/compare/main...CBG-4293-full-refactor but I do not think it is worthwhile to actually pursue, we can just add a quicker regression test when needed to support a new platform.

bbrks:
> I agree with your point about too risky for 3.2.1. I irrationally fear too much the fact that someone else is going to create another rhel port name.

If that were the case, we'd have to adopt it as a supported platform and at that point it doesn't feel like much of a stretch to include that future distro in this list as well. The chances of that happening seem lower than us introducing or breaking something that we weren't testing well.

bbrks (approved):

PR #7151, opened 2024-10-11
CBG-4212: Trigger attachment migration job upon db startup

gregns1:
CBG-4212

- Changes to `InitSyncInfo` to return requires migration if the collection needs it (if sync info is missing altogether or is `metaVersion` is not equal to 4.0)
- Stores collections that require migration on the db context 
- Initiates a new Attachment migration manager upon db startup, will trigger this job based upon existence of collections that require migrating 
- Changes as purging of dcp checkpoints weren't happening as expected
- Changed to way we build collection IDs for dcp client. Changes will mean we only run dcp feeds against collections wee need for migration. This will make less overhead for updating a db to include a new collection when the other collection(s) have already been migrated. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2766/

### DIFF
--- a/base/constants_syncdocs.go
+++ b/base/constants_syncdocs.go
@@ -385,13 +385,14 @@ type SyncInfo struct {
 //  1. If syncInfo doesn't exist, it is created for the specified metadataID
 //  2. If syncInfo exists with a matching metadataID, returns requiresResync=false
 //  3. If syncInfo exists with a non-matching metadataID, returns requiresResync=true
-func InitSyncInfo(ds DataStore, metadataID string) (requiresResync bool, err error) {
+//     If syncInfo exists and has metaVersion equal to 4.0, return requiresAttachmentMigration=flase, else requiresAttachmentMigration=true
+func InitSyncInfo(ds DataStore, metadataID string) (requiresResync bool, requiresAttachmentMigration bool, err error) {
 
 	var syncInfo SyncInfo
 	_, fetchErr := ds.Get(SGSyncInfo, &syncInfo)
 	if IsDocNotFoundError(fetchErr) {
 		if metadataID == "" {
-			return false, nil
+			return false, true, nil
 		}
 		newSyncInfo := &SyncInfo{MetadataID: metadataID}
 		_, addErr := ds.Add(SGSyncInfo, 0, newSyncInfo)
@@ -399,18 +400,24 @@ func InitSyncInfo(ds DataStore, metadataID string) (requiresResync bool, err err
 			// attempt new fetch
 			_, fetchErr = ds.Get(SGSyncInfo, &syncInfo)
 			if fetchErr != nil {
-				return true, fmt.Errorf("Error retrieving syncInfo (after failed add): %v", fetchErr)
+				return true, true, fmt.Errorf("Error retrieving syncInfo (after failed add): %v", fetchErr)
 			}
 		} else if addErr != nil {
-			return true, fmt.Errorf("Error adding syncInfo: %v", addErr)
+			return true, true, fmt.Errorf("Error adding syncInfo: %v", addErr)
 		}
 		// successfully added
-		return false, nil
+		return false, true, nil
 	} else if fetchErr != nil {
-		return true, fmt.Errorf("Error retrieving syncInfo: %v", fetchErr)
+		return true, true, fmt.Errorf("Error retrieving syncInfo: %v", fetchErr)
+	}
+	// check for meta version, if we don't have meta version of 4.0 we need to run migration job
+	if syncInfo.MetaDataVersion != "4.0" {
+		requiresAttachmentMigration = true
+	} else {
+		requiresAttachmentMigration = false
 	}
 
-	return syncInfo.MetadataID != metadataID, nil
+	return syncInfo.MetadataID != metadataID, requiresAttachmentMigration, nil
 }
 
 // SetSyncInfoMetadataID sets syncInfo in a DataStore to the specified metadataID, preserving metadata version if present
--- a/base/constants_syncdocs.go
+++ b/base/constants_syncdocs.go
@@ -11,6 +11,7 @@ licenses/APL2.txt.
 package base
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"strconv"
@@ -386,7 +387,7 @@ type SyncInfo struct {
 //  2. If syncInfo exists with a matching metadataID, returns requiresResync=false
 //  3. If syncInfo exists with a non-matching metadataID, returns requiresResync=true
 //     If syncInfo exists and has metaDataVersion greater than or equal to 4.0, return requiresAttachmentMigration=false, else requiresAttachmentMigration=true to bring migrate metadata attachments.
-func InitSyncInfo(ds DataStore, metadataID string) (requiresResync bool, requiresAttachmentMigration bool, err error) {
+func InitSyncInfo(ctx context.Context, ds DataStore, metadataID string) (requiresResync bool, requiresAttachmentMigration bool, err error) {
 
 	var syncInfo SyncInfo
 	_, fetchErr := ds.Get(SGSyncInfo, &syncInfo)
@@ -406,7 +407,7 @@ func InitSyncInfo(ds DataStore, metadataID string) (requiresResync bool, require
 			return true, true, fmt.Errorf("Error adding syncInfo: %v", addErr)
 		}
 		// successfully added
-		requiresAttachmentMigration, err = CompareMetadataVersion(syncInfo.MetaDataVersion)
+		requiresAttachmentMigration, err = CompareMetadataVersion(ctx, syncInfo.MetaDataVersion)
 		if err != nil {
 			return syncInfo.MetadataID != metadataID, true, err
 		}
@@ -415,7 +416,7 @@ func InitSyncInfo(ds DataStore, metadataID string) (requiresResync bool, require
 		return true, true, fmt.Errorf("Error retrieving syncInfo: %v", fetchErr)
 	}
 	// check for meta version, if we don't have meta version of 4.0 we need to run migration job
-	requiresAttachmentMigration, err = CompareMetadataVersion(syncInfo.MetaDataVersion)
+	requiresAttachmentMigration, err = CompareMetadataVersion(ctx, syncInfo.MetaDataVersion)
 	if err != nil {
 		return syncInfo.MetadataID != metadataID, true, err
 	}
@@ -477,31 +478,37 @@ func SerializeIfLonger(name string, length int) string {
 
 // CompareMetadataVersion Will build comparable build version for comparison with meta version defined in syncInfo, then
 // will return true if we require attachment migration, false if not.
-func CompareMetadataVersion(metaVersion string) (bool, error) {
+func CompareMetadataVersion(ctx context.Context, metaVersion string) (bool, error) {
 	var syncInfoVersion *ComparableBuildVersion
+	var requiresAttachmentMigration bool
 	var err error
 	if metaVersion != "" {
 		syncInfoVersion, err = NewComparableBuildVersionFromString(metaVersion)
 		if err != nil {
 			return true, err
 		}
-	}
-	requiresAttachmentMigration, err := CheckRequireAttachmentMigration(syncInfoVersion)
-	if err != nil {
-		return true, err
+		requiresAttachmentMigration, err = CheckRequireAttachmentMigration(ctx, syncInfoVersion)
+		if err != nil {
+			return true, err
+		}
+	} else {
+		// no meta version passed in, thus attachment migration should take place
+		requiresAttachmentMigration = true
 	}
 	return requiresAttachmentMigration, nil
 }
 
 // CheckRequireAttachmentMigration will return true if current metaVersion < 4.0.0, else false
-func CheckRequireAttachmentMigration(version *ComparableBuildVersion) (bool, error) {
+func CheckRequireAttachmentMigration(ctx context.Context, version *ComparableBuildVersion) (bool, error) {
 	if version == nil {
-		return true, nil
+		AssertfCtx(ctx, "failed to build comparable build version for syncInfo metaVersion")
+		return true, fmt.Errorf("corrupt syncInfo metaVersion value")
 	}
 	minVerStr := "4.0.0" // minimum meta version that needs to be defined for metadata migration. Any version less than this will require attachment migration
 	minVersion, err := NewComparableBuildVersionFromString(minVerStr)
 	if err != nil {
-		return true, nil
+		AssertfCtx(ctx, "failed to build comparable build version for minimum version for attachment migration")
+		return true, err
 	}
 
 	if minVersion.AtLeastMinorDowngrade(version) {
--- a/base/constants_syncdocs.go
+++ b/base/constants_syncdocs.go
@@ -479,23 +479,15 @@ func SerializeIfLonger(name string, length int) string {
 // CompareMetadataVersion Will build comparable build version for comparison with meta version defined in syncInfo, then
 // will return true if we require attachment migration, false if not.
 func CompareMetadataVersion(ctx context.Context, metaVersion string) (bool, error) {
-	var syncInfoVersion *ComparableBuildVersion
-	var requiresAttachmentMigration bool
-	var err error
-	if metaVersion != "" {
-		syncInfoVersion, err = NewComparableBuildVersionFromString(metaVersion)
-		if err != nil {
-			return true, err
-		}
-		requiresAttachmentMigration, err = CheckRequireAttachmentMigration(ctx, syncInfoVersion)
-		if err != nil {
-			return true, err
-		}
-	} else {
+	if metaVersion == "" {
 		// no meta version passed in, thus attachment migration should take place
-		requiresAttachmentMigration = true
+		return true, nil
 	}
-	return requiresAttachmentMigration, nil
+	syncInfoVersion, err := NewComparableBuildVersionFromString(metaVersion)
+	if err != nil {
+		return true, err
+	}
+	return CheckRequireAttachmentMigration(ctx, syncInfoVersion)
 }
 
 // CheckRequireAttachmentMigration will return true if current metaVersion < 4.0.0, else false
--- a/base/constants_syncdocs.go
+++ b/base/constants_syncdocs.go
@@ -385,7 +385,7 @@ type SyncInfo struct {
 //  1. If syncInfo doesn't exist, it is created for the specified metadataID
 //  2. If syncInfo exists with a matching metadataID, returns requiresResync=false
 //  3. If syncInfo exists with a non-matching metadataID, returns requiresResync=true
-//     If syncInfo exists and has metaVersion equal to 4.0, return requiresAttachmentMigration=flase, else requiresAttachmentMigration=true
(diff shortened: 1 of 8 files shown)


base/constants_syncdocs.go
```diff
@@ -385,32 +385,39 @@ type SyncInfo struct {
 //  1. If syncInfo doesn't exist, it is created for the specified metadataID
 //  2. If syncInfo exists with a matching metadataID, returns requiresResync=false
 //  3. If syncInfo exists with a non-matching metadataID, returns requiresResync=true
-func InitSyncInfo(ds DataStore, metadataID string) (requiresResync bool, err error) {
+//     If syncInfo exists and has metaVersion equal to 4.0, return requiresAttachmentMigration=flase, else requiresAttachmentMigration=true
```
torcolvin:
```suggestion
//     If syncInfo exists and has metaDataVersion equal to 4.0, return requiresAttachmentMigration=false, else requiresAttachmentMigration=true to bring migrate metadata attachments.
```

base/constants_syncdocs.go
```diff
@@ -385,32 +385,39 @@ type SyncInfo struct {
...
 		}
 		// successfully added
-		return false, nil
+		return false, true, nil
 	} else if fetchErr != nil {
-		return true, fmt.Errorf("Error retrieving syncInfo: %v", fetchErr)
+		return true, true, fmt.Errorf("Error retrieving syncInfo: %v", fetchErr)
+	}
+	// check for meta version, if we don't have meta version of 4.0 we need to run migration job
+	if syncInfo.MetaDataVersion != "4.0" {
```
torcolvin: This should use a function like `CheckMinorDowngrade` but implemented for a major version less than 4.x. You should write a test for 4.0, 4.1, 5.0, 5.2, etc.

db/background_mgr_attachment_migration.go
```diff
@@ -325,3 +340,26 @@ func (a *AttachmentMigrationManager) resetDCPMetadataIfNeeded(ctx context.Contex
 	}
 	return nil
 }
+
+// getCollectionIDsForMigration will get all collection IDs required for DCP client on migration run
+func getCollectionIDsForMigration(db *DatabaseContext) ([]uint32, error) {
+	collectionIDs := make([]uint32, 0)
+
+	// if all collections are included in RequireAttachmentMigration then we need to run against all collections,
```
torcolvin, line 350: We are doing a weird check here for `len(db.CollectionByID) == len(db.RequireAttachmentMigration)` because there's an unlikely chance these could actually be _out of date_ collections. I think you should just replace this with:

```
if len(db.RequireAttachmentMigration) == 0 {
	return db.GetCollectionIDs()
}
```

for readability. Otherwise it seems fine to iterate through the list again.

db/background_mgr_attachment_migration.go
```diff
@@ -325,3 +340,26 @@ func (a *AttachmentMigrationManager) resetDCPMetadataIfNeeded(ctx context.Contex
...
+	// being triggered by rest api (even after job was previously completed)
+	if len(db.CollectionByID) == len(db.RequireAttachmentMigration) || len(db.RequireAttachmentMigration) == 0 {
+		// get all collection IDs
+		collectionIDs = db.GetCollectionIDs()
+	} else {
+		// iterate through and grab collectionIDs we need
+		for _, v := range db.RequireAttachmentMigration {
+			collection, err := db.GetDatabaseCollection(v.ScopeName(), v.CollectionName())
+			if err != nil {
+				return nil, fmt.Errorf("failed to find ID for collection %s.%s", base.MD(v.ScopeName()).Redact(), base.MD(v.CollectionName()).Redact())
```
torcolvin: nit:

```suggestion
				return nil, base.RedactErrorf("failed to find ID for collection %s.%s", base.MD(v.ScopeName()), base.MD(v.CollectionName())
```

db/background_mgr_attachment_migration_test.go
```diff
@@ -82,9 +78,9 @@ func TestAttachmentMigrationTaskMixMigratedAndNonMigratedDocs(t *testing.T) {
 
 }
 
-func getAttachmentMigrationStats(resyncManager BackgroundManagerProcessI) ResyncManagerResponseDCP {
-	var resp ResyncManagerResponseDCP
-	rawStatus, _, _ := resyncManager.GetProcessStatus(BackgroundManagerStatus{})
+func getAttachmentMigrationStats(migrationManager BackgroundManagerProcessI) AttachmentMigrationManagerResponse {
+	var resp AttachmentMigrationManagerResponse
+	rawStatus, _, _ := migrationManager.GetProcessStatus(BackgroundManagerStatus{})
 	_ = base.JSONUnmarshal(rawStatus, &resp)
```
torcolvin: Check the errors here

db/database.go
```diff
@@ -2479,6 +2488,15 @@ func (db *DatabaseContext) StartOnlineProcesses(ctx context.Context) (returnedEr
 	}
 	db.backgroundTasks = append(db.backgroundTasks, bgtSyncTime)
 
+	db.AttachmentMigrationManager = NewAttachmentMigrationManager(db)
+	// if we have collections requiring migration, run the job
+	if len(db.RequireAttachmentMigration) > 0 && !db.BucketSpec.IsWalrusBucket() {
+		err := db.AttachmentMigrationManager.Start(ctx, nil)
+		if err != nil {
+			base.WarnfCtx(ctx, "Error trying to migrate attachments for %s with error: %v", db.Name, err)
```
torcolvin, line 2496: If we can't start the attachment manager, we should fail taking the database online, this is a fatal error. `Start` is asynchronous.

This might be worth logging a specific Info message for because we are taking an explicit action., like "Migrating attachment metadata automatically to Sync Gateway 4.0+ for collections %s"

db/database.go
```diff
@@ -2479,6 +2488,15 @@ func (db *DatabaseContext) StartOnlineProcesses(ctx context.Context) (returnedEr
 	}
 	db.backgroundTasks = append(db.backgroundTasks, bgtSyncTime)
 
+	db.AttachmentMigrationManager = NewAttachmentMigrationManager(db)
+	// if we have collections requiring migration, run the job
+	if len(db.RequireAttachmentMigration) > 0 && !db.BucketSpec.IsWalrusBucket() {
+		err := db.AttachmentMigrationManager.Start(ctx, nil)
```
torcolvin, line 2494: I think this might need to be `Run` because we don't want to take the database online while attachments are in a mixed format? Or are we allowing that?

db/database_test.go
```diff
@@ -3715,3 +3715,41 @@ func TestSettingSyncInfo(t *testing.T) {
...
+//   - Purpose is to test code pathways inside the InitSyncInfo function will return requires attachment migration
+//     as expected.
+//   - Have a test case for meta version being 4.0 in the bucket and assert it returns false in that scenario
+func TestRequireMigration(t *testing.T) {
+	db, ctx := setupTestDB(t)
+	defer db.Close(ctx)
+	collection, _ := GetSingleDatabaseCollectionWithUser(ctx, t, db)
+	ds := collection.GetCollectionDatastore()
+
+	// test no sync info in bucket and set metadataID, returns requireMigration
```
torcolvin, line 3802: I think these tests would be easier to read and reason about if they were parameterized tests:

We want each test, with `InitSyncInfo` with matching metadata ID, empty metadata id, non matching metdata id. These will get tested for no metadataVersion, 4.x metadataVersion, 5.x metadataVersion (and other cases you think are important, as above).

We don't need to create a database here, we can use a bucket directly. Between each test, we should be sure to delete `base.SyncInfo` but I don't think each test needs a new bucket.

db/util_testing.go
```diff
@@ -917,3 +917,13 @@ func MoveAttachmentXattrFromGlobalToSync(t *testing.T, ctx context.Context, docI
 	_, err = dataStore.WriteWithXattrs(ctx, docID, 0, cas, value, map[string][]byte{base.SyncXattrName: newSync}, []string{base.GlobalXattrName}, opts)
 	require.NoError(t, err)
 }
+
+func RequireBackgroundManagerState(t *testing.T, ctx context.Context, mgr *BackgroundManager, expState BackgroundProcessState) {
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		var status BackgroundManagerStatus
+		rawStatus, err := mgr.GetStatus(ctx)
+		require.NoError(c, err)
+		require.NoError(c, base.JSONUnmarshal(rawStatus, &status))
```
torcolvin: in testify, you have to use assert.NoError until https://github.com/stretchr/testify/commit/b074924938f86d417f1c9a845c7e8b0784d7f937 gets merged and we update it, otherwise it panics the test haness.

rest/attachment_migration_test.go
```diff
@@ -0,0 +1,455 @@
...
+//   - Create a db
+//   - Grab attachment migration manager and assert it has run upon db startup
+//   - Assert job has written syncInfo metaVersion as expected to the bucket
+func TestMigrationJobStartOnDbStart(t *testing.T) {
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("rosmar does not support DCP client, pending CBG-4249")
+	}
+	rt := NewRestTesterPersistentConfig(t)
+	defer rt.Close()
+	ctx := base.TestCtx(t)
```
torcolvin: you can use `rt.Context()` here.

rest/attachment_migration_test.go
```diff
@@ -0,0 +1,455 @@
...
+	mgr := dbCtx.AttachmentMigrationManager
+
+	// wait for migration job to finish
+	db.RequireBackgroundManagerState(t, ctx, mgr, db.BackgroundProcessStateCompleted)
+
+	// assert that sync info with metadata version written to the collection
+	syncInf := base.SyncInfo{}
+	_, err := ds.Get(base.SGSyncInfo, &syncInf)
+	require.NoError(t, err)
+	assert.Equal(t, "4.0", syncInf.MetaDataVersion)
```
torcolvin: This code is going to fail as soon as we bump to 4.1, watch where you use 4.0 as a string that isn't for test purposes. I'd be inclined _not_ to use 4.0 as an explicit string anywhere except for where you want to do a major version test.

rest/attachment_migration_test.go
```diff
@@ -0,0 +1,455 @@
...
+		_, writeErr := ds0.WriteWithXattrs(ctx, key, 0, 0, bodyBytes, xattrsInput, nil, opts)
+		require.NoError(t, writeErr)
+	}
+	for i := 0; i < 4000; i++ {
+		key := fmt.Sprintf("%s_%d", t.Name()+"greg", i)
+		xattrsInput := map[string][]byte{
+			"_xattr": []byte(`{"some":"xattr"}`),
+		}
+		_, writeErr := ds1.WriteWithXattrs(ctx, key, 0, 0, bodyBytes, xattrsInput, nil, opts)
+		require.NoError(t, writeErr)
```
torcolvin: I'd probably combine this into the above loop unless you care about the order.

rest/attachment_migration_test.go
```diff
@@ -0,0 +1,455 @@
...
+
+	resp := rt.CreateDatabase(dbName, dbConfig)
+	RequireStatus(t, resp, http.StatusCreated)
+
+	dbCtx := rt.GetDatabase()
+	mgr := dbCtx.AttachmentMigrationManager
+	assert.Len(t, dbCtx.RequireAttachmentMigration, 1)
+	scNames := base.ScopeAndCollectionNames{}
+	scName := base.ScopeAndCollectionName{Scope: scope, Collection: collection1}
+	scNames = append(scNames, scName)
```
torcolvin:
```suggestion
	scNames := base.ScopeAndCollectionNames{base.ScopeAndCollectionName{Scope: scope, Collection: collection1}}
```

rest/attachment_migration_test.go
```diff
@@ -0,0 +1,455 @@
...
+	scope := dataStoreNames[0].ScopeName()
+	collection1 := dataStoreNames[0].CollectionName()
+	collection2 := dataStoreNames[1].CollectionName()
+	delete(scopesConfigC1Only[scope].Collections, collection2)
+
+	scopesConfigBothCollection := GetCollectionsConfig(t, tb, 2)
+
+	// Create a db1 with one collection initially
+	dbConfig := rt.NewDbConfig()
+	dbConfig.AutoImport = false
```
torcolvin, line 114: Why we setting to automatic import? Does this hurt running the tests? Do we need to run the tests with import? Just mark why the decision is made.

rest/attachment_migration_test.go
```diff
@@ -0,0 +1,455 @@
...
+	RequireStatus(t, resp, http.StatusCreated)
+
+	// wait for attachment migration job to start and finish
+	dbCtx = rt.GetDatabase()
+	mgr = dbCtx.AttachmentMigrationManager
+	assert.Len(t, dbCtx.RequireAttachmentMigration, 2)
+	scName = base.ScopeAndCollectionName{Scope: scope, Collection: collection2}
+	scNames = append(scNames, scName)
+	assert.ElementsMatch(t, scNames, dbCtx.RequireAttachmentMigration)
+	db.RequireBackgroundManagerState(t, ctx, mgr, db.BackgroundProcessStateRunning)
```
torcolvin, line 139: Is there a chance this could run fast enough that we miss it? That's probably the reason you are using 4000 docs, which I would allude to in a comment.

rest/attachment_migration_test.go
```diff
@@ -0,0 +1,455 @@
...
+	resp = rt.UpsertDbConfig(dbName, dbConfig)
+	RequireStatus(t, resp, http.StatusCreated)
+
+	// wait for attachment migration job to start and finish
+	dbCtx = rt.GetDatabase()
+	mgr = dbCtx.AttachmentMigrationManager
+	assert.Len(t, dbCtx.RequireAttachmentMigration, 2)
+	scName = base.ScopeAndCollectionName{Scope: scope, Collection: collection2}
+	scNames = append(scNames, scName)
+	assert.ElementsMatch(t, scNames, dbCtx.RequireAttachmentMigration)
```
torcolvin:
```suggestion
	scNames = append(scNames, base.ScopeAndCollectionName{Scope: scope, Collection: collection2})
	assert.ElementsMatch(t, scNames, dbCtx.RequireAttachmentMigration)
```

Then we will see the expected and difference right away.

torcolvin (commented):
I didn't finish looking through the tests, but I did the first part of them. I'll let you go through the comments and apply them to other parts of the tests.

I'd think about if there are ways to make the tests more readable (shorter), maybe helper functions to get status struct and be able to compare it?

db/database.go
gregns1: Do you still think this should be done? Given we don't block for going online this migration (like resync) my feeling is we should still take db online but with sufficient warnings for it in the logs.

torcolvin: This was my misunderstanding about the requirements of how attachment migration works.

I think we should log that creating a database is starting a background task to perform attachment migration, but this can be as is.

base/constants_syncdocs.go
```diff
@@ -385,32 +385,45 @@ type SyncInfo struct {
...
-func InitSyncInfo(ds DataStore, metadataID string) (requiresResync bool, err error) {
+//     If syncInfo exists and has metaDataVersion greater than or equal to 4.0, return requiresAttachmentMigration=false, else requiresAttachmentMigration=true to bring migrate metadata attachments.
+func InitSyncInfo(ds DataStore, metadataID string) (requiresResync bool, requiresAttachmentMigration bool, err error) {
 
 	var syncInfo SyncInfo
 	_, fetchErr := ds.Get(SGSyncInfo, &syncInfo)
 	if IsDocNotFoundError(fetchErr) {
 		if metadataID == "" {
-			return false, nil
+			return false, true, nil
```
torcolvin, line 396: I was running `go test -coverpkg ./... -cover=cover.out ./db -run TestSyncInfo` to see if lines were covered and I was surprised this line isn't hit.

gregns1: Not sure what test TestSyncInfo is referring to? Doesn't seem to be a test I've written, with regards to that particular line being covered in tests the test `TestInitSyncInfoRequireMigrationEmptyBucket` covers this line

base/constants_syncdocs.go
```diff
@@ -385,32 +385,45 @@ type SyncInfo struct {
...
-				return true, fmt.Errorf("Error retrieving syncInfo (after failed add): %v", fetchErr)
+				return true, true, fmt.Errorf("Error retrieving syncInfo (after failed add): %v", fetchErr)
 			}
 		} else if addErr != nil {
-			return true, fmt.Errorf("Error adding syncInfo: %v", addErr)
+			return true, true, fmt.Errorf("Error adding syncInfo: %v", addErr)
 		}
 		// successfully added
-		return false, nil
+		return false, true, nil
```
torcolvin: If there was a fetch in the middle to get a new value (401), you need to check the metadata version again and run a version check.

base/constants_syncdocs.go
```diff
@@ -457,10 +467,45 @@ func SetSyncInfoMetaVersion(ds DataStore, metaVersion string) error {
...
+
+// CheckRequireAttachmentMigration will return true if current metaVersion < 4.0.0, else false
+func CheckRequireAttachmentMigration(version *ComparableBuildVersion) (bool, error) {
+	if version == nil {
+		return true, nil
+	}
+	minVerStr := "4.0.0" // minimum meta version that needs to be defined for metadata migration. Any version less than this will require attachment migration
+	minVersion, err := NewComparableBuildVersionFromString(minVerStr)
+	if err != nil {
+		return true, nil
```
torcolvin: We aren't passing the error here, but I think we should at least log this case with a warning and fail this case with cb_sg_devmode.

base/constants_syncdocs.go
```diff
@@ -457,10 +467,45 @@ func SetSyncInfoMetaVersion(ds DataStore, metaVersion string) error {
...
+	if err != nil {
+		return true, err
+	}
+	return requiresAttachmentMigration, nil
+}
+
+// CheckRequireAttachmentMigration will return true if current metaVersion < 4.0.0, else false
+func CheckRequireAttachmentMigration(version *ComparableBuildVersion) (bool, error) {
+	if version == nil {
+		return true, nil
```
torcolvin: Is this a condition that needs a warning log and a failure?

torcolvin (changes requested):

base/constants_syncdocs.go
```diff
@@ -457,10 +468,51 @@ func SetSyncInfoMetaVersion(ds DataStore, metaVersion string) error {
...
+		requiresAttachmentMigration, err = CheckRequireAttachmentMigration(ctx, syncInfoVersion)
+		if err != nil {
+			return true, err
+		}
+	} else {
+		// no meta version passed in, thus attachment migration should take place
+		requiresAttachmentMigration = true
+	}
+	return requiresAttachmentMigration, nil
+}
```
torcolvin: nit: this is probably easier to read like this:

```suggestion
    // no meta version passed in, thus attachment migration should take place
    if metaVersion == "" {
    	return requiresAttachmentMigration, nil
    }
	syncInfoVersion, err := NewComparableBuildVersionFromString(metaVersion)
	if err != nil {
		return true, err
	}
	return CheckRequireAttachmentMigration(ctx, syncInfoVersion)
}
```

torcolvin (approved):

PR #7158, opened 2024-10-15
CBG-4254 implement Couchbase Server peer

torcolvin:
- Skipped tests in https://docs.google.com/spreadsheets/d/1U2FxFlMo5vT-tk9mmDgw68zONX1JCPsy9gxzH7uWk6E/edit?gid=0#gid=0 in future tickets.
- Implement Write and Delete functions to always overwrite with the latest update. The Delete will technically be able to delete something that has already been deleted, but I actually think that is desired.
- Read the hlv implicitly out the xattrs to construct a DocVersion, there is a concept of a CV without a _vv, that is just implicitly sourceID@cas (when a document is written directly to couchbase server and not replicated by XDCR).

### DIFF
--- a/topologytest/couchbase_server_peer_test.go
+++ b/topologytest/couchbase_server_peer_test.go
@@ -10,6 +10,7 @@ package topologytest
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"testing"
 	"time"
@@ -66,7 +67,8 @@ func (p *CouchbaseServerPeer) String() string {
 	return p.name
 }
 
-func (p *CouchbaseServerPeer) ctx() context.Context {
+// Context returns the context for the peer.
+func (p *CouchbaseServerPeer) Context() context.Context {
 	return base.TestCtx(p.tb)
 }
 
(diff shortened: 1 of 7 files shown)


topologytest/couchbase_server_peer_test.go
```diff
@@ -174,11 +206,83 @@ func (p *CouchbaseServerPeer) CreateReplication(passivePeer Peer, config PeerRep
...
+	docVersion := rest.DocVersion{}
+	hlvBytes, ok := xattrs[base.VvXattrName]
+	if ok {
+		var hlv *db.HybridLogicalVector
+		require.NoError(peer.TB(), json.Unmarshal(hlvBytes, &hlv))
+		docVersion.CV = db.Version{SourceID: hlv.SourceID, Value: hlv.Version}
+	} else {
+		docVersion.CV = db.Version{SourceID: peer.SourceID(), Value: cas}
+	}
+	sync, ok := xattrs[base.SyncXattrName]
```
adamcfraser, line 234: RevTreeID will not (necessarily) be consistent across XDCR replications, so from that perspective I might have expected a CouchbaseServerPeer to ignore SyncXattr/RevTreeID as those are concepts only relevant when talking about SGW and CBL peers.  Are there some test cases that require this information to be populated?

adamcfraser (commented):
Overall looks fine, one question about populating revTreeID for couchbase peers.

topologytest/couchbase_server_peer_test.go
torcolvin: I think I was populating it so it would be present, but not necessarily using it for anything at this time. You could imagine that CBL would set this value, and then Couchbase Server would read it. It is helpful for debugging to know what revid it was, because sometimes I can correlate with where that came from, in addition to Source ID.

I think that there could/will be a need to add `PV` to this version of certain types of assertions, not yet implemented.

adamcfraser (approved):

PR #7161, opened 2024-10-17
CBG-2838: Rely on index 'IF NOT EXISTS' and idempotent deferred index builds fo…

bbrks:
CBG-2838

Rely on index `IF NOT EXISTS` and idempotent deferred index builds for simpler and lower-overhead index creation.

According to the query docs, this is perfectly fine to do, and eliminates a lot of our manual `system:indexes` queries and retry/error handling logic?

Now the only `system:indexes` call we have to make is the final one for each collection, to make sure all of the indexes are finished building and are online.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2756/
- [x] `CBS 7.0.5` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2754/


db/indexes.go
```diff
@@ -384,22 +348,19 @@ func InitializeIndexes(ctx context.Context, n1QLStore base.N1QLStore, options In
...
-		isDeferred, err := sgIndex.createIfNeeded(ctx, n1QLStore, options)
+		err := sgIndex.createIfNeeded(ctx, n1QLStore, options)
 		if err != nil {
 			return base.RedactErrorf("Unable to install index %s: %v", base.MD(sgIndex.simpleName), err)
 		}
 
-		if isDeferred {
-			deferredIndexes = append(deferredIndexes, fullIndexName)
-		}
+		deferredIndexes = append(deferredIndexes, fullIndexName)
```
torcolvin: These are no longer necessarily deferred indexes, they could actually be already created.  This would just build all indexes and it is idempotent if they are present.

I think just a change for the variable name is all that is necessary here.

I think with this then the `if len(deferredIndexes) > 0` is not necessary but not harmful. I would probably use `fullIndexNames`

torcolvin (commented):
As discussed offline, some improvements to this PR should be made:

- look for `IsIndexerRetryIndexError` usage as that is where we are taking action on a retryable error. Potentially with `CREATE INDEX .. IF NOT EXISTS` this error type will not be returned.
- test this on CBS <7.1 which does not have `IF NOT EXISTS`

db/indexes.go
bbrks: done this, and added a typed error `ErrIndexBackgroundRetry` to enable those to be swallowed conditionally (based on IF NOT EXISTS) further up the call stack

base/collection_n1ql_common.go
```diff
@@ -128,7 +132,32 @@ func (im *indexManager) GetAllIndexes() ([]gocb.QueryIndex, error) {
 //	  CreateIndex("myIndex", "field1, field2, nested.field", "field1 > 0", N1qlIndexOptions{numReplica:1})
 //	CREATE INDEX myIndex on myBucket(field1, field2, nested.field) WHERE field1 > 0 WITH {"numReplica":1}
 func CreateIndex(ctx context.Context, store N1QLStore, indexName string, expression string, filterExpression string, options *N1qlIndexOptions) error {
-	createStatement := fmt.Sprintf("CREATE INDEX `%s` ON %s(%s)", indexName, store.EscapedKeyspace(), expression)
+	return createIndex(ctx, store, indexName, expression, filterExpression, false, options)
+}
+
+// CreateIndexIfNotExists issues a CREATE INDEX query in the current bucket, using the form:
```
torcolvin: This is more accurately create an index in the current collection, I think?

base/collection_n1ql_common.go
```diff
@@ -171,14 +204,12 @@ func createIndex(ctx context.Context, store N1QLStore, indexName string, createS
 
 	if IsIndexerRetryIndexError(err) {
```
torcolvin: Could you combine these two if statements to something like `if IsIndexerRetryIndexError(err) || IsCreateDuplicateIndexError(err)`?

I don't think that either of these are necessarily need to be flagged as Info, especially the "duplicate index creation", since I think this is likely to happen in the case that multiple nodes are running through the index creation code simultaneously, in the case of a `PUT /db` with `start_offline: true` on a multi node cluster. The first node will start the initialization and then call `BootstrapConfig.UpdateConfig` for the database, and the subsequent nodes will pick up that config with bootstrap polling. However, with a sufficiently size index creation, the idea that all nodes will exist at the same time is more likely.

Without `start_offline: true`, this would be less likely to happen since the index creation is synchronous to the HTTP rest call.

I think that dropping these statements to debug, or lessening the `Error:` / `error` warning is appropriate. Maybe something like:

```
Index %s already being created on server. N1QL output: %s", indexName, err
```

base/collection_n1ql_common.go
```diff
@@ -213,51 +244,31 @@ func waitForIndexExistence(ctx context.Context, store N1QLStore, indexName strin
...
-	}
 
-	if len(deferredIndexes) == 0 {
-		return nil
-	}
-
-	InfofCtx(ctx, KeyQuery, "Building deferred indexes: %v", deferredIndexes)
-	buildErr := buildIndexes(ctx, s, deferredIndexes)
-	return buildErr
+	InfofCtx(ctx, KeyQuery, "Built deferred indexes: %v", indexSet)
```
torcolvin: The old message should be right:

```
InfofCtx(ctx, KeyQuery, "Building deferred indexes: %v", deferredIndexes)
```

This only issues a build command, but doesn't actually mean the indexes are built yet.

db/indexes.go
```diff
@@ -330,30 +300,29 @@ func (i *SGIndex) createIfNeeded(ctx context.Context, bucket base.N1QLStore, opt
...
+		err = bucket.CreateIndexIfNotExists(ctx, indexName, indexExpression, filterExpression, n1qlOptions)
 		if err != nil {
-			// If index has already been created (race w/ other SG node), return without error
-			if err == base.ErrAlreadyExists {
-				isDeferred = false // Index already exists, don't need to update.
-				return false, nil, nil
-			}
 			if strings.Contains(err.Error(), "not enough indexer nodes") {
 				return false, fmt.Errorf("Unable to create indexes with the specified number of replicas (%d).  Increase the number of index nodes, or modify 'num_index_replicas' in your Sync Gateway database config.", options.NumReplicas), nil
+			} else if errors.Is(err, base.ErrIndexBackgroundRetry) {
```
torcolvin: Maybe drop a comment that we only think this will happen with <7.1 ?

db/indexes.go
```diff
@@ -330,30 +300,29 @@ func (i *SGIndex) createIfNeeded(ctx context.Context, bucket base.N1QLStore, opt
...
+				// bubble up without retry so the caller can handle waiting for index readiness
+				base.DebugfCtx(ctx, base.KeyQuery, "Index creation failed with background retry error - will wait for index readiness: %v", err)
```
torcolvin: We should probably log the name of the index here, and maybe change the comment, to indicate that this was a failure returned by N1QL but that the index is expected to be created on the server.

db/indexes_test.go
```diff
@@ -104,7 +105,99 @@ func TestInitializeIndexes(t *testing.T) {
...
+
+			n1qlStore, ok := base.AsN1QLStore(collection.dataStore)
+			require.True(t, ok)
+
+			if test.clusterN1QLStore {
+				n1qlStore, err = base.NewClusterOnlyN1QLStore(gocbBucket.GetCluster(), gocbBucket.BucketName(), collection.ScopeName, collection.Name)
+				require.NoError(t, err)
+			}
+
+			// drop and restore indexes between tests, to the way the bucket pool would
```
torcolvin: I'd be very tempted to move this to `db/indextest` because we know that dropping and readding indexes takes a lot of time and also causes a lot of load on the server.

There's a decent amount of reusable test code there as well, this could probably just call https://github.com/couchbase/sync_gateway/blob/main/db/indextest/indextest_test.go#L248 in a loop with different ctxs

base/collection_n1ql_common.go
bbrks: Removed, since we already had the "Building" line above.

base/collection_n1ql_common.go
bbrks: Yep, old comment stolen from `CreateIndex`. Have removed all bucket references.

db/indexes.go
bbrks: This error can happen on all versions. It's not related to the `IF NOT EXISTS` change, and has been added to change from per-index wait to collection-wide `waitForIndexesOnline`

base/collection_n1ql_common.go
```diff
@@ -123,7 +123,7 @@ func (im *indexManager) GetAllIndexes() ([]gocb.QueryIndex, error) {
 	return im.cluster.GetAllIndexes(im.bucketName, opts)
 }
 
-// CreateIndex issues a CREATE INDEX query in the current bucket, using the form:
```
torcolvin, line 122: Do you want to change example also? This might be pedantic since this will be bucket.scope.collection

base/collection_n1ql_common.go
```diff
@@ -149,23 +149,23 @@ func CreateIndexIfNotExists(ctx context.Context, store N1QLStore, indexName stri
 
 // createIndex is a common function for CreateIndex and CreateIndexIfNotExists
 func createIndex(ctx context.Context, store N1QLStore, indexName string, expression string, filterExpression string, ifNotExists bool, options *N1qlIndexOptions) error {
-	createClause := "CREATE INDEX"
-
+	var ifNotExistsStr string
 	// Server 7.1+ - we can still safely _not_ use this when it's not available, because we have equivalent error handling inside this function to swallow `ErrAlreadyExists`.
```
torcolvin, line 153: this now needs to be 7.6+

bbrks: Nope, still 7.1+

base/collection_n1ql_common.go
bbrks: It's still a valid example if using the default collection and it doesn't seem necessary to change.

db/indexes_test.go
bbrks: Done, had to add some additional stuff to `setupN1QLStore` to make it work, and it's not really compatible with the other associated test, given the permutations intentionally ignoring the test env vars for better coverage.

torcolvin (approved):

PR #7162, opened 2024-10-18
CBG-4300 improve rosmar XDCR handling

torcolvin:
- copy HLV if set on from collection, and construct HLV if needed
- check _mou.pCAS when deciding whether to XDCR document

### DIFF
--- a/xdcr/rosmar_xdcr.go
+++ b/xdcr/rosmar_xdcr.go
@@ -79,7 +79,7 @@ func (r *rosmarManager) processEvent(ctx context.Context, event sgbucket.FeedEve
 		}
 
 		// Have to use GetWithXattrs to get a cas value back if there are no xattrs (GetWithXattrs will not return a cas if there are no xattrs)
-		_, toXattrs, toCas, err := col.GetWithXattrs(ctx, docID, []string{base.VvXattrName, base.MouXattrName})
+		_, targetXattrs, toCas, err := col.GetWithXattrs(ctx, docID, []string{base.VvXattrName, base.MouXattrName})
 		if err != nil && !base.IsDocNotFoundError(err) {
 			base.WarnfCtx(ctx, "Skipping replicating doc %s, could not perform a kv op get doc in toBucket: %s", event.Key, err)
 			r.errorCount.Add(1)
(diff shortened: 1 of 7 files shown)


xdcr/rosmar_xdcr.go
```diff
@@ -123,7 +123,46 @@ func (r *rosmarManager) processEvent(ctx context.Context, event sgbucket.FeedEve
 			return true
 		}
 
-		err = opWithMeta(ctx, col, r.fromBucketSourceID, toCas, toXattrs, event)
+		hlv, mou, body, err := getBodyHLVAndMou(event)
```
adamcfraser: When mou.cas == cas on the target (i.e. there was an import), I believe we should be using toCas = hlv.cvCAS to decide whether to replicate the source mutation.  So I think we require something above line 120 like:
```
 if toXattrs.mou.cas == toCas {
   toCas = toXattrs.hlv.cvCAS
}
```

xdcr/rosmar_xdcr.go
```diff
@@ -123,7 +123,46 @@ func (r *rosmarManager) processEvent(ctx context.Context, event sgbucket.FeedEve
...
+		if err != nil {
+			base.WarnfCtx(ctx, "Replicating doc %s, could not get body, hlv, and mou: %s", event.Key, err)
+			r.errorCount.Add(1)
+			return false
+		}
+		if hlv != nil && mou != nil {
+			pCAS := base.HexCasToUint64(mou.PreviousCAS)
+			if pCAS <= toCas {
```
adamcfraser: We shouldn't be considering pCAS when deciding whether to replicate or not - I believe only mou.cas, cvCAS and cas should be considered.

xdcr/rosmar_xdcr.go
```diff
@@ -123,7 +123,46 @@ func (r *rosmarManager) processEvent(ctx context.Context, event sgbucket.FeedEve
...
+				base.TracefCtx(ctx, base.KeyWalrus, "Skipping replicating doc %s, cas %d <= %d", docID, event.Cas, toCas)
+				return true
+			}
+		}
+		if hlv == nil {
+			newHlv := db.NewHybridLogicalVector()
+			hlv = &newHlv
+			err := hlv.AddVersion(db.Version{
+				SourceID: r.fromBucketSourceID,
+				Value:    event.Cas,
```
adamcfraser: Here's where we want to use pCAS.  If the source mutation (event) was an import (event.Cas == mou.cas), we should set the value to event.mou.pCas.

xdcr/rosmar_xdcr.go
```diff
@@ -123,7 +123,46 @@ func (r *rosmarManager) processEvent(ctx context.Context, event sgbucket.FeedEve
...
+			})
+			if err != nil {
+				base.WarnfCtx(ctx, "Replicating doc %s, could not set hlv.AddVersion: %s", event.Key, err)
+				r.errorCount.Add(1)
+				return false
+			}
+			hlv.CurrentVersionCAS = event.Cas
```
adamcfraser: Similar to the comment above, if the source mutation was an import, should set cvCAS to pCAS.

xdcr/rosmar_xdcr.go
adamcfraser: I'm wondering if naming everything 'sourceHLV, sourceMou, sourceBody' and 'targetHLV, targetMou etc' will make it easier to follow the logic here.

adamcfraser (changes requested):

xdcr/rosmar_xdcr.go
torcolvin: This was a bug, fixed.

xdcr/rosmar_xdcr.go
torcolvin: Since XDCR doesn't handle this case, I'm not going to handle this case for rosmar XDCR.

xdcr/rosmar_xdcr.go
torcolvin: Since XDCR doesn't handle this case, I'm not going to handle this case for rosmar XDCR.

torcolvin: This was evaluating the `_mou` for the case where HLV is not created by import, which will not happen for Sync Gateway 4.x, this could only exist from Sync Gateway 3.2+ clients.

In this case, the import would be echoed between clusters, but only once.

adamcfraser (dismissed):

adamcfraser (approved):

PR #7171, opened 2024-10-22
CBG-4263 preserve _sync xattr on the target

torcolvin:
Make sure that target xattrs are overwritten, except for `_sync` which is preserved.

### DIFF
--- a/xdcr/rosmar_xdcr.go
+++ b/xdcr/rosmar_xdcr.go
@@ -79,7 +79,7 @@ func (r *rosmarManager) processEvent(ctx context.Context, event sgbucket.FeedEve
 		}
 
 		// Have to use GetWithXattrs to get a cas value back if there are no xattrs (GetWithXattrs will not return a cas if there are no xattrs)
-		_, targetXattrs, toCas, err := col.GetWithXattrs(ctx, docID, []string{base.VvXattrName, base.MouXattrName})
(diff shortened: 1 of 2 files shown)


xdcr/rosmar_xdcr.go
```diff
@@ -135,33 +135,31 @@ func (r *rosmarManager) processEvent(ctx context.Context, event sgbucket.FeedEve
...
 		}
-		if targetXattrs == nil {
-			targetXattrs = make(map[string][]byte, 1) //  length for _vv
+		var newXattrs map[string][]byte
+		if sourceXattrs == nil {
+			newXattrs = make(map[string][]byte)
+		} else {
+			newXattrs = sourceXattrs
+			for _, xattrName := range []string{base.VvXattrName, base.MouXattrName, base.SyncXattrName} {
+				delete(newXattrs, xattrName)
```
adamcfraser: It looks like we're already deleting _vv and _mou in getBodyHLVAndMou.  Do you think it would make the implementation more straightforward clear if the sourceXattrs returned by getBodyHLVAndMou were called 'non-mobile xattrs', and excluded _vv, _mou and _sync?

adamcfraser (changes requested):

adamcfraser (approved):

PR #7174, opened 2024-10-30
CBG-4323: fix for per shard memory based eviction

gregns1:
CBG-4323

- Hit a panic in 3.2.1 testing that was due to using the overall memory stat for memory based eviction when using sharded rev cache, this panic was from trying to evict from empty rev cache shard. As I was using the overall stat for rev cache memory usage when I hit a point where memory based eviction was needed (as overall memory exceeded the parameter), per chance I hit this point when trying to add a new doc that was going to empty shard. When adding this doc eviction as hit and then we tried to evict from the current shard but to was empty. 
- Fix is to keep separate count for memory usage per shard and evict per shard aligning with what we do for num items eviction 
- Had to do this through another atomic as we don't always hold rev cache lock when incrementing/decrementing memory and reacquiring lock will probably be more expensive
- Have minimum of 50MB for cache config 
- Removed 10% buffer per shard 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2764/

### DIFF
--- a/db/revision_cache_lru.go
+++ b/db/revision_cache_lru.go
@@ -82,16 +82,17 @@ func (sc *ShardedLRURevisionCache) Remove(docID, revID string, collectionID uint
 
 // An LRU cache of document revision bodies, together with their channel access.
 type LRURevisionCache struct {
-	backingStores    map[uint32]RevisionCacheBackingStore
-	cache            map[IDAndRev]*list.Element
-	lruList          *list.List
-	cacheHits        *base.SgwIntStat
-	cacheMisses      *base.SgwIntStat
-	cacheNumItems    *base.SgwIntStat
-	lock             sync.Mutex
-	capacity         uint32
-	memoryCapacity   int64
-	cacheMemoryBytes *base.SgwIntStat
+	backingStores        map[uint32]RevisionCacheBackingStore
+	cache                map[IDAndRev]*list.Element
+	lruList              *list.List
+	cacheHits            *base.SgwIntStat
+	cacheMisses          *base.SgwIntStat
+	cacheNumItems        *base.SgwIntStat
+	lock                 sync.Mutex
+	capacity             uint32
+	MaxMemoryCapacity    int64
+	currMemoryCapacity   base.AtomicInt
+	cacheMemoryBytesStat *base.SgwIntStat
 }
 
 // The cache payload data. Stored as the Value of a list Element.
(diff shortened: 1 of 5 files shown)


db/revision_cache_lru.go
```diff
@@ -82,16 +82,17 @@ func (sc *ShardedLRURevisionCache) Remove(docID, revID string, collectionID uint
...
-	cacheMemoryBytes *base.SgwIntStat
+	backingStores        map[uint32]RevisionCacheBackingStore
+	cache                map[IDAndRev]*list.Element
+	lruList              *list.List
+	cacheHits            *base.SgwIntStat
+	cacheMisses          *base.SgwIntStat
+	cacheNumItems        *base.SgwIntStat
+	lock                 sync.Mutex
+	capacity             uint32           // Max number of items capacity of LRURevisionCache
+	MaxMemoryCapacity    int64            // Max memory capacity of LRURevisionCache
```
adamcfraser: It doesn't look like this needs to be a public property (can be `maxMemoryCapacity`).  Actually - I'm not clear why this needed to change from `memoryCapacity`.

db/revision_cache_lru.go
```diff
@@ -82,16 +82,17 @@ func (sc *ShardedLRURevisionCache) Remove(docID, revID string, collectionID uint
...
+	currMemoryCapacity   base.AtomicInt   // count of number of bytes used currently in the LRURevisionCache
+	cacheMemoryBytesStat *base.SgwIntStat // stat for overall cache memory usage in bytes
```
```suggestion
	cacheMemoryBytesStat *base.SgwIntStat // stat for overall revision cache memory usage in bytes.  When using sharded cache, will be shared by all shards.
```

db/revision_cache_lru.go
```diff
@@ -82,16 +82,17 @@ func (sc *ShardedLRURevisionCache) Remove(docID, revID string, collectionID uint
...
+	backingStores        map[uint32]RevisionCacheBackingStore
+	cache                map[IDAndRev]*list.Element
+	lruList              *list.List
+	cacheHits            *base.SgwIntStat
+	cacheMisses          *base.SgwIntStat
+	cacheNumItems        *base.SgwIntStat
+	lock                 sync.Mutex
+	capacity             uint32           // Max number of items capacity of LRURevisionCache
+	MaxMemoryCapacity    int64            // Max memory capacity of LRURevisionCache
+	currMemoryCapacity   base.AtomicInt   // count of number of bytes used currently in the LRURevisionCache
```
```suggestion
	currMemoryUsage   base.AtomicInt   // count of number of bytes used currently in the LRURevisionCache
```

db/revision_cache_lru.go
```diff
@@ -541,9 +542,19 @@ func (rc *LRURevisionCache) performEviction() {
...
 }
+
+// updateRevCacheMemoryUsage atomically increases overall memory usage for cache and the actual rev cache objects usage
+func (rc *LRURevisionCache) updateRevCacheMemoryUsage(bytesCount int64) {
+	// We need to keep track of the current LRURevisionCache memory usage AND the overall usage of the cache. We need
+	// overall memory usage for the stat added to show rev cache usage plus we need the current rev cache capacity of the
+	// LRURevisionCache object for sharding the rev cache. This way we can perform eviction on per shard basis much like
+	// we do with the number of items capacity eviction
+	rc.currMemoryCapacity.Add(bytesCount)
+	rc.cacheMemoryBytesStat.Add(bytesCount)
```
adamcfraser, line 559: In other scenarios like this (for the channel cache in particular), we compute aggregate stats at stat collection time (see UpdateCalculatedStats).  The above approach may be fine for 3.2.1 (based on performance results), but filing a tracking ticket to use this approach going forward would be a good idea.

adamcfraser (dismissed):
Generally looks good with some naming/comment suggestions for clarity and a suggestion for a future enhancement ticket.

db/revision_cache_lru.go
gregns1: This was the approach I wanted to take, but it came with its complexities with fetching the current memory usage. Would require a new interface method on the rev cache which I thought at this time may not be wanted (given the time constraints). I can file a ticket to explore this in a future release.

db/revision_cache_lru.go
gregns1: Making public was mistake, the renaming was just to make it more clear in terms of distinction from `currMemoryCapacity` but after the rename of that to `currMemoryUsage` I can revert it.

bbrks (approved):

PR #7178, opened 2024-11-04
Cancel in-progress GH actions jobs when the PR or ref has been superseded

bbrks:
We don't need the old runs to continue to completion for PRs or refs that are being re-run.
Cancel them to save unnessesary CI work. Thanks @snej!

https://turso.tech/blog/simple-trick-to-save-environment-and-money-when-using-github-actions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,10 @@ on:
       - 'release/*'
       - 'beryllium'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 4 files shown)


torcolvin (commented):
This does the right thing for cancelling PRs associated with a pull request, but we don't want this to cancel requests if they are push and main,release* branches.

I think we want to add something like https://docs.github.com/en/enterprise-cloud@latest/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#example-only-cancel-in-progress-jobs-on-specific-branches

torcolvin (approved):

PR #7179, opened 2024-11-05
CBG-4335 update to go 1.23, and minor versions for 3.1/3.2 branches

torcolvin:
- go2xunit fails with go 1.23 on a full integration test run, so use 1.22. I tried briefly to use https://github.com/jstemmer/go-junit-report but it doesn't work as a drop in replacement. I'm not sure what the long term plan for creating xunit files would be, but I think this solution will continue to work for some time.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2774/ (go2xunit changes only)
- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2768/ (fails with only go2xunit)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: 1.23.2
+          go-version: 1.23.3
       - name: go-build
         run: go build "./..."
 
@@ -43,7 +43,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: 1.23.2
+          go-version: 1.23.3
       - run: go install github.com/google/addlicense@latest
       - run: addlicense -check -f licenses/addlicense.tmpl .
 
@@ -54,7 +54,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: 1.23.2
+          go-version: 1.23.3
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v4
         with:
@@ -81,7 +81,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: 1.23.2
+          go-version: 1.23.3
       - name: Build
         run: go build -v "./..."
       - name: Run Tests
@@ -101,7 +101,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: 1.23.2
+          go-version: 1.23.3
       - name: Run Tests
         run: go test -tags cb_sg_devmode -race -shuffle=on -timeout=30m -count=1 -json -v "./..." | tee test.json | jq -s -jr 'sort_by(.Package,.Time) | .[].Output | select (. != null )'
         shell: bash
@@ -120,7 +120,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: 1.23.2
+          go-version: 1.23.3
       - name: Run Tests
         run: go test -tags cb_sg_devmode -shuffle=on -timeout=30m -count=1 -json -v "./..." | tee test.json | jq -s -jr 'sort_by(.Package,.Time) | .[].Output | select (. != null )'
         shell: bash
@@ -139,7 +139,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: 1.23.2
+          go-version: 1.23.3
       - name: Run Tests
         run: go test -tags cb_sg_devmode -shuffle=on -timeout=30m -count=1 -json -v "./..." | tee test.json | jq -s -jr 'sort_by(.Package,.Time) | .[].Output | select (. != null )'
         shell: bash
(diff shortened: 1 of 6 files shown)


jenkins-integration-build.sh
```diff
@@ -57,7 +57,10 @@ if [ "${USE_GO_MODULES:-}" == "false" ]; then
     go get -u -v github.com/AlekSi/gocov-xml
 else
     # Install tools to use after job has completed
-    go install -v github.com/tebeka/go2xunit@latest
+    # go2xunit will fail with 1.23 with name mismatch (try disabling parallel mode), but without any t.Parallel()
+    go install golang.org/dl/go1.22.8@latest
+    ~/go/bin/go1.22.8 download
+    ~/go/bin/go1.22.8 install -v github.com/tebeka/go2xunit@latest
```
bbrks, line 63: As discussed this isn't going to fix the issue (it seems like this issue is sporadic and not related to Go version)

```suggestion
    go install -v github.com/tebeka/go2xunit@latest
```

.github/workflows/ci.yml
```diff
@@ -54,12 +54,11 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: 1.22.5
-          cache: false
+          go-version: 1.23.2
```
bbrks: What's the reason for dropping `cache: false`?

torcolvin: I was looking at the latest version of packages and this isn't the recommendation. I remember writing this code but I'm not sure why. It seems like it would be better to cache go version when available.

I'm happy to leave this code in though.

jenkins-integration-build.sh
torcolvin: I tested this on a bunch of files from Jenkins locally, and there are way more cases where this will fail with go 1.23 and not go 1.22

I do not think this is a robust solution and I think switching to go-junit-report with either a patch or post processing the xml will be the right solution. I'll file a ticket for this since we now saw this elsewhere. I do think that some of the problems came from parsing the panics in https://github.com/couchbase/sync_gateway/commit/142f2156bedfaec8e824d9724ce875f5de20818c.

.github/workflows/ci.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: 1.22.5
+          go-version: 1.23.2
```
bbrks:
```suggestion
          go-version: 1.23.3
```

.github/workflows/ci.yml
```diff
@@ -43,7 +43,7 @@ jobs:
...
```
bbrks:
```suggestion
          go-version: 1.23.3
```

.github/workflows/ci.yml
bbrks:
```suggestion
          go-version: 1.23.3
```

.github/workflows/ci.yml
```diff
@@ -82,15 +81,15 @@ jobs:
...
```
bbrks:
```suggestion
          go-version: 1.23.3
```

.github/workflows/ci.yml
```diff
@@ -102,13 +101,13 @@ jobs:
...
```
bbrks:
```suggestion
          go-version: 1.23.3
```

.github/workflows/ci.yml
```diff
@@ -121,13 +120,13 @@ jobs:
...
```
bbrks:
```suggestion
          go-version: 1.23.3
```

.github/workflows/ci.yml
```diff
@@ -140,13 +139,13 @@ jobs:
...
```
bbrks:
```suggestion
          go-version: 1.23.3
```

.github/workflows/ci.yml
```diff
@@ -192,7 +191,7 @@ jobs:
...
```
bbrks:
```suggestion
          go-version: 1.23.3
```

.github/workflows/service.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       # build sync gateway since the executable is needed for service installation
       - uses: actions/setup-go@v5
         with:
-          go-version: 1.22.5
+          go-version: 1.23.2
```
bbrks:
```suggestion
          go-version: 1.23.3
```

Jenkinsfile
```diff
@@ -16,7 +16,7 @@ pipeline {
     }
 
     tools {
-        go '1.22.5'
+        go '1.23.2'
```
bbrks:
```suggestion
        go '1.23.3'
```

manifest/product-config.json
```diff
@@ -6,7 +6,7 @@
             "release_name": "Couchbase Sync Gateway",
             "production": true,
             "interval": 30,
-            "go_version": "1.22.5",
+            "go_version": "1.23.2",
```
bbrks:
```suggestion
            "go_version": "1.23.3",
```

manifest/product-config.json
```diff
@@ -621,7 +621,7 @@
             "release_name": "Couchbase Sync Gateway 4.0.0",
             "production": true,
             "interval": 30,
-            "go_version": "1.22.2",
+            "go_version": "1.23.2",
```
bbrks:
```suggestion
            "go_version": "1.23.3",
```

bbrks (approved):

PR #7183, opened 2024-11-06
CBG-4213: add attachment migration api

gregns1:
CBG-4213

- Addition of the api end point to trigger attachment migration and to get status of it
- Move tests into separate package in rest 
- Api docs for this new endpoint

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2778/

### DIFF
--- a/db/background_mgr_attachment_migration.go
+++ b/db/background_mgr_attachment_migration.go
@@ -66,8 +66,8 @@ func (a *AttachmentMigrationManager) Init(ctx context.Context, options map[strin
 		var statusDoc AttachmentMigrationManagerStatusDoc
 		err := base.JSONUnmarshal(clusterStatus, &statusDoc)
 
-		reset, ok := options["reset"].(bool)
-		if reset && ok {
+		reset, _ := options["reset"].(bool)
+		if reset {
 			base.InfofCtx(ctx, base.KeyAll, "Attachment Migration: Resetting migration process. Will not resume any partially completed process")
 		}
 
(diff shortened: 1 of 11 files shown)


db/background_mgr_attachment_migration.go
```diff
@@ -66,9 +66,14 @@ func (a *AttachmentMigrationManager) Init(ctx context.Context, options map[strin
 		var statusDoc AttachmentMigrationManagerStatusDoc
 		err := base.JSONUnmarshal(clusterStatus, &statusDoc)
 
+		reset, ok := options["reset"].(bool)
+		if reset && ok {
```
bbrks: Nit: We could rely on the `false` zero value to avoid an extra bool being passed around for evaluation like this? Would slightly clean up the code below and has the same behaviour - any non-bool passed would evaluate to `false`.

```suggestion
		reset, _ := options["reset"].(bool)
		if reset {
```

db/background_mgr_attachment_migration.go
```diff
@@ -66,9 +66,14 @@ func (a *AttachmentMigrationManager) Init(ctx context.Context, options map[strin
...
+			base.InfofCtx(ctx, base.KeyAll, "Attachment Migration: Resetting migration process. Will not resume any partially completed process")
+		}
+
 		// If the previous run completed, or there was an error during unmarshalling the status we will start the
 		// process from scratch with a new migration ID. Otherwise, we should resume with the migration ID, stats specified in the doc.
-		if statusDoc.State == BackgroundProcessStateCompleted || err != nil {
+		if statusDoc.State == BackgroundProcessStateCompleted || err != nil || (reset && ok) {
```
bbrks:
```suggestion
		if statusDoc.State == BackgroundProcessStateCompleted || err != nil || reset {
```

bbrks (dismissed):
Looks good, one nitpick but feel free to merge as-is

bbrks (approved):

PR #7186, opened 2024-11-07
CBG-4348 clarify logs around changes feed

torcolvin:
CBG-4348 clarify logs around changes feed

- If admin user, log that is the admin user.
- If admin user, the channels will start at sequence 0, which is omited from TimedSet.String(). Create a RedactorBuilder on TimedSet to log the zero values.
- Change redaction from <ud>A.1,B.2</ud> to <ud>A</ud>.1,<ud>B</ud> to be able to identify redacted channels. Sequence numbers are not sensitive data.

Old logs:

```
Changes: c:#003 db:db MultiChangesFeed(channels: {<ud>*</ud>}, options: {Since: 0, Limit: 0, Conflicts: false, IncludeDocs: false, Wait: false, Continuous: false, HeartbeatMs: 0, TimeoutMs: 300000, ActiveOnly: false, Revocations: false, RequestPlusSeq: 0}) ...
Changes+: c:#003 db:db MultiChangesFeed: channels expand to "" ... <ud>  (to ADMIN)</ud>
```

New logs:

```
Changes: c:#003 db:db MultiChangesFeed(channels: {<ud>*</ud>}, options: {Since: 0, Limit: 0, Conflicts: false, IncludeDocs: false, Wait: false, Continuous: false, HeartbeatMs: 0, TimeoutMs: 300000, ActiveOnly: false, Revocations: false, RequestPlusSeq: 0}) ... <ud>  (to ADMIN)</ud>
Changes+: c:#003 db:db MultiChangesFeed: channels expand to "<ud>*</ud>:0" ... <ud>  (to ADMIN)</ud>
```

### DIFF
--- a/channels/timed_set.go
+++ b/channels/timed_set.go
@@ -10,7 +10,7 @@ package channels
 
 import (
 	"fmt"
-	"sort"
+	"slices"
 	"strconv"
 	"strings"
 
@@ -334,9 +334,19 @@ func TimedSetFromSequenceOnlySet(sequenceOnlySet map[string]uint64) TimedSet {
 // Encodes a TimedSet as a string (as sent in the public _changes feed.)
 // This string can later be turned back into a TimedSet by calling TimedSetFromString().
 func (set TimedSet) String() string {
-	var items []string
+	// setting to false is necessary to preserve the round trip compatbility.
+	items := set.asStrings(false, nil)
+	return strings.Join(items, ",")
+}
+
+// asStrings returns a sorted list of items in the list, appropriate for logging or serialization. show_zero_sequence is the preferred format unless the string needs to be backward compatible. Passing a redactorFunc is optional.
+func (set TimedSet) asStrings(show_zero_sequence bool, redactorFunc func(any) base.RedactorFunc) []string {
+	items := make([]string, 0, len(set))
 	for channel, vbSeq := range set {
-		if vbSeq.Sequence > 0 {
+		if show_zero_sequence || vbSeq.Sequence > 0 {
+			if redactorFunc != nil {
+				channel = redactorFunc(channel).Redact()
+			}
 			if vbSeq.VbNo != nil {
 				items = append(items, fmt.Sprintf("%s:%d.%d", channel, *vbSeq.VbNo, vbSeq.Sequence))
 			} else {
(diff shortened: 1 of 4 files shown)


bbrks:
I'm trying to wrap my head around what a channel with a `0` sequence even means to review this (in particular the changes mentioned in the 2nd bullet) and I cannot find a definitive answer anywhere, but I can see a lot of uses of `0` sequence (via `TimedSet.Add()`) - which makes me think this change is incorrect.

Does a `0` sequence channel mean it was not granted, and thus no access - which I think is what this logging change implies... or that it _was_ granted and there is no start sequence for which the grant applies - in which case we'll _want_ to log that channel.

Both I can see being true and I'm not sure which one is actually the case.

torcolvin:
> I'm trying to wrap my head around what a channel with a `0` sequence even means to review this and I cannot find a definitive answer anywhere, but I can see a lot of uses of `0` sequence (via `TimedSet.Add()`)
> 
> Does it mean it was not granted, and thus no access, or it _was_ granted and there is no start sequence for which the grant applies? Both I can see being true and I'm not sure which one is actually the case.

Maybe there's a better way to log this, the state that this happens is always https://github.com/couchbase/sync_gateway/blob/848960a5b3952de5e60eec0ca7bc0e1176664fa8/db/changes.go#L712 which means take all the channels explicitly for the admin user. It should only happen for the admin user. These channels are not expanded, so maybe we shouldn't have logs about: `base.DebugfCtx(ctx, base.KeyChanges, "MultiChangesFeed: channels expand to %#v ... %s", base.UD(channelsSince), base.UD(to))` but rather `base.DebugfCtx(ctx, base.KeyChanges, "MultiChangesFeed: channels are %#v ... for admin user", base.UD(channelsSince), base.UD(to))`. A different way that might be more logical is to change `channelsSince = channels.AtSequence(chans, 0)` to `channelsSince = channels.AtSequence(chans, 1)`, which I think would be equivalent in functionality.

This came out of a bug in 2.8.3 where deleting a user between the web auth and starting the changes feed accidentally upgraded from a user scope to admin. This was incidentally fixed, but it was confusing to debug.

bbrks (approved):

PR #7187, opened 2024-11-08
CBG-4263 create single actor tests

torcolvin:
CBG-4263 create the full single actor tests, though they are mostly skipped due to intermittent failures and lack of implementation of CBL. There are subsequent tickets for addressing them.

I added more optional data to `DocVersion` to allow seeing more information on assertions. This information isn't always going to be populated, but it is useful for printing out log message to see debugging.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2794/

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -1797,7 +1797,7 @@ func (db *DatabaseCollectionWithUser) storeOldBodyInRevTreeAndUpdateCurrent(ctx 
 	// Store the new revision body into the doc:
 	doc.setRevisionBody(ctx, newRevID, newDoc, db.AllowExternalRevBodyStorage(), newDocHasAttachments)
 	doc.SyncData.Attachments = newDoc.DocAttachments
-	doc.metadataOnlyUpdate = newDoc.metadataOnlyUpdate
+	doc.MetadataOnlyUpdate = newDoc.MetadataOnlyUpdate
 
 	if doc.CurrentRev == newRevID {
 		doc.NewestRev = ""
@@ -1808,7 +1808,7 @@ func (db *DatabaseCollectionWithUser) storeOldBodyInRevTreeAndUpdateCurrent(ctx 
 		if doc.CurrentRev != prevCurrentRev {
 			doc.promoteNonWinningRevisionBody(ctx, doc.CurrentRev, db.RevisionBodyLoader)
 			// If the update resulted in promoting a previous non-winning revision body to winning, this isn't a metadata only update.
-			doc.metadataOnlyUpdate = nil
+			doc.MetadataOnlyUpdate = nil
 		}
 	}
 }
@@ -2088,8 +2088,8 @@ func (col *DatabaseCollectionWithUser) documentUpdateFunc(
 		return
 	}
 
-	// compute mouMatch before the callback modifies doc.metadataOnlyUpdate
-	mouMatch := doc.metadataOnlyUpdate != nil && base.HexCasToUint64(doc.metadataOnlyUpdate.CAS) == doc.Cas
+	// compute mouMatch before the callback modifies doc.MetadataOnlyUpdate
+	mouMatch := doc.MetadataOnlyUpdate != nil && base.HexCasToUint64(doc.MetadataOnlyUpdate.CAS) == doc.Cas
 	// Invoke the callback to update the document and with a new revision body to be used by the Sync Function:
 	newDoc, newAttachments, createNewRevIDSkipped, updatedExpiry, err := callback(doc)
 	if err != nil {
@@ -2322,8 +2322,8 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 			updatedDoc.Spec = appendRevocationMacroExpansions(updatedDoc.Spec, revokedChannelsRequiringExpansion)
 
 			updatedDoc.IsTombstone = currentRevFromHistory.Deleted
-			if doc.metadataOnlyUpdate != nil {
-				if doc.metadataOnlyUpdate.CAS != "" {
+			if doc.MetadataOnlyUpdate != nil {
+				if doc.MetadataOnlyUpdate.CAS != "" {
 					updatedDoc.Spec = append(updatedDoc.Spec, sgbucket.NewMacroExpansionSpec(XattrMouCasPath(), sgbucket.MacroCas))
 				}
 			} else {
@@ -2386,8 +2386,8 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 		} else if doc != nil {
 			// Update the in-memory CAS values to match macro-expanded values
 			doc.Cas = casOut
-			if doc.metadataOnlyUpdate != nil && doc.metadataOnlyUpdate.CAS == expandMacroCASValueString {
-				doc.metadataOnlyUpdate.CAS = base.CasToString(casOut)
+			if doc.MetadataOnlyUpdate != nil && doc.MetadataOnlyUpdate.CAS == expandMacroCASValueString {
+				doc.MetadataOnlyUpdate.CAS = base.CasToString(casOut)
 			}
 			// update the doc's HLV defined post macro expansion
 			doc = postWriteUpdateHLV(doc, casOut)
--- a/db/crud.go
+++ b/db/crud.go
@@ -901,9 +901,9 @@ func (db *DatabaseCollectionWithUser) updateHLV(ctx context.Context, d *Document
 	hasHLV := d.HLV != nil
 	if d.HLV == nil {
 		d.HLV = &HybridLogicalVector{}
-		base.DebugfCtx(ctx, base.KeySGTest, "No existing HLV for doc %s", base.UD(d.ID))
+		base.DebugfCtx(ctx, base.KeyVV, "No existing HLV for doc %s", base.UD(d.ID))
 	} else {
-		base.DebugfCtx(ctx, base.KeySGTest, "Existing HLV for doc %s before modification %+v", base.UD(d.ID), d.HLV)
+		base.DebugfCtx(ctx, base.KeyVV, "Existing HLV for doc %s before modification %+v", base.UD(d.ID), d.HLV)
 	}
 	switch docUpdateEvent {
 	case ExistingVersion:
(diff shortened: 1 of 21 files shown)


db/crud.go
```diff
@@ -896,11 +896,14 @@ func (db *DatabaseCollectionWithUser) OnDemandImportForWrite(ctx context.Context
 }
 
 // updateHLV updates the HLV in the sync data appropriately based on what type of document update event we are encountering. mouMatch represents if the _mou.cas == doc.cas
-func (db *DatabaseCollectionWithUser) updateHLV(d *Document, docUpdateEvent DocUpdateType, mouMatch bool) (*Document, error) {
+func (db *DatabaseCollectionWithUser) updateHLV(ctx context.Context, d *Document, docUpdateEvent DocUpdateType, mouMatch bool) (*Document, error) {
 
 	hasHLV := d.HLV != nil
 	if d.HLV == nil {
 		d.HLV = &HybridLogicalVector{}
+		base.DebugfCtx(ctx, base.KeySGTest, "No existing HLV for doc %s", base.UD(d.ID))
```
bbrks: It feels like we should be putting these under the CRUD log key, rather than something test-specific?

```suggestion
		base.DebugfCtx(ctx, base.KeyCRUD, "No existing HLV for doc %s", base.UD(d.ID))
```

rest/utilities_testing.go
```diff
@@ -2448,12 +2448,15 @@ func WaitAndAssertBackgroundManagerExpiredHeartbeat(t testing.TB, bm *db.Backgro
...
 // DocVersion represents a specific version of a document in an revID/HLV agnostic manner.
 type DocVersion struct {
-	RevTreeID string
-	CV        db.Version
+	RevTreeID     string                  // RevTreeID is the rev treee ID of a document, may be empty not present
+	HLV           *db.HybridLogicalVector // HLV is the hybrid logical vector of the document, may not be present
+	Mou           *db.MetadataOnlyUpdate  // Mou is the metadata only update of the document, may not be present
+	Cas           uint64                  // Cas is the cas value of the document
+	HasImplicitCV bool                    // If true, the HLV was constructed from cas@sourceID instead of from _vv, used for documents written to Couchbase Server without Sync Gateway
 }
```
bbrks: This change is about equivalent to storing the whole rev tree inside `DocVersion` -  deviating from the original intention of `DocVersion` to only represent a single version of a document. By doing this we're overloading the term "Version" again to mean something else.

Refer to this thread attempting to disambiguate the names of these things from last year:
https://couchbase.slack.com/archives/GCCJPDTRV/p1702317050503909 

Looking at how `DocVersion` is mainly used (i.e. `UpdateDoc`, `WaitForVersion`, etc.), this doesn't feel like an appropriate change to make to this struct.

I agree it might be useful for some diagnosis if we don't otherwise have access to the document itself to grab the extended metadata - but how often is that the case, if at all?

topologytest/hlv_test.go
```diff
@@ -29,50 +28,182 @@ func getSingleDsName() base.ScopeAndCollectionName {
...
+}
+
+// TestHLVUpdateDocumentSingleActor tests creating a document with a single actor in different topologies.
+func TestHLVUpdateDocumentSingleActor(t *testing.T) {
+
+	base.SetUpTestLogging(t, base.LevelDebug, base.KeyCRUD, base.KeyImport, base.KeySGTest)
+	for _, tc := range getSingleActorTestCase() {
+		t.Run(tc.description(), func(t *testing.T) {
+			if strings.HasPrefix(tc.activePeerID, "cbl") {
+				t.Skip("Skipping Couchbase Lite test, returns unexpected body in proposeChanges: [304], CBG-4335")
```
bbrks: Wrong Jira ticket (Go upgrade)

topologytest/hlv_test.go
```diff
@@ -29,50 +28,182 @@ func getSingleDsName() base.ScopeAndCollectionName {
...
+			}
+			if base.UnitTestUrlIsWalrus() {
+				t.Skip("rosmar failure to investigate CBG-4329")
```
bbrks: Wrong Jira ticket (refers to CB Server only?)

topologytest/hlv_test.go
```diff
@@ -29,50 +28,182 @@ func getSingleDsName() base.ScopeAndCollectionName {
...
+// TestHLVResurrectDocumentSingleActor tests resurrect a document with a single actor in different topologies.
+func TestHLVResurrectDocumentSingleActor(t *testing.T) {
+
+	base.SetUpTestLogging(t, base.LevelDebug, base.KeyImport, base.KeySGTest)
+	for _, tc := range getSingleActorTestCase() {
+		t.Run(tc.description(), func(t *testing.T) {
+			if strings.HasPrefix(tc.activePeerID, "cbl") {
+				t.Skip("Skipping Couchbase Lite test, does not know how to push a deletion yet")
+			}
+			t.Skip("Skipping Sync Gateway test for rosmar, intermittent failures CBG-4239")
```
bbrks: Wrong Jira ticket (doesn't exist?)

bbrks (changes requested):
- I am unsure about the change to `DocVersion`
- The linked Jira tickets in the `t.Skip` seem wrong, they refer to things other than the skip reason

db/crud.go
torcolvin: I was very reluctant to put these under the existing CRUD key - these are going to be _very_ noisy for anything but topology tests and we wouldn't have this on in production. I am very open to ideas of how to solve the problem.

I thought about:

- Putting these at `KeyCRUD` but Trace, but then if you want this logging, you have to opt-in for trace logging for everything, and this is prohibitive to analyze for.
- Create a specific topology test key (but KeySGTest is so ill used anyway, I thought it was the same)
- Create a function like `base.TestInfofCtx` that would compile away if `cb_sg_devmode` was not used - we pretty much don't need this outside of running a specific test environment, that I'm aware of.

That said, if something _did_ go wrong in production or say a QE/perf test this is useful to know about.

rest/utilities_testing.go
torcolvin: I do think I'm overloading this datastructure, but this information turns out to be vital to debugging and it is very hard to look up retrospectively, if at all. If this is rosmar, the information will have gone out of memory. Additionally, since XDCR is in play, the information is differently when this is logged vs when you look at it in on the server, so this information is effectively lost.

I think we could create a datastructure specific to the topologytests to hold this to store this information, it doesn't need to be in the REST packages. I don't know the right answer to this, but this might be useful when the blip tests are modified to make this comparison worthwhile https://github.com/couchbase/sync_gateway/blob/50438755357f8dffc3ca2a96572050068a2421d1/rest/utilities_testing.go#L2459.

In practice an equality comparison is either for `DocVersion.CV` or `DocVersion.RevTreeID`, and possibly both? There are some weird situations you end up in a testing scenario where you might add a document using SDK and only have a CV, or a testing scenario where you add a document via CBL 3.x and only have a revtree ID.

I think I'd want to have some datastructure that represents this in the topology tests. The reason that I pushed this down to the rest tester is that I was thinking about implementing `SyncGatewayPeer.WriteDocument` using functions like `RestTester.PutDocDirectly`. I believe that I ripped this out for two reasons:

- `RestTester` put functions all use optimistic concurrency via revtreeids and we don't have access to those.
- ultimately the topology tests are written with LWW and so I just dropped out the middle functions of `RestTester` to use a RetryLoop on crud.go directly. I'm not sure we want to drop this. But you could imagine a function that does optimistic concurrency at the RestTester level.

I'm not committed to one way of doing this, but I definitely want to leave enough code here that I'm not constantly re-adding tracing code to topologytests to tell what was going on. I also am probably too cautious to avoid our many levels of blip testing wrapper code.

db/crud.go
bbrks: I'm certainly open to keeping at debug with a version-vector specific log key, at least for the beta and early 4.x versions until we're more confident about the implementation.

This is more reason to drop the SGTest key, I think we're going to need it for QE and beta issues.

rest/utilities_testing.go
bbrks: > There are some weird situations you end up in a testing scenario where you might add a document using SDK and only have a CV, or a testing scenario where you add a document via CBL 3.x and only have a revtree ID.
> RestTester put functions all use optimistic concurrency via revtreeids and we don't have access to those.

That was the point of this struct. So we have a single type that we are able to use to pass into MVCC functions that will work in all cases. Optimistic concurrency is the default behaviour of Couchbase Mobile, and I don't think we should be dropping that generally.

A subset of tests that rely on LWW seems OK, but it's not really how real-world deployments work and we should strive to keep the test APIs with MVCC in mind where we can, and allow simpler LWW fallback options where we do not want to deal with the complexity of multiple writers.

bbrks: > vital to debugging and it is very hard to look up retrospectively, if at all. If this is rosmar, the information will have gone out of memory. 

Can you elaborate on how we lose information about a documents full HLV? This seems unexpected to me.

torcolvin: A lot of confusion comes around the fact that documents are continually getting imported in these tests, and also by the nature of XDCR not copying over the `_sync` data structure.

Let's take a pretty simple case:

```mermaid
flowchart LR
    subgraph clusterA
        CBS1
        SG1
    end
    subgraph clusterB
        CBS2
        SG2
    end
    CBS1<-->CBS2 
```

If we have a single actor test that is doing the following write a document, then update the document on CBS1 as an SDK write, there are actually numerous states that the document will be in, and it can be challenging to debug.

- SDK write to CBS1, no `_vv`, `_mou`, cas1
- xdcr to CBS2, `_vv: { cvCAS: cas1, ver: cas1, src: cbs1}`, cas1
- import on CBS1: creates `_vv: { cvCAS: cas1, ver: cas1, src: cbs1}`,  `_mou: {cas: cas2, pCas: cas1}`, cas2
- import on CBS2: creates `_vv: { cvCAS: cas1, ver: cas1, src: cbs1}`,  `_mou: {cas: cas3, pCas: cas1}`, cas3

At this point, all documents are synchronized with a CV, but mismatched cas values. There are other scenarios at this point depending on whether import on CBS1 occurred first, then CBS2 would replicate the document with cas2, but there would still be a subsequent import.

Then, going to the second part of the test, an update of the document. Not knowing what the state of the cas, _vv, _mou mean it can be really challenging to debug where the values came from. If `doc.cas = _mou.cas` the evaluation of XDCR will be different. In this case we can see that the `_mou` looks different, but as soon as another write occurs it will be challenging to determine where it is from.

I agree that we aren't always going to want LWW, which is why I thought about pushing the extra information down to the `DocVersion` data structure.

torcolvin: Addressed by creating a datastructure local to topologytest as `DocMetadata`.

bbrks (approved):

PR #7194, opened 2024-11-15
Cover `UpdateSessionUUID` with auth unit test

bbrks:
Cover `user.UpdateSessionUUID` with a unit test to ensure it invalidates previously created sessions.

The use of the Admin API endpoint was covered in `TestSessionAPI` but it doesn't acutally verify existing sessions are made unusable.

- n/a

### DIFF
--- a/auth/session_test.go
+++ b/auth/session_test.go
@@ -312,3 +312,40 @@ func TestUserWithoutSessionUUID(t *testing.T) {
 	require.NoError(t, err)
 
 }
+
+// TestUserDeleteAllSessions changes the session UUID on a user such that existing sessions should not be usable.
+func TestUserDeleteAllSessions(t *testing.T) {
+	ctx := base.TestCtx(t)
+	testBucket := base.GetTestBucket(t)
+	defer testBucket.Close(ctx)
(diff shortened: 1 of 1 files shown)


auth/session_test.go
```diff
@@ -312,3 +312,40 @@ func TestUserWithoutSessionUUID(t *testing.T) {
...
+	_, err = auth.AuthenticateCookie(request, recorder)
+	require.NoError(t, err)
+
+	// h.deleteUserSessions() equivalent
+	user.UpdateSessionUUID()
+	err = auth.Save(user)
+	require.NoError(t, err)
+
+	_, err = auth.AuthenticateCookie(request, recorder)
+	require.EqualError(t, err, "401 Session no longer valid for user")
```
torcolvin, line 350: nit, take or leave:

I worry a bit about string matching and the formatting of this text and would have a slight preference for defining. `base.HTTPErrorf(http.StatusUnauthorized, "Session no longer valid for user")` as a constant and doing a errors.Is comparison. We do want to ultimately make sure that this comes back with 401 error and doing that means passing HTTPError up the stack.

torcolvin (approved):

PR #7201, opened 2024-11-18
CBG-4370 create a cancel context inside BlipSyncContext

torcolvin:
CBG-4370 create a cancel context inside BlipSyncContext

This cancel context allows a forceable closure of the underlying blip connection.

In the case there is a continuous pull replication and there is an error on the changes feed, the only way to stop the pull replication is to shut down the connection. CBL clients do not listen to unsolicited error messages.

This leverages https://github.com/couchbase/sync_gateway/commit/6317f1977e7bab1390ba2fc2cf339913763066be 

I added some leaky bucket code to be able to trigger the error.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2826/


db/changes.go
```diff
@@ -1416,6 +1416,7 @@ func GenerateChanges(ctx context.Context, database *DatabaseCollectionWithUser,
 	var lastSeq SequenceID
 	var feed <-chan *ChangeEntry
 	var timeout <-chan time.Time
+	var feedErr error
```
adamcfraser, line 1419: This looks like it's shadowed by the error on line 1440.  Is that intentional or should line 1440 now be removed?

db/blip_handler.go
```diff
@@ -429,7 +434,7 @@ func (flag changesDeletedFlag) HasFlag(deletedFlag changesDeletedFlag) bool {
 }
 
 // Sends all changes since the given sequence
-func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions) (isComplete bool) {
+func (bh *blipHandler) sendChanges(sender *blip.Sender, opts *sendChangesOptions) (bool, error) {
```
adamcfraser, line 436: If we're going to remove the 'isComplete' name, you probably want to add a comment to describe what the return value is supposed to be.

db/active_replicator.go
```diff
@@ -208,9 +208,10 @@ func connect(arc *activeReplicatorCommon, idSuffix string) (blipSender *blip.Sen
 	arc.replicationStats.NumConnectAttempts.Add(1)
 
 	var originPatterns []string // no origin headers for ISGR
-	// NewSGBlipContext doesn't set cancellation context - active replication cancellation on db close is handled independently
-	blipContext, err := NewSGBlipContext(arc.ctx, arc.config.ID+idSuffix, originPatterns, nil)
+	cancelCtx, cancelFunc := context.WithCancel(context.Background())
+	blipContext, err := NewSGBlipContext(arc.ctx, arc.config.ID+idSuffix, originPatterns, cancelCtx)
 	if err != nil {
+		cancelFunc()
```
adamcfraser, line 214: When considering possible backports of this fix, should we omit the ISGR changes?

db/blip_sync_context.go
```diff
@@ -34,7 +34,10 @@ const (
 
 var ErrClosedBLIPSender = errors.New("use of closed BLIP sender")
 
-func NewBlipSyncContext(ctx context.Context, bc *blip.Context, db *Database, contextID string, replicationStats *BlipSyncStats) *BlipSyncContext {
+func NewBlipSyncContext(ctx context.Context, bc *blip.Context, db *Database, contextID string, replicationStats *BlipSyncStats, ctxCancelFunc context.CancelFunc) (*BlipSyncContext, error) {
+	if ctxCancelFunc == nil {
```
adamcfraser, line 38: Is this required for cases where the BlipSyncContext is for an active peer (ISGR)?  Closing the connection in that case feels like a different problem than the passive peer case.

rest/blip_api_crud_test.go
```diff
@@ -3185,3 +3186,44 @@ func TestBlipDatabaseClose(t *testing.T) {
 		}, time.Second*10, time.Millisecond*100)
 	})
 }
+
+// Starts a continuous pull replication then updates the db to trigger a close.
+func TestChangesFeedExitDisconnect(t *testing.T) {
+
+	base.SetUpTestLogging(t, base.LevelInfo, base.KeyHTTP, base.KeySync, base.KeySyncMsg, base.KeyChanges, base.KeyCache)
+	btcRunner := NewBlipTesterClientRunner(t)
+	var shouldChanenlQueryError atomic.Bool
```
adamcfraser:
```suggestion
	var shouldChannelQueryError atomic.Bool
```

rest/blip_sync.go
```diff
@@ -51,17 +51,22 @@ func (h *handler) handleBLIPSync() error {
 	// error is checked at the time of database load, and ignored at this time
 	originPatterns, _ := hostOnlyCORS(h.db.CORS.Origin)
 
+	cancelCtx, cancelCtxFunc := context.WithCancel(h.db.DatabaseContext.CancelContext)
 	// Create a BLIP context:
-	blipContext, err := db.NewSGBlipContext(h.ctx(), "", originPatterns, h.db.DatabaseContext.CancelContext)
+	blipContext, err := db.NewSGBlipContext(h.ctx(), "", originPatterns, cancelCtx)
 	if err != nil {
+		cancelCtxFunc()
```
adamcfraser, line 58: Is this the only error exit where we need to ensure cancelCtxFunc() is called?  Is it needed in the PostHandshakeCallback below, or is that handled elsewhere?

adamcfraser (changes requested):
A few comments/questions.

rest/blip_sync.go
torcolvin: This is handled in `BlipSyncContext.Close`

db/changes.go
torcolvin: fixed

db/active_replicator.go
torcolvin: I think that we should keep ISGR fixes because it can hang in the same way, see new tests.

adamcfraser (approved):
LGTM - thanks for the additional ISGR push test, that was informative.

PR #7202, opened 2024-11-19
CBG-4302: create multi actor tests

gregns1:
[CBG-4302](https://jira.issues.couchbase.com/browse/CBG-4302)

- add multi actor tests for non conflicting mutations 
- skips in place for intermittent failures - links to appropriate tickets 

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2796/

[CBG-4302]: https://couchbasecloud.atlassian.net/browse/CBG-4302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/topologytest/couchbase_server_peer_test.go
+++ b/topologytest/couchbase_server_peer_test.go
@@ -170,9 +170,10 @@ func (p *CouchbaseServerPeer) waitForDocVersion(dsName sgbucket.DataStoreName, d
 		var xattrs map[string][]byte
 		var cas uint64
 		docBytes, xattrs, cas, err = p.getCollection(dsName).GetWithXattrs(p.Context(), docID, []string{base.VvXattrName})
-		if !assert.NoError(c, err) {
+		if err != nil && base.IsDocNotFoundError(err) {
 			return
 		}
+		assert.NoError(c, err)
 		// have to use p.tb instead of c because of the assert.CollectT doesn't implement TB
 		version = getDocVersion(docID, p, cas, xattrs)
 		assert.Equal(c, expected.CV(), version.CV(), "Could not find matching CV on %s for peer %s (sourceID:%s)\nexpected: %+v\nactual:   %+v\n          body: %+v\n", docID, p, p.SourceID(), expected, version, string(docBytes))
--- a/topologytest/couchbase_server_peer_test.go
+++ b/topologytest/couchbase_server_peer_test.go
@@ -170,10 +170,9 @@ func (p *CouchbaseServerPeer) waitForDocVersion(dsName sgbucket.DataStoreName, d
 		var xattrs map[string][]byte
 		var cas uint64
 		docBytes, xattrs, cas, err = p.getCollection(dsName).GetWithXattrs(p.Context(), docID, []string{base.VvXattrName})
-		if err != nil && base.IsDocNotFoundError(err) {
+		if !assert.NoError(c, err) {
 			return
 		}
-		assert.NoError(c, err)
 		// have to use p.tb instead of c because of the assert.CollectT doesn't implement TB
 		version = getDocVersion(docID, p, cas, xattrs)
 		assert.Equal(c, expected.CV(), version.CV(), "Could not find matching CV on %s for peer %s (sourceID:%s)\nexpected: %+v\nactual:   %+v\n          body: %+v\n", docID, p, p.SourceID(), expected, version, string(docBytes))
--- a/topologytest/couchbase_server_peer_test.go
+++ b/topologytest/couchbase_server_peer_test.go
@@ -18,7 +18,6 @@ import (
 	sgbucket "github.com/couchbase/sg-bucket"
 	"github.com/couchbase/sync_gateway/base"
 	"github.com/couchbase/sync_gateway/db"
-	"github.com/couchbase/sync_gateway/rest"
 	"github.com/couchbase/sync_gateway/xdcr"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -290,12 +289,4 @@ func getBodyAndVersion(peer Peer, collection sgbucket.DataStore, docID string) (
 	var body db.Body
 	require.NoError(peer.TB(), base.JSONUnmarshal(docBytes, &body))
 	return getDocVersion(docID, peer, cas, xattrs), body
-}
-
-// getFunctionalCV returns a functional version from a DocVersion, which will  CV from a DocVersion.
-func getCVBeforeImport(version rest.DocVersion) db.Version {
-	if version.Cas != 0 && version.Mou != nil && version.HLV.CurrentVersionCAS != version.Cas {
-		return db.Version{SourceID: version.HLV.SourceID, Value: base.HexCasToUint64(version.Mou.PreviousCAS)}
-	}
-	return db.Version{SourceID: version.HLV.SourceID, Value: version.HLV.Version}
 }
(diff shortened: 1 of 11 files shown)


topologytest/couchbase_server_peer_test.go
```diff
@@ -170,9 +170,10 @@ func (p *CouchbaseServerPeer) waitForDocVersion(dsName sgbucket.DataStoreName, d
 		var xattrs map[string][]byte
 		var cas uint64
 		docBytes, xattrs, cas, err = p.getCollection(dsName).GetWithXattrs(p.Context(), docID, []string{base.VvXattrName})
-		if !assert.NoError(c, err) {
+		if err != nil && base.IsDocNotFoundError(err) {
```
torcolvin: What's the purpose of this? If the error is doc not found then it won't fall through to the assertions and I think the test will pass, even though no document is found.

topologytest/hlv_test.go
```diff
@@ -73,12 +120,30 @@ func TestHLVCreateDocumentSingleActor(t *testing.T) {
...
+	for _, tc := range getMultiActorTestCases() {
+		t.Run(tc.description(), func(t *testing.T) {
+			peers, _ := setupTests(t, tc.topology, "")
+
+			for peerName, _ := range peers {
+				tc.originalActor = peerName
+				docID := tc.docID(tc.getOriginalActor())
+				docBody := []byte(fmt.Sprintf(`{"peer": "%s", "topology": "%s"}`, peerName, tc.description()))
+				docVersion := peers[peerName].CreateDocument(tc.collectionName(), docID, docBody)
+				waitForVersionAndBody(t, tc, peers, docVersion, docBody)
```
torcolvin: I think these tests might want to do the write in a loop and then wait in a separate loop. Otherwise this is very similar to the single actor tests.

You could even imagine doing the writes _very_ simultaneously in a goroutine but I think starting in a loop is fine.

topologytest/hlv_test.go
```diff
@@ -93,22 +158,64 @@ func TestHLVUpdateDocumentSingleActor(t *testing.T) {
...
+			peers, _ := setupTests(t, tc.topology, "")
+
+			for peerName, _ := range peers {
+				tc.originalActor = peerName
+				docID := tc.docID(tc.getOriginalActor())
+				body1 := []byte(fmt.Sprintf(`{"originPeer": "%s", "topology": "%s", "writePeer": "%s"}`, peerName, tc.description(), peerName))
+				createVersion := peers[peerName].CreateDocument(tc.collectionName(), docID, body1)
+				waitForVersionAndBody(t, tc, peers, createVersion, body1)
+
+				peerList := maps.Keys(peers)
```
torcolvin: On the single actor tests I put `PeerNames` on the interface, and sorted them to provide some determinism when writing the tests. I think putting this on the general interface will shorten where you are iterating over peers in each test.

We could then decide later if we wanted to do a different order.

topologytest/hlv_test.go
```diff
@@ -93,22 +158,64 @@ func TestHLVUpdateDocumentSingleActor(t *testing.T) {
...
+				for _, updatePeer := range peerList {
+					if updatePeer == peerName {
+						// skip current peer in topology
+						continue
+					}
+					newBody := []byte(fmt.Sprintf(`{"originPeer": "%s", "topology": "%s", "writePeer": "%s"}`, peerName, tc.description(), updatePeer))
+					updateVersion := peers[updatePeer].WriteDocument(tc.collectionName(), docID, newBody)
+					t.Logf("createVersion: %+v, updateVersion: %+v", createVersion, updateVersion)
+					t.Logf("waiting for document %s peer write %s on all peers", docID, updatePeer)
+					waitForVersionAndBodyOnNonActivePeers(t, tc, peers, updateVersion, newBody, updatePeer)
```
torcolvin: I think this wait should be outside of the loop, and potentially have the write + update occur for each peer simultaneously.

torcolvin (commented):
I put comments on a few tests, but I think they apply to all of them, we want to make sure the actions are actually occurring simultaneously.

topologytest/couchbase_server_peer_test.go
gregns1: I'll change this. Need a way of accounting for time for doc to be replicated. I noticed sometimes after creating a doc a test would show not found on another peer. My reading into that was it wasn't waiting long enough to replicate.

topologytest/hlv_test.go
```diff
@@ -73,12 +133,41 @@ func TestHLVCreateDocumentSingleActor(t *testing.T) {
...
+	for _, tc := range getMultiActorTestCases() {
+		t.Run(tc.description(), func(t *testing.T) {
+			peers, _ := setupTests(t, tc.topology, "")
+
+			var docVersionList []BodyAndVersion
+			peerList := tc.PeerNames()
+
+			// grab sorted peer list and create a list to store expected version,
+			// doc body
+			for _, peerName := range peerList {
```
torcolvin, line 154: readability, the shorter the test code the easier to read IMO

```suggestion
			for peerName := range tc.PeerNames() {
```

topologytest/hlv_test.go
```diff
@@ -73,12 +133,41 @@ func TestHLVCreateDocumentSingleActor(t *testing.T) {
...
+				tc.originalActor = peerName
+				docID := tc.docID(tc.getOriginalActor())
```
torcolvin, line 156: I think just change this function `docID` to take an activePeer parameter and have topology be immutable.

topologytest/hlv_test.go
```diff
@@ -176,6 +346,49 @@ func TestHLVResurrectDocumentSingleActor(t *testing.T) {
...
+				resBody := []byte(fmt.Sprintf(`{"peer": "%s", "topology": "%s", "write": "resurrection on peer %s"}`, tc.originalActor, tc.description(), peerName))
+				updateVersion := peers[peerName].WriteDocument(tc.collectionName(), docID, resBody)
+				//waitForVersionAndBodyOnNonActivePeers(t, tc, peers, updateVersion, resBody, peerName)
+				docVersionList = append(docVersionList, BodyAndVersion{body: resBody, docMeta: updateVersion, updatePeer: peerName})
+			}
+
+			for i, updatePeer := range peerList {
+				tc.originalActor = updatePeer
+				docVersion := docVersionList[i]
+				waitForVersionAndBodyOnNonActivePeers(t, tc, peers, docVersion.docMeta, docVersion.body, docVersion.updatePeer)
```
torcolvin, line 386: You can have this function take docID so that topology.originalActor is not modified.

topologytest/hlv_test.go
```diff
@@ -187,18 +400,30 @@ func stripInternalProperties(body db.Body) {
...
-		body := peer.WaitForDocVersion(testCase.collectionName(), testCase.docID(), expectedVersion)
+		t.Logf("waiting for doc version on %s, written from %s", peer, testCase.getOriginalActor())
+		body := peer.WaitForDocVersion(testCase.collectionName(), testCase.docID(testCase.getOriginalActor()), expectedVersion)
+		requireBodyEqual(t, expectedBody, body)
+	}
+}
+
+// waitForVersionAndBodyOnNonActivePeers waits for a document version on all peers that the update didn't occur on
+func waitForVersionAndBodyOnNonActivePeers(t *testing.T, testCase ActorTest, peers map[string]Peer, expectedVersion DocMetadata, expectedBody []byte, writingPeer string) {
+	delete(peers, writingPeer)
```
torcolvin, line 416: A map is a pointer here, so I might have the caller filter out the peer names since this worries me to do a modification. You should be able to just do `slices.Delete(maps.Keys(peerNames), peerNames)`.

torcolvin (commented):
Looks good, just some comments on how to improve readability of these tests. I have a strong desire to make `Topology` effectively immutable, so I wonder if we can just make functions to pass in `actor` where you are setting `tc.originalActor`.

PR #7203, opened 2024-11-20
CBG-4369 optionally return CV on rest API

torcolvin:
CBG-4369 optionally return CV on rest API

Implements getting cv out of the requests used in https://github.com/couchbaselabs/couchbase-lite-tests/blob/main/client/src/cbltest/api/syncgateway.py

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2807/

### DIFF
--- a/rest/bulk_api.go
+++ b/rest/bulk_api.go
@@ -25,6 +25,25 @@ import (
 	"github.com/couchbase/sync_gateway/db"
 )
 
+// allDocsRowValue is a struct that represents possible values returned in a document from /ks/_all_docs endpoint
+type allDocsRowValue struct {
+	Rev      string              `json:"rev"`
+	CV       string              `json:"cv,omitempty"`
+	Channels []string            `json:"channels,omitempty"`
+	Access   map[string]base.Set `json:"access,omitempty"` // for admins only
+}
+
+// allDocsRow is a struct that represents a linefrom /ks/_all_docs endpoint
+type allDocsRow struct {
+	Key       string           `json:"key"`
+	ID        string           `json:"id,omitempty"`
+	Value     *allDocsRowValue `json:"value,omitempty"`
+	Doc       json.RawMessage  `json:"doc,omitempty"`
+	UpdateSeq uint64           `json:"update_seq,omitempty"`
+	Error     string           `json:"error,omitempty"`
+	Status    int              `json:"status,omitempty"`
+}
+
 // HTTP handler for _all_docs
 func (h *handler) handleAllDocs() error {
 	// http://wiki.apache.org/couchdb/HTTP_Bulk_Document_API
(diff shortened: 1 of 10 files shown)


rest/bulk_api.go
```diff
@@ -276,9 +285,10 @@ func (h *handler) handleDump() error {
 func (h *handler) handleRepair() error {
 	// TODO: If repair is re-enabled, it may need to be modified to support xattrs and GSI
 	err := errors.New("_repair endpoint disabled")
-	if err != nil {
+	/*if err != nil {
```
adamcfraser: This looks like a change in auditing behaviour (to not audit when a no-op call to repair is made).  Does it make sense in this PR?

rest/bulk_api.go
```diff
@@ -364,6 +374,7 @@ func (h *handler) handleBulkGet() error {
 
 	includeAttachments := h.getBoolQuery("attachments")
 	showExp := h.getBoolQuery("show_exp")
+	showCV := h.getBoolQuery("show_cv")
```
adamcfraser, line 376: We should be consistent with the query parameter used to include cv in the responses.  I'm fine with `cvs=true` like you've got in the all docs case.

rest/doc_api.go
```diff
@@ -30,6 +30,7 @@ func (h *handler) handleGetDoc() error {
 	revid := h.getQuery("rev")
 	openRevs := h.getQuery("open_revs")
 	showExp := h.getBoolQuery("show_exp")
+	showCV := h.getBoolQuery("show_cv")
```
adamcfraser, line 33: As above, let's pick a single query param name and use it throughout.

rest/doc_api_test.go
```diff
@@ -173,3 +176,70 @@ func TestGuestReadOnly(t *testing.T) {
 	RequireStatus(t, response, http.StatusForbidden)
 
 }
+
+func TestGetDocWithCV(t *testing.T) {
```
adamcfraser, line 180: Do we have a test case for bulk_get?  I may just have missed it.

adamcfraser (changes requested):

rest/bulk_api.go
torcolvin: golangci-lint ci didn't pass without this change, I am not immediately sure why because it's not changed code.

rest/bulk_api.go
torcolvin: I did `show_cv` somewhat arbitrarily becuase I thought it made it clearer that it is an output parameter, especially when `rev` or `revs` can be input parameters to some functions.

Also cvs reminds me of a US pharmacy or the legacy version control system. I am happy to switch to cvs though.

rest/doc_api_test.go
torcolvin: Added missing test.

rest/bulk_api.go
torcolvin: Now it _did_ pass with this removed, I fixed this issue on main but anemone hasn't been rebased recently.

adamcfraser (approved):

PR #7205, opened 2024-11-22
CBG-4303: conflicting writes muti actor tests, skipping failures

gregns1:
CBG-4303

- Adds test for multi actor conflicting writes
- Skipping all tests due to general flake issues and panics against rosmar. tickets are [CBG-4379](https://jira.issues.couchbase.com/browse/CBG-4379) and [CBG-4378](https://jira.issues.couchbase.com/browse/CBG-4378) respectively 

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2812/

[CBG-4379]: https://couchbasecloud.atlassian.net/browse/CBG-4379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CBG-4378]: https://couchbasecloud.atlassian.net/browse/CBG-4378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/topologytest/couchbase_server_peer_test.go
+++ b/topologytest/couchbase_server_peer_test.go
@@ -54,11 +54,13 @@ func (r *CouchbaseServerReplication) PassivePeer() Peer {
 
 // Start starts the replication
 func (r *CouchbaseServerReplication) Start() {
+	r.t.Logf("starting XDCR replication")
 	require.NoError(r.t, r.manager.Start(r.ctx))
 }
 
 // Stop halts the replication. The replication can be restarted after it is stopped.
 func (r *CouchbaseServerReplication) Stop() {
+	r.t.Logf("stopping XDCR replication")
 	require.NoError(r.t, r.manager.Stop(r.ctx))
 }
 
@@ -83,16 +85,15 @@ func (p *CouchbaseServerPeer) GetDocument(dsName sgbucket.DataStoreName, docID s
 }
 
 // CreateDocument creates a document on the peer. The test will fail if the document already exists.
-func (p *CouchbaseServerPeer) CreateDocument(dsName sgbucket.DataStoreName, docID string, body []byte) DocMetadata {
-	p.tb.Logf("%s: Creating document %s", p, docID)
-	p.tb.Logf("bucket %s", p.bucket.GetName())
+func (p *CouchbaseServerPeer) CreateDocument(dsName sgbucket.DataStoreName, docID string, body []byte) BodyAndVersion {
+	p.tb.Logf("%s: Creating document %s in bucket %s", p, docID, p.bucket.GetName())
 	// create document with xattrs to prevent XDCR from doing a round trip replication in this scenario:
 	// CBS1: write document (cas1, no _vv)
 	// CBS1->CBS2: XDCR replication
 	// CBS2->CBS1: XDCR replication, creates a new _vv
 	cas, err := p.getCollection(dsName).WriteWithXattrs(p.Context(), docID, 0, 0, body, map[string][]byte{"userxattr": []byte(`{"dummy": "xattr"}`)}, nil, nil)
 	require.NoError(p.tb, err)
-	return DocMetadata{
+	docMetadata := DocMetadata{
 		DocID: docID,
 		Cas:   cas,
 		ImplicitCV: &db.Version{
@@ -100,10 +101,15 @@ func (p *CouchbaseServerPeer) CreateDocument(dsName sgbucket.DataStoreName, docI
 			Value:    cas,
 		},
 	}
+	return BodyAndVersion{
+		docMeta:    docMetadata,
+		body:       body,
+		updatePeer: p.name,
+	}
 }
 
 // WriteDocument writes a document to the peer. The test will fail if the write does not succeed.
-func (p *CouchbaseServerPeer) WriteDocument(dsName sgbucket.DataStoreName, docID string, body []byte) DocMetadata {
+func (p *CouchbaseServerPeer) WriteDocument(dsName sgbucket.DataStoreName, docID string, body []byte) BodyAndVersion {
 	p.tb.Logf("%s: Writing document %s", p, docID)
 	// write the document LWW, ignoring any in progress writes
 	callback := func(_ []byte) (updated []byte, expiry *uint32, shouldDelete bool, err error) {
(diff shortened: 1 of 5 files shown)


topologytest/couchbase_server_peer_test.go
```diff
@@ -85,6 +85,7 @@ func (p *CouchbaseServerPeer) GetDocument(dsName sgbucket.DataStoreName, docID s
 // CreateDocument creates a document on the peer. The test will fail if the document already exists.
 func (p *CouchbaseServerPeer) CreateDocument(dsName sgbucket.DataStoreName, docID string, body []byte) DocMetadata {
 	p.tb.Logf("%s: Creating document %s", p, docID)
+	p.tb.Logf("bucket %s", p.bucket.GetName())
```
torcolvin:
```suggestion
	p.tb.Logf("%s: Creating document %s in bucket %s", p, docID, p.bucket.GetName())
```

topologytest/hlv_test.go
```diff
@@ -64,6 +73,60 @@ func getSingleActorTestCase() []singleActorTest {
...
+		tests = append(tests, multiActorTest{topology: tc})
+	}
+	return tests
+}
+
+// BodyAndVersion struct to hold doc update information to assert on
+type BodyAndVersion struct {
+	docMeta    DocMetadata
+	body       []byte
+	updatePeer string
```
torcolvin: Can you document what updatePeer means?

topologytest/hlv_test.go
```diff
@@ -64,6 +73,60 @@ func getSingleActorTestCase() []singleActorTest {
...
+}
+
+func stopPeerReplications(t *testing.T, peerReplications []PeerReplication) {
+	t.Logf("stopping replications for peers")
```
torcolvin: Is it worth logging which peers are getting stopped, or loggin gin each replication?

topologytest/hlv_test.go
```diff
@@ -64,6 +73,60 @@ func getSingleActorTestCase() []singleActorTest {
...
+	for _, replication := range peerReplications {
+		replication.Stop()
+	}
+}
+
+func startPeerReplications(t *testing.T, peerReplications []PeerReplication) {
+	t.Logf("starting replications for peers")
```
torcolvin: same as above, should we log which replications are starting /stopping?

topologytest/hlv_test.go
```diff
@@ -72,9 +135,47 @@ func TestHLVCreateDocumentSingleActor(t *testing.T) {
 		t.Run(tc.description(), func(t *testing.T) {
 			peers, _ := setupTests(t, tc.topology, tc.activePeerID)
 
+			docID := tc.docID("")
```
torcolvin: should we have this just be tc.activePeerID for readability and matching the behavior?

topologytest/hlv_test.go
```diff
@@ -72,9 +135,47 @@ func TestHLVCreateDocumentSingleActor(t *testing.T) {
...
+	for _, tc := range getMultiActorTestCases() {
+		if strings.Contains(tc.description(), "CBL") {
+			// Test case flakes given the WaitForDocVersion function only waits for a docID on the cbl peer. We need to be
+			// able to wait for a specific version to arrive over pull replication
+			t.Skip("We need to be able to wait for a specific version to arrive over pull replication, CBG-4257")
+		}
+		t.Run(tc.description(), func(t *testing.T) {
+			peers, replications := setupTests(t, tc.topology, "")
+
+			stopPeerReplications(t, replications)
```
torcolvin: Is there a reason to stop replications before creating conflicting docs? Is this because `CreateDocument` can fail?

I think this is a good idea to debug tests at first but I think we'll want these writes to occur eventually while peers are active. I think this is worth marking as such and filing a followup ticket.

I think it might improve readability if we provide something like options to `setupTests` to say `startReplications` and `activePeer`. I am a big fan of `setupTestsOptions` structs so it's easy to understand what arguments like bools and strings mean.

 Actually I think `activePeer` might be now unused...

topologytest/hlv_test.go
```diff
@@ -72,9 +135,47 @@ func TestHLVCreateDocumentSingleActor(t *testing.T) {
...
+
+			docID := tc.docID("")
```
torcolvin: What do you think about having something like `singleDocID()` with no argument and `peerDocID(peerName string)` for readability?

topologytest/hlv_test.go
```diff
@@ -95,16 +196,67 @@ func TestHLVUpdateDocumentSingleActor(t *testing.T) {
 			}
 			peers, _ := setupTests(t, tc.topology, tc.activePeerID)
 
+			docID := tc.docID("")
 			body1 := []byte(fmt.Sprintf(`{"peer": "%s", "topology": "%s", "write": 1}`, tc.activePeerID, tc.description()))
-			createVersion := peers[tc.activePeerID].CreateDocument(tc.collectionName(), tc.docID(), body1)
+			createVersion := peers[tc.activePeerID].CreateDocument(tc.collectionName(), docID, body1)
```
torcolvin, line 193: should these documents return `BodyAndVersion` so you don't have to construct it everytime?

torcolvin (commented):
LGTM in concept, lots of readability comments which you can think about and take or leave.

Speaking of readability, do you think it would make sense to put the different test scenarios (single actor, multi actor non conflicting, multi actor conflicting) into their own files? I can see these getting out of hand but perhaps there aren't enough tests yet to warrant that.

topologytest/hlv_test.go
gregns1: So I'm stopping replications as the spec says to do so to create conflicting mutations. I think it makes sense at this stage to ensure we don't have timing issues etc. But I do agree we should have some concurrent updates but that maybe should be separate ticket.

torcolvin: Definitely a separate ticket!

gregns1: Filed [CBG-4380](https://jira.issues.couchbase.com/browse/CBG-4380) for it!

torcolvin (approved):

PR #7206, opened 2024-11-22
CBG-4365 rosmar xdcr, use _mou.cas for conflict resolution

torcolvin:
CBG-4365 rosmar xdcr, use _mou.cas for conflict resolution

I ran topologytests without a skip on `TestHLVUpdateDocumentSingleActor` and they now intermittently pass, rather than consistently fail without this patch.

The guts of this ticket are to consider the cas for conflict resolution in xdcr as actual cas, unless `_mou.cas == doc cas`, in which case, using `_hlv.cvCas` is correct. This matches the behavior of server XDCR.

`MetadataOnlyUpdate` is one of the only datastructures that we have that stores cas as a string, and that's fine for serialization but problematic for comparison. I decided to write method to convert the data rather than writing yet more comparisons in xdcr code. I don't think it makes sense to change the data on marshal and unmarshal due to the use of string comparison in `computeMetadataOnlyUpdate` but maybe having it convert all the values to hex on marshal/unmarshal would make more sense.

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2810/

### DIFF
--- a/xdcr/rosmar_xdcr.go
+++ b/xdcr/rosmar_xdcr.go
@@ -350,12 +350,13 @@ func getHLVAndMou(xattrs map[string][]byte) (*db.HybridLogicalVector, *db.Metada
 	return hlv, mou, nil
 }
 
-func updateHLV(xattrs map[string][]byte, sourceHLV *db.HybridLogicalVector, sourceMou *db.MetadataOnlyUpdate, sourceID string, copiedCas uint64) error {
+// updateHLV will update the xattrs on the target document considering the source's HLV, _mou, sourceID and cas.
+func updateHLV(xattrs map[string][]byte, sourceHLV *db.HybridLogicalVector, sourceMou *db.MetadataOnlyUpdate, sourceID string, sourceCas uint64) error {
 	// TODO: read existing targetXattrs[base.VvXattrName] and update the pv CBG-4250. This will need to merge pv from sourceHLV and targetHLV.
 	var targetHLV *db.HybridLogicalVector
 	// if source vv.cvCas == cas, the _vv.cv, _vv.cvCAS from the source is correct and we can use it directly.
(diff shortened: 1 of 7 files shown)


xdcr/rosmar_xdcr.go
```diff
@@ -373,3 +388,21 @@ func updateHLV(xattrs map[string][]byte, sourceHLV *db.HybridLogicalVector, sour
 	}
 	return nil
 }
+
+// getConflictResolutionCas returns cas for conflict resolution. If thethe conflict resolution cas.
```
gregns1: Typo here? I think it would be good to include on the comment the expected interaction between actual cas and mou cas, something like the below:
If we have:
- the _mou.cas == actualCas we will return _vv.cvCAS for conflict resolution, 
- if mou.Cas is out of date we will use _vv.ver
- if mou.Cas is not out of date but there is no HLV we return 0 to match XDCR

xdcr/rosmar_xdcr.go
```diff
@@ -342,21 +350,24 @@ func getHLVAndMou(xattrs map[string][]byte) (*db.HybridLogicalVector, *db.Metada
 	return hlv, mou, nil
 }
 
-func updateHLV(xattrs map[string][]byte, sourceHLV *db.HybridLogicalVector, sourceMou *db.MetadataOnlyUpdate, sourceID string, sourceCas uint64) error {
+func updateHLV(xattrs map[string][]byte, sourceHLV *db.HybridLogicalVector, sourceMou *db.MetadataOnlyUpdate, sourceID string, copiedCas uint64) error {
```
adamcfraser: I'm not sure what 'copiedCas' implies here (as opposed to source cas) - maybe a comment to explain?

adamcfraser (changes requested):

adamcfraser (approved):

PR #7213, opened 2024-11-26
Automatically collect heap profiles from high memory in tests

torcolvin:
This collects high heap profiles in the directory that the tests exist, every 5 secs. I could turn this into an opt-in variable - I think I always want this running locally. Having this run in github actions is useless because we don't upload the artifacts. Using in jenkins might be useful since often times the high memory usage comes from couchbase server. In this case, we could re-run locally however.

Unrelated, except in debugging the same problem, I changed the bucket names for rosmar so that it was easier to read.

### DIFF
--- a/base/main_test_bucket_pool.go
+++ b/base/main_test_bucket_pool.go
@@ -513,6 +513,10 @@ func (tbp *TestBucketPool) createTestBuckets(numBuckets, bucketQuotaMB int, buck
 	// create required number of buckets (skipping any already existing ones)
 	for i := 0; i < numBuckets; i++ {
 		testBucketName := fmt.Sprintf(tbpBucketNameFormat, tbpBucketNamePrefix, i, bucketNameTimestamp)
+		// shorten the name for rosmar buckets since they are in memory only
+		if UnitTestUrlIsWalrus() {
+			testBucketName = fmt.Sprintf("bucket%d", i)
+		}
 		ctx := BucketNameCtx(context.Background(), testBucketName)
 
 		// Bucket creation takes a few seconds for each bucket,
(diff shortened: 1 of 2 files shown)


base/main_test_bucket_pool.go
```diff
@@ -513,6 +513,10 @@ func (tbp *TestBucketPool) createTestBuckets(numBuckets, bucketQuotaMB int, buck
 	// create required number of buckets (skipping any already existing ones)
 	for i := 0; i < numBuckets; i++ {
 		testBucketName := fmt.Sprintf(tbpBucketNameFormat, tbpBucketNamePrefix, i, bucketNameTimestamp)
+		// shorten the name for rosmar buckets since they are in memory only
+		if UnitTestUrlIsWalrus() {
+			testBucketName = fmt.Sprintf("bucket%d", i)
+		}
```
bbrks, line 519: I'm OK with this change, but not really the profiling one.

It's probably best just to split this unrelated stuff even if it adds a few mins of branch/PR effort because it makes merging or reviewing easier.

bbrks (changes requested):
Did you see `SG_TEST_PROFILE_FREQUENCY`? That should do what this is doing (plus goroutines currently)

base/main_test_bucket_pool.go
torcolvin: Fair, I did see `SG_TEST_PROFILE_FREQUENCY` but the type of error I was trying to track down is an error that occurs only when running couchbase server integration tests for a long period of time, so I was worried about the number of profiles collected. But disk space is cheap and the profiles are small, so even for 30+ minute run this is fine.

I think I'll drop this PR.

PR #7226, opened 2024-12-03
CBG-4270: Implement sequence-based storage for BlipTesterClient

bbrks:
CBG-4270

BlipTesterClient now maintains a client-side sequence number, associated with every document update made on the client. This is independent of the Sync Gateway sequence.

Document storage is sequence-based, and we store all versions of documents for the purposes of history and test assertions. No cleanup is made on these old versions, since it's not expected that we'll be writing extremely long-running tests using this.

We build a client-side "changes feed" by iterating over all sequences and emitting the latest version of each document. There is no client-side checkpointing, but tests can start changes feeds from a specific sequence when required. Most tests can just use a continous changes feed running in the background.

The client's changes feed produces `ProposeChanges` in batches of only `1` item at a time, rather than a more real-world example of `20` - since the code to produce batches and a flush timeout of changes is not implemented. Filed CBG-4401 as follow-up.

## TODO
- [x] Self-review/cleanup
- [x] Ctx/close investigation
- [x] fix data race

```
2024-12-03T17:55:25.9788096Z ==================
2024-12-03T17:55:25.9788177Z WARNING: DATA RACE
2024-12-03T17:55:25.9788285Z Write at 0x00c000d1c108 by goroutine 198918:
2024-12-03T17:55:25.9788567Z   github.com/couchbase/sync_gateway/rest.(*BlipTesterCollectionClient).upsertDoc()
2024-12-03T17:55:25.9788906Z       /home/runner/work/sync_gateway/sync_gateway/rest/blip_client_test.go:1294 +0xcf5
2024-12-03T17:55:25.9789163Z   github.com/couchbase/sync_gateway/rest.(*BlipTesterCollectionClient).AddRev()
2024-12-03T17:55:25.9789496Z       /home/runner/work/sync_gateway/sync_gateway/rest/blip_client_test.go:1315 +0x99
2024-12-03T17:55:25.9789784Z   github.com/couchbase/sync_gateway/rest.(*BlipTestClientRunner).AddRev()
2024-12-03T17:55:25.9790112Z       /home/runner/work/sync_gateway/sync_gateway/rest/blip_client_test.go:1668 +0x55
2024-12-03T17:55:25.9790477Z   github.com/couchbase/sync_gateway/rest.TestMinRevPosWorkToAvoidUnnecessaryProveAttachment.func1()
2024-12-03T17:55:25.9790865Z       /home/runner/work/sync_gateway/sync_gateway/rest/attachment_test.go:2417 +0x86b
2024-12-03T17:55:25.9791099Z   github.com/couchbase/sync_gateway/rest.(*BlipTestClientRunner).Run.func2()
2024-12-03T17:55:25.9791424Z       /home/runner/work/sync_gateway/sync_gateway/rest/blip_client_test.go:857 +0xbd
2024-12-03T17:55:25.9791513Z   testing.tRunner()
2024-12-03T17:55:25.9791811Z       /opt/hostedtoolcache/go/1.23.3/x64/src/testing/testing.go:1690 +0x226
2024-12-03T17:55:25.9791921Z   testing.(*T).Run.gowrap1()
2024-12-03T17:55:25.9792207Z       /opt/hostedtoolcache/go/1.23.3/x64/src/testing/testing.go:1743 +0x44
2024-12-03T17:55:25.9792216Z 
2024-12-03T17:55:25.9792356Z Previous read at 0x00c000d1c108 by goroutine 199068:
2024-12-03T17:55:25.9792847Z   github.com/couchbase/sync_gateway/rest.(*BlipTesterCollectionClient).docsSince.func1.(*BlipTesterCollectionClient).OneShotDocsSince.1()
2024-12-03T17:55:25.9793171Z       /home/runner/work/sync_gateway/sync_gateway/rest/blip_client_test.go:109 +0x7ce
2024-12-03T17:55:25.9793454Z   github.com/couchbase/sync_gateway/rest.(*BlipTesterCollectionClient).docsSince.func1()
2024-12-03T17:55:25.9793778Z       /home/runner/work/sync_gateway/sync_gateway/rest/blip_client_test.go:119 +0x321
```

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2844/
- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2855/
- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2863/

### DIFF
--- a/base/collection_xattr.go
+++ b/base/collection_xattr.go
@@ -265,7 +265,7 @@ func (c *Collection) subdocGetBodyAndXattrs(ctx context.Context, k string, xattr
 				xattrs[xattrKey] = xattr
 			}
 
-			if len(xattrErrors) == len(xattrs) {
+			if len(xattrErrors) == len(xattrKeys) {
 				// No doc, no xattrs means the doc isn't found
 				return false, ErrNotFound, cas
 			}
--- a/base/collection_xattr.go
+++ b/base/collection_xattr.go
@@ -265,7 +265,7 @@ func (c *Collection) subdocGetBodyAndXattrs(ctx context.Context, k string, xattr
 				xattrs[xattrKey] = xattr
 			}
 
-			if len(xattrErrors) == len(xattrKeys) {
+			if len(xattrErrors) == len(xattrs) {
 				// No doc, no xattrs means the doc isn't found
 				return false, ErrNotFound, cas
 			}
--- a/db/crud.go
+++ b/db/crud.go
@@ -2189,8 +2189,15 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 		base.InfofCtx(ctx, base.KeyCRUD, "doc %q / %q, has been pruned, it has not been inserted into the revision cache", base.UD(docid), newRevID)
 	}
 
+	// If this update branched the revision tree, make a note of it in the 'Stored doc' log.
+	// The next time we'll see this is when the changes feed sends this revision.
+	inConflictLogSuffix := ""
+	if inConflict {
+		inConflictLogSuffix = " (branched)"
+	}
+
 	// Now that the document has successfully been stored, we can make other db changes:
-	base.DebugfCtx(ctx, base.KeyCRUD, "Stored doc %q / %q as #%v", base.UD(docid), newRevID, doc.Sequence)
+	base.DebugfCtx(ctx, base.KeyCRUD, "Stored doc %q / %q%s as #%v", base.UD(docid), newRevID, inConflictLogSuffix, doc.Sequence)
 
 	leafAttachments := make(map[string][]string)
 	if !skipObsoleteAttachmentsRemoval {
--- a/db/crud.go
+++ b/db/crud.go
@@ -2189,15 +2189,8 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 		base.InfofCtx(ctx, base.KeyCRUD, "doc %q / %q, has been pruned, it has not been inserted into the revision cache", base.UD(docid), newRevID)
 	}
 
-	// If this update branched the revision tree, make a note of it in the 'Stored doc' log.
-	// The next time we'll see this is when the changes feed sends this revision.
-	inConflictLogSuffix := ""
-	if inConflict {
-		inConflictLogSuffix = " (branched)"
-	}
-
 	// Now that the document has successfully been stored, we can make other db changes:
-	base.DebugfCtx(ctx, base.KeyCRUD, "Stored doc %q / %q%s as #%v", base.UD(docid), newRevID, inConflictLogSuffix, doc.Sequence)
+	base.DebugfCtx(ctx, base.KeyCRUD, "Stored doc %q / %q as #%v", base.UD(docid), newRevID, doc.Sequence)
 
 	leafAttachments := make(map[string][]string)
 	if !skipObsoleteAttachmentsRemoval {
--- a/rest/api_test_helpers.go
+++ b/rest/api_test_helpers.go
@@ -43,7 +43,7 @@ type PutDocResponse struct {
 
 // PutNewEditsFalse builds a new_edits=false style put to create a revision with the specified revID.
 // If parentRevID is not specified, treated as insert
-func (rt *RestTester) PutNewEditsFalse(docID string, newVersion DocVersion, parentVersion DocVersion, bodyString string) DocVersion {
+func (rt *RestTester) PutNewEditsFalse(docID string, newVersion DocVersion, parentVersion *DocVersion, bodyString string) *DocVersion {
 
 	var body db.Body
 	marshalErr := base.JSONUnmarshal([]byte(bodyString), &body)
@@ -55,10 +55,12 @@ func (rt *RestTester) PutNewEditsFalse(docID string, newVersion DocVersion, pare
 	revisions := make(map[string]interface{})
 	revisions["start"] = newRevGeneration
 	ids := []string{newRevDigest}
-	if parentVersion.RevID != "" {
-		_, parentDigest := db.ParseRevID(base.TestCtx(rt.TB()), parentVersion.RevID)
+	if parentVersion != nil {
+		parentVersionCopy := *parentVersion
+		_, parentDigest := db.ParseRevID(base.TestCtx(rt.TB()), parentVersionCopy.RevID)
 		ids = append(ids, parentDigest)
 	}
+	// TODO: Revs invalid???
 	revisions["ids"] = ids
 
 	requestBody[db.BodyRevisions] = revisions
@@ -69,7 +71,7 @@ func (rt *RestTester) PutNewEditsFalse(docID string, newVersion DocVersion, pare
 
 	rt.WaitForPendingChanges()
 
-	return DocVersionFromPutResponse(rt.TB(), resp)
+	return base.Ptr(DocVersionFromPutResponse(rt.TB(), resp))
 }
 
 func (rt *RestTester) RequireWaitChanges(numChangesExpected int, since string) ChangesResults {
--- a/rest/api_test_helpers.go
+++ b/rest/api_test_helpers.go
@@ -60,7 +60,6 @@ func (rt *RestTester) PutNewEditsFalse(docID string, newVersion DocVersion, pare
 		_, parentDigest := db.ParseRevID(base.TestCtx(rt.TB()), parentVersionCopy.RevID)
 		ids = append(ids, parentDigest)
 	}
-	// TODO: Revs invalid???
 	revisions["ids"] = ids
 
 	requestBody[db.BodyRevisions] = revisions
--- a/rest/attachment_test.go
+++ b/rest/attachment_test.go
@@ -2406,10 +2406,9 @@ func TestMinRevPosWorkToAvoidUnnecessaryProveAttachment(t *testing.T) {
 		proveAttachmentAfter := btc.pushReplication.replicationStats.ProveAttachment.Value()
 		assert.Equal(t, proveAttachmentBefore, proveAttachmentAfter)
 
-		// start another push to run in the background since the last doc version
-		doc, ok := btcRunner.SingleCollection(btc.id).getClientDoc(docID)
-		require.True(t, ok)
-		btcRunner.StartPushWithOpts(btc.id, BlipTesterPushOptions{Continuous: true, Since: strconv.Itoa(int(doc.latestSeq))})
+		// start another push to run in the background from where we last left off
+		latestSeq := btcRunner.SingleCollection(btc.id).lastSeq()
+		btcRunner.StartPushWithOpts(btc.id, BlipTesterPushOptions{Continuous: true, Since: strconv.Itoa(int(latestSeq))})
 
 		// Push another bunch of history, this time whilst a replicator is actively pushing them
 		for i := 25; i < 50; i++ {
--- a/rest/attachment_test.go
+++ b/rest/attachment_test.go
@@ -720,7 +720,7 @@ func TestConflictWithInvalidAttachment(t *testing.T) {
 	require.NoError(t, base.JSONUnmarshal(response.Body.Bytes(), &body))
 
 	// Modify Doc
-	parentRevList := [3]string{"foo3", "foo2", version.Digest()}
+	parentRevList := [3]string{"foo3", "foo2", version.RevIDDigest()}
 	body["_rev"] = "3-foo3"
 	body["rev"] = "3-foo3"
 	body["_revisions"].(map[string]interface{})["ids"] = parentRevList
@@ -786,13 +786,13 @@ func TestConflictingBranchAttachments(t *testing.T) {
 
 	// //Create diverging tree
 
-	reqBodyRev2 := `{"_rev": "2-two", "_revisions": {"ids": ["two", "` + version.Digest() + `"], "start": 2}}`
+	reqBodyRev2 := `{"_rev": "2-two", "_revisions": {"ids": ["two", "` + version.RevIDDigest() + `"], "start": 2}}`
 	response := rt.SendAdminRequest("PUT", "/{{.keyspace}}/doc1?new_edits=false", reqBodyRev2)
 	RequireStatus(t, response, http.StatusCreated)
 
 	docVersion2 := DocVersionFromPutResponse(t, response)
 
-	reqBodyRev2a := `{"_rev": "2-two", "_revisions": {"ids": ["twoa", "` + version.Digest() + `"], "start": 2}}`
+	reqBodyRev2a := `{"_rev": "2-two", "_revisions": {"ids": ["twoa", "` + version.RevIDDigest() + `"], "start": 2}}`
 	response = rt.SendAdminRequest("PUT", "/{{.keyspace}}/doc1?new_edits=false", reqBodyRev2a)
 	RequireStatus(t, response, http.StatusCreated)
 	docVersion2a := DocVersionFromPutResponse(t, response)
@@ -2104,10 +2104,10 @@ func TestAttachmentRemovalWithConflicts(t *testing.T) {
 	losingVersion3 := rt.UpdateDoc(docID, version, `{"_attachments": {"hello.txt": {"revpos":2,"stub":true,"digest":"sha1-Kq5sNclPz7QV2+lfQIuc6R7oRu0="}}}`)
 
 	// Create doc conflicting with previous revid referencing previous attachment too
-	winningVersion3 := rt.PutNewEditsFalse(docID, NewDocVersionFromFakeRev("3-b"), version, `{"_attachments": {"hello.txt": {"revpos":2,"stub":true,"digest":"sha1-Kq5sNclPz7QV2+lfQIuc6R7oRu0="}}, "Winning Rev": true}`)
+	winningVersion3 := rt.PutNewEditsFalse(docID, NewDocVersionFromFakeRev("3-b"), &version, `{"_attachments": {"hello.txt": {"revpos":2,"stub":true,"digest":"sha1-Kq5sNclPz7QV2+lfQIuc6R7oRu0="}}, "Winning Rev": true}`)
 
 	// Update the winning rev 3 and ensure attachment remains around as the other leaf still references this attachment
-	finalVersion4 := rt.UpdateDoc(docID, winningVersion3, `{"update": 2}`)
+	finalVersion4 := rt.UpdateDoc(docID, *winningVersion3, `{"update": 2}`)
 
 	type docResp struct {
 		Attachments db.AttachmentsMeta `json:"_attachments"`
@@ -2158,7 +2158,7 @@ func TestAttachmentsMissing(t *testing.T) {
 
 	version2 := rt.UpdateDoc(docID, version1, `{"_attachments": {"hello.txt": {"revpos":1,"stub":true,"digest":"sha1-Kq5sNclPz7QV2+lfQIuc6R7oRu0="}}, "testval": ["xxx","xxx"]}`)
 
-	_ = rt.PutNewEditsFalse(docID, NewDocVersionFromFakeRev("2-b"), version1, `{"_rev": "2-b", "_revisions": {"ids": ["b", "ca9ad22802b66f662ff171f226211d5c"], "start": 2}, "Winning Rev": true}`)
+	_ = rt.PutNewEditsFalse(docID, NewDocVersionFromFakeRev("2-b"), &version1, `{"_rev": "2-b", "_revisions": {"ids": ["b", "ca9ad22802b66f662ff171f226211d5c"], "start": 2}, "Winning Rev": true}`)
 
 	rt.GetDatabase().FlushRevisionCacheForTest()
 
@@ -2177,7 +2177,7 @@ func TestAttachmentsMissingNoBody(t *testing.T) {
 
 	version2 := rt.UpdateDoc(docID, version1, `{"_attachments": {"hello.txt": {"revpos":1,"stub":true,"digest":"sha1-Kq5sNclPz7QV2+lfQIuc6R7oRu0="}}}`)
 
-	_ = rt.PutNewEditsFalse(docID, NewDocVersionFromFakeRev("2-b"), version1, `{}`)
+	_ = rt.PutNewEditsFalse(docID, NewDocVersionFromFakeRev("2-b"), &version1, `{}`)
 
 	rt.GetDatabase().FlushRevisionCacheForTest()
 
@@ -2254,6 +2254,8 @@ func TestAttachmentDeleteOnExpiry(t *testing.T) {
 
 }
 func TestUpdateExistingAttachment(t *testing.T) {
+	base.SetUpTestLogging(t, base.LevelTrace, base.KeyCRUD, base.KeySync, base.KeySyncMsg)
+
 	rtConfig := &RestTesterConfig{
 		GuestEnabled: true,
 	}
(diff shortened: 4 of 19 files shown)


rest/blip_api_crud_test.go
```diff
@@ -3143,7 +3143,8 @@ func TestOnDemandImportBlipFailure(t *testing.T) {
 				btcRunner.WaitForDoc(btc2.id, markerDoc)
 
 				// Validate that the latest client message for the requested doc/rev was a norev
-				msg, ok := btcRunner.SingleCollection(btc2.id).GetBlipRevMessage(docID, revID.RevID)
+				// FIXME: Norev support
```
bbrks: done - comment needs removing

rest/blip_api_delta_sync_test.go
```diff
@@ -811,19 +818,21 @@ func TestBlipDeltaSyncPush(t *testing.T) {
...
 		assert.Equal(t, `{"greetings":[{"hello":"world!"},{"hi":"alice"}]}`, string(data))
 		// create doc1 rev 2-abc on client
-		newRev, err := btcRunner.PushRev(client.id, docID, version, []byte(`{"greetings":[{"hello":"world!"},{"hi":"alice"},{"howdy":"bob"}]}`))
+		newRev, err := btcRunner.AddRev(client.id, docID, &version, []byte(`{"greetings":[{"hello":"world!"},{"hi":"alice"},{"howdy":"bob"}]}`))
 		assert.NoError(t, err)
 
 		// Check EE is delta, and CE is full-body replication
 		msg := client.waitForReplicationMessage(collection, 2)
 
+		// FIXME: Delta sync support for push replication
```
bbrks: done - comment needs removing

rest/blip_api_delta_sync_test.go
```diff
@@ -856,6 +865,7 @@ func TestBlipDeltaSyncPush(t *testing.T) {
 		assert.Equal(t, map[string]interface{}{"howdy": "bob"}, greetings[2])
 
 		// tombstone doc1 (gets rev 3-f3be6c85e0362153005dae6f08fc68bb)
+		// FIXME: Not replicated to client?
```
bbrks: done - comment needs removing

rest/blip_client_test.go
```diff
@@ -750,6 +968,183 @@ func (btcRunner *BlipTestClientRunner) Collection(clientID uint32, collectionNam
...
+						var status int
+						if i >= len(response) {
+							// trailing zeros are removed - treat as 0 from now on
+							status = 0
+						} else {
+							status = response[i]
+						}
+						switch status {
+						case 0:
+							// FIXME: `change.activeRev()` could change if the test is writing concurrently. We'll have to store the specific version we sent in proposeChanges and iterate on that.
```
bbrks: this may be a source of flakiness but not sure it will actually surface itself unless there's a test that is writing a lot of revs on the same doc whilst pushing

base/collection_xattr.go
```diff
@@ -265,7 +265,7 @@ func (c *Collection) subdocGetBodyAndXattrs(ctx context.Context, k string, xattr
 				xattrs[xattrKey] = xattr
 			}
 
-			if len(xattrErrors) == len(xattrs) {
+			if len(xattrErrors) == len(xattrKeys) {
```
bbrks: unrelated - @torcolvin I think I mentioned this to you the other day, would like your opinions on whether this needs fixing in another PR

db/crud.go
```diff
@@ -2189,8 +2189,15 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
...
+	// If this update branched the revision tree, make a note of it in the 'Stored doc' log.
+	// The next time we'll see this is when the changes feed sends this revision.
+	inConflictLogSuffix := ""
+	if inConflict {
+		inConflictLogSuffix = " (branched)"
+	}
+
 	// Now that the document has successfully been stored, we can make other db changes:
-	base.DebugfCtx(ctx, base.KeyCRUD, "Stored doc %q / %q as #%v", base.UD(docid), newRevID, doc.Sequence)
+	base.DebugfCtx(ctx, base.KeyCRUD, "Stored doc %q / %q%s as #%v", base.UD(docid), newRevID, inConflictLogSuffix, doc.Sequence)
```
bbrks: unrelated - file new ticket/move to another PR

rest/api_test_helpers.go
```diff
@@ -55,10 +55,12 @@ func (rt *RestTester) PutNewEditsFalse(docID string, newVersion DocVersion, pare
...
 	revisions["start"] = newRevGeneration
 	ids := []string{newRevDigest}
-	if parentVersion.RevID != "" {
-		_, parentDigest := db.ParseRevID(base.TestCtx(rt.TB()), parentVersion.RevID)
+	if parentVersion != nil {
+		parentVersionCopy := *parentVersion
+		_, parentDigest := db.ParseRevID(base.TestCtx(rt.TB()), parentVersionCopy.RevID)
 		ids = append(ids, parentDigest)
 	}
+	// TODO: Revs invalid???
```
bbrks: fixed - remove comment

base/collection_xattr.go
torcolvin: I think we could safely remove this, but we don't have coverage for all these lines anyway.

At least the comment is wrong here, because it would hit `gocbcore.ErrMemdSubDocBadMulti`. I did however try to write tests and fix this in https://github.com/couchbase/sync_gateway/commit/94169b373169c73f3674434a2ba2c3608790a449. We could file a ticket to fix this but I honestly don't think it will be worth prioritizing.

rest/api_test_helpers.go
```diff
@@ -69,7 +70,7 @@ func (rt *RestTester) PutNewEditsFalse(docID string, newVersion DocVersion, pare
 
 	rt.WaitForPendingChanges()
 
-	return DocVersionFromPutResponse(rt.TB(), resp)
+	return base.Ptr(DocVersionFromPutResponse(rt.TB(), resp))
```
torcolvin, line 73: Is there a reason to return this as a Pointer? I can see passing this in a ptr perhaps, but maybe it is even more explicit to have a DocVersion that is something like "NoParentDocVersion" - that was the intent behind `EmptyDocVersion` which is mostly passed everywhere.

I think fixing this could be out of scope, but I am just curious if there's a good reason for it.

rest/attachment_test.go
```diff
@@ -2272,38 +2272,41 @@ func TestUpdateExistingAttachment(t *testing.T) {
...
 
+		btcRunner.StartPull(btc.id)
+		btcRunner.StartPush(btc.id)
+
 		doc1Version := rt.PutDoc(doc1ID, `{}`)
 		doc2Version := rt.PutDoc(doc2ID, `{}`)
-
 		rt.WaitForPendingChanges()
-		btcRunner.StartOneshotPull(btc.id)
+
```
torcolvin, line 2281: For this test, this doesn't matter - but is there a reason to change from a one shot pull to a continuous pull? In this test, we are doing a pull before push so that documents with the revs exist on the blip tester client.

The reason I think this never matters is that there is only one blip tester client and so it none of the changes will be replicated via the pull.

OTOH, if there was a single shot pull replication that would be a simpler test to debug if you are looking at logging, because the pull would be done before a push ever happened.

If you do decide to leave the changes as is (which I'm okay with), the `rt.WaitForPendingChanges` is unnecessary since that was need after `PutDoc` but before the one shot pull happened so that the latest sequence for pull would be up to date.

rest/attachment_test.go
```diff
@@ -2272,38 +2272,41 @@ func TestUpdateExistingAttachment(t *testing.T) {
...
+		var err error
+		doc1Version, err = btcRunner.AddRev(btc.id, doc1ID, &doc1Version, []byte(`{"key": "val", "_attachments": {"attachment": {"data": "`+attachmentAData+`"}}}`))
 		require.NoError(t, err)
-		doc2Version, err = btcRunner.PushRev(btc.id, doc2ID, doc2Version, []byte(`{"key": "val", "_attachments": {"attachment": {"data": "`+attachmentBData+`"}}}`))
+		doc2Version, err = btcRunner.AddRev(btc.id, doc2ID, &doc2Version, []byte(`{"key": "val", "_attachments": {"attachment": {"data": "`+attachmentBData+`"}}}`))
 		require.NoError(t, err)
 
-		assert.NoError(t, rt.WaitForVersion(doc1ID, doc1Version))
-		assert.NoError(t, rt.WaitForVersion(doc2ID, doc2Version))
+		require.NoError(t, rt.WaitForVersion(doc1ID, doc1Version))
```
torcolvin, line 2294: This is out of scope for this PR but probably `WaitForVersion` doesn't need to check for an error, we are always going to want to stop test execution .

rest/attachment_test.go
```diff
@@ -2376,25 +2380,45 @@ func TestMinRevPosWorkToAvoidUnnecessaryProveAttachment(t *testing.T) {
...
 		initialVersion := btc.rt.PutDoc(docID, `{"_attachments": {"hello.txt": {"data": "aGVsbG8gd29ybGQ="}}}`)
 
 		// Replicate data to client and ensure doc arrives
 		btc.rt.WaitForPendingChanges()
-		btcRunner.StartOneshotPull(btc.id)
 		btcRunner.WaitForVersion(btc.id, docID, initialVersion)
 
-		// Push a revision with a bunch of history simulating doc updated on mobile device
-		// Note this references revpos 1 and therefore SGW has it - Shouldn't need proveAttachment
+		// Create a set of revisions before we start the replicator to ensure there's a significant amount of history to push
```
torcolvin, line 2393: Does this run into the revsLimit set on the replicator?

rest/blip_client_test.go
```diff
@@ -861,17 +1244,89 @@ func (btr *BlipTesterReplicator) sendMsg(msg *blip.Message) (err error) {
...
+
+	return &rev, nil
+}
+
+// AddRev creates a revision on the client.
 // The rev ID is always: "N-abc", where N is rev generation for predictability.
-func (btc *BlipTesterCollectionClient) PushRev(docID string, parentVersion DocVersion, body []byte) (DocVersion, error) {
-	revid, err := btc.PushRevWithHistory(docID, parentVersion.RevID, body, 1, 0)
-	return DocVersion{RevID: revid}, err
+func (btc *BlipTesterCollectionClient) AddRev(docID string, parentVersion *DocVersion, body []byte) (DocVersion, error) { // Inline attachment processing
```
torcolvin, line 1433: Potentially out of scope:

Each use of `AddRev` looks like it immediately requires err == nil after it.

rest/blip_api_attachment_test.go
```diff
@@ -297,35 +303,33 @@ func TestBlipPushPullNewAttachmentCommonAncestor(t *testing.T) {
 		btc := btcRunner.NewBlipTesterClientOptsWithRT(rt, opts)
 		defer btc.Close()
 
-		btcRunner.StartPull(btc.id)
+		btcRunner.StartPushWithOpts(btc.id, BlipTesterPushOptions{Continuous: false})
```
torcolvin: This code doesn't seem right in either the old code or the new code.

In the old code it is doing a pull replication, but there are no revs to ever be pulled as far as I can see.

In the new code, the push is with `Continuous: false` and that seems like it isn't going to push the code that `AddRev` calls.

rest/blip_api_attachment_test.go
```diff
@@ -297,35 +303,33 @@ func TestBlipPushPullNewAttachmentCommonAncestor(t *testing.T) {
...
 
-		// CBL creates revisions 1-abc,2-abc on the client, with an attachment associated with rev 2.
-		bodyText := `{"greetings":[{"hi":"alice"}],"_attachments":{"hello.txt":{"data":"aGVsbG8gd29ybGQ="}}}`
-		err := btcRunner.StoreRevOnClient(btc.id, docID, "2-abc", []byte(bodyText))
+		docVersion, err := btcRunner.AddRev(btc.id, docID, nil, []byte(`{"greetings":[{"hi": "alice"}]}`))
```
torcolvin, line 308: I think I understand why you changed the bodies here to be alice,bob,charlie,dave - to recognize each version, but just making sure I understand the intention behind this change.

rest/blip_client_test.go
```diff
@@ -50,8 +51,12 @@ type BlipTesterClientOpts struct {
 
 	// sendReplacementRevs opts into the replacement rev behaviour in the event that we do not find the requested one.
 	sendReplacementRevs bool
+
+	revsLimit *int // defaults to 20
 }
 
+const defaultBelipTesterClientRevsLimit = 20
```
torcolvin:
```suggestion
// defaultBlipTesterClientRevsLimit is the number of revisions stored by the blip test client, old revisions will be discarded
const defaultBlipTesterClientRevsLimit = 20
```

not sure the comment is right but belip is definitely a typo

rest/blip_client_test.go
```diff
@@ -65,19 +70,153 @@ type BlipTesterClient struct {
...
+		c.seqLock.Unlock()
+		c.TB().Logf("OneShotDocsSince: since=%d, _seqLast=%d - iterating", since, seqLast)
+		for seq := since; seq <= seqLast; seq++ {
+			doc, ok := c.getClientDocForSeq(seq)
+			// filter non-latest entries in cases where we haven't pruned _seqStore
+			if !ok {
+				continue
+			} else if doc.latestSeq != seq {
+				// this entry should be cleaned up from _seqStore?
+				base.AssertfCtx(context.TODO(), "seq %d found in _seqStore but latestSeq for doc %d - this should've been pruned out!", seq, doc.latestSeq)
```
torcolvin: I'd do ctx = base.TestCtx(c.TB()) here

rest/blip_client_test.go
```diff
@@ -65,19 +70,153 @@ type BlipTesterClient struct {
...
+		for {
+			c.TB().Logf("docsSince: sinceVal=%d", sinceVal)
+			for _, doc := range c.OneShotDocsSince(sinceVal) {
+				select {
+				case <-ctx.Done():
+					return
+				default:
+				}
+				c.TB().Logf("sending doc %q", doc.id)
+				ch <- doc
```
torcolvin: I think that this might be where you want to do an early return if the ctx is cancelled? otherwise you can use if ctx.Err() != nil in the select loop above

rest/utilities_testing_resttester.go
```diff
@@ -131,7 +131,11 @@ func (rt *RestTester) WaitForVersion(docID string, version DocVersion) error {
 		}
 		var body db.Body
 		require.NoError(rt.TB(), base.JSONUnmarshal(rawResponse.Body.Bytes(), &body))
-		return body.ExtractRev() == version.RevID
+		if body.ExtractRev() != version.RevID {
+			rt.TB().Logf("Retrying WaitForVersion for doc %s, expected rev %s, got body %s", docID, version.RevID, body)
```
torcolvin: This is probably going to end up being quite noisy for test, if you want to change this I'd probably use require.EvenutallyWithT and have the log messages on the assertion, so that you see the error on the last failing pass.

rest/utilities_testing.go
```diff
@@ -55,7 +56,7 @@ type RestTesterConfig struct {
 	ImportFilter                    string                      // put the import filter function source in here (optional)
 	DatabaseConfig                  *DatabaseConfig             // Supports additional config options.  BucketConfig, Name, Sync, Unsupported will be ignored (overridden)
 	MutateStartupConfig             func(config *StartupConfig) // Function to mutate the startup configuration before the server context gets created. This overrides options the RT sets.
-	InitSyncSeq                     uint64                      // If specified, initializes _sync:seq on bucket creation.  Not supported when running against walrus
+	InitSyncSeq                     uint64                      // If specified, initializes _sync:clientSeq on bucket creation.  Not supported when running against walrus
```
torcolvin: This change is not correct?

rest/blip_client_test.go
```diff
@@ -65,19 +70,153 @@ type BlipTesterClient struct {
...
+
+// OneShotDocsSince is an iterator that yields client sequence and document pairs that are newer than the given since value.
+func (c *BlipTesterCollectionClient) OneShotDocsSince(since clientSeq) iter.Seq2[clientSeq, *clientDoc] {
+	return func(yield func(clientSeq, *clientDoc) bool) {
+		c.seqLock.Lock()
+		seqLast := c._seqLast
+		for c._seqLast <= since {
+			// block until new seq
+			c.TB().Logf("OneShotDocsSince: since=%d, _seqLast=%d - waiting for new sequence", since, c._seqLast)
+			c._seqCond.Wait()
```
torcolvin, line 95: I'm worried about the case where a test assertion fails in the middle of the iteration, or a test times out? A sync.Cond will block forever, and we can't cancel with a context. For the changes feed we use the idiom of Broadcast() and then checking under wait to see if the context was cancelled.

I'm not sure this is a problem, but I am worried about a failing test failing to take down its rest tester, either as a data leak or failing to exit appropriately.

rest/blip_client_test.go
```diff
@@ -65,19 +70,153 @@ type BlipTesterClient struct {
...
+type clientSeq uint64
+
+// clientDocRev represents a revision of a document stored on this client, including any metadata associated with this specific revision.
+type clientDocRev struct {
+	clientSeq     clientSeq
+	version       DocVersion
+	parentVersion *DocVersion
+	body          []byte
+	isDelete      bool
+	isNoRev       bool
```
torcolvin: What is the difference between a document being a norev and a deletion? I don't think I understand this abstraction.

I don't necessarily understand the differences between a nil body, isDelete, isNoRev?

rest/blip_client_test.go
```diff
@@ -65,19 +70,153 @@ type BlipTesterClient struct {
...
+	message       *blip.Message // rev or norev message associated with this revision
```
torcolvin: I'm confused by the presence of the message property. If this was a pull replication, this is storing the rev messages. What's the advantage of storing the rev messages with the other information?

torcolvin (commented):
I only looked through part of this and mostly not the changes in blip_client_test.go but it would be helpful to me to have an overall design strategy possibly even in a comment for how document bodies and revisions are stored on the body for a pull replication and push replication.

It might make sense for me to meet and have you walk me through the design strategy a bit to make the review faster for me to do.

rest/utilities_testing.go
bbrks: oops, find/replace mistake :)

rest/attachment_test.go
bbrks: yes (intentionally so)

rest/blip_client_test.go
bbrks: `norev` is sent when a requested revID is not found at all - it indicates a doc was purged or expired (not deleted) on SG between the time at which the change entry was sent and the rev was requested, but that doesn't mean it is a delete that needs replicating to the client

rest/blip_client_test.go
bbrks: so tests can assert on the replication protocol behaviour specifically, rather than the eventual outcome (e.g. identifying whether this rev was sent as a delta, and introspecting what message properties it had)

Most tests use the `WaitForBlipRevMessage` helpers to pull this for a specific replicated doc version

rest/blip_client_test.go
bbrks: in the middle of fixing this race, since it's loosely related to other race detector failures

rest/blip_client_test.go
bbrks: Fair comment - was mostly relying on existing test code that had error returns
 There's only one test (`TestBlipPusherUpdateDatabase`) that expects a non-nil error which I guess could be handled separately.

rest/blip_client_test.go
bbrks: I think maybe calling Broadcast() on close and doing a ctx check to return would resolve this?  I'm not overly familiar with sync.Cond idioms

rest/blip_client_test.go
bbrks: Additionally delete is a flag on a message. There's no guarantee the body is nil AFAIK

rest/blip_client_test.go
bbrks: FYI, I still need to do this and the other context/close investigation we talked about.

rest/blip_client_test.go
```diff
@@ -65,19 +71,229 @@ type BlipTesterClient struct {
...
+			// block until new seq
+			c.TB().Logf("OneShotDocsSince: since=%d, _seqLast=%d - waiting for new sequence", since, c._seqLast)
+			c._seqCond.Wait()
+			// Check to see if we were woken because of Close()
+			if ctx.Err() != nil {
+				c.seqLock.Unlock()
+				return
+			}
+			seqLast = c._seqLast
+			c.TB().Logf("OneShotDocsSince: since=%d, _seqLast=%d - woke up", since, c._seqLast)
```
torcolvin: I'd probably drop the logging here, or turn this into optional debugging logging for the Blip Tester itself, I'm not sure that this is going to be meaningful for running tests in general.

I'm not going to comment on each `c.TB().Logf` line here.

rest/blip_client_test.go
```diff
@@ -65,19 +71,229 @@ type BlipTesterClient struct {
...
+		c.seqLock.Unlock()
+		c.TB().Logf("OneShotDocsSince: since=%d, _seqLast=%d - iterating", since, seqLast)
+		for seq := since; seq <= seqLast; seq++ {
+			doc, ok := c.getClientDocForSeq(seq)
+			// filter non-latest entries in cases where we haven't pruned _seqStore
+			if !ok {
+				continue
+			} else if latestDocSeq := doc.latestSeq(); latestDocSeq != seq {
+				// this entry should've been cleaned up from _seqStore
+				base.AssertfCtx(base.TestCtx(c.TB()), "seq %d found in _seqStore but latestSeq for doc %d - this should've been pruned out!", seq, latestDocSeq)
```
torcolvin: I have a preference for constructing these as:


```suggestion
				require.FailNow(c.TB(), fmt.Sprintf("seq %d found in _seqStore but latestSeq for doc %d - this should've been pruned out!", seq, latestDocSeq))
```

Since panicking the test harness should be avoided. The same comment applies to lower lines of code.

rest/blip_client_test.go
```diff
@@ -65,19 +71,229 @@ type BlipTesterClient struct {
...
+					return
+				case ch <- doc:
+					c.TB().Logf("sent doc %q to changes feed", doc.id)
+					sinceVal = doc.latestSeq()
+				}
+			}
+			if !continuous {
+				c.TB().Logf("opts.Continuous=false, breaking changes loop")
+				break
+			}
```
torcolvin: readability nit here:

```suggestion
			defer close(ch)
			for _, doc := range c.OneShotDocsSince(ctx, sinceVal) {
				select {
				case <-ctx.Done():
					return
				case ch <- doc:
					sinceVal = doc.latestSeq()
				}
			}
			if !continuous {
				break
			}
```

Without this change, I don't understand how the loop closes in the non continuous case.

rest/blip_client_test.go
```diff
@@ -750,6 +1032,228 @@ func (btcRunner *BlipTestClientRunner) Collection(clientID uint32, collectionNam
...
+	sinceFromStr, err := db.ParsePlainSequenceID(opts.Since)
+	require.NoError(btcc.TB(), err)
+	seq := clientSeq(sinceFromStr.SafeSequence())
+	btcc.goroutineWg.Add(1)
+	go func() {
+		defer btcc.goroutineWg.Done()
+		for {
+			if btcc.ctx.Err() != nil {
+				return
+			}
```
torcolvin: I'd probably drop this context check and keep the one inside `docSince` function and the one in the `docSince` loop. But this is just readability.

rest/blip_client_test.go
```diff
@@ -65,19 +71,229 @@ type BlipTesterClient struct {
...
+// Can be used for lookups in clientDoc.revisionsBySeq
+func (cd *clientDoc) docRevSeqsNewestToOldest() []clientSeq {
+	cd.lock.RLock()
+	defer cd.lock.RUnlock()
+	return cd._docRevSeqsNewestToOldest()
+}
+
+func (cd *clientDoc) _docRevSeqsNewestToOldest() []clientSeq {
+	cd.lock.RLock()
+	defer cd.lock.RUnlock()
```
torcolvin: I assume this is a mistake since `docRevSeqsNewestToOldest` is also called with this lock.

torcolvin (commented):
I think the code generally looks fine. I have some readability improvements I've made to avoid so many `ctx.Err() != nil` checks which I think were made defensively, and split up some of the complexity in `docsSince`. As discussed offline, this code might not live forever anyway and get replaced by active replicator code.

I did spend a considerable amount of time debugging why some of the blip delta sync tests fail on my local machine, only when running multiple tests at once. I never got to the bottom of why they failed and I am not sure that they were totally successfully passing before either. Running these on a different system leads to passing. That said, the delta sync tests are not run through github actions, so they might fail.

I will push up a branch (on top of this one) where you can look at the changes I've made to help debug, and decide to keep or remove them.

- remove AssertfCtx / err requirements so that there is a better traceback in the case of failure. In the case of `PushUnsolicitedRev` we need more information from the error message, but generally we just want a blip message to fail the test harness, preferably without passing the error back up. This results in a lot of code changes.
- removes most of the logging
- removes channel close from docSince since this is unused
- remove very confusing `UnsubPushChanges` function.
- Added some docstrings to try to understand the code better.

I am fine if you don't want this branch.

I think it is worth considering whether some of the assertion handling to push into testify require assertions with tracebacks will make it easier to debug the code on anemone, and if we want to do a painful merge once. Maybe evaluate how the merge/rebase of this PR will be onto anemone and whether significant changes were made there.

rest/blip_client_test.go
bbrks: It closes from the context when that is cancelled, but otherwise doesn't.

The loop will sit waiting in the `OneShotDocsSince` iterator until a new sequence is available on the client via the `c._seqCond.Wait()`

torcolvin (approved):

PR #7228, opened 2024-12-04
CBG-4389: extract cv from known revs according to spec

gregns1:
CBG-4389

In logs from CBL I noticed the following logging:
```
2024-11-28T23:39:55.204Z [DBG] Sync+: c:[453971bf] db:travel col:hotels Sending rev "<ud>hotel_1</ud>" 3-88f59fdda5466dbe3e2b98db4efac4b6 as delta. DeltaSrc:180c463ead590000@JpwZW6meh8irNPB2yMnTMA; 18080b8abea70000@rc2hc9uJRjeUpkduy5kYtQ
```
Leading to the following error when getting rev cache miss for this version and thus attempting a load from the bucket:
```
2024-11-28T23:39:55.204Z [DBG] CRUD+: c:[453971bf] db:travel col:hotels mismatch between specified current version and fetched document current version for doc <ud>hotel_1</ud>
```
DeltaSrc seems to be a CV and PV pair. Currently for known revs we take the first entry as the 'rev', it seems CBL is sending the full version vector in this array entry, thus we need to extract the cv from this full version vector and use that as the known rev for the document. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2838/


db/blip_sync_context.go
```diff
@@ -357,7 +357,16 @@ func (bsc *BlipSyncContext) handleChangesResponse(ctx context.Context, sender *b
 			// The first element of the knownRevsArray returned from CBL is the parent revision to use as deltaSrc
```
torcolvin, line 357: Is this comment wrong for HLVs?

torcolvin (commented):
Can you write an example test with a blip body so we can compare this?

db/blip_sync_context.go
```diff
@@ -357,7 +357,16 @@ func (bsc *BlipSyncContext) handleChangesResponse(ctx context.Context, sender *b
...
 			if bsc.useDeltas && len(knownRevsArray) > 0 {
 				if revID, ok := knownRevsArray[0].(string); ok {
-					deltaSrcRevID = revID
+					if bsc.useHLV() {
+						msgHLV, err := extractHLVFromBlipMessage(revID)
+						if err != nil {
+							base.ErrorfCtx(ctx, "Invalid known rev format for hlv")
```
torcolvin, line 363: This probably needs to say the docid, and the rev id body for debugging

PR #7229, opened 2024-12-04
CBG-4366 enable resurrection tests

torcolvin:
Switch tests to use fewer buckets: for each topology use use one bucket and then each actor will run tests in sequence. Using subtests allows to run each test individually if necessary.

Need to update tb on each peer for each subtest to avoid `subtest may have called FailNow on a parent test` error.

The integration tests can still fail but they don't seem to fail more frequently than the old code. I haven't seen failures in jenkins, but I might not have run them frequently enough.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2839/

### DIFF
--- a/rest/utilities_testing_resttester.go
+++ b/rest/utilities_testing_resttester.go
@@ -36,6 +36,11 @@ func (rt *RestTester) Run(name string, test func(*testing.T)) {
 	})
 }
 
+func (rt *RestTester) UpdateTB(t *testing.T) {
+	var tb testing.TB = t
+	rt.testingTB.Store(&tb)
+}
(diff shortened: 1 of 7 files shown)


gregns1 (approved):
LGTM, I ran tests in goland and it feels much nicer having each subtest under each topology. 
I was trying to think of a way to integrate the updatePeersT call into the test so whoever is writing a test wouldn't have to remember to do it (I would totally forget) but I'm not sure there is an easy way to do that so I'm good with it as is! 😄

PR #7237, opened 2024-12-09
CBG-4389: extract cv from known revs and store backup rev by cv

gregns1:
CBG-4389

- Extract cv from known revs 
- Disable delta sync for revTree clients 
- Switch to backup rev by cv, using hash of cv for concerns about key length
- Move backup for current rev to post update hlv function as due to macro expansion we don't have a version for cv at the backup rev function

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2862/

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -1084,9 +1084,6 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 	if deltaSrcRevID, isDelta := revMessage.DeltaSrc(); isDelta && !revMessage.Deleted() {
 		if !bh.sgCanUseDeltas {
 			return base.HTTPErrorf(http.StatusBadRequest, "Deltas are disabled for this peer")
-		} else if !bh.useHLV() {
-			// Disable delta sync for protocol versions < 4, CBG-3748 (backwards compatibility for revID delta sync)
-			return base.HTTPErrorf(http.StatusBadRequest, "backwards compatibility for revTree deltas not yet implemented")
 		}
 
 		//  TODO: Doing a GetRevCopy here duplicates some rev cache retrieval effort, since deltaRevSrc is always
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -772,7 +772,7 @@ func (bh *blipHandler) handleChanges(rq *blip.Message) error {
 	}
 	output.Write([]byte("]"))
 	response := rq.Response()
-	if bh.sgCanUseDeltas {
+	if bh.sgCanUseDeltas && bh.useHLV() {
 		base.DebugfCtx(bh.loggingCtx, base.KeyAll, "Setting deltas=true property on handleChanges response")
 		response.Properties[ChangesResponseDeltas] = trueProperty
 		bh.replicationStats.HandleChangesDeltaRequestedCount.Add(int64(nRequested))
@@ -865,7 +865,7 @@ func (bh *blipHandler) handleProposeChanges(rq *blip.Message) error {
 	}
 	output.Write([]byte("]"))
 	response := rq.Response()
-	if bh.sgCanUseDeltas {
+	if bh.sgCanUseDeltas && bh.useHLV() {
 		base.DebugfCtx(bh.loggingCtx, base.KeyAll, "Setting deltas=true property on proposeChanges response")
 		response.Properties[ChangesResponseDeltas] = trueProperty
 	}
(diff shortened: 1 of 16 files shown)


db/blip_handler.go
```diff
@@ -1082,6 +1084,9 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 	if deltaSrcRevID, isDelta := revMessage.DeltaSrc(); isDelta && !revMessage.Deleted() {
 		if !bh.sgCanUseDeltas {
 			return base.HTTPErrorf(http.StatusBadRequest, "Deltas are disabled for this peer")
+		} else if !bh.useHLV() {
```
adamcfraser: For an incoming delta that has a revTreeID as deltaSrc, we should be able to accept if when deltaSrc is the current revTreeID.  That case shouldn't require any temporary revision bodies.

db/blip_sync_context.go
```diff
@@ -354,10 +354,20 @@ func (bsc *BlipSyncContext) handleChangesResponse(ctx context.Context, sender *b
 				knownRevsByDoc[docID] = knownRevs
 			}
 
-			// The first element of the knownRevsArray returned from CBL is the parent revision to use as deltaSrc
+			// The first element of the knownRevsArray returned from CBL is the parent revision to use as deltaSrc for
+			// revtree clients, for HLV clients the first element is the HLV
```
adamcfraser:
```suggestion
			// For HLV clients, use the cv as deltaSrc
```

db/blip_sync_context.go
```diff
@@ -354,10 +354,20 @@ func (bsc *BlipSyncContext) handleChangesResponse(ctx context.Context, sender *b
...
 			if bsc.useDeltas && len(knownRevsArray) > 0 {
 				if revID, ok := knownRevsArray[0].(string); ok {
-					deltaSrcRevID = revID
+					if bsc.useHLV() {
+						msgHLV, err := extractHLVFromBlipMessage(revID)
+						if err != nil {
+							base.ErrorfCtx(ctx, "Invalid known rev format for hlv on doc: %s", docID)
```
adamcfraser: If we can't parse this, should we fall back to sending full body instead of erroring?  Or will it cause problems elsewhere?

adamcfraser: On a related note, does the loop in line 374 also need to handle HLV format, or does that happen elsewhere?

db/crud.go
```diff
@@ -879,27 +879,21 @@ func (db *DatabaseCollectionWithUser) getAvailableRevAttachments(ctx context.Con
...
-	}
 
 	// Find an ancestor that still has JSON in the document:
 	var json []byte
 	ancestorRevId := newDoc.RevID
 	for {
 		if ancestorRevId = doc.History.getParent(ancestorRevId); ancestorRevId == "" {
-			// No ancestors with JSON found.  Check if we need to back up current rev for delta sync, then return
-			db.backupRevisionJSON(ctx, doc.ID, newDoc.RevID, "", newBodyBytes, nil, doc.Attachments)
+			// No ancestors with JSON found. Return early
```
adamcfraser, line 888: Why wouldn't we want to back up the current rev (as cv) in this case as well?

adamcfraser (changes requested):
A few questions/comments.

db/crud.go
gregns1: Had to move the backing up of the current version of the doc to the function postWriteUpdateHLV as in this area here we don't actually have a cv to work with as we haven't macro expanded the cas into cv yet.

db/blip_handler.go
gregns1: Good point I think as long as the delta source rev is in the rev cache its fine, I think it makes sense for me to remove this upon reflection. If we receive a rev delta source, look in rev cache for it and fail to find it (then fail to load as no backup by rev id), the function will error anyway on line 1110.

db/blip_sync_context.go
gregns1: Good spot! Have updated to reflect this, let me know what you think!

db/crud.go
```diff
@@ -2611,6 +2605,24 @@ func postWriteUpdateHLV(doc *Document, casOut uint64) *Document {
 	if doc.HLV.CurrentVersionCAS == expandMacroCASValueUint64 {
 		doc.HLV.CurrentVersionCAS = casOut
 	}
+	// backup new revision to the bucket now we have a doc assigned a CV (post macro expansion) for delta generation purposes
+	backupRev := db.deltaSyncEnabled() && db.deltaSyncRevMaxAgeSeconds() != 0
+	if db.UseXattrs() && backupRev {
```
adamcfraser, line 2610: I think the UseXattrs() check is unnecessary here (we don't support HLV with non-xattrs)

adamcfraser (approved):
LGTM

PR #7239, opened 2024-12-11
CBG-4331: legacy rev handling for version 4 replication protocol

gregns1:
CBG-4331

- Add check for vector rev in propose changes handling, if no vector present fall into legacy rev handling 
- When failing to parse hlv out of known revs AND we're in protocol version 4 then we can assume we have received a legacy rev, then i'm passing down boolean into sendRevision to build out the history for a legacy rev message 
- Handling for legacy rev messages in processRev when in protocol version 4.
- Test coverage using blip tester, adding custom handlers when necessary to force a legacy rev response
- [CBG-4409](https://jira.issues.couchbase.com/browse/CBG-4409) will be the ticket to add new BlipTesterClient tests for each test case in the document

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2868/

[CBG-4409]: https://couchbasecloud.atlassian.net/browse/CBG-4409?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/db/blip_sync_context.go
+++ b/db/blip_sync_context.go
@@ -363,7 +363,7 @@ func (bsc *BlipSyncContext) handleChangesResponse(ctx context.Context, sender *b
 					if versionVectorProtocol {
 						msgHLV, err := extractHLVFromBlipMessage(revID)
 						if err != nil {
-							base.DebugfCtx(ctx, base.KeySync, "Invalid known rev format for hlv on doc: %s falling back to full body replication.", base.MD(docID))
+							base.DebugfCtx(ctx, base.KeySync, "Invalid known rev format for hlv on doc: %s falling back to full body replication.", base.UD(docID))
 							deltaSrcRevID = "" // will force falling back to full body replication below
 						} else {
 							deltaSrcRevID = msgHLV.GetCurrentVersionString()
--- a/db/blip_sync_context.go
+++ b/db/blip_sync_context.go
@@ -363,7 +363,7 @@ func (bsc *BlipSyncContext) handleChangesResponse(ctx context.Context, sender *b
 					if versionVectorProtocol {
 						msgHLV, err := extractHLVFromBlipMessage(revID)
 						if err != nil {
-							base.DebugfCtx(ctx, base.KeySync, "Invalid known rev format for hlv on doc: %s falling back to full body replication.", docID)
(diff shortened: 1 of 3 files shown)


rest/blip_legacy_revid_test.go
```diff
@@ -0,0 +1,425 @@
...
+	err = base.JSONUnmarshal(body, &changeList)
+	require.NoError(t, err)
+
+	assert.Len(t, changeList, 0)
+}
+
+// TestProposeChangesHandlingWithExistingRevs:
+//   - Build up propose changes request for conflicting and non conflicting docs with legacy revs
+//   - Assert that the response sent from SGW is as expected
+func TestProposeChangesHandlingWithExistingRevs(t *testing.T) {
```
adamcfraser, line 67: We don't have a case here where the current rev on SGW matches the legacy rev (where we'd expect to return ProposedRev_Exists).  Is that challenging to test today?  For the purposes of testing it would be probably be ok if the SGW write has an HLV, if that's needed (i.e. the legacy rev is matching the revTreeID generated for a SGW 4.0 update).

adamcfraser (changes requested):
Changes look good - made a comment about one use case that I thought the tests could include.

db/blip_sync_context.go
```diff
@@ -363,7 +363,7 @@ func (bsc *BlipSyncContext) handleChangesResponse(ctx context.Context, sender *b
 					if versionVectorProtocol {
 						msgHLV, err := extractHLVFromBlipMessage(revID)
 						if err != nil {
-							base.DebugfCtx(ctx, base.KeySync, "Invalid known rev format for hlv on doc: %s falling back to full body replication.", docID)
+							base.DebugfCtx(ctx, base.KeySync, "Invalid known rev format for hlv on doc: %s falling back to full body replication.", base.MD(docID))
```
adamcfraser: doc IDs are user data, and should get UD()

adamcfraser (dismissed):

bbrks (approved):

PR #7245, opened 2024-12-16
CBG-4420: handle rev tree in history on processRev

gregns1:
CBG-4420

- Add code to pull rev tree from history if it exists
- Add conflict check fro rev tree history if that exists on rev message 
- Add tests for conflict and non conflict scenarios

## Pre-review checklist


db/blip_handler.go
```diff
@@ -1080,13 +1081,23 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
...
 		newDoc.HLV = incomingHLV
 	}
+	if legacyRevPresent {
+		// split out rev tree history here into separate list
+		if historyStr != "" {
+			history = append(history, strings.Split(historyStr, ",")...)
+		}
+		if len(incomingHLV.PreviousVersions) > 0 || len(incomingHLV.MergeVersions) > 0 {
+			// we have hlv entries at start of history string, remove this as we have already parsed these values above
+			history = history[1:]
```
adamcfraser: In the case where PV has multiple entries, I believe there can be more than one comma-delimited HLV entry in the history before reaching legacy revTreeIDs.  Would it make more sense to have extractHLVFromBlipMessage return the actual list of legacyRevIDs since it's already doing that parsing work?

rest/blip_legacy_revid_test.go
```diff
@@ -223,6 +223,123 @@ func TestProcessLegacyRev(t *testing.T) {
...
+	// remove hlv here to simulate a legacy rev
+	require.NoError(t, ds.RemoveXattrs(ctx, docID2, []string{base.VvXattrName}, docVersion.CV.Value))
+	rt.GetDatabase().FlushRevisionCacheForTest()
+
+	// Have CBL send an update to that doc, with history in HLV + revTreeID format
+	history = []string{"1000@CBL2", rev1ID}
+	sent, _, _, err = bt.SendRevWithHistory("doc2", "1001@CBL1", history, []byte(`{"some": "update"}`), blip.Properties{})
+	assert.True(t, sent)
+	require.NoError(t, err)
+
```
adamcfraser, line 285: There's a non-conflict case here that I don't think is implemented yet:
```
// 3. CBL sends rev=1010@CBL1, history=1000@CBL2,2-abc,1-abc when SGW has current rev 1-abc (document underwent multiple legacy and p2p updates before being pushed to SGW)
```
In this case 2-abc should be getting added to the revtree history, _in addition to_ the new rev that gets auto-generated for `1010@CBL1`.

rest/blip_legacy_revid_test.go
```diff
@@ -223,6 +223,123 @@ func TestProcessLegacyRev(t *testing.T) {
...
+	docVersion = rt.PutDocDirectly(docID2, db.Body{"test": "doc"})
+	rev1ID = docVersion.RevTreeID
+
+	docVersion = rt.UpdateDocDirectly(docID2, docVersion, db.Body{"some": "update"})
+	rev2ID = docVersion.RevTreeID
+
+	docVersion = rt.UpdateDocDirectly(docID2, docVersion, db.Body{"some": "update2"})
+
+	history = []string{rev2ID, rev1ID}
+	sent, _, _, err = bt.SendRevWithHistory(docID, "3-abc", history, []byte(`{"key": "val"}`), blip.Properties{})
```
adamcfraser: I think this is supposed to be using docID2.

rest/blip_legacy_revid_test.go
```diff
@@ -223,6 +223,123 @@ func TestProcessLegacyRev(t *testing.T) {
...
+
+	docVersion = rt.UpdateDocDirectly(docID, docVersion, db.Body{"some": "update2"})
+
+	// remove hlv here to simulate a legacy rev
+	require.NoError(t, ds.RemoveXattrs(base.TestCtx(t), docID, []string{base.VvXattrName}, docVersion.CV.Value))
+	rt.GetDatabase().FlushRevisionCacheForTest()
+
+	// Test case: same as above but not having the rev be legacy on SGW side (don't remove the hlv)
+	history := []string{rev2ID, rev1ID}
+	sent, _, _, err := bt.SendRevWithHistory(docID, "3-abc", history, []byte(`{"key": "val"}`), blip.Properties{})
```
adamcfraser, line 366: Can we include conflict cases where the client is sending a mix of HLV and legacy revID.  e.g.:

Case 1:
- SGW has 2-abc
- CBL sends rev=1010@CBL1, history=1000@CBL2,1-abc

Case 2:
- SGW has 1000@CBL2, 2-abc
- CBL sends rev=1010@CBL1, history=2-abc

rest/blip_legacy_revid_test.go
adamcfraser: I think there's a similar case for create, like:
```
// 4. CBL sends rev=1010@CBL1, history=1-abc when SGW does not have the doc (document underwent multiple legacy and p2p updates before being pushed to SGW)
```
Here we'd expect the resultant rev to be 2-something (the autogenerated rev tree ID for the HLV update), but have 1-abc as it's parent in the revtree.

db/blip_handler.go
gregns1: Good point sorry i missed this.

PR #7251, opened 2024-12-19
CBG-4432: Implement Delete on the BlipTesterClient and enable delete-related topology tests

bbrks:
CBG-4432

- Allows BlipTesterClient to delete docs and push them as tombstones.
- Enables topology tests utilizing deletes

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2873/
- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2874/

### DIFF
--- a/rest/blip_api_attachment_test.go
+++ b/rest/blip_api_attachment_test.go
@@ -379,7 +379,7 @@ func TestBlipPushPullNewAttachmentNoCommonAncestor(t *testing.T) {
 		bodyText := `{"greetings":[{"hi":"alice"}],"_attachments":{"hello.txt":{"data":"aGVsbG8gd29ybGQ="}}}`
 		rev := NewDocVersionFromFakeRev("2-abc")
 		// FIXME CBG-4400:  docID: doc1 was not found on the client - expecting to update doc based on parentVersion RevID: 2-abc
-		err := btcRunner.StoreRevOnClient(btc.id, docID, &rev, []byte(bodyText))
+		_, err := btcRunner.AddRev(btc.id, docID, &rev, []byte(bodyText))
 		require.NoError(t, err)
 
 		bodyText = `{"greetings":[{"hi":"alice"}],"_attachments":{"hello.txt":{"revpos":2,"length":11,"stub":true,"digest":"sha1-Kq5sNclPz7QV2+lfQIuc6R7oRu0="}}}`
@@ -608,7 +608,7 @@ func TestBlipLegacyAttachNameChange(t *testing.T) {
 		docVersion := client1.GetDocVersion(docID)
 
 		// Store the document and attachment on the test client
-		err := btcRunner.StoreRevOnClient(client1.id, docID, &docVersion, rawDoc)
+		_, err := btcRunner.AddRev(client1.id, docID, &docVersion, rawDoc)
 		// FIXME CBG-4400: docID: doc was not found on the client - expecting to update doc based on parentVersion RevID: 1-5fc93bd36377008f96fdae2719c174ed
 		require.NoError(t, err)
 
@@ -672,7 +672,7 @@ func TestBlipLegacyAttachDocUpdate(t *testing.T) {
 
 		// Store the document and attachment on the test client
 		// FIXME CBG-4400: docID: doc was not found on the client - expecting to update doc based on parentVersion RevID: 1-5fc93bd36377008f96fdae2719c174ed
-		err := btcRunner.StoreRevOnClient(client1.id, docID, &version, rawDoc)
+		_, err := btcRunner.AddRev(client1.id, docID, &version, rawDoc)
 		require.NoError(t, err)
 		btcRunner.AttachmentsLock(client1.id).Lock()
 		btcRunner.Attachments(client1.id)[digest] = attBody
--- a/rest/utilities_testing_blip_client.go
+++ b/rest/utilities_testing_blip_client.go
@@ -1593,12 +1593,10 @@ func (btc *BlipTesterCollectionClient) upsertDoc(docID string, parentVersion *Do
 		return nil, err
 	}
 
-	btc._seqLast++
-	newSeq := btc._seqLast
-
 	var docVersion DocVersion
 	if btc.UseHLV() {
-		newVersion := db.Version{SourceID: fmt.Sprintf("btc-%d", btc.parent.id), Value: uint64(newSeq)}
+		// TODO: CBG-4440 Construct a HLC for Value - UnixNano is not accurate enough on Windows to generate unique values, and seq is not comparable across clients.
+		newVersion := db.Version{SourceID: fmt.Sprintf("btc-%d", btc.parent.id), Value: uint64(time.Now().UnixNano())}
 		if err := hlv.AddVersion(newVersion); err != nil {
 			return nil, err
 		}
@@ -1609,6 +1607,8 @@ func (btc *BlipTesterCollectionClient) upsertDoc(docID string, parentVersion *Do
 		docVersion = DocVersion{RevTreeID: newRevID}
 	}
 
+	btc._seqLast++
+	newSeq := btc._seqLast
 	rev := clientDocRev{clientSeq: newSeq, version: docVersion, body: body, HLV: hlv, isDelete: body == nil}
 	doc.addNewRev(rev)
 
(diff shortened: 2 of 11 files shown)


rest/blip_api_attachment_test.go
```diff
@@ -379,7 +379,7 @@ func TestBlipPushPullNewAttachmentNoCommonAncestor(t *testing.T) {
 		bodyText := `{"greetings":[{"hi":"alice"}],"_attachments":{"hello.txt":{"data":"aGVsbG8gd29ybGQ="}}}`
 		rev := NewDocVersionFromFakeRev("2-abc")
 		// FIXME CBG-4400:  docID: doc1 was not found on the client - expecting to update doc based on parentVersion RevID: 2-abc
-		err := btcRunner.StoreRevOnClient(btc.id, docID, &rev, []byte(bodyText))
+		_, err := btcRunner.AddRev(btc.id, docID, &rev, []byte(bodyText))
```
bbrks, line 382: `StoreRevOnClient` was obsolete and was just a wrapper around `AddRev` - removed for simplicity.

topologytest/peer_test.go
```diff
@@ -411,40 +411,34 @@ func TestPeerImplementation(t *testing.T) {
...
-
 			// Delete
 			deleteVersion := peer.DeleteDocument(collectionName, docID)
 			require.NotEmpty(t, deleteVersion.CV(t))
 			require.NotEqual(t, deleteVersion.CV(t), updateVersion.docMeta.CV(t))
 			require.NotEqual(t, deleteVersion.CV(t), createVersion.docMeta.CV(t))
-			if tc.peerOption.Type == PeerTypeCouchbaseServer {
-				require.Empty(t, deleteVersion.RevTreeID)
-			} else {
+			if tc.peerOption.Type == PeerTypeSyncGateway {
```
bbrks, line 413: I saw this change elsewhere previously - and now that CBL is enabled, SG is the only peer that is writing RevTreeIDs.

topologytest/topologies_test.go
```diff
@@ -146,7 +146,6 @@ var Topologies = []Topology{
 			"sg1":  {Type: PeerTypeSyncGateway, BucketID: PeerBucketID1},
 			"sg2":  {Type: PeerTypeSyncGateway, BucketID: PeerBucketID2, Symmetric: true},
 			"cbl1": {Type: PeerTypeCouchbaseLite},
-			// TODO: CBG-4270, push replication only exists empemerally
```
bbrks, line 149: Should've been removed when `cbl2` below got added I think?

rest/utilities_testing_blip_client.go
```diff
@@ -314,6 +314,32 @@ func (btcc *BlipTesterCollectionClient) GetDoc(docID string) ([]byte, *DocVersio
 	return latestRev.body, &latestRev.version
 }
 
+// IsTombstoned returns true if the latest version of the doc is a tombstone.
```
torcolvin, line 317: I am actually not sure if couchbase lite uses the terminology tombstone? I think this function is fine but I wonder if the name is confusing because we consider having server tombstones, and sync gateway tombstones as separate things and this is neither.

rest/utilities_testing_blip_client.go
```diff
@@ -314,6 +314,32 @@ func (btcc *BlipTesterCollectionClient) GetDoc(docID string) ([]byte, *DocVersio
...
+func (btcc *BlipTesterCollectionClient) IsTombstoned(docID string) (bool, error) {
+	doc, exists := btcc.getClientDoc(docID)
+	if !exists {
+		return false, base.ErrNotFound
```
torcolvin, line 321: is wrapping these errors with the docid going to be helpful if they fail?

rest/utilities_testing_blip_client.go
```diff
@@ -1587,6 +1621,15 @@ func (btc *BlipTesterCollectionClient) upsertDoc(docID string, parentVersion *Do
 	return &rev, nil
 }
 
+// Delete creates a tombstone for the document.
+func (btc *BlipTesterCollectionClient) Delete(docID string, parentVersion *DocVersion) (DocVersion, error) {
```
torcolvin, line 1626: should this require that parentVersion != nil? While you can create a tombstone on sync gateway I don't think this is a valid thing for lite to do.

topologytest/couchbase_lite_mock_peer_test.go
```diff
@@ -135,20 +145,36 @@ func (p *CouchbaseLiteMockPeer) WaitForDocVersion(dsName sgbucket.DataStoreName,
...
 // WaitForTombstoneVersion waits for a document to reach a specific version, this must be a tombstone. The test will fail if the document does not reach the expected version in 20s.
-func (p *CouchbaseLiteMockPeer) WaitForTombstoneVersion(_ sgbucket.DataStoreName, _ string, _ DocMetadata) {
-	require.Fail(p.TB(), "WaitForTombstoneVersion not yet implemented CBG-4257")
+func (p *CouchbaseLiteMockPeer) WaitForTombstoneVersion(dsName sgbucket.DataStoreName, docID string, expected DocMetadata) {
+	client := p.getSingleSGBlipClient().CollectionClient(dsName)
+	expectedVersion := db.DocVersion{CV: expected.CV(p.TB())}
+	require.EventuallyWithT(p.TB(), func(c *assert.CollectT) {
+		isTombstone, err := client.IsVersionTombstone(docID, expectedVersion)
+		require.NoError(c, err)
+		require.True(c, isTombstone, "expected docID %s on peer %s to be deleted", docID, p)
```
torcolvin, line 153: The thing that is very helpful here is to actually show the latest version of the document when this fails.

After the test fails, the replications are still running, so examining the bucket doesn't always yield the right result.

topologytest/couchbase_lite_mock_peer_test.go
```diff
@@ -135,20 +145,36 @@ func (p *CouchbaseLiteMockPeer) WaitForDocVersion(dsName sgbucket.DataStoreName,
...
 }
 
 // RequireDocNotFound asserts that a document does not exist on the peer.
-func (p *CouchbaseLiteMockPeer) RequireDocNotFound(sgbucket.DataStoreName, string) {
-	// not implemented yet in blip client tester
-	// _, err := p.btcRunner.GetDoc(p.btc.id, docID)
-	// base.RequireDocNotFoundError(p.btcRunner.TB(), err)
+func (p *CouchbaseLiteMockPeer) RequireDocNotFound(dsName sgbucket.DataStoreName, docID string) {
+	client := p.getSingleSGBlipClient().CollectionClient(dsName)
+	isTombstone, err := client.IsTombstoned(docID)
```
torcolvin: See above, the thing that is really helpful is to show the body + version if this function fails.

This function isn't actually used in tests IIRC.

topologytest/peer_test.go
torcolvin: This test was just to make sure the testing code works correctly, this seems right.

torcolvin (commented):
LGTM generally.

Since the tests that are failing are new, I'd be tempted to disable the `TestMultiActorConflictDelete/2x_CBL<->SG<->CBS_XDCR_only_1.3` with a skip with a specific investigation. It might be addressed by Adam's changes?

rest/utilities_testing_blip_client.go
bbrks: I don't think so. The caller knows the doc ID and can choose to log it/wrap it at that point

rest/utilities_testing_blip_client.go
bbrks: Yep, valid point - might as well add this guard to prevent misuse.

rest/utilities_testing_blip_client.go
bbrks: I think tombstone is standard terminology across all of Couchbase for a soft or replicated delete. I'm not sure this is a misuse of the term.

topologytest/couchbase_lite_mock_peer_test.go
bbrks: I considered this, but the combination of `require` and `EventuallyWithT` means it's not that easy to grab additional information in the failure cases.

As an aside, `WaitForDeletion` and `RequireDocNotFound` don't seem to be used by topology tests... so maybe we can just delete this code and shrink the peer interface?

torcolvin: Sounds good to me, I wrote them before I wrote the test.

torcolvin (approved):

adamcfraser (approved):

PR #7252, opened 2024-12-20
CBG-4416 Use waitForTombstoneVersion to validate tombstone versions

adamcfraser:
Validates CV for deletes.  Uptakes rosmar fix and includes a few test tweaks to ensure CBL version isn't used for lastWrite until CBL deletes are implemented in test framework.

CBG-4416

### DIFF
--- a/topologytest/hlv_test.go
+++ b/topologytest/hlv_test.go
@@ -58,14 +58,14 @@ func waitForVersionAndBody(t *testing.T, dsName base.ScopeAndCollectionName, pee
 	}
 }
 
-func waitForDeletion(t *testing.T, dsName base.ScopeAndCollectionName, peers Peers, docID string, deleteActor string) {
-	for peerName, peer := range peers {
+func waitForTombstoneVersion(t *testing.T, dsName base.ScopeAndCollectionName, peers Peers, docID string, expectedVersion BodyAndVersion) {
+	for peerName, peer := range peers.SortedPeers() {
 		if peer.Type() == PeerTypeCouchbaseLite {
 			t.Logf("skipping deletion check for Couchbase Lite peer %s, CBG-4432", peerName)
 			continue
 		}
-		t.Logf("waiting for doc to be deleted on %s, written from %s", peer, deleteActor)
-		peer.WaitForDeletion(dsName, docID)
+		t.Logf("waiting for tombstone version %#v on %s, written from %s", expectedVersion, peer, expectedVersion.updatePeer)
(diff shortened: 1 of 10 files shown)


xdcr/rosmar_xdcr.go
```diff
@@ -111,6 +111,11 @@ func (r *rosmarManager) processEvent(ctx context.Context, event sgbucket.FeedEve
 			base.WarnfCtx(ctx, "Skipping replicating doc %s, could not perform a kv op get doc in toBucket: %s", event.Key, err)
 			r.errorCount.Add(1)
 			return false
+		} else if base.IsDocNotFoundError(err) {
+			// Log if the target document exists as tombstone but doesn't have xattrs we care about
+			if actualTargetCas != 0 {
+				base.DebugfCtx(ctx, base.KeyVV, "Target document exists as a tombstone, but does not have _vv, _mou, _sync")
+			}
```
torcolvin, line 118: Is this useful to leave in debugging? At this point it might or might not get replicated depending on conflict resolution, despite having no _vv

topologytest/hlv_test.go
```diff
@@ -138,7 +138,11 @@ func deleteConflictDocs(t *testing.T, dsName base.ScopeAndCollectionName, peers
 			continue
 		}
 		deleteVersion := peer.DeleteDocument(dsName, docID)
-		t.Logf("deleteVersion: %+v", deleteVersion)
+		if peer.Type() == PeerTypeCouchbaseLite {
+			t.Logf("Don't include deleteVersion from Couchbase Lite peers when determining lastWrite %s, CBG-4432", peerName)
+			continue
+		}
+		t.Logf("deleteVersion on peer %s: %+v, implicit version: %+v", peerName, deleteVersion, deleteVersion.ImplicitHLV)
```
torcolvin: Is implicit version helpful to log here? I think logging with `%#v` will log all the details. Whether or not something has an implicit hlv only indicates whether it was written by couchbase server, not via XDCR.

torcolvin (approved):
LGTM.

xdcr/rosmar_xdcr.go
adamcfraser: I left it in as it's a bit of an unusual case that I don't expect us to hit in future (when we've removed the need for the dummy xattr), and it would be useful to see when it continues to show up in that case.

topologytest/hlv_test.go
adamcfraser: Agree - will remove the redundant implicitVersion..

torcolvin (approved):

bbrks (approved):

PR #7256, opened 2025-01-03
CBG-4373: implement flusher interface

gregns1:
CBG-4373

Implement flusher interface on counted and non counted response writers.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2875/

### DIFF
--- a/rest/api_test.go
+++ b/rest/api_test.go
@@ -2873,17 +2873,12 @@ func TestBufferFlush(t *testing.T) {
 	require.NoError(t, err)
 	require.NoError(t, a.Save(user))
 
-	// create some changes
-	for i := 0; i < 10; i++ {
-		rt.PutDoc(fmt.Sprint(i), `{"some":"doc", "channels":["foo"]}`)
-	}
-
 	var wg sync.WaitGroup
 	var resp *TestResponse
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		resp = rt.SendUserRequest(http.MethodGet, "/{{.keyspace}}/_changes?feed=continuous&since=0&timeout=5000&include_docs=true", "", "foo")
+		resp = rt.SendUserRequest(http.MethodGet, "/{{.keyspace}}/_changes?feed=continuous&since=0&timeout=500&include_docs=true", "", "foo")
 		RequireStatus(t, resp, http.StatusOK)
 	}()
 	wg.Wait()
--- a/rest/api_test.go
+++ b/rest/api_test.go
@@ -2857,3 +2857,37 @@ func TestAllDbs(t *testing.T) {
 	RequireStatus(t, resp, http.StatusOK)
 	require.Equal(t, fmt.Sprintf(`[{"db_name":"%s","bucket":"%s","state":"Online"}]`, rt.GetDatabase().Name, rt.GetDatabase().Bucket.GetName()), resp.Body.String())
 }
+
+// TestBufferFlush will test for http.ResponseWriter implements Flusher interface
+func TestBufferFlush(t *testing.T) {
+	rt := NewRestTester(t, &RestTesterConfig{
+		SyncFn: channels.DocChannelsSyncFunction,
+	})
+	defer rt.Close()
+	ctx := base.TestCtx(t)
+
+	a := rt.ServerContext().Database(ctx, "db").Authenticator(ctx)
+
+	// Create a test user
+	user, err := a.NewUser("foo", "letmein", channels.BaseSetOf(t, "foo"))
+	require.NoError(t, err)
+	require.NoError(t, a.Save(user))
+
+	// create some changes
+	for i := 0; i < 10; i++ {
+		rt.PutDoc(fmt.Sprint(i), `{"some":"doc", "channels":["foo"]}`)
+	}
+
+	var wg sync.WaitGroup
+	var resp *TestResponse
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		resp = rt.SendUserRequest(http.MethodGet, "/{{.keyspace}}/_changes?feed=continuous&since=0&timeout=5000&include_docs=true", "", "foo")
+		RequireStatus(t, resp, http.StatusOK)
+	}()
+	wg.Wait()
+
+	// assert that the response is a flushed response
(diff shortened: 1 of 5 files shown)


rest/api_test.go
```diff
@@ -2857,3 +2857,37 @@ func TestAllDbs(t *testing.T) {
 	RequireStatus(t, resp, http.StatusOK)
 	require.Equal(t, fmt.Sprintf(`[{"db_name":"%s","bucket":"%s","state":"Online"}]`, rt.GetDatabase().Name, rt.GetDatabase().Bucket.GetName()), resp.Body.String())
 }
+
+// TestBufferFlush will test for http.ResponseWriter implements Flusher interface
+func TestBufferFlush(t *testing.T) {
+	rt := NewRestTester(t, &RestTesterConfig{
```
torcolvin, line 2863: I think we can shorten this test, because we don't necessarily care about having any documents. I also think we don't have to test with a continuous changes feed.

What we do want to test is that all our response writers support flush. I think I'd be inclined to write a different type of test like https://github.com/couchbase/sync_gateway/blob/88edb9a72a24e4d1fd00e4ec844d2e2d034a58b4/rest/counted_response_writer_test.go#L132 using the same testcases and `getResponseWriter` but have the `httptest.NewRecorder` returned by `getResponseWriter`.

rest/counted_response_writer.go
```diff
@@ -92,3 +92,10 @@ func (w *CountedResponseWriter) isHijackable() bool {
 	_, ok := w.writer.(http.Hijacker)
 	return ok
 }
+
+func (w *CountedResponseWriter) Flush() {
```
torcolvin, line 96: I think we should add some compile time assertions for all the interfaces that we implement them.

```
var _ http.Flusher = &CountedResponseWriter{}
var _ http.Flusher = &NonCountedResponseWriter{}
var _ http.Flusher = &EncodedResponseWriter{}
```

while we are there, I think we should assert on a hijacker interface, since I made that mistake too, before.

```
var _ http.Hijacker = &CountedResponseWriter{}
var _ http.Hijacker = &NonCountedResponseWriter{}
var _ http.Hijacker = &EncodedResponseWriter{}
```

rest/handler.go
```diff
@@ -671,6 +671,11 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 		}
 	}
 	h.updateResponseWriter()
+	// ensure wrapped ResponseWriter implements http.Flusher
+	_, ok := h.response.(http.Flusher)
+	if !ok {
+		return fmt.Errorf("ResponseWriter does not implement Flusher interface")
```
```suggestion
		return fmt.Errorf("http.ResponseWriter %T does not implement Flusher interface", h.response)
```

bbrks:
Might be worth changing the code in `(h *handler) flush()` to drop the type check and rely on compilation failure if `h.response` does not implement `Flush()`.

```diff
 func (h *handler) flush() {
-       switch r := h.response.(type) {
-       case http.Flusher:
-               r.Flush()
-       }
+       h.response.Flush()
 }
```

Avoids the silent failure we were bitten by before, and wouldn't need extra type checking or test coverage since it's not even compilable.

I think the following type of approach to implement Flush for the wrappers is just as susceptible to silent failures, where we call `Flush()`, but actually the underlying response will never get flushed.

```go
func (w *NonCountedResponseWriter) Flush() {
	f, ok := w.ResponseWriter.(http.Flusher)
	if ok {
		f.Flush()
	}
}
```

Would much prefer to let this panic if Flush is not implemented and we actually call `Flush()`

bbrks (changes requested):
Might be worth changing the code in `(h *handler) flush()` to drop the type check and rely on compilation failure if `h.response` does not implement `Flush()`.

```diff
 func (h *handler) flush() {
-       switch r := h.response.(type) {
-       case http.Flusher:
-               r.Flush()
-       }
+       h.response.Flush()
 }
```

Avoids the silent failure we were bitten by before, and wouldn't need extra type checking or test coverage since it's not even compilable.

I think the following type of approach to implement Flush for the wrappers is just as susceptible to silent failures, where we call `Flush()`, but actually the underlying response will never get flushed.

```go
func (w *NonCountedResponseWriter) Flush() {
	f, ok := w.ResponseWriter.(http.Flusher)
	if ok {
		f.Flush()
	}
}
```

Would much prefer to let this panic if Flush is not implemented and we actually call `Flush()`

rest/handler.go
```diff
@@ -1595,10 +1607,8 @@ func (h *handler) writeMultipart(subtype string, callback func(*multipart.Writer
 }
 
 func (h *handler) flush() {
-	switch r := h.response.(type) {
-	case http.Flusher:
-		r.Flush()
-	}
+	r, _ := h.response.(http.Flusher)
+	r.Flush()
```
bbrks: This is still a runtime check of `http.Flusher` on the concrete type, and the panic message is not super clear since you're relying on a `nil` `r`:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x18 pc=0x103491948]

goroutine 358 [running]:
github.com/couchbase/sync_gateway/rest.(*handler).flush(...)
	/Users/benbrooks/dev/cb/sg/rest/handler.go:1599
github.com/couchbase/sync_gateway/rest.(*handler).sendContinuousChangesByHTTP.func1({0x14000808090?, 0x1?, 0x700000005?})
	/Users/benbrooks/dev/cb/sg/rest/changes_api.go:501 +0xb8
```

Preferably switch to non-interface call for compilation failure:

```suggestion
	h.response.Flush()
```

```
# github.com/couchbase/sync_gateway/rest
rest/handler.go:1598:13: h.response.Flush undefined (type CountableResponseWriter has no field or method Flush)
```

Or, lower preference since it's a runtime check: A shorthand type assertion will give a better panic message at runtime:


```suggestion
	h.response.(http.Flusher).Flush()
```

```
panic: interface conversion: *rest.CountedResponseWriter is not http.Flusher: missing method Flush
```

bbrks (changes requested):

bbrks (approved):

PR #7259, opened 2025-01-03
CBG-4444 create an env var override for bucket op timeout

torcolvin:
This address a problem where running a test like `TestWriteTombstone` immediately after starting the test harness will fail with: `ambiguous timeout | {"operation_id":"MutateIn","opaque":"9","time_observed":10000808,"last_dispatched_to":"172.17.0.3:11210","last_dispatched_from":"172.17.0.1:55254","last_connection_id":"7e7da8507602d388/c099ed3d3ef4cbbf"}
`

The problem arises from flush bucket not being a synchronous operation and each vbucket needs to be able ready to accept writes. This can also occur after flushing a bucket between tests, even after the initial creation.

The reason it happens more frequently on my computer is that the tests are generally faster to run, and so Couchbase Server has less time to process its bucket operations before it calls `MutateIn`. This occurs on a Linux machine running server versions 7.2->7.6.

Remove the hard coded timeout of 10 seconds since that is the default anyway. https://github.com/couchbase/sync_gateway/blob/d61c39261271d3f04c8d714c3455c1b6ad667230/base/constants.go#L154

### DIFF
--- a/base/main_test_bucket_pool_config.go
+++ b/base/main_test_bucket_pool_config.go
@@ -58,6 +58,9 @@ const (
 
 	// Creates buckets with a specific number of number of replicas
 	tbpEnvBucketNumReplicas = "SG_TEST_BUCKET_NUM_REPLICAS"
+
+	// bucket op timeout is the number of seconds that kv operations will retry. Increasing this can be necessary when flushing buckets rapidly.
+	tbpEnvBucketOpTimeout = "SG_TEST_BUCKET_OP_TIMEOUT"
 )
 
 // TestsUseNamedCollections returns true if the tests use named collections.
--- a/base/main_test_bucket_pool_config.go
+++ b/base/main_test_bucket_pool_config.go
@@ -58,9 +58,6 @@ const (
 
 	// Creates buckets with a specific number of number of replicas
 	tbpEnvBucketNumReplicas = "SG_TEST_BUCKET_NUM_REPLICAS"
-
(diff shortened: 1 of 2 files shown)


bbrks (dismissed):
I wouldn't be opposed to bumping the timeout instead of adding a config option but will leave up to you.

torcolvin:
> I wouldn't be opposed to bumping the timeout instead of adding a config option but will leave up to you.

I would prefer to bump the timeout lest I forget to set this variable or AWS machines have different performance implications. The thing that gives me pause is that this will become a different timeout than we use for production uses. This hasn't bitten any customers that we know of, and usually operations are retried anyway at some level. I could mitigate this by only setting this when we use FlushBucketEmptier.

Separately, Brett and I experimented with what would happen if we used gocbcorex (not possible in production, but just to see if it was suspectible to the same things), and the answer was that it wouldn't happen, but this comes down to timing issues of when gocbcorex decides a bucket is ready.  

I guess that a customer who had a problem could set `kv_timeout=20` in the connection string, as could I in `SG_TEST_COUCHBASE_SERVER_URL`.

Does any of this information the way you feel? I think either I'll add something to my connection string in `SG_TEST_COUCHBASE_SERVER_URL` and drop this PR or we should bump the timeout for all tests.

bbrks:
Different test vs. production default timeouts is a fair concern though not one I'm particularly worried about. Customers aren't going to be flushing buckets at the frequency we are and I agree that it's going to be easy to forget to change this when relevant for testing.

bbrks (approved):

PR #7263, opened 2025-01-06
golangci-lint: skip prealloc/fieldalignment in all test files

torcolvin:
There are test files named `*test*.go` like `utilities_testing.go` or `hlv_utilities_testing.go` and we do not care about aligning struct fields or preallocating maps/arrays for these.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -33,28 +33,27 @@ linters:
     - unconvert # Remove unnecessary type conversions
     - unparam # Reports unused function parameters
     - unused # (megacheck) Checks Go code for unused constants, variables, functions and types
-  disable:
-    - asciicheck # Simple linter to check that your code does not contain non-ASCII identifiers
-    - depguard # Go linter that checks if package imports are in a list of acceptable packages
-    - dogsled # Checks assignments with too many blank identifiers # (e.g. x, _, _, _, := f())
-    - err113 # Golang linter to check the errors handling expressions
-    - funlen # Tool for detection of long functions
-    - gochecknoglobals # Checks that no globals are present in Go code
-    - gochecknoinits # Checks that no init functions are present in Go code
-    - gocognit # Computes and checks the cognitive complexity of functions
-    - gocyclo # Computes and checks the cyclomatic complexity of functions
-    - godot # Check if comments end in a period
-    - godox # Tool for detection of FIXME, TODO and other comment keywords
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -70,10 +70,12 @@ linters-settings:
 # Disable goconst in test files, often we have duplicated strings across tests, but don't make sense as constants.
 issues:
   exclude-rules:
-    - path: (_test\.go|utilities_testing\.go)
+    # cover utilities_testing.go and _test.go files
+    - path: (_test.*go)
```
bbrks: I think the `.go` suffix still needs escaping with `\.go`

```suggestion
    - path: (_test.*\.go)
```

bbrks (changes requested):

.golangci.yml
```diff
@@ -70,10 +70,12 @@ linters-settings:
...
 issues:
   exclude-rules:
-    - path: (_test\.go|utilities_testing\.go)
+    # cover utilities_testing.go and _test.go files
+    - path: (_test.*\.go)
       linters:
         - goconst
-    - path: (_test\.go|utilities_testing\.go)
+        - prealloc
+    - path: (_test.*go)
```
bbrks: ditto here - sorry didn't leave suggestion in both places

 ```suggestion
    - path: (_test.*\.go)
```

bbrks (changes requested):

bbrks (dismissed):

.golangci.yml
bbrks: These previous suggestions/fixes have been lost in the force push

bbrks (changes requested):

torcolvin (commented):
I did another force push and I updated the rules to account for base/util_testing.go

bbrks:
> I did another force push and I updated the rules to account for base/util_testing.go

Maybe I just don't understand this syntax...

Wouldn't `util_testing.go` already be covered by the `.*` part of the `_test.*\.go` rule, since `ing` is a suffix of `_test` preceeding `.go`?

torcolvin:
> > I did another force push and I updated the rules to account for base/util_testing.go
> 
> Maybe I just don't understand this syntax...
> 
> Wouldn't `util_testing.go` already be covered by the `.*` part of the `_test.*\.go` rule, since `ing` is a suffix of `_test` preceeding `.go`?

user error, stand by

bbrks (approved):

PR #7265, opened 2025-01-07
CBG-4336: add updated at field for persisted configs

gregns1:
CBG-4336

- Add updated at timestamp to persisted configs
- The following configs are included:
  - Registry 
  - Db configs 
  - Replication configs 
  - User and role docs 

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2884/

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -632,7 +632,6 @@ func (auth *Authenticator) UpdateUserEmail(u User, email string) error {
 			return nil, err
 		}
 		currentUser.SetUpdatedAt()
-		currentUser.SetCreatedAt(currentUser.GetCreatedAt())
 
 		return currentUser, nil
 	}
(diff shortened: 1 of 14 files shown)


bbrks:
@JRascagneres Can you review and make sure the `updated_at` name and format/UTC aligns with stuff on your side? And input on whether `created_at` is necessary as well.

bbrks:
As discussed - add a `created_at` field as well, but don't care about timestamp format - keep that as-is.

auth/auth.go
```diff
@@ -631,6 +631,9 @@ func (auth *Authenticator) UpdateUserEmail(u User, email string) error {
 		if err != nil {
 			return nil, err
 		}
+		currentUser.SetUpdatedAt()
+		currentUser.SetCreatedAt(currentUser.GetCreatedAt())
```
bbrks: I don't understand why the `Set(Get())` pattern is required. I think it's not nescessary to call `SetCreatedAt` in these places at all, since the time is already set.

bbrks (changes requested):
Looks good, wonder if we can simplify a bit around Principal with the `SetCreatedAt` usage - but nitpicking.

auth/auth.go
gregns1: Good point, not sure why I thought this was needed. Will remove this now

bbrks (approved):

PR #7268, opened 2025-01-07
CBG-4402 do not terminate attachment migration error on document deletion

torcolvin:
A deleted document should not even be processed, before it would error in `UnmarshalDocumentSyncDataFromFeed`.

I also changed the code so that it doesn't error the job, but just skips processing this document. In the case that a document has invalid `_sync` data, I think other documents shouldn't also fail.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2883/

### DIFF
--- a/db/background_mgr_attachment_migration.go
+++ b/db/background_mgr_attachment_migration.go
@@ -100,18 +100,15 @@ func (a *AttachmentMigrationManager) Run(ctx context.Context, options map[string
 	}
 	defer persistClusterStatus()
 
-	var processFailure error
-	failProcess := func(err error, format string, args ...interface{}) bool {
-		processFailure = err
-		terminator.Close()
-		base.WarnfCtx(ctx, format, args...)
-		return false
-	}
-
 	callback := func(event sgbucket.FeedEvent) bool {
 		docID := string(event.Key)
 		collection := db.CollectionByID[event.CollectionID]
 		base.TracefCtx(ctx, base.KeyAll, "[%s] Received DCP event %d for doc %v", migrationLoggingID, event.Opcode, base.UD(docID))
+
+		// Ignore non-mutation events: Deletion, Backfill, etc.
+		if event.Opcode != sgbucket.FeedOpMutation {
+			return true
+		}
 
 		// Ignore documents without xattrs, to avoid processing unnecessary documents
 		if event.DataType&base.MemcachedDataTypeXattr == 0 {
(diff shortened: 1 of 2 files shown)


db/background_mgr_attachment_migration.go
```diff
@@ -76,11 +78,15 @@ func (a *AttachmentMigrationManager) Init(ctx context.Context, options map[strin
 		if statusDoc.State == BackgroundProcessStateCompleted || err != nil || reset {
 			return newRunInit()
 		}
-		a.MigrationID = statusDoc.MigrationID
-		a.SetStatus(statusDoc.DocsChanged, statusDoc.DocsProcessed)
-		a.SetCollectionIDs(statusDoc.CollectionIDs)
+		a.lock.Lock()
```
adamcfraser: Since the stats are all atomics I'm assuming the lock is for migrationID and collectionIDs.  We don't use locking anywhere else when accessing migrationID (including this function), and have an existing function (SetCollectionIDs) to manage locking for CollectionIds, so the lock/defer unlock approach doesn't feel like a good fit.

db/background_mgr_attachment_migration.go
```diff
@@ -216,46 +220,49 @@ func (a *AttachmentMigrationManager) Run(ctx context.Context, options map[string
...
 	defer a.lock.Unlock()
 
-	a.DocsProcessed.Set(0)
-	a.DocsChanged.Set(0)
-	a.CollectionIDs = nil
+	a.docsProcessed.Store(0)
+	a.docsChanged.Store(0)
+	a.docsFailed.Store(0)
+	a.collectionIDs = nil
+	a.migrationID = ""
```
adamcfraser: The changes to migrationID handling in this PR seem unrelated to the fix (not terminating on document deletion).  

Can you provide a bit of context on why it's needed?  Is it a bug related to DCP connection names when restarting a replication?

adamcfraser (dismissed):

bbrks (approved):

db/background_mgr_attachment_migration.go
torcolvin: I removed this code. I do not think it is right but it is outside the scope of this problem.

db/background_mgr_attachment_migration.go
torcolvin: I removed this code. I do not think it is right but it is outside the scope of this problem.

PR #7271, opened 2025-01-08
CBG-4418 Assert on full HLV in topology tests

adamcfraser:
CBG-4418

Modify waitForVersionAndBody to wait for and assert on full HLV instead of cv only.

Includes additional fixes required to get tests passing when checking full HLV:
- When calling HLV.AddVersion, remove existing entry for source from PV (when present)
- in CouchbaseServerPeer.WriteDocument, do not populate lastXattrs for document resurrection case (Couchbase doesn't preserve xattrs on resurrection)
- in getDocVersion (peer tests), remove unnecessary copy of HLV (previous HLV is being directly unmarshalled into Implicit HLV)

Also reduces pollInterval for topology tests to reduce test execution time (avoids minimum required wait time for each check).

### DIFF
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -13,6 +13,7 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"fmt"
+	"maps"
 	"sort"
 	"strconv"
 	"strings"
@@ -187,22 +188,15 @@ func (hlv *HybridLogicalVector) Equals(other *HybridLogicalVector) bool {
 	if hlv.Version != other.Version {
 		return false
 	}
-	if len(hlv.PreviousVersions) != len(other.PreviousVersions) {
+
+	if !maps.Equal(hlv.PreviousVersions, other.PreviousVersions) {
 		return false
 	}
-	for k, v := range hlv.PreviousVersions {
-		if other.PreviousVersions[k] != v {
-			return false
-		}
-	}
-	if len(hlv.MergeVersions) != len(other.MergeVersions) {
+
+	if !maps.Equal(hlv.MergeVersions, other.MergeVersions) {
 		return false
 	}
-	for k, v := range hlv.MergeVersions {
-		if other.MergeVersions[k] != v {
-			return false
-		}
-	}
+
 	return true
 }
 
@@ -234,7 +228,7 @@ func (hlv *HybridLogicalVector) DominatesSource(version Version) bool {
 
 }
 
-// AddVersion adds newVersion to the in memory representation of the HLV.
+// AddVersion adds newVersion as the current version to the in memory representation of the HLV.
 func (hlv *HybridLogicalVector) AddVersion(newVersion Version) error {
 	var newVersionCAS uint64
 	hlvVersionCAS := hlv.Version
@@ -275,9 +269,7 @@ func (hlv *HybridLogicalVector) AddVersion(newVersion Version) error {
 	}
 
 	// If new version already exists PV, need to remove it
-	if _, ok := hlv.PreviousVersions[newVersion.SourceID]; ok {
-		delete(hlv.PreviousVersions, newVersion.SourceID)
-	}
+	delete(hlv.PreviousVersions, newVersion.SourceID)
 
 	hlv.Version = newVersion.Value
 	hlv.SourceID = newVersion.SourceID
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -180,6 +180,32 @@ func NewHybridLogicalVector() *HybridLogicalVector {
 	}
 }
 
+func (hlv *HybridLogicalVector) Equals(other *HybridLogicalVector) bool {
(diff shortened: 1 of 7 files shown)


db/hybrid_logical_vector.go
```diff
@@ -180,6 +180,32 @@ func NewHybridLogicalVector() *HybridLogicalVector {
...
+		}
+	}
+	if len(hlv.MergeVersions) != len(other.MergeVersions) {
+		return false
+	}
+	for k, v := range hlv.MergeVersions {
+		if other.MergeVersions[k] != v {
+			return false
+		}
+	}
```
torcolvin: fyi, there's stdlib map comparison functions so you can do:

```suggestion
	if !maps.Equal(hlv.PreviousVersions, other.PreviousVersions) {
		return false
	}
	if !maps.Equal(hlv.MergeVersions, other.MergeVersions) {
		return false
	}
```

db/hybrid_logical_vector.go
```diff
@@ -247,6 +273,12 @@ func (hlv *HybridLogicalVector) AddVersion(newVersion Version) error {
 		// source doesn't exist in PV so add
 		hlv.PreviousVersions[hlv.SourceID] = hlv.Version
 	}
+
+	// If new version already exists PV, need to remove it
+	if _, ok := hlv.PreviousVersions[newVersion.SourceID]; ok {
+		delete(hlv.PreviousVersions, newVersion.SourceID)
+	}
```
torcolvin: take or leave, delete() when it doesn't exist is a no-op 
```suggestion
	delete(hlv.PreviousVersions, newVersion.SourceID)
```

topologytest/couchbase_server_peer_test.go
```diff
@@ -183,6 +187,7 @@ func (p *CouchbaseServerPeer) waitForDocVersion(dsName sgbucket.DataStoreName, d
 		}
 		// have to use p.tb instead of c because of the assert.CollectT doesn't implement TB
 		version = getDocVersion(docID, p, cas, xattrs)
+		assert.True(c, version.IsHLVEqual(expected), "Actual HLV does not match expected on %s for peer %s.  Expected: %#v, Actual: %#v", docID, p, expected, version)
 		assert.Equal(c, expected.CV(c), version.CV(c), "Could not find matching CV on %s for peer %s\nexpected: %#v\nactual:   %#v\n          body: %#v\n", docID, p, expected, version, string(docBytes))
```
torcolvin: should we drop the CV assertion here?

topologytest/couchbase_lite_mock_peer_test.go
```diff
@@ -133,6 +133,7 @@ func (p *CouchbaseLiteMockPeer) WaitForDocVersion(dsName sgbucket.DataStoreName,
 		if !assert.NotNil(c, actual, "Could not find docID:%+v on %p\nVersion %#v", docID, p, expected) {
 			return
 		}
+		assert.True(c, actual.IsHLVEqual(expected), "Actual HLV does not match expected on %s for peer %s.  Expected: %s, Actual: %s", docID, p, expected.HLV, actual.HLV)
 		assert.Equal(c, expected.CV(c), actual.CV(c), "Could not find matching CV on %s for peer %s (sourceID:%s)\nexpected: %#v\nactual:   %#v\n          body: %+v\n", docID, p, p.SourceID(), expected, actual, string(data))
```
torcolvin: should we drop the CV assertion?

topologytest/couchbase_lite_mock_peer_test.go
```diff
@@ -133,6 +133,7 @@ func (p *CouchbaseLiteMockPeer) WaitForDocVersion(dsName sgbucket.DataStoreName,
 		if !assert.NotNil(c, actual, "Could not find docID:%+v on %p\nVersion %#v", docID, p, expected) {
 			return
 		}
+		assert.True(c, actual.IsHLVEqual(expected), "Actual HLV does not match expected on %s for peer %s.  Expected: %s, Actual: %s", docID, p, expected.HLV, actual.HLV)
```
torcolvin: For debugging this, I had implemented a method like `GoString()` on `HybridLogicalVector` because `%s` will print out pointer values for `PreviousVersions` and also having int representations for `HybridLogicalVector.Value` is nice?

Was this something you felt was helpful?

torcolvin (approved):
Everything looks right to me, I did note some style and debuggability stuff but that can be ignored if you want.

topologytest/couchbase_lite_mock_peer_test.go
adamcfraser: Yes, I used that for some of the failing tests.  I don't think I had any failures at this point so hadn't noticed this wasn't switched over to use %#v - I've made that change in the latest commit.

The logs are still fairly dense to parse when triaging failures, but they do a good job of including everything that's needed to trace the replication/HLV states.

torcolvin (approved):

torcolvin (approved):

PR #7284, opened 2025-01-10
Topology test fixes

adamcfraser:
Test logging enhancements:
- verbose HLV logging in a few more places
- turn on Sync and CRUD log keys for topology tests

Flaky test fixes:
- Fix changes processing in blip test client to not request revs when version in changes message is already known to the client
- Change StopPush in blipTestClient to synchronously wait for push processing to be stopped
- rosmar XDCR: replicate _mou to target when it is not obsolete

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -1242,7 +1242,7 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 			}
 		} else {
 			if doc.HLV.isDominating(newDocHLV) {
-				base.DebugfCtx(ctx, base.KeyCRUD, "PutExistingCurrentVersion(%q): No new versions to add", base.UD(newDoc.ID))
+				base.DebugfCtx(ctx, base.KeyCRUD, "PutExistingCurrentVersion(%q): No new versions to add.  existing: %#v  new:%#v", base.UD(newDoc.ID), doc.HLV, newDocHLV)
 				return nil, nil, false, nil, base.ErrUpdateCancel // No new revisions to add
 			}
 			if newDocHLV.isDominating(doc.HLV) {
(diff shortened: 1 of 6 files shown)


db/crud.go
```diff
@@ -1242,7 +1242,7 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 			}
 		} else {
 			if doc.HLV.isDominating(newDocHLV) {
-				base.DebugfCtx(ctx, base.KeyCRUD, "PutExistingCurrentVersion(%q): No new versions to add", base.UD(newDoc.ID))
+				base.DebugfCtx(ctx, base.KeyCRUD, "PutExistingCurrentVersion(%q): No new versions to add.  existing: %#v  new:%#v", base.UD(newDoc.ID), doc.HLV, newDocHLV)
```
torcolvin, line 1245: This should possibly be KeyVV to match above

rest/utilities_testing_blip_client.go
```diff
@@ -1436,8 +1444,17 @@ func (btc *BlipTesterCollectionClient) StartPullSince(options BlipTesterPullOpti
...
-	require.True(btc.TB(), btc.pushRunning.CASRetry(true, false), "can't stop push replication - not running")
+	require.True(btc.TB(), btc.pushRunning.IsTrue(), "can't stop push replication - not running")
 	btc.pushCtxCancel()
+
+	// Wake up any waiting push loops to check for cancellation
+	btc._seqCond.Broadcast()
+
+	// wait for push replication to stop running
+	require.EventuallyWithT(btc.TB(), func(c *assert.CollectT) {
+		require.True(c, btc.pushRunning.IsTrue() == false)
```
torcolvin, line 1455:
```suggestion
		assert.False(c, btc.pushRunning.IsTrue())
```

the way that assertion is written will actually fail immediately if `btc.pushRunning.IsTrue() == true` and it won't do a backoff.

torcolvin (approved):

PR #7286, opened 2025-01-10
Multi-actor/Conflict tests and BTC pull replication conflict resolution

bbrks:
Enable multi-actor and conflict tests and allow BlipTesterClient to resolve conflicts and push up as new version.

Spun off two failing subtests into new ticket (CBG-4458) to get other improvements into `release/anemone`

### DIFF
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -491,9 +491,9 @@ func appendRevocationMacroExpansions(currentSpec []sgbucket.MacroExpansionSpec, 
 
 // ExtractHLVFromBlipMessage extracts the full HLV a string in the format seen over Blip
 // blip string may be the following formats
-//  1. cv only:    		cv
-//  2. cv and pv:  		cv;pv1,pv2
-//  3. cv, pv, and mv: 	cv;mv1,mv2;pv1,pv2
+//  1. cv only:			cv
+//  2. cv and pv:		cv;pv1,pv2
+//  3. cv+mv and pv:	cv,mv1,mv2;pv1,pv2
 //
 // Function will return list of revIDs if legacy rev ID was found in the HLV history section (PV)
 // TODO: CBG-3662 - Optimise once we've settled on and tested the format with CBL
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -346,13 +346,13 @@ func (hlv *HybridLogicalVector) AddNewerVersions(otherVector *HybridLogicalVecto
 		// for source if the local version for that source is lower
 		for i, v := range otherVector.PreviousVersions {
 			if hlv.PreviousVersions[i] == 0 {
-				hlv.setPreviousVersion(i, v)
+				hlv.SetPreviousVersion(i, v)
 			} else {
 				// if we get here then there is entry for this source in PV so we must check if its newer or not
 				otherHLVPVValue := v
 				localHLVPVValue := hlv.PreviousVersions[i]
 				if localHLVPVValue < otherHLVPVValue {
-					hlv.setPreviousVersion(i, v)
+					hlv.SetPreviousVersion(i, v)
 				}
 			}
 		}
@@ -384,8 +384,8 @@ func (hlv *HybridLogicalVector) computeMacroExpansions() []sgbucket.MacroExpansi
 	return outputSpec
 }
 
-// setPreviousVersion will take a source/version pair and add it to the HLV previous versions map
-func (hlv *HybridLogicalVector) setPreviousVersion(source string, version uint64) {
+// SetPreviousVersion will take a source/version pair and add it to the HLV previous versions map
+func (hlv *HybridLogicalVector) SetPreviousVersion(source string, version uint64) {
 	if hlv.PreviousVersions == nil {
 		hlv.PreviousVersions = make(HLVVersions)
 	}
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -384,7 +384,7 @@ func (hlv *HybridLogicalVector) computeMacroExpansions() []sgbucket.MacroExpansi
 	return outputSpec
 }
 
-// SetPreviousVersion will take a source/version pair and add it to the HLV previous versions map
+// SetPreviousVersion will take a source/version pair and sets the value for the given source in the previous versions map
 func (hlv *HybridLogicalVector) SetPreviousVersion(source string, version uint64) {
 	if hlv.PreviousVersions == nil {
 		hlv.PreviousVersions = make(HLVVersions)
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -443,6 +443,41 @@ func (hlv *HybridLogicalVector) ToHistoryForHLV() string {
 	return s.String()
 }
 
+func FromHistoryForHLV(history string) (*HybridLogicalVector, error) {
+	hlv := NewHybridLogicalVector()
+	// split the history string into PV and MV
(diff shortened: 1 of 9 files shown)


db/hybrid_logical_vector.go
```diff
@@ -375,8 +375,8 @@ func (hlv *HybridLogicalVector) computeMacroExpansions() []sgbucket.MacroExpansi
 	return outputSpec
 }
 
-// setPreviousVersion will take a source/version pair and add it to the HLV previous versions map
-func (hlv *HybridLogicalVector) setPreviousVersion(source string, version uint64) {
+// SetPreviousVersion will take a source/version pair and add it to the HLV previous versions map
```
torcolvin: This is probably contextually obvious from how hlvs work, but I might add a comment to the affect of: version passed for a given source is updated in the previous versions map for that source.

rest/utilities_testing_blip_client.go
```diff
@@ -563,20 +587,41 @@ func (btr *BlipTesterReplicator) initHandlers(btc *BlipTesterClient) {
...
-				v, err := db.ParseVersion(revID)
+				incomingCV, err := db.ParseVersion(revID)
 				require.NoError(btr.TB(), err, "error parsing version %q: %v", revID, err)
-				newVersion = DocVersion{CV: v}
-				require.NoError(btr.TB(), hlv.AddVersion(v))
+				incomingVersion = DocVersion{CV: incomingCV}
+
+				clientCV := doc._currentVersion(btc.TB())
+				// incoming rev older than stored client version and comes from a different source - need to resolve
+				if incomingCV.Value < clientCV.Value && incomingCV.SourceID != clientCV.SourceID {
```
torcolvin: Possibly there are two circumstances to handle:

- Sync Gateway sends a `rev` message for something that the client already has. It probably shouldn't do this if `proposeChanges` is working. If that's true, we should possibly error.
- Is there a case where we should look up the incomingCV in the pv as well, if there was a local write? If there's separate history in pv from incoming write and local write (e.g. p2p or xdcr), should we resolve that?

rest/utilities_testing_blip_client.go
```diff
@@ -786,24 +835,53 @@ func (btr *BlipTesterReplicator) initHandlers(btc *BlipTesterClient) {
 		doc.lock.Lock()
 		defer doc.lock.Unlock()
 
-		var newVersion DocVersion
+		var incomingVersion DocVersion
```
torcolvin, line 859: Is it worth deduplicating this from deleted rev handling above?

torcolvin (commented):
I forgot this was a WIP, please feel free to disregard comments if irrelevant.

rest/utilities_testing_blip_client.go
bbrks: Added assertion to ensure SG isn't sending a rev for an already known CV - it wasn't but there in case it regresses.

rest/utilities_testing_blip_client.go
```diff
@@ -571,28 +589,67 @@ func (btr *BlipTesterReplicator) initHandlers(btc *BlipTesterClient) {
...
+
+				// fetch client's latest version to do conflict check and resolution
+				latestClientRev, err := doc._latestRev()
+				require.NoError(btc.TB(), err, "couldn't get latest revision for doc %q", docID)
+				if latestClientRev != nil {
+					clientCV := latestClientRev.version.CV
+
+					// safety check - ensure SG is not sending a rev that we already had - ensures changes feed messaging is working correctly to prevent
+					if clientCV.SourceID == incomingCV.SourceID && clientCV.Value == incomingCV.Value {
+						require.FailNow(btc.TB(), "incoming revision %v is equal to client revision %v - should've been filtered via changes response before ending up as a rev", incomingCV, clientCV)
```
torcolvin: I'm about to throw up a PR to check this with ruleguard, but `FailNow` isn't fmt.Sprintf CBG-4455 to not panic- this needs to be:

```suggestion
						require.FailNow(btc.TB(), fmt.Sprintf("incoming revision %v is equal to client revision %v - should've been filtered via changes response before ending up as a rev", incomingCV, clientCV))
```

rest/utilities_testing_blip_client.go
```diff
@@ -1007,6 +1109,12 @@ func (btcRunner *BlipTestClientRunner) NewBlipTesterClientOptsWithRT(rt *RestTes
 	if !opts.AllowCreationWithoutBlipTesterClientRunner && !btcRunner.initialisedInsideRunnerCode {
 		require.FailNow(btcRunner.TB(), "must initialise BlipTesterClient inside Run() method")
 	}
+	if opts.ConflictResolver == "" {
+		opts.ConflictResolver = ConflictResolverDefault
+	}
+	if !opts.ConflictResolver.IsValid() {
+		require.FailNow(btcRunner.TB(), "invalid conflict resolver %q", opts.ConflictResolver)
```
torcolvin:
```suggestion
		require.FailNow(btcRunner.TB(), fmt.Sprintf("invalid conflict resolver %q", opts.ConflictResolver))
```

topologytest/sync_gateway_peer_test.go
```diff
@@ -63,7 +63,11 @@ func (p *SyncGatewayPeer) GetDocument(dsName sgbucket.DataStoreName, docID strin
 	collection, ctx := p.getCollection(dsName)
 	doc, err := collection.GetDocument(ctx, docID, db.DocUnmarshalAll)
 	require.NoError(p.TB(), err)
-	return DocMetadataFromDocument(doc), doc.Body(ctx)
+	var body db.Body
+	if !doc.IsDeleted() {
+		body = doc.Body(ctx)
```
torcolvin, line 68: this will be nil when this occurs, are we intentionally returning empty body rather than nil?

rest/utilities_testing_blip_client.go
bbrks: fixed all instances of this

topologytest/sync_gateway_peer_test.go
bbrks: Some commentary on this change in the `CouchbaseLiteMockPeer` implementation - since for a tombstone body, there were variations of `[]byte({})` vs `nil` vs `[]byte{}`.

I don't think it really matters as long as all peers treat it the same. It was required sincd `waitForConvergingVersion` is also comparing doc bodies.

```go
// GetDocument returns the latest version of a document. The test will fail the document does not exist.
func (p *CouchbaseLiteMockPeer) GetDocument(dsName sgbucket.DataStoreName, docID string) (DocMetadata, db.Body) {
	bodyBytes, meta := p.getLatestDocVersion(dsName, docID)
	require.NotNil(p.TB(), meta, "docID:%s not found on %s", docID, p)
	var body db.Body
	// it's easier if all clients can return consistent bodies for tombstones
	// lets just settle on nil, since we still need special handling anyway for `` vs `{}` so unmarshal doesn't barf
```

torcolvin (dismissed):

db/hybrid_logical_vector.go
```diff
@@ -427,6 +427,41 @@ func (hlv *HybridLogicalVector) ToHistoryForHLV() string {
...
+		// MV
+		mvs := strings.Split(versionSets[1], ",")
+		for _, mv := range mvs {
+			v, err := ParseVersion(mv)
+			if err != nil {
+				return nil, err
+			}
+			hlv.MergeVersions[v.SourceID] = v.Value
+		}
+		fallthrough
```
bbrks, line 464: This was for pre-beta implementation and needs removing/changing

torcolvin:
I'm redoing this code in CBG-4434, and I've already incorporated the parts I can use. This PR is quite stale.

PR #7293, opened 2025-01-14
CBG-4459 validate cv when passed in

torcolvin:
CBG-4459 validate cv when passed in

- validate that sourceID is base64, requires some test changes
- pass back better error messages if hlv was not found

The sourceID from CBL/CBS never include padding, so I used `base64.RawStdEncoding` to generate sourceIDs. Not all ASCII strings are valid base64, so I changed some strings.

I considered doing regexp parsing for invalid characters, since what was happening before this PR was that "1@abc; 2@def" was getting parsed as SourceID=`abc; 2@def` Value=`1`. I decided to just parse as base64 since this will provide more robust validation. There are two possible downsides to this: it is slightly more expensive to call `ParseVersion` and we can't use any artibrary string for source id for debugging.

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2899/

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -1076,14 +1076,21 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 			history = append(history, strings.Split(historyStr, ",")...)
 		}
 	} else {
-		versionVectorStr := rev
-		if historyStr != "" {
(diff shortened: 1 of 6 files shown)


torcolvin:
This PR is now quite out of date and will not be implemented as in this PR.

PR #7295, opened 2025-01-15
CBG-4309: update cbgt to address lack of janitor kick during rollback

gregns1:
Uptake a fix in cbgt from MB-64885 on a special branch of cbgt that contains this fix on top of v1.3.9 - https://github.com/couchbase/cbgt/tree/sg_v139

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2916/


rest/importtest/import_test.go
```diff
@@ -2316,6 +2316,127 @@ func TestImportRollback(t *testing.T) {
...
+		DatabaseConfig: &rest.DatabaseConfig{
+			DbConfig: rest.DbConfig{
+				ImportPartitions: base.Uint16Ptr(2),
+			},
+		},
+	})
+
+	// create doc id's for vb0 and vb800
+	vb0DocIDs := []string{"abbacomes", "abdicate", "accrescent", "aconitum", "acrux", "adduction", "affrication", "algraphy", "allantoinuria", "altiloquent"}
+	vb800DocIDs := []string{"abrook", "accept", "accompaniment", "acoemeti", "adiposeness", "alkyd", "alnage", "ambulance", "anasazi", "anhydroxime"}
```
torcolvin, line 2348: so there's no vb800 on mac, ideally you can pick keys that would match for 1024 vbuckets (non mac) or 64 buckets (mac)

base/dcp_sharded.go
```diff
@@ -337,6 +337,16 @@ func initCBGTManager(ctx context.Context, bucket Bucket, spec BucketSpec, cfgSG
 		sourceUUID:    bucketUUID,
 	}
 
+	// Define rollback hook to be called by cbgt during rollback. This is used to kick the janitor to reassign feeds
+	// after a full rollback of a pindex
+	cbgt.RollbackHook = func(phase cbgt.RollbackPhase, pindex *cbgt.PIndex) (err error) {
```
torcolvin: Since this modifies a _global_ in cbgt, can we move this to `init` where we add the other globals? https://github.com/couchbase/sync_gateway/blob/a68b162bba09dddbd333e2c4fda733f36d91488a/base/dcp_feed_type.go#L119

gregns1: I agree this makes sense but we need access to the manager to kick the janitor, which doesn't look like it can be done there?

torcolvin: Then we need to figure out how to run a function that passes in a manager for a specific bucket `initCBGTManager` is called once per bucket and then a second database would overwrite this callback for the first.

I think we might be able to do something with a wrapper function to look up mgr based on pindex and store this globally, or we could ask cbgt team to change their API so this isn't required? I don't know if they have one manager per instance, but that seems wacky.

torcolvin: Abandoned this approach for fix in cbgt.

adamcfraser (approved):

PR #7298, opened 2025-01-16
CBG-4468 do not check legacy revtree if hlv is dominating

torcolvin:
CBG-4468 do not check legacy revtree if hlv is dominating

It feels like there has to be a better way to express this in the code but I can not think of what it might be. Open to suggestions.

## Pre-review checklist

### DIFF
--- a/rest/blip_legacy_revid_test.go
+++ b/rest/blip_legacy_revid_test.go
@@ -984,3 +984,34 @@ func TestConflictBetweenPostUpgradeCBLMutationAndPostUpgradeSGWMutation(t *testi
 	assert.Equal(t, rev1ID, bucketDoc.CurrentRev)
 	assert.Equal(t, docVersion.CV.String(), bucketDoc.HLV.GetCurrentVersionString())
 }
+
+func TestLegacyRevNotInConflict(t *testing.T) {
+	base.SetUpTestLogging(t, base.LevelDebug, base.KeyHTTP, base.KeySync, base.KeySyncMsg, base.KeyCRUD)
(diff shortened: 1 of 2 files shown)


rest/blip_legacy_revid_test.go
```diff
@@ -984,3 +984,34 @@ func TestConflictBetweenPostUpgradeCBLMutationAndPostUpgradeSGWMutation(t *testi
...
+	sent, _, _, err := bt.SendRevWithHistory(docID, "100@CBL1", history, []byte(`{"key": "val"}`), blip.Properties{})
+	assert.True(t, sent)
+	require.NoError(t, err)
+
+	// assert that the bucket doc is as expected
+	bucketDoc, _, err := collection.GetDocWithXattrs(ctx, docID, db.DocUnmarshalAll)
+	require.NoError(t, err)
+	assert.Equal(t, "100@CBL1", bucketDoc.HLV.GetCurrentVersionString())
+	assert.NotNil(t, bucketDoc.History[rev1ID])
+	assert.Equal(t, docVersion.CV.Value, bucketDoc.HLV.PreviousVersions[docVersion.CV.SourceID])
```
adamcfraser, line 1020: Can we add a test step to make sure that the SGW still returns conflict if the document is updated on the server after 1-abc, prior to CBL sending their rev?

adamcfraser (changes requested):
Change looks reasonable, would like to make sure the actual conflict case is still treated as a conflict in a test.

adamcfraser (approved):

PR #7300, opened 2025-01-16
CBG-4469 Evaluate full HLV when available during proposeChanges

adamcfraser:
Until client-side check of HLV (using returned cv with conflict) is supported, use provided full HLV to identify whether proposed change is in conflict, regardless of previousRev.

CBG-4469

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -3211,7 +3211,7 @@ func (db *DatabaseCollectionWithUser) CheckProposedRev(ctx context.Context, doci
 }
 
 // CheckProposedVersion - given DocID and a version in string form, check whether it can be added without conflict.
-func (db *DatabaseCollectionWithUser) CheckProposedVersion(ctx context.Context, docid, proposedVersionStr string, previousRev string) (status ProposedRevStatus, currentVersion string) {
+func (db *DatabaseCollectionWithUser) CheckProposedVersion(ctx context.Context, docid, proposedVersionStr string, previousRev string, proposedHLVString string) (status ProposedRevStatus, currentVersion string) {
 
 	proposedVersion, err := ParseVersion(proposedVersionStr)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


db/crud.go
```diff
@@ -3211,7 +3211,7 @@ func (db *DatabaseCollectionWithUser) CheckProposedRev(ctx context.Context, doci
 }
 
 // CheckProposedVersion - given DocID and a version in string form, check whether it can be added without conflict.
-func (db *DatabaseCollectionWithUser) CheckProposedVersion(ctx context.Context, docid, proposedVersionStr string, previousRev string) (status ProposedRevStatus, currentVersion string) {
+func (db *DatabaseCollectionWithUser) CheckProposedVersion(ctx context.Context, docid, proposedVersionStr string, previousRev string, proposedHLVString string) (status ProposedRevStatus, currentVersion string) {
```
torcolvin, line 3217: For clarity, would changing this help?

Given that this is only called in one place, is it simpler to have this function take `proposedRev` instead of `proposedVersionStr` and have the beginning look like?

```
proposedVersion, err := ParseVersion(ExtractCVFromProposeChangesRev(rev))
```

Or document in the docstring what these args are, because then `proposedVersionStr` should actually be "proposedCV"?

db/crud.go
```diff
@@ -3259,7 +3259,20 @@ func (db *DatabaseCollectionWithUser) CheckProposedVersion(ctx context.Context,
 		// previousVersion didn't match, but proposed version and server CV have matching source, and proposed version is newer
 		return ProposedRev_OK, ""
 	} else {
-		// Conflict, return the current cv.  This may be a false positive conflict if the client has replicated
+		// Temporary (CBG-4466): check the full HLV that's being sent by CBL with proposeChanges messages.
+		// If the current server cv is dominated by the incoming HLV (i.e. the incoming HLV has an entry for the same source
+		// with a version that's greater than or equal to the server's cv), then we can accept the proposed version.
+		proposedHLV, _, err := ExtractHLVFromBlipMessage(proposedHLVString)
+		if err != nil {
+			base.DebugfCtx(ctx, base.KeyCRUD, "CheckProposedVersion for doc %s unable to extract proposedHLV from rev message, treating as conflict: %v", base.UD(docid), err)
```
torcolvin: If this happens, this actually seems more fatal than debug, this means that CBL violated the contract/protocol and we would want to know about it. I think we should log as much information as we can here, like `proposedHLVString`

db/crud.go
```diff
@@ -3259,7 +3259,20 @@ func (db *DatabaseCollectionWithUser) CheckProposedVersion(ctx context.Context,
...
+		} else {
+			if proposedHLV.DominatesSource(localDocCV) {
+				base.DebugfCtx(ctx, base.KeyCRUD, "CheckProposedVersion returning OK for doc %s because incoming HLV dominates cv", base.UD(docid))
+				return ProposedRev_OK, ""
+			}
+		}
```
torcolvin: nit: simplification

```suggestion
		} else if proposedHLV.DominatesSource(localDocCV) {
			base.DebugfCtx(ctx, base.KeyCRUD, "CheckProposedVersion returning OK for doc %s because incoming HLV dominates cv", base.UD(docid))
			return ProposedRev_OK, ""
		}
```

db/crud.go
```diff
@@ -3259,7 +3259,20 @@ func (db *DatabaseCollectionWithUser) CheckProposedVersion(ctx context.Context,
...
+
+		//Conflict, return the current cv.  This may be a false positive conflict if the client has replicated
```
torcolvin: I don't understand what this means exactly, can you note an example of the strings that would happen here, for me to figure out later?

torcolvin (approved):

PR #7302, opened 2025-01-16
CBG-4367: Retain minimum number of recent sequences during compaction

gregns1:
CBG-4367

- Retail minimum of 5 recent sequences during compaction of recent sequences in the sync data of a document

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2901/

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -26,7 +26,7 @@ import (
 
 const (
 	kMaxRecentSequences            = 20    // Maximum number of sequences stored in RecentSequences before pruning is triggered
-	kMinRecentSequences            = 5     // Minimum number of sequences stored in RecentSequences before pruning is triggered
+	kMinRecentSequences            = 5     // Minimum number of sequences that should be left stored in RecentSequences during compaction
 	unusedSequenceWarningThreshold = 10000 // Warn when releasing more than this many sequences due to existing sequence on the document
 )
 
--- a/db/crud.go
+++ b/db/crud.go
@@ -26,6 +26,7 @@ import (
 
 const (
 	kMaxRecentSequences            = 20    // Maximum number of sequences stored in RecentSequences before pruning is triggered
+	kMinRecentSequences            = 5     // Minimum number of sequences stored in RecentSequences before pruning is triggered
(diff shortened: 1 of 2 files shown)


db/crud.go
```diff
@@ -1712,13 +1713,19 @@ func (db *DatabaseContext) assignSequence(ctx context.Context, docSequence uint6
 		// so we're allowing more 'recent sequences' on the doc (20) before attempting pruning
 		stableSequence := db.changeCache.GetStableSequence(doc.ID).Seq
 		count := 0
+		// we want to keep at least kMinRecentSequences recent sequences in the recent sequences list to reduce likelihood
+		// races between compaction of resent sequences and a coalesced DCP mutation resulting in skipped/abandoned sequences
+		maxToCompact := len(doc.RecentSequences) - kMinRecentSequences
```
torcolvin, line 1718: This number can be negative, in which case we will end up compacting all sequences. Is it possible to write a unit test which would hit this and make sure that this function is robust to `len(doc.RecentSequences) < 5` ?

db/crud.go
```diff
@@ -1712,13 +1713,19 @@ func (db *DatabaseContext) assignSequence(ctx context.Context, docSequence uint6
...
 			// as the mutations that this node used to rev nextSequence will at some point be delivered
 			// to each node.
 			if seq < stableSequence {
 				count++
+				if count == maxToCompact {
+					break
+				}
 			} else {
 				break
 			}
```
torcolvin, line 1731: I wonder if this is easier to read, though you would want to do a check for len sequences outside of this function. You could create a function that _just_ prunes sequences to avoid longer logic and allow for early returns for len(doc.RecentSequences) < 5, etc.

Not set on this idea, just a thought on how to make it easier to read.

```suggestion
			if seq >= stableSequence {
				break
			}
			count++
			if count == maxCount {
				break
			}
```

db/crud.go
bbrks: We can't get here without

```go
if len(doc.RecentSequences) >= kMaxRecentSequences {
```

`maxToCompact` can't be negative

bbrks (approved):
I'm OK with this 👍 thanks!

PR #7303, opened 2025-01-16
CBG-4466 support mv

torcolvin:
CBG-4466 support mv

commit e9c4d1c571a07c6d3f91f80d3fa7e04a04087290 (first commit) is just rewriting the test so that I can make sense of the new format. I think the `createHLVForTest` is hazardous because it has its own shorthand syntax. Once we are happy with the blip message type, I think we could use that for constructing a string.

- Switched some tests to use small enough integers so that hex representation and decimal representation are the same for testing
- `TestPutRev4` used mv, but it had duplicate mv and pv values, so the test needed to be changed

Added:

- `TestBlipMergeVersions` which will round trip a merge version (working). This uses a rev message directly because there's not a great way with the `BlipTesterClient` to push a message with custom properties. This would be a future enhancement. It uses btcRunner to pull revisions. This test can be extended to support other features.

TODO:

- [x] Change `HybridLogicalVector.AddNewerVersions` to update mv. Cases: (1) move mv to pv (2) move existing mv to pv, and replace with incoming mv.   These are started in `TestHLVUpdateMV`.
- [ ] Run `BenchmarkExtractHLVFromBlipMessage` to see if extra validation to check for incorrect duplicate values in cv/mv/pv is a problem.
- [ ] Evaluate test cases for non rev messages and processing full hlv. Cases: (1) CBL response to changes message (with and without delta). (2) `proposeChanges`

## Pre-review checklist

### DIFF
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -239,6 +239,7 @@ func (hlv *HybridLogicalVector) AddVersion(newVersion Version) error {
 	if hlv.SourceID == "" {
 		hlv.Version = newVersion.Value
 		hlv.SourceID = newVersion.SourceID
+		hlv.InvalidateMV()
 		return nil
 	}
 	// if new entry has the same source we simple just update the version
@@ -247,6 +248,7 @@ func (hlv *HybridLogicalVector) AddVersion(newVersion Version) error {
 			return fmt.Errorf("attempting to add new version vector entry with a CAS that is less than the current version CAS value for the same source. Current cas: %d new cas %d", hlv.Version, newVersion.Value)
 		}
 		hlv.Version = newVersion.Value
+		hlv.InvalidateMV()
 		return nil
 	}
 	// if we get here this is a new version from a different sourceID thus need to move current sourceID to previous versions and update current version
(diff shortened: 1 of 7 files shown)


db/hybrid_logical_vector.go
```diff
@@ -383,6 +404,14 @@ func (hlv *HybridLogicalVector) setPreviousVersion(source string, version uint64
 	hlv.PreviousVersions[source] = version
 }
 
+// setMergeVersion will take a source/version pair and add it to the HLV merge versions map
+func (hlv *HybridLogicalVector) setMergeVersion(source string, version uint64) {
+	if hlv.MergeVersions == nil {
```
torcolvin, line 409: I am not entirely sure why we don't require initialization of the maps and using constructor `NewHybridLogicalVector`. I wasn't sure if it was for perf reasons so I left it.

Is there a reason to have `setMergeVersion` in a loop rather than just assign the incoming merge versions to the existing hlv?

db/utilities_hlv_testing.go
```diff
@@ -183,6 +185,7 @@ func EncodeTestHistory(historyString string) (encodedString string) {
 			if index > 0 {
 				encodedString += ","
 			}
+			fmt.Printf("version: %s encoded: %s\n", version, EncodeTestVersion(version))
```
torcolvin:
```suggestion
```
I left this debugging in by mistake.

db/hybrid_logical_vector_test.go
```diff
@@ -827,3 +1065,199 @@ func TestVersionDeltaCalculation(t *testing.T) {
...
+				MergeVersions: map[string]uint64{
+					"a": 1,
+					"b": 2,
+				},
+			},
+			incomingHLV: HybridLogicalVector{
+				Version:  4,
+				SourceID: "c",
+			},
+			// does not update with incoming versions?
```
torcolvin: is this comment correct?

db/hybrid_logical_vector.go
adamcfraser: I agree it seems inconsistent, but don't want to open that can of worms at this point.  Leaving setMergeVersion as a single entry API to match setPreviousVersion and for potential future usage.

adamcfraser (approved):

PR #7305, opened 2025-01-17
CBG-4422: move setting the collection set on resync status to init fu…

gregns1:
CBG-4422

- move setting the collection set on resync status to init function to avoid delay between status running and collection set appearing in status
- create a new struct to hold collection information needed in Run() function 

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2902/

### DIFF
--- a/db/background_mgr_resync_dcp.go
+++ b/db/background_mgr_resync_dcp.go
@@ -32,7 +32,14 @@ type ResyncManagerDCP struct {
 	VBUUIDs             []uint64
 	useXattrs           bool
 	ResyncedCollections map[string][]string
-	lock                sync.RWMutex
+	resyncCollectionInfo
+	lock sync.RWMutex
+}
+
+// resyncCollectionInfo contains information on collections included on resync run, populated in init() and used in Run()
+type resyncCollectionInfo struct {
+	hasAllCollections bool
+	collectionIDs     []uint32
 }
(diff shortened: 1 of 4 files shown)


db/background_mgr_resync_dcp.go
```diff
@@ -65,6 +75,16 @@ func (r *ResyncManagerDCP) Init(ctx context.Context, options map[string]interfac
 		return nil
 	}
 
+	// Get collectionIds and store in manager for use in DCP client later
```
torcolvin, line 78: I think we might only want to set this in `newRunInit`. I don't think this will change from run to run but we don't need to do this calculation.

gregns1: It shouldn't change from run to run but it could (currently we don't handle that). I filed CBG-4473 for this last week.

My thinking of putting this here is for when CBG-4473 is done it will be easy to find a difference between the passed in collections through the endpoint and the previous interrupted run. In this case we should fail and force the user to reset resync. 
Example:
- Run resync with collection A 
- Stop resync 
- Call action=start again it with payload of Collection A and B 
- Here we should fail and force user to reset but we don't currently

torcolvin (approved):

PR #7306, opened 2025-01-17
CBG-4474 Fix for MV invalidation ordering

adamcfraser:
Ensure that MV is invalidated before cv is moved to pv.

Created more extensive testing for AddVersion, and refactored AddVersion to reduce complexity based on the additional test coverage.

CBG-4474

### DIFF
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -230,52 +230,40 @@ func (hlv *HybridLogicalVector) DominatesSource(version Version) bool {
 
 // AddVersion adds newVersion as the current version to the in memory representation of the HLV.
 func (hlv *HybridLogicalVector) AddVersion(newVersion Version) error {
-	var newVersionCAS uint64
-	hlvVersionCAS := hlv.Version
-	if newVersion.Value != expandMacroCASValueUint64 {
-		newVersionCAS = newVersion.Value
-	}
+
 	// check if this is the first time we're adding a source - version pair
 	if hlv.SourceID == "" {
 		hlv.Version = newVersion.Value
 		hlv.SourceID = newVersion.SourceID
-		hlv.InvalidateMV()
 		return nil
 	}
-	// if new entry has the same source we simple just update the version
+
+	// If the new version is older than an existing version for the same source, return error
+	existingValueForSource, found := hlv.GetValue(newVersion.SourceID)
+	if found && existingValueForSource > newVersion.Value {
+		return fmt.Errorf("attempting to add new version vector entry with a value that is less than the existing value for the same source. New version: %v, Existing HLV: %v", newVersion, hlv)
+	}
+
+	// Move existing mv to pv before adding the new version
+	hlv.InvalidateMV()
+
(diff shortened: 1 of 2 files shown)


db/hybrid_logical_vector.go
```diff
@@ -230,52 +230,40 @@ func (hlv *HybridLogicalVector) DominatesSource(version Version) bool {
...
-	// if new entry has the same source we simple just update the version
+
+	// If the new version is older than an existing version for the same source, return error
+	existingValueForSource, found := hlv.GetValue(newVersion.SourceID)
+	if found && existingValueForSource > newVersion.Value {
+		return fmt.Errorf("attempting to add new version vector entry with a value that is less than the existing value for the same source. New version: %v, Existing HLV: %v", newVersion, hlv)
+	}
+
+	// Move existing mv to pv before adding the new version
+	hlv.InvalidateMV()
```
torcolvin, line 248: The comment in `InvalidateMV` is wrong:

```
// InvalidateMV will move all merge versions to PV, except merge version entries that share a source with cv
```

I think this should be:
```
// InvalidateMV will move all merge versions to previous versions. This function should be called before placing current version in previous versions, since a cv will override an mv.
```

db/hybrid_logical_vector.go
```diff
@@ -230,52 +230,40 @@ func (hlv *HybridLogicalVector) DominatesSource(version Version) bool {
...
-	} else {
-		// source doesn't exist in PV so add
-		hlv.PreviousVersions[hlv.SourceID] = hlv.Version
-	}
 
-	// If new version already exists PV, need to remove it
-	delete(hlv.PreviousVersions, newVersion.SourceID)
+	hlv.PreviousVersions[hlv.SourceID] = hlv.Version
 
+	// If new version source already existed in PV, need to remove it
```
torcolvin, line 263: Is there a circumstance (I guess this would be a bug) where

existing hlv:

```
cv: 1@a
mv: 2@a, 3@b
```

incoming hlv: `3@c`

and then the new hlv would be:

```
cv: 3@c
pv: 2@a, 3@b
```

or do we actually always want to just pick the latest? I think either way you decide we should add a test so we can ensure we aren't changing behavior.

Or are we assuming that the existing HLV could never happen? I worry that it might happen if it was written by a function that is not this one or replicated with xdcr.

db/hybrid_logical_vector.go
adamcfraser: I think the first version is actually accurate? (merge versions for the source cv will not be moved to pv)

db/hybrid_logical_vector.go
adamcfraser: Yes, that existing HLV is invalid.  We don't currently have any handling to repair or recover from invalid HLVs.  It's something that we could consider in a future enhancement, perhaps.

db/hybrid_logical_vector.go
torcolvin: I misread this function, looks good.

torcolvin (approved):

PR #7308, opened 2025-01-20
CBG-4455 do not panic when calling `FailNow`

torcolvin:
CBG-4455 do not panic when calling `FailNow`

Confusingly, require.FailNow's signature is `FailNow(t TestingT, failureMessage string, msgAndArgs ...interface{})`. There are a few consequences to this, one of which is that expecting `failureMessage` to be formatted will not happen. The second problem is if the first argument in `msgAndArgs` is not a format string, then the test harness can panic.

- Created a ruleguard rule to find the misuse of this. This prevents use of % followed by an letter but most "real" uses of % are followed by a space, number, or punctuation. Forcing passing to fmt.Sprintf will cause the compiler to error if the wrong argument types are passed. Updated ruleguard rules in linter to account for new checks, and changed the string to only pick up `rules-*.go` files so the test files in that directory aren't picked up.
- Fixed existing misuse.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2898/ (known, unrelated failure)


bbrks (dismissed):

.golangci-strict.yml
```diff
@@ -34,37 +34,36 @@ linters:
...
-    - wsl # Whitespace Linter - Forces you to use empty lines!
+    # - asciicheck # Simple linter to check that your code does not contain non-ASCII identifiers
+    # - depguard # Go linter that checks if package imports are in a list of acceptable packages
+    # - dogsled # Checks assignments with too many blank identifiers # (e.g. x, _, _, _, := f())
+    # - err113 # Golang linter to check the errors handling expressions
+    # - funlen # Tool for detection of long functions
+    # - gochecknoglobals # Checks that no globals are present in Go code
+    # - gochecknoinits # Checks that no init functions are present in Go code
+    # - gocognit # Computes and checks the cognitive complexity of functions
+    # - gocyclo # Computes and checks the cyclomatic complexity of functions
```
bbrks: There's a bunch of commits being pulled into this PR that are already in `main` - it looks like a bad rebase to me

bbrks (changes requested):

bbrks (approved):

PR #7309, opened 2025-01-20
CBG-4312 improve metrics documentation

torcolvin:
CBG-4312

- remove /_metrics endpoint, which does not need public API documentation, as it never should have been implemented
- change copy to match public docs for description
- add links to upstream files

This upstreams changes done in https://github.com/couchbase/docs-sync-gateway/pull/839/files

@simon-dew, I did not make the same changes in the files https://github.com/couchbase/sync_gateway/blob/main/docs/api/metric-capella.yaml and https://github.com/couchbase/sync_gateway/blob/main/docs/api/admin-capella.yaml . I am happy to make the corresponding changes in these files or we can wait until you figure out what the appropriate `externalDocs` sections are.

Can you verify that I ported all the changes? I did not port the line `/_metrics` because I think it is wrong, we can wait until https://github.com/couchbase/docs-sync-gateway/pull/839/files is resolved.

```
          application/json:
            schema:
              $ref: metrics_metadata.json
```

## Pre-review checklist


docs/api/metric.yaml
```diff
@@ -30,15 +34,17 @@ servers:
 paths:
   /_ping:
     $ref: ./paths/common/_ping.yaml
-  /_metrics:
-    $ref: ./paths/metric/metrics.yaml
```
bbrks: Isn't this what `x-internal` was for? I'm not sure how I feel about the API becoming only partially documented based on internal/external opinions again, rather than a comprehensive spec of the API SG is actually serving.

torcolvin: Good catch.

docs/api/paths/metric/metrics.yaml
simon-dew, line 1: Do you want to update the summary of the `get_metrics` operation to match `get__expvar`?

> Get debugging and monitoring runtime stats in Prometheus Exposition format

docs/api/paths/metric/_expvar.yaml
```diff
@@ -18,9 +18,13 @@ get:
...
+  tags:
+    - JSON
   responses:
     '200':
-      description: Returned statistics
+      description: |-
+          Successfully returned statistics.
+          For details, see [JSON Metrics](stats-monitoring-json.html).
       content:
         application/javascript:
```
simon-dew: Is this actually correct, or should the response be `application/json`?

docs/api/admin.yaml
simon-dew, line 1: While we are here, do we want to add natural language summaries for all the operations in the Admin REST API which don't currently have them? For example, `head_keyspace-_changes`, `post_db-_ensure_full_commit`.

simon-dew (changes requested):
Looks like you have incorporated almost all of the changes from https://github.com/couchbase/docs-sync-gateway/pull/839 👍 

I have a couple of comments though:

docs/api/admin.yaml
torcolvin: https://jira.issues.couchbase.com/browse/CBG-4503 I'm putting this in a separate ticket because it is confusing and tricky.

simon-dew: Fair enough!

bbrks (approved):

simon-dew (approved):

PR #7326, opened 2025-01-28
CBG-4487: Avoid deadlock in revision cache between Get and On demand import

gregns1:
CBG-4487

- Refactor eviction to not acquire value lock whilst hold rev cache lock 
- made item bytes atomic 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2920/

### DIFF
--- a/db/revision_cache_lru.go
+++ b/db/revision_cache_lru.go
@@ -14,6 +14,7 @@ import (
 	"container/list"
 	"context"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	sgbucket "github.com/couchbase/sg-bucket"
@@ -108,7 +109,7 @@ type revCacheValue struct {
 	lock        sync.RWMutex
 	deleted     bool
 	removed     bool
-	itemBytes   int64
+	itemBytes   atomic.Int64
 }
 
 // Creates a revision cache with the given capacity and an optional loader function.
@@ -265,13 +266,13 @@ func (rc *LRURevisionCache) Upsert(ctx context.Context, docRev DocumentRevision,
 	}
 
 	// Purge oldest item if over number capacity
-	var numItemsRemoved int
-	for len(rc.cache) > int(rc.capacity) {
-		rc.purgeOldest_()
-		numItemsRemoved++
+	numItemsRemoved, numBytesEvicted := rc._numberCapacityEviction()
+	if numBytesEvicted > 0 {
+		rc.updateRevCacheMemoryUsage(-numBytesEvicted)
 	}
+	rc.lock.Unlock() // release lock after eviction finished
 	if numItemsRemoved > 0 {
-		rc.cacheNumItems.Add(int64(-numItemsRemoved))
+		rc.cacheNumItems.Add(-numItemsRemoved)
 	}
 
 	docRev.CalculateBytes()
@@ -280,16 +281,7 @@ func (rc *LRURevisionCache) Upsert(ctx context.Context, docRev DocumentRevision,
 	value.store(docRev)
 
 	// check we aren't over memory capacity, if so perform eviction
-	numItemsRemoved = 0
-	if rc.memoryCapacity > 0 {
-		for rc.currMemoryUsage.Value() > rc.memoryCapacity {
-			rc.purgeOldest_()
-			numItemsRemoved++
-		}
-		rc.cacheNumItems.Add(int64(-numItemsRemoved))
-	}
-
-	rc.lock.Unlock()
+	rc.revCacheMemoryBasedEviction()
 }
 
 func (rc *LRURevisionCache) getValue(docID, revID string, collectionID uint32, create bool) (value *revCacheValue) {
@@ -307,15 +299,16 @@ func (rc *LRURevisionCache) getValue(docID, revID string, collectionID uint32, c
 		rc.cache[key] = rc.lruList.PushFront(value)
 		rc.cacheNumItems.Add(1)
 
-		// evict if over number capacity
-		var numItemsRemoved int
-		for len(rc.cache) > int(rc.capacity) {
-			rc.purgeOldest_()
-			numItemsRemoved++
+		numItemsRemoved, numBytesEvicted := rc._numberCapacityEviction()
+		if numBytesEvicted > 0 {
+			rc.updateRevCacheMemoryUsage(-numBytesEvicted)
 		}
+		rc.lock.Unlock() // release lock as eviction is finished
 		if numItemsRemoved > 0 {
-			rc.cacheNumItems.Add(int64(-numItemsRemoved))
+			rc.cacheNumItems.Add(-numItemsRemoved)
 		}
+		// return early as rev cache mutex has been released at this point
+		return
 	}
 	rc.lock.Unlock()
 	return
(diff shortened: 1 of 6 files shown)


db/revision_cache_lru.go
```diff
@@ -265,13 +266,13 @@ func (rc *LRURevisionCache) Upsert(ctx context.Context, docRev DocumentRevision,
...
 	// Purge oldest item if over number capacity
-	var numItemsRemoved int
-	for len(rc.cache) > int(rc.capacity) {
-		rc.purgeOldest_()
-		numItemsRemoved++
+	numItemsRemoved, numBytesEvicted := rc._numberCapacityEviction()
+	if numBytesEvicted > 0 {
+		rc.updateRevCacheMemoryUsage(-numBytesEvicted)
 	}
+	rc.lock.Unlock() // release lock after eviction finished
```
torcolvin, line 274: rc.updateRevCacheMemoryUsage doesn't require a lock, it updates atomics, so we can move this up.

db/revision_cache_lru.go
```diff
@@ -307,15 +299,16 @@ func (rc *LRURevisionCache) getValue(docID, revID string, collectionID uint32, c
...
-		// evict if over number capacity
-		var numItemsRemoved int
-		for len(rc.cache) > int(rc.capacity) {
-			rc.purgeOldest_()
-			numItemsRemoved++
+		numItemsRemoved, numBytesEvicted := rc._numberCapacityEviction()
+		if numBytesEvicted > 0 {
+			rc.updateRevCacheMemoryUsage(-numBytesEvicted)
 		}
+		rc.lock.Unlock() // release lock as eviction is finished
```
torcolvin, line 307: As above, we can move unlock since `updateRevCacheMemoryUsage` doesn't require a lock

db/revision_cache_lru.go
```diff
@@ -349,11 +342,16 @@ func (rc *LRURevisionCache) removeValue(value *revCacheValue) {
 	rc.lock.Unlock()
 }
 
-func (rc *LRURevisionCache) purgeOldest_() {
-	value := rc.lruList.Remove(rc.lruList.Back()).(*revCacheValue)
-	delete(rc.cache, value.key)
-	// decrement memory overall size
-	rc.updateRevCacheMemoryUsage(-value.getItemBytes())
+// _numberCapacityEviction will iterate removing the last element in cache til we fall below the maximum number of items
+// threshold for this shard, retuning the bytes evicted and number of items evicted
```
torcolvin, line 347:
```suggestion
// threshold for this shard, returning the bytes evicted and number of items evicted. Holding LRURevisionCache.lock is required.
```

db/revision_cache_lru.go
```diff
@@ -539,13 +535,22 @@ func (rc *LRURevisionCache) revCacheMemoryBasedEviction() {
...
+		for bytesNeededToRemove > numBytesRemoved {
+			value := rc.lruList.Remove(rc.lruList.Back()).(*revCacheValue)
+			delete(rc.cache, value.key)
+			numItemsRemoved++
+			valueBytes := value.getItemBytes()
+			numBytesRemoved += valueBytes
+		}
 	}
+	rc.updateRevCacheMemoryUsage(-numBytesRemoved) // need update rev cache memory stats before release lock to stop other goroutines evicting based on outdated stats
+	rc.lock.Unlock()                               // release lock after removing items from cache
```
torcolvin: We can move the lock above updateRevCacheMemory usage.

db/revision_cache_lru.go
gregns1: Issue is if we don't update whilst holding lock and another goroutine is waiting on acquiring lock for eviction once this is released the other goroutine newly acquiring lock will be acting on incorrect memory stats and will over evict.

db/revision_cache_lru.go
```diff
@@ -539,13 +535,22 @@ func (rc *LRURevisionCache) revCacheMemoryBasedEviction() {
...
-	var numItemsRemoved int64
-	for rc.currMemoryUsage.Value() > rc.memoryCapacity {
-		rc.purgeOldest_()
-		numItemsRemoved++
+	var numItemsRemoved, numBytesRemoved int64
+	rc.lock.Lock() // hold rev cache lock to remove items from cache until we're below memory threshold for the shard
+	// check if we are over memory capacity after holding rev cache mutex (protect against another goroutine evicting whilst waiting for mutex above)
+	if rc.currMemoryUsage.Value() > rc.memoryCapacity {
+		// find amount of bytes needed to evict till below threshold
+		bytesNeededToRemove := rc.currMemoryUsage.Value() - rc.memoryCapacity
```
bbrks: Just avoids a duplicate atomic load

```suggestion
	if currMemoryUsage := rc.currMemoryUsage.Value(); currMemoryUsage > rc.memoryCapacity {
		// find amount of bytes needed to evict till below threshold
		bytesNeededToRemove := currMemoryUsage - rc.memoryCapacity
```

db/revision_cache_lru.go
bbrks: It kinda looks like everywhere we decrement this we need to be holding the lock, and everywhere we increment we don't. Is that right? Makes me wonder if they should be separate functions with better indication of which lock is required.

bbrks (changes requested):
Will take another pass tomorrow to double check locking and sync up on locking during `updateRevCacheMemoryUsage`

db/revision_cache_lru.go
```diff
@@ -152,7 +153,8 @@ func (rc *LRURevisionCache) UpdateDelta(ctx context.Context, docID, revID string
 	if value != nil {
 		outGoingBytes := value.updateDelta(toDelta)
 		if outGoingBytes != 0 {
-			rc.updateRevCacheMemoryUsage(outGoingBytes)
+			rc.currMemoryUsage.Add(outGoingBytes)
+			rc.cacheMemoryBytesStat.Add(outGoingBytes)
```
bbrks, line 157: I think we can just do this here, right?

```suggestion
			rc.incrRevCacheMemoryUsage(outGoingBytes)
```

db/revision_cache_lru.go
```diff
@@ -539,22 +536,51 @@ func (rc *LRURevisionCache) revCacheMemoryBasedEviction() {
...
+// You should be holding rev cache lock in using this function to avoid eviction processes over evicting items
+func (rc *LRURevisionCache) _decrRevCacheMemoryUsage(bytesCount int64) {
+	// We need to keep track of the current LRURevisionCache memory usage AND the overall usage of the cache. We need
+	// overall memory usage for the stat added to show rev cache usage plus we need the current rev cache capacity of the
+	// LRURevisionCache object for sharding the rev cache. This way we can perform eviction on per shard basis much like
+	// we do with the number of items capacity eviction
+	if bytesCount > 0 {
+		// function is for decrementing memory usage, so return if incrementing
+		return
+	}
```
bbrks: I like that there's a guard here and in the Incr function, but I worry about us hitting them silently.
It seems really easy to mistakenly pass in a positive number into a decr function (arguably this is how it should work?)

We should probably be using `AssertfCtx` before the return, which will panic in dev mode, and emit a warning in prod code, since it's a coding error and needs to be addressed.

bbrks (changes requested):
A few nits, otherwise I'm OK with this

db/revision_cache_lru.go
gregns1: My predicament here is `outGoingBytes` could be incrementing bytes count or decrementing it depending on how the updating the delta can effect the overall memory footprint of the value.

bbrks: Ah gotcha - didn't realise outGoingBytes could be positive or negative.

bbrks (approved):

PR #7328, opened 2025-01-28
CBG-4494 do not put documents into rev cache for an on demand import

torcolvin:
CBG-4494 do not put documents into rev cache when they are being loaded

Used the same test as original bug.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2917/

### DIFF
--- a/db/import.go
+++ b/db/import.go
@@ -32,7 +32,7 @@ const (
 )
 
 // Imports a document that was written by someone other than sync gateway, given the existing state of the doc in raw bytes
-func (db *DatabaseCollectionWithUser) ImportDocRaw(ctx context.Context, docid string, value []byte, xattrs map[string][]byte, isDelete bool, cas uint64, expiry *uint32, mode ImportMode, updateRevCache bool) (docOut *Document, err error) {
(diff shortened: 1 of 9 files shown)


db/import.go
```diff
@@ -32,7 +32,7 @@ const (
 )
 
 // Imports a document that was written by someone other than sync gateway, given the existing state of the doc in raw bytes
-func (db *DatabaseCollectionWithUser) ImportDocRaw(ctx context.Context, docid string, value []byte, xattrs map[string][]byte, isDelete bool, cas uint64, expiry *uint32, mode ImportMode) (docOut *Document, err error) {
+func (db *DatabaseCollectionWithUser) ImportDocRaw(ctx context.Context, docid string, value []byte, xattrs map[string][]byte, isDelete bool, cas uint64, expiry *uint32, mode ImportMode, updateRevCache bool) (docOut *Document, err error) {
```
adamcfraser: I think ImportMode (already present on the signature) is sufficient to determine the `updateRevCache` value?

adamcfraser (commented):
Approach looks good, comment on one place I think we can use existing information.

db/import.go
torcolvin: Good catch.

adamcfraser (approved):

PR #7335, opened 2025-01-30
CBG-4499: fix leaking goroutines

gregns1:
CBG-4499

Fix leaking goroutines likely to cause test timeouts we saw previously.

## Pre-review checklist

### DIFF
--- a/db/revision_cache_test.go
+++ b/db/revision_cache_test.go
@@ -1486,7 +1486,7 @@ func TestRevCacheOnDemandImport(t *testing.T) {
 	revID, _, err := collection.Put(ctx, docID, Body{"ver": "1"})
 	require.NoError(t, err)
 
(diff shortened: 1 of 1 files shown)


db/revision_cache_test.go
```diff
@@ -1488,13 +1485,22 @@ func TestRevCacheOnDemandImport(t *testing.T) {
 	docID := "doc1"
 	revID, _, err := collection.Put(ctx, docID, Body{"ver": "1"})
 	require.NoError(t, err)
+
+	testCtx, testCtxCancel := context.WithCancel(base.TestCtx(t))
```
torcolvin:
```suggestion
	ctx, testCtxCancel := context.WithCancel(ctx)
```

I think it makes conceptual sense to me to not worry about which context to use then we can always just use the one known by `ctx` if we expand the test?

bbrks (approved):

PR #7338, opened 2025-01-30
CBG-4450: test flake where nanoseconds are the same in the config between updates

gregns1:
CBG-4450

- Add sleep to ensure nanosecond difference between config updates
- Particularly flakes on windows, as result re ran CI 3 times for windows - no flake

## Pre-review checklist

### DIFF
--- a/rest/replicatortest/replicator_test.go
+++ b/rest/replicatortest/replicator_test.go
@@ -8681,6 +8681,9 @@ func TestReplicationConfigUpdatedAt(t *testing.T) {
 	currTime := configResponse.UpdatedAt
 	createdAtTime := configResponse.CreatedAt
 
+	// avoid flake where update at seems to be the same (possibly running to fast)
+	time.Sleep(500 * time.Nanosecond)
+
 	resp = activeRT.SendAdminRequest("PUT", "/{{.db}}/_replicationStatus/replication1?action=stop", "")
 	rest.RequireStatus(t, resp, http.StatusOK)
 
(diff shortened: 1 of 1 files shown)


rest/replicatortest/replicator_test.go
```diff
@@ -8681,6 +8681,9 @@ func TestReplicationConfigUpdatedAt(t *testing.T) {
 	currTime := configResponse.UpdatedAt
 	createdAtTime := configResponse.CreatedAt
 
+	// avoid flake where update at seems to be the same (possibly running to fast)
+	time.Sleep(500 * time.Nanosecond)
```
bbrks: We should bump this more - we can tolerate a few extra milliseconds of test time no problem

```suggestion
	// avoid flake where update at seems to be the same (possibly running to fast)
	time.Sleep(10 * time.Millisecond)
```

rest/replicatortest/replicator_test.go
```diff
@@ -8681,6 +8681,9 @@ func TestReplicationConfigUpdatedAt(t *testing.T) {
...
+
 	resp = activeRT.SendAdminRequest("PUT", "/{{.db}}/_replicationStatus/replication1?action=stop", "")
 	rest.RequireStatus(t, resp, http.StatusOK)
```
bbrks, line 8689: Down below where we do `assert.Greater` we can use `AssertTimeGreaterThan` - it has built in handling for this particular flavour of Windows-specific failure to tolerate equal times even if the sleep does fail.

If we wrote a bug around this where we never update timestamp, it'd show up on other OSes

bbrks (dismissed):

rest/replicatortest/replicator_test.go
torcolvin: Since this is only on windows that it fails IIRC, maybe we should not sleep, because I don't know if it will help? But I agree about this timestamp thing.

torcolvin (approved):

torcolvin (approved):

torcolvin (approved):

PR #7345, opened 2025-01-31
CBG-4390: provide cluster UUID at top level endpoint

gregns1:
CBG-4390

- Add cluster uuid to runtime server config response 
- If there are db's defined we can just grab off a db context, else we need to request it from couchbase server. 

Screenshot 2025-02-06 at 12 14 39

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2940/


rest/admin_api.go
```diff
@@ -409,6 +414,29 @@ func (h *handler) handleGetConfig() error {
...
+			bucketCfg := getBucketConfigFromBoostrap(h.server.Config.Bootstrap)
+			spec, err := GetBucketSpec(h.ctx(), &DatabaseConfig{
+				DbConfig: DbConfig{
+					BucketConfig: bucketCfg,
+				},
+			}, h.server.Config)
+			if err != nil {
+				return err
+			}
+			clusterUUID, err = h.server.BootstrapContext.Connection.GetClusterUUID(h.ctx(), spec)
```
torcolvin: I think if persistent config is diable that this bootstrap connection might not exist?

base/bootstrap.go
```diff
@@ -282,6 +285,31 @@ func (cc *CouchbaseCluster) GetConfigBuckets() ([]string, error) {
...
+		return "", err
+	}
+	randBucket := lits[rand.Intn(len(lits))]
+	b, teardown, err := cc.getBucket(ctx, randBucket)
+	if err != nil {
+		return "", err
+	}
+	defer teardown()
+
+	// construct what's needed for GocbV2Bucket to fetch cluster UUID
```
torcolvin: Just move this function https://github.com/couchbase/sync_gateway/blob/main/db/database.go#L344 to base package?

torcolvin: But we have an agent here, so I think maybe moving the function to not require a bucket would be great. It just needs a gocbcore.Agent to get mgmt endpoint

 https://github.com/couchbase/sync_gateway/blob/8a9c7f27bf4b527199552eb9466902568905ccdd/base/collection.go#L450
 
 Then you can have a way to get mgmteps from either couchbase cluster or bucket.

rest/config_test.go
```diff
@@ -3240,3 +3240,57 @@ func TestRoleUpdatedAtField(t *testing.T) {
 	assert.Greater(t, newTime.UnixNano(), currTime.UnixNano())
 	assert.Equal(t, timeCreated.UnixNano(), newCreated.UnixNano())
 }
+
+func TestServerUUIDRuntimeServerConfig(t *testing.T) {
+	rt := NewRestTester(t, &RestTesterConfig{PersistentConfig: true})
```
torcolvin: we should parameterize this test for persistent config and not

rest/admin_api.go
```diff
@@ -409,6 +415,25 @@ func (h *handler) handleGetConfig() error {
 			}
 		}
 
+		// grab cluster uuid for runtime config
+		var clusterUUID string
+		if !base.ServerIsWalrus(h.server.Config.Bootstrap.Server) {
```
torcolvin: I'd probably implement a `CouchbaseCluster.GetClusterUUID` and `RosmarCluster.UUID`, which I think would be more readable. That would eliminate needing to reconstruct `getBucketConfigFromBoostrap` 

Is `h.server.Config.Boostrap` actually defined if sync gateway is started without persistent config? I think you can have non persistent config without any databases as well, I'm not sure this is easily tested with `RestTester` but you can probably test with with non persistent config + remove database. If this isn't easily testable like that, you can consider dropping the test for this case, but you should make sure that non persistent config without databases works because I do not think SG validates you necessarily have a database, even in non persistent config. https://github.com/couchbase/sync_gateway/blob/b77c0b493aea1f619023a2f3ea03945f015dc177/rest/server_context.go#L118

base/gocb_utils.go
```diff
@@ -161,3 +163,36 @@ func getRootCAs(ctx context.Context, caCertPath string) (*x509.CertPool, error)
 	}
 	return rootCAs, nil
 }
+
+func GetServerUUIDWithAgent(a *gocbcore.Agent, mgmtEp, username, password string) (string, error) {
```
torcolvin: I'd probably create a common function name with `MgmtRequest` with this call string and have `GocbV2Bucket.MgmtRequest` call into this.

Then you can replace reuse this in https://github.com/couchbase/sync_gateway/blob/b77c0b493aea1f619023a2f3ea03945f015dc177/base/bucket.go#L375.

rest/admin_api.go
gregns1: I originally took this approach but it has issue with legacy config. There is no couchbase cluster on bootstrap context when running in legacy config and you have no db running.  So If you startup in legacy config, delete db then call this endpoint when trying to access couchbase cluster it will panic. 

Bootstrap credentials are filled out in bootstrap as far as I can tell as you still need creds to bootstrap to a bucket for the db to run against. This is done here -> https://github.com/couchbase/sync_gateway/blob/b77c0b493aea1f619023a2f3ea03945f015dc177/rest/config_legacy.go#L154

My test already runs with a db defined then deletes the db then calls the endpoint to retrieve the cluster UUID and it passes fine in that respect. I already did also do some testing locally to see if you can startup with no db defined in legacy config but could not get that to work. I also cannot see anywhere in legacy config struct where you can define this information -> https://github.com/couchbase/sync_gateway/blob/b77c0b493aea1f619023a2f3ea03945f015dc177/rest/config_legacy.go#L28 I also couldn't see an example of this in our examples folder in the repo.

rest/admin_api.go
```diff
@@ -352,9 +352,14 @@ func (h *handler) handleGetDbConfig() error {
 
 type RunTimeServerConfigResponse struct {
 	*StartupConfig
+	RuntimeClusterUUID
```
bbrks: I'd like to put this in a nested "runtime" struct in the response if possible - somewhere we can add more runtime-only info in future without interfering or being lost inside the actual `StartupConfig` properties.

I'm thinking of things like post-bootstrap server addresses (since we can bootstrap from a single node and get many servers back), server node latency information, and even non-server runtime info (uptime, command line flags, host timezone, etc.)

bbrks (changes requested):
Just a comment about how we populate the API response for extensibility - otherwise looks good

rest/admin_api.go
gregns1: Good shout, updated screenshot in description to reflect the change.

bbrks (approved):

PR #7351, opened 2025-02-04
CBG-4371 Make sure not to redact profiles

torcolvin:
CBG-4371 Make sure not to redact profiles

With -redact set, `pprof*.gz` files would be redacted with `add_gzip_file_task(sourcefile_path=filename, salt=salt)`. This meant that when the files get opened in text mode https://github.com/couchbase/sync_gateway/blob/b77c0b493aea1f619023a2f3ea03945f015dc177/tools/tasks.py#L534 then might be doing something unwanted to gzipped contents as well as then trying to do a regex replacement on it https://github.com/couchbase/sync_gateway/blob/b77c0b493aea1f619023a2f3ea03945f015dc177/tools/tasks.py#L128

Since I centralized the code to collect log files, run `pathlib.Path.resolve()` on the path of the file to normalize a path from duplicate slashes or same directories. It is possible to add a path with duplicate names with `log_file_path` and also in `/home/sync_gateway/logs`. 

This code is quite well tested by existing test `test_make_collect_logs_tasks`

### DIFF
--- a/tools-tests/parser_test.py
+++ b/tools-tests/parser_test.py
@@ -16,9 +16,9 @@ def test_parser():
 def test_parser_log_redaction_salt():
     parser = sgcollect_info.create_option_parser()
     options, _ = parser.parse_args(["--log-redaction-salt=a", "foo.zip"])
-    assert options.salt_value == b"a"
+    assert options.salt_value == "a"
 
     options, _ = parser.parse_args(["foo.zip"])
-    # assert this is bytes uuid4
-    assert isinstance(options.salt_value, bytes)
-    assert len(options.salt_value) == 16
+    # assert this is a str repr of uuid4
+    assert isinstance(options.salt_value, str)
+    assert len(options.salt_value) == 36
--- a/tools-tests/parser_test.py
+++ b/tools-tests/parser_test.py
@@ -11,3 +11,20 @@ import sgcollect_info
 
 def test_parser():
     sgcollect_info.create_option_parser()
+
+
+def test_parser_log_redaction_salt():
+    parser = sgcollect_info.create_option_parser()
+    options, _ = parser.parse_args(["--log-redaction-salt=a", "foo.zip"])
+    assert options.salt_value == b"a"
+
+    options, _ = parser.parse_args(["foo.zip"])
+    # assert this is bytes uuid4
+    assert isinstance(options.salt_value, bytes)
+    assert len(options.salt_value) == 16
+
+
+def test_parser_sg_url():
+    parser = sgcollect_info.create_option_parser()
+    options, _ = parser.parse_args(["foo.zip"])
+    assert options.sync_gateway_url is None
--- a/tools-tests/parser_test.py
+++ b/tools-tests/parser_test.py
@@ -22,9 +22,3 @@ def test_parser_log_redaction_salt():
     # assert this is bytes uuid4
     assert isinstance(options.salt_value, bytes)
     assert len(options.salt_value) == 16
-
-
-def test_parser_sg_url():
-    parser = sgcollect_info.create_option_parser()
-    options, _ = parser.parse_args(["foo.zip"])
-    assert options.sync_gateway_url is None
--- a/tools/sgcollect_info
+++ b/tools/sgcollect_info
@@ -16,6 +16,7 @@ import glob
 import json
 import optparse
 import os
+import pathlib
 import platform
 import re
 import ssl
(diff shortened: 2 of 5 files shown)


tools/tasks.py
```diff
@@ -407,11 +409,7 @@ def redact_and_zip(self, filename, log_type, salt, node):
         redactor = LogRedactor(salt, self.tmpdir)
 
         for name, fp in self.files.items():
-            if not (
-                ".gz" in name
-                or "expvars.json" in name
-                or os.path.basename(name) == "sync_gateway"
-            ):
+            if not redactable_file(name) or ".gz" in name:
```
bbrks: I know this is covered by unit tests so I know it works, but the precendece here is quite difficult to parse.

Does `not` negate both `or` clauses, or just the first? The brackets made this obvious before, but it's not that easy to determine any more.

The main concern here is that pprof files have the pb.gz suffix so end up matching both.

Would be kinda nice if `redactable_file` could handle this case inside the function?

tools/sgcollect_info
```diff
@@ -385,9 +395,22 @@ def make_collect_logs_tasks(
...
+    sg_tasks: List[PythonTask] = []
+
+    def add_file_collection_task(filename: pathlib.Path):
+        canonical_filename = pathlib.Path(filename)
+        canonical_filename.resolve()
+        if canonical_filename in sg_log_file_paths:
+            return
+        sg_log_file_paths.add(canonical_filename)
+        if not should_redact or not redactable_file(filename):
+            task = add_file_task(sourcefile_path=filename)
```
bbrks: I wonder if we should use the canonicalised version of the filename here or not? It won't make much material difference other than the path that gets included in the file header I guess

tools/sgcollect_info
```diff
@@ -385,9 +395,22 @@ def make_collect_logs_tasks(
...
+            task.no_header = True
+            sg_tasks.append(task)
+        else:
+            sg_tasks.append(add_gzip_file_task(sourcefile_path=filename, salt=salt))
```
bbrks: Not all of these files are actually gzipped. For example, this glob: `"{0}*{1}".format(name, ext))` grabs `sg_info.log`

It used to be run through `add_file_task` instead of `add_gzip_file_task`

tools/sgcollect_info
torcolvin: These are written without headers, so I am not sure this is helpful. I am not opposed to writing them with headers.

tools/tasks.py
torcolvin: I rewrote this in a kind of aggressive way but I think this is much more robust for anything we add to sgcollect.

The reason that there is this wacky parsing is that text files are opened in latin1. We don't need to figure out the encoding of the files, although I am certain our log files are never latin1. By opening the files in binary mode, we can just do redaction on the bytes.

I created an open function which will detect if it is gzip or not, so it will automatically handle gzipped files or non gzipped files.

This would actually probably mean the pprof files wouldn't get corrupted since almost certainly these regexes wouldn't match. That said, we are now avoiding redacting these files.

tools/sgcollect_info
torcolvin: Good catch, fixed as above.

tools-tests/parser_test.py
```diff
@@ -11,3 +11,14 @@
...
+
+def test_parser_log_redaction_salt():
+    parser = sgcollect_info.create_option_parser()
+    options, _ = parser.parse_args(["--log-redaction-salt=a", "foo.zip"])
+    assert options.salt_value == b"a"
+
+    options, _ = parser.parse_args(["foo.zip"])
+    # assert this is bytes uuid4
+    assert isinstance(options.salt_value, bytes)
+    assert len(options.salt_value) == 16
```
bbrks: I think this may be an incorrect assertion / change. See comparions against `main` here:

Screenshot 2025-02-21 at 17 12 13

Screenshot 2025-02-21 at 17 32 32


Seems like we would want to preserve the string version of `salt_value`, or at least make it a stringified UUID stored as bytes.

The hash is consistent if explicitly given a salt, so perhaps we can just change the `default=` initialization to correct this and accept `b'salt_value'` in the unredacted output.

bbrks (changes requested):

tools-tests/parser_test.py
torcolvin: Really good catch, fixed.

bbrks (approved):

PR #7352, opened 2025-02-04
CBG-4322 use unique output files

torcolvin:
CBG-4322 use unique output files

Written on top of https://github.com/couchbase/sync_gateway/pull/7351 code

In the case that there is a file with the same contents in multiple places, this fixes the idea that this writes this to the same file.

e.g. `log_file_path` in config.json is /var/log/sync_gateway

There is /var/log/sync_gateway/sg_info.log and /home/sync_gateway/logs/sg_info.log

The two will be concatenated together before this fix. There's another fix in https://github.com/couchbase/sync_gateway/pull/7351 which canonicalizes the paths with `path.Pathlib.resolve` before uploading.

I've considered whether to add some `log` messages to say where the files are picked up from, so we'd know why we got a duplicate. log is only written to stderr, but we could write tee to a file  like sgcollect_info-output.log so we'd have an idea why things might go wrong.

New tests were written.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,4 +2,4 @@
 extend-include = ["tools/sgcollect_info"]
 
 [tool.ruff.lint]
-extend-select = ["I"]
+extend-select = ["I","B006"]  # isort, mutable default argument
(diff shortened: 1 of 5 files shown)


bbrks (commented):
The code reivew is fine - I see nothing jumping out at me.

I'd like to spent an hour or so testing it though before merging similar to #7351

bbrks (approved):

PR #7358, opened 2025-02-05
CBG-4391: bubble up errors from startup db process to db summaries

gregns1:
CBG-4391

- Part 1 of better error reporting for database startup for capella, part 2 will be handled in different ticket (failures in start online processes)
- Utilise the invalid db config tracking to bubble up error code + message to the all db's endpoint for use in capella
- This also allows any misconfiguration to be correct by the customer by using the create db endpoint again with a correct config
- Tried to only report errors that may be relevant to capella users (for example, you cannot toggle auto import so not reporting auto import option being invalid) 
- Test to repro the collection being deleted underneath sync gateway error

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2936/

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -308,6 +308,7 @@ type invalidConfigInfo struct {
 	configBucketName    string
 	persistedBucketName string
 	collectionConflicts bool
+	databaseError       *DatabaseError
 }
 
 type invalidDatabaseConfigs struct {
@@ -350,7 +351,7 @@ func (sc *ScopesConfig) HasNewCollection(previousCollectionMap map[string]struct
 
 // addInvalidDatabase adds a db to invalid dbconfig map if it doesn't exist in there yet and will log for it at warning level
 // if the db already exists there we will calculate if we need to log again according to the config update interval
-func (d *invalidDatabaseConfigs) addInvalidDatabase(ctx context.Context, dbname string, cnf DatabaseConfig, bucket string) {
+func (d *invalidDatabaseConfigs) addInvalidDatabase(ctx context.Context, dbname string, cnf DatabaseConfig, bucket string, databaseErr *DatabaseError) {
 	d.m.Lock()
 	defer d.m.Unlock()
 	if d.dbNames[dbname] == nil {
@@ -359,6 +360,7 @@ func (d *invalidDatabaseConfigs) addInvalidDatabase(ctx context.Context, dbname 
 			configBucketName:    *cnf.Bucket,
 			persistedBucketName: bucket,
 			collectionConflicts: cnf.Version == invalidDatabaseConflictingCollectionsVersion,
+			databaseError:       databaseErr,
 		}
 	}
 
@@ -373,6 +375,8 @@ func (d *invalidDatabaseConfigs) addInvalidDatabase(ctx context.Context, dbname 
 	} else if cnf.Version == invalidDatabaseConflictingCollectionsVersion {
 		base.SyncGatewayStats.GlobalStats.ConfigStat.DatabaseRollbackCollectionCollisions.Add(1)
 		logMessage += " Conflicting collections detected"
+	} else if databaseErr != nil {
+		logMessage += " Error encountered loading database."
 	} else {
 		// Nothing is expected to hit this case, but we might add more invalid sentinel values and forget to update this code.
 		logMessage += " Database was marked invalid. See logs for details."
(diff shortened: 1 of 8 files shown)


rest/config.go
```diff
@@ -1998,9 +2002,9 @@ func (sc *ServerContext) FetchConfigs(ctx context.Context, isInitialStartup bool
 }
 
 // _applyConfigs takes a map of dbName->DatabaseConfig and loads them into the ServerContext where necessary.
-func (sc *ServerContext) _applyConfigs(ctx context.Context, dbNameConfigs map[string]DatabaseConfig, isInitialStartup bool) (count int) {
+func (sc *ServerContext) _applyConfigs(ctx context.Context, dbNameConfigs map[string]DatabaseConfig, isInitialStartup bool, loadFromBucket bool) (count int) {
```
bbrks, line 2005: I'm pretty sure `isInitialStartup` is the same flag logically as the new `loadFromBucket`. Is there anywhere where these two bools aren't the same?

rest/server_context_test.go
```diff
@@ -971,3 +971,61 @@ func TestDatabaseStartupFailure(t *testing.T) {
...
+	scopesConfig := GetCollectionsConfig(t, rt.TestBucket, 3)
+	dataStoreNames := GetDataStoreNamesFromScopesConfig(scopesConfig)
+	scope := dataStoreNames[0].ScopeName()
+
+	// remove datastore
+	b, err := base.AsGocbV2Bucket(rt.GetDatabase().Bucket)
+	require.NoError(t, err)
+	dsList, err := b.ListDataStores()
+	require.NoError(t, err)
+	require.NoError(t, b.DropDataStore(dsList[0]))
```
bbrks, line 993: I think this might disrupt latter tests that are relying on there being an intact set of collections on this bucket

Either have the test reinstate the datastore we deleted, or create a new datastore just for this test that we're ok with dropping.

rest/database_error.go
```diff
@@ -0,0 +1,46 @@
...
+	DatabaseInitialisationIndexError
+	DatabaseCreateDatabaseContextError
+	DatabaseSGRClusterError
+	DatabaseCreateReplicationError
+)
+
+var DatabaseErrorString = []string{
+	DatabaseBucketConnectionError:      "Error connecting to bucket",
+	DatabaseInvalidDatastore:           "Collection(s) not available",
+	DatabaseIndexError:                 "Error creating/building/waiting index",
```
bbrks: Is this different than `DatabaseInitialisationIndexError`? Can't see it used so may just be an old entry

rest/database_error.go
```diff
@@ -0,0 +1,46 @@
...
+// DatabaseError denotes an error that occurred during database startup
+type DatabaseError struct {
+	ErrMsg string            `json:"error_message"`
+	Code   DatabaseErrorType `json:"error_code"`
+}
+
+type DatabaseErrorType uint8
+
+const (
+	DatabaseBucketConnectionError DatabaseErrorType = iota
```
bbrks: The `iota` pattern isn't great for things that have external consumers, since it's really easy to add or remove an entry in the middle of the list and change the subsequent error codes.

Think we'll just have to declare all of these codes explicitly to be safe.

bbrks (changes requested):

rest/database_error.go
gregns1: Good spot, forgot to remove!

rest/config.go
gregns1: Its not the same for the goroutine that polls for new configs https://github.com/couchbase/sync_gateway/blob/65c5ae456a553e3018a3ae7ee938eaedde2f0174/rest/server_context.go#L2116 

My thinking was handling incorrect data migration scenarios where you copy dbconfig from on prem with different collections that what are defined in capella, then you would still fall into this handling? I could be going down the wrong rabbit hole there though and would accept if we should only fall into this handling on node startup.

rest/database_error.go
gregns1: Just pushed up commit to have global map for error coder to message, I feel this fixes the issue of adding/removing error codes keeping that unique 1-1 mapping? 

Or do you think actually having to define error code and message constants separate would be best?

rest/config.go
bbrks: Oh I see - this also applies to nodes that load a newly created db config that was created on another node.

rest/database_error.go
```diff
@@ -0,0 +1,44 @@
...
+	DatabaseInitSyncInfoError:          "Error initialising sync info",
+	DatabaseInitialisationIndexError:   "Error initialising database indexes",
+	DatabaseCreateDatabaseContextError: "Error creating database context",
+	DatabaseSGRClusterError:            "Error with fetching SGR cluster definition",
+	DatabaseCreateReplicationError:     "Error creating replication during database init",
+}
+
+type databaseErrorCode uint8
+
+const (
```
bbrks, line 30:
```suggestion
// Error codes exposed for each error a database can encounter on load. These codes are consumed by Capella so must remain stable.
const (
```

bbrks (approved):

PR #7361, opened 2025-02-05
CBG-4186: Put ISGR into an errored stopped state when a reconnect times out

bbrks:
CBG-4186

Put ISGR into an errored stopped state when a reconnect times out
- Allows replicator to be started again after timing out with a failed reconnect loop
- Ensures replicator can be restarted again from this state in new test

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2941/

### DIFF
--- a/db/active_replicator_common.go
+++ b/db/active_replicator_common.go
@@ -196,14 +196,26 @@ func (a *activeReplicatorCommon) reconnectLoop() {
 		return err != nil, err, nil
 	}
 
-	err, _ := base.RetryLoop(ctx, "replicator reconnect", retryFunc, sleeperFunc)
+	retryErr, _ := base.RetryLoop(ctx, "replicator reconnect", retryFunc, sleeperFunc)
 	// release timer associated with context deadline
 	if deadlineCancel != nil {
 		deadlineCancel()
 	}
-	if err != nil {
+	if retryErr != nil {
+		switch ctxErr := ctx.Err(); ctxErr {
+		case context.Canceled:
+			// replicator was stopped - appropriate state has already been set
+			base.InfofCtx(ctx, base.KeyReplicate, "exiting reconnect loop: %v", ctxErr)
+			return
+		case context.DeadlineExceeded:
+			// timeout on reconnecting state
+			base.WarnfCtx(ctx, "aborting reconnect loop after timeout: %v", ctxErr)
+		default:
+			// unexpected error from retry loop
+			base.WarnfCtx(ctx, "aborting reconnect loop after error: %v", retryErr)
+		}
+
 		a.replicationStats.NumReconnectsAborted.Add(1)
-		base.WarnfCtx(ctx, "couldn't reconnect replicator: %v", err)
 		a.lock.Lock()
 		defer a.lock.Unlock()
 		a.setState(ReplicationStateError)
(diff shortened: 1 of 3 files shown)


bbrks (commented):
This appears to have unintended consequences for manually stopping a reconnecting replicator (via `TestActiveReplicatorReconnectSendActions`). Will spend some more time on it.

db/active_replicator_common.go
```diff
@@ -196,14 +196,31 @@ func (a *activeReplicatorCommon) reconnectLoop() {
...
+		default:
+			// unexpected error from retry loop
+			base.WarnfCtx(ctx, "aborting reconnect loop after error: %v", retryErr)
+		}
+
 		a.replicationStats.NumReconnectsAborted.Add(1)
-		base.WarnfCtx(ctx, "couldn't reconnect replicator: %v", err)
+		a.lock.Lock()
+		defer a.lock.Unlock()
+		a.setState(ReplicationStateError)
```
torcolvin: In the case that it stops due to a timeout or cancellation, should we update `setLastError` message to say this was the last error and that we stopped connecting? 

I can see arguments both ways here.

db/active_replicator_common.go
```diff
@@ -196,14 +196,31 @@ func (a *activeReplicatorCommon) reconnectLoop() {
...
 	}
 
-	err, _ := base.RetryLoop(ctx, "replicator reconnect", retryFunc, sleeperFunc)
+	retryErr, _ := base.RetryLoop(ctx, "replicator reconnect", retryFunc, sleeperFunc)
 	// release timer associated with context deadline
 	if deadlineCancel != nil {
 		deadlineCancel()
 	}
-	if err != nil {
+	if retryErr != nil {
```
torcolvin: does an early return make this easier to understand? Not sure.

torcolvin (commented):
Looks good to me, just asking some questions, feel free to drop.

db/active_replicator_common.go
torcolvin: I think it would be good to say that this preserves `setLastError` from above in a comment.

db/active_replicator_common.go
```diff
@@ -201,27 +200,33 @@ func (a *activeReplicatorCommon) reconnectLoop() {
...
-		a.setState(ReplicationStateError)
-		a._publishStatus()
-		a._stop()
+	// Exit early if no error or if context was cancelled
+	if retryErr == nil {
+		return
+	}
+	if ctx.Err() == context.Canceled {
+		// replicator was stopped - appropriate state has already been set
+		base.InfofCtx(ctx, base.KeyReplicate, "exiting reconnect loop: %v", ctx.Err())
```
torcolvin: the context error isn't going to contain anything interesting, so I think you might want logic like:

https://github.com/couchbase/sync_gateway/blob/dccc524c1d6bf046bb2a5463810e074a21ab5e12/base/util.go#L470 where is says "canceled" or "timed out" in the log and then logs the last known error (or not, since that will already be in the logs).

I think this means that  should never get hit because the only ways to leave the loop. We should definitely log this but I don't think will get hit and you can probably comment to that effect.

```
// unexpected error from retry loop
		base.WarnfCtx(ctx, "aborting reconnect loop after error: %v", retryErr)
```

Sorry I didn't catch this before, I'm just trying to figure out how to have the logs be useful AND meaningful.

This is a future enhancement, but I often wonder whether https://github.com/couchbase/sync_gateway/blob/dccc524c1d6bf046bb2a5463810e074a21ab5e12/base/util.go#L475 should wrap the ctx error.

torcolvin (approved):

PR #7369, opened 2025-02-07
CBG-4454: Add Database Init and Online fatal error stats

bbrks:
CBG-4454

Add new stats for errors that cause databases to fail to come online (both with `Init` and `StartOnline`).
The stats are marked Internal, and are not committed for external or customer use - at least until we have a better understanding of their use-case.

- [ ] I would like a review from App Services team to actually make sure these stats are useful given the other changes to API for error reporting. I am unsure what the expectation of action is from these stats. Do they monitor and then use the API to find out more information?


rest/server_context.go
```diff
@@ -1010,6 +1013,7 @@ func (sc *ServerContext) asyncDatabaseOnline(nonCancelCtx base.NonCancellableCon
 	// Before bringing the database online, ensure that the database hasn't been modified while we waited for initialization to complete
 	currentDbVersion := sc.GetDbVersion(dbc.Name)
 	if currentDbVersion != version {
+		dbc.DbStats.DatabaseStats.TotalOnlineFatalErrors.Add(1)
```
torcolvin: I don't think this is a fatal error, the newer version is going to get picked up?

JRascagneres:
>  I am unsure what the expectation of action is from these stats. Do they monitor and then use the API to find out more information?

Essentially stats are super easy to alert on as all the infrastructure is built around stat alerts. We can't really alert on something that requires a call to SGW. The intention is that the stats can trigger alerts, either for us (as support) to action or for the customer to action. The intention is that the API side can provide more information through the UI which calls to SGW on demand. 

TLDR: Yes there is still useful information to be gained by having stats alongside the error reporting through the API.

torcolvin (approved):

PR #7376, opened 2025-02-11
CBG-4512: pick up gocb fix for bucket names with . in

gregns1:
CBG-4512

- pick up commit https://github.com/couchbase/gocb/commit/8ef6d90105113012ba7d73a833918808f8539408 from gocb, this seems to also update gocbcore with it
- Add test that fails pre this uptake

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2954/

### DIFF
--- a/rest/config_database_test.go
+++ b/rest/config_database_test.go
@@ -10,6 +10,7 @@ package rest
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"testing"
 	"time"
@@ -92,7 +93,7 @@ func TestConfigToBucketPointName(t *testing.T) {
 
 	rt := NewRestTester(t, nil)
 	defer rt.Close()
-	const testBucketName = "sg_int_my.Bucket"
+	testBucketName := base.CreateTestBucketName(fmt.Sprintf(".%d", time.Now().Unix()))
 
 	// create db config to point to bucket with . in the name
 	dbConfig := rt.NewDbConfig()
--- a/rest/config_database_test.go
+++ b/rest/config_database_test.go
@@ -88,12 +88,15 @@ func TestConfigToBucketPointName(t *testing.T) {
 	if base.UnitTestUrlIsWalrus() {
 		t.Skip("Need cbs bucket for this test")
 	}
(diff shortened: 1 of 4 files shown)


torcolvin (dismissed):
I looked at gocbcore commits from v10.5.2 to 134c149 and I do not think any of the code affects Sync Gateway's usage.

torcolvin (commented):
I am a bit worried about this test but I think I will leave it for now. The issue is that there might not be enough room (memory) to create buckets on a cluster and this bucket could exist on a server. I actually think there's probably low value and high risk to doing this test, and we should perhaps remove it.

rest/config_database_test.go
```diff
@@ -82,3 +83,48 @@ func TestDbConfigUpdatedAtField(t *testing.T) {
...
+
+func TestConfigToBucketPointName(t *testing.T) {
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("Need cbs bucket for this test")
+	}
+	base.SetUpTestLogging(t, base.LevelInfo, base.KeyHTTPResp, base.KeyHTTP)
+
+	rt := NewRestTester(t, nil)
+	defer rt.Close()
+	const testBucketName = "sg_int_my.Bucket"
```
torcolvin: Can you make a function like this public in base/util_testing.go so we can use it like we do in a bucket pool?

```
	// Append a timestamp to all of the bucket names to ensure uniqueness across a single package.
	// Not strictly required, but can help to prevent (index) resources from being incorrectly reused on the server side for recently deleted buckets.
	bucketNameTimestamp := time.Now().UnixNano()

	// create required number of buckets (skipping any already existing ones)
	for i := 0; i < numBuckets; i++ {
		bucketName := fmt.Sprintf(tbpBucketNameFormat, tbpBucketNamePrefix, i, bucketNameTimestamp)
```

This is from main_test_bucket_pool.go ?		

I'm just trying to avoid a failure from a test collision.

torcolvin (approved):

PR #7379, opened 2025-02-12
CBG-4518: better error to pass up stack when xattr config doesn't exist

gregns1:
CBG-4518

- Pass up stack more descriptive error when metadata doc is found with no xattr config 
- Added test to test different scenarios returning the new xattr error type added 
- Added code to add db's ties to a bucket into invalid db tracking map when we hit this issue on config poll for a bucket. 
- Added code to fail config update and take tb offline, avoiding a config rollback when this is hit during a db config update. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2966/


base/config_persistence.go
```diff
@@ -108,7 +108,7 @@ func (xbp *XattrBootstrapPersistence) loadRawConfig(ctx context.Context, c *gocb
 		xattrContErr := res.ContentAt(0, &rawValue)
 		if xattrContErr != nil {
 			DebugfCtx(ctx, KeyCRUD, "No xattr config found for key=%s, path=%s: %v", key, cfgXattrConfigPath, xattrContErr)
-			return rawValue, 0, ErrNotFound
```
torcolvin, line 111: I am very reluctant to change this error type because this has an impact on the retry behavior - the usage of `IsDocNotFoundError` in each call in `config_manager.go` depends on this https://github.com/couchbase/sync_gateway/blob/aaa9f0136925df4469b36f4919fa932b381089be/rest/config_manager.go#L438

I'd feel safer about _wrapping_ this `ErrNotFound` vs retyping this error but I do not know if this error would bubble up in this case.

rest/config_test.go
```diff
@@ -3301,3 +3301,43 @@ func TestServerUUIDRuntimeServerConfig(t *testing.T) {
...
+	dbConfig1 := rt.NewDbConfig()
+	RequireStatus(t, rt.CreateDatabase("db1", dbConfig1), http.StatusCreated)
+
+	// delete created registry
+	err := rt.CustomTestBucket.DefaultDataStore().Delete(base.SGRegistryKey)
+	require.NoError(t, err)
+
+	// create registry document in the bucket with no xattr config
+	_, err = rt.CustomTestBucket.DefaultDataStore().AddRaw(base.SGRegistryKey, 0, []byte("{}"))
+	require.NoError(t, err)
```
torcolvin: What happens with config polling at this point? Does it unload the database? The existing code will not be able to unload the database here.

torcolvin (commented):
I feel pretty nervous about changing this error still, given how the retry code works here https://github.com/couchbase/sync_gateway/blob/aaa9f0136925df4469b36f4919fa932b381089be/rest/config_manager.go#L375

I'd make sure that we have tests in the rest package that cover each of the places `IsDocNotFoundError` where the error would be hit in config_manager for xattr config.

I think we should pursue writing a document describing the ways that a user can end up this scenario.

torcolvin:
It is not clear that this is the correct solution for this problem overall. This adds complexity to server context and might not solve the original problem.

This probably won't be merged as is, and rethinking the problem and deciding if this right approach is correct. It may be that this is OK, but this ticket is not a current priority so closing this PR. This can be considered parked, and reopened when we next work on this ticket.

PR #7382, opened 2025-02-13
CBG-4452 avoid checking revpos on attachments

torcolvin:
CBG-4452 avoid checking revpos on attachments

The test failed due to a race between when rev 26-abc was stored on server and when the BlipTesterClient had updated the last known server value of a document. `rt.WaitForVersion` would wait for when the document is stored on Couchbase Server but the last known server revision is stored when the `rev` message comes back. It would have been possible to fix this test, but it was testing behavior that currently exists, so decided to remove it.

- Changed Blip Tester to only send 20 revs in history to match CBL behavior

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2958/


db/blip_handler.go
```diff
@@ -1128,22 +1128,18 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
...
-				commonAncestor := currentDoc.History.findAncestorFromSet(currentDoc.CurrentRev, history)
-				minRevpos, _ = ParseRevID(bh.loggingCtx, commonAncestor)
-				minRevpos++
 				rawBucketDoc = rawDoc
 				currentBucketDoc = currentDoc
-			} else {
-				minRevpos, _ = ParseRevID(bh.loggingCtx, history[len(history)-1])
 			}
 		}
+		// updatedRevPos is the revpos of the new revision, to be added to attachment metadata if needed for CBL<4.0 compatibility. revpos is not used
```
adamcfraser: Can we identify which versions of CBL require revpos, if any?  I was under the impression that revpos is strictly a SGW concept and shouldn't actually be required for any clients as long as SGW is using 1:1 attachment:document storage.

torcolvin: I don't think any code requires revpos, I looked back to release/2.8. Some lite code looks like it sends revpos to the server but I do not think this matters.

I ran the tests in couchbase-lite-core that use real sync gateway but I think https://github.com/couchbaselabs/couchbase-lite-tests/blob/4d49b6f9f71744ce8b121b74922f35ae717e3d76/tests/test_replication_blob.py#L95 does an absolute diff on the document and that's why I didn't remove populating revpos in https://github.com/couchbase/sync_gateway/commit/8986872dddf91054811fccaed7bc78e094c8cc88

revpos is also used by couchdb, but I'm not sure how.

torcolvin:
This code preserves revpos on output for potential couchdb compatibility, but it does not use revpos for any calculations. The revpos is added to metadata on input. Existing tests cover this behavior.

db/blip_handler.go
adamcfraser:
```suggestion
		// updatedRevPos is the revpos of the new revision, to be added to attachment metadata if needed for CBL<4.0 compatibility. revpos is no longer used by Sync Gateway.
```

db/blip_handler.go
```diff
@@ -1190,23 +1186,18 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
...
-				if !ok {
-					return base.HTTPErrorf(http.StatusBadRequest, "Invalid attachment")
-				}
-
 				// Compare the revpos and attachment digest. If incoming revpos is less than or equal to minRevPos and
 				// digest is different we need to override the revpos and set it to the current revision to ensure
-				// the attachment is requested and stored
-				if int(incomingAttachmentRevpos) <= minRevpos && currentAttachmentDigest != incomingAttachmentDigest {
-					bodyAtts[name].(map[string]interface{})["revpos"], _ = ParseRevID(bh.loggingCtx, revID)
+				// the attachment is requested and stored. revpos provided for SG/CBL<4.0 compatibility but is not used.
```
adamcfraser:
```suggestion
				// the attachment is requested and stored. revpos provided for SG/CBL<4.0 compatibility but is no longer used by Sync Gateway.
```

adamcfraser (dismissed):

adamcfraser (approved):

PR #7389, opened 2025-02-20
CBG-4400 re-enable skipped attachment tests

torcolvin:
CBG-4400 enable skipped attachment tests

- Rather than having "fake revs", have real revs.
- Create function to prune old versions out of BlipTesterClient
- Remove versionvector test code on main (since those tests are always skipped)
- changed the test sync metadata to use sequence 1 so it doesn't show up as a skipped sequence in test

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2960/

### DIFF
--- a/rest/attachment_test.go
+++ b/rest/attachment_test.go
@@ -1900,10 +1900,8 @@ func TestBasicAttachmentRemoval(t *testing.T) {
 		attBody := []byte(`hi`)
 		digest := db.Sha1DigestKey(attBody)
 		attKey := db.MakeAttachmentKey(db.AttVersion1, docID, digest)
-		rawDoc := rawDocWithAttachmentAndSyncMeta()
-
 		// Create a document with legacy attachment.
-		CreateDocWithLegacyAttachment(t, rt, docID, rawDoc, attKey, attBody)
+		CreateDocWithLegacyAttachment(t, rt, docID, rawDocWithAttachmentAndSyncMeta(rt), attKey, attBody)
 
 		// Get the document and grab the revID.
 		version, _ := rt.GetDoc(docID)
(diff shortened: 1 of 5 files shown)


rest/attachment_test.go
```diff
@@ -2699,13 +2685,17 @@ func retrieveAttachmentMeta(t *testing.T, rt *RestTester, docID string) (attMeta
 	return attachments
 }
 
-func rawDocWithAttachmentAndSyncMeta() []byte {
-	return []byte(`{
+// rawDocWithAttachmentAndSyncMeta returns a raw document with an attachment and sync metadata inline. RestTester is used to determine the sequence for the document.
+func rawDocWithAttachmentAndSyncMeta(rt *RestTester) []byte {
+	latestSeq, err := rt.GetDatabase().LastSequence(rt.Context())
+	require.NoError(rt.TB(), err)
+	docSeq := latestSeq + 1
```
bbrks: I can't decide if this is correct logically and not safe, or is actually not a valid use of "latestSeq + 1".

Are we relying on a subsequent import to write a _new_ set of sync metadata (and sequence number), or are we just migrating this current `_sync` data into xattrs via on-demand import?

The code written is assuming no other writes will happen between this doc being written and the import allocating a new sequence number.

Would suggest exposing `nextSequence()` in order to get a valid and allocated sequence to assign/use in this doc meta if this is a sequence value we want to use/keep.

bbrks (changes requested):
Looks good - I just have a question about how `CreateDocWithLegacyAttachment` allocates its sequence

rest/attachment_test.go
torcolvin: Good call.

bbrks (approved):

PR #7402, opened 2025-02-27
CBG-4528 remove separate doc lock and seq lock

torcolvin:
CBG-4528 remove separate doc lock and seq lock to make write operations atomic

- By keeping a single lock, and requiring access to it when accessing `_seqStore` or any document, this can prevent out of sync data. There is still separate data for the attachment storage, and the message storage, but I considered it out of scope for this PR to handle, and I'm not sure that it is required to fix.
- To avoid misusing the API, do not allow direct access to `clientDoc` structure, since you need to hold the lock on the parent structure to make sure no one else is accessing it.
- The unit of transaction switched from `clientDoc` to `proposeChangeBatchEntry`. `proposeChangeBatchEntry` isn't the best name for the data structure, but it contains all the information needed to generate and send a change, while making a copy of the data to not worry about underlying data structure being mutated. It does not contain the document body, since this is unique and fetchable later.
 
Previously the separate locks led to buggy circumstances around code such as this, since the sequence was locked but the document could be mutated before `_seqsByVersions` set and `setLatestServerVersion`. 

```
		btcc._seqStore[newClientSeq] = doc
		btcc._seqFromDocID[docID] = newClientSeq

		if replacedRev != "" {
			// store the new sequence for a replaced rev for tests waiting for this specific rev
			doc.lock.Lock()
			doc._seqsByVersions[DocVersion{RevID: replacedRev}] = newClientSeq
			doc.lock.Unlock()
		}
		doc.setLatestServerVersion(newVersion)
```

since the lastServerVersion is separately from teh document body getting changed.

The reviewer should check: Are the locks present in all cases? Are they correctly read/write?			

I am open to a better way of abstracting things, since now everything needs to go through `BlipTesterCollectionClient`. I have run a lot of tests with a lot of race detector, but it is hard to spot any of the failing tests by eyes, but to the best of my testing, this hasn't introduced any new bugs.

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2972/

### DIFF
--- a/rest/blip_client_test.go
+++ b/rest/blip_client_test.go
@@ -1208,7 +1208,6 @@ func (btr *BlipTesterReplicator) sendMsg(msg *blip.Message) {
 func (btcc *BlipTesterCollectionClient) upsertDoc(docID string, parentVersion *DocVersion, body []byte) *clientDocRev {
 	btcc.seqLock.Lock()
 	defer btcc.seqLock.Unlock()
-	newSeq := btcc._nextSequence()
 	oldSeq, ok := btcc._seqFromDocID[docID]
 	var doc *clientDoc
 	if ok {
@@ -1233,6 +1232,7 @@ func (btcc *BlipTesterCollectionClient) upsertDoc(docID string, parentVersion *D
 	digest := "abc" // TODO: Generate rev ID digest based on body hash?
 
 	newRevID := fmt.Sprintf("%d-%s", newGen, digest)
+	newSeq := btcc._nextSequence()
 	rev := clientDocRev{clientSeq: newSeq, version: DocVersion{RevID: newRevID}, body: body}
 	doc._addNewRev(rev)
 
--- a/rest/blip_client_test.go
+++ b/rest/blip_client_test.go
@@ -1032,19 +1032,18 @@ func (btcc *BlipTesterCollectionClient) sendRev(ctx context.Context, change prop
 	revRequest.Properties[db.RevMessageRev] = change.version.RevID
 	revRequest.Properties[db.RevMessageHistory] = change.historyStr()
 
-	docBody, deltaBody, deltaVersion := btcc.getBodyAndDeltas(change.docID, change.version, change.latestServerVersion)
-
+	var deltaVersion *DocVersion
+	var docBody []byte
+	if deltasSupported {
+		docBody, deltaVersion = btcc.getDeltaBody(change.docID, change.version, change.latestServerVersion)
+	} else {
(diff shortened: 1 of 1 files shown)


rest/blip_client_test.go
```diff
@@ -1332,6 +1208,7 @@ func (btr *BlipTesterReplicator) sendMsg(msg *blip.Message) {
 func (btcc *BlipTesterCollectionClient) upsertDoc(docID string, parentVersion *DocVersion, body []byte) *clientDocRev {
 	btcc.seqLock.Lock()
 	defer btcc.seqLock.Unlock()
+	newSeq := btcc._nextSequence()
```
bbrks: I think "allocating" a next sequence right before its used is safer in case we add any early returns/exits in this code.

I don't think it would cause functional issues but it would at least avoid unused allocated sequences in the client - I'm not sure whether the code is set up to handle that.

rest/blip_client_test.go
```diff
@@ -1456,19 +1325,42 @@ func (btcc *BlipTesterCollectionClient) WaitForVersion(docID string, docVersion
...
+	defer btcc.seqLock.RUnlock()
+	doc, ok := btcc._getClientDoc(docID)
+	require.True(btcc.TB(), ok, "docID %q not found", docID)
+	var body db.Body
+	require.NoError(btcc.TB(), body.Unmarshal(doc._getRev(btcc.TB(), version).body))
+	return body
+}
+
+// getBodyAndDeltas returns the body and delta from version to serverVersion. If serverVersion was a tombstone, do not return delta. This will fail the test harness if not present.
+func (btcc *BlipTesterCollectionClient) getBodyAndDeltas(docID string, version DocVersion, serverVersion DocVersion) (body []byte, deltaBody db.Body, parentVersion *DocVersion) {
```
bbrks: This function isn't returning a delta - though it arguably should. It is only returning a full server version body right now.

We need to also move the `Diff()` code down into here to to achieve that.

bbrks (changes requested):
I'm Ok with these changes - just a couple of comments about moving some bits of code around.

rest/blip_client_test.go
torcolvin: Good call.

rest/blip_client_test.go
torcolvin: Hopefully I made this code more clear now.

bbrks (approved):

100manbul (approved):

PR #7405, opened 2025-02-28
CBG-4541: rev cache changes merge into anemone

gregns1:
[CBG-4541](https://jira.issues.couchbase.com/browse/CBG-4541)

- Take rev cache deadlock changes into anemone
- Also fixes some memory based eviction issues 
- Parametrised tests for cv and revid code paths for eviction 
- Move some high churn tests into file for no race detector to be run as these test fail on test on race (asserting on map contents while goroutines teardown).
- Have a new test for race condition fix in `revision_cache_test.go`

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2981/

[CBG-4541]: https://couchbasecloud.atlassian.net/browse/CBG-4541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rest/importtest/import_test.go
```diff
@@ -2326,6 +2329,8 @@ func TestImportRollback(t *testing.T) {
 // - Test is much like TestImportRollback, but with multiple partitions and multiple vBuckets rolling back
 // - Test case rollbackWithoutFailover will only rollback one partition
 func TestImportRollbackMultiplePartitions(t *testing.T) {
+	t.Skip("test will fail on this branch, no cbgt update on here yet")
```
torcolvin: tag with https://jira.issues.couchbase.com/browse/CBG-4505

db/revision_cache_no_race_test.go
```diff
@@ -0,0 +1,233 @@
+//go:build !race
```
torcolvin: These don't need to be marked no-race anymore, do they?

gregns1: These tests do, the tests themselves race sometimes. Teardown of goroutines acting on rev cache and asserting on rev cache length etc. These are definetly test only races. To show the race condition from before is fixed I added a test back to the main test file that previously failed on race detector.

I think its worth keeping these tests in as they show that rev cache lookup maps are in sync after high eviction churn on the cache.

torcolvin: These tests are enabled on the race detector on main, what makes the anemone code different?

gregns1: I added the extra assertions here given 1. the race condition found on anemone and 2. we have two look up maps and wanted to have ways to assert that under high churn the maps are aligned

gregns1: I can just remove these assertions and put them back if you think it worth doing that?

adamcfraser (commented):
LGTM

adamcfraser (approved):

PR #7410, opened 2025-03-07
CBG-4549: Refactor resync Init so the reason for a reset is always logged

bbrks:
CBG-4549

- Refactor resync Init so the reason for a reset is always logged.
Scenarios covered by subtests in `TestResyncDCPInit`:
```
Initialize_new_run_with_empty_cluster_state db:db Resync: Resetting resync process with ID: "..." - no previous run found
Reinitialize_existing_run db:db Resync: Resuming resync with ID: "..."
Reinitialize_completed_run db:db Resync: Resetting resync process with ID: "..." - previous run completed
Force_restart_existing_run db:db Resync: Resetting resync process with ID: "..." - reset option requested
```
- Remove outdated resync information (since 3.0 we have cluster-aware resync which won't allow multiple nodes to run resync concurrently)

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2985/

### DIFF
--- a/db/background_mgr_resync_dcp.go
+++ b/db/background_mgr_resync_dcp.go
@@ -74,33 +74,32 @@ func (r *ResyncManagerDCP) Init(ctx context.Context, options map[string]interfac
 	// add collection list to manager for use in status call
 	r.SetCollectionStatus(collectionNames)
 
-	// If the previous run completed, or there was an error during unmarshalling the status we will start the
-	// process from scratch with a new resync ID. Otherwise, we should resume with the resync ID, stats specified in the doc.
-	var resetReason string
+	// If the previous run completed, or we couldn't determine, we will start the resync with a new resync ID.
+	// Otherwise, we should resume with the resync ID, and the previous stats specified in the doc.
+	var resetMsg string // an optional message about why we're resetting
 	var statusDoc ResyncManagerStatusDocDCP
 	if clusterStatus == nil {
-		resetReason = "no previous run found"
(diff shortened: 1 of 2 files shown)


db/background_mgr_resync_dcp.go
```diff
@@ -85,29 +74,34 @@ func (r *ResyncManagerDCP) Init(ctx context.Context, options map[string]interfac
...
+		if err != nil {
+			return err
 		}
-		r.ResyncID = statusDoc.ResyncID
-		r.SetStatus(statusDoc.DocsChanged, statusDoc.DocsProcessed)
-
-		base.InfofCtx(ctx, base.KeyAll, "Resync: Attempting to resume resync with resync ID: %s", r.ResyncID)
-
+		r.ResyncID = newID.String()
+		base.InfofCtx(ctx, base.KeyAll, "Resync: Resetting resync process with ID: %q - %s", r.ResyncID, resetReason)
```
torcolvin: I am fine with this change but I wonder if the two common scenarios here actually should not have the log line reset? I am not convinced by this, but I worry reset implies dropping previous bad state.

`clusterStatus == nil` - technically this is a reset because you are "setting" but I think that this is better stated as just "Starting resync process".

Similarly with `statusDoc.State == BackgroundProcessStateCompleted` but this case I can see how reset seems like the right word.

torcolvin (dismissed):

torcolvin (approved):

PR #7419, opened 2025-03-12
CBG-4560 mark always replicator reconnecting

torcolvin:
CBG-4560 mark always replicator reconnecting

Addresses a test flake where `TestPanicInCheckpointHash` would panic when reconnect loop is called from https://github.com/couchbase/sync_gateway/blob/main/db/active_replicator.go#L221 

Before this change calling `ActiveReplicator.Stop` would not correctly wait for reconnection if it occurred in the `OnExitCallback`. The panic occurs from the passive side when https://github.com/couchbase/sync_gateway/blob/main/db/active_replicator.go#L261 was called.

Renamed the test to not include the word `Panic` to make it easier to find real panics in test results.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2989/

### DIFF
--- a/db/active_replicator_common.go
+++ b/db/active_replicator_common.go
@@ -273,14 +273,17 @@ func (arc *activeReplicatorCommon) reconnectLoop() {
 		ctx)
 
 	retryFunc := func() (shouldRetry bool, err error, _ interface{}) {
-		select {
-		case <-ctx.Done():
+		// check before and after acquiring lock to make sure to exit early if ActiveReplicatorCommon.Stop() was called.
+		if ctx.Err() != nil {
 			return false, ctx.Err(), nil
-		default:
 		}
 
 		arc.lock.Lock()
 		defer arc.lock.Unlock()
+
+		if ctx.Err() != nil {
+			return false, ctx.Err(), nil
+		}
 
 		// preserve lastError from the previous connect attempt
 		arc.setState(ReplicationStateReconnecting)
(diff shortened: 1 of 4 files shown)


db/active_replicator_common.go
```diff
@@ -165,7 +165,6 @@ func (arc *activeReplicatorCommon) Start(ctx context.Context) error {
 			defer arc.ctxCancel()
 		} else {
 			base.InfofCtx(arc.ctx, base.KeyReplicate, "Attempting to reconnect in background: %v", err)
-			arc.reconnectActive.Set(true)
 			go arc.reconnectLoop()
```
bbrks: I feel like this change may potentially cause other races, where `reconnectActive.Set(true)` was previously being called inside the `arc.Lock`, and if we move it down into the `reconnectLoop` goroutine, there's a time at which there's an unlocked replicator that is about to go into a reconnecting state but hasn't yet set the flag?

Maybe races of this kind are guarded by other things, but I'm not as confident in this change as adding the `reconnectActive.Set` in the `OnExitCallback`. That seems like an obvious omission to me that doesn't affect any locking mechanisms.

Let me know if you disagree. I know this approach is easier to forget instances of `reconnectActive` for new `go arc.reconnectLoop()` invocations- but perhaps both lines could just be moved into a synchronous function?

bbrks (commented):
Have a question about this change

db/active_replicator_common.go
torcolvin: I think it would be fine to put the check inside the goroutine, I don't think there's a race to protect from. That said, I did spot a race which is in the reconnect function that it locks, then tries to connect. But if it is locked because `stop` was called (which acquires the lock), then the reconnect loop could reconnect after stop was called. Added a context cancel check again right after the lock to protect against it.

I think this refactor splits the difference between what we were going for.

Also, renamed misnamed `reconnect` function to `disconnect` in the refactor when I renamed `reconnectLoop` to `reconnect`.

bbrks (approved):

PR #7428, opened 2025-03-18
CBG-4548 allow creation of partitioned indexes

torcolvin:
CBG-4548 allow creation of partitioned indexes

 - create optional database config `index.num_partitions` indicating how many partitions to create. By default, indexes will not be partitioned.
 - partition only the largest indexes, `channels` and `allDocs` which scale for the number of docs and the number of channels in each document.
 - move `num_index_replicas` to `index.num_replicas` in configuration to allow more index options to be created in the future.
 	- `num_index_replicas` is backward compatible
    - move test code to use `index.num_replicas` or remove the definition when unnecessary (it is set by `RestTester`)
- prevent xattrs=false from using partitioned indexes, to make testing easier

I think there's test coverage for everything here, but the things I am most worried about is roundtripping `/db/_config` and making sure that `DbConfig.Index` doesn't get set automatically so a `GET` followed by a `PUT` would result in a mismatched configuration. Additionally, there is concern/requirement to make this work for legacy configuration.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2998/


base/collection_n1ql_common.go
```diff
@@ -161,11 +161,16 @@ func createIndex(ctx context.Context, store N1QLStore, indexName string, express
 	if filterExpression != "" {
 		filterExpressionStr = " WHERE " + filterExpression
 	}
+	var partitionExpresionStr string
+	if options.NumPartitions != nil {
```
bbrks: I think we also need a `> 1` check or else you'll end up with partitioned indexes even when the value is set to 1

base/collection_n1ql_common.go
```diff
@@ -161,11 +161,16 @@ func createIndex(ctx context.Context, store N1QLStore, indexName string, express
...
+		partitionExpresionStr = " PARTITION BY HASH(META().id)"
```
bbrks, line 166: We need to pass in the actual value of `options.NumPartitions` otherwise it's just Query's default 8, but that will have to be put into `options` rather than appended to the query here

https://docs.couchbase.com/server/current/n1ql/n1ql-language-reference/index-partitioning.html#index-with

db/database.go
```diff
@@ -1915,6 +1916,14 @@ func (context *DatabaseContext) UseXattrs() bool {
 	return context.Options.EnableXattr
 }
 
+// numIndexPartitions returns the number of index partitions to use for the database's indexes.
+func (context *DatabaseContext) numIndexPartitions() uint32 {
+	if context.Options.NumIndexPartitions != nil {
+		return *context.Options.NumIndexPartitions
+	}
+	return 1
```
bbrks:
```suggestion
	return DefaultNumIndexPartitions
```

db/indextest/indextest_test.go
```diff
@@ -310,15 +317,42 @@ func TestInitializeIndexesConcurrentMultiNode(t *testing.T) {
...
+
+	setupN1QLStore(ctx, t, database.Bucket, serverless, numPartitions)
+	gocbBucket, err := base.AsGocbV2Bucket(database.Bucket)
+	require.NoError(t, err)
+	for _, dsName := range []sgbucket.DataStoreName{db.GetSingleDatabaseCollection(t, database.DatabaseContext).GetCollectionDatastore(), database.MetadataStore} {
+		allIndexes, err := gocbBucket.GetCluster().Bucket(gocbBucket.BucketName()).Scope(dsName.ScopeName()).Collection(dsName.CollectionName()).QueryIndexes().GetAllIndexes(nil)
+		require.NoError(t, err)
+		for _, index := range allIndexes {
+			if strings.HasPrefix(index.Name, "sg_allDocs") || strings.HasPrefix(index.Name, "sg_channels") {
+				require.True(t, strings.HasSuffix(index.Name, "x1_p8"), "expected 8 partitions for %+v", index)
```
bbrks: Is it possible to assert on the _actual_ number of partitions using this API? I can't tell whether it will be in `index.Params`

If so, changing `numPartitions` to something other than 8 and doing that assertion seems like an easy enhancement to this test to cover the options on query creation - likewise for the else case below - ensuring we only have 1 partition for everything else

bbrks (changes requested):
Not sure whether this PR was intended to implement custom partition counts other than 8. The ticket and PR were a bit vague in terms of just stating "opt-in" - left comments below along those lines though just in case it was not intended

base/collection_n1ql_common.go
torcolvin: This works by magic by adding options struct to with statement https://github.com/couchbase/sync_gateway/blob/a110b1a276b157dc6eaff3970c18c57cbc8c2f78/base/collection_n1ql_common.go#L191

You need both `PARTITION BY HASH` and `WITH { "num_partition": 8}`

db/indextest/indextest_test.go
torcolvin: I added an assertion in a gross way, querying all the index nodes and then aggregating information but I don't think there's a better way to do this.

base/collection_n1ql_common.go
bbrks: Oh my mistake - I didn't notice that `N1qlIndexOptions` was the struct being marshalled as options.

rest/config_database.go
```diff
@@ -153,17 +153,19 @@ func DefaultDbConfig(sc *StartupConfig, useXattrs bool) *DbConfig {
...
+		OldRevExpirySeconds:   base.Uint32Ptr(base.DefaultOldRevExpirySeconds),
+		ViewQueryTimeoutSecs:  base.Uint32Ptr(base.DefaultViewTimeoutSecs),
+		LocalDocExpirySecs:    base.Uint32Ptr(base.DefaultLocalDocExpirySecs),
+		EnableXattrs:          base.BoolPtr(base.DefaultUseXattrs),
+		SecureCookieOverride:  base.BoolPtr(sc.API.HTTPS.TLSCertPath != ""),
+		SessionCookieName:     "",
+		SessionCookieHTTPOnly: base.BoolPtr(false),
+		AllowConflicts:        base.BoolPtr(base.DefaultAllowConflicts),
+		Index: &IndexConfig{
+			NumReplicas: base.UintPtr(DefaultNumIndexReplicas),
```
bbrks, line 167: This doesn't change anything other than displaying "1" as default when `include_runtime=true` is set on a GET config request but I think we should do it.

```suggestion
			NumReplicas:   base.UintPtr(DefaultNumIndexReplicas),
			NumPartitions: base.Ptr(uint32(db.DefaultNumIndexPartitions),
```

rest/config.go
```diff
@@ -171,7 +171,8 @@ type DbConfig struct {
 	SessionCookieName                string                           `json:"session_cookie_name,omitempty"`                  // Custom per-database session cookie name
 	SessionCookieHTTPOnly            *bool                            `json:"session_cookie_http_only,omitempty"`             // HTTP only cookies
 	AllowConflicts                   *bool                            `json:"allow_conflicts,omitempty"`                      // Deprecated: False forbids creating conflicts
-	NumIndexReplicas                 *uint                            `json:"num_index_replicas,omitempty"`                   // Number of GSI index replicas used for core indexes
+	NumIndexReplicas                 *uint                            `json:"num_index_replicas,omitempty"`                   // Number of GSI index replicas used for core indexes, deprecated for IndexConfig.NumReplicas
+	Index                            *IndexConfig                     `json:"index,omitempty"`                                // Index options
```
bbrks, line 175: Requires API spec changes

bbrks (dismissed):
This looks good 👍 
- One nit about `DefaultDbConfig` to ensure the default (logical) value of 1 is shown
- Missing API docs for db config endpoints

rest/config.go
torcolvin: Thanks, can you review text copy for me?

rest/config_database.go
torcolvin: I think this is fine but let me run through jenkins before merging.

docs/api/components/schemas.yaml
```diff
@@ -1719,9 +1719,23 @@ Database:
...
+    index:
+      type: object
+      description: Settings for Global Secondary Indexes (GSI).
+      properties:
+        num_replicas:
+          description: This is the number of Global Secondary Indexes (GSI) to use for core indexes.
+          type: number
+          default: 1
+        num_partitions:
+          description: The number of partitions to use for the Global Secondary Indexes (GSI) for large indexes. It is not recommended to set this unless you have performance issues. If set, the recommended number is 8.
```
bbrks: Going to preempt some user-facing documentation around this which calls out performance tradeoffs and any required migration steps, since if this option is available, some users are going to try without completely comprehending.

Feel free to push back or change based on what you know/think we'll be doing around docs.

I used this page/section as a reference for this suggestion:

https://docs.couchbase.com/server/current/n1ql/n1ql-language-reference/index-partitioning.html#performance-considerations 

```suggestion
          description: The number of partitions to use for the large indexes created by Sync Gateway. It is not recommended to set this unless you require additional horizontal scalability for individual indexes and have appropriately scaled your Query nodes to handle the increased query parallelism. If set, the recommended number is 8 and does not need to be directly related to the number of your Query nodes. Ensure documentation is read to understand the performance tradeoffs and instructions for migration if you have previously run with only one partition.
```

bbrks (commented):
All looks good, with one suggestion for documentation which I'll leave for @adamcfraser to review the wording of as well.

db/indexes.go
```diff
@@ -175,6 +177,10 @@ var (
 			"ORDER BY [op.name, LEAST($sync.`sequence`, op.val.seq),IFMISSING(op.val.rev,null),IFMISSING(op.val.del,null)] " +
 			"LIMIT 1",
 	}
+	partionableIndexes = map[SGIndexType]bool{
```
adamcfraser:
```suggestion
	partitionableIndexes = map[SGIndexType]bool{
```

db/indexes.go
```diff
@@ -214,18 +221,23 @@ type SGIndex struct {
...
-func (i *SGIndex) indexNameForVersion(version int, useXattrs bool) string {
+func (i *SGIndex) indexNameForVersion(version int, useXattrs bool, numPartitions uint32) string {
 	xattrsToken := ""
 	if useXattrs {
 		xattrsToken = "x"
 	}
-	return fmt.Sprintf(indexNameFormat, i.simpleName, xattrsToken, version)
+	indexName := fmt.Sprintf(indexNameFormat, i.simpleName, xattrsToken, version)
+	if i.partionable && numPartitions > 1 {
+		indexName = fmt.Sprintf("%s_p%d", indexName, numPartitions)
```
adamcfraser: Would be nice to define `%s_p%d` as a const (partitionedIndexNameFormat) so that it can be documented at the top of this file for quick reference beside `indexNameFormat`.

db/indexes.go
```diff
@@ -281,8 +293,10 @@ func (i *SGIndex) shouldCreate(options InitializeIndexOptions) bool {
 // Creates index associated with specified SGIndex if not already present.  Always defers build - a subsequent BUILD INDEX
 // will need to be invoked for any created indexes.
 func (i *SGIndex) createIfNeeded(ctx context.Context, bucket base.N1QLStore, options InitializeIndexOptions) error {
-
-	indexName := i.fullIndexName(options.UseXattrs)
+	if options.NumPartitions < 1 {
```
adamcfraser, line 297: Elsewhere if `options.NumPartitions<=1`, it's treated as "do not use partitions" (i.e. the check seems to consistently be `if options.NumPartitions > 1`).  This might be the right approach, just want to make sure we're not triggering a fatal error when options.NumPartitions isn't initialized (is zero), if it's not actually a fatal condition.

integration-test/start_server.sh
```diff
@@ -71,7 +71,7 @@ if [ "${MULTI_NODE:-}" == "true" ]; then
     ${DOCKER_COMPOSE} up -d --force-recreate --renew-anon-volumes --remove-orphans
 else
     # single node
-    docker run --rm -d --name ${SG_TEST_COUCHBASE_SERVER_DOCKER_NAME} --volume "${WORKSPACE_ROOT}:/workspace" -p 8091-8096:8091-8096 -p 11207:11207 -p 11210:11210 -p 11211:11211 -p 18091-18094:18091-18094 "${COUCHBASE_DOCKER_IMAGE_NAME}"
+    docker run --rm -d --name ${SG_TEST_COUCHBASE_SERVER_DOCKER_NAME} --volume "${WORKSPACE_ROOT}:/workspace" -p 8091-8096:8091-8096 -p 9102:9102 -p 11207:11207 -p 11210:11210 -p 11211:11211 -p 18091-18094:18091-18094 -p 19102 "${COUCHBASE_DOCKER_IMAGE_NAME}"
```
adamcfraser:
```suggestion
    docker run --rm -d --name ${SG_TEST_COUCHBASE_SERVER_DOCKER_NAME} --volume "${WORKSPACE_ROOT}:/workspace" -p 8091-8096:8091-8096 -p 9102:9102 -p 11207:11207 -p 11210:11210 -p 11211:11211 -p 18091-18094:18091-18094 -p 19102:19102 "${COUCHBASE_DOCKER_IMAGE_NAME}"
```
I'm guessing this is needed based on other ports, but I'm not a docker expert.

base/collection.go
```diff
@@ -468,6 +468,19 @@ func (b *GocbV2Bucket) QueryEpsCount() (int, error) {
 	return len(agent.N1qlEps()), nil
 }
 
+// GSIEps returns the GSI endpoints for the bucket for querying information about indexes.
+func (b *GocbV2Bucket) GSIEps() (url []string, err error) {
```
adamcfraser: Requires access to an internal port (9102), might want to call that out for future usage (currently test-only).

base/bucket_n1ql.go
```diff
@@ -20,7 +20,8 @@ const PrimaryIndexName = "#primary"
 
 // IndexOptions used to build the 'with' clause
 type N1qlIndexOptions struct {
-	NumReplica      uint `json:"num_replica,omitempty"`          // Number of replicas
-	IndexTombstones bool `json:"retain_deleted_xattr,omitempty"` // Whether system xattrs on tombstones should be indexed
-	DeferBuild      bool `json:"defer_build,omitempty"`          // Whether to defer initial build of index (requires a subsequent BUILD INDEX invocation)
+	NumReplica      uint    `json:"num_replica,omitempty"`          // Number of replicas
+	IndexTombstones bool    `json:"retain_deleted_xattr,omitempty"` // Whether system xattrs on tombstones should be indexed
+	DeferBuild      bool    `json:"defer_build,omitempty"`          // Whether to defer initial build of index (requires a subsequent BUILD INDEX invocation)
+	NumPartitions   *uint32 `json:"num_partition"`                  // The number of partitions to use for the index. 1 will be a non-partitioned index.
```
adamcfraser: I think we want omitempty here to avoid sending {"num_partition":null} when NumPartitions isn't defined?

adamcfraser (changes requested):
Generally looks great, a few minor comments/questions.

db/indexes.go
torcolvin: I split the validation for this, once you get to this part of the code we do expect  >= 1 to make it simpler to not have to convert everytime here and do a conversion from uninitialized to set to one. I thought this was conceptually easier because it forces the caller to set this value, so the caller decides when to actually make the indexes partitionable.

This was easier for me in testing to make sure I was passing this value through the code but I'm not committed to this approach.

integration-test/start_server.sh
torcolvin: I added some extra ports based on public documentation and fixed this.

adamcfraser (approved):

PR #7429, opened 2025-03-20
CBG-4540: Clone byte slices coming over DCP when transforming into `sgbucket.FeedEvent`

bbrks:
CBG-4540

- Clone the byte slices coming over DCP when building `FeedEvent` to ensure independent memory allocations of data between gocbcore and Sync Gateway.
- Guards against potential cache corruption in the event of gocbcore/memd refactors to reuse the allocated byte slices in the `memd.Packet` `sync.Pool`
- Avoids retaining `memd.Packet` memory when we keep a direct reference to `Key` and/or `Value` byte slices in our caches.
  - Should significantly improve caching feed memory usage since we're only storing `Key`.
  - Should slightly improve revcache usage since we're storing both.
- Overall it's safer to _always_ do this copy, when crossing the gocbcore/memd/sgbucket package boundaries than conditionally when we're about to store things in the SG caches - even though we may be paying for a copy without needing it.
- Performance testing has shown that the bytes copy implementation doesn't significantly impact import throughput (the copy is very cheap compared to JSON unmarshalling for instance) and _improves_ memory usage associated with import because of above reasons for more efficient `sync.Pool` reuse.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2996/

### DIFF
--- a/base/dcp_client_stream_event.go
+++ b/base/dcp_client_stream_event.go
@@ -54,15 +54,14 @@ type mutationEvent struct {
 }
 
 // asFeedEvent converts a mutationEvent to a sgbucket.FeedEvent.
-// The byte slices must by copied to ensure that memory associated with the memd mutationEvent and Packet are independent and can be released or reused by gocbcore as needed.
 func (e mutationEvent) asFeedEvent() sgbucket.FeedEvent {
 	return sgbucket.FeedEvent{
 		Opcode:       sgbucket.FeedOpMutation,
 		Flags:        e.flags,
 		Expiry:       e.expiry,
 		CollectionID: e.collection,
-		Key:          EfficientBytesClone(e.key),
-		Value:        EfficientBytesClone(e.value),
+		Key:          e.key,
+		Value:        e.value,
 		DataType:     e.datatype,
 		Cas:          e.cas,
 		RevNo:        e.revNo,
@@ -84,13 +83,12 @@ type deletionEvent struct {
 }
 
 // asFeedEvent converts a deletionEvent to a sgbucket.FeedEvent.
-// The byte slices must be copied to ensure that memory associated with the memd mutationEvent and Packet are independent and can be released or reused by gocbcore as needed.
 func (e deletionEvent) asFeedEvent() sgbucket.FeedEvent {
 	return sgbucket.FeedEvent{
 		Opcode:       sgbucket.FeedOpDeletion,
 		CollectionID: e.collection,
-		Key:          EfficientBytesClone(e.key),
-		Value:        EfficientBytesClone(e.value),
+		Key:          e.key,
+		Value:        e.value,
 		DataType:     e.datatype,
 		Cas:          e.cas,
 		RevNo:        e.revNo,
--- a/base/dcp_client_stream_event.go
+++ b/base/dcp_client_stream_event.go
@@ -54,7 +54,7 @@ type mutationEvent struct {
 }
 
 // asFeedEvent converts a mutationEvent to a sgbucket.FeedEvent.
-// The byte slices are copied to ensure that memory associated with the memd mutationEvent and Packet are independent and can be released or reused by gocbcore as needed.
+// The byte slices must by copied to ensure that memory associated with the memd mutationEvent and Packet are independent and can be released or reused by gocbcore as needed.
 func (e mutationEvent) asFeedEvent() sgbucket.FeedEvent {
 	return sgbucket.FeedEvent{
 		Opcode:       sgbucket.FeedOpMutation,
@@ -84,7 +84,7 @@ type deletionEvent struct {
 }
 
 // asFeedEvent converts a deletionEvent to a sgbucket.FeedEvent.
-// The byte slices are copied to ensure that memory associated with the memd mutationEvent and Packet are independent and can be released or reused by gocbcore as needed.
+// The byte slices must be copied to ensure that memory associated with the memd mutationEvent and Packet are independent and can be released or reused by gocbcore as needed.
 func (e deletionEvent) asFeedEvent() sgbucket.FeedEvent {
 	return sgbucket.FeedEvent{
 		Opcode:       sgbucket.FeedOpDeletion,
--- a/base/dcp_client_stream_event.go
+++ b/base/dcp_client_stream_event.go
@@ -54,14 +54,15 @@ type mutationEvent struct {
 }
 
 // asFeedEvent converts a mutationEvent to a sgbucket.FeedEvent.
+// The byte slices are copied to ensure that memory associated with the memd mutationEvent and Packet are independent and can be released or reused by gocbcore as needed.
 func (e mutationEvent) asFeedEvent() sgbucket.FeedEvent {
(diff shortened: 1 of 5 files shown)


base/dcp_common.go
```diff
@@ -522,15 +522,16 @@ func dcpKeyFilter(key []byte, metaKeys *MetadataKeys) bool {
 }
 
 // Makes a feedEvent that can be passed to a FeedEventCallbackFunc implementation
+// The byte slices are copied to ensure that memory associated with the memd mutationEvent and Packet are independent and can be released or reused by gocbcore as needed.
 func makeFeedEvent(key []byte, value []byte, dataType uint8, cas uint64, expiry uint32, vbNo uint16, collectionID uint32, opcode sgbucket.FeedOpcode) sgbucket.FeedEvent {
 
 	// not currently doing rq.Extras handling (as in gocouchbase/upr_feed, makeUprEvent) as SG doesn't use
```
torcolvin: Can you update this comment? I think that it can probably be removed entirely, but we shouldn't refer to cbdatastource or gocouchbase anymore.

I think it might be worth adding another comment here that _any_ byte slices passed into this function need to be copied. I tried to think if there was a way in the compiler that we can force that these values get copied if a new value is added. OTOH, there's nothing really going to change about the memcached protocol probably so maybe this is over cautious.

For your information, we do have expiry here. flags are not useful to us ever, they are arbitrarily set by the SDK and can be used to speed up some of their transcoding. But they are not guaranteed to ever be set, since e.g. C API can't set the flags. We use datatype to determine information about the document. I can't see us ever needing to handle snapshots because we don't care about snapshot boundaries, these would only be useful for checkpoints and those are handled separately.

https://github.com/couchbase/kv_engine/blob/master/docs/dcp/documentation/commands/mutation.md for documentation and https://github.com/couchbase/kv_engine/blob/master/docs/dcp/documentation/commands/extended_meta/ext_meta_ver1.md for why extras is not populated or something we need to use.

torcolvin (commented):
LGTM

base/dcp_common.go
```diff
@@ -522,15 +522,15 @@ func dcpKeyFilter(key []byte, metaKeys *MetadataKeys) bool {
...
 
 // Makes a feedEvent that can be passed to a FeedEventCallbackFunc implementation
+// The byte slices must be copied to ensure that memory associated with the memd mutationEvent and Packet are independent and can be released or reused by gocbcore as needed.
 func makeFeedEvent(key []byte, value []byte, dataType uint8, cas uint64, expiry uint32, vbNo uint16, collectionID uint32, opcode sgbucket.FeedOpcode) sgbucket.FeedEvent {
 
-	// not currently doing rq.Extras handling (as in gocouchbase/upr_feed, makeUprEvent) as SG doesn't use
-	// expiry/flags information, and snapshot handling is done by cbdatasource and sent as
-	// SnapshotStart, SnapshotEnd
+	// not currently doing rq.Extras handling (as was done in go-couchbase/upr_feed, makeUprEvent) since SG doesn't use
+	// flags information, and snapshot handling is done by cbgt and sent as SnapshotStart, SnapshotEnd
```
torcolvin: I'm being fussy here but in further consideration I think we should drop this comment.

It's unlikely we'd _ever_ get an values like `SnapshotStart` / `SnapshotEnd` and there are other messages that are also ignored. https://github.com/couchbase/gocbcore/blob/b7d79c7b009d1084bcf0caef2fef3d7592b80c41/dcp.go#L211 lists the types of messages processed and we really only have mutation/deletion because other messages don't even _have_ a key or value necessarily. 

I think the comment was helpful in the migration but now just refers to an API that no longer exists.

bbrks: > I think the comment was helpful in the migration but now just refers to an API that no longer exists.

That's fair - and I think I agree. Will just remove it.

bbrks:
```suggestion
```

torcolvin (dismissed):

base/dcp_client_stream_event.go
```diff
@@ -54,14 +54,15 @@ type mutationEvent struct {
 }
 
 // asFeedEvent converts a mutationEvent to a sgbucket.FeedEvent.
+// The byte slices must by copied to ensure that memory associated with the memd mutationEvent and Packet are independent and can be released or reused by gocbcore as needed.
```
adamcfraser: Would there be a benefit in doing the bytes clone when we create the mutationEvent/deletionEvent (in dcp_client_stream_observer), so that the copy is done before the event is potentially waiting for some time in the DCPWorker queue?  It might not be a significant difference in time under most conditions, but doing it as we process the gocbcore.DcpMutation seems like it might be the right place.

bbrks: I did originally have an implementation that did it there, and I'm OK moving it back. I only had a few small reasons for pushing it down one more level, and wasn't thinking about the delay in callback invocation:

- Both CBGT/normal DCP feeds build a `FeedEvent`, so the crossover between DCP clients and SG code was more clear here across both DCP implementations. Also made it easier to cover all 3 with a single test.
- We lose the `dcpKeyFilter` early-return before doing the copy. Given the key prefixes we check for and exclude, this isn't a big deal, since those metadata docs aren't that large in volume or size.

base/dcp_common.go
```diff
@@ -522,15 +522,13 @@ func dcpKeyFilter(key []byte, metaKeys *MetadataKeys) bool {
...
 func makeFeedEvent(key []byte, value []byte, dataType uint8, cas uint64, expiry uint32, vbNo uint16, collectionID uint32, opcode sgbucket.FeedOpcode) sgbucket.FeedEvent {
 
-	// not currently doing rq.Extras handling (as in gocouchbase/upr_feed, makeUprEvent) as SG doesn't use
-	// expiry/flags information, and snapshot handling is done by cbdatasource and sent as
-	// SnapshotStart, SnapshotEnd
 	event := sgbucket.FeedEvent{
 		Opcode:       opcode,
-		Key:          key,
-		Value:        value,
+		Key:          EfficientBytesClone(key),
```
adamcfraser, line 530: This cloning can be removed now that the incoming key/value are already copied, correct?

adamcfraser (changes requested):

base/dcp_client_stream_event.go
adamcfraser: I agree about the key filtering, but I think the key filtering should probably get the same handling in future (potentially in 3.3.0) - move it up so it gets performed prior to the event being added to the worker queue.

base/dcp_common.go
bbrks: This codepath is the one still used via CBGT (Invoked via `DCPDest`->`DataUpdate`->`makeFeedEventForDest`) for EE/Sharded import.

The other ones that changed with `mutationEvent`/`deletionEvent` are for the caching feed via `StartGocbDCPFeed`->`DCPClient`->`Mutation`

adamcfraser (approved):

PR #7432, opened 2025-03-21
CBG-4570: stop audit events logging when no audit config is provided at db level

gregns1:
CBG-4570

- When not providing a db level audit log config and you have global audit enabled you will get all audit events logged for that db
- To fix we should explicitly set enabled to false in audit log config on database when no config is povided
- This PR incudes a lot of test fixes too, found that some tests were wrongly asserting on events 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/2999/

### DIFF
--- a/base/util.go
+++ b/base/util.go
@@ -896,10 +896,6 @@ func IntPtr(i int) *int {
 	return &i
 }
 
-func UintSlicePtr(u []uint) *[]uint {
-	return &u
-}
-
 // IntDefault returns ifNil if i is nil, or else returns dereferenced value of i
 func IntDefault(i *int, ifNil int) int {
 	if i != nil {
--- a/base/util.go
+++ b/base/util.go
@@ -896,6 +896,10 @@ func IntPtr(i int) *int {
 	return &i
 }
 
+func UintSlicePtr(u []uint) *[]uint {
+	return &u
+}
+
 // IntDefault returns ifNil if i is nil, or else returns dereferenced value of i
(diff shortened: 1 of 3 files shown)


base/util.go
```diff
@@ -896,6 +896,10 @@ func IntPtr(i int) *int {
 	return &i
 }
 
+func UintSlicePtr(u []uint) *[]uint {
+	return &u
+}
+
```
bbrks: There's a new-ish generic `Ptr()` function that will take any type now :) we can rip out/not introduce new type-specific ones

rest/config.go
```diff
@@ -2298,7 +2298,11 @@ func RegisterSignalHandler(ctx context.Context) {
 func (c *DbConfig) toDbLogConfig(ctx context.Context) *base.DbLogConfig {
 	l := c.Logging
 	if l == nil || (l.Console == nil && l.Audit == nil) {
-		return nil
+		return &base.DbLogConfig{
+			Audit: &base.DbAuditLogConfig{
+				Enabled: false,
```
bbrks:
```suggestion
				Enabled: base.DefaultDbAuditEnabled,
```

bbrks (changes requested):
Looks good - just a couple of tweaks to tidy up

base/util.go
gregns1: I had no idea this was there! Thanks!

bbrks (approved):

PR #7436, opened 2025-03-27
CBG-4578: have maximum threshold for releasing sequences in nextSequenceGreaterThan

gregns1:
CBG-4578

- New error type to be returned for when number of sequcnes to release is above a threshold 
- Tests for doc update, resync, import feed and on demand import 
- New stat for when we encounter this, this will be used for capella alerting 
- Will log at warning level when this is encountered at teh sequcne allocator and document update level (this can be changed if this is too mcuh warning logging) 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3005/

### DIFF
--- a/base/stats.go
+++ b/base/stats.go
@@ -1756,7 +1756,7 @@ func (d *DbStats) initDatabaseStats() error {
 	if err != nil {
 		return err
 	}
-	resUtil.CorruptSequenceCount, err = NewIntStat(SubsystemDatabaseKey, "corrupt_sequence_count", StatUnitNoUnits, CorruptSequenceCountDesc, StatAddedVersion3dot3dot0, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, labelKeys, labelVals, prometheus.CounterValue, 0)
+	resUtil.CorruptSequenceCount, err = NewIntStat(SubsystemDatabaseKey, "corrupt_sequence_count", StatUnitNoUnits, CorruptSequenceCountDesc, StatAddedVersion3dot2dot4, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, labelKeys, labelVals, prometheus.CounterValue, 0)
 	if err != nil {
 		return err
 	}
--- a/base/stats.go
+++ b/base/stats.go
@@ -645,6 +645,8 @@ type DatabaseStats struct {
 	SequenceReleasedCount *SgwIntStat `json:"sequence_released_count"`
 	// The total number of sequences reserved by Sync Gateway.
 	SequenceReservedCount *SgwIntStat `json:"sequence_reserved_count"`
+	// The total number of corrupt sequences above the MaxSequencesToRelease threshold seen at the sequence allocator
+	CorruptSequenceCount *SgwIntStat `json:"corrupt_sequence_count"`
 	// The total number of warnings relating to the channel name size.
 	WarnChannelNameSizeCount *SgwIntStat `json:"warn_channel_name_size_count"`
 	// The total number of warnings relating to the channel count exceeding the channel count threshold.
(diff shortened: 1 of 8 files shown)


base/stats.go
```diff
@@ -1754,6 +1756,10 @@ func (d *DbStats) initDatabaseStats() error {
 	if err != nil {
 		return err
 	}
+	resUtil.CorruptSequenceCount, err = NewIntStat(SubsystemDatabaseKey, "corrupt_sequence_count", StatUnitNoUnits, CorruptSequenceCountDesc, StatAddedVersion3dot3dot0, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, labelKeys, labelVals, prometheus.CounterValue, 0)
```
bbrks: Maybe expect this to be going into 3.2.4 as well 

```suggestion
	resUtil.CorruptSequenceCount, err = NewIntStat(SubsystemDatabaseKey, "corrupt_sequence_count", StatUnitNoUnits, CorruptSequenceCountDesc, StatAddedVersion3dot2dot4, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, labelKeys, labelVals, prometheus.CounterValue, 0)
```

db/crud.go
```diff
@@ -1882,6 +1882,9 @@ func (col *DatabaseCollectionWithUser) documentUpdateFunc(ctx context.Context, d
 
 	unusedSequences, err = col.assignSequence(ctx, previousDocSequenceIn, doc, unusedSequences)
 	if err != nil {
+		if errors.Is(err, base.ErrMaxSequenceReleasedExceeded) {
+			base.WarnfCtx(ctx, "Doc %s / %s had an existing sequence %d that is more than %d larger than expected. Document update will be cancelled. This may indicate documents being migrated between databases by an external process.", base.UD(newDoc.ID), prevCurrentRev, doc.Sequence, MaxSequencesToRelease)
```
bbrks: Since this indicates manual intervention to strip doc metadata, I think this is one of few times that Error level is appropriate to log at.

Also reworded log message and used `previousDocSequenceIn` to be safe (in case `col.assignSequence` changed `doc.Sequence`

```suggestion
			base.ErrorfCtx(ctx, "Doc %s / %s had a much larger sequence (%d) than the current sequence number. Document update will be cancelled, since we don't want to allocate sequences to fill a gap this large. This may indicate document metadata being migrated between databases where it should've been stripped and re-imported.", base.UD(newDoc.ID), prevCurrentRev, previousDocSequenceIn, MaxSequencesToRelease)
```

db/sequence_allocator.go
```diff
@@ -286,6 +289,15 @@ func (s *sequenceAllocator) nextSequenceGreaterThan(ctx context.Context, existin
 	numberToRelease := existingSequence - syncSeq
 	numberToAllocate := s.sequenceBatchSize
 	incrVal := numberToRelease + numberToAllocate
+
+	// if sequences to release are above the max allowed, return error to cancel update
+	if numberToRelease > MaxSequencesToRelease {
+		base.WarnfCtx(ctx, "Number of sequences to release (%d) to catch up with doc current sequence (%d) exceeds MaxSequencesToRelease (%d). Doc update will be cancelled.", numberToRelease, existingSequence, MaxSequencesToRelease)
```
bbrks: Already effectively doing this logging (with more contextual data (doc ID)) in the caller.
Would prefer to remove this unless there's information here that we cannot log above. I can see we're not capturing `numberToRelease` in caller, but I don't think that's important to know.

bbrks (changes requested):
A few comments mostly around the logging - open to different ideas or opinions though

base/stats.go
gregns1: Ah yeah good point, if we have 3.2.4 added version will that break anything if we backport to 3.2.2.1?

db/crud.go
gregns1: I agree, moved it to error. I've left it as doc.Sequence as upon using `previousDocSequenceIn` I was getting 0 sequence values in the log line. It looks like from inside `updateAndReturnDoc` this value is always 0 when its passed into `documentUpdateFunc`. It seems to only be used post `documentUpdateFunc` after a sequence is allocated. When i fixed this (to the actual current doc seq value) I had a few test failures, looking at the code inside `assignSequence` it drops into this https://github.com/couchbase/sync_gateway/blob/5bd0d640ad377b80a2c385023d0381d10b0919a3/db/crud.go#L1669 when the sequence that has been passed in is non zero. The more I look at this code I feel like something is not right here... but i don't want to change this code too much given we hope to get 3.2.2.1 out with minimal changes and I thought we should keep risk low? Happy to hear your thoughts though.

Anyway I think its fine as `doc.Sequence` as we won't have assigned the seq to the doc if we hit this at that point.

base/stats.go
bbrks: Nah this version is just for documentation purposes and we can forget about an MP version for that

db/crud.go
bbrks: Sounds good, agree with minimizing the changes

bbrks (approved):

db/crud.go
gregns1: Upon review of the code `previousDocSequenceIn` being 0 when passed into the this function makes sense. Given the callback this function sits in code that can be invoked multiple times (for example, if we got CAS failure that we should retry on). It's used for `unused_sequence` handling so the change cache doesn't get hung up on not receiving sequences that were allocated but are unusable.

May file an enhancement for future to improve variable naming/add some comments around the doc sequence variable being passed down from the callback to ensure no confusion in future.

PR #7437, opened 2025-03-27
CBG-4503 use natural language for openapi

torcolvin:
Update/remove any summaries where it doesn't have a natural language summary.

- Remove `head` section anytime it can't be used to check existence.
- Remove unused `OpenID Connect` tag from admin API


docs/api/paths/admin/db-_ensure_full_commit.yaml
```diff
@@ -8,7 +8,7 @@
 parameters:
   - $ref: ../../components/parameters.yaml#/db
 post:
-  summary: '/{db}/_ensure_full_commit'
+  summary: 'Commits any recent changes to the specified database to disk'
```
bbrks: Since it doesn't really do anything maybe best to keep it plain

```suggestion
  summary: Ensure Full Commit
```

docs/api/paths/public/db-_ensure_full_commit.yaml
```diff
@@ -8,7 +8,7 @@
 parameters:
   - $ref: ../../components/parameters.yaml#/db
 post:
-  summary: '/{db}/_ensure_full_commit'
+  summary: Commits any recent changes to the specified database to disk
```
bbrks: ditto

```suggestion
  summary: Ensure Full Commit
```

docs/api/paths/admin/keyspace-_raw-docid.yaml
```diff
@@ -113,7 +113,7 @@ get:
     - Document
   operationId: get_keyspace-_raw-docid
 head:
-  summary: /{keyspace}/_raw/{docid}
+  summary: Check if a document exists
```
bbrks: I think this would be `HEAD /db/docid` rather than the `/_raw/` version of the endpoint. Would prefer to just remove this head entry in favor of that one.

bbrks (changes requested):

bbrks (approved):

PR #7458, opened 2025-04-02
CBG-4581 - Add sequence value stats for last allocated and last reserved

bbrks:
CBG-4581

Add new stats to track last allocated and last reserved sequence values
- Centralized sequence stat updates into the two main functions `_incrementSequence` and `_nextSequence`
  - Additional locking required for new value stats to ensure concurrent sequence allocations do not overwrite previous values, so stat updates moved inside the mutex guarded methods.
  - Deduplicates stat updates by moving them down one level.
- Extended existing test assertion helper `assertNewAllocatorStats` to cover new seq value stats.

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3019/

### DIFF
--- a/db/sequence_allocator.go
+++ b/db/sequence_allocator.go
@@ -179,14 +179,11 @@ func (s *sequenceAllocator) nextSequence(ctx context.Context) (sequence uint64, 
 	if err != nil {
 		return 0, err
 	}
-
 	// If sequences were reserved, send notification to the release sequence monitor, to start the clock for releasing these sequences.
 	// Must be done after mutex is released.
 	if sequencesReserved {
 		s.reserveNotify <- struct{}{}
 	}
-
-	s.dbStats.SequenceAssignedCount.Add(1)
 	return sequence, nil
 }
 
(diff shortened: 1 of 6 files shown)


db/sequence_allocator.go
```diff
@@ -229,6 +225,7 @@ func (s *sequenceAllocator) nextSequenceGreaterThan(ctx context.Context, existin
 	if targetSequence <= s.max {
 		releaseFrom := s.last + 1
 		s.last = targetSequence
+		s.dbStats.LastSequenceAssignedValue.Set(int64(targetSequence))
```
torcolvin, line 228: Does it make sense to abstract this into a function so if `s.last` is set then this is always incremented in case we move this?

Should `s.last` be named `s._last` to indicate you need a lock for read/write?

bbrks: I started to do this refactor and it unraveled a slight difference between `s.last`'s value and an actually allocated/assigned sequence (reported via the stat), which aren't always the same - in cases of releasing unused sequences or allocating a new batch without allocating a sequence.

It made stuff more complicated for the special cases where we don't want to update the stat, and it actually broke stuff in my attempts.

I'm 50/50 about naming as well. The mutex guards most of the whole struct really, and renaming `_last`, `_max`, `_sequenceBatchSize`, and `_lastSequenceReserveTime` becomes pretty noisy in diff and readability.

torcolvin (approved):

PR #7463, opened 2025-04-03
CBG-4599 update to golanci-lint v2 configs

torcolvin:
CBG-4599 update to golanci-lint v2 configs

Don't merge this until devs know to update golangci-lint locally and we update the maintainance release branches also.

- Used `golangci-lint migrate` and then re-added the comments.
- Manually added any staticchecks that fail

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,9 +60,9 @@ jobs:
         with:
           go-version: 1.24.2
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v6
+        uses: golangci/golangci-lint-action@v7
         with:
-          version: v1.64.8
+          version: v2.0.2
           args: --config=.golangci-strict.yml
 
   test:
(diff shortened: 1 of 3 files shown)


bbrks:
Gonna merge this since my local dev environment has been broken for weeks already anyway.

The JetBrains plugin still does not yet work with the new output format flags, and requires a wrapper as a temporary fix until the plugin is updated. I think this is soon since there's already an RC that fixes this.

https://plugins.jetbrains.com/plugin/12496-go-linter

```sh
#!/bin/bash

set -o errexit
set -o pipefail
set -o nounset
# set -o xtrace

# This script is a bridge to use golangci-lint v2 with the v1 --out-format flag.
# It allows JetBrains IDEs to remain usable while we migrate to v2.
main ()
{
    tmpfile="$(mktemp)"
    trap 'rm -f "${tmpfile}"' EXIT

    local -a args=()
    local prev_arg=""

    for arg in "$@"
    do
        if [ "${prev_arg}" = "--out-format" ]
        then
            args+=("${tmpfile}")
            prev_arg=""
            continue
        fi

        if [ "${arg}" = "--out-format" ]
        then
            args+=("--output.json.path")
        else
            args+=("${arg}")
        fi

        prev_arg="${arg}"
    done

    golangci-lint "${args[@]}"
    cat "${tmpfile}"
}

main "$@"
```

bbrks (approved):

PR #7462, opened 2025-04-03
CBG-4596 use go 1.24.2

torcolvin:
CBG-4596 use go 1.24.2

This commit is specific to release/3.2.4. Did `git checkout origin/main .github/workflows/ci.yml` to pull in updated versions for github actions and Jenkinsfile to pick up version for pipeline jobs.

PR #7465, opened 2025-04-03
[3.2.4] CBG-4564 update go deps post go 1.24 upgrade

torcolvin:
follow up to https://github.com/couchbase/sync_gateway/pull/7462 and https://github.com/couchbase/sync_gateway/pull/7444 since we decided to bump to go 1.24.2 on 3.2.4

did a `git checkout origin/main go.mod` and rolled back go-jose v2 update, gocb/cbgt upgrades.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3023/


go.mod
```diff
@@ -1,11 +1,11 @@
 module github.com/couchbase/sync_gateway
 
-go 1.22.0
+go 1.23.0
```
bbrks, line 3:
```suggestion
go 1.24.0
```

or not? I don't actually know

torcolvin: I could bump this requirement, but we _can_ build on 1.23.0 so I left it here because it matches what we did on main. My thought was we don't need to bump this until a dependency requires it? But it doesn't matter either way.

I think there's something to putting `toolchain 1.24.2` in the files to enforce using that version of go, which is sort of a belt and suspenders approach to also remembering to update the Jenkinsfile, github actions, product-config.json - since it will enforce using that version. It would also force using that version locally.

I'm inclined to leave this as is for release/3.2.4 though and puzzle the right thing to do on main when it is less likely to break builds.

bbrks: That's fine by me. Is this PR now ready for review/merge?

bbrks (approved):

PR #7467, opened 2025-04-03
CBG-4546: Detect loading non-xattr config document in xattr bootstrap mode

bbrks:
CBG-4546

- Add stat to track loading non-xattr config document in xattr bootstrap mode
- Change debug to warning and make slightly more descriptive (retain original message for grep)

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3024/

### DIFF
--- a/base/config_persistence.go
+++ b/base/config_persistence.go
@@ -108,7 +108,7 @@ func (xbp *XattrBootstrapPersistence) loadRawConfig(ctx context.Context, c *gocb
 		xattrContErr := res.ContentAt(0, &rawValue)
 		if xattrContErr != nil {
 			SyncGatewayStats.GlobalStats.ConfigStat.XattrFormatMismatches.Add(1)
-			WarnfCtx(ctx, "Found config document but No xattr config found for key=%s, path=%s: %v", key, cfgXattrConfigPath, xattrContErr)
+			DebugfCtx(ctx, KeyCRUD, "Found config document but No xattr config found for key=%s, path=%s: %v", key, cfgXattrConfigPath, xattrContErr)
 			return rawValue, 0, ErrNotFound
 		}
 		return rawValue, res.Cas(), nil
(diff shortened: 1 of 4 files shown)


base/config_persistence.go
```diff
@@ -107,7 +107,8 @@ func (xbp *XattrBootstrapPersistence) loadRawConfig(ctx context.Context, c *gocb
 		// config
 		xattrContErr := res.ContentAt(0, &rawValue)
 		if xattrContErr != nil {
-			DebugfCtx(ctx, KeyCRUD, "No xattr config found for key=%s, path=%s: %v", key, cfgXattrConfigPath, xattrContErr)
+			SyncGatewayStats.GlobalStats.ConfigStat.XattrFormatMismatches.Add(1)
+			WarnfCtx(ctx, "Found config document but No xattr config found for key=%s, path=%s: %v", key, cfgXattrConfigPath, xattrContErr)
```
torcolvin:
```suggestion
			WarnfCtx(ctx, "Found config document %s without expected xattr %s: %v, treating as not found", key, cfgXattrConfigPath, xattrContErr)
```

base/config_persistence.go
```diff
@@ -181,7 +182,8 @@ func (xbp *XattrBootstrapPersistence) loadConfig(ctx context.Context, c *gocb.Co
 		// config
 		xattrContErr := res.ContentAt(0, valuePtr)
 		if xattrContErr != nil {
-			DebugfCtx(ctx, KeyCRUD, "No xattr config found for key=%s, path=%s: %v", key, cfgXattrConfigPath, xattrContErr)
+			SyncGatewayStats.GlobalStats.ConfigStat.XattrFormatMismatches.Add(1)
+			WarnfCtx(ctx, "Found config document but No xattr config found for key=%s, path=%s: %v", key, cfgXattrConfigPath, xattrContErr)
```
torcolvin:
```suggestion
			WarnfCtx(ctx, "Found config document %s without expected xattr %s: %v, treating as not found", key, cfgXattrConfigPath, xattrContErr)

```

torcolvin (commented):
I'm a bit worried about changing this debug to warn.

This is only going to affect app services, so we don't have to worry about on prem customers. However, I think this will log warning/stat every 10s for potentially 2 or more documents (registry + dbconfig). Will this warning end up getting shown to customers? Is this going to needlessly fill up warning logs which might not be necessary since app services always logs at debug anyway?

base/config_persistence.go
bbrks: This is worded weirdly to keep the original message intact. I know support already use it for grepping for this issue.

bbrks:
I agree about the change to warn potentially being too disruptive. We'll find out who's doing this with the stat even without the logs so I will put it back.

base/config_persistence.go
bbrks:
```suggestion
			DebugfCtx(ctx, KeyCRUD, "Found config document but No xattr config found for key=%s, path=%s: %v", key, cfgXattrConfigPath, xattrContErr)
```

base/config_persistence.go
bbrks:
```suggestion
			DebugfCtx(ctx, KeyCRUD, "Found config document but No xattr config found for key=%s, path=%s: %v", key, cfgXattrConfigPath, xattrContErr)
```

torcolvin:
LGTM

torcolvin (approved):

PR #7474, opened 2025-04-08
CBG-4602: benchmarking for procesEntry

gregns1:
CBG-4602

Benchmarking for process entry. This can be an iterative approach but for now we have something in place to generate data and throw at `processEntry`. We can build upon this in future if we find we need more data points. 

High level view is we have a sgw node abstraction each with their own allocator which links to central atomic counter much like sync seq. We batch allocations to nodes and this is configurable, along with number of nodes and the delays associated with nodes (simulating nodes writing at differing speeds. We have a max delay of 150 ms as this is the slowest a nodes can write through 10 sequences without releasing any and reallocating a batch. 

We needed some methods adding to testing files in db package to expose the `processEntry` and init of change cache to the testing tool itself. 

## Pre-review checklist

### DIFF
--- a/tools/cache_perf_tool/main.go
+++ b/tools/cache_perf_tool/main.go
@@ -67,19 +67,19 @@ func main() {
 		cpuProfBuf := bytes.Buffer{}
 		err = pprof.StartCPUProfile(&cpuProfBuf)
 		if err != nil {
-			log.Println("Error starting CPU profile: %v", err)
+			log.Printf("Error starting CPU profile: %v", err)
 			return
 		}
 		defer func() {
 			pprof.StopCPUProfile()
 			cpuProfileBuffer, err := io.ReadAll(&cpuProfBuf)
 			if err != nil {
-				log.Println("error reading cpuProfBuf: %v", err)
+				log.Printf("error reading cpuProfBuf: %v", err)
 				return
 			}
 			err = os.WriteFile("cpu.prof", cpuProfileBuffer, os.ModePerm)
 			if err != nil {
-				log.Println("error writing cpu profile to file: %v", err)
+				log.Printf("error writing cpu profile to file: %v", err)
 				return
 			}
 		}()
@@ -104,7 +104,7 @@ func main() {
 		},
 	})
 	if err != nil {
-		log.Println("Error creating database context: %v", err)
+		log.Printf("Error creating database context: %v", err)
 		return
 	}
 	defer dbContext.Close(ctx)
@@ -124,7 +124,7 @@ func main() {
 			}
 			// need to have a delay for each node defined so we have variable write throughput
 			if len(delayList) != *nodes-1 {
-				log.Println("invalid number of delays: %d", len(delayList))
+				log.Printf("invalid number of delays: %d", len(delayList))
 				return
 			}
 		}
@@ -132,7 +132,7 @@ func main() {
 		// create new sgw node abstraction and spawn write goroutines
 		p.spawnDocCreationGoroutine(ctx)
 	} else {
-		log.Println("Invalid mode: %s", *mode)
+		log.Printf("Invalid mode: %s", *mode)
 		return
 	}
 
@@ -209,13 +209,13 @@ func heapProfiling(ctx context.Context) {
 			heapProfBuf := bytes.Buffer{}
 			err := pprof.WriteHeapProfile(&heapProfBuf)
 			if err != nil {
-				log.Println("Error writing heap profile: %v", err)
+				log.Printf("Error writing heap profile: %v", err)
 				return
 			}
 
 			err = os.WriteFile(fileName, heapProfBuf.Bytes(), os.ModePerm)
 			if err != nil {
-				log.Println("Error writing heap profile to file: %v", err)
+				log.Printf("Error writing heap profile to file: %v", err)
 				return
 			}
 		}
@@ -236,12 +236,12 @@ func mutexProfiling(ctx context.Context) {
 			mutexProfBuf := bytes.Buffer{}
 			err := pprof.Lookup("mutex").WriteTo(&mutexProfBuf, 0)
 			if err != nil {
-				log.Println("Error writing mutex profile: %v", err)
+				log.Printf("Error writing mutex profile: %v", err)
 				return
 			}
 			err = os.WriteFile(fileName, mutexProfBuf.Bytes(), os.ModePerm)
 			if err != nil {
-				log.Println("Error writing mutex profile to file: %v", err)
+				log.Printf("Error writing mutex profile to file: %v", err)
 				return
 			}
 		}
@@ -262,13 +262,13 @@ func blockProfiling(ctx context.Context) {
 			blockProfBuf := bytes.Buffer{}
 			err := pprof.Lookup("block").WriteTo(&blockProfBuf, 0)
 			if err != nil {
-				log.Println("Error writing block profile: %v", err)
+				log.Printf("Error writing block profile: %v", err)
 				return
 			}
 
 			err = os.WriteFile(fileName, blockProfBuf.Bytes(), os.ModePerm)
 			if err != nil {
-				log.Println("Error writing block profile to file: %v", err)
+				log.Printf("Error writing block profile to file: %v", err)
 				return
 			}
 		}
@@ -289,13 +289,13 @@ func goroutineProfiling(ctx context.Context) {
 			goroutineProfBuf := bytes.Buffer{}
 			err := pprof.Lookup("goroutine").WriteTo(&goroutineProfBuf, 1)
 			if err != nil {
-				log.Println("Error writing goroutine profile: %v", err)
+				log.Printf("Error writing goroutine profile: %v", err)
 				return
 			}
 
 			err = os.WriteFile(fileName, goroutineProfBuf.Bytes(), os.ModePerm)
 			if err != nil {
-				log.Println("Error writing goroutine profile to file: %v", err)
+				log.Printf("Error writing goroutine profile to file: %v", err)
 				return
 			}
 		}
--- a/tools/cache_perf_tool/main.go
+++ b/tools/cache_perf_tool/main.go
@@ -16,9 +16,11 @@ import (
 	"io"
 	"log"
 	"os"
+	"runtime"
 	"runtime/pprof"
 	"strconv"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
@@ -26,27 +28,34 @@ import (
 	"github.com/couchbase/sync_gateway/base"
 	"github.com/couchbase/sync_gateway/db"
 	"github.com/couchbaselabs/rosmar"
+	"github.com/felixge/fgprof"
 )
 
 var numGoroutines atomic.Int32
 
+var once sync.Once
+
 func main() {
 	mode := flag.String("mode", "processEntry", "Mode for the tool to run in, either dcp or processEntry.")
 	nodes := flag.Int("sgwNodes", 1, "Number of sgw nodes to abstract.")
-	bachSize := flag.Int("batchSize", 10, "Batch size for the sequence allocator.")
-	timeToRun := flag.Int("duration", 5, "Duration to run the test for in minutes.")
-	dalays := flag.String("writeDelay", "150", "Delay between writes in milliseconds. Must be entered in format <delayMS>,<delayMS>,<delayMS>.")
-	profile := flag.Bool("profile", false, "Enable profiling.")
+	batchSize := flag.Int("batchSize", 10, "Batch size for the sequence allocator.")
+	timeToRun := flag.Duration("duration", 5*time.Minute, "Duration to run the test for in minutes. Examples:  3m for 3 minutes, 30s for 30 seconds etc")
+	delays := flag.String("writeDelay", "0", "Delay between writes in milliseconds. Must be entered in format <delayMS>,<delayMS>,<delayMS>.")
+	profileInterval := flag.Duration("profileInterval", 30*time.Second, "Interval for profiling to be triggered on, example 10s would be every 10 seconds.") //flag.Bool("profileInterval", false, "Enable profiling.")
+	numChannels := flag.Int("numChannels", 1, "Number of channels to create per document.")
 	flag.Parse()
 
 	if *nodes < 1 {
 		log.Fatalf("Invalid number of nodes: %d", *nodes)
 	}
-	if *bachSize < 1 {
-		log.Fatalf("Invalid batch size: %d", *bachSize)
+	if *batchSize < 1 || *batchSize > 10 {
+		log.Fatalf("Invalid batch size: %d", *batchSize)
 	}
 	if *timeToRun < 1 {
 		log.Fatalf("Invalid duration: %d", *timeToRun)
+	}
+	if *numChannels < 1 {
+		log.Fatalf("Invalid number of channels: %d", *numChannels)
 	}
 
 	parentCtx := context.Background()
@@ -54,7 +63,7 @@ func main() {
 
 	// Need a bucket type for creating the database context
 	var walrusBucket *rosmar.Bucket
-	bucketName := "cahceTest" + "rosmar_"
+	bucketName := "cacheTest" + "rosmar_"
 	url := rosmar.InMemoryURL
 	walrusBucket, err := rosmar.OpenBucket(url, bucketName, rosmar.CreateOrOpen)
 	if err != nil {
@@ -62,7 +71,7 @@ func main() {
 	}
 	defer walrusBucket.Close(parentCtx)
 
-	if *profile {
+	if profileInterval.Seconds() > 1 {
 		// start CPU profiling here
 		cpuProfBuf := bytes.Buffer{}
 		err = pprof.StartCPUProfile(&cpuProfBuf)
@@ -83,10 +92,25 @@ func main() {
 				return
 			}
 		}()
-		go heapProfiling(ctx)
-		go mutexProfiling(ctx)
-		go blockProfiling(ctx)
-		go goroutineProfiling(ctx)
+		fileName := fmt.Sprintf("fprof-%s.prof", time.Now().Format(time.RFC3339))
+		fProfBuf := bytes.Buffer{}
+		stopFn := fgprof.Start(&fProfBuf, fgprof.FormatPprof)
+		defer func() {
+			err := stopFn()
+			if err != nil {
+				log.Printf("Error stopping fprof profile: %v", err)
+				return
+			}
+			err = os.WriteFile(fileName, fProfBuf.Bytes(), os.ModePerm)
+			if err != nil {
+				log.Printf("Error writing fprof profile to file: %v", err)
+				return
+			}
+		}()
+		go heapProfiling(ctx, *profileInterval)
+		go mutexProfiling(ctx, *profileInterval)
+		go blockProfiling(ctx, *profileInterval)
+		go goroutineProfiling(ctx, *profileInterval)
 	}
 
 	// new syncSeqMock to be used for the sequence allocator
@@ -109,6 +133,9 @@ func main() {
 	}
 	defer dbContext.Close(ctx)
 
+	// stats goroutine
+	go csvStats(ctx, dbContext)
+
 	// init change cache and unlock mutex for the test
 	dbContext.StartChangeCache(t, parentCtx)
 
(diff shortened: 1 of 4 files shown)


tools/cache_perf_tool/main.go
```diff
@@ -0,0 +1,303 @@
...
+	if *timeToRun < 1 {
+		log.Fatalf("Invalid duration: %d", *timeToRun)
+	}
+
+	parentCtx := context.Background()
+	ctx, cancelFunc := context.WithCancel(parentCtx)
+
+	// Need a bucket type for creating the database context
+	var walrusBucket *rosmar.Bucket
+	bucketName := "cahceTest" + "rosmar_"
```
bbrks:
```suggestion
	bucketName := "cacheTest" + "rosmar_"
```

tools/cache_perf_tool/main.go
```diff
@@ -0,0 +1,303 @@
...
+	// init change cache and unlock mutex for the test
+	dbContext.StartChangeCache(t, parentCtx)
+
+	// mode selection logic
+	if *mode == "dcp" {
+		// todo: add dcp mode code
+	} else if *mode == "processEntry" {
+		var delayList []time.Duration
+		if *nodes > 1 { // if we have one node then we don't have a delay (node will write as fast as possible)
+			delayList, err = extractDelays(*dalays)
```
bbrks:
```suggestion
			delayList, err = extractDelays(*delays)
```

tools/cache_perf_tool/main.go
```diff
@@ -0,0 +1,303 @@
...
+)
+
+var numGoroutines atomic.Int32
+
+func main() {
+	mode := flag.String("mode", "processEntry", "Mode for the tool to run in, either dcp or processEntry.")
+	nodes := flag.Int("sgwNodes", 1, "Number of sgw nodes to abstract.")
+	bachSize := flag.Int("batchSize", 10, "Batch size for the sequence allocator.")
+	timeToRun := flag.Int("duration", 5, "Duration to run the test for in minutes.")
+	dalays := flag.String("writeDelay", "150", "Delay between writes in milliseconds. Must be entered in format <delayMS>,<delayMS>,<delayMS>.")
```
bbrks:
```suggestion
	delays := flag.String("writeDelay", "150", "Delay between writes in milliseconds. Must be entered in format <delayMS>,<delayMS>,<delayMS>.")
```

tools/cache_perf_tool/main.go
```diff
@@ -0,0 +1,303 @@
...
+	"github.com/couchbaselabs/rosmar"
+)
+
+var numGoroutines atomic.Int32
+
+func main() {
+	mode := flag.String("mode", "processEntry", "Mode for the tool to run in, either dcp or processEntry.")
+	nodes := flag.Int("sgwNodes", 1, "Number of sgw nodes to abstract.")
+	bachSize := flag.Int("batchSize", 10, "Batch size for the sequence allocator.")
+	timeToRun := flag.Int("duration", 5, "Duration to run the test for in minutes.")
```
bbrks:
```suggestion
	timeToRun := flag.Int("duration", 5, "Duration to run the test for in minutes.")
```

Switch to `time.Duration` to allow for lower values like `1s` for devlopment

tools/cache_perf_tool/main.go
```diff
@@ -0,0 +1,303 @@
...
+	dalays := flag.String("writeDelay", "150", "Delay between writes in milliseconds. Must be entered in format <delayMS>,<delayMS>,<delayMS>.")
+	profile := flag.Bool("profile", false, "Enable profiling.")
```
bbrks:
```suggestion
	profile := flag.Bool("profile", false, "Enable profiling.")
```

Switch to profileInterval so we can grab profiles every N seconds

tools/cache_perf_tool/processEntry.go
```diff
@@ -0,0 +1,101 @@
...
+				sgwSeqno := node.seqAlloc.nextSeq()
+				docCount++
+				chanMap := make(channels.ChannelMap)
+				chanMap["test"] = nil
+				logEntry := &db.LogEntry{
+					Sequence:     sgwSeqno,
+					DocID:        fmt.Sprintf("key-%d-%d", docCount, sgwSeqno),
+					RevID:        "1-abc",
+					Flags:        0,
+					TimeReceived: time.Now(),
```
bbrks: grab `time.Now()` once and share for both fields - avoids 2 syscalls

tools/cache_perf_tool/processEntry.go
```diff
@@ -0,0 +1,101 @@
...
+			case <-ctx.Done():
+				return
+			default:
+				sgwSeqno := node.seqAlloc.nextSeq()
+				docCount++
+				chanMap := make(channels.ChannelMap)
+				chanMap["test"] = nil
+				logEntry := &db.LogEntry{
+					Sequence:     sgwSeqno,
+					DocID:        fmt.Sprintf("key-%d-%d", docCount, sgwSeqno),
```
bbrks: Sprintf is kinda slow, would suggest building this doc ID string using strconv and string concat

https://pkg.go.dev/strconv#FormatInt

`"key-"+"-"+ `

tools/cache_perf_tool/main.go
bbrks: allow `0` to support multiple nodes writing as fast as possible

tools/cache_perf_tool/main.go
```diff
@@ -0,0 +1,303 @@
...
+			}
+			err = os.WriteFile("cpu.prof", cpuProfileBuffer, os.ModePerm)
+			if err != nil {
+				log.Printf("error writing cpu profile to file: %v", err)
+				return
+			}
+		}()
+		go heapProfiling(ctx)
+		go mutexProfiling(ctx)
+		go blockProfiling(ctx)
```
bbrks: set fractions to grab data for these profiles

tools/cache_perf_tool/main.go
```diff
@@ -0,0 +1,303 @@
...
+		fmt.Println("end of test stats")
+		dbContext.UpdateCalculatedStats(ctx)
+		fmt.Println("high seq of cache", dbContext.DbStats.Database().HighSeqFeed.Value())
+		fmt.Println("pending seq length", dbContext.DbStats.Cache().PendingSeqLen.Value())
+		fmt.Println("high seq stable", dbContext.DbStats.Cache().HighSeqStable.Value())
+		fmt.Println("num current skipped", dbContext.DbStats.Cache().NumCurrentSeqsSkipped.Value())
+		fmt.Println("cumulative skipped", dbContext.DbStats.Cache().NumSkippedSeqs.Value())
+		fmt.Println("skipped length", dbContext.DbStats.Cache().SkippedSeqLen.Value())
+		fmt.Println("skipped capacity", dbContext.DbStats.Cache().SkippedSeqCap.Value())
+		fmt.Println("dcp cache count", dbContext.DbStats.Database().DCPCachingCount.Value())
```
bbrks: java properties format output for jenkins graph plots

```
STAT1=val
STAT2=val
STAT3=val
```

tools/cache_perf_tool/main.go
```diff
@@ -0,0 +1,303 @@
...
+			}
+		}
+		p := &processEntryGen{t: t, dbCtx: dbContext, delays: delayList, seqAlloc: seqAllocator, numNodes: *nodes, batchSize: *bachSize}
+		// create new sgw node abstraction and spawn write goroutines
+		p.spawnDocCreationGoroutine(ctx)
+	} else {
+		log.Printf("Invalid mode: %s", *mode)
+		return
+	}
+
```
bbrks: new ticker goroutine to dump stats every second (CSV)

tools/cache_perf_tool/main.go
```diff
@@ -117,42 +144,29 @@ func main() {
...
+		if err != nil {
+			return
 		}
-		p := &processEntryGen{t: t, dbCtx: dbContext, delays: delayList, seqAlloc: seqAllocator, numNodes: *nodes, batchSize: *bachSize}
+		// need to have a delay for each node defined so we have variable write throughput
+		if len(delayList) != *nodes {
+			log.Printf("invalid number of delays, number of input delays should match number of nodes: "+
+				"Delays=%d and number of nodes=%d", len(delayList), *nodes)
+			return
+		}
```
bbrks: Try and do this validation after you've parsed the flags and doing other validation. It will mean parsing `delayList` ahead of time and storing that in a variable for later use, but that's OK.

It might mean you're doing this evne if not running in `processEntry` mode? I'm not sure if these options would apply to other modes.

This avoids this late exit after you've started up the db context and profiling - one of which actually panics if it stops too quickly (from the error)

```
$ go build && ./cache_perf_tool -duration 1s -sgwNodes 4 -writeDelay 80,20,100
2025-04-11T11:57:53.047+01:00 [INF] c:CleanAgedItems db:db Created background task: "CleanAgedItems" with interval 1m0s
2025-04-11T11:57:53.047+01:00 [INF] db:db Using default sync function "function(doc){channel(doc.channels);}" for database db._default._default
2025-04-11T11:57:53.047+01:00 [INF] c:InsertPendingEntries Created background task: "InsertPendingEntries" with interval 2.5s
2025-04-11T11:57:53.047+01:00 [INF] c:CleanSkippedSequenceQueue Created background task: "CleanSkippedSequenceQueue" with interval 30m0s
2025/04/11 11:57:53 invalid number of delays, number of input delays should match number of nodes: Delays=3 and number of nodes=4
panic: runtime error: integer divide by zero

goroutine 1 [running]:
github.com/felixge/fgprof.(*wallclockProfile).exportPprof(0x140001a8080, 0x0, {0x10dd2dd0f0481?, 0x14002f0c6f0?, 0x1042bd100?}, {0x67f8f5b1?, 0x140004cd7f8?, 0x1042bd100?})
	/Users/benbrooks/go/pkg/mod/github.com/felixge/fgprof@v0.9.5/fgprof.go:230 +0xaa0
github.com/felixge/fgprof.(*wallclockProfile).Export(0xc1f65b4c42f0c6f0?, {0x10385cd38, 0x14000454000}, {0x1031d8b85?, 0xc831d7?}, 0x1042bd100?, {0x140004cd868?, 0x1030140e0?, 0x1042bd100?}, {0x140004cd7f0?, ...})
	/Users/benbrooks/go/pkg/mod/github.com/felixge/fgprof@v0.9.5/fgprof.go:187 +0x7c
github.com/felixge/fgprof.Start.func2()
	/Users/benbrooks/go/pkg/mod/github.com/felixge/fgprof@v0.9.5/fgprof.go:73 +0x174
main.main.func2()
	/Users/benbrooks/dev/cb/sg/tools/cache_perf_tool/main.go:99 +0x44
main.main()
	/Users/benbrooks/dev/cb/sg/tools/cache_perf_tool/main.go:155 +0xd58
```

tools/cache_perf_tool/main.go
```diff
@@ -167,13 +181,59 @@ outerloop:
...
+	// calculate here avg time to cache seq in ms
+	count := dbStats.Database().DCPCachingCount.Value()
+	timeNano := dbStats.Database().DCPCachingTime.Value()
+	avgTimeNano := float64(timeNano) / float64(count)
+	avgTimeMs := avgTimeNano / 1e6
+	_, _ = fmt.Fprintf(os.Stdout, "high_seq_feed=%d\npending_seq_len=%d\nhigh_seq_stable=%d\ncurrent_skipped_seq_count=%d\nnum_skipped_seqs=%d\n"+
+		"skipped_seq_len=%d\nskipped_seq_cap=%d\ndcp_caching_count=%d\ndcp_caching_time=%d\navg_time_per_seq_ms=%f\n",
+		dbStats.Database().HighSeqFeed.Value(), dbStats.Cache().PendingSeqLen.Value(), dbStats.Cache().HighSeqStable.Value(), dbStats.Cache().NumCurrentSeqsSkipped.Value(),
+		dbStats.Cache().NumSkippedSeqs.Value(), dbStats.Cache().SkippedSeqLen.Value(), dbStats.Cache().SkippedSeqCap.Value(), count,
+		timeNano, avgTimeMs)
```
bbrks: This would be far easier to read and edit if it were one line per stat.

```suggestion
	_, _ = fmt.Fprintf(os.Stdout, "high_seq_feed=%d\n", dbStats.Database().HighSeqFeed.Value())
	_, _ = fmt.Fprintf(os.Stdout, "pending_seq_len=%d\n", dbStats.Cache().PendingSeqLen.Value())
	_, _ = fmt.Fprintf(os.Stdout, "high_seq_stable=%d\n", dbStats.Cache().HighSeqStable.Value())
	_, _ = fmt.Fprintf(os.Stdout, "current_skipped_seq_count=%d\n", dbStats.Cache().NumCurrentSeqsSkipped.Value())
	_, _ = fmt.Fprintf(os.Stdout, "num_skipped_seqs=%d\n", dbStats.Cache().NumSkippedSeqs.Value())
	_, _ = fmt.Fprintf(os.Stdout, "skipped_seq_len=%d\n", dbStats.Cache().SkippedSeqLen.Value())
	_, _ = fmt.Fprintf(os.Stdout, "skipped_seq_cap=%d\n", dbStats.Cache().SkippedSeqCap.Value())
	_, _ = fmt.Fprintf(os.Stdout, "dcp_caching_count=%d\n", count)
	_, _ = fmt.Fprintf(os.Stdout, "dcp_caching_time=%d\n", timeNano)
	_, _ = fmt.Fprintf(os.Stdout, "avg_time_per_seq_ms=%f\n", avgTimeMs)
```

tools/cache_perf_tool/main.go
```diff
@@ -167,13 +181,59 @@ outerloop:
...
+			})
+			dbContext.UpdateCalculatedStats(ctx)
+			dbStats := dbContext.DbStats
+			// calculate here avg time to cache seq in ms
+			count := dbStats.Database().DCPCachingCount.Value()
+			timeNano := dbStats.Database().DCPCachingTime.Value()
+			avgTimeNano := float64(timeNano) / float64(count)
+			avgTimeMs := avgTimeNano / 1e6
+			_, _ = fmt.Fprintf(os.Stderr, "%s,%d,%d,%d,%d,%d,%d,%d,%d,%d,%f\n",
+				time.Now().Format(time.RFC3339), dbStats.Database().HighSeqFeed.Value(), dbStats.Cache().PendingSeqLen.Value(), dbStats.Cache().HighSeqStable.Value(),
```
bbrks: It's slightly easier to plot these if the time is a numerical timestamp, since you don't need to fiddle around with date formatting in a spreadsheet to get it parsing correctly.

`time.Now().Unix()`

tools/cache_perf_tool/main.go
```diff
@@ -167,13 +181,59 @@ outerloop:
...
+	defer numGoroutines.Add(-1)
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			once.Do(func() {
+				_, _ = fmt.Fprintf(os.Stderr, "timestamp,high_seq_feed,pending_seq_len,high_seq_stable,current_skipped_seq_count,"+
+					"num_skipped_seqs,skipped_seq_len,skipped_seq_cap,dcp_caching_count,dcp_caching_time,avg_time_per_seq_ms\n")
```
bbrks:
```suggestion
_, _ = fmt.Fprintf(os.Stderr, "timestamp,")
_, _ = fmt.Fprintf(os.Stderr, "high_seq_feed,")
_, _ = fmt.Fprintf(os.Stderr, "pending_seq_len,")
_, _ = fmt.Fprintf(os.Stderr, "high_seq_stable,")
_, _ = fmt.Fprintf(os.Stderr, "current_skipped_seq_count,")
_, _ = fmt.Fprintf(os.Stderr, "num_skipped_seqs,")
_, _ = fmt.Fprintf(os.Stderr, "skipped_seq_len,")
_, _ = fmt.Fprintf(os.Stderr, "skipped_seq_cap,")
_, _ = fmt.Fprintf(os.Stderr, "dcp_caching_count,")
_, _ = fmt.Fprintf(os.Stderr, "dcp_caching_time,")
_, _ = fmt.Fprintf(os.Stderr, "avg_time_per_seq_ms")
_, _ = fmt.Fprintf(os.Stderr, "\n")
```

tools/cache_perf_tool/main.go
```diff
@@ -167,13 +181,59 @@ outerloop:
...
+			dbStats := dbContext.DbStats
+			// calculate here avg time to cache seq in ms
+			count := dbStats.Database().DCPCachingCount.Value()
+			timeNano := dbStats.Database().DCPCachingTime.Value()
+			avgTimeNano := float64(timeNano) / float64(count)
+			avgTimeMs := avgTimeNano / 1e6
+			_, _ = fmt.Fprintf(os.Stderr, "%s,%d,%d,%d,%d,%d,%d,%d,%d,%d,%f\n",
+				time.Now().Format(time.RFC3339), dbStats.Database().HighSeqFeed.Value(), dbStats.Cache().PendingSeqLen.Value(), dbStats.Cache().HighSeqStable.Value(),
+				dbStats.Cache().NumCurrentSeqsSkipped.Value(), dbStats.Cache().NumSkippedSeqs.Value(), dbStats.Cache().SkippedSeqLen.Value(), dbStats.Cache().SkippedSeqCap.Value(),
+				count, timeNano, avgTimeMs)
```
bbrks:
```suggestion
_, _ = fmt.Fprintf(os.Stderr, "%d,", time.Now().Unix())
_, _ = fmt.Fprintf(os.Stderr, "%d,", dbStats.Database().HighSeqFeed.Value())
_, _ = fmt.Fprintf(os.Stderr, "%d,", dbStats.Cache().PendingSeqLen.Value())
_, _ = fmt.Fprintf(os.Stderr, "%d,", dbStats.Cache().HighSeqStable.Value())
_, _ = fmt.Fprintf(os.Stderr, "%d,", dbStats.Cache().NumCurrentSeqsSkipped.Value())
_, _ = fmt.Fprintf(os.Stderr, "%d,", dbStats.Cache().NumSkippedSeqs.Value())
_, _ = fmt.Fprintf(os.Stderr, "%d,", dbStats.Cache().SkippedSeqLen.Value())
_, _ = fmt.Fprintf(os.Stderr, "%d,", dbStats.Cache().SkippedSeqCap.Value())
_, _ = fmt.Fprintf(os.Stderr, "%d,", count)
_, _ = fmt.Fprintf(os.Stderr, "%d,", timeNano)
_, _ = fmt.Fprintf(os.Stderr, "%f", avgTimeMs)
_, _ = fmt.Fprintf(os.Stderr, "\n")
```

tools/cache_perf_tool/main.go
```diff
@@ -234,11 +294,14 @@ func mutexProfiling(ctx context.Context) {
 		case <-ticker.C:
 			fileName := fmt.Sprintf("mutex-%s.prof", time.Now().Format(time.RFC3339))
 			mutexProfBuf := bytes.Buffer{}
+			runtime.SetMutexProfileFraction(1)
+			time.Sleep(5 * time.Second)
```
bbrks: We can sleep for the interval here, to ensure the profile captures the full span - otherwise you can enable 1s profiling with a 5 second sleep which would not work correctly.

```suggestion
			time.Sleep(interval)
```

tools/cache_perf_tool/main.go
```diff
@@ -260,11 +323,14 @@ func blockProfiling(ctx context.Context) {
 		case <-ticker.C:
 			fileName := fmt.Sprintf("block-%s.prof", time.Now().Format(time.RFC3339))
 			blockProfBuf := bytes.Buffer{}
+			runtime.SetBlockProfileRate(1)
+			time.Sleep(5 * time.Second)
```
bbrks: ditto

tools/cache_perf_tool/sgwNodeAbstraction.go
```diff
@@ -54,6 +56,13 @@ func (s *sequenceAllocator) nextSeq() uint64 {
 		max := s.syncSeqMock.nextBatch(uint64(s.batchSize))
 		s.maxSeqInBatch = max
 		s.lastSeq = max - uint64(s.batchSize)
+		// channel has cap of 1 so to avoid blocking here we need select here
+		// the motivation to have cap 1 on this buffered channel is to sort of siluate dedupe from KV engine upon
```
bbrks:
```suggestion
		// the motivation to have cap 1 on this buffered channel is to sort of simulate dedupe from KV engine upon
```

tools/cache_perf_tool/sgwNodeAbstraction.go
```diff
@@ -54,6 +56,13 @@ func (s *sequenceAllocator) nextSeq() uint64 {
...
+		// rapid updates to _sync:seq doc
+		select {
+		case s.syncSeqEvent <- struct{}{}:
```
bbrks: I'm not quite sure I understand what's happening here. There's nothing to drain `s.syncSeqEvent`, and the `default:` case is going to let things through even if the channel is full?

That seems like the same thing as this whole select not being here unless I did misunderstand something.

tools/cache_perf_tool/main.go
```diff
@@ -16,53 +16,62 @@ import (
...
 	mode := flag.String("mode", "processEntry", "Mode for the tool to run in, either dcp or processEntry.")
 	nodes := flag.Int("sgwNodes", 1, "Number of sgw nodes to abstract.")
-	bachSize := flag.Int("batchSize", 10, "Batch size for the sequence allocator.")
-	timeToRun := flag.Int("duration", 5, "Duration to run the test for in minutes.")
-	dalays := flag.String("writeDelay", "150", "Delay between writes in milliseconds. Must be entered in format <delayMS>,<delayMS>,<delayMS>.")
-	profile := flag.Bool("profile", false, "Enable profiling.")
+	batchSize := flag.Int("batchSize", 10, "Batch size for the sequence allocator.")
+	timeToRun := flag.Duration("duration", 5*time.Minute, "Duration to run the test for in minutes. Examples:  3m for 3 minutes, 30s for 30 seconds etc")
+	delays := flag.String("writeDelay", "0", "Delay between writes in milliseconds. Must be entered in format <delayMS>,<delayMS>,<delayMS>.")
+	profileInterval := flag.Duration("profileInterval", 30*time.Second, "Interval for profiling to be triggered on, example 10s would be every 10 seconds.") //flag.Bool("profileInterval", false, "Enable profiling.")
```
bbrks:
```suggestion
	profileInterval := flag.Duration("profileInterval", 30*time.Second, "Interval for profiling to be triggered on, example 10s would be every 10 seconds.")
```

tools/cache_perf_tool/main.go
```diff
@@ -16,53 +16,62 @@ import (
...
 
 	"github.com/couchbase/sync_gateway/base"
 	"github.com/couchbase/sync_gateway/db"
 	"github.com/couchbaselabs/rosmar"
+	"github.com/felixge/fgprof"
 )
 
 var numGoroutines atomic.Int32
 
+var once sync.Once
```
bbrks: Doesn't feel appropriate to be a global here with a generic name. Either make it specific to CSV header or (which I think is also possible, rework that bit of code so it doesn't need to rely on a `sync.Once`.

It seems like we can just print the CSV header before we enter the ticker loop?

bbrks (changes requested):

tools/cache_perf_tool/main.go
```diff
@@ -0,0 +1,369 @@
...
+			dbStats := dbContext.DbStats
+			// calculate here avg time to cache seq in ms
+			count := dbStats.Database().DCPCachingCount.Value()
+			timeNano := dbStats.Database().DCPCachingTime.Value()
+			avgTimeNano := float64(timeNano) / float64(count)
+			avgTimeMs := avgTimeNano / 1e6
+			_, _ = fmt.Fprintf(os.Stderr, "%s,%d,%d,%d,%d,%d,%d,%d,%d,%d,%f\n",
+				time.Now().Format(time.RFC3339), dbStats.Database().HighSeqFeed.Value(), dbStats.Cache().PendingSeqLen.Value(), dbStats.Cache().HighSeqStable.Value(),
+				dbStats.Cache().NumCurrentSeqsSkipped.Value(), dbStats.Cache().NumSkippedSeqs.Value(), dbStats.Cache().SkippedSeqLen.Value(), dbStats.Cache().SkippedSeqCap.Value(),
+				count, timeNano, avgTimeMs)
```
bbrks: Make sure we're using consistent time units, otherwise these become impossible to plot on the same graph scales.

I don't mind either way, milliseconds or nanoseconds but as long as they are the same units

bbrks (changes requested):

tools/cache_perf_tool/sgwNodeAbstraction.go
gregns1: Apologies this is soemeting that leaked out of the DCP side implementation, will remove form this PR

tools/cache_perf_tool/main.go
gregns1: So I went with this approach as when we create the node we print logging out to show SGW nodes with their associated delay. I found that printing before loop would print the header in the middle of this logging and you'd end up with csv output and logging mixed together. I'll remove and remove the logging around nodes and associated delays as i don't feel there is much info the user gains from this.

bbrks (approved):

PR #7475, opened 2025-04-09
CBG-4603 Send 422 on delta sync for _removed in body

torcolvin:
CBG-4603: Send 422 on delta sync for _removed in body

- implement logic in blip test code to resend rev without delta

- [ ] TODO: write test for ISGR

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3026/


bbrks (approved):
LGTM

`TestUnprocessableDeltas` is a good ISGR test covering the same 422 handling that we could tweak for additional testing of `_removed` if you feel it's necessary, but I don't think it is.

The code paths are identical for ISGR, and we already have test coverage of 422 for other reasons so I am not concerned about needing more testing here.

PR #7476, opened 2025-04-11
CBG-4604 document missing run states

torcolvin:
CBG-4604 document missing run states

- update online/offline endpoints to discourage use vs `/{db}/_config` with `offline` in JSON.

This should probably be updated as well https://docs.couchbase.com/sync-gateway/current/resync.html#steps-to-resync and https://docs.couchbase.com/sync-gateway/current/database-offline.html

I think it is fine to POST to `/{db}/_config` even if you aren't using persistent config, but let me know if I misunderstand this and we need include the legacy way of doing online/offline.

I do not like the offline endpoint since https://github.com/couchbase/sync_gateway/blob/580e727f73807632715d5764c175e525597ad64f/db/database.go#L819 doesn't close the feeds correctly since the ctx isn't cancelled https://github.com/couchbase/sync_gateway/commit/6317f1977e7bab1390ba2fc2cf339913763066be

### DIFF
--- a/docs/api/paths/admin/db-_offline.yaml
+++ b/docs/api/paths/admin/db-_offline.yaml
@@ -10,9 +10,9 @@ parameters:
 post:
   summary: Take the database offline
   description: |-
-    If using persistent config, call [POST /{db}/_config](#operation/post_db-_config) with `{"offline": true}` to set the database to offline.
-
     This will take the database offline on this node only. Actions can be taken without disrupting current operations ungracefully or having the restart the Sync Gateway instance.
+
+    If using persistent config, call [POST /{db}/_config](#operation/post_db-_config) with `{"offline": true}` to set the database to offline.
 
     This will not take the backing Couchbase Server bucket offline.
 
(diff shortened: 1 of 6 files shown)


docs/api/paths/admin/db-_offline.yaml
```diff
@@ -10,7 +10,9 @@ parameters:
 post:
   summary: Take the database offline
   description: |-
-    This will take the database offline meaning actions can be taken without disrupting current operations ungracefully or having the restart the Sync Gateway instance.
+    If using persistent config, call [POST /{db}/_config](#operation/post_db-_config) with `{"offline": true}` to set the database to offline.
+
+    This will take the database offline on this node only. Actions can be taken without disrupting current operations ungracefully or having the restart the Sync Gateway instance.
```
bbrks: This reads as though the 2nd paragraph is the result of calling `POST /{db}/_config` since it kind of just runs on into it.

I think the `/db/_online` one is worded better since the order of paragraphs is reversed, and would advise the same here.

bbrks (changes requested):
Nit on phrasing

bbrks (approved):

PR #7480, opened 2025-04-14
fix end of stats test format

gregns1:
Small change to change end of test stats to csv format for plot job.

### DIFF
--- a/tools/cache_perf_tool/main.go
+++ b/tools/cache_perf_tool/main.go
@@ -162,7 +162,7 @@ func main() {
 		log.Printf("Invalid mode: %s", *mode)
 		return
 	}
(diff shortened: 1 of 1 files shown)


torcolvin (approved):
LGTM. Suggestion for future:

I'd probably use https://pkg.go.dev/encoding/csv

and create a structure that is 

```
struct fields type {
	headerName string
	value int64 
}
```

Or make values []int64 if you expect to have multiple lines. You could also do something like []csvVal and do a construction like:

```
func newCsvVal[T any](v T) string {
	return fmt.Printf("%", v)
}
```
All of this is probably overkill for this test tool and I don't think you need to do this but just spitballing to make it easier for future projects.

bbrks (approved):

PR #7481, opened 2025-04-14
CBG-4557 document endpoints for partitioned indexes

torcolvin:
This tracks the new endpoints for partitioned indexes.

This doesn't need to be reviewed until the implementation of the indexes is created, but can be used as a reference. I expect the bulk of the documentation is in the API reference but there will be a small amount of text about the new feature and how to do a downtime operation.

### DIFF
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -3088,13 +3088,17 @@ IndexInitStatus:
       description: The status of the current operation.
       type: string
       enum:
-        - running
         - completed
         - error
+        - running
+        - stopped
+        - stopping
       x-enumDescriptions:
         running: Indexes are being created.
         completed: All indexes were created.
         error: The index initialization operation has failed.
+        stopped: The index initialization operation has been stopped. These indexes may still on Couchbase Server.
+        stopping: The index initialization operation is in the process of being stopped.
     start_time:
       description: The ISO-8601 date and time the index initialization operation was started.
       type: string
@@ -3105,8 +3109,18 @@ IndexInitStatus:
       description: 'scope name with one or more collection names and the status of their index creation'
       type: object
       additionalProperties:
-        $ref: '#/CollectionNames'
-      example: {"scopeName": ["collection1", "collection2"]}
+        x-additionalPropertiesName: scopename
+        description: An object keyed by scope, containing a set of collections and the status of their index creation.
+        type: object
+        additionalProperties:
+          x-additionalPropertiesName: collectionname
+          type: string
+          enum:
+            - "completed"
+            - "in progress"
+          x-enumDescriptions:
+            completed: All indexes were created.
+            "in progress": Indexes are being created.
     settings:
       allOf:
         - $ref: '#/IndexSettings'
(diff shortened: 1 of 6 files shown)


docs/api/components/schemas.yaml
```diff
@@ -3075,3 +3067,53 @@ doc_access_spans:
...
+      example: {"scopeName": ["collection1", "collection2"]}
+    settings:
+      allOf:
+        - $ref: '#/IndexSettings'
+  required:
+    - status
+    - start_time
+    - last_error
+    - docs_changed
+    - docs_processed
```
```suggestion
    - index_status
```

docs/api/components/schemas.yaml
```diff
@@ -3075,3 +3067,53 @@ doc_access_spans:
...
+      type: string
+    last_error:
+      description: The last error that occurred in the index initialization operation (if any).
+      type: string
+    index_status:
+      description: 'scope name with one or more collection names and the status of their index creation'
+      type: object
+      additionalProperties:
+        $ref: '#/CollectionNames'
+      example: {"scopeName": ["collection1", "collection2"]}
```
bbrks: This doesn't quite seem right. I think it should be this?

```suggestion
      example: {"scopeName": {"collection1": "ready", "collection2": "in progress"}
```

docs/api/components/schemas.yaml
```diff
@@ -3075,3 +3067,53 @@ doc_access_spans:
...
+  description: The status of an asynchronous indexes initizational operation.
+  type: object
+  properties:
+    status:
+      description: The status of the current operation.
+      type: string
+      enum:
+        - running
+        - completed
+        - error
```
bbrks: > stop: Stops tracking the indexes by Sync Gateway. These indexes will still be created on Couchbase Server.

Suggests "stopping" and "stopped" still need to be valid enum states for this?

```suggestion
        - running
        - completed
        - stopping
        - stopped
        - error
```

docs/api/paths/admin/db-_index_init.yaml
```diff
@@ -0,0 +1,79 @@
...
+    1. Start the process of creating new indexes with [POST /{db}/_index_init](#operation/post-db_index_init).
+    2. Wait for index initialization to complete with [GET /{db}/_index_init](#operation/get_db-_index_init).
+    3. Update the database configuration to use these new indexes with [POST /{db}/_config](#operation/post_db-_config).
+    4. Call [POST /_post_upgrade](#operation/post__post_upgrade) to remove the original indexes.
+
+    This operation will start creation of indexes, and the creation of indexes can not be stopped on Couchbase Server once it has been started.
+
+    Required Sync Gateway RBAC roles:
+
+    * Sync Gateway Architect
```
bbrks, line 24: I have a feeling this needs to be Dev Ops, rather than Architect - but would like to verify this.

bbrks (changes requested):
Do you agree with these comments? I can change if required with the implementation PR

bbrks:
Assigning to self to uptake into feature PR

bbrks:
Rolled into #7496

PR #7485, opened 2025-04-15
Replace `[Type]Ptr` and `[Type]Default` usages with generic versions

bbrks:
Eliminates unnecessary type-specific code without performance penalty
- Replace old `[Type]Ptr` method usages with already existing generic version
- Add generic version of `[Type]Default`
- Some untyped constants got type hints which will give us compile-time guarantees about overflows - but were otherwise a requirement to use with `Ptr` because it had to determine the type at compile time.
- Won't impact future backports, since we're now building everywhere with modern Go versions that have generic support. `Ptr` is already backported.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3030/


base/bucket.go
```diff
@@ -34,7 +34,7 @@ const (
 	DefaultPool = "default"
 )
 
-const DefaultViewTimeoutSecs = 75 // 75s
+const DefaultViewTimeoutSecs uint32 = 75 // 75s
```
torcolvin, line 37: Nit: take or leave: make this


```suggestion
const DefaultViewTimeoutSecs = 75 * time.Second
```

Then use as `DefaultViewTimeoutSecs.Milliseconds()` in return value?

db/attachment_compaction_test.go
```diff
@@ -921,18 +921,18 @@ func createDocWithInBodyAttachment(t *testing.T, ctx context.Context, docID stri
 
 		attachmentSyncDataBytes, err := base.JSONMarshal(attachmentSyncData)
 		if err != nil {
-			return nil, base.Uint32Ptr(0), false, err
+			return nil, base.Ptr(uint32(0)), false, err
```
torcolvin, line 924: this is just test code, change to more sensible API:

```suggestion
			return nil, nil, false, err
```

db/attachment_compaction_test.go
```diff
@@ -921,18 +921,18 @@ func createDocWithInBodyAttachment(t *testing.T, ctx context.Context, docID stri
...
 		}
 
 		updated, err = base.InjectJSONPropertiesFromBytes(current, base.KVPairBytes{
 			Key: "_attachments",
 			Val: attachmentSyncDataBytes,
 		})
 		if err != nil {
-			return nil, base.Uint32Ptr(0), false, err
+			return nil, base.Ptr(uint32(0)), false, err
```
torcolvin, line 932: this is just test code, change to more sensible API:

```suggestion
			return nil, nil, false, err
```

torcolvin (approved):

db/attachment_compaction_test.go
bbrks: `0` and `nil` are semantically different for an Update callback (set to no expiry vs. leave as-is)

base/bucket.go
bbrks: Still have to do the awkward uint32 conversion after converting from a duration to milliseconds though in real usage.

Given this is still old view code I'm not going to change it.

PR #7489, opened 2025-04-18
CBG-4616 make WaitForChanges more strict

torcolvin:
CBG-4616 make WaitForChanges more strict

This addresses some subtle race conditions in test where we could be waiting for the wrong number of changes, since `WaitForChanges` would return if there were greater than the expected number of changes.

https://github.com/couchbase/sync_gateway/blob/main/rest/attachment_test.go#L1776 is an example of a problematic test. I changed this test to just do an on demand get, but this number is wrong since there are changes from the previous subtest. I thought about making it do a wait for import check, but an on-demand import seemed more explicit here.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3039/

### DIFF
--- a/rest/blip_api_no_race_test.go
+++ b/rest/blip_api_no_race_test.go
@@ -17,8 +17,11 @@ import (
 	"sync"
 	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/couchbase/sync_gateway/base"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 // TestBlipPusherUpdateDatabase starts a push replication and updates the database underneath the replication.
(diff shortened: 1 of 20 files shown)


rest/blip_api_no_race_test.go
```diff
@@ -76,8 +74,7 @@ func TestBlipPusherUpdateDatabase(t *testing.T) {
 		}()
 
 		// and wait for a few to be done before we proceed with updating database config underneath replication
-		_, err := rt.WaitForChanges(5, "/{{.keyspace}}/_changes", "", true)
-		require.NoError(t, err)
+		rt.WaitForChanges(5, "/{{.keyspace}}/_changes", "", true)
```
bbrks: I am not sure how this test is going to be reliable with the new exact length checks.

Does the test need changing, or a new helper that is tolerant to more changes than expected?

bbrks (changes requested):
In general I'm OK with the approach, but there's at least one test in particular (see inline comment) that is relying on the `>=` behaviour of the existing implementation.

rest/blip_api_no_race_test.go
torcolvin: Good catch, this _doesn't_ work correctly but I didn't run this test enough times to expose this problem.

torcolvin: I changed this test but it is skipped which is why it didn't flag as a failure. The test doesn't work correctly even when it is unskipped, but that is covered in CBG-1908. I don't think this test is relevant anymore as written, and it doesn't work correctly after the BlipTester changes.

bbrks (approved):

PR #7493, opened 2025-04-22
CBG-4006: Default log retention of 90 days for stats

bbrks:
CBG-4006

Override the default value of `MaxAge` for the stats logger to 90 days.
- Doing this instead of simply increasing `statsMinAge` because existing users may have already configured a value >= 3 but < 45 and we don't want that to now fail. Min age 3 still seems like a reasonable limit to configuration without affecting the default value of `MaxAge`.
- Size limit of 1GB (per level) is still in effect, so the new default of 90 days is safe enough.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3043/

### DIFF
--- a/base/logger_audit.go
+++ b/base/logger_audit.go
@@ -197,7 +197,7 @@ func NewAuditLogger(ctx context.Context, config *AuditLoggerConfig, logFilePath 
 		config.CollationBufferSize = Ptr(defaultFileLoggerCollateBufferSize)
 	}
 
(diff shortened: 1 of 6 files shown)


torcolvin (approved):
This seems fine.

At this point, I'd probably be inclined to make an options struct due to close together int, *int, * but I don't feel strongly about this.

PR #7494, opened 2025-04-22
CBG-4456 fix unit test

torcolvin:
mobile jenkins caught this because it runs with `-tags cb_sg_enterprise` but not `-race`. We do not have anything in the test matrix for this.

I would like to move this test out of the no race package but there's still a data race in `AssertLogContains` that I can not quite pin down yet.

### DIFF
--- a/rest/audit_test.go
+++ b/rest/audit_test.go
@@ -964,7 +964,7 @@ func TestNoAuditWhenDisabledAtDb(t *testing.T) {
 
 	// test event that is enabled by default on db (when audit is enabled)
 	output = base.AuditLogContents(t, func(t testing.TB) {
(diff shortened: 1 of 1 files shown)


bbrks (approved):

bbrks:
> mobile jenkins caught this because it runs with -tags cb_sg_enterprise but not -race. We do not have anything in the test matrix for this.

Integration tests cover this combination

torcolvin:
> > mobile jenkins caught this because it runs with -tags cb_sg_enterprise but not -race. We do not have anything in the test matrix for this.
> 
> Integration tests cover this combination

I'm OK that just mobile jenkins covered this but I was confused because I didn't see any pre-merge failures.

PR #7495, opened 2025-04-22
Disable TestChangesLargeSequence

torcolvin:
Disabling the test, to make views integration tests pass.

This test fails in unique ways for GSI, views, and rosmar. In each case the sequence value of the document is incorrect, but a different number of incorrect. In GSI and views, the value of sequence via query will always return `math.MaxInt64`. In rosmar, it will return a number that is much greater than `math.MaxInt64` that is not equal to the sequence value.

A better change would be to restrict sync sequences to less than max.Int64.

Fix the test to actually behave as intended, even if the assertions don't pass. We can use `InitSyncSeq` in rosmar and the sequence numbers are higher because the principal doc `_user/alice` gets a sequence before the testdoc.

### DIFF
--- a/rest/utilities_testing.go
+++ b/rest/utilities_testing.go
@@ -221,7 +221,7 @@ func (rt *RestTester) Bucket() base.Bucket {
 	rt.TestBucket = testBucket
 
 	if rt.PersistentConfig != false {
-		require.Zero(rt.TB(), rt.InitSyncSeq, "RestTesterConfig.InitSyncSeq is not supported on rosmar")
+		require.Zero(rt.TB(), rt.InitSyncSeq, "RestTesterConfig.InitSyncSeq is not supported with RestTesterConfig.PersistentConfig = true")
 	}
 
 	corsConfig := &auth.CORSConfig{
--- a/rest/utilities_testing.go
+++ b/rest/utilities_testing.go
@@ -220,17 +220,8 @@ func (rt *RestTester) Bucket() base.Bucket {
 	}
 	rt.TestBucket = testBucket
 
-	if rt.InitSyncSeq > 0 {
-		if base.TestsUseNamedCollections() {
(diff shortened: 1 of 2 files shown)


rest/utilities_testing.go
```diff
@@ -220,17 +220,8 @@ func (rt *RestTester) Bucket() base.Bucket {
...
-		}
-
-		log.Printf("Initializing %s to %d", base.DefaultMetadataKeys.SyncSeqKey(), rt.InitSyncSeq)
-		tbDatastore := testBucket.GetMetadataStore()
-		_, incrErr := tbDatastore.Incr(base.DefaultMetadataKeys.SyncSeqKey(), rt.InitSyncSeq, rt.InitSyncSeq, 0)
-		if incrErr != nil {
-			rt.TB().Fatalf("Error initializing %s in test bucket: %v", base.DefaultMetadataKeys.SyncSeqKey(), incrErr)
-		}
+	if rt.PersistentConfig != false {
+		require.Zero(rt.TB(), rt.InitSyncSeq, "RestTesterConfig.InitSyncSeq is not supported on rosmar")
```
bbrks: I still don't think we should or need to do this.

`InitSyncSeq` isn't related to persistent config and initializing a non-zero seq _should_ work just fine on Rosmar? I do take your point that the above skipped test is the only one using `InitSyncSeq` but I don't think that should be reason to add these restrictions.

bbrks (changes requested):

rest/utilities_testing.go
torcolvin: This is actually just a typo in the description, I just wanted to protect the caller from setting `InitSyncSeq` with `PersistentConfig: true` where this would be a no-op. Fixed the text here.

bbrks (approved):

PR #7496, opened 2025-04-23
CBG-4587: Add async index init REST API to allow partitioned index creation without downtime

bbrks:
CBG-4587

Add async index init REST API to allow partitioned index creation without downtime
- `GET`/`POST` `/db/_index_init`
- Supports cross-node start/stop and status reporting with the `AsyncIndexInitManager` wrapper
- Uses the existing `DatabaseInitManager` to integrate with existing index init code and support index creation request deduplication
- Index creation states can be one of `queued`, `in progress`, `ready`, `error`
- No `/_post_upgrade` cleanup support yet - will be in a follow-up ticket/PR
- "E2E" type test that covers the expected and documented user steps to do a no-downtime migration to partitioned indexes

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3047/
- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3064/

### DIFF
--- a/db/background_mgr.go
+++ b/db/background_mgr.go
@@ -348,9 +348,6 @@ func (b *BackgroundManager) resetStatus() {
 }
 
 func (b *BackgroundManager) Stop() error {
-	if b == nil {
-		return nil
-	}
 	if err := b.markStop(); err != nil {
 		return err
 	}
--- a/db/background_mgr.go
+++ b/db/background_mgr.go
@@ -348,8 +348,10 @@ func (b *BackgroundManager) resetStatus() {
 }
 
 func (b *BackgroundManager) Stop() error {
-	err := b.markStop()
-	if err != nil {
+	if b == nil {
+		return nil
+	}
+	if err := b.markStop(); err != nil {
 		return err
 	}
 
--- a/db/background_mgr.go
+++ b/db/background_mgr.go
@@ -124,9 +124,12 @@ func (b *BackgroundManager) Start(ctx context.Context, options map[string]interf
 		return err
 	}
 
+	// disassociate with request context lifespan for background goroutines
+	backgroundCtx := context.WithoutCancel(ctx)
+
 	if b.isClusterAware() {
 		b.backgroundManagerStatusUpdateWaitGroup.Add(1)
-		go func(terminator *base.SafeTerminator) {
+		go func(ctx context.Context, terminator *base.SafeTerminator) {
 			defer b.backgroundManagerStatusUpdateWaitGroup.Done()
 			ticker := time.NewTicker(BackgroundManagerStatusUpdateIntervalSecs * time.Second)
 			for {
@@ -141,10 +144,10 @@ func (b *BackgroundManager) Start(ctx context.Context, options map[string]interf
 					return
 				}
 			}
-		}(b.terminator)
+		}(backgroundCtx, b.terminator)
 	}
 
-	go func() {
+	go func(ctx context.Context) {
 		err := b.Process.Run(ctx, options, b.UpdateStatusClusterAware, b.terminator)
 		if err != nil {
 			base.ErrorfCtx(ctx, "Error: %v", err)
@@ -173,7 +176,7 @@ func (b *BackgroundManager) Start(ctx context.Context, options map[string]interf
 			// Note: We can ignore the error, worst case is the user has to wait until the heartbeat doc expires
 			_ = b.clusterAwareOptions.metadataStore.Delete(b.clusterAwareOptions.HeartbeatDocID())
 		}
-	}()
+	}(backgroundCtx)
 
 	if b.isClusterAware() {
 		err := b.UpdateStatusClusterAware(ctx)
--- a/db/background_mgr.go
+++ b/db/background_mgr.go
@@ -124,12 +124,9 @@ func (b *BackgroundManager) Start(ctx context.Context, options map[string]interf
 		return err
 	}
 
-	// disassociate with request context lifespan for background goroutines
-	backgroundCtx := context.WithoutCancel(ctx)
-
 	if b.isClusterAware() {
 		b.backgroundManagerStatusUpdateWaitGroup.Add(1)
-		go func(ctx context.Context, terminator *base.SafeTerminator) {
+		go func(terminator *base.SafeTerminator) {
 			defer b.backgroundManagerStatusUpdateWaitGroup.Done()
 			ticker := time.NewTicker(BackgroundManagerStatusUpdateIntervalSecs * time.Second)
 			for {
@@ -144,10 +141,10 @@ func (b *BackgroundManager) Start(ctx context.Context, options map[string]interf
 					return
 				}
 			}
-		}(backgroundCtx, b.terminator)
+		}(b.terminator)
 	}
 
-	go func(ctx context.Context) {
+	go func() {
 		err := b.Process.Run(ctx, options, b.UpdateStatusClusterAware, b.terminator)
 		if err != nil {
 			base.ErrorfCtx(ctx, "Error: %v", err)
@@ -176,7 +173,7 @@ func (b *BackgroundManager) Start(ctx context.Context, options map[string]interf
 			// Note: We can ignore the error, worst case is the user has to wait until the heartbeat doc expires
 			_ = b.clusterAwareOptions.metadataStore.Delete(b.clusterAwareOptions.HeartbeatDocID())
 		}
-	}(backgroundCtx)
+	}()
 
 	if b.isClusterAware() {
 		err := b.UpdateStatusClusterAware(ctx)
--- a/db/background_mgr_async_index_init.go
+++ b/db/background_mgr_async_index_init.go
@@ -1,0 +1,84 @@
+//  Copyright 2025-Present Couchbase, Inc.
+//
+//  Use of this software is governed by the Business Source License included
+//  in the file licenses/BSL-Couchbase.txt.  As of the Change Date specified
+//  in that file, in accordance with the Business Source License, use of this
+//  software will be governed by the Apache License, Version 2.0, included in
+//  the file licenses/APL2.txt.
+
+package db
+
+import (
+	"context"
+	"sync"
+
+	"github.com/couchbase/sync_gateway/base"
+)
+
+// AsyncIndexInitManager is a background manager process that manages the cross-node job state and status of an async invocation of DatabaseInitManager (via /db/_index_init)
+// This manager does not do the actual work to initialize indexes, due to go package boundaries and import-cycles, but is being fed status updates from callbacks passed into rest.DatabaseInitManager
+// This status can be viewed cross-node, and similarly the start/stop actions can be used cross-node with this AsyncIndexInitManager layer.
+type AsyncIndexInitManager struct {
+	lock      sync.Mutex
+	statusMap *map[string]map[string]string
+	doneChan  chan error
+}
+
+// Init is called synchronously to set up a run for the background manager process. See Run() for the async part.
+func (a *AsyncIndexInitManager) Init(ctx context.Context, options map[string]interface{}, clusterStatus []byte) error {
+	a.statusMap = options["statusMap"].(*map[string]map[string]string)
+	a.doneChan = options["doneChan"].(chan error)
+	return nil
+}
+
+// Run is called inside a goroutine to perform the job of the job. This function should block until the job is complete.
+func (a *AsyncIndexInitManager) Run(ctx context.Context, options map[string]interface{}, persistClusterStatusCallback updateStatusCallbackFunc, terminator *base.SafeTerminator) error {
+	return <-a.doneChan
+}
+
+type AsyncIndexInitManagerResponse struct {
+	BackgroundManagerStatus
+	IndexStatus map[string]map[string]string `json:"index_status"` // scope->collection->status
+}
+
+type AsyncIndexInitManagerStatusDoc struct {
+	AsyncIndexInitManagerResponse `json:"status"`
+}
+
+func (a *AsyncIndexInitManager) GetProcessStatus(status BackgroundManagerStatus) (statusOut []byte, meta []byte, err error) {
+	a.lock.Lock()
+	defer a.lock.Unlock()
+
+	var statusMap map[string]map[string]string
+	if a.statusMap != nil {
+		statusMap = *a.statusMap
+	}
+
+	retStatus := AsyncIndexInitManagerResponse{
+		BackgroundManagerStatus: status,
+		IndexStatus:             statusMap,
+	}
+
+	statusJSON, err := base.JSONMarshal(retStatus)
+	return statusJSON, nil, err
+}
+
+func (a *AsyncIndexInitManager) ResetStatus() {
+	a.statusMap = nil
+	return
+}
+
+var _ BackgroundManagerProcessI = &AsyncIndexInitManager{}
+
+func NewAsyncIndexInitManager(metadataStore base.DataStore, metaKeys *base.MetadataKeys) *BackgroundManager {
+	return &BackgroundManager{
+		name:    "index_init",
+		Process: &AsyncIndexInitManager{},
+		clusterAwareOptions: &ClusterAwareBackgroundManagerOptions{
+			metadataStore: metadataStore,
+			metaKeys:      metaKeys,
+			processSuffix: "index_init",
+		},
+		terminator: base.NewSafeTerminator(),
+	}
+}
--- a/db/background_mgr_async_index_init.go
+++ b/db/background_mgr_async_index_init.go
@@ -19,22 +19,27 @@ import (
 // This manager does not do the actual work to initialize indexes, due to go package boundaries and import-cycles, but is being fed status updates from callbacks passed into rest.DatabaseInitManager
 // This status can be viewed cross-node, and similarly the start/stop actions can be used cross-node with this AsyncIndexInitManager layer.
 type AsyncIndexInitManager struct {
-	lock      sync.Mutex
-	statusMap *IndexStatusByCollection
-	doneChan  chan error
+	lock       sync.Mutex
+	_statusMap *IndexStatusByCollection // _statusMap is updated by the DatabaseInitManager's callbacks. Here for BackgroundManager persistence.
+	_doneChan  chan error               // _doneChan is a DatabaseInitManager worker's done channel. Here to allow Run to block until complete.
 }
 
 // Init is called synchronously to set up a run for the background manager process. See Run() for the async part.
 func (a *AsyncIndexInitManager) Init(ctx context.Context, options map[string]interface{}, clusterStatus []byte) error {
-	a.statusMap = options["statusMap"].(*IndexStatusByCollection)
-	a.doneChan = options["doneChan"].(chan error)
+	a.lock.Lock()
+	defer a.lock.Unlock()
+	a._statusMap = options["statusMap"].(*IndexStatusByCollection)
+	a._doneChan = options["doneChan"].(chan error)
 	return nil
 }
 
 // Run is called inside a goroutine to perform the job of the job. This function should block until the job is complete.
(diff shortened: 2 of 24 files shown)


docs/api/admin.yaml
```diff
@@ -107,8 +107,14 @@ paths:
     $ref: './paths/admin/keyspace-_config-sync.yaml'
   '/{keyspace}/_config/import_filter':
     $ref: './paths/admin/keyspace-_config-import_filter.yaml'
+  '/{db}/':
+    $ref: './paths/admin/db-.yaml'
+  /_all_dbs:
+    $ref: ./paths/admin/_all_dbs.yaml
```
bbrks: accidentally pulled in via bad cherry-pick/rebase

```suggestion
```

db/background_mgr_async_index_init.go
```diff
@@ -0,0 +1,84 @@
...
+// Init is called synchronously to set up a run for the background manager process. See Run() for the async part.
+func (a *AsyncIndexInitManager) Init(ctx context.Context, options map[string]interface{}, clusterStatus []byte) error {
+	a.statusMap = options["statusMap"].(*map[string]map[string]string)
+	a.doneChan = options["doneChan"].(chan error)
+	return nil
+}
+
+// Run is called inside a goroutine to perform the job of the job. This function should block until the job is complete.
+func (a *AsyncIndexInitManager) Run(ctx context.Context, options map[string]interface{}, persistClusterStatusCallback updateStatusCallbackFunc, terminator *base.SafeTerminator) error {
+	return <-a.doneChan
```
torcolvin: Should this get cancelled when the background manager is shut down? Do we need to pass a cancelation context or use a terminator like the other context managers when the database shuts down?

db/background_mgr_async_index_init.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+type AsyncIndexInitManagerResponse struct {
+	BackgroundManagerStatus
+	IndexStatus map[string]map[string]string `json:"index_status"` // scope->collection->status
```
torcolvin: Is this worth making a type in case we change it?

db/database.go
```diff
@@ -666,6 +667,14 @@ func (context *DatabaseContext) stopBackgroundManagers() []*BackgroundManager {
 		}
 	}
 
+	if context.AsyncIndexInitManager != nil {
```
torcolvin: I'd be tempted to just construct just to `for _, mgr := range []*BackgroundManager{context.ResyncManager, context.AttachmentCompactionManager...}` to dedupe this code a bit.

docs/api/components/schemas.yaml
```diff
@@ -3147,3 +3139,65 @@ DatabaseState:
...
+        - completed
+        - error
+        - running
+        - stopped
+        - stopping
+      x-enumDescriptions:
+        running: Indexes are being created.
+        completed: All indexes were created.
+        error: The index initialization operation has failed.
+        stopped: The index initialization operation has been stopped. These indexes may still on Couchbase Server.
```
torcolvin:
```suggestion
        stopped: The index initialization operation has been stopped. These indexes may still exist on Couchbase Server.
```

docs/api/paths/admin/db-_index_init.yaml
```diff
@@ -0,0 +1,79 @@
...
+      description: Defines whether the index creation operation is being started or stopped.
+      schema:
+        type: string
+        default: start
+        enum:
+          - start
+          - stop
+        x-enumDescriptions:
+          start: Starts the creation of indexes.
+          stop: Stops tracking the indexes by Sync Gateway. These indexes will still be created on Couchbase Server.
```
torcolvin:
```suggestion
          stop: Stops tracking the index creation by Sync Gateway. These indexes will still be created on Couchbase Server.
```

docs/api/paths/admin/db-_index_init.yaml
```diff
@@ -0,0 +1,79 @@
...
+  requestBody:
+    content:
+      application/json:
+        schema:
+          allOf:
+            - $ref: ../../components/schemas.yaml#/IndexSettings
+  responses:
+    '200':
+      description: successfully changed the status of the resync operation
```
torcolvin:
```suggestion
      description: successfully changed the status of the index initialization operation
```

rest/indextest/index_init_api_test.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	base.SetUpTestLogging(t, base.LevelDebug, base.KeyHTTP, base.KeyConfig, base.KeyQuery)
+	require.False(t, base.TestsDisableGSI(), "Test requires GSI to be enabled")
+	const (
+		dbName            = "db"
+		initialPartitions = uint32(2)
+		newPartitions     = uint32(4)
+	)
+
+	rt := rest.NewRestTesterPersistentConfigNoDB(t)
```
torcolvin: Does this work with and without persistent config? Some of our k8s customers do not use k8s.

rest/indextest/index_init_api_test.go
```diff
@@ -0,0 +1,123 @@
...
+//  2. Use Index Init API to initialize a new set of indexes with different partitions.
+//  3. Use the regular db config API to change the number of partitions to match the set created.
+//  4. Ensure the database starts up using the indexes created in step 2 without any delay.
+//  5. Use the _post_upgrade endpoint to clean up the old set of indexes.
+func TestChangeIndexPartitions(t *testing.T) {
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("This test only works against Couchbase Server")
+	}
+	base.SetUpTestLogging(t, base.LevelDebug, base.KeyHTTP, base.KeyConfig, base.KeyQuery)
+	require.False(t, base.TestsDisableGSI(), "Test requires GSI to be enabled")
```
torcolvin: This test should? fail with views enabled, or at least this assertion will not work.

rest/admin_api.go
```diff
@@ -266,6 +267,100 @@ func (h *handler) handleDbOffline() error {
...
+		return base.HTTPErrorf(http.StatusBadRequest, "num_partitions must be greater than 0")
+	}
+	return nil
+}
+
+// handleIndexInit allows async index initialization to be run or managed
+func (h *handler) handlePostIndexInit() error {
+	action := cmp.Or(h.getQuery("action"), "start")
+
+	if action == "stop" {
```
torcolvin, line 298: What do you think about a set of tests to test validation of input:

- action incorrectly
- calling this function when GSI is not enabled
- bad json
- mismatched partitions

rest/admin_api.go
```diff
@@ -266,6 +267,100 @@ func (h *handler) handleDbOffline() error {
...
+	}
+
+	var newDbConfig DatabaseConfig
+	if err := base.DeepCopyInefficient(&newDbConfig, currentDbConfig); err != nil {
+		return err
+	}
+	newDbConfig.Index.NumPartitions = req.NumPartitions
+
+	var statusMap = make(map[string]map[string]string, len(newDbConfig.Scopes))
+	for scope, _ := range newDbConfig.Scopes {
```
torcolvin: micro nit:

```suggestion
	for scope := range newDbConfig.Scopes {
```

rest/admin_api.go
```diff
@@ -266,6 +267,100 @@ func (h *handler) handleDbOffline() error {
...
+		statusMap[scope] = make(map[string]string)
+	}
+	var statusCallback CollectionCallbackFunc = func(dbName string, scName base.ScopeAndCollectionName, status CollectionIndexStatus) {
+		// even though the scope map is initialized above,
+		// it's possible we have a named scope and this triggers a `_default` index init,
+		// so we need to be defensive here and create if necessary
+		if _, ok := statusMap[scName.ScopeName()]; !ok {
+			statusMap[scName.ScopeName()] = make(map[string]string, 1)
+		}
```
torcolvin: Is there a chance this can cause a data race if two indexes in `_default` collection complete simultaneously and there are only explicit named collections? I think we could avoid this case by _always_ initializing `_default` scope above.

rest/database_init_manager.go
```diff
@@ -250,6 +278,9 @@ func (w *DatabaseInitWorker) Run() {
 		keyspaceCtx := base.KeyspaceLogCtx(w.ctx, w.n1qlStore.BucketName(), scName.ScopeName(), scName.CollectionName())
 		indexErr = db.InitializeIndexes(keyspaceCtx, w.n1qlStore, collectionIndexOptions)
 		if indexErr != nil {
+			if w.collectionStatusCallback != nil {
+				w.collectionStatusCallback(w.dbName, scName, CollectionIndexStatusError)
+			}
 			break
 		}
```
torcolvin, line 277: The code below here is weird, because it actually should look like:

```
		// Check for context cancellation after each collection is processed - if cancelled, return cancellation error
		// to all watchers end exit
		if w.ctx.Err() != nil {
			indexErr = errors.New("Database initialization cancelled")
			break
        }
```

We check for `indexErr` in the case of a non ctx error above.

rest/database_init_manager.go
```diff
@@ -264,8 +295,8 @@ func (w *DatabaseInitWorker) Run() {
...
 
-		if w.collectionCompleteCallback != nil {
-			w.collectionCompleteCallback(w.dbName, scName.CollectionName())
+		if w.collectionStatusCallback != nil {
+			w.collectionStatusCallback(w.dbName, scName, CollectionIndexStatusReady)
```
torcolvin: I think maybe this should be moved above the context check?

torcolvin (commented):
I'm a bit worried about the state of `doneChan` in general. These are cases that concern me (and might warrant tests):

1. start index init, takes a while to build
2. stop index init
3. start new index init, with different parameters

1. shut down database while creation is running
2. assert no goroutine leaks

1. swap database to offline while index creation is running (with `{"offline": true}` and /db/_offline
2. make sure that there are no hanging goroutines

I'm OK if some of these don't work correctly, but I'd like to know what the limitations are.

rest/admin_api.go
bbrks: > if two indexes in _default collection complete simultaneously

I don't think this is possible with how the index init works today, it doesn't have that level of concurrency... But I think always initializing a _default scope above without any collections makes more sense and avoids this check in the callback.

rest/database_init_manager.go
bbrks: Thanks, this was from a previous iteration that had no per-index error state

rest/indextest/index_init_api_test.go
bbrks: Not sure I follow this comment. Are you suggesting the test case should run with views?

rest/admin_api.go
bbrks: agreed, will add

rest/indextest/index_init_api_test.go
bbrks: it should work but I'll add a different test to verify

rest/indextest/index_init_api_test.go
torcolvin: I'm not suggesting this test should work with views but I think as written this assertion would fail with couchbase server and views? This test could/should be skipped.

bbrks: I am not sure how `base.TestsDisableGSI()` can return false when using views and would allow this test to run? Are you sure you read the assertion and condition correctly?

torcolvin: Running `SG_TEST_BACKING_STORE=couchbase go test ./rest/indextest -run TestChangeIndexPartitions` will fail. I'm OK to skip this test, but this require statement will fail.

I think we could have this endpoint fail if views are enabled on the database but that doesn't need to be in this test.

bbrks: Oh my god I am so sorry... I was reading `require.False` as a skip somehow in my head ...

bbrks:
I have greatly increased the scope of testing on this now and done a lot of polishing. It's about where it should've been when opening PR :)

db/background_mgr.go
```diff
@@ -124,9 +124,12 @@ func (b *BackgroundManager) Start(ctx context.Context, options map[string]interf
 		return err
 	}
 
+	// disassociate with request context lifespan for background goroutines
+	backgroundCtx := context.WithoutCancel(ctx)
```
torcolvin: I don't know if this is correct, at least not for a generic background task. The context used by the db init manager task, so is the concern that you could still use a background manager even after you stop the database?

My concern about having it be non cancellable is you might actually expect to be able to use this for cancellation, and generally having any work on the background manager will be non functional. It is weird to have a terminator and a context, but I'm just not sure why this is the way it is. Normally you'd need a bucket for the lifetime of a background job.

db/background_mgr_async_index_init.go
```diff
@@ -0,0 +1,96 @@
...
+	retStatus := AsyncIndexInitManagerResponse{
+		BackgroundManagerStatus: status,
+		IndexStatus:             statusMap,
+	}
+
+	statusJSON, err := base.JSONMarshal(retStatus)
+	return statusJSON, nil, err
+}
+
+func (a *AsyncIndexInitManager) ResetStatus() {
```
torcolvin, line 83: does this end up leaking doneChan? I think that doneChans responsibility is on the caller's side ,but it might be worth noting that.

rest/indextest/index_init_api_test.go
```diff
@@ -0,0 +1,323 @@
...
+
+			// update db config with new partition count - shouldn't create indexes at this point since they already exist
+			if dbConfig.Index == nil {
+				dbConfig.Index = &rest.IndexConfig{}
+			}
+			dbConfig.Index.NumPartitions = base.Ptr(test.newPartitions)
+			t.Logf("dbconfig: %+v", dbConfig)
+			rest.RequireStatus(t, rt.ReplaceDbConfig(dbName, dbConfig), http.StatusCreated)
+
+			// cleanup old indexes
```
torcolvin: add jira ticket to be able to find this test in the next PR

rest/indextest/index_init_api_test.go
```diff
@@ -0,0 +1,323 @@
...
+				}
+			}, 30*time.Second, 1*time.Second)
+			assertNumSGIndexPartitions(t, database)
+
+			// update db config with new partition count - shouldn't create indexes at this point since they already exist
+			if dbConfig.Index == nil {
+				dbConfig.Index = &rest.IndexConfig{}
+			}
+			dbConfig.Index.NumPartitions = base.Ptr(test.newPartitions)
+			t.Logf("dbconfig: %+v", dbConfig)
```
torcolvin:
```suggestion
```

rest/database_init_manager.go
```diff
@@ -106,10 +104,15 @@ func (m *DatabaseInitManager) InitializeDatabase(ctx context.Context, startupCon
...
+	indexOptions := m.buildIndexOptions(dbConfig, useLegacySyncDocsIndex)
+
+	// allow the test callback to be overridden by the caller if desired
+	if statusCallback == nil {
+		statusCallback = m.testCollectionStatusUpdateCallback
+	}
 
 	// Create new worker and add this caller as a watcher
-	worker := NewDatabaseInitWorker(ctx, dbConfig.Name, n1qlStore, collectionSet, indexOptions, m.collectionCompleteCallback)
+	worker := NewDatabaseInitWorker(context.WithoutCancel(ctx), dbConfig.Name, n1qlStore, collectionSet, indexOptions, statusCallback)
```
torcolvin, line 115: I'm careful to watch for when we need to mark the context as non cancellable in a PR and I wonder if this is the right place to do to it. I think this is fine, but in the case where this is called from the bootstraping polling that context will not be cancelled.

The pattern we use elsewhere is https://github.com/couchbase/sync_gateway/blob/01fe76185503ce0302b6c18caf419ef6e9dd94e6/rest/admin_api.go#L120 but I do like using std library `WithoutCancel` and also having a typed non cancellable context if we need one in `InitializeDatabase` or in BackgroundManager if appropriate.

There's a quirk about logging here, where setting up a non cancellable context in different ways means there might be a correlation ID from an http request here where there is not typically normally. https://github.com/couchbase/sync_gateway/blob/01fe76185503ce0302b6c18caf419ef6e9dd94e6/rest/admin_api.go#L120

I think I like having the http correlation ID here so we know from when this was called, but this means that the correlation id will be logged after the request has been returned, and I do not know if this is confusing.

rest/admin_api.go
```diff
@@ -266,6 +267,112 @@ func (h *handler) handleDbOffline() error {
...
+	done, err := h.server.DatabaseInitManager.InitializeDatabaseWithStatusCallback(h.ctx(), h.server.initialStartupConfig, &newDbConfig, statusCallback, useLegacySyncDocsIndex)
+	if err != nil {
+		return err
+	}
+
+	opts := map[string]interface{}{
+		"statusMap": &statusMap,
+		"doneChan":  done,
+	}
+	err = h.db.AsyncIndexInitManager.Start(h.ctx(), opts)
```
torcolvin, line 363: I commented above about this, but there's some thought as to whether this should be a non cancellable context, but if this is an annoying refactor, let's do it later.

But when we call `GetStatus` and return an error, having the correlation id on this context would be useful.

torcolvin (commented):
I left a comment with questions but not a decision about how I'd like the API to work. I'm OK with deferring fixing this, or fixing it nicely to get this PR in and hitting in a follow up PR.

I do think this is probably very easily to get wrong if the operations in tests just happen to be shorter  than the request context cancellation.

db/background_mgr_async_index_init.go
bbrks: I think I can `nil` it here like the map, although with some additional locking required.

It will be soon overwritten with the new doneChan via the next invocation (ResetStatus only gets called when a new job runs), but feels safe enough to do.

I don't think it leaks anything as-is though. I left a comment in the struct to state who's responsibility this field is.

db/background_mgr.go
bbrks: I think this was an overzealous bugfix. Isn't required here which would affect other things more generally than the bug I was trying to fix. Reverted.

rest/database_init_manager.go
bbrks: The actual reason for `WithoutCancel` here is to avoid the job being killed by being associated with the original HTTP request context - this can be cancelled once the initial request finishes and was added as a bugfix. Unfortunately `RestTester` does not reproduce this bug, since it's not making actual HTTP requests through the stdlib http package. I found this bug via manual testing.

In all other cases where we use a `NonCancellableContext`, the new process runs indefinitely and logs a lot (starting a database via REST API call), where we definitely don't want the original HTTP request ID to be kept around for logging.

In this case, I think it's slightly different, in that it's a background job that is expected to finish, even if it's hours long, and only `WaitForIndexesOnline` logging. Resync does similar but generates a unique ID for each run. I think that is a bit overkill for this instead of just keeping the original HTTP request ID.

rest/indextest/index_init_api_test.go
```diff
@@ -123,18 +125,17 @@ func TestChangeIndexPartitions(t *testing.T) {
 						assert.Equal(c, db.CollectionIndexStatusReady, status)
 					}
 				}
-			}, 30*time.Second, 1*time.Second)
+			}, 1*time.Minute, 1*time.Second)
```
torcolvin, line 128: Did you find this really takes a long time? I find server is slow and if you have multiple nodes it is really slow.

Is it worth polling more frequently than 1 sec to make the test faster, or do these take 20-30 secs?

Because this is so unexpected, I might add partitionIndexPollingTime, partitionIndexPollingExpiration for times since this is duplicated.

torcolvin (approved):
LGTM

rest/indextest/index_init_api_test.go
bbrks: It's usually reasonably quick (~7 sec locally), but I saw it fail once on Jenkins where it took slightly longer than 30sec 

I don't think it's worth polling more frequently because at most you save 1 second, but there's also the interaction between this test retry loop and the `WaitForIndexesOnline` backoff retry loop which can sleep for much longer.

PR #7499, opened 2025-04-24
CBG-4614 Add toggle for using separate principal indexes

torcolvin:
CBG-4614 Add toggle for using separate principal indexes

- This will force all code to use the new indexes, but there will be a followup PR to actually decide when to use the new indexes. This forced all the
- Removed now used `Serverless` option from `DatabaseContext`. This still exists in rest code, but I don't want to change that code as it is too risky.
- simplified the code that used channels to use waitgroups for easier readability

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3084/

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -175,7 +175,6 @@ type DatabaseContextOptions struct {
 	GroupID                       string
 	JavascriptTimeout             time.Duration // Max time the JS functions run for (ie. sync fn, import filter)
 	UseLegacySyncDocsIndex        bool
-	Serverless                    bool // If running in serverless mode
 	Scopes                        ScopesOptions
 	MetadataStore                 base.DataStore // If set, use this location/connection for SG metadata storage - if not set, metadata is stored using the same location/connection as the bucket used for data storage.
 	MetadataID                    string         // MetadataID used for metadata storage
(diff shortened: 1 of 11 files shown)


rest/indextest/resync_test.go
```diff
@@ -58,7 +58,11 @@ func TestResyncWithoutIndexes(t *testing.T) {
 		// the sync docs index)
```
gregns1: NIT: remove this comment given the new configuration can have syncDocs and roles index?

db/util_testing.go
```diff
@@ -375,7 +375,7 @@ var viewsAndGSIBucketInit base.TBPBucketInitFunc = func(ctx context.Context, b b
 			UseXattrs:                  base.TestUseXattrs(),
 			NumReplicas:                0,
 			WaitForIndexesOnlineOption: base.WaitForIndexesDefault,
-			LegacySyncDocsIndex:        true, // Change in CBG-4614
+			LegacySyncDocsIndex:        false, // Change in CBG-4614
```
gregns1: Remove comment here?
```suggestion
			LegacySyncDocsIndex:        false,
```

db/database.go
```diff
@@ -1013,60 +1009,31 @@ func (c *DatabaseCollection) processForEachDocIDResults(ctx context.Context, cal
...
 	go func() {
-		defer close(ch)
-		roles, err := db.getRoleIDsUsingIndex(ctx, true)
-		ch <- data{
-			value: roles,
-			err:   err,
-		}
+		defer wg.Done()
+		includeDeleted := true
+		roles, getRolesErr = db.getRoleIDsUsingIndex(ctx, includeDeleted)
```
gregns1, line 1027: NIT: any reason why we don't just pass true in instead of using includeDeleted?

gregns1 (commented):
LGTM generally, just couple nit picks you can take or leave

db/database.go
torcolvin: I have a preference for defining constants so this code is easier to read, since `true` doesn't tell use what the argument would be. Probably doesn't matter much, but this is reasoning.

This is an ancient recommendation now, but this idea comes from `Code Complete` book from Steve McConnell.

gregns1 (approved):

PR #7500, opened 2025-04-24
Make tests work with views=true,xattrs=false

torcolvin:
- `user_xattr_key` is only available when xattrs are enabled
- `TestResyncRegenerateSequencesCorruptDocumentSequence` could be made to work without xattrs, but this doesn't feel a worthwhile use of time.

- [x] `GSI=true,xattrs=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3076/
- [x] `GSI=true,xattrs=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3081/
- [x] `GSI=false,xattrs=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3078/ has failures, but they seem to occur only when running a fuller set of tests together and have to do with views service being unhappy
- [x] `GSI=false,xattrs=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3079/ has failures, but they seem to occur only when running a fuller set of tests together and have to do with views service being unhappy

### DIFF
--- a/db/util_testing.go
+++ b/db/util_testing.go
@@ -191,8 +191,6 @@ func purgeWithDCPFeed(ctx context.Context, dataStore sgbucket.DataStore, tbp *ba
 	}
 
 	purgeCallback := func(event sgbucket.FeedEvent) bool {
-		var purgeErr error
-
 		processedDocCount.Add(1)
 		// We only need to purge mutations/deletions
 		if event.Opcode != sgbucket.FeedOpMutation && event.Opcode != sgbucket.FeedOpDeletion {
@@ -206,11 +204,7 @@ func purgeWithDCPFeed(ctx context.Context, dataStore sgbucket.DataStore, tbp *ba
 
 		key := string(event.Key)
 
-		if base.TestUseXattrs() {
-			purgeErr = dataStore.DeleteWithXattrs(ctx, key, []string{base.SyncXattrName})
-		} else {
-			purgeErr = dataStore.Delete(key)
-		}
+		purgeErr := dataStore.DeleteWithXattrs(ctx, key, []string{base.SyncXattrName})
 		if base.IsDocNotFoundError(purgeErr) {
 			// If key no longer exists, need to add and remove to trigger removal from view
 			_, addErr := dataStore.Add(key, 0, purgeBody)
--- a/db/util_testing.go
+++ b/db/util_testing.go
@@ -204,6 +204,7 @@ func purgeWithDCPFeed(ctx context.Context, dataStore sgbucket.DataStore, tbp *ba
 
 		key := string(event.Key)
 
+		// always delete with xattrs, since sometimes even SG_TEST_USE_XATTRS=false write xattrs to test for migration
 		purgeErr := dataStore.DeleteWithXattrs(ctx, key, []string{base.SyncXattrName})
 		if base.IsDocNotFoundError(purgeErr) {
 			// If key no longer exists, need to add and remove to trigger removal from view
--- a/rest/adminapitest/admin_api_test.go
+++ b/rest/adminapitest/admin_api_test.go
@@ -1385,14 +1385,17 @@ func TestResyncStopUsingDCPStream(t *testing.T) {
 	defer rt.Close()
 
 	numOfDocs := 1000
-	for i := 0; i < numOfDocs; i++ {
+	// gsi is slower than views, so update the number of documents for views so stopping the resync will not complete
+	if base.TestsDisableGSI() {
+		numOfDocs = 5000
+	}
+	for i := range numOfDocs {
 		rt.CreateTestDoc(fmt.Sprintf("doc%d", i))
 	}
 
-	err := rt.WaitForCondition(func() bool {
-		return int(rt.GetDatabase().DbStats.Database().SyncFunctionCount.Value()) == numOfDocs
-	})
-	assert.NoError(t, err)
+	rt.WaitForPendingChanges()
+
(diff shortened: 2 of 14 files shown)


rest/persistent_config_test.go
```diff
@@ -1493,15 +1493,18 @@ func persistentConfigTestCases() []persistentConfigTestCase {
...
+			name:        "document_persistence",
+			xattrConfig: false,
+		},
+	}
+	if base.TestUseXattrs() {
+		testCases = append(testCases, persistentConfigTestCase{
+			name:        "xattr_persistence",
+			xattrConfig: true,
+		})
+	}
```
bbrks: This doesn't seem right and I'm curious if anything broke with this because there's an underlying bug, or whether this was just cleanup?

We should continue testing both unconditionally, because it's valid to set `UseXattrConfig` for SG config even when `"enable_shared_bucket_access": false` (which is what `SG_TEST_USE_XATTRS` is intended for)

db/util_testing.go
```diff
@@ -206,11 +204,7 @@ func purgeWithDCPFeed(ctx context.Context, dataStore sgbucket.DataStore, tbp *ba
 
 		key := string(event.Key)
 
-		if base.TestUseXattrs() {
-			purgeErr = dataStore.DeleteWithXattrs(ctx, key, []string{base.SyncXattrName})
-		} else {
-			purgeErr = dataStore.Delete(key)
-		}
+		purgeErr := dataStore.DeleteWithXattrs(ctx, key, []string{base.SyncXattrName})
```
bbrks, line 208: This confused me at first but does make sense since `DeleteWithXattrs` will handle both xattr and non-xattr states of documents the same way.

Might be worth a comment?

bbrks (dismissed):

rest/persistent_config_test.go
torcolvin: I thought this would just make it more obvious what was going on, testing xattr persistence is for capella and non xattr is not that common. This issue is fixed by switching the bucket emptier to `purgeErr := dataStore.DeleteWithXattrs(ctx, key, []string{base.SyncXattrName})` since what happens with xattr persistence tests is that the sync metadata is preserved between tests, and an earlier test is not cleaned up for a later test.

This is actually a bug with xattrs=false,views=true since the views bucket pool uses bucket flush and that's less of a problem.

db/util_testing.go
torcolvin: Added a comment for this.

rest/persistent_config_test.go
torcolvin: Readded these tests.

rest/adminapitest/admin_api_test.go
```diff
@@ -2898,9 +2901,9 @@ func TestConfigsIncludeDefaults(t *testing.T) {
 	tb2 := base.GetTestBucket(t)
 	defer tb2.Close(ctx)
 
-	resp = rest.BootstrapAdminRequest(t, sc, http.MethodPut, "/db2/",
-		`{"bucket": "`+tb2.GetName()+`", "num_index_replicas": 0, "unsupported": {"disable_clean_skipped_query": true}}`,
-	)
+	resp = rest.BootstrapAdminRequest(t, sc, http.MethodPut, "/db2/", fmt.Sprintf(
+		`{"bucket": "%s", "num_index_replicas": 0, "num_index_replicas": 0, "enable_shared_bucket_access": %t, "use_views": %t, "unsupported": {"disable_clean_skipped_query": true}}`, tb2.GetName(), base.TestUseXattrs(), base.TestsDisableGSI(),
+	))
```
gregns1, line 2906: Is there a reason for `"num_index_replicas"` repeated here?

rest/xattr_upgrade_test.go
```diff
@@ -224,8 +227,10 @@ func TestCheckForUpgradeFeed(t *testing.T) {
 	assert.NoError(t, err, "Error writing doc w/ xattr")
 	rt.WaitForSequence(1)
 
+	// Use non admin channels since it is in the channel cache. Using admin using * channel which will not find this document.
+
 	// Attempt to update the documents via Sync Gateway.  Should trigger checkForUpgrade handling to detect metadata in xattr, and update normally.
-	changes := rt.PostChangesAdmin("/{{.keyspace}}/_changes", "{}")
+	changes := rt.GetChanges("/{{.keyspace}}/_changes", user)
```
gregns1: I am confused about what this test is testing for here. But changing this to get changes by user means this will just return the user doc. If that is what the test is trying to do (I'm really not sure myself) then we should alter the above comments to make it clear that the documnet we expect to see on changes here is the user doc and not `TestCheckForUpgrade` doc. 

If this test is trying to see the document `TestCheckForUpgrade` added above this changes call doesn't return that I think because the documents metadata doesn't actually have channel A allocated to it `"channels": [null]`.

gregns1 (commented):
Mostly looks good just couple of questions I have

rest/adminapitest/admin_api_test.go
torcolvin: accident, I've added a helper function for this.

torcolvin: I didn't add a helper function but I removed this.

rest/xattr_upgrade_test.go
torcolvin: Skipped with a comment, as per discussion with Adam. This doesn't work with GSI and that's expected. This test was for a rolling upgrade from non xattrs->xattrs and the non xattr nodes will already have this value in the cache.

This is an edge case that we won't fix so skipped the test.

gregns1 (approved):

PR #7504, opened 2025-04-28
Needs rework: `/_post_upgrade` cleanup handling for `num_partitions`

bbrks:
During development of this feature, I realized this approach is not going to work and will introduce more edge cases for `/_post_upgrade` that seems unacceptable.

We may need to change how `ServerContext.PostUpgrade()` operates, such that it can consider all database configurations during its evaluation of which indexes to cleanup. Doing this in a per-database loop is not enough to prevent accidental removal of in-use indexes for databases with varying configurations.

This was also a bug with xattrs, but was far less likely to happen than differing numbers of `num_partitions`.

See `FIXME` in code to see thought process and where the current approach started to unravel.

- n/a


torcolvin:
Reimplementing this in a different way.

PR #7505, opened 2025-04-28
CBG-4607: Allow separate principal index creation using the async index REST API

bbrks:
CBG-4607

Allow separate principal index creation using the async index REST API
- Make `num_partitions` optional since now users might only care about `separate_principal_indexes` on this API
- Specifying `"separate_principal_indexes": true` will cause the new user and role specific indexes to be created
- Omitting this value will fall back to whatever the database context has for `UseLegacySyncDocsIndex()`

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3094/

### DIFF
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -3144,6 +3144,10 @@ IndexSettings:
       description: The number of partitions to use for the large indexes created by Sync Gateway. It is not recommended to set this unless you require additional horizontal scalability for individual indexes and have appropriately scaled your Query nodes to handle the increased query parallelism. If set, the recommended number is 8 and does not need to be directly related to the number of your Query nodes. Ensure documentation is read to understand the performance tradeoffs and instructions for migration if you have previously run with only one partition. See [/{db}/_index_init](#operation/post_db-_index_init) for more information.
       type: number
       default: 1
+    separate_principal_indexes:
+      description: Whether to create separate indexes for users and roles instead of a single larger syncDocs index.
+      type: boolean
+      default: false
 IndexInitStatus:
   description: The status of an asynchronous indexes initialization operation.
   type: object
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -3144,8 +3144,11 @@ IndexSettings:
       description: The number of partitions to use for the large indexes created by Sync Gateway. It is not recommended to set this unless you require additional horizontal scalability for individual indexes and have appropriately scaled your Query nodes to handle the increased query parallelism. If set, the recommended number is 8 and does not need to be directly related to the number of your Query nodes. Ensure documentation is read to understand the performance tradeoffs and instructions for migration if you have previously run with only one partition. See [/{db}/_index_init](#operation/post_db-_index_init) for more information.
       type: number
       default: 1
-    separate_principal_indexes:
-      description: Whether to create separate indexes for users and roles instead of a single larger syncDocs index.
+    create_separate_principal_indexes:
+      description: |-
+        Whether to create separate indexes for users and roles instead of a single larger syncDocs index.
+
+        The separate principal indexes are smaller and used automatically for new database deployments. To remove the syncDocs index, wait for this to complete, restart all Sync Gateway instances and run  [POST /_post_upgrade](#operation/post__post_upgrade).
       type: boolean
       default: false
 IndexInitStatus:
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -322,9 +322,6 @@ func (h *handler) handlePostIndexInit() error {
 	}
 
 	currentDbConfig := h.server.GetDatabaseConfig(h.db.Name)
-	if req.NumPartitions != nil && currentDbConfig.NumIndexPartitions() == *req.NumPartitions {
-		return base.HTTPErrorf(http.StatusBadRequest, "num_partitions is already %d", *req.NumPartitions)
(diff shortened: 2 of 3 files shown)


docs/api/components/schemas.yaml
```diff
@@ -3147,6 +3147,10 @@ IndexSettings:
       description: The number of partitions to use for the large indexes created by Sync Gateway. It is not recommended to set this unless you require additional horizontal scalability for individual indexes and have appropriately scaled your Query nodes to handle the increased query parallelism. If set, the recommended number is 8 and does not need to be directly related to the number of your Query nodes. Ensure documentation is read to understand the performance tradeoffs and instructions for migration if you have previously run with only one partition. See [/{db}/_index_init](#operation/post_db-_index_init) for more information.
       type: number
       default: 1
+    separate_principal_indexes:
+      description: Whether to create separate indexes for users and roles instead of a single larger syncDocs index.
```
torcolvin: We should document this behavior. I think the default is do not do anything, but if there are already separate principal indexes.

Thinking more about this, we don't need to support going the opposite direction, so maybe this should just be called `create_separate_principal_indexes` if that is simpler. And then it can no-op if you call it again.

The num_partitions does not check the existing db state and I don't know if it's helpful to have this be restricted more strongly?

bbrks: It doesn't check db state but it does check if the existing config has a matching partition count

```go
	currentDbConfig := h.server.GetDatabaseConfig(h.db.Name)
	if req.NumPartitions != nil && currentDbConfig.NumIndexPartitions() == *req.NumPartitions {
		return base.HTTPErrorf(http.StatusBadRequest, "num_partitions is already %d", *req.NumPartitions)
	}
```

I assumed `UseLegacySyncDocsIndex()` would be based on the existence of an index on startup but maybe that's wrong - there's not really an equivalent for this option AFAIK, since we're not planning on exposing it to `DbConfig`?

rest/admin_api.go
```diff
@@ -350,7 +351,10 @@ func (h *handler) handlePostIndexInit() error {
 		}
 	}
 
-	useLegacySyncDocsIndex := h.db.UseLegacySyncDocsIndex() // TODO: CBG-4607: Change when this API is updated to support the new principal indexes
+	useLegacySyncDocsIndex := h.db.UseLegacySyncDocsIndex()
+	if req.SeparatePrincipalIndexes != nil {
+		useLegacySyncDocsIndex = !(*req.SeparatePrincipalIndexes)
```
adamcfraser, line 353: If this doesn't change the value of useLegacySyncDocsIndex (from line 354), do you think we should be returning an error similar to the num_partitions handling on line 326?

adamcfraser (commented):
Question about the no-op case, otherwise looks fine.

rest/admin_api.go
torcolvin: I thought about this and this was in Ben's PR originally. I do like making an API more restrictive.

The reason _not_ to do this is if a database is running but an index was deleted, this would let you recreate it without taking the database down.

This is a pretty niche case and I can be swayed either way. Since it is pretty much idempotent and pretty quick if all the indexes are recreated, and it doesn't put a lot of load on the system, this didn't seem bad.

rest/admin_api.go
```diff
@@ -278,14 +278,15 @@ func (h *handler) handleGetIndexInit() error {
...
-	NumPartitions *uint32 `json:"num_partitions"`
+	NumPartitions            *uint32 `json:"num_partitions"`
+	SeparatePrincipalIndexes *bool   `json:"create_separate_principal_indexes"`
 }
 
 func (req PostIndexInitRequest) Validate() error {
-	if req.NumPartitions == nil {
-		return base.HTTPErrorf(http.StatusBadRequest, "num_partitions is required")
+	if req.NumPartitions == nil && req.SeparatePrincipalIndexes == nil {
+		return base.HTTPErrorf(http.StatusBadRequest, "at least one of num_partitions or separate_principal_indexes is required")
```
```suggestion
		return base.HTTPErrorf(http.StatusBadRequest, "at least one of num_partitions or create_separate_principal_indexes is required")
```

rest/admin_api.go
bbrks: I'm caught up on this now. I'm certainly open to the idea of removing the error for `num_partitions` and letting the request run through init/readiness checks again.

bbrks (commented):
outdated field name in error msg and input below on no-op case

rest/indextest/index_init_api_test.go
```diff
@@ -227,22 +227,6 @@ func TestChangeIndexPartitionsErrors(t *testing.T) {
...
-		t.Skip("This test only works against Couchbase Server with GSI enabled")
-	}
-
-	// requires index init
-	base.LongRunningTest(t)
-
-	rt := rest.NewRestTester(t, &rest.RestTesterConfig{DatabaseConfig: &rest.DatabaseConfig{DbConfig: rest.DbConfig{Index: &rest.IndexConfig{NumPartitions: base.Ptr(uint32(2))}}}})
-	defer rt.Close()
-
-	resp := rt.SendAdminRequest(http.MethodPost, "/{{.db}}/_index_init", `{"num_partitions":2}`)
```
bbrks, line 241: Should be able to move this `2 -> 2` use-case into a `TestChangeIndexPartitions` subtest now just for coverage.

adamcfraser (approved):

PR #7507, opened 2025-04-30
CBG-4619: skip assigning nil body attachments to _attachments property if _attachments is uses as json value

gregns1:
CBG-4619

We currently do a check for the presence of _attachment in byte array of the incoming body on replication as optimisation to avoid attachment processing if its not needed. Issue is this will pick up the presence of _attachments if the key word is used as value in json document not the key. 
When using _attachments as json value in a document body and there is a previous version of the doc available we do some extra processing and attempt to fetch the attachment metadata from the body. When doing so and there is no attachment meta available the `GetBodyAttachments` function will return nil AttachmentMeta type and further down we will assign a nil `bodyAtts` to the _attachments property in the body object. This causes the `downloadOrVerifyAttachments` function to fail as _attachments property is in the body but is nil thus is invalid. See example error below:
```
2025-04-28T11:53:03.831+01:00 [ERR] c:[636a9852] db:db col:sg_test_0 Error during downloadOrVerifyAttachments for doc <ud>doc1</ud>/2-abc: 400 Invalid _attachments -- db.(*blipHandler).processRev() at blip_handler.go:1203
```
This fix here is to perform a nil check on body[BodyAttachments] and if nil skip all the attachment processing.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3079/

### DIFF
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -1127,9 +1127,7 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 	// Pull out attachments
 	if injectedAttachmentsForDelta || bytes.Contains(bodyBytes, []byte(BodyAttachments)) {
 		body := newDoc.Body(bh.loggingCtx)
-		// The above check for presence of _attachments in the json body will pass even if _attachments is
-		// used for json value rather than json key. So we need to perform a nil check on _attachments body property
-		// and if it returns nil we should skip any processing below on the document body.
+		// The bytes.Contains([]byte(BodyAttachments)) check will pass even if _attachments is not a toplevel key but rather a nested key or subkey. That check is an optimization to avoid having to unmarshal the document if there are no attachments. Therefore, check again that the unmarshalled body contains BodyAttachments.
 		if body[BodyAttachments] != nil {
 
 			var currentBucketDoc *Document
--- a/db/blip_handler.go
+++ b/db/blip_handler.go
@@ -1150,54 +1150,60 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 
 		// Do we have a previous doc? If not don't need to do this check
 		if currentBucketDoc != nil {
+			// CBG-4619: The above check for presence of _attachments in the json body will pass even if _attachments is
+			// used for json value rather than json key. So we need to perform a nil check after fetching Body attachments
+			// in GetBodyAttachments() and if it returns nil we should skip any processing below on the document body
+			// attachments. downloadOrVerifyAttachments() will effectively be a no-op when no document body attachments are found
 			bodyAtts := GetBodyAttachments(body)
-			currentDigests = make(map[string]string, len(bodyAtts))
-			for name, value := range bodyAtts {
-				// Check if we have this attachment name already, if we do, continue check
-				currentAttachment, ok := currentBucketDoc.Attachments[name]
-				if !ok {
-					// If we don't have this attachment already, ensure incoming revpos is greater than minRevPos, otherwise
-					// update to ensure it's fetched and uploaded
-					bodyAtts[name].(map[string]interface{})["revpos"] = updatedRevPos
-					continue
+			if bodyAtts != nil {
+				currentDigests = make(map[string]string, len(bodyAtts))
+				for name, value := range bodyAtts {
+					// Check if we have this attachment name already, if we do, continue check
+					currentAttachment, ok := currentBucketDoc.Attachments[name]
+					if !ok {
+						// If we don't have this attachment already, ensure incoming revpos is greater than minRevPos, otherwise
+						// update to ensure it's fetched and uploaded
+						bodyAtts[name].(map[string]interface{})["revpos"] = updatedRevPos
+						continue
+					}
+
+					currentAttachmentMeta, ok := currentAttachment.(map[string]interface{})
+					if !ok {
+						return base.HTTPErrorf(http.StatusInternalServerError, "Current attachment data is invalid")
+					}
+
+					currentAttachmentDigest, ok := currentAttachmentMeta["digest"].(string)
+					if !ok {
+						return base.HTTPErrorf(http.StatusInternalServerError, "Current attachment data is invalid")
+					}
+					currentDigests[name] = currentAttachmentDigest
+
+					incomingAttachmentMeta, ok := value.(map[string]interface{})
+					if !ok {
+						return base.HTTPErrorf(http.StatusBadRequest, "Invalid attachment")
+					}
+
+					// If this attachment has data then we're fine, this isn't a stub attachment and therefore doesn't
+					// need the check.
+					if incomingAttachmentMeta["data"] != nil {
+						continue
+					}
+
+					incomingAttachmentDigest, ok := incomingAttachmentMeta["digest"].(string)
+					if !ok {
+						return base.HTTPErrorf(http.StatusBadRequest, "Invalid attachment")
+					}
+
+					// Compare the revpos and attachment digest. If incoming revpos is less than or equal to minRevPos and
+					// digest is different we need to override the revpos and set it to the current revision to ensure
+					// the attachment is requested and stored. revpos provided for SG/CBL<4.0 compatibility but is no longer used by Sync Gateway.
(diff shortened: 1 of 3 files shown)


rest/blip_api_attachment_test.go
```diff
@@ -670,3 +670,64 @@ func TestBlipLegacyAttachDocUpdate(t *testing.T) {
...
+
+		opts := &BlipTesterClientOpts{SupportedBLIPProtocols: SupportedBLIPProtocols}
+		btc := btcRunner.NewBlipTesterClientOptsWithRT(rt, opts)
+		defer btc.Close()
+
+		btcRunner.StartPush(btc.id)
+
+		// add rev with _attachments property as value in json
+		doc1Version := btcRunner.AddRev(btc.id, doc1ID, EmptyDocVersion(), []byte(`{"data": "_attachments"}`))
+		rt.WaitForVersion(doc1ID, doc1Version)
```
torcolvin: I'd probably just parameterize this test with bodies := string{}

I like the test cases here but I'd be tempted to add one more for ignoring a non toplevel key:

`{"parent": { "_attachments": "data" } }`

If it's easy to write as non a string for readability you could use `base.MustJSONMarshal(t, body))` get bytes for AddRev

rest/blip_api_attachment_test.go
```diff
@@ -670,3 +670,64 @@ func TestBlipLegacyAttachDocUpdate(t *testing.T) {
...
+		rt.WaitForVersion(doc1ID, doc1Version2)
+
+		doc2Version2Body := []byte(`{"data": {"textfield": "_attachments"}}`)
+		doc2Version2 := btcRunner.AddRev(btc.id, doc2ID, &doc2Version, doc2Version2Body)
+		rt.WaitForVersion(doc1ID, doc2Version2)
+
+		doc3Version2Body := []byte(`{"data1": {"data": {"textfield": "_attachments"}}}`)
+		doc3Version2 := btcRunner.AddRev(btc.id, doc3ID, &doc3Version, doc3Version2Body)
+
+		// assert that the bodies that are replicated are as expected
```
torcolvin: I don't think this assertion is correct, you want to test that SGW has this document. I think as soon as you call `AddRev` to the document that this version will be present on btcRunner, which now supports storing data without a replication. I think you want rt.WaitForVersion.

rest/blip_api_attachment_test.go
```diff
@@ -670,3 +670,64 @@ func TestBlipLegacyAttachDocUpdate(t *testing.T) {
...
+		rt.WaitForVersion(doc1ID, doc1Version)
+
+		// add some revs with _attachments at value in nested json
+		doc2Version := btcRunner.AddRev(btc.id, doc2ID, EmptyDocVersion(), []byte(`{"data": {"textfield": "_attachments"}}`))
+		rt.WaitForVersion(doc1ID, doc2Version)
+
+		doc3Version := btcRunner.AddRev(btc.id, doc3ID, EmptyDocVersion(), []byte(`{"data": {"data": {"textfield": "_attachments"}}}`))
+		rt.WaitForVersion(doc1ID, doc3Version)
+
+		// add rev2 for each doc and wait for each to be replicated to SGW
```
torcolvin: I would add a comment here to say make sure updating new content onto a parent document works OK, since there is different logic for pushing new data, since SG will scan the parent doc for attachment keys

db/blip_handler.go
```diff
@@ -1150,54 +1150,60 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 
 		// Do we have a previous doc? If not don't need to do this check
 		if currentBucketDoc != nil {
+			// CBG-4619: The above check for presence of _attachments in the json body will pass even if _attachments is
+			// used for json value rather than json key. So we need to perform a nil check after fetching Body attachments
+			// in GetBodyAttachments() and if it returns nil we should skip any processing below on the document body
+			// attachments. downloadOrVerifyAttachments() will effectively be a no-op when no document body attachments are found
```
torcolvin: nit: I would remove this jira ticket on this comment, since this ticket was just a bug and the origin of the bucket isn't important if there's a regression test. Once code gets moved around, it's probably better to check git history.

I was still confused about the comment because I didn't understand why the `for name, value := range bodyAtts` didn't cover this case as a no-op.

Possibly to make this more foolproof I'd create a function like `func createAttachmentsMeta(previousDoc * Document, incomingBody db.Body, updatedRevPos) (existingAttachmentDigests, map[string]string, meta *AttachmentsMeta, err error)` with a docstring like:

```createAttachmentsMeta constructs the metadata for attachments for the incoming rev. Validates the data against the previous Document if one is available. If no attachments are found, there will be no error and nil AttachmentsMeta. If there is validation error for any fields in the attachment section of the body a base.HTTPError will be returned.```

Then we can be thoughtful on how to update the body lower down, since `GetBodyAttachments` is actually called a second time lower:

```
		newDoc.DocAttachments = GetBodyAttachments(body)
		delete(body, BodyAttachments)
		newDoc.UpdateBody(body)
```

I don't think that `downloadOrVerifyAttachments` even needs a body and could take `AttachmentsMeta` but `ForEachStubAttachment` call signature needs changing.

LMK what you think about changing this and see if it improves readability. I broke this code handling more than one writing anemone code recently so I'm sensitive to ways it could be improved.

Comment (gregns1) [reply db/blip_handler.go]: There is definitely ways this code could be tidied up and you suggest some things I wanted to do especially removing passing the body to download and verify and thought about moving `bodyAtts := GetBodyAttachments(body)` to outside the if statement on `currentBucketDoc` so we can avoid potentially calling it again here: 
```
newDoc.DocAttachments = GetBodyAttachments(body) 
delete(body, BodyAttachments) 
newDoc.UpdateBody(body)
```

I assumed given its going into a potential MP we would want to keep changes minimal to reduce risk here so decided against it. 

The case wasn't a no-op as even though we wouldn't iterate through bodyAtts it would then subsequently be assigned to the _attachment body property after the iteration. That then broke things inside download and verify. 

I am pushing a new commit that simplifies things and will skip all this attachment processing altogether if body[BodyAttachments] is nil. We can maybe discuss later if its worth tidying the rest of this up as you mention above.

Comment (torcolvin) [reply db/blip_handler.go]: Does `injectedAttachmentsForDelta` do anything here? I can't see it doing anything but I am honestly not sure. It seems like deltaSrc updates `newDoc` but not `body`. Can you write some tests to verify that and maybe add it to parametrized test suite just in case?

I agree with you that somewhere in comment I got overzealous about refactoring.

File: rest/blip_api_attachment_test.go (hunk 2 above)
Comment (gregns1) [reply rest/blip_api_attachment_test.go]: Totally make mistake here, wanted to verify the do bodes are as expected on rest tester, making a change now to do that.

File: db/blip_handler.go (hunk 4 above)
Comment (gregns1) [reply db/blip_handler.go]: Wrote delta sync test for this although I think we're mostly safe on that, seems to update newDoc and in this code we're changing we grab body from newDoc anyway in `body := newDoc.Body(bh.loggingCtx)`

File: db/blip_handler.go [hunk 5]
```diff
@@ -1127,87 +1127,93 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
...
-				// Check if we have this attachment name already, if we do, continue check
-				currentAttachment, ok := currentBucketDoc.Attachments[name]
-				if !ok {
-					// If we don't have this attachment already, ensure incoming revpos is greater than minRevPos, otherwise
-					// update to ensure it's fetched and uploaded
-					bodyAtts[name].(map[string]interface{})["revpos"] = updatedRevPos
-					continue
+		// The above check for presence of _attachments in the json body will pass even if _attachments is
+		// used for json value rather than json key. So we need to perform a nil check on _attachments body property
+		// and if it returns nil we should skip any processing below on the document body.
```
torcolvin:
```suggestion
		// The bytes.Contains([]byte(BodyAttachments)) check will pass even if _attachments is not a toplevel key but rather a nested key or subkey. That check is an optimization to avoid having to unmarshal the document if there are no attachments. Therefore, check again that the unmarshalled body contains BodyAttachments.
```

one more wording tweak

torcolvin (dismissed):

torcolvin (approved):

PR #7514, opened 2025-05-07
Allow RetryLoop to be type safe

torcolvin:
- Use generics to allow use of type safe RetryLoops
- Drop need in tests to always assert on errr != nil for some rest utility test functions

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3096/ (known independent flake)

### DIFF
--- a/rest/utilities_testing.go
+++ b/rest/utilities_testing.go
@@ -961,19 +961,19 @@ func (rt *RestTester) SendUserRequest(method, resource, body, username string) *
 	return rt.Send(RequestByUser(method, rt.mustTemplateResource(resource), body, username))
 }
 
-func (rt *RestTester) WaitForNUserViewResults(numResultsExpected int, viewUrlPath string, user auth.User, password string) (viewResult sgbucket.ViewResult, err error) {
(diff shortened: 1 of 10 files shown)


rest/utilities_testing.go
```diff
@@ -2404,12 +2383,12 @@ func WaitAndAssertBackgroundManagerState(t testing.TB, expected db.BackgroundPro
 
 func WaitAndAssertBackgroundManagerExpiredHeartbeat(t testing.TB, bm *db.BackgroundManager) bool {
```
gregns1, line 2384: This doesn't look like its used at all anywhere, should we just remove it?

gregns1 (approved):
LGTM - can take or leave the comment I have made

rest/utilities_testing.go
torcolvin: I'll remove in a different PR, I have one that wipes out all the public functions that are unused using https://pkg.go.dev/golang.org/x/tools/cmd/deadcode

Since Sync Gateway isn't a library, we can easily remove dead public functions.

PR #7515, opened 2025-05-07
fix processEntry mode on caching tool

gregns1:
After #7483 got merged processEnrty mode was broken due to some changes. Also needed to add ability to add configured number of channels per doc from system channels. 

## Pre-review checklist

### DIFF
--- a/tools/cache_perf_tool/processEntry.go
+++ b/tools/cache_perf_tool/processEntry.go
@@ -50,8 +50,8 @@ func (p *processEntryGen) nodeWrites(ctx context.Context, node *sgwNode, delay t
 	if delay.Nanoseconds() == 0 {
 		// mutate as fast as possible
 		for {
-			chanMap = make(channels.ChannelMap)
-			for i := 0; i < p.numChansPerDoc; i++ {
+			chanMap = make(channels.ChannelMap, p.numChansPerDoc)
+			for range p.numChansPerDoc {
 				if chanCountZeroWait == p.totalChans {
 					// when count gets to total number configured channels we rest index to 0
 					chanCountZeroWait = 0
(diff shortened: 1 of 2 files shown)


tools/cache_perf_tool/processEntry.go
```diff
@@ -74,6 +85,16 @@ func (p *processEntryGen) nodeWrites(ctx context.Context, node *sgwNode, delay t
 	ticker := time.NewTicker(delay)
 	defer ticker.Stop()
 	for {
+		chanMap = make(channels.ChannelMap)
+		for i := 0; i < p.numChansPerDoc; i++ {
```
torcolvin: very nit:

```suggestion
		chanMap = make(channels.ChannelMap, len(p.numChansPerDoc)
		for i := range p.numChansPerDoc {
```

tools/cache_perf_tool/processEntry.go
```diff
@@ -74,6 +85,16 @@ func (p *processEntryGen) nodeWrites(ctx context.Context, node *sgwNode, delay t
 	ticker := time.NewTicker(delay)
 	defer ticker.Stop()
 	for {
```
torcolvin, line 87: Should this be a function to deduplicate above where this passes in a time.Ticker with 0 so that the first one is deduplicated from this?

If the delay is actually 0, are we supposed to run this function twice?

tools/cache_perf_tool/processEntry.go
```diff
@@ -74,6 +85,16 @@ func (p *processEntryGen) nodeWrites(ctx context.Context, node *sgwNode, delay t
...
+		for i := 0; i < p.numChansPerDoc; i++ {
+			if chanCountWait == p.totalChans {
+				// when count gets to total number configured channels we rest index to 0
+				chanCountWait = 0
+			}
+			chanName := "test-" + strconv.Itoa(chanCountWait)
+			chanMap[chanName] = nil
+			chanCountWait++
+		}
 		timeStamp := time.Now()
```
torcolvin: Do you want timestamp to be before the select, or do you want this to be after the ticker is selected?

tools/cache_perf_tool/processEntry.go
gregns1: Can't pass 0 into ticket hence why I just went for this approach. No wouldn't expect this to run twice if node has 0 delay.

torcolvin (approved):

PR #7518, opened 2025-05-09
CBG-4637 fix example for /_all_dbs

torcolvin:
CBG-4637 fix example for /_all_dbs

Analyzed use of `oneOf` in our documentation. 
- removed oneOf in `userCtx` since it always returns a sequence number, never compound sequence
- oneOf bug only hits arrays https://github.com/Redocly/redoc/issues/2676 and not objects in getting db audit configuration


IsaacLambat (approved):
Looking much better 🎉

PR #7528, opened 2025-05-13
CBG-4634: remove mutex around initial sequence inside getInitialSequence()

gregns1:
CBG-4634

- Was expecting to have to move this around a bit but seems that race detectors don't complain so kept it simple. 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3109/

### DIFF
--- a/db/change_cache_test.go
+++ b/db/change_cache_test.go
@@ -921,8 +921,6 @@ func TestChannelQueryCancellation(t *testing.T) {
 	assert.NoError(t, err, "Put failed with error: %v", err)
 	require.NoError(t, db.changeCache.waitForSequence(ctx, 4, base.DefaultWaitForSequence))
 
-	// Flush the cache, to ensure view query on subsequent changes requests
-
 	// Issue two one-shot since=0 changes request.  Both will attempt a view query.  The first will block based on queryWg,
 	// the second will block waiting for the view lock
 	initialQueryCount := db.DbStats.Cache().ViewQueries.Value()
--- a/db/change_cache_test.go
+++ b/db/change_cache_test.go
@@ -922,7 +922,6 @@ func TestChannelQueryCancellation(t *testing.T) {
 	require.NoError(t, db.changeCache.waitForSequence(ctx, 4, base.DefaultWaitForSequence))
 
 	// Flush the cache, to ensure view query on subsequent changes requests
-	// require.NoError(t, db.FlushChannelCache(ctx))
 
 	// Issue two one-shot since=0 changes request.  Both will attempt a view query.  The first will block based on queryWg,
 	// the second will block waiting for the view lock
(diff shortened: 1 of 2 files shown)


torcolvin (commented):
Is it worth dropping this function?

The comment here https://github.com/couchbase/sync_gateway/blob/7456780e9e7689a7cf307d84d7c6575c86bb2bae/db/change_cache.go#L55 would need to be updated

From looking at this, I don't think there's a race unless `FlushChannelCache` will run at the same time as changes are coming in.

I think it's worth running the tests that call `FlushChannelCache` with a really high number of `-count` to ensure that there aren't data races in these tests that reset `initialSequence`

I'd remove https://github.com/couchbase/sync_gateway/blob/7456780e9e7689a7cf307d84d7c6575c86bb2bae/db/change_cache_test.go#L925 since it is unused.

Eyeballing the code, I think https://github.com/couchbase/sync_gateway/blob/7456780e9e7689a7cf307d84d7c6575c86bb2bae/rest/replicatortest/replicator_test.go#L4955 could race here because `Purge` doesn't wait for the cache to be updated. I suspect this could be fixed with `WaitForPendingChanges`.

The uses in `rest/changestest/changes_api_test.go` look OK assuming the cache waiters in each case are waiting for the right number of values. I'd be curious comments from members of the code base why `NewDCPCachingCountWaiter` is used rather than just `WaitForPendingChanges`. Maybe this code predated `WaitForPendingChanges`, since it just seems more robust than waiting for specific numbers of documents which you have to add. Unless you are doing an SDK write and waiting for import and then subsequent caching, `WaitForPendingChanges` naively seems better.

gregns1:
> https://github.com/couchbase/sync_gateway/blob/7456780e9e7689a7cf307d84d7c6575c86bb2bae/rest/replicatortest/replicator_test.go#L4955
> could race here because `Purge` doesn't wait for the cache to be updated. I suspect this could be fixed with `WaitForPendingChanges`.

This won't race as purging a doc will be ignored over the caching feed/import feed (delete mutation with no xattr) 

I ran changes test that call the test function `FlushChannelCache` many each 100 times with race and none failed so. I think we're fine. I did consider stopping then starting change cache in the test function to be extra safe but may be overkill if there are no test issues.

db/change_cache_test.go
```diff
@@ -922,7 +922,6 @@ func TestChannelQueryCancellation(t *testing.T) {
 	require.NoError(t, db.changeCache.waitForSequence(ctx, 4, base.DefaultWaitForSequence))
 
 	// Flush the cache, to ensure view query on subsequent changes requests
```
torcolvin: Remove this comment

torcolvin (approved):

PR #7530, opened 2025-05-13
CBG-4636: have check for skipped sequences on recent sequence handling

gregns1:
CBG-4636

We need to check if its possible that a recent_sequence got pushed to skipped on recent sequence handing, when pending gets to max or is flushed due to age we move the next expected sequence at the cache to the oldest pending sequence. Hence this loop was not running on some sequences that were in recent sequences on a doc but were also present in skipped sequences. This meant skipped would grow exponentially with no prospect of ever removing the sequences until we abandoned them. This lead to high sequence stable being stuck until abandoned sequences job was run. 

I chose not to do an actual check in skipped given this is done inside `processEntry` and it would be less expensive to just check if seq in recent sequence is less than current sequence and it is greater than or equal to the oldest skipped sequence.

Before changes high_seq_stable:
high_seq_stable_pre

After changes high_seq_stable:
high_seq_stable_post

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3114/

### DIFF
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -445,22 +445,21 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent) {
 
 	if len(syncData.RecentSequences) > 0 {
 		nextSequence := c.getNextSequence()
-		oldestSkipped := c.getOldestSkippedSequence(ctx)
 
 		for _, seq := range syncData.RecentSequences {
 			// seq < currentSequence means the sequence is not the latest allocated to this document
 			// seq >= nextSequence means this sequence is a pending sequence to be expected in the cache
 			// the two conditions above together means that the cache expect ue to run processEntry on this sequence as its pending
-			// If seq >= c.getOldestSkippedSequence(ctx) and seq < current sequence allocated to the doc this means
-			// that this sequence never arrived over the caching feed due to deduplication and was pushed to a skipped sequence list
-			// We need check for 0 oldestSkipped because if the skipped list is empty 0 is returned
-			isSkipped := (seq < currentSequence && seq >= oldestSkipped) && oldestSkipped != 0
+			// If seq < current sequence allocated to the doc and seq is in skipped list this means that this sequence
+			// never arrived over the caching feed due to deduplication and was pushed to a skipped sequence list
+			isSkipped := (seq < currentSequence && seq < nextSequence) && c.WasSkipped(seq)
 			if (seq >= nextSequence && seq < currentSequence) || isSkipped {
 				base.InfofCtx(ctx, base.KeyCache, "Received deduplicated #%d in recent_sequences property for (%q / %q)", seq, base.UD(docID), syncData.CurrentRev)
 				change := &LogEntry{
 					Sequence:     seq,
 					TimeReceived: timeReceived,
 					CollectionID: event.CollectionID,
+					Skipped:      isSkipped,
 				}
 
 				// if the doc was removed from one or more channels at this sequence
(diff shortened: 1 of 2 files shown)


db/change_cache.go
```diff
@@ -475,7 +475,9 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent) {
 		nextSequence := c.getNextSequence()
 
 		for _, seq := range syncData.RecentSequences {
-			if seq >= nextSequence && seq < currentSequence {
+			// we need to check if it's possible that one of the recent sequences has been pushed to skipped
+			isSkipped := seq < currentSequence && seq >= c.getOldestSkippedSequence(ctx)
+			if (seq >= nextSequence && seq < currentSequence) || isSkipped {
```
torcolvin, line 456: I think it would be helpful to document what each of these states mean. I wrote a guess but I'm not actually sure. One of the if statements ends up being duplicated in both predicates `seq < currentSequence`.

```suggestion
            // seq < currentSequence means the sequence is not the latest allocated, and is not ahead of the latest allocated (i.e. migrated in from another cluster)
            // seq >= nextSequence means this sequence is a pending sequence to be expected in a subsequent call to processEntry
            // seq >= c.getOldestSkippedSequence(ctx) means that this sequence never arrived over the caching feed and was pushed to a skipped sequence
			if seq < currentSequence && ( seq >= nextSequence || seq >= c.getOldestSkippedSequence(ctx)) {
```

`seq >= c.getOldestSkippedSequence(ctx)` acquires a read lock and while I do not think this is contentious because I doubt `recent_sequences` is large, does it make sense to move this outside the loop. Alternatively, is it wrong to move outside of the loop because there could be new skipped sequences while this is processing?

gregns1: I originally too this approach but went for this option given I wasn't totally sure at the time if it was avoidable. 

I've thought more about it today and reran test with moving the get oldest skipped call and I feel its safe to move it out of the loop.

db/change_cache.go
```diff
@@ -473,9 +473,16 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent) {
 
 	if len(syncData.RecentSequences) > 0 {
 		nextSequence := c.getNextSequence()
+		oldestSkipped := c.getOldestSkippedSequence(ctx)
```
bbrks: What happens when this is `0` because we've not skipped any sequences?

Everywhere else this is used it has a zero check, but the way its used below makes me think we'll be running `processEntry` many times for the same sequences in `RecentSequences` again since `isSkipped` evaluates to true until the list is pruned.

bbrks (changes requested):
Have a question about how this change interacts when we don't have any skipped sequences but do have recent sequences present on a doc

db/change_cache.go
```diff
@@ -473,9 +473,16 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent) {
...
 
 		for _, seq := range syncData.RecentSequences {
-			if seq >= nextSequence && seq < currentSequence {
+			// seq < currentSequence means the sequence is not the latest allocated to this document
```
adamcfraser, line 450: Functionally it looks like what you're doing here is:
1. Ignore recent sequences that aren't less than the doc's current sequence (existing corner case handling)
2. For any others, let n be the lowest sequence of [oldest skipped, nextSequence].  Send any recent sequences through process entry if they are greater than or equal to n.

This lets process entry do the work of determining whether they are skipped, pending, or already processed.

If that's accurate, I think you can reorganize the code to compute n outside the loop, which would make the the logic inside the loop easier to reason about.

adamcfraser (dismissed):

db/change_cache.go
```diff
@@ -447,12 +447,19 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent) {
 		nextSequence := c.getNextSequence()
 
 		for _, seq := range syncData.RecentSequences {
-			if seq >= nextSequence && seq < currentSequence {
+			// seq < currentSequence means the sequence is not the latest allocated to this document
+			// seq >= nextSequence means this sequence is a pending sequence to be expected in the cache
+			// the two conditions above together means that the cache expect ue to run processEntry on this sequence as its pending
```
```suggestion
			// the two conditions above together means that the cache expects us to run processEntry on this sequence as its pending
```

db/change_cache.go
```diff
@@ -700,9 +709,13 @@ func (c *changeCache) processEntry(ctx context.Context, change *LogEntry) channe
 	//   - principal mutations that don't increment sequence
 	// We can cancel processing early in these scenarios.
 	// Check if this is a duplicate of an already processed sequence
-	if sequence < c.nextSequence && !c.WasSkipped(sequence) {
-		base.DebugfCtx(ctx, base.KeyCache, "  Ignoring duplicate of #%d", sequence)
-		return nil
+	if sequence < c.nextSequence && !change.Skipped {
+		// check for presence in skippedSeqs, it's possible that change.skipped can be false but subsequently change
```
```suggestion
		// check for presence in skippedSeqs, it's possible that change.skipped can be false but a subsequent change
```

bbrks (dismissed):
Just a few comment typos but the actual change and test looks good to me

bbrks (approved):

PR #7531, opened 2025-05-13
CBG-4650 Use /db/_config offline= for persistent config in tests

torcolvin:
Calling `/db/_offline` doesn't actually stop all the processes like setting `{"offline": true}` does https://github.com/couchbase/sync_gateway/blob/c109db310a7ad121a12363f2b0b10c3eb631f3d6/db/database.go#L810

I keep getting burned by this in tests because notably import is still running. In addition, pull ISGR would be running, since only the changes feeds are shut down. Any active push replication can also continue while the database is offline. But any new connections will receive a 503, so the problem might be more limited in scope?

I think this change makes more sense as a default, and I had previously improved logging to suggest not to use the offline endpoints. https://github.com/couchbase/sync_gateway/commit/a82bd9305da18cc361e9786c56cef8335d5d6dc7 It might be worth changing https://docs.couchbase.com/sync-gateway/current/resync.html#updating-the-sync-function documentation as well.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3113/

### DIFF
--- a/rest/adminapitest/admin_api_test.go
+++ b/rest/adminapitest/admin_api_test.go
@@ -3017,7 +3017,8 @@ func TestDbOfflineConfigPersistent(t *testing.T) {
 	// Get config values before taking db offline
 	resp := rt.SendAdminRequest(http.MethodGet, "/{{.db}}/_config", "")
 	rest.RequireStatus(t, resp, http.StatusOK)
-	dbConfigBeforeOffline := resp.Body.String()
+	var dbConfigBeforeOffline rest.DatabaseConfig
+	require.NoError(t, base.JSONUnmarshal(resp.BodyBytes(), &dbConfigBeforeOffline))
 
 	resp = rt.SendAdminRequest(http.MethodGet, "/{{.keyspace}}/_config/import_filter", "")
 	rest.RequireStatus(t, resp, http.StatusOK)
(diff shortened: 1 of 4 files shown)


rest/adminapitest/admin_api_test.go
```diff
@@ -3033,7 +3034,17 @@ func TestDbOfflineConfigPersistent(t *testing.T) {
 	// Check offline config matches online config
 	resp = rt.SendAdminRequest(http.MethodGet, "/{{.db}}/_config", "")
 	rest.RequireStatus(t, resp, http.StatusOK)
```
bbrks: This test case is actually testing what used to happen and we should replace `rt.TakeDbOffline()` with something that _doesn't_ modify the stored config for this test instead of adding code to accommodate the changed fields.

We're ensuring that the database is able to restart and come back online, or that a single node can be taken offline without affecting others.

rest/adminapitest/resync_test.go
```diff
@@ -220,6 +220,7 @@ func TestResyncRegenerateSequencesPrincipals(t *testing.T) {
 				require.NotEqual(t, originalUserSeq, user.Sequence())
 			}
 
+			collection, ctx = rt.GetSingleTestDatabaseCollection()
```
bbrks, line 223: Is this required because taking the db offline w/ persistent config makes the collection reference unusable?

bbrks (changes requested):
> notably import is still running. In addition, pull ISGR would be running, since only the changes feeds are shut down. Any active push replication can also continue while the database is offline. But any new connections will receive a 503, so the problem might be more limited in scope?

This sounds like more like the underlying bug than test not taking databases offline in the right way. Is there a ticket already for correcting this behavior? I can still see users running into this even using persistent config by using the `/db/_offline` endpoints. Taking a single node offline still seems like a valid use-case for these older endpoints to me.

Given this isn't an area of focus for us right now, I am more than happy to defer this to a Jira ticket and close up the testing reliability gap with this change.

rest/adminapitest/resync_test.go
torcolvin: yep

torcolvin:
> > notably import is still running. In addition, pull ISGR would be running, since only the changes feeds are shut down. Any active push replication can also continue while the database is offline. But any new connections will receive a 503, so the problem might be more limited in scope?
> 
> This sounds like more like the underlying bug than test not taking databases offline in the right way. Is there a ticket already for correcting this behavior? I can still see users running into this even using persistent config by using the `/db/_offline` endpoints. Taking a single node offline still seems like a valid use-case for these older endpoints to me.
> 
> Given this isn't an area of focus for us right now, I am more than happy to defer this to a Jira ticket and close up the testing reliability gap with this change.

I do _not_ like the current behavior at all, and I worry that some customers are indeed using it.  https://jira.issues.couchbase.com/browse/CBG-4651 describes this behavior.

bbrks (approved):

PR #7533, opened 2025-05-15
CBG-4594: Move key filtering on mutation feed higher up stack

gregns1:
CBG-4594

- Moves the key filtering higher up stack to avoid sending docs we don't care about to DCP workers 
- Involves adding metadata keys to the dcp client to allow this filtering which meant changes to feed args in sgbucket given we use a shared interface to start the dcp feed for mutation feed between sgw and rosmar, I can't import metadata keys to rosmar given that would create a circular dependency. 
- In doing this change I noticed we were then subsequently filtering on docs on three places, the new DCP level, `ProcessFeedEvent` and then finally in `processEntry`. This felt wasteful to me so I added a doc type field to `FeedEvent` in sgbucket to avoid having to filter on key prefix in multiple places. 

I'm overall I'm not sure if this change will deliver much or a meaningful change and has required more changes than I was anticipating, so I'm interested in hearing others thoughts on whether its still worth having this change.

- [ ] Link upstream PRs - https://github.com/couchbase/sg-bucket/pull/135 https://github.com/couchbaselabs/rosmar/pull/48

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3117/

### DIFF
--- a/base/dcp_client_stream_observer.go
+++ b/base/dcp_client_stream_observer.go
@@ -9,8 +9,6 @@
 package base
 
 import (
-	"strings"
-
 	"github.com/couchbase/gocbcore/v10"
 	sgbucket "github.com/couchbase/sg-bucket"
 )
@@ -35,9 +33,13 @@ func (dc *DCPClient) SnapshotMarker(snapshotMarker gocbcore.DcpSnapshotMarker) {
 
 func (dc *DCPClient) Mutation(mutation gocbcore.DcpMutation) {
 
-	filterKey, dcpEventType := dc.filteredKey(mutation.Key)
-	if filterKey {
-		return
+	var dcpEventType sgbucket.FeedFilterType
+	var filterKey bool
+	if dc.FilterFunc != nil {
+		filterKey, dcpEventType = dc.FilterFunc(mutation.Key)
+		if filterKey {
+			return
+		}
 	}
 
 	e := mutationEvent{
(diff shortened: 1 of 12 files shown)


base/dcp_client_stream_observer.go
```diff
@@ -136,6 +144,32 @@ func (dc *DCPClient) SeqNoAdvanced(seqNoAdvanced gocbcore.DcpSeqNoAdvanced) {
 	})
 }
 
-func (dc *DCPClient) filteredKey(key []byte) bool {
-	return false
+// filteredKey will filter keys we don't care about off the mutation DCP feed and will return DCP event type on non-filtered docs
+func (dc *DCPClient) filteredKey(key []byte) (bool, sgbucket.FeedItemDocType) {
```
torcolvin: I did some experimental code for this by having this a function that you could pass into `sgbucket.FeedArguments` and have it optionally called there or not.

In the first implementation of this PR, this code is duplicated in rosmar, which I think is quite hazardous. Additionally, this actually doesn't take place for the cbgt implementation of the import feed, but this code is functional for CE import feed and CE caching feed. This is actually pretty confusing!

I think we would get most of the benefit, if you want, of having this logic at the top of `c *changeCache) DocChanged` to return a `docType` (defined in SG).

Moving this logic for the purposes of testing the caching without CBS DCP isn't really correct, since this function will always be called on Sync Gateway side before this function is executed. But if you did want to skip this processing, you could have a separate function in `changeCache` that works after this filter function is called.

torcolvin: I think that this isn't worth duplicating in 3 places, and we could put this in `processFeedEvent` but I'm not sure this optimization will gain us much.

PR #7538, opened 2025-05-20
CBG-4472 remove warnings for null document bodies

torcolvin:
CBG-4472 remove warnings for null document bodies

When writing this test, discovered that rosmar returns `sgbucket.XattrMissingError` rather than `ErrXattrNotFound` for these two cases, and so swapped to use `base.IsXattrNotFoundError`. I think this code could be dropped for a 3.2.5 backport if it is too risky, but I do not think it is.

The tests use the new blip test handling code in 3.3, so might be a bit tricky to backport, but I do not think that's critical.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3121/

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -2685,7 +2685,7 @@ func (db *DatabaseCollectionWithUser) RevDiff(ctx context.Context, docid string,
 
 	doc, err := db.GetDocSyncDataNoImport(ctx, docid, DocUnmarshalHistory)
 	if err != nil {
-		if !base.IsDocNotFoundError(err) && !errors.Is(err, base.ErrXattrNotFound) {
+		if !base.IsDocNotFoundError(err) && !base.IsXattrNotFoundError(err) {
 			base.WarnfCtx(ctx, "RevDiff(%q) --> %T %v", base.UD(docid), err, err)
 		}
 		missing = revids
@@ -2747,7 +2747,7 @@ func (db *DatabaseCollectionWithUser) CheckProposedRev(ctx context.Context, doci
 	}
 	doc, err := db.GetDocSyncDataNoImport(ctx, docid, level)
 	if err != nil {
-		if !base.IsDocNotFoundError(err) && !errors.Is(err, base.ErrXattrNotFound) {
+		if !base.IsDocNotFoundError(err) && !base.IsXattrNotFoundError(err) {
 			base.WarnfCtx(ctx, "CheckProposedRev(%q) --> %T %v", base.UD(docid), err, err)
 			return ProposedRev_Error, ""
 		}
--- a/db/crud_test.go
+++ b/db/crud_test.go
@@ -1807,3 +1807,24 @@ func TestDocUpdateCorruptSequence(t *testing.T) {
 		return db.DbStats.Database().CorruptSequenceCount.Value()
 	}, 1)
 }
+
+func TestPutResurrection(t *testing.T) {
+	base.SetUpTestLogging(t, base.LevelDebug, base.KeyAll)
+
+	db, ctx := setupTestDB(t)
+	defer db.Close(ctx)
+
+	startWarnCount := base.SyncGatewayStats.GlobalStats.ResourceUtilization.WarnCount.Value()
+	collection, ctx := GetSingleDatabaseCollectionWithUser(ctx, t, db)
+
+	const docID = "doc1"
+	_, _, err := collection.Put(ctx, docID, Body{"foo": "bar"})
+	require.NoError(t, err)
+
+	require.NoError(t, collection.Purge(ctx, docID, false))
+	// assert no warnings when re-pushing a resurrection
+	_, _, err = collection.Put(ctx, docID, Body{"resurrect": true})
+	require.NoError(t, err)
+
+	require.Equal(t, startWarnCount, base.SyncGatewayStats.GlobalStats.ResourceUtilization.WarnCount.Value())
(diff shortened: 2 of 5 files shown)


bbrks (commented):
I think I'm OK with the logging change, aside from the potentially non cherry-pickable stuff - but I'm still not clear _why_ we see this warning in this case.

Is it that the purged doc body is empty? What codepath is being exercised that results in that doc body attempting to be read? That doesn't seem expected to me.

torcolvin:
> I think I'm OK with the logging change, aside from the potentially non cherry-pickable stuff - but I'm still not clear _why_ we see this warning in this case.
> 
> Is it that the purged doc body is empty? What codepath is being exercised that results in that doc body attempting to be read? That doesn't seem expected to me.

There are actually a few cases where this is called. I'll cover the first case in `TestPushResurrection`. 

- `DatabaseCollectionWithUser.Put(ctx, docid, Body{"resurrect": true})` https://github.com/couchbase/sync_gateway/blob/7d271dbd43f2c9078d3504228fa862bd37e7e8a6/db/crud.go#L852
- `DatabaseCollectionWithUser.updateAndReturnDoc(...)` where `existingDoc=nil` https://github.com/couchbase/sync_gateway/blob/7d271dbd43f2c9078d3504228fa862bd37e7e8a6/db/crud.go#L891
    - `WriteUpdateWithXattrs()` is called https://github.com/couchbase/sync_gateway/blob/7d271dbd43f2c9078d3504228fa862bd37e7e8a6/db/crud.go#L2014 with an inline callback. 
    - `WriteUpdateWithXattrs` calls `subdocGetBodyAndXattrs` https://github.com/couchbase/sync_gateway/blob/7d271dbd43f2c9078d3504228fa862bd37e7e8a6/base/collection_xattr_common.go#L207
        - `subdocGetBodyAndXattrs` will return nil body, `ErrNotFound` but a non zero cas. This seems weird on the face, but a cas safety can be required in general for `WriteUpdateWithXattrs` https://github.com/couchbase/sync_gateway/blob/7d271dbd43f2c9078d3504228fa862bd37e7e8a6/base/collection_xattr.go#L251. 
            - if tombstone -> non tombstone, cas value is ignored so 0 or non zero cas would be okay (this is our case and always the cas in `Put`)
            - if tombstone -> tombstone, cas check is required (this is the case if say CBL pushed a deleted document (rev 1-abc) on top of a resurrection)
    - The `WriteUpdateWithXattrs` callback function is invoked with currentValue = nil, currentXattrs = nil, cas = nil
        - doc, err = db.unmarshalDocumentWithXattrs will return `Document{cas=nonzero}` with all other fields as default. In the case that there was not a tombstone, this would return `Document` with fields as default. https://github.com/couchbase/sync_gateway/blob/7d271dbd43f2c9078d3504228fa862bd37e7e8a6/db/crud.go#L2016C7-L2016C48
    - ... `updateAndReturnDoc` calls `documentUpdateFunc` which will ultimately invoke the callback back in `Put`  https://github.com/couchbase/sync_gateway/blob/7d271dbd43f2c9078d3504228fa862bd37e7e8a6/db/crud.go#L2034
- `Document.IsSGWrite(rawBody=ressurectionBody)` is called to determine if we need to perform an on demand import https://github.com/couchbase/sync_gateway/blob/7d271dbd43f2c9078d3504228fa862bd37e7e8a6/db/crud.go#L897
    -  Then reaches this call to `BodyBytes` because it passed the above check of `doc.Cas != doc.SyncData.GetSyncCas()`, since `doc.Cas == nonzero` and `doc.SyncData.GetSyncCas()` == 0  https://github.com/couchbase/sync_gateway/blob/7d271dbd43f2c9078d3504228fa862bd37e7e8a6/db/document.go#L576        
    
It would be possible, but riskier to no-op through `IsSGWrite` call in the special case where the document body = nil. In this case, you have to account for both body and xattr being nil. Since `IsSGWrite` will end up returning `false` in this case, we go on to call to actually try to import the document (import will end up being cancelled). This is not optimal, but this change would be much more drastic.

In the case of a blip push, this can use `DatabaseCollection.PutExistingRevWithConflictResolution`. Given this, it might actually make sense to drop the tests in the rest package.

I thought there was another pathway to hit this error through blip code which undergoes special validation to check for properties in a document like `_attachments` and `_expiry` before deciding to export, which I spotted looking at these failure messages in existing unit tests but I can't find it right now.

I did explore having `BodyBytes` and `Body` return a specific error like `ErrNilBody` but then when I looked at lot of the cases most the cases it would make sense to ignore the nil body anyway even if it is probably very unlikely or impossible to be hit. A non exhaustive list of examples:

- https://github.com/couchbase/sync_gateway/blob/7d271dbd43f2c9078d3504228fa862bd37e7e8a6/rest/admin_api.go#L1455-L1455
- https://github.com/couchbase/sync_gateway/blob/7d271dbd43f2c9078d3504228fa862bd37e7e8a6/db/document.go#L645-L645
- https://github.com/couchbase/sync_gateway/blob/7d271dbd43f2c9078d3504228fa862bd37e7e8a6/db/attachment.go#L186

bbrks:
Ok, I think the `Document{cas=nonzero}` explanation fills in a large understanding gap I had and it definitely explains why we see this with compact as per original Jira ticket. Thanks for explanation

bbrks (approved):

PR #7539, opened 2025-05-22
CBG-4658: Uptake Otto stack depth limit with tests

bbrks:
CBG-4658

Prevents SO in the event somebody has screwed up their sync/import function, or is just doing too much recursion.
Limit is set to 10,000 JS stack frames, which aligns with the max Go stdlib JSON nesting depth (golang/go#31789) in case there are legitimate recursion use-cases.

Added unit test and more 'real-world' import feed sync function test:
```
2025-05-22T11:17:33.283+01:00 [DBG] CRUD+: t:TestImportFeedWithRecursiveSyncFunction db:db col:sg_test_0 Invoking sync on doc "<ud>TestImportFeedWithRecursiveSyncFunction</ud>" rev 1-2b2a84503be944f6ba87b49256345b6f
2025-05-22T11:17:33.333+01:00 [WRN] t:TestImportFeedWithRecursiveSyncFunction db:db col:sg_test_0 Sync fn exception: RangeError: Maximum call stack size exceeded; doc "<ud>TestImportFeedWithRecursiveSyncFunction</ud>" / "1-2b2a84503be944f6ba87b49256345b6f" -- db.(*DatabaseCollectionWithUser).getChannelsAndAccess() at crud.go:2522
2025-05-22T11:17:33.333+01:00 [DBG] CRUD+: t:TestImportFeedWithRecursiveSyncFunction db:db col:sg_test_0 Did not update document "<ud>TestImportFeedWithRecursiveSyncFunction</ud>" w/ xattr: 500 Exception in JS sync function
2025-05-22T11:17:33.333+01:00 [INF] Import: t:TestImportFeedWithRecursiveSyncFunction db:db col:sg_test_0 Error importing doc "<ud>TestImportFeedWithRecursiveSyncFunction</ud>": 500 Exception in JS sync function
2025-05-22T11:17:33.333+01:00 [DBG] Import+: t:TestImportFeedWithRecursiveSyncFunction db:db col:sg_test_0 Did not import doc "<ud>TestImportFeedWithRecursiveSyncFunction</ud>" - external update will not be accessible via Sync Gateway.  Reason: 500 Exception in JS sync function
```

## Dependencies
- [x] https://github.com/couchbase/sg-bucket/pull/136

- n/a

### DIFF
--- a/channels/main_test.go
+++ b/channels/main_test.go
@@ -24,7 +24,7 @@ func TestMain(m *testing.M) {
 	teardownFuncs := make([]func(), 0)
 	teardownFuncs = append(teardownFuncs, base.SetUpGlobalTestLogging(ctx))
 	teardownFuncs = append(teardownFuncs, base.SetUpGlobalTestProfiling(m))
-	teardownFuncs = append(teardownFuncs, base.SetUpGlobalTestMemoryWatermark(m, 128))
+	teardownFuncs = append(teardownFuncs, base.SetUpGlobalTestMemoryWatermark(m, 256))
 
 	base.SkipPrometheusStatsRegistration = true
 
--- a/rest/importtest/import_test.go
+++ b/rest/importtest/import_test.go
@@ -74,32 +74,24 @@ func TestImportFeed(t *testing.T) {
 }
 
 func TestImportFeedWithRecursiveSyncFunction(t *testing.T) {
-
 	base.SkipImportTestsIfNotEnabled(t)
-
-	rtConfig := rest.RestTesterConfig{
-		SyncFn:     `function access(doc) { access("foo", "bar"); }`,
-		AutoImport: base.Ptr(true),
-	}
-
-	rt := rest.NewRestTester(t, &rtConfig)
-	defer rt.Close()
-	dataStore := rt.GetSingleDataStore()
 
 	base.SetUpTestLogging(t, base.LevelDebug, base.KeyImport, base.KeyCRUD)
 
+	rt := rest.NewRestTester(t, &rest.RestTesterConfig{
+		SyncFn:     `function access(doc) { access("foo", "bar"); }`,
+		AutoImport: base.Ptr(true),
+	})
+	defer rt.Close()
+
 	// Create doc via the SDK
 	mobileKey := t.Name()
-	mobileBody := make(map[string]interface{})
-	mobileBody["channels"] = "ABC"
-	_, err := dataStore.Add(mobileKey, 0, mobileBody)
-	assert.NoError(t, err, "Error writing SDK doc")
+	mobileBody := map[string]any{"channels": "ABC"}
+	added, err := rt.GetSingleDataStore().Add(mobileKey, 0, mobileBody)
+	require.NoError(t, err, "Error writing SDK doc")
+	require.True(t, added)
 
-	// Wait for import
-	err = rt.WaitForCondition(func() bool {
-		return rt.GetDatabase().DbStats.SharedBucketImportStats.ImportErrorCount.Value() == 1
-	})
-	require.NoError(t, err)
+	base.RequireWaitForStat(t, rt.GetDatabase().DbStats.SharedBucketImportStats.ImportErrorCount.Value, 1)
 
 	// Attempt to get the document via Sync Gateway.
 	response := rt.SendAdminRequest("GET", "/{{.keyspace}}/"+mobileKey, "")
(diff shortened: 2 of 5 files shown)


channels/main_test.go
```diff
@@ -24,7 +24,7 @@ func TestMain(m *testing.M) {
 	teardownFuncs := make([]func(), 0)
 	teardownFuncs = append(teardownFuncs, base.SetUpGlobalTestLogging(ctx))
 	teardownFuncs = append(teardownFuncs, base.SetUpGlobalTestProfiling(m))
-	teardownFuncs = append(teardownFuncs, base.SetUpGlobalTestMemoryWatermark(m, 128))
+	teardownFuncs = append(teardownFuncs, base.SetUpGlobalTestMemoryWatermark(m, 256))
```
bbrks, line 27: The new `TestSyncFunctionRecursion` pushes just over 128MB but 256 is still a reasonable limit IMO

rest/importtest/import_test.go
```diff
@@ -73,6 +73,42 @@ func TestImportFeed(t *testing.T) {
...
+
+	// Wait for import
+	err = rt.WaitForCondition(func() bool {
+		return rt.GetDatabase().DbStats.SharedBucketImportStats.ImportErrorCount.Value() == 1
+	})
+	require.NoError(t, err)
+
+	// Attempt to get the document via Sync Gateway.
+	response := rt.SendAdminRequest("GET", "/{{.keyspace}}/"+mobileKey, "")
+	assert.Equal(t, 404, response.Code)
```
torcolvin: I know this code is copypasta from another test but I think writing it in a different way would be shorter and also fails on debug. `Add` will return no error if the document already exists, so to use `Add` robustly, you have to check the `added=true` return value. You can use `Set` to make sure the document exists. Using `base.RequireWaitForStat` is type safe and we can change to do better error reporting eventually.

```suggestion
	mobileKey := t.Name()
	require.NoError(t, dataStore.Set(mobileKey, 0, nil, db.Body{"channels": "ABC"}))

	// Wait for import
	base.RequireWaitForStat(t, rt.GetDatabase().DbStats.SharedBucketImportStats.ImportCount.Value, 1)

	// Attempt to get the document via Sync Gateway.
	response := rt.SendAdminRequest("GET", "/{{.keyspace}}/"+mobileKey, "")
	rest.RequireStatus(t, response, http.StatusOK)
```

rest/importtest/import_test.go
```diff
@@ -73,6 +73,42 @@ func TestImportFeed(t *testing.T) {
...
+
+	// Verify this didn't trigger an on-demand import
+	assert.Equal(t, int64(0), rt.GetDatabase().DbStats.SharedBucketImportStats.ImportCount.Value())
```
torcolvin, line 101: Add assertion:

base.RequireWaitForStat(t, rt.GetDatabase().DbStats.SharedBucketImportStats.ImportCount.Value, 2)

torcolvin (commented):
I didn't look exhaustively for this but I was curious to make sure that an error from import would move forward the checkpoints for this document.

The answer to this is that we'll always move the checkpoints forward if we can because https://github.com/couchbase/sync_gateway/blob/7d271dbd43f2c9078d3504228fa862bd37e7e8a6/db/import_listener.go#L167 return true for all cases.

I think this is well outside the scope of this ticket, but it's interesting that an error like this would require a resync to fix. But other types of transient errors could also require a resync, like an temporary kv error where processing the document failed but then there was a recovery later and the checkpoint moved forward.

rest/importtest/import_test.go
bbrks: Why would I have two imports here? I have zero.

Making this a WaitForStat will make it take longer to fail under the failure case when the expected value is `0`

rest/importtest/import_test.go
bbrks: Will use wait for stat helpers, but I don't agree with using Set to hide the fact that doc may exist in bucket. Is it not a good assumption that a test bucket is empty? Why not?

bbrks:
> I think this is well outside the scope of this ticket, but it's interesting that an error like this would require a resync to fix. But other types of transient errors could also require a resync, like an temporary kv error where processing the document failed but then there was a recovery later and the checkpoint moved forward.

The expectation I believe is that we can advance the checkpoint forwards because we can always rely on on-demand imports to avoid running resync for a semi-automated recovery.

rest/importtest/import_test.go
torcolvin: I meant:

assert.Equal(t, rt.GetDatabase().DbStats.SharedBucketImportStats.ImportErrorCount.Value(), 2)

bbrks: Sure.

torcolvin (approved):

PR #7540, opened 2025-05-23
CBG-4617: Add DisablePublicAllDocs db config option

bbrks:
CBG-4617

- Add `DisablePublicAllDocs` db config option (defaults to `false`)
- Add deprecation warning when this option evaluates to `false`
- Add test
  - Note: This test was originally AI generated but subsequently cleaned up and modified

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3123/

### DIFF
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -1915,6 +1915,14 @@ Database:
                   name:
                     description: The name of the role for which audit logging is disabled.
                     type: string
+    disable_public_all_docs:
+      description: |-
+        This controls whether the `_all_docs` REST API endpoint is publicly accessible or not.
+
+        If set to `true`, the `_all_docs` endpoint will not be publicly accessible, and will only be available on the Admin API.
+        Setting this to `false`, or leaving it as the default value is deprecated, and will default to `true` in a future release.
+      type: boolean
+      default: false
   title: Database-config
 Disabled-users-and-roles:
   type: object
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -1917,10 +1917,12 @@ Database:
                     type: string
     disable_public_all_docs:
       description: |-
-        This controls whether the `_all_docs` REST API endpoint is publicly accessible or not.
+        This controls whether the [GET /{keyspace}/_all_docs](#operation/get_keyspace-_all_docs) REST API endpoint is publicly accessible or not.
+        Disabling this endpoint is recommended for larger datasets or production workloads.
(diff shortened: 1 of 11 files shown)


docs/api/components/schemas.yaml
```diff
@@ -1915,6 +1915,14 @@ Database:
                   name:
                     description: The name of the role for which audit logging is disabled.
                     type: string
+    disable_public_all_docs:
+      description: |-
+        This controls whether the `_all_docs` REST API endpoint is publicly accessible or not.
+
+        If set to `true`, the `_all_docs` endpoint will not be publicly accessible, and will only be available on the Admin API.
+        Setting this to `false`, or leaving it as the default value is deprecated, and will default to `true` in a future release.
```
torcolvin: I suggest adding some information about why you might want to do this:

This endpoint is present for CouchDB compatibility, but has serious performance implication on GSI nodes and memory of Sync Gateway when this is called. As alternatives to get the documents, use /ks/_changes to get all docs, or /ks/_bulk_get if you know the specific document ids.

docs/api/components/schemas.yaml
```diff
@@ -1915,6 +1915,14 @@ Database:
                   name:
                     description: The name of the role for which audit logging is disabled.
                     type: string
+    disable_public_all_docs:
+      description: |-
+        This controls whether the `_all_docs` REST API endpoint is publicly accessible or not.
```
torcolvin: Nice to have but maybe add links:

[_all_docs](#operation/get_keyspace-_all_docs)

This link will resolve within the admin endpoint, but I think it is still helpful to see what the link does.

docs/api/paths/public/keyspace-_all_docs.yaml
```diff
@@ -9,7 +9,10 @@ parameters:
   - $ref: ../../components/parameters.yaml#/keyspace
 get:
   summary: Gets all the documents in the database with the given parameters
-  description: Returns all documents in the database based on the specified parameters.
+  description: |-
+    Returns all documents in the database based on the specified parameters.
```
torcolvin, line 13: I think we should add the caveats and warnings about calling this endpoint here.

This document is also in `paths/admin/keyspace-all_docs.yaml`. Before we added some extra documentation I would have suggested deduplicating these yaml files, but we can leave them.

bbrks: I can reword the recommendation to be applicable to both admin and public but can't de-duplicate if we're specifying the 403 response. The info is going to be the same either way.

torcolvin (approved):

PR #7548, opened 2025-05-29
DOC-12510 add rules to enforce additional typechecking

torcolvin:
This was inspired by https://github.com/couchbase/docs-site/pull/823 and I briefly thought that we would need to specify `type: object` on any schema that didn't have one. It doesn't seem like openapi-spec-validator, swagger, or redocly force this, so this assumption was baked into the that PR.

However, when trying to look at this, `uvx openapi-spec-validator docs/api/*yaml` did find issues with defaults having mismatched types and default values being outside the minimum and maximum range. I ended up writing one new redocly linter than catches defaults and type missing. I looked for a redocly linter to do this and I couldn't find one! This seems like something that people would have and would exist in the default (or extensible) linters. It seems like it would be good if someone could double check I'm not missing something

Additionally, I added a linter that checks `additionalProperties` for either an example of `x-additionalPropertiesName` names since they show up poorly in both redocly and in our json database parsing, see https://docs.couchbase.com/sync-gateway/current/configuration-schema-database.html#event_handlers-db_state_changed-options-{additionalProperties}


docs/api/plugins/rules/typecheck-defaults.js
```diff
@@ -0,0 +1,125 @@
+module.exports = TypeCheckDefaults;
+
+function checkMinMaxInteger(schema, ctx) {
```
osfameron, line 13: nice sense-check

docs/api/paths/public/keyspace-_changes.yaml
```diff
@@ -67,11 +67,10 @@ get:
           type: string
     - name: heartbeat
       in: query
-      description: The interval (in milliseconds) to send an empty line (CRLF) in the response. This is to help prevent gateways from deciding the socket is idle and therefore closing it. This is only applicable to `feed=longpoll` or `feed=continuous`. This will override any timeouts to keep the feed alive indefinitely. Setting to 0 results in no heartbeat. The maximum heartbeat can be set in the server replication configuration.
+      description: The interval (in milliseconds) to send an empty line (CRLF) in the response. This is to help prevent gateways from deciding the socket is idle and therefore closing it. This is only applicable to `feed=longpoll` or `feed=continuous`. This will override any timeouts to keep the feed alive indefinitely. Setting to 0 results in no heartbeat. The maximum heartbeat can be set in the server replication configuration. If heartbeat is non zero, it must be at least 25000 milliseconds.
       schema:
         type: integer
         default: 0
-        minimum: 25000
```
osfameron, line 74: ah, this was removed because of the sensecheck?
hahaha, that's a weird case

osfameron (commented):
Looks good to me - I don't think any of those things are *necessary* on our side, but consistency is good, and the extra rules look useful 👍 

(Not Approving as I'm not a code-owner anyway)

torcolvin:
From looking at https://docs.couchbase.com/sync-gateway/current/configuration-schema-database.html#event_handlers-db_state_changed-options-{additionalProperties} - I wanted to remove this `additionalProperties` by name. I know event handlers are deprecated but I added an additional linter to find these issues (and fixed them in the openapi specs).

bbrks (approved):

PR #7550, opened 2025-05-30
CBG-4629 add documentation for changes_request_plus

torcolvin:
CBG-4629 add documentation for changes_request_plus

### DIFF
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -1827,7 +1827,7 @@ Database:
       type: boolean
       default: false
       description: |-
-        When true, ensures all valid documents written prior to the request being issued are included in the response. This is only applicable for non-continuous feeds. Setting this value at a database level is required for Couchbase Lite replications.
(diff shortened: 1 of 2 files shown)


docs/api/components/schemas.yaml
```diff
@@ -1823,6 +1823,13 @@ Database:
         Defaults to true when running in serverless mode otherwise defaults to false.
       type: boolean
       default: false
+    changes_request_plus:
+      type: boolean
+      default: false
+      description: |-
+        When true, ensures all valid documents written prior to the request being issued are included in the response. This is only applicable for non-continuous feeds. Setting this value at a database level is required for Couchbase Lite replications.
```
bbrks: > Setting this value at a database level is required for Couchbase Lite replications.

This makes it sound like users _must_ enable this config option to use Couchbase Lite, which is not the case.
Also this db config option just sets the default value of `request_plus` for changes feeds, rather than enforcing it.

I would suggest rephrasing to:

```suggestion
        Sets the default value of `request_plus` for one-shot/non-continuous changes feeds, which when true, ensures all valid documents written prior to the request being issued are included in the response. Setting this option at the database level is required to ensure Couchbase Lite utilizes this changes feed mode.
```

bbrks (changes requested):

bbrks (approved):

PR #7551, opened 2025-06-02
CBG-603: have broadcast goroutine with ticker value to broadcast changes

gregns1:
CBG-603

- Have `BroadcastChanges` function that spawns a goroutine to call Broadcast if the change listener count changes
- Adds a few tests to assert the boolean is correctly changed when skipped sequences are present and replication works as expected 
- Use pushing and removing from skipped to return whether the skipped list is empty or not, this will avoid any extra contention on skipped list mutex 
- Note: new skipped list data structure to go in soon so code to return whether or not skipped is empty will change very soon

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3128/

### DIFF
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -589,9 +589,10 @@ func (c *changeCache) processUnusedRange(ctx context.Context, fromSequence, toSe
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
+	var isEmpty bool
 	if toSequence < c.nextSequence {
 		// batch remove from skipped
-		c.skippedSeqs.processUnusedSequenceRangeAtSkipped(ctx, fromSequence, toSequence)
+		isEmpty = c.skippedSeqs.processUnusedSequenceRangeAtSkipped(ctx, fromSequence, toSequence)
 	} else if fromSequence >= c.nextSequence {
 		// whole range to pending
 		c._pushRangeToPending(fromSequence, toSequence, timeReceived)
@@ -607,6 +608,11 @@ func (c *changeCache) processUnusedRange(ctx context.Context, fromSequence, toSe
 		// a duplicate entry with a sequence within the bounds of the range was previously present
 		// in pending.
 		base.WarnfCtx(ctx, "unused sequence range of #%d to %d contains duplicate sequences, will be ignored", fromSequence, toSequence)
+	}
+	if isEmpty {
+		if c.db.mutationListener.SkippedSequenceBroadcast.Load() {
+			c.db.mutationListener.SkippedSequenceBroadcast.Store(false)
+		}
 	}
 	return allChangedChannels
 }
@@ -941,7 +947,12 @@ func (h *LogPriorityQueue) Pop() interface{} {
 // ////// SKIPPED SEQUENCE QUEUE
 
 func (c *changeCache) RemoveSkipped(x uint64) error {
-	err := c.skippedSeqs.removeSeq(x)
+	isEmpty, err := c.skippedSeqs.removeSeq(x)
+	if isEmpty {
+		if c.db.mutationListener.SkippedSequenceBroadcast.Load() {
+			c.db.mutationListener.SkippedSequenceBroadcast.Store(false)
+		}
+	}
 	return err
 }
 
@@ -955,6 +966,9 @@ func (c *changeCache) PushSkipped(ctx context.Context, startSeq uint64, endSeq u
 		return
 	}
 	c.skippedSeqs.PushSkippedSequenceEntry(NewSkippedSequenceRangeEntry(startSeq, endSeq))
+	if !c.db.mutationListener.SkippedSequenceBroadcast.Load() {
+		c.db.mutationListener.SkippedSequenceBroadcast.Store(true)
+	}
 }
 
 // waitForSequence blocks up to maxWaitTime until the given sequence has been received.
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -946,10 +946,13 @@ func (h *LogPriorityQueue) Pop() interface{} {
 
 func (c *changeCache) RemoveSkipped(x uint64) error {
 	numSkipped, err := c.skippedSeqs.removeSeq(x)
+	if err != nil {
+		return err
+	}
 	if numSkipped == 0 {
 		c.db.mutationListener.BroadcastSlowMode.CompareAndSwap(true, false)
 	}
-	return err
+	return nil
 }
 
 func (c *changeCache) WasSkipped(x uint64) bool {
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -589,10 +589,10 @@ func (c *changeCache) processUnusedRange(ctx context.Context, fromSequence, toSe
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
-	var isEmpty bool
+	var numSkipped int64
 	if toSequence < c.nextSequence {
 		// batch remove from skipped
-		isEmpty = c.skippedSeqs.processUnusedSequenceRangeAtSkipped(ctx, fromSequence, toSequence)
+		numSkipped = c.skippedSeqs.processUnusedSequenceRangeAtSkipped(ctx, fromSequence, toSequence)
 	} else if fromSequence >= c.nextSequence {
 		// whole range to pending
 		c._pushRangeToPending(fromSequence, toSequence, timeReceived)
@@ -609,10 +609,8 @@ func (c *changeCache) processUnusedRange(ctx context.Context, fromSequence, toSe
 		// in pending.
 		base.WarnfCtx(ctx, "unused sequence range of #%d to %d contains duplicate sequences, will be ignored", fromSequence, toSequence)
 	}
-	if isEmpty {
-		if c.db.mutationListener.SkippedSequenceBroadcast.Load() {
-			c.db.mutationListener.SkippedSequenceBroadcast.Store(false)
-		}
+	if numSkipped == 0 {
+		c.db.mutationListener.SkippedSequenceBroadcast.CompareAndSwap(true, false)
 	}
 	return allChangedChannels
 }
@@ -947,11 +945,9 @@ func (h *LogPriorityQueue) Pop() interface{} {
 // ////// SKIPPED SEQUENCE QUEUE
 
 func (c *changeCache) RemoveSkipped(x uint64) error {
-	isEmpty, err := c.skippedSeqs.removeSeq(x)
-	if isEmpty {
-		if c.db.mutationListener.SkippedSequenceBroadcast.Load() {
-			c.db.mutationListener.SkippedSequenceBroadcast.Store(false)
-		}
+	numSkipped, err := c.skippedSeqs.removeSeq(x)
+	if numSkipped == 0 {
+		c.db.mutationListener.SkippedSequenceBroadcast.CompareAndSwap(true, false)
 	}
 	return err
 }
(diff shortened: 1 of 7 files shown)


db/change_cache.go
```diff
@@ -608,6 +609,11 @@ func (c *changeCache) processUnusedRange(ctx context.Context, fromSequence, toSe
 		// in pending.
 		base.WarnfCtx(ctx, "unused sequence range of #%d to %d contains duplicate sequences, will be ignored", fromSequence, toSequence)
 	}
+	if isEmpty {
+		if c.db.mutationListener.SkippedSequenceBroadcast.Load() {
+			c.db.mutationListener.SkippedSequenceBroadcast.Store(false)
+		}
```
bbrks: Can use `CompareAndSwap` here, since a separate atomic `Load` and `Store` will still allow for a race between the two lines from another goroutine.

CAS is implemented in one CPU instruction, so it's no more expensive than a `Load()` would be, even when successful.

```suggestion
		_ = c.db.mutationListener.SkippedSequenceBroadcast.CompareAndSwap(true, false)
```

db/change_cache.go
```diff
@@ -941,7 +947,12 @@ func (h *LogPriorityQueue) Pop() interface{} {
 // ////// SKIPPED SEQUENCE QUEUE
 
 func (c *changeCache) RemoveSkipped(x uint64) error {
-	err := c.skippedSeqs.removeSeq(x)
+	isEmpty, err := c.skippedSeqs.removeSeq(x)
+	if isEmpty {
+		if c.db.mutationListener.SkippedSequenceBroadcast.Load() {
+			c.db.mutationListener.SkippedSequenceBroadcast.Store(false)
+		}
```
```suggestion
		_ = c.db.mutationListener.SkippedSequenceBroadcast.CompareAndSwap(true, false)
```

db/change_cache.go
```diff
@@ -955,6 +966,9 @@ func (c *changeCache) PushSkipped(ctx context.Context, startSeq uint64, endSeq u
 		return
 	}
 	c.skippedSeqs.PushSkippedSequenceEntry(NewSkippedSequenceRangeEntry(startSeq, endSeq))
+	if !c.db.mutationListener.SkippedSequenceBroadcast.Load() {
+		c.db.mutationListener.SkippedSequenceBroadcast.Store(true)
+	}
```
```suggestion
	_ = c.db.mutationListener.SkippedSequenceBroadcast.CompareAndSwap(false, true)
```

db/change_listener.go
```diff
@@ -212,10 +221,49 @@ func (listener *changeListener) Notify(ctx context.Context, keys channels.Set) {
 	}
 	base.DebugfCtx(ctx, base.KeyChanges, "Notifying that %q changed (keys=%q) count=%d",
```
bbrks: Should this be moved into the ticker as well?

db/skipped_sequence.go
```diff
@@ -215,9 +215,9 @@ func binarySearchFunc(a *SkippedSequenceListEntry, seq uint64) int {
 }
 
 // _removeSeqRange will remove sequence between x and y from the skipped sequence slice
-func (s *SkippedSequenceSlice) _removeSeqRange(ctx context.Context, startSeq, endSeq uint64) error {
+func (s *SkippedSequenceSlice) _removeSeqRange(ctx context.Context, startSeq, endSeq uint64) (bool, error) {
```
bbrks: These method signatures are confusing without names - normally `bool, error` or `something, bool` indicates whether an operation was successful or not (without an error)

I would be tempted to go as far as returning the number of skipped sequences, instead of just a bool, and make the caller turn that into what they want using `length == 0`

It's a better API if you think about testing the result of these operations, since you can do assertions on the length result.

db/change_listener.go
```diff
@@ -212,10 +221,49 @@ func (listener *changeListener) Notify(ctx context.Context, keys channels.Set) {
...
 		base.MD(listener.bucketName), base.UD(keys), listener.counter)
-	listener.tapNotifier.Broadcast()
 	listener.tapNotifier.L.Unlock()
 }
 
+func (listener *changeListener) BroadcastChanges(ctx context.Context) {
```
bbrks: Rename to make this more obvious that it is starting a goroutine/process

```suggestion
func (listener *changeListener) StartNotifierBroadcaster(ctx context.Context) {
```

db/change_listener.go
```diff
@@ -24,23 +25,29 @@ import (
...
+	tapNotifier              *sync.Cond             // Posts notifications when documents are updated
+	FeedArgs                 sgbucket.FeedArguments // The Tap Args (backfill, etc)
+	counter                  uint64                 // Event counter; increments on every doc update
+	_terminateCheckCounter   uint64                 // Termination Event counter; increments on every notifyCheckForTermination
+	keyCounts                map[string]uint64      // Latest count at which each doc key was updated
+	OnChangeCallback         DocChangedFunc
+	terminator               chan bool          // Signal to cause DCP feed to exit
+	sgCfgPrefix              string             // SG config key prefix
+	started                  base.AtomicBool    // whether the feed has been started
+	SkippedSequenceBroadcast atomic.Bool        // bool to indicate if a skipped sequence ticker value should be used to notify changes feeds of changes
```
bbrks: I had wondered about making this an atomic that directly controls the desired ticker frequency is a better approach than a slow/fast bool - but it does mean that you can't do CAS when setting the value, which means you'll be calling `Set()` over repeatedly, which I think is less desirable than CAS.

db/change_listener.go
```diff
@@ -212,10 +221,49 @@ func (listener *changeListener) Notify(ctx context.Context, keys channels.Set) {
...
+					// change the ticker from default value to use skipped sequence value
+					base.DebugfCtx(ctx, base.KeyChanges, "Changing broadcast changes interval to the skipped sequence value for %q", base.MD(listener.bucketName))
+					ticker.Reset(SkippedSequenceBroadcastChangesTime)
+					defaultTickerValue = false // update local bool to indicate ticker value is no longer using the default value
+				} else if !shouldUseSkippedValue && !defaultTickerValue {
+					// change the ticker from skipped sequence value to default value
+					base.DebugfCtx(ctx, base.KeyChanges, "Changing broadcast changes interval to the default value for %q", base.MD(listener.bucketName))
+					ticker.Reset(DefaultBroadcastChangesTime)
+					defaultTickerValue = true // update local bool to indicate ticker value is using the default value
+				}
```
bbrks: If we switch both bools to mean the same thing (e.g. `slowMode`), we can simplify the logic when changing to something like this, because anything non-matching means it needs to change:

where `tickerValForBroadcastSpeed` is just a function that returns a new ticker duration for the given `true`/`false` value of `slowMode`

```suggestion
				// change the ticker duration if required
				newSlowMode := listener.broadcastSlowMode.Load()
				if newSlowMode != broadcastSlowMode {
					newDuration := tickerValForBroadcastSpeed(slowMode)
					base.DebugfCtx(ctx, base.KeyChanges, "Updating broadcast changes interval for %q to %v", base.MD(listener.bucketName), newDuration)
					broadcastSlowMode = newSlowMode
					ticker.Reset(newDuration)
				}
```

bbrks (changes requested):

db/change_cache.go
```diff
@@ -589,9 +589,10 @@ func (c *changeCache) processUnusedRange(ctx context.Context, fromSequence, toSe
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
+	var isEmpty bool
```
torcolvin: i would rename this var removedSkippedSeqs to represent what this means better.

db/skipped_sequence.go
```diff
@@ -226,78 +226,82 @@ func (s *SkippedSequenceSlice) _removeSeqRange(ctx context.Context, startSeq, en
...
 
 // removeSeq will remove a given sequence from the slice if it exists
-func (s *SkippedSequenceSlice) removeSeq(x uint64) error {
+func (s *SkippedSequenceSlice) removeSeq(x uint64) (bool, error) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
 
 	if len(s.list) == 0 {
-		return fmt.Errorf("skipped sequence list is empty, unable to remove sequence %d", x)
+		return true, fmt.Errorf("skipped sequence list is empty, unable to remove sequence %d", x)
```
torcolvin: Is this intetnional to return true here?

db/change_listener.go
```diff
@@ -212,10 +221,49 @@ func (listener *changeListener) Notify(ctx context.Context, keys channels.Set) {
...
-	listener.tapNotifier.Broadcast()
 	listener.tapNotifier.L.Unlock()
 }
 
+func (listener *changeListener) BroadcastChanges(ctx context.Context) {
+	ticker := time.NewTicker(DefaultBroadcastChangesTime)
+	// boolean to indicate whether ticker is using the default value, this is needed so we don't call reset on ticker
+	// for a value it already has
+	defaultTickerValue := true
+	terminator := listener.terminator // take copy of this to avoid race when restarting change listener in tests
```
torcolvin: I don't think this makes a copy, this is a channel which is implemented as a pointer

db/change_listener.go
gregns1: I'm unsure. I chose to leave as is in first iteration of this PR but i'm open to hear you opinion. It felt like it was right to leave it as given that even though the function doesn't broadcast those changes any more, it still increments the keys that have changed. If I move this into the ticker goroutine we will have to remove the keys from the log message which felt like valid info that could be being removed? 

Maybe the best option is to change this log line to something like "keys %q have changed" then add a new log line inside ticker to show that a notify is being broadcast?

db/skipped_sequence.go
gregns1: Yeah as the list is empty here

db/change_cache.go
gregns1: This was a variable to indicate if the skipped list is empty or not after removal so I don't think that makes sense. 

This has now been changed to return actual count not boolean based on review above

db/change_listener.go
gregns1: I'll remove the comment, I don't know how better to word it.

db/change_listener.go
bbrks: Yes I think the reword and additional log makes sense - we want to know when its broadcast, but I think we do also want to know when each change was bumping the waiter count

db/change_listener.go
bbrks: Might be clearer if this is passed into the goroutine function instead to capture it - it's functionally equivalent to copying the pointer address.

```go
go func(terminator chan bool) {
	// ...
}(listener.terminator)
```

db/change_cache.go
```diff
@@ -941,7 +945,10 @@ func (h *LogPriorityQueue) Pop() interface{} {
 // ////// SKIPPED SEQUENCE QUEUE
 
 func (c *changeCache) RemoveSkipped(x uint64) error {
-	err := c.skippedSeqs.removeSeq(x)
+	numSkipped, err := c.skippedSeqs.removeSeq(x)
+	if numSkipped == 0 {
+		c.db.mutationListener.BroadcastSlowMode.CompareAndSwap(true, false)
+	}
 	return err
```
bbrks: This is probably worth an error check now? Although the only cases we really care about works with the code as-is whether there's an error check or not (empty list or not found sequence)

I'm thinking if we were to add an error case we'd want to leave the state of `BroadcastSlowMode` alone and not switch back to fast mode, given `return 0, err` is a common enough pattern I can see that being written instead of returning the current skipped seq length.

```suggestion
	if err != nil {
		return err
	}
	if numSkipped == 0 {
		c.db.mutationListener.BroadcastSlowMode.CompareAndSwap(true, false)
	}
	return nil
```

bbrks (changes requested):

db/change_cache.go
gregns1: Good point, will adjust this.

bbrks (approved):

PR #7552, opened 2025-06-02
CBG-4661: Pass doc type down from ProcessFeedEvent to avoid extra key filtering

gregns1:
CBG-4661

- Passing a doc type parameter down stack into `DocChanged` from the filtering we do in `PorocessFeedEvent` to avoid the same key filtering being done twice

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3129/
- [x] https://jenkins.sgwdev.com/job/SyncGateway-Integration/3132/
  - unrelated flakes

### DIFF
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -40,12 +40,13 @@ const (
 type DocumentType uint8
 
 const (
-	DocTypeDocument       DocumentType = iota // Customer data document type
+	DocTypeUnknown        DocumentType = iota // Unknown document type
+	DocTypeDocument                           // Customer data document type
 	DocTypeUser                               // User document
 	DocTypeRole                               // Role document
 	DocTypeUnusedSeq                          // Unused sequence notification
 	DocTypeUnusedSeqRange                     // Unused sequence range notification
-	DocTypeSGConfig                           // Sync Gateway config document
+	DocTypeSGCfg                              // Cfg docs for import feed management
 )
 
 // Enable keeping a channel-log for the "*" channel (channel.UserStarChannel). The only time this channel is needed is if
@@ -318,25 +319,22 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent, docType DocumentType)
 	timeReceived := channels.NewFeedTimestamp(&event.TimeReceived)
 	// ** This method does not directly access any state of c, so it doesn't lock.
 	// Is this a user/role doc for this database?
-	if docType == DocTypeUser {
+	switch docType {
+	case DocTypeUnknown:
+		return // no-op unknown doc type
+	case DocTypeUser:
 		c.processPrincipalDoc(ctx, docID, docJSON, true, timeReceived)
 		return
-	} else if docType == DocTypeRole {
+	case DocTypeRole:
 		c.processPrincipalDoc(ctx, docID, docJSON, false, timeReceived)
 		return
-	}
-
-	// Is this an unused sequence notification?
-	if docType == DocTypeUnusedSeq {
+	case DocTypeUnusedSeq:
 		c.processUnusedSequence(ctx, docID, timeReceived)
 		return
-	}
-	if docType == DocTypeUnusedSeqRange {
+	case DocTypeUnusedSeqRange:
 		c.processUnusedSequenceRange(ctx, docID)
 		return
-	}
-
-	if docType == DocTypeSGConfig {
+	case DocTypeSGCfg:
 		if c.cfgEventCallback != nil {
 			c.cfgEventCallback(docID, event.Cas, nil)
 		}
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -36,6 +36,18 @@ const (
 	unusedSeqCollectionID         = 0                // Collection ID used by ChangeWaiter to mark unused sequences
 )
 
+// DocumentType indicates the type of document being processed in the change cache (e.g. User doc etc).
+type DocumentType uint8
+
+const (
+	DocTypeDocument       DocumentType = iota // Customer data document type
+	DocTypeUser                               // User document
+	DocTypeRole                               // Role document
+	DocTypeUnusedSeq                          // Unused sequence notification
+	DocTypeUnusedSeqRange                     // Unused sequence range notification
+	DocTypeSGConfig                           // Sync Gateway config document
+)
+
 // Enable keeping a channel-log for the "*" channel (channel.UserStarChannel). The only time this channel is needed is if
 // someone has access to "*" (e.g. admin-party) and tracks its changes feed.
 var EnableStarChannelLog = true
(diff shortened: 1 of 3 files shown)


db/change_listener.go
```diff
@@ -128,17 +128,24 @@ func (listener *changeListener) ProcessFeedEvent(event sgbucket.FeedEvent) bool
 	if event.Opcode == sgbucket.FeedOpMutation || event.Opcode == sgbucket.FeedOpDeletion {
```
bbrks, line 141: This function is maybe a bit unwieldy now.

I think it would benefit a lot, even for just readability, but potentially for performance, from early returns for at least the most common cases:
- mutation or deletion for non-sg docs (prioritising this in evaluation)
- backfill opcodes (return true)
- dcp checkpoint case (return false)

And then if we think the if/else is still a bit too much, could go further. I'm thinking of something like this, but this might be too much refactoring, will let you decide:

```go
func (listener *changeListener) DocumentType(key []byte) DocumentType {
	if bytes.HasPrefix(key, []byte(listener.metaKeys.UserKeyPrefix())) {
		return DocTypeUser
	} else if bytes.HasPrefix(key, []byte(listener.metaKeys.RoleKeyPrefix())) {
		return DocTypeRole
	} else if bytes.HasPrefix(key, []byte(listener.metaKeys.UnusedSeqPrefix())) {
		return DocTypeUnusedSeq
	} else if bytes.HasPrefix(key, []byte(listener.metaKeys.UnusedSeqRangePrefix())) {
		return DocTypeUnusedSeqRange
	}
	return DocTypeUnknown
}

// ProcessFeedEvent is invoked for each mutate or delete event seen on the server's mutation feed (TAP or DCP).  Uses document
// key to determine handling, based on whether the incoming mutation is an internal Sync Gateway document.
func (listener *changeListener) ProcessFeedEvent(event sgbucket.FeedEvent) (requiresCheckpointPersistence bool) {
	// happy/fast-path: non-metadata doc mutations/deletions
	if event.Opcode == sgbucket.FeedOpMutation || event.Opcode == sgbucket.FeedOpDeletion {
		if !bytes.HasPrefix(event.Key, []byte(base.SyncDocPrefix)) {
			listener.OnDocChanged(event, DocTypeDocument)
			return true
		}
	} else {
		// backfill or unknown opcodes
		return true
	}

	// SG DCP checkpoint docs (including other config group IDs)
	if bytes.HasPrefix(event.Key, []byte(base.DCPCheckpointRootPrefix)) {
		// Do not require checkpoint persistence when DCP checkpoint docs come back over DCP - otherwise
		// we'll end up in a feedback loop for their vbucket if persistence is enabled
		// NOTE: checkpoint persistence is disabled altogether for the caching feed.  Leaving this check in place
		// defensively.
		return false
	}

	// Cfg callback supports both mutation and deletion events
	if !bytes.HasPrefix(event.Key, []byte(listener.sgCfgPrefix)) {
		listener.OnDocChanged(event, DocTypeSGConfig)
		return true
	}

	if event.Opcode != sgbucket.FeedOpMutation {
		// nothing more to handle
		return true
	}

	docType := listener.DocumentType(event.Key)
	if docType == DocTypeUser || docType == DocTypeRole {
		// defer to notify after callback completion
		defer listener.notifyKey(listener.ctx, string(event.Key))
	}

	listener.OnDocChanged(event, docType)
	return true
}
```

db/change_cache.go
```diff
@@ -36,6 +36,18 @@ const (
...
+// DocumentType indicates the type of document being processed in the change cache (e.g. User doc etc).
+type DocumentType uint8
+
+const (
+	DocTypeDocument       DocumentType = iota // Customer data document type
+	DocTypeUser                               // User document
+	DocTypeRole                               // Role document
+	DocTypeUnusedSeq                          // Unused sequence notification
+	DocTypeUnusedSeqRange                     // Unused sequence range notification
+	DocTypeSGConfig                           // Sync Gateway config document
```
bbrks: This is probably slightly better described as a SGCfg document (for a cbgt `Cfg` implementation) - it's not really a SG config.

db/change_cache_test.go
```diff
@@ -1406,7 +1406,7 @@ func TestLateArrivingSequenceTriggersOnChange(t *testing.T) {
 		CollectionID: collectionID,
 		DataType:     dataType,
 	}
-	db.changeCache.DocChanged(feedEventDoc2)
+	db.changeCache.DocChanged(feedEventDoc2, 0)
```
bbrks: Should use the const here instead of `0`, since we know we're writing an actual customer document.


```suggestion
	db.changeCache.DocChanged(feedEventDoc2, DocTypeDocument)
```

db/change_cache_test.go
```diff
@@ -1958,7 +1958,7 @@ func BenchmarkDocChanged(b *testing.B) {
 			b.ResetTimer()
 			for i := 0; i < b.N; i++ {
 				feedEntry := bm.feed.Next()
-				changeCache.DocChanged(feedEntry)
+				changeCache.DocChanged(feedEntry, 0)
```
bbrks:
```suggestion
				changeCache.DocChanged(feedEntry, DocTypeDocument)
```

db/change_cache_test.go
```diff
@@ -1415,7 +1415,7 @@ func TestLateArrivingSequenceTriggersOnChange(t *testing.T) {
 		Value:        doc1DCPBytes,
 		CollectionID: collectionID,
 	}
-	db.changeCache.DocChanged(feedEventDoc1)
+	db.changeCache.DocChanged(feedEventDoc1, 0)
```
bbrks:
```suggestion
	db.changeCache.DocChanged(feedEventDoc1, DocTypeDocument)
```

db/change_cache.go
```diff
@@ -36,6 +36,18 @@ const (
 	unusedSeqCollectionID         = 0                // Collection ID used by ChangeWaiter to mark unused sequences
 )
 
+// DocumentType indicates the type of document being processed in the change cache (e.g. User doc etc).
+type DocumentType uint8
+
+const (
+	DocTypeDocument       DocumentType = iota // Customer data document type
```
```suggestion
	DocTypeUnknown       DocumentType = iota // Unknown document type
	DocTypeCustomerData                      // Customer data document type
```

Maybe this instead? So we can treat the `0` value as unknown for all non-recognized `_sync:` docs, and then have `DocTypeCustomerData` be used for everything that isn't `_sync:` prefixed.

It makes the switch/case suggested easier if there's a doc type that represents this that can no-op.

db/change_cache.go
```diff
@@ -306,25 +318,25 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent) {
 	timeReceived := channels.NewFeedTimestamp(&event.TimeReceived)
 	// ** This method does not directly access any state of c, so it doesn't lock.
 	// Is this a user/role doc for this database?
-	if strings.HasPrefix(docID, c.metaKeys.UserKeyPrefix()) {
+	if docType == DocTypeUser {
```
bbrks: Switch/case is going to be easier to read here (and probably slightly more performant) now that there's `DocumentType` to switch on

bbrks (changes requested):

bbrks (approved):

PR #7553, opened 2025-06-03
CBG-4554 create automaticReconnectOnRollback to force reopening a stream on rollback

torcolvin:
This is option 2 for implementing rollback fixes in `MB-66749`. This is _much_ simpler code than option 1 and the only case I have definitely working.

To run:

`env SG_TEST_BACKING_STORE=couchbase go test -tags cb_sg_enterprise,cb_sg_devmode ./rest/importtest -run TestImportRollbackAll -count 1`

This is intended as proof of concept for cbgt changes. The guts of the change are in https://github.com/couchbase/cbgt/commit/24afb2e9402e9c32fcddbeee0ad60c5959526af4 which I didn't push to gerrit because I wasn't sure how to incorporate that into go.mod

## Pre-review checklist
- [x] Removed debug logging (`fmt.Print`, `log.Print`, ...)
- [x] Logging sensitive data? Make sure it's tagged (e.g. `base.UD(docID)`, `base.MD(dbName)`)
- [x] Updated relevant information in the API specifications (such as endpoint descriptions, schemas, ...) in `docs/api`

## Dependencies (if applicable)
- [x] Link upstream PRs
- [ ] Update Go module dependencies when merged

## [Integration Tests](https://jenkins.sgwdev.com/job/SyncGateway-Integration/build?delay=0sec)
- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/000/

PR #7554, opened 2025-06-03
CBG-4554 just reopen stream on rollback 

torcolvin:
Unlike option 2, this involves a lot of changes to Sync Gateway https://github.com/couchbase/sync_gateway/pull/7553. The primary change involves passing `cbgt.Manager` into each `cbgt.Dest` instance, which feels very clumsy and perhaps like we are having dependencies go the wrong way around in a circular fashion. My understand of the data organization is that this is the class structure.

```mermaid
flowchart TD
      Manager
      Manager --> Feed
      Feed --> Dest1
      Feed --> Dest2
```      

Particularly fragile is the logic in `DCPCommon.rollbackEx` where `c.cbgtManager.CurrentMaps()` is called to find the right dest. This passes in my test but this feels very not robust. I think I'd have to consider what the right behavior is if we don't find the right feed or dest, which is probably to try to restart the feed?

This depends on https://github.com/couchbase/cbgt/commit/ec8d9453abd4af30e21e4fee8e1d5744d53dbe0d and this is intended as a proof of concept.

To run this code `env SG_TEST_BACKING_STORE=couchbase go test -tags cb_sg_enterprise,cb_sg_devmode ./rest/importtest -run TestImportRollbackAll -count 1`

## Pre-review checklist

- [ ] Link upstream PRs


torcolvin:
This was a proof of concept and was deemed too complicated compared to https://github.com/couchbase/sync_gateway/pull/7553.

PR #7555, opened 2025-06-03
CBG-4670: only update xattr on import

gregns1:
CBG-4670

- This code was incorrectly changed as part of the mou updates. We should only update xattr on import. This code reverst back to how we used to handle this.
- Resync is fine as is.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3142/

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -2062,8 +2062,9 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 			doc.SetCrc32cUserXattrHash()
 			var rawSyncXattr, rawMouXattr, rawDocBody []byte
 			rawDocBody, rawSyncXattr, rawMouXattr, err = doc.MarshalWithXattrs()
-			// if isImport is true, we don't want to update the document body, only the xattrs
-			if !isImport && len(rawDocBody) > 0 {
+			// If isImport is true, we don't generally want to update the document body, only the xattrs. One exception
+			// being when a import is resurrecting a document then we need a body to write back
+			if (!isImport && len(rawDocBody) > 0) || (isImport && doc.Deleted && len(rawDocBody) > 0) {
 				updatedDoc.Doc = rawDocBody
 				docBytes = len(updatedDoc.Doc)
 			}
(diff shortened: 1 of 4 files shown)


rest/importtest/import_test.go
```diff
@@ -2467,3 +2467,31 @@ func TestImportUpdateExpiry(t *testing.T) {
...
+		AutoImport: base.Ptr(true),
+	})
+	defer rt.Close()
+
+	docID := t.Name()
+
+	collection, _ := rt.GetSingleTestDatabaseCollectionWithUser()
+	ds := collection.GetCollectionDatastore()
+
+	specBody := []byte(`{"test":"<>"}`) // use a special character that is currently escaped to ensure it is not modified by the import process
```
bbrks: I think this import codepath is updating the `doc_writes_bytes` stat via:

```go
docBytes := 0   // Track size of document written, for write stats

// ...

// if isImport is true, we don't want to update the document body, only the xattrs
if !isImport && len(rawDocBody) > 0 {
	updatedDoc.Doc = rawDocBody
	docBytes = len(updatedDoc.Doc)

// ...

db.dbStats().Database().DocWritesBytes.Add(int64(docBytes))
```

which might be a useful assertion that continues working for if/when we fix the `<>` bug

bbrks (changes requested):
Looks good, but have a suggestion about better assertion

bbrks (dismissed):

db/crud.go
```diff
@@ -2062,7 +2062,9 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 			doc.SetCrc32cUserXattrHash()
 			var rawSyncXattr, rawMouXattr, rawDocBody []byte
 			rawDocBody, rawSyncXattr, rawMouXattr, err = doc.MarshalWithXattrs()
-			if len(rawDocBody) > 0 {
+			// If isImport is true, we don't generally want to update the document body, only the xattrs. One exception
+			// being when a import is resurrecting a document then we need a body to write back
+			if (!isImport && len(rawDocBody) > 0) || (isImport && doc.Deleted && len(rawDocBody) > 0) {
```
bbrks: This can be simplified to:

```suggestion
			if len(rawDocBody) > 0 && (!isImport || doc.Deleted) {
```

bbrks (changes requested):

bbrks (approved):

PR #7556, opened 2025-06-03
CBG-4638 add missing require_resync to docs

torcolvin:
CBG-4638 add missing require_resync to documentation

Evaluated GET /_status and added missing `vendor` and `cluster_uuid` properties.

### DIFF
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -826,6 +826,9 @@ Retrieved-replication:
         - error
         - starting
         - reconnecting
+    cluster_uuid:
+      description: The cluster unique identifier.
+      type: string
   title: Replication
 Replication:
   description: Properties of a replication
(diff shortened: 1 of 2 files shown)


docs/api/components/schemas.yaml
```diff
@@ -3005,6 +3014,10 @@ Status:
 
         Blank if `api.hide_product_version=true` in the startup configuration.
       type: string
+    vendor:
+      description: Product name
+      type: string
```
bbrks: This is a struct

```go
type vendor struct {
	Name    string `json:"name"`
	Version string `json:"version"`
}
```

bbrks (changes requested):

docs/api/components/schemas.yaml
torcolvin: good catch, thanks

bbrks (approved):

PR #7558, opened 2025-06-04
CBG-4669: Silence resync-specific logging (move to trace)

bbrks:
CBG-4669

- Resync logging moved to trace since we have regular logging already present via `updateChannels` and `updateAccess`

- n/a

### DIFF
--- a/db/document.go
+++ b/db/document.go
@@ -910,7 +910,7 @@ func (doc *Document) updateChannels(ctx context.Context, newChannels base.Set) (
 		}
 	}
 	if changed != nil {
-		base.DebugfCtx(ctx, base.KeyCRUD, "\tDoc %q / %q in channels %q", base.UD(doc.ID), doc.CurrentRev, base.UD(newChannels))
+		base.InfofCtx(ctx, base.KeyCRUD, "\tDoc %q / %q in channels %q", base.UD(doc.ID), doc.CurrentRev, base.UD(newChannels))
 		changedChannels, err = channels.SetFromArray(changed, channels.KeepStar)
 	}
 	return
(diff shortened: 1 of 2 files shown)


torcolvin:
Can you show the difference in logging for a CBL push? I'm worried that losing the logging for CBL debugging might be worse than logging every time for resync, since that's a much less common operation.

In either case, it looks like the crud logging only occurs if there's a change in roles or channels. Many cases of repeated pushing (or resync) will not change the access so that has less utility. Do we log channels when we save a document? If `access()` is called to add a channel to a user doc, is that logged at info or debug?

The use for this debugging typically is "why did my document not get replicated" and we can use channel information to determine that the user didn't have access to the channel at that time. But this might be uncommon, since you can get channel history in sync metadata, and do not need to get this from logs.

bbrks:
Some of this overlaps with the discussion about the additional log level we had today.

I'll revert the `db/document.go` and just remove the unnecessary resync logging for 3.3

db/document.go
```diff
@@ -910,7 +910,7 @@ func (doc *Document) updateChannels(ctx context.Context, newChannels base.Set) (
 		}
 	}
 	if changed != nil {
-		base.InfofCtx(ctx, base.KeyCRUD, "\tDoc %q / %q in channels %q", base.UD(doc.ID), doc.CurrentRev, base.UD(newChannels))
+		base.DebugfCtx(ctx, base.KeyCRUD, "\tDoc %q / %q in channels %q", base.UD(doc.ID), doc.CurrentRev, base.UD(newChannels))
```
bbrks:
```suggestion
		base.InfofCtx(ctx, base.KeyCRUD, "\tDoc %q / %q in channels %q", base.UD(doc.ID), doc.CurrentRev, base.UD(newChannels))
```

db/document.go
```diff
@@ -991,7 +991,7 @@ func (accessMap *UserAccessMap) updateAccess(ctx context.Context, doc *Document,
 		if accessMap == &doc.RoleAccess {
 			what = "role"
 		}
-		base.InfofCtx(ctx, base.KeyAccess, "Doc %q grants %s access: %v", base.UD(doc.ID), what, base.UD(*accessMap))
+		base.DebugfCtx(ctx, base.KeyAccess, "Doc %q grants %s access: %v", base.UD(doc.ID), what, base.UD(*accessMap))
```
bbrks:
```suggestion
		base.InfofCtx(ctx, base.KeyAccess, "Doc %q grants %s access: %v", base.UD(doc.ID), what, base.UD(*accessMap))
```

torcolvin (approved):

PR #7563, opened 2025-06-06
CBG-4065: ISGR and Blipsync User-Agent support

bbrks:
CBG-4065

- Defines a Sync Gateway `User-Agent` header for use by ISGR on `/_blipsync` requests
- Blipsync logs UA at debug for now (tagged as `<UD>` until we're capable of parsing elements)

## Example test output (official builds will also contain build number)
```go
2025-06-06T15:50:56.097+01:00 [INF] HTTP+: c:[310fafbf] db:db #004:     --> 101 [310fafbf] Upgraded to WebSocket protocol BLIP_3+CBMobile_4 (as <ud>AL_1c.e-@</ud>)  (0.0 ms)
2025-06-06T15:50:56.097+01:00 [DBG] HTTP+: c:[310fafbf] db:db BLIP Client had User-Agent: <ud>CouchbaseSyncGateway/3.3.0-EE (ISGR; darwin/arm64)</ud>
```

- n/a

### DIFF
--- a/base/util.go
+++ b/base/util.go
@@ -35,6 +35,7 @@ import (
 	"strings"
 	"sync/atomic"
 	"time"
+	"unicode"
 
 	"github.com/couchbase/gomemcached"
 	"github.com/gorilla/mux"
@@ -1715,4 +1716,14 @@ func EfficientBytesClone(b []byte) []byte {
 	nb := make([]byte, len(b))
 	copy(nb, b)
 	return nb
+}
+
+// removeAllWhitespace removes all whitespace characters from the input string.
+func removeAllWhitespace(s string) string {
+	return strings.Map(func(r rune) rune {
+		if unicode.IsSpace(r) {
+			return -1
+		}
+		return r
+	}, s)
 }
--- a/rest/blip_sync.go
+++ b/rest/blip_sync.go
@@ -94,7 +94,7 @@ func (h *handler) handleBLIPSync() error {
 		// ActiveSubprotocol only available after handshake via ServeHTTP(), so have to get go-blip to invoke callback between handshake and serving BLIP messages
 		subprotocol := blipContext.ActiveSubprotocol()
 		h.logStatus(http.StatusSwitchingProtocols, fmt.Sprintf("[%s] Upgraded to WebSocket protocol %s+%s%s", blipContext.ID, blip.WebSocketSubProtocolPrefix, subprotocol, h.formattedEffectiveUserName()))
-		base.DebugfCtx(h.ctx(), base.KeyHTTP, "BLIP Client had User-Agent: %s", h.rq.Header.Get(base.HTTPHeaderUserAgent))
+		base.DebugfCtx(h.ctx(), base.KeyHTTP, "BLIP Client had User-Agent: %s", base.UD(h.rq.Header.Get(base.HTTPHeaderUserAgent)))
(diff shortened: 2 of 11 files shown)


base/util.go
```diff
@@ -1716,3 +1717,13 @@ func EfficientBytesClone(b []byte) []byte {
 	copy(nb, b)
 	return nb
 }
+
+// removeAllWhitespace removes all whitespace characters from the input string.
+func removeAllWhitespace(s string) string {
```
torcolvin, line 1722: no preference here but you can use: ```re := regexp.MustCompile(`\s+`); re.ReplaceAllString(s, "")```

rest/blip_sync.go
```diff
@@ -94,6 +94,7 @@ func (h *handler) handleBLIPSync() error {
 		// ActiveSubprotocol only available after handshake via ServeHTTP(), so have to get go-blip to invoke callback between handshake and serving BLIP messages
 		subprotocol := blipContext.ActiveSubprotocol()
 		h.logStatus(http.StatusSwitchingProtocols, fmt.Sprintf("[%s] Upgraded to WebSocket protocol %s+%s%s", blipContext.ID, blip.WebSocketSubProtocolPrefix, subprotocol, h.formattedEffectiveUserName()))
+		base.DebugfCtx(h.ctx(), base.KeyHTTP, "BLIP Client had User-Agent: %s", base.UD(h.rq.Header.Get(base.HTTPHeaderUserAgent)))
```
torcolvin: should we log this in the line above with `Updated to WebSocket` so this gets logged at info and not something that might get lost in debug?

not committed to this, but I don't know I'd look for this in a different line, like maybe https://github.com/couchbase/sync_gateway/blob/a7326093dea3d4673d857b608dc74170ef0525a6/rest/handler.go#L752

I don't have strong feelings about this if you want to drop this.

torcolvin (dismissed):

base/util.go
bbrks: I can, but is there a reason to? It's 5x slower even if using precompiled regexp and not any more readable.

```
BenchmarkRemoveAll/strings.Map_unicode.IsSpace-10 11619644  106.6 ns/op   32 B/op  1 allocs/op
BenchmarkRemoveAll/regexp-10                       1000000 1092   ns/op 1027 B/op 16 allocs/op
BenchmarkRemoveAll/regexp-precompiled-10           2260296  524.4 ns/op   88 B/op  4 allocs/op
```

rest/blip_sync.go
bbrks: I do agree. Just wasn't quite sure where to put it or how to format it and adding new line was the lazy approach.

I think I'll put it at the end of the existing `Upgraded to WebSocket protocol` line.

New example:

```go
2025-06-06T18:09:55.016+01:00 [INF] HTTP+: c:[3ec703b8] db:db #004:     --> 101 [3ec703b8] Upgraded to WebSocket protocol BLIP_3+CBMobile_4 from "CouchbaseSyncGateway/3.3.0-EE (ISGR; darwin/arm64)"  (as <ud>AL_1c.e-@</ud>)  (0.0 ms)
```

base/util.go
torcolvin: wow, no reason to do this. I was surprised that this is slower!

torcolvin (approved):

PR #7564, opened 2025-06-06
Skip TestImportRollbackMultiplePartitions if vbucket count not 1024

bbrks:
Test fails on MacOS' 64 vBuckets (and won't work with 100 either)

Skip for now - ticket was already filed to adapt this test CBG-4544 for other counts

### DIFF
--- a/rest/importtest/import_test.go
+++ b/rest/importtest/import_test.go
@@ -2301,6 +2301,12 @@ func TestImportRollbackMultiplePartitions(t *testing.T) {
 	bucket := base.GetTestBucket(t)
 	defer bucket.Close(ctx)
 
+	numVBuckets, err := bucket.GetMaxVbno()
+	require.NoError(t, err)
+	if numVBuckets != 1024 {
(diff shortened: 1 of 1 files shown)


rest/importtest/import_test.go
```diff
@@ -2301,6 +2301,12 @@ func TestImportRollbackMultiplePartitions(t *testing.T) {
 	bucket := base.GetTestBucket(t)
 	defer bucket.Close(ctx)
 
+	numVBuckets, err := bucket.GetMaxVbno()
+	require.NoError(t, err)
+	if numVBuckets != 1024 {
+		t.Skipf("Test requires 1024 vBuckets, but only had %d", numVBuckets)
```
```suggestion
		t.Skipf("Test requires 1024 vBuckets, but only had %d, will fix in CBG-4544", numVBuckets)
```

just so it is easier to find the test from the code, and to know we have a ticket where we are planning to fix when reading the code.

torcolvin (approved):

PR #7575, opened 2025-06-10
CBG-4678: reject writes when unsupported option is set for skipped sequences

gregns1:
CBG-4678

- Reject writes from rest api and blip writes when unsupported option is enabled together with the slow broadcast boolean being set on the node. 
- Test for api pathway, testing with blip tester client not possible as test wil fail as the write will be rejected at sync gateway 
- Tested with new doc pusher and pusher doesn't stop with this error code
- Moved atomic boolean to db context and had to include db context on change listener 

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3145/

### DIFF
--- a/rest/doc_api.go
+++ b/rest/doc_api.go
@@ -601,6 +601,14 @@ func (h *handler) handlePostDoc() error {
 		return base.HTTPErrorf(http.StatusForbidden, auth.GuestUserReadOnly)
 	}
 
+	if h.db.DatabaseContext.Options.UnsupportedOptions != nil && h.db.DatabaseContext.Options.UnsupportedOptions.RejectWritesWithSkippedSequences {
+		// if we are in slow broadcast mode reject write with 503 and increment rejected writes stat
+		if h.db.BroadcastSlowMode.Load() {
+			h.db.DbStats.DatabaseStats.NumDocWritesRejected.Add(1)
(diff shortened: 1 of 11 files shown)


rest/doc_api.go
```diff
@@ -442,6 +442,14 @@ func (h *handler) handlePutDoc() error {
 		return base.HTTPErrorf(http.StatusForbidden, auth.GuestUserReadOnly)
 	}
 
+	if h.db.DatabaseContext.Options.UnsupportedOptions != nil && h.db.DatabaseContext.Options.UnsupportedOptions.RejectWritesWithSkippedSequences {
+		// if we are in slow broadcast mode reject write with 503 and increment rejected writes stat
+		if h.db.BroadcastSlowMode.Load() {
```
bbrks, line 447: Should have coverage for the other APIs that can modify documents, even if they're not being used by the perf tests we're looking at

- `/db/bulk_docs`
- `POST /db/`
- `DELETE /db/doc`

bbrks (changes requested):

rest/doc_api.go
gregns1: Added these into the PR, included them in the test too. Decided to leave `_purge` alone given purged docs should be ignored at the cache when they arrive over dcp.

bbrks (approved):

PR #7576, opened 2025-06-10
CBG-4676: add stats around the public all docs endpoint

gregns1:
CBG-4676

- adds three new stats around the public all docs endpoint 
- Number of requests (this includes any that aren't successful due to some error)
- Number of docs pre filtering 
- Number of docs post filtering (this will include docs the user doesn't have access to for example when you specify the document id in the "keys" property. This is because these docs will be included in the response on this case but will have an error associated with them). 
- All stats added have been tagged as internal - I don't believe the customer learns much from these

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3146/

### DIFF
--- a/rest/bulk_api.go
+++ b/rest/bulk_api.go
@@ -50,9 +50,9 @@ type allDocsRow struct {
 func (h *handler) handleAllDocs() error {
 
 	var numDocsPreFilter, numDocsPostFilter uint64
-	defer func() {
-		// increment public all docs stats here
-		if h.privs != adminPrivs {
+	if h.privs != adminPrivs {
+		defer func() {
+			// public _all_docs stats
 			h.db.DbStats.DatabaseStats.NumPublicAllDocsRequests.Add(1)
 			if numDocsPreFilter > 0 {
 				h.db.DbStats.DatabaseStats.NumDocsPreFilterPublicAllDocs.Add(int64(numDocsPreFilter))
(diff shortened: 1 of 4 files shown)


rest/bulk_api.go
```diff
@@ -48,6 +48,21 @@ type allDocsRow struct {
...
+		if h.privs != adminPrivs {
+			h.db.DbStats.DatabaseStats.NumPublicAllDocsRequests.Add(1)
+			if numDocsPreFilter > 0 {
+				h.db.DbStats.DatabaseStats.NumDocsPreFilterPublicAllDocs.Add(int64(numDocsPreFilter))
+			}
+			if numDocsPostFilter > 0 {
+				h.db.DbStats.DatabaseStats.NumDocsPostFilterPublicAllDocs.Add(int64(numDocsPostFilter))
+			}
+		}
+	}()
```
bbrks: Pull the conditional up out so we avoid the defer cost - `h.privs` won't change mid-request, so we know up-front whether to do this or not.

```suggestion
	// public _all_docs stats
	if h.privs != adminPrivs {
		defer func() {
			h.db.DbStats.DatabaseStats.NumPublicAllDocsRequests.Add(1)
			if numDocsPreFilter > 0 {
				h.db.DbStats.DatabaseStats.NumDocsPreFilterPublicAllDocs.Add(int64(numDocsPreFilter))
			}
			if numDocsPostFilter > 0 {
				h.db.DbStats.DatabaseStats.NumDocsPostFilterPublicAllDocs.Add(int64(numDocsPostFilter))
			}
		}()
	}
```

bbrks (changes requested):
Looks good - one minor suggestion

bbrks (approved):

PR #7581, opened 2025-06-12
Improvements for ISGR docs.

torcolvin:
- move username/password to bottom of docs to represent that they are deprecated and remote_username/remote_password should be used.
- Deduplicate the GET and POST replication configuration.
- Add descriptions for the ISGR replication state.
- Note that remote_username/remote_password are used for all replications, not just push replications.
- Fix the code sample in `custom_conflict_resolver` to be valid JS
- Modify description of delta_sync to match the fact that "" is use the database and only explicitly setting to false will have an effect.
- Correct name of doc_write_failures and doc_write_conflicts stats

I've uncovered some real confusing copy in the docs DOC-13297 which came from an old version of openapi docs, so I went through to try to validate this for correctness.

### DIFF
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -732,9 +732,7 @@ Replication:
       default: false
     enable_delta_sync:
       description: |-
-        If the field is omitted, this will use the delta-sync setting of the database configuration `delta_sync.enabled`.
-
-        If it is set to false, then delta-sync will not be used for this replication.
+        This will turn on delta-sync for the replication. In order to enable delta-sync for a replication, the database level setting `delta_sync.enabled` must also be set to true.
 
         Using delta-sync is an Enterprise Edition only feature.
       type: boolean
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -599,236 +599,44 @@ HTTP-Error:
     - error
     - reason
   title: HTTP-Error
+ISGRReplicationState:
+  description: This is the state that the replicator is in or that trying to transition in to.
+  type: string
+  enum:
+    - running
+    - stopped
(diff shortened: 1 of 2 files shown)


docs/api/components/schemas.yaml
```diff
@@ -707,13 +732,11 @@ Retrieved-replication:
       default: false
     enable_delta_sync:
       description: |-
-        This will turn on delta- sync for the replication. This works in conjunction with the database level setting `delta_sync.enabled`
+        If the field is omitted, this will use the delta-sync setting of the database configuration `delta_sync.enabled`.
```
bbrks: Is that actually true? My reading of the code matches the old description. An omitted value means `false`/turn off delta sync.

bbrks (changes requested):

docs/api/components/schemas.yaml
torcolvin: I read this like this:

Blip delta sync is set on initialization: 
https://github.com/couchbase/sync_gateway/blob/e46771e9e3b10cee76e1a896dcd6cb0d7a5359f9/db/blip_sync_context.go#L56-L56

And it is only set to false here https://github.com/couchbase/sync_gateway/blob/e46771e9e3b10cee76e1a896dcd6cb0d7a5359f9/db/active_replicator.go#L245

bbrks: But `arc.config.DeltasEnabled` is a bool so the zero value will match the if condition and turn it off.

See also `DefaultReplicationConfig()` which explicitly sets the default value of `DeltasEnabled` to false in the replication config.

torcolvin: How confusing, thanks.

bbrks (approved):

PR #7582, opened 2025-06-12
CBG-3935: return error for invalid sync data in import codepath

gregns1:
CBG-3935

- Added check for valid sync data for imports
- Returns error for on-demand imports and feed
- Import feed panic recovery split into #7598

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3155/
- [x] `GSI=true,xattrs=false` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3157/


db/import.go
```diff
@@ -110,6 +111,13 @@ func (db *DatabaseCollectionWithUser) ImportDoc(ctx context.Context, docid strin
 //	mode - ImportMode - ImportFromFeed or ImportOnDemand
 func (db *DatabaseCollectionWithUser) importDoc(ctx context.Context, docid string, body Body, expiry *uint32, isDelete bool, existingDoc *sgbucket.BucketDocument, mode ImportMode) (docOut *Document, err error) {
 
+	// if sync gateway panics during import processing we want to be able to recover instead of bringing the node down
+	defer func() {
+		if r := recover(); r != nil {
+			base.WarnfCtx(ctx, "[%s] Unexpected panic importing document %s - skipping import: \n %s", r, base.UD(docid), debug.Stack())
```
torcolvin: I think this is an error condition and we should definitely log as an error. I do not think this panic recovery is in the right place, nor appropriate at warning.

If we do a panic recovery, we want to pass the error back to the caller, minimally.

But I don't think that we should generally recover from a panic in this subset of code.

```mermaid
flowchart TD
    ProcessFeedEvent --> ImportFeedEvent
    ImportFeedEvent --> ImportDocRaw
    ImportDocRaw --> importDoc
    ImportDoc --> importDoc
    OnDemandImportForWrite --> ImportDoc
    Put --> OnDemandImportForWrite
    handlePutAttachment --> Put
    PutExistingRevWithConflictResolution --> OnDemandImportForWrite
    blip.ProcessRev --> PutExistingRevWithConflictResolution
    PutExistingRev --> PutExistingRevWithConflictResolution
    PutExistingRevWithBody --> PutExistingRev
    handlePutDoc --> PutExistingRevWithBody
```

blip panics are handled with with https://github.com/couchbase/sync_gateway/blob/0f885d7e9adb8287f2bafdaf0095f69f46589f33/db/blip_sync_context.go#L179-L183 - but this typically just causes the handler to fail, without a response.

With a blip replication, you would want to consider what you'd want to happen:

* Should this stop a push or pull replication? Is there an error that CBL should handle? I think for a pull replication an argument should be made that this would send a `norev`, which is what I think we do for other cases.
* For a push replication, do we just want to overwrite the bad contents of the file and not do an on demand import? 

With HTTP handlers:

* Calling GET `/{ks}/doc` or PUT `/{ks}/doc` should probably return an error if they panic. The behavior of ServeHTTP is that any handler function that panics will not take down the process, but log the panic and close the request. https://pkg.go.dev/net/http#Handler


I think we should move the panic handling to separate ticket and consider how we would want to handle the general case after fixing this immediate problem. Minimally, I think we would want the behavior of the panic to be in `ProcessFeedEvent` and send `false` back so that this is checkpointed. If this panics for other reasons (internal gocb errors, that are actually recoverable when the cluster goes online), we don't want the checkpoints to run ahead. The checkpoint would run ahead if there was a document after this one that succeeded, but that is probably OK.

Another common scenario is that many bucket ops will panic such as https://jira.issues.couchbase.com/browse/CBG-4001 and we've been considering whether to add panic recovery on import.

I think we should remove the panic handling from this PR, and make sure we are getting the right response for push/pull/PUT/GET/Import

db/crud.go
```diff
@@ -75,6 +75,11 @@ func (c *DatabaseCollection) GetDocumentWithRaw(ctx context.Context, docid strin
 		// If existing doc wasn't an SG Write, import the doc.
 		if !isSgWrite {
 			var importErr error
+			if !doc.SyncIsEmpty() && !doc.HasValidSyncData() {
+				base.WarnfCtx(ctx, "Invalid sync data for doc %s - not importing.", base.UD(docid))
+				c.dbStats().SharedBucketImportStats.ImportErrorCount.Add(1)
```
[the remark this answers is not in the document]
bbrks: This is the one thing that immediately jumped out at me as well when I skimmed through yesterday.

I think it can be changed quite easily to a function that returns just an error to handle the different returns. It only needs to take a reference to the db to update stats, and `doc`, since you can get `docid` via `doc.ID` for the log line.

bbrks (changes requested):

db/import.go
```diff
@@ -205,7 +205,6 @@ func (db *DatabaseCollectionWithUser) importDoc(ctx context.Context, docid strin
 			}
 			metadataOnlyUpdate = false
 			// If document still requires import post-migration attempt, continue with import processing based on the body returned by migrate
-			doc = migratedDoc
```
bbrks, line 208: Won't this have the result of not having an up to date `CAS` value for `doc` if migrated and written back to bucket?

Not quite sure what the intention of this removal was

gregns1: Easier to just talk through an example. Lets the an example where you have inline sync data and its empty. We call into this `importDoc` callback here https://github.com/couchbase/sync_gateway/blob/0bf4737cee565f9075b3f9c019abffd92f541979/db/crud.go#L1833 Then hit this condition here where we identify sync data is not valid -> https://github.com/couchbase/sync_gateway/blob/0bf4737cee565f9075b3f9c019abffd92f541979/db/import.go#L389 and looks to let the import process fix it essentially. But this line above assigns a new document pointer to doc and further down here -> https://github.com/couchbase/sync_gateway/blob/0bf4737cee565f9075b3f9c019abffd92f541979/db/import.go#L298 when we're adding rev to history on that pointer. Then once we jump back into `documentUpdateFunc` after the callback is finished, doc here is still referencing the old pointer, thus the rev we just created/added is not there and we panic in here https://github.com/couchbase/sync_gateway/blob/0bf4737cee565f9075b3f9c019abffd92f541979/db/crud.go#L1844 as the rev isn't in that history map.

Its possible I am just changing something I shouldn’t and this code needs updating to error in this case and not let import feed fix this.

torcolvin: I think it's very likely we _don't_ want to modify this code. I was wondering if there was a way that we aren't going to panic calling functions with invalid data. I've attached some code before, but this doesn't work either.

I don't want to do a panic recovery, at least not as part of this PR, because of the reasons stated above, we could be swallowing panics we don't want to.

I think we should handle the individual functions that can panic with invalid data. We don't ever want this to panic, even if it is supposed to impossible to call them this way. Writing a strict unit test to just call those functions would be great. I identified two: `setNonWinningRevisionBody` and `setRevisionBody`, since I think the programming contract for go implies that it should never panic, even with bad data.


The example code below is wrong though and causes unrelated tests to fail, and I'm not sure it makes all the new tests pass either. I think this is a good idea even if we want to keep `validateSyncDataForImport`.

The code for `validateSyncDataForImport` feels like it is in the wrong place, but I can't put my finger on it yet where it should be. Is it in `OnDemandImportForWrite`/`OnDemandImportForRead`  or in `ImportDoc/`importDoc`? Or is handling these specific errors sufficient for passing the error back up the stack? I will look at this more but I ran out of time and wanted to provide intermediate feedback.

```
diff --git a/db/crud.go b/db/crud.go
index 538c1125b..dde141b08 100644
--- a/db/crud.go
+++ b/db/crud.go
@@ -75,9 +75,6 @@ func (c *DatabaseCollection) GetDocumentWithRaw(ctx context.Context, docid strin
 		// If existing doc wasn't an SG Write, import the doc.
 		if !isSgWrite {
 			var importErr error
-			if syncDataErr := doc.validateSyncDataForImport(ctx, c.dbCtx, docid); syncDataErr != nil {
-				return nil, nil, syncDataErr
-			}
 			doc, importErr = c.OnDemandImportForGet(ctx, docid, rawBucketDoc.Body, rawBucketDoc.Xattrs, rawBucketDoc.Cas)
 			if importErr != nil {
 				return nil, nil, importErr
@@ -167,10 +164,6 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
 		// If existing doc wasn't an SG Write, import the doc.
 		if !isSgWrite {
 			var importErr error
-			if syncDataErr := doc.validateSyncDataForImport(ctx, c.dbCtx, docid); syncDataErr != nil {
-				return emptySyncData, syncDataErr
-			}
-
 			doc, importErr = c.OnDemandImportForGet(ctx, docid, rawDoc, xattrs, cas)
 			if importErr != nil {
 				return emptySyncData, importErr
@@ -909,9 +902,6 @@ func (db *DatabaseCollectionWithUser) Put(ctx context.Context, docid string, bod
 		// (Be careful: this block can be invoked multiple times if there are races!)
 		// If the existing doc isn't an SG write, import prior to updating
 		if doc != nil && !isSgWrite && db.UseXattrs() {
-			if syncDataErr := doc.validateSyncDataForImport(ctx, db.dbCtx, docid); syncDataErr != nil {
-				return nil, nil, false, nil, syncDataErr
-			}
 			err := db.OnDemandImportForWrite(ctx, newDoc.ID, doc, deleted)
 			if err != nil {
 				if db.ForceAPIForbiddenErrors() {
@@ -1039,9 +1029,6 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithConflictResolution(ctx c
 
 		// If the existing doc isn't an SG write, import prior to updating
 		if doc != nil && !isSgWrite && db.UseXattrs() {
-			if syncDataErr := doc.validateSyncDataForImport(ctx, db.dbCtx, newDoc.ID); syncDataErr != nil {
-				return nil, nil, false, nil, syncDataErr
-			}
 			err := db.OnDemandImportForWrite(ctx, newDoc.ID, doc, newDoc.Deleted)
 			if err != nil {
 				return nil, nil, false, nil, err
@@ -1511,7 +1498,7 @@ func (doc *Document) updateWinningRevAndSetDocFlags(ctx context.Context) {
 	}
 }
 
-func (db *DatabaseCollectionWithUser) storeOldBodyInRevTreeAndUpdateCurrent(ctx context.Context, doc *Document, prevCurrentRev string, newRevID string, newDoc *Document, newDocHasAttachments bool) {
+func (db *DatabaseCollectionWithUser) storeOldBodyInRevTreeAndUpdateCurrent(ctx context.Context, doc *Document, prevCurrentRev string, newRevID string, newDoc *Document, newDocHasAttachments bool) error {
 	if doc.HasBody() && doc.CurrentRev != prevCurrentRev && prevCurrentRev != "" {
 		// Store the doc's previous body into the revision tree:
 		oldBodyJson, marshalErr := doc.BodyBytes(ctx)
@@ -1559,10 +1546,14 @@ func (db *DatabaseCollectionWithUser) storeOldBodyInRevTreeAndUpdateCurrent(ctx
 		if marshalErr != nil {
 			base.WarnfCtx(ctx, "Unable to marshal document body properties for storage in rev tree: %v", marshalErr)
 		}
-		doc.setNonWinningRevisionBody(prevCurrentRev, oldBodyJson, db.AllowExternalRevBodyStorage(), oldDocHasAttachments)
+		doc.setNonWinningRevisionBody(ctx, prevCurrentRev, oldBodyJson, db.AllowExternalRevBodyStorage(), oldDocHasAttachments)
 	}
 	// Store the new revision body into the doc:
-	doc.setRevisionBody(ctx, newRevID, newDoc, db.AllowExternalRevBodyStorage(), newDocHasAttachments)
+	err := doc.setRevisionBody(ctx, newRevID, newDoc, db.AllowExternalRevBodyStorage(), newDocHasAttachments)
+	if err != nil {
+		base.WarnfCtx(ctx, "Unable to set revision body for doc %s: %v", base.UD(doc.ID), err)
+		return err
+	}
 	doc.SyncData.Attachments = newDoc.DocAttachments
 	doc.metadataOnlyUpdate = newDoc.metadataOnlyUpdate
 
@@ -1578,6 +1569,7 @@ func (db *DatabaseCollectionWithUser) storeOldBodyInRevTreeAndUpdateCurrent(ctx
 			doc.metadataOnlyUpdate = nil
 		}
 	}
+	return nil
 }
 
 func (db *DatabaseCollectionWithUser) prepareSyncFn(doc *Document, newDoc *Document) (mutableBody Body, metaMap map[string]interface{}, newRevID string, err error) {
@@ -1855,7 +1847,10 @@ func (col *DatabaseCollectionWithUser) documentUpdateFunc(ctx context.Context, d
 	prevCurrentRev := doc.CurrentRev
 	doc.updateWinningRevAndSetDocFlags(ctx)
 	newDocHasAttachments := len(newAttachments) > 0
-	col.storeOldBodyInRevTreeAndUpdateCurrent(ctx, doc, prevCurrentRev, newRevID, newDoc, newDocHasAttachments)
+	err = col.storeOldBodyInRevTreeAndUpdateCurrent(ctx, doc, prevCurrentRev, newRevID, newDoc, newDocHasAttachments)
+	if err != nil {
+		return
+	}
 
 	syncExpiry, oldBodyJSON, channelSet, access, roles, err := col.runSyncFn(ctx, doc, mutableBody, metaMap, newRevID)
 	if err != nil {
diff --git a/db/document.go b/db/document.go
index a3ec0f992..7fb5587f1 100644
--- a/db/document.go
+++ b/db/document.go
@@ -696,24 +696,23 @@ func (doc *Document) pruneRevisions(ctx context.Context, maxDepth uint32, keepRe
 }
 
 // Adds a revision body (as Body) to a document.  Removes special properties first.
-func (doc *Document) setRevisionBody(ctx context.Context, revid string, newDoc *Document, storeInline, hasAttachments bool) {
+func (doc *Document) setRevisionBody(ctx context.Context, revid string, newDoc *Document, storeInline, hasAttachments bool) error {
 	if revid == doc.CurrentRev {
 		doc._body = newDoc._body
 		doc._rawBody = newDoc._rawBody
-	} else {
-		bodyBytes, _ := newDoc.BodyBytes(ctx)
-		doc.setNonWinningRevisionBody(revid, bodyBytes, storeInline, hasAttachments)
 	}
+	bodyBytes, _ := newDoc.BodyBytes(ctx)
+	return doc.setNonWinningRevisionBody(ctx, revid, bodyBytes, storeInline, hasAttachments)
 }
 
 // Adds a revision body (as []byte) to a document.  Flags for external storage when appropriate
-func (doc *Document) setNonWinningRevisionBody(revid string, body []byte, storeInline bool, hasAttachments bool) {
+func (doc *Document) setNonWinningRevisionBody(ctx context.Context, revid string, body []byte, storeInline bool, hasAttachments bool) error {
 	revBodyKey := ""
 	if !storeInline && len(body) > MaximumInlineBodySize {
 		revBodyKey = generateRevBodyKey(doc.ID, revid)
 		doc.addedRevisionBodies = append(doc.addedRevisionBodies, revid)
 	}
-	doc.History.setRevisionBody(revid, body, revBodyKey, hasAttachments)
+	return doc.History.setRevisionBody(revid, body, revBodyKey, hasAttachments)
 }
 
 // persistModifiedRevisionBodies writes new non-inline revisions to the bucket.
diff --git a/db/document_test.go b/db/document_test.go
index 027a08bda..eaa4823b4 100644
--- a/db/document_test.go
+++ b/db/document_test.go
@@ -13,6 +13,7 @@ package db
 import (
 	"bytes"
 	"encoding/binary"
+	"fmt"
 	"log"
 	"testing"
 
@@ -354,3 +355,19 @@ func getSingleXattrDCPBytes() []byte {
 	dcpBody = append(dcpBody, body...)
 	return dcpBody
 }
+
+func TestDocumentInvalidRevtree(t *testing.T) {
+	ctx := base.TestCtx(t)
+	for _, storeInline := range []bool{true, false} {
+		t.Run(fmt.Sprintf("storeInline=%t", storeInline), func(t *testing.T) {
+			for _, hasAttachments := range []bool{true, false} {
+				t.Run(fmt.Sprintf("hasAttachments=%t", hasAttachments), func(t *testing.T) {
+					existingDoc := NewDocument("testDocID")
+					newDoc := NewDocument("testDocID")
+					err := existingDoc.setRevisionBody(ctx, "1-1234567890abcdef", newDoc, storeInline, hasAttachments)
+					require.Error(t, err)
+				})
+			}
+		})
+	}
+}
diff --git a/db/revtree.go b/db/revtree.go
index 08c2dc11b..7d5ad22df 100644
--- a/db/revtree.go
+++ b/db/revtree.go
@@ -407,20 +407,19 @@ func (tree RevTree) getRevisionBody(revid string, loader RevLoaderFunc) ([]byte,
 	return info.Body, true
 }
 
-func (tree RevTree) setRevisionBody(revid string, body []byte, bodyKey string, hasAttachments bool) {
+func (tree RevTree) setRevisionBody(revid string, body []byte, bodyKey string, hasAttachments bool) error {
 	if revid == "" {
-		// TODO: CBG-1948
-		panic("Illegal empty revision ID")
+		return fmt.Errorf("setRevisionBody called with empty revid: %q", revid)
 	}
 	info, found := tree[revid]
 	if !found {
-		// TODO: CBG-1948
-		panic(fmt.Sprintf("rev id %q not found", revid))
+		return fmt.Errorf("setRevisionBody called for revid %s not in history: %v", revid, tree)
 	}
 
 	info.BodyKey = bodyKey
 	info.Body = body
 	info.HasAttachments = hasAttachments
+	return nil
 }
 
 func (tree RevTree) removeRevisionBody(ctx context.Context, revid string) (deletedBodyKey string) {
diff --git a/db/revtree_test.go b/db/revtree_test.go
index 0b34f42e5..695371a2e 100644
--- a/db/revtree_test.go
+++ b/db/revtree_test.go
@@ -1176,6 +1176,41 @@ func (tree RevTree) LongestBranch() int {
 
 }
 
+func TestInvalidRevTree(t *testing.T) {
+	// assure these cases have an error returned
+	testCases := []struct {
+		name    string
+		revTree RevTree
+		revID   string
+		body    []byte
+	}{
+		{
+			name:    "Empty RevTree",
+			revTree: RevTree{},
+			revID:   "1-foo",
+			body:    []byte(`{"foo":"bar"}`),
+		},
+		{
+			name: "No Rev for revid specified",
+			revTree: RevTree{
+				"1-foo": {ID: "1-foo"},
+			},
+			revID: "2-bar",
+			body:  []byte(`{"foo":"bar"}`),
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			for _, hasAttachments := range []bool{false, true} {
+				t.Run(fmt.Sprintf("hasAttachments=%t", hasAttachments), func(t *testing.T) {
+					bodyKey := "bodyKey"
+					require.Error(t, tc.revTree.setRevisionBody(tc.revID, tc.body, bodyKey, hasAttachments))
+				})
+			}
+		})
+	}
+}
+
 // Create body content as map of 100 byte entries.  Rounds up to the nearest 100 bytes
 func createBodyContentAsMapWithSize(docSizeBytes int) map[string]string {
 ```

File: db/import.go (hunk 1 above)
Comment (torcolvin) [reply db/import.go]: Filed https://jira.issues.couchbase.com/browse/CBG-4697 to do the general panic handling, which should be done in a separate PR.

File: db/import.go (hunk 3 above)
Comment (torcolvin) [reply db/import.go:208]: I want test like this to make sure that a replication from CBL will be robust to this issue and the handling of this document will not stop a replication. In a pull replication this should send a `norev`. In a push replication, I'm not sure if this should skip this document or whether the replication stops. There are some cases of this we've encountered before https://jira.issues.couchbase.com/browse/CBG-4028 but I'm not sure if this is handled for push.:

- no auto import
- write invalid doc1 to bucket
- CBL write doc1
- CBL continuous replication starts
- CBL write doc2
- make sure this gets replicated

I want something similar on the pull side but I'm not sure if this is possible?

- no auto import
- cbl continuous pull replication
- write doc 1 to sync gateway (valid)
- validate CBL has doc1
- write invalid doc directly to bucket
- does this document show up in the changes feed?
- write doc2 to sync gateway (valid)
- make sure that CBL has doc2

I am not 100% sure what should happen in these scenarios but I think it's worthwhile to write tests in order to find out.

File: db/import.go [hunk 4]
```diff
@@ -376,19 +369,19 @@ func (db *DatabaseCollectionWithUser) importDoc(ctx context.Context, docid strin
...
+		return nil, unmarshalErr
 	}
 	doc.Cas = existingDoc.Cas
 
 	// If no sync metadata is present, return for import handling
 	if !doc.HasValidSyncData() {
-		base.InfofCtx(ctx, base.KeyMigrate, "During migrate, doc %q doesn't have valid sync data.  Falling back to import handling.  (cas=%d)", base.UD(docid), doc.Cas)
-		return doc, true, nil
+		base.InfofCtx(ctx, base.KeyMigrate, "During migrate, doc %q doesn't have valid sync data.", base.UD(docid))
+		return doc, fmt.Errorf("not imported, invalid sync data found")
```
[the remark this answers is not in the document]
bbrks: Disagree - we're not in the context of a HTTP request. Maybe an error variable for reuse with the one other place, but it's a nitpick if anything.

db/crud.go
```diff
@@ -236,11 +235,15 @@ func (db *DatabaseCollection) GetDocSyncDataNoImport(ctx context.Context, docid
 
 // OnDemandImportForGet.  Attempts to import the doc based on the provided id, contents and cas.  ImportDocRaw does cas retry handling
 // if the document gets updated after the initial retrieval attempt that triggered this.
-func (c *DatabaseCollection) OnDemandImportForGet(ctx context.Context, docid string, rawDoc []byte, xattrs map[string][]byte, cas uint64) (docOut *Document, err error) {
+func (c *DatabaseCollection) OnDemandImportForGet(ctx context.Context, docid string, doc *Document, rawDoc []byte, xattrs map[string][]byte, cas uint64) (docOut *Document, err error) {
 	isDelete := rawDoc == nil
 	importDb := DatabaseCollectionWithUser{DatabaseCollection: c, user: nil}
 	var importErr error
```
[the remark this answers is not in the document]
bbrks: Overly defensive IMO and error would be misleading for users. We have other problems further up stack if `doc` is nil.

bbrks (approved):

PR #7584, opened 2025-06-12
Fix Sync Gateway 3.3 index documentation

torcolvin:
- Fix missing `database.index.num_replicas` section
- Fix bad links

### DIFF
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -1737,8 +1737,14 @@ Database:
       type: number
       default: 1
     index:
-      allOf:
-        - $ref: '#/IndexSettings'
+      description: "Global Secondary Index Settings"
+      properties:
+        num_partitions:
+          $ref: '#/NumIndexPartitions'
+        num_replicas:
+          description: This is the number of Global Secondary Indexes (GSI) to use for core indexes.
+          type: number
+          default: 1
     use_views:
       description: Force the use of views instead of GSI.
       type: boolean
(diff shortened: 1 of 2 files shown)


simon-dew (dismissed):
Okay so I think this is now correct. Just checking for confirmation:

* In `POST {db}/_index_init`, the request body contains `create_separate_principal_indexes` and `num_partitions` (not `num_replicas`)
* In `GET {db}/_index_init`, the `settings` property in the response contains `create_separate_principal_indexes` and `num_partitions` (not `num_replicas`)
* In `PUT` and `POST {db}/_config`, the `index` property in the request body now contains `num_replicas` and `num_partitions`
* Likewise in `GET {db}/_config`, the `index` property in the response now contains `num_replicas` and `num_partitions`

If that's expected, then this LGTM 👍

simon-dew:
Note that `GET {db}/_index_init` has a period at the end of the summary ("Get status of index initialization.") which is inconsistent with all the other operations. Annoying, but maybe outside the scope of this pull request?

torcolvin:
> Okay so I think this is now correct. Just checking for confirmation:
> 
> * In `POST {db}/_index_init`, the request body contains `create_separate_principal_indexes` and `num_partitions` (not `num_replicas`)
> * In `GET {db}/_index_init`, the `settings` property in the response contains `create_separate_principal_indexes` and `num_partitions` (not `num_replicas`)
> * In `PUT` and `POST {db}/_config`, the `index` property in the request body now contains `num_replicas` and `num_partitions`
> * Likewise in `GET {db}/_config`, the `index` property in the response now contains `num_replicas` and `num_partitions`
> 
> If that's expected, then this LGTM 👍

This is correct! It was quite wrong before.

torcolvin:
> Note that `GET {db}/_index_init` has a period at the end of the summary ("Get status of index initialization.") which is inconsistent with all the other operations. Annoying, but maybe outside the scope of this pull request?

Happy to make changes like this, I'm definitely not attuned to these sorts of things when I write them. I'm also happy to have anyone from docs team make PRs for stuff like this. 

I made the call to put a lot of technical information into the openapi spec on how to use partitioned indexes, figuring this was going to be less likely to get separated from the nitty gritty on how to use this. If there is a more clear way to express something, I will defer to your expertise.

bbrks (approved):

PR #7587, opened 2025-06-13
CBG-4473: Reset Resync on resume when changing collections

bbrks:
CBG-4473

- Store collection IDs in resync meta
- Reset resync ID/checkpoints when collection set changes

## Dependencies
- [x] #7585 
- [x] #7586 
- [x] Rebase for above

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3163/

### DIFF
--- a/base/util.go
+++ b/base/util.go
@@ -1727,3 +1727,20 @@ func removeAllWhitespace(s string) string {
 		return r
 	}, s)
 }
+
+// SlicesEqualIgnoreOrder checks if two slices are equal, ignoring order.
+func SlicesEqualIgnoreOrder[T comparable](a, b []T) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	aLookup := make(map[T]struct{}, len(a))
+	for _, aItem := range a {
+		aLookup[aItem] = struct{}{}
+	}
+	for _, bItem := range b {
+		if _, ok := aLookup[bItem]; !ok {
+			return false
+		}
+	}
+	return true
+}
--- a/base/util.go
+++ b/base/util.go
@@ -1733,14 +1733,15 @@ func SlicesEqualIgnoreOrder[T comparable](a, b []T) bool {
 	if len(a) != len(b) {
 		return false
 	}
-	aLookup := make(map[T]struct{}, len(a))
+	aCount := make(map[T]int, len(a))
 	for _, aItem := range a {
-		aLookup[aItem] = struct{}{}
+		aCount[aItem]++
 	}
 	for _, bItem := range b {
-		if _, ok := aLookup[bItem]; !ok {
+		if count, ok := aCount[bItem]; !ok || count == 0 {
 			return false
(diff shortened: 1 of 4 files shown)


base/util.go
```diff
@@ -1727,3 +1727,20 @@ func removeAllWhitespace(s string) string {
 		return r
 	}, s)
 }
+
+// SlicesEqualIgnoreOrder checks if two slices are equal, ignoring order.
+func SlicesEqualIgnoreOrder[T comparable](a, b []T) bool {
```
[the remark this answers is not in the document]
gregns1: When b has multiple occurrences of the same value that occurs in a then this fails, for example `a = []any{1, 2, 3}` and `b = []any{1, 2, 2}` this function passes as true when they aren’t equal. 

I think given this specific use case that collection ID's should be unique for each bucket so there should be no multiple occurrences in each list of the same value. It may just be worth adding a comment to this function to make it clear the limitations of this function.

db/background_mgr_resync_dcp.go
```diff
@@ -86,6 +86,8 @@ func (r *ResyncManagerDCP) Init(ctx context.Context, options map[string]interfac
 		resetMsg = "failed to unmarshal cluster status"
 	} else if statusDoc.State == BackgroundProcessStateCompleted {
 		resetMsg = "previous run completed"
+	} else if !base.SlicesEqualIgnoreOrder(r.collectionIDs, statusDoc.CollectionIDs) {
```
gregns1, line 89: Do we know if the reset process here clears any old checkpoints up from the bucket? I feel if not it could be out of scope for this PR but an enhancement we could make in future to avoid unnecessary files being in the bucket.

gregns1 (commented):
Generally looks good to me just couple of comments.

db/background_mgr_resync_dcp.go
bbrks: It doesn't :( unfortunately this affects more than just the new collections case - any reset will leave behind the old DCP checkpoints.

We'll have to clean them up at reset time when we still have access to the previous ResyncID, and unfortunately this means we can't clean up any pre-existing dcp checkpoints.

I'll file a CBG and work on this separately, but it sounds like we may want to get this into 3.3 if at all possible without affecting the schedule, as the longer that gap exists the worse the problem will get for users.

bbrks: CBG-4692

base/util.go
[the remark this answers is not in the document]
bbrks: Copilot and yourself totally caught this and it was an easy enough fix to make it work.

Thanks!

bbrks:
@gregns1 sorry for the larger diff now... I fixed a reocurring typo `resyc` -> `resync`

gregns1 (approved):

PR #7588, opened 2025-06-13
improve shared_bucket_access documentation

torcolvin:
Inspired from https://jira.issues.couchbase.com/browse/DOC-12474

- `enable_shared_bucket_access` is going to removed in 4.0, so officially mark this as removed.
- clarify that `import_docs` implies for DCP import feed, but not on demand import.
- add upper and lower limit for import partitions

### DIFF
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -1714,7 +1714,7 @@ Database:
       description: Whether to use extended attributes to store Sync Gateway document (`_sync`) metadata.
       type: boolean
       default: true
-      deprecated: false
+      deprecated: true
     session_cookie_secure:
       description: |-
         Override the session cookie `secure` flag. If set, the cookie will have the `secure` flag.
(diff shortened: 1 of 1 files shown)


docs/api/components/schemas.yaml
```diff
@@ -1712,6 +1714,7 @@ Database:
       description: Whether to use extended attributes to store Sync Gateway document (`_sync`) metadata.
       type: boolean
       default: true
+      deprecated: false
```
bbrks: This is more misleading than not specifying it IMO... The `false` value for this option is deprecated (since 3.0), and arguably _using_ this config at all is deprecated... since the default value is already what we want. I would set this to `true` if anything at all.

bbrks (changes requested):

docs/api/components/schemas.yaml
torcolvin: whoops this was a brain typo, I was going to mark this value as deprecated.

iamfortune:
This lgtm

iamfortune (approved):

bbrks (approved):

PR #7589, opened 2025-06-13
CBG-4554 ask cbgt to perform a reconnect after rollback

torcolvin:
CBG-4554 ask cbgt to perform a reconnect after rollback
CBG-3451 Log on successful sharded DCP stream open

- uptake cbgt changes:
	- MB-66789 to add debug logging
	- MB-66749 create option to reconnect streams on re-open
	
Change the way that cbgt rollbacks are handled by Sync Gateway. If on opening a DCP Stream, kv detects a rollback, it will return with the updated vbuuid + sequence number to be used to reconnect to a rollback. This value is already set by `cbgt.Dest.RollbackEx` and Sync Gateway sets this new value in its in `DCPDest.RollbackEx` with `DCPCommon.setMetadata`.

Before this code change, after setting the metadata the janitor rollback function was called. This would call the janitor rollback function which closes the pindex and recreates a new pindex. This would happen up to 1024 (number of vbuckets). This is very heavy weight for what sync gateway needs to do, and created complications around locking.

Sync Gateway is fine with just re initiating an open stream call with the new metadata for a single vbucket, and avoiding a janitor rollback greatly simplifies the code pathway for Sync Gateawy. Additionally, there is no requirement for locking across vBuckets. The behavior for Sync Gateway differs from fts, since fts needs to close the pindex because checkpoints for all vbuckets in a pindex are tied together, and fts does additional locking.

Some code cleanup which could be dropped to limit the scope of the PR:

- dropped passing `BucketSpec` through `cbgtFeedParams` since Xattrs have to be enabled to run an import feed
- drop `janitorRollback` function through `DCPDest` and `DCPCommon`, since we never want to call the function
- use `cbgt.DCPFeedParams` embedded struct to prevent string matching problems 

To pick up the logging changes which are at the debug level, add `clog.SetLevel(clog.LevelDebug)` since clog will filter https://github.com/couchbase/clog/blob/f935d1fdfc36541b505cf86fea4822e4067f9c39/clog.go#L318 debug logging. These were previously hidden from Sync Gateway logs. There is only one case of debug logging from cbgt that is not related to opening DCP streams, which shouldn't be called often. https://github.com/couchbase/cbgt/blob/be3e8c315a106901044aaab3ac150a4f7d463aed/cmd/planner.go#L89 and I think logging cbgt debug logs at sync gateway debug level is fine.

## Pre-review checklist

- [x] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3167/


base/dcp_common.go
```diff
@@ -145,14 +145,15 @@ func (c *DCPCommon) getMetaData(vbucketId uint16) (
 	return value, lastSeq, nil
 }
 
-// rollbackEx is called when a DCP stream issues a rollback. The metadata persisted for a given uuid and sequence number and then cbgt.Mgr JANITOR_ROLLBACK_PINDEX is issued via janitorRollback function.
-func (c *DCPCommon) rollbackEx(vbucketId uint16, vbucketUUID uint64, rollbackSeq uint64, rollbackMetaData []byte, janitorRollback func()) error {
+// rollbackEx is called when a DCP open stream a rollback. The metadata persisted for a given uuid and sequence number and stream reopening is deferred to cbgt via AutoReconnectAfterRollback feed parameter.
+func (c *DCPCommon) rollbackEx(vbucketId uint16, vbucketUUID uint64, rollbackSeq uint64, rollbackMetaData []byte) error {
```
adamcfraser, line 149: Not strictly related to this PR, but given that there are many normal situations for rollback, I feel like we should downgrade the logging on line 150 to Info.  What do you think?

base/dcp_feed_type.go
```diff
@@ -161,12 +152,13 @@ type SGFeedIndexParams struct {
...
-	feedParams.DbName = dbName
-
-	if spec.UseXattrs {
-		feedParams.IncludeXAttrs = true
+func cbgtFeedParams(ctx context.Context, scope string, collections []string, dbName string) (string, error) {
+	feedParams := &SGFeedSourceParams{
+		DbName: dbName,
+		DCPFeedParams: cbgt.DCPFeedParams{
+			AutoReconnectAfterRollback: true,
+			IncludeXAttrs:              true,
```
adamcfraser, line 160: This change isn't related to the PR (to always set xattrs=true), and it's hard to tell the impact of this change.  I'm assuming that elsewhere we're not using cbgt in the first place if we're not using xattrs?  How confident are you that there aren't unintended side-effects here?

torcolvin: I agree this isn't related to the PR, but there is no way we are running an import feed without xattrs being enabled - this would be pointless.

base/dcp_common.go
torcolvin: Agreed this isn't a warning situation.

adamcfraser (approved):

PR #7592, opened 2025-06-17
CBG-4693: close done chan in correct location

gregns1:
CBG-4693

- calling close done chan in wrong place so stop change listener wasn't waiting for broadcast to close

### DIFF
--- a/db/change_listener.go
+++ b/db/change_listener.go
@@ -264,12 +264,12 @@ func (listener *changeListener) StartNotifierBroadcaster(ctx context.Context) {
 	// for a value it already has
 	broadcastSlowMode := false
 	go func(terminator chan bool) {
+		defer func() {close(listener.broadcastChangesDoneChan)}()
 		var currCount uint64
 		for {
 			select {
 			case <-terminator:
 				ticker.Stop()
-				close(listener.broadcastChangesDoneChan)
 				return
 			case <-ticker.C:
 				// if the counter has changed, notify waiting clients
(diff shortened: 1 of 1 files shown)


db/change_listener.go
```diff
@@ -270,6 +269,7 @@ func (listener *changeListener) StartNotifierBroadcaster(ctx context.Context) {
 			select {
 			case <-terminator:
 				ticker.Stop()
+				close(listener.broadcastChangesDoneChan)
```
bbrks: I'd keep the defer and move it to the top of the goroutine function personally.

We don't know if/when we'll introduce another case in here that can cause the goroutine to exit (e.g. let's say we started using `context` to close this in addition to the terminator channel.

bbrks (changes requested):

db/change_listener.go
```diff
@@ -260,11 +260,13 @@ func (listener *changeListener) Notify(ctx context.Context, keys channels.Set) {
...
 func (listener *changeListener) StartNotifierBroadcaster(ctx context.Context) {
 	ticker := time.NewTicker(DefaultBroadcastChangesTime)
-	defer func() { close(listener.broadcastChangesDoneChan) }()
 	// boolean to indicate whether ticker is using the default value, this is needed so we don't call reset on ticker
 	// for a value it already has
 	broadcastSlowMode := false
 	go func(terminator chan bool) {
+		defer func() {
+			close(listener.broadcastChangesDoneChan)
+		}()
```
bbrks: Not 100% sure about this, but you might have to pass `listener.broadcastChangesDoneChan` into the goroutine like we did for terminator to take a copy - the race detector was flagging this when `listener` changed I think?

bbrks (changes requested):

db/change_listener.go
gregns1: Good catch, forgot about that.

bbrks (approved):

PR #7593, opened 2025-06-17
CBG-4493: Optimize happy-path for collections that have all indexes online

bbrks:
CBG-4493

- Skips the per-index `CREATE INDEX IF NOT EXISTS` and `WaitForIndexesOnline` steps based on an initial online check for the whole collection.
- In the best-case of an already initialized database/collection, we only run a single query instead of up to 7 
- Other minor optimizations by reusing the same set of index `sgIndex` instead of re-evaluating `sgIndex.shouldCreate` every time

Manual testing with a locally running Couchbase server and 3 named collections showed a proportional 6-7x improvement on database startup time - from ~340ms to ~50ms

I evaluated options to reduce further by checking all indexes on the entire set of collections with a single query, but that proved challenging to implement due to a lot of per-collection index option logic. I decided it was not worth the extra effort and code complexity.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3170/

### DIFF
--- a/db/indexes.go
+++ b/db/indexes.go
@@ -427,7 +427,6 @@ func InitializeIndexes(ctx context.Context, n1QLStore base.N1QLStore, options In
 			// remove from list of indexes to create/wait for
 			return true
 		}
-		base.DebugfCtx(ctx, base.KeyAll, "Index %s is not online - will issue CREATE INDEX statement and wait for readiness", idx.fullIndexName)
 		return false
 	})
 
--- a/db/indexes.go
+++ b/db/indexes.go
@@ -376,61 +376,89 @@ type InitializeIndexOptions struct {
 	LegacySyncDocsIndex        bool                            // if true, create legacy sync docs index (for backwards compatibility)
 }
 
+type sgIndexToBuild struct {
+	fullIndexName string
+	sgIndex       SGIndex
+}
+
+type sgIndexesToBuild []sgIndexToBuild
+
+func (s sgIndexesToBuild) FullIndexNames() []string {
+	names := make([]string, 0, len(s))
+	for _, sgIndex := range s {
+		names = append(names, sgIndex.fullIndexName)
+	}
+	return names
+}
+
 // Initializes Sync Gateway indexes for datastore.  Creates required indexes if not found, then waits for index readiness.
 func InitializeIndexes(ctx context.Context, n1QLStore base.N1QLStore, options InitializeIndexOptions) error {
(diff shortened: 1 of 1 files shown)


db/indexes.go
```diff
@@ -376,61 +376,89 @@ type InitializeIndexOptions struct {
...
-		err := sgIndex.createIfNeeded(ctx, n1QLStore, options)
-		if err != nil {
-			if !errors.Is(err, base.ErrIndexBackgroundRetry) {
-				return base.RedactErrorf("Unable to install index %s: %v", base.MD(sgIndex.simpleName), err)
-			}
-		}
+	if len(requiredIndexes) == 0 {
+		// not expected to get here - all our collections require at least one index ...
+		base.AssertfCtx(ctx, "No indexes to create - expected at least one index to be created, but none were found")
+		return nil
```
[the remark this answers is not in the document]
bbrks: Disagree - SG assertion code is safe and will just Warn in production.

db/indexes.go
```diff
@@ -376,61 +376,89 @@ type InitializeIndexOptions struct {
...
+	// Drop indexes from the list if they're already online to avoid any per-index init cost.
+	requiredIndexes = slices.DeleteFunc(requiredIndexes, func(idx sgIndexToBuild) bool {
+		meta, exists := indexesMeta[idx.fullIndexName]
+		if exists && meta.State == base.IndexStateOnline {
+			base.DebugfCtx(ctx, base.KeyAll, "Index %s is already online", idx.fullIndexName)
+			// remove from list of indexes to create/wait for
+			return true
 		}
-	}
+		base.DebugfCtx(ctx, base.KeyAll, "Index %s is not online - will issue CREATE INDEX statement and wait for readiness", idx.fullIndexName)
```
torcolvin: I think this statement could be deceptive and I'd be inclined to drop it - while we do CREATE INDEX IF NOT EXISTS, this could be confusing in the case that the index exists but is in the process of building.

It looks like we don't log create index anyway but we do log in `WaitForIndexesOnline` when it is ready and when it is not ready and I think that's sufficient.

torcolvin (dismissed):

db/indexes.go
bbrks: Yep that's fair - we already have enough logging in the existing codepath.

```suggestion
```

bbrks (commented):
.

gregns1 (approved):

PR #7597, opened 2025-06-19
Add uv lock and mypy to do checking

torcolvin:
This python code is fragile and buggy and doing anything to help validate it can be helpful.

- Use `uv` to pin the environment and make it easier to run CI and do testing. `uv` will cache modules and make testing easier.
	- `uv run -- pytest` to run tests
	- `uv run -- mypy` to run type checking
	- `uv run -- ruff check` to run linter
	- `uv run -- ruff format` to auto format file
- Straight move `sgcollect_info` bulk to `sgcollect.py` so it is importable via mypy and the tools-test directory can be reported. As long as `tools/sgcollect_info` is executable, it will be packaged up in https://github.com/couchbase/build/blob/0961883a4a6afcc1fb82e9e876af454f43944298/scripts/jenkins/mobile/sgw_unix_build.sh#L350 `sgcollect.py` should match the origin/main `sgcollect_info` except for removing `__main__` and one type check change.
- use mypy to do type checking. I had to fix one spot in `sgcollect.py` to pass the correct type `add_file_collection`.
- turned mypy options strict except for pieces that do not pass (yet)


torcolvin:
> Comments suppressed due to low confidence (1)
> **tools/tasks.py:199**
> 
> * The code references `threading.Event` and `threading.Timer` but `threading` is not imported. Please add `import threading` at the top of the file.
> 
> ```
>         timer_fired = threading.Event()
> ```

This is just wrong, `import threading` is at the top of the file.

bbrks (approved):
LGTM - verified `tools/sgcollect.py` is largely unchanged too - since the diff here wasn't helpful.

```diff
$ diff main-sgcollect_info sgcollect.py
406c406
<     def add_file_collection_task(filename: pathlib.Path):
---
>     def add_file_collection_task(filename: str):
1026,1029d1025
<
<
< if __name__ == "__main__":
<     main()
```

PR #7601, opened 2025-06-23
CBG-4712 /_sgcollect_info endpoint to work with non default port

torcolvin:
CBG-4712 /_sgcollect_info endpoint to work with non default port

This fix was driven by the desire to have integration testing from go, and doing integration testing in go requires listening on ports that are not :4985 - it uses :0 and picks a port randomly.

Fixes made:

- pass `--sync-gateway-url` from `/_sgcollect_info`
- change the way URL is picked for SG (`--sync-gateway-url`, `http://127.0.0.1:4985`, `https://127.0.0.1:4985`)
- always use `--sync-gateway-executable` first if present, rather than picking up default locations
- remove `get_absolute_path` function since this would always throw an exception since upgrade to python3 (found with parsing log output for errors, then using python typing). This function mixes up bytes and str.
- Close the io.Pipe for stderr/stdout, without this, each time you run sgcollect 2 goroutines are leaked.

Testing only improvements:

- scope `sgCollect` instance to the `ServerContext`. In practice, this is the same for a sync gateway executable but for testing it means not worrying about cleaning up globals since we can modify the ServerContext and it is torn down between tests. This also allows `sgCollectPaths` to use a ctx with logging information on it, for filtering.
- Create `sgCollectOutputStream` struct to make handling output a bit easier. This needs to be closed if the command doesn't start (e.g. path not found). For testing, we need to wait for the output to be closed before reading it. Because it is buffered output, it lasts a bit longer than `cmd.Wait()` and the race detector is tripped. 
- Add `sgCollect.stderr` / `sgCollect.stdout` to monitor errors from sgcollect output.
- Return `sgCollectBinary` as a []string{} so it can be triggered via `python sgcollect_info`. This is only needed for testing, in normal production this will be a single executable that self extracts a python installation.

I think I'd like to set a build flag or environment variable so this test doesn't run as a part of standard dev runs, it is just too slow.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3179/

### DIFF
--- a/rest/sgcollect.go
+++ b/rest/sgcollect.go
@@ -262,10 +262,12 @@ func (o *sgCollectOutputStream) Close(ctx context.Context) {
 	select {
 	case <-o.stdoutDoneChan:
 	case <-time.After(5 * time.Second):
+		base.WarnfCtx(ctx, "sgcollect_info: timed out waiting for stdout processing to finish")
 	}
 	select {
 	case <-o.stderrDoneChan:
 	case <-time.After(5 * time.Second):
+		base.WarnfCtx(ctx, "sgcollect_info: timed out waiting for stderr processing to finish")
 	}
 }
 
--- a/rest/sgcollect.go
+++ b/rest/sgcollect.go
@@ -55,10 +55,10 @@ type sgCollectOutputStream struct {
 	stderrDoneChan   chan struct{}  // Channel to signal stderr processing completion
 }
 
+// sgCollect manages the state of a running sgcollect_info process.
 type sgCollect struct {
-	cancel           context.CancelFunc
+	cancel           context.CancelFunc // Function to cancel a running sgcollect_info process, set when status == sgRunning
 	status           *uint32
-	context          context.Context
 	sgPath           string    // Path to the Sync Gateway executable
 	sgCollectPath    []string  // Path to the sgcollect_info executable
 	sgCollectPathErr error     // Error if sgcollect_info path could not be determined
@@ -67,7 +67,7 @@ type sgCollect struct {
 }
 
 // Start will attempt to start sgcollect_info, if another is not already running.
-func (sg *sgCollect) Start(logFilePath string, ctxSerialNumber uint64, zipFilename string, params sgCollectOptions) error {
+func (sg *sgCollect) Start(ctx context.Context, logFilePath string, zipFilename string, params sgCollectOptions) error {
 	if atomic.LoadUint32(sg.status) == sgRunning {
 		return ErrSGCollectInfoAlreadyRunning
 	}
@@ -81,7 +81,7 @@ func (sg *sgCollect) Start(logFilePath string, ctxSerialNumber uint64, zipFilena
 		// If no output directory specified, default to the configured LogFilePath
 		if logFilePath != "" {
 			params.OutputDirectory = logFilePath
-			base.DebugfCtx(sg.context, base.KeyAdmin, "sgcollect_info: no output directory specified, using LogFilePath: %v", params.OutputDirectory)
+			base.DebugfCtx(ctx, base.KeyAdmin, "sgcollect_info: no output directory specified, using LogFilePath: %v", params.OutputDirectory)
 		} else {
 			// If LogFilePath is not set, and DefaultLogFilePath is not set via a service script, error out.
 			return errors.New("no output directory or LogFilePath specified")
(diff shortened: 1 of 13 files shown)


rest/sgcollect.go
```diff
@@ -212,6 +195,80 @@ func validateOutputDirectory(dir string) error {
...
+		base.WarnfCtx(ctx, "sgcollect_info: failed to close stderr pipe writer: %v", err)
+	}
+	err = o.stdoutPipeWriter.Close()
+	if err != nil {
+		base.WarnfCtx(ctx, "sgcollect_info: failed to close stdout pipe writer: %v", err)
+	}
+	// Wait for the goroutines to finish processing the output streams, or exit after 5 seconds.
+	select {
+	case <-o.stdoutDoneChan:
+	case <-time.After(5 * time.Second):
```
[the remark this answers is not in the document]
bbrks: Seems fine given context

tools/sgcollect.py
```diff
@@ -608,69 +621,60 @@ def make_config_tasks(
...
-            data = json.load(response)
-        except urllib.error.URLError as e:
-            print("WARNING: Unable to connect to Sync Gateway: {0}".format(e))
+    if not sg_url:
+        return None, None
+
+    try:
+        response = urlopen_with_basic_auth(expvar_url(sg_url), sg_username, sg_password)
+    except urllib.error.URLError as e:
+        print("WARNING: Unable to connect to Sync Gateway: {0}".format(e))
```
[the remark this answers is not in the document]
bbrks: `print("WARNING: ..."` is standard in sgcollect, for better or for worse

rest/sgcollect.go
```diff
@@ -90,69 +95,46 @@ func (sg *sgCollect) Start(logFilePath string, ctxSerialNumber uint64, zipFilena
...
-	go func() {
-		scanner := bufio.NewScanner(stdoutPipeReader)
-		for scanner.Scan() {
-			base.InfofCtx(sg.context, base.KeyAll, "sgcollect_info: %v", scanner.Text())
-		}
-		if err := scanner.Err(); err != nil {
-			base.ErrorfCtx(sg.context, "sgcollect_info: unexpected error: %v", err)
-		}
-	}()
+	base.InfofCtx(ctx, base.KeyAdmin, "sgcollect_info started with cmdline: %v", base.UD(cmdline))
```
bbrks: I know this was how it was before this PR, but let's remove the args for additional clarity

```
2025-06-24T14:28:26.672+01:00 [INF] HTTP: c:#001 POST /_sgcollect_info (as <ud>Administrator</ud> as ADMIN)
2025-06-24T14:28:26.673+01:00 [DBG] Admin+: c:SGCollect-001 sgcollect_info: no output directory specified, using LogFilePath: /tmp/sglog-resync-test
2025-06-24T14:28:26.677+01:00 [INF] Admin: c:SGCollect-001 sgcollect_info started with cmdline: [ <ud>/Users/benbrooks/dev/cb/sg/tools/sgcollect_info</ud> <ud>--sync-gateway-username</ud> <ud>Administrator</ud> <ud>--sync-gateway-password</ud> <ud>password</ud> <ud>--sync-gateway-url</ud> <ud>http://127.0.0.1:4985</ud> <ud>--sync-gateway-executable</ud> <ud>/Users/benbrooks/dev/cb/sg/sync_gateway</ud> <ud>/tmp/sglog-resync-test/sgcollectinfo-2025-06-24t132826-sg@192.168.68.210.zip</ud> ]
```

It's still worth printing out `zipPath` since that's a generated name and users would probably want to know where it will be

```suggestion
	base.InfofCtx(ctx, base.KeyAdmin, "sgcollect_info started with output zip: %v", base.UD(zipPath))
```

bbrks (changes requested):
One suggested change to log output from a `/_sgcollect_info` REST request, otherwise looks good and tested fine

rest/sgcollect_test.go
```diff
@@ -222,3 +226,62 @@ func TestSgcollectOptionsArgs(t *testing.T) {
 		})
 	}
 }
+
+func TestSGCollectIntegration(t *testing.T) {
```
bbrks: You've probably already seen - but something is a bit screwed up on Windows in this test. It eventually failed after 420 seconds

```go
2025-06-24T14:11:44.0631070Z 2025-06-24T14:10:15.092Z [INF] HTTP: c:#5360 GET /_sgcollect_info (as ADMIN)
2025-06-24T14:11:44.0631219Z 2025-06-24T14:10:17.092Z [INF] HTTP: c:#5361 GET /_sgcollect_info (as ADMIN)
2025-06-24T14:11:44.0631335Z     sgcollect_test.go:272: 
2025-06-24T14:11:44.0631645Z         	Error Trace:	D:/a/sync_gateway/sync_gateway/rest/sgcollect_test.go:276
2025-06-24T14:11:44.0632218Z         	            				C:/hostedtoolcache/windows/go/1.24.4/x64/src/runtime/asm_amd64.s:1700
2025-06-24T14:11:44.0632339Z         	Error:      	Not equal: 
2025-06-24T14:11:44.0632530Z         	            	expected: "stopped"
2025-06-24T14:11:44.0632720Z         	            	actual  : "running"
2025-06-24T14:11:44.0632832Z         	            	
2025-06-24T14:11:44.0632999Z         	            	Diff:
2025-06-24T14:11:44.0633170Z         	            	--- Expected
2025-06-24T14:11:44.0633314Z         	            	+++ Actual
2025-06-24T14:11:44.0633471Z         	            	@@ -1 +1 @@
2025-06-24T14:11:44.0633613Z         	            	-stopped
2025-06-24T14:11:44.0633746Z         	            	+running
2025-06-24T14:11:44.0633860Z     sgcollect_test.go:272: 
2025-06-24T14:11:44.0634159Z         	Error Trace:	D:/a/sync_gateway/sync_gateway/rest/sgcollect_test.go:272
2025-06-24T14:11:44.0634329Z         	Error:      	Condition never satisfied
2025-06-24T14:11:44.0634514Z         	Test:       	TestSGCollectIntegration
2025-06-24T14:11:44.0634736Z         	Messages:   	sgcollect_info did not stop running in time
2025-06-24T14:11:44.0634907Z 2025-06-24T14:10:19.091Z [INF] HTTP: c:#5362 GET /_sgcollect_info (as ADMIN)
2025-06-24T14:11:44.0635072Z 2025-06-24T14:10:19.092Z [INF] HTTP: c:#5363 DELETE /_sgcollect_info (as ADMIN)
2025-06-24T14:11:44.0635318Z 2025-06-24T14:10:19.092Z [INF] HTTP: t:TestSGCollectIntegration Closing HTTP Server: 127.0.0.1:54226
2025-06-24T14:11:44.0635555Z 2025-06-24T14:10:19.093Z [INF] HTTP: t:TestSGCollectIntegration Closing HTTP Server: 127.0.0.1:54227
2025-06-24T14:11:44.0635792Z 2025-06-24T14:10:19.093Z [INF] HTTP: t:TestSGCollectIntegration Closing HTTP Server: 127.0.0.1:54228
2025-06-24T14:11:44.0636042Z 2025-06-24T14:10:19.093Z [INF] HTTP: t:TestSGCollectIntegration Closing HTTP Server: 127.0.0.1:54225
2025-06-24T14:11:44.0636144Z --- FAIL: TestSGCollectIntegration (420.12s)
```

bbrks (changes requested):

tools/tasks.py
```diff
@@ -178,7 +178,7 @@ def execute(self, fp):
             p = subprocess.Popen(
                 self.command,
                 bufsize=-1,
-                stdin=subprocess.PIPE,
+                stdin=subprocess.DEVNULL,
```
bbrks: I was surprised to see this is not a change already made in cbcollect_info if it's actually required, and I am also not sure quite why this change?

Surely sgcollect_info was not completely broken on Windows up until this point?

The commit message isn't really indicative of the change being made. There's nothing Windows or CI specific about this change, and we're not closing stdin.

Will this change impact interactive password prompts in future? I guess maybe not since that's not stdin, and it's instead just a prompt from the program.

torcolvin: We close stdin right after launching this, but this makes sure it is already closed and is more correct. I wondered if there was a race condition. sgcollect isn't designed to collect any data from stdin for any diagnostic programs it runs.

https://github.com/couchbase/sync_gateway/actions/runs/15852522483/job/44689423355?pr=7601

Shows `wmic` running at:
```
2025-06-24T14:03:48.830Z [INF] c:SGCollect-5382 sgcollect_info: Computer system (wmic computersystem) - OK
```

And then the test runs until:

```
2025-06-24T14:10:39.220Z [INF] HTTP: c:#5592 GET /_sgcollect_info (as ADMIN)
```

which is 7 minutes. I guessed on two reasons for this:

1. wmic is slow - it would have been running `WindowsTask("Installed software", "wmic product get name, version"),` for 7 minutes, which seems weird, because that seems like it would normally be fast. That's why I thought it could be asking for something from stdin.
2. wmic is also known to be slow, when getting the event log. That's the most useful behavior it can do, and also the slowest. So this is expensive for the last 30 days, but on a github actions machine it doesn't stay running for very long. This is the equivalent of the systemd/dmesg logs for windows.

wmic is in fact deprecated and going to be removed - https://learn.microsoft.com/en-us/windows/win32/wmisdk/wmic

That said, I'm going to have another commit for moving the tests to a separate package so it doesn't bang up against the 10 minute timeout for the rest package. When running the tests via github, this will also increase parallelism since packages are run simultaneously.

bbrks (approved):

PR #7604, opened 2025-06-24
CBG-4687 have sgcollect_info use a token

torcolvin:
CBG-4687 have sgcollect_info use a token

- Created a field on handlers that marks if is sgcollect that indicates whether sgcollect will be called.
- Create a specific token for performing these operations
- Switch sgcollect code pass auth headers rather than explicit username/password

For the reviewer:

- Manual testing with real SG and /_sgcollect_info endpoint and sgcollect_info executable

This was done on top of https://github.com/couchbase/sync_gateway/pull/7601. Will turn into a non draft PR when that is merged.

## Pre-review checklist

- [x] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3181/


rest/handler.go
```diff
@@ -289,6 +292,18 @@ func (h *handler) invoke(method handlerMethod, accessPermissions []Permission, r
 	return method(h) // Call the actual handler code
 }
 
+// shouldCheckAdminRBAC returns true if the request needs to check the server for permissions to run
+func (h *handler) shouldCheckAdminRBAC() bool {
+	if h.sgcollect && h.server.sgcollect.hasValidToken(h.rq.Header) {
+		return false
```
bbrks: Is it worth detecting when a handler is called with an sgcollect token and the handler _doesn't_ have `h.sgcollect` set?

It indicates we have added a new endpoint to sgcollect but forgotten to set the handler option which seems like an easy thing to forget.

`AssertfCtx` should let us catch this at dev time via the existing sgcollect tests we have.

rest/sgcollect.go
```diff
@@ -157,6 +171,36 @@ func (sg *sgCollect) IsRunning() bool {
...
+}
+
+// hasValidToken checks if the provided headers contain a valid token for sgcollect_info.
+func (sg *sgCollect) hasValidToken(headers http.Header) bool {
+	auth := headers.Get("Authorization")
+	authPrefix := "SGCollect "
+	if !strings.HasPrefix(auth, authPrefix) {
+		return false
+	}
+	if time.Since(sg.tokenAge) > 30*time.Minute {
```
bbrks: Move this token timeout into a named const for clarity.

bbrks: I am wondering whether 30 minutes is long enough, or whether we'll actually get cases where `sgcollect_info` takes longer to run if there's a really big set of logs to process?

Do we have any information about how long we expect sgcollect_info to run for? Even locally it takes ~2 minutes to run.

There is probably data we could gather from previous collections to be sure this is long enough not to cause problems, but the easier alternative is to give us much more headroom to eliminate the chance of exceeding during a valid run, yet still have an upper-limit token lifespan (e.g. 6 hours) to avoid indefinite token lifespan in the case where we were unable to clear the token manually.

bbrks (changes requested):
A couple of small suggestions - otherwise looks good to me.

rest/sgcollect.go
bbrks: I would potentially like a third opinion on this - the token expiry was a suggestion which I think is a useful defensive measure in case we or sgcollect screws up and keeps a token active for a long period, but I'm not quite sure what the expiry value should be.

/cc @adamcfraser

bbrks: 12hr

rest/handler.go
torcolvin: Added this with tests at some complexity of the code.

rest/sgcollect.go
torcolvin: We decided on 12 hours offline.

rest/handler_test.go
```diff
@@ -151,3 +153,67 @@ func Benchmark_parseKeyspace(b *testing.B) {
...
+						}
+						require.True(t, metricsHandler.shouldCheckAdminRBAC(), "expected invalid token to still require auth")
+					} else {
+						require.False(t, adminHandler.shouldCheckAdminRBAC())
+						require.False(t, metricsHandler.shouldCheckAdminRBAC())
+
+					}
+					// with valid sgcollect token, but sgcollect on the handler is disabled
+					require.NoError(t, sc.sgcollect.createNewToken())
+					adminHandler = newHandler(sc, adminPrivs, adminServer, httptest.NewRecorder(), &http.Request{Header: http.Header{"Authorization": []string{"SGCollect invalid"}}}, handlerOptions{sgcollect: false})
```
bbrks: this should ideally be a valid token but the current implementation for the assertion doesn't care whether the token was valid or not, so it's fine

bbrks (dismissed):
LGTM

bbrks (approved):

PR #7605, opened 2025-06-24
CBG-3561 make --sync-gateway-password obsolete

torcolvin:
CBG-3561 make --sync-gateway-password obsolete

- `--sync-gateway-password=arg` is going to return `SG_PASSWORD_ERROR` explaining what to do.
- `SG_USERNAME` and `SG_PASSWORD` are supported for headless access
- remove hidden argument `--watch-stdin` since if you called this, `getpass` would fail since `--watch-stdin` would exit sgcollect_info as soon as the password was typed.
- always prompt for a password if we have a username.

Precedence (last wins):

- SG_USERNAME, --sync-gateway-username
- SG_PASSWORD, --sync-gateway-password

For future consideration, CBG-3562 which would require passing this variable or answering that we can't provide it. It would be good to try to get SG URL, and then ask for username/password. I thought this was scoped a bit outside of this review.

It would be more idiomatic to allow `--sync-gateway-password` if it has no arguments to prompt for a password, but this isn't possible using the long deprecated `optparse` cmdline parsing library.

## Pre-review checklist

### DIFF
--- a/tools/sgcollect.py
+++ b/tools/sgcollect.py
@@ -20,6 +20,7 @@ import os
 import pathlib
 import platform
 import re
+import shutil
 import ssl
 import sys
 import urllib.error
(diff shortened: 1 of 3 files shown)


tools/sgcollect.py
```diff
@@ -178,12 +185,15 @@ def create_option_parser():
...
     )
     parser.add_option(
         "--sync-gateway-password",
-        dest="sync_gateway_password",
-        help="Sync Gateway Admin API password ",
+        action="callback",
+        callback=sync_gateway_option_callback,
+        dest="do_not_use",
+        type=str,
+        help="Sync Gateway Admin API password. Can be specified with SG_PASSWORD env variable.",
```
bbrks: Since any use of this flag will error now, steer people away from it.

```suggestion
        help="Sync Gateway Admin API password. Should be specified with SG_PASSWORD env variable or input interactively.",
```

bbrks (dismissed):
LGTM - I left a suggestion about help text wording to more forcibly direct people towards env var or interactive prompt.

tools/sgcollect.py
bbrks: @torcolvin Do we want this change or leave it?

torcolvin:
@bbrks can you verify `sgcollect_info` output and `sgcollect_info -h` output?

bbrks (approved):

PR #7606, opened 2025-06-25
CBG-4716: remove repeated calls to was skipped in processEntry

gregns1:
CBG-4716

Looking at the perf profiling for the new skiplist, I see that we use a bit too much time in WasSkipped function. We do call this twice on skipped sequences, once to check for duplicates and once to mark as skipped. We should just be able to call this once per skipped entry. 

Also noticed info logging taking some time up so dropped some to debug given we only log sequence related info at debug usually. Plus moved some info logging out of loops. 

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3182/

### DIFF
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -436,16 +436,18 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent, docType DocumentType)
 	c.db.DbStats.Database().DCPReceivedCount.Add(1)
 
 	// If the doc update wasted any sequences due to conflicts, add empty entries for them:
-	for _, seq := range syncData.UnusedSequences {
-		base.InfofCtx(ctx, base.KeyCache, "Received unused #%d in unused_sequences property for (%q / %q)", seq, base.UD(docID), syncData.CurrentRev)
-		change := &LogEntry{
-			Sequence:       seq,
-			TimeReceived:   timeReceived,
-			CollectionID:   event.CollectionID,
-			UnusedSequence: true,
+	if len(syncData.UnusedSequences) > 0 {
+		for _, seq := range syncData.UnusedSequences {
+			change := &LogEntry{
+				Sequence:       seq,
+				TimeReceived:   timeReceived,
+				CollectionID:   event.CollectionID,
+				UnusedSequence: true,
+			}
+			changedChannels := c.processEntry(ctx, change)
+			changedChannelsCombined = changedChannelsCombined.Update(changedChannels)
 		}
-		changedChannels := c.processEntry(ctx, change)
-		changedChannelsCombined = changedChannelsCombined.Update(changedChannels)
+		base.InfofCtx(ctx, base.KeyCache, "Received unused sequences %d in unused_sequences property for (%q / %q)", syncData.UnusedSequences, base.UD(docID), syncData.CurrentRev)
 	}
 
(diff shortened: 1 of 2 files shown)


db/change_cache.go
```diff
@@ -436,16 +436,18 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent, docType DocumentType)
...
+				TimeReceived:   timeReceived,
+				CollectionID:   event.CollectionID,
+				UnusedSequence: true,
+			}
+			changedChannels := c.processEntry(ctx, change)
+			changedChannelsCombined = changedChannelsCombined.Update(changedChannels)
 		}
-		changedChannels := c.processEntry(ctx, change)
-		changedChannelsCombined = changedChannelsCombined.Update(changedChannels)
+		base.InfofCtx(ctx, base.KeyCache, "Received unused sequences %d in unused_sequences property for (%q / %q)", syncData.UnusedSequences, base.UD(docID), syncData.CurrentRev)
```
bbrks: I'd reformat this log just to move the list to the end, since it could get somewhat long. I think this also may make sense to put to Debug since any investigation around specific documents or sequences would require that anyway.

```suggestion
		base.DebugfCtx(ctx, base.KeyCache, "Received unused sequences in unused_sequences property for (%q / %q): %v", base.UD(docID), syncData.CurrentRev, syncData.UnusedSequences)
```

db/change_cache.go
```diff
@@ -491,6 +494,9 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent, docType DocumentType)
 				changedChannelsCombined = changedChannelsCombined.Update(changedChannels)
 			}
 		}
+		if len(seqsCached) > 0 {
+			base.InfofCtx(ctx, base.KeyCache, "Received deduplicated seqs %d in recent_sequences property for (%q / %q)", seqsCached, base.UD(docID), syncData.CurrentRev)
```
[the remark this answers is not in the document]
bbrks: Similar here - move the list to the end of the log message

```suggestion
			base.DebugfCtx(ctx, base.KeyCache, "Received deduplicated seqs in recent_sequences property for (%q / %q): %v", base.UD(docID), syncData.CurrentRev, seqsCached)
```

bbrks (changes requested):

db/change_cache.go
```diff
@@ -436,16 +436,18 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent, docType DocumentType)
...
 
 	// If the doc update wasted any sequences due to conflicts, add empty entries for them:
-	for _, seq := range syncData.UnusedSequences {
-		base.InfofCtx(ctx, base.KeyCache, "Received unused #%d in unused_sequences property for (%q / %q)", seq, base.UD(docID), syncData.CurrentRev)
-		change := &LogEntry{
-			Sequence:       seq,
-			TimeReceived:   timeReceived,
-			CollectionID:   event.CollectionID,
-			UnusedSequence: true,
+	if len(syncData.UnusedSequences) > 0 {
```
torcolvin, line 439: This if statement is not necessary, if `syncData.UnusedSequences` is empty, then the for loop will no-op

gregns1: Need this here so we don't log when the list is empty

bbrks (approved):

PR #7615, opened 2025-06-30
sort openapi properties for readabilty

torcolvin:
I'm open to dropping this PR but it is hard to find values in the configuration as is. The idea is to sort lexically, but sort deprecated properties at the top. The current sorting is closest to the structure of the go structs, which doesn't have meaning, and the output is sorted, due to json.Marshal rules.

By doing this, I actually found some other properties that should be deprecated but are not.

* I only applied this to two common field sections: `Startup-config` and `Database` since these are the biggest and most common, with the idea we could extend this later.
* I limited the scope of what it can detect, it won't sort `allOf`, `oneOf`, `anyOf`.

**Testing**:

* `yq --prettyPrint 'sort_keys(..)'` on each input file and diff the output after separating the `Database` and `Startup-config` sections.

** Implementation**:

Step 1: make sure the yaml roundtrips for all documents:

```
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "ruamel.yaml",
# ]
# ///
from ruamel.yaml import YAML
import argparse
import pathlib
import sys
import io


def main() -> None:
    parser = argparse.ArgumentParser(
        description="Validate ruamel will round trip without reformatting data."
    )
    parser.add_argument(
        "base_dir", type=pathlib.Path, help="Base directory of yaml to search"
    )
    args = parser.parse_args()

    for root, _, files in args.base_dir.walk():
        for file in files:
            if file.endswith(".yaml") or file.endswith(".yml"):
                roundtrip_yaml(pathlib.Path(root) / file)


def roundtrip_yaml(file_path: pathlib.Path) -> None:
    yaml = YAML()
    yaml.preserve_quotes = True
    yaml.indent(mapping=2, sequence=4, offset=2)
    yaml.width = 10000
    with file_path.open() as fp:
        data = yaml.load(fp)

    with io.BytesIO() as buf:
        yaml.dump(data, buf)
        out = buf.getvalue().decode("utf-8")
    if out != file_path.read_text():
        print(f"File {file_path} does not round trip correctly.", file=sys.stderr)
        print(out, file=sys.stdout)
        sys.exit(1)


if __name__ == "__main__":
    main()
```

Step 2: copy/paste sections to sort them.

```
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "ruamel.yaml",
# ]
# ///
from ruamel.yaml import YAML, CommentedMap
import argparse
import pathlib
import sys


def main() -> None:
    parser = argparse.ArgumentParser(
        description="Sort Redocly properties in a YAML file."
    )
    parser.add_argument(
        "file", type=pathlib.Path, help="Path to the YAML file to sort."
    )
    args = parser.parse_args()

    yaml = YAML()
    yaml.preserve_quotes = True
    yaml.indent(mapping=2, sequence=4, offset=2)
    yaml.width = 10000
    with args.file.open() as fp:
        data = yaml.load(fp)

    find_properties(data)

    yaml.dump(data, sys.stdout)


def find_properties(data: CommentedMap) -> None:
    for _, val in data.items():
        if isinstance(val, CommentedMap):
            find_properties(val)
            properties = val.get("properties")
            if properties:
                val["properties"] = sort_commented_map(properties)


def sort_commented_map(data: CommentedMap) -> CommentedMap:
    out = CommentedMap()
    non_deprecated_keys = [k for k, v in data.items() if not v.get("deprecated")]
    deprecated_keys = [k for k, v in data.items() if v.get("deprecated")]
    for k in sorted(non_deprecated_keys):
        out[k] = data[k]
    for k in sorted(deprecated_keys):
        out[k] = data[k]
    return out


if __name__ == "__main__":
    main()
```


torcolvin:
I still think that this would make docs more readable, but I forgot jsoniter doesn't sort map keys so we wouldn't see sorted output, I was testing standard library encoding/json.

bbrks (approved):

PR #7617, opened 2025-07-04
DOC-13103-updated the description for bootstrap.server

iamfortune:
[DOC-13103](https://jira.issues.couchbase.com/browse/DOC-13103)

Describe your PR here...
- I updated the description for bootstrap.server, made it more in-depth like the legacy. 
- 

I didn't find a style guide for updating these, let me know if there's one.  

Screenshot 2025-07-04 at 2 50 23 pm

[DOC-13103]: https://couchbasecloud.atlassian.net/browse/DOC-13103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -2042,7 +2042,76 @@ Startup-config:
           type: string
           default: 10s
         server:
-          description: Couchbase Server connection string/URL.
+          description: |
+            Couchbase Server connection string/URL for bootstrap configuration.
+            
+            The connection string should only reference Couchbase Server Data (KV) nodes. 
+            Using other node types (Query, Index, Analytics, or Search nodes) may cause 
+            connection failures and unreliable operation.
+            
+            **Connection String Format**
+            Sync Gateway supports the ability to specify multiple hosts in the configuration. 
+            Sync Gateway supports both the couchbase:// and http:// schemes for specifying 
+            connection endpoints.
+
+            Sync Gateway also supports SSL in the connection to Couchbase Server; use the 
+            couchbases:// scheme for this. As with the Couchbase Server SDKs, the https:// 
+            scheme is not supported.
+
+            **Examples of valid server values for IPv4:**
+            - `couchbase://host1`
+            - `couchbases://host1`
+            - `couchbase://host1,host2`
+            - `couchbase://host1:11210,host2`
+            - `couchbases://host1:11207,host2`
+            - `http://host1:8091`
+            - `http://host1,host2:8091`
+            - `http://foo:bar@host1:8091`
+
+            **Examples of valid server values for IPv6:**
+            - `http://[2001:db8::8811]:8091` // single node IPv6 - http scheme with default server port
+            - `couchbases://[2001:db8::8811]` // single node SSL IPv6 - default port (omitted)
+            - `couchbase://[2001:db8::8811],[2001:db8::8822]:888` // node1 default port, node2 port 888
+
+            **Port Requirements**
+            When using the couchbase:// or couchbases:// schemes, the port is not required, 
+            but if specified should be the external/internal bucket ports (defaults are 11210 
+            or 11207 respectively). Attempting to use the admin ports (8091/18091) will result 
+            in a startup error.
+
+            **Alternate Addresses**
+            On startup, Sync Gateway will try each hostname that is provided until it is able 
+            to connect successfully.
+
+            By default, if a remote cluster has an external address set, then when SG connects 
+            it will apply a heuristic to determine whether to choose between external or default 
+            (internal) addresses.
+
+            The choice is based on the host names supplied in the connection string:
+            - Sync Gateway uses external networking only when none of the supplied host names match any 
+              of Couchbase Server's internal node addresses, and an external address is defined.
+            - In all other cases Sync Gateway uses the default (internal) networking.
+
+            However, it is possible to override this behavior by adding a network parameter 
+            to the connection string.
+
+            The network parameter can be:
+            - `auto`: this is the default value if no parameter is provided. In this case the 
+              heuristic described above is applied to determine the address used; so effectively 
+              there is no override.
+            - `external`: to always force use of the external address
+            - `default`: to always force use of the internal address
+
+            Example: `"server": "couchbases://my-cbs-server?network=default"`
+
+            Will force the connection to ignore any alternative external addresses configured 
+            on the Couchbase Server node.
+
+            **Lost Connections**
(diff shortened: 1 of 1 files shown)


docs/api/components/schemas.yaml
```diff
@@ -2042,7 +2042,76 @@ Startup-config:
           type: string
           default: 10s
         server:
-          description: Couchbase Server connection string/URL.
+          description: |
+            Couchbase Server connection string/URL for bootstrap configuration.
+            
+            The connection string should only reference Couchbase Server Data (KV) nodes. 
+            Using other node types (Query, Index, Analytics, or Search nodes) may cause
```
torcolvin:
```suggestion
            Using other node types (Query, Index, Analytics, or Search nodes) is not supported.
```

docs/api/components/schemas.yaml
```diff
@@ -2042,7 +2042,76 @@ Startup-config:
...
+          description: |
+            Couchbase Server connection string/URL for bootstrap configuration.
+            
+            The connection string should only reference Couchbase Server Data (KV) nodes. 
+            Using other node types (Query, Index, Analytics, or Search nodes) may cause 
+            connection failures and unreliable operation.
+            
+            **Connection String Format**
+            Sync Gateway supports the ability to specify multiple hosts in the configuration. 
+            Sync Gateway supports both the couchbase:// and http:// schemes for specifying
```
torcolvin:
```suggestion
            Sync Gateway supports both the couchbase:// and couchbases:// schemes for specifying 
```

https://docs.couchbase.com/go-sdk/current/howtos/managing-connections.html#connection-strings

docs/api/components/schemas.yaml
```diff
@@ -2042,7 +2042,76 @@ Startup-config:
...
+            Example: `"server": "couchbases://my-cbs-server?network=default"`
+
+            Will force the connection to ignore any alternative external addresses configured
+            on the Couchbase Server node.
+
+            **Lost Connections**
+            If the connection to Couchbase Server is lost during normal operations, Sync Gateway
+            will automatically re-connect to another node in the cluster. During that re-connection
+            period, the Sync Gateway will appear offline and documents will not be replicated
+            to mobile clients.
```
torcolvin: I eventually realized that this is from https://docs.couchbase.com/sync-gateway/current/configuration-properties-legacy.html#databases-this_db-server

The comments I'm going to make here apply to the other text identically. Much of this information comes from

https://docs.couchbase.com/go-sdk/current/howtos/managing-connections.html

```suggestion
            **Connection String Format**
            Sync Gateway supports the ability to resolve DNS SRV records for alternate hostnames, or specifying multiple hostnames explicitly.
            Sync Gateway supports both the couchbases:// for TLS and couchbase:// schemes for  insecure connection. 

			The supported schemes match that of the Couchbase Server SDKs.

            **Examples of valid server values:**
            - `couchbases://nodeA.example.com`
            - `couchbase://nodeA.example.com`
            - `couchbases://nodeA.example.com,nodeB.example.com`
            - `couchbase://nodeA.example.com,nodeB.example.com`
            - `couchbases://nodeA.example.com:1234,nodeB.example.com:1234`
            - `couchbase://nodeA.example.com:1234,nodeB.example.com:1234`
            - `couchbases://127.0.0.1`
            - `couchbase://127.0.0.1`

            On startup, Sync Gateway will try all hostnames until it is able
            to connect successfully.

            **Alternate Addresses**
            
            If your Couchbase Server cluster is running in a containerized, port mapped, or otherwise NATd environment like Docker or Kubernetes, Sync Gateway might need more information to connect to the cluster. In many cases the client is able to automatically select the correct set of addresses. 
            
            If the detection heuristic fails in your environment
            By default, if a remote cluster has an external address set, then when SG connects
            it will apply a heuristic to determine whether to choose between external or default
            (internal) addresses.

            The choice is based on the host names supplied in the connection string:
            - Sync Gateway uses external networking only when none of the supplied host names match any
              of Couchbase Server's internal node addresses, and an external address is defined.
            - In all other cases Sync Gateway uses the default (internal) networking.

            However, it is possible to override this behavior by adding a network parameter
            to the connection string.

            The network parameter can be:
            - `external`: Force the use alternate addresses of Couchbase Server. Used when Sync Gateway should not share network used by Couchbase Server internally.
            - `default`: Do not allow use of alternate addresses of Couchbase Server. Used when Sync Gateway and Couchbase Server are on the same network.

            Example: `"server": "couchbases://my-cbs-server?network=default"`

            Will force the connection to ignore any alternative external addresses configured
            on the Couchbase Server node.

            **Lost Connections**
            If the connection to Couchbase Server is lost during normal operations, Sync Gateway
            will automatically re-connect to another node in the cluster.
```

I changed some of the text copy here, but here are the bullet points.

- http:// connections haven't been supported since Sync Gateway 1.x and they definitely should be be included anywhere
- couchbases:// (TLS) should be called out before couchbase:// because Sync Gateway will in fact not connect to `couchbase://` without also calling `bootstrap.use_tls_server: false` explicitly. I don't think we need to call out `use_tls_server` parameter, but similarly, I'd not highlight an examples that avoid use of TLS.
- 11207 (TLS) and 11210 (non TLS) are the default ports for Couchbase Server https://docs.couchbase.com/server/current/install/install-ports.html#ports-listed-by-communication-path see `client-to-node` and if they are used, they don't need to be listed explicitly.
- I didn't think it was needed to call out ipv6 addresses, given that I do not believe those were in a valid format. I did explicitly call out DNS or IP in the examples using example.com
- I don't think there is a good way to link this into the SDK documentation but I think emphasizing that using DNS srv records and pointing to a single DNS hostname is more typical, see https://docs.couchbase.com/go-sdk/current/howtos/managing-connections.html#using-dns-srv-records.
- The heuristic for choosing `network=external` or `network=default` has the wrong description. I omitted the description because I do not know what this is. I would say it would be a misconfiguration of Couchbase Server if they do not set this up correctly and they need to specify this. This is maintained by the couchbase go SDK team, so I'd prefer to omit rather than misspeak.

torcolvin (commented):
I did not preview the suggested changes, but I notice some markdown oddities. To preview this document locally:

```
npx @redocly/cli@latest preview-docs admin
```

and expand response on http://127.0.0.1:8080/#tag/Server/operation/get__config

iamfortune:
I have updated these, let me know if you've got any more feedback or if I missed anything.

torcolvin (approved):

PR #7626, opened 2025-07-11
Add a test for import feed panic

torcolvin:
Take or leave this PR, it is a followup to https://github.com/couchbase/sync_gateway/commit/92cfc7c76ceb3d5578fd56761601b0833d302497

When reading the code setting: `shouldPersistCheckpoint = false // Don't persist checkpoint if we panicked` in the defer has no effect since the code was changed from a named to an unnamed return. When I read this code, I was sure that this code would not function because it was no longer a named return. I wrote a test, and then was surprised it did in fact work. It works since it will return the default value of `bool` after a recover.


bbrks (approved):
👍 it's a bit more obvious with this additional comment

PR #7631, opened 2025-07-15
make XDCR stop integration test more robust

torcolvin:
- When goxdcr.log is rotating, it occasionally does not exist. Do a retry when getting last timestamp.
- Use 4 buckets in bucket pool for XDCR tests as this speeds things up, since the next set of buckets will be ready as buckets are always allocated in 2s.
- Fix gross misuse of `CreateSleeperFunc`

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3215 (xdcr package only)

### DIFF
--- a/base/main_test_bucket_pool.go
+++ b/base/main_test_bucket_pool.go
@@ -64,7 +64,9 @@ type TestBucketPool struct {
 	skipCollections bool
 	// numCollectionsPerBucket is the number of collections to create in each bucket
 	numCollectionsPerBucket int
-	useExistingBucket       bool
+	// numBuckets is the number of buckets managed by the pool
+	numBuckets        int
+	useExistingBucket bool
 
 	// skipMobileXDCR may be true for older versions of Couchbase Server that don't support mobile XDCR enhancements
 	skipMobileXDCR bool
@@ -79,6 +81,7 @@ type TestBucketPoolOptions struct {
 	RequireXDCR             bool // Test buckets will be performing XDCR, requires Server > 7 for integration test robustness
 	ParallelBucketInit      bool
 	NumCollectionsPerBucket int      // setting this value in main_test.go will override the default
+	DefaultNumBuckets       int      // setting this value will indicate the default value for the number of buckets to create in the pool, SG_TEST_NUM_BUCKETS
 	TeardownFuncs           []func() // functions to be run after Main is completed but before standard teardown functions run
 }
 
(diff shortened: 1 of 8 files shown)


base/main_test_bucket_pool.go
```diff
@@ -79,6 +81,7 @@ type TestBucketPoolOptions struct {
 	RequireXDCR             bool // Test buckets will be performing XDCR, requires Server > 7 for integration test robustness
 	ParallelBucketInit      bool
 	NumCollectionsPerBucket int      // setting this value in main_test.go will override the default
+	DefaultNumBuckets       int      // setting this value will indicate the default value for the number of buckets to create in the pool, SG_TEST_NUM_BUCKETS
```
bbrks: Did you mean `SG_TEST_BUCKET_POOL_SIZE`? I can't see what this env var is referring to

xdcr/main_test.go
```diff
@@ -19,6 +19,11 @@ import (
 
 func TestMain(m *testing.M) {
 	ctx := context.Background() // start of test process
-	tbpOptions := base.TestBucketPoolOptions{MemWatermarkThresholdMB: 2048, RequireXDCR: true}
+	tbpOptions := base.TestBucketPoolOptions{
+		MemWatermarkThresholdMB: 2048,
+		RequireXDCR:             true,
+		DefaultNumBuckets:       4, // use 4 buckets for testing since each test will use 2 buckets for XDCR
```
bbrks: This does make sense and I'd be open to bumping the default to 4 globally if that makes more sense than per-package overrides?
Less complexity and would even allow ISGR tests to run quicker since those also require 2 buckets per-test.

I think 3 was settled on only as a a reasonably low value - at the time - there were deminishing returns beyond a few buckets. That might be different now.

topologytest/main_test.go
```diff
@@ -27,7 +27,12 @@ func TestMain(m *testing.M) {
 		base.SkipTestMain(m, "Tests are disabled for Couchbase Server by default, to enable set %s=true environment variable", base.TbpEnvTopologyTests)
 		return
 	}
-	tbpOptions := base.TestBucketPoolOptions{MemWatermarkThresholdMB: 2048}
+	tbpOptions := base.TestBucketPoolOptions{
+		MemWatermarkThresholdMB: 2048,
+		NumCollectionsPerBucket: 1, // tests use one named collection
+		RequireXDCR:             true,
+		DefaultNumBuckets:       4, // XDCR tests run in pairs, so having four buckets means the tests run faster
```
bbrks: see other comment

bbrks (changes requested):

base/main_test_bucket_pool.go
torcolvin: Removed this.

torcolvin (commented):
Update to remove the optional setting, I think 4 is fine too, especially since https://github.com/couchbase/sync_gateway/commit/44d6aa5b208f18ff9e247d85eab83a9755406a16 means we don't have to wait for all the buckets to spin up.

base/util.go
```diff
@@ -583,6 +582,11 @@ func CreateSleeperFunc(maxNumAttempts, timeToSleepMs int) RetrySleeper {
 
 }
 
+// CreateLinearSleeperFunc creates a linear sleeper function that sleeps up to maxNumAttempts, sleeping timeToSleep each attempt.
+func CreateLinearSleeperFunc(totalTime, timeToSleep time.Duration) RetrySleeper {
```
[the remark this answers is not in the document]
torcolvin: The API of this function doesn't lend itself to error reporting since this is used inside of another function.

bbrks (approved):

PR #7633, opened 2025-07-18
Add Test Coverage: TestCollectionsPublicChannel to ensure changes feed coverage for syncfn granted public docs

bbrks:
Ensures that docs being put in the public channel via sync function in a named collection are accessible over the changes feed as expected (they are)

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3221/

### DIFF
--- a/rest/api_collections_test.go
+++ b/rest/api_collections_test.go
@@ -213,6 +213,7 @@ func TestCollectionsPublicChannelViaSyncFn(t *testing.T) {
 	require.Len(t, changesResults.Results, 3)
 	t.Logf("changes results: %s", changesResults.Summary())
 
+	// since we're filtering only to channel A, no docs are returned
 	changesResults = rt.WaitForChanges(1, "/{{.keyspace}}/_changes?filter="+base.ByChannelFilter+"&channels=A", username, false)
 	require.Len(t, changesResults.Results, 1)
 	t.Logf("changes results: %s", changesResults.Summary())
--- a/rest/api_collections_test.go
+++ b/rest/api_collections_test.go
@@ -117,7 +117,7 @@ func TestCollectionsPutDocInKeyspace(t *testing.T) {
 func TestCollectionsPublicChannel(t *testing.T) {
 	const (
 		username = "alice"
-		password = "pass"
+		password = RestTesterDefaultUserPassword
 	)
 
 	rt := NewRestTester(t, &RestTesterConfig{
@@ -155,6 +155,9 @@ func TestCollectionsPublicChannel(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, 1, alldocsresp.TotalRows)
 	assert.Len(t, alldocsresp.Rows, 1)
+
+	changesResp := rt.GetChangesOneShot(t, "keyspace", 0, username, 2)
+	t.Logf("changes resp: %s", changesResp.BodyBytes())
 }
 
 // TestNoCollectionsPutDocWithKeyspace ensures that a keyspace can't be used to insert a doc on a database not configured for collections.
--- a/rest/api_collections_test.go
+++ b/rest/api_collections_test.go
@@ -188,6 +188,7 @@ func TestCollectionsPublicChannelViaSyncFn(t *testing.T) {
 	RequireStatus(t, resp, http.StatusOK)
 
 	pathPrivate := "/{{.keyspace}}/docstillpublic"
+	// The "channels" property in the document is ignored because the SyncFn routes all documents to the "!" channel.
 	resp = rt.SendAdminRequest(http.MethodPut, pathPrivate, `{"channels":["a"]}`)
 	RequireStatus(t, resp, http.StatusCreated)
 	resp = rt.SendUserRequestWithHeaders(http.MethodGet, pathPrivate, "", nil, username, password)
--- a/rest/api_collections_test.go
+++ b/rest/api_collections_test.go
@@ -156,8 +156,66 @@ func TestCollectionsPublicChannel(t *testing.T) {
 	assert.Equal(t, 1, alldocsresp.TotalRows)
 	assert.Len(t, alldocsresp.Rows, 1)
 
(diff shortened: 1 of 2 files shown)


rest/api_collections_test.go
```diff
@@ -155,6 +155,69 @@ func TestCollectionsPublicChannel(t *testing.T) {
...
+	// The "channels" property in the document is ignored because the SyncFn routes all documents to the "!" channel.
+	resp = rt.SendAdminRequest(http.MethodPut, pathPrivate, `{"channels":["a"]}`)
+	RequireStatus(t, resp, http.StatusCreated)
+	resp = rt.SendUserRequestWithHeaders(http.MethodGet, pathPrivate, "", nil, username, password)
+	RequireStatus(t, resp, http.StatusOK)
+
+	resp = rt.SendUserRequestWithHeaders(http.MethodGet, "/{{.keyspace}}/_all_docs?include_docs=true", "", nil, username, password)
+	RequireStatus(t, resp, http.StatusOK)
+	t.Logf("all docs resp: %s", resp.BodyBytes())
+	var alldocsresp struct {
```
torcolvin: this is `allDocsResponse` https://github.com/couchbase/sync_gateway/blob/5108469f708de4bf7b1c1f7e63e1887409c1a176/rest/bulk_api.go#L28

rest/api_collections_test.go
```diff
@@ -155,6 +155,69 @@ func TestCollectionsPublicChannel(t *testing.T) {
...
+
+// TestCollectionsPublicChannelViaSyncFn ensures that docs routed to the public channel via the sync function are accessible by a user with no other access.
+func TestCollectionsPublicChannelViaSyncFn(t *testing.T) {
+	const (
+		username = "alice"
+		password = RestTesterDefaultUserPassword
+	)
+
+	rt := NewRestTester(t, &RestTesterConfig{
+		SyncFn: `function(doc){channel('!');}`,
```
torcolvin: Do you think this test is different than above?

If I understand correctly, when collections are enabled, `SyncFn: channels.DocChannelsSyncFunction` does execute javascript because that is no longer the implicit sync function.

torcolvin (dismissed):

rest/api_collections_test.go
bbrks: Eh not really, thinking about it some more. Maybe I'll just add the channel-filter tests/assertions up above instead.

The non-JS vs. JS thing is a subtle implementation detail - the test itself is still assigning to `!` via JS via the doc property, as you say.

rest/api_collections_test.go
```diff
@@ -132,6 +132,13 @@ func TestCollectionsPublicChannel(t *testing.T) {
 	})
 	defer rt.Close()
 
+	// grant 'b' so we can check a non-public but accessible doc as well
+	if rt.GetDatabase().OnlyDefaultCollection() {
+		require.NoError(t, rt.SetAdminChannels(username, rt.GetDatabase().Name, "b"))
+	} else {
+		require.NoError(t, rt.SetAdminChannels(username, rt.GetSingleKeyspace(), "b"))
+	}
```
torcolvin: These actually are the same code, since `rt.GetSingleKeyspace()` will resolve to `db` if `SG_TEST_USE_DEFAULT_COLLECTION=true`. The other way to write this code which I'd be inclined to do for new tests is to remove the code to add a user in dbconfig but use `rt.CreateUser("alice", []string{"b"})`.

This is just informational for writing future tests, I'm happy to leave it.


```suggestion
	// grant 'b' so we can check a non-public but accessible doc as well
	require.NoError(t, rt.SetAdminChannels(username, rt.GetSingleKeyspace(), "b"))
```

rest/api_collections_test.go
```diff
@@ -144,17 +151,35 @@ func TestCollectionsPublicChannel(t *testing.T) {
...
 	err := json.Unmarshal(resp.BodyBytes(), &alldocsresp)
 	require.NoError(t, err)
-	assert.Equal(t, 1, alldocsresp.TotalRows)
-	assert.Len(t, alldocsresp.Rows, 1)
+	assert.Equal(t, 2, alldocsresp.TotalRows)
+	assert.Len(t, alldocsresp.Rows, 2)
+
+	// user doc and accessible doc
+	changesResults := rt.WaitForChanges(3, "/{{.keyspace}}/_changes", username, false)
+	require.Len(t, changesResults.Results, 3)
```
torcolvin: `WaitForChanges` already does this assertion https://github.com/couchbase/sync_gateway/blob/497e38bfcb8ba017288ff7e348d6943d92cf9532/rest/utilities_testing.go#L908
```suggestion
```

rest/api_collections_test.go
```diff
@@ -144,17 +151,35 @@ func TestCollectionsPublicChannel(t *testing.T) {
...
+	t.Logf("changes results: %s", changesResults.Summary())
+
+	// explicitly filtering for public only
+	changesResults = rt.WaitForChanges(2, "/{{.keyspace}}/_changes?filter="+base.ByChannelFilter+"&channels=!", username, false)
+	require.Len(t, changesResults.Results, 2)
```
torcolvin:
```suggestion
```
As above, `WaitForChanges` already does this assertion.

rest/api_collections_test.go
```diff
@@ -144,17 +151,35 @@ func TestCollectionsPublicChannel(t *testing.T) {
...
+	t.Logf("changes results: %s", changesResults.Summary())
+
+	// since we're filtering only to channel 'a' and we don't have access - don't expect any docs
+	changesResults = rt.WaitForChanges(1, "/{{.keyspace}}/_changes?filter="+base.ByChannelFilter+"&channels=a", username, false)
+	require.Len(t, changesResults.Results, 1)
```
torcolvin:
```suggestion
```
As above `WaitForChanges` already does this.

torcolvin (dismissed):
Take or leave nit comments.

torcolvin (approved):

PR #7637, opened 2025-07-21
CBG-4280 Improve test assertions for attachments

torcolvin:
CBG-4280 Improve test assertions for attachments

This is improving the assertions for refactoring,

The plan of refactoring is to consolidate `Document.GlobalSyncData.Attachments` and `Document.DocAttachments` and rename `Document.SyncData.Attachments` -> `Document.SyncData.AttachmentsPre4dot0`. When referring to a document in memory, `Document.Attachments` should be the canonical way to do it, since `Document.GlobalSyncData` is an embedded struct.

The goal of attachment handling in 4.0 is to be able to read attachments in different formats:

- inline body (pre xattrs)
- `_sync.attachments`
- `_globalSync.attachments_meta`

When the document is written to the collection then only place attachment data should be is `globalSync.attachments_meta`.

- `/ks/doc` or `/_bulk_get` -> `_attachments` as a key inline to the body. Sometimes this appears to have the full contents of the attachment.
- blip message -> `_attachments` as a key inline to the body, with stub always.

These test changes should not be changing the code, just adding full assertions for the types. The reviewer should validate this.

I made use of `DocAttachment` data structure and changed `AttachmentMap` to use the concrete `DocAttachment` rather than a pointer so I could do `require.Equal(t, AttachmentMap{}, m)`. I can't see when this would cause problems or result in an extra copies because right now this is only used in test code. Aspirationally a future ticket could imagine using `DocAttachment` or similar structure to be `globalSync.attachments_meta` rather than `map[string]any`.

The one piece I'm worried about is making `Data` json marshalable and whether I've changed production code. That's why I verify the full contents of the body. I've considered whether to allow this be unmarshable but not marshable, but that seemed to blow up the scope of this PR.

# Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3223/

### DIFF
--- a/db/attachment.go
+++ b/db/attachment.go
@@ -46,8 +46,12 @@ type updatedAttachment struct {
 // updatedAttachments holds the user facing attachment name and raw contents
 type updatedAttachments map[string]updatedAttachment
 
-// A map of keys -> DocAttachments.
-type AttachmentMap map[string]*DocAttachment
+// AttachmentMap can represent:
+//   - _sync.attachments
+//   - _globalSync.attachments_meta
+//   - inline body `_attachments` from blip rev messages
+//   - inline body `_attachments`from pre Sync Gateway 2.5
+type AttachmentMap map[string]DocAttachment
 
 // A struct which models an attachment.  Currently only used by test code, however
 // new code or refactoring in the main codebase should try to use where appropriate.
@@ -58,7 +62,7 @@ type DocAttachment struct {
 	Revpos      int    `json:"revpos,omitempty"`
 	Stub        bool   `json:"stub,omitempty"`
 	Version     int    `json:"ver,omitempty"`
-	Data        []byte `json:"-"` // tell json marshal/unmarshal to ignore this field
+	Data        []byte `json:"data,omitempty"` // tell json marshal/unmarshal to ignore this field
 }
 
 // ErrAttachmentTooLarge is returned when an attempt to attach an oversize attachment is made.
--- a/db/attachment.go
+++ b/db/attachment.go
@@ -53,8 +53,10 @@ type updatedAttachments map[string]updatedAttachment
 //   - inline body `_attachments`from pre Sync Gateway 2.5
 type AttachmentMap map[string]DocAttachment
 
-// A struct which models an attachment.  Currently only used by test code, however
-// new code or refactoring in the main codebase should try to use where appropriate.
+// DocAttachment models an attachment. This could be attachment metadata or the full body of an attachment.
+//   - `_sync.attachments` : Data field should not be present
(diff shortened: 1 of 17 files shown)


rest/attachment_test.go
```diff
@@ -1130,19 +1127,18 @@ func TestBasicAttachmentRemoval(t *testing.T) {
...
+				ContentType: "content/type",
+				Digest:      "sha1-CTJaowVFZ4ozgmvBageTH9w+OKU=",
+				Length:      38,
+				Revpos:      2,
+			},
+		}, attachments)
 
 		// Retrieve the key used for internal attachment storage and retrieval.
 		attKey := retrieveAttachmentKey(t, docID, attName)
-		require.NotEmpty(t, attKey)
```
gregns1, line 1145: Any reason this assertion is dropped?

gregns1: Scratch that I see the `RequireDocNotFound` this is passed into will fail if this is empty.

rest/attachment_test.go
```diff
@@ -2958,20 +2918,20 @@ func TestAttachmentMigrationToGlobalXattrOnUpdate(t *testing.T) {
...
-	require.NoError(t, base.JSONUnmarshal(xattrs[base.GlobalXattrName], &globalXattr))
-
-	assert.Nil(t, bucketSyncData.Attachments)
-	assert.NotNil(t, globalXattr.GlobalAttachments)
-	attMeta := globalXattr.GlobalAttachments["camera.txt"].(map[string]interface{})
-	assert.Equal(t, float64(20), attMeta["length"])
+	require.Empty(t, db.GetRawSyncXattr(t, ds, docID).Attachments)
+	revpos := 1
+	if !base.IsEnterpriseEdition() {
+		revpos = 2 // this is not intentional, and maybe should be fixed. However, neither CBL nor SG since 3.0 use revpos, so a fix is low priority
```
gregns1: Is there a ticket for this to link here?

db/attachment.go
```diff
@@ -58,7 +62,7 @@ type DocAttachment struct {
 	Revpos      int    `json:"revpos,omitempty"`
 	Stub        bool   `json:"stub,omitempty"`
 	Version     int    `json:"ver,omitempty"`
-	Data        []byte `json:"-"` // tell json marshal/unmarshal to ignore this field
+	Data        []byte `json:"data,omitempty"` // tell json marshal/unmarshal to ignore this field
```
gregns1: Should we update this comments now?

torcolvin: Updated comments, thanks.

gregns1 (dismissed):

gregns1 (approved):

PR #7640, opened 2025-07-22
CBG-4734: fix for race when removing a value that is being loaded into rev cache

gregns1:
CBG-4734

Added a new function to remove only from rev lookup and have extra logic around eviction to enure we are evicting values we should be

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3235/

### DIFF
--- a/db/revision_cache_lru.go
+++ b/db/revision_cache_lru.go
@@ -520,7 +520,7 @@ func (rc *LRURevisionCache) removeFromCacheByCV(ctx context.Context, docID strin
 	elem := element.Value.(*revCacheValue)
 
 	// we can only remove this value if the value has finished loading from the bucket
-	_, _ = base.RetryLoop(ctx, "remove from revision cache by rev", func() (shouldRetry bool, err error, _ any) {
+	_, _ = base.RetryLoop(ctx, "remove from revision cache by cv", func() (shouldRetry bool, err error, _ any) {
 		if !elem.canEvict.Load() {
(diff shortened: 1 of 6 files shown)


db/revision_cache_lru.go
```diff
@@ -518,6 +518,16 @@ func (rc *LRURevisionCache) removeFromCacheByCV(ctx context.Context, docID strin
 	}
 	// grab the revid key from the value to enable us to remove the reference from the rev lookup map too
 	elem := element.Value.(*revCacheValue)
+
+	// we can only remove this value if the value has finished loading from the bucket
+	_, _ = base.RetryLoop(ctx, "remove from revision cache by cv", func() (shouldRetry bool, err error, _ any) {
```
bbrks: Is this just working around the fact that we may need to acquire the `revCacheValue` lock when we're removing? The Load already holds a lock for the value - I've not completely thought this through so take this comment with a pinch of salt

A retry loop waiting for this to be false just smells like a lock is missing somewhere though for better synchronization.

gregns1: Yeah essentially we want to avoid holding the value lock and the rev cache lock at the same time to avoid any deadlocking similar to the deadlock we hit a while back. So this function already holds the rev cache lock so we added this atomic bool for eviction/removal previously which avoids us having to acquire value lock

bbrks (approved):
LGTM

PR #7643, opened 2025-07-22
Use WaitUntilReady retry strategy for test pool

torcolvin:
Followup to https://github.com/couchbase/sync_gateway/pull/7642. I don't know why this only hits MasterIntegration and not Sync Gateway Integration Tests but I could reproduce locally.

### DIFF
--- a/base/main_test_cluster.go
+++ b/base/main_test_cluster.go
@@ -39,7 +39,11 @@ type tbpCluster struct {
 
 // newTestCluster returns a cluster based on the driver used by the defaultBucketSpec.
 func newTestCluster(ctx context.Context, clusterSpec CouchbaseClusterSpec) (*tbpCluster, error) {
-	agent, err := NewClusterAgent(ctx, clusterSpec, TestClusterReadyTimeout)
+	agent, err := NewClusterAgent(ctx, clusterSpec,
+		CouchbaseClusterWaitUntilReadyOptions{
+			Timeout:       TestClusterReadyTimeout,
+			RetryStrategy: gocbcore.NewBestEffortRetryStrategy(nil),
+		})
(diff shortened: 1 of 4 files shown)


bbrks:
I'm not sure I understand what this change actually does.

Everywhere `RetryStrategy` is set is `nil`, either explicitly or implicitly?

base/main_test_cluster.go
```diff
@@ -39,7 +39,11 @@ type tbpCluster struct {
 
 // newTestCluster returns a cluster based on the driver used by the defaultBucketSpec.
 func newTestCluster(ctx context.Context, clusterSpec CouchbaseClusterSpec) (*tbpCluster, error) {
-	agent, err := NewClusterAgent(ctx, clusterSpec, TestClusterReadyTimeout)
+	agent, err := NewClusterAgent(ctx, clusterSpec,
+		CouchbaseClusterWaitUntilReadyOptions{
+			Timeout:       TestClusterReadyTimeout,
+			RetryStrategy: gocbcore.NewBestEffortRetryStrategy(nil),
```
bbrks, line 45: nvm, I see - `nil` is a parameter...

bbrks (approved):

torcolvin:
> I'm not sure I understand what this change actually does.
> 
> Everywhere `RetryStrategy` is set is `nil`, either explicitly or implicitly?

The default retry strategy is a fast failure retry strategy, from https://github.com/couchbase/gocbcore/blob/e5f9929c8e4dcd4a4d6dcfda751f0748db71cced/agent.go#L182

It is potentially not correct to have a fast failure retry strategy for the cluster agent. We do not have one for the bucket connections, but it is out of scope to change this for this PR.

PR #7647, opened 2025-07-25
CBG-4280 consolidate Attachment on document

torcolvin:
CBG-4280 consolidate Attachment on document

- merge `Document.SyncData.Attachments`, `Document.GlobalSyncData.GlobalAttachments`, and `Document.DocAttachments` into `Document._globalSync.Attachments`
- `Document.SyncData.Attachments` becomes `Document.SyncData.AttachmentsPre4dot0` and is used for unmarshalling, but should be cleared before writing a document
- rename `GlobalSyncData.GlobalAttachments` to `GlobalSyncData.Attachments`

I feel pretty comfortable with the amount of testing that this has, especially after I did https://github.com/couchbase/sync_gateway/commit/7ded49843d5813c4cec983e783a10028f1fa418b

I've considered whether to make `GlobalSyncData` not embedded into `Document`, but ultimately I decided not to for the ease of referring to `Document.Attachments` in the code. This makes the in memory data structure s

Things I don't love:

- `Document.SyncData.AttachmentsPre4dot0` has to exist to unmarshal SyncData, but we need to be sure to clear this when we are writing this data. This is handled in `unmarshalDocumentWithXattrs`. An alternate approach to what I did would be to create a `MarshalJSON` function to make sure that this field never got marshalled, but I couldn't think of a good way to do this.
- `Document.MarshalJSON` has to re-add attachments into SyncData into order to have `ImportDoc` work correctly. This is already noted as weird in this TODO https://github.com/couchbase/sync_gateway/blob/feebe5af5583520ff4be57424ca95b38d1923be4/db/import.go#L78. I can't think of a place that this is called outside `ImportDoc` because you almost never want to marshal metadata into an inline body. I think this hack in `Document.MarshalJSON` could be removed if `ImportDoc` got refactored. See `TestLegacyAttachmentMigrationToGlobal` for the only code that tests this pathway.

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3227/


db/crud.go
```diff
@@ -1037,18 +1037,13 @@ func (c *DatabaseCollectionWithUser) MigrateAttachmentMetadata(ctx context.Conte
...
 			return base.RedactErrorf("Failed to Unmarshal global sync data when attempting to migrate sync data attachments to global xattr with id: %s. Error: %v", base.UD(docID), err)
 		}
-		// add the sync data attachment metadata to global xattr
-		for i, v := range syncData.Attachments {
-			globalData.GlobalAttachments[i] = v
-		}
-	} else {
-		globalData.GlobalAttachments = syncData.Attachments
 	}
+	globalData.Attachments = mergeAttachments(syncData.AttachmentsPre4dot0, globalData.Attachments)
```
gregns1, line 1039: NIT: could we do some clean up of this function if its going to be used here so the code doesn't look so pre 2.5 attachment specific?

db/document.go
```diff
@@ -87,24 +87,24 @@ func (m *MetadataOnlyUpdate) PreviousCAS() uint64 {
...
+	Channels            channels.ChannelMap  `json:"channels,omitempty"`
+	Access              UserAccessMap        `json:"access,omitempty"`
+	RoleAccess          UserAccessMap        `json:"role_access,omitempty"`
+	Expiry              *time.Time           `json:"exp,omitempty"`                     // Document expiry.  Information only - actual expiry/delete handling is done by bucket storage.  Needs to be pointer for omitempty to work (see https://github.com/golang/go/issues/4357)
+	Cas                 string               `json:"cas"`                               // String representation of a cas value, populated via macro expansion
+	Crc32c              string               `json:"value_crc32c"`                      // String representation of crc32c hash of doc body, populated via macro expansion
+	Crc32cUserXattr     string               `json:"user_xattr_value_crc32c,omitempty"` // String representation of crc32c hash of user xattr
+	TombstonedAt        int64                `json:"tombstoned_at,omitempty"`           // Time the document was tombstoned.  Used for view compaction
+	AttachmentsPre4dot0 AttachmentsMeta      `json:"attachments,omitempty"`
```
gregns1: Should we add a comment around this now to state that this is only used for pre migrated docs/pre migrated inline sync docs?

gregns1 (commented):
LGTM, just couple of nit pick comments

db/document.go
```diff
@@ -1127,6 +1130,11 @@ func (doc *Document) UnmarshalJSON(data []byte) error {
 }
 
 func (doc *Document) MarshalJSON() (data []byte, err error) {
+	// The calling code in ImportDoc will marshal the document since it takes a body with _sync inline.
+	// Make sure that _sync._attachments is present in the document body since _globalSync._attachments_meta
+	// can not be seen in an inline body.
+	doc.SyncData.AttachmentsPre4dot0 = doc.Attachments()
```
adamcfraser, line 1171: How sure are we that the on-demand import case in ImportDoc is the only use case for this function, given that it's also the standard override for marshalling of Document?  Were you able to confirm that there aren't any other places we end up calling json.Marshal() on a *Document (including things like passing a *Document to gocb, etc)?

db/document.go
```diff
@@ -215,13 +215,13 @@ type Document struct {
 	Deleted        bool
 	DocExpiry      uint32
 	RevID          string
-	DocAttachments AttachmentsMeta
```
adamcfraser, line 253: I believe that this set of properties (Deleted, DocExpiry, RevID, DocAttachments) was originally added to facilitate passing this set of properties around in cases where SyncData wasn't fully populated yet (e.g. incoming writes).   I think it's reasonable to consider these properties technical debt (they've never really belonged in Document), so it's possible that what you're doing in this PR is addressing that technical debt for DocAttachments specifically.  

However, I'm still a little bit concerned regarding cases where the attachments were previously being written to this "transport" helper property, and we're now writing them directly to _globalSync.Attachments (e.g. the doc_api.go change).  Are we certain there aren't any cases (possibly cas retry cases) where stale values populated in this way are being retained in _globalSync and getting persisted when they shouldn't?

db/document.go
torcolvin: I can't be 100% sure that this is the only caller, but if this isn't, then this preserves the pre-existing behavior. I feel more confident about this because I can't think of a case where we'd serialize `_sync` inline. The only case might be in the GET or `_bulk_get` endpoints, but they use `GetMutable1xBody()`. The raw endpoint used to use this before https://github.com/couchbase/sync_gateway/pull/7649

Backing up old revisions for delta sync up the body + attachments and doesn't serialize the full `_sync` body.

If this was called, it could be in test code to write the old style inline sync docs, but I evaluated that in https://github.com/couchbase/sync_gateway/commit/7ded49843d5813c4cec983e783a10028f1fa418b and we typically write the full old style document.

db/document.go
torcolvin: I feel reasonably comfortable with this. It used to be more complicated when `_sync` could ever be inline, but now it is not. I am not going to evaluate the non xattr use cases.

1. PUT use cases, pull `_attachments` out of an inline body to add to a SyncData that is being constructed. This is unambiguously the right thing to do when serializing a document.
    1. `DatabaseCollection.Put` https://github.com/couchbase/sync_gateway/blob/feebe5af5583520ff4be57424ca95b38d1923be4/db/crud.go#L1097
    2. `DatabaseCollection.PutWithExistingRev` https://github.com/couchbase/sync_gateway/blob/feebe5af5583520ff4be57424ca95b38d1923be4/db/crud.go#L1517 
    3. `processRev` https://github.com/couchbase/sync_gateway/blob/feebe5af5583520ff4be57424ca95b38d1923be4/db/blip_handler.go#L1304
    4. replicate2 put https://github.com/couchbase/sync_gateway/blob/696430cbbeb47eca5dcac1868cb077dfa3248e09/rest/doc_api.go#L593
2. conflict resolution, these are similar to PUT cases    
    1. `DatabaseCollection.resolveDocWithLocalWins` https://github.com/couchbase/sync_gateway/blob/feebe5af5583520ff4be57424ca95b38d1923be4/db/crud.go#L1517
    2. `DatabaseCollection.docMerge` https://github.com/couchbase/sync_gateway/blob/feebe5af5583520ff4be57424ca95b38d1923be4/db/crud.go#L1824
3.  Sync Function Dry Run, used by non-yet used diagnostic API https://github.com/couchbase/sync_gateway/blob/feebe5af5583520ff4be57424ca95b38d1923be4/db/crud.go#L1813
4. `importDoc` https://github.com/couchbase/sync_gateway/blob/696430cbbeb47eca5dcac1868cb077dfa3248e09/db/import.go#L330 In this case, the old code would zero the `_sync.attachments` property if it was a delete, but not directly modify the DocAttachments

When writing a document, `doc.SyncData.Attachments` was always used: https://github.com/couchbase/sync_gateway/blob/696430cbbeb47eca5dcac1868cb077dfa3248e09/db/document.go#L1306

The only time `doc.DocAttachments` is used is when `storeAttachments` is called, in this case only trying to add the new documents? I think the case it is trying to deal with is https://github.com/couchbase/sync_gateway/blob/main/db/attachment.go#L138-L141 but in this case, it doesn't care about the difference of the values that are passed in, so I can't see why this would matter.

adamcfraser: Discussed this further offline, and with the switch to allow_conflicts=false in 4.0.0, there shouldn't be a non-conflicting CAS retry case where the changes to DocAttachments would cause any problems.  Even if DocAttachments (mapped to _globalSync.Attachments) were updated as part of the write process, a CAS conflict with an underlying change to _globalSync.Attachments in the bucket would be a conflict state, and not permitted.

adamcfraser (approved):

PR #7648, opened 2025-07-25
Cleanup jenkins scripts

torcolvin:
The motivation to do this was that `SG_TEST_BUCKET_POOL` is explicitly set by this script and I think we should move to having these default values determined by the code and jenkins could optionally overwrite them.

This is a low priority change to review.

This change is no-op, there are no functional changes to the script except not requiring setting `SG_TEST_BUCKET_POOL_SIZE` and 
`SG_TEST_BUCKET_POOL_DEBUG` before you call this script. Note, these are always currently set by the jenkins pipelines anyway, so the values would pass through. So this is also not a functional change with the way the jenkins jobs are set up.

- Added shellcheck action because I use this locally and it helps when refactoring.
    - Excluded all non test scripts from checking, but it will
- Fixed the scripts to make shellcheck happy
    - quoting
    - switch comparison to consistent for bash, `if [ "${SG_TEST_X509:-}" == "true" -a "${COUCHBASE_SERVER_PROTOCOL}" != "couchbases" ]; then` to use `&&` instead of `-a`
    - switch `GO_TEST_FLAGS` to use a bash array so that it could be correctly quoted to appease shellcheck
- add REQUIRED_VARS check in the case I have to run the script locally, so it runs faster
- Play with -x / +x so that it doesn't output all the test `[` `]` and is easy to read. I used the output to debug things but I think there's enough information to debug. This might just add noise.
- Remove `GO_MODULES` and `XATTRS` since `GO_MODULES` isn't relevant since 3.1 and `XATTRS` is no longer relevant as of 4.0. This script now is always sourced from the repo, so I'm not worried about changes.
- Remove unused `DOCKER_CBS_ROOT_DIR` since `/workspace` is the only volume used for cbcollects.
- Remove `\t` from `stdbuf -oL grep -a -E '(--- (FAIL|PASS|SKIP):|github.com/couchbase/sync_gateway(/.+)?\t|TEST: |panic: )'` since grep always complainted about it, and it didn't do what it said. The scripts still find the values we need for the `ok` or `failed` for finishing package.

## Testing

- ran `./jenkins-integration-build.sh -m` locally to simulate `MasterIntegration` job.
- use `./start_server.sh` locally (I already use this to create testing)

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3228/


jenkins-integration-build.sh
```diff
@@ -83,38 +77,38 @@ if [ "${TEST_DEBUG:-}" == "true" ]; then
...
 fi
 
 if [ "${SG_TEST_PROFILE_FREQUENCY:-}" == "true" ]; then
     export SG_TEST_PROFILE_FREQUENCY=${SG_TEST_PROFILE_FREQUENCY}
 fi
 
 if [ "${RUN_WALRUS}" == "true" ]; then
-    set +e
+    set +e -x
+    set -x +e
```
bbrks: one of these redundant?

jenkins-integration-build.sh
```diff
@@ -44,75 +54,63 @@ export GOPRIVATE=github.com/couchbaselabs/go-fleecedelta
...
+    GO_TEST_FLAGS+=(-test.timeout="${PACKAGE_TIMEOUT}")
 else
     echo "Defaulting package timeout to ${DEFAULT_PACKAGE_TIMEOUT}"
-    GO_TEST_FLAGS="${GO_TEST_FLAGS} -test.timeout=${DEFAULT_PACKAGE_TIMEOUT}"
+    GO_TEST_FLAGS+=(-test.timeout="${DEFAULT_PACKAGE_TIMEOUT}")
 fi
 
 if [ "${DETECT_RACES:-}" == "true" ]; then
-    GO_TEST_FLAGS="${GO_TEST_FLAGS} -race"
+    GO_TEST_FLAGS=(-race)
```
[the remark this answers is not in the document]
bbrks: I think this comment was an accurate one FWIW

integration-test/start_server.sh
```diff
@@ -54,36 +50,42 @@ if [[ -n "${AMAZON_LINUX_2}" ]]; then
...
 else
+    set -x # Output all executed shell commands
     # single node
     docker run --rm -d --name ${SG_TEST_COUCHBASE_SERVER_DOCKER_NAME} --volume "${WORKSPACE_ROOT}:/workspace" -p 8091-8097:8091-8097 -p 9102:9102 -p 9123:9123 -p 11207:11207 -p 11210:11210 -p 11211:11211 -p 18091-18097:18091-18097 -p 19102:19102 "${COUCHBASE_DOCKER_IMAGE_NAME}"
 fi
 
 # Test to see if Couchbase Server is up
 # Each retry min wait 5s, max 10s. Retry 20 times with exponential backoff (delay 0), fail at 120s
-curl --retry-all-errors --connect-timeout 5 --max-time 10 --retry 20 --retry-delay 0 --retry-max-time 120 'http://127.0.0.1:8091'
+docker exec couchbase curl --fail --silent --retry-all-errors --connect-timeout 5 --max-time 10 --retry 20 --retry-delay 0 --retry-max-time 120 'http://127.0.0.1:8091'
```
bbrks, line 81: why `docker exec curl` here and not on line 88?

bbrks (changes requested):
few comments and the EditorConfig conflict from #7787 (Sorry)

integration-test/start_server.sh
torcolvin: I probably hit this with accidental debugging, I'd rather have it call curl from inside the container than outside given we don't know the network constraints.

bbrks (approved):

PR #7649, opened 2025-07-25
CBG-4767: Make `_raw` endpoint return actual state of persisted doc

bbrks:
Switch `/db/_raw/doc` endpoint to fetch actual persisted document contents. This is a pre-req for 4.0 because we need to include `_vv` and some other non-`_sync` xattrs in this API.

The xattr keys are returned nested underneath an `_xattrs` property returned by `_raw` (i.e. `_sync` has moved down one level) - the intention here is to know the difference between inline sync data and xattr-stored sync data.

Screenshot 2025-07-29 at 10 06 07

### Side-effects (on-demand import)

Doing the bucket-level fetch has a (good) side-effect of not performing an on-demand import, so we can inspect the state of the document and xattrs in the bucket without changing it.

### Documentation

This API is a diagnostic API and is already marked in our docs as subject to change so I think we're OK doing this, especially in 4.0

The docs are intentionally left vague about schema, since it's not committed - only describing the high-level xattr keys and generic objects, not the contents of each. There's a detailed real-world example provided as a reference.

https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/couchbase/sync_gateway/CBG-4767/docs/api/admin.yaml#tag/Document/operation/get_keyspace-_raw-docid

Screenshot 2025-07-29 at 10 04 08

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3232/


rest/utilities_testing_resttester.go
```diff
@@ -491,3 +496,46 @@ func (rt *RestTester) PutDocWithAttachment(docID string, body string, attachment
 	}
 	return rt.PutDocDirectly(docID, rawBody)
 }
+
+type RawDocResponse struct {
+	Xattrs  RawDocXattrs `json:"_xattrs"`
+	Deleted bool         `json:"_deleted"` // FIXME: no longer true in 4.0
```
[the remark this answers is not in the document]
bbrks: I'll address this FIXME

rest/admin_api.go
```diff
@@ -1423,78 +1424,123 @@ func (h *handler) handleDeleteDB() error {
...
-	docid := h.PathVar("docid")
 
-	includeDoc, includeDocSet := h.getOptBoolQuery("include_doc", true)
-	redact, _ := h.getOptBoolQuery("redact", false)
-	salt := h.getQuery("salt")
+	var (
+		docID                     = h.PathVar("docid")
+		includeDoc, includeDocSet = h.getOptBoolQuery("include_doc", true)
+		redact                    = h.getBoolQuery("redact")
+		redactSalt                = h.getQuery("redactSalt")
```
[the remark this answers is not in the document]
bbrks: Shit - this was an accidental refactor - I wanted to change only the variable name. Will fix.

rest/admin_api.go
```diff
@@ -1423,78 +1424,123 @@ func (h *handler) handleDeleteDB() error {
...
-		if salt == "" {
-			salt = uuid.New().String()
+	xattrsObject := []byte(base.EmptyDocument)
+
+	// inject xattrs as top-level properties in the doc - this isn't *quite* representative of the actual stored doc,
+	// since they are outside the doc body as separate objects, but we'll stamp a marker value in each xattr object to tell us whether it was injected or inline in the original doc body.
+	for _, k := range xattrKeys {
+		v := xattrValues[k]
+		if !redact {
+			xattrsObject, err = base.InjectJSONProperties(xattrsObject, base.KVPair{Key: k, Val: json.RawMessage(v)})
```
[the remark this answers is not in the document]
bbrks: This is intentional (I want to emit `null` values for missing xattrs)

rest/utilities_testing_resttester.go
```diff
@@ -491,3 +496,45 @@ func (rt *RestTester) PutDocWithAttachment(docID string, body string, attachment
...
+}
+
+func (t *RawDocXattrs) UnmarshalJSON(p []byte) error {
+	if err := json.Unmarshal(p, &t.RawDocXattrsWellKnown); err != nil {
+		return err
+	}
+	return json.Unmarshal(p, &t.RawDocXattrsOthers)
+}
+
+var _ json.Unmarshaler = &RawDocXattrs{}
```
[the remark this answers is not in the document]
bbrks: test struct - only expect unmarshalling

rest/adminapitest/admin_api_test.go
```diff
@@ -1798,34 +1798,27 @@ func TestRawRedaction(t *testing.T) {
...
 	// Test redacted
-	body = map[string]interface{}{}
 	res = rt.SendAdminRequest("GET", "/{{.keyspace}}/_raw/testdoc?redact=true&include_doc=false", ``)
-	err = base.JSONUnmarshal(res.Body.Bytes(), &body)
-	assert.NoError(t, err)
-	syncData = body[base.SyncPropertyName]
-	require.NotNil(t, syncData)
-	assert.NotEqual(t, map[string]interface{}{"achannel": nil}, syncData.(map[string]interface{})["channels"])
-	assert.NotEqual(t, []interface{}{[]interface{}{"achannel"}}, syncData.(map[string]interface{})["history"].(map[string]interface{})["channels"])
+	rawResponseStr = res.Body.String()
```
torcolvin, line 1820:
```suggestion
    rest.RequireStatus(t, res, http.StatusOK)
	rawResponseStr = res.BodyString()
```

rest/adminapitest/admin_api_test.go
```diff
@@ -1798,34 +1798,27 @@ func TestRawRedaction(t *testing.T) {
...
+	require.NotContains(t, rawResponseStr, "achannel")
+	require.NotContains(t, rawResponseStr, "foo")
+	require.NotContains(t, rawResponseStr, "bar")
 
 	// Test include doc false doesn't return doc
-	body = map[string]interface{}{}
 	res = rt.SendAdminRequest("GET", "/{{.keyspace}}/_raw/testdoc?include_doc=false", ``)
-	assert.NotContains(t, res.Body.String(), "foo")
+	rawResponseStr = res.Body.String()
```
torcolvin, line 1831:
```suggestion
    rest.RequireStatus(t, http.StatusOK, res)
	rawResponseStr = res.BodyString()
```

rest/adminapitest/admin_api_test.go
```diff
@@ -1845,20 +1838,24 @@ func TestRawTombstone(t *testing.T) {
 
 	resp = rt.SendAdminRequest(http.MethodGet, "/{{.keyspace}}/_raw/"+docID, ``)
```
torcolvin, line 1863:
```suggestion
	resp = rt.SendAdminRequest(http.MethodGet, "/{{.keyspace}}/_raw/"+docID, ``)
	rest.RequireStatus(t, resp, http.StatusOK)
```

rest/adminapitest/admin_api_test.go
```diff
@@ -1845,20 +1838,24 @@ func TestRawTombstone(t *testing.T) {
...
+	rawResponseStr := resp.Body.String()
+	assert.NotContains(t, rawResponseStr, `"_id":"`+docID+`"`)
+	assert.NotContains(t, rawResponseStr, `"_rev":"`+version.RevTreeID+`"`)
+	assert.Contains(t, rawResponseStr, `"foo":"bar"`)
+	assert.NotContains(t, rawResponseStr, `"_deleted":true`)
 
 	// Delete the doc
 	deletedVersion := rt.DeleteDoc(docID, version)
 
 	resp = rt.SendAdminRequest(http.MethodGet, "/{{.keyspace}}/_raw/"+docID, ``)
```
torcolvin, line 1875:
```suggestion
	resp = rt.SendAdminRequest(http.MethodGet, "/{{.keyspace}}/_raw/"+docID, ``)
	rest.RequireStatus(t, resp, http.StatusOK)
```

rest/adminapitest/admin_api_test.go
```diff
@@ -1798,34 +1798,27 @@ func TestRawRedaction(t *testing.T) {
 
 	// Test redact being disabled by default
 	res = rt.SendAdminRequest("GET", "/{{.keyspace}}/_raw/testdoc", ``)
-	var body map[string]interface{}
-	err := base.JSONUnmarshal(res.Body.Bytes(), &body)
-	assert.NoError(t, err)
-	syncData := body[base.SyncPropertyName]
-	assert.Equal(t, map[string]interface{}{"achannel": nil}, syncData.(map[string]interface{})["channels"])
-	assert.Equal(t, []interface{}{[]interface{}{"achannel"}}, syncData.(map[string]interface{})["history"].(map[string]interface{})["channels"])
+	rawResponseStr := res.Body.String()
```
torcolvin, line 1803:
```suggestion
    rest.RequireStatus(t, res, http.StatusOK)
	rawResponseStr := res.Body.String()
```

rest/adminapitest/admin_api_test.go
```diff
@@ -3016,7 +3014,7 @@ func TestDbOfflineConfigLegacy(t *testing.T) {
 	// Get config values before taking db offline
 	resp = rt.SendAdminRequest("GET", "/db/_config", "")
 	require.Equal(t, http.StatusOK, resp.Code)
-	dbConfigBeforeOffline := string(resp.BodyBytes())
+	dbConfigBeforeOffline := resp.Body.String()
```
torcolvin: Prefer `resp.BodyString()` and `resp.BodyBytes()` otherwise a single read like `resp.Body.String()` will clear from subsequent callers.

```suggestion
	dbConfigBeforeOffline := resp.BodyString()
```

rest/adminapitest/admin_api_test.go
```diff
@@ -3025,7 +3023,7 @@ func TestDbOfflineConfigLegacy(t *testing.T) {
 	// Check offline config matches online config
 	resp = rt.SendAdminRequest("GET", "/db/_config", "")
 	assert.Equal(t, http.StatusOK, resp.Code)
-	assert.Equal(t, dbConfigBeforeOffline, string(resp.BodyBytes()))
+	assert.Equal(t, dbConfigBeforeOffline, resp.Body.String())
```
torcolvin:
```suggestion
	assert.Equal(t, dbConfigBeforeOffline, resp.BodyString())
```

rest/importtest/import_test.go
```diff
@@ -108,6 +109,12 @@ func TestImportFeedWithRecursiveSyncFunction(t *testing.T) {
 	base.RequireWaitForStat(t, rt.GetDatabase().DbStats.SharedBucketImportStats.ImportErrorCount.Value, 2)
 }
 
+func getRawDocWithOnDemandImport(rt *rest.RestTester, docID string) *rest.TestResponse {
+	// trigger on-demand import - since 4.0 _raw will not do this
+	rt.TriggerOnDemandImport(docID)
+	return rt.SendAdminRequest("GET", "/{{.keyspace}}/_raw/"+docID, "")
+}
```
torcolvin: I'd be tempted to change the code for 2 reasons:

1. Check for status in this function since it seems like the caller would always have to do the check
2. Unmarshal document for fewer lines of code in the callers, and avoid accidentally unmarshalling into the same output struct without clearing it

```suggestion
// getRawDocWithOnDemandImport does an on demand import and the
func getRawDocWithOnDemandImport(rt *rest.RestTester, docID string) *rest.RawDocResponse {
	// trigger on-demand import - since 4.0 _raw will not do this
	rt.TriggerOnDemandImport(docID)
	resp := rt.SendAdminRequest("GET", "/{{.keyspace}}/_raw/"+docID, "")
	rest.RequireStatus(t, resp, http.StatusOK)
	var output rest.RawDocResponse
	require.NoError(t, base.JSONUnmarshal(resp.BodyBytes(), &output))
	return output
}
```

rest/importtest/import_test.go
```diff
@@ -911,10 +913,11 @@ func TestMigrateLargeInlineRevisions(t *testing.T) {
 	// Get raw to retrieve metadata, and validate bodies have been moved to xattr
 	rawResponse := rt.SendAdminRequest("GET", "/{{.keyspace}}/_raw/"+key, "")
 	assert.Equal(t, 200, rawResponse.Code)
-	var doc treeDoc
+	rawResponse.DumpBody()
```
```suggestion
```

rest/importuserxattrtest/rawdoc_test.go
```diff
@@ -41,7 +41,11 @@ func TestUserXattrsRawGet(t *testing.T) {
 	require.NoError(t, err)
 
 	ctx := rt.Context()
-	_, err = rt.GetSingleDataStore().UpdateXattrs(ctx, docKey, 0, cas, map[string][]byte{xattrKey: base.MustJSONMarshal(t, "val")}, nil)
+	_, err = rt.GetSingleDataStore().UpdateXattrs(ctx, docKey, 0, cas,
+		map[string][]byte{
+			xattrKey: base.MustJSONMarshal(t, "val"),
+		},
+		nil)
 	assert.NoError(t, err)
```
torcolvin:
```suggestion
	require.NoError(t, err)
```

rest/admin_api.go
```diff
@@ -1423,78 +1424,123 @@ func (h *handler) handleDeleteDB() error {
...
+			return base.HTTPErrorf(http.StatusBadRequest, "redact and include_doc cannot be true at the same time. "+
+				"If you want to redact you must specify include_doc=false")
+		}
+		// if include_doc was not explicitly set, and we want to redact, then automatically set include_doc to false
+		includeDoc = false
+	} else {
+		if redactSalt != "" {
+			return base.HTTPErrorf(http.StatusBadRequest, "redactSalt can only be used when redact=true")
+		}
 	}
```
torcolvin: nit: collapse if

```suggestion
	} else if redactSalt != "" {
			return base.HTTPErrorf(http.StatusBadRequest, "redactSalt can only be used when redact=true")
   }
```

rest/admin_api.go
```diff
@@ -1423,78 +1424,123 @@ func (h *handler) handleDeleteDB() error {
...
 
-	if redact && !includeDocSet {
-		includeDoc = false
+	// collect the set of xattrs to fetch that Sync Gateway is interested in.
+	xattrKeys := []string{
```
torcolvin: I would also like to put revseqno here.

`Database.syncGlobalSyncMouRevSeqNoAndUserXattrKeys()` might be correct, but actually that doesn't include revseqno! 

RevSeqNo will be really helpful for debugging XDCR

I think using a common function means if we added a future xattr that we might be more likely to pick it up?

rest/admin_api.go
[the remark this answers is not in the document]
torcolvin: I think this is a worth adding a comment for, I also wondered about the loop. I think it's also worth stating in the api docs that this will be null if the property is not found.

rest/admin_api.go
```diff
@@ -1423,78 +1424,123 @@ func (h *handler) handleDeleteDB() error {
...
+	// inject xattrs as top-level properties in the doc - this isn't *quite* representative of the actual stored doc,
+	// since they are outside the doc body as separate objects, but we'll stamp a marker value in each xattr object to tell us whether it was injected or inline in the original doc body.
+	for _, k := range xattrKeys {
+		v := xattrValues[k]
+		if !redact {
+			xattrsObject, err = base.InjectJSONProperties(xattrsObject, base.KVPair{Key: k, Val: json.RawMessage(v)})
+			if err != nil {
+				return base.HTTPErrorf(http.StatusInternalServerError, "couldn't inject xattr %q into response: %s", k, err)
+			}
+		} else {
```
torcolvin: Will this write non redactable xattrs like mou if redact is true?

rest/admin_api.go
```diff
@@ -1423,78 +1424,123 @@ func (h *handler) handleDeleteDB() error {
...
+		// try to extract a revision from the sync data
+		docCurrentRev := "unknown"
+		if syncData, ok := xattrValues[base.SyncXattrName]; ok {
+			var sdRev struct {
+				Rev channels.RevAndVersion `json:"rev"`
+			}
+			if err := json.Unmarshal(syncData, &sdRev); err != nil {
+				return base.HTTPErrorf(http.StatusInternalServerError, "couldn't unmarshal rev from SyncData for doc %q: %s", base.UD(docID), err)
+			}
+			docCurrentRev = sdRev.Rev.RevTreeID
```
torcolvin, line 1484: This is almost `db.revOnlySyncData`. I'm fine with this, but I had to look up why this was not using `CurrentRev` using `RevTreeID` (becuase it's not sync data).

rest/doc_api_test.go
```diff
@@ -126,11 +126,11 @@ func TestDocumentNumbers(t *testing.T) {
 
 			// Check channel assignment
 			getRawResponse := rt.SendAdminRequest("GET", fmt.Sprintf("/{{.keyspace}}/_raw/%s?redact=false", test.name), "")
-			var rawResponse RawResponse
+			var rawResponse RawDocResponse
```
torcolvin, line 129:
```suggestion
            RequireStatus(t, getRawResponse, http.StatusOK)
			var rawResponse RawDocResponse
```

rest/doc_api_test.go
```diff
@@ -126,11 +126,11 @@ func TestDocumentNumbers(t *testing.T) {
...
 			require.NoError(ts, base.JSONUnmarshal(getRawResponse.Body.Bytes(), &rawResponse))
 			log.Printf("raw response: %s", getRawResponse.Body.Bytes())
-			assert.Equal(ts, 1, len(rawResponse.Sync.Channels))
-			assert.True(ts, HasActiveChannel(rawResponse.Sync.Channels, test.expectedFormatChannel), fmt.Sprintf("Expected channel %s was not found in document channels (%s)", test.expectedFormatChannel, test.name))
+			assert.Equal(ts, 1, len(rawResponse.Xattrs.Sync.Channels))
+			assert.True(ts, HasActiveChannel(rawResponse.Xattrs.Sync.Channels, test.expectedFormatChannel), fmt.Sprintf("Expected channel %s was not found in document channels (%s)", test.expectedFormatChannel, test.name))
```
torcolvin: I've thought about adding an assertion to `base/util_testing.go`:

`RequireMapKeysEqual` with the implementation:

`require.ElementsEqual(t, expectedValues, slices.Collect(maps.Keys(actual)), "Actual map: %#+v", actual)`

then you would know what the actual values where whereas `HasActiveChannel` doesn't tell you what the keys are. The difference in `HasActiveChannel` is that it would tell you if the `* ChannelRemoval` was null.

rest/utilities_testing_resttester.go
```diff
@@ -66,6 +66,11 @@ func (rt *RestTester) GetDoc(docID string) (DocVersion, db.Body) {
 	return DocVersion{RevTreeID: *r.RevID}, body
 }
 
+func (rt *RestTester) TriggerOnDemandImport(docID string) {
+	// Trigger on-demand import for the document, don't care about response
```
torcolvin: Adding a docstring allows this to be picked up by editors.

```suggestion
// TriggerOnDemandImport will use the REST API to trigger on an demand import via GET. This function intentionally does not check error codes in case the document does not exist or is invalid to be imported.
func (rt *RestTester) TriggerOnDemandImport(docID string) {
	// Trigger on-demand import for the document, don't care about response
```

Is there a use for not checking the error code? Because if this has an error code, it doesn't get imported.

There's mildly different code for GET `OnDemandImportForGet` and PUT `OnDemandImportForPut` which is why I noted this. I can't think of a reason that the GET version would fail. The `PUT` version can fail if there are conflicts, it would do an import of the document in the bucket, but then fail.

rest/utilities_testing_resttester.go
```diff
@@ -491,3 +496,45 @@ func (rt *RestTester) PutDocWithAttachment(docID string, body string, attachment
...
+// RawDocXattrsWellKnown contains well known fields - see RawDocXattrsOthers for those not known until runtime.
+type RawDocXattrsWellKnown struct {
+	Sync       db.SyncDataJSON `json:"_sync"`
+	GlobalSync map[string]any  `json:"_globalSync"`
+	VV         map[string]any  `json:"_vv"`
+	MOU        map[string]any  `json:"_mou"`
+}
+
+// RawDocXattrsOthers contains fields that can only be known at runtime (e.g. user xattr names)
+type RawDocXattrsOthers map[string]interface{}
```
torcolvin: very nit:

```suggestion
type RawDocXattrsOthers map[string]any
```

rest/utilities_testing_resttester.go
```diff
@@ -491,3 +496,45 @@ func (rt *RestTester) PutDocWithAttachment(docID string, body string, attachment
...
+}
+
+// RawDocXattrs is a non-exhaustive set of xattrs returned by the _raw endpoint. Used for test assertions.
+type RawDocXattrs struct {
+	RawDocXattrsWellKnown
+	RawDocXattrsOthers
+}
+
+// RawDocXattrsWellKnown contains well known fields - see RawDocXattrsOthers for those not known until runtime.
+type RawDocXattrsWellKnown struct {
```
torcolvin, line 512: I wonder if it makes sense to use the same structure for writing, I think `map[string]any` could include json.RawMessage. Not sure, if you don't want to rewrite this.

rest/utilities_testing_resttester.go
```diff
@@ -491,3 +496,45 @@ func (rt *RestTester) PutDocWithAttachment(docID string, body string, attachment
...
+// RawDocXattrsOthers contains fields that can only be known at runtime (e.g. user xattr names)
+type RawDocXattrsOthers map[string]interface{}
+
+func (t *RawDocXattrs) MarshalJSON() ([]byte, error) {
+	// overwrite fields in Main into Extra
+	data, err := json.Marshal(t.RawDocXattrsWellKnown)
+	if err != nil {
+		return nil, err
+	}
+	if err := json.Unmarshal(data, &t.RawDocXattrsOthers); err != nil {
```
torcolvin, line 528: This seems very weird to me that `RawDocXattrs` has side effects. It doesn't feel right for caching, especially in a test function, and you don't know someone has mutated `RawDocXattrs`. Are you trying to avoid a copy of `t.RawDocXattrsOthers`?

rest/utilities_testing_resttester.go
```diff
@@ -491,3 +496,45 @@ func (rt *RestTester) PutDocWithAttachment(docID string, body string, attachment
...
+		return nil, err
+	}
+	return json.Marshal(t.RawDocXattrsOthers)
+}
+
+func (t *RawDocXattrs) UnmarshalJSON(p []byte) error {
+	if err := json.Unmarshal(p, &t.RawDocXattrsWellKnown); err != nil {
+		return err
+	}
+	return json.Unmarshal(p, &t.RawDocXattrsOthers)
```
torcolvin: In this case, I think you probably want to delete the keys of `_vv`, `_sync` etc, so these aren't listed twice.

rest/admin_api.go
bbrks: 👍 included full `$document` virtual xattr - agree with importance for 4.0 debugging

rest/admin_api.go
bbrks: Thanks!! This is a bug :)

rest/admin_api.go
bbrks: Yeah, the persisted version of SyncData looks a bit different than the Unmarshalled version which is why I had to redefine some things like this.

There's not many places (if anywhere) we interact with sync data directly as it is represented marshalled.

rest/utilities_testing_resttester.go
bbrks: gah old habits

rest/utilities_testing_resttester.go
bbrks: I'll add a TODO but probably just punt on this. It doesn't cause any issues for its current usage in tests and is only a problem if you go looking for the known keys inside the "Others" map, which is too awkward to accidentally do, vs. accessing the struct fields.

This will go away with the v2 implementation of JSON with the 'inline' feature - but this code is an awkward hack to get a set of known fields out, and then a map of everything else.

https://github.com/golang/go/issues/6213

rest/utilities_testing_resttester.go
bbrks: Dumb implementation of JSON v2's `inline` feature

rest/utilities_testing_resttester.go
bbrks: I don't want a strongly typed struct for writing since I want it to tolerate future xattr changes well, but for ease of access, the struct is used in tests to navigate the known objects.

rest/doc_api_test.go
bbrks: `assert.Contains` works well instead. This is just old code.

docs/api/paths/admin/keyspace-_raw-docid.yaml
```diff
@@ -43,7 +43,7 @@ get:
             additionalProperties: true
             properties:
               _xattrs:
-                description: The extended attributes of the document, which can include Sync Gateway metadata, such as revision history, channel information, and other sync-related data. This is non-exhaustive and only includes xattrs that are relevant to Sync Gateway.
+                description: The extended attributes of the document, which can include Sync Gateway metadata, such as revision history, channel information, and other sync-related data. This is non-exhaustive and only includes xattrs that are relevant to Sync Gateway. Any `null` values indicate the xattr is not present on the document.
```
torcolvin, line 46: The way in openapi to do this is in OpenAPI 3.0 (currently set):

`nullable: true` https://swagger.io/docs/specification/v3_0/data-models/data-types/#null

At a soon but later point, we can update to 3.1 (not a big deal, mostly just changes null types), and you can do `type: [object, null]`

db/document.go
```diff
@@ -131,6 +133,21 @@ func (sd *SyncData) getCurrentChannels() base.Set {
 	return ch
 }
 
+// RedactRawSyncData runs HashRedact on the given sync data.
+func RedactRawSyncData(syncData []byte, salt string) ([]byte, error) {
```
gregns1: We should have a way to redact attachment names in `_globalSync` like we do for sync data redaction

torcolvin (approved):

PR #7651, opened 2025-07-28
Return when sending a 422 to a rev

torcolvin:
Fixes some wacky pathways where the HLV gets updated twice and this rev gets added to the BlipTesterClient data store. This isn't the correct behavior. 

I think this actually doesn't usually result in a problem since the same contents are written twice, but the second time it should have a different (full) body.

I didn't find that this was a problem until I rewrote some the blip client tester code to resolve conflicts, but I wanted to pull this out into a change sooner rather than later due to unexpected behavior.

## Pre-review checklist


bbrks (approved):
good find

PR #7657, opened 2025-08-01
CBG-4267 implement LWW conflict resolution for blip tester

torcolvin:
CBG-4267 implement conflict resolution for blip tester

- only implements LWW
    - put the incoming rev + existing CBL rev into mv
    - create new CBL rev
    - mark last replicated version as the incoming rev, so the new rev is replicated to SG
- Rename `HybridLogicalVector.Equals` to `HybridLogicalVector.Equal` to match go idioms
- Update `proposeChanges` message to include full HLV to be able to push the version back to SG
- It is OK for `sendRev` to return a 409, do not cause blip tester to fail. CBL getting a 409 is expected until CBL resolves the version locally.

This was tested with the topology tests, but it does not fully work. There's a chance the implementation needs to be improved, but this is a good start. The only test that hits this code pathway for conflicts is the new test I wrote, the topology tests are currently skipped.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3238/


rest/utilities_testing_blip_client.go
```diff
@@ -276,6 +299,50 @@ func (cd *clientDoc) _proposeChangesEntryForDoc() *proposeChangeBatchEntry {
 	return &proposeChangeBatchEntry{docID: cd.id, version: latestRev.version, revTreeIDHistory: revTreeIDHistory, hlvHistory: latestRev.HLV, latestServerVersion: cd._latestServerVersion, seq: cd._latestSeq, isDelete: latestRev.isDelete}
 }
 
+// _resolveConflict will resolve a conflict for the given incoming version. This will mutate the hlv, and return the expected body and version to write.
```
adamcfraser: Usually I'd consider "resolving the conflict" as a separate step from "detecting a conflict", like the way we do for ISGR (see db.IsIllegalConflict and db.resolveConflict).  I think this is doing both, and would be easier to follow if the two tasks were separated out.  Let me know if there's something about HLVs that makes it more logical to do both at once.

rest/utilities_testing_blip_client.go
```diff
@@ -276,6 +299,50 @@ func (cd *clientDoc) _proposeChangesEntryForDoc() *proposeChangeBatchEntry {
...
+		}
+		opts.hlv.SetMergeVersion(opts.incomingVersion.SourceID, opts.incomingVersion.Value)
+		opts.hlv.SetMergeVersion(clientCV.SourceID, clientCV.Value)
+		newVersion := db.Version{SourceID: opts.localSourceID, Value: uint64(opts.hlc.Now())}
+		return latestLocalRev.body, newVersion
+	default:
+		opts.t.Fatalf("Unknown conflict resolver %q - cannot resolve detected conflict", opts.conflictResolver)
+	}
+	opts.t.Fatalf("Unreachable code in _resolveConflict - should never reach here")
+	return latestLocalRev.body, latestLocalRev.version.CV // unreachable, but required to satisfy return type
```
adamcfraser: Can this be moved into the default for the above switch avoid the "unreachable code" Fatalf?

rest/utilities_testing_blip_client.go
```diff
@@ -276,6 +299,38 @@ func (cd *clientDoc) _proposeChangesEntryForDoc() *proposeChangeBatchEntry {
...
+
+const (
+	remoteDocWinner conflictResolutionWinner = iota // The remote document is the winner in the conflict resolution
+	localDocWinner                                  // The local document is the winner in the conflict resolution
+)
+
+func (r *clientDocRev) _pickConflictWinner(t testing.TB, incomingCV db.Version, conflictResolver BlipTesterClientConflictResolverType) conflictResolutionWinner {
+	localCV := r.HLV.ExtractCurrentVersionFromHLV()
+	// safety check - ensure SG is not sending a rev that we already had - ensures changes feed messaging is working correctly to prevent
+	if localCV.Equal(incomingCV) {
```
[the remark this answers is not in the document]
torcolvin: this is not how go works

rest/utilities_testing_blip_client.go
```diff
@@ -276,6 +299,38 @@ func (cd *clientDoc) _proposeChangesEntryForDoc() *proposeChangeBatchEntry {
...
+		require.FailNow(t, fmt.Sprintf("incoming CV %#+v is equal to local CV revision %#+v - this should've been filtered via changes response before ending up as a rev", incomingCV, r))
```
adamcfraser: This is probably valid for the current topologies being tested using the BlipTesterClient, but in real-world scenarios it's possible that a client received the revision via another replication (P2P, replication with a different SGW cluster) in-between the changes response and the rev being sent.  Will need to tweak this if we add those topologies in future - maybe add a comment to that effect beside the FailNow so that it's easy to recognize when someone hits that?

rest/utilities_testing_blip_client.go
```diff
@@ -276,6 +299,38 @@ func (cd *clientDoc) _proposeChangesEntryForDoc() *proposeChangeBatchEntry {
...
+	}
+	if incomingCV.SourceID == localCV.SourceID {
```
adamcfraser: I think this should also check incomingCV.Value == localCV.Value.  I think you're basing this on that if the SourceIDs are the same that this isn't actually a conflict and we shouldn't be invoking this code in the first place, but as in the comment above I think there are race conditions between the changes response and the rev message where either the incomingCV or the localCV should be the winner, and we need to wait for the conflictResolver to make that decision.

adamcfraser: I think this should also check incomingCV.Value == localCV.Value.  Possibly you've omitted that based on the idea that if the SourceIDs are the same that this isn't actually a conflict and we shouldn't be invoking this code in the first place?  But otherwise if the two revisions are by the same source, I think we'd always want to pick the one with the larger Value, which isn't necessarily the remote (for the same race conditions described above)

rest/utilities_testing_blip_client.go
```diff
@@ -565,28 +621,31 @@ func (btr *BlipTesterReplicator) handleRev(ctx context.Context, btc *BlipTesterC
...
-			var hlv db.HybridLogicalVector
+			var incomingVersion DocVersion
+			var incomingHLV *db.HybridLogicalVector
 			if btc.UseHLV() {
+				cv, err := db.ParseVersion(revID)
+				require.NoError(btr.TB(), err, "error parsing version %q: %v", revID, err)
+				incomingVersion = DocVersion{CV: cv}
 				if revHistory != "" {
-					existingVersion, _, err := db.ExtractHLVFromBlipMessage(revHistory)
+					// why is rev history not populated for a deleted message?
```
adamcfraser: I agree with this comment - I think the HLV should be getting sent with a tombstone as well, when present.  Is that not the case?  Or is this just asking whether the revHistory !="" check is not needed?  If it's the latter, there wouldn't be a revHistory if the actor doing the create and the delete were the same source.

rest/utilities_testing_blip_client.go
```diff
@@ -565,28 +621,31 @@ func (btr *BlipTesterReplicator) handleRev(ctx context.Context, btc *BlipTesterC
...
+					incomingHLV, _, err = db.ExtractHLVFromBlipMessage(revHistory)
```
adamcfraser: I didn't think the revHistory in the rev message includes the CV (since it's sent in the rev property). It's possible you're getting around that using the AddVersion on line 637 but I don't think that's what you want to do (as that function is going to do things like move the MV to the PV).   If it's correct that the revHistory is just MV and PV, it might be a better approach to either:
- prepend the CV string to that before calling ExtractHLVFromBlipMessage
- Explicitly set SourceID and Version on incomingHLV using cv on line 637, instead of using AddVersion

rest/utilities_testing_blip_client.go
```diff
@@ -747,28 +806,31 @@ func (btr *BlipTesterReplicator) handleRev(ctx context.Context, btc *BlipTesterC
...
-		var hlv db.HybridLogicalVector
+		var incomingVersion DocVersion
+		var incomingHLV *db.HybridLogicalVector
 		if btc.UseHLV() {
+			cv, err := db.ParseVersion(revID)
+			require.NoError(btr.TB(), err, "error parsing version %q: %v", revID, err)
+			incomingVersion = DocVersion{CV: cv}
 			if revHistory != "" {
-				existingVersion, _, err := db.ExtractHLVFromBlipMessage(revHistory)
+				incomingHLV, _, err = db.ExtractHLVFromBlipMessage(revHistory)
```
adamcfraser: Same comments as above for the revHistory handling.

rest/utilities_testing_blip_client.go
```diff
@@ -1260,6 +1332,11 @@ func (btcc *BlipTesterCollectionClient) sendRev(ctx context.Context, change prop
 		btcc.sendRev(ctx, change, false)
 		return
 	}
+	if errorCode == strconv.Itoa(http.StatusConflict) {
+		// If there is a conflict created between the proceeding proposeChanges and this rev message.
```
adamcfraser, line 1356: Do you think some info/debug logging for this case is warranted so that this scenario is more visible in test output?  I feel like in many cases it would be unexpected?

rest/utilities_testing_blip_client.go
```diff
@@ -1923,32 +2000,77 @@ func (btcc *BlipTesterCollectionClient) pruneVersion(docID string, version DocVe
...
-		HLV:       opts.hlv,
+		HLV:       *updatedHLV,
 	}
+	switch winner {
+	case remoteDocWinner:
+		docRev.body = opts.body
+		docRev.version = opts.incomingVersion
+		if btcc.UseHLV() {
+			// Add the incoming HLV to the local HLV, regardless of winner
+			require.NoError(btcc.TB(), docRev.HLV.AddNewerVersions(opts.incomingHLV))
```
adamcfraser: In a non-conflict case, I think we just be using the incomingHLV as the HLV, as we've determined that localRev.HLV is already included in its causal history.

There's a corner case where there has been asymmetric pruning of the HLV on client and server, but I don't think that's tested by the BlipTestClient.  And even if it was, and we wanted to preserve those, I think we'd want to restore those by doing something like incomingHLV.AddNewerVersions(docRev.HLV).

More generally, I think this is going to be harder to maintain/debug if the non-conflict code path and the "conflict, remoteWins" cases are the same - I think this goes back to not using 'pickConflictWinner' unless there's actually a conflict.

rest/utilities_testing_blip_client.go
```diff
@@ -1923,32 +2000,77 @@ func (btcc *BlipTesterCollectionClient) pruneVersion(docID string, version DocVe
...
 	newClientSeq := btcc._nextSequence()
 
+	winner := remoteDocWinner
+	var localRev *clientDocRev
+	doc, hasLocalDoc := btcc._getClientDoc(docID)
+	// No conflict resolution for revtree at this time
+	var updatedHLV *db.HybridLogicalVector
+	if hasLocalDoc {
+		localRev = doc._latestRev(btcc.TB())
+		winner = localRev._pickConflictWinner(btcc.TB(), opts.incomingVersion.CV, btcc.parent.ConflictResolver)
```
adamcfraser: This still isn't really separated out in the way I was thinking.  I was suggesting to separate "detecting a conflict" from "resolving a conflict", where:

- "detecting a conflict": Do the incoming and local HLVs represent a conflict?
- "resolving a conflict": Running the conflict resolver to compute the winner and the new body (if merge), and computing the combined HLV

Here "resolving a conflict" is only run if it's actually a conflict, otherwise it's the normal update case.

rest/utilities_testing_blip_client.go
adamcfraser: Also, we still need to write MV in the case of an actual conflict where we pick the remote as winner, right?  That doesn't look like it's being handled here.

rest/utilities_testing_blip_client.go
```diff
@@ -1923,32 +2000,77 @@ func (btcc *BlipTesterCollectionClient) pruneVersion(docID string, version DocVe
...
-	if !ok {
+		if btcc.UseHLV() {
+			// move all versions from remote HLV to local HLV
+			require.NoError(btcc.TB(), docRev.HLV.AddNewerVersions(opts.incomingHLV))
+			// manually reconstruct the HLV
+			// - remove the any pv that contain the merge sourceIDs
+			// - add the merge sourceIDs with the incoming version and local version
+			// - update the new CV
+			delete(docRev.HLV.PreviousVersions, opts.incomingHLV.SourceID)
+			delete(docRev.HLV.PreviousVersions, btcc.parent.SourceID)
```
adamcfraser: Shouldn't this use localRev.HLV.SourceID and not btcc.parent.SourceID?

rest/utilities_testing_blip_client.go
```diff
@@ -1963,7 +2085,12 @@ func (btcc *BlipTesterCollectionClient) addRev(docID string, opts revOptions) {
 		doc._seqsByVersions[*opts.replacedVersion] = newClientSeq
 	}
 	if opts.updateLatestServerVersion {
-		doc._latestServerVersion = docRev.version
+		doc._latestServerVersion = opts.incomingVersion
+	}
+	// if we resolved a conflict, then we might need to push this conflict back
+	if winner == localDocWinner {
```
adamcfraser: Similar to the above, localDocWinner isn't the only possibility when resolving a conflict.

adamcfraser (changes requested):

rest/utilities_testing_blip_client.go
torcolvin: What is the right behavior for:

- incomingCV.sourceID == localCV.sourceID && incomingCV.Value == localCV.Value
    - Is it possible that these would be equal but the pv or mv are not equal?
- incomingCV.sourceID == localCV.sourceID && incomingCV.Value < localCV.Value
        - Is it possible that these would be equal but the pv or mv are not equal?
        
I propose to fail right now because this is a condition isn't possible in the current implementation and to implement this in a future ticket.

rest/utilities_testing_blip_client.go
```diff
@@ -500,6 +555,7 @@ func (btr *BlipTesterReplicator) handleChanges(btc *BlipTesterClient) func(*blip
 						knownRevs[i] = []interface{}{} // sending empty array means we've not seen the doc before, but still want it
 						continue
 					} else if localHLV.DominatesSource(changesVersion) {
+						base.DebugfCtx(ctx, base.KeySGTest, "Skipping changes for incoming doc %q with rev %d@%s as we already have a newer version %#+v", docID, changesVersion.Value, changesVersion.SourceID, localHLV)
```
torcolvin, line 616: I think I want to change rev %d@%s

rest/utilities_testing_blip_client.go
torcolvin: Put the CV ahead of the other strings, and pulled this into a function.

rest/utilities_testing_blip_client.go
torcolvin: This makes sense to me, it is the case that it is the same actor doing create and delete.

rest/utilities_testing_blip_client.go
torcolvin: I think what we care about removing the sourceID that comes from merge versions and the sourceID from cv (btcc.parent.SourceID)

rest/utilities_testing_blip_client.go
adamcfraser: For the first case, it's the same revision and it's not a conflict (regardless of whether PV or MV are equal due to asymmetric pruning, etc). 
For the second case this isn't a conflict (local dominates incoming).

rest/utilities_testing_blip_client.go
```diff
@@ -276,6 +300,95 @@ func (cd *clientDoc) _proposeChangesEntryForDoc() *proposeChangeBatchEntry {
...
+
+	// local revision is newer than incoming revision. Common case:
+	// - CBL writes document 1@cbl1
+	// - CBL pushes to SG as 1@cbl1
+	// - CBL pulls document 1@cbl1
+	//
+	// NOTE: without P2P replication, this should not be the case and we would not get this revision, since CBL
+	// would respond to a SG changes message that CBL does not need this revision
+	if localHLV.DominatesSource(*incomingCV) {
+		require.FailNow(t, fmt.Sprintf("incoming CV %#+v has lower version than the local revision %#+v - this should've been filtered via changes response before ending up as a rev. blip tester would reply that to Sync Gateway that it doesn't need this revision", incomingCV, localHLV))
```
adamcfraser, line 347: I'm not sure this should fail the test at this step (I think so that we can validate that changes processing is working as expected?).  Wouldn't it be preferable to have this match CBL behaviour?  I'm concerned that this could incorrectly handle other scenarios where the changes message isn't sufficient to identify whether local dominates incoming (maybe CBL switching clusters?).

rest/utilities_testing_blip_client.go
```diff
@@ -276,6 +300,95 @@ func (cd *clientDoc) _proposeChangesEntryForDoc() *proposeChangeBatchEntry {
...
+	btcc.TB().Fatalf("Unknown conflict resolver %q - cannot resolve detected conflict", btcc.parent.ConflictResolver)
+	return nil, db.HybridLogicalVector{}
+}
+
+func (btcc *BlipTesterCollectionClient) _resolveConflictLWW(incomingHLV *db.HybridLogicalVector, incomingBody []byte, latestLocalRev *clientDocRev) (body []byte, hlv db.HybridLogicalVector) {
+	latestLocalHLV := latestLocalRev.HLV
+	updatedHLV := latestLocalRev.HLV.Copy()
+	// resolve conflict in favor of remote document
+	if incomingHLV.Version > latestLocalHLV.Version {
+		require.NoError(btcc.TB(), updatedHLV.AddNewerVersions(incomingHLV))
```
adamcfraser, line 373: Is CBL not updating the merge version in this scenario?  That might be reasonable, and is similar to what CBL used to do for revTrees.  If so we will want to update the design doc accordingly.

adamcfraser (commented):
Changes look good to me.  No functional concerns, just a question about whether FailNow is appropriate in one spot, and a clarifying question about CBL behaviour.

rest/utilities_testing_blip_client.go
torcolvin: I agree that we should relax this behavior, but I'd prefer to do it in a separate PR when I can find out that is necessary and how to handle it. Right now it is an assertion of the correct behavior while development.

Until I enable tests that would cause this I do not want to relax the assertion since it is more likely to catch problems.

adamcfraser (dismissed):

adamcfraser (approved):

PR #7659, opened 2025-08-01
CBG-4753: Add `version_type` preference for `/_changes` to allow CV-aware consumers

bbrks:
CBG-4753

- Adds a `version_type` option to REST API Changes feeds which has two possible options - `rev` (default) or `cv`
  - This controls whether the change entries preferentially show a revtree ID or current version value to notify clients of a new document change.
  - If old documents do not have a cv (they were written prior to 4.0, we'll still continue to send the `rev` version of the change).
  - This parameter is documented as a preference only, and notes the case where `rev` may still be produced on a feed requesting `cv`

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/37/

### DIFF
--- a/db/changes.go
+++ b/db/changes.go
@@ -181,7 +181,6 @@ func (db *DatabaseCollectionWithUser) AddDocToChangeEntryUsingRevCache(ctx conte
 	if err != nil {
 		return err
 	}
-	// TODO: Stamp CV?
 	entry.Doc, err = rev.As1xBytes(ctx, db, nil, nil, false)
 	return err
 }
--- a/db/changes.go
+++ b/db/changes.go
@@ -1366,9 +1366,9 @@ func createChangesEntry(ctx context.Context, docid string, db *DatabaseCollectio
 
 	switch options.VersionType {
 	case ChangesVersionTypeCV:
-		row.Changes = []ChangeByVersionType{{options.VersionType: populatedDoc.HLV.GetCurrentVersionString()}}
+		row.Changes = []ChangeByVersionType{{ChangesVersionTypeCV: populatedDoc.HLV.GetCurrentVersionString()}}
 	case "", ChangesVersionTypeRevTreeID:
-		row.Changes = []ChangeByVersionType{{options.VersionType: populatedDoc.CurrentRev}}
+		row.Changes = []ChangeByVersionType{{ChangesVersionTypeRevTreeID: populatedDoc.CurrentRev}}
 	default:
 		base.AssertfCtx(ctx, "createChangeEntry called with an unsupported VersionType: %s", options.VersionType)
 	}
--- a/db/changes.go
+++ b/db/changes.go
@@ -1364,14 +1364,13 @@ func createChangesEntry(ctx context.Context, docid string, db *DatabaseCollectio
 		return nil
 	}
 
-	row.Changes = []ChangeByVersionType{{ChangesVersionTypeRevTreeID: populatedDoc.CurrentRev}}
 	switch options.VersionType {
 	case ChangesVersionTypeCV:
-		row.Changes[0] = ChangeByVersionType{options.VersionType: populatedDoc.HLV.GetCurrentVersionString()}
-	case ChangesVersionTypeRevTreeID:
-		fallthrough
+		row.Changes = []ChangeByVersionType{{options.VersionType: populatedDoc.HLV.GetCurrentVersionString()}}
+	case "", ChangesVersionTypeRevTreeID:
+		row.Changes = []ChangeByVersionType{{options.VersionType: populatedDoc.CurrentRev}}
 	default:
-		// already initialized with a 'rev' change entry above
+		base.AssertfCtx(ctx, "createChangeEntry called with an unsupported VersionType: %s", options.VersionType)
 	}
 
 	row.Deleted = populatedDoc.Deleted
--- a/db/changes.go
+++ b/db/changes.go
@@ -26,32 +26,51 @@ import (
 // Options for changes-feeds.  ChangesOptions must not contain any mutable pointer references, as
 // changes processing currently assumes a deep copy when doing chanOpts := changesOptions.
 type ChangesOptions struct {
-	Since          SequenceID      // sequence # to start _after_
-	Limit          int             // Max number of changes to return, if nonzero
-	Conflicts      bool            // Show all conflicting revision IDs, not just winning one?
-	IncludeDocs    bool            // Include doc body of each change?
-	Wait           bool            // Wait for results, instead of immediately returning empty result?
-	Continuous     bool            // Run continuously until terminated?
-	RequestPlusSeq uint64          // Do not stop changes before cached sequence catches up with requestPlusSeq
-	HeartbeatMs    uint64          // How often to send a heartbeat to the client
-	TimeoutMs      uint64          // After this amount of time, close the longpoll connection
-	ActiveOnly     bool            // If true, only return information on non-deleted, non-removed revisions
-	Revocations    bool            // Specifies whether revocation messages should be sent on the changes feed
-	clientType     clientType      // Can be used to determine if the replication is being started from a CBL 2.x or SGR2 client
-	ChangesCtx     context.Context // Used for cancelling checking the changes feed should stop
+	Since          SequenceID         // sequence # to start _after_
+	Limit          int                // Max number of changes to return, if nonzero
+	Conflicts      bool               // Show all conflicting revision IDs, not just winning one?
+	IncludeDocs    bool               // Include doc body of each change?
+	Wait           bool               // Wait for results, instead of immediately returning empty result?
+	Continuous     bool               // Run continuously until terminated?
+	RequestPlusSeq uint64             // Do not stop changes before cached sequence catches up with requestPlusSeq
+	HeartbeatMs    uint64             // How often to send a heartbeat to the client
+	TimeoutMs      uint64             // After this amount of time, close the longpoll connection
+	ActiveOnly     bool               // If true, only return information on non-deleted, non-removed revisions
+	Revocations    bool               // Specifies whether revocation messages should be sent on the changes feed
+	clientType     clientType         // Can be used to determine if the replication is being started from a CBL 2.x or SGR2 client
+	ChangesCtx     context.Context    // Used for cancelling checking the changes feed should stop
+	VersionType    ChangesVersionType // The type of version to use for the changes feed. This is used to determine whether to use send revtree IDs or CV in the changes feed entries.
+}
+
+type ChangesVersionType string
+
+const (
+	ChangesVersionTypeRevTreeID ChangesVersionType = "rev" // Use revtree IDs in changes feed entries
+	ChangesVersionTypeCV        ChangesVersionType = "cv"  // Use current version in changes feed entries
+)
+
+func ParseChangesVersionType(s string) (ChangesVersionType, error) {
+	switch ChangesVersionType(s) {
+	case "", ChangesVersionTypeRevTreeID:
+		return ChangesVersionTypeRevTreeID, nil
+	case ChangesVersionTypeCV:
+		return ChangesVersionTypeCV, nil
+	default:
+		return "", fmt.Errorf("unknown changes version type: %q", s)
+	}
 }
 
 // A changes entry; Database.GetChanges returns an array of these.
 // Marshals into the standard CouchDB _changes format.
 type ChangeEntry struct {
-	Seq            SequenceID      `json:"seq"`
-	ID             string          `json:"id"`
-	Deleted        bool            `json:"deleted,omitempty"`
-	Removed        base.Set        `json:"removed,omitempty"`
-	Doc            json.RawMessage `json:"doc,omitempty"`
-	Changes        []ChangeRev     `json:"changes"`
-	Err            error           `json:"err,omitempty"` // Used to notify feed consumer of errors
-	allRemoved     bool            // Flag to track whether an entry is a removal in all channels visible to the user.
+	Seq            SequenceID            `json:"seq"`
+	ID             string                `json:"id"`
+	Deleted        bool                  `json:"deleted,omitempty"`
+	Removed        base.Set              `json:"removed,omitempty"`
+	Doc            json.RawMessage       `json:"doc,omitempty"`
+	Changes        []ChangeByVersionType `json:"changes"`
+	Err            error                 `json:"err,omitempty"` // Used to notify feed consumer of errors
+	allRemoved     bool                  // Flag to track whether an entry is a removal in all channels visible to the user.
 	branched       bool
 	backfill       backfillFlag // Flag used to identify non-client entries used for backfill synchronization (di only)
 	principalDoc   bool         // Used to indicate _user/_role docs
@@ -74,14 +93,10 @@ const (
 	BackfillFlag_Complete
 )
 
-type ChangeRev map[string]string // Key is always "rev", value is rev ID
+type ChangeByVersionType map[ChangesVersionType]string // Keyed by the type of version in the value
 
 type ViewDoc struct {
 	Json json.RawMessage // should be type 'document', but that fails to unmarshal correctly
-}
-
-func (db *DatabaseCollectionWithUser) AddDocToChangeEntry(ctx context.Context, entry *ChangeEntry, options ChangesOptions) {
-	db.addDocToChangeEntry(ctx, entry, options)
 }
 
 // Adds a document body and/or its conflicts to a ChangeEntry
@@ -159,7 +174,7 @@ func (db *DatabaseCollectionWithUser) AddDocInstanceToChangeEntry(ctx context.Co
 					entry.Deleted = false
 				}
 				if !(options.ActiveOnly && leaf.Deleted) {
-					entry.Changes = append(entry.Changes, ChangeRev{"rev": leaf.ID})
+					entry.Changes = append(entry.Changes, ChangeByVersionType{"rev": leaf.ID})
 				}
 			}
 		})
@@ -300,7 +315,7 @@ func (db *DatabaseCollectionWithUser) buildRevokedFeed(ctx context.Context, ch c
 					continue
 				}
 
-				change := makeRevocationChangeEntry(logEntry, seqID, singleChannelCache.ChannelID())
+				change := makeRevocationChangeEntry(ctx, logEntry, seqID, singleChannelCache.ChannelID(), options.VersionType)
 
 				base.DebugfCtx(ctx, base.KeyChanges, "Channel feed processing revocation seq: %v in channel %s ", seqID, base.UD(singleChannelCache.ChannelID().Name))
 
@@ -459,7 +474,7 @@ func (db *DatabaseCollectionWithUser) changesFeed(ctx context.Context, singleCha
 					TriggeredBy: options.Since.TriggeredBy,
 				}
 
-				change := makeChangeEntry(logEntry, seqID, singleChannelCache.ChannelID())
+				change := makeChangeEntry(ctx, logEntry, seqID, singleChannelCache.ChannelID(), options.VersionType)
 				lastSeq = logEntry.Sequence
 
 				// Don't include deletes or removals during initial channel backfill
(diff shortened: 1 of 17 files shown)


db/changes.go
```diff
@@ -26,32 +26,62 @@ import (
...
+		return ChangesVersionTypeRevTreeID, nil
+	case ChangesVersionTypeCV:
+		return ChangesVersionTypeCV, nil
+	default:
+		return "", fmt.Errorf("unknown changes version type: %q", s)
+	}
+}
+
+// ChangeVersionString attempts to return the version string for the preferred ChangesVersionType, but will fall back to rev if cv is not available when requested.
+func (ce *ChangeEntry) ChangeVersionString(versionType ChangesVersionType) string {
```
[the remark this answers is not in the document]
bbrks: `ChangeEntry` is guaranteed to have at least 1 change (more in the case of conflicts and ?style=all_docs requested)

db/changes.go
```diff
@@ -26,32 +26,62 @@ import (
...
+	if s, ok := ce.Changes[0][versionType]; ok {
+		return s
+	}
+	// requested version type not found, return `rev` as a fallback.
+	return ce.Changes[0][ChangesVersionTypeRevTreeID]
```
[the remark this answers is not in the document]
bbrks: This is safe - always 1 element and it will return zero value

db/changes.go
```diff
@@ -494,20 +532,33 @@ func (db *DatabaseCollectionWithUser) changesFeed(ctx context.Context, singleCha
...
+		Changes:      []ChangeByVersionType{{ChangesVersionTypeRevTreeID: logEntry.RevID}},
 		branched:     (logEntry.Flags & channels.Branched) != 0,
 		principalDoc: logEntry.IsPrincipal,
 		collectionID: logEntry.CollectionID,
 	}
+
+	switch versionType {
+	case ChangesVersionTypeCV:
+		if logEntry.SourceID != "" {
+			change.Changes[0] = ChangeByVersionType{versionType: Version{SourceID: logEntry.SourceID, Value: logEntry.Version}.String()}
```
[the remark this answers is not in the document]
bbrks: intentional

db/changes.go
```diff
@@ -1308,9 +1359,16 @@ func createChangesEntry(ctx context.Context, docid string, db *DatabaseCollectio
 		return nil
 	}
 
-	changes := make([]ChangeRev, 1)
-	changes[0] = ChangeRev{"rev": populatedDoc.CurrentRev}
-	row.Changes = changes
+	row.Changes = []ChangeByVersionType{{ChangesVersionTypeRevTreeID: populatedDoc.CurrentRev}}
+	switch options.VersionType {
+	case ChangesVersionTypeCV:
+		row.Changes[0] = ChangeByVersionType{options.VersionType: populatedDoc.HLV.GetCurrentVersionString()}
```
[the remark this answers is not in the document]
bbrks: intentional

db/changes.go
```diff
@@ -26,32 +26,62 @@ import (
...
+	case ChangesVersionTypeCV:
+		return ChangesVersionTypeCV, nil
+	default:
+		return "", fmt.Errorf("unknown changes version type: %q", s)
+	}
+}
+
+// ChangeVersionString attempts to return the version string for the preferred ChangesVersionType, but will fall back to rev if cv is not available when requested.
+func (ce *ChangeEntry) ChangeVersionString(versionType ChangesVersionType) string {
+	if s, ok := ce.Changes[0][versionType]; ok {
```
torcolvin: Since this is pulled into a function, do you think this makes sense to check if `len(ce.Changes) > 1` before going into this function to avoid a panic if this is called in a place that is unexpected and there are somehow no changes?

I think in practice a ChangeEntry should always have changes but it is sometimes constructed from an empty struct so there could be a case where it is buggy and I don't want Sync Gateway to panic in prod.

db/changes.go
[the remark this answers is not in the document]
torcolvin: I think this statement is true once things get to the REST api, but there are actually several places where `ChangeEntry` are constructed without changes. These should only be where `ChangeEntry.Err != nil`.

I'd feel better if this code couldn't panic at all and just return an empty string in this case.

db/changes.go
```diff
@@ -124,11 +153,14 @@ func (db *DatabaseCollectionWithUser) addDocToChangeEntry(ctx context.Context, e
...
-		db.AddDocInstanceToChangeEntry(ctx, entry, doc, options)
+		if err := db.AddDocInstanceToChangeEntry(ctx, entry, doc, options); err != nil {
+			base.WarnfCtx(ctx, "Changes feed: error adding doc %q to change entry: %v", base.UD(entry.ID), err)
+			return
+		}
 
 	} else if options.IncludeDocs {
 		// Retrieve document via rev cache
-		revID := entry.Changes[0]["rev"]
+		revID := entry.ChangeVersionString(options.VersionType)
```
torcolvin: I'd be tempted to name this version rather than revID. I found this code particularly confusing because I worried that AddDocToChangeEntryUsingRevCache was not going to be safe for CV or revtree ids. I am not sure whether or not this will be confusing outside this PR.

I do sometimes use revID to mean CV or revtree id, but it's not always clear when this is the case to me.revision id makes sense as a CV or revtree id and we use that name for the output format for blip & rest because the blip property is `_rev` inline or `rev`. My thought is that `version` is used when it could be either in the code, but the transport layers could use `rev`.

If you don't find this confusing this is fine, when I read the whole review I am not confused anymore.

db/changes.go
```diff
@@ -1308,9 +1359,16 @@ func createChangesEntry(ctx context.Context, docid string, db *DatabaseCollectio
...
-	row.Changes = changes
+	row.Changes = []ChangeByVersionType{{ChangesVersionTypeRevTreeID: populatedDoc.CurrentRev}}
+	switch options.VersionType {
+	case ChangesVersionTypeCV:
+		row.Changes[0] = ChangeByVersionType{options.VersionType: populatedDoc.HLV.GetCurrentVersionString()}
+	case ChangesVersionTypeRevTreeID:
+		fallthrough
+	default:
+		// already initialized with a 'rev' change entry above
+	}
```
torcolvin: slight preference for this code

```suggestion

	switch options.VersionType {
	case ChangesVersionTypeCV:
		row.Changes = []ChangeByVersionType{{options.VersionType: populatedDoc.HLV.GetCurrentVersionString()}}
	case ChangesVersionTypeRevTreeID:
	    row.Changes = []ChangeByVersionType{{ChangesVersionTypeRevTreeID: populatedDoc.CurrentRev}}
	default:
	     base.AssertfCtx(ctx, "createChangeEntry called with an unsupported VersionType: %s", options.VersionType)
	}
```

db/changes.go
```diff
@@ -1352,15 +1410,18 @@ func createChangesEntry(ctx context.Context, docid string, db *DatabaseCollectio
...
+		}
 	}
 
 	return row
 }
 
 func (options ChangesOptions) String() string {
 	return fmt.Sprintf(
-		`{Since: %s, Limit: %d, Conflicts: %t, IncludeDocs: %t, Wait: %t, Continuous: %t, HeartbeatMs: %d, TimeoutMs: %d, ActiveOnly: %t, Revocations: %t, RequestPlusSeq: %d}`,
+		`{Since: %s, Limit: %d, Conflicts: %t, IncludeDocs: %t, Wait: %t, Continuous: %t, HeartbeatMs: %d, TimeoutMs: %d, ActiveOnly: %t, Revocations: %t, RequestPlusSeq: %d, VersionType: %s}`,
```
torcolvin: If I read this right, VersionType will be empty string for blip or active replicator. Do you think this is confusing if this code was used later? or we should log as %q?

For blip changes feeds probably using cv mode makes sense, but it also doesn't really matter. I'm trying to figure out a way to make debugging this the least amount of confusing.

Should we enforce when calling `MultiChangesFeed` that `VersionType` is not an empty string, and then I think this would make some of the other code easier to understand:

```
	row.Changes = []ChangeByVersionType{{ChangesVersionTypeRevTreeID: populatedDoc.CurrentRev}}
	switch options.VersionType {
	case ChangesVersionTypeCV:
		row.Changes[0] = ChangeByVersionType{options.VersionType: populatedDoc.HLV.GetCurrentVersionString()} 2Code has comments. Press enter to view.
	case ChangesVersionTypeRevTreeID:
		fallthrough
	default:
		// already initialized with a 'rev' change entry above
	}
```

could become which is more obvious to me than knowing everywhere that options.VersionType = "" means ChangesVersionTypeRevTreeID:


```
	switch options.VersionType {
	case ChangesVersionTypeCV:
		row.Changes[0] = ChangeByVersionType{options.VersionType: populatedDoc.HLV.GetCurrentVersionString()} 2Code has comments. Press enter to view.
	case ChangesVersionTypeRevTreeID:
        row.Changes = []ChangeByVersionType{{ChangesVersionTypeRevTreeID: populatedDoc.CurrentRev}}
	default:
         < err here>
	}
```

docs/api/paths/admin/keyspace-_changes.yaml
```diff
@@ -102,6 +102,15 @@ get:
...
         default: false
+    - name: version_type
+      in: query
+      description: The preferred type of document versioning to use for the changes feed.
+      schema:
+        type: string
+        default: rev
+        enum:
+          - rev
+          - cv
```
torcolvin, line 113:
```suggestion
        enum:
          - rev
          - cv
        x-enumDescriptions:
          rev: Sync Gateway <4.0 style revisions. For example: 1-abc, 2-def
          cv: Sync Gateway 4.0+ style revisions. For example: 1@example  
```

db/changes.go
```diff
@@ -147,19 +179,23 @@ func (db *DatabaseCollectionWithUser) AddDocToChangeEntryUsingRevCache(ctx conte
 }
 
 // Adds a document body and/or its conflicts to a ChangeEntry
-func (db *DatabaseCollectionWithUser) AddDocInstanceToChangeEntry(ctx context.Context, entry *ChangeEntry, doc *Document, options ChangesOptions) {
+func (db *DatabaseCollectionWithUser) AddDocInstanceToChangeEntry(ctx context.Context, entry *ChangeEntry, doc *Document, options ChangesOptions) error {
 
 	includeConflicts := options.Conflicts && entry.branched
 
-	revID := entry.Changes[0]["rev"]
+	revID := entry.ChangeVersionString(options.VersionType)
```
torcolvin: I can't make a comment on this line, but in this function if you have a CV and `options.IncludeDocs` is specified then `get1xRevFromDoc` is called with a CV. I am not sure that all of the the code downstream from `get1xRevFromDoc` can take a CV, I am thinking `db.authorizeDoc` specifically.

db/changes.go
```diff
@@ -26,32 +26,62 @@ import (
...
-	Doc            json.RawMessage `json:"doc,omitempty"`
-	Changes        []ChangeRev     `json:"changes"`
-	Err            error           `json:"err,omitempty"` // Used to notify feed consumer of errors
-	allRemoved     bool            // Flag to track whether an entry is a removal in all channels visible to the user.
+	Seq            SequenceID            `json:"seq"`
+	ID             string                `json:"id"`
+	Deleted        bool                  `json:"deleted,omitempty"`
+	Removed        base.Set              `json:"removed,omitempty"`
+	Doc            json.RawMessage       `json:"doc,omitempty"`
+	Changes        []ChangeByVersionType `json:"changes"`
```
torcolvin, line 89: nit:

Not a strong preference here but is `ChangeVersion` a bad name here? to match `DocVersion` etc.

docs/api/paths/admin/keyspace-_changes.yaml
```diff
@@ -166,6 +175,13 @@ post:
               description: 'When true, ensures all valid documents written prior to the request being issued are included in the response.  This is only applicable for non-continuous feeds.'
               type: boolean
               default: false
+            version_type:
+              description: The preferred type of document versioning to use for the changes feed.
+              type: string
+              default: rev
+              enum:
+                - rev
+                - cv
```
torcolvin, line 187:
```suggestion
              enum:
                - rev
                - cv
	        x-enumDescriptions:
	          rev: Sync Gateway <4.0 style revisions. For example: 1-abc, 2-def
	          cv: Sync Gateway 4.0+ style revisions. For example: 1@example      
```

db/changes.go
```diff
@@ -494,20 +532,33 @@ func (db *DatabaseCollectionWithUser) changesFeed(ctx context.Context, singleCha
...
+		fallthrough
+	default:
+		// already initialized with a 'rev' change entry
+	}
+
 	// populate CurrentVersion entry if log entry has sourceID and Version populated
 	// This allows current version to be nil in event of CV not being populated on log entry
 	// allowing omitempty to work as expected
 	if logEntry.SourceID != "" {
+		// TODO: Remove this if we change BLIP generateBlipSyncChanges and tests to use versionType and read the value from the normal Changes array... no reason to store this info in two places.
```
torcolvin: I think this TODO should get a ticket if we want to do this - and it could be marked in `type Change`.

db/changes.go
```diff
@@ -1352,15 +1410,18 @@ func createChangesEntry(ctx context.Context, docid string, db *DatabaseCollectio
 
 	row.Removed = base.SetFromArray(removedChannels)
 	if options.IncludeDocs || options.Conflicts {
-		db.AddDocInstanceToChangeEntry(ctx, row, populatedDoc, options)
+		if err := db.AddDocInstanceToChangeEntry(ctx, row, populatedDoc, options); err != nil {
+			base.WarnfCtx(ctx, "Unable to add doc instance to change entry for %s: %v", base.UD(docid), err)
+			return nil
```
torcolvin: Is returning nil here correct? If it is correct, I think I'd comment why here because it looks wrong.

db/crud.go
```diff
@@ -361,20 +361,36 @@ func (db *DatabaseCollectionWithUser) Get1xRevBodyWithHistory(ctx context.Contex
 //   - attachmentsSince is nil to return no attachment bodies, otherwise a (possibly empty) list of
 //     revisions for which the client already has attachments and doesn't need bodies. Any attachment
 //     that hasn't changed since one of those revisions will be returned as a stub.
-func (db *DatabaseCollectionWithUser) getRev(ctx context.Context, docid, revid string, maxHistory int, historyFrom []string) (revision DocumentRevision, err error) {
-	if revid != "" {
+func (db *DatabaseCollectionWithUser) getRev(ctx context.Context, docid, revOrCV string, maxHistory int, historyFrom []string) (DocumentRevision, error) {
```
torcolvin, line 364: Some general thoughts that aren't actionable necessarily:

I don't love that this could be revOrCV here. I think a totally more correct way would be to pass a `DocVersion` here that is only used in tests. I experimented a bit with this and I couldn't make it work without refactoring. We don't have a `db.Version` for `DocVersion.CV` in the calling functions here. I didn't have time to consider whether it would be more appropriate to have ChangeEntry.Changes actually have a non string version of the CV.

I don't think we have time for this for a release, but it might be worth thinking about in general.

Other places in crud the API is `revTreeID string, cv *db.Version` as in `documentRevisionForRequest` and so changing this would involve downstream changes outside of the scope of this PR.

docs/api/components/schemas.yaml
```diff
@@ -549,9 +549,17 @@ Changes-feed:
...
-                  description: The new revision that was caused by that change.
+                  description: |
+                    The new Revision ID associated with the change.
+                    This is usually omitted in favour of the `cv` property when the `version_type` preference is set to `cv`,
+                    however data written prior to Sync Gateway 4.0 can still emit `rev` only change entries if we do not have a `cv` available for that document revision.
+                  type: string
+                cv:
+                  description: The new Current Version associated with the change. This value requires the `version_type` preference set to `cv`.
                   type: string
             uniqueItems: true
```
torcolvin: Not sure if a better way to express this is:


```
            items:
              oneOf:
                - type: object
                  properties:
                    rev:
                      description: The new Revision ID associated with the change.
                      type: string
                  title: Revision ID
                - type: object
                  properties:
                    cv:
                      description: The new Current Version associated with the change. This value requires the `version_type` preference set to `cv`.
                      type: string
                  title: Current Version
```                  

Or if this would imply that the all of the items are of the same type.

docs/api/components/schemas.yaml
```diff
@@ -549,9 +549,17 @@ Changes-feed:
             type: array
             items:
               type: object
+              minProperties: 1
```
torcolvin: Above description is wrong, at least the part that says `rev`:

```
description: List of document leafs with each leaf containing only a `rev` field
```

db/changes.go
```diff
@@ -26,32 +26,62 @@ import (
...
+// ChangesVersionType determines the preferred version type to use in changes feed entries.
+// If the requested version type is not available, the feed will attempt to fall back to ChangesVersionTypeRevTreeID which should always be available.
+type ChangesVersionType string
+
+const (
+	ChangesVersionTypeRevTreeID ChangesVersionType = "rev" // Use revtree IDs in changes feed entries
+	ChangesVersionTypeCV        ChangesVersionType = "cv"  // Use current version in changes feed entries
+)
+
+func ParseChangesVersionType(s string) (ChangesVersionType, error) {
```
torcolvin, line 56: Add docstring that explains that "" will default to revtree if not specified.

db/changes.go
```diff
@@ -517,15 +568,15 @@ func makeChangeEntry(logEntry *LogEntry, seqID SequenceID, channel channels.ID)
...
+func makeRevocationChangeEntry(ctx context.Context, logEntry *LogEntry, seqID SequenceID, channel channels.ID, versionType ChangesVersionType) ChangeEntry {
+	entry := makeChangeEntry(ctx, logEntry, seqID, channel, versionType)
 	entry.Revoked = true
 
 	return entry
 }
 
 // AuditReadEvent issues a read event for this change entry. If there is no document body, there will be no event used.
-func (ce *ChangeEntry) AuditReadEvent(ctx context.Context) {
+func (ce *ChangeEntry) AuditReadEvent(ctx context.Context, versionType ChangesVersionType) {
```
torcolvin: I think you could actually avoid changing this API this by just figuring out what ce.changes[0] key is.

It might also be possible to just change `ce.ChangeVersionString` to just return whichever key is present, since it seems like it should always use `cv` if it is present and otherwise fall back to `rev`

One way I thought of to make `ce.ChangeVersionString` easier (or other code) is to change `ChangeByVersionType` to be a different type of struct:

```
type ChangeVersion struct {
	revtreeID string
	cv string
}

// basically this validates that we never have both when we go to marshal the data to conform to the rest api or blip api?
func (c *ChangeVersion) MarshalJSON() ([]byte, error) {
	if c.revtreeID != "" && c.cv != "" {
		return nil, fmt.Errorf("having both revtree %q and cv %q is invalid", c.revtreeID, c.cv)
	}
	<marshal as map>
}
```

rest/utilities_testing.go
```diff
@@ -883,8 +883,8 @@ func (cr ChangesResults) Summary() string {
 	return strings.Join(revs, ", ")
 }
 
-// RequireChangeRevVersion asserts that the given ChangeRev has the expected version for a given entry returned by _changes feed
-func RequireChangeRevVersion(t *testing.T, expected DocVersion, changeRev db.ChangeRev) {
+// RequireChangeRevVersion asserts that the given ChangeByVersionType has the expected version for a given entry returned by _changes feed
+func RequireChangeRevVersion(t *testing.T, expected DocVersion, changeRev db.ChangeByVersionType) {
 	RequireDocVersionEqual(t, expected, DocVersion{RevTreeID: changeRev["rev"]})
```
torcolvin: Is this always supposed to assert on rev here, if so rename this function with revtree id, otherwise make the assertion more flexible for cv.

db/crud.go
torcolvin, line 364: update docstring to replace revid with revOrCV and explain cv

torcolvin (commented):
There are some general thoughts in this code review but the

db/changes.go
bbrks: I think this is more accurately described as `ChangesByChangesVersionType` but that feels very awkward. Any suggestions? Otherwise I'll just use it.

torcolvin: If this was `ChangeRev` then `ChangeVersion` doesn't seem weird to me, but I don't have a better idea.

You could probably drop Type and just do `ChangeByVersion` if you think that's better.

db/crud.go
bbrks: Some of this is tied to CBG-4776 because there will be a new requirement for passing "one of cv or revid" down from REST APIs in a fetch (which eventually reaches this codepath)

The problem is all of these areas of rev/cv have been added piecemeal with no real design or consistency - which I completely agree is a problem, but one that doesn't _need_ solving right now.

`string` is dumb but achieves this "oneOf" requirement easily and with sufficient naming and documentation I don't see it being a problem in reality. If we're accidentally passing in docID or something then we'll find out about that pretty quickly in our tests.

db/changes.go
bbrks: What does this change give us other than indirection in the JSON Marshalling? Accessing fields by `.revtreeID` instead of `["rev"]`?

Do you think that's more beneficial than a closer representation of the wire-protocol for this data type? I'm not convinced

bbrks: Ok, I changed to use a `ce.ChangeVersionString()`, which just returns the first version type it found in `ce`

We have to assume there's only one version type present with this approach - which _should_ be true. That's the cost of being less explicit with preferred version type than the approach previously in this PR.

db/changes.go
bbrks: You're right - `get1xRevFromDoc` isn't going to work with CV.

I'll get test coverage of this, but maybe just queue wiring this code up behind CBG-4776 - that's required to fetch body by CV in REST API and would use same code.

db/changes.go
bbrks: I think it is better to return `row` instead of `nil`, which will allow the irretrievable body to be skipped in the error case.

There was no existing codepath that would result in `AddDocInstanceToChangeEntry` returning an error so this is a new edge-case just for CVs.

db/changes.go
bbrks: Replaced with `%q` to make the empty string logging clearer.

There's already a ticket filed and a TODO to make BLIP use the version type option, but it's not really relevant for this PR.

I kind of agree on setting explicit `VersionType` values and removing the fallbacks, but it's affecting BLIP code that isn't implemented yet, and it would technically introduce more error codepaths deeper into changes code, which are awkward to handle, since there's no fallback available.

db/changes.go
bbrks: I added an `AssertfCtx` - all of this codepath isn't set up to handle bubbling up errors, but realistically it's an assertion failure anyway and at least avoids a potential panic in prod.

db/changes.go
bbrks: Keeping as `ChangeByVersionType`, since I think it makes it clearer the map keys inside the slice are denoting what version type is being used for the change.

`ChangeByVersion` makes me think the versions are the map keys.

torcolvin (approved):

PR #7660, opened 2025-08-04
CBG-3661: allow configurable sub-protocol version for active replicator

gregns1:
CBG-3661

- Adds ability to explicitly set sub-protocol version for active replicator

## Pre-review checklist

- [ ] Link upstream PRs

### DIFF
--- a/db/active_replicator.go
+++ b/db/active_replicator.go
@@ -222,12 +222,16 @@ func connect(arc *activeReplicatorCommon, idSuffix string) (blipSender *blip.Sen
 	cancelCtx, cancelFunc := context.WithCancel(context.WithoutCancel(ctx)) // separate cancel context from parent cancel context
 
 	var originPatterns []string // no origin headers for ISGR
+	var blipContext *blip.Context
+
+	// CBG-4780: WIll hard code to use < 4 protocols for now, as the ISGR doesn't support 4+ protocols.
+	arc.config.SupportedBLIPProtocols = []string{CBMobileReplicationV3.SubprotocolString()}
 	// NewSGBlipContext doesn't set cancellation context - active replication cancellation on db close is handled independently
-	// TODO: CBG-3661 ActiveReplicator subprotocol versions
-	// - make this configurable for testing mixed-version replications
-	// - if unspecified, default to v2 and v3 until VV is supported with ISGR, then also include v4
-	protocols := []string{CBMobileReplicationV3.SubprotocolString(), CBMobileReplicationV2.SubprotocolString()}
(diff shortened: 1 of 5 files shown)


db/active_replicator.go
```diff
@@ -222,12 +222,16 @@ func connect(arc *activeReplicatorCommon, idSuffix string) (blipSender *blip.Sen
 	cancelCtx, cancelFunc := context.WithCancel(context.WithoutCancel(ctx)) // separate cancel context from parent cancel context
 
 	var originPatterns []string // no origin headers for ISGR
+	var blipContext *blip.Context
+
+	// CBG-4780: WIll hard code to use < 4 protocols for now, as the ISGR doesn't support 4+ protocols.
+	arc.config.SupportedBLIPProtocols = []string{CBMobileReplicationV3.SubprotocolString()}
```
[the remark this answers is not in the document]
gregns1: Temporarily meant to be like this. Changes haven't been made to support version 4 yet

db/active_replicator_pull.go
```diff
@@ -30,6 +30,8 @@ func NewPullReplicator(ctx context.Context, config *ActiveReplicatorConfig) (*Ac
 	apr := ActivePullReplicator{
 		activeReplicatorCommon: replicator,
 	}
+	// CBG-4780: WIll hard code to use < 4 protocols for now, as the ISGR doesn't support 4+ protocols.
+	replicator.config.SupportedBLIPProtocols = []string{CBMobileReplicationV3.SubprotocolString()}
```
torcolvin: Are you intentionally dropping `CBMobileReplicationV2.SubprotocolString()`?

I'd move this into `newActiveReplicatorCommon` since this code is shared between push and pull.

I think there's two ways of thinking about it, but right now this is confusing.

1. Require `SupportedBLIPProtocols []string` as an argument when the active replicator is created to be non empty. Fail creating the active replicator if this isn't set. This would mean changing a lot of test code.
2. Have `ForceSupportedBLIPProtocols` work as an override for all protocols (but document this behavior). ActiveReplicator will choose supported protocols unless `ForceSupportedBLIPProtocols` is specified.

gregns1: Will move to common. This is only for testing. I don't expect the main body of code to be setting this. If nothing is set here `supportedSubprotocols()` will be used to create blip context.

torcolvin: Or you could just keep that logic in connect as is, check config and use it fi specificed

torcolvin (approved):

PR #7665, opened 2025-08-05
CBG-4288: some assertion cleanups for active replicator tests

gregns1:
CBG-4288

- Changed equal function top check rev ID of defined then CV of defined. 
- I have added some assertions to CV on `requireDocumentVersion` 
- Overall in two minds if this PR is even necessary, I will probably add better assertion functions for testing ISGR when we actually have some parts implemented. For example, until we have the new rev tree property implemented I can't realistically assert on both CV and revID on remote compared to local. Given we will only be sending HLV in messages. Same for < 4 protocol, we can't assert on what a CV will be at remote if the only property being sent is revID properties. So its not totally correct we can assert on both at the same time always. 

## Pre-review checklist

- [ ] Link upstream PRs

### DIFF
--- a/db/utilities_hlv_testing.go
+++ b/db/utilities_hlv_testing.go
@@ -32,11 +32,19 @@ func (v *DocVersion) String() string {
 	return fmt.Sprintf("RevTreeID: %s", v.RevTreeID)
 }
 
-func (v DocVersion) Equal(o DocVersion) bool {
-	if v.RevTreeID != o.RevTreeID {
-		return false
+// RevOrCVEqual compares two DocVersions for equality. It checks RevID if defined or CV is defined.
+func (v DocVersion) RevOrCVEqual(o DocVersion) bool {
+	if v.RevTreeID != "" && o.RevTreeID != "" {
+		if v.RevTreeID == o.RevTreeID {
+			return true
+		}
 	}
-	return true
+	if !v.CV.IsEmpty() && !o.CV.IsEmpty() {
+		if v.CV.String() == o.CV.String() {
+			return true
+		}
+	}
+	return false
 }
 
 func (v DocVersion) GetRev(useHLV bool) string {
--- a/rest/replicatortest/replicator_test.go
+++ b/rest/replicatortest/replicator_test.go
@@ -2072,7 +2072,7 @@ func TestActiveReplicatorPullBasic(t *testing.T) {
 	rt2.CreateUser(username, []string{username})
 
 	docID := t.Name() + "rt2doc1"
-	version := rt2.PutDoc(docID, `{"source":"rt2","channels":["`+username+`"]}`)
+	version := rt2.PutDocDirectly(docID, rest.JsonToMap(t, `{"source":"rt2","channels":["`+username+`"]}`))
 
 	rt2collection, rt2ctx := rt2.GetSingleTestDatabaseCollection()
 	remoteDoc, err := rt2collection.GetDocument(rt2ctx, docID, db.DocUnmarshalAll)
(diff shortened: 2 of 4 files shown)


db/utilities_hlv_testing.go
```diff
@@ -32,11 +32,19 @@ func (v *DocVersion) String() string {
...
+			return true
+		}
+	}
+	if !v.CV.IsEmpty() && !o.CV.IsEmpty() {
+		if v.CV.String() == o.CV.String() {
+			return true
+		}
 	}
-	return true
+	return false
```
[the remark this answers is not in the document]
bbrks: lol wtf no

db/utilities_hlv_testing.go
```diff
@@ -32,11 +32,19 @@ func (v *DocVersion) String() string {
...
 }
```
torcolvin, line 48: I don't love this function, because I think the behavior should be explicit if we can for specific functions, especially while ISGR really needs `DocVersion.RevTreeEqual`:

- `DocVersion.RevTreeEqual`
- `DocVersion.CVEqual`

I think for this commit we mostly need `DocVersion.RevTreeEqual`.

I also could keep this function with the idea to move the behavior into `DocVersion.Equal` in a specific future ticket number. I think changing `DocVersion.Equal` is a future ticket, but I'm not sure exactly which one.

rest/replicatortest/replicator_test.go
```diff
@@ -2114,7 +2114,7 @@ func TestActiveReplicatorPullBasic(t *testing.T) {
 	doc, err := rt1collection.GetDocument(rt1ctx, docID, db.DocUnmarshalAll)
 	assert.NoError(t, err)
 
-	requireDocumentVersion(t, version, doc)
+	requireDocumentVersion(t, version, doc, false)
```
torcolvin, line 2117: I would rather do something like this here:

```
rt1DocVersion, rt1Body := rt1.GetDoc(docID)
RequireDocVersionEqual(t, version, rt1DocVersion)
require.Equal(t, "rt2", rt1Body["source"])
```

which I think is temporarily:

```
rt1DocVersion, rt1Body := rt1.GetDoc(docID)
// CBG-4780: switch assertion to RequireDocVersionEqual(t, version, rt1DocVersion)
// <insert ticket here to change RequireDocVersionEqual to use CV, or create a temporary function to do full equality
RequireDocVersionRevTreeEqual(t, version, rt1DocVersion)
require.Equal(t, "rt2", rt1Body["source"])
```

db/utilities_hlv_testing.go
```diff
@@ -32,11 +32,19 @@ func (v *DocVersion) String() string {
 	return fmt.Sprintf("RevTreeID: %s", v.RevTreeID)
 }
 
-func (v DocVersion) Equal(o DocVersion) bool {
-	if v.RevTreeID != o.RevTreeID {
-		return false
+// RevOrCVEqual compares two DocVersions for equality. It checks RevID if defined or CV is defined.
+func (v DocVersion) RevOrCVEqual(o DocVersion) bool {
```
torcolvin, line 36: Do you think this function should return false if there the version is empty? That might be better suited to `RequireDocVersionEqual*` functions

torcolvin (commented):
I can accept this PR as is, it is an improvement over before.

I don't really like the `requireDocumentVersion` which I'm not sure is necessary since it mostly just calls a single function underneath. I think having a parameter that is boolean means it is not clear from reading the code what `false` or `true` means. I think if we keep the `requireDocumentVersion` we should just have it say `requireDocumentVersionRevTree` to indicate what it is actually doing.

I think we should be cautious about how much time is spent on this change, but it may be that the suggestions I made below are easy. If you don't like the suggestions, I can definitely approve this change.

PR #7668, opened 2025-08-07
CBG-4780: basic ISGR push and pull for 4.0

gregns1:
CBG-4780

- PR does the minimum to get ISGR running in 4.0 replication. Conflicts are not handled yet. 
- Some test are skipped due to conflict issues (not implemented) 
- Some test assertions are changed given we don't currently keep rev IDs in sync each side of the replication (diff revIDs being generated each side even on new doc writes) so CV is only asserted at this time. I have linked tickets to change these when work takes place in other tickets. 
- I have also linked work in the write REST API work to change assertions based on that work too
- One particular test is forced to run in 3.0 mode due to asserting docs created with same revIDs and bodies each each side get marked as peer doesn't want and asserts some check pointing behaviour around this. This is not applicable to 4.0 replication. Given we don’t generate same versions for docs with same id's and bodies like we do with revIDs. 
- New function to check incoming changes on changes message. I have tried to keep same sort of structure as RevsDiff given I think we can use known revs property for the rev tree property added later. 
- I will enhance conflict detecting for push replications on changes message in ticket CBG-4792 (similar behaviour to what `proposeChanges` does) 
- I have moved assertion changes from #7665 to here and will close it 

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/3242/


rest/replicatortest/replicator_test.go
```diff
@@ -8035,50 +8102,6 @@ func requireBodyEqual(t *testing.T, expected string, doc *db.Document) {
 	require.Equal(t, expectedBody, doc.Body(base.TestCtx(t)))
 }
 
-// TestReplicatorUpdateHLVOnPut:
-//   - For purpose of testing the PutExistingRev code path
-//   - Put a doc on a active rest tester
-//   - Create replication and wait for the doc to be replicated to passive node
-//   - Assert on the HLV in the metadata of the replicated document
-func TestReplicatorUpdateHLVOnPut(t *testing.T) {
```
bbrks, line 8043: Try a REST API PUT with `new_edits=false` to ensure we're generating a CV for a rev-ID write.

bbrks (approved):

PR #7681, opened 2025-08-13
CBG-4790: reconcile rev tree on write for ISGR

gregns1:
CBG-4790

- PR to take the new `revTree` property in ISGR and includes those revisions in the write
- There are some test cases that I would still like to write but require conflict resolution of some sort to be in place 
- Changes to tests that were temporarily asserting on CV only due to missing reconciliation of revTree. These tests now assert on CV and revID.

Updated some info on ISGR changes thus far in this document: https://docs.google.com/document/d/1SrN_om0pBm3yGnZF1cenFrkODhwMYWfPkjPbTq_2cEE/edit?usp=sharing 

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/32/ (test flake)


db/blip_sync_context.go
```diff
@@ -386,18 +386,28 @@ func (bsc *BlipSyncContext) handleChangesResponse(ctx context.Context, sender *b
...
 					if err != nil {
-						// assume we have received legacy rev if we cannot parse hlv from known revs, and we are in vv replication
+						// assume we have received legacy rev if the following conditions are met:
+						//  - we cannot parse cv from known revs
+						//  - we are in vv enabled replication
+						// 	- the first element of known rev array was NOT a CV
 						if versionVectorProtocol {
-							legacyRev = true
+							if !cvInKnownRevs {
+								// if cv wasn't the first element in known revs array, this will be legacy doc
```
torcolvin: This should probably never happen, but this comment isn't correct:

If you had known revs as:

["1@abc", "1-abc", "2@def"]

this would end up incorrect. I find this logic incredibly hard to follow and I was thinking whether we could split this into a function with some table driven inputs.

I'm thinking something like:

```getKnownRevs(changeRequestRow []any) (deltaSrcRevID string, knownRevs map[string]bool)```

Then we could list valid conditions for

- 3.x replications
- 4.x replications
- mixed?

I am having trouble holding in my head what the expected input and output is, and having a test for this can serve as a type of documentation.

File: db/crud.go [hunk 2]
```diff
@@ -1211,7 +1211,7 @@ func (db *DatabaseCollectionWithUser) Put(ctx context.Context, docid string, bod
 	return newRevID, doc, err
 }
 
-func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Context, newDoc *Document, newDocHLV *HybridLogicalVector, existingDoc *sgbucket.BucketDocument, revTreeHistory []string) (doc *Document, cv *Version, newRevID string, err error) {
+func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Context, newDoc *Document, newDocHLV *HybridLogicalVector, existingDoc *sgbucket.BucketDocument, revTreeHistory []string, revTreeISGRProperty bool) (doc *Document, cv *Version, newRevID string, err error) {
```
torcolvin: I think you should describe the arguments to this function in a docstring.

I'm trying to wrap my head around why you'd have revTreeHistory but revTreeISGRProperty is false. This could be done in some type of test as well.

One way to make this more clear (not sure) is to have something that is:

```
type putRevTreeHistory struct {
      history []string // is this always going be ["1-abc", "2-abc"]
      forceAlignment bool // do not check for revtree conflicts?
}
```

I don't like the above because I'm in fact not sure what this does.

My very shaky understanding is:

1. revTreeHistory could contain the incoming cv
2. revTreeHistory could contain the revtree from the incoming source? which would be the previous revisions? Could this be a 2@abc, 1-def?

Do we think that we should clarify with some type of typed information to make the subsequent calls easier to understand?

I really don't think this should mention ISGR at all, that's ultimately an implementation detail upstream. The boolean `revTreeISGRProperty` controls:

1. If true, generate a new revtree id rather than putting the existing revtree id
2. If true, will take all of this revtree history and add it to the existing revtree? or overwrite the history if there are conflicts?
3. avoid rev tree history conflict checking?

Because this behavior is connected to `revHistory` that's why I thought they might belong in the same struct.

db/crud.go
```diff
@@ -3328,6 +3366,47 @@ func (db *DatabaseCollectionWithUser) CheckProposedVersion(ctx context.Context,
...
+	}
+
+	err := doc.addNewerRevisionsToHistory(newDoc, currentRevIndex, parent, revTreeHistory)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (doc *Document) addNewerRevisionsToHistory(newDoc *Document, currentRevIndex int, parent string, docHistory []string) error {
```
torcolvin: It feels worth describing this function, potentially with examples?

- What is `currentRevIndex`?
- What is `parent`?
- Should this function indicate that only affects RevTree? since revisions can imply hlv type revisions but this is strictly for revtreeids
- When can this error? It seems like this will error if any revtreeids  in `docHistory` exist in `doc.History` already or `parent` exists is `doc.History`

db/crud.go
```diff
@@ -1317,24 +1343,15 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 		// rev tree conflict check if we have rev tree history to check against + finds current rev index to allow us
 		// to add any new revision to rev tree below.
 		// Only check for rev tree conflicts if we haven't already checked above
-		if !revTreeConflictChecked && len(revTreeHistory) > 0 {
+		if !revTreeConflictChecked && len(revTreeHistory) > 0 && !revTreeISGRProperty {
```
torcolvin: Should we declare `revTreeConflictChecked` true if `revTreeISGRProperty` is set above? If not, I think that we should add to comment above why this isn't checked

rest/replicatortest/replicator_revtree_test.go
```diff
@@ -0,0 +1,382 @@
...
+
+	// call CheckChangeVersion with fake VV and the docID from above
+	missing, possible := collectionRT1.CheckChangeVersion(rt1Ctx, t.Name(), "123@src")
+
+	// assert that the missing revision is the CV and known revID is current revIDS of the document
+	assert.Len(t, missing, 1)
+	assert.Len(t, possible, 2)
+	// rt1's CV and revID of the doc should be returned as known revs
+	assert.Equal(t, versionRT1.CV.String(), possible[0])
+	assert.Equal(t, versionRT1.RevTreeID, possible[1])
```
torcolvin:
```suggestion
	assert.Equal(t, []string{"123@src"}, missing)
	assert.Equal(t,  []string{versionRT1.CV.String(),  versionRT1.RevTreeID}, possible[1])
```

rest/replicatortest/replicator_revtree_test.go
```diff
@@ -0,0 +1,382 @@
...
+
+	"github.com/couchbase/sync_gateway/base"
+	"github.com/couchbase/sync_gateway/channels"
+	"github.com/couchbase/sync_gateway/db"
+	"github.com/couchbase/sync_gateway/rest"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestKnownRevsForCheckChangeVersion(t *testing.T) {
```
torcolvin: Move this test into db package? I don't think this needs rest tester at all

rest/replicatortest/replicator_revtree_test.go
```diff
@@ -0,0 +1,382 @@
...
+		ChangesBatchSize:    200,
+		ReplicationStatsMap: dbReplicatorStats(t),
+		CollectionsEnabled:  !rt1.GetDatabase().OnlyDefaultCollection(),
+		Continuous:          false,
+	})
+	require.NoError(t, err)
+	defer func() { assert.NoError(t, ar.Stop()) }()
+
+	// Start the replicator
+	assert.NoError(t, ar.Start(ctx1))
```
torcolvin:
```suggestion
	require.NoError(t, ar.Start(ctx1))
```

rest/replicatortest/replicator_revtree_test.go
```diff
@@ -0,0 +1,382 @@
...
+
+	docHistoryList := make([]string, 0, 11)
+	docID := t.Name() + "rt2doc1"
+	version := rt2.PutDocDirectly(docID, rest.JsonToMap(t, `{"source":"rt1","channels":["alice"]}`))
+	docHistoryList = append(docHistoryList, version.RevTreeID)
+
+	rt2collection, rt2ctx := rt2.GetSingleTestDatabaseCollection()
+	remoteDocRT2, err := rt2collection.GetDocument(rt2ctx, docID, db.DocUnmarshalAll)
+	require.NoError(t, err)
+	assert.Equal(t, version.RevTreeID, remoteDocRT2.CurrentRev)
```
torcolvin: Should this end up being:

rt2Version, _ := rt2.GetDoc(docID)

Since we presumably want to check for CV? Or was this part of something that you were verifying the behavior?

rest/replicatortest/replicator_revtree_test.go
```diff
@@ -0,0 +1,382 @@
...
+	remoteDocRT1, err := rt1collection.GetDocument(rt1ctx, docID, db.DocUnmarshalAll)
+	require.NoError(t, err)
+	rest.RequireDocVersionEqual(t, version, remoteDocRT1.ExtractDocVersion())
+
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		assert.Equal(c, db.ReplicationStateStopped, ar.GetStatus(ctx1).Status)
+	}, time.Second*20, time.Millisecond*100)
+
+	for i := 0; i < 10; i++ {
+		version = rt2.UpdateDocDirectly(docID, version, rest.JsonToMap(t, `{"source":"rt1","channels":["alice"]}`))
```
torcolvin: I think for debugging, you should add something to indicate what revision of the doc this is, `"write": i`

rest/replicatortest/replicator_revtree_test.go
```diff
@@ -0,0 +1,382 @@
...
+
+	// wait for the document originally written to rt1 to arrive at rt2
+	changesResults := rt1.WaitForChanges(1, "/{{.keyspace}}/_changes?since=0", "", true)
+	assert.Equal(t, docID, changesResults.Results[0].ID)
+
+	// assert on version
+	rt1collection, rt1ctx := rt1.GetSingleTestDatabaseCollection()
+	remoteDocRT1, err := rt1collection.GetDocument(rt1ctx, docID, db.DocUnmarshalAll)
+	require.NoError(t, err)
+	rest.RequireDocVersionEqual(t, version, remoteDocRT1.ExtractDocVersion())
```
torcolvin: Here, why not use rt1.GetDoc() to get the docversion?

rest/replicatortest/replicator_revtree_test.go
```diff
@@ -0,0 +1,382 @@
...
+	// start again for new revisions
+	assert.NoError(t, ar.Start(ctx1))
+
+	// wait for the document originally written to rt1 to arrive at rt2
+	changesResults = rt1.WaitForChanges(1, "/{{.keyspace}}/_changes?since="+changesResults.Last_Seq.String(), "", true)
+	assert.Equal(t, docID, changesResults.Results[0].ID)
+
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		assert.Equal(c, db.ReplicationStateStopped, ar.GetStatus(ctx1).Status)
+	}, time.Second*20, time.Millisecond*100)
```
torcolvin: For readability, I would be inclined to use:

```
rt1.WaitForReplicationStatus(replicationID, db.ReplicationStateStopped)
```

you might need to use `RestTester.CreateReplication`

rest/replicatortest/replicator_revtree_test.go
```diff
@@ -0,0 +1,382 @@
...
+
+	rt1Doc, err := rt1collection.GetDocument(rt1ctx, docID, db.DocUnmarshalAll)
+	require.NoError(t, err)
+	assert.Equal(t, version.RevTreeID, rt1Doc.CurrentRev)
+	assert.Len(t, rt1Doc.History.GetLeaves(), 1)
+	assert.Len(t, rt1Doc.History, 11) // 1 base + 10 updates
+
+	rest.RequireDocVersionEqual(t, version, rt1Doc.ExtractDocVersion())
+	assert.ElementsMatch(t, slices.Collect(maps.Keys(rt1Doc.History)), docHistoryList)
```
torcolvin: I created base.RequireKeysEqual for this if you want

rest/replicatortest/replicator_revtree_test.go
```diff
@@ -0,0 +1,382 @@
...
+}
+
+func TestActiveReplicatorPushRevTreeReconciliation(t *testing.T) {
```
torcolvin: Is this just the same test as the pull test? should these be parameterized? If this makes it too complicated I don't want to do that, but also if it really is mostly the same code then it might be more helpful?

The same is true for large size tests. If it doesn't make sense to use push and pull, does it make sense to parameterize doc numbers for the test?

db/crud.go
```diff
@@ -3328,6 +3366,47 @@ func (db *DatabaseCollectionWithUser) CheckProposedVersion(ctx context.Context,
 	}
 }
 
+func (doc *Document) alignRevTreeHistory(ctx context.Context, newDoc *Document, revTreeHistory []string) error {
```
torcolvin, line 3382: Add docstring and table test for this function with the expectations.

My understanding is that align might actually mean merge and overwrite if required, but it would be easier to see if we can have concrete examples.

rest/replicatortest/replicator_revtree_test.go
gregns1: I want to be using the active replicator object here not the rest api for these so will have to leave these assertions as is. 

Its on roadmap to have these active replicator tests running like blip tester client does in two modes, it may not make sense to have these running in two modes given they test 4.0 features but I would still like that option.

rest/replicatortest/replicator_revtree_test.go
gregns1: Parameterized these tests

db/crud.go
gregns1: Done and added test

db/crud.go
gregns1: Added comment here for clarity, will be updating the doc in the description with more legacy rev handling info soon too for team knowledge.

db/crud.go
gregns1: Changed the naming and added comment to describe input parameters to the function.

db/blip_sync_context.go
gregns1: Moved to function and test runs on that function.

db/crud.go
gregns1: I will be documenting legacy rev handling better soon too for knowledge of the team given this area can get very confusing in the code IMO.

db/blip_sync_context.go
```diff
@@ -870,3 +847,64 @@ func (bsc *BlipSyncContext) useHLV() bool {
...
+	knownRevs = make(map[string]bool)
+	// The first element of the knownRevsArray returned from CBL is the parent revision to use as deltaSrc for
+	// revtree clients. For HLV clients, use the cv as deltaSrc
+	if bsc.useDeltas && len(knownRevsArray) > 0 {
+		if revID, ok := knownRevsArray[0].(string); ok {
+			if bsc.useHLV() {
+				// extract cv from the known revs array
+				msgHLV, _, err := extractHLVFromBlipString(revID)
+				if err != nil {
+					base.DebugfCtx(ctx, base.KeySync, "Invalid known rev format for hlv on doc: %s falling back to full body replication.", base.UD(docID))
```
torcolvin: we are swallowing the error here, so we don't have much chance of ever debugging this

db/blip_sync_context.go
```diff
@@ -367,45 +366,23 @@ func (bsc *BlipSyncContext) handleChangesResponse(ctx context.Context, sender *b
...
+			//	- [revID] indicating the clients wants this doc but already has a revision and this revision is a pre
+			// 	upgraded revision (a document without a HLV yet as it hasn't been updated since upgrade)
+			//	ISGR ONLY:
+			// 	- [cv, revID] indicating the clients wants this doc but already has a revision and this revision. First
+			//	element must always be a CV here which will be used for delta sync if enabled. The subsequent revID
+			//	will be used to determine what revIDs will need to be included in the revTree property for ISGR
+
+			deltaSrcRevID, legacyRev, knownRevs, err = bsc.getKnownRevs(ctx, docID, knownRevsArray)
+			if err != nil {
+				base.ErrorfCtx(ctx, "Invalid response to 'changes' message")
```
torcolvin: log this error message

db/blip_sync_context.go
```diff
@@ -870,3 +847,64 @@ func (bsc *BlipSyncContext) useHLV() bool {
 func (bsc *BlipSyncContext) sendRevTreeProperty() bool {
 	return bsc.activeCBMobileSubprotocol >= CBMobileReplicationV4 && bsc.clientType == BLIPClientTypeSGR2
 }
+
+// getKnownRevs will return deltaSrcRev (if delta sync is enabled), whether the known rev array represents a legacy
+// document, the parsed known revs in a map and error if unknown format is received.
```
torcolvin:
```suggestion
// getKnownRevs will return deltaSrcRev (if delta sync is enabled), whether the known rev array represents a legacy
// document without an HLV, the parsed known revs in a map and error if unknown format is received.
```

db/blip_sync_context.go
```diff
@@ -870,3 +847,64 @@ func (bsc *BlipSyncContext) useHLV() bool {
...
+				// extract cv as string
+				revID = msgHLV.GetCurrentVersionString()
+				// mark that we have seen a CV in known revs so we can mark as non legacy rev
+				cvInKnownRevs = true
+			}
+			// we can assume here that if we fail to parse hlv, we have received a rev id in known revs. If we don't fail to parse hlv
+			// then we have extracted cv from it and can assign the cv string to known revs here
+			knownRevs[revID] = true
+		} else {
+			return deltaSrcRev, changeIsLegacyRev, knownRevs, errors.New("invalid known rev format")
```
torcolvin, line 901: log the rev body here, we probably will not have this.

db/blip_sync_context.go
```diff
@@ -870,3 +847,64 @@ func (bsc *BlipSyncContext) useHLV() bool {
...
+			// extract cv from the known revs array
+			msgHLV, _, err := extractHLVFromBlipString(revID)
+			if err != nil {
+				// assume we have received legacy rev if the following conditions are met:
+				//  - we cannot parse cv from known revs
+				//  - we are in vv enabled replication
+				// 	- the first element of known rev array was NOT a CV
+				if bsc.useHLV() {
+					if !cvInKnownRevs {
+						// if cv wasn't the first element in known revs array, this will be legacy doc
```
torcolvin:
```suggestion
						// if cv wasn't the first element in known revs array, assume all elements will be revtree ids.
```

db/crud.go
```diff
@@ -1211,7 +1211,15 @@ func (db *DatabaseCollectionWithUser) Put(ctx context.Context, docid string, bod
...
 }
 
-func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Context, newDoc *Document, newDocHLV *HybridLogicalVector, existingDoc *sgbucket.BucketDocument, revTreeHistory []string) (doc *Document, cv *Version, newRevID string, err error) {
+// PutExistingCurrentVersion:
+//   - newDoc: new incoming doc
+//   - newDocHLV: new incoming doc's HLV
+//   - existsingDoc: existing doc in bucket (if present)
+//   - revTreeHistory: list of revID's from the incoming docs history (including docs current rev).
+//   - alignRevTrees: if this is true then we will align the new write with the incoming docs rev tree. If this is
+//     false and len(revTreeHistory) > 0 then this meaans the lcoal version of this doc is legacy doc so this parameter
```
```suggestion
//     false and len(revTreeHistory) > 0 then this means the local version of this doc does not have an HLV so this parameter
```

db/crud.go
```diff
@@ -3147,6 +3180,14 @@ func (db *DatabaseCollectionWithUser) CheckChangeVersion(ctx context.Context, do
 		missing = append(missing, rev)
 		return
 	}
+	if doc.HLV == nil {
+		// no hlv on local doc, mark as missing but send current rev as known rev (will be handled as legacy
+		// rev document on changes response handler)
+		base.DebugfCtx(ctx, base.KeyChanges, "Doc %s has no HLV, marking change version %s as missing", base.UD(docid), base.UD(rev))
```
torcolvin: Do you think debug is too verbose for this? This seems common. If this is useful for testing, maybe KeyVV?

db/crud.go
```diff
@@ -3328,6 +3376,56 @@ func (db *DatabaseCollectionWithUser) CheckProposedVersion(ctx context.Context,
...
+	for i, revid := range revTreeHistory {
+		if doc.History.contains(revid) {
+			currentRevIndex = i
+			parent = revid
+			break
+		}
+	}
+
+	if parent != doc.CurrentRev {
+		base.DebugfCtx(ctx, base.KeyCRUD, "incoming rev tree history has different history than local doc %s, aligning local doc history with incoming rev tree history", base.UD(doc.ID))
```
```suggestion
		base.DebugfCtx(ctx, base.KeyCRUD, "incoming rev tree history has different history than local doc %s, overwriting the revision history to match the incoming history", base.UD(doc.ID))
```

torcolvin (commented):
LGTM, some logging suggestions

db/blip_sync_context.go
gregns1: I don't think we want to error here fully given it can fall back to full body replication. Its possible that the first elem may be a revID (if the change response for this doc is a legacy rev) in this case we wouldn't want to error but maybe log to ensure its clear why its not being sent as a delta. 

I have added the error message and the revID string to the log line though so we can see why we fall into this.

db/crud.go
gregns1: Good point, i'll drop to trace

torcolvin (approved):

PR #7683, opened 2025-08-15
CBG-4806: move conflict detection to common function

gregns1:
CBG-4806

- moving conflict detection code to shared function to be used in a centralised place. This way blip tester client and `PutExistingCurrentVersion` work in the same way. 

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGateway-Integration/33/

### DIFF
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -9,7 +9,9 @@
 package db
 
 import (
+	"context"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"maps"
 	"sort"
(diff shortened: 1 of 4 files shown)


rest/utilities_testing_blip_client.go
```diff
@@ -325,35 +324,16 @@ func (cd *clientDoc) _hasConflict(t testing.TB, incomingHLV *db.HybridLogicalVec
...
-	// - CBL writes document 1@cbl1
-	// - CBL pushes to SG as 1@cbl1
-	// - CBL pulls document 1@cbl1
-	//
-	// NOTE: without P2P replication, this should not be the case and we would not get this revision, since CBL
-	// would respond to a SG changes message that CBL does not need this revision
-	if localHLV.DominatesSource(*incomingCV) {
+	inConflict, err := db.IsInConflict(nil, &localHLV, incomingHLV)
+	if err != nil {
 		require.FailNow(t, fmt.Sprintf("incoming CV %#+v has lower version than the local revision %#+v - this should've been filtered via changes response before ending up as a rev. blip tester would reply that to Sync Gateway that it doesn't need this revision", incomingCV, localHLV))
```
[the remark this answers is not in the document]
gregns1: Will leave this as is given the extra context in original message.

db/hybrid_logical_vector.go
```diff
@@ -689,3 +691,50 @@ func (hlv *HybridLogicalVector) UnmarshalJSON(inputjson []byte) error {
...
+//   - Local CV matches incoming CV, so no new versions to add
+func IsInConflict(ctx context.Context, localHLV, incomingHLV *HybridLogicalVector) (bool, error) {
+	incomingCV := incomingHLV.ExtractCurrentVersionFromHLV()
+	localCV := localHLV.ExtractCurrentVersionFromHLV()
+
+	// check if incoming CV and local CV are the same. This is needed here given that if both CV's are the same the
+	// below check to check local revision is newer than incoming revision will pass given the check and will add the
+	// incoming versions even if CV's are the same
+	if localCV.Equal(*incomingCV) {
+		base.DebugfCtx(ctx, base.KeyCRUD, "incoming CV %#+v is equal to local revision %#+v", incomingCV, localCV)
```
torcolvin: This is going to be way too noisy to leave at debug, I might put this at KeyVV

db/hybrid_logical_vector.go
```diff
@@ -689,3 +691,50 @@ func (hlv *HybridLogicalVector) UnmarshalJSON(inputjson []byte) error {
...
+// or not and will error for the following cases:
+//   - Local HLV dominates incoming HLV (meaning local version is a newer version that the incoming one)
+//   - Local CV matches incoming CV, so no new versions to add
+func IsInConflict(ctx context.Context, localHLV, incomingHLV *HybridLogicalVector) (bool, error) {
+	incomingCV := incomingHLV.ExtractCurrentVersionFromHLV()
+	localCV := localHLV.ExtractCurrentVersionFromHLV()
+
+	// check if incoming CV and local CV are the same. This is needed here given that if both CV's are the same the
+	// below check to check local revision is newer than incoming revision will pass given the check and will add the
+	// incoming versions even if CV's are the same
```
torcolvin:
```suggestion
	// incoming versions even if CVs are the same
```

db/hybrid_logical_vector.go
```diff
@@ -689,3 +691,50 @@ func (hlv *HybridLogicalVector) UnmarshalJSON(inputjson []byte) error {
...
+	if localCV.Equal(*incomingCV) {
+		base.DebugfCtx(ctx, base.KeyCRUD, "incoming CV %#+v is equal to local revision %#+v", incomingCV, localCV)
+		return false, ErrNoNewVersionsToAdd
+	}
+
+	// standard no conflict case. In the simple case, this happens when:
+	//  - SG writes document 1@cbs1
+	//  - CBL pulls document 1@cbs1
+	//  - SG writes document 2@cbs1
```
torcolvin: Now that we have moved this out of the blip tester code, let's make the examples more general.

- localHLV cv:1@src1
- remoteHLV cv:2@src1

Let's do this for all the comments.

There are other cases that aren't the simple case, where 1@src1 would exist in mv or pv, but I don't think they need to be listed.

db/hybrid_logical_vector.go
```diff
@@ -689,3 +691,51 @@ func (hlv *HybridLogicalVector) UnmarshalJSON(inputjson []byte) error {
...
+//   - Local CV matches incoming CV, so no new versions to add
+func IsInConflict(ctx context.Context, localHLV, incomingHLV *HybridLogicalVector) (bool, error) {
+	incomingCV := incomingHLV.ExtractCurrentVersionFromHLV()
+	localCV := localHLV.ExtractCurrentVersionFromHLV()
+
+	// check if incoming CV and local CV are the same. This is needed here given that if both CV's are the same the
+	// below check to check local revision is newer than incoming revision will pass given the check and will add the
+	// incoming versions even if CVs are the same
+	if localCV.Equal(*incomingCV) {
+		base.TracefCtx(ctx, base.KeyVV, "incoming CV %#+v is equal to local revision %#+v", incomingCV, localCV)
```
torcolvin, line 710: I generally like %#+v for logging, but I think %#v might be better for HLV and CV since it displays versions in decimal instead of a hexadecimal.

Can you check this? If this is true, can you file a ticket to create a `ruleguard` rule so that we can detect this for the future?

If not, disregard this comment.

gregns1: It only seems to wrap the formatted bit into quotes. 

Internally we handle the HLV in uint64 type so logging here for example should be fine although %#v seems to format a uint64 into hex.

torcolvin: I think it's worth filing a ticket to evaluate the places we log this and make sure we don't log in hex, it just makes everything harder to handle. But I can assert https://github.com/couchbaselabs/sg_dev_tools/pull/37 will handle this too.

torcolvin (approved):

db/hybrid_logical_vector.go
gregns1: https://jira.issues.couchbase.com/browse/CBG-4811

PR #7687, opened 2025-08-18
CBG-4813 refactor tests to use consistent HLV format

torcolvin:
CBG-4813 refactor tests to use consistent HLV format so these test cases can be lifted to validate CBL.

@adamcfraser can you validate the expected output for the FIXME tests and see if there are other tests that should be added?

## Pre-review checklist

### DIFF
--- a/db/hybrid_logical_vector_test.go
+++ b/db/hybrid_logical_vector_test.go
@@ -9,9 +9,7 @@
 package db
 
 import (
-	"encoding/base64"
 	"math/rand/v2"
-	"reflect"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


db/hybrid_logical_vector_test.go
```diff
@@ -103,61 +101,61 @@ func TestInternalHLVFunctions(t *testing.T) {
...
+			inputListHLVA:  "20@cluster1;10@cluster2",
+			inputListHLVB:  "10@cluster2;15@cluster1",
 			expectedResult: true,
 		},
 		{
 			name:           "B CV older than A's PV for same source",
-			inputListHLVA:  []string{"cluster1@20", "cluster2@10"},
-			inputListHLVB:  []string{"cluster2@10", "cluster1@15"},
+			inputListHLVA:  "20@cluster1;10@cluster2",
+			inputListHLVB:  "10@cluster2;15@cluster1",
```
adamcfraser: This matches the previous test, I think inputListHLVA is supposed to be `"20@cluster1;15@cluster2"`

db/hybrid_logical_vector_test.go
```diff
@@ -103,61 +101,61 @@ func TestInternalHLVFunctions(t *testing.T) {
 func TestConflictDetectionDominating(t *testing.T) {
```
adamcfraser: Some MV cases for this test would make sense (particularly since createHLVForTest should now be able to parse those from string).  Cases like:
1. a.MV entry dominates b.CV (expected true)
2. a.MV entry doesn't dominate b.CV (expected false)
3. b.CV.source occurs in both a.CV and a.MV, dominates both (expected true)
3. b.CV.source occurs in both a.CV and a.MV, dominates only a.MV (expected false)

db/hybrid_logical_vector_test.go
```diff
@@ -171,69 +169,98 @@ func TestConflictDetectionDominating(t *testing.T) {
...
+			incomingHLV: "4@c,5@b,6@a",
+			finalHLV:    "4@c,5@b,6@a",
 		},
+		/* FIXME, this does not work yet.
+		{
+			name:        "incoming does not dominate pv",
+			existingHLV: "3@c;5@b,6@a",
+			incomingHLV: "4@c,1@b,2@a",
+			finalHLV:    "3@c;5@b,6@a",
+		},
```
adamcfraser, line 262: Does 'not work' mean that AddNewerVersions shouldn't have been called in the first place (since incoming doesn't dominate pv?).  I see how we might not want to put guardrails for that in AddNewerVersions, because it's computation overhead that we expect to have already completed by a previous call to IsDominating?

db/hybrid_logical_vector_test.go
```diff
@@ -171,69 +169,98 @@ func TestConflictDetectionDominating(t *testing.T) {
...
+		*/
+		{
+			name:        "incoming mv partially overlaps with pv",
+			existingHLV: "3@c;2@b,1@a",
+			incomingHLV: "4@c,2@b,6@a",
+			finalHLV:    "4@c,2@b,6@a",
+		},
+		/* FIXME: this doesn't work yet
+		{
+			name:        "incoming mv partially overlaps with pv, but incoming mv should be wiped out",
```
adamcfraser: I think this is an impossible scenario (where `3@c` has `6@a` in its history, but `4@c` represents a merge that includes `1@a`).  Can you come up with a sequence of mutations that would lead to this?

Or are you thinking about this as a type of repair scenario for invalid input?

db/hybrid_logical_vector_test.go
```diff
@@ -1072,191 +1099,51 @@ func stringHexToUint(t testing.TB, value string) uint64 {
...
-				SourceID: "a",
-				Version:  2,
-				PreviousVersions: map[string]uint64{
-					"b": 2,
-					"d": 1,
-				},
-			},
+			name:        "cv mv pv, duplicate sources",
+			existingHLV: "2@a,2@b,2@c;1@b,1@d", // this is an invalid HLV, since a source ID exists in mv and pv
+			expectedHLV: "2@a;1@b,2@c,1@d",
```
adamcfraser: Similar to the above, maybe we should be thinking about this in terms of "invalidateMV will repair invalid MV/PV combinations"?

db/hybrid_logical_vector_test.go
torcolvin: I agree this is an impossible situation if an HLV is well formed. However, an HLV could come from a replication where it is malformed.

After offline discussion, `AddNewerVersions` should:

 - Preserve incoming merge versions even if they happen to be older than existing pv. In this case, the existing pv should be wiped out.
 - Keep mv if there is a singular value. This shouldn't happen, but since Sync Gateway doesn't process merge values, there's no harm in leaving this alone.
 - If incoming mv has merge versions with duplicate IDs to pv that exist on the incoming mv, move everything from mv into pv and invalidate mv on the resulting HLV.

db/hybrid_logical_vector_test.go
torcolvin: InvalidateMV can only be called once it is parsed from the wire format, so I'm not very worried about that for incoming from blip.

Removing this test.

adamcfraser (approved):
LGTM

PR #7693, opened 2025-08-20
CBG-4751: Allow CV to be used as OCC value in REST API for document writes (updates/deletes)

bbrks:
CBG-4751

- Allow CV to be used in OCC writes (Document Update/Delete and the associated attachment endpoints) using existing `rev` parameter - which can automatically detect RevTreeID/CV.
- Switch test helpers to use CV in writes when available for coverage, and allow fallback to RevTreeID. This is a prereq for getting #7658 merged  
- Prevent CV OCC value in doc updates for docs in conflict or against non-current versions, since we don't maintain linear version history like we do for RevTrees and can't correlate an old CV with a branched revision ID.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/47/
  - fixed in final commit


db/document.go
```diff
@@ -1484,3 +1485,60 @@
...
+func (v *DocVersion) RevIDGeneration() int {
+	if v == nil {
+		return 0
+	}
+	gen, err := strconv.ParseInt(strings.Split(v.RevTreeID, "-")[0], 10, 64)
+	if err != nil {
+		base.AssertfCtx(context.TODO(), "Error parsing generation from rev ID %q: %v", v.RevTreeID, err)
+		return 0
+	}
+	return int(gen)
```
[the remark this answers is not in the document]
bbrks: will switch to int64 to simplify

db/crud.go
```diff
@@ -3342,6 +3373,7 @@ func (db *DatabaseCollectionWithUser) CheckProposedVersion(ctx context.Context,
 	// previousRev may be revTreeID or version
 	var previousVersion Version
 	previousRevFormat := "version"
+	// TODO: CBG-4812 Use base.IsRevTreeID
 	if !strings.Contains(previousRev, "@") {
```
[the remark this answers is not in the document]
bbrks: Intentionally delayed to mentioned Jira ticket.

db/blip_handler.go
```diff
@@ -852,6 +852,7 @@ func (bh *blipHandler) handleProposeChanges(rq *blip.Message) error {
 
 		changeIsVector := false
 		if versionVectorProtocol {
+			// TODO: CBG-4812 Use base.IsRevTreeID
 			changeIsVector = strings.Contains(rev, "@")
```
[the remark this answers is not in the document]
bbrks: Intentionally delayed to mentioned Jira ticket.

db/blip_handler.go
```diff
@@ -1094,6 +1095,7 @@ func (bh *blipHandler) processRev(rq *blip.Message, stats *processRevStats) (err
 	var incomingHLV *HybridLogicalVector
 	// Build history/HLV
 	var legacyRevList []string
+	// TODO: CBG-4812 Use base.IsRevTreeID
 	changeIsVector := strings.Contains(rev, "@")
```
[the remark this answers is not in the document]
bbrks: Intentionally delayed to mentioned Jira ticket.

db/crud.go
```diff
@@ -1146,21 +1150,42 @@ func (db *DatabaseCollectionWithUser) Put(ctx context.Context, docid string, bod
...
-					generation, _ = ParseRevID(ctx, matchRev)
-					generation++
+
+		// OCC check of matchCV against CV on the doc
+		if matchCV != "" {
+			if matchCV == doc.HLV.GetCurrentVersionString() {
+				// set matchRev to the current revision ID and allow existing codepaths to perform RevTree-based update.
+				matchRev = doc.CurrentRev
+				// bump generation based on retrieved RevTree ID
+				generation, _ = ParseRevID(ctx, matchRev)
```
torcolvin, line 1160: I'm pretty sure that this came from before here and this should never happen unless `matchRev` is not a valid revid, but should we do an error check and warn here?

I see the value in continuing and assuming the revision is 1.

db/crud.go
```diff
@@ -1146,21 +1150,42 @@ func (db *DatabaseCollectionWithUser) Put(ctx context.Context, docid string, bod
...
+				generation++
+			} else if doc.hasFlag(channels.Conflict | channels.Hidden) {
+				// Can't use CV as an OCC Value when a document is in conflict, or we're updating the non-winning leaf
+				// There's no way to get from a given old CV to a RevTreeID to perform the update correctly, since we don't maintain linear history for a given SourceID.
+				// Reject the request and force the user to resolve the conflict using RevTree IDs which does have linear history available.
+				conflictErr = base.HTTPErrorf(http.StatusBadRequest, "Cannot use CV to modify a document in conflict - resolve first with RevTree ID")
```
torcolvin, line 1166: Is there any way that this is going to get hit for a blip code pathway? If it does, will the blip code know how to handle a 400 with this error message?

db/document.go
```diff
@@ -1484,3 +1485,60 @@ func (doc *Document) ExtractDocVersion() DocVersion {
...
+}
+
+// RevIDGeneration returns the Rev ID generation for the current version
+func (v *DocVersion) RevIDGeneration() int64 {
+	if v == nil {
+		return 0
+	}
+	gen, err := strconv.ParseInt(strings.Split(v.RevTreeID, "-")[0], 10, 64)
+	if err != nil {
+		base.AssertfCtx(context.TODO(), "Error parsing generation from rev ID %q: %v", v.RevTreeID, err)
```
torcolvin: Now that this isn't just test code, have this take a ctx and return an error.

db/document.go
```diff
@@ -1484,3 +1485,60 @@ func (doc *Document) ExtractDocVersion() DocVersion {
...
+
+// DocVersionRevTreeEqual compares two DocVersions based on their RevTreeID.
+func (v DocVersion) DocVersionRevTreeEqual(o DocVersion) bool {
+	if v.RevTreeID != o.RevTreeID {
+		return false
+	}
+	return true
+}
+
+// GetRev returns the revision ID for the DocVersion.
```
torcolvin: Suggested added text:

This could return an empty string if not set.

db/document.go
```diff
@@ -1484,3 +1485,60 @@ func (doc *Document) ExtractDocVersion() DocVersion {
...
+	}
+	return gen
+}
+
+// RevIDDigest returns the Rev ID digest for the current version
+func (v *DocVersion) RevIDDigest() string {
+	if v == nil {
+		return ""
+	}
+	return strings.Split(v.RevTreeID, "-")[1]
```
torcolvin: Make sure this code can't panic.

docs/api/paths/public/keyspace-docid.yaml
```diff
@@ -40,7 +40,7 @@ get:
                 description: The revision ID of the document.
                 type: string
               _cv:
-                description: The current version of version vector of the document.
+                description: The Current Version of the document.
```
torcolvin, line 43: Is it worth dropping an example string? Not that people should know what to do with it.

rest/doc_api.go
```diff
@@ -331,6 +331,58 @@ func (h *handler) handleGetAttachment() error {
...
+func (h *handler) getOCCValue(optionalBody db.Body) (occValue string, occValueType occVersionType, err error) {
+	// skipBodyMatchValidation can skip the validation of the occValue inside the body, since in some cases we're pulling that value out of the body anyway.
+	var skipBodyMatchValidation bool
+
+	// occValue is the optimistic concurrency control value, which can be either the current rev ID or CV - used to prevent lost updates.
+	// we grab occValue from either query param, Etag header, or request body (in that order)
+	if revQuery := h.getQuery("rev"); revQuery != "" {
+		occValue = revQuery
+		occValueType = guessOCCVersionTypeFromValue(occValue)
+	} else if ifMatch, err := h.getEtag("If-Match"); err != nil {
```
torcolvin, line 354: Should etag check be first?

rest/doc_api.go
```diff
@@ -331,6 +331,58 @@ func (h *handler) handleGetAttachment() error {
 
 }
 
+// getOCCValue retrieves the optimistic concurrency control value for a document (Revision ID or CV) from several possible sources:
+// - Query parameter "rev"
+// - If-Match header
+// - Body field "_rev" or "_cv" (if the body is provided in the request)
+// It also validates that the provided OCC value exactly matches the corresponding body field if it exists in multiple places.
+func (h *handler) getOCCValue(optionalBody db.Body) (occValue string, occValueType occVersionType, err error) {
```
torcolvin, line 339: I am not 100% sure if this is intentional or an accident. `occVersionType` is an "enum" and defined as `VersionTypeRevTreeID occVersionType = iota // Revision Tree ID (RevTreeID / RevID)`

This means if none of the if/else blocks get hit it is detected as `VersionTypeRevTreeID`.

This will get blocked if it is validated since `_rev` and `_cv` will be empty, but if it is not validated, it will get returned as rev.

If this is intentional, I'd probably call it out in this function or use iota+1 and explicitly set this in an else block.

The combination of named return default values and the lack of enum types makes this hazardous.

rest/utilities_testing.go
```diff
@@ -1138,6 +1138,16 @@ func (rt *RestTester) GetDocumentSequence(key string) (sequence uint64) {
 	return rawResponse.Xattrs.Sync.Sequence
 }
 
+// GetRawDoc returns the raw document response for a given document ID using the _raw endpoint.
+func (rt *RestTester) GetRawDoc(key string) RawDocResponse {
+	response := rt.SendAdminRequest("GET", fmt.Sprintf("/{{.keyspace}}/_raw/%s", key), "")
+	require.Equal(rt.TB(), http.StatusOK, response.Code, "Error getting raw document %s", response.Body.String())
+	response.DumpBody()
```
torcolvin:
```suggestion
```

I'd be inclined to drop the response in the general case, but I don't mind too much

rest/utilities_testing_resttester.go
```diff
@@ -106,20 +107,42 @@ func (rt *RestTester) UpdateDocRev(docID, revID string, body string) string {
 
 // UpdateDoc updates a document at a specific version and returns the new version.
```
torcolvin:
```suggestion
// UpdateDoc updates a document at a specific version and returns the new version. Uses CV for REST API if present in DocVersion, otherwise fall back to RevTreeID.
```

rest/utilities_testing_resttester.go
```diff
@@ -106,20 +107,42 @@ func (rt *RestTester) UpdateDocRev(docID, revID string, body string) string {
...
+	if isRespUseRevTreeIDInstead(resp) {
+		// trying to update a document in-conflict with a CV - try again with RevTreeID
+		// this is a pretty narrow edge-case and one that customers would deal with in the same way (get the document out of conflict using RevTreeID before using CV)
+		resp = rt.SendAdminRequest(http.MethodPut, fmt.Sprintf("/%s/%s?rev=%s", rt.GetSingleKeyspace(), docID, version.RevTreeID), body)
+	}
+	RequireStatus(rt.TB(), resp, http.StatusCreated)
+	return DocVersionFromPutResponse(rt.TB(), resp)
 }
 
 // DeleteDoc deletes a document at a specific version. The test will fail if the revision does not exist.
```
torcolvin:
```suggestion
// DeleteDoc deletes a document at a specific version. The test will fail if the revision does not exist. Uses CV for REST API if present in DocVersion, otherwise fall back to RevTreeID.
```

torcolvin (commented):
LGTM in general.

db/crud.go
bbrks: There's warnings inside `ParseRevID` already - given matchRev is coming from sync data, I don't think this is something that needs additional error handling (the 2nd return param is not an error, it's the rev digest)

db/crud.go
bbrks: Nope, it can't happen.

BLIP goes through `PutExistingCurrentVersion` since the client is the one generating a new CV value and pushing it, not relying on SG to generate the HLV entry. Put is SG/REST API-only writes.

db/document.go
bbrks: Now I look closer, this method doesn't seem nessesary to have at all. There are only 2 test uasges, and `ParseRevID` does the job of this in a more generic way.

docs/api/paths/public/keyspace-docid.yaml
bbrks: There are already examples a few lines below for the full schema which does the same thing in the Reodcly output.

rest/doc_api.go
bbrks: This is the existing order of precedence that I don't want to change.

torcolvin (dismissed):

torcolvin (dismissed):

torcolvin (approved):

PR #7696, opened 2025-08-21
CBG-4778: LWW conflict resolution for ISGR

gregns1:
CBG-4778

- Adds HLV conflict resolution to active replicator. We need two defined given when we build active replicator we don't know what sub-protocol we're running in but also we need reference to pre 4.0 conflict resolution for pre upgraded docs. 
- Adds lower level test around local wins and remote wins 
- Adds higher level tests for conflict resolution around local wins and remote wins cases
- Needed to have the ability to mock the HLV on a doc to look a certain way for test cases around MV so added test helper `AlterHLVForTest` 
- PR will not reconcile rev tree's at this time, later ticket for this
- PR takes into account work for [CBG-4789](https://jira.issues.couchbase.com/browse/CBG-4789) 

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/53/

[CBG-4789]: https://couchbasecloud.atlassian.net/browse/CBG-4789?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -1301,7 +1301,6 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 		revTreeConflictChecked := false
 		var parent string
 		var currentRevIndex int
-		var conflictResolved bool
 
 		// Conflict check here
 		// if doc has no HLV defined this is a new doc we haven't seen before, skip conflict check
@@ -1366,7 +1365,7 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 					// temp remove history alignment here, pending CBG-4791
 					alignRevTrees = false
 					revTreeConflictChecked = true
-					previousRevTreeID = doc.CurrentRev
+					previousRevTreeID = doc.GetRevTreeID()
 					newGeneration = prevGeneration + 1
 					base.InfofCtx(ctx, base.KeyCRUD, "conflict detected between the two HLV's for doc %s, incoming version %v, local version %v", base.UD(doc.ID), newDocHLV.ExtractCurrentVersionFromHLV(), doc.HLV.ExtractCurrentVersionFromHLV())
 					if conflictResolver == nil {
@@ -1375,23 +1374,14 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 					}
 					// resolve conflict
 					var newHLV *HybridLogicalVector
-					newHLV, err = db.ResolveHLVConflict(ctx, doc, newDoc, conflictResolver)
+					newHLV, err = db.resolveHLVConflict(ctx, doc, newDoc, conflictResolver)
 					if err != nil {
 						base.InfofCtx(ctx, base.KeyCRUD, "Failed to resolve HLV conflict for doc %s, error: %v", base.UD(doc.ID), err)
 						return nil, nil, false, nil, err
 					}
-					conflictResolved = true // mark that we resolved the conflict so we don't overwrite the HLV MV below
 					// overwrite the existing HLV with the new one
 					doc.HLV = newHLV
 				}
-			}
-		}
-		// we only want to populate the merge versions if we have not already resolved the conflict above given the
-		// special handling for conflict cases around merge versions.
-		if !conflictResolved {
-			// populate merge versions
-			if newDocHLV.MergeVersions != nil {
-				doc.HLV.MergeVersions = newDocHLV.MergeVersions
 			}
 		}
 		// rev tree conflict check if we have rev tree history to check against + finds current rev index to allow us
@@ -1710,21 +1700,16 @@ func (db *DatabaseCollectionWithUser) revTreeConflictCheck(ctx context.Context, 
 	return parent, currentRevIndex, nil
 }
 
-func (db *DatabaseCollectionWithUser) ResolveHLVConflict(ctx context.Context, localDoc, incomingDoc *Document, resolver *ConflictResolver) (*HybridLogicalVector, error) {
-	if resolver == nil {
-		return nil, errors.New("Conflict resolution function is nil for resolveConflict")
-	}
-
-	localRevID := localDoc.SyncData.CurrentRev
+func buildResolverBody(localDoc, incomingDoc *Document) (Body, Body, error) {
+	localRevID := localDoc.SyncData.GetRevTreeID()
 	localAttachments := localDoc.Attachments()
 	localExpiry := localDoc.SyncData.Expiry
 
-	remoteRevID := incomingDoc.RevID
 	remoteAttachments := incomingDoc.Attachments()
 
 	localDocBody, err := localDoc.GetDeepMutableBody()
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	localDocBody[BodyRev] = localRevID
 	localDocBody[BodyId] = localDoc.ID
@@ -1735,13 +1720,27 @@ func (db *DatabaseCollectionWithUser) ResolveHLVConflict(ctx context.Context, lo
 
 	remoteDocBody, err := incomingDoc.GetDeepMutableBody()
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
-	remoteDocBody[BodyRev] = remoteRevID
+	remoteDocBody[BodyRev] = incomingDoc.SyncData.GetRevTreeID()
 	remoteDocBody[BodyId] = incomingDoc.ID
 	remoteDocBody[BodyAttachments] = remoteAttachments
 	remoteDocBody[BodyDeleted] = incomingDoc.Deleted
 	remoteDocBody[BodyCV] = incomingDoc.HLV.GetCurrentVersionString()
+
+	return localDocBody, remoteDocBody, nil
+}
+
+func (db *DatabaseCollectionWithUser) resolveHLVConflict(ctx context.Context, localDoc, incomingDoc *Document, resolver *ConflictResolver) (*HybridLogicalVector, error) {
+	if resolver == nil {
+		return nil, errors.New("Conflict resolution function is nil for resolveConflict")
+	}
+
+	localDocBody, remoteDocBody, err := buildResolverBody(localDoc, incomingDoc)
+	if err != nil {
+		base.InfofCtx(ctx, base.KeyReplicate, "Error when building local and remote documents for conflict resolution for doc %s: %v", base.UD(localDoc.ID), err)
+		return nil, err
+	}
 
 	conflict := Conflict{
 		LocalDocument:  localDocBody,
@@ -1782,37 +1781,11 @@ func (db *DatabaseCollectionWithUser) resolveConflict(ctx context.Context, local
 		return "", nil, errors.New("Conflict resolution function is nil for resolveConflict")
 	}
 
-	// Local doc (localDoc) is persisted in the bucket unlike the incoming remote doc (remoteDoc).
-	// Internal properties of the localDoc can be accessed from sync metadata.
-	localRevID := localDoc.SyncData.GetRevTreeID()
-	localAttachments := localDoc.Attachments()
-	localExpiry := localDoc.SyncData.Expiry
-
-	remoteRevID := remoteDoc.RevID
-	remoteAttachments := remoteDoc.Attachments()
-
-	// TODO: Make doc expiry (_exp) available over replication.
-	// remoteExpiry := remoteDoc.Expiry
-
-	localDocBody, err := localDoc.GetDeepMutableBody()
+	localDocBody, remoteDocBody, err := buildResolverBody(localDoc, remoteDoc)
 	if err != nil {
+		base.InfofCtx(ctx, base.KeyReplicate, "Error when building local and remote documents for conflict resolution for doc %s: %v", base.UD(localDoc.ID), err)
 		return "", nil, err
 	}
-	localDocBody[BodyId] = localDoc.ID
-	localDocBody[BodyRev] = localRevID
-	localDocBody[BodyAttachments] = localAttachments
-	localDocBody[BodyExpiry] = localExpiry
-	localDocBody[BodyDeleted] = localDoc.IsDeleted()
-
-	remoteDocBody, err := remoteDoc.GetDeepMutableBody()
-	if err != nil {
-		return "", nil, err
-	}
-	remoteDocBody[BodyId] = remoteDoc.ID
-	remoteDocBody[BodyRev] = remoteRevID
-	remoteDocBody[BodyAttachments] = remoteAttachments
-	// remoteDocBody[BodyExpiry] = remoteExpiry
-	remoteDocBody[BodyDeleted] = remoteDoc.Deleted
 
 	conflict := Conflict{
 		LocalDocument:  localDocBody,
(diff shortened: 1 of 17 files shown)


db/crud.go
```diff
@@ -1692,7 +1777,7 @@ func (db *DatabaseCollectionWithUser) resolveConflict(ctx context.Context, local
 	}
 
 	// Local doc (localDoc) is persisted in the bucket unlike the incoming remote doc (remoteDoc).
-	// Internal properties of the localDoc can be accessed from syc metadata.
+	// Internal properties of the localDoc can be accessed from sync metadata.
```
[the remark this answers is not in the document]
gregns1: This is fixed?

torcolvin:
I plan to review this more, but I've created some helper functions https://github.com/couchbase/sync_gateway/pull/7704 to deal with updating HLV. These duplicate/obviate some of the code here. I needed to create these standalone functions to write unit tests that couchbase lite can use to ensure matching behavior.

Those functions will be used for blip client testing and for ISGR.

db/crud.go
```diff
@@ -1682,6 +1704,69 @@ func (db *DatabaseCollectionWithUser) revTreeConflictCheck(ctx context.Context,
 	return parent, currentRevIndex, nil
 }
 
+func (db *DatabaseCollectionWithUser) ResolveHLVConflict(ctx context.Context, localDoc, incomingDoc *Document, resolver *ConflictResolver) (*HybridLogicalVector, error) {
```
torcolvin: What is the benefit to having this function vs `DatabaseCollectionWithUser.resolveConflict` function? If I understand correctly this function is basically a duplicate except that it will update the HLV.

db/crud.go
```diff
@@ -1901,6 +1986,27 @@ func (db *DatabaseCollectionWithUser) resolveDocMerge(ctx context.Context, local
 	return mergedRevID, docHistory, nil
 }
 
+// resolveLocalWinsHLV will update remote doc's body and attachments to match the local doc, and return a new HLV for local wins
+func (db *DatabaseCollectionWithUser) resolveLocalWinsHLV(ctx context.Context, localDoc, remoteDoc *Document) (*HybridLogicalVector, error) {
```
torcolvin, line 1969: There seem to be a few differences between `resolveLocalWinsHLV` and `resolveDocLocalWins`, some of which seem right, and some of which seem not right.

1. There's not a new revtree id that is created in this function. You would need a local revtree id to support 3.x CBL replication even after 4.x ISGR replication.
2. There's handling of branching tombstones. I'm not sure there will ever be branching since the local revtree is already overwritten by remote revtree by the time we are here?  but I'm actually not sure.
3. tombstoneActiveRevision means that the old revision is added to the doc.History for revtree ids.

db/crud.go
```diff
@@ -1356,15 +1357,36 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
...
-		if newDocHLV.MergeVersions != nil {
-			doc.HLV.MergeVersions = newDocHLV.MergeVersions
+		// we only want to populate the merge versions if we have not already resolved the conflict above given the
+		// special handling for conflict cases around merge versions.
+		if !conflictResolved {
+			// populate merge versions
+			if newDocHLV.MergeVersions != nil {
+				doc.HLV.MergeVersions = newDocHLV.MergeVersions
+			}
 		}
```
torcolvin: I'm reasonably sure that this logic is just completely wrong, and I am going to move this into its own function as part of https://github.com/couchbase/sync_gateway/pull/7704. This problem pre-exists this PR.

I want to move the updating the HLV into its own function that can be used by blip tester and crud API.

If this code is ever called when a merge version is created, this is wrong, as you've noted.

db/hybrid_logical_vector.go
```diff
@@ -739,3 +739,126 @@ func IsInConflict(ctx context.Context, localHLV, incomingHLV *HybridLogicalVecto
...
+		// remote document wins
+		return conflict.RemoteDocument, nil
+	}
+	return conflict.LocalDocument, nil
+}
+
+// localWinsConflictResolutionForHLV will alter the HLV for a local wins conflict resolution. Preserving local MV
+// unless incoming MV has a src common with local MV and has a higher version, in which case local MV is invalidated and moved to PV.
+// In the eventuality that local CV is <= to incoming CV, a new CV will be created with this db's sourceID
+func localWinsConflictResolutionForHLV(ctx context.Context, localHLV, incomingHLV *HybridLogicalVector, docID, sourceID string) (*HybridLogicalVector, error) {
```
torcolvin, line 843: I didn't read `localWinsConflictResolutionForHLV` or `remoteWinsConflictResolutionForHLV` when doing https://github.com/couchbase/sync_gateway/pull/7709 but I think they will obviate these functions.

db/hybrid_logical_vector.go
```diff
@@ -739,3 +739,126 @@ func IsInConflict(ctx context.Context, localHLV, incomingHLV *HybridLogicalVecto
...
+// remoteWinsConflictResolutionForHLV will alter the HLV for a remote wins conflict resolution. Preserving incoming MV
+// unless local MV has a src common with incoming MV and has a higher version, in which case incoming MV is invalidated and moved to PV.
+func remoteWinsConflictResolutionForHLV(ctx context.Context, docID string, localHLV, incomingHLV *HybridLogicalVector) (*HybridLogicalVector, error) {
+	if localHLV == nil || incomingHLV == nil {
+		return nil, errors.New("local or incoming hlv is nil for resolveConflict")
+	}
+
+	newHLV := incomingHLV.Copy()
+
+	newHLV.addNewerVersionTmp(localHLV)
```
torcolvin: I think that the case where:

> Preserving incoming MV unless local MV has a src common with incoming MV and has a higher version, in which case incoming MV is invalidated and moved to PV.

should not be possible as per https://github.com/couchbase/sync_gateway/pull/7687#discussion_r2289443416

Let me know if there's a case you are thinking of that doesn't meet this classification.

db/sg_replicate_cfg.go
```diff
@@ -609,17 +609,30 @@ func (m *sgReplicateManager) NewActiveReplicatorConfig(config *ReplicationCfg) (
...
 	rc.Direction = config.Direction
 
-	// Set conflict resolver for pull replications
+	// Set conflict resolver for pull replications. Currently, at this point we don't know if the replication will
+	// be sub-protocol version 4 or < 4 so we must set both conflict resolver functions. In addition, we must have
+	// both available for the possibility of replicating a document that exists both side fo the replication already
+	// but both are legacy pre upgrades docs without HLV.
 	if rc.Direction == ActiveReplicatorTypePull || rc.Direction == ActiveReplicatorTypePushAndPull {
 		if config.ConflictResolutionType == "" {
 			rc.ConflictResolverFunc, err = NewConflictResolverFunc(m.loggingCtx, ConflictResolverDefault, "", m.dbContext.Options.JavascriptTimeout)
```
torcolvin: nit: I'd add `var err error` to block scope err to this function, to avoid leaking err variable, now that the error is handled inside this function.

db/sg_replicate_cfg.go
```diff
@@ -609,17 +609,30 @@ func (m *sgReplicateManager) NewActiveReplicatorConfig(config *ReplicationCfg) (
...
-
+			if err != nil {
+				return nil, err
+			}
+			rc.ConflictResolverFuncForHLV, err = NewConflictResolverFuncForHLV(m.loggingCtx, ConflictResolverDefault, "", m.dbContext.Options.JavascriptTimeout)
+			if err != nil {
+				return nil, err
+			}
 		} else {
 			rc.ConflictResolverFunc, err = NewConflictResolverFunc(m.loggingCtx, config.ConflictResolutionType, config.ConflictResolutionFn, m.dbContext.Options.JavascriptTimeout)
```
torcolvin: nit: I'd add `var err error` to block scope err to this function, to avoid leaking err variable, now that the error is handled inside this function.

db/sg_replicate_conflict_resolver.go
```diff
@@ -51,10 +51,13 @@ const (
...
 // Conflict is the input to all conflict resolvers.  LocalDocument and RemoteDocument
 // are expected to be document bodies with metadata injected into the body following
-// the same approach used for doc and oldDoc in the Sync Function
+// the same approach used for doc and oldDoc in the Sync Function. LocalHLV and RemoteHLV
+// will be the documents HLVs, these are needed for HLV ready conflict resolution.
 type Conflict struct {
 	LocalDocument  Body `json:"LocalDocument"`
 	RemoteDocument Body `json:"RemoteDocument"`
+	LocalHLV       *HybridLogicalVector
+	RemoteHLV      *HybridLogicalVector
```
torcolvin, line 60: Are there going to be cases where only one HLV is present?

db/sg_replicate_conflict_resolver.go
```diff
@@ -51,10 +51,13 @@ const (
 
 // Conflict is the input to all conflict resolvers.  LocalDocument and RemoteDocument
 // are expected to be document bodies with metadata injected into the body following
-// the same approach used for doc and oldDoc in the Sync Function
+// the same approach used for doc and oldDoc in the Sync Function. LocalHLV and RemoteHLV
+// will be the documents HLVs, these are needed for HLV ready conflict resolution.
```
torcolvin:
```suggestion
// will be the documents HLVs.
```

Document if these can be nil. In the case of legacy rev with v4 protocol, will this get run? What about incoming legacy rev with existing hlv?

db/sg_replicate_conflict_resolver.go
```diff
@@ -137,6 +140,38 @@ func (c *ConflictResolver) Resolve(ctx context.Context, conflict Conflict) (winn
...
+	winner, err = c.crf(ctx, conflict)
+	if err != nil {
+		return winner, "", err
+	}
+
+	winningRev, ok := winner[BodyCV]
+	if !ok {
+		c.stats.ConflictResultMergeCount.Add(1)
+		return winner, ConflictResolutionMerge, nil
+	}
```
torcolvin, line 166: What are the cases where this gets hit vs the end of this function

db/sg_replicate_conflict_resolver.go
```diff
@@ -137,6 +140,38 @@ func (c *ConflictResolver) Resolve(ctx context.Context, conflict Conflict) (winn
...
+		return winner, ConflictResolutionLocal, nil
+	}
+
+	remoteRev, ok := conflict.RemoteDocument[BodyCV]
+	if ok && remoteRev == winningRev {
+		c.stats.ConflictResultRemoteCount.Add(1)
+		return winner, ConflictResolutionRemote, nil
+	}
+
+	base.InfofCtx(ctx, base.KeyReplicate, "Conflict resolver returned non-empty revID (%s) not matching local (%s) or remote (%s), treating result as merge.", winningRev, localRev, remoteRev)
```
torcolvin: Does this indicate user error in the custom conflict resolver, or can/should we prevent the user from writing `_cv` by the user?

It seems like it would be nice if we could prevent this caveat from the docs https://docs.couchbase.com/sync-gateway/current/sync/sync-inter-syncgateway-conflict-resolution.html#conflict-resolution-approaches

> When creating a new revision, do not provide a _rev property. A new revision ID will be generated. Use delete mergedDoc._rev to invalidate the property.

db/hybrid_logical_vector.go
gregns1: Not sure its possible but this documnet suggests that it should be handled. https://docs.google.com/document/d/19mC9-TptN7ynuF5XR9TQ_MckHWsSrjUfI2p71G_DD1U/edit?tab=t.0#heading=h.iyhds217i9i It indicates any version across the hlv that is newer than the M<V then the MV should be invalided. I think in the context of this PR this way always a temp function till you worked on better shared functions for this.

db/crud.go
gregns1: 1. New RevID is generated after this function is done inside PutExistsingCurrentVersion
2 & 3. There is work after this to reconcile revTree's on conflict resolution, the idea here is work for this will take place there

db/crud.go
gregns1: The functions have different behaviour due to the difference between hlv an revtree, but my main motivation was to keep code paths different to make easier for debugging (I sometimes feel its just easier this way then constantly having to switch on a hlv boolean) + easier to eventually one day remove rev tree handling completely. 

But I agree that I could move some of this into shared function (building up the doc body for example)

db/sg_replicate_conflict_resolver.go
gregns1: Only both HLV's will be here, if one is missing it may be given an implicit HLV or we fall back to rev tree conflict handling.

db/sg_replicate_conflict_resolver.go
gregns1: The aim here was to keep in sync with current existing conflict resolution functions.

db/sg_replicate_conflict_resolver.go
gregns1: I agree this is probably preferable, will make this change.

db/crud.go
```diff
@@ -1362,16 +1362,28 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
...
+					previousRevTreeID = doc.GetRevTreeID()
+					newGeneration = prevGeneration + 1
+					base.InfofCtx(ctx, base.KeyCRUD, "conflict detected between the two HLV's for doc %s, incoming version %v, local version %v", base.UD(doc.ID), newDocHLV.ExtractCurrentVersionFromHLV(), doc.HLV.ExtractCurrentVersionFromHLV())
+					if conflictResolver.hlvConflictResolver == nil {
+						// cancel rest of update, HLV is in conflict and no resolver is present
+						return nil, nil, false, nil, base.HTTPErrorf(http.StatusConflict, "Document revision conflict")
+					}
+					// resolve conflict
+					var newHLV *HybridLogicalVector
+					newHLV, err = db.resolveHLVConflict(ctx, doc, newDoc, conflictResolver.hlvConflictResolver)
```
torcolvin, line 1377: nit

```suggestion
					newHLV, err := db.resolveHLVConflict(ctx, doc, newDoc, conflictResolver.hlvConflictResolver)
```

db/crud.go
```diff
@@ -1907,6 +1965,27 @@ func (db *DatabaseCollectionWithUser) resolveDocMerge(ctx context.Context, local
...
+	}
+
+	newHLV, err := localWinsConflictResolutionForHLV(ctx, localDoc.HLV, remoteDoc.HLV, localDoc.ID, db.dbCtx.EncodedSourceID)
+	if err != nil {
+		return nil, err
+	}
+
+	remoteDoc.RemoveBody()
+	remoteDoc.Deleted = localDoc.IsDeleted()
+	remoteDoc.SetAttachments(localDoc.Attachments().ShallowCopy())
```
torcolvin, line 1983: I think this is fine for now, but is it possible that resolving an HLV conflict will also mean that affects a revtree as well?

I suspect it will need to call https://github.com/couchbase/sync_gateway/blob/02429a56456ff283c6cdc7c6d4e155212e70eefc/db/crud.go#L1824-L1862 as well.

torcolvin (approved):
LGTM, I think we can sort out the other stuff in a future PR.

db/crud.go
gregns1: This will be taken care of in CBG-4791

PR #7704, opened 2025-08-25
CBG-4829 add test cases for hlv conflicts

torcolvin:
CBG-4829 add test cases for hlv conflicts

I am torn on whether `IsInConflict` should return three states (conflict, no conflict, no conflict but shouldn't update). It isn't super expensive to check twice and makes the logic easier, but on the other hand these operations always happen at the same time.

Fix bug where incoming mv on an hlv conflict merge needs to move to final pv.

## Pre-review checklist


db/hybrid_logical_vector.go
```diff
@@ -728,14 +785,43 @@ func IsInConflict(ctx context.Context, localHLV, incomingHLV *HybridLogicalVecto
...
 	// - If merge versions do not match, then it has not been resolved.
 	if len(incomingHLV.MergeVersions) != 0 && len(localHLV.MergeVersions) != 0 && maps.Equal(incomingHLV.MergeVersions, localHLV.MergeVersions) {
-		base.DebugfCtx(ctx, base.KeyVV, "merge versions match between local HLV %#v and incoming HLV %#v, conflict previously resolved", localHLV, incomingCV)
-		return false, nil
+		return HLVNoConflict
 	}
-	return true, nil
+	return HLVConflict
+}
+
```
adamcfraser, line 754: Looking at these the two Update functions below, it seems to me like what we're really missing is a function like
```
// Update the PV of hlv with any entries from incomingHLV that are newer than hlv's entry for the same source
func (hlv *HybridLogicalVector) UpdateHistory(incomingHLV *HybridLogicalVector) error {
  // For any sources in incoming HLV's CV, MV, PV:
  //   - if the source is not present in hlv, add it to hlv.PV
  //   - if the source is present in hlv.PV but incomingHLV's value is newer, update hlv.PV
  //   - if the source is present in hlv.CV or hlv.MV, ignore it (don't overwrite CV or MV when updating history)
  }
  ```
  This would simplify the functions below, I think, to:
  UpdateFromIncomingRemoteWins:
      incomingHLV.UpdateHistory(hlv)
      
   UpdateFromIncomingWithNewCV:
      hlv.AddVersion(newCV)
      hlv.UpdateHistory(incomingHLV)
  
  I think it's also quite a bit easier to reason about - what we're really doing in both cases is "updating the history of the winning revision with any newer entries from the losing revision".  You could call it MergeHistory instead of UpdateHistory if you wanted to.

File: db/hybrid_logical_vector.go [hunk 2]
```diff
@@ -728,14 +785,43 @@ func IsInConflict(ctx context.Context, localHLV, incomingHLV *HybridLogicalVecto
...
+	// - remove the any pv that contain the merge sourceIDs
+	// - add the merge sourceIDs with the incoming version and local version
+	// - update the new CV
+	delete(hlv.PreviousVersions, incomingHLV.SourceID)
+	delete(hlv.PreviousVersions, newCV.SourceID)
+	delete(hlv.PreviousVersions, previousSourceID)
+	hlv.SetMergeVersion(incomingHLV.SourceID, incomingHLV.Version)
+	hlv.SetMergeVersion(previousSourceID, previousVersion)
+	hlv.SourceID = newCV.SourceID
+	hlv.Version = newCV.Value
```
torcolvin: I expect this value will be `expandMacroCASValueString` in cases where this isn't blip.

I am not sure if we need to update cvcas here or not, quite possibly yes?

gregns1: https://docs.google.com/document/d/19mC9-TptN7ynuF5XR9TQ_MckHWsSrjUfI2p71G_DD1U/edit?tab=t.0#heading=h.qb4ommtyrgfp 

Case two here suggests a new CV needs generating when local CV ver <= incoming CV ver. But otherwise we don't need to generate a new CV. In the case we don't need to generate a new CV then merge versions don't get set. It feels like this function should handle this?

torcolvin: It is the responsibility of the caller to pay attention to `HLVNoConflictRevAlreadyPresent` from `IsInConflict` and not add the cv.

torcolvin: I'm not clear when this would happen.

Is this when there is a conflict, and there is custom conflict resolution? In normal case where  `local CV ver <= incoming CV ver` happens then the it would invoke the remote wins pathway?

Is the concern that this function should be renamed to just merge or is there a functional problem?

torcolvin: This is true, but as discussed this offline, we aren't going to implement this right now because we aren't sure CBL has enough time to pick up the change and it will also need changes in the topology tests. I will file a post 4.0.0 ticket for this enhancement.

adamcfraser (approved):

PR #7706, opened 2025-08-25
CBG-4787 allow updating document that have been tombstoned

torcolvin:
CBG-4787 allow updating document that have been tombstoned

1. Calling `Purge` on a document leaves behind a _vv that refers to a _past_ mutation of a document. If that revision would conflict with CBL, CBL would never be able to pull that tombstone as in the test `TestBlipPushRevOnResurrection`. The fix for this is in `HasConflict` which is to disable conflict checking while there is a _vv but no _hlv, to match rev tree id resurrection.
2. The fix for import code running and canceling when trying to import a deletion is in https://github.com/couchbase/sync_gateway/blob/cde6f6789baf97833e56961ca761fdfff6bc8457/db/import_listener.go#L201 because it checks `syncData == nil` which was true in 3.3 and not in 4.0. I propose an alternate way of resolving this in https://github.com/couchbase/sync_gateway/pull/7710.

- Create SyncData.IsEmpty() to allow checking for no _sync when _vv exists
- Change calls for SyncData == nil to use SyncData.IsEmpty()
- Remove a bit of dead code

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/49/

A different case could be be broken, and addressing it would be best done in a different PR:

1. doc1 is converged on CBS1 and CBS2, CBL1 with `1@abc`
2. delete doc1 on CBS1 to create `2@cbs1;1@abc`
3. delete doc1 on CBS2 to create `3@cbs2;1@abc`
4. CBL1 pull revision from CBS2 `2@cbs1;1@abc`
 5. Turn on XDCR
 6. Both CBS1 and CBS2 will have `3@cbs2;1@abc`. However, CBL1 will remain at `2@cbs1`.
 7. TODO: validate that CBL1 which isn't source ID of `abc` will be able to push a resurrection to CBS1. I'm not 100% sure this is possible, but I actually don't think this PR.

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -72,65 +72,38 @@ func (c *DatabaseCollection) GetDocumentWithRaw(ctx context.Context, docid strin
 	if key == "" {
 		return nil, nil, base.HTTPErrorf(400, "Invalid doc ID")
 	}
-	if c.UseXattrs() {
-		doc, rawBucketDoc, err = c.GetDocWithXattrs(ctx, key, unmarshalLevel)
-		if err != nil {
-			return nil, nil, err
-		}
-		isSgWrite, crc32Match, _ := doc.IsSGWrite(ctx, rawBucketDoc.Body)
-		if crc32Match {
-			c.dbStats().Database().Crc32MatchCount.Add(1)
-		}
-
-		// If existing doc wasn't an SG Write, import the doc.
-		if !isSgWrite {
-			// reload to get revseqno for on-demand import
-			doc, rawBucketDoc, err = c.getDocWithXattrs(ctx, key, append(c.syncGlobalSyncAndUserXattrKeys(), base.VirtualXattrRevSeqNo), unmarshalLevel)
-			if err != nil {
-				return nil, nil, err
-			}
-			isSgWrite, _, _ := doc.IsSGWrite(ctx, rawBucketDoc.Body)
-			if !isSgWrite {
-				var importErr error
-				doc, importErr = c.OnDemandImportForGet(ctx, docid, doc, rawBucketDoc.Body, rawBucketDoc.Xattrs, rawBucketDoc.Cas)
-				if importErr != nil {
-					return nil, nil, importErr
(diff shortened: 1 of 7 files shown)


db/crud.go
```diff
@@ -72,65 +72,38 @@ func (c *DatabaseCollection) GetDocumentWithRaw(ctx context.Context, docid strin
...
 			return nil, nil, err
 		}
-		isSgWrite, crc32Match, _ := doc.IsSGWrite(ctx, rawBucketDoc.Body)
-		if crc32Match {
-			c.dbStats().Database().Crc32MatchCount.Add(1)
-		}
-
-		// If existing doc wasn't an SG Write, import the doc.
+		isSgWrite, _, _ := doc.IsSGWrite(ctx, rawBucketDoc.Body)
 		if !isSgWrite {
```
gregns1: Why do we need to call `IsSGWrite` again here?

torcolvin: I didn't actually modify this code, just dedent it, but I agree this is weird so I fixed it.

db/document.go
```diff
@@ -121,6 +121,12 @@ type SyncData struct {
 	removedRevisionBodyKeys map[string]string // keys of non-winning revisions that have been removed (and so may require deletion), indexed by revID
 }
 
+// IsEmpty returns true if the SyncData structure is empty. An empty SyncData structure can have a non
+// HLV attribute, representing the unmarshaling of a document with a _vv xattr but no _sync xattr.
```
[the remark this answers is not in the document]
torcolvin: I don't think this is helpful.

gregns1 (approved):

torcolvin:
Implementing via https://github.com/couchbase/sync_gateway/pull/7710/files instead.

torcolvin:
Closing this in favor of https://github.com/couchbase/sync_gateway/pull/7710

PR #7710, opened 2025-08-26
CBG-4787 move SyncData.HLV to Document.HLV

torcolvin:
Move SyncData.HLV to Document.HLV

This is a bigger refactor which also fixes https://github.com/couchbase/sync_gateway/pull/7706, by making `SyncData` actually allowed to be nil, which was possible in 3.3 but not in 4.0 and causes CBG-4787 and other cases where SyncData == nil was checked before.

There is risk to this type of refactor, but I did changes in https://github.com/couchbase/sync_gateway/pull/7706 to check for `SyncData.IsEmpty()` in places that it wasn't a problem that it didn't get flagged by tests before.

I think even if this isn't considered for 4.0.0 there are benefits to doing this in the future:

- drop `SyncDataJSON` data structure and simplify marshal
- remove the confusing behavior about what `CurrentRev` means, which baffles me a bit. I renamed this to `RevTreeID` to indicate what it means. I often found `CurrentRev` confusing, but perhaps this is in contrast to `NewestRev`, which is a probably outside the scope.
- Clarity when `SyncData` is nil from import feed or GET and hlv is populated.

Downsides:

- If HLV exists on Document, the CV value in `SyncData` has to be kept up to date for use in `channels`, `_rev.src`, `_rev.ver`

I actually feel fine about doing this for 4.0.0 with the likelihood it might find new bugs before the release and there's pretty good test coverage for most of the circumstances we care about, but I am OK rejecting or deferring this. I wanted to showcase this as an approach I did. I felt like https://github.com/couchbase/sync_gateway/pull/7647 ended up paying off pretty quickly in both confusion and finding bugs.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/52/


db/document.go
```diff
@@ -119,6 +117,19 @@ type SyncData struct {
 
 	addedRevisionBodies     []string          // revIDs of non-winning revision bodies that have been added (and so require persistence)
 	removedRevisionBodyKeys map[string]string // keys of non-winning revisions that have been removed (and so may require deletion), indexed by revID
+
+	RevAndVersion channels.RevAndVersion `json:"rev"` // Current revision and HLV version
```
adamcfraser: Let's move this up to the top of SyncData, for visibility.

db/document.go
```diff
@@ -510,73 +527,64 @@ func UnmarshalDocumentSyncData(data []byte, needHistory bool) (*SyncData, error)
...
-// Unmarshals sync metadata for a document arriving via DCP.  Includes handling for xattr content
+// UnmarshalDocumentSyncDataFromFeed sync metadata for a document arriving via DCP.  Includes handling for xattr content
 // being included in data.  If not present in either xattr or document body, returns nil but no error.
 // Returns the raw body, in case it's needed for import.
 
 // TODO: Using a pool of unmarshal workers may help prevent memory spikes under load
-func UnmarshalDocumentSyncDataFromFeed(data []byte, dataType uint8, userXattrKey string, needHistory bool) (result *SyncData, rawBody []byte, rawXattrs map[string][]byte, err error) {
-	var body []byte
+func UnmarshalDocumentSyncDataFromFeed(data []byte, dataType uint8, userXattrKey string, needHistory bool) (*RawDocument, error) {
+	rawDoc := &RawDocument{}
```
adamcfraser: We're including globalSync in the fetch below (line 541), but it's not returned with RawDocument.  I can't tell if this is harmless (we don't need to reference attachments during feed processing), or an error in this PR, or an error in https://github.com/couchbase/sync_gateway/pull/7647.  Conceptually it feels like globalSync should be included in RawDocument, but let me know whether you think this is an issue.

db/background_mgr_attachment_migration.go
```diff
@@ -131,14 +131,14 @@ func (a *AttachmentMigrationManager) Run(ctx context.Context, options map[string
 		}
 
 		a.docsProcessed.Add(1)
-		syncData, _, _, err := UnmarshalDocumentSyncDataFromFeed(event.Value, event.DataType, collection.userXattrKey(), false)
+		doc, err := UnmarshalDocumentSyncDataFromFeed(event.Value, event.DataType, collection.userXattrKey(), false)
```
adamcfraser: I'd name this something other than 'doc' since this naming refers to a Document everywhere else in the code (it's RawDocument here).

db/change_cache.go
```diff
@@ -377,7 +377,7 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent, docType DocumentType)
 	}
 
 	// First unmarshal the doc (just its metadata, to save time/memory):
-	syncData, rawBody, rawXattrs, err := UnmarshalDocumentSyncDataFromFeed(docJSON, event.DataType, collection.userXattrKey(), false)
+	doc, err := UnmarshalDocumentSyncDataFromFeed(docJSON, event.DataType, collection.userXattrKey(), false)
```
adamcfraser: Same comment about naming (rawDoc instead of doc).

db/document.go
```diff
@@ -510,73 +527,64 @@ func UnmarshalDocumentSyncData(data []byte, needHistory bool) (*SyncData, error)
...
 
-// Unmarshals sync metadata for a document arriving via DCP.  Includes handling for xattr content
+// UnmarshalDocumentSyncDataFromFeed sync metadata for a document arriving via DCP.  Includes handling for xattr content
 // being included in data.  If not present in either xattr or document body, returns nil but no error.
 // Returns the raw body, in case it's needed for import.
 
 // TODO: Using a pool of unmarshal workers may help prevent memory spikes under load
-func UnmarshalDocumentSyncDataFromFeed(data []byte, dataType uint8, userXattrKey string, needHistory bool) (result *SyncData, rawBody []byte, rawXattrs map[string][]byte, err error) {
-	var body []byte
+func UnmarshalDocumentSyncDataFromFeed(data []byte, dataType uint8, userXattrKey string, needHistory bool) (*RawDocument, error) {
```
adamcfraser: I think the reason we need to use RawDocument (and not just Document) is the way the consumers are using rawXattrs later.  Is that accurate?  Can we avoid the need for RawDocument by having this function return (*Document, rawXattrs)?

db/crud.go
```diff
@@ -225,43 +225,16 @@ func (db *DatabaseCollection) unmarshalDocumentWithXattrs(ctx context.Context, d
...
-						// No valid sync data in doc, but doc has been upgraded to use xattrs
-						syncData = upgradeDoc.SyncData
-					} else {
-						base.WarnfCtx(ctx, "No valid sync data nor xattrs in doc %q", base.UD(docid))
-						err = base.HTTPErrorf(404, "Not imported")
-					}
-				}
-			}
-		}
+func (db *DatabaseCollection) GetDocSyncDataNoImport(ctx context.Context, docid string, level DocumentUnmarshalLevel) (*SyncData, *HybridLogicalVector, error) {
```
adamcfraser, line 229: What's the idea behind the switch to *SyncData -  to return nil when there's no sync xattr?  I'm not sure that's happening as it's currently implemented.

db/import_listener.go
```diff
@@ -179,7 +179,7 @@ func (il *importListener) ProcessFeedEvent(event sgbucket.FeedEvent) bool {
 }
 
 func (il *importListener) ImportFeedEvent(ctx context.Context, collection *DatabaseCollectionWithUser, event sgbucket.FeedEvent) {
-	syncData, rawBody, rawXattrs, err := UnmarshalDocumentSyncDataFromFeed(event.Value, event.DataType, collection.userXattrKey(), false)
+	doc, err := UnmarshalDocumentSyncDataFromFeed(event.Value, event.DataType, collection.userXattrKey(), false)
```
adamcfraser: rawDoc instead of doc?

adamcfraser (changes requested):
Generally this looks good to me.  Some questions around RawDocument but otherwise I've been through everything and I think it's reasonable for 4.0

db/document.go
```diff
@@ -119,6 +117,19 @@ type SyncData struct {
...
+func (sd *SyncData) SetRevTreeID(revTreeID string) {
+	sd.RevAndVersion.RevTreeID = revTreeID
+}
+
+func (sd *SyncData) GetRevTreeID() string {
+	if sd == nil {
+		return ""
+	}
+	return sd.RevAndVersion.RevTreeID
 }
```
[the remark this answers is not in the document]
torcolvin: I don't think this is necessary.

db/document.go
torcolvin: Using `Document` means that we have to implement a `Document.SyncData` == nil and many of the callers require accessing raw body and xattrs. I switched to use `sgbucket.BucketDocument` and returned SyncData unmarshalled.

e.g. https://github.com/couchbase/sync_gateway/blob/8ee0ee37786b8f67cd2111116a59950a4f6b147f/db/import.go#L42

This code is a bit silly for import because we actually have to do a remarshal into bytes in one case https://github.com/couchbase/sync_gateway/blob/8ee0ee37786b8f67cd2111116a59950a4f6b147f/db/import.go#L78 but fixing this to take a `Document` was too much work for this PR.

db/raw_document.go
```diff
@@ -0,0 +1,26 @@
...
+//
+// Use of this software is governed by the Business Source License included
+// in the file licenses/BSL-Couchbase.txt.  As of the Change Date specified
+// in that file, in accordance with the Business Source License, use of this
+// software will be governed by the Apache License, Version 2.0, included in
+// the file licenses/APL2.txt.
+
+package db
+
+// RawDocument respresents a raw document with unmarshalled metadata for _sync, and _vv.
```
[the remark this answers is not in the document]
torcolvin: Removed this code.

db/document.go
torcolvin: This is messy since `importDocRaw` expects raw xattrs. I think this might be a worthwhile enhancement in the future, see https://github.com/couchbase/sync_gateway/pulls#discussion_r2301840605

db/crud.go
torcolvin: Yeah, just switch to a nil ptr rather than constructing an empty object. I changed the behavior of the function to not return a nil struct if _sync exists but _vv doesn't.

db/change_cache.go
```diff
@@ -497,27 +497,27 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent, docType DocumentType)
...
 		TimeReceived: timeReceived,
 		Channels:     syncData.Channels,
 		CollectionID: event.CollectionID,
 	}
-	if syncData.HLV != nil {
-		change.SourceID = syncData.HLV.SourceID
-		change.Version = syncData.HLV.Version
+	if len(doc.Xattrs[base.VvXattrName]) > 0 {
+		change.SourceID = syncData.RevAndVersion.CurrentSource
+		change.Version = base.HexCasToUint64(syncData.RevAndVersion.CurrentVersion)
```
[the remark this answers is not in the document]
torcolvin: I think this is fine.

db/document.go
adamcfraser: This sent me looking at import.go, and it seems like there might be an existing bug that we're not setting VvXattrName on line 95 of import.go:
https://github.com/couchbase/sync_gateway/pull/7710/files#diff-aab2fd2f8c58970f5dcd4d3868987d6116b3db2773456c6704d0e8449d9884e9R95

In the delete case, I think we still want to consider any existing HLV when performing the import (the same way we're doing in the non-delete case).  If you agree I think it makes sense to include the fix in this PR, since it's all about handling for HLV xattr.

rest/utilities_testing_resttester.go
```diff
@@ -519,10 +519,10 @@ type RawDocXattrs struct {
 
 // RawDocXattrsWellKnown contains well known fields - see RawDocXattrsOthers for those not known until runtime.
 type RawDocXattrsWellKnown struct {
-	Sync       db.SyncDataJSON `json:"_sync"`
-	GlobalSync map[string]any  `json:"_globalSync"`
-	VV         map[string]any  `json:"_vv"`
-	MOU        map[string]any  `json:"_mou"`
+	Sync       db.SyncData    `json:"_sync"`
```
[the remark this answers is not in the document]
torcolvin: This isn't a correct name.

db/document.go
torcolvin: fixed.

adamcfraser (dismissed):

adamcfraser (approved):

PR #7714, opened 2025-08-28
CBG-4752: Bulk APIs for CV (read and write)

bbrks:
CBG-4752

Read and write support for CVs in bulk APIs:
- `_all_docs`
  - Removed `show_cv` option. Always return CV like we do for other APIs.
- `_bulk_get`
  - Removed `show_cv` option.
  - Renamed/cleanup (make it obvious that RevTree IDs or CVs are accepted with new variable name `revOrCV`)
- `_bulk_docs`
  - Renamed/cleanup
  - Use returned RevTree ID and CV for response
- Test fixes
  - Mostly removing the full/hardcoded bulk API responses since CV is now present

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/55/

### DIFF
--- a/rest/api_test.go
+++ b/rest/api_test.go
@@ -402,48 +402,129 @@ func assertGatewayStatus(t *testing.T, response *TestResponse, expected int) {
 }
 
 func TestBulkDocs(t *testing.T) {
-	rt := NewRestTester(t, nil)
-	defer rt.Close()
+	tests := []struct {
+		occKey string
+	}{
+		{occKey: db.BodyRev},
+		{occKey: db.BodyCV},
+	}
 
-	input := `{"docs": [{"_id": "bulk1", "n": 1}, {"_id": "bulk2", "n": 2}, {"_id": "_local/bulk3", "n": 3}]}`
-	response := rt.SendAdminRequest("POST", "/{{.keyspace}}/_bulk_docs", input)
-	RequireStatus(t, response, 201)
+	for _, test := range tests {
+		t.Run(test.occKey, func(t *testing.T) {
+			rt := NewRestTester(t, nil)
+			defer rt.Close()
 
-	var docs []interface{}
-	assert.NoError(t, base.JSONUnmarshal(response.Body.Bytes(), &docs))
-	assert.Len(t, docs, 3)
-	assert.Equal(t, map[string]interface{}{"rev": "1-50133ddd8e49efad34ad9ecae4cb9907", "id": "bulk1"}, docs[0])
+			const bulk1DocID = "bulk1"
+			const bulk2DocID = "bulk2"
+			const bulk3LocalDocID = "_local/bulk3"
 
-	response = rt.SendAdminRequest("GET", "/{{.keyspace}}/bulk1", "")
-	RequireStatus(t, response, 200)
-	var respBody db.Body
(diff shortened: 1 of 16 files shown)


rest/api_test.go
```diff
@@ -402,48 +402,129 @@ func assertGatewayStatus(t *testing.T, response *TestResponse, expected int) {
...
+			bulk1Version, _ := rt.GetDoc(bulk1DocID)
+			bulk2Version, _ := rt.GetDoc(bulk2DocID)
+			bulk3Version, _ := rt.GetDoc(bulk3LocalDocID)
+
+			var docs []interface{}
+			require.NoError(t, base.JSONUnmarshal(response.Body.Bytes(), &docs))
+			require.Len(t, docs, 3)
+			assert.Equal(t, map[string]interface{}{"rev": bulk1Version.RevTreeID, "cv": bulk1Version.CV.String(), "id": bulk1DocID}, docs[0])
+			assert.Equal(t, map[string]interface{}{"rev": bulk2Version.RevTreeID, "cv": bulk2Version.CV.String(), "id": bulk2DocID}, docs[1])
+			assert.Equal(t, map[string]interface{}{"rev": bulk3Version.RevTreeID, "id": bulk3LocalDocID}, docs[2])
```
torcolvin: nit:

If you are doing this level of assertion does it makes sense to do:

```
assert.Equal(t, []interface{}{
    map[string]interface{}{"rev": bulk1Version.RevTreeID, "cv": bulk1Version.CV.String(), "id": bulk1DocID},
	map[string]interface{}{"rev": bulk2Version.RevTreeID, "cv": bulk2Version.CV.String(), "id": bulk2DocID},
	map[string]interface{}{"rev": bulk3Version.RevTreeID, "id": bulk3LocalDocID}, docs)

File: rest/bulk_api.go [hunk 2]
```diff
@@ -577,37 +573,36 @@ func (h *handler) handleBulkDocs() error {
...
 			status["status"] = code
 			status["error"] = base.CouchHTTPErrorName(code)
 			status["reason"] = msg
-			base.InfofCtx(h.ctx(), base.KeyAll, "\tBulkDocs: Doc %q --> %d %s (%v)", base.UD(docid), code, msg, err)
-			err = nil // wrote it to output already; not going to return it
+			base.InfofCtx(h.ctx(), base.KeyAll, "\tBulkDocs: Doc %q --> %d %s (%v)", base.UD(docid), code, msg, docErr)
 		} else {
-			status["rev"] = revid
+			status["rev"] = newDoc.GetRevTreeID()
+			status["cv"] = newDoc.RevAndVersion.CV()
```
torcolvin: `Document.RevAndVersion.CV()` will return `@` if there is no cv present.

If this is a no-op modification, will this show up as an error, or is there a possibility that this is a no-op but also a legacy rev and therefore it will not have a CV?

rest/api_test.go
bbrks: Thanks! That definitely makes more sense... tunnel vision from previous test assertion style

rest/bulk_api.go
bbrks: `@` seems like a dangerous zero value since parsers will see the `@` and could assume source and value are both well-formed instead of empty.

Is there any reason why `CV()` wouldn't return `""` instead?

torcolvin (approved):

PR #7718, opened 2025-08-29
CBG-3748: Support legacy RevTree IDs for Delta Sync

bbrks:
CBG-3748

Store a "pointer" revision backup doc keyed from a RevTree ID that points to the CV-based key for retrieval for 3.x clients and delta sync from RevTree IDs.

Can be disabled for customers that only have 4.x clients with a db-level config option, that can also control non-delta sync temporary revision backups (behind a TODO for 4.0, since it requires changes to restore previous functionality)

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/61/

### DIFF
--- a/db/revision.go
+++ b/db/revision.go
@@ -229,25 +229,28 @@ func (body Body) getExpiry() (uint32, bool, error) {
 	return *expiry, true, err
 }
 
-// nonJSONPrefix is used to ensure old revision bodies aren't hidden from N1QL/Views.
-const nonJSONPrefix = byte(1)
-
-// Looks up the raw JSON data of a revision that's been archived to a separate doc.
+// getOldRevisionJSON looks up the raw JSON data of a revision that's been archived to a separate doc.
 // If the revision isn't found (e.g. has been deleted by compaction) returns 404 error.
-func (c *DatabaseCollection) getOldRevisionJSON(ctx context.Context, docid string, rev string) ([]byte, error) {
-	data, _, err := c.dataStore.GetRaw(oldRevisionKey(docid, rev))
+func (c *DatabaseCollection) getOldRevisionJSON(ctx context.Context, docid string, revOrCV string) ([]byte, error) {
+	data, _, err := c.dataStore.GetRaw(oldRevisionKey(docid, revOrCV))
 	if base.IsDocNotFoundError(err) {
-		base.DebugfCtx(ctx, base.KeyCRUD, "No old revision %q / %q", base.UD(docid), rev)
-		err = ErrMissing
+		base.DebugfCtx(ctx, base.KeyCRUD, "No old revision %q / %q", base.UD(docid), revOrCV)
+		return nil, ErrMissing
+	} else if err != nil {
+		return nil, err
 	}
-	if data != nil {
-		// Strip out the non-JSON prefix
-		if len(data) > 0 && data[0] == nonJSONPrefix {
-			data = data[1:]
-		}
-		base.DebugfCtx(ctx, base.KeyCRUD, "Got old revision %q / %q --> %d bytes", base.UD(docid), rev, len(data))
+	kind, revData := stripNonJSONPrefix(data)
+	switch kind {
+	case nonJSONPrefixKindRevBody:
+		base.DebugfCtx(ctx, base.KeyCRUD, "Got old revision %q / %q --> %d bytes", base.UD(docid), revOrCV, len(revData))
+		return revData, nil
+	case nonJSONPrefixKindRevPtr:
+		// pointer to a CV-keyed backup revision - do another fetch by CV for the body
+		base.DebugfCtx(ctx, base.KeyCRUD, "Found old revision pointer %q -> %q / %q", revOrCV, base.UD(docid), string(revData))
+		return c.getOldRevisionJSON(ctx, docid, string(revData))
+	default:
+		return nil, fmt.Errorf("unexpected prefix %b for old revision doc %q / %q", kind, base.UD(docid).String(), revOrCV)
 	}
-	return data, err
 }
 
 // Makes a backup of revision body for use by delta sync, and in-flight replications requesting an old revision.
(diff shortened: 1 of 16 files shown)


db/revision_old.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// setOldRevisionJSONPtr stores a pointer from the old revision's RevTreeID to the Current Version hash used to fetch the body.
+func (db *DatabaseCollectionWithUser) setOldRevisionJSONPtr(ctx context.Context, doc *Document, expiry uint32) error {
+	ptrKey := oldRevisionKey(doc.ID, doc.GetRevTreeID())
+	revHash := base.Crc32cHashString([]byte(doc.HLV.GetCurrentVersionString()))
+	ptrBytes := withNonJSONPrefix(nonJSONPrefixKindRevPtr, []byte(revHash))
+	err := db.dataStore.SetRaw(ptrKey, expiry, nil, ptrBytes)
+	if err == nil {
+		base.DebugfCtx(ctx, base.KeyCRUD, "Backed up revision body legacy RevTree ID pointer %q -> %q/%q (%d bytes, ttl:%d)", doc.GetRevTreeID(), base.UD(doc.ID), revHash, len(ptrBytes), db.deltaSyncRevMaxAgeSeconds())
```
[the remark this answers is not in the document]
bbrks: RevTree ID isn't user data

db/revision.go
```diff
@@ -229,25 +229,28 @@ func (body Body) getExpiry() (uint32, bool, error) {
...
+	switch kind {
+	case nonJSONPrefixKindRevBody:
+		base.DebugfCtx(ctx, base.KeyCRUD, "Got old revision %q / %q --> %d bytes", base.UD(docid), revOrCV, len(revData))
+		return revData, nil
+	case nonJSONPrefixKindRevPtr:
+		// pointer to a CV-keyed backup revision - do another fetch by CV for the body
+		base.DebugfCtx(ctx, base.KeyCRUD, "Found old revision pointer %q -> %q / %q", revOrCV, base.UD(docid), string(revData))
+		return c.getOldRevisionJSON(ctx, docid, string(revData))
+	default:
+		return nil, fmt.Errorf("unexpected prefix %b for old revision doc %q / %q", kind, base.UD(docid).String(), revOrCV)
```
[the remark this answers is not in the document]
bbrks: revOrCV isn't user data

db/revision_old.go
```diff
@@ -0,0 +1,57 @@
...
+	nonJSONPrefixKindUnknown nonJSONPrefix = iota
+	// nonJSONPrefixKindRevBody is used for old revision bodies and denotes that everything following is a standard JSON document.
+	nonJSONPrefixKindRevBody
+	// nonJSONPrefixKindRevPtr is used for old revision RevTreeID->CV pointers. The bytes following this are the CV string to be used to fetch the actual body.
+	nonJSONPrefixKindRevPtr
+)
+
+// withNonJSONPrefix returns a new byte slice prefixed with a non-JSON byte. The input slice is not modified.
+func withNonJSONPrefix(kind nonJSONPrefix, body []byte) []byte {
+	buf := make([]byte, 1, len(body)+1)
```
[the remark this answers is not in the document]
bbrks: I don't think this is a real problem. Go has bounds checks for slice indexes.

I could add guards or use alternate methods `bytes.Join` but they're less readable and don't really make anything better.

db/revision.go
torcolvin, line 252: Do you want to do, or are these equivalent?


```suggestion
		return nil, base.RedactErrorf("unexpected prefix %b for old revision doc %q / %q", kind, base.UD(docid).String(), revOrCV)
```

docs/api/components/schemas.yaml
```diff
@@ -1634,6 +1634,13 @@ Database:
       description: 'The amount of milliseconds a N1QL query should run before logging a warning. '
       type: number
       default: 500
+    store_legacy_revtree_data:
+      description: |-
+        Controls whether Sync Gateway stores additional legacy revision tree pointer data to support 3.x/early 4.x clients that still use RevTree IDs (for example when used as delta sources).
```
torcolvin, line 1639: Is there any case other than delta sync where this is relevant?

I might phrase this in closer terms to functional impact. I'm not fussed on the particularly wording, take or leave this.

```
Controls whether Sync Gateway stores a mapping per document to map between a Sync Gateway < 4.0 revision ID and a Sync Gateway 3.0 revision ID.

This can be turned to false if delta_sync is disabled, or all clients (Couchbase Lite, Inter-Sync Gateway Replication, Sync Gateway) have documents that have been modified by Sync Gateway or Couchbase Lite 4.0+
```

If this only affects delta sync, should the parameter be tied to delta sync? Probably not, since you might lose the ability to enable delta sync after a document has been replicated.

torcolvin (approved):

docs/api/components/schemas.yaml
bbrks: It was a config option in `delta_sync` originally but after discussing with Adam in Sync Up last week, this is also related to a feature where we store old revision bodies for 5 mins to cover old 2.x/3.x clients running replication underneath changing documents.

In those cases, the older clients don't have the replacement rev feature we added recently so they need to still be able to access the rev between the time the changes message sent and when it's actually replicated.

With that in mind, I moved the config option up and it can operate independently of `delta_sync`. It's also possible that `store_legacy_revtree_data` could be extended in the far future to actually rip out 3.x/RevTree support completely, but we're not quite there yet.

db/revision.go
bbrks: They're mostly equivalent here. The only difference that can occur when we use `RedactErrorf` is if we're using the error in a returned HTTP response, where the response gets the unredacted version, and the logging includes the redacted version.

In this case it _can_ do that, but it's not a huge deal and I don't expect this default case to be hit anyway.

PR #7723, opened 2025-09-01
CBG-4826 do not managle history if only mv present

torcolvin:
Fixes an issue in topologytests:

```
2025-08-27T18:09:39.310Z [INF] CRUD: c:[2782b301] db:sg1 col:sg_test_0 CheckProposedVersion for doc <ud>doc_MultiActorConflictResurrect_2x_CBL<->SG<->CBS_XDCR_only_1.3</ud> unable to extract proposedHLV from rev message, will be treated as conflict: invalid hlv in changes message received, more than one semi-colon: "6@cbl1;4@rosmar2,3@rosmar1;"
2025-08-27T18:09:39.310Z [DBG] TEST+: c:cbl1 proposeChanges response: [409]
```

The fix is in `proposeChangesBatchEntry.MarshalJSON` but I wrote a test to indicate valid HLV format:

`cv,mv,mv;` is equivalent to `cv,mv,mv`

https://github.com/couchbase/sync_gateway/pull/7722 helps debug this.

NOTE: this doesn't fix all the test failures, but fixes one subset of them.

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/60/
- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/TopologyTests/38/

### DIFF
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -479,6 +479,9 @@ func appendRevocationMacroExpansions(currentSpec []sgbucket.MacroExpansionSpec, 
 //  1. cv only:    		cv
 //  2. cv and pv:  		cv;pv
 //  3. cv, pv, and mv: 	cv,mv;pv
+//  4. cv, mv only:
+//     a. cv,mv;
+//     b. cv,mv
 //
 // Function will return list of revIDs if legacy rev ID was found in the HLV history section (PV)
 // TODO: CBG-3662 - Optimise once we've settled on and tested the format with CBL
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -478,7 +478,7 @@ func appendRevocationMacroExpansions(currentSpec []sgbucket.MacroExpansionSpec, 
 // blip string may be the following formats
 //  1. cv only:    		cv
 //  2. cv and pv:  		cv;pv
-//  3. cv, pv, and mv: 	cv;mv;pv
+//  3. cv, pv, and mv: 	cv,mv;pv
 //
 // Function will return list of revIDs if legacy rev ID was found in the HLV history section (PV)
 // TODO: CBG-3662 - Optimise once we've settled on and tested the format with CBL
--- a/rest/utilities_testing_blip_client.go
+++ b/rest/utilities_testing_blip_client.go
@@ -1249,7 +1249,13 @@ func (e proposeChangeBatchEntry) MarshalJSON() ([]byte, error) {
 	if e.useHLV() {
 		// Until CBG-4461 is implemented the second value in the array is the full HLV.
 		if e.historyStr() != "" {
-			rev += ";" + e.historyStr()
+			// if mv present, this will already contain a ;
+			// else add one
+			if strings.Contains(e.historyStr(), ";") {
+				rev += "," + e.historyStr()
+			} else {
+				rev += ";" + e.historyStr()
+			}
 		}
 	}
 	fmt.Fprintf(output, `["%s","%s"`, e.docID, rev)
--- a/rest/utilities_testing_blip_client.go
+++ b/rest/utilities_testing_blip_client.go
@@ -591,7 +591,7 @@ func (btr *BlipTesterReplicator) handleChanges(ctx context.Context, btc *BlipTes
 						knownRevs[i] = []interface{}{} // sending empty array means we've not seen the doc before, but still want it
 						continue
 					} else if localHLV.DominatesSource(changesVersion) {
-						base.DebugfCtx(ctx, base.KeySGTest, "Skipping changes for incoming doc %q with rev %#v as we already have a newer version %#v", docID, changesVersion.Value, changesVersion.SourceID, localHLV)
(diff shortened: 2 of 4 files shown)


rest/utilities_testing_blip_client.go
```diff
@@ -591,7 +591,7 @@ func (btr *BlipTesterReplicator) handleChanges(ctx context.Context, btc *BlipTes
 						knownRevs[i] = []interface{}{} // sending empty array means we've not seen the doc before, but still want it
 						continue
 					} else if localHLV.DominatesSource(changesVersion) {
-						base.DebugfCtx(ctx, base.KeySGTest, "Skipping changes for incoming doc %q with rev %d@%s as we already have a newer version %#+v", docID, changesVersion.Value, changesVersion.SourceID, localHLV)
+						base.DebugfCtx(ctx, base.KeySGTest, "Skipping changes for incoming doc %q with rev %#v as we already have a newer version %#v", docID, changesVersion.Value, changesVersion.SourceID, localHLV)
```
bbrks: This isn't a correct format string - still passing in two distinct values for `%d@%s`

rest/utilities_testing_test.go
```diff
@@ -333,3 +333,68 @@ func TestRequest(t *testing.T) {
 		_ = Request(http.MethodGet, "%", "")
 	})
 }
+
+func TestUtilitiesProposeChangesBatch(t *testing.T) {
+	// remove test when CBG-4461, this tests existing CBL behavior but not necessarily desired behavior
```
bbrks: CBG-4461 is closed as won't do - and the last comment in there suggests we actually won't be doing it?

rest/utilities_testing_blip_client.go
```diff
@@ -1249,7 +1249,13 @@ func (e proposeChangeBatchEntry) MarshalJSON() ([]byte, error) {
...
 		// Until CBG-4461 is implemented the second value in the array is the full HLV.
 		if e.historyStr() != "" {
-			rev += ";" + e.historyStr()
+			// if mv present, this will already contain a ;
+			// else add one
+			if strings.Contains(e.historyStr(), ";") {
+				rev += "," + e.historyStr()
+			} else {
+				rev += ";" + e.historyStr()
+			}
```
bbrks: `e.historyStr()` isn't cheap, and this is a relatively hot-path in replication - evaluate it once in the `if` and reuse the value?

db/hybrid_logical_vector.go
```diff
@@ -478,7 +478,10 @@ func appendRevocationMacroExpansions(currentSpec []sgbucket.MacroExpansionSpec,
 // blip string may be the following formats
 //  1. cv only:    		cv
 //  2. cv and pv:  		cv;pv
-//  3. cv, pv, and mv: 	cv;mv;pv
+//  3. cv, pv, and mv: 	cv,mv;pv
+//  4. cv, mv only:
+//     a. cv,mv;
+//     b. cv,mv
```
bbrks, line 500: can you add a comment comment describing who uses each of these and when?

The trailing `;` seems weird to me but if it's for a CBL quirk we should document it

bbrks (changes requested):

rest/utilities_testing_test.go
torcolvin: Fixed and added a comment to https://jira.issues.couchbase.com/browse/CBG-4461?focusedCommentId=1257894 explaining that we will never do CBG-4461. When CBL-6633 is implemented post release, this code pathway could be removed provided that we have some testing to make sure we are still detecting and stripping history in proposeChanges. It's primarily an optimization for Couchbase Lite.

db/hybrid_logical_vector.go
torcolvin: I don't know if this is for a CBL quirk or not, it's a Sync Gateway quirk because of https://github.com/couchbase/sync_gateway/blob/588b1e0037d474a0ef5c0eac7c862524f9b47d00/db/hybrid_logical_vector.go#L464

I'm going to inquire as to the desired behavior and get back to this. I will do this in a separate PR if appropriate. Because this is the behavior as exists, I thought I wanted to test this.

I was going to document this in https://github.com/couchbase/couchbase-lite-core/blob/master/modules/docs/pages/replication-protocol.adoc#version-vectors, was there a different place for documentation that you would look?

rest/utilities_testing_blip_client.go
torcolvin: I don't know why this wasn't caught by the compiler, is this something we should add a ruleguard type rule to?

bbrks: I guess the compiler only catches direct usages not stuff that goes through intermediary layers before getting to printf

Probably should add a rule eventually for this, yes. We already have one that catches error wrapping verb `%w` inside logging (which we don't support) - so maybe easy enough to add on to that?

db/hybrid_logical_vector.go
bbrks: I was just thinking inside this godoc header where `a` and `b` are listed.

bbrks (approved):

PR #7729, opened 2025-09-03
CBG-4316:  Remove allow_conflicts option

RIT3shSapata:
CBG-4316

Describe your PR here...
- Removed allow_conflicts option
- Users cannot create a database with `allow_conflicts=true`
- Any existing config loaded with `allow_conflicts=true` will result in the database going offline with an error
- Added test cases to validate `allow_conflicts=true` cannot be set
- Added workaround to existing test cases to enable allow conflicts

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/65/

### DIFF
--- a/rest/api_test.go
+++ b/rest/api_test.go
@@ -3423,17 +3423,23 @@ func TestDocCRUDWithCV(t *testing.T) {
 }
 
 // TestAllowConflictsConfig verifies that the database configuration does not allow
-// the `allow_conflicts` property to be set to true.
-// `allow_conflicts` is no longer supported in SGW 4.0
+// the `allow_conflicts` property to be set to true. `allow_conflicts` is no longer
+// supported in SGW 4.0. This test ensures that a config loaded will add the
+// database to invalid configs. The config will be fixed when a database is created
+// without `allow_conflicts`
 func TestAllowConflictsConfig(t *testing.T) {
+
+	ctx := t.Context()
+
 	rt := NewRestTester(t, &RestTesterConfig{
 		PersistentConfig: true,
 	})
 	defer rt.Close()
 
 	const (
-		dbName  = "db1"
-		errResp = "allow_conflicts cannot be set to true"
+		dbName    = "db1"
+		errResp   = "allow_conflicts cannot be set to true"
+		allDBsURL = "/_all_dbs?verbose=true"
 	)
 
 	dbConfig := rt.NewDbConfig()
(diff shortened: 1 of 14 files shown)


rest/server_context_test.go
```diff
@@ -238,6 +238,15 @@ func TestGetOrAddDatabaseFromConfig(t *testing.T) {
...
 	assert.Equal(t, bucketName, dbContext.BucketSpec.BucketName)
+
+	dbConfig = DbConfig{
+		Name:           "imdb1",
+		AllowConflicts: base.Ptr(true),
+		BucketConfig:   bucketConfig,
+	}
+	_, err = serverContext.AddDatabaseFromConfig(ctx, DatabaseConfig{DbConfig: dbConfig})
+
+	assert.Error(t, err)
```
bbrks: Can you add a stricter assertion here that checks for the error reason?

Something like:

```go
	assert.Contains(t, err.Error(), "x was an invalid option")
```

rest/server_context_test.go
```diff
@@ -238,6 +238,15 @@ func TestGetOrAddDatabaseFromConfig(t *testing.T) {
 	assert.NoError(t, err, "No error while trying to get the existing database name")
 	assert.Equal(t, rosmar.InMemoryURL, dbContext.BucketSpec.Server)
 	assert.Equal(t, bucketName, dbContext.BucketSpec.BucketName)
+
+	dbConfig = DbConfig{
```
bbrks, line 243: add comment describing what this part of the test is doing - it's not immediately obvious outside of the context of this PR

```suggestion
	// config with disallowed allow_conflicts=true
	dbConfig = DbConfig{
```

rest/utilities_testing.go
```diff
@@ -2638,6 +2638,10 @@ func (rt *RestTester) NewDbConfig() DbConfig {
 	return config
 }
 
+func (rt *RestTester) EnableAllowConflicts(dbName string) {
+	rt.RestTesterServerContext.databases_[dbName].Options.AllowConflicts = base.Ptr(true)
+}
```
bbrks: It's probably better to make this a function on `DatabaseContext`, enforce test-only (one that takes in a `testing.TB` parameter, even if unused), since that would allow `db` package tests to enable this option for specific conflict testing or setup as well, not just `RestTester`

Access to functions like this are still possible from `RestTester`-based tests like:
```go
rt.GetDatabase().EnableAllowConflicts(rt.TB())
```

rest/server_context.go
```diff
@@ -986,6 +986,14 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
 		return nil, replicationErr
 	}
 
+	if config.AllowConflicts != nil {
+		base.WarnfCtx(ctx, "AllowConflicts options is no longer supported. Do not use it in the config")
```
bbrks: Since this error message ends up on users, it probably makes sense to use the JSON name (`allow_conflicts`), though I'm not sure we even need to warn if this is set to false.

rest/server_context.go
```diff
@@ -986,6 +986,14 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
...
+		if *config.AllowConflicts {
+			sc._handleInvalidDatabaseConfig(ctx, spec.BucketName, config, db.NewDatabaseError((db.DatabaseAllowConflictsError)))
+			return nil, errors.New("Allow conflicts is true")
```
bbrks: Similarly, improve this error message so users know what they've done wrong, and what they can do to fix it.

This error is not informative enough about why `Allow conflicts is true` is a problem.

rest/api_test.go
```diff
@@ -3406,3 +3406,73 @@ func TestDocCRUDWithCV(t *testing.T) {
...
+		errResp = "allow_conflicts cannot be set to true"
+	)
+
+	dbConfig := rt.NewDbConfig()
+	dbConfig.AllowConflicts = base.Ptr(true)
+
+	resp := rt.CreateDatabase(dbName, dbConfig)
+
+	RequireStatus(t, resp, http.StatusBadRequest)
+	assert.Contains(t, resp.Body.String(), errResp)
```
bbrks, line 3451: You can probably combine this test and the one below:
- do the failed creation before inserting a valid configuration
- then move onto the `ReloadDatabase` / offline testing

bbrks (changes requested):
The actual functionality looks good - just comments around testing and what messages/errors end up being passed back to users in some cases

bbrks (approved):
LGTM 👍 thanks!

PR #7731, opened 2025-09-04
CBG-3748 follow-up: Remove test skips / re-enable some BLIP legacy rev delta sync codepaths

bbrks:
CBG-3748

Follow-up missed in #7718

- Remove test skips
- Restore disabled BLIP replication codepaths for delta sync when using legacy revs
- Remaining test cases still skipped until we store legacy rev revision body backups (non-delta sync case - tracked with another ticket)

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/68/

### DIFF
--- a/db/crud_test.go
+++ b/db/crud_test.go
@@ -1081,8 +1081,15 @@ func BenchmarkHandleRevDelta(b *testing.B) {
 }
 
 func TestGetAvailableRevAttachments(t *testing.T) {
-	t.Skip("Revs are backed up by hash of CV now, test needs to fetch backup rev by revID, CBG-3748 (backwards compatibility for revID)")
 	db, ctx := setupTestDB(t)
+	db.Options.StoreLegacyRevTreeData = true
+
+	// TODO: CBG-4840 Only requires delta sync until restoration of non-delta sync RevTree ID revision body backups"
+	if !base.IsEnterpriseEdition() {
+		t.Skip("CBG-4840 - temp skip see above comment")
+	}
+	db.Options.DeltaSyncOptions = DeltaSyncOptions{Enabled: true, RevMaxAgeSeconds: 300}
+
 	defer db.Close(ctx)
 	collection, ctx := GetSingleDatabaseCollectionWithUser(ctx, t, db)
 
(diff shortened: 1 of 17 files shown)


db/revision_cache_test.go
```diff
@@ -623,10 +623,17 @@ func TestRevisionCacheInternalProperties(t *testing.T) {
...
+	base.SetUpTestLogging(t, base.LevelDebug, base.KeyCRUD)
 
 	db, ctx := setupTestDB(t)
+	db.Options.StoreLegacyRevTreeData = true
+
+	// TODO: CBG-4840 Only requires delta sync until restoration of non-delta sync RevTree ID revision body backups"
+	if !base.IsEnterpriseEdition() {
+		t.Skip("CBG-4840 - temp skip see above comment")
+	}
+	db.Options.DeltaSyncOptions = DeltaSyncOptions{Enabled: true, RevMaxAgeSeconds: 300}
```
torcolvin, line 637: should this be in an else block? since you can't set delta sync on CE

db/crud_test.go
```diff
@@ -1081,8 +1081,15 @@ func BenchmarkHandleRevDelta(b *testing.B) {
...
 func TestGetAvailableRevAttachments(t *testing.T) {
-	t.Skip("Revs are backed up by hash of CV now, test needs to fetch backup rev by revID, CBG-3748 (backwards compatibility for revID)")
 	db, ctx := setupTestDB(t)
+	db.Options.StoreLegacyRevTreeData = true
+
+	// TODO: CBG-4840 Only requires delta sync until restoration of non-delta sync RevTree ID revision body backups"
+	if !base.IsEnterpriseEdition() {
+		t.Skip("CBG-4840 - temp skip see above comment")
+	}
+	db.Options.DeltaSyncOptions = DeltaSyncOptions{Enabled: true, RevMaxAgeSeconds: 300}
```
torcolvin, line 1093: should this be in an else block? since you can't set delta sync on CE

rest/attachment_test.go
```diff
@@ -687,12 +687,18 @@ func TestBulkGetBadAttachmentReproIssue2528(t *testing.T) {
...
 	defer rt.Close()
 
 	dbConfig := rt.NewDbConfig()
 	dbConfig.AllowConflicts = base.Ptr(true)
+
+	// TODO: CBG-4840 Only requires delta sync until restoration of non-delta sync RevTree ID revision body backups"
+	if !base.IsEnterpriseEdition() {
+		t.Skip("CBG-4840 - temp skip see above comment")
+	}
+	dbConfig.DeltaSync = &DeltaSyncConfig{Enabled: base.Ptr(true), RevMaxAgeSeconds: base.Ptr(uint32(300))}
```
torcolvin, line 700: should this be in an else block? since you can't set delta sync on CE

rest/blip_api_crud_test.go
```diff
@@ -1893,8 +1892,9 @@ func TestGetRemovedDoc(t *testing.T) {
 
 	// Delete any temp revisions in case this prevents the bug from showing up (didn't make a difference)
 	tempRevisionDocID := base.RevPrefix + "foo:5:3-cde"
-	err = rt.GetSingleDataStore().Delete(tempRevisionDocID)
-	assert.NoError(t, err, "Unexpected Error")
+	_ = rt.GetSingleDataStore().Delete(tempRevisionDocID)
+	// TODO: CBG-4840 - Requires restoration of non-delta sync RevTree revision backups
+	// assert.NoError(t, err, "Unexpected Error")
```
torcolvin, line 1897: do you want to actually have this assert on non found error, even though the desired behavior is to succeed, so we know if we when we change this code?

Take or leave this.

bbrks: The assertion can be unconditionally restored as soon as CBG-4840 is done, but until then it currently depends on whether delta sync is enabled or not, and it didn't seem worth the hassle to make it work for both

rest/attachment_test.go
bbrks: There's a skip above?

db/crud_test.go
bbrks: There's a skip above?

db/revision_cache_test.go
bbrks: There's a skip above?

bbrks:
@gregns1 maybe cast your eye over the non-test changes here - I'm effectively removing the subprotocol checks to allow delta sync to work for 3.x and 4.x clients - the changes to make this work by storing revtree based revision backups was done in an earlier PR but I missed these changes

torcolvin (approved):

PR #7748, opened 2025-09-11
CBG-3796: Disable setting `enable_star_channel` to false

RIT3shSapata:
CBG-3796

Describe your PR here...
- Removed ability to set `enable_star_channel` to false
- The value will always default to true
- Added test case to check if a DB is created by setting the value to false

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/76/

### DIFF
--- a/rest/api_test.go
+++ b/rest/api_test.go
@@ -3490,3 +3490,30 @@ func TestAllowConflictsConfig(t *testing.T) {
 	RequireStatus(t, resp, http.StatusOK)
 	require.Equal(t, fmt.Sprintf(`[{"db_name":"%s","bucket":"%s","state":"Online"}]`, rt.GetDatabase().Name, rt.GetDatabase().Bucket.GetName()), resp.Body.String())
 }
+
+func TestDisableAllowStarChannel(t *testing.T) {
+	rt := NewRestTester(t, &RestTesterConfig{
+		PersistentConfig: true,
+		SyncFn:           channels.DocChannelsSyncFunction,
+	})
+	defer rt.Close()
+
+	const (
+		dbName  = "db"
+		errResp = "enable_star_channel cannot be set to false"
+		docID   = "doc1"
+		docBody = `{"channels":["testChannel"]}`
+	)
+
+	dbConfig := rt.NewDbConfig()
+	dbConfig.Name = dbName
+	dbConfig.CacheConfig = &CacheConfig{
+		ChannelCacheConfig: &ChannelCacheConfig{
+			EnableStarChannel: base.Ptr(false),
+		},
+	}
+
+	resp := rt.CreateDatabase(dbName, dbConfig)
+	RequireStatus(t, resp, http.StatusBadRequest)
+	assert.Contains(t, resp.Body.String(), errResp)
+}
--- a/rest/api_test.go
+++ b/rest/api_test.go
@@ -3492,11 +3492,11 @@ func TestAllowConflictsConfig(t *testing.T) {
 }
 
 // TestDisableAllowStarChannel verifies that the database configuration does not allow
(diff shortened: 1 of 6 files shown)


rest/api_test.go
```diff
@@ -3490,3 +3490,35 @@ func TestAllowConflictsConfig(t *testing.T) {
 	RequireStatus(t, resp, http.StatusOK)
 	require.Equal(t, fmt.Sprintf(`[{"db_name":"%s","bucket":"%s","state":"Online"}]`, rt.GetDatabase().Name, rt.GetDatabase().Bucket.GetName()), resp.Body.String())
 }
+
+// TestDisableAllowStarChannel verifies that the database configuration does not allow
+// the `enable_star_channel` property to be set to false. The star channel is a special
+// channel in Sync Gateway that provides access to all documents, and disabling it is
+// not supported. This test ensures that an attempt to disable the star channel results
+// in a bad request error.
+func TestDisableAllowStarChannel(t *testing.T) {
```
bbrks, line 3498: This is kinda doing the same thing as `TestDatabaseConfigValidation` just at a higher-level, but what we might want to do here instead is load a config from a bucket that contains the `=false` setting, avoiding validation - not creating/updating through the API (which will run the validation)

Similar to these lines from the `allow_conflicts` change:

https://github.com/couchbase/sync_gateway/pull/7729/files#diff-cb4346bfd3dff6c8e36847bd103c89be564e7ce5d1c8dd0d275404061ada619cR3464-R3474

rest/api_test.go
```diff
@@ -3490,3 +3490,35 @@ func TestAllowConflictsConfig(t *testing.T) {
...
+		PersistentConfig: true,
+		SyncFn:           channels.DocChannelsSyncFunction,
+	})
+	defer rt.Close()
+
+	const (
+		dbName  = "db"
+		errResp = "enable_star_channel cannot be set to false"
+		docID   = "doc1"
+		docBody = `{"channels":["testChannel"]}`
```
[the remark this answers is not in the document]
bbrks: 👍

rest/server_context.go
```diff
@@ -1166,9 +1166,8 @@ func dbcOptionsFromConfig(ctx context.Context, sc *ServerContext, config *DbConf
 				cacheOptions.CacheSkippedSeqMaxWait = time.Duration(*config.CacheConfig.ChannelCacheConfig.MaxWaitSkipped) * time.Millisecond
 			}
 			// set EnableStarChannelLog directly here (instead of via NewDatabaseContext), so that it's set when we create the channels view in ConnectToBucket
-			if config.CacheConfig.ChannelCacheConfig.EnableStarChannel != nil {
-				db.EnableStarChannelLog = *config.CacheConfig.ChannelCacheConfig.EnableStarChannel
-				base.WarnfCtx(ctx, `Deprecation notice: enable_star_channel config option is due to be removed, in future star channel will always be enabled`)
+			if config.CacheConfig.ChannelCacheConfig.EnableStarChannel != nil && !*config.CacheConfig.ChannelCacheConfig.EnableStarChannel {
+				base.WarnfCtx(ctx, `enable_star_channel config option is set to false, changing the value to true`)
```
bbrks: For other cases where we're removing config in 4.0 we've errored instead of warn and override the invalid configuration.
Should we be consistent in how we handle this? (returning error and ensuring the customer realises this config is/needs changing)

If time doesn't allow, especially if this adds additional complexities, I'm OK getting this in as-is, but would've preferred to be consistent.

bbrks (dismissed):
I'll approve but review the comments, and see if you think we should be changing this.

If we weren't up against the release I'd have pushed back more.

bbrks (approved):

PR #7751, opened 2025-09-12
CBG-4828: pre upgraded document conflict resolution

gregns1:
CBG-4828

- Pre upgraded conflicting mutation test cases for ISGR
- Enhancement has been made to not request any changes sent from another peer that have a revID generated CV that is already present locally. This avoids an echo back to a peer after legacy rev has been replicated. 
- If a local doc has no hlv and the rev tree sent by a client is in conflict with local rev tree, we now assign implicit hlv based on local SourceID and local revID and use that to run through HLV conflict resolution. 

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/79/

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -1330,12 +1330,7 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 				// conflict detected between incoming rev and local rev revtrees, allow hlv conflict resolvers
 				// below to fix (for ISGR pull)
 				doc.HLV = NewHybridLogicalVector()
-				cvValue, err := LegacyRevToRevTreeEncodedVersion(doc.GetRevTreeID())
-				cvValue.SourceID = db.dbCtx.EncodedSourceID // give it this nodes sourceID
-				if err != nil {
-					return nil, nil, false, nil, err
-				}
-				err = doc.HLV.AddVersion(cvValue)
+				doc.HLV, err = legacyRevToHybridLogicalVector(opts.NewDoc.ID, doc.GetRevTreeID())
 				if err != nil {
 					return nil, nil, false, nil, err
 				}
@@ -3446,6 +3441,13 @@ func parseIncomingChange(docid, rev string) (cvValue Version, err error) {
 	return cvValue, nil
 }
 
+func (db *DatabaseCollectionWithUser) getHLVWithLocalSourceIDFromLegacyRev(docID, revID string) (hlv *HybridLogicalVector, err error) {
+	// no hlv on local doc, convert revID into CV and use that as the local doc hlv
+	hlv, err = legacyRevToHybridLogicalVector(docID, revID)
+	hlv.SourceID = db.dbCtx.EncodedSourceID
+	return hlv, err
+}
+
 func (db *DatabaseCollectionWithUser) CheckChangeVersion(ctx context.Context, docid, rev string) (missing, possible []string) {
 	if strings.HasPrefix(docid, "_design/") && db.user != nil {
 		return // Users can't upload design docs, so ignore them
@@ -3460,14 +3462,12 @@ func (db *DatabaseCollectionWithUser) CheckChangeVersion(ctx context.Context, do
 		return
 	}
 	if hlv == nil {
-		// no hlv on local doc, convert revID into CV and use that as the local doc hlv
-		hlv, err = legacyRevToHybridLogicalVector(docid, syncData.GetRevTreeID())
 		// give local version a sourceID of this node. Needed given if we have a pull replication running and
 		// locally we have 3-def and remote has 3-abc, if we give the legacy rev sourceID to both nodes locally
 		// we will say we don't need 3-abc given 3-def resolves to higher version value (local dominates) but in
 		// reality we need to pull this conflict rev to solve it and push it back up. If we give local hlv our sourceID
 		// then we will say we need 3-abc as it is from a different sourceID and local hlv no longer dominates
-		hlv.SourceID = db.dbCtx.EncodedSourceID
+		hlv, err = db.getHLVWithLocalSourceIDFromLegacyRev(docid, syncData.GetRevTreeID())
 		if err != nil {
 			base.WarnfCtx(ctx, "%s", err)
 			missing = append(missing, rev)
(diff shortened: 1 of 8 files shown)


db/crud.go
```diff
@@ -1315,14 +1315,34 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
...
 		// 1. incoming doc is a tombstone and local doc is also a tombstone (for ISGR pull replications)
 		// 2. local doc has no hlv
 		allowConflictingTombstone := opts.ForceAllowConflictingTombstone && doc.IsDeleted()
+
+		// if local doc is a pre-upgraded mutation and the rev tree is conflicting with incoming, assign this local
+		// version an implicit hlv based on its rev tree ID + this nodes sourceID to allow HLV conflict resolution to occur
+		// between the two versions.
+		if doc.HLV == nil && doc.Cas != 0 && !allowConflictingTombstone {
+			_, _, conflictErr := db.revTreeConflictCheck(ctx, opts.RevTreeHistory, doc, opts.NewDoc.Deleted)
+			if conflictErr != nil {
```
torcolvin, line 1329:
```suggestion
			if errors.Is(conflictErr, base.ErrDocumentConflict) {
```

you'd have to define `ErrDocumentConflict = base.HTTPErrorf(http.StatusConflict, "Document revision conflict")` but this would future proof this function if it returns a different type of error in the future, and it was not obvious to me that this function would only return this error type

db/crud.go
```diff
@@ -3452,6 +3462,12 @@ func (db *DatabaseCollectionWithUser) CheckChangeVersion(ctx context.Context, do
 	if hlv == nil {
 		// no hlv on local doc, convert revID into CV and use that as the local doc hlv
 		hlv, err = legacyRevToHybridLogicalVector(docid, syncData.GetRevTreeID())
+		// give local version a sourceID of this node. Needed given if we have a pull replication running and
+		// locally we have 3-def and remote has 3-abc, if we give the legacy rev sourceID to both nodes locally
+		// we will say we don't need 3-abc given 3-def resolves to higher version value (local dominates) but in
+		// reality we need to pull this conflict rev to solve it and push it back up. If we give local hlv our sourceID
+		// then we will say we need 3-abc as it is from a different sourceID and local hlv no longer dominates
+		hlv.SourceID = db.dbCtx.EncodedSourceID
```
torcolvin: I'd wrap this up in a specific function,  but I'm not sure how to name it, which isn't a great sign. The logic is duplicated in `PutExistingCurrentVersion`.

I've considered: `func (c *DatabaseCollection) getHLVWithLocalSourceIDFromLegacyRev`

or `encodeRevTreeIDAsHybridLogicalVector(revid string, sourceID string)`

db/crud.go
```diff
@@ -3483,6 +3504,17 @@ func (db *DatabaseCollectionWithUser) CheckChangeVersion(ctx context.Context, do
 	return
 }
 
+// implicitCVEqual will check if two implicit CVs are equal.
+func (db *DatabaseCollectionWithUser) implicitCVEqual(localCV, incomingCV Version) bool {
```
torcolvin: I've used the word `implicit HLV` to mean what happens if XDCR copies a document and it doesn't have an HLV but it is actually `cas@currentSourceID`, so I don't think that's a great word here.

torcolvin (commented):
I need to give this some thought as to whether encoding with <legacyrevtree>@activeSource is the right idea, but I have dropped a few comments that you do not need to address until we decide that this is the right approach.

db/crud.go
```diff
@@ -1315,14 +1315,29 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
...
+		// if local doc is a pre-upgraded mutation and the rev tree is conflicting with incoming, assign this local
+		// version an implicit hlv based on its rev tree ID + this nodes sourceID to allow HLV conflict resolution to occur
+		// between the two versions.
+		if doc.HLV == nil && doc.Cas != 0 && !allowConflictingTombstone {
+			_, _, conflictErr := db.revTreeConflictCheck(ctx, opts.RevTreeHistory, doc, opts.NewDoc.Deleted)
+			if conflictErr != nil {
+				// conflict detected between incoming rev and local rev revtrees, allow hlv conflict resolvers
+				// below to fix (for ISGR pull)
+				doc.HLV = NewHybridLogicalVector()
+				doc.HLV, err = legacyRevToHybridLogicalVector(opts.NewDoc.ID, doc.GetRevTreeID())
```
```suggestion
				doc.HLV, err = legacyRevToHybridLogicalVector(opts.NewDoc.ID, doc.GetRevTreeID())
```

db/crud.go
```diff
@@ -1360,11 +1374,9 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 					}
 				}
 			case HLVConflict:
-				// if the legacy rev list is from ISGR property, we should not do a conflict check
+				// if we have been supplied a rev tree from cbl, perform conflict check on rev tree history
```
torcolvin, line 1376: Why is this only in CBL and not ISGR? Because `PutExistingCurrentVersion` is only used by CBL?

Either way, I might just drop "from cbl"

db/crud.go
```diff
@@ -3483,6 +3504,17 @@ func (db *DatabaseCollectionWithUser) CheckChangeVersion(ctx context.Context, do
 	return
 }
 
+// revIDGeneratedCVEqual will check if two implicit CVs are equal.
```
torcolvin: This code still is weird to me, maybe write something about how this is only used for ISGR conflict checking and should not be used generally.

If I understood correctly, this code doesn't end up serializing this as the HLV and is only used for in flight operations? If it did as encodedver@ISGRSource, then it could conflict if there was another writer.

Also, it probably is only likely to get hit for a match in ISGR because the low ver numbers.

If I don't understand how this is supposed to work, that's fine but I don't know that we should ever have something that uses the revtree encoding version but not `Revision+Tree+Encoding` as the source ID, at least not for something that ends up ever getting written to the bucket.

gregns1: This is never written to the bucket. Essentially the issue is here if you as active peer pull 2-abc and you have 2-def as your local doc giving them both legacy source ID doesn't work. This would evaluate as localHLV dominates incoming hlv given 2-abc generates a lower value to 2-def. In reality for ISGR we want to pull these revisions for conflict resolution. My idea of giving the local version sourceID of the local node would avoid this happening, then local hlv will not dominate incoming hlv. 

I'm going to play around with seeing if I can extract generation from this generated cv to see if there's a way around confusing people like this.

db/crud.go
gregns1: This code block will only be executed for CBL writes given AlignRevTree is only true for ISGR writes here. I plan to clean this up a bit because it does seem confusing but that can take place next week.

torcolvin (approved):

PR #7761, opened 2025-09-16
CBG-4855 Disable concurrent builds for pipeline jobs

torcolvin:
This depends on https://github.com/couchbase/sync_gateway/pull/7760

For freestyle jobs, this parameter can be set, but for pipeline jobs it is not set. If multiple concurrent builds occur on a single instance, it will corrupt the workspace.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,7 +3,7 @@ pipeline {
 
     options {
         timeout(time: 60, unit: 'MINUTES')
-        disableConcurrentBuilds()
+        disableConcurrentBuilds(abortPrevious: true)
     }
 
     environment {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -3,6 +3,7 @@ pipeline {
 
     options {
         timeout(time: 60, unit: 'MINUTES')
+        disableConcurrentBuilds()
```
bbrks: Do we want to abort previous runs instead? Similar to GH actions?

```suggestion
        disableConcurrentBuilds(abortPrevious: true)
```

https://www.jenkins.io/doc/book/pipeline/syntax/#options

bbrks (approved):

Jenkinsfile
torcolvin: good catch

bbrks (dismissed):

bbrks (approved):

PR #7764, opened 2025-09-18
CBG-4811: logging changes for versions

gregns1:
CBG-4811

Small PR to make some changes to how versions are logged to use GoString() interface instead of String() interface.

Some logging has not changed given I feel like it makes more sense we log in transport format for replication communication. Such as:
```
2025-09-17T13:44:45.445Z [DBG] Changes+: c:TestActiveReplicatorMultiCollection-push db:rt1_active col:sg_test_0 MultiChangesFeed sending <ud>{Seq:1, ID:active-doc1, Changes:[map[cv:186615cf67fb0000@Gtq2Kq2aBVMhTdta1sBE9w]]}</ud> <ud>  (to ADMIN)</ud

2025-09-17T13:44:45.447Z [DBG] Sync+: c:TestActiveReplicatorMultiCollection-push db:rt1_active col:sg_test_0 Sending rev "<ud>active-doc1</ud>" 186615cf67fb0000@Gtq2Kq2aBVMhTdta1sBE9w based on 0 known, digests: []

2025-09-17T13:45:58.984Z [DBG] SyncMsg+: c:TestActiveReplicatorRemoteWinsCases/local_doc_has_MV_defined_before_conflict_res-pull db:activedb col:sg_test_0 #3: Type:rev Id:<ud>doc1_</ud> Collection:0 Rev:186615e099660000@1tnlc9bYt6RQa0QFcyHoQA Sequence:2

2025-09-17T13:46:40.072Z [DBG] Sync+: c:[5dafa359] db:passivedb Peer didn't want revision <ud>doc1_merge</ud> / 186615ea285f0000@Gtq2Kq2aBVMhTdta1sBE9w (seq:3)
``` 

## Pre-review checklist

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -1375,7 +1375,7 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 				if len(opts.RevTreeHistory) > 0 && !opts.AlignRevTrees {
 					parent, currentRevIndex, err := db.revTreeConflictCheck(ctx, opts.RevTreeHistory, doc, opts.NewDoc.Deleted)
 					if err != nil {
-						base.InfofCtx(ctx, base.KeyCRUD, "conflict detected between the two HLV's for doc %s, incoming version %#v, local version %#v, and conflict found in rev tree history", base.UD(doc.ID), opts.NewDocHLV.ExtractCurrentVersionFromHLV(), doc.HLV.ExtractCurrentVersionFromHLV())
+						base.DebugfCtx(ctx, base.KeyCRUD, "conflict detected between the two HLV's for doc %s, and conflict found in rev tree history", base.UD(doc.ID))
 						return nil, nil, false, nil, err
 					}
 					_, err = doc.addNewerRevisionsToRevTreeHistory(opts.NewDoc, currentRevIndex, parent, opts.RevTreeHistory)
@@ -1385,7 +1385,7 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 					revTreeAlignedForCBL = true // we have aligned the rev tree for CBL push here so skip later in function
 					doc.HLV.UpdateWithIncomingHLV(opts.NewDocHLV)
 				} else {
-					base.DebugfCtx(ctx, base.KeyCRUD, "conflict detected between the two HLV's for doc %s, incoming version %#v, local version %#v", base.UD(doc.ID), opts.NewDocHLV.ExtractCurrentVersionFromHLV(), doc.HLV.ExtractCurrentVersionFromHLV())
+					base.DebugfCtx(ctx, base.KeyCRUD, "conflict detected between the two HLV's for doc %s", base.UD(doc.ID))
 					if opts.ConflictResolver == nil {
 						// cancel rest of update, HLV is in conflict and no resolver is present
 						return nil, nil, false, nil, base.HTTPErrorf(http.StatusConflict, "Document revision conflict")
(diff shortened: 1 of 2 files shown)


db/crud.go
```diff
@@ -1375,7 +1375,7 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 				if len(opts.RevTreeHistory) > 0 && !opts.AlignRevTrees {
 					parent, currentRevIndex, err := db.revTreeConflictCheck(ctx, opts.RevTreeHistory, doc, opts.NewDoc.Deleted)
 					if err != nil {
-						base.InfofCtx(ctx, base.KeyCRUD, "conflict detected between the two HLV's for doc %s, incoming version %s, local version %s, and conflict found in rev tree history", base.UD(doc.ID), opts.NewDocHLV.GetCurrentVersionString(), doc.HLV.GetCurrentVersionString())
+						base.InfofCtx(ctx, base.KeyCRUD, "conflict detected between the two HLV's for doc %s, incoming version %#v, local version %#v, and conflict found in rev tree history", base.UD(doc.ID), opts.NewDocHLV.ExtractCurrentVersionFromHLV(), doc.HLV.ExtractCurrentVersionFromHLV())
```
torcolvin: I think this logging should be dropped to at least Debug, and maybe moved to VV for test only logging.

What do we log for revtrees in this circumstance? It might just be that this is a 409 conflict as a rev message that 

```suggestion
						base.DebugfCTx(ctx, base.KeyCRUD, "conflict detected between the two HLVs for doc %s, incoming version %#v, local version %#v, and conflict found in rev tree history", base.UD(doc.ID), opts.NewDocHLV.ExtractCurrentVersionFromHLV(), doc.HLV.ExtractCurrentVersionFromHLV())
```

If we do log this I think we probably want not just CV, because without  full history we won't know why this conflicts.

db/crud.go
```diff
@@ -1385,7 +1385,7 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 					revTreeAlignedForCBL = true // we have aligned the rev tree for CBL push here so skip later in function
 					doc.HLV.UpdateWithIncomingHLV(opts.NewDocHLV)
 				} else {
-					base.DebugfCtx(ctx, base.KeyCRUD, "conflict detected between the two HLV's for doc %s, incoming version %v, local version %v", base.UD(doc.ID), opts.NewDocHLV.ExtractCurrentVersionFromHLV(), doc.HLV.ExtractCurrentVersionFromHLV())
+					base.DebugfCtx(ctx, base.KeyCRUD, "conflict detected between the two HLV's for doc %s, incoming version %#v, local version %#v", base.UD(doc.ID), opts.NewDocHLV.ExtractCurrentVersionFromHLV(), doc.HLV.ExtractCurrentVersionFromHLV())
```
torcolvin: This should be dropped, or at best be VV. This is standard operating procedure for any conflict and we don't need to know about it.

If we do need to know about it, we should make the message sound as "positive" as possible, because this isn't an error scenario. Additionally, if we did need to know about it, we probably want the full HLV and not just the CV, since it might not be obvious why it is a conflict.

I think this can be dropped though.

db/crud.go
gregns1: dropped to debug and removed hlv info.

db/crud.go
gregns1: We log at debug for rev tree conflicts but no rev tree info, dropped the HLV info from the log line but certainly think we should be logging at debug for conflicts. It will help us debug if there is an issue around conflict resolution etc.

torcolvin (approved):

PR #7772, opened 2025-09-22
CBG-4849: updates to api docs for ISGR

gregns1:
CBG-4849

- Details the new default policy 
- Note to remove the _rev and _cv properties for doc merges. 

## Pre-review checklist

- [ ] Link upstream PRs

### DIFF
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -709,9 +709,14 @@ Replication:
         - custom
       x-enumDescriptions:
         default: |-
-          In priority order, this will cause
+          For Sync Gateway versions < 4.x, in priority order, this will cause
             - Deletes to always win (the delete with the longest revision history wins if both revisions are deletes)
             - The revision with the longest revision history to win. This means the the revision with the most changes and therefore the highest Revision Tree ID will win.
+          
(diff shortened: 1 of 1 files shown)


docs/api/components/schemas.yaml
```diff
@@ -711,7 +711,7 @@ Replication:
         default: |-
           For Sync Gateway versions < 4.x, in priority order, this will cause
             - Deletes to always win (the delete with the longest revision history wins if both revisions are deletes)
-            - The revision with the longest revision history to win. This means the the revision with the most changes and therefore the highest Revision Tree ID will win.
+            - The revision with the longest revision history to win. This means the revision with the most changes and therefore the highest Revision Tree ID will win.
```
torcolvin: I don't think we need to mention this because the API docs are tied to a release version. We can consult with the team.

torcolvin (commented):
Do we ever explain in the public documentation what `_rev` and `_cv` are in the custom conflict resolver section? I can see that we don't because it's hard to know how you would use them, but are they visible to users who might want to do something like custom merge if there is a flag, or LWW otherwise?

What happens if you include `_rev` or `_cv`? Does ISGR reject this document?

torcolvin (approved):

PR #7775, opened 2025-09-23
CBG-4864 Import document if cv changes, even if body is unchanged

torcolvin:
CBG-4864 make sure to import document if cv changes

- Check _sync.rev.ver and _sync.rev.src against _vv.ver and _vv.src to decide to do an import even if there is a matching body.

Note: this causes behavior change in CBG-4878.

Notes for reviewer:

- Do I need to care about when HLV is nil, and whether I should include implicit CV in these cases?
- Did I test legacy rev sufficiently?
- Are there unit tests that would work better?

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/100/

### DIFF
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -395,17 +395,12 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent, docType DocumentType)
 		if syncData == nil {
 			return
 		}
-		var cv *Version
 		vv := doc.Xattrs[base.VvXattrName]
 		if len(vv) > 0 {
-			cv, err = getCurrentVersionFromVVXattr(vv)
-			if err != nil {
-				base.WarnfCtx(ctx, "Error unmarshalling %s xattr (%q) for doc %q: %v", base.VvXattrName, string(vv), base.UD(docID), err)
+			isSGWrite, _, _ := syncData.IsSGWrite(ctx, event.Cas, doc.Body, rawUserXattr, base.Ptr(rawHLV(vv)))
+			if !isSGWrite {
+				return
 			}
-		}
-		isSGWrite, _, _ := syncData.IsSGWrite(event.Cas, doc.Body, rawUserXattr, cv)
-		if !isSGWrite {
-			return
 		}
 	}
 
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -395,7 +395,15 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent, docType DocumentType)
 		if syncData == nil {
 			return
 		}
-		isSGWrite, _, _ := syncData.IsSGWrite(event.Cas, doc.Body, rawUserXattr)
+		var cv *Version
+		vv := doc.Xattrs[base.VvXattrName]
+		if len(vv) > 0 {
+			cv, err = getCurrentVersionFromVVXattr(vv)
+			if err != nil {
+				base.WarnfCtx(ctx, "Error unmarshalling %s xattr (%q) for doc %q: %v", base.VvXattrName, string(vv), base.UD(docID), err)
+			}
+		}
+		isSGWrite, _, _ := syncData.IsSGWrite(event.Cas, doc.Body, rawUserXattr, cv)
 		if !isSGWrite {
 			return
 		}
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -395,12 +395,14 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent, docType DocumentType)
 		if syncData == nil {
 			return
 		}
+		var rawVV *rawHLV
 		vv := doc.Xattrs[base.VvXattrName]
 		if len(vv) > 0 {
-			isSGWrite, _, _ := syncData.IsSGWrite(ctx, event.Cas, doc.Body, rawUserXattr, base.Ptr(rawHLV(vv)))
-			if !isSGWrite {
-				return
-			}
+			rawVV = base.Ptr(rawHLV(vv))
+		}
+		isSGWrite, _, _ := syncData.IsSGWrite(ctx, event.Cas, doc.Body, rawUserXattr, rawVV)
+		if !isSGWrite {
+			return
 		}
 	}
 
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -395,7 +395,7 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent, docType DocumentType)
 		if syncData == nil {
 			return
 		}
-		isSGWrite, _, _ := syncData.IsSGWrite(event.Cas, doc.Body, rawUserXattr)
+		isSGWrite, _, _ := syncData.IsSGWrite(ctx, event.Cas, doc.Body, rawUserXattr, doc.Xattrs[base.VvXattrName])
 		if !isSGWrite {
 			return
 		}
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -395,7 +395,7 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent, docType DocumentType)
 		if syncData == nil {
 			return
 		}
-		isSGWrite, _, _ := syncData.IsSGWrite(ctx, event.Cas, doc.Body, rawUserXattr, doc.Xattrs[base.VvXattrName])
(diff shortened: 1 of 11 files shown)


db/document.go
```diff
@@ -661,7 +661,38 @@ func HasUserXattrChanged(userXattr []byte, prevUserXattrHash string) bool {
 	return userXattrCrc32cHash(userXattr) != prevUserXattrHash
 }
 
-// SyncData.IsSGWrite - used during feed-based import
+// CVOutdated returns true if the provided CV does not match _sync.rev.ver and _sync.rev.src. The caller is responsible for testing if the values are non-empty.
+// This compares _vv.ver and _vv.src to _sync.rev.ver and _sync.rev.src
+func (s *SyncData) CVOutdated(cvSourceID string, cvVersion uint64) bool {
```
adamcfraser:
```suggestion
func (s *SyncData) CVEqual(cvSourceID string, cvVersion uint64) bool {
```
I feel like calling this CVEqual is more straightforward (but would need to flip the return values from true/false), as "Outdated" suggests to me a "greater than" comparison, as opposed to matching.

db/crud.go
```diff
@@ -66,6 +66,15 @@ func (c *DatabaseCollection) GetDocument(ctx context.Context, docid string, unma
 	return doc, err
 }
 
+// IsSGWrite is used to determine if the document was written by Sync Gateway or needs to be imported.
+func (c *DatabaseCollection) IsSGWrite(ctx context.Context, doc *Document, rawBody []byte) (isSGWrite bool, crc32Match bool, bodyChanged bool) {
+	cv := Version{SourceID: c.dbCtx.EncodedSourceID, Value: doc.Cas}
```
adamcfraser: In the case that there's no HLV defined, I don't think we want to perform the cv comparison at all when deciding when to import, and shouldn't be trying to use implicit cv for this.

Was there a failing scenario in the case where doc.HLV was nil?

db/document.go
```diff
@@ -661,7 +661,38 @@ func HasUserXattrChanged(userXattr []byte, prevUserXattrHash string) bool {
...
+	if cvSourceID != s.RevAndVersion.CurrentSource {
+		return true
+	}
+	return cvVersion != base.HexCasToUint64(s.RevAndVersion.CurrentVersion)
+}
+
+// needsImport checks if a document needs to be imported. Used when Document.IsSGWrite can not be used.
+func (s *SyncData) needsImport(ctx context.Context, cas uint64, rawBody []byte, rawUserXattr []byte, rawHLV []byte) (isSGWrite bool, crc32Match bool) {
```
adamcfraser: I expected SyncData.needsImport and SyncData.IsSGWrite to be identical.  I see that SyncData.IsSGWrite is still used in some places ([like the change cache ](https://github.com/couchbase/sync_gateway/blob/main/db/change_cache.go#L398)).  

There should only be one set of criteria for "IsSGWrite" - either the document has been processed by SGW and can be used (for caching, etc), or it requires import and should not be used until it gets imported.  Given that, I didn't think this PR should be making any changes to how/where IsSGWrite is called, just on the criteria used inside it.

db/document.go
```diff
@@ -661,7 +661,38 @@ func HasUserXattrChanged(userXattr []byte, prevUserXattrHash string) bool {
...
+	isSGWrite, crc32Match, _ = s.IsSGWrite(cas, rawBody, rawUserXattr)
+	if !isSGWrite || rawHLV == nil {
+		return isSGWrite, crc32Match
+	}
+	limitedHLV := struct {
```
adamcfraser: Can we define 'limitedHLV' in hybrid_logical_version, and maybe a helper function there to just unmarshal cv?

adamcfraser (changes requested):

db/crud.go
torcolvin: There were some odd tests that I think were fundamentally wrong where they had _sync.rev.ver data in blip_legacy_revid_test. I was trying to think of whether XDCR creates a vv when it copies a document without a vv, but it definitely does, because it has to know the sourceID where it comes from!

I can't come up with a circumstance where this is true. But, do I actually need to be looking at _mou? I don't think so, since if `_mou.cas` is updated then so is `_sync.cas` by definition.

db/document.go
torcolvin: Good call.

db/document.go
torcolvin: I've changed this to your suggestion but I could still see the merit to the approach I made, there's basically two different functions:

* SyncData.IsSGWrite that has CV already unmarshalled (if it comes from `doc.HLV`), called from on demand import
* SyncData.IsSGWrite where you don't have vv unmarshalled. I was trying to optimize for not unmarshalling _vv if we don't need to in mutation feed or import feed.

I think I could possibly do both via having an interface for either Version/string that will extract the CV only after we do the other comparisons, but maybe this is an overoptimization?

db/change_cache.go
```diff
@@ -395,9 +395,12 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent, docType DocumentType)
 		if syncData == nil {
 			return
 		}
-		isSGWrite, _, _ := syncData.IsSGWrite(event.Cas, doc.Body, rawUserXattr)
-		if !isSGWrite {
-			return
+		vv := doc.Xattrs[base.VvXattrName]
+		if len(vv) > 0 {
```
adamcfraser, line 400: I think moving the IsSGWrite inside the vv check would result in incorrect behaviour for documents with syncData but not vv.  For example, a pre-4.0 SGW write that is updated via the SDK shouldn't be cached until it's been imported, but it looks like this would attempt to cache it since it would have _sync and not _vv.

db/document.go
```diff
@@ -678,21 +686,26 @@ func (s *SyncData) IsSGWrite(cas uint64, rawBody []byte, rawUserXattr []byte) (i
...
 
+	if s.RevAndVersion.CurrentVersion != "" || s.RevAndVersion.CurrentSource != "" {
+		extractedCV, err := cv.ExtractCV()
+		if !errors.Is(err, base.ErrNotFound) {
+			if err != nil {
+				base.WarnfCtx(ctx, "Error extracting cv during IsSGWrite write check: %v", err)
+				return true, true, false
+			}
+			if !s.CVEqual(*extractedCV) {
+				return false, true, false
```
adamcfraser: When this function returns crc32match=true, we increment the Crc32MatchCount stat.  I think the intention of that stat is to track the number of documents that aren't imported because the body hasn't changed.  Given that, I wonder if we should be returning false, false, false here.

db/hybrid_logical_vector.go
```diff
@@ -248,6 +248,14 @@ func (hlv *HybridLogicalVector) ExtractCurrentVersionFromHLV() *Version {
 	return &currVersion
 }
 
+// ExtractCV is used to sastify CV only interface. Since it can never return an error, consider ExtractCurrentVersionFromHLV or GetCurrentVersion instead.
```
adamcfraser: The comment here (about never returning an error) needs updating.

db/document.go
```diff
@@ -678,21 +686,26 @@ func (s *SyncData) IsSGWrite(cas uint64, rawBody []byte, rawUserXattr []byte) (i
 		return false, false, false
 	}
 
+	if s.RevAndVersion.CurrentVersion != "" || s.RevAndVersion.CurrentSource != "" {
+		extractedCV, err := cv.ExtractCV()
+		if !errors.Is(err, base.ErrNotFound) {
+			if err != nil {
+				base.WarnfCtx(ctx, "Error extracting cv during IsSGWrite write check: %v", err)
```
adamcfraser: I think info level logging here is sufficient, and we should remove the word 'error'.   Maybe something like "Unable to extract cv during IsSGWrite check - skipping matching cv check: %v"

adamcfraser (changes requested):

db/change_cache.go
torcolvin: This was just a programming mistake, good catch.

db/document.go
torcolvin: Ah, that's why the UserXattr comparison returns crc32 match  = false even though the crc32 match was true. I thought that was a bug, but I didn't want to change for UserXattr comparison, but I'll change and add a comment here.

adamcfraser (approved):
LGTM - thanks

PR #7776, opened 2025-09-25
CBG-4839 get maxcas for each vbucket

torcolvin:
CBG-4839

Non goals for this ticket:

- Do not need to poll/query for updates if user changes bucket setting, will rely on reloading database.

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/103/
- [ ] `topologytests` TODO

### DIFF
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -52,7 +52,7 @@ const (
 )
 
 const (
-	unknownSourceID = "unknownSource" // used if the document was written before ECCV was enabled
+	unknownSourceID = "Unknown+Source" // used if the document was written before ECCV was enabled
 )
 
 type HLVVersions map[string]uint64 // map of source ID to version uint64 version value
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -51,6 +51,10 @@ const (
 	minPVEntriesRetained         = 3 // minPVEntriesRetained defines the minimum number of PV entries that should be retained after compaction, to avoid removing all history for infrequently updated/replicated documents.
 )
 
+const (
+	unknownSourceID = "unknownSource" // used if the document was written before ECCV was enabled
+)
+
 type HLVVersions map[string]uint64 // map of source ID to version uint64 version value
 
 // sorted will iterate through the map returning entries in a stable sorted order. Used by testing to make it easier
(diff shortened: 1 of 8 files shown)


bbrks:
> - [x] In GetCCVStartingCas, I believe vBucketsMaxCas will not show up if enableCrossClusterVersioning is not set to true. Test manually in unit test framework by removing [e67f6dc/base/main_test_bucket_pool.go#L650C5-L650C29](https://github.com/couchbase/sync_gateway/blob/e67f6dced72f502e44b553e06882daf8de570419/base/main_test_bucket_pool.go#L650C5-L650C29). ECCV is a one way operation

Done

> - [ ] If enableCrossClusterVersioning is not set, or vBucketsMaxCas, initialize values to MaxUint64 (i.e. no cas will be greater).

Not required - needs the enabled condition anyway and falls back to `0`

> - [x] Call GetCCVStartingCas and store on database at load time. I think it's possible we might want to live reload this when someone enables ECCV with a running database. but not as part of this ticket. That future enhancement could use bootstrap polling to test for cluster map changes, which can be used to reset this value.

Kinda done - Same pattern as metadata purge, version pruning window. Lazy init on first call and uses cached values - still allows for force refresh in future.

> - [x] Create interface for rosmar which is going to return these starting cas values as 0. rosmar always supports ECCV effectively. This should simplify DCP code.

Kinda. Rosmar is handled by the `!ok` check

> - [x] Compare cas of document to be imported with CCV starting cas. If greater or equal, use DatabaseContext.EncodedSourceID, otherwise use base64 encoded "Unknown" str.

Done - with additional enabled condition.

> - [x] Change setting source ID based on comparison above.

Done with hardcoded `unknownSource` const

> - [ ] Verify this is used for on demand and feed imports.

Still required!

db/hybrid_logical_vector.go
```diff
@@ -51,6 +51,10 @@ const (
 	minPVEntriesRetained         = 3 // minPVEntriesRetained defines the minimum number of PV entries that should be retained after compaction, to avoid removing all history for infrequently updated/replicated documents.
 )
 
+const (
+	unknownSourceID = "unknownSource" // used if the document was written before ECCV was enabled
```
adamcfraser: Should we use the format `Unknown+Source` for consistency of format with the other well-known sourceID?

adamcfraser (approved):

PR #7798, opened 2025-10-06
CBG-4894: Consider local/remote tombstones in the default conflict resolver for ISGR

gregns1:
CBG-4894

- Adds logic to have local or remote wins from the default resolver if the local or remote document is a tombstone
- This was a bug in 4.0.0 implementation and aligns with what CBL does
- I did think about making this change to blip tester client but topology tests do not seem to like this at all so if we want to make that change I feel it needs its separate ticket given the debugging effort involved to understand why this is happening

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/127/

### DIFF
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -952,6 +952,15 @@ func DefaultLWWConflictResolutionType(ctx context.Context, conflict Conflict) (B
 	if conflict.LocalHLV == nil || conflict.RemoteHLV == nil {
 		return nil, errors.New("local or incoming document is nil for resolveConflict")
 	}
+	localDeleted, _ := conflict.LocalDocument[BodyDeleted].(bool)
+	remoteDeleted, _ := conflict.RemoteDocument[BodyDeleted].(bool)
+	if localDeleted && !remoteDeleted {
+		return conflict.LocalDocument, nil
+	}
+	if remoteDeleted && !localDeleted {
+		return conflict.RemoteDocument, nil
+	}
+
 	// resolve conflict in favor of remote document, remote wins case
 	if conflict.RemoteHLV.Version > conflict.LocalHLV.Version {
 		// remote document wins
--- a/db/hybrid_logical_vector.go
+++ b/db/hybrid_logical_vector.go
@@ -952,8 +952,8 @@ func DefaultLWWConflictResolutionType(ctx context.Context, conflict Conflict) (B
 	if conflict.LocalHLV == nil || conflict.RemoteHLV == nil {
 		return nil, errors.New("local or incoming document is nil for resolveConflict")
 	}
-	localDeleted, _ := conflict.LocalDocument[BodyDeleted].(bool)
-	remoteDeleted, _ := conflict.RemoteDocument[BodyDeleted].(bool)
+	localDeleted := conflict.LocalDocument.IsDeleted()
+	remoteDeleted := conflict.RemoteDocument.IsDeleted()
 	if localDeleted && !remoteDeleted {
 		return conflict.LocalDocument, nil
 	}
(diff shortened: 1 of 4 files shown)


db/hybrid_logical_vector.go
```diff
@@ -952,6 +952,15 @@ func DefaultLWWConflictResolutionType(ctx context.Context, conflict Conflict) (B
 	if conflict.LocalHLV == nil || conflict.RemoteHLV == nil {
 		return nil, errors.New("local or incoming document is nil for resolveConflict")
 	}
+	localDeleted, _ := conflict.LocalDocument[BodyDeleted].(bool)
+	remoteDeleted, _ := conflict.RemoteDocument[BodyDeleted].(bool)
+	if localDeleted && !remoteDeleted {
+		return conflict.LocalDocument, nil
+	}
+	if remoteDeleted && !localDeleted {
+		return conflict.RemoteDocument, nil
```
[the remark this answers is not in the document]
gregns1: No we should be returning the tombstone

db/hybrid_logical_vector.go
```diff
@@ -952,6 +952,15 @@ func DefaultLWWConflictResolutionType(ctx context.Context, conflict Conflict) (B
 	if conflict.LocalHLV == nil || conflict.RemoteHLV == nil {
 		return nil, errors.New("local or incoming document is nil for resolveConflict")
 	}
+	localDeleted, _ := conflict.LocalDocument[BodyDeleted].(bool)
+	remoteDeleted, _ := conflict.RemoteDocument[BodyDeleted].(bool)
```
torcolvin: nit:

I think I'd be inclined to create a `isDeleted` function on `Body` which has a lot of methods like `ExtractDeleted` and `getExpiry`

rest/replicatortest/replicator_conflict_test.go
```diff
@@ -2119,6 +2119,168 @@ func TestActiveReplicatorConflictRemoveCVFromCache(t *testing.T) {
...
+			ReplicationStatsMap:        replicationStats,
+			ConflictResolverFuncForHLV: resolverFunc,
+			CollectionsEnabled:         !rt1.GetDatabase().OnlyDefaultCollection(),
+			Continuous:                 true,
+			SupportedBLIPProtocols:     sgrRunner.SupportedSubprotocols,
+		})
+		require.NoError(t, err)
+		defer func() {
+			require.NoError(t, ar.Stop())
+		}()
```
torcolvin: This is fine but I wonder you can just use `rt1.CreateReplication` to reduce boilerplate? If this isn't easy, I wonder if we should make it easier?

Additionally, I have preference for using passiveRT and activeRT rather than remember that rt1 and rt2 are active

rest/replicatortest/replicator_conflict_test.go
```diff
@@ -2119,6 +2119,168 @@ func TestActiveReplicatorConflictRemoveCVFromCache(t *testing.T) {
...
+		newRev := getRevTreeID(t, rt2Version.RevTreeID, docBodyBytes)
+		conflictResVersion := rt1Version
+		conflictResVersion.RevTreeID = newRev
+
+		// expect local doc to win and still be tombstone with remote docs revision history
+		sgrRunner.WaitForTombstone(docID, rt1, conflictResVersion)
+
+		rt1Collection, rt1ctx := rt1.GetSingleTestDatabaseCollectionWithUser()
+		rt1Doc, err := rt1Collection.GetDocument(rt1ctx, docID, db.DocUnmarshalAll)
+		require.NoError(t, err)
```
torcolvin: I wrote a helper function for rt.GetDocument(docID) to shorten this a bit

torcolvin (commented):
LGTM, I wonder if you want to look if we can shorten the boilerplate for the tests to make them easier to read and this is a good time in the release cycle to think about this.

rest/replicatortest/replicator_conflict_test.go
gregns1: Added a comment to the ticket CBG-4882 for refactoring out tests this way. I think it will fix the issue seen in that ticket and totally makes sense given its less boiler plate code.

torcolvin (approved):

PR #7801, opened 2025-10-06
Test blipsync error response for protocol upgrade

torcolvin:
Validate that Sync Gateway will pass through https://github.com/couchbase/go-blip/blob/d73efab2df065030271cb1007b9257580234fef9/context.go#L258


torcolvin:
This is an example test, I don't think it is required.

PR #7816, opened 2025-10-10
CBG-4345 don't panic if role documents can't be read

torcolvin:
CBG-4345 don't panic if role documents can't be read

If a role can't be read, we want to end the connection because we don't know the right role information to use to use. In the case of BLIP, we should exit with a 503 to indicate to the app to retry. At the point we have a `auth.User` object, we have authenticated and that's probably OK, but there could be an error reading the role document (SDK error, etc).

I changed the handler code to exit with the auth error if we can't read the role document from the user. This is a behavioral change, but I think is the correct one to do, we don't know information about the user. An alternative would be to log an info or error. This is not that dissimilar to the behavior we have now (GetRole would panic, the HTTP handler would exit 500).

This code will need careful review to make sure that the functions and goroutines will terminate correctly.

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/148/


db/changes.go
```diff
@@ -1315,12 +1352,16 @@ func (db *DatabaseCollectionWithUser) DocIDChangesFeed(ctx context.Context, user
 
 	// Subroutine that creates a response row for a document:
 	output := make(chan *ChangeEntry, len(explicitDocIds))
+	defer close(output)
```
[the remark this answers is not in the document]
torcolvin: This code uses this channel to signal that this is done, if the channel isn't closed an early return is needed, then this was a latent bug.

db/design_doc.go
```diff
@@ -268,20 +268,23 @@ func (db *Database) QueryDesignDoc(ctx context.Context, ddocName string, viewNam
...
+func filterViewResult(input sgbucket.ViewResult, user auth.User, applyChannelFiltering bool) (result sgbucket.ViewResult, err error) {
 	hasStarChannel := false
 	var visibleChannels ch.TimedSet
 	if user != nil {
 		// Views only support default collection, so filter based on default collection channels
-		visibleChannels = user.InheritedCollectionChannels(base.DefaultScope, base.DefaultCollection)
+		visibleChannels, err = user.InheritedCollectionChannels(base.DefaultScope, base.DefaultCollection)
 		hasStarChannel = !visibleChannels.Contains("*")
 		if !applyChannelFiltering {
 			return // this is an error
```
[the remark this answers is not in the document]
torcolvin: Leaving existing code as is.

rest/bulk_api.go
```diff
@@ -97,10 +97,14 @@ func (h *handler) handleAllDocs() error {
...
-		availableChannels = h.user.InheritedCollectionChannels(h.collection.ScopeName, h.collection.Name)
+		var err error
+		availableChannels, err = h.user.InheritedCollectionChannels(h.collection.ScopeName, h.collection.Name)
+		if err != nil {
+			return err
+		}
 		if availableChannels == nil {
-			// TODO: CBG-1948
-			panic("no channels for user?")
+			base.AssertfCtx(h.ctx(), "User %q has no channels in handleAllDocs", base.UD(h.user.Name()))
```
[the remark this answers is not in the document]
torcolvin: I do want to assert here, and we still return the same error as a panic but with a better message.

auth/user.go
```diff
@@ -567,15 +579,15 @@ func (user *userImpl) SetPassword(password string) error {
...
+func (user *userImpl) GetRoles() ([]Role, error) {
 	if user.roles == nil {
 		roles := make([]Role, 0, len(user.RoleNames()))
 		deletedRoles := make([]Role, 0)
 		for name := range user.RoleNames() {
 			role, err := user.auth.GetRoleIncDeleted(name)
 			// base.InfofCtx(user.auth.LogCtx, base.KeyAccess, "User %s role %q = %v", base.UD(user.Name_), base.UD(name), base.UD(role))
 			if err != nil {
-				panic(fmt.Sprintf("Error getting user role %q: %v", name, err))
+				return nil, fmt.Errorf("Error getting user role %q: %w", name, err)
```
adamcfraser: Since it's going in logs, I think we want base.UD(name).

db/changes.go
```diff
@@ -674,14 +674,23 @@ func (col *DatabaseCollectionWithUser) checkForUserUpdates(ctx context.Context,
 		base.DebugfCtx(ctx, base.KeyChanges, "MultiChangesFeed reloading user %+v", base.UD(col.user))
 
 		if col.user != nil {
-			previousChannels = col.user.InheritedCollectionChannels(col.ScopeName, col.Name)
+			previousChannels, err = col.user.InheritedCollectionChannels(col.ScopeName, col.Name)
+			if err != nil {
+				base.WarnfCtx(ctx, "Error getting previous channels for user %q: %v", base.UD(col.user.Name()), err)
```
adamcfraser:
```suggestion
				base.WarnfCtx(ctx, "Unable to retrieve inherited channels for user %q: %v", base.UD(col.user.Name()), err)
```

db/changes.go
```diff
@@ -674,14 +674,23 @@ func (col *DatabaseCollectionWithUser) checkForUserUpdates(ctx context.Context,
...
 			previousRoles := col.user.RoleNames()
 			if err := col.ReloadUser(ctx); err != nil {
 				base.WarnfCtx(ctx, "Error reloading user %q: %v", base.UD(col.user.Name()), err)
 				return false, 0, nil, err
 			}
 			// check whether channel set has changed
-			changedChannels = col.user.InheritedCollectionChannels(col.ScopeName, col.Name).CompareKeys(previousChannels)
+			channels, err := col.user.InheritedCollectionChannels(col.ScopeName, col.Name)
+			if err != nil {
+				base.WarnfCtx(ctx, "Error getting channels for user %q: %v", base.UD(col.user.Name()), err)
```
adamcfraser:
```suggestion
				base.WarnfCtx(ctx, "Unable to retrieve inherited channels for user %q: %v", base.UD(col.user.Name()), err)
```

db/changes.go
```diff
@@ -763,7 +772,17 @@ func (col *DatabaseCollectionWithUser) SimpleMultiChangesFeed(ctx context.Contex
 		var channelsSince channels.TimedSet
 		if col.user != nil {
 			var channelsRemoved []string
-			channelsSince, channelsRemoved = col.user.FilterToAvailableCollectionChannels(col.ScopeName, col.Name, chans)
+			var err error
+			channelsSince, channelsRemoved, err = col.user.FilterToAvailableCollectionChannels(col.ScopeName, col.Name, chans)
+			if err != nil {
+				base.WarnfCtx(ctx, "Error filtering to available channels for user %q: %v", base.UD(col.user.Name()), err)
```
adamcfraser:
```suggestion
				base.WarnfCtx(ctx, "Unable to filter to available channels for user %q: %v", base.UD(col.user.Name()), err)
```

db/changes.go
```diff
@@ -1141,7 +1169,16 @@ func (col *DatabaseCollectionWithUser) SimpleMultiChangesFeed(ctx context.Contex
 				return
 			}
 			if userChanged && col.user != nil {
-				newChannelsSince, _ := col.user.FilterToAvailableCollectionChannels(col.ScopeName, col.Name, chans)
+				newChannelsSince, _, err := col.user.FilterToAvailableCollectionChannels(col.ScopeName, col.Name, chans)
+				if err != nil {
+					base.WarnfCtx(ctx, "Error filtering to available channels for user %q: %v", base.UD(col.user.Name()), err)
```
adamcfraser:
```suggestion
					base.WarnfCtx(ctx, "Unable to filter to available channels for user %q: %v", base.UD(col.user.Name()), err)
```

db/changes.go
```diff
@@ -1390,7 +1429,11 @@ func createChangesEntry(ctx context.Context, docid string, db *DatabaseCollectio
 		//   - the active revision is in a channel the user can see (removal==nil)
 		//   - the doc has been removed from a user's channel later the requested since value (removal.Seq > options.Since.Seq).  In this case, we need to send removal:true changes entry
 		for channel, removal := range populatedDoc.Channels {
-			if db.user.CanSeeCollectionChannel(db.ScopeName, db.Name, channel) && (removal == nil || removal.Seq > options.Since.Seq) {
+			canSee, err := db.user.CanSeeCollectionChannel(db.ScopeName, db.Name, channel)
+			if err != nil {
+				return nil, err
```
adamcfraser: In other places we're wrapping the underlying error with additional context.  I think that makes sense to help identify exactly where the failure occurred during changes processing - any reason to not do so here as well?

rest/diagnostic_api.go
```diff
@@ -43,7 +43,10 @@ func (h *handler) getAllUserChannelsResponse(user auth.User) (map[string]map[str
 
 	for _, dsName := range h.db.DataStoreNames() {
 		keyspace := dsName.ScopeName() + "." + dsName.CollectionName()
-		currentChannels := user.InheritedCollectionChannels(dsName.ScopeName(), dsName.CollectionName())
+		currentChannels, err := user.InheritedCollectionChannels(dsName.ScopeName(), dsName.CollectionName())
+		if err != nil {
+			return nil, fmt.Errorf("error getting user channels: %w", err)
```
adamcfraser:
```suggestion
			return nil, fmt.Errorf("error retrieving inherited user channels: %w", err)
```

rest/handler.go
```diff
@@ -915,56 +918,69 @@ func (h *handler) checkPublicAuth(dbCtx *db.DatabaseContext) (err error) {
...
+
+	username := ""
+	if h.isGuest() {
+		username = base.GuestUsername
+	} else if h.user != nil {
+		username = h.user.Name()
+	}
+	roleNames, err := getSGUserRolesForAudit(dbCtx, h.user)
+	if err != nil {
+		base.InfofCtx(h.ctx(), base.KeyHTTP, "Error getting user roles for audit logging: %v", err)
```
adamcfraser:
```suggestion
		base.InfofCtx(h.ctx(), base.KeyHTTP, "Unable to retrieve user roles for audit logging, will be omitted from audit entry: %v", err)
```

adamcfraser (changes requested):
Looks fine.  Generally I don't like including the text "Error" in warning or info logging, made a few suggestions to avoid that, and a few other error handling comments.

adamcfraser (approved):

PR #7828, opened 2025-10-17
CBG-4579: Add uint64 support to few metrics in sgw

RIT3shSapata:
CBG-4579

Describe your PR here...
- Added `SgwUIntStat` to use `uint64` instead of `int64` for few metrics such as sequence and timestamps
- `int64` stats have a possibility of overflowing which would then make the number negative
- Fixed few type conversions in unit tests which use the updated metrics

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/141/

### DIFF
--- a/base/stats.go
+++ b/base/stats.go
@@ -475,7 +475,7 @@ type CacheStats struct {
 	HighSeqCached *SgwUIntStat `json:"high_seq_cached"`
 	// The highest contiguous sequence number that has been cached.
 	HighSeqStable         *SgwUIntStat `json:"high_seq_stable"`
-	NonMobileIgnoredCount *SgwIntStat `json:"non_mobile_ignored_count"`
+	NonMobileIgnoredCount *SgwIntStat  `json:"non_mobile_ignored_count"`
 	// The total number of active channels.
 	NumActiveChannels *SgwIntStat `json:"num_active_channels"`
 	// The total number of skipped sequences. This is a cumulative value.
--- a/base/stats.go
+++ b/base/stats.go
@@ -76,7 +76,7 @@ const (
 	StatUnitUnixTimestamp = "unix timestamp"
 
 	StatFormatInt      = "int"
-	StatFormatUint     = "uint"
+	StatFormatUint     = "uint64"
 	StatFormatFloat    = "float"
 	StatFormatDuration = "duration"
 	StatFormatBool     = "bool"
@@ -1085,39 +1085,7 @@ func (s *SgwIntStat) String() string {
 	return strconv.FormatInt(s.Value(), 10)
 }
 
-// NewUIntStat creates and returns a new unsigned integer Sync Gateway stat (SgwUint64Stat).
-// The stat is initialized to initialValue and, unless SkipPrometheusStatsRegistration is true,
-// is registered with Prometheus. It returns an error if required metadata is missing or if
-// Prometheus registration fails.
-//
-// Parameters:
-//
-//	subsystem: The Prometheus subsystem segment (e.g. resource_utilization, database) used to build the fully qualified metric name.
-//	key: The short metric key appended to the subsystem to form the final metric name.
-//	unit: The unit of measurement (e.g. bytes, seconds). Used for metadata export tooling.
-//	description: Human-readable help text for the metric. Must be non-empty.
-//	addedVersion: The Sync Gateway version the stat was introduced. Must be non-empty.
-//	deprecatedVersion: The version the stat was deprecated, or empty if not deprecated.
-//	stability: The stability level (committed, volatile, or internal). Must be non-empty.
-//	labelKeys: Slice of label keys for constant labels. Must align index-wise with labelVals.
-//	labelVals: Slice of label values corresponding to labelKeys. Length must match labelKeys.
-//	statValueType: The Prometheus value type (counter or gauge) controlling exposition semantics.
-//	initialValue: The initial uint64 value assigned to the stat's atomic counter.
-//
-// Behavior:
-//   - Validates required metadata (description, addedVersion, stability).
-//   - Builds a Prometheus descriptor with any constant labels.
-//   - Sets the initial value atomically.
-//   - Registers the metric with Prometheus unless SkipPrometheusStatsRegistration is true.
-//   - Returns (*SgwUint64Stat, error) where error is non-nil on validation or registration failure.
-//
-// Concurrency:
-//
-//	The underlying value uses atomic.Uint64 for safe concurrent mutation by callers.
-//
-// Errors:
-//
-//	Returned if required fields are missing or if prometheus.Register fails (e.g. duplicate registration).
+// NewUIntStat creates a uint64 stat. The stat is initialized to initialValue.
 func NewUIntStat(subsystem, key, unit, description, addedVersion, deprecatedVersion, stability string, labelKeys, labelVals []string, statValueType prometheus.ValueType, initialValue uint64) (*SgwUint64Stat, error) {
 	stat, err := newSGWStat(subsystem, key, unit, description, addedVersion, deprecatedVersion, stability, labelKeys, labelVals, statValueType)
 	if err != nil {
@@ -1140,6 +1108,7 @@ func NewUIntStat(subsystem, key, unit, description, addedVersion, deprecatedVers
 	return wrappedStat, nil
 }
 
+// FormatString returns a name of the stat type used for generating documentation.
 func (s *SgwUint64Stat) FormatString() string {
 	return StatFormatUint
 }
@@ -1148,6 +1117,7 @@ func (s *SgwUint64Stat) Describe(ch chan<- *prometheus.Desc) {
 	ch <- s.statDesc
 }
 
+// Collect satisfies the prometheus.Collector interface communicates the stat to prometheus. This is known to truncate to float64, which is the only format that prometheus can accept.
 func (s *SgwUint64Stat) Collect(ch chan<- prometheus.Metric) {
 	ch <- prometheus.MustNewConstMetric(s.statDesc, s.statValueType, float64(s.Value()))
 }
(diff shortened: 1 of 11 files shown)


base/stats.go
```diff
@@ -653,19 +653,19 @@ type DatabaseStats struct {
 	// The total amount of bytes read over the public REST api
 	PublicRestBytesRead *SgwIntStat `json:"public_rest_bytes_read"`
 	// The value of the last sequence number assigned. Callers using Set should be holding a mutex or ensure concurrent updates to this value are otherwise safe.
-	LastSequenceAssignedValue *SgwIntStat `json:"last_sequence_assigned_value"` // TODO: CBG-4579 - Replace with SgwUintStat stat
+	LastSequenceAssignedValue *SgwUIntStat `json:"last_sequence_assigned_value"` // TODO: CBG-4579 - Replace with SgwUintStat stat
```
torcolvin:
```suggestion
	LastSequenceAssignedValue *SgwUIntStat `json:"last_sequence_assigned_value"`
```

base/stats.go
```diff
@@ -653,19 +653,19 @@ type DatabaseStats struct {
...
 	// The total number of high sequence lookups.
 	SequenceGetCount *SgwIntStat `json:"sequence_get_count"`
 	// The total number of times the sequence counter document has been incremented.
 	SequenceIncrCount *SgwIntStat `json:"sequence_incr_count"`
 	// The total number of unused, reserved sequences released by Sync Gateway.
-	SequenceReleasedCount *SgwIntStat `json:"sequence_released_count"`
+	SequenceReleasedCount *SgwUIntStat `json:"sequence_released_count"`
 	// The value of the last sequence number reserved (which may not yet be assigned). Callers using Set should be holding a mutex or ensure concurrent updates to this value are otherwise safe.
-	LastSequenceReservedValue *SgwIntStat `json:"last_sequence_reserved_value"` // TODO: CBG-4579 - Replace with SgwUintStat stat
+	LastSequenceReservedValue *SgwUIntStat `json:"last_sequence_reserved_value"` // TODO: CBG-4579 - Replace with SgwUintStat stat
```
torcolvin:
```suggestion
	LastSequenceReservedValue *SgwUIntStat `json:"last_sequence_reserved_value"`
```

base/stats.go
```diff
@@ -1079,6 +1084,48 @@ func (s *SgwIntStat) String() string {
 	return strconv.FormatInt(s.Value(), 10)
 }
 
+func NewUIntStat(subsystem, key, unit, description, addedVersion, deprecatedVersion, stability string, labelKeys, labelVals []string, statValueType prometheus.ValueType, initialValue uint64) (*SgwUIntStat, error) {
```
torcolvin: nit:

I like to add docstrings to any functions, but especially any public functions.

https://tip.golang.org/doc/comment#func is the reference, but these become findable if you run godoc locally and will show up in an editor.

https://go.dev/wiki/CodeReviewComments#doc-comments

https://revive.run/r#exported

We don't flag this as strict, but often the act of writing a comment makes sure your function makes sense.

base/stats.go
```diff
@@ -986,6 +986,11 @@ type SgwIntStat struct {
 	AtomicInt
 }
 
+type SgwUIntStat struct {
+	SgwStat
+	AtomicUInt
```
torcolvin: We used to use `AtomicUInt` before `atomic.Uint64` was part of the std library https://pkg.go.dev/sync/atomic#Uint64

If you use atomic.Uint64, you'll have to create some wrappers, which I think is fine. Eventually I hope we'll be able to drop `base.AtomicUint` and related types.

base/stats.go
```diff
@@ -986,6 +986,11 @@ type SgwIntStat struct {
 	AtomicInt
 }
 
+type SgwUIntStat struct {
```
torcolvin: very much a nit:

I get that this name is trying to match `AtomicUInt` but I'd prefer the SgwUintStat or SgwUint64Stat to match atomic.Uint64

torcolvin (commented):
Very small nits on this.

base/stats.go
RIT3shSapata: removed the comment

base/stats.go
RIT3shSapata: removed the comment

base/stats.go
RIT3shSapata: removed `base.AtomicUint` and embedded `atomic.utint64` directly in struct and wrote wrapper methods for it

base/stats.go
RIT3shSapata: Renamed it to SgwUint64Stat

base/stats.go
RIT3shSapata: done

base/stats.go
```diff
@@ -1079,6 +1085,101 @@ func (s *SgwIntStat) String() string {
...
+//   - Registers the metric with Prometheus unless SkipPrometheusStatsRegistration is true.
+//   - Returns (*SgwUint64Stat, error) where error is non-nil on validation or registration failure.
+//
+// Concurrency:
+//
+//	The underlying value uses atomic.Uint64 for safe concurrent mutation by callers.
+//
+// Errors:
+//
+//	Returned if required fields are missing or if prometheus.Register fails (e.g. duplicate registration).
```
torcolvin: I think copilot is a bit too verbose here, we want to make the docs useful without being too overwhelming. This will show up in editor suggestions. Suggestions:


```suggestion
// NewUIntStat creates a uint64 stat. The stat is initialized to initialValue.
```

This is just a general comment about style, we should not take any action here.


I don't think we should document all of the parameters, but a different design to make documenting all the parmeters here would be an options struct so `subsystem, key, unit, description, addedVersion, deprecatedVersion, stability string, labelKeys, labelVals []string, statValueType prometheus.ValueType` are defined once for each stat and all the restrictions and requirements are defined once. Additionally, if it is defined one on each line in a struct, it can be easier to read and grok.

base/stats.go
```diff
@@ -1079,6 +1085,101 @@ func (s *SgwIntStat) String() string {
...
+		err := prometheus.Register(wrappedStat)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return wrappedStat, nil
+}
+
+func (s *SgwUint64Stat) FormatString() string {
```
torcolvin, line 1112:
```suggestion
// FormatString returns a name of the stat type used for generating documentation.
func (s *SgwUint64Stat) FormatString() string {
```

This shows up in documentation https://docs.couchbase.com/sync-gateway/current/manage/stats-monitoring-prometheus.html and https://docs.couchbase.com/sync-gateway/current/manage/stats-monitoring-json.html

but curiously neither page use this value and it isn't generated by the generator that is used to update these pages. https://github.com/couchbase/sync_gateway/blob/da8b47516f6018f209eb2a94666bb1b87fa4313f/tools/stats-definition-exporter/stat_definition.go#L25

base/stats.go
```diff
@@ -76,6 +76,7 @@ const (
 	StatUnitUnixTimestamp = "unix timestamp"
 
 	StatFormatInt      = "int"
+	StatFormatUint     = "uint"
```
torcolvin: I would name this as uint64, see comment in `FormatString`

base/stats.go
```diff
@@ -1079,6 +1085,101 @@ func (s *SgwIntStat) String() string {
...
+	}
+
+	return wrappedStat, nil
+}
+
+func (s *SgwUint64Stat) FormatString() string {
+	return StatFormatUint
+}
+
+func (s *SgwUint64Stat) Describe(ch chan<- *prometheus.Desc) {
```
torcolvin, line 1116:
```suggestion
// Describe satifies the prometheus.Collector interface communicates the description for prometheus text output.
func (s *SgwUint64Stat) Describe(ch chan<- *prometheus.Desc) {
```

base/stats.go
```diff
@@ -1079,6 +1085,101 @@ func (s *SgwIntStat) String() string {
...
+	ch <- s.statDesc
+}
+
+func (s *SgwUint64Stat) Collect(ch chan<- prometheus.Metric) {
```
torcolvin, line 1121:
```suggestion
// Collect satifies the prometheus.Collector interface communicates the stat to prometheus. This is known to truncate to float64, which is the only format that prometheus can accept.
func (s *SgwUint64Stat) Collect(ch chan<- prometheus.Metric) {
```

base/stats.go
```diff
@@ -1079,6 +1085,101 @@ func (s *SgwIntStat) String() string {
...
+			return
+		}
+
+		if s.CompareAndSwap(cur, value) {
+			return
+		}
+	}
+}
+
+func (s *SgwUint64Stat) Value() uint64 {
```
torcolvin, line 1152:
```suggestion
// Value returns the value. This satifies the AtomicUint interface but new code should prefer SgwUint64Stat.Load to match atomic.Uint64 interface.
func (s *SgwUint64Stat) Value() uint64 {
```

base/stats.go
```diff
@@ -1079,6 +1085,101 @@ func (s *SgwIntStat) String() string {
...
+
+func (s *SgwUint64Stat) MarshalJSON() ([]byte, error) {
+	return []byte(strconv.FormatUint(s.Value(), 10)), nil
+}
+
+func (s *SgwUint64Stat) String() string {
+	return strconv.FormatUint(s.Value(), 10)
+}
+
+func (s *SgwUint64Stat) Set(value uint64) {
```
torcolvin, line 1134:
```suggestion
// Set updates the existing value. This satifies the AtomicUint interface but code can prefer SgwUint64Stat.Store to match atomic.Uint64.
func (s *SgwUint64Stat) Set(value uint64) {
```

torcolvin (commented):
Just some fiddly nits.

base/stats.go
RIT3shSapata: I've updated a the comment. About the struct to pass as a parameter instead of everything, do you want to me file another Jira ticket to track this?

base/stats.go
RIT3shSapata: done

base/stats.go
torcolvin: Nah, this is a comment about how to better design things in the future.

torcolvin (approved):

PR #7829, opened 2025-10-17
CBG-4895: Allow tuning of dcp_read_buffer and kv_buffer server connstr parameters

RIT3shSapata:
CBG-4895

Describe your PR here...
- Currently `dcp_read_buffer` and `kv_buffer` are only supported if CBS is serverless, need to enable them for regular CBS as well for perf tests in capella
-  remove serverless check for unsupported config options such as dcp_read_buffer and kv_buffer

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/142/
- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/146/
- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/147/

### DIFF
--- a/base/constants.go
+++ b/base/constants.go
@@ -252,7 +252,7 @@ func ServerIsTLS(server string) bool {
 // Equivalent to the old regexp: `^(walrus:|file:|/|\.)`
 func ServerIsWalrus(server string) bool {
 	return strings.HasPrefix(server, "walrus:") ||
-		strings.HasPrefix(server, "rosmar:") ||
+		strings.HasPrefix(server, "rosmar") ||
 		strings.HasPrefix(server, "file:") ||
 		strings.HasPrefix(server, "/") ||
 		strings.HasPrefix(server, ".")
--- a/base/constants.go
+++ b/base/constants.go
@@ -252,7 +252,7 @@ func ServerIsTLS(server string) bool {
 // Equivalent to the old regexp: `^(walrus:|file:|/|\.)`
 func ServerIsWalrus(server string) bool {
 	return strings.HasPrefix(server, "walrus:") ||
-		strings.HasPrefix(server, "rosmar") ||
+		strings.HasPrefix(server, "rosmar:") ||
 		strings.HasPrefix(server, "file:") ||
 		strings.HasPrefix(server, "/") ||
 		strings.HasPrefix(server, ".")
--- a/rest/api_test.go
+++ b/rest/api_test.go
@@ -3533,7 +3533,7 @@ func TestDisableAllowStarChannel(t *testing.T) {
 	base.DebugfCtx(t.Context(), base.KeySGTest, "additional logs")
 }
 
-func TestUnsupportedOptions( t *testing.T){
+func TestUnsupportedOptions(t *testing.T) {
 	RequireBucketSpecificCredentials(t)
 	tests := []struct {
 		name            string
--- a/rest/api_test.go
+++ b/rest/api_test.go
@@ -3533,47 +3533,121 @@ func TestDisableAllowStarChannel(t *testing.T) {
 	base.DebugfCtx(t.Context(), base.KeySGTest, "additional logs")
 }
 
-func TestUnsupportedOptions(t *testing.T) {
-	RequireBucketSpecificCredentials(t)
+func TestUnsupportedServerConfigOptions(t *testing.T) {
 	tests := []struct {
 		name            string
 		expectedConnStr string
 		kvBuffer        int
 		dcpBuffer       int
+		serverless      bool
+		params          string
 	}{
 		{
-			name:            "unsupported options specified",
+			name:            "serverless-no_query_param-no_unsupported_options",
+			serverless:      true,
+			expectedConnStr: "?dcp_buffer_size=1048576&idle_http_connection_timeout=90000&kv_buffer_size=1048576&kv_pool_size=1&max_idle_http_connections=64000&max_perhost_idle_http_connections=256",
+		},
+		{
+			name:            "non_serverless-no_query_param_and_no_unsupported_options",
+			serverless:      false,
+			expectedConnStr: "?idle_http_connection_timeout=90000&kv_pool_size=2&max_idle_http_connections=64000&max_perhost_idle_http_connections=256",
+		},
+		{
+			name:            "serverless-no_query_param-unsupported_options",
+			serverless:      true,
 			expectedConnStr: "?dcp_buffer_size=3000&idle_http_connection_timeout=90000&kv_buffer_size=2000&kv_pool_size=1&max_idle_http_connections=64000&max_perhost_idle_http_connections=256",
 			kvBuffer:        2000,
 			dcpBuffer:       3000,
 		},
 		{
-			name:            "default serverless",
-			expectedConnStr: "?dcp_buffer_size=1048576&idle_http_connection_timeout=90000&kv_buffer_size=1048576&kv_pool_size=1&max_idle_http_connections=64000&max_perhost_idle_http_connections=256",
+			name:            "non_serverless-no_query_param-unsupported_options",
+			serverless:      false,
+			expectedConnStr: "?dcp_buffer_size=3000&idle_http_connection_timeout=90000&kv_buffer_size=2000&kv_pool_size=2&max_idle_http_connections=64000&max_perhost_idle_http_connections=256",
+			kvBuffer:        2000,
+			dcpBuffer:       3000,
+		},
+		{
+			name:            "serverless-dcp_buffer_query_param-kv_buffer_unsupported_option",
+			serverless:      true,
+			params:          "?dcp_buffer_size=20",
+			expectedConnStr: "?dcp_buffer_size=20&idle_http_connection_timeout=90000&kv_buffer_size=2000&kv_pool_size=1&max_idle_http_connections=64000&max_perhost_idle_http_connections=256",
+			kvBuffer:        2000,
+		},
+		{
+			name:            "non_serverless-dcp_buffer_query_param-kv_buffer_unsupported_option",
+			serverless:      false,
+			params:          "?dcp_buffer_size=20",
+			expectedConnStr: "?dcp_buffer_size=20&idle_http_connection_timeout=90000&kv_buffer_size=2000&kv_pool_size=2&max_idle_http_connections=64000&max_perhost_idle_http_connections=256",
+			kvBuffer:        2000,
+		},
+		{
+			name:            "serverless-dcp_buffer_query_param-dcp_buffer_unsupported_option",
+			serverless:      true,
+			params:          "?dcp_buffer_size=20",
+			expectedConnStr: "?dcp_buffer_size=20&idle_http_connection_timeout=90000&kv_buffer_size=1048576&kv_pool_size=1&max_idle_http_connections=64000&max_perhost_idle_http_connections=256",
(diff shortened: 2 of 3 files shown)


rest/server_context.go
```diff
@@ -600,22 +600,20 @@ func GetBucketSpec(ctx context.Context, config *DatabaseConfig, serverConfig *St
...
-	if serverConfig.IsServerless() {
-		params := base.DefaultServerlessGoCBConnStringParams()
-		if config.Unsupported != nil {
-			if config.Unsupported.DCPReadBuffer != 0 {
-				params.DcpBufferSize = config.Unsupported.DCPReadBuffer
-			}
-			if config.Unsupported.KVBufferSize != 0 {
-				params.KvBufferSize = config.Unsupported.KVBufferSize
-			}
+	params := base.DefaultServerlessGoCBConnStringParams()
```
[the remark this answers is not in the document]
torcolvin: I don't think this is a nitpick comment from copilot, I think this is critical: see https://github.com/couchbase/sync_gateway/blob/d22c9589bea59a94e780f3fb4e9b0b7ca963418e/rest/main.go#L435-L439 for how we apply this. We _don't_ want the serverless parameters to override the other parameters.

rest/api_test.go
```diff
@@ -3532,3 +3532,48 @@ func TestDisableAllowStarChannel(t *testing.T) {
...
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			ctx := base.TestCtx(t)
+			tb := base.GetTestBucket(t)
+			defer tb.Close(ctx)
+			bucketServer := tb.BucketSpec.Server
+			test.expectedConnStr = bucketServer + test.expectedConnStr
+
+			rt := NewRestTester(t, &RestTesterConfig{CustomTestBucket: tb.NoCloseClone(), PersistentConfig: true, serverless: true})
```
torcolvin: When testing this, you want to make sure that serverless is true and false here, we want to make sure these values are obeyed with or without serverless on the server configuration.

rest/api_test.go
```diff
@@ -3532,3 +3532,48 @@ func TestDisableAllowStarChannel(t *testing.T) {
 	assert.Error(t, err, errResp)
 	base.DebugfCtx(t.Context(), base.KeySGTest, "additional logs")
 }
+
+func TestUnsupportedOptions(t *testing.T) {
+	RequireBucketSpecificCredentials(t)
+	tests := []struct {
```
torcolvin, line 3537: I think we want some extra permutations including:

-  serverless, no query parameters on `server` specification in bootstrap, no unsupported parameters
- non serverless, no query parameters on `server` specification in bootstrap, no unsupported parameters
- serverless, no query parameters on `server` specification in bootstrap, unsupported parameters set
- no serverless, no query parameters on `server` specification in bootstrap, unsupported parameters set
- serverless, bootstrap is `couchbase://foo?dcp_buffer_size=20`, only `unsupported.KvBufferSize` set. 
- no serverless, bootstrap is `couchbase://foo?dcp_buffer_size=20`, only `unsupported.KvBufferSize` set
- serverless, bootstrap is `couchbase://foo?dcp_buffer_size=20`, only `unsupported.DCPReadBuffer` set. 
- no serverless, bootstrap is `couchbase://foo?dcp_buffer_size=20`, only `unsupported.DCPReadBuffer` set.
- serverless, bootstrap is `couchbase://foo?kv_buffer_size=20`, only `unsupported.DCPReadBuffer` set. 
- no serverless, bootstrap is `couchbase://foo?kv_buffer_size=20`, only `unsupported.DCPReadBuffer` set
- serverless, bootstrap is `couchbase://foo?kv_buffer_size=20`, only `unsupported.KvBufferSize` set. 
- no serverless, bootstrap is `couchbase://foo?kv_buffer_size=20`, only `unsupported.KvBufferSize` set.

Basically, we want to make sure we don't change the behavior of which setting wins in a given circumstance.

rest/api_test.go
```diff
@@ -3532,3 +3532,48 @@ func TestDisableAllowStarChannel(t *testing.T) {
...
+		},
+		{
+			name:            "default serverless",
+			expectedConnStr: "?dcp_buffer_size=1048576&idle_http_connection_timeout=90000&kv_buffer_size=1048576&kv_pool_size=1&max_idle_http_connections=64000&max_perhost_idle_http_connections=256",
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			ctx := base.TestCtx(t)
+			tb := base.GetTestBucket(t)
```
torcolvin: I think we can worry about this later (I can come back with a recommendation) but I would be inclined to try to test this with a single bucket or maybe no bucket at all since this is manipulating string parameters. The only reason for this is that `GetTestBucket` takes many seconds when running.

rest/api_test.go
```diff
@@ -3532,3 +3532,48 @@ func TestDisableAllowStarChannel(t *testing.T) {
 	assert.Error(t, err, errResp)
 	base.DebugfCtx(t.Context(), base.KeySGTest, "additional logs")
 }
+
+func TestUnsupportedOptions(t *testing.T) {
+	RequireBucketSpecificCredentials(t)
```
torcolvin: RequireBucketSpecificCredentials should be changed to specify that it needs CBS RBAC credentials.

This test probably needs `SG_TEST_BACKING_STORE=couchbase` to fall into the gocb sections, but it would actually be served by a function that said:

```
		t.Skip("This test only works against Couchbase Server since it requires a gocb bucket")
	}
```

rest/api_test.go
torcolvin: Here's a way to make this faster:

```
func TestUnsupportedServerConfigOptions(t *testing.T) {
	tests := []struct {
		name            string
		expectedConnStr string
		kvBuffer        int
		dcpBuffer       int
	}{
		{
			name:            "unsupported options specified",
			expectedConnStr: "?dcp_buffer_size=3000&idle_http_connection_timeout=90000&kv_buffer_size=2000&kv_pool_size=1&max_idle_http_connections=64000&max_perhost_idle_http_connections=256",
			kvBuffer:        2000,
			dcpBuffer:       3000,
		},
		{
			name:            "default serverless",
			expectedConnStr: "?dcp_buffer_size=1048576&idle_http_connection_timeout=90000&kv_buffer_size=1048576&kv_pool_size=1&max_idle_http_connections=64000&max_perhost_idle_http_connections=256",
		},
	}
	for _, test := range tests {
			ctx := base.TestCtx(t)
			serverBase := "couchbases://example.com"
			sc := &StartupConfig{
				Bootstrap: BootstrapConfig{
					Server: serverBase,
				},
			}
			dbConfig := &DatabaseConfig{
				DbConfig: DbConfig{
					Unsupported: &db.UnsupportedOptions{
						KVBufferSize:  test.kvBuffer,
						DCPReadBuffer: test.dcpBuffer,
					},
				},
			}
			spec, err := GetBucketSpec(ctx, dbConfig, sc)
			require.NoError(t, err)
			require.Equal(t, serverBase+test.expectedConnStr, spec.Server)
		})
	}
}
```

rest/api_test.go
RIT3shSapata: I don't think this is applicable, given, the unit test has been refactored to directly generate the bucketspec

rest/api_test.go
RIT3shSapata: refactored the unit test

rest/api_test.go
RIT3shSapata: added these test cases

rest/api_test.go
RIT3shSapata: done

rest/server_context.go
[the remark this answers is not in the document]
RIT3shSapata: fixed this

rest/server_context.go
```diff
@@ -600,8 +600,14 @@ func GetBucketSpec(ctx context.Context, config *DatabaseConfig, serverConfig *St
 	} else {
 		server = serverConfig.Bootstrap.Server
 	}
-	if serverConfig.IsServerless() {
-		params := base.DefaultServerlessGoCBConnStringParams()
+
+	if strings.HasPrefix(server, "couchbase") {
```
torcolvin:
```suggestion
	if !base.ServerIsWalrus(server) {
```

rest/server_context.go
```diff
@@ -601,7 +601,7 @@ func GetBucketSpec(ctx context.Context, config *DatabaseConfig, serverConfig *St
 		server = serverConfig.Bootstrap.Server
 	}
 
-	if !base.ServerIsWalrus(server) {
+	if !base.ServerIsWalrus(server) && server != "" {
```
torcolvin: This is production code so I'm cautious about modifying it, what's an example test that fails? It feels like we should define a Server in all cases.

RIT3shSapata: `TestDeleteDatabasePointingAtSameBucket` kept failing if I did not add `server != ""`. When I dug into it, the test tries to create a db on the same bucket. 
When the createDB handler handles this request and then tries to setup the config passed [here](https://github.com/couchbase/sync_gateway/blob/dfee3e5193c05702c044375c3eb62b5e03b462a8/rest/admin_api.go#L159), it updates the config string to `""` from `nil`, while inheriting the connection string from bootstrapConfig, [here](https://github.com/couchbase/sync_gateway/blob/dfee3e5193c05702c044375c3eb62b5e03b462a8/rest/config.go#L430). Since this is a unit test, so we do not set the bootstrap config, therefore, I think it is expected for the `Config.Server` to be set to `""`. 
This is why I added this additional check, I know this condition looks very suspicious, but I'm not not sure as to how else do I prevent the test from failing

RIT3shSapata: A slight correction to my comment earlier, 
> while inheriting the connection string from bootstrapConfig, here. Since this is a unit test, so we do not set the bootstrap config, therefore, I think it is expected for the Config.Server to be set to "".

It does not inherit the existing Bootstrap config, instead it inherits an empty bootstrap config

rest/server_context.go
```diff
@@ -595,13 +595,19 @@ func (sc *ServerContext) getOrAddDatabaseFromConfig(ctx context.Context, config
 func GetBucketSpec(ctx context.Context, config *DatabaseConfig, serverConfig *StartupConfig) (base.BucketSpec, error) {
 
 	var server string
-	if config.Server != nil {
+	if config.Server != nil && *config.Server != "" {
```
RIT3shSapata, line 598: I've moved the check for empty string from `_getOrAddDatabaseFromConfig`, i'm doing the check inside the `GetBucketSpec`

base/constants.go
```diff
@@ -252,7 +252,7 @@ func ServerIsTLS(server string) bool {
 // Equivalent to the old regexp: `^(walrus:|file:|/|\.)`
 func ServerIsWalrus(server string) bool {
 	return strings.HasPrefix(server, "walrus:") ||
-		strings.HasPrefix(server, "rosmar:") ||
+		strings.HasPrefix(server, "rosmar") ||
```
```suggestion
		strings.HasPrefix(server, "rosmar:") ||
```

Can we leave this code the way it used to look?

torcolvin (commented):
LGTM otherwise.

torcolvin (approved):

PR #7840, opened 2025-10-28
CBG-4930: Add Revtree to RevMessage's stringer function

RIT3shSapata:
CBG-4930

Describe your PR here...
- Added `RevTree` to `RevMessage`'s stringer function

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/149/

### DIFF
--- a/db/blip_sync_messages.go
+++ b/db/blip_sync_messages.go
@@ -393,11 +393,6 @@ func (rm *RevMessage) Rev() (rev string, found bool) {
 	return rev, found
 }
 
-func (rm *RevMessage) RevTree() (revTree string, found bool) {
-	revTree, found = rm.Properties[RevMessageTreeHistory]
-	return revTree, found
-}
-
 func (rm *RevMessage) Deleted() bool {
 	deleted, found := rm.Properties[RevMessageDeleted]
 	if !found {
@@ -461,6 +456,10 @@ func (rm *RevMessage) String() string {
 
 	if id, foundId := rm.ID(); foundId {
 		buffer.WriteString(fmt.Sprintf("Id:%v ", base.UD(id).Redact()))
+	}
+
+	if rev, foundRev := rm.Rev(); foundRev {
+		buffer.WriteString(fmt.Sprintf("Rev:%v ", rev))
 	}
 
 	if coll, ok := rm.Collection(); ok {
--- a/db/blip_sync_messages.go
+++ b/db/blip_sync_messages.go
@@ -454,6 +454,7 @@ func (rm *RevMessage) SetProperties(properties blip.Properties) {
 	maps.Copy(rm.Properties, properties)
 }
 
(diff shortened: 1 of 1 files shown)


RIT3shSapata:
I'm not sure as to how to test this change as I could not find any other test cases for testing adding properties

torcolvin:
One way I find to make sure that this code works and how it is executed is to put a `panic` in the function and see what code executes the code.

In this case I found that https://github.com/couchbase/sync_gateway/blob/953c1986fba6df6f7d7c12966da72d1c2ca53708/db/blip_handler.go#L1053 will log at debug every line. @gregns1 brought up the idea that we don't want to log the full history for each message at debug - it might not be helpful.

I think for this ticket rather than implementing this as I suggested, we want to add a comment in this function to say that not all properties are implemented in `String()` listing specficially `History` and `RevTree` history.

This is different from how I filed the ticket which I did by looking at code. It is important when adding a property that we check if a class has a custom stringer and to implement it.

Apologies for leading this ticket in the wrong direction initially.

db/blip_sync_messages.go
```diff
@@ -461,10 +467,6 @@ func (rm *RevMessage) String() string {
 		buffer.WriteString(fmt.Sprintf("Collection:%v ", coll))
 	}
 
-	if rev, foundRev := rm.Rev(); foundRev {
-		buffer.WriteString(fmt.Sprintf("Rev:%v ", rev))
-	}
```
torcolvin, line 466: I don't think we want to remove Rev here

db/blip_sync_messages.go
```diff
@@ -393,6 +393,11 @@ func (rm *RevMessage) Rev() (rev string, found bool) {
 	return rev, found
 }
 
+func (rm *RevMessage) RevTree() (revTree string, found bool) {
+	revTree, found = rm.Properties[RevMessageTreeHistory]
+	return revTree, found
+}
```
torcolvin: this change is no longer needed, I think I just want a comment change and that's it.

db/blip_sync_messages.go
```diff
@@ -457,14 +458,14 @@ func (rm *RevMessage) String() string {
...
-		buffer.WriteString(fmt.Sprintf("Collection:%v ", coll))
-	}
-
 	if rev, foundRev := rm.Rev(); foundRev {
 		buffer.WriteString(fmt.Sprintf("Rev:%v ", rev))
 	}
 
+	if coll, ok := rm.Collection(); ok {
+		buffer.WriteString(fmt.Sprintf("Collection:%v ", coll))
+	}
```
torcolvin, line 463: This moves the order of this. It probably doesn't matter, but I'd not even change anything that doesn't need it.

torcolvin (approved):

PR #7841, opened 2025-10-28
CBG-4962 create cookie with SameSite=None if CORS enabled

torcolvin:
CBG-4962 create cookie with SameSite=None if CORS enabled

Preserve existing behavior unless CORS is enabled. If CORS is enabled, pass `SameSite=None` to allow cross origin requests.

## Pre-review checklist

### DIFF
--- a/auth/session.go
+++ b/auth/session.go
@@ -135,7 +135,7 @@ func (auth *Authenticator) GetSession(sessionID string) (*LoginSession, error) {
 	return &session, nil
 }
 
-func (auth *Authenticator) MakeSessionCookie(session *LoginSession, secureCookie bool, httpOnly bool) *http.Cookie {
+func (auth *Authenticator) MakeSessionCookie(session *LoginSession, secureCookie bool, httpOnly bool, sameSite http.SameSite) *http.Cookie {
 	if session == nil {
 		return nil
 	}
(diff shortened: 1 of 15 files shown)


auth/session.go
```diff
@@ -145,6 +145,7 @@ func (auth *Authenticator) MakeSessionCookie(session *LoginSession, secureCookie
 		Expires:  session.Expiration,
 		Secure:   secureCookie,
 		HttpOnly: httpOnly,
+		SameSite: sameSite,
```
adamcfraser, line 149: I had to go verify that http.Cookie doesn't emit the SameSite attribute when SameSiteDefaultMode is specified, to ensure we won't see any functional change for users not using CORS or setting the flag.
It's possible that's worth mentioning the default handling in a function description comment, to make that clear.  (on the fence about the necessity of that, take it or leave it)

docs/api/components/schemas.yaml
```diff
@@ -1698,6 +1698,15 @@ Database:
         remote_config_tls_skip_verify:
           description: Enable self-signed certificates for external JavaScript load.
           type: boolean
+        same_site_cookie:
+          description: |-
+            Override the session cookie SameSite behavior. By default, if CORS is enabled, a session cookie will have SameSite:None. `Default` will omit any SameSite option from the cookie.
```
adamcfraser:
```suggestion
            Override the session cookie SameSite behavior. By default, a session cookie will have SameSite:None if CORS is enabled, and will have no SameSite attribute if CORS is not enabled.  Setting this property to`Default` will omit the SameSite attribute from the cookie.
```

adamcfraser (changes requested):
Looks good to me.  One documentation suggestion, otherwise looks fine.

adamcfraser (approved):

PR #7845, opened 2025-10-30
CBG-4966 create a one time session id for blipsync

torcolvin:
CBG-4966 create a one time session id for blipsync

- Create a `one_time: true` parameter for public `/db/_session` end point that will return a one time session in `one_time_session_id` property
- If handler is `/ks/_blipsync` check `Sec-WebSocket-Protocol` header for a value `SyncGatewaySession={sessionID}`. If found, use this for auth. Modify `http.Request.Header` to remove this header to remove this from logging https://github.com/couchbase/go-blip/blob/main/context.go#L256. It isn't harmful to keep it in, it will just get logged if a client connects to blip that doesn't speak any of the supported protocols. At that point the session will be invalid.

I'm not sure I like the refactoring I did in `handleSessionPOST`, but `getUserFromSessionRequestBody` was hiding a lot of errors, including guest auth. There might be a less risky way to do that, and I'm open to going over it again.

Blocks https://github.com/couchbase/couchbase-lite-js/pull/67

## Pre-review checklist

- [ ] Link upstream PRs

### DIFF
--- a/auth/session.go
+++ b/auth/session.go
@@ -77,6 +77,11 @@ func (auth *Authenticator) AuthenticateCookie(rq *http.Request, response http.Re
 		base.InfofCtx(auth.LogCtx, base.KeyAuth, "Session no longer valid for user %s", base.UD(session.Username))
 		return nil, base.HTTPErrorf(http.StatusUnauthorized, "Session no longer valid for user")
 	}
+	err = auth.deleteOneTimeSession(auth.LogCtx, &session)
+	if err != nil {
+		return nil, err
+	}
+
 	return user, err
 }
 
@@ -88,19 +93,30 @@ func (auth *Authenticator) AuthenticateOneTimeSession(ctx context.Context, sessi
 		return nil, base.HTTPErrorf(http.StatusUnauthorized, "Session Invalid")
 	}
 
-	if session.OneTime != nil && *session.OneTime {
-		err = auth.datastore.Delete(auth.DocIDForSession(sessionID))
-		if err != nil {
-			// If doc is not found, it probably means someone else is simultaneously using the one-time session, error.
-			// If the delete error comes from another source, still treat this as an error, expecting the client to retry
-			// due to a temporary KV issue.
-			if !base.IsDocNotFoundError(err) {
-				base.InfofCtx(ctx, base.KeyAuth, "Unable to delete one-time session %s. Not allowing login: %v", base.UD(sessionID), err)
-			}
-			return nil, base.HTTPErrorf(http.StatusUnauthorized, "Session Invalid")
-		}
+	err = auth.deleteOneTimeSession(ctx, session)
+	if err != nil {
+		return nil, err
 	}
 	return user, nil
+}
+
+// deleteOneTimeSession deletes the session if it is marked as a one-time session. If the session can be not deleted, or
+// was already deleted an error is returned.
+func (auth *Authenticator) deleteOneTimeSession(ctx context.Context, session *LoginSession) error {
+	if session.OneTime == nil || !*session.OneTime {
+		return nil
+	}
+	err := auth.datastore.Delete(auth.DocIDForSession(session.ID))
+	if err != nil {
+		// If doc is not found, it probably means someone else is simultaneously using the one-time session, error.
+		// If the delete error comes from another source, still treat this as an error, expecting the client to retry
+		// due to a temporary KV issue.
+		if !base.IsDocNotFoundError(err) {
+			base.InfofCtx(ctx, base.KeyAuth, "Unable to delete one-time session %s. Not allowing login: %v", base.UD(session.ID), err)
+		}
+		return base.HTTPErrorf(http.StatusUnauthorized, "Session Invalid")
+	}
+	return nil
 }
 
 // CreateSession creates a new login session for the specified user with the specified TTL. If oneTime is true, the
--- a/auth/session.go
+++ b/auth/session.go
@@ -25,6 +25,7 @@ type LoginSession struct {
 	Expiration  time.Time     `json:"expiration"`
 	Ttl         time.Duration `json:"ttl"`
 	SessionUUID string        `json:"session_uuid"` // marker of when the user object changes, to match with session docs to determine if they are valid
+	OneTime     *bool         `json:"one_time,omitempty"`
 }
 
 const DefaultCookieName = "SyncGatewaySession"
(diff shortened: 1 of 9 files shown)


auth/session.go
```diff
@@ -79,7 +80,25 @@ func (auth *Authenticator) AuthenticateCookie(rq *http.Request, response http.Re
...
+func (auth *Authenticator) AuthenticateOneTimeSession(ctx context.Context, sessionID string) (User, error) {
+	_, user, err := auth.GetSession(sessionID)
+	if err != nil {
+		return nil, base.HTTPErrorf(http.StatusUnauthorized, "Session Invalid")
+	}
+	err = auth.datastore.Delete(auth.DocIDForSession(sessionID))
+	if err != nil && !base.IsDocNotFoundError(err) {
+		base.InfofCtx(ctx, base.KeyAuth, "Error deleting one-time session %s: %v, it will expire due to TTL soon", base.UD(sessionID), err)
+	}
+	if err != nil {
```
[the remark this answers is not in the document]
torcolvin: good catch!

auth/session.go
```diff
@@ -79,7 +80,22 @@ func (auth *Authenticator) AuthenticateCookie(rq *http.Request, response http.Re
 	return user, err
 }
 
-func (auth *Authenticator) CreateSession(ctx context.Context, user User, ttl time.Duration) (*LoginSession, error) {
+// AuthenticateOneTimeSession authenticates a session and deletes it upon successful authentication.
+func (auth *Authenticator) AuthenticateOneTimeSession(ctx context.Context, sessionID string) (User, error) {
```
adamcfraser, line 90: Technically multiple requests could successfully GET (concurrently) and auth with the same session.  

One option to avoid this would be to return error and not auth if the Delete gets a DocNotFound.  Do you think that's safe/feasible, or are there corner cases where we expect a DocNotFound?  I think a race with TTL is fine to reject as non-authenticated.

auth/session.go
```diff
@@ -79,7 +80,22 @@ func (auth *Authenticator) AuthenticateCookie(rq *http.Request, response http.Re
...
+	_, user, err := auth.GetSession(sessionID)
+	if err != nil {
+		return nil, base.HTTPErrorf(http.StatusUnauthorized, "Session Invalid")
+	}
+	err = auth.datastore.Delete(auth.DocIDForSession(sessionID))
```
adamcfraser: It feels like we should be checking for session.OneTime actually being set in the session here before deleting the session.

rest/blip_sync_test.go
```diff
@@ -70,3 +73,50 @@ func TestHostOnlyCORS(t *testing.T) {
 		})
 	}
 }
+
+func TestOneTimeSessionBlipSyncAuthentication(t *testing.T) {
```
adamcfraser, line 81: Do we have tests covering removal of blipSessionIDPrefix from secWebSocketProtocolHeader when there are multiple protocols defined (e.g. multiple mobile protocols as well as the blipSession entry, with different ordering)

adamcfraser (changes requested):
Generally looks good - a few comments/questions.

auth/session.go
torcolvin: Should we let the session authenticate? I'm inclined to, but I could be flexible.

auth/session.go
```diff
@@ -79,7 +80,32 @@ func (auth *Authenticator) AuthenticateCookie(rq *http.Request, response http.Re
...
+	}
+
+	if session.OneTime != nil && *session.OneTime {
+		err = auth.datastore.Delete(auth.DocIDForSession(sessionID))
+		if err != nil {
+			// If doc is not found, it probably means someone else is simultaneously using the one-time session, error.
+			// If the delete error comes from another source, still treat this as an error, expecting the client to retry
+			// due to a temporary KV issue.
+			if !base.IsDocNotFoundError(err) {
+				base.InfofCtx(ctx, base.KeyAuth, "Error deleting one-time session %s. Not allowing login: %v", base.UD(sessionID), err)
```
adamcfraser:
```suggestion
				base.InfofCtx(ctx, base.KeyAuth, "Unable to delete one-time session %s. Not allowing login: %v", base.UD(sessionID), err)
```

rest/handler.go
```diff
@@ -1033,6 +1044,30 @@ func (h *handler) checkPublicAuth(dbCtx *db.DatabaseContext) (err error) {
...
+// verifyWebsocketToken checks for a valid websocket token in the request headers. If present, invalidate the token so it can't be reused.
+//
+// Returns a nil error if there is no token on the header but no user object. If the header is found,
+func (h *handler) getWebsocketToken() string {
+	// go blip expects only one Sec-WebSocket-Protocol header, with comma separated values
+	protocolHeaders := h.rq.Header.Get(secWebSocketProtocolHeader)
+	var outputHeaders []string
+	var sessionID string
+	for _, header := range strings.Split(protocolHeaders, ",") {
+		header := strings.TrimSpace(header)
```
adamcfraser: The TrimSpace here means that outputHeaders might not exactly match the incoming headers.  I don't see how this would cause a problem, but it would be preferable to avoid changing anything other than the blipSessionIDPrefix header here?

rest/session_api.go
```diff
@@ -41,50 +47,69 @@ func (h *handler) handleSessionPOST() error {
 		return err
 	}
 
-	// NOTE: handleSessionPOST doesn't handle creating users from OIDC - checkPublicAuth calls out into AuthenticateUntrustedJWT.
-	// Therefore, if by this point `h.user` is guest, this isn't creating a session from OIDC.
-	if h.db.Options.DisablePasswordAuthentication && (h.user == nil || h.user.Name() == "") {
```
adamcfraser, line 46: This early return (and the associated comment) looks like it should be valid under the changes in this PR.  Should we leave as-is to minimize risk of changes in behaviour?

rest/session_api.go
```diff
@@ -41,50 +47,69 @@ func (h *handler) handleSessionPOST() error {
...
-		return h.makeSession(h.user)
-	} else {
+	var params struct {
+		Name     string `json:"name"`
+		Password string `json:"password"`
+		OneTime  bool   `json:"one_time"`
+	}
+	user := h.user
+	if len(body) > 0 {
+		err := base.JSONUnmarshal(body, &params)
```
adamcfraser: What's the reason to not use readJSONInto here, as was previously done?

adamcfraser (changes requested):

rest/session_api.go
torcolvin: readJSONInto will return an EOF error if there's no body - I want to read the body once.

But I'll switch to using `ReadJSONFromMime` to accomplish the same thing.

I couldn't even catch EOF error since https://github.com/couchbase/sync_gateway/blob/e886b0844fa5d462f5d29fcf0aa9d45fc75148a2/rest/multipart.go#L48 will drop the underlying error and the underlying error isn't even io.EOF anyway.

rest/session_api.go
torcolvin: I can put this code back, but I had trouble with this case in unit tests which primarily existed to check the case where you could fall through and accidentally validate a user as a guest after removal of `user, err := h.getUserFromSessionRequestBody()`

I'm pretty confident this case is checked below because it is checked when params.Name != "" is checked but I don't mind keeping this either.

rest/session_api.go
```diff
@@ -46,45 +54,70 @@ func (h *handler) handleSessionPOST() error {
...
+	}
+
+	var params struct {
+		Name     string `json:"name"`
+		Password string `json:"password"`
+		OneTime  bool   `json:"one_time"`
+	}
+	user := h.user
+	if len(body) > 0 {
+		err := ReadJSONFromMIME(h.rq.Header, io.NopCloser(bytes.NewReader(body)), &params)
```
adamcfraser: Can we just use h.requestBody here?  (and avoid reading the body above)

adamcfraser (changes requested):

rest/session_api.go
torcolvin: We can't detect if the body was empty or malformed using this because I can't identify the error from https://github.com/couchbase/sync_gateway/blob/e886b0844fa5d462f5d29fcf0aa9d45fc75148a2/rest/multipart.go#L48 whether it is invalid JSON or empty json - even if this error were wrapped (which it isn't), it isn't an `io.EOF`.

adamcfraser (approved):

PR #7846, opened 2025-10-31
CBG-4892: Add eccv to cluster_info endpoint

RIT3shSapata:
CBG-4892

Describe your PR here...
- Add ECCV values for each bucket as a response for `_cluster_info` endpoint
- This endpoint would be used  by `sgcollect_info` to display this setting of a bucket

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/150/

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -2294,7 +2294,7 @@ type ClusterInfo struct {
 
 type BucketInfo struct {
 	Registry                     *GatewayRegistry `json:"registry,omitempty"`
-	EnableCrossClusterVersioning bool            `json:"enable_cross_cluster_versioning"`
+	EnableCrossClusterVersioning bool             `json:"enable_cross_cluster_versioning"`
 }
 
 // Get SG cluster information.  Iterates over all buckets associated with the server, and returns cluster
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -2293,7 +2293,7 @@ type ClusterInfo struct {
 }
 
 type BucketInfo struct {
-	Registry                     GatewayRegistry `json:"registry,omitempty"`
+	Registry                     *GatewayRegistry `json:"registry,omitempty"`
 	EnableCrossClusterVersioning bool            `json:"enable_cross_cluster_versioning"`
 }
 
@@ -2325,7 +2325,7 @@ func (h *handler) handleGetClusterInfo() error {
 			}
 
 			bucketInfo := BucketInfo{
-				Registry:                     *registry,
+				Registry:                     registry,
 				EnableCrossClusterVersioning: eccv[bucketName],
 			}
 			clusterInfo.Buckets[bucketName] = bucketInfo
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -2293,7 +2293,7 @@ type ClusterInfo struct {
 }
 
 type BucketInfo struct {
-	Registry                     GatewayRegistry `json:"registry"`
+	Registry                     GatewayRegistry `json:"registry,omitempty"`
 	EnableCrossClusterVersioning bool            `json:"enable_cross_cluster_versioning"`
 }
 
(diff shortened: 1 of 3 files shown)


rest/api_test.go
```diff
@@ -3533,6 +3533,39 @@ func TestDisableAllowStarChannel(t *testing.T) {
...
+	type BucketInfoResponse struct {
+		EnableCrossClusterVersioning bool `json:"enable_cross_cluster_versioning"`
+	}
+	type ClusterInfoResponse struct {
+		LegacyConfig bool                          `json:"legacy_config,omitempty"`
+		Buckets      map[string]BucketInfoResponse `json:"buckets,omitempty"`
+	}
+	var clusterInfo ClusterInfoResponse
+	err := base.JSONUnmarshal(resp.Body.Bytes(), &clusterInfo)
+	assert.NoError(t, err)
```
gregns1:
```suggestion
	require.NoError(t, err)
```

rest/admin_api.go
```diff
@@ -2322,8 +2323,11 @@ func (h *handler) handleGetClusterInfo() error {
 				continue
 			}
 
+			eccv := h.server.getBucketCCVSettings(bucketName)
```
gregns1: I would probably change this to avoid iterating multiple times through the map of database contexts. Maybe move this outside the bucket name iteration and have this function just iterate through database context map and return a map of bucket name to eccv value. Then you can use this map to lookup bucket name eccv value here. 

This just avoids iterating through db context map for each bucket name here.

rest/admin_api.go
gregns1, line 2309: We still support persistent config in 4.0. So we want to return eccv value for non persistent config mode too. 

We will also want a test for this case too, you may be able to parameterise the test you already have to have a test case for persistent and non persistent config modes.

rest/api_test.go
```diff
@@ -3533,6 +3533,39 @@ func TestDisableAllowStarChannel(t *testing.T) {
...
+}
```
gregns1, line 3593: I would also add assertion to this to assert that the eccv value is true (given we set it to true for our test suite)

gregns1 (changes requested):
Generally looks good just a few suggestions and the need for the eccv value to be presented in legacy config mode.

rest/api_test.go
```diff
@@ -3533,6 +3533,66 @@ func TestDisableAllowStarChannel(t *testing.T) {
...
+			}
+			type ClusterInfoResponse struct {
+				LegacyConfig bool                          `json:"legacy_config,omitempty"`
+				Buckets      map[string]BucketInfoResponse `json:"buckets,omitempty"`
+			}
+			var clusterInfo ClusterInfoResponse
+			err := base.JSONUnmarshal(resp.Body.Bytes(), &clusterInfo)
+			require.NoError(t, err)
+			assert.True(t, clusterInfo.Buckets[bucketName].EnableCrossClusterVersioning)
+			//RequireStatus(t, rt.SendAdminRequest(http.MethodDelete, fmt.Sprintf("/%s/", tests.dbName), ""), http.StatusOK)
```
gregns1: Should this commented out line be removed?

rest/admin_api.go
```diff
@@ -2293,17 +2293,19 @@ type ClusterInfo struct {
 }
 
 type BucketInfo struct {
-	Registry GatewayRegistry `json:"registry"`
+	Registry                     GatewayRegistry `json:"registry,omitempty"`
```
gregns1: For omit empty to work correctly here I think you'll have to make this GatewayRegistry struct a pointer.

Running in legacy config mode as it stands prints empty registry for the bucket:

Image

gregns1 (changes requested):
Generally looks good just couple of small tweaks and should be ready to go!

gregns1 (approved):

PR #7861, opened 2025-11-03
CBG-4746: Deduplicate channel set outside processEntry

gregns1:
CBG-4746

- Changed delay slightly in tool to allow toll to run faster
- Switch to return list of channel ID not maps from `processEntry` then deduplicate the channel IDs outside `processEntry` mutex

Before:
Screenshot 2025-11-03 at 11 09 17

After:
Screenshot 2025-11-03 at 11 08 57

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/153/

### DIFF
--- a/db/change_cache.go
+++ b/db/change_cache.go
@@ -277,11 +277,9 @@ func (c *changeCache) InsertPendingEntries(ctx context.Context) error {
 	// Trigger _addPendingLogs to process any entries that have been pending too long:
 	c.lock.Lock()
 	changedChannels := c._addPendingLogs(ctx)
-	if c.notifyChangeFunc != nil && len(changedChannels) > 0 {
-		channelSet := channels.SetFromArrayNoValidate(changedChannels)
-		c.notifyChangeFunc(ctx, channelSet)
-	}
 	c.lock.Unlock()
+
+	c.notifyChange(ctx, changedChannels)
 
 	return nil
 }
@@ -563,6 +561,13 @@ type cachePrincipal struct {
 	Sequence uint64 `json:"sequence"`
 }
 
+func (c *changeCache) notifyChange(ctx context.Context, chs []channels.ID) {
+	if c.notifyChangeFunc == nil || len(chs) == 0 {
+		return
+	}
+	c.notifyChangeFunc(ctx, channels.SetFromArrayNoValidate(chs))
+}
+
 func (c *changeCache) Remove(ctx context.Context, collectionID uint32, docIDs []string, startTime time.Time) (count int) {
 	return c.channelCache.Remove(ctx, collectionID, docIDs, startTime)
 }
@@ -634,7 +639,8 @@ func (c *changeCache) releaseUnusedSequenceRange(ctx context.Context, fromSequen
 	}
 
 	// push unused range to either pending or skipped lists based on current state of the change cache
-	allChangedChannels = c.processUnusedRange(ctx, fromSequence, toSequence, allChangedChannels, timeReceived)
+	changedChannels := c.processUnusedRange(ctx, fromSequence, toSequence, timeReceived)
+	allChangedChannels.Update(channels.SetFromArrayNoValidate(changedChannels))
 
 	if c.notifyChangeFunc != nil {
 		c.notifyChangeFunc(ctx, allChangedChannels)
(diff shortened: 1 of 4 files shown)


db/change_cache.go
```diff
@@ -277,8 +277,9 @@ func (c *changeCache) InsertPendingEntries(ctx context.Context) error {
 	// Trigger _addPendingLogs to process any entries that have been pending too long:
 	c.lock.Lock()
 	changedChannels := c._addPendingLogs(ctx)
-	if c.notifyChange != nil && len(changedChannels) > 0 {
-		c.notifyChange(ctx, changedChannels)
+	channelSet := channels.SetFromArrayNoValidate(changedChannels)
+	if c.notifyChange != nil && len(channelSet) > 0 {
+		c.notifyChange(ctx, channelSet)
 	}
```
torcolvin: `channels.SetFromArrayNoValidate` is very cheap to run (one empty allocation) if `changedChannels` is empty, but there also no reason to run the function conceptually.

```suggestion
	if c.notifyChange != nil && len(changedChannels) > 0 {
		channelSet := channels.SetFromArrayNoValidate(changedChannels)
		c.notifyChange(ctx, channelSet)
}
```

db/change_cache.go
```diff
@@ -860,8 +870,8 @@ func (c *changeCache) _addToCache(ctx context.Context, change *LogEntry) channel
 // Add the first change(s) from pendingLogs if they're the next sequence.  If not, and we've been
 // waiting too long for nextSequence, move nextSequence to skipped queue.
 // Returns the channels that changed.
```
torcolvin:
```suggestion
// Returns the channels that changed. This may return a channel more than once.
```

db/change_cache.go
```diff
@@ -732,13 +740,15 @@ func (c *changeCache) processPrincipalDoc(ctx context.Context, docID string, doc
...
-	if c.notifyChange != nil && len(changedChannels) > 0 {
-		c.notifyChange(ctx, changedChannels)
+	channelSet := channels.SetFromArrayNoValidate(changedChannels)
+
+	if c.notifyChange != nil && len(channelSet) > 0 {
+		c.notifyChange(ctx, channelSet)
 	}
 }
 
 // Handles a newly-arrived LogEntry.
```
torcolvin:
```suggestion
// processEntry handles newly-arrived LogEntry and returns the channels affected by this revision. This will be any existing, removed, or added channels. Channels can exist multiple times in this array.
```

I'm not sure that's the best description.

But that left me with another idea, what if you do the deduplication inside `processEntry` but no inside the `lock`? You run the function as is that produces duplicates and then deduplicate before returning to the user? I don't know if this makes any code easier to read.

db/change_cache.go
```diff
@@ -592,14 +596,16 @@ func (c *changeCache) releaseUnusedSequence(ctx context.Context, sequence uint64
...
 	} else {
-		changedChannels.Add(unusedSeqChannelID)
+		channelSet = channels.SetFromArrayNoValidate(changedChannels)
+		channelSet.Add(unusedSeqChannelID)
 	}
-	if c.notifyChange != nil && len(changedChannels) > 0 {
-		c.notifyChange(ctx, changedChannels)
+	if c.notifyChange != nil && len(channelSet) > 0 {
+		c.notifyChange(ctx, channelSet)
 	}
```
torcolvin: Are there any cases of `notifyChange` that you don't have to change? You could consider something like this that might? be more readable. If it's not don't worry about it.

Rename `changeCache.notifyChange` on the struct to `changeChange.notifyChangeFunc`

Then create:

```
func (c *changeCache) notifyChange(ctx context.Context, chs []channels.ID) {
    if !c.notifyChangeFunc || len(chs) == 0 {
		return
	}
    c.notifyChangeFunc(ctx,   channels.SetFromArrayNoValidate(chs))
}
```

torcolvin (commented):
Concept seems fine, I think a quick look to see if we can hide the duplication in our our functions is reasonable, and we should definitely document which functions which could return duplicate channels in output.

db/change_cache.go
gregns1: I agree with the comment needing updating, I've changed this. 

With the deduplicate in `processEntry` but outside the mutex I feel it makes sense when calling the function not in it. I know if we did inside it would deduplicate some code but I think I'd personally prefer to keep the `defer c.lock.Unlock()` in `processEntry` and if we did this change we would have to remove that and call before each return.

db/change_cache.go
gregns1: I have renamed to `notifyChangeFunc` I like the suggestion to have this in separate function but there are some areas in unused sequence processing we call notify anyway. I'm not sure if it adds confusion to add this and maybe a separate function for calling whether channels is empty or not.

db/change_cache.go
```diff
@@ -277,8 +277,9 @@ func (c *changeCache) InsertPendingEntries(ctx context.Context) error {
 	// Trigger _addPendingLogs to process any entries that have been pending too long:
 	c.lock.Lock()
 	changedChannels := c._addPendingLogs(ctx)
-	if c.notifyChange != nil && len(changedChannels) > 0 {
-		c.notifyChange(ctx, changedChannels)
+	if c.notifyChangeFunc != nil && len(changedChannels) > 0 {
+		channelSet := channels.SetFromArrayNoValidate(changedChannels)
+		c.notifyChangeFunc(ctx, channelSet)
 	}
```
torcolvin: I don't think we want to lock here.

gregns1: That's strange we seem to hold the lock for notify here, will move the unlock to after addPendingLogs

torcolvin (approved):

PR #7864, opened 2025-11-04
CBG-4980: support /db/_flush for internal testing

torcolvin:
CBG-4980 support /db/_flush for internal testing

Also addresses:

- CBG-4982 Database recreation after flush doesn't work with persistent config
- CBG-4980 Couchbase Buckets aren't flushable

NOTE: _flush is a database call, so any other databases that are backed by the same bucket will be destroyed.

- /db/_flush for Couchbase Server buckets has been broken since 3.2 as part of https://github.com/couchbase/sync_gateway/commit/617b8732dc724fda92fd5d05fd5e79bd4e25e623
- /db/_flush for persistent config never worked correctly before this commit, because it would delete the database registry and only re-add the database from `AddDatabaseFromConfig`. It would also not block database polling while this happened which could result in a crash.

Behavior changes:

- Support the /db/_flush endpoint which will flush a gocb bucket or delete a rosmar bucket
- Support persistent and non persistent configuration by using `InsertConfig` or `AddDatabaseFromConfig`. Test to make sure config polling no-ops after calling /db/_flush
- Support using serialized rosmar buckets by calling `OpenBucketIn` in `RosmarCluster`. When using `rosmar.InMemoryURL`, `OpenBucket` and `OpenBucketIn` behave identically, but they do not otherwise.
- If a bucket was closed, `rosmar.Bucket.DefaultDataStore()` would just return nil, and we didn't do a nil check. Probably https://github.com/couchbaselabs/rosmar/blob/3b9ac157a3cd52c076649d31cbe385e76386787d/bucket_api.go#L122 should panic, or we change the `sg-bucket.Bucket.DefaultDataStore` API to return errors. I thought this was out of scope for this PR, and we might want to backport this to older releases where we might not want to switch rosmar.

When flushing or deleting a bucket, you need to create a copy of the bucket object.

## Possible Different Approach

Since `/db/_flush` is database scoped, but it flushes the entire bucket, this means other databases that are backed by the same bucket are deleted. I have considered removing this API entirely, and maybe replacing it with an API that is like `/_internal/rosmar/bucketName` that could delete a bucket that is either in memory or on disk. This would be more clear in the behavior but add an API that we don't to use outside of some E2E testing.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/156/

### DIFF
--- a/base/bucket_test.go
+++ b/base/bucket_test.go
@@ -471,9 +471,6 @@ func TestBaseBucket(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			baseBucket := GetBaseBucket(test.bucket)
-			_, ok := baseBucket.(*rosmar.Bucket)
-			assert.True(t, ok, "Base bucket is rosmar")
 			_, err := AsRosmarBucket(test.bucket)
 			require.NoError(t, err)
 		})
--- a/base/bucket_test.go
+++ b/base/bucket_test.go
@@ -473,7 +473,9 @@ func TestBaseBucket(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			baseBucket := GetBaseBucket(test.bucket)
 			_, ok := baseBucket.(*rosmar.Bucket)
-			assert.True(t, ok, "Base bucket wasn't walrus bucket")
+			assert.True(t, ok, "Base bucket is rosmar")
+			_, err := AsRosmarBucket(test.bucket)
+			require.NoError(t, err)
 		})
 	}
 }
--- a/base/collection.go
+++ b/base/collection.go
@@ -28,7 +28,6 @@ import (
 	"github.com/couchbase/gocb/v2"
 	"github.com/couchbase/gocbcore/v10"
 	sgbucket "github.com/couchbase/sg-bucket"
-	"github.com/couchbaselabs/rosmar"
 	pkgerrors "github.com/pkg/errors"
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
@@ -204,15 +203,6 @@ func AsGocbV2Bucket(bucket Bucket) (*GocbV2Bucket, error) {
 		return gocbv2Bucket, nil
 	}
 	return nil, fmt.Errorf("bucket is not a gocb bucket (type %T)", baseBucket)
-}
-
-// AsRosmar returns a bucket as a rosmar.Bucket, or an error if it is not one.
-func AsRosmarBucket(bucket Bucket) (*rosmar.Bucket, error) {
-	baseBucket := GetBaseBucket(bucket)
-	if rosmarBucket, ok := baseBucket.(*rosmar.Bucket); ok {
-		return rosmarBucket, nil
-	}
-	return nil, fmt.Errorf("bucket is not a rosmar bucket (type %T)", baseBucket)
 }
 
 func (b *GocbV2Bucket) GetName() string {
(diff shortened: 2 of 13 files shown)


base/bucket_test.go
```diff
@@ -473,7 +473,9 @@ func TestBaseBucket(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			baseBucket := GetBaseBucket(test.bucket)
 			_, ok := baseBucket.(*rosmar.Bucket)
-			assert.True(t, ok, "Base bucket wasn't walrus bucket")
+			assert.True(t, ok, "Base bucket is rosmar")
+			_, err := AsRosmarBucket(test.bucket)
```
adamcfraser, line 474: It looks like AsRosmarBucket is doing the same work as lines 474-476 - can those be removed or is there a subtle difference there?

base/collection.go
```diff
@@ -204,6 +206,15 @@ func AsGocbV2Bucket(bucket Bucket) (*GocbV2Bucket, error) {
 	return nil, fmt.Errorf("bucket is not a gocb bucket (type %T)", baseBucket)
 }
 
+// AsRosmar returns a bucket as a rosmar.Bucket, or an error if it is not one.
```
adamcfraser: It feels odd to stick a single *rosmar.Bucket helper function in the middle of the GocbV2Bucket implementation of BucketStore (and related interfaces), but I suppose there isn't really a natural location for it elsewhere?  I might prefer it in rosmar_cluster just for the grouping of rosmar functionality, even if it's not cluster-related?  If you feel it belongs where you've got it I'm ok with that, if it was an intentional choice.

base/rosmar_cluster.go
```diff
@@ -20,90 +25,145 @@ var _ BootstrapConnection = &RosmarCluster{}
...
+	if serverURL != rosmar.InMemoryURL {
+		u, err := url.Parse(serverURL)
+		if err != nil {
+			return nil, err
+		}
+		directory := u.Path
+		if runtime.GOOS == "windows" {
+			directory = strings.TrimPrefix(directory, "/")
+		}
+		err = os.MkdirAll(directory, 0750)
```
adamcfraser: Why are group permissions needed here?  (0750 vs 0700)

rest/api.go
```diff
@@ -274,77 +273,52 @@ func (h *handler) handleCompact() error {
...
 			}
-
-		}
-
-		// Re-open database and add to Sync Gateway
-		_, err2 := h.server.AddDatabaseFromConfig(h.ctx(), config.DatabaseConfig)
-		if err2 != nil {
-			return err2
+			defer bucket.Close(h.ctx())
+			// Flush the bucket (assuming it conforms to sgbucket.DeleteableStore interface
```
adamcfraser: It doesn't look like we're using the DeleteableStore interface check any more?

adamcfraser (changes requested):
Generally looks good, a few questions on some of the specifics.

base/collection.go
torcolvin: Yeah, this I'll move this. I kept it close by to `AsGocbV2Bucket` but I think it is weird too.

base/rosmar_cluster.go
torcolvin: Fixed, no good reason.

rest/api.go
torcolvin: Good call, this actually requires Flush anyway.

adamcfraser (approved):

PR #7865, opened 2025-11-06
CBG-4923: Prevent Excessive conflict logging in PutCurrentVersion

RIT3shSapata:
CBG-4923

Describe your PR here...
- Added additional property in document struct to store the revision conflict check to prevent multiple `revTreeConflictCheck`
- Tested this out using the topology test `TestActiveReplicatorConflictPreUpgradedVersionOneSide`

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/155/

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -1352,16 +1352,15 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 		allowConflictingTombstone := opts.ForceAllowConflictingTombstone && doc.IsDeleted()
 
 		// variables to keep track if there was rev tree conflict check and storing the status of the check
-		revTreeConflictCheck := false
+		revTreeConflictChecked := false
 		revTreeConflictCheckStatus := false
 
 		// if local doc is a pre-upgraded mutation and the rev tree is conflicting with incoming, assign this local
 		// version an implicit hlv based on its rev tree ID + this nodes sourceID to allow HLV conflict resolution to occur
 		// between the two versions.
 		if doc.HLV == nil && doc.Cas != 0 && !allowConflictingTombstone {
-			revTreeConflictCheck = true
-			var conflictErr error
-			_, _, conflictErr = db.revTreeConflictCheck(ctx, opts.RevTreeHistory, doc, opts.NewDoc.Deleted)
+			revTreeConflictChecked = true
+			_, _, conflictErr := db.revTreeConflictCheck(ctx, opts.RevTreeHistory, doc, opts.NewDoc.Deleted)
 			if conflictErr != nil {
 				// conflict detected between incoming rev and local rev revtrees, allow hlv conflict resolvers
 				// below to fix (for ISGR pull)
@@ -1387,7 +1386,7 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 				}
 			}
 		} else {
-			conflictStatus := doc.IsInConflict(ctx, db, opts.NewDocHLV, opts, revTreeConflictCheck, revTreeConflictCheckStatus)
+			conflictStatus := doc.IsInConflict(ctx, db, opts.NewDocHLV, opts, revTreeConflictChecked, revTreeConflictCheckStatus)
 			switch conflictStatus {
 			case HLVNoConflictRevAlreadyPresent:
 				base.DebugfCtx(ctx, base.KeyCRUD, "PutExistingCurrentVersion(%q): No new versions to add.  existing: %#v  new:%#v", base.UD(opts.NewDoc.ID), doc.HLV, opts.NewDocHLV)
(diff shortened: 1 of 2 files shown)


db/crud.go
```diff
@@ -1363,6 +1363,7 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 				if err != nil {
 					return nil, nil, false, nil, err
 				}
+				doc.RevConflict = true
```
gregns1: I would potentially avoid adding this boolean to the document struct and just have a boolean declared above this block something like revTreeConflictChecked and mark it true if revTreeConflictCheck has been run. So here you would mark this true before on a new line below line 1368

We _might_ need two different booleans, one to indicate that the check has been run, one to indicate if the rev tree is in conflict.

db/document.go
```diff
@@ -1494,8 +1495,7 @@ func (doc *Document) IsInConflict(ctx context.Context, db *DatabaseCollectionWit
 	}
 	// we need to check if local CV version was generated from a revID if so we need to perform conflict check on rev tree history
 	if doc.HLV.HasRevEncodedCV() {
-		_, _, isConflictErr := db.revTreeConflictCheck(ctx, putOpts.RevTreeHistory, doc, putOpts.NewDoc.Deleted)
-		if isConflictErr != nil {
+		if doc.RevConflict {
```
gregns1: We still need potentially need a check for revtree conflict here, take the scenario:

- Local doc has rev tree generated CV (before receiving the incoming doc over the replicator) 
- Incoming doc arrives and also has legacy generated CV, we get to function `PutExistingCurrentVersion` given local doc has CV we won't enter the code block on line https://sourcegraph.com/github.com/couchbase/sync_gateway/-/blob/db/crud.go?L1357 
- So when checking if documents are in conflict we need to use rev tree history to do so (given all legacy rev generated CV's have the source `Revision+Tree+Encoding` so the HLV's will never be in conflict) 

So essentially we want to keep `revTreeConflictCheck` here but only run if we haven't already done so I think.

db/crud.go
gregns1, line 1419: We also need to avoid this call to `revTreeConflictCheck` here if the check has been done above/rev tree has already been marked as conflict. 

Examples of when we will enter this block if:

- SGW has CV: 1000@CBL2 CurrentRev 2-abc
- CBL pushes: rev: 1010@CBL1, history=2-abc

Even though the HLV's are in conflict, SGWs current rev (2-abc) is the parent of the version CBL is sending.

Or

- SGW doc has no HLV (so gets rev generated one in the code block starting line 1357)
- Incoming rev is legacy rev generated CV
- Underlying rev trees for the documents are in conflict 

Here we will have checked for conflict in the block on 1357, inside the `IsInConflict` function and here in this block.

gregns1 (changes requested):
There are some subtle edge cases around this area involving legacy rev replication. We need to be careful that we don't break these cases but also achieve the outcome of avoiding running `revTreeConflictCheck` multiple times.

db/crud.go
```diff
@@ -1351,18 +1351,25 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
...
+		revTreeConflictCheckStatus := false
+
 		// if local doc is a pre-upgraded mutation and the rev tree is conflicting with incoming, assign this local
 		// version an implicit hlv based on its rev tree ID + this nodes sourceID to allow HLV conflict resolution to occur
 		// between the two versions.
 		if doc.HLV == nil && doc.Cas != 0 && !allowConflictingTombstone {
-			_, _, conflictErr := db.revTreeConflictCheck(ctx, opts.RevTreeHistory, doc, opts.NewDoc.Deleted)
+			revTreeConflictCheck = true
+			var conflictErr error
+			_, _, conflictErr = db.revTreeConflictCheck(ctx, opts.RevTreeHistory, doc, opts.NewDoc.Deleted)
```
gregns1: NIT: Is there a reason for moving the declaration `conflictErr` I think it was fine before using :=? Unless I'm missing something

db/crud.go
```diff
@@ -1351,18 +1351,25 @@ func (db *DatabaseCollectionWithUser) PutExistingCurrentVersion(ctx context.Cont
 		// 2. local doc has no hlv
 		allowConflictingTombstone := opts.ForceAllowConflictingTombstone && doc.IsDeleted()
 
+		// variables to keep track if there was rev tree conflict check and storing the status of the check
+		revTreeConflictCheck := false
```
gregns1: Bit of a nit pick sorry but could we rename this boolean to `revTreeConflictChecked` I think it reads a little better.

gregns1 (commented):
Looks generally really good! Just couple of nit picks I feel help with readability and then it should be ready to go.

db/crud.go
RIT3shSapata: oh yes, I had defined it to be reused elsewhere, but then removed it later, forgot to remove this, will fix it

gregns1 (approved):

PR #7868, opened 2025-11-08
Bad History panic example.

torcolvin:
Passing a rev message with:

- Rev: 1-abc
- History: 3-def

Will result in a panic when trying to pull the message.

```
2025-11-08T09:57:23.908-05:00 [WRN] c:[67613fc5] db:db col:sg_test_0 PANIC handling 'changes' response: runtime error: slice bounds out of range [1:0]
goroutine 135 [running]:
runtime/debug.Stack()
	/Users/torcolvin/sdk/go1.25.1/src/runtime/debug/stack.go:26 +0x64
github.com/couchbase/sync_gateway/db.(*BlipSyncContext).handleChangesResponse.func1()
	/Users/torcolvin/repos/sync_gateway/db/blip_sync_context.go:296 +0x90
panic({0x105a7a500?, 0x14000534150?})
	/Users/torcolvin/sdk/go1.25.1/src/runtime/panic.go:783 +0x120
github.com/couchbase/sync_gateway/db.toHistory(0x105ab64e0?, 0x1400030d0e0, 0x0)
	/Users/torcolvin/repos/sync_gateway/db/blip_sync_context.go:808 +0x118
github.com/couchbase/sync_gateway/db.(*BlipSyncContext).sendRevision(0x140001a7560, {0x105b89078, 0x1400037c3c0}, 0x14000480300, {0x14000328d78, 0x4}, {0x14000328d90, 0x5}, {0x140007ccb58?, 0x104411db8?, ...}, ...)
	/Users/torcolvin/repos/sync_gateway/db/blip_sync_context.go:759 +0xcb4
github.com/couchbase/sync_gateway/db.(*BlipSyncContext).handleChangesResponse(0x140001a7560, {0x105b89078, 0x1400037c3c0}, 0x14000480300, 0x140000eb050, {0x140001b5308, 0x1, 0xc8?}, {0xc23bf474f61e4368?, 0x16aff6a?, ...}, ...)
	/Users/torcolvin/repos/sync_gateway/db/blip_sync_context.go:394 +0xa00
github.com/couchbase/sync_gateway/db.(*blipHandler).sendBatchOfChanges.func1(...)
	/Users/torcolvin/repos/sync_gateway/db/blip_handler.go:629
created by github.com/couchbase/sync_gateway/db.(*blipHandler).sendBatchOfChanges in goroutine 182
	/Users/torcolvin/repos/sync_gateway/db/blip_handler.go:628 +0x350
[...]

PR #7869, opened 2025-11-08
Bad History panic example

torcolvin:
Counterpart of https://github.com/couchbase/sync_gateway/pull/7868 against 3.3


torcolvin:
This is a proof of concept test for https://jira.issues.couchbase.com/browse/CBG-4995, since we aren't planning on fixing the code right away, closing this ticket. This test can be used for reference or implementation at that time.

PR #7870, opened 2025-11-10
CBG-4714: collect goroutine stack traces on signal

gregns1:
CBG-4714

- Collects a stack trace for all running goroutines for signal SIGUSR1 
- Endpoint via REST API to collect stack trace for all running goroutines 
- sgcollect changes to collect these files
- sgcollect will call endpoint to collect stack trace when running

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/159/


rest/server_context.go
```diff
@@ -1844,6 +1873,38 @@ func (sc *ServerContext) logStats(ctx context.Context) error {
 
 }
 
+func (sc *ServerContext) logStackTraces(ctx context.Context, timestamp string) {
+
+	base.InfofCtx(ctx, base.KeyAll, "Collecting stack trace for all goroutines")
+	stackTrace := base.GetStackTrace()
+
+	// log to console
+	_, _ = fmt.Fprintf(os.Stderr, "Stack trace:\n%s\n", stackTrace)
```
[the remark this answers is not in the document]
gregns1: I think we want this logging direct to Stderr for console, we will be writing files for collection by sgcollect

rest/api.go
```diff
@@ -693,6 +693,14 @@ func (h *handler) handleFgprof() error {
 	return stopFn()
 }
 
+func (h *handler) handleCollectStackTrace() error {
+
+	stackTrace := base.GetStackTrace()
```
torcolvin: Is there a reason not to use:

pprof.Lookup("goroutine").WriteTo(os.Stdout, 1)

tools/sgcollect.py
```diff
@@ -279,6 +279,25 @@ def make_http_client_pprof_tasks(
     return pprof_tasks
 
 
+def make_http_client_stack_trace_task(
+    sg_url: str, auth_headers: dict[str, str]
+) -> PythonTask:
+    """
+    This task uses the python http client to collect the raw stack trace data
+    """
+    stack_trace_url = "{0}/_debug/stacktrace".format(sg_url)
```
torcolvin: I think we can just change to `/_debug/pprof/goroutine?debug=2` which will collect all the goroutines in plain text format https://github.com/couchbase/sync_gateway/blob/618bb6aa5d9d9549908d2d99c52b178634af5136/tools/sgcollect.py#L260

I'd be tempted to keep the format as is (no debug=) which works for go tool pprof

and add output in plain text. I believe that /_debug/pprof/goroutine?debug=2 calls pprof.Lookup("goroutine") and won't condense matching stack traces, but you should verify that.

rest/stack_trace_handler_uinx.go
```diff
@@ -0,0 +1,54 @@
+//go:build !windows
```
torcolvin: change filename spelling for unix, if you end with `_unix.go` it implies the build tags

rest/stack_trace_handler_uinx.go
```diff
@@ -0,0 +1,54 @@
...
+	go func() {
+		select {
+		case sig := <-signalChannel:
+			base.InfofCtx(ctx, base.KeyAll, "Handling signal: %v", sig)
+			switch sig {
+			case syscall.SIGUSR1:
+				// stack trace signal received
+				currentTime := time.Now()
+				timestamp := currentTime.Format(time.RFC3339)
+				sc.logStackTraces(ctx, timestamp)
```
torcolvin: I think this might be good to log that we are logging the stack trace to stderr, and also log the stack stack trace with the traditional InfofCtx logging so that it gets picked up if someone isn't grabbing stderr output.

rest/stack_trace_handler_uinx.go
```diff
@@ -0,0 +1,54 @@
...
+	"context"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/couchbase/sync_gateway/base"
+)
+
+// registerSignalHandlerForStackTrace will register a signal handler to capture stack traces
```
torcolvin: Should https://github.com/couchbase/sync_gateway/blob/618bb6aa5d9d9549908d2d99c52b178634af5136/rest/config.go#L2324 be put into the same area / function?

I didn't investigate this but it looks generic-ish.

rest/api.go
gregns1: Happy to use this method but will probably have to call it twice to put once to console and then to file.

rest/stack_trace_handler_uinx.go
gregns1: I mentioned this in the description, I need access to the server context to get the log file path to wrote a file to it, couldn't really think of a nice way to do that here. 

I guess its possible I can move RegisterSignalHandler to after parseFlags in serverMain?

rest/stack_trace_handler_uinx.go
gregns1: I have logging to indicate that a stack trace is being collected but feel logging at stderr and Info level and writing to the file is over kill. I have left out logging to info level given we write this stuff in a file for sgcollect to collect up anyway.

rest/stack_trace_handler_uinx.go
gregns1: This was typo, fixed

base/util.go
```diff
@@ -1831,3 +1834,60 @@ func IsRevTreeID(s string) bool {
...
+// GetStackTrace will return goroutine stack traces for all goroutines in Sync Gateway.
+func GetStackTrace() (bytes.Buffer, error) {
+	profBuf := bytes.Buffer{}
+	err := pprof.Lookup("goroutine").WriteTo(&profBuf, 2)
+	return profBuf, err
+}
+
+// RotateProfilesIfNeeded will remove old files if there are more than
+// 10 matching the given filename pattern.
+func RotateProfilesIfNeeded(filename string) error {
```
torcolvin: rename this to RotateFilenamesIfNeeded

base/util.go
```diff
@@ -1831,3 +1834,60 @@ func IsRevTreeID(s string) bool {
...
+	filename := filepath.Join(logDirectory, StackFilePrefix+timestamp+".log")
+	file, err := os.Create(filename)
+	defer func() {
+		closeErr := file.Close()
+		if closeErr != nil {
+			WarnfCtx(ctx, "Error closing stack trace file %s: %v", filename, closeErr)
+		}
+	}()
+	if err != nil {
+		WarnfCtx(ctx, "Error opening stack trace file %s: %v", filename, err)
```
torcolvin, line 1893: return here

base/util.go
```diff
@@ -1831,3 +1834,60 @@ func IsRevTreeID(s string) bool {
...
+	if err != nil {
+		WarnfCtx(ctx, "Error writing stack trace to file %s: %v", filename, err)
+	}
+
+	rotatePath := filepath.Join(logDirectory, StackFilePrefix+"*.log")
+	err = RotateProfilesIfNeeded(rotatePath)
+	if err != nil {
+		WarnfCtx(ctx, "Error rotating stack trace files in path %s: %v", rotatePath, err)
+	}
+}
```
torcolvin, line 1903: I don't think this is ever going to rotate the file you most recently opened but linux/mac can modify/delete files that are open but windows can not.

I think to make this more fool proof, create private function that writes the stack trace file and then closes it with a defer. Then call the file rotation. That would also address the problem needing to return if you can't open the file

tools/sgcollect.py
```diff
@@ -279,6 +279,25 @@ def make_http_client_pprof_tasks(
...
+    """
+    This task uses the python http client to collect the raw stack trace data
+    """
+    stack_trace_url = "{0}/_debug/pprof/goroutine?debug=2".format(sg_url)
+
+    stack_trace_task = make_curl_task(
+        name="Collect stack trace via http client",
+        auth_headers=auth_headers,
+        url=stack_trace_url,
+        log_file="sg_stack_trace.log",
```
torcolvin:
```suggestion
        log_file="goroutines.log",
```

I don't care what we call this but sg_stack_trace.log will mean that this output overwrites the output from the signal handler, so it needs to be a different name.

Basically this is just a plain text representation of pprof_goroutines.pb.gz so I think that would be a reasonable name.

gregns1: I will alter the name for all the files to make it goroutine specific but this doesn't overwrite the signal files, the signal files will have timestamp in the file name, sgcollect http call will not.

torcolvin: Oh, that's fine then.

rest/register_handler_unix.go
```diff
@@ -0,0 +1,60 @@
+//go:build !windows
+// +build !windows
```
torcolvin, line 2: I think this line is actually unnecessary and will be flagged by govet in go 1.25, but not sure, we can leave this.

tools/sgcollect.py
```diff
@@ -367,6 +386,7 @@ def make_collect_logs_tasks(
         "sg_debug.log": "sg_debug.log",
         "sg_trace.log": "sg_trace.log",
         "sg_stats.log": "sg_stats.log",
+        "sg_goroutines.log": "sg_goroutines.log",
```
torcolvin: I think now that I understand that one log will be the stacktraces at a particular time (from SIGUSR1) and one will be the stack traces at sgcollect I could go back around to stack traces.

But basically this part of the python code is how it finds files from disk directly and the above part of the python code with `make_curl_task` is how it makes requests against Sync Gateway.

torcolvin (dismissed):

tools/sgcollect.py
torcolvin: I would go back to sg_stack_trace now that I understand this, but I feel bad about making you change the names again.

rest/register_handler_unix.go
gregns1: I use go 1.25 locally and my linter didn't pick that up so should be okay fro now (I think)

torcolvin (approved):

PR #7873, opened 2025-11-11
CBG-4747: remove serialisation on critical path

gregns1:
CBG-4747

- Remove serialisation on channel ID struct, pay for it at logging time
- This is to get some perf results form Dragos on how this looks

## Pre-review checklist

- [ ] Link upstream PRs

### DIFF
--- a/channels/set.go
+++ b/channels/set.go
@@ -32,9 +32,8 @@ const AllChannelWildcard = "*"    // wildcard for 'all channels'
 
 // ID represents a single channel inside a collection
 type ID struct {
-	Name          string // name of channel
-	CollectionID  uint32 // collection it belongs to
-	serialization string // private method for logging and matching inside changeWaiter notification
+	Name         string // name of channel
+	CollectionID uint32 // collection it belongs to
 }
 
 func (c ID) String() string {
--- a/channels/set.go
+++ b/channels/set.go
@@ -38,15 +38,14 @@ type ID struct {
 }
 
 func (c ID) String() string {
-	return c.serialization
+	return strconv.FormatUint(uint64(c.CollectionID), 10) + "." + base.UserDataPrefix + c.Name + base.UserDataSuffix
 }
 
 // NewID returns a new ChannelID
 func NewID(channelName string, collectionID uint32) ID {
 	return ID{
-		Name:          channelName,
-		CollectionID:  collectionID,
-		serialization: strconv.FormatUint(uint64(collectionID), 10) + "." + base.UserDataPrefix + channelName + base.UserDataSuffix,
+		Name:         channelName,
+		CollectionID: collectionID,
 	}
 }
 


channels/set.go
```diff
@@ -32,21 +32,25 @@ const AllChannelWildcard = "*"    // wildcard for 'all channels'
...
 
 func (c ID) String() string {
-	return c.serialization
+	var sb strings.Builder
+	sb.WriteString(strconv.FormatUint(uint64(c.CollectionID), 10))
+	sb.WriteByte('.')
+	sb.WriteString(base.UserDataPrefix)
+	sb.WriteString(c.Name)
+	sb.WriteString(base.UserDataSuffix)
+	return sb.String()
```
torcolvin: I'd be tempted by a sync.Once here as well, there's a bit more storage here at the expense of speed.

sync.Once will allocate 1 bool, 1 sync.Mutex to run

It isn't just in logging that the channel id is used, we use this for lookup when you call `channelCacheImpl.channelCaches.Get`

I think testing is the way to find this, but I'd be surprised if this didn't perform worse in your tests, and if it doesn't perform worse then this probably isn't in a critical path.

One thing I'd do is turn off all logging in your test environment and see when the string function is actually called to see where it is used.

The reason this serialization format exists is when the channel cache was extended to be collection aware a channel `A` had to now be distinguished in different collections. We wanted to share a single cache so that the cache size wouldn't ballon per collection, but the lookup keys to the cache are all strings, thus the serialization.

Ultimately the question is whether there actually ever is a case where String() isn't called, because I'm not sure there would be.

gregns1: Yeah the channel ID struct is used to key channel caches for lookup but if we remove the serialisation this will be removed from the keys. I am sure that channel name and collection id on the struct is enough to keep each channel cache distinct/unique.

torcolvin: From some discussion offline, I think that if we are constructing a `channels.ID` then `channels.ID.String()` is always going to be called, notably in the `Notify` function. https://github.com/couchbase/sync_gateway/blob/9722cde508f7b95b9f2bb0d81ef187de912d5a62/db/change_listener.go#L254 It could be possible that there are places this is called without `Notify`, I did not look exhaustively.

The reason that it is serialized to a string is that the change waiters use a string based notification system which was written before collections. To add collections support, this string based serialization of collection ID is only needed for the change waiters https://github.com/couchbase/sync_gateway/pull/5815#discussion_r995164034

For the case of logging, we actually could have logging not log the serialization. I think in all cases for logging we have (or should have) a collection ID https://github.com/couchbase/sync_gateway/blob/9722cde508f7b95b9f2bb0d81ef187de912d5a62/db/change_cache.go#L364 . 

Any or all of these options can be used:

- We could use `String()` for logging and only log the channel name without doing an int to ascii conversion. This might even be preferable anyway for log readbility reasons since 1:chanName doesn't have meaning since `1` represents the collection ID which isn't easy to translate back to a real collection name. We have/should have the collection name on the log context.
- Continue to use serialization 1:chanName for the change waiter, and change the types for the changeWaiter to be something like `channelKey` which is a string (but you could imagine potentially changing to a struct later).
- Keep the serialization but do it under sync.Once so it is delayed until needed.
- Make sure that the construction of channel serialization is done outside holding any mutexes (maybe delay until `Notify`.

I'm also OK pushing changes, watching what happens in a perf run, then keeping/reverting the change.

PR #7875, opened 2025-11-12
CBG-3689: CheckProposedRev does not unmarshall the entire history

RIT3shSapata:
CBG-3689

Describe your PR here...
- `CheckProposedRev` does not unmarshall the entire history to check if the current version is deleted or not
- Added a new level for unmarshal that will unmarshall only the current version and the flags
- Added unit test to test CheckProposedRev

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/162/

### DIFF
--- a/db/crud_test.go
+++ b/db/crud_test.go
@@ -2282,3 +2282,97 @@ func TestSyncDataCVEqual(t *testing.T) {
 		})
 	}
 }
+
+func TestProposedRev(t *testing.T) {
+
+	base.SetUpTestLogging(t, base.LevelDebug, base.KeyAll)
+
+	db, ctx := setupTestDB(t)
+	defer db.Close(ctx)
+	collection, ctx := GetSingleDatabaseCollectionWithUser(ctx, t, db)
+
+	// create 3 documents
+	body := Body{"key1": "value1", "key2": 1234}
+	_, doc1, err := collection.Put(ctx, "doc1", body)
+	require.NoError(t, err)
+	doc1Rev := doc1.GetRevTreeID()
+
+	_, doc2, err := collection.Put(ctx, "doc2", body)
+	require.NoError(t, err)
+	doc2Rev1 := doc2.GetRevTreeID()
+	_, doc2, err = collection.Put(ctx, "doc2", Body{"_rev": doc2Rev1, "key1": "value2", "key2": 5678})
+	require.NoError(t, err)
+	doc2Rev2 := doc2.GetRevTreeID()
+
+	_, doc3, err := collection.Put(ctx, "doc3", body)
+	require.NoError(t, err)
+	doc3Rev1 := doc3.GetRevTreeID()
+	_, doc3, err = collection.Put(ctx, "doc3", Body{"_rev": doc3Rev1, "_deleted": true})
+	require.NoError(t, err)
+
+	testCases := []struct {
+		name           string
+		revID          string
+		parentRevID    string
+		expectedStatus ProposedRevStatus
+		currentRev     string
+		docID          string
+	}{
+		{
+			name:           "no existing document",
+			revID:          "1-abc",
+			parentRevID:    "",
+			expectedStatus: ProposedRev_OK_IsNew,
+			currentRev:     "",
+			docID:          "doc",
+		},
+		{
+			name:           "existing revision with no previous version",
+			revID:          doc1Rev,
+			parentRevID:    "",
+			expectedStatus: ProposedRev_Exists,
+			currentRev:     "",
+			docID:          "doc1",
+		},
+		{
+			name:           "existing revision with previous version",
+			revID:          doc2Rev2,
+			parentRevID:    doc2Rev1,
+			expectedStatus: ProposedRev_Exists,
+			currentRev:     "",
+			docID:          "doc2",
+		},
+		{
+			name:           "new revision with previous version",
+			revID:          "2-abc",
+			parentRevID:    doc1Rev,
+			expectedStatus: ProposedRev_OK,
+			currentRev:     "",
+			docID:          "doc1",
+		},
+		{
+			name:           "new revision with previous revision as tombstone",
+			revID:          "1-abc",
+			parentRevID:    "",
+			expectedStatus: ProposedRev_OK,
+			currentRev:     "",
+			docID:          "doc3",
+		},
+		{
+			name:           "conflicting revision with previous version",
+			revID:          "2-abc",
+			parentRevID:    doc2Rev1,
+			expectedStatus: ProposedRev_Conflict,
+			currentRev:     doc2Rev2,
+			docID:          "doc2",
+		},
+	}
+
+	for _, tc := range testCases {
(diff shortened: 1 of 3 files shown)


db/crud_test.go
```diff
@@ -2282,3 +2282,97 @@ func TestSyncDataCVEqual(t *testing.T) {
 		})
 	}
 }
+
+func TestProposedRev(t *testing.T) {
+
+	base.SetUpTestLogging(t, base.LevelDebug, base.KeyAll)
```
torcolvin: nitpicks:

I think this line is useful for debugging, but LevelDebug with KeyAll displays a lot of logging for rosmar that isn't helpful. When committing this, I think it's fine for this test to remove all the logging.

db/crud_test.go
```diff
@@ -2282,3 +2282,97 @@ func TestSyncDataCVEqual(t *testing.T) {
...
+
+	db, ctx := setupTestDB(t)
+	defer db.Close(ctx)
+	collection, ctx := GetSingleDatabaseCollectionWithUser(ctx, t, db)
+
+	// create 3 documents
```
torcolvin, line 2291: Nit:

To make this easier to read and debug in the future I think I would do something like name these documents with what they are:

const (
    singleRevDoc = "singleRevDoc"
    twoRevDoc = "twoRevDoc"
    tombstonedDoc = "tombstonedDoc"
)

db/crud_test.go
```diff
@@ -2282,3 +2282,97 @@ func TestSyncDataCVEqual(t *testing.T) {
...
+	doc3Rev1 := doc3.GetRevTreeID()
+	_, _, err = collection.Put(ctx, "doc3", Body{"_rev": doc3Rev1, "_deleted": true})
+	require.NoError(t, err)
+
+	testCases := []struct {
+		name           string
+		revID          string
+		parentRevID    string
+		expectedStatus ProposedRevStatus
+		currentRev     string
```
torcolvin:
```suggestion
		expectedCurrentRev     string
```

This indicates it is the return value of the struct

db/crud_test.go
```diff
@@ -2282,3 +2282,97 @@ func TestSyncDataCVEqual(t *testing.T) {
...
+		docID          string
+	}{
+		{
+			name:           "no existing document",
```
torcolvin: General comment, but I think this is worth addressing for this PR.

When writing a test, you want to think about the way the code is called, not just the implementation. Some of the idea of testing is to cover the cases that don't have different code pathways today but do have different code pathways in the future.

So there are several to account for and in a fast test it would be great to hit all of the possibilities

- type of document: (no document, multiple rev, tombstone)
- arguments (revID, parentRevID)
- revID possibilities:
	- revID doesn't exist
    - revID does exist and is current rev
    - revID does exist and is in history
- parentRevID
    - parentRevID doesn't exist
    - parentRevID does exist, and is the immediate parent
    - parentRevID does exist in the history but _isn't_ the immediate parent. (In this case, you can test this by having the documents have three revisions so you can test for a past parent)

I'm not going to list all these permutations, but given how fast this test runs I think it would be fine to hit most of them. I think this is 3 * 3 * 3 which is a manageable number. If it was too many, then I think consider skip more, but this gives more coverage to this function. If this does really blow up because I did the math wrong, we definitely can consider shortening the number of cases.

the first example that strikes me is what happens when you have:

- no existing document
- rev: 2-def
- parentRev: 1-def

That's a common scenario where Couchbase Lite is pushing a document that has two local revisions and Sync Gateway has none.

db/document.go
```diff
@@ -167,6 +168,14 @@ func (sd *SyncData) SetCV(hlv *HybridLogicalVector) {
 	sd.RevAndVersion.CurrentVersion = string(base.Uint64CASToLittleEndianHex(hlv.Version))
 }
 
+func (sd *SyncData) hasFlag(flag uint8) bool {
+	return sd.Flags&flag != 0
+}
+
+func (sd *SyncData) IsDeleted() bool {
```
torcolvin, line 177: I think get in the habit of writing docstrings, especially for public functions.


```suggestion
// IsDeleted returns true if the document metadata indicates it is a tombstone.
func (sd *SyncData) IsDeleted() bool {
```

db/document.go
```diff
@@ -167,6 +168,14 @@ func (sd *SyncData) SetCV(hlv *HybridLogicalVector) {
 	sd.RevAndVersion.CurrentVersion = string(base.Uint64CASToLittleEndianHex(hlv.Version))
 }
 
+func (sd *SyncData) hasFlag(flag uint8) bool {
```
torcolvin, line 172:
```suggestion
// hasFlag returns true if the document has this bit flag.
func (sd *SyncData) hasFlag(flag uint8) bool {
```

db/document.go
```diff
@@ -306,6 +315,11 @@ type revOnlySyncData struct {
 	CurrentRev channels.RevAndVersion `json:"rev"`
 }
 
+type revAndFlagsSyncData struct {
```
torcolvin, line 320:
```suggestion
// revAndFlagsSyncData is a limited set of the SyncData suitable for faster unmarshalling. It only contains rev and flags properties
type revAndFlagsSyncData struct {
```

db/document.go
```diff
@@ -1336,6 +1342,23 @@ func (doc *Document) UnmarshalWithXattrs(ctx context.Context, data, syncXattrDat
 			doc.SyncData = SyncData{}
 		}
 		doc._rawBody = data
+	case DocUnmarshalRevAndFlags:
+		// Unmarshal rev, cas and flags from sync metadata
+		if syncXattrData != nil {
+			var revOnlyMeta revAndFlagsSyncData
+			unmarshalErr := base.JSONUnmarshal(syncXattrData, &revOnlyMeta)
+			if unmarshalErr != nil {
+				return pkgerrors.WithStack(base.RedactErrorf("Failed to UnmarshalWithXattrs() doc with id: %s (DocUnmarshalRevAndFlags).  Error: %v", base.UD(doc.ID), unmarshalErr))
```
torcolvin, line 1353: nit: general go style

pkgerrors WithStack was used before go implemented its own error wrapping in go 1.13. Some of Sync Gateway's code base was written before this existed. Preferred way of doing this is to use `%w` on base.RedactErrorf or fmt.Errorf to use "native" go error wrapping.


```suggestion
				return base.RedactErrorf("Failed to UnmarshalWithXattrs() doc with id: %s (DocUnmarshalRevAndFlags).  Error: %w", base.UD(doc.ID), unmarshalErr)
```

torcolvin (commented):
LGTM in general, I have left some more comments to pick up on go, Sync Gateway style, and the level of testing I prefer. Some of this is idealism but I'd rather note it here when it is easy to achieve and we can discuss how much to do on future tickets when this is harder to achieve.

torcolvin (approved):

PR #7879, opened 2025-11-13
CBG-4249: Create an abstract DCPClient to be able to work with rosmar

torcolvin:
Create an abstract DCPClient to be able to work with rosmar

- Enable DCP to work with attachment migration and resync. attachment compaction since this uses hierarchical paths in rosmar for xattr subdoc operations https://jira.issues.couchbase.com/browse/CBG-4232
- Remove `GocbV2Bucket.StartDCPFeed` usage since `DCPClientOptions` has more features than `sgbucket.FeedArguments`. `sgbucket.Bucket.StartDCPFeed` is needed to call into rosmar. A future ticket could try to merge these two options. For use in Sync Gateway, use `base.StartDCPFeed` or `base.DCPClient`
- Moves guts of `StartDCPFeed` to `StartGocbDCPFeed`
- Fixes a few data race conditions due to fast DCP (`DatabaseContext.RequireAttachmentMigration` and `ResyncManagerDCP.VBUUIDs`)
- remove `assert.Len(t, dbCtx.RequireAttachmentMigration, 1)` assertions when attachment migration runs quickly. Exposes an existing bug `CBG-5038` where Attachment Migration changes syncinfo. Skip those tests under rosmar

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/476


base/dcp_sharded.go
```diff
@@ -64,14 +64,18 @@
 // StartShardedDCPFeed initializes and starts a CBGT Manager targeting the provided bucket.
 // dbName is used to define a unique path name for local file storage of pindex files
 func StartShardedDCPFeed(ctx context.Context, dbName string, configGroup string, uuid string, heartbeater Heartbeater, bucket Bucket, spec BucketSpec, scope string, collections []string, numPartitions uint16, cfg cbgt.Cfg) (*CbgtContext, error) {
+	fmt.Printf("cfg=%+v\n", cfg)
```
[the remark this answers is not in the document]
torcolvin: fxied

base/dcp_client.go
```diff
@@ -106,21 +105,21 @@ func newDCPClientWithForBuckets(ctx context.Context, ID string, callback sgbucke
...
+		spec:                       bucket.GetSpec(),
+		supportsCollections:        bucket.IsSupported(sgbucket.BucketStoreFeatureCollections),
+		terminator:                 make(chan bool),
+		doneChannel:                make(chan error, 1),
+		failOnRollback:             options.FailOnRollback,
+		checkpointPrefix:           options.CheckpointPrefix,
+		dbStats:                    options.DbStats,
+		agentPriority:              options.AgentPriority,
+		collectionIDs:              options.CollectionIDs,
+		checkpointPersistFrequency: options.CheckpointPersistFrequency,
```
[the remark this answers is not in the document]
torcolvin: This isn't right.

db/util_testing.go
```diff
@@ -187,26 +186,19 @@ func purgeWithDCPFeed(ctx context.Context, bucket base.Bucket, tbp *base.TestBuc
...
-			collections[collection.GetCollectionID()] = collection
-		}
-	}
-
-	dcpClientOpts := base.DCPClientOptions{
-		OneShot:           true,
-		FailOnRollback:    false,
-		CollectionIDs:     slices.Collect(maps.Keys(collections)),
-		MetadataStoreType: base.DCPMetadataStoreInMemory,
+		collectionNames.Add(dataStoreName)
```
bbrks, line 197: Since Add takes a variadic and Add itself is doing a reasonable amount of work to handle the de-dupe, it might make more sense to push this down outside of the loop?

base/collection_common.go
```diff
@@ -45,3 +49,16 @@ func (s ScopeAndCollectionNames) ScopeAndCollectionNames() []string {
 func FullyQualifiedCollectionName(bucketName, scopeName, collectionName string) string {
 	return bucketName + "." + scopeName + "." + collectionName
 }
+
+// Add adds any collections to the collections. Any duplicates will be ignored.
+func (c CollectionNames) Add(ds ...sgbucket.DataStoreName) {
+	for _, d := range ds {
+		if _, ok := c[d.ScopeName()]; !ok {
+			c[d.ScopeName()] = []string{}
+		} else if slices.Contains(c[d.ScopeName()], d.CollectionName()) {
```
bbrks: I don't really like the `slices.Contains` inside the `ds` iteration - especially thinking about in the future when we're considering increasing the collection limits.

It kind of feels like this would be better suited as a `map[string]map[string]struct{}` or similar to avoid the slice iterations at insertion time, and handle dedupe.

db/background_mgr_resync_dcp.go
```diff
@@ -406,22 +390,28 @@ func initializePrincipalDocsIndex(ctx context.Context, db *Database) error {
 	return InitializeIndexes(ctx, n1qlStore, options)
 }
 
+func getResyncDCPPrefix(resyncID string) string {
+	return fmt.Sprintf(
+		"sg-%v:resync:%v",
```
bbrks: We should decide on whether CBG-4286 means we should make this prefix `sg:` or one that includes a version. I would be inclined to say `sg:` and then group version numbers in the trailing part of the name.

base/dcp_client.go
```diff
@@ -0,0 +1,111 @@
...
+		ErrorfCtx(ctx, "Finished calling async close error from DCP Feed %q for bucket %q: %v", feedName, MD(bucketName), err)
+		if doneChan != nil {
+			<-doneChan
+		}
+		return nil, err
+	}
+	InfofCtx(ctx, KeyDCP, "Started DCP Feed %q for bucket %q", feedName, MD(bucketName))
+	go func() {
+		select {
+		case err := <-doneChan:
```
adamcfraser, line 165: I thought that for a one-shot feed, a closed doneChan wasn't an error case.

base/gocb_dcp_client.go
```diff
@@ -273,14 +273,13 @@ func (dc *DCPClient) Start() (doneChan chan error, err error) {
 	return dc.doneChannel, nil
 }
 
-// Close is used externally to stop the DCP client. If the client was already closed due to error, returns that error
-func (dc *DCPClient) Close() error {
+// Close is used externally to stop the DCP client.
+func (dc *GoCBDCPClient) Close() {
 	dc.close()
-	return dc.getCloseError()
```
adamcfraser, line 305: Do we not care about reporting any potential close error, or is that being done elsewhere?

base/gocb_dcp_client.go
```diff
@@ -40,17 +40,17 @@ type endStreamCallbackFunc func(e endStreamEvent)
 
 var ErrVbUUIDMismatch = errors.New("VbUUID mismatch when failOnRollback set")
 
-type DCPClient struct {
+type GoCBDCPClient struct {
```
adamcfraser, line 43: Nitpick, take it or leave it, but I think GocbDCPClient and GocbDCPClientOptions are more readable.

db/database.go
```diff
@@ -2573,11 +2572,19 @@ func (db *DatabaseContext) GetCollectionIDs() []uint32 {
 }
 
 // PurgeDCPCheckpoints will purge all DCP metadata from previous run in the bucket, used to reset dcp client to 0
-func PurgeDCPCheckpoints(ctx context.Context, database *DatabaseContext, checkpointPrefix string, taskID string) error {
+func PurgeDCPCheckpoints(ctx context.Context, database *DatabaseContext, checkpointPrefix string, feedPrefix string) error {
 
 	bucket, err := base.AsGocbV2Bucket(database.Bucket)
 	if err != nil {
-		return err
+		checkpoint := checkpointPrefix + ":" + feedPrefix
```
adamcfraser, line 2511: Is this Rosmar-specific handling - it seems to require some knowledge of the naming convention for rosmar DCP checkpoints.  Seems like an unexpected place for rosmar-specific code.

adamcfraser (commented):
A few initial comments, haven't done a full review.

bbrks (commented):
Dropping the few comments I had from yesterday's review

base/gocb_dcp_client.go
torcolvin: I walked these changes out for https://jira.issues.couchbase.com/browse/CBG-2234, which we might not do.

db/util_testing.go
torcolvin: Switched to a different constructor.

db/background_mgr_resync_dcp.go
torcolvin: Resolved separately in CBG-4286

base/collection_common.go
torcolvin: Created `CollectionNameSet` to avoid this problem. It was particularly a problem if you specify duplicate collection IDs when opening a gocb dcp stream - the open stream request just returns EOF and is really hard to debug.

db/database.go
torcolvin: I'm going to try to move this separate from this PR.

base/gocb_dcp_feed.go
```diff
@@ -47,123 +45,71 @@ func getHighSeqMetadata(cbstore CouchbaseBucketStore) ([]DCPMetadata, error) {
...
-func StartGocbDCPFeed(ctx context.Context, bucket *GocbV2Bucket, bucketName string, args sgbucket.FeedArguments, callback sgbucket.FeedEventCallbackFunc, dbStats *expvar.Map, metadataStoreType DCPMetadataStoreType, groupID string) error {
+func newGocbDCPClient(ctx context.Context, bucket *GocbV2Bucket, opts DCPClientOptions) (*GoCBDCPClient, error) {
+	var collectionIDs []uint32
 
-	feedName, err := GenerateDcpStreamName(args.ID)
+	cm, err := bucket.GetCollectionManifest()
 	if err != nil {
-		return err
+		return nil, err
 	}
```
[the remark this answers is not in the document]
torcolvin: With Server 7.6 as a requirement, this isn't an issue.

base/dcp_client.go
```diff
@@ -0,0 +1,112 @@
...
+			// client.Close will always show error over doneChan
+			_ = client.Close()
+			dcpCloseErr := <-doneChan
+			if dcpCloseErr != nil {
+				WarnfCtx(ctx, "Error on closing DCP Feed %q for %q: %v", feedName, MD(bucketName), dcpCloseErr)
+			}
+			break
+		}
+	}()
+	return doneChan, err
```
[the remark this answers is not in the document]
torcolvin: This is fine. We don't actually care about the closeError from start. https://jira.issues.couchbase.com/browse/CBG-2234 would fix this.

db/import_listener.go
```diff
@@ -79,16 +77,17 @@ func (il *importListener) StartImportFeed(dbContext *DatabaseContext) (err error
...
+			Terminator:        il.terminator,
+			CheckpointPrefix:  il.checkpointPrefix,
+			MetadataStoreType: base.DCPMetadataStoreCS,
+			CollectionNames:   dbContext.collectionNameSet(),
+			Callback:          il.ProcessFeedEvent,
 		}
 
-		return dbContext.Bucket.StartDCPFeed(il.loggingCtx, feedArgs, il.ProcessFeedEvent, importFeedStatsMap.Map)
+		_, err := base.StartDCPFeed(ctx, dbContext.Bucket, feedArgs)
+		return err
```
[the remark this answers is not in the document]
torcolvin: I am fine to drop this, this is only for CE.

torcolvin:
I'm working through some flaky tests that were probably flaky before but now have faster timing issues. I don't think this will affect the bulk of this review.

base/rosmar_dcp_client.go
```diff
@@ -0,0 +1,86 @@
...
+	return doneChan, nil
+}
+
+// Close the DCP feed. This is a non blocking operation to allow for use in a callback function.
+func (dc *RosmarDCPClient) Close() error {
+	if dc.terminator != nil {
+		close(dc.terminator)
+		dc.terminator = nil
+	}
+	return nil
```
bbrks: Wrap in a `sync.Once` or maybe have a closing bool like GoCBDCPClient? This isn't safe to call concurrently as-is

db/import_listener.go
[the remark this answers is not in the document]
bbrks: Is it fine to drop this for CE? It seems like an undesirable breaking change

db/database.go
```diff
@@ -2550,11 +2552,19 @@ func (db *DatabaseContext) GetCollectionIDs() []uint32 {
...
+		err := database.MetadataStore.Delete(checkpoint)
+		if err != nil && !base.IsDocNotFoundError(err) {
+			return err
+		}
+		if base.IsDocNotFoundError(err) {
+			return nil
+		}
```
bbrks:
```suggestion
```

xdcr/rosmar_xdcr.go
bbrks: Should this channel read be returned now it returns `error`?

db/background_mgr_resync_dcp.go
```diff
@@ -415,14 +416,15 @@ func initializePrincipalDocsIndex(ctx context.Context, db *Database) error {
 // getResyncDCPClientOptions returns the default set of DCPClientOptions suitable for resync. collectionIDs
 // represent Couchbase Server collection IDs. prefix represents the prefixed name of the checkpoint documents
 // used to store DCP checkpoints.
-func getResyncDCPClientOptions(db *DatabaseContext, resyncID string, collectionIDs []uint32) *base.DCPClientOptions {
+func getResyncDCPClientOptions(db *DatabaseContext, resyncID string, collectionNames base.CollectionNameSet, callback sgbucket.FeedEventCallbackFunc) *base.DCPClientOptions {
 	return &base.DCPClientOptions{
```
bbrks: Caller dereferences anyway, and compaction version of this function returns a non-pointer too.

```suggestion
func getResyncDCPClientOptions(db *DatabaseContext, resyncID string, collectionNames base.CollectionNameSet, callback sgbucket.FeedEventCallbackFunc) base.DCPClientOptions {
return base.DCPClientOptions{
```

bbrks (changes requested):
A few nits, question about Close() error value and import stats

db/import_listener.go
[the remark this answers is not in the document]
torcolvin: The only stat that gets logged is expvar only dcp_rollback_count but I agree I should put it back.

db/background_mgr_attachment_compaction.go
```diff
@@ -133,7 +133,10 @@ func (a *AttachmentCompactionManager) Run(ctx context.Context, options map[strin
 		a.SetPhase("mark")
 		worker := func() (shouldRetry bool, err error, value any) {
 			persistClusterStatus()
-			_, a.VBUUIDs, metadataKeyPrefix, err = attachmentCompactMarkPhase(ctx, dataStore, collectionID, database, a.CompactID, terminator, &a.MarkedAttachments)
+			_, dcpClient, err := attachmentCompactMarkPhase(ctx, dataStore, collectionID, database, a.CompactID, terminator, &a.MarkedAttachments)
+			if dcpClient != nil {
+				a.VBUUIDs = base.GetVBUUIDs(dcpClient.GetMetadata())
+			}
 			if err != nil {
 				shouldRetry, err = a.handleAttachmentCompactionRollbackError(ctx, options, dataStore, database, err, MarkPhase, metadataKeyPrefix)
```
bbrks: `metadataKeyPrefix` now always empty here

should probably be using `dcpClient.GetMetadataKeyPrefix()`?

rest/upgradetest/upgrade_registry_test.go
```diff
@@ -185,6 +185,9 @@ func getDbConfigFromLegacyConfig(rt *rest.RestTester) string {
 
 }
 func TestLegacyMetadataID(t *testing.T) {
+	if base.UnitTestUrlIsWalrus() {
+		t.Skip("fix in CBG-5038")
```
bbrks: CBG-5038 is now fixed via #8059 

These skips shouldn't be needed

xdcr/rosmar_xdcr.go
```diff
@@ -225,9 +225,11 @@ func (r *rosmarManager) Start(ctx context.Context) error {
 	r.collectionsLock.Lock()
 	defer r.collectionsLock.Unlock()
 	r.terminator = make(chan bool)
-	r.doneChan = make(chan struct{})
+	//r.doneChan = make(chan struct{})
```
bbrks: nit whilst there's another round to do

```suggestion
```

base/gocb_dcp_feed.go
```diff
@@ -47,123 +45,71 @@ func getHighSeqMetadata(cbstore CouchbaseBucketStore) ([]DCPMetadata, error) {
...
+		OneShot:           opts.OneShot,
+		FailOnRollback:    opts.FailOnRollback,
+		InitialMetadata:   opts.InitialMetadata,
 	}
 
-	if args.Backfill == sgbucket.FeedNoBackfill {
+	if opts.FromLatestSequence {
+		if len(opts.InitialMetadata) > 0 {
+			return nil, fmt.Errorf("DCPClientOptions.InitialMetadata cannot be provided when FromLatestSequence is true")
+		}
```
bbrks: `NewDCPClient` already checks this combination (and some other cases)

I think it's reasonable to leave all the validation up inside `NewDCPClient` - especially since there are no other callers to this specific function that don't go through the wrapper.

base/gocb_dcp_feed.go
```diff
@@ -47,123 +45,71 @@ func getHighSeqMetadata(cbstore CouchbaseBucketStore) ([]DCPMetadata, error) {
...
 		metadata, err := getHighSeqMetadata(bucket)
 		if err != nil {
-			return err
+			return nil, err
 		}
 		options.InitialMetadata = metadata
 	}
 
-	dcpClient, err := NewDCPClient(
+	return NewGocbDCPClient(
```
bbrks: The unexported function calling the exported function of the same name seems a little odd to me.

I am not sure which function is more appropriate for a caller to use (if any) - since they have the same name, almost same parameters and no docstring.

It feels like there are 1 or 2 too many wrapper functions - I think probably because tests want a simpler version via (`NewGocbDCPClient`?)

`NewDCPClientForTest` is another variant sometimes used that seems candidate for cleanup.

db/util_testing.go
```diff
@@ -956,7 +953,7 @@ func WaitForBackgroundManagerHeartbeatDocRemoval(t testing.TB, mgr *BackgroundMa
 // RequireBackgroundManagerState waits for a BackgroundManager to reach a given state or fails test harness.
 func RequireBackgroundManagerState(t testing.TB, mgr *BackgroundManager, expState BackgroundProcessState) BackgroundManagerStatus {
 	waitTime := 10 * time.Second
-	if !base.UnitTestUrlIsWalrus() {
+	if !base.UnitTestUrlIsWalrus() || base.IsRaceDetectorEnabled(t) || os.Getenv("CI") == "" {
```
bbrks: Does this CI check need to be inverted?

```suggestion
	if !base.UnitTestUrlIsWalrus() || base.IsRaceDetectorEnabled(t) || os.Getenv("CI") != "" {
```

bbrks (changes requested):

base/gocb_dcp_feed.go
torcolvin: I think it is fine to remove this, but this is used by cache_perf_tool.

base/dcp_client.go
```diff
@@ -0,0 +1,116 @@
...
```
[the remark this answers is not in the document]
torcolvin: I agree this is not quite right but I think it is out of scope to fix.

db/background_mgr_attachment_compaction.go
```diff
@@ -133,9 +132,12 @@ func (a *AttachmentCompactionManager) Run(ctx context.Context, options map[strin
...
 		worker := func() (shouldRetry bool, err error, value any) {
 			persistClusterStatus()
-			_, a.VBUUIDs, metadataKeyPrefix, err = attachmentCompactMarkPhase(ctx, dataStore, collectionID, database, a.CompactID, terminator, &a.MarkedAttachments)
+			_, dcpClient, err := attachmentCompactMarkPhase(ctx, dataStore, collectionID, database, a.CompactID, terminator, &a.MarkedAttachments)
+			if dcpClient != nil {
+				a.VBUUIDs = base.GetVBUUIDs(dcpClient.GetMetadata())
+			}
 			if err != nil {
-				shouldRetry, err = a.handleAttachmentCompactionRollbackError(ctx, options, dataStore, database, err, MarkPhase, metadataKeyPrefix)
+				shouldRetry, err = a.handleAttachmentCompactionRollbackError(ctx, options, dataStore, database, err, MarkPhase, dcpClient.GetMetadataKeyPrefix())
```
bbrks, line 146: Panic if `attachmentCompactMarkPhase` fails and `dcpClient` is nil (I actually don't know how to avoid this and correctly do rollback - perhaps there is no need and it's OK to just do this rollback when `dcpClient` is not nil - I'll leave for you to figure out.

bbrks (changes requested):
One comment - the others I had were all deferred as follow-up fixes (`doneChan` missing events, less specific/obscure rosmar checkpoint purging)

db/background_mgr_attachment_compaction.go
torcolvin: If there was no rollback error, this is a non retryable error. Thanks for the good catch.

bbrks (approved):

bbrks:
Worth a fresh integration test run before merge?

PR #7878, opened 2025-11-13
CBG-4928 do not log warning for closed blip sender

torcolvin:
CBG-4928 do not log warning for closed blip sender

This should prevent a log message as below when a blip connection is closed and it can not longer send changes. I believe a blip connection could be closed due to normal network disconnections: 

```
[WRN] c:[231d979a] db:db col:col2 Error from bh.handleChangesResponse: use of closed BLIP sender -- db.(*blipHandler).sendBatchOfChanges.func1() at blip_handler.go:614
```

It is OK to not enter fatalErrorCallback https://github.com/couchbase/sync_gateway/blob/b8b689bb534be1670f71808945ebbbc90a3eaecf/db/active_replicator_push.go#L211-L213 since if the sender is closed it will enter https://github.com/couchbase/sync_gateway/blob/main/db/active_replicator.go#L238 which will close and reconnect.

This was hard to unit test, but is observed in real world logs, so pushing this change to avoid unexpected logs. In order to test this, I believe you have to have the sender closed at the right time.

NOTE: this is only one part of CBG-4928 and there is a second part needed to address:

```
[WRN] c:[393a64c4] db:db col:col1 Error retrieving changes for channel "<ud>b2</ud>": Changes feed cancelled while waiting for view lock -- db.(*DatabaseCollectionWithUser).changesFeed.func1() at changes.go:437
[WRN] c:[393a64c4] db:db col:col1 MultiChangesFeed got error reading changes feed: Error while building channel feed -- db.(*DatabaseCollectionWithUser).SimpleMultiChangesFeed.func1() at changes.go:914
```

PR #7880, opened 2025-11-14
CBG-4928: Update log messages on changes feed being cancelled

RIT3shSapata:
CBG-4928

Describe your PR here...
- Currently SGW logs the changes feed cancellation messages at warn level with "error" in the message
- These are not necessarily errors, therefore the message level and messages have been updated appropriately
- The other part of this ticket is implemented in #7878 

## Pre-review checklist

- [ ] Link upstream PRs

### DIFF
--- a/base/error.go
+++ b/base/error.go
@@ -41,7 +41,7 @@ var (
 	ErrNotFound              = &sgError{"Not Found"}
 	ErrUpdateCancel          = &sgError{"Cancel update"}
 	ErrImportCancelledPurged = HTTPErrorf(http.StatusNotFound, "Import Cancelled Due to Purge")
-	ErrChannelFeed           = &sgError{"Failed to build channel feed"}
+	ErrChannelFeed           = &sgError{"Error while building channel feed"}
 	ErrTimeout               = &sgError{"Operation timed out"}
 	ErrPathNotFound          = sgbucket.ErrPathNotFound
 	ErrPathExists            = sgbucket.ErrPathExists
--- a/base/error.go
+++ b/base/error.go
@@ -41,7 +41,7 @@ var (
 	ErrNotFound              = &sgError{"Not Found"}
 	ErrUpdateCancel          = &sgError{"Cancel update"}
 	ErrImportCancelledPurged = HTTPErrorf(http.StatusNotFound, "Import Cancelled Due to Purge")
-	ErrChannelFeed           = &sgError{"Error while building channel feed"}
+	ErrChannelFeed           = &sgError{"Failed to build channel feed"}
 	ErrTimeout               = &sgError{"Operation timed out"}
 	ErrPathNotFound          = sgbucket.ErrPathNotFound
 	ErrPathExists            = sgbucket.ErrPathExists
--- a/db/changes.go
+++ b/db/changes.go
@@ -995,8 +995,10 @@ func (col *DatabaseCollectionWithUser) SimpleMultiChangesFeed(ctx context.Contex
 							feeds[i] = nil
 						} else {
 							// On feed error, send the error and exit changes processing
-							if options.ChangesCtx.Err() != nil {
-								base.WarnfCtx(ctx, "MultiChangesFeed got error reading changes feed: %v", current[i].Err)
+							if current[i].Err == base.ErrChannelFeed {
+								if options.ChangesCtx.Err() == nil {
+									base.WarnfCtx(ctx, "MultiChangesFeed got error reading changes feed: %v", current[i].Err)
+								}
 								select {
 								case <-options.ChangesCtx.Done():
 								case output <- current[i]:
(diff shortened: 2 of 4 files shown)


db/channel_cache_single.go
```diff
@@ -413,7 +413,7 @@ func (c *singleChannelCacheImpl) GetChanges(ctx context.Context, options Changes
 	// Check whether the changes process has been terminated while we waited for the view lock, to avoid the view
 	// overhead in that case (and prevent feedback loop on query backlog)
 	if options.ChangesCtx.Err() != nil {
-		return nil, fmt.Errorf("Changes feed cancelled while waiting for view lock")
+		return nil, fmt.Errorf("Changes feed cancelled")
```
torcolvin:
```suggestion
		return nil, fmt.Errorf("Changes feed cancelled %w", options.ChangesCtx.Err())
```

This would tell us in downstream while it was canceled, particularly if we are able to determine a cause, this will make the above easier to reason about I think.

db/changes.go
```diff
@@ -480,7 +480,7 @@ func (db *DatabaseCollectionWithUser) changesFeed(ctx context.Context, singleCha
 			base.TracefCtx(ctx, base.KeyChanges, "Querying channel %q with options: %+v", base.UD(singleChannelCache.ChannelID().Name), paginationOptions)
 			changes, err := singleChannelCache.GetChanges(ctx, paginationOptions)
 			if err != nil {
-				base.WarnfCtx(ctx, "Error retrieving changes for channel %q: %v", base.UD(singleChannelCache.ChannelID().Name), err)
+				base.InfofCtx(ctx, base.KeyChanges, "Could not retrieve changes for channel %q: %v", base.UD(singleChannelCache.ChannelID().Name), err)
```
torcolvin: Assuming the context cancellation is passed back in the error message:

```
								if !errors.Is(err, context.Canceled) && errors.Is(err, context.DeadlineExceeded) {
									base.WarnfCtx(ctx, "Error retrieving changes for channel %q: %v", base.UD(singleChannelCache.ChannelID().Name), err)
								}
```

However, I think to match the behavior of below if we don't always wrap the context error message:

```suggestion
								if  options.ChangesCtx.Err() != nil {
									base.WarnfCtx(ctx, "Error retrieving changes for channel %q: %v", base.UD(singleChannelCache.ChannelID().Name), err)
								}
```

db/changes.go
```diff
@@ -996,7 +996,7 @@ func (col *DatabaseCollectionWithUser) SimpleMultiChangesFeed(ctx context.Contex
 						} else {
 							// On feed error, send the error and exit changes processing
 							if current[i].Err == base.ErrChannelFeed {
-								base.WarnfCtx(ctx, "MultiChangesFeed got error reading changes feed: %v", current[i].Err)
```
torcolvin, line 999:
```suggestion
								if !options.ChangesCtx.Err() {
									base.WarnfCtx(ctx, "MultiChangesFeed got error reading changes feed: %v", current[i].Err)
								}
```

In this case, it's not clear whether that err from current[i].Err is better than the error returned.

base/error.go
```diff
@@ -41,7 +41,7 @@ var (
 	ErrNotFound              = &sgError{"Not Found"}
 	ErrUpdateCancel          = &sgError{"Cancel update"}
 	ErrImportCancelledPurged = HTTPErrorf(http.StatusNotFound, "Import Cancelled Due to Purge")
-	ErrChannelFeed           = &sgError{"Error while building channel feed"}
+	ErrChannelFeed           = &sgError{"Failed to build channel feed"}
```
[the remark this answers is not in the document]
torcolvin: I think we can drop this, because I think this error wouldn't show up if we avoid logging it at all.

db/channel_cache_single.go
```diff
@@ -413,7 +413,7 @@ func (c *singleChannelCacheImpl) GetChanges(ctx context.Context, options Changes
 	// Check whether the changes process has been terminated while we waited for the view lock, to avoid the view
 	// overhead in that case (and prevent feedback loop on query backlog)
```
torcolvin:
```suggestion
// Check whether the changes process has been terminated before running a query
```

base/error.go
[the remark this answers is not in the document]
RIT3shSapata: I think we cannot remove this, as its being used in multiple other places

db/changes.go
RIT3shSapata: A few tests were not being passed because the Querytimeout would just throw an error and no context will be cancelled

base/error.go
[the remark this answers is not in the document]
torcolvin: I meant, we don't need to change the message. I think "Failure" and "Error" are roughly equivalent in severity and I think there's value to leaving the messages the same to track across releases for supportability.

It is worth changing a message if it downgrades it from an issue that customers/support has to look at, or provides additional clarity.

db/changes.go
```diff
@@ -995,7 +995,7 @@ func (col *DatabaseCollectionWithUser) SimpleMultiChangesFeed(ctx context.Contex
 							feeds[i] = nil
 						} else {
 							// On feed error, send the error and exit changes processing
-							if current[i].Err == base.ErrChannelFeed {
+							if options.ChangesCtx.Err() != nil {
```
torcolvin: We _do_ want to warn when it is an error but we don't have a context cancellation.

We don't want to warn if we have a context cancellation.

torcolvin (approved):

PR #7885, opened 2025-11-17
CBG-4747: remove serialisation from caching process

gregns1:
CBG-4747

- Draft of proposal to move serialisation off the critical path at the cache but also change the change listener to be keyed by ID struct not string to avid serialisation at the change waiter level 
- Supplementary doc to do with this PR -> https://docs.google.com/document/d/1LiUy_jvr3HjIYYB9uY3GPUQ_2CWXr80NdeVi3YJZ5Ns/edit?usp=sharing 

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/165/

### DIFF
--- a/channels/set.go
+++ b/channels/set.go
@@ -36,7 +36,7 @@ type ID struct {
 }
 
 func (c ID) String() string {
-	return base.UserDataPrefix + c.Name + base.UserDataSuffix
+	return c.Name
 }
 
 // NewID returns a new ChannelID
@@ -193,7 +193,7 @@ func (redactorSet RedactorSet) GetRedactionString(shouldRedact bool) string {
 	tmp := []byte("{")
 	iterationCount := 0
 	for setItem, _ := range redactorSet.set {
-		tmp = append(tmp, redactorSet.redactorFunc(setItem).String()...)
+		tmp = append(tmp, redactorSet.redactorFunc(setItem).Redact()...)
 		iterationCount++
 		if iterationCount != len(redactorSet.set) {
 			tmp = append(tmp, ", "...)
--- a/channels/set.go
+++ b/channels/set.go
@@ -35,6 +35,8 @@ type ID struct {
 	CollectionID uint32 // collection it belongs to
 }
 
+// String returns the channel name as a string, omitting the collection ID, so beware when using for logging purposes
+// that the collection name is logged from the context on the log line
 func (c ID) String() string {
 	return c.Name
 }
--- a/channels/set.go
+++ b/channels/set.go
@@ -11,7 +11,6 @@ package channels
 import (
 	"fmt"
 	"sort"
-	"strconv"
 	"strings"
 
 	"github.com/couchbase/sync_gateway/base"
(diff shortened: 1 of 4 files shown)


db/change_listener.go
```diff
@@ -310,9 +310,10 @@ func tickerValForBroadcastSpeed(skippedSequencePresent bool) time.Duration {
 
 // Changes the counter, notifying waiting clients. Only use for a key update.
 func (listener *changeListener) notifyKey(ctx context.Context, key string) {
+	mapKey := channels.NewID(key, 0)
```
torcolvin: This is only right because the callers are always principal documents, I think this should just change to notifyChannel with key channels.ID and then the caller constructs it.

If/when we move principal docs to system collection this will be wrong.

channels/set.go
```diff
@@ -32,21 +31,19 @@ const AllChannelWildcard = "*"    // wildcard for 'all channels'
...
-	Name          string // name of channel
-	CollectionID  uint32 // collection it belongs to
-	serialization string // private method for logging and matching inside changeWaiter notification
+	Name         string // name of channel
+	CollectionID uint32 // collection it belongs to
 }
 
 func (c ID) String() string {
-	return c.serialization
+	return base.UserDataPrefix + c.Name + base.UserDataSuffix
```
torcolvin: It feels weird to me to put redaction here, is this a performance benefit?

Does this log correctly if you do `base.UD(channels.ID{Name: Foo, CollectionID: 1})`

torcolvin (commented):
I like this so much better than the other approaches.

The thing I'd like to look for is to analyze logging to make sure that the collection is logged if we need it (probably by `col:colname` in context log key).


Some random ideas I have:

Look for is `unusedSeqChannelID` which is a sentinel value. I think just make sure the logging for this makes sense.

Another idea I had is whether we actually need to create `channels.Set` at all, and whether notify functions should just take `... channels.ID` variadic arguments. We could actually just deduplicate in one loop to avoid having to construct a set before notify, since it seems like mostly we just deal in []channels.ID now.

channels/set.go
gregns1: I was having issues with `TestSetString` without having the UD tag serialisation here. And without this tagging here this log line https://sourcegraph.com/github.com/couchbase/sync_gateway/-/blob/db/change_listener.go?L256 was logging without the ud tags around the keys.

torcolvin: Does switching from `%q` to `%s` make this better?

db/change_listener.go
```diff
@@ -181,7 +181,8 @@ func (listener *changeListener) ProcessFeedEvent(event sgbucket.FeedEvent) bool
 	docType := listener.DocumentType(event.Key)
 	if docType == DocTypeUser || docType == DocTypeRole {
 		// defer to notify after callback completion
-		defer listener.notifyKey(listener.ctx, string(event.Key))
+		key := channels.NewID(string(event.Key), 0)
```
torcolvin:
```suggestion
		key := channels.NewID(string(event.Key), event.CollectionID)
```

db/change_listener.go
```diff
@@ -505,9 +507,9 @@ func (waiter *ChangeWaiter) RefreshUserKeys(user auth.User, metaKeys *base.Metad
 		if len(waiter.userKeys) == 1 && len(user.RoleNames()) == 0 {
 			return
 		}
-		waiter.userKeys = []string{metaKeys.UserKey(user.Name())}
+		waiter.userKeys = []channels.ID{channels.NewID(metaKeys.UserKey(user.Name()), 0)}
```
torcolvin: This has also hardcoded the metdata id, and we should pull from the event / database / etc.

torcolvin: Although now I see the benefit of always putting zero here if we want to move the principals between collections.

Maybe keep the 0 and use constant like principalDocCollectionChannelID and have it be zero to match existing data? and then just use that constant everywhere

bbrks (approved):
General approach and the implementation both look good to me 👍

PR #7886, opened 2025-11-17
Remove goutils logger

torcolvin:
Remove goutils logger

This is just a proposal. I can see the benefit of leaving this in if cbgt decides to use this for something else. However, they have their own logging utility so this seems unlikely.

I believe this was used only from go-couchbase. This drops a direct dependency to an indirect dependency from cbgt only. It is used by cbgt to access cbauth which Sync Gateway is never going to use. https://github.com/couchbase/cbgt/blob/74dbd1c34db313c2b65a65f6e3cecca7eacb00b3/cbauth.go#L24 / https://github.com/couchbase/cbauth

Inspired from https://github.com/couchbase/sync_gateway/pull/7881

### DIFF
--- a/base/logger_external.go
+++ b/base/logger_external.go
@@ -12,13 +12,10 @@ package base
 
 import (
 	"context"
-	"fmt"
-	"os"
 
 	"github.com/couchbase/clog"
 	"github.com/couchbase/gocb/v2"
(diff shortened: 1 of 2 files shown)


bbrks (approved):
👍 agree with all of this

PR #7889, opened 2025-11-19
CBG-5006: Split ServerContext mutexes

bbrks:
CBG-5006

Have locks for each part of `ServerContext` instead of a shared mutex - avoids R/W deadlocks between HTTP Server list and Databases and reduces scope of each lock.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/170/

### DIFF
--- a/rest/api_test.go
+++ b/rest/api_test.go
@@ -2989,7 +2989,9 @@ func TestCreateDBWithXattrsDisabled(t *testing.T) {
 	resp = rt.CreateDatabase(dbName, dbConfig)
 	RequireStatus(t, resp, http.StatusCreated)
 
-	rt.RestTesterServerContext._dbConfigs[dbName].DatabaseConfig.EnableXattrs = base.Ptr(false)
+	rt.updatePersistedConfig(dbName, func(config *DatabaseConfig) {
+		config.EnableXattrs = base.Ptr(false)
+	})
 
 	_, err := rt.RestTesterServerContext.ReloadDatabase(t.Context(), dbName, false)
 	assert.Error(t, err, errResp)
@@ -3481,8 +3483,10 @@ func TestAllowConflictsConfig(t *testing.T) {
 	require.Equal(t, fmt.Sprintf(`[{"db_name":"%s","bucket":"%s","state":"Online"}]`, rt.GetDatabase().Name, rt.GetDatabase().Bucket.GetName()), resp.Body.String())
 	bucketName := rt.GetDatabase().Bucket.GetName()
 
-	// Attempt to set the `AllowConflicts` property to true in the database configuration.
-	rt.RestTesterServerContext._dbConfigs[dbName].DatabaseConfig.DbConfig.AllowConflicts = base.Ptr(true)
+	// Set the `AllowConflicts` property to true in the persisted database configuration and try to load it (this covers cases where the config value was previously set from an older SGW version that allowed this configuration).
+	rt.updatePersistedConfig(dbName, func(config *DatabaseConfig) {
+		config.DbConfig.AllowConflicts = base.Ptr(true)
+	})
 
 	// Reload the database configuration and verify that an error is returned.
 	_, err := rt.RestTesterServerContext.ReloadDatabase(ctx, dbName, false)
@@ -3525,7 +3529,9 @@ func TestDisableAllowStarChannel(t *testing.T) {
 	RequireStatus(t, resp, http.StatusCreated)
 
 	// Attempting to disable `enable_star_channel`
-	rt.ServerContext()._dbConfigs[dbName].DatabaseConfig.CacheConfig.ChannelCacheConfig.EnableStarChannel = base.Ptr(false)
+	rt.updatePersistedConfig(dbName, func(config *DatabaseConfig) {
+		config.CacheConfig.ChannelCacheConfig.EnableStarChannel = base.Ptr(false)
+	})
 
 	// Reloading the database after updating the config
 	_, err := rt.ServerContext().ReloadDatabase(ctx, dbName, false)
--- a/rest/api_test.go
+++ b/rest/api_test.go
@@ -2989,7 +2989,7 @@ func TestCreateDBWithXattrsDisabled(t *testing.T) {
 	resp = rt.CreateDatabase(dbName, dbConfig)
 	RequireStatus(t, resp, http.StatusCreated)
 
-	rt.RestTesterServerContext.dbConfigs[dbName].DatabaseConfig.EnableXattrs = base.Ptr(false)
+	rt.RestTesterServerContext._dbConfigs[dbName].DatabaseConfig.EnableXattrs = base.Ptr(false)
 
 	_, err := rt.RestTesterServerContext.ReloadDatabase(t.Context(), dbName, false)
 	assert.Error(t, err, errResp)
@@ -3482,7 +3482,7 @@ func TestAllowConflictsConfig(t *testing.T) {
 	bucketName := rt.GetDatabase().Bucket.GetName()
 
 	// Attempt to set the `AllowConflicts` property to true in the database configuration.
-	rt.RestTesterServerContext.dbConfigs[dbName].DatabaseConfig.DbConfig.AllowConflicts = base.Ptr(true)
+	rt.RestTesterServerContext._dbConfigs[dbName].DatabaseConfig.DbConfig.AllowConflicts = base.Ptr(true)
 
 	// Reload the database configuration and verify that an error is returned.
 	_, err := rt.RestTesterServerContext.ReloadDatabase(ctx, dbName, false)
@@ -3525,7 +3525,7 @@ func TestDisableAllowStarChannel(t *testing.T) {
 	RequireStatus(t, resp, http.StatusCreated)
 
 	// Attempting to disable `enable_star_channel`
-	rt.ServerContext().dbConfigs[dbName].DatabaseConfig.CacheConfig.ChannelCacheConfig.EnableStarChannel = base.Ptr(false)
+	rt.ServerContext()._dbConfigs[dbName].DatabaseConfig.CacheConfig.ChannelCacheConfig.EnableStarChannel = base.Ptr(false)
 
 	// Reloading the database after updating the config
 	_, err := rt.ServerContext().ReloadDatabase(ctx, dbName, false)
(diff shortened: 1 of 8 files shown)


rest/server_context.go
```diff
@@ -275,21 +279,23 @@ func (sc *ServerContext) Close(ctx context.Context) {
...
+	for _, db := range sc._databases {
 		db.Close(ctx)
 		_ = db.EventMgr.RaiseDBStateChangeEvent(ctx, db.Name, "offline", "Database context closed", &sc.Config.API.AdminInterface)
 	}
-	sc.databases_ = nil
+	sc._databases = nil
 	sc.invalidDatabaseConfigTracking.dbNames = nil
 
+	sc._httpServersLock.Lock()
+	defer sc._httpServersLock.Unlock()
```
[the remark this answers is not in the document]
bbrks: Isn't an issue in reality but I rearranged the code to make a bit more sense in terms of ordering of Close operations (and guards against the nested lock anyway)

rest/api_test.go
```diff
@@ -2990,7 +2989,7 @@ func TestCreateDBWithXattrsDisabled(t *testing.T) {
 	resp = rt.CreateDatabase(dbName, dbConfig)
 	RequireStatus(t, resp, http.StatusCreated)
```
torcolvin, line 2992: this line should be a lock but actually i'd just change the test so it makes a different REST api call

I don't want to accidentally have the race detector pick up this error because we are doing something naughty in the tests and config polling runs simultaneously.

rest/api_test.go
```diff
@@ -3483,7 +3482,7 @@ func TestAllowConflictsConfig(t *testing.T) {
 	bucketName := rt.GetDatabase().Bucket.GetName()
 
 	// Attempt to set the `AllowConflicts` property to true in the database configuration.
-	rt.RestTesterServerContext.dbConfigs[dbName].DatabaseConfig.DbConfig.AllowConflicts = base.Ptr(true)
+	rt.RestTesterServerContext._dbConfigs[dbName].DatabaseConfig.DbConfig.AllowConflicts = base.Ptr(true)
```
torcolvin: See above about potential race and just use /db/_config API

rest/api_test.go
```diff
@@ -3526,7 +3525,7 @@ func TestDisableAllowStarChannel(t *testing.T) {
 	RequireStatus(t, resp, http.StatusCreated)
 
 	// Attempting to disable `enable_star_channel`
-	rt.ServerContext().dbConfigs[dbName].DatabaseConfig.CacheConfig.ChannelCacheConfig.EnableStarChannel = base.Ptr(false)
+	rt.ServerContext()._dbConfigs[dbName].DatabaseConfig.CacheConfig.ChannelCacheConfig.EnableStarChannel = base.Ptr(false)
```
torcolvin: As above, use REST API

rest/config.go
```diff
@@ -1657,21 +1657,21 @@ func (sc *ServerContext) fetchAndLoadConfigs(ctx context.Context, isInitialStart
...
 		for dbName, fetchedConfig := range fetchedConfigs {
-			if dbConfig, ok := sc.dbConfigs[dbName]; ok && dbConfig.cfgCas >= fetchedConfig.cfgCas {
+			if dbConfig, ok := sc._dbConfigs[dbName]; ok && dbConfig.cfgCas >= fetchedConfig.cfgCas {
 				sc.invalidDatabaseConfigTracking.remove(dbName)
 				base.DebugfCtx(ctx, base.KeyConfig, "Database %q bucket %q config has not changed since last update", fetchedConfig.Name, *fetchedConfig.Bucket)
 				delete(fetchedConfigs, dbName)
 			}
 		}
-		sc.lock.RUnlock()
+		sc._databasesLock.RUnlock()
```
torcolvin, line 1674: I want to file a separate ticket for this, if there was a panic between this line and and the readLock, then I want this to unlock.

If the panic happens in config polling, I believe that this will take down Sync Gateway. However, this can get called from any db scoped REST API handler:

`GetInactiveDatabase` (if serverless) -> `FetchAndLoadDatabaseSince` -> `fetchAndLoadConfigs`

Since this is only serverless, this is low priority.

For the issue that spawned this ticket (see https://jira.issues.couchbase.com/browse/CBG-5010 for the full logs), something locked the serverContext.lock mutex but didn't unlock it. I don't think it's this line but this is one of the only cases where this is called outside a defer.

I think it would be fine to do this as part of a separate defensive ticket.

rest/config.go
```diff
@@ -1895,9 +1895,9 @@ func (sc *ServerContext) _handleInvalidDatabaseConfig(ctx context.Context, bucke
 func (sc *ServerContext) bucketNameFromDbName(ctx context.Context, dbName string) (bucketName string, found bool) {
 	// Minimal representation of config struct to be tolerant of invalid database configurations where we still need to find a database name
 	// see if we find the database in-memory first, otherwise fall back to scanning buckets for db configs
-	sc.lock.RLock()
-	dbc, ok := sc.databases_[dbName]
-	sc.lock.RUnlock()
+	sc._databasesLock.RLock()
+	dbc, ok := sc._databases[dbName]
+	sc._databasesLock.RUnlock()
```
torcolvin, line 1900: It is so unlikely that this would panic, but due to excessive paranoia I am almost tempted to wrap this in a function in case `sc._databases` was nil when this function was called in a case where panics are handled.

Feel free to disregard.

rest/config.go
```diff
@@ -1948,7 +1948,7 @@ func (sc *ServerContext) fetchConfigsSince(ctx context.Context, refreshInterval
 		sc.fetchConfigsLastUpdate = time.Now()
 	}
 
-	return sc.dbConfigs, nil
+	return sc._dbConfigs, nil
```
torcolvin, line 1951: This function is weird because it returns a value that would require you to hold a lock. This behavior should be noted in the docstring and considered.

I didn't look through the implications of this.

rest/config.go
```diff
@@ -2211,8 +2211,8 @@ func StartServer(ctx context.Context, config *StartupConfig, sc *ServerContext)
 }
 
 func sharedBucketDatabaseCheck(ctx context.Context, sc *ServerContext) (errors error) {
```
torcolvin, line 2213: Should we lock this function or rename to _sharedBucketDatabaseCheck and note which lock you are required to hold?

rest/config.go
torcolvin, line 2253: Should we lock this function or rename to _sharedBuckets and note which lock you are required to hold?

rest/server_context.go
```diff
@@ -320,9 +328,9 @@ func (sc *ServerContext) GetDatabase(ctx context.Context, name string) (*db.Data
 // GetActiveDatabase attempts to return the DatabaseContext of a loaded database. If not found, the database name will be
 // validated to make sure it's valid and then an error returned.
 func (sc *ServerContext) GetActiveDatabase(name string) (*db.DatabaseContext, error) {
-	sc.lock.RLock()
-	dbc := sc.databases_[name]
-	sc.lock.RUnlock()
+	sc._databasesLock.RLock()
+	dbc := sc._databases[name]
```
torcolvin, line 332: This could panic if `sc._databases` is nil, which I don't think would happen but might be worth protecting against?

rest/server_context.go
```diff
@@ -352,9 +360,9 @@ func (sc *ServerContext) GetInactiveDatabase(ctx context.Context, name string) (
 			dbConfigFound, _ = sc.fetchAndLoadDatabase(base.NewNonCancelCtx(), name, false)
 		}
 		if dbConfigFound {
-			sc.lock.RLock()
-			defer sc.lock.RUnlock()
-			dbc := sc.databases_[name]
+			sc._databasesLock.RLock()
+			defer sc._databasesLock.RUnlock()
+			dbc := sc._databases[name]
```
torcolvin, line 365: Out of scope, do we want to unlock inside this loop?

torcolvin (commented):
I wrote a lot of comments about paranoia about grabbing the lock but possibly panicking or not letting go of the lock because I believe the issue that spawned this ticket had a ServerContext.lock.Lock() that was not unlocked but no longer existed in the stack trace that we had.

I think that fixing these issues can be deferred to a separate ticket and some cases were really extra paranoid and probably don't need to be addressed.

rest/config.go
bbrks: There is already a ticket to review usage of the lock. This ticket is simply splitting out the http server lock.

I tested this by running `/_ping` and `/db/_config` in a loop to ensure that the db config update does not interfere with the ping response (which it now does not)

Will review with the other larger `sc._databasesLock` changes in CBG-5012

rest/api_test.go
bbrks: This test case isn't quite a regular `/db/_config` operation - it's effectively doing a persistent config load of an old configuration - bypassing the upfront validation that `/db/_config` would've run.

Either way, I can move it to a function to wrap in a read mutex (the read lock is required to read the dbName entry, but nothing more to update the underlying configuration)

rest/api_test.go
bbrks: Nah, as above - this test is ensuring coverage of loading an older pre 4.0 config, where AllowConflicts was allowed to be set. I cannot set this value through the REST API anymore (intentionally so)

I have created a named function to do this to make it more obvious why it's being done this way.

rest/api_test.go
bbrks: ditto

rest/config.go
bbrks: Will review with the other larger `sc._databasesLock` changes in CBG-5012

rest/config.go
bbrks: Yeah I agree this is weird and seems wrong - will review with the other larger `sc._databasesLock` changes in CBG-5012

rest/config.go
bbrks: Will review with the other larger `sc._databasesLock` changes in CBG-5012

rest/config.go
bbrks: Will review with the other larger `sc._databasesLock` changes in CBG-5012

rest/server_context.go
bbrks: Will review with the other larger `sc._databasesLock` changes in CBG-5012

rest/server_context.go
bbrks: Will review with the other larger `sc._databasesLock` changes in CBG-5012

torcolvin (approved):

PR #7891, opened 2025-11-19
CBG-4932 improve ResyncManger test flakes

torcolvin:
- the problem seems to be that some tests take a long time to run, mostly after churn of the bucket pool to do query operations after resync finishes
- clean up to use test functions for waiting

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/175/ (known independent test failure)

### DIFF
--- a/db/util_testing.go
+++ b/db/util_testing.go
@@ -938,8 +938,8 @@ func MoveAttachmentXattrFromGlobalToSync(t *testing.T, dataStore base.DataStore,
 func RequireBackgroundManagerState(t testing.TB, mgr *BackgroundManager, expState BackgroundProcessState) BackgroundManagerStatus {
 	waitTime := 10 * time.Second
 	if !base.UnitTestUrlIsWalrus() {
-		// Increase wait time for CI tests against Couchbase Server with GSI disabled (views), some queries take a
-		// longer time to run
+		// Increase wait time for CI tests against Couchbase Server, they can take longer to run.
+		// Generally everything runs in 10 seconds, but when it does not, it is not worth flagging the failures.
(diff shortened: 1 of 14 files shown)


db/util_testing.go
```diff
@@ -935,14 +935,23 @@ func MoveAttachmentXattrFromGlobalToSync(t *testing.T, dataStore base.DataStore,
 
 // RequireBackgroundManagerState waits for a BackgroundManager to reach a given state within 10 seconds or fails test
 // harness.
-func RequireBackgroundManagerState(t testing.TB, ctx context.Context, mgr *BackgroundManager, expState BackgroundProcessState) {
+func RequireBackgroundManagerState(t testing.TB, mgr *BackgroundManager, expState BackgroundProcessState) BackgroundManagerStatus {
+	waitTime := 10 * time.Second
+	if !base.UnitTestUrlIsWalrus() {
+		// Increase wait time for CI tests against Couchbase Server with GSI disabled (views), some queries take a
+		// longer time to run
```
gregns1: Should there be an additional check here for gsi being enabled given the comment suggests this is views specific?

torcolvin: I think the comment is wrong actually. It takes longer for views, but it just sometimes also takes a long time for Couchbase Server if I've ~angered~ overloaded it. I decided not to investigate a lot because I already did a boring and unfruitful views investigation before I realised I didn't use `SG_TEST_USE_GSI=true`. I think the benefits of waiting longer outweigh potentially not returning early.

gregns1 (approved):

PR #7892, opened 2025-11-19
CBG-3690 don't re-read document for resync

torcolvin:
CBG-3690 don't re-read document for resync

- one fix to create hlv if none exists from https://github.com/couchbase/sync_gateway/pull/7890
- remove non xattr code pathway
- removed redundant args from resyncDocument and unused return param

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/171/ (unrelated flake in TestRemoveIndexesUseViewsTrueAndFalse)

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -1838,7 +1838,6 @@ func (db *DatabaseCollectionWithUser) ResyncDocument(ctx context.Context, docid 
 		if err != nil {
 			return sgbucket.UpdatedDoc{}, err
 		}
-		fmt.Printf("%+v\n", doc)
 		updatedDoc, shouldUpdate, updatedExpiry, _, updatedUnusedSequences, err = db.getResyncedDocument(ctx, doc, regenerateSequences, nil)
 		if err != nil {
 			return sgbucket.UpdatedDoc{}, err
@@ -1854,7 +1853,6 @@ func (db *DatabaseCollectionWithUser) ResyncDocument(ctx context.Context, docid 
 
 		// Update MetadataOnlyUpdate based on previous Cas, MetadataOnlyUpdate
 		doc.MetadataOnlyUpdate = computeMetadataOnlyUpdate(doc.Cas, doc.RevSeqNo, doc.MetadataOnlyUpdate)
-		fmt.Printf("updatedDoc= %+v\n", updatedDoc)
 		_, rawSyncXattr, _, rawMouXattr, rawGlobalXattr, err := updatedDoc.MarshalWithXattrs()
 		updatedDoc := sgbucket.UpdatedDoc{
 			Doc: nil, // Resync does not require document body update
--- a/db/database.go
+++ b/db/database.go
@@ -1824,96 +1824,67 @@ func (db *DatabaseCollectionWithUser) getResyncedDocument(ctx context.Context, d
 
 // ResyncDocument will re-run the sync function on the document and write an updated version to the bucket. If
 // the sync function doesn't change any channels or access grants, no write will be performed.
-func (db *DatabaseCollectionWithUser) ResyncDocument(ctx context.Context, docid, key string, regenerateSequences bool, unusedSequences []uint64) (updatedHighSeq uint64, updatedUnusedSequences []uint64, err error) {
+func (db *DatabaseCollectionWithUser) ResyncDocument(ctx context.Context, docid string, previousDoc *sgbucket.BucketDocument, regenerateSequences bool) (updatedUnusedSequences []uint64, err error) {
 	var updatedDoc *Document
 	var shouldUpdate bool
 	var updatedExpiry *uint32
-	if db.UseXattrs() {
-		writeUpdateFunc := func(currentValue []byte, currentXattrs map[string][]byte, cas uint64) (sgbucket.UpdatedDoc, error) {
-			// There's no scenario where a doc should from non-deleted to deleted during UpdateAllDocChannels processing,
-			// so deleteDoc is always returned as false.
-			if currentValue == nil || len(currentValue) == 0 {
-				return sgbucket.UpdatedDoc{}, base.ErrUpdateCancel
-			}
-			doc, err := db.unmarshalDocumentWithXattrs(ctx, docid, currentValue, currentXattrs, cas, DocUnmarshalAll)
-			if err != nil {
-				return sgbucket.UpdatedDoc{}, err
-			}
-			updatedDoc, shouldUpdate, updatedExpiry, updatedHighSeq, unusedSequences, err = db.getResyncedDocument(ctx, doc, regenerateSequences, unusedSequences)
-			if err != nil {
-				return sgbucket.UpdatedDoc{}, err
-			}
-			if !shouldUpdate {
-				return sgbucket.UpdatedDoc{}, base.ErrUpdateCancel
-			}
-			base.TracefCtx(ctx, base.KeyAccess, "Saving updated channels and access grants of %q", base.UD(docid))
-			if updatedExpiry != nil {
-				updatedDoc.UpdateExpiry(*updatedExpiry)
-			}
-			doc.SetCrc32cUserXattrHash()
-
-			// Update MetadataOnlyUpdate based on previous Cas, MetadataOnlyUpdate
-			if db.useMou() {
-				doc.MetadataOnlyUpdate = computeMetadataOnlyUpdate(doc.Cas, doc.RevSeqNo, doc.MetadataOnlyUpdate)
-			}
-
-			_, rawSyncXattr, _, rawMouXattr, rawGlobalXattr, err := updatedDoc.MarshalWithXattrs()
-			updatedDoc := sgbucket.UpdatedDoc{
-				Doc: nil, // Resync does not require document body update
-				Xattrs: map[string][]byte{
-					base.SyncXattrName: rawSyncXattr,
-				},
-				Expiry: updatedExpiry,
-			}
-			if db.useMou() {
-				updatedDoc.Xattrs[base.MouXattrName] = rawMouXattr
-				if doc.MetadataOnlyUpdate.HexCAS == expandMacroCASValueString {
-					updatedDoc.Spec = append(updatedDoc.Spec, sgbucket.NewMacroExpansionSpec(XattrMouCasPath(), sgbucket.MacroCas))
-				}
-			}
-			if rawGlobalXattr != nil {
-				updatedDoc.Xattrs[base.GlobalXattrName] = rawGlobalXattr
-			}
-			return updatedDoc, err
+	writeUpdateFunc := func(currentValue []byte, currentXattrs map[string][]byte, cas uint64) (sgbucket.UpdatedDoc, error) {
+		// resyncDocument is not called on tombstoned documents, so this value will only be empty if the document was
+		// deleted between DCP event and calling this function. In any case, we do not need to update it.
+		if len(currentValue) == 0 {
+			return sgbucket.UpdatedDoc{}, base.ErrUpdateCancel
 		}
-		opts := &sgbucket.MutateInOptions{
-			MacroExpansion: macroExpandSpec(base.SyncXattrName),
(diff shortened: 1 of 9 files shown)


torcolvin:
When looking at this PR, evaluate this for safety to put into 4.0.2 to address CBG-5017 and obviate the need for https://github.com/couchbase/sync_gateway/pull/7890

db/database.go
```diff
@@ -1822,97 +1823,70 @@ func (db *DatabaseCollectionWithUser) getResyncedDocument(ctx context.Context, d
...
-				if doc.MetadataOnlyUpdate.HexCAS == expandMacroCASValueString {
-					updatedDoc.Spec = append(updatedDoc.Spec, sgbucket.NewMacroExpansionSpec(XattrMouCasPath(), sgbucket.MacroCas))
-				}
-			}
-			if rawGlobalXattr != nil {
-				updatedDoc.Xattrs[base.GlobalXattrName] = rawGlobalXattr
-			}
-			return updatedDoc, err
+		_, rawSyncXattr, rawVvXattr, rawMouXattr, rawGlobalXattr, err := updatedDoc.MarshalWithXattrs()
+		updatedDoc := sgbucket.UpdatedDoc{
```
gregns1, line 1854: NIT: This declaration here shadows the declaration on line 1827

db/document.go
```diff
@@ -1578,3 +1563,42 @@ func (d DocVersion) CVOrRevTreeID() string {
...
+	return &sgbucket.BucketDocument{
+		Body:        body,
+		Xattrs:      xattrs,
+		Cas:         event.Cas,
+		Expiry:      event.Expiry,
+		IsTombstone: event.Opcode == sgbucket.FeedOpDeletion,
+	}, nil
+}
+
+func getBucketDocument(ctx context.Context, collection *DatabaseCollection, docID string) (*sgbucket.BucketDocument, error) {
```
adamcfraser: Is this a test-only function at this point in time?  I was going to suggest that import could rely on preserveExpiry at this point, but it looks like this function doesn't have any non-test usage?

db/database.go
```diff
@@ -1822,97 +1823,70 @@ func (db *DatabaseCollectionWithUser) getResyncedDocument(ctx context.Context, d
...
-				}
-
-				updatedBytes, marshalErr := base.JSONMarshal(updatedDoc)
-				return updatedBytes, updatedExpiry, false, marshalErr
-			} else {
-				return nil, nil, false, base.ErrUpdateCancel
-			}
-		})
+		if rawGlobalXattr != nil {
+			updatedDoc.Xattrs[base.GlobalXattrName] = rawGlobalXattr
```
adamcfraser: Could this be removed, since resync won't be adding/removing attachments?

db/database.go
```diff
@@ -1822,97 +1823,70 @@ func (db *DatabaseCollectionWithUser) getResyncedDocument(ctx context.Context, d
...
+		doc.SetCrc32cUserXattrHash()
 
-			// Update MetadataOnlyUpdate based on previous Cas, MetadataOnlyUpdate
-			if db.useMou() {
-				doc.MetadataOnlyUpdate = computeMetadataOnlyUpdate(doc.Cas, doc.RevSeqNo, doc.MetadataOnlyUpdate)
-			}
+		// Update MetadataOnlyUpdate based on previous Cas, MetadataOnlyUpdate
+		doc.MetadataOnlyUpdate = computeMetadataOnlyUpdate(doc.Cas, doc.RevSeqNo, doc.MetadataOnlyUpdate)
+		mouMatch := false // after writing this document, mou.cas != doc.cas since otherwise shouldUpdate would be false
+		doc, err = db.updateHLV(ctx, doc, Resync, mouMatch)
```
gregns1: Any reason not to just pass in false here given mouMatch is always false?

db/database.go
```diff
@@ -1822,97 +1823,70 @@ func (db *DatabaseCollectionWithUser) getResyncedDocument(ctx context.Context, d
...
+		}
+		doc.SetCrc32cUserXattrHash()
 
-			// Update MetadataOnlyUpdate based on previous Cas, MetadataOnlyUpdate
-			if db.useMou() {
-				doc.MetadataOnlyUpdate = computeMetadataOnlyUpdate(doc.Cas, doc.RevSeqNo, doc.MetadataOnlyUpdate)
-			}
+		// Update MetadataOnlyUpdate based on previous Cas, MetadataOnlyUpdate
+		doc.MetadataOnlyUpdate = computeMetadataOnlyUpdate(doc.Cas, doc.RevSeqNo, doc.MetadataOnlyUpdate)
+		mouMatch := false // after writing this document, mou.cas != doc.cas since otherwise shouldUpdate would be false
```
gregns1: I find this comment confusing as after this update has been completed mou.cas should be == to doc cas right?

torcolvin:
To consider:

Is the latest set of changes better than https://github.com/couchbase/sync_gateway/pull/7892/files/eb809bf36ded144a008de3741c629a11a999d827 ? Does it add more risk to the ticket to make it not worthwhile? The idea behind the latest set of changes is to avoid having to set `mouMatch` and moving the update to a single point in code. This means that we move

The refactor exposes something that is tricky which is that `Document.RevSeqNo` is present on the `Document` struct but only populated when an import is occurring, since we do not want to fetch an extra xattr if needed. This means the value is often 0, so I added code to check if `Document.RevSeqNo = 0` and not write the mou in this case. An error here would be a developer problem.

I thought about if the right thing to do is to add `RevSeqNo` to `sgbucket.BucketDocument` which almost fixes the problem, except for on of the on demand import fields on write. Because `documentUpdateFunc` takes `db.Document` object, we need to be able to pass rev seq no in that function. That refactor became even bigger than the existing refactor that I did to consolidate mou handling.

db/database.go
```diff
@@ -1822,97 +1823,70 @@ func (db *DatabaseCollectionWithUser) getResyncedDocument(ctx context.Context, d
...
```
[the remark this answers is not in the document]
torcolvin: This code is now obsolete.

db/database.go
torcolvin: Obviated this code in a separate PR.

db/database.go
torcolvin: Obviated this code in a separate PR.

torcolvin:
> To consider:
> 
> Is the latest set of changes better than [couchbase/sync_gateway/pull/7892/files/eb809bf36ded144a008de3741c629a11a999d827](https://github.com/couchbase/sync_gateway/pull/7892/files/eb809bf36ded144a008de3741c629a11a999d827) ? Does it add more risk to the ticket to make it not worthwhile? The idea behind the latest set of changes is to avoid having to set `mouMatch` and moving the update to a single point in code. This means that we move
> 
> The refactor exposes something that is tricky which is that `Document.RevSeqNo` is present on the `Document` struct but only populated when an import is occurring, since we do not want to fetch an extra xattr if needed. This means the value is often 0, so I added code to check if `Document.RevSeqNo = 0` and not write the mou in this case. An error here would be a developer problem.
> 
> I thought about if the right thing to do is to add `RevSeqNo` to `sgbucket.BucketDocument` which almost fixes the problem, except for on of the on demand import fields on write. Because `documentUpdateFunc` takes `db.Document` object, we need to be able to pass rev seq no in that function. That refactor became even bigger than the existing refactor that I did to consolidate mou handling.

This entire comment is obviated since it was determined to always update `_mou` and never `_vv` in https://github.com/couchbase/sync_gateway/commit/693e4b55bc80478471db883068c2ba330f084794

db/database_collection.go
```diff
@@ -323,9 +323,20 @@ func (c *DatabaseCollection) syncGlobalSyncAndUserXattrKeys() []string {
...
+	}
+	return xattrKeys
+}
+
 // syncGlobalSyncMouRevSeqNoAndUserXattrKeys returns the xattr keys for the user, mou, revSeqNo and sync xattrs.
 func (c *DatabaseCollection) syncGlobalSyncMouRevSeqNoAndUserXattrKeys() []string {
-	xattrKeys := []string{base.SyncXattrName, base.VvXattrName}
+	xattrKeys := []string{base.SyncXattrName, base.VvXattrName, base.VirtualXattrRevSeqNo}
 	if c.useMou() {
 		xattrKeys = append(xattrKeys, base.MouXattrName, base.GlobalXattrName)
```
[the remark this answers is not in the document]
torcolvin: Removed unused useMou() here.

db/document.go
```diff
@@ -1578,3 +1563,18 @@ func (d DocVersion) CVOrRevTreeID() string {
...
+	body, xattrs, err := sgbucket.DecodeValueWithAllXattrs(event.Value)
+	if err != nil {
+		return nil, err
+	}
+	return &sgbucket.BucketDocument{
+		Body:        body,
+		Xattrs:      xattrs,
+		Cas:         event.Cas,
+		Expiry:      event.Expiry,
+		IsTombstone: event.Opcode == sgbucket.FeedOpDeletion,
```
adamcfraser, line 1578: Does this handle all tombstone cases?  e.g. if we modify a system xattr on a document that's already a tombstone, will that be sent as FeedOpDeletion or FeedOpMutation?

db/database.go
```diff
@@ -1824,96 +1824,65 @@ func (db *DatabaseCollectionWithUser) getResyncedDocument(ctx context.Context, d
...
-			updatedDoc, shouldUpdate, updatedExpiry, updatedHighSeq, unusedSequences, err = db.getResyncedDocument(ctx, doc, regenerateSequences, unusedSequences)
-			if err != nil {
-				return nil, nil, false, err
-			}
-			if shouldUpdate {
-				base.TracefCtx(ctx, base.KeyAccess, "Saving updated channels and access grants of %q", base.UD(docid))
-				if updatedExpiry != nil {
-					updatedDoc.UpdateExpiry(*updatedExpiry)
-				}
+		updatedDoc, shouldUpdate, updatedExpiry, _, updatedUnusedSequences, err = db.getResyncedDocument(ctx, doc, regenerateSequences, nil)
```
adamcfraser: Did you consider removing highSeq from getResyncedDocument since it doesn't appear to be used?  Or maybe a follow-up PR to remove it?

db/database.go
```diff
@@ -1824,96 +1824,65 @@ func (db *DatabaseCollectionWithUser) getResyncedDocument(ctx context.Context, d
...
+		_, rawSyncXattr, _, rawMouXattr, rawGlobalXattr, err := updatedDoc.MarshalWithXattrs()
+		updatedDoc := sgbucket.UpdatedDoc{
+			Doc: nil, // Resync does not require document body update
+			Xattrs: map[string][]byte{
+				base.SyncXattrName: rawSyncXattr,
+				base.MouXattrName:  rawMouXattr,
+			},
+			Expiry: updatedExpiry,
+		}
+		if doc.MetadataOnlyUpdate != nil {
```
adamcfraser: Should line 1861 (adding the mou xattr) be inside this if clause?  Or is there a scenario where rawMouXattr is non-nil, but doc.MetadataOnlyUpdate is nil?

adamcfraser (changes requested):
Generally looks good, a few questions on the details.

db/document.go
torcolvin: Yes, any tombstone mutation will show up as `FeedOpDeletion`.

db/database.go
torcolvin: yup

db/database.go
torcolvin: This if statement doesn't need to exist anymore, it did for <4.0 since Sync Gateway didn't write `_mou` unless CBS was >=7.6

db/document.go
torcolvin: Since I moved this to test only code, I was going to just leave as is.

db/database.go
torcolvin: I did decide to remove this.

adamcfraser (approved):
LGTM - thanks for all the review feedback.

PR #7895, opened 2025-11-24
CBG-5020: fix panic in eviction based on out of sync stats

gregns1:
CBG-5020

- Defensive checks when evicting based off of memory 
- Moves some code around to have defer use for unlocking 

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/173/

### DIFF
--- a/db/revision_cache_lru.go
+++ b/db/revision_cache_lru.go
@@ -940,8 +940,12 @@ func (rc *LRURevisionCache) evictBasedOffMemoryUsage(ctx context.Context) int64 
 					// list is empty, nothing more to evict but stats are wrong so zero stats and return
 					base.DebugfCtx(ctx, base.KeyCache, "Revision cache memory stats inconsistent for this shard, resetting to zero")
 					correctionVal := rc.currMemoryUsage.Value()
-					rc.currMemoryUsage.Add(-correctionVal)
+					// Correct overall memory stat across shards to remove this shards memory usage. We cannot set this
+					// to 0 as other shards usage is included in this
 					rc.cacheMemoryBytesStat.Add(-correctionVal)
+					// Set this shards memory usage to zero. We can set this to 0 given this count is for
+					// this particular shard.
+					rc.currMemoryUsage.Set(0)
 					break
 				}
 			}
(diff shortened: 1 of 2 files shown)


db/revision_cache_lru.go
```diff
@@ -912,17 +918,32 @@ func (rc *LRURevisionCache) revCacheMemoryBasedEviction(ctx context.Context) {
...
-				break
+				if rc.lruList.Len() > 0 {
+					// no more values ready for eviction
+					break
+				} else {
+					// list is empty, nothing more to evict but stats are wrong so zero stats and return
+					base.DebugfCtx(ctx, base.KeyCache, "Revision cache memory stats inconsistent for this shard, resetting to zero")
+					correctionVal := rc.currMemoryUsage.Value()
+					rc.currMemoryUsage.Add(-correctionVal)
+					rc.cacheMemoryBytesStat.Add(-correctionVal)
```
bbrks, line 945: It would be much clearer and potentially safer to use Set(0) for this than doing two separate Value and Add IMO

db/revision_cache_test.go
```diff
@@ -2442,3 +2443,41 @@ func TestEvictionWhenStaleCVRemoved(t *testing.T) {
 	revValueRevEntry := valRevEntry.Value.(*revCacheValue)
 	assert.Equal(t, "2-abc", revValueRevEntry.revID)
 }
+
+func TestUpdateDeltaRevCacheMemoryStatPanic(t *testing.T) {
```
bbrks, line 2450: Would be worth adding a comment to this test that describes why we're interleaving two revcache operations to simulate the concurrency bug.

Perhaps also clearer comments for each "concurrent" operation being run (i.e. `// Thread 1: UpdateDelta (start)` `// Thread 2: Remove (start)`

bbrks (changes requested):

db/revision_cache_lru.go
gregns1: I have changed this, I have added some comments on this area, we can;t quite set `cacheMemoryBytesStat` to 0 given this is the memory usage across all shards.

bbrks (approved):

PR #7900, opened 2025-11-27
CBG-5020 Avoid double-counting memory correction

adamcfraser:
CBG-5020

Return instead of break to avoid double-counting entries that were removed prior to the empty list state.

Added a second test to cover the multi-value case.


bbrks (approved):
LGTM 👍 thanks for catching this!

PR #7916, opened 2025-12-02
CBG-4286: Add sg prefix to dcp feed names

RIT3shSapata:
CBG-4286

Describe your PR here...
- Added prefix to DCP feed names as per KV conventions

## Pre-review checklist

- [ ] Link upstream PRs

### DIFF
--- a/base/dcp_common.go
+++ b/base/dcp_common.go
@@ -298,7 +298,7 @@ func GenerateDcpStreamName(feedID string) (string, error) {
 	commitTruncated := StringPrefix(GitCommit, 7)
 
 	return fmt.Sprintf(
-		"%v-v-%v-commit-%v-uuid-%v",
+		"sg:%v-v-%v-commit-%v-uuid-%v",
 		feedID,
 		ProductAPIVersion,
 		commitTruncated,


torcolvin (commented):
I think this is going to add "sg" to places where SG is already set - 

See how this is called through db/import_listener.go and db/change_listener.go

I do try to fix this in https://github.com/couchbase/sync_gateway/pull/7879 but not entirely.

A good way to see what it used is to add print debugging https://github.com/couchbase/sync_gateway/blob/693e4b55bc80478471db883068c2ba330f084794/base/dcp_client.go#L371 to see the ID and run the integration tests (unfortunately this is SG_TEST_BACKING_STORE=couchbase until https://github.com/couchbase/sync_gateway/pull/7879, so you could wait for this PR)

The other place this is used is by cbgt for EE import feed https://github.com/couchbase/cbgt/blob/74dbd1c34db313c2b65a65f6e3cecca7eacb00b3/gocbcore_agents.go#L232

It would be good to know the names for each of the dcp feed types we run:

- import feed
- caching feed
- resync background process
- attachment migration background process
- attachment compaction background process

I think that we sort of mix and match names and versions in this.

The other thing I think this PR should do is actually check the value will be <200, since this https://github.com/couchbase/sync_gateway/blob/693e4b55bc80478471db883068c2ba330f084794/base/dcp_common_test.go#L27 tests in the non production case.

bbrks:
Will close this PR for now - since it was decided that this DCP name change would be done either after some of Tor's DCP refactoring or rolled up into it (CBG-4249)

We can reopen if this changes.

PR #7918, opened 2025-12-03
CBG-5023: refactor memory tracking around deltas at the cache

gregns1:
CBG-5023

**In progress**

The overall problem here is given we have to measure the size of a delta now we have issues with inconsistency with memory stats when rev cache state changes underneath the update delta process. 

Problem 1: 
- Thread 1: loading from bucket and in doing so is setting the item bytes on the value itself 
- Thread 2: is trying to update the delta on this value (this thread will acquire rev cache lock when thread 1 is done creating the value and over onto populating the value from the bucket)
- Issue arises that thread 2 will update our stats with size of delta and also update the value bytes size itself with size of delta then thread 1 goes to increment the stats with the memory footprint of the value (at this point it has been updated to increment the size of delta) 
    - Thus at this stage we end up double counting the delta size (once in update delta process and then once in increment memory stats post load from bucket) 

Problem 2: 
- Thread 1: calls update delta
- Thread 2: calls remove item, this thread will wait until rev cache mutex is released from thread 1, once this has taken place this thread will proceed with removing the value and reading decrementing overall memory footprint of cache with this values memory value
	- The issue arises here, thread 2 is removing the value underneath the update delta process, so thread 2 is decrementing the overall memory stats by the values footprint, this could be before or after thread 1 has updated the values memory footprint 
	- This means UpdateDelta thread (thread 1) is updating the stats with different view than what thread 2 is doing, we can end up with inconsistent stats here
	- This case is the case we think was hit in https://jira.issues.couchbase.com/browse/CBG-5020 

Overall delta sync is incompatible with memory based eviction as it stands. Enough workload through this path will lead to stats being incorrect. The problem we have is we cannot let the `UpdateDelta` process release the rev cache lock, given this is why we sometimes can end up with a value being removed underneath the UpdateDelta process. But we also cannot acquire the rev cache value lock here as we already have the rev cache lock (this should be avoided due to risk of deadlock). 

With recent changes that went in I tried using the new delta lock on the value, the issue here is holding this lock and rev cache lock seemed to lead to some deadlock scenarios in testing. 

Summary of changes:
- Have rev cache lock be held for the whole update delta process. This stops this particular value being removed under the update delta process. We need this to stop the removal process decrementing by some amount that doesn't reflect what the UpdateDelta process is incrementing by. 
- Have update delta process not follow through if the value itself hasn't been loaded/stored into yet. This is using the atomic boolean we have on the value. It is not safe to continue if some other thread is altering the value struct (during load or store functions).
- I have changed the decrement memory stats function. This function will decrement the memory value of the value itself AND the delta (if one exists). 
- Change to make `itemBytes` on the rev cache value just be a count of the rev cache items bytes (not the delta if one exists on the value). This means the increment memory stats function will ONLY increment memory stats by the values memory footprint (and not the delta) and we now leave the delta memory tracking to `UpdateDelta` 

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/196/

### DIFF
--- a/db/revision_cache_lru.go
+++ b/db/revision_cache_lru.go
@@ -119,23 +119,23 @@ type LRURevisionCache struct {
 
 // The cache payload data. Stored as the Value of a list Element.
 type revCacheValue struct {
-	err          error
-	history      Revisions
-	channels     base.Set
-	expiry       *time.Time
-	attachments  AttachmentsMeta
-	delta        *RevisionDelta
-	id           string
-	cv           Version
-	hlvHistory   string
-	revID        string
-	bodyBytes    []byte
-	lock         sync.RWMutex
-	deleted      bool
-	removed      bool
-	itemBytes    atomic.Int64
-	collectionID uint32
-	canEvict     atomic.Bool
+	err            error
+	history        Revisions
+	channels       base.Set
+	expiry         *time.Time
+	attachments    AttachmentsMeta
+	delta          *RevisionDelta
+	id             string
+	cv             Version
+	hlvHistory     string
+	revID          string
+	bodyBytes      []byte
+	lock           sync.RWMutex
+	deleted        bool
+	removed        bool
+	itemBytes      atomic.Int64
+	collectionID   uint32
+	valuePopulated atomic.Bool
 	deltaLock    sync.Mutex // synchronizes GetDelta across multiple clients for each fromRevision
 }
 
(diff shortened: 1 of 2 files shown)


gregns1:
Closing as we should take the approach of re architecting the rev cache to deal with this problem more sufficiently.

PR #7923, opened 2025-12-05
CBG-4915: Fix flakiness of `TestMultiActorLosingConflictUpdateRemovingAttachments`

RIT3shSapata:
CBG-4915

Describe your PR here...
- The test was flaky because the test would progress from the `require.EventuallyWithT` even though the condition is not completely met
- Refactored the code to use `RestTester`'s `WaitForVersion` method instead

## Pre-review checklist

- [ ] Link upstream PRs

### DIFF
--- a/rest/utilities_testing_resttester.go
+++ b/rest/utilities_testing_resttester.go
@@ -237,6 +237,11 @@ func (rt *RestTester) WaitForVersionRevIDOnly(docID string, version DocVersion) 
 	rt.WaitForVersion(docID, version)
 }
 
+func (rt *RestTester) WaitForVersionHLVOnly(docID string, version DocVersion) {
+	version.RevTreeID = ""
+	rt.WaitForVersion(docID, version)
+}
+
 // WaitForTombstone waits for a the document version to exist and be tombstoned. If the document is not found, the test will fail.
 func (rt *RestTester) WaitForTombstone(docID string, deleteVersion DocVersion) {
 	collection, ctx := rt.GetSingleTestDatabaseCollectionWithUser()
--- a/rest/utilities_testing_resttester.go
+++ b/rest/utilities_testing_resttester.go
@@ -237,6 +237,7 @@ func (rt *RestTester) WaitForVersionRevIDOnly(docID string, version DocVersion) 
 	rt.WaitForVersion(docID, version)
 }
 
+// WaitForCV waits for the document's current version to match the expectedVersion. Fails the test harness. WaitForVersion should be used in the general case to test revtree and and cv behavior.
 func (rt *RestTester) WaitForVersionHLVOnly(docID string, version DocVersion) {
 	version.RevTreeID = ""
 	rt.WaitForVersion(docID, version)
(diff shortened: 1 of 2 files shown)


xdcr/attachment_test.go
```diff
@@ -77,12 +77,8 @@ func TestMultiActorLosingConflictUpdateRemovingAttachments(t *testing.T) {
...
 
 	// wait for doc to replicate to rtB
-	var rtBVersion rest.DocVersion
-	require.EventuallyWithT(t, func(c *assert.CollectT) {
-		rtBVersion, _ = rtB.GetDoc(docID)
-		assert.Equal(c, rtAVersion.CV.String(), rtBVersion.CV.String())
-		assert.Equal(c, rtAVersion.RevTreeID, rtBVersion.RevTreeID)
-	}, time.Second*5, time.Millisecond*100)
+	rtB.WaitForVersion(docID, rtAVersion)
+	rtBVersion, _ := rtB.GetDoc(docID)
```
torcolvin, line 81: I think given this assertion, that rtBVersion and rtAVersion are the same.

I would perhaps assert that the version has CV and Revtree.

It is possible that this assertion will only work on CV, which will wait for the document to be copied (CV matches) but not imported.

I haven't looked into this, but we can discuss.

xdcr/attachment_test.go
torcolvin: This needs to get changed too

torcolvin: We should move these to a defer. In the case of rosmar, the xdcr replications will stop when the buckets disappear but in CBS they won't, and so this could leak into future tests.

Maybe this is actually already handled by defer at the top.

xdcr/attachment_test.go
RIT3shSapata: I checked the implementation of `WaitForVersion` and I see that it is asserting for both RevTree and CV. The document is fetched via the Admin API, so this would mean that the document has been imported right? It's not directly reading from the datastore

xdcr/attachment_test.go
RIT3shSapata: Yes I think its handled by the defer above

RIT3shSapata: done

torcolvin: If this is handled by the defer above, I think we can drop this just as part of cleanup, so I don't ask the same question again next time we have to look at this code.

The easier we can make the tests to read will make the tests easier to debug.

rest/utilities_testing_resttester.go
```diff
@@ -237,6 +237,11 @@ func (rt *RestTester) WaitForVersionRevIDOnly(docID string, version DocVersion)
 	rt.WaitForVersion(docID, version)
 }
 
+func (rt *RestTester) WaitForVersionHLVOnly(docID string, version DocVersion) {
+	version.RevTreeID = ""
+	rt.WaitForVersion(docID, version)
+}
```
torcolvin, line 244: for readability:



```suggestion
// WaitForCV waits for the document's current version to match the expectedVersion. Fails the test harness. WaitForVersion should be used in the general case to test revtree and and cv behavior.
func (rt *RestTester) WaitForCV(docID string, expectedVersion DocVersion) {
	expectedVersion.RevTreeID = ""
	rt.WaitForVersion(docID, expectedVersion)
}
```

torcolvin (approved):

PR #7924, opened 2025-12-10
CBG-5050: fix for no rev messages being sent as changes on blip tester client

gregns1:
CBG-5050

- Recent merge uncovered some issues with no rev messages being sent as changes to SGW and subsequent rev messages for these changes being sent with nil bodies. This was causing issues in SGW 

## Pre-review checklist

- [ ] Link upstream PRs

### DIFF
--- a/rest/utilities_testing_blip_client.go
+++ b/rest/utilities_testing_blip_client.go
@@ -113,6 +113,10 @@ func (btcc *BlipTesterCollectionClient) getProposeChangesForSeq(seq clientSeq) (
 	if !ok {
 		return nil, false
 	}
+	latestRev := doc._revisionsBySeq[doc._latestSeq]
+	if latestRev.noRev {
+		return nil, false
+	}
 	return doc._proposeChangesEntryForDoc(), true
 }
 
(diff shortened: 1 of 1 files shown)


torcolvin (commented):
I think a norev should just not result in a document write.

Can you write a test outside the topologytests that will exhibit this behavior? Why is a norev getting sent at all?

torcolvin:
As discussed offline, I'm OK with this as a temporary fix. This allows topologytests to pass and also `TestSendReplacementRevision` to pass.

In the future we should:

- write a specific test that sends a norev to blip tester outside the topology tests. This would be a case where the change exists in the change cache, but isn't available when actually fetched. This should be done for replacement revs but also non replacement rev cases.
- Not write a local revision in our blip tester storage, a no rev doesn't represent a revision. If there were checkpoints in the blip tester, we would advance the checkpoint.
- Make sure that receiving a norev will delete the document from the blip tester storage, a norev represents a document deletion, even if the initial changes message was not a deletion.

Some of these might exist, but I think making these changes is out of scope for this ticket, which was intended to implement more accurate CBL behavior. The future enhancements would be to better match CBL behavior, but are not blocking any current tests.

torcolvin (approved):

PR #7925, opened 2025-12-10
CBG-5051 make sure to marshal sequence on norev

torcolvin:
CBG-5051 make sure to marshal sequence on norev

- Asserts that the `sequence` property for `rev` and `norev` matches that of the changes messages. The reason that this wasn't checkpointed by CBL is that for any requested revision from a changes messages, CBL is waiting for a matching string from the `sequence` property on a `rev` or `norev`.
- Added err checking rather than asserts just because it seemed like a bad smell, but I can drop this out. I don't think it makes the code riskier.

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/201/

### DIFF
--- a/db/blip.go
+++ b/db/blip.go
@@ -12,6 +12,7 @@ package db
 
 import (
 	"context"
+	"fmt"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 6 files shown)


gregns1 (approved):
LGTM

PR #7932, opened 2025-12-16
CBG-4411: Improvements to Diagnostic Sync Function dry run endpoint

RIT3shSapata:
CBG-4411

Describe your PR here...
- The endpoint can now run with a custom function provided by the user as well as a document
- The `SyncFnDryRun` function will now create an ephemeral `SyncRunner` if a custom sync function is passed
- Refactored the unit tests for the Sync Function dry run

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/202/
- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/207/

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -1694,58 +1694,11 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
 
 }
 
-func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, oldDoc *Document, body Body, docID, syncFn string) (*channels.ChannelMapperOutput, error) {
 // SyncFnDryrun Runs the given document body through a sync function and returns expiry, channels doc was placed in,
 // access map for users, roles, handler errors and sync fn exceptions.
 // If syncFn is provided, it will be used instead of the one configured on the database.
+func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, newDoc, oldDoc *Document, syncFn string) (*channels.ChannelMapperOutput, error) {
 
-	delete(body, BodyId)
-
-	// Get the revision ID to match, and the new generation number:
-	matchRev, _ := body[BodyRev].(string)
-	generation, _ := ParseRevID(ctx, matchRev)
-	if generation < 0 {
-		return nil, base.HTTPErrorf(http.StatusBadRequest, "Invalid revision ID")
-	}
-	generation++
-
-	// Create newDoc which will be used to pass around Body
-	newDoc := &Document{
-		ID: docID,
-	}
-	// Pull out attachments
-	newDoc.SetAttachments(GetBodyAttachments(body))
-	delete(body, BodyAttachments)
-
-	delete(body, BodyRevisions)
-
-	err := validateAPIDocUpdate(body)
-	if err != nil {
-		return nil, err
-	}
-	bodyWithoutInternalProps, wasStripped := StripInternalProperties(body)
-	canonicalBytesForRevID, err := base.JSONMarshalCanonical(bodyWithoutInternalProps)
-	if err != nil {
-		return nil, err
-	}
-
-	// We needed to keep _deleted around in the body until we generated a rev ID, but now we can ditch it.
-	_, isDeleted := body[BodyDeleted]
-	if isDeleted {
-		delete(body, BodyDeleted)
-	}
-
-	// and now we can finally update the newDoc body to be without any special properties
-	newDoc.UpdateBody(body)
-
-	// If no special properties were stripped and document wasn't deleted, the canonical bytes represent the current
-	// body.  In this scenario, store canonical bytes as newDoc._rawBody
-	if !wasStripped && !isDeleted {
-		newDoc._rawBody = canonicalBytesForRevID
-	}
-
-	newRev := CreateRevIDWithBytes(generation, matchRev, canonicalBytesForRevID)
-	newDoc.RevID = newRev
 	mutableBody, metaMap, _, err := db.prepareSyncFn(oldDoc, newDoc)
 	if err != nil {
 		base.InfofCtx(ctx, base.KeyDiagnostic, "Failed to prepare to run sync function: %v", err)
--- a/db/crud.go
+++ b/db/crud.go
@@ -1697,28 +1697,7 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
 // SyncFnDryrun Runs a document through the sync function and returns expiry, channels doc was placed in, access map for users, roles, handler errors and sync fn exceptions.
 // If a docID is a non empty string, the document will be fetched from the bucket, otherwise the body will be used. If both are specified, this function returns an error.
 // The first error return value represents an error that occurs before the sync function is run. The second error return value represents an exception from the sync function.
-func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, body Body, docID, syncFn string) (*channels.ChannelMapperOutput, error, error) {
-	doc := &Document{
-		ID:    docID,
-		_body: body,
-	}
-	oldDoc := doc
-	if docID != "" {
-		if docInBucket, err := db.GetDocument(ctx, docID, DocUnmarshalAll); err == nil {
-			oldDoc = docInBucket
-			if len(doc._body) == 0 {
-				body = oldDoc.Body(ctx)
-				doc._body = body
-				// If no body is given, use doc in bucket as doc with no old doc
-				oldDoc._body = nil
-			}
-			doc._body[BodyRev] = oldDoc.SyncData.GetRevTreeID()
-		} else {
-			return nil, err, nil
-		}
-	} else {
-		oldDoc._body = nil
-	}
+func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, oldDoc *Document, body Body, docID, syncFn string) (*channels.ChannelMapperOutput, error) {
 
 	delete(body, BodyId)
 
@@ -1726,7 +1705,7 @@ func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, body Bod
 	matchRev, _ := body[BodyRev].(string)
 	generation, _ := ParseRevID(ctx, matchRev)
 	if generation < 0 {
-		return nil, base.HTTPErrorf(http.StatusBadRequest, "Invalid revision ID"), nil
+		return nil, base.HTTPErrorf(http.StatusBadRequest, "Invalid revision ID")
 	}
 	generation++
 
@@ -1742,12 +1721,12 @@ func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, body Bod
 
 	err := validateAPIDocUpdate(body)
 	if err != nil {
-		return nil, err, nil
+		return nil, err
 	}
 	bodyWithoutInternalProps, wasStripped := StripInternalProperties(body)
 	canonicalBytesForRevID, err := base.JSONMarshalCanonical(bodyWithoutInternalProps)
 	if err != nil {
-		return nil, err, nil
+		return nil, err
 	}
 
 	// We needed to keep _deleted around in the body until we generated a rev ID, but now we can ditch it.
@@ -1770,30 +1749,33 @@ func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, body Bod
 	mutableBody, metaMap, _, err := db.prepareSyncFn(oldDoc, newDoc)
 	if err != nil {
 		base.InfofCtx(ctx, base.KeyDiagnostic, "Failed to prepare to run sync function: %v", err)
-		return nil, err, nil
+		return nil, err
 	}
 
 	syncOptions, err := MakeUserCtx(db.user, db.ScopeName, db.Name)
 	if err != nil {
-		return nil, err, nil
+		return nil, err
 	}
 	var output *channels.ChannelMapperOutput
 	if syncFn == "" {
 		output, err = db.ChannelMapper.MapToChannelsAndAccess(ctx, mutableBody, string(oldDoc._rawBody), metaMap, syncOptions)
+		if err != nil {
+			err = fmt.Errorf("%s%s", base.ErrSyncFnDryRun, err)
+		}
 	} else {
 		jsTimeout := time.Duration(base.DefaultJavascriptTimeoutSecs) * time.Second
 		syncRunner, err := channels.NewSyncRunner(ctx, syncFn, jsTimeout)
 		if err != nil {
-			return nil, fmt.Errorf("failed to create sync runner: %v", err), nil
+			return nil, fmt.Errorf("failed to create sync runner: %v", err)
 		}
 		jsOutput, err := syncRunner.Call(ctx, mutableBody, string(oldDoc._rawBody), metaMap, syncOptions)
 		if err != nil {
-			return nil, err, nil
+			err = fmt.Errorf("%s%s", base.ErrSyncFnDryRun, err)
 		}
 		output = jsOutput.(*channels.ChannelMapperOutput)
 	}
 
-	return output, nil, err
+	return output, err
 }
 
 // revTreeConflictCheck checks for conflicts in the rev tree history and returns the parent revid, currentRevIndex
--- a/db/crud.go
+++ b/db/crud.go
@@ -1694,7 +1694,9 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
 
 }
 
-// SyncFnDryrun Runs a document through the sync function and returns expiry, channels doc was placed in, access map for users, roles, handler errors and sync fn exceptions
+// SyncFnDryrun Runs a document through the sync function and returns expiry, channels doc was placed in, access map for users, roles, handler errors and sync fn exceptions.
+// If a docID is a non empty string, the document will be fetched from the bucket, otherwise the body will be used. If both are specified, this function returns an error.
+// The first error return value represents an error that occurs before the sync function is run. The second error return value represents an exception from the sync function.
 func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, body Body, docID, syncFn string) (*channels.ChannelMapperOutput, error, error) {
 	doc := &Document{
 		ID:    docID,
@@ -1704,7 +1706,7 @@ func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, body Bod
 	if docID != "" {
 		if docInBucket, err := db.GetDocument(ctx, docID, DocUnmarshalAll); err == nil {
 			oldDoc = docInBucket
-			if doc._body == nil || len(doc._body) == 0 {
+			if len(doc._body) == 0 {
 				body = oldDoc.Body(ctx)
 				doc._body = body
 				// If no body is given, use doc in bucket as doc with no old doc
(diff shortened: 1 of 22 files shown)


db/crud.go
```diff
@@ -1695,7 +1695,7 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
 }
 
 // SyncFnDryrun Runs a document through the sync function and returns expiry, channels doc was placed in, access map for users, roles, handler errors and sync fn exceptions
-func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, body Body, docID string) (*channels.ChannelMapperOutput, error, error) {
+func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, body Body, docID, syncFn string) (*channels.ChannelMapperOutput, error, error) {
```
torcolvin: Let's document the behavior of this function and add a pre condition to help the callers do the right thing:

```suggestion
// SyncFnDryrun Runs a document through the sync function and returns expiry, channels doc was placed in, access map for users, roles, handler errors and sync fn exceptions.
// If a docID is a non empty string, the document will be fetched from the bucket, otherwise the body will be used. If both are specified, this function returns an error.
// The first error return value represents an error that occurs before the sync function is run. The second error return value represents an exception from the sync function.
func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, body Body, docID, syncFn string) (*channels.ChannelMapperOutput, error, error) {
	if docID != "" && len(body) > 0 {
		return nil, fmt.Errorf("cannot call SyncFnDryrun with both a body and a docID"), nil
```

A better abstraction for this is to read the docID before you call this function and then pass the body into this function, so you don't have multiple pathways.

It is also extremely weird to return two errors, and that could be changed but is a more complicated change. I think it would be good to write a test that would return different types of errors on this function and see what they are. Two ways of changing this could be:

1. Create a different struct that like:

```
type SyncDryRunOutput {
	output *channels.ChannelMapperOutput
    err error // represents an exception running the sync function
}
```

2. Wrap the error in a specific error type like:

`const ErrSyncFn = errors.New("Exception occurred while calling a sync function")`

Then if `Call` returns an exception, return `fmt.Errorf("%w: %w", err)` and you can use `errors.Is` by the caller.

I think changing the error might be out of scope for this change, but I think it's important to make sure there's a test for that code pathway.

db/crud.go
```diff
@@ -1704,7 +1704,7 @@ func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, body Bod
 	if docID != "" {
 		if docInBucket, err := db.GetDocument(ctx, docID, DocUnmarshalAll); err == nil {
 			oldDoc = docInBucket
-			if doc._body == nil {
+			if doc._body == nil || len(doc._body) == 0 {
```
torcolvin: nit about go: checking for nil is redundant

```suggestion
			if len(doc._body) == 0 {
```

rest/diagnostic_doc_api.go
```diff
@@ -69,15 +75,19 @@ func (h *handler) handleGetDocChannels() error {
...
 	docid := h.getQuery("doc_id")
+	contentType, _, _ := mime.ParseMediaType(h.rq.Header.Get("Content-Type"))
 
-	body, err := h.readDocument()
-	if err != nil {
-		if docid == "" {
-			return fmt.Errorf("no doc id provided for dry run and error reading body: %s", err)
-		}
+	if contentType != "application/json" && contentType != "" {
+		return base.HTTPErrorf(http.StatusUnsupportedMediaType, "Invalid Content type %s", contentType)
```
torcolvin:
```suggestion
		return base.HTTPErrorf(http.StatusUnsupportedMediaType, "Invalid Content-Type header: %s. Needs to be empty or application/json", contentType)
```

rest/diagnostic_doc_api_test.go
torcolvin, line 71:
```suggestion
	rt := NewRestTester(t, &RestTesterConfig{
		PersistentConfig: true,
		ImportFilter: function(doc) { if (doc.user.num) { return true; } else { return false; } }"},
		SyncFn: "function(doc,oldDoc){if (doc.user.num >= 100) {channel(doc.channel);} else {throw({forbidden: 'user num too low'});}if (oldDoc){ console.log(oldDoc); if (oldDoc.user.num > doc.user.num) { access(oldDoc.user.name, doc.channel);} else {access(doc.user.name[0], doc.channel);}}}"`
		})
```

Using this format will allow you to call `NewDbConfig` below and have this work with collections.

rest/diagnostic_doc_api_test.go
```diff
@@ -72,80 +72,18 @@ func TestGetDocDryRuns(t *testing.T) {
...
-
-	err = json.Unmarshal(response.BodyBytes(), &respMap)
-	assert.NoError(t, err)
-	assert.Equal(t, respMap.Roles, channels.AccessMap{"user": channels.BaseSetOf(t, "role1")})
 	newSyncFn := `"function(doc,oldDoc){if (doc.user.num >= 100) {channel(doc.channel);} else {throw({forbidden: 'user num too low'});}if (oldDoc){ console.log(oldDoc); if (oldDoc.user.num > doc.user.num) { access(oldDoc.user.name, doc.channel);} else {access(doc.user.name[0], doc.channel);}}}"`
-	resp = rt.SendAdminRequest("POST", "/db/_config", fmt.Sprintf(
+	resp := rt.SendAdminRequest("PUT", "/db/", fmt.Sprintf(
 		`{"bucket":"%s", "num_index_replicas": 0, "enable_shared_bucket_access": %t, "sync":%s, "import_filter":%s}`,
 		bucket, base.TestUseXattrs(), newSyncFn, ImportFilter))
 	RequireStatus(t, resp, http.StatusCreated)
```
torcolvin, line 79: To work with collections see above for `RestTesterConfig` change.

```suggestion
resp := rt.CreateDatabase("db", rt.NewDbConfig())`
```

torcolvin (commented):
I'm going to review test cases later today, I wanted to give some comments.

db/crud.go
RIT3shSapata: done

db/crud.go
RIT3shSapata: I've made the changes the changes to the documentation. I do agree that it is strange that there are two errors are being returned, please let me know if you would like me to make the changes to the way error is handled. 

The first method of resolving this, already exists. In the handler there is a struct `SyncFnDryRun` which stores the error returned from the Sync Function and stores it in the `Exception` property. 

There are already test cases in the unit tests which try to cover the codepath of when there is an error in sync function. Please check test cases 6,7 and 11. Did you want coverage similar to the ones in these cases or a different one?

torcolvin: I think we can leave this weird code pathway for now in favor of getting things done for this sprint, but if it ends up causing confusion while you are working I think it might make sense to fix?

The answer sort of depends on how long you think it would take you to fix, and I think it is okay to leave it.

RIT3shSapata: Fixed it in the latest commit.

RIT3shSapata:
Hey @bbrks @torcolvin , I've refactored bit of redundant code, which was pointed out @torcolvin . I've also updated the API docs for the diagnostic API as well.

db/crud.go
```diff
@@ -1694,37 +1694,18 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
...
-		} else {
-			return nil, err, nil
-		}
-	} else {
-		oldDoc._body = nil
-	}
+// SyncFnDryrun Runs a document through the sync function and returns expiry, channels doc was placed in, access map for users, roles, handler errors and sync fn exceptions.
+// If a docID is a non empty string, the document will be fetched from the bucket, otherwise the body will be used. If both are specified, this function returns an error.
+// The first error return value represents an error that occurs before the sync function is run. The second error return value represents an exception from the sync function.
+func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, oldDoc *Document, body Body, docID, syncFn string) (*channels.ChannelMapperOutput, error) {
```
bbrks: Similarly, `docID` is now unused and is making the usage of the function ambiguous without reading the implementation.

```suggestion
func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, oldDoc *Document, body Body, syncFn string) (*channels.ChannelMapperOutput, error) {
```

db/crud.go
```diff
@@ -1694,37 +1694,18 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
...
-			doc._body[BodyRev] = oldDoc.SyncData.GetRevTreeID()
-		} else {
-			return nil, err, nil
-		}
-	} else {
-		oldDoc._body = nil
-	}
+// SyncFnDryrun Runs a document through the sync function and returns expiry, channels doc was placed in, access map for users, roles, handler errors and sync fn exceptions.
+// If a docID is a non empty string, the document will be fetched from the bucket, otherwise the body will be used. If both are specified, this function returns an error.
+// The first error return value represents an error that occurs before the sync function is run. The second error return value represents an exception from the sync function.
```
bbrks: This comment is outdated since it's now up to the caller to pass in document body.

Also only returns one error.

```suggestion
// SyncFnDryrun Runs the given document body through a sync function and returns expiry, channels doc was placed in, access map for users, roles, handler errors and sync fn exceptions.
// If syncFn is provided, it will be used instead of the one configured on the database.
```

db/crud.go
```diff
@@ -1768,16 +1749,34 @@ func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, body Bod
...
 	if err != nil {
-		return nil, err, nil
+		return nil, err
+	}
+	var output *channels.ChannelMapperOutput
+	var syncErr error
+	if syncFn == "" {
+		output, err = db.ChannelMapper.MapToChannelsAndAccess(ctx, mutableBody, string(oldDoc._rawBody), metaMap, syncOptions)
+		if err != nil {
+			syncErr = fmt.Errorf("%s%s", base.ErrSyncFnDryRun, err)
```
bbrks: ~I'd be tempted to use `%w` instead of `%s` here for `err` so the original can still be used by `errors.Is`~

```suggestion
			syncErr = fmt.Errorf("%s %w", base.ErrSyncFnDryRun, err)
```

**Edit:** This comment is superseded by the error handling comment in the REST API handler function.

db/crud.go
```diff
@@ -1768,16 +1749,34 @@ func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, body Bod
...
+		}
+	} else {
+		jsTimeout := time.Duration(base.DefaultJavascriptTimeoutSecs) * time.Second
+		syncRunner, err := channels.NewSyncRunner(ctx, syncFn, jsTimeout)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create sync runner: %v", err)
```
bbrks, line 1723:
```suggestion
			return nil, fmt.Errorf("failed to create sync runner: %w", err)
```

**Edit:** This comment is superseded by the error handling comment in the REST API handler function.

db/crud.go
```diff
@@ -1775,7 +1775,22 @@ func (db *DatabaseCollectionWithUser) SyncFnDryrun(ctx context.Context, body Bod
 	if err != nil {
 		return nil, err, nil
 	}
-	output, err := db.ChannelMapper.MapToChannelsAndAccess(ctx, mutableBody, string(oldDoc._rawBody), metaMap, syncOptions)
+	base.InfofCtx(ctx, base.KeyHTTP, "old doc: %s, new doc: %s", string(oldDoc._rawBody), mutableBody)
+	var output *channels.ChannelMapperOutput
+	if syncFn == "" {
+		output, err = db.ChannelMapper.MapToChannelsAndAccess(ctx, mutableBody, string(oldDoc._rawBody), metaMap, syncOptions)
+	} else {
+		jsTimeout := time.Duration(base.DefaultJavascriptTimeoutSecs) * time.Second
```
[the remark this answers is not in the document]
bbrks: It's fine as-is

rest/diagnostic_doc_api.go
```diff
@@ -69,24 +76,47 @@ func (h *handler) handleGetDocChannels() error {
 // If docid is specified and the document does not exist in the bucket, it will return error
 func (h *handler) handleSyncFnDryRun() error {
 	docid := h.getQuery("doc_id")
+	contentType, _, _ := mime.ParseMediaType(h.rq.Header.Get("Content-Type"))
```
bbrks: `h.readJSONInto` already does this mime type checking (via the `processContentEncoding` function) and handles things like gzip compression.

```suggestion
```

docs/api/paths/diagnostic/keyspace-sync.yaml
```diff
@@ -7,16 +7,39 @@
 # the file licenses/APL2.txt.
 parameters:
   - $ref: ../../components/parameters.yaml#/keyspace
```
bbrks, line 9: We should include the available query parameters in here (`doc_id`) - appreciate this was already missing before this PR but let's fix it up now.

```suggestion
  - $ref: ../../components/parameters.yaml#/keyspace
  - $ref: ../../components/parameters.yaml#/doc_id
```

rest/diagnostic_doc_api.go
```diff
@@ -69,15 +75,19 @@ func (h *handler) handleGetDocChannels() error {
...
-	if err != nil {
-		if docid == "" {
-			return fmt.Errorf("no doc id provided for dry run and error reading body: %s", err)
-		}
+	if contentType != "application/json" && contentType != "" {
+		return base.HTTPErrorf(http.StatusUnsupportedMediaType, "Invalid Content type %s", contentType)
+	}
+
+	var syncDryRunPayload SyncFnDryRunPayload
+	err := h.readJSONInto(&syncDryRunPayload)
```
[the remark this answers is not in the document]
bbrks: I agree with this comment.

Based on the table in the API documentation describing how this endpoint works - in the case where you provide a doc ID to act as `oldDoc`, and provide some invalid JSON for `doc` - this error will be a silent failure (running the sync function on an empty body).

This should be able to be covered by a unit test case.

rest/diagnostic_doc_api.go
```diff
@@ -69,24 +76,47 @@ func (h *handler) handleGetDocChannels() error {
...
-		return err
-	}
-	if syncFnErr != nil {
-		resp := SyncFnDryRun{
-			Exception: syncFnErr.Error(),
+		if strings.Contains(err.Error(), base.ErrSyncFnDryRun.Error()) {
+			errMsg := strings.ReplaceAll(err.Error(), base.ErrSyncFnDryRun.Error(), "")
+			resp := SyncFnDryRun{
+				Exception: errMsg,
+			}
```
bbrks: Something like this would be better achieved using `errors.As`, where you assert on the type of error being returned instead of string comparisons.

https://go.dev/blog/go1.13-errors

This would involve making `ErrSyncFnDryRun` a new type instead of an instance of `sgError`:

```go
const syncFnDryRunErrorPrefix = "Error returned from Sync Function"

// SyncFnDryRunError is returned when the sync function dry run returns an error.
// It wraps the original error for errors.Is and the type supports errors.As
type SyncFnDryRunError struct {
	Err error
}

func (e *SyncFnDryRunError) Error() string {
	if e == nil {
		return syncFnDryRunErrorPrefix
	}
	if e.Err == nil {
		return syncFnDryRunErrorPrefix
	}
	return syncFnDryRunErrorPrefix + ": " + e.Err.Error()
}

func (e *SyncFnDryRunError) Unwrap() error {
	if e == nil {
		return nil
	}
	return e.Err
}
```

In use:
```go
return &base.SyncFnDryRunError{Err: err}
```

This then allows the error handling to use `errors.As` and still grab the underlying error value.

```go
	output, err := h.collection.SyncFnDryrun(h.ctx(), oldDoc, syncDryRunPayload.Doc, docid, syncDryRunPayload.Function)
	if err != nil {
		var syncFnDryRunErr *base.SyncFnDryRunError
		if !errors.As(err, &syncFnDryRunErr) {
			return err
		}

		errMsg := syncFnDryRunErr.Error()
		if syncFnDryRunErr.Unwrap() != nil {
			errMsg = syncFnDryRunErr.Err.Error()
		}
		resp := SyncFnDryRun{
			Exception: errMsg,
		}
		h.writeJSON(resp)
		return nil
	}
```

It's a little more code, but less brittle than string comparisons and is more idiomatic (allows the error/type to play nicely with the stdlib errors package)

rest/diagnostic_doc_api_test.go
```diff
@@ -1079,3 +998,241 @@ func TestGetUserDocAccessDuplicates(t *testing.T) {
...
+		document        string
+		docID           string
+		existingDoc     bool
+		existingDocID   string
+		existingDocBody string
+		expectedOutput  SyncFnDryRun
+		expectedStatus  int
+	}{
+		{
+			name:           "test_case_1",
```
bbrks: It would be useful to have these test names reflect what is being tested for each subtest (e.g. For this one, since the provided sync function is identical to the database one, we can't determine which ran. What are we testing here?)

This helps us ensure we've got coverage of edge cases and makes it easier to figure out what is going on when these tests need to be looked at.

bbrks (changes requested):
Let me know if you want to pair up or talk through any of these comments - I'd be happy to

db/crud.go
RIT3shSapata: doc ID is required for the preprocessing of the newDoc. I've moved all the preprocessing inside the handler. I wanted to know if its okay to do all the preprocessing in the handler instead of the Dry run method.

db/crud.go
```diff
@@ -1694,90 +1694,43 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
...
+		syncRunner, err := channels.NewSyncRunner(ctx, syncFn, jsTimeout)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create sync runner: %v", err)
+		}
+		jsOutput, err := syncRunner.Call(ctx, mutableBody, string(oldDoc._rawBody), metaMap, syncOptions)
+		if err != nil {
+
+			return nil, fmt.Errorf("failed to create sync runner: %v", err)
```
bbrks: Does this need to be a typed `SyncFnDryRunError` as well? The error is a runtime error from the javascript execution not creating a sync runner.

It seems like we may end up with slightly different behaviour between a custom sync function and the actual sync function because of this.

rest/diagnostic_doc_api.go
```diff
@@ -70,20 +76,89 @@ func (h *handler) handleGetDocChannels() error {
 func (h *handler) handleSyncFnDryRun() error {
 	docid := h.getQuery("doc_id")
 
-	body, err := h.readDocument()
+	var syncDryRunPayload SyncFnDryRunPayload
+	err := h.readJSONInto(&syncDryRunPayload)
+	// Only require a valid JSON payload if docid is not provided.
+	// If docid is provided, the sync function will use the document from the bucket, and the payload is optional.
```
bbrks: These comments are from the old behaviour now and can be removed.

bbrks (changes requested):
This looks mostly good until the CBG-5061 refactor - mainly just one comment about the returned error type for a custom sync function and then we can get this merged.

rest/diagnostic_doc_api.go
```diff
@@ -70,20 +76,89 @@ func (h *handler) handleGetDocChannels() error {
...
+	if err != nil {
+		var syncFnDryRunErr *base.SyncFnDryRunError
+		if !errors.As(err, &syncFnDryRunErr) {
+			return err
+		}
+
+		errMsg := syncFnDryRunErr.Error()
+		if syncFnDryRunErr.Unwrap() != nil {
+			errMsg = syncFnDryRunErr.Err.Error()
+		}
```
bbrks: Stole comment from the import PR since I spotted it there first.

---

I don't think the `Unwrap` step isn't necessary since `Error()` on the typed error returns the underlying error (if present) - and if we have a nil typed error, we can use that without a problem since it returns `syncFnDryRunErrorPrefix` as a generic error message.


```suggestion
		errMsg := syncFnDryRunErr.Error()
```

bbrks (approved):
LGTM 👍 thanks!

PR #7934, opened 2025-12-18
CI: Add govulncheck / Bump main Go version

bbrks:
- Add `govulncheck` to CI workflow
- Bump to latest Go version in `main`
- Bump `golangci-lint` action and tool due to Go version upgrade
- Tag `activeReplicatorCommon` async reconnect function with `go:norace` directive - fails reliably after Go version change in GH Actions

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/212/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - uses: actions/setup-go@v5
         with:
           go-version: 1.25.5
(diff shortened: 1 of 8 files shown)


bbrks:
FYI this looks like a legitimate race in ActiveReplicator `arc.ctx` that may be more reliably detectable now? I'll check it out a little more tomorrow.

.github/workflows/ci.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       - uses: actions/checkout@v4
```
torcolvin: if you update other actions versions

checkout is now version 5, but it doesn't look like it has relevant changes for us

no need to change

torcolvin (dismissed):
LGTM

github now provide yaml anchors if we wanted to deduplicate the constants in the future https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#yaml-anchors-and-aliases

torcolvin (approved):

PR #7935, opened 2025-12-19
CBG-4412: Add ability to pass custom import filter into diagnostic API

RIT3shSapata:
CBG-4412

Describe your PR here...
- Upstream [PR](https://github.com/couchbase/sync_gateway/pull/7932) for CBG-4411
- Adds ability to pass custom import filter into diagnostic API
- Refactor unit tests

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/208/
- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/211/


rest/diagnostic_doc_api.go
```diff
@@ -183,20 +187,40 @@ func (h *handler) handleSyncFnDryRun() error {
...
-	body, err := h.readDocument()
-	if err != nil {
-		if docid == "" {
-			return fmt.Errorf("Error reading body: %s, no doc id provided for dry run", err)
-		}
+	var importFilterPayload ImportFilterDryRunPayload
+	err := h.readJSONInto(&importFilterPayload)
+	// Only require a valid JSON payload if docid is not provided.
+	// If docid is provided, the sync function will use the document from the bucket, and the payload is optional.
+	if err != nil && docid == "" {
```
bbrks: Same issue here as on the sync function PR - I don't think we should swallow this error when `docid` is empty since we'd be allowing this endpoint to run on invalid JSON.

rest/diagnostic_doc_api.go
```diff
@@ -183,20 +187,40 @@ func (h *handler) handleSyncFnDryRun() error {
...
+		return base.HTTPErrorf(http.StatusBadRequest, "Error reading sync function payload: %v", err)
```
bbrks:
```suggestion
		return base.HTTPErrorf(http.StatusBadRequest, "Error reading import filter payload: %v", err)
```

rest/diagnostic_doc_api.go
```diff
@@ -183,20 +187,40 @@ func (h *handler) handleSyncFnDryRun() error {
...
 	errorMsg := ""
 	if err != nil {
+		var importFilterDryRunErr *base.ImportFilterDryRunError
+		if !errors.As(err, &importFilterDryRunErr) {
+			return err
+		}
 		errorMsg = err.Error()
+		if importFilterDryRunErr.Unwrap() != nil {
+			errorMsg = importFilterDryRunErr.Err.Error()
+		}
```
bbrks: I don't think the `Unwrap` step isn't necessary since `Error()` on the typed error returns the underlying error (if present) - and if we have a nil typed error, we can use that without a problem since it returns `importFilterErrorPrefix` as a generic error message.


```suggestion
```

base/error.go
```diff
@@ -386,3 +386,28 @@ func (e *SyncFnDryRunError) Unwrap() error {
...
+}
+
+func (e *ImportFilterDryRunError) Error() string {
+	if e == nil {
+		return importFilterErrorPrefix
+	}
+	if e.Err == nil {
+		return importFilterErrorPrefix
+	}
+	return e.Err.Error()
```
bbrks: Would recommend prepending the prefix here so that it's clear what the underlying error is coming from.


```suggestion
	return importFilterErrorPrefix + ": " + e.Err.Error()
```

bbrks (changes requested):
Mostly looks good 👍

db/import.go
```diff
@@ -545,20 +545,43 @@ func (i *ImportFilterFunction) EvaluateFunction(ctx context.Context, doc Body, d
...
+// ImportFilterDryRun Runs a document through the import filter and returns a boolean and error
+func (db *DatabaseCollectionWithUser) ImportFilterDryRun(ctx context.Context, doc Body, importFn string) (bool, error) {
+
+	var shouldImport bool
+	if importFn == "" {
+		importFilter := db.importFilter()
+		if importFilter == nil {
+			return false, base.HTTPErrorf(http.StatusBadRequest, "No import filter specified")
+		}
+		output, err := importFilter.EvaluateFunction(ctx, doc, true)
```
bbrks: This is using the shared pool of ImportFilter which may cause contention with the real import process and the diagnostic API.

Would prefer to be able to run the function in a new instance like we did for the sync function.

db/import.go
```diff
@@ -545,20 +545,43 @@ func (i *ImportFilterFunction) EvaluateFunction(ctx context.Context, doc Body, d
...
-				doc = docInBucket.Body(ctx)
+
+// ImportFilterDryRun Runs a document through the import filter and returns a boolean and error
+func (db *DatabaseCollectionWithUser) ImportFilterDryRun(ctx context.Context, doc Body, importFn string) (bool, error) {
+
+	var shouldImport bool
+	if importFn == "" {
+		importFilter := db.importFilter()
+		if importFilter == nil {
+			return false, base.HTTPErrorf(http.StatusBadRequest, "No import filter specified")
```
bbrks: As discussed, this should be `return true, nil` since no filter function allows docs to be imported without error.

db/import.go
```diff
@@ -545,20 +545,43 @@ func (i *ImportFilterFunction) EvaluateFunction(ctx context.Context, doc Body, d
...
+			return false, &base.ImportFilterDryRunError{Err: err}
+		}
+		shouldImport = output
+	} else {
+		jsTimeout := time.Duration(base.DefaultJavascriptTimeoutSecs) * time.Second
+		importRunner, err := newImportFilterRunner(ctx, importFn, jsTimeout)
+		if err != nil {
+			return false, errors.New("failed to create import filter runner: " + err.Error())
+		}
+		importOuput, err := importRunner.Call(ctx, doc)
```
bbrks: Should probably error check here first (including the errors.As handling for an expected runtime error to return in the API) before we do the type switch. It's best practice/idiomatic to do error checking first before trying to use a returned value.

db/import.go
```diff
@@ -545,20 +545,40 @@ func (i *ImportFilterFunction) EvaluateFunction(ctx context.Context, doc Body, d
...
+			break
+		case string:
+			boolResult, err := strconv.ParseBool(result)
+			if err != nil {
+				return false, err
 			}
-		} else {
-			return false, err
+			shouldImport = boolResult
+			break
```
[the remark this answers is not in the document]
bbrks: 👍

db/import.go
```diff
@@ -545,20 +545,40 @@ func (i *ImportFilterFunction) EvaluateFunction(ctx context.Context, doc Body, d
...
+		}
+		shouldImport = output
+	} else {
+		jsTimeout := time.Duration(base.DefaultJavascriptTimeoutSecs) * time.Second
+		importRunner, err := newImportFilterRunner(ctx, importFn, jsTimeout)
+		if err != nil {
+			return false, errors.New("failed to create import filter runner: " + err.Error())
+		}
+		importOuput, err := importRunner.Call(ctx, doc)
+		switch result := importOuput.(type) {
```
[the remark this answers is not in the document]
bbrks: 👍

db/import.go
```diff
@@ -545,20 +545,43 @@ func (i *ImportFilterFunction) EvaluateFunction(ctx context.Context, doc Body, d
...
+	if importFn == "" {
+		importFilter := db.importFilter()
+		if importFilter == nil {
+			return false, base.HTTPErrorf(http.StatusBadRequest, "No import filter specified")
+		}
+		output, err := importFilter.EvaluateFunction(ctx, doc, true)
+		if err != nil {
+			return false, &base.ImportFilterDryRunError{Err: err}
+		}
+		shouldImport = output
```
bbrks: If we `return output` here, we can avoid the nested `} else {` block below since we're returning early.

This is more idiomatic Go than writing if else chains and deep nesting.

bbrks (changes requested):

rest/diagnostic_doc_api.go
```diff
@@ -195,24 +200,49 @@ func (h *handler) handleSyncFnDryRun() error {
...
 	if err != nil {
-		if docid == "" {
-			return fmt.Errorf("Error reading body: %s, no doc id provided for dry run", err)
-		}
+		return base.HTTPErrorf(http.StatusBadRequest, "Error reading import filter payload: %v", err)
 	}
 
-	if docid != "" && body != nil {
+	// Cannot pass both doc_id and body in the request body
+	if len(importFilterPayload.Doc) > 0 && docid != "" {
```
bbrks, line 210: Should we also check that at least one of these is set?

If both are empty, we're getting all the way into `ImportFilterDryRun` with a nil body and then returning `true` which is misleading/incorrect (we wouldn't import a nil document body since we require that all mobile-aware documents are [JSON objects](https://jira.issues.couchbase.com/browse/DOC-10952))

`TestImportFilterDryRunErrors` should be an easy place to drop the additional test case for no doc ID and no body.

rest/diagnostic_doc_api.go
```diff
@@ -195,24 +200,49 @@ func (h *handler) handleSyncFnDryRun() error {
...
 		return base.HTTPErrorf(http.StatusBadRequest, "doc body and doc id provided. Please provide either the body or a doc id for the import filter dry run")
 	}
-	shouldImport, err := h.collection.ImportFilterDryRun(h.ctx(), body, docid)
+
+	var doc db.Body
+	if docid != "" {
+		docInBucket, err := h.collection.GetDocument(h.ctx(), docid, db.DocUnmarshalSync)
```
bbrks: If we unmarshal the doc body as soon as possible with the Get operation, we can get better error reporting if the document is invalid JSON - since calling `Body()` below with invalid JSON will only warn and return `nil` as the function does not have an error returned.

```suggestion
		docInBucket, err := h.collection.GetDocument(h.ctx(), docid, db.DocUnmarshalAll)
```

rest/diagnostic_doc_api_test.go
```diff
@@ -1361,6 +1291,454 @@ func TestSyncFuncDryRunErrors(t *testing.T) {
...
+					Errors: []string{"This is a console.error log from doc.logerror"},
+					Info:   []string{"This is a console.log log from doc.loginfo", "one more info for good measure..."},
+				},
+			},
+			expectedStatus: http.StatusOK,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
```
bbrks: We have to use `rt.Run` instead of `t.Run` when using `RestTester`.

It ensures that the subtests use the correct `testing.T` reference when things fail. Not doing this can result in things like this: `subtest may have called FailNow on a parent test`

```suggestion
		rt.Run(test.name, func(t *testing.T) {
```

rest/diagnostic_doc_api_test.go
```diff
@@ -1361,6 +1291,454 @@ func TestSyncFuncDryRunErrors(t *testing.T) {
...
+			assert.Equal(t, test.expectedOutput, output)
+		})
+	}
+}
+
+func TestImportFilterDryRunErrors(t *testing.T) {
+	rt := NewRestTester(t, nil)
+	defer rt.Close()
+
+	// doc ID not found
```
bbrks, line 1736: As mentioned in another comment, adding a test case for no doc id and no doc body would be good here.

db/import.go
```diff
@@ -545,20 +549,43 @@ func (i *ImportFilterFunction) EvaluateFunction(ctx context.Context, doc Body, d
...
+	case bool:
+		shouldImport = result
+	case string:
+		boolResult, err := strconv.ParseBool(result)
+		if err != nil {
 			return false, err
 		}
+		shouldImport = boolResult
+	default:
+		return false, &base.ImportFilterDryRunError{Err: err}
```
bbrks: `err` is always nil here, but the real function returns `errors.New("Import filter function returned non-boolean value.")` and we should do the same for dry run.

Really this whole switch/case feels like a good chunk of code to pull into a separate function to be used in both places.

db/import.go
bbrks: This function still has the `dryRun` parameter from the old implementation that calls into the shared JSServer instance, but it's never `true` and should be completely removed.

bbrks (changes requested):
Mostly looks good. See inline comments for specific changes/comments.

bbrks (approved):
LGTM 👍 thanks!

PR #7949, opened 2026-01-07
CBG-5052 send noRev if history can't be parsed

torcolvin:
CBG-5052 send noRev if history can't be parsed

A document with the given revtree will panic:

```mermaid
graph TD
    1a["1-a"] --> 2b["2-b"]
    2b --> 3d["3-d"]
    3d --> 3c["3-c"]
```

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/234/


rest/blip_api_crud_test.go
```diff
@@ -3549,3 +3551,182 @@ func TestTombstoneCount(t *testing.T) {
...
+					"3-c",
+					"1-a",
+					"2-b"
+				]
+			},
+			"rev": "3-c",
+			"sequence": {{.seq}},
+			"value_crc32c": "expand"
+		}`
+		tmpl := template.Must(template.New("badSync").Option("missingkey=error").Parse(badSyncRaw))
```
bbrks, line 3598: Templating seems pretty overkill for one single-use sequence value rather than `fmt.Sprintf` - but it's done now and this is very much a nitpick.

rest/blip_api_crud_test.go
```diff
@@ -3549,3 +3551,182 @@ func TestTombstoneCount(t *testing.T) {
 	})
 
 }
+
+func TestBlipNoRevOnCorruptHistory(t *testing.T) {
+	base.SetUpTestLogging(t, base.LevelTrace, base.KeyHTTP, base.KeySync, base.KeySyncMsg)
```
bbrks: Guessing this was used to debug - can probably remove altogether, definitely shouldn't remain as trace though

```suggestion
```

rest/blip_api_crud_test.go
```diff
@@ -3549,3 +3551,182 @@ func TestTombstoneCount(t *testing.T) {
...
+			Username: user,
+		})
+		defer btc.Close()
+
+		ctx := rt.Context()
+		seq, err := rt.GetDatabase().NextSequence(ctx)
+		require.NoError(t, err)
+		// document contains an invalid revtree
+		//
+		// 3-c is a child of 3-d, two revisions of the same generation can not exist in a non branched revtree
```
bbrks: more specific description of why this revtree is wrong - it's not really related to branched/unbranched revtrees

```suggestion
		// 3-c is a child of 3-d, a revision must be one or more generations higher than it its parent, not equal
```

rest/blip_api_crud_test.go
```diff
@@ -3549,3 +3551,182 @@ func TestTombstoneCount(t *testing.T) {
...
+		ctx := rt.Context()
+		docID := SafeDocumentName(t, t.Name())
+		docRev1 := rt.CreateDocNoHLV(docID, db.Body{"delta": true, "channels": []string{channelA}})
+		btcRunner.StartOneshotPull(btc.id)
+		btcRunner.WaitForVersion(btc.id, docID, DocVersion{RevTreeID: docRev1.GetRevTreeID()})
+		seq, err := rt.GetDatabase().NextSequence(ctx)
+		require.NoError(t, err)
+		// document contains an invalid revtree
+		//
+		// 3-c is a child of 3-d, two revisions of the same generation can not exist in a non branched revtree
```
bbrks:
```suggestion
		// 3-c is a child of 3-d, a revision must be one or more generations higher than it its parent, not equal
```

base/util_testing.go
```diff
@@ -1030,3 +1030,10 @@ func TestRequiresViews(t testing.TB) {
 		t.Skip("Skipping test - views not supported with this bucket configuration")
 	}
 }
+
+// TestRequiresDeltaSync will skip the current test if not running with EE.
+func TestRequiresDeltaSync(t testing.TB) {
+	if !IsEnterpriseEdition() {
+		t.Skipf("Skipping test due to running in EE mode")
```
bbrks: This skip message is the opposite of what is happening

```suggestion
		t.Skipf("Skipping test - Delta Sync requires EE")
```

db/blip_handler.go
```diff
@@ -927,14 +927,26 @@ func (bsc *BlipSyncContext) sendRevAsDelta(ctx context.Context, sender *blip.Sen
 	if redactedRev != nil {
 		var history []string
 		var revTreeProperty []string
-		if !bsc.useHLV() {
-			history = toHistory(redactedRev.History, knownRevs, maxHistory)
+		if !bsc.useHLV() || bsc.sendRevTreeProperty() {
+			var err error
+			history, err = toHistory(redactedRev.History, knownRevs, maxHistory)
+			if err != nil {
+				err := base.RedactErrorf("Could not get rev tree history for replacement rev in sendRevAsDelta %s %s: %w, sending a noRev to skip this revision for replication at sequence %d.", base.UD(docID), revID, err, seq)
```
bbrks: `%d` is the wrong verb for seq here. Should be `%s` so we get compound sequences hidden if zero.

```
... replication at sequence {0 0 3}. -- db.(*BlipSyncContext).sendRevAsDelta() at blip_handler.go:935
```

```suggestion
				err := base.RedactErrorf("Could not get rev tree history for replacement rev in sendRevAsDelta %s %s: %w, sending a noRev to skip this revision for replication at sequence %s.", base.UD(docID), revID, err, seq)
```

db/blip_handler.go
```diff
@@ -927,14 +927,26 @@ func (bsc *BlipSyncContext) sendRevAsDelta(ctx context.Context, sender *blip.Sen
...
+			}
 		} else {
 			history = append(history, redactedRev.HlvHistory)
 		}
 		if bsc.sendRevTreeProperty() {
 			revTreeProperty = append(revTreeProperty, redactedRev.RevID)
-			revTreeProperty = append(revTreeProperty, toHistory(redactedRev.History, knownRevs, maxHistory)...)
+			history, err := toHistory(redactedRev.History, knownRevs, maxHistory)
+			if err != nil {
+				err := base.RedactErrorf("Could not get rev tree history for replacement rev when sending in sendRevAsDelta %s %s: %w, sending a noRev to skip this document for replication at sequence %d.", base.UD(docID), redactedRev.RevID, err, seq)
```
bbrks:
```suggestion
				err := base.RedactErrorf("Could not get rev tree history for replacement rev when sending in sendRevAsDelta %s %s: %w, sending a noRev to skip this document for replication at sequence %s.", base.UD(docID), redactedRev.RevID, err, seq)
```

db/blip_sync_context.go
```diff
@@ -768,7 +768,12 @@ func (bsc *BlipSyncContext) sendRevision(ctx context.Context, sender *blip.Sende
 	}
 	var history []string
 	if !bsc.useHLV() || localIsLegacyRev {
-		history = toHistory(docRev.History, knownRevs, maxHistory)
+		var err error
+		history, err = toHistory(docRev.History, knownRevs, maxHistory)
+		if err != nil {
+			err := base.RedactErrorf("Could not get rev tree history for %s %s: %w, sending a noRev to skip this revision for replication at sequence %d.", base.UD(docID), revID, err, seq)
```
bbrks:
```suggestion
			err := base.RedactErrorf("Could not get rev tree history for %s %s: %w, sending a noRev to skip this revision for replication at sequence %s.", base.UD(docID), revID, err, seq)
```

db/blip_sync_context.go
```diff
@@ -791,7 +801,13 @@ func (bsc *BlipSyncContext) sendRevision(ctx context.Context, sender *blip.Sende
 		// should only be sent if the local + remote revisions are not a legacy revisions as the rev tree in this case will be
 		// sent in history property.
 		revTreeHistoryProperty = append(revTreeHistoryProperty, docRev.RevID) // we need current rev
-		revTreeHistoryProperty = append(revTreeHistoryProperty, toHistory(docRev.History, knownRevs, maxHistory)...)
+		history, err := toHistory(docRev.History, knownRevs, maxHistory)
+		if err != nil {
+			err := base.RedactErrorf("Could not get rev tree history for %s %s when local and remote revision are hlv aware: %w, sending a noRev to skip this revision for replication at sequence %d.", base.UD(docID), revID, err, seq)
```
bbrks:
```suggestion
			err := base.RedactErrorf("Could not get rev tree history for %s %s when local and remote revision are hlv aware: %w, sending a noRev to skip this revision for replication at sequence %s.", base.UD(docID), revID, err, seq)
```

db/blip_sync_context.go
```diff
@@ -818,9 +834,14 @@ func digests(meta []AttachmentStorageMeta) []string {
...
 }
 
-func toHistory(revisions Revisions, knownRevs map[string]bool, maxHistory int) []string {
+// toHistory returns an array of revision IDs in a descending array. Truncates array at maxHistory length, or at the first rev in knownRevs. Returns an error with invalid history.
+func toHistory(revisions Revisions, knownRevs map[string]bool, maxHistory int) ([]string, error) {
 	// Get the revision's history as a descending array of ancestor revIDs:
-	history := revisions.ParseRevisions()[1:]
+	revs := revisions.ParseRevisions()
+	if len(revs) == 0 {
+		return nil, fmt.Errorf("invalid revision history, no revisions found")
```
bbrks: This is misleading - there are revisions, but they cannot be parsed and get returned as empty up to this point.

It feels like this may be the wrong place to create the error?

rest/blip_api_crud_test.go
bbrks, line 3556: I can't immediately see the source of the flake nor can I reproduce, but maybe enabling logging has avoided it due to timing?

bbrks (changes requested):

db/blip_sync_context.go
```diff
@@ -765,7 +765,12 @@ func (bsc *BlipSyncContext) sendRevision(ctx context.Context, sender *blip.Sende
 	}
 	var history []string
 	if !bsc.useHLV() || localIsLegacyRev {
-		history = toHistory(docRev.History, knownRevs, maxHistory)
+		var err error
+		history, err = toHistory(docRev.History, knownRevs, maxHistory)
+		if err != nil {
+			err := base.RedactErrorf("Could not get rev tree history for %s %s: %w, sending a noRev to skip this revision for replication at sequence %d.", base.UD(docID), revID, err, seq)
```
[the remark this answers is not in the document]
torcolvin: I don't agree with these comments because I actually think it is better to overwrite the variable but I'm curious what @bbrks thinks.

rest/blip_api_crud_test.go
torcolvin: I copied this from the delta case where I had to use a lot of variables and I kept screwing up Sprintf but I agree with this.

rest/blip_api_crud_test.go
torcolvin: I can't reproduce this but I'm inclined to leave this logging in when inevitably mobile jenkins reproduces this problem. It's pretty light on logging in general with restricted log keys.

db/blip_sync_context.go
torcolvin: I am open to doing something different here, I don't know the right way to handle this error.

I would like to catch the error earlier, but it actually happens when reading the document. If I fail to unmarshal the document, then I think REST api calls would fail in addition to blip. I’m not sure what the right answer is.
The abbreviated stack trace is:

* revCacheLoaderForDocument https://github.com/couchbase/sync_gateway/blob/f57d57b52f6585d55edee399244750a13d638913/db/revision_cache_interface.go#L461
* https://github.com/couchbase/sync_gateway/blob/f57d57b52f6585d55edee399244750a13d638913/db/revtree.go#L832 is where this problem occurs because Start=3 digests=[c d b a] . Potentially logging here is the right thing to do, but then it is part of rev cache loader and we don’t know why the document is loaded.
* then it really is mostly a problem for blip because this is the only case where revisions.ParseRevisions is called because it returns an empty ids https://github.com/couchbase/sync_gateway/blob/f57d57b52f6585d55edee399244750a13d638913/db/revtree.go#L822 because len(digests) is 4 and Start is 3


I don’t feel like it is totally correct to fail to unmarshal the document but I also don’t know how to “fix” the history. I don’t know if this is the only place the history is wrong. Additionally, bubbling the error up the caller is hard because it is in unmarshal and if I catch errors in `encodeRevisions` then there are a bunch of cases that handling this is OK.

The idea is that in the rare case this exists, this document would be skipped for replication and then it could be analyzed with `/ks/_raw/doc` endpoint. It's not even clear how to fix this problem exactly because you need to purge the document and recreate it and has to be done on a case by case basis for the "right" way to resolve the issue in replication.

The idea is to have an error like this not halt the entire pull replication in the case there are one or two bad documents. We do not know how documents like this were created in the first place.

bbrks:
```suggestion
		return nil, fmt.Errorf("invalid revision history")
```

bbrks: will keep as-is but tweak error message

bbrks (approved):

PR #7950, opened 2026-01-08
CBD-6543 bump build repo SHA to pick up recent changes to pyinstaller

mhocouchbase:
CBD-6543

Describe your PR here...
-Bump build repo SHA to pick up recent changes to pyinstaller

## Pre-review checklist

- [ ] Link upstream PRs

### DIFF
--- a/manifest/3.3.xml
+++ b/manifest/3.3.xml
@@ -17,7 +17,7 @@ licenses/APL2.txt.
 
     <!-- Build Scripts (required on CI servers) -->
     <project name="product-texts" path="product-texts" remote="couchbase"/>
-    <project name="build" path="cbbuild" remote="couchbase" revision="b1dc2359603ab411ce0788854ea879ce804c9055">
+    <project name="build" path="cbbuild" remote="couchbase" revision="0ae8f452cece1e1f48f99e370a103b55bfea2dbe">
         <annotation name="VERSION" value="3.3.3" keep="true"/>
         <annotation name="BLD_NUM" value="@BLD_NUM@" keep="true"/>
         <annotation name="RELEASE" value="@RELEASE@" keep="true"/>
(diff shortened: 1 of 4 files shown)


torcolvin (commented):
We typically copy manifests from previous versions and technically 3.0.xml and 3.1.xml are not end of life until next month. 3.2.xml is an active maintenance release.

If changing this xml file will cause a new build, then we can wait to bump those values, but if it doesn't, let's change them now because it would be required by the build system.

mhocouchbase:
> We typically copy manifests from previous versions and technically 3.0.xml and 3.1.xml are not end of life until next month. 3.2.xml is an active maintenance release.
> 
> If changing this xml file will cause a new build, then we can wait to bump those values, but if it doesn't, let's change them now because it would be required by the build system.

It is going to trigger new builds.  let's wait until 3.0.xml and 3.1.xml reach EOL.  I modified the build jobs to check out build repo separately at the moment.  So, it is safe to leave this open for a while.

torcolvin:
> We typically copy manifests from previous versions and technically 3.0.xml and 3.1.xml are not end of life until next month. 3.2.xml is an active maintenance release.
> 
> If changing this xml file will cause a new build, then we can wait to bump those values, but if it doesn't, let's change them now because it would be required by the build system.

I think then let's change 3.3.xml, 4.0.xml, and default xml since 3.3.2.2 was linux only for capella, and then we can close this ticket.

bbrks:
I'm OK merging this for 4.0.3, 3.3.3 and 4.1.0

3.3.2.2 we've already delivered the relevant builds to Capella and we don't expect anybody else to be picking up this version.
If we remove the 3.3.2.2 change we can just roll forwards with the changes from the above unreleased versions.

torcolvin:
Picked up these changes into a separate PR https://github.com/couchbase/sync_gateway/pull/7977.

PR #7954, opened 2026-01-14
CBG-5089: Add support for user defined metadata in diagnostic api

RIT3shSapata:
CBG-5089

Describe your PR here...
- Added ability for user to pass the user defined xattrs into Sync Function Diagnostic API

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/221/

### DIFF
--- a/rest/diagnostic_doc_api.go
+++ b/rest/diagnostic_doc_api.go
@@ -104,9 +104,13 @@ func (h *handler) handleSyncFnDryRun() error {
 		if !exists {
 			return base.HTTPErrorf(http.StatusBadRequest, "Missing xattrs in meta")
 		}
-		_, exists = xattrs[h.collection.UserXattrKey()]
+		userXattrKey := h.collection.UserXattrKey()
+		if userXattrKey == "" {
+			return base.HTTPErrorf(http.StatusBadRequest, "no user xattr key configured for this database")
+		}
+		_, exists = xattrs[userXattrKey]
 		if !exists {
-			return base.HTTPErrorf(http.StatusBadRequest, "xattr key passed does not match db defined xattr key")
+			return base.HTTPErrorf(http.StatusBadRequest, "configured user xattr key %q not found in provided xattrs", userXattrKey)
 		}
 	}
 
--- a/rest/diagnostic_doc_api.go
+++ b/rest/diagnostic_doc_api.go
@@ -40,10 +40,13 @@ type ImportFilterDryRun struct {
 	Logging      DryRunLogging `json:"logging"`
 }
 
+type SyncFnDryRunMetaMap struct {
+	Xatrrs map[string]any `json:"xatrrs"`
+}
 type SyncFnDryRunPayload struct {
-	Function string         `json:"sync_function"`
-	Doc      db.Body        `json:"doc,omitempty"`
-	Meta     map[string]any `json:"meta,omitempty"`
+	Function string              `json:"sync_function"`
+	Doc      db.Body             `json:"doc,omitempty"`
+	Meta     SyncFnDryRunMetaMap `json:"meta,omitempty"`
 }
 
 type ImportFilterDryRunPayload struct {
@@ -98,20 +101,20 @@ func (h *handler) handleSyncFnDryRun() error {
 		return base.HTTPErrorf(http.StatusBadRequest, "no doc_id or document provided")
 	}
 
+	var userXattrs map[string]any
 	// checking user defined metadata
-	if syncDryRunPayload.Meta != nil {
-		xattrs, exists := syncDryRunPayload.Meta["xattrs"].(map[string]any)
-		if !exists {
-			return base.HTTPErrorf(http.StatusBadRequest, "Missing xattrs in meta")
-		}
+	if syncDryRunPayload.Meta.Xatrrs != nil {
+		xattrs := syncDryRunPayload.Meta.Xatrrs
 		userXattrKey := h.collection.UserXattrKey()
 		if userXattrKey == "" {
 			return base.HTTPErrorf(http.StatusBadRequest, "no user xattr key configured for this database")
 		}
-		_, exists = xattrs[userXattrKey]
+		_, exists := xattrs[userXattrKey]
 		if !exists {
 			return base.HTTPErrorf(http.StatusBadRequest, "configured user xattr key %q not found in provided xattrs", userXattrKey)
 		}
+		userXattrs = make(map[string]any)
+		userXattrs["xattrs"] = syncDryRunPayload.Meta.Xatrrs
 	}
 
 	oldDoc := &db.Document{ID: docid}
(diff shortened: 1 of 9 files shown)


rest/diagnostic_doc_api.go
```diff
@@ -97,6 +98,18 @@ func (h *handler) handleSyncFnDryRun() error {
...
 
+	// checking user defined metadata
+	if syncDryRunPayload.Meta != nil {
+		xattrs, exists := syncDryRunPayload.Meta["xattrs"].(map[string]any)
+		if !exists {
+			return base.HTTPErrorf(http.StatusBadRequest, "Missing xattrs in meta")
+		}
+		_, exists = xattrs[h.collection.UserXattrKey()]
+		if !exists {
+			return base.HTTPErrorf(http.StatusBadRequest, "xattr key passed does not match db defined xattr key")
```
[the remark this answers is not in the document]
RIT3shSapata: Hey @bbrks , would like your opinion on this, the error handling suggested by copilot is also good, do you think I should change the current code to this?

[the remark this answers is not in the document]
bbrks: Yeah I think the suggested error handling is good - it at least informs the user more accurately of an unset xattr key

rest/diagnostic_doc_api_test.go
```diff
@@ -1255,8 +1349,16 @@ func TestSyncFuncDryRun(t *testing.T) {
 				RequireStatus(t, rt.SendAdminRequest(http.MethodDelete, "/{{.keyspace}}/_config/sync", ""), http.StatusOK)
 			}
 
+			if test.xattrKey != "" {
+				RequireStatus(t, rt.SendAdminRequest(http.MethodPost, "/{{.db}}/_config", fmt.Sprintf(`{"user_xattr_key": "%s"}`, test.xattrKey)), http.StatusCreated)
```
bbrks, line 1408: This is going to leave the database in a slightly different state (with `user_xattr_key`) for any subsequent subtests which isn't great.

It's probably worth moving this to a new more targeted user xattr test which is EE only which sets up the database once.

At least then you still get test coverage of the other functionality in CE mode and without the subtest db config update

rest/diagnostic_doc_api_test.go
```diff
@@ -1278,9 +1380,21 @@ func TestSyncFuncDryRun(t *testing.T) {
 }
 
 func TestSyncFuncDryRunErrors(t *testing.T) {
-	rt := NewRestTester(t, nil)
+	if !base.IsEnterpriseEdition() {
+		t.Skipf("Requires EE for some config properties")
```
bbrks: Likewise - if we have a subset of a feature EE-only, either make the specific assertions run in EE mode only as a conditional, or new test for that sub-feature.

bbrks (changes requested):
Probably move the user xattr test cases to a test that is testing that specific sub-feature, so all the other tests can still run in CE mode.

rest/diagnostic_doc_api.go
```diff
@@ -97,6 +98,22 @@ func (h *handler) handleSyncFnDryRun() error {
 		return base.HTTPErrorf(http.StatusBadRequest, "no doc_id or document provided")
 	}
 
+	// checking user defined metadata
+	if syncDryRunPayload.Meta != nil {
```
bbrks: This isn't ensuring there's **nothing else** set in the map (e.g. `meta.foo`)

I think that's mostly why I prefer the struct approach, though it also makes access easier since there's less type assertions too:

```go
	Meta *SyncFnDryRunMeta `json:"meta,omitempty"`
}

type SyncFnDryRunMeta struct {
	Xattrs map[string]any `json:"xattrs,omitempty"`
}
```

We'd also need to length-check the Xattrs map to ensure only exactly 1 (matching) key is present as well - it's too easy to unintentionally pass extra data into the sync function otherwise that wouldn't be available in the real one.

e.g.

`meta.xattrs[my_user_xattr_key]={"my_xattr_data":true}`
`meta.xattrs[bar]="buzz"`

dry run could technically still be able to access `meta.xattrs.bar` despite not being available in a real sync function.

bbrks (changes requested):
Need a bit more validation around the `meta` object being passed into the request really. See inline.

rest/diagnostic_doc_api.go
```diff
@@ -97,6 +101,22 @@ func (h *handler) handleSyncFnDryRun() error {
...
+		userXattrKey := h.collection.UserXattrKey()
+		if userXattrKey == "" {
+			return base.HTTPErrorf(http.StatusBadRequest, "no user xattr key configured for this database")
+		}
+		_, exists := xattrs[userXattrKey]
+		if !exists {
+			return base.HTTPErrorf(http.StatusBadRequest, "configured user xattr key %q not found in provided xattrs", userXattrKey)
+		}
+		userXattrs = make(map[string]any)
+		userXattrs["xattrs"] = syncDryRunPayload.Meta.Xattrs
```
bbrks:
```suggestion
		userXattrs["xattrs"] = xattrs
```

rest/diagnostic_doc_api.go
```diff
@@ -97,6 +101,22 @@ func (h *handler) handleSyncFnDryRun() error {
 		return base.HTTPErrorf(http.StatusBadRequest, "no doc_id or document provided")
 	}
 
+	var userXattrs map[string]any
+	// checking user defined metadata
+	if syncDryRunPayload.Meta.Xattrs != nil {
+		xattrs := syncDryRunPayload.Meta.Xattrs
```
bbrks, line 107: Still missing the length check on xattrs. Without this, users could still be passing extra data into the sync function as long as one of them matched the configured xattr key:

```json
"xattrs": {
  "my_real_xattr_key": {},
  "something_else": {}
}
```

```suggestion
		xattrs := syncDryRunPayload.Meta.Xattrs
		if len(xattrs) > 1 {
			return base.HTTPErrorf(http.StatusBadRequest, "Only one xattr key can be specified in meta")
		}
```

rest/diagnostic_doc_api_test.go
```diff
@@ -1293,6 +1450,26 @@ func TestSyncFuncDryRunErrors(t *testing.T) {
...
+
+	dbConfig := rt.NewDbConfig()
+	dbConfig.Name = "db1"
+	dbConfig.UserXattrKey = base.Ptr("channelXattrs")
+
+	RequireStatus(t, rt.CreateDatabase("db1", dbConfig), http.StatusCreated)
+	// Invalid meta body
+	RequireStatus(t, rt.SendDiagnosticRequest(http.MethodPost, "/{{.keyspace}}/_sync", `{"meta":{"foo": "bar"}}`), http.StatusBadRequest)
+	// Invalid meta body
+	RequireStatus(t, rt.SendDiagnosticRequest(http.MethodPost, "/{{.keyspace}}/_sync", `{"meta":{"xattrs": {"foo": "bar"}}}`), http.StatusBadRequest)
```
bbrks: Add a test for 2 xattr keys (one matching the configured name)

bbrks (changes requested):

bbrks (approved):

PR #7959, opened 2026-01-15
CBG-5102: don't add backup rev loads to opposite lookup map that you have fetched by

gregns1:
CBG-5102

-  Don't revs to to cv lookup map when the revision you have loaded from the bucket is not current 
- Same inverse, don't add to revID lookup map when CV you have fetched is not current 
- There is an issue with history being sent correctly for legacy revs on delta sync, this will be addressed separately
- Also removed code to add legacy docs that don't yet have a CV assigned to the CV lookup map by generating encoded cv for them. This is problematic given we end up with rev value with CV defined but not reflected in the bucket on the document. This was added when we designed the rev cache to always be in sync between lookup maps, now that isn't the case we can safely remove this. 
- Removed test `TestIncorrectStatsWhenAddingToLookupMapKeyThatExists` given that tests stats around the above bullet point that has been removed

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/225/ (known flake)

### DIFF
--- a/db/blip_sync_context.go
+++ b/db/blip_sync_context.go
@@ -588,7 +588,7 @@ func (bsc *BlipSyncContext) sendDelta(ctx context.Context, sender *blip.Sender, 
 
 	var history []string
 	var revTreeProperty []string
-	// now CV can be empty on rev cache items, only by CV when its available
+	// CV can now be empty on rev cache items; only use CV when it's available
 	// history being sent here is potentially incorrect pending CBG-5106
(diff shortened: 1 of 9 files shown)


bbrks (dismissed):
This looks good - I agree with the Copilot nitpicks - but am OK merging as-is.

torcolvin (approved):

PR #7960, opened 2026-01-15
Add POC dockertest-based upgradetest package

bbrks:
Definitely a few things in here that need cleanup/refining/tweaking but open to general opinions of this implementation.

Future expansion:
- Allow a special value which represents the current branch/codebase (to test upgrading to in-development SG versions before GA)

`TestRollingUpgradeWithCRUD` currently fails since 3.x can produce sporadic 500 errors when reading 4.x metadata - due to now broken sync metadata assertion


bbrks:
will reopen when/if we're ready for in-repo upgrade testing

PR #7963, opened 2026-01-15
CBG-5109 Fix & vet log *fCtx function arguments

torcolvin:
CBG-5109 Fix & vet log *fCtx function arguments

As per the documentation in https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/printf:

> To enable printf checking on a function that is not found by this analyzer's heuristics (for example, because control is obscured by dynamic method calls), insert a bogus call:

> ```func MyPrintf(format string, args ...any) {
	if false {
		_ = fmt.Sprintf(format, args...) // enable printf checking
	}
	...
}```

- add bogus calls to allow go vet (and hence golangci-lint) to catch errors.
- Fix type errors in logger functions and require a constant format string.

### DIFF
--- a/base/logging.go
+++ b/base/logging.go
@@ -143,6 +143,10 @@ func PanicfCtx(ctx context.Context, format string, args ...any) {
 
 // FatalfCtx logs the given formatted string and args to the error log level and given log key and then exits.
 func FatalfCtx(ctx context.Context, format string, args ...any) {
+	// no-op call to enable golang.org/x/tools/go/analysis/passes/printf checking in go vet
+	if false {
+		_ = fmt.Sprintf(format, args...)
+	}
 	// Fall back to stdlib's log.Panicf if SG loggers aren't set up.
 	if errorLogger.Load() == nil {
 		log.Fatalf(format, args...)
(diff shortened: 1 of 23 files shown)


base/logging.go
```diff
@@ -193,6 +221,10 @@ func RecordStats(statsJson string) {
 // logTo is the "core" logging function. All other logging functions (like Debugf(), WarnfCtx(), etc.) end up here.
 // The function will fan out the log to all of the various outputs for them to decide if they should log it or not.
 func logTo(ctx context.Context, logLevel LogLevel, logKey LogKey, format string, args ...any) {
+	// no-op call to enable golang.org/x/tools/go/analysis/passes/printf checking in go vet
+	if false {
+		_ = fmt.Sprintf(format, args...)
+	}
```
bbrks, line 215: Forcing the check just once here in `logTo` allows basically all of the other logging functions to still be checked - there's no need to define this in every place.

It greatly reduces the impact of the noise of enabling this check (which I think is extremely valuable)

There are still some exceptions, like the `RedactableError` and test logger - but otherwise this looks good.

bbrks (changes requested):
Nice that it found all of these legitimate formatting issues - completely gross way of having to force the check :(

I think we can minimise the impact by just adding the noop to the lowest level of the logging callstack?

base/logging.go
torcolvin: Does this work? It didn't work for me because I thought it only checks one layer of functions?

I would be happy if you were to play with this (but not too long) because maybe I just missed something.

`go vet` plus `-printfuncs` argument which didn't seem to work for this. I did try to add logging to a copy of [golang.org/x/tools/go/analysis/passes/printf](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/printf) but I wasn't able to determine why it couldn't detect the functions.

I did wonder if renaming the functions InfoCtxf or just Infof would fix the problem but that would make backports awful. Though that would align with the naming conventions.

torcolvin:
Do you think this PR is worth backporting once we sort out the details?

base/logging.go
bbrks: I tried it locally with just `go vet ./base` and it worked. I couldn't push it any further down though. Idk if there's a limit on how far it checks?

bbrks:
Yes I think so - at the very least the actual fixes to the wrong formatting directives

base/logging.go
torcolvin: Thanks! I really thought I verified this before but it seems to work fine like this.

Adding to TestBucketPool.Fatalf is required because of the `addPrefixes` call.

bbrks (approved):

PR #7999, opened 2026-01-21
[3.3.3 backport] CBG-5116: Diagnostic API Improvements

RIT3shSapata:
CBG-5116

Describe your PR here...
- This PR contains all the commits of Diagnostic API

## Pre-review checklist

- [ ] Link upstream PRs

### DIFF
--- a/db/import.go
+++ b/db/import.go
@@ -238,13 +238,13 @@ func (db *DatabaseCollectionWithUser) importDoc(ctx context.Context, docid strin
 
 			if isDelete && body == nil {
 				deleteBody := Body{BodyDeleted: true}
-				shouldImport, err = importFilter.EvaluateFunction(ctx, deleteBody, false)
+				shouldImport, err = importFilter.EvaluateFunction(ctx, deleteBody)
 			} else if isDelete && body != nil {
 				deleteBody := body.ShallowCopy()
 				deleteBody[BodyDeleted] = true
-				shouldImport, err = importFilter.EvaluateFunction(ctx, deleteBody, false)
+				shouldImport, err = importFilter.EvaluateFunction(ctx, deleteBody)
 			} else {
-				shouldImport, err = importFilter.EvaluateFunction(ctx, body, false)
+				shouldImport, err = importFilter.EvaluateFunction(ctx, body)
 			}
 
 			if err != nil {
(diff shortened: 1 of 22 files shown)


db/import.go
```diff
@@ -499,16 +503,43 @@ func NewImportFilterFunction(ctx context.Context, fnSource string, timeout time.
 }
 
 // Calls a jsEventFunction returning an interface{}
-func (i *ImportFilterFunction) EvaluateFunction(ctx context.Context, doc Body, dryRun bool) (bool, error) {
+func (i *ImportFilterFunction) EvaluateFunction(ctx context.Context, doc Body) (bool, error) {
 
 	result, err := i.Call(ctx, doc)
 	if err != nil {
-		if !dryRun {
-			base.WarnfCtx(ctx, "Unexpected error invoking import filter for document %s - processing aborted, document will not be imported.  Error: %v", base.UD(doc), err)
```
bbrks, line 507: I've just noticed we're removing this warning logging and relying on the caller which only logs at info. Was this intentional? It seems like customers may miss this message at info level.

db/import.go
```diff
@@ -518,26 +549,7 @@ func (i *ImportFilterFunction) EvaluateFunction(ctx context.Context, doc Body, d
 		}
 		return boolResult, nil
 	default:
-		if !dryRun {
-			base.WarnfCtx(ctx, "Import filter function returned non-boolean result %v Type: %T", result, result)
```
bbrks, line 522: Likewise, this warning going means that customers won't see this since the caller logs errors at info level.

rest/audit_test.go
```diff
@@ -483,6 +484,34 @@ func TestAuditLoggingFields(t *testing.T) {
...
+					Doc: db.Body{
+						"accessChannel": []string{"dynamicChan5412"},
+						"accessUser":    "user",
+						"channel":       []string{"dynamicChan222"},
+						"expiry":        10,
+					},
+				}
+				bodyBytes, err := json.Marshal(request)
+				require.NoError(t, err)
+				RequireStatus(t, rt.SendDiagnosticRequest(http.MethodPost, "/{{.keyspace}}/_sync", string(bodyBytes)), http.StatusOK)
```
bbrks, line 509: Just needs the extra auth test commit pulled into this PR branch from #8006

bbrks (changes requested):
Looks fine but I have a few questions (that may also affect `main`/`4.0.3`)

db/import.go
RIT3shSapata: We do return the error back to the handler, which will send then return this error as part of exception in the response body

db/import.go
RIT3shSapata: Similarly, the error will be sent back to the user as a exception in the response

db/import.go
bbrks: As discussed - there is no request here to get back to a user. An import is sourced from a doc coming over the DCP feed to Sync Gateway. Warnings are the only way of communicating this error back.

bbrks:
See #8017 for restoration of warning logs - we can cherry-pick that into this

Edit: I'll follow up with this cherry-pick so you can review independently.

bbrks (approved):

PR #7996, opened 2026-01-20
CBG-5113: Add diagnostic server audit event

RIT3shSapata:
CBG-5113

Describe your PR here...
- Add Diagnostic Server audit event

## Pre-review checklist
- [x] Removed debug logging (`fmt.Print`, `log.Print`, ...)
- [x] Logging sensitive data? Make sure it's tagged (e.g. `base.UD(docID)`, `base.MD(dbName)`)
- [x] Updated relevant information in the API specifications (such as endpoint descriptions, schemas, ...) in `docs/api`

## Dependencies (if applicable)
- [ ] Link upstream PRs
- [ ] Update Go module dependencies when merged

## [Integration Tests](https://jenkins.sgwdev.com/job/SyncGatewayIntegration/build?delay=0sec)
- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/0000/

PR #8000, opened 2026-01-21
CBG-5129: Enable exhaustive linter

torcolvin:
Enable exhaustive linter to find missing uses of enums in switch statements.

This did find https://github.com/couchbase/sync_gateway/pull/7996 so seems useful. The other cases do not seem critical to fix since `gocb.LogMaxVerbosity` exists but is unused and `LogLevelEnabled` is only called at this time with Info or lower.

### DIFF
--- a/.golangci-strict.yml
+++ b/.golangci-strict.yml
@@ -9,6 +9,7 @@
 version: "2"
 linters:
   enable:
+    - exhaustive
     - gocritic # The most opinionated Go source code linter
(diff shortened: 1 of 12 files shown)


torcolvin:
I dropped the checking of the jwt algorithms since I did not have time to test enabling EdDSA, and this can be done as a separate ticket. Adding `//exhasutive:enforce` there may be worth considering.

bbrks (approved):

PR #8044, opened 2026-01-29
CBG-5070: Fix _mou not being cleared when tombstoning document 

RIT3shSapata:
CBG-5070

Describe your PR here...
- Fixed the scenario where the _mou is not being cleared when a document is tombstoned

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/251/

### DIFF
--- a/rest/api_test.go
+++ b/rest/api_test.go
@@ -3794,7 +3794,8 @@ func TestMOUDeletedOnTombstone(t *testing.T) {
 		rt.WaitForTombstone(docID, tombStoneVersion)
 
 		rawDoc := rt.GetRawDoc(docID)
-		assert.NotContains(t, rawDoc.Xattrs.RawDocXattrsOthers, base.MouXattrName)
+		mou, _ := rawDoc.Xattrs.RawDocXattrsOthers[base.MouXattrName]
+		assert.Nil(t, mou)
 
 	})
 }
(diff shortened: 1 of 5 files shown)


rest/api_test.go
```diff
@@ -3835,3 +3754,48 @@ func TestGetConfigAfterFailToStartOnlineProcess(t *testing.T) {
 	resp = rt.SendAdminRequest(http.MethodGet, "/_config?include_runtime=true", "")
 	RequireStatus(t, resp, http.StatusOK)
 }
+
+func TestMOUDeletedOnTombstone(t *testing.T) {
```
gregns1: Could we move this test into blip_api_crud_test.go?

gregns1 (changes requested):
Generally looks good just think we should move the test into different file. 

Wonder if we could also have a lower level test maybe in `collection_xattr_test` or something? A test that would use `WriteWithXattrs` to write a document xattr1 and xattr2, then calls `WriteTombstoneWithXattrs` to tombstone the doc an remove one of the xattrs? This will help us catch regressions even if `_mou` handling changes for some reason in future.

gregns1 (approved):

PR #8046, opened 2026-01-29
add test for existing DCP checkpoint prefixes

torcolvin:
CBG-5144 add test for existing DCP checkpoint prefixes

Add a test to make sure any future DCP refactoring doesn't break checkpoint names.

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/247/

### DIFF
--- a/db/background_mgr_attachment_migration_test.go
+++ b/db/background_mgr_attachment_migration_test.go
@@ -84,8 +84,6 @@ func TestAttachmentMigrationManagerResumeStoppedMigration(t *testing.T) {
 	defer db.Close(ctx)
 	collection, ctx := GetSingleDatabaseCollectionWithUser(ctx, t, db)
 
-	// create some docs with attachments defined, a large number is needed to allow us to stop the migration midway
-	// through without it completing first
 	for i := range 4000 {
 		docBody := Body{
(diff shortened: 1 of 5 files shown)


db/background_mgr_attachment_migration_test.go
```diff
@@ -255,3 +253,78 @@ func TestMigrationManagerDocWithSyncAndGlobalAttachmentMetadata(t *testing.T) {
...
+
+	migrationID := "1234"
+	testCases := []struct {
+		name          string
+		collectionIDs []uint32
+		groupID       string
+		expected      string
+	}{
+		{
+			name:          "default collection, no metadata id",
```
RIT3shSapata: shouldn't these be named as groupID instead of metadata id?

RIT3shSapata (commented):
a bit confused as to why the test case names have metadata id when group IDs are being used in the actual test case

db/background_mgr_attachment_migration_test.go
torcolvin: fixed!

RIT3shSapata (approved):

PR #8054, opened 2026-02-03
CBG-5143: Add EdDSA to set of supported algorithms for OIDC/JWTs

bbrks:
CBG-5143

Manually verified with Keycloak and OIDC Auth Code Flow.
Can demo setup in end of sprint if desired.

Screenshot 2026-02-03 at 19 02 38
Screenshot 2026-02-03 at 19 03 16
Screenshot 2026-02-03 at 19 03 09

- n/a

### DIFF
--- a/auth/jwt.go
+++ b/auth/jwt.go
@@ -39,6 +39,7 @@ var SupportedAlgorithms = map[jose.SignatureAlgorithm]bool{
 	jose.PS256: true,
 	jose.PS384: true,
 	jose.PS512: true,
+	jose.EdDSA: true,
 }
 
 // Full list of supported algorithms is used to initially parse the JWT.  The more restricted list (based
@@ -53,6 +54,7 @@ var SupportedAlgorithmsSlice = []jose.SignatureAlgorithm{
 	jose.PS256,
 	jose.PS384,
 	jose.PS512,
+	jose.EdDSA,
 }
 
 // JWTConfigCommon groups together configuration options common to both OIDC and local JWT authentication.


auth/jwt.go
```diff
@@ -39,6 +39,7 @@ var SupportedAlgorithms = map[jose.SignatureAlgorithm]bool{
 	jose.PS256: true,
 	jose.PS384: true,
 	jose.PS512: true,
+	jose.EdDSA: true,
 }
```
torcolvin: I'd be tempted to add this linter, since I caught this bug before it was reported using CBG-5129

```suggestion
} //exhaustive:enforce
```

auth/jwt.go
```diff
@@ -53,6 +54,7 @@ var SupportedAlgorithmsSlice = []jose.SignatureAlgorithm{
 	jose.PS256,
 	jose.PS384,
 	jose.PS512,
+	jose.EdDSA,
 }
```
torcolvin: I'd be tempted to add this linter, since I caught this bug before it was reported using CBG-5129

```suggestion
} //exhaustive:enforce
```

torcolvin (dismissed):

auth/jwt.go
bbrks: I made this change but it now fails on HMAC tokens and I am pretty sure that _is_ intentional.

I don't think we want to make this exhaustive even if it did highlight an upstream algorithm addition. I will revert this suggestion.

```
  Error: auth/jwt.go:32:27: missing keys in map of key type jose.SignatureAlgorithm: jose.HS256, jose.HS384, jose.HS512 (exhaustive)
  var SupportedAlgorithms = map[jose.SignatureAlgorithm]bool{
```

torcolvin: I was thinking we'd add false here for algorithms we know we don't want to support, but I don't care that much. I don't think this is of high value here, but this is an interesting linter to spot some types of issues in the future.

If you did add those algorithms as false, you would need to change https://github.com/couchbase/sync_gateway/blob/14bb857021b221d1d4f61ef0c884fb6fd5f17310/rest/config.go#L1017-L1019

torcolvin (approved):

PR #8056, opened 2026-02-04
CBG-5160 stub index creation for test purposes

torcolvin:
CBG-5160 stub index creation for test purposes

Creating and destroying indexes takes a lot of time for N1QL nodes. Sometimes they give up after 1 minute when running full integration tests.

I don't think to test the behavior of the `AsyncIndexInitManager` that we need to actually create the indexes, since we aren't verifying that they are being created. There is code elsewhere that does test whether these are created.

Note:

- All tests in rest/indextest drop all indexes as part of the bucket readier. They will get recreated when a database starts up. I would propose to move the tests that use `SetInitializeIndexesFunc` to rest package or something that isn't dropping indexes each time.
- Fixed a bug that I should pull into a separate ticket where `DatabaseContext.AsyncIndexInitManager` would be nil if the database was started offline and cause a panic.
-  There are still slow tests in `indextest` which are candidates for mocking, especially `TestChangeIndexPartitions`. The first half of this test would work for mocking, but the second part of the test what tests `_post_upgrade` removal process can't be tested using this framework, so I did not modify it.

I'm open to this not being the right thing to do, but I'm reluctant to increase the wait time for indexes to be created above 1 minute because I think it could take a really long time and our integration tests are now pushing two hours. I don't think the real creation of these indexes is that useful, we do test the partitioned index creation in db/indextest package. I think it would be fine to leave a small number of these tests preesent.

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/268/


rest/indextest/index_init_api_test.go
```diff
@@ -441,13 +425,21 @@ func runIndexInit(rt *rest.RestTester, body string) {
...
-		rest.AssertStatus(t, resp, http.StatusOK)
-		var body db.AsyncIndexInitManagerResponse
-		err := base.JSONUnmarshal(resp.BodyBytes(), &body)
-		require.NoError(c, err)
-		require.Equal(c, db.BackgroundProcessStateCompleted, body.State)
-	}, 1*time.Minute, 500*time.Millisecond)
+	db.RequireBackgroundManagerState(t, rt.GetDatabase().AsyncIndexInitManager, db.BackgroundProcessStateCompleted)
+}
+
+// getNoopBlockingInitializeIndexes is a replacement initialize index function that blocks forever. It is cancelled when DatabaseInitManager.Cancel is called. Used to avoid churn on n1ql nodes.
```
torcolvin:
```suggestion
// getNoopBlockingInitializeIndexes is a replacement initialize index function that blocks forever. It is cancelled when DatabaseInitManager.Cancel is called. Used to avoid churn on N1QL nodes.
```

db/database.go
```diff
@@ -598,6 +598,8 @@ func NewDatabaseContext(ctx context.Context, dbName string, bucket base.Bucket,
 	dbContext.ResyncManager = NewResyncManagerDCP(metadataStore, dbContext.UseXattrs(), metaKeys)
 	dbContext.AsyncIndexInitManager = NewAsyncIndexInitManager(dbContext.MetadataStore, dbContext.MetadataKeys)
```
bbrks, line 606: Existing line needs removal?


```suggestion
```

bbrks (changes requested):

bbrks (approved):

PR #8057, opened 2026-02-05
CBG-5153: Implement resync using CBGT for a single node

RIT3shSapata:
CBG-5153

Describe your PR here...

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/386/ (failed)
- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/398/
- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/399/
- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/400/


db/background_mgr_resync_dcp.go
```diff
@@ -181,6 +187,85 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+		if err != nil {
+			return fmt.Errorf("Error creating resync cfg: %v", err)
+		}
+
+		cbStore, ok := base.AsCouchbaseBucketStore(bucket)
+		if !ok {
+			// TODO: How to handle walrus for distributed resync?
+			base.InfofCtx(loggingCtx, base.KeyResync, "walrus bucket detected while running distributed resync")
+			return nil
+		}
```
torcolvin: If this is Walrus or CE, distributed resync should not be available, you can have this fail much earlier

db/background_mgr_resync_dcp.go
```diff
@@ -181,6 +187,85 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+		// TODO: make num partitions as a user configurable value for resync
+		numpartitions := uint16(1)
```
torcolvin: I think we can just use ImportPartitions here from db

db/background_mgr_resync_dcp.go
torcolvin: this code would belong in an else statement

torcolvin: This code will go into an else statement for non distributed resync

db/background_mgr_resync_dcp_test.go
```diff
@@ -240,47 +240,53 @@ func TestResyncManagerDCPStart(t *testing.T) {
...
+				"regenerateSequences": false,
+				"collections":         ResyncCollections{},
+			}
 
-		err := db.ResyncManager.Start(ctx, options)
-		require.NoError(t, err)
+			// REsyncManager.Dist = true
+			if distributed {
+				// TODO: not sure how to set Distributed as true
+			}
```
```suggestion
        // temporarily force setting resync here
		db.ResyncManager.Distributed = distributed
```

db/background_mgr_resync_dcp_test.go
```diff
@@ -240,47 +240,53 @@ func TestResyncManagerDCPStart(t *testing.T) {
...
+			err := db.ResyncManager.Start(ctx, options)
+			require.NoError(t, err)
 
-		RequireBackgroundManagerState(t, db.ResyncManager, BackgroundProcessStateCompleted)
+			RequireBackgroundManagerState(t, db.ResyncManager, BackgroundProcessStateCompleted)
```
torcolvin: For distributed resync until stop is implemented, this will have to call

waitForResyncDocsProcessed(int64(docsToCreate)) and then manually stop resync

base/dcp_sharded.go
```diff
@@ -120,15 +133,15 @@ func GenerateLegacyIndexName(dbName string) string {
...
+func createCBGTIndex(ctx context.Context, c *CbgtContext, dbName string, configGroupID string, bucket Bucket, scope string, collections []string, numPartitions uint16, resyncIndex bool) error {
 	sourceType := SOURCE_DCP_SG
 
 	sourceParams, err := cbgtFeedParams(ctx, scope, collections, dbName)
 	if err != nil {
 		return err
 	}
 
-	indexParams, err := cbgtIndexParams(ImportDestKey(dbName, scope, collections))
+	indexParams, err := cbgtIndexParams(DestKey(dbName, scope, collections, ImportDestType))
```
torcolvin: I think this code has to be incorrect here since you use resyncIndex above and this is hard coded for import

torcolvin (commented):
A few comments before I look more, I think the test is passing right now because it starts the non distributed resync.

db/background_mgr_resync_dcp_test.go
RIT3shSapata: So the thing is `db.ResyncManager` is of type `*BackgroundManager` and the `Distributed` is defined inside `ResyncManagerDCP`, so there is no way to directly access the `Distributed` property

db/background_mgr_resync_dcp_test.go
RIT3shSapata: Yes I agree, but right now, I'm not sure how to test this out, as mentioned in my previous comment

db/background_mgr_resync_dcp_test.go
torcolvin: is this just db.ResyncManager.Process.Distributed = true?

base/dcp_common.go
```diff
@@ -35,6 +35,7 @@ const kCheckpointTimeThreshold = 1 * time.Minute
 // DCP Feed IDs are used to build unique DCP identifiers
 const DCPCachingFeedID = "SG"
 const DCPImportFeedID = "SGI"
+const DCPResyncFeedID = "SGR"
```
torcolvin: I'm going to change this in https://jira.issues.couchbase.com/browse/CBG-4286 but I think that this value is only used for logging when calling `NewDCPDest`. If that's true, then this value doesn't have to be `dcp_common.go` and should say "resync" with the ID of the resync process `resyncLoggingID`.

base/dcp_sharded.go
torcolvin: This code will force the use of `import` in the index names, this needs to be handled.

base/dcp_sharded.go
```diff
@@ -42,6 +43,18 @@ var firstVersionToSupportCollections = &ComparableBuildVersion{
 	patch: 0,
 }
 
+type DestType int
+
+const (
+	ImportDestType DestType = iota // "import"
+	ResyncDestType                 // "resync"
+)
```
torcolvin: you could just make this a typed string so that you don't need to store  destTypeNames as well.

```suggestion
type DestType string

const (
	ImportDestType DestType = "import"
	ResyncDestType = "resync"
)
```

base/dcp_sharded.go
```diff
@@ -63,7 +76,7 @@ type CbgtContext struct {
 
 // StartShardedDCPFeed initializes and starts a CBGT Manager targeting the provided bucket.
 // dbName is used to define a unique path name for local file storage of pindex files
-func StartShardedDCPFeed(ctx context.Context, dbName string, configGroup string, uuid string, heartbeater Heartbeater, bucket Bucket, spec BucketSpec, scope string, collections []string, numPartitions uint16, cfg cbgt.Cfg) (*CbgtContext, error) {
+func StartShardedDCPFeed(ctx context.Context, dbName string, configGroup string, uuid string, heartbeater Heartbeater, bucket Bucket, spec BucketSpec, scope string, collections []string, numPartitions uint16, cfg cbgt.Cfg, resyncIndex bool) (*CbgtContext, error) {
```
torcolvin: This could take `DestType` instead of a boolean parameter, which would be generally preferred since it would be more reasonable. The name `DestType` might not be appropriate, but it could be ShardedDCPFeedType instead?

It might be worth figuring out below what is actually needed to do to split on the feed type and whether to pass these arguments into this function or pass a type parameter. Review this comment after looking through the rest of the code and the comments.

base/dcp_sharded.go
```diff
@@ -85,7 +98,7 @@ func StartShardedDCPFeed(ctx context.Context, dbName string, configGroup string,
 	ctx = CorrelationIDLogCtx(ctx, DCPImportFeedID)
```
torcolvin: This is wrong here because it is specific to import, but I think that the place to fix this is probably in the caller of `StartShardedDCPFeed` rather than putting an if statement here.

base/dcp_sharded.go
torcolvin: This needs to be unique to a db and a particular resync ID run. That is, if resync?action=reset happens, it should probably not reuse the same run.

This says this is length constrained, but I'm not sure what the length constraint is. It might be 200 as per https://jira.issues.couchbase.com/browse/CBG-4286 but you have to look up how it is used by cbgt.

If it is length restricted, return an error if this is too long.

torcolvin: This needs to be handled.

base/heartbeat.go
```diff
@@ -45,7 +45,7 @@ type HeartbeatListener interface {
 	Stop()
 }
 
-// couchbaseHeartBeater is a Heartbeater implementation that uses Couchbase document expiry for heartbeat detection.
+// CouchbaseHeartBeater is a Heartbeater implementation that uses Couchbase document expiry for heartbeat detection.
```
torcolvin: Does this need to be changed to be public?

base/log_keys.go
```diff
@@ -51,6 +51,7 @@ const (
 	KeyMigrate
 	KeyQuery
 	KeyReplicate
+	KeyResync
```
torcolvin: Is there a requirement to have new log keys? I would not create new log keys without an explicit requirement. The log key definitions are tied into capella here and creating a new log key would mean that it would have to be coreleased on app services https://docs.couchbase.com/app-services/monitoring/configure-log-streaming-app-endpoint.html - let's keep the log keys the same unless we want to change for all resync logs, in which case this should be a separate ticket.

Adding new log keys also would necessitate a change to API docs.

db/background_mgr_resync_dcp.go
```diff
@@ -110,8 +113,14 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
 	regenerateSequences := options["regenerateSequences"].(bool)
 	resyncCollections := options["collections"].(ResyncCollections)
 
+	resyncCbgtContext := &base.CbgtContext{}
+	resyncHB := &base.CouchbaseHeartBeater{}
+
 	resyncLoggingID := "Resync: " + r.ResyncID
 
+	var doneChan chan error
+	dcpClient := &base.GoCBDCPClient{}
```
torcolvin: Cbgtcontext, CouchbaseHeartBeater, GoCBDCPClient can't be used outside of their New functions.

Prefer to define these inside `Distributed` block if possible, or use nil pointers if it is not.

db/background_mgr_resync_dcp.go
```diff
@@ -181,22 +190,104 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+		cbStore, ok := base.AsCouchbaseBucketStore(bucket)
+		if !ok {
+			base.InfofCtx(loggingCtx, base.KeyResync, "walrus bucket detected while running distributed resync")
+			return nil
+		}
+
+		if !base.IsEnterpriseEdition() {
+			base.WarnfCtx(loggingCtx, "CE SGW detected while running distributed resync")
+			return nil
+		}
```
torcolvin: I'll probably push this code up in a separate PR but I was thinking of a call like:

```
func (db *DatabaseContext) useShardedDCP() bool {
	return !base.IsEnterpriseEdition()  &&!db.BucketSpec.IsWalrusBucket()
}
```

and this could be used by both import feed and resync. If this is an unsupported mode, it would be appropriate to just return an error.

db/background_mgr_resync_dcp.go
```diff
@@ -181,22 +190,104 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+		resyncHBPrefix := db.MetadataKeys.ResyncHeartbeaterPrefix(db.Options.GroupID)
+		resyncHB, err = base.NewCouchbaseHeartbeater(db.MetadataStore, resyncHBPrefix, db.UUID)
+		if err != nil {
+			return fmt.Errorf("Error creating resync heartbeater: %v", err)
+		}
+		err = resyncHB.StartSendingHeartbeats(ctx)
+		if err != nil {
+			return fmt.Errorf("Error starting resync heartbeater: %v", err)
+		}
+
```
torcolvin, line 247: If there's _any_ error below this, we need to stop sending heartbeats, this probably gets done with a defer.

db/import_pindex.go
```diff
@@ -20,32 +20,35 @@ import (
...
-		&cbgt.PIndexImplType{
-			New:       getNewPIndexImplType(ctx),
-			Open:      OpenImportPIndexImpl,
-			OpenUsing: getOpenImportPIndexImplUsing(ctx),
-			Description: "general/syncGateway-import " +
-				" - import processing for shared bucket access",
-		})
+	for _, pIndexName := range []string{base.CBGTIndexTypeSyncGatewayImport, base.CBGTIndexTypeSyncGatewayResync} {
+		pIndexType := pIndexName + configGroup
+		// pIndexType := [ base.CBGTIndexTypeSyncGatewayImport + configGroup, resyncPindexName]
```
torcolvin:
```suggestion
```

db/import_pindex.go
```diff
@@ -20,32 +20,35 @@ import (
...
 
-// registerImportPindexImplMutex locks access to cbgt.RegisterImportPindexImpl.
+// registerImportPindexImplMutex locks access to cbgt.RegisterPindexImpl.
 var registerImportPindexImplMutex = sync.Mutex{}
 
-// RegisterImportPindexImpl registers the PIndex type definition.  This is invoked by cbgt when a Pindex (collection of
+// RegisterPindexImpl registers the PIndex type definition.  This is invoked by cbgt when a Pindex (collection of
 // vbuckets) is assigned to this node.
-func RegisterImportPindexImpl(ctx context.Context, configGroup string) {
+func RegisterPindexImpl(ctx context.Context, configGroup string) {
```
torcolvin: Suggest rename to:

RegisterPindexImpls

db/import_pindex.go
```diff
@@ -20,32 +20,35 @@ import (
...
+		pIndexType := pIndexName + configGroup
+		// pIndexType := [ base.CBGTIndexTypeSyncGatewayImport + configGroup, resyncPindexName]
+		base.InfofCtx(ctx, base.KeyDCP, "Registering PindexImplType for %s", pIndexType)
+		cbgt.RegisterPIndexImplType(pIndexType,
+			&cbgt.PIndexImplType{
+				New:       getNewPIndexImplType(ctx),
+				Open:      OpenPIndexImpl,
+				OpenUsing: getOpenPIndexImplUsing(ctx),
+				Description: "general/syncGateway-import " +
+					" - import processing for shared bucket access",
```
torcolvin: I don't think this value is ever visible, but this string says import

db/import_pindex.go
```diff
@@ -20,32 +20,35 @@ import (
...
+			})
+	}
 }
 
 // getListenerForIndex looks up the importListener for the dbName specified in the index params
```
torcolvin:
```suggestion
// getListenerForIndex looks up the Dest instance for a given set of cbgt index parameters.
```

db/import_pindex.go
torcolvin:
```suggestion
	base.DebugfCtx(ctx, base.KeyDCP, "Fetching dest for %v", base.MD(sgIndexParams.DestKey))
```

torcolvin:
```suggestion
		return nil, fmt.Errorf("error retrieving dest for indexParams %v: %v", indexParams, fetchErr)
```

db/import_pindex.go
```diff
@@ -71,29 +74,29 @@ func getListenerImportDest(ctx context.Context, indexParams string, restart func
...
-	newImportPIndexImpl := func(indexType, indexParams, path string, restart func()) (cbgt.PIndexImpl, cbgt.Dest, error) {
+	// OpenImportindexImplUsing is called.
+	newPIndexImpl := func(indexType, indexParams, path string, restart func()) (cbgt.PIndexImpl, cbgt.Dest, error) {
 		defer base.FatalPanicHandler()
 
-		importDest, err := getListenerImportDest(ctx, indexParams, restart)
+		dest, err := getListenerDest(ctx, indexParams, restart)
 		if err != nil {
 			// This error can occur when a stale index definition hasn't yet been removed from the plan (e.g. on update to db config)
 			base.InfofCtx(ctx, base.KeyDCP, "No importDest found for indexParams - usually an obsolete index pending removal. %v", err)
```
torcolvin:
```suggestion
			base.InfofCtx(ctx, base.KeyDCP, "No dest found for indexParams - usually an obsolete index pending removal. %v", err)
```

db/background_mgr_resync_dcp.go
```diff
@@ -269,15 +360,20 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
 		}
 	case <-terminator.Done():
 		base.DebugfCtx(ctx, base.KeyAll, "[%s] Terminator closed. Ending Resync process.", resyncLoggingID)
-		err = dcpClient.Close()
-		if err != nil {
-			base.WarnfCtx(ctx, "[%s] Failed to close resync DCP client! %v", resyncLoggingID, err)
-			return err
-		}
+		if r.Distributed {
+			resyncCbgtContext.Stop()
```
torcolvin: I think that the context needs to be stopped if there are any errors earlier, see also for the heartbeat prefix.

db/import_pindex.go
```diff
@@ -20,32 +20,35 @@ import (
 	"github.com/couchbase/sync_gateway/base"
 )
 
-// registerImportPindexImplMutex locks access to cbgt.RegisterImportPindexImpl.
+// registerImportPindexImplMutex locks access to cbgt.RegisterPindexImpl.
 var registerImportPindexImplMutex = sync.Mutex{}
```
torcolvin: This code no longer just applies to import, so consider renaming this variable.

torcolvin (commented):
I gave a non exhaustive review for any cases where import was used for code, but I don't think I found all cases. Definitely give a look through the code pathways to see if you can find other locations.

base/heartbeat.go
RIT3shSapata: I had to change this to public because starting the resync manager's is present in the `db` package, while this is in the `base` package

base/dcp_sharded.go
RIT3shSapata: It is length restricted but the names are generated by using `Crc32Hash`, which returns a `uint32`. So when its formatted using `%x`, it would be hex without leading zeros, which I think would be upto 8 hex characters. So the maximum length would be for resync index of 34. So do you still think it makes sense to return an error if the length is beyond 200

db/background_mgr_resync_dcp.go
RIT3shSapata: I've defined them outside the `Distributed` block, because these are later required to be closed on receiving a terminate request

.github/workflows/ci.yml
```diff
@@ -32,7 +32,7 @@ concurrency:
   cancel-in-progress: ${{ !contains(github.ref, 'release/') && !contains(github.ref, 'main') }}
 
 env:
-  GO_VERSION: &go_version 1.25.6
+  GO_VERSION: &go_version 1.25.7
```
torcolvin:
```suggestion
  GO_VERSION: &go_version 1.25.6
```

let's leave this out of this PR

base/dcp_sharded.go
```diff
@@ -104,31 +112,39 @@ func StartShardedDCPFeed(ctx context.Context, dbName string, configGroup string,
...
-func GenerateLegacyIndexName(dbName string) string {
-	return dbName + "_import"
+// Given a dbName and feedType, generates a name based on the approach used prior to CBG-626.  Used for upgrade handling
+func GenerateLegacyIndexName(dbName string, feedType ShardedDCPFeedType) string {
+	if feedType == ImportShardedDCPFeedType {
+		return dbName + "_import"
+	} else {
+		return dbName + "_resync"
+	}
 }
```
torcolvin: I don't think this code is ever going to be called because it relies on finding the feed from before. I think you can leave this function as as, document / rename it so you know it is only going to get called from import.

base/dcp_sharded.go
```diff
@@ -104,31 +112,39 @@ func StartShardedDCPFeed(ctx context.Context, dbName string, configGroup string,
...
 
 // Given a dbName, generate a unique and length-constrained index name for CBGT to use as part of their DCP name.
-func GenerateIndexName(dbName string) string {
+func GenerateIndexName(dbName string, feedID string) string {
```
torcolvin: We can pull this into a separate ticket but I think that this value needs to be <200 characters and we need to shorten this.

We can't break code that might work for import but as we make the index name longer, we want to make sure that this code works if there is a long database name. https://github.com/couchbase/sync_gateway/blob/aa83a21e9c7a77f31018a37467aab4f99de6e4ee/rest/config_manager.go#L832 is an example of how we handle this for a metadataID

I think creating a separate ticket is a good plan for this.

base/dcp_sharded.go
```diff
@@ -176,14 +197,14 @@ func createCBGTIndex(ctx context.Context, c *CbgtContext, dbName string, configG
 }
 
 // dcpSafeIndexName returns an index name and previousIndexUUID to handle upgrade scenarios from the
-// legacy index name format ("dbname_import") to the new length-safe format ("db[crc32]_index").
+// legacy index name format ("dbname_import") or ("dbname_resync") to the new length-safe format ("db[crc32]_index").
 // Handles removal of legacy index definitions, except for the case where the legacy index is
 // the only index defined, and the name is safe.  In that case, continue using legacy index name
 // to avoid restarting the import processing from zero
-func dcpSafeIndexName(ctx context.Context, c *CbgtContext, dbName string) (safeIndexName, previousUUID string) {
+func dcpSafeIndexName(ctx context.Context, c *CbgtContext, dbName string, feedType ShardedDCPFeedType, feedID string) (safeIndexName, previousUUID string) {
```
torcolvin: I think we can pass feedType to both functions here, and through the callstack.

db/background_mgr_resync_dcp_test.go
```diff
@@ -240,47 +240,63 @@ func TestResyncManagerDCPStart(t *testing.T) {
...
+				rs, ok := db.ResyncManager.Process.(*ResyncManagerDCP)
+				require.True(t, ok)
+				rs.Distributed = true
+			}
+			err := db.ResyncManager.Start(ctx, options)
+			require.NoError(t, err)
 
-		RequireBackgroundManagerState(t, db.ResyncManager, BackgroundProcessStateCompleted)
+			if distributed {
+				waitforResycnDocsChanged(t, db, int64(docsToCreate))
```
```suggestion
				waitforResyncDocsChanged(t, db, int64(docsToCreate))
```

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/couchbase/sync_gateway
 
-go 1.25.6
+go 1.25.7
```
torcolvin:
```suggestion
go 1.25.6
```

drop this change from this PR

Jenkinsfile
```diff
@@ -15,7 +15,7 @@ pipeline {
     }
 
     tools {
-        go '1.25.6'
+        go '1.25.7'
```
torcolvin:
```suggestion
        go '1.25.6'
```

drop this change from this PR

db/background_mgr_resync_dcp.go
```diff
@@ -269,15 +359,20 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+			resyncHB.Stop(ctx)
```
torcolvin: I think you can move this up into the defer block, because you want to stop the heartbeater regardless in this function. That will mean that you can make `resyncCbgtContext` and `resyncHB` local variables.

db/background_mgr_resync_dcp.go
```diff
@@ -181,22 +190,103 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+		}
+
+		numpartitions := db.Options.ImportOptions.ImportPartitions
+		resyncCbgtContext, err = base.StartShardedDCPFeed(loggingCtx, db.Name, db.Options.GroupID, db.UUID, resyncHB, bucket,
+			cbStore.GetSpec(), scopeName, collectionNamesByScope[scopeName], numpartitions, resyncCfg, base.ResyncShardedDCPFeedType, dcpFeedKey)
+
+		if err != nil {
+			return fmt.Errorf("Error starting sharded dcp feed: %v", err)
+		}
+		defer resyncCbgtContext.Stop()
```
torcolvin:
```suggestion
		defer func() {
			resyncCbgtContext.Stop()
			resyncHB.Stop()
		}
```

db/background_mgr_resync_dcp.go
```diff
@@ -181,22 +190,103 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+
+		cbStore, _ := base.AsCouchbaseBucketStore(bucket)
+		// Dest creation
+		for sn := range db.Scopes {
+			scopeName = sn
+		}
+
+		for _, collection := range db.CollectionByID {
+			if bucket.IsSupported(sgbucket.BucketStoreFeatureCollections) && !db.OnlyDefaultCollection() {
+				collectionNamesByScope[collection.ScopeName] = append(collectionNamesByScope[collection.ScopeName], collection.Name)
```
torcolvin: We should make a common function for getting collection names. I think I have some code to do this, and I'm OK leaving it in this intermediate state to clean up

db/background_mgr_resync_dcp.go
```diff
@@ -181,22 +190,103 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+		if db.OnlyDefaultCollection() {
+			resyncDestKey = base.DestKey(db.Name, "", []string{}, base.ResyncShardedDCPFeedType)
+		} else {
+			resyncDestKey = base.DestKey(db.Name, scopeName, collectionNamesByScope[scopeName], base.ResyncShardedDCPFeedType)
+		}
+
+		maxVbNo, err := bucket.GetMaxVbno()
+
+		// TODO: Should I use separate checkpoints or can the same import checkpoints be used?
+		checkPointPrefix := db.MetadataKeys.DCPVersionedCheckpointPrefix(db.Options.GroupID, 0)
```
torcolvin: These can not be the same checkpoint names since the format of checkpoints is:

| type | name | struct |
|-------------- |---------------------------- | ---------------------------------------------------------------------------- |
| cbgt                | `{checkpointPrefix}{vbNo}` | https://github.com/couchbase/cbgt/blob/f9f7a06e180707e25299e45b198c73d953817a2b/feed_dcp_gocbcore.go#L1140 |
| sync gateway | `{checkpointPrefix}{workerNo}` | map[uint16]DCPMetadata |

The worker numbers are from 0-7 and so will overlap with the vbucket ids. Additionally, the names are tied to a particular implementation of resync - CBG-5144.

We probably want to file a ticket to handle distributed resync checkpoint cleanup, and link it into CBG-5041.

db/background_mgr_resync_dcp.go
```diff
@@ -181,22 +190,103 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+			}
+		}
+		sort.Strings(collectionNamesByScope[scopeName])
+		if db.OnlyDefaultCollection() {
+			resyncDestKey = base.DestKey(db.Name, "", []string{}, base.ResyncShardedDCPFeedType)
+		} else {
+			resyncDestKey = base.DestKey(db.Name, scopeName, collectionNamesByScope[scopeName], base.ResyncShardedDCPFeedType)
+		}
+
+		maxVbNo, err := bucket.GetMaxVbno()
```
torcolvin: This value is `DatabaseContext.numVBuckets`, so we can avoid error checking here.

db/background_mgr_resync_dcp.go
```diff
@@ -181,22 +190,103 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+			resyncDest, _, err := base.NewDCPDest(loggingCtx, callback, db.Bucket, maxVbNo, true, nil, dcpFeedKey, nil, checkPointPrefix, db.MetadataKeys)
+			if err != nil {
+				return nil, fmt.Errorf("Error creating resync dest: %v", err)
+			}
+			return resyncDest, nil
+		}
+
+		base.StoreDestFactory(loggingCtx, resyncDestKey, resyncDestFunc)
+
+		base.InfofCtx(loggingCtx, base.KeyJavascript, "ResyncID: %s Starting DCP resync for bucket: %q ", resyncLoggingID, base.UD(bucket.GetName()))
```
torcolvin:
```suggestion
		base.InfofCtx(loggingCtx, base.KeyAll, "Resync: Starting DCP resync for bucket: %q ",  base.UD(bucket.GetName()))
```

I believe that the loggingCtx should already contain the CorrelationID and so it doesn't need to be duplicated. You should make sure that is true.

db/background_mgr_resync_dcp.go
```diff
@@ -181,22 +190,103 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+
+		// CFG creation:
+		resyncCfg, err := base.NewCfgSG(ctx, db.MetadataStore, db.MetadataKeys.ResyncCfgPrefix(db.Options.GroupID))
+		if err != nil {
+			return fmt.Errorf("Error creating resync cfg: %v", err)
+		}
+
+		numpartitions := db.Options.ImportOptions.ImportPartitions
+		resyncCbgtContext, err = base.StartShardedDCPFeed(loggingCtx, db.Name, db.Options.GroupID, db.UUID, resyncHB, bucket,
+			cbStore.GetSpec(), scopeName, collectionNamesByScope[scopeName], numpartitions, resyncCfg, base.ResyncShardedDCPFeedType, dcpFeedKey)
```
torcolvin: cbStore.GetSpec can be replaced with db.BucketSpec to simplify this code

db/background_mgr_resync_dcp.go
```diff
@@ -181,22 +190,103 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+
+		if err != nil {
+			return fmt.Errorf("Error starting sharded dcp feed: %v", err)
```
torcolvin:
```suggestion
			return fmt.Errorf("Error starting resync sharded dcp feed: %v", err)
```

db/background_mgr_resync_dcp.go
```diff
@@ -181,22 +190,103 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+		var scopeName string
+		collectionNamesByScope := make(map[string][]string)
+
+		// TODO: Handle multiple scopes
+
+		loggingCtx := db.AddBucketUserLogContext(ctx)
+
+		if !db.useShardedDCP(){
+			base.WarnfCtx(loggingCtx, "running distributed resync is not supported")
+			return nil
```
torcolvin: we can just return an error here without logging

db/background_mgr_resync_dcp.go
RIT3shSapata: `loggingCtx` does have `correlationID` field but it is empty in this scenario. On checking the code, `UserLogCtx` does not populate the `correlationID`. So I think this is necessary here?

db/background_mgr_resync_dcp.go
RIT3shSapata: Filed - [CBG-5175](https://jira.issues.couchbase.com/browse/CBG-5175)

base/dcp_sharded.go
RIT3shSapata: I did not understand this, could you please elaborate more?

base/dcp_sharded.go
RIT3shSapata: Like I mentioned in my comment earlier, I don't think it will ever exceed 34:
https://github.com/couchbase/sync_gateway/pull/8057#discussion_r2815486962
we can pair up and discuss it further

base/dcp_sharded.go
torcolvin: I'm inclined to just change the code so that it returns an error to bubble up if we do make a mistake in the future.

base/heartbeat.go
torcolvin: I think it's weird to have it non public when it is used but ultimately this is still used in db package already. https://github.com/couchbase/sync_gateway/blob/189cc8a18997f268674419c6f132efe5015be382/db/database.go#L2292-L2294

We are probably writing a slightly different heartbeater, so this isn't worth spending too much time on.

base/dcp_sharded.go
torcolvin: Can you pass just `feedType ShardedDCPFeedType`? I think that the feedID should actually always be the same for resync, and doesn't need to be unique per run. 

That was my mistake in the earlier specification.

db/background_mgr_resync_dcp.go
torcolvin: Agree, this is out of scope for fix in this PR, let me see about fixing this separately. We should still change the log key to be KeyAll and `Resync:` is redundant.

base/constants_syncdocs.go
```diff
@@ -224,6 +233,28 @@ func (m *MetadataKeys) SGCfgPrefix(groupID string) string {
...
+	if groupID != "" {
+		return m.resyncHeartbeaterPrefix + groupID + ":"
+	}
+	return m.resyncHeartbeaterPrefix
+}
+
+// ResyncCfgPrefix returns a document prefix to use for resync cfg documents
+//
+//	format: _sync:{m_$}:cfg[groupID:]   (collections)
+//	format: _sync:cfg:[groupID:]   (default)
```
torcolvin:
```suggestion
//	format: _sync:{m_$}:resync_cfg[groupID:]   (collections)
//	format: _sync:resync_cfg:[groupID:]   (default)
```

base/dcp_sharded.go
```diff
@@ -151,12 +163,17 @@ func createCBGTIndex(ctx context.Context, c *CbgtContext, dbName string, configG
 	}
 
 	// Determine index name and UUID
-	indexName, previousIndexUUID := dcpSafeIndexName(ctx, c, dbName)
+	indexName, previousIndexUUID := dcpSafeIndexName(ctx, c, dbName, shardedDcpFeedType, feedID)
```
adamcfraser: I had thought that dcpSafeIndexName was specifically around supporting upgrade from instances that had been using the non-length-safe import index name.  We shouldn't need that for new feed types?

base/dcp_sharded.go
torcolvin: Can you file a separate ticket for this and resolve this when you are done?

base/dcp_sharded.go
```diff
@@ -459,8 +476,10 @@ func (c *CbgtContext) RemoveFeedCredentials(dbName string) {
 	// CBG-4394: removing root certs for the bucket should be done, but it is keyed based on the bucket UUID, and multiple dbs can use the same bucket
 }
 
-// ImportDestKey is used for retrieval of import dest from cbgtDestFactories
-func ImportDestKey(dbName string, scope string, collections []string) string {
+// Format of dest key for retrieval of import dest from cbgtDestFactories
+func DestKey(dbName string, scope string, collections []string, shardedDcpFeedType ShardedDCPFeedType) string {
+	// ImportDestKey is used for retrieval of import dest from cbgtDestFactories
+	//func ImportDestKey(dbName string, scope string, collections []string) string {
```
torcolvin:
```suggestion
```

base/dcp_sharded_test.go
torcolvin: change name of this test to `TestDestKey` and all the cases for resync as well

db/background_mgr_resync_dcp.go
```diff
@@ -194,22 +200,101 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
 
-	base.InfofCtx(ctx, base.KeyAll, "[%s] Starting DCP feed for resync", resyncLoggingID)
-	doneChan, err := dcpClient.Start()
-	if err != nil {
-		base.WarnfCtx(ctx, "[%s] Failed to start resync DCP feed! %v", resyncLoggingID, err)
-		_ = dcpClient.Close()
-		return err
-	}
-	base.DebugfCtx(ctx, base.KeyAll, "[%s] DCP client started.", resyncLoggingID)
+		// TODO: Handle multiple scopes
```
torcolvin:
```suggestion
```

db/background_mgr_resync_dcp.go
```diff
@@ -194,22 +200,101 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+
+		loggingCtx := db.AddBucketUserLogContext(ctx)
```
torcolvin, line 205: I would be inclined to shadow ctx here because it would be easy to not use the right context.

```suggestion
		ctx = db.AddBucketUserLogContext(ctx)
```

That said, I don't think this seems right. I think we should defer this into a future ticket that you should file and note here.

title: ensure resync context logging is consistent

Verify use of context.Context used in `ResyncManager` to include the correlation ID, bucket, and collection when appropriate. All operations should be appropriately tagged

db/background_mgr_resync_dcp.go
```diff
@@ -194,22 +200,101 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+		} else {
+			resyncDestKey = base.DestKey(db.Name, scopeName, collectionNamesByScope[scopeName], base.ResyncShardedDCPFeedType)
+		}
+
+
+		// TODO: Use different checkpoint names, to be fixed part of CBG-5144
+		checkPointPrefix := db.MetadataKeys.DCPVersionedCheckpointPrefix(db.Options.GroupID, 0)
+		if err != nil {
+			return fmt.Errorf("Error getting max VB number: %v", err)
+		}
```
torcolvin:
```suggestion
```

This code is dead now

db/background_mgr_resync_dcp.go
```diff
@@ -194,22 +200,101 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+		if err != nil {
+			return fmt.Errorf("getting collection names by scope failed: %v", err)
+		}
+
+		sort.Strings(collectionNamesByScope[scopeName])
+		if db.OnlyDefaultCollection() {
+			resyncDestKey = base.DestKey(db.Name, "", []string{}, base.ResyncShardedDCPFeedType)
+		} else {
+			resyncDestKey = base.DestKey(db.Name, scopeName, collectionNamesByScope[scopeName], base.ResyncShardedDCPFeedType)
+		}
```
torcolvin: I think the DestKey will do the right thing even if you specify only `_default._default`. That said, I don't actually think we need to special case this even in the dest key, and I'd not special case this. The reason it had to be special cased for import is that prior to 3.1 and collection support was added to Sync Gateway, the checkpoint names weren't included. Because we are implementing these from scratch, we can just unilaterally include checkpoint names.

```
resyncDestKey = base.DestKey(db.Name, scopeName, collectionNamesByScope[scopeName], base.ResyncShardedDCPFeedType)
```

db/background_mgr_resync_dcp.go
```diff
@@ -194,22 +200,101 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+
+		numpartitions := db.Options.ImportOptions.ImportPartitions
+		resyncCbgtContext, err := base.StartShardedDCPFeed(loggingCtx, db.Name, db.Options.GroupID, db.UUID, resyncHB, bucket,
+			 scopeName, collectionNamesByScope[scopeName], numpartitions, resyncCfg, base.ResyncShardedDCPFeedType, clientOptions.CheckpointPrefix)
 
-	r.VBUUIDs = base.GetVBUUIDs(dcpClient.GetMetadata())
+		if err != nil {
+			return fmt.Errorf("Error starting resync sharded dcp feed: %v", err)
+		}
+		defer resyncCbgtContext.Stop()
```
torcolvin: This is redundant with 277


```suggestion
```

db/database.go
```diff
@@ -2616,3 +2616,17 @@ func (db *DatabaseContext) usingRosmar() bool {
...
+	if err != nil {
+		return collectionNamesByScope, err
+	}
+	for _, collection := range db.CollectionByID {
+		if bucket.IsSupported(sgbucket.BucketStoreFeatureCollections) && !db.OnlyDefaultCollection() {
+			collectionNamesByScope[collection.Name] = append(collectionNamesByScope[collection.Name], collection.Name)
+		}
+	}
+	return collectionNamesByScope, nil
+}
```
torcolvin: on main , this is `db.collectionNames` now.

```suggestion
```

db/background_mgr_resync_dcp.go
RIT3shSapata: Filed: https://jira.issues.couchbase.com/browse/CBG-5243

db/background_mgr_resync_dcp.go
RIT3shSapata: filed: https://jira.issues.couchbase.com/browse/CBG-5243

base/dcp_sharded.go
RIT3shSapata: Filed: https://jira.issues.couchbase.com/browse/CBG-5244

db/background_mgr_resync_dcp_test.go
```diff
@@ -240,47 +240,63 @@ func TestResyncManagerDCPStart(t *testing.T) {
...
-		collectionName := scopeAndCollectionName.CollectionName()
+			dbc, _ := GetSingleDatabaseCollectionWithUser(ctx, t, db)
+			scopeAndCollectionName := dbc.ScopeAndCollectionName()
+			scopeName := scopeAndCollectionName.ScopeName()
+			collectionName := scopeAndCollectionName.CollectionName()
 
-		initialStats := getResyncStats(t, db)
-		log.Printf("initialStats: processed[%v] changed[%v]", initialStats.DocsProcessed, initialStats.DocsChanged)
+			initialStats := getResyncStats(t, db)
+			log.Printf("initialStats: processed[%v] changed[%v]", initialStats.DocsProcessed, initialStats.DocsChanged)
```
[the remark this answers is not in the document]
RIT3shSapata: this is useful information for running the unit tests

db/background_mgr_resync_dcp_test.go
```diff
@@ -625,7 +641,15 @@ func waitForResyncDocsProcessed(t testing.TB, db *Database, count int64) {
...
-	}, 10*time.Second, 1*time.Millisecond)
+	}, 1*time.Minute, 1*time.Millisecond)
+}
+
+func waitforResyncDocsChanged(t testing.TB, db *Database, count int64) {
+
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		stats := getResyncStats(t, db)
+		assert.Equal(c, stats.DocsChanged, count)
+	}, 5*time.Minute, 1*time.Millisecond)
```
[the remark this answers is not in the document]
RIT3shSapata: changing this to 10ms

base/dcp_test.go
```diff
@@ -229,7 +245,8 @@ func TestCBGTIndexCreation(t *testing.T) {
 			defer context.Manager.Stop()
 
 			// Define index type
-			configGroup := "configGroup" + t.Name()
+			//configGroup := "configGroup" + t.Name()
+			configGroup := ""
 			indexType := CBGTIndexTypeSyncGatewayImport + configGroup
 			cbgt.RegisterPIndexImplType(indexType,
 				&cbgt.PIndexImplType{})
```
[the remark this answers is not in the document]
RIT3shSapata: done

db/background_mgr_resync_dcp.go
```diff
@@ -194,22 +200,90 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+		loggingCtx := db.AddBucketUserLogContext(ctx)
+
+		if !db.useShardedDCP() {
+			return fmt.Errorf("running distributed resync is not supported")
+		}
+
+		// Dest creation
+		for sn := range db.Scopes {
+			scopeName = sn
+		}
```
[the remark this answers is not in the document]
RIT3shSapata: only one scope is allowed in SGW

base/constants_syncdocs.go
```diff
@@ -39,6 +39,8 @@ const (
 	MetaKeyRolePrefix                                          // "role:"
 	MetaKeyUserEmailPrefix                                     // "useremail:"
 	MetaKeySessionPrefix                                       // "session:"
+	MetaKeyResyncHeartBeaterPrefix                             // "resync_hb:"
+	MetaKeyResyncCfgPrefix                                     // "resync_cfg
```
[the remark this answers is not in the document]
torcolvin: This comment looks correct from copilot.

base/constants_syncdocs.go
```diff
@@ -224,6 +233,28 @@ func (m *MetadataKeys) SGCfgPrefix(groupID string) string {
...
+func (m *MetadataKeys) ResyncHeartbeaterPrefix(groupID string) string {
+	if groupID != "" {
+		return m.resyncHeartbeaterPrefix + groupID + ":"
+	}
+	return m.resyncHeartbeaterPrefix
+}
+
+// ResyncCfgPrefix returns a document prefix to use for resync cfg documents
+//
+//	format: _sync:{m_$}:resync_cfg[groupID:]   (collections)
```
torcolvin:
```suggestion
//	format: _sync:{m_$}:resync_cfg:[groupID:]   (collections)
```

I think this is correct?

base/constants_syncdocs_test.go
```diff
@@ -140,6 +140,8 @@ func TestMetadataKeys(t *testing.T) {
 		sessionKey                       string
 		backgroundProcessHeartbeatPrefix string // with backgroundID
 		backgroundProcessStatusPrefix    string // with backgroundID
+		resyncHeatbeatPrefix             string
+		resyncCfgPrefix                  string
 	}{
```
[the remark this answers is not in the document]
torcolvin: This comment is correct I think.

base/dcp_sharded.go
```diff
@@ -32,6 +32,7 @@ const (
 	CBGTCfgNodeDefsKnown              = SyncDocPrefix + "cfgnodeDefs-known"
 	CBGTCfgNodeDefsWanted             = SyncDocPrefix + "cfgnodeDefs-wanted"
 	CBGTCfgPlanPIndexes               = SyncDocPrefix + "cfgplanPIndexes"
+	CBGTIndexTypeSyncGatewayResync    = "syncGateway-resync-"
```
torcolvin: I think we decided _not_ to scope resync by groupID so we can just name this:


```suggestion
	CBGTIndexTypeSyncGatewayResync    = "syncGateway-resync"
```

base/dcp_sharded.go
```diff
@@ -153,12 +165,17 @@ func createCBGTIndex(ctx context.Context, c *CbgtContext, dbName string, configG
...
-	indexName, previousIndexUUID := dcpSafeIndexName(ctx, c, dbName)
+	indexName, previousIndexUUID := dcpSafeIndexName(ctx, c, dbName, shardedDcpFeedType, feedID)
 	InfofCtx(ctx, KeyDCP, "Creating cbgt index %q for db %q", indexName, MD(dbName))
 
 	// Index types are namespaced by configGroupID to support delete and create of a database targeting the
 	// same bucket in a config group
-	indexType := CBGTIndexTypeSyncGatewayImport + configGroupID
+	var indexType string
+	if shardedDcpFeedType == ResyncShardedDCPFeedType {
+		indexType = CBGTIndexTypeSyncGatewayResync + configGroupID
```
torcolvin: See above, I do not think we want to scope this by groupID.

db/background_mgr_resync_dcp_test.go
```diff
@@ -240,47 +240,63 @@ func TestResyncManagerDCPStart(t *testing.T) {
 		assert.Equal(t, db.DbStats.Database().SyncFunctionCount.Value(), int64(docsToCreate))
 	})
 
-	t.Run("Resync with updated sync function", func(t *testing.T) {
-		docsToCreate := 100
-		db, ctx := setupTestDBForResyncWithDocs(t, docsToCreate, true)
-		defer db.Close(ctx)
+	for _, distributed := range []bool{false, true} {
+		t.Run("Resync with updated sync function", func(t *testing.T) {
+			docsToCreate := 100
```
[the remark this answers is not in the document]
torcolvin: This comment is correct.

db/pindex.go
```diff
@@ -23,25 +23,27 @@ import (
...
-	cbgt.RegisterPIndexImplType(pIndexType,
-		&cbgt.PIndexImplType{
-			New:    getNewPIndexImplType(ctx),
-			Open:   openPIndexImpl,
-			OpenEx: getOpenExPIndexImpl(ctx),
-			Description: "general/syncGateway-import " +
-				" - import processing for shared bucket access",
-		})
+	for _, indexType := range []string{base.CBGTIndexTypeSyncGatewayImport, base.CBGTIndexTypeSyncGatewayResync} {
+		pIndexType := indexType + configGroup
```
torcolvin: We don't want to use config group for resync.

db/background_mgr_resync_dcp.go
```diff
@@ -194,22 +200,104 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
```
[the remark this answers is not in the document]
RIT3shSapata: multiple scopes is not supported

db/background_mgr_resync_dcp.go
```diff
@@ -194,22 +200,104 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+
+		resyncDestFunc := func(janitorRollback func()) (cbgt.Dest, error) {
+			resyncDest, err := base.NewDCPDest(loggingCtx, callback, db.MetadataStore, db.numVBuckets, true, nil, nil, checkPointPrefix)
+			if err != nil {
+				return nil, fmt.Errorf("Error creating resync dest: %v", err)
+			}
+			return resyncDest, nil
+		}
+
+		base.StoreDestFactory(loggingCtx, resyncDestKey, resyncDestFunc)
```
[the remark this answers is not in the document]
RIT3shSapata: @torcolvin curious to hear your thoughts on this

db/background_mgr_resync_dcp.go
```diff
@@ -194,22 +200,104 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+
+		base.InfofCtx(loggingCtx, base.KeyJavascript, "ResyncID: %s Starting DCP resync for bucket: %q ", resyncLoggingID, base.UD(bucket.GetName()))
```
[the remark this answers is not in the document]
RIT3shSapata: @torcolvin do you think it makes sense to have a separate key for resync related logs?

db/background_mgr_resync_dcp.go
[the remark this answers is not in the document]
RIT3shSapata: discussed offline and decided to defer and remove the factory

db/pindex.go
```diff
@@ -23,25 +23,26 @@ import (
...
-	base.InfofCtx(ctx, base.KeyDCP, "Registering PindexImplType for %s", pIndexType)
-	cbgt.RegisterPIndexImplType(pIndexType,
-		&cbgt.PIndexImplType{
-			New:    getNewPIndexImplType(ctx),
-			Open:   openPIndexImpl,
-			OpenEx: getOpenExPIndexImpl(ctx),
-			Description: "general/syncGateway-import " +
-				" - import processing for shared bucket access",
-		})
+	for _, pIndexType := range []string{base.CBGTIndexTypeSyncGatewayImport + configGroup, base.CBGTIndexTypeSyncGatewayResync} {
```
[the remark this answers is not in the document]
RIT3shSapata: updated documentation

db/background_mgr_resync_dcp.go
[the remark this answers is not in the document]
RIT3shSapata: made it key all, additional work related to logging will be done in https://jira.issues.couchbase.com/browse/CBG-5243

base/constants_syncdocs.go
```diff
@@ -224,6 +233,28 @@ func (m *MetadataKeys) SGCfgPrefix(groupID string) string {
...
+// ResyncCfgPrefix returns a document prefix to use for resync cfg documents
+//
+//	format: _sync:{m_$}:resync_cfg:[groupID:]   (collections)
+//	format: _sync:resync_cfg:[groupID:]   (default)
+func (m *MetadataKeys) ResyncCfgPrefix(groupID string) string {
+	if groupID != "" {
+		return m.resyncCfgPrefix + groupID + ":"
+	}
+	return m.resyncCfgPrefix
+}
```
torcolvin, line 250: These are not going to be group id scoped.

base/heartbeat.go
torcolvin: I made this comment before but I'm pretty sure this is true now, we don't have to make this public.

torcolvin:
One more change I just thought of - I think I want to name the checkpoint names differently so that we can see the difference.

torcolvin (approved):

PR #8059, opened 2026-02-06
CBG-5038 allow metadata id to be nil while syncinfo exists

torcolvin:
CBG-5038 allow metadata id to be nil while syncinfo exists

Fix a gap where setting up a non persistent configuration database and running attachment migration would cause the next start of the database to require resync.

attachment migration will write metadata_version and will preserve metadataID, but the metadata id might not exist if non persistent config is used.

With non persistent config, metadataID is not written, but running attachment migration will write a metadata_version.

Turn metadataID into an optional value.


rest/upgradetest/upgrade_registry_test.go
```diff
@@ -295,7 +298,7 @@ func TestMetadataIDWithConfigGroups(t *testing.T) {
 	require.Equal(t, db.RunStateString[db.DBOnline], dbRoot.State)
```
gregns1:
```suggestion
	require.Equal(t, db.RunStateString[db.DBOnline], dbRoot.State, "Database did not come online after upgrade %#+v", dbRoot)
```

gregns1 (changes requested):
Generally looks good to me just one minor test nit

gregns1 (approved):

PR #8071, opened 2026-02-18
Add initial AGENTS.md

bbrks:
Add AGENTS.md (mostly generated, with some manual post-processing/tweaks)

- Information derived from Opus 4.6 exploring repo, [DeepWiki](https://deepwiki.com/couchbase/sync_gateway) and the [public SGW documentation](https://docs.couchbase.com/sync-gateway/current/introduction.html)
- Added symlink stubs for Gemini and Claude - since they don't directly use the generic/standard `AGENTS.md` (yet?)

- n/a

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -118,7 +118,6 @@ These variables are read by Go test code via `os.Getenv` and work with `go test`
 | Variable | Purpose | Default |
 |----------|---------|---------|
 | `SG_TEST_BUCKET_POOL_SIZE` | Number of buckets to pre-create in the test pool | `4` |
-| `SG_TEST_BUCKET_QUOTA_MB` | RAM quota per test bucket (MB) | `200` |
 
 ## Gotchas
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -90,7 +90,8 @@ Never add the `cb_sg_enterprise` tag to test commands unless intentionally testi
 - REST tests use `rest.NewRestTester(t, &RestTesterConfig{...})`.
 - Bucket tests use `base.GetTestBucket(t)`.
 - Tests run with `-shuffle=on` by default.
-- Test timeout defaults to 20 minutes per package.
+- Go tests timeout at 10 minutes per package, but should be increased to 45 minutes if using `SG_TEST_BACKING_STORE=Couchbase`
+- If running more than one test, pipe the test output to a temp file for later analysis and ONLY READ SMALL PARTS OF THIS FILE in the event of a failure, it will be very large! Make sure to clean up the temp log file once you're done with it too.
 
 ### REST API changes
 - When modifying REST handlers, query parameters, or response schemas, update the OpenAPI specs in `docs/api/`.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -117,7 +117,7 @@ These variables are read by Go test code via `os.Getenv` and work with `go test`
 
 | Variable | Purpose | Default |
 |----------|---------|---------|
-| `SG_TEST_BUCKET_POOL_SIZE` | Number of buckets to pre-create in the test pool | `4` |
+| `SG_TEST_BUCKET_POOL_SIZE` | Number of buckets to pre-create in the test pool. Single tests usually only need one or two buckets and will speed up testing. Leave this as default if running more than one test. | `4` |
 
 ## Gotchas
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,7 +5,7 @@ Sync Gateway is a horizontally scalable web server that securely manages access 
 ## Build & Test
 
 - Build: `go build -o bin/sync_gateway .`
-- Building or testing EE (requires private repo SSH access) must use the `cb_sg_enterprise` build tag for all Go commands. E.g: `go build -tags cb_sg_enterprise .`
+- Building or testing EE (requires private repo SSH access) must use the `cb_sg_enterprise, cb_sg_devmode` build tags for all Go commands. E.g: `go build -tags cb_sg_enterprise,cb_sg_devmode .`
 - Run all unit tests (Rosmar/in-memory, no Couchbase Server): `go test ./...`
 - Run a single test: `go test -run ^TestFunctionName$ github.com/couchbase/sync_gateway/rest`
 - Run a single package: `go test github.com/couchbase/sync_gateway/rest`
@@ -73,7 +73,7 @@ Never add the `cb_sg_enterprise` tag to test commands unless intentionally testi
 ## Conventions & Patterns
 
 ### Go style
-- Go 1.25+. Tabs for indentation (standard `gofmt`).
+- Use the Go version declared in `go.mod`. Tabs for indentation (standard `goimports`).
 - 120-char soft line limit (`.editorconfig`).
 - Use `github.com/stretchr/testify` for assertions (`require` for fatal, `assert` for non-fatal).
 - Use stdlib and base utilities where possible unless a module is already referenced by `go.mod`.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -125,4 +125,4 @@ These variables are read by Go test code via `os.Getenv` and work with `go test`
 - EE build requires SSH access to `github.com/couchbaselabs/go-fleecedelta` private repo.
 - Integration tests must always force `-count=1 -p 1` (serial execution) to avoid cross-package interference.
 - Rosmar is an in-memory Couchbase bucket simulator used for unit tests; it lives in `github.com/couchbaselabs/rosmar`.
-  - It does not have 100% feature-parity, so any features heavily using Couchbase-Server should be tested in both modes to ensure compatability.
+  - It does not have 100% feature-parity, so any features heavily using Couchbase-Server should be tested in both modes to ensure compatibility.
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -0,0 +1,128 @@
+# Couchbase Sync Gateway
+
+Sync Gateway is a horizontally scalable web server that securely manages access and synchronization between Couchbase Lite clients and Couchbase Server. Written in Go, it exposes REST and BLIP (WebSocket-based) APIs.
+
+## Build & Test
+
+- Build: `go build -o bin/sync_gateway .`
+- Building or testing EE (requires private repo SSH access) must use the `cb_sg_enterprise` build tag for all Go commands. E.g: `go build -tags cb_sg_enterprise .`
```
torcolvin: should we add 
```suggestion
- Building or testing EE (requires private repo SSH access) must use the `cb_sg_enterprise, cb_sg_devmode` build tags for all Go commands. E.g: `go build -tags cb_sg_enterprise,cb_sg_devmode .`
```
?

we can add to a separate line

AGENTS.md
```diff
@@ -0,0 +1,128 @@
...
+- Run all unit tests (Rosmar/in-memory, no Couchbase Server): `go test ./...`
+- Run a single test: `go test -run ^TestFunctionName$ github.com/couchbase/sync_gateway/rest`
+- Run a single package: `go test github.com/couchbase/sync_gateway/rest`
+- Run integration tests (requires local Couchbase Server): `SG_TEST_BACKING_STORE=Couchbase go test ./...`
+- Run a specific benchmark: `go test -bench=^BenchmarkSomething$ -run=- ./...`
+- Lint: `golangci-lint run`
+- Python tooling lint/typecheck: `uv run ruff check tools/ tools-tests/` and `uv run mypy`
+- Python tests: `uv run pytest`
```
torcolvin, line 16: Do we need to add sgcollect?

AGENTS.md
```diff
@@ -0,0 +1,128 @@
...
+## Build & Test
+
+- Build: `go build -o bin/sync_gateway .`
+- Building or testing EE (requires private repo SSH access) must use the `cb_sg_enterprise` build tag for all Go commands. E.g: `go build -tags cb_sg_enterprise .`
+- Run all unit tests (Rosmar/in-memory, no Couchbase Server): `go test ./...`
+- Run a single test: `go test -run ^TestFunctionName$ github.com/couchbase/sync_gateway/rest`
+- Run a single package: `go test github.com/couchbase/sync_gateway/rest`
+- Run integration tests (requires local Couchbase Server): `SG_TEST_BACKING_STORE=Couchbase go test ./...`
+- Run a specific benchmark: `go test -bench=^BenchmarkSomething$ -run=- ./...`
+- Lint: `golangci-lint run`
```
torcolvin, line 14: should we do --config=.golangci-strict.yml ?

AGENTS.md
```diff
@@ -0,0 +1,128 @@
...
+Sync Gateway ships two editions controlled by the `cb_sg_enterprise` build tag:
+- **CE** (Community Edition): default, no build tag needed. Files: `*_ce.go`
+- **EE** (Enterprise Edition): `go build -tags cb_sg_enterprise`. Files: `*_ee.go`
+
+Never add the `cb_sg_enterprise` tag to test commands unless intentionally testing EE features.
+
+## Conventions & Patterns
+
+### Go style
+- Go 1.25+. Tabs for indentation (standard `gofmt`).
```
torcolvin:
```suggestion
- Go 1.25+. Tabs for indentation (standard `goimports`).
```

torcolvin: Should we say to use a go version in go.mod so we don't have to update this?

AGENTS.md
```diff
@@ -0,0 +1,128 @@
...
+- Wrap User Data (doc IDs, usernames, PII) with `base.UD()`.
+- Wrap Metadata (db names, config keys) with `base.MD()`.
+
+### Testing patterns
+- Each top-level package has a `main_test.go` with `TestMain` that sets up `TestBucketPool`.
+- Default test backing store is **Rosmar** (in-memory). Set `SG_TEST_BACKING_STORE=Couchbase` for CBS.
+- REST tests use `rest.NewRestTester(t, &RestTesterConfig{...})`.
+- Bucket tests use `base.GetTestBucket(t)`.
+- Tests run with `-shuffle=on` by default.
+- Test timeout defaults to 20 minutes per package.
```
torcolvin:
```suggestion
- Test timeout defaults to 10 minutes per package unless SG_TEST_BACKING_STORE=Couchbase where it is 45 seconds.
```

AGENTS.md
```diff
@@ -0,0 +1,128 @@
...
+|----------|---------|---------|
+| `SG_TEST_BACKING_STORE` | Set to `Couchbase` to test against a real CBS cluster instead of Rosmar | Rosmar (in-memory) |
+| `SG_TEST_COUCHBASE_SERVER_URL` | Couchbase Server URL for integration tests | `couchbase://localhost` |
+
+**Integration test tuning** (only relevant with `SG_TEST_BACKING_STORE=Couchbase`):
+
+| Variable | Purpose | Default |
+|----------|---------|---------|
+| `SG_TEST_BUCKET_POOL_SIZE` | Number of buckets to pre-create in the test pool | `4` |
+| `SG_TEST_BUCKET_QUOTA_MB` | RAM quota per test bucket (MB) | `200` |
```
torcolvin:
```suggestion
```

I don't think this is important

AGENTS.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Gotchas
+
+- EE build requires SSH access to `github.com/couchbaselabs/go-fleecedelta` private repo.
+- Integration tests must always force `-count=1 -p 1` (serial execution) to avoid cross-package interference.
```
torcolvin, line 126:
```suggestion
- Integration tests must always force `-p 1` (serial execution) to avoid cross-package interference.
```

AGENTS.md
bbrks: I was trying to figure out what would be best for an agent to use, and it's probably the default lint config which shows a wider range of issues, rather than a smaller but stricter (i.e. fail CI) subset?

AGENTS.md
bbrks: Elaborate please. Add a new bulletpoint describing sgcollect, or are you suggesting we remove these python lines?

I'd probably appreciate the extra help for Python/sgcollect changes via an agent.

AGENTS.md
bbrks:
```suggestion
- Use the Go version declared in `go.mod`. Tabs for indentation (standard `goimports`).
```

AGENTS.md
bbrks: Huh, where is that even defined? Why would a package using CB timeout after 45s???

AGENTS.md
bbrks: agree

AGENTS.md
bbrks: I would rather not deal with test caching issues if an agent is running tests.

AGENTS.md
```diff
@@ -0,0 +1,128 @@
...
+|----------|---------|---------|
+| `SG_TEST_BACKING_STORE` | Set to `Couchbase` to test against a real CBS cluster instead of Rosmar | Rosmar (in-memory) |
+| `SG_TEST_COUCHBASE_SERVER_URL` | Couchbase Server URL for integration tests | `couchbase://localhost` |
+
+**Integration test tuning** (only relevant with `SG_TEST_BACKING_STORE=Couchbase`):
+
+| Variable | Purpose | Default |
+|----------|---------|---------|
+| `SG_TEST_BUCKET_POOL_SIZE` | Number of buckets to pre-create in the test pool | `4` |
+| `SG_TEST_BUCKET_QUOTA_MB` | RAM quota per test bucket (MB) | `200` |
```
bbrks:
```suggestion
```

AGENTS.md
torcolvin: I think this is fine, I guess I was just thinking to note that the python code is not typically needed to look at for development and it should not be scanned typically.

AGENTS.md
torcolvin: So I think this is worth mentioning something for Couchbase integration test, the default is whatever go test does.

If I forget -timeout 45m when running the tests like jenkins integration tests do, 10m is not enough.

AGENTS.md
torcolvin: I don't know that I have ever had a test caching issue?

Re-running and re-parsing the output, which is verbose, is quite expensive for AI or a human.

I don't run -count 1 typically, only use -count if there's a known flaky test. In that case, I can ask to rerun the tests to catch flakes. If this needs to be fixed, I'd rather fix it for humans as well as bots.

AGENTS.md
bbrks: Probably something for a specialized skill instead of generic agent but should be good enough for now and isn't bloating the agent context much.

AGENTS.md
bbrks:
```suggestion
- Go tests timeout at 10 minutes per package, but should be increased to 45 minutes if using `SG_TEST_BACKING_STORE=Couchbase`
- If running more than one test, pipe the test output to a temp file for later analysis and ONLY READ SMALL PARTS OF THIS FILE in the event of a failure, it will be very large! Make sure to clean up the temp log file once you're done with it too.
```

Added some guidance - and also tried to avoid piping huge log outputs into the LLM. That sounds expensive...

AGENTS.md
bbrks: For Rosmar this is mostly not an issue, but for tests relying on an external resource like Couchbase Server - test caching can really hide/obscure issues, since the cache isn't invalidated if the environment the test is running against changes.

torcolvin (approved):

PR #8074, opened 2026-02-19
Reduce duplication in DCP code

torcolvin:
Creates common code for sharded DCP to share between import and resync.

- Create `CollectionNames` struct for common use. This type already exists in `sgbucket.FeedArguments` and probably could be pushed to sg-bucket in the future. It would probably be better to have a type that is `map[string]map[string]{}struct` in the future.
- Add some error checking for ResyncManger options structs.
- move `ctx = CorrelationIDLogCtx(ctx, DCPImportFeedID)` back to import listener in preparation for resync usage

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/323/ (known test flake in TestResyncWithoutIndexes)


db/database.go
```diff
@@ -2557,6 +2563,20 @@ func (db *DatabaseContext) EnableAllowConflicts(tb testing.TB) {
...
+	return base.IsEnterpriseEdition() && !db.BucketSpec.IsWalrusBucket()
+}
+
+// collectionNames returns the names of the collections on this database.
+func (db *DatabaseContext) collectionNames() base.CollectionNames {
+	names := base.NewCollectionNames()
+	for _, col := range db.CollectionByID {
+		names.Add(col.dataStore)
+	}
+	return names
```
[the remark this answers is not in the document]
torcolvin: I don't think sorting is necessary here, we can sort later if required. As a reviewer, make sure the code is sorted where needed.

RIT3shSapata (dismissed):

torcolvin:
Deleted the test `TestGetDatabaseCollectionWithUserDefaultCollection` because it is no longer useful after restricting the database to one scope. Additionally, this code is covered everytime a `DatabaseContext` is created since `Scopes` is required.

base/util_testing.go
```diff
@@ -824,68 +822,6 @@ func SkipImportTestsIfNotEnabled(t *testing.T) {
 	}
 }
 
-// CreateBucketScopesAndCollections will create the given scopes and collections within the given BucketSpec.
-func CreateBucketScopesAndCollections(ctx context.Context, bucketSpec BucketSpec, scopes map[string][]string) error {
```
RIT3shSapata, line 828: Why is this utility function being removed?

torcolvin: I tried to use this function in `TestGetDatabaseCollectionWithUserDefaultCollection` to replace code that existing by creating a datastore on `_default.customdatastore`. I actually found in jenkins that it would fail sometimes. The code is ultimately duplicated by `Bucket.CreateDataStore` and I'd rather have it one place.

The code in `Bucket.CreateDataStore` can still fail, and I wasn't able to get to the bottom of why. It seems like `gocb.Bucket("foo").Scope("_default").Collection("customdatastore").Exists("foo")` would pass indicating it existed, but then SG `Collection.setCollectionID` would fail which is called in `NewCollection` with the collection not existing. I assume this is somehow a mismatch between kv and ns_server but I didn't get to the bottom of this.

This code change isn't really relevant to the commit as of now, but I do not regret removing it.

RIT3shSapata (approved):

PR #8075, opened 2026-02-20
Decouple BucketSpec from DatabaseContext

torcolvin:
Decouple BucketSpec from DatabaseContext

This fixes an issue where `DatabaseContext.BucketSpec` was only set by the initialization of `NewDatabaseContext` in `server_context.go`. This meant that a call like `db.BucketSpec.IsWalrusBucket()` would only be valid if called via `server_context.go` and not in the db package. This addresses problems running unit tests, but does not affect production code.

I am planning to add a check for using rosmar bucket to avoid sharded DCP code, and it was adding this code where I realized there was a problem.

I considered alternate approaches, and I'm open to these or others:

1. Requiring `BucketSpec` as a parameter to `DatabaseContextOptions`. This would work but I would have had to modify more test code.
2. Constructing a bucket spec inside `NewDatabaseContext` by either finding bucketSpec after type assertion to `GocbV2Bucket` or reconstructing the value from rosmar. The problem is that once a rosmar bucket has been opened, I would have to modify `rosmar.Bucket` to be able to find `BucketSpec.Server` https://github.com/couchbase/sync_gateway/blob/61cf3ebcca647463e4cc2e116cd5c127f2a8c27e/base/bucket.go#L287
3. Move `BucketSpec` code into sgbucket and add `BucketStore.GetSpec` interface. However, the `BucketSpec` is pretty much only gocb options, with the exception of Server (typically rosmar in memory URL, but can be an on disk URL). This felt overkill.

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/281/

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -106,7 +106,7 @@ type DatabaseContext struct {
 	UUID                        string             // UUID for this database instance. Used by cbgt and sgr
 	MetadataStore               base.DataStore     // Storage for database metadata (anything that isn't an end-user's/customer's documents)
 	Bucket                      base.Bucket        // Storage
-	BucketSpec                  base.BucketSpec    // The BucketSpec
+	bucketUsername              string             // name of the connecting user for audit logging
 	BucketUUID                  string             // The bucket UUID for the bucket the database is created against
 	EncodedSourceID             string             // The md5 hash of bucket UUID + cluster UUID for the bucket/cluster the database is created against but encoded in base64
 	BucketLock                  sync.RWMutex       // Control Access to the underlying bucket object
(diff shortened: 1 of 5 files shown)


torcolvin:
This has some implications on code I wanted to do for https://github.com/couchbase/sync_gateway/pull/8074 / https://github.com/couchbase/sync_gateway/pull/8057 but I can drop this and determine in a better way if a bucket is a rosmar bucket and add this to DatabaseContext struct if I don't want to do `AsRosmarBucket(db.Bucket)`.

However, the code in https://github.com/couchbase/sync_gateway/blob/d4290f6a879d989eba94a7e91d98ac065733996f/db/database.go#L2140 is wrong for logging in the `db` package since spec.Server == ""

bbrks (approved):

PR #8082, opened 2026-02-25
Test: Change default Couchbase Server URL to 127.0.0.1 from localhost

bbrks:
For extremely annoying and entirely unnecessary reasons, having the VPN on causes `localhost` lookups to take forever (>100ms) to resolve for me despite several complaints.

This has mostly been worked around by me manually setting `SG_TEST_COUCHBASE_SERVER_URL=couchbase://127.0.0.1` for a couple of years now, but I think this change might as well just go in. It's effectively the same but avoids DNS entirely.

- n/a


base/constants.go
```diff
@@ -25,7 +25,7 @@ const (
 	GuestUsername = "GUEST"
 	ISO8601Format = "2006-01-02T15:04:05.000Z07:00"
 
-	kTestCouchbaseServerURL = "couchbase://localhost"
+	kTestCouchbaseServerURL = "couchbase://127.0.0.1"
```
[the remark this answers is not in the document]
bbrks: 👍 Applied a find/replace for all instances of this

examples/legacy_config/basic-couchbase-bucket.json
```diff
@@ -7,7 +7,7 @@
   },
   "databases": {
     "db": {
-      "server": "couchbase://localhost",
+      "server": "couchbase://127.0.0.1",
```
torcolvin: I'm not so sure I would change the example documentation here.

torcolvin (dismissed):
I think this is fine, I can't think of a case where switching to an IP is bad. If we enhance the test framework to use certificate based auth with tls verification, I think there might be some trickiness but the bucket pool code does not consistently work in this way right now.

I'd be inclined not to change an example files though.

I tested this with local docker setup on mac and linux with `./integration-test/start_server.sh` and a basic bucket connection test and it seems to work.

examples/legacy_config/basic-couchbase-bucket.json
bbrks: Gah thanks - I'd tried to ignore config/docs/examples like this but my find/replace skills were lacking :)

torcolvin (approved):

PR #8083, opened 2026-02-26
CBG-5192: add ability to identify if pushed encoded CV is derived from our local revID

gregns1:
CBG-5192

- CBL were actually replying on some buggy behaviour form the rev cache ion version 4.0.0 - 4.0.2. With the old behaviour the documnet was being added to CBS with 3-abc@RTE as CV and 4-abc as revID. But due to bug in rev cache we were linking the rev cache value as 3-abc@RTE  and revID 3-abc. So the rest api fetch of the doc post replication was returning stale revID. So the overall replication behaviour hasn’t changes between versions just a bug being fixed at the rev cache. 
- This doesn’t take away from the fact that a resolved conflict on CBL for remote wins should not result in new mutation being pushed top SGW. The issue we have is CBL is resolving conflict and saving the remote revID as rev tree encoded CV. We currently have no handling on SGW to see that this proposed change is equal to our local revID. 
- We end up with the following being pushed in proposed changes after resolved conflict [3-abc@RTE, 3-abc, docID]
- SGW currently just sees the server revID CBL sends matches our revID and accepts this as new mutation. We should eb comparing encoded CV being sent against our revID before accepting change. 

Sheets doc explains the state of each step above:
https://docs.google.com/spreadsheets/d/12cHAi8xR7huqUYVfDflXKxejQnVVwH8glTGsdAiX4nM/edit?usp=sharing

Ran this against the test this was found in and passes. 

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/292/

### DIFF
--- a/rest/blip_legacy_revid_test.go
+++ b/rest/blip_legacy_revid_test.go
@@ -1087,7 +1087,6 @@ func TestLegacyRevBlipTesterClient(t *testing.T) {
 }
 
 func TestCBLPushEncodedCVDerivedFromSGWLocalRevID(t *testing.T) {
-	base.SetUpTestLogging(t, base.LevelTrace, base.KeySyncMsg, base.KeyCRUD)
 	btcRunner := NewBlipTesterClientRunner(t)
 
 	btcRunner.SkipSubtest[RevtreeSubtestName] = true // vv specific test
--- a/rest/blip_legacy_revid_test.go
+++ b/rest/blip_legacy_revid_test.go
@@ -308,6 +308,47 @@ func TestProcessLegacyRev(t *testing.T) {
 	})
 }
 
+func TestSendUnsolicitedRevWithRTEDerivedFromLocalRevID(t *testing.T) {
+	bt := NewBlipTesterFromSpec(t, BlipTesterSpec{
+		allowConflicts: false,
+		GuestEnabled:   true,
+		blipProtocols:  []string{db.CBMobileReplicationV4.SubprotocolString()},
+	})
+	defer bt.Close()
+	rt := bt.restTester
+
+	doc := rt.CreateDocNoHLV(t.Name(), db.Body{"key": "val"})
+	sgwVersion := doc.ExtractDocVersion()
+
+	encodedVersion, err := db.LegacyRevToRevTreeEncodedVersion(sgwVersion.RevTreeID)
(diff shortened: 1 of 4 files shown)


rest/blip_legacy_revid_test.go
```diff
@@ -1045,6 +1045,45 @@ func TestLegacyRevBlipTesterClient(t *testing.T) {
 	})
 }
 
+func TestCBLPushEncodedCVDerivedFromSGWLocalRevID(t *testing.T) {
+	base.SetUpTestLogging(t, base.LevelTrace, base.KeySyncMsg, base.KeyCRUD)
```
torcolvin:
```suggestion
```

rest/utilities_testing_blip_client.go
```diff
@@ -1648,7 +1649,13 @@ func (btcc *BlipTesterCollectionClient) upsertDoc(docID string, opts blipTesterU
 			digest = opts.specificDigest
 		}
 		newRevID := fmt.Sprintf("%d-%s", newGen, digest)
-		docVersion = DocVersion{RevTreeID: newRevID}
+		if opts.writeAsEncodedCV {
```
torcolvin: I don't feel like this is the right place for this if statement because it is adding something as an HLV, but it's ignoring the previous versions that come before it.

I feel like what needs to happen is closer to the block above where `newVersion` is passed in:

```
	if btcc.UseHLV() {
		require.NoError(btcc.TB(), hlv.AddVersion(newVersion))
		docVersion = DocVersion{CV: *hlv.ExtractCurrentVersionFromHLV()}
```

I will think on this a bit more, I think that potentially a way to do this is to change the option with:

blipTesterUpsertOptions.revType

and have:

* auto (pick from version)
* legacyrevid (force non hlv doc version)
* encodedrevid

torcolvin (commented):
I definitely think this is the right behavior but I think you need to add more tests for other circumstances.

* What happens if `changes` is sent with this rather than `proposeChanges` ? Do you need to check the revision?
* What happens if there is an unsolicited rev?

torcolvin (approved):
LGTM, I think you should re-run jenkins after the test changes to be sure.

PR #8084, opened 2026-02-26
CBG-5225: sg-bucket KV range scan implementations

bbrks:
**Note:** This was coded primarily using Opus 4.6 - with manual guidance/verification/review.

Adds a test that can do a KV range scan on both Couchbase via Gocb and Rosmar via an sg-bucket interface.

## Dependencies
- [x] https://github.com/couchbase/sg-bucket/pull/139
- [x] https://github.com/couchbaselabs/rosmar/pull/57

- [x] https://jenkins.sgwdev.com/job/SyncGatewayIntegration/653/

### DIFF
--- a/base/collection_rangescan.go
+++ b/base/collection_rangescan.go
@@ -39,7 +39,7 @@ func (c *Collection) Scan(_ context.Context, scanType sgbucket.ScanType, opts sg
 		return nil, err
 	}
 
-	return &gocbScanResultIterator{result: result, idsOnly: opts.IDsOnly}, nil
+	return &gocbScanResultIterator{result: result}, nil
 }
 
 func toGocbScanType(scanType sgbucket.ScanType) (gocb.ScanType, error) {
@@ -59,9 +59,8 @@ func toGocbScanType(scanType sgbucket.ScanType) (gocb.ScanType, error) {
 }
 
 type gocbScanResultIterator struct {
-	result  *gocb.ScanResult
-	idsOnly bool
-	err     error
+	result *gocb.ScanResult
+	err    error
 }
 
 func (it *gocbScanResultIterator) Next(_ context.Context) *sgbucket.ScanResultItem {
@@ -76,7 +75,7 @@ func (it *gocbScanResultIterator) Next(_ context.Context) *sgbucket.ScanResultIt
 		ID:  item.ID(),
 		Cas: uint64(item.Cas()),
 	}
-	if !it.idsOnly {
+	if !item.IDOnly() {
 		if err := item.Content(&result.Body); err != nil {
 			it.err = fmt.Errorf("failed to decode scan result body for %s: %w", item.ID(), err)
 			return nil
--- a/base/collection_rangescan.go
+++ b/base/collection_rangescan.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2025-Present Couchbase, Inc.
+
+Use of this software is governed by the Business Source License included in
+the file licenses/BSL-Couchbase.txt.  As of the Change Date specified in that
+file, in accordance with the Business Source License, use of this software will
+be governed by the Apache License, Version 2.0, included in the file
+licenses/APL2.txt.
+*/
+
+package base
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/couchbase/gocb/v2"
+	sgbucket "github.com/couchbase/sg-bucket"
+)
+
+var _ sgbucket.RangeScanStore = &Collection{}
+
+func (c *Collection) Scan(_ context.Context, scanType sgbucket.ScanType, opts sgbucket.ScanOptions) (sgbucket.ScanResultIterator, error) {
+	c.Bucket.waitForAvailKvOp()
+	defer c.Bucket.releaseKvOp()
+
+	gocbScanType, err := toGocbScanType(scanType)
+	if err != nil {
+		return nil, err
+	}
+
+	scanOpts := &gocb.ScanOptions{
+		IDsOnly:    opts.IDsOnly,
+		Transcoder: gocb.NewRawBinaryTranscoder(),
+	}
+
+	result, err := c.Collection.Scan(gocbScanType, scanOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	return &gocbScanResultIterator{result: result, idsOnly: opts.IDsOnly}, nil
+}
+
+func toGocbScanType(scanType sgbucket.ScanType) (gocb.ScanType, error) {
+	switch st := scanType.(type) {
+	case sgbucket.RangeScan:
+		rs := gocb.RangeScan{}
+		if st.From != nil {
+			rs.From = &gocb.ScanTerm{Term: st.From.Term, Exclusive: st.From.Exclusive}
+		}
+		if st.To != nil {
+			rs.To = &gocb.ScanTerm{Term: st.To.Term, Exclusive: st.To.Exclusive}
+		}
+		return rs, nil
+	default:
+		return nil, fmt.Errorf("unsupported scan type: %T", scanType)
+	}
+}
+
+type gocbScanResultIterator struct {
+	result  *gocb.ScanResult
+	idsOnly bool
+	err     error
+}
+
+func (it *gocbScanResultIterator) Next(_ context.Context) *sgbucket.ScanResultItem {
+	if it.err != nil {
+		return nil
+	}
+	item := it.result.Next()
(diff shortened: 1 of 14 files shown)


base/collection_rangescan.go
```diff
@@ -0,0 +1,86 @@
...
+	result := &sgbucket.ScanResultItem{
+		ID:     item.ID(),
+		Cas:    uint64(item.Cas()),
+		IDOnly: item.IDOnly(),
+	}
+	if !item.IDOnly() {
+		var body []byte
+		if err := item.Content(&body); err == nil {
+			result.Body = body
+		}
```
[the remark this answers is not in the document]
torcolvin: Does gocb.RangeScan ignore tombstones? Do we care about tombstones for usage? If so, we should probably align the behavior in rosmar.

I would write a test to see what happens if there is a literal `null` body and a tombstone and then make sure rosmar and gocb are aligned in behavior.

base/collection_rangescan.go
```diff
@@ -0,0 +1,86 @@
...
+		return rs, nil
+	default:
+		return nil, fmt.Errorf("unsupported scan type: %T", scanType)
+	}
+}
+
+type gocbScanResultIterator struct {
+	result  *gocb.ScanResult
+	idsOnly bool
```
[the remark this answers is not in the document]
torcolvin: Agree with this comment.

base/collection_rangescan_test.go
```diff
@@ -0,0 +1,337 @@
...
+		"doc_d": []byte(`{"name":"delta"}`),
+		"doc_e": []byte(`{"name":"echo"}`),
+	}
+	for k, v := range docs {
+		require.NoError(t, dataStore.SetRaw(k, 0, nil, v))
+	}
+
+	allDocIDs := []string{"doc_a", "doc_b", "doc_c", "doc_d", "doc_e"}
+
+	// CBS range scan may not immediately reflect recent writes (requires persistence).
```
torcolvin: Is this true? This seems perilous if so. Does this depend of the type of storage used by the backing bucket?

base/collection_rangescan_test.go
```diff
@@ -0,0 +1,337 @@
...
+	require.NoError(t, err)
+}
+
+// TestRangeScanRBACUser verifies that KV range scan operations succeed when the bucket
+// connection uses an RBAC user with the mobile_sync_gateway role (EE) or
+// bucket_full_access role (CE), rather than the cluster administrator.
+func TestRangeScanRBACUser(t *testing.T) {
+	if UnitTestUrlIsWalrus() {
+		t.Skip("Requires Couchbase Server")
+	}
```
torcolvin:
```suggestion
	TestsRequireMobileRBAC(t)
```

base/collection_rangescan_test.go
```diff
@@ -0,0 +1,337 @@
...
+
+	SetUpTestLogging(t, LevelDebug, KeyAll)
```
torcolvin: do not like this pattern for committing this code because gocb / rosmar logging is long

base/collection_rangescan_test.go
```diff
@@ -0,0 +1,337 @@
...
+	roles := []string{
+		fmt.Sprintf("bucket_full_access[%s]", bucketName),
+		fmt.Sprintf("bucket_admin[%s]", bucketName),
+	}
+	if !TestsUseServerCE() {
+		roles = append(roles, fmt.Sprintf("mobile_sync_gateway[%s]", bucketName))
+	}
+
+	makeRBACUser(t, httpClient, mgmtEp, rbacUsername, rbacPassword, roles)
+	defer deleteRBACUser(t, httpClient, mgmtEp, rbacUsername)
```
torcolvin: Move `rest.MakeUser` and `rest.DeleteUser` to base package.

base/collection_rangescan_test.go
```diff
@@ -0,0 +1,337 @@
...
+
+	// --- Open a second connection to the same bucket using RBAC credentials ---
+	// RBAC user creation may take a moment to propagate to the KV service,
+	// so retry the connection a few times before giving up.
```
torcolvin: I don't think this is true either because we don't do this in the REST package.

base/collection_rangescan_test.go
```diff
@@ -0,0 +1,337 @@
...
+		BucketName: bucketName,
+		Auth: TestAuthenticator{
+			Username:   rbacUsername,
+			Password:   rbacPassword,
+			BucketName: bucketName,
+		},
+		UseXattrs:       bucket.BucketSpec.UseXattrs,
+		TLSSkipVerify:   bucket.BucketSpec.TLSSkipVerify,
+		BucketOpTimeout: bucket.BucketSpec.BucketOpTimeout,
+	}
```
torcolvin: I would call `getTestBucketSpec`. Without this, if your cluster requires a TLS cert it would fail.

base/collection_rangescan_test.go
```diff
@@ -0,0 +1,337 @@
...
+	var rbacBucket *GocbV2Bucket
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		var connectErr error
+		rbacBucket, connectErr = GetGoCBv2Bucket(ctx, rbacSpec)
```
torcolvin: I don't think this is necessary to do in a retry loop, but I'm curious if so.

base/collection_rangescan_test.go
```diff
@@ -0,0 +1,337 @@
...
+		rbacDataStore = rbacBucket.DefaultDataStore()
+	} else {
+		var dsErr error
+		rbacDataStore, dsErr = rbacBucket.NamedDataStore(ScopeAndCollectionName{
+			Scope:      adminDataStore.ScopeName(),
+			Collection: adminDataStore.CollectionName(),
+		})
+		require.NoError(t, dsErr, "Failed to open named collection %s.%s on RBAC bucket",
+			adminDataStore.ScopeName(), adminDataStore.CollectionName())
+	}
```
torcolvin: ds := rbacBucket.GetSingleDataStore()

base/collection_rangescan_test.go
```diff
@@ -0,0 +1,337 @@
...
+			if item == nil {
+				break
+			}
+			ids = append(ids, item.ID)
+			assert.True(t, item.IDOnly)
+			assert.Nil(t, item.Body, "Expected nil body for IDsOnly scan, key %s", item.ID)
+		}
+		sort.Strings(ids)
+		require.Equal(t, allDocIDs, ids)
+	})
```
torcolvin: It feels worth parameterizing this whole test with the standard bucket connection above rather than duplicate.

base/leaky_datastore.go
```diff
@@ -547,7 +547,16 @@ func (lds *LeakyDataStore) GetIndexes() (indexes []string, err error) {
...
+	if !ok {
+		return nil, fmt.Errorf("underlying datastore %T does not support range scan", lds.dataStore)
+	}
+	return rss.Scan(scanType, opts)
+}
+
 // Assert interface compliance:
 var (
-	_ sgbucket.DataStore = &LeakyDataStore{}
+	_ sgbucket.DataStore      = &LeakyDataStore{}
```
torcolvin, line 570: Would it make it easier of sgbucket.DataStore just implemented RangeScanStore so we don't have to check for compliance everywhere?

bbrks: Probably - will keep for now for consistency.

We can think about whether these finer-grained sgbucket interfaces make sense more generally

gregns1 (dismissed):
LGTM, will let you update dependencies where and will re-approve

base/collection_rangescan_test.go
bbrks: there's no simple scan consistency setting available to use to guard against this. We could pass in vb snapshots to avoid - but I don't think it's worth it given we're not expecting to one-shot the range scan to migrate all data in one go

it will be an iterative process until no eligible data is remaining to move (i.e. guard against any writes that may be coming in underneath a running range scan)

base/collection_rangescan_test.go
bbrks: This is actually true - without the retry I get consistent failures. May be related to fastFail retry strategy but that's a different issue/fix.

base/collection_rangescan_test.go
bbrks: without the retry I get consistent failures. May be related to fastFail retry strategy but that's a different issue/fix.

gregns1 (approved):

PR #8093, opened 2026-03-04
CBG-4651: change _online/_offline endpoints to work for both persistent and non persistent config

gregns1:
CBG-4651

- Change for online/offline endpoints to be aligned between persistent and non persistent config 
- Each endpoint will alter the db config to set offline true/false. Persistent config will persist this change to be picked up by other nodes, non persistent config will not 
- Both modes will disconnect clients as the db is reloaded for each mode

## Pre-review checklist

- [ ] Link upstream PRs

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/318/

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -855,47 +855,6 @@ func (context *DatabaseContext) NotifyTerminatedChanges(ctx context.Context, use
 	context.mutationListener.NotifyCheckForTermination(ctx, base.SetOf(base.UserPrefixRoot+username))
 }
 
-func (dc *DatabaseContext) TakeDbOffline(ctx context.Context, reason string) error {
-
-	if atomic.CompareAndSwapUint32(&dc.State, DBOnline, DBStopping) {
-		// notify all active _changes feeds to close
-		close(dc.ExitChanges)
-
-		// Block until all current calls have returned, including _changes feeds
-		dc.AccessLock.Lock()
-		defer dc.AccessLock.Unlock()
-
-		dc.changeCache.Stop(ctx)
-
-		// set DB state to Offline
-		atomic.StoreUint32(&dc.State, DBOffline)
-
-		if err := dc.EventMgr.RaiseDBStateChangeEvent(ctx, dc.Name, "offline", reason, dc.Options.AdminInterface); err != nil {
-			base.InfofCtx(ctx, base.KeyCRUD, "Error raising database state change event: %v", err)
-		}
-
-		return nil
-	} else {
-		dbState := atomic.LoadUint32(&dc.State)
-		// If the DB is already transitioning to: offline or is offline silently return
-		if dbState == DBOffline || dbState == DBResyncing || dbState == DBStopping {
-			return nil
-		}
-
-		msg := "Unable to take Database offline, database must be in Online state but was " + RunStateString[dbState]
-		if dbState == DBOnline {
-			msg = "Unable to take Database offline, another operation was already in progress. Please try again."
-		}
-
-		base.InfofCtx(ctx, base.KeyCRUD, "%s", msg)
-		return base.NewHTTPError(http.StatusServiceUnavailable, msg)
-	}
-}
-
-func (db *Database) TakeDbOffline(nonContextStruct base.NonCancellableContext, reason string) error {
-	return db.DatabaseContext.TakeDbOffline(nonContextStruct.Ctx, reason)
-}
-
 func (context *DatabaseContext) Authenticator(ctx context.Context) *auth.Authenticator {
 	context.BucketLock.RLock()
 	defer context.BucketLock.RUnlock()
--- a/db/database.go
+++ b/db/database.go
@@ -129,7 +129,6 @@ type DatabaseContext struct {
 	AttachmentCompactionManager *BackgroundManager
 	AttachmentMigrationManager  *BackgroundManager
 	AsyncIndexInitManager       *BackgroundManager
-	ExitChanges                 chan struct{}        // Active _changes feeds on the DB will close when this channel is closed
 	OIDCProviders               auth.OIDCProviderMap // OIDC clients
 	LocalJWTProviders           auth.LocalJWTProviderMap
 	ServerUUID                  string // UUID of the server, if available
@@ -2415,8 +2414,6 @@ func (db *DatabaseContext) StartOnlineProcesses(ctx context.Context) (returnedEr
 	if err := base.RequireNoBucketTTL(ctx, db.Bucket); err != nil {
 		return err
 	}
-
-	db.ExitChanges = make(chan struct{})
 
 	// Start checking heartbeats for other nodes.  Must be done after caching feed starts, to ensure any removals
 	// are detected and processed by this node.
--- a/docs/api/paths/admin/db-_offline.yaml
+++ b/docs/api/paths/admin/db-_offline.yaml
@@ -10,7 +10,9 @@ parameters:
 post:
   summary: Take the database offline
   description: |-
-    If using persistent config this will alter the database config in the bucket to take the database offline across all nodes. If you're not using persistent config, this will bring the database online on this node only. Actions can be taken without disrupting current operations ungracefully or having the restart the Sync Gateway instance.
+    Transition the state of the database to offline. This database will no longer accept traffic and only maintenance operations can be performed. This operation blocks until all REST traffic is idle.
+
+    If using persistent config this will affect all Sync Gateway nodes, otherwise this is operation is local to the node.
 
     This will not take the backing Couchbase Server bucket offline.
 
--- a/docs/api/paths/admin/db-_offline.yaml
+++ b/docs/api/paths/admin/db-_offline.yaml
@@ -10,9 +10,7 @@ parameters:
 post:
   summary: Take the database offline
   description: |-
-    This will take the database offline on this node only. Actions can be taken without disrupting current operations ungracefully or having the restart the Sync Gateway instance.
-
-    If using persistent config, call [POST /{db}/_config](#operation/post_db-_config) with `{"offline": true}` to set the database to offline.
+    If using persistent config this will alter the database config in the bucket to take the database offline across all nodes. If you're not using persistent config, this will bring the database online on this node only. Actions can be taken without disrupting current operations ungracefully or having the restart the Sync Gateway instance.
 
     This will not take the backing Couchbase Server bucket offline.
 
--- a/docs/api/paths/admin/db-_online.yaml
+++ b/docs/api/paths/admin/db-_online.yaml
@@ -12,7 +12,7 @@ post:
   description: |-
     Transition the state of the database to online.
 
-    If using persistent config this will affect all Sync Gateway nodes, otherwise this is operation is local to the node.
+    If using persistent config this will affect all Sync Gateway nodes, otherwise this operation is local to the node.
 
     Bringing a database online will:
     * Close the database connection to the backing Couchbase Server bucket.
--- a/docs/api/paths/admin/db-_online.yaml
+++ b/docs/api/paths/admin/db-_online.yaml
@@ -10,7 +10,9 @@ parameters:
 post:
   summary: Bring the database online
   description: |-
-    If using persistent config this will alter the database config in the bucket to take the database online across all nodes. If you're not using persistent config, this will bring the database online on this node only.
+    Transition the state of the database to online.
+
+    If using persistent config this will affect all Sync Gateway nodes, otherwise this is operation is local to the node.
 
     Bringing a database online will:
     * Close the database connection to the backing Couchbase Server bucket.
--- a/docs/api/paths/admin/db-_online.yaml
+++ b/docs/api/paths/admin/db-_online.yaml
@@ -10,9 +10,7 @@ parameters:
 post:
   summary: Bring the database online
   description: |-
-    This will bring the database online on this node only so the Public and full Admin REST API requests can be served.
-
-    If using persistent config, call [POST /{db}/_config](#operation/post_db-_config) with `{"offline": false}` to set the database to online.
+    If using persistent config this will alter the database config in the bucket to take the database online across all nodes. If you're not using persistent config, this will bring the database online on this node only.
 
     Bringing a database online will:
     * Close the database connection to the backing Couchbase Server bucket.
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -264,18 +264,24 @@ func (h *handler) handleDbOnline() error {
 				currentDatabaseCfg := &DatabaseConfig{DbConfig: currentConfig}
 
 				dbCreds, _ := h.server.Config.DatabaseCredentials[h.db.Name]
-				if err := currentDatabaseCfg.setup(h.ctx(), h.db.Name, h.server.Config.Bootstrap, dbCreds, nil, false); err != nil {
+				if err := currentDatabaseCfg.setup(ctx.Ctx, h.db.Name, h.server.Config.Bootstrap, dbCreds, nil, false); err != nil {
+					// Reset DB state back to Offline on setup failure to avoid being stuck in DBStarting.
+					atomic.StoreUint32(&h.db.State, db.DBOffline)
 					base.WarnfCtx(ctx.Ctx, "config setup failed taking db online, Err: %v", err)
+					return
 				}
 				if err := h.server.ReloadDatabaseWithConfig(contextNoCancel, *currentDatabaseCfg); err != nil {
+					// Reset DB state back to Offline on setup failure to avoid being stuck in DBStarting.
+					atomic.StoreUint32(&h.db.State, db.DBOffline)
 					base.WarnfCtx(ctx.Ctx, "database reload failed during db online, Err: %v", err)
+					return
 				}
 			} else {
 				// persistent config here
 				bucket := h.db.Bucket.GetName()
 				dbName := h.db.Name
 				var updatedDbConfig *DatabaseConfig
-				cas, err := h.server.BootstrapContext.UpdateConfig(h.ctx(), bucket, h.server.Config.Bootstrap.ConfigGroupID, dbName, func(bucketDbConfig *DatabaseConfig) (updatedConfig *DatabaseConfig, err error) {
+				cas, err := h.server.BootstrapContext.UpdateConfig(contextNoCancel.Ctx, bucket, h.server.Config.Bootstrap.ConfigGroupID, dbName, func(bucketDbConfig *DatabaseConfig) (updatedConfig *DatabaseConfig, err error) {
 					if h.headerDoesNotMatchEtag(bucketDbConfig.Version) {
 						return nil, base.HTTPErrorf(http.StatusPreconditionFailed, "Provided If-Match header does not match current config version")
 					}
@@ -284,11 +290,11 @@ func (h *handler) handleDbOnline() error {
 					// set config to offline false to start online processes
 					bucketDbConfig.StartOffline = base.Ptr(false)
 
-					if err := bucketDbConfig.validateConfigUpdate(h.ctx(), oldConfig, false); err != nil {
+					if err := bucketDbConfig.validateConfigUpdate(contextNoCancel.Ctx, oldConfig, false); err != nil {
 						return nil, base.NewHTTPError(http.StatusBadRequest, err.Error())
 					}
 
-					bucketDbConfig.Version, err = GenerateDatabaseConfigVersionID(h.ctx(), bucketDbConfig.Version, &bucketDbConfig.DbConfig)
+					bucketDbConfig.Version, err = GenerateDatabaseConfigVersionID(ctx.Ctx, bucketDbConfig.Version, &bucketDbConfig.DbConfig)
 					if err != nil {
 						return nil, err
 					}
@@ -299,14 +305,18 @@ func (h *handler) handleDbOnline() error {
 				})
 				if err != nil {
 					base.WarnfCtx(ctx.Ctx, "database reload failed during db online, Err: %v", err)
+					// Reset DB state back to Offline on setup failure to avoid being stuck in DBStarting.
+					atomic.StoreUint32(&h.db.State, db.DBOffline)
 					return
 				}
 				updatedDbConfig.cfgCas = cas
 
 				dbCreds, _ := h.server.Config.DatabaseCredentials[dbName]
 				bucketCreds, _ := h.server.Config.BucketCredentials[bucket]
-				if err := updatedDbConfig.setup(h.ctx(), dbName, h.server.Config.Bootstrap, dbCreds, bucketCreds, h.server.Config.IsServerless()); err != nil {
+				if err := updatedDbConfig.setup(contextNoCancel.Ctx, dbName, h.server.Config.Bootstrap, dbCreds, bucketCreds, h.server.Config.IsServerless()); err != nil {
 					base.WarnfCtx(ctx.Ctx, "config setup failed taking db online, Err: %v", err)
+					// Reset DB state back to Offline on setup failure to avoid being stuck in DBStarting.
+					atomic.StoreUint32(&h.db.State, db.DBOffline)
 					return
 				}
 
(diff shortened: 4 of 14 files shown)


db/database.go
torcolvin, line 854: ExitChanges is unused now

torcolvin (commented):
The API docs will need to be updated as noted below and possibly also the public facing document.

rest/admin_api.go
```diff
@@ -260,9 +331,86 @@ func (h *handler) handleDbOnline() error {
...
-	if err := h.db.TakeDbOffline(base.NewNonCancelCtx(), "ADMIN Request"); err != nil {
-		base.InfofCtx(h.ctx(), base.KeyCRUD, "Unable to take Database : %v, offline", base.MD(h.db.Name))
-		return err
+
+	var oldConfig DbConfig
+	contextNoCancel := base.NewNonCancelCtx()
+	// Block until all current calls have returned, including _changes feeds
+	h.db.AccessLock.Lock()
+	defer h.db.AccessLock.Unlock()
+
```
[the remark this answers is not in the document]
gregns1: I don't think this is true? AccessLock read lock is acquired inside validateAndWriteHeaders released after function is finished. Only other places this is used is on Online/Offline operations.

rest/admin_api.go
```diff
@@ -247,11 +247,82 @@ func (h *handler) handleDbOnline() error {
...
+					bucketDbConfig.SGVersion = base.ProductVersion.String()
+					updatedDbConfig = bucketDbConfig
+					return bucketDbConfig, nil
+				})
+				if err != nil {
+					base.WarnfCtx(ctx.Ctx, "database reload failed during db online, Err: %v", err)
+					return
+				}
+				updatedDbConfig.cfgCas = cas
+
```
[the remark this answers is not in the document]
torcolvin: I don't think this is true? @gregns1 do you think you can verify?

docs/api/paths/admin/db-_offline.yaml
```diff
@@ -10,9 +10,7 @@ parameters:
 post:
   summary: Take the database offline
   description: |-
-    This will take the database offline on this node only. Actions can be taken without disrupting current operations ungracefully or having the restart the Sync Gateway instance.
-
-    If using persistent config, call [POST /{db}/_config](#operation/post_db-_config) with `{"offline": true}` to set the database to offline.
+    If using persistent config this will alter the database config in the bucket to take the database offline across all nodes. If you're not using persistent config, this will bring the database online on this node only. Actions can be taken without disrupting current operations ungracefully or having the restart the Sync Gateway instance.
```
torcolvin:
```suggestion
    Transition the state of the database to offline. This database will no longer accept traffic and only maintenance operations can be performed. This operation blocks until all REST traffic is idle.
    
    If using persistent config this will affect all Sync Gateway nodes, otherwise this is operation is local to the node.
```

docs/api/paths/admin/db-_online.yaml
```diff
@@ -10,9 +10,7 @@ parameters:
 post:
   summary: Bring the database online
   description: |-
-    This will bring the database online on this node only so the Public and full Admin REST API requests can be served.
-
-    If using persistent config, call [POST /{db}/_config](#operation/post_db-_config) with `{"offline": false}` to set the database to online.
+    If using persistent config this will alter the database config in the bucket to take the database online across all nodes. If you're not using persistent config, this will bring the database online on this node only.
```
torcolvin:
```suggestion
    Transition the state of the database to online.
    
    If using persistent config this will affect all Sync Gateway nodes, otherwise this is operation is local to the node.
```

rest/admin_api.go
```diff
@@ -247,11 +247,94 @@ func (h *handler) handleDbOnline() error {
...
 	base.InfofCtx(h.ctx(), base.KeyCRUD, "Taking Database : %v, online in %v seconds", base.MD(h.db.Name), input.Delay)
-	go func() {
+	// allow db to come online in background in its own goroutine
+	go func(ctx base.NonCancellableContext) {
 		time.Sleep(time.Duration(input.Delay) * time.Second)
-		h.server.TakeDbOnline(base.NewNonCancelCtx(), h.db.DatabaseContext)
-	}()
+		h.db.AccessLock.Lock()
+		defer h.db.AccessLock.Unlock()
+		if atomic.CompareAndSwapUint32(&h.db.State, db.DBOffline, db.DBStarting) {
```
torcolvin: Can you have this share code with `handlePutDbConfig`? I think you can abstract part of that function from where it updates the config.

You could do something like at the top of the modification function:

```
defer func() {
      if err != nil {
							// Reset DB state back to Offline on setup failure to avoid being stuck in DBStarting.
					atomic.StoreUint32(&h.db.State, db.DBOffline)
	}
}()
```       

I think it's OK to audit log that this is a configuration change, because it is. You can audit log Online after.

rest/changes_api.go
```diff
@@ -443,10 +443,6 @@ func (h *handler) sendSimpleChanges(channels base.Set, options db.ChangesOptions
 				base.InfofCtx(h.ctx(), base.KeyChanges, "Connection lost from client")
 				forceClose = true
 				break loop
-			case <-h.db.ExitChanges:
```
torcolvin, line 446: I think you want to listen for `h.db.CancelCtx` here.

I actually think there is a problem with the `AccessLock` exemplified by this code. If you were running a continuous changes rest API call and then tried to take the database offline, it would never go offline because the continuous changes REST API call would be holding `AccessLock` and the db offline would never be able to acquire it.

I think we can probably avoid acquiring an `AccessLock` for now and just take the db offline without waiting for connections to drain. It might be worth consulting with the team whether we would actually want a function that does a draining operation. Notably a `sync.Mutex` wouldn't do a drain anyway because callers that lock the mutex are not guaranteed to be accessed in a FIFO order, so new REST requests could come in and block the code.

rest/api_test.go
```diff
@@ -3284,6 +3288,25 @@ func TestBufferFlush(t *testing.T) {
...
+		resp := rt.SendAdminRequest(http.MethodGet, "/{{.keyspace}}/_changes?feed=continuous&since=0&timeout=50000&include_docs=true", "")
+		RequireStatus(t, resp, http.StatusOK)
+	}()
+
+	base.RequireWaitForStat(t, func() int64 {
+		return rt.GetDatabase().DbStats.Database().NumReplicationsActive.Value()
+	}, 1)
+
+	// assert that we can take db offline still
+	rt.TakeDbOffline()
```
bbrks, line 3303: We should check we've not leaked a changes feed goroutine here as well? I would expect changes feeds to be closed.

rest/admin_api.go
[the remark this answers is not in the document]
bbrks: Yeah I agree with your analysis - the read lock is released before calling `method(h)` in `h.invoke()` which would be the blocking operation in a cont. changes feed

rest/changes_api.go
bbrks: > because the continuous changes REST API call would be holding AccessLock and the db offline would never be able to acquire it.

As far as I can tell that's not true. See response to Copilot above.

rest/admin_api.go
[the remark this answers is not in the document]
torcolvin: I agree with above, I was tricked by the comment here which implied it was blocking https://github.com/couchbase/sync_gateway/blob/39636bd83a0b3c135bf5b965249f1e71f968031e/rest/handler.go#L500

Given this, does AccessLock actually have any use at all? It would only block access when you are deciding auth/keyspace/etc so I don't think this actually adds any value. We don't do this during standard config updates.

rest/changes_api.go
gregns1: Tests i've added I think prove this is wrong. See `TestContinuousChangesDoesNotBlockIOffline` and `TestBlipDisconnectOnDbOffline`

rest/admin_api.go
```diff
@@ -247,11 +247,83 @@ func (h *handler) handleDbOnline() error {
 
 	_ = base.JSONUnmarshal(body, &input)
```
torcolvin: I think you should add error checking to this, just in case.

rest/admin_api.go
```diff
@@ -247,11 +247,83 @@ func (h *handler) handleDbOnline() error {
...
+				// Reset DB state back to Offline on setup failure to avoid being stuck in DBStarting.
+				atomic.StoreUint32(&h.db.State, db.DBOffline)
+			}
+		}()
 		time.Sleep(time.Duration(input.Delay) * time.Second)
-		h.server.TakeDbOnline(base.NewNonCancelCtx(), h.db.DatabaseContext)
-	}()
+		h.db.AccessLock.Lock()
+		defer h.db.AccessLock.Unlock()
+		if atomic.CompareAndSwapUint32(&h.db.State, db.DBOffline, db.DBStarting) {
```
torcolvin: nit (don't have to do this): 

for readability I'd be inclined to use an early return so this is less indented

rest/admin_api.go
```diff
@@ -247,11 +247,83 @@ func (h *handler) handleDbOnline() error {
...
+
+				dbCreds, _ := h.server.Config.DatabaseCredentials[dbName]
+				bucketCreds, _ := h.server.Config.BucketCredentials[bucket]
+				if err = updatedDbConfig.setup(contextNoCancel.Ctx, dbName, h.server.Config.Bootstrap, dbCreds, bucketCreds, h.server.Config.IsServerless()); err != nil {
+					base.WarnfCtx(ctx.Ctx, "config setup failed taking db online, Err: %v", err)
+					return
+				}
+
+				h.server._databasesLock.Lock()
+				defer h.server._databasesLock.Unlock()
```
torcolvin: I think you actually want to lock this at the top of this function before checking `if atomic.CompareAndSwapUint32(&h.db.State, db.DBOffline, db.DBStarting)`. That way if this function is called while a config update you will not lose the config update if it "wins".

If this is the correct behavior that's OK but I want to think about it and document why this is OK. This is one of the reasons that deduplicating these functions would be good.

rest/admin_api.go
```diff
@@ -247,11 +247,83 @@ func (h *handler) handleDbOnline() error {
 
 	_ = base.JSONUnmarshal(body, &input)
 
+	contextNoCancel := base.NewNonCancelCtx()
```
torcolvin: I think this drops the db from the context.

I think we should keep db on the context so we don't have to worry about logging the database name everywhere. You could also drop all the context like above but then you want to make sure all logging will have database name.

IMO, I'd be fine with keeping all the context information including the correlation ID so we know what triggered this database update, vs bootstrap polling. This might be confusing because the request will be done, but some of the context will live on. I think this would depart from the way other code works so I'm fine just preserving the db name. If you did go with this approach `ctx.WithoutCancel` will drop the cancellations from any context.

rest/admin_api.go
```diff
@@ -247,11 +247,83 @@ func (h *handler) handleDbOnline() error {
...
+						return nil, err
+					}
+
+					bucketDbConfig.SGVersion = base.ProductVersion.String()
+					updatedDbConfig = bucketDbConfig
+					return bucketDbConfig, nil
+				})
+				if err != nil {
+					base.WarnfCtx(ctx.Ctx, "database reload failed during db online, Err: %v", err)
+					return
```
torcolvin: I wonder if this code should be its own function. This already has drift between other places that update config:

- handlePutCollectionConfigSync
- handleDeleteCollectionConfigSync
- handleDeleteCollectionConfigImportFilter
- handlePutCollectionConfigImportFilter

There seems like there are three steps:

Several steps:

1. In `UpdateConfig`:
	i. Use the bucketconfig to update with one new value. This is going to unique to every function
    ii. validate config update and produce a new bucketDbConfig (this should match for all functions)
2. After the config has been updated, call setup and reload the configuration
3. If steps 1 and 2 succeed, then audit with the appropriate logging.

rest/admin_api.go
```diff
@@ -260,10 +332,65 @@ func (h *handler) handleDbOnline() error {
 // Take a DB offline
 func (h *handler) handleDbOffline() error {
 	h.assertAdminOnly()
-	if err := h.db.TakeDbOffline(base.NewNonCancelCtx(), "ADMIN Request"); err != nil {
-		base.InfofCtx(h.ctx(), base.KeyCRUD, "Unable to take Database : %v, offline", base.MD(h.db.Name))
-		return err
+
+	var oldConfig DbConfig
+	contextNoCancel := base.NewNonCancelCtx()
+	// Block until all current calls have returned, including _changes feeds
```
torcolvin, line 330: This code is duplicated from online step, I wonder  if the abstraction is something like pass `func(bucketDbConfig *DatabaseConfig)` and then all the functions above that modify the configuration and expect a bucket reload and call this.

rest/blip_api_crud_test.go
```diff
@@ -3247,6 +3256,65 @@ func TestBlipDatabaseClose(t *testing.T) {
 	})
 }
 
+func TestBlipDisconnectOnDbOffline(t *testing.T) {
```
torcolvin, line 3250: nit:

I don't know I need these tests, I think they were useful as a proof of concept. I'm fine to leave them too.

I'm wary of too many tests that don't provide utility. If I did keep them I'd probably do something like `TestNonBlockingContinuousConnectionsDbOffline`.

I think mostly these prove that I don't think `AccessLock` is doing what it was originally doing. I'd also be inclined to remove `AccessLock` entirely. I think that it existed at a time when `validateAndWriteHeaders` didn't exist yet and `AccessLock` actually blocked when the handlers were running.

rest/server_context_test.go
```diff
@@ -950,27 +954,29 @@ func TestDatabaseStartupFailure(t *testing.T) {
...
+			JWTChannels: base.SetOf("asdf"),
+		},
 	}
 
-	// this call does use the leaky bucket
-	dbContext, err := rt.ServerContext().AddDatabaseFromConfigWithBucket(rt.Context(), t, dbConfig, rt.Bucket())
-	require.ErrorIs(t, err, touchErr)
+	// reload db with invalid config, should fail online process and put db in offline state
+	dbContext, err := rt.ServerContext().ReloadDatabase(ctx, "db", false)
+	require.Error(t, err)
```
torcolvin: I would do something like:

require.ErrorContains(t, err, "must either specify all OIDC properties or none")

to make sure this is the error we expect.

rest/server_context_test.go
```diff
@@ -950,27 +954,29 @@ func TestDatabaseStartupFailure(t *testing.T) {
 		t.Skip("LeakyBucketConfig not supported on CBS")
 	}
```
torcolvin:
```suggestion
```

The way that you rewrote the test, it doesn't need EE or rosmar anymore.

It does fail when rosmar is not used but I can't see why immediately and that makes me think the code isn't doing what what we think.

rest/utilities_testing_resttester.go
```diff
@@ -466,26 +466,15 @@ func (rt *RestTester) PersistDbConfigToBucket(dbConfig DbConfig, bucketName stri
 
 // TakeDbOffline takes the database offline.
```
torcolvin, line 467:
```suggestion
// TakeDbOffline takes the database offline. This request is asynchronous. 
```

I'd be somewhat tempted to make this code synchronous, we actually rely on this being synchronous for each test that relies on resync. https://github.com/couchbase/sync_gateway/blob/39636bd83a0b3c135bf5b965249f1e71f968031e/rest/adminapitest/resync_test.go#L198

Are there any cases where we are trying to test its asynchronous behavior? If so, consider if this is more or fewer than rely on synchronous behavior.

rest/server_context_test.go
```diff
@@ -950,27 +954,29 @@ func TestDatabaseStartupFailure(t *testing.T) {
...
-			},
-		},
+		CustomTestBucket: tb.NoCloseClone(),
 		PersistentConfig: true,
 	})
 	defer rt.Close()
+	ctx := rt.Context()
+
+	dbCfg := rt.NewDbConfig()
+	RequireStatus(t, rt.CreateDatabase("db", dbCfg), http.StatusCreated)
```
torcolvin:
```suggestion
rt := NewRestTesterPersistentConfig(t), 
defer rt.Close()
	ctx := rt.Context()
```

The way you rewrote the test, it doesn't rely on a leaky bucket anymore so you can simplify this.

rest/utilities_testing_resttester.go
gregns1: Do you mean online?

rest/admin_api.go
gregns1: I don't think this needs to move? This lock I'm sure if for the databases maps on server context.

rest/blip_api_crud_test.go
gregns1: I feel like this is good to catch regressions if offline doesn't close blip connections.

rest/admin_api.go
gregns1: Good catch, there is a few areas of code using this so i've filed this https://jira.issues.couchbase.com/browse/CBG-5212 to address this.

rest/admin_api.go
gregns1: I have refactored non persistent config to reduce deuplciated code. I have also moved the code after UpdateConfig into shared function called `updateConfigAndReloadDatabase`. I feel like the update callback itself has some subtle differences between uses in other functions, If we want to consolidate this code I think this should be a different ticket given that seems a bit out of scope for this as it think it needs some thought on how to best to do this.

rest/admin_api.go
```diff
@@ -245,13 +245,77 @@ func (h *handler) handleDbOnline() error {
...
+		h.db.AccessLock.Lock()
+		defer h.db.AccessLock.Unlock()
+		if !atomic.CompareAndSwapUint32(&h.db.State, db.DBOffline, db.DBStarting) {
+			base.InfofCtx(contextNoCancel.Ctx, base.KeyCRUD, "Unable to take Database : %v online , database must be in Offline state", base.UD(h.db.Name))
+			return
+		}
+		var oldConfig DbConfig
+		if !h.server.persistentConfig {
+			newDbConfig := &DbConfig{StartOffline: base.Ptr(false)}
+			err := h.updateNonPersistentDbConfig(contextNoCancel, h.db.Name, false, false, true, newDbConfig)
```
bbrks: This function and the deferred error handling relies on `err` declared on line 260.

By doing a `:=` assignment here you're shadowing `err` resulting in the `err` in the defer remaining nil

```suggestion
			err = h.updateNonPersistentDbConfig(contextNoCancel, h.db.Name, false, false, true, newDbConfig)
```

rest/admin_api.go
```diff
@@ -260,10 +324,50 @@ func (h *handler) handleDbOnline() error {
 // Take a DB offline
 func (h *handler) handleDbOffline() error {
```
bbrks, line 325: Nowhere is checking `db.State` anymore before taking the db offline - is that intentional? E.g What if this database is currently resyncing?

rest/admin_api.go
```diff
@@ -260,10 +324,50 @@ func (h *handler) handleDbOnline() error {
...
+			}
+
+			bucketDbConfig.SGVersion = base.ProductVersion.String()
+			updatedDbConfig = bucketDbConfig
+			return bucketDbConfig, nil
+		})
+		if err != nil {
+			return err
+		}
+		if err := h.updateConfigAndReloadDatabase(h.ctx(), dbName, bucket, cas, updatedDbConfig); err != nil {
```
bbrks: use of `h.ctx()` throughout seems wrong, think we need to create a non-cancel ctx to ensure all of these operations are run OK even if the initial request has timed out or been cancelled to ensure the database is in a good state (i.e. not half-updated but not reloaded)

rest/admin_api.go
```diff
@@ -641,6 +745,58 @@ func (h *handler) handlePutConfig() error {
 	return base.HTTPErrorf(http.StatusOK, "Updated")
 }
 
+// updateConfigAndReloadDatabase applies the post-UpdateConfig steps shared by several handlers: it stamps
+// the CAS on the config, runs setup (credential injection etc.), acquires _databasesLock and reloads
+// the database. ctx must be non-cancellable because it is used inside a goroutine or across long-lived
+// operations.
+func (h *handler) updateConfigAndReloadDatabase(ctx context.Context, dbName, bucket string, cas uint64, updatedDbConfig *DatabaseConfig) error {
```
bbrks: this would enforce the godoc comment above about non-cancellable context and the PR comment above

```suggestion
func (h *handler) updateConfigAndReloadDatabase(ctx base.NonCancellableContext, dbName, bucket string, cas uint64, updatedDbConfig *DatabaseConfig) error {
```

rest/admin_api.go
```diff
@@ -641,6 +745,58 @@ func (h *handler) handlePutConfig() error {
...
+	if validateConfigUpdate {
+		err := updatedDbConfig.validateConfigUpdate(h.ctx(), oldDBConfig,
+			validateOIDC)
+		if err != nil {
+			return base.NewHTTPError(http.StatusBadRequest, err.Error())
+		}
+	}
+
+	dbCreds, _ := h.server.Config.DatabaseCredentials[dbName]
+	if err := updatedDbConfig.setup(h.ctx(), dbName, h.server.Config.Bootstrap, dbCreds, nil, false); err != nil {
```
bbrks: Mix of `h.ctx()` and `ctx` here - I think it needs to be the non-cancellable one?

rest/admin_api.go
```diff
@@ -245,13 +245,77 @@ func (h *handler) handleDbOnline() error {
...
+				atomic.StoreUint32(&h.db.State, db.DBOffline)
+			}
+		}()
 		time.Sleep(time.Duration(input.Delay) * time.Second)
-		h.server.TakeDbOnline(base.NewNonCancelCtx(), h.db.DatabaseContext)
-	}()
+		h.db.AccessLock.Lock()
+		defer h.db.AccessLock.Unlock()
+		if !atomic.CompareAndSwapUint32(&h.db.State, db.DBOffline, db.DBStarting) {
+			base.InfofCtx(contextNoCancel.Ctx, base.KeyCRUD, "Unable to take Database : %v online , database must be in Offline state", base.UD(h.db.Name))
```
bbrks: database names aren't user data - MD or drop the redaction 

```suggestion
			base.InfofCtx(contextNoCancel.Ctx, base.KeyCRUD, "Unable to take Database : %v online , database must be in Offline state", base.MD(h.db.Name))
```

docs/api/paths/admin/db-_online.yaml
```diff
@@ -10,9 +10,9 @@ parameters:
 post:
   summary: Bring the database online
   description: |-
-    This will bring the database online on this node only so the Public and full Admin REST API requests can be served.
+    Transition the state of the database to online.
 
-    If using persistent config, call [POST /{db}/_config](#operation/post_db-_config) with `{"offline": false}` to set the database to online.
+    If using persistent config this will affect all Sync Gateway nodes, otherwise this is operation is local to the node.
```
bbrks:
```suggestion
    If using persistent config this will affect all Sync Gateway nodes, otherwise this operation is local to the node.
```

rest/api_test.go
```diff
@@ -3284,6 +3284,30 @@ func TestBufferFlush(t *testing.T) {
 	assert.True(t, resp.Flushed)
 }
 
+func TestContinuousChangesDoesNotBlockIOffline(t *testing.T) {
```
bbrks:
```suggestion
func TestContinuousChangesDoesNotBlockOffline(t *testing.T) {
```

rest/admin_api.go
```diff
@@ -245,13 +245,77 @@ func (h *handler) handleDbOnline() error {
...
+			return
+		}
+		var oldConfig DbConfig
```
bbrks: `oldConfig` not used at this scope - move down into the else block

bbrks (changes requested):
A few things - mostly around context usage

bbrks (approved):

PR #8094, opened 2026-03-05
CBG-5205 test speed up: avoid sleep on db start

torcolvin:
CBG-5205 test speed up: avoid sleep on db start

When starting a database that has a non zero _sync:seq, the database will sleep 1.5 to wait for released sequences.
In a test environment, this occurs often when there is a single RestTester running an a database configuration
occurs.

Fixes a bug where if users or roles are specified on database configuration, they would allocate a sequence
before the initial sequence was computed, so this would always result in waiting.

Set DisableSequenceWaitOnDbRestart anywhere there was a single RestTester. Keep this value for ISGR tests that
actually use multiple RestTesters.

Remove LongRunningTest for any test that runs faster than 50ms. This affects 37 tests.

I think it might be worthwhile to drop the comments around the usage of DisableSequenceWaitOnDbRestart?

## Pre-review checklist

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/436/

### DIFF
--- a/rest/adminapitest/admin_api_test.go
+++ b/rest/adminapitest/admin_api_test.go
@@ -547,8 +547,6 @@ func TestDBOfflinePutDbConfig(t *testing.T) {
 // Tests that the users returned in the config endpoint have the correct names
 // Reproduces #2223
 func TestDBGetConfigNamesAndDefaultLogging(t *testing.T) {
-	base.LongRunningTest(t)
-
 	p := "password"
 	rt := rest.NewRestTester(t,
 		&rest.RestTesterConfig{
@@ -2287,8 +2285,6 @@ func TestHandleGetStackTrace(t *testing.T) {
 }
 
 func TestConfigRedaction(t *testing.T) {
-	base.LongRunningTest(t)
-
 	rt := rest.NewRestTester(t, &rest.RestTesterConfig{DatabaseConfig: &rest.DatabaseConfig{DbConfig: rest.DbConfig{Users: map[string]*auth.PrincipalConfig{"alice": {Password: base.Ptr("password")}}}}})
 	defer rt.Close()
 
--- a/db/change_cache_test.go
+++ b/db/change_cache_test.go
@@ -2430,8 +2430,6 @@ func TestReleasedSequenceRangeHandlingEverythingPendingAndProcessPending(t *test
 //   - Then add another pending entry to take over pending capacity and assert that the range is processed correctly
 //   - Then add another range to completely unblock pending
 func TestReleasedSequenceRangeHandlingEverythingPendingLowPendingCapacity(t *testing.T) {
-	base.LongRunningTest(t)
-
 	base.SetUpTestLogging(t, base.LevelDebug, base.KeyCache)
 
 	ctx := base.TestCtx(t)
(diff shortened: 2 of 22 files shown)


rest/adminapitest/admin_api_test.go
```diff
@@ -2932,10 +2928,12 @@ func TestPutDbConfigChangeName(t *testing.T) {
 }
 
 func TestSwitchDbConfigCollectionName(t *testing.T) {
-	base.LongRunningTest(t)
-
 	base.TestRequiresCollections(t)
 	base.RequireNumTestDataStores(t, 2)
+	// speed up test by not sleeping for _sync:seq when database reloads
+	// this sleep is used for multiple Sync Gateway nodes starting up simultaneously, but this test is only
+	// using a single node
```
bbrks: These comments throughout are definitely redundant as you mentioned.

This function name does a reasonable job at communicating the intention already, and the godoc comment has a good explanation of why. Maybe you can just improve the godoc comment if you feel like it needs it?

rest/adminapitest/admin_api_test.go
```diff
@@ -3296,6 +3294,10 @@ func TestDbConfigPersistentSGVersions(t *testing.T) {
 
 func TestDeleteFunctionsWhileDbOffline(t *testing.T) {
 	base.LongRunningTest(t)
+	// speed up test by not sleeping for _sync:seq when database reloads
+	// this sleep is used for multiple Sync Gateway nodes starting up simultaneously, but this test is only using a
+	// single node
+	db.DisableSequenceWaitOnDbRestart(t)
```
bbrks: Since single node is the "normal" perhaps making this the default behaviour for RestTester would be beneficial rather than introducing this into a large number of tests?

There is already an under-utilised framework in place for "multi-node" RestTesters (`RestTesterCluster`) which may be worth thinking about rolling out more widely if there are tests doing a weird DIY multi-node thing?

db/database.go
```diff
@@ -2245,6 +2245,15 @@ func (db *DatabaseContext) StartOnlineProcesses(ctx context.Context) (returnedEr
...
 
+	// Get current value of _sync:seq
+	initialSequence, seqErr := db.sequences.lastSequence(ctx)
+	if seqErr != nil {
+		return seqErr
+	}
+	initialSequenceTime := time.Now()
+
+	base.InfofCtx(ctx, base.KeyCRUD, "Database has _sync:seq value on startup of %d", initialSequence)
+
```
bbrks: I am finding it difficult to reason about what reordering this code means in practice. Sequences reserved but not allocated during user creation below will no longer block the change cache from starting up even on a single node?

Is that intentional? I think you'd hit this situation if you had a bunch of users to create that forced sequence batching but didn't use a full batch.

bbrks (changes requested):

rest/adminapitest/admin_api_test.go
torcolvin: I think we could bypass this wait time without losing anything important in our tests. The places that set up multiple nodes all call `addActiveRT`. My understanding of the wait is that it helps if multiple nodes are starting up at exactly the same time.

If we wanted to test that behavior, we should probably do it in a way that is more robust and a faster polling window than incidental 1.5 seconds.

db/database.go
torcolvin: My read on this was that installing the user creation was entirely accidental to be this early and just a result of how `StartOnlineProcess` was created. I think I could move the code to `InstallPrincipals` to after the change cache starts up. Moving it in the middle caused this data race https://github.com/couchbase/sync_gateway/pull/8094/changes/7e9da5acb9c1ff7f4462f9491250294b785cdbed which is how I ended up the sequence allocation to the middle.

I agree this is a weird way to do it, and I'm not fond of it. If we have `DisableSequenceWaitOnDbRestart` running all the times, I don't think we have to close this gap.

One thing that makes the principal creation odd is that they are only modified if they don't already exist (except for guest user) as per https://github.com/couchbase/sync_gateway/blob/39636bd83a0b3c135bf5b965249f1e71f968031e/db/users.go#L92 so they rarely would ever actually be created on multiple nodes simultaneously, since this only occurs once on the initial creation of the database.

bbrks (approved):

PR #8113, opened 2026-03-17
[3.3.4] CBG-5200: Add guardrail to prevent disabling of xattrs

RIT3shSapata:
CBG-5200

Describe your PR here...
- Add guardrails to prevent disabling of xattrs
- Added unit tests

## Pre-review checklist

- [ ] Link upstream PRs

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/390

### DIFF
--- a/rest/adminapitest/admin_api_test.go
+++ b/rest/adminapitest/admin_api_test.go
@@ -3998,3 +3998,23 @@ func RequireEventCount(t *testing.T, runtimeConfig *rest.RuntimeDatabaseConfig, 
 	}
 	require.Equal(t, expectedCount, actualCount)
 }
+
+// TestDisableXattrs verifies that disabling xattrs (extended attributes) on a database
+// is handled correctly through the admin API. It ensures that the configuration change
+// is applied and that the database behaves as expected when xattrs are disabled.
+func TestDisableXattrs(t *testing.T) {
+	for _, persistent := range []bool{true, false} {
+		rt := rest.NewRestTester(t, &rest.RestTesterConfig{
+			PersistentConfig: persistent,
+		})
+		defer rt.Close()
+
+		dbConfig := rt.NewDbConfig()
+		dbConfig.EnableXattrs = base.Ptr(true)
+
+		rest.RequireStatus(t, rt.CreateDatabase("db1", dbConfig), http.StatusCreated)
+
+		dbConfig.EnableXattrs = base.Ptr(false)
+		rest.RequireStatus(t, rt.UpsertDbConfig("db1", dbConfig), http.StatusBadRequest)
+	}
+}
--- a/rest/adminapitest/admin_api_test.go
+++ b/rest/adminapitest/admin_api_test.go
@@ -4004,17 +4004,19 @@ func RequireEventCount(t *testing.T, runtimeConfig *rest.RuntimeDatabaseConfig, 
 // is applied and that the database behaves as expected when xattrs are disabled.
 func TestDisableXattrs(t *testing.T) {
 	for _, persistent := range []bool{true, false} {
-		rt := rest.NewRestTester(t, &rest.RestTesterConfig{
-			PersistentConfig: persistent,
+		t.Run(fmt.Sprintf("persistentConfig=%t", persistent), func(t *testing.T) {
+			rt := rest.NewRestTester(t, &rest.RestTesterConfig{
+				PersistentConfig: persistent,
+			})
+			defer rt.Close()
+
+			dbConfig := rt.NewDbConfig()
+			dbConfig.EnableXattrs = base.Ptr(true)
+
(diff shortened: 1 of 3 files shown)


rest/config.go
```diff
@@ -725,6 +725,13 @@ func (dbConfig *DbConfig) validateConfigUpdate(ctx context.Context, old DbConfig
 // validateChanges compares the current DbConfig with the "old" config, and returns an error if any disallowed changes
 // are attempted.
 func (dbConfig *DbConfig) validateChanges(ctx context.Context, old DbConfig) error {
+	// add guardrails to prevent disabling enabled_shared_bucket_access.
```
gregns1:
```suggestion
	// add guardrails to prevent disabling enable_shared_bucket_access.
```

rest/config.go
```diff
@@ -725,6 +725,13 @@ func (dbConfig *DbConfig) validateConfigUpdate(ctx context.Context, old DbConfig
...
+	// nil is treated as false (not enabled), so transitioning from explicitly-enabled to nil is also disallowed.
+	oldXattrsEnabled := old.EnableXattrs != nil && *old.EnableXattrs
+	newXattrsEnabled := dbConfig.EnableXattrs != nil && *dbConfig.EnableXattrs
+	if oldXattrsEnabled && !newXattrsEnabled {
+		return fmt.Errorf("cannot disable enabled_shared_bucket_access after enabling it")
```
gregns1:
```suggestion
		return fmt.Errorf("cannot disable enable_shared_bucket_access after enabling it")
```

gregns1 (dismissed):
Generally looks good to me. Just one thing I think it would be good to have a higher level test to catch regressions in case any of the validation changes inside `handlePutConfig` in future 3.3.x releases. 
Have higher level test just try disable `enable_shared_bucket_access` via db config update endpoint. I think it would be good to have this test parameterised to run in persistent config, non persistent config and using PUT/POST.

rest/adminapitest/admin_api_test.go
```diff
@@ -3998,3 +3998,23 @@ func RequireEventCount(t *testing.T, runtimeConfig *rest.RuntimeDatabaseConfig,
...
+
+// TestDisableXattrs verifies that disabling xattrs (extended attributes) on a database
+// is handled correctly through the admin API. It ensures that the configuration change
+// is applied and that the database behaves as expected when xattrs are disabled.
+func TestDisableXattrs(t *testing.T) {
+	for _, persistent := range []bool{true, false} {
+		rt := rest.NewRestTester(t, &rest.RestTesterConfig{
+			PersistentConfig: persistent,
+		})
+		defer rt.Close()
```
bbrks: `defer` only works in a function scope, not loops, etc.

This should probably be a subtest function anyway and would fix this.

rest/config.go
```diff
@@ -725,6 +725,13 @@ func (dbConfig *DbConfig) validateConfigUpdate(ctx context.Context, old DbConfig
 // validateChanges compares the current DbConfig with the "old" config, and returns an error if any disallowed changes
 // are attempted.
 func (dbConfig *DbConfig) validateChanges(ctx context.Context, old DbConfig) error {
+	// add guardrails to prevent disabling enable_shared_bucket_access.
+	// nil is treated as false (not enabled), so transitioning from explicitly-enabled to nil is also disallowed.
+	oldXattrsEnabled := old.EnableXattrs != nil && *old.EnableXattrs
+	newXattrsEnabled := dbConfig.EnableXattrs != nil && *dbConfig.EnableXattrs
+	if oldXattrsEnabled && !newXattrsEnabled {
+		return fmt.Errorf("cannot disable enable_shared_bucket_access after enabling it")
```
bbrks: This wording is awkward because of the "disable enable" thing.

```suggestion
		return fmt.Errorf("enable_shared_bucket_access=true is a one-way operation, cannot disable it after previously enabling it")
```

rest/config.go
```diff
@@ -725,6 +725,13 @@ func (dbConfig *DbConfig) validateConfigUpdate(ctx context.Context, old DbConfig
 // validateChanges compares the current DbConfig with the "old" config, and returns an error if any disallowed changes
 // are attempted.
 func (dbConfig *DbConfig) validateChanges(ctx context.Context, old DbConfig) error {
+	// add guardrails to prevent disabling enable_shared_bucket_access.
+	// nil is treated as false (not enabled), so transitioning from explicitly-enabled to nil is also disallowed.
```
bbrks: > nil is treated as false (not enabled)

Only in versions prior to 3.0 - see #5123 when we changed the default

Given this, we should _allow_ any transition to `true`, or `nil`->`nil` - everything else is forbidden.

bbrks (changes requested):

bbrks (approved):

PR #8120, opened 2026-03-18
CBG-4363: Xattr-only caching feed

bbrks:
CBG-4363 / CBG-4640

Disables document bodies on the caching DCP feed with a KV Get fallback for subset of documents that require it (metadata docs and IsSGWrite fallback)

## Dependencies
- [x] https://github.com/couchbase/sg-bucket/pull/140
- [x] https://github.com/couchbaselabs/rosmar/pull/59

- [ ] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/446/


db/change_listener.go
```diff
@@ -94,10 +94,11 @@ func (listener *changeListener) Start(ctx context.Context, bucket base.Bucket, d
...
-		ID:         base.DCPCachingFeedID,
-		Backfill:   sgbucket.FeedNoBackfill,
-		Terminator: listener.terminator,
-		DoneChan:   make(chan struct{}),
+		ID:          base.DCPCachingFeedID,
+		Backfill:    sgbucket.FeedNoBackfill,
+		Terminator:  listener.terminator,
+		DoneChan:    make(chan struct{}),
+		FeedContent: sgbucket.FeedContentXattrOnly,
 	}
```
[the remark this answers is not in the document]
bbrks: 4.0+ can only run in xattr mode - not an issue

db/change_cache.go
```diff
@@ -387,35 +392,38 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent, docType DocumentType)
 		return
 	}
 
-	// If using xattrs and this isn't an SG write, we shouldn't attempt to cache.
+	if syncData == nil {
+		return
+	}
```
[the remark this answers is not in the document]
bbrks: 4.0+ can only run in xattr mode - not an issue

base/dcp_client_test.go
```diff
@@ -752,6 +752,189 @@ func TestDCPFeedEventTypes(t *testing.T) {
...
+					var eventMu sync.Mutex
+					allFound := make(chan struct{})
+
+					checkAllFound := func() {
+						if gotBodyOnly.Load() && gotXattr.Load() && gotCounter.Load() {
+							select {
+							case allFound <- struct{}{}:
+							default:
+							}
+						}
```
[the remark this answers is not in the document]
bbrks: will switch to use `base.RequireChan` helpers

db/change_cache.go
```diff
@@ -318,20 +318,25 @@ func (c *changeCache) CleanSkippedSequenceQueue(ctx context.Context) error {
...
 	timeReceived := channels.NewFeedTimestamp(&event.TimeReceived)
+
 	// ** This method does not directly access any state of c, so it doesn't lock.
 	// Is this a user/role doc for this database?
 	switch docType {
 	case DocTypeUnknown:
 		return // no-op unknown doc type
 	case DocTypeUser:
-		c.processPrincipalDoc(ctx, docID, docJSON, true, timeReceived)
+		if docBody := c.fetchMetadataDocBody(ctx, docID, event.Cas); docBody != nil {
```
torcolvin: Is it ever a problem to not have a document body here? Previously, we would pass docJSON as nil before.

Does this amount of an optimization? I don't think we can do anything if it is a nil body but I'm just curious for the motivation.

I suspect it would only be a nil body if it got corrupted because we should ignore deletion events without xattrs

db/document.go
```diff
@@ -700,6 +700,47 @@ func (s *SyncData) IsSGWrite(ctx context.Context, cas uint64, rawBody []byte, ra
...
+		return false, false
+	}
+
+	// 4. CV mismatch — SDK write
+	if s.RevAndVersion.CurrentVersion != "" || s.RevAndVersion.CurrentSource != "" {
+		extractedCV, err := cv.ExtractCV()
+		if !errors.Is(err, base.ErrNotFound) {
+			if err != nil {
+				base.InfofCtx(ctx, base.KeyImport, "Unable to extract cv during IsSGWriteXattrOnly check - skipping cv match check: %v", err)
+				return true, false
```
```suggestion
				return false, false
```

This might be worth logging what the cv is, this document is basically corrupt as is. We could return an error from this function and pass it to the caller to log as a info/warning that this document will _not_ be processed by the changes feed.

bbrks: Good catch, there's equivalent code that also looks incorrect in the original `SyncData.IsSGWrite()`

```go
		if !errors.Is(err, base.ErrNotFound) {
			if err != nil {
				base.InfofCtx(ctx, base.KeyImport, "Unable to extract cv during IsSGWrite write check - skipping cv match check: %v", err)
				return true, true, false
			}
```

db/change_cache.go
bbrks: Tweaked this a little bit to explicitly handle doc not found and cas errors silently.
Both of those cases are OK to ignore error for - either deletes or a CAS mismatch means there's an in-flight mutation that will arrive again over the feed.
The other times we'll log the error and skip.

torcolvin:
Consider picking up the latest version of gocb- https://github.com/couchbase/sync_gateway/pull/8148

base/stats.go
```diff
@@ -1474,6 +1476,10 @@ func (d *DbStats) initCacheStats() error {
 	if err != nil {
 		return err
 	}
+	resUtil.IsSGWriteKVFetchCount, err = NewIntStat(SubsystemCacheKey, "issgwrite_kv_fetch_count", StatUnitNoUnits, IsSGWriteKVFetchCountDesc, StatAddedVersion4dot0dot0, StatDeprecatedVersionNotDeprecated, StatStabilityCommitted, labelKeys, labelVals, prometheus.CounterValue, 0)
```
torcolvin: Is this planning to go into 4.0.5 or 3.3.5 or 4.1.0?

db/change_cache.go
```diff
@@ -387,35 +392,37 @@ func (c *changeCache) DocChanged(event sgbucket.FeedEvent, docType DocumentType)
...
+	if vv := doc.Xattrs[base.VvXattrName]; len(vv) > 0 {
+		rawVV = base.Ptr(rawHLV(vv))
+	}
+	isDelete := event.Opcode == sgbucket.FeedOpDeletion
+
+	if isSGWrite, isSGWriteAmbiguous := syncData.IsSGWriteXattrOnly(ctx, event.Cas, isDelete, rawUserXattr, rawVV); isSGWriteAmbiguous {
+		// CRC is the only remaining check, but we need to fetch the doc body now
+		c.db.DbStats.Cache().IsSGWriteKVFetchCount.Add(1)
+		docBody, cas, err := collection.GetCollectionDatastore().GetRaw(docID)
+		if err != nil {
```
torcolvin, line 410: Maybe too detailed thoughts:

Do we want to just ignore DocNotFound here entirely if the tombstone got purged? I think this is unlikely but if so the debug logging would not be helpful, maybe.

Also, does this work for tombstones?

base/stats.go
bbrks: My bad - 4.1 only I think. There's not enough benefit to backporting.

db/change_cache.go
bbrks: Case 2 in `IsSGWriteXattrOnly` handles server-side generated tombstones - the intention is to drop them and wait for the import and process that mutation. We won't get into this conditional block, and will just return from `!isSGWrite` below.

```go
// 2. Deletion with non-SG CRC - SG deletions always store DeleteCrc32c
	if isDelete && s.Crc32c != base.DeleteCrc32c {
		return false, false
	}
```

torcolvin (approved):

torcolvin (dismissed):

torcolvin (approved):

PR #8133, opened 2026-03-24
CBG-5233: have single key type on revision cache

gregns1:
CBG-5233

- remove revID lookup and hlv lookup and have single lookup by string to list element
- This means two items can correspond to the same doc if a fetch is done by revID then CV
- For Put and Upsert, these only operate on CV as it stands. I think this makes sense instead of a Put or Upsert potentially adding two new items for one doc. In any case this is off by default.
- Some refactoring to make code shared
- Kept lookup map keyed by struct type, to avoid string formatting

## Pre-review checklist

- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/410/ (disable rev cache)
- [x] `GSI=true,xattrs=true` https://jenkins.sgwdev.com/job/SyncGatewayIntegration/409/ (flaky test failure)

### DIFF
--- a/db/revision_cache_bypass.go
+++ b/db/revision_cache_bypass.go
@@ -118,7 +118,7 @@ func (rc *BypassRevisionCache) GetActive(ctx context.Context, docID string, coll
 }
 
 // Peek is a no-op for a BypassRevisionCache, and always returns a false 'found' value.
-func (rc *BypassRevisionCache) Peek(ctx context.Context, docID, revID string, collectionID uint32) (docRev DocumentRevision, found bool) {
+func (rc *BypassRevisionCache) Peek(ctx context.Context, docID string, key RevCacheKey) (docRev DocumentRevision, found bool) {
 	return DocumentRevision{}, false
 }
 
--- a/db/revision_cache_bypass.go
+++ b/db/revision_cache_bypass.go
@@ -32,48 +32,28 @@ func NewBypassRevisionCache(backingStores map[uint32]RevisionCacheBackingStore, 
 }
 
 // GetUsingRevID fetches the revision for the given docID and revID immediately from the bucket.
-func (rc *BypassRevisionCache) GetUsingRevID(ctx context.Context, docID, revID string, collectionID uint32, includeDelta bool) (docRev DocumentRevision, err error) {
+func (rc *BypassRevisionCache) Get(ctx context.Context, docID, versionString string, collectionID uint32, includeDelta, loadBackup bool) (docRev DocumentRevision, err error) {
 	doc, err := rc.backingStores[collectionID].GetDocument(ctx, docID, DocUnmarshalSync)
 	if err != nil {
 		return DocumentRevision{}, err
 	}
 
 	docRev = DocumentRevision{
-		RevID:                  revID,
 		DocID:                  docID,
 		RevCacheValueDeltaLock: &sync.Mutex{}, // initialize the mutex for delta updates
 	}
-	var hlv *HybridLogicalVector
-	docRev.BodyBytes, docRev.History, docRev.Channels, docRev.Removed, docRev.Attachments, docRev.Deleted, docRev.Expiry, hlv, err = revCacheLoaderForDocument(ctx, rc.backingStores[collectionID], doc, revID)
-	if err != nil {
-		return DocumentRevision{}, err
-	}
-	if hlv != nil {
-		docRev.CV = hlv.ExtractCurrentVersionFromHLV()
-		docRev.HlvHistory = hlv.ToHistoryForHLV()
-	}
-
-	rc.bypassStat.Add(1)
-
-	return docRev, nil
-}
-
-// GetUsingCV fetches the Current Version for the given docID and CV immediately from the bucket.
-func (rc *BypassRevisionCache) GetUsingCV(ctx context.Context, docID string, cv *Version, collectionID uint32, includeDelta bool, loadBackup bool) (docRev DocumentRevision, err error) {
-
-	docRev = DocumentRevision{
-		CV:                     cv,
-		DocID:                  docID,
-		RevCacheValueDeltaLock: &sync.Mutex{}, // initialize the mutex for delta updates
-	}
-
-	doc, err := rc.backingStores[collectionID].GetDocument(ctx, docID, DocUnmarshalSync)
-	if err != nil {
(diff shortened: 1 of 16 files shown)


db/revision_cache_interface.go
```diff
@@ -524,3 +509,11 @@ func (c *DatabaseCollection) getCurrentVersion(ctx context.Context, doc *Documen
...
 	return bodyBytes, attachments, channels, deleted, err
 }
+
+func CreateRevisionCacheRevIDKey(docID, revID string, collectionID uint32) string {
+	return fmt.Sprintf("%s:%s:%d", docID, revID, collectionID)
+}
+
+func CreateRevisionCacheCVKey(docID string, cv *Version, collectionID uint32) string {
+	return fmt.Sprintf("%s:%s:%d", docID, cv.String(), collectionID)
+}
```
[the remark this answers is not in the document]
gregns1: Switched to use struct type to avoid formatting strings

db/revision_cache_interface.go
```diff
@@ -524,3 +483,13 @@ func (c *DatabaseCollection) getCurrentVersion(ctx context.Context, doc *Documen
 	}
 	return bodyBytes, attachments, channels, deleted, err
 }
+
+type RevCacheKey struct {
```
bbrks: Nit: Can remain unexported, since nobody outside of `db` needs this

```suggestion
type revCacheKey struct {
```

db/revision_cache_bypass.go
```diff
@@ -31,49 +31,29 @@ func NewBypassRevisionCache(backingStores map[uint32]RevisionCacheBackingStore,
 	}
 }
 
-// GetWithRev fetches the revision for the given docID and revID immediately from the bucket.
-func (rc *BypassRevisionCache) GetWithRev(ctx context.Context, docID, revID string, collectionID uint32, includeDelta bool) (docRev DocumentRevision, err error) {
+// GetUsingRevID fetches the revision for the given docID and revID immediately from the bucket.
```
bbrks:
```suggestion
// Get fetches the revision for the given docID and revID immediately from the bucket.
```

db/revision_cache_interface.go
```diff
@@ -29,44 +29,27 @@ const (
 // RevisionCache is an interface that can be used to fetch a DocumentRevision for a Doc ID and Rev ID pair.
 type RevisionCache interface {
 
-	// GetWithRev returns the given revision, and stores if not already cached.
+	// GetUsingRevID returns the given revision, and stores if not already cached.
```
bbrks:
```suggestion
	// Get returns the given revision, and stores if not already cached.
```

db/revision_cache_lru.go
```diff
@@ -391,229 +311,87 @@ func (rc *LRURevisionCache) upsertDocToCache(ctx context.Context, cvKey IDandCV,
...
-	rc.removeFromCVLookup(ctx, docID, cv, collectionID)
-}
-
-// RemoveRevOnly removes a rev from revision cache lookup map, if present.
-func (rc *LRURevisionCache) RemoveRevOnly(ctx context.Context, docID, revID string, collectionID uint32) {
-	// This will only remove the entry from the rev lookup map, not the lru list
-	rc.removeFromRevLookup(ctx, docID, revID, collectionID)
+// Remove removes a rev from revision cache lookup map, if present.
+func (rc *LRURevisionCache) Remove(ctx context.Context, docID, versionString string, collectionID uint32) {
+	rc.removeFromRevCache(ctx, CreateRevisionCacheKey(docID, versionString, collectionID))
```
bbrks: Effectively an empty wrapper function now - could just pull code in `LRURevisionCache.removeFromRevCache()` directly up into here?

db/revision_cache_lru.go
```diff
@@ -391,229 +311,87 @@ func (rc *LRURevisionCache) upsertDocToCache(ctx context.Context, cvKey IDandCV,
...
-		value.canEvict.Store(false)
-		newElem := rc.lruList.PushFront(value)
-		rc.hlvCache[key] = newElem
+		value = &revCacheValue{id: docID, collectionID: collectionID, itemKey: key}
+		if base.IsRevTreeID(docVersionString) {
+			value.revID = docVersionString
+		} else {
+			cv, err := ParseVersion(docVersionString)
+			if err != nil {
+				return nil
```
bbrks: Not a huge fan of dropping this error - I think most callers could take a returned error OK?

I think there is that test-helper method that is not set up to tolerate a `nil` value being returned which would end up panicing the tests if it ran into it. Maybe add an assertion there or explicit error check?

db/revision_cache_lru.go
```diff
@@ -391,229 +311,87 @@ func (rc *LRURevisionCache) upsertDocToCache(ctx context.Context, cvKey IDandCV,
...
 	rc.lock.Lock()
 	defer rc.lock.Unlock()
-	element, ok := rc.hlvCache[key]
+	elem, ok := rc.cache[key]
 	if !ok {
 		return
 	}
-	// grab the revid key from the value to enable us to remove the reference from the rev lookup map too
-	elem, ok := element.Value.(*revCacheValue)
+	revValue, ok := elem.Value.(*revCacheValue)
```
bbrks, line 364: Reminder to self: I need to take another look at this and compare with `removeValue` to figure out why they are different (if they are at all)_

bbrks (commented):
I think this looks OK - a few nits and one more thing I need to follow up on to double check tomorrow

db/revision_cache_lru.go
gregns1: So I noticed a bit lots of tests fail when intentionally sending a bad rev in the request. I have changed this to have the same processing as `getRev` used to for parsing version string. If we can't parse CV we assume its revID and if its some garbage string passed in let the 404 bubble up.

db/revision_cache_lru.go
gregns1: This function decrements the memory usage and `removeValue` doesn't. This is because we don't actually increment memory usage for failed loads so no need to do it here.
I have moved the `removeFromRevCache` into `Remove` no need to keep separate now we don't have two removal methods. I have also renamed this to be explicit of its use.

nntld (approved):

bbrks (approved):

PR #8141, opened 2026-03-30
CBG-5190: Add node poller for shardedDCPFeedListener

RIT3shSapata:
CBG-5190

Describe your PR here...
- Added a node poller that polls for all the CBGT subscribed documents
- This node poller is used only for distributed resync
- Whenever there is a change in any of the documents used by CBGT, an event is fired to reload the nodes
- added test coverage for the node poller

## Pre-review checklist

- [ ] Link upstream PRs

- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/445/
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/477/

### DIFF
--- a/base/dcp_sharded.go
+++ b/base/dcp_sharded.go
@@ -566,7 +566,7 @@ func registerHeartbeatListener(ctx context.Context, heartbeater Heartbeater, cbg
 		return nil, errors.New("Unable to register heartbeat listener with nil manager, cfg or heartbeater")
 	}
 
-	// Register listener for import, uses cfg and manager to manage set of participating nodes
+	// Register listener for shardedDCP, uses cfg and manager to manage set of participating nodes
 	shardedDCPHeartbeatListener, err := NewShardedDCPHeartbeatListener(ctx, cbgtContext)
 	if err != nil {
 		return nil, err
@@ -582,7 +582,7 @@ func registerHeartbeatListener(ctx context.Context, heartbeater Heartbeater, cbg
 
 // shardedDCPHeartbeatListener uses cbgt's cfg to manage node list
 type shardedDCPHeartbeatListener struct {
-	cfg        cbgt.Cfg      // cbgt cfg being used for import
+	cfg        cbgt.Cfg      // cbgt cfg being used for shardedDCP
 	mgr        *cbgt.Manager // cbgt manager associated with this import node
 	ctx        *CbgtContext
 	terminator chan struct{} // close cfg subscription on close
--- a/base/dcp_sharded.go
+++ b/base/dcp_sharded.go
@@ -37,13 +37,6 @@ const (
 	CBGTIndexTypeSyncGatewayResync    = "syncGateway-resync"
 )
 
-type ShardedDCPFeedType string
-
-const (
-	ShardedDCPFeedTypeImport ShardedDCPFeedType = "import"
-	ShardedDCPFeedTypeResync ShardedDCPFeedType = "resync"
-)
-
 // firstVersionToSupportCollections represents the earliest Sync Gateway release that supports collections.
 var firstVersionToSupportCollections = &ComparableBuildVersion{
 	epoch: 0,
@@ -63,7 +56,7 @@ type CbgtContext struct {
 	Manager           *cbgt.Manager                // Manager is main entry point for initialization, registering indexes
 	Cfg               cbgt.Cfg                     // Cfg manages storage of the current pindex set and node assignment
 	heartbeater       Heartbeater                  // Heartbeater used for failed node detection
-	heartbeatListener *shardedDCPHeartbeatListener // Listener subscribed to failed node alerts from heartbeater
+	heartbeatListener *shardedDCPHeartBeatListener // Listener subscribed to failed node alerts from heartbeater
 	eventHandlers     *sgMgrEventHandlers          // Event handler callbacks
 	dbName            string                       // Database name
 	sourceName        string                       // cbgt source name. Store on CbgtContext for access during teardown
@@ -183,14 +176,14 @@ func StartShardedDCPFeed(ctx context.Context, opts ShardedDCPOptions) (*CbgtCont
 
 // GenerateCBGTIndexName creates an index name for cbgt using the database name suitable for the distributed DCP feed.
 // Given a dbName and feedType, generate a unique and length-constrained index name for CBGT to use as part of their DCP name.
-func GenerateCBGTIndexName(dbName string, feedType ShardedDCPFeedType) (string, error) {
+func GenerateCBGTIndexName(dbName string, feedType string) (string, error) {
 	// Index names *must* start with a letter, so we'll prepend 'db' before the per-database checksum (which starts with '0x')
 	// Don't use Crc32cHashString here because this is intentionally non zero padded to match
 	// existing values.
 	var indexName string
-	if feedType == ShardedDCPFeedTypeImport {
+	if feedType == CBGTIndexTypeSyncGatewayImport {
 		indexName = fmt.Sprintf("db0x%x_index", Crc32cHash([]byte(dbName)))
-	} else if feedType == ShardedDCPFeedTypeResync {
+	} else if feedType == CBGTIndexTypeSyncGatewayResync {
 		indexName = fmt.Sprintf("db0x%x_resync_index", Crc32cHash([]byte(dbName)))
 	} else {
 		return "", fmt.Errorf("unknown index type %s", feedType)
@@ -366,7 +359,7 @@ func initCBGTManager(ctx context.Context, bucket Bucket, spec BucketSpec, cfgSG 
 	//   avoids file system usage, in conjunction with managerLoadDataDir=false in options.
 	dataDir := ""
 
-	eventHandlersCtx, eventHandlersCancel := context.WithCancelCause(ctx)
+	eventHandlersCtx, eventHandlersCancel := context.WithCancel(ctx)
 	eventHandlers := &sgMgrEventHandlers{ctx: eventHandlersCtx, ctxCancel: eventHandlersCancel}
 
 	// Specify one feed per pindex
@@ -513,7 +506,7 @@ func getMinNodeVersion(cfg cbgt.Cfg) (*ComparableBuildVersion, error) {
 // Stop unregisters the listener from the heartbeater, and stops it and associated handlers.
 func (c *CbgtContext) Stop(ctx context.Context) {
 	if c.eventHandlers != nil {
-		c.eventHandlers.ctxCancel(errors.New("CbgtContext is stopping, cancelling event handlers"))
+		c.eventHandlers.ctxCancel()
 	}
 
 	if c.heartbeatListener != nil {
@@ -544,7 +537,7 @@ func (c *CbgtContext) RemoveFeedCredentials(dbName string) {
 }
 
 // Format of dest key for retrieval of import dest from cbgtDestFactories
-func DestKey(dbName string, scope string, collections []string, feedType ShardedDCPFeedType) string {
+func DestKey(dbName string, scope string, collections []string, feedType string) string {
 	sort.Strings(collections)
 	collectionString := ""
 	onlyDefault := true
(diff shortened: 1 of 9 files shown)


base/dcp_sharded.go
```diff
@@ -586,6 +595,69 @@ func registerHeartbeatListener(ctx context.Context, heartbeater Heartbeater, cbg
 	return shardedDCPHeartbeatListener, nil
 }
 
+type cfgNodePoller struct {
+	nodeDefsKnownKey  string
+	nodeDefsWantedKey string
+	nodeDefsKnownCAS  uint64 // to store the last known cas of nodeDefsKnown
+	nodeDefsWantedCAS uint64 // to store the last known cas of nodeDefsWanted
```
torcolvin: We want to construct this as map[string]uint64 of doc name -> cas so we can extend this to multiple documents that we want to track.

From the perspective of heartbeat we only care about nodeDefsKnown and nodeDefsWanted but for cbgt and cfg we care about all files that `cbgt.Cfg.Subscribe` has been called on. This is should be all the files in https://github.com/couchbase/cbgt/blob/99fce75d027f0e89f8f8f5a70a686e831c594af7/manager.go#L1446

I think we can do something where this struct links into `NewCfgSG`.

You can add an optional `cfgNodePoller` to the the `NewCfgSG` constructor. If this is set, you know we _always_ want to pay attention to these document keys but we also want to pay attention to anything that is called by `CfgSG.Subscribe` and register that document name in this map.

base/dcp_sharded.go
```diff
@@ -586,6 +595,69 @@ func registerHeartbeatListener(ctx context.Context, heartbeater Heartbeater, cbg
...
+	cfg               *CfgSG
+}
+
+func newCfgNodePoller(datastore DataStore, cfg cbgt.Cfg) (*cfgNodePoller, error) {
+
+	nodeDefsKnownKey := cbgt.CfgNodeDefsKey(cbgt.NODE_DEFS_KNOWN)
+	_, nodeDefsKnownCas, err := datastore.GetRaw(nodeDefsKnownKey)
+	if err != nil {
+		return nil, err
+	}
```
torcolvin: We shouldn't need to necessarily do this from the constructor, we can do this when we start polling.

Note that you want to test scenarios including but not limited to these circumstances:

1. start with document not present
2. start with document present
3. turn document into a tombstone
4. resurrect document
5. purge documents

DocNotFound is not an error condition for getting a cas value. We can log at WRN if there's any other type of error, I'd expect those to be things like circuitbreaker tripped and only happen when CBS goes down.

torcolvin (commented):
Think about how to write tests for this code, we want to make sure this is robust for a variety of errors. Doc not found errors are going to look different for rosmar and CBS. You can make the polling window very fast for the sake of doing a unit test.

base/dcp_sharded.go
```diff
@@ -586,6 +595,75 @@ func registerHeartbeatListener(ctx context.Context, heartbeater Heartbeater, cbg
 	return shardedDCPHeartbeatListener, nil
 }
 
+type cfgNodePoller struct {
```
torcolvin, line 602: Let's document what this does:


```suggestion
// cfgNodePoller is a companion to a cbgt.Cfg instance that polls a datastore for changes to selected document keys.
//
// cfgNodePoller.Register is used to listen for changes to keys.
// fireEvent is called whenever a document's cas value changes. This should map to cbgt.Cfg.FireEvent function.
type cfgNodePoller struct {
```

base/dcp_sharded.go
```diff
@@ -586,6 +595,75 @@ func registerHeartbeatListener(ctx context.Context, heartbeater Heartbeater, cbg
...
+	ctx          context.Context
+	keyWatcher   map[string]uint64
+	datastore    DataStore
+	fireEvent    CfgEventNotifyFunc
+	lock         sync.Mutex
+	pollInterval time.Duration
+}
+
+func newCfgNodePoller(ctx context.Context, datastore DataStore, fireEvent CfgEventNotifyFunc, pollInterval time.Duration) *cfgNodePoller {
```
torcolvin, line 612:
```suggestion
// newCfgNodePoller constructs a poller instance on the datastore. fireEvent is a cbgt.Cfg.FireEvent function called whenever a document cas changes. 
func newCfgNodePoller(ctx context.Context, datastore DataStore, fireEvent CfgEventNotifyFunc, pollInterval time.Duration) *cfgNodePoller {
```

base/dcp_sharded.go
```diff
@@ -586,6 +595,75 @@ func registerHeartbeatListener(ctx context.Context, heartbeater Heartbeater, cbg
...
+		datastore:    datastore,
+		fireEvent:    fireEvent,
+		pollInterval: pollInterval,
+	}
+
+	p.StartPolling(ctx)
+	return p
+}
+
+func (p *cfgNodePoller) Register(key string) error {
```
torcolvin, line 643:
```suggestion
// Register adds a document key for polling.
func (p *cfgNodePoller) Register(key string) error {
```

base/dcp_sharded.go
```diff
@@ -586,6 +595,75 @@ func registerHeartbeatListener(ctx context.Context, heartbeater Heartbeater, cbg
...
+
+	cas, err := p.datastore.Get(key, nil)
+	if err != nil && !IsDocNotFoundError(err) {
+		return err
+	}
+	p.keyWatcher[key] = cas
+	return nil
+}
+
+func (p *cfgNodePoller) Poll() {
```
torcolvin:
```suggestion
// poll reads the cas for all of the registered documents and calls fireEvent if any cas have changed since the last time poll was called.
func (p *cfgNodePoller) poll() {
```

nit: suggest making `poll` "private" since this should not be called from external interface.

base/dcp_sharded.go
```diff
@@ -586,6 +595,75 @@ func registerHeartbeatListener(ctx context.Context, heartbeater Heartbeater, cbg
...
+
+	p := &cfgNodePoller{
+		ctx:          ctx,
+		keyWatcher:   make(map[string]uint64),
+		datastore:    datastore,
+		fireEvent:    fireEvent,
+		pollInterval: pollInterval,
+	}
+
+	p.StartPolling(ctx)
```
torcolvin:
```suggestion
	p.startPolling(ctx)
```

base/dcp_sharded.go
```diff
@@ -586,6 +595,75 @@ func registerHeartbeatListener(ctx context.Context, heartbeater Heartbeater, cbg
...
+			p.lock.Lock()
+			defer p.lock.Unlock()
+			p.keyWatcher[key] = newCas
+
+			p.fireEvent(key, newCas, nil)
+		}
+	}
+}
+
+func (p *cfgNodePoller) StartPolling(ctx context.Context) {
```
torcolvin:
```suggestion
// startPolling launches an indefinite loop to listen for document cas changes.
func (p *cfgNodePoller) startPolling(ctx context.Context) {
```

base/dcp_sharded.go
```diff
@@ -586,6 +595,75 @@ func registerHeartbeatListener(ctx context.Context, heartbeater Heartbeater, cbg
...
+	ticker := time.NewTicker(p.pollInterval)
+	go func() {
+		defer FatalPanicHandler()
+		for {
+			select {
+			case <-ctx.Done():
+				ticker.Stop()
+				return
+			case <-ticker.C:
+				p.Poll()
```
torcolvin:
```suggestion
				p.poll()
```

base/dcp_sharded.go
```diff
@@ -586,6 +595,75 @@ func registerHeartbeatListener(ctx context.Context, heartbeater Heartbeater, cbg
...
+			WarnfCtx(p.ctx, "error reading doc: %s %v", key, err)
+			continue
+		}
+		if oldCas != newCas {
+			p.lock.Lock()
+			defer p.lock.Unlock()
+			p.keyWatcher[key] = newCas
+
+			p.fireEvent(key, newCas, nil)
+		}
```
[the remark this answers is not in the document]
torcolvin: This is extremely important to do. I think you can probably just lock at the top of this function while polling happens and it makes this easier but this is the sort of thing that a test should have caught:

1. change one document
2. change second document
3. run poll
4. <deadlock>

base/sg_cluster_cfg.go
```diff
@@ -44,7 +45,7 @@ var ErrCfgCasError = &cbgt.CfgCASError{}
 //
 // urlStr: single URL or multiple URLs delimited by ';'
 // bucket: couchbase bucket name
```
torcolvin:
```suggestion
// NewCfgSG returns a Cfg implementation that reads/writes its entries
// from/to a couchbase datastore. All document names will start with keyPrefix.
// If useNodePoller is true, then any document changes are received by node polling. If useNodePoller is not true, then the caller needs to register event changes itself by calling FireEvent.
//
// The caching feed implements FireEvent calls by looking for document changes starting with keyPrefix and calling FireEvent.
```

base/sg_cluster_cfg.go
```diff
@@ -133,6 +138,10 @@ func (c *CfgSG) Subscribe(cfgKey string, ch chan cbgt.CfgEvent) error {
 	}
 	c.subscriptions[cfgKey] = append(a, ch)
 
+	if c.nodePoller != nil {
+		c.nodePoller.Register(c.sgCfgBucketKey(cfgKey))
+	}
+
 	c.lock.Unlock()
```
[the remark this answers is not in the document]
torcolvin: yes, this lock needs to be a defer and also not blocked for error. It's OK to lock for this entire function for readability. If this was a performance bottleneck it would need to lock more locally.

db/background_mgr_resync_dcp.go
```diff
@@ -246,13 +246,16 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
 		}
 
 		// CFG creation:
-		resyncCfg, err := base.NewCfgSG(ctx, db.MetadataStore, db.MetadataKeys.ResyncCfgPrefix())
+		ctx, cancel := context.WithCancelCause(ctx)
```
torcolvin: The govet issue here needs to be addressed:

lostcancel: the cancel function is not used on all paths (possible context leak) (govet)

db/background_mgr_resync_dcp.go
torcolvin, line 284: lostcancel: the cancel function is not used on all paths (possible context leak) (govet)

The govet issue here need to be addressed.

One way of addressing this is use err in named defer when we define cancel:

```
    defer func() {
        if err != nil {
            cancel(fmt.Errorf("resync run function failed: %w", err)
        } else {
            cancel(fmt.Errorf("resync run terminated normally"))
       }
```

db/background_mgr_resync_dcp.go
RIT3shSapata: I'd added inside defer goroutine at the end, but I'm now moving this as a defer right after this

base/dcp_sharded.go
```diff
@@ -586,6 +595,107 @@ func registerHeartbeatListener(ctx context.Context, heartbeater Heartbeater, cbg
...
+// getWatcher returns a copy of the keyWatcher map containing the current CAS values for all registered keys.
+// This method is thread-safe and creates a new map to avoid external modifications to the internal state.
+func (p *cfgNodePoller) getWatcher() map[string]uint64 {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	watcher := make(map[string]uint64, len(p.keyWatcher))
+	for key, cas := range p.keyWatcher {
+		watcher[key] = cas
+	}
+	return watcher
```
torcolvin: nit:

You can use `maps.Clone` here.

base/dcp_sharded.go
```diff
@@ -586,6 +595,107 @@ func registerHeartbeatListener(ctx context.Context, heartbeater Heartbeater, cbg
...
+// the fireEvent callback is invoked with the key, new CAS value, and nil error. The keyWatcher is updated with
+// the new CAS values. This method is thread-safe and uses optimistic locking to avoid blocking during I/O operations.
+func (p *cfgNodePoller) poll() {
+
+	watcher := p.getWatcher()
+
+	for key, oldCas := range watcher {
+		newCas, err := p.datastore.Get(key, nil)
+		if err != nil && !IsDocNotFoundError(err) {
+			WarnfCtx(p.ctx, "error reading doc: %s %v", UD(key), err)
```
torcolvin:
```suggestion
			WarnfCtx(p.ctx, "cfgNodePoller: error polling doc: %s %v, skipping polling", UD(key), err)
```

base/dcp_sharded.go
```diff
@@ -586,6 +595,107 @@ func registerHeartbeatListener(ctx context.Context, heartbeater Heartbeater, cbg
...
+			continue
+		}
+		p.lock.Lock()
+		currentCas, ok := p.keyWatcher[key]
+		if !ok || currentCas != oldCas {
+			p.lock.Unlock()
+			continue
+		}
+		p.keyWatcher[key] = newCas
+		p.lock.Unlock()
```
torcolvin: It's an anti pattern to not call Unlock() in a defer in case this code crashes.

The way around this is to create a function like:

```
func(p *cfgNodePoller) updateCas(key string, newCas) {
        p.Lock()
        defer p.lock.Unlock()
		p.keyWatcher[key] = newCas
}
```

to mark whether the function is updated.

I don't see how we need to check the cas _again_ here, I think you can just have a function to update the cas.

base/sg_cluster_cfg.go
```diff
@@ -127,13 +131,19 @@ func (c *CfgSG) Subscribe(cfgKey string, ch chan cbgt.CfgEvent) error {
...
 	}
 	c.subscriptions[cfgKey] = append(a, ch)
 
-	c.lock.Unlock()
+	if c.nodePoller != nil {
+		err := c.nodePoller.Register(c.sgCfgBucketKey(cfgKey))
+		if err != nil {
+			return err
+		}
+	}
```
torcolvin: very small nit, prefer non nested code when possible.

```suggestion
	if c.nodePoller == nil {
		return nil
	}
	return c.nodePoller.Register(c.sgCfgBucketKey(cfgKey))
```

torcolvin (commented):
LGTM mostly, I think waiting on this PR to write some more tests is OK in case it changes the code.

Some nits which are mostly go idioms but the Lock should be addressed.

db/background_mgr_resync_dcp.go
```diff
@@ -246,13 +246,17 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
 		}
 
 		// CFG creation:
-		resyncCfg, err := base.NewCfgSG(ctx, db.MetadataStore, db.MetadataKeys.ResyncCfgPrefix())
+		resyncCtx, cancelResync := context.WithCancelCause(ctx)
+		defer cancelResync(nil)
```
torcolvin, line 258: The way this should be canceled needs to be robust to any failure conditions. This means it should probably be done with a named return and then to cancel in a defer:

change:
```
func (r *ResyncManagerDCP) Init(ctx context.Context, options map[string]any, clusterStatus []byte) error {
```

to
```
func (r *ResyncManagerDCP) Init(ctx context.Context, options map[string]any, clusterStatus []byte) (err error) {
```

Then you can do:

```
defer func() {
    if err != nil {
          cancelResync(err)
    } else {
           cancelResync(errors.New("resync ended normally"))
    }
}()
```

This means you use the existing error return values for resync but it will still cancel the resync normally and it is traced through with that error.

db/background_mgr_resync_dcp.go
```diff
@@ -246,13 +246,17 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
 		}
 
 		// CFG creation:
-		resyncCfg, err := base.NewCfgSG(ctx, db.MetadataStore, db.MetadataKeys.ResyncCfgPrefix())
+		resyncCtx, cancelResync := context.WithCancelCause(ctx)
```
torcolvin, line 250: Pending https://github.com/couchbase/sync_gateway/pull/8172 but I think this probably belongs at the top of this function, I can't think of too many reasons we want it much lower.

db/background_mgr_resync_dcp.go
torcolvin: looking at these cancel functions makes me see that that `resyncHB.Stop(ctx)` needs to happen immediately after 

```
		err = resyncHB.Start(ctx)
		if err != nil {
			return fmt.Errorf("Error starting resync heartbeater: %v", err)
		}
```

or it will leak this go routine.

db/background_mgr_resync_dcp.go
RIT3shSapata: Do you want me to rebase it to #8172?

torcolvin: No, this is fine, I'll merge #8172 after this one.

torcolvin (approved):

PR #8145, opened 2026-04-02
CBG-5263 workaround TestNewlyCreateSGWPermissions failure

torcolvin:
CBG-5263 workaround TestNewlyCreateSGWPermissions failure

If you keep persistent config but move the online call back to the original location, this intermittent test failure is consistently reproduced.

## Pre-review checklist

### DIFF
--- a/rest/admin_api_auth_test.go
+++ b/rest/admin_api_auth_test.go
@@ -1010,9 +1010,11 @@ func TestNewlyCreateSGWPermissions(t *testing.T) {
 	rt := NewRestTester(t, &RestTesterConfig{
 		AdminInterfaceAuthentication:    true,
 		enableAdminAuthPermissionsCheck: true,
+		PersistentConfig:                true,
 	})
 	defer rt.Close()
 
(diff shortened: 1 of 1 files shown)


gregns1 (changes requested):
It feels like the correct fix for this test is to have a wait for online state retry loop for after the `sync_gateway_configurator` `_online` update. This feels prone to someone adding some more assertion to the test cases below the `_online` test case and hitting the same problem again. 

It also should not be the case that we panic here. There is some weirdness here given my local debugging shows the bucket as not nil when returned from `GetActiveDatabase` but seems to be nil after getting `AccessLock`. Not sure if this is out of scope for this ticket but certainly something we should look into.

torcolvin:
I want to push this PR as a workaround so the integration tests don't fail and then have you actually fix this ticket separately.

Once this PR is merged, leave the ticket open to fix. I think this is a follow-on issue to CBG-4651. I think the issue existed before but CBG-4651 made it a lot worse.

gregns1 (approved):

PR #8151, opened 2026-04-07
CBG-5232: cache memory managment

gregns1:
CBG-5232

Memory based eviction for delta cache and revision cache. 
Slight difference to original design, Original design was that a memory controller would control the eviction process, when implementing I felt it was cleaner for the orchestrator to do this given it.s job is to orchestrator between the caches. So memory controller simply sits on each shard and counts memory for that shard and lets the orchestrator know when its over capacity. 

## Pre-review checklist

- [ ] Link upstream PRs

- [x]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/433/
- [x] https://jenkins.sgwdev.com/job/SyncGatewayIntegration/434/ (disable rev cache)

### DIFF
--- a/db/cache_memory_controller.go
+++ b/db/cache_memory_controller.go
@@ -1,0 +1,71 @@
+/*
+Copyright 2026-Present Couchbase, Inc.
+
+Use of this software is governed by the Business Source License included in
+the file licenses/BSL-Couchbase.txt.  As of the Change Date specified in that
+file, in accordance with the Business Source License, use of this software will
+be governed by the Apache License, Version 2.0, included in the file
+licenses/APL2.txt.
+*/
+
+package db
+
+import (
+	"context"
+	"sync/atomic"
+
+	"github.com/couchbase/sync_gateway/base"
+)
+
+// CacheMemoryController manages a single memory budget shared between a revision cache and
+// its associated delta cache. All accounting is done via atomic operations so the hot-path
+// capacity check (IsOverCapacity) is entirely lock-free.
+type CacheMemoryController struct {
+	capacity       int64            // total byte limit; 0 = unlimited
+	currBytesCount atomic.Int64     // bytes held by the revision cache + delta cache shard
+	totalBytesStat *base.SgwIntStat // shared observable stat. Overall footprint of the caches combined.
+}
+
+// globalCacheAccessCounter is a monotonically increasing counter stamped onto every
+// cache item on insert and on access. It provides a total ordering across both the
+// revision and delta caches for cross-cache LRU eviction decisions.
+// Using a counter rather than time.Now() keeps the hot-path cost to a single
+// atomic increment (~1–2 ns vs ~25 ns for a clock read).
+var globalCacheAccessCounter atomic.Uint64
+
+func nextAccessOrder() uint64 {
+	return globalCacheAccessCounter.Add(1)
+}
+
+func newCacheMemoryController(capacityBytes int64, totalBytesStat *base.SgwIntStat) *CacheMemoryController {
+	return &CacheMemoryController{
+		capacity:       capacityBytes,
+		totalBytesStat: totalBytesStat,
+	}
+}
+
+// IsOverCapacity will check if overall memory footprint is over the configured for this shard/cache.
+func (mc *CacheMemoryController) IsOverCapacity() bool {
+	if mc.capacity == 0 {
+		return false
+	}
+	return mc.currBytesCount.Load() > mc.capacity
+}
+
+func (mc *CacheMemoryController) bytesToEvict() int64 {
+	excess := mc.currBytesCount.Load() - mc.capacity
+	if excess < 0 {
+		return 0
+	}
+	return excess
+}
+
+func (mc *CacheMemoryController) incrementBytesCount(ctx context.Context, n int64) {
+	mc.currBytesCount.Add(n)
+	mc.totalBytesStat.Add(n)
+}
+
+func (mc *CacheMemoryController) decrementBytesCount(ctx context.Context, n int64) {
+	mc.currBytesCount.Add(-n)
+	mc.totalBytesStat.Add(-n)
+}
--- a/db/cache_memory_controller.go
+++ b/db/cache_memory_controller.go
@@ -11,7 +11,6 @@ licenses/APL2.txt.
 package db
 
 import (
-	"context"
 	"sync/atomic"
 
 	"github.com/couchbase/sync_gateway/base"
@@ -60,12 +59,12 @@ func (mc *CacheMemoryController) bytesToEvict() int64 {
 	return excess
 }
 
-func (mc *CacheMemoryController) incrementBytesCount(ctx context.Context, n int64) {
+func (mc *CacheMemoryController) incrementBytesCount(n int64) {
 	mc.currBytesCount.Add(n)
 	mc.totalBytesStat.Add(n)
 }
 
-func (mc *CacheMemoryController) decrementBytesCount(ctx context.Context, n int64) {
+func (mc *CacheMemoryController) decrementBytesCount(n int64) {
 	mc.currBytesCount.Add(-n)
 	mc.totalBytesStat.Add(-n)
 }
--- a/db/cache_memory_controller.go
+++ b/db/cache_memory_controller.go
@@ -37,6 +37,8 @@ func (mc *CacheMemoryController) IsOverCapacity() bool {
 	if mc.capacity == 0 {
 		return false
 	}
+	// IsOverCapacity returns true only when usage strictly exceeds the limit,
+	// so a cache exactly at capacity is not evicted.
 	return mc.currBytesCount.Load() > mc.capacity
 }
 
--- a/db/cache_memory_controller.go
+++ b/db/cache_memory_controller.go
@@ -20,15 +20,15 @@ import (
 // its associated delta cache. All accounting is done via atomic operations so the hot-path
 // capacity check (IsOverCapacity) is entirely lock-free.
 type CacheMemoryController struct {
-	capacity       int64            // total byte limit; 0 = unlimited
-	currBytesCount atomic.Int64     // bytes held by the revision cache + delta cache shard
-	totalBytesStat *base.SgwIntStat // shared observable stat. Overall footprint of the caches combined.
+	capacity           int64            // total byte limit; 0 = unlimited
+	bytesInUseForShard atomic.Int64     // bytes held by the revision cache + delta cache shard
+	globalUsageStat    *base.SgwIntStat // shared observable stat. Overall footprint of the caches combined.
 }
 
 func newCacheMemoryController(capacityBytes int64, totalBytesStat *base.SgwIntStat) *CacheMemoryController {
 	return &CacheMemoryController{
-		capacity:       capacityBytes,
-		totalBytesStat: totalBytesStat,
+		capacity:        capacityBytes,
+		globalUsageStat: totalBytesStat,
 	}
 }
 
@@ -39,11 +39,11 @@ func (mc *CacheMemoryController) IsOverCapacity() bool {
 	}
 	// IsOverCapacity returns true only when usage strictly exceeds the limit,
 	// so a cache exactly at capacity is not evicted.
-	return mc.currBytesCount.Load() > mc.capacity
+	return mc.bytesInUseForShard.Load() > mc.capacity
 }
 
 func (mc *CacheMemoryController) bytesToEvict() int64 {
-	excess := mc.currBytesCount.Load() - mc.capacity
+	excess := mc.bytesInUseForShard.Load() - mc.capacity
 	if excess < 0 {
 		return 0
 	}
@@ -51,11 +51,11 @@ func (mc *CacheMemoryController) bytesToEvict() int64 {
 }
 
 func (mc *CacheMemoryController) incrementBytesCount(n int64) {
-	mc.currBytesCount.Add(n)
-	mc.totalBytesStat.Add(n)
+	mc.bytesInUseForShard.Add(n)
+	mc.globalUsageStat.Add(n)
 }
 
 func (mc *CacheMemoryController) decrementBytesCount(n int64) {
-	mc.currBytesCount.Add(-n)
-	mc.totalBytesStat.Add(-n)
+	mc.bytesInUseForShard.Add(-n)
+	mc.globalUsageStat.Add(-n)
 }
--- a/db/cache_memory_controller.go
+++ b/db/cache_memory_controller.go
@@ -25,17 +25,6 @@ type CacheMemoryController struct {
 	totalBytesStat *base.SgwIntStat // shared observable stat. Overall footprint of the caches combined.
 }
 
-// globalCacheAccessCounter is a monotonically increasing counter stamped onto every
-// cache item on insert and on access. It provides a total ordering across both the
-// revision and delta caches for cross-cache LRU eviction decisions.
-// Using a counter rather than time.Now() keeps the hot-path cost to a single
-// atomic increment (~1–2 ns vs ~25 ns for a clock read).
-var globalCacheAccessCounter atomic.Uint64
-
-func nextAccessOrder() uint64 {
-	return globalCacheAccessCounter.Add(1)
-}
-
 func newCacheMemoryController(capacityBytes int64, totalBytesStat *base.SgwIntStat) *CacheMemoryController {
 	return &CacheMemoryController{
 		capacity:       capacityBytes,
--- a/db/revision_cache_bypass.go
+++ b/db/revision_cache_bypass.go
@@ -31,7 +31,8 @@ func NewBypassRevisionCache(backingStores map[uint32]RevisionCacheBackingStore, 
 	}
 }
 
-// Get fetches the revision for the given docID and revID immediately from the bucket.
+// Get fetches the revision for the given docID and revID immediately from the bucket. Always returns false for checking
+// for memory based eviction as there is nothing to evict from when using bypass revision cache.
 func (rc *BypassRevisionCache) Get(ctx context.Context, docID, versionString string, collectionID uint32, loadBackup bool) (docRev DocumentRevision, b bool, err error) {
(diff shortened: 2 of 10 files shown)


db/revision_cache_lru.go
```diff
@@ -341,13 +366,13 @@ func (rc *LRURevisionCache) getValue(ctx context.Context, docID, docVersionStrin
 		} else {
 			value.cv = currentVersion
 		}
-		value.itemLoaded.Store(false)
+		value.accessOrder.Store(nextAccessOrder()) // stamp on insert
 		rc.cache[key] = rc.lruList.PushFront(value)
 		rc.cacheNumItems.Add(1)
```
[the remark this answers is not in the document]
gregns1: This is by design, I don't want to be adding an extra atomic increment for each operation at the cache. This does mean when eviction fires it could evict the item more recently used technically but this is a worth while technicality for one less atomic increment on every cache hit.

db/cache_memory_controller.go
```diff
@@ -0,0 +1,70 @@
...
+	currBytesCount atomic.Int64     // bytes held by the revision cache + delta cache shard
+	totalBytesStat *base.SgwIntStat // shared observable stat. Overall footprint of the caches combined.
+}
+
+// globalCacheAccessCounter is a monotonically increasing counter stamped onto every
+// cache item on insert and on access. It provides a total ordering across both the
+// revision and delta caches for cross-cache LRU eviction decisions.
+// Using a counter rather than time.Now() keeps the hot-path cost to a single
+// atomic increment (~1–2 ns vs ~25 ns for a clock read).
+var globalCacheAccessCounter atomic.Uint64
```
bbrks: This is process-global - should be scoped to a database!

db/revision_cache_orchestrator.go
```diff
@@ -55,16 +77,59 @@ func (c *RevisionCacheOrchestrator) UpdateDelta(ctx context.Context, docID, from
...
+	docRev, _, err := c.revisionCache.Get(ctx, docID, fromVersionString, collectionID, RevCacheLoadBackupRev)
 	if err != nil {
 		return docRev, err
 	}
 	if c.deltaCache != nil {
 		cachedDelta := c.deltaCache.getCachedDelta(ctx, docID, fromVersionString, toVersionString, collectionID)
 		docRev.Delta = cachedDelta
 	}
+	// check for memory based eviction
+	c.triggerMemoryEviction()
```
bbrks: This doesn't need to be called if getCachedDelta is not creating an entry right? Similarly if revisionCache.Get doesn't insert anything

I know there's an early return inside but it's still doing the work to acquire the lock to check.

db/revision_cache_orchestrator.go
```diff
@@ -55,16 +77,59 @@ func (c *RevisionCacheOrchestrator) UpdateDelta(ctx context.Context, docID, from
...
+	}
+	for bytesNeededToEvict > numBytesRemoved {
+		if c.deltaCache != nil {
+			deltaCandidateOrder = c.deltaCache.peekLRUTailAccessOrder()
+		}
+		revCandidateOrder := c.revisionCache.peekLRUTailAccessOrder()
+		if revCandidateOrder == 0 && deltaCandidateOrder == 0 {
+			// Both caches are empty — nothing more to evict.
+			break
+		}
```
bbrks, line 123: `evictLRUTail()` already tolerates the empty cache case. I don't think this is needed?

db/revision_cache_bypass.go
```diff
@@ -97,7 +98,7 @@ func (rc *BypassRevisionCache) GetActive(ctx context.Context, docID string, coll
 
 	rc.bypassStat.Add(1)
 
-	return docRev, nil
+	return docRev, checkForMemoryEviction, nil
```
bbrks: This function always returns `false` for `checkForMemoryEviction` and I think it reads clearer if it isn't using the variable name without a set.

db/revision_cache_orchestrator.go
```diff
@@ -10,28 +10,48 @@ licenses/APL2.txt.
...
+)
 
 // RevisionCacheOrchestrator orchestrates between the revisionCache and a deltaCache.
 type RevisionCacheOrchestrator struct {
-	revisionCache *LRURevisionCache
-	deltaCache    *LRUDeltaCache
+	revisionCache    *LRURevisionCache      // holds document revisions
+	deltaCache       *LRUDeltaCache         // holds computed deltas, only initialized when delta sync is enabled
+	memoryController *CacheMemoryController // used to control memory usage of revision cache and delta cache combined
+	evictionLock     sync.Mutex             // This is to synchronise the eviction process so we don;t have multiple goroutines fighting to evict
```
bbrks:
```suggestion
	evictionLock     sync.Mutex             // This is to synchronise the eviction process so we don't have multiple goroutines fighting to evict
```

db/revision_cache_lru.go
```diff
@@ -392,18 +428,22 @@ func (rc *LRURevisionCache) removeValueForFailedLoad(value *revCacheValue) {
...
-		// delete item from lookup map
 		delete(rc.cache, value.itemKey)
 		numItemsEvicted++
-		numBytesEvicted += value.getItemBytes()
+		// Only count bytes if they were already accounted in the stat (memStateSized).
+		// If the item was still in memStateLoading the in-flight CAS will fail and skip
+		// the increment, so there is nothing to decrement.
+		if value.memState.Swap(memStateRemoved) == memStateSized {
+			numBytesEvicted += value.getItemBytes()
+		}
```
bbrks, line 443: Not specific to this code, but the whole eviction, size tracking, LRU/monotonic counter approach seems complicated, and I'd potentially like a walkthrough of the design to properly understand this.

bbrks (changes requested):
I am conscious we're trying to simplify the cache generally, but this seems more complicated conceptually to wrap my head around the safety of.

With the complexity of the memory tracking/eviction, LRU/monotonic counter in particular, I'd potentially like a walkthrough of the design to properly understand this.

db/revision_cache_test.go
```diff
@@ -2657,3 +2678,664 @@ func TestMultipleRevCacheItemsForSameDocs(t *testing.T) {
...
+		}()
+		go func() {
+			defer wg.Done()
+			orchestrator.Remove(ctx, docID, cv.String(), testCollectionID)
+		}()
+	}
+	wg.Wait()
+
+	// With MaxItemCount=1, 1 item should remain.
+	// get copilot to suggest assertion here
```
bbrks: Stale comment I Think?

db/util_testing.go
```diff
@@ -1098,11 +1098,16 @@ func (c *collectionRevisionCache) PutRevEntry(t *testing.T, ctx context.Context,
 	} else {
 		cache = (*c.revCache).(*RevisionCacheOrchestrator)
 	}
+	val := cache.revisionCache.getValue(ctx, docRev.DocID, docRev.RevID, c.collectionID, false)
 	// Remove any existing entry so that store() below is not a no-op.
 	cache.Remove(ctx, docRev.DocID, docRev.RevID, c.collectionID)
+	if val.memState.Swap(memStateRemoved) == memStateSized {
```
bbrks: Maybe need a `nil` check here for an uncached value?

```suggestion
	if val != nil && val.memState.Swap(memStateRemoved) == memStateSized {
```

db/revision_cache_orchestrator.go
```diff
@@ -55,16 +78,79 @@ func (c *RevisionCacheOrchestrator) UpdateDelta(ctx context.Context, docID, from
...
+		numBytesRemoved += bytes
+	}
+	c.memoryController.decrementBytesCount(numBytesRemoved)
+}
+
+// evictOneItem removes one item from either the revision or delta cache using round-robin
+// selection. If the primary cache is empty, it immediately falls back to the other cache.
+// Returns (bytes freed, true) when an item was removed, or (0, false) when both are empty.
+// Must only be called while holding evictionLock.
+func (c *RevisionCacheOrchestrator) evictOneItem() (int64, bool) {
```
bbrks: Since this requires the `evictionLock` - we should name it with a leading underscore to make that more obvious. I was confused about how `c.evictNextFromRev` was thread-safe before realising this.

```suggestion
// Must only be called while holding evictionLock.
func (c *RevisionCacheOrchestrator) _evictOneItem() (int64, bool) {
```

db/revision_cache_orchestrator.go
```diff
@@ -55,16 +78,79 @@ func (c *RevisionCacheOrchestrator) UpdateDelta(ctx context.Context, docID, from
...
+		}
+		// Rev cache exhausted — fall back to delta immediately without burning a loop iteration.
+		return c.deltaCache.evictLRUTail()
+	}
+
+	if bytes, evicted := c.deltaCache.evictLRUTail(); evicted {
+		return bytes, true
+	}
+	// Delta cache exhausted — fall back to rev immediately without burning a loop iteration.
+	return c.revisionCache.evictLRUTail()
```
bbrks: If we're doing this fallback anyway, maybe the function is better written like this? We're basically just flipping the order of the function calls then without repeating ourselves.

```suggestion
	// flip eviction toggle
	c.evictNextFromRev = !c.evictNextFromRev
	first, second := c.revisionCache.evictLRUTail, c.deltaCache.evictLRUTail
	if !c.evictNextFromRev {
		first, second = second, first
	}
	if bytes, evicted := first(); evicted {
		return bytes, true
	}
	return second()
```

db/cache_memory_controller.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/couchbase/sync_gateway/base"
+)
+
+// CacheMemoryController manages a single memory budget shared between a revision cache and
+// its associated delta cache. All accounting is done via atomic operations so the hot-path
+// capacity check (IsOverCapacity) is entirely lock-free.
+type CacheMemoryController struct {
+	capacity       int64            // total byte limit; 0 = unlimited
+	currBytesCount atomic.Int64     // bytes held by the revision cache + delta cache shard
+	totalBytesStat *base.SgwIntStat // shared observable stat. Overall footprint of the caches combined.
```
bbrks: There's a bit of duplication here - I understand why by passing down the stat pointer and keeping the capacity tracking separate from the exported stat, but they should be at least named consistently (`curr` vs `total`)

db/revision_cache_test.go
```diff
@@ -2657,3 +2678,664 @@ func TestMultipleRevCacheItemsForSameDocs(t *testing.T) {
...
+			docID := fmt.Sprintf("doc-%d", i)
+			orchestrator.Put(ctx, makeTestRevision(docID, cv, `{"k":"val"}`), testCollectionID)
+		}()
+	}
+	wg.Wait()
+
+	// After all concurrent Puts and evictions settle, the memory stat should be at or below capacity.
+	assert.Equal(t, int64(44), revStats.cacheMemoryStat.Value(), "we should have 4 items worth of memory")
+	assert.Equal(t, int64(4), revStats.cacheNumItemsStat.Value(), "we should have 4 items in cache")
+}
```
bbrks, line 3343: Is it worth doing a concurrency stress test that is exercising most of the cache methods available to ensure memory tracking is consistent after a sustained workload?

I got Claude to write this which I think does a good job of covering the cases we'd be worried about?

```go
// TestMemoryStatLongTermConsistency runs a high-concurrency mixed workload (Put, Upsert,
// Get, UpdateDelta, GetWithDelta, Remove) against an orchestrator with both rev and delta
// caches enabled and a memory limit that forces frequent memory-based eviction. After the
// workload settles, it audits the full internal state of both caches against the published
// memory stat and the controller's internal counter — catching any divergence between the
// stat and reality across all operation types and eviction paths.
//
// Mirrors the production invariant that any given CV maps to exactly one body: Put/Upsert
// use a writer-side CV, Get/GetWithDelta use the backing-store's CV, and the body bytes for
// each (docID, CV) pair are deterministic. Without this, the "Put-on-existing-Sized" path
// is reached (different bodies for the same cache key) which is impossible in production.
func TestMemoryStatLongTermConsistency(t *testing.T) {
		t.Skip("delta cache is EE only")
	}

	ctx := base.TestCtx(t)

	revStats := newTestRevCacheStats()
	deltaStats := newTestDeltaStats()

	var docCounter, revCounter base.SgwIntStat
	bs := &testBackingStore{getDocumentCounter: &docCounter, getRevisionCounter: &revCounter}

	// MaxBytes deliberately small so memory-based eviction fires constantly throughout the run.
	// MaxItemCount large so number-based eviction stays out of the way (covered by other tests).
	const maxBytes = int64(2000)
	opts := &RevisionCacheOptions{MaxItemCount: 10000, MaxBytes: maxBytes}
	orchestrator := NewRevisionCacheOrchestrator(
		opts, CreateTestSingleBackingStoreMap(bs, testCollectionID),
		revStats, deltaStats, true,
	)

	// loadedCV is what the backing store returns for any doc — Get/GetWithDelta cache under this key.
	// writeCV is what Put/Upsert cache under — distinct from loadedCV so the two paths never collide
	// on the same cache key (production invariant: each write produces a fresh CV).
	loadedCV := Version{Value: 123, SourceID: "test"}
	writeCV := Version{Value: 999, SourceID: "writer"}

	const numDocs = 100
	const opsPerWorker = 1000
	const numWorkers = 250

	docID := func(i int) string { return fmt.Sprintf("doc-%d", i) }
	// bodyForDoc is deterministic per docID so any Put or Upsert on (docID, writeCV) produces the
	// same byte count — upholding the "one CV ⇒ one body" invariant.
	bodyForDoc := func(i int) string {
		return fmt.Sprintf(`{"docID":%d,"payload":"abcdefghijklmnop"}`, i)
	}

	put := func(i int) {
		orchestrator.Put(ctx, makeTestRevision(docID(i), writeCV, bodyForDoc(i)), testCollectionID)
	}
	upsert := func(i int) {
		orchestrator.Upsert(ctx, makeTestRevision(docID(i), writeCV, bodyForDoc(i)), testCollectionID)
	}
	get := func(i int) {
		_, _, _ = orchestrator.Get(ctx, docID(i), loadedCV.String(), testCollectionID, false)
	}
	updateDelta := func(i int) {
		toCV := Version{Value: uint64(1000 + i), SourceID: "test"}.String()
		d := RevisionDelta{DeltaBytes: []byte(fmt.Sprintf("delta-payload-for-doc-%d", i))}
		d.CalculateDeltaBytes()
		orchestrator.UpdateDelta(ctx, docID(i), loadedCV.String(), toCV, testCollectionID, d)
	}
	getWithDelta := func(i int) {
		toCV := Version{Value: uint64(1000 + i), SourceID: "test"}.String()
		_, _ = orchestrator.GetWithDelta(ctx, docID(i), loadedCV.String(), toCV, testCollectionID)
	}
	removeWriter := func(i int) {
		orchestrator.Remove(ctx, docID(i), writeCV.String(), testCollectionID)
	}
	removeLoaded := func(i int) {
		orchestrator.Remove(ctx, docID(i), loadedCV.String(), testCollectionID)
	}

	ops := []func(int){put, upsert, get, updateDelta, getWithDelta, removeWriter, removeLoaded}

	var wg sync.WaitGroup
	for w := range numWorkers {
		wg.Add(1)
		go func(seed int) {
			defer wg.Done()
			r := rand.New(rand.NewSource(int64(seed)))
			for range opsPerWorker {
				op := ops[r.Intn(len(ops))]
				op(r.Intn(numDocs))
			}
		}(w)
	}
	wg.Wait()

	// Audit: walk both caches under their own locks and sum the bytes that should be
	// reflected in the memory stat. Items in memStateLoading must not be counted (their
	// bytes were never added). Items in memStateRemoved must not be present in the map at all.
	var revBytes, deltaBytes, revCount, deltaCount int64

	orchestrator.revisionCache.lock.Lock()
	for _, elem := range orchestrator.revisionCache.cache {
		val := elem.Value.(*revCacheValue)
		switch val.memState.Load() {
		case memStateSized:
			revBytes += val.itemBytes.Load()
		case memStateRemoved:
			t.Errorf("rev cache map still contains a removed entry for key %+v", val.itemKey)
		}
		revCount++
	}
	revListLen := int64(orchestrator.revisionCache.lruList.Len())
	orchestrator.revisionCache.lock.Unlock()

	orchestrator.deltaCache.lock.Lock()
	for _, elem := range orchestrator.deltaCache.cache {
		val := elem.Value.(*deltaCacheValue)
		deltaBytes += val.delta.totalDeltaBytes
		deltaCount++
	}
	deltaListLen := int64(orchestrator.deltaCache.lruList.Len())
	orchestrator.deltaCache.lock.Unlock()

	totalBytes := revBytes + deltaBytes

	// Stat published to the rest of the system must match the bytes actually held.
	assert.Equal(t, totalBytes, revStats.cacheMemoryStat.Value(),
		"published memory stat must equal sum of bytes in rev cache (%d) + delta cache (%d)", revBytes, deltaBytes)

	// Internal controller counter must agree with the published stat — they are mutated
	// together in incrementBytesCount/decrementBytesCount, so any drift between them
	// indicates a missed write site.
	assert.Equal(t, revStats.cacheMemoryStat.Value(), orchestrator.memoryController.currBytesCount.Load(),
		"internal controller counter must equal published memory stat")

	// Item-count stats must match the actual number of entries in each cache.
	assert.Equal(t, revCount, revStats.cacheNumItemsStat.Value(),
		"rev cache num-items stat must equal number of map entries")
	assert.Equal(t, deltaCount, deltaStats.DeltaCacheNumItems.Value(),
		"delta cache num-items stat must equal number of map entries")

	// Map size and list length must agree (no orphaned list entries or leaked map keys).
	assert.Equal(t, revCount, revListLen, "rev cache map size must equal LRU list length")
	assert.Equal(t, deltaCount, deltaListLen, "delta cache map size must equal LRU list length")

	// Memory must remain at or below the configured limit (eviction may briefly go over, but
	// the controller drives back to <= maxBytes once triggerMemoryEviction completes).
	assert.LessOrEqual(t, revStats.cacheMemoryStat.Value(), maxBytes,
		"after workload settles, memory stat must be at or below maxBytes — eviction failed to keep up")
}
```

bbrks (changes requested):

bbrks (approved):
LGTM 👍

PR #8152, opened 2026-04-07
CBG-5061: Sync Function Refactor

RIT3shSapata:
CBG-5061

Describe your PR here...
- Refactored `prepareSyncFn` with new function signature
- made the document preparation for all the processes to use `prepareDocForSyncFn`
- updated downstream paths that used `_rawBody` to perform lazy marshalling correctly

## Pre-review checklist

- [ ] Link upstream PRs

- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/447/ (failed)
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/453/ (failed)
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/468/ (failed)
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/470/ (failed)
- [x]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/471/
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/475/

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -2175,18 +2175,18 @@ func (db *DatabaseCollectionWithUser) prepareDocForSyncFn(ctx context.Context, d
 		return
 	}
 
-	err = validateNewBody(body)
+	mutableBody = body.DeepCopy(ctx)
+	err = validateNewBody(mutableBody)
 	if err != nil {
 		return
 	}
-	mutableBody = body.DeepCopy(ctx)
 	newRevID = revID
-	if tombstone {
-		mutableBody[BodyDeleted] = true
-	}
 
 	mutableBody[BodyId] = doc.ID
 	mutableBody[BodyRev] = newRevID
+	if tombstone {
+		mutableBody[BodyDeleted] = true
+	}
 
 	return
 }
--- a/db/crud.go
+++ b/db/crud.go
@@ -1174,7 +1174,7 @@ func (db *DatabaseCollectionWithUser) Put(ctx context.Context, docid string, bod
 		if conflictErr != nil {
 			if db.ForceAPIForbiddenErrors() {
 				// Make sure the user has permission to modify the document before confirming doc existence
-				mutableBody, metaMap, newRevID, err := db.prepareDocForSyncFn(ctx, doc, newDoc.Body(ctx), newDoc.RevID, false, newDoc.Deleted)
+				mutableBody, metaMap, newRevID, err := db.prepareDocForSyncFn(ctx, doc, newDoc.Body(ctx), newDoc.RevID, newDoc.Deleted)
 				if err != nil {
 					base.InfofCtx(ctx, base.KeyCRUD, "Failed to prepare to run sync function: %v", err)
 					return nil, nil, false, nil, ErrForbidden
@@ -1623,7 +1623,7 @@ func (db *DatabaseCollectionWithUser) PutExistingRevWithBody(ctx context.Context
 // access map for users, roles, handler errors and sync fn exceptions.
 // If syncFn is provided, it will be used instead of the one configured on the database.
 func (db *DatabaseCollectionWithUser) SyncFnDryRun(ctx context.Context, newDoc, oldDoc *Document, userMeta, syncOptions map[string]any, syncFn string, errorLogFunc, infoLogFunc func(string)) (*channels.ChannelMapperOutput, error) {
-	mutableBody, metaMap, _, err := db.prepareDocForSyncFn(ctx, oldDoc, newDoc.Body(ctx), newDoc.RevID, false, newDoc.Deleted)
+	mutableBody, metaMap, _, err := db.prepareDocForSyncFn(ctx, oldDoc, newDoc.Body(ctx), newDoc.RevID, newDoc.Deleted)
 	if err != nil {
 		base.InfofCtx(ctx, base.KeyDiagnostic, "Failed to prepare to run sync function: %v", err)
 		return nil, err
@@ -2169,37 +2169,20 @@ func (db *DatabaseCollectionWithUser) storeOldBodyInRevTreeAndUpdateCurrent(ctx 
 	}
 }
 
-func (db *DatabaseCollectionWithUser) prepareDocForSyncFn(ctx context.Context, doc *Document, body Body, revID string, getAncestor bool, tombstone bool) (mutableBody Body, metaMap map[string]any, newRevID string, err error) {
+func (db *DatabaseCollectionWithUser) prepareDocForSyncFn(ctx context.Context, doc *Document, body Body, revID string, tombstone bool) (mutableBody Body, metaMap map[string]any, newRevID string, err error) {
 	metaMap, err = doc.GetMetaMap(db.UserXattrKey())
 	if err != nil {
 		return
 	}
 
-	if getAncestor {
-		var bodyBytes []byte
-		var ancestorRevID string
-		bodyBytes, ancestorRevID, _, err = db.getAvailableRev(ctx, doc, revID)
-		if err != nil {
-			return
-		}
-		err = mutableBody.Unmarshal(bodyBytes)
-		if err != nil {
-			return
-		}
-		if ancestorRev, ok := doc.History[ancestorRevID]; ok && ancestorRev != nil && ancestorRev.Deleted {
-			mutableBody[BodyDeleted] = true
-		}
-		newRevID = ancestorRevID
-	} else {
-		err = validateNewBody(body)
-		if err != nil {
-			return
-		}
-		mutableBody = body
-		newRevID = revID
-		if tombstone {
-			mutableBody[BodyDeleted] = true
-		}
+	err = validateNewBody(body)
+	if err != nil {
+		return
+	}
+	mutableBody = body.DeepCopy(ctx)
+	newRevID = revID
+	if tombstone {
+		mutableBody[BodyDeleted] = true
 	}
 
 	mutableBody[BodyId] = doc.ID
@@ -2223,7 +2206,20 @@ func (db *DatabaseCollectionWithUser) recalculateSyncFnForActiveRev(ctx context.
 	// In some cases an older revision might become the current one. If so, get its
 	// channels & access, for purposes of updating the doc:
 
-	curBody, _, _, err := db.prepareDocForSyncFn(ctx, doc, nil, doc.GetRevTreeID(), true, false)
+	ancestorBody, ancestorRevID, _, err := db.getAvailableRev(ctx, doc, doc.GetRevTreeID())
+	if err != nil {
+		return
+	}
+	var mutableBody Body
+	err = mutableBody.Unmarshal(ancestorBody)
+	if err != nil {
+		return
+	}
+	var isTombstone bool
+	if ancestorRev, ok := doc.History[ancestorRevID]; ok && ancestorRev != nil && ancestorRev.Deleted {
+		isTombstone = true
+	}
+	curBody, _, _, err := db.prepareDocForSyncFn(ctx, doc, mutableBody, doc.GetRevTreeID(), isTombstone)
 	if err != nil {
 		return
 	}
@@ -2474,14 +2470,7 @@ func (col *DatabaseCollectionWithUser) documentUpdateFunc(
 		return
 	}
 
-	// Ensure _rawBody is initialized for downstream paths, and use
-	// the raw-body-backed mutable body to preserve previous validation behavior.
-	mutableBodyForSyncFn, err := newDoc.GetDeepMutableBody()
-	if err != nil {
-		return
-	}
-
-	mutableBody, metaMap, newRevID, err := col.prepareDocForSyncFn(ctx, doc, mutableBodyForSyncFn, newDoc.RevID, false, newDoc.Deleted)
+	mutableBody, metaMap, newRevID, err := col.prepareDocForSyncFn(ctx, doc, newDoc.Body(ctx), newDoc.RevID, newDoc.Deleted)
 	if err != nil {
 		return
 	}
@@ -2688,7 +2677,7 @@ func (db *DatabaseCollectionWithUser) updateAndReturnDoc(ctx context.Context, do
 			doc.SetCrc32cUserXattrHash()
 
 			var rawSyncXattr, rawMouXattr, rawVvXattr, rawGlobalSync, rawDocBody []byte
-			rawDocBody, rawSyncXattr, rawVvXattr, rawMouXattr, rawGlobalSync, err = doc.MarshalWithXattrs()
+			rawDocBody, rawSyncXattr, rawVvXattr, rawMouXattr, rawGlobalSync, err = doc.MarshalWithXattrs(ctx)
 			if err != nil {
 				return updatedDoc, err
 			}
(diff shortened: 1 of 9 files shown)


RIT3shSapata:
@bbrks I've reverted the use of prepareSyncFn in both the functions for the following reasons:
- `getResyncedDocument`: [copilot's comment](https://github.com/couchbase/sync_gateway/pull/8152#discussion_r3045390483), made a lot of sense of how by using prepare sync function we're just recomputing the access for the same document content.
- `recalculateSyncFnForActiveRev`: By using `prepareSyncFn` in here, resulted in lot of failure in tests with tombstones.

So I just decided to remove the keys that are not required for Sync Function after the `bodyBytes` is fetched.

rest/admin_api_auth_test.go
```diff
@@ -1438,6 +1428,16 @@ func TestNewlyCreateSGWPermissions(t *testing.T) {
...
+		{
+			Method:   "POST",
+			Endpoint: "/db/_offline",
+			Users:    []string{syncGatewayConfigurator},
+		},
+		{
+			Method:   "POST",
+			Endpoint: "/db/_online",
+			Users:    []string{syncGatewayConfigurator},
+		},
```
bbrks: This change is undoing part of the change made on `main` in #8154 

I think we've already covered how to do rebases to resolve conflicts with force pushes, but this squash merge either doesn't seem to be doing the right thing or is error-prone.

db/crud.go
```diff
@@ -2496,7 +2476,14 @@ func (col *DatabaseCollectionWithUser) documentUpdateFunc(
 		return
 	}
 
-	mutableBody, metaMap, newRevID, err := col.prepareSyncFn(doc, newDoc)
+	// Ensure _rawBody is initialized for downstream revision backup paths, and use
+	// the raw-body-backed mutable body to preserve previous validation behavior.
+	mutableBodyForSyncFn, err := newDoc.GetDeepMutableBody()
+	if err != nil {
+		return
+	}
```
bbrks: Why does this codepath require a mutable body (a deep copy - not cheap to make), whereas others can just use `doc.Body(ctx)` to pass in?

The comment I don't think is good justification alone, since `_rawBody` is a lazily marshalled version of the doc, but can always be reproduced from the unmarshalled version when needed. Curious what the previous validation behavior is that we're preserving too - is that just json marshalling validation?

We're paying the cost of the `DeepCopy` already inside `prepareDocForSyncFn` to make any body being passed in a mutable body so this feels like duplicated effort in this case.

db/database.go
```diff
@@ -1750,19 +1750,13 @@ func (db *DatabaseCollectionWithUser) getResyncedDocument(ctx context.Context, d
...
-		var body Body
-		if err := body.Unmarshal(bodyBytes); err != nil {
-			base.WarnfCtx(ctx, "Error unmarshalling body %s/%s for sync function %s", base.UD(docid), rev.ID, err)
-			return
-		}
-		metaMap, err := doc.GetMetaMap(db.UserXattrKey())
+
+		body, metaMap, _, err := db.prepareDocForSyncFn(ctx, doc, nil, rev.ID, true, false)
 		if err != nil {
+			base.WarnfCtx(ctx, "Unable to prepare doc for rev %s: %v", rev.ID, err)
```
bbrks: Include Doc ID in message. RevID alone is not enough to be useful.

db/crud.go
```diff
@@ -2197,30 +2169,43 @@ func (db *DatabaseCollectionWithUser) storeOldBodyInRevTreeAndUpdateCurrent(ctx
...
+			return
+		}
+		err = mutableBody.Unmarshal(bodyBytes)
+		if err != nil {
+			return
+		}
+		if ancestorRev, ok := doc.History[ancestorRevID]; ok && ancestorRev != nil && ancestorRev.Deleted {
+			mutableBody[BodyDeleted] = true
+		}
+		newRevID = ancestorRevID
```
bbrks: This bool feels like the callers need to be doing the `getAvailableRev` bit and pass in the `ancestorRevID` and `tombstone` just like a regular document.

What makes the ancestor case special that it has to be handled differently here? Just avoiding the `validateNewBody` work? 

Could we perhaps add a `shouldValidate` or `isNewDoc` bool instead to control that behavior?

db/crud.go
```diff
@@ -2197,30 +2169,43 @@ func (db *DatabaseCollectionWithUser) storeOldBodyInRevTreeAndUpdateCurrent(ctx
...
+		}
+		mutableBody = body.DeepCopy(ctx)
+		newRevID = revID
+		if tombstone {
+			mutableBody[BodyDeleted] = true
+		}
 	}
 
-	newRevID = newDoc.RevID
+	//newRevID = revID
```
```suggestion
```

can remove now

bbrks (changes requested):
This looks pretty good now - just a few small things

db/crud.go
RIT3shSapata: There were a few tests, that were failing because I initially did not perform a mutable copy. The body would not be sanitized for the next test case and fail due to this. The only difference between `prepareSyncFn` and `prepareDocForSyncFn` was making the mutable copy. The tests that failed were `TestImportInternalPropertiesHandling` and `TestBackupOldRevisionWithAttachments`. 

There was a downstream function that which required newDoc's `rawBody` to be present, `postWriteUpdateHLV` was one function that I observed, didn't check for other functions in downstream. The `rawDoc` being updated was only possible if `newDoc.GetDeepMutableBody()` was invoked. Since I was not passing `newDoc` I couldn't do this inside the `prepareDocForSyncFn`. That's why I make I called the `GetDeepMutableBody` method outside.

I missed removing the deep copy inside the `prepareDocForSyncFn`, will remove that. I would like to know what more explanation you would like in the comment.

db/crud.go
RIT3shSapata: The ancestor case has to be separate for `recalculateSyncFnForActiveRev` which is used to calculate the previous rev's access which will used to handle current rev conflicts. And for the resync case, we need ancestors as we need to recompute the access to all the revisions in the history.

This was previously done separately under their respective functions. Since we wanted to ensure the use of `prepareDocForSyncFn` for all scenarios, I had to add this ancestor case.

The other case is just doing the same thing as what `prepareSyncFn` used to do before. 

I had to add the `tombstone` bool because `newDoc` contains if the document is deleted or not, which is required for the non ancestor case. So i'm not sure if this can be controlled by a single bool value, I will still need two bool values to ensure the behavior doesn't change

db/database.go
```diff
@@ -1750,19 +1750,13 @@ func (db *DatabaseCollectionWithUser) getResyncedDocument(ctx context.Context, d
...
-		var body Body
-		if err := body.Unmarshal(bodyBytes); err != nil {
-			base.WarnfCtx(ctx, "Error unmarshalling body %s/%s for sync function %s", base.UD(docid), rev.ID, err)
-			return
-		}
-		metaMap, err := doc.GetMetaMap(db.UserXattrKey())
+
+		body, metaMap, _, err := db.prepareDocForSyncFn(ctx, doc, nil, rev.ID, true, false)
 		if err != nil {
+			base.WarnfCtx(ctx, "Unable to prepare doc %s for rev %s: %v", doc.ID, rev.ID, err)
```
bbrks: All document IDs need tagging as User Data for log redaction purposes.

This is in our PR templates as a self-review checklist item, and also the Copilot instructions - which will probably catch this if it were re-reviewed.

db/crud.go
```diff
@@ -2197,30 +2169,41 @@ func (db *DatabaseCollectionWithUser) storeOldBodyInRevTreeAndUpdateCurrent(ctx
...
+		if ancestorRev, ok := doc.History[ancestorRevID]; ok && ancestorRev != nil && ancestorRev.Deleted {
+			mutableBody[BodyDeleted] = true
+		}
+		newRevID = ancestorRevID
+	} else {
+		err = validateNewBody(body)
+		if err != nil {
+			return
+		}
+		mutableBody = body
```
bbrks: `mutableBody` is not a mutable-safe copy at this point!

If someone (the sync function) were to change the document in some way, the `body` passed by reference would change too. This is the main point of deep-copying doc bodies before passing into the sync function.

rest/admin_api_auth_test.go
```diff
@@ -1393,16 +1393,6 @@ func TestNewlyCreateSGWPermissions(t *testing.T) {
...
-		{
-			Method:   "POST",
-			Endpoint: "/db/_offline",
-			Users:    []string{syncGatewayConfigurator},
-		},
-		{
-			Method:   "POST",
-			Endpoint: "/db/_online",
-			Users:    []string{syncGatewayConfigurator},
-		},
```
bbrks, line 1405: This is now removing the change made in `main`, rather than basing your code on top of it cleanly

db/crud.go
bbrks: I commented inline but we don't want to completely remove the DeepCopy - it is necessary to avoid document bodies being mutated by the sync function.

The original comment was more in terms of finding an appropriate place to do that work (and only once)

If downstream code is relying on `_rawBody` being present, then change that to use `doc.BodyBytes()` to do the lazy marshalling correctly - that's a bug.

db/crud.go
bbrks: I was basically thinking that `prepareDocForSyncFn` continues to have a single responsibility.

Even the validate function feels a little out of place but I appreciate that's existing code. That feels better suited for something up in the write-codepath, since we'd never need to validate an existing doc body (resync or the ancestor special-case) - but we could always address that separately. I don't think it's doing much harm staying here for now.

E.g.

```go
// prepareDocForSyncFn returns a document body and metaMap suitable for running through the sync function.
// The returned mutableBody is safe for modification without affecting the given body.
func (db *DatabaseCollectionWithUser) prepareDocForSyncFn(ctx context.Context, doc *Document, body Body, revID string, validateBody bool, tombstone bool) (mutableBody Body, metaMap map[string]any, err error) {
	if doc == nil {
		return nil, nil, fmt.Errorf("prepareDocForSyncFn: doc cannot be nil")
	}
	if body == nil {
		return nil, nil, fmt.Errorf("prepareDocForSyncFn: body cannot be nil")
	}

	metaMap, err = doc.GetMetaMap(db.UserXattrKey())
	if err != nil {
		return
	}

	mutableBody = body.DeepCopy(ctx)
	if tombstone {
		mutableBody[BodyDeleted] = true
	}
	mutableBody[BodyId] = doc.ID
	mutableBody[BodyRev] = revID

	return
}
```

and in the cases where we need to do that ancestor lookup, do that (there's only one place):

```go
	// In some cases an older revision might become the current one. If so, get its
	// channels & access, for purposes of updating the doc:
	ancestorBody, ancestorRevID, _, err := db.getAvailableRev(ctx, doc, doc.GetRevTreeID())
	if err != nil {
		return
	}
	var mutableBody Body
	err = mutableBody.Unmarshal(ancestorBody)
	if err != nil {
		return
	}
	var isTombstone bool
	if ancestorRev, ok := doc.History[ancestorRevID]; ok && ancestorRev != nil && ancestorRev.Deleted {
		isTombstone = true
	}
```

In the resync case, that's slightly different and only needs this to fetch backup revision bodies:
```go
	doc.History.forEachLeaf(func(rev *RevInfo) {
		revBodyBytes, _, _, err := db.getRevision(ctx, doc, rev.ID)
		if err != nil {
			base.WarnfCtx(ctx, "Unable to retrieve body for doc %s rev %s: %v", base.UD(doc.ID), rev.ID, err)
			return
		}
		var mutableBody Body
		if err := mutableBody.Unmarshal(revBodyBytes); err != nil {
			base.WarnfCtx(ctx, "Unable to unmarshal body for doc %s rev %s: %v", base.UD(doc.ID), rev.ID, err)
			return
		}
```

bbrks (changes requested):
last set of changes are not safe - the mutable body copy is not being done in all cases any more... See inline comments for more detail and suggestions.

Please let me know if you are struggling with the refactor/test failures and need a hand. The test failures are a suggestion that something is wrong and it may not be in the area of the code you're touching.

db/crud.go
```diff
@@ -2948,15 +2928,19 @@ func (db *DatabaseCollectionWithUser) postWriteUpdateHLV(ctx context.Context, do
 	// we don't need to store revision body backups without delta sync in 4.0, since all clients know how to use the sendReplacementRevs feature
 	backupRev := db.deltaSyncEnabled() && db.deltaSyncRevMaxAgeSeconds() != 0
 	if db.UseXattrs() && backupRev {
-		var newBodyWithAtts = doc._rawBody
+		newBodyWithAtts, err := doc.BodyBytes(ctx)
+		if err != nil {
+			base.WarnfCtx(ctx, "Unable to marshal new revision body during backupRevisionJSON: doc=%q rev=%q cv=%q err=%v ", base.UD(doc.ID), base.UD(doc.GetRevTreeID()), base.UD(doc.HLV.GetCurrentVersionString()), err)
+
+		}
```
bbrks, line 2942: This is unnecessary marshalling work when there are no attachments on a doc. Move inside the attachment if block.

This is on the replicator codepath, and thus sensitive to throughput/performance concerns.

db/crud.go
```diff
@@ -2948,15 +2928,19 @@ func (db *DatabaseCollectionWithUser) postWriteUpdateHLV(ctx context.Context, do
...
 		if len(doc.Attachments()) > 0 {
 			var err error
 			newBodyWithAtts, err = base.InjectJSONProperties(doc._rawBody, base.KVPair{
```
bbrks: accessing `doc._rawBody` here reads weirdly since it's still just relying on the side effect of `BodyBytes` marshalling and setting `_rawBody` on doc. Use the byte slice returned by the `BodyBytes` function.

db/crud.go
```diff
@@ -2197,28 +2169,28 @@ func (db *DatabaseCollectionWithUser) storeOldBodyInRevTreeAndUpdateCurrent(ctx
 	}
 }
 
-func (db *DatabaseCollectionWithUser) prepareSyncFn(doc *Document, newDoc *Document) (mutableBody Body, metaMap map[string]any, newRevID string, err error) {
-	// Marshal raw user xattrs for use in Sync Fn. If this fails we can bail out so we should do early as possible.
+func (db *DatabaseCollectionWithUser) prepareDocForSyncFn(ctx context.Context, doc *Document, body Body, revID string, tombstone bool) (mutableBody Body, metaMap map[string]any, newRevID string, err error) {
```
bbrks: Taking a revID parameter and also returning the same rev ID tells us there's something wrong here.

There's no need to return a rev ID any more after this refactoring work, since the callers always know it. Another clue is most callers ignore the revID parameter.

db/document.go
```diff
@@ -1401,11 +1401,15 @@ func (doc *Document) UnmarshalWithXattrs(ctx context.Context, data, syncXattrDat
...
 	// Grab the rawBody if it's already marshalled, otherwise unmarshal the body
-	if doc._rawBody != nil {
+	docBytes, err := doc.BodyBytes(ctx)
+	if err != nil {
+		return nil, nil, nil, nil, nil, pkgerrors.WithStack(base.RedactErrorf("Failed to MarshalWithXattrs() doc body with id: %s.  Error: %v", base.UD(doc.ID), err))
+	}
+	if docBytes != nil {
 		if !doc.IsDeleted() {
-			data = doc._rawBody
+			data = docBytes
```
bbrks: I am curious if this change is just papering over a test or codepath that was erroneously relying on the presence of `doc._rawBody` without calling through the `BodyBytes()` method to do the lazy marshalling? Which test failed without this change?

This function is already doing the work to marshal `doc._body` when needed, so maybe a better fix would be to set `doc._rawBody` once marshalled? (line ~1426) instead of always calling `BodyBytes()` to rely on that side-effect?

bbrks (changes requested):
This feels really close for the sync function cleanup task - but I still have questions around some of the test fixes being done.

I think we are papering over broken code with some of these `BodyBytes` changes being inside generic helper methods, rather than fixing where the `doc._rawBody` is being accessed directly in more specific areas of code.

db/crud.go
RIT3shSapata: I think this should stay outside itself, because the new Body is used later by `setOldRevisionJSON`.

RIT3shSapata:
closing this PR for now, will work on this later.

PR #8154, opened 2026-04-07
CBG-5263: fix TestNewlyCreateSGWPermissions

gregns1:
CBG-5263

Underlying problem will be fixed in different ticket. This gets the test to pass. 

## Pre-review checklist

- [ ] Link upstream PRs

- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/458/ (running test in question 100 times) 
- [ ] https://jenkins.sgwdev.com/job/SyncGatewayIntegration/457/ (rest package run)

### DIFF
--- a/rest/handler.go
+++ b/rest/handler.go
@@ -497,11 +497,10 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 			// When the lock is returned we know that the db state will not be changed by
 			// any other call
 
-			// defer releasing the dbContext until after the handler method returns
-			dbContext.DbStateLock.RLock()
-			defer dbContext.DbStateLock.RUnlock()
-			if dbContext.IsClosed() {
-				return base.HTTPErrorf(http.StatusServiceUnavailable, "DB is not available - try again later")
+			// defer releasing the dbContext until after the handler method returns, unless it's a blipsync request
+			if !h.isBlipSync() {
+				dbContext.AccessLock.RLock()
+				defer dbContext.AccessLock.RUnlock()
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


rest/handler.go
```diff
@@ -501,6 +501,14 @@ func (h *handler) validateAndWriteHeaders(method handlerMethod, accessPermission
 			if !h.isBlipSync() {
 				dbContext.AccessLock.RLock()
 				defer dbContext.AccessLock.RUnlock()
+
+				// Check if the dbContext was closed while we were waiting to acquire AccessLock.
+				// Another goroutine may have held the write lock and closed the DB before we got here.
+				// Re-fetching and re-locking would introduce a new race, so we return 503
+				// instead — signalling the client to retry once the DB is available again.
+				if dbContext.IsClosed() {
+					return base.HTTPErrorf(http.StatusServiceUnavailable, "DB is not available - try again later")
```
torcolvin: I think you want to do this even if it is a blipsync request because the same problem would exist. In fact, Do we also need to do this if the handler runs offline?

Won't the database go from offline -> close -> online when you take the database online?

This also seems like it would fix the issue in validateAndWriteHeaders, but couldn't the dbContext be nil once this function exists and invoke is called?

I'm not sure how to reconcile this but it still doesn't seem right.

gregns1: This is only needed for the block of code inside shouldCheckAdminAuth if statement. Otherwise the db state handling code line 644 will handle this issue.

torcolvin:
I think this got worse in two ways:

I moved the code that checks for db state so that it wouldn't panic here:

https://github.com/couchbase/sync_gateway/blob/a120bcc398d72cb0d29cb8f45e7eeb540fe72126/rest/handler.go#L633

The goal is to:

1. Don't leak information about a database's existence if we don't know, so if the database is going offline/online we return the same thing as if the database is not found / or doesn't have RBAC credentials

Maybe the AccessLock code is OK because it can't be modified again until the AccessLock is reacquired. That said, we don't acquire an access lock for PUT /db/_config so I'm not sure why that is different.

I will give this more thought later today.

torcolvin (approved):

PR #8159, opened 2026-04-09
CBG-5222: implement MetadataStore type

gregns1:
CBG-5222

New MetadataStore type that implements the DataStore interface. Relevant testing for all reads going primary then fallback to fallback datastore when error is not found error **AND** migration is not complete. All writes go primary only. 

## Pre-review checklist

- [ ] Link upstream PRs

- [x]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/455/ (known flake)

### DIFF
--- a/base/dual_metadata_store.go
+++ b/base/dual_metadata_store.go
@@ -35,6 +35,21 @@ func NewMetadataStore(primary, fallback DataStore) *MetadataStore {
 	}
 }
 
+// Primary will return the primary data store for dual metadata store type (_system._mobile)
+func (ms *MetadataStore) Primary() DataStore {
+	return ms.primary
+}
+
+// Fallback will return the fallback data store for dual metadata store type (_default._default)
+func (ms *MetadataStore) Fallback() DataStore {
+	return ms.fallback
+}
+
+// SetMigrationComplete will set migration complete to true to stop reads falling back to fallback datastore.
+func (ms *MetadataStore) SetMigrationComplete() {
+	ms.migrationComplete.Store(true)
+}
+
 // readFromFallback returns true when err indicates the key was not found in primary and metadata migration has
 // not yet complete, meaning the operation should be retried against the fallback DataStore.
 func (ms *MetadataStore) readFromFallback(err error) bool {
--- a/base/dual_metadata_store.go
+++ b/base/dual_metadata_store.go
@@ -121,7 +121,7 @@ func (ms *MetadataStore) GetExpiry(ctx context.Context, k string) (expiry uint32
 
 func (ms *MetadataStore) Exists(k string) (exists bool, err error) {
 	exists, err = ms.primary.Exists(k)
-	if err != nil && !ms.readFromFallback(nil, err) {
+	if err != nil && !ms.readFromFallback(context.TODO(), err) {
 		return false, err
 	}
 	if exists || ms.migrationComplete.Load() {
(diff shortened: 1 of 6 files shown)


base/dual_metadata_store.go
```diff
@@ -0,0 +1,225 @@
...
+		primary:  primary,
+		fallback: fallback,
+	}
+}
+
+// readFromFallback returns true when err indicates the key was not found in primary and metadata migration has
+// not yet complete, meaning the operation should be retried against the fallback DataStore.
+func (ms *MetadataStore) readFromFallback(err error) bool {
+	return IsDocNotFoundError(err) && !ms.migrationComplete.Load()
+}
```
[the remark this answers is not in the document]
gregns1: SetMigrationComplete added but not needed yet.

base/dual_metadata_store.go
```diff
@@ -0,0 +1,249 @@
...
+
+func (ms *MetadataStore) Exists(k string) (exists bool, err error) {
+	exists, err = ms.primary.Exists(k)
+	if err == nil && exists {
+		return true, nil
+	}
+	if ms.readFromFallback(err) || (!exists && err == nil && !ms.migrationComplete.Load()) {
+		return ms.fallback.Exists(k)
+	}
+	return exists, err
```
bbrks: I'd rewrite this as something like this, to give you idiomatic early-return errors and lose the complex if condition. 

```suggestion
	exists, err = ms.primary.Exists(k)
	if err != nil && !ms.readFromFallback(err) {
		return false, err
	}
	if exists || ms.migrationComplete.Load() {
		return exists, err
	}
	return ms.fallback.Exists(k)
```

base/dual_metadata_store.go
```diff
@@ -0,0 +1,249 @@
...
+	return ms.primary.Delete(k)
+}
+
+func (ms *MetadataStore) Remove(k string, cas uint64) (casOut uint64, err error) {
+	return ms.primary.Remove(k, cas)
+}
+
+func (ms *MetadataStore) Update(k string, exp uint32, callback sgbucket.UpdateFunc) (casOut uint64, err error) {
+	return ms.primary.Update(k, exp, callback)
+}
```
bbrks, line 174: There's a bit of a weird edge-case with `Update` that I've just thought about, which may need more complex implementation to guard against. Applies to other CAS-Safe write operations too where the Get got a value from the fallback.

If callback in Update is reading from the fallback data store (and finds an existing doc to update), the CAS for the write isn't going to be correct and we'll just continuously get a CAS-fail loop.

Can you write a test that runs Update on 2 datastores, where there's no data to update in primary, but there is in fallback? I think that would be enough to trigger this. The existing Update test has no data in either datastore

base/dual_metadata_store.go
```diff
@@ -0,0 +1,249 @@
...
+
+// MigrationComplete reports whether metadata migration has finished.
+func (ms *MetadataStore) MigrationComplete() bool {
+	return ms.migrationComplete.Load()
+}
+
+// readFromFallback returns true when err indicates the key was not found in primary and metadata migration has
+// not yet complete, meaning the operation should be retried against the fallback DataStore.
+func (ms *MetadataStore) readFromFallback(err error) bool {
+	return IsDocNotFoundError(err) && !ms.migrationComplete.Load()
```
bbrks: Don't know if it makes sense to log in here at Debug/Trace level that we're falling back, or whether to do that in each bucket method? I'm open to either.

I'd prefer it to be debug so we can at least track this in real deployments at least until the feature has had a bit of time to settle.

bbrks (changes requested):
This mostly looks good - a few inline comments

bbrks (approved):

PR #8162, opened 2026-04-10
CBG-5265: Generate Grafana dashboards from stats-definition-exporter

bbrks:
CBG-5265

Allows the `stats-definition-exporter` tool to output Grafana dashboards based on actual available SG codebase stats.

- Dashboards group stats by their 'subsystem' as defined by SG code.
- Text descriptions are available in each stat pane.
- Filtering is possible on node, database and collections.
- Annotations visible for SG node restarts.

Output JSON requires adding a `"version": x` field in the JSON to allow Grafana to do a CAS-style update of the dashboard, but otherwise is good to paste straight into Grafana's Dashboard JSON editor. Verified on Capella Grafana and Supportal Grafana.

```shell
$ go build && ./stats-definition-exporter -format supportal-grafana -no-file | pbcopy
```

- n/a


torcolvin:
Is it possible to show us a preview of these dashboards? Send links offline.

torcolvin:
Suggestions, take or leave. I love the look of these! 

- Add to the description the version of Sync Gateway the version the stat started to appear in, sometimes I wonder why the stat is empty.
- revision_cache_total_memory description should probably be marked as bytes and it would be great if the scale was bytes.
- database_config_bucket_mismatches is bytes scale, not integer
- database_config_collection_mismatches is bytes scale, not integer
- xattr_format_mistmatches is bytes scale, not integer
- dcp_received_time doesn't have a y axis
- Be able to do something with the very long units, I would love to have 1 million be represented as 1 000 000 or 1,000,000 
- num_pull_active_repl_active_one_shot shows fractional in the y axis but this might be because it is 0->1. There might be no way to mark that it is integer rather than float
- pull_rev_send_latency has no y axis
- label "replication_push" and "replication_pull" as ISGR
- I forgot which one is right but we should keep either utilization_node_cpu_percent or utilization_process_node_cpu_utilization , whichever one is not closer to the `load` from top.
- resync_num_changed / resync_num_processed is bytes scale, not integer
- Some of the bucket stats are duplicated by collection stats, is that worth calling on in the description? I always get confused and look at the wrong import stats. Or we could just drop the duplicated bucket stats from grafana entirely, they basically exist for backward compatibility. I can't think of why we would need a separate graph for these.

bbrks:
> pull_rev_send_latency has no y axis

This is a Grafana bug when the value overflows 2^53 - can't fix this

bbrks:
I refreshed the Grafana dashboards with latest version of this branch.

I have largely refactored this to use Grafana's SDK/Builder API instead of hand-crafting JSON and the required types. Sorry if you reviewed the old implementation in depth. I think this is a much better approach though. The outputs are nearly identical.

bbrks:
I actually need to revert some build/module changes.

The build script relies on the existing behaviour:

https://github.com/couchbase/build/blob/847121d4f4cbf50a639b428ae12ceef54cc80769/scripts/jenkins/mobile/sgw_unix_build.sh#L92-L106

torcolvin:
Screenshot 2026-04-16 at 1 42 22 PM

After I asked for shortened units, one thing I notice is if you have shortened units on the y axis, then the tooltip also has shortened units. Is this possible to manipulate? This is true for bytes and time (minutes).

Otherwise, LGTM I'm very excited about these.

bbrks:
I can't decouple the axis and tooltip values/formatting.

Choices are:
- As is now: Decimals(0) for integer types (bytes) - but they get very poorly displayed in the UI like above
- Decimals(1-2) and deal with seeing usless decimal places for low integer values - this is probably the more rare case? Don't think we have many byte or nanosecond values that we expect to remain low.

tools/stats-definition-exporter/grafana.go
```diff
@@ -0,0 +1,283 @@
...
+		WithTarget(prometheus.NewDataqueryBuilder().
+			RefId("A").
+			Expr(buildExpr(metricName, stat, config)).
+			LegendFormat(legendForLabels(stat.Labels, config)).
+			EditorMode(prometheus.QueryEditorModeCode).
+			Range())
+
+	// Force integer y-axis ticks for stats whose underlying values are whole numbers.
+	if stat.Format == "int" || stat.Format == "uint64" {
+		b = b.Decimals(0)
```
bbrks: Even though we have integer stats and decimals don't make sense, often these are bytes or nanoseconds, in which we'd like to see the additional precision for when converted to GB, Seconds, etc.

```suggestion
		b = b.Decimals(2)
```

bbrks:
Refreshed dashboards again

Screenshot 2026-04-17 at 15 00 10

good now?

torcolvin (approved):

torcolvin:
Do you want to add instructions somewhere for how to actually get these onto our grafana instances? Doesn't need to be in this repo, could be in confluence or google docs.

PR #8167, opened 2026-04-13
CBG-5285: flake fix for TestActiveReplicatorPushAndPullConflict

gregns1:
CBG-5285

Remove assertions on stats to assert that both push and pull are complete. The issue is there is no gurentee on these stats. For example, 
- Setup push and pull replicator
-  pull starts up before push does and pulls remotes version
-  Solves for remote and writes remotes version as its current version. 
- Then push starts up after this has happened
- Sends its newly resolved confoict version lcoally 
- Remote rejects chnage as it alrwady has it 

So in this case `DocWriteConflict` doesn't get incremented as the push replication doesn't push the conflict before it gets resolved on active. Its a race condition and these assertions aren't always going to be true for it, i've removed them as really assertions below this assert the conflict is correctly resolved. I don't think these assertions are needed. 

## Pre-review checklist

- [ ] Link upstream PRs

- [x]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/469/

### DIFF
--- a/rest/replicatortest/replicator_test.go
+++ b/rest/replicatortest/replicator_test.go
@@ -4261,20 +4261,6 @@ func TestActiveReplicatorPushAndPullConflict(t *testing.T) {
 				require.NoError(t, ar.Start(ctx1))
 				t.Logf("Replicator started")
 
-				// wait for both push and pull to complete:
-				// - the document originally written to rt2 to arrive at rt1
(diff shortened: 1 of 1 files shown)


torcolvin:
If the push and pull might occur in different orders, is the conflict resolution always going to be the same? I think so, but I just want you to verify.

torcolvin (approved):

PR #8169, opened 2026-04-14
CBG-5289: nil check on map niled by closure of server context

gregns1:
CBG-5289

- Protect against panic for online endpoint when server context is closed underneath it. 
- This is a test only issue. 

## Pre-review checklist

- [x]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/472/

### DIFF
--- a/rest/admin_api.go
+++ b/rest/admin_api.go
@@ -265,6 +265,16 @@ func (h *handler) handleDbOnline() error {
 			}
 		}()
 		time.Sleep(time.Duration(input.Delay) * time.Second)
+		// If the server was closed during the delay, skip the reload entirely.
+		// The nil-guard in _getOrAddDatabaseFromConfig is the definitive safety net,
+		// but this early check avoids acquiring locks and doing expensive work
+		// (bucket connection, index setup) against a torn-down ServerContext.
+		select {
+		case <-h.server.serverCtx.Done():
+			base.InfofCtx(ctx.Ctx, base.KeyCRUD, "Database %v online cancelled: server context closed", base.MD(h.db.Name))
+			return
+		default:
+		}
 		h.db.AccessLock.Lock()
 		defer h.db.AccessLock.Unlock()
 		if !atomic.CompareAndSwapUint32(&h.db.State, db.DBOffline, db.DBStarting) {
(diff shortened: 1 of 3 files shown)


bbrks (dismissed):
I'm OK with this approach as a defensive measure, but it would be nice if the `handleDbOnline` goroutine received a cancellation from `Stop()` such that it prevented doing anything after the initial sleep (or even exiting early if it got the stop signal before the sleep fired - you can do a time.After instead of Sleep to push it down into a select to get this)

Probably requires adding a context.Context and a cancel function in `ServerContext` to do this properly?

rest/admin_api.go
```diff
@@ -265,6 +265,16 @@ func (h *handler) handleDbOnline() error {
...
+		// If the server was closed during the delay, skip the reload entirely.
+		// The nil-guard in _getOrAddDatabaseFromConfig is the definitive safety net,
+		// but this early check avoids acquiring locks and doing expensive work
+		// (bucket connection, index setup) against a torn-down ServerContext.
+		select {
+		case <-h.server.serverCtx.Done():
+			base.InfofCtx(ctx.Ctx, base.KeyCRUD, "Database %v online cancelled: server context closed", base.MD(h.db.Name))
+			return
+		default:
+		}
```
bbrks: The sleep is going to be delaying the close/cancel. Would prefer if these were in the same select condition to allow early-exit. [Since Go 1.23 we don't need to release an inlined time.After timer either](https://pkg.go.dev/time#After)

```suggestion
		// If the server was closed during the delay, skip the reload entirely.
		// The nil-guard in _getOrAddDatabaseFromConfig is the definitive safety net,
		// but this early check avoids acquiring locks and doing expensive work
		// (bucket connection, index setup) against a torn-down ServerContext.
		select {
		case <-h.server.serverCtx.Done():
			base.InfofCtx(ctx.Ctx, base.KeyCRUD, "Database %v online cancelled: server context closed", base.MD(h.db.Name))
			return
		case <-time.After(time.Duration(input.Delay) * time.Second):
			// continue
		}
```

bbrks (changes requested):

bbrks (approved):

PR #8173, opened 2026-04-16
CBG-5191: Add distributed test simulation for nodepoller

RIT3shSapata:
CBG-5191

Describe your PR here...
- add tests to simulate distributed node polling

## Pre-review checklist

- [ ] Link upstream PRs

- [x]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/479/
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/502/

### DIFF
--- a/base/dcp_sharded.go
+++ b/base/dcp_sharded.go
@@ -660,7 +660,7 @@ func (p *cfgNodePoller) poll() {
 	watcher := p.getWatcher()
 
 	for key, oldCas := range watcher {
-		newCas, err := p.datastore.Get(key, nil)
+		newCas, err := p.datastore.GetCas(key)
 		if err != nil && !IsDocNotFoundError(err) {
 			WarnfCtx(p.ctx, "cfgNodePoller: error polling doc: %s %v, skipping polling", UD(key), err)
 			continue
--- a/base/dcp_sharded.go
+++ b/base/dcp_sharded.go
@@ -655,12 +655,12 @@ func (p *cfgNodePoller) Register(key string) error {
 // poll checks all registered keys for CAS changes. For each key whose CAS value has changed since the last check,
 // the fireEvent callback is invoked with the key, new CAS value, and nil error. The keyWatcher is updated with
 // the new CAS values. This method is thread-safe and uses optimistic locking to avoid blocking during I/O operations.
-func (p *cfgNodePoller) poll() {
+func (p *cfgNodePoller) poll(ctx context.Context) {
 
 	watcher := p.getWatcher()
 
 	for key, oldCas := range watcher {
-		newCas, err := p.datastore.GetCas(key)
+		_, newCas, err := p.datastore.GetXattrs(ctx, key, []string{"$document.CAS"})
 		if err != nil && !IsDocNotFoundError(err) {
 			WarnfCtx(p.ctx, "cfgNodePoller: error polling doc: %s %v, skipping polling", UD(key), err)
 			continue
@@ -686,7 +686,7 @@ func (p *cfgNodePoller) startPolling(ctx context.Context) {
 				ticker.Stop()
 				return
 			case <-ticker.C:
-				p.poll()
+				p.poll(ctx)
 			}
 		}
 	}()
--- a/base/dcp_sharded.go
+++ b/base/dcp_sharded.go
@@ -660,7 +660,7 @@ func (p *cfgNodePoller) poll(ctx context.Context) {
 	watcher := p.getWatcher()
 
 	for key, oldCas := range watcher {
-		_, newCas, err := p.datastore.GetXattrs(ctx, key, []string{"$document.CAS"})
+		_, newCas, err := p.datastore.GetXattrs(ctx, key, []string{VirtualDocumentXattrCAS})
 		if err != nil && !IsDocNotFoundError(err) {
 			WarnfCtx(p.ctx, "cfgNodePoller: error polling doc: %s %v, skipping polling", UD(key), err)
 			continue
--- a/base/dcp_sharded_test.go
+++ b/base/dcp_sharded_test.go
@@ -1094,7 +1094,6 @@ func TestCfgNodePollerDistributed(t *testing.T) {
 		var errA, errB error
 		var casANew, casBNew uint64
 
-		wg.Add(2)
 		wg.Go(func() {
 			defer wg.Done()
 			casANew, errA = nodeA.Set(key, []byte(`{"source": "nodeA", "ver": 2}`), cas)
--- a/base/dcp_sharded_test.go
+++ b/base/dcp_sharded_test.go
@@ -17,6 +17,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/couchbase/cbgt"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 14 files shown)


base/dcp_sharded_test.go
```diff
@@ -958,3 +960,815 @@ func TestCfgNodePoller_StartPolling(t *testing.T) {
...
+
+	ctx := t.Context()
+	bucket := GetTestBucket(t)
+	defer bucket.Close(ctx)
+	dataStore := bucket.GetSingleDataStore()
+	testPollInterval := 10 * time.Millisecond
+
+	// Helper to create a CfgSG with poller
+	createNode := func(prefix string) (*CfgSG, chan cbgt.CfgEvent) {
+		cfg, err := NewCfgSG(ctx, dataStore, prefix, true, testPollInterval)
```
torcolvin: This needs to be closed in the test environment, you could do `t.Cleanup(cfg.Close())` but then you need to make this a function that isn't inside `TestCfgNodePollerDistributed` because you want this to be scoped to the subtest testing.T functions. That's fine and potentially more readable anyway.

base/dcp_sharded_test.go
```diff
@@ -958,3 +960,815 @@ func TestCfgNodePoller_StartPolling(t *testing.T) {
...
+		require.EventuallyWithT(t, func(t *assert.CollectT) {
+			select {
+			case event := <-eventsB:
+				assert.Equal(t, key, event.Key)
+				assert.Equal(t, cas, event.CAS)
+				assert.NoError(t, event.Error)
+			default:
+				assert.Fail(t, "no event received")
+			}
+		}, 100*time.Millisecond, 10*time.Millisecond)
```
torcolvin: There's a helpful helper function for this `base.RequireChanRecvWithTimeout`:

https://github.com/couchbase/sync_gateway/blob/d6c31d5d8fc291738ca188a1e3292fe2fad226fb/base/util_testing.go#L1027

base/dcp_sharded_test.go
```diff
@@ -958,3 +960,815 @@ func TestCfgNodePoller_StartPolling(t *testing.T) {
...
+
+		// Bot nodes try to update the document with the same CAS
```
torcolvin:
```suggestion
		// Both nodes try to update the document with the same CAS
```

base/dcp_sharded_test.go
```diff
@@ -958,3 +960,815 @@ func TestCfgNodePoller_StartPolling(t *testing.T) {
...
+
+		wg.Add(2)
+		go func() {
+			defer wg.Done()
+			casANew, errA = nodeA.Set(key, []byte(`{"source": "nodeA", "ver": 2}`), cas)
+		}()
+		go func() {
+			defer wg.Done()
+			casBNew, errB = nodeB.Set(key, []byte(`{"source": "nodeB", "ver": 2}`), cas)
+		}()
```
torcolvin: nit: prefer `wg.Do` API when spawning goroutines to avoid API mistakes:

```suggestion
		wg.Do(func() {
			casANew, errA = nodeA.Set(key, []byte(`{"source": "nodeA", "ver": 2}`), cas)
		})
		wg.Do(func() {
			casBNew, errB = nodeB.Set(key, []byte(`{"source": "nodeB", "ver": 2}`), cas)
		})
```

base/dcp_sharded_test.go
```diff
@@ -958,3 +960,815 @@ func TestCfgNodePoller_StartPolling(t *testing.T) {
...
+		wg.Wait()
+
+		var successCas uint64
+		var failedNodeEvents chan cbgt.CfgEvent
+		// Exactly one update should succeed and the other should fail with CAS Error
+		require.True(t, (errA == nil && errB != nil) || (errA != nil && errB == nil))
+		if errA != nil {
+			require.ErrorIs(t, errA, ErrCfgCasError)
```
torcolvin, line 1113: Is this always `ErrCfgCasError with an without rosmar?

base/dcp_sharded_test.go
```diff
@@ -958,3 +960,815 @@ func TestCfgNodePoller_StartPolling(t *testing.T) {
...
+			default:
+				assert.Fail(t, "losing node should detect winner's update")
+			}
+		}, 100*time.Millisecond, 10*time.Millisecond)
+	})
+
+	t.Run("delete_detection_across_nodes", func(t *testing.T) {
+		if UnitTestUrlIsWalrus() {
+			t.Skip("This test requires Couchbase Server - deletion behavior differs in Rosmar")
+		}
```
torcolvin: What's the differing behavior with rosmar? This feels like either rosmar is detecting a race you aren't seeing with CBS, or potentially we need to handle different error conditions.

There's no reason that this heartbeater should have couchbase server dependency.

base/dcp_sharded_test.go
```diff
@@ -958,3 +960,815 @@ func TestCfgNodePoller_StartPolling(t *testing.T) {
...
+			case event := <-eventsA:
+				assert.Equal(collectT, key, event.Key)
+				assert.Equal(collectT, uint64(0), event.CAS, "Node A: CAS should be 0 for deletion")
+				assert.NoError(collectT, event.Error)
+			default:
+				assert.Fail(collectT, "Node A should detect its own deletion")
+			}
+		}, 100*time.Millisecond, 10*time.Millisecond)
+
+		// Node B should detect deletion via polling (Couchbase Server: tombstone has CAS=0)
```
torcolvin: What does rosmar do here?

A tombstone shouldn't have zero cas unless it is purged, that might mean that the Get call we need to make has to be tombstone aware.

It's definitely possible we should create a function that is does `GetCas` on datastore https://github.com/couchbase/sg-bucket/blob/190459c31cd8ba61afbdfe6c6e4ceb9d965af7da/bucket.go#L273 to make sure this works on a tombstone or no tombstone.

But this seems like this isn't right here.

db/database.go
```diff
@@ -509,7 +509,7 @@ func NewDatabaseContext(ctx context.Context, dbName string, bucket base.Bucket,
 	// Initialize sg cluster config.  Required even if import and sgreplicate are disabled
 	// on this node, to support replication REST API calls
 	if base.IsEnterpriseEdition() {
-		sgCfg, err := base.NewCfgSG(ctx, metadataStore, metaKeys.SGCfgPrefix(dbContext.Options.GroupID), false)
+		sgCfg, err := base.NewCfgSG(ctx, metadataStore, metaKeys.SGCfgPrefix(dbContext.Options.GroupID), false, 0)
```
torcolvin: I'd pass base.DefaultHeartbeatPoolInterval here, or even better is basically to create a function that allows custom variable for tests, but then do not require this of the callers in the db package::

```
func newCfgSG(ctx context.Context, datastore sgbucket.DataStore, keyPrefix string, useNodePoller bool, pollInterval time.Duration) (*CfgSG, error) {
```

and then have the public function just call the private one:
```
func NewCfgSG(ctx context.Context, datastore sgbucket.DataStore, keyPrefix string, useNodePoller bool) (*CfgSG, error) {
	return newCfgSG(ctx, datastore, keyPrefix, useNodePoller, DefaultHeartbeatPollInterval)
```

base/dcp_sharded_test.go
RIT3shSapata: There is no explicit `Close` method for `CfgSG` or `nodepoller` so I'm going to pass contexts, that will cancel at the end of every test case

base/dcp_sharded_test.go
RIT3shSapata: The `Set` function always returns `ErrCfgCasError` https://github.com/couchbase/sync_gateway/blob/6922536a5752b5736482d8396fd6207a7f2d95fd/base/sg_cluster_cfg.go#L97-L116

base/dcp_sharded_test.go
torcolvin: If that's the case then we need to cancel the context

base/dcp_sharded.go
```diff
@@ -660,7 +660,7 @@ func (p *cfgNodePoller) poll() {
 	watcher := p.getWatcher()
 
 	for key, oldCas := range watcher {
-		newCas, err := p.datastore.Get(key, nil)
+		newCas, err := p.datastore.GetCas(key)
```
torcolvin: Can we do p.datastore.GetXattrs(ctx, key, "$document.CAS") for this? you'd need to add support for this in rosmar but this seems like it would make sense it me.

base/sg_cluster_cfg.go
```diff
@@ -58,12 +54,31 @@ func NewCfgSG(ctx context.Context, datastore sgbucket.DataStore, keyPrefix strin
...
+// If useNodePoller is true, then any document changes are received by node polling with the specified pollInterval.
+// If useNodePoller is not true, then the caller needs to register event changes itself by calling FireEvent.
+//
+// The caching feed implements FireEvent calls by looking for document changes starting with keyPrefix and calling FireEvent.
+func NewCfgSG(ctx context.Context, datastore sgbucket.DataStore, keyPrefix string, useNodePoller bool) (*CfgSG, error) {
+	pollInterval := DefaultHeartbeatPollInterval
+	if !useNodePoller {
+		pollInterval = 0
+	}
+	return newCfgSG(ctx, datastore, keyPrefix, useNodePoller, pollInterval)
```
torcolvin:
```suggestion
	return newCfgSG(ctx, datastore, keyPrefix, useNodePoller, DefaultHeartbeatPollInterval)
```

torcolvin (dismissed):

torcolvin (approved):

PR #8178, opened 2026-04-18
CBG-5298 add test for repro

torcolvin:
CBG-5298 add test for repro

These tests will fail as expected, there is a "simple" test in the auth package and a more complicated test in OIDC pcakage.


bbrks:
Thanks for this! Going to roll these up along with the eventual fixes since this isn't mergeable as-is

PR #8179, opened 2026-04-20
CBG-4665: Only put an ActiveReplicator into an error state when the connect error is a fatal one

bbrks:
CBG-4665

Fixes a prominent flake in `TestReplicatorReconnectTimeout` where the `NumReconnectsAborted` stat was incremented prior to the replicator state changing.

- Fix attempt 1: Reorder stat/error state change so error state is always before the stat change.
- Fix attempt 2: Instead of reordering the stat or state change, only put the replicator into an error state when it's a fatal error on startup. This avoids a transient 'error' state which the test was getting into before the aborted stat had a chance to be set and the 2nd error state set.

- [ ] https://jenkins.sgwdev.com/job/SyncGatewayIntegration/498/

### DIFF
--- a/db/active_replicator_common.go
+++ b/db/active_replicator_common.go
@@ -337,13 +337,13 @@ func (arc *activeReplicatorCommon) synchronousReconnect() {
 	}
 
 	base.WarnfCtx(ctx, "aborting reconnect loop: %v", retryErr)
+	arc.replicationStats.NumReconnectsAborted.Add(1)
 	arc.lock.Lock()
+	defer arc.lock.Unlock()
 	// use setState to preserve last error from retry loop set by setLastError
 	arc.setState(ReplicationStateError)
 	arc._publishStatus()
 	arc._stop("stopping after failed reconnect attempts: " + retryErr.Error())
-	arc.lock.Unlock()
-	arc.replicationStats.NumReconnectsAborted.Add(1)
 }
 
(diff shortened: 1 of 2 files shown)


torcolvin:
I'm not opposed to this change but the root of this problem is interesting:

I believe that it comes from https://github.com/couchbase/sync_gateway/blob/bad45fe8be825a2bad808e688f053efdf5746de4/db/active_replicator_common.go#L161

If the first connection fails, it will set the error state as part of `setError`. I think `setError` should only be called in the case of when this is fatal, and not when you are about to reconnect.

I don't know how you feel about the state of "connecting" and an error message from `setLastError`,  I think that seems reasonable to me.

bbrks:
> I'm not opposed to this change but the root of this problem is interesting:
> 
> I believe that it comes from
> 
> https://github.com/couchbase/sync_gateway/blob/bad45fe8be825a2bad808e688f053efdf5746de4/db/active_replicator_common.go#L161
> 
> If the first connection fails, it will set the error state as part of `setError`. I think `setError` should only be called in the case of when this is fatal, and not when you are about to reconnect.
> 
> I don't know how you feel about the state of "connecting" and an error message from `setLastError`, I think that seems reasonable to me.

I actually agree with this assessment/approach - working on that now.

torcolvin (approved):

PR #8182, opened 2026-04-20
RedactableError support for errors.Is/As/AsType

torcolvin:
RedactableError support for errors.Is/As/AsType

## Pre-review checklist

- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/493/

### DIFF
--- a/base/redactable_error.go
+++ b/base/redactable_error.go
@@ -57,7 +57,11 @@ func (re *RedactableError) Redact() string {
 	return fmt.Errorf(re.fmt, redactedArgs...).Error()
 }
 
+// Is returns true if any error in the chain matches the target error.
 func (re *RedactableError) Is(target error) bool {
+	if errors.Is(re, target) {
+		return true
+	}
 	for _, arg := range re.args {
 		err, ok := arg.(error)
 		if !ok {
@@ -70,7 +74,11 @@ func (re *RedactableError) Is(target error) bool {
 	return false
 }
 
+// As returns true if any error in the chain matches the target type and sets target as the value.
 func (re *RedactableError) As(target any) bool {
+	if errors.As(re, target) {
+		return true
+	}
 	for _, arg := range re.args {
 		err, ok := arg.(error)
 		if !ok {
--- a/base/redactable_error.go
+++ b/base/redactable_error.go
@@ -10,7 +10,10 @@ licenses/APL2.txt.
 
 package base
 
-import "fmt"
+import (
+	"errors"
+	"fmt"
+)
 
 // RedactableError is an error that can be used as a drop-in replacement for an error,
 // which has the ability to redact any sensitive data by calling redact() on all of its args.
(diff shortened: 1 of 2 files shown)


base/redactable_error.go
```diff
@@ -53,3 +53,8 @@ func (re *RedactableError) Redact() string {
 	// can't use Sprintf as it doesn't support `%w`
 	return fmt.Errorf(re.fmt, redactedArgs...).Error()
 }
+
+// Unwrap returns the underlying error(s) created by fmt.Errorf. Allows use of errors.As, errors.AsType, errors.Is
+func (re *RedactableError) Unwrap() error {
+	return fmt.Errorf(re.fmt, re.args...)
```
bbrks: I'm not against this but it does cause all of the redaction/type information to be lost, which seems more prone to accidental use (e.g. returning an unwrapped error to reduce noise/context from a known error-handling codepath after an errors.Is/As)

I think if we want to do this, we should do it in a way that allows Unwrap without building a new error.

bbrks (changes requested):

gregns1 (dismissed):

base/redactable_error_test.go
```diff
@@ -46,6 +48,77 @@ func TestRedactErrorf(t *testing.T) {
...
+		err2 := &customErr{msg: "err2"}
+		re2 := RedactErrorf("err1: %w, err2: %w", err1, err2)
+
+		assert.True(t, errors.Is(re2, err1))
+		assert.True(t, errors.Is(re2, err2))
+
+		var target *customErr
+		assert.True(t, errors.As(re2, &target))
+		assert.Equal(t, err1, target) // errors.As returns the first one it finds
+	})
```
bbrks, line 112: Looks pretty good. Maybe just one test case to cover the wrapped redactable error handling?

```go
  t.Run("Redactor interface survives Unwrap", func(t *testing.T) {
      inner := RedactErrorf("user %s", UD("Bob"))
      outer := RedactErrorf("wrapped: %w", inner)

      // Outer redacts correctly
      assert.Contains(t, outer.Redact(), "user <ud>Bob</ud>")

      // errors.As extracts inner, still satisfies Redactor
      var target *RedactableError
      require.True(t, errors.As(outer, &target))
      require.NotSame(t, outer, target)

      // Unwrapped inner still redacts
      var r Redactor = target
      assert.Contains(t, r.Redact(), "user <ud>Bob</ud>")
      assert.NotContains(t, r.Redact(), "user Bob")
  })
```

bbrks (changes requested):
I'm OK with this last iteration of Unwrap - we should just get some test coverage of redacting the pre and post-unwrap stages

base/redactable_error_test.go
torcolvin: I added this test and one like it but notably 

```
      var target *RedactableError
      require.True(t, errors.As(outer, &target))
      require.NotSame(t, outer, target)
```

Actually would fail because errors.As here gets the outer error.

bbrks (approved):

PR #8187, opened 2026-04-22
CBG-5182: Add Database state management with polling

RIT3shSapata:
CBG-5182

Describe your PR here...
- Added a Database state manager, that performs the following:
    - Creates a database state document
    - Polls it periodically and compares the CAS to check for difference
    - When a change is detected, notifies the ResyncHandler
- Added test coverage for all the database state manager
- Integrated it with `DatabaseContext` and is initialised when the database is started in offline mode

image

## Pre-review checklist

- [ ] Link upstream PRs

- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/504/ (failed)
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/505/ (failed)
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/511/ (failed - flaky test)
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/523/ (failed - couchbase connection)
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/535/
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/543/
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/561/
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/593/
- [x]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/610/
- [x]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/620/
- [x]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/634/


rest/server_context.go
```diff
@@ -1075,6 +1075,7 @@ func (sc *ServerContext) _getOrAddDatabaseFromConfig(ctx context.Context, config
 	}
 
 	if !startOnlineProcesses {
+		dbcontext.DBStateMgr.StartPolling(dbcontext.CancelContext)
```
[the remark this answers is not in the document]
RIT3shSapata: @torcolvin curious to know if you want me to start the polling only if offline mode + needs resync or just offline mode is enough

db/database_state.go
```diff
@@ -0,0 +1,142 @@
...
+		metadataStore:   metadataStore,
+		pollingInterval: 10 * time.Second,
+	}
+}
+
+// UpdateState persists the given DatabaseState to the metadata store using a CAS write, then updates the
+// in-memory State and CAS on success. Returns an error on CAS mismatch or store failure.
+func (dbMgr *DatabaseStateMgr) UpdateState(state DatabaseState) (err error) {
+	dbMgr.lock.Lock()
+	defer dbMgr.lock.Unlock()
```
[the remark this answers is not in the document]
torcolvin: This is definitely correct, there are a few ways of thinking about addressing it:

1. Do a cas unsafe write always, assume last one who wins is OK.
2. Design this API to only update a subdoc and make _just_ the subdoc operation fail.
3. Use `Datastore.Update` function which handles cas safety for us. I think this is a approach we want to do, which is basically going to be you have a new state and it is always going to update with Last One Wins, but doing this means we have a bit of an opportunity to do some intelligent updating.

db/database_state.go
```diff
@@ -0,0 +1,142 @@
...
+	cas, err := dbMgr.metadataStore.WriteCas(dbMgr.dbStateID, 0, dbMgr.CAS, state, 0)
+	if err != nil {
+		return err
+	}
+	dbMgr.State = state
```
torcolvin: I don't think we actually need to store the state at all, I would only store the cas. If we need to re-read it, it is fine to do this.

db/database_state.go
```diff
@@ -0,0 +1,142 @@
...
+	dbMgr.CAS = cas
+	return
+}
+
+// GetState reads the current DatabaseState document from the metadata store. Returns the state and its CAS value.
+// A doc-not-found error is treated as a zero-value state.
+func (dbMgr *DatabaseStateMgr) GetState() (state DatabaseState, cas uint64, err error) {
```
torcolvin: The polling part of GetState might return CAS but if there's a reason that a non polling caller would need cas, then they don't need this.

Also, this API should use `*DatabaseState` in the return value and return nil. Otherwise you have to worry about an empty vs nil problem.

db/database_state.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	return
+}
+
+// DeleteState removes the database state document from the metadata store using a CAS-protected Remove.
+// Returns an error on CAS mismatch or if the document does not exist.
+func (dbMgr *DatabaseStateMgr) DeleteState() (err error) {
+	dbMgr.lock.Lock()
+	defer dbMgr.lock.Unlock()
+	_, err = dbMgr.metadataStore.Remove(dbMgr.dbStateID, dbMgr.CAS)
```
torcolvin: You can use dataStore.Update to delete a document in a to avoid these races

db/database_state.go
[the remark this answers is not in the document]
RIT3shSapata: changed it to use `datastore.Update`

db/database_state.go
```diff
@@ -0,0 +1,144 @@
...
+	dbMgr.lock.Lock()
+	defer dbMgr.lock.Unlock()
+	state, cas, err := dbMgr.GetState()
+	if err != nil {
+		if base.IsDocNotFoundError(err) && cas != dbMgr.CAS {
+			dbMgr.CAS = cas
+			dbMgr.resyncHandler(false)
+			return
+		}
+		base.WarnfCtx(ctx, "error while polling for offline database state: %v", err)
```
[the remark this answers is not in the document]
torcolvin: I think this comment is correct but there might be a more elegant way to resolve this. Add a test case for this.

db/database_state.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	if cas == dbMgr.CAS {
+		return
+	}
+	dbMgr.resyncHandler(state.ResyncRunning)
+	dbMgr.CAS = cas
```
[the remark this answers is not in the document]
torcolvin: I don't think we need to lock while we are calling this function. Do we need a lock at all? What is the circumstance where we want a lock?

db/database_state_test.go
```diff
@@ -0,0 +1,238 @@
...
+	"time"
+
+	"github.com/couchbase/sync_gateway/base"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestUpdateState verifies that UpdateState persists the state document and updates the in-memory State and
+// CAS, and returns an error when the CAS is stale.
+func TestUpdateState(t *testing.T) {
```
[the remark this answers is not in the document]
torcolvin: This comment seems valid, and I'd rename this to be TestDatabaseStateUpdate so the test name tells you what code is running

rest/api.go
```diff
@@ -385,6 +385,10 @@ func (h *handler) handlePostResync() error {
 			if err != nil {
 				return err
 			}
+			err = h.db.DBStateMgr.UpdateState(db.DatabaseState{ResyncRunning: true})
+			if err != nil {
+				base.WarnfCtx(h.ctx(), "Error updating state for 'resyncing' database: %v", err)
+			}
```
torcolvin: We don't need to update this, this would get handled as part of background manager anyway.

rest/api.go
```diff
@@ -417,6 +421,10 @@ func (h *handler) handlePostResync() error {
 		if err != nil {
 			return err
 		}
+		err = h.db.DBStateMgr.DeleteState()
+		if err != nil {
+			return err
+		}
```
torcolvin: We don't need to update this, this would get handled as part of background manager anyway.

rest/server_context.go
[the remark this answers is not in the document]
torcolvin: I'd like to move this code into a function in database called `DatabaseContext.InitializeOfflineMode()` so that this can be tested from within db package. You don't need to worry about the TempResyncHandler code for now, but keeping it within the db package is an easier abstraction. `server_context.go` can use call `dbContext.InitializeOfflineMode()`.

db/database_state.go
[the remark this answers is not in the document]
RIT3shSapata: We need the lock because we are updating the CAS value in this function, I've changed it to make two different lock calls and then immediately lock. I've made sure that the callback function is called and only after which the poll function will hold the lock.

db/database_state.go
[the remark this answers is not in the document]
RIT3shSapata: Yes makes sense, changed this and added a test case for this scenario

rest/api.go
RIT3shSapata: got it, moved this under background manager

db/database.go
```diff
@@ -610,6 +611,9 @@ func NewDatabaseContext(ctx context.Context, dbName string, bucket base.Bucket,
 	dbContext.ResyncManager = NewResyncManagerDCP(metadataStore, dbContext.UseXattrs(), metaKeys)
 	dbContext.AsyncIndexInitManager = NewAsyncIndexInitManager(dbContext.MetadataStore, dbContext.MetadataKeys)
 
+	// Initialize DatabaseStateManager
+	dbContext.DBStateMgr = NewDatabaseStateMgr(metadataStore, metaKeys.DatabaseStateKey())
```
torcolvin: Very nit picky, drop the comment because it is pretty obvious what is going on from the function name (great). I would use `Manager` vs `Mgr` because it matches the names of the other ResyncManager, etc.


```suggestion
	dbContext.DBStateManager = NewDatabaseStateMgr(metadataStore, metaKeys.DatabaseStateKey())
```

db/database_state.go
```diff
@@ -0,0 +1,149 @@
...
+// Returns an error on delete failure
+func (dbMgr *DatabaseStateMgr) DeleteState() (err error) {
+	dbMgr.lock.Lock()
+	defer dbMgr.lock.Unlock()
+	_, err = dbMgr.metadataStore.Update(dbMgr.dbStateID, 0, func(current []byte) (updated []byte, expiry *uint32, delete bool, err error) {
+		return nil, nil, true, nil
+	})
+	if err != nil {
+		return err
+	}
```
torcolvin: If you want to do a cas unsafe removal `metadataStore.Delete` will do this. I wonder if I mislead you earlier.

db/database_state.go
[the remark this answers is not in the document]
torcolvin: If cas is the only value that is locked, I think we should store it as atomic.Pointer[uint64]. Having `Unlock` not in a defer makes me really anxious, and it is the cause of many subtle bugs.

If we need to lock for other things, I think having a function `setCas(uint64)` that just does lock / unlock there is more appropriate.

These types of bugs are very subtle and hard to catch when they happen to customers which is why I'm so picky about it.

db/database_state_test.go
```diff
@@ -0,0 +1,254 @@
...
+		require.NoError(t, mgr.DeleteState())
+		require.Zero(t, mgr.CAS)
+	})
+
+	t.Run("returns no error when doc already deleted", func(t *testing.T) {
+		docID := base.NewMetadataKeys(t.Name()).DatabaseStateKey()
+		mgr := NewDatabaseStateMgr(metadataStore, docID)
+		require.NoError(t, mgr.UpdateState(DatabaseState{ResyncRunning: true}))
+		require.NoError(t, mgr.DeleteState())
+		require.NoError(t, mgr.DeleteState())
```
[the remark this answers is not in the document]
torcolvin: What do you think about this comment?

db/database_state.go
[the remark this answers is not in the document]
RIT3shSapata: changed it to atomic.Uint64

base/util_testing.go
```diff
@@ -1084,3 +1084,32 @@ func WaitWithTimeout(t testing.TB, wg *sync.WaitGroup, timeout time.Duration) {
 		require.FailNow(t, fmt.Sprintf("Timed out waiting after %.2f sec", timeout.Seconds()))
 	}
 }
+
+// CaptureLogOutput captures log output during the execution of a function.
+// It creates a temporary logger with output redirected to a buffer, atomically swaps
+// the global logger for the duration of the function, and returns the captured output.
+// This approach is thread-safe and properly flushes buffered logs before returning.
+func CaptureLogOutput(t testing.TB, fn func()) string {
```
bbrks: We already havesimilar utility functions `AssertLogContains` and `AssertLogNotContains`, but in my experience this type of assertion is quite flaky and prone to race conditions with the global logger being swapped.

I would prefer not to use it if possible, favouring functional testing rather than log output. In the case of `Polling without state being created`, I don't think any specific test coverage is really required.

db/database_state.go
```diff
@@ -0,0 +1,144 @@
...
+// Returns an error on delete failure
+func (dbMgr *DatabaseStateMgr) DeleteState() (err error) {
+	dbMgr.lock.Lock()
+	defer dbMgr.lock.Unlock()
+	err = dbMgr.metadataStore.Delete(dbMgr.dbStateID)
+	if err != nil {
+		return err
+	}
+	dbMgr.CAS.Store(0)
+	return
```
bbrks: avoid naked returns - easier to read

```suggestion
	return nil
```

db/database.go
```diff
@@ -610,6 +611,8 @@ func NewDatabaseContext(ctx context.Context, dbName string, bucket base.Bucket,
 	dbContext.ResyncManager = NewResyncManagerDCP(metadataStore, dbContext.UseXattrs(), metaKeys)
 	dbContext.AsyncIndexInitManager = NewAsyncIndexInitManager(dbContext.MetadataStore, dbContext.MetadataKeys)
```
bbrks: I know this is strictly unrelated to this PR, but worthy of a tiny cleanup whilst we're here for consistency reasons. `dbContext.MetadataStore` comes from the `metadata` variable up at the top of the function.

```suggestion
	dbContext.AsyncIndexInitManager = NewAsyncIndexInitManager(metadataStore, dbContext.MetadataKeys)
```

db/database.go
```diff
@@ -2591,3 +2596,13 @@ func (db *DatabaseContext) usingRosmar() bool {
 func (db *DatabaseContext) WaitForSequenceNotSkipped(ctx context.Context, targetSequence uint64) error {
 	return db.changeCache.waitForSequenceNotSkipped(ctx, targetSequence, defaultWaitForSequence)
 }
+
+// InitializeOfflineMode starts polling the database state when the database transitions to offline mode.
+// This enables the database to detect state changes (such as resync requests) from other nodes in the cluster
+// while it is offline. The polling mechanism watches the metadata store for updates to the database state
+// document and invokes registered handlers when changes are detected.
+func (db *DatabaseContext) InitializeOfflineMode() {
+	// TODO: Add the appropriate handler function to handle this
```
bbrks: Does this TODO need to be done in this PR or in a follow-up ticket? What is that Jira ticket if so?

db/database_state.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// DeleteState removes the database state document from the metadata store using a CAS-protected Remove.
+// Returns an error on delete failure
+func (dbMgr *DatabaseStateMgr) DeleteState() (err error) {
+	dbMgr.lock.Lock()
+	defer dbMgr.lock.Unlock()
+	err = dbMgr.metadataStore.Delete(dbMgr.dbStateID)
+	if err != nil {
```
bbrks, line 85: Maybe only return error here if error is not `DocNotFound`?

I am not sure if multiple SG nodes could be racing to delete the state or not, but it would leave `dbMr.CAS` outdated on all but one node if so. Not sure what the implications are there.

db/database_state.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// GetState reads the current DatabaseState document from the metadata store. Returns the state and its CAS value.
+// A doc-not-found error is treated as a zero-value state.
+func (dbMgr *DatabaseStateMgr) GetState() (state *DatabaseState, cas uint64, err error) {
+	cas, err = dbMgr.metadataStore.Get(dbMgr.dbStateID, &state)
+	if err != nil && !base.IsDocNotFoundError(err) {
+		return nil, cas, err
+	}
+	return
```
bbrks: Here is why naked returns are a little problematic for readability. You have to read the full function and figure out what values are set where and whether that is a valid combination.

It's more obvious this is a success case if this were:

```suggestion
	return state, cas, nil
```

db/database_state.go
```diff
@@ -0,0 +1,144 @@
...
+		if err != nil {
+			return nil, nil, false, err
+		}
+		return bodyBytes, nil, false, nil
+	})
+	if err != nil {
+		return err
+	}
+	dbMgr.CAS.Store(cas)
+	return
```
bbrks:
```suggestion
	return nil
```

db/background_mgr_resync_dcp.go
```diff
@@ -142,8 +142,17 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
...
+		err := db.DBStateManager.DeleteState()
+		if err != nil {
+			base.WarnfCtx(ctx, "failed to delete the database state: %v", err)
+		}
 	}()
 
+	err = db.DBStateManager.UpdateState(DatabaseState{ResyncRunning: true})
+	if err != nil {
+		return fmt.Errorf("error updating state for 'resyncing' database: %w", err)
+	}
```
bbrks: If this errors then the deferred `DeleteState` could also error/warn with a not found. Maybe this is acceptable enough? Just calling it out.

db/database_state.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	return
+}
+
+// DeleteState removes the database state document from the metadata store using a CAS-protected Remove.
+// Returns an error on delete failure
+func (dbMgr *DatabaseStateMgr) DeleteState() (err error) {
+	dbMgr.lock.Lock()
+	defer dbMgr.lock.Unlock()
+	err = dbMgr.metadataStore.Delete(dbMgr.dbStateID)
```
bbrks: I thought the db state document was generic for the database and potentially going to be used for things other than resync in future? That means we can't delete it. We can only ever update values inside of it in a subdoc style fashion.

db/database_state.go
```diff
@@ -0,0 +1,144 @@
...
+// in-memory State and CAS on success. Returns an error store failure.
+func (dbMgr *DatabaseStateMgr) UpdateState(state DatabaseState) (err error) {
+	dbMgr.lock.Lock()
+	defer dbMgr.lock.Unlock()
+	cas, err := dbMgr.metadataStore.Update(dbMgr.dbStateID, 0, func(current []byte) (updated []byte, expiry *uint32, delete bool, err error) {
+		bodyBytes, err := base.JSONMarshal(state)
+		if err != nil {
+			return nil, nil, false, err
+		}
+		return bodyBytes, nil, false, nil
```
bbrks, line 70: Similar comment to delete:
If dbstate doc is used for anything other than resync, this will be overwriting the existing values.

Do we not need to do an get->unmarshal->set fields type update?

bbrks (changes requested):
@torcolvin Please see if my `Delete`/`Update` comments make sense. I think this implementation breaks the assumption that the db state document can be used for anything other than resync.

base/util_testing.go
RIT3shSapata: yes makes sense, will remove it

db/database.go
RIT3shSapata: added the CBG in the comment

db/database_state.go
RIT3shSapata: The race condition is possible but not entirely sure how frequently it will happen. The CAS being outdated in other nodes is the intended behaviour. In the next poll run on a node, the CAS mismatch will be detected and updated.

RIT3shSapata: On second thought, I've added a check for DocNotFoundError, we can discuss this further offline

db/background_mgr_resync_dcp.go
RIT3shSapata: removed DeleteState function

db/database_state.go
RIT3shSapata: updated it

db/database_state.go
RIT3shSapata: removed the deletestate function

db/database.go
```diff
@@ -2510,6 +2515,29 @@ func (db *DatabaseContext) GetCollectionIDs() []uint32 {
...
+	if err != nil {
+		return err
+	}
+
+	datastore := database.MetadataStore
+	metadata := base.NewDCPMetadataCS(ctx, datastore, numVbuckets, base.DefaultNumWorkers, checkpointPrefix)
+	metadata.Purge(ctx, base.DefaultNumWorkers)
+	return nil
+}
+
```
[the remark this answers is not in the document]
bbrks: +1

db/database_state.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	if cas == dbMgr.CAS.Load() {
+		return
+	}
+	dbMgr.resyncHandler(state.ResyncRunning)
+	dbMgr.CAS.Store(cas)
+}
+
+// StopPolling signals the background polling goroutine started by StartPolling to exit.
+func (dbMgr *DatabaseStateMgr) StopPolling() {
```
bbrks: Does this need wiring up to `handleDbOnline` or some other area of code that is related to bring the database back online? Or is this expected to be a follow-up? Or is this being stopped as a side-effect of the existing DatabaseContext being Closed when a database is being brought back online (reloaded?)

I think the idea is we'd only be polling for an offline database - but want to check with you/@torcolvin

db/database_state.go
```diff
@@ -0,0 +1,142 @@
...
+		bodyBytes, err := base.JSONMarshal(state)
+		if err != nil {
+			return nil, nil, false, err
+		}
+		return bodyBytes, nil, false, nil
+	})
+	if err != nil {
+		return err
+	}
+	dbMgr.CAS.Store(cas)
```
bbrks: I am not sure why we are setting CAS here, since it would prevent this node from running `poll()` loop and actually invoking the resync function.

Was the intention that the node triggering resync would've already kicked off the resync function outside of the polling loop?

If so, there's another race condition where it's possible for the poll loop to run using an outdated CAS (when `poll()` runs between line 73 and 76) - which AFAIK can only be fixed by holding the lock inside `poll()`.

Given that there's a requirement for all code to use the lock, I am not sure why this CAS value needs to be atomic. It is hiding a race condition that the race detector should pick up.

db/background_mgr_resync_dcp.go
```diff
@@ -159,8 +159,17 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
 		} else {
 			cancelResync(errors.New("resync ended normally"))
 		}
+		err := db.DBStateManager.UpdateState(DatabaseState{ResyncRunning: base.Ptr(false)})
+		if err != nil {
+			base.WarnfCtx(ctx, "failed to delete the database state: %v", err)
```
bbrks: outdated message

```suggestion
			base.WarnfCtx(ctx, "failed to update the database state: %v", err)
```

db/database_state.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+type DatabaseState struct {
+	ResyncRunning *bool `json:"resync,omitempty"`
+}
+
+type ResyncHandler func(resume *bool)
+
+func TempResyncHandler(resume *bool) {
+	return
```
bbrks, line 26: I know this is temporary until resync is available, but maybe worth considering adding a log message or something else that we can run a test assertion on (e.g. a simple counter so we can "assert ResyncHandler fired 3 times")

db/database_state.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return nil, cas, err
+	}
+	return &state, cas, nil
+}
+
+// AddResyncFunc registers the callback that is invoked by the polling loop when a state change is detected.
+// The callback receives true if a resync should resume, or false if the state document was removed (resync complete).
+func (dbMgr *DatabaseStateMgr) AddResyncFunc(resyncFunc ResyncHandler) {
+	dbMgr.resyncHandler = resyncFunc
```
bbrks: Nit: This is a `Set` operation rather than add (append)

db/database_state.go
```diff
@@ -0,0 +1,142 @@
...
+	"time"
+
+	"github.com/couchbase/sync_gateway/base"
+)
+
+type DatabaseState struct {
+	ResyncRunning *bool `json:"resync,omitempty"`
+}
+
+type ResyncHandler func(resume *bool)
```
bbrks: What is the meaning of `nil` being passed here? Is that intentional or a side-effect of making the state document take an update?

db/database_state.go
```diff
@@ -0,0 +1,142 @@
...
+	if cas == dbMgr.CAS.Load() {
+		return
+	}
+	dbMgr.resyncHandler(state.ResyncRunning)
+	dbMgr.CAS.Store(cas)
+}
+
+// StopPolling signals the background polling goroutine started by StartPolling to exit.
+func (dbMgr *DatabaseStateMgr) StopPolling() {
+	dbMgr.terminator.Close()
```
bbrks: We should make this close syncronous so that callers know `poll` isn't in the middle of doing an iteration. It could result in a bucket op where the bucket has been `nil`ed during a test close.

Would advise following pattern in changeCache (using `waitForBGTCompletion` as a helper method that waits for a doneChan to close)

bbrks (changes requested):
Let me know if you want to pair up on this

db/database_state.go
RIT3shSapata: The thought process was that the polling will only run when a database is taken offline. When the database is then reloaded, i.e. brought back online, it would hit the codepath which would call `DatabaseContext.Close()` which would call the `StopPolling()` method.

bbrks: Understood - thanks for clarification.

db/database_state.go
RIT3shSapata: I've added the counter and assertions in the test, didn't change the `TempResyncHandler`

db/database_state.go
```diff
@@ -0,0 +1,148 @@
...
+			}
+			return
+		}
+		base.WarnfCtx(ctx, "error while polling for offline database state: %v", err)
+		return
+	}
+	if cas == dbMgr.CAS.Load() {
+		return
+	}
+	dbMgr.resyncHandler(*state.ResyncRunning)
```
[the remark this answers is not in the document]
bbrks: +1

we need to anticipate something else writing state doc that won't be writing the resync field in future

db/background_mgr_resync_dcp.go
```diff
@@ -159,6 +159,10 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
 		} else {
 			cancelResync(errors.New("resync ended normally"))
 		}
+		err := db.DBStateManager.UpdateState(ctx, DatabaseState{ResyncRunning: base.Ptr(false)})
```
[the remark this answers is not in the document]
bbrks: I don't think we have any SDK operations that cancel based on context right now, but it's probably worth defending against by doing the WithoutCancel wrapper for this op

db/database_state.go
bbrks: These test assertions aren't covering a concurrent `poll` and `UpdateState` to catch the comment about CAS above.

`no callback when CAS unchanged` is close but the polling loop only starts after the update is called

db/database_state.go
bbrks: I think this is still an issue.

`UpdateState` writes then stores the new CAS, but the poll goroutine reads CAS without taking dbMgr.lock. A poll running between the `Update` op and `dbMgr.CAS.Store(cas)` will see the new CAS in the bucket, compare
  against the old in-memory CAS, and invoke the handler when it shouldn't have.

With the current noop `TempResyncHandler` this is basically unnoticeable (I left a comment in the previous round to add a counter to assert on to catch this type of thing - but the assertions were in a test handler not using a concurrent polling loop)

db/background_mgr_resync_dcp.go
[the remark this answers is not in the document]
RIT3shSapata: done

db/background_mgr_resync_dcp.go
```diff
@@ -154,6 +154,10 @@ func (r *ResyncManagerDCP) Run(ctx context.Context, options map[string]any, pers
 	ctx = base.CorrelationIDLogCtx(ctx, r.ResyncID)
 	ctx, cancelResync := context.WithCancelCause(ctx)
 	defer func() {
+		stateErr := db.DBStateManager.UpdateState(ctx, DatabaseState{ResyncRunning: base.Ptr(false)})
+		if stateErr != nil {
+			base.WarnfCtx(ctx, "failed to update the database state: %v", err)
```
bbrks:
```suggestion
			base.WarnfCtx(ctx, "failed to update the database state: %v", stateErr)
```

bbrks (changes requested):
This all looks good now with the updated locking in place to avoid the node seeing its own CAS value and triggering another resync function invocation.
Just need 1 `err` variable fix and we can merge this.

bbrks (approved):

PR #8203, opened 2026-04-27
CBG-5261 lock on StartOnlineProcesses

torcolvin:
CBG-5261 lock on StartOnlineProcesses

There are multiple places where calling `Close` with `StartOnlineProcesses` can fail. The way this happens is if the database is called with: `StartOffline: true` and then the configuration is updated again in rapid succession.

There are multiple types of panics that occur. I'm not sure how valuable the test provided is, it was mostly helpful to run thousands of times to make sure all races occur.

The behavior of `changeCache.Init` to lock the change cache but not unlock until `changeCache.Start` would result in a hanging `Close`. This was latent non test only bug.

## Pre-review checklist

- [x]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/537/

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -2313,10 +2313,10 @@ func (db *DatabaseContext) StartOnlineProcesses(ctx context.Context) (returnedEr
 		_ = db.sequences.waitForReleasedSequences(ctx, initialSequenceTime)
 	}
 
+	needChangeCacheUnlock = false
 	if err := db.changeCache.Start(initialSequence); err != nil {
 		return err
 	}
-	needChangeCacheUnlock = false
 
 	// If this is an xattr import node, start import feed.  Must be started after the caching DCP feed, as import cfg
 	// subscription relies on the caching feed.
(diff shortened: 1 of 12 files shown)


db/database.go
```diff
@@ -2194,14 +2194,26 @@ func (dbc *DatabaseContext) GetRequestPlusSequence() (uint64, error) {
 }
 
 func (db *DatabaseContext) StartOnlineProcesses(ctx context.Context) (returnedError error) {
+	// It is possible to call StartOnlineProcesses while Close is called if Database is loaded with
+	// StartOffline: true
+	db.BucketLock.RLock()
+	defer db.BucketLock.RUnlock()
 
+	if db.IsClosed() {
```
bbrks: `IsClosed()` is acquiring another `RLock` recursively, which is [not allowed](https://pkg.go.dev/sync#RWMutex). This will deadlock if this sequence of events happens:
1. `StartOnlineProcesses()`
2. `Close()` - blocks waiting for 1 to unlock
3. `IsClosed()` - blocks waiting for 2's write lock

Your test may be kinda covering this case but is very reliant on a race condition to trigger it and I'm not sure you are hitting it.

bbrks (changes requested):

db/database.go
torcolvin: Great catch! I don't think the tests do catch issues like this which is why I think we could drop the test. Even running 1000s of times I still didn't encounter this bug again.

bbrks (approved):

PR #8209, opened 2026-04-28
CBG-5325: Add REST API spec for Document channel history compaction

RIT3shSapata:
CBG-5325

Describe your PR here...
- Added REST API spec for Document channel history compaction

## Pre-review checklist

- [ ] Link upstream PRs

### DIFF
--- a/docs/api/admin.yaml
+++ b/docs/api/admin.yaml
@@ -115,8 +115,10 @@ paths:
     $ref: './paths/admin/db-_index_init.yaml'
   '/{keyspace}/_purge':
     $ref: './paths/admin/keyspace-_purge.yaml'
-  '/{keyspace}/_prune':
-    $ref: './paths/admin/keyspace-_prune.yaml'
+  '/{keyspace}/_history':
+    $ref: './paths/admin/keyspace-_history.yaml'
+  '/{keyspace}/_history/compact':
+    $ref: './paths/admin/keyspace-_history-compact.yaml'
   '/{db}/_flush':
     $ref: './paths/admin/db-_flush.yaml'
   '/{db}/_online':
--- a/docs/api/admin.yaml
+++ b/docs/api/admin.yaml
@@ -115,6 +115,8 @@ paths:
     $ref: './paths/admin/db-_index_init.yaml'
   '/{keyspace}/_purge':
     $ref: './paths/admin/keyspace-_purge.yaml'
+  '/{keyspace}/_prune':
+    $ref: './paths/admin/keyspace-_prune.yaml'
   '/{db}/_flush':
     $ref: './paths/admin/db-_flush.yaml'
   '/{db}/_online':
--- a/docs/api/paths/admin/keyspace-_history-compact.yaml
+++ b/docs/api/paths/admin/keyspace-_history-compact.yaml
@@ -18,7 +18,8 @@ post:
 
     Required Sync Gateway RBAC roles:
 
-    * Sync Gateway Architect
+    * Sync Gateway Application
+    * Sync Gateway Application Read Only
   requestBody:
     content:
       application/json:
@@ -27,9 +28,7 @@ post:
           properties:
             doc_ids:
               description: |-
-                List of document IDs whose inactive channel history should be compacted.
-                Only the inactive channel entries for these specific documents will be affected.
-                Other documents and active channels in the keyspace will not be modified.
+                List of document IDs whose inactive channel should be compacted.
               type: array
               items:
                 type: string
@@ -38,10 +37,7 @@ post:
                 - doc2
             seq:
               description: |-
-                The sequence number threshold. Inactive channel entries for sequences before this number
-                (sequences at which the document left the channel) will be removed from the specified documents.
-                Active channels and inactive channel entries at or after this sequence number will be preserved.
-                This allows you to clean up old history while keeping recent changes intact.
+                Channel history for inactive channels having end sequences earlier than this sequence will be removed from the specified document's metadata.
               type: number
               example: 12345
   responses:
--- a/docs/api/paths/admin/keyspace-_history-compact.yaml
+++ b/docs/api/paths/admin/keyspace-_history-compact.yaml
@@ -19,16 +19,18 @@ post:
     Required Sync Gateway RBAC roles:
 
     * Sync Gateway Application
-    * Sync Gateway Application Read Only
   requestBody:
     content:
       application/json:
         schema:
           type: object
+          required:
+            - doc_ids
+            - seq
           properties:
             doc_ids:
               description: |-
-                List of document IDs whose inactive channel should be compacted.
+                List of document IDs whose inactive channels should be compacted.
               type: array
               items:
                 type: string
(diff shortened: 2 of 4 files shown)


docs/api/admin.yaml
```diff
@@ -115,6 +115,8 @@ paths:
     $ref: './paths/admin/db-_index_init.yaml'
   '/{keyspace}/_purge':
     $ref: './paths/admin/keyspace-_purge.yaml'
+  '/{keyspace}/_prune':
```
adamcfraser: `_prune` is a relatively generic description (i.e. there's nothing about the name _prune to indicate that it's specifically about "removing channel history from a document".  Given that, I feels like we should either:

1. Extend the existing _compact endpoint with a new 'type' for this functionality, or
2. Come up with more specific naming for the new endpoint

I think the downside of #1 might be around documentation - how to document the different payloads that should be used based on type.

docs/api/paths/admin/keyspace-_prune.yaml
```diff
@@ -0,0 +1,87 @@
...
+      content:
+        application/json:
+          schema:
+            type: object
+            additionalProperties:
+              type: array
+              items:
+                type: string
+              description: |-
+                Array of channel names that remain for the corresponding document after pruning.
```
adamcfraser: Why would we want to return all channel names that remain for each document?  That could be a long list, and I'm not sure what the functional use case would be.  If anything I think we might want to remove the list of channels that *were* pruned.

docs/api/paths/admin/keyspace-_prune.yaml
```diff
@@ -0,0 +1,87 @@
...
+                changes intact.
+              type: number
+              example: 12345
+  responses:
+    "200":
+      description: |-
+        Successfully pruned channel history from the specified documents.
+        Returns a mapping of document IDs to the list of channels that remain after pruning.
+
+        **Note:** If a document ID is not present in the response, it means no channels were pruned for that document.
```
adamcfraser: Similar to the comment below, it might make more sense to have the response list the channels that were pruned, and have that list be empty for documents that weren't pruned (to provide feedback that the document in question exists, but didn't have anything eligible for pruning).

docs/api/paths/admin/keyspace-_prune.yaml
```diff
@@ -0,0 +1,87 @@
...
+                Other documents in the keyspace will not be modified.
+              type: array
+              items:
+                type: string
+              example:
+                - doc1
+                - doc2
+            seqNo:
+              description: |-
+                The sequence number threshold. All channel entries recorded before this sequence number
```
adamcfraser: Description should always describe channels as 'inactive channels', since active channels are not affected.

The sequence being evaluated is the sequence at which the document left the channel - that should be described in this description, instead of just saying 'recorded' which isn't specific.

docs/api/paths/admin/keyspace-_prune.yaml
```diff
@@ -0,0 +1,87 @@
...
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/schemas.yaml#/HTTP-Error
+    "403":
+      $ref: ../../components/responses.yaml#/Unauthorized-database
+
+  tags:
+    - Document
+  operationId: post_keyspace-_prune
```
adamcfraser: Similar to the other case, do we need some type of GET operation to allow people to view the channel history on a document?

adamcfraser (changes requested):

docs/api/paths/admin/keyspace-_prune.yaml
RIT3shSapata: changed it

docs/api/paths/admin/keyspace-_prune.yaml
RIT3shSapata: added it

docs/api/paths/admin/keyspace-_history.yaml
```diff
@@ -0,0 +1,71 @@
...
+# the file licenses/APL2.txt.
+parameters:
+  - $ref: ../../components/parameters.yaml#/keyspace
+get:
+  summary: Get Channel History of Document
+  description: |-
+    Retrieves all inactive channels in the channel history before a particular sequence number.
+
+    This endpoint returns the inactive channels associated with a document (sequences at which the document left the channel)
+    that have history entries before the specified sequence number.
```
adamcfraser: This description seems to be saying the same thing twice?

docs/api/paths/admin/keyspace-_history.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+    Required Sync Gateway RBAC roles:
+
+    * Sync Gateway Architect
```
adamcfraser: I think this should this instead be:
    * Sync Gateway Application
    * Sync Gateway Application Read Only

Architect isn't intended to be able to access application data like documents.

docs/api/paths/admin/keyspace-_history-compact.yaml
```diff
@@ -0,0 +1,86 @@
...
+    * Sync Gateway Architect
+  requestBody:
+    content:
+      application/json:
+        schema:
+          type: object
+          properties:
+            doc_ids:
+              description: |-
+                List of document IDs whose inactive channel history should be compacted.
```
adamcfraser:
```suggestion
                List of document IDs whose inactive channels should be compacted.
```

docs/api/paths/admin/keyspace-_history-compact.yaml
```diff
@@ -0,0 +1,86 @@
...
+                Only the inactive channel entries for these specific documents will be affected.
```
adamcfraser: Can remove this line, redundant with the previous one.

docs/api/paths/admin/keyspace-_history-compact.yaml
```diff
@@ -0,0 +1,86 @@
...
+                Other documents and active channels in the keyspace will not be modified.
```
adamcfraser: I think this line is also redundant?

docs/api/paths/admin/keyspace-_history-compact.yaml
```diff
@@ -0,0 +1,86 @@
...
+              type: array
+              items:
+                type: string
+              example:
+                - doc1
+                - doc2
+            seq:
+              description: |-
+                The sequence number threshold. Inactive channel entries for sequences before this number
+                (sequences at which the document left the channel) will be removed from the specified documents.
```
adamcfraser:
```suggestion
                Channel history for inactive channels having end sequences earlier than this sequence will be removed from the specified document's metadata.
```

docs/api/paths/admin/keyspace-_history-compact.yaml
```diff
@@ -0,0 +1,86 @@
...
+                Active channels and inactive channel entries at or after this sequence number will be preserved.
+                This allows you to clean up old history while keeping recent changes intact.
```
adamcfraser: I think this line can be removed

adamcfraser (changes requested):
A few notes, and review the copilot output, it's relevant.

adamcfraser (approved):

PR #8210, opened 2026-04-28
CBG-5329: Add REST API spec for user channel history pruning

RIT3shSapata:
CBG-5329

Describe your PR here...
- Added REST API spec for user channel history pruning

## Pre-review checklist

- [ ] Link upstream PRs

### DIFF
--- a/docs/api/admin.yaml
+++ b/docs/api/admin.yaml
@@ -51,8 +51,8 @@ paths:
     $ref: './paths/admin/db-_user-name-_session.yaml'
   '/{db}/_user/{name}/_session/{sessionid}':
     $ref: './paths/admin/db-_user-name-_session-sessionid.yaml'
-  '/{db}/_user/{name}/_prune':
-    $ref: './paths/admin/db-_user-name-_prune.yaml'
+  '/{db}/_user/{name}/_history':
+    $ref: './paths/admin/db-_user-name-_history.yaml'
   '/{db}/_role/':
     $ref: './paths/admin/db-_role-.yaml'
   '/{db}/_role/{name}':
--- a/docs/api/admin.yaml
+++ b/docs/api/admin.yaml
@@ -53,6 +53,8 @@ paths:
     $ref: './paths/admin/db-_user-name-_session-sessionid.yaml'
   '/{db}/_user/{name}/_history':
     $ref: './paths/admin/db-_user-name-_history.yaml'
+  '/{db}/_user/{name}/_history/compact':
+    $ref: './paths/admin/db-_user-name-_history-compact.yaml'
   '/{db}/_role/':
     $ref: './paths/admin/db-_role-.yaml'
   '/{db}/_role/{name}':
--- a/docs/api/admin.yaml
+++ b/docs/api/admin.yaml
@@ -51,6 +51,8 @@ paths:
     $ref: './paths/admin/db-_user-name-_session.yaml'
   '/{db}/_user/{name}/_session/{sessionid}':
     $ref: './paths/admin/db-_user-name-_session-sessionid.yaml'
+  '/{db}/_user/{name}/_prune':
+    $ref: './paths/admin/db-_user-name-_prune.yaml'
   '/{db}/_role/':
     $ref: './paths/admin/db-_role-.yaml'
   '/{db}/_role/{name}':
--- a/docs/api/components/schemas.yaml
+++ b/docs/api/components/schemas.yaml
@@ -3379,17 +3379,21 @@ RegistryScope:
       items:
         type: string
 
-UserChannelHistory:
-  description: User's channel history organized by scope and collection.
+UserHistory:
+  description: User history containing various history types organized by scope and collection.
   type: object
-  additionalProperties:
-    description: An object keyed by scope name.
-    type: object
-    x-additionalPropertiesName: scopename
-    additionalProperties:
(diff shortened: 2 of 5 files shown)


docs/api/admin.yaml
```diff
@@ -51,6 +51,8 @@ paths:
     $ref: './paths/admin/db-_user-name-_session.yaml'
   '/{db}/_user/{name}/_session/{sessionid}':
     $ref: './paths/admin/db-_user-name-_session-sessionid.yaml'
+  '/{db}/_user/{name}/_prune':
```
adamcfraser: My preference would be to use `_compact` instead of `_prune`, for consistency with our other REST API endpoints and how we generally describe this kind of operation.

docs/api/paths/admin/db-_user-name-_prune.yaml
```diff
@@ -0,0 +1,105 @@
...
+# in that file, in accordance with the Business Source License, use of this
+# software will be governed by the Apache License, Version 2.0, included in
+# the file licenses/APL2.txt.
+parameters:
+  - $ref: ../../components/parameters.yaml#/db
+  - $ref: ../../components/parameters.yaml#/user-name
+get:
+  summary: Get user's channel information
+  description: |-
+    Retrieve the channel access information for a user, including admin channels, all channels, and collection access configuration.
```
adamcfraser: This API should be working with the channel *history*, not the active channels.  The intention here is to remove entries from the channel history, not from active channels (removing the latter can already be done through the standard processes of removing admin or dynamic grants).  The entries we're attempting to compact here are the ones used to trigger channel revocation processing - stored as channel_history and collection_access.scope.collection.channel_history.

docs/api/paths/admin/db-_user-name-_prune.yaml
```diff
@@ -0,0 +1,105 @@
...
+#
+# Use of this software is governed by the Business Source License included
+# in the file licenses/BSL-Couchbase.txt.  As of the Change Date specified
+# in that file, in accordance with the Business Source License, use of this
+# software will be governed by the Apache License, Version 2.0, included in
+# the file licenses/APL2.txt.
+parameters:
+  - $ref: ../../components/parameters.yaml#/db
+  - $ref: ../../components/parameters.yaml#/user-name
+get:
```
adamcfraser: Calling GET /db/_user/_compact to fetch the user's channel history isn't really intuitive.  Let's discuss offline what options make sense here.

docs/api/admin.yaml
adamcfraser: See additional discussion offline about using a more generic `_history` to support future extensibility for managing role history as well.

docs/api/paths/admin/db-_user-name-_prune.yaml
```diff
@@ -0,0 +1,105 @@
...
+    * Sync Gateway Architect
+  responses:
+    '200':
+      description: User channel information retrieved successfully
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/schemas.yaml#/UserChannelInfo
+          example:
+            admin_channels:
```
adamcfraser: I think we could remove the root level options (that exists on other APIs for backward compatibility), and require callers to use the collection_access approach even when working with _default._default.

adamcfraser (changes requested):
Added a few comments, also see additional Slack discussion.

docs/api/paths/admin/db-_user-name-_history.yaml
```diff
@@ -0,0 +1,39 @@
...
+# in that file, in accordance with the Business Source License, use of this
+# software will be governed by the Apache License, Version 2.0, included in
+# the file licenses/APL2.txt.
+parameters:
+  - $ref: ../../components/parameters.yaml#/db
+  - $ref: ../../components/parameters.yaml#/user-name
+get:
+  summary: Get user's channel history information
+  description: |-
+    Retrieve the Channel history of the specified user.
```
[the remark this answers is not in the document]
adamcfraser: +1

docs/api/paths/admin/db-_user-name-_history.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+    Required Sync Gateway RBAC roles:
+
+    * Sync Gateway Architect
```
adamcfraser, line 18: Should Sync Gateway Application also have permission for this endpoint (and Sync Gateway Application Read Only), or was it decided that this should be Architect only?

docs/api/components/schemas.yaml
```diff
@@ -3378,3 +3378,18 @@ RegistryScope:
       type: array
       items:
         type: string
+
+UserChannelHistory:
```
adamcfraser: We discussed wanting to design this so that it could be extended in future to also work on role history.  The schema as designed doesn't seem like it will support this, as the channel information is at the root level and not within a 'channels' entry.

adamcfraser (changes requested):

docs/api/paths/admin/db-_user-name-_history.yaml
RIT3shSapata: Yes Sync Gateway Application should also have permission for this endpoint. Will add it

adamcfraser (approved):
LGTM

PR #8215, opened 2026-04-29
CBG-5180 add a multi node mode to BackgroundManager

torcolvin:
CBG-5180 add a multi node mode to BackgroundManager

Add a multi node mode existing modes of background manager for purposes of supporting resync. This mode is intended to be cluster aware and support calling Start() on one node and have all other nodes pick up and call the `BackgroundManager.Run` function. If another node calls `Stop` then all noes should stop.

The intended use case is for distributed resync to be able to call `ResyncManager.Run` on all nodes simultaneously.

This should support in this new mode:

- If Start is called and the cluster status is marked as "Running", then allow running to occur
- Update the status document on all nodes, since any node could trigger stop. The status document has entirely identical information on each node (except for stats, to be continued in a future PR). If the current manager is running but another node has moved the state to a stopped state, trigger stop on the local node.
- In this mode, there is no need for a heartbeat document since the intended behavior is to always allow a non running node to pick up this task. The intention behind this is to allow a distributed resync process to resume even if the entire Sync Gateway cluster rolls over. If it was a background manager task that couldn't resume in that case, that logic should be handled otherwise.
- Don't overwrite start_time if another node has already started.

Changes that might seem unrelated:

- Created a new function for updating the status document for the multi node background manager. This is to avoid modifying the existing background manager code too much; support for stopping the update if we detect the cluster status == stopped but our status is running; and allow using `DataStore.Update` in a future PR to allow some fancy stats modification.
- Because `b.State` is read inside `updateMultiNodeClusterAwareStatus` document, I had to make this a lockable property. Since you can not recursively lock a mutex, I decided to create a statusLock to separate modifying the status parameters on BackgroundManager separately. This could be pulled into a separate PR and was only discovered due to some race conditions.

## Pre-review checklist

- [x]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/551/

### DIFF
--- a/db/background_mgr.go
+++ b/db/background_mgr.go
@@ -41,7 +41,7 @@ const (
 
 // errBackgroundManagerAlreadyStopping is returned when a Start or Stop is called while the process is in the Stopping
 // state.
-var errBackgroundManagerAlreadyStopping = base.HTTPErrorf(http.StatusServiceUnavailable, "Process currently stopping. Wait until stopped to retry")
+var errBackgroundManagerStatusAlreadyStopping = base.HTTPErrorf(http.StatusServiceUnavailable, "Process currently stopping. Wait until stopped to retry")
 
 // errBackgroundManagerStatusNotRunning is returned when the bucket status is not running but the local status is.
 var errBackgroundManagerStatusNotRunning = errors.New("status in bucket is not running, but local status is, avoiding overwriting local status to bucket")
@@ -82,7 +82,7 @@ type ClusterAwareBackgroundManagerOptions struct {
 	metadataStore base.DataStore
 	metaKeys      *base.MetadataKeys
 	processSuffix string
-	MultiNode     bool // If true, the background manager is expected to run on all nodes of a Sync Gateway cluster.
+	multiNode     bool // If true, the background manager is expected to run on all nodes of a Sync Gateway cluster.
 
 	lastSuccessfulHeartbeatUnix base.AtomicInt
 }
@@ -269,7 +269,7 @@ func (b *BackgroundManager) markStart(ctx context.Context) error {
 
 	if b.mode() == backgroundManagerModeMultiNode {
 		if b.clusterStateIs(ctx, BackgroundProcessStateStopping) {
-			return errBackgroundManagerAlreadyStopping
+			return errBackgroundManagerStatusAlreadyStopping
 		}
 	}
 
@@ -278,7 +278,7 @@ func (b *BackgroundManager) markStart(ctx context.Context) error {
 	}
 
 	if b.GetRunState() == BackgroundProcessStateStopping {
-		return errBackgroundManagerAlreadyStopping
+		return errBackgroundManagerStatusAlreadyStopping
 	}
 
 	// Now we know that we're the only running process we should instantiate these values
@@ -378,6 +378,8 @@ func (b *BackgroundManager) getStatusFromCluster(ctx context.Context) ([]byte, e
 		return nil, err
 	}
 
+	// In multi node mode, there is no heartbeat document. Each time a node comes online, it is expected to resume the
+	// background process.
 	if b.mode() == backgroundManagerModeMultiNode {
 		return status, nil
 	}
@@ -444,21 +446,13 @@ func (b *BackgroundManager) clearLastErrorMessage() {
 //
 // This will return an error if the status is not in a running state, as already stopped or stopping.
 func (b *BackgroundManager) Stop(ctx context.Context) error {
-	if err := b.markStop(); err != nil {
-		if errors.Is(err, errBackgroundManagerProcessAlreadyStopped) || errors.Is(err, errBackgroundManagerAlreadyStopping) {
+	if err := b.markStop(ctx); err != nil {
+		if errors.Is(err, errBackgroundManagerProcessAlreadyStopped) || errors.Is(err, errBackgroundManagerStatusAlreadyStopping) {
 			return nil
 		}
 		return err
 	}
-
-	if b.mode() == backgroundManagerModeMultiNode {
-		err := b.UpdateStatusClusterAware(ctx)
-		if err != nil {
-			base.WarnfCtx(ctx, "Failed to update cluster status to stopping: %v", err)
-		}
-	}
-
-	b.Terminate()
+	b.stopProcess(ctx)
 	return nil
 }
 
@@ -469,10 +463,11 @@ func (b *BackgroundManager) Terminate() {
 	b.backgroundManagerStatusUpdateWaitGroup.Wait()
 }
 
-func (b *BackgroundManager) markStop() error {
+func (b *BackgroundManager) markStop(ctx context.Context) error {
 	b.lock.Lock()
 	defer b.lock.Unlock()
 
+	currentState := b.GetRunState()
 	if b.mode() == backgroundManagerModeSingleNode {
 		_, _, err := b.clusterAwareOptions.metadataStore.GetRaw(b.clusterAwareOptions.HeartbeatDocID())
 		if err != nil {
@@ -488,22 +483,21 @@ func (b *BackgroundManager) markStop() error {
 		}
 
 		// If this is the node running the service
-		if b.GetRunState() == BackgroundProcessStateRunning {
+		if currentState == BackgroundProcessStateRunning {
 			b.setRunState(BackgroundProcessStateStopping)
 		}
 
 		return nil
 	}
 
-	if b.GetRunState() == BackgroundProcessStateStopping {
-		return errBackgroundManagerAlreadyStopping
+	if currentState == BackgroundProcessStateStopping {
+		return errBackgroundManagerStatusAlreadyStopping
 	}
 
-	if b.GetRunState() == BackgroundProcessStateCompleted || b.GetRunState() == BackgroundProcessStateStopped || b.GetRunState() == BackgroundProcessStateError {
+	if slices.Contains([]BackgroundProcessState{BackgroundProcessStateCompleted, BackgroundProcessStateStopped, BackgroundProcessStateError}, currentState) {
 		return errBackgroundManagerProcessAlreadyStopped
 	}
 
-	b.setRunState(BackgroundProcessStateStopping)
 	return nil
 }
 
@@ -671,16 +665,14 @@ func (b *BackgroundManager) startPollingMultiNodeStatus(ctx context.Context, ter
 	for {
 		select {
 		case <-ticker.C:
-			triggerStop, err := b.watchStatusDocMultiNode(ctx)
+			err := b.updateMultiNodeClusterAwareStatus(ctx)
 			if err != nil {
-				base.WarnfCtx(ctx, "Failed to watch poll status doc: %v, will retry", err)
-			}
-			if triggerStop {
-				continue
-			}
-			err = b.updateMultiNodeClusterAwareStatus(ctx)
-			if err != nil && !errors.Is(err, errBackgroundManagerStatusNotRunning) {
-				base.DebugfCtx(ctx, base.KeyAll, "Failed to update multi node cluster aware status: %v, will retry", err)
+				if errors.Is(err, errBackgroundManagerStatusNotRunning) {
+					b.stopProcess(ctx)
+					return
+				} else {
+					base.DebugfCtx(ctx, base.KeyAll, "Failed to update multi node cluster aware status: %v, will retry", err)
+				}
 			}
 		case <-terminator.Done():
 			ticker.Stop()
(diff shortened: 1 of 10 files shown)


db/background_mgr.go
```diff
@@ -140,8 +164,13 @@ func (b *BackgroundManager) Start(ctx context.Context, options map[string]any) e
 				}
 			}
 		}(b.terminator)
-	}
 
+	}
+	if b.mode() == backgroundManagerModeMultiNode {
+		b.backgroundManagerStatusUpdateWaitGroup.Go(func() {
+			b.startPollingMultiNodeStatus(ctx, b.terminator)
+		})
```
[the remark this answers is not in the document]
torcolvin: Go 1.25+ does support this syntax.

db/background_mgr.go
```diff
@@ -341,15 +427,37 @@ func (b *BackgroundManager) resetStatus() {
...
 	if err := b.markStop(); err != nil {
+		if errors.Is(err, errBackgroundManagerProcessAlreadyStopped) || errors.Is(err, errBackgroundManagerAlreadyStopping) {
+			return nil
+		}
 		return err
 	}
 
+	if b.mode() == backgroundManagerModeMultiNode {
+		// Use a detached context for status update as the original context might be cancelled
+		err := b.UpdateStatusClusterAware(ctx)
```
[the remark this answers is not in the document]
torcolvin: This comment is just wrong.

db/background_mgr_test.go
```diff
@@ -0,0 +1,323 @@
...
+		managers[i] = &BackgroundManager{
+			name:                fmt.Sprintf("mgr%d", i),
+			clusterAwareOptions: options,
+			Process:             processes[i],
+		}
+	}
+
+	// Start all simultaneously
+	var wg sync.WaitGroup
+	for i := range numNodes {
```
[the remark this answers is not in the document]
torcolvin: go supports this syntax in go 1.25

db/background_mgr.go
```diff
@@ -64,6 +82,7 @@ type ClusterAwareBackgroundManagerOptions struct {
 	metadataStore base.DataStore
 	metaKeys      *base.MetadataKeys
 	processSuffix string
+	MultiNode     bool // If true, the background manager is expected to run on all nodes of a Sync Gateway cluster.
```
adamcfraser: Is it required for this to be public (unlike the other ClusterAwareBackgroundManagerOptions)?

db/background_mgr.go
```diff
@@ -341,15 +428,36 @@ func (b *BackgroundManager) resetStatus() {
...
+func (b *BackgroundManager) Stop(ctx context.Context) error {
 	if err := b.markStop(); err != nil {
+		if errors.Is(err, errBackgroundManagerProcessAlreadyStopped) || errors.Is(err, errBackgroundManagerAlreadyStopping) {
+			return nil
+		}
 		return err
 	}
 
+	if b.mode() == backgroundManagerModeMultiNode {
+		err := b.UpdateStatusClusterAware(ctx)
```
adamcfraser: Is the following summary of how this works accurate:
- For single-node cluster aware processes, cross-node Stop notification is done through the heartbeat document (shouldStop)
- For multi-node cluster aware processes, cross-node Stop notification is done through the status document (using status.state)

If yes, I think it would be clearer to move this logic down into markStop, and don't use markStop in watchStatusDocMultiNode.  (see separate comment on that usage)

db/background_mgr.go
```diff
@@ -484,6 +663,71 @@ func (b *BackgroundManager) UpdateHeartbeatDocClusterAware(ctx context.Context)
...
+	state, err := b.getClusterStatusState(ctx)
+	if err != nil {
+		if base.IsDocNotFoundError(err) {
+			return false, nil
+		}
+		return false, err
+	}
+
+	if slices.Contains([]BackgroundProcessState{BackgroundProcessStateStopping, BackgroundProcessStateStopped, BackgroundProcessStateCompleted, BackgroundProcessStateError}, state) && b.GetRunState() == BackgroundProcessStateRunning {
+		_ = b.markStop()
```
adamcfraser: This is using markStop just to update the local run state to Stopping (and also reusing the RunState transition checking).  I think that muddies the original intention of markStop ("call markStop when you want to notify all nodes to stop").  It also acquires b.lock when we only really need the new statusLock?
Should we create a new method like "setRunStateStopped" to do this work instead?

db/background_mgr.go
```diff
@@ -381,43 +488,80 @@ func (b *BackgroundManager) markStop() error {
...
 	}
 
-	if b.State == BackgroundProcessStateCompleted || b.State == BackgroundProcessStateStopped {
-		return processAlreadyStoppedErr
+	if b.GetRunState() == BackgroundProcessStateCompleted || b.GetRunState() == BackgroundProcessStateStopped || b.GetRunState() == BackgroundProcessStateError {
+		return errBackgroundManagerProcessAlreadyStopped
 	}
 
-	b.State = BackgroundProcessStateStopping
+	b.setRunState(BackgroundProcessStateStopping)
```
adamcfraser: This code has ended up a bit clunky with the new Status lock - it's acquiring and releasing b.statusLock 5 times in this function.  
Is the expected contract that setRunState should only be called while holding b.lock, so there no chance of the status changing in between condition checks on line 502?  If not, the behaviour of that line seems unpredictable.  If yes, we should be able to fetch the status once and reuse it in this method.

db/background_mgr.go
```diff
@@ -35,6 +39,19 @@ const (
 	BackgroundProcessStateError     BackgroundProcessState = "error"
 )
 
+// errBackgroundManagerAlreadyStopping is returned when a Start or Stop is called while the process is in the Stopping
+// state.
+var errBackgroundManagerAlreadyStopping = base.HTTPErrorf(http.StatusServiceUnavailable, "Process currently stopping. Wait until stopped to retry")
```
```suggestion
var errBackgroundManagerProcessAlreadyStopping = base.HTTPErrorf(http.StatusServiceUnavailable, "Process currently stopping. Wait until stopped to retry")
```
for consistency with other error naming.

db/background_mgr.go
```diff
@@ -103,34 +122,52 @@ func (b *BackgroundManager) GetName() string {
 func (b *BackgroundManager) Start(ctx context.Context, options map[string]any) error {
 	err := b.markStart(ctx)
 	if err != nil {
+		if b.mode() == backgroundManagerModeMultiNode && errors.Is(err, errBackgroundManagerProcessAlreadyRunning) {
```
adamcfraser, line 125: Can we push this check down into markStart (since this is the only caller)?

db/background_mgr.go
```diff
@@ -103,34 +122,52 @@ func (b *BackgroundManager) GetName() string {
...
+
+	// If we're resuming a cluster-aware process, try to reuse the previous start time
+	if processClusterStatus != nil {
+		var clusterStatus struct {
+			Status BackgroundManagerStatus `json:"status"`
+		}
+
+		err := base.JSONUnmarshal(processClusterStatus, &clusterStatus)
+		if err != nil {
+			base.InfofCtx(ctx, base.KeyAll, "Could not get the cluster status before calling BackgroundManager.Run %v", err)
```
```suggestion
			base.InfofCtx(ctx, base.KeyAll, "Could not unmarshal the cluster status before calling BackgroundManager.Run %v", err)
```

db/background_mgr.go
```diff
@@ -103,34 +122,52 @@ func (b *BackgroundManager) GetName() string {
...
+		}
+		if clusterStatus.Status.State == BackgroundProcessStateRunning && !clusterStatus.Status.StartTime.IsZero() {
```
adamcfraser, line 152: Why is there a limitation on persisted Status.State=running to reuse the start time?

db/background_mgr.go
```diff
@@ -484,6 +663,71 @@ func (b *BackgroundManager) UpdateHeartbeatDocClusterAware(ctx context.Context)
...
+	ticker := time.NewTicker(BackgroundManagerStatusUpdateIntervalSecs * time.Second)
+	for {
+		select {
+		case <-ticker.C:
+			triggerStop, err := b.watchStatusDocMultiNode(ctx)
+			if err != nil {
+				base.WarnfCtx(ctx, "Failed to watch poll status doc: %v, will retry", err)
+			}
+			if triggerStop {
+				continue
```
adamcfraser: This feels like a bit of a convoluted way of handling this (watchStatusDocMultiNode closes the terminator, and so we continue and then fall through to line 685).  If triggerStop=true, maybe just duplicate the ticker.Stop() and return here so it's explicit what's going on?

db/background_mgr.go
```diff
@@ -484,6 +663,71 @@ func (b *BackgroundManager) UpdateHeartbeatDocClusterAware(ctx context.Context)
...
+			}
+			err = b.updateMultiNodeClusterAwareStatus(ctx)
+			if err != nil && !errors.Is(err, errBackgroundManagerStatusNotRunning) {
```
adamcfraser: What's the intended behaviour when err is errBackgroundManagerStatusNotRunning?  It looks like it's going to silently ignore this (occurs when the bucket status is stopped/stopping/completed/error and the local status is running) - is the idea to fall through and perform the same check in the next ticker call to watchStatusDocMultiNode?

db/background_mgr.go
```diff
@@ -140,8 +177,13 @@ func (b *BackgroundManager) Start(ctx context.Context, options map[string]any) e
...
-	}
 
+	}
+	if b.mode() == backgroundManagerModeMultiNode {
+		b.backgroundManagerStatusUpdateWaitGroup.Go(func() {
+			b.startPollingMultiNodeStatus(ctx, b.terminator)
+		})
+	}
 	go func() {
 		err := b.Process.Run(ctx, options, b.UpdateStatusClusterAware, b.terminator)
```
adamcfraser, line 188: Is the idea for backgroundManagerModeMultiNode that as soon as one node's Run completes, the process is stopped for all nodes?  Or said another way - it's the responsibility of the individual process implementation to ensure that a given node doesn't exit Run() until all nodes are done?  That would be useful to document somewhere.

db/background_mgr.go
```diff
@@ -295,6 +378,10 @@ func (b *BackgroundManager) getStatusFromCluster(ctx context.Context) ([]byte, e
 		return nil, err
 	}
 
+	if b.mode() == backgroundManagerModeMultiNode {
```
adamcfraser, line 383: Can you add a comment about why the handling below isn't needed in multi-node mode?

adamcfraser (changes requested):
General approach looks fine, a few questions on some of the specifics.

db/background_mgr.go
torcolvin: Good catch, this was the part of the code that was written by gemini before I wrote this. It really likes to make everything public.

db/background_mgr.go
torcolvin: I like this. There are two things that need to happen:

1. change the status to stopped, and close the terminator.

db/background_mgr.go
torcolvin: I'm not sure the right thing to do here, I agree that this sort of belongs in `markStart`. The intention behind this is to have `Start` be idempotent if it is already running in multi node mode. What needs to happen is that `Start` returns immediately and doesn't spawn another goroutine and call `Run`.

db/background_mgr.go
torcolvin: This is done in a clunky way but above the modified code, we want to use the latest time:

b.setStartTime(time.Now().UTC())

If we use clusterStatus time if it isn't running and happens to be in the `Stopped` state, then it would pick up an earlier time.

db/background_mgr.go
torcolvin: I think the expectation is:

If one node exits `Run` this is a fatal error (or a success) and all nodes should enter stopped state. I don't think there's an expectation that they all will exit at the same time.

For distributed resync, they will not exit until we determine that all DCP feeds are done, but the first node who exits can call `Stopping`. 

Now that I say this out loud, I wonder if this is wrong because the state will transition to Stop before the background process is actually done on all nodes. I think it would be closer to correct to transition to `Stopping` until all nodes are done. The problem here is that I don't actually know as part of the background manager who all the nodes are. I wonder if this needs a rethink to the implementation.

db/background_mgr.go
torcolvin: agree, it is a mess.

db/background_mgr.go
torcolvin: I moved this around a bit to trigger Stop here, which I think is more clear. I'm not sure if this is what this comment intended.

db/background_mgr.go
torcolvin: I'm changing this to use `stopProcess` to avoid this error, and removed the watch function entirely since it is handled by `updateMultiNodeClusterAwareStatus`.

db/background_mgr.go
adamcfraser: Ah ok - so 'start time' is intended to represent 'the most recent start' in the case where the task is manually stopped then started.  That makes sense.

db/background_mgr.go
adamcfraser: That makes sense - thanks for the clarification.

db/background_mgr.go
adamcfraser: As discussed offline, the contract will be that the individual process is responsible for not exiting Run() until all nodes are done.  Current implementation makes sense with that information.

adamcfraser (approved):

PR #8218, opened 2026-04-30
CBG-5326: Function to handle document channel history compaction

RIT3shSapata:
CBG-5326

Describe your PR here...
- Added a function to compact the channel history of a given document based on sequence number
- Added a high level test coverage for the same

## Pre-review checklist

- [ ] Link upstream PRs

- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/547/
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/563/
- [x]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/590/
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/596/
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/631/
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/682/

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -245,6 +245,19 @@ func (c *DatabaseCollection) CompactDocChannelHistory(ctx context.Context, docid
 		c.dbStats().Database().Crc32MatchCount.Add(1)
 	}
 
+	if !isSgWrite {
+		var importErr error
+
+		doc, importErr = c.OnDemandImportForGet(ctx, docid, doc, rawDoc, xattrs, cas)
+		if importErr != nil {
+			return nil, importErr
+		}
+		if doc == nil {
+			return nil, fmt.Errorf("skipping compaction of document %s, %v ", base.UD(docid), base.ErrNotFound)
+		}
+		cas = doc.Cas
+	}
+
 	compactedChannels := make([]string, 0)
 
 	doc.SyncData.ChannelSetHistory = slices.DeleteFunc(doc.SyncData.ChannelSetHistory, func(channel ChannelSetEntry) bool {
@@ -296,11 +309,9 @@ func (c *DatabaseCollection) CompactDocChannelHistory(ctx context.Context, docid
 	// build macro expansion for sync data. This will avoid the update to xattrs causing an extra import event (i.e. sync cas will be == to doc cas)
 	opts := &sgbucket.MutateInOptions{}
 	// Only update _sync.cas and _mou.cas if the pre-compaction doc had already been imported by SGW
-	if isSgWrite {
-		opts.MacroExpansion = []sgbucket.MacroExpansionSpec{
-			sgbucket.NewMacroExpansionSpec(XattrMouCasPath(), sgbucket.MacroCas),
-			sgbucket.NewMacroExpansionSpec(xattrCasPath(base.SyncXattrName), sgbucket.MacroCas),
-		}
+	opts.MacroExpansion = []sgbucket.MacroExpansionSpec{
+		sgbucket.NewMacroExpansionSpec(XattrMouCasPath(), sgbucket.MacroCas),
+		sgbucket.NewMacroExpansionSpec(xattrCasPath(base.SyncXattrName), sgbucket.MacroCas),
 	}
 	opts.PreserveExpiry = true // if doc has expiry, we should preserve this
 
--- a/db/crud.go
+++ b/db/crud.go
@@ -231,7 +231,7 @@ func (c *DatabaseCollection) CompactDocChannelHistory(ctx context.Context, docid
 	}
 
 	xattrKeys := []string{base.SyncXattrName, base.VirtualXattrRevSeqNo, base.MouXattrName}
-	_, xattrs, cas, err := c.dataStore.GetWithXattrs(ctx, key, xattrKeys)
+	rawDoc, xattrs, cas, err := c.dataStore.GetWithXattrs(ctx, key, xattrKeys)
 	if err != nil {
 		return err
 	}
@@ -239,6 +239,20 @@ func (c *DatabaseCollection) CompactDocChannelHistory(ctx context.Context, docid
 	doc, err := c.unmarshalDocumentWithXattrs(ctx, key, nil, xattrs, cas, DocUnmarshalSync)
 	if err != nil {
 		return err
+	}
+
+	isSgWrite, crc32Match, _ := doc.IsSGWrite(ctx, rawDoc)
+	if crc32Match {
+		c.dbStats().Database().Crc32MatchCount.Add(1)
+	}
+
+	if !isSgWrite {
+		var importErr error
+
+		doc, importErr = c.OnDemandImportForGet(ctx, docid, doc, rawDoc, xattrs, cas)
+		if importErr != nil {
+			return importErr
+		}
 	}
 
 	// Store lengths before compaction to detect if any changes occur
@@ -274,7 +288,8 @@ func (c *DatabaseCollection) CompactDocChannelHistory(ctx context.Context, docid
 
 	revSeqNo, err := unmarshalRevSeqNo(xattrs[base.VirtualXattrRevSeqNo])
 	if err != nil {
-		base.InfofCtx(ctx, base.KeyCRUD, `Could not determine the revSeqNo when attempting to compact channel history for doc: %s. Error: %v`, base.UD(docid), err)
+		base.WarnfCtx(ctx, `Could not determine the revSeqNo when attempting to compact channel history for doc %s - history will not be compacted: %v`, base.UD(docid), err)
+		return base.RedactErrorf(`Could not determine the revSeqNo when attempting to compact channel history for doc %s - history will not be compacted: %v`, base.UD(docid), err)
 	}
 
 	metadataOnlyUpdate := computeMetadataOnlyUpdate(doc.Cas, revSeqNo, doc.MetadataOnlyUpdate)
--- a/db/crud.go
+++ b/db/crud.go
@@ -223,7 +223,7 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
 }
 
 // CompactDocChannelHistory removes channel history entries that ended at or before the given sequence number.
-// This is used to truncate stale channel assignment history to reduce storage overhead.
+// This is used to prune stale channel assignment history to reduce storage overhead.
 func (c *DatabaseCollection) CompactDocChannelHistory(ctx context.Context, docid string, seq uint64) ([]string, error) {
 	key := realDocID(docid)
 	if key == "" {
@@ -243,19 +243,6 @@ func (c *DatabaseCollection) CompactDocChannelHistory(ctx context.Context, docid
 	isSgWrite, crc32Match, _ := doc.IsSGWrite(ctx, rawDoc)
 	if crc32Match {
 		c.dbStats().Database().Crc32MatchCount.Add(1)
-	}
-
-	if !isSgWrite {
-		var importErr error
-
-		doc, importErr = c.OnDemandImportForGet(ctx, docid, doc, rawDoc, xattrs, cas)
-		if importErr != nil {
-			return nil, importErr
-		}
-		if doc == nil {
-			return nil, base.ErrNotFound
-		}
-		cas = doc.Cas
 	}
 
 	compactedChannels := make([]string, 0)
@@ -309,8 +296,10 @@ func (c *DatabaseCollection) CompactDocChannelHistory(ctx context.Context, docid
 	// build macro expansion for sync data. This will avoid the update to xattrs causing an extra import event (i.e. sync cas will be == to doc cas)
 	opts := &sgbucket.MutateInOptions{}
 	spec := []sgbucket.MacroExpansionSpec{
-		sgbucket.NewMacroExpansionSpec(xattrCasPath(base.SyncXattrName), sgbucket.MacroCas),
 		sgbucket.NewMacroExpansionSpec(XattrMouCasPath(), sgbucket.MacroCas),
+	}
+	if isSgWrite {
+		spec = append(spec, sgbucket.NewMacroExpansionSpec(xattrCasPath(base.SyncXattrName), sgbucket.MacroCas))
 	}
 	opts.MacroExpansion = spec
 	opts.PreserveExpiry = true // if doc has expiry, we should preserve this
--- a/db/crud.go
+++ b/db/crud.go
@@ -224,21 +224,21 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
 
 // CompactDocChannelHistory removes channel history entries that ended at or before the given sequence number.
 // This is used to truncate stale channel assignment history to reduce storage overhead.
-func (c *DatabaseCollection) CompactDocChannelHistory(ctx context.Context, docid string, seq uint64) error {
+func (c *DatabaseCollection) CompactDocChannelHistory(ctx context.Context, docid string, seq uint64) ([]string, error) {
 	key := realDocID(docid)
 	if key == "" {
-		return base.HTTPErrorf(400, "Invalid doc ID")
+		return nil, base.HTTPErrorf(400, "Invalid doc ID")
 	}
 
 	xattrKeys := []string{base.SyncXattrName, base.VirtualXattrRevSeqNo, base.MouXattrName}
 	rawDoc, xattrs, cas, err := c.dataStore.GetWithXattrs(ctx, key, xattrKeys)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	doc, err := c.unmarshalDocumentWithXattrs(ctx, key, nil, xattrs, cas, DocUnmarshalSync)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	isSgWrite, crc32Match, _ := doc.IsSGWrite(ctx, rawDoc)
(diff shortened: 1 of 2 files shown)


db/crud.go
```diff
@@ -216,6 +216,67 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
...
+		return
+	}
+
+	doc, err := c.unmarshalDocumentWithXattrs(ctx, docid, nil, xattrs, cas, DocUnmarshalSync)
+	if err != nil {
+		return
+	}
+
+	var compactedChannelHistory []ChannelSetEntry
+	for _, channel := range doc.SyncData.ChannelSetHistory {
```
adamcfraser: The compact option needs to be removing inactive channels from three places in SyncData:

- Channels, based on ChannelRemoval.Seq
- ChannelSet, based on ChannelSetEntry.End
- ChannelSetHistory, based on ChannelSetEntry.End

adamcfraser (changes requested):
Left one comment - have a look at that as well as the copilot comments and assign back once you've updated the PR.

db/crud.go
RIT3shSapata: done

db/crud.go
```diff
@@ -216,6 +216,83 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
...
+	rawSyncXattr, err := base.JSONMarshal(doc.SyncData)
+	if err != nil {
+		return base.RedactErrorf("failed to marshal sync data when trying to compact channel history for doc:%s. Error: %v", base.UD(docid), err)
+	}
+
+	revSeqNo, err := unmarshalRevSeqNo(xattrs[base.VirtualXattrRevSeqNo])
+	if err != nil {
+		base.InfofCtx(ctx, base.KeyCRUD, `Could not determine the revSeqNo when attempting to compact channel history for doc: %s. Error: %v`, base.UD(docid), err)
+	}
+
```
adamcfraser, line 301: I believe you should use the existing computeMetadataUpdate function here to build the MetadataOnlyUpdate, to cover the case where the channel history compaction is being performed on top of a previous metadata-only update (like import).  In that situation prevCas is set differently.

db/crud.go
```diff
@@ -216,6 +216,83 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
...
+// This is used to truncate stale channel assignment history to reduce storage overhead.
+func (c *DatabaseCollection) CompactDocChannelHistory(ctx context.Context, docid string, seq uint64) (err error) {
+	key := realDocID(docid)
+	if key == "" {
+		return base.HTTPErrorf(400, "Invalid doc ID")
+	}
+
+	_, xattrs, cas, err := c.dataStore.GetWithXattrs(ctx, key, c.syncGlobalSyncMouRevSeqNoAndUserXattrKeys())
+	if err != nil {
+		return
```
adamcfraser: Avoid naked returns (here and elsewhere)

db/crud.go
```diff
@@ -216,6 +216,83 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
...
 
+// CompactDocChannelHistory removes channel history entries that ended at or before the given sequence number.
+// This is used to truncate stale channel assignment history to reduce storage overhead.
+func (c *DatabaseCollection) CompactDocChannelHistory(ctx context.Context, docid string, seq uint64) (err error) {
+	key := realDocID(docid)
+	if key == "" {
+		return base.HTTPErrorf(400, "Invalid doc ID")
+	}
+
+	_, xattrs, cas, err := c.dataStore.GetWithXattrs(ctx, key, c.syncGlobalSyncMouRevSeqNoAndUserXattrKeys())
```
adamcfraser: I generally try to avoid shadowing the return variable - in my experience that helps avoid subtle bugs.

db/crud.go
```diff
@@ -216,6 +216,83 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
...
+	if err != nil {
+		return
+	}
+
+	doc, err := c.unmarshalDocumentWithXattrs(ctx, docid, nil, xattrs, cas, DocUnmarshalSync)
+	if err != nil {
+		return
+	}
+
+	var compactedChannelSetHistory []ChannelSetEntry
```
adamcfraser: Removal from ChannelSetHistory and ChannelSet seem like a good candidates for slices.DeleteFunc, to avoid having to allocate a new slice every time.

db/crud.go
```diff
@@ -216,6 +216,83 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
...
+	}
+	doc.SyncData.ChannelSet = compactedChannelSet
+
+	compactedChannels := make(channels.ChannelMap)
+	for chanName, chanEntry := range doc.SyncData.Channels {
+		if chanEntry == nil || chanEntry.Seq > seq {
+			compactedChannels[chanName] = chanEntry
+		}
+	}
+	doc.SyncData.Channels = compactedChannels
```
adamcfraser: It's safe in go to remove entries from a map while you range over it, you don't need to allocate a separate compactedChannels here.

db/crud.go
```diff
@@ -216,6 +216,83 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
...
+		PreviousRevSeqNo: revSeqNo,
+	}
+	rawMouXattr, err := base.JSONMarshal(metadataOnlyUpdate)
+	if err != nil {
+		return base.RedactErrorf("failed to marshal _mou when attempting to compact channel history for doc: %s. Error: %v", base.UD(docid), err)
+	}
+
+	// build macro expansion for sync data. This will avoid the update to xattrs causing an extra import event (i.e. sync cas will be == to doc cas)
+	opts := &sgbucket.MutateInOptions{}
+	spec := append(macroExpandSpec(base.SyncXattrName), sgbucket.NewMacroExpansionSpec(XattrMouCasPath(), sgbucket.MacroCas))
```
adamcfraser: We don't actually need to be recomputing crc32c (since the body hasn't changed) - that's getting picked up by macroExpandSpec.  It might be harmless (other than extra overhead), but I might consider a simplified version of macroExpandSpec that's sync.cas only here.

adamcfraser (changes requested):

db/crud.go
```diff
@@ -216,6 +217,85 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
...
+	}
+
+	rawSyncXattr, err := base.JSONMarshal(doc.SyncData)
+	if err != nil {
+		return base.RedactErrorf("failed to marshal sync data when trying to compact channel history for doc:%s. Error: %v", base.UD(docid), err)
+	}
+
+	revSeqNo, err := unmarshalRevSeqNo(xattrs[base.VirtualXattrRevSeqNo])
+	if err != nil {
+		base.InfofCtx(ctx, base.KeyCRUD, `Could not determine the revSeqNo when attempting to compact channel history for doc: %s. Error: %v`, base.UD(docid), err)
```
```suggestion
		base.InfofCtx(ctx, base.KeyCRUD, `Could not determine the revSeqNo when attempting to compact channel history for doc %s - history will not be compacted:  %v`, base.UD(docid), err)
```
Generally for info-level logging of non-actionable errors:
- Better to avoid the word "Error" in logs, to avoid ambiguity around the severity of the issue
- Describe how the system is handling the situation

adamcfraser: Is it actually safe to continue here if we weren't able to unmarshal revSeqNo?  I think storing an mou with an incorrect revSeqNo has implications for XDCR conflict resolution.

db/crud.go
```diff
@@ -216,6 +217,85 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
...
+
+	rawMouXattr, err := base.JSONMarshal(metadataOnlyUpdate)
+	if err != nil {
+		return base.RedactErrorf("failed to marshal _mou when attempting to compact channel history for doc: %s. Error: %v", base.UD(docid), err)
+	}
+
+	// build macro expansion for sync data. This will avoid the update to xattrs causing an extra import event (i.e. sync cas will be == to doc cas)
+	opts := &sgbucket.MutateInOptions{}
+	spec := []sgbucket.MacroExpansionSpec{
+		sgbucket.NewMacroExpansionSpec(xattrCasPath(base.SyncXattrName), sgbucket.MacroCas),
```
adamcfraser: It's correct to update sync.cas here, because we don't want this compaction to trigger a new document import.  However, this means that we need to make sure this processing isn't being run on documents that haven't yet been imported - otherwise this cas update will make it appear as if this document has already been imported (and so the sync metadata would be incorrect).  

I think the best way to avoid this is to trigger on-demand import as needed when initially fetching the document.  I think you can do that by using GetDocument instead of GetWithXattrs when originally fetching the document on line 229 - this has the on-demand import work built-in.

Will want to add a test case for that (running compact on a non-imported document).

rest/api_test.go
```diff
@@ -3908,3 +3908,269 @@ func TestFetchBackupRevisionByCVThroughAPI(t *testing.T) {
...
+	}
+	rt := NewRestTester(t, &RestTesterConfig{SyncFn: channels.DocChannelsSyncFunction})
+	defer rt.Close()
+
+	collection, ctx := rt.GetSingleTestDatabaseCollection()
+
+	t.Run("basic compaction", func(t *testing.T) {
+		var body db.Body
+		// Create a document with a single channel assignment
+		resp := rt.SendAdminRequest(http.MethodPut, "/{{.keyspace}}/doc1", `{"channels": ["test"]}`)
```
adamcfraser: Using the RT helper functions can reduce the test boilerplate.  rt.PutDoc in this case, rt.UpdateDoc or rt.UpdateDocRev for subsequent updates.

adamcfraser (changes requested):
I like the new approach for computing changes.  Some comments about handling for on-demand import and minor style suggestions.

db/crud.go
RIT3shSapata: changed the log level to warn and am returning the same error, therefore exiting the compaction process

db/crud.go
RIT3shSapata: Added check for non imported document and triggering on demand import in the same function. I went through the GetDocument method and the xattr keys which it fetches is `base.SyncXattrName, base.VvXattrName, base.GlobalXattrName`, the xattr keys which are required for compaction are `base.SyncXattrName, base.VirtualXattrRevSeqNo, base.MouXattrName`. 

Let me know your thoughts on this approach, or would you prefer I use `GetDocument` itself.

Also while writing the test for this scenario, ran into a few challenges, will discuss this further offline.

db/crud.go
```diff
@@ -216,6 +217,100 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
...
+
+	isSgWrite, crc32Match, _ := doc.IsSGWrite(ctx, rawDoc)
+	if crc32Match {
+		c.dbStats().Database().Crc32MatchCount.Add(1)
+	}
+
+	if !isSgWrite {
+		var importErr error
+
+		doc, importErr = c.OnDemandImportForGet(ctx, docid, doc, rawDoc, xattrs, cas)
```
gregns1, line 251: Make sure you have a nil check on doc here too to avoid the same panic found in https://jira.issues.couchbase.com/browse/CBG-5355 

Code else where for reference: https://github.com/couchbase/sync_gateway/blob/d09f0fd3bf44c3575cf103e9d275eefd0fa60a77/db/crud.go#L198

We have a ticket in backlog to make this code better but not sure if we'll get to it in the 4.1 timeframe so be good to protect against it here.

db/crud.go
```diff
@@ -221,6 +222,107 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
 
 }
 
+// CompactDocChannelHistory removes channel history entries that ended at or before the given sequence number.
+// This is used to truncate stale channel assignment history to reduce storage overhead.
```
```suggestion
// This is used to prune channel assignment history to reduce storage overhead.
```

db/crud.go
```diff
@@ -221,6 +222,107 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
...
+	}
+
+	revSeqNo, err := unmarshalRevSeqNo(xattrs[base.VirtualXattrRevSeqNo])
+	if err != nil {
+		base.WarnfCtx(ctx, `Could not determine the revSeqNo when attempting to compact channel history for doc %s - history will not be compacted: %v`, base.UD(docid), err)
+		return nil, base.RedactErrorf(`Could not determine the revSeqNo when attempting to compact channel history for doc %s - history will not be compacted: %v`, base.UD(docid), err)
+	}
+
+	metadataOnlyUpdate := computeMetadataOnlyUpdate(doc.Cas, revSeqNo, doc.MetadataOnlyUpdate)
+
```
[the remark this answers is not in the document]
adamcfraser: @RIT3shSapata The above comment is correct - if an on-demand import occurs, the document's revSeqNo will have changed and will need to be re-fetched.  
Thinking about this further, I'm wondering if we should take a different approach for documents that are pending import.  Let's discuss offline.

adamcfraser (changes requested):

db/crud.go
```diff
@@ -221,6 +222,96 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
...
+	if err != nil {
+		return nil, base.RedactErrorf("failed to marshal _mou when attempting to compact channel history for doc: %s. Error: %v", base.UD(docid), err)
+	}
+
+	// build macro expansion for sync data. This will avoid the update to xattrs causing an extra import event (i.e. sync cas will be == to doc cas)
+	opts := &sgbucket.MutateInOptions{}
+	spec := []sgbucket.MacroExpansionSpec{
+		sgbucket.NewMacroExpansionSpec(XattrMouCasPath(), sgbucket.MacroCas),
+	}
+	if isSgWrite {
```
```suggestion
	// Only update _sync.cas if the pre-compaction doc had already been imported by SGW
	if isSgWrite {
```

db/crud.go
```diff
@@ -221,6 +222,96 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
...
+
+	rawMouXattr, err := base.JSONMarshal(metadataOnlyUpdate)
+	if err != nil {
+		return nil, base.RedactErrorf("failed to marshal _mou when attempting to compact channel history for doc: %s. Error: %v", base.UD(docid), err)
+	}
+
+	// build macro expansion for sync data. This will avoid the update to xattrs causing an extra import event (i.e. sync cas will be == to doc cas)
+	opts := &sgbucket.MutateInOptions{}
+	spec := []sgbucket.MacroExpansionSpec{
+		sgbucket.NewMacroExpansionSpec(XattrMouCasPath(), sgbucket.MacroCas),
```
adamcfraser, line 313: Looking at the import code, we also need to skip updating the mou when !isSgWrite, because otherwise the eventual import is going to (incorrectly) fail to add an entry to the HLV:
https://github.com/couchbase/sync_gateway/blob/main/db/crud.go#L1023

adamcfraser: It would be ideal to have a unit test for this scenario to ensure it works as expected. 

I think this should just be a matter of modifying TestCompactNonImportedDocWithAutoImport to ensure that the cv is being updated after step 9 (i.e. it shouldn't match the cv after step 5)

adamcfraser (changes requested):
Handling needed for _mou in the non-sgWrite case.

db/crud.go
```diff
@@ -295,10 +295,10 @@ func (c *DatabaseCollection) CompactDocChannelHistory(ctx context.Context, docid
 
 	// build macro expansion for sync data. This will avoid the update to xattrs causing an extra import event (i.e. sync cas will be == to doc cas)
 	opts := &sgbucket.MutateInOptions{}
-	spec := []sgbucket.MacroExpansionSpec{
-		sgbucket.NewMacroExpansionSpec(XattrMouCasPath(), sgbucket.MacroCas),
-	}
+	spec := []sgbucket.MacroExpansionSpec{}
+	// Only update _sync.cas if the pre-compaction doc had already been imported by SGW
```
adamcfraser: Comment needs updated to cover that we're also only updating _mou.cas if the pre-compaction doc had been imported.

db/crud.go
```diff
@@ -295,10 +295,10 @@ func (c *DatabaseCollection) CompactDocChannelHistory(ctx context.Context, docid
 
 	// build macro expansion for sync data. This will avoid the update to xattrs causing an extra import event (i.e. sync cas will be == to doc cas)
 	opts := &sgbucket.MutateInOptions{}
-	spec := []sgbucket.MacroExpansionSpec{
-		sgbucket.NewMacroExpansionSpec(XattrMouCasPath(), sgbucket.MacroCas),
-	}
+	spec := []sgbucket.MacroExpansionSpec{}
```
adamcfraser: Now that we only need `spec` when isSgWrite=true, you should move this and line 304 inside the if clause.

rest/api_test.go
```diff
@@ -4181,6 +4183,8 @@ func TestCompactNonImportedDocWithAutoImport(t *testing.T) {
 	require.NoError(t, err)
 	// History should be compacted
 	assert.Less(t, len(syncData.ChannelSetHistory), len(syncDataBefore.ChannelSetHistory))
+	// CV must be updated by the import triggered during compaction
```
adamcfraser, line 4195: Compaction no longer triggers import (after the latest changes), so this comment is incorrect.  The import occurring at this point is actually an on-demand import triggered by the GetDocSyncData call on line 4182.  However, this means that the test isn't really distinguishing between a bug that would update the CV during compaction, and the on-demand import.  I think you should add a check in between step 8 and 9 that retrieves the document metadata directly (using GetRawDoc) and ensure the cv has *not* changed at that point, and then continue with the verification that you already have to ensure that the on-demand import *does* change the cv.

adamcfraser (changes requested):
Generally looks good - needs a test tweak and a suggestion for coding style.

rest/api_test.go
RIT3shSapata: yes makes sense, missed that part,  changed the way the syncdata is read

db/crud.go
```diff
@@ -221,6 +222,96 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
...
+	}
+
+	doc, err := c.unmarshalDocumentWithXattrs(ctx, key, nil, xattrs, cas, DocUnmarshalSync)
+	if err != nil {
+		return nil, err
+	}
+
+	isSgWrite, crc32Match, _ := doc.IsSGWrite(ctx, rawDoc)
+	if crc32Match {
+		c.dbStats().Database().Crc32MatchCount.Add(1)
```
adamcfraser, line 245: Since we're no longer performing an on-demand import, we shouldn't be incrementing the crc32match stat - that's specifically about tracking cases where an import isn't performed because the body hasn't changed.

rest/api_test.go
```diff
@@ -4179,12 +4179,14 @@ func TestCompactNonImportedDocWithAutoImport(t *testing.T) {
...
 	// Step 9: Verify compaction succeeded and history was removed
-	syncData, err := collection.GetDocSyncData(ctx, nonImportedDocID)
+	require.NoError(t, err)
+	syncData, _, err := collection.GetDocSyncDataNoImport(ctx, nonImportedDocID, db.DocUnmarshalSync)
 	require.NoError(t, err)
 	// History should be compacted
 	assert.Less(t, len(syncData.ChannelSetHistory), len(syncDataBefore.ChannelSetHistory))
 	// CV must be updated by the import triggered during compaction
-	assert.NotEqual(t, cvBeforeCompaction, syncData.CVOrRevTreeID(), "cv should be updated after compaction import")
+	// This assertion should be changed after CBG-5397
```
adamcfraser: Why will this assertion change after CBG-5397?

rest/api_test.go
```diff
@@ -3916,3 +3916,286 @@ func TestFetchBackupRevisionByCVThroughAPI(t *testing.T) {
...
+
+	// Step 8: Call CompactDocChannelHistory - this will trigger the auto-import check
+	// which verifies the document is imported (has valid _sync xattr) before compacting
+	compactedChannels, err := collection.CompactDocChannelHistory(ctx, nonImportedDocID, docSeq-1)
+	require.NoError(t, err)
+	assert.Equal(t, []string{"test_channel"}, compactedChannels)
+
+	// Step 9: Verify compaction succeeded and history was removed
+	require.NoError(t, err)
+	syncData, _, err := collection.GetDocSyncDataNoImport(ctx, nonImportedDocID, db.DocUnmarshalSync)
```
adamcfraser: I made a change to this test locally to check the values in _mou at this point, and that highlighted a problem.  At this point in the test, the value of the mou xattr is:

```
mou: {"cas":"expand","pCas":"0x0000d3ad4ba6b118","pRev":4}
```

Let's discuss offline, I think we need to re-evaluate on-demand import as a requirement for compaction.

adamcfraser (changes requested):

rest/api_test.go
RIT3shSapata: my bad, i got confused and assumed that the compaction would import it too. this is an incorrect comment, will fix it

rest/api_test.go
RIT3shSapata: changed it to perform on demand import

db/crud.go
RIT3shSapata: as discussed offline, we are performing on demand import now

rest/api_test.go
adamcfraser: You should validate the value of mou at this point as well, to guard against the kind of issues we saw on previous commits.  At minimum should validate that _mou.cas is being properly set.  Could use rt.GetSingleDataStore().GetXattrs to fetch without triggering an on-demand import.

adamcfraser: Should also verify that the sync.cas update is working as expected (sync.cas == document.cas)

adamcfraser (changes requested):
Implementation looks good, but I think the test needs to assert that the macro expansion is working as intended, given some of the subtlety there - added a suggestion there.

rest/api_test.go
RIT3shSapata: done

adamcfraser (dismissed):

adamcfraser (approved):

PR #8225, opened 2026-05-05
CBG-5355: fix panic for on demand get when ErrImportCancelled is returned from import

gregns1:
CBG-5355

When `ErrImportCancelled` is encountered in `importDoc` we swallow the error and return silently nil doc but also nil error. This results in us returning nil, nil for `OnDemandImportForGet` inside `GetDocSyncData` which can lead to panic when returning doc.SyncData. 

`GetDocSyncData` is called from `buildRevokedFeed` which when hitting this panic can bring down the whole process via `FatalPanicHandler`. 

## Pre-review checklist

- [ ] Link upstream PRs

- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/570/

### DIFF
--- a/db/crud.go
+++ b/db/crud.go
@@ -193,6 +193,12 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
 			if importErr != nil {
 				return emptySyncData, importErr
 			}
+			// importDoc swallows ErrImportCancelled (e.g. SG purge race), returning
+			// (nil, nil). Treat that the same as not found.
+			if doc == nil {
+				base.DebugfCtx(ctx, base.KeyImport, "Unable to import doc %q during on demand import for get - will be treated as not found.", base.UD(docid))
+				return emptySyncData, base.ErrNotFound
+			}
 		}
 
 		return doc.SyncData, nil
--- a/db/crud.go
+++ b/db/crud.go
@@ -196,7 +196,6 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
 			// importDoc swallows ErrImportCancelled (e.g. SG purge race), returning
 			// (nil, nil). Treat that the same as not found.
 			if doc == nil {
-				base.DebugfCtx(ctx, base.KeyImport, "Unable to import doc %q during on demand import for get - will be treated as not found.", base.UD(docid))
 				return emptySyncData, base.ErrNotFound
 			}
 		}
--- a/db/import_test.go
+++ b/db/import_test.go
@@ -17,6 +17,7 @@ import (
 	"log"
 	"math"
 	"net/http"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


db/crud.go
```diff
@@ -193,6 +193,12 @@ func (c *DatabaseCollection) GetDocSyncData(ctx context.Context, docid string) (
 			if importErr != nil {
 				return emptySyncData, importErr
 			}
+			// importDoc swallows ErrImportCancelled (e.g. SG purge race), returning
+			// (nil, nil). Treat that the same as not found.
+			if doc == nil {
+				base.DebugfCtx(ctx, base.KeyImport, "Unable to import doc %q during on demand import for get - will be treated as not found.", base.UD(docid))
+				return emptySyncData, base.ErrNotFound
+			}
```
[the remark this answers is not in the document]
gregns1: Went for this approach as I feel this is the less risky approach considering this is candidate for backport into maintenance releases. Other callers correctly guard against this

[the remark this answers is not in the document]
torcolvin: Can you make sure we have a followup ticket to fix this? I see that the other case of OnDemandImportForGet is covered but this is a footgun.

I actually don't even think we need to log for this case.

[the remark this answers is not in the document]
gregns1: Filed https://jira.issues.couchbase.com/browse/CBG-5362

db/import_test.go
```diff
@@ -1488,6 +1489,76 @@ func TestImportWithSyncCVAndNoVV(t *testing.T) {
...
+		// UpdateCallback fires AFTER the import callback returns but BEFORE import
+		// commits the tombstone write. Calling SetRaw here resurrects the tombstone as a
+		// live document with no _sync xattr, advancing the CAS. SGW detects the mismatch and retries. On retry the import
+		// callback sees body != nil and _sync RevTreeID == "", satisfying the
+		// isDelete && GetRevTreeID() == "" condition that returns ErrImportCancelled.
+		if key != docID || !importTriggered {
+			return
+		}
+		resurrectOnce.Do(func() {
+			err := docDatastore.SetRaw(key, 0, nil, []byte(`{"foo":"resurrected"}`))
```
torcolvin: SetRaw behaves differently for rosmar and this would get marked as a binary document, I am surprised this fails in the same way, can you verify if this is the case and this doesn't hit https://jira.issues.couchbase.com/browse/CBG-4496?

db/import_test.go
```diff
@@ -1488,6 +1489,76 @@ func TestImportWithSyncCVAndNoVV(t *testing.T) {
...
+	err = docDatastore.Delete(docID)
+	require.NoError(t, err)
+
+	// UpdateCallback now acts only during the import write, not Put.
+	importTriggered = true
+
+	// Ensure GetDocSyncData will handle a nil doc returned from an on-demand import event when ErrImportCancelled is returned.
+	_, err = collection.GetDocSyncData(ctx, docID)
+	require.Error(t, err, "expected an error when import is cancelled mid-flight, not a panic")
+	require.True(t, base.IsDocNotFoundError(err), "expected a document-not-found error after import is cancelled mid-flight, got: %v", err)
```
torcolvin: nit, take or leave:

```suggestion
	base.RequireDocNotFoundError(t, err)
```

db/import_test.go
```diff
@@ -1488,6 +1489,76 @@ func TestImportWithSyncCVAndNoVV(t *testing.T) {
...
+	leakyDataStore.SetUpdateCallback(func(key string) {
+		// UpdateCallback fires AFTER the import callback returns but BEFORE import
+		// commits the tombstone write. Calling SetRaw here resurrects the tombstone as a
+		// live document with no _sync xattr, advancing the CAS. SGW detects the mismatch and retries. On retry the import
+		// callback sees body != nil and _sync RevTreeID == "", satisfying the
+		// isDelete && GetRevTreeID() == "" condition that returns ErrImportCancelled.
+		if key != docID || !importTriggered {
+			return
+		}
+		resurrectOnce.Do(func() {
```
torcolvin: Why does this get hit multiple times and need a sync.Once? Is it better to just have a boolean to say, only resurrect once? I'm OK either way, I am just curious why this is necessary and trying to put myself in the shoes of future me debugging these tests.

Is it worth making sure that this code _does_ get hit with a an assertion that this code does actually get hit, in the case that the write method is changed from `WriteUpdateWithXattrs` to `Update` or something like that? I recently debugged some code where it turned out the callback wasn't getting hit and it was relying on different behavior?

db/import_test.go
```diff
@@ -1488,6 +1489,76 @@ func TestImportWithSyncCVAndNoVV(t *testing.T) {
...
+//  2. The first import callback succeeds and produces an updatedDoc for the
+//     import tombstone write.
+//  3. LeakyDataStore.UpdateCallback fires SetRaw resurrects the tombstone as a live document with
+//     no _sync xattr, advancing the CAS.
+//  4. CAS mismatch detected and retries. On retry it reads the now-live doc
+//     (body != nil, no _sync). The CAS mismatch block in the import callback re-fetches
+//     the body. Execution falls through to isDelete && doc.GetRevTreeID() == "", which fires and returns ErrImportCancelled.
+//  5. importDoc's switch has no return statement in the ErrImportCancelled case, so it
+//     falls through to return docOut, nil with docOut==nil.
+func TestGetDocSyncDataPanicOnImportCancelled(t *testing.T) {
```
torcolvin: I greatly prefer if you don't include the word `Panic` in the test name so it is easier to find a real panic when searching test output.

Do we need to do a similar test for feed import, or do you think this is covered?

torcolvin (dismissed):
Code looks good, just some questions about the tests.

db/import_test.go
gregns1: There is a comment about why this is necessary but a boolean does the same thing and can add an assertion on that being true at end of test to validate the calback was run

db/import_test.go
gregns1: Rosmar still hits this case but have change to use Set in light of this

torcolvin (approved):

PR #8233, opened 2026-05-05
CBG-5175: resync - purge checkpoints if needed

torcolvin:
CBG-5175: resync - purge checkpoints if needed

Create a function to `PurgeShardedDCPCheckpoints` that can be called when necessary. Currently wired up to purge whenever we purged the existing code. I tried this style of purging for attachment migration and compaction and somehow some subtle timing differences surface bugs in the code I don't want to uncover. There's a future jira ticket for this.

It's expected that in Sync Gateway 4.1, EE will use distributed mode and CE will use non distributed mode which is why both pathways are present.

I do not yet purge the checkpoints when the feed completes normally as I do not have the code in cbgt or hooks to determine when that is.

## Pre-review checklist

- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/594/

### DIFF
--- a/base/dcp_sharded_test.go
+++ b/base/dcp_sharded_test.go
@@ -1568,10 +1568,10 @@ func TestCfgNodePollerDistributed(t *testing.T) {
 type mockHeartbeater struct{}
 
 func (m *mockHeartbeater) RegisterListener(listener HeartbeatListener) error { return nil }
-func (m *mockHeartbeater) UnregisterListener(name string)                  {}
-func (m *mockHeartbeater) Start(context.Context) error                      { return nil }
-func (m *mockHeartbeater) StartSendingHeartbeats(context.Context) error       { return nil }
-func (m *mockHeartbeater) StartCheckingHeartbeats(context.Context) error      { return nil }
+func (m *mockHeartbeater) UnregisterListener(name string)                    {}
+func (m *mockHeartbeater) Start(context.Context) error                       { return nil }
+func (m *mockHeartbeater) StartSendingHeartbeats(context.Context) error      { return nil }
+func (m *mockHeartbeater) StartCheckingHeartbeats(context.Context) error     { return nil }
 func (m *mockHeartbeater) Stop(context.Context)                              {}
 
 type mockPIndexImpl struct {
@@ -1745,7 +1745,7 @@ func TestShardedDCPCheckpointCleanup(t *testing.T) {
 	require.True(t, foundAny, "Should have found at least one checkpoint")
 
 	// Purge
-	err = PurgeDCPCheckpoints(ctx, bucket.GetMetadataStore(), checkpointPrefix, "", DCPFeedSharded)
+	err = PurgeDCPCheckpoints(ctx, bucket.GetMetadataStore(), checkpointPrefix, DCPFeedSharded)
 	require.NoError(t, err)
 
 	// Verify gone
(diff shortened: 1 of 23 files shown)


db/background_mgr_resync_dcp.go
```diff
@@ -112,52 +119,37 @@ func (r *ResyncManagerDCP) Init(ctx context.Context, options map[string]any, clu
...
 }
 
+// purgeCheckpoints removes checkpoints for a given resyncID.
+func (r *ResyncManagerDCP) purgeCheckpoints(ctx context.Context, db *Database, resyncID string) error {
+	if resyncID == "" {
+		return errors.New("resyncID is required to delete checkpoints")
+	}
+	checkpointPrefix := GetResyncDCPCheckpointPrefix(db.DatabaseContext, resyncID, r.Distributed)
+	if !r.Distributed {
+		dcpClient, err := r.newDCPClient(ctx, db, false)
```
adamcfraser: newDCPClient seems like it does a decent amount of setup work (including loading and initializing metadata from the checkpoints we're about to delete).  Is this just to handle the differences in purge between Rosmar and Gocb implementations?  

Any way of having an equivalent function to base.PurgeShardedDCPCheckpoints for non-sharded feeds that accepts a MetadataStore and numWorkers?

adamcfraser (changes requested):

base/dcp_sharded_test.go
```diff
@@ -1562,3 +1564,211 @@ func TestCfgNodePollerDistributed(t *testing.T) {
 	})
 
 }
+
+type mockHeartbeater struct{}
```
adamcfraser: Bit of a nitpick, but should this be called something like noopHeartbeater?

adamcfraser: I'd also be fine defining this (as public) in base/heartbeat.go to make it available for other consumers in future.

base/rosmar_dcp_client.go
```diff
@@ -97,8 +97,7 @@ func (dc *RosmarDCPClient) metadataStore() sgbucket.DataStore {
 // PurgeCheckpoints deletes the checkpoint document for the feed. Calling this function while the feed is running
 // will not alter the feed nor remove the checkpoint for the future.
 func (dc *RosmarDCPClient) PurgeCheckpoints() error {
-	checkpoint := dc.opts.CheckpointPrefix + ":" + dc.opts.FeedID
-	err := dc.metadataStore().Delete(checkpoint)
+	err := dc.metadataStore().Delete(dc.opts.CheckpointPrefix)
```
adamcfraser: This feels a bit brittle (if the rosmar implementation changes how it stores checkpoints, this will start failing silently), but I suppose the same can be said for the previous implementation, so not a blocker for this PR.

adamcfraser (dismissed):
Looks good to me.  Just a few nitpicks/style suggestions.  Feel free to ignore and merge if you're happy with it as-is.

base/dcp_sharded_test.go
torcolvin: Yeah this is a better name. To be honest, I had AI write this and didn't review its implementation, but I think it is worthwhile to do so.

base/rosmar_dcp_client.go
torcolvin: That's why I wrote the test (which was broken). I did siphon all the code into one implementation though.

adamcfraser (approved):
LGTM.  Do we need a ticket to track the TODO?

torcolvin:
> LGTM. Do we need a ticket to track the TODO?

Ben's picking this up at the start of next week and asked me not to create more conflicts with his code changes.

PR #8238, opened 2026-05-06
CBG-5291: CAS-safe Update on dual MetadataStore wrapper

bbrks:
CBG-5291

`MetadataStore.Update` previously delegated the whole read-modify-write to the fallback store, and `WriteUpdateWithXattrs` went straight to primary without consulting fallback.
Both now share a wrapper-level CAS retry loop that always lands writes in primary, even when the doc only exists in fallback.

- Check primary for existence; if it holds the doc, directly delegate to the primary data store.
- Otherwise read from fallback, run the callback, and insert into primary with cas=0 to force insertion.
  - On CAS failure, retry via `RetryLoopCas` / `DefaultRetrySleeper`. It's expect that a CAS failure will subsequently find an entry in the primary - since all writes are routed there. Only reads have the fallback.
- Deletes/tombstones on fallback-held docs apply to fallback, because the primary data store never held the doc, so a primary tombstone would just shadow it, instead of actually removing the original.
- `WriteUpdateWithXattrs` short-circuits to primary when `previous.Cas != 0`, since the wrapper reads only return non-zero CAS for primary hits and so can rely on a regular update.

- [x]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/600/

### DIFF
--- a/base/bucket_gocb_test.go
+++ b/base/bucket_gocb_test.go
@@ -2998,9 +2998,11 @@ func TestMetadataStoreUpdateExistingPrimary(t *testing.T) {
 	assert.Equal(t, fallbackBody, fallbackRaw, "fallback is never written by Update")
 }
 
-// TestMetadataStoreUpdateDeleteFallbackOnly verifies the agreed CBG-5291 behaviour: when the
-// callback requests a delete and the doc exists only in fallback, the wrapper is a no-op against
-// primary (no tombstone written) and leaves the fallback copy alone for the migration sweep.
+// TestMetadataStoreUpdateDeleteFallbackOnly verifies the CBG-5291 delete behaviour for a doc
+// that exists only in fallback: the wrapper writes the delete to fallback (the only store
+// that ever held the doc) and does not create a primary tombstone. Without this, callers
+// reading after delete would still observe the doc via the fallback read path until the
+// migration sweep ran.
 func TestMetadataStoreUpdateDeleteFallbackOnly(t *testing.T) {
 	ctx := TestCtx(t)
 	bucket := GetTestBucket(t)
@@ -3015,20 +3017,71 @@ func TestMetadataStoreUpdateDeleteFallbackOnly(t *testing.T) {
 	require.True(t, ok)
 
 	cas, err := metaStore.Update(docID, 0, func(current []byte) (updated []byte, expiry *uint32, isDelete bool, err error) {
+		assert.Equal(t, originalBody, current, "callback must observe fallback value")
 		return nil, nil, true, nil
 	})
 	require.NoError(t, err)
 	assert.Zero(t, cas)
 
-	// Primary stays absent.
+	// Primary stays absent - we did not write a primary tombstone.
 	exists, err := metaStore.Primary().Exists(docID)
 	require.NoError(t, err)
 	assert.False(t, exists)
 
-	// Fallback is untouched.
-	fallbackRaw, _, err := metaStore.Fallback().GetRaw(docID)
+	// Fallback delete landed - subsequent reads must surface not-found.
+	_, _, err = metaStore.Fallback().GetRaw(docID)
+	require.True(t, IsDocNotFoundError(err), "expected fallback delete, got err=%v", err)
+
+	// Wrapper-level read also reports not-found (no read fallthrough to a stale doc).
+	_, _, err = metaStore.GetRaw(docID)
+	require.True(t, IsDocNotFoundError(err), "wrapper read must report not-found after delete, got err=%v", err)
+}
+
+// TestMetadataStoreUpdateRetriesOnCASMismatch exercises the wrapper's CAS retry: the doc starts
+// life in fallback only; from inside the callback (between the wrapper's primary.Exists probe
+// and its primary.WriteCas insert) a concurrent writer lands a doc in primary. The first
+// WriteCas(cas=0) must hit a CAS mismatch and the loop must retry — on retry primary now holds
+// the doc, so the wrapper delegates to primary.Update, the callback fires again against the
+// primary value, and the second result is what finally lands in primary.
+func TestMetadataStoreUpdateRetriesOnCASMismatch(t *testing.T) {
+	ctx := TestCtx(t)
+	bucket := GetTestBucket(t)
+	defer bucket.Close(ctx)
+
+	metaStore := NewMetadataStore(bucket.GetMobileSystemDataStore(), bucket.DefaultDataStore())
+
+	docID := t.Name()
+	originalBody := []byte(`{"src":"fallback"}`)
+	racerBody := []byte(`{"src":"racer"}`)
+
+	ok, err := metaStore.Fallback().Add(docID, 0, originalBody)
 	require.NoError(t, err)
-	assert.Equal(t, originalBody, fallbackRaw)
+	require.True(t, ok)
+
+	var calls int
+	var callbackInputs [][]byte
+	cas, err := metaStore.Update(docID, 0, func(current []byte) (updated []byte, expiry *uint32, isDelete bool, err error) {
+		calls++
+		callbackInputs = append(callbackInputs, append([]byte(nil), current...))
+		if calls == 1 {
+			// Simulate a concurrent writer beating us to the punch on primary.
+			added, addErr := metaStore.Primary().Add(docID, 0, racerBody)
+			require.NoError(t, addErr)
+			require.True(t, added)
+			return []byte(`{"src":"call1"}`), nil, false, nil
+		}
+		return []byte(`{"src":"call2"}`), nil, false, nil
+	})
+	require.NoError(t, err)
+	require.NotZero(t, cas)
+	require.Equal(t, 2, calls, "wrapper must retry through primary.Update after CAS race")
+	assert.Equal(t, originalBody, callbackInputs[0], "first callback observes the fallback value")
+	assert.Equal(t, racerBody, callbackInputs[1], "second callback observes the racer's value from primary")
+
+	primaryRaw, primaryCas, err := metaStore.Primary().GetRaw(docID)
+	require.NoError(t, err)
+	assert.Equal(t, []byte(`{"src":"call2"}`), primaryRaw, "the second callback's result wins")
+	assert.Equal(t, cas, primaryCas)
 }
 
 // TestMetadataStoreUpdateAfterMigrationComplete verifies that once SetMigrationComplete has been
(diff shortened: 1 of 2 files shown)


base/bucket_gocb_test.go
```diff
@@ -2919,6 +2921,441 @@ func TestMetadataStoreKVStoreWriteOperations(t *testing.T) {
...
+	require.NoError(t, err)
+	require.True(t, ok)
+
+	var calls int
+	var callbackInputs [][]byte
+	cas, err := metaStore.Update(docID, 0, func(current []byte) (updated []byte, expiry *uint32, isDelete bool, err error) {
+		calls++
+		callbackInputs = append(callbackInputs, append([]byte(nil), current...))
+		if calls == 1 {
+			// Simulate a concurrent writer beating us to the punch on primary.
```
torcolvin, line 3067: The test that might seem missing is what happens if you have a retry here but you also have a writer that writes to the _secondary_ location for read and you have to update it?

base/bucket_gocb_test.go
```diff
@@ -2919,6 +2921,441 @@ func TestMetadataStoreKVStoreWriteOperations(t *testing.T) {
...
+	primaryRaw, _, err := metaStore.Primary().GetRaw(docID)
+	require.NoError(t, err)
+	assert.Equal(t, updatedBody, primaryRaw)
+
+	fallbackRaw, _, err := metaStore.Fallback().GetRaw(docID)
+	require.NoError(t, err)
+	assert.Equal(t, fallbackBody, fallbackRaw, "fallback is never written by Update")
+}
+
+// TestMetadataStoreUpdateDeleteFallbackOnly verifies the CBG-5291 delete behaviour for a doc
```
torcolvin, line 3001: Anytime you write tombstones it makes me think about resurrections. I think  that the behavior is probably the same, but `Update` can do a resurrection with preserving existing xattrs? I think I'm comfortable the tests as is, but if you aren't, you can write some.

base/dual_metadata_store.go
```diff
@@ -168,9 +169,95 @@ func (ms *MetadataStore) Remove(k string, cas uint64) (casOut uint64, err error)
...
+		writeExp := exp
+		if cbExpiry != nil {
+			writeExp = *cbExpiry
+		}
+
+		if isDelete {
+			// Delete the fallback copy — primary never held it, so a primary tombstone is
+			// pointless and would shadow the fallback for nothing. Remove with the CAS we
+			// just observed; on a concurrent fallback mutation we retry the loop.
+			_, removeErr := ms.fallback.Remove(k, fallbackCas)
```
torcolvin: Do we actually want to Purge this document?

base/dual_metadata_store.go
```diff
@@ -168,9 +169,95 @@ func (ms *MetadataStore) Remove(k string, cas uint64) (casOut uint64, err error)
...
+// When the callback requests a delete on a fallback-only doc, the wrapper applies the delete
+// directly to the fallback store (the only store that ever held the doc). Without this,
+// subsequent reads would still surface the doc via the read-fallback path until the migration
+// sweep ran.
+func (ms *MetadataStore) Update(k string, exp uint32, callback sgbucket.UpdateFunc) (uint64, error) {
+	primaryUpdateFn := func() (uint64, error) {
+		return ms.primary.Update(k, exp, callback)
+	}
+	step := func() (uint64, error) {
+		fallbackValue, fallbackCas, err := ms.fallback.GetRaw(k)
```
torcolvin: This is going to return false for a tombstone, but we don't return false for a tombstone in the general case of Update, is this the desired behavior?

base/dual_metadata_store.go
```diff
@@ -168,9 +169,95 @@ func (ms *MetadataStore) Remove(k string, cas uint64) (casOut uint64, err error)
...
+			_, removeErr := ms.fallback.Remove(k, fallbackCas)
+			switch {
+			case removeErr == nil, IsDocNotFoundError(removeErr):
+				return 0, nil
+			case IsCasMismatch(removeErr):
+				return 0, errRetryUpdateLoop
+			}
+			return 0, removeErr
+		}
+		casOut, writeErr := ms.primary.WriteCas(k, writeExp, 0, newValue, 0)
```
torcolvin, line 232: WriteCas has some undesirable behavior in rosmar until https://github.com/couchbaselabs/rosmar/commit/0013da7629832d7585cc8090cbf9ce33c29583b0 goes in, just make sure to pick this up.

base/dual_metadata_store.go
```diff
@@ -168,9 +169,95 @@ func (ms *MetadataStore) Remove(k string, cas uint64) (casOut uint64, err error)
...
+		if writeErr == nil {
+			return casOut, nil
+		}
+		if IsCasMismatch(writeErr) {
+			return 0, errRetryUpdateLoop
+		}
+		return 0, writeErr
```
torcolvin: The upstream update function returns cas if found and there is an error. I'm not sure if this can ever occur meaningfully but I remember this being a weird quirk of the API. This may be a case where rosmar has divergent behavior from CBS.

base/dual_metadata_store.go
```diff
@@ -168,9 +169,95 @@ func (ms *MetadataStore) Remove(k string, cas uint64) (casOut uint64, err error)
...
+		if IsDocNotFoundError(err) {
+			return primaryUpdateFn()
+		}
+		if err != nil {
+			return 0, err
+		}
+
+		newValue, cbExpiry, isDelete, cbErr := callback(fallbackValue)
+		if cbErr != nil {
+			return 0, cbErr
```
torcolvin: The upstream update function returns cas even if there is an error. I'm nto sure if this is intentional or not, but it might be a really subtle set of bugs to not match behavior.

base/dual_metadata_store.go
```diff
@@ -225,8 +312,80 @@ func (ms *MetadataStore) DeleteWithXattrs(ctx context.Context, k string, xattrKe
...
+		if err != nil && !IsXattrNotFoundError(err) && !errors.Is(err, ErrXattrPartialFound) {
+			return 0, err
+		}
+
+		updatedDoc, cbErr := callback(fallbackBody, fallbackXattrs, 0)
+		if cbErr == ErrCasFailureShouldRetry {
+			return 0, errRetryUpdateLoop
+		}
+		if cbErr != nil {
+			return 0, cbErr
```
torcolvin: Unlike `Update`, this function's parent _does_ return 0 cas.

base/dual_metadata_store.go
```diff
@@ -225,8 +312,80 @@ func (ms *MetadataStore) DeleteWithXattrs(ctx context.Context, k string, xattrKe
...
+			return primaryDelegate()
+		}
+		// A doc with no/partial xattrs is still a fallback hit — surface what we have to the
+		// callback rather than treat it as not-found.
+		if err != nil && !IsXattrNotFoundError(err) && !errors.Is(err, ErrXattrPartialFound) {
+			return 0, err
+		}
+
+		updatedDoc, cbErr := callback(fallbackBody, fallbackXattrs, 0)
+		if cbErr == ErrCasFailureShouldRetry {
```
torcolvin: I would much prefer errors.Is

base/dual_metadata_store.go
```diff
@@ -225,8 +312,80 @@ func (ms *MetadataStore) DeleteWithXattrs(ctx context.Context, k string, xattrKe
...
+		}
+		// xattrsToDelete is intentionally nil on this branch: this is an insert (cas=0) and
+		// the underlying primary store rejects xattrsToDelete on insert
+		// (sgbucket.ErrDeleteXattrOnDocumentInsert). Anything the callback wanted dropped is
+		// already absent from updatedDoc.Xattrs, so it never lands in primary.
+		casOut, writeErr := ms.primary.WriteWithXattrs(ctx, k, writeExp, 0, updatedDoc.Doc, updatedDoc.Xattrs, nil, opts)
+		if writeErr == nil {
+			return casOut, nil
+		}
+		if IsCasMismatch(writeErr) || IsDocNotFoundError(writeErr) {
```
torcolvin, line 381: In the parent function this sets `previousLoopCas`. I don't know why this was necessary but I recall removing it and causing problems.

torcolvin (commented):
There's some definitely odd existing behavior around `Update` and `UpdateWithXattrs` sometimes returning a cas value on error instead of 0. I think this code might have divergence in rosmar/CBS but I think it might be worthwhile to match the upstream behavior.

The other major issue is potentially whether we want to handle tombstone documents. We allow dbconfig to be tombstones in app services and then we will resurrect them. I wonder if we want to be more explicit about tombstone handling and removal. I don't have this part of the code fully loaded in my head, but if you do I'm happy, or at least make sure that we have test coverage for the expected behavior.

base/bucket_gocb_test.go
bbrks: The design of the MetadataStore dual-collection wrapper means that all^ writes only ever go to primary once it is in use.

^caveats:
* Deletes can apply to the fallback which are technically a write... but there's handling in this PR to cover that case by simply retrying the update.
* The timing of all SG nodes actually enabling/using MetadataStore consistently is a problem to be solved with future tickets to close up any gaps where existing SG nodes are writing metadata to the fallback.

By design _most_ Update operations aren't actually 'CAS safe' in the traditional sense, since the caller hasn't actually got a CAS to work with. The read->callback mutation->write steps are the only CAS-safe bit.

base/bucket_gocb_test.go
bbrks: Done - resurrections restore into primary and reinstate the original xattrs

base/dual_metadata_store.go
bbrks: `Purge` meaning what in this context? Removing any system xattrs too?

base/dual_metadata_store.go
bbrks: There is an unbounded `for {` in the datastore implementations. Here at least we are guarded by a bounded retry loop.

base/dual_metadata_store.go
bbrks: Only ever want to return a non-zero CAS if these wrappers are successful by design. Added a lot of comments around this.

The intention is the retry loop will pick up new CAS values each time.

base/dual_metadata_store.go
bbrks: Only ever want to return a non-zero CAS if these wrappers are successful by design. Added a lot of comments around this.

The intention is the retry loop will pick up new CAS values each time.

base/dual_metadata_store.go
bbrks: Only ever want to return a non-zero CAS if these wrappers are successful by design. Added a lot of comments around this.

The intention is the retry loop will pick up new CAS values each time.

base/dual_metadata_store.go
bbrks: Can you re-check this - I'm not quite sure what you mean and some of the improved naming/comments may help answer this.

torcolvin (approved):

PR #8240, opened 2026-05-07
CBG-5290: Wire up ctx into sg-bucket apis

bbrks:
CBG-5290

Wire up context into sg-bucket APIs

## Dependencies
- [x] https://github.com/couchbase/sg-bucket/pull/144
- [x] https://github.com/couchbaselabs/rosmar/pull/72

- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/585/

### DIFF
--- a/base/collection.go
+++ b/base/collection.go
@@ -209,7 +209,7 @@ func (b *GocbV2Bucket) GetName() string {
 	return b.bucket.Name()
 }
 
-func (b *GocbV2Bucket) UUID() (string, error) {
+func (b *GocbV2Bucket) UUID(_ context.Context) (string, error) {
 	config, configErr := b.getConfigSnapshot()
 	if configErr != nil {
 		return "", fmt.Errorf("Unable to determine bucket UUID for collection %v: %w", b.GetName(), configErr)
@@ -314,7 +314,7 @@ func (b *GocbV2Bucket) GetStatsVbSeqno(maxVbno uint16, useAbsHighSeqNo bool) (uu
 	return GetStatsVbSeqno(genericStats, maxVbno, useAbsHighSeqNo)
 }
 
-func (b *GocbV2Bucket) GetMaxVbno() (uint16, error) {
+func (b *GocbV2Bucket) GetMaxVbno(ctx context.Context) (uint16, error) {
 
 	config, configErr := b.getConfigSnapshot()
 	if configErr != nil {
@@ -358,7 +358,7 @@ func (b *GocbV2Bucket) GetCCVSettings(ctx context.Context) (ccvEnabled bool, max
 		return false, nil, nil
 	}
 
-	numVBuckets, err := b.GetMaxVbno()
+	numVBuckets, err := b.GetMaxVbno(ctx)
 	if err != nil {
 		return false, nil, fmt.Errorf("error getting vbucket count: %v", err)
 	}
@@ -455,13 +455,13 @@ func (b *GocbV2Bucket) Flush(ctx context.Context) error {
 // BucketItemCount first tries to retrieve an accurate bucket count via N1QL,
 // but falls back to the REST API if that cannot be done (when there's no index to count all items in a bucket)
 func (b *GocbV2Bucket) BucketItemCount(ctx context.Context) (itemCount int, err error) {
-	dataStoreNames, err := b.ListDataStores()
+	dataStoreNames, err := b.ListDataStores(ctx)
 	if err != nil {
 		return 0, err
 	}
 
 	for _, dsn := range dataStoreNames {
-		ds, err := b.NamedDataStore(dsn)
+		ds, err := b.NamedDataStore(ctx, dsn)
 		if err != nil {
 			return 0, err
 		}
@@ -655,7 +655,7 @@ func (b *GocbV2Bucket) releaseQueryOp() {
 	<-b.queryOps
 }
 
-func (b *GocbV2Bucket) ListDataStores() ([]sgbucket.DataStoreName, error) {
+func (b *GocbV2Bucket) ListDataStores(_ context.Context) ([]sgbucket.DataStoreName, error) {
 	if !b.IsSupported(sgbucket.BucketStoreFeatureCollections) {
 		return []sgbucket.DataStoreName{ScopeAndCollectionName{Scope: DefaultScope, Collection: DefaultCollection}}, nil
 	}
@@ -678,7 +678,7 @@ func (b *GocbV2Bucket) ListDataStores() ([]sgbucket.DataStoreName, error) {
 }
 
 // DropDataStore removes a collection from the bucket. This function will return immediately but the collection may take some time to delete.
-func (b *GocbV2Bucket) DropDataStore(name sgbucket.DataStoreName) error {
+func (b *GocbV2Bucket) DropDataStore(ctx context.Context, name sgbucket.DataStoreName) error {
 	if b.cluster == nil {
 		return fmt.Errorf("bucket %s has been closed", MD(b.GetName()))
 	}
@@ -710,7 +710,7 @@ func (b *GocbV2Bucket) CreateDataStore(ctx context.Context, name sgbucket.DataSt
 }
 
 // DefaultDataStore returns the default collection for the bucket.
-func (b *GocbV2Bucket) DefaultDataStore() sgbucket.DataStore {
+func (b *GocbV2Bucket) DefaultDataStore(_ context.Context) sgbucket.DataStore {
 	return &Collection{
 		Bucket:     b,
 		Collection: b.bucket.DefaultCollection(),
@@ -718,7 +718,7 @@ func (b *GocbV2Bucket) DefaultDataStore() sgbucket.DataStore {
 }
 
 // NamedDataStore returns a collection on a bucket within the given scope and collection.
-func (b *GocbV2Bucket) NamedDataStore(name sgbucket.DataStoreName) (sgbucket.DataStore, error) {
+func (b *GocbV2Bucket) NamedDataStore(_ context.Context, name sgbucket.DataStoreName) (sgbucket.DataStore, error) {
 	c, err := NewCollection(
 		b,
 		b.bucket.Scope(name.ScopeName()).Collection(name.CollectionName()))
--- a/base/constants_syncdocs.go
+++ b/base/constants_syncdocs.go
@@ -422,16 +422,16 @@ type SyncInfo struct {
 func InitSyncInfo(ctx context.Context, ds DataStore, metadataID string) (requiresResync bool, requiresAttachmentMigration bool, err error) {
 
 	var syncInfo SyncInfo
-	_, fetchErr := ds.Get(SGSyncInfo, &syncInfo)
+	_, fetchErr := ds.Get(ctx, SGSyncInfo, &syncInfo)
 	if IsDocNotFoundError(fetchErr) {
 		if metadataID == "" {
 			return false, true, nil
 		}
 		newSyncInfo := &SyncInfo{MetadataID: Ptr(metadataID)}
-		_, addErr := ds.Add(SGSyncInfo, 0, newSyncInfo)
+		_, addErr := ds.Add(ctx, SGSyncInfo, 0, newSyncInfo)
 		if IsCasMismatch(addErr) {
 			// attempt new fetch
-			_, fetchErr = ds.Get(SGSyncInfo, &syncInfo)
+			_, fetchErr = ds.Get(ctx, SGSyncInfo, &syncInfo)
 			if fetchErr != nil {
 				return true, true, fmt.Errorf("Error retrieving syncInfo (after failed add): %v", fetchErr)
 			}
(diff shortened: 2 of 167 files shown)


base/collection.go
```diff
@@ -678,7 +678,7 @@ func (b *GocbV2Bucket) ListDataStores() ([]sgbucket.DataStoreName, error) {
 }
 
 // DropDataStore removes a collection from the bucket. This function will return immediately but the collection may take some time to delete.
-func (b *GocbV2Bucket) DropDataStore(name sgbucket.DataStoreName) error {
+func (b *GocbV2Bucket) DropDataStore(ctx context.Context, name sgbucket.DataStoreName) error {
 	if b.cluster == nil {
 		return fmt.Errorf("bucket %s has been closed", MD(b.GetName()))
```
torcolvin, line 683: out of scope and not very important since we don't redact MD but this should be RedactErrorf, we could write a linter or code to find fmt.Errorf with MD/UD

base/constants_syncdocs.go
```diff
@@ -468,13 +468,13 @@ func (s *SyncInfo) requiresResync(metadataID string) bool {
...
 // SetSyncInfoMetadataID sets syncInfo in a DataStore to the specified metadataID, preserving metadata version if present
-func SetSyncInfoMetadataID(ds DataStore, metadataID string) error {
+func SetSyncInfoMetadataID(ctx context.Context, ds DataStore, metadataID string) error {
 
 	// If the metadataID isn't defined, don't persist SyncInfo.  Defensive handling for legacy use cases.
 	if metadataID == "" {
 		return nil
 	}
-	_, err := ds.Update(SGSyncInfo, 0, func(current []byte) (updated []byte, expiry *uint32, delete bool, err error) {
+	_, err := ds.Update(ctx, SGSyncInfo, 0, func(_ context.Context, current []byte) (updated []byte, expiry *uint32, delete bool, err error) {
```
torcolvin: This is an interesting API. In this case we aren't using context variable but I think you possibly _do_ want to shadow ctx through this and not use the parent?

That is replace `_ context.Context` with `ctx context.Context`. I suspect these are always going to be the same but there's something that feels a bit weird about this callback API specifically.

db/attachment_compaction.go
```diff
@@ -311,7 +311,7 @@ func attachmentCompactSweepPhase(ctx context.Context, dataStore base.DataStore,
 	// Iterate over v1 attachments and if not marked with supplied compactionID we can purge the attachments.
 	// In the event of an error we can return but continue - Worst case is an attachment which should be deleted won't
 	// be deleted.
-	callback := func(event sgbucket.FeedEvent) bool {
+	callback := func(_ context.Context, event sgbucket.FeedEvent) bool {
```
torcolvin: I have definitely wanted this before to have this take a ctx method _but_ I have misgivings about not showing the ctx here. I don't think we need to change this code here necessarily.

This might just be a prior bad code smell where this is complicated nested function.

I'm curious if you had thoughts about what to do in short or long term. If we are never using to use this second context in future code, then maybe it is better to not have it?

db/background_mgr.go
```diff
@@ -582,7 +582,7 @@ func (b *BackgroundManager) UpdateSingleNodeClusterAwareStatus(ctx context.Conte
 // updateMultiNodeClusterAwareStatus updates the cluster status document with the current local status. If the bucket status is in a stopping / stopped / completed / error state but the local status is running, then this method will not update the bucket status and instead return. The caller is responsible for taking appropriate action.
 func (b *BackgroundManager) updateMultiNodeClusterAwareStatus(ctx context.Context) error {
 	docID := b.clusterAwareOptions.StatusDocID()
-	_, err := b.clusterAwareOptions.metadataStore.Update(docID, 0, func(current []byte) ([]byte, *uint32, bool, error) {
+	_, err := b.clusterAwareOptions.metadataStore.Update(ctx, docID, 0, func(_ context.Context, current []byte) ([]byte, *uint32, bool, error) {
```
torcolvin: This is new code you can do func(ctx context.Context) here.

db/background_mgr_resync_dcp_test.go
```diff
@@ -408,7 +408,7 @@ func TestResyncManagerDCPResumeStoppedProcessChangeCollections(t *testing.T) {
 		require.NotNil(t, col)
 
 		// required to avoid missing audit fields in PUT
-		ctx := col.AddCollectionContext(ctx)
+		ctx = col.AddCollectionContext(ctx)
```
torcolvin: This is intentionally ctx := so that the collectio name doesn't leak outside of the for loop

db/change_listener.go
```diff
@@ -152,7 +152,7 @@ func (listener *changeListener) DocumentType(key []byte) DocumentType {
 
 // ProcessFeedEvent is invoked for each mutate or delete event seen on the server's mutation feed (TAP or DCP).  Uses document
 // key to determine handling, based on whether the incoming mutation is an internal Sync Gateway document.
-func (listener *changeListener) ProcessFeedEvent(event sgbucket.FeedEvent) bool {
+func (listener *changeListener) ProcessFeedEvent(_ context.Context, event sgbucket.FeedEvent) bool {
```
torcolvin: I don't think we need to change this code now, but I think this is an interesting case of being able to pass context, and not have `listener.ctx` as a struct variable

db/import_listener.go
```diff
@@ -119,7 +119,7 @@ func (il *importListener) StartImportFeed(dbContext *DatabaseContext) (err error
 // executed concurrently for multiple events from different vbuckets.  Filters out
 // internal documents based on key, then checks sync metadata to determine whether document needs to be imported.
 // Returns true if the checkpoints should be persisted.
-func (il *importListener) ProcessFeedEvent(event sgbucket.FeedEvent) bool {
+func (il *importListener) ProcessFeedEvent(_ context.Context, event sgbucket.FeedEvent) bool {
```
torcolvin: Another point where I think it would be interesting to be able to use passed ctx vs il.loggingCtx

db/indexes.go
```diff
@@ -559,14 +559,14 @@ func GetOnlinePrincipalIndexes(ctx context.Context, collection base.N1QLStore, u
...
+		dataStore, err := bucket.NamedDataStore(ctx, dsName)
 		if err != nil {
 			errs = errs.Append(fmt.Errorf("failed to get datastore %s: %w", base.MD(dsName), err))
 			continue
 		}
-		n1qlStore, ok := dsName.(base.N1QLStore)
+		n1qlStore, ok := dataStore.(base.N1QLStore)
 		if !ok {
-			errs = errs.Append(fmt.Errorf("datastore %s(%T) is not a N1QLStore", base.MD(dsName), dsName))
+			errs = errs.Append(fmt.Errorf("datastore %s(%T) is not a N1QLStore", base.MD(dsName), dataStore))
```
torcolvin: This needs to be RedactErrof

db/util_testing.go
```diff
@@ -199,7 +199,7 @@ func purgeWithDCPFeed(ctx context.Context, bucket base.Bucket, tbp *base.TestBuc
 
 	}
 
-	purgeCallback := func(event sgbucket.FeedEvent) bool {
```
torcolvin, line 202: given this is test code, I'd be inclined to use ctx context.Context here

rest/config_test.go
```diff
@@ -1423,31 +1423,32 @@ func deleteTempFile(t *testing.T, file *os.File) {
...
 	// setting just a log key should enable logging
 	config.Logging.Console = &base.ConsoleLoggerConfig{LogKeys: []string{"CRUD"}}
-	require.NoError(t, config.SetupAndValidateLogging(base.TestCtx(t)))
+	require.NoError(t, config.SetupAndValidateLogging(ctx))
 	assert.Equal(t, base.LevelInfo, *base.ConsoleLogLevel())
 	assert.Equal(t, []string{"CRUD"}, base.ConsoleLogKey().EnabledLogKeys())
 }
 
 func TestSetupServerContext(t *testing.T) {
+	ctx := base.TestCtx(t)
```
torcolvin: I think this is now in the wrong place, it should be in t.Run

rest/persistent_config_test.go
```diff
@@ -116,6 +117,7 @@ func TestAutomaticConfigUpgrade(t *testing.T) {
 }
 
 func TestAutomaticConfigUpgradeError(t *testing.T) {
+	ctx := base.TestCtx(t)
```
torcolvin: This also is now in the wrong scope of subtests

rest/server_context_test.go
```diff
@@ -436,6 +436,7 @@ func TestTLSSkipVerifyCombinations(t *testing.T) {
 // CBG-1518 - Test GetbucketSpec() ServerTLSSkipVerify and empty CA Cert behaviour.
 // Does not test validation of having CA Cert provided and TLS Skip verify on. See TestTLSSkipVerifyCombinations for that.
 func TestTLSSkipVerifyGetBucketSpec(t *testing.T) {
+	ctx := base.TestCtx(t)
```
torcolvin: This is in the wrong scope for subtests

rest/server_context_test.go
```diff
@@ -612,6 +612,7 @@ func TestServerContextSetupCollectionsSupport(t *testing.T) {
 func TestLogFlush(t *testing.T) {
 	// FIXME: CBG-1869 flaky test
 	t.Skip("CBG-1869: Flaky test")
+	ctx := base.TestCtx(t)
```
torcolvin: Wrong scope for subtests

rest/server_context_test.go
```diff
@@ -759,7 +759,6 @@ func TestDisableScopesInLegacyConfig(t *testing.T) {
 		for _, scopes := range []bool{false, true} {
 			t.Run(fmt.Sprintf("persistent_config=%t", persistentConfig), func(t *testing.T) {
 
-				ctx := base.TestCtx(t)
```
torcolvin, line 762: This is intentionally shadowing ctx here to mark which subtest this belongs to

rest/adminapitest/admin_api_test.go
```diff
@@ -3835,6 +3837,7 @@ func setServerPurgeInterval(t *testing.T, rt *rest.RestTester, newPurgeInterval
 }
 
 func TestTombstoneCompactionPurgeInterval(t *testing.T) {
+	ctx := base.TestCtx(t)
```
torcolvin: this is just unused and then shadowed?

xdcr/rosmar_xdcr.go
```diff
@@ -262,7 +262,7 @@ func (r *rosmarManager) Start(ctx context.Context) error {
 			break
 		}
 	}
-	callback := func(event sgbucket.FeedEvent) bool {
+	callback := func(_ context.Context, event sgbucket.FeedEvent) bool {
```
torcolvin: Given this is test code, I feel fine about have ctx used from the callback.

rest/view_api_test.go
```diff
@@ -671,6 +672,7 @@ func TestViewQueryWrappers(t *testing.T) {
 
 func TestViewQueryWithXattrAndNonXattr(t *testing.T) {
 
+	ctx := base.TestCtx(t)
```
torcolvin, line 675: I'm often inclined to move this below where RestTester is configured to show both test and database, but this is fine, and I haven't commented everywhere I saw this.

rest/user_api_test.go
```diff
@@ -1072,6 +1070,7 @@ func TestFunkyUsernames(t *testing.T) {
 	}
 }
 func TestRemovingUserXattr(t *testing.T) {
+	ctx := base.TestCtx(t)
```
torcolvin: This is unused and only used inside of the testCases where it is a different ctx.

rest/server_context_test.go
```diff
@@ -898,6 +897,7 @@ func TestCompactIntervalFromConfig(t *testing.T) {
 }
 
 func TestHeapProfileValuesPopulated(t *testing.T) {
+	ctx := base.TestCtx(t)
```
torcolvin: this now has ctx not scoped to subtests

rest/server_context_test.go
```diff
@@ -853,6 +853,7 @@ func TestOfflineDatabaseStartup(t *testing.T) {
 }
 
 func TestCompactIntervalFromConfig(t *testing.T) {
+	ctx := base.TestCtx(t)
```
torcolvin: this dropped ctx scoped to subtest

torcolvin (commented):
There are three things that I see.

- `Update` is changed to pass a context and then it is always ignored. If we always want to ignore it, then I don't think we should have this be an argument at all. If we want to consider its usage in the future and not modify existing code at this time, that's OK too.
- Similar to `Update` the `FeedEventFunc` is changed to pass a context and then it is always ignored. If we always want to ignore it, then I don't think we should have this be an argument at all. If we want to consider its usage in the future and not modify existing code at this time, that's OK too.
- There are a bunch of places in the test code where ctx := base.TestCtx(t) was moved from the subtest to the parent test. I did not exhaustively search test code.

db/attachment_compaction.go
bbrks: I'm leaning towards the callback function not needing ctx.

Update-specific code can log using `ctx` passed, and then the caller can already include their own log context as a closure that isn't passed as a parameter.

db/background_mgr_resync_dcp_test.go
torcolvin: This comment is still relevant, but it is test only code.

db/query_dual_metadata_test.go
```diff
@@ -114,18 +114,19 @@ func TestMergedDedupIteratorFallbackOnly(t *testing.T) {
...
 	assert.Equal(t, "_sync:user:carol", rows[0].Id)
 	assert.Equal(t, "_sync:role:editor", rows[1].Id)
 }
 
 // TestMergedDedupIteratorEmpty verifies that empty iterators are handled cleanly.
 func TestMergedDedupIteratorEmpty(t *testing.T) {
+	ctx := base.TestCtx(t)
 	iter := newDualMetadataStorePrincipalDedupIterator[PrincipalRow](newMockIterator(nil), newMockIterator(nil), 0)
 	var row PrincipalRow
 	assert.False(t, iter.Next(context.Background(), &row))
```
torcolvin:
```suggestion
	assert.False(t, iter.Next(ctx, &row))
```

torcolvin (dismissed):
Two quick test fixes, not very important. I'll re-approve when upstream changes get merged.

torcolvin (approved):

PR #8242, opened 2026-05-07
CBG-5177 allow stats to be collated for distributed background manager

torcolvin:
CBG-5177 allow stats to be collated for distributed background manager

## Pre-review checklist

- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/595/


db/background_mgr.go
```diff
@@ -106,9 +106,19 @@ type BackgroundManagerStatus struct {
 // BackgroundManagerProcessI is an interface satisfied by any of the background processes
 // Examples of this: ReSync, Compaction, Attachment Migration
 type BackgroundManagerProcessI interface {
+	// Init is called before Run for setup purposes. If Init errors, Run will not happen.
 	Init(ctx context.Context, options map[string]any, clusterStatus []byte) error
+	// Run implements all of the work of the process.
 	Run(ctx context.Context, options map[string]any, persistClusterStatusCallback updateStatusCallbackFunc, terminator *base.SafeTerminator) error
-	GetProcessStatus(status BackgroundManagerStatus) (statusOut []byte, meta []byte, err error)
+	// GetProcessStatus passes the BackgroundManagerStatus, the previousStatus that was serialized. previousStatus is
```
```suggestion
	// GetProcessStatus accepts the current BackgroundManagerStatus and the previousStatus that was serialized. previousStatus is
```

db/background_mgr_resync_dcp.go
```diff
@@ -448,18 +449,66 @@ type ResyncManagerResponseDCP struct {
 	CollectionsProcessing map[string][]string `json:"collections_processing,omitempty"`
 }
 
-func (r *ResyncManagerDCP) GetProcessStatus(status BackgroundManagerStatus) ([]byte, []byte, error) {
+// resyncStats is a part of the ResyncManagerResponseDCP struct that only contains the stats fields for efficincy.
+type resyncStats struct {
```
adamcfraser, line 482: Should we embed this type in ResyncManagerResponseDCP to avoid future divergence with the versions of the properties there?

db/background_mgr_resync_dcp.go
```diff
@@ -28,12 +28,14 @@ import (
...
-	DocsProcessed       base.AtomicInt
-	DocsChanged         base.AtomicInt
-	ResyncID            string
-	VBUUIDs             []uint64
-	useXattrs           bool
-	ResyncedCollections base.CollectionNames
+	docsProcessedSerialized      atomic.Int64
+	docsChangedSerialized        atomic.Int64
+	docsChangedSinceLastUpdate   atomic.Int64
+	docsProcessedSinceLastUpdate atomic.Int64
```
adamcfraser: I think it would make the stats more intuitive with just a little rearrangement, to:
- docsProcessedLocal
- docsProcessedLocalSerialized

This really just shuffles around the arithmetic, but I think it makes it a lot easier to think about:
- docsProcessedLocal is the total number of docs processed by this node
- docsProcessedLocalSerialized is the total number of docs processed by this node that have been persisted to status

Under this arrangement, GetProcessStatus would incrementing the previous stats with whatever hadn't been serialized (docsProcessedLocal - docsProcessedLocalSerialized), and SetProcessStatus would just be incrementing docProcessedLocalSerialized by ( newStats.DocsProcessed - previousStats.DocsProcessed).

What do you think?

adamcfraser (changes requested):
Overall approach looks fine given the current limitations of BackgroundManager status.  Main suggestion is to switch around the stats a bit to make them more intuitive, without really changing the overall approach.

db/background_mgr_resync_dcp.go
torcolvin: I'm not opposed to changing the name, but the reason I did it this way was so that a user didn't accidentally use `ResyncManager.docsProcessed` when they need to call `ResyncManager.DocsProcessed()` - this code might work OK unless you tested this in a distributed system.

I might not understand this comment.

adamcfraser: As discussed, this would require an additional stat in order to keep track of the cross-node 'docsProcessed' for logging at the completion of the job.

torcolvin: I think I changed this to be what you requested.

db/background_mgr_resync_dcp.go
```diff
@@ -457,24 +462,75 @@ func (r *ResyncManagerDCP) SetCollectionStatus(collectionNames base.CollectionNa
...
+	}
+
 	response := ResyncManagerResponseDCP{
 		BackgroundManagerStatus: status,
 		ResyncID:                r.ResyncID,
-		DocsChanged:             r.DocsChanged.Value(),
-		DocsProcessed:           r.DocsProcessed.Value(),
-		CollectionsProcessing:   r.ResyncedCollections,
+		resyncStats: resyncStats{
+			DocsChanged:   previousStats.DocsChanged + (r.docsChangedLocalSerialized.Load() - r.docsChangedLocal.Load()),
```
adamcfraser: Shouldn't this be:
previousStats.DocsChanged +
(r.docsChangedLocal.Load() - r.docsChangedLocalSerialized.Load())

(i.e. previous changed plus whatever hasn't been serialized on this node)

db/background_mgr_resync_dcp.go
```diff
@@ -457,24 +462,75 @@ func (r *ResyncManagerDCP) SetCollectionStatus(collectionNames base.CollectionNa
...
+			DocsProcessed: previousStats.DocsProcessed + (r.docsProcessedLocalSerialized.Load() + r.docsProcessedLocal.Load()),
```
adamcfraser: Same as above?

adamcfraser (changes requested):
I like the new approach, thanks for switching.  One bit about the GetProcessStats handling that doesn't look exactly right?

torcolvin:
I think the counting is correct now (as per tests that were written), but I'd appreciate if you can double check this.

adamcfraser (commented):
LGTM

adamcfraser (approved):

PR #8245, opened 2026-05-08
CBG-5330: Add function to compact user channel history

RIT3shSapata:
CBG-5330

Describe your PR here...
- Added function to compact the channel history based

## Pre-review checklist

- [ ] Link upstream PRs

- [x]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/598/
- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/612/

### DIFF
--- a/auth/role.go
+++ b/auth/role.go
@@ -54,13 +54,11 @@ func (timedSet TimedSetHistory) PruneHistory(partitionWindow time.Duration) []st
 }
 
 func (timedSet TimedSetHistory) CompactChannelHistory(channel string) bool {
-	for chanName := range timedSet {
-		if chanName == channel {
-			delete(timedSet, chanName)
-			return true
-		}
+	if _, ok := timedSet[channel]; !ok {
+		return false
 	}
-	return false
+	delete(timedSet, channel)
+	return true
 }
 
 type GrantHistory struct {
--- a/auth/role.go
+++ b/auth/role.go
@@ -53,8 +53,8 @@ func (timedSet TimedSetHistory) PruneHistory(partitionWindow time.Duration) []st
 	return prunedChannelHistory
 }
 
-func (timedSet TimedSetHistory) CompactChannelHistory(channel string) bool{
-	for chanName, _ := range timedSet {
+func (timedSet TimedSetHistory) CompactChannelHistory(channel string) bool {
+	for chanName := range timedSet {
 		if chanName == channel {
 			delete(timedSet, chanName)
 			return true
--- a/auth/role.go
+++ b/auth/role.go
@@ -53,6 +53,16 @@ func (timedSet TimedSetHistory) PruneHistory(partitionWindow time.Duration) []st
 	return prunedChannelHistory
 }
 
+func (timedSet TimedSetHistory) CompactChannelHistory(channel string) bool{
+	for chanName, _ := range timedSet {
+		if chanName == channel {
+			delete(timedSet, chanName)
+			return true
+		}
+	}
+	return false
+}
+
 type GrantHistory struct {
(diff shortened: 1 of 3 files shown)


auth/role.go
```diff
@@ -53,6 +53,14 @@ func (timedSet TimedSetHistory) PruneHistory(partitionWindow time.Duration) []st
 	return prunedChannelHistory
 }
 
+func (timedSet TimedSetHistory) CompactChannelHistory(channel string) bool {
```
adamcfraser: I agree with the Copilot comment - this is actually a more generic operation on TimedSetHistory that adds information to the delete operation about whether an element was actually removed.  On its own I don't know if that API is that useful - I think a better approach would be to follow the pattern used by the existing TimedSetHistory.PruneHistory - write a custom method to remove entries from the map based on key, moving some of the functionality you currently have in CompactChannelHistory.  

Something like
func (timedSet TimedSetHistory) PruneHistoryByKey(keys []string) removedKeys []string

auth/user.go
```diff
@@ -750,6 +750,20 @@ func (user *userImpl) GetAddedChannels(channels ch.TimedSet) (base.Set, error) {
 	return output, nil
 }
 
+func (user *userImpl) CompactChannelHistory(scope, collection string, channels []string) []string {
```
adamcfraser, line 753: As described above, propose moving a lot of this functionality down into TimedSetHistory.  That should also remove the need to set the data back to the set on line 762 - you can just delete in place.

auth/user_test.go
```diff
@@ -475,6 +475,84 @@ func TestUserKeysHash(t *testing.T) {
...
+	base.SetUpTestLogging(t, base.LevelDebug, base.KeyAuth)
+
+	ctx := base.TestCtx(t)
+	testBucket := base.GetTestBucket(t)
+	defer testBucket.Close(ctx)
+	dataStore := testBucket.GetSingleDataStore()
+
+	const (
+		scope      = "scope1"
+		collection = "collection1"
```
adamcfraser, line 488: I think this could use some additional testing for cases where the user has channels defined for multiple collections, and verifying error handling for scope/collection mismatches.

auth/user_test.go
```diff
@@ -475,6 +475,84 @@ func TestUserKeysHash(t *testing.T) {
...
+		assert.Empty(t, compacted)
+
+		afterHistory := u.CollectionChannelHistory(scope, collection)
+		require.Len(t, afterHistory, 3, "all three channels should remain in history")
+		_, ch1Present := afterHistory["ch1"]
+		_, ch2Present := afterHistory["ch2"]
+		_, ch3Present := afterHistory["ch3"]
+		assert.True(t, ch1Present)
+		assert.True(t, ch2Present)
+		assert.True(t, ch3Present)
```
adamcfraser, line 552: Other scenarios you'll want to ensure work as expected:
- mix of existing and non-existing channels
- duplicate channel names in the input to CompactChannelHistory
- using the default collection

adamcfraser (changes requested):

auth/user.go
```diff
@@ -750,6 +750,12 @@ func (user *userImpl) GetAddedChannels(channels ch.TimedSet) (base.Set, error) {
...
+	chanHistory := user.CollectionChannelHistory(scope, collection)
```
adamcfraser, line 754: Do you need a nil check here in case the user doesn't have history for the scope/collection?

adamcfraser (dismissed):

auth/user.go
RIT3shSapata: added the nil check

bbrks (approved):
LGTM - let's merge this and then get the additional testing around real replication revocation behaviour once wired up to the REST API

PR #8259, opened 2026-05-12
CBG-5298: Cleanup for GetUser

bbrks:
CBG-5298

- [x]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/615/

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -719,6 +719,9 @@ func (auth *Authenticator) casUpdatePrincipal(p Principal, callback casUpdatePri
 		if err != nil {
 			return fmt.Errorf("Error reloading principal after CAS failure: %w", err)
 		}
+		if p == nil {
+			return base.ErrNotFound
+		}
 	}
 	base.InfofCtx(auth.LogCtx, base.KeyAuth, "Unable to update principal after %d attempts.  Principal:%s Error:%v", PrincipalUpdateMaxCasRetries, base.UD(p.Name()), err)
 	return err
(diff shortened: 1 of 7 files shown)


auth/session.go
```diff
@@ -72,8 +72,7 @@ func (auth *Authenticator) AuthenticateCookie(rq *http.Request, response http.Re
 	if err != nil {
 		return nil, err
 	}
-
-	if session.SessionUUID != user.GetSessionUUID() {
+	if user == nil || session.SessionUUID != user.GetSessionUUID() {
 		base.InfofCtx(auth.LogCtx, base.KeyAuth, "Session no longer valid for user %s", base.UD(session.Username))
```
gregns1, line 76: Take or leave this: would it be useful to have more specific logging here for user == nil case for debugging purposes?

auth/auth.go
```diff
@@ -964,7 +967,14 @@ func (auth *Authenticator) RegisterNewUser(username, email string) (User, error)
 
 	err = auth.Save(user)
 	if base.IsCasMismatch(err) {
-		return auth.GetUser(username)
+		existing, getErr := auth.GetUser(username)
+		if getErr != nil {
+			return nil, getErr
+		}
+		if existing == nil {
+			return nil, err
```
gregns1, line 975: Take or leave: returning cas mismatch error here which I think is fine but wonder your opinion on new error with better context on it.

gregns1 (approved):
Generally looks good to me, couple of comments you can take or leave.

PR #8266, opened 2026-05-12
CBG-5184 use cbgt one shot mode for distributed resync

torcolvin:
CBG-5184 use cbgt one shot mode

This does not yet work in multi node mode (expected) and all tests are not enabled. I'm working on why some of the tests fail or flake. I do not believe there will be major changes from the failing tests, so I am putting this up for review as is. Most tests do pass in distributed resync mode (where distributed resync is only run on a single node).

- Register high sequence numbers `endSeqNos` with cbgt via `sourceParams`. This marks when cbgt will stop processing documents.
- when specifying a high sequence number to kv, it will always stream sequences to the end of the snapshot, which might be higher than the specified endSeqNo. Avoid checkpointing these sequence numbers in `DCPCommon.updateSeq` to avoid a ERANGE error when re-opening the feed. https://github.com/couchbase/cbgt/blob/6e34ff79a0e6e97d5fb5b548c217a8a9d279a94b/feed_dcp_gocbcore.go#L1034
- Add `ForceCheckpointWrite` when the feed shuts down to force the checkpoints to be written. This is really only helpful for tests, but `DCPCommon.setMetaData` will only be called on a 1 minute interval, so the last checkpoints are not set. For a test that makes sure that some documents are processed, and then resumed from that point, this test fails without this call.
- `cbgt.EventHandler.OnUnregisterFeed` is now used to discover when vBuckets are completed. This is tracked locally (for now, see TODO) and used to close `dcpDoneChan` to indicate when the DCP operations should finish. There are N feeds for N partitions used by cbgt. This function is called whenever a feed (set of vBuckets for a partition) completes. This is true on reaching the endSeqNos, or when unregistering a feed (normal shutdown, such as `/db/resync?action=stop`)

Small fixes:

- Created an options struct for `DCPDestOptions` to avoid having a another nil-able optino.
- cbgt changed a type from string to `ConsistencyLevel`, this is type alias.
- Allow `dbExpVarsStat` to be nil, especially in tests. I do not have a place for this stat yet in distributed resync.

## Pre-review checklist

- [ ]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/702/

### DIFF
--- a/base/dcp_common.go
+++ b/base/dcp_common.go
@@ -244,7 +244,7 @@ func (c *DCPCommon) shouldProcessSequence(vBucketID uint16, seq uint64) bool {
 // vbucket and log a warning in that case.  The valid case for setting warnOnLowerSeqNo to
 // false is when it's a rollback scenario.  See https://github.com/couchbase/sync_gateway/issues/1098 for dev notes.
 func (c *DCPCommon) updateSeq(vbucketId uint16, seq uint64, warnOnLowerSeqNo bool) {
-	if c.shouldProcessSequence(vbucketId, seq) {
+	if !c.shouldProcessSequence(vbucketId, seq) {
 		return
 	}
 	c.m.Lock()
--- a/base/dcp_common.go
+++ b/base/dcp_common.go
@@ -232,7 +232,8 @@ func (c *DCPCommon) shouldProcessSequence(vBucketID uint16, seq uint64) bool {
 	// DCP will provide mutations that run to the end of the snapshot that contains the end sequence number.
 	endSeq, ok := c.endSeqNos[vBucketID]
 	if !ok {
-		AssertfCtx(c.loggingCtx, "Received DCP event for vbno %d which is not tracked by the expected endSeqNos %#+v. This means that endSeqNos was specified with the incorrect number of vBuckets", vBucketID, c.endSeqNos)
+		AssertfCtx(c.loggingCtx, "Received DCP event for vbno %d which is not tracked by the expected endSeqNos %#+v. This means that endSeqNos was specified with the incorrect number of vBuckets. Processing this sequence anyway", vBucketID, c.endSeqNos)
+		return true
 	}
 	return seq <= endSeq
 }
--- a/base/dcp_common.go
+++ b/base/dcp_common.go
@@ -231,10 +231,8 @@ func (c *DCPCommon) updateSeq(vbucketId uint16, seq uint64, warnOnLowerSeqNo boo
 	if c.endSeqNos != nil {
 		endSeq, ok := c.endSeqNos[vbucketId]
 		if !ok {
-			AssertfCtx(c.loggingCtx, "Received DCP event for vbno %d which is greater than endSeqNos we have for that vbno", vbucketId)
-		}
-		// Checkpoints
-		if seq > endSeq {
+			AssertfCtx(c.loggingCtx, "Received DCP event for vbno %d which is not tracked by the expected endSeqNos %#+v. This means that endSeqNos was specified with the incorrect number of vBuckets", vbucketId, c.endSeqNos)
+		} else if seq > endSeq {
 			return
 		}
 	}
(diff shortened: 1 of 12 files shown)


base/dcp_common.go
```diff
@@ -60,37 +60,44 @@ type DCPCommon struct {
...
+	if c.endSeqNos != nil {
+		endSeq, ok := c.endSeqNos[event.VbNo]
+		if !ok {
+			AssertfCtx(c.loggingCtx, "Received DCP event for vbno %d which is greater than endSeqNos we have for that vbno", event.VbNo)
+		}
+		// Checkpoints
+		if seq > endSeq {
+			return
+		}
+	}
```
torcolvin: This code is duplicated in updateSeq, should be removed.

Consider whether we actually want to run the callback or not for sequences greater than endSeq

base/dcp_common.go
```diff
@@ -226,6 +224,19 @@ func (c *DCPCommon) updateSeq(vbucketId uint16, seq uint64, warnOnLowerSeqNo boo
 	c.m.Lock()
 	defer c.m.Unlock()
 
+	// Check the expected maximum sequence number when running a one shot feed. Do not checkpoint if the incoming
+	// sequence is greater than the expected maximum sequence number.
```
adamcfraser: Can we do this check earlier (in dataUpdate) and also avoid callback processing for any sequences higher than endSeqNo?

base/dcp_dest.go
```diff
@@ -30,6 +30,7 @@ func init() {
 type SGDest interface {
 	cbgt.Dest
 	cbgt.DestEx
+	ForceCheckpointWrite()
```
adamcfraser, line 33: As discussed, PR doesn't currently have any callers of ForceCheckpointWrite.  Review whether that's intentional.

db/background_mgr_resync_dcp.go
```diff
@@ -597,6 +644,44 @@ func (r *ResyncManagerDCP) GetProcessStatus(status BackgroundManagerStatus, prev
 	return statusJSON, metaJSON, err
 }
 
+// getUnregisterFeedFunc returns a callback function to be called when a cbgt feed exits. This function will close
+// doneChan when all vBuckets have completed, which will allow the resync process to finish.
+func (r *ResyncManagerDCP) getUnregisterFeedFunc(ctx context.Context, totalVBuckets uint16) base.CbgtUnregisterFeedCallback {
+	return func(feed cbgt.Feed) {
+		for _, d := range feed.Dests() {
```
adamcfraser: As discussed, don't think we have a use case where our feed will have multiple Dests, but to be defensive can make this code just look for *at least one* SGDest to use for the ForceCheckpoint call (and exit the loop when it finds it)

adamcfraser (changes requested):
Generally looks fine, a few comments to look at.

base/dcp_common.go
```diff
@@ -215,28 +218,37 @@ func (c *DCPCommon) persistCheckpoint(vbNo uint16, value []byte) error {
...
+	if c.endSeqNos == nil {
+		return true
+	}
+	// Check the expected maximum sequence number when running a one shot feed. Do not checkpoint if the incoming
+	// sequence is greater than the expected maximum sequence number.
+	//
+	// DCP will provide mutations that run to the end of the snapshot that contains the end sequence number.
+	endSeq, ok := c.endSeqNos[vBucketID]
+	if !ok {
+		AssertfCtx(c.loggingCtx, "Received DCP event for vbno %d which is not tracked by the expected endSeqNos %#+v. This means that endSeqNos was specified with the incorrect number of vBuckets", vBucketID, c.endSeqNos)
```
adamcfraser: In the case where the endSeq isn't found, it looks like the code will skip the mutation (endSeq will be zero for line 237).  Would it be more defensive to return true in that scenario?

adamcfraser (changes requested):
Changes look good - one minor question on handling the error case in shouldProcessSequence

adamcfraser (dismissed):

adamcfraser (approved):

PR #8267, opened 2026-05-13
CBG-5381: fix resync regenerate sequences set sync info no-op

gregns1:
CBG-5381

Fixes endless resync loop when moving collection from a standard metadata ID to a default metadata id. 

Notes on the fix: 

How `metadataID` now flows through the system with this fix across three possible distinct states:

### 1. Legacy (non-persistent) config
* **Source:** computeMetadataID is never called
* **config.MetadataID:** `""` (empty)
* **contextOptions.MetadataID:** `""` (empty)
* **NewMetadataKeys result:** `DefaultMetadataKeys` (unprefixed)
* **SetSyncInfoMetadataID arg:** `""` → **no-op** (correct: no syncInfo doc for legacy)
* **InitSyncInfo arg:** `""` → no resync required when doc missing, also doesn’t persist anything (correct: no syncInfo doc for legacy)

---

### 2. Persistent config (Default)
* **Source:** database includes `_default._default`
* **config.MetadataID:** `"_default"`
* **contextOptions.MetadataID:** `"_default"`
* **NewMetadataKeys result:** `DefaultMetadataKeys` (unprefixed) — *NewMetadataKeys now treats "_default" same as ""*
* **SetSyncInfoMetadataID arg:** `"_default"` → **writes syncInfo doc with _default metadata id**
* **InitSyncInfo arg:** `"_default"` → compares against stored value, writes one if not exists.

---

### 3. Persistent config (Named Collections)
* **Source:** standard ID
* **config.MetadataID:** `"mydb"`
* **contextOptions.MetadataID:** `"mydb"`
* **NewMetadataKeys result:** Prefixed keys (`m_mydb:...`)
* **SetSyncInfoMetadataID arg:** `"mydb"` → **writes syncInfo doc with `"mydb"` metadata id**
* **InitSyncInfo arg:** `"mydb"` → compares against stored value, writes one if not exists.

---

## Bug Analysis

The bug hit was **State 2** above:
1. `InitSyncInfo` was called with `"_default"` (from `config.MetadataID`).
2. `SetSyncInfoMetadataID` was called with `""` (from `Options.MetadataID`, which had the `"_default"` → `""` translation applied).
3. `SetSyncInfoMetadataID("")` is a no-op, so the `syncInfo` document was never updated after resync.
4. Every subsequent `InitSyncInfo("_default")` found a stale `metadataID` mismatch.

**The Fix:**
Now `options.MetadataID` is always `config.MetadataID` — no translation. Both `InitSyncInfo` and `SetSyncInfoMetadataID` receive `"_default"`, so the `syncInfo` document is correctly updated and subsequent startup finds a match. The `"_default"` → legacy unprefixed keys mapping is now handled entirely within `NewMetadataKeys`, which is the only consumer that needs it.

## Safety of effects on legacy config
Legacy config never calls `computeMetadataID`, so `config.MetadataID` remains `""`. The `""` value flows through identically to before: `NewMetadataKeys("")` returns `DefaultMetadataKeys`, `SetSyncInfoMetadataID("")` is a no-op, and `InitSyncInfo("")` identifies empty metadata id and doesn’t persist anything. **No behaviour change for legacy.**

## Pre-review checklist

- [ ] Link upstream PRs

- [x]  https://jenkins.sgwdev.com/job/SyncGatewayIntegration/622/

### DIFF
--- a/base/constants_syncdocs.go
+++ b/base/constants_syncdocs.go
@@ -22,6 +22,8 @@ const MetadataIdPrefix = "m_"                                  // Prefix for met
 const SyncDocMetadataPrefix = SyncDocPrefix + MetadataIdPrefix // Prefix for all namespaced Sync Gateway metadata documents
 const DCPCheckpointPrefix = "dcp_ck:"
 
+const DefaultMetadataID = "_default" // MetadataID assigned to databases that include _default._default, signalling legacy (unprefixed) metadata key format.
+
 const minimumAttachmentMigrationMetadataVersion = "4.0.0" // minimum metadata version that needs to be defined for metadata migration.
 
 // Sync Gateway Metadata document types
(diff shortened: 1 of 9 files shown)


bbrks (approved):
LGTM 👍

